PR #4882, opened 2016-01-30
core: update doc after auto executing in micro state.

levindu:
This is useful if :doc is a function to return the state which may be
modified by :execute-binding-on-enter.

### DIFF
--- a/core/core-micro-state.el
+++ b/core/core-micro-state.el
@@ -114,13 +114,13 @@ used."
               ,(format "%S micro-state." name)
               (interactive)
               ,@on-enter
+              ,(when exec-binding
+                 (spacemacs//micro-state-auto-execute bindings))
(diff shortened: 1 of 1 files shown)


robbyoconnor:
Now we're good :heart:

nixmaniack:
I doubt we're using this in `develop` anymore since the migration to `hydra` based `transient-state`

StreakyCobra:
@nixmaniack Indeed, let's ask @justbur as he worked on this recently :-)

justbur:
Except in one case the micro-state macro is not being used anymore

syl20bnr:
Indeed micro-state are almost a thing from the past, transient-state FTW!
I still cherry-picked this contribution in the mean time (may push it to master in next release).

Thank you :+1:

PR #4823, opened 2016-01-26
Fix layouts transient state doc and tweak formatting of hint 

justbur:

PR #4883, opened 2016-01-30
Change all references of micro-state in codebase to transient-state

balajisivaraman:
The term micro-state was still being referred to all over the codebase. So I've created this PR in-order to change all of them to transient-state in order to ensure consistency, and also for feedback in case I screw something up. :smiley: 

- [x] Change all `.org` files in `doc` folder to refer to transient states
- [x] Change all layer-specific `README` files to refer to transient states
- [ ] There are some variables across the codebase (For eg: some `faces` vars) that refer to micro-states. I can change these as well if it's OK with you folks.
- [ ] There are places such as `ido` and `workspaces` that refer to micro-states, even though they have been changed to transient states. These might need to be verified to ensure they are not bugs.
- [ ] Magit-blame mode is still a micro-state (only one left). So I've not changed anything related to it as of now.
- [ ] Right now the screenshot files used have simply been renamed. But they should probably be taken anew as well. I'll do that.

I just noticed #4823 updates the workspaces related docs. So I'll change them back to the way they were so that the other PR can be merged.


syl20bnr:
Thank you for the PR :+1: 
I preferred to merge it before waiting for all the check boxes to avoid future merge conflicts.
I created an issue with the todo list here: https://github.com/syl20bnr/spacemacs/issues/4913

PR #4874, opened 2016-01-30
Ensure that `pcache-directory` ends in a slash

rpglover64:
Otherwise, `pcache` doesn't seem to function, at least in the case of `unicode-fonts`; see https://github.com/rolandwalker/unicode-fonts/issues/20.

PR #4885, opened 2016-01-30
Ensure that `pcache-directory` ends in a slash

rpglover64:
Resubmit of #4874, against `develop` this time.

Otherwise, `pcache` doesn't seem to function, at least in the case of `unicode-fonts`; see https://github.com/rolandwalker/unicode-fonts/issues/20.


syl20bnr:
Nice catch, thank you!

Issue #4386, opened 2016-01-04
Change default value for dotspacemacs-whitespace-cleanup


syl20bnr:
We should change it to `nil`, it can be surprising for some users.
Source: https://www.reddit.com/r/emacs/comments/3ze6gf/spacemacs_0105_released_new_website_and_layouts/cylowm2

StreakyCobra:
:cry: A soooooo sane default gone

xfq:
It also needs better documentation.  For example, what are the differences between `all` and `trailing`?

Maybe I should report this as another issue.

echohack:
Gonna weigh in on this one. This was a poor change IMO -- hunting down this default and changing it was not a delightful experience. As a new user, this not being enabled was surprising.

VS Code has this enabled by default, for instance.

<img width="593" alt="Screen Shot 2019-03-19 at 22 03 26" src="https://user-images.githubusercontent.com/3253989/54660252-e76ef500-4a92-11e9-98ca-3691a9bc4477.png">

Compro-Prasad:
@echohack For your kind information, not all projects have linters that remove unneeded whitespace but most of them use a version control software. Providing this default will produce large diffs which is not good practise.

PR #4889, opened 2016-01-30
Ensure trailing newline when saving

fbergroth:


robbyoconnor:
No. Make this an option, I hate this.

fbergroth:
I think this should be the default. No need for an option that just wraps a variable, users can just change the value in their dotfile.

robbyoconnor:
There's something missing here then...How are users supposed to know to change this?

fbergroth:
I think most users would expect this behaviour. A lot of the stuff in `spacemacs-base/config.el` seems to be sane defaults.

We can't possibly document every emacs setting in spacemacs. :)

robbyoconnor:
I am saying to **DOCUMENT IT!**

fbergroth:
Troll somewhere else.

StreakyCobra:
I don't think it's a better default to set it to true in all cases. People should be wanting not to have newline on some text files, and they will except no modifications on parts of the files they have not touched when editing existing files: this goes against POLA, especially as it can easily be unnoticed. We should probably document in the FAQ how to enable it instead, IMHO.

robbyoconnor:
@fbergroth and I wasn't trolling you, I was telling you to document it. Fucks sake.

robbyoconnor:
@fbergroth and I wasn't trolling you, I was telling you to document it. Fucks sake. If you add something like this, always document it.

StreakyCobra:
Oh, calm down @robbyoconnor, this kind of saying shouldn't exist at all in the Spacemacs community.

@fbergroth The fact that I don't agree doesn't mean it can't be discussed or that other developers will not like it :smiley: Wait for other people opinion, reopen it if you feel so :relaxed:

robbyoconnor:
Okay, rephrase: @fbergroth: I wasn't trolling you, I was telling you to document it...that's what makes spacemacs so great, it's heavily documented.

robbyoconnor:
Clearing out any previous objections.

fbergroth:
Just to be clear, this setting ensures that the final line in a text document ends with a newline character, as per the [POSIX definition](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206).

syl20bnr:
I agree with @StreakyCobra. We cannot set a default that modifies the buffer of a user behind a scene. This kind of setting must be explicitly opt-in by the user.

Maybe this PR can be modified to add an HOW-TO instead ?

fbergroth:
Alright. I'm not really interested in writing documentation (I never read it anyway, I prefer reading the code), so I'm closing this.

I'm not a fan of documenting basic emacs settings in spacemacs, I think it would be more worthwhile to document `Info-mode` and how to read/search the emacs info manual.

StreakyCobra:
@fbergroth I'm not really sure, but doesn't `ws-butler` already [takes care of this](https://github.com/lewang/ws-butler/blob/master/ws-butler.el#L89)? If it is the case it is already nice because it let the user decide if it should be done on the whole file, on changed lines only, or not done at all (see `dotspacemacs-whitespace-cleanup`).

fbergroth:
@StreakyCobra [it seems](https://github.com/lewang/ws-butler/blob/a998a23ee8713808ac1fe3d1523ea1861be4da57/ws-butler.el#L107-L109) to only do it if `require-final-newline` is `t`?

StreakyCobra:
Indeed, I was thinking beyond this and I thought it would be nice to have this done only when the ending line is modified, like what can be done for trailing whitespaces removal with `'changed`, hence my `ws-butler` link. But it doesn't solves the problem at all of course. And even, when trailing spaces where removed by default *only on changed lines*, users where complaining about it (see #4386) :pensive:

kini:
I think the POSIX definition of "line" cited by @fbergroth is a pretty convincing argument that we're not "modifying the buffer of a user behind the scenes" as @syl20bnr said. According to POSIX, every line just ends with a \n. The reason there's no empty line shown at the end of the buffer is because that empty line would also have to end with \n. So, without an empty line, the file ends with a \n character, and with an empty line, the file ends with \n\n.

Also, wasn't the default evil behavior to always ensure a trailing newline character in files? It's definitely the default behavior in vim.

PR #4813, opened 2016-01-26
mu4e: use helm-comp-read

jeizsm:
helm-completing-read doesn't exist

PR #4894, opened 2016-01-31
Fix nonexistent helm completing read function

darkfeline:
(Please check that a function exists before using it.)

### DIFF
--- a/layers/+email/mu4e/packages.el
+++ b/layers/+email/mu4e/packages.el
@@ -37,7 +37,7 @@
       (setq mu4e-completing-read-function
             (if (configuration-layer/layer-usedp 'spacemacs-ivy)
                 'ivy-completing-read
-              'helm-completing-read))
+              'helm--completing-read-default))
 
       (add-to-list 'mu4e-view-actions
                    '("View in browser" . mu4e-action-view-in-browser) t)


layers/+email/mu4e/packages.el
```diff
@@ -37,7 +37,7 @@
       (setq mu4e-completing-read-function
             (if (configuration-layer/layer-usedp 'spacemacs-ivy)
                 'ivy-completing-read
-              'helm-completing-read))
+              'helm--completing-read-default))
```
nixmaniack: The doc for `helm--completing-read-default` says don't use this directly but instead use `helm-comp-read`. I haven't tested this but just pointing out if you've missed.

darkfeline:
~~~
This function should be used only as a `completing-read-function'.
~~~

In this case, I'm using it as a `completing-read-function`.

~~~
Don't use it directly, use instead `helm-comp-read' in your programs.
~~~

This is talking about packages and functions using helm directly for completing read.

Or the cop-out answer: `helm--completing-read-default` works and `helm-comp-read` doesn't.

nixmaniack: I see. Didn't read between the lines. Thanks for explaining. :smile:

StreakyCobra:
Changing this is already proposed in #4813. I had the same reflex has @nixmaniack to ask replacing it with `helm-comp-read` there, as the documentation says. If you said it's not working then I'll ask there to change it back. I think we have a first-come first-served philosophy here, so the other PR should probably be merged :-)

TheBB:
Closing this since I merged the older PR #4813. Thanks, though!

PR #4904, opened 2016-01-31
support package archive on local disk

emacs18:
This change avoids error that would result if package archives are located on local disk, i.e., if package-archives specifies local directories rather than URLs.

Without this change I'm forced to URL such as http://localhost/.
Yes this avoids internet access if URL points to a host on local intranet.
However using local disk rather than URL is much much faster.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -243,7 +243,8 @@ refreshed during the current session."
                                     "\nError while contacting %s repository!"
                                     (car archive)))
                                   'error))
-                             (url-retrieve-synchronously (cdr archive))))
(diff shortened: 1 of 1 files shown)


syl20bnr:
Unfortunately it breaks the fetching of the archives, Emacs tries to ssh first and output an error.

emacs18:
I changed the patch so that rather than calling file-directory-p, the path is parsed to see if it contains hostname component.

syl20bnr:
@emacs18 Thank you for the PR, your repository is gone so I cannot merge it, please reopen a new one if you want. Sorry for this.

PR #4914, opened 2016-02-01
Add ob-ipython to ipython-notebook layer

AlejandroCatalina:
I think having this package would improve `ipython-notebook` layer for Org. I wasn't sure if adding it to `ipython-layer` was the right choice, instead of adding it to org-layer. It seemed to belong here better, thought, and I think the sml layer also added `ob-sml` to its layer and not to org-layer, so I left it here.

Suggestions and corrections/improvements be welcome!


StreakyCobra:
There is another example in the code and it is [also placed in the topic layer](https://github.com/syl20bnr/spacemacs/blob/18874c9a3eee11c82528d105dacec9034c5432e1/layers/%2Blang/sml/packages.el#L58-L62).

AlejandroCatalina:
Yup I meant that, seems fine then.

izahn:
I would very much like to see ob-ipython added to this layer. I've cleaned up merge conflicts with the develop branch and made some small improvements in https://github.com/izahn/spacemacs/tree/layers/add-ob-ipython. @AlejandroCatalina do you want to clean this up this PR or should I submit a new one from my version of your `layers/add-ob-ipython` branch?

AlejandroCatalina:
Cleaned! Thank you for notifying me, I've been very busy these months, but now I'm up again ;)

layers/+lang/ipython-notebook/packages.el
```diff
@@ -157,4 +158,11 @@
         ("9" ein:notebook-worksheet-open-last)
         ("+" ein:notebook-worksheet-insert-next)
         ("-" ein:notebook-worksheet-delete)
-        ("x" ein:notebook-close)))))
+        ("x" ein:notebook-close))
+      (spacemacs/set-leader-keys "ein" 'spacemacs/ipython-notebook-transient-state/body))))
+
+(defun ipython-notebook/init-ob-ipython ()
+  (use-package ob-ipython
+    :defer t
```
izahn: If this is deferred here what causes it to be loaded? I tested it out and it seems like ob-ipython is not loaded when visiting a `.org` buffer, but has to be explicitly `require`d. Should we remove `:defer t` from this config?

AlejandroCatalina: I think `:defer t` causes it to not be loaded upon start time but when the layer is loaded. This worked on my setup, isn't it working anymore?

izahn: Oops, my mistake. I had some other settings that were causing problems. Sorry for the noise and thanks again for pushing this forward. I hope we get it into spacemacs soon.

syl20bnr:
Thank you ! 👍 

PR #4918, opened 2016-02-01
auto-complete: Better binding for hippie-expand

justbur:
Using a remap is better, because it will "automatically toggle" the key
binding if hybrid-mode is toggled.

The previous version just disabled the key binding completely if someone
used hybrid style.

@TheBB  this is a small improvement on @nixmaniack's last commit.


TheBB:
:+1:

PR #4921, opened 2016-02-01
Open restclient with a keybinding

vise890:
It seemed like opening `restclient` with a shortcut namespaced under `applications` would be useful.

Here's a PR implementing it. What do you think?

### DIFF
--- a/layers/restclient/README.org
+++ b/layers/restclient/README.org
@@ -22,6 +22,7 @@ file.
 
 | Key Binding | Description                                                 |
 |-------------+-------------------------------------------------------------|
+| ~SPC a h~   | Switch to a new buffer in restclient-mode                   |
 | ~SPC m s~   | Send and stay in window (pretty-print response if possible) |
 | ~SPC m S~   | Send and switch window (pretty-print response if possible)  |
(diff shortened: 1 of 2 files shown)


TheBB:
:+1: for me, but this is about the twentieth PR I see trying to take `SPC a r`, which is [taken](https://github.com/TheBB/spacemacs/blob/develop/layers/+tools/ranger/packages.el#L21). You'll have to find something else.

vise890:
Aaah of course.

Ok, what about `h` (as in `http`) ?

Compro-Prasad:
Seems nice @vise890 .

syl20bnr:
I think it is not necessary, you can `SPC f J http RET` or use any kind
of jump facility to existing .http file.

vise890:
The only limitation I can see in the file jump trick is that you to have a `.http` file. Either way, I'm super happy for this to be closed if there's no interest :smile:

syl20bnr:
As any file with `.http` extension will have restclient or org (ob-http) activated I don't feel this shortcut is need. Thank you for the PR though and I hope it won't prevent your from contributing again to Spacemacs 💜

PR #4925, opened 2016-02-02
Remove `gnus-fetch-old-headers` customization

### LINKED ISSUE #1331: Reading gmane groups really slow with gnus layer.
This is due to setting gnus-fetch-old-headers to t.
Its help says:
> This feature can seriously impact performance it ignores all
> locally cached header entries.  Setting it to t for groups for a
> server that doesn't expire articles (such as news.gmane.org),
> leads to very slow summary generation.

Setting it nil fixes the issue. Please consider not setting it to t by default.

StreakyCobra:
Fix #1331

@andreas-h @amardeep (and maybe @cpaulik) You are the only users of `gnus` that I know. Do you agree on this consensus, discussed in the cited issue?

### DIFF
--- a/layers/+email/gnus/packages.el
+++ b/layers/+email/gnus/packages.el
@@ -48,7 +48,6 @@
        gnus-sum-thread-tree-single-leaf "╰► "
        gnus-sum-thread-tree-vertical "│"
        gnus-article-browse-delete-temp t
-       gnus-fetch-old-headers t
        gnus-treat-strip-trailing-blank-lines 'last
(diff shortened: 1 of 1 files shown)


cpaulik:
I don't use it anymore but from what I know it seems reasonable.

gour:
Here is new Gnus user and I'd like to have smooth Gmane experience. :wink:

StreakyCobra:
Thank you :+1: Che… bla bla bla

PR #4932, opened 2016-02-02
fix paste micro -> paste transient

nixmaniack:


TheBB:
Thanks! Cherry-picked.

PR #4933, opened 2016-02-02
Enable emmet-mode in sass and scss modes.

utkarshkukreti:
emmet-mode handles detecting the major-mode and inserting the correct expansion for the major-mode automatically.


TheBB:

PR #4935, opened 2016-02-02
Fix `0` as prefix arg in paste transient state

### LINKED ISSUE #4930: Pressing any number key doesn't exit pasting transient state
#### Description
Pressing any number key doesn't exit pasting transient state.

#### Reproduction guide
- Start Emacs
- Paste something with `p`
- Press `0`

*Observed behaviour:*
Nothing happens.

*Expected behaviour:*
Point should move to the beginning of line.

#### System Info
- OS: darwin
- Emacs: 24.5.1
- Spacemacs: 0.105.9
- Spacemacs branch: develop (rev. 18874c9)
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(osx lispy spacemacs-helm emacs-lisp syntax-checking common-lisp git github
     (shell :variables shell-default-shell 'eshell)
     spacemacs-layouts erc html pdf-tools
     (elfeed :variables rmh-elfeed-org-files
             (list "/Users/andre0991/Dropbox/backup/emacs/elfeed-feeds.org"))
     themes-megapack semantic
     (auto-completion :variables auto-completion-private-snippets-directory "/Users/andre0991/Dropbox/backup/emacs/yasnippet-snippets/" auto-completion-return-key-behavior nil auto-completion-tab-key-behavior 'complete auto-completion-enable-help-tooltip t auto-completion-enable-sort-by-usage t auto-completion-enable-snippets-in-popup t :disabled-for latex)
     org
     (ranger :variables ranger-show-preview t)
     markdown latex spell-checking version-control andre-common-lisp andre-elfeed andre-osx andre-erc andre-eww)
```

Thanks.

AlejandroCatalina:
When in paste-transient-state, 0 won't exit the transient state and so
it won't take the point to the beginning of line.

I don't know if this is the best idiomatic way to do it @justbur .

Fix #4930

### DIFF
--- a/layers/+distribution/spacemacs-base/packages.el
+++ b/layers/+distribution/spacemacs-base/packages.el
@@ -341,6 +341,18 @@ Example: (evil-map visual \"<\" \"<gv\")"
         "n>" 'spacemacs/scroll-transient-state/spacemacs/scroll-half-page-down)
 
       ;; pasting transient-state
+      (evil-define-command spacemacs//transient-state-0 ()
+        :keep-visual t
+        :repeat nil
+        (interactive)
+        (if current-prefix-arg
+            (progn
+              (setq this-command #'digit-argument)
+              (call-interactively #'digit-argument))
+          (setq this-command #'evil-beginning-of-line
+                hydra-deactivate t)
+          (call-interactively #'evil-beginning-of-line)))
+
       (spacemacs|define-transient-state paste
         :title "Pasting Transient State"
         :doc "\n[%s(length kill-ring-yank-pointer)/%s(length kill-ring)] \
@@ -350,7 +362,8 @@ below. Anything else exits."
         ("J" evil-paste-pop)
         ("K" evil-paste-pop-next)
         ("p" evil-paste-after)
-        ("P" evil-paste-before))
+        ("P" evil-paste-before)
+        ("0" spacemacs//transient-state-0))
       (when dotspacemacs-enable-paste-transient-state
         (define-key evil-normal-state-map "p" 'spacemacs/paste-transient-state/evil-paste-after)
         (define-key evil-normal-state-map "P" 'spacemacs/paste-transient-state/evil-paste-before))


justbur:
I would add it to the transient-state definition itself in this case, but otherwise the idea is fine. If the digit keys are a more general problem we can come up with an option to disable them

AlejandroCatalina:
I don't think they're a problem *in general*, but in this case pasting *0* times something doesn't make sense. 

Other thing to note in this solution is that we won't be able to type `0` in any prefix argument sequence; if you type `110` as the argument to `p`, it will break at `0`. Anything about it?

justbur:
Try this
```elisp
      (unless (eq dotspacemacs-editing-style 'emacs)
        (defvar spacemacs-paste-transient-state-add-bindings nil)
        (push '("0" evil-digit-argument-or-evil-beginning-of-line :exit t)
              spacemacs-paste-transient-state-add-bindings))
```

AlejandroCatalina:
Almost. I've been trying with it and the problem is that `:exit t` will make `0` exit the transient state, although the `0` is behaving as a prefix argument. If only there were a way to pass `:exit (exp)`...

Anyway, your expression works pretty good as after entering `10` you can reenter the paste-transient-state by typing `p`, so it might be enough. Is not an option modifying hydra to evaluate the parameter to `:exit` right?

justbur:
Not that I know of. you want 0 to exit if it comes first but be a digit arg otherwise and not exit?

AlejandroCatalina:
Exactly that.

justbur:
Here you go
```elisp
      (evil-define-command spacemacs//transient-state-0 ()
        :keep-visual t
        :repeat nil
        (interactive)
        (if current-prefix-arg
            (progn
              (setq this-command #'digit-argument)
              (call-interactively #'digit-argument))
          (setq this-command #'evil-beginning-of-line
                hydra-deactivate t)
          (call-interactively #'evil-beginning-of-line)))
```
Use this without `:exit`

AlejandroCatalina:
That's it, like a charm, I'm committing.

TheBB:

PR #4937, opened 2016-02-02
Delay `vi-tilde-fringe` loading after display init

### LINKED ISSUE #1303: no tildes in the fringe after EOF when using emacs-client
when using emacs-client instead of emacs there are no tildes in the fringe after the end of file. That's with everything else configured exactly the same.

StreakyCobra:
Fix #1303


syl20bnr:
Nice fix :+1: 
Thank you!

PR #4944, opened 2016-02-03
Add more doc to the underscores word motion.

franciscoj:
I tried to change the word motion to include underscores without success by using the provided example.

I added a couple of examples more and changed the example to be a more specific on where to put the code.

### DIFF
--- a/doc/FAQ.org
+++ b/doc/FAQ.org
@@ -264,12 +264,16 @@ binding in the mode's map, e.g. for =magit-status-mode=,
 #+END_SRC
 
 ** Include underscores in word motions?
-You can modify the syntax table of the mode in question. For example, for Python
-mode:
+You can modify the syntax table of the mode in question. To do so you can
+include this on your =dotspacemacs/user-config=.
 
 #+BEGIN_SRC emacs-lisp
-(with-eval-after-load 'python
-  (modify-syntax-entry ?_ "w" python-mode-syntax-table))
+;; For python
+(add-hook 'python-mode-hook #'(lambda () (modify-syntax-entry ?_ "w")))
+;; For ruby
+(add-hook 'ruby-mode-hook #'(lambda () (modify-syntax-entry ?_ "w")))
+;; For Javascript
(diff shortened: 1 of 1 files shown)


doc/FAQ.org
```diff
@@ -268,8 +268,13 @@ You can modify the syntax table of the mode in question. For example, for Python
 mode:
 
 #+BEGIN_SRC emacs-lisp
-(with-eval-after-load 'python
-  (modify-syntax-entry ?_ "w" python-mode-syntax-table))
+;; In your dotspacemacs/user-init()
```
StreakyCobra: I think this comment should be placed out of the code block, in the org text. You can modify the sentence just above to remove the explicit python mention, and add there the information where it should be placed. Also this would probably better be placed in `dotspacemacs/user-config` (I haven't tested it though)

franciscoj: :ok_hand: 

About placing it on `dotspacemacs/user-config` I tried and couldn't make it work but my understanding of emacs and spacemacs is very small so didn't understand why :(

doc/FAQ.org
```diff
@@ -264,12 +264,16 @@ binding in the mode's map, e.g. for =magit-status-mode=,
 #+END_SRC
 
 ** Include underscores in word motions?
-You can modify the syntax table of the mode in question. For example, for Python
-mode:
+You can modify the syntax table of the mode in question. To do so you can
+include this on your =dotspacemacs/user-init=.
```
TheBB: Idiomatically it should be in `dotspacemacs/user-config`.

franciscoj: You're both right. I just tried it again  and it worked on `dotspacemacs/user-config`. I'll fix the doc. Thanks!

franciscoj:
@TheBB @StreakyCobra seems better now? Shall I sqash the commits or is it OK as it is?

StreakyCobra:
> Shall I sqash the commits or is it OK as it is?

If you know how to do it, yes, it will simplify our work :-)

franciscoj:
Done

StreakyCobra:
:+1: Thx

StreakyCobra:

Issue #4120, opened 2015-12-09
emacsclient -nw reuses window layout from another frame


agriffis:
This is a problem I'm seeing on develop which I wasn't seeing on master.

Consider two terminals. The first is running `emacsclient -nw` with three files:

![screenshot from 2015-12-09 07-50-44](https://cloud.githubusercontent.com/assets/50637/11686061/e5a5ad00-9e4b-11e5-9e76-3a82ec3e14e6.png)

In the second terminal, we run `emacsclient -nw file4.txt`. We expect this should show *only* the requested file, but instead it shows the same layout as the first frame, substituting the most recently used file with the new one. This is easiest to see in a screencast:

![out](https://cloud.githubusercontent.com/assets/50637/11686090/1ceccdac-9e4c-11e5-844a-0f6420d7d0a7.gif)

This is pretty much driving me nuts!

StreakyCobra:
It is due to `spacemacs-layout`. This also happen on GUI mode with emacsclient. In short layouts are used to keep and restore windows positions and buffers. There is a default layout, let's call it `Home`, that contains all buffers and remember the windows positions. When you open one frame with `emacsclient`, the default layout is selected and when you split the screen or open a buffer, the state is stored. When you open a second one with emacsclient, the same layout is **restored** with its positions and buffers, and then it probably replace the first buffer with the file passed as argument.

So I would say it's "as expected", but there is probably some enhancement to do here. @CestDiego @syl20bnr Any idea how to deal with emacsclient and layouts?

agriffis:
@StreakyCobra But the problem is that the first emacsclient is still open. There are two things I think we know in this case: (1) If the first emacsclient is still open, we probably don't want to reuse a layout that's still in use. (2) If we explicitly mention file(s) on the command-line, we probably don't want to reuse a previous layout anyway.

(I'm suspicious that I don't want `spacemacs-layout` anyway, but that's a different story.)

agriffis:
By the way, I don't have `(setq-default dotspacemacs-configuration-layers '(spacemacs-layouts))` in my `.spacemacs` but it seems to be loading anyway.

StreakyCobra:
[...]

PR #4946, opened 2016-02-03
persp-mode: switch back to melpa

bmag:
I don't understand the problem that caused us to switch to a fork, but the switch was supposed to be temporary and the problem is claimed to be fixed upstream. In the meantime there were a few more commits upstream, so let's switch back.
I didn't experience any errors switching back to upstream, but I haven't tested thoroughly either.

Refs:
https://github.com/Bad-ptr/persp-mode.el/pull/25
https://github.com/syl20bnr/spacemacs/issues/4120

### DIFF
--- a/layers/+window-management/spacemacs-layouts/packages.el
+++ b/layers/+window-management/spacemacs-layouts/packages.el
@@ -9,11 +9,7 @@
 ;;
 ;;; License: GPLv3
 (setq spacemacs-layouts-packages
-      '(;; temporary switch on a fork to fix
-        ;; https://github.com/syl20bnr/spacemacs/issues/4120
-        (persp-mode :location (recipe :fetcher github
-                                      :repo "syl20bnr/persp-mode.el"
-                                      :branch "fix-emacsclient-crash"))
+      '(persp-mode
         spaceline
         eyebrowse
         helm


TheBB:
:+1: I wanted to ask this too.

syl20bnr:
The bug is still present, to reproduce it:
- open emacs in GUI mode
- open a new terminal and type `emacsclient -t toto.txt`

The client crashes.

When you changed the package list did you delete the persp-mode directory in `elpa` to be sure to retrieve the actual package from melpa ?

bmag:
@syl20bnr ok, I see the client crash here as well, and in the \*Messages\* buffer (accessible in the GUI):
> [persp-mode] Warning: Could not restore window configuration, because of the error -- (error Window #<window 11 on persp-mode.el> too small to accommodate state)

And I can't quit with `SPC q q` or `C-x C-c` because of:
> apply: Device 1 is not a termcap terminal device

This is with persp-mode-20160215.807, Ubuntu 15.10, bash and gnome-terminal.

syl20bnr:
The quit error is still here even with my crappy fork (which I would like to get ride of :-)).

bmag:
When (if?) we switch back to melpa, we'll need to merge #5229

PR #4947, opened 2016-02-03
Fix name of google-translate's init function

bmag:
`spacemacs/init-google-translate` should be `spacemacs-language/init-google-translate`, `google-translate` is not loaded because of this.


StreakyCobra:
Nice catch. Thank you :+1: Cherry-picked into `develop`. You can safely delete your branch.

Issue #4945, opened 2016-02-03
Loading Racket will destroy company.


hujianxin:
When I startup Emacs and check into a c file for a python file, company works well. But, once I open a Racket file, company will stop working, even in other file.

And the error messages is:
```
Company: An error occurred in post-command
Company: frontend company-pseudo-tooltip-frontend error "Wrong type argument: number-or-marker-p, nil" on command post-command
```

hujianxin:
I deleted racket-mode and company-mode file and restart emacs, this issue reproduced.

hujianxin:
well
#### System Info
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.9
- Spacemacs branch: develop (rev. 8c8da65)
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
(spacemacs-ivy auto-completion better-defaults git github
               (shell :variables shell-default-height 30 shell-default-position 'bottom)
               spell-checking syntax-checking version-control dockerfile puppet django org markdown emacs-lisp
               (c-c++ :variables c-c++-enable-clang-support t)
               common-lisp ess extra-lang html javascript java latex python racket scala scheme sql shell-scripts yaml command-log dash pandoc ycmd evil-cleverparens evil-commentary evil-snipe unimpaired vim-empty-lines vinegar eyebrowse spacemacs-layouts colors cscope deft gtags semantic smex speed-reading themes-megapack)
```

greghendershott:
Although I'm not able to test with spacemacs right now, I did confirm that company-mode and latest racket-mode are still working fine with each other on Emacs 24.5.1 on OS X and Emacs 25.0.50 on Ubuntu. I tried both `global-company-mode` and plain `company-mode`.

---

Using <kbd>C-h f</kbd> I searched for `company-pseudo-tooltip-frontend` from the error message you shared. This seems to be the company-mode code that shows the "popup" completion list. `"Wrong type argument: number-or-marker-p, nil" on command post-command` sounds like a problem with a buffer position. Since I can't repro the problem, I can't tell where. But in any case a buffer position isn't a value that would come directly from `racket-mode`, AFAICT. At the moment it doesn't _seem_ like it's a racket-mode issue I could address.

StreakyCobra:
@greghendershott Thanks for your feedback. If there is no bug on a stock Emacs, then it's probably a problem on our end.

[...]

PR #4951, opened 2016-02-04
Disable company-quickhelp-mode only when it is already enabled

syohex:
This is related to #4945.
CC: @hujianxin @greghendershott

`(company-quickhelp-mode -1)` restores `company-tooltip-minimum-width` value, but when `company-quickhelp-mode` is never enabled,  `company-tooltip-minimum-width` is set `nil`. Its type must be integer, so #4945 error is raised. I think this issue is reproduced only on `emacs -nw`,  `company-quickhelp-mode` is not enabled on `emacs -nw`

### DIFF
--- a/layers/+lang/racket/packages.el
+++ b/layers/+lang/racket/packages.el
@@ -16,7 +16,8 @@
   ;; company-quickhelp calls it. Note hook is appendended for proper ordering.
   (add-hook 'company-mode-hook
             '(lambda ()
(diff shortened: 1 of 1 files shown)


hujianxin:
@syohex Thanks to syohex, you are my hero!

StreakyCobra:
:+1: Nice catch @syohex, thanks!

syl20bnr:
Thank you :+1: 

PR #4954, opened 2016-02-04
prevent auto-highlight-symbol from setting up its own default map

bling:
when `auto-highlight-symbol` activates, it sets up `M-S-<left>` and `M-S-<right>` as keybindings, which is a big no-no for anyone that uses org-mode.

since spacemacs already sets up the microstate, it's harmless to disable the original.


nixmaniack:
:+1:
This branch has conflicts!

syl20bnr:
I think it is OK to do this, thank you again :-)

PR #4957, opened 2016-02-04
scala: make some keybindings work in emacs editing style

jerrypnz:
Some of the keybindings in the scala layer are defined with
`evil-define-key` for `insert` state. Since evil mode is no longer
activated in emacs editing style, these keybindings don't work.

### DIFF
--- a/layers/+lang/scala/packages.el
+++ b/layers/+lang/scala/packages.el
@@ -32,10 +32,9 @@
     (progn
       (setq user-emacs-ensime-directory ".cache/ensime")
 
-      (evil-define-key 'insert ensime-mode-map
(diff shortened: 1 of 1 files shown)


syl20bnr:
Putting this PR on hold, it may be unnecessary since I plan to put back evil in holy mode.

moonranger:
:+1:

rgrinberg:
@moonranger @syl20bnr Can this be closed? I'm not entirely sure what holy mode is, but it seems like it uses evil again.

jerrypnz:
Sorry I've not been using spacemacs for a while and I'm not sure if this change is still relevant. I'll close it for now.

PR #4960, opened 2016-02-04
Show function signature for Golang buffers.

### LINKED ISSUE #4368: improvement: go layer autocomplete must show function signature
Hi guys, I think than would be cool than the autocomplete in the go layer would show the signature (expected arguments and result for every function) inside the box or in other windows (similar to how ternjs works) 

in my emacs config this works and show the function signature but I don't know why the spacemacs layer is  different, I suppose than must be a configuration in the layer because scala with ensime works as expected.

thanks!!!...

carlosgaldino:
This fixes #4368.

### DIFF
--- a/layers/+lang/go/packages.el
+++ b/layers/+lang/go/packages.el
@@ -100,7 +100,9 @@
       :if (configuration-layer/package-usedp 'company)
       :defer t
       :init
-      (push 'company-go company-backends-go-mode))))
+      (progn
+        (setq company-go-show-annotation t)
+        (push 'company-go company-backends-go-mode)))))
 
 (defun go/init-go-oracle()
   (let ((go-path (getenv "GOPATH")))


layers/+lang/go/packages.el
```diff
@@ -93,7 +93,11 @@
 
 (when (configuration-layer/layer-usedp 'auto-completion)
   (defun go/post-init-company ()
-    (spacemacs|add-company-hook go-mode))
+    (spacemacs|add-company-hook go-mode)
+    (add-hook 'go-mode-hook
+              (lambda ()
+                (setq-local company-go-show-annotation t)
```
fbergroth: `company-go-show-annotation` is part of the `company-go` package, it should be set there.

carlosgaldino: In their repo? Or on `go/init-company-go` in this same file?

fbergroth: Right, in `go/init-company-go` . Replace the `:init` section with
```el
(progn
  (setq company-go-show-annotation t)
  (push 'company-go company-backends-go-mode))
```

carlosgaldino: Changed. Thank you.

fbergroth:
:+1: although `company-tooltip-align-annotations` can be discussed. I personally prefer having it set to `nil` for the golang backend.

carlosgaldino:
@fbergroth Yeah, you're right. It is much better not aligned. I removed it.

syl20bnr:
Thank you !

PR #4963, opened 2016-02-04
Add a layer for nginx config files

waisbrot:
Simple [nginx-mode](https://github.com/ajc/nginx-mode), for editing an nginx config file.

Is this an appropriate layer to add (given the simplicity) or would it be better to give Spacemacs some way of creating simple layers (e.g. you'd list `nginx` in your `dotspacemacs-configuration-layers` and when it didn't find an appropriate layer it would try `(use-package nginx)` and `(use-package nginx-mode)`?


layers/+config-files/nginx/packages.el
```diff
@@ -0,0 +1,18 @@
...
+;;
+;;; License: GPLv3
+
+(defconst nginx-packages
+  '(
+    nginx-mode
+    ))
+
+(defun nginx/init-nginx-mode ()
+  (use-package nginx-mode))
```
AlejandroCatalina: As a general recommendation, every package should be deferred, so adding `:defer t` would be better in my opinion. Unless it is strictly necessary to load it at startup time.

Let's hear other's opinions as well.

waisbrot: I had no particular reason for not setting `:defer t` beyond a habit of taking the defaults for everything that I don't have a reason to set to a non-default value.

AlejandroCatalina: Well, the reason here would be spacemacs conventions, if i am not
mistaken. @StreakyCobra am i right?

Nathaniel Waisbrot writes:

>> +;; Copyright (c) 2012-2016 Sylvain Benner & Contributors
>> +;;
>> +;; Author: Nathaniel Waisbrot <code@waisbrot.net>
>> +;; URL: https://github.com/syl20bnr/spacemacs
>> +;;
>> +;; This file is not part of GNU Emacs.
>> +;;
>> +;;; License: GPLv3
>> +
>> +(defconst nginx-packages
>> +  '(
>> +    nginx-mode
>> +    ))
>> +
>> +(defun nginx/init-nginx-mode ()
>> +  (use-package nginx-mode))
>
> I had no particular reason for not setting `:defer t` beyond a habit of taking the defaults for everything that I don't have a reason to set to a non-default value.
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/syl20bnr/spacemacs/pull/4963/files#r51932723


-- 
Un saludo,
Alejandro Catalina Feliu

TheBB: Yes, every package should have deferred loading unless explicitly required on startup.

As an aside I'm thinking if this package doesn't work better in the extra-langs layer.

waisbrot:
I've corrected the `:defer` flag. 

Regarding the suggestion to place this in extra-langs: everything in there seems like an actual programming language. Even if it might be painful to do so, you can use MATLAB to write a graphical tic-tac-toe game or to periodically scan your system for files named "*.exe". Nginx config files are much less powerful, I think.

AlejandroCatalina:
I think it might be in both places. The key argument in my case to
decide wether to put it in extra or not would be how widely used is it.

If it is quite useful i'll leave it there, if it is more like an extra
thing, something good to have but not so used... Well, then I think it
could be better off in extra.

This is just my opinion, and please note i do not have a clear one on
this topic.

Nathaniel Waisbrot writes:

> I've corrected the `:defer` flag. 
>
> Regarding the suggestion to place this in extra-langs: everything in there seems like an actual programming language. Even if it might be painful to do so, you can use MATLAB to write a graphical tic-tac-toe game or to periodically scan your system for files named "*.exe". Nginx config files are much less powerful, I think.
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/syl20bnr/spacemacs/pull/4963#issuecomment-180062714


-- 
Un saludo,
Alejandro Catalina Feliu

fbergroth:
I also think this is the correct place for this layer. 

Though, I don't see much benefit to adding single-package layers that adds nothing over just adding the package to `dotspacemacs-additional-packages`. I guess it serves a purpose of saying that this package has been tested in Spacemacs.

@TheBB Currently it seems that extra-langs has become a layer that just accumulates unrelated packages with tiny configurations. I don't think it's very elegant to just dump packages in there, as a user is usually just interested in a few of them.

waisbrot:
Oh, `dotspacemacs-additional-packages` is basically what I was thinking of. I didn't notice it.

Maybe this should just be a README list somewhere of "additional packages that work with spacemacs"? Everything in extra-pangs also seems like it would fall into that category.

syl20bnr:
There is now a value to add layer with single packages, this value is the lazy install feature. Spacemacs can ask the user to install a layer when opening a file based on `auto-mode-alist`.

So I'm for keeping it :-)

syl20bnr:
I finished the support for lazy loading so now there is some value to add simple layers like this because Spacemacs can propose to the user to install the packages needed to handle a filetype.

Thank you :+1: 

nixmaniack:
Lazy loading of layers is cool! Spacemacs should have a blog which introduces these kind of features to wider public.

PR #4958, opened 2016-02-04
[RDY] Add emberjs layer

robbyoconnor:
Ready for merge 

I excluded keybindings for working with Ember views since they are not used anymore.

PR #4964, opened 2016-02-04
Add super-save layer that auto-saves buffers

mmainz:
This layer is inspired by Bozhidar Batsovs
super-save (https://github.com/bbatsov/super-save), but is adapted to
work with Spacemacs. Using the original package proved to be impractical
for integrating with Spacemacs, and since it is not a large package, the
functionality was reimplemented.
When you enable this layer, switching windows or Spacemacs losing focus
causes an automatic save to the current buffer.

I want to say I'm not very experienced in Emacs LISP, so if you see something weird in the code, it's probably because of that. Then again, there are only minor modifications to the original super-save code.

If for some reason you don't like to have this layer in the Spacemacs repository, no hard feelings. I can just put it in its own repository and use it from there. But I thought there might be other people like me that have forgotten how to save manually since using super-save :)

### DIFF
--- a/layers/super-save/README.org
+++ b/layers/super-save/README.org
@@ -1,0 +1,24 @@
+#+TITLE: super-save layer
+#+HTML_HEAD_EXTRA: <link rel="stylesheet" type="text/css" href="../css/readtheorg.css" />
+
+#+CAPTION: logo
+
+* Table of Contents                                        :TOC_4_org:noexport:
+ - [[Description][Description]]
+ - [[Install][Install]]
+ - [[Key bindings][Key bindings]]
+
+* Description
+Inspired by https://github.com/bbatsov/super-save but adapted for Spacemacs.
+Automatically saves your buffers when you select another window or Spacemacs loses focus.
+
+* Install
+To use this contribution add it to your =~/.spacemacs=
+
+#+begin_src emacs-lisp
+  (setq-default dotspacemacs-configuration-layers '(super-save))
+#+end_src
+
+* Key bindings
+
+No key bindings.
--- a/layers/super-save/packages-config.el
+++ b/layers/super-save/packages-config.el
@@ -1,0 +1,15 @@
+;;; packages-config.el --- super-save layer packages file for Spacemacs.
+;;
+;; Copyright (c) 2012-2016 Mario Mainz & Contributors
+;;
+;; Author: Mario Mainz <mainz.mario@googlemail.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+(defconst super-save-packages
+  '(window-numbering))
+
+;;; packages-config.el ends here
--- a/layers/super-save/packages.el
+++ b/layers/super-save/packages.el
@@ -1,0 +1,90 @@
+;;; packages.el --- super-save layer packages file for Spacemacs.
+;;
+;; Copyright (c) 2012-2016 Mario Mainz & Contributors
+;;
+;; Author: Mario Mainz <mainz.mario@googlemail.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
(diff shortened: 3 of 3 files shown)


robbyoconnor:
Define your packages in `packages-config.el`

and initialize them in `packages.el`

I fixed the wizard but @syl20bnr is waiting a bit it seems

robbyoconnor:
See [this example](https://github.com/syl20bnr/spacemacs/pull/4958/files).

mmainz:
Thanks for the hint! I updated the PR. I amended the commit to satisfy the one commit rule.

robbyoconnor:
:+1:

thimios:
:+1:

robbyoconnor:
What is the difference between this and what we currently have?

joehillen:
@robbyoconnor Spacemacs currently doesn't save the file if the buffer loses focus. That would be nice to have. It would save me a lot of mistaken compiles.

robbyoconnor:
Oh nevermind :)

sirech:
I come from prelude, and would love to have this layer in as well. I am not used to saving stuff manually anymore :+1:

joehillen:
@sirech Please :+1: the PR. The maintainers looks that those numbers.

syl20bnr:
@mmainz Thank you for the PR, it is on the top of my queue of PR to merge but first I want to know why the original package does not play well with Spacemacs, what issues did you fix with your own implementation ?

mmainz:
@syl20bnr I honestly forgot what the problem was :)
I'd have to check again. Maybe the original problem I tried to circumvent doesn't exist anymore, since super-save has been updated since I made this PR as well. I could try again to use the package instead once I find some time, but I'm currently busy moving, so it might take a while.

zhengpd:
Any update on this? It would be nice to auto save buffer when losing focus.

sirech:
This PR broke with the update to 0.200.8, as it relies on `window-numbering`, whereas `winum` is used by _spacemacs_ now. I fixed it this way:

```diff
--- a/layers/super-save/packages.el
+++ b/layers/super-save/packages.el
@@ -10,7 +10,7 @@
 ;;; License: GPLv3

 (defconst super-save-packages
-  '(window-numbering))
+  '(winum))

 (defgroup super-save nil
   "Smart-saving of buffers."
Stage this hunk [y,n,q,a,d,/,j,J,g,e,?]? y
@@ -29,17 +29,17 @@
     "evil-window-down"
     "evil-window-left"
     "evil-window-right"
-    "select-window-by-number"
-    "select-window-0"
-    "select-window-1"
-    "select-window-2"
-    "select-window-3"
-    "select-window-4"
-    "select-window-5"
-    "select-window-6"
-    "select-window-7"
-    "select-window-8"
-    "select-window-9")
+    "winum-select-window-by-number"
+    "winum-select-window-0"
+    "winum-select-window-1"
+    "winum-select-window-2"
+    "winum-select-window-3"
+    "winum-select-window-4"
+    "winum-select-window-5"
+    "winum-select-window-6"
+    "winum-select-window-7"
+    "winum-select-window-8"
+    "winum-select-window-9")
   "A list of commands which would trigger `super-save-command'."
   :group 'super-save
   :type '(repeat string))
Stage this hunk [y,n,q,a,d,/,K,j,J,g,e,?]? y
@@ -87,7 +87,7 @@
    (super-save-mode (super-save-initialize))
    (t (super-save-stop))))

-(defun super-save/post-init-window-numbering ()
+(defun super-save/post-init-winum ()
   (super-save-mode +1))

 ;;; packages.el ends here
```

xged:
(defadvice switch-to-buffer (before save-buffer-now activate)
      (when buffer-file-name (save-buffer)))
    (defadvice other-window (before other-window-now activate)
      (when buffer-file-name (save-buffer)))

Above snippet auto-saves on every `spacemacs/evil-mc-paste-after` (`p`). Will it not be the case with this layer?

beta1440:
Why not add `super-save` to the Spacemacs distribution?

Compro-Prasad:
@mmainz Are you still on this?

PR #4966, opened 2016-02-05
[restclient] Added keybinding for copy curl command

d1egoaz:


layers/restclient/README.org
```diff
@@ -22,6 +22,7 @@ file.
 
 | Key Binding | Description                                                 |
 |-------------+-------------------------------------------------------------|
+| ~SPC m c~   | Copy query under the cursor as a curl command               |
```
Andre0991: I have never used this package so I might be wrong, but  I think `y` (which stands for `yank`) instead of `c` would
 fit better Spacemacs' conventions

d1egoaz: Thanks, that makes sense/

StreakyCobra:

I have removed trailing whitespaces in 6055ea9.

PR #4967, opened 2016-02-05
Fix git-gutter linum setup.

person808:


StreakyCobra:

PR #4962, opened 2016-02-04
ivy: Add custom transient state

justbur:
Replace the hydra usually bound to C-o

@Andre0991 I saw your discussion and came up with this alternative.

PR #4750, opened 2016-01-22
ivy/helm: Allow spacemacs-helm with spacemacs-ivy

justbur:
Introduce spacemacs-helm-basic-config variable that prevents helm from
making changes to global configuration elements. This allows
spacemacs-ivy to play with spacemacs-helm.

PR #4969, opened 2016-02-05
Collected ivy PRs

justbur:
I collected my PRs for ivy into one branch, mainly so it's more convenient for people to use (including myself). I can add more commits to this if someone else wants to contribute.

Currently includes:
 - #4962 
 - #4953 
 - #4919 
 - #4750


robbyoconnor:
You're amazing

myrjola:
Thank you @justbur! Much more convenient this way. Moving the commits together made me discover the goodness in #4962.

justbur:
@myrjola I'm not sure I understand [your comment](https://github.com/syl20bnr/spacemacs/pull/4969#commitcomment-15931359). They aren't disabled.

justbur:
@myrjola @syl20bnr I changed the API for these functions to use a hook. The changes are in the most recent commit.

myrjola:
@justbur, very nice! I also tested that toggling holy-mode and hybrid-mode changes the bindings in helm and ivy as one would expect. Can't wait to have these commits merged to develop.

abo-abo:
@justbur buffers generated by `ivy-occur` seem to work only in "insert" mode - nothing works in "normal" mode.

mkomitee:
It looks like the implementation of `spacemacs/ivy-persp-buffer` doesn't actually limit ivy to the buffers in the current perspective. I think `ivy-switch-buffer` maintains its own listing for buffers, so the `with-persp-buffer-list` macro doesn't help.

justbur:
> @justbur buffers generated by ivy-occur seem to work only in "insert" mode - nothing works in "normal" mode.

@abo-abo I'm not sure what you mean by that

justbur:
@mkomitee The helm function doesn't seem to work this way either.

abo-abo:
> @justbur buffers generated by ivy-occur seem to work only in "insert" mode - nothing works in "normal" mode.
> 
> @abo-abo I'm not sure what you mean by that

Key bindings like <kbd>RET</kbd> (`ivy-done`) don't work in `*ivy-occur*...` buffers when in normal mode. I have to press <kbd>i</kbd> to get into the insert mode in order for the bindings to start working.

justbur:
Thanks @abo-abo. I hadn't played with the occur stuff yet. Should be fixed in https://github.com/syl20bnr/spacemacs/commit/80395b9118a30f570d782b9a09185478a950bba5

justbur:
To anyone who cares on this thread, I got wgrep working with the spacemacs search functions. To try it

 1. `SPC /` (or your favorite search func)
 2. search for something
 3. `C-c C-o` for `ivy-occur`
 4. `,w` in normal mode for `wgrep`
 5. edit the results and `,,` to "commit" (`,a` to abort)

robbyoconnor:
> edit the results and ,, to "commit" (,a to abort)

@justbur why not `,c` as is convention?

justbur:
@robbyoconnor fixed, but `,,` is a convention too

robbyoconnor:
I do wonder however if `,c

robbyoconnor:
I do wonder if `,c` and `,k` will clash with magit now?

sooheon:
@justbur wgrep is pretty amazing. Should g/G be swapped in the hydra though? I know going to bottom is more common, hence the desire to put it on an easier key, but it's literally the opposite of vim, so it's confusing. Also, opening the custom transient state with C-o, then pressing keyboard `<escape>` bugs out --- the next keypress still registers to the transient state, for some reason.

syl20bnr:
:+1: 
@justbur Remove the [WIP] prefix when you judge it is ready to me merged.

layers/+completion/spacemacs-helm/config.el
```diff
@@ -0,0 +1,14 @@
...
+;; Author: Sylvain Benner <sylvain.benner@gmail.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+(defvar spacemacs-helm-basic-config nil
+  "If non nil, don't bind keys globally in spacemacs-helm and
+don't enable helm-mode.")
```
syl20bnr: I'm not for this variable, it does not make sense, Spacemacs should always configure Helm whenever it is used by a package. The issue is that with the current layer system we cannot say "configure this package because it is a dependency of another package".

But we can hack around by adding a local package called `basic-helm-config` to `spacemacs` layer. This local package would have no file in local directory.

I guess we will have to do the same for Ivy but we can keep this for another PR, this one being rather big already.

syl20bnr: Actually the local package is not a good solution.

Consider creating a new layer `spacemacs-helm-basic` which is automatically added by `spacemacs-helm`. This way we keep the isolation and the user can choose between a minimal helm config and a the full blown config.

Mmmmmhh I wonder if `spacemacs-ivy` should automatically include `spacemacs-helm-basic` as well. Would make sense to me.

syl20bnr: `spacemacs-helm-basic` can be rename to something better like `spacemacs-helm-base` maybe ? We already have `spacemacs-base` layer.

justbur:
@sooheon thanks, I agree. Those issues should be fixed now. 

@syl20bnr I don't really like the separate layer for key bindings idea, so I came up with an alternative which is more explicit. I added `spacemacs-helm-bind-keys` and `spacemacs-ivy-bind-keys` as layer variables to each layer. They default to `t`. So to configure spacemacs-helm and spacemacs-ivy to work together you would do this in the dotfile

```elisp
(spacemacs-helm :variables spacemacs-helm-bind-keys nil)
spacemacs-ivy
```

and you can also do the reverse. This is a nice convention to me that could be used across other layers that you might want to load but not bind keys for. I'm removing the WIP, because I don't have anything more to do here, unless there are other suggestions of course.

justbur:
I also rebased this

syl20bnr:
Are we sure there is only key bindings there ? What about the other UX components like the presentation ?

Not sure this PR is the right place to work on this but I believe that we should adopt a different strategy for Ivy and Helm:

Since packages can define either Helm of Ivy as dependencies (the two most popular completion framework) I think Spacemacs should provide de facto a base configuration for both of them without forcing their installation (let the dependencies tree do it if required). Then the configuration bit is not about wether the use wants the basic config or not, it is about if they want Helm or Ivy as default completion framework.
This strategy expands to Ido minus the fact that we won't support Ido as default completion framework.

justbur:
> Are we sure there is only key bindings there ? What about the other UX
components like the presentation ?

What are you referring to?

The point of this was to allow people to use helm and ivy. Helm for
somethings and ivy for others. Mostly this was because we didn't have ivy
versions of everything yet, but I think that concern in most taken care of
now.

What if someone wants ivy for everything except helm-ag for example?

abo-abo:
I'm not really experienced with layers, but if I were to do it for my config, I would add 4 things:

```elisp
(eval-after-load 'ivy
  ;; basic Ivy config, even if the user wants Helm bindings
  ;; e.g. set `ivy-height' etc
  ;; If the user e.g. does M-x `counsel-M-x' this should be loaded.
  )

(eval-after-load 'helm
  ;; basic Helm config, even if the user wants Ivy bindings
  ;; e.g. set `helm-candidate-number-limit' etc
  ;; If the user e.g. does M-x `helm-M-x' this should be loaded.
  )
  
(define-layer 'spacemacs-ivy
  ;; all key bindings go here
)

(define-layer 'spacemacs-helm
  ;; all key bindings go here
)
```

The first two statements, which configure Ivy or Helm after they were loaded (possibly due to a package dependency), should be always present. The user can have them both at once, without conflict.

The other two statements, it's up to the user to decide which layer to prefer, but they're almost exclusive. At least, of the two there should be a dominant one - the one that takes up the key bindings if it can. And the other one can be used as backup, in case the dominant one doesn't define a particular key binding.

syl20bnr:
@abo-abo :+1: 

The `eval-after...` bits can be encapsulated within local packages. This was my original answer but I modified it while writing for the base layers instead.

justbur:
In that case, I removed the helm/ivy integration code from this pr so you can do it the way you want to @syl20bnr. 

This means that `spacemacs-ivy` (as of this pr) assumes that either helm or ivy is used but not together.

syl20bnr:
Yes we can do it in another PR :+1:

syl20bnr:
Thank you :purple_heart: :purple_heart: :purple_heart:
I changed the key binding for `ivy-resume` to `SPC r l`  to match the binding for `helm-resume`.
Cherry-picked and rebased into develop branch, you can safely delete your branch.

abo-abo:
Thanks @justbur and @syl20bnr, I'm testing it right now and it looks very good.

Two questions: 

1. I couldn't self-update on the develop branch; I got "void symbol `projectile-mode`". Then I had to delete the whole ELPA just to get things working. Is this a common bug?
   
2. In my opinion, `ace-window` is nearly useless on <kbd>SPC w SPC</kbd>, while commands with less utility have much better bindings, like `other-window` on <kbd>SPC ww</kbd> and `other-frame` on <kbd>SPC wo</kbd>. I don't know if it's common for Spacemacs users to bring up requests for bindings changes, so I do it in a comment that can be easily ignored if warranted :)

robbyoconnor:
To update, you just git pull --rebase

--Rob

abo-abo:
> To update, you just git pull --rebase

That's what I did. Then restarted, and it didn't go well.

nixmaniack:
1. To update develop: you have to do it via git pull and update packages via "Update packages" on Spacemacs home buffer. 

2. I definitely find ace-window much more pleasing than `SPC ww` and `SPC wo`, it should be discussed in an issue.

StreakyCobra:
> I don't know if it's common for Spacemacs users to bring up requests for bindings changes

@abo-abo Yes of course :relaxed: Go ahead, open an issue, or even a PR!

justbur:
fwiw I have ace-window on SPC ww

Andre0991:
@justbur Where can we continue giving suggestions?
Seems that `SPC f L` is bound to `helm-locate` instead of `counsel-locate` (I have both `spacemacs-ivy` and `spacemacs-helm` enabled). Also, `SPC f f` is `helm-find-files`. I'm not sure how `ivy` and `helm` work together currently, which one takes precedence? Is it still OK to use both `spacemacs-ivy` and `spacemacs-helm`?

justbur:
@Andre0991 

> Where can we continue giving suggestions?

I don't have more changes in the works, so you can either open a pr or an issue about them.

> Seems that SPC f L is bound to helm-locate instead of counsel-locate (I have both spacemacs-ivy and spacemacs-helm enabled)

Might be because helm overwrites the binding. See below.

> Is it still OK to use both spacemacs-ivy and spacemacs-helm?

No, I removed my integration code because of the discussion above. Well... yes, but I'm not responsible for what happens :wink:

Andre0991:
@justbur Right. I want to use both so I think I'll just rebind some keys.

robbyoconnor:
Hi,

I'd like to add you to my professional network on LinkedIn.

- Robert

Confirm that you know Robert: https://www.linkedin.com/e/v2?e=0-ini67nz6-tm&t=ssuw&tracking=eml-guest-invite-cta&ek=invite_guest&sharedKey=EvHMphma&invitationId=6130896991145779201








You received an invitation to connect. LinkedIn will use your email address to make suggestions to our members in features like People You May Know. Unsubscribe here: https://www.linkedin.com/e/v2?e=0-ini67nz6-tm&t=lun&midToken=AQFmrFVhGEd36Q&ek=invite_guest&loid=AQHDz14rTef6-gAAAVRVSbnc5rRhu4BMqAMJKHltFe_ISiO4Rw5M2REsdi_Rbs-CplAEPOi8rSua-tF8MJIb_O3wTAOSXSUQr9XbmYjhAyZvsA4bJkE2ZHTz5QUhnem-O1wbP-1Z5EXSPYIMQ8dICQKf8nhNlURx9qhsvPZvSiz7EHx3zTfNRgrwNhRAAsZVFWLJ&eid=0-ini67nz6-tm


This email was sent to reply@reply.github.com.

If you need assistance or have questions, please contact LinkedIn Customer Service: https://www.linkedin.com/e/v2?e=0-ini67nz6-tm&a=customerServiceUrl&ek=invite_guest

&copy; 2016 LinkedIn Corporation, 2029 Stierlin Court, Mountain View CA 94043. LinkedIn and the LinkedIn logo are registered trademarks of LinkedIn.

syl20bnr:
In develop, Ivy users now get a base configuration for helm encapsulated in `with-eval-after-load` (and vice-versa), so everybody should get a consistent experience if a package uses the other completion method. To test it. an Ivy user can `SPC SPC helm-find-files` and they should get a true Spacemacsy helm experience :-)

PR #4975, opened 2016-02-05
Eww gitignore fix

ahyatt:
I realize that eww support in spacemacs is still not done, but at least using eww files should be in .gitignore.  This is that fix (although it's possible this does not include every kind of eww file).


AlejandroCatalina:
This might need a little rebase.

robbyoconnor:
you need to branch from `develop` -- you branched from `master` here

ahyatt:
Although I'm sure there must be a way to fix up this pull request by doing clever git things, I basically have no idea how to do that, so I'll just close it instead and open a new one.

ahyatt:
New request is #4976.

PR #4976, opened 2016-02-05
Add files used by eww to .gitignore.

ahyatt:
I realize that eww support in spacemacs is still not done, but at least using eww files should be in .gitignore. This is that fix (although it's possible this does not include every kind of eww file).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,12 +7,14 @@ eproject.lst
 \#*
 .places
 .cache
+/eww-bookmarks
 eshell/history
 .emacs.desktop
 .emacs.desktop.lock
 eshell/alias
 eshell/lastdir
 /url/cookies
+/url/cache
 my-org/
 org-files/
 semanticdb/
@@ -33,6 +35,7 @@ tmp/
 /history
 .python-environments/
 server/
+/network-security.data
 *~
 *.elc
 *.pyc


robbyoconnor:
There we go :+1:

AlejandroCatalina:
Yup, much better now.

StreakyCobra:
Just to be sure, can some `eww` users check if there is a way to set these locations to be in `.cache/` instead? It is the prefered way to do it in Spacemacs :-) If it is not possible then this is fine.

fandag:
Not sure about `/eww-bookmarks`. The file is created like this in the eww code (and read the same way):

    (defun eww-write-bookmarks ()
      (with-temp-file (expand-file-name "eww-bookmarks" user-emacs-directory)
        (insert ";; Auto-generated file; don't edit\n")
        (pp eww-bookmarks (current-buffer))))


`/url/cache` is part of the `url` configuration, which can be modified by setting `url-configuration-directory`

`/network-security.data` is part of the nsm configuration in Emacs25, which can be modified by setting `nsm-settings-file`

robbyoconnor:
yeah this whole PR can be made moot if we fix the layer :)

StreakyCobra:
Thanks @fandag . So I think `eww-bookmarks` can be added to `.gitignore` but I would prefer configuring the two others to use the `.cache` folder instead, if it makes sense.

syl20bnr:
url cache directory is already set to `.cache`, I guess it was an old cache directory in your config.
I let the two others on gitignore for now since I don't have Emacs 25 right now.

Thank you :+1: 

PR #4978, opened 2016-02-06
[command-log] fix readme css path

d1egoaz:
Current command-log layer isn't showing custom css.
LHS (without) - RHS (with css)
image


StreakyCobra:

PR #4981, opened 2016-02-06
hide export tags and TOC annotation in README.org files

JAremko:
Reduce visual noise in Specemacs'es docs.

Affects only` README.org` files  opened  with `org` and only in the `~/.emacs.d/layers/` folder or it's sub folders.

### DIFF
--- a/layers/org/packages.el
+++ b/layers/org/packages.el
@@ -290,6 +290,21 @@ Will work on both org-mode and any mode that accepts plain html."
       (spacemacs/set-leader-keys
         "Cc" 'org-capture)
 
+      (unless (version< emacs-version "24.4")
+
+        (defun spacemacs/org-clearer-view ()
+          "Hide export tags and annotations when viewing documentation with `SPC` `h` `SPC`"
+          (interactive)
+           (when (string-match (concat (expand-file-name "~") "/.emacs.d/layers/.*README\.org")  buffer-file-name)
+             (progn
+               (push '("#+TITLE: " . ?⁣) prettify-symbols-alist)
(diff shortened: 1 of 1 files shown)


layers/org/packages.el
```diff
@@ -290,6 +290,21 @@ Will work on both org-mode and any mode that accepts plain html."
       (spacemacs/set-leader-keys
         "Cc" 'org-capture)
 
+      (unless (version< emacs-version "24.4")
+
+        (defun spacemacs/org-clearer-view ()
```
JAremko: `clear-view` was taken :smile:

JAremko:
I propose tag `dirty-hacks` :joy:

JAremko:
First I was thinking that it should be optional, but then I realized how awesome it is and decided that it should be mandatory.

JAremko:
My best work so far @StreakyCobra :sheep:  :it:

JAremko:
Also it encourages those plebs on old Emacses to upgrade thus reducing maintenance cost. 
It's like awesome with extra layer of the same stuff.

layers/org/packages.el
syl20bnr: Why not `spacemacs/org-hide-export-tags-and-annotations` ? :-)

JAremko: @syl20bnr Because people will think I'm being serious. Am I? I don't even know :confused: 
I should reduce the amount of :hankey:  in it before giving diz `smelly-foo` any proper name

JAremko:
Closed in favor #4989

PR #4984, opened 2016-02-06
[WIP] centered-cursor-mode: override H M L zt zz zb

e-matteson:
Previously, when centered-cursor-mode was enabled, the evil bindings
zt/zz/zb/M had no visible effect, and H/L scrolled the window up/down.
This commit adds new functions that behave more intuitively in
centered-cursor-mode, and overrides the evil normal state bindings.
H/M/L change the fixed vertical position to the top/center/bottom by
moving the cursor to the line at that new position without scrolling the
window. zt/zz/zb change the fixed vertical position to the
top/center/bottom by scrolling the window without moving the cursor.

This all works for me, but I'm not completely confident it respects all relevant conventions and overrides evil bindings in the proper way. Suggestions are welcomed.

### DIFF
--- a/layers/+spacemacs/spacemacs-ui/local/centered-cursor/centered-cursor-mode.el
+++ b/layers/+spacemacs/spacemacs-ui/local/centered-cursor/centered-cursor-mode.el
@@ -94,7 +94,7 @@
 ;; This file is *NOT* part of GNU Emacs.
 
 ;;; Code:
- 
+
 (defgroup centered-cursor nil
   "Makes the cursor stay vertically in a defined position (usually centered).
 Instead the cursor the text moves around the cursor."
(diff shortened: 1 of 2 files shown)


e-matteson:
I noticed a possible downside of overriding the evil normal state map with the centered-cursor-mode map. centered-cursor-mode binds C-v to ccm-scroll-up. After my commit, this overrides evil-visual-block, which is probably more useful to evil users.

Possible solutions include:
1. Don't add any new bindings to the centered-cursor-mode map, just add the new functions and let users bind them to H/M/L etc. in .spacemacs if they want.
2. Remove the C-v binding from centered-cursor-mode.
3. Somehow make the C-v binding apply only to holy-mode?

Thoughts?

e-matteson:
Closing this, because it's incomplete, I'm not sure how best to improve it, I'm not using centered-cursor-mode anymore, and no one else seems interested in the feature.

PR #4985, opened 2016-02-07
paste-ts: Change key bindings to C-j/C-k

justbur:
J was annoying for those who wanted to use join lines right after a
paste.


StreakyCobra:

Issue #4980, opened 2016-02-06
New layer format


darkfeline:
I'd like clarification on the new layer format.  Relevant links:

Layers documentation: https://github.com/syl20bnr/spacemacs/blob/develop/doc/LAYERS.org
Relevant commit: https://github.com/syl20bnr/spacemacs/commit/09af60594c315b956b284bd58e5f2af1beddd011

## Old format

- `packages.el`: Set layer packages and configure packages (`init-<package>`).
- `funcs.el`: Function definitions, including named functions that are used in `init-<package>`.
- `keybindings.el`: Any keybindings that don't go into `init-<package>` (not package specific).
- `config.el`: All remaining configuration that don't go into `init-<package>` (not package specific).

## New format

- `packages.el`: Set layer packages.
- `packages-config.el`:  Configure packages (`init-<package>`).
- `packages-funcs.el`:  Function definitions, only named functions that are used in `init-<package>` (package related).
- `funcs.el`: Remaining function definitions (not package specific).
- `keybindings.el`: Any keybindings that don't go into `init-<package>` (not package specific).
- `config.el`: All remaining configuration that don't go into `init-<package>` (not package specific).

## Questions

1. Is my understanding correct?
2. Migrating existing layers involves: moving init functions `packages.el` -> `packages-config.el`, moving some definitions `funcs.el` -> `packages-funcs.el`?
3. Is having a separate `keybindings.el` necessary?  I believe these will be things like calls to `spacemacs/set-leader-keys`, `evil-define-key`, etc.  Can these not simply go into `config.el`?
4. Regarding where to put functions, say a function `my-layer-foo-function` gets used in `init-foo` as `(setq foo-function 'my-layer-foo-function)`, then it goes into `packages-funcs.el`.  An interactive command `my-layer-frob-paragraph` goes into `funcs.el`.  Does this sound right?
5. Is the new layer format "final"?  Should existing layers be migrated?
6. What's the motivation behind the change?

StreakyCobra:
@syl20bnr is the best person to answer all this :-) But I can give some indications about what I know:

> What's the motivation behind the change?

It has to do with [lazy layers installation](https://github.com/syl20bnr/spacemacs/blob/f94263919b53faf0bd83910100185a29179e3281/core/templates/.spacemacs.template#L14-L16). It *seems* it could also help solving #3430.

> Should existing layers be migrated?

[...]

PR #4986, opened 2016-02-07
Fix layer anatomy documentation

darkfeline:
Fix documentation per https://github.com/syl20bnr/spacemacs/issues/4980


robbyoconnor:
Looks good to me :+1:

syl20bnr:
This is not right, see my answer in #4980 ;-)

darkfeline:
Closing then, I'll point out that the plantuml layer needs fixing.

robbyoconnor:
Why does it need fixing?

PR #4988, opened 2016-02-07
scope the aliases under eshell so they are not defined globally

bling:


layers/shell/packages.el
```diff
@@ -140,8 +140,8 @@ is achieved by adding the relevant text properties."
       (require 'esh-opt)
 
       ;; quick commands
-      (defalias 'e 'find-file-other-window)
-      (defalias 'd 'dired)
+      (defalias 'eshell/e 'find-file-other-window)
+      (defalias 'eshell/d 'dired)
```
syl20bnr: Well, this alias is now longer to type than `dired`:-)

syl20bnr:
I did not realise that the alias were globals which sucks.
On the other hand if we add `eshell/

bling:
not sure what you mean?  even though the function name is `eshell/d` you can still type just `d` in eshell.

TheBB:

@syl20bnr eshell looks up aliases under the `eshell/` prefix, so this is harmless.

PR #4993, opened 2016-02-07
layouts: add funct to switch between open projects

AlejandroCatalina:
This will allow a user to switch between already open projects' layouts more easily than looking for the desired project in `SPC p l`. I find this really helpful since I'm working on several projects at a time and it's a pretty common operation for me.

It still has no keybinding because I didn't know where to put, something like `SPC p L` might me nice. When a keybinding is chosen i'll add the proper documentation. It's also missing a helm function that i'll add as soon as i can.

Hope it helps! Any suggestions and corrections welcome.

### DIFF
--- a/layers/+window-management/spacemacs-layouts/funcs.el
+++ b/layers/+window-management/spacemacs-layouts/funcs.el
@@ -96,8 +96,15 @@ perspectives does."
                            (helm-marked-candidates))))))))
 
 ;; Helm Projectile related functions ---------------------------------------
+(defun spacemacs//layout-search-project (project)
+  "Look for a project perspective among current layouts."
+  (let ((persps (persp-names-current-frame-fast-ordered)))
+    (do ((idx 0 (1+ idx))
+         (curr (car persps) (nth idx persps)))
+        ((equal curr project) idx))))
 
 (defun spacemacs/helm-persp-switch-project (arg)
+  "Switch to a project perspective using Helm."
   (interactive "P")
   (helm
    :sources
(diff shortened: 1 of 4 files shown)


robbyoconnor:
:purple_heart: :yellow_heart: :heart: :blue_heart: :green_heart:

AlejandroCatalina:
This seems ready to me, aside keybinding discussion and probably some idiomatic expression or so.

StreakyCobra:
You can probably remove the WIP then :-)

AlejandroCatalina:
Right! Thanks

layers/+window-management/spacemacs-layouts/funcs.el
```diff
@@ -117,6 +122,25 @@ perspectives does."
                       (projectile-switch-project-by-name project)))))))
    :buffer "*Helm Projectile Layouts*"))
```
jmiven: You've forgotten to `(push project layouts-projectile)` in `spacemacs/helm-persp-switch-project`. :wink:

layers/+window-management/spacemacs-layouts/funcs.el
```diff
@@ -117,6 +122,25 @@ perspectives does."
...
 
+(defun spacemacs/helm-persp-switch-open-project (arg)
+  (interactive "P")
+  (helm
+   :sources
+   (helm-build-in-buffer-source "*Helm Switch Project Layout*"
+     :data (lambda ()
+             (if (projectile-project-p)
+                 layouts-projectile
+               projectile-known-projects))
```
jmiven: I'm not very familiar with helm, but shouldn't the list of known projects be available even if some projects are already opened?

jmiven: Also, switching to a known project with this function makes Emacs freeze. I don't know yet how to debug that. :confused:

jmiven:
I like this feature! :+1:

layers/+window-management/spacemacs-layouts/funcs.el
AlejandroCatalina: Thanks! I don't use helm so I had it pending to test.

layers/+window-management/spacemacs-layouts/funcs.el
AlejandroCatalina: I'll work on the freezing thing. 

I didn't add the known project as this is only to switch between already open ones. To see the known projects you have `SPC p l`.

AlejandroCatalina:
Actually fixing the `(push project layouts-projectile)` fixed the freezing thing, that was caused trying to access to the -1 persp.

I also decided that the helm function will only show already open projects. If you guys prefer showing instead all the known projects when there is not open ones i'll add it, but i thought it is more consistent this way , since the purpose of the function is to switch between open projects.

jmiven:
> I also decided that the helm function will only show already open projects. If you guys prefer showing instead all the known projects when there is not open ones i'll add it, but i thought it is more consistent this way , since the purpose of the function is to switch between open projects.

What I had in mind was a behavior à la helm-mini which lists the opened buffers, but also the recent files.

But anyway you're right. Thanks, I already adopted the feature. :smile:

Andre0991:
>What I had in mind was a behavior à la helm-mini which lists the opened buffers, but also the recent files.
But anyway you're right. Thanks, I already adopted the feature. :smile: 

Good idea. `SPC l l` could use multiple sources to distinguish projects layouts and other layouts.

layers/+window-management/spacemacs-layouts/funcs.el
```diff
@@ -113,10 +118,28 @@ perspectives does."
                 (lambda (project)
                   (let ((persp-reset-windows-on-nil-window-conf t))
                     (persp-switch project)
+                    (push project layouts-projectile)
                     (let ((projectile-completion-system 'helm))
                       (projectile-switch-project-by-name project)))))))
    :buffer "*Helm Projectile Layouts*"))
 
+(defun spacemacs/helm-persp-switch-open-project (arg)
```
syl20bnr: Is the `arg` argument needed?

syl20bnr: Can you document the function ?

layers/+window-management/spacemacs-layouts/funcs.el
```diff
@@ -127,11 +150,20 @@ perspectives does."
...
                       (let ((persp-reset-windows-on-nil-window-conf t))
                         (persp-switch project)
+                        (push project layouts-projectile)
                         (let ((projectile-completion-system 'ivy))
-                          (projectile-switch-project-by-name project))))
-            )
+                          (projectile-switch-project-by-name project))))))
 
-  )
+(defun spacemacs/ivy-persp-switch-open-projects (arg)
```
syl20bnr: Same comments, is `arg` needed ?
Can you document the function ?

AlejandroCatalina: The arg is needed as long as we receive an interactive argument (prefix argument), so i left it there. I'm documenting the functions.

AlejandroCatalina:
Rebased and updated @syl20bnr

nixmaniack:
@AlejandroCatalina This looks like good addition. Any reason to close this PR?

AlejandroCatalina:
This looks done to me!

PR #4998, opened 2016-02-08
Fix react magic mode

EMayej:
Fix regex syntax, the original one doesn't work.


TheBB:

PR #5001, opened 2016-02-08
helm should always be fuzzy by default

bling:


bmag:
1. What's wrong with the current setting?

bling:
nothing wrong -- it's just an incomplete list.  there's a couple sources that don't have fuzzy matching turned on.  this patch ensures that any source, old and new, will have fuzzy turned on.

syl20bnr:
Thank you, nice trick :+1: 
Let's see how it goes.

bmag:
Apparently, this change causes #9631. @bling do you think you can investigate and fix it? seems that `helm--fuzzy-regexp-cache` can be empty during `SPC f f` when following the repro steps in the linked issue, causing an error in `helm-fuzzy-match`.

PR #5003, opened 2016-02-08
include plantuml as a babel language if jar is defined

bling:


robbyoconnor:
Okay -- as much as I support this -- this Why just plantuml? We should add more than just plantuml..

StreakyCobra:
@robbyoconnor Because we have to start somewhere I suppose :-)

robbyoconnor:
Let's expend this -- I mean I'm not saying this PR should be rejected -- I have this in my `dotspacemacs/user-config`

layers/org/packages.el
```diff
@@ -110,6 +110,12 @@
             ;; `helm-org-headings-max-depth'.
             org-imenu-depth 8)
 
+      (with-eval-after-load 'org
+        (when (boundp 'org-plantuml-jar-path)
+          (org-babel-do-load-languages
+           'org-babel-load-languages
+           '((plantuml . t)))))
+
```
syl20bnr: Org layer should not know about plantuml. Do it the other way around:
add `org` to the `plantuml` layer package list and move this configuration to `plantuml/post-init-org` function. Since the configuration lies in the `plantuml` layer you don't need to check for `org-plantum-jar-path`, and `plantuml/post-init-org` is executed only if the `org` layer is used.
You can either keep the `with-eval-after-load` or use a `use-package` hook with `spacemacs|use-package-add-hook` to inject your configuration in the `:config` block for instance.

layers/+lang/plantuml/packages-config.el
```diff
@@ -14,6 +14,6 @@
 ;;; Code:
 
 (defconst plantuml-packages
-  '(puml-mode))
+  '(org puml-mode))
```
robbyoconnor: I think each package should be on its own line

layers/+lang/plantuml/packages.el
```diff
@@ -21,4 +21,13 @@
...
+
+(defun plantuml/post-init-org ()
+  (spacemacs|use-package-add-hook org
+    :post-config
+    (when (boundp 'org-plantuml-jar-path)
+      (org-babel-do-load-languages
+       'org-babel-load-languages
+       '((plantuml . t))))))
+
 ;;; packages.el ends here
```
robbyoconnor: :+1:

robbyoconnor:
Please also update the documentation @bling

layers/+lang/plantuml/README.org
```diff
@@ -31,7 +32,6 @@ robbyoconnor->theOtherPerson : And they thinks it's funny? Yup, they definitely
 [[file:img/dia.png]]
 
 
-
```
robbyoconnor: What is this?

StreakyCobra: @robbyoconnor Probably an extra line removed, I don't see the problem there.

robbyoconnor: It's not hurting anyone -- just curious

syl20bnr:
Thank you :+1: 

PR #5012, opened 2016-02-09
changed binding `SPC j =` to `SPC x i` for indenting buffer

adrianstaniec:
imho it belongs better together with align-\* commands 
rather than with jumps and joins
`x i` for teXt Indent seems also an easier mnemonic
at least for the 'holy' guys

### DIFF
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -2200,6 +2200,7 @@ Text related commands (start with ~x~):
     | ~SPC x g l~ | set languages used by translate commands                      |
     | ~SPC x g t~ | translate current word using Google Translate                 |
     | ~SPC x g T~ | reverse source and target languages                           |
+    | ~SPC x i~   | corrent indentation of region or buffer                       |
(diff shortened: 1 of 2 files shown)


StreakyCobra:
Same as for the other PR, need documentation :wink:

adrsta:
now i'm thinking about maybe `SPC b i` or `SPC b I`
which would fit to buffer functions like `SPC b Y` and `SPC b P`
but less to the region side of things

syl20bnr:
Should be `SPC b =`, `=` is used to format in Vim.

Compro-Prasad:
@adrianstaniec, It is better to assign to `SPC b =` as suggested by Sylvain. It would by default indent whole buffer and would work on regions too which is a plus point. A newcomer will definitely look into `SPC b` for new things and having indentation functionality there is a plus point.

robbyoconnor:
This needs to be rebased.

adrianstaniec:
Sorry, I've lost the interest in spacemacs already.

Issue #3920, opened 2015-11-23
Environment variable warnings


zhexuany:
I just updated my spacmeacs in develop branch. 
And emacs gave me a warning as follow:
Warning (emacs): You appear to be setting environment variables ("PATH") in your .bashrc or .zshrc: those files are only read by interactive shells, so you should instead set environment variables in startup files like .bash_profile or .zshenv.  Refer to your shell's man page for more info.  In future, exec-path-from-shell will not read variables set in the wrong files.

I have no idea what is going on about this. :(

ntcong:
Are you using OSX?

In OSX, program launch from Application/Dock/Spotlight doesn't inherit the interactive shell environment (.bashrc or .zshrc). So spacemacs have exec-path-from-shell in order to get PATH and GOPATH (not sure about other uses) from your shell rc.

exec-path-from-shell recently has a commit warning about the PATH setting in wrong file (it should be in bash_profile instead of bashrc)

idoo:
@ntcong do you have any ideas, how to fix it? move i moved PATH rows from .zshrc to .zshenv, but still receiving message. ( I'm oh my zsh user)

magthe:
I get this on Linux, but for `MANPATH`.

My shell of choice is *zsh*, and I'm using *antigen* and (parts of) OMZ.  (I saw OMZ mentioned in https://github.com/purcell/exec-path-from-shell/issues/34)

Anyway, `MANPATH` is actually unset in my setup (both *zsh* and *bash*), so I'm rather confused at this warning!

ntcong:
there's a running issue https://github.com/purcell/exec-path-from-shell/issues/34 . The commit is recent, so maybe there some edge cases.

nixmaniack:
There might be cases where some other utilities in your startup rc files are setting up environment variables. 
e.g in my case, after removing PATH, MANPATH etc, I was still getting this error. After analysing my rc file, I found `pyenv`, `rbenv` which modify and set PATH hence moving them to `.zshenv` in my case fixed this warning. 
So watch out for such tools.

zhexuany:
After I delete all PATH variable, then the warning vanishes.  I am currently using OSX and ZSH.

magthe:
How did you find out what caused it?

In my case it's `MANPATH`, and it's unset in *zsh*:

```
$ print $MANPATH

```

It's also unset in *emacs*, `getenv` tells me so.

[...]

PR #5024, opened 2016-02-11
Disable exec path from shell under X

nicot:
I'm running into a bug where my path is set incorrectly in Emacs. Specifically, it is not picking up exports I make in my ~/.zshenv file. I'm not sure if this is the correct solution, but this is the best way for me to get Emacs to properly get my environment variables. I think I am setting the env correctly, as it is in my ~/.zshenv and works with this change.

I would disable exec-path-from-shell, but the Go and rust layers need it.

Is anybody more familiar with the Emacs environment able to confirm that this will not break user's setups?

### DIFF
--- a/layers/+distribution/spacemacs-base/packages.el
+++ b/layers/+distribution/spacemacs-base/packages.el
@@ -466,7 +466,7 @@ Example: (evil-map visual \"<\" \"<gv\")"
 
 (defun spacemacs-base/init-exec-path-from-shell ()
   (use-package exec-path-from-shell
-    :init (when (memq window-system '(mac ns x))
+    :init (when (memq window-system '(mac ns))
             (exec-path-from-shell-initialize))))
 
 (defun spacemacs-base/init-fill-column-indicator ()


StreakyCobra:
I don't think it's a better default to disable it, there will be people who will ask to add it back. According to the package author, it's a package that is [also required form some X11 setups](https://github.com/syl20bnr/spacemacs/issues/3920#issuecomment-159706493). And AFAIK it has not caused more problems than it has solved in the past.

It's also more reliable to ask users to add it to the `dotspacemacs-excluded-packages` list if it causes problems than asking people to configure it in their `.spacemacs`.

nicot:
The problem I am having is not the warning (I don't even get a warning.) My problem is that the plugin changes my path to an incorrect one. Is there a way to disable the plugin initialization while still letting it function in the Go and Rust mode?

StreakyCobra:
I understood that your problem is not the warning, I was just pointing out that disabling it for X11 is not a good general solution.

If you don't want your path to be changed, you can probably change [`exec-path-from-shell-variables`](https://github.com/purcell/exec-path-from-shell/blob/c2ca275d3243e8253513ced73e3ac21dc352e303/exec-path-from-shell.el#L79-L83) in `dotspacemacs/user-init`:

```elisp
(setq-default exec-path-from-shell-variables '("GO…" "RUST…"))
```

robbyoconnor:
This is also against `master` please close this and do it against `develop` as the [contribution guidelines](https://github.com/syl20bnr/spacemacs/blob/develop/CONTRIBUTING.org) state. Ignorance isn't an excuse when it's linked quite prominently when you open a PR or issue.

nicot:
@robbyoconnor Thanks Robby. The point of this PR was to offer a suggestion to people whose setups were broken by this, and a search for a better solution.

@StreakyCobra Thanks for the information! That worked perfectly. Should I add this to documentation somewhere, or am I the only one running into this problem?

robbyoconnor:
@nicot Could you put this PR out of its misery please?

StreakyCobra:
@nicot I propose you to create a PR documenting this on the `doc/FAQ.org` file in the `How do I` section, and to close this one.

It would be nice to read the contributing guidelines pointed by @robbyoconnor before creating the new one :-)

syl20bnr:
I agree with @StreakyCobra, it is better to keep it and add a FAQ entry.

Issue #3171, opened 2015-09-28
Enabling structured haskell mode breaks vim indentation


tko:
With plain haskell layer hitting <kbd>= =</kbd> in normal mode indents as one would expect. With `(haskell :variables haskell-enable-shm-support t)` it just produces an error instead:

> haskell-mode-suggest-indent-choice: You tried to do an indentation command, but an indentation mode has not been enabled yet.
>
> Run M-x describe-variable haskell-mode-hook for a list of such modes.

and the description doesn't have anything to say about indentation modes:

> haskell-mode-hook is a variable defined in `haskell-mode.el'.
> Its value is (spacemacs/init-haskell-mode structured-haskell-mode ghc-init flycheck-mode spacemacs//init-company-haskell-mode company-mode)
>
>
>    This variable may be risky if used as a file-local variable.
>
> Documentation:
> Hook run after entering Haskell mode.
> No problems result if this variable is not bound.
> `add-hook' automatically binds it.  (This is true for all hook variables.)

tko:
Related: https://github.com/chrisdone/structured-haskell-mode/issues/81

d12frosted:
Since officially Chris Done marked `shm` and `evil` as mode-conflict and wont-fix, I think we should close this issue. But I would like to leave #1048 open, because I believe that we could add some workaround for evil users. Gonna think about it in the near future. 

@TheBB what do you think?

TheBB:
Yeah, I'm liable to agree. Let's do SHM stuff in #1048.

syl20bnr:
We finally removed it from haskell layer in develop branch.
It may come back in a better shape at some point :-)

syl20bnr:
oops already closed :-D

PR #2617, opened 2015-08-11
update shm key bindings

d12frosted:
By default `structured-haskell-mode` binds `C-j` to `shm/newline-indent-proxy`. And this is very important functionality of `structured-haskell-mode` as it inserts code that you would insert manually and also allows better indentation insertion in certain contexts. So I think it shouldn't be thrown away when `haskell-enable-shm-support` is `t`.

Also, when `C-j` is set to `nil`, `haskell-mode` argues that indentation isn't configured. So it doesn't looks nice.

The only thing I am not sure about is how `C-j` should behave in `normal` state. I don't use it all, so not sure if it's a good practice for `C-j` to insert code in certain situations, but anyway, this is how Chris Done intended it to behave, so I removed this rebind for `normal` state as well. 

For normal state there is also a mess for `RET` key. And by default, when `electric-mode` is on, the default bind from `structured-haskell-mode` will do `shm/newline-indent`. So I've removed them as well.

And about `C-k`. The idea of `structured-haskell-mode` is that you operate on context rather than on line. So I don't understand the point of making `C-k` to remove line.

What do you think? :smile:

Issue #1048, opened 2015-04-03
Rethink structured-haskell-mode keybindings


EDmitry:
I have been trying to wrap my head around structured haskell mode.  By default `)` and `(` are mapped to movements. The problem is that `o` and `O` functionality is missing because you have to choose the right block before you can create a newline otherwise the indentation will be incorrect. I have found that binding ")" in normal mode to switch to insert mode and do `shm/close-paren` improved the workflow a bit. I have tried doing that without switching to insert mode but then the cursor is at the wrong position in the end of line (common evil-mode issue).

What's interesting though is that basically structured-haskell-mode is trying to emulate vim text objects in some way. It already has some sort of selection which is different from evil's selection. I wonder if there is a better way to integrate it with evil-mode.

EDmitry:
Also, maybe we could use some ideas from evil-lisp-state. It seems like SHM is a similar concept but has less actions.

gitfoxi:
I agree that SHM is not doing much in spacemacs. I wonder where would be a good place to work on it?

* Spacemacs
* evil-lisp-mode
* A new evil-haskell-mode

I'd love to know how SHM author @chrisdone thinks it should work.

jb55:
A new `evil-haskell-mode` would be great :+1:

d12frosted:
> I agree that SHM is not doing much in spacemacs.

I am using Spacemacs with emacs editing style and for me `shm` is priceless. But having `evil-haskell-mode` will not hurt me - I will still use `shm` :smile:

syl20bnr:
We finally removed it from haskell layer in develop branch.
It may come back in a better shape at some point :-)

PR #5026, opened 2016-02-11
shm: undo awkward normal-map rebindings

jb55:
I was trying to use structured-haskell-mode in Spacemacs but I found it awkward:

  * The custom shm key bindings remap some important vim key bindings
  * It unmaps `<C-j>` and maps it to normal mode enter, which means you can't use it for `shm/newline-indent`ing lists and strings. You would have to `fdl<enter>` for each list item
  * They are somewhat opinionated and don't match any spacemacs conventions.

We should look into making something similar to evil-lisp-state, which would be really cool.

cc @bjarkevad @d12frosted

### DIFF
--- a/layers/+lang/haskell/packages.el
+++ b/layers/+lang/haskell/packages.el
@@ -282,18 +282,6 @@
         (define-key shm-map (kbd "C-c S") 'shm/case-split)
         (define-key shm-map (kbd "C-c C-s") 'shm/do-case-split))
 
-      (evil-define-key 'normal shm-map
-        (kbd "RET") nil
-        (kbd "C-k") nil
-        (kbd "C-j") nil
-        (kbd "D") 'shm/kill-line
-        (kbd "R") 'shm/raise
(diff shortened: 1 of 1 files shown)


d12frosted:
I'll leave some references that might be interesting: https://github.com/syl20bnr/spacemacs/issues/3171, https://github.com/syl20bnr/spacemacs/issues/1048 and https://github.com/syl20bnr/spacemacs/pull/2617

Honestly, I think that shm should be removed from haskell layer since it doesn't really play well with non-emacs editing style. For those who use emacs editing style and want to use shm - it's easy to add it - and there is nothing spacemacs specific. 

But it would be nice to have someone who uses both evil and shm to fix most of issues with shm in spacemacs :)

jb55:
I agree. A half baked evil shm mode isn't really useful, and could probably be safely removed until someone makes a solid evil state for it.

syl20bnr:
I vote for removing it.

jb55:
Closing in favor of #5087

PR #5028, opened 2016-02-11
html: Don't enable web-mode for jsx files

lunaryorn:
- Conceptually it just doesn't feel right: JSX is no HTML, this would be the wrong layer.
- It seems redundant as well: There's a `react` layer that provides dedicated JSX support.
- Finally, there are other modes for JSX (`js2-jsx-mode` for instance) that user might want to use,
  without affecting Web Mode for template files.


robbyoconnor:
:+1:

syl20bnr:
Nice catch :+1: 

Thank you !

Issue #5029, opened 2016-02-11
ranger-search-files does not work


psamim:
#### Description
In default ranger `f` is bound to `ranger-search-files` while in ranger layer in Spacemacs this key is shadowed by the default evil-mode behaviour of `f` key. I suggest using different key binding for `ranger-search-files`.

#### Reproduction guide
- Start Emacs
- Be sure to use `vim` editing mode
- Be sure ranger layer is loaded
- Open ranger with `SPC a r`
- Press `f`

*Observed behaviour:*
I observe Emacs is waiting for the next letter, because `f<X>` puts the cursor on the next occurrence of the letter <X>.

*Expected behaviour:*
I expect it brings `ranger-search-files` as it is the default ranger behaviour.

#### System Info
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.9
- Spacemacs branch: nil (rev. 0562f05)
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
((auto-completion :variables auto-completion-enable-help-tooltip t)
 better-defaults emacs-lisp javascript clojure haskell python lua c-c++ html shell-scripts ess sql go scala java react git markdown org restclient selectric erc dockerfile latex
 (evil-snipe :variables evil-snipe-enable-alternate-f-and-t-behaviors t)
 evil-commentary spell-checking syntax-checking version-control ranger semicolon org-latex-classes org-psamim psamim-defaults)
```

StreakyCobra:
It works as expected here. Can you update your issue to follow [this template](https://github.com/syl20bnr/spacemacs/wiki/Debugging#how-to-make-a-great-bug-report)?

psamim:
@StreakyCobra Sure, I updated the issue.

StreakyCobra:
Can you try to remove `evil-snipe` from you layer list and try again?

psamim:
I did and now it is working as expected. Should we add a hook to disable evil-snipe in ranger?

StreakyCobra:
Yes probably. @ralesi any opinion on this?

ralesi:
@streakycobra That can work, I will investigate how evil-snipe hooks into evil-mode bindings.

d12frosted:
Fixed with released of Spacemacs v0.200. Let us know if you still have any problems with this issue.

PR #5030, opened 2016-02-11
Add key binding for ranger-search-files

psamim:
The default binding `f` is shadowed by `evil-mode`.
Regarding https://github.com/syl20bnr/spacemacs/issues/5029


StreakyCobra:
(see my comment in #5029)

TheBB:
I'm closing this since we went with #5196. Thanks though!

PR #5032, opened 2016-02-11
company: Fix :init and use default company-frontends

fbergroth:
The default value of `company-frontends` includes the following frontends:

- `company-pseudo-tooltip-unless-just-one-frontend`
  2016-02-11-112558_219x61_scrot
  Display the "popup" with candidates if there are more than 1 candidate.

- `company-preview-if-just-one-frontend`
  2016-02-11-113101_185x23_scrot
  If there is only one candidate, display it inline ("preview").

- `company-echo-metadata-frontend`
  rsz_screenshot_2016-02-11_11-28-28
  Show metadata in echoarea.

The previous setting only enabled the "popup" frontend. I'm not sure why it was set, maybe because `company-echo-metadata-frontend` interferes a bit with the `company-quickhelp-frontend`. I'd argue that the default is better, and we can disable the echo frontend if quickhelp is enabled.

This also fixes the messed up `:init` form.

### DIFF
--- a/layers/auto-completion/packages.el
+++ b/layers/auto-completion/packages.el
@@ -76,21 +76,21 @@
             company-minimum-prefix-length 2
             company-require-match nil
             company-dabbrev-ignore-case nil
-            company-dabbrev-downcase nil
-            company-frontends '(company-pseudo-tooltip-frontend)))
-            (defvar-local company-fci-mode-on-p nil)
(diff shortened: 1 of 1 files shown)


robbyoconnor:
:+1:

syl20bnr:
Excellent :purple_heart: 

Thank you :+1: 

PR #5038, opened 2016-02-11
Remove ruby-end-mode and use smartparens instead

mmainz:
ruby-end-mode does not always behave correctly. An example for that is
if statements, which don't have an end statement if they are written on
one line. Also, it feels dirty to use a Ruby package to implement Elixir
syntax behaviour.
Smartparens has functionality to define pairs for certain major modes.
This adds smartparens pairs for the "do" keyword and the arrow (->)
operator, so that the "end" keyword is automatically inserted and
properly indented on enter.

This also seems to be the recommended way to get this kind of behaviour.
See here: https://github.com/tonini/alchemist.el/issues/97

### DIFF
--- a/layers/+lang/elixir/packages-config.el
+++ b/layers/+lang/elixir/packages-config.el
@@ -102,6 +102,25 @@
   (spacemacs|add-company-hook elixir-mode)
   (spacemacs|add-company-hook alchemist-iex-mode))
 
+(defun elixir-do-end-close-action (id action context)
+  (when (eq action 'insert)
+    (newline-and-indent)
+    (forward-line -1)
+    (indent-according-to-mode)))
+
+(defun elixir/post-init-smartparens ()
+  (sp-with-modes '(elixir-mode)
+    (sp-local-pair "->" "end"
+                   :when '(("RET"))
+                   :post-handlers '(:add elixir-do-end-close-action)
+                   :actions '(insert)))
+
+  (sp-with-modes '(elixir-mode)
+    (sp-local-pair "do" "end"
+                   :when '(("SPC" "RET"))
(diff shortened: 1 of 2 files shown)


robbyoconnor:
Will this still automatically insert the end for if ?

robbyoconnor:
Ah this is just elixir nevermind.

mmainz:
Yes, this is just Elixir. And I think the only case where you have to type `end` without first having to type `do` is `fn -> end`, which is the second rule introduced in my changes :)

mmainz:
Also, the way it is implemented now, typing an if statement results in code like below:
```
if condition do
|
end

end
```
Because `ruby-end-mode` inserts an `end` on `if` and another one on `do`, because that is also a keyword for starting blocks in Ruby.

robbyoconnor:
I have never added a do with if statements in ruby

mmainz:
Yeah, you only need that in Elixir. That's why you get this weird behaviour when you use `ruby-end-mode` for typing in Elixir.

syl20bnr:
Nice :purple_heart: 

Thank you :+1: 

jojojames:
Would this be better?

    (sp-local-pair "fn" "end"
                   :when '(("SPC" "RET"))
                   :post-handlers '(sp-ruby-def-post-handler)
                   :actions '(insert navigate))

You can have -> without the end so they aren't a true pair, with -> you'll get unintentional 'ends' added when you type code like this.

    if x = 3 do
      case y = :ok do
        :ok -> :ok
        :error -> a = "it's an error"
      end
    else
      z = 11
    end

http://elixir-lang.readthedocs.io/en/latest/technical/scoping.html

fn examples would be

fn end
fn y -> end

PR #5045, opened 2016-02-12
Add majapahit-theme

kostajh:
This is a nice theme with light & dark versions.

### DIFF
--- a/core/core-themes-support.el
+++ b/core/core-themes-support.el
@@ -145,6 +145,8 @@
     (colorsarenice-light . colorsarenice-theme)
     (hemisu-dark  . hemisu-theme)
     (hemisu-light . hemisu-theme)
+    (majapahit-dark . majapahit-theme)
+    (majapahit-light . majapahit-theme)
     (material-light . material-theme)
     (minimal-light . minimal-theme)
     (moe-dark  . moe-theme)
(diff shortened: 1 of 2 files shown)


cpaulik:
You will also have to add the lookup from the theme name to the package to core theme support.

See [here](https://github.com/syl20bnr/spacemacs/blob/ef5a2ad5ee6362588244374364fe4ac2df066f37/core/core-themes-support.el#L116-L117) for an example

kostajh:
@cpaulik thanks! I had looked at another PR adding support for a theme, and didn't see that. I'll do that and resubmit.

cpaulik:
It depends on the theme. If the theme name is the same as the package name on MELPA then it works out of the box. This step is necessary if multiple variants like a dark and a light theme are in one MELPA package.

kostajh:
That makes sense, thanks for explaining!


On Fri, Feb 12, 2016, at 10:24 AM, Christoph Paulik wrote:
> It depends on the theme. If the theme name is the same as the package
> name on MELPA then it works out of the box. This step is necessary if
> multiple variants like a dark and a light theme are in one MELPA
> package.


> — Reply to this email directly or view it on GitHub[1].


>



Links:

  1. https://github.com/syl20bnr/spacemacs/pull/5045#issuecomment-183372199

kostajh:
OK, reworked & this works fine locally. Thanks again @cpaulik

syl20bnr:
Thank you :+1: 

PR #5046, opened 2016-02-12
Add smart closing parenthesis

StreakyCobra:
When pressing `)` in insert mode, the point will go over the next closing parenthesis/bracket/quote/etc… For instance (the `|` character being the cursor position):

    [("{|}")]

pressing 6 times `)` will do:

    [("{}|")]
    [("{}"|)]
    [("{}")|]
    [("{}")]|
    [("{}")])|
    [("{}")]))|

It also takes care of creating closing surrounding as in:

    [({|
    [({}|
    [({})|
    [({})]|
    [({})])|

Let split the behavior in two cases:

**The surroundings are balanced**

- If there is a closing surrounding *on the same line*, pass over it
- Otherwise, insert the `)` character

**The surroundings are unbalanced**

- Insert the char closing the next surrounding, even if it is not open on the same line.
- I noticed one border case problem though:
  - `{"` fails if there is no newline at the end of the line, or becomes `{"}}}` otherwise.

Despite this small quirk, I think it could be added as it's an opt-in feature anyway.

BTW pressing `C-q` before a surrounding will enter it textually, this can be use when people want to write `")"` (try it :wink:) with this enabled. `" C-q ) )` works.

### DIFF
--- a/layers/+spacemacs/spacemacs-editing/packages.el
+++ b/layers/+spacemacs/spacemacs-editing/packages.el
@@ -261,7 +261,7 @@ It will toggle the overlay under point or create an overlay of one character."
 (defun spacemacs-editing/init-smartparens ()
   (use-package smartparens
     :defer t
-    :commands (sp-split-sexp sp-newline)
+    :commands (sp-split-sexp sp-newline sp-up-sexp)
     :init
     (progn
       (spacemacs/add-to-hooks (if dotspacemacs-smartparens-strict-mode
@@ -321,6 +321,22 @@ It will toggle the overlay under point or create an overlay of one character."
       (sp-pair "{" nil :post-handlers
                '(:add (spacemacs/smartparens-pair-newline-and-indent "RET")))
       (sp-pair "[" nil :post-handlers
-               '(:add (spacemacs/smartparens-pair-newline-and-indent "RET"))))))
+               '(:add (spacemacs/smartparens-pair-newline-and-indent "RET")))
 
-
+      (defun spacemacs/smart-closing-parenthesis ()
+        (interactive)
+        (let* ((sp-navigate-close-if-unbalanced t)
+               (current-pos (point))
+               (current-line (line-number-at-pos current-pos))
+               (next-pos (save-excursion
+                           (sp-up-sexp)
+                           (point)))
+               (next-line (line-number-at-pos next-pos)))
+          (cond
+           ((and (= current-line next-line)
+                 (not (= current-pos next-pos)))
+            (sp-up-sexp))
+           (t
+            (insert-char ?\))))))
+      (when dotspacemacs-smart-closing-parenthesis
+          (define-key evil-insert-state-map ")" 'spacemacs/smart-closing-parenthesis)))))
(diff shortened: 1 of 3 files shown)


StreakyCobra:
Also let me know if this should also be enabled for emacs mode users.

AlejandroCatalina:
Since it is customizable, why not adding it to emacs users too :P. Lispy also works by implicit *special positions* and it's pretty nice for emacs users, so IMO it'd be nice to have it for emacsers too.

layers/+spacemacs/spacemacs-editing/packages.el
```diff
@@ -321,6 +321,25 @@ It will toggle the overlay under point or create an overlay of one character."
...
       (sp-pair "[" nil :post-handlers
-               '(:add (spacemacs/smartparens-pair-newline-and-indent "RET"))))))
-
+               '(:add (spacemacs/smartparens-pair-newline-and-indent "RET")))
 
+      (defun spacemacs/smart-closing-parenthesis ()
+        (interactive)
+        (let* ((sp-navigate-close-if-unbalanced t)
+               (current-pos (point))
+               (current-line (what-line))
```
fbergroth: `(what-line)` is meant for interactive use, you probably want `(line-number-at-pos)` here.

layers/+spacemacs/spacemacs-editing/packages.el
```diff
@@ -321,6 +321,25 @@ It will toggle the overlay under point or create an overlay of one character."
...
+               (next-line (save-excursion
+                            (sp-up-sexp)
+                            (what-line))))
+          (cond
+           ((and (string-equal current-line next-line)
+                 (not (= current-pos next-pos)))
+            (sp-up-sexp))
+           (t
+            (insert-char ?\)))
+           )))
```
fbergroth: Help, we're dangling! Save us :)

layers/+spacemacs/spacemacs-editing/packages.el
StreakyCobra: Thanks for the info, it's indeed better!

layers/+spacemacs/spacemacs-editing/packages.el
StreakyCobra: The family has been regrouped :-)

person808:
Is this like https://github.com/emacs-berlin/general-close?

StreakyCobra:
@person808 Probably. @TheBB mentioned it on a gitter discussion if I recall correctly.

Do you have any experience with it? If it works better I'll update my PR to use it instead, my code is more a kind of proof-of-concept :-)

person808:
Never used it, was just wondering

StreakyCobra:
@person808 I tested it a little bit, it seems that `general-close` seems to close an opened surrounding, but not passing over an already closed one, i.e:

This is working:

    [({|
    [({}|
    [({})|
    [({})]|

This is not working:

    [("{}|")]
    [("{}|")]
    [("{}"|)]
    [("{}")|]
    [("{}")]|

Also it seems to have problem adding closing parenthesis alone, i.e.

    []
    []) <- not possible

syl20bnr:
This is great, may take some time to have the reflex to use `C-q` sometimes but I see the benefit of such feature.

Thank you :+1: 

PR #5048, opened 2016-02-12
install-package: don't set lazy-install pkg is nil

### LINKED ISSUE #5020: wrong-type-argument eieio-object-p nil
I keep getting this after updating packages.  Removing elpa fixes it, but that would only be ok if it happened once or twice.  It happens often.

What gives?

#### System Info
- OS: darwin
- Emacs: 24.5.7
- Spacemacs: 0.105.9
- Spacemacs branch: develop (rev. de7eb40)
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(unscroll auto-completion xkcd emacs-lisp git dash yaml github javascript html markdown extra-langs python ruby org osx python django syntax-checking spell-checking version-control lua colors spacemacs-layouts vimscript spacemacs-helm
          (c-c++ :variables c-c++-enable-clang-support t)
```

bmag:
When the list of package names returned by `configuration-layer//get-uninstalled-packages` contains a package that isn't contained in `configuration-layer--packages`, then `configuration-layer//install-package` throws a "wrong-argument-type: eieio-object-p nil" error at startup. `configuration-layer//install-packages` catches the error and prints it in the home buffer.

This PR fixes `configuration-layer//install-package` to not set the `lazy-loaded` property of a package, when Spacemacs doesn't have an entry for that package in `configuration-layer--packages`.

Fixes #5020.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -724,7 +724,7 @@ path."
           (cond
            ((or (null pkg) (eq 'elpa location))
             (configuration-layer//install-from-elpa pkg-name)
-            (oset pkg :lazy-install nil))
+            (when pkg (oset pkg :lazy-install nil)))
            ((and (listp location) (eq 'recipe (car location)))
             (configuration-layer//install-from-recipe pkg)
             (oset pkg :lazy-install nil))


syl20bnr:
Nice catch :purple_heart: 

Thank you :+1: 

PR #5062, opened 2016-02-14
evil-magit: Use horizontal movement option

justbur:
New option in evil-magit that moves h to H, l to L, and L to C-l to
allow h and l to move horizontally in the buffer.

This is probably a little more natural for vim users.

### DIFF
--- a/layers/+source-control/git/packages.el
+++ b/layers/+source-control/git/packages.el
@@ -30,6 +30,7 @@
 (defun git/init-evil-magit ()
   (unless (eq dotspacemacs-editing-style 'emacs)
     (with-eval-after-load 'magit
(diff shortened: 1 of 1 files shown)


nixmaniack:
:+1:

StreakyCobra:
Two questions:

- AFAIK all operations on magit are done line by line. So is moving with `h` and `l` really needed as a default?
- Being set in a `with-eval-after-load` of `magit` within the `init` of `evil-magit`, is it still easy do disable it from user configuration?

justbur:
I'm torn on this one. I think that @syl20bnr will want it, but on second thought I think it would be better to make people aware of the option through the docs instead of making it the default.

PR #5066, opened 2016-02-14
Restart Spacemacs restoring the layouts

### LINKED ISSUE #5058: spacemacs/restart-emacs should remember the open files
Since emacs already have support for this (desktop-save) it shouldn't be particular hard?

StreakyCobra:
With this change:

- `SPC q r` will restart emacs restoring the layouts.
- `SPC q R` will restart emacs without restoring the layouts.

I put restart resuming the layouts on `SPC q r` as it's probably the intended users' behavior, and used `SPC q R` for restart without resuming the layouts as it's a more destructive operation.

Also it's possible now to use `emacs --resume-layouts` to temporary start an emacs instance resuming the previous layouts.

Note: Setting directly `dotspacemacs-auto-resume-layouts` in the command line parsing is not possible because it get overridden when parsing the `.spacemacs`. I created a new variable instead.

Fix #5058

### DIFF
--- a/core/core-command-line.el
+++ b/core/core-command-line.el
@@ -9,6 +9,10 @@
 ;;
 ;;; License: GPLv3
 
+(defvar spacemacs-force-resume-layout nil
+  "If non-nil force the current emacs instance to resume layouts
+  at start time despite the value of `dotspacemacs-auto-resume-layouts'.")
+
 (defun spacemacs//parse-command-line (args)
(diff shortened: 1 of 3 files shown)


syl20bnr:
:+1: 
Makes me think we need to document the Spacemacs specific command line options at some point before 0.106.0 (I have no date for this version though).

syl20bnr:
Thank you :+1: 

PR #5075, opened 2016-02-15
fix renamed cider-test-run-tests function

AlejandroCatalina:
The function was renamed in a recent commit to `cider-test-run-ns-tests`, see
https://github.com/clojure-emacs/cider/commit/568263143bed3bfdb15c661d6b0c5b69aa9e972e#diff-f8e2b8e5fa0f8c5cda946e72ab461994R572
for more info.

It was reported on gitter, so let's update it.

### DIFF
--- a/layers/+lang/elm/packages-funcs.el
+++ b/layers/+lang/elm/packages-funcs.el
@@ -10,30 +10,25 @@
 ;;; License: GPLv3
 
 
+;; elm-mode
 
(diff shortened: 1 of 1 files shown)


syl20bnr:
Nice :+1: 
Thank you again!
Merged in develop, I keep this PR opened because of the required hotfix for 0.105.

PR #5078, opened 2016-02-15
remove structured-haskell-mode

d12frosted:
Just as the title says - by this commit we remove shm. The main reason is that shm doesn't play well with vim editing style and causing more problems that are not so easily fixed. But still, it's very easy to add it to personal configs if you are using emacs editing style. 

Related issues: #5026, #1048 and #3171(this one has more explanations). 

> officially Chris Done marked `shm` and `evil` as mode-conflict and wont-fix

P. S. Do I need to add anything about how to enable shm to haskell layer readme? With warning 'use it for your own risk and preferably with emacs editing style' :) 

P. P. S. @syl20bnr thanks for the green light :)

### DIFF
--- a/layers/+lang/haskell/README.org
+++ b/layers/+lang/haskell/README.org
@@ -16,7 +16,6 @@
      - [[GHCi-ng support][GHCi-ng support]]
        - [[Stack users][Stack users]]
        - [[ghc-mod users][ghc-mod users]]
-     - [[structured-haskell-mode][structured-haskell-mode]]
      - [[hindent][hindent]]
  - [[Key bindings][Key bindings]]
    - [[Documentation][Documentation]]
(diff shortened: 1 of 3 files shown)


d12frosted:
I believe that some people would not like this decision, but again, if someone has a good solution for this problem - PRs are warmly welcome. I think that shm is incredibly great thing, it's just doesn't work with vim editing style as good as it works with emacs editing style. In case Chris Done reading this - thanks for shm - it makes me happy when I code in Haskell. :D

joehillen:
:+1:

syl20bnr:
We should mention shm in the README and explain our motivation behind the removal so people are aware of it and may try to fix it.

robbyoconnor:
Why not conditionally install it based on editing style?

robbyoconnor:
Why not conditionally install it based on editing style? Rather than remove it altogether.

robbyoconnor:
Also this is against `master` and not `develop` @d12frosted  ;) You know the drill :stuck_out_tongue:

jb55:
:+1: after trying it out multiple times already. Until there's evil support, it's probably better to just install it with additional packages.

d12frosted:
> Also this is against master and not develop @d12frosted ;) You know the drill

Just wow. Haha. My bad, will redo (in several hours). :-P

> Why not conditionally install it based on editing style?

Sure we can install it only for emacs editing style (and even not for hybrid). There are actually 3 (or 2) configs that will be common for everyone. And then comes some per-person configs. Also, shm is not supported for all themes. For example, I asked nashamri to update Spacemacs theme so it supports shm highlights. But from my experience only few themes has this support - in other themes it would look like broken thing. So I think that shm configs are personal. I will just add notes about what everyone wants. How does it sound to you, @robbyoconnor?

> We should mention shm in the README and explain our motivation behind the removal so people are aware of it and may try to fix it.

Ok, I'll write up some notes.

d12frosted:
Closing this PR in order to make @robbyoconnor happy :) 

P. S. will reopen it very soon.

StreakyCobra:
> Also this is against master and not develop

@d12frosted You don't contribute enough often to avoid this kind of mistake it seems :wink:

d12frosted:
@StreakyCobra yeah, it seems so - hope to fix it soon :)

PR #5081, opened 2016-02-15
spacemacs-editing-vis: Add anzu explicitly

justbur:
anzu was pulled in by evil-anzu, but it's useful for all users so we
should configure it as such. This change only affects those who exclude
the spacemacs-evil layer but include spacemacs-editing-visual

### DIFF
--- a/layers/+spacemacs/spacemacs-editing-visual/packages.el
+++ b/layers/+spacemacs/spacemacs-editing-visual/packages.el
@@ -13,6 +13,7 @@
       '(
         ;; default
         adaptive-wrap
+        anzu
         auto-highlight-symbol
         highlight-indentation
         highlight-numbers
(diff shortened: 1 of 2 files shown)


bmag:
Nice one. How about remapping `query-replace` and `query-replace-regexp` to `anzu-query-replace` and `anzu-query-replace-regexp` in the global key map?

justbur:
@bmag see what you think now

bmag:
I tested it a bit and it looks good :+1:  I've got just a few more minor nitpicks.
- A prefix of `sr` or `sR` for "search-replace" instead of "text-replace" seems more intuitive to me, but that's purely subjective I guess. `xr` is also fine.
- might want to bind `anzu-query-replace-at-cursor` as well. It query-replaces all occurrences of the thing at point from point to end of buffer. ("query" here means to ask the user for confirmation for each occurrence).
- IMO `anzu-replace-at-cursor-thing` is better than `anzu-query-replace-at-cursor-thing`. They both replace just the one occurrence at point, so querying for confirmation is kind of redundant.

justbur:
@bmag Good suggestions. I changed the prefix.

I have the reverse behavior between the -at-cursor and the -at-cursor-thing variants though. -at-cursor-thing replaces all occurrences for me.

I don't see the point of replacing just the one occurrence at the point. That's not really search and replace is it?

bmag:
Hmm... -at-cursor-thing behaves differently for me in python and elisp buffers. In python it does replace all, in elisp only the current one. I'm not sure what's the difference between the two, now.

I could see myself using occasionally a command for replacing just one occurrence quickly, if I was using emacs editing style. But since I was wrong about the command's behaviour (it replaces more than once) the point is moot.

justbur:
maybe @syohex can explain the difference

syohex:
> Hmm... -at-cursor-thing behaves differently for me in python and elisp buffers. In python it does replace all, in elisp only the current one. I'm not sure what's the difference between the two, now.

Could you show me sample python code and tell me how to reproduce your issue ? (anzu.el uses `bounds-of-thing-at-point` in such commands for finding beginning and end of thing)

syl20bnr:
Do we really need to put those functions under `SPC s r` as well ? I find the UX of these commands awkward and non intuitive so let's keep only the remaps.

bmag:
@syl20bnr I'm fine with only the remaps.

@syohex I seem to have misunderstood how -at-cursor-thing works. I didn't realize it's supposed to replace only in the current function (it seems that thing=function in this context).

sample python file: (`|` is where point is at)
```python
def plus1(|num):
    return num + 1

def plus2(num):
    return num + 2
```
`anzu-query-replace-at-cursor-thing` replaces both occurrences of `num` in `plus1`. `anzu-query-replace-at-cursor` replaces all 4 occurrences of `num`.

sample elisp file: `anzu-20160130.1855/anzu.el `, point in `anzu-query-replace-at-cursor` definition, before the `a` of `anzu--query-replace-common`.
`anzu-query-replace-at-cursor-thing` replaces only the one occurrence inside of `anzu-query-replace-at-cursor`, while `anzu-query-replace-at-cursor` replaces all 5 occurrences from point to end of buffer.

syohex:
Ah, my command naming is confusing, sorry.

justbur:
@syl20bnr rebased and removed the leader bindings

@syohex I understand now. Thanks for the explanation. I didn't see that `anzu-replace-at-cursor-thing` is a custom variable.

justbur:
It's not necessary to do this as a separate package if everyone will be using evil again

PR #5082, opened 2016-02-15
google-translate: Cleanup use-package declaration

justbur:
Declare :commands so that the package is properly deferred. Also wrap
the whole init section in progn


syl20bnr:
Thank you again :-)

PR #5083, opened 2016-02-15
Explain helm-descbind prompt more detailed

triklsbg:
To prevent missunderstandings, times before "SPC" meant the spacebar. This time not.
"a blank" is still missing in the enumeration of the letters to type. Please correct me.

### DIFF
--- a/doc/QUICK_START.org
+++ b/doc/QUICK_START.org
@@ -145,7 +145,7 @@ type a pattern like this regular expression:
     SPC\ b
 #+END_EXAMPLE
 
-which would list all =buffer= related bindings.
+which would list all =buffer= related bindings. *Note:* You are at the /HELM-Descbind/ prompt, the pattern consists of 6 letters: uppercase ~SPC~, a backslash and a lowercase ~b~.
 
 ** Describe functions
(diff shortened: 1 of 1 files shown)


syl20bnr:
I'm not sure to understand what you mean here. What problem do you want to resolve  ?

Maybe a screenshot would be better ?

triklsbg:
There is no "real problem". My intention is to point out the difference between the Spacebar denoted as "SPC" and the letterwise "SPC" in the meaning 3 single letters which must be typed at the pattern prompt for the firstime reader. Background: My personal attempt ended in the mistake that i typed the Spacebar when i saw SPC\ b instead of the single uppercase letters... The text in the doc is correct but i think an additional explanation helps to avoid misunderstandings.

triklsbg:
erroneously closed.

syl20bnr:
Ok, thank you, I understand now.

TheBB:

I made some further clarifications in https://github.com/syl20bnr/spacemacs/commit/2442431d92307cd86670141ebfd579d129216bf9, as well as splitting the long line.

Issue #5060, opened 2016-02-14
Exiting symbol editing mode puts Emacs in Vim style editing


shivamkalra:
#### Description
Existing symbol editing mode puts Emacs in Vim style mode. 

#### Reproduction guide
- Start Emacs
- Press `M-m s e` to enter symbol editing mode
- Press `e` to make an edit
- Do editing and exiting symbol editing mode by pressing escape.

*Observed behaviour:*
It puts Emacs in Vim mode and cannot switch back to Emacs style editing.

*Expected behaviour:*
It should come back to whichever editing style user was using before.

![result](https://cloud.githubusercontent.com/assets/1386352/13030950/fed51fc4-d288-11e5-94c7-399f6d1658b8.gif)

#### System Info
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.9
- Spacemacs branch: develop (rev. 9f862c8)
- Distribution: spacemacs
- Editing style: emacs
- Completion: helm
- Layers:
```elisp
((auto-completion :variables auto-completion-enable-help-tooltip t)
 better-defaults spacemacs-helm emacs-lisp gtags c-c   version-control git markdown org gnus shell syntax-checking csharp sql csharp lua elpy xkcd haskell extra-langs emoji spell-checking ipython-notebook
 (erc :variables erc-sasl-server-regexp-list
      '("irc\.freenode\.net"))
 shell-scripts theming javascript html colors ibuffer smex github themes-megapack fasd wakatime notmuch mu4e ipython-notebook matlab search-engine tools
 (elfeed :variables rmh-elfeed-org-files
         (list "~/.emacs.d/private/elfeed.org"))
 (latex :variables latex-enable-auto-fill t))
```

StreakyCobra:
I think it is expected right now because `M-m s e` is bound to `evil-iedit-state` which is meant to work with `evil`. Can you try with [`C-;`](https://github.com/victorhge/iedit/blob/2d9017daeaeb408568992f5c9cad5df5ca4f1d12/iedit.el#L97) instead?

shivamkalra:
@StreakyCobra Thanks. Yes, `C-;` works way I would expect it to. However, wouldn't it be better if `M-m s e` could decide this based on `dotspacemacs-edit-style` variable?

```
(if (eq dotspacemacs-editing-style 'emacs)
           (iedit-mode)
         (evil-iedit-state))
```

StreakyCobra:
[...]

PR #5091, opened 2016-02-16
iedit-mode: bind iedit-mode for emacs users

AlejandroCatalina:
If the user has emacs editing style, bind iedit-mode instead.

I don't know if this is the right place to put this, i thought so as here is where the binding is being set. Anyway, any discussion is warmly welcome.

Address #5060

### DIFF
--- a/layers/+spacemacs/spacemacs-evil/packages.el
+++ b/layers/+spacemacs/spacemacs-evil/packages.el
@@ -69,7 +69,9 @@
 (defun spacemacs-evil/init-evil-iedit-state ()
   (use-package evil-iedit-state
     :commands (evil-iedit-state evil-iedit-state/iedit-mode)
-    :init (spacemacs/set-leader-keys "se" 'evil-iedit-state/iedit-mode)
+    :init (spacemacs/set-leader-keys "se" (if (eq dotspacemacs-editing-style 'emacs)
+                                              'iedit-mode
+                                            'evil-iedit-state/iedit-mode))
     :config
     ;; activate leader in iedit and iedit-insert states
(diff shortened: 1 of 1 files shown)


justbur:
This is not the way I would do this, but @syl20bnr is probably going to disagree with me. IMO emacs users should not be using the spacemacs-evil layer at all since they are currently not using evil-mode. Therefore the binding should go in a place where it will affect emacs style users. 

There is also the issue of the `e` command in the symbol highlight transient state that needs to be fixed.

AlejandroCatalina:
I agree with that, but to my surprise spacemacs-evil is loaded even if I'm using emacs editing style, so I thought this would make sense.

If the way this is changes, I'd happily do it other way.

On Feb 16, 2016 4:55 PM, Justin Burkett <notifications@github.com> wrote:

This is not the way I would do this, but @syl20bnr is probably going to disagree with me. IMO emacs users should not be using the spacemacs-evil layer at all since they are currently not using evil-mode. Therefore the binding should go in a place where it will affect emacs style users. 

There is also the issue of the e command in the symbol highlight transient state that needs to be fixed.

justbur:
I can't remember why now but that's how @syl20bnr wanted to do it.

syl20bnr:
I agree with the PR as long as it does not prevent from switching the editing style on the fly which seems to be the case here :+1:

As for the current split of packages, there is nothing really decided yet and I would like to keep this subject for later so don't make any assumptions based on this split.

syl20bnr:
Thank you :+1: 
I made the modification for `ahs`.

PR #5093, opened 2016-02-16
Remove H and L remap (fixes #4645)

### LINKED ISSUE #4645: orgmode H and L mapping should be removed
In orgmode H and L are mapped to start and end of line.
This conflict with normal vim behavior (and therefore other mode) where H and L go at the top and bottom of the page.
Moreover, these mapping are unnecessary as the same can already by achieved with ^,0 and $.
Unless I'm missing something ...

choppsv1:


syl20bnr:
Thank you :+1: 

PR #5094, opened 2016-02-16
Fix toggle-maximize-buffer

justbur:
Was using the symbol _ as the name of a register when the docs for
register-alist state that the name of a register should be a character.


justbur:
I should mention that this fixes an error when using evil-show-registers

syl20bnr:
Nice catch :+1: 

Thank you !

PR #5103, opened 2016-02-17
Allow `tsfmt` to use the local project's config.

overminder:
By adding `--baseDir=.` to the command line arguments. This option is documented [here](https://github.com/vvakame/typescript-formatter).

### DIFF
--- a/layers/+lang/typescript/packages.el
+++ b/layers/+lang/typescript/packages.el
@@ -89,7 +89,7 @@
                           (progn
                             (with-current-buffer outputbuf (erase-buffer))
                             (write-region nil nil tmpfile)
-                            (if (zerop (apply 'call-process "tsfmt" nil outputbuf nil (list tmpfile)))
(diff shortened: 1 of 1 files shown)


StreakyCobra:
Wouldn't it better to set the `baseDir` to `(projectile-project-root)` if the file is in a projectile project? I don't know anythink about typescript, so maybe what I'm saying is just nonsense…

overminder:
Thanks for your suggestion! I actually just started to use spacemacs and don't know most of the existing conventions :P

I will definitely look into that. How does projectile find the root directory of a project? I currently only know that it looks for `.git` and `.projectile`.

StreakyCobra:
> I currently only know that it looks for .git and .projectile.

It's doing kind of this I think.

So if it make sense to set it to the project's root, you should check if there is a project with `(projectile-project-p)`, if yes set it to the project root, if not set it to the current file's folder. I'm not really sure because I don't know typescript, so please double-check :-)

syl20bnr:
Thank you !

PR #5104, opened 2016-02-17
org: move global keys from :config to :init

nixmaniack:
Allows autoloading of org via these commands.


TheBB:

PR #5109, opened 2016-02-17
rustfmt: Remove obsolete package

fbergroth:
This package was merged into rust-mode.

Edit: instead of marking `rust-enable-rustfmt-on-save` as an obsolete variable, I simply deleted it since it never reached the master branch.


TheBB:

PR #5110, opened 2016-02-17
update lua readme

d12frosted:
Add a note about luacheck

---

Didn't know that this layer can do this :) Probably good thing to mention.

### DIFF
--- a/layers/+lang/lua/README.org
+++ b/layers/+lang/lua/README.org
@@ -11,20 +11,21 @@
    - [[Commands][Commands]]
 
 * Description
-
 This layer adds support for editing Lua.
 
 ** Features:
 - Editing lua files using [[https://github.com/immerrr/lua-mode][lua-mode]]
 - Sending code to a lua REPL
+- Code linting using [[https://github.com/mpeterv/luacheck][Luacheck]]
 
 * Install
 To use this configuration layer, add it to your =~/.spacemacs=. You will need to
 add =lua= to the existing =dotspacemacs-configuration-layers= list in this
 file.
 
-* Key Bindings
+In order to enable code linting, install [[https://github.com/mpeterv/luacheck][Luacheck]].
 
+* Key Bindings
 ** Commands
 
 | Key Binding | Description                                |


fbergroth:
Flycheck is likely already supported via the lua compiler checker.

You can run `SPC e v` in any buffer to see what checkers are available.

d12frosted:
> Flycheck is likely already supported via the lua compiler checker.

Don't really understand what you mean. 

In any case, I think it worth mentioning external dependency.

fbergroth:
I just thought that the wording was a bit strange since basic syntax checking already works with the luac checker. 

I don't disagree that mentioning the other checker is good. :)

robbyoconnor:
:+1:

d12frosted:
> basic syntax checking already works with the luac checker

what is `luac checker`? 

---

Syntax check works only when `luacheck` is installed. See [flycheck](https://github.com/flycheck/flycheck/blob/d4c302c09fdbbc2190a72406d71bfb9b3dcd5c45/flycheck.el#L7222). 

And this is really funny, because I've read lua layer readme and didn't find any mention about syntax checker. So I started to look for one. Then I found `luacheck` and decided to try. Imagine how was I surprised to see it working automatically in Spacemacs. That's why I've added that line - I want to save other people time :)

robbyoconnor:
Yeah all flycheck does is run the program and parse its output using a regular expression

fbergroth:
Sorry for the confusion. Flycheck comes with two checkers for lua, the other checker called [lua](https://github.com/flycheck/flycheck/blob/d4c302c09fdbbc2190a72406d71bfb9b3dcd5c45/flycheck.el#L7245-L7256).

You wrote
>In order to enable =flycheck= support ...

The lua checker will probably be enabled for most people, so I meant to suggest that it should written like "In order to enable linting via Luacheck ...".

d12frosted:
> Yeah all flycheck does is run the program and parse its output using a regular expression

Yes, I know. But it's not the point of our discussion. See my response to @fbergroth 

> Flycheck comes with two checkers for lua

Wow, didn't know. But why then it didn't work? 

> The lua checker will probably be enabled for most people

Indeed, you are right then :) Right, I'll change that. Thanks for information :)

fbergroth:
> Wow, didn't know. But why then it didn't work?

See enabled checkers with `SPC e v`. If you have troubles with a particular checker, you can run `M-x flycheck-compile`. Probably you just don't have `luac` in your path. :)

d12frosted:
@fbergroth I've updated it :) 

Thanks for the help and for `SPC e v`.

TheBB:

PR #5111, opened 2016-02-17
[RDY] Don't set insert-state for emacs users

AlejandroCatalina:
When switching to repl in python-mode don't set evil-insert-state if the
user is using emacs editing style.

I don't know if this happens in other repls, i'll look for them :P.

### DIFF
--- a/layers/+distribution/spacemacs-base/packages.el
+++ b/layers/+distribution/spacemacs-base/packages.el
@@ -226,6 +226,12 @@
       (defun evil-insert-state-cursor-hide ()
         (setq evil-insert-state-cursor '((hbar . 0))))
 
+      (defun spacemacs/normal-to-insert-state ()
+        "Switches to evil-insert-state if the current state is
+evil-normal state."
+        (when (evil-normal-state-p)
+          (evil-insert-state)))
+
       (unless (eq dotspacemacs-editing-style 'emacs)
         (evil-mode 1)))
     :config
--- a/layers/+lang/clojure/packages.el
+++ b/layers/+lang/clojure/packages.el
@@ -57,7 +57,7 @@ the focus."
 `insert state'."
         (interactive)
         (cider-insert-last-sexp-in-repl t)
-        (evil-insert-state))
+        (spacemacs/normal-to-insert-state))
 
       (defun spacemacs/cider-send-region-to-repl (start end)
         "Send region to REPL and evaluate it without changing
@@ -72,7 +72,7 @@ the focus."
         (interactive "r")
         (cider-insert-in-repl
          (buffer-substring-no-properties start end) t)
-        (evil-insert-state))
+        (spacemacs/normal-to-insert-state))
 
       (defun spacemacs/cider-send-function-to-repl ()
         "Send current function to REPL and evaluate it without changing
@@ -85,7 +85,7 @@ the focus."
 `insert state'."
         (interactive)
         (cider-insert-defun-in-repl t)
-        (evil-insert-state))
+        (spacemacs/normal-to-insert-state))
 
       (defun spacemacs/cider-send-ns-form-to-repl ()
         "Send buffer's ns form to REPL and evaluate it without changing
@@ -98,7 +98,7 @@ the focus."
 `insert state'."
         (interactive)
         (cider-insert-ns-form-in-repl t)
-        (evil-insert-state))
+        (spacemacs/normal-to-insert-state))
 
       (defun spacemacs/cider-send-buffer-in-repl-and-focus ()
         "Send the current buffer in the REPL and switch to the REPL in
@@ -106,7 +106,7 @@ the focus."
         (interactive)
         (cider-load-buffer)
         (cider-switch-to-repl-buffer)
-        (evil-insert-state))
+        (spacemacs/normal-to-insert-state))
 
       (defun spacemacs/cider-test-run-focused-test ()
         (interactive)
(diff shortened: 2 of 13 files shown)


justbur:
There's nothing wrong with this pr, but I'd like to propose a different approach to this sort of problem. 

@syl20bnr has said that he wants users to be able to switch between editing styles during a session by toggling hybrid-mode and/or holy-mode. We don't currently change the value of dotspacemacs-editing-style when this happens though.

Therefore it would make more sense to either check for whether holy-mode is active at the moment in a case like this, or the more fool proof way is to check to see if evil-local-mode is active, which is what you really want to know.

AlejandroCatalina:
That's right and that would be much more accurate. What's the variable that stores the current editing style?

On Feb 17, 2016 16:50, Justin Burkett <notifications@github.com> wrote:

There's nothing wrong with this pr, but I'd like to propose a different approach to this sort of problem. 

@syl20bnr has said that he wants users to be able to switch between editing styles during a session by toggling hybrid-mode and/or holy-mode. We don't currently change the value of dotspacemacs-editing-style when this happens though.

Therefore it would make more sense to either check for whether holy-mode is active at the moment in a case like this, or the more fool proof way is to check to see if evil-local-mode is active, which is what you really want to know.

justbur:
I believe the intention is just to not enable evil when it's not active already, so I would do that exactly like this
```elisp
(when evil-local-mode
  (evil-insert-state))
```

You could also check if `holy-mode` is active with this
```elisp
(when holy-mode
  (evil-insert-state))
```
but I prefer the former, because someone could quite reasonably disable `evil-local-mode` in all repl buffers but use vim editing style. The first case accounts for that, but the latter doesn't

AlejandroCatalina:
Yeah checking `evil-local-mode` I think it's better. I'll update when I get home ;). Thanks!

justbur:
I just did a quick search and there's a bunch of cases where a spacemacs function uses `(evil-insert-state)`, `(evil-normal-state)`, etc. 

Maybe we can correct this in the evil repo. Let me check.

justbur:
Never mind, I guess it makes sense to enable evil when these functions are called, but we should check for all of them. 

Maybe a good solution here is to write functions like `spacemacs/safe-evil-insert-mode` that check the appropriate cases centrally, and then replace all corresponding points in the code with these versions.

AlejandroCatalina:
I agree, enable it unless holy mode is on. That fiction should be in the core right?

AlejandroCatalina:
Function*

justbur:
I'd say in the evil configuration would be the first place I would look for it

StreakyCobra:
> That fiction should be in the core right?

I would probably add it to the `config` of `init-evil`.

AlejandroCatalina:
Yes that makes sense. I'm on it. Let's hear @syl20bnr on this.

syl20bnr:
Not sure where to put this function. If we put it in `init-evil` then the `spacemacs-evil` layer becomes mandatory and currently I have no clear idea where we are going with it so I guess a better location is in `core-funcs.el`, this brings only the evil package dependency instead of the whole evil layer and the current version of Spacemacs needs evil no matter what, so putting the new functions in core is OK to me.

justbur:
To clarify something `spacemacs-evil` does not include `evil`. It's the packages that depended on evil that were in `spacemacs` and *not* `spacemacs-base`. `evil` was and still is in `spacemacs-base`, so it is not optional at the moment.

syl20bnr:
Oh ok, so indeed we can put them with `init-evil` but maybe in `:init` instead of `:config` to be sure the functions are defined when Emacs finishes loading.

justbur:
I wasn't voting one way or the other. I just thought there was some
confusion on what spacemacs-evil was.

AlejandroCatalina:
Moved to :init and rebased @syl20bnr

justbur:
@AlejandroCatalina I would go ahead and fix all of these instances if you can. If you search the repo there's a bunch of similar calls to `evil-insert-state` when switching to repls (clojure, elm, fsharp, etc)

Also, to be even more careful about how this works, I had the thought that we should only force insert state if `evil-local-mode` is active and we're in normal-state (someone might want to be in evil's emacs state for example), so instead of checking `evil-local-mode` you could just check the value of `(evil-normal-state-p)`.

AlejandroCatalina:
I'll go for it, that totally makes sense. I'll check the repls and so. Any suggestion welcome.

AlejandroCatalina:
@justbur done, i think i didn't miss anyone, but tell me if you or anyone else find other ocurrence.

justbur:
@AlejandroCatalina looks ok to me. One more nitpick, sorry, you should explain in the docstring why it's "safe", or now that the function is more specific maybe we can call it `spacemacs/normal-to-insert-state` or something more descriptive like that. What do you think?

AlejandroCatalina:
Yeah that'd suit the function better now. I forgot the docstring so I'll add it later.

Thanks!

AlejandroCatalina:
@justbur is it fine now? :D

justbur:
:+1: nice

AlejandroCatalina:
@syl20bnr this is ready to me unless more improvements are required.

TheBB:
Shouldn't we be setting the initial state for those modes instead? I'm one of those weirdos who like to end up in normal state when I open a REPL and I use evil's own configuration options to make sure that happens.

Anyway this PR looks fine to me. It doesn't break anything that wasn't already broken at least, with respect to the above. :-D

TheBB:

Issue #4892, opened 2016-01-31
mu4e: Enabling mu4e-maildirs-extension by default


zakkak:
Hello @all,

In bb07eeccade0af51b15b7093d4856fe4f2fdb532 the [`mu4e-maildirs-extension`](https://github.com/agpchil/mu4e-maildirs-extension) is enabled by default in the mu4e layer.

I personally find this extension too intrusive/noisy to be enabled by default. I believe that the default options should reflect the common use case and I don't believe that most mu4e users do want mu4e-maildirs-extension, but I might be totally wrong so I would like to hear what the rest of you think.

mskorzhinskiy:
I'm also found this extension too noisy, and I think that it should be disabled by default.

jackbaty:
I'm curious what you mean by "common use case" that would cause most users to want it disabled. I find the extension a useful and obvious (to me) feature of my mail reader. I might have missed it if it needed to be enabled separately (yes, that would have been my fault, but still). 

My experience was improved by it being enabled by default.

zakkak:
I also noticed that the navigation in `mu4e-main-mode` is broken for `evil-mode` users, since `j` is overridden to allow users to jump to directories (the default binding in `mu4e-main-mode`.  As a result, `evil-mode` users cannot navigate to the folders listing of `mu4e-maildirs-extension` using `evil-mode` navigation keybindings.

@jackbaty with "common use case" I refer to the perspective I have obtained about how people use and configure mu4e.  I have obtained this perspective by using mu4e and discussing with other users in its repository and mailing list.  As I point out in my first post, this perspective might be totally wrong and that is why I want to discuss it with everybody else here. Of course if most people think it should be enabled by default, then that should be the case.

PS: In general the use of folders, although intuitive, is

jackbaty:
Thanks for the clarification, @zakkak. I'm (slowly) coming around to the archive-and-search mindset, after years of managing dozens of folders via fcc-hooks etc. in Mutt. 

Thinking about it further, having the maildirs extension enabled by default might encourage people to use mu4e in a way that it wasn't designed for. That, along with the idea of a "common use case" makes me lean toward your suggestion of disabling it by default.

nixmaniack:
I already discussed my opinion about this in that commit comment thread. 

[...]

PR #5112, opened 2016-02-17
mu4e: Don't use maildirs-extension by default.

JorisE:
The maildirs extension is no longer turned on by default. It is now
installed only if the variable mu4e-use-maildirs-extension is non-nil.

Related issue: #4892

### DIFF
--- a/layers/+email/mu4e/packages.el
+++ b/layers/+email/mu4e/packages.el
@@ -70,12 +70,12 @@
               (mu4e-alert-enable-mode-line-display)))))
 
 (defun mu4e/init-mu4e-maildirs-extension ()
+  "If mu4e-use-maildirs-extension is non-nil, set
+mu4e-use-maildirs-extension-load to be evaluated after mu4e has been loaded."
   (use-package mu4e-maildirs-extension
-    :defer t
+    :if mu4e-use-maildirs-extension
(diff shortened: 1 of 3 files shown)


AlejandroCatalina:
Like it :+1:

layers/+email/mu4e/packages.el
```diff
@@ -73,12 +73,13 @@
               (mu4e-alert-enable-mode-line-display)))))
 
 (defun mu4e/init-mu4e-maildirs-extension ()
+  "If mu4e-use-maildirs-extension is non-nil, set
+mu4e-use-maildirs-extension-load to be evaluated after mu4e has been loaded."
   (use-package mu4e-maildirs-extension
     :defer t
-    :init (with-eval-after-load 'mu4e (mu4e-maildirs-extension-load))))
+    :init (when mu4e-use-maildirs-extension
```
StreakyCobra: Wouldn't it better to use `:if` in `use-package` instead?

JorisE: I did not know about that option, but it looks like a good idea.

JorisE:
Would it be rude to ask if anyone can take a look at this?

syl20bnr:
Thank you! 💜 

My sincere apologies, this PR has not been merged in due time and has been replaced by a more recent PR from you 😄 (https://github.com/syl20bnr/spacemacs/pull/9586).

PR #5114, opened 2016-02-17
Fixes #4490 - mu4e: Use evil-mu4e to set up keybindings.

### LINKED ISSUE #4490: mu4e - autoevilification makes it difficult to work with; needs better keybindings
The auto-evilification of mu4e makes the layer rather difficult to work with as some of the more important bindings get very inconvenient compared to the ones in the mu4e manual (http://www.djcbsoftware.nl/code/mu/mu4e/Keybindings.htm)
# Problems:

1) Moving to previous/next message from the search results is pretty important to process email quickly. Without evilification this is bound to `n`,`p`. Both easily accessible; after evilification this become `C-n` and `p`. Note the lack of symmetry and moreover it makes one of the most used commands a difficult emacs-like key combo, i.e. `C-n` instead of `n`. Way too painful.

2) Automatic remapping of keys in the mu4e layer fails for some of the keys as the mu4e layer emits the following warning message in the message buffer:

```
(Spacemacs) Warning: Auto-evilication could not remap these functions in map `mu4e-headers-mode-map':
   - `mu4e-headers-query-prev' originally mapped on `\'
```

Moving back to the previous search result is used often, so painful that it gets lost. So probably worth mapping this to something more sensible than the `M-left` it now defaults to.  

btw) The key-bindings changed are rather extensive compared to the mu4e manual. It would be good to get the full list of the actual keybindings in the documentation. 
#### System Info
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.4
- Spacemacs branch: master (rev. da94a25)
- Distribution: spacemacs
- Layers:

``` elisp
(auto-completion emacs-lisp eyebrowse git markdown mu4e org
                 semantic spell-checking syntax-checking version-control)
```

JorisE:
Evil-mu4e is an external package to set up the keybindings for mu4e.

Related issue: #4490


StreakyCobra:
> but we may have to wait merging untill we are clear on the keybindings...

@JorisE You have to set the title to start with `[WIP]` then :-)

JorisE:
@StreakyCobra: That sounds like a smart convention :)

layers/+email/mu4e/packages.el
```diff
@@ -26,32 +27,10 @@
...
-
-      (evilified-state-evilify-map
-       mu4e-headers-mode-map
-       :mode mu4e-headers-mode
-       :bindings
-       (kbd "C-j") 'mu4e-headers-next
-       (kbd "C-k") 'mu4e-headers-prev
-       (kbd "J") (lambda ()
-                   (interactive)
-                   (mu4e-headers-mark-thread nil '(read))))
```
myrjola: I would be sad to see the `J` binding go from the mu4e-header-mode and mu4e-view-mode. It's my primary way of processing mail. Basically they jump to the next thread marking the previous thread as read.

JorisE: I have added the same functionality under J.

layers/+email/mu4e/packages.el
```diff
@@ -14,7 +14,7 @@
         (mu4e :location site)
         mu4e-alert
         mu4e-maildirs-extension
-        org
```
aroig: @JorisE I think you should not remove org here, otherwise org-mu4e will not load.

syl20bnr:
Thank you ! 👍 

I Cherry-picked into develop branch but unfortunately had to revert it as `evil-mu4e` is not yet ready to replace the auto-evilification.

The reason is that we must be able to hot change the editing style and key bindings must stay consistent, i.e. when I'm in Vim editing style then `evil-mu4e` must be activate but as soon as I switch to Emacs editing style with `SPC t E e` then `evil-mu4e` must undo its changes. This is an important feature to allow pair-programming even though for an email application it is less likely to happen...
see commit: https://github.com/syl20bnr/spacemacs/commit/f55b6254

Recently we added support for this behavior with `evil-magit` in commit https://github.com/syl20bnr/spacemacs/commit/4a34ef46 so it should give you some pointer on how to to the same with `evil-mu4e`,  of course your package must support to deactivate itself properly, you can check how it is done in `evil-magit`.

PR #5116, opened 2016-02-17
Improve toggle-maximize-centered-buffer

justbur:
1. The register name should be a character not a symbol.

2. Combine toggle-maximize-centered-buffer and bzg-big-fringe-mode to
   make new minor mode that combines the two effects.

3. Fix problem where the mode could only be enabled if there was more
   than one window, which makes no sense.

4. Use window-configuration-to-register function


TheBB:

PR #5121, opened 2016-02-18
markdown: Add binding for live preview mode

lunaryorn:
Recent versions of markdown provide markdown-live-preview-mode which shows a live preview in another window using Emacs' built-in EWW browser.  This pull request adds a binding to toggle this mode.  I'm not sure whether I should have implemented a proper spacemacs toggle for this so I went with the simpler route and bound the mode directly.  If you'd like to have a toggle please tell me.

Note that as it builds upon EWW this mode may not be available for older versions of Emacs.  As I'm using Emacs head I don't know however when EWW was introduced to Emacs, but I can find out if you'd like to add that to the documentation.

### DIFF
--- a/layers/+lang/markdown/README.org
+++ b/layers/+lang/markdown/README.org
@@ -112,18 +112,19 @@ To generate a table of contents type on top of the buffer:
 
 ** Buffer-wide commands
 
-| Key Binding | Description          |
-|-------------+----------------------|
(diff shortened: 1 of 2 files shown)


syl20bnr:
Eww was part of 24.5 if I recall correctly so it should be OK. thank you!

StreakyCobra:

PR #5124, opened 2016-02-18
cider: yet another missed occurrence of cider-test-run-tests

AlejandroCatalina:
Okay this was my bad as i didn't see this occurrence when I fixed the binding. This was not directly called but would result in another bug, so here's the fix.

If there is any other occurrence, please let me know. Should be hotfix?


syl20bnr:
Thank you ! :+1: 
Cherry-picked into develop branch and released in 0.105.10, you can safely delete your branch.

PR #5125, opened 2016-02-18
Add C-S-(JK) to helm-follow-action

CestDiego:
I noticed that one can go down and up with `C-j` and `C-k` but
helm-follow-action bindings are M-down or M-up, which is not EVIL at all
:P just putting it out there @syl20bnr this should not break anything


robbyoconnor:
I want this bad

TheBB:

PR #5126, opened 2016-02-18
Add RMarkdown support to ESS layer.

cpaulik:


izahn:
Why just RMarkdown? `polymode` supports Sweave, org-mode, and several other
literate programming systems. And it is not specific to R either, so
polymode should probably have its own layer.

I'm not against merging this as a temporary measure, but I do think it
needs some thought and a coherent way to integrate polymode into spacemacs
generally.

cpaulik:
I was just playing around with RMarkdown and saw that it was not supported. Since it seems more clumsy than using org mode directly I have no real need for it. So no strong opinion on this from my side.

roryk:
polymode doesn't work very well with evil-mode: https://github.com/vspinu/polymode/issues/41

izahn:
@roryk I can't reproduce the problems described in https://github.com/vspinu/polymode/issues/41. Can you? It is true that insert/normal state is not synchronized across indirect buffers, but I don't really see that as a problem, and certainly not something that should block integration in spacemacs.

roryk:
Yeah, I totally can, I enabled polymode again to edit a Rmarkdown file to check to see if it was fixed and disabled it a couple of days ago because it wasn't. It seems like it works fine at first but has problems. I'm not sure what is exactly how to reproduce the behavior, but you can get into a state where undo gets destroyed. It might be you have to get to a more complicated undo history like undoing in the two different blocks before it happens.

roryk:
I can't seem to reproduce it though, I'll keep it turned on for a while and see if I can figure out when and why it is happening.

The insert/normal state not being synced is pretty brutal. I can't even make named blocks:

```
```{r
````

traps me in a look up the function mode in spacemacs.

**Comment by izahn** [comment]
On Thu, Feb 25, 2016 at 1:40 PM, Rory Kirchner <notifications@github.com>
wrote:

> I can't seem to reproduce it though, I'll keep it turned on for a while
> and see if I can figure out when and why it is happening.
>
> The insert/normal state not being synced is pretty brutal. I can't even
> make named blocks:
>
> ```{r
>
> traps me in a look up the function mode in spacemacs.
>
​For me it just switches to ​normal state as soon as polymode recognizes
the delimiter, requiring an extra "i" to get back to insert state. Is that
what you are describing, or another problem? For me this is annoying but
not a deal breaker.

**Comment by roryk** [comment]
Yup, you are right.

**Comment by vspinu** [comment]
The issue has been fixed on the polymode side, at least to the extent I can test it as a non evil user.

**Comment by biocyberman** [comment]
Org-mode seems to be more stable than knitr; it supports many other languages; and managing document structure with it is very pleasing. So I tried to use org-mode to write org+R documents as a replacement for rmarkdown. However, here are the reasons I switched back to knitr (rmarkdown) for now:

- I could not find out how to highlight R code with colors.
- Knitr supporst better when I want to include plots from R code chunks, adding figure captions.

So yes, I still want to have support for Rmarkdown/knitr in `ess`. Implementing this pull request (enabling `polymode`) with some slight modifications, I find that `polymode` does not play well with evil-mode when:

- opening a code chunk. After  ````{r `, the whole line start jumping to the right on any keystroke.
- yanking and pasting. If the text content contains a whole code chunk, i.e. including the  start and the end ```, the paste is truncated to the first line of the yanked region.

Switching to emacs key (by looping with `Ctrl+z`) eliminates these problems. But you know that, vim key bindings rock. I can't live without `evil`.

**Comment by cpaulik** [comment]
For syntax highlighting you must have the `ess` layer installed and probably also do `  (add-to-list 'org-babel-load-languages '(R . t))` somewhere. I don't know about the figure captions but [this](http://archive.indianstatistics.org/tools/orgpapers.pdf) may help. What exactly was the problem with the figure captions?

**Comment by biocyberman** [comment]
@cpaulik Clarification: Syntax highlight while editing is working fine. I was taking about syntax highlighting for code chunks in rendered PDF or HTML documents.

**Comment by cpaulik** [comment]
@biocyberman Strange. HTML syntax highlighting should work fine out of the box AFAIK. For latex you have to use the `minted` package.

Something like 

```
      (setq org-latex-listings 'minted)
      ;; setup minted to have frame, small text and line numbers
      (setq org-latex-minted-options
                 '(("frame" "lines")
                   ("fontsize" "\\scriptsize")
                   ("linenos" "")))

      ;; setup of latex processing
      (setq org-latex-pdf-process '("latexmk -synctex=1 -shell-escape -pdflatex=xelatex -pdf %f"))
```
in your config should do if you use latexmk. Otherwise change the `org-latex-pdf-process` variable accordingly.

**Comment by izahn** [comment]
I can reproduce the `evil-yank` problem, I guess it should be reported upstream. I don't see anything like

> After ```{r, the whole line start jumping to the right on any keystroke.

@biocyberman can you give more detailed reproduction steps for this one?

**Comment by biocyberman** [comment]
@cpaulik and @izahn Don't want to hit and run, just because I don't work with R lately so I do not have input to reproduce this problem.

**Comment by Compro-Prasad** [comment]
@cpaulik @biocyberman Any updates on this PR?

**Comment by Compro-Prasad** [comment]
@duianto This seems a simple one. Can it be merged?
@cpaulik Please add an entry to `CHANGELOG.develop`.

**Comment by duianto** [comment]
I tried opening a `.rmd` example file:
https://raw.githubusercontent.com/tgirke/GEN242/gh-pages/_vignettes/07_Rbasics/sample.Rmd

But it opens in `fundamental mode` and the messages buffer shows:
>Loading c:/Users/username/.emacs.d/elpa/26.2/develop/julia-mode-20190407.2119/julia-latexsubs...done
File mode specification error: (error Autoloading file c:/Users/username/.emacs.d/elpa/26.2/develop/ess-20190616.1211/ess-site.elc failed to define function poly-markdown+r-mode)

I found this page:
https://www.reddit.com/r/spacemacs/comments/9ciefe/polymode_for_rmd/

Where the comment says:
>It looks like Polymode received a major update. You need to specify now the poly-modes you want to install (see here)
>
>Thus, for .Rmd file support it should be
```
polymode
poly-R
poly-noweb
poly-markdown
```

When I add the last three packages (`polymode` is added to the `ess` layer by this PR).

Now the same `.rmd` file opens in `markdown` major mode with `poly-markdown+r-mode` minor mode.

All three packages might not be needed, because the same thing happens when only `poly-R` is added.

#### Syntax highlighting issue

The author of the reddit thread seems to have had some syntax highlighting issues:
> If I open a .Rmd, some chunks have proper syntax highlighting while others don't, and adding more text/code results in somewhat sporadic behavior. Code highlights/unhighlights unexplicably.

And there's a small syntax highlighting issue for me as well. The highlighting doesn't appear until the cursor is moved. I even tried waiting 30 seconds but nothing happened until the cursor was moved, so it doesn't seem to be delayed.

When the highlighting does appear then it seems to be fine. Except for one place where it seems like there's an attempt at adding an indented code block within a code block:
```
    ```
	```\{r code_chunk_name, eval=FALSE\}
	x <- 1:10
	```
    ```
```
(everything in the code block is indented 4 spaces to make the backticks visible)
Then the highlighting marks everything until the start of the next code block as code. But this is unrelated to this PR, it seems to be a limitation (unsupported feature) in `markdown-mode`.

#### System Info :computer:
- OS: windows-nt
- Emacs: 26.2
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 4b20bb791)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(autohotkey colors emacs-lisp ess git helm markdown multiple-cursors org spell-checking syntax-checking treemacs version-control)
```

vspinu:
Issues with polymode are better reported on polymode tracker. 

> I found this page:

Polymode has been developed actively for the past months, so most of the issues people had in the past are either resolved or are no longer relevant. Setups and configurations in older posts/blogs might no longer be correct either. 

> All three packages might not be needed, because the same thing happens when only poly-R is added.

Only poly-R is needed. All others are dependencies of poly-R. Assignment into auto-mode-alist is not needed either, it's done automatically. 

> The highlighting doesn't appear until the cursor is moved.

This could come from the emacs re-display engine. In emacs 26 I have started seeing delays in re-display on various occasions in non-polymode buffer. I have never seen them in polymode buffers though. User input trigers the re-display so it's not a such a big deal. I haven't seen such issues with emacs 27 though. So maybe it has been solved already.

> Except for one place where it seems like there's an attempt at adding an indented code block within a code block:

Fenced code blocks cannot be nested. It doesn't make much sense in general. Indented blocks do work.

duianto:
>Issues with polymode are better reported on polymode tracker.

This seems to just be a setup issue in Spacemacs.

When I add `poly-R` to the `ess` layers `ess-packages` variable, and an entry for:
```elisp
(defun ess/init-poly-R ())
```

Then it works without having to manually add the package to the `dotspacemacs-additional-packages` variable in `.spacemacs`.

---

>Only poly-R is needed. All others are dependencies of poly-R.

Confirmed, it still works when these are commented out:
- `polymode` from the `ess-packages` variable
- and the `ess/init-polymode` function
```elisp
(defun ess/init-polymode ()
  (use-package poly-R
    :defer t)
  (use-package poly-markdown
    :defer t)
  )
```

---

>This could come from the emacs re-display engine. In emacs 26 I have started seeing delays in re-display on various occasions in non-polymode buffer. I have never seen them in polymode buffers though. User input trigers the re-display so it's not a such a big deal. I haven't seen such issues with emacs 27 though. So maybe it has been solved already.

Confirmed, it seems to be an Emacs 26 issue, syntax highlighting are present instantly when the `sample.Rmd` file opens with the latest snapshot:
GNU Emacs 27.0.50 (build 1, x86_64-w64-mingw32) of 2019-06-09

---

>Fenced code blocks cannot be nested. It doesn't make much sense in general. Indented blocks do work.

Then it's just an anomaly by the author of the sample file:
https://raw.githubusercontent.com/tgirke/GEN242/gh-pages/_vignettes/07_Rbasics/sample.Rmd

PR #5128, opened 2016-02-18
[Core] <SPC> f y working in dired buffers

dcluna:
Currently, the `spacemacs/show-and-copy-buffer-filename` does not work in dired buffers, this PR fixes that.


StreakyCobra:

PR #5137, opened 2016-02-19
Change helm-themes and counsel-load-theme keybinding to SPC T s.

person808:
The current keybinding is SPC T h which is different from the binding for counsel-load-theme. The new mnemonic for both ivy and helm would be`UI themes/color themes` which IMO makes more sense than the helm specific mnemonic of `UI themes/helm`.

### DIFF
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -429,7 +429,7 @@ To install it, just add =themes-megapack= to your =~/.spacemacs= like so:
 (setq-default dotspacemacs-configuration-layers '(themes-megapack))
 #+END_SRC
 
-You have now installed around 100 themes you are free to try with ~SPC T h~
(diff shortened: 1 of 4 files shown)


justbur:
`Tc` was actually for counsel to match the helm one, but I agree that it's not the right mnemonic. **T**hemes **c**olors is a little odd to me. How about **Themes** **s**elect?

sooheon:
I actually think `Th` for *Th*emes is a fine mnemonic, and as a bonus it's much easier to type than `Ts` or `Tc`.

StreakyCobra:
> it's much easier to type than Ts or Tc.

People always forget about the different existing keyboard-layouts :wink:

syl20bnr:
Thank you !
`SPC T s` was taken to toggle the semantic stickyfunction header, I moved it to `SPC T S`since it is something less used.

PR #5143, opened 2016-02-19
Enable OSX non-native fullscreen on regular builds

aaronjensen:
The mac port build supports this out of the box, but the version
installed with `brew install emacs --with-cocoa` requires this variable to be
cleared.

### DIFF
--- a/layers/+distribution/spacemacs-base/config.el
+++ b/layers/+distribution/spacemacs-base/config.el
@@ -244,6 +244,8 @@ or lists of these.")
     (if dotspacemacs-maximized-at-startup
         (add-hook 'window-setup-hook 'toggle-frame-maximized))))
 
+(setq ns-use-native-fullscreen (not dotspacemacs-fullscreen-use-non-native))
+
 ;; ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


d12frosted:
FYI, `--cocoa` is [deprecated](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/emacs.rb#L32), use `--with-cocoa`. See my

aaronjensen:
@d12frosted oh yeah, thanks, I updated the commit message.

TheBB:
Wouldn't it make sense to set it inversely too?

```elisp
(setq ns-use-native-fullscreen (not dotspacemacs-fullscreen-use-non-native))
```

aaronjensen:
@TheBB not sure, I think it defaults to false and I wanted to not do anything that is unnecessary. I'm happy to change it if you think we should.

aaronjensen:
@TheBB I updated it, thanks.

aaronjensen:
@syl20bnr This one should be good to go, need anything else?

syl20bnr:
Thank you !

PR #5144, opened 2016-02-19
Add `C-[hjkl]` bindings in deft

StreakyCobra:


layers/deft/packages.el
```diff
@@ -37,4 +37,9 @@
         "d" 'deft-delete-file
         "i" 'deft-toggle-incremental-search
         "n" 'deft-new-file
-        "r" 'deft-rename-file))))
+        "r" 'deft-rename-file)
+      (evil-define-key 'insert deft-mode-map
+        (kbd "C-h") 'deft-filter-clear
+        (kbd "C-j") 'evil-next-line
```
StreakyCobra: Hum, I just thought that this will not work for non-evil users :-( Not event sure `C-k` should be remapped as discussed in https://github.com/syl20bnr/spacemacs/issues/5237.

Compro-Prasad:
@StreakyCobra Is the hold over? It is about to be 2 years on hold.

PR #5150, opened 2016-02-19
Remove references to Solarized themes

dumbl3d0re:
The documentation and configuration files shouldn't assume that the user have installed themes that isn't part of the default distribution in their general documentation.

### DIFF
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -98,8 +98,6 @@ of a list then all discovered layers will be installed.")
 
 (defvar dotspacemacs-themes '(spacemacs-dark
                               spacemacs-light
-                              solarized-dark
-                              solarized-light
                               leuven)
   "List of themes, the first of the list is loaded when spacemacs starts.
 Press <SPC> T n to cycle to the next theme in the list (works great
--- a/core/templates/.spacemacs.template
+++ b/core/templates/.spacemacs.template
@@ -101,8 +101,6 @@ values."
    ;; with 2 themes variants, one dark and one light)
    dotspacemacs-themes '(spacemacs-dark
                          spacemacs-light
-                         solarized-light
-                         solarized-dark
                          leuven
                          monokai
                          zenburn)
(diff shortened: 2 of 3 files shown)


cpaulik:
These themes will automatically be installed if loaded through `SPC T n`

Otterhall:
But if they aren't installed by default, they shouldn't be in the configuration and documentation files.

Frankly I don't believe they are necessary because the spacemacs dark/light themes has a similar colorscheme, and installed by default. But if the user want to use them specifically, they can install it without much trouble by installing the `theme-megapack`.

We shouln't give examples in the documentation that **doesn't work** by default. Better to use a theme that's installed by default, and is a official spacemacs theme.

StreakyCobra:
I am in favor to only define `spacemacs-[dark|light]` in `dotspaecmacs-themes`, both in `core/core-dotspacemacs.el` and `core/templates/.spacemacs.template`. Removing all the other.

I don't agree with the change in `doc/DOCUMENTATION.org` though. It's an example to show how to define and use other themes. `solarized` is a good example for this.

Otterhall:
> I don't agree with the change in doc/DOCUMENTATION.org though. It's an example to show how to define and use other themes. solarized is a good example for this. @StreakyCobra 

As I understand it the the changes is `doc/DOCUMENTATION.org` is about the same variable (`dotspacemacs-themes`) that I've changed in `core/templates/.spacemacs.template`. I.e. in `doc/DOCUMENTATION.org` gives an example of how to change the default theme at startup from `spacemacs-dark` to another theme.

But correct me if I'm wrong.

StreakyCobra:
@Otterhall Yes, it's about the same variable, so it's better to show that this can also be used with other themes, like solarized.

In other words, I don't see the problem of having `solarized` used as an example here?

nixmaniack:
Since `solarized-dark` doesn't work out of the box in terminal, it's not the "best" example from UX PoV.

StreakyCobra:
Ok, this could be a good reason :-)

syl20bnr:
I agree with this.
I even kept only `spacemacs-dark` and `spacemacs-light` in the default value for `dotspacemacs-themes`. But kept in documentation there are leuven, monokai and zenburn mentioned.
Thank you (and thanks again for your patience) !

PR #5161, opened 2016-02-20
[WIP] change OS X installation instructions

d12frosted:
So not it's recommended to install GNU emacs instead of emacs-mac as
discussed in #4955.

---

Not sure what to do with [these notes](https://github.com/syl20bnr/spacemacs/blob/develop/doc/FAQ.org#why-are-the-powerline-colors-not-correct-on-os-x) in FAQ. 

P. S. I will remove WIP once we decide what to do with mentioned problem and after `--with-spacemacs-icon` become available for `emacs` brew formulae. 

P. P. S. Even if you try to install using `--with-spacemacs-icon` option, emacs will be installed, just without spacemacs icon. Already sent PR to emacs formula. Details in https://github.com/Homebrew/homebrew/pull/49371.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,11 +131,10 @@ this version. The same may be true for other distributions as well.
 
 ### OS X
 
-We recommend the homebrew [emacs-mac-port][] formula:
+The recommended way of installing Emacs on OS X is using [homebrew][]:
 
 ```sh
-$ brew tap railwaycat/homebrew-emacsmacport
-$ brew install emacs-mac --with-spacemacs-icon  # OR, brew cask install emacs-mac
+$ brew install emacs --with-cocoa --with-spacemacs-icon --with-gnutls --with-rsvg --with-imagemagick
 $ brew linkapps
 ```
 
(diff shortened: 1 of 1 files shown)


syl20bnr:
We can replace the advice to use emacs-mac-port by the solution to set the variable to nil and explain that it modifies the color rendering of the whole Emacs. Also some users recommend to change the powerline separator to something else not subject to the graphical glitches.

d12frosted:
@syl20bnr as it turned out, brew maintainers are against of adding option for spacemacs icon. So we have an option to create a separate tap for emacs with spacemacs icon. If you wish I can create one.

syl20bnr:
Thank you !

I edited the PR to remove the --spacemacs-icon parameter. I'm not sure we should create a tap, I want people to explicitly install Emacs and then add Spacemacs on top, I don't want to "hide" Emacs and get people confused about what it Spacemacs.
So we could add the steps in the FAQ to add the Spacemacs icon on OS X.

d12frosted:
> I want people to explicitly install Emacs and then add Spacemacs on top, I don't want to "hide" Emacs and get people confused about what is Spacemacs.

Good point. People are already confused a little bit.

PR #5164, opened 2016-02-20
Exclude untracked files when checking dirty repo

glittershark:
Exclude untracked files (porcelain lines of the form "?? FILENAME") when
checking for a dirty repo during updates, since these won't be squashed
by any git commands


glittershark:
oops, master

PR #5168, opened 2016-02-21
Fixes #5170 - Set default values only if not already customized

### LINKED ISSUE #5170: Org layer overrides user configuration set in org-customize
Fixed in #5168

opie4624:
Thank you for your pull request. In order for your pull request to be accepted please check the following.
- [X] I have read the Contribution Guidelines at https://github.com/syl20bnr/spacemacs/tree/develop/CONTRIBUTING.org
- [X] My pull request is against the `develop` branch and not `master`
- [X] I have rebased my contribution against `develop` and not `master` – Changes were made while on the `develop` branch
- [X] I have written a descriptive commit message that tells people what I did
- [X] My pull request has been rebased and squashed into one commit

Failure to check these will result in your pull request being rejected. You may describe your pull request below.

### DIFF
--- a/core/core-spacemacs.el
+++ b/core/core-spacemacs.el
@@ -304,4 +304,10 @@ with the message pre-filled with template and information."
     (setq url (replace-regexp-in-string "#" "%23" url))
     (browse-url url)))
 
+(defun spacemacs/check-for-customized (arg)
+  "Non-nil if a value has been set via customize"
+  (or (get arg 'customized-value)
+      (get arg 'customized-face)
+      (get arg 'saved-value)))
+
 (provide 'core-spacemacs)
--- a/layers/org/packages.el
+++ b/layers/org/packages.el
@@ -267,7 +267,8 @@ Will work on both org-mode and any mode that accepts plain html."
         "aol" 'org-store-link))
     :config
     (progn
-      (setq org-default-notes-file "notes.org")
+      (unless (spacemacs/check-for-customized 'org-default-notes-file)
+          (setq org-default-notes-file "notes.org"))
       (font-lock-add-keywords
        'org-mode '(("\\(@@html:<kbd>@@\\) \\(.*\\) \\(@@html:</kbd>@@\\)"
                     (1 font-lock-comment-face prepend)


robbyoconnor:
@opie4624 thanks <3

opie4624:
I think this could stand to see a bit more polish over time. There are lots of defaults that layers override that could stand to have a better "standardized" way to set them that would still honor user's settings via Customize.

Relevant discussion in chat: https://gitter.im/syl20bnr/spacemacs?at=56b391733bd55a660df5688c

syl20bnr:
Sorry but what's the issue you want to fix ? Can you open an issue to describe the problem first ? A PR is not the right place to do this.

reactormonk:
I've just recently had the issue of scala-mode overriding my customize settings. The correct check is according to #emacs `custom-variable-p` and I'd go with a macro to replace the setq call. I'll take a look.

TheBB:
I think this, or something like it, is what we need to use.

`custom-variable-p` is unreliable because it doesn't seem to know whether a
variable is customizable until its definition has been loaded, and it also
returns true regardless of whether it actually has a custom value or not.

```elisp
(defmacro spacemacs|setq (&rest args)
  (let (result)
    (while args
      (let ((variable (pop args))
            (form (pop args)))
        (push `(or (not (get ',variable 'customized-value))
                   (not (get ',variable 'saved-value))
                   (setq ,variable ,form))
              result)))
    `(progn ,@(reverse result))))
```

opie4624:
When I was researching this for my org-mode issues, I found a comment buried on the customize page of emacswiki that stated `customized-value` is set if it has been customized and loaded from .emacs while `saved-value` is set if the value has been customized within the interactive customize interface during this session. So @TheBB's code is checking the right things.

Also `customized-face` is set for customized fonts. Again, `saved-value` is used when fonts are customized interactively in this session. In the code I submitted for the PR, I checked all 3 in the same place since it seemed a nice, clean way to check for customization anywhere in once place.

TheBB:
Yes, I skipped the face stuff because you wouldn't use a `setq` macro for that. And I think going with a macro is the right choice here so that you wouldn't have to replicate the checking logic everywhere.

syl20bnr:
Thank you for the PR, I pushed a commit which delays the application of custom settings, they are now applied after `user-config`.

See commit: https://github.com/syl20bnr/spacemacs/commit/e699f18

PR #5169, opened 2016-02-21
org: fix layout when org-agenda-files isn't a list

jmiven:
The org custom layout is currently broken when the org-agenda-files variable is the name of a directory or a file.


jmiven:
I've added the autoload for the org-agenda-files command.

TheBB:

PR #5184, opened 2016-02-22
add 'cider-apropos functionalty to clojure layer

birdspider:
I'm unsure if this is important enough for inclusion or falls under "user key-bindings".


AlejandroCatalina:
It is really nice to have this, IMHO. :+1:

TheBB:

Issue #5188, opened 2016-02-22
German Keyboard on OSX: Symbols behind ⌥ - key are unreachable


marcopaga:
#### Description
I want to enter '[' or ' ]' in OSX with a german keyboard. ⌥5 / ⌥6 doesn’t work. This will switch the window due to the `windows-numbering` package.

#### Reproduction guide
- Start Emacs
- Enter ⌥5 with a german keyboard

*Observed behaviour:*
- Will try to switch to window 5

*Expected behaviour:*
- Expected is [

#### System Info
- OS: darwin
- Emacs: 24.5.1
- Spacemacs: 0.105.11
- Spacemacs branch: master (rev. 93e55bb)
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(emacs-lisp)
```

#### Backtrace
```
BACKTRACE IF RELEVANT
```

marcopaga:
When deactivating the window-numbering package it doesn't work, too.

E.g. This config doesn't work: https://gist.github.com/marcopaga/c6b8f99f825d577c169a

sooheon:
I think your problem is that the `osx-use-option-as-meta` doesn't work as intended. With that var set as nil, the value of `mac-option-modifier` is still `meta`. Try setting that explicitly to `nil`. You can set left and right option modifiers separately, so one side does the system option key, and the other side sends meta.

tko:
FWIW for me `mac-option-modifier` has default value of `(:function alt :mouse alt)` and I can type special characters just fine (Finnish keyboard layout.) `window-numbering` also seems to work though I get to use `ESC` since I don't have meta.

Maybe the default value depends on which emacs build one uses? I have emacs-24.5-z-mac-5.15 from the emacs-mac brew cask.

d12frosted:
I also relay on `OPTION-` symbols, so I have set in `dotspacemacs/user-config` function:

```
(if (spacemacs/system-is-mac)
      (setq mac-command-modifier 'meta
            mac-option-modifier  'none))
```

So CMD acts as `M` (meta) and Option is free for special symbols.

d12frosted:
And another update.

In linked `.spacemacs` file there is `osx` layer enabled, but I don't see it in `SPC h d s` output. Try to restart emacs.

marcopaga:
You are right. Needed to restart. This seems to be fine.

#### System Info
- OS: darwin
- Emacs: 24.5.1
[...]

PR #5189, opened 2016-02-22
Tip regarding option on international keyboards

marcopaga:
This is a follow up for: https://github.com/syl20bnr/spacemacs/issues/5188

More config information regarding option on international keyboards.

### DIFF
--- a/layers/osx/README.org
+++ b/layers/osx/README.org
@@ -47,6 +47,12 @@ for e.g. Finnish and Swedish keyboard layouts, you'll probably want to leave the
      (osx :variables osx-use-option-as-meta nil)))
 #+END_SRC
 
+If you have problem entering symbols that are behind the ~⌥~ key you may want to added this to the user-init in the .spacemacs-File. This will allow you to use the right ~⌥~ key to write symbols. The left ~⌥~ key can be used as the Meta key. 
+
+#+BEGIN_SRC emacs-lisp
+  (setq-default mac-right-option-modifier nil)
+#+END_SRC
+
 ** Coreutils
 
 To get =gls= install coreutils homebrew:


syl20bnr:
:+1:

d12frosted:
Another good solution is using `cmd` as `meta`:

```
(setq mac-command-modifier 'meta
      mac-option-modifier  'none)
```

The only caveat is when you use GNU Emacs (25?) and want to `cmd-h`. You could think that `(bind-key "M-h" 'ns-do-hide-emacs)` is enough to solve this problem, but some modes define `M-h`. That's why I use following function:

```
(defun d12/setup-M-h ()
  "Setup M-h key binding on OS X in GUI."
  (when (and (spacemacs/system-is-mac)
             (display-graphic-p))
    (bind-key "M-h" 'ns-do-hide-emacs)
    (-map (lambda (mode)
            (add-hook (intern (concat (symbol-name mode) "-hook"))
                      `(lambda ()
                         (define-key
                           (symbol-value (intern ,(concat (symbol-name mode) "-map")))
                           (kbd "M-h")
                           nil))))
          '(org-mode
            company-quickhelp-mode))))
```

This is far from ideal, but I believe that some people could find it useful.

TheBB:

We break long lines around these here parts so I went ahead and did that for you in https://github.com/syl20bnr/spacemacs/commit/9ccad9c2c1da21301fffc126a5beaf785de9a0d7. Also formatted the rest of the file to match Spacemacs conventions.

Akii:
As an additional note: You need the emacs-mac-port, not the default emacs installed by homebrew for this setting to work.

To check if the setting is available:

```
SPC h d v mac-right
```

should output

```
mac-right-option-modifier is a variable defined in ‘C source code’.
Its value is nil
Original value was left

Documentation:
Modifier key assumed when the Mac right option key is pressed.
The value is of the form either SYMBOL or ‘(:ordinary SYMBOL :function
SYMBOL :mouse SYMBOL)’.  The latter allows us to specify different
behaviors among ordinary keys, function keys, and mouse operations.
```

PR #5196, opened 2016-02-23
Disable evil-snipe in ranger

### LINKED ISSUE #5029: ranger-search-files does not work

TheBB:
Fix #5029, possibly supersedes #5030.

CC @StreakyCobra @ralesi @psamim

### DIFF
--- a/layers/+vim/evil-snipe/packages.el
+++ b/layers/+vim/evil-snipe/packages.el
@@ -2,6 +2,7 @@
       '(
         evil-snipe
         magit
+        ranger
         ))
 
 (defun evil-snipe/init-evil-snipe ()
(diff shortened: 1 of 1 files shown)


StreakyCobra:
I prefer this over #5030

psamim:
It is a better solution.

ralesi:
This works,  but is there a way we can just set up either:
- an alist of modes in evilsnipe to not apply the hook
- bind the evilsnipe commands directly to evil-motion-state-map so they don't interfere with keys applied on evil. 

 I believe this is going to come up for any mode that is using evil-motion-state

TheBB:
I merged it as an initial solution. We can write something more elegant later.

PR #5209, opened 2016-02-24
fix spacemacs/swiper-region-or-symbol

d12frosted:
`swiper--ivy` takes candidates list and initial input. That's why in current implementation `spacemacs/swiper-region-or-symbol` is not working as expected. Looks like `swiper` is what it should be.


TheBB:

PR #5210, opened 2016-02-24
[RDY]add annotation when exporting instead of embedding.

JAremko:
Add function `spacemacs//add-org-meta-readtheorg-css` that wraps around `org-html-publish-to-html`. The function inserts `readtheorg.css` link with a proper path if the document to be published doesn't have it.

It removes the need to clutter Spacemacs documentation with `#+HTML_HEAD_EXTRA: <link rel="stylesheet" type="text/css" href="../css/readtheorg.css" />` in each org file.  Also it reduces amount of bugs, because it's hard to get the path right.

Also I removed those links from the org files.

### DIFF
--- a/core/core-documentation.el
+++ b/core/core-documentation.el
@@ -13,6 +13,7 @@
 (require 'ox-publish)
 (require 's)
 (require 'dash)
+(require 'f)
 
 (defvar spacemacs--category-names
   '(("config-files" . "Configuration files")
@@ -85,11 +86,41 @@
         (format "%s\n%s%s" beginning-of-content toc-string rest-of-content)
       content)))
 
+(defun spacemacs//add-org-meta-readtheorg-css (origfunc &rest args)
+  (save-current-buffer
+    (let* ((head-css-extra-readtheorg-head (concat
+                                            "#+HTML_HEAD_EXTRA:"
+                                            "<link rel=\"stylesheet\" "
+                                            "type=\"text/css\" "
+                                            "href=\""))
+           (head-css-extra-readtheorg-tail "css/readtheorg.css\" />\n")
+           (filename (car (nthcdr 1 args)))
+           (visitingp (find-buffer-visiting filename)))
+      (with-temp-buffer
+        (when visitingp (with-current-buffer visitingp (setq buffer-file-name nil)))
+        (insert-file-contents filename t)
+        (save-match-data
+          (if (or (re-search-forward "\+HTML_HEAD_EXTRA\:.*\/css\/readtheorg\.css" nil t nil)
+                  (string= filename "LAYERS.org"))
+              (apply origfunc args)
+            (progn (goto-char (point-min))
+                            (if (search-forward "#+TITLE:" nil t nil)
+                                (beginning-of-line 2)
+                              (error "Can't find #+TITLE:"))
+                            (insert (concat head-css-extra-readtheorg-head
+                                            (f-relative user-emacs-directory
+                                                        (file-name-directory filename))
+                                            head-css-extra-readtheorg-tail))
+                            (apply origfunc args)
+                            (not-modified)))))
+        (when visitingp (with-current-buffer visitingp (setq buffer-file-name filename))))))
+
 (defun spacemacs/publish-doc ()
   "Publishe the documentation to doc/export/."
   (interactive)
   (advice-add 'org-html-toc :filter-return #'spacemacs//format-toc)
   (advice-add 'org-html-template :filter-return #'spacemacs//format-content)
+  (advice-add 'org-html-publish-to-html :around #'spacemacs//add-org-meta-readtheorg-css)
   (let* ((header
           "<link rel=\"stylesheet\" type=\"text/css\"
                  href=\"http://www.pirilampo.org/styles/readtheorg/css/htmlize.css\"/>
@@ -137,6 +168,7 @@
              :publishing-function org-publish-attachment))))
     (org-publish-project "spacemacs"))
   (advice-remove 'org-html-toc #'spacemacs//format-toc)
-  (advice-remove 'org-html-template #'spacemacs//format-content))
+  (advice-remove 'org-html-template #'spacemacs//format-content)
+  (advice-remove 'org-html-publish-to-html #'spacemacs//add-org-meta-readtheorg-css))
 
 (provide 'core-documentation)
--- a/doc/CONVENTIONS.org
+++ b/doc/CONVENTIONS.org
@@ -1,5 +1,4 @@
 #+TITLE: Spacemacs Conventions
-#+HTML_HEAD_EXTRA: <link rel="stylesheet" type="text/css" href="../css/readtheorg.css" />
 
 * Spacemacs conventions                                     :TOC_4_org:noexport:
  - [[Code guidelines][Code guidelines]]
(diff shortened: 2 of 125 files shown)


layers/org/packages.el
```diff
@@ -135,6 +136,24 @@
...
+              (with-current-buffer tmpbf
+                (progn
+                  (when (string= (buffer-substring (point-min) 3) "* ")
+                    (delete-region (point-min) 3))
+                  (goto-char (point-min))
+                  (insert "#+TITLE: ")
+                  (beginning-of-line 2)
+                  (insert spacemacs--org-export-css-annotation)
+                  (org-html-export-to-html)))
+              (kill-buffer tmpbf)))))
```
JAremko: I think it might be better to simply use `save-excursion` and `undo`  Instead of creating a new buffer +  this way org export thingy won't prompt for the file name. Or work with a buffer attached to a file and then restore it from the tmpbf.

TheBB:
This is a good start, thank you. Can you put the export function in `core/core-documentation.el` and make sure that `spacemacs/publish-doc` uses it? This isn't a function that the end user will need to use so let's keep it in that file.

I don't agree with substituting a first-level headline for the title though. Now the title is on the same level as the table of contents and the other headlines.

layers/org/packages.el
```diff
@@ -135,6 +136,24 @@
...
+            (progn
+              (copy-to-buffer tmpbf (point-min) (point-max))
+              (with-current-buffer tmpbf
+                (progn
+                  (when (string= (buffer-substring (point-min) 3) "* ")
+                    (delete-region (point-min) 3))
+                  (goto-char (point-min))
+                  (insert "#+TITLE: ")
+                  (beginning-of-line 2)
+                  (insert spacemacs--org-export-css-annotation)
```
TheBB: This has to be a function that computes how many `../` to add depending on the path of the file. They're not equal!

layers/org/packages.el
TheBB: I believe `org-publish` needs to have access to the filename, therefore we can't use temporary buffers. Otherwise how would it know to export `layers/auto-completion/README.org` to `doc/layers/auto-completion/README.html`?

layers/org/packages.el
```diff
@@ -155,6 +174,7 @@ Will work on both org-mode and any mode that accepts plain html."
         "e" 'org-export-dispatch
         "f" 'org-set-effort
         "P" 'org-set-property
+        "S" 'spacemacs/org-html-export-annotated
```
TheBB: We don't need this. It's not an end user function.

layers/org/packages.el
JAremko: @TheBB  Hm... You right!  Well, it won't be too hard.

syl20bnr:
Nice :-)

JAremko:
If it gonna work from the first time I will kiss myself :smile:

JAremko:
Aaaand I actually misunderstood how URLs are structured.

JAremko:
Who knows why [this](https://github.com/syl20bnr/spacemacs/blame/ef5a2ad5ee6362588244374364fe4ac2df066f37/layers/speed-reading/README.org#L2) layer has `../css/readtheorg.css` but [that](https://github.com/syl20bnr/spacemacs/blame/723a7921d7ee2ed8d61aa3e7cc09b0b24949c3bb/layers/smex/README.org#L2)  has `../../css/readtheorg.css` ? They are at the same depth :confused:

core/core-documentation.el
```diff
@@ -111,7 +147,7 @@
              :base-directory ,spacemacs-docs-directory
              :base-extension "org"
              :publishing-directory ,(concat publish-target "doc/")
-             :publishing-function org-html-publish-to-html
+             :publishing-function spacemacs//spacemacs-doc-publishing-function
```
JAremko: I think I can use `advice` instead. But I have to use `org-html-publish-to-htm` in two different places with the two different advices. 

This way the code is clearer.

core/core-documentation.el
```diff
@@ -85,6 +85,42 @@
...
+                                        "type=\"text/css\""
+                                        "href=\"../../../css/readtheorg.css\" />\n")))
+        (unless (search-forward head-css-extra-readtheorg nil t nil)
+          (progn
+            (goto-char (point-min))
+            (when (search-forward "#+TITLE:" nil t nil)
+              (beginning-of-line 2))
+            (insert head-css-extra-readtheorg)))))
+
+    (org-html-publish-to-html plist filename pub-dir)))
```
JAremko: Probably should remove the added lines when finished.

JAremko:
Ok we almost there.  :smile:

JAremko:
Look close enough to be good enough. 
It's, probably, a good time for someone to review this. Before I gone too far  :smirk:

JAremko:
Looks like it's ready :shipit:

TheBB:
Thanks, I'll have a look when I get time.

core/core-documentation.el
```diff
@@ -85,6 +85,52 @@
...
+  (progn
+    (find-file filename)
+    (unless (re-search-forward "\+HTML_HEAD_EXTRA\:.*\/css\/readtheorg.css" nil t nil)
+      (let ((head-css-extra-readtheorg (concat
+                                        "#+HTML_HEAD_EXTRA:"
+                                        "<link rel=\"stylesheet\" "
+                                        "type=\"text/css\" "
+                                        "href=\"../css/readtheorg.css\" />\n")))
+        (progn
+          (setq buffer-read-only nil)
```
JAremko: As an alternative we can set `inhibit-read-only` before starting the publishing process and turn it off afterwards.

JAremko:
Refactoring time!  :smiley:

fbergroth:
You can clean up the path shenanigans with some `f.el`, such as `f-relative`. :)

JAremko:
@fbergroth  thix I'll check it out. 
Now I have bigger problem.  Apparently, if an org file is already visited by some buffer, then `org-html-publish-to-html` will use it instead of my temp buffer. So I has to use some dirty hack  :confused:

JAremko:
Or I can use `find-file` instead, but then I'll have to inhibit `read-only` and `undo` my change. Well it's not too bad.

JAremko:
@TheBB  [RDY] now for realz! :relieved:

core/core-documentation.el
```diff
@@ -85,11 +86,41 @@
...
+        (when visitingp (with-current-buffer visitingp (setq buffer-file-name nil)))
+        (insert-file-contents filename t)
+        (save-match-data
+          (if (or (re-search-forward "\+HTML_HEAD_EXTRA\:.*\/css\/readtheorg\.css" nil t nil)
+                  (string= filename "LAYERS.org"))
+              (apply origfunc args)
+            (progn (goto-char (point-min))
+                            (if (search-forward "#+TITLE:" nil t nil)
+                                (beginning-of-line 2)
+                              (error "Can't fine #+TITLE:"))
```
JAremko: `Can't fine` :facepalm:

core/core-documentation.el
```diff
@@ -85,11 +86,41 @@
...
+                            (insert (concat head-css-extra-readtheorg-head
+                                            (f-relative user-emacs-directory
+                                                        (file-name-directory filename))
+                                            head-css-extra-readtheorg-tail))
+                            (apply origfunc args)
+                            (not-modified)))))
+        (when visitingp (with-current-buffer visitingp (setq buffer-file-name filename))))))
+
 (defun spacemacs/publish-doc ()
   "Publishe the documentation to doc/export/."
```
JAremko: typo.

JAremko:
@TheBB This PR overlaps with my TOC link fix thingy. I'll compose big PR with separate commit to keep the whole thing conflict free.

Also I found typo in the original file and removed  unnecessary stuff from this commit.

Issue #5200, opened 2016-02-23
Update annoyances


justbur:
This is a list of things I find annoying about "Update Packages". Some might be necessary, but I suspect that they're not. 

 1. Having to restart emacs: Is this really necessary? What problem does it solve? As far as I know updating through `list-packages` works fine without a restart, so why does spacemacs need a restart?

 2. No knowledge of dependencies: Every time `magit` updates with `with-editor` I have to update twice (2! restarts if you follow the spacemacs update procedure), because `with-editor` is deleted, which `magit` needs so `magit` fails to be installed the first time, requiring another restart.

 3. It's getting better but it is still difficult to keep track of what packages were updated and what their versions were. A simple  log would be really nice here (Ref #3798)

TheBB:
Another thing I would like (because I just experienced a need for it) is to be able to hold packages back from being updated.

syl20bnr:
1- This is necessary because of bugs we encountered when not doing so. The current process of updating package has been built empirically given the numerous bugs we got in the first end of 2-14 and beginning half of 2015. I never had any issue since then I use the `Udpate packages` button of Spacemacs.

2- Dependencies are handled by `package.el`.

3- We can much improve this indeed.

>  hold packages back from being updated

This is a planned feature in my todo list (I may know why you needed it :-D).

syl20bnr:
For the magit issue, it may be due to it being too smart by asynchronously updating itself to avoid the bugs we cannot encounter in Spacemacs. Let me find the thread about it. May or may not be related.

justbur:
> 2- Dependencies are handled by package.el.

Yes, which is why it refuses to install magit the first time.

 1. Update Packages
 2. Updates for with-editor and magit detected so these are deleted (well moved to rollback)
 3. Restart
 4. Spacemacs appears to process the to be installed packages in alphabetical order
 5. magit is scheduled to be installed first, but with-editor is not available so the install fails.
 6. with-editor installs fine
 7. Restart
 8. magit can now install because with-editor is available. 

5 seems completely unnecessary and avoidable. You could check that with-editor is about to be installed

syl20bnr:
See: https://github.com/magit/magit/pull/2156

justbur:
No, this is not the same issue. This is a spacemacs issue with the order in which packages are processed.

syl20bnr:
> magit is scheduled to be installed first, but with-editor is not available so the install fails.

[...]

PR #5217, opened 2016-02-24
Use unless-debug version of condition-case

justbur:
It's not possible to get full backtraces in these instances when an
error is signaled, and I think we should generate full backtraces when
debug-on-error is enabled as a general rule.

I had to make this change at one place to track down the root of problem
2 in #5200.

### DIFF
--- a/layers/+distribution/spacemacs-base/packages.el
+++ b/layers/+distribution/spacemacs-base/packages.el
@@ -300,7 +300,7 @@ Example: (evil-map visual \"<\" \"<gv\")"
         i.e. `SPC m h h`, to lookup the source of the definition,
         while falling back to `evil-lookup'"
         (interactive)
-        (condition-case nil
+        (condition-case-unless-debug nil
             (execute-kbd-macro (kbd (concat dotspacemacs-leader-key " mhh")))
           (error (evil-lookup))))
       (define-key evil-normal-state-map (kbd "K") 'spacemacs/evil-smart-doc-lookup)
@@ -311,7 +311,7 @@ Example: (evil-map visual \"<\" \"<gv\")"
         i.e. `SPC m g g`, to lookup the source of the definition,
         while falling back to `evil-goto-definition'"
         (interactive)
-        (condition-case nil
+        (condition-case-unless-debug nil
             (execute-kbd-macro (kbd (concat dotspacemacs-leader-key " mgg")))
           (error (evil-goto-definition))))
       (define-key evil-normal-state-map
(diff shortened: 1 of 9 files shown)


StreakyCobra:
:+1:

layers/+distribution/spacemacs-base/packages.el
```diff
@@ -300,7 +300,7 @@ Example: (evil-map visual \"<\" \"<gv\")"
         i.e. `SPC m h h`, to lookup the source of the definition,
         while falling back to `evil-lookup'"
         (interactive)
-        (condition-case nil
+        (condition-case-unless-debug nil
```
TheBB: I'm not sure if this is needed. An error here is expected in many modes, no?

layers/+distribution/spacemacs-base/packages.el
```diff
@@ -311,7 +311,7 @@ Example: (evil-map visual \"<\" \"<gv\")"
         i.e. `SPC m g g`, to lookup the source of the definition,
         while falling back to `evil-goto-definition'"
         (interactive)
-        (condition-case nil
+        (condition-case-unless-debug nil
```
TheBB: Same?

layers/+distribution/spacemacs-base/packages.el
justbur: My perspective is that you enable debugging because you want a backtrace and I'd just assume always generate one. 

If you're sure that someone would never want the backtrace here I'm ok removing it.

justbur: Actually this is just a poorly written function. It should be checking if the key binding exists directly rather than relying on some key sequence throwing an error.

TheBB: I agree :+1:

> My perspective is that you enable debugging because you want a backtrace and I'd just assume always generate one.

Coming from a Python perspective, errors are frequently used as part of an interface. That seems to be the case here, even though as you say it should have been written differently to begin with.

IOW, *errors* are not always *exceptions*.

justbur: Even if this function is written to catch an exception it shouldn't catch
everything just the one it anticipates.

TheBB: I know. I said I agree.

justbur: I realized that afterwards :)

TheBB:

Had to tweak a few things: https://github.com/syl20bnr/spacemacs/commit/65702c9285f6f7636d08f5a03d7b7bec9e1e849d

PR #4989, opened 2016-02-07
[RIP] hide org meta when using `SPC` `h` `SPC`

JAremko:
I reworked #4981 from scratch.

PROS compered to the old PR:
  - Doesn't requires Emacs version "24.4".
  - Intelligently hides lines instead of exact match.
  - You still can select hidden lines if you want.
  - Less code smell = happy :nose:  (it's a nose).
  - Can be disabled.
  - Doesn't do stuff if there is nothing to do.  

CONS:
  - A bit more complicated.
  - Proper global state management might require extra hookery.  (So far looks good)

*I found it rather neat to have. So I removed all the :hankey: added some more and created this brand new PR.*

PR #5222, opened 2016-02-24
[RDY]hide org-mode meta tags in the view-mode

JAremko:
Add function  `spacemacs//toggle-transparent-org-meta` that decorates `view-mode`. The function visually hides  and unhides org meta tags when `view-mode` is toggled. It makes documentation clearer and easier to read.

This is my third and final attempt to make it right. The code smells like :rose: 

The wall of shame:
#4989
#4981


JAremko:
I formatted it a bit. Looks good.   :shipit:

JAremko:
I'll use it as a base for my other commit -they're overlapping.

PR #5223, opened 2016-02-24
Fix Window Manipulation Transient State typo

darkfeline:
Fix Window Manipulation Transient State typo


TheBB:

PR #5226, opened 2016-02-25
fix for git-blame transient state

catesandrew:
looks like a typo, use transient state vs micro-state for magit-blame


nixmaniack:
It was intentional as far as I remember. https://github.com/syl20bnr/spacemacs/commit/de44ab1

catesandrew:
closing per [this revert](https://github.com/syl20bnr/spacemacs/commit/de44ab1) as nixmaniack mentioned above.

PR #5228, opened 2016-02-25
fix typo for js2-imenu extras mode

catesandrew:
previously the shortcut was `<spc> sl` but now its closer to either `<spc> sj` or `<spc> ij`

### DIFF
--- a/layers/+lang/javascript/packages.el
+++ b/layers/+lang/javascript/packages.el
@@ -84,7 +84,7 @@
     :init
     (progn
       (add-to-list 'auto-mode-alist '("\\.js\\'" . js2-mode))
-      ;; required to make `<SPC> s l' to work correctly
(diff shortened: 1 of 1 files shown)


layers/+lang/javascript/packages.el
```diff
@@ -84,7 +84,7 @@
     :init
     (progn
       (add-to-list 'auto-mode-alist '("\\.js\\'" . js2-mode))
-      ;; required to make `<SPC> s l' to work correctly
+      ;; required to make `<leader> s j' or `<leacer> i j` to work correctly
```
person808: Typo (`<leacer>`)

TheBB:
Can't you just name the feature instead of the keybinding in that comment? Then we don't have to come back to this later.

TheBB:
Thanks! Cherry-picked in develop. You can safely delete your branch. I fixed it up myself.

PR #5229, opened 2016-02-25
fix for perspective handler to have two args

catesandrew:
the `persp-before-switch-functions` handler have to have two arguments.
they are ignored anyways but to reduce friction and unnecessary harm,
adding the signature for two params


bmag:
**Don't merge this PR yet!**

We use a fork of persp-mode, not the persp-mode from melpa. The signature was changed in melpa's version, but not in the fork we use. Merging the PR now would cause an error when switching perspectives.

@catesandrew I suggest replacing `(_new-persp-name _frame-or-window)` with `(&rest _args)`, so it matches any signature change upstream.

catesandrew:
thanks bmag, i'll go ahead and close it since i'm sure when spacemacs switches over from the branch it will be addressed at that time too.

PR #5230, opened 2016-02-25
fix call to correctly go to transient-state

catesandrew:
we want to call workspaces-transient-state vs workspaces-micro-state.
this was probably just an oversight, but nonetheless keeps the fn from
working properly and needs to be updated.


TheBB:

PR #5233, opened 2016-02-25
Provide a way to enter the nil perspective

russell:
Some buffers end up in the nil perspective, this provides a way to
switch back to the nil perspective using the same mechanism that you
use to leave it.

### DIFF
--- a/layers/+window-management/spacemacs-layouts/funcs.el
+++ b/layers/+window-management/spacemacs-layouts/funcs.el
@@ -103,18 +103,20 @@ perspectives does."
    :sources
    (helm-build-in-buffer-source "*Helm Switch Project Layout*"
      :data (lambda ()
-             (if (projectile-project-p)
-                 (cons (abbreviate-file-name (projectile-project-root))
-                       (projectile-relevant-known-projects))
-               projectile-known-projects))
+             (cons persp-nil-name
+                   (if (projectile-project-p)
+                       (cons (abbreviate-file-name (projectile-project-root))
+                             (projectile-relevant-known-projects))
+                     projectile-known-projects)))
      :fuzzy-match helm-projectile-fuzzy-match
      :mode-line helm-read-file-name-mode-line-string
      :action '(("Switch to Project Perspective" .
                 (lambda (project)
                   (let ((persp-reset-windows-on-nil-window-conf t))
                     (persp-switch project)
-                    (let ((projectile-completion-system 'helm))
-                      (projectile-switch-project-by-name project)))))))
+                    (unless (equal project persp-nil-name)
+                      (let ((projectile-completion-system 'helm))
+                        (projectile-switch-project-by-name project))))))))
    :buffer "*Helm Projectile Layouts*"))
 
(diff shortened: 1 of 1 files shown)


russell:
This patch could possibly activate helm-mini on re-entry into the nil perspective? at the moment it just switches back to the perspective.

bmag:
FYI `SPC l 1` switches to the nil perspective, so there is already a way to enter the nil perspective. But another way won't hurt.

layers/+window-management/spacemacs-layouts/funcs.el
```diff
@@ -103,18 +103,20 @@ perspectives does."
...
+                     projectile-known-projects)))
      :fuzzy-match helm-projectile-fuzzy-match
      :mode-line helm-read-file-name-mode-line-string
      :action '(("Switch to Project Perspective" .
                 (lambda (project)
                   (let ((persp-reset-windows-on-nil-window-conf t))
                     (persp-switch project)
-                    (let ((projectile-completion-system 'helm))
-                      (projectile-switch-project-by-name project)))))))
+                    (when (not (equal project persp-nil-name))
```
bmag: Could use `(unless ...` instead of `(when (not ...`

TheBB:
Doesn't `SPC l h` switch to the nil perspective also?

bmag:
It does, somehow I missed it when looking at `SPC l ?` before...

layers/+window-management/spacemacs-layouts/funcs.el
russell: Thanks, too much python for me i think...

russell:
Ah yeah, i didn't find `SPC l` I think I was too busy looking with desc-bindings.  Thanks, for the tip.

russell:
Ok, after some time to think about it, I believe that this patch could be better.

1. It mixes perspective sources and projectile sources, so that means that when the different source provides different behaviour it's surprising to the user.
2. If there are multiple perspectives, then it doesn't list them.

I have an alternative implementation below that instead of adding the nil perspective to the list, it adds in the `spacemacs//helm-perspectives-source` below the projectile source.  Does this seem more reasonable?  I think it would probably provide a better switching experience.

```emacs-lisp
(defun spacemacs/helm-persp-switch-project (arg)
  (interactive "P")
  (helm
   :sources
   `(,(helm-build-in-buffer-source "Switch Project"
     :data (lambda ()
             (if (projectile-project-p)
                 (cons (abbreviate-file-name (projectile-project-root))
                       (projectile-relevant-known-projects))
               projectile-known-projects))
     :fuzzy-match helm-projectile-fuzzy-match
     :mode-line helm-read-file-name-mode-line-string
     :action '(("Switch to Project Perspective" .
                (lambda (project)
                  (let ((persp-reset-windows-on-nil-window-conf t))
                    (persp-switch project)
                    (let ((projectile-completion-system 'helm))
                      (projectile-switch-project-by-name project)))))))
     ,(spacemacs//helm-perspectives-source))
   :buffer "*Helm Projectile Layouts*"))
``

PR #5236, opened 2016-02-25
vim/hybrid: rebind C-k, C-h to M-k, M-h in helm

bmag:
In Helm, when Vim or Hybrid editing style is used, we shadow the useful original bindings of `C-k` and `C-h`. This PR rebinds the missing commands to `M-k` and `M-h`.


nixmaniack:
:+1:

choppsv1:
For hybrid mode C-k should be the normal kill functionality, and not up selection (#5237), so this suggestion should only apply to vim mode, which if #5237 gets fixed would be the only place that needed it. :)

PR #5245, opened 2016-02-25
dvorak-layer: remapping some keybindings

AlejandroCatalina:
Make spacemacs keybindings dvorak programmer friendly.

@StreakyCobra as you can see, your functions are here ;P. We should make them generic and load them at startup time.


StreakyCobra:
Thanks! So we definitively need to group most of the parts as they are nearly identical. I'm not in favor in loading them at startup time though, because most of Spacemacs users don't need this at all.

I have another idea how to regroup both layer in an elegant way, but I have to test it first. Stay tuned :-)

MrOutis:
:+1:

StreakyCobra:
@AlejandroCatalina It's still a WIP, but it would be nice if you can already test my new [`keyboard-layout`](https://github.com/StreakyCobra/spacemacs/tree/contrib/keyboard-layout/layers/%2Bintl/keyboard-layout) layer. You will have to enable it with `(keyboard-layout :variables kl-layout 'dvorak)` for dvorak.

I will be able to test all bepo functionalities, but I'll need you to test all dvorak ones later.

AlejandroCatalina:
Sure, on it!

syl20bnr:
Thank you for the PR 👍 
I'm closing it since the new keyboard-layout is now in develop branch, feel free to reopen PRs against this new layer.

StreakyCobra:
Actually this PR was used as a basis to create the keyboard-layout layer, so probably no need to reopen.

PR #5251, opened 2016-02-25
create align-repeat-backslash command

nocash:
I frequently like to align the backslash character in my code when breaking string and such across multiple lines. This PR adds a menu option for backslash to the align menu.

```bash
 for f in \
   foo    \
   barbaz \
   qux    \
; do
```

### DIFF
--- a/layers/+distribution/spacemacs-base/funcs.el
+++ b/layers/+distribution/spacemacs-base/funcs.el
@@ -689,6 +689,7 @@ the right."
 (spacemacs|create-align-repeat-x "bar" "|")
 (spacemacs|create-align-repeat-x "left-paren" "(")
 (spacemacs|create-align-repeat-x "right-paren" ")" t)
+(spacemacs|create-align-repeat-x "backslash" "\\\\")
 
 ;; END align functions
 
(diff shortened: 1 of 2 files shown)


layers/+distribution/spacemacs-base/funcs.el
```diff
@@ -689,6 +689,7 @@ the right."
 (spacemacs|create-align-repeat-x "bar" "|")
 (spacemacs|create-align-repeat-x "left-paren" "(")
 (spacemacs|create-align-repeat-x "right-paren" ")" t)
+(spacemacs|create-align-repeat-x "backslash" "\\\\")
```
nocash: I was getting a regexp parse error when I tried writing this as `"\\"`. Alternatively this works as `"[\\]"` if that style would be preferred.

layers/+distribution/spacemacs-base/keybindings.el
```diff
@@ -341,6 +341,7 @@
   "xa:" 'spacemacs/align-repeat-colon
   "xa;" 'spacemacs/align-repeat-semicolon
   "xa=" 'spacemacs/align-repeat-equal
+  "xa\\" 'spacemacs/align-repeat-backslash
```
nocash: I didn't align this to the rest of the options listed here for fear that it would obscure the change but can do so if desired.

fbergroth:
This is useful elsewhere, but I'd like to mention that the `shell-scripts` layer binds `SPC m \` to `sh-backslash-region`.

nocash:
> This is useful elsewhere, but I'd like to mention that the shell-scripts layer binds SPC m \ to sh-backslash-region.

:+1: Ah, that's handy to know too. Thanks for pointing it out!

TheBB:

PR #5257, opened 2016-02-26
Disable tilde fringe in shell/comint modes

russell:
Disable the tilde fringe mode in shells, this seems to be the same behaviour that would be expected in VIM.


TheBB:
Note that the package may be exluded by some users, in which case this will break their config because you are calling functions that don't exist.

Can you move the configuration to a `post-init-vi-tilde-fringe-mode` function?

StreakyCobra:

PR #5260, opened 2016-02-26
New layer: column-enforce-mode

dcluna:
This adds the [column-enforce-mode](https://github.com/jordonbiondo/column-enforce-mode/) package for controlling line length.


syl20bnr:
Thank you ! :+1: 
A layer is too much for this but I like the functionality so I replaced the layer by a toggle on `SPC t 8` and `SPC t C-8`, for now we cannot pass an universal argument to a toggle so the length is fixed to 80 columns, PR welcome to improve this :-) Related commit: https://github.com/syl20bnr/spacemacs/commit/828ae81

PR #5263, opened 2016-02-26
Exit layout transient state on select custom layout

CestDiego:
This was asked for in the gitter chat. I thought at first it would be
great for when people wanted to open many perspectives at the same time,
but now I don't think the same. It'd be very hard to modify the macro
for adding custom layouts to make this argument optional, so I leave it
at @syl20bnr discretion.


AlejandroCatalina:
This would be actually great :+1:

TheBB:

Issue #5244, opened 2016-02-25
Org mode agenda view switch


lengyueyang:
I have a problem after update the spacemacs today.

When I opended the agenda view, the RET, switch to the same window, but before its the other window

TheBB:
Can you update your issue with precise reproduction steps, expected and actual effect?

NJBS:
This seems like expected behavior if I understand your issue correctly. RET will open the org file in the same window while TAB will open it in another window.

lengyueyang:
Thanks TheBB and NJBS
I want to use RET to open org file in another window.

NJBS:
You can add the following to your .spacemacs configuration file:
``` elisp
  (with-eval-after-load 'org-agenda
    (define-key org-agenda-mode-map (kbd "RET") 'org-agenda-goto))
```

TheBB:
`TAB` and `RET` are bound to the same function on my system, and it opens the entry in the other window.

NJBS:
On my system 'TAB' is bound to 'org-agenda-goto' whilst 'RET' is bound to 'org-agenda-switch-to'. It seems like this is consistent with the documentation [here](http://orgmode.org/manual/Agenda-commands.html).

StreakyCobra:
On my system I have the same behavior as @TheBB :stuck_out_tongue:

StreakyCobra:
The only place in the code where `org-agenda-switch-to` is defined is in the org agenda transient state.

NJBS:
I don't have the chance to test this on `develop` at the moment, do you still experience the same behavior on `master`?

StreakyCobra:
Oh yeah `master`, I always forget about this one :sweat:

Indeed, in master they have the behavior you described. So this is fixed in `develop`. @lengyueyang If all your questions have been answered, you may consider closing the issue, otherwise let us know what you need more :-)

lengyueyang:
Thanks for all guys, I sloved the problem as NJBS's comments

tommyjiang:
Just want to remind someone else if wants to map TAB to `org-agenda-switch-to` can use the code as following:

```elisp
  (with-eval-after-load 'org-agenda
    (define-key org-agenda-mode-map (kbd "<tab>") 'org-agenda-switch-to))
```

PR #5267, opened 2016-02-26
Fix org-agenda RET behavior

NJBS:
This should fix the inconsistencies seen in https://github.com/syl20bnr/spacemacs/issues/5244


StreakyCobra:
Isn't it already the correct behavior on `develop`?

TheBB:

PR #5270, opened 2016-02-27
Latex layer: add more key bindings

kccai:


TheBB:

PR #4792, opened 2016-01-25
Add IRC buffers to respective perspectives

CestDiego:
This attempts to fix https://github.com/syl20bnr/spacemacs/issues/4781

Also add configurability for the default binding for the IRC perspectives.

PR #5272, opened 2016-02-27
Layouts and IRC

### LINKED ISSUE #4781: spacemacs-layouts: erc doesn't play well with layouts
When creating a layout and then opening an erc connection, the erc buffers land in the default layout and not in the currently viewed one.

TheBB:
This PR includes #4792 and fixes #4781. I noticed some problems with it while merging, and the fixes became extensive enough that it warrants a separate PR for review.

The problem is that use-package expands macros eagerly, which means that `spacemacs|define-custom-layout` has to be ready very early in the loading process. Therefore I've moved all the defuns and defmacros into `packages-funcs.el`.

The other problem is that `spacemacs|define-custom-layout` doesn't work with symbols for the name and the binding, so they have to be evaled.

Finally, since persp-mode loads itself during its own init function, any post-config hooks added in post-init functions won't run. I took the chance to simplify the rcirc and erc post-init functions a bit. (It would work if they were pre-init functions, but I prefer this to the use-package hooks.)

A feature I would like is to have a centralized list of servers to join. I redefine the custom ERC layout to make it just log on to the servers I am interested in instead of asking me every time. Unfortunately this means I also have to copy the add-buffer-to-persp code, otherwise if I run my custom ERC layout that won't work. But this PR does not implement that. (At least not yet.)

CC @CestDiego  @choppsv1


CestDiego:
I see this problem .. I wish I could have more time to solve it...I'll try to solve the problem explained in the last sentences this week..

TheBB:
@syl20bnr @CestDiego I was about to merge this but there was a non-trivial rebase involved, so I rebased and will leave it to warm for another day or two.

TheBB:
In it goes. Let's hope there are no bugs.

PR #5280, opened 2016-02-27
Add predicate to exclude open files from being added to recentf

### LINKED ISSUE #4973: agenda and todo in spacemacs-buffer pollutes recent files
With `agenda` and `todos` specified in the spacemacs buffer, the top of my recentf list now includes all org files.  I have noticed this behavior with default org usage as well.  Is it possible to advise org to work with recentf?  I don't want to but an exclusion for all org files, but I think files not interactively accessed should now be added to the list.

ralesi:
Fixes #4973 , recentf will no longer add files to the list that have never been viewed.


ralesi:
This should clear all the org files loading from `org-agenda` and any other major mode that loads files for information without the intent to view them.  Because we are taking a snapshot of the recentf-list at file load to keep the files that were not viewed this session, it might take clearing the recentf-list completely to get the non-viewed files off the list.

layers/+distribution/spacemacs-base/packages.el
```diff
@@ -847,7 +852,23 @@ below. Anything else exits."
...
+
+      (defsubst spacemacs//recentf-viewed-predicate (file)
+        "Return non-nil if FILE was previously in recentf list or has been
+displayed during this session."
+        (and (recentf-keep-default-predicate file)
+             (or (member file spacemacs-recentf-on-load)
+                 (spacemacs//file-was-viewed-p file))))
+
+      ;; And, of course, you now need:
+      (setf recentf-keep '(spacemacs//recentf-viewed-predicate)))))
```
bmag: why `setf` and not `add-to-list`? By overriding the existing value, we risk interfering with user's personal configuration.

layers/+distribution/spacemacs-base/packages.el
```diff
@@ -847,7 +852,23 @@ below. Anything else exits."
...
+      (defsubst spacemacs//file-was-viewed-p (file)
+        "Return non-nil if FILE's buffer exists and has been displayed."
+        (interactive)
+        (let ((buf (find-buffer-visiting file)))
+          (> (buffer-local-value 'buffer-display-count buf) 0)))
+
+      (defsubst spacemacs//recentf-viewed-predicate (file)
+        "Return non-nil if FILE was previously in recentf list or has been
+displayed during this session."
+        (and (recentf-keep-default-predicate file)
```
bmag: If the `setf` below is replaced with `add-to-list`, then we don't need to check `recent-keep-default-predicate` here.

kccai:
Maybe I met a new issue here or maybe not. But I'd like to ask you guys first ...

If I set the `dotspacemacs-startup-lists` as follows,
```
   dotspacemacs-startup-lists '(agenda recents bookmarks projects)
```
then the `recentf-exclude` variable will be **reset** to `nil` (originally, spacemacs sets it to be `("COMMIT_EDITMSG\\'" "/home/user/.emacs.d/elpa" "/home/user/.emacs.d/.cache/")`

But if I **change the order** of the startup list like as follows,
```
   dotspacemacs-startup-lists '(recents agenda bookmarks projects)
```
then the `recentf-exclude` will be the default value set by spacemacs. 

I don't know why this happened... I really want the agenda list to be at the top of the home buffer without affecting the `recentf-exclude` set by spacemacs ... Could you help?

ralesi:
@bmag, awesome I didn't realize recentf-keep accepted lists of funcs. Will update the PR.

ralesi:
@kccai, is this after applying this PR,  or in the current develop branch?

kccai:
@ralesi 
I am in the current develop branch, not applying this PR. Hoping that this PR could solve this problem. Thanks~

ralesi:
@bmag, I updated with your suggestions, much better just adding to the `recentf-keep` list.  Thanks.

ralesi:
@kccai , I do not see the same issue as you.  Perhaps you can make an issue with the steps to repo?

kccai:
@kccai Thanks, I will check it again and create an issue if this problem persists.

ralesi:
I am finding that org files are still being added, testing out applying changes as an exclude predicate.

syl20bnr:
What's the status on this PR ?

ralesi:
@syl20bnr, I have git it working,  but if agenda is listed before recent files in the startup list,  the org files still get added.   Will get it fixed and update the PR tonight.

syl20bnr:
@ralesi thank you :+1:

ralesi:
@syl20bnr , I take that back.  I cannot come up with a method to keep the `recentf-list` clean during startup.  The method with the predicate works, but requires one to run `recentf-cleanup` after load, which is enough of a preference that I don't think it should be added to one of the default hooks.  Optimally, we would have an exclusion predicate placed on `recentf-exclude` if the opened file is not viewed, but that hook runs right after `after-find-file` and the buffer is not completely set by `set-window`, which is what increments the buffer display counter. 

If anyone else has other ideas to get this to work consistently with the additional org file loading at startup, I would be open to ideas.

ralesi:
@syl20bnr, i updated and the current implementation works for me.  I created a variable that will bypass recentf saving if let bound.  I have also bound it on my call to org-refile in my own config to prevent that function from adding all the org files to my recent list.

PR #5281, opened 2016-02-27
[WIP] Cleanse the org file headlines

JAremko:
This way the doc headers are more consistent. Also it will help with making org files work both in Emacs and GitHub.

### DIFF
--- a/layers/+frameworks/react/README.org
+++ b/layers/+frameworks/react/README.org
@@ -38,7 +38,7 @@ add =react= to the existing =dotspacemacs-configuration-layers= list in this
 file.
 
 You will also need to install =tern= to use the auto-completion and
-documentation features:
+documentation features
 #+BEGIN_SRC sh
   $ npm install -g tern
 #+END_SRC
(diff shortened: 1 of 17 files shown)


JAremko:
I think I'll also write a test for validating org files in PR's so we will not need to do it all over again. But It will be separate PR

robbyoconnor:
> Clense 

 :laughing:

JAremko:
After some extensive grepery and sedery I forget [the list](https://gist.github.com/JAremko/0d9b65c1ed3758acd306) most of those lines has to be replaced with something that works on the GitHub. 

@robbyoconnor  When it sounds epic I feel less lazy :smiley:

robbyoconnor:
Nothing, I'm just in a funny mood

JAremko:
Actually GitHub seems to work well with multiple capitalisations in org files.

robbyoconnor:
:heart:

JAremko:
Ok. Now I know what to do for realz  :smiling_imp:

PR #5285, opened 2016-02-28
Fix Java layer command documentation.

Tinche:
The "Open project management buffer" shortcut is actually SPC m p p, as per 
https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Blang/java/packages.el#L121
not SPC m p s which does nothing.


robbyoconnor:
I wound recommend closing this and doing it against a feature branch as you now can't update your `develop` to track upstream changes... otherwise this is fine.

TheBB:

PR #5286, opened 2016-02-28
Improve ivy keybindings for hybrid editing style

sooheon:
Previously, there was no distinction between hybrid and vim styles,
which caused hybrid style to inherit some "vim-isms" like C-hjkl
navigation, which led to conflicts. This commit lets hybrid keybindings
adhere more closely to emacs defaults.

The addition of a ivy-want-escape-quit variable allows user selection of
escape key behavior, defaulting to 'minibuffer-keyboard-quit for vim and
hybrid styles.


sooheon:
Now obsolete.

PR #5289, opened 2016-02-28
clojure: add clojure-snippets to clojure layer

AlejandroCatalina:
The purpose of this is to have available right out of the box some nice snippets on clojure mode.

It would be pretty easy to install the package alone without putting it into the layer, but I think this is easier for the user, who might not even know the existence of this package.

I had doubts about putting this on auto-completion layer or clojure layer, I finally put this here as it's directly related to the mode and so.

### DIFF
--- a/layers/+lang/clojure/packages.el
+++ b/layers/+lang/clojure/packages.el
@@ -4,11 +4,12 @@
     cider-eval-sexp-fu
     clj-refactor
     clojure-mode
+    (clojure-snippets :toggle (configuration-layer/layer-usedp 'auto-completion))
     company
     popwin
     rainbow-delimiters
     subword
-   ))
+    ))
 
 
 (defun clojure/init-cider ()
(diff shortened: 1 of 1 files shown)


layers/+lang/clojure/packages.el
```diff
@@ -4,11 +4,12 @@
     cider-eval-sexp-fu
     clj-refactor
     clojure-mode
+    clojure-snippets
```
person808: Maybe use `(clojure-snippets :toggle (configuration-layer/layer-usedp 'auto-completion)` to install package only if the `auto-completion` layer is installed

AlejandroCatalina: That's fine, done!

layers/+lang/clojure/packages.el
```diff
@@ -4,11 +4,12 @@
     cider-eval-sexp-fu
     clj-refactor
     clojure-mode
+    (clojure-snippets :toggle (configuration-layer/layer-usedp 'auto-completion))
```
robbyoconnor: I love this syntax -- it's freaking beautiful

TheBB:

Definitely is correct to put it in the clojure layer.

PR #5290, opened 2016-02-28
avoid being evil in deft with emacs editing style

d12frosted:


StreakyCobra:
I think @AlejandroCatalina implemented a core method recently for doing this.

AlejandroCatalina:
Exactly. I'm on the phone so I don't have the reference, maybe #5111?

On Feb 28, 2016 13:58, Fabien Dubosson <notifications@github.com> wrote:

I think @AlejandroCatalina implemented a core method recently for doing this.

d12frosted:
Yeah, I aslo feel like it's not ideal, that's why I opened a related issue #5291. Thanks for the link though.

d12frosted:
Ah, so you propose to use `spacemacs/normal-to-insert-state` function instead of this check, right?

AlejandroCatalina:
Yes I think that would work.

On Feb 28, 2016 14:12, Boris <notifications@github.com> wrote:

Ah, so you propose to use spacemacs/normal-to-insert-state function instead of this check, right?

d12frosted:
Done, thanks @AlejandroCatalina

TheBB:

PR #5293, opened 2016-02-28
also exit layout selection on number key

choppsv1:
Recently layout microstate was changed to exit on layout selection using the key sequence `SPC l o ...` this change provides the symetric behavior for `SPC l ` followed by a number.


TheBB:
We need to keep `SPC l` and `SPC l w` as consistent as possible. Can you see to it?

choppsv1:
Eivind Fonn <notifications@github.com> writes:

> We need to keep `SPC l` and `SPC l w` as consistent as possible. Can you see to it?

I've updated the PR to include changing workspaces to exit on
"switch-to" as well.

Thanks,
Chris.

TheBB:

PR #5297, opened 2016-02-28
Fix inline code block

johankj:
Added a zero-width space between `,` and `~` because emphasis blocks won't get parsed if they end with a comma.

This makes the inline code block show up correctly on GitHub and hopefully [the website](http://spacemacs.org/doc/DOCUMENTATION.html#orgheadline145) as well.
It's not a pretty hack, but it works.

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -263,7 +263,7 @@ error during the download of the starting theme.
   in a future version.
 *** New conventions
 - Commit and abort commands conventions:
-  - ~SPC m ,~ and ~SPC m c~ to Valid/Confirm
+  - ~SPC m ,​~ and ~SPC m c~ to Valid/Confirm
   - ~SPC m a~ and ~SPC m k~ to Abort/Discard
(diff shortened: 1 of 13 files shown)


StreakyCobra:
This is not the only place in the code having this (remind me of https://github.com/syl20bnr/spacemacs/pull/2699). This may help you:

    $ cd .emacs.d/
    $ ag -Q ",~" -G org

johankj:
Good catch. I've updated the commit with all broken code blocks I could find (those starting/ending with `,`, `'` or `"`).

robbyoconnor:
This is also against `master` -- Please close and reopen against `develop` and read our [contribution guidelines](https://github.com/syl20bnr/spacemacs/blob/develop/CONTRIBUTING.org) before you do anymore contributing ;)

johankj:
Don't know how I missed that. Sorry about that.
See #5305 for a PR against develop.

robbyoconnor:
It's fine, this is why I was gentle about it

PR #5298, opened 2016-02-28
Spacemacs as an package

declanqian:
Use spacemacs as an package.  

https://www.reddit.com/r/emacs/comments/47y2eg/use_spacemacs_as_an_elisp_package/

It will not effect the other lisp code.

### DIFF
--- a/spacemacs.el
+++ b/spacemacs.el
@@ -19,4 +19,9 @@
   "Overwrite the spacemacs's `configuration-layer/delete-orphan-packages'
   to make it not remove any orphan packages.")
 
+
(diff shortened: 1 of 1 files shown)


robbyoconnor:
Why?

syl20bnr:
While I like the idea we won't support it officially because of the management overhead it will induce by requiring more tests and creating new issues.

If people wants to use this clever hack then they can use quelpa as shown in the reddit post.

declanqian:
@robbyoconnor For fun? Anyway ignore this. Another way of doing thing ha.

declanqian:
@syl20bnr Ok, I am closing this.

PR #5300, opened 2016-02-28
Update Scheme docs to curl over https

davbo:
Felt more comfortable running this over https. Not sure what the risks
are of piping files into a root privilege tar process, but the server
supports it so may as well use it.


TheBB:

PR #5306, opened 2016-02-29
Ignore rake.cache in the repository

asok:
The file is used by package [rake](https://github.com/asok/rake)

### DIFF
--- a/layers/+lang/ruby/README.org
+++ b/layers/+lang/ruby/README.org
@@ -15,11 +15,13 @@
    - [[Tests][Tests]]
      - [[RSpec-mode][RSpec-mode]]
      - [[Ruby-test-mode][Ruby-test-mode]]
+   - [[Rake][Rake]]
 
 * Description
 This layer provides support for the Ruby language with the following feature:
 - version manager (rbenv, rvm or chruby)
 - test runner (ruby-test and rspec)
+- rake runner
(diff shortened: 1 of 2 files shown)


TheBB:
https://github.com/asok/rake/blob/master/rake.el#L61-L65

Can you configure rake to place it in Spacemacs' cache directory instead?

asok:
I could. But I would need to handle two cases - when someone is using spacemacs and when someone is not using spacemacs.
I guess that the userbase of the `rake` package in majority are people who don't use spacemacs.

TheBB:
You misunderstand. Instead of changing gitignore, instead have Spacemacs change the value of `rake-cache-file` to live in .emacs.d/.cache instead. We do it with all the other cache-type files so we might as well do it with this one too.

TheBB:
Or to be blunt: please turn this PR into something that `(setq rake-cache-file ...)`.

asok:
> Or to be blunt: please turn this PR into something that (setq rake-cache-file ...)

Thanks. Indeed this looks much better.

I've updated the PR. I've added the `rake` package to the ruby layer.  The package was already included in `ruby-on-rails` layer (via the dependency on `projectile-rails` package). But for me it makes sense to have a rake runner for every ruby project since it's so omnipresent.

TheBB:

Issue #4851, opened 2016-01-28
Spacemacs introduction does not propose hybrid style


DamienCassou:
When starting spacemacs for the first time, `emacs` and `vim` styles are proposed, but not `hybrid`.

StreakyCobra:
I thought it has been corrected, it seems that no :-) Maybe there was a reason…

nixmaniack:
`hybrid` is the spacemacs invented style(AFAIK) and people aren't aware of it until they try spacemacs. OTOH, `emacs` and `vim` are well known styles.

izahn:
I think that in context the meaning of `hybrid` is clear. That is, given `vim/emacs/hybrid` most will correctly infer that hybrid uses both Emacs and Vim conventions and bindings. I don't think it will confuse to many people to include it, and IMO hybrid is actually the most useful of the three styles, so I would like to see if offered by the first-run wizard thingy.

sooheon:
there is no mention of `hybrid` in the `dotspacemacs-editing-style` docstring, either.

d12frosted:
Fixed by #5308

PR #5308, opened 2016-02-29
core: fix dotspacemacs-editing-style docstring

nixmaniack:
Fix https://github.com/syl20bnr/spacemacs/issues/4851#issuecomment-189995246


StreakyCobra:
You probably also need to synchronize `core/templates/.spacemacs` :-)

nixmaniack:
@StreakyCobra This is reverse sync! Template was updated already but not the variable.

TheBB:

PR #5309, opened 2016-02-29
Fix use of deprecated function avy--with-avy-keys

### LINKED ISSUE #5307: avy--with-avy-keys deprecated
After a recent update of avy, `spacemacs/ace-buffer-links` no longer works since `avy--with-avy-keys` is no longer defined.

Found this on the `avy` page: https://github.com/abo-abo/ace-link/issues/21

For now, I'm working around this with:

```elisp
  (with-eval-after-load 'avy
    (when (and (not (fboundp 'avy--with-avy-keys))
              (fboundp 'avy-with))
     (defalias 'avy--with-avy-keys 'avy-with)))
```

but should probably be changed in spacemacs.  Thanks for the great work!

NJBS:
Fixes #5307

### DIFF
--- a/layers/+spacemacs/spacemacs-ui/packages.el
+++ b/layers/+spacemacs/spacemacs-ui/packages.el
@@ -56,10 +56,10 @@
       (defun spacemacs/ace-buffer-links ()
         "Ace jump to links in `spacemacs' buffer."
         (interactive)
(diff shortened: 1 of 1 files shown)


sooheon:
It's a small thing but also fixing the indentation would be nice :)

TheBB:

PR #5310, opened 2016-02-29
Make cljr-keybinding-prefix configurable

sooheon:
Previously this was hardcoded as C-c C-f, which is a fine default, but
some users may prefer to change it.


a13ph:
> "fine default"
> 4 keys with 2 chords

PR #5311, opened 2016-02-29
Rename: ensime-refactor-inline-local -> ensime-refactor-diff-inline-l…

chessman:
Function has different name in ensime-emacs (see ensime-refactor.el). Inlining works now.


TheBB:
Thanks! Cherry-picked in develop.

I'm leaving this open for the hotfix.

Issue #5265, opened 2016-02-26
react: % key doesn't work for html/jsx tags


tko:
Say you have the following, possibly being returned from a function or something
```html
<div>
    <h1>Hello world</h1>
</div>
```
with the cursor placed in the opening `div` pressing <kbd>%</kbd> should move the cursor to the closing tag and vice versa.

In `web-mode` it works as expected (including jumping between `{` and `}`) but not in `react-mode`. In `react-mode` the cursor jumps to the `v` in both tags instead.

robbyoconnor:
You have to add evil-matchit hooks for react-mode -- PRs welcome.

robbyoconnor:
Be sure however to do it against `develop` and not `master`.

syl20bnr:
Fixed in 0.105.12

PR #5312, opened 2016-02-29
react: make evil-matchit jump between html/jsx tags as well

tko:
Fixes: #5265


TheBB:

PR #5315, opened 2016-03-01
Make pylookup use helm for completion

darkfeline:
Make pylookup use helm for completion

### DIFF
--- a/layers/+lang/python/packages.el
+++ b/layers/+lang/python/packages.el
@@ -199,7 +199,8 @@
       (let ((dir (configuration-layer/get-layer-local-dir 'python)))
         (setq pylookup-dir (concat dir "pylookup/")
               pylookup-program (concat pylookup-dir "pylookup.py")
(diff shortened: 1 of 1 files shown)


TheBB:
Is it possible to just set it to the value of `completing-read-function` to have it work with ivy as well?

darkfeline:
`pylookup-completing-read` is `ido-completing-read` for me.

In pylookup:

~~~ emacs-lisp
(defvar pylookup-completing-read 
  (if (null ido-mode) 'completing-read 'ido-completing-read)
  "Ido support with convenience")
~~~

I guess the right way would be to set it to `completing-read`?

darkfeline:
While we're at it, might as well fix this: https://github.com/syl20bnr/spacemacs/pull/5328

TheBB:
Yes, I think so. Thanks! :+1:

TheBB:

I took the liberty of changing the commit message since it's no longer about helm.

PR #5318, opened 2016-03-01
fix osx-use-options-as-meta on emacs 25

d12frosted:
By default it's already `meta`, so when you set `osx-use-options-as-meta` to `nil` - option is still `meta`. This fixes the issue on Emacs 25, and should still work on previous versions of Emacs (though I haven't tested).

CC @brenogil

### DIFF
--- a/layers/osx/keybindings.el
+++ b/layers/osx/keybindings.el
@@ -9,8 +9,8 @@
 
     (when osx-use-option-as-meta
       ;; Treat option as meta
-      (setq mac-option-key-is-meta t)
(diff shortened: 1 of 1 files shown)


d12frosted:
P. S. not sure what `mac-option-key-is-meta` is for.

sooheon:
emacs wiki says its a carbon-emacs specific var: https://www.emacswiki.org/emacs/MetaKeyProblems Kinda messy to set it for all distros, but not too harmful either, I guess.

TheBB:

PR #5319, opened 2016-03-01
shell-scripts: Enable flycheck in sh-mode

fbergroth:
`fish-mode` has no checkers, but maybe it should be enabled there as well, in case of a future checker?


TheBB:

PR #5331, opened 2016-03-02
Fixes #5301 - Define new variable: spacemacs-start-directory.

### LINKED ISSUE #5301: How to isolate *.cache* and *elpa* path?
I want to manage several profiles in one git repository. Each profile looks like something *user@host*. So, the convenient way is to put each profile into a directory and then link this directory as **~/.spacemacs.d**, and then link **spacemacs** as **~/.emacs.d**. The problem is that spacemacs installs packages into **ELPA** and keeps temporary files in **.cache**, which is not good for many profiles. How to isolate the two dirs and keep them as dirs in each profile?

exaos:
With this new variable, user can load spacemacs anywhere, e.g. "~/.emacs.d/spacemacs/",
and user can keep there temporary data inside their own "~/.emacs.d/". If user want to
use spacemacs this way, drop one line as the below into "~/.emacs.d/init.el":

  (setq spacemacs-start-directory "~/.emacs.d/spacemacs/")
  (load-file (concat spacemacs-start-directory "init.el"))

This also solve my own problem in this issue: https://github.com/syl20bnr/spacemacs/issues/5301

### DIFF
--- a/init.el
+++ b/init.el
@@ -20,7 +20,10 @@
     (message (concat "Your version of Emacs (%s) is too old. "
                      "Spacemacs requires Emacs version %d or above.")
              emacs-version spacemacs-emacs-min-version)
-  (load-file (concat user-emacs-directory "core/core-load-paths.el"))
+  (setq spacemacs-start-directory
+        (if (boundp 'spacemacs-start-directory)
+            spacemacs-start-directory user-emacs-directory))
+  (load-file (concat spacemacs-start-directory "core/core-load-paths.el"))
   (require 'core-spacemacs)
   (spacemacs/init)
   (spacemacs/maybe-install-dotfile)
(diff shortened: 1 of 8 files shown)


syl20bnr:
Put on hold.
We need to test it a bit, I'll merge this PR on my machines to test it during a couple of weeks.

init.el
```diff
@@ -20,7 +20,10 @@
     (message (concat "Your version of Emacs (%s) is too old. "
                      "Spacemacs requires Emacs version %d or above.")
              emacs-version spacemacs-emacs-min-version)
-  (load-file (concat user-emacs-directory "core/core-load-paths.el"))
+  (setq spacemacs-start-directory
```
bmag: A `defvar` is more appropriate here, and eliminates the need for a manual `boundp` check.
```elisp
(defvar spacemacs-start-directory user-emacs-directory)
```

syl20bnr:
Thank you for the PR :+1: 

I close this PR in favor of #5410 because it is much simpler and better fit Spacemacs.
This change is too invasive and I don't like shadowing the stock variable `user-emacs-directory`, it may surprise a lof of users.

PR #5334, opened 2016-03-02
evilified: Remove stray bind-map declaration

justbur:
It looks like it was not removed when the other one in the file was
added.


TheBB:

PR #5337, opened 2016-03-02
Fix scalastyle documentation

chessman:
http://www.flycheck.org/2015/11/14/flycheck-0.25.html

The scala-stylestyle syntax checker now expects a scalastyle executable in exec-path now. The flycheck-scalastyle-jar option which used to provide the location of the Scalastyle JAR file was removed.

### DIFF
--- a/layers/+lang/scala/README.org
+++ b/layers/+lang/scala/README.org
@@ -78,9 +78,8 @@ provided by Flycheck.
 
 To use scalastyle,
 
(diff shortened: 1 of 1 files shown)


StreakyCobra:
Thanks for your PR. In Spacemacs all PR have to be made against the `develop` branch. Please open a new PR applying your changes on top of it. You may also consider to read our [`CONTRIBUTING.org`](https://github.com/syl20bnr/spacemacs/blob/develop/CONTRIBUTING.org) guidelines first :-)

PR #5339, opened 2016-03-03
Fixes #5190 - core: Add :packages keyword for layers

### LINKED ISSUE #5190: Proposal to add :use keyword to layer declaration
In the spirit of being explicit about what you would like to use, rather than excluding what you don't want, a `:use` keyword for the `dotspacemacs-configuration-layers` list would be useful.

Specifically for adding on to `spacemacs-base`, currently the layers like `spacemacs-ui` `spacemacs-evil` etc. still come whole hog, and you end up adding most of `spacemacs` default back if you want something in the layers. This would allow you to be more selective, more easily.

```
'(better-defaults
  (spacemacs-ui :use '(info+ window-numbering))
  clojure)
```

justbur:
This implements my proposal in #5190.

To restrict layer to only use pkg1 and pkg2 use
(layer
  :packages (pkg1 pkg2))

To load all packages in a layer except pkg1 use
(layer
  :packages (not pkg1))

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -68,6 +68,10 @@
              :initform nil
              :type list
              :documentation "List of package symbols declared in this layer.")
+   (user-packages :initarg :user-packages
+             :initform nil
+             :type list
+             :documentation "List of package symbols declared by user.")
    (variables :initarg :variables
               :initform nil
               :type list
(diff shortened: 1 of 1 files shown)


justbur:
This is breaking the help facility for some reason

syl20bnr:
Which help facility does it break ?
Is this PR fixable ?

justbur:
I was getting nil packages in ivy-spacemacs-help. You can use it. I just
didn't want to spend the time digging into what was wrong.

syl20bnr:
Ok no problem, I'll look at it, thank you.

ralesi:
@syl20bnr , @justbur.  I can confirm the presence of nil layers in the `helm-spacemacs-help` and `ivy-spacemacs-help` variants; however this is found in the current develop branch, and is not tied to this PR.  I applied the changes in this PR to the current head and it seems to be working fine.   I did find a bug in `ivy-spacemacs-help`, but nothing pertaining to the `:packages` property implementation.

PR #5340, opened 2016-03-03
layouts: remove "toggle help" text string

nixmaniack:
Transient state's toggleable docstring should conventionally be on `?`
so it can be removed to save space.

before:
screenshot 2016-03-03 13 37 40

after:
screenshot 2016-03-03 13 52 59


bmag:
On the other hand, having that help line is good for discoverability. IMO the benefit of discoverability outweighs the lost screen estate. Lets see what others think.

Lknechtli:
why not just put it on the same line, possibly on the right side of the screen? No lost screen estate then.

robbyoconnor:
:-1:

nixmaniack:
It's not needed anymore as syl20bnr added some nice changes to transient state which includes help at right side.

PR #5341, opened 2016-03-03
highlight current layout and workspace in transient state

nixmaniack:
Not sure about the right face but this looks decent I think!

screenshot 2016-03-03 14 01 07

### DIFF
--- a/layers/+window-management/eyebrowse/packages.el
+++ b/layers/+window-management/eyebrowse/packages.el
@@ -106,13 +106,15 @@ workspace, preferably in the current window."
 
 
       (defun spacemacs//workspace-format-name (workspace)
-        (let ((current (eq (eyebrowse--get 'current-slot) (car workspace)))
-              (name (nth 2 workspace))
-              (number (car workspace)))
-          (concat
-           (if current "[" "")
-           (if (< 0 (length name)) name (int-to-string number))
-           (if current "]" ""))))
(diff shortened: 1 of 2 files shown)


fbergroth:
The `(when current ...)` checks are  not needed because of the parent `(if current ...)`. Maybe the brackets are not needed at all if we highlight it:
```
(if current
    (propertize caption 'face 'highlight)
  caption)
```

nixmaniack:
> The `(when current ...)` checks are not needed because of the parent `(if current ...)`

Right. I'll make the changes.

> Maybe the brackets are not needed at all if we highlight it

I'm not so sure about removing brackets and I leave the decision on syl20bnr. There might be users who don't like colors(or are colorblind) hence we should probably keep it.

fbergroth:
> There might be users who don't like colors(or are colorblind) hence we should probably keep it. 

Good point!

TheBB:
Thanks. I was going to do this myself. Can you do the same for the workspaces transient state?

nixmaniack:
@TheBB Done!

TheBB:

I found the warning face in Monokai to be disappointingly non-obvious. :-(

nixmaniack:
@TheBB I hear you. It's not obvious in Monokai. I think spacemacs should define custom faces here for providing more granularity in configuring faces. It would easier to override using theming then.

PR #5343, opened 2016-03-03
Fix rcirc logging where the channel name has '/' in it

aroig:
This happens when connecting to gitter via irc.gitter.im, for instance.
It turns out that channels are named like #syl20bnr/spacemacs. For the
logging to work, we need to create the directory #syl20bnr first.

### DIFF
--- a/layers/+chat/rcirc/packages.el
+++ b/layers/+chat/rcirc/packages.el
@@ -128,8 +128,6 @@
       (setq rcirc-log-flag t)
       (defun rcirc-write-log (process sender response target text)
         (when rcirc-log-directory
-          (when (not (file-directory-p rcirc-log-directory))
-            (make-directory rcirc-log-directory))
           (with-temp-buffer
             ;; Sometimes TARGET is a buffer :-(
             (when (bufferp target)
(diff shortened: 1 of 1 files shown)


CestDiego:
there are two approaches for this, for whoever plans to solve this;

- Either the directory is created  which would be nice because all of repos from syl20bnr would be under the `#syl20bnr` directory.
- We escape the `/` in the names

TheBB:
@CestDiego This is a PR so presumably @aroig plans to solve it (and maybe has, I haven't looked at it).

aroig:
This PR already solves it by creating the directory `#syl20bnr`.

CestDiego:
I missed this. Sorry. Good job

On Wed, Mar 9, 2016, 2:33 AM Abdó Roig-Maranges <notifications@github.com>
wrote:

> This PR already solves it by creating the directory #syl20bnr.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/syl20bnr/spacemacs/pull/5343#issuecomment-194153198>.
>
-- 
Diego Berrocal
http://cestdiego.github.io

CestDiego:
what's wrong with this PR? this looks like it's a fine solution

TheBB:
I don't think I said there's anything wrong with it. :-P

TheBB:

PR #5342, opened 2016-03-03
[WIP]add function that formats org files for GitHub

JAremko:
Add function `spacemacs/org-githubify-orgbuf`  that formats org files(opened in a buffer) to be viewed via GitHub. 

For now it only rebuilds the table of content with proper links for GitHub. Might be extended to fix all sorts of  render bugs and inconsistencies like removing trailing white spaces or `:` from headings. It can be fine tuned and it doesn't rely on the org mode.

*Part of solution for  #4399*

Issue #4399, opened 2016-01-05
Broken links in some .org files


xfq:
The TOC links in the following files are broken (404):

* https://github.com/syl20bnr/spacemacs/blob/master/doc/DOCUMENTATION.org
* https://github.com/syl20bnr/spacemacs/blob/master/doc/FAQ.org
* https://github.com/syl20bnr/spacemacs/blob/master/doc/CONVENTIONS.org
* https://github.com/syl20bnr/spacemacs/blob/master/doc/LAYERS.org
* https://github.com/syl20bnr/spacemacs/blob/master/doc/QUICK_START.org
* https://github.com/syl20bnr/spacemacs/blob/master/doc/VIMUSERS.org
* https://github.com/syl20bnr/spacemacs/blob/master/layers/LAYERS.org
* layers/**/README.org (I haven't tried all of them)

I think the broken links were introduced in 13c5b1d (Nov 2, 2015).

PS: the TOC links in https://github.com/syl20bnr/spacemacs/blob/master/CONTRIBUTING.org are not broken.

-----
**Update:** Not only TOC links, but most (if not all) internal links.

syl20bnr:
This is a known issue and we have yet to find a solution for this.

TheBB:
I thought we landed on wontfix for this, and to have http://spacemacs.org be the "official" way to view documentation online.

StreakyCobra:
> I thought we landed on wontfix for this, and to have http://spacemacs.org be the "official" way to view documentation online.

We landed on this yes, but imho this is not a viable long-term solution. Search engines are indexing those pages, some users read the documentation on github, some users use the github search box to find information in the project, etc.

robbyoconnor:
Add the wont fix label :D

StreakyCobra:
Nah, we need a better solution. Discussing with @TheBB recently we were seeing a potential solution to have links working everywhere, but we are both kind of busy those days.

louwers:
Add a [custom id](http://orgmode.org/manual/Internal-links.html) to every header and link to those like so:
```
* Spacemacs documentation                                   :TOC_4_org:noexport:
 - [[#core-pillars][Core Pillars]]
   - [[#mnemonic][Mnemonic]]
   - [[#discoverable][Discoverable]]
   - [[#consistent][Consistent]]
   - [[#crowd-configured][Crowd-Configured]]
 - [[#highlighted-feature][Highlighted feature]]
 - [[#screenshots][Screenshots]]
[...]

PR #5345, opened 2016-03-03
[WIP] Function for formatting githubified org files for Spacemacs

JAremko:
Add function `spacemacs/orgify-orgbuf` that formats org file(opened in a buffer) to be viewed via Spacemacs. (see #5342)

It seems to works as intended, but I haven't hooked it yet.  Not sure if it should be a separate PR from #5342 or not, but said PR can't destructively affect Spacemacs and can be safely merged and this one should be carefully evaluated.

Part of proposed solution for #4399

### DIFF
--- a/core/core-funcs.el
+++ b/core/core-funcs.el
@@ -277,4 +277,77 @@ buffer."
           . (,feature . ,repl-func))
         spacemacs-repl-list))
 
+(defun spacemacs/orgify-orgbuf ()
+  "Transforms current README.org file buffer from the GitHub friendly
+form(see `spacemacs/org-githubify-orgbuf' ) to the canonical."
+  (interactive)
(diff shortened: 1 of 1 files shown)


core/core-funcs.el
```diff
@@ -277,4 +277,77 @@ buffer."
           . (,feature . ,repl-func))
         spacemacs-repl-list))
 
+(defun spacemacs/orgify-orgbuf ()
```
JAremko: It might be a good idea to split the function and reuse its parts in  [spacemacs/org-githubify-orgbuf](https://github.com/syl20bnr/spacemacs/pull/5342/files#diff-c1f53f6886fdfa85b5282798acbc2f1eR11)

JAremko:
I'll move all #4399 related stuff to a single PR.

PR #5351, opened 2016-03-03
C-j/C-k bindings in shell only with vim style

sooheon:
This wraps C-k/C-j binding customization in a check for the
dotspacemacs-editing-style variable, and only applies them if vim style
is selected. For hybrid or emacs style users, C-k not killing line and
C-j not sending a form of return will be confusing, and they will be
more used to C-n/p or M-n/p bindings, which exist in the shells already.

### DIFF
--- a/layers/shell/packages.el
+++ b/layers/shell/packages.el
@@ -165,9 +165,10 @@ is achieved by adding the relevant text properties."
 
       ;; These don't work well in normal state
       ;; due to evil/emacs cursor incompatibility
-      (evil-define-key 'insert eshell-mode-map
-        (kbd "C-k") 'eshell-previous-matching-input-from-input
-        (kbd "C-j") 'eshell-next-matching-input-from-input))))
+      (when (eq dotspacemacs-editing-style 'vim)
+        (evil-define-key 'insert eshell-mode-map
+          (kbd "C-k") 'eshell-previous-matching-input-from-input
+          (kbd "C-j") 'eshell-next-matching-input-from-input)))))
 
 (defun shell/init-eshell-prompt-extras ()
   (use-package eshell-prompt-extras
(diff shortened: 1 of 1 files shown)


nixmaniack:
:+1:  +1 as it fixes my editing style i.e. hybrid :) 

Although, I think this would not solve the issue of dynamic switching between `hybrid` and `vim` (which is generally the case if you pair program!).

sooheon:
Ah you're right, I just did a naive check. What do you think would be a clean way to do it? Just add another form to check for the inverse and undo bindings there?

nixmaniack:
I am in the same boat, do not know of clean way here. May be @justbur can suggest. Also there was talk from syl20bnr of reverting to old implementation of hybrid which will have different keymaps which I think is the cleanest way.

In current situation, I think these bindings should be added on entry hook and remove in exit hook of states with guard for editing style. Do I make sense?

justbur:
[See here](https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Bdistribution/spacemacs-base/local/holy-mode/holy-mode.el#L54-L57) for how these bindings are added and removed in holy-mode. 

I don't really understand what is going to happen with hybrid mode now, so I don't have any other suggestions really. Instead of making one-off changes like this, I think @syl20bnr needs to decide on how he wants to design hybrid and holy mode and how to handle the key binding changes. He has suggested reverting both to older designs, and that could affect how this all works. I'm tired of arguing about these things and am just going to wait and work around whatever changes get made.

syl20bnr:
> I think @syl20bnr needs to decide on how he wants to design hybrid and holy mode and how to handle the key binding changes.

Will be something I'll experiment with next week-end, the idea is to put back evil at the center of Spacemacs and use evil states consistently, the effect is to embrace the evil API everywhere to keep a beautiful symmetry accros the system. I feel it is a better design but it has to support all the ergonomic details we discovered so I cannot decide yet.

PR #5354, opened 2016-03-04
Fix cider-pop-back function name

malchmih:
Alias `cider-jump-back` was deprecated and then removed from CIDER here: https://github.com/clojure-emacs/cider/commit/705133f2bdca33cb5bc6e0a5b4861965e0883ee4


syl20bnr:
Thank you :+1: 
Cherry-picked into develop branch but letting it open since it needs to be back ported.

PR #5357, opened 2016-03-04
Fix chruby in the ruby layer

asok:
This commit replaces the code used for finding out the ruby version to
use with `chruby-use-corresponding`.
Unfortunately the function is not marked as autoload thus
the require statement.

### DIFF
--- a/layers/+lang/ruby/packages.el
+++ b/layers/+lang/ruby/packages.el
@@ -57,22 +57,11 @@
 (defun ruby/init-chruby ()
   (use-package chruby
     :if (equal ruby-version-manager 'chruby)
+    :commands chruby-use-corresponding
     :defer t
     :init
     (progn
-      (defun spacemacs//enable-chruby ()
-        "Enable chruby, use .ruby-version if exists."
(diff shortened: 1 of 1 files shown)


nixmaniack:
You can add autoload cookies using `commnads` keyword of `use-package`

From [docs](https://github.com/jwiegley/use-package/tree/d867b0370e4e311c71665ccaa418374a15097461#key-binding)
> When you use the `:commands` keyword, it creates autoloads for those commands and defers loading of the module until they are used.

asok:
Updated

layers/+lang/ruby/packages.el
```diff
@@ -57,22 +57,11 @@
 (defun ruby/init-chruby ()
   (use-package chruby
     :if (equal ruby-version-manager 'chruby)
+    :commands 'chruby-use-corresponding
```
nixmaniack: Quoting isn't required.

asok: Done. Thanks.

TheBB:

PR #5367, opened 2016-03-05
Bind helm-ag--update-save-results

darkfeline:
This updates the helm-ag results buffer, very useful.


TheBB:

PR #5371, opened 2016-03-05
fix typo in dotspacemacs/install

d12frosted:


JAremko:
I remained me about an meme.  :wink:

d12frosted:
Which one? :D 

P. S. I am not a fan of such commits, but since it's in install flow - forced myself to send it :D

Issue #4277, opened 2015-12-23
I think the basic configuration of org should be in the spacemacs layer


justbur:
It's awkward to me at least that basic org mode stuff changes depending on whether you are using the org layer or not, especially since all of the docs are in org format now.

It would make more sense to me to have org be included in the spacemacs layer, and then have the extras in the current org layer become an org-extras or org-goodies layer. That way we can ensure that people have the same basic org configuration. See #4273 for example, where the method for following links depends on whether you use the layer.

nixmaniack:
I agree totally! I think this was mentioned in gitter as well with same issue of following links in org buffers. Since org is built-in package, it makes much sense to configure it in spacemacs in the spacemacs way. The chances of jumping into org files are more as we have documentation in org files and having consistent way of moving around should be configured in spacemacs layer or probably core itself.

syl20bnr:
Indeed.

What kind of rule can we make to decide whether a given org configuration should be in spacemacs or extra layer ?

justbur:
To start anything useful for the docs should be in spacemacs

TheBB:
Anything that is useful for someone who just drops into org mode on a casual basis should be in the org layer. If it's for people who are actually using org as a workflow tool I think it can stay in the org layer (e.g. the agenda).

vitiral:
@TheBB I think you meant "on a casual basis should be in the **spacemacs** layer."

TheBB:
Yep, thanks!

justbur:
Maybe org is sufficiently important to warrant its own spacemacs-org layer that is included in spacemacs by default?

a13ph:
It surely is sufficiently large to be modularized on it's own. It can match helm in numbers of plugins and such.

Some support for reading documentation in org format is important. Jump-to-heading certainly is (though it's not quite as advertised as it should be + I'm not sure about current status of it without org layer enabled)

justbur:
I guess no one wants to implement this

d12frosted:
No-no, reopen it. I was going to address it on these weekends (at least give it a try) :)

justbur:
Oh ok

justbur:
Closing now, because we have the PR for it now #5372

PR #5372, opened 2016-03-05
move org essentials to spacemacs distro

d12frosted:
Following #4277 

---

I haven't touched readme file, because I think it's good to have all description in one place. Also, I didn't rename `org` layer to anything else. If you think that it's a good idea - just tell me (or it could be done via another PR / commit). 

I tested my changes in different situations and everything seems to be working. But since I don't know all use-cases of org layer, I would like to take some help from org users and those who don't use org at all. 

In case you think that something is not in the right place (some configs should be moved from spacemacs-org to org or vice versa) just let's discuss :)

### DIFF
--- a/layers/+spacemacs/spacemacs-org/packages.el
+++ b/layers/+spacemacs/spacemacs-org/packages.el
@@ -1,0 +1,89 @@
+;;; packages.el --- spacemacs-org layer packages file for Spacemacs.
+;;
+;; Copyright (c) 2012-2016 Sylvain Benner & Contributors
+;;
+;; Author: Boris Buliga <d12frosted@d12frosted.local>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+;;; Commentary:
+
+;;; Code:
+
+(defconst spacemacs-org-packages
+  '(
+    (evil-org :location local)
+    flyspell
+    ;; org is installed by `org-plus-contrib'
+    (org :location built-in)
+    (org-plus-contrib :step pre)
+    org-bullets
+    toc-org
+    ))
+
+(defun spacemacs-org/init-evil-org ()
+  (use-package evil-org
+    :commands evil-org-mode
+    :init
+    (add-hook 'org-mode-hook 'evil-org-mode)
+    :config
+    (progn
+      (evil-define-key 'normal evil-org-mode-map
+        "O" 'evil-open-above)
+      (spacemacs|diminish evil-org-mode " ⓔ" " e"))))
+
+(defun spacemacs-org/post-init-flyspell ()
+  (spell-checking/add-flyspell-hook 'org-mode-hook))
+
+;; dummy init function to force installation of `org-plus-contrib'
+(defun spacemacs-org/post-init-org-plus-contrib ())
+
+(defun spacemacs-org/init-org ()
+  (use-package org
+    :mode ("\\.org$" . org-mode)
+    :commands (org-clock-out org-occur-in-agenda-files org-agenda-files)
+    :defer t
+    :init
+    (progn
+      ;; FIXME: This check has been disabled pending a resolution of
+      ;; https://github.com/syl20bnr/spacemacs/issues/3933
+      ;; (when (featurep 'org)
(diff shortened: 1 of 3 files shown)


JAremko:
:+1:  I don't understand why org isn't a part of Spacemacs. Pretty much everyone will want to read README.org files for layers.  And if we going to fix #4399 TOC will look  rather ugly in the fundamental mode.

```
* Spacemacs documentation                                   :TOC_4_org:noexport:
 - [[#core-pillars][Core Pillars]]
   - [[#mnemonic][Mnemonic]]
   - [[#discoverable][Discoverable]]
   - [[#consistent][Consistent]]
   - [[#crowd-configured][Crowd-Configured]]
 - [[#highlighted-feature][Highlighted feature]]
 - [[#screenshots][Screenshots]]
 - [[#who-can-benefit-from-this][Who can benefit from this?]]
```

a13ph:
to play devils advocate, orgmode is still a plaintext, readable fine even on paper or in any plain text editor. navigation somewhat suffers, and it's not as pretty, but it's not hell at all

That said, I agree that at least some support for orgmode is needed closer to the core - it's one of most popular packages and reasons to migrate, as well as currently one of the best documentation formats for emacs users

d12frosted:
> to play devils advocate, orgmode is still a plaintext, readable fine even on paper or in any plain text editor. navigation somewhat suffers, and it's not as pretty, but it's not hell at all

I read and edit org files on iPad as plain text files :/

layers/+spacemacs/spacemacs-org/packages.el
```diff
@@ -0,0 +1,96 @@
...
+      ;;     "Try removing org code from user initialization and private layers.") t))
+
+      (setq org-startup-with-inline-images t
+            org-src-fontify-natively t
+            ;; this is consistent with the value of
+            ;; `helm-org-headings-max-depth'.
+            org-imenu-depth 8)
+
+      (with-eval-after-load 'org-indent
+        (spacemacs|hide-lighter org-indent-mode))
```
syl20bnr: Now we disable org-indent by default we can remove this line.

layers/+spacemacs/spacemacs-org/packages.el
```diff
@@ -0,0 +1,96 @@
...
+
+      ;; Add global evil-leader mappings. Used to access org-agenda
+      ;; functionalities – and a few others commands – from any other mode.
+      (spacemacs/declare-prefix "ao" "org")
```
syl20bnr: Should be in `org` layer if there is no key bindings under this prefix in `spacemacs-org`

syl20bnr:
Nice work, the documentation will need to be split at some point but it can wait because this is something we have to rethink given we have multiple spacemacs layers now.

layers/+spacemacs/spacemacs-org/packages.el
d12frosted: Ah. Forgot that we don't enable it anymore. Good point.

d12frosted:
@syl20bnr thanks for review - fixed what you've asked.

d12frosted:
Ping, ping :cry:

JAremko:
This branch has conflicts that must be resolved :shipit:

d12frosted:
Ok, updated this one as well. Would be nice to have it merged as soon as possible :)

JAremko:
@syl20bnr  :shipit:

TheBB:
Now that we can sort PRs by :+1: reactions on the first post, that is a handy way for people to signal which PRs are most interesting to them. I've just added mine.

JAremko:
And two goes from me

nixmaniack:
> Now that we can sort PRs by :+1: reactions on the first post, that is a handy way for people to signal which PRs are most interesting to them.

This should be mentioned somewhere so that everyone is aware.

JAremko:
Also we can sort by the amount of confusion caused - neat.

d12frosted:
> Don't forger to selfthumb

@JAremko just did it :) 

But do I really want to see it in Spacemacs? Hm. I wonder.

syl20bnr:
Thank you ! :+1: 
I rearranged it a bit in commit https://github.com/syl20bnr/spacemacs/commit/ff7b839

PR #5383, opened 2016-03-06
Fix auto-completion toggle without default

TheBB:
There were some problems toggling auto-completion on if a default was not configured. (You'd get unbound variable error on `auto-complete-mode` for starters.) This commit also ensures that the frontend variable is always defined and that it always defaults to company.


syl20bnr:
Thank you :+1: 
Also released in 0.105.13

PR #5385, opened 2016-03-07
Fixes #5321 - Keep escape binding in evil-visual-state default

### LINKED ISSUE #5321: Why is evil-visual-state <escape> rebound in spacemacs-base?
https://github.com/syl20bnr/spacemacs/blob/master/layers/%2Bdistribution/spacemacs-base/packages.el#L248

`evil-exit-visual-state` is the default. Unless there is a compelling reason, I believe we should keep it the default, just because of strange edge cases. For example, this current rebinding causes problems when entering and exiting visual state and using multiple cursors (yes this is already buggy, but it can be made usable with a in-development branch).

sooheon:
While seemingly equivalent, this binding change can cause brittle
behavior in at least one case (used in conjunction with
multiple-cursors). Such a highly niche customization would probably be
better in a personal config, unless a case can be made that all
spacemacs users would benefit.

Addressed in #5321


syl20bnr:
Thank you ! :+1: 
Released in 0.105.14

PR #5386, opened 2016-03-07
srefactor requires Emas 24.4 or above, not older

tuhdo:


layers/+lang/c-c++/packages.el
```diff
@@ -23,7 +23,7 @@
     helm-cscope
     helm-gtags
     semantic
-    (srefactor :toggle (version< emacs-version "24.4"))
+    (srefactor :toggle (not (version< emacs-version "24.4")))
```
JAremko: `(version<= "24.4" emacs-version)`

tuhdo: No it's not. `srefactor` requires Emacs to be equal to 24.4 or above, not equal to 24.4 or older.

JAremko: That's why I swapped  `"24.4"` and `emacs-version`

tuhdo: Ah right I didn't notice the order.

nixmaniack: @JAremko I find the keyword `:toggle` a bit odd. Maybe it can just take a conditional form like `(if ...` or `(unless ...` etc if I'm understanding its purpose correctly.

TheBB:

PR #5390, opened 2016-03-07
projectile: autoload `projectile-project-p`

### LINKED ISSUE #5127: spacemacs/helm-persp-switch-project assumes projectile is loaded
The `spacemacs/helm-persp-switch-project` function (bound to `SPC p l` by default) assumes that the `projectile` package is loaded (specifically, it assumes that the `projectile-project-p` function is defined). This means that running `SPC p l` immediately after starting spacemacs yields an error. One solution is to add `(autoload 'projectile-project-p "projectile")` somewhere but I don't know if this is the "correct" solution or where this line should go.

bixuanzju:
Fix #5127

### DIFF
--- a/layers/+distribution/spacemacs-base/packages.el
+++ b/layers/+distribution/spacemacs-base/packages.el
@@ -772,14 +772,15 @@ below. Anything else exits."
                projectile-ag
                projectile-compile-project
                projectile-dired
-               projectile-grep
                projectile-find-dir
                projectile-find-file
                projectile-find-tag
                projectile-find-test-file
+               projectile-grep
                projectile-invalidate-cache
                projectile-kill-buffers
                projectile-multi-occur
+               projectile-project-p
                projectile-project-root
                projectile-recentf
                projectile-regenerate-tags


layers/+distribution/spacemacs-base/packages.el
```diff
@@ -781,6 +781,7 @@ below. Anything else exits."
                projectile-kill-buffers
                projectile-multi-occur
                projectile-project-root
+               projectile-project-p
```
StreakyCobra: One line above ;-)

d12frosted: `sort-lines` ! :D

bixuanzju: Sure thing! Since i am on that, `projectile-grep` is also misplaced.

StreakyCobra: :+1:

syl20bnr:
Thank you ! :+1: 
Released in 0.105.14

Issue #5266, opened 2016-02-26
add chrome layer variable for plain text email ?


dwang20151005:
I'd like to use the chrome layer, but I like my email in plain text.  

What do you think of adding a layer variable to the chrome layer to toggle between plain and rich text composition?  Rich text mode would be the current behavior, using gmail-message-mode to render markdown.  Plain text mode would use [edit-server-htmlize](https://github.com/frobtech/edit-server-htmlize) to support plain text email.  

I would be happy to submit a pull request if you approve of the general idea.

jb55:
(triage) it looks like this didn't get much attention/support. should we close this for now?

dwang20151005:
I actually wrote the pull request but closed it eventually.  I'm happy to re-open it if anyone would promise to look at it.  I would still like this to be included, so I don't have to maintain my own fork of the layer.

dwang20151005:
I still want this.

PR #5399, opened 2016-03-07
Support plain-text gmail via chrome layer variable

dwang20151005:
Those who use the chrome layer for gmail can now choose between rich-text
composition (default) or plain-text composition. In plain-text mode, the buffer
is de-htmlized before editing and re-htmlized afterward in a way compatible with
the Plain-text editing mode of gmail's web interface.

This implements #5266 .

As this is my first PR, it's very likely I've overlooked or misunderstood some part of your contribution guidelines. Corrections and advice will be received gratefully. Thanks!

### DIFF
--- a/layers/chrome/README.org
+++ b/layers/chrome/README.org
@@ -23,6 +23,10 @@ To use this configuration layer, add it to your =~/.spacemacs=. You will need to
 add =chrome= to the existing =dotspacemacs-configuration-layers= list in this
 file.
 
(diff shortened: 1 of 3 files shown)


dwang20151005:
Ping?  I don't really know the process.  What happens next?

StreakyCobra:
Unless there are pending comments for your code you don't have anything to do, just waiting. There are 168 other open PR, so it may take time :-)

dwang20151005:
Okay, that's fine now that I understand what's happening.  Thanks.

PR #5401, opened 2016-03-07
Fixes #5392 - Modified dotspacemacs/user-init//config comments

### LINKED ISSUE #5392: Improve description of user-init / user-config
The spacemacs config file has two places to put "user code":

```
(defun dotspacemacs/user-init ()
  "Initialization function for user code.
It is called immediately after `dotspacemacs/init'.  You are free to put almost
any user code here.  The exception is org related code, which should be placed
in `dotspacemacs/user-config'."
  )

(defun dotspacemacs/user-config ()
  "Configuration function for user code.
This function is called at the very end of Spacemacs initialization after
layers configuration. You are free to put any user code."
```

I find those documentation strings lack a bit of information: _Why_ would I choose one over the other / what kind of code goes into which block? Should I put everything that's not relatd to `org-mode` to `user-init`?

E.g. I had enabled `editorconfig-mode` using `setq` in `user-config` because it seemed to me like it was configuration, but it turns out it needs to already be done in init for that package to work.

triklsbg:
Clarify the execution order of dotspacemacs/user-init and dotspacemacs/user-config especially their pre- and post-layers character according to FAQ and issue #5392.

### DIFF
--- a/core/templates/.spacemacs.template
+++ b/core/templates/.spacemacs.template
@@ -247,15 +247,16 @@ values."
 
 (defun dotspacemacs/user-init ()
   "Initialization function for user code.
-It is called immediately after `dotspacemacs/init'.  You are free to put almost
-any user code here.  The exception is org related code, which should be placed
-in `dotspacemacs/user-config'."
+It is called immediately after `dotspacemacs/init', before layer configuration
+executes.  You are free to put almost any user code here.  The exception is
+org related code, which should be placed in `dotspacemacs/user-config'."
   )
 
 (defun dotspacemacs/user-config ()
   "Configuration function for user code.
 This function is called at the very end of Spacemacs initialization after
-layers configuration. You are free to put any user code."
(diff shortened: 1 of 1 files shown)


StreakyCobra:
It would be nice to add something like this saying:

In user-init: This function is mostly useful for variables that need to be set **before** packages are loaded. If you are unsure, you should try in setting them in `user-config` first.

In user-config: This is the place where most of your configurations should be done. Unless it is explicitly specified that a variable should be set before a package is loaded, you should place you code here.

CC @TheBB for opinion?

TheBB:
That sounds fine to me.

In fact personally I prefer it the other way: variables that could be set anywhere I put in `user-init`, but that's entirely cosmetic for the vast majority of them. It sounds reasonable that the rule of thumb should be: always try user-config first.

syl20bnr:
Thank you ! :+1: 
Released in 0.105.14
I modified it to reflect @StreakyCobra suggestions.

chu-:
It's more clear than before for me.

PR #5402, opened 2016-03-07
Add key binding to toggle clojure-enlighten-mode.

mahinshaw:
Add a keybinding to the clojure layer to toggle `cider-enlighten-mode`.

### DIFF
--- a/layers/+lang/clojure/README.org
+++ b/layers/+lang/clojure/README.org
@@ -119,22 +119,26 @@ As this state works the same for all files, the documentation is in global
 ** Leader
 *** Documentation
 
-| Key Binding | Description    |
-|-------------+----------------|
-| ~SPC m h a~ | cider apropos  |
-| ~SPC m h g~ | cider grimoire |
-| ~SPC m h h~ | cider doc      |
-| ~SPC m h j~ | cider javadoc  |
+| Key Binding | Description     |
+|-------------+-----------------|
+| ~SPC m h a~ | cider apropos   |
+| ~SPC m h g~ | cider grimoire  |
+| ~SPC m h h~ | cider doc       |
+| ~SPC m h j~ | cider javadoc   |
+| ~SPC m h n~ | cider browse ns |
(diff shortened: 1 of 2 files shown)


thenonameguy:
:+1:

agzam:
what do you guys think if we also add a shortcut for `'cider-repl-handle-shortcut` (in non-evil mode is `,` in cider-repl) as well? Can be done easily by adding `",," 'cider-repl-handle-shortcut` at: line:266 in packages.el? Then cider-repl's native menu would be available by pressing comma twice?

mahinshaw:
@agzam I like it.  I need to rebase this PR with what is currently on develop.  The cider-repl-mode bindings got merged with all the other mode bindings, and `cider-repl-handle-shortcut` is only relevant to cider mode.

mahinshaw:
Rebased patch so that changes in 63e1db9 are incorporated.  Added @agzam's suggestion, as well as `SPC m s d` to run `cider-repl-clear-buffer`, which was lost in the aforementioned commit (only for `cider-repl-mode`).

agzam:
What is the status of this PR? Will it ever be merged? Can we also expose clojure-mode bindings for "Cider scratch" (Clojure Interaction) mode? It requires an extremely small change, I think it totally can be part of this PR.

mahinshaw:
@agzam I am game for that. Any opinion on the binding? Maybe ~SPC m S~?

agzam:
@mahinshaw this is what I mean, 
here: https://github.com/syl20bnr/spacemacs/blob/dfad604faaa542080d252902babfdd392e354eea/layers/%2Blang/clojure/packages.el#L89

we just need to add `cider-clojure-interaction-mode` then major mode bindings would work in cider-scratch window

mahinshaw:
@agzam Good call. Thanks

mahinshaw:
Squashed into one commit

TheBB:

PR #5403, opened 2016-03-07
Add proselint layer

mkcode:
This layer hooks proselint into flycheck. It defines a toggle on `SPC t P` that turns it on and off. Getting this feedback is emacs is fun.

https://github.com/amperser/proselint/

Feedback is appreciated. I do not think that added this as a checker to flycheck is the ideal solution for this, because it can be always enabled like flyspell. When it enabled, it has the bad side effect of canceling the other flycheck checkers until it is toggled off. This works for me as is. Perhaps there is someway use `flycheck-add-next-checker` with a wildcard?

### DIFF
--- a/layers/proselint/README.org
+++ b/layers/proselint/README.org
@@ -1,0 +1,52 @@
+#+TITLE: proselint layer
+#+HTML_HEAD_EXTRA: <link rel="stylesheet" type="text/css" href="../css/readtheorg.css" />
+
+#+CAPTION: logo
+
+# The maximum height of the logo should be 200 pixels.
+[[img/proselint.png]]
+
+* Table of Contents                                        :TOC_4_org:noexport:
+ - [[Description][Description]]
+ - [[Install][Install]]
+ - [[Key bindings][Key bindings]]
+
+* Description
+
+  This layer hooks [[https://github.com/amperser/proselint][proselint]] into flycheck to making typing go good, obviously.
+
+  While flycheck-proselint is enabled (via the toggle or the config var), it
+  will take precedence and be the ONLY functioning flycheck checker. Toggling
+  it off will return the normal flycheck functioning to the buffer.
+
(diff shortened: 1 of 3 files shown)


mkcode:
I used a predicate rather than the modes for flycheck b/c I sometimes would like it in all modes (for comments in prog-modes). Uncommenting https://github.com/syl20bnr/spacemacs/pull/5403/files#diff-ca5347cef700e452e57c746cd67fd927R30 will apply it to all text modes.

Also, appending in line 32 https://github.com/syl20bnr/spacemacs/pull/5403/files#diff-ca5347cef700e452e57c746cd67fd927R30 has the effect of it being available but not enabled in all modes.

mkcode:
Hmm, maybe I should combine this with writegood-mode: https://github.com/bnbeckwith/writegood-mode

mkcode:
Stumbled across this today: https://news.ycombinator.com/item?id=11415815

Not the author. :-)

syl20bnr:
Interesting, what do you have in mind ? A `prose` layer ?

mkcode:
> Interesting, what do you have in mind ? A prose layer ?

Perhaps that is the way to go. There is a lot to play around with and explore here, with the the other writegood-mode and artbollocks-mode (I think we would want either one or the other).

I'm keen on having this implemented as a toggle for minor mode - rather than a major mode (or both) so that I could quickly get feedback for my english in any context.

JAremko:
Really cool, but I think it should be more general than just a linting layer. See #5996

JAremko:
@mkcode Can you rename it to prose layer  ?

mkcode:
> Can you rename it to prose layer ?

Sure. Will do later today.

Very glad this getting some attention now 👍

TheBB:
Superseded by #6012.

mkcode:
@JAremko - Done in #6012. which replaces this PR.

PR #5407, opened 2016-03-08
Latex layer: fix the all in one command (~SPC m a~)

kccai:
~TeX-command-run-all~ always asks to save file or not. Now with ~latex/all~
the .tex files are saved without asking.

### DIFF
--- a/layers/+lang/latex/funcs.el
+++ b/layers/+lang/latex/funcs.el
@@ -20,6 +20,14 @@
     ;; (when build-proc
     ;;   (set-process-sentinel build-proc 'latex//build-sentinel))))
 
+(defun latex/all ()
+  "Save, build and view in one command."
+  (interactive)
+  (progn
+    (let ((TeX-save-query nil))
+      (TeX-save-document (TeX-master-file)))
+    (TeX-command-run-all ())))
+
 (defun latex//build-sentinel (process event)
   (if (string= event "finished\n")
       (TeX-view)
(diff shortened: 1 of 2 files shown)


layers/+lang/latex/funcs.el
```diff
@@ -20,6 +20,14 @@
     ;; (when build-proc
     ;;   (set-process-sentinel build-proc 'latex//build-sentinel))))
 
+(defun latex/all ()
+  "Save, build and view in one command."
+  (interactive)
+  (progn
+    (let ((TeX-save-query nil))
+      (TeX-save-document (TeX-master-file)))
```
TheBB: Note that `(TeX-master-file)` may be a different file than the file the user is editing (which is presumably the one that is not saved).

TheBB: Oh, looks like `TeX-save-document` can deal with that. Neat.

Though I'm still not sure that I want a command that saves (other) files without explicit consent.

kccai: @TheBB Thanks for the comment :smiley: . 

In my case, it is a little annoying that the `TeX-command-run-all` keeps asking me to save the .tex files when I decide to compile them. I just need my files to be compiled after I finish editing them without asking me to do the saving. So I would assume that the users know their files would be saved before compiling, otherwise they would not compile the files.

kccai:
The keybindings `, a` and `, b` provide almost the same functions. I am closing this pull request and will try to find another way to avoid the duplicate functions.

PR #5410, opened 2016-03-08
Don't hard-code location of elpa directory, use package-user-dir where appropriate

bmag:
Currently the elpa directory is hard-coded to `<user-emacs-directory>/elpa` (usually `~/.emacs.d/elpa`).
With this PR, users can change the location of the elpa directory by changing the value of `package-user-dir`.
Users interested in changing the location of the elpa directory should set `package-user-dir` in `dotspacemacs/user-init`.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -1335,7 +1335,7 @@ to select one."
                  (pkg-dir-name (cdr apkg))
                  (installed-ver
                   (configuration-layer//get-package-version-string pkg))
-                 (elpa-dir (concat user-emacs-directory "elpa/"))
+                 (elpa-dir (file-name-as-directory package-user-dir))
                  (src-dir (expand-file-name
                            (concat rollback-dir (file-name-as-directory
                                                  pkg-dir-name))))
@@ -1436,7 +1436,7 @@ to select one."
     (cond
      ((version< emacs-version "24.3.50")
       (let* ((version (aref (cdr pkg-desc) 0))
-             (elpa-dir (concat user-emacs-directory "elpa/"))
+             (elpa-dir (file-name-as-directory package-user-dir))
              (pkg-dir-name (format "%s-%s.%s"
                                    (symbol-name pkg-name)
                                    (car version)
--- a/core/core-funcs.el
+++ b/core/core-funcs.el
@@ -49,7 +49,7 @@ FILE-TO-LOAD is an explicit file to load after the installation."
 
 (defun spacemacs//get-package-directory (pkg)
   "Return the directory of PKG. Return nil if not found."
-  (let ((elpa-dir (concat user-emacs-directory "elpa/")))
+  (let ((elpa-dir (file-name-as-directory package-user-dir)))
     (when (file-exists-p elpa-dir)
       (let ((dir (cl-reduce (lambda (x y) (if x x y))
                          (mapcar (lambda (x)


core/core-configuration-layer.el
```diff
@@ -1335,7 +1335,7 @@ to select one."
                  (pkg-dir-name (cdr apkg))
                  (installed-ver
                   (configuration-layer//get-package-version-string pkg))
-                 (elpa-dir (concat user-emacs-directory "elpa/"))
+                 (elpa-dir package-user-dir)
                  (src-dir (expand-file-name
                            (concat rollback-dir (file-name-as-directory
                                                  pkg-dir-name))))
```
emacs18: The default value of package-user-dir does not have trailing slash unlike the code that it is replacing.  Hence this change by itself is not a valid change.  You need to make matching change so that the code works as it did before whether or not package-user-dir has a trailing slash.  For example you can  change expand-file-name on line 1342 like this:

```
(expand-file-name (file-name-as-directory pkg-dir-name) elpa-dir)
```
This works because expand-file-name works whether or not the second argument has trailing slash.

core/core-configuration-layer.el
```diff
@@ -1436,7 +1436,7 @@ to select one."
     (cond
      ((version< emacs-version "24.3.50")
       (let* ((version (aref (cdr pkg-desc) 0))
-             (elpa-dir (concat user-emacs-directory "elpa/"))
+             (elpa-dir package-user-dir)
              (pkg-dir-name (format "%s-%s.%s"
                                    (symbol-name pkg-name)
                                    (car version)
```
emacs18: Same problem exists on line 1444 which can be fixed by changing it with
```
(expand-file-name pkg-dir-name elpa-dir)
```

core/core-funcs.el
```diff
@@ -49,7 +49,7 @@ FILE-TO-LOAD is an explicit file to load after the installation."
 
 (defun spacemacs//get-package-directory (pkg)
   "Return the directory of PKG. Return nil if not found."
-  (let ((elpa-dir (concat user-emacs-directory "elpa/")))
+  (let ((elpa-dir package-user-dir))
```
emacs18: The fact that you are potentially removing trailing slash has no impact in this case, because elpa-dir is used only in contexts where this does not matter.

emacs18:
This is an excellent submission. This change would enable users to
customize the package install directory merely by customizing
package-user-dir.

The ability to over-ride package-user-dir is important for those who use
multiple versions of emacs (e.g., 24.x and 25.x) where it is important
to avoid sharing byte compiled files.

bmag:
@emacs18 good catch. I've fixed it by wrapping `package-user-dir` with `file-name-as-directory` to ensure `elpa-dir` has a trailing slash.

syl20bnr:
Cool :+1: 
Thank you!

d12frosted:
Wow. I forgot about this PR. Thanks @bmag. This makes testing with different versions of Emacs much easier and pleasant. There is no need of funky scripts for this anymore :D

bmag:
You're welcome :)

mandarvaze:
@d12frosted I came to this page via your blog, where you asked to put the following into ``dotspacemacs/user-init`` 

    ;; setup package-user-dir to allow seamless switch between emacs versions
    (setq package-user-dir (file-name-as-directory
                        (concat user-emacs-directory "elpa/" emacs-version)))

Unfortunately, this breaks the following "powerline" related fix in the ``dotspacemacs/user-config`` 
    (setq powerline-default-separator 'utf-8)

May be the order is wrong ?
As soon as I comment ``package-user-dir`` line in the ``dotspacemacs/user-init``  (and **restart** spacemacs, the powerline is "fixed". Note : ``SPC f e R`` is not enough. I must ``SPC q q `` and start emacs again)

As you can imagine, I am newbie, so do not know intricacies of emacs.
Any suggestion on how to get both to work together would help.

bmag:
@mandarvaze those two code snippets are not related. In any case, on develop you don't need the `package-user-dir` code anymore. Instead, you should change the value of `dotspacemacs-elpa-subdirectory` in your dotfile from `nil` to `emacs-version`.

@d12frosted you might want to update that in your blog.

mandarvaze:
@bmag  I know (they are not related) But I see that they (somehow) don't work with each other
do you have any troubleshooting tips ?
BTW, I upgraded to emacs 25.1 (hence using this "trick") but I have my ``.emacs.d`` and ``.emacs.d/elpa`` from emacs 24.5. Could that be causing this issue ?
I'm on spacemacs version ``0.105.19``

d12frosted:
@bmag lol, indeed I should updated it. Once I update it in my config 😊 

@mandarvaze Ok, since you are not on develop, you have to use this snippet. I would recommend clearing `~/.emacs.d/elpa` before first run after you add that snippet (better to cache in case something unexpected happens).

But this is unrelated to your problem anyways. I believe that you actually have to set powerline separator in init function as well.

bmag:
@mandarvaze using the same elpa directory with different emacs versions is not recommended, because of incompatibility issues with the byte compiled (`.elc`) files. Using `.emacs.d/elpa` from emacs 24.5 could be causing an issue with 25.1, yes.

Specifically when changing something about powerline or spaceline, in master I belive you need to manually call `spaceline-compile` for the changes to take effect, so try that just in case.

It's possible that changing `package-user-dir` triggers an error that disrupt the loading of your configuration. Some troubleshooting tips:
- run emacs with `emacs --debug-init`, check for errors on startup and suspicious messages in the `*Messages*` buffer (`--debug-init` also makes spacemacs more verbose when loading)
- check the values of `package-user-dir` and `powerline-default-separator` with `SPC h d v` to verify that the values changed (or didn't)
- check if manually changing `powerline-default-separator` via `M-:` or `ielm` interface has any visible effects

PR #5412, opened 2016-03-08
set spacemacs badge url to spacemacs.org

### LINKED ISSUE #5409: Move the spacemacs badge to the spacemacs.org
I think we can move [Built with Spacemacs](http://github.com/syl20bnr/spacemacs) and it's link to [spacemacs.org](http://spacemacs.org) so the web site will get more search engine love.  

For html:
```html
<a href="http://spacemacs.org">
<img src="http://spacemacs.org/img/spacemacs-badge.svg" />
</a>
```

Im not sure how popular spacemacs badges are, 
but I think it won't hurt https://github.com/syl20bnr/spacemacs  too much, while it should be a meaningful improvement for the site.

The only problem that I see is that spacemacs.org doesn't work via https and it will make badges less popular. But the site should have ssl anyway. [In 2016..](https://letsencrypt.org/)

JAremko:
I changed the URL of the Spacemacs badge to http://spacemacs.org  so the web site will get  more attention and the search engine rank from links.

Closes #5409


syl20bnr:
Thank you :+1: 

PR #5415, opened 2016-03-09
[scala] Add function to insert types and add yank type at point

d1egoaz:
Add function to insert types and add yank type at point

- `, i y` Add function to yank type at point but just yank type name without full
package

- `, r a` Add function to insert type annotation i.e.:

Demo

Tested with the following examples:
```scala
  private val startTime = startDate.toDateTimeAtStartOfDay.getMillis
  private val endTime = endDate.toDateTimeAtStartOfDay.getMillis

  for {
    a <- List(1, 2)
    b <- Map(1 -> 1)
    (c, d) <- Map("" -> true)
  } yield 1

  def test = 1
  def test2() = 1
  def test3(x: Int) = 1
  def test4(x: Int = 3) = 1
  def test5(x: Int = 3, b: (Int, String) = (1, "")) = 1
  def test6(x: Int = 3, b: (Int, String) = (1, ""))(implicit z: (Int, Boolean)) = 1
  def sum[A](xs: List[A])(implicit m: List[A] ) = m
  def default1(dateStr: String = "a", b: (Int, Int))(implicit x: Int) = { List("") }

  def default(
    dateStr: String = "a", b: (Int, Int)
  )(implicit x: Int) = AuthenticatedAction { request =>
```

This function is being added to upstream package as well, https://github.com/ensime/ensime-emacs/pull/390 when it's got merged I'll update spacemacs reference to new function

### DIFF
--- a/layers/+lang/scala/README.org
+++ b/layers/+lang/scala/README.org
@@ -193,6 +193,7 @@ the ascii arrows back.
 | ~SPC m i i~ | inspect type at point                  |
 | ~SPC m i I~ | inspect type in other frame            |
 | ~SPC m i p~ | inspect project package                |
+| ~SPC m i y~ | yank type at point                     |
 
 *** Server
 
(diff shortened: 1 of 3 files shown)


d1egoaz:
Basic functionality to add type to a declaration or symbol.
I don't care about spaces/formatting as in Scala layer we can user `, r f` to format source.

d1egoaz:
^ Just updated, keybindings are more related to refactor rather than inspect

StreakyCobra:
Looks good to me :-)

d1egoaz:
Updated with upstream/ensime-emacs functions as change got merged there before here.
See https://github.com/ensime/ensime-emacs/pull/390

diego-alvarez-hs:
Just a reminder for administrators to merge this PR, Thanks :smile:

syl20bnr:
Thank you !

PR #5442, opened 2016-03-10
Use evil-define-minor-mode-key over define-key

### LINKED ISSUE #5418: spacemacs-toggle-visual-line-navigation-on is leaky
#### Description
The visual line navigation toggle (at `spc t L`) does two things, toggle visual-line-mode, and also rebind `j` and `k` keys to the visual line equivalents in motion states. The problem is that visual-line-mode is a buffer local mode, but the rebindings are global, thus leaking out to other buffers.

#### Reproduction guide
- Start Emacs
- Open the scratch buffer, and `spc t L`
- describe-key to confirm that `j` and `k` are visual line equivalents.
- `spc f e d` to open dotfile, describe-key confirms that `j/k` are still visual movement, but visual-line-mode should not be activated.

#### System Info
- OS: darwin
- Emacs: 24.5.1
- Spacemacs: 0.105.13
- Spacemacs branch: develop (rev. 895455d)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: ivy
- Layers:
```elisp
((auto-completion :variables auto-completion-tab-key-behavior 'cycle auto-completion-return-key-behavior nil auto-completion-enable-sort-by-usage t)
 better-defaults colors clojure dash emacs-lisp erc
 finance html git github org osx pandoc
 syntax-checking spacemacs-ivy themes-megapack vinegar sooheon)
```

sooheon:
Use evil-define-minor-mode-key over define-key

This keeps the keybinding specific to the visual-line minor mode, and
fixes #5418. Reverting the bindings on the :off toggle is no longer
necessary as well, turning off visual-line-mode will handle it.

### DIFF
--- a/layers/+distribution/spacemacs-base/keybindings.el
+++ b/layers/+distribution/spacemacs-base/keybindings.el
@@ -172,24 +172,15 @@
   :status visual-line-mode
   :on (progn
         (visual-line-mode)
-        (define-key evil-motion-state-map "j" 'evil-next-visual-line)
-        (define-key evil-motion-state-map "k" 'evil-previous-visual-line)
+        (evil-define-minor-mode-key 'motion 'visual-line-mode "j" 'evil-next-visual-line)
+        (evil-define-minor-mode-key 'motion 'visual-line-mode "k" 'evil-previous-visual-line)
         (when (bound-and-true-p evil-escape-mode)
           (evil-escape-mode -1)
           (setq evil-escape-motion-state-shadowed-func nil)
-          (define-key evil-motion-state-map "j" 'evil-next-visual-line)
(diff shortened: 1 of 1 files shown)


TheBB:
> Reverting the bindings on the :off toggle is no longer necessary as well, turning off visual-line-mode will handle it.

So is there still a reason why we need to enable the bindings in the `:on` as opposed to just doing it once and forgetting about it?

sooheon:
@TheBB Nothing I suppose, except the benefit of having the code in the place where it is relevant, and not running it when you don't need it? Is there a better place for it?

TheBB:
Hm, maybe not. I thought this was part of a package function for some reason.

syl20bnr:
Thank you :+1: 

PR #5457, opened 2016-03-11
Fix error when using `spacemacs/counsel-search' wihout initial input.

demon386:
Currently when `use-initial-input` is false, `initial-input` will become `nil` and will then make `(rxt-quote-pcre initial-input)` pop error.


TheBB:
Thanks! Cherry-picked in develop.  You can safely delete your branch.

PR #4836, opened 2016-01-27
Replace mu4e multiple accounts with contexts

darkfeline:
mu4e 0.9.16 now comes with a native contexts feature that completely
replaces the old hack for multiple accounts.

PR #5458, opened 2016-03-11
Fix problems with multi-account setup in mu4e

aroig:
Currently, account guessing for mu4e with multiple accounts has several problems:

- Detection of email in `To` fields is broken.
- It misses `CC` field or `From` field (replying to oneself)
- if we are replying and no guess can be made, it fails.

This commit rewrites the guessing function in a more robust maner, so that all the points above are addressed, and if no guess can be made, it just prompts the user asking for an account.

### DIFF
--- a/layers/+email/mu4e/funcs.el
+++ b/layers/+email/mu4e/funcs.el
@@ -9,39 +9,38 @@
 ;;
 ;;; License: GPLv3
 
-(defun mu4e//search-account-by-mail-address (mailto)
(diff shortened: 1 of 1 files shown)


aroig:
I just became aware of mu4e's 0.9.16 contexts, which I am going to use right away! In light of #4836 We can probably throw away this PR then... sorry for the noise!

StreakyCobra:
Closing as suggested. Feel free to reopen at any time if you change your mind :-)

Issue #5426, opened 2016-03-09
Add Vim-style remaps with Evil-Remap to ease transition of Vim users!


a13ph:
https://github.com/GuiltyDolphin/evil-remap/blob/master/evil-remap.el

Not too old of a package, should be fairly healthy

And vim remap style is very easy, so it might benefit others too.

PR #5465, opened 2016-03-12
Adds the evil-remap package to spacemacs layers

calebmeyer:
# Rationale

Spacemacs is intended (at least in part) for vimmers who have
never used emacs, but may already be very comfortable in vim. Often one
of the first things you learn in vim is to put something like
`imap jj <Esc> " use jj for returning to normal mode`
in your .vimrc

Spacemacs already covers this case pretty well, but there are lots of
other changes you might want to make. Another common one:
`nnoremap ; : " set ; to summon ex-mode`

Unfortunately, today there are many ways to set a keybinding in emacs.
global-set-key, local-set-key, define-key, etc. Worse still, most of the
documentation online doesn't take evil into account. Also, keys aren't
defined simply, via a string. You have to use (kbd "blah") instead.I
personally ran into this problem when starting with spacemacs. I wanted
to map a key in normal mode, but global-set-key was getting overridden
by the already defined behavior of that key in normal mode (a more
specific binding).

Evil-remap allows vim-users to bind keys simply. The functions all
follow the same format as vim:
`<mapping_type> <key to map> <behavior to map it to>`
# Implementation

This layer is very simple, since it's my first foray into emacs lisp.
Based on the documentation for creating a layer, I took the package
definition from my additional-packages and changed the require call to
be a use-package call. I also added documentation based on the
documentation on the package itself. I tested it locally in my private
layers, then moved it to make this PR.

### DIFF
--- a/layers/+vim/evil-remap/README.org
+++ b/layers/+vim/evil-remap/README.org
@@ -1,0 +1,47 @@
+#+TITLE: evil-remap layer
+#+HTML_HEAD_EXTRA: <link rel="stylesheet" type="text/css" href="../css/readtheorg.css" />
+
+* Table of Contents                                        :TOC_4_org:noexport:
+ - [[Description][Description]]
+ - [[Install][Install]]
+ - [[Functions][Functions]]
+ - [[Keybindings][Keybindings]]
+
+* Description
+This layer allows you to use vim-like commands to set new keybindings.
+
+For example, to map semicolon to open the ex modeline:
+
+Without this layer:
+#+begin_src emacs-lisp
+(define-key evil-normal-state-map (kbd ";") 'evil-ex)
+#+end_src
+
+With this layer:
+#+begin_src emacs-lisp
+(evil-nnoremap ";" 'evil-ex)
+#+end_src
+
+* Install
+To use this layer, add it to your =~/.spacemacs=
+
+#+begin_src emacs-lisp
+  (setq-default dotspacemacs-configuration-layers '(evil-remap))
+#+end_src
+
+* Functions
+
+| Function       | Description                                                                      |
+|----------------+----------------------------------------------------------------------------------|
+| evil-noremap   | Maps a key to a function in the normal, visual, and operator pending states.     |
+| evil-nnoremap  | Maps a key to a function in the normal state.                                    |
+| evil-vnoremap  | Maps a key to a function in the visual state.                                    |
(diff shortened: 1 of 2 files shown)


syl20bnr:
Thank you for the PR.

I'm not sure about this package because I feel that it does not serve the user well in the end for at least these reasons:
- the first obviously is that it hides one the most important things to learn in Emacs which is how to configure key bindings.
- then what are the limits of this package ? Is it complete enough to avoid violation of POLA for Spacemacs users ?
- Evil has the properties to be a pretty complete Vim UX emulation and to be an extensible layer. Does this cover the additional Evil states we have like `iedit` ?
- the package is young and we should be extra careful to propose it officially.

calebmeyer:
> the first obviously is that it hides one the most important things to learn in Emacs which is how to configure key bindings.

Do we have good documentation surrounding this today? If not, that may be a good alternative to this.

> then what are the limits of this package ? Is it complete enough to avoid violation of POLA for Spacemacs users ?

Huh, I've always heard it as the principle of least surprise. Threw me for a second. I'm not sure I understand the package well enough to answer this one.

> Evil has the properties to be a pretty complete Vim UX emulation and to be an extensible layer. Does this cover the additional Evil states we have like `iedit`?

It does not.

> the package is young and we should be extra careful to propose it officially.

Yeah, someone recommended on gitter that if I was going to use the package that I should make a layer of it :) It was a pain to set up manually, since all the other packages I've used have been on Melpa or elpa, and since I had to explicitly `(require 'evil-remap)` in my dotfile before I could use it.

a13ph:
solves #5426 <3

Kazark:
Given that this has not been merged, how does one go about `nnoremap ; :`? I am brand-new to Spacemacs and my muscle-memory is screwing me without this.

sdwolfz:
@Kazark There is an example in the PR itself.

```emacs-lisp
(define-key evil-normal-state-map (kbd ";") 'evil-ex)
```

Put this inside `dotspacemacs/user-config`.

Kazark:
@sdwolfz Thanks. That's the bootstrapping I needed as a complete n00b. One of the things I am looking forward to here is a chance to dive a little deeper into Lisp. I would much rather work in Lisp than VimScript, despite being moderately good at VimScript, which is one of the things that attracts me to Spacemacs. I am a diehard functional programmer, but have lamentably little experience with Lisps. Anyhow, thanks.

PR #5466, opened 2016-03-12
[mu4e] Use new layer format

darkfeline:
Use new layer format for mu4e


darkfeline:
This is woefully out of date

PR #5479, opened 2016-03-13
SPC cq: close more compilation windows

TheBB:
`SPC c q` only closes buffers called `*compilation*`, but compilation-mode buffers may have other names.

### DIFF
--- a/layers/+distribution/spacemacs-base/funcs.el
+++ b/layers/+distribution/spacemacs-base/funcs.el
@@ -900,7 +900,8 @@ is nonempty."
 (defun spacemacs/close-compilation-window ()
   "Close the window containing the '*compilation*' buffer."
   (interactive)
-  (delete-windows-on "*compilation*"))
+  (when compilation-last-buffer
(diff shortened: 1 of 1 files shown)


joehillen:
@TheBB should this binding be changed to `SPC c d` in order to be consistent with the other window/buffer closing functions?

TheBB:
Yeah, sounds like a good idea.

phinicota:
Maybe unrelated but, anyone considered using SPC c Q to close all compilation windows everywhere and SPC c q to close all opened compilations inside current frame?
I find this useful for when I have multiple emacsclient runnning and I want to close the *compilation* buffer in the current window to see the code better but no in the other emacsclient.

TheBB:
I pulled this in and changed it to `SPC c d`.

@phinicota You can open an issue with this suggestion maybe?

phinicota:
[here](https://github.com/syl20bnr/spacemacs/issues/5987) it is, added some stuff.

PR #5485, opened 2016-03-14
Add keybind: SPC-fj should run deer, not dired

sooheon:
This seems more likely to be the desired action if the user chooses the ranger layer.

### DIFF
--- a/layers/+tools/ranger/packages.el
+++ b/layers/+tools/ranger/packages.el
@@ -19,7 +19,8 @@
     (progn
       (spacemacs/set-leader-keys
         "ar" 'ranger
-        "ad" 'deer)
+        "ad" 'deer
+        "fj" 'deer)
 
       ;; set up image-dired to allow picture resize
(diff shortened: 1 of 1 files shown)


TheBB:
To me, this (as well as all other dired things) end up with deer already. CC @ralesi

StreakyCobra:
> end up with deer already

Not in my setup :-|

ralesi:
Set `ranger-override-dired` to always use deer.

TheBB:
Should we do that instead maybe?

sooheon:
@ralesi That setting seems to be buggy for me. It works when ranger/deer is initialized before dired, but if I directly call dired before ranger is initialized, then it wont' take effect.

syl20bnr:
Thank you for the PR!
May be fixed by #5563 @sooheon can you confirm ?

sooheon:
Can confirm, ranger-override-dried in pre-init fn now works as intended.

PR #5486, opened 2016-03-14
Added FAQ entry for remapping paste key

### LINKED ISSUE #5070: How to change paste behavior in evil mode
I am coming from vim world and I used to use yank -> paste to replace text with source text multiple times. 
Currently if I do [y] copy -> [p] paste -> [p] paste multiple times I get following result: 
animation

I would love to set default behavior to paste without remembering text that was replaced or to move this text to 2nd place and not replacing originaly yanked text. 
Shortly I would like to have result of above commands just `one` on each line.

Is there any way to do that in spacemacs?

Immortalin:
Fixes #5070


syl20bnr:
Thank you !

PR #5489, opened 2016-03-14
use toml-mode for all cargo configs

Stebalien:


layers/+lang/rust/packages.el
```diff
@@ -45,7 +45,7 @@
 
 (defun rust/init-toml-mode ()
   (use-package toml-mode
-    :defer t))
+    :mode "/\\(Cargo.lock\\|\\.cargo/config\\)\\'"))
```
Stebalien: I believe this is the correct way to do this and it works but I'm not entirely sure.

StreakyCobra:

PR #5490, opened 2016-03-14
fix file extension matching for filenames with newlines

Stebalien:
* "\\.ext$" matches both "abc.ext" and "abc.ext\nsomething".
* "\\.ext\\'" matches only "abc.ext".


joehillen:
Filenames with newlines!? Well I never.

syl20bnr:
Thank you !

PR #5493, opened 2016-03-15
Corrected key bindings for Swift major mode

xinranmsn:
Key bindings are not consistent with the documentation. Since the bindings are under major mode, they should not have the prefix 'm'.


syl20bnr:
Thank you !

PR #5497, opened 2016-03-15
spacemacs-ui-visual: enable vi-tilde-fringe only when graphic

nixmaniack:
Using emacs-nox, enabling `vi-tilde-fringe` produces following error on startup, hence exclude it on non-graphic display system.

```
vi-tilde-fringe-mode: Symbol's function definition is void: define-fringe-bitmap
Error in post-command-hook (global-vi-tilde-fringe-mode-check-buffers): (void-function define-fringe-bitmap)
```

### DIFF
--- a/layers/+spacemacs/spacemacs-ui-visual/packages.el
+++ b/layers/+spacemacs/spacemacs-ui-visual/packages.el
@@ -16,7 +16,7 @@
         neotree
         smooth-scrolling
         spaceline
-        vi-tilde-fringe
+        (vi-tilde-fringe :toggle (display-graphic-p))
         (zoom-frm :location local)))
 
 (defun spacemacs-ui-visual/init-fancy-battery ()


StreakyCobra:
Isn't [`spacemacs|do-after-display-system-init`](https://github.com/syl20bnr/spacemacs/blob/develop/layers/+spacemacs/spacemacs-ui-visual/packages.el#L370) enough? Because with this people starting in daemon mode will not be able to enable it in GUI frames.

nixmaniack:
@StreakyCobra You're right. If that's wrapped around in that macro, `vi-tilde-fringe` init block shouldn't have been executed. I don't understand that part throughly, may be @travisbhartwell could explain. 

From what I understand [here](https://github.com/syl20bnr/spacemacs/blob/82fdd9a511f1f5cc4f371b91ce691308686f8ec6/core/core-display-init.el#L35), what has happened in my case is, it has fallen back to normal loading which causes `vi-tilde-fringe`'s `:init` block to get executed causing failure. If my understanding is correct we might need to change this. 

Also I'm unable test server/client and GUI combination being on mac port at the moment.

nixmaniack:
Since using `display-graphic-p` isn't good solution here, closing this for now.

PR #5501, opened 2016-03-15
Define new variable: spacemacs-start-directory.

exaos:
With this new variable, user can load spacemacs anywhere, e.g. "~/.emacs.d/spacemacs/".
Only user's cache directory is still hard-coded as "~/.emacs.d/.cache/". If user want
to use spacemacs this way, drop one line as the below in "~/.emacs.d/init.el":

    (setq spacemacs-start-directory "~/.emacs.d/spacemacs/")
    (load-file (concat spacemacs-start-directory "init.el"))

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -32,11 +32,11 @@
   "Configuration layer templates directory.")
 
 (defconst configuration-layer-directory
-  (expand-file-name (concat user-emacs-directory "layers/"))
+  (expand-file-name (concat spacemacs-start-directory "layers/"))
   "Spacemacs contribution layers base directory.")
 
 (defconst configuration-layer-private-directory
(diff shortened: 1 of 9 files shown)


exaos:
I prefer this way to solve my issue https://github.com/syl20bnr/spacemacs/issues/5301
P.S. so many hard coded `user-emacs-directory` in spacemacs. A new variable might be good.

syl20bnr:
I was against this PR but the 👍  of @TheBB made me think about it and why not, it offers new possibilities for users. I documented this new approach in the README.md.
Thank you ! 👍 

d12frosted:
> Only user's cache directory is still hard-coded as "~/.emacs.d/.cache/"

What do you mean? There is `spacemacs-cache-directory`. So you can easily set that value as well. If `~/.emacs.d/.cache/` is used directly somewhere this is wrong and let us know or open PR 😸

PR #5503, opened 2016-03-15
Add TLS flag for ERC layer

jhenahan:
Just adding a layer variable to allow users to use `erc-tls` rather than `erc` in the custom layout. Happy to make any changes if there's a nicer way to do this.

### DIFF
--- a/layers/+irc/erc/packages.el
+++ b/layers/+irc/erc/packages.el
@@ -233,7 +233,9 @@
                           (persp-get-by-name
                            erc-spacemacs-layout-name)))
       (add-hook 'erc-mode-hook #'spacemacs-layouts/add-erc-buffer-to-persp)
-      (call-interactively 'erc))))
+      (if erc-enable-tls
+          (call-interactively 'erc-tls)
+        (call-interactively 'erc)))))
 
 (defun erc/post-init-smooth-scrolling ()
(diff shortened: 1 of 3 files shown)


TheBB:
I'd rather there were a configurable variable with the actual servers to use. No IRC user in the universe wants to type this info every time they connect anyway. Adding a TLS or no TLS variable makes it 99% annoying instead of 100%.

More or less what I said in https://github.com/syl20bnr/spacemacs/pull/5272.

layers/+irc/erc/packages.el
```diff
@@ -233,7 +233,9 @@
                           (persp-get-by-name
                            erc-spacemacs-layout-name)))
       (add-hook 'erc-mode-hook #'spacemacs-layouts/add-erc-buffer-to-persp)
-      (call-interactively 'erc))))
+      (if erc-enable-tls
+          (call-interactively 'erc-tls)
+        (call-interactively 'erc)))))
```
TheBB: `(call-interactively (if erc-enable-tls 'erc-tls 'erc))`

jhenahan:
@TheBB Isn't that what `erc-autojoin-channels-alist` is for?

jhenahan:
I think this is obsoleted by #5506, so I'll just close it.

PR #5506, opened 2016-03-16
Add server config to the ERC layer

jhenahan:
All right, so @TheBB has been talking about this for a bit (see #5272, for instance), so I took a crack at adding some support for default servers in the ERC layer. It's a tiny bit uglier than I might like, but it's markedly *less* ugly than manually typing in servers every time you connect.

This likely subsumes my last PR (#5503), though I could rework this to take it into account.

### DIFF
--- a/layers/+irc/erc/README.org
+++ b/layers/+irc/erc/README.org
@@ -9,6 +9,7 @@
    - [[OS X][OS X]]
    - [[Social graph][Social graph]]
    - [[Default servers][Default servers]]
+     - [[Security Note][Security Note]]
  - [[Key bindings][Key bindings]]
  - [[Spacemacs Layout Support][Spacemacs Layout Support]]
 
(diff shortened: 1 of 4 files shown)


syl20bnr:
Can we find a way to hide the password from the dotfile and implement and/or document it?

jhenahan:
I think ERC checks your `~/.authinfo.gpg`, by default, and you should be able to fetch the password any other way you like from the config (reading an env variable, or whatever). I can work on documenting some of the password management options.

TheBB:
I like it.

The password can be defined outside the dotfile as a variable, as I do it right now.

One nitpick, I think the API can be like this:

```elisp
              '((erc :variables
                     erc-server-list
                     '(("irc.freenode.net"
                        :port "6697"
                        :ssl t
                        :nick "some-user"
                        :password "secret")
                       ("irc.myworkirc.net"
                        :port "1234"
                        :nick "some-suit"
                        :password "hunter2")))))
```

And you can then define the `erc//server` function as something along these lines.

```elisp
(defun erc//server (erc-func &rest args)
  (apply erc-func :server args)
  ```

  With so many short there is probably not a huge benefit in splitting the functionality up, either?

**Comment by jhenahan** [comment]
That's exactly the kind of feedback I was looking for. My elisp is less than great. :D 

Splitting things up like that was mainly to make it easier to set the leader keybinding to allow it to work the same in the layout and outside of it, but I think with your pointers it should still work. Just a moment to test and I think we'll be all set with the new API.

File: layers/+irc/erc/README.org [hunk 1]
```diff
@@ -44,6 +46,41 @@ notifications via the OS X Notification Center.
...
+    that goes double if you version your config file. ERC allows for a number
+    of ways of protecting this information.
+
+    First, ERC will check your =~/.authinfo.gpg=, looking for lines like
+
+    #+BEGIN_SRC shell
+      machine <irc.server.url> login <yournick> password <yourpassword> port <portnumber>
+    #+END_SRC
+
+    You can leave password blank in this case.
```
calebmeyer: Does this mean 
```elisp
:password '')
;; or
:password nil)
```
or just omitting that key entirely?

TheBB:

Works very well for me. I had to do https://github.com/syl20bnr/spacemacs/commit/9e1909f5e0513bee51a85a79b3047179e4a7b45e and https://github.com/syl20bnr/spacemacs/commit/e6948533dff0a5e1ccf246908df92855bf405272.

PR #5525, opened 2016-03-17
Rust: add key binding for `cargo fmt`

dmit:
Rustfmt is an optional component, but it's a popular one and will likely be officially supported once it leaves the nursery. It is used by rust-mode and is already mentioned in the Install section of the README for the Rust layer.


StreakyCobra:

PR #5532, opened 2016-03-18
Allow = to be used to increase transparency

sooheon:
-,= is easy to go back and forth, without needing to press shift for +.


TheBB:

PR #5541, opened 2016-03-19
ansible: update auto-mode-alist for host/group vars

### LINKED ISSUE #5537: Ansible layer: Activate yaml mode for ansible-filename-re automatically
There is `ansible/ansible-filename-re` variable which defines patterns where should be ansible minor mode activated. Problem is that this is used in hook which is load after `yaml` mode is active - but it never happens for files not ending with .yaml extension.

I believe that this layer should be responsible for "full ansible experience". I expect it to automatically turn on `yaml` mode for files according to recommended ansible layout.

This code in user init provide expected behaviour:

    (add-to-list 'auto-mode-alist
                 '("\\(group_vars/.+\\|host_vars/.+\\)" . yaml-mode))

I'm not creating pull request as I'm not sure where and how exactly to add this code in right place.

Also I think that ansible mode should install `yaml` package and not rely on user to activate it. I'm not sure if there is way to depend on another layer or if it is problem to install one package from multiple layers. At least please add to README that `yaml` layer should be activated.

nixmaniack:
Fix #5537

### DIFF
--- a/layers/+config-files/ansible/packages.el
+++ b/layers/+config-files/ansible/packages.el
@@ -16,7 +16,10 @@
         yaml-mode))
 
 (defun ansible/init-ansible ()
-  (use-package ansible :defer t))
+  (use-package ansible
+    :defer t
+    :init (add-to-list 'auto-mode-alist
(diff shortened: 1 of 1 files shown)


TheBB:

PR #5547, opened 2016-03-20
Ignore tag files.

choppsv1:
ignore various tag files etags: TAGS, global: GPATH, GRTAGS, GTAGS, and id-utils: ID

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,8 @@
+GPATH
+GRTAGS
+GTAGS
+ID
+TAGS
 auto-save-list/
 elpa/
 export/


aaronjensen:
Unless spacemacs is generating these during normal usage, these should probably be in your [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) rather than in the project.

robbyoconnor:
I agree with @aaronjensen -- unless we generate these -- they belong in your gitignore -- not in spacemacs.

choppsv1:
This seems reasonable.

Issue #1823, opened 2015-06-02
Docs have some notation inconsistencies


MerelyAPseudonym:
There are two consistency problems that I've come across:
1. Sometimes keybindings are just written in backticks (example: `M-m f f`) and other times they're written using HTML `<kbd>` markup (example: <kbd>M-m f f</kbd>).
2. Sometimes the spacebar is written in keybindings with angle brackets (example: <kbd>&lt;SPC&gt; f f</kbd>) and other times without (example: <kbd>SPC f f</kbd>).

cpaulik:
While this should probably be consistent I think it might be not worth it to fix this in markdown in light of https://github.com/syl20bnr/spacemacs/issues/1568

MerelyAPseudonym:
@cpaulik good to know about. Thanks!

syl20bnr:
The export script will surely take care about the conversion so it is OK to fix them in markdown too. I'll make sure to merge the markdown fixes before converting to org :-)

person808:
I believe I have fixed these inconsistencies in #1939

StreakyCobra:
There is only few places left in the documentation containing either `<kbd>` or `<SPC>`. I'll let this issue for beginners.

PR #5549, opened 2016-03-20
Fix inconsistent notation of SPC key in docs

robert-m-johnson:
Replace all occurrences of `<SPC>` in org files with `SPC`.

Fixes syl20bnr/spacemacs#1823

### DIFF
--- a/core/templates/README.org.template
+++ b/core/templates/README.org.template
@@ -24,6 +24,6 @@ To use this contribution add it to your =~/.spacemacs=
 
 * Key bindings
 
-| Key Binding     | Description    |
-|-----------------+----------------|
-| ~<SPC> x x x~   | Does thing01   |
+| Key Binding | Description    |
+|-------------+----------------|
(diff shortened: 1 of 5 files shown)


robbyoconnor:
People will assume that they will need to type the letters `S`, `p`, and then `c` -- this is a concern. it should be: `SPC` -- use do something like this in the org files:

```org
=SPC xxx=
```

StreakyCobra:
@robbyoconnor What are you talking about? `SPC` is the current way of describing nearly all keybindings within Spacemacs, the few `<SPC>` should be corrected. And as far as I know this has never been a problem to understand for users. Also keybindings documentation in Spacemacs use `~` as surrounds, not `=`, so everything is fine on this side too.

After a quick check, this PR seems OK for me, correct PR branch, good commit message, org table aligned, …

robbyoconnor:
@StreakyCobra, maybe I read it fast -- but they didn't do that =foo=

StreakyCobra:

PR #5551, opened 2016-03-21
Fixes #5546 - Set popwin config for Help buffer :noselect t

### LINKED ISSUE #5546: company-show-doc-buffer should not select help buffer
#### Description
company-show-doc-buffer, bound to `f1` and `c-h` in company-active-map, pops up a temporary help buffer to show documentation for current completion. By default, this keeps point in the text buffer, and any further typing or movement dismisses the temp buffer. In spacemacs, focus moves to the new help window, which must then be dismissed manually before continuing typing.

#### Reproduction guide
- Start Emacs with auto-completion layer
- Go to an elisp buffer (`spc fed`)
- type `def`, and when the company candidates popup appears, press `C-h` or `f1`

*Observed behaviour:*

Point moves to help buffer. To continue typing, or selecting candidates, I need to dismiss help buffer with `q`
screen shot 2016-03-20 at 9 00 45 pm

*Expected behaviour:*

Point remains in place, so that continued input is immediate.
screen shot 2016-03-20 at 9 06 52 pm

I initially thought this was due to `(setq help-window-select t)`, but changing this variable has no effect (doesn't fix the issue in spacemacs, nor does it reproduce the issue in vanilla emacs with just company). I think it might have to do with spacemacs' use of winner-mode, but don't know enough about it. Any ideas?

#### System Info
- OS: darwin
- Emacs: 24.5.1
- Spacemacs: 0.105.14
- Spacemacs branch: develop (rev. 744cd0f)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: ivy
- Layers:
```elisp
((auto-completion :variables auto-completion-tab-key-behavior 'cycle auto-completion-return-key-behavior nil auto-completion-enable-sort-by-usage t)
 better-defaults colors clojure dash erc emacs-lisp
 finance html git github markdown org osx react
 syntax-checking spacemacs-ivy themes-megapack vinegar sooheon soo-clojure soo-parens soo-org)
```

sooheon:
Fixes issue #5546. Does not change behavior when calling help directly,
as with describe-function or describe-variable.

### DIFF
--- a/layers/+distribution/spacemacs-base/packages.el
+++ b/layers/+distribution/spacemacs-base/packages.el
@@ -734,7 +734,7 @@ below. Anything else exits."
       (setq popwin:special-display-config nil)
 
       ;; buffers that we manage
-      (push '("*Help*"                 :dedicated t :position bottom :stick t :noselect nil :height 0.4) popwin:special-display-config)
+      (push '("*Help*"                 :dedicated t :position bottom :stick t :noselect t :height 0.4) popwin:special-display-config)
       (push '("*compilation*"          :dedicated t :position bottom :stick t :noselect t   :height 0.4) popwin:special-display-config)
       (push '("*Shell Command Output*" :dedicated t :position bottom :stick t :noselect nil            ) popwin:special-display-config)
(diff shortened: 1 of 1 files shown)


syl20bnr:
Thank you !

PR #5559, opened 2016-03-21
Added quick keybinding for toggle-lisp-state

jgertm:
As already implemented for some lisps, pressing ~, ,~ while editing lisp files switches to lisp-state.
The documentation for each of the layers mentions this capability now.

### DIFF
--- a/layers/+lang/clojure/README.org
+++ b/layers/+lang/clojure/README.org
@@ -37,10 +37,10 @@ This layer adds support for [[http://clojure.org][Clojure]] language using [[htt
 
 * Features
 - REPL via [[https://github.com/clojure-emacs/cider][CIDER]]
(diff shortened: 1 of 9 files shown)


TheBB:
I don't think you should advertise it as `, ,` in the documentation when the major mode leader key may not be set to `,`. Just `SPC m ,` is fine.

The org layer does [this](https://github.com/TheBB/spacemacs/blob/boot-magic-regepx/layers/+emacs/org/packages.el#L223) by the way. Might be worth trying?

StreakyCobra:
@TheBB Broken link.

TheBB:
Sorry, fixed. It pointed to one of my PR branches that I deleted.

jgertm:
These changes have since found their way into develop.

PR #5563, opened 2016-03-22
Fixes for ranger layer to allow dired override

ralesi:
Defer blocked `ranger-override-dired` from operating correctly on initial load.  Added hook to `dired-mode-hook` and made common ranger commands accessible.

### DIFF
--- a/layers/+tools/ranger/packages.el
+++ b/layers/+tools/ranger/packages.el
@@ -15,12 +15,15 @@
 (defun ranger/init-ranger ()
   (use-package ranger
     :defer t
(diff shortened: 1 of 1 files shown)


StreakyCobra:
@ralesi Is the `:commands` really needed as all these functions have the `;;;autoload`? PS: Rebase needed :-)

StreakyCobra:
For the record: This PR will override https://github.com/syl20bnr/spacemacs/pull/5485 (right?)

syl20bnr:
Thank you ! 👍 
I move some code to dired post-init functions since it makes more sense to put it then I believe.

Issue #5567, opened 2016-03-22
Error when creating new layout


ilya-babanov:
#### Description
Hi! 
After recent changes in layouts logic it's impossible (for me) to create new layouts (`SPC-l-number`).
At every attempt to do that I get an error message.

#### Reproduction guide
- Start Emacs
- Try to create new layout (`SPC-l-number`)

*Observed behaviour:*
New layout isn't created. 
Instead you get this error message:
```
apply: Wrong number of arguments: (lambda (_new-persp-name) "Update and save current frame's eyebrowse workspace to its perspective.
Parameter _NEW-PERSP-NAME is ignored, and exists only for compatibility with
`persp-before-switch-functions'." (let* ((current-slot (eyebrowse--get (quote current-slot))) (current-tag (nth 2 (assoc current-slot (eyebrowse--get (quote window-configs)))))) (eyebrowse--update-window-config-element (eyebrowse--current-window-config current-slot current-tag))) (spacemacs/save-eyebrowse-for-perspective)), 2
```

*Expected behaviour:*
No errors, new layout is created.

*My investigation:*
It seems that problem with [`spacemacs/update-eyebrowse-for-perspective` function](https://github.com/syl20bnr/spacemacs/blob/82efa16dab91fcecb5996b91ed582e7ef5bc2855/layers/+spacemacs/spacemacs-layouts/funcs.el#L514). It is used as a hook for `persp-before-switch-functions`. But, according to [this code in persp-mode repository](https://github.com/Bad-ptr/persp-mode.el/blob/a425e09d0a9251971ea4f1fa748da773c3d3a11c/persp-mode.el#L1720-L1745), this hook is always called with two arguments, not one. And indeed - after adding second argument to `spacemacs/update-eyebrowse-for-perspective` everything is OK.

#### System Info
- OS: darwin
- Emacs: 24.5.1
- Spacemacs: 0.105.14
- Spacemacs branch: develop (rev. 82efa16)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(osx spacemacs-helm spacemacs-layouts
     (auto-completion :variables auto-completion-enable-sort-by-usage t)
     version-control git spell-checking
     (syntax-checking :variables syntax-checking-enable-tooltips nil)
     (shell :variables shell-default-term-shell "/usr/local/bin/zsh" shell-default-height 60)
[...]

PR #5568, opened 2016-03-23
layouts: fix update-eyebrowse-for-perspective missing argument

tridao:
Functions in `persp-before-switch-functions` must take two arguments -- a name of a perspective to switch and a frame or a window for which the switching takes place.


bmag:
duplicate of #5229 
see also https://github.com/syl20bnr/spacemacs/issues/5567#issuecomment-200021938

StreakyCobra:
Nice catch @bmag. The older has the priority, closing this one then :-)

Thanks anyway @tridao !

PR #5569, opened 2016-03-23
Add an example hook to evil-cleverparens layer doc

gilch:
Add an example `clojure-mode-hook` and remove misstatement

The layer doesn't add a hook to load with `emacs-lisp-mode`. A hook is required in the dotfile.

### DIFF
--- a/layers/+vim/evil-cleverparens/README.org
+++ b/layers/+vim/evil-cleverparens/README.org
@@ -9,16 +9,16 @@
 
 This simple layer adds the [[https://github.com/luxbock/evil-cleverparens][evil-cleverparens]] package, which overrides common
 normal-mode vim commands like D, dd, c, etc. to keep parentheses balanced. See
-the repository for more details. All the layer does right now is to add a hook
-to load evil-cleverparens with emacs-lisp-mode.
+the repository for more details.
 
 * Install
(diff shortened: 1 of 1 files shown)


bmag:
FWIW, I use this to enable evil-cleverparens:
```elisp
(add-hook 'smartparens-enabled-hook #'spacemacs/toggle-evil-cleverparens-on)
(add-hook 'smartparens-disabled-hook #'spacemacs/toggle-evil-cleverparens-off)
```

gilch:
Updating the doc is a minimal fix, but perhaps not an optimal one. I'm not sure, but if it generally makes sense to enable evil-cleverparens anytime smartparens is active, wouldn't it be better to add those hooks to the layer instead so the user doesn't have to configure it?

bmag:
That's a good suggestion. You're welcome to modify the PR :wink:

gilch:
I think it would be better in a new PR, since it would also require an update of this doc, but completely unrelated to my changes so far. It's not clear to me if that's a good default or not. Is smartparens only used for lisps? Does evil-cleverparens make sense in a non-lisp language? I needs its own discussion, even if you just open an issue. I'll close this PR if your PR gets merged first, since it would be obsolete, but if this gets merged first, no harm done, since you'll have to rewrite the doc anyway.

StreakyCobra:

PR #5571, opened 2016-03-23
add more usage for github gist

daleione:


robbyoconnor:
NO NO NO NO NO, add leader ksys for this functionality :)

layers/+source-control/github/README.org
```diff
@@ -79,6 +79,19 @@ In the gist list buffer:
...
+|-------------+--------------------------------|
+| C-x C-s     | save a new version of the gist |
+| C-x C-w     | rename some file               |
+
+Dired integration. From a dired buffer, you can:
+
+| Key Binding | Description                                                      |
+|-------------+------------------------------------------------------------------|
+| @           | make a gist out of marked files (with a prefix, make it private) |
+
```
robbyoconnor: Can you add some spacemacs keybindings? I don't use emacs editing style

robbyoconnor:
Add spacemacs leader key keybindings -- I don't think this should be merged as-is.

StreakyCobra:
@robbyoconnor Proposing a set of keys and pointers to existing keybindings definition code would be more helpful probably ;-)

robbyoconnor:
@StreakyCobra -- probably - I don't feel like coming up with them -- nor do I care -- I just don't think this should be merged as-is -- unless the namespace doesn't exist (happened for my emberjs layer) -- then it's fine -- but it exists

nixmaniack:
I think there's no added functionality in this PR. Functionality is already there, just documentation needs right keybindings stated. `SPC f s` for `save new version` and `SPC f R` for `rename` ? I haven't tested this, just pointing out.

robbyoconnor:
Why? Just rebase...

nixmaniack:
Docs aren't updated. Let's keep open!

smile13241324:
Sorry guys but this PR is now over 4 years old, we will not be able to merge this into the current code base anymore therefore I will close this. If we still need this, a new PR should be opened following the current conventions and bases to develop.

PR #5578, opened 2016-03-24
cider-turn-on-eldoc-mode is deprecated

tekacs:
It's reported in minibuffer (presumably by cider-mode) that cider-turn-on-eldoc-mode is deprecated and we should use eldoc-mode instead. So a simple update to do so.


tekacs:
Found this whilst investigating why eldoc-mode doesn't activate in my cider despite this hook, but I'll raise an issue for that elsewhere.

StreakyCobra:

Loading fixed in 5f93dc364f3dfdb73d117e1a6dbed8035d10f442.

StreakyCobra:
Reopening for hotfixing

syl20bnr:
Backported in 0.105.19

PR #5584, opened 2016-03-24
optionally enable ensime

channingwalton:
I've introduced a new variable, scala-auto-start-ensime, that determines if ensime starts
when a scala file is loaded.

The variable defaults to t to reflect current behaviour.

Its also been documented.


syl20bnr:
Thank you !

PR #5588, opened 2016-03-25
Disable large file check for certain major-modes

myrjola:
Also introduce a variable `spacemacs-large-file-modes-list` to configure these modes. I based it on a similar list used by vlf-mode https://github.com/m00natic/vlfi/blob/master/vlf-setup.el#L47-L51. Can you come up with any more modes that are supposed to open large files?

My problem is that I have certain large PDFs and images that I open in Emacs and that extra prompt by `spacemacs/check-large-file` is annoying in those cases.

### DIFF
--- a/layers/+distribution/spacemacs-base/funcs.el
+++ b/layers/+distribution/spacemacs-base/funcs.el
@@ -138,6 +138,16 @@ the current state and point position."
   :type 'number
   :group 'spacemacs)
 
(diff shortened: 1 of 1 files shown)


nixmaniack:
I have set `dotspacemacs-large-file-size` quite large for exactly same reason, opening large sized pdf files became annoying.

TheBB:

PR #5597, opened 2016-03-26
Fixes #5550 - Fix dired-x dired-jump autoload

### LINKED ISSUE #5550: Missing autoload on `dired-jump`
`SPC f j` `dired-jump` is missing an autoload.  `dired` must be loaded some other way before that binding/command will work.

On 744cd0f

darkfeline:
Properly add `dired-x` `dired-jump` autoload.

### DIFF
--- a/layers/+vim/vinegar/keybindings.el
+++ b/layers/+vim/vinegar/keybindings.el
@@ -1,4 +1,3 @@
-(require 'dired-x)
 (define-key evil-normal-state-map (kbd "-") 'dired-jump)
 
 (add-hook 'dired-mode-hook 'vinegar/dired-setup)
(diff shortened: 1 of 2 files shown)


darkfeline:
Anyone with commit access I can bug for this and https://github.com/syl20bnr/spacemacs/pull/5598?  These are pretty small fixes that have been pending for a week.

StreakyCobra:

layers/+vim/vinegar/keybindings.el
```diff
@@ -1,4 +1,3 @@
-(require 'dired-x)
```
caleb: I'm using the `develop` branch and if I have vinegar enabled I get an error that `dired-omit-mode` is void with this line deleted.

If I add the line back in everything works as expected, so it seems as though the `dired-x` module isn't being loaded by the time the `vinegar` layer is loaded.

PR #5598, opened 2016-03-26
Set SPC f b to bookmark-jump

darkfeline:
helm will overwrite this with its own command, but people not using
helm (only ivy at the moment) will need this defined.


StreakyCobra:

PR #5601, opened 2016-03-26
Make helm-persp-switch-project inherit helm actions

### LINKED ISSUE #5596: helm-persp-switch-project doesn't have the same actions as helm-projectile-switch-project
It would be nice if while in helm-persp-switch-project you could switch to eshell in the selected project's root directory with M-e or magit-status with M-g like you can with home-projectile-switch project.

honkhonkhonk:
Make helm-persp-switch-project have the same helm actions that helm-projectile-switch-project has. Fixes #5596

### DIFF
--- a/layers/+spacemacs/spacemacs-layouts/funcs.el
+++ b/layers/+spacemacs/spacemacs-layouts/funcs.el
@@ -315,6 +315,13 @@ perspectives does."
                    (mapcar 'persp-kill
                            (helm-marked-candidates))))))))
 
+
+(defun spacemacs//helm-persp-switch-project-helper (project func)
+  (let ((persp-reset-windows-on-nil-window-conf t))
+    (persp-switch project)
+    (funcall func project)
+    (delete-other-windows)))
+
 (defun spacemacs/helm-persp-switch-project (arg)
   (interactive "P")
   (helm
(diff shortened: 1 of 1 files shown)


layers/+distribution/spacemacs-base/packages.el
```diff
@@ -815,6 +815,7 @@ below. Anything else exits."
     :config
     (progn
       (projectile-global-mode)
+      (helm-projectile-on)
```
bmag: Is `helm-projectile` available if the user uses `spacemacs-ivy` layer instead of `spacemacs-helm`? Should probably move this to `helm-projectile`'s configuration in `layers/+completion/spacemacs-helm/packages.el`.

robbyoconnor: It's not. This will fail for those users. Helm is not always available.

honkhonkhonk:
oops. the function was being bound in init-projectile on master, but isn't in develop.

honkhonkhonk:
Okay, so currently it works when helm-projectile is loaded manually. I'm not sure of the best way to lazy load it.

honkhonkhonk:
For some reason adding this function to the :commands keyword isn't lazy loading helm-projectile for me. 

It works fine if i put (require 'helm-projectile) at the top of this function (helm-persp-switch-project), however this seems to be bad practice. 

Does anyone have any idea how I can get this working? I feel like I don't know what I'm doing..

Miciah:
@honkhonkhonk, what error are you seeing, and what change did you make to try to resolve the error?

Using your modified `spacemacs/helm-persp-switch-project`, I see the following error when I press `SPC p l` in a new Spacemacs session:

    let: Symbol’s function definition is void: helm-projectile-define-key

However, when I made the following modification and restarted Spacemacs, `SPC p l` worked with no problems:

```patch
diff --git a/layers/+completion/helm/packages.el b/layers/+completion/helm/packages.el
index 227bd4795..2feb28095 100644
--- a/layers/+completion/helm/packages.el
+++ b/layers/+completion/helm/packages.el
@@ -311,7 +311,8 @@
                helm-projectile-find-file
                helm-projectile-grep
                helm-projectile
-               helm-projectile-switch-project)
+               helm-projectile-switch-project
+               helm-projectile-define-key)
     :init
     (progn
       ;; needed for smart search if user's default tool is grep
```

layers/+spacemacs/spacemacs-layouts/funcs.el
```diff
@@ -315,6 +315,13 @@ perspectives does."
                    (mapcar 'persp-kill
                            (helm-marked-candidates))))))))
 
+
+(defun spacemacs//helm-persp-switch-project-helper (project func)
+  (let ((persp-reset-windows-on-nil-window-conf t))
+    (persp-switch project)
```
Miciah: Would it make sense to create a new workspace in the layout, if the layout already exists? I mean something like the following:
```diff
                            (helm-marked-candidates))))))))

 (defun spacemacs//helm-persp-switch-project-helper (project func)
-  (let ((persp-reset-windows-on-nil-window-conf t))
+  (let ((persp-reset-windows-on-nil-window-conf t)
+        (persp-already-exists (persp-p (persp-get-by-name project))))
     (persp-switch project)
+    (when persp-already-exists
+      (eyebrowse-create-window-config))
     (funcall func project)
     (delete-other-windows)))
```
This would avoid clobbering the existing workspace.  On the other hand, `SPC w u` should also undo said clobbering, and maybe that is good enough.

PR #5605, opened 2016-03-26
Setup flycheck-purescript for purescript layer

diogob:
I've added the package [flycheck-purescript](http://melpa.org/#/flycheck-purescript) and used the configuration from haskell layer as a template.


syl20bnr:
Thank you ! :+1: 

PR #5613, opened 2016-03-27
Improve Leiningen installation instructions in Clojure layer

### LINKED ISSUE #5612: Clojure layer nREPL cannot jack in (dependency versions)
I tried to start using ClojureScipt yesterday, and got stuck when I tried to launch the REPL from inside Spacemacs. I've tracked down the issue to stem from an incompatible version of Leiningen. It's very difficult to deduce this from the error that Spacemacs produces, and it's very likely to happen if you follow the installation instructions.

The installation instructions should explicitly mention the Leiningen version required instead of just recommending that the user install it from the OS package manager, because Ubuntu 14.X doesn't seem to package a recent enough version for it to work.

porglezomp:
The previous installation instructions suggested that you install lein via your OS package manager, but since the minimum required version to run the plugins is 2.5.2 and some OS package managers are still packaging versions in 1.x, linking to the official install instructions is better.

Fixes #5612

### DIFF
--- a/layers/+lang/clojure/README.org
+++ b/layers/+lang/clojure/README.org
@@ -83,7 +83,7 @@ Or set this variable when loading the configuration layer:
 - Run ~SPC m s i~ in any of the clojure source files to connect to the CIDER REPL.
 
 *** Quick Start with lein
-- Install =lein= via your OS package manager.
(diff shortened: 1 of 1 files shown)


syl20bnr:
Thank you !

PR #5620, opened 2016-03-28
add an option to use google-c-style in c-c++ mode

eklitzke:
This adds an option (which is disabled by default) to add `google-set-c-style` to the `c-common-mode-hooks`.

Again, this is disabled by default: it's only turned on when a user sets `'use-google` to true when enabling `c-c++` mode.

I've also updated the documentation to describe the new feature and how to enable it.

### DIFF
--- a/layers/+lang/c-c++/README.org
+++ b/layers/+lang/c-c++/README.org
@@ -12,6 +12,7 @@
    - [[#enable-clang-support][Enable Clang support]]
      - [[#clang-format][clang-format]]
      - [[#company-clang-and-flycheck][Company-clang and flycheck]]
+   - [[#enable-google-set-c-style][Enable google-set-c-style]]
  - [[#key-bindings][Key Bindings]]
 
 * Description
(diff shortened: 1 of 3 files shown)


StreakyCobra:
I think you are missing the `defvar` in `config.el`. Also try to use the naming scheme as other variables (e.g. `c-c++-enable-clang-support`)

eklitzke:
Thanks for your feedback, I will make those changes and then update my PR.

eklitzke:
I made both of these changes (and squashed my branch)

StreakyCobra:
:+1:

layers/+lang/c-c++/README.org
```diff
@@ -95,6 +96,27 @@ Not only does this allow proper autocomplete on projects with extra
 includes and flags, but there is also support for flycheck so that it
 doesn't complain about missing header files.
 
+** Enable google-set-c-style
+If you have clang enabled with =clang-format= as described earlier in this page
+you may not have a lot of neeed for =google-set-c-style= if you are already
+using a mode based on Google mode for most of your projects.
+
+However, if you don't have (or want) =clang-format=, or if you have to do a lot
+if you have to do a lot of [[https://www.emacswiki.org/emacs/TrampMode][Tramp]] remote editing on systems that don't have
```
emacs18: Following is duplicated: "if you have to do a lot".

layers/+lang/c-c++/README.org
```diff
@@ -95,6 +96,27 @@ Not only does this allow proper autocomplete on projects with extra
...
+=clang-format= installed, you may way =google-c-style= enabled and added to your
```
emacs18: Perhaps you meant "want" rather than "way" here.

eklitzke:
@emacs18 thanks for the suggestions, I made both changes in the current diff.

eklitzke:
I rebased this to (hopefully) fix my merge conflicts.

eklitzke:
I extended this so you can also set `c-c++-enable-google-newline` and get the google newline behavior (if it off by default)

syl20bnr:
Thank you ! 👍 

PR #5624, opened 2016-03-29
Gtags expand and cleanup

choppsv1:
- Add gtags support to many languages.
- Add documentation on using exuberant ctags and python pygments to enable this support in global.
- Cleanup and standardize how ggtags and helm-gtags is enabled. Theirs no need to check for the layer enabled as post-init-* won't be run unless that package has been enabled.

- This is a first round of changes. It's also worth looking into enabling eldoc and company in as many modes as possible who might otherwise not support those features.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -993,11 +993,11 @@ path."
 
 (defun configuration-layer//lazy-install-packages (layer-name mode)
   "Install layer with LAYER-NAME to support MODE."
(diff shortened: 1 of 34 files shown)


robbyoconnor:
May wanna squash when you're done :heart:

choppsv1:
Going to redo this so that it's one commit, couldn't manage to squash the commit b/c it had acquired a merge at some point.

PR #5625, opened 2016-03-29
expand gtags use to many more languages

choppsv1:
- Add gtags support to many languages.
- Add documentation on using exuberant ctags and python pygments to enable this support in global.
- Cleanup and standardize how ggtags and helm-gtags is enabled. There is no need to check for the layer enabled as `post-init-*` won't be run unless that package has been enabled.

This is a first round of changes. It's also worth looking into enabling eldoc and company in as many modes as possible who might otherwise not support those features.

### DIFF
--- a/layers/+lang/asm/packages.el
+++ b/layers/+lang/asm/packages.el
@@ -13,6 +13,8 @@
       '(
         ;; package names go here
         asm-mode
+        ggtags
+        helm-gtags
         nasm-mode
         x86-lookup
         ))
@@ -56,3 +58,9 @@
 (defun asm/post-init-company ()
   (spacemacs|add-company-hook asm-mode)
   (spacemacs|add-company-hook nasm-mode))
+
+(defun asm/post-init-ggtags ()
+  (add-hook 'asm-mode-hook #'spacemacs/ggtags-mode-enable))
+
+(defun asm/post-init-helm-gtags ()
+  (spacemacs/helm-gtags-define-keys-for-mode 'asm-mode))
--- a/layers/+tags/gtags/README.org
+++ b/layers/+tags/gtags/README.org
@@ -1,124 +1,252 @@
 #+TITLE: Helm Gtags layer
 
-* Table of Contents                                         :TOC_4_gh:noexport:
- - [[#description][Description]]
- - [[#features][Features]]
- - [[#install][Install]]
-   - [[#gnu-global][GNU Global]]
- - [[#usage][Usage]]
-   - [[#eldoc-integration][Eldoc integration]]
- - [[#key-bindings][Key bindings]]
+* Table of Contents                                         :TOC_4_org:noexport:
+ - [[Description][Description]]
+ - [[Features][Features]]
+ - [[Install][Install]]
+   - [[GNU Global (gtags)][GNU Global (gtags)]]
+     - [[Install on OSX using Homebrew][Install on OSX using Homebrew]]
+     - [[Install on *nix from source][Install on *nix from source]]
+       - [[Install recommended dependencies][Install recommended dependencies]]
+       - [[Install with recommended features][Install with recommended features]]
+       - [[Configure your environment to use pygments and ctags][Configure your environment to use pygments and ctags]]
+   - [[Emacs Configuration][Emacs Configuration]]
+ - [[Usage][Usage]]
+     - [[Language Support][Language Support]]
+       - [[Built-in languages][Built-in languages]]
+       - [[Exuberant ctags languages][Exuberant ctags languages]]
+       - [[Universal ctags languages][Universal ctags languages]]
+       - [[Pygments languages (plus symbol and reference tags)][Pygments languages (plus symbol and reference tags)]]
+   - [[Eldoc integration][Eldoc integration]]
+ - [[Key bindings][Key bindings]]
 
 * Description
-=helm-gtags= and =ggtags= are clients for GNU Global. GNU Global is a source
-code tagging system that allows querying symbol locations in source code, such
-as definitions or references.
+
+  =helm-gtags= and =ggtags= are clients for GNU Global. GNU Global is a source
+  code tagging system that allows querying symbol locations in source code, such
+  as definitions or references. Adding the =gtags= layer enables both of these modes.
 
 * Features
 
-- Select any tag in a project retrieves by gtags
-- Resume previous helm-gtags session
-- Jump to a location based on context
-- Find definitions
-- Find references
-- Present tags in current function only
-- Create a tag database
-- Jump to definitions in file
-- Show stack of visited locations
-- Manually update tag database
-- Jump to next location in context stack
-- Jump to previous location in context stack
-- Jump to a file in tag database
+  - Select any tag in a project retrieves by gtags
+  - Resume previous helm-gtags session
+  - Jump to a location based on context
+  - Find definitions
+  - Find references
+  - Present tags in current function only
+  - Create a tag database
+  - Jump to definitions in file
+  - Show stack of visited locations
+  - Manually update tag database
+  - Jump to next location in context stack
+  - Jump to previous location in context stack
+  - Jump to a file in tag database
+  - Enables =eldoc= in modes that otherwise might not support it.
+  - Enables =company complete= in modes that otherwise might not support it.
 
 * Install
-** GNU Global
 
-You can install =helm-gtags= from the software repository of your OS. For example, in Ubuntu:
+** GNU Global (gtags)
 
-#+BEGIN_SRC sh
-  sudo apt-get install global
-#+END_SRC
+   To use =gtags=, you first have to install [[https://www.gnu.org/software/global/download.html][GNU Global]].
 
-To use =helm-gtags=, you first have to install [[https://www.gnu.org/software/global/download.html][GNU Global]].
+   You can install =global= from the software repository of your OS; however,
+   many OS distributions are out of date, and you will probably be missing
+   support for =pygments= and =exuberant ctags=, and thus support for many
+   languages. We recommend installing from source. If not for example to install
+   on Ubuntu:
 
-Download the latest tar.gz archive, then run these commands:
+   #+begin_src sh
+     sudo apt-get install global
+   #+end_src
 
-#+BEGIN_SRC sh
-  tar xvf global-6.4.tar.gz
-  cd global-6.4
-  ./configure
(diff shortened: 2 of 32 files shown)


layers/+tags/gtags/packages.el
```diff
@@ -11,13 +12,24 @@
...
   (use-package ggtags
-    :defer t))
+    :defer t
+    :init
+    (progn
+      ;; modes that do not have a layer, add here.
+      (add-hook 'awk-mode-hook (lambda () (ggtags-mode 1)))
+      (add-hook 'shell-mode-hook (lambda () (ggtags-mode 1)))
+      (add-hook 'tcl-mode-hook (lambda () (ggtags-mode 1)))
+      (add-hook 'vhdl-mode-hook (lambda () (ggtags-mode 1)))
```
joehillen: I believe it's preferred to use a named function when adding hooks rather than a lambda for debugging purposes.

layers/+tags/gtags/README.org
```diff
@@ -5,121 +5,249 @@
...
+    your normal OS package manager, e.g., on Ubuntu
+
+    #+BEGIN_SRC sh
+      sudo apt-get install exuberant-ctags python-pygments
+    #+END_SRC
+
+    or e.g., Archlinux:
+
+    #+BEGIN_SRC sh
+      sudo pacman -S ctags python-pygments
```
joehillen: global is also available in the [AUR](https://aur.archlinux.org/packages/global/). Is that needed as well?

layers/+tags/gtags/README.org
```diff
@@ -5,121 +5,249 @@
...
+    #+END_SRC
 
-- From within Emacs, runs the command =helm-gtags-create-tags=, which is bound
-  to ~SPC m g c~. If the language is not directly supported by GNU Global, you
-  can choose =ctags= or =pygment= as a backend to generate tag database.
+**** Configure your environment to use pygments and ctags
 
-- From inside terminal, runs gtags at your project root in terminal:
+    To be able to use =pygments= and =ctags=, you need to copy the sample
+    =gtags.conf= either to =/etc/gtags.conf= or =$HOME/.globalrc=. For example:
```
joehillen: I had to extract gtags.conf from the global source tarball. Would it be better to distribute it with the layer itself and have it in the spacemacs source tree?

layers/+tags/gtags/README.org
choppsv1: Joe Hillenbrand <notifications@github.com> writes:
...
> global is also available in the
> [AUR](https://aur.archlinux.org/packages/global/). Is that needed as
> well?

Well the text is describing installing the requirements before
compiling global form source. I thought about mentioning AUR as well,
but since there's no standard command for installing from AUR it's not
really "provided by the OS package system" like yum or apt-get is. Plus,
most everyone running archlinux probably knows to check AUR anyway :)

Thanks,
Chris.

layers/+tags/gtags/README.org
choppsv1: Joe Hillenbrand <notifications@github.com> writes:

>>
>> -- From inside terminal, runs gtags at your project root in terminal:
>> +    To be able to use =pygments= and =ctags=, you need to copy the sample
>> +    =gtags.conf= either to =/etc/gtags.conf= or =$HOME/.globalrc=. For example:
>
> I had to extract gtags.conf from the global source tarball. Would it
> be better to distribute it with the layer itself and have it in the
> spacemacs source tree?

I don't think so. This file changes (and expands) from release to
various release of global. In another comment you mentioned AUR, at
least for me the AUR package installed gtags.conf in
/usr/share/gtags/gtags.conf. In fact my .globalrc is simply a symlink to
this file on my archlinux machines. The file should be installed with
any modern version of global.

Thanks,
Chris.

layers/+tags/gtags/packages.el
joehillen: Should it also be a toggle?

choppsv1: Joe Hillenbrand <notifications@github.com> writes:
>> +      (add-hook 'vhdl-mode-hook (lambda () (ggtags-mode 1)))
>
> I believe it's preferred to use a named function when adding hooks rather than a lambda for debugging purposes.

I've heard this about hooks as well; however, I don't believe this is a
spacemacs design requirement, and I don't think its meant to apply to
the simple case of a single function call.

Thanks,
Chris.

choppsv1: Joe Hillenbrand <notifications@github.com> writes:
>> +      (add-hook 'vhdl-mode-hook (lambda () (ggtags-mode 1)))
>
> Should it also be a toggle?

What circumstance are you imaging one would wish to toggle gtags mode on
and off?

Thanks,
Chris.

joehillen: > I don't think its meant to apply to the simple case of a single function call.

I've seen it requested in other reviews. @TheBB would know for sure.

layers/+tags/gtags/README.org
joehillen: > In another comment you mentioned AUR, at least for me the AUR package installed gtags.conf in /usr/share/gtags/gtags.conf. In fact my .globalrc is simply a symlink to this file on my archlinux machines. 

That's good to know. It might be worth mentioning in the docs here.

layers/+tags/gtags/packages.el
bmag: It should be `(add-hook <mode-hook> #'ggtags-mode)`.
Calling `ggtags-mode` without an argument turns it on (when called from elisp code, like in this case).

layers/+tags/gtags/README.org
bmag: I'd like to add that the version in Ubuntu's repository is old - 5.7, while the newest one is already 6.5, so it's probably better to build it from source on Ubuntu.

layers/+lang/asm/packages.el
```diff
@@ -56,3 +58,9 @@
 (defun asm/post-init-company ()
   (spacemacs|add-company-hook asm-mode)
   (spacemacs|add-company-hook nasm-mode))
+
+(defun asm/post-init-ggtags ()
+  (add-hook 'asm-mode-hook '(lambda () (ggtags-mode 1))))
```
bmag: Why do you not use `spacemacs/ggtags-enable-eldoc` here instead of `add-hook`?
Same goes for most other places where `(add-hook ... ggtags-mode)` is used.

layers/+tags/gtags/README.org
choppsv1: Bar <notifications@github.com> writes:
>> -* Usage
>> +    To take full advantage of global you should install 2 extra packages in
>> +    addition to global: pygments and ctags (exuberant). You can do this using
>> +    your normal OS package manager, e.g., on Ubuntu
>
> I'd like to add that the version in Ubuntu's repository is old - 5.7,
> while the newest one is already 6.5, so it's probably better to build
> it from source on Ubuntu.

If you look up to the start of this section I say:

   You can install =global= from the software repository of your OS; however,
   many OS distributions are out of date, and you will probably be missing
   support for =pygments= and =exuberant ctags=, and thus support for many
   languages. We recommend installing from source.

Which I think makes this point.

Thanks,
Chris.

layers/+lang/asm/packages.el
bmag: Or is it because you want to enable ggtags-mode but not use it to provide information to Eldoc? (which makes sense sometimes)

choppsv1: Bar <notifications@github.com> writes:
> Why do you not use `spacemacs/ggtags-enable-eldoc` here instead of `add-hook`?
> Same goes for most other places where `(add-hook ... ggtags-mode)` is used.

I believe in fact that I'm going to end up removing
spacemacs/ggtags-enable-eldoc as what its doing is no longer needed
(24.4+). The correct thing to do is enable the mode.

layers/+tags/gtags/packages.el
choppsv1: Bar <notifications@github.com> writes:
>> +      (add-hook 'vhdl-mode-hook (lambda () (ggtags-mode 1)))
>
> It should be `(add-hook <mode-hook> #'ggtags-mode)`.
> Calling `ggtags-mode` without an argument turns it on (when called from elisp code, like in this case).

I didn't know this! I will make the change, thanks!

Chris.

al3xandru:
Newbie question here: if this PR is against develop, shouldn't some ivy integration be presented as an alternative to helm?

choppsv1:
Alex Popescu <notifications@github.com> writes:

> Newbie question here: if this PR is against develop, shouldn't some ivy integration be presented as an alternative to helm?

Helm-gtags is a pre-existing package and support was already present. If
there's a ivy-gtags I would be open to adding support to it in another
feature PR. That said, I don't get what's so useful about helm-gtags,
ggtags-mode is plenty powerful.

Thanks,
Chris.

al3xandru:
@choppsv1 my question was mainly based on the fact that I've noticed the addition of the `helm-tags` dependency to a lot of language layers and the fact that develop will provide ivy as an alternative solution to helm. If helm-gtags is indeed optional, then it should be required iif helm is enabled.

choppsv1:
@al3xandru helm-gtags is optional. The inclusion in the packages array does not in and of itself cause the package to be loaded. Neither does the existence of the <layer>/post-init-<package> function, which is only called when a package is actually loaded after it's init function is called.

syl20bnr:
This is a big PR, I'll check it as soon as possible given its popularity. Thank you for your patience.

choppsv1:
@syl20bnr Should I do another update to fix the conflicts (i.e., do we think this is going to go in before any more change that would break it again)?

wamaral:
Needs more love :)

syl20bnr:
@choppsv1 You can let ut as is, I'll handle the conflicts. This PR is next on my merging list, just need enough time to correctly merge it.

dfischer:
Oh wow this looks like exactly what I need – hopefully merging soon 👍  <3

JohnLunzer:
Also looking forward to this, can't wait.

TheBB:
They say good things are worth waiting for! Thank you @choppsv1. I rebased and pushed, and applied some [fixes](https://github.com/syl20bnr/spacemacs/commit/9773ad611f6f03adc9d1fe80072537015861ac8f) of my own (mostly indentation in the readme).

PR #5628, opened 2016-03-29
Store mc-lists.el in .cache

voidlily:
This file is created by multiple-cursors and saves user preferences.

Not moving this file prevents updating spacemacs due to a dirty work
tree.

This change moves the file to .cache so updates to spacemacs work
without having to do anything.

See https://github.com/magnars/multiple-cursors.el#unknown-commands for
more information

### DIFF
--- a/layers/+spacemacs/spacemacs-evil/packages.el
+++ b/layers/+spacemacs/spacemacs-evil/packages.el
@@ -90,7 +90,9 @@
 
 (defun spacemacs-evil/init-evil-mc ()
   (use-package evil-mc
-    :defer t))
+    :defer t
+    :init
+    (setq mc/list-file (concat spacemacs-cache-directory "mc-lists.el"))))
 
 ;; other commenting functions in funcs.el with keybinds in keybindings.el
 (defun spacemacs-evil/init-evil-nerd-commenter ()


StreakyCobra:
Instead of adding this file to the `.gitignore`, it would be better to change the `mc/list-file` variable in order to move the file to `.cache/`.

voidlily:
Thanks, that sounds much better

voidlily:
(Wish it were easier to rename branches without opening a new PR. If you'd rather me rename the branch and open a new PR for semantics, please let me know)

StreakyCobra:
> If you'd rather me rename the branch and open a new PR for semantics, please let me know

We don't care at all about branch names :smiley:

layers/+spacemacs/spacemacs-evil/packages.el
```diff
@@ -90,7 +90,10 @@
 
 (defun spacemacs-evil/init-evil-mc ()
   (use-package evil-mc
-    :defer t))
+    :defer t
+    :init
+    (setq mc/list-file (concat spacemacs-cache-directory "mc-lists.el"))
+    ))
```
StreakyCobra: We usually group all closing parenthesis on the last line of the block :-)

voidlily: My bad, my personal style puts closing parens on their own line for cleaner diffs :)

nixmaniack:
The last time I checked `evil-mc` didn't use `multiple-cursors` so `mc/list-file` shouldn't be configured in `evil-mc` I think.

I do use `multiple-cursors` and I had this variable configured in my .spacemacs.

voidlily:
You know what, I think you're right. I probably had that file left around from what I tried `multiple-cursors` but switched to `evil-mc`

PR #5630, opened 2016-03-29
Fixes #5629 - Do not delete window when there is only one

### LINKED ISSUE #5629: Closing "ansi-term" buffers when sole window results in an error
#### Description

When closing ansi-term when it takes up the full window, and error is raised (and it makes the closing take a few seconds instead of being instantaneous).

#### Reproduction guide
- Start Emacs
- Have the shell layer loaded. On my system, this looks like:
```
     (shell :variables
            shell-default-height 30
            shell-default-position 'bottom)
```

*Observed behaviour:*
I frequently use ansi-term as part of my workflow. When I close a terminal, say with C-d, emacs freezes for a second or so and reports an error.

*Expected behaviour:*
I expect it to close immediately.

#### System Info
- OS: darwin
- Emacs: 24.5.1
- Spacemacs: 0.105.14
- Spacemacs branch: master (rev. a298a5a)
- Graphic display: t
- Distribution: spacemacs
- Editing style: emacs
- Completion: helm
- Layers:
```elisp
(emacs-lisp git markdown org
            version-control rust elm racket elixir osx haskell)
```

#### Backtrace
```
```

### Patch

I have a patch that fixes this problem locally. I'll make a PR for it immediately after this.

joelmccracken:
If ansi-term is a full-screen window, running `(delete-window)` in this
case would cause an error to appear. This commit guards against this
scenario.

See https://github.com/syl20bnr/spacemacs/issues/5629

### DIFF
--- a/layers/+tools/shell/packages.el
+++ b/layers/+tools/shell/packages.el
@@ -311,7 +311,8 @@ is achieved by adding the relevant text properties."
                                   (when (string-match "\\(finished\\|exited\\)"
                                                       change)
                                     (kill-buffer (process-buffer proc))
-                                    (delete-window))))))
+                                    (when (> (count-windows) 1)
+                                      (delete-window)))))))
       (add-hook 'term-mode-hook 'ansi-term-handle-close)
       (add-hook 'term-mode-hook (lambda () (linum-mode -1)))
 


syl20bnr:
Thank you ! :+1: 

PR #5654, opened 2016-04-01
enable c++11 mode in company-clang-arguments

eklitzke:
This adds a new variable, `c-c++-enable-c++11` which follows the existing `c-c++` variable naming conventions.

When the value is not nil (it is nil by default) then `company-clang-arguments` are set up to use `-std=c++11`

### DIFF
--- a/layers/+lang/c-c++/config.el
+++ b/layers/+lang/c-c++/config.el
@@ -14,6 +14,9 @@
 (defvar c-c++-enable-clang-support nil
   "If non nil Clang related packages and configuration are enabled.")
 
+(defvar c-c++-enable-c++11 nil
+  "If non nil then c++11 related features will be enabled")
+
 (spacemacs|defvar-company-backends c-mode-common)
 (spacemacs|defvar-company-backends cmake-mode)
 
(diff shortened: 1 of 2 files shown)


eklitzke:
I added another change like

```elisp
(when c-c++-enable-c++11
      (setq flycheck-clang-language-standard "c++11"))
```

This should help people using flycheck and getting spelling errors about words like `nullptr`

syl20bnr:
The variable is good but I would like the variable to be overridable at the project level via directory and file variable.
So we can find some hook to be able to read the variable `c-c++-enable-c++11` and set the flags accordingly.

eklitzke:
Let me think about this idea some more and then update the diff.

Beta1440:
Why not have a `c-c++-standard` variable that flycheck can read from instead of `c-c++-enable-c++11`? This avoids needing a `c-c++-enable-c++22` in the future.

syl20bnr:
Thank you again (and still one year and a half late :-)) !

@beta1440 I'm refactoring the `c-c++` layer, I'll take your remark into account.

PR #5656, opened 2016-04-02
Scala: fix ensime-typecheck-current-file to ensime-typecheck-current-buffer

hujianxin:
Now, ensime-emacs dose not have `ensime-typecheck-current-file` function. So, keybinding `SPC-m-c-t` should be binded to `ensime-typecheck-current-buffer`


syl20bnr:
You repo is down but I was able to do the change myself, see commit https://github.com/syl20bnr/spacemacs/commit/99d9993

Thank you !

PR #5664, opened 2016-04-03
python: explicitly autoload py-yapf-buffer

nixmaniack:
e511a1211cac48f03513669b35005c19e52f5713 added lazy-load for py-yapf which caused `py-yapf-buffer` to be undefined. 

Since it's a local package, there's no autoloads file. Adding it to `:commands` to let use-package add autoload for it.


syl20bnr:
Thank you ! :+1:

PR #5672, opened 2016-04-04
Revamp org-repo-todo

TheBB:
* Remove `SPC m g t` binding (it's not specific to emacs lisp mode)
* Add bindings for showing the todo lists of (a) the project, (b) all projects, and (c) all projects + the regular TODOs from `org-agenda-files`. Use `projectile-known-projects` for this.

Note: `org-repo-todo` does *not* use projectile to decide where to put the todo file.

I didn't put much thought into the actual bindings, so feel free to criticise.

I also changed the capture template because I didn't like it, but that's subjective of course, so feel free to criticise that too.

CC @CHedgren

### DIFF
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -498,14 +498,21 @@ Headline^^            Visit entry^^               Filter^^                    Da
 
 (defun org/init-org-repo-todo ()
   (use-package org-repo-todo
-    :defer t
+    :commands (ort/todo-root ort/find-root ort/todo-file)
     :init
     (progn
       (spacemacs/set-leader-keys
-        "Ct"  'ort/capture-todo
-        "CT"  'ort/capture-checkitem)
-      (spacemacs/set-leader-keys-for-major-mode 'org-mode
-        "gt" 'ort/goto-todos))))
+        "Ct" 'ort/capture-todo
+        "CT" 'ort/capture-checkitem
+        "aop" 'ort/list-project-todos)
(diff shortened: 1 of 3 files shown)


TheBB:
If we want a more robust solution that depends on (and integrates with) projectile from the bottom up, I can do that too.

CHedgren:
Yeah, using projectile is, i think, totally the way to go, please do that :)
This brief exchange has been really inspiring wrt. elisp.

layers/+emacs/org/packages.el
```diff
@@ -498,14 +498,21 @@ Headline^^            Visit entry^^               Filter^^                    Da
...
     (progn
       (spacemacs/set-leader-keys
-        "Ct"  'ort/capture-todo
-        "CT"  'ort/capture-checkitem)
-      (spacemacs/set-leader-keys-for-major-mode 'org-mode
-        "gt" 'ort/goto-todos))))
+        "Ct" 'ort/capture-todo
+        "CT" 'ort/capture-checkitem
+        "aop" 'ort/list-project-todos)
+      (when (configuration-layer/package-usedp 'projectile)
```
syl20bnr: If projectile is a mandatory dependency for `org-repo-todo` then `when` can wrap the whole `defun`.

TheBB: It's not a dependency, just a dependency for those additional functions I wrote.

syl20bnr:
In the box !

PR #5673, opened 2016-04-04
Fixes #5273 - core: make retaining visual mode on shift optional

### LINKED ISSUE #5273: Remapping `>` to `>gv` in visual mode breaks repeat (and deviates from vim)
#### Description
Spacemacs remaps`>` to `>gv` in visual mode  (`layers/+distribution/spacemacs-base/packages.el:286-287`). This

1. Breaks repeat (`v>.` doesn't indent twice).
2. Significantly deviates from vim and isn't documented anywhere.

#### Reproduction guide
- Start Emacs (evil mode)
- Type some text in the scratch buffer.
- From normal mode, enter `v>.` (enter visual mode, shift right, repeat).

*Observed behaviour:*
It claims that `.` is undefined. And only indents once.

*Expected behaviour:*
I expected it to indent twice.

#### System Info
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.11
- Spacemacs branch: master (rev. 93e55bb)
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(better-defaults git
                 (auto-completion :variables auto-completion-return-key-behavior nil)
                 org spell-checking syntax-checking
                 emacs-lisp markdown yaml c-c   html haskell go javascript python
                 (rust :variables rust-format-on-save t)
                 (latex :variables latex-build-command "LaTeX"))
```

Stebalien:
This breaks repeated indentations (>.) in visual mode and deviates from vim.

Fixes #5273

### DIFF
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -159,6 +159,9 @@ size to make separators look not too crappy.")
 (defvar dotspacemacs-remap-Y-to-y$ nil
   "If non nil `Y' is remapped to `y$' in Evil states.")
 
+(defvar dotspacemacs-retain-visual-mode-on-shift nil
+  "If non nil `>' is remapped to `>gv' and `<' is remapped to `<gv' in visual mode.")
+
 (defvar dotspacemacs-ex-substitute-global nil
   "If non nil, inverse the meaning of `g' in `:substitute' Evil ex-command.")
 
--- a/layers/+distribution/spacemacs-bootstrap/packages.el
+++ b/layers/+distribution/spacemacs-bootstrap/packages.el
@@ -108,8 +108,9 @@
   (add-hook 'after-change-major-mode-hook 'spacemacs//set-evil-shift-width 'append)
 
   ;; Keep the region active when shifting
-  (evil-map visual "<" "<gv")
-  (evil-map visual ">" ">gv")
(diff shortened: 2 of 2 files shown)


syl20bnr:
Thank you for the PR. I moved the issue to `won't fix` you can see the motivation on the issue thread.

syl20bnr:
Actually I'm open to a dotfile variable with the raw Vim configuration as default, see issue thread.

Stebalien:
@syl20bnr done, I think. Is this what you were looking for?

syl20bnr:
Yep thank you :-)

TheBB:

In https://github.com/syl20bnr/spacemacs/commit/5b719df5002c977eb50d22696cbad67e2a066e7d I:

* Renamed to use `state` instead of `mode` as this is the preferred evil term
* Changed the default value to `t` (can be discussed obviously)
* Added the variable to the dotfile template (!)
* Fixed your code (!!) ;-)

Stebalien:
Wow. I really dropped the ball here.

So, for some reason I assumed that the spacemacs template was auto-generated using some form of reflection and never bothered to check.

As for the bugs... I really should stop assuming something is so simple I don't need to test it. I'm really sorry about that.

TheBB:
No worries!

PR #5686, opened 2016-04-05
Set iedit default state

TheBB:
Emacs style users end up in normal state otherwise.


syl20bnr:
I don't know who had this issue but I think he/she was on master branch.
Since the refactor of the holy-mode this behavior cannot happen because we redirect the evil-state to emacs state, see https://github.com/syl20bnr/spacemacs/blob/develop/layers/+distribution/spacemacs-base/local/holy-mode/holy-mode.el#L31-L47

But kudos for the spot of the new variable `evil-iedit-state-default-state` which I added to fix this before I realized that advices were a better solution.

TheBB:
Hm, okay.

PR #5699, opened 2016-04-06
remove duplicated configs of eldoc

zilongshanren:
There are two exactly the same copy of eldoc config in spacemacs-base/packages.el


syl20bnr:
WOW nice spot :+1: 
Thank you !

Issue #776, opened 2015-03-07
In c++, flycheck mode "next error" clobbers compile errors


spott:
When I compile a c++ file, and there are errors, "next error" (`<SPC> e n`) doesn't work, instead finding next flycheck error instead of next compile error.

It would be nice if when the compile error window is open, "next error" went to the next compile error.

tuhdo:
While waiting for it to be fixed, you can still use stock key bindings `M-g n` and `M-g p` to go to next and previous compile error.

spott:
Awesome.  Thanks.



—
Andrew

jb55:
Related: https://github.com/flycheck/flycheck/issues/95

Looks like @jwiegley has a fork somewhere fixing this.

The `M-g {n,p}` workaround isn't very satisfying when writing Haskell code, since 95% of the time you are jumping through compile errors. It would be nice to have the less RSI `SPC e n` always work for those cases.

Not sure how to fix this since flycheck says they're not going to stop clobbering the next-error binding.

jb55:
After looking around some more it looks like you can disable the next-error behaviour with a setting:

https://github.com/flycheck/flycheck/issues/202

jb55:
But that doesn't work because it looks like `spacemacs/next-errors` calls it anyway... :(

StreakyCobra:
Can anyone check if this bug is still happening?

jb55:
I've muscle memory'd M-g {n, p}, but it would be nice to have proper bindings for these

dennishamester:
I really like what @spott suggested, that `spacemacs/next-error` should go to the next compilation error if the compilation window is open.

Here is a quick fix: dennishamester@0304a2ef03c276cb2b3673d8e7e0a4a3c9360eb3

What do you think?

StreakyCobra:
PR it, that will be better discussed there :-)

dennishamester:
done: #5712

d12frosted:
Fixed with released of Spacemacs v0.200. Let us know if you still have any problems with this issue.

PR #5712, opened 2016-04-07
Fix next-error behavior when flycheck is active

dennishamester:
@spott suggested in #776, that `spacemacs/next-error` should go to the next compilation error if the compilation window is open, regardless of whether flycheck is enabled or not. IMO, that makes the available keybindings much more usable for people who compile from inside emacs (like me :wink:)


syl20bnr:
Thank you ! :+1: 

PR #5716, opened 2016-04-07
[github layer] git-link working as expected in git-timemachine buffers

dcluna:
This commit allows an user to run git-link in a git-timemachine bufer and get the current revision instead of the latest commit in that branch.

### DIFF
--- a/layers/+source-control/github/packages.el
+++ b/layers/+source-control/github/packages.el
@@ -92,6 +92,12 @@
         (let (git-link-open-in-browser)
           (call-interactively 'git-link-commit)))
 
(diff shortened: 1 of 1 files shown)


syl20bnr:
Nice :+1: 
I changed the advice for a `cl-letf` which is local to the new function `spacemacs/git-link`, see commit: https://github.com/syl20bnr/spacemacs/commit/43bc881
Also I extend the idea and making it work with `git-link-commit` as well, see commit: https://github.com/syl20bnr/spacemacs/commit/ecffbb4

Thank you !

Issue #5719, opened 2016-04-08
Multiterm shell popup won't open


nwolfe:
#### Description
The multi-term shell fails to popup due to incorrect number of args passed to function.

#### Reproduction guide
- Start Emacs
- SPC a s m

*Observed behaviour:*
Error message at the bottom saying:
> Wrong number of arguments: (lambda (_) "Wrapper to be able to call multi-term from shell-pop" (interactive) (multi-term)), 0

*Expected behaviour:*
Popup window with multi-term

#### System Info
- OS: darwin
- Emacs: 24.5.1
- Spacemacs: 0.105.16
- Spacemacs branch: develop (rev. 2b4a039)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
((auto-completion :variables auto-completion-enable-help-tooltip t auto-completion-enable-snippets-in-popup t)
 better-defaults
 (clojure :variables clojure-enable-fancify-symbols t)
 (colors :variables colors-enable-rainbow-identifiers t)
 deft emacs-lisp git github markdown
 (mu4e :variables mu4e-installation-path "/usr/local/share/emacs/site-lisp")
 org osx puppet python ruby
 (shell :variables shell-default-shell 'multi-term shell-default-height 40 shell-default-term-shell "/bin/zsh")
 spacemacs-layouts spotify
 (spell-checking :variables spell-checking-enable-by-default nil)
 version-control yaml)
```

#### Backtrace
```
Debugger entered--Lisp error: (wrong-number-of-arguments (lambda (_) "Wrapper to be able to call multi-term from shell-pop" (interactive) (multi-term)) 0)
  multiterm()
  call-interactively(multiterm nil)
  (lambda nil (call-interactively (quote multiterm) nil))()
  shell-pop--switch-to-shell-buffer(1)
  shell-pop-up(1)
  shell-pop(nil)
  (if (equal (quote (4)) index) (call-interactively (quote multiterm)) (shell-pop--set-shell-type (quote shell-pop-shell-type) (quote ("multiterm" "*multiterm*" (lambda nil (call-interactively (quote multiterm) nil))))) (shell-pop index))
  spacemacs/shell-pop-multiterm(nil)
  call-interactively(spacemacs/shell-pop-multiterm nil nil)
  command-execute(spacemacs/shell-pop-multiterm)
```

nwolfe:
[...]

PR #5720, opened 2016-04-08
(5719) Fix multi-term shell popup

nwolfe:
This commit removes the unused argument from the wrapper lambda, which
appears to be no longer provided when the function is called.


nwolfe:
Hopefully this is the proper fix for https://github.com/syl20bnr/spacemacs/issues/5719

syl20bnr:
Thank you ! :+1: 

PR #5723, opened 2016-04-08
Fix MAJOR_VERSION, URL and URL2 in makefile for pylookup

hemcsec:
Fix MAJOR_VERSION, URL and URL2 in makefile for pylookup.


hemcsec:
Close this PR for incorrectly use PR without a dedicated branch.

PR #5726, opened 2016-04-08
Set `comment-inline-offset` to 2 for python-mode.

xiaohanyu:
Since python's PEP8 recommends two spaces for inline spaces.

Ref:
- https://www.python.org/dev/peps/pep-0008/#inline-comments

### DIFF
--- a/layers/+lang/python/packages.el
+++ b/layers/+lang/python/packages.el
@@ -241,6 +241,7 @@
               fill-column python-fill-column
               ;; auto-indent on colon doesn't work well with if statement
               electric-indent-chars (delq ?: electric-indent-chars))
+        (setq-local comment-inline-offset 2)
         (annotate-pdb)
         ;; make C-j work the same way as RET
         (local-set-key (kbd "C-j") 'newline-and-indent))


layers/+lang/python/config.el
```diff
@@ -28,3 +28,6 @@
   "Automatically set pyenv version from \".python-version\".
 
 Possible values are `on-visit', `on-project-switch' or `nil'.")
+
+(add-hook 'python-mode-hook
+          (lambda () (set (make-local-variable 'comment-inline-offset) 2)))
```
bmag: 1. Can use `setq-local` here
2. Why not set the variable inside of `python-default` instead? (see `python/init-python`in packages.el)

xiaohanyu: Thanks for reminding. First time to use `use-package`, still not quite familiar with it. Will modify this PR soon.

layers/+lang/python/packages.el
```diff
@@ -238,6 +238,7 @@
       (defun python-default ()
         (setq mode-name "Python"
               tab-width 4
+              comment-inline-offset 2
               fill-column python-fill-column
```
xiaohanyu: @bmag repushed new code. Should I use `setq-local` here or just `setq`?

bmag: I think `setq-local`, since the variable isn't buffer-local by default.

xiaohanyu: Yeah, you're right, spent one hour diving into the details of `(make-local-variable)`, (buffer-local-variables)`, `(setq-default)`, `(setq-local)`, etc. Repushed code again.

bmag: Looks good now :+1:

xiaohanyu:
@syl20bnr please just take one minute to merge this PR?

fbergroth:
Yes, elpy does this as well. :+1:

syl20bnr:
Thank you !

PR #5727, opened 2016-04-08
Improve `SPC s c` to support evil-search-module

### LINKED ISSUE #1326: When `evil-search-module` is `evil-search` then `SPC s c` doesn't work
It would be nice if `SPC s c` worked for `evil-search` mode too, I prefer `evil-search` since it's a lot like Vim's search rather than Emacs' isearch.

For some reason search highlights also look different when `evil-search` is the selected search module, I don't understand why.

### LINKED ISSUE #5101: use evil-search in evil buffers
In Vim, I seem to be able to do `/ c-r 0 Enter` to search for the contents of the yank register, but this doesn't work in Spacemacs. I didn't find an issue for this specifically upstream, only https://bitbucket.org/lyro/evil/issues/166/some-key-bindings-are-missing-in-command.

Am I missing something?

StreakyCobra:
Fix #1326
Fix #5101

### DIFF
--- a/layers/+spacemacs/spacemacs-evil/packages.el
+++ b/layers/+spacemacs/spacemacs-evil/packages.el
@@ -179,9 +179,15 @@
   (use-package evil-search-highlight-persist
     :init
     (progn
+      (defun spacemacs/evil-search-clear-highlight ()
+        "Clear evil-search or evil-ex-search persistent highlights."
+        (interactive)
+        (case evil-search-module
+          ('isearch (evil-search-highlight-persist-remove-all))
+          ('evil-search (evil-ex-nohighlight))))
       (global-evil-search-highlight-persist)
       ;; (set-face-attribute )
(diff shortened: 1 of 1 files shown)


edrex:
Awesome! Should this be pushed upstream as a change to `evil-search-highlight-persist-remove-all` ([#L150](https://github.com/juanjux/evil-search-highlight-persist/blob/master/evil-search-highlight-persist.el#L150))? Oh [I see](https://github.com/juanjux/evil-search-highlight-persist/issues/6#issuecomment-160068581), someone would have to fork and maintain that package. (hot potato!)

StreakyCobra:
I don't know, I don't have the motivation to maintain the package, so for now it's clean enough to keep it here.

syl20bnr:
I was not aware that the package maintainer gave up on Emacs. Since it is a rather short package we could import it as a local package if no one is willing to take its ownership.

Thank you !

PR #5728, opened 2016-04-08
Fix shells asking for "Run program"

StreakyCobra:
After 63dc8f7ea0b262e3bd1e164bda87d7e1265adfcf `SPC '` started to ask
for "Run program" instead of opening the terminal with the defined shell.

This commit remove the `call-interactively` that are causing this
problem. `SPC u SPC '` is still working as far as I tested.


syl20bnr:
Thank you ! :+1: 

PR #5730, opened 2016-04-08
(,r , ,r)

kuangdash:

=> anyway, here is a problem:

The origin "(r . ,r)" will trigger "wrong type argument, stringp: r"

### DIFF
--- a/layers/+distribution/spacemacs-base/local/helm-spacemacs/helm-spacemacs.el
+++ b/layers/+distribution/spacemacs-base/local/helm-spacemacs/helm-spacemacs.el
@@ -124,7 +124,7 @@
                ((string-equal r "VIMUSERS.org")
                 `("Vim users migration guide" . ,r))
                (t
-                `(r . ,r))))
+                `(,r . ,r))))
             result)))
 
 (defun helm-spacemacs//documentation-action-open-file (candidate)


syl20bnr:
Indeed: 

> Before you submit this pull request, please ensure it is against the develop branch and not master

:smile:

kuangdash:
I see it, but there will encounter a problem if the "doc" directory has org-files other than the default.

syl20bnr:
I think your fix is still useful in develop, problem is the file you modified has moved in develop branch, I know this is not ideal but development is performed in the develop branch so you have to check it out to be sure to have the last changes, in this case, the file at the correct place.

kuangdash:
I can't boot the emacs on the develop-branch (win 10 system): "emacs can't require evil" which seems to be the problem of the package "bind-map". 

so... for the people who use the master-branch …

kuangdash:
The *backtrace* picture :
[picture](http://imgur.com/J3DS1dS)

syl20bnr:
Thank you for the screenshot, which Emacs version is it ? Where did you find the binaries ?
There is definitely something wrong with the Emacs for windows, it seems autoload feature does not work for bootstrap packages on develop.

kuangdash:
emacs 24.5 on msys2, I install it by "pacman -S"

I think that the problem comes from:

```
(bind-map spacemacs-default-map
    :prefix-cmd spacemacs-cmds
    :keys (dotspacemacs-emacs-leader-key)
    :evil-keys (dotspacemacs-leader-key)
    :override-minor-modes t
    :override-mode-name spacemacs-leader-override-mode)
```

On the function of "bind-map"
here comes:
`(when evil-keys '((require 'evil)))`

While evil is not loaded …

syl20bnr:
That's weird I'll have to watch more closely at in on a Windows box.

kuangdash:
Ok, I resolve it. The problem is the version of the package "bind-map".
The difference between bind-map-2015 and bind-map-2016 cause the problem.

syl20bnr:
Nice :+1: 
Waiting for you PR against develop branch then :smile:

kuangdash:
Anyway, please merge this too. :joy:

syl20bnr:
closing in favor of #5731

syl20bnr:
@kuangdash It will be backported to master when a new release is ready.

Issue #5729, opened 2016-04-08
helm-eshell-history and emacs editing mode change to evil mode


j4:
#### Description
Use helm-eshell-history (` Alt-M m H` ) in emacs mode passes into evil mode

#### Reproduction guide
- Start Emacs
- `Alt-M a s e`
- `Alt-M m H`

*Observed behaviour:*
The problem is coming from here :
https://github.com/syl20bnr/spacemacs/blob/master/layers/shell/packages.el#L185

*Expected behaviour:*

Stay in the emacs mode

#### System Info
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.16
- Spacemacs branch: master (rev. cf8563b)
- Graphic display: t
- Distribution: spacemacs
- Editing style: emacs
- Completion: helm
- Layers:
```elisp
(auto-completion themes-megapack emacs-lisp erc elfeed git github ansible python go ibuffer shell restclient haskell markdown semantic gtags org
                 (shell :variables shell-default-shell 'eshell)
                 (shell :variables shell-default-position 'bottom shell-default-height 30)
                 (shell :variables shell-protect-eshell-prompt nil)
                 spell-checking spotify syntax-checking version-control yaml html php search-engine javascript vagrant mu4e)
```

j4:
Fix in dev. #5734

PR #5734, opened 2016-04-09
Fix helm eshell history switch to evil mode

j4:
Patch #5729 

Check if we are not in the emacs edit mode then enter evil-insert-state.


nixmaniack:
This shouldn't be required with recent changes of state handling in develop branch. We have advised these functions based on currently active editing style. Hence these guards will not be needed anymore.

Can you try this on develop? I don't seem to reproduce this on develop.

j4:
Yes it's work fine in dev :+1:

PR #5743, opened 2016-04-10
Revert "(5719) Fix multi-term shell popup"

nwolfe:
This reverts commit 8526ee7145e3b4cf6c0a96f6e21480d4df5ee2e9.

---

The cause of https://github.com/syl20bnr/spacemacs/issues/5719 was reverted, and therefore the change needs to be reverted since the multiterm popup is now broken again.


shaief:
I'm still suffering from it. Are you sure it was fixed by this PR?

nwolfe:
@shaief Yes, this fix is still accurate.  I just verified the popup is broken on head of `develop`, and this PR still fixes it.

StreakyCobra:

PR #5744, opened 2016-04-10
Fix 'makefile' for pylookup

hemcsec:
Fix 'makefile' for pylookup.


StreakyCobra:

PR #5754, opened 2016-04-11
spacemacs-evil: add evil-visual-mark-mode

### LINKED ISSUE #5619: Add a toggle to reveal the marks of evil.
Marks are not used as much in Vim as vi, thanks to visual mode and text objects, but they are still useful and Evil also has them. Evil's marks are invisible like vi's are. The only way to tell where they are is with `:marks`. Vim has several plugins to help with this. Evil also has one: the [evil-visual-mark-mode](https://github.com/roman/evil-visual-mark-mode) package, which shows them in-line as an overlay.

It would also be nice to briefly turn this on after hitting `m`, ``` ` ```, or `'`, but before selecting which character. This is consistent with the way Spacemacs shows other overlays only when needed (e.g. `:s` shows what will change as an overlay even before you do it).

The package is on MELPA, so it's possible now to add the package to the dotfile and add a custom toggle under `SPC o`, but it seems like something Spacemacs should include by default. I'm also not sure how to make the second part work.

nixmaniack:
This adds toggle for [`evil-visual-mark-mode`](https://github.com/roman/evil-visual-mark-mode) on `` SPC t ` ``

Fix #5619

### DIFF
--- a/layers/+spacemacs/spacemacs-evil/packages.el
+++ b/layers/+spacemacs/spacemacs-evil/packages.el
@@ -28,7 +28,8 @@
         ;; https://github.com/7696122/evil-terminal-cursor-changer/issues/8
         (evil-terminal-cursor-changer :excluded t)
         evil-tutor
-        (evil-unimpaired :location local)))
+        (evil-unimpaired :location local)
+        evil-visual-mark-mode))
 
 (defun spacemacs-evil/init-evil-anzu ()
   (use-package evil-anzu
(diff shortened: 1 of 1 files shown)


robbyoconnor:
@robbyoconnor wants this bad.

TheBB:
@robbyoconnor Use the thumbs up reaction to indicate that you want it.

robbyoconnor:
I did but I REALLY wanted it -- I feel strongly

syl20bnr:
Thank you ! :+1: 

PR #5763, opened 2016-04-12
Diminish space-doc-mode lighter

hemcsec:
Diminish space-doc-mode lighter to avoid sometimes incorrect display.


syl20bnr:
Thank you !

@JAremko do you think we should remove the heart again ? :-)

JAremko:
@syl20bnr  Even diminished heart is better than to be heartless :smiley:

PR #5782, opened 2016-04-13
support underscores in Go test functions

jaffee:
When trying to run a single test function with LEADER m t t in a Golang
test file, spacemacs failed to find functions which had underscores in
their names. This commit simply adds an underscore to the regular
expression which spacemacs uses when searching for the current test
function.


syl20bnr:
Thank you ! 👍 

PR #5785, opened 2016-04-13
Org download layer

krakapwa:
This layer adds org-download support to spacemacs. It allows to drag-and-drop images from your browser or file-system into an org file. You can also insert images straight from gnome-screenshot (or the like).

For further details about the org-download package, have a look at the original package homepage: https://github.com/abo-abo/org-download

I use this for my research: I can quickly copy figures/images/graphs that I find on website or pdf files (with gnome-screenshot) straight into my org document.

### DIFF
--- a/layers/+tools/org-download/packages-config.el
+++ b/layers/+tools/org-download/packages-config.el
@@ -1,0 +1,26 @@
+;;; packages-config.el --- org-download layer packages file for Spacemacs.
+;;
+;; Copyright (c) 2012-2016 Sylvain Benner & Contributors
+;;
+;; Author: Laurent Lejeune <olol85@gmail.com>
+;; URL: https://github.com/krakapwa
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+;;; Commentary:
+
+;;; Code:
+(defun org-download/init-org-download()
+  (use-package org-download 
+    :config
+    ))
+
+(defun org-download/post-init-org ()
+  (spacemacs/set-leader-keys-for-major-mode 'org-mode
+    "iy" 'org-download-yank
+    "is" 'org-download-screenshot
+    ))
+
+;;; packages-config.el ends here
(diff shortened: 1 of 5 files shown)


nixmaniack:
You should rebase your changes on top of latest `develop` and force push to this PR itself.

layers/+tools/org-download/packages-config.el
```diff
@@ -0,0 +1,26 @@
...
+;;
+;;; License: GPLv3
+
+;;; Commentary:
+
+;;; Code:
+(defun org-download/init-org-download()
+  (use-package org-download 
+    :config
+    ))
```
nixmaniack: If there's isn't any config, you can remove use-package itself as it's a `noop`

layers/+tools/org-download/packages.el
```diff
@@ -0,0 +1,22 @@
...
+
+;;; Commentary:
+
+;;; Code:
+
+(defconst org-download-packages
+  '(
+    org
+    org-download
+    )
```
nixmaniack: Also I think this belongs to `org` layer itself rather than a separate layer. Let's see what others say.

cpaulik:
I agree with @nixmaniack that this can and should go into the org layer.

krakapwa:
OK let me know when you reach a final decision on the separate layer vs. org layer debate.

NJBS:
I also agree that this should go into the org layer. The layer already includes support for additional org related packages like `org-pomodoro` and `org-present` so having this package be its own layer seems inconsistent.

layers/+tools/org-download/packages-config.el
bmag: `(use-package <package> :defer t)` is a noop, but `(use-package <package>)` is the same as `(require '<package>)`.
@nixmaniack what you're suggesting will work only if org-download defines proper autoloads.

robbyoconnor:
FYI you can push to the same branch and it will be reflected on the PR -- close this one as you have a more recent one.

layers/+tools/org-download/packages-config.el
nixmaniack: @bmag You're right, I assumed `:defer t` was there. Thanks for mentioning it.

syl20bnr:
close in favor of #5807

PR #5786, opened 2016-04-13
Allow multiple python test runners and per project test runner.

cpaulik:
This allows python-test-runner to also be a list of either `'(nose
pytest)` or `'(pytest nose)` . Dispatch functions call the primary test runner by default unless the prefix argument is set in which case the other test runner is called. 

This also adds documentation about using directory local variables to set this per project.

### DIFF
--- a/layers/+lang/python/funcs.el
+++ b/layers/+lang/python/funcs.el
@@ -68,3 +68,71 @@ when this mode is enabled since the minibuffer is cleared all the time."
 (defun spacemacs//python-imenu-create-index-use-semantic ()
   "Use semantic if the layer is enabled."
   (setq imenu-create-index-function 'semantic-create-imenu-index))
+
+(defun python-get-main-testrunner ()
+  "Get the main test runner."
+  (if (listp python-test-runner) (car python-test-runner) python-test-runner))
+
+(defun python-get-secondary-testrunner ()
+  "Get the secondary test runner"
+  (cdr (assoc (python-get-main-testrunner) '((pytest . nose)
+                                             (nose . pytest)))))
+
+(defun python-call-correct-test-function (arg funcalist)
+  "Call a test function based on the chosen test framework.
+ARG is the universal-argument which chooses between the main and
+the secondary test runner. FUNCALIST is an alist of the function
+to be called for each testrunner. "
+  (let ((test-runner (if arg
+                         (python-get-secondary-testrunner)
+                       (python-get-main-testrunner))))
+    (funcall (cdr (assoc test-runner funcalist)))))
+
+(defun python-test-all (arg)
+  "Run all tests."
+  (interactive "P")
+  (python-call-correct-test-function arg '((pytest . pytest-all)
+                                           (nose . nosetests-all))))
+
+(defun python-test-pdb-all (arg)
+  "Run all tests in debug mode."
+  (interactive "P")
+  (python-call-correct-test-function arg '((pytest . pytest-pdb-all)
+                                           (nose . nosetests-pdb-all))))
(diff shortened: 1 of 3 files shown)


syl20bnr:
All the test functions don't use the prefix argument. What do you think about leveraging the universal prefix argument for this instead ?
We can create interactive functions for each test command and dispatch given the use of `C-u` or `SPC u`. For instance to use the second test engine it will look likes `SPC u , t a`. There is another advantage to do so, given it is dynamic dispatching we can use local file or directory variable to change the value of `python-test-runner` depending on the project.

cpaulik:
Yes I think that is a good idea, I especially like the part of choosing the right test runner automatically per project. I'll update this PR.

syl20bnr:
When I talked about file and directory local variable I was referring to https://www.gnu.org/software/emacs/manual/html_node/emacs/Directory-Variables.html :-) In Spacemacs you can add them with `SPC f v ...`.

The idea is to avoid to bind the keys statically at configuration time by defining functions to read the variable `python-test-runner` and dispatch the action to the correct test functions. So you can work with different projects with different test engines setup in the same emacs session. The keys are bound to those dispatch functions instead of directly to the package test functions.

cpaulik:
Updated according to your suggestions. Thanks for the guidance @syl20bnr .

layers/+lang/python/funcs.el
```diff
@@ -68,3 +68,76 @@ when this mode is enabled since the minibuffer is cleared all the time."
...
   (setq imenu-create-index-function 'semantic-create-imenu-index))
+
+(defun python-get-main-testrunner ()
+  "Get the main test runner."
+  (cond ((eq 'pytest python-test-runner) 'pytest)
+        ((and (listp python-test-runner)
+              (eq 'pytest (car python-test-runner))) 'pytest)
+        ((eq 'nose python-test-runner) 'nose)
+        ((and (listp python-test-runner)
+              (eq 'nose (car python-test-runner))) 'nose)))
```
TheBB: How about

```elisp
(if (listp python-test-runner) (car python-test-runner) python-test-runner)
```

Or did you write it like that for safety?

layers/+lang/python/funcs.el
```diff
@@ -68,3 +68,76 @@ when this mode is enabled since the minibuffer is cleared all the time."
...
+
+(defun python-call-correct-test-function (arg funcalist)
+  "Call a test function based on the chosen test framework.
+ARG is the universal-argument which chooses between the main and
+the secondary test runner. FUNCALIST is an alist of the function
+to be called for each testrunner. "
+  (let ((test-runner (if arg
+                         (python-get-secondary-testrunner)
+                       (python-get-main-testrunner))))
+        (funcall (cdr (assoc test-runner funcalist)))))
```
TheBB: Misleading indentation.

TheBB:
I like it :+1:

When everything is just symbols it's more efficient to use `eq`, `assq` and `memq` than `equal`, `assoc` and `member`.

layers/+lang/python/funcs.el
cpaulik: It would not be that safe since there is not default condition. I just copied it from the previous version of the PR without thinking much about it. It is updated to your simpler version now.

cpaulik:
@TheBB Thanks for the feedback. I really should read more on the types in elisp.

TheBB:
Cool, this looks good to me.

syl20bnr:
Thank you ! 💜 
I prefixed the function with `spacemacs` and also did it for other functions not part of this PR.

PR #5790, opened 2016-04-14
Let counsel-search respect ivy-count-format

sooheon:
There is no reason to hardcode the formatting string for this search function in particular. Let it be set by a variable, so user can change the ivy-format as they wish.


syl20bnr:
Thank you again :-)

PR #5592, opened 2016-03-25
Disable semantic-idle-summary in Python mode.

cpaulik:
Anaconda provides more useful information but can not do it properly
when this mode is enabled since the minibuffer is cleared all the time.

Fix #5564

PR #5804, opened 2016-04-15
set semantic-idle-summary-mode on hook

nbecker:

Thank you <3

### DIFF
--- a/layers/+lang/python/funcs.el
+++ b/layers/+lang/python/funcs.el
@@ -63,4 +63,6 @@
   " Disable semantic-idle-summary in Python mode.
 Anaconda provides more useful information but can not do it properly
 when this mode is enabled since the minibuffer is cleared all the time."
-  (semantic-idle-summary-mode 0))
+  (add-hook 'semantic-mode-hook (lambda()
+                                  (semantic-idle-summary-mode 0))
+            ))


layers/+lang/python/funcs.el
```diff
@@ -63,4 +63,6 @@
   " Disable semantic-idle-summary in Python mode.
 Anaconda provides more useful information but can not do it properly
 when this mode is enabled since the minibuffer is cleared all the time."
-  (semantic-idle-summary-mode 0))
+  (add-hook 'semantic-mode-hook (lambda()
+                                  (semantic-idle-summary-mode 0))
+            ))
```
NJBS: Closing parentheses should not got on their own line.

https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html

StreakyCobra:
A little bit of context or an explanation about the motivations of this change would be better than the default template message.

nbecker:
Without this patch, attempt to enter python-mode resulted in an error: semantic-idle-summary-mode is undefined.

I don't know if this is actually the best fix.

nixmaniack:
Disabling `semantic-idle-summary-mode` should be in `post-init-semantic` as per #5592. Looks like later [changes](https://github.com/syl20bnr/spacemacs/commit/6c2665e0f78ae82891b7e2b1ddacea70e9d81933) moved it to `anaconda-mode` `:init` block. So in case `semantic` layer isn't used, it'll throw an error. 

So I think it should be moved back to post-init-semantic with anaconda-mode guard as proposed in #5592. 

cc: @cpaulik

cpaulik:
@nixmaniack Thanks, Yes I think it should be in `post-init-semantic`. @nbecker Your fix disables `idle-summary-mode` for all layers when the python layer is used. So if you want to work on it then please change your PR.

I did not really realize that @syl20bnr moved it although he said so in his comment. Also how is this merged? I don't see this commit in `develop`

syl20bnr:
This is already merged but not pushed, I indeed moved some stuff to semantic post-init. I will push soon.

syl20bnr:
Thank you ! 
Corrections were made in commit https://github.com/syl20bnr/spacemacs/commit/47776e8

PR #5807, opened 2016-04-16
Org download

krakapwa:
Added layer for org-download support. It is a package that allows to drag-and-drop images from a web-browser or file-system into org files. Also, it can be used to insert images straight from gnome-screenshot (or the like). The images are automatically saved into a user-defined folder with timestamps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,6 +240,21 @@ HOME=~/spacemacs emacs
 Note: If you're on Fish shell, you will need to modify the last command to: `env
 HOME=$HOME/spacemacs emacs`
 
+## Installation alongside another configuration
+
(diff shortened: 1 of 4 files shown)


StreakyCobra:
Your PR should only contain the last commit: "Added org-download support". All the others shouldn't be there :-)

syl20bnr:
Thank you !
See modifications made in commit https://github.com/syl20bnr/spacemacs/commit/14d3529
It would be handy to define some default `org-download-screenshot-method` for all OSes.

krakapwa:
OK I'll look into it. Thanks!

PR #5809, opened 2016-04-16
Clojure layer cleanup

sooheon:
Clean up obsolete variables, respect user dotspacemacs-editing-style, fix errors on cider-test wrappers, and improve test functions. I can rebase into single commit once reviewed and ready for merge.

### DIFF
--- a/layers/+lang/clojure/packages.el
+++ b/layers/+lang/clojure/packages.el
@@ -345,17 +345,8 @@ If called with a prefix argument, uses the other-window instead."
       (add-to-list 'magic-mode-alist '(".* boot" . clojure-mode)))
     :config
     (progn
-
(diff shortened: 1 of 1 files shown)


syl20bnr:
Thank you ! 👍 
I did not cherry-pick https://github.com/syl20bnr/spacemacs/pull/5809/commits/429689ddd7a6364547cef68d7f3fe4bedd7ab98e and https://github.com/syl20bnr/spacemacs/pull/5809/commits/fcd74b251247c75775671ce13b20b65cc15a3812 because I think removing duplication does not worth obfuscating function definitions (we lose the hability to nagivate to the function with SPC h d f for instance); and the last commit was not necessary because the key bindings are already isolated (i.e. defined only for normal state).

PR #5814, opened 2016-04-17
Wrap company kb changes in check for user prefs

sooheon:
Only use evilified bindings in company-active-map if user wants vim
style or hjkl bindings. 

As a side note, I think wrapping that predicate in its own var somewhere globally once may be nice.

### DIFF
--- a/layers/+completion/auto-completion/packages.el
+++ b/layers/+completion/auto-completion/packages.el
@@ -103,13 +103,16 @@
       (spacemacs//auto-completion-set-RET-key-behavior 'company)
       (spacemacs//auto-completion-set-TAB-key-behavior 'company)
       (spacemacs//auto-completion-setup-key-sequence 'company)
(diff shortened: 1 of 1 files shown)


TheBB:
I think the idea is to use `spacemacs-editing-style-hook` for this, so that it can be dynamically updated when the editing style changes.

syl20bnr:
Thank you !
I changed the commit to separate the non-evil bindings and the evil-bindings, I also defined `C-n` `C-p` and `C-f ` for emacs style. All these bindings are not defined by company so we don't overwrite anything.
See commit: https://github.com/syl20bnr/spacemacs/commit/667a7f6

sooheon:
@syl20bnr quick question: if I remove the company binding fn from spacemacs-editing-style-hook in a post-config-company function, will it not take any effect at all?

PR #5825, opened 2016-04-18
Only disable semantic-idle-summary in python if anaconda-mode is used.

cpaulik:
Some people have problems with anaconda mode and exclude it. This
enables them to at least have the semantic summary in their modeline.


syl20bnr:
Thank you ! 👍 
Hotfixed in master as well.

Issue #5523, opened 2016-03-17
Unable to activate package 'helm-gitignore', required package 'request-0.1.0' is unavailable


mjm:
#### Description
Upon updating my packages to get around the issue with smooth-scrolling, my Emacs setup can't finish installing packages because it can't find 'request-0.1.0', which helm-gitignore depends on.

I have two machines that I updated my packages on, running the same version of Emacs. Only one of them is stuck here.

#### Reproduction guide
- Start Emacs
- Let missing packages try to install

*Observed behaviour:*
Eventually:

```
Warning (spacemacs): 
Error connection time out for gnu repository!
Warning (initialization): An error occurred while loading ‘/Users/matt/.emacs.d/init.el’:

error: Unable to activate package ‘helm-gitignore’.
Required package ‘request-0.1.0’ is unavailable

To ensure normal operation, you should investigate and remove the
cause of the error in your initialization file.  Start Emacs with
the ‘--debug-init’ option to view a complete error backtrace.
```

Now this occurs right away upon Emacs startup.

*Expected behaviour:*
All packages install correctly.

#### System Info
- OS: darwin
- Emacs: 25.0.92.1
- Spacemacs: 0.105.14
- Spacemacs branch: master (rev. a298a5a)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(auto-completion better-defaults emacs-lisp git markdown org syntax-checking version-control common-lisp html javascript ruby shell-scripts yaml restclient themes-megapack osx gnus
                 (mu4e :variables mu4e-installation-path "/Volumes/Work/Brew/share/emacs/site-lisp/mu/mu4e"))
```

NJBS:
Can you try nuking `emacs.d/elpa` and restarting emacs with the `--insecure` flag?

mjm:
This is what happens when I did so. I'm not sure if the --insecure flag is working.

```
Warning (spacemacs): 
Error connection time out for melpa repository!
Warning (spacemacs): 
Error connection time out for gnu repository!
Warning (spacemacs): An error occurred while retrieving the theme, using default theme. (error: (error Package ‘color-theme-sanityinc-tomorrow-’ is unavailable))
Warning (initialization): An error occurred while loading ‘/Users/matt/.emacs.d/init.el’:

error: Package ‘evil-’ is unavailable

To ensure normal operation, you should investigate and remove the
cause of the error in your initialization file.  Start Emacs with
[...]

PR #5831, opened 2016-04-19
Don't mark request pacakage as built-in.

Szunti:
@syl20bnr Not sure if you only meant this in #5523. 

helm-gitignore seems to work and the request cache is created in .emacs.d/.cache/request for me . I have no idea what the removed comment means.

### DIFF
--- a/layers/+distribution/spacemacs-base/packages.el
+++ b/layers/+distribution/spacemacs-base/packages.el
@@ -39,9 +39,7 @@
         (process-menu :location built-in)
         projectile
         (recentf :location built-in)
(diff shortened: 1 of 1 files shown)


TheBB:
By the way you can use `request`, not `(request)`

Szunti:
You are totally right, changed it.

syl20bnr:
Thank you !

We may have a regression here, cc @bixuanzju 
:point_up: [21 avril 2016 22:42](https://gitter.im/syl20bnr/spacemacs?at=57198fac3307b26736e3a56c)

PR #5832, opened 2016-04-19
Add code block surround operator for org-mode

TheBB:


layers/+emacs/org/funcs.el
```diff
@@ -0,0 +1,18 @@
+;;; packages.el --- rebox layer function file for Spacemacs.
```
nixmaniack: org layer function file for Spacemacs.

TheBB: Of course!

layers/+emacs/org/packages.el
```diff
@@ -57,11 +57,9 @@
...
 (defun org/post-init-evil-surround ()
   (defun spacemacs/add-org-surrounds ()
-    (push '(?: . spacemacs//surround-drawer) evil-surround-pairs-alist))
-  (add-hook 'org-mode-hook 'spacemacs/add-org-surrounds)
-  (defun spacemacs//surround-drawer ()
-    (let ((dname (read-from-minibuffer "" "")))
-      (cons (format ":%s:\n" (or dname "")) "\n:END:"))))
+    (push '(?: . spacemacs//surround-drawer) evil-surround-pairs-alist)
+    (push '(?# . spacemacs//surround-code) evil-surround-pairs-alist))
+  (add-hook 'org-mode-hook 'spacemacs/add-org-surrounds))
```
syl20bnr: Do we have to do this in a hook or could it be done only once with a `spacemacs|add-use-package-hook` ?

TheBB: I think it has to be in a hook because `evil-surround-pairs-alist` must be buffer-local.

TheBB:
Merged.

PR #5833, opened 2016-04-19
Add org-insert-subheading keybinding

jgertm:
This commit allows for easy creation of subheadings in orgmode buffers, similar to how headings can be inserted with `SPC m h i`


layers/+emacs/org/README.org
```diff
@@ -233,10 +233,11 @@ You can tweak the bullets displayed in the org buffer in the function
 
 | Key Binding | Description                      |
 |-------------+----------------------------------|
```
joehillen: The heading needs to be realigned too.

syl20bnr:
Thank you !

PR #5835, opened 2016-04-19
Fix font definition for first frame in daemon mode

StreakyCobra:
:-)


syl20bnr:
:-)

syl20bnr:
Magic ! Maybe we should change the message because it feels strange to read it when Emacs start (for some reason there is a pause and we have a lot of time to read it).

Thank you !

StreakyCobra:
> Maybe we should change the message because it feels strange to read it when Emacs start (for some reason there is a pause and we have a lot of time to read it).

Oh, I can't see it with my setup. But yes, would be better to change it then :joy:

PR #5836, opened 2016-04-20
Fix null error

tuhdo:

Thank you <3
char-after can return nil. For that reason, we check to ensure it does
not cause error.


syl20bnr:
Thank you ! 👍 

PR #5837, opened 2016-04-20
Defer space-doc

NJBS:
`space-doc` currently adds a large chunk of time to initial load time since it has to require `org`.

### DIFF
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -535,4 +535,5 @@ a Markdown buffer and use this command to convert it.
 
 (defun org/init-space-doc ()
   (use-package space-doc
+    :commands space-doc-mode
     :config (spacemacs|diminish space-doc-mode " ❤" " d")))


TheBB:
I think you need `:commands space-doc-mode` instead? Autoloads aren't automatically generated for local packages.

JAremko:
Btw @syl20bnr [made `space-doc` lazy loaded](https://github.com/NJBS/spacemacs/commit/eb14645d0fbfe78fcfc95df1dcba17fd9704165b) by directly requiring it in [`spacemacs/prettify-org-buffer`](https://github.com/syl20bnr/spacemacs/blob/develop/core/core-funcs.el#L160) 

I'm not sure if [diminishing heart symbol](https://github.com/NJBS/spacemacs/commit/c921eb6bbca0bb261e7f0191337ed2b90cf890d4) instead of replacing it in the mode definition worth all the troubles. :confused:

JAremko:
@NJBS  Can we replace [this](https://github.com/NJBS/spacemacs/commit/eb14645d0fbfe78fcfc95df1dcba17fd9704165b#diff-f92348317ee9191057a5943114613618R179) with `use-package` shenenigans ? So `space-doc`  will load only  when required.  I think it's a simpler solution. Or use `with-eval-after-load` on `org mode.

NJBS:
@JAremko Is there a reason that `space-doc` is spread across core and org? If a user does not have the `org` layer installed the require will fail every time the user tries to view documentation.

cc @syl20bnr

syl20bnr:
Good point, it should be in core and we should probably get ride of use-package to setup the mode. It looks weird to have core functions relying on a mode setup in a layer.

syl20bnr:
Thank you !
Merged for now, we can consider to move it to core at some point.

JAremko:
@NJBS  Yeah. We forgot to move it when #5372 was merged.

Issue #5839, opened 2016-04-20
spacemacs-ivy should not override prefix-help-command


sooheon:
From: https://github.com/justbur/emacs-which-key/issues/111

Found out it was caused by this commit: https://github.com/syl20bnr/spacemacs/commit/9edc44d

This breaks default behaviour of which-key. PR fix incoming :)

PR #5840, opened 2016-04-20
ivy: don't override prefix-help-command

sooheon:
Fixes: #5839 

counsel-descbinds is accessible through SPC-?, and shadowing the C-h key
for which key makes pagination impossible. As this deviates from
which-key defaults, this should be up to user to config, not a base in
spacemacs-ivy.


syl20bnr:
Thank you !

PR #5852, opened 2016-04-21
fix yapf still being in extension, causing its loading to fail

Tritlo:
A very simple fix for the yapf bug.

### DIFF
--- a/layers/+lang/python/extensions.el
+++ b/layers/+lang/python/extensions.el
@@ -18,7 +18,6 @@
     nose
     pylookup
     python-compile
-    py-yapf
     ))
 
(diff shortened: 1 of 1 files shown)


nixmaniack:
This PR is against `master` and as per [contribution guidelines](https://github.com/syl20bnr/spacemacs/blob/master/CONTRIBUTING.org#pull-request) wouldn't be accepted. And it seems this is already fixed in `develop`.

Tritlo:
This was mostly to draw TheBB's attention and to show how I fixed it on gitter.

syl20bnr:
This is Ok I'll merge it against master since this is a hotfix.

PR #5858, opened 2016-04-22
Fixes #5897 - Fix init error for emacs in terminal mode

### LINKED ISSUE #5897: vi-tilde-fringe-mode: Symbol's function definition is void: define-fringe-bitmap
I am using a emacs configured without X (--without-x), and spacemacs report this error when booting.
I can't use (SPC h I) to report the issue, as this key is undefined.

neutralevil:
The behavior of loading `vi-tilde-fringe` only for graphical emacs
was removed in d95f9df, and changed to load it after display init.
But the `spacemacs|do-after-display-system-init` macro is no help
to terminal mode emacs.

### DIFF
--- a/core/core-display-init.el
+++ b/core/core-display-init.el
@@ -32,7 +32,7 @@ created."
   `(let ((init (cond ((boundp 'ns-initialized) 'ns-initialized)
                      ((boundp 'w32-initialized) 'w32-initialized)
                      ((boundp 'x-initialized) 'x-initialized)
-                     (t 't))))           ; fallback to normal loading behavior
+                     (t (display-graphic-p))))) ; fallback to normal loading behavior only for GUI client
      (if (symbol-value init)
          (progn
            ,@body)


StreakyCobra:
@syl20bnr Do we want `spacemacs|do-after-display-system-init` to be run only for GUI instances of Emacs? If yes it would be more general to do the checking there than guarding this package. Otherwise :+1: for this change.

syl20bnr:
Indeed we can move the check to the macro, @neutralevil can you rebase it with the check moved to the macro ?

neutralevil:
@syl20bnr I've modified the macro, can you review it?

TheBB:
The return value of the `cond` form is passed to `symbol-value`. You should probably sanitize the return value of `display-graphic-p` so that it's definitely valid as a symbol with a truthy value then, or write it in a different way:

```elisp
    (let ((init (or (bound-and-true-p ns-initialized)
                    (bound-and-true-p w32-initialized)
                    (bound-and-true-p x-initialized)
                    (display-graphic-p)
      ...)
```

I know this isn't 100% functionally identical but it should work.

neutralevil:
@TheBB How about this way:
```
(let ((init (cond ((boundp 'ns-initialized) 'ns-initialized)
                      ((boundp 'w32-initialized) 'w32-initialized)
                      ((boundp 'x-initialized) 'x-initialized)
                      ((display-graphic-p) 't)
                      (t 'nil)))))
  ...)
```
It maintain the original behavior as much as possible.
But as a newbie to elisp, I don't understand why we should return a symbol but not a value.

TheBB:
> But as a newbie to elisp, I don't understand why we should return a symbol but not a value.

The value of `init` is used in `(symbol-value init)` later on, extracting the value bound to a symbol.

That's why there is a problem with you using `(display-graphic-p)` like this. In the other cases you are returning a symbol to be looked up but in the fallback case you are evidently returning the actual value to be used directly.

Therefore this only makes sense if all the other cases also get the value directly and you drop the call to `(symbol-value)` later.

syl20bnr:
Thank you ! 👍 

PR #5860, opened 2016-04-22
use node_modules version of eslint or global if not available

axyz:
This would hopefully make eslint always work with the local version when available

### DIFF
--- a/layers/+frameworks/react/README.org
+++ b/layers/+frameworks/react/README.org
@@ -8,13 +8,9 @@
  - [[#install][Install]]
  - [[#optional-configuration][Optional Configuration]]
  - [[#key-bindings][Key Bindings]]
-   - [[#js2-mode][js2-mode]]
-   - [[#folding-js2-mode][Folding (js2-mode)]]
-   - [[#refactoring-js2-refactor][Refactoring (js2-refactor)]]
    - [[#formatting-web-beautify][Formatting (web-beautify)]]
      - [[#documentation-js-doc][Documentation (js-doc)]]
    - [[#auto-complete-and-documentation-tern][Auto-complete and documentation (tern)]]
-   - [[#json][JSON]]
 
 * Description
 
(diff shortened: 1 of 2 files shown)


axyz:
mhm probably is missing an check for the case when you have a node_modules directory but no local eslint, in those cases should fallback to eslint

axyz:
now should work

axyz:
now should work

robbyoconnor:
Why not implement similar for the javascript layer?

axyz:
well, the plan is to try it out on the react layer in order to decide if it may be worth replacing the current js2 based javascript layer with something similar based on web-mode. Also this eslint trick is something surely to be shared with the js layer anyway, but first we should also agree on having eslint as the default linter (i think now is still jshint mixed with the js2 syntax checker).

Right now I'm using the react layer on all my js files to check if there are some issues, so far tern seems to work nicely as well as jsdoc, basic skewer integration should be easy as well (though would be great to find a way to let it work with es6 transpiling with babel before sending the code to the skewer server).

without js2 we are loosing some come complex refactor, but tern mode has some good basic ones.
js2 syntax checking is more annoying than useful especially if you use es7 code with eslint.

default code comment is a bit ugly as it goes for /* code */ instead of // code.    However the good part is that it seems to work effectively when commenting out jsx, js or mixed codes.

robbyoconnor:
I suppose...

CestDiego:
given that webpack is a great tool that many people use and there is hot-module replacement already I wouldn't worry too much about skewer for react. :( 

Although having es6 traspiled would be great though. if it included sourcemaps that is.

syl20bnr:
Thank you ! 👍 
Let's see how it goes.

joefiorini:
@axyz @syl20bnr I get a warning now on Spacemacs load:

> (Spacemacs) Warning: More than one init function found for package tern. Previous owner was react, replacing it with layer javascript.

I'm assuming this is because previous only the `javascript` layer had `init-tern` and with this PR both layers now have it. I'm not sure if it causes an issue, but I can say I'm missing almost all of my keyboard shortcuts for React mode (including the tern ones) under `SPC m`. I'd like to help troubleshoot this, is there a way I can verify that react/init-tern is actually running?

syl20bnr:
@joefiorini I looked at the code seems that the init function in react layer should be a post-init.
I'm fixing it.

syl20bnr:
@joefiorini I pushed the fix, can you try it and tell me if you have correct tern config in both js2-mode and react-mode ?

joefiorini:
@syl20bnr Looks good, thanks!

PR #5861, opened 2016-04-22
Let's try to get this message unnoticed

StreakyCobra:
syl20bnr reported this message can be read on startup because there is a
small delay after it, so let's try to change it to something less weird.


StreakyCobra:
Thank me :+1: Cherry-picked into `develop`, I can safely delete my branch.

PR #5879, opened 2016-04-24
Remove duplicate definition of symbol-highlight

sooheon:
There is another one at line 174 :)


StreakyCobra:
Nice catch :smiley: Thank you :+1: Cherry-picked into `develop`, you can safely delete your branch.

PR #5887, opened 2016-04-25
Adapt hunk commands to other diff tools

iurifq:
The current keybinding only works for diff-hl. Using version-control commands abstracts the diff tool dependency away.


TheBB:
There has to be some check on whether the version control layer is enabled here. You can't just bind to functions that may not exist.

iurifq:
I get your point. Well, they are there because they are related to unimpaired. IMO moving them to the version-control layer doesn't fix the problem as unimpaired might not be configured. Besides, we would be scattering the logic related to unimpaired.

Isn't there a way to check if these functions are defined? I'm a noob in emacs, so I don't know how to do that 😔

TheBB:
I did it myself: https://github.com/syl20bnr/spacemacs/commit/325067bc1015c4ee102f49ed5d6d3c6d35b79068

This is the whole purpose of the layer system so it makes sense to leverage it in this case.

TheBB:
Sorry, I merged it before I saw your comment. Hopefully my commit shows how to achieve this kind of thing. :-)

PR #5894, opened 2016-04-26
Diminish setting for visual-line-mode.

xiaohanyu:
I'm a heavy user of `visual-line-mode`, and I guess there're lots of users as me who will turn on `visual-line-mode` globally. So I think it's better to set diminish for `visual-line-mode`.

### DIFF
--- a/layers/+distribution/spacemacs-base/packages.el
+++ b/layers/+distribution/spacemacs-base/packages.el
@@ -48,6 +48,7 @@
         undo-tree
         (uniquify :location built-in)
         (url :location built-in)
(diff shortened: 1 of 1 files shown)


TheBB:

sooheon:
I think this should be changed to `L`. Remember the convention is that the diminish character is the same as what you would use to toggle with (spc t foo). This helps prevent clashes.

TheBB:
That's a good point.

TheBB:
Fixed https://github.com/syl20bnr/spacemacs/commit/975b83b68bf9eb160aafaf402547e5a5f339d610

PR #5895, opened 2016-04-26
deft: Fix create new file with filter text

andschwa:
The binding of `SPC m n` calls `deft-new-file`, which only uses the filter text `(when deft-use-filter-string-for-filename)`, which is set to nil originally. So we need to make it true.

### DIFF
--- a/layers/+tools/deft/packages.el
+++ b/layers/+tools/deft/packages.el
@@ -33,6 +33,7 @@
         (evil-insert-state nil)))
     :config
     (progn
+      (setq deft-use-filter-string-for-filename t)
       (spacemacs/set-leader-keys-for-major-mode 'deft-mode
         "d" 'deft-delete-file
         "i" 'deft-toggle-incremental-search


andschwa:
@StreakyCobra I don't think this is an enhancement; it's a bug fix. The behavior as documented is not the actual behavior; this corrects that.

andschwa:
Rebased on latest develop.

amardeep:
@syl20bnr could you please merge this.

andschwa:
@syl20bnr @StreakyCobra What do I need to do to enable merging this behavior fix?

StreakyCobra:
Nothing, just wait for a developer to review and merge it :-) There are quite a lot of PR out there :-)

syl20bnr:
Thank you ! It is merged now :-)
I move the setting to `:init` so users can more easily customize it in their `user-config` function.

andschwa:
Awesome! Thanks very much :smile:

Issue #4687, opened 2016-01-19
Reduce gc-cons-threshold for responsiveness


tuhdo:
It's worth to have a look at this the article in this thread and its discussion: https://www.reddit.com/r/emacs/comments/41m7x3/why_are_you_changing_gcconsthreshold/. In summary, higher `gc-cons-threshold` speed up Emacs, but lower `gc-cons-threshold` gives more responsiveness. Given Spacemacs sometimes causes random freeze, I think it's worth to try setting `gc-cons-threshold` back to its default value after Spacemacs done initializing.

CestDiego:
this ^ I approve. @tuhdo what do you think of the fix that they propose though?

syl20bnr:
I repeat: https://github.com/syl20bnr/spacemacs/issues/3011#issuecomment-139950569

:-)

d12frosted:
Fixed with released of Spacemacs v0.200. Let us know if you still have any problems with this issue.

PR #5898, opened 2016-04-26
Reduce gc-cons-threshold after init

tampix:
Reset the value of `gc-cons-threshold` after init to improve overall responsiveness.
See https://github.com/syl20bnr/spacemacs/issues/4687


CestDiego:
YAS

syl20bnr:
Thank you !

PR #5900, opened 2016-04-26
Fixed agda-post-packages -> agda-packages (develop)

sindikat:
Agda-layer

Previously Agda wouldn't load at all, now everything should work.

### DIFF
--- a/layers/+lang/agda/packages.el
+++ b/layers/+lang/agda/packages.el
@@ -9,7 +9,7 @@
 ;;
 ;;; License: GPLv3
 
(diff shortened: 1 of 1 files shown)


TheBB:
Thanks, can you put everything back in packages.el though please? I think we're not going down that road after all.

sindikat:
Done.

Can you please link me to the most recent discussion on how layers should be structured? I can't find anything on Issues tab.

TheBB:

There's no official discussion on it AFAIK.

PR #5908, opened 2016-04-27
Doc: clarify search-replace tool dependecy

bmag:
Make it clearer that search-replace via helm-ag requires ag, pt, or ack.


StreakyCobra:

PR #5913, opened 2016-04-27
Don't switch to error-list window if already on it

bmag:
It seems unnecessary to open the buffer if it's already displayed in the current window, especially because using `switch-to-buffer-other-window` in that case causes `flycheck-error-list-buffer` to be opened in *two* windows.


StreakyCobra:

Maybe some people were using this to focus the error-list when already open, but indeed it was bogus if already in it… Let's see if there are complains about this change :-)

PR #5914, opened 2016-04-28
Fixes #5906 - Fix Initial Scratch Buffer Message.

### LINKED ISSUE #5906: initial-scratch-message
hi,  `(setq initial-scratch-message "aaaa")` . put it in `dotspacemacs/user-config ()`.  Why it's not working?

mayanez:
Related to issue https://github.com/syl20bnr/spacemacs/issues/5906


StreakyCobra:
We can't do only this. With this change this put back the welcome message that we do not want in the `*scratch*` buffer:

    ;; This buffer is for notes you don't want to save, and for Lisp evaluation.
    ;; If you want to create a file, visit that file with C-x C-f,
    ;; then enter the text in that file's own buffer.

TheBB:
Agree with @StreakyCobra, this won't be the solution.

Issue #5812, opened 2016-04-17
Error happens when search something in project using ag


hujianxin:
#### Description
There are errors when using `ag`. The error messages is: 
```
error in process sentinel: time-since: Wrong type argument: number-or-marker-p, nil
error in process sentinel: Wrong type argument: number-or-marker-p, nil
```

#### Reproduction guide
- Start Emacs
- Open a git project using `SPC p l`
- Type `SPC s p` to search something in this project
- Type anything you what to search
- Error happens

*Observed behaviour:*
The lines contain the words I what to search.

*Expected behaviour:*
```
error in process sentinel: time-since: Wrong type argument: number-or-marker-p, nil
error in process sentinel: Wrong type argument: number-or-marker-p, nil
```

#### System Info
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.18
- Spacemacs branch: develop (rev. 5ee970f)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
(spacemacs-ivy
 (auto-completion :variables auto-completion-enable-snippets-in-popup t)
 better-defaults git version-control
 (shell :variables shell-default-height 30 shell-default-position 'bottom)
 spell-checking syntax-checking dockerfile org markdown latex django emacs-lisp
 (c-c   :variables c-c  -enable-clang-support t)
 (ess :variables ess-enable-smart-equals nil ess-enable-smartparens t)
 html javascript python
 (scala :variables scala-auto-insert-asterisk-in-comments t)
 (clojure :variables clojure-enable-fancify-symbols t)
 scheme racket sql shell-scripts yaml ranger evil-cleverparens evil-snipe colors deft cscope gtags restclient scala2)
```

sooheon:
Can confirm I see this with all ivy-based ag searches.

syl20bnr:
Confirmed that it happens only with ivy.

d12frosted:
Fixed with released of Spacemacs v0.200. Let us know if you still have any problems with this issue.

PR #5918, opened 2016-04-28
pcre2el is not loaded in `spacemacs-base`

ralesi:
pcre2el is not loaded by `spacemacs-base`, but is needed for ivy support.  Should address #5812


StreakyCobra:

PR #5919, opened 2016-04-28
Fix evil-map warning.on load

ralesi:
Definition of `evil-map` in the bootstrap funcs.el causes a warning at startup, as evil-normal-state-map is still undefined at this point.  Renamed funcs.el to packages-funcs.el, as we don't need these functions until evaluation of packages.el anyway.


syl20bnr:
Thank you again !
This is the one I did not cherry-pick, I just added a `with-eval-after-load` to wrap the defmacro.

PR #5920, opened 2016-04-28
Async is not present in `spacemacs-base`

ralesi:
async is not loaded in `spacemacs-base`; however `core-release-management` and the check for new versions makes use of the `async-start` macro.  This checks if `async is present, and if not, skips the check for new version at startup.


syl20bnr:
Thank you !
I fixed it another way (hopefully) by adding async package to `spacemacs-boostrap` layer.

syl20bnr:
@ralesi Actually I cherry-picked this commit, I then added async to `spacemacs-boostrap` layer.

PR #5922, opened 2016-04-28
Add shell shell-default-full-span variable to shell layer

dubnde:
Enhancement to add `shell-default-full-span` variable to allow customisation of full span window for the shell.


syl20bnr:
Thank you !

PR #2299, opened 2015-07-10
Add support for fancy frame title

usharf:
@syl20bnr  Follow-up to #2139, this is a somewhat naive, I believe, attempt allow users to format the frame title dynamically.

The default behavior is that only when there is more than one frame, the frame-title is updated to show buffer/file name in the title. At least for me this is the default behavior.

This solution uses the built-in format specs (which are passed verbatim) and adds couple of new ones, like:

%a abbreviated-file-name (if buffer visiting a file, or buffer name)
%t name of projectile project (or "-")
%S system-name
%I invocation-name (name of the program, which is "Emacs")
%U user name (taken for $USER, if defined)

I wasn't sure where in `core-spacemacs.el` put the initialization code, if it even belongs there. 

Right now, as is, frame-title will blank before becoming visible again, so I guess it's not the right place.

Also, default for configuration template is "%I@%S", which gives me "Emacs@imac.local" (my hostname), which is what you get by default with spacemacs or without any startup init at all.

Example user customization  (in .spacemacs)

```emacs-lisp
   dotspacemacs-frame-title-format "%a [%t] (%U@%S)"
```

I should also add that not all original specs work as expected, and I'm not even sure if they should work when used for  frame-title (notably the %+, %- etc.)

Issue #2139, opened 2015-06-28
Fancy frame-title-format


usharf:
I wasn't able to find anything relating to setting up a title format. 

I'm using something like the following, which I've adapted from someone's dotemacs to remove dependency on `cl-lib` (sorry, but did't keep record). 

It'll set the title on an Emacs frame (GUI) to the `abbreviated-file-name [projectile-project-name]` - if projectile is enabled and there's an active project, otherwise, set title to buffer name.

for example

```
"~/spacemacs-private/.spacemacs [spacemacs-private]"
```

Taught it might be useful.

```emacs-lisp
  (defun spacemacs//frame-title-format ()
    "Return frame title with current project name, where applicable."
    (let ((file buffer-file-name))
       (if file
           (concat (abbreviate-file-name file)
                   (when (and (bound-and-true-p projectile-mode)
                              (projectile-project-p))
                     (format " [%s]" (projectile-project-name))))
           "%b")))

  (when (display-graphic-p)
    (setq frame-title-format '((:eval (spacemacs//frame-title-format)))))
```

syl20bnr:
This is great :+1:  I would suggest to make it a dotspacemacs variable taking some sort of format where we invent that `%f` is the filename, `%F` is the filepath, `%p` the project name etc...

usharf:
Looking into that!

usharf:
@syl20bnr looking into that!

usharf:
@syl20bnr tentative PR at #2299

PR #5924, opened 2016-04-28
Add support for fancy frame title (cont'd)

cviebig:
Dear contributors,

I would like to re-submit the change by @usharf from pull request #2299 (originating in #2139) for merging into current `develop`. I've rebased the change and resolved the merge conflicts. Unfortunately `develop` is non functional for me at the moment so I could only test the change on `master` where it works like a charm.

The change introduces the possibility to configure the formatting and contents of frame titles. By doing this users can now customize frame titles to contain for example the names of the currently active buffer or project. The default behauvior remains unchanged. Therefore the frame title will be formatted as `user@hostname` if the user does not customize this setting.

Thank you very much in advance

Best regards
Christoph

### DIFF
--- a/core/core-spacemacs.el
+++ b/core/core-spacemacs.el
@@ -82,6 +82,9 @@ the final step of executing code in `emacs-startup-hook'.")
   (setq dotspacemacs-editing-style (dotspacemacs//read-editing-style-config
                                     dotspacemacs-editing-style))
   (configuration-layer/initialize)
+  ;; frame title init
+  (when (and (display-graphic-p) dotspacemacs-frame-title-format)
+    (setq frame-title-format '((:eval (spacemacs/frame-title-prepare)))))
   ;; default theme
   (let ((default-theme (car dotspacemacs-themes)))
     (spacemacs/load-theme default-theme)
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -119,6 +119,10 @@ of a list then all discovered layers will be installed.")
 Press `SPC T n' to cycle to the next theme in the list (works great
 with 2 themes variants, one dark and one light")
 
+(defvar dotspacemacs-frame-title-format nil
+  "Default format string for a frame title bar, using the
+  original format spec, and additional customizations.")
+
 (defvar dotspacemacs-colorize-cursor-according-to-state t
   "If non nil the cursor color matches the state color in GUI Emacs.")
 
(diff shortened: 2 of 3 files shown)


cviebig:
Should we introduce a new layer? This might reflect the community response to this pull request and would keep additional ballast out of the spacemacs core.

syl20bnr:
I like this PR and we can put it directly in core.
Can you add some documentation in DOCUMENTATION.ORG about the additional supported format?

core/core-spacemacs.el
```diff
@@ -82,6 +82,9 @@ the final step of executing code in `emacs-startup-hook'.")
   (setq dotspacemacs-editing-style (dotspacemacs//read-editing-style-config
                                     dotspacemacs-editing-style))
   (configuration-layer/initialize)
+  ;; frame title init
+  (when (and (display-graphic-p) dotspacemacs-frame-title-format)
+    (setq frame-title-format '((:eval (spacemacs/frame-title-prepare)))))
```
mskorzhinskiy: I would assume the **icon-title-format** should be defined here too. Or separate variable should be introduced that will control **icon-title-format**.

mskorzhinskiy:
There are any chance to add support of **which-func-mode** (https://www.emacswiki.org/emacs/WhichFuncMode)?

core/core-spacemacs.el
cviebig: I'm afraid that I quite did not get what an icon title is. Does it configure only the icon or more?

mskorzhinskiy: From documentation (GNU Emacs 24.5.1):
> Template for displaying the title bar of an iconified frame.
> (Assuming the window manager supports this feature.)

You can see iconified frames with title bars in a tiling windows manager
with a tabs support. When one window is displaying and other are not,
that windows that are not displaying are called "tabbified" or
"iconfied". But their title bars are still displaying as a tabs.

So without setting this variable you will get different frame name
format for different frames.

Despite the fact that this behavior irritates me, I would say that for
some use-cases it can be useful.

kepi: I think that for tiling window managers the **icon-title-format** is more imporant than classic title because you can use it to differentiate between tabs.

As you can see on this screenshot, only active window have title and all others (where it would be really usefull) doesn't.

Iconified Titles in Xmonad

It make sense to have two special option for icon title as you may need much shorter version. I'm using layout where all tabs are equal sized, but  it may be more common to have active tab larger than non-active (at least icon name suggest this behavior).

kepi:
@cviebig do you plan to rebase this to current develop and prepare documentation @syl20bnr requested? If there are no plans maybe somebody can take over? Thanks.

kepi:
As @cviebig is not responding I resubmitted this PR against current `develop` with documentation and support for icon-title-format in #8824

syl20bnr:
Thank you ! 👍 

chrissound:
Is there any documentation for this feature?

kepi:
Yes, it is described in standard documentation. It is only in develop currently so head to https://github.com/syl20bnr/spacemacs/blob/develop/doc/DOCUMENTATION.org#frame-title

PR #5934, opened 2016-04-29
Make pug files use jade-mode since jade is now called pug

robbyoconnor:
Jade was renamed to pug -- we can't just yank jade support yet -- but the language itself hasn't changed -- so add .pug to the mode list


syl20bnr:
Thank you !

robbyoconnor:
Flycheck won't work for pug files (for now)

syl20bnr:
Ok, thanks for the warning.

PR #5938, opened 2016-04-29
[WIP] Fixes powerline separators on OS X

calebmeyer:
I recently uninstalled railwaycat's emacs-mac-port (the old recommendation for OS X) and installed @d12frosted's emacs-plus.

When I did that my powerline separators became just slightly off in color from the rest of the spaceline:
screen shot 2016-04-29 at 8 59 57 am

Asking on chat yielded that ns-use-srgb-colorspace is set to true by default but should be nil.

I don't know if I put the code in the right place, but I'm open to suggestions.


calebmeyer:
After the change:
screen shot 2016-04-29 at 9 33 42 am

d12frosted:
I am not sure that this is proper solution, but it would be great if it really fixes this problem in GNU Emacs :)

syl20bnr:
The issue of setting it to `nil` is that colors are washed out so I prefer users to take explicit actions to disable it.

d12frosted:
Than we should at least change the note in FAQ. Anyway, current proposed
solution conflicts with actual instructions in readme file :)

~ d12frosted

syl20bnr:
Yep, also changing the powerline separator is another solution for this.

d12frosted:
But in his screenshot he already changed separator. I think.

syl20bnr:
Only the variable `ns-use-srgb-colorspace` is set to nil.
Changing the separator would change the shape of the separators, from arrow to a straight bar for example so the user can still let `ns-use-srgb-colorspace` to t.

nixmaniack:
Yeah, I think using `alternate` and `bar` style diminishes the difference in color. And using `utf-8` makes it go away completely.

calebmeyer:
+1 using one of those separator types makes this unnecessary. However, I believe the default separator type still has this problem. 

If it is preferred, we could move the documentation mentioning the variable into the README under the OS X install emacs section.

syl20bnr:
Indeed it should be mentioned directly on the README.

syl20bnr:
I did not see that this is a PR, @calebmeyer can you update the PR to keep the value to nil but improve the README with the various solutions mentioned in this PR ?

nixmaniack:
I as well did not see this is a PR! I think setting `ns-use-srgb-colorspace` to `nil` isn't proper solution(as @d12frosted pointed) since it makes whole emacs colors little off. Hence we should document solutions discussed here.

calebmeyer:
@syl20bnr yep I'll update it as soon as I'm back to my computer. Might be tomorrow, we have a busy weekend.

calebmeyer:
Looks like the build failed due to `s.el` not being available.

d12frosted:
Thank you :+1: Squashed and cherry-picked into `develop`, you can safely delete your branch. I've made several changes in 93cfb7f4ffc2450deca38709995eb2062399d8a3.

calebmeyer:
Will do. Thanks

PR #5939, opened 2016-04-29
add support for c++-mode to gtags layer

michalslonina:
GTags seams to work find in c++-mode.


michalslonina:
will add support for c-mode too

PR #5940, opened 2016-04-29
add C and C++ mode support in gtags layer

michalslonina:
GTags works nice for both C and C++ modes.


NJBS:
C and C++ have their own layer which enables gtag support.

See its [packages.el](https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Blang/c-c%2B%2B/packages.el#L105-L108)

TheBB:
Yep, this is already in. Sorry!

PR #5941, opened 2016-04-29
Feature/add company ghci to haskell layer

dysinger:
Aloha

I added company-ghci code completion to the haskell layer.

History: it used to be that we needed ghc-mod to do things like code completion.  These days with ghc 7.10 we can just ask ghci for this information (IIRC it works with ghci-ng too for older versions of ghc).

-Tim


dysinger:
Sorry for the noise. I accidentally PR against master from develop

PR #5942, opened 2016-04-29
Adds company-ghci to haskell layer

dysinger:
Aloha

I added company-ghci code completion to the haskell layer.

History: it used to be that we needed ghc-mod to do things like code completion.  These days with ghc 7.10 we can just ask ghci for this information (IIRC it works with ghci-ng too for older versions of ghc).

-Tim

### DIFF
--- a/layers/+lang/haskell/packages.el
+++ b/layers/+lang/haskell/packages.el
@@ -15,6 +15,7 @@
     company
     company-cabal
     company-ghc
+    company-ghci
     flycheck
     flycheck-haskell
     ghc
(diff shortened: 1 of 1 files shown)


syl20bnr:
Thank you ! 👍 
I refactored the company related packages using `:toggle` keyword in package list, it simplifies a bit the configuration. Related commit: https://github.com/syl20bnr/spacemacs/commit/425e6ba

d12frosted:
At last I've got opportunity to test this package. Thank you for bringing it up, so far it works like a charm. I even considering to vote it to be default completion backend instead of `ghc-mod` as `company-ghci` is easier to setup (no need to install external dependency) and it works with different `ghc` versions, while `ghc-mod` is tied to one version of `ghc` (the one that was used to build `ghc-mod`) - which leads sometimes to very unfriendly behaviour. 

Going to play with it a little bit more on my weekends. 😸

PR #5945, opened 2016-04-29
core: add another code folding method using origami to spacemacs-editing

nixmaniack:
While working on some heavily nested JSON data(works with other major modes as well), I was finding evil fold capabilities quite limiting. Hence I looked at other folding packages and found [origami](https://github.com/gregsexton/origami.el) to be much nicer. I've configured it since then and thought of sharing it here.  I'm mostly using `TAB` a lot to show/hide like org does for nested structure. Let me know the feedback if it's useful or needs improvement or trash it if it's not helpful! 

Functionality is [documented](https://github.com/gregsexton/origami.el#what-can-it-do) on origami repo nicely. Do check it out. 

To keep it consistent with evil-fold commands, I remap them to substituting origami commands when new dotspacemacs variable `dotspacemacs-folding-method` is set to `origami`. Code folding transient state is available on `SPC z .`

Demos:
origami-demo

origami-demo-2

* layers/+spacemacs/spacemacs-editing/packages.el (spacemacs-editing/init-origami):
  add origami package and initialize it
* core/core-dotspacemacs.el (dotspacemacs-override-evil-folding): new variable to allow choosing between different code folding methods. Currently supported `evil` and `origami`

### DIFF
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -166,6 +166,9 @@ if used there.")
 (defvar dotspacemacs-ex-substitute-global nil
   "If non nil, inverse the meaning of `g' in `:substitute' Evil ex-command.")
 
+(defvar dotspacemacs-folding-method 'evil
+  "Code folding method. Possible values are `evil' and `origami'.")
+
 (defvar dotspacemacs-default-layout-name "Default"
   " Name of the default layout.")
 
--- a/layers/+spacemacs/spacemacs-editing/packages.el
+++ b/layers/+spacemacs/spacemacs-editing/packages.el
@@ -21,6 +21,7 @@
         link-hint
         lorem-ipsum
         move-text
+        origami
         pcre2el
         smartparens
         uuidgen))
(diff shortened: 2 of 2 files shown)


core/core-dotspacemacs.el
```diff
@@ -162,6 +162,9 @@ size to make separators look not too crappy.")
 (defvar dotspacemacs-ex-substitute-global nil
   "If non nil, inverse the meaning of `g' in `:substitute' Evil ex-command.")
 
+(defvar dotspacemacs-override-evil-folding nil
```
StreakyCobra: Maybe a `dotspacemacs-folding-method` taking as option `'evil` or `'origiami` will be easier to extend to eventual future methods.

nixmaniack: That sounds nice. Do you want to have a transient state for `'evil` folding method then?

nixmaniack: @StreakyCobra I have made the changes as suggested. Do check out and review! :)

syl20bnr:
How does it compare with https://github.com/mrkkrp/vimish-fold ?

nixmaniack:
@syl20bnr I hadn't checked it. But I tried it now and it only works on `active regions` which isn't helpful if I want to fold all possible foldable code. 

Since origami is based on parsers, it works pretty well for supported modes else it fallbacks to indentation based folding.

TheBB:
I just tried this out and it's really cool. There's a bug (at least in elisp) when there's a function on the first line in the file though (I can't toggle that). Otherwise this is brilliant. Love the transient state.

ralesi:
@nixmaniack, this is a good work, but it might be better to use the internal evil folding alist to map origami to the standard evil maps.  This will allow the user to use the same bindings depending on which folding mode is active and preferred for each filetype instead of just having a dotspacemacs variable.  I am away from the computer right now, but can give you some more specifics later.

TheBB:
I was under the impression that the evil API is a bit limited compared to origami's.

nixmaniack:
@TheBB Thanks for the feedback. Indeed it looks like a bug, would report it upstream.

@ralesi Any help is appreciated as I don't have enough understanding of internals of evil. 

I really like the idea of being able to switch the folding methods dynamically since in Emacs world we don't have one package to cover all cases. But my concern was, all these folding packages keep some state about foldings and overlays, mixing them together might create more issues. 

Also at this stage, origami provides much more functionality than what evil's folding api provides, in which case we will have to maintain these additional functions(e.g. next/previous fold, close rec, show only etc), in that case it'd be good to propose these changes upstream or maintain a package with these additional vimish functions(for origami, maybe evil-origami). And this would be the case with any other folding method we support.

Regarding, using folding method based on mode and type, do you have any suggestions? I found origami to work almost everywhere.

ralesi:
@nixmaniack ,

I started working on a package to help make this seamless, but I haven't completed it yet.  Below is some of the code to it.  The idea being that you call evil-fold-action using a custom list with all the actions you are interested in adding.  We can completely overwrite the standard evil-fold-list with all the enhancements and add additional properties to the list in the case of origami.  The benefit of going this route is that we define the mappings once, but can add functionality to other folding minor modes (like 'vimish-fold', 'outline-minor-mode', or 'hs-minor-mode') using these same properties.

I was originally thinking we can make a completely separate `folding` layer.  Would this make sense instead of appending to `spacemacs-editing`? 

````
;; evil fold actions
(defun evil-hideout-decrement-level ()
  (interactive)
  (evil-fold-action evil-hideout-fold-list :decrement-level))

(defun evil-hideout-increment-level ()
  (interactive)
  (evil-fold-action evil-hideout-fold-list :increment-level))

(defun evil-hideout-cycle ()
  (interactive)
  (evil-fold-action evil-hideout-fold-list :cycle))
```

```
(setq evil-hideout-fold-list
      `(((hs-minor-mode)
         :open-all   hs-outline-show-all
         :close-all  hs-hide-all
         :toggle     hs-toggle-hiding
         :open       hs-show-block
         :open-rec   nil
         :cycle      hs-cycle-level
         :decrement-level hs-decrement-level
         :increment-level hs-increment-level
         :close      hs-hide-block)
        ((hide-ifdef-mode)
         :open-all   show-ifdefs
         :close-all  hide-ifdefs
         :toggle     nil
         :open       show-ifdef-block
         :open-rec   nil
         :close      hide-ifdef-block)
        ((outline-mode
          outline-minor-mode
          org-mode)
         :cycle      outline-cycle
         :open-all   outline-show-all
         :close-all  ,(lambda ()
                        (with-no-warnings (outline-hide-sublevels 1)))
         :toggle     outline-toggle-children
         :open       ,(lambda ()
                        (with-no-warnings
                          (outline-show-entry)
                          (outline-show-children)))
         :open-rec   outline-show-subtree
         :close      outline-hide-subtree
         :decrement-level outline-hide-more
         :increment-level outline-show-more))
      )
````

nixmaniack:
@ralesi That looks better and should definitely be in its own layer or package. And since my elisp and evil fu isn't great and you're already working on this, it would be good if you take this. I can help wherever I can!

sooheon:
Btw, @TheBB, because you say you just tried it out; if I'm not a contributor to this repo, is there a simple way I can check out a PR like this with git, or do people just copy paste the code?

TheBB:
You have to add the contributor's fork as a remote:

    git remote add nixmaniack git@github.com:nixmaniack/spacemacs

Then fetch:

    git fetch nixmaniack

Then check out the branch:

    git checkout nixmaniack/origami-code-folding

StreakyCobra:
@sooheon Or use an helper like [`hub`](https://github.com/github/hub) (that can be aliased instead of `git`). This let you do stuff like this:

`git checkout https://github.com/syl20bnr/spacemacs/pull/5945`

:smiley:

sooheon:
@TheBB @StreakyCobra Thanks guys, the @TheBB's method works easily with magit, so I'll use that :)

@nixmaniack I think 127 characters for the transient state is a bit long; would you consider this as a more compact version that respects 80 chars?

"
Close^^           Open^^            Toggle^^            Goto^^         Other^^
───────^^──────── ─────^^────────── ─────^^──────────── ──────^^────── ─────^^─────────
[_c_] at point    [_o_] at point    [_a_] at point      [_n_] next     [_s_] single out
[_C_] recursively [_O_] recursively [_A_] all           [_p_] previous [_R_] reset
[_m_] all         [_r_] all         [_TAB_] recursively ^^             [_q_] quit"

Also, I think adding ("C-g" nil :exit t) and ("<escape>" nil :exit t) actions would be nice.

nixmaniack:
Not sure why I didn't get any notification for the activity on this thread after my last reply. :-/

@sooheon Sure, I'll update the docstring to respect 80 chars. 

Regarding trying out some PR, I follow this https://help.github.com/articles/checking-out-pull-requests-locally/

ralesi:
@nixmaniack,  after I finish with a couple other commits, I plan on working on folding.  Are we set on just "folding"  for the layer,  or "evil-folding"?

syl20bnr:
Nice 👍 
I split the config, the evil part is in the `init-evil`, related commit https://github.com/syl20bnr/spacemacs/commit/430a3e4
Thank you !

Issue #3875, opened 2015-11-19
Spacemacs-layout is reset after sync-configuration-layers


dsdshcym:
After `sync-configuration-layers`, the current layout configuration will lost.
And there only left one `Default` layout.

I don't know if this is a feature or a bug. But I think it's better if we can keep the opening layouts.

bmag:
I can reproduce this with latest develop, with these 3 simple steps:
1. start spacemacs: I now have one layout - the default one.
2. open a new layout with `SPC l 2`: I now have two layouts.
3. reload config with `SPC f e R`: I now have only one layout again - the default one.

#### System Info
- OS: gnu/linux
- Emacs: 24.5.2
- Spacemacs: 0.105.0
- Spacemacs branch: develop (rev. 8191a66)
- Distribution: spacemacs
- Layers:
```elisp
((auto-completion :variables auto-completion-enable-help-tooltip nil auto-completion-enable-sort-by-usage nil auto-completion-enable-snippets-in-popup t)
 (git :variables git-magit-status-fullscreen t)
 version-control markdown org
 (syntax-checking :variables flycheck-check-syntax-automatically
                  '(save mode-enabled))
 evil-cleverparens c-c++
 (clojure :variables clojure-enable-fancify-symbols t)
 emacs-lisp python
 (shell :variables shell-default-shell 'shell)
 search-engine ibuffer smex themes-megapack theming eyebrowse window-purpose colors)
```

d12frosted:
The same here. Repo steps are really easy. Just as @bmag noticed.

AlejandroCatalina:
I'm looking at this and so far I think we could add some variable that will be bound to `nil` if a restart is needed and `t` if we want to preserve the current perspectives after syncing. 

The thing is to know where this variable should take effect (obviously not when an actual emacs restart is hapennig). At first sight, I thought this variable would be defined and set in the layer, and *somehow* look at it before resetting. 

I'm digging into the syncing process to see how exactly it goes, but I'm not sure if it is possible to check variables and perform alternative actions at syncing time or if we should change the current core behavior on this thing.

Any thoughts and points welcome.

syl20bnr:
Close it since it is part of a milestone.

PR #5952, opened 2016-05-01
Fixes #5925 - layouts: don't double activate persp-mode

### LINKED ISSUE #5925: SPC-f-e-R resets the layout of all frames created with make-frame/emacsclient -c
#### Description
For the purposes of this bug emacsclient -c/make-frame/new-frame has the same behaviour so I'll just use make-frame in the description.

A new frame created using make-frame will inherit the layout of the last active frame if it was also created using make-frame. Subsequent calls to make-frame et. al. repeats this process. Edit: make-frame will work as intended if it's called from a frame created with 'emacsclient -c $file' if no calls to make-frame from a frame created with make-frame has been made.

Edit: The above issue is reproducible in bare emacs with persp-mode. Ie. running 'emacs --debug-init --daemon -q --execute "(progn (package-initialize) (setq persp-auto-resume-time -1) (require 'persp-mode) (persp-mode))"'. (But syncing can't be tested of course)

Calling dotspacemacs/sync-configuration-layouts will set the layout of all frames created by make-frame et. al. to the most recent inherited layout.

Reproduced using a clean .emacs.d/ and the .spacemacs template. The forked persp-mode (#4120, #4121) fixes it somehow, so doesn't reproduce in 0.105.19. Excluding perps-mode or using the fork fixes the issue in develop. But that's obviously suboptimal.

#### Reproduction guide
- Start Emacs (without an explicit file)
- change the layout in some way, eg.:
  - switch to another buffer
  - move the cursor to some point
  - create a new sub-window
  -  lets call the resulting layout layout_1
- make-frame / emacsclient -c
- modify the layout of both frames, eg:
  - switch to another buffer (buffer-2)
  - etc.
- spc-f-e-R or spacemacs/reload

*Observed behaviour:*
The new frame inherits layout_1 when created. After any modification spc-f-e-R
will reset both frames to layout_1.

*Expected behaviour:*
The new frame should inherit the active buffer not the whole layout if created with make-frame, if created with emacsclient -c the buffer should be the default buffer. (This seems to be both sane and standard emacs behaviour).
The sync should preserve the modified layouts.

#### System Info
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.19
- Spacemacs branch: develop (rev. b21c717)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(spacemacs-helm emacs-lisp)
```

bmag:
If `persp-mode` is already activated, don't re-activate it. Should fix bug where running `dotspacemacs/sync-configuration-layers` (<kbd>SPC f e R</kbd>) resets all the layouts.
I didn't test it thoroughly, but it seems ok.

Ref: #5925 #3875 
CC participants of those discussions: @hedning @dsdshcym @d12frosted @AlejandroCatalina

### DIFF
--- a/layers/+spacemacs/spacemacs-layouts/packages.el
+++ b/layers/+spacemacs/spacemacs-layouts/packages.el
@@ -108,8 +108,11 @@
             persp-reset-windows-on-nil-window-conf nil
             persp-set-last-persp-for-new-frames nil
             persp-save-dir spacemacs-layouts-directory)
-      ;; always activate persp-mode
-      (persp-mode)
+      ;; always activate persp-mode, unless it is already active (e.g. don't
+      ;; re-activate during `dotspacemacs/sync-configuration-layers' - see
+      ;; issues #5925 and #3875)
+      (unless (bound-and-true-p persp-mode)
+        (persp-mode))
       ;; layouts transient state
       (spacemacs|transient-state-format-hint layouts
         spacemacs--layouts-ts-full-hint


syl20bnr:
Nice !! 💜 
Thank you ! 👍 

PR #5959, opened 2016-05-02
layouts: add a function to spacemacs-layout layer to open a file in its project

aroig:
This PR adds `spacemacs/find-file-in-project`, that opens a file, in a new persp, enabling projectile mode on it. The project opened is the innermost project in the directory tree that contains the file, and the perspective is named as the root directory for the project.

This is useful to open a project from the command line. I use this bash alias, for instance

```
ep () { emacsclient -e '(spacemacs/find-file-in-project "$1")'; }
```

### DIFF
--- a/layers/+spacemacs/spacemacs-layouts/funcs.el
+++ b/layers/+spacemacs/spacemacs-layouts/funcs.el
@@ -55,6 +55,25 @@ current perspective."
   (helm-mini)
   (add-hook 'ido-make-buffer-list-hook  #'persp-restrict-ido-buffers))
 
+(defun spacemacs/find-file-in-project (filename)
+  "Open a file like find-file. If the file belongs to a project, creates
+   a new persp and enables projectile mode for it."
+  (interactive (list (read-file-name "Find file: " nil default-directory (confirm-nonexistent-file-or-buffer))))
+  ;; need the require since the projectile functions used here are not auto-loadable.
+  (require 'projectile)
+  (let* ((persp-reset-windows-on-nil-window-conf t)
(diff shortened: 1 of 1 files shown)


layers/+spacemacs/spacemacs-layouts/funcs.el
```diff
@@ -55,6 +55,25 @@ current perspective."
   (helm-mini)
   (add-hook 'ido-make-buffer-list-hook  #'persp-restrict-ido-buffers))
 
+(defun spacemacs/find-file-in-project (filename)
+  "Open a file like find-file. If the file belongs to a project, creates
+   a new persp and enables projectile mode for it."
+  (interactive "P")
```
cpaulik: Doesn't `(interactive "P")` mean that the first function argument will be the numeric prefix argument?

aroig: > Doesn't (interactive "P") mean that the first function argument will be the numeric prefix argument?

Yes it does... I copy-pasted that from the wrong place!

robbyoconnor:
99% certain the CI failure isn't the fault of this PR

aroig:
> 99% certain the CI failure isn't the fault of this PR

Yup, I've squashed the commits to trigger the CI. They pass now!

By the way I fixed the the `(interactive ...)` thing. Now it uses helm when called interactively.

mukhali:
https://github.com/technomancy/find-file-in-project
maybe we need this package?

WarFox:
This is exactly what I was looking for. This automatically adds a new perspective/layout when opening a file from new project directory. Super useful

Issue #5958, opened 2016-05-02
Propose to rename spacemacs-buffer.el to spacemacs-buffer.txt


zilongshanren:
Since I enable semantic layer for emacs-lisp-mode, when Spacemacs starts up, it will load the spacemacs-buffer.el and it will also activate semantic and other minor mode for emacs-lisp-mode which is very time consuming...

So I have renamed the suffix of spacemacs-buffer.el to spacemacs-buffer.txt and it reduce about 1.5s for the loading time 😄

zilongshanren:
![image](https://cloud.githubusercontent.com/assets/785541/14953334/55e8255e-109b-11e6-8776-d56d81e0b797.png)

JAremko:
Impressive :shipit:

bmag:
Wow, nice find, I had no idea. I assume by "renamed" you mean "changed the value of `spacemacs-buffer--cache-file`"?

Loading all these minor modes and other file-opening-related stuff is done because `spacemacs/dump-vars-to-file` uses `find-file-noselect` internally.

Rewriting `spacemacs/dump-vars-to-file` with `with-temp-file` should be simpler and avoid all the extra loading. I'm thinking something along the lines of:
```elisp
(with-temp-file filename
  (spacemacs/dump varlist buf)
  (make-directory (file-name-directory filename) t))
```
But feel free to see what works and make a PR. (not sure I'll have the time to do it myself before the weekend)

zilongshanren:
@bmag 
I mean change the suffix name of `spacemacs-buffer.el` file to the `.txt` suffix, thus when loading `spacemacs-buffer.el` at startup, it will not activate many time consuming operations related to emacs-lisp-mode.

Or we could avoid loading `spacemacs-buffer.el` when spacemacs start up.

bmag:
AFAICT `spacemacs-buffer.el` is a cache file and is accessed for two reasons: to read the cached value of `spacemacs-buffer--release-note-version`, and to write it back when it changes.
When writing the value to `spacemacs-buffer.el`, the current implementation causes Emacs to visit the file, although that isn't really necessary.
I can't test now, but I think the writing is done automatically at startup only if the release notes are displayed automatically at startup.

bmag:
> Or we could avoid loading spacemacs-buffer.el when spacemacs start up.

The slowness is due to saving the file, not because of loading it. (surprisingly)

zilongshanren:
@bmag  
[...]

PR #5960, opened 2016-05-02
Better spacemacs/dump-vars-to-file and shorter startup time

bmag:
Visiting the dump file before writing to it is slow, because it triggers all the regular actions that happen when a user opens a file, e.g. enable major mode and minor modes. We don't need all that, `with-temp-file` is what we really want.

This rewrite of `spacemacs/dump-vars-to-file` removes the startup lag reported by @zilongshanren in #5958.

### DIFF
--- a/core/core-funcs.el
+++ b/core/core-funcs.el
@@ -86,17 +86,12 @@ and its values are removed."
       (push (pop tail) result))
     (nreverse result)))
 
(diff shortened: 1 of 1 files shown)


TheBB:

I didn't see a big reduction in startup time from this, but anyway it's an improvement.

PR #5962, opened 2016-05-02
Defer loading of persp mode until after user-config

### LINKED ISSUE #3881: dotspacemacs-auto-resume-layouts interferes with hooks
After enabling `dotspacemacs-auto-resume-layouts` (running on the develop branch) I've noticed that all buffers are loaded before my `dotspacemacs/user-config` runs, which means any hooks I add in there don't fire when the buffers load. 

Is this expected / is that the wrong place to configure my hooks?

For now, I've been moving these types of hooks configuration into `custom-set-variables`, which I'd rather not have to do because I can't comment them or format them.

cpaulik:
This should fix #3881

Tested it with `(add-hook 'python-mode-hook 'highlight-indentation-mode)` in my `user-config`

### DIFF
--- a/layers/+spacemacs/spacemacs-layouts/packages.el
+++ b/layers/+spacemacs/spacemacs-layouts/packages.el
@@ -108,11 +108,15 @@
             persp-reset-windows-on-nil-window-conf nil
             persp-set-last-persp-for-new-frames nil
             persp-save-dir spacemacs-layouts-directory)
-      ;; always activate persp-mode, unless it is already active (e.g. don't
-      ;; re-activate during `dotspacemacs/sync-configuration-layers' - see
-      ;; issues #5925 and #3875)
-      (unless (bound-and-true-p persp-mode)
-        (persp-mode))
+
+      (defun spacemacs//activate-persp-mode ()
+        "Always activate persp-mode, unless it is already active.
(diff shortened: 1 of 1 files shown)


aaronjensen:
Tested and confirmed that it fixes the issues I was running into with #3881

layers/+spacemacs/spacemacs-layouts/packages.el
```diff
@@ -112,7 +112,7 @@
       ;; re-activate during `dotspacemacs/sync-configuration-layers' - see
       ;; issues #5925 and #3875)
       (unless (bound-and-true-p persp-mode)
-        (persp-mode))
+        (spacemacs/defer-until-after-user-config #'persp-mode))
```
bmag: I'm not sure how the deferring works here, but won't it be more robust to defer the `bound-and-true-p` guard as well?

cpaulik: You are right. Reading the comment above actually explains it.

cpaulik: Put the whole thing into a function right there. Hope that is OK.

TheBB:
Let's try it. :-)

PR #5964, opened 2016-05-03
fix the README for the rebox layer

WillianPaiva:
change the prefix on the README from "i" to "x" to correspond to the keybinds used on the package.el


TheBB:

PR #5998, opened 2016-05-06
Fix #5966, slime-eval-last-sexp is not defined

### LINKED ISSUE #5966: [SPC m e e] in Lisp mode runs the command slime-eval-last-sexp, which is not defined
[SPC m e e] in Lisp mode runs the command slime-eval-last-sexp, which is not defined.
But there is a slime-eval-last-expression, should it be the right function?

111lll1l1l:
I don't know if i am right to change it to slime-eval-last-expression, there is also an eval-last-sexp.


TheBB:
`eval-last-sexp` is for Emacs lisp.

syl20bnr:
Thank you !

PR #6001, opened 2016-05-06
Prevent warnings when toggles are turned on

curtmack:
A recent commit introduced a warning when toggles are turned on. This is due to improper quoting in the `spacemacs|add-toggle` macro, which I've corrected here.


sooheon:
This also prevents spacemacs/default-pop-shell from working properly.

TheBB:

PR #6004, opened 2016-05-06
Add Darkokai theme to themes-megapack

ekmecic:
Add the new[ Darkokai theme](https://github.com/sjrmanning/darkokai) to themes-megapack.
The theme was added to MELPA today.

I feel that it's different enough from the proliferation of other Monokai-inspired themes to merit inclusion in the layer.

### DIFF
--- a/layers/+themes/themes-megapack/packages.el
+++ b/layers/+themes/themes-megapack/packages.el
@@ -32,6 +32,7 @@
     dakrone-theme
     darkburn-theme
     darkmine-theme
+    darkokai-theme
     darktooth-theme
     django-theme
     dracula-theme


JAremko:
Looks good man

joehillen:
This needs `(setq darkokai-mode-line-padding 1)` else the modeline will have thick border like this.

screenshot from 2016-05-09 12-13-14

syl20bnr:
Thank you ! 👍 
I also pushed the changes for mode-line padding, thank you @joehillen.

myme:
I noticed the modeline border issue happening when trying out `darkokai` the other day. It does seem to correct itself when toggling through themes, but when `darkokai` is the selected theme on bootup it's broken (first item in `dotspacemacs-themes`). Manually setting `darkokai-mode-line-padding` like @joehillen mentioned in `dotspacemacs/user-init` resolves the issue on boot as well.

May be that the way this is handled in the themes-megapack layer isn't entirely waterproof? I guess the fact that there's code there for this means that nobody wants users to have to add this to their `.spacemacs`.

I just switched over to trying out Emacs (through Spacemacs), so I barely know what I'm talking about here...

syl20bnr:
Does it stil happen with last develop ?

myme:
I can't reproduce in `develop` (f6ccc1eb6f5ce252c17f025c30bb2ae7723caa40).

I see that now it's keeping the Spacemacs dark theme during initialization, then switching to `darkokai`, which would allow for the theme init function to run.

👍

joehillen:
@syl20bnr Seems to be fixed. Putting `darkokai-mode-line-padding` in `user-init` is no longer required.

joehillen:
@syl20bnr However, if darkokai is my default theme, I end up with amalgam between darkokai and spacemacs-dark at startup until I switch to a different theme and back.

syl20bnr:
@joehillen Indeed, I pushed the fix on develop, I guess you should set `darkokai-mode-line-padding`again as it was a side effect of the bug.

Issue #5999, opened 2016-05-06
SPC SPC is undefined on latest develop


aldanor:
After pulling the most recent `develop` today, `SPC SPC` is no longer working (`M-x` still does):

```
SPC SPC is undefined
```

#### System Info
- OS: windows-nt
- Emacs: 24.5.1
- Spacemacs: 0.105.19
- Spacemacs branch: develop (rev. 78f11f4)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
((auto-completion :variables auto-completion-enable-sort-by-usage t)
 (c-c++ :variables c-c++-default-mode-for-headers 'c++-mode c-c++-enable-clang-support t)
 csv deft emacs-lisp eyebrowse
 (git :variables git-magit-status-fullscreen t magit-save-repository-buffers 'dontask magit-revision-show-gravatars nil)
 (github :variables gh-profile-default-profile "aldanor")
 html javascript ipython-notebook markdown org
 (python :variables python-fill-column 99 python-test-runner 'pytest)
 rust search-engine
 (shell :variables shell-default-shell 'eshell shell-enable-smart-eshell t)
 shell-scripts spacemacs-helm syntax-checking themes-megapack theming version-control yaml aldanor)

StreakyCobra:
I can't reproduce on `develop` (same revision as you) + packages updated.

aldanor:
Interesting, I can't reproduce it either on my Mac with *exactly* the same config, the same develop revision etc. Wonder why could it be suddenly messed up on Windows? Will confirm on Monday.

WillianPaiva:
I'm on linux and have the same problem

aldanor:
Ok so it's not just me then. How do we reproduce this though?

StreakyCobra:
FWIW I also tried a totally fresh install (i.e. no `.spacemacs`, fresh `.emacs.d` on `develop`) and it's working as expected.

WillianPaiva:
i did this change on my version and it works now 

https://github.com/syl20bnr/spacemacs/pull/6009

TheBB:
@WillianPaiva That is not likely the correct solution. The key binding is set here: https://github.com/syl20bnr/spacemacs/blob/develop/layers/+completion/spacemacs-helm/packages.el#L225

You should probably check if you have any errors reported in `dotspacemacs/user-config` which prevents this hook from running as expected.

[...]

PR #6009, opened 2016-05-07
fix SPC SPC missing

WillianPaiva:
fix the missing SPC SPC for the helm-M-x

### DIFF
--- a/layers/+completion/spacemacs-helm/packages.el
+++ b/layers/+completion/spacemacs-helm/packages.el
@@ -180,7 +180,10 @@
 
       ;; use helm by default for M-x, C-x C-f, and C-x b
       (unless (configuration-layer/package-usedp 'smex)
-        (global-set-key (kbd "M-x") 'helm-M-x))
+        (global-set-key (kbd "M-x") 'helm-M-x)
+        (spacemacs/set-leader-keys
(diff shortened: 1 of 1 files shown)


layers/+completion/spacemacs-helm/packages.el
```diff
@@ -180,7 +180,10 @@
 
       ;; use helm by default for M-x, C-x C-f, and C-x b
       (unless (configuration-layer/package-usedp 'smex)
-        (global-set-key (kbd "M-x") 'helm-M-x))
+        (global-set-key (kbd "M-x") 'helm-M-x)
+        (spacemacs/set-leader-keys
+          "SPC" 'helm-M-x)
```
StreakyCobra: I'm not sure about this solution as it is discussed in #5999 that not everybody has this problem. But for sure `dotspacemacs-emacs-command-key` should be used instead of literal `"SPC"`.

StreakyCobra: Also, it is already done here: https://github.com/syl20bnr/spacemacs/blob/develop/layers/+completion/spacemacs-helm/packages.el#L219-L225

WillianPaiva:
yep i see the error it was the command-key we can close this pr.

PR #6010, opened 2016-05-07
Fix Elisp = prefix

NJBS:
It wasn't actually being set :smile:


syl20bnr:
Nice catch 👍 
Thank you !

Issue #5996, opened 2016-05-06
Add natural languages layer


JAremko:
**I think Spacemacs can benefit from a [prose](http://emacs.stackexchange.com/a/14238) layer**

 It should Include something like spell check (we have it), grammar check, punctuation, sentence navigation/manipulation, translation, synonyms  search etc.  Looks like Emacs has it all [[1]](https://www.emacswiki.org/emacs/CategoryWriting), [[2]](https://www.emacswiki.org/emacs/ThesauriAndSynonyms), [[3]](https://github.com/istib/wordsmith-mode), [[4]](http://unconj.ca/blog/linting-prose-in-emacs.html), [[5]](https://github.com/gbgar/wordlint.el). 
 We can organize the packages in one clear layer. But how to manage it? It's not like it should be enabled each time someone views txt, org, md, files. Might be wrap it in a "prose" minor mode?

**Any suggestions? What should be in the layer. How to approach it ?**

StreakyCobra:
Somehow related: #5403

JAremko:
@StreakyCobra   And somehow I even commented there. 
> The table layout is broken on GitHub. Probably because it has a gap - | should be --|
Then :alien:  come and wiped my  mind.

mkcode:
@JAremko - I think we should continue to use this issue as the overall discussion for a prose layer. Any single PR would only be a piece of it.

I left from documentation of current issues with proselint-flycheck here: https://github.com/mkcode/spacemacs/blob/95539151fd5c3f289f47aa94b1ff02c3306df89e/layers/prose/README.org#current-issues.

> Any suggestions? What should be in the layer. How to approach it ?

There are so many different ways to go with a general prose mode, like you showed with the 5 links in the top post. We should check out https://github.com/bnbeckwith/writegood-mode and  https://github.com/sachac/artbollocks-mode. My choice of what to do next would be to get rid of the toggle and have the feedback always on like how flyspell is done. But as of now, that flyicheck-next issue I described in the README is preventing that.

Also, I'm certain that I would like the prose layer to be developed in a way where we could use it in any text or prog modes. Whatever feedback I want for writing, I would like everywhere, including in my long form code comments.

Btw - Most likely I will not be able to contribute the amount of time required to make this really nice so please, anyone else can pick this up.

JAremko:
WIP wishlist. I think we can put most of it behind `SPC x` (text) and `SPC t` (toggle)

[...]

PR #6012, opened 2016-05-08
Add natural layer

mkcode:
Adds the beginnings of a prose layer.

Replaces #5403 according to the discussion in #5996.


layers/prose/packages.el
```diff
@@ -0,0 +1,48 @@
...
+(defconst prose-packages
+  '(flycheck))
+
+(defun prose/post-init-flycheck ()
+
+  (defun prose-proselint-enabled-p ()
+    "Workaround for proselint flycheck checker predicate"
+    prose-proselint-enabled)
+
+  (flycheck-define-checker proselint
```
JAremko: Mb it will better to add this https://github.com/amperser/proselint/blob/master/plugins/flycheck/ to the load path instead of embedding? This way we won't need to maintain the code.

JAremko: I think I'll submit a  PR to proselint  repo

JAremko:
It looks like flycheck/flycheck#939 will add proselint.

mkcode:
> It looks like flycheck/flycheck#939 will add proselint.

Nice find @JAremko. The body of the checker, using json, looks a lot better than what I have in this PR. On the other hand, I'm not happy with the `modes` option to `flycheck-define-checker` in the PR. This will make it so it is only available in those modes. If we want it available to be used everywhere, we should use the `predicate` option instead like I have in this PR.

What is your opinion of what we should do? I guess the ideal would be to use the upstream flycheck checker and replace `modes` to `predicate` within spacemacs?

JAremko:
@mkcode It hasn't been merged yet! I think you should ask in the PR before it's too late

The :ship:  might sail away  :shipit:

JAremko:
> This will make it so it is only available in text and markdown modes. If we want it available to be used everywhere, we should use the predicate option instead like I have in this PR.

@mkcode  I think we can use `flycheck-add-mode` with a config variable.  So a user will be able to specify where the linter should be enabled.  I'll  prototype something. But I think we should wait unit the flycheck PR is merged.

mkcode:
@JAremko - Great feedback. 👍  I asked @lunaryorn for instruction for the best way to do this. He should know best.

mkcode:
@JAremko - I think I fixed the issue with 25d0a841fc2ea555ec3ec19f!!! 🎉 Its working well for me at least. Now it looks like I can leave proselint running, get its feedback in prog modes, and also get my prog mode checkers feedback.

JAremko:
Great idea.  Now we need  @syl20bnr  feedback.   

Why `text-mode` in `:modes` doesn't count as the superset mode of  all text related modes? Aren't they supposed to be derived from it ?

mkcode:
@JAremko - flycheck doesn't check for derived modes: https://github.com/flycheck/flycheck/blob/master/flycheck.el#L1601-L1602

JAremko:
@mkcode I updated readme to  the new standard also renamed  and moved the layer. New name fits better if we want to extend it in the future.  

You can cherrypick it or simply copypasta :smile: https://github.com/mkcode/spacemacs/commit/8dbc974065bba79b55c10c7f7de6928909aa676f

mkcode:
@JAremko - cherry picked your commit. What was the motivation to rename as natural mode? Perhaps I missed something in gitter?

Also - I'll squash later. I need to get to my real job, now... :-)

JAremko:
@mkcode  I only renamed the layer. layers on the develop branch are organized in folders. So the layer should go in some category.  `+lang` already has non programming languages so it should fit. And the name of the layer should be broder if we want to add more tools #5996   Better to come up with a perfect name now because renaming a layer that is in use will be a pain in the rear end for sure.

JAremko:
*If @syl20bnr  is ok with the way it is done and the new name/place we'll also need to rename prose to natural in the code.*

jazen:
Any update?

beta1440:
This layer should be called "language" instead of "natural" since this layer is about language, not nature.

viccuad:
I would call it prose, as language or natural sounds too open, and "prose mode" is self explicative. I would like my bikeshed blue, too :).

robbert-vdh:
It would be nice to have other tools for working with natural language (like LanguageTool #8899) integrated into the layer. That would probably require having to put the different tools behind flags as most of them have got external dependencies, but it would make it a lot easier to add more tools without cluttering things up too much.

jazen:
What's missing to get this PR merged?

layers/+lang/natural/packages.el
```diff
@@ -0,0 +1,59 @@
...
+    ;; :modes (text-mode markdown-mode gfm-mode))
+
+  (add-to-list 'flycheck-checkers 'proselint)
+
+  (defun prose-reset-proselint-next-checkers ()
+    "Adds the usual mode checkers as next checkers to the proselint checker"
+    ;; Reset the proselint next-checkers prop to nil
+    (setf (flycheck-checker-get 'proselint 'next-checkers) nil)
+    ;; Add the regular mode and predicate checkers for this buffer
+    (dolist (checker (flycheck-possibly-suitable-checkers))
```
jazen: `flycheck-possibly-suitable-checkers` is not available (anymore?) in flycheck 32. I raised a simiilar issue here https://github.com/abingham/flycheck-vale/issues/10#issuecomment-385277741 but it seems that flycheck needs to resolve a [long-running issue](https://github.com/flycheck/flycheck/pull/1072) first.

duianto:
The following PRs changes have been applied to the `develop` branch:
Add a layer for LanguageTool support #8899

smile13241324:
I am not sure if integrating proselint is still necessary given that the project looks to be pretty stale, see here https://github.com/amperser/proselint/, last commit is from before 4 years. If it would still be useful, I would volunteer to put it into the spell-checking layer.

Though I would prefer to have the build in flycheck package be extracted to its own package and offered to flycheck in general to take the rest of the emacs world with us. But this is something I can do on the way.

Let me know if this is still needed.

smile13241324:
Looks like this is so old and fragmented we will be better of starting anew with a fresh design for a such a layer. Therefore I will close this PR.

PR #6017, opened 2016-05-09
Update init.el

YasuharuIida:
- suppress error on emacs ( it's version is < spacemacs-emacs-min-version).
spacemacs-emacs-min-version 's type is not int but string.


syl20bnr:
Nice 👍 
Thank you !

PR #6031, opened 2016-05-11
Layouts: graphical/terminal frames should have different workspaces

### LINKED ISSUE #5978: emacsclient --tty crashes immediately (regression)
#### Description
I can no longer connect tty clients to a graphical emacs on osx. I bisected, and
this is the offending commit:

https://github.com/syl20bnr/spacemacs/commit/cf3068e4efcaff792fb6e7f140492213eebfdf74

Because of its nature (it is simply re-enabling a hook that was disabled by an
upstream change), reverting the commit is probably not the right thing to do. 

#### Reproduction guide
- Start Emacs
- Run `emacsclient --tty something`

*Observed behaviour:*
emacsclient starts and crashes immediately

This error appears in messages:

```
Error during redisplay: (eval (window-numbering-get-number-string)) signaled (wrong-type-argument hash-table-p nil) [2 times]
```

*Expected behaviour:*
emacsclient does not crash

#### System Info
- OS: darwin
- Emacs: 25.0.93.1
- Spacemacs: 0.105.19
- Spacemacs branch: nil (rev. cf3068e)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: helm
- Layers:
```elisp
(better-defaults spacemacs-helm emacs-lisp markdown
                 (auto-completion :variables auto-completion-enable-sort-by-usage nil auto-completion-enable-snippets-in-popup t)
                 colors osx github javascript react deft floobits ruby
                 spell-checking ranger version-control rcirc tmux evil-little-word auto-correct frame-geometry cleverparens-lispy)
```

bmag:
Use two different sets of workspaces for each perspective - one set for graphical frames, and one set for terminal frames. This is required because workspaces from graphical frames are not compatible for use in terminal frames.

Should fix #5978, but I didn't have time for extensive testing.

### DIFF
--- a/layers/+spacemacs/spacemacs-layouts/funcs.el
+++ b/layers/+spacemacs/spacemacs-layouts/funcs.el
@@ -492,16 +492,52 @@ STATE is a window-state object as returned by `window-state-get'."
 
 ;; Eyebrowse and Persp integration
 
+(defun spacemacs//get-persp-workspace (&optional persp frame)
+  "Get the correct workspace parameters for perspective.
+PERSP is the perspective, and defaults to the current perspective.
+FRAME is the frame where the parameters are expected to be used, and
+defaults to the current frame."
+  (let ((param-names (if (display-graphic-p frame)
+                         '(gui-eyebrowse-window-configs
+                           gui-eyebrowse-current-slot
+                           gui-eyebrowse-last-slot)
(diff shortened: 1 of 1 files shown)


aaronjensen:
This works for me

syl20bnr:
@bmag Nice work with the layouts as always 💯 

Thank you !

Issue #5436, opened 2016-03-10
linum-mode is too slow in large org-mode buffer


levindu:
When in large org-mode buffer (6K+ lines), turning on linum-mode make cursor movement very slow, almost unbearable.

I suggest to use nlinum-mode, which works quite well in my situation.

Or provide an option of line number backend for user to choose linum-mode or nlinum-mode?

a13ph:
I've seen some linum alternatives which posit they work well in large buffers (by jit display or smth else)

PR #6036, opened 2016-05-12
add nlinum layer

CodeFalling:
A layer to replace linum-mode with [nlinum-mode](https://elpa.gnu.org/packages/nlinum.html), and line-relative-mode with [nlinum-relative-mode](https://github.com/CodeFalling/nlinum-relative).

They would have better performance with large complex files(such as large org file)

### DIFF
--- a/layers/nlinum/README.org
+++ b/layers/nlinum/README.org
@@ -1,0 +1,46 @@
+#+TITLE: nlinum layer
+
+# TOC links should be GitHub style anchors.
+* Table of Contents                                        :TOC_4_gh:noexport:
+ - [[#description][Description]]
+ - [[#feature][Feature]]
+ - [[#config][Config]]
+ - [[#important-note][Important Note]]
+ - [[#install][Install]]
+ - [[#key-bindings][Key bindings]]
+
+* Description
+This layer provide a faster drop-in replacement for =linum-mode= and =linum-relative-mode=.
+
+* Feature
++ [[https://elpa.gnu.org/packages/nlinum.html][nlinum-mode]] as replacement of linum-mode
++ [[https://github.com/CodeFalling/nlinum-relative][nlinum-relative]] as replacement of nlinum-relative-mode
+
+* Config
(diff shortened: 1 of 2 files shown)


TheBB:
This is great, but I don't think we need a layer for this. It's something we should use by default when the user has line numbers enabled.

CodeFalling:
In this case I need take some time to optimize `nlinum-relative-mode` first.

StreakyCobra:
Related: #5436

TheBB:
Is nlinum-relative any worse than linum-relative?

CodeFalling:
@TheBB I just create nlinum-relative today so it may be not so stable. 

- It can't specify current line number's face for now. 
- It won't update line-number when moving, scrolling etc (Or to say, only update when idle).

StreakyCobra:
If it become the spacemacs' default, it would be nice to put them on MELPA instead of using github fetcher :-)

TheBB:
Okay... well anyway I was thinking of using nlinum by default for a while now anyway. It's possible for the moment to fall back on linum if the user wants relative, or to introduce another option for this package... or just let them specify the delay. Not sure. :-)

But as @StreakyCobra says, for sure you should put your package on MELPA.

CodeFalling:
MELPA's PR is waiting for merge in fact😂 

Be patient, I'll try to fix shortcomings of `nlinum-relative` first.

CodeFalling:
I have tried set delay of `nlinum-relative`  to `0`, then it become as slow as `linum-relative` in large org file but get a immediate updating.

So I am going to add current line face custom for `nlinum-relative`, then make them all default and provide a option to custom delay as need. 

- [ ] add current line face custom for `nlinum-relative`
- [ ] put `nlinum-relative` on MELPA
- [ ] make another PR to make `nlinum` and `nlinum-relative` default
- [ ] provide option to custom delay

StreakyCobra:
Should this be closed in favor of #6037?

PR #6037, opened 2016-05-12
Repalce to nlinum and nlinum-relative

CodeFalling:
Move from https://github.com/syl20bnr/spacemacs/pull/6036

- [X] add current line face custom for `nlinum-relative`
- [X] put `nlinum-relative` on MELPA
- [X] make another PR to make `nlinum` and `nlinum-relative` default
- [X] provide option to custom delay(just `nlinum-relative-delay` is fine)
- [X] make `nlinum-relative` smooth in even large orgmode file

### DIFF
--- a/layers/+distribution/spacemacs-base/config.el
+++ b/layers/+distribution/spacemacs-base/config.el
@@ -117,13 +117,7 @@ It runs `tabulated-list-revert-hook', then calls `tabulated-list-print'."
 (setq window-combination-resize t)
 ;; Show column number in mode line
 (setq column-number-mode t)
-;; Activate linum-mode in all prog-mode and text-mode buffers if the setting is
-;; enabled.
-(when dotspacemacs-line-numbers
-  (add-hook 'prog-mode-hook 'linum-mode)
-  (add-hook 'text-mode-hook 'linum-mode))
-;; line number
-(setq linum-format "%4d")
+
 ;; highlight current line
 (global-hl-line-mode t)
 ;; no blink
--- a/layers/+distribution/spacemacs-base/keybindings.el
+++ b/layers/+distribution/spacemacs-base/keybindings.el
@@ -246,9 +246,9 @@
   :documentation "Move point according to visual lines."
   :evil-leader "tL")
 (spacemacs|add-toggle line-numbers
-  :status linum-mode
-  :on (linum-mode)
-  :off (linum-mode -1)
+  :status nlinum-mode
+  :on (nlinum-mode)
+  :off (nlinum-mode -1)
   :documentation "Show the line numbers."
   :evil-leader "tn")
 (spacemacs|add-toggle auto-fill-mode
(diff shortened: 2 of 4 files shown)


TheBB:
Thanks :-)

Maybe I'm misunderstanding something but judging from your [README](https://github.com/CodeFalling/nlinum-relative#usage) the delay option is already in?

CodeFalling:
Yes, should it be a `spacemacs-xxx` variable, or just keep current situation ?

TheBB:
I'm fine with the current situation.

syl20bnr:
@CodeFalling we try to avoid duplicating package variables unless for a few cases.

I suggest 0.2 seconds delay since it is what is used generally in Spacemacs (indent guides, rainbow delimiters, parens highlighting etc...)

CodeFalling:
@syl20bnr With a delay it could move smoothly in even large org file, but line number won't get update immediately(see below gif). Is it what we want?

nlinum-relative-delay

syl20bnr:
Nope, I believe that the delay what only to transform absolute to relative numbers. I.e. the absolute numbers are update in real time but the relative number are displayed after 0.2s

CodeFalling:
Current situation is display relative number in `normal mode` and absolute number in `insert mode`. 

What do you mean is always display absolute number when editing, moving, scrolling etc. Then after a 0.2s delay **auto** transform to relative numbers(back to absolute number when operate)? Is there any misunderstanding?

CodeFalling:
Or "absolute numbers" refer to "current line number"(which is highlighted), and other line(relative numbers) should be displayed after 0.2s?

CodeFalling:
nlinum-relative-large-org

I fixed performance bug of `nlinum-relative`, now we can just drop `delay` at all. It work with large org file smooth with real time update.

CodeFalling:
Everything is ok, this PR is ready for merge now.

CodeFalling:
`Invalid face reference: linum` in new frame seems a bug fixed in emacs 25 nightly version

syl20bnr:
Nice PR 💜 
Thank you ! 👍 

luxbock:
For future googlers, I had `(add-hook 'prog-mode-hook 'linum-mode)` in my config, which breaks nlinum.

phinicota:
I'm experiencing some kind of input lag with a delay lower than 0.05 for instance with cpp files.
This is not the case when I open simple files (with less parsing?) like .spacemacs.
Is this related?
This fixed it.

```
 (setq nlinum-relative-redisplay-delay 0.2)
```

#### System Info
- OS: gnu/linux
- Emacs: 24.5.2
- Spacemacs: 0.105.20
- Spacemacs branch: develop (rev. 463e477)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(helm
 (auto-completion :variables auto-completion-complete-with-key-sequence-delay 0.04)
 emacs-lisp
 (ranger :variables ranger-override-dired t ranger-show-preview t ranger-cleanup-eagerly t)
 games python
 (c-c++ :variables c-c++-enable-clang-support t c-c++-default-mode-for-headers 'c++-mode)
 wakatime semantic spotify git extra-langs
 (shell :variables shell-default-height 30 shell-default-term-shell "/bin/bash" shell-enable-smart-eshell t shell-default-position 'bottom)
 syntax-checking)
```

PR #6038, opened 2016-05-13
Add text object for org-link

et2010:
This PR add a text object for org-link so we can easily select  a whole link in org-mode buffer with `v i l` while in normal state.

### DIFF
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -109,12 +109,10 @@
                 (save-match-data (org-element-property :end (org-element-context))))))
 
       (add-hook 'org-mode-hook (lambda ()
-                                 (let ((old-map evil-outer-text-objects-map)
-                                       (new-map (make-sparse-keymap)))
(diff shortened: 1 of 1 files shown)


TheBB:
Could you make it valid only in org-mode at least? People who use evil-little-word have other things on `l` that are valid in all modes.

et2010:
@TheBB OK, I'll try to make it org-only and move the key-binding to `k`

et2010:
@TheBB OK, it's done. Now it's available only in org mode. Maybe I should also update the documentation?

et2010:
Sorry for the ugliness. Let me know if I should close this and send a new PR.

Miciah:
I think https://github.com/syl20bnr/spacemacs/pull/9041 supersedes this PR because with recent evil-org, `v i e` (`evil-org-inner-object`) selects the entire link.

Compro-Prasad:
@syl20bnr @et2010 Seems like this can be closed.

PR #6041, opened 2016-05-13
Define ahk lookup options.

ralesi:
Accommodate both lookup functions in ahk-mode.


StreakyCobra:

syl20bnr:
Thank you !
I moved `SPC m h r` to `SPC m h H` since it is a modifier of the same action.

PR #3109, opened 2015-09-23
Add imenu-list layer

bmag:
imenu-list provides a side bar that shows all imenu entries of the current buffer. More info at https://github.com/bmag/imenu-list

I chose `SPC b i` as the binding to toggle imenu-list, as mnemonic for "buffer index", but of course that can be changed.

How imenu-list differs from NeoTree and sr-speedbar: NeoTree only shows files and directories, sr-speedbar shows files, directories and imenu indexes, and imenu-list shows only imenu indexes.

How imenu-list differs from `spacemacs/jump-in-buffer` (<kbd>SPC s l</kbd>): imenu-list is a side bar that stays visible and updates automatically when switching buffers and serves as a visual reminder of the buffer's structure, while `spacemacs/jump-in-buffer` serves as a jump command and doesn't remain visible.

PR #6042, opened 2016-05-13
Let "bt" goto todo buffer

ralesi:
`SPC b t` seems sensible to goto the project todo buffer.

### DIFF
--- a/layers/+emacs/org/README.org
+++ b/layers/+emacs/org/README.org
@@ -409,6 +409,7 @@ org-present must be activated explicitly by typing: ~SPC SPC org-present~
 
 | Key Binding | Description                                                |
 |-------------+------------------------------------------------------------|
+| ~SPC b t~   | Switch to the TODO list buffer                             |
 | ~SPC C t~   | capture a TODO item for the current project                |
 | ~SPC C T~   | capture a checklist item for the current project           |
 | ~SPC a o p~ | show the TODO list for the current project                 |
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -487,6 +487,7 @@ Headline^^            Visit entry^^               Filter^^                    Da
     :init
     (progn
       (spacemacs/set-leader-keys
+        "bt" 'ort/goto-todos
         "Ct" 'ort/capture-todo
         "CT" 'ort/capture-checkitem
         "aop" 'ort/list-project-todos)


StreakyCobra:

ralesi:
Documentation added

agzam:
Although it seems to be natural choice, IMO we should be very careful and try to not pollute groups of "core bindings". Not everyone uses Org-mode all the time. I think this particular binding belongs to `SPC a o` group

d12frosted:
@agzam I see your point. But this is in `org` layer. If someone enables it - he or she definitely uses it.

agzam:
> Not everyone uses Org-mode all the time

I guess I should've accentuated it on "all the time". What I meant to say that I think we reached a point when every single keybinding proposed to be added to a "core" (firmly established) group of bindings should be carefully reviewed and discussed. The project reached its maturity and people rely on Spacemacs to work on real-life projects. Changing things that they already become accustomed to could make people sad. Adding/removing/changing a single keybinding would visually move existing elements around and although to most of us it's not big of a deal, to many it is. I personally witnessed a guy moving back to it's own custom setup, because quote: "Spacemacs guys have changed lots of keybindings"

One could even go further and suggest that `SPC C t` and `SPC C T` are actually belong to `SPC a o` group, but as I already said, let's probably be wise about changing things. Unless it's certainly beneficial to everyone (something like `SPC SPC` for `M-x`)

syl20bnr:
I agree with @agzam. In my mind if we add this key binding it feels like Spacemacs supports a todo buffer in its base distribution and it is not the case, it is in the org layer. This is because `SPC b` is for buffers and does not convey any layer specific meaning, whereas `SPC a o` or `SPC C` are clearly about org (I mean the prefix names are related to org).
Now I see that it is just a matter of feeling and perception since org _is_ in base Emacs but I'm more comfortable to put this binding somewhere in `SPc a o` simply because we already have a key prefix for org, so it is more accurate.
Users can add a key binding behind `SPC o` if they want a more accessible key binding. Or we also have the possibility to put it behind `SPC C`.

d12frosted:
@agzam, good point. I didn't think about this from such perspective. Thanks.

ralesi:
I can see where you are coming from. I figured since this was only opening a buffer it wouldn't be an issue, but I don't have any preference either way, was just trying to integrate the layers better. Should other layers like imenu-list not use the b prefix as well?

syl20bnr:
`SPC b i` is a good case, actually I wrote this when merging the PR with `imenu-list`:

> Very nice, I see this package in default distribution at some point sharing the same window as neotree with window-purpose.
I like SPC b i and it seems that you foresee the changes on SPC j since the same functionality is under SPC j i which matches perfectly :-)

source: https://github.com/syl20bnr/spacemacs/pull/3109#issuecomment-207636217
So it is somewhat consistent with what has been said on this thread.

But now I'm thinking about it, we could reserver `SPC b t` for the todo buffer. I'm not found of `ort/goto-todos` for now but this is a feature we can improve in the future by supporting other "todos framework".

So I'm ok with `SPC b t` for the todo buffer, not so happy with ort though.

ralesi:
I agree,  ort/goto-todos is not ideal,  maybe we can add a baked in one.

TheBB:
I think you might want to look at #6329. I like the `SPC b t` binding though.

syl20bnr:
Do we keep `SPC b t` or `SPC a o P` is enough ?

TheBB:
I made this change in the org-projectile setting. Thanks! Cherry-picked in develop. You can safely delete your branch.

PR #6044, opened 2016-05-13
Set powershell-indent based off tab-width.

ralesi:
This will allow tab spacing to be consistent with the users existing setup.

### DIFF
--- a/layers/+lang/windows-scripts/packages.el
+++ b/layers/+lang/windows-scripts/packages.el
@@ -53,6 +53,8 @@
         (spacemacs|define-text-object "$" "dollarparen" "$(" ")")
         )
       (add-hook 'powershell-mode-hook 'powershell/define-text-objects)
+      ;; set powershell indent based on tab-width
+      (setq powershell-indent tab-width)
       (spacemacs/set-leader-keys
         "asp" 'powershell)
       (spacemacs/set-leader-keys-for-major-mode 'powershell-mode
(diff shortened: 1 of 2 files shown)


StreakyCobra:
Maybe this list can be used to set this? https://github.com/syl20bnr/spacemacs/blob/develop/layers/+distribution/spacemacs-bootstrap/config.el#L13

layers/+lang/windows-scripts/packages.el
```diff
@@ -53,6 +53,8 @@
         (spacemacs|define-text-object "$" "dollarparen" "$(" ")")
         )
       (add-hook 'powershell-mode-hook 'powershell/define-text-objects)
+      ;; set powershell indent to tab-width
+      (setq powershell-indent tab-width)
```
bmag: Doesn't it matter that `tab-width` is buffer-local? There's `(default-value 'tab-width)` if using the default value (set by `setq-default`) makes more sense here. (I don't actually know what's appropriate here, just raising a concern)

ralesi:
@StreakyCobra,  this alist will probably work fine,  I will update this to give it a try.

ralesi:
@StreakyCobra, this alist is used only for evil-shift-width, but I have updated including this change as well.

StreakyCobra:
@ralesi Oh right, but indeed it's nice to had it there too. (And thanks for the key-bindings documentation update on your other PRs, I didn't want to spam you, so doing it here ;-) )

PR #6045, opened 2016-05-13
Provide additional jump options.

ralesi:
Moved `jf` and `jv` to their uppercase versions to be consistent with the rest of Spacemacs.  The lowercase will prompt to find any function or variable.  Also added a library lookup as another option.


StreakyCobra:

ralesi:
Yes,  I know =).  If it seems the PR will go through,  I typically edit the documentation before the final push.

ralesi:
Documentation updated

syl20bnr:
Thank you again :-)
I removed `SPC j L` since I think `SPC f e l` does the same job. Maybe `SPC f e l` should be changed to not be helm specific though, another topic for another PR ;-)

PR #6046, opened 2016-05-13
Add additional custom file actions for ivy.

ralesi:
This provides a `defvar` with custom actions used with ivy find-file based commands.  At startup, we recreate the list to work within the `spacemacs-ivy/counsel-search` command so that we can use these commands with `grep`-like output as well.

### DIFF
--- a/layers/+distribution/spacemacs-base/funcs.el
+++ b/layers/+distribution/spacemacs-base/funcs.el
@@ -868,6 +868,28 @@ Compare them on count first,and in case of tie sort them alphabetically."
                                                (region-end))))
  (evil-end-undo-step))
 
+;; find file functions in split
+(defun spacemacs//display-in-split (buffer alist)
+  "Split selected window and display BUFFER in the new window.
+BUFFER and ALIST have the same form as in `display-buffer'. If ALIST contains
+a split-side entry, its value must be usable as the SIDE argument for
+`split-window'."
+  (let ((window (split-window nil nil (cdr (assq 'split-side alist)))))
+    (window--display-buffer buffer window 'window alist)
+    window))
+
+(defun spacemacs/find-file-vsplit (file)
+  "find file in vertical split"
+  (interactive "FFind file (vsplit): ")
+  (let ((buffer (find-file-noselect file)))
+    (pop-to-buffer buffer '(spacemacs//display-in-split (split-side . right)))))
+
+(defun spacemacs/find-file-split (file)
+  "find file in horizonatl split"
+  (interactive "FFind file (split): ")
+  (let ((buffer (find-file-noselect file)))
+    (pop-to-buffer buffer '(spacemacs//display-in-split (split-side . below)))))
+
 (defun spacemacs//intersperse (seq separator)
   "Returns a list with `SEPARATOR' added between each element
 of the list `SEQ'."
@@ -908,4 +930,3 @@ is nonempty."
   (interactive)
   (when compilation-last-buffer
     (delete-windows-on compilation-last-buffer)))
-
--- a/layers/+completion/ivy/config.el
+++ b/layers/+completion/ivy/config.el
@@ -30,3 +30,19 @@ with options to run in the shell.")
 than this amount.")
 
 (defvar spacemacs--counsel-initial-number-cand 100)
+
+(defvar spacemacs--ivy-file-actions
+  '(("f" find-file-other-frame "other frame")
+    ("w" find-file-other-window "other window")
+    ("v" spacemacs/find-file-vsplit "in vertical split")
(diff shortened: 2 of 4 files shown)


layers/+completion/spacemacs-ivy/packages.el
```diff
@@ -265,6 +300,26 @@ Helm hack."
...
+      (ivy-set-actions
+       'counsel-find-file
+       spacemacs-ivy-file-actions)
+
+      (defun spacemacs-ivy/find-file-vsplit (file)
+        "find file in vertical split"
+        (interactive)
+        (select-window (split-window-right))
+        (find-file file)
+        (balance-windows))
```
bmag: I think we shouldn't call `balance-windows` here, because the user doesn't always want the windows to be balanced (I know I don't).

layers/+completion/spacemacs-ivy/packages.el
```diff
@@ -265,6 +300,26 @@ Helm hack."
...
+      ;; set additional ivy actions
+      (ivy-set-actions
+       'counsel-find-file
+       spacemacs-ivy-file-actions)
+
+      (defun spacemacs-ivy/find-file-vsplit (file)
+        "find file in vertical split"
+        (interactive)
+        (select-window (split-window-right))
+        (find-file file)
```
bmag: Calling `select-window` and then `find-file` assumes that `find-file` always uses the selected window, but it may choose another window if the user customizes (or advises) `find-file` or `switch-to-buffer`. Maybe there's a `find-file*` variant that takes an `action` argument similar to `pop-to-buffer`? If I have time later I will offer an alternative implementation.

layers/+completion/spacemacs-ivy/packages.el
```diff
@@ -265,6 +300,26 @@ Helm hack."
...
+        (balance-windows))
+
+      (defun spacemacs-ivy/find-file-split (file)
+        "find file in horizontal split"
+        (interactive)
+        (select-window (split-window-below))
+        (find-file file)
+        (balance-windows))
```
bmag: Same comments I made for `spacemacs-ivy/find-file-vsplit` apply here as well.

ralesi:
@bmag, I removed the balance-window commands, not quite sure what you mean for your other issue.

bmag:
@ralesi I meant something like this:
```elisp
(defun spacemacs-ivy/find-file-vsplit (file)
  "find file in vertical split"
  (interactive "F")
  (let ((buffer (find-file-noselect file)))
    (pop-to-buffer buffer '(spacemacs/display-in-split (split-side . right)))))

;; this can be placed in spacemacs-base
(defun spacemacs/display-in-split (buffer alist)
  "Split selected window and display BUFFER in the new window.
BUFFER and ALIST have the same form as in `display-buffer'. If ALIST contains
a split-side entry, its value must be usable as the SIDE argument for
`split-window'."
  (let ((window (split-window nil nil (cdr (assq 'split-side alist)))))
    (window--display-buffer buffer window 'window alist)
    window))
```
My point is that user configuration is more likely to break the `select-window`+`find-file` approach. More to the point, there are window-management packages, such as window-purpose (disclaimer: I wrote it), that change the behavior of `switch-to-buffer`. My proposed implementation is less likely to conflict with those packages.

bmag:
Unrelated, I noticed now that the interactive forms are wrong - they should be `(interactive "FFind file vsplit: ")` or something like that.

ralesi:
@syl20bnr ,  I updated this to work with your recent changes to the completion layers.

layers/+distribution/spacemacs-base/funcs.el
```diff
@@ -868,6 +868,28 @@ Compare them on count first,and in case of tie sort them alphabetically."
...
+  "find file in vertical split"
+  (interactive "FFind file (vsplit): ")
+  (let ((buffer (find-file-noselect file)))
+    (pop-to-buffer buffer '(spacemacs//display-in-split (split-side . right)))))
+
+(defun spacemacs/find-file-split (file)
+  "find file in horizonatl split"
+  (interactive "FFind file (split): ")
+  (let ((buffer (find-file-noselect file)))
+    (pop-to-buffer buffer '(spacemacs//display-in-split (split-side . bottom)))))
```
bmag: should be `(split-side . below)` instead of `(split-side . bottom)`

ralesi:
@bmag, Thanks again.  Seems to be working right now.

d12frosted:
I want this PR to be merged. After that I'll share some other actions :-P

syl20bnr:
Thank you again 👍 
I renamed perspective to layout since this is the word we use in spacemacs. See https://github.com/syl20bnr/spacemacs/commit/8f45a94

Issue #5120, opened 2016-02-18
Home buffer at full screen not centered


don-han:
#### Description
The home buffer is not centered when window is maximized. I tested it both on `master` and on `develop`

![image](https://cloud.githubusercontent.com/assets/5600864/13136906/44db2490-d5d2-11e5-8f2d-ad8ba471e9af.png)

#### Expected Behavior
Not sure if this is a feature, but I expected the home buffer to be on the center

#### System Info
- OS: darwin
- Emacs: 24.5.1
- Spacemacs: 0.105.9
- Spacemacs branch: master (rev. 0562f05)
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
((auto-completion :variables auto-completion-return-key-behavior 'complete auto-completion-tab-key-behavior 'cycle auto-completion-complete-with-key-sequence nil)
 better-defaults emacs-lisp markdown syntax-checking version-control
 (latex :variables latex-enable-auto-fill t)
 themes-megapack python)
```

StreakyCobra:
When https://github.com/syl20bnr/spacemacs/pull/4262 will be merged it will be easy to do probably.

TheBB:
I'm leaning towards wontfix for this. I'm not sure starting up in a distraction-free mode would be very useful.

StreakyCobra:
To be clear, from my point of view this should not be the default, but it should be easy to do at user-config level once a distraction-free mode would be merged.

syl20bnr:
Same here, it would feel weird for a lot of users.

a13ph:
Centering might be not the preferred solution - though I fail to see how, tbh, you don't need to activate distraction-free modes for that - but having lots of unused space with left alignment does suck.

StreakyCobra:
I was just saying this would be possible *easily* at user level with a distraction-free mode, not that this should be the Spacemacs solution. If people want to create a PR that does this, please feel free. Also be sure to deal with the arbitrary length of the recent files list.

StreakyCobra:
This has been fixed in `develop` by #6048

mosic:
How can I switch back to the old layout? Centered does look better on small screens, but it's pretty bad on larger ones (especially with release notes visible):

![image](https://cloud.githubusercontent.com/assets/1229422/15291466/6424bc84-1b7e-11e6-8554-080e27077c1a.png)

ralesi:
Are we just looking to change the width of the release notes?

[...]

PR #6048, opened 2016-05-13
Center the startup banner

ralesi:
This centers the banner and button widgets in the spacemacs-buffer based on the initial window size on startup.  Terminal windows should use the pre-defined 75 columns.  In addition, provides a refresh button `C-r` to allow refresh of the buffer if window size changes.

### DIFF
--- a/core/core-spacemacs-buffer.el
+++ b/core/core-spacemacs-buffer.el
@@ -161,14 +161,14 @@ If ALL is non-nil then truly all banners can be selected."
            (spec (create-image banner))
            (size (image-size spec))
            (width (car size))
-           (left-margin (floor (- spacemacs-buffer--banner-length width) 2)))
+           (left-margin (max 0 (floor (- spacemacs-buffer--banner-length width) 2))))
       (goto-char (point-min))
       (insert "\n")
       (insert (make-string left-margin ?\ ))
       (insert-image spec)
       (insert "\n\n")
-      (insert (make-string (floor (/ (- spacemacs-buffer--banner-length
-                                        (+ (length title) 1)) 2)) ?\ ))
+      (insert (make-string (max 0 (floor (/ (- spacemacs-buffer--banner-length
+                                        (+ (length title) 1)) 2))) ?\ ))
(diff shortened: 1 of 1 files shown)


ralesi:
Here is a screenshot:
image

d12frosted:
I am not offended by non-centered home buffer, but your version looks very nice 😸

d12frosted:
The only thing - I think it should auto-refresh right after spacemacs applies initial frame size settings.

StreakyCobra:
Partially related discussion: https://github.com/syl20bnr/spacemacs/issues/5120

syl20bnr:
I like it :-)

ralesi:
@d12frosted, which method are you using to resize windows at start?

d12frosted:
I am setting `dotspacemacs-maximized-at-startup` to `t`, `dotspacemacs-fullscreen-at-startup` and `dotspacemacs-fullscreen-use-non-native` to `nil`. I am on OS X, GNU Emacs 25.

d12frosted:
Also, I think it should not by turned on by default. At least, for me it looks weird in maximised window 😸 Because buttons and icon are centered, while recent files and other widgets are not centered. In tiny window your centering looks very nice. 👏

StreakyCobra:
Agreed upon @d12frosted. I don't have an opinion if it should be default or not, but people would appreciate a way to disable it. This can look really weird on big screen :-)

syl20bnr:
I really like it even on a wide screen, I don't find that the rest of the buffer is not centered, actually what would feel weird is to have everything centered. Only the title feels really slick to me.

@ralesi Thank you! 💜 

Note: I forced the display of warning buffers when installing and loading packages at the bottom of the screen so the home buffer is not messed up. Related commit https://github.com/syl20bnr/spacemacs/commit/1ebbe18

syl20bnr:
Here is a screenshot on a wide screen:

capture d ecran 2016-05-14 a 22 11 38

d12frosted:
:+1: great, thanks

~ d12frosted

PR #6049, opened 2016-05-13
Switch to an alist for startup lists.

ralesi:
Defines an `alist` that is used to define the lists to show in the spacemacs-buffer and specifying maximum list size for each list.  This allows each user to limit the output of long lists (like todo, or projects).

### DIFF
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -302,13 +302,12 @@ tool of the list. Supported tools are `ag', `pt', `ack' and `grep'.")
 specified with an installed package.
 NOT USED FOR NOW :-)")
 
-(defvar dotspacemacs-startup-lists '(recents projects)
-  "List of items to show in the startup buffer. If nil it is disabled.
-Possible values are: `recents' `bookmarks' `projects' `agenda' `todos'.")
-
-(defvar dotspacemacs-startup-recent-list-size 5
-  "Number of recent files to show in the startup buffer. Ignored if
-`dotspacemacs-startup-lists' doesn't include `recents'.")
+(defvar dotspacemacs-startup-lists '((recents  . 5)
+                                    (projects . 7))
+  "Association list of items to show in the startup buffer of the form
+`(list-type . list-size)`. If nil it is disabled.
+Possible values for list-type are:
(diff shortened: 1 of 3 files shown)


syl20bnr:
Is it backward compatible or does it break when people update ?

ralesi:
@syl20bnr,  I created a new variable, so the buffer won't break when people update; however, the old startup list will no longer be respected.  We could add a note if the standard list is still defined in the init file to let people know it is deprecated.

ralesi:
Do we want to make this backward compatible?

syl20bnr:
@smeevil @tomekowal

Hey guys,
I added your flycheck checkers in the elixir layer in develop.
`credo` seems to work fine but `mix compile` and `test result` display no information, since `mix compile` checker is inspired from `test result` checker and they both rely on a utility script in Elixir I wonder if you could help to debug what's going on.

Commits:
- credo: https://github.com/syl20bnr/spacemacs/commit/5b97df8
- mix compile: https://github.com/syl20bnr/spacemacs/commit/dd572af
- test result: https://github.com/syl20bnr/spacemacs/commit/4996f52

Notes:
For `mix compile` the checker is disabled by default since it is not safe, you have to setup the layer variable =elixir-enable-compilation-checking= to t or use a directory local variable (see README.org).

Thank you !

syl20bnr:
Ideally yes, people will be confused if their home buffer is modified when updating. Also would be better to keep the same variable name and make it work for both old and new values.

ralesi:
OK, we can make this backward compatible, but how should we deal with the conflict when a value is defined for recents in the alist and the value already existing in the recent-list-size variable we have in dotspacemacs?  They provide the same functionality.  If the list is set as an alist, use those values?

syl20bnr:
Mmmhh I would deprecate recent-list-size so users using it will get the default size and make startup-list support the both forms.

ralesi:
@syl20bnr, updated to use `string` or `cons` pairs.  Maintains backwards compatibility with previous list and can use interchangably, i.e.

```
   dotspacemacs-startup-list '((recents .  7)
                               bookmarks
                               (projects  . 7))
```

syl20bnr:
Thank you ! 👍 

PR #6053, opened 2016-05-14
[ruby layer] Automatically switch from common Ruby compilation modes to interact with a debugger

dcluna:
This commit adds the 'inf-ruby-auto-enter function to 'compilation-filter-hook, automatically entering inf-ruby-mode when it hits a breakpoint (see the [inf-ruby-manual](https://github.com/nonsequitur/inf-ruby/#manual) for more details).

This might have the side-effect of slightly slowing down compilation modes for other languages, as the hook is run whenever a line is inserted in a comint buffer. We might want to conditionally enable it
via an optional variable.

### DIFF
--- a/layers/+lang/ruby/packages.el
+++ b/layers/+lang/ruby/packages.el
@@ -149,6 +149,10 @@
     ;; setup the hook via an autoload
     :config
     (progn
+      (defun spacemacs//inf-ruby-auto-enter ()
+        "Automatically enters inf-ruby-mode in ruby modes' debugger breakpoints."
+        (add-hook 'compilation-filter-hook 'inf-ruby-auto-enter nil t))
+      (add-hook 'rspec-compilation-mode-hook 'spacemacs//inf-ruby-auto-enter)
       (spacemacs|hide-lighter rspec-mode)
       (dolist (mode '(ruby-mode enh-ruby-mode))
         (spacemacs/set-leader-keys-for-major-mode mode


layers/+lang/ruby/packages.el
```diff
@@ -128,6 +128,7 @@
         (spacemacs/declare-prefix-for-mode mode "mg" "ruby/goto")
         (spacemacs/declare-prefix-for-mode mode "mh" "ruby/docs")
         (spacemacs/declare-prefix-for-mode mode "ms" "ruby/repl")
+        (add-hook 'compilation-filter-hook 'inf-ruby-auto-enter) ;; automatically enters inf-ruby mode from a compilation buffer
```
syl20bnr: You can make the hook local to the current buffer with the last argument of `add-hook`.
So what can be done to limit it to ruby buffer is to wrap the `add-hook` into a function and add this function to `ruby-mode-hook`.

layers/+lang/ruby/packages.el
```diff
@@ -149,6 +149,10 @@
     ;; setup the hook via an autoload
     :config
     (progn
+      (defun spacemacs//inf-ruby-auto-enter ()
+        "Automatically enters inf-ruby-mode in ruby modes' debugger breakpoints."
+        (add-hook 'compilation-filter-hook 'inf-ruby-auto-enter nil t))
+      (add-hook 'rspec-compilation-mode-hook 'spacemacs//inf-ruby-auto-enter)
```
dcluna: @syl20bnr It has to be made local to this hook, but other than that, your suggestion worked great. Thanks!

syl20bnr:
Nice work!
Thank you again :-)

PR #6058, opened 2016-05-16
add common-lisp layer prefix-names for which-key

gilch:
Add the `SPC m` prefix key names to `common-lisp` layer for `which-key`.


syl20bnr:
Thank you !

Issue #6069, opened 2016-05-17
Haskell-mode debug keys don't work


user471:
#### Reproduction guide

- Start Emacs
- open haskell file
- run debug `, d d`

*Observed behaviour:*

- key `g - refresh` doesn't work. `g` opens `which-key` window

*Expected behaviour:*

- key `g` should work

#### System Info
- OS: windows-nt
- Emacs: 24.5.1
- Spacemacs: 0.105.20
- Spacemacs branch: develop (rev. 121ee00)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
((auto-completion :variables auto-completion-enable-help-tooltip t auto-completion-enable-sort-by-usage t)
 git emacs-lisp haskell colors javascript csharp dash org purescript syntax-checking markdown semantic elm ranger)
```

d12frosted:
You can use `, d r` to call refresh while in `haskell-mode` (not in debug buffer). 

I believe that it's just due to vim editing style. I believe that haskell debug mode should be evilified (or how is it called)?

user471:
>You can use , d r to call refresh while in haskell-mode

Yes, but I think pressing one key, especially for something like step, would be a better option.

>I believe that it's just due to vim editing style. I believe that haskell debug mode should be evilified (or how is it called)? 

Couldn't Spacemacs just ignore specific window and don't interrupt keys on it?

d12frosted:
Well, you are in evil state - that's why it interpreters keys differently.
That's why I think that buffer should be evilified (if I understand what it
means). You can switch to insert state in that buffer and use `g`.

~ d12frosted

d12frosted:
But that makes sense. See #6072. It will allow you to use `g` for refresh (and some other short bindings).

d12frosted:
Fixed with released of Spacemacs v0.200. Let us know if you still have any problems with this issue.

PR #6072, opened 2016-05-17
setup keybindings in haskell-debug-mode

d12frosted:
Following #6069 

Ideal solution is to evilify that buffer, but I'd leave it to someone who uses evil part of Spacemacs 😸

### DIFF
--- a/layers/+lang/haskell/README.org
+++ b/layers/+lang/haskell/README.org
@@ -20,13 +20,14 @@
  - [[Key bindings][Key bindings]]
    - [[Documentation][Documentation]]
    - [[Debug][Debug]]
+   - [[Debug Buffer][Debug Buffer]]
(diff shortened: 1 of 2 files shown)


d12frosted:
Ok, evilification is not hard, so I updated PR.

TheBB:
The typography layer packages can also be considered for moving to prose.

TheBB:
Ugh, wrong PR.

user471:
Wouldn't it be better if you get keys directly from https://github.com/haskell/haskell-mode/blob/master/haskell-debug.el#L76 ? In this case it will automatically update if they add new keys.

syl20bnr:
Thank you ! 👍 
I unified the key bindings in commit https://github.com/syl20bnr/spacemacs/commit/a1d60c0

PR #4790, opened 2016-01-25
haskell README: Fix proposed ghc-mod workaround

ilyagr:
In my setup, the command

```
(spacemacs/set-leader-keys-for-major-mode 'haskell-mode
        "mht"  'ghc-show-type)
```
  very confusingly makes the actual key combination by `<SPC> m m h t` instead of `<SPC> m h t`. I think that the leading `m` is assumed. I also added a command for `mhi` since `mhi` and `mht` seem to go together. This works on my setup.

Issue #4770, opened 2016-01-23
Haskell: problems with 'mht' and 'mhi'


ilyagr:
I think that the resolution of PR #3197 in commit b27d6bd94e83e93b23c92c0b9ce19a387fdd8bd2 is causing problems on my system. After a fresh install (no haskell-specific configuration in .spacemacs) and opening a .hs file, both `<SPC> mht` and `<SPC> mhi` are bound to the function `haskell-mode-enable-process-minor-mode`, which simply prints a (rather unhelpful) error message.

It seems that for the keybindings to work, one must enable the "process-minor-mode" in such a way that it happens before the keybinding of `<SPC> mht` happens. I'm not sufficiently versed in (space)macs to know how to do it in my .spacemacs, exactly. Also, it seems that if I ever changed the "process-minor-mode", the keybindings would not get updated as they should.

d12frosted:
Please, share you configuration for haskell layer. If you are using `ghci-ng` or not using `ghc-mod` then it makes sense. In order to use `mhi` or `mht` you need an associated interactive session.

ilyagr:
I do not have any Haskell-specific config for spacemacs, beyond adding the Haskell layer. I have stack and latest ghc-mod in my path, no ghci-ng.

The main problem is that creating an interactive session does not rebind 'mht' to a useful function, it keeps just printing the same error message. A workaround would be to setup ghc-mod or an interactive session before the lines in the linked commit are executed, but I'm not sure what the proper way to do that is.

nebtrx:
I'm having the same issue. So far now I've found a workaround but it's kind of annoying: After open a hs buffer resync dotspacemacs config (M-m fe R) and that's it, all comes back to normal.

I'm using ghc-mod and this is my spacemacs haskell related config:

...
 (haskell :variables
              ;;haskell-enable-hindent-style "chris-done"

              haskell-enable-ghci-ng-support nil
              ;;haskell-process-type 'ghci
              haskell-enable-shm-support t)

...
...
(defun dotspacemacs/user-config ()
  (spacemacs/set-leader-keys-for-major-mode 'haskell-mode "mht"  'ghc-show-type)

  ;; always delete trailing space automatically on save
  (add-hook 'before-save-hook 'delete-trailing-whitespace)
  ;; automatic file reloading if it changes externally
  (global-auto-revert-mode t)
  (add-hook 'haskell-mode-hook 'haskell-indent-mode)
[...]

PR #6074, opened 2016-05-17
Fix proposed ghc-mod workaround in haskell layer README

hasufell:
"m" is extraneous.

Also see #4770 and #4790


syl20bnr:
Thank you ! 👍 

PR #6076, opened 2016-05-18
Add make-frame shortcut, SPC w F

JP-Ellis:
Placed it under the `SPC w` prefix since the related `other-frame` is
set to `SPC w o`.


StreakyCobra:
> [ALWAYS document your new keybindings or keybindings changes inside the relevant documentation file. It should be the layer’s README.org file for layer’s keybindings, or DOCUMENTATION.org for general Spacemacs key bindings.](https://github.com/syl20bnr/spacemacs/blob/develop/CONTRIBUTING.org#contributing-a-keybinding)

:wink:

JP-Ellis:
Sorry, my bad!  I initially couldn't find where `SPC w o` was documented, so I thought that it went undocumented (!)... but turns out I just missed it :/

syl20bnr:
Thank you !

PR #6079, opened 2016-05-18
Add Cedet layer

declanqian:
I have been trying to evaluate the possibilities of using Emacs as full function IDE. And the 
video [The Editor of a Lifetime](https://www.youtube.com/watch?v=VADudzQGvU8) mentioned
[CEDET](http://cedet.sourceforge.net/), which is still a active project recently add some android development supports. And I found it useful. 

And I do search from the issues. Some of them mentioned CEDET. And there is already ~Semantic layer~ which use the build-in CEDET. I think this won't conflict with it, since basically it just remove the build-in CEDET from the ~load-path~ and replace it with the new one.

What do you guys think, is is worthy to include this in Spacemacs?


layers/+emacs/cedet/README.org
```diff
@@ -0,0 +1,36 @@
...
+You can clone the git repo with:
+
+#+BEGIN_SRC bash
+ git clone git://git.code.sf.net/p/cedet/git cedet 
+#+END_SRC
+
+Also I host a mirror on [[https://github.com/declanqian/cedet][Github]].
+
+If the installation directory of CEDET is not in Emacs’ load path, you can set
+the layer variable =mu4e-installation-path=, for example:
```
TheBB: cedet-installation-path…

layers/+emacs/cedet/packages.el
```diff
@@ -0,0 +1,39 @@
...
+
+(defun cedet/init-cedet-contrib()
+  "Initialize my package"
+  (use-package cedet-contrib-load
+    :defer t
+    :commands semantic-scala-cedet-support
+    :init
+    (progn
+      (unless (featurep 'cedet-devel-load)
+              (require 'cedet-devel-load)))
```
TheBB: I'm not sure how CEDET works exactly but we try to avoid loading packages on startup whenever possible.

Also `(require 'cedet-devel-load)` is a no-op if it's loaded already so you don't have to check for `featurep`. You don't need the `progn` either since it's just a single sexp.

layers/+emacs/cedet/README.org
declanqian: Should I change this variable name? It is the cedet git repo dir name.

layers/+emacs/cedet/packages.el
declanqian: Yeah, what should I do if the ~cedet/init-cedet-contrib~ rely on the ~cedet/init-cedet~. Is there 
some example codes?

layers/+emacs/cedet/README.org
TheBB: No but you shouldn't call it "mu4e" in the documentation. ;-)

tuhdo:
This is redundant with `semantic` layer. You should add these changes to `semantic` instead.

declanqian:
Maybe not to include cedet to Spacemacs. Close this one.

PR #6085, opened 2016-05-18
Add beginners tutorial

Nikolaiam:
Spacemacs has good documentation for the more computer-savvy, but
for non-programmers and those who have never heard of Emacs or Vim
before the process of installation and initial setup can be a bit
overwhelming, especially as a the information has to be collected from
multiple sources.

This detailed tutorial walks the user through the
setup process and shows the basic functionality of the program. It is
meant to serve as the one document one needs to get up-and-running with
Spacemacs.

Feedback is very welcome!

### DIFF
--- a/doc/BEGINNERS_TUTORIAL.org
+++ b/doc/BEGINNERS_TUTORIAL.org
@@ -1,0 +1,297 @@
+#+TITLE: Beginners tutorial
+
+* Beginners tutorial                                      :TOC_4_gh:noexport:
+ - [[#why-spacemacs][Why Spacemacs?]]
+ - [[#install][Install]]
+   - [[#1-install-emacs][1. Install Emacs]]
+   - [[#2-install-git][2. Install Git]]
+   - [[#3-install-spacemacs][3. Install Spacemacs]]
+     - [[#note-for-windows-users][Note for Windows users]]
+   - [[#4-install-the-default-font][4. Install the default font]]
+   - [[#5-open-spacemacs-and-choose-default-editing-style-][5. Open Spacemacs and choose default editing style ]]
+ - [[#getting-started][Getting started]]
+   - [[#keybinding-notation][Keybinding notation]]
+   - [[#modal-text-editing---why-and-how-][Modal text editing - why and how? ]]
+   - [[#start-the-vim-tutorial-][Start the Vim tutorial ]]
+   - [[#using-the-spacebar-to-launch-commands][Using the spacebar to launch commands]]
+   - [[#buffers-windows-and-frames][Buffers, windows and frames]]
+ - [[#configuring-spacemacs][Configuring Spacemacs]]
+   - [[#adding-language-support-and-other-features-using-layers][Adding language support and other features: using layers]]
+   - [[#changing-the-colour-theme][Changing the colour theme]]
+   - [[#starting-maximized][Starting maximized]]
+   - [[#quitting-restarting][Quitting (restarting)]]
+ - [[#additional-features-tips-and-troubleshooting][Additional features, tips and troubleshooting]]
+   - [[#org-mode][Org mode]]
+   - [[#version-control---the-intelligent-way][Version control - the intelligent way]]
+   - [[#daemon-mode-and-instant-startup-linux-][Daemon mode and instant startup (Linux) ]]
+   - [[#swap-caps-lock-and-esc-keys-on-your-keyboard][Swap caps lock and esc keys on your keyboard]]
+   - [[#faq-and-troubleshooting][FAQ and troubleshooting]]
+
+* Why Spacemacs?
+- Unparallelled text and structure editing for all types of writing tasks:
+  creative writing, blogging, note-taking, todo-lists, scientific papers... 
+- Powerful modes for programming in dozens of programming languages
+- Deeply customizable yet beginner-friendly
+ 
+* Install
+ Spacemacs is a beginner-friendly and powerful extension of a popular text
+ editor called Emacs. To install Spacemacs you need to first install base Emacs
+ and then download and install the Spacemacs extension files, which is easiest
+ done by using a program called Git. The steps are easy and outlined below.
+
+** 1. Install Emacs
+Follow the instructions in https://github.com/syl20bnr/spacemacs#prerequisites.
+
+** 2. Install Git
+Go to https://git-scm.com/downloads and install Git on your platform.
+
+** 3. Install Spacemacs
+Open a terminal or command prompt, paste the following code to it:
+
+#+BEGIN_SRC sh
+git clone https://github.com/syl20bnr/spacemacs ~/.emacs.d
+#+END_SRC
+
+Press enter to execute the code and the program you installed in the previous step,
+Git, will download the Spacemacs extension files.
+
+*** Note for Windows users
+If you use windows, you have to modify the git command by inserting the correct
+path to your .emacs.d folder. The dot before the folder means that it is hidden,
+so you'll have to search for hidden files to find the folder. When you have
+found the folder, substitute the original path with the correct one. The proper
+code would look something like this:
+
+#+BEGIN_SRC sh
+git clone https://github.com/syl20bnr/spacemacs /path/to/your/.emacs.d
+#+END_SRC
+
+** 4. Install the default font
+The default font used by Spacemacs is Source Code Pro by Adobe. It is
+recommended to [[https://github.com/adobe-fonts/source-code-pro][install]] it on your system to ensure correct visual representation.
+
+** 5. Open Spacemacs and choose default editing style 
+Open Spacemacs by clicking the Emacs icon in your applications menu. The first
+time Spacemacs launches, it will load and install packages and prompt you for
+your preferred editing style. You have two options: Vim ("Among the stars aboard
+the Evil flagship") and Emacs. If you haven't used Emacs before or are unsure
+about the differences of the editing styles, we recommend selecting the default,
+Vim, by pressing enter. Using this configuration is introduced more thoroughly
+in the next section. If you are already familiar with Emacs or do not plan to
+switch into modal editing style, select Emacs with the left and right arrow keys. 
+There is also a third option, "Hybrid", for more advanced users willing to use
+both styles. All of these choices can be easily changed later by editing the
+dotspacemacs-editing-style variable in the dotfile (see [[#configuring-spacemacs][Configuring Spacemacs]]),
+so if modal editing does not sweep you away, you can switch to the Emacs style
+later.
+
+Next, you will be prompted for the distribution you would like to start with.
+The standard distribution is recommended, press enter to select it. 
+
+Now Spacemacs will download and install required packages. This will take some
+minutes depending on your connection. After everything is installed (you will
+see the text "n packages loaded in x s" appear in the list under the Spacemacs
+logo), restart Spacemacs. 
+
+Now your installation process is complete, congratulations! For troubleshooting,
+see the last section.
+
+* Getting started
+** Keybinding notation
+The power of Spacemacs lies in its efficient keybindings. Because it is built on
+Emacs, we will use Emacs conventions for keybinding notation. The most important
+modifier keys are:
+
+~SPC~ = ~Space~, used as the leader key in Vim editing style.
+~C-~ = ~Ctrl~
+~M-~ (for "meta") = ~Alt~ 
+~S-~ = ~Shift~
+
+The modifier keys can be used either in a sequence or as key chords by pressing
+two keys at the same time. ~SPC 1~ is notation for a key sequence and means
+pressing ~Space~ first and pressing ~1~ after it. Key chords are notated by
+writing a ~-~ between the keys. Thus ~C-c~ means pressing ~Ctrl~ and the letter ~c~ 
+simultaneously. Key chords and sequences can also be combined: ~C-c a~ means
+"First press ~Ctrl~ and ~c~ simultaneously, then press ~a~". ~C-c C-a~ means
+"First press ~Ctrl~ and ~c~ simultaneously, then press ~Ctrl~ and ~a~
+simultaneously".
+
+This document assumes you chose the "Vim" editing style and notates accordingly.
+If you chose the Emacs editing style, just substitute ~SPC~ with ~M-m~ in all
+the commands that begin with ~SPC~.
+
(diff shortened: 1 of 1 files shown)


doc/BEGINNERS_TUTORIAL.org
```diff
@@ -0,0 +1,189 @@
...
+called "frames". A frame is what pops up when you launch Spacemacs from your
+desktop shortcut. A frame contains windows and buffers.
+
+Windows in spacemacs are the visual spaces a frame is divided into. The default
+is one, but windows can be split to allow editing multiple files in one frame.
+Let's try this. Press SPC to bring up the menu. You can see different letters
+having different submenus associated with them, usually with a mnemonic for
+easier recall. The letter w is assigned for "windows": press it. A new menu
+opens with further options. We will write the character / (that is, shift and
+seven) to split the currently active window vertically into two.
```
cpaulik: That is only true for certain keyboard layouts.

doc/BEGINNERS_TUTORIAL.org
```diff
@@ -0,0 +1,189 @@
...
+the (hidden) .git folder in the same folder the tracked file(s) are in. You can
+make further commits the same way.
+** Daemon mode and instant startup (Linux) 
+Emacs can be used in daemon mode: a daemon runs in the background and launches
+clients. This way new frames launch instantly without delay.
+https://www.emacswiki.org/emacs/EmacsAsDaemon tells more about the daemon
+and how to set it to launch automatically on startup.
+** Swap caps lock and esc keys on your keyboard
+This is useful outside of Spacemacs as well!
+** Troubleshooting
```
cpaulik: Maybe it would be best to link to the FAQ here.

doc/BEGINNERS_TUTORIAL.org
```diff
@@ -0,0 +1,189 @@
...
+Now you should see two windows of this tutorial, and the one on the left should
+be active, as can be seen from the modeline in the bottom or by moving the
+cursor around using the navigation keys. This isn't very useful, as we
+would probably want to see a different file on the right.
+
+First, activate the window on the right with SPC 2. Now that the window on the 
+right is active, we can open a different buffer for a different file. We'll use
+the scratch buffer, which can be used like a notepad. Be warned, unlike other buffers it doesn't prompt you 
+whether you want to save the changes you've made when quitting Spacemacs!
+Press SPC b to open the buffers menu and then switch to the scratch buffer. Now
```
syl20bnr: missing the `s` key to actually switch to the scratch buffer.

doc/BEGINNERS_TUTORIAL.org
```diff
@@ -0,0 +1,189 @@
...
+keeping notes, maintaining todo lists, planning projects, and authoring
+documents with a fast and effective plain-text system", but this gives only a
+small inkling of its versatility. If you do any kind of writing at all, chances
+are that Org mode will make it easier and more fun. This tutorial was written in
+Org mode.
+
+Install the Org layer and open this tutorial. Press shift-TAB repeatedly and observe that
+this cycles the visibility of the contents of different headlines. Press t in
+normal mode and observe that you can add TODO tags on headlines. Press M-k
+(Alt-k, M stands for "modifier") or M-j in normal mode and see how you can
```
syl20bnr: M stands for meta.

doc/BEGINNERS_TUTORIAL.org
```diff
@@ -0,0 +1,189 @@
...
+Install the git layer, restart Spacemacs and open a file you want to version
+control. You can check the status of your file by pressing SPC g s. Select the
+folder your file is in. You will be prompted whether you want to create a
+repository in the folder. Select yes. You will see a list of "Untracked files":
+navigate to the file you want to track and press s to "stage changes". You might
+be prompted to save the file: save it if necessary. Now the new file needs to be
+commited: press c and c again. Two windows pop up: one showing the changes
+you've made since the last edit (in this case, the whole document) and another
+prompting for a commit message. Write "Initial commit", press ESC to exit back
+to normal mode and press :wq to write and quit the commit message.
```
syl20bnr: instead of `:wq` most people use `SPC m c` to confirm and `SPC m a` to abort or `C-c C-c` and `C-c C-k`.

bmag:
I like it. A few notes:
- key shortcuts look nicer if you format them. e.g. `~SPC b b~` instead of `SPC b b`
- absolute beginners don't know how to read Emacs key notation: `SPC` is space, `C-` is Ctrl, `M-` is Alt (or something else on Mac?), `s-` is Super, etc. `C-c a` means "Ctrl+c, then a", `C-c C-a` means "Ctrl+c, then Ctrl+a", etc. I couldn't find a good external resource that explains it, although I'm sure I've read one in the past.

doc/BEGINNERS_TUTORIAL.org
```diff
@@ -0,0 +1,184 @@
...
+ and install the Spacemacs extension files, which is easiest done by using a program
+ called Git. The steps are easy and outlined below.
+** 1. Install Emacs
+Follow the instructions in https://github.com/syl20bnr/spacemacs#prerequisites.
+** 2. Install Git
+Go to https://git-scm.com/downloads and install Git on your platform.
+** 3. Install Spacemacs
+Open a terminal or command prompt, paste the following code to it:
+
+git clone https://github.com/syl20bnr/spacemacs ~/.emacs.d
```
d12frosted: Should be wrapped in `BEGIN_SRC` as:

```
#+BEGIN_SRC shell
$ git clone https://github.com/syl20bnr/spacemacs ~/.emacs.d
#+END_SRC

File: doc/BEGINNERS_TUTORIAL.org [hunk 7]
```diff
@@ -0,0 +1,184 @@
...
+  scientific papers... 
+- Powerful modes for programming in dozens of programming languages
+- Deeply customizable yet beginner-friendly
+* Installation
+ Spacemacs is a beginner-friendly and powerful distribution of a popular text
+ editor called Emacs. To install Spacemacs you need to first install base Emacs and then download
+ and install the Spacemacs extension files, which is easiest done by using a program
+ called Git. The steps are easy and outlined below.
+** 1. Install Emacs
+Follow the instructions in https://github.com/syl20bnr/spacemacs#prerequisites.
```
bmag: I think we should point out that Windows users need Gnu-TLS support for HTTPS to work (although it's possible to use HTTP instead). The recommended 64-bit build has it (http://emacsbinw64.sourceforge.net/), but the official 32-bit build doesn't (https://www.gnu.org/software/emacs/download.html).
If necessary, Gnu-TLS can be installed from http://www.gnutls.org/download.html. To check if Emacs supports Gnu-TLS:
```
M-: (gnutls-available-p)
```

doc/BEGINNERS_TUTORIAL.org
```diff
@@ -0,0 +1,184 @@
...
+** 2. Install Git
+Go to https://git-scm.com/downloads and install Git on your platform.
+** 3. Install Spacemacs
+Open a terminal or command prompt, paste the following code to it:
+
+git clone https://github.com/syl20bnr/spacemacs ~/.emacs.d
+
+Press enter to execute the code and the program you installed in the previous step,
+Git, will download the Spacemacs extension files.
+** 4. Install the default font
```
d12frosted: Very good point. 👍 💯

doc/BEGINNERS_TUTORIAL.org
bmag: I don't think `~/.emacs.d` works in Windows (unless you use cygwin or similar). Easiest way I found to discover the real `.emacs.d` path on windows is to open `~` in Dired: `M-x dired ~/`

doc/BEGINNERS_TUTORIAL.org
```diff
@@ -0,0 +1,184 @@
...
+The default font used by Spacemacs is Source Code Pro by Adobe. It is
+recommended to [[https://github.com/adobe-fonts/source-code-pro][install]] it on your system to ensure correct visual representation.
+** Done!
+Open Spacemacs by clicking the Emacs icon in your applications menu. The first
+time Spacemacs launches, it will ask you for your preferred editing style. We
+recommend selecting the default option, "Evil", more of which is talked about in
```
d12frosted: I would recommend either `evil` or `emacs` based on user background. Emacs users trying Spacemacs for the first time without any vim background should not use `evil` unless they are here for `evil` part. 

I think the main point I would like to make here is that user might choose any editing style based on his preference and background. `hybrid` style is generally for advanced users. **But** it's very important to say that any decisions made here can be easily changed later on.

doc/BEGINNERS_TUTORIAL.org
```diff
@@ -0,0 +1,184 @@
...
+the next section. For troubleshooting, see the last section.
+* Getting started
+** Modal text editing - why and how?
```
bmag: Not usre, but maybe we should note that we also support a non-modal editing style - `emacs` (with regular Emacs key bindings)

doc/BEGINNERS_TUTORIAL.org
d12frosted: Ah, I don't know how it works on Windows (never really used it oO).

My point was about document structure :)

d12frosted:
General advice - don't forget to wrap inline code and key bindings with tilde like `~code~`. Also you should wrap code blocks with following:

```
#+BEGIN_SRC lang-name

#+END_SRC
```

It renders beautifully in `org-mode` and on GitHub Pages.

doc/BEGINNERS_TUTORIAL.org
d12frosted: YES, YES and YES! Not everyone use Spacemacs for `evil` part. Remove `emacs` editing style and you'll make it impossible for me to use Spacemacs.

d12frosted:
Also, please read and follow [these conventions](https://github.com/syl20bnr/spacemacs/blob/develop/doc/CONVENTIONS.org#writing-documentation) about writing documentation. 😸

doc/BEGINNERS_TUTORIAL.org
```diff
@@ -0,0 +1,184 @@
...
+The modal editing features of Spacemacs originate from a text editor called VI,
+and thus the modal editing tutorial is called eVIl tutor. Press SPC h T (that
+is, the spacebar followed by h and T) to familiarize yourself with 
+modal editig. 
+** Using the spacebar to launch commands
+Now that you are familiar with writing and editing text it is time to put the "Space"
+into Spacemacs. Because the spacebar is the most accessible key on the keyboard
+and is pressed by the strongest fingers (the thumbs), it is a natural choice for
+launching commands. You can think of it as the start menu of Spacemacs.
+** Buffers, windows and frames
```
d12frosted: Very nice to bring up this topic. 💯

doc/BEGINNERS_TUTORIAL.org
```diff
@@ -0,0 +1,184 @@
...
+Open a terminal or command prompt, paste the following code to it:
+
+git clone https://github.com/syl20bnr/spacemacs ~/.emacs.d
+
+Press enter to execute the code and the program you installed in the previous step,
+Git, will download the Spacemacs extension files.
+** 4. Install the default font
+The default font used by Spacemacs is Source Code Pro by Adobe. It is
+recommended to [[https://github.com/adobe-fonts/source-code-pro][install]] it on your system to ensure correct visual representation.
+** Done!
```
bmag: Almost done. Part of the installation process (IMO) is to run Spacemacs for the first time, choose editing style and distribution (and helm/ivy?), then wait for Spacemacs to download lots of packages. After that, restart Spacemacs (to get a clean state) and only then the installation is complete.

doc/BEGINNERS_TUTORIAL.org
```diff
@@ -0,0 +1,184 @@
...
+and is pressed by the strongest fingers (the thumbs), it is a natural choice for
+launching commands. You can think of it as the start menu of Spacemacs.
+** Buffers, windows and frames
+Because of its roots in Emacs, which was developed in the 80's
+before the advent of modern graphical user interfaces, Spacemacs has 
+a different name of what we normally call "windows": in Spacemacs these are
+called "frames". A frame is what pops up when you launch Spacemacs from your
+desktop shortcut. A frame contains windows and buffers.
+
+Windows in spacemacs are the visual spaces a frame is divided into. The default
```
d12frosted: > Windows in spacemacs

First of all, 'Spacemacs' and not 'spacemacs' 😸 

Also, I think it's better to use 'Emacs' instead of 'Spacemacs' when you are talking about general behaviour or concepts that are not really related to Spacemacs. The main point is to avoid confusion about what Spacemacs is.

doc/BEGINNERS_TUTORIAL.org
```diff
@@ -0,0 +1,184 @@
+* Why Spacemacs?
+- Unparallelled text and structure editing for all types of writing tasks: creative writing, blogging, note-taking, todo-lists,
+  scientific papers... 
+- Powerful modes for programming in dozens of programming languages
+- Deeply customizable yet beginner-friendly
+* Installation
+ Spacemacs is a beginner-friendly and powerful distribution of a popular text
```
d12frosted: > Spacemacs is ... distribution

I wonder if we should call it 'distribution'. Because people often confuse Spacemacs with replacement for Emacs, while it's actually extension, a set of configurations. But I am not native English speaker, so that might be my problem in understanding what 'distribution' actually means.

Nikolaiam: I think you're right, extension is a better description of what Spacemacs is!

doc/BEGINNERS_TUTORIAL.org
Nikolaiam: I'm somewhat unfamiliar with this topic, but do you think this information is best placed in the beginners tutorial? If I understood correctly, the recommended build has this and http can be used, so by following these instructions, no problems should arise? I'm just thinking that the more concise the instructions are, the better, and maybe this information would find a better place [https://github.com/syl20bnr/spacemacs#prerequisites](here)?

Nikolaiam:
After the newest amendments the Travis CI build failed and I'm at a total loss why. Any suggestions on what the problem is?

syl20bnr:
This is my fault, will fix this evening.

syl20bnr:
Thank you ! 👍 

PR #6089, opened 2016-05-19
Fixes #6051, checks for existence of git-gutter.

### LINKED ISSUE #6051: git gutter toggle default off not working
I'm have the following layer:
```
     (version-control
         :variables
            version-control-global-margin nil
      )
```

But this does not seem to work, in that <space>Td gives:
```
Symbol's value as variable is void: git-gutter+-mode
```

When i set `version-control-global-margin t`, toggling the margin works fine

ralesi:
Non-global vcs margin's are not autoloaded in git-gutter and git-gutter+.  This checks for it's existence when `version-control-global-margin` is nil.


syl20bnr:
Thank you !

Issue #5751, opened 2016-04-11
Move autogenerated package list out of .spacemacs


necto:
Hello,
I keep my `.spacemacs` under VCS, and it is annoying that each time when I change anything, or the Spacemacs gets updated, it regenerates the package plain list:

```
 '(package-selected-packages
   (quote
    (powerline spinner s parent-mode projectile pkg-info epl request flx magit-popup git-commit with-editor smartparens iedit highlight pos-tip yasnippet packed dash avy helm-core async auto-complete utop ocp-indent merlin tuareg caml magit color-theme-sanityinc-solarized marmalade material-theme bracketed-paste anzu helm xterm-color shell-pop multi-term eshell-prompt-extras esh-help smeargle orgit mmm-mode markdown-toc markdown-mode magit-gitflow helm-gitignore helm-company helm-c-yasnippet gitignore-mode gitconfig-mode gitattributes-mode git-timemachine git-messenger git-gutter-fringe+ git-gutter-fringe git-gutter+ git-gutter gh-md evil-magit diff-hl company-statistics company-quickhelp company auto-yasnippet ac-ispell zenburn-theme monokai-theme solarized-theme ws-butler window-numbering volatile-highlights vi-tilde-fringe spaceline smooth-scrolling restart-emacs rainbow-delimiters popwin popup persp-mode pcre2el paradox page-break-lines open-junk-file neotree move-text macrostep lorem-ipsum linum-relative leuven-theme info+ indent-guide ido-vertical-mode hungry-delete hl-todo highlight-parentheses highlight-numbers highlight-indentation help-fns+ helm-themes helm-swoop helm-projectile helm-mode-manager helm-make helm-flx helm-descbinds helm-ag google-translate golden-ratio flx-ido fill-column-indicator fancy-battery expand-region exec-path-from-shell evil-visualstar evil-tutor evil-surround evil-search-highlight-persist evil-numbers evil-nerd-commenter evil-mc evil-matchit evil-lisp-state evil-indent-plus evil-iedit-state evil-exchange evil-escape evil-args evil-anzu eval-sexp-fu elisp-slime-nav define-word clean-aindent-mode buffer-move auto-highlight-symbol auto-compile aggressive-indent adaptive-wrap ace-window ace-link ace-jump-helm-line quelpa package-build use-package which-key bind-key bind-map evil spacemacs-theme)))
```

As it disturbs my flow and leads to meaningless changes in my configuration commits.
[...]

PR #6090, opened 2016-05-19
Discussion: Move customized settings to custom.el

ralesi:
As discussed in #5751, proposes to move customized settings in a file within the /.cache directory.  One thing missing in this PR is some indication that the current mess at the bottom of the dotspacemacs file can be deleted by the user.

### DIFF
--- a/core/core-load-paths.el
+++ b/core/core-load-paths.el
@@ -45,10 +45,17 @@
 (defconst user-home-directory
   (expand-file-name "~/")
   "User home directory (~/).")
+
 (defconst pcache-directory
   (concat spacemacs-cache-directory "pcache/"))
 (unless (file-exists-p spacemacs-cache-directory)
-    (make-directory spacemacs-cache-directory))
+  (make-directory spacemacs-cache-directory))
+
+(setq custom-file
+      (concat spacemacs-cache-directory "custom.el"))
+(unless (file-exists-p custom-file)
+  (with-temp-buffer (write-file custom-file))
+  (custom-save-all))
 
 (defconst user-dropbox-directory
   (expand-file-name (concat user-home-directory "Dropbox/"))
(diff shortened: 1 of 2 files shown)


syl20bnr:
I have to think about it. If someone can give a summary of the cons of the current method I would gladly appreciate it.
One thing I like about the current method of having everything in the dotfile is that it makes it simpler for things like the Nix package manager to install spacemacs but I'm far to be an expert with this package manager (cc @travisbhartwell and @CestDiego). OTOH we support also `spacemacs.d` directory and with this method having a `custom.el` is easy.
Maybe we should create a `˜/.spacemacs-custom.el` ? It would be only `custom.el` in `spacemacs.d`. `spacemacs-custom.el` maybe confusing though, we could name it `.emacs-custom.el` but then people may believe it is an Emacs standard thing...

ralesi:
@syl20bnr, for the most part, spacemacs doesn't make use the customization system, so another benefit would be that we specify when the custom file is loaded in the spacemacs.

ralesi:
@syl20bnr, one con I have found in the current method, is that some packages make use of defcustom to define variables as functions, in these cases they sometimes are incompatible with defer loading.  `org-trello` is one example. One benefit of controlling the time of loading the custom settings is that the user can incorporate fixes in either user-init and user-config depending on the need.  In the current setup, I don't think either functions are run prior to the evaluation of custom settings.

syl20bnr:
@nixmaniack indeed this would be great to control when to load the settings (I guess we would load them between user-init and user-config). But we cannot put the custom file in .cache because it may contain real configuration as important as the config in .spacemacs.

I thought about a crazy thing à la @JAremko ;-) we could wrap them in a `defun` at the bottom of the .spacemacs file. The trick would be to have a smart and clean enough function to wrap them on the fly before actually `load` the file. Opening it in a temp buffer and using `sexp` navigation and search functions should do the trick cleanly. For the name of the function I propose `dotspacemacs/emacs-custom-settings`.

ralesi:
@syl20bnr, does the latest custom settings changes invalidate this PR?

PR #6095, opened 2016-05-19
fix #6084 space-doc-mode bugs

### LINKED ISSUE #6084: space-doc-mode bugs
I decided to test some edge cases that I overlooked before and I found that:
 - [This code](https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Bspacemacs/spacemacs-org/local/space-doc/space-doc.el#L290) should check  if the cookie is actually bound.

 - [And this one](https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Bspacemacs/spacemacs-org/local/space-doc/space-doc.el#L109) should check if `(match-string 4)` - *emphasized region* is one or more `*`. Sometimes  org-mode emphasizer matches `***` as bold `*`  so we need to check for it and do nothing.

JAremko:
Add necessary checkers.

### DIFF
--- a/layers/+spacemacs/spacemacs-org/local/space-doc/space-doc.el
+++ b/layers/+spacemacs/spacemacs-org/local/space-doc/space-doc.el
@@ -105,7 +105,12 @@ appearances in the current buffer. The function uses
 
   ;; `org-do-emphasis-faces' returns non-nil value when it
   ;; found a region to emphasize.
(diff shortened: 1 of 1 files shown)


JAremko:
@syl20bnr  :shipit:

syl20bnr:
Thank you ! 👍 

PR #6103, opened 2016-05-20
[haskell] fix type and info hey bindings

d12frosted:
At some point it became broken again. I noticed it by accident as I use `C-c C-t` and `C-c C-i` whenever I need to get type / info. By 'broken' I mean that those bindings call function that shows error about misconfiguration, while `C-c C-t` and `C-c C-i` are properly set. 

---

I left the part where it sets `, h t` and `, h i` to whatever is set to `C-c C-t` and `C-c C-i` respectively just to make sure that it shows error when something goes wring (but if someone thinks that it just leads to more confusion - feel free to remove them).

### DIFF
--- a/layers/+lang/haskell/packages.el
+++ b/layers/+lang/haskell/packages.el
@@ -91,8 +91,10 @@
           "me" 'ghc-expand-th
           "mn" 'ghc-goto-next-hole
           "mp" 'ghc-goto-prev-hole
-          "m>"  'ghc-make-indent-deeper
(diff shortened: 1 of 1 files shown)


joehillen:
I really wish the Haskell bindings weren't nested so much. We should refactor them.

d12frosted:
It's just because we have to set them both for literate haskell mode and for regular haskell mode 😸 But I feel you.

syl20bnr:
Thank you ! 👍 

PR #6106, opened 2016-05-20
Improve ivy-hydra config

sooheon:
1. Define ivy--matcher-desc inline for ivy-hydra compatibility while
dependency issues get resolved.

2. Improve ivy-transient-state.
   - Keeps the same functionality.
   - More compact view; fits in 80 chars and 4 vertical lines rather
   than 7.
   - Slight keybinding change: keeps the bindings for toggling auto
   calling and case fold as default (c, C), uses `f` to call ivy-action
   function once without exiting. Closer to defaults, and `f` is prime
   real estate on the keyboard which should be used for a common action,
   rather than to toggle case-fold.

Screenshot comparison:
Old:
screen shot 2016-05-21 at 8 38 53 am
New:
screen shot 2016-05-21 at 8 38 02 am

### DIFF
--- a/layers/+spacemacs/spacemacs-completion/packages.el
+++ b/layers/+spacemacs/spacemacs-completion/packages.el
@@ -140,11 +140,11 @@
     (define-key ivy-minibuffer-map "\C-o"
       (defhydra spacemacs/ivy-transient-state (:hint nil :color pink)
         "
- Move ^^^^^^^^^^     | Call^^         ^^ | Cancel^^ | Options^^     | Action _w_/_s_/_a_: %s(ivy-action-name)
--^-^-^-^-^-^-^-^-^-^-+-^---^----------^^-+-^-^------+-^-^-----------+-----------------------------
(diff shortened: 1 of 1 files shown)


ralesi:
:+1:

syl20bnr:
Thank you for the PR.
The result is visually pleasing and I appreciate the effort you put into it but the docstring is unreadable in the source code which is unfortunately a no go for me :-(
I prefer a simple Hydra docstring with a good balance between the runtime rendering and the source code readability.
Maybe removing the unnecessary unicode chars with the crosshair at the beginning can help to make it less messy. Actually I am against those crosshairs.

sooheon:
So for readability of source code, would you prefer that the 8 movement keys be listed horizontally like the old version? Or I could keep the visual layout, but use `|` instead of the unicode arrows, but that doesn't change the readability much.

syl20bnr:
@sooheon yep, drop the cross and use the same list style everywhere, an example is the window transient state on `SPC w .`.

zilongshanren:
@sooheon  `ivy-hydra` now is a separate package in melpa now.

PR #6108, opened 2016-05-21
Add pivot-window-split function

sooheon:
This function will toggle the orientation of window split between
vertical and horizontal. I have bound to SPC-wi, for "window (i)nvert" or
"window p(i)vot".

Only works when there are two windows, but is still useful much of the time.


agzam:
I don't get it. It's already possible to do with `SPC w L` `SPC w H` (push window right/left) and `SPC w J` `SPC w K`. (push window up/down)

sooheon:
Wow it is. Thats really useful!

Issue #6100, opened 2016-05-20
+lang/scala MELPA scala-mode2 not found


bowerth:
#### Description
The scala-mode2 package listed in the +lang/scala layer cannot be found on melpa

#### Reproduction guide
- Start Emacs
- edit .spacemacs file SPC f e d
- add `scala` to `dotspacemacs-configuration-layers`

*Observed behaviour:*
Found 1 new package(s) to install...
--> refreshing package archive: gnu... [3/3]
--> installing package: scala-mode2@scala... [1/1]
Package scala-mode2 is unavailable. Is the package name misspelled?

*Expected behaviour:*
Installation of scala-mode2 package succeeds

#### System Info
- OS: windows-nt
- Emacs: 25.0.93.1
- Spacemacs: 0.105.20
- Spacemacs branch: master (rev. 8c0a86f)
- Graphic display: t
- Distribution: spacemacs
- Editing style: emacs
- Completion: helm
- Layers:
```elisp
(c-c   clojure emacs-lisp ess eyebrowse git go latex markdown mu4e org polymode python scala yaml)
```

syohex:
https://github.com/melpa/melpa/pull/3906

scala-mode2 is dead.

syl20bnr:
Fixed in develop, I will release a new 0.105 this week.

syl20bnr:
Backported to 0.105.21

PR #6111, opened 2016-05-22
using scala-mode

j-martin:
Fix https://github.com/syl20bnr/spacemacs/issues/6100
Using scala-mode instead of scala-mode-2 as per https://github.com/melpa/melpa/pull/3906

It seems to be working the same way as it used to. As far I can tell.

### DIFF
--- a/layers/+lang/scala/packages.el
+++ b/layers/+lang/scala/packages.el
@@ -14,7 +14,7 @@
     ensime
     noflet
     sbt-mode
-    scala-mode2
+    scala-mode
     ))
 
 (defun scala/init-ensime ()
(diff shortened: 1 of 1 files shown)


diwa:
+1 for this. Scala is currently broken

diego-alvarez-hs:
Great @j-martin I think you also need to update the references on these lines https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Blang/scala/packages.el#L196-L197

j-martin:
Good catch (see previous commit).

syl20bnr:
Thank you ! 👍 
Keeping it opened until it is backported into 0.105.

syl20bnr:
Backported to 0.105.21

Issue #6099, opened 2016-05-20
How to revert custom-set-faces in auto-completion/init-company?


sooheon:
I prefer the standard look, but there is a custom-set-faces for company's init function that overrides whatever I customize it to, even when I reset all customizations. Shouldn't custom-set be left to the end user? While variables can be overriden, custom-set becomes pretty set in stone.

https://github.com/syl20bnr/spacemacs/blob/master/layers/auto-completion/packages.el#L114

d12frosted:
Fixed with released of Spacemacs v0.200. Let us know if you still have any problems with this issue.

PR #6116, opened 2016-05-23
Move custom-set-faces out of company config

sooheon:
Keep it in a recommendation in the README for the layer. Eval'ing the
custom-set-faces directly in the config makes it difficult to revert the
customisations.

Addresses #6099


syl20bnr:
Thank you again !

syl20bnr:
My position on this is that we should let the themes handle such faces, and the spacemacs themes are nice without it so let's remove it :-)

PR #6118, opened 2016-05-23
Refactor cases of layer-usedp for layer composability

sooheon:
The rationale is that users may not have included `auto-completion` layer, but manage `company` themselves. They would want all these other layers to setup company related config, but in each layer, we are currently only checking `layer-usedp 'auto-completion`, etc.


syl20bnr:
I like it and it make a lot of sense, I had to add support for `package-usedp` in `packages.el` files, my solution is a bit of a hack but there is not a lot of choice given the current implementation, related commit: https://github.com/syl20bnr/spacemacs/commit/37dfa47
Roughly we delay package declaration for layers with missing config functions.
cc @StreakyCobra @TheBB 

Thank you ! 👍 

Issue #5194, opened 2016-02-23
spacemacs|dimish does not work when spaceline is excluded


sooheon:
#### Description

Calls to `spacemacs|diminish` when spaceline is excluded from dotspacemacs is ineffectual. Is there some unnecessary coupling there?
#### Reproduction guide
- Add `spaceline` to `dotspacemacs-excluded-packages`
- Set `dotspacemacs-mode-line-unicode-symbols nil`
- Restart Emacs
- Observe non-diminished modeline
- Interactively (or in code) call something like `(spacemacs|diminish evil-cleverparens-mode nil " cp")`

_Observed behaviour:_
No diminishing!

_Expected behaviour:_
Diminishing according to given string.
#### System Info
- OS: darwin
- Emacs: 24.5.1
- Spacemacs: 0.105.11
- Spacemacs branch: develop (rev. fd93264)
- Distribution: spacemacs
- Editing style: hybrid
- Completion: ivy
- Layers:

``` elisp
((auto-completion :variables auto-completion-tab-key-behavior 'cycle auto-completion-return-key-behavior 'complete auto-completion-enable-snippets-in-popup t auto-completion-enable-sort-by-usage t)
 better-defaults colors clojure dash emacs-lisp
 (evil-snipe :variables evil-snipe-enable-alternate-f-and-t-behaviors t)
 finance html markdown
 (shell :variables shell-default-position 'bottom shell-default-shell 'eshell)
 git org osx pandoc rcirc syntax-checking spacemacs-ivy version-control
 (vinegar :variables vinegar-reuse-dired-buffer t)
 sooheon)
```

TheBB:
Yes, this doesn't work if spaceline is not enabled.

person808:
Seems like the function `spacemacs//prepare-diminish` is responsible

justbur:
I just ran into this. Here's a workaround

```elisp
  (defun spacemacs/diminish-hook (_)
    (cl-loop for (mode uni nouni) in spacemacs--diminished-minor-modes
             do
             (diminish mode
                       (if dotspacemacs-mode-line-unicode-symbols
                           uni nouni))))
  (add-hook 'after-load-functions 'spacemacs/diminish-hook)
```

TheBB:
Seems like a reasonable thing to put in `spacemacs-base/init-diminish` if `spaceline` is not present.

d12frosted:
[...]

PR #6119, opened 2016-05-23
Workaround for diminish without spaceline

sooheon:
Applies @justbur's workaround, addresses #5194.
https://github.com/syl20bnr/spacemacs/issues/5194#issuecomment-189389164


syl20bnr:
Thank you again :-)
I updated the function to take into account the magic value to be able to have unicode in GUI and non unicode in terminal with an emacs daemon, related commit https://github.com/syl20bnr/spacemacs/commit/818afdf

PR #6129, opened 2016-05-25
We don't need to declare pcre2el in ivy

sooheon:


syl20bnr:
Thank you !

PR #6147, opened 2016-05-28
[purescript] Pass mode rather than hooks to add-flycheck-hook

actionshrimp:
The signature of `add-flycheck-hook` changed (compared to master) and it looks like the purescript layer still uses the old signature.


syl20bnr:
Nice catch 👍 
Thank you !

PR #6155, opened 2016-05-29
Don’t explicitly set box attribute

TheBB:
Since the face inherits from mode-line this is unnecessary, and makes it difficult to change the mode-line box property.


syl20bnr:
Thank you ! 👍 

PR #6164, opened 2016-05-30
[react layer] Project wide jump to definition

axyz:
dumb-jump seems to do a great work with javascript and finally allow a decent project wide jump to definition.

For now I'm keeping tern keybindings as the behavior is different for searching on same buffer  and by name.

I've also opened an [issue](https://github.com/jacktasia/dumb-jump/issues/41) on dumb-jump to see if we can end up recognizing jsx components and their definitions.

### DIFF
--- a/layers/+frameworks/react/README.org
+++ b/layers/+frameworks/react/README.org
@@ -9,7 +9,8 @@
  - [[#optional-configuration][Optional Configuration]]
  - [[#key-bindings][Key Bindings]]
    - [[#formatting-web-beautify][Formatting (web-beautify)]]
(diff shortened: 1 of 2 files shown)


syl20bnr:
That's really cool, I need to delay the PR a bit because I'm thinking about dumb-jump, maybe we should make it part of the tag layer revamp.
Anyway I'll merge your PR no matter what but I delay it a little bit.

axyz:
just an update: with the last releases of dumb-jump it now works as expected with jsx components as well

TheBB:
After #6851 this should be in now. Let me know if not (you may have to define `SPC mgg` in react.)

Issue #6167, opened 2016-05-30
Running with dotspacemacs-configuration-layers 'all [Collecting bugs] 


JAremko:
I'm trying to run Spacemacs with `dotspacemacs-configuration-layers 'all` and `dotspacemacs-enable-lazy-installation 'unused` and this is what makes my life hard:

- - - - - 
ranger `spc` `a` `r` throws error:

#### Description :octocat:
Error (probably due to missing `fasd`) executable + some other obscure stuff.

#### Reproduction guide :beetle:
- Start Emacs
- try `spc` `a` `r`  

#### System Info :computer:
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.20
- Spacemacs branch: nil (rev. nil)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(rebox floobits spacemacs-editing spacemacs-editing-visual spacemacs-completion spacemacs-ui-visual spacemacs-org spacemacs-evil spacemacs-ui spacemacs-language spacemacs-misc spacemacs-layouts syntax-checking spell-checking nixos osx gtags cscope react django ruby-on-rails bepo nlinum spacemacs spacemacs-bootstrap spacemacs-base erc jabber rcirc dockerfile systemd salt ansible nginx terraform puppet cfengine smex ibuffer typography better-defaults semantic org vim-powerline evil-snipe vinegar vim-empty-lines evil-cleverparens evil-commentary selectric games emoji xkcd themes-megapack colors theming perforce github version-control git chinese latex clojure typescript vimscript emacs-lisp haskell windows-scripts java faust nim sql asciidoc shell-scripts ess scheme go ipython-notebook plantuml bibtex common-lisp javascript html lua racket yaml ruby elm sml extra-langs csharp rust fsharp csv d purescript graphviz python idris agda erlang markdown swift php scala autohotkey c-c++ elixir octave ocaml asm elfeed evernote spotify twitter search-engine wakatime mu4e gnus helm ivy auto-completion ranger finance dash fasd pdf-tools speed-reading deft command-log vagrant restclient chrome imenu-list prodigy shell pandoc ycmd tmux geolocation)
```

#### Message :paw_prints:
```
Fasd executable cannot be found. It is required by `fasd.el'. Cannot add file/directory to the fasd db
Eval error in the `c-lang-defconst' for `c-mode-menu' in php-mode:
[...]

PR #6168, opened 2016-05-30
core: remove support for all layers

TheBB:
We have lazy installation now, and some layers are conflicting, while others are not really intended to be part of all (bepo).


StreakyCobra:
:+1:

Related to #6167

Also adding that the notion of `+distribution` can replace this feature. We should provide a `spacemacs-complete` distribution loading all non-conflicting layers.

syl20bnr:
Thank you !

See commit https://github.com/syl20bnr/spacemacs/commit/60f5a3a for the replacement feature.

PR #6171, opened 2016-05-30
Breaking Bug Fix ERC related errors

CestDiego:
Is there a change in the way things are loaded?

### DIFF
--- a/layers/+chat/erc/packages.el
+++ b/layers/+chat/erc/packages.el
@@ -123,6 +123,7 @@
   (spacemacs|use-package-add-hook erc
     :post-config
     (use-package erc-sasl
+      :defer t
       :if erc-enable-sasl-auth
       ;; Following http://www.emacswiki.org/emacs/ErcSASL
       ;; Maybe an advice would be better?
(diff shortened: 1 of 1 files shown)


syl20bnr:
I removed the quoted `erc`'s in `with-eval-after-load` and somehow it breaks. `with-eval-after-load` is a macro and it should not be quoted but maybe use-package does something with it.
I forgot to test the change, sorry.

TheBB:
Just because it's a macro doesn't mean it shouldn't be quoted, really depends on what the macro does.

```
ELISP> (macroexpand '(with-eval-after-load erc (message "hi")))
(eval-after-load erc
  (lambda nil
    (message "hi")))
```

It shouldn't be quoted in `use-package-add-hook` but it should be quoted in `with-eval-after-load`.

syl20bnr:
So the other `with-eval-after-load` in the code base are wrong since they are not quoted.

syl20bnr:
wow seems my brain bugged last night, everything is quoted in the code base...

syl20bnr:
OK... I understand what happened now, thank you guys.

CestDiego:
It's a pleasure. :)  I'm happy we found this soonish enough

PR #6175, opened 2016-05-31
fix cannot load youdao-dictionary package

TinySong:
use configuration-layer/package-usedp to load "youdao-dictionary" package,
the function argument is wrong


TheBB:
The toggle should be `chinese-enable-youdao-dict`. There shouldn't be a function call at all here.

TinySong:
@TheBB ,OK, is my fault， will open another PR， close this PR

PR #6176, opened 2016-05-31
core: misc. goodies for describe-package

TheBB:
- Use completing-read when called interactively
- Check for string locations in addition to built-in etc.
- Don’t double-print on/off for toggles
- Bind to SPC hdP


syl20bnr:
Cool 👍 
I added a TODO to try to make it work with `configuration-layer/get-all-packages`, I quickly try it and there were some error with missing layer objects.
Thank you !

PR #6178, opened 2016-05-31
core: misc. goodies for toggles

TheBB:
- Add a `:mode` to `add-toggle` since so many toggles are just minor modes anyway.
- Create a predicate function for each toggle
- Use predicate function to show toggle status in SPC h SPC


syl20bnr:
💜

PR #6183, opened 2016-06-01
Consider adding vmd support in markdown mode

bixuanzju:
From its [project](https://github.com/blak3mill3r/vmd-mode) README:
> Fast Github-flavored Markdown previews synchronized with changes to an emacs buffer (no need to save).

In my opinion this is much better than the current live preview support in markdown mode. The only problem might be users need to explicitly install `vmd` binary (this is just as simple as  `npm install -g vmd`).

If you guys are interested, I can then document this functionality in later commits.

### DIFF
--- a/layers/+lang/markdown/README.org
+++ b/layers/+lang/markdown/README.org
@@ -25,6 +25,7 @@ This layer adds markdown support to Spacemacs.
 
 ** Features:
 - markdown files support via [[http://jblevins.org/git/markdown-mode.git/][markdown-mode]]
+- Fast Github-flavored live preview via [[https://github.com/blak3mill3r/vmd-mode][vmd-mode]]
 - TOC generation via [[https://github.com/ardumont/markdown-toc][markdown-toc]]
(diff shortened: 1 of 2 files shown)


syl20bnr:
Thank you 👍 
Can you document it in the README.org ?

bixuanzju:
@syl20bnr No problem!

bixuanzju:
@syl20bnr Done!

syl20bnr:
Thank you ! 👍 
I added anew layer variable to choose the live preview engine, `SPC m c P` is then used both for eww and vmd. To use vmd set the layer variable `markdown-live-preview-engine` to `'vmd`. See commit https://github.com/syl20bnr/spacemacs/commit/ea77a5d

syl20bnr:
BTW, I tried it and electron crashes when the vmd process starts :-(

bixuanzju:
@syl20bnr That's odd. I've been using it for several days without problem.

PR #6184, opened 2016-06-01
invalidate projectile cache on delete and rename

d12frosted:
Invalidate projectile cache when removing or renaming file in project. Doesn't make process longer but saves some key strokes.

### DIFF
--- a/layers/+distribution/spacemacs-base/funcs.el
+++ b/layers/+distribution/spacemacs-base/funcs.el
@@ -297,6 +297,9 @@ argument takes the kindows rotate backwards."
                (when (fboundp 'recentf-add-file)
                    (recentf-add-file new-name)
                    (recentf-remove-if-non-kept filename))
+               (when (and (configuration-layer/package-usedp 'projectile)
+                          (projectile-project-p))
+                 (call-interactively #'projectile-invalidate-cache))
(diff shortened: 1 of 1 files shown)


d12frosted:
(I am voting for my change!)

syl20bnr:
Side question: Do we enable the cache by default in Spacemacs ? I thought it was disabled.

d12frosted:
At first I also thought that by default we are not using caching and this is me who turned it on. So I disabled from my configs and tried cache invalidation and it worked. So I thought no matter what - it would be helpful. But your message forced me to check again. And here is what I found.

From documentation of `projectile-enable-caching`:

> Project caching is automatically enabled by default if you’re
> using the native indexing method.

We are using `alien` as `projectile-indexing-method`. And `projectile-enable-caching` is `nil`. But somehow I still see removed files in `SPS p f` **unless** I invalidate cache. So actually it might be bug in `projectile`.

d12frosted:
@syl20bnr I player with it a little bit more and found that it's actually a bug in my configs. So... this might sound pointless in most cases, because in most cases caching is disabling. But for those who use caching for any reason this thing might be useful. And most importantly - cache invalidation leads to no errors or problems even if caching is disabled. Plus it's consistent with recently merged PR that adds `delete-file` and `rename-file` functions which are also invalidating cache.

syl20bnr:
Thank you ! 👍 

PR #6187, opened 2016-06-01
implement rename and delete file functions

d12frosted:
The only similar functions I've found in Spacemacs are about current buffer and associated file. So I've diced to slightly tweak two functions from my private configs and add them to spacemacs-base. 

I wanted to reuse new function in buffer related functions, but new functions work in `file -> buffer` way, while existing functions work in `buffer -> file` way. Existing functions can be generalised, but I'll leave it for the next time (or another person). 

P. S. and again I vote for my PR. Be nice person and vote for it as well :trollface:

PR #6188, opened 2016-06-01
[ivy] add delete and remove file actions

d12frosted:
Depends on https://github.com/syl20bnr/spacemacs/pull/6187


StreakyCobra:
Why not in the same PR?

d12frosted:
Because they are about different things :) I thought it's better to have two different PRs each with one commit than one PR with two commits that are not really about the same thing.

syl20bnr:
And the second one :-)
Thank you !

PR #2930, opened 2015-09-08
Color identifiers mode

ralesi:
Add color identifiers mode, toggled by `SPC t C i`.

PR #6192, opened 2016-06-01
Allow user to enable rainbow-identifiers-mode without making it default

curtmack:
When you add the colors layer with the `rainbow-identifiers` package enabled (`colors-enable-rainbow-identifiers` variable set to `t`), currently the layer always adds `'rainbow-identifiers-mode` to the `prog-mode-hook`, essentially making it the default. This PR adds a new variable, `colors-rainbow-identifiers-add-hook`, which controls this behavior. Defaults to `t` (previous behavior); if set to `nil`, this hook is not created. This allows a user to install `rainbow-identifiers` through the preexisting layer, still getting the `SPC t C i` toggle binding, without turning it on by default.

Example usage has also been added to the README.org for the colors layer.

I've been running with this for a few months now, but just now got around to prettying it up for release; let me know if there's any style/naming issues.

### DIFF
--- a/layers/+themes/colors/README.org
+++ b/layers/+themes/colors/README.org
@@ -28,22 +28,12 @@ file.
 
 ** Enable rainbow-identifiers
 
-To enable the package =rainbow-identifiers= set the variable
-=colors-enable-rainbow-identifiers= to =t=:
+To enable =rainbow-identifiers-mode= by default, set the variable
+=colors-enable-rainbow-identifiers-by-default= to =t=:
 
 #+BEGIN_SRC emacs-lisp
   (setq-default dotspacemacs-configuration-layers '(
-    (colors :variables colors-enable-rainbow-identifiers t)))
-#+END_SRC
-
-You may also choose not to enable =rainbow-identifiers= by default
-by setting the =colors-rainbow-identifiers-add-hook= variable to =nil=:
-
-#+BEGIN_SRC emacs-lisp
-  (setq-default dotspacemacs-configuration-layers '(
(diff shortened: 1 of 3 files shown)


TheBB:
Can't people who want this just use `(remove-hook 'prog-mode-hook 'rainbow-identifiers-mode)`?

curtmack:
That's also a perfectly fine solution; I made this change originally because creating the hook automatically seemed somewhat rude, but that's admittedly a very "feely" justification.

If no one else has any better reason for wanting this feature, I'll happily close the pull request.

syl20bnr:
I see your point, we have this kind of control on flycheck and some other layers.
I think we don't need a new variable to achieve what you want.

What I would do is to remove this line https://github.com/syl20bnr/spacemacs/blob/develop/layers/+themes/colors/packages.el#L46
And use the layer variable `colors-enable-rainbow-identifiers` to guard the hook: https://github.com/syl20bnr/spacemacs/blob/develop/layers/+themes/colors/packages.el#L65
We can also rename this variable `colors-enable-rainbow-identifiers-by-default` and add an alias on the old name for backward compatibility.

Can you update your PR with the above modifications ?

curtmack:
Okay, that's done.

syl20bnr:
Thank you ! 👍 

In the end I removed this variable for a new variable `color-colorize-identifiers`, in your case you should set it to `nil`. More info in this comment: https://github.com/syl20bnr/spacemacs/pull/2930#issuecomment-223734889

PR #6198, opened 2016-06-02
Show file name when Spacemacs opens a large file

stuhacking:
When Spacemacs encounters a large file it displays a
prompt to open the file literally. However, if the file
is opened as a result of another action (e.g. Visiting a
TAGS table) it might not be obvious which file is triggering
the prompt. This change adds the file name to the prompt.


syl20bnr:
Nice 👍 
Thank you !

PR #6205, opened 2016-06-02
helm-help: pass toggle candidates as list

TheBB:
A problem I found with https://github.com/syl20bnr/spacemacs/pull/6178 is that helm evaluates the candidates with the current buffer being the *helm* buffer, thus most local toggles will have bogus statuses. This fixes that.

Can someone check whether the same has to be done with ivy?


syl20bnr:
Thank you !

PR #6207, opened 2016-06-02
make visual line move configurable

Stebalien:
This is a non-standard key binding and really should be configurable (really, it
should be in a vim-tweaks layer but that's a different story). For people like
me with muscle memory from vim, deviations from vim can be very frustrating.


mindshell:
"For people like me with muscle memory from vim, deviations from vim can be very frustrating."

^ indeed

syl20bnr:
Thank you ! 👍 
I renamed the variable to `dotspacemacs-visual-line-move-text` and changed the default value to nil.

Stebalien:
Thanks!

PR #6208, opened 2016-06-02
Use defconst in LAYER documentation

darkfeline:


TheBB:
Can you update the indentation of the rest of the code samples then, maybe?

darkfeline:
I recall reading that we decided to not use any extra indentation in code blocks, is that right?

syl20bnr:
Thank you ! 👍 

TheBB:
@darkfeline I mean that Emacs Lisp style guidelines indicate that you should indent like

```elisp
(defconst my-var
  "some-value")
```

and not like

```elisp
(defconst my-var
      "some-value")
```

Which is what you get if you just replace `setq` with `defconst`, since for `setq` the following is correct.

```elisp
(setq my-var
      "some-value")
```

PR #6212, opened 2016-06-03
[scala] Using upstream commands to print and yank type at point

d1egoaz:
- Improve copy to clipboard
- changed and added a few commands keymap


syl20bnr:
Thank you ! 👍 
I moved back the print type functions to `SPC m h t` and also used `t` for short type and `T` for full type.
`SPC m h t` is the "standard" (read, also used in other language) key binding for this.

PR #6215, opened 2016-06-03
Add mu4e support to the bepo layer

CharlesHD:
In mu4e headers and view mode `hjkl` are movements keys.
This commit replace these with `csrt`.
`t` and `s` are in both mode bind to important mu4e functions for tagging mail and searching.
I propose to remap them respectively to `è` which is the original `t` place on Azerty keyboard and `/` which is the robust search binding everywhere in spacemacs.

### DIFF
--- a/layers/+keyboard-layouts/bepo/packages.el
+++ b/layers/+keyboard-layouts/bepo/packages.el
@@ -26,6 +26,7 @@
     helm
     ivy
     magit
+    mu4e
     neotree
     org
     org-agenda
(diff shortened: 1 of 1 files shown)


StreakyCobra:
Thank you, nice work :heart: :smiley: :+1: Cherry-picked into `develop`, you can safely delete your branch.

CharlesHD:
Thanks, dunno if it'll help someone though. 

I'd like to add other bindings for twitter mode and other spacemacs spot where bepo is missing (like some transient states).

StreakyCobra:
> Thanks, dunno if it'll help someone though.

I thought the same when creating the bepo layer… and it's useful for some people now, not a lot, but still. It's the same for this, some people will probably appreciate mu4e being supported :-)

> I'd like to add other bindings for twitter mode and other spacemacs spot where bepo is missing (like some transient states).

Please do.

PR #6218, opened 2016-06-04
csharp: enable evil-matchit for C-like tags

andschwa:
The evil-matchit plugin does not have built-in support for C#, but I followed their instructions (and compared to some other code in Spacemacs) to set it up for C# with simple and C tag support.

Thus, `%` now works as expected on, say `#if ... #else ... #endif`, and `switch` statements as well as their cases.

This is an enhancement.


syl20bnr:
Thank you !

PR #6224, opened 2016-06-04
ahs-to-iedit: restrict iedit region to ahs range

### LINKED ISSUE #6223: iedit doesn't respect restricted range
#### Description :octocat:

iedit doesn't respect restricted range.

- Start Emacs
- Create Python file with content:
~~~
def foo(spam):
    return spam
          
def bar(spam):
    return spam
~~~      
- Use `* r r e` on `spam` to restrict range to function and start iedit.
- Replace `spam` with, e.g., `eggs`

*Observed behaviour:* :eyes:
`spam` in both functions are changed.

*Expected behaviour:* :broken_heart:
Only `spam` in one function is changed.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.20
- Spacemacs branch: update-layer-doc (rev. 9f0a983)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
(ivy
 (auto-completion :variables auto-completion-enable-sort-by-usage t)
 better-defaults gtags ibuffer spell-checking syntax-checking semantic typography
 (chrome :variables edit-server-new-frame nil)
 dash elfeed
 (shell :variables shell-default-shell 'shell shell-default-term-shell
        (mir/with-executable "/usr/local/bin/fish" "/usr/bin/fish" "/bin/bash" "/bin/sh")
        shell-default-height 50)
 version-control git c-c   common-lisp csv dockerfile go html javascript latex lua markdown python scheme shell-scripts sql yaml mir-magit-annex mir-pass mir-rc mir-rc-qualia mir-rc-mail)
```

bmag:
Fixes #6223. AHS (auto-highlight-search) can restrict the search to "ranges" (whole buffer, function, display area). Now when `ahs-to-iedit` calls Iedit it also restricts Iedit to AHS's active range.
Each range is called in AHS a "plugin", and this fix should work also for user-defined AHS plugins.

@darkfeline can you confirm that the issue is fixed?

### DIFF
--- a/layers/+spacemacs/spacemacs-editing-visual/packages.el
+++ b/layers/+spacemacs/spacemacs-editing-visual/packages.el
@@ -234,10 +234,14 @@
         (cond
          ((and (not (eq dotspacemacs-editing-style 'emacs))
                (configuration-layer/package-usedp 'evil-iedit-state))
-          (evil-iedit-state/iedit-mode))
+          (evil-iedit-state/iedit-mode)
+          (iedit-restrict-region (ahs-current-plugin-prop 'start)
+                                 (ahs-current-plugin-prop 'end)))
          ((and (eq dotspacemacs-editing-style 'emacs)
                (configuration-layer/package-usedp 'iedit))
-          (iedit-mode))
+          (iedit-mode)
+          (iedit-restrict-region (ahs-current-plugin-prop 'start)
+                                 (ahs-current-plugin-prop 'end)))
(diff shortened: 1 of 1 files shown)


darkfeline:
Works for all three ranges.

syl20bnr:
Thank you ! 💜 

Issue #4929, opened 2016-02-02
kill-this-buffer stops working on OS X


jupl:
#### Description
When using OS X with `dotspacemacs-persistent-server` enabled, buffers can be closed with the keystrokes `SPC b d`. However, when closing window and opening a new one after keystroke no longer works. `SPC b k ENTER` does work. (extra keystroke required though) This is happening in both master and develop.

#### Reproduction guide
- Start Emacs
- `SPC f e d` to open a buffer
- `SPC b d` to close the buffer
- `:q` to close window
- Click on app in dock (app should still be running) to open a new window
- `SPC f e d` to open a buffer
- `SPC b d` to close the buffer

*Observed behaviour:*
Only the first use of `SPC b d` in the reproduction steps closes the buffer. The second one does not work. Messages buffer does not show any error messages.

*Expected behaviour:*
Both instances of `SPC b d` should close the buffer as expected.

robbyoconnor:
@jupl Thanks, but follow this template: https://github.com/syl20bnr/spacemacs/wiki/Debugging#how-to-make-a-great-bug-report note that if you're in develop `SPC h I` will open an issue prefilled with all that template.

sooheon:
FWIW, I can't seem to reproduce this after `:q` or `spc q z`

jupl:
@robbyoconnor Done.

kccai:
I cannot reproduce the issue. But `SPC b d` does not work occasionally (don't know why). I have to restart emacs every time this happens.

mineo:
`SPC b d` stops working for me as well after some time. `menu-bar-menu-frame-live-and-visible-p` is always nil at that point, which means that `kill-this-buffer` simply doesn't do anything. Restarting Emacs fixes that, `SPC q z` and opening a new window makes `menu-bar-...` be nil again.

kccai:
Yes, I also have this problem on my Arch Linux machine. I do not think this is a OS X specific issue.

levindu:
This also happens in Linux.
But if  SPC b d not works, just run 'M-x menu-bar-mode' twice (toggle menu bar), and it work again. No need to restart emacs.

bluesd:
Have this problem too (Linux and Windows), seems to be related with kill-this-buffer when using emacsclient. 

**You can reproduce it just doing:**
* Open `emacs`
* `SPC q z`
* Open `emacs -c`
* `SPC b d`. -> Does nothing.

[...]

PR #6225, opened 2016-06-04
spacemacs-base: Always kill buffers with SPC b d

toupeira:
The `kill-this-buffer` function is from `menu-bar.el` and has some checks which caused https://github.com/syl20bnr/spacemacs/issues/4929. This replaces it with our own naive implementation.

The menu-bar version also calls `abort-recursive-edit` if the minibuffer is open, not sure if we need that here. Is it possible to trigger the `SPC` leader from the minibuffer?

### DIFF
--- a/layers/+distribution/spacemacs-base/funcs.el
+++ b/layers/+distribution/spacemacs-base/funcs.el
@@ -402,6 +402,14 @@ removal."
       (buffer-disable-undo)
       (fundamental-mode))))
 
+;; our own implementation of kill-this-buffer from menu-bar.el
+(defun spacemacs/kill-this-buffer ()
+  "Kill the current buffer."
+  (interactive)
+  (if (window-minibuffer-p)
+      (abort-recursive-edit)
+    (kill-buffer (current-buffer))))
+
 ;; found at http://emacswiki.org/emacs/KillingBuffers
 (defun spacemacs/kill-other-buffers ()
   "Kill all other buffers."
--- a/doc/VIMUSERS.org
+++ b/doc/VIMUSERS.org
@@ -247,7 +247,7 @@ To map ~<Leader>~ keybindings, use the =spacemacs/set-leader-keys= function.
 #+begin_src emacs-lisp
   (spacemacs/set-leader-keys key function) ; Syntax
   ;; Map killing a buffer to <Leader> b c
-  (spacemacs/set-leader-keys "bc" 'kill-this-buffer)
+  (spacemacs/set-leader-keys "bc" 'spacemacs/kill-this-buffer)
   ;; Map opening a link to <Leader> o l only in org-mode (works for any major-mode)
   (spacemacs/set-leader-keys-for-major-mode 'org-mode
     "ol" 'org-open-at-point)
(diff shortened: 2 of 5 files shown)


bmag:
> Is it possible to trigger the SPC leader from the minibuffer?

Apparently it is. For example: `SPC b b M-m b d`

toupeira:
> Apparently it is. For example: `SPC b b M-m b d`

Hmm `M-m` triggers `helm-toggle-all-marks` here. But I added the check for `window-minibuffer-p` anyway and was able to verify that it closes the minibuffer through Emacs mode and `M-: (spacemacs/kill-this-buffer)`.

toupeira:
Let me know if I should update anything else, like docs or the changelog.

bmag:
> M-m triggers helm-toggle-all-marks here

I'm currently using Hybrid+Ivy, so that explains the difference.

> Let me know if I should update anything else, like docs or the changelog.

I found 4 more places where we use `kill-this-buffer`, one of them documentation, which we probably want to change:
- `./layers/+distribution/spacemacs-base/keybindings.el:421`: https://github.com/syl20bnr/spacemacs/blob/3130cc6f2e39a6690f98ddfe4d3deed466365a1c/layers/%2Bdistribution/spacemacs-base/keybindings.el#L421
- `./layers/+completion/ivy/local/ivy-spacemacs-help/ivy-spacemacs-help.el:101`: https://github.com/syl20bnr/spacemacs/blob/3130cc6f2e39a6690f98ddfe4d3deed466365a1c/layers/%2Bcompletion/ivy/local/ivy-spacemacs-help/ivy-spacemacs-help.el#L101
- `./layers/+completion/helm/local/helm-spacemacs-help/helm-spacemacs-help.el:179`: https://github.com/syl20bnr/spacemacs/blob/3130cc6f2e39a6690f98ddfe4d3deed466365a1c/layers/%2Bcompletion/helm/local/helm-spacemacs-help/helm-spacemacs-help.el#L179
- `./doc/VIMUSERS.org:250`: https://github.com/syl20bnr/spacemacs/blob/3130cc6f2e39a6690f98ddfe4d3deed466365a1c/doc/VIMUSERS.org#L250

toupeira:
Thanks, I should have thought to do a quick grep myself ;-) Amended!

bmag:
Looks good to me :+1:

layers/+distribution/spacemacs-base/funcs.el
```diff
@@ -402,6 +402,14 @@ removal."
       (buffer-disable-undo)
       (fundamental-mode))))
 
+;; our own implementation of kill-this-buffer from menu-bar.el
+(defun spacemacs/kill-this-buffer ()
+  "Kill the current buffer."
+  (interactive)
+  (if (window-minibuffer-p)
+    (abort-recursive-edit)
+    (kill-buffer (current-buffer))))
```
syl20bnr: Are you sure the `kill-buffer` sexp should be _inside_ the `if` ? If yes then it is not indented properly, if no you should use `when` instead of `if`.

toupeira: I indeed want to use `if`, indenting is fixed now.

syl20bnr:
Thank you ! 👍 

PR #6227, opened 2016-06-04
Work around the fact that w32-initialized is t too early for us

pfmoore:
Unlike the symbols ```x-initialized``` and ```ns-initialized```, the symbol ```w32-initialized``` is true in daemon mode before GUI functions like ```(find-font)``` work. As a result, ```spacemacs|do-after-display-system-init ``` does not work as expected on Windows in daemon mode.

This patch changes the behaviour on Windows to queue the action if ```(font-family-list)``` is still nil, on the basis that once ```(font-family-list)``` is available, we know that the GUI must be set up.

It's possible that the same check would be just as good on Unix/OSX, but I don't have these systems to check on, and I didn't want to modify working code for no good reason.

### DIFF
--- a/core/core-display-init.el
+++ b/core/core-display-init.el
@@ -29,11 +29,15 @@ the display system initialized."
   "If the display-system is initialized, run `BODY', otherwise,
 add it to a queue of actions to perform after the first graphical frame is
 created."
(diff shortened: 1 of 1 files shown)


syl20bnr:
Nice find 👍 
For now I'm OK with this approach, it is well documented with the comments you added.

Thank you !

PR #6237, opened 2016-06-05
Add javascript-standard as a flycheck checker to react-mode

shahinism:
By default the only available checker in `react-mode` is
javascript-eslint. `javascript-standard` is based on `javascript-eslint`
and has some modification on rules. Since it's not defined as a checker
for this mode, it's also impossible to select it in a buffer.


syl20bnr:
Thank you ! 👍 

PR #6238, opened 2016-06-06
update flyspell-correct after breaking changes

d12frosted:
Original `flyspell-correct` soon will be updated on MELPA with some breaking changes. I am uploading other packages on MELPA and once they are available - it will be good to have changes proposed in PR. 

Oh, and also I switched to using `flyspell-correct-previous-word-generic` instead of `flyspell-correct-word-generic` as it's much pleasant.


syl20bnr:
You can use `:toggle` keyword in the package list instead of `when` sexp in the file, it is more effective to use `:toggle`, `when` sexp forces a reload of the file (to be able to resolve correctly the `package-usedp` calls).

Can we replace `flyspell-popup` with `flyspell-correct-popup` ? Do you think it's possible to have both popup and `(or helm ivy)` interfaces at the same time ?

d12frosted:
I've updated my PR, so now it uses toggles. Also, all recipes are on MELPA now. 

> Can we replace flyspell-popup with flyspell-correct-popup ?

Currently `flyspell-correct` gives `flyspell-correct-previous-word-generic` function that `flyspell-popup` doesn't have. But `flyspell-popup` has this `auto-correct-mode`. I can implement it as well, so it'll be optionally available for any interface.

> Do you think it's possible to have both popup and (or helm ivy) interfaces at the same time ?

Not right now, but it's easy to make it available. Currently correction functions use `flyspell-correct-interface` variable, but it's easy to make it optional variable of correction functions, so one can override interface for specific call. This would make it possible to use different interfaces at the same time without the need to change value of `flyspell-correct-interface`.

But what is the reasoning that you have in mind? Do you want to use popup interface for auto-correct-mode?

syl20bnr:
> Do you want to use popup interface for auto-correct-mode?

I imagine popup interface to work like company but with an idle timer instead when point is on the word. Must be easy to dismiss to not get too much in the way. Not sure it is a good UX though.

syl20bnr:
Thank you ! 👍 

Keeping it open to backport it.

Stebalien:
Note, switching to `flyspell-correct-previous-word-generic` means that words can't be corrected when the cursor is over first character of the misspelled word (at least in evil normal mode):

```
|a|sdf
 ^--- cursor box
```

d12frosted:
It's now possible. Just update the package :)

Stebalien:
Are you referring to https://github.com/d12frosted/flyspell-correct/commit/07431d6d61a730985b27bc4c6aa6f3b6b12448d4? It doesn't appear to fix my problem. To reproduce, open a scratch file, switch to evil-mode (important), type `test foobar`, put the cursor over the `f`, and hit `SPC S c`. I believe that problem is that the cursor is really right before the word.

d12frosted:
@Stebalien thanks for reporting! Indeed it didn't work. I pushed a fix that should be available on MELPA within several hours.

Stebalien:
Thanks!

syl20bnr:
Hotfix not required since flyspell-correct is not in 0.105, closing this PR.

PR #6241, opened 2016-06-06
bootstrap: spacemacs/set-state-faces function

TheBB:
This function is called from core-themes-support.el if it exists, but it doesn't exist. This fixes some annoying faces in daemon mode for me.

I'm not sure why it was removed though.


syl20bnr:
Nice find 👍 
Thank you !

Issue #6182, opened 2016-06-01
Elixir new checkers for flycheck in develop, need some help


syl20bnr:
@smeevil @tomekowal

Hey guys,
I added your flycheck checkers in the elixir layer in develop.
credo seems to work fine but mix compile and test result display no information, since mix compile checker is inspired from test result checker and they both rely on a utility script in Elixir I wonder if you could help to debug what's going on.

Commits:

credo: 5b97df8
mix compile: dd572af
test result: 4996f52
Notes:
For mix compile the checker is disabled by default since it is not safe, you have to setup the layer variable =elixir-enable-compilation-checking= to t or use a directory local variable (see README.org).

Thank you !

tomekowal:
Hi, thanks for your interest! In the `mix compile` checker there need to be three things in place for the checker to actually display anything:

* the `mix_compile_helper` needs to be in path. Is there a way to tell emacs that file shipped with package should be visible as an executable for flycheck? This would remove manual step of copying the file to path.
* the project needs to be valid mix project containing `mix.exs` configuration file otherwise it does nothing
* there have to be errors in place

I will try to debug the problem some time next week.

syl20bnr:
I think all these 3 requirements where met when I tested on ecto.
The utility helper is added to exec-path variable and pressing `SPC e v` in an elixir buffer shows that the checker is operational (provided you set the variable `elixir-enable-compilation-checker` to t).

@tomekowal if you refactor your repo to not be like a Spacemacs layer it would be possible to directly point to it in the elixir layer instead of having the files in Spacemacs.

aaronjensen:
fwiw, the test result checker appears to not work at all, i'm not sure if it was ever completed. Even with `mix_test_helper` in the path, it prints no output. If I modify it to print output, it does not work. @smeevil should https://github.com/smeevil/flycheck-elixir-testresult be working?

smeevil:
Hi,

Sorry I tried a few months ago to switch from the intelli-j range of editors to spacemacs. The initial learning cure was quite big, but after a month I got the hang of it and started to appreciate the editor. I tried to get several of the flycheck plugins to work, credo was one of the first things amongst them. I did get it to work eventually on my setup, but a clean install would not work correctly IIRC. In the mean time I switched back to intelli-j again because i missed the deep integration of the code and other utils it offers. I'm afraid i can be of little help here :/

Gerard.

aaronjensen:
[...]

PR #6249, opened 2016-06-07
Replace local elixir-flycheck-mix-compile package with flycheck-mix f…

tomekowal:
References #6182 

This PR replaces local package `elixir-flycheck-mix-compile` with `flycheck-mix` from melpa. I was able to make it run without additional executables so the configuration is simpler.


syl20bnr:
Thank you ! 👍 

PR #6254, opened 2016-06-07
jabber: don't overwrite mini-buffer and other annoyances

toshism:
jabber.el has the annoying habit of overwriting the mini-buffer with status messages, even when the mini-buffer is active (you are entering a command or whatever), this stops that.

sets status to online when connecting

jabber-whitespace-ping-start to help keep from getting disconnected


syl20bnr:
Thank you ! 👍 
I simplified the layer bit in commit https://github.com/syl20bnr/spacemacs/commit/56844ce

PR #6259, opened 2016-06-08
Auto center spacemacs-buffer on window resize.

ralesi:
Addresses some of the concerns in #5120.  Will resize spacemacs-buffer in current frame if any window is resized.

### DIFF
--- a/core/core-spacemacs-buffer.el
+++ b/core/core-spacemacs-buffer.el
@@ -819,42 +819,67 @@ border."
     (force-mode-line-update)
     (spacemacs-buffer/goto-link-line)))
 
-(defun spacemacs-buffer/goto-buffer ()
+(defvar spacemacs-buffer--last-width nil
+  "Previous width of spacemacs-buffer")
+
+(defun spacemacs-buffer/goto-buffer (&optional refresh)
   "Create the special buffer for `spacemacs-buffer-mode' if it doesn't
 already exist, and switch to it."
   (interactive)
-  (unless (buffer-live-p (get-buffer spacemacs-buffer-name))
-    ;; revise banner length in GUI
-    (when (display-graphic-p)
-      (setq spacemacs-buffer--banner-length
-            (- (window-total-width) 2)))
-    (with-current-buffer (get-buffer-create spacemacs-buffer-name)
-      (page-break-lines-mode)
-      (save-excursion
-        (spacemacs-buffer/set-mode-line "")
-        ;; needed in case the buffer was deleted and we are recreating it
-        (setq spacemacs-buffer--note-widgets nil)
-        (spacemacs-buffer/insert-banner-and-buttons)
-        ;; non-nil if emacs-startup-hook was run
-        (if (bound-and-true-p spacemacs-initialized)
-            (progn
-              (when dotspacemacs-startup-lists
-                (spacemacs-buffer/insert-startupify-lists))
-              (spacemacs-buffer//insert-footer)
-              (spacemacs-buffer/set-mode-line spacemacs--default-mode-line)
-              (force-mode-line-update)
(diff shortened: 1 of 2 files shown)


core/core-spacemacs-buffer.el
```diff
@@ -811,18 +811,22 @@ border."
...
     ;; revise banner length in GUI
     (when (display-graphic-p)
       (setq spacemacs-buffer--banner-length
-            (- (window-total-width) 2)))
+            (window-width)))
     (with-current-buffer (get-buffer-create spacemacs-buffer-name)
       (page-break-lines-mode)
       (save-excursion
+        (when refresh
+          (let ((buffer-read-only nil))
```
bmag: `inhibit-read-only` was made for this

core/core-spacemacs-buffer.el
```diff
@@ -841,12 +845,20 @@ already exist, and switch to it."
...
 
+(add-hook 'window-size-change-functions 'spacemacs-buffer//resize-on-hook)
+
+(defun spacemacs-buffer//resize-on-hook (frame)
+  (let ((space-win (get-buffer-window spacemacs-buffer-name))
+        (frame-win (frame-selected-window frame)))
+    (when (and space-win
+               (not (window-minibuffer-p frame-win)))
+      (save-selected-window
+        (select-window space-win)
```
bmag: `with-selected-window` is more concise

core/core-spacemacs-buffer.el
ralesi: How is one better than another for single command usage?

core/core-spacemacs-buffer.el
ralesi: I agree, I tried that first and it didn't work, but I think that was likely because I was still killing the spacemacs-buffer instead of erasing it.

core/core-spacemacs-buffer.el
bmag: `inhibit-read-only` also ignores read-only text properties (see the variable's documentation)

core/core-spacemacs-buffer.el
syl20bnr: I tried with `with-selected-window` and it works.

syl20bnr:
That's really great 💜 
Do you think it is also possible to resize when the actual windows manager's window (frame) is resized ? For instance in i3 it does not refresh the buffer when I resize the X window.

ralesi:
@syl20bnr, I found bugs in different builds of Emacs 25.X that would allow/disallow it to work with frame resizing.  The latest build of 25. seemed to work for windows when a build from back in March did not trigger on frame resize.  Can you try the latest build of Emacs to see if it works?    Latest changes suggested have addressed the code suggestions.

ralesi:
@syl20bnr, I updated with a few fixes, and it now uses `window-configuration-change-hook`. So should work everywhere.  Also checks for change in buffer width and only fires on change.

deb0ch:
I tried it out and when homepage banner is the random ascii ones, everytime you resize it will load a new different one. It does not break functionnality, but is just a little bit disconcerting.

ralesi:
I also notice with the random banner that the ASCII is not centered, will have to look into how to do that

syl20bnr:
@ralesi not needed there is already a PR for this :-)

ralesi:
@syl20bnr,  which PR is that?

d12frosted:
@ralesi here it is: #6502 💃

TheBB:

TheBB:
Btw, I guess you didn't try this commit with `dotspacemacs-startup-banner` set to `random`. :joy:

https://github.com/syl20bnr/spacemacs/commit/df255da69a180c25c9fce5d7db7618799af284f7

PR #6260, opened 2016-06-08
Add custom evil-register listing with ivy.

ralesi:
Add function to display list of `evil-registers` in `ivy`, overwriting basic evil-register binding.


ralesi:
Also reorganizes `ivy` packages alphabetically.

syl20bnr:
Thank you ! 👍 
I generalized this approach for both helm and ivy so it is cleaner, if none of the helm or ivy layer is used then ahs does not have `b`, `f` and `/` key bindings. I also "improved" the ahs transient state doc (this is subjective but I prefer it like this). Related commits https://github.com/syl20bnr/spacemacs/commit/f5799ab and https://github.com/syl20bnr/spacemacs/commit/2b44ce3

Issue #2452, opened 2015-07-29
`SPC /`: jump to the next match?


xged:
How to jump to the next (previous) match with `SPC /` (`spacemacs/helm-project-smart-do-search-region-or-symbol`) or with one of other search tools (`SPC s ...`) after you selected a candidate.

Vim (grep) equivalent: `:cn`.

justbur:
These are close to what you are asking for:

- TAB on each match in the helm buffer will open the file showing the match in another window without closing helm.

- C-c C-f enables follow mode which automatically shows the match in another window as you C-j C-k through the possible matches.

- C-c ? will show all of the options for the current helm session.

mmcnl:
If you haven't opened another Helm buffer in the meantime, you can return to the last search results using `SPC h l`. 

I too, however, would love it to have an way (like in Vim) to easily switch to the next matching file without having to explictly re-visit the search.

xged:
All good tips. This minimizes the issue, nonetheless I do too think the functionality is needed. One more tip -
`F3`: in a *helm* buffer, convert a *helm* search buffer into a regular buffer.

syl20bnr:
I would like to have such feature too.
It should be provided by `SPC e n` (next-error) but somehow the search with helm is not compatible with next-error :-(

PR #6263, opened 2016-06-08
GNE (Generalized next-error)

TheBB:
This PR does a few different things.

- Refactor the `spacemacs/next-error` and `spacemacs/previous-error` to use a common function to choose which system to use (flycheck or emacs). Also make sure that this function looks for *any* `next-error` capable buffer, and not just one named `*compilation*`.
- Implement a small "generalized" next-error interface that can run in any buffer that stores the results from searches. It plugs in the Emacs next-error system.
- Implement an "error" transient state on `SPC e .` that works with all these systems: it shows which system is in use, and displays additional information from GNE if the next-error buffer in use supports it.
- Advice the functions from helm-grep and helm-ag that saves the results in a buffer (bound to F3 in both cases) to set up the GNE in those buffers.
- Add a keymap to the Spacemacs counsel search with an action on F3 that does the same as the helm variants: writes the results to a buffer. Also set up the GNE in this buffer.

This should fix at least https://github.com/syl20bnr/spacemacs/issues/2452.

This PR is WIP only because I am at a conference right now with limited Emacs capabilities so I can't check if I did everything correctly, but I had it working yesterday in my own config so I'll double check tonight if I got it all right.

### DIFF
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -2216,13 +2216,17 @@ called =pt=.
 
 **** Useful key bindings
 
-| Key Binding            | Description                                                              |
-|------------------------+--------------------------------------------------------------------------|
-| ~F3~                   | in a =helm= buffer, convert a =helm= search buffer into a regular buffer |
-| ~SPC r l~              | resume the last =completion= buffer                                      |
-| ~SPC r s~ or ~SPC s l~ | resume search buffer (completion or converted search buffer)             |
-| ~SPC s `~              | go back to the previous place reached with =helm-ag=                     |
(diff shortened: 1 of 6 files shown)


TheBB:
Okay this is ready now.

TheBB:
Okay, this is ready now.

syl20bnr:
💜 
I wonder if it could be part of a stand-alone package. Some follow up feature could be to make it work with hidden buffers, maybe resurrecting/redisplaying the last one, also putting the point on the current error in the buffer would be great.
Wonderful job 💯 

Thank you !

PR #6265, opened 2016-06-08
rainbow-identifiers: do not overwrite theme values

ksjogo:
Check if a theme is already setting rainbow identifiers saturation before forcing it.


syl20bnr:
Thank you ! 👍 

PR #6271, opened 2016-06-09
fix startup crash if powerline is excluded

ksjogo:
This prohibits a crash when powerline is excluded in dotspacemacs-additional-packages.


syl20bnr:
Thank you !

PR #6273, opened 2016-06-09
Parse CLI args after dotspacemacs/init

TheBB:
Not sure when this one snuck in, but CLI arguments should take priority over variables like `dotspacemacs-elpa-https`.


syl20bnr:
Thank you!
I think parsing CLI parameters after calling `dotspacemacs/init` is too late, for instance the debug parameters are needed before calling this function.
So I reverted this commit and introduced a new variable `spacemacs-insecure` which overrides correctly the dotfile variables. Related commit https://github.com/syl20bnr/spacemacs/commit/419eb51

PR #6277, opened 2016-06-09
Enable volatile highlights for evil and undo-tree

ohspite:
These are not enabled by default.  This patch uses a modified version of the examples [here](https://github.com/k-talo/volatile-highlights.el).


syl20bnr:
Thank you ! 👍 
Somehow there were some error messages printed in `*Message*` because the mode was enabled before defining the extensions, I fixed this, see commit https://github.com/syl20bnr/spacemacs/commit/9d7259d

ohspite:
Cool.  Sorry I didn't catch it, and thanks for fixing that.

Issue #6233, opened 2016-06-05
Consider adding `intero` mode to Haskell layer


bixuanzju:
Just keep a note here for anyone interested in integrating `intero` into Haskell layer.

[Intero](http://commercialhaskell.github.io/intero/)

chrisbarrett:
For cribbing:

https://github.com/chrisbarrett/spacemacs-layers/blob/e3e38a38d41edc3b757ab6dfe62f0d7f11851c4d/cb-new-haskell/packages.el#L186

reckbo:
This would be good

syl20bnr:
`new-haskell` ... interesting. Several Haskell users would like a new/more modern/more opinionated/less bloated haskell layer :-)

d12frosted:
> more opinionated / less bloated

More options easily lead to bloating 💃

@chrisbarrett I like your clean layer. I don't see anything for completion though. I think `company-ghci` would nicely fit. 

Re `interno` - I've seen it and it was on my list to try.

chrisbarrett:
@d12frosted: interno provides a company backend.

d12frosted:
Yeah, just noticed. :D

d12frosted:
> “I just tried this and it works. It just works. I have never had this experience with Haskell + Emacs. Thank you so much for this.”

From interno site.

chrisbarrett:
Yes, it's been solid for me over the past day of hacking.

d12frosted:
I tried your layer, I like it so far. Just several notes:

1. For some reason there is no `haskell-flycheck-holes` in `local` dir, though `haskell-flycheck-holes` is used in `packages.el`.
2. You use `use-package :functions` in many places where it looks like `:commands` is what you wanted. 
3. Shouldn't [it](https://github.com/chrisbarrett/spacemacs-layers/blob/master/cb-new-haskell/packages.el#L313) be `init` instead of `config`?
4. `haskell-mode` misses several established key bindings. 

But anyway, thanks for that layer! Gonna steal some stuff :-P

d12frosted:
P. S. can anyone explain me why haskell ecosystem has so much tooling with so much problems? And instead of solving existing problems the usual solution is to bring yet another tool.

TheBB:
> Avoid success at all cost.

—Simon Peyton Jones

d12frosted:
And write it using comics sans, yes 💃 

But still, I believe you understand my struggle. 😸

[...]

PR #6282, opened 2016-06-10
refine haskell completion backends

d12frosted:
- add `haskell-completion-backend` variable that should be used to select
  desired completion backend
- add support for `intero`
- remove `ghci-ng` support (`intero` ate it)

---

I am sharing this to get some feedback. I tested all backends, but mostly I was focused on `intero`. And it works very good in my point of view. 

Documentation is not ready yet. I'll update it in several days.

Default value of `haskell-completion-backend` is `ghci`, which means that `company-ghci` is used as backend. I used that value for two reasons. Firstly, it works with and without `stack`. Secondly, it doesn't require from user to install anything manually (as opposed to `intero` and `ghc-mod`).

Following #6233 

CC @magthe and anyone willing to test it.

P. S. `intero` just works (c)

### DIFF
--- a/layers/+lang/haskell/README.org
+++ b/layers/+lang/haskell/README.org
@@ -9,12 +9,11 @@
    - [[#layer][Layer]]
    - [[#dependencies][Dependencies]]
    - [[#setup-path][Setup PATH]]
-   - [[#ghc-mod-support][ghc-mod support]]
-     - [[#stack-users][Stack users]]
+   - [[#completion-support][Completion support]]
+     - [[#company-ghci][=company-ghci=]]
+     - [[#intero][=intero=]]
+     - [[#ghc-mod][=ghc-mod=]]
    - [[#optional-extras][Optional extras]]
-     - [[#ghci-ng-support][GHCi-ng support]]
-       - [[#stack-users][Stack users]]
-       - [[#ghc-mod-users][ghc-mod users]]
      - [[#structured-haskell-mode][structured-haskell-mode]]
      - [[#hindent][hindent]]
  - [[#key-bindings][Key bindings]]
@@ -22,15 +21,16 @@
    - [[#debug][Debug]]
    - [[#debug-buffer][Debug Buffer]]
    - [[#repl][REPL]]
+   - [[#intero-repl][Intero REPL]]
    - [[#cabal-commands][Cabal commands]]
    - [[#cabal-files][Cabal files]]
    - [[#refactor][Refactor]]
-   - [[#ghc-mod][Ghc-mod]]
+   - [[#ghc-mod-1][Ghc-mod]]
      - [[#insert-template][Insert template]]
  - [[#syntax-checking][Syntax checking]]
    - [[#flycheck][Flycheck]]
    - [[#hlint][HLint]]
-   - [[#ghc-mod][ghc-mod]]
+   - [[#ghc-mod-2][ghc-mod]]
    - [[#interactive-haskell-mode][Interactive haskell-mode]]
    - [[#flymake][Flymake]]
    - [[#troublesshooting][Troublesshooting]]
@@ -55,7 +55,7 @@ This layer adds support for the [[https://www.haskell.org/][Haskell]] language.
 
 ** Features:
 - syntax highlighting for [[https://github.com/haskell/haskell-mode][haskell source]], [[https://github.com/haskell/haskell-mode][cabal files]], [[https://github.com/bgamari/cmm-mode][C-- source]],
-- auto-completion with [[https://github.com/iquiw/company-ghc][company-ghc]].
+- auto-completion with one of selected backends (=intero=, =ghci= or =ghc-mod=).
 
 *This layer is in construction, it needs your contributions and bug reports.*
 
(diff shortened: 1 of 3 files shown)


bjarkevad:
Looks great! I'm partial to the use of `haskell-completion-backend`, as it would be nicer if it wasn't something that you'd have to worry about. I think moving completely to `intero` eventually might make sense. For now though, it's nice to have the choice.
I'll use this today and tomorrow, and return with any feedback I have.

kRITZCREEK:
@bjarkevad "completely" moving to intero is not an option as it only supports stack projects and a haskell layer ought to be able to work in cabal projects.

bjarkevad:
@kRITZCREEK I should've clarified - I meant using intero as the default, and then fall back to ghci when working on cabal projects.

magthe:
Works beautifully!

Started with the `ghci` completion backend, then switched to `intero`. Nice, nice, nice!

bixuanzju:
Thank you for your layer!  

The repl-related functions seem not working, e.g., `intero-repl-load` gives `apply: Symbol's function definition is void: haskellintero//preserve-focus
`

d12frosted:
@bixuanzju that was a typo, try again :)

d12frosted:
That was a typo 💃 Pushed updated.

bixuanzju:
@d12frosted This time new error message `spacemacs//init-company-intero-repl-mode: Symbol's value as variable is void: company-backends-intero-repl-mode`

bixuanzju:
@d12frosted FYI

>Debugger entered--Lisp error: (void-variable company-backends-intero-repl-mode)
  (set (make-variable-buffer-local (quote company-backends)) company-backends-intero-repl-mode)
  spacemacs//init-company-intero-repl-mode()
  run-hooks(change-major-mode-after-body-hook comint-mode-hook intero-repl-mode-hook)
  apply(run-hooks (change-major-mode-after-body-hook comint-mode-hook intero-repl-mode-hook))
  run-mode-hooks(intero-repl-mode-hook)
  intero-repl-mode()
  intero-repl-buffer()
  (let ((buffer (intero-repl-buffer))) (if (get-buffer-window buffer (quote visible)) nil (display-buffer (intero-repl-buffer))))
  haskell-intero/display-repl()
  call-interactively(haskell-intero/display-repl nil nil)
  command-execute(haskell-intero/display-repl)

bixuanzju:
looks like adding
```
(spacemacs|defvar-company-backends intero-repl-mode)
```
to `config.el` solves the problem!

bixuanzju:
And the prompt looks rather ugly (where is the cute λ??)

screen shot 2016-06-12 at 5 32 48 pm

bjarkevad:
@bixuanzju It looks like that's an Intero related thing:
https://github.com/commercialhaskell/intero/blob/master/elisp/intero.el#L669 
I'm not quite sure why '\4' is used for the prompt, but it renders like that weird question mark.

bixuanzju:
@bjarkevad Seems like this was deliberate https://github.com/commercialhaskell/intero/issues/40 (they call it interrobang)

bjarkevad:
@bixuanzju then I guess it's a non-issue

d12frosted:
@bixuanzju Fixed that. Also diminished `intero-mode`.

d12frosted:
Updated documentation:
  - document haskell-completion-backend variable
  - remove installation notes for ghc-mod as they are not relevant
    anymore
  - remove ghci-ng section
  - overall readme file fixes and improvements

d12frosted:
If no one finds new issues with that in several days, it's a green light for merging :)

Rydgel:
What's the way of testing it, can I tell spacemacs that I want the dev version of a layer?

d12frosted:
@Rydgel it's pretty easy to do. Assuming that your `origin` points to https://github.com/syl20bnr/spacemacs you can use following snippet in terminal. 

```
$ git fetch origin pull/6282/head:haskell/layer-rework
$ git checkout haskell/layer-rework
```

Rydgel:
@d12frosted working great so far

d12frosted:
Oops. Travis check failed.

Rydgel:
It seems that the build phase is using Cabal by default (`SPC m c c`).
Is there a variable to switch to Stack instead?

d12frosted:
@Rydgel 

You can set `haskell-compile-cabal-build-command` to something like `cd %s && stack build`. For example, put following snippet in `user-config` function.

```
(setq haskell-compile-cabal-build-command "cd %s && stack build")
```

If you wish this to be available without restart, you can also `M-: (setq haskell-compile-cabal-build-command "cd %s && stack build") <return>`. Just note that in case you don't use sh-compatible shell (`sh`, `bash`, `zsh` etc) then you have to also change `&&` to suitable operator. For example, for fish that command should be `cd %s; and stack build` or just `cd %s; stack build`.

Thanks for bringing this up. Actually this should be fixed upstream. Meanwhile use proposed solution.

d12frosted:
Ok, looks like no one has problems with this change. So probably it's good to be merged, @syl20bnr .

bjarkevad:
The repl prompt symbol discussed here: https://github.com/commercialhaskell/intero/issues/40 is not working with Source Code Pro it seems - this might need a look?

d12frosted:
Hm. It works for me both in GUI and Terminal. How does it look for you?

angerman:
I don't see a problem with Source Code Pro. @bjarkevad, what do you see?
screenshot 2016-06-19 10 22 35

While I do understand that the use of the interrobang (‽), that `intero` uses for the prompt, maybe 
debatable, I would consider that orthogonal to this PR.

bjarkevad:
@d12frosted @angerman I just reinstalled the fonts on my machine, and now it looks right. I just got the 'missing symbol square' before.

syl20bnr:
@d12frosted will you work on the support for dir local variable in this PR or a new one ?

d12frosted:
In a new one. Also, having this PR in develop will force more feedback :)

syl20bnr:
@d12frosted OK I merged it. The README is impressive, around 500 lines!
Thank you ! 💜 

d12frosted:
Haha, thanks.

Yay, the 🐊 is in the 🏠 .

PR #6288, opened 2016-06-11
Add a new amazing quote from @patrick-motard

robbyoconnor:
This has to be done, the quote was too good!


patrick-motard:
I am patrick-motard and I support this message. Patrick-motard for president 2016.

Issue #5671, opened 2016-04-04
ipython-notebook layer commands not working as expected (i.e. SPC f s doesn't really save the notebook)


jbmorgado:
#### Description

Some ipython-notebook layer commands not working as expected. For instance: `C-r` doesn't call the `ein:notebook-rename-command` as expected, and worst `SPC f s` doesn't `save the notebook like you would a regular buffer` and as expressed in the documentation, instead it saves the all ein environment.
#### Reproduction guide
- Start Emacs
- Open a Jupyter notebook in command line.
- Create a Notebook in Emacs with `SPC a i n` and then choosing new notebook (or open an existing one).

_Observed behaviour:_
- `SPC f s` calls `basic-save-buffer` instead of the intended `ein:notebook-save-notebook-command`
- `C-r` doesn't do anything

_Expected behaviour:_
- `SPC f s` should save the notebook correctly by calling `ein:notebook-save-notebook-command`
- `C-r` should rename the notebook correctly by calling `ein:notebook-rename-command`
#### System Info
- OS: darwin
- Emacs: 24.5.1
- Spacemacs: 0.105.14
- Spacemacs branch: master (rev. 3aff734)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:

``` elisp
((auto-completion :variables auto-completion-enable-snippets-in-popup t auto-completion-complete-with-key-sequence-delay 0.8 auto-completion-enable-help-tooltip t auto-completion-private-snippets-directory "~/Dropbox/Apps/Yasnippet")
 emacs-lisp ess extra-langs
 (git :variables magit-repository-directories
      '("~/repos/"))
 html ipython-notebook
 (latex :variables latex-build-command "LatexMk")
 markdown
 (org :variables org-enable-github-support t)
 osx
 (python :variables python-enable-yapf-format-on-save t)
 shell
 (spell-checking :variables spell-checking-enable-auto-dictionary t spell-checking-enable-by-default nil)
 (syntax-checking :variables syntax-checking-enable-tooltips nil syntax-checking-enable-by-default nil)
 (version-control :variables version-control-diff-tool 'diff-hl version-control-global-margin t))
```

AlejandroCatalina:
Actually, those bindings are intented to be for major mode, so my guess is it works pressing `SPC m f s` and so.

cpaulik:
[...]

PR #6289, opened 2016-06-11
WIP: Restore SPC f s in ipython notebook layer.

cpaulik:
Fix https://github.com/syl20bnr/spacemacs/issues/5671

### DIFF
--- a/layers/+lang/ipython-notebook/packages.el
+++ b/layers/+lang/ipython-notebook/packages.el
@@ -68,7 +68,9 @@
         "+" 'ein:notebook-worksheet-insert-next
         "-" 'ein:notebook-worksheet-delete
         "x" 'ein:notebook-close
-        "u" 'ein:worksheet-change-cell-type
+        "u" 'ein:worksheet-change-cell-type)
+
+      (evil-leader/set-key-for-mode 'ein:notebook-multilang-mode
         "fs" 'ein:notebook-save-notebook-command)
 
       ;; keybindings for ipython notebook traceback mode


layers/+lang/ipython-notebook/packages.el
```diff
@@ -68,7 +68,9 @@
         "+" 'ein:notebook-worksheet-insert-next
         "-" 'ein:notebook-worksheet-delete
         "x" 'ein:notebook-close
-        "u" 'ein:worksheet-change-cell-type
+        "u" 'ein:worksheet-change-cell-type)
+
+      (evil-leader/set-key-for-mode 'ein:notebook-multilang-mode
```
bmag: shouldn't it be `spacemacs/set-leader-keys-for-major-mode`? (the same as the key-bindings above)

cpaulik: No. Because then the keybinding would be `SPC m f s`

bmag: Did you test it? The docstring for `evil-leader/set-key-for-mode` tells me that
> `evil-leader/set-key-for-mode` is an alias for `spacemacs/set-leader-keys-for-major-mode` in `core-keybindings.el`.

cpaulik: Ups, I did this in a hurry. It seems that there no longer is a function that can do what we need here. I don't have time right now to dive into the spacemacs keybinding logic to set this up so it works with all the editing modes.

cpaulik:
Now I understand. The way `evil-leader/set-key-for-mode` works was changed in https://github.com/syl20bnr/spacemacs/commit/ee6a486d4cf837cfd340b8fd4715ce92cce95737 . That's why I thought we could just use it as before. Anyway I don't know enough about keymaps to change `spacemacs//init-leader-mode-map` in a way that would allow the behaviour we want here. Closing for now.

PR #6291, opened 2016-06-11
Add keybinding for magit-reset-popup

robbyoconnor:
I felt like this was missing. It matches `magit`'s keybinding.


robbyoconnor:
Okay this exists so closing this

PR #6295, opened 2016-06-11
clojure: enable smartparens in cider

TheBB:
cider-repl-mode doesn’t derive from comint, therefore doesn't get smartparens (unless, as it happens, strict mode is enabled).


syl20bnr:
Thank you 👍

PR #6305, opened 2016-06-12
Provide more details for ycmd installation and configuration

dathinaios:
A few additions to the `ycmd` layer documentation. I could not find how to get rid of the extra line at the bottom of the code blocks or how to add one above it.

How do you preview the `org-mode` to `markdown` conversion? I had to push and check every time! Is there a better way?

Hope that is useful,

Dionysis


TheBB:
Thanks!

You don't need to quote the variable name when using `setq` (which stands for "set quoted")

```elisp
(setq some-var some-value)
```

syl20bnr:
Nice 👍 
Thank you !

PR #6306, opened 2016-06-12
Fix duplicated layer string format error

CarlQLange:
I had a duplicated layer, but the message that's printed broke spacemacs starting up because it was a format string but wasn't in a `format` form. Simple fix. Spacemacs works again! Yay!

This issue was introduced in https://github.com/syl20bnr/spacemacs/commit/c0851ddcb3cd0f18e41c37f7bfe6b195117f3e27, according to a quick `git bisect`. :)


bmag:
Startup was also broken for me because of a spacemacs warning. Merging this PR locally fixed it for me.

PR #6311, opened 2016-06-12
Fix auto complete with key sequence in Hybrid mode

madand:
Implement handling of `evil-hybrid-state-map` keymap to make completion
with a key sequence work in `Hybird` state.


TheBB:

PR #6314, opened 2016-06-13
Bind "C-c C-e" to spacemacs/helm-find-files-edit for helm-files buffer.

lululau:
I thought the key binding of `helm/find-files-edit` may be a typo, cause I can't find a func which is named as it. And the `spacemacs/helm-find-files-edit` func may be the right one.


syl20bnr:
Oh nice catch ! 💯 
And this key binding is really cool, I should use it :-)
Thank you !

PR #6315, opened 2016-06-13
fix spacemacs buffer for -nw and --batch

ksjogo:
The new spacemacs buffer functionality broke -nw and --batch, fixed this.


siddharthist:
In what way are they broken?

ksjogo:
-nw had/has problems with the graphics as the option is t as it is compiled into the binary, but the console will not display the image.
--batch was not working with the new progress bar as it will calculate negative values. I think another patch already fixed that, but this way it is still improving performance as there is no bar displayed anyway.

siddharthist:
@ksjogo I think you accidentally checked in a commit labeled "proportional layer" BTW

ksjogo:
Oh, yeah, it was tracking my develop branch.

Issue #6317, opened 2016-06-13
Error on docker layer


kostajh:
#### Description :octocat:
New error appears on latest `develop` branch.

```
Error (use-package): docker :config: Symbol’s value as variable is void: docker-container-mode-map
```

#### Reproduction guide :beetle:
- Enable `docker` layer
- restart Emacs
- Open a Dockerfile and manually switch `dockerfile-mode` on 

*Observed behaviour:* :eyes: :broken_heart:
Error (use-package): docker :config: Symbol’s value as variable is void: docker-container-mode-map
No syntax highlighting present

*Expected behaviour:* :heart: :smile:
Syntax highlighting present, no errors

#### System Info :computer:
- OS: gnu/linux
- Emacs: 25.1.50.2
- Spacemacs: 0.105.21
- Spacemacs branch: develop (rev. 1b77fd8)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
(ansible ruby kostajh
         (ivy :variables ivy-initial-inputs-alist nil)
         auto-completion finance docker slack better-defaults pdf-tools themes-megapack
         (erc :variables erc-join-buffer 'bury erc-track-exclude-types
              '("NICK" "333" "353" "MODE" "JOIN" "PART")
              erc-hide-list
              '("JOIN" "PART" "QUIT" "NICK" "MODE")
              erc-fools
              '("SLACK" "omega-deployments" "redmine"))
         emacs-lisp
         (ranger :variables ranger-show-preview t)
         git prodigy markdown
         (elfeed :variables rmh-elfeed-org-files
                 (list "~/.spacemacs.d/feeds.org"))
         (twitter :variables twittering-use-master-password t)
         (gtags :variables helm-gtags-auto-update nil)
         github dash spotify pandoc sql yaml restclient semantic html
         (chrome :variables edit-server-url-major-mode-alist
                 '(("github\\.com" . org-mode)
                   ("pm\\.savaslabs\\.com" . markdown-mode)))
         javascript shell-scripts
         (geolocation :variables geolocation-enable-weather-forecast t geolocation-enable-location-service t)
         emoji vagrant
         php
         (shell :variables shell-default-shell 'eshell shell-enable-smart-eshell t shell-defaul-term-shell "/bin/fish" shell-default-height 40 shell-default-position 'bottom)
         (spell-checking :variables spell-checking-enable-by-default t)
         syntax-checking version-control)
```

[...]

PR #6318, opened 2016-06-13
docker: fix evilification

TheBB:
These maps aren't defined in docker.el.

Should fix https://github.com/syl20bnr/spacemacs/issues/6317

### DIFF
--- a/layers/+tools/docker/packages.el
+++ b/layers/+tools/docker/packages.el
@@ -34,13 +34,13 @@
         "DP" 'docker-push
         "Dp" 'docker-pause
         "Dr" 'docker-restart
-        "Ds" 'docker-start))
-    :config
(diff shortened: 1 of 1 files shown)


kostajh:
Hmm, that fixes the error, but any idea why `dockerfile-mode` isn't enabled automatically when I visit a `Dockerfile`?

TheBB:
It seems to work for me, if by "Dockerfile" you mean "a file called Dockerfile". I haven't used docker at all so that's what I did to test.

robbyoconnor:
@kostajh I encountered this in #6454 as well.

syl20bnr:
Thank you ! 👍 
@robbyoconnor does it fix your issue as well ?

PR #6320, opened 2016-06-13
Support for Hybrid state in Iedit Insert state

MadAnd:
I use `hybrid` editing style. When I enable `iedit-insert-state`, the `insert-state` is activated instead of the expected `hybrid-state`.

This PR addresses the issue with support for toggling of Hybrid mode.


syl20bnr:
Very cool hack 👍 
Thank you !

PR #6323, opened 2016-06-13
restclient: add key bindings for jump to next / previous query

tko:


tko:
Not entirely sure if both of the next/prev key bindings are useful as it seems little redundant, but `gj` seems conventional enough but it's only for vim style.

layers/+tools/restclient/packages.el
```diff
@@ -35,8 +38,29 @@
       (unless restclient-use-org
         (add-to-list 'auto-mode-alist '("\\.http\\'" . restclient-mode)))
       (spacemacs/set-leader-keys-for-major-mode 'restclient-mode
+        "n" 'restclient-jump-next
```
d12frosted: I think it would be better to have it under `SPC g`. So next is `SPC g n` and previous is `SPC g p`. What do you think?

layers/+tools/restclient/packages.el
```diff
@@ -10,9 +10,12 @@
 ;;; License: GPLv3
 (setq restclient-packages
       '(
+        company
+        company-restclient
         golden-ratio
         ob-http
         restclient
+        restclient-helm
```
d12frosted: This package should be used only when `helm` layer is used. 

```
(restclient-helm :toggle (configuration-layer/layer-usedp 'helm))
```

d12frosted:
Thanks for this PR! Please squash your commits as asked in [Contributing Guide](https://github.com/syl20bnr/spacemacs/blob/master/CONTRIBUTING.org#ideally-for-simple-prs-most-of-them).

layers/+tools/restclient/packages.el
tko: Isn't `SPC g` prefix for git? Looking at the docs I think there's some precedent for `SPC m g n` but that seems little unnecessarily cumbersome when almost the whole major mode prefix is unused. If either the toplevel were more full or the `g` sublevel had more commands it would be a good fit, but as it stands two toplevel commands seems better tradeoff as it is more shallow and takes only one extra key compared to one submenu with only two commands.

*shrug* no strong feelings

layers/+tools/restclient/README.org
```diff
@@ -32,6 +32,7 @@ Also there is an [[http://emacsrocks.com/e15.html][Emacs Rocks!]] episode of it.
 |-------------+-------------------------------------------------------------|
 | ~gj~        | Jump to next request                                        |
 | ~gk~        | Jump to previous request                                    |
+| ~SPC m j~   | Jump to a variable or request in current buffer             |
```
tko: Should this be `SPC s j` ? In some major modes it seems similar in function but different somehow. In restclient it gives Imenu error but I know nothing about Imenu.

layers/+tools/restclient/packages.el
d12frosted: Sorry, I mean't `,` instead of `SPC`. So actually, `SPC m g n` and `SPC m g p`. But I am not good at setting good key bindings, so let's wait for more input 😸

layers/+tools/restclient/packages.el
tko: Is there a meaningful difference between `layer-usedp` and `package-usedp` in this context? Seems like the latter is used more often, but the distinction escapes me.

tko:
There's only two hard things in computer science, cache invalidation and key bindings... Split to separate PRs so this one can focus on bikeshedding ;-)

syl20bnr:
Thank you ! 👍 
I removed the bindings gj and gk to keep the default Vim behavior.

PR #6325, opened 2016-06-14
swap c-j/c-k in tide reference mode

zilongshanren:
generally we prefer `j` for next and "k" for previous.


syl20bnr:
Nice catch 👍 
Thank you !

PR #6327, opened 2016-06-14
ipython-notebook: fix loading of keymap

TheBB:
This keymap wasn't defined in ein.el, nor included from there, so it was possible to trigger an error (in this case, by loading org).


TheBB:
This was driving me up the wall so I merged it.

PR #6328, opened 2016-06-14
org: explicitly load org-loaddefs.el

TheBB:
This file ships with org and contains a bunch of autoloads that are not applied since it's never loaded. I'm not sure why these are not in org-plus-contrib-autoloads.el (which **is** loaded), but it's possible to trigger loading errors if this isn't done.


syl20bnr:
Do you have a repro step to trigger one error ?
I remember a package which was overriding this file and thus triggered the same kind of errors, was ox-gfm I believe.

TheBB:
I don't remember any more, but it had something to do with ein.

TheBB:
Okay, I got it. Add org and ipython-notebook layers. Start Emacs,, go to core-documentation.el and eval buffer. You'll get an error. If you load org-loaddefs.el first, you will not get an error.

PR #6330, opened 2016-06-14
Fix the loading of `vmd` for daemon mode

StreakyCobra:
When Spacemacs is used in daemon mode, the `PATH` is not always filled
up with user-defined run paths when Spacemacs load the packages. The
`toggle` of the `vmd` package is checking for the `vmd` executable which
may be in one of the user path, what prevent the package to be loaded,
even if the executable is available at execution time.

If people want to use `vmd`, setting `markdown-live-preview-engine`
should be enough, there is no need for checking for the executable to be
present, this is the responsability of the user to ensure this.

### DIFF
--- a/layers/+lang/markdown/packages.el
+++ b/layers/+lang/markdown/packages.el
@@ -19,8 +19,7 @@
     markdown-toc
     mmm-mode
     smartparens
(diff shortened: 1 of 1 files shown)


syl20bnr:
You are right, this check is unnecessary since it is already an opt-in 👍 
Thank you !

PR #6338, opened 2016-06-14
Don't un-exclude an excluded package

darkfeline:
If a layer using a package is loaded after a layer excluding it, the
exclude flag should not be overwritten.


darkfeline:
Should the tests be fixed, or should I approach this bug from a different direction?

syl20bnr:
Thank you ! 💯 
I fixed the tests and made them more granular in: https://github.com/syl20bnr/spacemacs/commit/8549437

syl20bnr:
Seems that is was the 3000th closed commit so you get the 3000th closed PR achievement!

aaronjensen:
This prevents explicitly including packages that have been excluded like `evil-terminal-cursor-changer`

PR #6342, opened 2016-06-15
don't ask users to spaceline-compile

d12frosted:
Due to 738f4252


d12frosted:
I went ahead and cherry-picked into `develop` branch. Just for fun I'll leave following message.

PR #6352, opened 2016-06-17
Update README.org

tonylotts:
Update Ensime SBT plugin installation instructions for new version and name.

### DIFF
--- a/layers/+lang/scala/README.org
+++ b/layers/+lang/scala/README.org
@@ -50,7 +50,7 @@ Configure the ENSIME sbt plugin by adding the following to
 =~/.sbt/0.13/plugins/plugins.sbt=:
 
 #+BEGIN_SRC scala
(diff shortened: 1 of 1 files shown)


d12frosted:
Hey, thanks for contribution.

I have a question about versioning. Is it necessary? Why just don't ask users to install latest available version?

tonylotts:
@d12frosted, I agree.  The version is a moving target.  I think in this case the source block should be removed, and link to the Ensime documentation instead.  

Closing this, I'll follow up with another PR.

d12frosted:
OK, waiting for new PR 💃 

FYI, it's not necessary to open new PR if you wish to fix something in existing - you can do some changes, commit and then squash / fixup your commits and then force push into your branch. That would look as a clean as new branch, but there will be no need of opening new PR.

Issue #6334, opened 2016-06-14
kill-buffer-and-window


freddie-freeloader:
After using all the time `SPC b d` and `SPC w d`, I got tired of it and I use now `kill-buffer-and-window` (per default `C-x 4 0`).

It would be nice, if one could trigger this with `SPC b D` since `SPC b d` just deletes the buffer and `SPC w D` is already used.

Does this make sense to anyone else?

bmag:
That does sound nice. I've used `kill-buffer-and-window` in the past and I'm sure I'd use it more if it had a better binding than `C-x 4 0`. Are you planning to submit a pull request?

freddie-freeloader:
I'm reading *Beginner's Guide to Contributing a Pull Request to Spacemacs* right know. :smile:

PR #6356, opened 2016-06-17
Add kill-buffer-and-window keybinding

freddie-freeloader:
`SPC b D` would be bound to `kill-buffer-and-window`, which kills the buffer and windows.

This concerns #6334.

I simply added this keybinding and it seems to work fine. 
If you use it on the only existing window, `spacemacs` prevents you from killing it completely, which is the behaviour I'd expect.  

btw: Kudos to everyone who worked on the **Contribution guidelines**. :smiley:

### DIFF
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -1840,6 +1840,7 @@ Buffer manipulation commands (start with ~b~):
 | ~SPC TAB~   | switch to alternate buffer in the current window (switch back and forth) |
 | ~SPC b b~   | switch to a buffer using =helm=                                          |
 | ~SPC b d~   | kill the current buffer (does not delete the visited file)               |
+| ~SPC b D~   | kill the current buffer and its window                                   |
 | ~SPC b e~   | erase the content of the buffer (ask for confirmation)                   |
 | ~SPC b h~   | open =*spacemacs*= home buffer                                           |
 | ~SPC b k~   | kill a buffer                                                            |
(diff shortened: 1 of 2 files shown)


TheBB:
> btw: Kudos to everyone who worked on the Contribution guidelines. :smiley:

On behalf of all those, thank you, but it does say there to always document new keybindings, so please do that as well. ;-)

freddie-freeloader:
Github obviously doesn't like force pushing and even after committing my changes again, it doesn't reopen this issue. Any experience with that? 
I'm sorry for causing trouble. :disappointed:

freddie-freeloader:
You caught me! :laughing: 

I added description in `DOCUMENTATION.org`.

d12frosted:

syl20bnr:
I'm not happy with this binding, if we have a way to kill a buffer and the window then we must have a consistent way to kill a window and a buffer. This is not the case here.

freddie-freeloader:
I thought *and* would be commutative. :smile:

syl20bnr:
Actually there were a lot of work to be done to harmonize the key bindings between `SPC w` and `SPC b`, I'm pretty happy with the result: https://github.com/syl20bnr/spacemacs/commit/b30c393

What it changes for this related PR is that deleting both the buffer and the window is done with the universal prefix argument and it works both in `SPC w` and `SPC b` commands, it also work with all the `ace-window` version of the commands which are on a capital letter.
For instance `SPC u SPC b d` and `SPC u SPC w d`  do the same thing as expected, as well as the ace-window versions `SPC u SPC b D` and `SPC u SPC w D`.

freddie-freeloader:
Pretty nice! This is so much better!

PR #6357, opened 2016-06-17
Add keybindings for hy-mode in python layer

dannyfreeman:
The keybindings allow for connecting to an inferior-lisp repl when in hy-mode.
The hy package must be installed via pip in order to connect to the repl.

### DIFF
--- a/layers/+lang/python/packages.el
+++ b/layers/+lang/python/packages.el
@@ -113,21 +113,19 @@
     :defer t
     :init
     (progn
-      (add-hook 'hy-mode-hook 'paredit-mode)
-      (add-hook 'hy-mode-hook #'smartparens-strict-mode)
       (let ((hy-path (executable-find "hy")))
-        (if hy-path
-            (progn
-              (setq hy-mode-inferior-lisp-command (concat hy-path " --spy"))
-              (spacemacs/set-leader-keys-for-major-mode 'hy-mode
-                "sr" 'inferior-lisp
-                "sb" 'switch-to-lisp
-                "sf" 'lisp-load-file
-
-                "ee" 'lisp-eval-last-sexp
-                "er" 'lisp-eval-region
-                "ef" 'lisp-eval-defun
-                )))))))
+        (when hy-path
+          (setq hy-mode-inferior-lisp-command (concat hy-path " --spy"))
+          (spacemacs/set-leader-keys-for-major-mode 'hy-mode
+            "si" 'inferior-lisp
+            "sb" 'lisp-load-file
+            "sB" 'switch-to-lisp
+            "se" 'lisp-eval-last-sexp
+            "sf" 'lisp-eval-defun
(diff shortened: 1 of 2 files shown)


layers/+lang/python/packages.el
```diff
@@ -110,7 +110,24 @@
...
   (use-package hy-mode
-    :defer t))
+    :defer t
+    :init
+    (progn
+      (add-hook 'hy-mode-hook 'paredit-mode)
+      (add-hook 'hy-mode-hook #'smartparens-strict-mode)
+      (let ((hy-path (executable-find "hy")))
+        (if hy-path
+            (progn
```
bmag: Use `(when <condition> <body>)` instead of `(if <condition> (progn <body>))`

layers/+lang/python/packages.el
```diff
@@ -110,7 +110,24 @@
 
 (defun python/init-hy-mode ()
   (use-package hy-mode
-    :defer t))
+    :defer t
+    :init
+    (progn
+      (add-hook 'hy-mode-hook 'paredit-mode)
+      (add-hook 'hy-mode-hook #'smartparens-strict-mode)
```
bmag: Should honor `dotspacemacs-smartparens-strict-mode`

layers/+lang/python/packages.el
```diff
@@ -110,7 +110,24 @@
 
 (defun python/init-hy-mode ()
   (use-package hy-mode
-    :defer t))
+    :defer t
+    :init
+    (progn
+      (add-hook 'hy-mode-hook 'paredit-mode)
```
bmag: Why `paredit-mode` and not `smartparens-mode`? Either way, the user can exclude both Paredit and Smartparens, perhaps use `with-eval-after-load` or move to `python/post-init-smartparens`?

bmag:
> The hy package must be installed via pip in order to connect to the repl.

1. doesn't the package need to be installed anyway so the user can run Hy programs?
2. it should be documented

layers/+lang/python/packages.el
dannyfreeman: Paredit-mode is shown in the [hy-mode](https://github.com/hylang/hy-mode) repository. That was my only reason. I will change this to `smartparens-mode` and add it to the `python/post-init-smartparens` soon.

bmag:
Looks good now

syl20bnr:
Good 👍 
There as some functions that are not defined for me though, and `lisp-eval-last-sexp` should probably go under `SPC m e e` instead of `SPC m s e` (see convention here: https://github.com/syl20bnr/spacemacs/blob/develop/doc/CONVENTIONS.org#evaluation and https://github.com/syl20bnr/spacemacs/blob/develop/doc/CONVENTIONS.org#send-code) but I was not able to find this function, PR welcome to correct it.

PR #6359, opened 2016-06-18
Added glsl layer for OpenGL shader syntax highlighting

Ell:
Adds a layer for highlighting the syntax of OpenGL shaders.  

Will add company-mode support when https://github.com/Kaali/company-glsl is added to melpa.

### DIFF
--- a/layers/+lang/glsl/README.org
+++ b/layers/+lang/glsl/README.org
@@ -1,0 +1,19 @@
+#+TITLE: GLSL layer
+#+HTML_HEAD_EXTRA: <link rel="stylesheet" type="text/css" href="../../../css/readtheorg.css" />
+
+* Table of Contents
+- [[Desscription][Description]]
+- [[Install][Install]]
+
+* Description
+This layer provides syntax highlighting for GLSL shader files. 
+By default it will provide highlighting for files ending in:
+- .glsl
+- .vert
+- .frag
+- .geom
+
+* Install
+To use this configuration layer, add it to your =~.spacemacs=.
+You will need to add =yaml= to the existing =dotspacemacs-configuration-layers=
+list in this file.
--- a/layers/+lang/glsl/packages.el
+++ b/layers/+lang/glsl/packages.el
@@ -1,0 +1,19 @@
+;;; packages.el --- GLSL Layer packages File for Spacemacs
+;;
+;; Copyright (c) 2012-2016 Sylvain Benner & Contributors
+;;
+;; Author: Alan Love <alan@cattes.us>
+;; URL: https://github.com/ell
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GLPv3
+(setq glsl-packages '(glsl-mode))
+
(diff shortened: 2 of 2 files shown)


layers/+lang/glsl/README.org
```diff
@@ -0,0 +1,19 @@
...
+By default it will provide highlighting for files ending in:
+- .glsl
+- .vert
+- .frag
+- .geom
+
+* Install
+To use this configuration layer, add it to your =~.spacemacs=.
+You will need to add =yaml= to the existing =dotspacemacs-configuration-layers=
+list in this file.
```
syl20bnr: You can create a file `layers.el` and declare this dependency in this file, you can find an example here: 
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+tools/ansible/layers.el#L12

layers/+lang/glsl/packages.el
```diff
@@ -0,0 +1,19 @@
...
+;;; License: GLPv3
+(setq glsl-packages '(glsl-mode))
+
+(defun glsl/init-glsl-mode ()
+  "Initialize GLSL mode"
+  (use-package glsl-mode
+    :mode (("\\.glsl\\'" . glsl-mode)
+           ("\\.vert\\'" . glsl-mode)
+           ("\\.frag\\'" . glsl-mode)
+           ("\\.geom\\'" . glsl-mode))))
```
syl20bnr: `:mode` is not needed here, there are already present in the source code: https://github.com/jimhourihan/glsl-mode/blob/master/glsl-mode.el#L132

layers/+lang/glsl/packages.el
```diff
@@ -0,0 +1,19 @@
...
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GLPv3
+(setq glsl-packages '(glsl-mode))
+
+(defun glsl/init-glsl-mode ()
+  "Initialize GLSL mode"
+  (use-package glsl-mode
+    :mode (("\\.glsl\\'" . glsl-mode)
```
syl20bnr: add `:defer t` for lazy loading.

d12frosted:
> Will add company-mode support when https://github.com/Kaali/company-glsl is added to melpa.

You can prepare it without `company-glsl` being on MELPA. Just use [recipe](https://github.com/syl20bnr/spacemacs/blob/develop/doc/LAYERS.org#packagesel) with github fetcher.

layers/+lang/glsl/README.org
d12frosted: Actually this was copy-paste from `yaml` layer. `Shaders` layer doesn't require `yaml` 😸 And I don't see why it should 💃

d12frosted:

I've also [made some changes](https://github.com/syl20bnr/spacemacs/commit/44a792f83c2f21010b7cbf1ae17f4ae10a3eb481) to this layer. Also see https://github.com/syl20bnr/spacemacs/issues/6931 for completion.

PR #6360, opened 2016-06-18
Clarification on using packages without a layer

lverns:


lverns:
I realized that the example code in this section was straight up incorrect, so I removed it. The documentation points users to the correct spot in `~/.spacemacs` which should be all that is needed.

lverns:
I've now squashed the commits and formated the message as specified in [CONTRIBUTING.org](https://github.com/lverns/spacemacs/blob/docs/CONTRIBUTING.org)

d12frosted:
> I've now squashed the commits and formated the message as specified in CONTRIBUTING.org

👍 Wanted to ask you to do so 💃 

> I realized that the example code in this section was straight up incorrect

What's wrong? `setq` vs `setq-default`?

lverns:
> What's wrong? setq vs setq-default? 

I guess the example wasn't technically wrong, so much as confusing or encouraging bad practice.

It turns out that you can set `dotspacemacs-additional-packages` using either `setq` or `setq-default`. However, the variable is already in `.spacemacs` inside a big call to `setq-default` so users should just modify that instead of making a separate call to `setq`/`setq-default`.  Indeed, setting it in multiple places will result in the last call overriding earlier ones.

d12frosted:
Ah I see. I remember several people reporting problems with layers activation. As it turned out, they made new `setq` call instead of modifying existing variable. My usual advice - read the whole `.spacemacs` file and all the comments before reporting errors. 

In any case, I understand your intention and I like it. Though having example is useful. What do you think about using following snippet?

```
(defun dotspacemacs/layers ()
  "Configuration Layers declaration."
  (setq-default
   ;; ...
   dotspacemacs-additional-packages '(llvm-mode dts-mode)
   ;; ...
   ))
```

d12frosted:
FYI https://github.com/syl20bnr/spacemacs/blob/develop/doc/FAQ.org#install-a-package-not-provided-by-a-layer

lverns:
Question:
Is there a way to build the html [here](http://spacemacs.org/doc/DOCUMENTATION.html) for viewing on my local machine?

syl20bnr:
@lverns Yes you can, load the file `core/core-documentation.el` either with `load-file` or `require` then `SPC SPC spacemacs/publish-doc`. You should have them in `export` folder.

syl20bnr:
Thank you !

PR #6364, opened 2016-06-18
Add mwim (move where i mean) to layer better-defaults

deb0ch:
mwim (move where i mean) is an elpa package for enhancing C-a and C-e behaviours: subsequent strokes will toggle between beginning of line and beginning of code, or between end of line and end of code (before/after comments) respectively.

Before this PR only one behaviour was offered for C-a: going to beginning of line first.

This PR allows for choosing whether you prefer to move to beginning of line or code first (defaults tot the same as before) and adds the equivalent for end of code (before or after comments).

Two layer variables allow for tweaking C-a and C-e behaviours:

- better-defaults-move-to-beginning-of-code-first
- better-defaults-move-to-end-of-code-first

### DIFF
--- a/layers/+emacs/better-defaults/config.el
+++ b/layers/+emacs/better-defaults/config.el
@@ -1,0 +1,20 @@
+;;; config.el --- Better Emacs Defaults Layer configuration variables File
+;;
+;; Copyright (c) 2012-2016 Sylvain Benner & Contributors
+;;
+;; Author: Thomas de Beauchêne <thomas.de.beauchene@gmail.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+(defvar better-defaults-move-to-beginning-of-code-first t
+  "when t, first stroke of C-a will move the cursor to the beginning of code.
+When nil, first stroke will go to the beginning of line.
+Subsequent strokes will toggle between beginning of line and beginning of code.")
+
+(defvar better-defaults-move-to-end-of-code-first t
+  "when t, first stroke of C-e will move the cursor to the end of code (before comments).
+When nil, first stroke will go to the end of line (after comments).
+Subsequent strokes will toggle between end of line and end of code.")
--- a/layers/+emacs/better-defaults/README.org
+++ b/layers/+emacs/better-defaults/README.org
@@ -4,10 +4,11 @@
 
 * Table of Contents                                         :TOC_4_gh:noexport:
  - [[#description][Description]]
- - [[#install][Install]]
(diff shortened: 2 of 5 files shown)


TheBB:
Thanks for the PR! A few things: the defvars should be in config.el, and the keybindings should be in the package init function (so that the package can be excluded without errors).

deb0ch:
Ok, working on it !

deb0ch:
Done

layers/+emacs/better-defaults/config.el
```diff
@@ -0,0 +1,20 @@
...
+;;
+;;; License: GPLv3
+
+(defvar better-defaults-mwim-beginning-of-code-first t
+  "when t, first stroke of C-a will move the cursor to the beginning of code.
+When nil, first stroke will go to the beginning of line.
+Subsequent strokes will toggle between beginning of line and beginning of code.")
+
+(defvar better-defaults-mwim-end-of-code-first t
+  "when t, first stroke of C-a will move the cursor to the end of code (before comments).
```
d12frosted: Shouldn't it be `C-e` instead of `C-a` here?

deb0ch: Well spoted ! I changed it along with the variable names which were a bit mysterious if you don't know the package mwim.

d12frosted:
Haven't tried it yet, but I do like the idea behind this PR.

deb0ch:
He he glad you like it :smile: 

For now I set the defaults to `t` so that people will notice it, but if we decide to set defaults to `nil`, the change would be more 'invisible' but still convenient.

I really don't know which approach would be best, seamless or noticeable.

If you guys have any advice on the matter inputs would gladly be appreciated :smiley:

d12frosted:
Oh. Just noticed that you didn't update README file.

1. Update keybindigns section. 
2. (Optionally) add notes about new vars.

deb0ch:
According to the new vars' values the bindings won't be exactly the same, I don't know what is the best way of expressing that.

Any suggestions ?

deb0ch:
Done, I hope it's clear enough.

d12frosted:
Hey, thanks for the update! I like it.

I made several changes in 237d114:

1. Set `better-defaults-move-to-end-of-code-first` to `nil` by default as it's less intrusive and doesn't change previous behaviour (while I find it useful, some people might find it very confusing).
2. I've made several improvements to README file following [conventions](https://github.com/syl20bnr/spacemacs/blob/develop/doc/CONVENTIONS.org#writing-documentation). Also (not yet in conventions) I've filled paragraphs. 

P. S. 👍 Cherry-picked into `develop`, you can safely delete your branch.

PR #6365, opened 2016-06-18
Integrate pdf-tools with latex layer.

cpaulik:
This enables viewing of tex output using pdf-tools. Includes jumping to
the part of the pdf under the cursor in the .tex file.

The only problem it has is that `auctex` seems to not play nice with lazy loading. It uses `(unless (featurep 'pdf-tools))` [here](https://github.com/giordano/auctex/blob/master/tex.el#L1265).
The code works fine if the `unless` is removed. Is there a way around this or could this be considered a bug in `auctex`?

### DIFF
--- a/layers/+lang/latex/README.org
+++ b/layers/+lang/latex/README.org
@@ -10,11 +10,12 @@
    - [[#layer][Layer]]
    - [[#auto-completion][Auto-completion]]
    - [[#previewing][Previewing]]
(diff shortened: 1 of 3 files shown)


cpaulik:
It works now since we can use our own function for performing the syncing the views.

cpaulik:
The behavior in auctex was fixed so we don''t need our own function any more.

PR #6373, opened 2016-06-20
Add lazy installation support for terraform files

fbergroth:
:smile_cat:


siddharthist:
One thought: HCL support is part of the terraform layer, and if we only load the layer on *.tf files, we'll miss out on HCL syntax highlighting, etc. on other HCL files. Maybe HCL should be its own language layer?

fbergroth:
> HCL support is part of the terraform layer

HCL just happends to work because hcl-mode is a dependency of terraform-mode. I agree though, the terraform layer should be renamed hcl and the terraform-mode package should be part of that layer. I may create a PR in a couple of days.

PR #6376, opened 2016-06-20
add geben to php layer

ksjogo:
add geben (xdebug interface) to php layer

### DIFF
--- a/layers/+lang/php/packages.el
+++ b/layers/+lang/php/packages.el
@@ -15,6 +15,7 @@
         drupal-mode
         eldoc
         flycheck
+        geben
         ggtags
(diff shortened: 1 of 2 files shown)


layers/+lang/php/packages.el
```diff
@@ -39,6 +41,20 @@
 (defun php/post-init-flycheck ()
   (add-hook 'php-mode-hook 'flycheck-mode))
 
+(defun php/init-geben ()
+  (use-package geben
+    :defer t))
```
ksjogo: Probably should add  (setq geben-display-window-function (quote popwin:display-buffer))
as well.

PR #6377, opened 2016-06-20
[WIP] proportional font layer

ksjogo:
Layer for general use of proportional fonts (easier on the eyes).
Adaptive configuration to use mono fonts in specific modes.
Did not do any layer development before, would be glad about feedback and recommendations.
Docs will be fixed.

### DIFF
--- a/layers/+emacs/proportional/README.org
+++ b/layers/+emacs/proportional/README.org
@@ -1,0 +1,61 @@
+#+TITLE: Typography layer
+
+* Table of Contents                                         :TOC_4_gh:noexport:
+ - [[#description][Description]]
+ - [[#install][Install]]
+ - [[#key-bindings][Key bindings]]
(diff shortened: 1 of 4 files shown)


syl20bnr:
> Docs will be fixed.

I should read up to the end before jumping to the code review :-)

What are the criteria for using monospace font over the proportional one ?
I wonder if it would be better to do it the other way around, setting a monospaced font by default and set hook for the proportional one, sound safer to me.

ksjogo:
The monospace font is activated for modes relying on spaces for table formatting. That is mainly tabulated-list-mode (for which for some reason the hook inheritance does not work) and different popups (magit, which-key).

CarlQLange:
This would be nice to have :) Especially with the [Input](http://input.fontbureau.com) font family.

PR #6390, opened 2016-06-21
Prevent find-file-at-point from pinging what looks like domains

### LINKED ISSUE #2654: Pinging clock.be (The Kingdom of Belgium)...
In the screenshot below, I have a `clock` object with a method named `begin`.  Now when I type `clock.be` and then press tab,  I get stuck with the following message in minibuffer -
    
     Pinging clock.be (The Kingdom of Belgium)...

spacemacs-bug

mineo:
Previously,

* typing `gf` in domain names (for example google.com) in normal buffers
* typing `<tab>` in domain names in the minibuffer

would start pinging that domain name. This is not really useful, so
disable it.

Fixes #2654.

I'm not sure if that's the correct place to put this line, but I couldn't find a better one.


syl20bnr:
Thank you ! 👍 
Nice to see someone trying to fix one year old issues :-D

PR #6392, opened 2016-06-21
Fix clone line in README.me

loxaxs:
I replaced the clone address to be consistent with rest of the file. Also, it wouldn't have worked for some unlucky people like me.


d12frosted:
Haha. What is this? oO

PR #6393, opened 2016-06-21
Fix a `git clone` line in README.md

loxaxs:
Fixed the shell code line in "Installation alongside another configuration" paragraph of README.me to use a http address rather than a git@ address. This is more consistent with the rest of file, as well as more compatible with git installations.


d12frosted:

PR #6395, opened 2016-06-22
add counsel-dash

ralesi:
Use counsel-dash when choosing ivy.


d12frosted:
Please, update README file :D

ralesi:
This is pretty seamless, nothing really changes in the documentation.  `counsel-dash` is actually just a wrapper for helm-dash anyway, so it requires it as well.  The bindings remain the same for either ivy or helm completion.

d12frosted:
Ok, now I see that it's just a wrapper. But still, readme says (for example):

> Lookup thing at point in `helm-dash`

If I am `ivy` user I would not use that key binding, because I don't use `helm` :D Confuses me a bit.

ralesi:
Documentation updated to reference both now.

syl20bnr:
Thank you ! 👍 

PR #6396, opened 2016-06-22
show ielm in a popwin buffer, default in insert-mode.

ralesi:
Get ielm to work with popwin.  Default to insert-mode if evil is active.

### DIFF
--- a/layers/+lang/emacs-lisp/packages.el
+++ b/layers/+lang/emacs-lisp/packages.el
@@ -35,7 +35,8 @@
         (spacemacs/declare-prefix-for-mode mode "ms" "ielm")
         (spacemacs/set-leader-keys-for-major-mode mode
           "'" 'ielm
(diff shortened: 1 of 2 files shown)


bmag:
I voted against because (1) I tend to use IELM as a normal interactive shell, the same as I would use a Python terminal, so using popwin and hiding the buffer on `C-g` would be uncomfortable, and (2) because it's inconsistent with all other terminals and REPLs (at least those that I've used), except for the `shell` layer's shells.

d12frosted:
@syl20bnr is it still in merging state? 😸 I don't see it on develop.

syl20bnr:
Thank you!

I was merging it and I asked myself the same questions as @bmag, I came here to comment and saw @bmag's comment so I'm for not merging this one for the sake of consistency with the other REPLs.

Issue #6105, opened 2016-05-20
spacemacs-centered-buffer-mode visual bug


JAremko:
[`spacemacs-centered-buffer-mode` (`SPC` `w` `M`)](https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Bdistributions/spacemacs-base/funcs.el#L192)  looks weird with flycheck 

![http://i.imgur.com/DpSwisq.png](http://i.imgur.com/DpSwisq.png)

And when I go to ranger (`SPC` `a` `r`) it brakes:
![http://i.imgur.com/RjM0asR.png](http://i.imgur.com/RjM0asR.png)

The doc string says *"Minor mode to use big fringe in the current buffer."*  - but it doesn't seems to be buffer local.

*Btw `SPC` buffer also centered, I don't think that it is a desirable behavior.*

JAremko:
I'm thinking. Can we instead center buffer(its content) in the same way `org-indent` does?

JAremko:
Also I can't toggle it if:
 1. I used it
 2. Exited Emacs
 3. Started Emacs from emacs-desktop
 4. Now it doesn't work until I switch buffers.
 5. After switching  it sometimes glitches ![http://i.imgur.com/2pUv2zz.png](http://i.imgur.com/2pUv2zz.png)

JAremko:
fixed in #6397

JAremko:
fixed in #6473

PR #6397, opened 2016-06-22
[WIP]Center buffer with space-doc-mode

JAremko:
Since we are :bomb:  old Emacs versions we can use some extra shenanigans from emacs 24.4+ like `window-text-pixel-size` to achieve great success.  Precisely speaking, center a buffer with the respect of its content maximum width. **Like so**:

**or so:**

*^^^^ they have different width*

*I'm not using `spacemacs-centered-buffer-mode` because it's glitchy #6105 and provides only fixed size centering.*

*Powered by :hocho:  :gun:  :bomb:  :fire:   - be careful!*

### DIFF
--- a/layers/+distributions/spacemacs-base/funcs.el
+++ b/layers/+distributions/spacemacs-base/funcs.el
@@ -206,6 +206,16 @@ automatically applied to."
     (when (assoc ?_ register-alist)
       (jump-to-register ?_))))
 
+(defun spacemacs/ace-center-window ()
+  "Ace center window."
+  (interactive)
+  (require 'ace-window)
+  (aw-select
+   " Ace - Center Window"
+   (lambda (window)
+     (aw-switch-to-window window)
+     (spacemacs-centered-buffer-mode))))
+
 (defun spacemacs/useless-buffer-p (buffer)
   "Determines if a buffer is useful."
   (let ((buf-paren-major-mode (get (with-current-buffer buffer
@@ -374,7 +384,9 @@ removal."
 ;; from magnars
 (defun spacemacs/sudo-edit (&optional arg)
   (interactive "p")
-  (let ((fname (if (or arg (not buffer-file-name)) (read-file-name "File: ") buffer-file-name)))
+  (let ((fname (if (or arg (not buffer-file-name))
+                   (read-file-name "File: ")
+                 buffer-file-name)))
     (find-file
      (cond ((string-match-p "^/ssh:" fname)
             (with-temp-buffer
@@ -383,7 +395,8 @@ removal."
               (let ((last-match-end nil)
                     (last-ssh-hostname nil))
                 (while (string-match "@\\\([^:|]+\\\)" fname last-match-end)
-                  (setq last-ssh-hostname (or (match-string 1 fname) last-ssh-hostname))
+                  (setq last-ssh-hostname (or (match-string 1 fname)
+                                              last-ssh-hostname))
                   (setq last-match-end (match-end 0)))
                 (insert (format "|sudo:%s" (or last-ssh-hostname "localhost"))))
               (buffer-string)))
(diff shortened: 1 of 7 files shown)


JAremko:
Seems to be working as intended.

JAremko:
@syl20bnr  :shipit:

syl20bnr:
That's nice 👍 
I'm having a bug though if I `SPC w V` to open a new window while reading a doc 💔

layers/+spacemacs/spacemacs-org/local/space-doc/space-doc.el
```diff
@@ -102,11 +106,46 @@ It is set by `spacemacs//space-doc-set-space-doc-cache'.")
...
+                                      2) 50)))
+               (delete-other-windows)
+               (set (make-local-variable 'spacemacs--space-doc-origin-left-fringe-width)
+                    (or left-fringe-width 0))
+               (set (make-local-variable 'spacemacs--space-doc-origin-right-fringe-width)
+                    (or right-fringe-width 0))
+               (when (> fringe-width 0)
+                 (setq left-fringe-width fringe-width
+                       right-fringe-width fringe-width)
+                 (setq-local split-width-threshold nil)
```
JAremko: @syl20bnr  hmm.  Apparently nil doesn't work. But `(setq-local split-width-threshold 2000)` will do the trick.

JAremko:
hmmm

syl20bnr:
Going to bed, I let you hmmm, tell me when the light is green again to test it :-D

JAremko:
@syl20bnr  Well... With all the dirty trick in the book.. It seems to be close enough... Probably.  
  :shipit:

TheBB:
Maybe I missed a memo here but why do we want this?

JAremko:
@TheBB   We might not want this. It's up to @syl20bnr   

Pretty much standard situation with my commits  ¯\_(ツ)_/¯

JAremko:
@TheBB  Do you think that centering isn't better or do you think that the code stinks too much? I value your input :heart:     

My reasoning is that it makes documentation easier to read on the big screens, without resizing frame etc.  So it should be good for beginners.  Also it's kinda neat. Docs are one of the first places new Spacemacsians are visiting so it should be impressive, it should show the effort we put into the project.    

So I'm thinking.   If we can make it work reliably (even if it takes some extreme hackery)  it will be cool addition.

tonylotts:
This is a great idea.

layers/+spacemacs/spacemacs-org/local/space-doc/space-doc.el
```diff
@@ -102,11 +106,69 @@ It is set by `spacemacs//space-doc-set-space-doc-cache'.")
...
+
+    (setq  left-fringe-width spacemacs--space-doc-origin-left-fringe-width
+           right-fringe-width spacemacs--space-doc-origin-right-fringe-width)
+    (remove-hook 'window-configuration-change-hook
+                 'spacemacs//space-doc-on-window-width-changed-hook
+                 t)
+    (kill-local-variable org-startup-folded))
+
+  ;; FIXME: Find a better way to update fringes.
+  (let (this-buffer-name (buffer-name))
```
bmag: Did you try `(set-window-buffer nil (current-buffer))`? The documentation for `left-fringe-width` says:
> To make the change take effect, call `set-window-buffer'.

JAremko: @bmag  thanks!

layers/+spacemacs/spacemacs-org/local/space-doc/space-doc.el
```diff
@@ -102,11 +106,69 @@ It is set by `spacemacs//space-doc-set-space-doc-cache'.")
 
 (defun spacemacs//space-doc-hide-line-numbers (&optional enable)
   "If ENABLE is non-nil then toggle off the line numbers."
+  ;; FIXME: Can we prevent it from producing a message?
```
bmag: Merging #6417 should prevent it from producing a message here

TheBB:
> Do you think that centering isn't better or do you think that the code stinks too much?

I don't think it's better. I think it looks silly to have the buffer centered when the text is still left-aligned. But so long as I get https://github.com/syl20bnr/spacemacs/pull/6418 in and can disable it, I'm fine with it.

JAremko:
> don't think it's better. I think it looks silly to have the buffer centered when the text is still left-aligned.  
@TheBB    I think I know how to make it better. Thanks!

JAremko:
> I don't think it's better. I think it looks silly to have the buffer centered when the text is still left-aligned. 

It's pretty popular in web design. I think the difference is that the web pages with centered container and left-aligned text often have some kind of border that makes it look ok.   So I added fringe face inversion - should do the trick.

JAremko:
@syl20bnr  :shipit:

JAremko:
@syl20bnr  :shipit:

JAremko:
Line numbers should be toggled-off again when centering auto disables. Gonna fix it. And might be pick a better color system for fringes.

JAremko:
@TheBB  how about this:





Looks good to me.

JAremko:
@syl20bnr  :shipit:

layers/+distributions/spacemacs-base/funcs.el
```diff
@@ -189,50 +189,55 @@ automatically applied to."
...
+        (when (> fringe-width 0)
+          (setq left-fringe-width fringe-width
+                right-fringe-width fringe-width)
+          (setq-local org-startup-folded nil)))
+
+    (setq  left-fringe-width spacemacs--centered-buffer-mode-origin-left-fringe-width
+           right-fringe-width spacemacs--centered-buffer-mode-origin-right-fringe-width)
+    (face-remap-remove-relative spacemacs--centered-buffer-mode-fringe-face-cookie)
+    (when (and dotspacemacs-line-numbers
+               (not (and space-doc-mode
```
JAremko: space-doc-mode should be wrapped it  `bound-and-true-p`

JAremko:
This is getting ridiculous. Even more than usual :smile:

syl20bnr:
I like it but I would like to be less intrusive with the following changes:
- only toggle centering if there is only one window, if there is a split then display the buffer normally.
- when "maximizing" a doc window (suppressing splits to have only one window) then center the buffer again.

syl20bnr:
Icing on the cake would be to be able to differentiate between vertical and horizontal split, we want to disable it only with vertical split.

A generalization would be to disable it as soon as there is not enough space, i.e. the fringes length are negative.

JAremko:
Ok cool.  I'll do my best :+1:

JAremko:
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -



Now when it is called interactively it will kill its horizontal neighbors otherwise it will enable only if it has enough space in its own window. Also it will disabled when its width is shrinked.

@syl20bnr   It's testing time :shipit:

JAremko:
@syl20bnr  Im thinking.  Mb it will be better if we have `SPC` `w` `M` bound to something like `centered-buffer-in-frame` and  `SPC` `w` `m` to `centered-buffer-in-window` one will kill other windows and another will not.  But `SPC` `w` `m` is taken by `maximize-buffer` :confused:

JAremko:
Hm... sometimes fringes are getting corrupted when scrolled. Probably should add `redraw-frame` somewhere. :disappointed:

JAremko:
This one was a real pain in the rear...

JAremko:
Looks like I done it! I finished. It was horrible :scream:  but fun :smirk: 

@syl20bnr  :shipit:

robbyoconnor:
It has always bugged me that the home buffer acts really weird if the window is maximized

JAremko:
Squashed commits :muscle:

JAremko:
Damn. I think it's too dirty again. I'll wait a day. Need to clear-up my head.  I added a lot of garbage 2day .It should be much simpler than this.

syl20bnr:
@JAremko with all the knowledge you've got with edge cases and fixing it in a spacemacs env. I'm wondering if it would not be better to make it a separate minor mode that we can reuse elsewhere so we can profit from all the hacks for any window.

JAremko:
It's getting better. Don't panic :wink:

PR #6400, opened 2016-06-22
incorporate new functionality from psc-ide

kRITZCREEK:
I've added lots of new keybindings for functionality we added in `psc-ide-emacs`. I'm not entirely sure how to document the new layer variables and would welcome suggestions. Also I'm a spacemacs layer "newbie" so feel free to nitpick ;)

EDIT: @bsermons maybe you could take a look aswell? thanks!

### DIFF
--- a/layers/+lang/purescript/README.org
+++ b/layers/+lang/purescript/README.org
@@ -10,6 +10,9 @@
      - [[#imports][Imports]]
      - [[#psc-ide][psc-ide]]
    - [[#repl][REPL]]
+ - [[#layer-variables][Layer Variables]]
+   - [[#purescript-add-import-on-completion-default-t-][~purescript-add-import-on-completion (default t)~ ]]
+   - [[#purescript-enable-rebuild-on-save-default-nil][~purescript-enable-rebuild-on-save (default nil)~]]
 
 * Description
 
@@ -24,10 +27,12 @@ To use this configuration layer, add it to your =~/.spacemacs=. You will need to
 add =purescript= to the existing =dotspacemacs-configuration-layers= list in this
 file.
 
-If you want to get autocompletions and type information you have to install [[https://github.com/kRITZCREEK/psc-ide][psc-ide]].
+You'll also need to make sure the PureScript compiler and its associated
+binaries (psc-ide-server, psci,...) are on your path. Installation instructions
+can be found [[http://www.purescript.org/download/][here]].
 
 * Key bindings
-
+  
 ** Purescript
 
 *** Imports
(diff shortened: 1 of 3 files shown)


sectore:
@kRITZCREEK It seems that `psc-ide-flycheck-insert-suggestion` (`,-m-i-s`) is not setting the type above the value. The value is removed by its type instead. Please check line 7 of following example:


psc-ide-flyecheck-insert-suggestion

kRITZCREEK:
Thanks for finding that bug! That's actually a bug in `psc-ide` though, so I'll report the bug over there and link to it here.

sectore:
@kRITZCREEK Here are some more `beta` tests based on [psc-ide-emacs`](https://github.com/epost/psc-ide-emacs#usage) examples ;)

**[OK]** [Get completions from the modules you imported while you type](https://github.com/epost/psc-ide-emacs#get-completions-from-the-modules-you-imported-while-you-type-through-company-mode)

**[OK]** [Get completions from all modules in your project with company-complete](https://github.com/epost/psc-ide-emacs#get-completions-from-all-modules-in-your-project-with-company-complete)

**[OK]** [Show type for identifier under cursor](https://github.com/epost/psc-ide-emacs#show-type-for-identifier-under-cursor-c-c-c-t)

**[ERROR]** [Add a clause for the function definition under cursor](https://github.com/epost/psc-ide-emacs#add-a-clause-for-the-function-definition-under-cursor-c-c-c-a)

image

**[OK]** [Add an import for the identifier under cursor](https://github.com/epost/psc-ide-emacs#add-an-import-for-the-identifier-under-cursor-c-c-c-i)

**[NO]** [Add an import on completion](https://github.com/epost/psc-ide-emacs#add-an-import-on-completion-turned-on-by-default) For example: Writing `filter even [1,2,3]`does provide either autocompletion nor autoimport. `Data.Array` and `Data.Int` were installed into the project by bower and the project was build before testing.

kRITZCREEK:
The error you get when adding a clause is actually a typed hole, which is just what we want in that situation since It can't fill out the implementation for you. Instead a typed hole gives you as much information as the compiler can, to help you implement the function.

Adding imports on completions works for me however, so that sounds like a genuine failure... If the module in question eg. `Data.Array` is not imported in your current file, you might need to manually trigger the autocompletion via `company-complete` in order to select `Data.Array.filter`.

syl20bnr:
Thank you ! 👍 
Minor reorganization of the README in the commit https://github.com/syl20bnr/spacemacs/commit/482277d

PR #6401, opened 2016-06-22
User configurable default company-backends

izahn:
Currently `spacemacs|defvar-company-backends` hard codes company backends that will be enabled any time `company-mode` is active (at least that appears to be the case, please correct me if that assumption is wrong). Configuring `company-backends` in `.spacemacs` has no effect, because this variable is not used when constructing the mode-specific company backends. This makes it difficult for the user to configure the default set of enabled company backends.

This PR fixes that by using `company-backends` in `spacemacs|defvar-company-backends`. This allows the user to configure a default set of backends that specific modes can then add to. For example, I really only want file completion everywhere, so I have `(auto-completion :variables company-backends '(company-files))` in my `dotspacemacs-configuration-layors`.

Note that I have set `company-backends` to the old hard-coded list, so this PR should not change anything for those who have not set `company-backends` in their `.spacemacs`. For those who have set this variable the PR makes spacemacs respect it when constructing mode specific company backends.

### DIFF
--- a/core/core-auto-completion.el
+++ b/core/core-auto-completion.el
@@ -11,12 +11,19 @@
 
 ;; Company -------------------------------------------------------------------
 
+(defvar spacemacs-default-company-backends
+  '((company-dabbrev-code company-gtags company-etags company-keywords)
+    company-files company-dabbrev)
+  "The list of default company backends used by spacemacs.
+This variable is used to configure mode-specific company backends in spacemacs.
+Backends in this list will always be active in these modes, as well as any
+backends added by individual spacemacs layers.")
+
 (defmacro spacemacs|defvar-company-backends (mode)
   "Define a MODE specific company backend variable with default backends.
 The variable name format is company-backends-MODE."
   `(defvar ,(intern (format "company-backends-%S" mode))
-     '((company-dabbrev-code company-gtags company-etags company-keywords)
-       company-files company-dabbrev)
+     `,spacemacs-default-company-backends
      ,(format "Company backend list for %S" mode)))
 
 (defmacro spacemacs|add-company-hook (mode)
--- a/layers/+completion/auto-completion/README.org
+++ b/layers/+completion/auto-completion/README.org
@@ -14,8 +14,9 @@
    - [[#add-auto-completion-in-a-layer][Add auto-completion in a layer]]
      - [[#in-configel][In =config.el=]]
      - [[#in-packagesel][In =packages.el=]]
+   - [[#completion-back-ends][Completion back ends]]
    - [[#improved-faces][Improved faces]]
- - [[#key-bindings][Key Bindings]]
+ - [[#key-bindings-1][Key Bindings]]
    - [[#company][Company]]
    - [[#auto-complete][Auto-complete]]
    - [[#yasnippet][Yasnippet]]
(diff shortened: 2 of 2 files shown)


TheBB:
Maybe we can instead change the hook function to take into account the default value of `company-backends` when forming the buffer-local value. That way we don't have to track upstream changes to this default value.

TheBB:
I'm talking about [this one](https://github.com/syl20bnr/spacemacs/blob/develop/core/core-auto-completion.el#L30-L38).

That's fine, but if the value we use is not default, we probably shouldn't name the variable `company-backends`. :-)

izahn:
@TheBB Ah, I see. I don't have any opinion about whether it is better to change this in `spacemacs|add-company-hook` or in `spacemacs|defvar-company-backends`. I changed it in `spacemacs|defvar-company-backends` because that was where it is hard coded it it was obvious to me how to fix it. On the other hand it is not clear to me how to adjust `spacemacs|add-company-hook` to produce the desired result.

syl20bnr:
what about `default-company-backends` ?

izahn:
What problem would be solved by introducing a `default-company-backends`
variable? If there is a reason I don't mind adding it to this PR, but I
don't see the advantage.

syl20bnr:
`default-company-backends` is injected to the buffer local variable `company-backends` and it better conveys the meaning of such variable. Modifying the global `company-backends` variable sounds wrong to me as we loose the ability to use it with `global-company-mode`. I may be wrong but your PR modifies the `company-backends` variable which is a company variable, we don't do that in Spacemacs, we first make this variable file local before modifying it, thus we never modify the global value. A user can then enable globally company and have the default behavior of company with all the backends setup.

core/core-auto-completion.el
```diff
@@ -11,12 +11,14 @@
 
 ;; Company -------------------------------------------------------------------
 
+(setq company-backends '((company-dabbrev-code company-gtags company-etags company-keywords)
```
MadAnd: Better use `defvar`, because then Emacs will show the original value if user changed the variable.

izahn:
I'm not convinced adding `default-company-backends` is a good idea, but I defer to @syl20bnr and @TheBB, and I appreciate the desire not to mess with the default values set by package developers.

Perhaps a better alternative is to use the value of `company-backends` without modifying it. That way company configuration in spacemacs will be consistent with the company package documentation and users can change it if they don't like the default. 

My concern is that adding `default-company-backends` increases the complexity of the system without improving things. Anyone who reads the company package documentation, ports parts of their old `.emacs` file, or does a web search for [emacs company add backend](https://www.google.com/search?hl=en&q=emacs%20company%20add%20backend) is going to try modifying `company-backends`. If we instead use `default-company-backends` that won't work, and a) we will have to write and maintain documentation about this variable and b) users will have to find and read spacemacs documentation in order to configure company back ends. 

If there is still consensus that  a new variable should be introduced let me know and I will add it along with documentation. Otherwise I think using `company-backends` without modification is a good idea.

syl20bnr:
We cannot use company-backends because the company backend packages push to it and this defeat the purpose of having mode-specific backends lists. It is also problematic for users wanting to enable company globally and benefit from company for packages not configured by Spacemacs. This is why I don't want to touch this variable.

Currently it is not possible to alter the default list of backends added to the mode-specific backends lists so I propose a Spacemacs specific variable for this. And it makes sense because this mode-specific approach is a Spacemacs construct, not a company one.

izahn:
OK, thanks for all the feedback, I appreciate everyone who took time to think about this and make suggestions for improving this PR.

 I've added a `spacemacs-default-company-backends` variable and use it instead of `company-backends`. I documented the new variable in the auto-completion layer README.org. I hope this is close to being ready. Please let me know if there are other changes or improvements that should be made to the PR.

TheBB:

PR #6402, opened 2016-06-22
Ignore eww created url directory

tonylotts:
Eww creates a url directory inside of .emacs.d.  This prevents updates because the git status is not clean.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,7 +13,7 @@ eshell/history
 .emacs.desktop.lock
 eshell/alias
 eshell/lastdir
-/url/cookies
+url/*
 my-org/
 org-files/
 semanticdb/


syl20bnr:
Do you use the develop branch or does the url directory come from the master branch ? This directory should be fixed in develop branch so it is not necessary to add it to  the gitignore file.

tonylotts:
@syl20bnr, I use the master branch.  This directory is created when the user loads a page in eww, or views a git reference in magit (magic displays the commuter's gravatar).

syl20bnr:
I cannot reproduce it. Are you able to reproduce it with a clean installation ? Do you think it is possible that the `url` directory has been created a long time ago while it was not set by Spacemacs ?
Do you have a `url` directory in the `.cache` directory ?

tonylotts:
I can reproduce this by viewing a git commit with a Magit.  In the Magit Log buffer, place the cursor on a reference, and press enter.  I specifically reproduced this, by viewing 0283f64 in the spacemacs repo.

The following file is created within '.emacs.d' 'url/cache/<user>/http/com/gravatar/www/<guid>'

tonylotts:
Another way to reproduce this is to load a web page using eww.

syl20bnr:
You are able to reproduce it with a clean Spacemacs and `.spacemacs` file? All the files you mentioned here are in `.cache` directory for me.

tonylotts:
I just tried with a clean Spacemacs install, and fresh .spacemacs file.

- Rename .spacemacs.
- Quit emacs, and delete ~/.emacs.d.
- Clone Spacemacs to ~/.emacs.d.
- Launch emacs (emacs-plus 24.5).
- Select evil mode.
- After compilation, quit emacs, and re-launch.
- Add 'git' layer to 'dotspacemacs-configuration-layers' (no other changes to .spacemacs), and reload.
- View git status of ~/.emacs.d in magit: spc u, spc g s, select ~/.emacs.d
- View log: spc g l l
- View the commit at HEAD: press enter/return
This is the point where the url directory is created within ~/.emacs.d.  A cache entry is created for the author's, and committer's gravatar.

syl20bnr:
Thank you for the great steps, I will check them soon.

joehillen:
This is still happening on develop

syl20bnr:
Thank you !
I have it occasionally and I cannot reproduce it consistently so let's ignore it.

PR #6403, opened 2016-06-22
remove old custom layout on override

d12frosted:
For example, creating new custom layout `["e" . "Name"]` clashes with
`["e" . "Spacemacs"]`. After override `SPC l o` should not show both of
them, but only the override.

### DIFF
--- a/layers/+spacemacs/spacemacs-layouts/funcs.el
+++ b/layers/+spacemacs/spacemacs-layouts/funcs.el
@@ -197,7 +197,10 @@ Available PROPS:
        (if ,already-defined?
            (unless (equal ,already-defined? ,name)
              (spacemacs-buffer/warning "Replacing existing binding \"%s\" for %s with %s"
-                                ,binding ,already-defined? ,name)
+                                       ,binding ,already-defined? ,name)
+             (setq spacemacs--custom-layout-alist
+                   (delete (assoc ,binding spacemacs--custom-layout-alist)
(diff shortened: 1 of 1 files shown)


TheBB:
I think you need to `setq` the result. AFAIK `delete` is destructive but returns the result.

> Write \`(setq foo (delq element foo))' to be sure of correctly changing
the value of a list \`foo'.

d12frosted:
Yes, `delete` is destructive. Didn't know that it's advices to use `setq` even with `delete`.

d12frosted:
@TheBB Updated 😸

TheBB:
In particular,

```
ELISP> (setq test '(a b c d))
(a b c d)

ELISP> (delete 'a test)
(b c d)

ELISP> test
(a b c d)

ELISP> (delete 'b test)
(a c d)

ELISP> test
(a c d)
```

d12frosted:
Hm. The funny thing is that it worked for me without `setq`. Anyway, thanks for fixing problem upfront 😸

TheBB:
Yeah, it'll work if the first element doesn't have to be deleted. There's no way to delete the first element other than pointing the variable to the next cons cell. The others can be deleted by modifying the cdrs to point past the deleted element.

d12frosted:
Oh, didn't know and didn't think about it. Thanks for clarifying that!

syl20bnr:
Thank you ! 👍 
And thank you @TheBB for the explication.

PR #6406, opened 2016-06-23
github: Add bindings for the addition of remotes

IvanMalison:
`github-clone-fork-remote` forks an existing remote
`github-clone-add-existing-remote` Adds an existing remote using completion on any existing forks of the source remote.
`github-clone-add-source-remote` Adds the original parent of a remote that is a fork.

These functions only exist in the newest version of github-clone.el (they were merged here https://github.com/dgtized/github-clone.el/commit/467b40ca60a6c26257466ebc43c74414df7f19cc), but they are definitely already in melpa. Can we do anything to ensure that people are using a properly up to date github-clone.el when these bindings are present?

### DIFF
--- a/layers/+source-control/github/packages.el
+++ b/layers/+source-control/github/packages.el
@@ -72,7 +72,10 @@
     :defer t
     :init
     (spacemacs/set-leader-keys
-      "gh C-c" 'github-clone)))
+      "gh C-c" 'github-clone
+      "ghr" 'github-clone-add-existing-remote
+      "ghf" 'github-clone-fork-remote
+      "ghu" 'github-clone-add-source-remote))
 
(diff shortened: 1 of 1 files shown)


robbyoconnor:
I'd just like to say great work!

syl20bnr:
Thank you !
I moved them all behind `SPC g h c` prefix, it is longer but I was not happy with spreading them all over the place and I don't think these key bindings are often used so it should do it :-)
Note that I made a big refactor on key bindings in git and github layer in order to free up key bindings and be in a more favorable position for the future, related commits:
- https://github.com/syl20bnr/spacemacs/commit/b92daec 
- https://github.com/syl20bnr/spacemacs/commit/8f22e0c

PR #6411, opened 2016-06-23
Fix spacemacs-start-directory variable definition

leonlinc:
Should use defvar to define the spacemacs-start-directory so that user can override the default value.

### DIFF
--- a/core/core-load-paths.el
+++ b/core/core-load-paths.el
@@ -11,6 +11,9 @@
 (defun add-to-load-path (dir) (add-to-list 'load-path dir))
 
 ;; paths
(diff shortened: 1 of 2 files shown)


d12frosted:
For reference: https://github.com/syl20bnr/spacemacs/commit/cbb61070fd1cd0a64ce6e894b8c3bd9697cf3345

leonlinc:
I moved the definition of spacemacs-start-directory to core-load-paths.el.  Otherwise the core library will depend on the init.el and that breaks the unit tests.

syl20bnr:
Very Nice 💯 

Thank you !

PR #6412, opened 2016-06-23
Add leader key for projectile-run-shell

eskimor:


eskimor:
I just discovered there is also a projectile-run-term. Would projectile-run-term be better?

d12frosted:
Hey, thanks you for your contribution 😸 

> Would projectile-run-term be better?

Not better, not worse :D Everyone uses whatever suits his / her needs. Also, there is `projectile-run-eshell`, which is also popular. 

So it would be nice to have keybindings for all of them. For example, under `SPC p x` (just like `C-c p x`) or `SPC p s` (consistent with `SPC a s`). With `t` for `term`, `s` for `shell` and `e` for `eshell`.

eskimor:
That makes sense :-) On the other hand - it gets lengthy. Maybe in addition ps with a customize option for using one of the three?

eskimor:
Added all shells as suggested. The customize thing does not really make sense, as you can equally well just add the binding your self in .spacemacs.

d12frosted:
Hey, thanks for the update. Two comments:

1. You should also document these key bindings [here](https://github.com/syl20bnr/spacemacs/blob/develop/doc/DOCUMENTATION.org#managing-projects). 
2. You [should](https://github.com/syl20bnr/spacemacs/blob/develop/CONTRIBUTING.org#ideally-for-simple-prs-most-of-them) squash your commits into one.

eskimor:
Thank you! Is it ok now?

d12frosted:
@eskimor ya, it's great. Thank you 💃

syl20bnr:
Why not follow `SPC a s` prefix ?
- `SPC p s e`  for eshell
- `SPC p s i` for interactive shell
- `SPC p s T` for term

Would be ideal to provide also the missing `SPC p s m` (multiterm) and `SPC p s t` (ansi-term) :-)

eskimor:
@syl20bnr for consistency with projectiles default key bindings: C-c p x [est]
also this way p s stays free, so one can map his/here preferred shell there. (I did that in my .spacemacs already)

eskimor:
There is no projectile-run-multiterm nor run-ansi-term. Otherwise - why not ;-)

d12frosted:
I am really bad at key bindings, so... @syl20bnr, @TheBB or @bmag could you share your thoughts?

eskimor:
Closing, because it's stale.

PR #6417, opened 2016-06-24
Toggles: print message only in interactive usage

bmag:
Prevent `spacemacs/toggle-*` functions from echoing "enabled"/"disabled" messages when called non-interactively. In other words, we don't need to print the result of the toggle if it wasn't explicitly toggles by the user. The same condition is used for regular Emacs minor mode toggles.


JAremko:
This is awesome!

syl20bnr:
Nice 💜 
Thank you !

PR #6418, opened 2016-06-24
Allow user more control over space-doc mode

TheBB:
* Move org-indent-mode and view-mode from spacemacs/prettify-buffer to space-doc.el.
* Make spacemacs--space-doc-modificators a variable instead of a constant, so users can change it.


JAremko:
I think it will be useful.  But it probably should be documented somewhere. Mb add it to FAQ?

syl20bnr:
I like it, we can document it latter as it is a rather advanced settings right now (some function names are not user friendly for instance).
Thank you ! 👍 

PR #6422, opened 2016-06-24
Fixes display of per-binding docs in micro-state

dwang20151005:
Previous behavior, if a micro-state binding had :doc not nil, was an error
  Wrong type argument: stringp, 1

Detailed reproduction:
cat > /tmp/test.el <<EOF
(spacemacs|define-micro-state test
  :doc "[a] a [b] b [q] quit"
  :bindings
  ("a" nil :doc "a")
  ("b" nil :doc "b")
  ("q" nil :doc "quit" :exit t))

(spacemacs/test-micro-state)
EOF
emacs --load /tmp/test.el

Type a, b, or q and see the error in the echo buffer.

### DIFF
--- a/core/core-micro-state.el
+++ b/core/core-micro-state.el
@@ -164,17 +164,20 @@ used."
          (doc-body
           `((let ((bdoc ,@binding-doc)
                   (defdoc ,@default-doc))
-              (if bdoc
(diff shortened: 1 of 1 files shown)


syl20bnr:
Thank you ! 👍 

Issue #6419, opened 2016-06-24
Hooks are executed before .dir-locals are applied


d12frosted:
It seems that hooks are executed before directory locals are applied, so it breaks ability to setup desired tools on per project (directory) basis. For example, setting `ruby-test-runner` to `rspec` in `.dir-locals.el` doesn't enable `rspec-mode`. 

You can also check it by using following content of `.dir-locals.el`.

```
((nil . ((eval . (progn
                   (message " >>>> dir-locals.el"))))))
```

And then execute following elisp code.

```
(defun test-hook-vs-dir-locals ()
  (message " >>>> hook"))

(add-hook #'prog-mode-hook #'test-hook-vs-dir-locals)
```

And then open any file that runs in `prog-mode` derived major mode and observe following output in `*Messages*`.

```
 >>>> hook
 >>>> dir-locals.el
```

At first I thought that there is something different between Emacs 25 and Emacs 24, but this behaviour doesn't change upon switching to Emacs 24.

@syl20bnr do you have any ideas?

TheBB:
Some options in the first answer [here](http://stackoverflow.com/questions/5147060/how-can-i-access-directory-local-variables-in-my-major-mode-hooks). The first one looks quite elegant and easy to implement for us.

d12frosted:
Thanks, I'll take a look. I just thought that it must be working, since @syl20bnr was talking about supporting such flow for haskell layer[1] as it's supported for ruby layer.

[1]: It's pretty much finished, just doesn't work because of this issue

TheBB:
There might be a bug in the ruby layer then. I thought everything that checked the variables in question was postponed until the actual bound functions were called, but I see now that is not the case.

syl20bnr:
I had to use hack-local-variables at some point but I don't remember where, if you search in the code base you should find it. Or maybe it was for elixir when I played with the flycheck checkers but the code may be gone. I'll search for you when I have a chance.

syl20bnr:
It is still here: https://github.com/syl20bnr/spacemacs/blob/develop/layers/+lang/elixir/funcs.el#L24

The ruby layer should not work indeed... :-(

syl20bnr:
Seems that @TheBB link provides cleaner implementations, since this will be a recurrent pattern it would be cool to have an API to easily dispatch sexps based on dir local vars.

d12frosted:
[...]

PR #6424, opened 2016-06-24
implement API for local-vars-hook

d12frosted:
Following https://github.com/syl20bnr/spacemacs/issues/6419#issuecomment-228298464 and https://github.com/syl20bnr/spacemacs/issues/6419#issuecomment-228327497.

### DIFF
--- a/layers/+distributions/spacemacs-base/config.el
+++ b/layers/+distributions/spacemacs-base/config.el
@@ -189,3 +189,11 @@ It runs `tabulated-list-revert-hook', then calls `tabulated-list-print'."
 (defun server-remove-kill-buffer-hook ()
   (remove-hook 'kill-buffer-query-functions 'server-kill-buffer-query-function))
 (add-hook 'server-visit-hook 'server-remove-kill-buffer-hook)
+
+;; ---------------------------------------------------------------------------
+;; Other
+;; ---------------------------------------------------------------------------
+
+;; hook into `hack-local-variables' in order to allow switching spacemacs
+;; configurations based on local variables
+(add-hook 'hack-local-variables-hook #'spacemacs//run-local-vars-mode-hook)
--- a/layers/+lang/elixir/packages.el
+++ b/layers/+lang/elixir/packages.el
@@ -124,8 +124,8 @@
                    (cons 'elixir-enable-compilation-checking nil))
       (add-to-list 'safe-local-variable-values
                    (cons 'elixir-enable-compilation-checking t))
-      (add-hook 'elixir-mode-hook
-                'spacemacs//elixir-enable-compilation-checking t))))
+      (spacemacs/add-to-local-vars-hook 'elixir-mode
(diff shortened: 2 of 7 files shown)


d12frosted:
I've also pushed to this branch haskell layer update to support this flow and fix for ruby layer.

@TheBB @syl20bnr I can separate it to tree distinct PRs if you think that it would be better

layers/+distributions/spacemacs-base/config.el
```diff
@@ -189,3 +189,11 @@ It runs `tabulated-list-revert-hook', then calls `tabulated-list-print'."
...
   (remove-hook 'kill-buffer-query-functions 'server-kill-buffer-query-function))
 (add-hook 'server-visit-hook 'server-remove-kill-buffer-hook)
+
+;; ---------------------------------------------------------------------------
+;; Other
+;; ---------------------------------------------------------------------------
+
+;; hook into `hack-local-variables' in order to allow switching spacemacs
+;; configurations based on local variables
+(add-hook #'hack-local-variables-hook #'spacemacs//run-local-vars-mode-hook)
```
TheBB: Should probably not be `#` on the first one.

layers/+lang/elixir/packages.el
```diff
@@ -124,8 +124,8 @@
                    (cons 'elixir-enable-compilation-checking nil))
       (add-to-list 'safe-local-variable-values
                    (cons 'elixir-enable-compilation-checking t))
-      (add-hook 'elixir-mode-hook
-                'spacemacs//elixir-enable-compilation-checking t))))
+      (spacemacs/add-to-local-vars-hook 'elixir-mode-hook
```
TheBB: Judging from the two functions in spacemacs-base/funcs.el you have to use `elixir-mode` here?

layers/+distributions/spacemacs-base/config.el
d12frosted: Indeed, it's var, not func. 👍

layers/+lang/elixir/packages.el
d12frosted: Thanks for catching this.

d12frosted:
Also I think that this feature should be documented. General note in documentation and links from layers that support such configuration.

syl20bnr:
Excellent 💯 
I removed the helper because it was abstracting a one liner so we are better to use the standard functions to hook functions.
Also instead of `(concat ....)` with symbol names you can use `(format "%S" ...)`
Thank you ! 👍 

d12frosted:
> I removed the helper because it was abstracting a one liner so we are better to use the standard functions to hook functions.

Makes sense. And looks much shorter. Thanks.

As per `concat` vs `format` - didn't know about `%S`. Thanks. 

P. S. commit for reference: https://github.com/syl20bnr/spacemacs/commit/89dc1af9a60e5d8bdf58e345c46b78df39835fbd

PR #6428, opened 2016-06-25
allow to set dotspacemacs-filepath

d12frosted:
Following the trend. This is useful in addition to allowing to set `spacemacs-start-directory` and `package-user-dir`. But not sure if it should be documented.


syl20bnr:
Thank you ! 👍 

PR #6439, opened 2016-06-27
org-agenda: Add missed key binding for pomodoro

mskorzhinskiy:
Just realized that keybinding for pomodoro is missed in agenda buffer.


syl20bnr:
Thank you !
I moved the key binding to `org-pomodoro` instead since anyone could exclude this package if they want, related commit https://github.com/syl20bnr/spacemacs/commit/0a719b1

PR #6441, opened 2016-06-27
Add duplicate-line-or-region under SPC x l d

deb0ch:
from https://www.emacswiki.org/emacs/CopyingWholeLines

Duplicate current line, or currently selected region if active.
With argument N, make N copies.
With negative N, comment out original line and use the absolute value.

Default keybinding is C-d to allow fast access in emacs editing style, configurable with a layer variable.
Also accessible under `SPC x l d` for "text - line - duplicate".

**Ongoing reflection:**
- `SPC x l d` seems like the most logical but it is quite long, if you have an idea of something shorter your advice is very welcomed.
- `C-d` seems like a fine shortcut to me: fast access and mnemonic, but as it might (?) break someone's workflow I put it in a configuration variable.
  Still not sure if that configuration variable is really needed or if we should hard-code it. Input needed.

### DIFF
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -2562,6 +2562,7 @@ Text related commands (start with ~x~):
 | ~SPC x j r~ | set the justification to right                                |
 | ~SPC x J~   | move down a line of text (enter transient state)              |
 | ~SPC x K~   | move up a line of text (enter transient state)                |
+| ~SPC x l d~ | duplicate line or region                                      |
 | ~SPC x l s~ | sort lines                                                    |
 | ~SPC x l u~ | uniquify lines                                                |
 | ~SPC x o~   | use avy to select a link in the frame and open it             |
--- a/layers/+distributions/spacemacs-base/funcs.el
+++ b/layers/+distributions/spacemacs-base/funcs.el
@@ -710,6 +710,29 @@ the right."
   (interactive)
   (call-interactively 'write-file))
 
+;; from https://www.emacswiki.org/emacs/CopyingWholeLines
+(defun spacemacs/duplicate-line-or-region (&optional n)
+  "Duplicate current line, or region if active.
+With argument N, make N copies.
+With negative N, comment out original line and use the absolute value."
+  (interactive "*p")
+  (let ((use-region (use-region-p)))
+    (save-excursion
+      (let ((text (if use-region        ; Get region if active, otherwise line
+                      (buffer-substring (region-beginning) (region-end))
+                    (prog1 (thing-at-point 'line)
+                      (end-of-line)
+                      (if (< 0 (forward-line 1)) ; Go to beginning of next line, or make a new one
+                          (newline))))))
+        (dotimes (i (abs (or n 1)))     ; Insert N times, or once if not specified
+          (insert text))))
+    (if use-region nil                  ; Only if we're working with a line (not a region)
+      (let ((pos (- (point) (line-beginning-position)))) ; Save column
+        (if (> 0 n)                             ; Comment out original with negative arg
+            (comment-region (line-beginning-position) (line-end-position)))
+        (forward-line 1)
+        (forward-char pos)))))
+
 (defun spacemacs/uniquify-lines ()
   "Remove duplicate adjacent lines in region or current buffer"
   (interactive)
--- a/layers/+distributions/spacemacs-base/keybindings.el
+++ b/layers/+distributions/spacemacs-base/keybindings.el
@@ -418,6 +418,7 @@
   "xjl" 'set-justification-left
   "xjn" 'set-justification-none
   "xjr" 'set-justification-right
+  "xld" 'spacemacs/duplicate-line-or-region
   "xls" 'spacemacs/sort-lines
   "xlu" 'spacemacs/uniquify-lines
   "xtc" 'transpose-chars


nixmaniack:
`C-d` would override default `delete-char` which I assume is much commonly used keybinding in `emacs`/`hybrid` style. So it shouldn't be default.

d12frosted:
Thank you for your contribution. Several comments:

1. `C-d` is a no go, because as @nixmaniack already said, it's important key binding for `emacs` and `hybrid` editing styles. 
2. Having separate configuration variable for second key binding looks redundant, as it just adds additional layer of abstraction, which can be reduced by directly using `global-set-key`.

> `SPC x l d` seems like the most logical but it is quite long

Yes, it's longer than `C-d`, but it seems that line duplication is not among top used functions, so probably it doesn't deserve such a short key binding :D

deb0ch:
Ok ok, I give in ^^

It is one of my very top-used functions, that's why I wanted to propose it. When bound to something quick I find it very powerful.

I wasn't sure if `C-d` was used by anyone because there is `DEL` doing the same and better, as stated in emacs help:

```
C-d runs the command delete-char, which is an interactive built-in function in
`C source code'.

It is bound to C-d.

(delete-char N &optional KILLFLAG)

...

The command `delete-forward-char' is preferable for interactive use, e.g.
because it respects values of `delete-active-region' and `overwrite-mode'.
```
Emacs' own documentation do not advise it for interactive use (why is it even there then ?!) and recommends `DEL` instead.

Having no idea who uses what, I do not know at all if anybody uses `C-d` over `DEL`.

I felt like `C-d` is a good match because it is fast and mnemonic for 'duplicate', but if it conflicts with some people's habits I will just keep it in my own init file :smile:

d12frosted:
> here is DEL doing the same and better

Depends on your keyboard. I am always using laptops, so for me `DEL` is `fn-BACKSPACE`. Which is far from being good. On PC keyboards IIRC `DEL` is somewhere far as well, and you have to move your right hand from home row, which is also not cool. That's why I use `C-d` :D 

> why is it even there then ?!

Haha, funny indeed. `C-d` is mapped to `delete-char` by default in Emacs. 

But I have it bound to `sp-delete-char` which has no drawbacks of regular `delete-char`.

> I do not know at all if anybody uses C-d over DEL.

We should divide `delete-char` vs `delete-forward-char` (where obviously second function is a better choice) and `C-d` vs `DEL` as keybindings (which is hard question). 

> I felt like C-d is a good match because it is fast and mnemonic for 'duplicate'

Yeah, it's good intention. What about `C-c d`? It's also mnemonic and fast to type. Or `C-c C-d`, but it overlaps with `sh-cd-here`.

deb0ch:
I find that `C-c d` or `C-c C-d` are also very good ideas (even if I will keep my own in my own init file :stuck_out_tongue_winking_eye: ).

`C-c d` is already used only in disaster (to disassemble C code at point) and Helm, and `C-c C-d` is used in magit, git-gutter+, gnuplot and slime as local keymaps.

As they are defined as local, these all have higher priority so there would be no conflicts with either of these.

I let you choose which one you prefer, in the mean time I will remove that configuration variable.

Cheers

d12frosted:
> even if I will keep my own in my own init file 😜

haha, it's Emacs after all 😜

> C-c d is already used only in disaster (to disassemble C code at point) and Helm

Didn't know about it. Thanks for bringing it up.

> I let you choose which one you prefer

Personally I prefer `C-c d` (this is what I was using before I realised that I don't use line duplication). But I would like to hear what others think about this key binding. We should be cautious about it :D 

> in the mean time I will remove that configuration variable

👍

deb0ch:
Updated: changed `C-d` to `C-c d` and removed configuration variable.

syl20bnr:
better-defaults layer is for replacing default Emacs key-bindings by enhanced ones. Is is the case here ? I don't use Emacs style but at first sight it seems to not be the case.
It is not necessary to add `SPC x l d` since this is not the purpose of this layer.

d12frosted:
AFAIK, there is no line duplication in Emacs out of box.

P. S. Totally forgot about the purpose of this layer..

deb0ch:
Ok, I did not really understand then, thanks for the explanation.

Should I move it somewhere else then ? like where the other line treatment functions are ?

deb0ch:
Update: now in spacemacs-base along with other `SPC l *` functions.

Also removed `C-c d` binding, still accessible under `SPC l d`.

WarFox:
now, `SPC l` seems to be bound to `layouts-micro-state`

I think @deb0ch original suggestion `SPC x l d`, seems to be best logical default, since it's under text -> line -> duplicate

WarFox:
I've been experimenting this with `s-d` as will, works great for me.

`s-d` was bound to isearch-repeat-backward, which I can live without

d12frosted:
Line duplication is a nice feature. But I think it should go somewhere else, not into `better-defaults` layer. And probably it should be under `SPC x l d`, just like you initially proposed.

deb0ch:
Already moved to spacemacs-base.el and only under `SPC x l d` !

d12frosted:
Oh indeed! That's nice!

d12frosted:
You will not believe it, but I merged 💃 thanks for the PR, it took so long to bring it here 😸

deb0ch:
Yay ! :tada: :confetti_ball: :smiley_cat:

Issue #5333, opened 2016-03-02
magit-status fails with "Unknown upattern `(quote magit-log-mode)'"


brabalan:
#### Description
When starting magit-status I get the error `pcase--u1: Unknown upattern `(quote magit-log-mode)'`

#### Reproduction guide
- Start Emacs
- Go to a git directory
- `SPC g s`

*Observed behaviour:*
I get an error.

*Expected behaviour:*
`magit-status` starts.

#### System Info
- OS: darwin
- Emacs: 24.5.1
- Spacemacs: 0.105.11
- Spacemacs branch: master (rev. 93e55bb)
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
((auto-completion :variables auto-completion-enable-snippets-in-popup t auto-completion-tab-key-behavior 'cycle)
 elfeed emacs-lisp finance git gnus javascript latex ocaml org
 (spell-checking :variables spell-checking-enable-auto-dictionary t))
```

#### Backtrace
```
  signal(error ("Unknown upattern `(quote magit-log-mode)'"))
  error("Unknown upattern `%s'" (quote magit-log-mode))
  pcase--u1(((match major-mode quote magit-log-mode)) #[257 "\302\301\242>\204

fbergroth:
Fixed in https://github.com/magit/magit/commit/637f771893ed3b59036c7f1de687362343da97ba

brabalan:
I updated the packages and everything works now.

PR #6444, opened 2016-06-28
replace apostrophes with backticks inside pcases

korayal:
It looks like when pcase is being used, backticks is the right way to go as mentioned in https://github.com/syl20bnr/spacemacs/issues/5333

On my setup, when I try to open a project in haskell-mode, I get the error below in Messages:

```
File local-variables error: (error "Unknown upattern `(quote ghci)'")
```

### DIFF
--- a/layers/+lang/haskell/config.el
+++ b/layers/+lang/haskell/config.el
@@ -17,7 +17,7 @@
 (spacemacs|defvar-company-backends haskell-cabal-mode)
 (spacemacs|defvar-company-backends intero-repl-mode)
 
-(defvar haskell-completion-backend 'ghci
+(defvar haskell-completion-backend `ghci
   "Completion backend used by company.
 Available options are `ghci', `intero' and `ghc-mod'. Default is
 `ghci'.")
--- a/layers/+lang/haskell/funcs.el
+++ b/layers/+lang/haskell/funcs.el
@@ -15,9 +15,9 @@
     (add-hook 'haskell-mode-hook 'interactive-haskell-mode))
   (when (configuration-layer/package-usedp 'company)
     (pcase haskell-completion-backend
-      ('ghci (spacemacs-haskell//setup-ghci))
-      ('ghc-mod (spacemacs-haskell//setup-ghc-mod))
-      ('intero (spacemacs-haskell//setup-intero)))))
+      (`ghci (spacemacs-haskell//setup-ghci))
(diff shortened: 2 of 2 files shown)


d12frosted:
Hm. Recently I also faced similar problem. Didn't know that backquotes should be used instead.

a13ph:
@korayal You [confusingly](http://www.cs.sfu.ca/~ggbaker/reference/characters/) use wrong term, [backquote and backtick are the same character `] (http://english.stackexchange.com/questions/181810/backquote-vs-backtick). 
`'` is apostrophe. Or, informally and historically, single quote.

korayal:
@a13ph thanks for the correction :) I assumed the information in [this commit](https://github.com/magit/magit/commit/637f771893ed3b59036c7f1de687362343da97ba)'s title is correct. I'm fixing the title now.

puffnfresh:
@syl20bnr was this merged to develop? I'm getting an error message from this problem.

d12frosted:
@puffnfresh I believe @syl20bnr merged it locally, but didn't push yet, so just wait a little bit 😸

layers/+lang/haskell/config.el
```diff
@@ -17,7 +17,7 @@
 (spacemacs|defvar-company-backends haskell-cabal-mode)
 (spacemacs|defvar-company-backends intero-repl-mode)
 
-(defvar haskell-completion-backend 'ghci
+(defvar haskell-completion-backend `ghci
```
d12frosted: Shouldn't it be `'ghci`?

syl20bnr: I merged it locally and I made the fix. I will push later today.

d12frosted: Thanks 😸 Didn't know that you made a fix, no hurry 😸 Just wanted to make sure that it comes to develop without any issues 💃

syl20bnr: No problem, I could have overlooked it very easily. Thank you!

korayal: thanks, I think they work all the same (I didn't have any issues on this case) but, as you said, the convention seems to be the backticks(`)  are used only when it's needed specifically. (as in the `pcase` use here)

syl20bnr:
Thank you ! 👍
I made the fix mentiond by @d12frosted in commit https://github.com/syl20bnr/spacemacs/commit/15c392b 

PR #6453, opened 2016-06-28
Add hook to ediff org files unfolded

mazinbokhari:
Add hook to ediff org files unfolded instead of folded as is the current implementation.


mazinbokhari:
I would have made this configurable, but I really don't think there is ever a reason to have an ediff folded. I often have crowd sourced notes for courses and such, and often run into this prolbme.

syl20bnr:
Nice, makes sense 👍 
Thank you !

PR #6457, opened 2016-06-29
fix home buffer random banner choice

deb0ch:
Random banner choice was broken by the presence of the `img` directory.
Even if `random` was set instead of `random*` in .spacemacs file, the
`cate` banner would still show (but not the `doge` banner).

This commit fixes this bug while simplifying the existing code by adding
a regexp filter to only choose "*.txt" files.


syl20bnr:
Nice ! 💯 
To be honest I knew about this bug for a long time but never got the courage to look at it and it was fun to have people get the doge sometimes without digging the code :-D
So now it is back to normal and more robust, great!

Thank you !

PR #6463, opened 2016-06-30
[shell-scripts] add note about shellcheck

d12frosted:
it might be obvious for some users, but I was using this layer for a long time and only today by accident found this gem.


ikhthiandor:
Man, I found through you. This is really awesome. I've been looking for something to make writing and managing shellscripts a bit saner.

syl20bnr:
Thank you!

Issue #5863, opened 2016-04-22
Add installation notes about Source Code Pro


d12frosted:
We don't mention that default `.spacemacs` uses Source Code Pro. And people run into problems. We should update installation notes. 

For example: https://github.com/syl20bnr/spacemacs/issues/3477#issuecomment-207830837

d12frosted:
Fixed with released of Spacemacs v0.200. Let us know if you still have any problems with this issue.

Issue #2864, opened 2015-09-02
Spacemacs not loading font on OSX


afabbro:
I'm pretty confident that I'm followed the instructions for the Powerline-patched fonts. They're on my system and available in other applications.

However, I get:

(Spacemacs) Warning: Cannot find font "Sauce Code Powerline"! 

when running Spacemacs and it hits the default font setting line.

Where does spacemacs (and/or powerline) look for the fonts?

d12frosted:
Are you trying it in terminal? Then you need to set it in terminal preferences for your current profile.

timstott:
Facing identical issue here. I'm using spacemacs from the terminal (iTerm2 | OS X).

![screen shot 2015-10-16 at 10 57 00](https://cloud.githubusercontent.com/assets/1334474/10539039/acbe49a6-73f4-11e5-8235-b928bedbed7c.png)

```
;; .spacemacs
dotspacemacs-default-font '("Droid Sans Mono for Powerline"
                            :size 12
                            :weight normal
                            :width normal
                            :powerline-scale 1.3)  
```

@d12frosted any thoughts? :cake:

d12frosted:
@timstott 
In the terminal this warning is expected. Just ignore it. :smile:

StreakyCobra:
We should probably document somewhere that the font in terminal mode is set by the terminal and can not be changed by emacs.

d12frosted:
I would vote for both.

d12frosted:
Fixed with released of Spacemacs v0.200. Let us know if you still have any problems with this issue.

PR #6464, opened 2016-06-30
improve font documentation

d12frosted:
- make it clear that setting dotspacemacs-default-font has no effect in
  terminal
- make it clear that it's user responsibility to install Source Code Pro
  font

Related issues: #2864, #5863


syl20bnr:
Thank you !

PR #6467, opened 2016-06-30
New layer: +misc/smart-tabs

curtmack:
This layer streamlines the process of setting up smart-tabs-mode with Spacemacs; it automates disabling Evil's built-in tab replacement and enabling indent-tabs-mode for supported languages (while leaving those settings alone in other buffers).

This is my first contributed layer, so please let me know if there's any style or naming issues that need to be corrected.

Make what you want of the introductory text, the joke had to be made.

### DIFF
--- a/layers/+misc/smart-tabs/README.org
+++ b/layers/+misc/smart-tabs/README.org
@@ -1,6 +1,7 @@
 #+TITLE: Smart Tabs layer
 
-# TOC links should be GitHub style anchors.
+The best whitespace is neither spaces nor tabs, it's spaces /and/ tabs!
+
 * Table of Contents                                        :TOC_4_gh:noexport:
  - [[#description][Description]]
  - [[#install][Install]]
@@ -10,18 +11,20 @@
    - [[#respecting-evil-customizations][Respecting evil customizations]]
 
 * Description
-The best whitespace is neither spaces nor tabs, it's spaces *and* tabs!
-
-This layer brings =smart-tabs-mode= integration, a minor mode that intelligently
-indents your code with tabs, while preserving spaces for alignment and ASCII
-art. In this way, multiple team members can choose their preferred tab length
-without breaking alignment. To enable =smart-tabs-mode= for a given language,
-you *insinuate* it (see instructions below, under [[*configuration][Configuration]]).
+This layer installs =smart-tabs-mode=, a minor mode that intelligently indents
+your code with tabs, while preserving spaces for alignment and ASCII art. In
+this way, multiple team members can choose their preferred tab length without
+breaking alignment of other code elements. To enable =smart-tabs-mode= for a
+given language, you /insinuate/ it (see instructions below, under
+[[#configuration][Configuration]]).
 
 If =indent-tabs-mode= is set to nil, you'll always indent with spaces, even if
 =smart-tabs-mode= is on. Nevertheless, it is recommended that =indent-tabs-mode=
 is left =nil= by default; the layer will handle turning it on for the languages
-you insinuate. =smart-tabs-mode= will use =tab-width= as its indent offset.
+you insinuate.
+
+Note that =smart-tabs-mode= will use =tab-width= as its indent offset,
+overriding the normal indent offset for whatever major mode is active.
 
 Please see [[https://www.emacswiki.org/emacs/SmartTabs][Smart Tabs on the Emacs Wiki]] if you would like more information on
 the underlying mode.
@@ -53,19 +56,19 @@ By default, none of these will be insinuated. To use =smart-tabs-mode= with any
 #+BEGIN_SRC emacs-lisp
   (setq-default dotspacemacs-configuration-layers '(
     (smart-tabs :variables
-                smart-tabs-default-insinuations '(c c++ cperl javascript java))
-   ))
+                smart-tabs-default-insinuations '(c c++ cperl javascript java))))
 #+END_SRC
 
 ** Adding support for additional languages
-You should use the new =smart-tabs-add-language-and-insinuate= macro to add
-support for additional language. You can call this in your
-=dotspacemacs/user-config= function to enable support for any other languages
-you use. The arguments are:
+You can use the =smart-tabs-add-language-and-insinuate= macro in your
+=dotspacemacs/user-config= to add support for additional languages. The
+arguments are:
 
- - A symbol to represent the language. This value is mostly unimportant; the
-   only rule is that you can't use the same symbol twice.
+ - A symbol to represent the language. It's recommended to use the name of the
+   language's major mode minus the =-mode= suffix, but the actual value is
+   unimportant; the only rule is that you can't use the same symbol twice.
  - The mode hook for the major mode you want to associate with this language.
+   Hooks will be installed to enable =smart-tabs-mode= and =indent-tabs-mode=.
  - An alist of indent rules. Each rule is a cons cell containing an indent
    function followed by the indent offset variable used by that function.
  - Optionally, any number of body parameters to be executed in the language's
@@ -76,21 +79,19 @@ Here's an example that adds support for Groovy via =groovy-mode=.
 #+BEGIN_SRC emacs-lisp
   (defun dotspacemacs/user-config ()
 
-  ;; ...
+    ;; ...
 
-    (smart-tabs-add-language-and-insinuate groovy groovy-mode-hook
-     ;; Note that Groovy mode, like many other C-like languages, simply reuses
-     ;; Emacs' core C indentation commands
+    (smart-tabs-add-language-and-insinuate
+     groovy
+     groovy-mode-hook
      ((c-indent-line . c-basic-offset)
       (c-indent-region . c-basic-offset))))
 #+END_SRC
 
 ** Respecting evil customizations
-Because =evil= uses its own indentation functions, by default =smart-tabs-mode=
-will not work with evil's indentation operator, ~=~. It's possible to work
-around this by making buffer-local changes to =evil-shift-width= and
-=evil-indent-convert-tabs=, two customizations provided by the evil package. By
-default, the layer will handle this for you; however, if you have customized
+Normally, the layer will make buffer-local changes to =evil-shift-width= and
+=evil-indent-convert-tabs=, two customizations provided by the evil package, in
+order to make evil's indentation operators play nice. If you have customized
 those values yourself and want the layer to respect your values, you can disable
 this behavior by setting the layer variable
 =smart-tabs-respect-evil-customizations= to =t=.
--- a/layers/+misc/smart-tabs/README.org
+++ b/layers/+misc/smart-tabs/README.org
@@ -50,8 +50,8 @@ this file.
 | XML        | =nxml-mode=   | =nxml=       |
 
 By default, none of these will be insinuated. To use =smart-tabs-mode= with any
-   of these languages, set the layer variable =smart-tabs-default-insinuations=
-   to a list of any or all of the symbols listed above.
+of these languages, set the layer variable =smart-tabs-default-insinuations=
+to a list of any or all of the symbols listed above.
 
 #+BEGIN_SRC emacs-lisp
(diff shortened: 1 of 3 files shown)


layers/+misc/smart-tabs/README.org
```diff
@@ -0,0 +1,103 @@
+#+TITLE: Smart Tabs layer
+
+The best whitespace is neither spaces nor tabs, it's spaces /and/ tabs!
```
d12frosted: I like it. 💃

layers/+misc/smart-tabs/config.el
```diff
@@ -0,0 +1,44 @@
...
+;;; Code:
+
+;;; Utility function that simply enables indent-tabs-mode
+;;; This is used as an argument to add-hook
+(defun smart-tabs/enable-tabs-mode ()
+  (setq indent-tabs-mode t))
+
+;;; Macro that combines adding new language support, insinuating it, and hooking
+;;; indent-tabs-mode in one place
+(defmacro smart-tabs-add-language-and-insinuate (lang mode-hook advice-list &rest body)
```
bmag: spacemacs convention is for macros to have a `|` separator: `smart-tabs|add-language-and-insinuate`

layers/+misc/smart-tabs/packages.el
```diff
@@ -0,0 +1,41 @@
...
+(defconst smart-tabs-packages
+  '(smart-tabs-mode))
+
+(defun smart-tabs/init-smart-tabs-mode ()
+  (use-package smart-tabs-mode
+    :diminish smart-tabs-mode
+    :config
+    (progn
+      (add-hook 'smart-tabs-mode-hook
+                (lambda ()
```
bmag: it's good practice to name functions that go into a hook. Instead of:
```elisp
(add-hook <some-hook> (lambda () ...))
```
Is better to use:
```elisp
(defun whatever () ...)
(add-hook <some-hook> #'whatever)
```

curtmack:
I've made those changes; I also renamed an internal function to use the `//` convention as well. There was also a function that I had written initially, that didn't end up getting used, so I removed that.

layers/+misc/smart-tabs/README.org
robbyoconnor: I too love this :dancer:

skebanga:
What's the status of this please?

d12frosted:
Ping, @syl20bnr :D

SteveJobzniak:
@curtmack Wow, thanks a lot for creating this!

@syl20bnr This isn't fully ready to merge yet, but it's a GREAT package.

First I'll demonstrate what it does:

**Indenting with normal spaces (we all know and love/hate those):**

01-spaces

**Indenting with regular Emacs tabs (notice how it TERRIBLY assumes tab-width=4 always, and adds tabs until the required column, then fills in with spaces to get the "proper" but non-portable alignment):**

02-regulartabs_tabwidth4

**What regular Emacs tabs created at tab-width 4 look like when they are viewed in a text editor with tab-width 2 instead (UH OH! ALIGNMENT GOES TO HELL!):**

03-regulartabs_tabwidth2

**Smart tabs enabled, note how it indents with one tab per indentation level, and then uses spaces for alignment (this example was created at tab-width 4):**

04-smarttabs_tabwidth4

**Smart-tabs from above, viewed by someone at tab-width 8 instead:**

05-smarttabs_tabwidth8

**Smart-tabs from above, viewed by someone at tab-width 2 instead:**

06-smarttabs_tabwidth2

As you can all see, Smart Tabs are incredibly awesome. They let you use Tab (for more efficient, fluidly changeable indentation) but still perfectly align variables.

The only slight downside is that smart-tabs doesn't have a lot of language support built in. So I haven't yet researched how to create a hook for PHP. But it's possible to support any language by writing your own hooks.

SteveJobzniak:
Now for the not-so-great problems with this package, which need to be fixed before it's good enough to be merged:

**Highlighting "erroneous, mixed whitespace" is enabled:**

See the images above for the big, yellow bars in <SPC t w> whitespace mode. Not sure if this is fixable. But it would be great if it's possible to tell all smart-tab enabled languages to never highlight <tab><spaces> as wrong anymore.

**And the most serious one: >> and << operator in Evil mode are broken.**

Before using >>:

broken-01

After using >>:

broken-02

I know that you very nicely fixed the "=" auto-indent operator. But I am not sure this one can be fixed as easily. I think the code should be something like:

"indent by 1" = "add 1 tab to the start of each line to be affected"
"unindent by 1" = "remove 1 tab from the start of each line to be affected; do NOT affect spaces"

@curtmack Do you think you can do that? I have no idea how to fix these bugs, sorry.

SteveJobzniak:
@d12frosted @robbyoconnor @skebanga 

For you and anyone else who wants to use the preliminary layer as @curtmack wrote it so far, there are two options. Both will place it in your "private" directory so that it doesn't break Spacemacs update mechanism.

Clone my copy of it:
```bash
cd ~/.emacs.d/private
git clone https://github.com/SteveJobzniak/spacemacs-private-smart-tabs.git smart-tabs
```

Create your own instance of everything in this current pull request (this is identical to my repo, at the moment):
```bash
cd ~/.emacs.d/private
mkdir smart-tabs
curl -L -O "https://github.com/syl20bnr/spacemacs/pull/6467.patch"
patch -p4 < 6467.patch
rm -f 6467.patch
```

Then just add it to your list of active layers. You can view help with `<SPC h l>` and typing `smart-tabs` to see how to configure it.

Here's my configuration which enables all of the built-in language supports:

In "dotspacemacs-configuration-layers":
```lisp
     (smart-tabs :variables
                 smart-tabs-default-insinuations '(c c++ cperl javascript java python ruby nxml))
```

In "dotspacemacs/user-config":
```list
  "Smart-tabs: Use spaces globally, then enable smart tabs per-language"
  (setq-default indent-tabs-mode nil)
```

SteveJobzniak:
The highlight-warning "bug" can be overlooked since most people do not run in `<SPC t w>` mode.

But here's the code that _badly needs_ to be overridden (evil-indent was already fixed by @curtmack, but the shifts aren't):

minor_modes_todo

SteveJobzniak:
@curtmack Found one more bug in this layer.

You use "setq-local evil-shift-width tab-width". I am not fully sure where or how that runs (I descend from a long line of Vim users), but the problem is that tab-width is later modified individually by various language modes. Python for example sets it to 4 when a new document is loaded.

As a result of that problem, my Python buffer has these variable values:

```
tab-width is a variable defined in ‘C source code’.
Its value is 4
Original value was 8
```

```
evil-shift-width is a variable defined in ‘evil-vars.el’.
Its value is 8
Original value was 4
```

As you can see, evil-shift-width has been set to the original value of tab-width from before the major mode was loaded. So if I use >> or <<, it shifts by 8 spaces (2 tabs) instead of 4 as expected by the document.

Edit: Actually... this is serious but it may only affect python-mode. When python-mode opens a document, it scans it to see if the author used 2, 4, 8 etc spaces, and then it sets tab-width to the same value to ensure that the code will work properly. If you open a new document it says "[Can't guess python-indent-offset, using defaults: 4](http://stackoverflow.com/questions/18778894/emacs-24-3-python-cant-guess-python-indent-offset-using-defaults-4)". So Python-mode itself changes the tab-width setting _after_ you've told Evil to use the earlier value.


------


So yeah, 3 bugs:

- Highlighting whitespace errors. May not be possible/worth fixing since almost nobody codes with whitespace visible all the time. And the highlighting can be seen as a "feature" to ensure smart-tabs are working.
- Evil shifting is not working. Seems to need a total rewrite since there are no evil environment variables for controlling shift behavior of tabs/space selection, as far as I can see. Maybe this can be fixed upstream in evil-shift, just like "evil-indent-convert-tabs nil" exists for regular "=" auto-indenting.
- The way that the layer's "evil-setup" function affects "evil-shift-width" needs to be improved so that it always matches tab-width of the document, or _at least_ so that it's delayed until after the document is fully loaded so that it doesn't use an outdated value.

SteveJobzniak:
**So yeah, there are 3 bugs preventing merging right now:**

- Highlighting whitespace errors. May not be possible/worth fixing since almost nobody codes with whitespace visible all the time. And the highlighting can be seen as a "feature" to ensure smart-tabs are working.
- Evil shifting is not working. Seems to need a total rewrite since there are no evil environment variables for controlling shift behavior of tabs/space selection, as far as I can see. Maybe this can be fixed upstream in evil-shift, just like "evil-indent-convert-tabs nil" exists for regular "=" auto-indenting.
- The way that the layer's "evil-setup" function affects "evil-shift-width" needs to be improved so that it always matches tab-width of the document, or _at least_ so that it's delayed until after the document is fully loaded so that it doesn't use an outdated value.

SteveJobzniak:
I tried to research the shift-bug, and looked at all indentation commands Emacs has built-in:

https://www.gnu.org/software/emacs/manual/html_node/elisp/Region-Indent.html

The closest thing to what we need is "indent-rigidly start end count" which moves a region of text right or left by the desired number of columns. So if you tell it to move 4 right (and tabstop is set to 4) then that function will add one tab. The problems are: 1) It inserts spaces if you don't give it the exact number of columns matching one tab. 2) It converts the rest of the line to tabs!

So it looks like the last two remaining solutions for the shift bug are:
- Option 1: Try to get Evil to give us a shift mode which doesn't convert spaces on the line. This is unlikely to happen.
- Option 2: Write our own replacement for `evil-shift-right` (all other evil shift functions refer to that one's code).

Here's the idea I have for that (but I am not going to do it, I am not good enough at Lisp yet):

```lisp
(eval-after-load "evil"
  'save a copy/reference of the original evil-shift-right somehow
  '(evil-define-operator evil-shift-right (parameters here...)
     ...))
```

Then we'd do this:
1. Add the above injection method to the smart-tabs package. Which replaces `evil-shift-right` with our own code instead, but saves a reference to the original code.
2. On a per-buffer basis, whenever our modified `evil-shift-right` runs, it checks: Is smart-tabs disabled for this document? Then run the original `evil-shift-right`. If smart-tabs enabled? Then use our code instead...
3. In our code, we'd simply add X number of `\t` tab characters to the left of the desired lines, if we're indenting by X. And we'd simply remove X number o `\t` characters to the left of the desired lines, if we're unindenting by X.
4. We never, ever remove spaces when de-indenting. Only tabs.
5. With this, evil shifting would be working kinda-as-intended in smart-tabs mode. Unlike the normal shifting, it wouldn't fix badly indented lines. But all we'd have to do is `==` to first indent broken lines normally, and THEN (or if the document was already fine) we can start using `>>` and `<<` to shift the tab indentation to the right or left.

SteveJobzniak:
Does anyone want a small, fun lisp coding task to help out with the development of this spacemacs layer? I am new to lisp.

We need a `tabshift` function which takes 3 arguments: `firstline, lastline, tabshiftamount`.

For each of the lines of code to process, do:
- If `tabshiftamount` is >= 1, insert that many `\t` leading tab characters at the start of all of those lines.
- If `tabshiftamount` is <= -1, remove that many `\t` leading tab characters at the start of all of those lines.
- For any lines that don't have enough tabs to remove at the start, look for spaces instead and treat each chunk of `tab-width` (the per-buffer variable) spaces as one tab and remove those.
- So if a line has `<tab><7 spaces>` and it's told to shift -2 tabs, it would remove the tab and 4 spaces (if `tab-width` was set to 4).
- In both the indent and unindent cases, always preserve all other leading whitespace on the line exactly as-is.

If someone is skilled enough at lisp to come up with this, then we'll be on our way to finishing this smart-tabs layer!

PrgrmAtCeleritas:
I also noticed that if you have ``dotspacemacs-whitespace-cleanup 'changed`` Spacemacs will overwrite the alignment of spaces with tabs. May need to look into that.

SteveJobzniak:
@PrgrmAtCeleritas Nice discovery. So yeah there's a list of about 3-4 things that need to be fixed before this is stable.

For now I've gone back to 100% spaces instead of tabs. Someday in the future I would like to be a Smart-Tabs user. ;-)

PR #6472, opened 2016-07-01
fix some function names in space-doc

d12frosted:
It seems that naming was changed and it totally broke several markers.

Thanks to @JAremko for finding it.

But it's still not ideal and throwing some errors. For example,

```elisp
Error during redisplay: (jit-lock-function 1) signaled (wrong-type-argument arrayp nil)
Error during redisplay: (jit-lock-function 15168) signaled (wrong-type-argument arrayp nil)
```

Also, using `space-doc-mode` on Documentation doesn't look very good :/


JAremko:
@d12frosted  pushed the rest of the fixes https://github.com/JAremko/spacemacs-pr/commit/55ac19913f36cf5202daeb4b582e92ab0f89d57e  can youcherry-pick them?

syl20bnr:
Thank you ! 👍 
And sorry for my bad pushes :-(

PR #6477, opened 2016-07-02
Update `brew linkapps` to be more specific

jkrmr:
Users may not want to link all possible apps and doing so isn't required to link Spacemacs individually. (I accidentally linked all my apps because I always forget that Homebrew linkapps can accept a list of formulae 😬)


d12frosted:
Oh, good point 😸 

PR #6484, opened 2016-07-03
Adds elm-format section and advice for absolute paths with npm

benansell:
- Use of relative paths for npm global bin directory caused the following error in the browser when trying to preview a buffer with elm-reactor:

`A web handler threw an exception. Details:`
`elm-make: readCreateProcessWithExitCode: runInteractiveProcess: exec: does not exist (No such file or directory)`

- Added a section for elm-format so that hints on how to use homebrew can be given


syl20bnr:
Thank you ! 👍 

PR #6491, opened 2016-07-04
colorize compilation buffers by processing ansi color sequences

deb0ch:
By default, compilation buffers do not handle ansi color SGRs and print them instead, making the output messy and not readable.

This commit brings some color and readability to compilation buffers by treating ansi color SGRs.

The approach taken here uses [the function](https://github.com/emacs-mirror/emacs/blob/master/lisp/ansi-color.el#L233) used in [the function used in comint-mode](https://github.com/emacs-mirror/emacs/blob/master/lisp/comint.el#L389).

### DIFF
--- a/layers/+spacemacs/spacemacs-ui-visual/packages.el
+++ b/layers/+spacemacs/spacemacs-ui-visual/packages.el
@@ -10,7 +10,8 @@
 ;;; License: GPLv3
 
 (setq spacemacs-ui-visual-packages
-      '(fancy-battery
+      '((ansi-colors :location built-in)
+	fancy-battery
         fill-column-indicator
         golden-ratio
         hl-todo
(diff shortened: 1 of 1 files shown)


layers/+emacs/better-defaults/packages.el
```diff
@@ -10,9 +10,19 @@
...
+  '((ansi-colors :location built-in)
+    mwim)
   "The list of Lisp packages required by the mwim layer.")
 
+(defun better-defaults/init-ansi-colors ()
+  :init
+  (progn
+    (defun better-defaults/compilation-buffer-apply-ansi-colors ()
+      (setq inhibit-read-only t)
+      (ansi-color-apply-on-region compilation-filter-start (point-max))
```
deb0ch: **NB:** I used `point-max` here, but it also worked using `point`. I don't know if it is of any importance or if it might impact performance. Insights welcomed.

syl20bnr:
Nice, I would put this in the layer `spacemacs-ui-visual` instead of `better-defaults` which is more reserved for improved default emacs key bidnings.

layers/+emacs/better-defaults/packages.el
```diff
@@ -10,9 +10,19 @@
...
+      (setq inhibit-read-only nil))
```
syl20bnr: Use a let binding for `inhibit-read-only` instead of globally modifying the variable with two `setq`, example here: https://github.com/syl20bnr/spacemacs/blob/develop/layers/+tools/shell/packages.el#L137-L138

deb0ch:
Aye, working on it:

- use `let` instead of `setq`
- integrate into `spacemacs-ui-visual`

Thanks for feedback !

deb0ch:
Done

syl20bnr:
Thank you !
I cleaned up the PR a bit in commit https://github.com/syl20bnr/spacemacs/commit/939cbb6

PR #6508, opened 2016-07-06
Add Dropbox user dir to load path only if it exists

madand:
Currently the item `~/Dropbox/emacs` is added to the `load-path`, even if that directory not exists in the filesystem. This may couse troubles. For example, the following code:
```emacs-lisp
(require 'url-about)
(url-probe-protocols)
```
triggers an error:
``` emacs-lisp
Debugger entered--Lisp error: (file-error "Opening directory" "no such file or directory" "/home/madand/Dropbox/emacs/")
  directory-files("/home/madand/Dropbox/emacs/" nil "^url-.*\\.el$")
  #[(d) "\301\302\303\304\305#\"\207" [d mapc #[(f) "\302\303\"\205

### DIFF
--- a/core/core-load-paths.el
+++ b/core/core-load-paths.el
@@ -10,6 +10,11 @@
 ;;; License: GPLv3
 (defun add-to-load-path (dir) (add-to-list 'load-path dir))
 
+(defun add-to-load-path-if-exists (dir)
+  "If DIR exists in the file system, add it to `load-path'."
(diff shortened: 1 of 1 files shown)


syl20bnr:
Thank you ! 👍 
I forgot about this variable and I just removed it in commit https://github.com/syl20bnr/spacemacs/commit/7e82535 since it makes no sense to have DropBox mentioned in the default config.

PR #6509, opened 2016-07-06
Run `diff-hl-magit-post-refresh` after Magit commit

bixuanzju:
Otherwise `diff-hl-mode` still highlights uncommitted changes. See https://github.com/dgutov/diff-hl#magit


syl20bnr:
Thank you ! 👍 

PR #6525, opened 2016-07-07
fix chinese layer readme

zilongshanren:


d12frosted:

PR #6527, opened 2016-07-07
fix call to show-all

### LINKED ISSUE #6500: `outline-show-all` void when using `ediff`
When trying to use `ediff` through `SPC f e D` I'm getting the error: `Symbol function definition is void: outline-show-all`.

This bug might have been introduced recently. It happened just after I pulled from `develop` today.

#### System Info :computer:
- OS: darwin
- Emacs: 24.5.1
- Spacemacs: 0.105.21
- Spacemacs branch: develop (rev. a650877)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(helm html auto-completion emacs-lisp git github haskell html java markdown osx
      (shell :variables shell-default-shell 'eshell shell-enable-smart-eshell t)
      scala spell-checking syntax-checking version-control yaml)
```

d12frosted:
As @syohex pointed out, `outline-show-all` was introduced in Emacs 25.1. Shouldn't we use `show-all` instead which is aliased to `outline-show-all` in Emacs 25.1 (not by Spacemacs though, but in `outline.el`).

Also, am I missing relation between `org` and `outline`? I mean, why `(with-eval-after-load 'org ...)`? Because of this I moved code back to `spacemacs-base` layer.

Fixes #6500

### DIFF
--- a/layers/+distributions/spacemacs-base/packages.el
+++ b/layers/+distributions/spacemacs-base/packages.el
@@ -126,6 +126,9 @@
        ;; emacs is evil and decrees that vertical shall henceforth be horizontal
        ediff-split-window-function 'split-window-horizontally
        ediff-merge-split-window-function 'split-window-horizontally)
+      ;; show org ediffs unfolded
+      (require 'outline)
+      (add-hook 'ediff-prepare-buffer-hook #'show-all)
       ;; restore window layout when done
       (add-hook 'ediff-quit-hook #'winner-undo))))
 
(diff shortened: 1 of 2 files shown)


d12frosted:
Cherry-picked.

NJBS:
Sorry, just saw these messages.

So the relation between org and outline-show-all as that the org-compat package defines an alias for outline-show-all for Emacs 24.5. Since the hook was only intended to be triggered for org files I figured only adding the hook after org was loaded (and the compatibility alias was defined) would be better than adding in another require especially if the user wasn't intending on diffing an org file to begin with.

This should fix the issue regardless :+1:

d12frosted:
@NJBS 👍 thanks for explanation

PR #6529, opened 2016-07-07
space-doc/centered-buffer improvements.

JAremko:
Slightly reduced amount of hacks and moved them in the place where there is less chance that someone notice them. Also made dedicated window noop for `centered-buffer-mode` as @syl20bnr  hinted.

### DIFF
--- a/layers/+distributions/spacemacs-base/local/centered-buffer-mode/centered-buffer-mode.el
+++ b/layers/+distributions/spacemacs-base/local/centered-buffer-mode/centered-buffer-mode.el
@@ -120,12 +120,12 @@ Assume to be called interactively when INTERACT has non nil value."
                 (face-remap-add-relative 'fringe :background
                                          spacemacs-centered-buffer-mode-fringe-color)
                 (set-window-buffer window indirect-buffer)
-                (advice-add 'spacemacs/previous-useful-buffer
+                (advice-add 'previous-buffer
                             :before
-                            #'spacemacs//centered-buffer-mode-prev-next-useful-buffer-advice)
-                (advice-add 'spacemacs/next-useful-buffer
+                            #'spacemacs//centered-buffer-mode-prev-next-buffer-advice)
+                (advice-add 'next-buffer
                             :before
-                            #'spacemacs//centered-buffer-mode-prev-next-useful-buffer-advice)
+                            #'spacemacs//centered-buffer-mode-prev-next-buffer-advice)
                 (add-hook 'buffer-list-update-hook
                           'spacemacs//centered-buffer-buffer-list-update-fringes)
                 (add-hook 'window-configuration-change-hook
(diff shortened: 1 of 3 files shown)


layers/+distributions/spacemacs-base/local/centered-buffer-mode/centered-buffer-mode.el
```diff
@@ -46,82 +46,136 @@ that differed modifications won't cause an overflow."
...
+    (setq spacemacs-centered-buffer-mode
+          nil
+          spacemacs--centered-buffer-mode-indirect-buffer
+          indirect-buffer)
+    (when (derived-mode-p 'org-mode)
+      (setq-local org-startup-folded nil)
+      (outline-show-all))
+    (switch-to-buffer indirect-buffer nil t)
+    (with-mode-disabled page-break-lines-mode
+                        (let* ((fringe-w (spacemacs//centered-buffer-calc-fringe
```
syl20bnr: better to put `page-break-lines-mode` on a new line to avoid the huge indentation.

JAremko: @syl20bnr  done.

JAremko:
Hm.. Fond something strange. Gotta test it more :confused:

JAremko:
@syl20bnr  ok looks good :shipit:

JAremko:
Simplified `space-doc`  behavior customization, improved documentation. Refactored the code a bit.

@syl20bnr  :up:

JAremko:
There are two more thing on my mind:

  - Replace overlays with font-lock. It will improve performance (not sure if it's a problem right now, but still faster = better) Also it will remove the need in `advice-org-do-emphasis-faces` (it should be hidden as a part of the implementation anyway)
  -  Add .spacemacs config. Something like ` dotspacemacs-center-documentation`   With the default value of the minimum required window(or frame) width for it to be enabled. It also can be `nil` - disable.

a13ph:
i wouldn't say faster perf is always better, without taking context and actual thresholds/constraints into account.
also, idk if relevant, but there's now ~ `jit-fontlock`

JAremko:
@a13ph Let's put it this way. When slower is better? :shipit:

JAremko:
MELPA NOO!!!111  :cry:

d12frosted:
It's been very unstable lately 😟

JAremko:
@TheBB  I think we better merge it - it fixes stuff :shipit:

TheBB:

TheBB:
I still think we need to discuss centering by default, though.

nixmaniack:
> I still think we need to discuss centering by default, though.

FWIW, I'm not getting centering by default (I think after this update) even though I have it enabled. Might file an issue later if needed.

JAremko:
@nixmaniack  I [added](https://github.com/syl20bnr/spacemacs/pull/6529/files#diff-15b3d16ce5d3935442dbbd4974b6f25fR85  ) minimal Emacs frame aspect ration configuration. First I was planning to put it into dotfile, but we probably want to keep it small and I'm not sure how important this configuration is.  Probably should document it in the FAQ :confused: 

@TheBB  Thanks for finding time! I know you are busy :heart: 

>   I still think we need to discuss centering by default, though.

I have an idea.

nixmaniack:
@JAremko Okay. It looked like a bug when documentation wasn't centering. Thanks for the pointer to configuration.

PR #6530, opened 2016-07-08
Update ensime buffer name

Astrac:
The name of the buffer for Ensime changed in the latest releases so the spacemacs integration layer is trying to open a new instance for every file you open. This should fix it.

### DIFF
--- a/layers/+lang/scala/funcs.el
+++ b/layers/+lang/scala/funcs.el
@@ -39,7 +39,7 @@
   (let ((default-directory (file-name-directory file)))
     (-when-let (project-name (projectile-project-p))
       (--first (-when-let (bufname (buffer-name it))
-                 (and (s-contains? "inferior-ensime-server" bufname)
+                 (and (s-contains? "ENSIME-backend" bufname)
                       (s-contains? (file-name-nondirectory project-name) bufname)))
                (buffer-list)))))
 


reactormonk:
Could this PR be merged? Without it, spacemacs will try to start a new ensime instance for each buffer, which basically disables ensime for that buffer, because it doesn't connect to the existing instance.

diego-alvarez-hs:
autostart option ON by default is annoying.

I can send a PR making it default OFF but not sure how many are interested in it.

Astrac:
Personally I would prefer it to be off by default as well; could you share the setting to turn it off by configuration please?

d1egoaz:
@astrac https://github.com/syl20bnr/spacemacs/tree/develop/layers/%2Blang/scala#disable-auto-start

Astrac:
I just discovered that this doesn't work, the buffer name is actually `ENSIME-[project name]`, the `backend` bit is the unfortunately ambiguous name of the project I was working on when I discovered the issue.

Astrac:
@d1egoaz thanks for the pointer but that doesn't seem to work and there is no reference to that key in the code as far as I can tell so maybe that piece of the docs is a bit out of sync.

diego-alvarez-hs:
@Astrac weird, here is used in the code https://github.com/syl20bnr/spacemacs/blob/9773ad611f6f03adc9d1fe80072537015861ac8f/layers/%2Blang/scala/packages.el#L35
And it worked for me :shrug:

Are you in the develop branch?

d12frosted:
> I just discovered that this doesn't work, the buffer name is actually ENSIME-[project name], the backend bit is the unfortunately ambiguous name of the project I was working on when I discovered the issue.

Hey, sorry for the delay. Could you please fix this in this PR?

diego-alvarez-hs:
I just created the PR for disable autostart https://github.com/syl20bnr/spacemacs/pull/6674

d1egoaz:
@Astrac might want to take a look at how ensime generates the buffer name, I think it is defined here https://github.com/ensime/ensime-emacs/blob/master/ensime-startup.el#L182

Astrac:
@d1egoaz the buffer name is given by `*ENSIME-{projectName}*`; the hard bit I think is finding out the project name as that's something that is provided in the `.ensime` file. We could just check for buffers starting with `*ENSIME-` but:

- I think that if you track the development branch ensime will create multiple of those buffer
- It won't start different ensime servers for different projects

None of those issues are really big but I suspect that the best fix for all this problems really is just not auto-starting ensime!

reactormonk:
I would also suggest not auto-starting ensime, because it tends to take up quite some memory, so you'll have to make sure you don't have too many instances running.

d12frosted:
Hey @Astrac,

As a side note, I don't use Ensime. Just looking at the code linked by @d1egoaz I think that it's possible to get proper name by using something their functions. For example, I see it loads `config-file` using `ensime-config-load` and then it gets interactive buffer name by `(concat ensime-default-buffer-prefix (ensime--get-name config))`. But I believe that there should be function for getting current project interactive buffer name.

TheBB:
What's the story on this?

TheBB:
bump

TheBB:
I'm closing this, since I can't make out from the above discussion any sort of consensus on whether this is necessary or even correct. Please resubmit as needed.

PR #6532, opened 2016-07-08
Suppress company-complete call if tooltip is visible in scala-mode

iravid:
Currently, the function `scala/completing-dot` causes `company-mode` to be called whenever `.`
is inserted, even if the company tooltip is visible.

This causes the currently displayed candidate to be inserted and doesn't seem to be the required
behavior.

This change suppresses the call to `company-mode` if the tooltip is visible.

### DIFF
--- a/layers/+lang/scala/funcs.el
+++ b/layers/+lang/scala/funcs.el
@@ -90,7 +90,8 @@ point to the position of the join."
                     (buffer-substring (line-beginning-position) (point)))
     (delete-horizontal-space t))
   (insert ".")
-  (company-complete))
+  (when (not (company-tooltip-visible-p))
+    company-complete))
 
 ;;; Flyspell
 


gilbertw1:
Hi, I just found this pull request after submitting one of my own to solve this exact problem: https://github.com/syl20bnr/spacemacs/pull/6545

The primary difference being that I solved the problem by calling ```company-abort``` before insert the ```.``` and calling ```company-complete```. 

I didn't mean to duplicate effort, I only noticed this one when I was looking to get an idea of how many outstanding PRs were out there. I was wondering if you had any thoughts on which method is superior, the one thing I like about the ```company-abort``` approach is that it inserts the dot and still shows possible completions regardless of whether there was a company candidate visible or not, which I believe is the overall idea with this ```completing-dot``` function.

iravid:
@gilbertw1 - hey! I think that your method is better, so closing this one out :-)

PR #6544, opened 2016-07-10
Replace py-yapf with yapfify.

JorisE:
Yapfify uses project the settings that would apply to a file if you would run YAPF on it from the command line.

It also shows a help-window if yapf fails to farmat the buffer with an error.

### DIFF
--- a/layers/+lang/python/packages.el
+++ b/layers/+lang/python/packages.el
@@ -30,11 +30,11 @@
     pytest
     (python :location built-in)
     pyvenv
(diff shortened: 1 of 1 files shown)


TheBB:

Made some changes in https://github.com/syl20bnr/spacemacs/commit/3c18a4a6bb2be3ca5c4fcece32b44ced04e6738a, in particular the `SPC m y` binding feels out of place (and was undocumented). We can add it back if we find a good key for it.

Issue #6547, opened 2016-07-10
Renaming from lowercase to uppercase doesn't work.


vimuel:
#### Description
`rename-current-buffer-file` doesn't work when only changing case. Is this intended behaviour?

#### Reproduction guide
- Start Spacemacs
- Open `somefile.txt`
- Using [SPC f R] rename to `SomeFile.txt`

*Observed behaviour:*
Filename is still `somefile.txt`.

*Expected behaviour:*
Filename should get changed to `SomeFile.txt`.

#### System Info
- OS: darwin
- Emacs: 25.1.50.1
- Spacemacs: 0.105.21
- Spacemacs branch: master (rev. 0283f64)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(auto-completion emacs-lisp git haskell markdown org pandoc
                 (spell-checking :variables spell-checking-enable-by-default nil)
                 syntax-checking version-control)
```

cpaulik:
This works for me. Have you made sure that `somefile.py` is selected when pressing `RET` and not `SomeFile.py`. Helm will automatically select an existing file if it matches the search string.

vimuel:
Yes, `SomeFile.txt` was selected. See screenshots below:
![Screenshot 1](https://raw.githubusercontent.com/vimuel/screenshots/master/spacemacs/20160701_spacemacs_1.png)
![Screenshot 2](https://raw.githubusercontent.com/vimuel/screenshots/master/spacemacs/20160701_spacemacs_2.png)
![Screenshot 3](https://raw.githubusercontent.com/vimuel/screenshots/master/spacemacs/20160701_spacemacs_3.png)
![Screenshot 4](https://raw.githubusercontent.com/vimuel/screenshots/master/spacemacs/20160701_spacemacs_4.png)

TheBB:
This is an OS X issue. See https://github.com/syl20bnr/spacemacs/issues/6547

vimuel:
My bad.

nixmaniack:
Another point to note here is that if you use `ivy`, it lets you rename correctly(looks like it passes correct `dir` value to `read-file-name`). But with `helm` the issue persists and as explained in the referenced issue, if you pass correct value of `dir` to `read-file-name` it allows you to rename on OS X as well. We can patch that function for `'darwin`.

vimuel:
[...]

PR #6558, opened 2016-07-12
fix rename-current-buffer-file for OS X

### LINKED ISSUE #4383: File rename broken for changing capitalization
I'm running 0.105.0 but this was an issue on earlier versions as well. Renaming `./Foo.md` to `./FOO.md` using `SPC f R` fails (I receive a prompt telling me I'm attempting to rename Foo.md to Foo.md). I suspect that this is due to the autocomplete changing FOO back to Foo without showing so to the user. I didn't see any way to disable this behavior to test this.

nixmaniack:
`read-file-name` expects directory and not filename.
Fix #4383 #6547


d12frosted:

I am glad that solution doesn't have any OS-specific code.

Issue #6523, opened 2016-07-07
Error during evil-unimpaired loading (can't find f package)


bixuanzju:
#### Workaround for this problem

Add `evil-unimpaired` to list of excluded packages in your `.spacemacs`.

```
(defun dotspacemacs/layers ()
  "Configuration Layers declaration."
  (setq-default
   ;; ...
   dotspacemacs-excluded-packages '(evil-unimpaired)
   ;; ...
))
```

#### Description :octocat:
Commit c98f862 assumes we installed `f`, but in my case, I didn't.

Note that I have to revert c98f862 (in my system, it's 2a04d47) to be able to use the reporting mechanism.

#### Reproduction guide :beetle:
- Start Emacs
- and behold

*Observed behaviour:* :eyes: :broken_heart:
(file-error "Cannot open load file" "No such file or directory" "f")

*Expected behaviour:* :heart: :smile:
No such error

#### System Info :computer:
- OS: darwin
- Emacs: 25.0.95.1
- Spacemacs: 0.105.21
- Spacemacs branch: bug (rev. 2a04d47)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
(ivy
 (auto-completion :variables auto-completion-enable-sort-by-usage t)
 emacs-lisp
 (git :variables git-magit-status-fullscreen t)
 (markdown :variables markdown-live-preview-engine 'vmd)
 org
 (shell :variables shell-default-height 30 shell-default-position 'bottom)
 spell-checking
 (syntax-checking :variables syntax-checking-enable-tooltips nil)
 (version-control :variables version-control-diff-tool 'diff-hl)
 osx colors idris
 (haskell :variables haskell-completion-backend 'intero haskell-enable-hindent-style "johan-tibell")
 (latex :variables latex-build-command "LatexMk")
 (evil-snipe :variables evil-snipe-enable-alternate-f-and-t-behaviors t)
 ocaml lua sml yaml
 (c-c++ :variables c-c++-enable-clang-support t)
 proof-general)
```

#### Backtrace :paw_prints:
```
Debugger entered--Lisp error: (file-error "Cannot open load file" "No such file or directory" "f")
  require(f)
[...]

PR #6559, opened 2016-07-12
ugly workaround for #6523

d12frosted:
Actually this is to show how `location local` should be translated in `core-configuration-layer.el`. I tried to implement this, but it looks like it requires more work than I expected 😸 If someone wants to implement it faster, this PR can be used as reference. 

I don't expect it to be merged 😸 

CC @bmag @syl20bnr @TheBB


d12frosted:
Actually I've started to work on installing local packages using quelpa [here](https://github.com/d12frosted/spacemacs/commit/664946f1e6403f7fc7e77397707339c9bae7bfa3). It's not ready yet and has following issues

- [ ] for some reason `nyan-mode` can't be installed with error about missing `version` or `package-version` value. I tried doing very different things (even totally changing content of `nyan-mode.el` with something that obviously works). The problem might be with my local cache (though I am sure I cleared everything related).
- [ ] for some reason on subsequent runs packages are installed over and over (and only some of them).

d12frosted:
There is no point in this PR anymore.

PR #6561, opened 2016-07-12
Change multi buffer search to use swoop

quicknir:
Previously, helm-ag searches were used to search in multiple buffers, while swoop was used for single buffer searches. Using swoop in all buffers had a special shortcut that was a bit awkward (s C-s). Swoop has a nicer interface then helm-ag, so this commit makes it so that swoop is used for multiple buffers, consistent with single buffer search, and eliminates the need for special shortcuts.

Change-Id: I0e80ab9c571513f771859003cba9ceac8b8aa8e0


quicknir:
I think swoop is the logical choice for searching all open buffers. It has follow mode on by default which is generally a win for buffers that are already open, but is slow and ends up opening many redundant buffers if used when searching on disk; in that sense I think that multi buffer search should be consistent with single buffer search and not with on-disk searches (file & project). It also has nicer formatted output than the various helm-ags (nice grouping of hits by file with clear fonts). 

From talking to people in gitter, it seems like there's a lot of love for swoop and a few people thought this would be a good idea. Another win (in my book) is that it actually reduces the number of shortcuts; I think as spacemacs gets larger and more complex, opportunities to simplify and reduce the number of shortcuts we use without loss of functionality should be taken.

Another thing that might be nice is to be able to transition from symbol highlight mode directly into single file swoop. I think I recall someone asking about this before in gitter. However, right now when I add the text it looks all misaligned so I decided to leave it for a separate commit. Maybe someone else could add that or tell me how to do it?

quicknir:
Another change that might be nice actually, is to make swoop all pop up consistently with how helm-ag pops up. Helm-ag pops up in the normal global helm way, swoop seems to "associate" itself with the current buffer and only pop up there. So if you have vertical splits, swoop is only half width. For single buffer swoop this perhaps makes sense as you are only searching in that buffer, but not so much for multi buffer. I'll look into this as well.

nixmaniack:
> Should I add it to the commit?

That sounds right to me. Add it to this PR itself.

quicknir:
Update on this. I noticed that it's slightly slower than helm-ag. So I tested it with 50 open buffers. Although it's not unbearable, it's definitely noticeable and I'm not sure if this is the right thing any more. Right now I'm chasing with both helm-ag and swoop; I'm trying to see how to make the former follow by default and have nicer output, and how to make the latter perform better. Hope to hear back from one of these guys soon and then decide appropriately.

PR #6562, opened 2016-07-12
Allow to install local package using quelpa

d12frosted:
This PR adds ability to mark local package to be installed using `quelpa`.

Following #6523 (initial issue), #6559 (proposed self-claimed ugly workaround) and discussion around https://github.com/d12frosted/spacemacs/commit/664946f1e6403f7fc7e77397707339c9bae7bfa3#commitcomment-18211735. 

This might be not the cleanest way to achieve the goal, but still lets us to install local packages differently.

```
;; this will be installed by `load-path' modification
(package-name :location local)

;; this will be installed using quelpa
(package-name :location (recipe :fetcher local))
```

Any tips and ideas are welcome!

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -477,7 +477,7 @@ layer directory."
   "Make `cfgl-layer' objects from the passed layer SYMBOLS."
   (delq nil (mapcar 'configuration-layer/make-layer symbols)))
 
-(defun configuration-layer/make-package (pkg &optional obj togglep)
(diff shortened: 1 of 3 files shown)


syl20bnr:
I love it :purple_heart:
One little detail though :-)
I would use `:fetcher` instead of `:url`:

```
(package-name :location (recipe :fetcher local))
```

d12frosted:
@syl20bnr nice idea! Makes so much more sense to use `:fetcher local` rather than `:url local`. 💯

d12frosted:
Cherry-picked right into `develop` branch.

Issue #5776, opened 2016-04-13
Spacemacs-layouts - Special buffers are not added to the current layout


syl20bnr:
### Repro steps

_Need python layer enabled and installed_
- Start Emacs
- `SPC l 2 y foo RET`
- `SPC f J py RET`
- `SPC m ' RET y`
- `ESC SPC b b`
### Observed behavior

Only buffers `*scratch*` and `xxxx.py` are listed.
### Expected behavior

The REPL buffer should be listed as well.
### Work around

`SPC b a` to add the REPL manually to the layout.

cpaulik:
Could this also the reason why e.g. the `*pytest*` buffer can show up in any layout if it goes into debug mode?

syl20bnr:
I don't know if it shares the same cause but we can add this unexpected behaviour in this issue.

d12frosted:
AFAIU, it's already in master, no? 😸

d12frosted:
Closing this one for now. Let me know if it's not perfect yet and I'll reopen (or reopen yourself) 😸

cpaulik:
This is probably not fixed. Any new `*python*` or `*pytest*` buffer is not added to the current perspective. I'm on latest develop fea919c

braham-snyder:
Not looking fixed for me on develop either.

I think we can easily improve the default behavior quite a bit. Currently, for example, in any non-default layout, after pressing `<return>` on a magit diff to visit the associated file, `previous-buffer` does not return to that magit diff (even after adding `"\\*magit"` to the useful buffers regexp list).

Instead, at the moment, I've just set `persp-add-buffer-on-after-change-major-mode` to `t`--that's worked well for me so far, but unsure if there's a better solution.

For reference, here's what looks like a similar but slightly less general issue (closed "for inactivity") with some discussion that's probably applicable: https://github.com/syl20bnr/spacemacs/issues/4956

Perhaps one of these two issues should be reopened?

ngoonee:
Agree that https://github.com/syl20bnr/spacemacs/issues/4956 or this should be re-opened. My current bug-bear is that opening terms (SPC-') when in a layout doesn't add the term to the layout, which is basically what's being discussed in this linked issue (and would be solved with a proper solution to this issue).

d12frosted:
@bmag do you have any ideas?

bmag:
[...]

Issue #4956, opened 2016-02-04
spacemac-layouts: In non default layouts popping buffers is behaving wrong 


shivamkalra:
#### Description
I'm using `spacemac-layouts` and in all the non `Default` layout, behavior of popping buffers is wrong. 

#### Reproduction guide
- Start Emacs
- Create new layout by pressing `SPC-l 2`
- In new layout, open any file (make sure it is added to current layout `SPC-l b`)
- Now open `dired`, navigate to some sub directories
- From a sub-directory press `q`
- You would expect to go back to parent directory but you get to file buffer instead.

*Observed behaviour:*
Since `dired` is not part of current layout, on quitting the buffer, it shows me the last buffer on `spacemacs-layouts` buffer stack instead of global buffer stack.

This is making `dired` almost impossible to use, I can only navigate parent directory by pressing `^`. Same happens for various Emacs plugins that create buffers with `*` in front, they are not added to layouts and their history is not tracked.

*Expected behaviour:*
On existing the current buffer, it must go back to previous buffer I was on.

It must add all the buffer opened in current layout no matter whether it is `dired`, `mu4e` or `*` in front of it. However, on saving them to `file` it can write exclusively.

#### System Info
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.9
- Spacemacs branch: nil (rev. a739e1f)
- Distribution: spacemacs
- Editing style: hybrid
- Completion: helm
- Layers:
```elisp
((auto-completion :variables auto-completion-enable-help-tooltip t)
 better-defaults spacemacs-helm emacs-lisp gtags c-c   version-control git markdown org gnus shell syntax-checking csharp sql csharp lua elpy elfeed xkcd haskell extra-langs spell-checking ipython-notebook
 (erc :variables erc-sasl-server-regexp-list
      '("irc\.freenode\.net"))
 shell-scripts theming javascript html colors ibuffer smex github themes-megapack fasd wakatime notmuch mu4e tools
 (latex :variables latex-enable-auto-fill t))
```
Annoying Behavior
![optimised](https://cloud.githubusercontent.com/assets/1386352/12807825/5e8f4698-cae0-11e5-83d3-397537c165b6.gif)

robbyoconnor:
Are you using `develop` or `master`?

shivamkalra:
`develop`

Andre0991:
[...]

PR #6574, opened 2016-07-14
Set buffer-predicate to spacemacs/useful-buffer-p

### LINKED ISSUE #6159: Never auto-switch to helm menu buffers
#### Description :octocat:

If there's only one normal buffer left, `spacemacs/alternate-buffer` will switch to a random helm menu (if it exists). After deleting all buffers, spacemacs will switch to an old helm menu.

Preferably, these would never be displayed. I don't know how fixable this is but it's kind of an annoying papercut.

- Start Emacs
- Hit `SPC SPC` then `ESC` to open and close a helm menu.
- repeatedly hit `SPC bd` to close all open buffers.
- Eventually, you'll switch to some helm menu.

*Observed behaviour:* :eyes:

I switch to old helm menus.

*Expected behaviour:* :broken_heart:

When deleting the last normal buffer, I expect to see a new `*scratch*` buffer. When calling `spacemacs/alternate-buffer` on the last buffer, I expect nothing to happen.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.20
- Spacemacs branch: develop (rev. 1623ff6)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(helm better-defaults git
      (auto-completion :variables auto-completion-return-key-behavior nil)
      org spell-checking syntax-checking
      emoji emacs-lisp markdown latex bibtex csv yaml c-c   html haskell go javascript python systemd
      (rust :variables rust-format-on-save nil))
```

bmag:
Prevents `next-buffer`, `other-buffer`, etc. from choosing useless buffers.
No need for `spacemacs/next-useful-buffer`, `spacemacs/previous-useful-buffer` anymore, so I replaced their bindings with `next-buffer` and `previous-buffer`.

Fixes `spacemacs/alternate-buffer` to respect `buffer-predicate`. It no longer uses `evil-alternate-buffer`.

When spacemacs-layouts is used, buffer-predicate filters useful buffer that belong to the current layout. For example, `SPC b n` will not show useless buffers or buffers that don't belong to the current layout.

Caveats:
- `spacemacs/alternate-buffer` shows buffer from current layout, so it is effected by `persp-mode` not adding certain buffers to current layout. See #4956, #5776.
- Switching layouts sometimes doesn't immediately update the buffer-predicate persp-mode uses. Instead, it's updated ~2-3 seconds after a layout switch. As a result, switching a layout and then pressing `SPC TAB` can show a buffer from the previous layout.

Fixes #6159 as much as is reasonably possible.


TheBB:

izahn:
This breaks stuff, see https://github.com/syl20bnr/spacemacs/pull/6579. Can we revert until we figure out why and how to fix it?

TheBB:
You mean https://github.com/syl20bnr/spacemacs/issues/6678

izahn:
Yes https://github.com/syl20bnr/spacemacs/issues/6678

Issue #4165, opened 2015-12-12
Pinning specific packages to alternative repository?


a13ph:
Make it easier to set packages which are to be installed from melpa-stable (or org elpa, or smth else even) - there is no information on where to put those, while attempts to do so by usual ways fail

As @grammati [said in gitter](https://gitter.im/syl20bnr/spacemacs?at=566c9283187e75ea0e485cbf)

> user-init is too early (package-archives not defined), and 
> user-config is too late (already got an error about melpa-stable not being an archive)

Of course one could hijack spacemacs files to inject those settings at appropriate part of loading process, but we better do so collectively :)

a13ph:
#3034 might be relevant, but I can't wrap my head around whether it is

a13ph:
[Old solution](https://github.com/syl20bnr/spacemacs/issues/2057#issuecomment-137960576) doesn't seem to work anymore

> package-archives is undefined in dotspacemacs/init

a13ph:
I actually think this should be reported upstream to be handled by Paradox itself, since it [will have](http://endlessparentheses.com/new-in-emacs-25-1-archive-priorities-and-downgrading-packages.html) or already has archive priorities and downgrading - very close features

gfZeng:
I found most Clojure developer need this feature. We need use stable cider

tonylotts:
See the temporary fix for #6531 as an example of how to pin CIDER.

Issue #3542, opened 2015-10-25
How to use MELPA Stable  as default repository


robsonpeixoto:
I'd like to use stable version, but I'd like to use unstable version if the package aren't available on stable.

Is possible to configure MELPA Stable as default repository?

Thanks

robsonpeixoto:
Maybe it can help https://github.com/jwiegley/use-package#for-packageel-users

bmag:
In Emacs 25 you can do so by setting `package-archive-priorities`. Right now Spacemacs doesn't use Melpa Stable, so you will also need to add Melpa Stable to `configuration-layer--elpa-archives`.

Issue #2057, opened 2015-06-20
Pinned package dependencies


bonega:
As far as I know Spacemacs always install bleeding edge packages from ELPA.

This leaves us with the fun situation where Spacemacs stability is a function of todays date.
E.g Spacemacs 0.102.2 will probably become more and more broken as packages change.

This is an unwanted behaviour for a supposed stable version(or any version).

Dependencies should be pinned to specific versions.
Also it is debatable if the bulk of packages should be installed from bleeding edge ELPA.
Rather something like melpa-stable(or elpa-stable if it exists).

The Clojure-repl integration Cider changes daily I think.
Since Elpa uses 0.9.1-SNAPSHOT.
A fair target would be 0.9.0 that was released as stable a few days ago.

I would think that a great deal of issues is dependent on the specific version of a package.
That makes it very hard to debug user problems in a safe way.

dab:
:+1: +1
packages versions should be hardcoded in releases

codygman:
(thumbsup) Every once in a while I switch to spacemacs for a work week.
This would definitely help make that smoother.

sooheon:
This is coming to emacs itself in 25: http://endlessparentheses.com/new-in-emacs-25-1-archive-priorities-and-downgrading-packages.html

syl20bnr:
I hear you.
We are not ready for such a move and the best is to stick with MELPA for everything for now. But rest assure that stable packages are planed:
1- merge extensions and packages into a single keyworded list
2- add support for Quelpa
3- from here we will be able to fetch the packages from source at a given commit which will give us full flexibility to tailor the spacemacs distribution.
4- I will be able to tag spacemacs 1.0 :-)

I will be able to work on the first step for 0.104.x

For the CIDER example, we have to move forward and use 0.9.1-SNAPSHOT like we used 0.9.0-SNAPSHOT before.

bmag:
Hey, I just came across some short instructions on how to build a [custom melpa archive](https://github.com/milkypostman/melpa/wiki/Custom-Melpa-Archive). I'm not exactly sure what it entails, but I suspect that we could skip this step:
```bash
git rm recipes/*
```
Before a release we simply re-build all recipes with:
```bash
./melpa build
```
And use our own custom melpa archive (hosted on spacemacs.org?) instead of the bleeding-edge melpa.org archive.

bmag:
[...]

PR #6575, opened 2016-07-14
Allow to pin packages to specific archive

### LINKED ISSUE #6531: Scala layer: ensime needs to be pinned to stable.melpa.org
Ensime reached a milestone 1.0.0 release; and now releases are being tracked on stable.melpa.org, whereas melpa.org tracks the development versions which are quite volatile.

Currently, we don't have an entry in 'package-archives for stable.melpa.org.
Locally, I've tried adding a melpa-stable entry, and pinning ensime in (use-package), however it still downloads from melpa.org (not stable.melpa.org).  It's possible that I didn't use ':pin' correctly, or there could be a bug related to ':pin' (possibly upstream).

d12frosted:
Somewhat fixes or moves towards fixing #6531, #2057 and #3542.

Also #4165 is somewhat related.

Basically thanks to @tonylotts I realised that pinning packages to stable MELPA can be easily automated.

This leverages described problem. But doesn't solve it fully.
# Hot it works

This PR does following.
1. Adds MELPA stable to list of package archives. This has minor drawback - longer update. But this is very minor.
2. Exposes function `configuration-layer/pin-package` which can be used to pin specific package to one of available archives. One can use it manually, but there is a better choice.
3. Adds custom handling for packages with `:archive archive-name`. To be accurate - it just pins that package.
# Usage

``` elisp
(defun dotspacemacs/layers ()
  "Configuration Layers declaration."
  (setq-default
   ;; ...
   dotspacemacs-additional-packages '(;; ...
                                      (ensime :archive melpa-stable))
   ;; ...
   )
```

This will force Spacemacs to download `ensime` from MELPA stable.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -210,7 +210,8 @@ is not set for the given SLOT."
 (defvar configuration-layer--elpa-archives
   '(("melpa" . "melpa.org/packages/")
     ("org"   . "orgmode.org/elpa/")
-    ("gnu"   . "elpa.gnu.org/packages/"))
+    ("gnu"   . "elpa.gnu.org/packages/")
+    ("melpa-stable" . "stable.melpa.org/packages/"))
   "List of ELPA archives required by Spacemacs.")
 
 (defvar configuration-layer-exclude-all-layers nil
@@ -560,6 +561,7 @@ If TOGGLEP is nil then `:toggle' parameter is ignored."
          (pkg-name-str (symbol-name pkg-name))
          (layer (unless (eq 'dotfile layer-name)
                   (configuration-layer/get-layer layer-name)))
+         (archive (when (listp pkg) (plist-get (cdr pkg) :archive)))
          (min-version (when (listp pkg) (plist-get (cdr pkg) :min-version)))
          (step (when (listp pkg) (plist-get (cdr pkg) :step)))
          (toggle (when (listp pkg) (plist-get (cdr pkg) :toggle)))
@@ -584,6 +586,8 @@ If TOGGLEP is nil then `:toggle' parameter is ignored."
     (cfgl-package-set-property obj :excluded
                                (and (configuration-layer/layer-usedp layer-name)
                                     (or excluded (oref obj :excluded))))
+    (when archive
+      (configuration-layer/pin-package pkg-name archive))
     (when location
       (if (and (listp location)
                (eq (car location) 'recipe)
@@ -1965,8 +1969,8 @@ to select one."
              (cadr (assq 'built-in stats))))
     (with-current-buffer (get-buffer-create spacemacs-buffer-name)
       (let ((buffer-read-only nil))
-	(spacemacs-buffer//center-line)
-	(insert "\n")))))
+  (spacemacs-buffer//center-line)
+  (insert "\n")))))
 
 (defun configuration-layer/load-or-install-protected-package
     (pkg &optional log file-to-load)
@@ -2009,6 +2013,19 @@ FILE-TO-LOAD is an explicit file to load after the installation."
       (setq configuration-layer-error-count
             (1+ configuration-layer-error-count))
     (setq configuration-layer-error-count 1)))
+
+(defun configuration-layer/pin-package (package archive)
+  "Pin PACKAGE to ARCHIVE.
+
+When PACKAGE is pinned to specific ARCHIVE, it is installed
+exactly from that ARCHIVE.
+
+ARCHIVE must be one of the following symbols: melpa, org, gnu,
+melpa-stable."
+  (let ((archive-name (symbol-name archive)))
+    (unless (assoc archive-name package-archives)
+    (error "Archive '%s' is not supported." archive))
+  (push (cons package archive-name) package-pinned-packages)))
 
 (provide 'core-configuration-layer)
 
--- a/doc/LAYERS.org
+++ b/doc/LAYERS.org
@@ -328,6 +328,9 @@ as follows:
                              :fetcher github
                              :repo "some/repo"))
 
+    ;; A pinned package
+    (some-package :archive melpa-stable)
+
     ;; An excluded package
     (some-package :excluded t)
     ))
@@ -337,6 +340,8 @@ The =:location= attribute specifies where the package may be found. Spacemacs
 currently supports packages on ELPA compliant repositories, local packages and
 MELPA recipes (through the Quelpa package). See the [[https://github.com/milkypostman/melpa#recipe-format][MELPA documentation]] for more
 information about recipes.
+
+By using =:archive= attribute one can pin specific package to specific archive.
 
 Packages may be /excluded/ by setting the =:excluded= property to true. This
 will prevent the package from being installed even if it is used by another


bmag:
Using a location recipe that isn't really a recipe kinda violates POLA. It may be better to implement a new syntax:
```elisp
(ensime :archive "melpa-stable")
```
Or maybe:
```elisp
(ensime :pin "melpa-stable")
```
In this case, `configuration-layer/pin-stable-package` should be renamed and take an additional `archive` parameter. The function could issue a warning if `archive` isn't listed in `package-archives`, mostly to protect from typos in the package declaration.

d12frosted:
Hey @bmag, 

What you say makes sense. I will update my PR to reflect this. Thanks for the suggestion.

d12frosted:
I've updated PR.

Instead of using strings - let's use symbols 😸 Checkout updated PR description.

core/core-configuration-layer.el
```diff
@@ -1866,7 +1870,18 @@ FILE-TO-LOAD is an explicit file to load after the installation."
...
 
+(defun configuration-layer/pin-package (package archive)
+  "Pin PACKAGE to ARCHIVE.
+
+When PACKAGE is pinned to specific ARCHIVE, it is installed
+exactly from that ARCHIVE.
+
+ARCHIVE must be one of the following symbols: melpa, org, gnu,
+melpa-stable."
+  (unless (memq archive '(melpa org gnu melpa-stable))
```
bmag: Should use `package-archives` instead of hard-coded archive names:
```elisp
;; if archive is a string
(assoc archive package-archives)
;; if archive is a symbol
(assoc (symbol-name archive) package-archives)
```

core/core-configuration-layer.el
```diff
@@ -1866,7 +1870,18 @@ FILE-TO-LOAD is an explicit file to load after the installation."
...
+    (error "Archive '%s' is not supported." archive))
+  (push `(,package . ,(symbol-name archive)) package-pinned-packages))
```
bmag: Nitpick: could use `cons` instead of backtick syntax

bmag:
I don't understand why you prefer symbols to strings here, but either way works :smile:

core/core-configuration-layer.el
d12frosted: Good catch. Thanks!

core/core-configuration-layer.el
d12frosted: Oh. Didn't know about this one. Thanks!

d12frosted:
@bmag thanks for all suggestions once again 😸 💃 

> I don't understand why you prefer symbols to strings here, but either way works 😄

Yeah, both of them works. My main point is that symbol here is like enum. Only several predefined values are acceptable. On the other hand, string has another meaning 😸

diego-alvarez-hs:
ping, would be awesome to have this on the develop branch

d12frosted:
Fixed conflicts.

d12frosted:
Fixed conflicts 😸

robsonpeixoto:
A noob question. Can it be used on `packages.el` http://spacemacs.org/doc/LAYERS.html#packagesel

d12frosted:
Yes

diego-alvarez-hs:
ship it! :D

liujoey:
Ship it please! 😸

babilen:
This sounds as if it is not possible to get all packages from stable by default, but would have to enumerate all of them. Would be great if you could allow users to define both the default repository along with per-package overrides.

Thanks! :)

clhodapp:
@babilen That sounds really likely to break things, as it's likely that the code within _some_ layer that most users have enabled on their systems would break if you changed the default repo to pull packages from. It seems a lot more manageable for the layer itself to specify which package version (from which repo) it's intended to work against.

babilen:
@clhodapp Well, sure. It's just that at the moment all packages (appear?) to be daily snapshots from melpa. I don't quite see how that is considered to be less likely to break things than pulling properly released versions by default. :)

clhodapp:
@babilen The problems is gradual vs all-at-once. Sure, _some_ layers break _some_ time, but by moving between melpa and melpa stable, you are likely changing the version of a whole bunch of packages at the same time (and changing them away from the version that is potentially a lot different from the version that most users are on to boot). It seems like a much safer option would be to have separate "stable" and "unstable" versions of layers (potentially both coexisting in the tree at the same time) and letting you pick which set you wanted to use (at the potential cost of losing out on some layers, especially at first). It could _maybe_ work to be able to mix-and-match from the sets, but then you could run the risk pre-inits and post-inits being broken (unless the system is crazy-smart and knows to use the pre-inits and post-inits from stable when a package is coming from a stable layer)... anyway, sorry for the rant ;)

d12frosted:
@babilen AFAIK, MELPA builds all packages continuously. Once build is finished, new one is started. If a package doesn't have any changes between builds, there is no new version. MELPA builds from specified branch (with `master` being the default). Stable MELPA build from specified branch (with `master` being the default), but based on tags, not on commits. As practice shows many developers forget to tag 'stable' commits. Some packages have very slow 'release' policy. So even if a package has stable version (which is not true for most packages on MELPA) it might be long away from latest commit, which might bring so many troubles to Spacemacs users. Also keep in mind that Spacemacs is not officially released and it's still in heavy development. But yeah, having support of MELPA stable is a good thing. Not sure when it will come though.

babilen:
@d12frosted That's exactly how it works. With support for `package-archive-priorities` in Emacs 25 the ecosystem will, hopefully, provide proper releases more frequently. I would, personally, much prefer to track properly tested and coherent releases than following the whims of `HEAD`.

`package-archive-priorities` also doesn't exclude installations from other repositories, but simply allows you to assign a higher priority to some repositories rather than relying solely on version number comparisons.

I understand that this requires a more rigorous development process than many package developers are willing to adopt, but exposing `package-archive-priorities` **and** allowing users to specify exceptions as proposed in this PR would be useful. That way you could treat packages that do not provide proper releases as exceptions rather than the norm. And if this causes too many problems users can still opt to not tinker with this setting.

d12frosted:
@babilen Unfortunately Spacemacs is not really suited for 'stable' releases, so `package-archive-priorities` will not solve problems at the moment. Because Spacemacs codebase targets latest available versions. If that version has any backward-incompatible changes - using older version breaks your setup. So you have to use new version.

We might provide something like 'snapshots'. The idea is that specific Spacemacs version is known to work with specific set of packages. There is a [mirror](https://github.com/syl20bnr/spacemacs-elpa-mirror) of popular package archives we maintain. Ideally I would like it to be used for these snapshots.

babilen:
@d12frosted But isn't that more a result of the fact that melpa stable is a comparably new thing, rather than a conscious and intentional decision?

In a way you probably want to develop spacemacs against a defined set of dependencies, rather than targeting `HEAD`. The way it is done right now would also mean that you cannot use the current spacemacs version in one, two or five years, whereas a development process that targets (and potentially pins) stable package releases wouldn't "break" over time.

d12frosted:
@babilen 

> But isn't that more a result of the fact that melpa stable is a comparably new thing, rather than a conscious and intentional decision?

It's hard to answer this question, but supporting Stable MELPA is not high on the priority list. There are several possible reasons for this.

1. Spacemacs is still not near it's release state. It just allows us to defer this issue 😸 
2. There are several ways to provide real stable Spacemacs releases (using Stable MELPA, pinning packages to specific version, using package snapshots à la stack lts packages). 
3. And most importantly - no one is working on this. I am planning to start, but I am busy with other things, so it might take some time until I begin doing anything 😸 

But keep in mind that I understand the need of such thing. It's just not critical for me 😸 I update not that often and I know how to revert / rollback updates (thanks to Spacemacs it's easy to do).

Current implementation of stable Spacemacs releases is not ideal as it targets specific versions of packages, but these versions are not written anywhere. Our solution to this issue - package fix releases. So in case you wish to use 'stable' release you have to use latest stable release with latest packages. If something breaks - new release will come. In case you are using old version of Spacemacs - you are on your own 😸 

P. S. Just in case. Everything I say is just my opinion and not something that can be treated as official project position. Other contributors might disagree with me. 😸

babilen:
@d12frosted Thank you for elaborating. I totally understand the reason why this is tricky to implement at the moment.  

I also believe that functionality that is easy to use and available will be used, so by making it easier to install, configure and track stable versions the community will, inadvertently, start using it. This could result in a more stable working platform that does not break over time. Rather than having to catch up with every whim (and bug) in `HEAD`, spacemacs development could focus on targeting certain major package releases.

Thanks for all the work on spacemacs and comments on this issue. Much appreciated!

syl20bnr:
I was working on this in late december but the solution was not satisfying, this is a tricky topic. The idea was to have package distribution channels:
- frozen (taken from a snapshot)
- stable
- latest

This is great but switching between channels may break Spacemacs since its configuration is bound to specific versions of packages.
Having priorities in Emacs 25 helps but I don't think we can leverage it given the limitation above.

So I start to think that it reduces only to a matter of choosing what we support in Spacemacs: either last stable packages or latest packages but not a mix of them. 

There is **no** good choice to this, since some people want stability and others want latest.

I prefer latest with a fast living ecosystem but there is a major downside to this:  sometimes a latest version of a package breaks fresh installs of Spacemacs. Can we find some solutions to this without pinning packages ?

I'm thinking about it, a sort of error detection + fallback may be a solution but not sure how hard it would be to implement.

tonylotts:
Is there anything holding this back from being merged?

achikin:
Any updates on this? I'm running into issues with the latest cider.

d12frosted:
seems like it's not going to happen 😸 closing PR

PR #6576, opened 2016-07-14
Fix magit-gh-pulls docs since #g isn't defined

robbyoconnor:
The docs weren't correct. This corrects them.

### DIFF
--- a/layers/+source-control/github/README.org
+++ b/layers/+source-control/github/README.org
@@ -44,13 +44,14 @@ be added to your =~/.gitconfig=
 
 In a =magit status= buffer (~SPC g s~):
 
-| Key Binding | Description                                     |
-|-------------+-------------------------------------------------|
-| ~# c~       | create a pull request                           |
-| ~# g~       | get a list of all PRs in the current repository |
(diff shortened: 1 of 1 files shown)


d12frosted:
Ya, first time I tried using `gh-pulls` run into the same problem - `#g` is not defined.

But, shouldn't it be changed by `#r`?

robbyoconnor:
@d12frosted fixed it -- and added everything else :)

d12frosted:
Now that I think about it - there is fundamental problem with `magit-gh-pulls` key bindings. The package itself [uses](https://github.com/sigma/magit-gh-pulls/blob/master/magit-gh-pulls.el#L594) `# g`. It's reflected in it's documentation and in the `magit` buffer. 

The problem is that `evil-magit` [makes](https://github.com/justbur/evil-magit/blob/master/evil-magit.el#L376) `magit-gh-pulls` popup evil in it's nature. But doesn't update documentation. 

I'll fetch solution for changing helper message in `magit-gh-pulls` today so it can reflect real key bindings.

Anyway, thanks @robbyoconnor 👍

d12frosted:

Issue #6571, opened 2016-07-13
markdown mmm scala


wookietreiber:
#### Description

I am using the markdown layer and have a markdown file containing Scala code. I would like to have syntax highlighting for the Scala code blocks. I guess this can be done with **mmm**.

I checked the **mmm-mode** via <kbd>SPC</kbd> <kbd>h</kbd> <kbd>SPC</kbd> and Scala is missing from `mmm-add-classes`.

Also, I don't know and cannot find any documentation about how to enable **mmm** within the **markdown layer**.
#### Reproduction guide
- Start Emacs
- open a markdown file with Scala code blocks

_Observed behavior:_
- Scala code blocks are colored uniformly 

_Expected behavior:_
- Scala code blocks are colored with Scala-specific syntax highlighting
#### System Info
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.21
- Spacemacs branch: master (rev. 0283f64)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: helm
- Layers:

``` elisp
(auto-completion emacs-lisp git lua markdown scala)
```

TheBB:
> Also, I don't know and cannot find any documentation about how to enable mmm within the markdown layer.

It's enabled, you just need to add scala to the list.

https://github.com/syl20bnr/spacemacs/blob/develop/layers/+lang/markdown/packages.el#L159-L208

wookietreiber:
@TheBB Am I supposed to or not to edit this file as a user? I thought I was only supposed to edit the `~/.spacemacs` file.

crshd:
@wookietreiber No, you're not. But you can put `mmm-add-class` and `mmm-add-mode-ext-class` in your `~/.spacemacs`, too.

TheBB:
@wookietreiber I was suggesting you make a pull request with this contribution, but @crshd beat you to it.

wookietreiber:
@crshd To which section? `dotspacemacs/user-init`?

@TheBB Yeah, sure :-) I would contribute, too, but I'm not that well versed yet with spacemacs to feel comfortable enough.

TheBB:
You can't do it in user-init, mmm won't be available that soon. As a rule, the only thing you can do there is to set variables.

crshd:
Nope, but in `dotspacemacs/user-config`.

wookietreiber:
[...]

PR #6579, opened 2016-07-14
Generate mmm-modes automatically based on a list

crshd:
This "fixes"  #6571 

**Question:**
Should we really add all possible combinations of languages, or let the user (Spacemacsian) take care of that themselves?

If we keep adding definitions to the core, we'll end up with a terribly long list...

### DIFF
--- a/layers/+lang/markdown/packages.el
+++ b/layers/+lang/markdown/packages.el
@@ -206,6 +206,11 @@ Will work on both org-mode and any mode that accepts plain html."
                           :face mmm-declaration-submode-face
                           :front "^```rust[\n\r]+"
                           :back "^```$")))
+      (mmm-add-classes '((markdown-scala
+                          :submode scala-mode
+                          :face mmm-declaration-submode-face
+                          :front "^```scala[\n\r]+"
+                          :back "^```$")))
       (setq mmm-global-mode t)
       (mmm-add-mode-ext-class 'markdown-mode nil 'markdown-python)
       (mmm-add-mode-ext-class 'markdown-mode nil 'markdown-java)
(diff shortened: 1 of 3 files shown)


wookietreiber:
Shouldn't there be a more DRY why of supporting **all** languages?

crshd:
I don't think there's a fully automated way. But maybe something like [this](http://jblevins.org/log/mmm)?

wookietreiber:
Yes, this looks definitely less boilerplaty.

layers/+lang/markdown/packages.el
```diff
@@ -42,6 +42,18 @@
 (defun markdown/post-init-smartparens ()
   (add-hook 'markdown-mode-hook 'smartparens-mode))
 
+;; from Jason Blevins http://jblevins.org/log/mmm
+(defun markdown/mmm-auto-class (lang &optional submode)
```
crshd: This should be `spacemacs//markdown-mmm-auto-class`, shouldn't it?

wookietreiber:
I would also like a small section in `layers/+lang/markdown/README.org` detailing how and where to add these settings to the `.spacemacs` file.

crshd:
[There is](https://github.com/crshd/spacemacs/tree/develop/layers/%2Blang/markdown#automatic-mmm-mode-generation).
I guess I could add another section for the not-directly relating modes. But I'm not yet sure wether I want to change how those are handled.

wookietreiber:
My bad, was looking at master branch. LGTM.

TheBB:
That's not ugly!

syl20bnr:
Thank you ! 💜 
Indeed this is not ugly at all!
I made minor cleanups in commit: https://github.com/syl20bnr/spacemacs/commit/4fd895a9

PR #6585, opened 2016-07-14
Unify auto-mode-alist entries to end in ' rather than $.

agriffis:
See https://www.emacswiki.org/emacs/AutoModeAlist


agriffis:
ouch, will try again

agriffis:
Ignoring my goofed pull-request which merged develop to master, this was already fixed in develop anyway.

Issue #6052, opened 2016-05-14
Mode-line separator can't be set to nil


jmfurlott:
<PLEASE REPLACE ALL UPPERCASE EXPRESSIONS AND DELETE THIS LINE>
#### Description
Mode-line not properly able to be set to `nil`

#### Reproduction guide
- Add following to spacemacs config: `(setq powerline-default-separator 'nil)`
- Reload config

*Observed behaviour:*
After updating to spacemacs 0.105.20, the curved icons are back in the mode line when they should just be "vertical".

![image](https://cloud.githubusercontent.com/assets/805688/15269737/ae7ee6be-19bd-11e6-97c9-d9ec23a1e73e.png)

*Expected behaviour:*
No curves at all -- just a "vertical" modeline

#### System Info
- OS: darwin
- Emacs: 25.1.50.1
- Spacemacs: 0.105.20
- Spacemacs branch: master (rev. 9c7250a)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(git html javascript
     (shell :variables shell-default-term-shell "/bin/zsh")
     erc rcirc restclient markdown org deft ruby-on-rails)
```

TheBB:
I can't reproduce this on develop. You did remember to do call `(spaceline-compile)` after setting `powerline-default-separator` to `nil`, right?

jmfurlott:
@TheBB I just compiled it and it worked.  Closing this.  Thanks!

Did not realize I had to do that -- don't remember ever having to after using spacemacs for at least a year

TheBB:
You probably upgraded Spaceline. It's a requirement with the new version. Packages change sometimes. ;-)

PR #6588, opened 2016-07-15
Update command to change powerline separator

jkrmr:
I was having issues changing the default powerline separator -- `(setq powerline-default-separator ‘arrow)` was setting the value as expected, but I could not see any change in the powerline separators displayed. It turns out that invoking `(spaceline-compile)` is now required for the change to take visible effect.

ref: #6052,
https://github.com/TheBB/spaceline#compiling-a-mode-line

### DIFF
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -1171,7 +1171,8 @@ file:
 (defun dotspacemacs/user-config ()
   "This is were you can ultimately override default Spacemacs configuration.
 This function is called at the very end of Spacemacs initialization."
-  (setq powerline-default-separator 'arrow))
+  (setq powerline-default-separator 'arrow)
+  (spaceline-compile))
 #+END_SRC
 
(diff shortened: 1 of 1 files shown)


NJBS:
We added `spaceline-compile` to the `post-user-config-hook` on the `develop` branch so it should be unnecessary to manually call it.

See [here](https://github.com/syl20bnr/spacemacs/blob/develop/layers/+spacemacs/spacemacs-ui-visual/packages.el#L247) and https://github.com/syl20bnr/spacemacs/pull/6342

Unfortunately, this hasn't made it to `master` yet so manually calling it is necessary there.

jkrmr:
@NJBS Good to know, thanks! I'll keep that in mind for whenever it makes it to master.
I'd also suggest cherry-picking this onto the README for master, though, since the manual `(spaceline-compile)` call is still necessary there. The README in its current form is a bit misleading.

PR #6590, opened 2016-07-15
Fix possibly calling bad code after theme change

TheBB:
See https://github.com/syl20bnr/spacemacs/commit/653da2a6aae9c44648035dec986ef6f7e84ed3e6#commitcomment-18253098

CC @mattrrichard


JAremko:
I think we can use `spacemacs-post-theme-change-hook` to reload `org-indent`  to update indentation symbols face(they might became visible otherwise).  

Also `space-doc` should be reloaded.

TheBB:
Merged.

PR #6596, opened 2016-07-16
[scala] Improve documentation for scalastyle using flycheck

d1egoaz:
Now scalastyle warnings are shown up by flycheck
image

### DIFF
--- a/layers/+lang/scala/packages.el
+++ b/layers/+lang/scala/packages.el
@@ -12,6 +12,7 @@
 (setq scala-packages
   '(
     ensime
+    flycheck
     ggtags
     helm-gtags
     noflet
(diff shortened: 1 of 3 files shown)


layers/+lang/scala/packages.el
```diff
@@ -191,6 +192,9 @@
       (when (configuration-layer/package-usedp 'expand-region)
         (require 'ensime-expand-region nil 'noerror)))))
 
+(defun scala/post-init-flycheck ()
+  (add-hook 'scala-mode-hook 'flycheck-mode))
```
d1egoaz: I added this as I always have to manually call `flycheck-mode` and I have enabled the `syntax-checking` layer ¯\_(ツ)_/¯

TheBB: https://github.com/syl20bnr/spacemacs/blob/develop/layers/+lang/python/packages.el#L97-L98

This is how you need to do it.

d1egoaz: LOL I took as reference a random lang layer (php) and that's why I used that 😆 
I just updated the PR with your suggestion (also fixed php)

TheBB: Thanks!

TheBB:

Made some changes for line length in https://github.com/syl20bnr/spacemacs/commit/bef1295094d298f97cf2eb047e2c02887ee2727f

PR #6597, opened 2016-07-16
add open config action to layer spacemacs help

d12frosted:
This PR adds `c` action to open `config.el` file when using `SPC h l` or `SPC h SPC` with `ivy` completion<a href="#sup_1"><sup>1</sup></a>. The only controversial thing is that message is shown when `config.el` is missing instead of opening new file. I feel like it's for better. 

<sup id="sup_1">1</sup>: Gonna implement the same thing for `helm` if you like the idea of having quick access to `config.el`.

### DIFF
--- a/layers/+completion/helm/local/helm-spacemacs-help/helm-spacemacs-help.el
+++ b/layers/+completion/helm/local/helm-spacemacs-help/helm-spacemacs-help.el
@@ -195,6 +195,8 @@
                 . helm-spacemacs-help//layer-action-open-readme)
                ("Open packages.el"
                 . helm-spacemacs-help//layer-action-open-packages)
+               ("Open config.el"
+                . helm-spacemacs-help//layer-action-open-config)
                ("Install Layer"
                 . helm-spacemacs-help//layer-action-install-layer)
                ("Open README.org (for editing)"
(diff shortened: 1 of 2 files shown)


nixmaniack:
I like the idea of displaying the message for missing file instead of opening a new file, I think we should do it for other actions as well.

d12frosted:
@nixmaniack 👍 

Since I made the change inside of `ivy-spacemacs-help//layer-action-open-file` - the message is show for other actions as well.

Ginna translate this for helm as well tomorrow.

d12frosted:
Also forgot to mention that I wanted to document these new key bindings but didn't find old ones in documentation. Am I missing something?

d12frosted:
Ok, could not resist and ported to `helm` as well.

ralesi:
I was thinking about making a function that just opens a find-file prompt inside the layer directory,  then we could open any file there.

d12frosted:
I cherry-picked it into develop. 

@ralesi PR is welcome :) I think it would be nice to have generic action for opening any file in layer. Like 'local' `projectile-find-file` 😸 So one could use that even to open layer local package. 

Also, `helm-spacemacs-help` and `ivy-spacemacs-help` have so much code duplication. I believe it's possible to extract one `spacemacs-help` package that is interface-ignorant with `help-spacemacs-help` and `ivy-spacemacs-help` providing only interfaces. But this is separate issue from what you described 😸

PR #6607, opened 2016-07-17
Fix typo in VIMUSERS Customization section

ivanbrennan:
Pluralize "four top-level functions".

_before_:
> There are four top-level function in the file...

_after_:
> There are four top-level functions in the file...


TheBB:

PR #6609, opened 2016-07-17
Fix #6602, Use Quelpa file fetcher instead of url for local packages

### LINKED ISSUE #6602: Failed to install evil-unimpaired in latest develop branch
#### Description :octocat:
Failed to install evil-unimpaired in latest develop branch

--> installing package: evil-unimpaired@spacemacs-evil... [1/1]

- Start Emacs

Found an buffer named "*ftp anonymous@c*" with these messages:

>无效帮助命令 foo。
>open c
>未知主机 c。

It may be something like this in English:

>Invalid help command foo
>open c
>Unknown host c

Install evil-unimpaired successfully

#### System Info :computer:
- OS: windows-nt
- Emacs: 25.0.94.2
- Spacemacs: 0.105.21
- Spacemacs branch: develop (rev. c64d04f)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
(html javascript python markdown c-c   auto-completion go git better-defaults csharp ivy syntax-checking emacs-lisp)
```

NJBS:
An upstream issue with Emacs causes local file urls to incorrectly be interpreted as FTP addresses on Windows, see #6602.

cc @d12frosted

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -498,11 +498,12 @@ If TOGGLEP is nil then `:toggle' parameter is ignored."
                (eq (plist-get (cdr location) :fetcher) 'local))
       (setq location
             `(recipe
-              :fetcher url
-              :url ,(format
-                     "file://%s%s/%s.el"
-                     (configuration-layer/get-layer-local-dir (oref layer :name))
-                     name-str name-str))))
+              :fetcher file
+              :path ,(expand-file-name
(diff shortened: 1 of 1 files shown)


d12frosted:

Indeed, using `file` fetched is better than `url` fetched in this case. 💯 Even if it doesn't fix initial issue.

PR #6612, opened 2016-07-17
Fix spelling in saltstack layer

Johnstone-Tech:


layers/+tools/salt/README.org
```diff
@@ -12,7 +12,7 @@ This layer provides syntax highlighting for Saltstack files.
```
d12frosted: > This layer provides syntax highlighting for Saltstack files.

Hey, could you fix the naming here as well?

Johnstone-Tech: I think it may be best to keep it as saltstack here as the product is known as saltstack to many.

d12frosted: Oh. I thought this is the line you changed in previous PR. My bad.

d12frosted:

Issue #6619, opened 2016-07-18
New gtags integration shadows key for find-symbol in elisp


jbeigel:
#### Description :octocat:
In commit b22702bc9e ggtags and helm-gtags were added to many language layers.

Unfortunately for emacs-lisp, SPC m g g previously worked very well (I think
it was using  elisp-slime-nav), but is now shadowed by helm-gtags-dwim.

Of course the workaround would be to not use the gtags layer, but for other
languages it is fine.

#### Reproduction guide :beetle:
- Start Emacs
- Open some elisp file

*Observed behaviour:* :eyes: :broken_heart:
- Position cursor over some symbol and press SPC m g g => helm-gtags asks
  if it should create a gtags database.
- There is no keybinding for the old "find-symbol" method.

*Expected behaviour:* :heart: :smile:
- Being able to navigate elisp without using gtags

#### System Info :computer:
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.21
- Spacemacs branch: develop (rev. 9773ad6)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(markdown auto-completion
          (c-c   :variables c-c  -default-mode-for-headers 'c  -mode)
          emacs-lisp git gtags happymake haskell helm my-magit-gerrit
          (org :variables org-enable-reveal-js-support t)
          osx python
          (rtags :variables rtags-layer-enable-auto-completion t rtags-layer-enable-syntax-checking t)
          shell
          (syntax-checking :variables syntax-checking-enable-tooltips nil)
          version-control yaml)
```

#### Backtrace :paw_prints:
```
<<BACKTRACE IF RELEVANT>>
```

TheBB:
Can we make a list of languages with better `,gg` so we can remove that binding in those languages?

As a workaround for now you can use `(gtags :disabled-for emacs-lisp)`

TheBB:
Relevant: https://github.com/syl20bnr/spacemacs/pull/6620

deb0ch:
Additionally, ggtags conflicts with helm-gtags as it overrides the `M-.` in c++-mode.

Which brings another question: why do we even need them both at the same time ?

Would it not be more reasonable to have something like 
```
(setq c-c++-packages
  '(
    cc-mode
    disaster
    clang-format
    cmake-mode
    company
    (company-c-headers :toggle (configuration-layer/package-usedp 'company))
[...]

PR #6620, opened 2016-07-18
[WIP] Additional gtags stuff

TheBB:
- Don’t override SPC m g g in some modes
- Diminish minor modes
- Set python-mode bindings in :init
- No need to check if mode is bound

See https://github.com/syl20bnr/spacemacs/issues/6619

### DIFF
--- a/layers/+lang/asm/packages.el
+++ b/layers/+lang/asm/packages.el
@@ -65,7 +65,7 @@
   (spacemacs|add-company-hook nasm-mode))
 
 (defun asm/post-init-ggtags ()
-  (add-hook 'asm-mode-hook #'spacemacs/ggtags-mode-enable))
+  (add-hook 'asm-mode-hook #'spacemacs/ggtags-mode-enable 'append))
 
 (defun asm/post-init-helm-gtags ()
   (spacemacs/helm-gtags-define-keys-for-mode 'asm-mode))
--- a/layers/+lang/c-c++/packages.el
+++ b/layers/+lang/c-c++/packages.el
@@ -92,8 +92,8 @@
     (spacemacs/add-to-hooks 'c-c++/load-clang-args '(c-mode-hook c++-mode-hook))))
 
 (defun c-c++/post-init-ggtags ()
-  (add-hook 'c-mode-hook #'spacemacs/ggtags-mode-enable)
-  (add-hook 'c++-mode-hook #'spacemacs/ggtags-mode-enable))
+  (add-hook 'c-mode-hook #'spacemacs/ggtags-mode-enable 'append)
+  (add-hook 'c++-mode-hook #'spacemacs/ggtags-mode-enable 'append))
 
 (defun c-c++/init-gdb-mi ()
   (use-package gdb-mi
--- a/layers/+lang/clojure/packages.el
+++ b/layers/+lang/clojure/packages.el
@@ -248,7 +248,7 @@
   (spacemacs|add-company-hook cider-repl-mode))
 
 (defun clojure/post-init-ggtags ()
-  (add-hook 'clojure-mode-hook #'spacemacs/ggtags-mode-enable))
+  (add-hook 'clojure-mode-hook #'spacemacs/ggtags-mode-enable 'append))
 
 (defun clojure/post-init-helm-gtags ()
   (spacemacs/helm-gtags-define-keys-for-mode 'clojure-mode))
--- a/layers/+lang/common-lisp/packages.el
+++ b/layers/+lang/common-lisp/packages.el
@@ -28,7 +28,7 @@
     "sI" 'spacemacs/helm-slime))
 
 (defun common-lisp/post-init-ggtags ()
-  (add-hook 'common-lisp-mode-hook #'spacemacs/ggtags-mode-enable))
+  (add-hook 'common-lisp-mode-hook #'spacemacs/ggtags-mode-enable 'append))
 
 (defun common-lisp/post-init-helm-gtags ()
   (spacemacs/helm-gtags-define-keys-for-mode 'common-lisp-mode))
--- a/layers/+lang/csharp/packages.el
+++ b/layers/+lang/csharp/packages.el
@@ -93,7 +93,7 @@
   (add-hook 'csharp-mode-hook 'turn-on-evil-matchit-mode))
 
 (defun csharp/post-init-ggtags ()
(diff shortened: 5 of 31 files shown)


ksjogo:
The github layer is interfering with the gg bindings. 
At the moment I strangely have not gtags bindings on leader.
Additionally ggtags-mode-map can shadow the M-. binding if it is loaded after helm-gtags. Maybe we should undef them, when helm-gtags is enabled.

TheBB:
Why is the github layer interfering? The github bindings aren't mode-specific, all the gtags bindings are.

ksjogo:
Oh, missed the mode specificity. Even though, they won't show up in php-mode in my emacs, actually no binding at all behind M-m m in php-mode.

kRITZCREEK:
I'll be submitting a new `SPC m g g` keybinding for PureScript PR this week. Does this PR mean I need to change anything but the PureScript layer?

TheBB:
@kRITZCREEK We don't have gtags in purescript anyway.

@ksjogo They certainly should be there if you have enabled the php layer.

ksjogo:
Yes, I would assume so, but there is none. Will investigate.

nixmaniack:
> Yes, I would assume so, but there is none. Will investigate.
 
I have found this occasionally that I can access mode specific bindings with `,` or `SPC m` (normal state) but they're not available under `M-m m`(hybrid state). Not sure if at all this is related.

deb0ch:
Additionally, ggtags conflicts with helm-gtags as it overrides the `M-.` in c++-mode.

Which brings another question: why do we even need them both at the same time ?

Would it not be more reasonable to have something like 
```
(setq c-c++-packages
  '(
    cc-mode
    disaster
    clang-format
    cmake-mode
    company
    (company-c-headers :toggle (configuration-layer/package-usedp 'company))
    company-ycmd
    flycheck
    gdb-mi
    ggtags :toggle (not (configuration-layer/package-usedp 'helm))
    helm-cscope
    helm-gtags :toggle (configuration-layer/package-usedp 'helm)
    semantic
    srefactor
    stickyfunc-enhance
    ycmd
    xcscope
    ))
```
in every language ?

ksjogo:
ggtags is providing the eldoc functionality afair.

tuhdo:
> ggtags is providing the eldoc functionality afair.

This is correct. `helm-gtags` provides similar features, but only `ggtags` provides `eldoc`, and is good at it. Imagine `eldoc` in just about any language, including esoteric one. It is really useful.

deb0ch:
from [gtags layer readme](http://spacemacs.org/layers/gtags/README.html#orgheadline5):
> (about eldoc)
> However, this feature is only activated for programming modes that are not one of these languages:
> 
>     - C mode
>     - C++ mode
>     - Common Lisp
>     - Emacs Lisp
>     - Python
>     - Ruby-mode
> 

right ?

ksjogo:
It is quite good with php
But we should fix the key shadowing

deb0ch:
Ooops, let me correct myself: why would it be activated in

- C mode
- C++ mode
- Common Lisp
- Emacs Lisp
- Python
- Ruby-mode

if in those modes, ggtags is not used to integrate eldoc ?

tuhdo:
> if in those modes, if helm-gtags is already used and ggtags is not used to integrate eldoc ?

`ggtags` and `helm-gtags` are supposed to be used together. `helm-gtags` for its commands with Helm interface, and `ggtags` solely for its `eldoc` integration. The modes you listed already have its own `eldoc` integration without requiring a tag database. However, it should be done for C/C++ modes, since not always people use `semantic` layer to get similar feature to `eldoc`.

deb0ch:
Then wouldn't a `ggtags :toggle (not (configuration-layer/package-usedp 'semantic))` be more appropriate in c/c++?

And does it also mean that ggtags can be deactivated in these other modes ?

TheBB:
If the problem is that `M-.` is being overridden, why don't we just unbind that?

ggtags-mode

* updates tags on save
* adds a completion-at-point function
* sets eldoc if not already set

So it does some useful things. Note specifically that it doesn't interfere with existing eldoc functionality so long as it's activated late, which should be easy to manage.

deb0ch:
Just making sure that it is not activated for nothing.

My concern is that startup is already long enough and keybindings crowded enough to not have to carry additional dead weight.

So for the same functionality, the less packages used, the simpler, the better, right ?

Although, it is still not crystal clear to me if ggtags is *really* useful even after your points: 

- **updates tags on save** : is it not dead simple to implement outside ggtags ?  
like `(add-to-list 'after-save-hook 'helm-gtags-update-tags)` ?
- **adds a completion-at-point function**: nobody ever complained for its absence before, am I wrong ? Plus, are there not already lots of other backends ?
- **advises current eldoc functionality** : only useful if semantic not activated *and* the user wants that feature (which btw interferes with echo messages).  
If the user wants that feature, shouldn't he just activate the semantic layer ?

It might well have a few (minor ?) additions to bring, but are they really worth the additional weight ?

tuhdo:
> My concern is that startup is already long enough

ggtags or helm-gtags should not affect startup time, since it's deferred.

> advises current eldoc functionality : only useful if semantic not activated and the user wants that feature (which btw interferes with echo messages).

The problem is that many users do not want to use Semantic, and Semantic only supports a few languages, while eldoc integration with `ggtags` is achieved for all languages that ctags/pygment supports..

deb0ch:
Even specifically for C/C++ modes ?

tuhdo:
> Even specifically for C/C++ modes ?

`semantic` not only supports C/C++ but other languages as well. The problem, it has its quirks that many people don't want to deal with. We never know if someone might or might not use `ggtags`'s features. If someone chooses to use `gtags` layer, then the eldoc feature from `ggtags` might be useful in their languages that has no other eldoc integration i.e. TCL or even Bash. In modes that `ggtags` has no use, we simply leave it there and it does nothing.

deb0ch:
> we simply leave it there and it does nothing.

Yeah, that was my concern. Dead weight.

But I guess OK, maybe that way is for now simpler for everyone.

TheBB:
I don't see the issue really. It's in a layer. You can choose not to use it.

choppsv1:
I agree, if you don't like ggtags don't enable it. Also, while I use helm I wouldn't want that fact to disable ggtags as an earlier comment suggested.

TheBB:
Closing this in favor of https://github.com/syl20bnr/spacemacs/pull/6815

Issue #6580, opened 2016-07-14
IPython 5 will cause problems.


cpaulik:
IPython5 is no longer using `readline` and they have basically dropped emacs `inferior shell` support.
See the last paragraph [here](http://ipython.readthedocs.io/en/stable/whatsnew/version5.html#id1) as well as the discussion in https://github.com/ipython/ipython/issues/9725 

When I have more time to look into this we might have to change a few settings in spacemacs. It seems that we might have to do something like https://github.com/jorgenschaefer/elpy/issues/908 or drop ipython support for the moment.

tacaswell:
https://github.com/matplotlib/matplotlib/pull/6734  will make the event loop integration much more automatic.  It should be v2.0 which should be released 'soon'.

cpaulik:
@tacaswell how does this relate to emacs support?

tacaswell:
Because the event loop integration was happen by chance because some of the qui frameworks register them selves into the inputhook that the readline based code looks at.

With the move away from readline you will need manually do the integration with the new IPython event loop.  That mpl PR makes the setup automatic.

cpaulik:
Aha, I have no idea how this all works. But this fix would only apply if matplotlib is imported?

TheBB:
It does seem peculiar that a matplitlib PR should fix an ipython issue…

tacaswell:
This was linked to from an issue related to mpl integration with ipython, I am probably a bit confused.

ThatGeoGuy:
If you don't all mind me chiming in, I think the problem may actually be solvable once the matplotlib folks remove whatever event loop stuff they've done to hook into the IPython readline. The issue referenced (which I filed) though seemed to be caused by not being able to use `jupyter console` within Emacs once I upgrade to IPython / jupyter-console 5.0. 

By default IPython will work in Emacs, provided you set the following: 

```elisp
(setq-default python-shell-interpreter "ipython"
              python-shell-interpreter-args "")
```

I think these are the defaults from the layer anyways in Spacemacs, so I don't think anything has to be done. The problem then, is using Emacs this way has traditionally meant that you can't use any of the matplotlib integrations, because the event loop would be tied to the process inside Emacs, and then any figures you load would lock up the event loop / process. To work around this, in the past you could get by with setting the following: 

```elisp
(setq-default python-shell-interpreter "jupyter"
              python-shell-interpreter-args "console")
```

[...]

PR #5997, opened 2016-05-06
Call python-setup-shell only once, not from hook

bmag:
`python-setup-shell` should only be called once. Fixes #5988.

PR #6627, opened 2016-07-18
Use --simple-prompt for IPython 5.

### LINKED ISSUE #6622: semantic and python layer cannot be used together
- Start Emacs
- add semantic and python layer
- create a .py file

Emacs hangs and cannot switch to the buffer.

Emacs successfully create and switch to the buffer.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.21
- Spacemacs branch: develop (rev. 9773ad6)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(semantic
 (python))
```

cpaulik:
IPython 5 does no longer use readline and so the emacs integration is
broken. See also
http://ipython.readthedocs.io/en/stable/whatsnew/version5.html#id1

Fix #6622 and related to #6580

### DIFF
--- a/layers/+lang/python/packages.el
+++ b/layers/+lang/python/packages.el
@@ -191,6 +191,9 @@
        (`on-project-switch
         (add-hook 'projectile-after-switch-project-hook
                   'spacemacs//pyenv-mode-set-local-version)))
+      ;; setup shell correctly on environment switch
+      (dolist (func '(pyenv-mode-set pyenv-mode-unset))
+        (advice-add func :after 'spacemacs/python-setup-shell))
       (spacemacs/set-leader-keys-for-major-mode 'python-mode
         "vu" 'pyenv-mode-unset
         "vs" 'pyenv-mode-set))))
@@ -204,7 +207,10 @@
         (spacemacs/set-leader-keys-for-major-mode mode
           "Va" 'pyvenv-activate
           "Vd" 'pyvenv-deactivate
-          "Vw" 'pyvenv-workon)))))
+          "Vw" 'pyvenv-workon))
+      ;; setup shell correctly on environment switch
+      (dolist (func '(pyvenv-activate pyvenv-deactivate pyvenv-workon))
+        (advice-add func :after 'spacemacs/python-setup-shell)))))
 
 (defun python/init-pylookup ()
   (use-package pylookup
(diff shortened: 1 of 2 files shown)


layers/+lang/python/packages.el
```diff
@@ -257,7 +257,10 @@
 
       (defun python-setup-shell ()
         (if (executable-find "ipython")
-            (setq python-shell-interpreter "ipython")
+            (progn (setq python-shell-interpreter "ipython")
+                   (when (string-equal
+                          (nth 1 (split-string (shell-command-to-string "ipython --version") "")) "5"))
```
nixmaniack: May be for making it future-proof? 
`(version<= "5" (replace-regexp-in-string "\n$" ""  (shell-command-to-string "ipython --version")))`

cpaulik: Thanks. That looks better. But I've used `(version<)`

layers/+lang/python/packages.el
```diff
@@ -257,7 +257,9 @@
 
       (defun python-setup-shell ()
         (if (executable-find "ipython")
-            (setq python-shell-interpreter "ipython")
+            (progn (setq python-shell-interpreter "ipython")
+                   (unless (version< "5" (replace-regexp-in-string "\n$" "" (shell-command-to-string "ipython --version")))
```
nixmaniack: Then you need to swap the order of arguments for `version<`

from docs:
>(version< V1 V2)

>Return t if version V1 is lower (older) than V2.

cpaulik: Thanks. I really did not pay attention yesterday evening. :smile:

cpaulik:
So I just noticed that this is a problem if there are multiple environments with different Ipython versions. Since `python-setup-shell` is only called once since https://github.com/syl20bnr/spacemacs/pull/5997 the interpreter is no longer adapted to a new environment.

@TheBB @bmag it might be good to advice the `pyenv-mode-set` and `pyvenv-activate` functions to set the correct shell interpreter. This should also not influence people who only use the global installation.

cpaulik:
This should now work with multiple environments. The shell is setup on every environment switch. It also checks if `ipython` is really installed in a `pyenv` environment since `executable-find` is fooled by the `pyenv` shims.

TheBB:
Do we need to potentially set these variables buffer-locally if they come from py(v)env?

yetone:
LGTM

cpaulik:
@TheBB  I don't think so. The py(v)env functions set the environment globally anyway. 

But unfortunately I just noticed that this solution does not work the first time if the first project you open is an ipython 5 project and your system/default ipython is version 4.

cpaulik:
The problem is because of `sematic-mode`. But unfortunately I have no idea yet why spacemacs hangs when switching from ipython4 to ipython5 for the first time. It also seems that this breaks semantic mode for other languages like elisp. Very strange. Another thing is that if the first switch is from ipython5 to ipython4 then semantic mode wants to run an inferior python interpreter when opening the first python file. :worried:

TheBB:

Seems likely we still have some work to do here because some of the symptoms I saw were very strange (as you say), but at least it's a lot more solid with this commit. :+1:

PR #6633, opened 2016-07-19
Issue template: direct link to bug report template

bmag:
Bugs that make Spacemacs unusable can't be reported from inside Spacemacs, so I added a direct link to the template used by `SPC h I`. The link is to the file in develop branch, and not master, because I assume it will be more up-to-date than master.

I made the PR against master (and not develop) because I think the change needs to be in master for users to see it when clicking the "New Issue" button.

### DIFF
--- a/.github/ISSUE_TEMPLATE
+++ b/.github/ISSUE_TEMPLATE
@@ -1,3 +1,4 @@
 If you are posting about a *bug* please close this issue and open a new one
-using [SPC h I] or [M-m h I] (I = capital i) from within Spacemacs.
+using [SPC h I] or [M-m h I] (I = capital i) from within Spacemacs. If the
+bug makes Spacemacs unusable, you can find the bug report template [here](https://github.com/syl20bnr/spacemacs/blob/develop/core/templates/REPORTING.template)
 You can delete this message to start typing.


a13ph:
I'm sorry, but why not just make it available as text on github directly? 
I assume I don't understand some reasons behind this decision to make users use emacs for bug reporting.
I agree that more issue-related information is better, but if we make it harder to post bad reports like this, we also make it harder to post issue reports in general as well. And that means less information, not more.
More issue reports is strictly better, especially since it allows for more interaction and contribution for wider audience.

bmag:
> why not just make it available as text on github directly?

IIRC it used to be that way, I don't know why it has changed. I'll let someone else answer this question.

>  I don't understand some reasons behind this decision to make users use emacs for bug reporting.

Provided Spacemacs isn't badly broken by the bug, it's easier to write the bug report from inside Emacs: system info is filled automatically, backtrace (if exists) is also added automatically (maybe it also needs to be the current buffer when calling `SPC h I`), Emacs window is bigger than GitHub's textbox.

> if we make it harder to post bad reports

Does this PR make it harder to post bad reports?

a13ph:
@bmag as noted by lack of @ mentioning you, I have no critique towards you here, or this PR itself. Only about more general reason of it's existence. 

 **Annoying pedant mode**, feel free to ignore:
_There might be other reasons why a person would prefer to post an issue through github_, so it's a rather prematurely excluded middle: 
- maintaining flow 
 - by continuing using github for issues 
   - (say i've remembered an issue while reading another one)
 - or by keeping emacs content the same, if it's crucial for the bug or other reasons
- not willing to publish information about themselves automatically without understanding the code which does it
- not being able to use Spacemacs for publishing issues despite it working quite well otherwise
 - not being proficient enough in emacs/evil to feel comfortable typing bug report in emacs 
 - having issues or restrictions with computer/network/...  running Spacemacs

TheBB:

PR #6661, opened 2016-07-23
Add ReST layer

wwguo:
The layer adds some functions to the built-in rst.el and supports Sphinx.


IbnSaeed:
Thanks, for providing this.

Can this be merged.

cpaulik:
Thank you for the PR. I think we should be careful here to not confuse [ReST](http://rest.elkstein.org/) and [reST](http://www.sphinx-doc.org/en/stable/rest.html) . So it might be good to rename the layer to `rst` or `restructured-text`? This would be less ambigous and probably easier to find.

I assume that you are the author of the extensions. It would be great if they could be made available on MELPA so that they are useful for the whole emacs ecosystem.

It would also be good to expose more keybindings. You bound some for hybrid state but they do not seem to be bound for evil or emacs users? 
The [conventions for plain text mark up languages](http://spacemacs.org/doc/CONVENTIONS.html#orgheadline27) might be useful here.

wwguo:
I was thinking of naming the layer as `rst'. Since the layer is based on the Emacs built-in rst.el and I'm not that familiar to Spacemacs, I'm afraid of creating conflicts between the layer and rst.el. If it will not cause any problem by naming it `rst', I'm OK with it. 

I'm the author of the extensions. Those extensions were written for the built-in rst.el many years ago. Some part of the extensions were actually added in the rst.el and the maintainer of rst.el was so kind listing me as an author of it. But I'm not a good coder and my extensions need to be reviewed carefully. Spacemacs is good that layers are isolated, so errors in my extensions will not cause much trouble to users. I prefer to use them as a part of a layer. 

I only use hybrid state and don't know what keybindings are convenient for users of other states. If possible, could you give me suggestions for the four keybindings used in other editing mode?

wwguo:
Thanks for suggestion!

I was thinking of naming the layer as `rst'. Since the layer is based on the Emacs built-in rst.el and I'm not that familiar to Spacemacs, I'm afraid of creating conflicts between the layer and rst.el. If it will not cause any problem by naming it `rst', I'm OK with it. 

I'm the author of the extensions. Those extensions were written for the built-in rst.el many years ago. Some part of the extensions were actually added in the rst.el and the maintainer of rst.el was so kind listing me as an author of it. But I'm not a good coder and my extensions need to be reviewed carefully. Spacemacs is good that layers are isolated, so errors in my extensions will not cause much trouble to users. I prefer to use them as a part of a layer. 

I only use hybrid state and don't know what keybindings are convenient for users of other states. If possible, could you give me suggestions for the four keybindings used in other editing mode?

TheBB:
There's no problem naming the layer `rst`.

I'm also not convinced that poor code quality is an argument for including code as local packages instead of putting them on MELPA. In spite of appearances, we don't actually *want* bugs here. :-) The Spacemacs layer system doesn't provide any more isolation than packages do.

Nevertheless I'm fine with it if it's understood to be temporary and/or an exception.

syl20bnr:
At least put your extensions in a GitHub repository if not already done and use a recipe to fetch them (look into the code for `:recipe` keyword to get examples), this way we avoid adding 1k lines of code to the Spacemacs repo and we avoid having to maintain it. If there is any bug related to the code we will add an issue on your repo.

wwguo:
OK. I might need some time to figure out how to do it in a proper way.

IbnSaeed:
@wwguo Any update ?

kxxoling:
Almost cannot wait for it! Any update?

istinspring:
is it merged?

d12frosted:
@istinspring since no one said that it's merged and PR is still open - no one merged it 😸 We are close to new release, so I am not sure if this layer gets in before release. Anyway, assigning this PR to @syl20bnr 😸

istinspring:
Yea would be cool. I donated him 10$ recently if it counts... ;-)

wwguo:
Sorry, I'm busy on personal issues recently. The layer still has two things need to be done. 

- Change the layer name to `rst'.
- Separated the extensions.

I will have a holiday early next month and will fix these two issues at that time.

d12frosted:
Oh welp, I missed that. Thanks for the heads up!

wwguo:
Hi guys, I need some help.  I put rst-mode extensions [here](https://github.com/wwguo/rst-mode-extensions) and changed the layer name. But I realized that I don't know how to tell the layer to download them. Could you give me an example? Then I don't need to go back to read all the references again. It has been a while since the last time I read the documents. I cannot remember those details. Thanks in advance.

wwguo:
I might find it. Is it just change  `(rst-directives :location local)` to  `(rst-directives :location (recipe :fetcher github :repo "wwguo/rst-mode-extensions/rst-directives")`. 

The layer uses three extensions. Is there any way to tell the layer to download them all instead of setting locations one by one?

syl20bnr:
Thank you ! 👍 

I made several changes:
- first, I renamed the layer to `restructuredtext` to remove the ambiguity with `restclient` layer
- then I moved the `sphinx` specific configuration to its own layer `sphinx`
- lastly I added support for auto-complete

Note that I kept only the leader key bindings since we don't usually bother with `C-c` key bindings in Spacemacs. You'll have to set those yourself in your own config.

syl20bnr:
@wwguo I pushed some improvements to `rst-sphynx` and applied more conventional key bindings (from CONVENTIONS.org), related commit: https://github.com/syl20bnr/spacemacs/commit/d1d58ef

Are you open for a pull request to your repo with theses changes?

wwguo:
Sure, it would be better to have my repo updated.

wwguo:
Just find I didn't push changes on my computer to my github repo. Don't know why I didn't do this.

PR #6671, opened 2016-07-25
Prefer $SPACEMACDIR/init.el over ~/.spacemacs

### LINKED ISSUE #6669: Wrong init file when using SPACEMACSDIR and ~/.spacemacs exists
I'm trying to make Emacs (on a Windows machine) work with a Spacemacs installation (.emacs.d) and configuration (.spacemacs.d) that are stored on a USB flash drive. If I have a `~/.spacemacs`, and I use `SPACEMACSDIR` environment variable to specify a directory on the USB drive, then Spacemacs tries to load `~/.spacemacs` instead of `%SPACEMACSDIR%/init.el`. The reason is that `dotspacemacs-filepath` is set wrong. PR incoming. There are no repro steps because I'm in the middle of trying to get it working.

bmag:
Previously, if SPACMEACSDIR is used, and both `$SPACEMACDIR/init.el` and `~/.spacemacs` exist, Spacemacs would use `~/.spacemacs` as the dotspacemacs file. Obviously, SPACEMACSDIR is for using a different config than the one in the home directory, so the correct thing to do is use `$SPACEMACSDIR/init.el`. Fixes #6669.

I am not sure what to do when SPACEMACSDIR is specified but `$SPACEMACSDIR/init.el` doesn't exist. For now I decided to fallback to `~/.spacemacs`, but I suspect the better option is to set `dotspacemacs-filepath` to `$SPACMEACSDIR/init.el` so that Spacemacs installs a new dotfile there.

### DIFF
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -15,38 +15,40 @@
 (defconst dotspacemacs-test-results-buffer "*dotfile-test-results*"
   "Name of the buffer to display dotfile test results.")
 
-(defconst dotspacemacs-directory
-  (let* ((env (getenv "SPACEMACSDIR"))
-         (env-dir (when env (expand-file-name (concat env "/"))))
-         (no-env-dir-default (expand-file-name
-                              (concat user-home-directory
-                                      ".spacemacs.d/"))))
+(let* ((env (getenv "SPACEMACSDIR"))
+       (env-dir (when env (expand-file-name (concat env "/"))))
+       (env-init (and env-dir (expand-file-name "init.el" env-dir)))
+       (no-env-dir-default (expand-file-name
+                            (concat user-home-directory
+                                    ".spacemacs.d/")))
+       (default-init (expand-file-name ".spacemacs" user-home-directory)))
+  (defconst dotspacemacs-directory
     (cond
      ((and env (file-exists-p env-dir))
       env-dir)
      ((file-exists-p no-env-dir-default)
       no-env-dir-default)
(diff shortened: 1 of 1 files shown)


core/core-dotspacemacs.el
```diff
@@ -15,38 +15,40 @@
...
-        spacemacs-dir-init
-      default))
-  "Filepath to the installed dotfile. If ~/.spacemacs exists,
-then this is used. If ~/.spacemacs does not exist, then check
-for init.el in dotspacemacs-directory and use this if it
-exists. Otherwise, fallback to ~/.spacemacs")
+      (cond
+       ((and env-init (file-exists-p env-init)) env-init)
+       ((file-exists-p default-init) default-init)
+       (and dotspacemacs-directory (file-exists-p spacemacs-dir-init) spacemacs-dir-init)
```
TheBB: For once Travis found a real bug. :-)

bmag: Fixed. Thank you Travis!

bmag:
closing due to some git problems, will open a new one

core/core-dotspacemacs.el
d12frosted: This should go to achievements! Nice work here, @bmag !

d12frosted:
> but I suspect the better option is to set dotspacemacs-filepath to $SPACMEACSDIR/init.el so that Spacemacs installs a new dotfile there.

Ya, I think it would be better.

PR #6672, opened 2016-07-25
Prefer $SPACEMACDIR/init.el over ~/.spacemacs

### LINKED ISSUE #6669: Wrong init file when using SPACEMACSDIR and ~/.spacemacs exists

bmag:
Previously, if SPACMEACSDIR is used, and both `$SPACEMACDIR/init.el` and `~/.spacemacs` exist, Spacemacs would use `~/.spacemacs` as the dotspacemacs file. Obviously, SPACEMACSDIR is for using a different config than the one in the home directory, so the correct thing to do is use `$SPACEMACSDIR/init.el`. Fixes #6669.

I am not sure what to do when SPACEMACSDIR is specified but `$SPACEMACSDIR/init.el` doesn't exist. For now I decided to fallback to `~/.spacemacs`, but I suspect the better option is to set `dotspacemacs-filepath` to `$SPACMEACSDIR/init.el` so that Spacemacs installs a new dotfile there.

### DIFF
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -40,15 +40,15 @@ directories exist, this variable will be nil.")
                                  (concat dotspacemacs-directory
                                          "init.el"))))
       (cond
-       ((and env-init (file-exists-p env-init)) env-init)
+       (env-init)
        ((file-exists-p default-init) default-init)
        ((and dotspacemacs-directory (file-exists-p spacemacs-dir-init)) spacemacs-dir-init)
        (t default-init)))
-    "Filepath to the installed dotfile. If SPACEMACSDIR is used
-and contains an init.el file, then this file is ued. Otherwise,
-if ~/.spacemacs exists, then this is used. If ~/.spacemacs does
-not exist, then check for init.el in dotspacemacs-directory and
-use this if it exists. Otherwise, fallback to ~/.spacemacs"))
+    "Filepath to the installed dotfile. If SPACEMACSDIR is given
+then SPACEMACSDIR/init.el is used. Otherwise, if ~/.spacemacs
+exists, then this is used. If ~/.spacemacs does not exist, then
+check for init.el in dotspacemacs-directory and use this if it
+exists. Otherwise, fallback to ~/.spacemacs"))
 
 (defvar dotspacemacs-distribution 'spacemacs
   "Base distribution to use. This is a layer contained in the directory
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -15,38 +15,40 @@
 (defconst dotspacemacs-test-results-buffer "*dotfile-test-results*"
   "Name of the buffer to display dotfile test results.")
 
-(defconst dotspacemacs-directory
-  (let* ((env (getenv "SPACEMACSDIR"))
-         (env-dir (when env (expand-file-name (concat env "/"))))
-         (no-env-dir-default (expand-file-name
-                              (concat user-home-directory
-                                      ".spacemacs.d/"))))
+(let* ((env (getenv "SPACEMACSDIR"))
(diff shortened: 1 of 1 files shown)


d12frosted:
Oh, gonna copy-paste my comment here.

---

> but I suspect the better option is to set dotspacemacs-filepath to $SPACMEACSDIR/init.el so that Spacemacs installs a new dotfile there.

Ya, I think it would be better.

bmag:
@d12frosted changed to use and install `$SPACEMACSDIR/init.el`also  when the file doesn't exist.
Will squash when there's a final decision.

TheBB:

dijonkitchen:
Why are these cherry-picked instead of merged into the branch? 

Also, what's the cherry-picked commit so that we can check if it's in master yet or not?

bmag:
> Why are these cherry-picked instead of merged into the branch?

To keep the history linear, mostly.

> what's the cherry-picked commit so that we can check if it's in master yet or not?

Don't know the commit ID, but you can search by commit message. Anyway, this commit is in master.

dijonkitchen:
Thanks, I found it here: https://github.com/syl20bnr/spacemacs/commit/6d60d2ef05869ca6007f72befdfd7c7aadd49c50

I figure [rebasing](https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase) to keep history linear would be easier than cherry-picking commits though.

d12frosted:
TL;DR it's a matter of preference 😸 

@dijonkitchen rebase (in terms of dealing with PR) is usually made on the feature branch and then you merge the rebased feature branch into master which equals to cherry-picking commits. Equals in the sense of the result. You have no way to tell the difference just by looking at the commit in the main branch (be it `master`, `develop` or whatever). Also, when the feature branch has only one commit it's usually simpler to use cherry-pick (you can simply cherry-pick the 'branch').

I don't know about the workflow of the others, but personally, I don't use GitHub buttons for merging PRs (I know that it can squash/rebase and all the sweet stuff) - I deal with PRs in Emacs. Sometimes I cherry-pick them into `develop` branch and use them a while before actually pushing. AFAIK, Sylvain uses a similar approach, but he even marks PRs that he cherry-picked locally with a 'Merging...' label.

d12frosted:
> what's the cherry-picked commit so that we can check if it's in master yet or not?

BTW, even if you use rebase-merge approach, the commit hash will change, because it most likely will have a different parent.

> You have no way to tell the difference just by looking at the commit in the main branch

Ok, I just realised that one of the side effects of using cherry-pick is that the cherry-picked commit is committed by the person performing cherry-pick. 

But when you rebase-merge PR that is already up to date with the main/target branch, committer is the author of PR, not the person doing rebase-merge. 

I hope I didn't miss something :)

dijonkitchen:
Thanks for taking the time to answer so clearly. It's definitely a matter of preference (and perhaps ease with Emacs); I figure the transparency for other users would be good on Github though so people can track things and what's where with little effort. Maybe the best approach if you all want to stick with what you're doing is to just mention the new commit hash in a comment for posterity. 

Thanks again for all your great work! Love Spacemacs!

d12frosted:
> Thanks for taking the time to answer so clearly

No problem :) 

> Maybe the best approach if you all want to stick with what you're doing is to just mention the new commit hash in a comment for posterity.

Good idea. Maybe I'll start doing this and we'll see how it's going.

PR #6674, opened 2016-07-25
[scala] Disable Ensime autostart

d1egoaz:
Disables Ensime autostart, default should be `false`.

When you work with multiple projects this will start an ensime instance every time a scala file is opened. It will be a huge memory consumer.

Also there is a bug with current spacemacs implementation of autostart: https://github.com/syl20bnr/spacemacs/pull/6530


TheBB:

PR #6675, opened 2016-07-25
Add mu installation note for OSX/Homebrew users

Karunamon:
I discovered this after a good deal of messing about trying to get mu to work, and figured it may be useful to include in the official docs.

### DIFF
--- a/layers/+email/mu4e/README.org
+++ b/layers/+email/mu4e/README.org
@@ -19,6 +19,14 @@
 In order to use this layer you must install mu and mu4e separately. Typically
 mu4e will be bundled with mu (this is the case on many Linux distributions).
 
+If you're on OS X and install mu using Homebrew, you must specify the
+location of your Emacs binary at install time using the EMACS environment
+variable, as well as passing the --with-emacs option:
+
+#+begin_src shell
+EMACS=$(which emacs) brew install homebrew --with-emacs
+#+end_src
+
 If the installation directory of mu4e is not in Emacs’ load path, you can set
 the layer variable =mu4e-installation-path=, for example:
 


layers/+email/mu4e/README.org
```diff
@@ -19,6 +19,14 @@
 In order to use this layer you must install mu and mu4e separately. Typically
 mu4e will be bundled with mu (this is the case on many Linux distributions).
 
+If you're on OS X and install mu using Homebrew, you must specify the
+location of your Emacs binary at install time using the EMACS environment
+variable, as well as passing the --with-emacs option:
+
+#+begin_src shell
+EMACS=$(which emacs) brew install homebrew --with-emacs
```
d12frosted: Hey, what is the meaning of this?

d12frosted: Did you just mean `brew install mu --with-emacs`?

d12frosted:
Hey, thanks for PR. It's a pity to hear that you struggled with this. 

I am not sure that `EMACS` environment variable is required. Though you have to install `mu` using `brew install mu --with-emacs` and then setup `mu4e-installation-path` as it mentioned in readme file. For you it should be something like `"/usr/local/Cellar/mu/0.9.16/share/emacs/site-lisp/"` with respect to latest version.

Karunamon:
Heh, you caught me halfway through writing this up.

The idea here was simplest minimum viable configuration. In the spirit of some of the OSX specifics sprinkled throughout the documentation, and the fact that just installing `mu` without the extra command line options won't give you a working config.

I had to do the `EMACS=` bit or else mu4e would not be correctly linked (I think it was putting it in a path accessible to the system emacs, rather than spacemacs). It also obviates the need to set the `mu4e-installation-path` option.

d12frosted:
Oh, just checked my configs. Totally forgot about another location of `brew` packages that should be used instead of installation path. You might want to use `/usr/local/opt/mu/share/emacs/site-lisp/`.

d12frosted:
> Heh, you caught me halfway through writing this up.

Haha. It happens 😸 💃

Karunamon:
I just ran a test again to confirm this, by backing up my `.emacs.d`, doing a fresh install, and blowing away `mu`.

Steps look like:

1. Install spacemacs
2. Edit `.emacs.d` and add the `mu4e` layer
3. Sync config
4. Attempt to run with: `SPC a m`:  `Cannot open load file: No such file or directory: mu4e`
5. Install `mu` using the `--with-emacs` argument to Homebrew
6. Relaunch spacemacs and attempt running again. **Same error**
7. Uninstall `mu`, reinstall `mu` with `EMACS=$(which emacs) brew install mu --with-emacs`
8. Relaunch spacemacs, attempt running: **It works!**

d12frosted:
I still don't get the point of `EMACS=$(which emacs) brew install mu --with-emacs`. Why is `brew` part in `EMACS` variable? Could you please try setting the value of `mu4e-installation-path` to `/usr/local/opt/mu/share/emacs/site-lisp/`? This is what actually README file asks users to do after installing `mu`. 

Don't get me wrong, I am all for mentioning that users must install `mu` with `brew` using `brew install mu --with-emacs`. But the part about `EMACS=$(which emacs) brew install mu --with-emacs` seems suspicious to me.

d12frosted:
Also, after you install `mu` using `brew install mu` executing `brew install mu --with-emacs` has no effect as it should warn you that `mu` was already installed. You should use `brew reinstall mu --with-emacs`.

TheBB:
> I still don't get the point of `EMACS=$(which emacs) brew install mu --with-emacs`. Why is `brew` part in `EMACS` variable?

What this does is it assigns the output of `which emacs` to the `EMACS` variable for the duration of the command `brew install mu --with-emacs`. So the `brew` is part of the command that runs, not the value being assigned.

d12frosted:
Thanks for explanation, @TheBB. But there is no separator between `EMACS=$(which emacs)` and `brew install mu --with-emacs`. So `EMACS=$(which emacs) brew install mu --with-emacs` actually only sets var. At least if I understand it properly 😸 💃 

But your comment forced me to investigate `mu` build script and it seems that setting `EMACS` environment variable has the same effect as using `--with-emacs` flag in the first place while installing `mu`. So I am against overcomplicating installation instructions. 

If you ever install `brew` package with wrong options, you have to do the following.

```
$ brew uninstall package
$ brew install package [options]
```

Even `reinstall` reuses previously used options, so when you really want to build with different options the best way is to use `uninstall` and `install` 😸 Hope it helps.

TheBB:
No, I think you're misunderstanding. What separator are you looking for? Try it:

```sh
$ TEST=test env | grep TEST
TEST=test
```

d12frosted:
@TheBB, my bad 😿  I really totally misunderstood that expression. Sorry for overall confusion around it. 

Anyway, the part about `mu` build script still stands. You just have to properly install it in the first place.

Karunamon:
>Could you please try setting the value of mu4e-installation-path to /usr/local/opt/mu/share/emacs/site-lisp/? This is what actually README file asks users to do after installing mu.

Tried, and that works after a restart of emacs (just doing an evaluate of the setq isn't enough).

So, apparently, this whole thing comes down to whether you want to modify the `brew install mu` command (which you have to do anyways with the `--with-emacs` flag) and have it Just Work, or set the path in emacs after the install.

Either will lead to a working installation, but the former way seems easier.

TheBB:
I dunno what you two have arrived at, but I am assuming it should be mu that's installed and not homebrew?

d12frosted:
@TheBB Well, I am all for adding a note that OS X and brew users must install `mu` using `brew install mu --with-emacs`. But that `var` set is not required as it has the same effect as setting using `--with-emacs` option. And more over, `EMACS=$(which emacs) brew install homebrew --with-emacs` doesn't work in `fish`, while `brew install mu --with-emacs` is universal.

TheBB:
Okay, I'll follow @d12frosted's suggestion. Also changed "homebrew" to "mu".

d12frosted:
@TheBB haha, indeed, `mu`, not `homebrew` :D

Issue #6591, opened 2016-07-15
haskell: swap keys for stylish-haskell and hindent


ljli:
I use hindent much more than stylish-haskell. I imagine that is quite common. Therefore it makes sense to me to have the slightly simpler `, f` for hindent and `, F` for stylish-haskell. Currently it's the other way around. I recognize this is pretty minor, but if no one objects I could just make a PR and we are done with it. I also think stylish-haskell could in principle be subsumed by hindent or one key could just do both (that is if the user wants both). The change seems beneficial anyway.

d12frosted:
Hey @ljli,

Yeah, it really makes sense. I never use `, f`, but time to time I do use `, F`. So PR is warmly welcome.

PR #6694, opened 2016-07-27
Swap haskell layer specific keys f and F

ljli:
Brought up in #6591. hindent is potentially much more used than stylish-haskell, therefore it now has the fewer keystrokes.


d12frosted:
Thanks for the PR! Also you (@ljli and others) should note that it's possible to enable automatic buffer stylish on save. At least, it works for me 😸 So I am really for this change 💃 

The only thing I would like you to update are [keybindings](https://github.com/syl20bnr/spacemacs/blob/develop/layers/+lang/haskell/README.org#key-bindings), keybindings 😸 💃

ljli:
Ah yes, I missed that. Updated!

d12frosted:

ljli:
I don't think this ever made it into develop.

d12frosted:
@ljli Sorry, for some reason it didn't get there. Fixed that.

Issue #3870, opened 2015-11-18
Racket layer dependencies


kalouantonis:
A minimal install of racket will not work with the racket layer. We should add a line in the README to notify people that the DrRacket package is required or at least the minimum required packages to be installed with raco.

There should at least be more informative output than what I got when first booting up with a minimal install: 

```
standard-module-name-resolver: collection not found
  for module path: (submod at-exp reader)
  collection: "at-exp"
  in collection directories:
   /Users/slacker/Library/Racket/6.2.1/collects
   /usr/local/Cellar/racket/6.2.1/share/racket/collects
   /usr/local/Cellar/racket/6.2.1/share/racket/pkgs/racket-lib
  context...:
   show-collection-err
   standard-module-name-resolver
   standard-module-name-resolver
   standard-module-name-resolver
   standard-module-name-resolver
```

lverns:
The docs can be viewed via `SPC h SPC` and then typing "racket". I can see that the documentation still hasn't been updated; however, I can't seem to reproduce the error that you have posted. Could you give the steps to reproduce?

kalouantonis:
I first installed the minimal distribution of racket using OSX brew, but its the same if you download if off the site. I then added racket to my `dotspacemacs-configuration-layers` as outlined in the README. Lastly, I execute `M-x racket-repl` and I get the error above.

To fix this I run `raco pkg install drracket`

lverns:
That **is** weird. With _no_ racket installed, I get a nice reasonable error. Once I install the "minimal distribution", I get the same error as you do. This is probably something that should be reported [upstream](https://github.com/greghendershott/racket-mode).

d12frosted:
Documentation change is now on master, so closing this.

PR #6696, opened 2016-07-28
Update racket-layer README.org

lverns:
This is in response to issue #3870.


d12frosted:

PR #6714, opened 2016-07-30
Add option for helm fuzzy matching

quicknir:
Adds an option in dotspacemacs which controls whether or not most helm sources have fuzzy matching enabled.

### DIFF
--- a/core/templates/.spacemacs.template
+++ b/core/templates/.spacemacs.template
@@ -186,6 +186,11 @@ values."
    ;; define the position to display `helm', options are `bottom', `top',
    ;; `left', or `right'. (default 'bottom)
    dotspacemacs-helm-position 'bottom
+   ;; Controls fuzzy matching in helm. If set to `always', force fuzzy matching
+   ;; in all non-asynchronous sources. If set to `source', preserve individual
+   ;; source settings. Else, disable fuzzy matching in all sources.
+   ;; Default 'always.
+   dotspacemacs-helm-use-fuzzy 'always
    ;; If non nil the paste micro-state is enabled. When enabled pressing `p`
    ;; several times cycle between the kill ring content. (default nil)
    dotspacemacs-enable-paste-transient-state nil
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -202,6 +202,12 @@ start.")
 (defvar dotspacemacs-helm-position 'bottom
   "Position in which to show the `helm' mini-buffer.")
 
+(defvar dotspacemacs-helm-use-fuzzy 'always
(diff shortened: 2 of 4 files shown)


quicknir:
Note that this should probably ideally be a layer variable, but cannot be because the functionality is defined in core. Presumably at some point there will be a refactor which moves the dotspacemacs helm- variables out into the layer.

core/templates/.spacemacs.template
```diff
@@ -186,6 +186,8 @@ values."
    ;; define the position to display `helm', options are `bottom', `top',
    ;; `left', or `right'. (default 'bottom)
    dotspacemacs-helm-position 'bottom
+   ;; If non nil, helm will use fuzzy matching in most sources
```
NJBS: Can you include the default value of the variable in the docstring?

quicknir:
@NJBS Done!

d12frosted:
Hey @quicknir 

Since `helm` is separate layer now, I think it would make more sense to move this configuration variable (`dotspacemacs-helm-use-fuzzy`) to the layer itself. What do you think?

quicknir:
@d12frosted I agree in principal and actually tried to do this for my first attempt. However, it was not successful; the code that actually ends up controlling this is in spacemacs-completion which is in the core. So I suspect that `spacemacs//helm-make-source` is defined before the variable from the layer is defined. So it would probably be more work (at least for me) to get this working in the way you describe.

IMHO, I think leaving this is a dotspacemacs variable is ok for now. There are a few other helm dotspacemacs variables and no layer variables currently, so a future refactor should move them all simultaneously (and should probably move helm-make-source to the helm layer as well). In some ways one could argue it's preferable to have all the config for helm in the same location, even if it's no longer the correct location.

Let me know what you think, if you still think it should be a layer variable then some advice on implementation would be appreciated because the naive approach did not work.

quicknir:
@d12frosted Update?

zzjjzzgggg:
Will this `dotspacemacs-helm-use-fuzzy` turn on/off helm fuzzy match globally? What if I want to define a `helm-source-async` with `:fuzzy-match nil`, but this `dotspacemacs-helm-use-fuzzy` is set to `t`?

TheBB:
Presumably there ought to be three choices: always on, always off, or according to source definition.

quicknir:
@zzjjzzgggg @TheBB Hopefully this addresses your concerns?

@d12frosted Sorry to ping again.

TheBB:
Looks good to me. Thanks! Cherry-picked in develop. You can safely delete your branch.

@syl20bnr We should maybe move helm dotspacemacs variables to layer variables.

d12frosted:
@quicknir sorry, was busy lately. Just saw your message :/

Yeah, we can do it in future. 😸

Issue #6027, opened 2016-05-10
Latest spaceline does not respect powerline-default-separator


carbohydratesn:
Hello, updating to spaceline-20160510.542 forces use of the default separators, even if another (box in this case) is specified in config.  

Rolling back to spaceline-20160507.302 is a temporary fix.

NJBS:
You'll have to call `spaceline-compile` in your `user-config` after setting `powerline-default-separator`.

See [discussion in gitter](https://gitter.im/syl20bnr/spacemacs?at=57321dffed393f3409b187a1) for more info.

TheBB:
@carbohydratesn If this fixed your problem (judging by your reactions) please close this issue. That goes for the other one too.

nwolfe:
Yeah, this seems like something that users will have to discover themselves when they start configuring their powerline/spaceline.

👍 for closing the issue

cpaulik:
Should that not be documented? https://github.com/syl20bnr/spacemacs/blob/develop/doc/DOCUMENTATION.org#powerline-separators says nothing about calling `(spaceline-compile)`

nwolfe:
Ah, forgot we had documentation around powerline already!  Yes we should definitely add mention of `(spaceline-compile)` to that section.

carbohydratesn:
Documentation sounds great, closed indeed

TheBB:
I updated the docs, thanks for finding that.

spijet:
After adding (spaceline-compile) to ~/.spacemacs, there's no more need to reload config (SPC-f-e-R) after starting Emacs in daemon mode and attaching a GUI emacsclient to get custom powerline separator right! :smile:

PR #6716, opened 2016-07-30
Update DOCUMENTATION.org - missing doc for #6027

mateuspv:

Thank you <3
Hi, let's add a missing documentation for:
"You'll have to call spaceline-compile in your user-config after setting powerline-default-separator."
described at issue #6027

Thanks!


TheBB:
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+spacemacs/spacemacs-ui-visual/packages.el#L247

We now call `spaceline-compile` after user-config, so no need for the user to do it. I realize this PR is against master and not develop, but I don't think we need a new release for this detail when 0.200 is so close. Therefore I will close this. Thank you though!

PR #6717, opened 2016-07-31
Make swoop optional and helm-ag follow

quicknir:
- Introduces a layer variable for helm to disable swoop
- helm-ag for single files and open buffers now
  starts in follow mode

Unfortunately there's a minor problem with my code I can't figure out. If I start spacemacs and do `SPC s s` or similar (with swoop disabled so helm-ag is launched with follow mode) then I get: helm-attr: Symbol's value as variable is void: helm-source-do-ag. However, if I run some other helm-ag command (`SPC /`), after that it's fine. So it seems like an issue with when things are being loaded. Not sure exactly how to fix it though.

### DIFF
--- a/layers/+completion/helm/packages.el
+++ b/layers/+completion/helm/packages.el
@@ -117,7 +117,17 @@
                 (lambda ()
                   (unless (configuration-layer/package-usedp 'smex)
                     (spacemacs/set-leader-keys
-                      dotspacemacs-emacs-command-key 'helm-M-x)))))
+                      dotspacemacs-emacs-command-key 'helm-M-x))))
+
+      ;; In follow mode, follow on keystrokes as well as up/down
+      (setq helm-follow-follow-on-update t)
+      ;; Run any search with follow mode on
+      (defun spacemacs//helm-do-search-with-follow (helm-source helm-command)
+        (let ((prev-follow-val (helm-attr 'follow (symbol-value helm-source))))
+          (helm-attrset 'follow 1 (symbol-value helm-source))
+          (call-interactively helm-command)
+          (helm-attrset 'follow prev-follow-val (symbol-value helm-source)))))
+
     :config
     (progn
       (helm-mode)
@@ -183,7 +193,7 @@
       (defun spacemacs/helm-file-do-ag (&optional _)
         "Wrapper to execute `helm-ag-this-file.'"
         (interactive)
-        (helm-ag-this-file))
+        (spacemacs//helm-do-search-with-follow 'helm-source-do-ag 'helm-do-ag-this-file))
 
       (defun spacemacs/helm-file-do-ag-region-or-symbol ()
         "Search in current file with `ag' using a default input."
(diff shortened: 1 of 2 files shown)


quicknir:
I think I understand the problem. Would it be alright to move the function definitions from :init into :config? Should I move all of them, or only the single-file and buffers ones (i.e. the one that need it) ?

TheBB:
You should move function definitions into `funcs.el`. They shouldn't be in `:init` *or* `:config`. This is sort of an ongoing process.

bmag:
>  If I start spacemacs and do SPC s s or similar (with swoop disabled so helm-ag is launched with follow mode) then I get: helm-attr: Symbol's value as variable is void: helm-source-do-ag.

Helm is loaded ~1 second after startup is done (`:defer 1` in `use-package` call), so the error is probably because Helm isn't loaded yet. Note that `helm-source-do-ag` is not auto-loaded either.

layers/+completion/helm/packages.el
```diff
@@ -117,7 +117,17 @@
...
-                      dotspacemacs-emacs-command-key 'helm-M-x)))))
+                      dotspacemacs-emacs-command-key 'helm-M-x))))
+
+      ;; In follow mode, follow on keystrokes as well as up/down
+      (setq helm-follow-follow-on-update t)
+      ;; Run any search with follow mode on
+      (defun spacemacs//helm-do-search-with-follow (helm-source helm-command)
+        (let ((prev-follow-val (helm-attr 'follow (symbol-value helm-source))))
+          (helm-attrset 'follow 1 (symbol-value helm-source))
+          (call-interactively helm-command)
```
bmag: Is there a reason to prefer `call-interactively` over `funcall`? If the intention is to change as little of the behavior as possible, then:
```elisp
(defun do-foo ()
  (interactive)
  (funcall 'foo))
```
Is closer to:
```elisp
(defun do-foo ()
  (interactive)
  (foo))
```
than:
```elisp
(defun do-foo ()
  (interactive)
  (call-interactively 'foo))
```

syl20bnr:
> Helm is loaded ~1 second after startup is done

I wonder if we should remove this.

quicknir:
@TheBB  What exactly should I move into funcs.el? The utility function that I wrote, all of the various helm-ag functions that we use to bind to, or both? Also if you have an example of a layer with sufficient complexity that does things correctly that I can use as a model, can you point me towards it?

@bmag I don't think it's related but I could be wrong, I didn't just wait one second.

@syl20bnr Do you want me to remove this, as a second commit, as part of this PR?

layers/+completion/helm/packages.el
quicknir: @bmag I can change the caall interatively, not a problem.

TheBB:
All defuns should be in funcs.el, but it's fine if you just move the ones you are touching rather than all the ones in the file. See the spacemacs-layouts layer for example. There are almost no defuns left in packages.el there.

quicknir:
This PR would need some work, I am going to close this for now. In the future I may tackle it again.

PR #6720, opened 2016-07-31
scala: add SPC m x for sbt-hydra

jdnavarro:
This PR introduces the main key binding to interact with hydra in `sbt-mode`. See section on *Hydra* in [ensime sbt-mode documentation[(https://ensime.github.io/editors/emacs/sbt-mode/).

I'm not sure `x` is the right key binding so please let me know any other key that makes more sense than `x`, or if you believe this belongs under an existing scala prefix. Notice also that `sbt-hydra` provides several functions so this binding may become a prefixed key under its own prefix section in the future.

### DIFF
--- a/layers/+lang/scala/funcs.el
+++ b/layers/+lang/scala/funcs.el
@@ -111,3 +111,8 @@ point to the position of the join."
   "Yank to kill ring and print full type name at point to the minibuffer."
   (interactive)
   (ensime-type-at-point t t))
+
+(defun sbt-hydra ()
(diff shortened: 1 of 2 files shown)


jdnavarro:
What about `SPC m b h`? Do you know any other language layer that already does subproject builds?

TheBB:

I moved it to SPC mbh. I also had to relocate the binding because this function belongs to the sbt package, not ensime: https://github.com/syl20bnr/spacemacs/commit/b8ca74e4b36d9d4fe1244d50ea148e93af7a00ed

PR #6725, opened 2016-07-31
move the eshell-z history file into the eshell cache directory

Stebalien:
Previously, it was saved directly to the user's home directory (`~/.z`) :scream:.

### DIFF
--- a/layers/+tools/shell/packages.el
+++ b/layers/+tools/shell/packages.el
@@ -186,7 +186,8 @@ is achieved by adding the relevant text properties."
     :defer t
     :init
     (with-eval-after-load 'eshell
-      (require 'eshell-z))))
+      (require 'eshell-z)
+      (setq eshell-z-freq-dir-hash-table-file-name (concat eshell-directory-name "/z")))))
 
 (defun shell/pre-init-helm ()
   (spacemacs|use-package-add-hook helm


d12frosted:
To be precise it saves to `$_Z_DATA` or `~/.z` 😸 I am not sure how people use this package, but I believe that it would be better to set `eshell-z-freq-dir-hash-table-file-name` to `(or (getenv "_Z_DATA") (concat eshell-directory-name "/z"))`, so people who use `$_Z_DATA` variable don't have to change the value of `eshell-z-freq-dir-hash-table-file-name`. 

What do you think?

Stebalien:
Ah... Actually, `z` appears to be a [thing](https://github.com/rupa/z) beyond emacs so it might be better to not change this at all (the env var made me a bit suspicious). However, I'm not happy with spacemacs creating an ambiguously named dotfile in `~/` by default for no obvious reason (I was very confused when I noticed this file in my home directory). Would it make sense to just disable this by default and make it a setting? I can always just disable the package (that's what I'm doing now); I submitted this PR primarily because it looked like a simple oversight.

TheBB:
In that case I don't think we need to do anything. It's a variable that users can set as they please. We can possibly document it if it's a problem.

Stebalien:
Meh. I honestly just don't care enough. From what I've seen, most people aren't quite so anal about keeping their home directory clean so this isn't really that big of an issue.

d12frosted:
> most people aren't quite so anal about keeping their home directory clean

Just the quote of the day 💃

PR #6737, opened 2016-08-02
rails: add which key prefixes for rails

nixmaniack:
* layers/+frameworks/ruby-on-rails/packages.el (ruby-on-rails/init-projectile-rails):
  added which-key prefixes for rails
* layers/+frameworks/ruby-on-rails/packages.el (ruby-on-rails/post-init-which-key):
  diminish `projectile-rails` prefix in which-key menu to make commands
  visible (since we're in rails menu, it makes sense to diminish common
  prefix)


TheBB:

PR #1428, opened 2015-04-28
[Fix #1409, #1407, #1424,#1442] Explicitly manage popwin 

tuhdo:
Set popwin:special-display-config to nil and let us manage it from
scratch. Doing so, we avoid popwin to mess up windows of some package
i.e. SLIME. Similar to how auto-completion layer is managed, if a layer
want to have popwin support, it must be explicitly specified in that
layer.

The default popwin:special-display-config contains obsolete config such
as anything (predecessor of Helm) and buffer that should be stick around
until we press C-g (like shell command output since we may need the
output for doing something else).

We don't need to use popwin:display-buffer since it can still mess up
window configuration, like this issue #1409. Simply add helm buffers to
display-buffer-alist, and use display-buffer-in-side-window; without any
argument, the default is bottom. We also inhibit Helm buffers to reuse
the existing windows with (inhibit-same-window . t) and a window
height (window-height . 0.4).

Also prevent popwin from piling up display-buffer-alist since everytime
it is reactivated, it adds its own buffers that already existed in
display-buffer-alist, by restoring display-buffer-alist after
popwin-mode is enabled again.

PR #1397, opened 2015-04-26
[Fix #1396] Do not use other function to display Helm buffer

tuhdo:
So when users add custom function to pop a buffer in some random window,
Helm won't be affected.

PR #6746, opened 2016-08-02
Move helm-popwin display config to helm layer

### LINKED ISSUE #1396: New helm window popup conflicts
The code introduced in https://github.com/syl20bnr/spacemacs/commit/bfe99322f7082466e24fed99091367374f9bf8b9 - #1360 breaks with my current window config. 

I have this in my config:
`(setq display-buffer-base-action '((display-buffer-use-some-window)))`

What that does for me is reuses my last 'other-window' for commands like `dired-other-window` (`o` in dired-mode by default) and `magit-status`. If I don't have this config set, these 2 commands split my current window, which I do not like. And if I do, helm shows up in my 'other-window' as well as on the bottom.

Any suggestions for making these compatible, and/or alternatively adopting this style as the default behavior?

edit: updated config.

### LINKED ISSUE #6634: Excluding popwin breaks helm
#### Description :octocat:
Excluding `popwin` causes error when opening helm sessions. Instead of helm session, there's an error message. As a result, helm can't be used, causing Spacemacs to be unusable.

- Enable `helm` layer
- Add `popwin` to `dotspacemacs-excluded-packages`
- Start Emacs
- Wait a second for Helm to load (Spacemacs loads Helm ~1 second after startup is done)
- Start helm session: `SPC f f`

Error message is printed in the echo area, Helm session does *not* open.
The error message: `Symbol’s function definition is void: popwin-mode`

With `SPC t D` (`spacemacs/toggle-debug-on-error`) before `SPC f f`, I get the backtrace posted below.

Helm session opens in the same place as usual, i.e. at the bottom with a height of 0.3 of the frame height. No error.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 25.0.95.1
- Spacemacs: 0.105.21
- Spacemacs branch: buffer-predicate (rev. 08f5291)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(helm emacs-lisp)
```

#### Backtrace :paw_prints:
```
Debugger entered--Lisp error: (void-function popwin-mode)
  (popwin-mode -1)
  (let ((display-buffer-base-action (quote (nil)))) (setq spacemacs-display-buffer-alist display-buffer-alist) (setq display-buffer-alist nil) (popwin-mode -1) (if helm-prevent-escaping-from-minibuffer (progn (define-key evil-motion-state-map [down-mouse-1] nil))))
  spacemacs//helm-prepare-display()
  run-hooks(helm-after-initialize-hook)
  helm-log-run-hook(helm-after-initialize-hook)
  helm-initial-setup("/home/bar/some")
  helm-initialize(nil "/home/bar/" "/home/bar/some" helm-source-find-files)
  helm-internal(helm-source-find-files "/home/bar/" "Find files or url: " nil nil "*helm find files*" nil "/home/bar/some" nil)
  apply(helm-internal (helm-source-find-files "/home/bar/" "Find files or url: " nil nil "*helm find files*" nil "/home/bar/some" nil))
  helm(helm-source-find-files "/home/bar/" "Find files or url: " nil nil "*helm find files*" nil "/home/bar/some" nil)
  apply(helm (helm-source-find-fil
```

bmag:
Rename and move `spacemacs//helm-prepare-display` and `spacemacs//restore-previous-display-config` to Helm layer. Use them only when `popwin` is used. Fixes #6634 

Misc:
- Move the code that prevents escaping the minibuffer with the mouse to separate functions, leave it in spacemacs-completion layer.
- Delete let-binding of `display-buffer-base-action`. It was introduced in #1397 to fix #1396, but didn't do anything since we stopped using `(popwin:display-buffer helm-buffer t)` in #1428.

### DIFF
--- a/layers/+completion/helm/funcs.el
+++ b/layers/+completion/helm/funcs.el
@@ -11,10 +11,32 @@
 
 
 
+(defvar helm--popwin-mode nil
+  "Temp variable to store `popwin-mode''s value.")
+
 (defun spacemacs//helm-cleanup ()
   "Cleanup some helm related states when quitting."
   ;; deactivate any running transient map (transient-state)
   (setq overriding-terminal-local-map nil))
+
+(defun spacemacs//helm-prepare-display ()
+  "Prepare necessary settings to make Helm display properly."
+  (setq spacemacs-display-buffer-alist display-buffer-alist)
+  ;; the only buffer to display is Helm, nothing else we must set this
+  ;; otherwise Helm cannot reuse its own windows for copyinng/deleting
+  ;; etc... because of existing popwin buffers in the alist
+  (setq display-buffer-alist nil)
+  (setq helm--popwin-mode popwin-mode)
+  (when popwin-mode
+    (popwin-mode -1)))
+
+(defun spacemacs//helm-restore-display ()
+  ;; we must enable popwin-mode first then restore `display-buffer-alist'
+  ;; Otherwise, popwin keeps adding up its own buffers to
+  ;; `display-buffer-alist' and could slow down Emacs as the list grows
+  (when helm--popwin-mode
+    (popwin-mode))
+  (setq display-buffer-alist spacemacs-display-buffer-alist))
 
 
 ;; REPLs integration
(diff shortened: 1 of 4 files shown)


TheBB:

I added a spacemacs variable name prefix in https://github.com/syl20bnr/spacemacs/commit/989dd3d774673a9fe06fb0edb2f9185d396baf94

PR #6747, opened 2016-08-02
Add permalinks to the doc headings

nashamri:
This PR is to add permalinks to the documentation headings and sections to make it easy to share links and to fix this website issue https://github.com/syl20bnr/spacemacs.org/issues/13.


TheBB:
This won't work because the HTML files aren't all at the same depth. :-/

nashamri:
I can't think of any way to overcome this other than just using the full path to that javascript file. It's already on http://develop.spacemacs.org/js/permalinks.js

Should I just change it to be http://www.spacemacs.org/js/permalinks.js (the stable site)?

nashamri:
I added the absolute path to that `js` file. It's not on the stable branch yet but on the `develop` branch of the website. If this is okay, we can move that file to the stable branch.

JAremko:
>can't think of any way to overcome this other than just using the full path to that javascript file. 

@nashamri  You can use the same trick as with css

https://github.com/nashamri/spacemacs/blob/21cfdf8b040c73fe01ede2dfaec60eb826d472a2/core/core-documentation.el#L152

syl20bnr:
Thank you ! 👍 
For now the absolute path will do it. We can go the injection route later (or when it breaks 😁).

PR #6764, opened 2016-08-05
Reduce yas loading warning

bixuanzju:
[Here](https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Bcompletion/auto-completion/packages.el#L192) it sets `spacemacs--auto-completion-dir` to be the `local` subdir, which we don't have, thus the warning.


TheBB:

PR #6770, opened 2016-08-05
activate ycmd-mode in c files as well

deb0ch:
ycmd-mode was only activated when entering c++-mode, making syntax
checking unavailable in c-mode as flycheck-ycmd-mode is hooked to
ycmd-mode.


TheBB:

PR #6771, opened 2016-08-05
Fix jumping backwards after calling spacemacs/evil-smart-goto-definition

### LINKED ISSUE #4942: evil-jump-backward not working as expected after `SPC s h d`
After doing a `symbol highlight` -> `go to definition`, `evil-jump-backward` does not jump back to the place before the jump to the definition. I guess the jump is not stored in the jump list, or no mark is set... I don't know the inner workings. Is this fixable?

Stebalien:
Ensure that the "previous position" mark is set correctly. Credit goes to @eltechnic0 to figuring this out.

Fixes #4942

### DIFF
--- a/layers/+distributions/spacemacs-bootstrap/funcs.el
+++ b/layers/+distributions/spacemacs-bootstrap/funcs.el
@@ -66,6 +66,14 @@
         (call-interactively binding)
       (evil-goto-definition))))
 
+;; Set the `:jump' property manually instead of just using `evil-define-motion'
+;; in an `eval-after-load' macro invocation because doing that prevents
+;; `describe-function' from correctly finding the source.
+;;
+;; See discussion on https://github.com/syl20bnr/spacemacs/pull/6771
+(with-eval-after-load 'evil
+  (evil-set-command-property 'spacemacs/evil-smart-goto-definition :jump t))
+
 (defun spacemacs//set-evil-shift-width ()
   "Set the value of `evil-shift-width' based on the indentation settings of the
 current major mode."


bmag:
1. makes more sense if `evil-set-command-property` comes right after the definition of `spacemacs/evil-smart-goto-definition`, instead of after the key binding.
2. perhaps use `evil-define-motion` to define `spacemacs/evil-smart-goto-definition`? You can see the source of `evil-goto-definition` for an example.

nixmaniack:
I didn't know the solution was out there in issues. This has bugged me a lot and I finally accepted this as behaviour from Vim! Thanks for the fix @eltechnic0 @Stebalien.

Stebalien:
@bmag, I've gone with option 2. Unfortunately, this means that `describe-function` can't find the definition of `spacemacs/evil-smart-goto-defintion` (because I have to put it in a `with-eval-after-load` block). Is there any way to fix this?

bmag:
Oh, I didn't realize that would happen. I don't know how to fix it.

Stebalien:
@bmag, I've gone back to using `evil-set-command-property` (but put it in an `eval-after-load` after the function definition) as I use `SPC hdk`/`SPC hdf` all the time and hate it when it links me to `init.el`.

bmag:
I'm ok with the current solution :+1:

TheBB:
Me too. Note that #6815 proposes a unified `SPC mgg` function, and since `spacemacs/evil-smart-goto-definition` merely runs `SPC mgg` I will probably move this over to the new function once that is merged. In fact it probably makes sense just to bind `gd` and `SPC mgg` to the same function in that case.

TheBB:
Or maybe not, if `evil-goto-definition` is useful. Anyway… thanks! Cherry-picked in develop. You can safely delete your branch.

PR #6774, opened 2016-08-05
go: replace deprecated oracle with guru

### LINKED ISSUE #6772: go: replace deprecated oracle with guru
[go layer doc: oracle section](https://github.com/syl20bnr/spacemacs/tree/master/layers/%2Blang/go#go-oracle)

[go team's emacs guru integration](https://github.com/golang/tools/blob/master/cmd/guru/go-guru.el)

[spacemacs' current oracle bindings](https://github.com/syl20bnr/spacemacs/blob/master/layers/%2Blang/go/extensions.el#L35), using go team's oracle integration. These are misfiled under `rename`. Hopefully folks don't have too much muscle memory built up and we can map them elsewhere.

[go guru design doc](https://docs.google.com/document/d/1_Y9xCEMj5S-7rv2ooHpZNH15JgRT5iM742gJkw5LtmQ/edit)

edrex:
updates go-oracle -> go-guru. It's the same command, probably renamed because of recent legal disputes..

Outstanding issues with this patch:

 - [x] change prefix for guru commands from `mr` to `mf` to separate from rename command
 - [ ] since rename prefix now only has a single item, surface that item as `mr`
 - [ ] Reuse `go-guru-map` rather than maintaining a separate list of bindings.
 - [ ] Calling `go-guru-definition` results in an error:
        `cdr: Wrong type argument: listp, "definition"`
        This might be due to upstream, further investigation is needed.
 - [ ] On OSX, `GOPATH` needs to be set via launchd or guru initialization is skipped, see http://stackoverflow.com/questions/135688/setting-environment-variables-in-os-x. Document this.
 - [ ] Can we do something smarter about providing an entry point package for some `guru` subcommands (`callers`, ...)? Maybe we can figure out the package name at the top level of the project and provide that as a default value in the prompt? What about persisting it across emacs sessions per-project?

Fixes #6772.

### DIFF
--- a/layers/+lang/go/packages.el
+++ b/layers/+lang/go/packages.el
@@ -10,7 +10,7 @@
         helm-gtags
         go-eldoc
         go-mode
-        (go-oracle :location site)
+        (go-guru :location site)
         (go-rename :location local)
         ))
 
@@ -113,31 +113,32 @@
 (defun go/init-go-eldoc()
   (add-hook 'go-mode-hook 'go-eldoc-setup))
 
-(defun go/init-go-oracle()
+(defun go/init-go-guru()
   (let ((go-path (getenv "GOPATH")))
     (if (not go-path)
         (spacemacs-buffer/warning
-         "GOPATH variable not found, go-oracle configuration skipped.")
+         "GOPATH variable not found, go-guru configuration skipped.")
       (when (load-gopath-file
-             go-path "/src/golang.org/x/tools/cmd/oracle/oracle.el")
-        (spacemacs/declare-prefix-for-mode 'go-mode "mr" "rename")
+             go-path "/src/golang.org/x/tools/cmd/guru/go-guru.el")
+        (spacemacs/declare-prefix-for-mode 'go-mode "mf" "guru")
         (spacemacs/set-leader-keys-for-major-mode 'go-mode
-          "ro" 'go-oracle-set-scope
-          "r<" 'go-oracle-callers
-          "r>" 'go-oracle-callees
-          "rc" 'go-oracle-peers
-          "rd" 'go-oracle-definition
-          "rf" 'go-oracle-freevars
-          "rg" 'go-oracle-callgraph
-          "ri" 'go-oracle-implements
-          "rp" 'go-oracle-pointsto
-          "rr" 'go-oracle-referrers
-          "rs" 'go-oracle-callstack
-          "rt" 'go-oracle-describe)))))
+          "fd" 'go-guru-describe
+          "ff" 'go-guru-freevars
+          "fi" 'go-guru-implements
+          "fc" 'go-guru-peers
+          "fr" 'go-guru-referrers
+          "fj" 'go-guru-definition
+          "fp" 'go-guru-pointsto
+          "fs" 'go-guru-callstack
+          "fe" 'go-guru-whicherrs
+          "f<" 'go-guru-callers
+          "f>" 'go-guru-callees
+          "fo" 'go-guru-set-scope)))))
 
 (defun go/init-go-rename()
   (use-package go-rename
     :init
+    (spacemacs/declare-prefix-for-mode 'go-mode "mr" "rename")
     (spacemacs/set-leader-keys-for-major-mode 'go-mode "rn" 'go-rename)))
 
 (defun go/init-flycheck-gometalinter()
--- a/layers/+lang/go/README.org
+++ b/layers/+lang/go/README.org
@@ -13,7 +13,7 @@
    - [[#tests][Tests]]
  - [[#working-with-go][Working with Go]]
    - [[#go-commands-start-with-m][Go commands (start with =m=):]]
-   - [[#go-oracle][Go Oracle]]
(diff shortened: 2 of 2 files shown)


TheBB:
@bogdanteleaga does this look ok to you?

TheBB:
Maybe you can have a look at the conventions document to make sure that the bindings make sense. `SPC m r` is for refactoring and there is no `SPC m f`. Some of these bindings look like they should be on `SPC m g` (like jump to definition).

layers/+lang/go/packages.el
```diff
@@ -113,31 +113,32 @@
...
+          "fs" 'go-guru-callstack
+          "fe" 'go-guru-whicherrs
+          "f<" 'go-guru-callers
+          "f>" 'go-guru-callees
+          "fo" 'go-guru-set-scope)))))
 
 (defun go/init-go-rename()
   (use-package go-rename
     :init
+    (spacemacs/declare-prefix-for-mode 'go-mode "mr" "rename")
```
robbyoconnor: Except you nuked all keybindings for this prefix.

edrex: oh you're right. I thought there were some "rename" bindings. what does the go-rename package do?

robbyoconnor: How did you miss the fact that no keybindings with that prefix existed?

edrex: I just quickly fixed this because I became blocked while doing some go work by an issue with oracle.

It looks like the main entry point for go rename is actually mapped to `mrr`.

I can just put the bindings back under `mr` if folks are happy with them.

robbyoconnor: hrm...I think I missed that.

edrex:
Anything I can do to move this forward? I can see the key remapping being possibly controversial; I'm happy to change it back.

Potential go-coding reviewers? @bogdanteleaga 

Thanks all

bogdanteleaga:
Hey, sorry, I saw the email initially and completely forgot about it.

Thanks for the contribution, I was not aware they were going to rename oracle, I could never use it that much because it didn't work properly for the projects I was working on. That being said, I always found the bindings being under `SPC m r` to be rather weird.

As for the point @TheBB made, some of them look like they could work as a replacement for go to defintion (go-guru-definition) or documentation (go-guru-describe). But I believe we should just let the whole thing sit under its own keybinding, mostly because guru/oracle doesn't work without setting a scope and that might be confusing to new users. Most of the functionality it provides is rather unique and doesn't really fit under the rest of the conventions we use in spacemacs as far as I am aware.

What I'd like is at least some warning sign(probably in the documentation), that oracle is deprecated and we should use guru now. Even though oracle users are most likely aware, I don't think it would hurt to have it around for a while.

tl;dr :+1:, but add an explanation in the docs.

edrex:
Ok, I'll update the docs warning users about oracle deprecation. Thanks for taking a look!

TheBB:
Ping! I was about to merge this but looks like it's not updated yet.

edrex:
@TheBB yeah I still need to do another pass, will update when ready

enzochiau:
Is it possible to reuse the `go-guru-map` instead of maintain another set in the `go/init-go-guru`?

edrex:
@enzochiau I'd like that. Example needed. Updated the issue text to add that as a todo.

layers/+lang/go/packages.el
```diff
@@ -113,31 +113,32 @@
...
   (let ((go-path (getenv "GOPATH")))
     (if (not go-path)
         (spacemacs-buffer/warning
-         "GOPATH variable not found, go-oracle configuration skipped.")
+         "GOPATH variable not found, go-guru configuration skipped.")
       (when (load-gopath-file
-             go-path "/src/golang.org/x/tools/cmd/oracle/oracle.el")
-        (spacemacs/declare-prefix-for-mode 'go-mode "mr" "rename")
+             go-path "/src/golang.org/x/tools/cmd/guru/go-guru.el")
+        (spacemacs/declare-prefix-for-mode 'go-mode "mf" "guru")
```
sunisdown: why use `mf`? what about `mj`？ ; j for jump(definition)

layers/+lang/go/packages.el
mirrorwis: any update?

jredville:
I'm new to emacs, but I use `go` and `go-guru` and want to get involved. Is there anything I can do to help out here?

TheBB:
I was waiting for this.

> Ok, I'll update the docs warning users about oracle deprecation.

If the PR works as is and someone submits a new one with documentation update on top, I can merge.

jredville:
I'll take a look this weekend

jredville:
@TheBB new one submitted building on this one

JAremko:
@TheBB  It  seems that golang.org/x/tools/cmd/oracle is gone now. So this PR should have high priority.

TheBB:
Thanks! Cherry-picked in develop as part of #7242. You can safely delete your branch..

PR #6780, opened 2016-08-06
Fix Agda layer for users without 'agda-mode' executable

FreeSalad:
Adds a layer variable (`agda-mode-path`) used to locate agda-mode, instead of always using the helper executable `agda-mode` (this functionality is still enabled by default with `'use-helper`).

### DIFF
--- a/layers/+lang/agda/README.org
+++ b/layers/+lang/agda/README.org
@@ -34,6 +34,11 @@ Then check that =agda= is available on your =$PATH= and seen by Emacs. For
 information about setting up =$PATH=, check out the corresponding section in the
 FAQ (~SPC h SPC $PATH RET~).
 
+By default the =agda-mode= executable bundled with most agda installations
+is used to locate the agda-mode package. If you don't have it and want to use
+a local agda-mode package, you can customize the layer variable =agda-mode-path=
+to your needs.
+
 * Key bindings
 The key bindings of this layer don't follow the Spacemacs conventions,
 we opted to a simple transcription of stock Agda mode key bindings to
--- a/layers/+lang/agda/config.el
+++ b/layers/+lang/agda/config.el
@@ -1,0 +1,15 @@
+;;; config.el --- Agda2 Layer config File for Spacemacs
+;;
+;; Copyright (c) 2012-2016 Sylvain Benner & Contributors
+;;
+;; Author: FreeSalad <freesalad@noreply.git>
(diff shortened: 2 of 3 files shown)


TheBB:
But what this does is it excludes all the agda-mode setup entirely if you turn off `agda-use-agda-mode`. From your description what the variable should do is just change the location of the package. Most of the `use-package` call should happen regardless.

FreeSalad:
But `(or (not agda-use-agda-mode) (...))` evaluates to `t` if `agda-use-agda-mode` is set to `nil`, right?
So either you've turned it off and `use-package` is evaluated, either you don't and it'll check if `agda-mode` exists. Depending on that it'll eval `use-package

I have it turned off on my machine and agda-mode works as expected.

TheBB:
Right, but then I'm not sure why you're still [running this](https://github.com/syl20bnr/spacemacs/pull/6780/files#diff-cc32d29b099feca4c8830cc74afe811aR19) if agda-mode is possibly not available at this point.

FreeSalad:
There's some kind of confusion here : installing Agda also installs the agda-mode package along with an `agda-mode` executable which is in charge of locating the package and compiling it, among other things.

The original Agda layer relied only on the executable to find the package, but some users could encounter an issue on installations without the executable (eg on Gentoo where the installing process is emulated and stripped of some useless things).

You don't actually need the executable if everything's already in the right place, that's what that variable enables you to do. It tells the layer to trust the user with its installation.

TheBB:
Right. Sorry. Your code just confused me.

I'd rather we solved this issue similar to how [mu4e](https://github.com/syl20bnr/spacemacs/blob/develop/layers/+email/mu4e/config.el#L12-L13) does it, for consistency's sake.

TheBB:
In agda the variable can for example take a symbol value which indicates to use the executable to locate it.

FreeSalad:
So, use `'use-helper`, `nil` or a string indicating where the package is located?

TheBB:
Yeah, for example.

FreeSalad:
Here, I've replaced `agda-use-agda-mode` with `agda-mode-path`.

TheBB:
Looks good! Thanks! Cherry-picked in develop. You can safely delete your branch.

I made some changes to the README and the indentation in https://github.com/syl20bnr/spacemacs/commit/dbbdca7d5d5e8079b96f7f26909b75c13af6016c.

PR #6783, opened 2016-08-07
New add cargo package to rust layer

kwrooijen:
Added cargo package for the Rust layer. 
Removed all the spacemacs defined cargo functions because they are all defined in the cargo package.
Added some extra cargo keybindings.


TheBB:
Can you (a) sort the bindings and (b) move tests to `SPC m t`?

kwrooijen:
Made the changes, if approved I can rebase it into a single commit

TheBB:

Had to make some changes in https://github.com/syl20bnr/spacemacs/commit/2651f67c4873ea000a6c08e6001fd82a50860df8. Without a `rust/init-cargo` function the package will not be installed!

Also noted that you removed the fmt function but not the binding to it, so I substituted the corresponding cargo-process function.

PR #1644, opened 2015-05-19
[Fix #1512] Don't enable smartparens when expanding snippets

tuhdo:
Otherwise the snippets will be messed up when expanding with
hippie-expand.

PR #6784, opened 2016-08-07
Fix smartparens remains disabled after a nested snippet expansion

### LINKED ISSUE #1512: Yasnippet expansion, hippie-expand and smartparens
Something strange is happening when expanding certain snippets using `hippie-expand` while `smartparens` is activated.

I can reproduce the issue on `develop` with a fresh `.spacemacs` template with only the `auto-completion` layer added:

1. clone spacemacs
2. initialize spacemacs, add `auto-completion` layer
3. start spacemacs
4. open a file `test.c`
5. type `main`
6. invoke `yas-expand` 
7. observe the correct result
8. delete buffer contents and type `main`
9. invoke `hippie-expand`
10. observe incorrect result (extra closing paren)

Sample of incorrect expansion:
```c
int main()int argc, char *argv[])
{
  
  return 0;
}
```

I have observed that toggling `smartparens` before and after like so:

```lisp
(add-to-list 'yas-before-expand-snippet-hook
             'spacemacs/toggle-smartparens)

(add-to-list 'yas-after-exit-snippet-hook
             'spacemacs/toggle-smartparens)
```

Resolves the issue, but I haven't been able to track down exactly why.

madand:
Current implementation, #1644, not handles the nested snippet
expansion properly, so you end up with smarparens disabled. Very
annoying.

I improved the implementation to remember the current state of
smartaperns before a top-level snippet expansion and preserve on any
nested snippet expansions.

Also, some housekeeping of the related code:
* Move hook handlers to the `funcs.el`.
* Add comments about yasnippet hooks subtleties relevant to the issue.

Fixes #1512

### DIFF
--- a/layers/+completion/auto-completion/funcs.el
+++ b/layers/+completion/auto-completion/funcs.el
@@ -220,3 +220,39 @@
   (interactive)
   (call-interactively 'aya-expand)
   (unless holy-mode (evil-insert-state)))
+
+
+;; Yasnippet and Smartparens
+
+;; If enabled, smartparens will mess snippets expanded by `hippie-expand`.
+;; We want to temporarily disable Smartparens during the snippet expansion and
+;; switch it back to the initial state when done.
+;;
+;; However, there is an asymmetry in Yasnippet's hooks:
+;; * `yas-before-expand-snippet-hook' is called for all snippet expansions,
+;; including the nested ones.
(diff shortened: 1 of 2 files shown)


TheBB:
Thanks, good detective work! Cherry-picked in develop. You can safely delete your branch.

PR #6797, opened 2016-08-11
WIP Added a geben configuration layer

rubberydub:
A small configuration layer providing key bindings for
gebben-on-emacs. The leader is SPC G.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,8 @@ server/
 *.db
 python-*-docs-html
 /.cask/
+geben/
+!layers/+tools/geben/
 
 # Private directory
 private/
--- a/layers/+tools/geben/packages.el
+++ b/layers/+tools/geben/packages.el
@@ -1,0 +1,72 @@
+;;; packages.el --- geben layer packages file for Spacemacs.
+;;
+;; Copyright (c) 2012-2016 Sylvain Benner & Contributors
+;;
+;; Author: Franz Luther Neulist Carroll <franzneulistcarroll@gmail.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+(defconst geben-packages '(geben))
+
+
+(defun geben/init-geben ()
+  (use-package geben
+    :init
+    (progn
+      (spacemacs/set-leader-keys
+        "Ga" 'geben
+        "Gr" 'geben-run
+        "Gx" 'geben-stop
+        "Ge" 'geben-end
+        "Gp" 'geben-proxy
+        "Gn" 'geben-proxy-end
+        "Gw" 'geben-where
+        "Gq" 'geben-quit-window
+        "Gv" 'geben-set-redirect
+        "Gh" 'geben-mode-help
+        "Gm" 'geben-toggle-pause-at-entry-line-flag
+        "Gj" 'geben-eval-expression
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -41,6 +41,8 @@ server/
 *.db
 python-*-docs-html
 /.cask/
+geben/
```
bmag: We probably want to set `geben-temporary-file-directory` to a location inside of the cache directory (`spacemacs-cache-directory`), instead of having it directly inside of `.emacs.d`.

In any case, we need to set `geben-temporary-file-directory` because the default value is hard-coded to use `"~/.emacs.d"` and not `user-emacs-directory`, limiting portability.

layers/+tools/geben/packages.el
```diff
@@ -0,0 +1,72 @@
...
+        "Gtm" 'geben-backtrace-mode
+        "Gth" 'geben-backtrace-mode-help
+        "Gtg" 'geben-backtrace-mode-goto
+        "Gth" 'geben-backtrace-mode-mouse-goto
+        "Gtc" 'geben-backtrace-mode-mode-context
+
+        "Gsi" 'geben-step-into
+        "Gso" 'geben-step-over
+        "Gsu" 'geben-step-out
+        "Gsa" 'geben-step-again
```
bmag: If the user is likely to use these commands several times consecutively, we should consider putting them in a transient state (see `spacemacs|define-transient-state`). For example, this way the user can type `SPC G s . o o o` (or similar) instead of `SPC g s o SPC g s o SPC g s o` to step-over multiple times.

I don't actually know anything about Geven, other from looking at the repo, so please excuse me if I've missed something relevant.

ksjogo:
Can I recommend also merging #6376 as the popwin config is quite useful.  I am developing upstream geben a bit and am a spacemacs user, so if any changes are recommended, I am happy to implement them.

rubberydub:
Thanks @bmag  and @ksjogo. 
@ksjogo I will definatly merge. I would like to implement the transient state, as I'm new to spacemacs and would like the experience.

rubberydub:
Closing as this is clearly too preemptive

rubberydub:
Closing as this is clearly too preemptive

ksjogo:
@rubberydub Did you continue working on the layer somehow? I am trying to push emacs xdebug support at the moment and it would be quite cool to also have a nice spacemacs integration.

rubberydub:
@ksjogo I've done [half the job](https://github.com/rubberydub/spacemacs-geben) and have forgotten about it until now. I'll form a new pull request soon (tomorrow). After I do, we can go over it?

rubberydub:
@ksjogo I took the popwin confs from your geben branch. I tried to merge in upstream develop, but there where other differences and I couldn't get it nice and clean. Is there another way take it to give you the credit?

rubberydub:
@ksjogo There is also the question of whether geben should be in the php layer or in a separate one. While it can be used for other langs, would anyone use it for anything besides php?

ksjogo:
Inside the php layer is probably better from a user experience, as the user most likely checks for/installs the php layer and not some arbitrarily named layer.
Just copy the popwin config, sure I will try it out locally.

dbrw:
Sorry if I am doing necrobump.
This issue is very interesting. It's helpful to have transient mode for geben either as separate layer or inside PHP layer. Currently I use geben exclusively with Xdebug.

PR #6799, opened 2016-08-11
Add protobuf layer

edrex:
Includes protobuf-mode and flycheck-protobuf packages

Current issues:
- [ ] When opening a `.proto` file on emacs 25, I see `File mode specification error: (void-variable company-backends-c-mode-common)` (but highlighting works)
- [ ] `flycheck-verify-setup` (`SPC e v`) reports `There are no syntax checkers for this buffer!` (need to add one, see https://github.com/edvorg/flycheck-protobuf)

### DIFF
--- a/layers/+lang/protobuf/packages.el
+++ b/layers/+lang/protobuf/packages.el
@@ -1,0 +1,21 @@
+(setq protobuf-packages
+      '(
+        flycheck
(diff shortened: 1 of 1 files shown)


robbyoconnor:
Add a README.

robbyoconnor:
You really should have used the function we have in spacemacs which creates everything for you from a template :)

edrex:
I never got back to this. Closing in preference to #7027.

PR #6802, opened 2016-08-12
Use spacemacs-core-directory const when building load-path

ivanbrennan:
As long as we've defined a spacemacs-core-directory, it seems odd not to use it when building the load-path. Is there any reason not to do so? Some small performance hit, maybe? The only thing I could think of was the fact the the characters don't line up as nicely. Then again, there's also fewer of them, so that's nice.


TheBB:

PR #6804, opened 2016-08-12
Optimize get-package-directory

ivanbrennan:
Saving a pkg-match regex to a temporary variable lets us avoid recomputing it on each iteration. I didn't benchmark the performance impact, but unless there's a downside to using `let*` vs `let`, it seems like a no-brainer.

Passing pkg-match to `directory-files` reduces the returned files to those that match the regex rolling our own reduction.


TheBB:
Solid stuff, thanks! Cherry-picked in develop. You can safely delete your branch.

PR #6805, opened 2016-08-12
Fix single missing character in documentation

splaspood:
Within the "Important Note" there is a single character (which written as wich) typo that this change corrects.


TheBB:

PR #6815, opened 2016-08-13
Refactor jump to definition

### LINKED ISSUE #6619: New gtags integration shadows key for find-symbol in elisp

TheBB:
According to discussion with @syl20bnr.

This commit defines:

- spacemacs-default-jump-handlers: a list of functions that can jump to
  definition in ALL modes.

- spacemacs-jump-handlers-MODE: a list of functions that can jump to
  definition in MODE.

- spacemacs-jump-handlers: a buffer-local list of functions that can
  jump to definition. This is made up of the values of the two previous
  variables whenever a given major mode is activated.

- spacemacs/jump-to-definition: a function that tries each function in
  spacemacs-jump-handlers in order, and stops when one of them takes us
  somewhere new.

- spacemacs|define-jump-handlers: a macro that
  * defines spacemacs-jump-handlers-MODE, possibly filled with initial
    functions
  * defines a function that is added to the hook of the given MODE
  * binds “SPC m g g” of that MODE to spacemacs/jump-to-definition

This is an attempt to harmonize all the different approaches to jumping.
Specifically,

- Existing intelligent jump packages that work for only a single mode
  should go to the beginning of spacemacs-jump-handlers-MODE. E.g.
  anaconda for python, ensime for scala, etc.

- Packages like gtags that work for several modes (but potentially not
  all) and which is dumber than the intelligent jumpers should go the
  the END of spacemacs-jump-handlers-MODE.

- Packages like dumb-jump that work for all modes should go to
  spacemacs-default-jump-handlers.

In all cases the order of the jump handlers in each list should be from
most to least intelligent.

Fixes #6619, and to some degree supersedes #6164 (CC @axyz re: dumb-jump)

### DIFF
--- a/core/core-jump.el
+++ b/core/core-jump.el
@@ -1,0 +1,61 @@
+;;; core-jump.el --- Spacemacs Core File
+;;
+;; Copyright (c) 2012-2016 Sylvain Benner & Contributors
+;;
+;; Author: Sylvain Benner <sylvain.benner@gmail.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+(defvar spacemacs-default-jump-handlers '()
+  "List of jump handlers available in every mode.")
+
+(defvar-local spacemacs-jump-handlers '()
+  "List of jump handlers local to this buffer.")
+
+(defmacro spacemacs|define-jump-handlers (mode &rest handlers)
+  "Defines jump handlers for the given MODE.
+This defines a variable `spacemacs-jump-handlers-MODE' to which
+handlers can be added, and a function added to MODE-hook which
+sets `spacemacs-jump-handlers' in buffers of that mode."
+  (let ((mode-hook (intern (format "%S-hook" mode)))
+        (func (intern (format "spacemacs//init-jump-handlers-%S" mode)))
+        (handlers-list (intern (format "spacemacs-jump-handlers-%S" mode))))
+    `(progn
+       (defvar ,handlers-list ',handlers
+         ,(format (concat "List of mode-specific jump handlers for %S. "
+                          "These take priority over those in "
+                          "`spacemacs-default-jump-handlers'.")
+                  mode))
+       (defun ,func ()
+         (setq spacemacs-jump-handlers
+               (append ,handlers-list
+                       spacemacs-default-jump-handlers)))
+       (add-hook ',mode-hook ',func)
+       (spacemacs/set-leader-keys-for-major-mode ',mode
+         "gg" 'spacemacs/jump-to-definition))))
+
+(defun spacemacs/jump-to-definition ()
+  (interactive)
+  (catch 'done
+    (let ((old-buffer (current-buffer))
+          (old-point (point)))
+      (dolist (handler spacemacs-jump-handlers)
+        (ignore-errors
+          (call-interactively handler))
+        (unless (and (eq old-point (point))
+                     (equal old-buffer (current-buffer)))
+          (throw 'done t))))
+    (message "No jump handler was able to find this symbol.")))
+
+;; Set the `:jump' property manually instead of just using `evil-define-motion'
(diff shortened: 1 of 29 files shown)


TheBB:
Note, with this is it potentially possible to make gtags work automatically in all modes that use spacemacs|define-jump-handlers without listing the package everywhere. I haven't done that here.

ksjogo:
Another option would be to use the new xref functions from 25 and create xref backends for the jumpers.
This route would introduce less spacemacs specific functionality and let emacs do the picking of the backends.
Based on the xref backends we could introduce just one common frontend in spacemacs e.g. one helm/ivy.

TheBB:
That looks promising, but we can't drop support for 24 for a good while yet. I assume we will also need to wait for upstream jumpers to integrate with xref.

layers/+lang/c-c++/packages.el
```diff
@@ -122,9 +125,11 @@
 
 (defun c-c++/post-init-ycmd ()
   (add-hook 'c++-mode-hook 'ycmd-mode)
-  (spacemacs/set-leader-keys-for-major-mode 'c++-mode
-    "gg" 'ycmd-goto
-    "gG" 'ycmd-goto-imprecise))
+  (add-hook 'spacemacs-jump-handlers-c++-mode 'ycmd-goto)
+  (add-hook 'spacemacs-jump-handlers-c-mode 'ycmd-goto)
+  (dolist (mode '(c++-mode c-mode))
+    (spacemacs/set-leader-keys-for-major-mode 'c++-mode
```
bmag: Should this not be a `dolist`, or should it be `mode` instead of `'c++-mode`?

core/core-jump.el
```diff
@@ -0,0 +1,54 @@
...
+                       spacemacs-default-jump-handlers)))
+       (add-hook ',mode-hook ',func)
+       (spacemacs/set-leader-keys-for-major-mode ',mode
+         "gg" 'spacemacs/jump-to-definition))))
+
+(defun spacemacs/jump-to-definition ()
+  (interactive)
+  (catch 'done
+    (dolist (handler spacemacs-jump-handlers)
+      (let ((old-buffer (current-buffer))
```
bmag: Should we move the `let` outside of `dolist`, because `old-buffer` and `old-point` aren't gonna change inside the loop?

TheBB: Yeah, that seems reasonable.

layers/+lang/c-c++/packages.el
TheBB: It should be a dolist, I just messed up the mode.

TheBB:
I have been using this for a few days now and I'm quite happy with it. If there are no complaints I might just merge it soon.

TheBB:
Merged.

PR #6825, opened 2016-08-14
Improve keybinding for beginning/end of line in pdf-tools

Andre0991:
`image-bol` and `image-eol` are more appropriate than`scroll-left` and `scroll-right`.


TheBB:

Issue #4453, opened 2016-01-06
Emacs hangs at contacting host


danprince:
Trying to install spacemacs on Ubuntu 15.04.

I've moved the spacemacs files to `~/.emacs.d/`, but when I start emacs I get a flash of the initial emacs window, then a black screen with

> Contacting host: melpa.org:443

Once it opened with `orgmode.org:80` instead, but I haven't been able to recreate that.
- Emacs works fine without the init file (`emacs -q`).
- Using emacs 24.4.1
- Using spacemacs 0.105.0
- `emacs --debug-init` produces no output
- I can `curl https://melpa.org` so not a network restriction

I've tried a fresh install of both emacs and spacemacs as well as using `emacs`, `emacs24` and `emacs24-x` (but I'm not exactly sure what the differences are).

I've been writing Clojure for the last few months and I'm keen to beat some of the struggles I've been having with Lisp + Vim, but after five or six attempts, I've given in and resorted to putting in an issue. Any ideas?

syl20bnr:
Can you try launching emacs with `emacs --insecure` to not use HTTPS to contact MELPA ?

danprince:
Aha, that's how created the `orgmode.org:80` hang. It starts with `melpa.org:80` then switches to orgmode and hangs there instead.

danprince:
@syl20bnr Is it possible to install the packages manually from the repository? If so, will that prevent emacs from making that initial network request?

greasynose:
@danprince  I'm also meet this problem, do you solve it ?

nickbdyer:
Did anyone find a solution to this?

danprince:
I haven't solved this problem yet. I just tried building emacs 24.4 from source and I still have a similar problem with Spacemacs.

Running `emacs` and `emacs-24.4` causes it to hang at:

> Opening TLS connection to `melpa.org`...done

Running `emacs --insecure` causes it to hang at:

> Contacting host: melpa.org

liangwang:
The same issue as OP reported, and the issue is only observed on Linux, no problems on Windows and MacOS.

Using latest spacemacs (v0.105.14), and emacs 24.5 for all three platforms.

wende:
I've got the same problem on OSX with Emacs v0.105.19

Running from finder instead of terminal fixed the issue

myme5261314:
[...]

PR #6828, opened 2016-08-15
Add package archive fallback mechanism

### LINKED ISSUE #6836: Cannot start spacemacs if melpa.org is down
That's all, really.

Running on OS X. Not a fresh install, although I tried a fresh install with the same results.

d12frosted:
Seeing all these issues with package installation due to MELPA availability problems I decided to improve status check for all elpa archives and error about archive being unavailable instead of showing errors like 'package xxx is not available' in the future.

Also I removed timeout heuristic as it's not 100% accurate. I don't see any reason for it to be here once `configuration-layer-check-archive-status` is used. 

Note that `configuration-layer-check-archive-status` is a little bit beefed (more than we need at this point), but it might become handy in future. 

Oh, and I am not sure about the error message though. @syl20bnr, you might want to improve it. 😸 

P. S. Hope I didn't miss anything. 

P. P. S. Somewhat fixes #6836. Another related issue is #4453.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -206,29 +206,17 @@ is not set for the given SLOT."
 
 (defvar configuration-layer--elpa-archives
   '(("melpa" . "melpa.org/packages/")
-    ("org"   . "orgmode.org/elpa/")
-    ("gnu"   . "elpa.gnu.org/packages/"))
-  "List of ELPA archives required by Spacemacs.")
-
-(defvar configuration-layer--elpa-archives-status
-  '(("melpa" . unknown)
-    ("org" . unknown)
-    ("gnu" . unknown))
-  "Status of ELPA archives.")
-
-(defvar configuration-layer-fallback-package-archives
-  '(("melpa" . "raw.githubusercontent.com/syl20bnr/spacemacs-elpa-mirror/master/melpa/")
+    ("melpa" . "raw.githubusercontent.com/syl20bnr/spacemacs-elpa-mirror/master/melpa/")
     ("melpa" . "elpa.zilongshanren.com/melpa/")
 
-    ("org" . "raw.githubusercontent.com/syl20bnr/spacemacs-elpa-mirror/master/org/")
-    ("org" . "elpa.zilongshanren.com/org/")
+    ("org"   . "orgmode.org/elpa/")
+    ("org"   . "raw.githubusercontent.com/syl20bnr/spacemacs-elpa-mirror/master/org/")
+    ("org"   . "elpa.zilongshanren.com/org/")
 
-    ("gnu" . "raw.githubusercontent.com/syl20bnr/spacemacs-elpa-mirror/master/gnu/")
-    ("gnu" . "elpa.zilongshanren.com/gnu/"))
-  "URL of fallback package archive.
-
-This archive is used when one of archives is down for any
-reason.")
+    ("gnu"   . "elpa.gnu.org/packages/")
+    ("gnu"   . "raw.githubusercontent.com/syl20bnr/spacemacs-elpa-mirror/master/gnu/")
+    ("gnu"   . "elpa.zilongshanren.com/gnu/"))
+  "List of ELPA archives required by Spacemacs.")
 
 (defvar configuration-layer-exclude-all-layers nil
   "If non nil then only the distribution layer is loaded.")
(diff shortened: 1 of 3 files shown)


core/core-configuration-layer.el
```diff
@@ -1953,6 +1939,26 @@ FILE-TO-LOAD is an explicit file to load after the installation."
...
+is `unknown'. Check is forced when RECHECK is non-nil."
+  (let* ((obj (assoc archive configuration-layer--elpa-archives-status))
+         (state (cdr obj))
+         (url (cdr (assoc archive package-archives))))
+    (message "archive %s state %s url %s" archive state url)
+    (when (and (or (eq state 'unknown) recheck)
+               url)
+      (condition-case nil
+          (if (url-http-file-exists-p url)
+              (setq state 'avaialble)
```
TheBB: available

core/core-configuration-layer.el
```diff
@@ -1953,6 +1939,26 @@ FILE-TO-LOAD is an explicit file to load after the installation."
...
+            (setq state 'unavailable))
+        ((error) (setq state 'unavailable)))
+      (setq configuration-layer--elpa-archives-status (delq obj configuration-layer--elpa-archives-status))
+      (add-to-list 'configuration-layer--elpa-archives-status (cons archive state)))
```
TheBB: You can do

    (setcdr (assq archive configuration-layer--elpa-archives-status) state)

TheBB:
If `dotspacemacs-elpa-timeout` isn't needed any more you should remove it (and the tests for it).

core/core-configuration-layer.el
d12frosted: My typing skills are weak 👀 Thanks for noticing this typo!

core/core-configuration-layer.el
d12frosted: Wow. Didn't know about `setcdr`. Thanks. Though in this case you have to use `assoc` instead of `assq` as we are dealing with strings.

d12frosted:
I have removed `dotspacemacs-elpa-timeout` as well as related test. Also I have updated error message in order to make it more clear.

d12frosted:
Just pushed another commit that adds fallback functionality when one of the archives is down. You can test it by returning `'unavailable` in `configuration-layer-check-archive-status` for any archive of your liking. For example,

```
(when (string-equal archive "melpa")
      (setq state 'unavailable))
```

So when any archive is not available, `configuration-layer/retrieve-package-archives` will fallback to `configuration-layer-fallback-package-archive` for unavailable archive. 

As fallback I am using snapshot provided by @zilongshanren. Thanks to https://github.com/melpa/melpa/issues/4156 I've found this nice 'mirror'. The good thing is that MELPA is going to create official mirror that people could use when MELPA is down. 

P. S. I've made two commits for now - gonna squash them later on demand. Just wanna keep first commit for now.

core/core-configuration-layer.el
```diff
@@ -201,6 +198,19 @@ LAYER has to be installed for this method to work properly."
     ("gnu"   . "elpa.gnu.org/packages/"))
   "List of ELPA archives required by Spacemacs.")
 
+(defvar configuration-layer--elpa-archives-status
+  '(("melpa" . unknown)
+    ("org" . unknown)
+    ("gnu" . unknown))
+  "Status of ELPA archives.")
+
+(defvar configuration-layer-fallback-package-archive
```
d12frosted: I am planning to convert it list of fallback archives.

core/core-configuration-layer.el
```diff
@@ -329,24 +337,16 @@ refreshed during the current session."
...
-                          (url-retrieve-synchronously (cdr archive))
-                        ('error
-                         (display-warning 'spacemacs
-                          (format
-                           "\nError while contacting %s repository!"
-                           (car archive)) :warning)
-                         'error))))
+        (if (eq 'unavailable
+                (configuration-layer-check-archive-status (car archive)))
+            (setcdr
```
d12frosted: Here I am going to check fallback archives as well and if all of them unavailable - error like in first commit.

d12frosted:
Pushed another commit. 

Instead of using one fallback url, use package-archives compatible list of fallback urls. If currently set url is not reachable, then try to connect to one of fallback urls. If all urls are unavailable - show an error with some hints and explanations.

Again, you might want to change error message. But I tried to make it as useful as possible. 😸

d12frosted:
And a good thing, now https://github.com/syl20bnr/spacemacs-elpa is used as primary fallback archive 😸

core/core-configuration-layer.el
```diff
@@ -201,6 +198,26 @@ LAYER has to be installed for this method to work properly."
     ("gnu"   . "elpa.gnu.org/packages/"))
   "List of ELPA archives required by Spacemacs.")
 
+(defvar configuration-layer--elpa-archives-status
+  '(("melpa" . unknown)
+    ("org" . unknown)
+    ("gnu" . unknown))
+  "Status of ELPA archives.")
+
+(defvar configuration-layer-fallback-package-archives
```
d12frosted: I wonder if we really need this. I mean, fallback archives can be added to the end of `configuration-layer--elpa-archives`, so instead of switching to values from `configuration-layer-fallback-package-archives` we can just remove first associated url and search for another. Once no urls are available for specific archive - error!

core/core-configuration-layer.el
```diff
@@ -295,41 +299,61 @@ cache folder.")
...
+(defun configuration-layer//check-archive-status (archive)
+  "Check ARCHIVE status.
+
+Available return values are available and unavailable."
+  (let ((url (format "%sarchive-contents"
+                     (cdr (assoc archive package-archives))))
+        (state 'unavailable))
+    (condition-case nil
+        (when (or (and (string-prefix-p "/" url)
+                       (file-readable-p url))
```
d12frosted: Not sure that this is the best way of checking that archive url is accessible. `file-accessible-directory-p`? I haven't tested situations when `archive-contents` file is readable, but all other files are not readable by current user.

d12frosted:
Ok, I've made several changes. I simplified fallback mechanism, removed two redundant variables and improved error message. Oh and also I made it possible to use local package archive urls.

d12frosted:
I spammed pretty much in this PR with commits. I'll squash them once @syl20bnr give the green light 😸 For now I wish to preserve history. 😸

ninrod:
This should close #7103 and provide a way for users behind corporate firewalls to use spacemacs in all it's glory.

d12frosted:
@ninrod is GitHub available for you behind your firewall?

ninrod:
@d12frosted yes it is. We have a severely slowed down connection pipe to and from it though. So the local clone is like heaven.

But really, the local melpa solution is so awesome that I'm using it in home, even without a firewall. 
With a single git clone you have the whole melpa, elpa and what not in a local folder that you control. No connection hiccups, I don't have to worry about melpa going down, reconfigurations are like instantaneous, I can tag a specific snapshot that I know for a fact that works with my config. What more can I ask? That should be the default. So robust, efficient, fast, configurable. Awesome.

d12frosted:
@ninrod I asked about GitHub just because some of packages used by Spacemacs are downloaded using `quelpa` from GitHub directly. This fallback mechanism doesn't cover such packages. E. g. when GitHub is down (?!) it's impossible to start Spacemacs from scratch even if you have local copies of elpa package archives (namely GNU ELPA, MELPA, Stable MELPA and Org-mode package archive). I know that I already said this in other thread, but I just want to make it clear for others as well. 

> reconfigurations are like instantaneous

Indeed. Also it makes bootstrap testing much easier. 

> I can tag a specific snapshot that I know for a fact that works with my config

And this as well. Might be very useful for corporations, where they want to use specific stable version of Spacemacs with specific package archives snapshot (so no upstream changes break flow). 

> That should be the default. So robust, efficient, fast, configurable. Awesome.

Glad you like it 💃

ninrod:
@d12frosted I should mention that I have the option of mirroring specific github repos inside an internal git server behind corp firewall. Out of curiosity, am I able to change those quelpa URLs to match my specific internal mirrors?

ninrod:
let's do this!

core/core-configuration-layer.el
syl20bnr: @d12frosted Where can we find the documentation about `file-readable-p` capable of checking URLs except in the C source ? This function does not block for a long time when the URL is not reachable ? What kind of tests did you do to check that `file-readable-p` is not blocking for a long time ?

syl20bnr: > What kind of tests did you do to check that file-readable-p is not blocking for a long time ?

I mean, did you test regular timeout method and `file-readable-p` with the same input ?

Issue #6819, opened 2016-08-14
need a way to disable ggtags


emacs18:
I typed "SPC h I" as instructed within my emacs, typed the message below, then hit C-c C-c.
However nothing seems to happen other than an *eww* buffer popping up with
a long URL starting with https://github.com/.

That is why I am creating this issue using the github web page.

#### Description :octocat:
On 2016/07/17, spacemacs/ggtags-mode-enable function was added on devel branch
to bunch of major mode hooks to enable ggtags in many modes. This change is
causing several problems.

One problem is that gtags simply is not robust enough to be used this way under
certain conditions such as on Red Hat 6. I often have to kill these processes
manually on Red Hat 6, because they eat up 100 percent of one CPU doing
who-knows-what for hours. I tried updating to newer GNU global few times, but
that did not help.

Another problem is that spacemacs fails to start if I remove certain ELPA
packages so that these are installed during emacs start up. Following is an
example stack trace where installing one of my own packages results in calling
"global" command which fails and shows the stack trace shown below.

#### Reproduction guide :beetle:
- Remove package-user-dir, i.e., all installed packages.
- Use default ~/.spacemacs except to add few packages to 
  dotspacemacs-additional-packages
- Start Emacs

*Observed behaviour:* :eyes: :broken_heart:
Stack trace shown above

*Expected behaviour:* :heart: :smile:
Emacs starts up without stack trace.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 25.1.1
- Spacemacs: 0.105.21
- Spacemacs branch: develop (rev. 36f3b63)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: helm
- Layers:
```elisp
((auto-completion :variables auto-completion-return-key-behavior 'complete auto-completion-tab-key-behavior 'cycle auto-completion-complete-with-key-sequence nil auto-completion-enable-help-tooltip t auto-completion-enable-sort-by-usage t auto-completion-private-snippets-directory nil)
 (c-c++ :variables c-c++-default-mode-for-headers 'c++-mode)
 clojure command-log emacs-lisp eyebrowse
 (git :variables git-magit-status-fullscreen t)
 github gtags html javascript latex markdown org perforce python shell shell-scripts helm spell-checking syntax-checking version-control my-navi)
```

#### Backtrace :paw_prints:
```
Debugger entered--Lisp error: (error "‘global’ non-zero exit: global: GTAGS not found.")
[...]

PR #6829, opened 2016-08-15
Further improve gtags layer

TheBB:
- Add option to disable by default
- Use local-vars hook to allow per-project enable/disable
- Don’t enable helm-gtags-mode (no need)
- Move emacs bindings from helm-gtags-mode-map to ggtags-mode-map

Fixes https://github.com/syl20bnr/spacemacs/issues/6819


TheBB:
Merged.

PR #6834, opened 2016-08-16
Added IBM Selectric image to the README.org

rbanffy:
Replaced the Hammond 12 image with an image from a 1961 IBM Selectric similar to the one from which the sound files were recorded.


d12frosted:
Thank you for this PR. Could you [please](https://github.com/syl20bnr/spacemacs/blob/master/CONTRIBUTING.org#ideally-for-simple-prs-most-of-them) squash all your commits into one?

PR #6846, opened 2016-08-17
Allow customization of spaceline additional segments.

riccardomurri:
Introduce a new customization variable
`spacemacs-spaceline-additional-segments', which is a list of the
additional segments that should be inserted in the modeline when it is
initialized.

### DIFF
--- a/layers/+distribution/spacemacs/packages.el
+++ b/layers/+distribution/spacemacs/packages.el
@@ -9,6 +9,18 @@
 ;;
 ;;; License: GPLv3
 
+(defcustom spacemacs-spaceline-additional-segments
+  '((new-version :when active))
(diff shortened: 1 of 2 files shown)


layers/+distribution/spacemacs/packages.el
```diff
@@ -9,6 +9,18 @@
 ;;
 ;;; License: GPLv3
 
+(defcustom spacemacs-spaceline-additional-segments
```
bmag: Layer variables usually go in `config.el`.

riccardomurri: Fixed in commit c69aea0. Thanks for pointing this out.

PR #6847, opened 2016-08-17
spacemacs-base: evilify archive-mode

lislon:
Basic evil keys are now work when viewing archives (zip, jar, ect.).


TheBB:

Issue #5667, opened 2016-04-04
Add more SPC m bindings for Clojure


gilch:
If you look through the menus in a Clojure buffer or Clojure REPL, you can see that a lot of the functions with Ctrl bindings in insert mode have no SPC bindings in normal mode. Is there some reason for leaving these out? There are plenty of letters left on the keyboard. Maybe CIDER has evolved since its layer was written.

kalouantonis:
If you have more specific examples, I'd be happy to add them

gilch:
> I've added all extra cider commands that my version supports (0.14.0snapshot). These include:
> cider-classpath
> cider-browse-ns
> cider-browse-ns-all
> cider-auto-test-mode
> cider-enlighten-mode

There are dozens more in the menus. They change a bit with context. I think you have to have a Clojure REPL running to see them. Use `SPC m s i` to jack in (some configuration/installation may be required).

In the Clojure major mode, hit `F10` and look through the `Clojure`, `CIDER`, `CIDER Eval`, and `CIDER Interactions` menus. Then hit `SPC m s s` to switch to the running Clojure REPL and hit `F10` and look through the `REPL` menus. Anything with a `C-` binding is probably important enough to have a `SPC m` binding in normal mode too.

gilch:
@TheBB hold it!

The PR from @kalouantonis was incomplete. This is **not** fixed in develop. I just said there are dozens of other bindings not covered:

## Missing SPC bindings for Clojure mode with CIDER minor mode
```
key             binding
---             -------

C-:             clojure-toggle-keyword-string

C-M-q           prog-indent-sexp
C-M-i           complete-symbol

C-c C-r a       clojure-unwind-all
C-c C-r f       clojure-thread-first-all
C-c C-r i       clojure-cycle-if
C-c C-r l       clojure-thread-last-all
C-c C-r p       clojure-cycle-privacy
C-c C-r t       clojure-thread
C-c C-r u       clojure-unwind

C-c C-r #       clojure-convert-collection-to-set
C-c C-r '       clojure-convert-collection-to-quoted-list
C-c C-r (       clojure-convert-collection-to-list
C-c C-r [       clojure-convert-collection-to-vector
C-c C-r {       clojure-convert-collection-to-map

C-c C-r n h     clojure-insert-ns-form-at-point
C-c C-r n i     clojure-insert-ns-form
C-c C-r n s     clojure-sort-ns
C-c C-r n u     clojure-update-ns

[...]

PR #6848, opened 2016-08-17
clojure: Add more SPC m bindings

kalouantonis:
Possible fix for issue #5667.

I've added all extra cider commands that my version supports (`0.14.0snapshot`). These include:
* cider-classpath
* cider-browse-ns
* cider-browse-ns-all
* cider-auto-test-mode
* cider-enlighten-mode


TheBB:

PR #6849, opened 2016-08-17
Allow customization of spaceline additional segments.

riccardomurri:
Introduce a new customization variable
`spacemacs-spaceline-additional-segments', which is a list of the
additional segments that should be inserted in the modeline when it is
initialized.

(This is basically the same patch as PR #6846 but applied to the "develop"
branch instead of "master".)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ my-org/
 org-files/
 semanticdb/
 edts/
+/recentf
 .recentf
 .recentf~
 projectile.cache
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -18,6 +18,7 @@ my-org/
 org-files/
 semanticdb/
 edts/
+/recentf
```
d12frosted: This change should not be part of this PR 😸

TheBB:
Thanks! Cherry-picked in develop (at least, the relevant commit). You can safely delete your branch.

PR #6850, opened 2016-08-17
Update emacs-eclim package name

CestDiego:
I think MELPA has an `old-names` back
[compatibility](https://github.com/melpa/melpa/issues/4159) measure for
cases like this, but for now it isn't working. And it's breaking all new
installs that use the java layer.

### DIFF
--- a/layers/+lang/java/packages.el
+++ b/layers/+lang/java/packages.el
@@ -12,7 +12,7 @@
 (setq java-packages
       '(
         company
-        emacs-eclim
+        eclim
         ggtags
         helm-gtags
         (java-mode :location built-in)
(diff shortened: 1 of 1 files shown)


purcell:
Yep, package has been renamed permanently in MELPA, so this PR is necessary.

TheBB:
Cherry-picked in develop. Leaving this open for now, have to think about master.

purcell:
> Cherry-picked in develop. Leaving this open for now, have to think about master.

Any reference to the old package name is just going to break.

TheBB:
I'm aware. I just seem to remember another commit that should have gone into master but I can't recall which. :-s

TheBB:
Released as 0.105.22.

PR #6853, opened 2016-08-18
Don't use golden ratio mode in Idris holes buffer

houli:
Drawing inspiration from the Agda layer https://github.com/syl20bnr/spacemacs/blob/b2d108877c290d6965f45e34df437346a47a90ea/layers/%2Blang/agda/packages.el#L80-L82
This change makes it so that the Idris type holes buffer does not open in golden ratio mode.


TheBB:

See https://github.com/syl20bnr/spacemacs/commit/da2af5c1fda0e1118364fd2f7e53aa76b84f561a for some changes I made. This is more idiomatic in Spacemacs.

PR #6856, opened 2016-08-18
Python: Set breakpoints correctly when pyenv is used.

cpaulik:
pyenv shims fool executable-find into thinking that ipdb is always
installed if it is installed in one environment.


TheBB:

PR #6857, opened 2016-08-18
SML: Improve indentation.

cpaulik:
I was suffering from the behaviour described at http://stackoverflow.com/questions/14374127/funky-indentation-in-sml-mode This should fix it as per the provided solution.


TheBB:

PR #6863, opened 2016-08-19
nixos: inhibit spacemacs indation in nix-mode

Profpatsch:
previously [this](https://github.com/NixOS/nix/issues/828#issuecomment-240698002) happened. Turns out it  was because of spacemacs' default indentation.

I looked into resolving the TODO (in the definition of the inhibit function), but I was not entirely sure where to put it in the other layers.


TheBB:

PR #6872, opened 2016-08-20
Share installed packages among multiple users

emacs18:
This PR allows me to install packages once which then can be used by all my colleagues without each
of them having to install the same pacakges.  There are potentially hundreds of my colleagues
who can be using the same shared file system so that the ability to share files is important not
only to save disk space, but also to reduce startup time.

Created dotspacemacs-use-shared-packages user option, so that if it is
set to non-nil, then directories listed in package-directory-list are
also searched for installed packages in addition to package-user-dir.
Site administrator can then install packages in one or more of the
directories in package-directory-list so that all users at the site can
share the same installed packages without each user having to install
them.

This change has no impact unless dotspacemacs-use-shared-packages is
explicitly changed to non-nil from the default value of nil.

### DIFF
--- a/core/core-funcs.el
+++ b/core/core-funcs.el
@@ -36,12 +36,18 @@
   (run-hooks 'text-mode-hook))
 
 (defun spacemacs//get-package-directory (pkg)
-  "Return the directory of PKG. Return nil if not found."
-  (let ((elpa-dir (file-name-as-directory package-user-dir)))
-    (when (file-exists-p elpa-dir)
-      (let* ((pkg-match (concat (symbol-name pkg) "-[0-9]+"))
-             (dir (car (directory-files elpa-dir 'full pkg-match))))
-        (when dir (file-name-as-directory dir))))))
+  "Return the installed directory of PKG. Return `nil' if not found."
+  (let ((directories (cons package-user-dir
+                           (and dotspacemacs-use-shared-packages
+                                package-directory-list)))
+        elpa-dir dir)
+    (while (and (not dir) directories)
+      (setq elpa-dir (pop directories))
+      (when (file-exists-p elpa-dir)
+        (let* ((pkg-match (concat (symbol-name pkg) "-[0-9]+"))
+               (dir (car (directory-files elpa-dir 'full pkg-match))))
+          (when dir (setq dir (file-name-as-directory dir))))))
+    dir))
 
 (defun spacemacs/mplist-get (plist prop)
   "Get the values associated to PROP in PLIST, a modified plist.
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -124,6 +124,12 @@ when the current branch is not `develop'. Note that checking for
 new versions works via git commands, thus it calls GitHub services
 whenever you start Emacs.")
 
+(defvar dotspacemacs-use-shared-packages nil
+  "If non-`nil', then look in all directories of `package-directory-list' when
+looking for the installed directory of a package which was not found under
+`package-user-dir'. Site administrators can install packages in one of
+these system-wide directories so that all users at the site can benefit.")
+
 (defvar dotspacemacs-configuration-layers '(emacs-lisp)
   "List of configuration layers to load.")
 


emacs18:
This is an update of PR 5464 which is now obsolete.

bmag:
I support the idea, but there must be a better way than this, no? AFAIK `site-lisp` (e.g. `/usr/local/share/emacs/24.5/site-lisp`) is intended for shared Emacs config, including shared packages. Can you explain what are the problems with it?

Also, I'm not sure why change `spacemacs//get-package-directory`, but not `configuration-layer//get-package-directory`. Is there a reason? Can you please explain the technical reasoning and impact of this PR?

core/core-funcs.el
```diff
@@ -36,12 +36,18 @@
...
 
 (defun spacemacs//get-package-directory (pkg)
-  "Return the directory of PKG. Return nil if not found."
-  (let ((elpa-dir (file-name-as-directory package-user-dir)))
-    (when (file-exists-p elpa-dir)
-      (let* ((pkg-match (concat (symbol-name pkg) "-[0-9]+"))
-             (dir (car (directory-files elpa-dir 'full pkg-match))))
-        (when dir (file-name-as-directory dir))))))
+  "Return the installed directory of PKG. Return `nil' if not found."
+  (let ((directories (cons package-user-dir
```
bmag: The result of this `cons` is not a list. I think you want `append` instead. Example from IELM session:
```elisp
ELISP> (cons (list 1 2 3 4) (list 5 6))
((1 2 3 4)
 5 6)

ELISP> (append (list 1 2 3 4) (list 5 6))
(1 2 3 4 5 6)
```

bmag: Nevermind, realised my mistake (that `package-user-dir` is not a list).

core/core-funcs.el
```diff
@@ -36,12 +36,18 @@
...
+                           (and dotspacemacs-use-shared-packages
+                                package-directory-list)))
+        elpa-dir dir)
+    (while (and (not dir) directories)
+      (setq elpa-dir (pop directories))
+      (when (file-exists-p elpa-dir)
+        (let* ((pkg-match (concat (symbol-name pkg) "-[0-9]+"))
```
bmag: `(pkg-match ...)` can be moved to the outer `let` and out of the loop , while we're at it.

core/core-funcs.el
```diff
@@ -36,12 +36,18 @@
...
+               (dir (car (directory-files elpa-dir 'full pkg-match))))
+          (when dir (setq dir (file-name-as-directory dir))))))
+    dir))
```
bmag: There's probably a more idiomatic/lispy way to refactor this `while` loop. Maybe `(catch 'found (dolist ... (throw 'found dir)))` or `cl-loop`?

core/core-funcs.el
emacs18: There is nothing wrong with site-lisp.  In fact I want to put installed packaged under site-lisp.  With the proposed code, I do have the installed files under site-lisp.  The issue is how does spacemacs know that the packages are installed already so that it does not install them under package-user-dir?

AFAIK package-directory-list was created to solve this problem, i.e., to allow site administrators to install packages under site-lisp directories.

emacs18: Why not change configuration-layer//get-package-directory?
It never occurred to me.  In my testing just applying the change in this PR was sufficient in that
spacemacs honored the packages installed under site-lisp and so it did not install them under
package-user-dir.

emacs18: After examining the value of package-alist, it appears that this variable already has the correct paths, i.e., 'dir' field of package-desc struct points to site-lisp directory rather than package-user-dir.

This brings up a question I had for a while.  Why does spacemacs not use package-alist to figure out the location of the installed package, and instead have its own code to search the file system?

core/core-funcs.el
emacs18: Sure.  Once the dust settles and if this PR is acceptable, then I'll update this PR with this change.

core/core-funcs.el
emacs18: I considered using catch/throw as well.  To me importance is not whether I use "idomatic" code.
To me writing simple code that most people can understand is a very good reason to stick with
simple syntax when it does not hurt that much.  I use catch/throw when I think it improves the
readability of the code.  In this particular case, I'm not sure that it matters too much.

core/core-funcs.el
emacs18: If spacemacs used package-alist to figure out package installation directory instead of searching package-user-dir,  then I think this PR would not be needed.  All I would have to do is just install packages under site-lisp, and package.el would take care of things.

core/core-funcs.el
emacs18: Following uses catch/throw.  If you like this one better, I can switch to using it.
It does not matter to me which one is used.

```
(defun spacemacs//get-package-directory (pkg)
  "Return the installed directory of PKG. Return `nil' if not found."
  (let ((directories (cons package-user-dir
                           (and dotspacemacs-use-shared-packages
                                package-directory-list)))
        (pkg-match (concat (symbol-name pkg) "-[0-9]+"))
        elpa-dir dir)
    (catch 'found
      (mapcar (lambda (elpa-dir)
                (when (file-exists-p elpa-dir)
                  (let ((dir (car (directory-files elpa-dir 'full pkg-match))))
                    (if dir
                        (throw 'found (file-name-as-directory dir))))))
              directories))))
```

bmag: A `catch` block returns its last argument if nothing was thrown. In this case, if the `mapcar` doesn't throw `found`, then `catch` will return the result of the `mapcar`. If you use `dolist`, this won't happen. Perhaps we can refactor this later, if the PR is accepted.

bmag:
@emacs18 I was trying to understand the idea behind the PR, because it looked weird to me that we need one. My questions were partly to help me understand the code, and partly to make you aware to parts of the code that I didn't know if you're aware of. I still don't understand all the background behind this PR, so I'll stop spamming this PR.

syl20bnr:
@emacs18 What's your current solution ? Can you provide more details about your use case ? What are the technologies involved, what is the OS etc ... ?

Users cannot mount the shared file system at the correct place to get Spacemacs happy ?

emacs18:
The use case is with hundreds of developers using many dozens of linux
servers all with NFS shared file system.  Starting emacs typically takes
over a minute with about 300 packages enabled.  Compare this with about 6
seconds on a laptop.  This is assuming that all packages have already been
installed.  If packages were not installed, then it takes over 10 minutes
to install 300 packages.  This is assuming that all packages are located
within the intranet so that
internet is not accessed to start and use emacs.

My goal was to make it as simple as possible for someone to try out
spacemacs by setting
up emacs so that no package need to be installed on startup.  However I
have not been able
to reduce the emacs startup time below about 25 seconds which is just about
the best time
I've seen on a linux server.  As I said before, typical time is over a
minute.  This is one key reason why I have pretty much have given up trying
to get colleagues to test drive spacemacs.  I have dozens of colleagues who
have been using emacs or vim for many years.  After trying for over a year,
I have not a single person who uses spacemacs besides myself.

emacs18:
I've adopted the following method to deploy emacs that multiple folks can use within my group.
- build emacs.  The install directory will be referred to as $root.
- install packages to $root/share/emacs/site-lisp/elpa
- install spacemacs under $root/share/emacs/site-lisp/spacemacs
- provide a common startup file anyone can use.

Issue #6266, opened 2016-06-08
Spacemacs-layouts - buffer isolation issue


pavkum:
#### Description
Buffers among the layouts are overlapping. I was redirected to [here](http://develop.spacemacs.org/doc/DOCUMENTATION.html#layouts-and-workspaces) from gitter chat which says layouts provide buffer isolation.

#### Reproduction guide
- Start Emacs
- Create a new layout by SPC l 2
- Create your custom window setup (split horizontally / vertically as per your choice)
- Save configuration by SPC l s
- Restart emacs
- Load layout by SPC l L
- switch to new layout SPC l 2
- open some random buffer
- switch to default layout 
- open buffer list for switch - C-x b (opens Helm buffers in my case)

*Observed behaviour:*
Buffer opened in 2nd layout is seen in 1st layout

*Expected behaviour:*
Buffer opened in 2nd layout shouldn't be visible in 1st layout

#### System Info
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.20
- Spacemacs branch: master (rev. 9c7250a)
- Graphic display: t
- Distribution: spacemacs
- Editing style: emacs
- Completion: helm
- Layers:
```elisp
(emacs-lisp html git typescript javascript eyebrowse org)
```

NJBS:
Can you still reproduce this when not using the default layout?

From http://develop.spacemacs.org/doc/DOCUMENTATION.html#the-default-layout :

The default layout is special because it has a global scope which means that all the opened buffers belong to it. So using only the default layout feels like not using layouts at all.

pavkum:
@NJBS Yes I can.

- created new layout at 3.
- Opened a random buffer 
- switched to 2nd layout
- Helm buffer shows the buffer created in 3

syl20bnr:
@pavkum I cannot reproduce it, maybe we need to backport a fix in master. Can you try with develop ?

pavkum:
@syl20bnr I got the cause.

I had added `ido` to exclude packages. This causes the issue. 
If I remove `ido` from exclude packages I cannot reproduce the issue.

pavkum:
@syl20bnr Basically I wanted kill buffer list to be same as `Helm`, by default I guess it shows `ido-buffers`. Instead of adding `ido` to exclude packages I disabled ido. Now in layers if I press kill switch `C-x k` helm buffers list all the buffers but if I enable ido and try I get only the buffers of that particular layout.

[...]

PR #6876, opened 2016-08-20
layouts: fix B b b not restricting to layout-local buffers

### LINKED ISSUE #5788: Spacemacs layouts - SPC b b not layout local if invoked before SPC B b if ido is excluded
#### Description
`SPC b b` is not layout local until `SPC B b` was invoked once when `ido` is in `dotspacemacs-excluded-packages`

#### Reproduction guide
- Start Emacs
- `SPC p l` `SPC b b`. This shows all buffers for me. 
- Then `SPC B b` `ESC` `SPC b b` shows only the buffers of the project.

*Expected behaviour:*
The first `SPC b b` should only show the layout buffer.

#### System Info
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.18
- Spacemacs branch: develop (rev. 5ee970f)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(ansible
 (auto-completion :variables auto-completion-enable-snippets-in-popup t auto-completion-private-snippets-directory "~/dotfiles/.spacemacs-config/snippets/" auto-completion-enable-help-tooltip nil)
 chrome clojure colors custom_org_config django
 (elfeed :variables rmh-elfeed-org-files
         (list "~/Dropbox/org/feeds.org"))
 emacs-lisp emoji erc finance games
 (git :variables git-gutter-use-fringe t)
 github gnus graphviz html ipython-notebook javascript latex lua markdown my-mu4e my_python org org-ipython org-page pandoc pdf-tools php
 (python :variables python-enable-yapf-format-on-save nil python-auto-set-local-pyenv-version 'on-project-switch python-test-runner
         '(pytest))
 ranger research-config restclient ruby rust search-engine semantic
 (shell :variables shell-default-shell 'term shell-default-term-shell "/bin/zsh")
 shell-scripts
 spacemacs-helm spacemacs-layouts syntax-checking systemd tmux vagrant
 yaml)
```

#### Backtrace
```
BACKTRACE IF RELEVANT
```

aroig:
This fixes two issues regarding layouts:

1. Since Bad-ptr/persp-mode.el@e950bf15, persp-mode requires setting
persp-hook-up-emacs-buffer-completion in order to install the hooks for
ido & friends. This variable is nil by default, making SPB b b not
restrict to layout-local buffers.

2. The function spacemacs-layouts/non-restricted-buffer-list removes a
hook and re-adds it later. This makes the assumption that the hook was
already present. If it was not (due to 1) then SPC B b changes global
state by adding that hook. Instead, just let-bind the hook variable
for the scope we need it changed.

This explains and partially fixes #5788 and #6266. It does not fix the
dependency on ido-mode. If ido-mode is excluded, persp-mode will not
install the hook for ido, and SPB b b will still be unrestricted.

### DIFF
--- a/layers/+spacemacs/spacemacs-layouts/funcs.el
+++ b/layers/+spacemacs/spacemacs-layouts/funcs.el
@@ -51,9 +51,8 @@ current perspective."
 
 (defun spacemacs-layouts/non-restricted-buffer-list ()
   (interactive)
-  (remove-hook 'ido-make-buffer-list-hook  #'persp-restrict-ido-buffers)
-  (helm-mini)
-  (add-hook 'ido-make-buffer-list-hook  #'persp-restrict-ido-buffers))
+  (let ((ido-make-buffer-list-hook (remove #'persp-restrict-ido-buffers ido-make-buffer-list-hook)))
+    (helm-mini)))
 
 
 ;; Persp transient-state
--- a/layers/+spacemacs/spacemacs-layouts/packages.el
+++ b/layers/+spacemacs/spacemacs-layouts/packages.el
@@ -107,7 +107,8 @@
             persp-nil-name dotspacemacs-default-layout-name
             persp-reset-windows-on-nil-window-conf nil
             persp-set-last-persp-for-new-frames nil
-            persp-save-dir spacemacs-layouts-directory)
+            persp-save-dir spacemacs-layouts-directory
+            persp-hook-up-emacs-buffer-completion t)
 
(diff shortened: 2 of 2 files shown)


TheBB:
That's strange, SPC b b is still restricted for me even with that setting `nil`. Anyway…  Thanks! Cherry-picked in develop. You can safely delete your branch.

PR #6886, opened 2016-08-21
Remove duplicated words in docstring

Andre0991:
"Open junk file" is duplicated in the docstring.


d12frosted:

PR #6891, opened 2016-08-22
HotFix. Add missing format function

CestDiego:
This kinda broke the javascript layer for me. This is the fix.


CestDiego:
Oh nvm this was just fixed with latest commit. :D

PR #6895, opened 2016-08-23
Include *idris-holes* and *idris-info* in golden ratio exclude

houli:
Missed these two with the previous PR #6853


TheBB:

Issue #6896, opened 2016-08-23
spacemacs align


vinurs:
in c function, i wrote this:
`char path[PATH_MAX] = {0};`
`        int ret             = -1;`
then press `space x a a`, it only align `=`, it seems that it cannot align variables.

TheBB:
`align` operates according to very specific rules that you can customize. You should read the documentation for this function. The results you want can be achieved with `align-regexp`:

* Select the region
* `M-x align-regexp C-u RET` (assuming helm)
* Enter `\(\s-+\)` as regexp
* 1
* 1
* y

`SPC x a r` is the easiest interface for this. Unfortunately it doesn't work with whitespace regexps, which is fixed by https://github.com/syl20bnr/spacemacs/pull/6897.

vinurs:
@TheBB thanks.
It seems that there is sth wrong with the default align rule.
Variables like these can be aligned correctly:
`        struct test_a     *a       = NULL;`
`        struct test_b     *b       = NULL;`
`        struct test_hello *c       = NULL;`
the variable names are aligned too.
but when i add array, it can not deal with correctly:
`        char     newpath[PATH_MAX] = {0};`

TheBB:
You're right, we don't have a fully fledged C parser for alignment. If you want more advanced behaviour you should implement it yourself (either as a regexp or an align rule).

TheBB:
To be clear, what you're seeing is because some types have a space in them, there's no way to know which is which unless you happen to be able to parse C, which we can't (at least not in this context).

TheBB:
Actually, I'm able to align that block with `SPC x a a`. Can you confirm?

vinurs:
@TheBB yes. 
the following block only aligned by `=`, the `a/b/c` variables are aligned too, but the last variable `newpath` doesn't aligned with `a/b/c`
        struct test_a     *a               = NULL;
        struct test_b     *b               = NULL;
        struct test_hello *c             = NULL;
        char newpath[PATH_MAX] = {0};

d12frosted:
Fixed with released of Spacemacs v0.200. Let us know if you still have any problems with this issue.

PR #6897, opened 2016-08-23
Allow align-repeat to work with empty regexps

TheBB:
This makes `SPC xar` align whitespace-delimited columns when given an empty regexp. Fix https://github.com/syl20bnr/spacemacs/issues/6896


TheBB:
Merged

PR #6910, opened 2016-08-24
Improve NeoTree discoverability: new TS, `?` hint, and clarified docs

### LINKED ISSUE #6867: It looks like Neotree has been configured for worse discoverability.
I'm bran new to emacs, so I had my start directory configured to the emacs installation folder which means I'd have to navigate out of that directory to open anything on emacs. I didn't know how to do that.

I tried various hotkeys that would make sense to me like shift-tab, alt-up arrow (the win10 hotkey), pressing enter  and h on the "/users/yxven/whatever" line. Nothing worked.

I pressed space to see if there was anything there. I didn't see anything related to neotree. (Apparently, I should have looked harder here. I know SPC h k would have helped now. Whatever, I'm new.)

I start googling. I find out the plugin I'm dealing with is neotree. I check the neotree docs. The neotree docs mention nothing about going to to a previous directory. https://www.emacswiki.org/emacs/NeoTree

Although, the screenshot suggests this isn't a problem for the default neotree installation because the default neotree installation has a line that says ".. (up a dir)" at the very top. It also has a "press ? for neotree help." Both of these are disabled in spacemacs.

In any case, I googled around a lot more and eventually learned that I should type K. It took me around an hour to learn how to navigate to the previous directory. 

tldr:
neo-show-updir-line should be enabled by default. Disabling it hurts discoverability for no good reason.

bmag:
Fixes #6867 

Changes `?` binding in NeoTree from `evil-search-backward` to `describe-mode` and adds a mtaching banner at the top of NeoTree buffer. The aim is to increase discoverability of NeoTree, and I think we can give up on `evil-search-backward` for that. Besides, `?` is bound to `describe-mode` also in un-configured NeoTree (because it inherits from `special-mode`).

Also adds some verbs to the docs describing NeoTree movement keys, to make it less laconic.

### DIFF
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -2091,17 +2091,17 @@ current state. With default =spacemacs-dark= theme:
 Navigation is centered on the ~hjkl~ keys with the hope of providing a fast
 navigation experience like in [[http://ranger.nongnu.org/][ranger]]:
 
-| Key Binding  | Description                                                            |
-|--------------+------------------------------------------------------------------------|
-| ~h~          | collapse expanded directory or go to parent node                       |
-| ~H~          | previous sibling                                                       |
-| ~j~          | next file or directory                                                 |
-| ~J~          | next expanded directory on level down                                  |
-| ~k~          | previous file or directory                                             |
-| ~K~          | parent directory, when reaching the root change it to parent directory |
-| ~l~ or ~RET~ | expand directory                                                       |
-| ~L~          | next sibling                                                           |
-| ~R~          | make a directory the root directory                                    |
+| Key Binding  | Description                                                                   |
+|--------------+-------------------------------------------------------------------------------|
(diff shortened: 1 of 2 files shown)


bmag:
@nixmaniack I added your changes from #6951. I had to fix up the TS a bit, but it's working now. I also just realized that the TS is available only for hybrid and vim style users. Emacs style users have `?` bound to `describe-mode`. I'll probably make a separate TS for emacs style (in this PR or a follow-up), because they have different keys.

TheBB:
Quality stuff. Thanks @bmag and @nixmaniack! Cherry-picked in develop. You can safely delete your branch.

nixmaniack:
@bmag I wasn't very sure of adding `j/k` to TS just because it's too common of the keybindings in `vim` world. 

> Emacs style users have `?` bound to `describe-mode`

I see. I didn't think of `emacs` style! It needs some work then.

PR #6914, opened 2016-08-25
document spacemacs-space-doc-center-buffer-mode-min-aspect-ratio

JAremko:
Add the `spacemacs-space-doc-center-buffer-mode-min-aspect-ratio` variable mention into the `space-doc` FAQ section.


JAremko:
I'll need to rework it.

PR #6915, opened 2016-08-25
Keep current random ascii banner and startup summary when auto-centering emacs buffer on window resize

deb0ch:
Written on top of #6259 by @ralesi.

Now, for having tested it, these two PRs combined offer a smooth home buffer experience :smile: 

This PR has several commits because they are tightly related and depend on each other.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -1913,15 +1913,21 @@ to select one."
         (configuration-layer//insert-lazy-install-form
          layer-name (car x) (cdr x))))))
 
+(defvar configuration-layer//spacemacs-startup-time nil
+  "Spacemacs full startup duration.")
+
 (defun configuration-layer/display-summary (start-time)
   "Display a summary of loading time."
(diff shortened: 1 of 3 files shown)


TheBB:
Ah, sorry, I jumped the gun on the random banner fix.

TheBB:
Thanks though. In the end I only needed the last commit. Note, Spacemacs conventions dictate `/` for functions and `-` for variables, so I renamed your variable.

TheBB:
Cherry-picked in develop. You can safely delete your branch.

deb0ch:
Thanks for the precision, I didn't get that yet. Just kept wondering why it is sometimes `/`, `//`, `-`, `--`, `|` or `||`.

Now that you say that I just took a look at CONVENTIONS and it is way clearer now :smile: 

I liked that how I did it it allowed for changing the banner when refreshing the startup page though. But I guess it is not really important and your way takes less code. Is it intentional, leaving the same even on refresh ?

PR #6920, opened 2016-08-26
Improve README of the Clojure layer

sebastianpoeplau:
Explicit configuration of leiningen or boot is not necessary anymore as of CIDER 0.11.


d12frosted:

PR #6931, opened 2016-08-27
add glsl-company to shaders layer

d12frosted:
Following #6359. Not pushing it directly mainly because `company-glsl` provides very simple completion. So simple that word expansion is much more powerful. 

So I am going to see if anyone is interested in this change. 

cc @Ell


d12frosted:
Cherry-picked this. Let the 🐊 to enter the box!

PR #6941, opened 2016-08-28
Add new base16 color schemes

belak:
The summerfruit theme was removed (and replaced with dark/light variants) and the github theme was added.

This happened upstream, followed by an update of the base16-emacs package.


TheBB:

PR #6942, opened 2016-08-28
Rust: Add binding for describing symbol at point

NJBS:
Adds a binding for the new [racer-describe](http://www.wilfred.me.uk/blog/2016/08/27/rustdoc-meets-the-self-documenting-editor/) function.

Seems to work fairly well out of the box.

### DIFF
--- a/layers/+lang/rust/README.org
+++ b/layers/+lang/rust/README.org
@@ -61,4 +61,5 @@ To enable automatic buffer formatting on save, set the variable =rust-format-on-
 | ~SPC m c u~ | update dependencies with Cargo              |
 | ~SPC m c x~ | execute a project with Cargo                |
 | ~SPC m g g~ | jump to definition                          |
+| ~SPC m h h~ | describe symbol at point                    |
 | ~SPC m t~   | run tests with Cargo                        |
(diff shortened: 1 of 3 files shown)


Stebalien:
Nice. Could you,

1. Evilify `racer-help-mode` mode (defaulting to motion mode would make the most sense).
2. Focus it by default (or provide an option to do so). That would make it easier to navigate/close.

NJBS:
@Stebalien Thanks for the suggestions!

The buffer is now focused by default since it seems more consistent with the built-in emacs describe-function and other implementations for major-mode help systems.

Additionally, I've evilified the the `racer-help-mode`. Although it seems like each call to `racer-describe` leaves behind a new buffer. I could see how this might be useful but it seems a bit inconsistent with how help buffers usually work and a long session could require the user to have to eventually manually purge the buffer list. One solution might be to bind `q` in the evilified state such that it also deletes the buffer when quitting. Unfortunately, I haven't had the chance to take a look at the implementation details yet so I unaware of any possible side-effects (perhaps @Wilfred could chime :smile:).

Suggestions welcome!

Wilfred:
See https://github.com/racer-rust/emacs-racer/issues/48 :)

I'll increasingly persuaded that we should create multiple buffers, to be honest.

Wilfred:
*should never create

bmag:
Thank you :+1: Cherry-picked into develop, you can safely delete your branch. I've added some fixes in 58c8fd515a9a0ab9bde21cfb531a5541c353b77a, such as:
- changed the formatting of funcs.el from dos to unix (newlines)
- `spacemacs/racer-describe` selects the window only if `help-window-select` is non-nil
- replaced `evilified-state-evilify` with `evilified-state-evilify-map` and moved it to `:config`

Issue #5821, opened 2016-04-18
version-control-diff right side on Emacs GUI and not working properly on Emacsclient GUI?


myme5261314:
I'm not sure why my version-control-diff symbols appear on the right side of the Emacs GUI like below:

`
#### System Info
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.18
- Spacemacs branch: develop (rev. b123b21)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(html spacemacs-helm spacemacs-ivy better-defaults github
      (version-control :variables version-control-diff-tool 'git-gutter version-control-global-margin t)
      semantic markdown
      (vinegar :variables vinegar-reuse-dired-buffer t)
      org prodigy search-engine
      (syntax-checking :variables syntax-checking-enable-by-default nil)
      (spell-checking :variables spell-checking-enable-by-default nil)
      yaml python lua html command-log emacs-lisp emoji fasd deft ranger gtags
      (spacemacs-layouts :variables layouts-enable-autosave t layouts-autosave-delay 300)
      (colors :variables colors-enable-nyan-cat-progress-bar t)
      (git :variables git-magit-status-fullscreen t magit-push-always-verify nil magit-save-repository-buffers 'dontask magit-revert-buffers 'silent magit-refs-show-commit-count 'all magit-revision-show-gravatars nil)
      (ibuffer :variables ibuffer-group-buffers-by 'projects)
      (c-c++ :variables c-c++-default-mode-for-headers 'c++-mode)
      (auto-completion :variables auto-completion-enable-sort-by-usage t auto-completion-enable-snippets-in-popup t)
      (shell :variables shell-default-position 'full shell-default-shell 'ansi-term shell-default-term-shell "/bin/zsh")
      (chinese :variables chinese-default-input-method 'pinyin chinese-enable-fcitx t chinese-enable-youdao-dict t)
      emacs-lisp markdown org dash my)
```
![image](https://cloud.githubusercontent.com/assets/1814831/14592988/ac5a2f6c-0557-11e6-9c6c-3e79f8ef69d8.png)

#### System Info
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.18
- Spacemacs branch: develop (rev. b123b21)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
[...]

Issue #2180, opened 2015-07-02
0.103: getting diff-hl in the margin also in a GUI setup


emmanueltouzery:
since 0.103 version-control uses diff-hl not anymore git-gutter.
in my .spacemacs I put:

      (setq diff-hl-side 'left)

same as I had left-fringe for git-gutter. But I doubt it has anything to do with my problem. The problem is that when just starting `emacs` diff-hl displays in the fringe as expected. However if I start it with `emacsclient -c "$@" --alternate-editor= &` then I get diff-hl consistently in the margin instead of the fringe.

if I comment these lines in the `version-control` layer then it's again all ok:

    (unless (display-graphic-p)
       (setq diff-hl-side 'left)
       (diff-hl-margin-mode))

it looks like emacsclient somehow triggers `display-graphic-p` to be false. I attempted changing the condition to `window-system` instead of `display-graphic-p` but the problem remains.

emmanueltouzery:
now added `(diff-hl-margin-mode 0)` in my `.spacemacs` as a workaround.

emmanueltouzery:
this might be related to this: http://stackoverflow.com/a/5801740/516188

> You can start emacs in daemon mode, then later tell it to create a graphical frame. [..]
> As a result of this, avoid putting code in your .emacs that depends on window-system. Instead, put code like your set-frame-size example in a hook function which runs after a frame is created

d12frosted:
Fixed with released of Spacemacs v0.200. Let us know if you still have any problems with this issue.

Issue #1894, opened 2015-06-08
emacs --daemon weird behaviour


CestDiego:
Hello, I share this to know if other people are having the same issue:

when running emacs as daemon I get this:
![alt](https://i.imgur.com/5SA3hII.png)

And when opening an emacsclient sth feels off (note the coloring on the window numbers)
![alt](https://i.imgur.com/Ocy5IyD.png)

And then when running something that takes input (such as helm-projectile) I get the prompt at the terminal where the daemon is running >very weird:
![alt](https://i.imgur.com/UsCemWq.png)

Also, `SPC :` doesn't seem to be defined, etc, etc I don't know if this is a problem with spacemacs but it looks as if spacemacs is not fully loading while being called as a daemon.

tuhdo:
Could you reproduce with stock Spacemacs?

CestDiego:
No I can't ;_; which means...I'm the problem right? dayum ;_;

tuhdo:
Not necessary. To find out, you can remove your personal config and private layer, and only leave the pre-configured layers active. Maybe there's a problem with a layer. If not, then, you now know where to look at :)

CestDiego:
:) ye, I'm gonna disable all my layers. then I'll se what's up with that.

CestDiego:
This is very weird, so I disable all my layers and just stay with spacemacs' ones, and then the problem rises again D: I know it stops after `keybindings`. And what's after that is
```
Loading /home/io/.emacs.d/.cache/recentf...done
Start checking for new version...
Open the quickhelp.

```

CestDiego:
ok So I deleted the nyan-cat stuff that checked if emacs was running as a graphical system. and now I don't see the prompt at the terminal and `SPC :` is defined, buuuut still 

![alt](https://i.imgur.com/baYHN3A.png)

Shows that there is sth wrong in the modeline.

syl20bnr:
@CestDiego what does the screenshot show ?

CestDiego:
@syl20bnr normally I would be able to have the numbers in black background, what I feel here, is that when I make a new frame after running as a daemon, none of the graphical features work, (spacemacs logo doesn't show, nyancat, etc)

robbyoconnor:
Affected by this -- just wanna be included

mistnim:
my spacemacs when running as daemon can't find source code pro, there are problems and the mode-line doesn't have fancy graphic.

CestDiego:
[...]

PR #6946, opened 2016-08-28
Fix some features in daemon mode

TheBB:
We can't reliably use `(display-graphic-p)` during startup. Therefore assume no graphic display and use `spacemacs|do-after-display-system-init` otherwise.

Probably fixes: https://github.com/syl20bnr/spacemacs/issues/5821, https://github.com/syl20bnr/spacemacs/issues/2180 and https://github.com/syl20bnr/spacemacs/issues/1894

Tested in daemon mode with all combinations of margin tool and frame type.

### DIFF
--- a/layers/+completion/auto-completion/packages.el
+++ b/layers/+completion/auto-completion/packages.el
@@ -14,7 +14,7 @@
         auto-complete
         ac-ispell
         company
-        company-quickhelp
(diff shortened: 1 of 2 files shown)


TheBB:
Added a fix for company-quickhelp-mode as well.

TheBB:
Merged.

PR #6953, opened 2016-08-29
Haskell: Add support for `intero-apply-suggestions`

Tehnix:
As per #6233 add support for `intero-apply-suggestions` behind `SPC m r s`, 
and move the prefix declaration for "haskell/refactor" to the other prefix declarations, 
since it is no longer only relevant for hlint.

### DIFF
--- a/layers/+lang/haskell/README.org
+++ b/layers/+lang/haskell/README.org
@@ -283,8 +283,12 @@ Refactor commands are prefixed by ~SPC m r~:
 |-------------+---------------------------------------------------|
 | ~SPC m r b~ | apply all HLint suggestions in the current buffer |
 | ~SPC m r r~ | apply the HLint suggestion under the cursor       |
+| ~SPC m r s~ | list all Intero suggestions                       |
 
 Only some of the HLint suggestions can be applied.
+
+To apply the intero suggestions, press `C-c C-c` when the window is open, which
+is also shown in the window that appears.
 
 ** Ghc-mod
 These commands are only available when ghc-mod is enabled.
--- a/layers/+lang/haskell/packages.el
+++ b/layers/+lang/haskell/packages.el
@@ -139,7 +139,8 @@
         (spacemacs/declare-prefix-for-mode mode "ms" "haskell/repl")
         (spacemacs/declare-prefix-for-mode mode "mc" "haskell/cabal")
         (spacemacs/declare-prefix-for-mode mode "mh" "haskell/documentation")
-        (spacemacs/declare-prefix-for-mode mode "md" "haskell/debug"))
+        (spacemacs/declare-prefix-for-mode mode "md" "haskell/debug")
+        (spacemacs/declare-prefix-for-mode mode "mr" "haskell/refactor"))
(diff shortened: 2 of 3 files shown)


Tehnix:
Some points that might be worth addressing:

- Should the shortcut be `SPC m r s`, I went with this because of the "suggestions" part
- Currently the user needs to press `C-c C-c` to actually apply the suggestions, which is clearly mentioned when the fix window opens (you can see an example in the image at the end). 

I considered if it was worth adding a shortcut for this (the `C-c C-c` command), but thought it might be confusing (unless there is some way to alter that text without changing the `intero` package?)

I also moved the "haskell/refactor" prefix declaration because that seemed to make more sense than having it under hlint-init specifically, but am more than willing to move that back if there are any opinions on it :)

screenshot 2016-08-29 11 44 38

layers/+lang/haskell/packages.el
```diff
@@ -67,6 +67,7 @@
     :defer t
     :config
     (progn
+      (spacemacs/set-leader-keys-for-major-mode 'haskell-mode "rs" 'intero-apply-suggestions)
```
d12frosted: I would move it to [this function](https://github.com/Tehnix/spacemacs/blob/064650b9421efb90be9f0e7b3d2c30acd759876c/layers/%2Blang/haskell/funcs.el#L49). Now we support several completion backends and `spacemacs-haskell//setup-completion-backend` calls proper setup function whenever `haskell` source file is opened. This flow allows user to change completion backend on per-project basis. 

This key binding doesn't conflict with key bindings from other completions... but who knows if it will in the future? :)

layers/+lang/haskell/packages.el
```diff
@@ -284,7 +286,6 @@
     :defer t
     :init
     (progn
-      (spacemacs/declare-prefix-for-mode 'haskell-mode "mr" "haskell/refactor")
```
d12frosted: Thanks. I like this change.

d12frosted:
Also, could you please add this key binding to readme file? 😸

Tehnix:
@d12frosted good catch! I'll add an entry and move the keybinding to the intero setup as well :)

layers/+lang/haskell/README.org
```diff
@@ -283,9 +283,12 @@ Refactor commands are prefixed by ~SPC m r~:
 |-------------+---------------------------------------------------|
 | ~SPC m r b~ | apply all HLint suggestions in the current buffer |
 | ~SPC m r r~ | apply the HLint suggestion under the cursor       |
+| ~SPC m r s~ | list all Intero suggestions                       |
 
 Only some of the HLint suggestions can be applied.
 
+To apply the intero suggestions, press `C-c C-c` when the window is open, which is also shown in the window that appears.
```
d12frosted: Hey,

The last thing, could you please `fill-paragraph` here? With emacs editing style it's `M-q` and should be the same for vim editing style, but I am not 100% sure. Thanks.

d12frosted:
Sorry, I don't get notifications when you make new commits in PR. Better to ping me 😸 

Issue #6949, opened 2016-08-29
emacs-eclim broken on the stable version?


emmanueltouzery:
#### Description
emacs-eclim doesn't work for me anymore with the latest stable release (0.105.22). It worked with 0.105.21 and the old eclim package.

#### Reproduction guide
- Start Emacs with the java layer enabled
- Open a java file

*Observed behaviour:*
Getting messages:

    Company backend ’company-emacs-eclim’ could not be initialized:
    Symbol’s function definition is void: company-emacs-eclim
    eclim--problems-mode-init: Symbol’s value as variable is void: G374

*Expected behaviour:*
The java file should open without warnings and if the eclim server is running, completion, go to definition, etc should work.

#### System Info
- OS: gnu/linux
- Emacs: 25.1.1
- Spacemacs: 0.105.22
- Spacemacs branch: master (rev. 9f9faa4)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(auto-completion :variables auto-completion-enable-help-tooltip t)
```
The `company-emacs-eclim` dependency does not appear to get installed, but even if it did, I think it would not work. I tried to force install it manually, to no avail:

    Contacting host: melpa.org:443
    Making version-control local to company-emacs-eclim-autoloads.el while let-bound!
    Generating autoloads for company-emacs-eclim.el...done
    Wrote /home/emmanuel/.emacs.d/elpa/company-emacs-eclim-20160813.1754/company-emacs-    eclim-autoloads.el
    Checking /home/emmanuel/.emacs.d/elpa/company-emacs-eclim-20160813.1754... [3 times]
    Compiling /home/emmanuel/.emacs.d/elpa/company-emacs-eclim-20160813.1754/company-emacs-eclim.el...done
    Wrote /home/emmanuel/.emacs.d/elpa/company-emacs-eclim-20160813.1754/company-emacs-eclim.elc
    Checking /home/emmanuel/.emacs.d/elpa/company-emacs-eclim-20160813.1754...
    Done (Total of 1 file compiled, 2 skipped)
    Package menu: Operation finished [Installed 1]
    ReportingService.java has auto save data; consider M-x recover-this-file
    eclim--problems-mode-init: Symbol’s value as variable is void: G374

TheBB:
[...]

PR #6954, opened 2016-08-29
Fix java layer after eclim package refactor

TheBB:
Fix https://github.com/syl20bnr/spacemacs/issues/6949

### DIFF
--- a/layers/+lang/java/packages.el
+++ b/layers/+lang/java/packages.el
@@ -12,6 +12,7 @@
 (setq java-packages
       '(
         company
+        (company-emacs-eclim :toggle (configuration-layer/package-usedp 'company))
         eclim
         ggtags
         helm-gtags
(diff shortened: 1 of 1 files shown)


emmanueltouzery:
The patch is a definite progress. However I'm still having issues, but it may not be spacemacs' fault anymore. I see errors like:

    eclim--problems-mode-init: Symbol’s value as variable is void: G374
    Company: An error occurred in auto-begin
    Company: backend company-emacs-eclim error "Invalid function: "-l"" with args (candidates main)

Opened https://github.com/emacs-eclim/emacs-eclim/issues/7 too.

TheBB:
Yeah, looks like an upstream bug.

TheBB:
I merged this. Please continue discussing the eclim issues in #6949.

PR #6955, opened 2016-08-29
ycmd layer: a better global_conf.py

deb0ch:
Rewrote global_conf.py to use compile_commands.json or .clang_complete to get compilation flags.

I used the following heuristics to get missing files from compile_commands.json:

1) If there is a compile_commands.json in a parent directory:
  a) If the file is a header file:
    - search for the header file itself in db
    - search for a sibling source file in the same directory (i.e. a source file with the same name but different extension)
    - search for a source file that includes our header's path
    - search for the nearest source file in db
  b) If the file is a source file:
    - search for the source file itself
    - search for the nearest source file in db
2) If no compile_commands.json, search for a .clang_complete:
    - get flags from .clang_complete
3) Always try to add extra flags from a .ycm_extra_flags file in a parent directory.
(like --sysroot="/path/to/your/toolchain/libc if you are cross-compiling)

This should bring the ycmd layer closer to working out of the box :smiley:

### DIFF
--- a/layers/+tools/ycmd/global_conf.py
+++ b/layers/+tools/ycmd/global_conf.py
@@ -1,84 +1,274 @@
+# global_conf.py --- ycmd global configuration file for Spacemacs
+#
+# Copyright (c) 2012-2016 Sylvain Benner & Contributors
+#
+# Author: Thomas de Beauchene <thomas.de.beauchene@gmail.com>
+# URL: https://github.com/syl20bnr/spacemacs
+#
+# This file is not part of GNU Emacs.
+#
+# License: GPLv3
+#
+# This script tries to get the compilation flags for a translation unit using
+# the following logic:
+#
+# 1) If there is a compile_commands.json in a parent directory:
+#   a) If the file is a header file:
+#     - search for the header file itself in db
+#     - search for a sibling source file in the same directory (i.e. a source
+#       file with the same name but different extension)
+#     - search for a source file that includes our header's path
+#     - search for the nearest source file in db
+#
+#   b) If the file is a source file:
+#     - search for the source file itself
+#     - search for the nearest source file in db
+#
+# 2) If no compile_commands.json, search for a .clang_complete:
+#   - get flags from .clang_complete
+#
+# 3) Always try to add extra flags from a .ycm_extra_flags file in a parent
+#    directory. (like --sysroot="/path/to/your/toolchain/libc" if you are cross-compiling)
+#
+# Thanks to Jonas Devlieghere and Gabor Marton for their work on which this code is based.
+# https://jonasdevlieghere.com/a-better-youcompleteme-config/
+# https://github.com/martong/ycm_extra_conf.jsondb
+
+import itertools
+import json
+import logging
 import os
+import os.path
+import re
 import ycm_core
 
-default_flags = [
-'-Wall',
-'-Wextra',
-'-Werror',
-'-Wc++98-compat',
-'-Wno-long-long',
-'-Wno-variadic-macros',
-'-fexceptions',
-'-DNDEBUG',
-'-std=c++11',
-'-x',
-'c++',
-# This path will only work on OS X, but extra paths that don't exist are not harmful
-'-isystem'
-'/System/Library/Frameworks/Python.framework/Headers',
-'-isystem',
-'/usr/include',
-'-isystem',
-'/usr/local/include',
-'-isystem',
-'/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1',
-'-isystem',
-'/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include',
-]
+SOURCE_EXTENSIONS = ['.cpp', '.cxx', '.cc', '.c', '.m', '.mm']
+HEADER_EXTENSIONS = ['.h', '.hxx', '.hpp', '.hh']
 
-def DirectoryOfThisScript():
-  return os.path.dirname( os.path.abspath( __file__ ) )
+# This function is called by ycmd.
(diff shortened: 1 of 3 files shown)


TheBB:
This loogs good to me, except:

1. The path for the python file needs to use `configuration-layer/get-layer-path` instead of `configuration-layer/get-layer-local-dir` (since it's not in the local dir). I'm not sure how this worked in the first place.

2. Catch the exception with `except Exception as e`, not `except Exception, e` which does not work on Python 3.

3. The logic for finding fallback files should be case insensitive.

4. Is there a reason clang_complete flags don't go through the relative to absolute path filter?

TheBB:
Does https://github.com/TheBB/spacemacs/commit/3d5ef09da09e9c6a10e41e13ff65feda87d66a1a look ok to you?

dubnde:
I really like this as I `use` ycmd a lot. Is it possible to make the location of where to look for the compilation database files `.clang_complete` or `compile_commands.json` configurable? That is, leave the default as is but give an option to change it. In my usage for example, I have out of source builds which create the compilation database not necessarily in the root folder.

Just a thought 😄

layers/+tools/ycmd/global_conf.py
```diff
@@ -1,84 +1,271 @@
...
-        curr_dir = new_dir
-        opt_fname = os.path.join(curr_dir, opt_basename)
+SOURCE_EXTENSIONS = ['.cpp', '.cxx', '.cc', '.c', '.m', '.mm']
+HEADER_EXTENSIONS = ['.h', '.hxx', '.hpp', '.hh']
+
+# This function is called by ycmd.
+def FlagsForFile(filename):
+    logging.info("%s: Getting flags for %s" % (__file__, filename))
+    root = os.path.realpath(filename);
+    compilation_db_flags = FlagsFromCompilationDatabase(root, filename)
```
dubnde: Is it possible to make the root configurable?

TheBB:
I'm not sure how feasible it is to pass config through ycmd to the python file. Maybe we could support a solution based on env vars, but isn't it simpler just to copy or symlink the compile_commands.json file? That's what I do… usually it doesn't change often.

dubnde:
@TheBB Yes. I should have mentioned it is not much effort to copy the file over but was just wondering and thought best to ask so we have it documented.

deb0ch:
@dubnde I didn't think of that particular use case, it's true that it would be nice to be able to do that.

Though I don't think it would be possible to configure it in elisp through emacs, as we don't have an accessible python API for that.

Environment variables might be an idea, but then how to go about it when you have several projects ?

We could also say that the location of the database can be stored in a configuration file at the root of the project, but then what is the advantage of that approach over a simple symlink ?

It is a long time that I'm thinking of putting more info in the .projectile files for things like that, but I don't know if that would be a good idea: I don't know if there would be any side effect, or even if you would want a .projectile file at that particular place.

I don't know what approach we could take for doing that in a way that would be more practical than a symbolic link.

@TheBB: Thanks for your work ! Glad that you took interest in it :smile: 

l46: Very good idea, I didn't think of case sensitivity.
l54: Yeah, thanks, it is more elegant that way. After a few reworks I guess it just went over my head.
l60: There is a problem here that I forgot to address before and that is still present, it is that flags can be `None` here and it would trigger an exception. I would like to always add the extra flags and keep the logs accurate. I will rework it sometime soon, if you don't before me.
l98: didn't know about that, thanks !
l129-131-133: ooops, forgot some crap
l151: I don't see the difference with what was before, can you explain ?

1) I didn't know about it actually, always used my own. Just assumed that the old one was called in the first place.
2) Thanks !
3) Yep.
4) No reason that I know of, we could try doing that.

deb0ch:
Just updated with the issue at l60 fixed and absolute paths for .clang_complete.

The rest is already taken care of by your commit, just the point at l151 that I don't really understand.

TheBB:
Ah, 151 is to be able to handle files like this.one.with.several.dots.cpp.

deb0ch:
Ah, that's what I thought. It is already taken care of, `os.path.splitext()` only splits the extension

TheBB:
Ah, of course. :+1:

deb0ch:
:smile_cat: 

I don't know what could be further improved. I'm sure a lot of things.

Tell me if you find something :smile:

TheBB:
No, I think it's good for now.

In the future we can add ycmd to other modes, https://github.com/syl20bnr/spacemacs/pull/6964 should make it simple to disable per-user.

deb0ch:
I think it is a good idea, it would make the behaviour of auto-completions more consistent accross languages (except for the ones without ycmd).

Though, except for fuzzy-matching I don't really see what difference it would make, let's say for a language like Python where there already is autocompletion and code navigation. Is the semantic analysis better with ycmd for these languages, compared to what there already is ?

TheBB:
No, probably not. Ycmd and anaconda-mode both use jedi in the back. I think for rust the situation is the same. It feels just a little peculiar to have a tool that supports several modes being restricted in Spacemacs to only one (or two) of them.

Besides, at least for me, the backend-handled fuzzy matching is a killer feature I'm really missing from other modes. I often have to skip out of company into helm with `C-/` to avoid typing prefixes. Mostly in emacs lisp of course, which ycmd doesn't do, but still.

deb0ch:
Yes, I really like the fuzzy matching and it's kind of missing in python. And I'd love it in elisp :smiley: 

Whatever, I also want to try rtags out sometime soon as I just realized that for code navigation ycmd restricts you to the current translation unit and its headers whereas rtags allegedly indexes all the files in the compilation db, but it doesn't have fuzzy matching.

I wonder if it would be worth having them both at the same time.

deb0ch:
By the way, I cherry-picked your changes into this PR, not sure if it is the way to go about it.

TheBB:
That's fine.

TheBB:

PR #6951, opened 2016-08-29
neotree: add ts for easy discoverabilty of keys

nixmaniack:
Attempted to fix #6867

It looks like this. Any suggestions/modifications are welcome. 

screen shot 2016-08-29 at 7 11 58 pm

Not sure about the warning in the echo area. @bmag any idea about it?

PR #6957, opened 2016-08-29
Assign 0 to NeoTree's window only when it is the top-left window

bmag:
When a transient state is invoked from NeoTree's window, an error message is shown:

window-number-double

The error happens because, for a very brief moment, two windows display NeoTree: the original, and the new window that is created for the TS. This PR solves the issue by assigning the number 0 only to the top-left NeoTree window.

(image copied from #6951)


TheBB:

Issue #6916, opened 2016-08-25
*FORUM* Should space-doc enable centered-buffer-mode by default?


JAremko:
### I wonder what other spacemacsers think

:+1:  YES 
- - - - - - -
:-1:  NO

#### WTF is `centered-buffer-mode` ?
*Note that it's configured to enable only on wide screens (frames)*

![http://i.imgur.com/7C9TLng.png](http://i.imgur.com/7C9TLng.png)
![http://i.imgur.com/xCeS5AN.png](http://i.imgur.com/xCeS5AN.png)
![http://i.imgur.com/p0c8T29.png](http://i.imgur.com/p0c8T29.png)

deb0ch:
I really like the idea behind your mode, and I think that it could be applied to way more things than just the docs :smile: 

Though I think it is wiser to not set it by default but under an easily accessible keybinding and let people discover and try it for some time before considering setting it as default.

JAremko:
@deb0ch  it's already bound to `w` `c`

deb0ch:
(I didn't mean the contrary :stuck_out_tongue_winking_eye: )

dubnde:
Is there a currently a setting to disable it by default?

TheBB:
Yes, you can configure `spacemacs-space-doc-modificators`

dubnde:
@TheBB ta

d12frosted:
While I do like the idea, I don't think I'll ever use it. Also what I really don't like - black background. It looks especially bad with light themes (which I always use). 

But I really believe that some plain text writes might like it :D Distraction free, centered text editing 💯

syl20bnr:
@JAremko what do you mean by purging it ? We can just disable it by default.

deb0ch:
Yeah, don't remove it please :smile:

TheBB:
Yes, my understanding is that this vote is about not enabling the mode by default when viewing docs. Keeping `SPC w c` is not up for discussion (at least, not here).

JAremko:
@syl20bnr   I will disable it by the default. But also it might make sense to remove the integration between the `space-doc` and `centered-buffer` modes(given no one seems to like it enabled) it might allow to remove a ton of hacks and simplify `space-doc`  FAQ.

Andre0991:
I agree with @d12frosted - the black background feels a bit weird depending on the theme. 

I wonder if there's  some way to match the theme colour (not using the exactly same background colour, but a similar one, I'd say).

[...]

PR #6958, opened 2016-08-29
space-doc-mode and centered-buffer-mode improvements

JAremko:
As promised in #6916

### DIFF
--- a/doc/FAQ.org
+++ b/doc/FAQ.org
@@ -528,7 +528,9 @@ You can modify the list of visual enhancements applied by the =space-doc-mode=:
         org-kbd-face-remap
         resize-inline-images))
 #+END_SRC
-It's the default value of the variable with all enhancements.
+By default only =center-buffer-mode= is disabled.
+Both =space-doc-mode= and =center-buffer-mode= can be customized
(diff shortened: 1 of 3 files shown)


JAremko:
It will be incremental PR.

JAremko:
### TODO

- [x]  Disabled by default centering in space-doc-mode.
- [x] 	Recenter buffer when content width increase.
- [ ] 	Disable space-doc if org-mode disabled (replaced with another major mode).
- [ ] 	Use test properties instead of overlays.
- [ ] 	*BONUS:*  Restore previous layout  after `SPC w C` `SPC w C`

JAremko:
@TheBB  :shipit:

JAremko:
@TheBB  ok I think we need to merge it asap so the changes will be battle-tested before the next release.

*It might be the last major rework.* :fireworks:

syl20bnr:
Thank you ! 👍 
I bet that

> Remove [WIP]

was the hardest of all! :-)

I noticed an issue with winner-undo, the fringe color is not reset. So I guess we should do the BONUS to be sure to correctly toggle the mode.

PR #6964, opened 2016-08-30
Implement :enabled-for layer keyword

TheBB:
This implements an `:enabled-for` feature that does the opposite of what `:disabled-for` already does.

Rationale: except for the sake of completeness, this might be useful for layers like ycmd and gtags. Ycmd shines in C/C++ but it also supports auto-completion in a number of different major modes where we already have other options (python, rust, C#). Gtags often competes with other jump-to-definition handlers (although, that has been largely resolved already with another method). In both cases it could be useful to restrict ycmd/gtags to work only in the modes where the user feels they are needed. This way we don't have to introduce superfluous layer variables to turn on or off support, and the user doesn't have to keep up with a potentially changing list of additional layers that they might have to disable.

Layers where this could be useful should probably explicitly note this possibility in their documentation.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -97,7 +97,12 @@ ROOT is returned."
    (disabled :initarg :disabled-for
              :initform nil
              :type list
-             :documentation "A list of layer where this layer is disabled."))
+             :documentation "A list of layers where this layer is disabled.")
+   (enabled :initarg :enabled-for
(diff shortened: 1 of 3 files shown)


TheBB:
CI failure looks unrelated, btw.

syl20bnr:
Excellent 💯  This is a perfect PR, code + doc + tests = me happy 💜

PR #6966, opened 2016-08-30
Option to choose intepreter for python-execute-file

bogdanvuk:
In the function python-execute-file, the name of the python interpreter
has been hard-coded to "python". If one uses Python 3, there is no way
to invoke "python3" interpreter. Invoke python interpreter by the name
set in python-shell-interpreter variable, i.e.:

(setq python-shell-interpreter "python3")

### DIFF
--- a/layers/+lang/python/packages.el
+++ b/layers/+lang/python/packages.el
@@ -327,9 +327,9 @@
         (setq universal-argument t)
         (if arg
             (call-interactively 'compile)
-
-          (setq compile-command (format "python %s" (file-name-nondirectory
-                                                     buffer-file-name)))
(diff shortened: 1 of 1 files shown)


TheBB:
> If one uses Python 3, there is no way to invoke "python3" interpreter.

Although I don't mind this PR, you should be using a virtual environment where the python executable is python 3. This solves one out of 50 potential problems whereas a virtual environment solves all 50. (Slight hyperbole.)

cpaulik:
I think it is useful for people who do only very little python programming and don't want to bother with environments.

TheBB:
At any rate, this change seems to have been made already. Sorry! Thank you for the effort though.

PR #6967, opened 2016-08-30
Use lowercase in bug report submit binding

tchajed:
"C-C C-C" is confusing since the Emacs convention is to use lowercase
letters in keybindings.


d12frosted:

I think it was because of overusing `M-u` 😸 

To be fair, this is not because of convention, but because the key binding is `C-c C-c` and not `C-C C-C` 😸

tchajed:
Ah yes, I was misremembering how shift keys are noted. Thanks for the quick turnaround!

Issue #6968, opened 2016-08-30
Upgrading spacemacs without upgrading packages results in error message about spacemacs//window-numbering-assign


tchajed:
It looks like commit 4b2f123a052 removed a workaround for nschum/window-numbering.el#10, which has been fixed upstream. Thus upgrading Spacemacs past 4b2f123a052 on the develop branch without also upgrading packages (or at least window-numbering.el) results in a fairly broken Emacs, where many operations (including `:` and quitting Emacs) results in failure and the error message `Symbol's value as variable is void: spacemacs//window-numbering-assign` in the echo area.

Is there a way for Spacemacs to warn that upgrading packages is needed, or better yet, specifically offer to upgrade window-numbering.el?

TheBB:
No, there's no such mechanism. Spacemacs quite explicitly tracks the latest packages, users should not upgrade one without also upgrading the other.

xged:
I updated packages, *pulled* latest `develop` branch and been having the same issue on `develop` (`master` is fine). *Chekingout* 4b2f123 did not help.

bmag:
@xged the package update must have gone wrong. In updated develop branch and latest window-numbering package, there is no code that treats `spacemacs//window-numbering-assign` as a variable, so the error above just isn't possible.

To make sure window-numbering is updated, find it's directory inside of `~/.emacs,d/elpa`, The directory should be called `window-numbering-<timestamp>`. The timestamp should be `20160809.1110`.

xged:
I forgot to mention that updates did went wrong (I believe it was `the tls connection was non-properly terminated`)
@bmag The timestamp of `window-numbering` is `20150228.1247` therefore not updated.

bmag:
The work-around was re-added, this issue can be closed.

PR #6974, opened 2016-08-31
re-fix window numbering bug

deb0ch:
4b2f123 recently removed a workaround which according to the comment was fixing some window-numbering bug. I just reapplied the workaround as it broke my emacs.

While it was working nice on my computer at work, it broke everything on my personal laptop: when trying to quit emacs, it wouldn't, but instead all commands would only split windows.

After manually sigterming the frame, I still had to use sigkill to get rid of the emacs process taking 100% CPU.

### DIFF
--- a/layers/+spacemacs/spacemacs-ui/packages.el
+++ b/layers/+spacemacs/spacemacs-ui/packages.el
@@ -301,4 +301,8 @@ debug-init and load the given list of packages."
                  ;; assigning 0 only to the top-left window
                  (eq (selected-window) (window-at 0 0)))
         0))
-    (setq window-numbering-assign-func #'spacemacs//window-numbering-assign)))
+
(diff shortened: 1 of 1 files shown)


NJBS:
Can you try updating your packages such that you have the latest window-numbering and seeing if this issue still persists?

https://github.com/syl20bnr/spacemacs/issues/6968

syl20bnr:
Thank you ! 👍 
Even if the bug is fixed with the last version of `windows-numbering` I prefer to merge this fix temporarily, in a few weeks we can revert it.

deb0ch:
Looks like I jumped the gun on that one, didn't even see the related issue or think of updating my packages.

But I agree that it makes sense to wait a bit so that everybody will have updated his packages before :smile:

PR #6981, opened 2016-09-01
add fill-or-unfill function

d12frosted:
Using this gem for pretty long time and really like it. People who doesn't have to unfill paragraphs even would not notice the change 😸 💃


d12frosted:

SteveJobzniak:
@d12frosted Thanks for making me aware of this but it's not the best version of the code... I suggest deleting this commit.

> Due to implementation details unfilling doesn't work when called twice via ~M-x~.

Here is the evolved version of the exact code, and it is much more advanced and handles `M-x` correctly:

https://github.com/purcell/unfill/blob/master/unfill.el

So undo this commit and merge a dependency on the unfill package instead.

PR #6983, opened 2016-09-01
elixir: flycheck on idle

aaronjensen:
`flycheck-mix` now uses `flycheck-buffer-saved-p` so it is safe to
flycheck on idle: https://github.com/tomekowal/flycheck-mix/pull/2


TheBB:

PR #6987, opened 2016-09-01
Fix Python jumping with async anaconda mode function.

### LINKED ISSUE #6985: Spacemacs jumping does not work correctly with anaconda mode.
It seems that the new implementation of jumping in spacemacs has some problems with anaconda mode.

1. Sometimes `dumb-jump-go` is used when `anaconda-mode-find-definitions` executed directly works without problems. I have not yet found a minimal working example for this case but I'll try to find one.
2. Spacemacs reports `"No jump handler was able to find this symbol."` but that is not true since a correct jump was made. 

  Example:
  1. create a file called `first.py` with the following contents:
  ```python
  def test():
      pass
  ```
  2. In the same folder create `second.py`:
  ```python
  from first import test
  ```
  In `second.py` put the cursor on `test` and `SPC m g g`. It hangs a little (sometimes) then jumps to the definition of `test` and writes the message `No jump handler was able to find this symbol.`. 

My guess is that these two issues are related and have to do with the fact that anaconda mode sends the responses async.

cpaulik:
Without this dumb-jump-go is often used if anaconda-mode-find-defintions
takes a little to return a result.

fix #6985

### DIFF
--- a/layers/+lang/python/packages.el
+++ b/layers/+lang/python/packages.el
@@ -49,7 +49,7 @@
             (concat spacemacs-cache-directory "anaconda-mode"))
       (add-hook 'python-mode-hook 'anaconda-mode)
       (add-hook 'spacemacs-jump-handlers-python-mode
-                'anaconda-mode-find-definitions))
+                '(anaconda-mode-find-definitions :async t)))
     :config
     (progn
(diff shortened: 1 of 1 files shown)


TheBB:

Issue #6918, opened 2016-08-26
Alias org-set-tags to counsel-org-tag when using ivy and org


darkfeline:
The default ivy completion and `org-set-tags` is fairly useless since it doesn't complete after the first tag.  `ivy`/`counsel` has a function for this that should replace `org-set-tags` entirely if `ivy` layer is used.

For reference: https://github.com/abo-abo/swiper/issues/177

github-god:
I double that.

syl20bnr:
@darkfeline thank you for the PR, merged in develop branch.

codygman:
I think this can be closed FYI.

duianto:
Any issue that's marked as `Fixed in develop` will be closed when the next version is released.
But we can close this now. Thanks for the suggestion.

PR #6993, opened 2016-09-02
Replace org-set-tags when ivy layer is used

darkfeline:
When ivy is used, the default org-set-tags will only complete for the
first tag, and any additional tags must be typed manually.  counsel
provides a replacement function to fix this.

See https://github.com/syl20bnr/spacemacs/issues/6918


TheBB:
Can you move this to a post-init-ivy function?

darkfeline:
Since the alias is overriding an `org` symbol, if I put it in `post-init-ivy`, wouldn't loading `org` overwrite the alias?

syl20bnr:
Thank you ! 👍 

I changed the config for a `post-init-org` function in commit https://github.com/syl20bnr/spacemacs/commit/14b20649

Issue #5636, opened 2016-03-30
Describe scss-lint setup in the html layer documentation.


JAremko:
[The HTML layer](https://github.com/syl20bnr/spacemacs/tree/develop/layers/%2Blang/html) provides `flycheck-mode` hook for `scss-mode` and it uses [scss-lint](https://github.com/brigade/scss-lint). I managed to setup it, but it was kinda tricky and I'm not sure this is the right way so it might be beneficent for those who came after us to say something, in the layer's docs, about `scss-lint` and how to use it.

I had to install required ruby gems directly from the github since [rubygems.org doesn't have a compitable version with flycheck](https://rubygems.org/gems/scss-lint/versions) - I think it requires at least [`SCSS-Lint 0.43.0`](https://github.com/brigade/scss-lint/releases/tag/v0.43.0)

This is my way:

```sh
   gem install specific_install 
   gem specific_install https://github.com/brigade/scss-lint.git 
   gem specific_install https://github.com/Sweetchuck/scss_lint_reporter_checkstyle.git 
```

Not sure if the best one.

JAremko:
Should also mention the configuration file https://github.com/brigade/scss-lint/blob/master/config/default.yml since the default linter indentation  is 2 (Spacemacs uses 4 for sass)  if not make it  equal there will be a ton of warnings after fmting

JAremko:
Solved in #6995

PR #6995, opened 2016-09-02
Describe scss-lint setup in the html layer documentati #5636

JAremko:
Solves #5636


JAremko:
@d12frosted  Can we merge it? It's not like it will break anything  :confused:

TheBB:

Issue #6997, opened 2016-09-02
REGRESSION: auto centering spacemacs-buffer makes resizing of windows incredibly choppy


aaronjensen:
#### Description :octocat:
*This probably only repros on macs*

Auto centering spacemacs-buffer (a7d4674) makes resizing of windows with mouse when spacemacs buffer is shown super choppy.

#### Reproduction guide :beetle:
- Start Emacs
- Open more than one window, at least one should be the spacemacs home buffer
- Try to resize a window with mouse

*Observed behaviour:* :eyes: :broken_heart:
You're able to resize the window but emacs freezes every couple of seconds.

*Expected behaviour:* :heart: :smile:
Emacs behaves as it does when the spacemacs home buffer is not shown, that is, resizing with the mouse works smoothly.

#### System Info :computer:
- OS: darwin
- Emacs: 25.1.1
- Spacemacs: 0.105.22
- Spacemacs branch: nil (rev. 4e7d80b)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: helm
- Layers:
```elisp
(better-defaults spacemacs-layouts helm emacs-lisp markdown
                 (syntax-checking :variables syntax-checking-enable-tooltips nil)
                 (auto-completion :variables auto-completion-enable-sort-by-usage nil auto-completion-enable-snippets-in-popup nil)
                 erlang elixir git dash html clojure
                 (org :variables org-enable-github-support t)
                 colors osx github javascript react deft ruby
                 (shell :variables shell-default-shell 'ansi-term shell-default-height 30 shell-default-position 'bottom)
                 spell-checking ranger version-control rcirc tmux yaml docker elm evil-little-word auto-correct frame-geometry cleverparens-lispy flow)
```

aaronjensen:
Related: #6986

TheBB:
I'm beginning to think there's something wrong with your build. This also works fine here.

aaronjensen:
@TheBB you're not on a mac. Mac window resizing is totally funky

aaronjensen:
Actually, if I keep the mouse held down, it does eventually go to where I'm holding, it just janks like crazy:

![jank](https://cloud.githubusercontent.com/assets/8588/18204498/feab7baa-70d1-11e6-86fe-55907924523d.gif)

aaronjensen:
[...]

PR #6998, opened 2016-09-02
core: Add resize hook later in startup

### LINKED ISSUE #6986: REGRESSION: auto centering spacemacs-buffer causes several second startup hang
#### Description :octocat:
Auto centering spacemacs-buffer (a7d4674) when combined with using
initial-frame-alist causes a several second hang during starutp.

- Add something like this to your init.el:
```
(setq initial-frame-alist
      '(
        (top . 23)
        (left . 0)
        (width . 122)
        (height . 40)))
```
- Start Emacs

Emacs freezes for several seconds after the window resizes.

Emacs does not freeze.

#### System Info :computer:
- OS: darwin
- Emacs: 25.1.1
- Spacemacs: 0.105.22
- Spacemacs branch: nil (rev. 4e7d80b)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: helm
- Layers:
```elisp
(better-defaults spacemacs-layouts helm emacs-lisp markdown
                 (auto-completion :variables auto-completion-enable-sort-by-usage nil auto-completion-enable-snippets-in-popup nil)
                 colors osx github javascript react deft ruby
                 spell-checking ranger version-control rcirc tmux yaml docker elm evil-little-word auto-correct frame-geometry cleverparens-lispy flow)
```

aaronjensen:
Fixes hang while resizing on startup with initial-frame-alist.

Fixes #6986

### DIFF
--- a/core/core-spacemacs-buffer.el
+++ b/core/core-spacemacs-buffer.el
@@ -903,7 +903,10 @@ already exist, and switch to it."
         (switch-to-buffer spacemacs-buffer-name)
         (spacemacs//redisplay)))))
 
-(add-hook 'window-configuration-change-hook 'spacemacs-buffer//resize-on-hook)
+(add-hook 'window-setup-hook
+          (lambda ()
+            (add-hook 'window-configuration-change-hook 'spacemacs-buffer//resize-on-hook)
+            (spacemacs-buffer//resize-on-hook)))
 
 (defun spacemacs-buffer//resize-on-hook ()
   (let ((space-win (get-buffer-window spacemacs-buffer-name))


TheBB:
That looks peculiar, but seems to make sense from the docstring. Thanks!

aaronjensen:
No problem. This still does not fix #6997 which would require some sort of debouncing and i'm not sure how to do that w/ emacs.

aaronjensen:
Given that #6986 ended up being a configuration change issue, this isn't necessary. It may be slightly more correct?

TheBB:
Yes, I would keep this anyhow.

TheBB:
I put your commit on top of https://github.com/syl20bnr/spacemacs/pull/7007 just to keep everything in one place.

PR #7007, opened 2016-09-02
Home buffer

TheBB:
As discussed in https://github.com/syl20bnr/spacemacs/issues/6986 and other places.

### DIFF
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -327,6 +327,9 @@ List sizes may be nil, in which case
 `spacemacs--buffer-startup-lists-length' takes effect.
 ")
 
+(defvar dotspacemacs-startup-buffer-responsive t
+  "True if the home buffer should respond to resize events.")
+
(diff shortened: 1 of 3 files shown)


JAremko:
I hope it doesn't break centering. It is my by far favorite feature in the upcoming release :confused:

deb0ch:
@JAremko Don't worry, by looking at it it doesn't seem to break anything, just resolve some performance issues and add a dotspacemacs var for deactivating the responsiveness on resize events.

Centering is nice, but it is even nicer if it works for everyone :smile: 

Really glad that the default value is `t` :smiley:

Btw, centering of startup lists is in progress !

TheBB:
IMO the main content shouldn't be centered. Lists are easier to read and browse through if they're adjusted.

TheBB:
Merged.

deb0ch:
I am not *actually* centering it because as you said it doesn't look nice.

The PR is ready and I should submit it soon, I guess you'll see then what I mean exactly.

PR #7011, opened 2016-09-03
Adjust or center startup lists in home buffer

deb0ch:
Just centering startup lists would look weird, especially when all the entries are short: the beginning would be almost in the middle of the page.

Moreover, an inconsistent position of the lists would make them less easy to read.

This is why the position of the beginning of the startup lists is relative to the position of the home buffer buttons, unless the lines are too long to fit the screen. In that case, the lists are centered.

### DIFF
--- a/core/core-spacemacs-buffer.el
+++ b/core/core-spacemacs-buffer.el
@@ -17,6 +17,10 @@
 (defconst spacemacs-buffer-logo-title "[S P A C E M A C S]"
   "The title displayed beneath the logo.")
 
+(defconst spacemacs-buffer-buttons-startup-lists-offset 25
+  "Relative position in characters of the home buffer buttons and the home
+ buffer startup lists.")
+
 (defconst spacemacs-buffer--banner-length 75
   "Width of a banner.")
 
@@ -42,6 +46,10 @@ version the release note it displayed")
 (defvar spacemacs-buffer--fresh-install
   (not (file-exists-p dotspacemacs-filepath))
   "Non-nil if this Emacs instance if a fresh install.")
+
+(defvar spacemacs-buffer--buttons-position nil
+  "Offset in characters between the edge of the screen and the beginning of the
+ home buffer buttons. Do not set this variable.")
 
 (defvar spacemacs-buffer-mode-map
   (let ((map (make-sparse-keymap)))
@@ -81,15 +89,15 @@ version the release note it displayed")
      (insert-file-contents file)
      (let ((banner-width 0))
        (while (not (eobp))
-	 (let ((line-length (- (line-end-position) (line-beginning-position))))
-	   (if (< banner-width line-length)
-	       (setq banner-width line-length)))
-	 (forward-line 1))
+         (let ((line-length (- (line-end-position) (line-beginning-position))))
+           (if (< banner-width line-length)
+               (setq banner-width line-length)))
+         (forward-line 1))
        (goto-char 0)
        (let ((margin (max 0 (floor (/ (- spacemacs-buffer--banner-length banner-width) 2)))))
-	 (while (not (eobp))
-	   (insert (make-string margin ?\ ))
-	   (forward-line 1))))
+         (while (not (eobp))
+           (insert (make-string margin ?\ ))
+           (forward-line 1))))
      (buffer-string))))
 
 (defun spacemacs-buffer/insert-banner-and-buttons ()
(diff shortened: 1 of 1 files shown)


deb0ch:
Damn, will take care of the conflicts very soon.

deb0ch:
By the way, I used `(window-width)` instead of `spacemacs-buffer--banner-length` that is usually used for centering because I don't really understand the point of it, and it looks like being just the same value.

syl20bnr:
Can you post before/after screenshots ?

deb0ch:
Hey, sure !

Before:
screenshot from 2016-09-04 21-35-34

After:
screenshot from 2016-09-04 21-36-58

screenshot from 2016-09-04 21-43-17

screenshot from 2016-09-04 21-46-52

When it wouldn't fit the screen anymore, it gets centered:

screenshot from 2016-09-04 23-08-08

JAremko:
@deb0ch  looks centered to me :+1:

core/core-spacemacs-buffer.el
```diff
@@ -590,7 +590,11 @@ border."
                    :mouse-face 'highlight
                    :follow-link "\C-m"
                    (propertize "Update Spacemacs" 'face 'font-lock-keyword-face))
-    (spacemacs-buffer//center-line)
+    (let ((len (- (line-end-position) (line-beginning-position))))
+      (spacemacs-buffer//center-line)
+      (setq spacemacs-buffer--buttons-position (- (- (line-end-position)
```
TheBB: `-` accepts more than one argument.

core/core-spacemacs-buffer.el
```diff
@@ -828,9 +826,53 @@ list. Return entire list if `END' is omitted."
...
+(defun spacemacs-buffer//center-startupify-lists ()
+  (let* ((lists-width (spacemacs-buffer//get-buffer-width))
+         (margin (max 0 (- spacemacs-buffer--buttons-position
+                           spacemacs-buffer-buttons-recent-lists-offset)))
+         (final-padding (if (< spacemacs-buffer--banner-length (+ margin lists-width))
+                            (max 0 (floor (/ (- spacemacs-buffer--banner-length lists-width) 2)))
+                          margin)))
+    (goto-char 0)
+    (while (not (eobp))
+      (progn (line-beginning-position)
```
TheBB: No need for `progn`.

core/core-spacemacs-buffer.el
```diff
@@ -828,9 +826,53 @@ list. Return entire list if `END' is omitted."
...
+          (if (< current-max line-length)
+              (setq current-max line-length)))
+        (forward-line 1))
+      current-max)))
+
+(defvar spacemacs-buffer--buttons-position nil
+  "Offset in characters between the edge of the screen and the beginning of the
+ home buffer buttons. Do not try to set this variable.")
+
+(defvar spacemacs-buffer-buttons-recent-lists-offset 25
```
TheBB: This seems to be independent of the actual width of the startup lists, i.e. not really centered. Am I missing something?

In any case this should probably not have `recent` in the name.

Would prefer to move the defvars to the top of the file if possible.

TheBB:
Please do excuse me for taking my time with this, but last time we went on a centering spree we dragged a ton of bugs in, for very little (IMO) gain, at a time when we're waiting for release. I'm not too keen on going through that again.

core/core-spacemacs-buffer.el
deb0ch: Changed, thanks for the tip !

core/core-spacemacs-buffer.el
deb0ch: Yes, it is independent of the width of the startup lists.  
The reason is that just centering based on the width of the startup lists would feel weird, so I did not just center it.  

The positioning algorithm is as follows:
- if there is enough space, the startup lists will have a fixed relative position to the actually centered elements (here the home buffer buttons)
- if a part of the startup lists would be off the screen just center them regularly, based on the width of the startup lists and the width of the window.

This way, the position of the beginning of the lines is consistent which makes reading easier, and the width of the buffer is always used at its best even when some entries are very long.

deb0ch: So, sometimes it is centered, sometimes it is adjusted.

TheBB: Yeah that's a fair point.

TheBB:

PR #7012, opened 2016-09-03
Buffer TS, :doc section, missing slash, 2 spaces

duianto:
"[N p] previous" was missing a slash between the "N" and "p" keys, and the listed keys were written with "," (commas) between the keys and a "." (period) at the end, that's different from most of the other transient states. By adding a ":doc" section then the slash can be added, the commas and period won't appear and the keys can be separated with 2 spaces instead of 1.


duianto:
This request was solved by #7529

PR #7014, opened 2016-09-03
org: Add org-expiry

darkfeline:
Add org-expiry package to be loaded lazily and autoload interactive
functions.

Provides no default configuration, but make it more convenient for users
to use the org-expiry package in their personal configuration.

### DIFF
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -23,6 +23,7 @@
     (ob :location built-in)
     (org :location built-in)
     (org-agenda :location built-in)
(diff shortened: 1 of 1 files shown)


darkfeline:
I've marked `org-expiry` as built-in. It ships with `org`, but not the default version of `org` shipped with Emacs stable (24.5). I'm not familiar with the use of `:builtin`, is this okay?

TheBB:
That's not a problem. Thanks! Cherry-picked in develop. You can safely delete your branch.

Issue #6631, opened 2016-07-19
Key remaps


buggymcbugfix:
OK, so I posted this in #6264, but not sure if anyone will notice it there, so posting again here:

I am a long-time Colemak user.

I would like to change as few keys as possible in spacemacs. I want to try a small cosmetic surgery that would implement the following behaviour:

![Colemak Vim](https://raw.githubusercontent.com/vimuel/screenshots/master/spacemacs/20160717_colemak_vim.png)

From what I understand, this would require simply swapping `j` and `k`.

How should I best do this?

d12frosted:
Hey,

I believe that `keyboard-layout` layer might help you with this. Just take a look how `dvorak` and `bepo` are [configured](https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Bintl/keyboard-layout/config.el#L34). You could do something very similar.

fmdkdd:
Ah!  A fellow Colemakian... ;)

I have a [private layer](https://github.com/fmdkdd/dotfiles/tree/master/spacemacs/.emacs.d/private/colemak-hjkl) that remaps HJKL to JKHL (so J is actually left and H is top in my configuration, unlike your drawing).

Basically, when I notice a wrong keybinding, I go to the source, copy its definition, and make my permutation into the code.  It's a bit of a pain, especially after a large Spacemacs update, but it has been stable for months now.

I did that before `keyboard-layout` was around, and it 'works for me', so I never bothered to change it.  Also, it seems `keyboard-layout` only permutes key for bindings defined by `evil`.  Many layers use HJKL as movement mnemonics, like org, magit, or web-mode.  When you are used to your permutation, you want that to be consistent across layers...

It's not as if we could just remap globally HJKL to JKHL, because sometimes these keys are _not_ used as movement keys (`C-j` splits lines for instance).  Ideally, all bindings that use HJKL as movement keys would be declared as such, and could all be overriden in a few lines of Elisp.

In the meantime, I hope my layer can save you some time.

vimuel:
Nice one, @fmdkdd — have installed your layer and it's working fine so far. Shall be testing it over the next few days and will keep you posted.

StreakyCobra:
@fmdkdd 

> Also, it seems keyboard-layout only permutes key for bindings defined by `evil`

Not true :-)

> Many layers use HJKL as movement mnemonics, like org, magit, or web-mode. When you are used to your permutation, you want that to be consistent across layers...

[...]

PR #7015, opened 2016-09-04
Add Colemak layout to keyboard-layout layer

lyallcooper:
The [Colemak keyboard layout](https://colemak.com/) is the third most popular English language keyboard layout (behind QWERTY and Dvorak), and also my keyboard layout of choice. So, I thought it'd be worth adding to the new keyboard-layout layer for myself and anyone else that might find it useful.

Overall, this is a pretty basic change that just adds the `hjkl` ↔ `hnei` mapping, specifies the home row keys and updates the README with a description of the layout.


ajgrf:
I'm glad I found this PR because I was about to start working on Colemak support myself. So thank you for getting it started! However, I think [these](https://github.com/ohjames/colemak) are smarter Colemak rebindings. So the rebinding table would look like this:

    (colemak . (("n" . "j")
                ("e" . "k")
                ("i" . "l")
                ("k" . "n")
                ("u" . "i")
                ("l" . "u")
                ("j" . "e"))))

They are still pretty minimal changes but it puts I, U, and N back into their QWERTY positions. With your bindings I would probably forever confuse I and U in both QWERTY and Colemak layouts!

What do you think? I know that there are a lot of Colemak users out there who have come up with their own incompatible vim bindings, so I hope this isn't too contentious.

lyallcooper:
That's an interesting remapping. Though I'd be hesitant to make it the default colemak mapping simply because I don't think it would be what people would expect. One solution could be to add a `colemak-alt` map, or perhaps let the user specify their own map.

I'll have a look to see if there's a nice way of doing it when I've got a little bit of time.

ajgrf:
I realized I may have over-emphasized swapping the insert/undo commands, but even if you don't do that I think putting the N bindings back into their QWERTY spot is still worthwhile.  So it would look like this:

    (colemak . (("n" . "j")
                ("e" . "k")
                ("i" . "l")
                ("k" . "n")
                ("l" . "i")
                ("j" . "e"))))

TheBB:
What's the relationship between this and https://github.com/syl20bnr/spacemacs/pull/7178?

allentiak:
#7178 has the same objective. Whereas #6631 contains the related discussion (!).
What about merging all these three issues?
I would say #7178 seems to be the best candidate, as it seems to be the most complete PR.

wbolster:
@ajgrf i agree with your preferred colemak remappings. my [evil-colemak-basics](https://github.com/wbolster/evil-colemak-basics) is an extended implementation of this idea.

syl20bnr:
Thank you !
There are multiple PR for colemak, see https://github.com/syl20bnr/spacemacs/commit/4461721e

Issue #7013, opened 2016-09-03
Evil Numbers Transient State, numeric argument, how to show [0-9] with colored numbers


duianto:
### Description
The Evil Numbers Transient State accepts a numeric prefix argument before pressing +/= or - to increase or decrease by that numeric amount. It should probably be indicated by adding "[0-9] amount" to the TS popup. 

### Question
How does one get the 0 and 9 numbers to show up with the bright red color: `#E12726` that indicates that the TS will stay open when the key is pressed.

It seems like the keys that are bound to run a function but not exit the TS,  are shown with the red color, and the bound keys that exit the TS are shown with the moderate blue color: `#4C8EC8`, here's how the keys are bound (q isn't bound in the current dev version but should probably be added):
https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Bspacemacs/spacemacs-evil/packages.el#L191
```
        :bindings
        ("+" evil-numbers/inc-at-pt)
        ("=" evil-numbers/inc-at-pt)
        ("-" evil-numbers/dec-at-pt)
        ("q" nil :exit t)) 
```

Since the 0 and 9 aren't bound keys but only listed in the ":doc" (see below) then that causes them to be shown with the default gray text color: `#B2B2B2`

Is it possible to just color the 0 and 9 numbers without them having to be bound to call a function, because the numbers need to be pressed several times for numeric arguments larger than a single digit.

### Underscores testing
When the 0, - (dash) and 9 keys are written with underscores around them like this:
`        :doc "\n[_+_/_=_] increase number  [_-_] decrease  [_0_-_9_] amount  [_q_] quit"`

Then the next time emacs is restarted, then a *Warnings* buffer opens with this text:
Warning (emacs): Unrecognized key: _0_
Warning (emacs): Unrecognized key: _9_

and when the Evil Numbers TS is opened, then the numbers are shown with underscores: `[_0_-_9_]`.

When the underscores are removed around the numbers:
`        :doc "\n[_+_/_=_] increase number  [_-_] decrease  [0-9] amount  [_q_] quit"`

and emacs is restarted, then the `*Warnings*` buffer doesn't open, and the numbers are shown without the underscores: [0-9]

It seems like the underscores are needed to show the correct colors for bound keys. For example if the underscores are removed around the q key: [q] then the q also is shown with the default gray text color.
<br>
If there's no clear way of coloring the 0-9 keys, then they probably could be left with the default gray text color.

[...]

PR #7020, opened 2016-09-04
Evil Nr TS, added 0..9 prefix and q quit

duianto:
This implements the changes in: #7013
Added "[0-9] amount" and "[q] quit", and separated the keys in the TS by 2 spaces.

### DIFF
--- a/layers/+spacemacs/spacemacs-evil/packages.el
+++ b/layers/+spacemacs/spacemacs-evil/packages.el
@@ -187,11 +187,12 @@
     (progn
       (spacemacs|define-transient-state evil-numbers
         :title "Evil Numbers Transient State"
-        :doc "\n[_+_/_=_] increase number [_-_] decrease"
+        :doc "\n[_+_/_=_] increase number  [_-_] decrease  [0..9] prefix  [_q_] quit"
         :bindings
         ("+" evil-numbers/inc-at-pt)
         ("=" evil-numbers/inc-at-pt)
-        ("-" evil-numbers/dec-at-pt))
+        ("-" evil-numbers/dec-at-pt)
+        ("q" nil :exit t)) 
       (spacemacs/set-leader-keys
         "n+" 'spacemacs/evil-numbers-transient-state/evil-numbers/inc-at-pt
         "n=" 'spacemacs/evil-numbers-transient-state/evil-numbers/inc-at-pt


duianto:
I just realized that the 1 to 8 numbers don't need to be bound. The reason for binding them to a function was just to have the [0-9] numbers show up in red.
Is it enough to just bind these two?
        ("0" digit-argument)
        ("9" digit-argument)

bmag:
I think this PR is misguided for two reasons.

First, during a hydra, `0`...`9` are bound to `hydra--digit-argument`. Binding them to `digit-argument` would possibly harm us down the line. I mean, there must be a reason for `hydra--digit-argument` existence.

Second, the doc is misleading. According to the doc, the valid amounts are 0-9, but that's not true. The digits are numeric prefix arguments, governed by the usual rules (modulo `hydra--digit-argument`), so these are also allowed: `150+` to increase by 150, `25-` to decrease by 25. Since these are the regular rules for prefix arguments, we can alter the doc to include something along the lines of:
> Pro-tip: use digit prefix arguments to increase/decrease by a bigger amount

duianto:
Maybe "[0-9] amount" could be written more generally as: 
[digits] prefix
[digit] prefix
[numbers] prefix
[number] prefix
[nr] prefix

I also stumbled over how to get the text to show up with the red color (won't close the TS), by binding the word in the [] (brackets) like this: `("digits" 'this-does-nothing)`.
Maybe there's a better way to bind it to something that does nothing.

bmag:
If coloring the digits requires binding them explicitly, I think it's not worth it. The colors have a meaning (blue - command exit hydra, red - doesn't exit), but digit arguments aren't normal commands. I don't think there's anything wrong with the digits being uncolorized. Besides, no coloring highlights the fact that the digits are "different" and don't do anything by themselves.

The `[numbers] prefix` seems the best option IMO. I'd suggest writing it as `[0..9] prefix`, but if you think one of the other phrasing is better then go for it, that would be better than if we started bikeshedding over this :wink:

duianto:
`[0..9] prefix` it is then, i'll keep them uncolored.

duianto:
Should i change title as well?

bmag:
> Should i change title as well?

I guess so.

>  should it state how it differs from the main spacemacs dev version?

You should squash the two commits into one, then change the commit message to say what are the changes from the dev version. It's easy to do this operation with magit, you need to:
1. open magit log: `SPC g s` to open status buffer, then `l l` to open log
2. go to the commit you want to rebase on (in this case, the earlier of the two commits you did)
3. start an interactive rebase: `r i`
4. squash the later commit onto the earlier one and commit: `s` to squash, `, c` to commit (there are key hints in the buffer)
5. in the commit buffer, re-write the message as necessary
6. after committing, you will need to force-push your branch (the `-f` flag)

There's a handy video here about using magit rebase: https://www.youtube.com/watch?v=mtliRYQd0j4, squashing starts at around 4:00 (the key-bindings are a bit different, because we use `evil-magit` on top of magit)

duianto:
I think it worked, thanks for the detailed explanation.

The only thing that got me a little confused was how to force-push, when i pressed P to open the 
magit-push-popup buffer, then at the top it says "Switches" and it shows two force commands:

-f Force with lease (--force-with-lease)
-F Force (--force)

i thought that maybe i would get to choose one of them when i pressed p, but when i tried it then i got this message:
```
Pushing to https://github.com/duianto/spacemacs
To https://github.com/duianto/spacemacs
 ! [rejected]        patch-2 -> patch-2 (non-fast-forward)
error: failed to push some refs to 'https://github.com/duianto/spacemacs'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
```
so i looked up the git push --help and searched for fast-forward. I happened to see that the title for that section was called \<refspec>, and when i looked through the options in the magit-push-popup buffer, then i saw in the Push section: "r explicit refspecs", and that got me to a prompt where i could type -f, and that seems to have worked :)

TheBB:
You can type -f from the git push popup directly. I can't tell from your story whether you figured that out or not but anyway, there you go.

duianto:
Even better, i didn't figure that out, now i see that it changes the color of the switches, Thanks

TheBB:

PR #7032, opened 2016-09-05
Move org-async-init.el into org/local dir

fintelkai:
Commit 1c4f685b13ab606f7ad71299547d22cf3eec6cc2 changed where org-export is looking for `org-async-init.el`. It replaced

```emacs
(let ((dir (configuration-layer/get-layer-property 'org :dir)))
  (setq org-export-async-init-file (concat dir "org-async-init.el")))
```

with

```emacs
(let ((dir (configuration-layer/get-layer-local-dir 'org)))
  (setq org-export-async-init-file (concat dir "org-async-init.el")))
```

Now, org-export is looking for `.../layers/+emacs/org/local/org-async-init.el` but the commit left `org-async-init.el` where it is (`.../layers/+emacs/org/org-async-init.el`). This makes org async export fail.

This PR simply moves `org-async-init.el` to where it's being looked for.


TheBB:

Issue #7042, opened 2016-09-06
Lisp error: (void-variable osx-use-dictionary-app)


hiberno:
I am running Spacemacs using the `develop` branch (mostly, because the `CSV` layer hasn't been merged into master yet) and pulled the latest changes this morning. Since then, Spacemacs aborts loading with the following error:

```
Symbol's value as variable is void: osx-use-dictionary-app
```

I suspect the error to stem from here: `layers/+spacemacs/spacemacs-languages/packages.el:13`

#### System Info :computer:
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.22
- Spacemacs branch: develop (rev. edf16ab)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
(auto-completion better-defaults c-c++ clojure colors csv emacs-lisp evil-cleverparens evil-commentary evil-snipe git github haskell html ivy java javascript latex markdown nixos notmuch org puppet rust scheme
                 (shell :variables shell-default-height 30 shell-default-position 'bottom)
                 shell-scripts spell-checking sql syntax-checking version-control weechat yaml)
```

StreakyCobra:
```
$ git blame layers/+spacemacs/spacemacs-language/packages.el
[...]
047b5722 (syl20bnr 2016-09-05 11:51:30 -0400 13)       '((define-word :toggle (not osx-use-dictionary-app))
[...]
```

:purple_heart: :grin:

kccai:
Just tested. This is fixed in #7044 :smile:

d12frosted:
Fix is in `develop`. Closing this issue since it's develop-branch only.

hiberno:
Thank you, that was quick! Now I know how to fix this kind of error, might open a PR myself next time. Well, thank you!

PR #7044, opened 2016-09-06
fix void variable when not using osx layer

nixmaniack:
Fix `Symbol's value as variable is void: osx-use-dictionary-app` when
not using osx layer.

Check: https://github.com/syl20bnr/spacemacs/commit/047b57223a8913699bc9ce52aa3046d934f40c9f#commitcomment-18905104

Fix: #7042


d12frosted:
Thanks for a quick fix. It's now in develop! ❤️

PR #7018, opened 2016-09-04
Common Lisp: enable fuzzy matching with comapny

tuhdo:

PR #7051, opened 2016-09-06
Common-lisp layer slime-company integration.

Chream:
Integration of slime-company contrib by default when using the common-lisp layer. Company auto complete will now work as well as all options from the slime-company (ie fuzzy matching). Also added a variable to disable/enable (enabled by default as no auto complete is working atm).

First contribution so tell me if I have done some conventions wrong and I will fix.

commit:

Add packages/init-functions. Add company conditional.

Initialize and define toggle variable.

### DIFF
--- a/layers/+lang/common-lisp/config.el
+++ b/layers/+lang/common-lisp/config.el
@@ -10,3 +10,8 @@
 ;;; License: GPLv3
 
 (spacemacs|define-jump-handlers lisp-mode slime-inspect-definition)
+
+;; Company integration
+(spacemacs|defvar-company-backends lisp-mode)
+(defcustom enable-slime-company t
+  "If non nil enables the slime-company package. DEFAULT: t")
--- a/layers/+lang/common-lisp/packages.el
+++ b/layers/+lang/common-lisp/packages.el
@@ -15,7 +15,16 @@
         ggtags
         helm
         helm-gtags
-        slime))
+        slime
+        auto-completion
+        (company :toogle lisp-mode-enable-slime-company)
+        (slime-company :toogle (configuration-layer/package-usedp 'company))))
+
+;; Hook company to common-lisp-mode
+(defun common-lisp/post-init-company ())
+
+;; Add the backend to the major mode specific backend list
+(defun common-lisp/init-slime-company ())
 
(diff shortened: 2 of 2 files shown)


layers/+lang/common-lisp/packages.el
```diff
@@ -15,7 +15,16 @@
...
         helm
         helm-gtags
-        slime))
+        slime
+        auto-completion
+        (company :toogle lisp-mode-enable-slime-company)
+        (slime-company :toogle (configuration-layer/package-usedp 'company))))
+
+;; Hook company to common-lisp-mode
+(defun common-lisp/post-init-company ())
```
bmag: You don't need an empty post-init function, it doesn't do anything.

layers/+lang/common-lisp/packages.el
```diff
@@ -15,7 +15,16 @@
         ggtags
         helm
         helm-gtags
-        slime))
+        slime
+        auto-completion
+        (company :toogle lisp-mode-enable-slime-company)
```
bmag: You have no company-specific configuration, so you don't need to list company in the packages list. `lisp-mode-enable-slime-company` should be part of slime-company's toggle.

layers/+lang/common-lisp/config.el
```diff
@@ -10,3 +10,8 @@
 ;;; License: GPLv3
 
 (spacemacs|define-jump-handlers lisp-mode slime-inspect-definition)
+
+;; Company integration
+(spacemacs|defvar-company-backends lisp-mode)
+(defcustom enable-slime-company t
```
bmag: We normally use `defvar` for layer variables, not `defcustom`. The name might need a `lisp-` or `spacemacs-` prefix as well, but I'm not sure we have a defined convention for the prefix of layer variables.

bmag:
@Chream can you take a look at #7018 and make your suggestions there? It also introduces `slime-company`, but doesn't use a layer variable.

Chream:
@bmag Thanks for the suggestions and help! Ill have a new look when I have some time. I have auto completion working now for now. Ill close this and post in the referenced PR.

AndreaOrru:
Not sure why this has not been merged... This is so useful.

Issue #7001, opened 2016-09-02
Spacemacs torrent client


JAremko:
It is surprising but it looks like there is no well developed Emacs package that adds something like transmission :scream: Or is there? It would be cool to have something like this in Spacemacs.

JAremko:
I found this one https://melpa.org/#/mentor 

Haven't tried it yet.

bmag:
here's another one to try: https://melpa.org/#/transmission

JAremko:
@bmag  Looks nice. Thanks!

JAremko:
ok. I'll make a new layer for transmission

JAremko:
I ran into a problem :confused:   holomorph/transmission#6

JAremko:
Solved in #7053

tidux:
These really should be set to auto-refresh by default.

PR #7053, opened 2016-09-06
Add transmission layer.

JAremko:
Solve #7001


layers/+tools/transmission/packages.el
```diff
@@ -0,0 +1,64 @@
...
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+(setq transmission-packages
+      '(transmission))
+
+(defun transmission/init-transmission ()
+  (use-package transmission
+    :config (progn
```
bmag: The package should be auto-loaded. That means using `:defer t`, and binding the entry points in `:init`. If the command `transmission` isn't auto-loaded, use the `:commands` directive to auto-load it.

JAremko: Thanks, I forgot to put `:defer`.  
But why I have to bind them in `:init` ?

d12frosted: > `(spacemacs/set-leader-keys "at" 'transmission)`

This one should be probably in init. Otherwise it will be never configured 😸

JAremko: @d12frosted Thank for clarifying :heart:  
I probably need  `use-package` checklist or a cheat sheet ,I always forget something.

JAremko:
:up:  :upside_down_face:

JAremko:
:up:  :open_umbrella:  :arrow_double_up:

JAremko:
Tested on Windows 7 and Linux. I would assume that it also works on Mac OS.

PR #7069, opened 2016-09-09
update scala layer readme javadoc variables

emanresusername:
changed `scala-use-java-doc-style` to `scala-enable-eldoc` to reflect what it said above in the doc

deleted the `scala-indent:use-javadoc-style` section cause i didn't see any references to it in the repo


TheBB:
The eldoc thing is fine, but I believe this variable is used in another package used by the scala layer, not by Spacemacs as such. So I'm reluctant to remove it just for this reason.

emanresusername:
appears to be fixed in current `develop` branch

PR #7081, opened 2016-09-10
Fix wrong filename in a layer file’s header comment

roryokane:
the header of `keybindings.el` in the `better-defaults` layer


d12frosted:

PR #7085, opened 2016-09-11
Use Magit 2.8's built-in fullscreen status display function

houli:
Magit 2.8(changelog here https://emacsair.me/2016/08/21/magit-2.8/) added it's own set of display functions one of which is a fullscreen display function that acts the same way as Spacemacs' version. These are documented here https://magit.vc/manual/magit/Switching-Buffers.html. This removes the need for Spacemacs to maintain its own version of a fullscreen status function.

### DIFF
--- a/layers/+source-control/git/packages.el
+++ b/layers/+source-control/git/packages.el
@@ -358,7 +358,7 @@ Press [_b_] again to blame further in the history, [_q_] to go up or quit."
         'spacemacs/magit-toggle-whitespace)
       ;; full screen magit-status
       (when git-magit-status-fullscreen
-        (setq magit-display-buffer-function 'spacemacs//fullscreen-magit)))))
+        (setq magit-display-buffer-function 'magit-display-buffer-fullframe-status-v1)))))
 
 (defun git/init-magit-gitflow ()
   (use-package magit-gitflow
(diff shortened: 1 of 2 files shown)


layers/+source-control/git/packages.el
```diff
@@ -358,7 +358,7 @@ Press [_b_] again to blame further in the history, [_q_] to go up or quit."
         'spacemacs/magit-toggle-whitespace)
       ;; full screen magit-status
       (when git-magit-status-fullscreen
-        (setq magit-display-buffer-function 'spacemacs//fullscreen-magit)))))
+        (setq magit-display-buffer-function 'magit-display-buffer-fullframe-status-v1)))))
```
TheBB: Maybe we can do away with the layer variable altogether and just document this one instead?

houli: Yes that's possible. Also those documentation sections in the README could also be cleaned up a bit as well to use the layer variable system rather than defining them in `user-init`

houli: Any thoughts about how to proceed with this PR?

TheBB:
Thanks! Cherry-picked in develop. Let's keep it like this for now. You can safely delete your branch.

PR #7088, opened 2016-09-11
auto load local .eslintrc

CodeFalling:
Load `.eslintrc` from project's root if existed.

### DIFF
--- a/layers/+frameworks/react/packages.el
+++ b/layers/+frameworks/react/packages.el
@@ -38,6 +38,14 @@
   (with-eval-after-load 'flycheck
     (dolist (checker '(javascript-eslint javascript-standard))
       (flycheck-add-mode checker 'react-mode)))
+  (defun react/use-eslintrc-from-project-root ()
+    (let ((rc-path (if (projectile-project-p)
+                       (concat (projectile-project-root) ".eslintrc"))))
+      (if (file-exists-p rc-path)
+          (progn
+            (message rc-path)
+          (setq flycheck-eslintrc rc-path)))))
+
   (defun react/use-eslint-from-node-modules ()
     (let* ((root (locate-dominating-file
                   (or (buffer-file-name) default-directory)
@@ -51,7 +59,7 @@
       (setq-local flycheck-javascript-eslint-executable eslint)))
 
   (add-hook 'react-mode-hook #'react/use-eslint-from-node-modules)
-
+  (add-hook 'react-mode-hook #'react/use-eslintrc-from-project-root)
   (spacemacs/add-flycheck-hook 'react-mode))
 
 (defun react/post-init-js-doc ()


AgIt0:
We should also make it configurable  somehow IMO, because a lot of projects I work on/I've seen keep the `.eslintrc` in a subfolder `assets` for instance

bleggett:
This is super-needed, but I think it belongs in the general Javascript layer.

robbyoconnor:
When will this awesomeness be merged?!?!

robbyoconnor:
When will this awesomeness be merged?!?! I agree that this needs to be in the javascript layer and not here.

layers/+frameworks/react/packages.el
```diff
@@ -38,6 +38,14 @@
...
     (dolist (checker '(javascript-eslint javascript-standard))
       (flycheck-add-mode checker 'react-mode)))
+  (defun react/use-eslintrc-from-project-root ()
+    (let ((rc-path (if (projectile-project-p)
+                       (concat (projectile-project-root) ".eslintrc"))))
+      (if (file-exists-p rc-path)
+          (progn
+            (message rc-path)
+          (setq flycheck-eslintrc rc-path)))))
+
```
robbyoconnor: Move to the `javascript` layer.

layers/+frameworks/react/packages.el
```diff
@@ -51,7 +59,7 @@
       (setq-local flycheck-javascript-eslint-executable eslint)))
 
   (add-hook 'react-mode-hook #'react/use-eslint-from-node-modules)
-
+  (add-hook 'react-mode-hook #'react/use-eslintrc-from-project-root)
```
robbyoconnor: Move to the `javascript` layer.

bleggett:
Looked into this more, and 

- Current versions of eslint automatically discover .eslintrc files by walking up the  folder tree from the file being checked.
- Current versions of flycheck just run `eslint --print-config .` which will return an exit code of 0 if eslint finds a config somewhere in the tree using the method above.
- Because of the above, the variable `flycheck-eslintrc` has been deprecated in flycheck.

So really, we shouldn't ever need to explicitly tell eslint or flycheck where the .eslintrc is, as long as it is somewhere in a parent folder or the project root, which is where eslint expects to find them.

TheBB:
Thank you for that @bleggett!

robbyoconnor:
@TheBB -- I'm pretty strong that this belongs in the `javascript` layer...not here...if this PR were to be accepted.

syl20bnr:
So this PR is not needed?

If so, please merge it and then correct it in a new commit.

robbyoconnor:
Shouldn't this be on the javascript layer though @syl20bnr?

CodeFalling:
As @bleggett said, this PR is not needed. Should we close this?

robbyoconnor:
Since it's in the wrong layer...yeah.

TheBB:
Closing then since not needed, no matter what layer it's in.

PR #7090, opened 2016-09-11
hint about where to put web-mode setq

hrldcpr:
Web-mode ignores settings set in `dotspacemacs/user-config` on the file opened at launch (though it picks them up for subsequent files), so it took me a while to figure out that I needed to put them in `dotspacemacs/user-init`. This might be nice to save some other people the anguish.

### DIFF
--- a/layers/+frameworks/react/README.org
+++ b/layers/+frameworks/react/README.org
@@ -82,7 +82,7 @@ Be sure to have the ~e4x~ option set to ~true~ on your ~.jsbeautifyrc~ here it i
 
 You may refer to the =web-mode= configuration for fine tuning the indenting behaviour.
 
-For example to have a consistent 2 spaces indenting both on =js= and =jsx= you may use these settings:
+For example to have a consistent 2 spaces indenting both on =js= and =jsx= you may use these settings in `dotspacemacs/user-init`:
 
 #+begin_src emacs-lisp
   (setq-default


d12frosted:
Quoting @NJBS 

> I don't understand why those settings need to be in `user-init`, and they seem to work fine on my end when put in `user-config`.

Does anyone else experience problems when placing configs in `user-config`?

hrldcpr:
Strange, I have quite a vanilla install of spacemacs so I'm not sure what would be causing this only for me… If you still can't reproduce I can try to reproduce it in a fresh spacemacs install in a docker image or something, to reduce other variables / so you can see for yourself.

NJBS:
If you can put together a list of reproduction steps and maybe an example file I'd be happy to try to reproduce 😄 !

Currently adding
``` elisp
   (setq-default
    ;; js2-mode
    js2-basic-offset 2
    ;; web-mode
    css-indent-offset 2
    web-mode-markup-indent-offset 2
    web-mode-css-indent-offset 2
    web-mode-code-indent-offset 2
    web-mode-attr-indent-offset 2)
```
to my `user-config` and opening a js file, css file, then html file and testing indentation in each of them gives me the indentation as set in my `user-config`.

hrldcpr:
Hi @NJBS sorry for the slow response.

The interesting part is that the issue only occurs when you open a jsx file from the command line, e.g. `emacs whatever.jsx` whereas things work fine if you open the file from within emacs.

Maybe that tip is enough for you to reproduce, but if not I'm able to reproduce it on a fresh Debian 8 (though I also get the same behavior on my Mac) as follows:

``` sh
apt update
apt install emacs-nox git
git clone https://github.com/syl20bnr/spacemacs ~/.emacs.d
emacs
```

Then add the `react` layer to `.spacemacs` and add the indent config to `dotspacemacs/user-config`:
``` el
  (setq-default
   ;; js2-mode
   js2-basic-offset 2
   ;; web-mode
   css-indent-offset 2
   web-mode-markup-indent-offset 2
   web-mode-css-indent-offset 2
   web-mode-code-indent-offset 2
   web-mode-attr-indent-offset 2
   )
```

Then open a jsx file (doesn't matter if it exists already or not):
``` sh
emacs test.jsx
```

If you write some code, you'll notice that the indent is 4 spaces.

And if you do `C-h v web-mode-code-indent-offset`, you'll get:
```
web-mode-code-indent-offset is a variable defined in `web-mode.el'.
Its value is 4
Local in buffer test.jsx; global value is 2
```

**However** if you instead open a jsx file with `C-x C-f test2.jsx` **or** place the config in `dotspacemacs/user-init`, then everything works as expected, and `C-h v web-mode-code-indent-offset` gives:

```
web-mode-code-indent-offset is a variable defined in `web-mode.el'.
Its value is 2
Original value was 4
Local in buffer test.jsx; global value is the same.
```

This only happens with jsx files, normal js is fine.

I hope that helps narrow it down, let me know if you have any questions. 👍

TheBB:
So this is about https://github.com/syl20bnr/spacemacs/pull/7202 actually. It's not a documentation problem IMO, and it certainly has nothing to do with web-mode.

hrldcpr:
Oh okay thanks, I'm glad it's a known bug.

I'm surprised more people don't encounter it; is it rare to open files from the command line?

Thanks for all your help, I'll close this now.

PR #7096, opened 2016-09-12
Add magit-repolist support

### LINKED ISSUE #7083: Magit Feature Request `magit-list-repositories`
I think it would be nice to have `<space>gL` binding for `magit-list-repositories`. It's trivial to add the binding, but where I got hung up is the buffer only responds to `emacs` mode bindings (ctrl-z then a return will open the status buffer). I'm not sure how to fix it.

In my `dotspacemacs/user-config()` I configured the feature with:

``` lisp
  (setq magit-repository-directories 
      '( "~/Development" ))
```

rphillips:
Adds Magit Repolist support.

Bindings:

`SPC gL`
`RET`
`g`

Fixes #7083

### DIFF
--- a/layers/+source-control/git/packages.el
+++ b/layers/+source-control/git/packages.el
@@ -158,7 +158,7 @@ Press [_b_] again to blame further in the history, [_q_] to go up or quit."
       (evilified-state-evilify-map magit-repolist-mode-map
           :mode magit-repolist-mode
           :bindings
-          (kbd "g") 'magit-list-repositories
+          (kbd "gr") 'magit-list-repositories
           (kbd "RET") 'magit-repolist-status)
       (unless (configuration-layer/package-usedp 'evil-magit)
         ;; use auto evilification if `evil-magit' is not used
--- a/layers/+source-control/git/packages.el
+++ b/layers/+source-control/git/packages.el
@@ -124,6 +124,7 @@
       (spacemacs/set-leader-keys
         "gb"  'spacemacs/git-blame-micro-state
         "gfh" 'magit-log-buffer-file
+        "gL"  'magit-list-repositories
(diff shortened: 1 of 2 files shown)


layers/+source-control/git/packages.el
```diff
@@ -154,6 +155,11 @@ Press [_b_] again to blame further in the history, [_q_] to go up or quit."
       (require 'git-rebase)
       ;; bind function keys
       ;; (define-key magit-mode-map (kbd "<tab>") 'magit-section-toggle)
+      (evilified-state-evilify-map magit-repolist-mode-map
+          :mode magit-repolist-mode
+          :bindings
+          (kbd "g") 'magit-list-repositories
```
nixmaniack: I think it should be bound to `gr` as we do in other buffers to _refresh_ the view. Binding it to `g` would override some evil bindings such `gg` to goto the top of buffer.

deb0ch:
Just tried to call the function `magit-list-repositories` and it just gave an empty page.

I guess it is because I did not set the options `magit-repository-directories' and `magit-repository-directories-depth'.

Would there be a nice default value to set that to ? like try to look in the entire home directory ? Or is it a bad idea ?

Maybe it would also be nice to bind `q` and `ESC` like in the other magit buffers ?

layers/+source-control/git/packages.el
rphillips: good call. tweaked in 386b0b1

rphillips:
@deb0ch `q` is bound already. I worry about a default directory, because I have _a lot_ of top level projects. It would be nice if someone knew more lisp, to display a warning if the directories are not set.

TheBB:
Thanks! Cherry-picked in develop. I made some small changes to the documentation. You can safely delete your branch.

PR #7098, opened 2016-09-12
axes.color_cycle is deprecated

Retorz:
It comes from an offical "UserWarning": axes.color_cycle is deprecated and replaced with axes.prop_cycle; please use the latter


TheBB:

PR #7101, opened 2016-09-13
Add open packages.el shortcut in spacemacs-help

CestDiego:
I find myself wanting to open the package.el and f3 is just too far away
in my keyboard. And I thought maybe this would be something people might
want. Also maybe we could have special binding for actions? like `C-1`
`C-2`, etc


TheBB:

Issue #4057, opened 2015-12-04
Spacemacs doesn't open file given as commandline argument


Immortalin:
When ` emacs <file name>` is ran, spacemacs goes directly to the home screen. The file that it was suppose to open shows up in the `Recent Files` section.

```sh
#### System Info
- OS: gnu/linux
- Emacs: 24.5.2
- Spacemacs: 0.105.0
- Spacemacs branch: develop (rev. c9de044)
- Distribution: spacemacs
- Layers:
```elisp
(auto-completion emacs-lisp clojure git go python elixir markdown org shell html javascript mu4e)
```

person808:
I'm pretty sure its caused by [this](https://github.com/syl20bnr/spacemacs/blob/develop/core/core-spacemacs.el#L83) line. I think doing something like this

```elisp
(unless file-argument-exists
    (switch-to-buffer (get-buffer-create spacemacs-buffer-name)))
```

would work, but I don't know how to detect a file argument (Note: `command-line-args` contains all arguments passed to emacs).

choppsv1:
A work around is to put the following into (defun dotspacemacs/user-config ()...

    (if inhibit-startup-screen
        (quit-window))

anakimluke:
Just letting you know I still get this bug when using the rolling release on emacs 27.1

lebensterben:
@anakimluke 

See https://github.com/syl20bnr/spacemacs/issues/15887

Issue #7016, opened 2016-09-04
*spacemacs* buffer disappears when using emacsclient


aroig:
#### Description :octocat:
Since a7d46742027, the `*spacemacs*` buffer disappears when connecting via emacsclient. As a side effect, this causes the following error message to pop up:

    Error (use-package): vi-tilde-fringe :init: No buffer named *spacemacs*

@ralesi any clue on what is going on?

#### Reproduction guide :beetle:
- Start Emacs in daemon mode: `emacs --daemon`
- open a new frame with emacsclient: `emacsclient -c`

*Observed behaviour:* :eyes: :broken_heart:
The `*spacemacs*` buffer is missing.

*Expected behaviour:* :heart: :smile:
There exists a `*spacemacs*` buffer.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.22
- Spacemacs branch: nil (rev. a7d4674)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(asm
 (auto-completion :variables auto-completion-enable-snippets-in-popup t auto-completion-enable-help-tooltip t auto-completion-return-key-behavior nil auto-completion-tab-key-behavior 'cycle auto-completion-private-snippets-directory "~/.spacemacs.d/snippets/" helm-mini-default-sources
                  '(helm-source-buffers-list))
 (colors :variables color-colorize-identifiers nil)
 finance git github
 (markdown :variables markdown-indent-on-enter nil)
 (org :variables org-enable-github-support t)
 (mu4e :variables mu4e-enable-mode-line nil)
 (rcirc :variables rcirc-spacemacs-layout-name "@rcirc" rcirc-spacemacs-layout-binding "i")
 jabber twitter
 (c-c   :variables c-c  -default-mode-for-headers 'c  -mode c-c  -enable-clang-support nil)
 syntax-checking asciidoc bibtex emacs-lisp extra-langs go haskell html ipython-notebook java javascript latex lua octave python restclient ruby shaders shell-scripts sql systemd theming vimscript yaml fasd spell-checking windows-scripts ab2-base ab2-clang ab2-devel ab2-latex ab2-org ab2-mu4e ab2-visual ab2-chat ab2-science ab2-audio)
```

deb0ch:
I was able to reproduce, and indeed it is coming from that particular commit.

Though it appears allright when creating it with `SPC b h`.

[...]

PR #7109, opened 2016-09-14
Fix initial-buffer-choice in daemon mode

TheBB:
After playing around a bit, this appears to check all the boxes.

- [x] Running `emacs` shows the home buffer.
- [x] Running `emacs <filename>` shows the file I asked for.
- [x] Running `emacs --daemon` and `emacsclient` shows the home buffer.
- [x] Running `emacs --daemon` and `emacsclient <filename>` shows the file I asked for.

Number three isn't working on current develop because `initial-buffer-choice` was set to nil to fix https://github.com/syl20bnr/spacemacs/issues/4057. However it can be set back later without ruining this behaviour.

This PR fixes the last remaining parts of https://github.com/syl20bnr/spacemacs/issues/7016, although I consider the release blocker to have already been fixed.

### DIFF
--- a/core/core-spacemacs.el
+++ b/core/core-spacemacs.el
@@ -120,6 +120,8 @@ the final step of executing code in `emacs-startup-hook'.")
     (spacemacs|do-after-display-system-init
      (kill-buffer (get-buffer spacemacs-buffer-name))
      (spacemacs-buffer/goto-buffer)))
+  ;; This is set to nil during startup to allow Spacemacs to show buffers opened
+  ;; as command line arguments.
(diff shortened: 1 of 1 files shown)


deb0ch:
I've been struggling so badly to try and find that one, I was ready to just rewrite the entire home buffer mechanism to try and fix it !

Glad that you fixed this one before I actually got to it.

I didn't think of the initial-buffer-choice mechanism. Didn't even know about its existence. Glad to learn something :smile:

TheBB:
Merged!

Issue #7105, opened 2016-09-13
Layouts refuse to auto save


gilbertw1:
#### Description
Layouts refuse to auto save on when closing Spacemacs or on the defined auto save interval.

#### Reproduction guide
- Start Emacs
- Create new perspective
- Close Emacs
- Start Emacs
- Layout was not auto saved.

*Observed behaviour:*
Any modifications to an existing layout are lost on restarting spacemacs.

*Expected behaviour:*
Any modifications to an existing layout should be saved automatically when closing the app or on the defined auto save interval, and should be visible on starting spacemacs.

#### System Info
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.22
- Spacemacs branch: master (rev. 9f9faa4)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(emacs-lisp
 (spacemacs-layouts :variables layouts-enable-autosave t layouts-autosave-delay 300)
 git markdown
 (org :variables org-enable-github-support t)
 games dash restclient
 (shell :variables shell-default-height 30 shell-default-position 'bottom)
 syntax-checking version-control xkcd auto-completion rust javascript
 (scala :variables scala-auto-start-ensime nil)
 haskell scheme
 (ranger :variables ranger-show-preview t ranger-cleanup-on-disable t)
 (clojure :variables clojure-enable-fancify-symbols t)
 themes-megapack github
 (evil-snipe :variables evil-snipe-enable-alternate-f-and-t-behaviors t)
 sql format-sql)
```

#### Backtrace
No Bactrace

Layouts have saved properly in the past, however after recently updating Spacemacs and packages I cannot get layouts to auto save anymore. 

I've tried removing and creating a completely clean emacs.d directory as well as switching to the 'develop' branch to no avail. I can place an old 'persp-auto-save' file (from the an old emacs.d directory) in the layouts cache directory and it properly loads the layouts on startup, but refuses to save any changes to the layouts. I can also manually save layouts with no problem. 

You can find my current .spacemacs config file [here.](https://gist.github.com/gilbertw1/e2582ab1bc17b0ec3530df49b49c72da) Any assistance or workarounds would be greatly appreciate as this is a rather large pain for me at the moment.

Thanks!

gilbertw1:
Looking throught the persp-mode github I found the following commit:

[...]

PR #7110, opened 2016-09-14
Run kill-emacs-query-functions when exiting

TheBB:
Fix https://github.com/syl20bnr/spacemacs/issues/7105

I have some better quit functionality in mind but this will be fine for now.

### DIFF
--- a/layers/+distributions/spacemacs-base/funcs.el
+++ b/layers/+distributions/spacemacs-base/funcs.el
@@ -588,12 +588,14 @@ dotspacemacs-persistent-server to be t"
 (defun spacemacs/kill-emacs ()
   "Lose all changes and exit Spacemacs"
   (interactive)
+  (run-hook-with-args-until-failure 'kill-emacs-query-functions)
   (setq spacemacs-really-kill-emacs t)
   (kill-emacs))
 
 (defun spacemacs/prompt-kill-emacs ()
   "Prompt to save changed buffers and exit Spacemacs"
   (interactive)
+  (run-hook-with-args-until-failure 'kill-emacs-query-functions)
   (setq spacemacs-really-kill-emacs t)
   (save-some-buffers)
   (kill-emacs))


layers/+distributions/spacemacs-base/funcs.el
```diff
@@ -588,12 +588,14 @@ dotspacemacs-persistent-server to be t"
...
   (setq spacemacs-really-kill-emacs t)
   (kill-emacs))
 
 (defun spacemacs/prompt-kill-emacs ()
   "Prompt to save changed buffers and exit Spacemacs"
   (interactive)
+  (run-hook-with-args-until-failure 'kill-emacs-query-functions)
   (setq spacemacs-really-kill-emacs t)
   (save-some-buffers)
   (kill-emacs))
```
bmag: Why do we call `save-some-buffers` and `kill-emacs` instead of just calling `save-buffers-kill-emacs`? That would also take care of running `kill-emacs-query-functions`.

layers/+distributions/spacemacs-base/funcs.el
```diff
@@ -588,12 +588,14 @@ dotspacemacs-persistent-server to be t"
 (defun spacemacs/kill-emacs ()
   "Lose all changes and exit Spacemacs"
   (interactive)
+  (run-hook-with-args-until-failure 'kill-emacs-query-functions)
```
bmag: `kill-emacs-query-functions` can ask the user if he/she really wants to quit Emacs, and in this case we just ignore the answer to the question and quit no matter what. That's weird.

bmag:
> this will be fine for now

I'm not sure `spacemacs/kill-emacs` really is fine, but if you think it is then ok.

layers/+distributions/spacemacs-base/funcs.el
TheBB: I don't know, I figured I'd fix this and solve that problem later. There is presumably a reason because elsewhere we *do* call `save-buffers-kill-emacs`.

TheBB:
You may be right. Would it be easier to add the hook back ourselves instead?

layers/+distributions/spacemacs-base/funcs.el
TheBB: You're right, I just copied the implementation from `save-buffers-kill-emacs` but I reliaze now that it's wrapped in a big `and`. That was confusing.

PR #7113, opened 2016-09-14
ycmd: improve search of nearest file in compilation database

deb0ch:
The old version wasn't thorough enough, as it only searched subdirectories recursively with `os.walk()`.

This version will search current directory recursively with `os.walk()` and then do the same recursively for parent directories until a match is found or we searched the database directory.


TheBB:

Issue #7116, opened 2016-09-14
Home buffer rerenders when opening helm


mahinshaw:
#### Description :octocat:
Home buffer rerenders when opening helm

#### Reproduction guide :beetle:
- Start Emacs
- On the Home huffer, initialize a helm buffer.
- This seems to be related to commit fd9ac4a

*Observed behaviour:* :eyes: :broken_heart:
- Watch the ui jump around while it tries to resize.

*Expected behaviour:* :heart: :smile:
- No jumping ui.

#### System Info :computer:
- OS: darwin
- Emacs: 25.1.1
- Spacemacs: 0.105.22
- Spacemacs branch: develop (rev. f373105)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(helm spacemacs-layouts
      (auto-completion :variables auto-completion-enable-help-tooltip t auto-completion-enable-sort-by-usage t)
      colors evil-commentary evil-cleverparens fasd
      (git :variables git-magit-status-fullscreen t)
      github nginx org restclient
      (shell :variables shell-default-height 30 shell-default-position 'bottom shell-default-shell 'multi-term)
      (spell-checking :variables spell-checking-enable-by-default nil)
      syntax-checking version-control vinegar clojure emacs-lisp erlang html markdown rust yaml)
```

deb0ch:
I don't see any different behaviour between before and after fd9ac4a on my end.

What do you mean exactly by "UI jumps" ?

mahinshaw:
If you open helm (I'm guessing ivy will cause this as well, but I don't use it), the text jumps on the screen.  When I say UI, I am more speaking to the UI components, as it is not only text that jumps. This only happens on the home buffer, which is why I am pointing to the aforementioned commit.  Furthermore, if I roll back to the prior commit, the "jumpy" behavior is not present.

TheBB:
Does the PR fix it?

mahinshaw:
@TheBB Yes. It does.

TheBB:
Cool, thanks. :+1:

mahinshaw:
And Thank You for all the work you do on Spacemacs!

TheBB:
Then this is fixed.

PR #7117, opened 2016-09-14
Additional home buffer fixes

TheBB:
CC @deb0ch. Does this make sense? Fixes https://github.com/syl20bnr/spacemacs/issues/7116

Rationale: Doesn't make sense to force redisplay every time the window configuration changes. Only when the width changes. I assume the condition should have checked `(window-width)` since the value of `--last-width` and `--banner-length` is the same (as per the `setq`), so the current condition will never fire. When I make these changes the redrawing correctly happens only when the width changes and not on every window configuration change.

Also don't redisplay every time even if nothing changes. That really shouldn't have to happen.


mahinshaw:
Works on my machine.

TheBB:
Merged.

Issue #7114, opened 2016-09-14
dotspacemacs/user-config not evaluated when running as a daemon


Stebalien:
#### Description :octocat:
Commit fd9ac4afce89d15130175d640e37194a8957b2ec causes `dotspacemacs/user-config` to not be evaluated on start when running emacs as a daemon.

#### Reproduction guide :beetle:
- Start Emacs

*Observed behaviour:* :eyes: :broken_heart:

None of the settings in `dotspacemacs/user-config` have been applied.

Unfortunately, I don't see any useful messages in `*Messages*`.

*Expected behaviour:* :heart: :smile:

The settings in `dotspacemacs/user-config` have been applied.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.22
- Spacemacs branch: develop (rev. fd9ac4a)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(helm better-defaults git github
      (shell :variables shell-default-shell 'eshell)
      org
      (auto-completion :variables auto-completion-tab-key-behavior 'cycle auto-completion-complete-with-key-sequence "jk" auto-completion-return-key-behavior nil)
      spell-checking syntax-checking emoji systemd emacs-lisp markdown latex bibtex csv yaml c-c   html haskell go javascript python csv asm
      (rust :variables rust-format-on-save nil))
```

TheBB:
Most likely there is something wrong with your user-config function. Mine is certainly being run, and I'd gather post people's, or we would be buried by reports about this already. Try replacing it with something that does nothing other than printing a message.

Stebalien:
I've already tried that; it doesn't run. However, re-loading the config `SPC f e R` causes it to be run. We're probably not buried in reports because:

1. Most people don't run spacemacs as a daemon.
2. The commit that introduced the bug was pushed last night (reverting this commit reliably fixes the bug).

---

 Anyways, I somehow managed to miss the fact that I had a `*Warnings*` buffer (I usually just look in `*Messages*`) with the following warning:

```
Warning (initialization): An error occurred while loading `/home/steb/.emacs.d/init.el':

Wrong type argument: wholenump, -29

To ensure normal operation, you should investigate and remove the
cause of the error in your initialization file.  Start Emacs with
the `--debug-init' option to view a complete error backtrace.
```

[...]

PR #7118, opened 2016-09-14
Never use negative widths when formatting the spacemacs buffer caption.

Stebalien:
Fixes: #7114


TheBB:
Thanks! Cherry-picked in develop (and fixed indentation). You can safely delete your branch.

PR #7138, opened 2016-09-18
Fix typo in nyan cat var docstring

eapolinario:
Simple typo: packges -> packages.


rski:
This is also present in `layers/+tools/evernote/config.el`

d12frosted:

@rski on develop branch there is no `config.el` in `evernote` layer + `evernote` layer moved from `+tools` 😸

Issue #6300, opened 2016-06-11
ivy layouts buffer switching doesn't respect layout


darkfeline:
`SPC l b` (switch to buffer respecting layouts) doesn't work with `ivy`.

The relevant function in the `ivy` layer:

``` emacs-lisp
(defun spacemacs/ivy-spacemacs-layout-buffer ()
  "Switch to layout buffer using ivy."
  (interactive)
  (let (ivy-use-virtual-buffers)
    (with-persp-buffer-list ()
                            (call-interactively 'ivy-switch-buffer))))
```

However, `ivy-switch-buffer` doesn't use `buffer-list` internally, so `with-persp-buffer-list` here doesn't do anything.

d12frosted:
Hm... indeed. But I think previously it worked. I mean, I remember `SPC l b` helping me a lot compared to `SPC b b`.

ahyatt:
FWIW, there is a gist with code to fix this linked from persp-mode: https://gist.github.com/Bad-ptr/1aca1ec54c3bdb2ee80996eb2b68ad2d#file-persp-ivy-el

bmag:
@ahyatt thanks, that looks promising.

TheBB:
Should be fixed in master.

PR #7142, opened 2016-09-19
layouts: persp-mode changed variable to enable ido-hooks

aroig:
Since persp-mode 2.5 the way to enable ido hooks changed again. Without this, SPC b b lists all buffers instead of persp-local ones.


TheBB:
Fixes https://github.com/syl20bnr/spacemacs/issues/6266 and https://github.com/syl20bnr/spacemacs/issues/6300

TheBB:
At least, I think it does.

aroig:
> Fixes #6266 and #6300

Yeah, I think those two were already fixed by #6876 (not sure about #6300). Unfortunately, it was broken again by upstream.

bmag:
Nope, doesn't fix #6300.

TheBB:

PR #7153, opened 2016-09-20
Yaml layer: enable flycheck

rski:


layers/+lang/yaml/packages.el
```diff
@@ -9,14 +9,19 @@
...
 (defun yaml/post-init-company ()
   (spacemacs|add-company-hook yaml-mode))
 
+(defun yaml/post-init-flycheck ()
+  (spacemacs/add-flycheck-hook 'yaml-mode))
+
 (defun yaml/init-yaml-mode ()
   "Initialize YAML mode"
   (use-package yaml-mode
+    :defer t
```
TheBB: `:mode` implies `:defer`, so no need for this.

rski: oh, i see, one sec then

robbyoconnor:
@TheBB -- actually all flycheck checkers require external tools. Literally all flycheck does is parse the output from the various tools...

layers/+lang/yaml/config.el
```diff
@@ -1,4 +1,4 @@
-;;; config.el --- YAML Layer packages File for Spacemacs
+;;; config.el --- YAML Layer configuration File for Spacemacs
```
d12frosted: Thanks for fixing this line. But it should be `config` 😸

rski: I can fix it, but I just looked it up from other layers:
```
~/Code/spacemacs/spacemacs on yaml_flycheck ⌚ 20:34:53
$ ag "config file"| wc -l
11

~/Code/spacemacs/spacemacs on  yaml_flycheck ⌚ 20:35:00
$ ag "configuration file"| wc -l
67
```

d12frosted: Hm... indeed 😸

d12frosted:

PR #7157, opened 2016-09-21
Re-map "pp" to counsel-projectile-switch-project

asok:
This is so it behaves the same as the default setup that uses `helm`.


d12frosted:
Oh ya, behaviour of `counsel-projectile` was changed recently. Didn't update yet 😸

d12frosted:

PR #7162, opened 2016-09-21
Update layer docs for packages.el to note where should local packages reside

xificurC:
The information about location of local packages is noted in the parent section. While this can be considered enough I kept re-reading `packages.el` and couldn't find the information. Adding the information here as well makes the section more self-contained.


robbyoconnor:
amend this commit to actually say what you did...this isn't clear.

xificurC:
@robbyoconnor is that better?

robbyoconnor:
Yeah and maybe update the title :)

xificurC:
@robbyoconnor done

TheBB:

PR #7163, opened 2016-09-21
Added ensime folder to .gitignore

brandondtb:
Ensime creates the folder "ensime" in .emacs.d on start to store local classpath info. Added it to .gitignore so that it doesn't dirty the repo state.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,6 +44,7 @@ python-*-docs-html
 /session.*
 /srecode-map.el
 /recentf
+ensime/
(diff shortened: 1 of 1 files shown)


d12frosted:
There should be a way to change default location this folder. Once you find a way to change it, please place it under `spacemacs-cache-directory`.

Compro-Prasad:
@brandondtb Can find out what variable is pointing to the `ensime` direcory?

PR #7070, opened 2016-09-09
python: really start REPL also in emacs 25.1

bmag:
Make `python-start-or-switch-repl` start a REPL when necessery also in Emacs 25.1. Fixes #6628.

The new version is supposed to work in Emacs 24.4, 24.5 and 25.1. For any daily python users: please confirm that this change doesn't break `SPC m '` if you're on 24.4 or 24.5, or confirm that it makes `SPC m '` work if you're on 25.1.

Edit: this PR also fixes #2872

PR #7166, opened 2016-09-22
fix python-start-or-switch-repl bug

### LINKED ISSUE #2872: SPC-m-s-i python-start-or-switch-repl does not start a REPL
`python-start-or-switch-repl` does not start a REPL

> python-shell-get-process-or-error: No inferior Python process running.

If I start a shell with `Ctrl-c Ctrl-p` then `SPC-m-s-i` switches to the REPL

I'm on the `develop` branch

akirakyle:
Fixes #2872
`python-shell-switch-to-shell` doesn't create a new python process if one doesn't already exist. Thus when `python-start-or-switch-repl` was called with `SPC m s i` without an existing python process running somewhere, you would get the error `No inferior Python process running.`

### DIFF
--- a/layers/+lang/python/packages.el
+++ b/layers/+lang/python/packages.el
@@ -311,8 +311,18 @@
       (defun python-start-or-switch-repl ()
         "Start and/or switch to the REPL."
         (interactive)
-        (python-shell-switch-to-shell)
-        (evil-insert-state))
+        (if (python-shell-get-process)
+            (progn
+              (python-shell-switch-to-shell)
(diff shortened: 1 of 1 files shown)


bmag:
Thanks for the contribution, however there's already an open PR (disclaimer: I wrote it) to fix the same problem, see #7070.

akirakyle:
Ah, I didn't see that! I was getting quite annoyed at `SPC m s i` not starting a new python process. Hopefully your PR gets merged in soon.

PR #7178, opened 2016-09-23
Add Colemak layout to keyboard-layout layer

fmdkdd:
This commit remaps the HJKL movement keys to JKHL.

See https://colemak.com/ for a description of the Colemak layout.

See #6631 for a background discussion.
### Remaining issues

I would consider this layour useable as is, but far from complete. 
At the very least, this coulp help Colemak users of Spacemacs to get something going without diving into too much Elisp.

Here are a couple of annoying issues that I did not manage to fix yet:
- Lost the binding to `evil-join` which was `J` by default.  Now `J` maps to `evil-window-top` and `evil-join` is unbound.
- In magit, `H` and `K` are correctly rebound to `evil-previous-visual-line` and `evil-next-visual-line`, but not when entering visual mode, where `H` becomes `evil-lookup` and `K` becomes `evil-previous-visual-line`.  This is puzzling to me,  as they are correctly bound in visual mode outside of magit.

cc @StreakyCobra.

### DIFF
--- a/layers/+intl/keyboard-layout/README.org
+++ b/layers/+intl/keyboard-layout/README.org
@@ -156,6 +156,23 @@ found at the [[https://colemak.com/][Colemak website]].
 
 [[file:img/colemak-layout.png]]
 
+The configuration `colemak-jkhl` reuses HJKL as movement keys, but permuted as
+JKHL:
+- ~j~ is now left
+- ~k~ is now down
+- ~h~ is now up
+- ~l~ is right, unchanged
+
+While not on the home row, the keys are still easily accessible and do not
+conflict with any other ~evil~ binding.
+
+Additionally, in normal and motion states, ~L~ is bound to ~evil-lookup~ while
+~H~ and ~K~ are respectively bound to ~evil-window-top~ and
+~evil-window-bottom~.
+
+*Note*: for technical reasons (see [[https://github.com/syl20bnr/spacemacs/pull/7178#issuecomment-249360301][GH-7178]]), the ~kl/pre-config-evil~ and
+~kl/post-config-evil~ hooks will be run /twice/.
+
 * Package Configurations
 The available configurations are:
 
--- a/layers/+intl/keyboard-layout/README.org
+++ b/layers/+intl/keyboard-layout/README.org
@@ -156,22 +156,17 @@ found at the [[https://colemak.com/][Colemak website]].
 
 [[file:img/colemak-layout.png]]
 
-The configuration `colemak-jkhl` reuses HJKL as movement keys, but permuted as
-JKHL:
-- ~j~ is now left
-- ~k~ is now down
-- ~h~ is now up
-- ~l~ is right, unchanged
+This layer offers three flavors of Colemak bindings:
 
-While not on the home row, the keys are still easily accessible and do not
-conflict with any other ~evil~ binding.
+- =colemak-hnei= remaps ~HJKL~ to ~HNEI~, keeping the same key location; useful
+  for people used to the ~HJKL~ scheme on a Qwerty keyboard.
+- =colemak-neio= remaps ~HJKL~ to ~NEIO~, shifted one key to the right for
+  easier access.
+- =colemak-jkhl= remaps ~HJKL~ to ~JKHL~, permuting the direction of the keys
+  without disturbing any other binding.
 
-Additionally, in normal and motion states, ~L~ is bound to ~evil-lookup~ while
-~H~ and ~K~ are respectively bound to ~evil-window-top~ and
-~evil-window-bottom~.
-
-*Note*: for technical reasons (see [[https://github.com/syl20bnr/spacemacs/pull/7178#issuecomment-249360301][GH-7178]]), the ~kl/pre-config-evil~ and
-~kl/post-config-evil~ hooks will be run /twice/.
+*Note*: for technical reasons (see [[https://github.com/syl20bnr/spacemacs/pull/7178#issuecomment-249360301][GH-7178]]), when using =colemak-jkhl=, the
+~kl/pre-config-evil~ and ~kl/post-config-evil~ hooks will be run /twice/.
 
 * Package Configurations
 The available configurations are:
--- a/layers/+intl/keyboard-layout/config.el
+++ b/layers/+intl/keyboard-layout/config.el
@@ -15,7 +15,8 @@
 
 (defvar kl-layout 'dvorak
   "The keyboard-layout to use. Possible values are
-`dvorak',`bepo', `colemak' and `colemak-jkhl'.")
+`dvorak',`bepo', `colemak-hnei', `colemak-hnei` and
+`colemak-jkhl'.")
 
 (defvar kl-enabled-configurations nil
   "If non nil, `keyboard-layout' will enable configurations only
@@ -42,15 +43,6 @@ case.")
              ("j" . "t")
              ("k" . "s")
              ("l" . "c")))
-    (colemak . (("n" . "h")
-                ("e" . "j")
-                ("i" . "k")
-                ("o" . "l")
-                ;;
-                ("h" . "n")
-                ("j" . "e")
-                ("k" . "i")
-                ("l" . "o")))
     (dvorak . (("h" . "h")
                ("t" . "j")
                ("n" . "k")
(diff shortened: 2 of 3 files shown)


StreakyCobra:
Looks fine :-) I can't really test colemak, but you know better than me if colemak hack/changes are needed! So feel free to do whatever is needed in `:colemak`, you (theoretically) can't broke the other layers by changing these sections.

fmdkdd:
@StreakyCobra Any idea why `evil-join` would end up unbound?  Since this is simply a permutation, I would expect the function to end up *somewhere* (preferably on `J`).  Otherwise, I guess I can always force these bindings to what I expect.

StreakyCobra:
I had unanswered question like this all the time. More or less all `HACK` comments for `bepo` are due to mysterious reasons that I wasn't able to identify :sweat_smile: 

Also, it may be nice to update the README with a colemak section to let users know that it's supported

fmdkdd:
@StreakyCobra Well, I can't say my own layer to rebind HJKL wasn't hackish as well.  I'll try to get around the two remaining issues and update the README.  Please wait before merging.

fmdkdd:
I've restored `evil-join` and fixed `h` and `k` in magit visual mode, and I've added a Colemak section to the README.

I've also managed to rebind `H`, `K` and `L` in normal mode.  Annoyingly, `L` is overridden by the `spacemacs-bootstrap` layer, and even with a spacemacs hooks on evil post-config, I couldn't redefine it.  I used a `post-init` function instead, but I used `kl|config` twice on the configuration `evil`.  Will that cause any issue?

Other than that, I would consider that configuration decent for a colemak user.  There are bindings missing, for `evil-window` or `org`, but  I don't have the energy to deal with them right now.

StreakyCobra:
> Will that cause any issue?

The configuration name is used for 2 purpose:

- Identify and run the associated `pre-config-<name>` and `post-init-<name>` (see https://github.com/syl20bnr/spacemacs/tree/develop/layers/%2Bintl/keyboard-layout#addoverride-key-bindings)
- Enable or disable the tweak by using it in `kl-enabled-configurations` respectively `kl-disabled-configuations`

The second one is not problematic, and maybe eve useful, because all `evil` related bindings will be activated/disabled at once with only `evil`. The first one *may* be problematic in some case, because the `pre/post-config-<name>` functions will be ran twice. If the users use these functions only for forcing a binding, no problem will arise, but if they are using functions with side effects, like swapping two keys, this will be applied twice and have no effect.

> Other than that, I would consider that configuration decent for a colemak user. There are bindings missing, for evil-window or org, but I don't have the energy to deal with them right now.

Yes, it doesn't have to be perfect, `dvorak` and `bepo` layouts are not fully functional as well. But from an user perspective is always better than nothing at all, especially regarding the complexity of the task.

I hope your experience with the `keyboard-layout` was good, and that it made it easy to implement new keyboard-layout support.

fmdkdd:
@StreakyCobra  I've added a comment and a note to the README indicating the quirk with hooks running twice.  Thanks for catching that.

Well adding the basic permutation through `keyboard-layout` was certainly a breeze.  But then adding the rest was harder than expected.  One thing that may be surprising is that you do not get any output of what was rebound by `kl/correct-keys` and other functions.  So you do not really know what effect the functions had, if any.  Next time I work on it, I will definitely add `message` calls inside these functions ;)

In my own [colemak-hjkl](https://github.com/fmdkdd/dotfiles/blob/master/spacemacs/.emacs.d/private/colemak-hjkl/packages.el) layer, the code is certainly more tedious, but also more straightforward: you see every binding redefinition, and there are no surprises.  Still, I will be adding a link `keyboard-layout` now that it supports Colemak, as it's a worthy alternative.

TheBB:
What's the relationship between this and https://github.com/syl20bnr/spacemacs/pull/7015?

TheBB:
This PR sems more complete, but there are still some things in the other one that aren't here. Could you perhaps pick the remainder from there?

fmdkdd:
Oh, I didn't see #7015, sorry for that.  One immediate difference is that the other PR maps `hjkl` to `hnei` (home row on right hand), while this one maps `hjkl` to `jkhl` (not home row, but not far, and doesn't move any other binding since it's just a permutation).

So the question would be, which one do we keep?

TheBB:
In other words, this one is faithful to the physical locations of the keys? In that case I'd prefer this one. As far as missing content is concerned I was thinking more about the avy keys and such.

ajgrf:
I made some ascii diagrams to help you compare the proposals as someone who doesn't know Colemak. This one looks like this:

    q w f p g  ← → u y ;
     a r s t d  ↑ n e i o
      z x c v b  ↓ m , . /

It doesn't rearrange any keys except hjkl, but as a result you lose out on Vim muscle memory.

For the next diagrams, the capital labels represent the non-direction keys that got rebound. They show which qwerty key it will act like instead of which letter gets typed in insert mode. For example, the key marked N below is actually the Colemak j key, but it will behave like the qwerty N key in normal mode.

So here's the other PR:

    q w f p g *N*I*u y ;
     a r s t d  ← ↓ ↑ → o
      z x c v b *E*m , . /

As you can see this one keeps the direction keys intact. New Colemak converts will find it messes with their muscle memory much less. I think it's better, but in the other discussion I have been advocating for the following tweak to put the "find next" command back into its qwerty position (by just switching E and N):

    q w f p g *E*I*u y ;
     a r s t d  ← ↓ ↑ → o
      z x c v b *N*m , . /

I also kind of want the following tweak as well, simply because I switch between Colemak and QWERTY semi-regularly and anyone who does the same will constantly confuse "undo" with "insert" in both layouts unless they have this tweak:

    q w f p g *E*U*I*y ;
     a r s t d  ← ↓ ↑ → o
      z x c v b *N*m , . /

However I recognize that probably not everyone will want that.

Anyway, I hope that helps and it wasn't too confusing.

StreakyCobra:
You can define two different keyboard layout variants, like name one "colemak-hnei" and the others "colemak-jkhl", each one with their own ":colemak-…" properties block in the configs.

fmdkdd:
Well if that's an option, then yes, let's keep both.   At the end of the day, it's a matter of individual preference.

I'll update this PR to call the layout `colemak-jkhl`.

fmdkdd:
@TheBB I've updated this PR to integrate the missing pieces (avy and ace-window keys), and I've also named the keyboard configuration `colemak-jkhl`, so there's room to add alternate configurations for colemak, should @ajgrf or @lyallcooper want to add them.

fmdkdd:
Also, you can probably close #6631 if this is merged (to help with the cleanup!).

syl20bnr:
Thank you :-)

I guess we can let the neiop variant be named colemak as it is the default "scheme" for the other layout. So we get colemak and colemak-jkhl.

allentiak:
Hi,
I also use Colemak.

From what I can see, there are two different aspects proposed here. One concerns the movement keys. The other one concerns other shortcuts.

At least for now, I would only change the movement keys. In this regard, I think the less shocking change would be to use H-N-E-I as "left-down-up-right". (Even though I would find it most pleasant to use N-E-I-O, à la Dvorak.)

Even if I don't have a strong opinion concertning the keys for the other shortcuts, in principle I would leave them as they are (as long as they don't interfere with any existing shortcut, of course).

Just my opinion.

wbolster:
Evil users using Colemak may be interested in my evil-colemak-basics package, designed as a smart hybrid between Colemak and Qwerty. The docs explain the rationale quite extensively.

wbolster:
Evil users using Colemak may be interested in my [evil-colemak-basics](https://github.com/wbolster/evil-colemak-basics) package, designed as a smart hybrid between Colemak and Qwerty. The docs explain the rationale quite extensively.

syl20bnr:
@fmdkdd  I merged two other PRs regarding Colemak layout. Can you rebase this PR on top of develop ?

fmdkdd:
@syl20bnr Sure.  That would bring the count of Colemak bindings to 3 in `kl--base-rebinding-maps`.  Currently in develop, two of the bindings are named `colemak`, which means only the first one will be picked up.  I could rename the two existing flavours using the scheme:

`colemak-neio` `colemak-hnei` `colemak-jkhl`

and succinctly present the flavours in the README.

fmdkdd:
@syl20bnr Should be good.

TheBB:

I had to do a fair bit of conflict resolution still. Maybe it would be good if you double-checked the result.

fmdkdd:
@TheBB As far as as colemak variants are concerned, LGTM.

allentiak:
Just curious... Was this merged into the 200.10 release? I keep on rebasing my own colemak variant on top of checkversion/master...

duianto:
@allentiak The hotfix only fixed two things:
- Fix Helm error when selecting a candidate from a Helm buffer (thanks to bmag)
- Fix window number assignation for Neotree buffer window (thanks to duianto)

source: https://github.com/syl20bnr/spacemacs/blob/master/core/info/release-notes/0.200.10.txt
and: https://github.com/syl20bnr/spacemacs/commits/master

allentiak:
So... Should I prepare a PR?

duianto:
@allentiak Sure, the other Colemak users might also like your variant. Just remember to make the PR against the develop branch.

source: https://github.com/syl20bnr/spacemacs/blob/master/CONTRIBUTING.org#pull-request

PR #7184, opened 2016-09-24
update macOS installation instructions

d12frosted:
Mostly to reflect latest changes in emacs-plus formulae. I thought that life is too short and too interesting to typo so much options.

Also I have changed OS X -> macOS. Let's get used to it 😸


d12frosted:

PR #7189, opened 2016-09-24
fix broken macOS paths for homebrew service

tfiala:
When Spacemacs is run as a macOS homebrew service, the window-system at startup time is not set as it is running essentially headless.  This causes the very-early call to exec-path-from-shell, which was checking for macOS via window-system, to fail the check.  This is pretty much game over for macOS and user-customized paths, as the exec paths are not updated and any extra user-path utilities (e.g. aspell, gls) won't be found.

This change switches over to using spacemacs/system-is-mac and spacemacs/system-is-linux as the primary checks, and keeps the window-system check for 'x in case other Unix variants are getting exec-path set via that check.

### DIFF
--- a/layers/+distributions/spacemacs-base/packages.el
+++ b/layers/+distributions/spacemacs-base/packages.el
@@ -177,7 +177,9 @@
 
 (defun spacemacs-base/init-exec-path-from-shell ()
   (use-package exec-path-from-shell
-    :init (when (memq window-system '(mac ns x))
+    :init (when (or (spacemacs/system-is-mac)
+                    (spacemacs/system-is-linux)
+                    (memq window-system '(x)))
             (exec-path-from-shell-initialize))))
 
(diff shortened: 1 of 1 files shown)


tfiala:
I'm a bit confused on which branch to base this on.  I used the 'develop' branch/tag as the initial PR request header message said to not use master and to use the 'development' branch (which isn't a named branch in the repo).  However, I see a lot of activity on the master branch, and the develop seems quite stale.

Please let me know if I need to rebase this against a different one.

Thanks for all the excellent work on Spacemacs!

nixmaniack:
`develop` is *the* development branch of spacemacs so you have done it right.  `master` might get hotfixes in between releases hence the activity. `develop` is always cutting-edge `master` might get stale.

tfiala:
Thanks, @nixmaniack!

TheBB:

PR #7192, opened 2016-09-25
Add phoenix layer

AgIt0:
- Remove elixir's layer project prefix, since it was used only for
  one binding which would be better suited in tests prefix
- Move `alchemist-project-find-test` to test prefix
- Add `phoenix` prefix and declare bindings


dongcarl:
Are there any updates on this?

syl20bnr:
I slowly look  at new layer, it will take some time until it lands in the develop branch. In the meantime you can pull this PR locally. Thank you for your patience.

syl20bnr:
I added a convention for frameworks key bindings: https://github.com/syl20bnr/spacemacs/commit/90452329

Should go under `SPC m f`, if you can update your PR accordingly it would be wonderful!

cc @sdwolfz (as you worked on this at some point).

sdwolfz:
Yes, `SPC m f` sounds great.

AgIt0:
@syl20bnr updated, looks a bit weird with current defined bindings(most of them are just `Find X`) but I think that's ok

sdwolfz:
@AgIt0 do you happen to have a sample project I could try this layer on before merging?

AgIt0:
@sdwolfz not an up to date one sadly if you feel like it there is this one that's simple and you can check out https://github.com/AgIt0/newsletter_phoenix
otherwise you can always try with https://github.com/gothinkster/elixir-phoenix-realworld-example-app which is fairly up to date.

Just keep in mind you would need elixir installed and all the packages that are required for the given project otherwise `alchemist` won't work

Issue #3808, opened 2015-11-14
emacs file-name


kvaneesh:
This doesn't result in us showing file-name buffer. Instead we always default to spacemacs buffer.

TheBB:
If anyone else have difficulty understand (like myself), he is saying that giving a filename as an argument to `emacs` on the command line opens the file, but doesn't show it.

Probably related to #3233.

kvaneesh:
That is correct. Thanks for explaining it better.

fleimgruber:
I can confirm that this is fixed at least in current develop.

Issue #3233, opened 2015-10-02
dotspacemacs/user-config run after file is loaded


beojan:
When Emacs is started with, for example `emacs main.cpp`, `dotspacemacs/user-config` is run after the file is loaded, so c++-mode hooks added in `dotspacemacs/user-config` don't get run.

syl20bnr:
I think this is because of the changed hook from `after-init-hook` to `emacs-startup-hook`.

emacs-startup-hook:

> Normal hook run after loading init files and handling the command line.

after-init-hook

> Normal hook run after initializing the Emacs session.

cc @vkz

StreakyCobra:
@syl20bnr Maybe you want to milestone this on for 0.105?

syl20bnr:
Pretty sure this bug is fixed in 0.200 :-D

hrldcpr:
As discussed in #7202, as far as I can tell this is no longer fixed or was never actually fixed.

PR #7202, opened 2016-09-27
Process user-config before command line arguments

BinaryKhaos:
emacs-startup-hook is run after the command line arguments are being processed (and not at all if in batch mode). This has several undesired side-effects. If Emacs is started with file arguments, user-config is being run after the respective major mode has been initialized. Thus, variables set and code defined in user-config will not have the desired effect in that particular scenario. Also, in batch mode, user-config will not be evaluated at all which in some cases is not what the user might want and/or expect.

This refactors the handling of user-config to the after-init-hook and introduces another needed setup function for init.el.

### DIFF
--- a/core/core-spacemacs.el
+++ b/core/core-spacemacs.el
@@ -174,8 +174,27 @@ defer call using `spacemacs-post-user-config-hook'."
       (funcall func)
     (add-hook 'spacemacs-post-user-config-hook func)))
 
+(defun spacemacs/setup-after-init-hook ()
+   "Add post init processing (before command line arguments are being processed)."
+  (add-hook
+   'after-init-hook
+   (lambda ()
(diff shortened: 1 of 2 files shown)


Miciah:
This issue seems to describe the same problem as #3233, which was closed as resolved in 0.200.  Was there a regression?

Also related: #3808.

hrldcpr:
@Miciah I'm using the latest version of spacemacs, but I still experience this issue.

In particular, I have this code in my `dotspacemacs/user-config`:

```elisp
(add-hook 'json-mode-hook 'prettier-js-mode)
```

…and prettier-js-mode is correctly enabled if I C-x C-f to a .json file, but not if I invoke spacemacs with `emacs whatever.json`

If, however, I move the code to `dotspacemacs/user-init`, then `emacs whatever.json` works correctly.

PR #7220, opened 2016-09-29
Correct the format of recipe. Fix #6610

### LINKED ISSUE #6610: Packages installed using file fetcher are always marked as outdated
#### Description :octocat:
Package installed using file fetcher are always marked as outdated when pressing 'update packages' button. 

- Start Emacs.
- Press 'update packages' button.
- Notice that `evil-unimpaired` has new version (and probably some other packages).
- Ask Emacs to update them.
- Restart Emacs so packages are updated.
- Press update button again. 

Package `evil-unimpaired` has new version.

Package `evil-unimpaired` wasn't changed. So it should not be marked as outdated.

#### System Info :computer:
- OS: darwin
- Emacs: 25.1.50.1
- Spacemacs: 0.105.21
- Spacemacs branch: develop (rev. 7a44019)
- Graphic display: t
- Distribution: spacemacs
- Editing style: emacs
- Completion: ivy
- Layers:
```elisp
(agda
 (auto-completion :disabled-for org)
 better-defaults colors dash erc
 (git :variables git-magit-status-fullscreen t)
 github
 pandoc ranger restclient syntax-checking spacemacs-layouts ivy spell-checking shell version-control search-engine no-dots d12frosted-core d12frosted-org d12frosted-lua d12frosted-haskell d12frosted-lisp d12frosted-visual d12frosted-spellchecking emacs-lisp
 (html :variables web-mode-markup-indent-offset 2 web-mode-attr-indent-offset 2 web-mode-code-indent-offset 2 web-mode-css-indent-offset 2 web-mode-sql-indent-offset 2 css-indent-offset 2)
 (javascript :variables js-indent-level 2 js2-basic-offset 2)
 lua markdown shell-scripts yaml react)
```

ynilu:
After a long dig, I finally found the cause of #6610 .

According to the example in the README of quelpa:

``` elisp
(quelpa '(rainbow-mode :fetcher file :path "/path/to/rainbow-mode/rainbow-mode.el"))
```
The car of the recipe should be the package name instead of "recipe"

P.S.
The upgrade checkout for packages using quelpa wont be done unless we set the `quelpa-upgrade-p` or add `:upgrade` in the recipe:

``` elisp
(defcustom quelpa-upgrade-p nil
  "When non-nil, `quelpa' will try to upgrade packages.
The global value can be overridden for each package by supplying
the `:upgrade' argument."
  :group 'quelpa
  :type 'boolean)
```

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -1419,18 +1419,18 @@ wether the declared layer is an used one or not."
     (when pkg
       (let ((location (oref pkg :location)))
         (when (and (listp location) (eq 'recipe (car location)))
-          location)))))
+          (cons pkg-name (cdr location)))))))
 
 (defun configuration-layer//new-version-available-p (pkg-name)
   "Return non nil if there is a new version available for PKG-NAME."
   (let ((recipe (configuration-layer//get-package-recipe pkg-name))
         (cur-version (configuration-layer//get-package-version-string pkg-name))
+        (quelpa-upgrade-p t)
         new-version)
     (when cur-version
       (setq new-version
             (if recipe
-                (quelpa-checkout recipe (expand-file-name (symbol-name pkg-name)
-                                                          quelpa-build-dir))
+                (or (quelpa-checkout recipe (expand-file-name (symbol-name pkg-name) quelpa-build-dir)) cur-version)
               (configuration-layer//get-latest-package-version-string
                pkg-name)))
       ;; (message "%s: %s > %s ?" pkg-name cur-version new-version)
--- a/tests/core/core-configuration-layer-utest.el
+++ b/tests/core/core-configuration-layer-utest.el
@@ -1746,7 +1746,7 @@
                                           'layer-no-recipe-1)
        ,(configuration-layer/make-package '(pkg2 :location elpa)
                                           'layer-no-recipe-1)) t)
-    (should (eq 'recipe
+    (should (eq 'pkg1
                 (car (configuration-layer//get-package-recipe 'pkg1))))))
 
 (ert-deftest test-get-package-recipe--return-nil-if-package-has-no-recipe ()


robbyoconnor:
@ynilu is awesome! 

@syl20bnr Can this PLEASE be in the next release?!?!?!

robbyoconnor:
@ynilu :100: :cake: :star:

d12frosted:
Very nice. I am happy some one found the problem. 💯

layers/+tools/command-log/packages.el
```diff
@@ -14,7 +14,7 @@
       ;; is an open PR to merge the fork into the original repo.
       ;; TODO when the PR is merged upstream, change to use the original package
       ;; from MELPA (IOW remove :location argument)
-      '((command-log-mode :location (recipe :fetcher github
+      '((command-log-mode :location (command-log-mode :fetcher github
```
d12frosted: I thought that packages with github fetcher are not marked as outdated every time. Does they?

ynilu: I didn't dig into the github fetcher part, but quelpa use the package build library from melpa to deal with the github fetcher. Since melpa uses the same recipe format as I mentioned, I think this should be the correct format of the command-log-mode recipe

syl20bnr:
`recipe` is a Spacemacs keyword and it is not passed to quelpa when calling the quelpa functions. Like @d12frosted said this does not happen with github repo whereas the recipe keyword is here.

I think you are on the right track but removing the recipe  keyword is not a good fix, recipe is important to tell the layer system that the location is a recipe.

ynilu:
@syl20bnr Thank you for point out the `recipe` is a Spacemacs keyword. So Maybe I should change the behavior of `configuration-layer//get-package-recipe` to return the right recipe.

ynilu:
@syl20bnr @d12frosted 
Does this change make sense? Or should I insert the package name outside the `configuration-layer//get-package-recipe` function?

I found that we need to set `quelpa-upgrade-p` to non-nil when we want to check for update, or `quelpa-checkout` won't get into the checkout part 

```elisp
(defun quelpa-checkout (rcp dir)
  "Return the version of the new package given a RCP.
Return nil if the package is already installed and should not be upgraded."
  (pcase-let ((`(,name . ,config) rcp)
              (package-build-stable quelpa-stable-p))
    (unless (or (and (assq name package-alist) (not quelpa-upgrade-p))
                (and (not config)
                     (quelpa-message t "no recipe found for package `%s'" name)))
      (let ((version (condition-case err
                         (package-build-checkout name config dir)
                       (error (quelpa-message t
                                              "failed to checkout `%s': `%s'"
                                              name
                                              (error-message-string err))
                              nil))))
        (when (quelpa-version>-p name version)
          version)))))
```

syl20bnr:
This looks good to me :+1: 

You can add the quelpa variable in this PR if it is needed.

ynilu:
I added the `quelpa-upgrade-p` part.

Should I also change the unit test for `configuration-layer//get-package-recipe`?
```elisp
(ert-deftest test-get-package-recipe--return-recipe-if-package-has-one ()
  (let ((configuration-layer--packages
         `(,(configuration-layer/make-package '(pkg1 :location (recipe blah)))
           ,(configuration-layer/make-package '(pkg2 :location elpa)))))
    (should (eq 'recipe
                (car (configuration-layer//get-package-recipe 'pkg1))))))
```

syl20bnr:
yes :-)

ynilu:
I corrected the unit test for `configuration-layer//get-package-recipe`

I think this PR is ready to be merged.

Please verify that.

Thanks

bixuanzju:
Any news on this PR?

d12frosted:
Sorry, I've been busy on these weekends. Gonna check it today / tomorrow.

d12frosted:
This works like a charm! Thank you so much for fixing this issue! Also, I just tried changing `evil-unimpaired` source code to see if it gets update during 'update packages' flow. And it does! 

PR #7224, opened 2016-09-30
Enable orgtbl-mode in markdown-mode

robbyoconnor:
This makes working with tables much nicer. Stolen shamelessly from [this StackOverflow answer](http://stackoverflow.com/a/26297700) and it works flawlessly.


d12frosted:
Ya, using this same trick. Good to bring it here 👍

d12frosted:

PR #7227, opened 2016-09-30
Move vi-tilde auxilliary functions to spacemacs-evil

bmag:
`spacemacs/disable-vi-tilde-fringe` and `spacemacs/disable-vi-tilde-fringe-read-only` are used by `spacemacs-evil` layer, but are defined in `spacemacs-ui-visual` layer. This PR moves the definitions to the
correct place. The old behavior causes a bug during startup, when using `spacemacs-base` with `spacemacs-evil` and without `spacemacs-ui-visual`.


TheBB:

PR #7253, opened 2016-10-03
Moved rebox to be under tools, as per documentation

Tritlo:
The rebox layer was directly under layers, while from the documentation (the changelog), it would seem that it should be under `+tools/`. This rectifies this mistake.


TheBB:

PR #7254, opened 2016-10-03
Fix changelog saying systemd is in +config-files

Tritlo:
The changelog says that systemd is under `+config-files`, when it is really under `+tools`. This fixes that mistake.


d12frosted:
Oh ya, it's because `+config-files` was renamed to `+tools`. Thanks for the catch!

d12frosted:

Issue #5130, opened 2016-02-18
Emmet support in React layer


favetelinguis:
Is it possible to add emmet support in the React Layer? 

When working with JSX files it is very nice to be able to use the HTML parts of Emmet, i know there is support in emmet for making class into className etc.

begizi:
Was able to do this by setting `emmet-expand-jsx-className? t` in `setq-default` in the `dotspacemacs/user-config` section of my `.spacemacs` config. Changes globally though so even in html it'll use `className` instead of `class`. Can maybe do something with mode hooks to scope this to jsx files? I'm pretty new to spacemacs/emacs though so I wouldn't know where to start to make that work.

TheBB:
Try this instead:

```elisp
(add-hook 'react-mode-hook
          (lambda ()
            (setq-local emmet-expand-jsx-className? t)))
```

If it works you can PR it.

begizi:
It looks like react-mode doesn't work with emett? I can only get it to work when in web-mode. In react-mode I get the message `No matches`

tko:
For me after opening a .jsx file I get `emmet-expand-jsx-className?` variable has local value `t` (global value is nil) and `emmet-mode` is `nil` -- I've never used emmet-mode before but seems to do the expected thing as long as you enable it. Seems there's explicit `(emmet-mode 0)` in the layer configuration that disables it for some reason?

begizi:
@TheBB That hook ended up working, but as @tko pointed out I had to enable emmet-mode in react-mode first by setting `(add-hook 'react-mode-hook 'emmet-mode)` in my configuration.

begizi:
Actually. @TheBB It looks like jsx className is enabled in react-mode by default. I just had to add a hook to enable emmet mode and it worked as expected.

TheBB:
Even better :+1:

begizi:
Any reason why emmet isn't enabled in react-mode by default?

TheBB:
No.

favetelinguis:
How does emmet know to use html in. Jsx or. React.js and not css expand?
Den 23 feb 2016 21:35 skrev "Eivind Fonn" <notifications@github.com>:

sarith:
Any resolution on this? @begizi tried everything you've laid out here and only have 'no matches' as you did earlier...

begizi:
[...]

PR #7263, opened 2016-10-03
Add emmet-mode to react-mode

kaipr:
This enables `emmet-mode` in `react-mode` (`*.jsx` files).
It automatically uses `className` instead of `class`, so no other configuration needed.

Put it there because thats how other modes are hooked, could also add it as a hook in the react layer if that would be preferred?

Closes syl20bnr/spacemacs#5130

### DIFF
--- a/layers/+lang/html/packages.el
+++ b/layers/+lang/html/packages.el
@@ -88,6 +88,7 @@
                                                 html-mode-hook
                                                 sass-mode-hook
                                                 scss-mode-hook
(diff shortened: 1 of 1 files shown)


TheBB:

I moved it to the react layer.

NdYAG:
@kaipr @TheBB 
Should we also delete this line?
https://github.com/syl20bnr/spacemacs/blob/master/layers/+frameworks/react/packages.el#L87

PR #7272, opened 2016-10-04
Use j/k in spacemacs buffer for forward/backward widget motions. Curr…

tobimensch:
…ently j/k were only navigating between lines,
but hitting RET wouldn't work on a recent file and the like,
when the cursor wasn't directly over the filename.

Changing j/k to widget navigation, ensures that the cursor is always on something activatable. This change should make the home buffer feel more at home for evil mode users.

Not sure this is the best solution and it doesn't take into account non-evil users. So take it as a suggestion.

I found myself again and again using j/k, ending up on the beginning of a line containing a filename, pressing RET, and being slightly annoyed that it didn't just work.

### DIFF
--- a/core/core-spacemacs-buffer.el
+++ b/core/core-spacemacs-buffer.el
@@ -53,9 +53,9 @@ version the release note it displayed")
 
 (defvar spacemacs-buffer-mode-map
   (let ((map (make-sparse-keymap)))
-    (define-key map [tab] 'widget-forward)
+    (define-key map (kbd "j") 'widget-forward)
     (define-key map (kbd "C-i") 'widget-forward)
-    (define-key map [backtab] 'widget-backward)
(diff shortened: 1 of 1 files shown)


TheBB:
Not sure how I feel about this. As a vim user, having j/k do something other than line up or down surely doesn't make me feel at home, nor does this ensure that the cursor is always on something clickable. I'd rather have this on J/K as per standard evilifaction rules.

Anway, thanks! Squashed and cherry-picked in develop. You can safely delete your branch.

Please sanitize your commits next time (leave merges out) and keep your commit message according to reasonable rules.

syl20bnr:
> I'd rather have this on J/K as per standard evilifaction rules.

When looking at the merged PR I had the exact same reaction, I will correct it.

Issue #7262, opened 2016-10-03
How to toggle transparency in a terminal


brezal:
Hi. I started trying out Spacemacs over the weekend and it's been nice.

I came across M-m T T for toggling transparency, and it works in the standalone gui version of emacs, but from within a terminal, it doesn't seem to work. I'm using iTerm 3.0.9.

**Edit:** I meant *terminal emulator*, sorry about the confusion.

TheBB:
Of course, the terminal is a text-only interface, and has no concept of things like transparency. You will have to configure the terminal emulator itself to enable transparency.

brezal:
Hi @TheBB, thanks for responding so quickly. The terminal is transparent, but when I start spacemacs using `emacs -nw`, the editor's background is a pretty black with no transparency and no way of toggling the background to be transparent.

jasonm23:
I assume the color theme `default` face isn't matching the default background color for the terminal.

If you set them to match, it should become transparent.

```
(set-face-attribute 'default nil :background  COLOR)
```

where color is the same as the terminal background.  

BTW What is the terminal emulator you're using?

TheBB:
> I'm using iTerm 3.0.9.

brezal:
@jasonm23 
(set-face-attribute 'default nil :background  "black") just sets the background to black. The background color of the term emulator is `#000000`, but there is transparency.

1. Has anyone successfully been able to have transparency in a terminal emulator?
2. Has anyone successfully used the transparency feature in iTerm?

jasonm23:
Sorry, I just tried this out, my other solution was untested.

Instead of 

```
(set-face-attribute 'default nil :background "black")
```

Try 

```
(set-face-attribute 'default nil :background nil)
```

Simpler (perhaps just more direct)

```
(set-face-background 'default nil)
```

This seems to work for me.

<img width="723" alt="screen shot 2016-10-09 at 1 54 47 am" src="https://cloud.githubusercontent.com/assets/71587/19214963/6ceb87dc-8dc3-11e6-8f9d-024ed3fde5fb.png">

#7288 is for GUI of course.

brezal:
<img width="1920" alt="screen shot 2016-10-08 at 9 15 05 pm" src="https://cloud.githubusercontent.com/assets/2381998/19217588/5b090a2a-8d9c-11e6-8350-8d905a1855f0.png">
[...]

PR #7288, opened 2016-10-05
Add optional frame arguments for transparency toggle/in/decrease.

TGThorax:
This way users can easily have transparency with all newly created frames.
```
  (spacemacs/toggle-transparency)
  (add-hook 'after-make-frame-functions
            (lambda (f)
              (spacemacs/toggle-transparency f)))
```
The increase and decrease function needn't be changed for the above code to work, but it's more consistent.

### DIFF
--- a/layers/+distributions/spacemacs-base/keybindings.el
+++ b/layers/+distributions/spacemacs-base/keybindings.el
@@ -584,11 +584,11 @@ otherwise it is scaled down."
 
 ;; Transparency transient-state
 
-(defun spacemacs/toggle-transparency ()
-  "Toggle between transparent or opaque display."
+(defun spacemacs/toggle-transparency (&optional frame)
+  "Toggle between transparent and opaque state for FRAME.
+If FRAME is nil, it defaults to the selected frame."
   (interactive)
-  (let* ((frame (selected-frame))
-         (alpha (frame-parameter frame 'alpha))
+  (let* ((alpha (frame-parameter frame 'alpha))
          (dotfile-setting (cons dotspacemacs-active-transparency
                                 dotspacemacs-inactive-transparency)))
     (set-frame-parameter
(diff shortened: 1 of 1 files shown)


layers/+distributions/spacemacs-base/keybindings.el
```diff
@@ -584,10 +584,10 @@ otherwise it is scaled down."
 
 ;; Transparency transient-state
 
-(defun spacemacs/toggle-transparency ()
-  "Toggle between transparent or opaque display."
+(defun spacemacs/toggle-transparency (&optional frame)
+  "Toggle between transparent or opaque display for given or current frame."
   (interactive)
-  (let* ((frame (selected-frame))
+  (let* ((frame (or frame (selected-frame)))
```
bmag: If `frame` is nil, `frame-parameter` defaults to the selected frame, so we don't need to explicitly set `frame`.

layers/+distributions/spacemacs-base/keybindings.el
```diff
@@ -584,10 +584,10 @@ otherwise it is scaled down."
 
 ;; Transparency transient-state
 
-(defun spacemacs/toggle-transparency ()
-  "Toggle between transparent or opaque display."
+(defun spacemacs/toggle-transparency (&optional frame)
+  "Toggle between transparent or opaque display for given or current frame."
```
bmag: I think it would be better to phrase to docstring similar to how Emacs phrases an optional frame argument. Example from `frame-parameter`:
> Return FRAME’s value for parameter PARAMETER.
> If FRAME is nil, describe the currently selected frame.

In this case, something like:
> Toggle between transparent or opaque display for FRAME.
> FRAME defaults to the currently selected frame.

layers/+distributions/spacemacs-base/keybindings.el
```diff
@@ -597,22 +597,24 @@ otherwise it is scaled down."
          dotfile-setting
        '(100 . 100)))))
 
-(defun spacemacs/increase-transparency ()
-  "Increase transparency of current frame."
+(defun spacemacs/increase-transparency (&optional frame)
+  "Increase transparency of given or current frame."
```
bmag: See above comment about docstring.

layers/+distributions/spacemacs-base/keybindings.el
```diff
@@ -597,22 +597,24 @@ otherwise it is scaled down."
...
   (interactive)
-  (let* ((current-alpha (car (frame-parameter (selected-frame) 'alpha)))
+  (let* ((frame (or frame (selected-frame)))
```
bmag: See above comment about not needing to set `frame` explicitly.

layers/+distributions/spacemacs-base/keybindings.el
```diff
@@ -597,22 +597,24 @@ otherwise it is scaled down."
...
          (increased-alpha (- current-alpha 5)))
     (when (>= increased-alpha frame-alpha-lower-limit)
-      (set-frame-parameter (selected-frame) 'alpha
+      (set-frame-parameter frame 'alpha
                            (cons increased-alpha increased-alpha)))))
 
-(defun spacemacs/decrease-transparency ()
-  "Decrease transparency of current frame."
+(defun spacemacs/decrease-transparency (&optional frame)
+  "Decrease transparency of given or current frame."
```
bmag: See above comment about docstring.

layers/+distributions/spacemacs-base/keybindings.el
```diff
@@ -597,22 +597,24 @@ otherwise it is scaled down."
...
   (interactive)
-  (let* ((current-alpha (car (frame-parameter (selected-frame) 'alpha)))
+  (let* ((frame (or frame (selected-frame)))
```
bmag: See above comment about not needing to set frame explicitly.

bmag:
The example can be shorter:
```elisp
(spacemacs/toggle-transparency)
(add-hook 'after-make-frame-functions #'spacemacs/toggle-transparency)
```

TGThorax:
@bmag Thanks for the suggestions!! :smile:  (Of course the example could be written like that, that was the whole point, stupid me :disappointed: )

brezal:
Is this related to #7262 by chance?

bmag:
Unlikely, but you are welcome to merge the PR locally and check.

TheBB:

PR #7295, opened 2016-10-05
Change 'SPC :' to 'SPC SPC' in quick start

andschwa:
This is a documentation update. The quick start guide should probably not use the deprecated keybinding 😄


TheBB:

PR #7299, opened 2016-10-06
[org] Add org-clock key bindings

darkfeline:
Add useful global org-clock key bindings.


TheBB:

PR #7300, opened 2016-10-06
better-defaults layer: update emacs logo

deb0ch:
The new emacs logo just looks better :stuck_out_tongue_winking_eye: 

Btw, I copy/pasted the image file from the emacs-lisp layer, relying on git to detect duplication and avoid uselessly wasted memory.

I don't know if that is the best approach, but I also didn't find a clean way of sharing that picture between the two layers.


houli:
The official new logo is the one on https://www.gnu.org/software/emacs/ which can be found in SVG here https://github.com/emacs-mirror/emacs/blob/master/etc/images/icons/hicolor/scalable/apps/emacs.svg

deb0ch:
Hmmm I thought that it was it :sweat_smile:
It looked newish and all with the flat design. And I prefer that version over the old one even if it's not the official, but that's beside the point.

What is this logo that is in the emacs-lisp layer then ?

And how should we go about it, replace them both by the new logo, leave it like that or accept this PR ?

TheBB:
The emacs lisp layer logo is also the old one, note how the E points slightly more upwards in the new.

JAremko:
> The new emacs logo just looks better :stuck_out_tongue_winking_eye:
But the first one takes 10X more work to make. Go figure. :confused:

deb0ch:
Done :smile: 

@JAremko Got more efficient :stuck_out_tongue:

TheBB:

PR #7301, opened 2016-10-06
ycmd layer: do not parse buffer on newline insertion

deb0ch:
Dramatically improves ycmd performance on large projects.


TheBB:

PR #7323, opened 2016-10-08
New option: auto-generate layout names

bmag:
If `spacemacs-auto-generate-layout-names` is non-nil, and the user tries to open a layout in a position that doesn't yet have a layout (e.g. with `SPC l 2`), then create a new layout with an automatically generated name. I tried to choose names that are familiar and easy to remember. To enable the feature, add the following to `user-config`:

``` elisp
(setq spacemacs-auto-generate-layout-names t)
```

Known issues:
- persp-mode doesn't really have a notion of a numbered layout. Because of this, creating a layout with `SPC l 7` (7 is only an example) will give it a name suitable for a 7th layout, but the actual position of the layout may be smaller. For example, if the user only had 4 layouts, then the "7th" layout is created in position 5.
- the new layout is created with a single window, but it doesn't show the home buffer. Instead, it will show the buffer of the previous layout's selected window. This is a difference I do not understand (or researched) between `(persp-switch "some-new-name")` and `(persp-switch nil)`.

### DIFF
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -198,6 +198,10 @@ if used there.")
   "If non nil then the last auto saved layouts are resume automatically upon
 start.")
 
+(defvar dotspacemacs-auto-generate-layout-names nil
+  "If non-nil, auto-generate layout name when creating new layouts.
+Only has effect when using the \"jump to layout by number\" commands.")
+
 (defvar dotspacemacs-max-rollback-slots 5
   "Maximum number of rollback slots to keep in the cache.")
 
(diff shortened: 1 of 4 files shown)


JAremko:
Nice addition. Thanks!

TheBB:
Do you mean grab *bag*? :-)

syl20bnr:
`spacemacs-auto-generate-layout-names` could be a dotspacemacs variable, check for the shape of other layout related dsm variables.

bmag:
@TheBB lol, of course
@syl20bnr ok, updated accordingly. Original version was modelled after `spacemacs-spaceline-additional-segments`.

TheBB:

I had to apply a fix: https://github.com/syl20bnr/spacemacs/commit/d4b004433cc26667a4ff1c18225fa74d1ce35842

Bizarrely, by default, `persp-get-by-name` will not return nil if it can't find a persp, but `:nil`.

PR #7337, opened 2016-10-09
Fix unclosed strings in spacemacs.template

arronmabrey:
I found two occurrences of unclosed strings in the `.spacemacs.template` file. Which caused issues with [parinfer-mode](https://github.com/DogLooksGood/parinfer-mode)'s `indent-mode`

This PR fixes them.

### DIFF
--- a/core/templates/.spacemacs.template
+++ b/core/templates/.spacemacs.template
@@ -114,7 +114,7 @@ values."
    ;; List of items to show in startup buffer or an association list of
    ;; the form `(list-type . list-size)`. If nil then it is disabled.
    ;; Possible values for list-type are:
-   ;; `recents' `bookmarks' `projects' `agenda' `todos'."
+   ;; `recents' `bookmarks' `projects' `agenda' `todos'.
    ;; Example for 5 recent files and 7 projects: '((recents . 5) (projects . 7))
    ;; List sizes may be nil, in which case
    ;; `spacemacs-buffer-startup-lists-length' takes effect.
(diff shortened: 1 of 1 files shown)


DogLooksGood:
BTW, This may be a bug of parinfer-mode, I'll try to reproduce it and address the problem.

In most case, parinfer-mode should not be affected by the content in comment.

core/templates/.spacemacs.template
```diff
@@ -114,7 +114,7 @@ values."
    ;; List of items to show in startup buffer or an association list of
    ;; the form `(list-type . list-size)`. If nil then it is disabled.
    ;; Possible values for list-type are:
-   ;; `recents' `bookmarks' `projects' `agenda' `todos'."
+   ;; "`recents' `bookmarks' `projects' `agenda' `todos'."
```
arronmabrey: yup, I'm down with that.

arronmabrey:
@DogLooksGood yeah I figured it might be a parinfer bug. I was planing to open an issue, glad you saw this. In fact I'm kinda surprised you saw it, Good Job 👍

DogLooksGood:
@arronmabrey https://shaunlebron.github.io/parinfer/#inserting-quotes

d12frosted:

I cherry-picked this because I don't like missing quotes even in comments. But that bug with parinfer should be fixed anyways (upstream) 😸 thanks everyone.

arronmabrey:
@d12frosted thanks! Yeah, unfortunately based on the [link about inserting-quotes](https://shaunlebron.github.io/parinfer/#inserting-quotes) @DogLooksGood posted. I don't think it's technically a bug, rather it's something you have to deal with because of how parinfer fundamentally works.

d12frosted:
@arronmabrey thanks for the link and additional info!

PR #7339, opened 2016-10-10
Add parinfer layer

DogLooksGood:
Add parinfer layer for spacemacs.  It should works fine with evil.


gilch:
I understand that `parinfer-mode` uses `aggressive-indent-mode` instead of the original "Paren Mode"? But Spacemacs already has a toggle for `aggressive-indent-mode`--`SPC t I`.

I'm concerned that it's possible to turn on `aggressive-indent-mode` and Parinfer's "Indent Mode" at the same time. These modes should be mutually exclusive. I see nothing in the layer to prevent that. Does the package already turn off Parinfer's "Indent Mode" if `aggressive-indent-mode` is activated manually?

Perhaps it would be preferable for `SPC t P` to alternately activate and deactivate Parinfer "Indent Mode", but only deactivate `aggressive-indent-mode`.

Then keep `SPC t I` current behavior to alternately activate and deactivate `aggressive-indent-mode`, but also deactivate Parinfer "Indent Mode" upon any activation of `aggressive-indent-mode`.

gilch:
I see you enabled Parinfer by default for both Emacs Lisp and Clojure.

Spacemacs has a layer for both Common Lisp and Scheme. I think you should use the hook for these other Lisps too.

There are various other lesser-known Lisp dialects with no Spacemacs layer (so far) which nonetheless have corresponding Emacs major modes. The layer doc could perhaps mention how to use a hook in the `.spacemacs` file to activate `parinfer-mode` for other major modes.

DogLooksGood:
Hi, @gilch.
`parinfer-mode` will remove the dependency `aggressive-indent-mode` in next version(currently `devel` branch) to make it a more independent.

The `Indent Mode` and `aggressive-indent-mode` can't work together because we can't modify the indentation when `aggressive-indent-mode` is enabled. `aggressive-indent-mode` works only when the current top-level sexp has balanced parens, and `parinfer-mode` did the same thing for its `Paren Mode`.  

I'm not familiar with how to deal with compatibility problems in layer.  But in my opinion, we should turn off `aggressive-indent-mode` when `parinfer-mode` is activated, and bind `parinfer-toggle-mode` to, maybe `SPC t I`.

DogLooksGood:
Yes, should enable `parinfer-mode` for all lisp dialects. 
I'm not clear about the hook names, may you provide?

gilch:
`common-lisp-mode-hook` and `scheme-mode-hook`, I think. `lisp-mode-hook` may also be used for nonspecific other Lisp dialects.

DogLooksGood:
Thanks, I'll add these hooks.

layers/+vim/parinfer/packages.el
```diff
@@ -0,0 +1,33 @@
...
+;;; License: GPLv3
+
+(defconst parinfer-packages
+  '(parinfer))
+
+(defun parinfer/init-parinfer ()
+  (use-package parinfer
+    :defer t
+    :diminish parinfer-mode
+    :config
```
bmag: Please remove `:config` keyword if there's no `:config` block.

gilch:
Other Lisp dialects may have their own major mode hooks, but I think it's best to let the user configure them when there's no Spacemacs layer, since we may have trouble finding them all. Hence my recommendation to mention using the hook in `.spacemacs` in this layer's doc.

I'm not sure how best to deal with incompatibilities either. Perhaps someone from the Spacemacs team will have more to say about it.

> parinfer-mode will remove the dependency aggressive-indent-mode in next version(currently devel branch) to make it a more independent.

Others may have a different preference, but I personally would prefer to continue using `aggressive-indent-mode` (rather than "Paren Mode") with `parinfer-mode`, since `aggressive-indent-mode` reduces the number of pointless choices I have to make when coding, and also makes it easier to keep a team project's style consistent.

DogLooksGood:
@gilch 
I just remove the dependencies, the behavior for `Paren Mode` will not change. Because we only use one function from `aggressive-indent-mode`, I just port that function to parinfer-mode.

layers/+vim/parinfer/packages.el
DogLooksGood: Thanks for review! I've remove the unnecessary `:config` and append commits.

sondr3:
Can I ask what benefit something like Parinfer has over something like Lispy/Paredit?

DogLooksGood:
I'm a `Lispy` user, and now I'm using `Parinfer` + `Lispy`. `Parinfer` is not a 100% covered solution for structural edit I think, but it provide more convenience over Paredit/Lispy in some cases. Especially when you want to insert code to an exist expression, you don't have to find the close parens anymore.

joakimmohn:
Big parinfer and spacemacs fan here! Whats the hold up for parinfer-mode? is there a way to get parinfer to work with spacemacs now?

DogLooksGood:
Hi, @joakimmohn, this PR should work. Just not merged yet.

lsdafjklsd:
Manually added these changes to my `~/.emacs.d/layers/+vim/`, works amazing! Just what this was missing :) Thanks again

ekhall:
This is super- thank you!

DogLooksGood:
Hi, 
Since this PR has been opened for a while. Is there any problem we need to solve to help merging this?

ekhall:
It worked very easily for me.

d12frosted:
Hey, thank you for this layer! I haven't used parinfer yet, but I am looking forward to try it very soon.

Since it's a new layer, it's @syl20bnr who has to look through it and merge it. Sorry for the delay.

gilch:
I'm getting a strange error. I installed the develop version of Spacemacs with a fresh `.spacemacs`, added the parinfer folder from this PR to `layers/+vim/`, and then added the `parinfer` layer to my `.spacemacs`. (I had to restart emacs a few times.) Then I created a `test.el` buffer, which started in Emacs-Lisp mode. I tried appending a character to the buffer and get the following error:
```
Error in post-command-hook (parinfer--invoke-parinfer-when-necessary): (invalid-function parinfer-save-excursion)
```
The point also jumps to the start of the buffer.

DogLooksGood:
Hi, @gilch. What's your  parinfer-mode version and emacs version? 
`parinfer-save-excursion` is https://github.com/DogLooksGood/parinfer-mode/blob/master/parinfer.el#L184

It should be `save-mark-and-excursion` in Emacs 25 and `save-excursion` in Emacs 24.

gilch:
That error was when running emacs 24.5.1 on Ubuntu. Parinfer is whatever Spacemacs downloaded given this layer. I'm not sure where to find the version for that. The header comment of `parinfer.el` shows version 0.4.7, if that's what you're asking for.

gilch:
Also, if I do `SPC h d f` on `parinfer-save-excursion`, it says that ```parinfer-save-excursion is an alias for `save-excursion' in `parinfer.el'.```, so it does appear to be `save-excursion` in Emacs 24 as expected.

DogLooksGood:
That it is.
I'll try to reproduce the problem.

sineer:
Thank you for this layer and parinfer-mode, it rocks!

ruliana:
Any progress on it? Any way I could help? :)

DogLooksGood:
@ruliana met any issues? In my case this works fine. @gilch said there's some issues but I could not reproduce.

ruliana:
@DogLooksGood No issues here. Just wondering if there is anything missing to merge it. It's ok to copy the PR to +vim/layers, but it would be nicer if it got merged ^_^ Anyway, thanks for the layer!

Akeboshiwind:
Hi @DogLooksGood, I'm getting the same issue as @glitch but in a Clojure buffer.
I'm running Ubuntu 16.04.1 LTS with Emacs version 24.5.1.

If you need any help reproducing the error I'm happy to help, I'd be great get this layer up and running.

DogLooksGood:
Hi @Akeboshiwind. It will be a great help if you provide your configuration and the steps to reproduce this issue.

Akeboshiwind:
@DogLooksGood, [here](https://github.com/Akeboshiwind/dotfiles/blob/master/.spacemacs) is my .spacemacs file.

**Steps to reproduce:**
1. Download the [parinfer](https://github.com/DogLooksGood/spacemacs/tree/develop/layers/%2Bvim/parinfer) directory from your branch of the repo.
2. Put the downloaded folder into the ~/.emacs.d/private/ folder on my local machine.
3. Add `parinfer` to the `dotspacemacs-configuration-layers` section of my .spacemacs like so:
image
4. Load spacemacs to install the layer.
5. Close and reopen spacemacs.
6. Open a clojure buffer.
7. Enter insert mode and press the backspace button.
8. The error appears.

I told you before but, **Version stuff:**
```
ake@ubuntu:~/.emacs.d/private$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 16.04.1 LTS
Release:        16.04
Codename:       xenial

ake@ubuntu:~/.emacs.d/private$ emacs --version
GNU Emacs 24.5.1
Copyright (C) 2015 Free Software Foundation, Inc.
GNU Emacs comes with ABSOLUTELY NO WARRANTY.
You may redistribute copies of Emacs
under the terms of the GNU General Public License.
For more information about these matters, see the file named COPYING.
```

If you like I can make a branch of my dotfiles repo with the changes in the steps above in it, although I think it's probably to do with the emacs version seeing as @gilch had the same issue with the same version.

DogLooksGood:
Hi, @Akeboshiwind . Thanks for help! I have reproduced this issue. 

the problem is that, AFAIK Emacs 25's `save-mark-and-excursion` did the same with `save-excursion` in Emacs 24's. So I write a alias for this:
```
(defalias 'parinfer-save-excursion
  (if (fboundp 'save-mark-and-excursion)
      'save-mark-and-excursion
    'save-excursion))
```

And this is the cause for problem, I don't know why it complains `parinfer-save-excursion` is an invalid function. Is there anyone can help to work this out?

Akeboshiwind:
Hmm, it's very strange.

I thought that it might be that the second parameter of `defalias` needs to resolve to a function, but even the GNU docs does the exact same thing [here](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html) (near the end).

Even when I play about with it in the \*scratch\* buffer it seems to work correctly, it would be good to know why your code doesn't work.

In the mean time does your fix work?

DogLooksGood:
@Akeboshiwind Have you tried to modify `parinfer.el` in parinfer-mode and  delete `parinfer.elc` to see if this fix works?

Akeboshiwind:
That seems to have fixed it yes, I'm not getting the issue anymore.

sineer:
Some strange behaviour creeped up on me using brew's Emacs 25.2.1 and spacemacs (tried both develop and current master branch):

Whenever parinfer mode is enabled and I try to paste over some selection the selection does not get overwritten. I'm pretty sure that used to work a few days ago and I have no clue what changed (not my emacs or my config, maybe some ELPA packages?)... emacs delete-selection-mode is on and works as expected. It's just whenever I try to overwrite a selection using OS X paste that it fails (the text get pasted but at cursor location and the selection does not get deleted).

I tried to find a fix or why it started to act like that but I failed :(

It is very annoying as pasting over selection is something I do all the time. I really wished I knew what broke this because I currently cannot find a fix and have to disable this sweet sweet parinfer-mode...

I'd be happy to help try any solution! anyone else having that issue?

DogLooksGood:
@sineer I personally don't use `delete-selection-mode`, but this issue exists in my Emacs too if I enable `delete-selection-mode`.

It is difficult to find where is broken.  Maybe we can write a new command for this behavior and bind it to cmd-V and C-y?

sineer:
@DogLooksGood Hi! Thanks for reply and your amazing work on parinfer-mode I like it very much!!

I just tried disabling delete-selection-mode and the issue remains.

Isn't it supposed to be normal behaviour when delete-selection-mode is disabled? You're saying that you don't use delete-selection-mode yet when you select a region and paste stuff it overwrites it instead of pasting it at cursor location?

I shall try it on my linux box later today. I'm a little confused because I think for a while this issue was not present... I only discovered parinfer-mode very recently so perhaps it was but I believe I would've noticed that earlier.

I'm curious to know how one can live without delete-selection-mode? I copy/paste code from the browser all the time or from other buffers, how do you go about pasting over stuff? You press the backspace key to delete selection prior to pasting every time? It is very counter intuitive for me it breaks my workflow pretty bad...

I tried looking at your code (which is nice by the way!) to figure it out but I did not get very far because I don't know enough about emacs internals.

It would be great if you could come up with a fix :) I'm willing to try anything.

Thank you for your time!

DogLooksGood:
@sineer recently don't have enough time to investigate  the problem.
maybe you need some command like this:
```
(defun yank-with-delete-selection ()
  (interactive)
  (if (region-active-p)
      (progn (delete-region (region-beginning) (region-end))
             (yank))
    (yank)))
```

sineer:
It works! 👍 

I replaced the (call-interactively 'yank) by your yank-with-delete-selection code and it properly overwrite selection when I paste.

Thank you very much friend! You are awesome.

JAremko:
@syl20bnr  Sooo... :roll_eyes:

jazen:
Any update?

t3chnoboy:
I've been using this layer for a while and I haven't had any problems with it.
One minor suggestion: it would be nice to automatically disable `smartparens` in this mode.

ClashTheBunny:
This is working wonderfully on Emacs 25.2.1 with Spacemacs 0.200.9 and all other plugins fully updated as of yesterday.  Thanks so much for all the hard work you put into this!  Loving it!

achikin:
Will it be merged?

TheBB:

I cleaned things up a bit in https://github.com/syl20bnr/spacemacs/commit/fab3b2a4e4cddb0cd76ea2be6beb21daf9451c09

In my brief testing I only experienced some minor glitches with smartparens, so I haven't done anything with that for the moment. I'm also considering moving the toggle onto a major mode keybinding.

robbyoconnor:
@DogLooksGood ensure trailing whitespace is stripped when you make PRs -- it's tricky to see it but you had it there. It's more visible in magit's diff view...

PR #7345, opened 2016-10-10
Higher visibility warnings + check if layers changed outside dotspacemacs/layers

TheBB:
The first commit stores the value of `dotspacemacs/layers` immediately after calling the function, and adds a check in `spacemacs-post-user-config-hook` to check if the value was changed. If it was, it issues a warning. (Note, the comparison is done with `eq`, not `equal`.) It also uses the saved value when reporting an issue.

The second commit shows warnings as a non-removable startup list.

Example:

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -375,6 +375,8 @@ refreshed during the current session."
   "Synchronize declared layers in dotfile with spacemacs.
 If NO-INSTALL is non nil then install steps are skipped."
   (dotspacemacs|call-func dotspacemacs/layers "Calling dotfile layers...")
+  (setq dotspacemacs--configuration-layers-saved
+        dotspacemacs-configuration-layers)
   (when (spacemacs-buffer//choose-banner)
     (spacemacs-buffer//inject-version))
   ;; declare used layers then packages as soon as possible to resolve
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -126,6 +126,9 @@ whenever you start Emacs.")
 (defvar dotspacemacs-configuration-layers '(emacs-lisp)
   "List of configuration layers to load.")
 
+(defvar dotspacemacs--configuration-layers-saved nil
+  "Saved value of `dotspacemacs-configuration-layers' after sync.")
+
 (defvar dotspacemacs-themes '(spacemacs-dark
                               spacemacs-light)
   "List of themes, the first of the list is loaded when spacemacs starts.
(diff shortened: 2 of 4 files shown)


nixmaniack:
I would really like to have this merged. Long back with `tern` like warnings I had included a popup to show the user that it's missing but it was a bit obtrusive and was moved to `*Messages*`. I think not many users actually go and check such warnings and miss some nice features. So :+1: for this addition.

d12frosted:
Very nice! 💯

TheBB:
Well I merged it, what's the worst that could happen.

kccai:
Thanks! But is there a way of toggle showing the warning in the home buffer? 
@TheBB

TheBB:
No, you're supposed to deal with them. What warnings are you getting?

kccai:
I have been having the following warnings for a very long time, but I just ignore them as my configurations work just fine ... 

```
- Auto-evilication could not remap these functions in map `org-agenda-mode-map': - 
`org-agenda-next-line' originally mapped on `C-n'      
- Auto-evilication could not remap these functions in map `org-agenda-mode-map': -
`org-agenda-toggle-time-grid' originally mapped on `G'      
- Auto-evilication could not remap these functions in map `org-agenda-mode-map': -
`org-agenda-filter-remove-all' originally mapped on `|'      
- Auto-evilication could not remap these functions in map `org-agenda-mode-map': -
`org-agenda-filter-by-tag-refine' originally mapped on `\'      
```

TheBB:
Okay, those are harmless enough, we can probably turn them into regular messages.

TheBB:
https://github.com/syl20bnr/spacemacs/pull/7465

nicbet:
Please forgive my ignorance, since the last update 0.200.x I am getting the warning on my home screen `dotspacemacs-configuration-layers' was changed outside of `dotspacemacs/layers'.  - as recently merged based no this discussion. 

Would you mind providing pointers as to what exactly is wrong with my installation/configuration to emit such warning, and how to repair? Clearly, since something is 'wrong' enough to warrant printing a warning, I should take some action?

TheBB:
What spacemacs does is it runs the function `dotspacemacs/layers` and then it reads the value of `dotspacemacs-configuration-layers`, which is set in that function. Then, later on, it runs `dotspacemacs/user-config`.

One of the most frequent errors people do is to set `dotspacemacs-configuration-layers` in `dotspacemacs/user-config`. This will not work because of the above order of things. Due to how often this happens we figured it'd be better to warn about it.

After `dotspacemacs/layers` runs the value of `dotspacemacs-configuration-layers` should not ever change. If you do that you are almost 100% doing something wrong without realizing it.

nicbet:
Thanks TheBB for the clarification. I had indeed done some nyan-cat layer configuration in the `user-config` function calling `set dotspacemacs-configuration-layers`

PR #7348, opened 2016-10-10
Check package excluded property before toggle

TheBB:
Makes https://github.com/syl20bnr/spacemacs/pull/7345 easier to swallow for people like me who have the javascript layer enabled but exclude tern.


TheBB:
In.

Issue #7298, opened 2016-10-06
Can't use org-edit-special in org-mode with plantuml (puml-mode)


EMayej:
#### Description :octocat:
Can't use org-edit-specil with plantuml

#### Reproduction guide :beetle:
- Start Emacs
- Open test.org
- Type in below text
```org
# BEGIN_SRC plantuml :file x.svg
  Alice -> Bob : Hello
# END_SRC
```
- `SPC m '` (which calls org-edit-special)

*Observed behaviour:* :eyes: :broken_heart:
emacs reports no plantuml-mode

*Expected behaviour:* :heart: :smile:
a new window opens with puml-mode

*Possible solution:*
https://github.com/skuro/puml-mode#integration-with-org-mode

#### System Info :computer:
- OS: darwin
- Emacs: 25.1.1
- Spacemacs: 0.105.22
- Spacemacs branch: develop (rev. cf951a5)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
(ivy
 (auto-completion :variables auto-completion-enable-snippets-in-popup t)
 clojure semantic dash emacs-lisp fasd git plantuml graphviz org osx ranger shell-scripts
 (python :variables python-test-runner 'pytest)
 (shell :variables shell-default-shell 'eshell shell-default-height 30 shell-default-position 'bottom)
 spell-checking syntax-checking)
```

Mitchicus:
Yes, and also for this layer it could be handy to document that you may need to set `puml-plantuml-jar-path`. I got tripped up on this when the layer was added and replaced my own config. 

I'll make a PR briefly to fix this issue

PR #7353, opened 2016-10-10
Fix plantuml org-babel loading

Mitchicus:
- Replace plantuml/post-init-org with a pre-init-org hook to properly
load the configuration.
- Associate plantuml with its major mode for org src buffers/blocks

Fixes: #7298

Sorry @NJBS as I didn't notice the Autumnal Cleanup 2016 beforehand and this was assigned to you it seems. Hopefully this saves you a trivial fix at least :smile:

### DIFF
--- a/layers/+lang/plantuml/packages.el
+++ b/layers/+lang/plantuml/packages.el
@@ -13,9 +13,10 @@
   '(org
     puml-mode))
 
-(defun plantuml/post-init-org ()
+(defun plantuml/pre-init-org ()
   (spacemacs|use-package-add-hook org
-    :post-config (add-to-list 'org-babel-load-languages '(plantuml . t))))
+    :post-config (add-to-list 'org-babel-load-languages '(plantuml . t))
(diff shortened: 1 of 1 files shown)


robbyoconnor:
Will this need tweaking due to  #7417 ???

layers/+lang/plantuml/packages.el
```diff
@@ -13,9 +13,10 @@
   '(org
     puml-mode))
 
-(defun plantuml/post-init-org ()
+(defun plantuml/pre-init-org ()
   (spacemacs|use-package-add-hook org
-    :post-config (add-to-list 'org-babel-load-languages '(plantuml . t))))
+    :post-config (add-to-list 'org-babel-load-languages '(plantuml . t))
+    (add-to-list 'org-src-lang-modes '("plantuml" . puml))))
```
d12frosted: this line needs change :D 

cc @robbyoconnor

TheBB:
I don't think this is necessary any more since the major mode is called plantuml-mode now. Therefore closing, let me know if I'm mistaken.

PR #7356, opened 2016-10-10
Fix connecting to default servers with TLS more than once with ERC

Hoot215:
The erc//servers function removes the :ssl variable from the list so it can be passed directly to erc-tls, but this means that subsequent connections will not use TLS. Thus, we should operate on a copy of the list instead.


TheBB:

PR #7359, opened 2016-10-11
A more intuitive window-split-right key binding 

liuchengxu:
`spc w -` means split-window-below, or "split-window-horizontal" personnally, where '-' could be seen as a horizontal line.

`spc w |` looks like more intuitive than `spc w /`, vertical line for vertical split.

### DIFF
--- a/layers/+distributions/spacemacs-base/keybindings.el
+++ b/layers/+distributions/spacemacs-base/keybindings.el
@@ -390,7 +390,7 @@
   "wv"  'split-window-right
   "wV"  'split-window-right-and-focus
   "ww"  'other-window
-  "w/"  'split-window-right
+  "w|"  'split-window-right
   "w="  'balance-windows
   "w_"  'spacemacs/maximize-horizontally)
 ;; text -----------------------------------------------------------------------


bmag:
1. why are there so many irrelevant commits in this PR?
2. `/` is easier to type than `|`, because it doesn't require pressing Shift. There's no harm in adding `SPC w |`, but please don't remove `SPC w /`.

crab9527:
1. It is my first PR actually, so maybe there are some misoperations unexpected..... I will be more careful....😳

2. 'l' is often used in my other programs's config, such as tmux. I thought this would be more intuitive. Thank you for your explanation. Maybe keeping this in my private config more approprivate.

TheBB:
You submitted a PR against develop but it looks like you branched off of master. You have to branch off of develop, too.

xuliuchengxlc:
@TheBB Thank you for your help!

tobimensch:
@bmag
> / is easier to type than |, because it doesn't require pressing Shift.

On a German keyboard layout you actually do need to press shift in order to type /. But you also need to type Alt Gr to type |, so both are about equally inconvenient.

robbyoconnor:
Okay you did something here... this needs to be rebased on top of `develop`...

PR #7361, opened 2016-10-11
Refactor spacemacs/python-execute-file

bmag:
- `compile-command` changes after every compilation, we don't need to reset it to `nil`.
- Let-bind `universal-argument` to minimize side effects (probably didn't hurt anything anyway).
- When running `SPC u SPC m c c` in python file, pre-fill `"python <filename>"` as the compile command, instead of whatever was in history. Old entries are still reachable via `M-p` and `M-n`.


TheBB:

Issue #6151, opened 2016-05-28
Toggle "Linum-Relative" functionality suggestion


TheGreenHabi:
Very often I toggle relative line numbers with `SPC t r``` without toggling "Line Numbers" first.
And "Relative Numbers" do not toggle automatically "Line Numbers" if they are not enabled.

I suggest to add functionality to enable "Line Numbers" when user enables "Relative Line Numbers" without activating "Line Numbers First"

syl20bnr:
I agree, would be more natural to have the same behavior for both toggles.

deb0ch:
#7365

TheBB:
Should be fixed in master.

PR #7365, opened 2016-10-11
Activating linum-relative now activates linum

deb0ch:
#6151

### DIFF
--- a/layers/+spacemacs/spacemacs-evil/funcs.el
+++ b/layers/+spacemacs/spacemacs-evil/funcs.el
@@ -57,6 +57,12 @@ Otherwise, revert to the default behavior (i.e. enable `evil-insert-state')."
                       :background nil
                       :foreground nil))
 
+(defun spacemacs/linum-relative-toggle ()
+  (interactive)
+  (if (not (bound-and-true-p linum-relative-mode))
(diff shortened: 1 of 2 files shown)


TheBB:
What's the desired behaviour if I turn it off though? Right now it will just go back to regular line numbers.

deb0ch:
Yes, it is intended. 

That's to not conflict with people (like me) who often switch between normal and relative.

This way I'm not changing the behaviour that already exists, and the only behaviour that *should* always be logical and was missing is the one change brought by that commit, activating linum when turning on relative linum.

Does this make sense to you ?

TheBB:
Yeah it makes sense, but maybe toggles isn't really the thing when there are three states. Maybe it would be more POLA to have three bindings that bring the user to no, regular and relative line numbers respectively, disregarding the current state.

TheBB:
Anyway I'm fine with this for now. There are conflicts though, maybe you can fix those.

deb0ch:
Ooops, forgot about the conflicts. It's all fixed now :smile: 

But I think that it is best if we can avoid adding an extra shortcut. They are cluttered enough like that.

TheBB:

PR #7382, opened 2016-10-13
Add leader binding for `flycheck-explain-error`

fmdkdd:
Flycheck recently gained a feature to explain the error at point (bound
under `C-c ! e` by default).  This commit adds a corresponding leader
binding to the flycheck layer.


TheBB:

PR #7387, opened 2016-10-13
Remove invalid keybindings of ensime

atreeyang:
The function **ensime-db-start** is obsoleted and removed in the newest ensime. 
In the fact, we can not find the function ensime-db-start in the spacemacs.
According to the document (http://ensime.github.io/editors/emacs/userguide/#debugging), now ensime need you use ensimeRunDebug command to start a debug session.

ensime-db-list-locals is unavailable as well.

So remove the invalid keybinding accordingly.


TheBB:

PR #7391, opened 2016-10-13
Add goto definition and find references keybindings to tsx files in typescript layer.

george-miller:
mode so we can properly edit/format the html that's inside.  However, with this
the current function/variable.  You also lose the ", g u" keybinding which
gives you all references of current function/variable.  This PR adds
those into the web mode for tsx files.

### DIFF
--- a/layers/+distributions/spacemacs-base/keybindings.el
+++ b/layers/+distributions/spacemacs-base/keybindings.el
@@ -234,6 +234,7 @@
   "cr" 'recompile
   "cd" 'spacemacs/close-compilation-window)
 (with-eval-after-load 'compile
+  (evil-define-key 'motion compilation-mode-map (kbd "gf") 'find-file-at-point)
   (define-key compilation-mode-map "r" 'recompile)
   (define-key compilation-mode-map "g" nil))
 ;; narrow & widen -------------------------------------------------------------


layers/+lang/typescript/packages.el
```diff
@@ -72,6 +72,8 @@
                 (tide-setup)
                 (flycheck-mode +1)
                 (eldoc-mode +1)
+                (define-key spacemacs-web-mode-map (kbd "g g") 'tide-jump-to-definition)
+                (define-key spacemacs-web-mode-map (kbd "g u") 'tide-references)
```
TheBB: Please use `spacemacs/set-leader-key-for-major-mode`

TheBB: Also, please use the new jumping API, don't bind `SPC g g` directly. See other layers for examples.

zilongshanren: @TheBB  I noticed that `tide-jump-to-definition` is already in `spacemacs-jump-handlers-typescript-mode` but it can't jump around. I could call the `tide-jump-to-definition` manually to jump around. Even if I add `:async t` to the jump handlers, it still doesn't work.

Any suggestions? thanks.

Issue #7383, opened 2016-10-13
Difficult to change sp-show-pair-delay


marienz:
#### Description :octocat:

Difficult to change `sp-show-pair-delay`
#### Reproduction guide :beetle:
- Set `sp-show-pair-delay` to 0 in either `dotspacemacs/user-init` or `dotspacemacs/user-config`
- Start Emacs

_Observed behaviour:_ :eyes: :broken_heart:
Emacs behaves as if `sp-show-pair-delay` is 0.2.

_Expected behaviour:_ :heart: :smile:
Emacs behaves as if `sp-show-pair-delay` is 0.
#### System Info :computer:
- OS: gnu/linux
- Emacs: 25.1.1
- Spacemacs: 0.200.1
- Spacemacs branch: master (rev. 3706a42)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: ivy
- Layers:

``` elisp
(ivy auto-completion better-defaults emacs-lisp
     (go :variables go-tab-width 4)
     python c-c   markdown sql systemd shell-scripts git
     (shell :variables shell-default-shell 'eshell shell-default-height 30 shell-default-position 'bottom)
     syntax-checking version-control)
```
#### Analysis

smartparens uses `sp-show-pair-delay` to initialize a timer as soon as `show-smartparens-mode` is activated in some buffer. When `dotspacemacs/user-config` runs, that timer is already set up. So setting `sp-show-pair-delay` in `dotspacemacs/user-config` is too late: the timer is never updated.

But `spacemacs-editing/init-smartparens` sets `sp-show-pair-delay`, and that seems to happen after `dotspacemacs/user-init`. So setting it in `dotspacemacs/user-init` is too early.

The default is 0.125, which spacemacs changes to 0.2. Maybe spacemacs can keep
the default, so I can change it in `dotspacemacs/user-init`? Or it could introduce `dotspacemacs-show-pair-delay`, and initialize `sp-show-pair-delay` to that (much like how `which-key-idle-delay` is initialized to `dotspacemacs-which-key-delay`)?

For now, I work around this by calling `(timer-set-idle-time
sp-show-pair-idle-timer sp-show-pair-delay t)` in `dotspacemacs/user-config`. But that seems a little clunky.

marienz:
2b7893ee1780b0c5727064ee6b9805d1b49d9a3f fixed this.

PR #7399, opened 2016-10-14
Fix for #7383

mahinshaw:
I added a var for setting `sp-show-pair-delay`, called `dotspacemacs-sp-show-pair-delay`.

### DIFF
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -304,6 +304,10 @@ derivatives. If set to `relative', also turns on relative line numbers.")
   over any automatically added closing parenthesis, bracket, quote, etc…
   This can be temporary disabled by pressing `C-q' before `)'. (default nil)")
 
+(defvar dotspacemacs-sp-show-pair-delay 0.2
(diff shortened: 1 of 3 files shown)


TheBB:
Maybe a more descriptive description?

mahinshaw:
A better description of the var?

TheBB:
Not necessarily, just something that tells a new user browsing their new dotfile what it does more or less.

syl20bnr:
Thank you for the PR. We usually try to not duplicate existing variables, so it would be better to update the DOCUMENTATION.org file about the vanilla variable `sp-show-pair-delay`.

mahinshaw:
@syl20bnr - This was in response to #7383. The issue was around initializing the `sp-show-pair-delay`.  Based on the bug report, setting it in user-init was too early, and user-config, was too late. Hence the shadowed variable.  If there is another approach, I am happy to alter or close this PR.

TheBB:
I pushed this fix instead. https://github.com/syl20bnr/spacemacs/commit/2b7893ee1780b0c5727064ee6b9805d1b49d9a3f

Thanks though!

PR #2583, opened 2015-08-08
python: use semantic for imenu only when semantic-mode is on

bmag:
Fixes #2569 
Sets `imenu-create-index-function` to `python/imenu-create-index-python-or-semantic` (new function) instead of `semantic-create-imenu-index` in python buffers. The new function knows when to use `python-imenu-create-index` and when to use `semantic-create-imenu-index`, so imenu always finds the correct entries.

PR #7405, opened 2016-10-15
Re-fix 2569

### LINKED ISSUE #2569: Imenu in python buffers stops working without semantic layer
Steps to reproduce:
1) make sure python layer is enabled and semantic layer is disabled in your `.spacemacs` file
2) start spacemacs
3) open a python file
4) hit <kbd>SPC s l</kbd> (`spacemacs/jump-in-buffer`). A helm buffer with some imenu entries pops up.
5) close the helm buffer and open a new python file
6) hit <kbd>SPC s l</kbd>. Expected behavior: same as in step 4. Actual behavior: there are no imenu entries.

Looks like imenu can't find any entries in any python file I open from here on. I've checked the value of `imenu-create-index-function`, which generates the imenu entries. In the first file its value is `python-imenu-create-index`, but in the second file its value is `semantic-create-imenu-index` - despite having disabled the semantic layer.

If I have the semantic layer enabled, I don't encounter this bug.

I'll research this bug further later.

bmag:
Use `spacemacs/python-imenu-create-index` to create imenu's index with semantic, if semantic-mode is enabled, or with python-mode's default method if semantic-mode is disabled.

In Emacs 24.5 and older, requiring `semantic` changes imenu's index function in python-mode to use semantic even if semantic-mode is disabled, which is bad, so we fix that with an advice.

Fixes #2569. This is a modified version of #2583, which was merged already, but the changes were removed at some point for no reason.

### DIFF
--- a/layers/+lang/python/funcs.el
+++ b/layers/+lang/python/funcs.el
@@ -88,9 +88,17 @@ Anaconda provides more useful information but can not do it properly
 when this mode is enabled since the minibuffer is cleared all the time."
   (semantic-idle-summary-mode 0))
 
-(defun spacemacs//python-imenu-create-index-use-semantic ()
+(defun spacemacs//python-imenu-create-index-use-semantic-maybe ()
   "Use semantic if the layer is enabled."
-  (setq imenu-create-index-function 'semantic-create-imenu-index))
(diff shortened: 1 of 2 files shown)


TheBB:

PR #7407, opened 2016-10-15
Add an example for non-standard node bin locations

futuro:
If you install your node bins in places outside of the system path, you
need to add those locations to the exec path before packages are loaded
or else spacemacs won't recognize them as existing when it initializes
the layers.


syl20bnr:
Thank you !

PR #7408, opened 2016-10-15
Update comment rules for evilified state

### LINKED ISSUE #5738: Rebinding rules for evilified-state are not documented
https://github.com/syl20bnr/spacemacs/blob/master/layers/%2Bdistributions/spacemacs-base/local/evil-evilified-state/evil-evilified-state.el#L32 says

```elisp
;; The shadowed original mode key bindings are automatically reassigned
;; following a set of rules:
```

However, that's the end of the comment, the rules are not listed anywhere.

EDIT: [@syl20bnr] updated broken link

philtothetop:
Added keybind rules for the evilified state, as requested in issue #5738 

Fixes #5738 
* Pointed out which keys are using as in Evil
* Pointed out the binding order pattern
* Added that the g binding will be directly bound to C-G


TheBB:

PR #7423, opened 2016-10-16
Fix org markup in the changelog

mineo:


d12frosted:

Issue #7427, opened 2016-10-17
When using ivy layout transient state stays open after selecting layout.


gilbertw1:
#### Description :octocat:
Layouts transient state does not close after pressing 'l' to select / create a layout when using ivy. After selecting a layout or creating a new one, the transient state stays open instead of closing as it does when using helm.

#### Reproduction guide :beetle:
- Start Emacs (using ivy)
- Press ```SPC l l```
- Select layout

*Observed behaviour:* :eyes: :broken_heart:
Layout is opened, but transient state is still active and needs to be closed before a buffer is focused.

*Expected behaviour:* :heart: :smile:
Layout is opened and a buffer is currently focused.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 25.1.1
- Spacemacs: 0.200.2
- Spacemacs branch: master (rev. 7bdde17)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
(python emacs-lisp
        (spacemacs-layouts :variables layouts-enable-autosave t layouts-autosave-delay 300)
        git
        (markdown :variables markdown-live-preview-engine 'vmd)
        (org :variables org-enable-github-support t)
        games gnus dash ivy restclient
        (shell :variables shell-default-height 30 shell-default-position 'bottom)
        syntax-checking version-control xkcd auto-completion rust javascript
        (scala :variables scala-auto-start-ensime nil)
        haskell scheme
        (ranger :variables ranger-show-preview t ranger-cleanup-on-disable t)
        (clojure :variables clojure-enable-fancify-symbols t)
        themes-megapack github
        (evil-snipe :variables evil-snipe-enable-alternate-f-and-t-behaviors t)
        sql html)
```

TheBB:
Should be fixed in master.

PR #7430, opened 2016-10-17
Fixed layout transient state not closing on layer select in ivy layer

gilbertw1:
Close layouts transient state after pressing 'l' to list layouts in ivy. This makes more sense than leaving the transient state open and matches the helm functionality. I use ```SPC l l``` frequently to jump between my layouts and the extra step of having to explicitly exit the transient state constantly throws me off.

This fixes:  #7427


TheBB:

PR #7438, opened 2016-10-17
Diminish purpose-mode

TheBB:
I made a dedicated segment in Spaceline for this, so let's diminish the minor mode lighter. Will merge when MELPA updates.

cc @bmag


TheBB:
In.

PR #7447, opened 2016-10-17
Rust layer - Fix error message about no eldoc support 

hicksy994:
There is an error when opening rust files that says there is "no ELDoc support in this buffer", however eldoc always worked anyway. This gets rid of that error message.


syl20bnr:
Thank you ! 👍 

PR #7448, opened 2016-10-17
Properly override purpose-mode-map

bmag:
Setting `purpose-mode-map` to nil before loading `window-purpose` didn't work, because several packages require `window-purpose` before we get to it. Setting to nil with `setcdr` is more reliable and doesn't depend as much on loading order (just need to run it after `window-purpose` was loaded).
cc @ksjogo


TheBB:

PR #7456, opened 2016-10-17
Remove trailing whitespace from all files

robbyoconnor:
Cosmetic change, strip trailing whitespace from all files.


deb0ch:
Trailing whitespaces are dirty :poop:

d12frosted:
Despite the fact that they are... well... white...

robbyoconnor:
I actually had to script this -- sed is amazing.

core/core-funcs.el
```diff
@@ -324,7 +324,7 @@ current window."
   (let ((comint-buffer-maximum-size 0))
     (comint-truncate-buffer)))
 
-
+
```
syl20bnr: Please update the script to keep the page breaks `^L` ;-)

syl20bnr: Please update the script to keep the page breaks ^L ;-)

robbyoconnor: Okay, will do.

robbyoconnor:
@syl20bnr done -- this should be scripted -- and we REALLY should have a git hook to keep this from happening again.

core/tools/spacefmt/tests/test.org
```diff
@@ -21,7 +21,7 @@
...
 * BAR
-** SUBBAR    ;     , .  
+** SUBBAR    ;     , .
```
robbyoconnor: LOL

robbyoconnor:
I used this: 

```sh 
find . -type f \( -name "*.md" -o -name "*.org" -o -name "*.txt" -o -name "*.el" \)  -not -path ./elpa/* -not -path ./.cache/* -not -path ./.cask/*  -not -path ./.git/* -not -path ./session.* -exec perl -p -i -e "s/[ \t]$//g" {} \;
```

core/tools/spacefmt/tests/test.org
syl20bnr: Is this whitespace in the test file on purpose ?

robbyoconnor: If it was -- I can fix it and adjust my command to exclude this file

robbyoconnor: I added it back.

JAremko:
> we REALLY should have a git hook to keep this from happening again.

@robbyoconnor  But will a git-hook with `find` work on Windows? Also we probably can get a list of modified files(or even file fragments) and apply it to them.  This way we can avoid conflicts, when two PRs automatically try to remove the same whitespace. And it probably will be the best if the modification will be in the separate commit.

robbyoconnor:
@JAremko -- Nope, it won't. That's just what I used -- if you look at it -- for every match it runs a perl script :-P

robbyoconnor:
I'm not sure how it'd work on windows...We really need to spotcheck PRs.  Like we need to have people run `SPC x  d w` on their files when they make PRs.

robbyoconnor:
I'm not sure how it'd work on windows...We really need to spotcheck PRs.  Like we need to have people run `SPC x  d w` on their files when they make PRs.  We COULD do it in fail if it finds any files using: 

```
find . -type f \( -name "*.md" -o -name "*.org" -o -name "*.txt" -o -name "*.el" \)  ! -path ./elpa/* ! -path ./.cache/* ! -path ./.cask/*  ! -path ./.git/* ! -path ./session.* ! -path core/tools/spacefmt/tests/test.org -exec perl -p -i -e "s/[ \t]$//g" {} \;
``` 

With a pre-commit hook, it would abort the commit operation if trailing whitespace exists. We can PROBABLY assume `sed` exists and use that.

JAremko:
@robbyoconnor I thinkg I know how to do it :stuck_out_tongue: But it will be preatty insane. 

We can create a bot account for Spacemacs and add its credentials to the Travis build script that will be ran only when someone with write privilege pushes changes to the develop branch (only those accounts have accsess to the secure varable that will hold bot's  credentials)
The script will run `spacefmt`  and other tools. Then it will run git status and if there are any changes it will add extra commit with formatting and stuff, or open a PR (for example, if  there is a conflict)

I was thinknig about making something like this to autmaticaly open PRs for spacemacs.org  when the documentation updates.


*I really need to finish this spacemacs-docker thingy ASAP. And start doing something useful :smile: *

robbyoconnor:
@JAremko -- You need to run this on all text, emacs lisp, org, and markdown files as well. People introduce trailing whitespace in all of those.

JAremko:
@robbyoconnor  sure. We need to expand `spacefmt`. And fix #7675

deb0ch:
@JAremko I think the idea of a githook is good, but IMHO it should just reject the push instead of adding a commit or a PR by itself. It doesn't sound very safe.

robbyoconnor:
the git hook should fix the trailing whitespace, but not generate a commit.

JAremko:
@robbyoconnor  @deb0ch  It can be in a hook but it will may be hard to make cross platform. Also it may be confusing with some github clients or the hook can be disabled. We can check it in a Travis script as a simple test(This way we don't need to care about compatibility) or make it produce PR with fixes - the biggest benefit of it will be that it will reduce the friction for contributors.

JAremko:
>  We can PROBABLY assume sed exists and use that. 

@robbyoconnor    It's is [gsed](https://github.com/syl20bnr/spacemacs/blob/master/core/tools/spacefmt/spacefmt#L24)  on MacOS and I doubt it exist on Windows in any shape or form :disappointed:

JAremko:
Actually, since we already have [the "wrong branch"](https://github.com/syl20bnr/spacemacs/blob/master/.travis-build.sh#L26) check in Travis, it may be even better if we extand it and insted of warning a contributer in the travis build log (that some people apparently don't read)  the bot will add a message to the PR thread with the list of the issues that have to be solved. Or even solve them and open a PR to the PR source fork with the fixes and then the contributor can review them and merge.

robbyoconnor:
@JAremko whatever solution we come up with...not having to do this again would be great :)

robbyoconnor:
@syl20bnr changes have been made :) I intentionally excluded @JAremko's test file. Let's think about an automated way of doing this. Trailing whitespace offends me greatly.

robbyoconnor:
Some more changes -- When merging PRs -- let's check for trailing whitespace folks! 

@d12frosted, @syl20bnr

JAremko:
@robbyoconnor  ok..  Doing it next :shipit:

robbyoconnor:
@JAremko -- I excluded your test org file :)

syl20bnr:
> I even created an account for the bot to honor the memories of @StreakyCobra https://github.com/robocobra

😂  I love the idea!

cc @StreakyCobra

syl20bnr:
Thank you ! 💜 
Very nice cleanup, I have some work to do in my house too if you are interested 😆 

robbyoconnor:
@syl20bnr -- unless that work can be scripted like this, nah. :)

StreakyCobra:
> I even created an account for the bot to honor the memories of @StreakyCobra https://github.com/robocobra

:joy: 

> But then it went rogue and started trolling my comrades

I'm not really surprized :thinking:

PR #7462, opened 2016-10-18
purpose: disable kill extension

### LINKED ISSUE #7460: Purpose mode make company mode really slow...
image

When I disabled purpose-mode, my company mode behaviors normal now.

@syl20bnr  Any workaround? thanks.

bmag:
`purpose-x-kill-setup` enables purpose's "kill" extension, which makes `kill-buffer` aware of purpose when looking for a new buffer to display in a window instead of the killed buffer. This is done by overriding `replace-buffer-in-windows`, and recently I have noticed that it can slow down some operations, so we should not enable it by default.

<s>Probably</s> fixes #7460


d12frosted:

Issue #7453, opened 2016-10-17
Improve purpose configuration


agzam:
#### Description :octocat:

Things won't stay in the same window.
#### Reproduction guide :beetle:
- Start Emacs
- Create couple of layouts (persp)
- Create vertical and horizontal window splits
- try changing buffers (especially to non-dedicated ones)
- try jumping to dired from a buffer

_Observed behaviour:_ :eyes: :broken_heart:
Things, instead of popping up in the same window (where the cursor is) often would open in other-window 
Even `SPC Tab` would jump to buffer in other window instead of opening previously used buffer in the same window
#### Additional info:

This is happening only in develop branch I have tried master and there everything works as it used to be. 
#### System Info :computer:
- OS: darwin
- Emacs: 25.1.1
- Spacemacs: 0.200.2
- Spacemacs branch: master (rev. 7bdde17)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:

``` elisp
(csv yaml
     (org :variables org-enable-reveal-js-support t)
     (auto-completion :variables auto-completion-tab-key-behavior 'cycle)
     emacs-lisp javascript html markdown shell-scripts clojure lua react
     (auto-completion :variables auto-completion-enable-snippets-in-popup t auto-completion-enable-help-tooltip t auto-completion-enable-sort-by-usage t)
     syntax-checking spell-checking helm
     (shell :variables shell-enable-smart-eshell t shell-default-term-shell "/bin/zsh")
     (deft :variables deft-directory "~/Google Drive/deft" deft-auto-save-interval 60)
     fasd ibuffer osx chrome
     (semantic :disabled-for emacs-lisp)
     restclient emoji ranger search-engine imenu-list pdf-tools docker git
     (version-control :variables version-control-diff-tool 'diff-hl version-control-global-margin t)
     (github :variables gist-view-gist t)
     agzam ag-dash synonyms)
```

bmag:
This is because of the new `spacemacs-purpose` layer (originally #1958). See the [layer's README](https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Bspacemacs/spacemacs-purpose/README.org) and [window-purpose](https://github.com/bmag/emacs-purpose) for details.

TheBB:
I'm confused though because the report claims @agzam is on master, but the symptoms perfectly match window-purpose behaviour.

bmag:
[...]

PR #7472, opened 2016-10-18
New dotvariable to control switch-to-buffer behavior

bmag:
With a nil value for `dotspacemacs-switch-to-buffer-prefers-purpose`, switch-to-buffer prefers using the current window, same as vanilla Emacs. With a non-nil value, switch-to-buffer prefers another window with the same purpose as the buffer. This affects actions like `spacemacs/alternate-buffer`, and opening buffers from Dired.

I chose `nil` as the default value so that `switch-to-buffer` stays roughly the same as vanila Emacs, and users who are familiar with Emacs don't get surprised (as in #7453). Personally I think setting to `dotspacemacs-switch-to-buffer-prefers-purpose` to `t` is better, I'm just not sure if it's good as the default behavior for unsuspecting users.

### DIFF
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -248,6 +248,14 @@ key sequence. Setting this variable is equivalent to setting
 right, and right-then-bottom. The last one will display on the
 right if possible and fallback to bottom if not.")
 
+(defvar dotspacemacs-switch-to-buffer-prefers-purpose nil
+  "Control where `switch-to-buffer' displays the buffer.
+If nil, `switch-to-buffer' displays the buffer in the current
+window even if another same-purpose window is available. If non
+nil, `switch-to-buffer' displays the buffer in a same-purpose
+window even if the buffer can be displayed in the current
+window.")
+
 (defvar dotspacemacs-loading-progress-bar t
   "If non nil a progress bar is displayed when spacemacs is loading. This
 may increase the boot time on some systems and emacs builds, set it to nil
--- a/core/templates/.spacemacs.template
+++ b/core/templates/.spacemacs.template
@@ -214,6 +214,12 @@ values."
    ;; right; if there is insufficient space it displays it at the bottom.
    ;; (default 'bottom)
    dotspacemacs-which-key-position 'bottom
+   ;; Control where `switch-to-buffer' displays the buffer. If nil,
+   ;; `switch-to-buffer' displays the buffer in the current window even if
+   ;; another same-purpose window is available. If non nil, `switch-to-buffer'
+   ;; displays the buffer in a same-purpose window even if the buffer can be
+   ;; displayed in the current window. (default nil)
+   dotspacemacs-switch-to-buffer-prefers-purpose nil
    ;; If non nil a progress bar is displayed when spacemacs is loading. This
    ;; may increase the boot time on some systems and emacs builds, set it to
    ;; nil to boost the loading time. (default t)
--- a/layers/+spacemacs/spacemacs-purpose/packages.el
+++ b/layers/+spacemacs/spacemacs-purpose/packages.el
@@ -109,6 +109,25 @@
       (purpose-mode))
     :config
     (progn
+      ;; change `switch-to-buffer' display preferences according to
+      ;; `dotspacemacs-switch-to-buffer-prefers-purpose'. This affects actions
+      ;; like `spacemacs/alternate-buffer', and opening buffers from Dired
+      (setcdr (assq 'switch-to-buffer purpose-action-sequences)
+              (if dotspacemacs-switch-to-buffer-prefers-purpose
+                  '(purpose-display-reuse-window-buffer
+                    purpose-display-reuse-window-purpose
+                    purpose-display-maybe-same-window
+                    purpose-display-maybe-other-window
+                    purpose-display-maybe-other-frame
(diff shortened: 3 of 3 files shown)


syl20bnr:
I was getting familiar with purpose before making my mind on this topic.

`SPC Tab` key binding is `switch to last buffer in current window`. We don't have an official binding to "switch to last window" (could be `SPC w Tab`).
Purpose is combination of both which is confusing for anyone new as it is not POLA and I would prefer to not complicate things by providing an option to change the simple and clear goal of `SPC Tab`.
What we can do instead is `SPC r Tab` for `switch to last purpose`, this command should _not_ change the buffer.
This way users can combine both actions very easily, create macros if they like, we just provide the simple a to the point bindings and actions.

syl20bnr:
What I wrote above is just a proposal, I'm not sur yet.

@bmag 
Here is an example of what I don't like with purpose right now, maybe you have a solution for this:
- I have a split with two code window
- I change to message buffer in one split
- Now I cannot open a code file in the window with message and cannot go back easily to the previous purpose of the window

Purpose _must_ be transparent if a user does not use it, so the design of the integration should not force one to use `SPC r`. Just like `SPC l`is purely optional.

bmag:
The violation of POLA is because Purpose changes the behavior of `switch-to-buffer`, and many actions (including `SPC TAB` and `find-file`) use `switch-to-buffer` and thus are affected. This PR changes back the behavior of `switch-to-buffer`, so it almost always uses the current window, as expected. The dotvariable is for users that prefer the current behavior (the surprising one), but is not necessary. The important part is to run something akin to this:
```elisp
(setcdr (assq 'switch-to-buffer purpose-action-sequences)
        '(purpose-display-maybe-same-window
          purpose-display-reuse-window-buffer
          purpose-display-reuse-window-purpose
          purpose-display-maybe-other-window
          purpose-display-maybe-other-frame
          purpose-display-maybe-pop-up-window
          purpose-display-maybe-pop-up-frame))
```

In `SPC TAB` we also have a bug that we don't pass the `force-same-window` argument, which we should. Nobody ran into it yet because using dedicated windows is rare. Even if you disable `purpose-mode`, and then run `SPC w t` (dedicate window) and `SPC TAB`, you'll see the alternate buffer opening in another window.

>  What we can do instead is `SPC r Tab` for `switch to last purpose`, this command should *not* change the buffer.

The purpose of a window is defined by the purpose of its current buffer, and each buffer has a single purpose (determined by Purpose's configuration, which is supposed to be reasonable by default). Changing a window's purpose must change its buffer (or the buffer's purpose, which is a bit more difficult).

syl20bnr:
> The purpose of a window is defined by the purpose of its current buffer, and each buffer has a single purpose (determined by Purpose's configuration, which is supposed to be reasonable by default). Changing a window's purpose must change its buffer (or the buffer's purpose, which is a bit more difficult).

Indeed it defeats the purpose of purpose :-)

Forget the my initial proposal.

We can try to find a way to make purpose transparent for users without the need to introduce a variable to modify the behavior of `SPC Tab`.

Maybe by default, Spacemacs could open everything in a general purpose ?
Using `SPC r` then allow to open buffer in a purpose other than general ?

We must have the purpose lighter in the modeline, would be could to make it very short, maybe using colors instead of a name ? Bring your ideas! :-D

syl20bnr:
> Maybe by default, Spacemacs could open everything in a general purpose ?
Using SPC r then allow to open buffer in a purpose other than general (can be predefined too) ?

Actually I love this apporach, it is exactly the same approach as spacemacs layouts.

bmag:
> We can try to find a way to make purpose transparent for users without the need to introduce a variable to modify the behavior of SPC Tab.

I can remove the dotvariable and just keep what I referred to as "the important part" in my previous comment.

> Maybe by default, Spacemacs could open everything in a general purpose ?
> Using SPC r then allow to open buffer in a purpose other than general (can be predefined too) ?

That's not exactly how Purpose is supposed to work. Purpose has a configuration that maps a single purpose to each buffer, depending on the buffer's name and major mode. The idea is that you have a configuration that tells Purpose what buffers belong together, you arrange the frame as you like (or load a predefined window template (what Purpose calls a "window layout")), and from then on the frame's window arrangement doesn't change (except for pop up windows). Code stays in the code window, REPL stays in the REPL window, etc.

That said, I can make all the buffers general by default, and I can write commands for interactively changing the purpose configuration. I don't think it's the right way to use Purpose, but I can make it possible.

bmag:
Oh, I forgot to add: in any case, Purpose won't be transparent for users who customize `display-buffer-alist`, and offers `purpose-special-action-sequences` instead. It also affects packages, but AFAIK the only packages that use `display-buffer-alist` are helm, popwin and imenu-list, all of which are taken care of already. (helm has a fix in Purpose upstream, popwin is replaced by purpose-popwin, imenu-list is aware of Purpose). I'm going to fix this in Purpose 2.0, but that is probably several months away.

syl20bnr:
I merged this one since it make purpose more "transparent", for sure I don't know yet how purpose will be integrated, let's play with it and feel its power first :-D One thing I feel though is that `spacemacs-layouts` is like the central point of integration for persp/eyebrowse/purpose, I mean `spacemacs-layouts` is a single feature composed on these 3 great packages.

syl20bnr:
Thank you ! 👍 

PR #7479, opened 2016-10-19
Do not override user customisation of projectile on windows.

puzl:
[Edit TheBB: removed text that specifically says to remove]

### DIFF
--- a/layers/+distributions/spacemacs-base/packages.el
+++ b/layers/+distributions/spacemacs-base/packages.el
@@ -321,8 +321,12 @@
       ;; note for Windows: GNU find or Cygwin find must be in path to enable
       ;; fast indexing
       (when (and (spacemacs/system-is-mswindows) (executable-find "find"))
-        (setq  projectile-indexing-method 'alien
-               projectile-generic-command "find . -type f"))
+        (progn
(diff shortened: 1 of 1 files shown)


TheBB:
Why though? Is it a problem to set these variables in dotspacemacs/user-config?

TheBB:
See https://github.com/syl20bnr/spacemacs/pull/5168 for discussion about a more comprehensive solution to the “customize problem”. Checking for boundp might not be sufficient.

puzl:
I really like using my customisation file because I set it to live outside of the .emacs.d directory.  I can then version control it and have it survive reinstalls of spacemacs etc.

I agree replacing the setq calls everywhere is laborious.  

Executing the custom-file after user-config has run might work.  Wouldn't it be better to have spacemacs do this automatically, whether or not custom-file has been modified by the user.

TheBB:
Of course, that is what I am proposing.

TheBB:
https://github.com/syl20bnr/spacemacs/pull/7483

syl20bnr:
Thank you for the PR, I pushed a commit which delays the application of custom settings, they are now applied after `user-config`.

See commit: https://github.com/syl20bnr/spacemacs/commit/e699f18

PR #7480, opened 2016-10-19
Cleaned up directory warning about duplicate layers.

puzl:
Cleaned up a warning that occurs when loading private layers
The layer discovery loop scans the private layer directory (
~/.spacemacs.d/layers) AND also ~/.spacemacs.d
The code then checks if a layer directory already exists
but the comparison fails due to trailing / in some paths.

The fix is to strip any trailing slash from both directories being
compared.

[Edit TheBB: removed text that specifically says to remove]


TheBB:
Might be more robust to use `directory-file-name`.

puzl:
Agreed, I'l make the change and make a new PR

PR #7482, opened 2016-10-19
Cleanup warning about duplicate layers

puzl:
Changed my previous PR to use 'directory-file-name' instead of regex replacing the trailing /.

This PR will cleanup the warnings produced for any private layers under ~/.spacemacs.d/layers


TheBB:
I don't understand why the layer system checks these directories multiple times, but the PR looks all right enough. Thanks! Squashed and cherry-picked in develop. You can safely delete your branch.

puzl:
Thanks.

syl20bnr:
We were adding some directories multiple times, I made some cleanup in commit: f5957eb

PR #7483, opened 2016-10-19
Reliable loading of custom-file after load

TheBB:
This lets people use customize to reliably override stuff, even variables that are set in Spacemacs. People who are already modifying the value of `custom-file` should see no change. People who are relying on customize in their dotfile should probably move them, but the existing forms should continue to work as well as they are now.

Fixes: https://github.com/syl20bnr/spacemacs/issues/5170

Obsoletes quite a few other PRs I expect.

### DIFF
--- a/layers/+distributions/spacemacs-base/config.el
+++ b/layers/+distributions/spacemacs-base/config.el
@@ -153,9 +153,13 @@ It runs `tabulated-list-revert-hook', then calls `tabulated-list-print'."
 ;; Session
 ;; ---------------------------------------------------------------------------
 
-;; save custom variables in ~/.spacemacs
+;; save custom variables in ~/.emacs.d/private/custom.el or
+;; ~/.spacemacs.d/custom.el by default
 (unless (bound-and-true-p custom-file)
-  (setq custom-file (dotspacemacs/location)))
+  (setq custom-file
+        (concat (or dotspacemacs-directory
+                    configuration-layer-private-directory)
+                "custom.el")))
 ;; scratch buffer empty
 (setq initial-scratch-message nil)
 ;; don't create backup~ files
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -377,6 +377,8 @@ If NO-INSTALL is non nil then install steps are skipped."
   (dotspacemacs|call-func dotspacemacs/layers "Calling dotfile layers...")
   (setq dotspacemacs--configuration-layers-saved
         dotspacemacs-configuration-layers)
+  (when (and custom-file (file-exists-p custom-file))
+    (load-file custom-file))
   (when (spacemacs-buffer//choose-banner)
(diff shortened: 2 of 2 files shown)


layers/+distributions/spacemacs-base/config.el
```diff
@@ -153,9 +153,13 @@ It runs `tabulated-list-revert-hook', then calls `tabulated-list-print'."
 ;; Session
 ;; ---------------------------------------------------------------------------
 
-;; save custom variables in ~/.spacemacs
+;; save custom variables in ~/.emacs.d/private/custom.el or
+;; ~/.spacemacs.d/custom.el by default
 (unless (bound-and-true-p custom-file)
-  (setq custom-file (dotspacemacs/location)))
+  (setq custom-file
+        (concat (or dotspacemacs-directory
```
bmag: Better to use `expand-file-name` here

jupl:
Is there going to be any way to avoid `custom-set-variables`? Originally every time I updated packages `custom-set-variables` keeps getting written by default to my `init.el` file which I didn't want. I ended up assigning `custom-file` to a file that gets ignored. With this change I can't see any way of avoid the auto inclusion of `custom-set-variables`.

syl20bnr:
I would like to avoid imposing a split by default for people using only the dotfile. People must be able to replicate a Spacemacs config simply by sharing the dotfile. Splitting part of the dotfile by default is a regression.

OTHO I see no problem to use a different file for people using the dotdirectory. We should then restrict the split to only this case.

For the dotfile case, I propose to make the split at runtime with some temp buffer magic. Actually once we have this, we could even avoid to split physically in the directory.

Also, this way we keep the default Emacs behavior which put customize settings in init.el (which is the dotfile in our case).

syl20bnr:
Mmhh I wonder if we could encapsulate empty customize sexps in a function. Maybe customize will detect them and keep them inside the function...

syl20bnr:
I pushed a solution with commit https://github.com/syl20bnr/spacemacs/commit/e699f18
It wraps the custom sexps in a new dotspacemacs function. I took some code from customize to replace the function in the dotfile, I will add some tests later 👊 

From the commit message:

>Customize cannot write its auto-generated sexps inside a function, to
accomplish this we trick Emacs by setting the custom file to a file
in `.cache` directory, the path to this file is defined by the variable
`spacemacs--custom-file`. At the startup of Emacs we read this file
to insert its content inside the function
`dotspacemacs/emacs-custom-settings` in the dotfile, this is done in the
function `spacemacs/write-custom-settings-to-dotfile`.

> I don't think we need to write the custom settings to the dotfile when
exiting Emacs as well, since we do it at startup at the very beginning
(i.e. before actually loading the dotfile) we should be OK.

syl20bnr:
I think we have now fixed the issue with the custom file so this PR is unnecessary.

PR #7488, opened 2016-10-19
Add toggle to display time in modeline

jupl:
`SPC t m T` will toggle display time using Emacs' built-in `display-time`.

### DIFF
--- a/layers/+distributions/spacemacs-base/keybindings.el
+++ b/layers/+distributions/spacemacs-base/keybindings.el
@@ -304,6 +304,12 @@
   :off (hidden-mode-line-mode)
   :documentation "Toggle the visibility of modeline."
   :evil-leader "tmt")
+(spacemacs|add-toggle display-time
+  :status display-time-mode
+  :on (display-time-mode)
+  :off (display-time-mode -1)
+  :documentation "Display time in modeline."
+  :evil-leader "tmT")
 (spacemacs|add-toggle transparent-frame
   :status nil
   :on (spacemacs/toggle-transparency)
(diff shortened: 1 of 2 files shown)


layers/+distributions/spacemacs-base/keybindings.el
```diff
@@ -304,6 +304,12 @@
   :off (hidden-mode-line-mode)
   :documentation "Toggle the visibility of modeline."
   :evil-leader "tmt")
+(spacemacs|add-toggle display-time
+  :status display-time-mode
+  :on (display-time-mode)
+  :off (display-time-mode -1)
```
TheBB: You can replace these three lines wth `:mode display-time-mode`

jupl: I didn't know that was available. I had this in my configuration for a while and I don't remember there being a `:mode` option. Updated.

TheBB: It was added… at some point… :-P

syl20bnr:
Thank you ! 👍 
I inverted key bindings with mode-line toggle because I find it better to have the whole line mode-line toggle on a capital letter: https://github.com/syl20bnr/spacemacs/commit/7cdd998

jupl:
Perfect, I actually prefer lower case. :D

facelesspanda:
As a general question:
why isn't this in a stable patch release yet? It's been almost a year.

I ended up modifying my own keybindings.el, so I'm not complaining, I just would like to know why shipping this small improvement should depend on the advancement of other more complex development tasks.

PR #7489, opened 2016-10-19
Add toggle for syntax highlighting (via font-lock)

jupl:
I titled it Syntax Highlighting as the description would make more sense to a newbie compared to font lock. Also, [it's referenced as syntax highlighting in the documentation](https://www.gnu.org/software/emacs/manual/html_node/efaq/Turning-on-syntax-highlighting.html).


syl20bnr:
Thank you ! 👍 
I moved it under the `SPC t h` prefix: https://github.com/syl20bnr/spacemacs/commit/2c5b628

PR #7490, opened 2016-10-20
Add support for ripgrep, fixes #7370

### LINKED ISSUE #7370: Support ripgrep search
It would be nice to have support for [ripgrep](https://github.com/BurntSushi/ripgrep) for searching in spacemacs.

Unfortunately its command line options are not directly compatible with any other search tool so that it could be used as a drop in replacement for one.

agriffis:
Working on a first contribution to spacemacs, be gentle. :-)

Note this puts ripgrep _first_ in the list, ahead of the silver searcher. This
is based on my own experience of ag being buggy with long lines and large
files, and it's the reason I'm pursuing ripgrep support in spacemacs. I think
when this PR is done, it's appropriate for ripgrep to take the first priority
slot.

Naturally up for discussion. ;-)
- [x] add to dotspacemacs-search-tools
- [x] add to documentation
- [x] add functions and bindings to helm layer
- [x] add functions and bindings to ivy layer
- [x] fix the regex joining problem for multiple words https://github.com/syohex/emacs-helm-ag/pull/278 https://github.com/syohex/emacs-helm-ag/pull/280
- [x] ~~figure out if this needs to depend on helm v2.3.0 which [added ripgrep support](https://github.com/emacs-helm/helm/pull/1625) as @koterpillar [mentioned](https://github.com/syl20bnr/spacemacs/issues/7370#issuecomment-254371441)~~ (I don't think this has any dependency on helm v2.3.0 since that code is distinct from the ag/rg support in helm-ag.)

### DIFF
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -326,9 +326,9 @@ to aggressively delete empty lines and long sequences of whitespace, `trailing'
 to delete only the whitespace at end of lines, `changed' to delete only
 whitespace for changed lines or `nil' to disable cleanup.")
 
-(defvar dotspacemacs-search-tools '("ag" "pt" "ack" "grep")
+(defvar dotspacemacs-search-tools '("rg" "ag" "pt" "ack" "grep")
   "List of search tool executable names. Spacemacs uses the first installed
-tool of the list. Supported tools are `ag', `pt', `ack' and `grep'.")
+tool of the list. Supported tools are `rg', `ag', `pt', `ack' and `grep'.")
 
 (defvar dotspacemacs-default-package-repository 'melpa-stable
   "The default package repository used if no explicit repository has been
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -2289,10 +2289,11 @@ You can toggle auto-save in a buffer by calling the command =auto-save-mode=.
 ** Searching
 *** With an external tool
 Spacemacs can be interfaced with different search utilities like:
-  - ack
-  - grep
+  - [[https://github.com/BurntSushi/ripgrep][rg]]
   - [[https://github.com/ggreer/the_silver_searcher][ag]]
   - [[https://github.com/monochromegane/the_platinum_searcher][pt]]
+  - [[http://beyondgrep.com/][ack]]
+  - grep
 
 The search commands in Spacemacs are organized under the ~SPC s~ prefix with the
 next key is the tool to use and the last key is the scope. For instance ~SPC s a
@@ -2304,9 +2305,9 @@ a B~ will search with symbol under point (if there is no active region).
 
 If the tool key is omitted then a default tool will be automatically selected
 for the search. This tool corresponds to the first tool found on the system of
-the list =dotspacemacs-search-tools=, the default order is =ag=, =pt=, =ack=
-then =grep=. For instance ~SPC s b~ will search in the opened buffers using =pt=
-if =ag= has not been found on the system.
+the list =dotspacemacs-search-tools=, the default order is =rg=, =ag=, =pt=,
+=ack= then =grep=. For instance ~SPC s b~ will search in the opened buffers
+using =pt= if =rg= and =ag= have not been found on the system.
 
 The tool keys are:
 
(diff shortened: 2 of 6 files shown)


agriffis:
This presently breaks in the same way that @swaroopch [mentioned](https://github.com/syl20bnr/spacemacs/issues/7370#issuecomment-254371193):

screenshot from 2016-10-19 23-07-33

so I'm looking into that now

agriffis:
Here's where the regex is constructed

https://github.com/syohex/emacs-helm-ag/blob/49e1f66fa80674513ca898e32d62d6dad875cb90/helm-ag.el#L855

bmag:
helm-ag [recommends](https://github.com/syohex/emacs-helm-ag#ripgrep) `"rg --vimgrep --no-heading"`, but you chose `"rg --smart-case --no-heading"`. Why?

agriffis:
@bmag Thanks for looking! Adding `--vimgrep` is the right thing to do, and I'll update the PR for that. I left it out initially because I wanted to reconcile the parameter sets recommended by helm versus helm-ag. I've read helm-ag now and see how it modifies its behavior internally for `--vimgrep`

Update on the regex problem: helm-ag uses forward lookahead assertions to match multiple words on a line in any order. Ripgrep's regex engine doesn't support lookahead assertions. However helm-ag already handles this for pt with the -e option (which has similar restrictions) so it's just a matter of extending that for ripgrep. I'll make a PR against helm-ag for this.

agriffis:
See helm-ag PR at https://github.com/syohex/emacs-helm-ag/pull/278

Unfortunately my helm-ag PR has a bug but I'm not sure what it is. Tests fail as a result. Anybody with more elisp experience please feel free to review and debug!

agriffis:
@bmag Updated with `--vimgrep`. Also I forgot to address `--smart-case` in my earlier response. I added that because it's the default on `ag` and seems like a sensible default for `rg` too. In other words, `helm-ag` didn't explicitly recommend it but their recommendation would result in different behavior between `ag` and `rg` out of the box.

agriffis:
All done here except waiting for the helm-ag PR (and subsequent release, I guess)

bmag:
I haven't tried `rg` yet, but it looks good :+1:

agriffis:
[helm-ag PR was merged](https://github.com/syohex/emacs-helm-ag/pull/280) and [released in 20161025.809](https://melpa.org/#/helm-ag)

@bmag Unless spacemacs needs something specific to call out the version dependency on helm-ag, this PR should now be ready to merge

bmag:
@agriffis I've tried it for a bit and it's really nice :smile: Behaves more or less like `ag` on my machine. Can you squash your commits, though?

agriffis:
@bmag squashed!

bmag:

Swoorup:
How I do specify file types though. Using `-t cpp blah` does not work. 
It gives `unrecognized file type: cpp.*bleh`

syohex:
@Swoorup Could you try it without space like `-tcpp blah` or use `=` as separator like `--type=cpp blah` ?

Swoorup:
No clue how but 
`-tcpp blah` worked! Could this be added to the documentation?

syohex:
Will do(at https://github.com/syohex/emacs-helm-ag)

izahn:
It's already documented, in the ripgrep man page.

sixshotx:
Is it possible to support the .ignore file when using rg in helm? The helm-ag uses "-p" option which is different from rg by default.

pkoch:
Just installed spacemas and I don't see this in master. Is that on purpose? When will this hit master?

duianto:
I don't know why it hasn't been included on the master branch yet, nor do I know when the next version will be released. The author is busy at the moment.

If you want to use it now, then you can install the develop branch. Just backup any current changes and remove `.spacemacs` and `.emacs.d` from the installation directory first. There has been some changes to the `.spacemacs` file on the develop branch, therefore it's best to let it create a new one.
```
git clone -b develop https://github.com/syl20bnr/spacemacs ~/.emacs.d
```
source: https://github.com/syl20bnr/spacemacs/tree/develop

It's stable.

Here are the changes that have occurred since the previous master branch.
https://github.com/syl20bnr/spacemacs/blob/develop/CHANGELOG.develop

PR #7500, opened 2016-10-20
Maximize frame earlier in the startup process

deb0ch:
Maximize frame as early as possible when `dotspacemacs-maximized-at-startup` is `t`.

This avoid this annoying effect that during all the loading and the displaying of the startup/update messages the frame is ridiculously small.

This code should work on windows as well.

### DIFF
--- a/core/core-spacemacs.el
+++ b/core/core-spacemacs.el
@@ -80,6 +80,9 @@ the final step of executing code in `emacs-startup-hook'.")
   (dotspacemacs/load-file)
   (require 'core-configuration-layer)
   (dotspacemacs|call-func dotspacemacs/init "Calling dotfile init...")
+  (when dotspacemacs-maximized-at-startup
+    (toggle-frame-maximized)
+    (add-to-list 'default-frame-alist '(fullscreen . maximized)))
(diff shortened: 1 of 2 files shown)


core/core-spacemacs.el
```diff
@@ -80,6 +80,9 @@ the final step of executing code in `emacs-startup-hook'.")
   (dotspacemacs/load-file)
   (require 'core-configuration-layer)
   (dotspacemacs|call-func dotspacemacs/init "Calling dotfile init...")
+  (if dotspacemacs-maximized-at-startup
+      (progn (toggle-frame-maximized)
+             (add-to-list 'default-frame-alist '(fullscreen . maximized))))
```
TheBB: `(if condition (progn ...))` is the same as `(when condition ...)`

deb0ch: Yep you're right, I'm not used enough to these forms :smile:

deb0ch:
By the way, it also has the nice addition that now frames created by `emacsclient -c` when running as daemon will also be maximized on creation.

syl20bnr:
Very nice 👍 
Thank you !

Issue #6693, opened 2016-07-27
Feature proposal, PR ready (ish): Moving & swapping buffers, ace-jump removal


quicknir:
This issue is to suggest an improvement and see how people feel, before pushing forward with a PR. I currently have this code working locally. The gist is to add a new set of bindings to `SPC b 1`, `SPC b 2`, etc. Without the prefix argument, this command moves the currently active buffer to the nth window (if the nth window does not exist, you get an error and nothing happens). Focus stays on the originally active _window_, i.e. does not follow the buffer. With the prefix argument, it swaps the currently active buffer with the active buffer at the nth window. Again, focus stays on the current window.

As you can see, this very nicely parallels the functionality of `SPC 1` etc. It's very concise, works nicely with the window numbering system, and together these keybindings with their optional prefix arguments cover the majority of what you would need to do other than creating new splits. In particular it covers functionality that until now was only really covered by ace window (ace-swap-window) . While ace window is a nice package, given that we are explicitly numbering all the windows anyhow, it's more convenient to use those window numberings (since you don't have to wait for the letters to pop up) and also more consistent (since `SPC 1` et al are very widely used). So I think we should remove it from the standard spacemacs distro.

A last note: what's the reasoning for not moving focus with the buffer? Well, there are cases where you'd want to do either one, so no matter which you choose the user will have to `SPC n` back to the window they want about half the time. This isn't the end of the world as that's only two more keystrokes. However, there is a particular use case: if you want to open a bunch of stuff from e.g. helm projectile, and put it all in specific windows. A very nice way to do that with the current bindings is to mark all the entries you want and hit enter; this opens all of those buffers and associates them with the current window. You can then repeatedly hit `SPC b n` to distribute/send those buffers to the exact windows you want. This is more convenient than any alternative currently available IMHO (the main current alternative is to switch to each buffer, repeatedly hitting `SPC r l` to resume your helm search, and open one file in that buffer). Of course, if focus moved each time this would be far more awkward.

bmag:
That's a very interesting suggestion, would like to see the code and try it out myself.

 > Without the prefix argument, this command moves the currently active buffer to the nth window

You didn't say what happens to the original window. Does its buffer change?

[...]

PR #7512, opened 2016-10-21
Add keybindings to move buffers by window number

quicknir:
Adds functions and keybindings that allow moving the current buffer to the nth window via SPC b n. Adding the prefix key causes a swap instead. Window focus follows the buffer, though we could later change this to a setting. Note that the associated buffers of the window are handled correctly; that is when the buffer is moved away from the window, that buffer is deleted from the window's history, so for instance SPC TAB does not switch back to the buffer you just moved away.

### DIFF
--- a/layers/+distributions/spacemacs-base/funcs.el
+++ b/layers/+distributions/spacemacs-base/funcs.el
@@ -184,6 +184,55 @@ Dedicated (locked) windows are left untouched."
   (interactive "p")
   (spacemacs/rotate-windows (* -1 count)))
 
+;; Note that this duplicates code from select-window-by-number, ideally should
+;; upstream this function into windows.el
+(defun spacemacs/get-window-by-number (i)
+  (let ((windows (car (gethash (selected-frame) window-numbering-table)))
+        window)
+    (if (and (>= i 0) (< i 10)
+             (setq window (aref windows i)))
+        window
+      (error "No window numbered %s" i))))
+
+(defun spacemacs/move-buffer-to-window (windownum follow-focus-p)
+  "Moves a buffer to a window, using the spacemacs numbering. follow-focus-p
+   controls whether focus moves to new window (with buffer), or stays on
+   current"
+  (interactive)
+  (let ((b (current-buffer))
+        (w1 (selected-window))
+        (w2 (spacemacs/get-window-by-number windownum)))
+    (unless (eq w1 w2)
+      (set-window-buffer w2 b)
+      (switch-to-prev-buffer)
+      (unrecord-window-buffer w1 b)))
+  (when follow-focus-p (select-window (spacemacs/get-window-by-number windownum))))
+
+(defun spacemacs/swap-buffers-to-window (windownum follow-focus-p)
+  "Swaps visible buffers between active window and selected window.
+   follow-focus-p controls whether focus moves to new window (with buffer), or
+   stays on current"
+  (interactive)
+  (let* ((b1 (current-buffer))
+         (w1 (selected-window))
+         (w2 (spacemacs/get-window-by-number windownum))
+         (b2 (window-buffer w2)))
+    (unless (eq w1 w2)
+      (set-window-buffer w1 b2)
+      (set-window-buffer w2 b1)
+      (unrecord-window-buffer w1 b1)
+      (unrecord-window-buffer w2 b2)))
+  (when follow-focus-p (select-window-by-number windownum)))
+
+(dotimes (i 9)
+  (let ((n (+ i 1)))
+    (eval `(defun ,(intern (format "buffer-to-window-%s" n)) (&optional arg)
+              ,(format "Move buffer to the window with number %i." n)
+              (interactive "P")
+              (if arg
+                  (spacemacs/swap-buffers-to-window ,n t)
+                (spacemacs/move-buffer-to-window ,n t))))))
+
 (defun spacemacs/rename-file (filename &optional new-filename)
   "Rename FILENAME to NEW-FILENAME.
 
@@ -907,4 +956,3 @@ a split-side entry, its value must be usable as the SIDE argument for
                                             text-scale-mode-amount) 1)
                                   (if (car (window-margins))
                                       (car (window-margins)) 1)))))
-
--- a/layers/+distributions/spacemacs-base/keybindings.el
+++ b/layers/+distributions/spacemacs-base/keybindings.el
@@ -130,6 +130,10 @@
   "bs"    'spacemacs/switch-to-scratch-buffer
   "bY"    'spacemacs/copy-whole-buffer-to-clipboard
   "bw"    'read-only-mode)
+(dotimes (i 9)
+  (let ((n (+ i 1)))
+    (spacemacs/set-leader-keys (format "b%i" n)
+      (intern (format "buffer-to-window-%s" n)))))
 ;; Cycling settings -----------------------------------------------------------
 (spacemacs/set-leader-keys "Tn" 'spacemacs/cycle-spacemacs-theme)
 ;; errors ---------------------------------------------------------------------


layers/+distributions/spacemacs-base/funcs.el
```diff
@@ -184,6 +184,49 @@ Dedicated (locked) windows are left untouched."
...
   (spacemacs/rotate-windows (* -1 count)))
 
+(defun spacemacs/move-buffer-to-window (windownum follow-focus-p)
+  "Moves a buffer to a window, using the spacemacs numbering. follow-focus-p
+   controls whether focus moves to new window (with buffer), or stays on
+   current"
+  (interactive)
+  (let ((b (current-buffer))
+        (w (selected-window)))
+    (save-selected-window
```
bmag: Saving the selected window, selecting another window and changing its buffers with `switch-to-buffer` is not the best way to change the buffer of a window. Partly because `switch-to-buffer` is not guaranteed to display the buffer in the selected window (you need to pass `force-same-window`). It is much better to use `set-window-buffer`:
```elisp
(let (;; ...
      (other-win (get-window-by-number windownum)))
  ;; ...
  (set-window-buffer other-win current-buf)
  ;; ...
  ))
```
(I'm assuming `get-window-by-number` or an equivalent function exists)

layers/+distributions/spacemacs-base/funcs.el
```diff
@@ -184,6 +184,49 @@ Dedicated (locked) windows are left untouched."
...
+  (if follow-focus-p (select-window-by-number windownum)))
+
+(defun spacemacs/swap-buffers-to-window (windownum follow-focus-p)
+  "Swaps visible buffers between active window and selected window.
+   follow-focus-p controls whether focus moves to new window (with buffer), or
+   stays on current"
+  (interactive)
+  (let ((b1 (current-buffer))
+        (w1 (selected-window)))
+    (save-selected-window
```
bmag: Also here `save-selected-window`+`select-window-by-number`+`switch-to-buffer` should be replaced by `get-window-by-number`+`set-window-buffer`.

layers/+distributions/spacemacs-base/funcs.el
```diff
@@ -184,6 +184,49 @@ Dedicated (locked) windows are left untouched."
...
+      (select-window-by-number windownum)
+      (let ((b2 (current-buffer))
+            (w2 (selected-window)))
+        (unless (eq w1 w2)
+          (switch-to-buffer b1)
+          (unrecord-window-buffer w2 b2)
+          (select-window w1)
+          (switch-to-buffer b2)
+          (unrecord-window-buffer w1 b1)))))
+  (if follow-focus-p (select-window-by-number windownum)))
```
bmag: When there is no `else` clause, `when` is better than `if` (both semantically and because it allows more than on sexp in the `then` clause).

layers/+distributions/spacemacs-base/funcs.el
quicknir: There isn't really a separate get-window-by-number function defined.

```
(defun select-window-by-number (i &optional arg)
  "Select window given number I by `window-numbering-mode'.
If prefix ARG is given, delete the window instead of selecting it."
  (interactive "P")
  (let ((windows (car (gethash (selected-frame) window-numbering-table)))
        window)
    (if (and (>= i 0) (< i 10)
             (setq window (aref windows i)))
        (if arg
            (delete-window window)
          (select-window window))
      (error "No window numbered %s" i))))
```

I could either repeat the first line where I call gethash, and do bounds checking. Or (my preference), if the problem is only with the `switch-to-buffer call`, then I can use `select-window-by-number` followed by `selected-window` to get the window, and then call `set-window-buffer`. It's a tiny bit roundabout, but it achieves the best reuse of existing code and is robust. I'd strongly prefer the latter; let me know if you are ok with it.

quicknir: Well, note that in any case it switches the window, because it follows the active buffer. So in the case that's currently being bound, it makes no difference. I did leave the boolean parameter, for future use (buffer transient state, plus maybe to create an option to control the behavior). If in the future we call without wanting to change window focus, then yes what I'm suggested will involve two window switches instead of none.

Personally, I don't like implementing `get-window-by-number`, because we can't immediately change `select-window-by-number`, so now we have pretty bad code duplication. But I will do it your way.

Note that I was trying to get my stuff merged into window-number.el at one point; I saw that the project was inactive and I asked the maintainer if I could take it over. He got back to me several months later and said "yes", by which point I was no longer interested. If you are willing, you could email him. If you become the maintainer for window-number.el then we can merge some changes and get the best of both worlds (no unnecessary switching, no code duplication).

layers/+distributions/spacemacs-base/funcs.el
```diff
@@ -184,6 +184,58 @@ Dedicated (locked) windows are left untouched."
...
+;; Note that this duplicates code from select-window-by-number, ideally should
+;; upstream this function into windows.el
+(defun spacemacs/get-window-by-number (i)
+  (let ((windows (car (gethash (selected-frame) window-numbering-table)))
+        window)
+    (if (and (>= i 0) (< i 10)
+             (setq window (aref windows i)))
+        window
+      (error "No window numbered %s" i)))
+  )
```
bmag: Didn't notice in previous version, but elisp convention is that closing parentheses don't deserve their own line. From the [elisp manual](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html):
> Don’t make a habit of putting close-parentheses on lines by themselves; Lisp programmers find this disconcerting.

It's not crucial, though.

layers/+distributions/spacemacs-base/funcs.el
```diff
@@ -184,6 +184,58 @@ Dedicated (locked) windows are left untouched."
...
+   current"
+  (interactive)
+  (let ((b (current-buffer))
+        (w1 (selected-window))
+        (w2 (spacemacs/get-window-by-number windownum)))
+    (unless (eq w1 w2)
+      (set-window-buffer w2 b)
+      (switch-to-prev-buffer)
+      (unrecord-window-buffer w1 b)
+      ))
```
bmag: Same parens comment as above.

layers/+distributions/spacemacs-base/funcs.el
```diff
@@ -184,6 +184,58 @@ Dedicated (locked) windows are left untouched."
...
+  (let* ((b1 (current-buffer))
+         (w1 (selected-window))
+         (w2 (spacemacs/get-window-by-number windownum))
+         (b2 (window-buffer w2)))
+    (unless (eq w1 w2)
+      (set-window-buffer w1 b2)
+      (set-window-buffer w2 b1)
+      (unrecord-window-buffer w1 b1)
+      (unrecord-window-buffer w2 b2)
+      ))
```
bmag: Same parens comment as above.

quicknir:
@bmag I fixed the style as you asked, for good measure. Curious, I saw that you approved the changes, but it was not merged. How does the review system work; do at least 2 maintainers have to approve or something like that? I was wanted to get started on the buffer transient mode PR, just want to be sure that this code will not undergo any non-trivial changes before I start that.

TheBB:
> but it was not merged

See label. @syl20bnr has this habit of pushing a million things in one go.

quicknir:
@TheBB I'm curious the reason for this; it's not great in the sense that it causes new work to be done on stale upstreams, and complicates multi PR projects; e.g. it would be great if I could fetch develop that already included my changes, instead of continuing to work on my branch, and rebasing in the future (and obviously I cannot actually start the new dependent PR until this one is merged).

TheBB:
Why don't you ask him? I know that we don't merge PRs as quickly as we should, but it's been three months since you posted your issue before this PR was posted, so in my opinion you have no reason to feel mistreated over a two day delay.

quicknir:
I don't feel mistreated; I am actually very impressed by @bmag's turnaround time in reviewing it, I was just curious about the reasons for the workflow, that is all.

syl20bnr:
Thank you ! These bindings are very well integrated into spacemacs 💜 
Sometime I start merging things and I cannot stop until I have not the time to close them, [waffle](https://waffle.io/syl20bnr/spacemacs) allow to easily mark issues as merged even if they have not been pushed yet. It is then easy to close the list of merged issues from the board later. I guess I should mark them as assigned to me also which I did not do.

deb0ch:
Original issue: #6693 (just for the link)

PR #7515, opened 2016-10-21
[keyboard-layout] Saner drovak defaults

Gonzih:
Hi there!

I tried to use keyboard-layout layer, but noticed that dvorak keys are off by 1 key. Instead of normal `hjkl` keys remapping uses `jkl;` keys for movement which is off from you would normally expect from vim. This is small fix for this with sane remappings for dvorak layout.

Thanks!


TheBB:
We can't really go about removing keyboard layouts. People are using them. Can't you add yours instead?

Gonzih:
Do really people use this layout that is shifted to the right by one key? I'm not trying to remove layout, I'm just trying to fix existing one. I heard about keyboard-layout some time ago and just tried to use it. It's off which is in my opinion is an issue. Feels like nobody really tested remapping before hand. Can somebody confirm that they are using misconfigured implementation?

TheBB:
Someone PR'd this and added it so I assume someone is using it. I don't mean to be too careful with the current state of Spacemacs but keyboard layouts are deeply personal things and you would IMO be in your right to be pissed off if someone messed with it. Changing it is by any practical means the same as removing it. We can add a new one, even rename this one if we feel like your idea is a more suitable default (which I agree with) but I would not like to remove the one that exist.

The reasoning behind the current implementation is that when the fingers rest on the home keys, your four right fingers are (assuming qwerty) on the `jkl;` keys, which is why many consider them to be more suitable “arrow keys” than `hjkl`. No finger rests on `h`

Gonzih:
I added new layout to not disturb any picky dvorak users :)

Kethku:
bump! I'd also say that this should be the default. The current default is different from what vim does. Since the keyboard layout layer's main purpose as far as I can tell is to allow people to use their hjkl muscle memory in other formats. It seems broken for the muscle memory to not be applicable for the dvorak keyboard. I can understand that some folks would say that having movement keys be jkl; (from qwerty's perspective) makes more sense but its simply not the default for vim and we should respect the default before the "better" option.

TheBB:

I didn't change the names. If people want to make dvp into dvorak, please open an issue about it.

PR #7527, opened 2016-10-23
Close layout transient state on 'buffer select' in ivy

gilbertw1:
Shortly after completing #7430, I noticed that the exact same problem existed when selecting the "buffer in layout" option in the layouts transient state when using ivy. That is the transient state did not close when entering 'buffer select' as it does when using helm.


syl20bnr:
Thank you ! 👍 

PR #7529, opened 2016-10-23
Overhaul transient buffer

quicknir:
Note that this PR includes my previous work, which has already been reviewed and approved. I make use of the new functions to add some nice bindings to buffer transient mode, making it useful for quickly rearranging many buffers.

When it's approved, as long as the other PR is merged before this one, the rebase should handle the duplicate commit.

### DIFF
--- a/layers/+distributions/spacemacs-base/funcs.el
+++ b/layers/+distributions/spacemacs-base/funcs.el
@@ -231,7 +231,14 @@ Dedicated (locked) windows are left untouched."
               (interactive "P")
               (if arg
                   (spacemacs/swap-buffers-to-window ,n t)
-                (spacemacs/move-buffer-to-window ,n t))))))
+                (spacemacs/move-buffer-to-window ,n t))))
+    (eval `(defun ,(intern (format "move-buffer-window-no-follow-%s" n)) ()
+             (interactive)
+             (spacemacs/move-buffer-to-window ,n nil)))
+    (eval `(defun ,(intern (format "swap-buffer-window-no-follow-%s" n)) ()
+             (interactive)
+             (spacemacs/swap-buffers-to-window ,n nil)))
+    ))
 
 (defun spacemacs/rename-file (filename &optional new-filename)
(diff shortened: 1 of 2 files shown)


syl20bnr:
You won't need to rebase because we cherry-pick instead of merging, so it is on our side to cherry-pick only the second commit and not the first one ;-)

quicknir:
@syl20bnr @d12frosted Hey guys, sorry to bother, but ping? I think this is a nice enhancement and I've been waiting for a review for a bit.

layers/+distributions/spacemacs-base/funcs.el
```diff
@@ -184,6 +184,62 @@ Dedicated (locked) windows are left untouched."
   (interactive "p")
   (spacemacs/rotate-windows (* -1 count)))
 
+;; Note that this duplicates code from select-window-by-number, ideally should
+;; upstream this function into windows.el
+(defun spacemacs/get-window-by-number (i)
+  (let ((windows (car (gethash (selected-frame) window-numbering-table)))
```
d12frosted: Where is `window-numbering-table` coming from? It's not defined for me 😢

d12frosted: And it seems that this file already contains the same definition of `spacemacs/get-window-by-number`.

quicknir:
@d12frosted Sorry for any confusion, I had pushed from this branch because the first commit contained necessary work for the second. When the first was merged, I offered to rebase this work to only have the second commit, but @syl20bnr wrote that it was not necessary in the comment above because the workflow is to cherry pick anyhow.

Would it be helpful to you if I just rebased anyhow?

d12frosted:
@quicknir oh, it's my fault, I didn't read it carefully what you have written in your message. I was more focused on the code itself. 

> Would it be helpful to you if I just rebased anyhow?

As Sylvain said - since we cherry-pick it's not needed. 

P. S. thanks for clarifications :)

d12frosted:

This works very nice 😸

PR #7530, opened 2016-10-23
More visible error in case of version mismatch

TheBB:
The current message is almost impossible to see unless you go looking for it.


syl20bnr:
Not sure when it has been pushed to develop but it's in there :-)

PR #7532, opened 2016-10-23
Add a layer for the vala programming language

Stebalien:
Includes:
- Syntax highlighting.
- ~Syntax checking.~
- Snippets (yasnippet)

### DIFF
--- a/layers/+lang/vala/packages.el
+++ b/layers/+lang/vala/packages.el
@@ -1,0 +1,34 @@
+;;; packages.el --- vala layer packages file for Spacemacs.
+;;
+;; Copyright (c) 2012-2016 Sylvain Benner & Contributors
+;;
+;; Author: Steven Allen <steven@stebalien.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+;;; Commentary:
+
+;;; Code:
+
+(defconst vala-packages
+  '(
+    vala-mode
+    yasnippet
+    (vala-snippets :toggle (configuration-layer/package-usedp 'yasnippet)))
+  "The list of Lisp packages required by the vala layer.")
+
+(defun vala/init-vala-mode ()
+  (use-package vala-mode
+    :defer t))
+
+(defun vala/post-init-yasnippet ()
+  (add-hook 'vala-mode-hook 'spacemacs/load-yasnippet))
+
+(defun vala/init-vala-snippets ()
+  (use-package vala-snippets
+    :defer t))
+
+;;; packages.el ends here
--- a/layers/+lang/vala/README.org
+++ b/layers/+lang/vala/README.org
@@ -1,0 +1,13 @@
+#+TITLE: Vala layer
+
+* Table of Contents                                        :TOC_4_gh:noexport:
+ - [[#description][Description]]
+ - [[#install][Install]]
+
+* Description
+Syntax highlighting and snippets for the [[https://wiki.gnome.org/Projects/Vala][Vala]] programming language.
+
+* Install
+To use this configuration layer, add it to your =~/.spacemacs=. You will need to
+add =vala= to the existing =dotspacemacs-configuration-layers= list in this
+file.


robbyoconnor:
Strike that -- I read syntax check...nevermind :)

bmag:
@Stebalien don't you want to include support for auto-completion and syntax-checking? Or is it somehow included already?

Stebalien:
@bmag, I don't know of any libraries that provide autoc-completion support, `flymake-vala` never worked properly for me, and none of the `flycheck-vala` libraries are on melpa.

However, there is a [`flycheck-vala`](https://marmalade-repo.org/packages/flycheck-vala) library on marmalade. What's the etiquette on adding other people's packages to melpa?

bmag:
Etiquette: see MELPA's [CONTRIBUTING.md](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.md), and their [pull request template](https://github.com/melpa/melpa/blob/master/.github/PULL_REQUEST_TEMPLATE.md). If that's problematic, you can use a recipe location in the layer's packages list.

However, I see last commit was a year ago, so I suggest you check that `flycheck-vala` even works.

Stebalien:
I've submitted an [issue](https://notabug.org/koz.ross/flycheck-vala/issues/1) on the `flycheck-vala` repo to see if the author is willing to have it added to melpa. In the mean-time, I've added a recipe that pulls directly from git.

layers/+lang/vala/packages.el
```diff
@@ -0,0 +1,47 @@
...
+                              :url "https://notabug.org/koz.ross/flycheck-vala.git")))
+  "The list of Lisp packages required by the vala layer.")
+
+(defun vala/post-init-flycheck ()
+  (spacemacs/add-flycheck-hook 'vala-mode))
+
+(defun vala/init-flycheck-vala ()
+  (use-package flycheck-vala
+    :defer t
+    :init (add-hook 'flycheck-mode-hook (lambda () (require 'flycheck-vala)))))
```
Stebalien: Is this the right way to do this? `flycheck-vala` doesn't appear to autoload any setup functions.

Stebalien:
`flycheck-vala` defines a few "safe" variables (`flycheck-vala-packages`, `flycheck-vala-enable-experimental-non-null`). Unfortunately, by the time `flycheck-vala` gets loaded, buffer-local variables have already to have been evaluated so the user ends up getting asked whether or not these these variables are safe. Is there a preferred way to handle this situation?

layers/+lang/vala/packages.el
```diff
@@ -0,0 +1,47 @@
...
+                              :url "https://notabug.org/koz.ross/flycheck-vala.git")))
+  "The list of Lisp packages required by the vala layer.")
+
+(defun vala/post-init-flycheck ()
+  (spacemacs/add-flycheck-hook 'vala-mode))
+
+(defun vala/init-flycheck-vala ()
+  (use-package flycheck-vala
+    :defer t
+    :init (add-hook 'vala-mode-hook (lambda () (require 'flycheck-vala)))))
```
Stebalien: I've changed this to use the `vala-mode-hook` instead of `flycheck-mode-hook` so that it gets loaded before file/directory local variables are read. Is this the correct way to do this?

bmag: A `require` is something that only happens once (subsequent calls don't do anything). The right thing is probably
```elisp
:init
  (require 'flycheck-vala))
```
Which I believe is the same as:
```elisp
:after flycheck
```
Update: for `vala-mode`, replace `flycheck` with `vala-mode` (or whatever is the name of the file that defines `vala-mode`)

Stebalien: Thanks, `:after flycheck` works.

tatsuhirosatou:
This would be great! I hope it gets added.

Stebalien:
So, it turns out that `flycheck-vala` doesn't support multi-file projects so it's not very useful for medium to large projects (projects where you really want on-the-fly syntax checking). I've removed it and this layer now only includes syntax highlighting and snippets.

bersace:
Hi,

I tested it, it's quite useful yet. Can we merge and iterate over this later ?

@Stebalien nice contribution :)

Regards

TheBB:

I moved it to the new `major-modes` layer since it was so small. Also applied the same fix as for matlab, to avoid having to special-case yasnippet.

PR #7535, opened 2016-10-24
don't show warning on layout key clashes

d12frosted:
Now that we show warning in the startup buffer I think it should not be considered warning when user defines his/her own layout and binds on the key with existing layout. Because, for example, not all people need Spacemacs layout bound on `e` (I use it for `environment`). And seeing warning in the startup buffer makes me think that for some reason it's not advices to rebind these layouts.


TheBB:

Issue #5737, opened 2016-04-09
evilification breaks anaconda-view-mode


mineo:
#### Description

The function that's bound to `n` in `anaconda-view-mode` (`next-error-no-select`) does not get rebound to any key.
#### Reproduction guide
- Start Emacs
- Open a python file
- Move cursor to a function definition
- `, g u` on the function name
- move to the new buffer
- press `n`

_Observed behaviour:_
- n tries to repeat the last search

_Expected behaviour:_
- Cursor moves to the next entry
#### System Info
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.14
- Spacemacs branch: master (rev. a298a5a)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:

``` elisp
((auto-completion :variables auto-completion-enable-help-tooltip t)
 (c-c  )
 emacs-lisp git github jabber
 (haskell :variables haskell-enable-ghc-mod-support t)
 html
 (latex :variables latex-build-command "LatexMk")
 markdown mineo mineo-rtags org python scala
 (shell :variables shell-default-position 'bottom shell-default-term-shell "/bin/zsh")
 (syntax-checking :variables syntax-checking-enable-tooltips nil)
 version-control)
```

mineo:
According to the rules linked to in #5738, `next-error-no-select` should be bound to `C-n` in evilified state, but it's not.

PR #7538, opened 2016-10-24
Map {j,k} to {next,previous}-error-no-select in anaconda-mode-view-mode

mineo:
This is similar to the mappings done for flycheck-error-list-mode-map
and works around the broken evilification in
anaconda-mode-view-mode (see #5737).

### DIFF
--- a/layers/+lang/python/packages.el
+++ b/layers/+lang/python/packages.el
@@ -58,7 +58,9 @@
         "gb" 'anaconda-mode-go-back
         "gu" 'anaconda-mode-find-references)
       (evilified-state-evilify anaconda-mode-view-mode anaconda-mode-view-mode-map
-        (kbd "q") 'quit-window)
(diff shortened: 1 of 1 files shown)


bmag:
IIRC anaconda's help buffer also uses `anaconda-mode-view-mode`, so this binds `j` and `k` also when the user views the documentation of some variable or function. Can you check if it's still the case, and if there's some way to bind `j` and `k` only when appropriate (e.g. viewing usages)?

mineo:
Hm, looks like you're right. Unfortunately anaconda-mode itself maps `{next,previous}-error-no-select` in anaconda-mode-view-mode (https://github.com/proofit404/anaconda-mode/blob/master/anaconda-mode.el#L716-L721), but only on C-n/C-p, which is less annoying if you only use it for viewing documentation than j/k here. I'll close this for now because I don't have the time to look into finding a better solution atm.

PR #7555, opened 2016-10-25
replaced "uninstall" with "delete" in comments on dotspacemacs-install-packages

cyberthal:
to avoid conflation of loading/ignoring with downloading/deleting.

Rationale:

The original confuses "uninstall" with "delete". The correct meaning is "delete".

Usage case study:

My goal was to avoid needless download time when switching between vanilla and modded configs. I like to switch back and forth rapidly to implement new features and track down problems. This works great with magit branch switching and lazy-load fast Spacemacs restart times. However, the bottleneck was that Spacemacs kept deleting and redownloading all the packages.

I read the original instructions above. They were unclear. I wasn't sure whether "keep-unused" would merely avoid deleting orphans, or actually load them. I had to test to verify that orphan layers were not loaded.

The description of 'all is confusing as well. I think it means that all Spacemacs layers are downloaded and activated regardless of whether they have been explicitly enabled, and that orphans are not deleted.

### DIFF
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -75,10 +75,10 @@ Paths must have a trailing slash (ie. `~/.mycontribs/')")
 (defvar dotspacemacs-install-packages 'used-only
   "Defines the behaviour of Spacemacs when installing packages.
 Possible values are `used-only', `used-but-keep-unused' and `all'. `used-only'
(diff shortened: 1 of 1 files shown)


syl20bnr:
Thank you !
I also updated the template of the dotfile which contains the same docstrings.

Issue #7546, opened 2016-10-25
Attempting to run helm-spacemacs-help results in `Symbol’s value as variable is void: fringe-mode`


jonaustin:
#### Description :octocat:

Attempting to run helm-spacemacs-help results in `Symbol’s value as variable is void: fringe-mode`

Environment:
- Mac OS X El Capitan
- ZSH / Oh My ZSH
- MacBook Pro Retina (2012)

I've tried the usual fixes for the void error - `rm -rf ~/.emacs/elpa` and restarting with `emacs --insecure`. No luck.

Also tried starting over with a brand new install (i.e. `git clone https://github.com/syl20bnr/spacemacs ~/.emacs.d
`, `rm ~/.spacemacs`), same thing.

Thanks for Spacemacs, it's great!
#### Reproduction guide :beetle:
- Start Emacs
- SPC h SPC

_Observed behaviour:_ :eyes: :broken_heart:
Simply returns above error. Nothing else happens.

_Expected behaviour:_ :heart: :smile:
It's never worked so I don't know what the expected behavior is.
#### System Info :computer:
- OS: darwin
- Emacs: 25.1.1
- Spacemacs: 0.200.3
- Spacemacs branch: master (rev. b7e51d7)
- Graphic display: nil
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:

``` elisp
(helm emacs-lisp)
```
- System configuration features: JPEG NOTIFY ACL LIBXML2 ZLIB

bmag:
Backtrace please. `SPC t D` before `SPC h SPC` will give you a backtrace when the error occurs. I suspect `fringe-mode` is not defined in non-GUI Emacs.

jonaustin:
Thanks, here you go:

Also, all the other `SPC h xxxx` commands seem to work fine.

```
Debugger entered--Lisp error: (void-variable fringe-mode)
  (equal fringe-mode 0)
  (not (equal fringe-mode 0))
  (and (or (and (symbolp (quote (not (equal fringe-mode 0)))) (boundp (quote (not (equal fringe-mode 0))))) (listp (quote (not (equal fringe-mode 0))))) (not (equal fringe-mode 0)))
  spacemacs/toggle-fringe-p()
  funcall(spacemacs/toggle-fringe-p)
[...]

PR #7564, opened 2016-10-27
Fix fringe mode void exception in terminal when calling helm-spacemacs-help

jonaustin:
This is a fix for #7546 

Note: I'm brand new to emacs, so apologies in advance if I messed something up!

### DIFF
--- a/layers/+distributions/spacemacs-base/keybindings.el
+++ b/layers/+distributions/spacemacs-base/keybindings.el
@@ -285,7 +285,7 @@
   :documentation "Toggle display of backtrace when an error happens."
   :evil-leader "tD")
 (spacemacs|add-toggle fringe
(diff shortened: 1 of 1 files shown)


TheBB:
You should probably use `boundp`. `fboundp` checks the function binding and `boundp` checks the variable binding. The latter is relevant here since we are using it as a variable.

I think a clearer way to write this is:

`(and (boundp 'fringe-mode) (not (equal fringe-mode 0)))`

TheBB:
Sorry about this but I think https://github.com/syl20bnr/spacemacs/pull/7571 is a more comprehensive solution. Since we already have `:if` for toggles we can use it, we just need to make sure that the toggle status function honors it.

jonaustin:
No worries, thanks.

Issue #7374, opened 2016-10-12
Dlang layer should list the tools it depends on


Dietr1ch:
Currently adding the `d` layer, may cause emacs to misbehave on D files.

I don't know the exact errors, but if I tell a friend to try the d layer, it should be straightforward to get it right after reading the layer's README.
The problem should be some unstated dependencies, as the layer works great for me.

This should apply for every layer, but currently the d layer has the issue.

TheBB:
PRs welcome.

Dietr1ch:
I know (=, I just have to further investigate what went wrong, but I won't get my hands on the machine until next week

Dietr1ch:
How should the requirements be listed?

Further thinking through this, It will be better to grab an empty machine, install barely emacs, and then checking what goes wrong. I'm willing to do that for arch, but testing every distro out there seems to be an overkill.

Some projects build per-distro requirements list. Maybe we should start having them on the end of each layer, what do you think of this? which distros should be listed, if any?

Dietr1ch:
I'm not sure if the installation instructions are now redundant. Installing dependencies is as easy as `pacman -S dcd` (or similar), and installing from source should imply that users know about `$PATH`.

Miciah:
Since #7576 / https://github.com/syl20bnr/spacemacs/commit/e4e6638d4df91e315e95cc7860c448ce9bcfca53 has been merged, can this issue be labeled "Fixed in develop"?

PR #7576, opened 2016-10-27
Add requirements list for D

Dietr1ch:
This adds an itemize (:P) with D requirements at the end (#7374).

Probably the installation instructions can be shrinked to the first paragraph, as having binaries on your PATH should be obvious.


TheBB:
@

PR #7579, opened 2016-10-28
spacemacs-purpose readme - Fix links and typos

CarlQLange:
I think there was a failed replace here :) The links now work correctly and the typos are gone. Yay!


d12frosted:

Issue #7585, opened 2016-10-29
lua-mode gtags layer shortcut is disable


wolfand11:
![bug1](https://cloud.githubusercontent.com/assets/1425676/19828173/dffd3eda-9df1-11e6-98b4-19040f56d58d.jpg)
![bug2](https://cloud.githubusercontent.com/assets/1425676/19828175/e2875564-9df1-11e6-8001-45c8e40f6450.jpg)

emacs-lisp-mode gtags layer shortcut is enable, but lua-mode gtags layer shortcut is disable,

TheBB:
https://github.com/syl20bnr/spacemacs/pull/7586

Miciah:
Since #7586 / https://github.com/syl20bnr/spacemacs/commit/26615a64cbe114cbae4df6ce0d55fed278b9ad90 has been merged, can this issue be labeled as "Fixed in develop"?

PR #7586, opened 2016-10-29
gtags: don’t check if mode function is bound

TheBB:
I'm not sure why this was there in the first place, but there's no guarantee that the mode function is bound when this function runs.

Fixes https://github.com/syl20bnr/spacemacs/issues/7585

### DIFF
--- a/layers/+tags/gtags/funcs.el
+++ b/layers/+tags/gtags/funcs.el
@@ -26,34 +26,34 @@ Otherwise does nothing."
 
 (defun spacemacs/helm-gtags-define-keys-for-mode (mode)
   "Define key bindings for the specific MODE."
-  (when (fboundp mode)
-    ;; The functionality of `helm-gtags-mode' is pretty much entirely superseded
-    ;; by `ggtags-mode', so we don't add this hook
-    ;; (let ((hook (intern (format "%S-hook" mode))))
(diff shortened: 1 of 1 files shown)


layers/+tags/gtags/funcs.el
```diff
@@ -26,34 +26,34 @@ Otherwise does nothing."
...
-      "gG" 'helm-gtags-dwim-other-window
-      "gi" 'helm-gtags-tags-in-this-function
-      "gl" 'helm-gtags-parse-file
-      "gn" 'helm-gtags-next-history
-      "gp" 'helm-gtags-previous-history
-      "gr" 'helm-gtags-find-rtag
-      "gR" 'helm-gtags-resume
-      "gs" 'helm-gtags-select
-      "gS" 'helm-gtags-show-stack
-      "gu" 'helm-gtags-update-tags)))
```
LemmingAvalanche: Here on the last hunk the only functional change is the deletion of the right
paren, but you’ve also de-indented the whole expression by two spaces. Was that 
intentional?

TheBB: Yes, the additional spaces was due to the `(when ...)` block which is now removed.

TheBB:

PR #7559, opened 2016-10-26
Add layer for Semantic Web formats

atextor:
- Syntax highlighting for N3, Turtle and SPARQL formats
- Company mode for SPARQL
- Execution of SPARQL queries against SPARQL HTTP endpoint

PR #7587, opened 2016-10-29
Added a layer for linked data files like ttl and n3

CarlQLange:
Based almost 100% on https://bitbucket.org/nxg/ttl-mode, but as a spacemacs layer.

For the (I expect) incredibly tiny intersection of sets of people who use spacemacs and use linked data!

It really doesn't do a lot beyond syntax highlighting, indentation and clicky links, but that's a huge improvement over plaintext.

### DIFF
--- a/layers/+lang/ttl/README.org
+++ b/layers/+lang/ttl/README.org
@@ -1,0 +1,25 @@
+#+TITLE: ttl layer
+
+# TOC links should be GitHub style anchors.
+* Table of Contents                                        :TOC_4_gh:noexport:
+ - [[#description][Description]]
+ - [[#install][Install]]
+ - [[#key-bindings][Key bindings]]
+
+* Description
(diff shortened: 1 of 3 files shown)


nixmaniack:
[`ttl-mode`](https://melpa.org/#/ttl-mode) is available on melpa. Since this layer doesn't do much configuration I find it not very useful as a layer. Using this package with `dotspacemacs-additional-packages` would make more sense.

atextor:
Hi, I'm in that tiny intersection :)
ttl-mode is part of syl20bnr/spacemacs#7559, which also contains sparql-mode. The PR is still open but I hope not for long.

CarlQLange:
Oh, sweet! In that case I'm happy to close this. My only request is that you enable goto-address-mode in those modes, so you can click all the URIs :D I suppose a post-init hook should do it fine.

Nice work!

PR #7590, opened 2016-10-29
Fix some minor mistakes and typos in the documentation

primeos:
Some minor mistakes and typos I found while reading through the documentation (I recently decided to switch from Vim to Spacemacs on my laptop and it's awesome so far! Good job guys :smile:).

PS: I didn't document the exact changes in my commit message/PR since looking at the diff/changes should be easier anyway (imho). But I could do that ofc if requested/recommended.


d12frosted:

Issue #4778, opened 2016-01-24
Buffers micro-state vim like bindings


Dietr1ch:
I feel that mnemonics for [n]next and [p]previous should be dropped for [j]next [p]previous. Even while it goes a bit against an easy way to remember, it uses your vim muscle memory for movement keys (bonus j/k are next to each other on qwerty and dvorak).

Maybe this change should be applied only if the user is using evil-mode.
There are probably a few more microstates that go through lists, they should be modified too to keep consistency.

Andre0991:
`j` and `k` are vertical movements. I think it would be weird to use them to go to, for example, the next and previous layouts, as it's a kind of horizontal movement. I think `n` and `p` are fine. Perhaps `h` and `l`, for that matter.

justbur:
The convention used in some other transient states is to bind `p` and `N` to "previous" while keeping `n` for next

StreakyCobra:
I think `n` and `p` are used a lot through Emacs and Spacemacs, and not only in micro/transient-state. For me we should keep these ones, but if you want to make (and maintain after) a layer that change these bindings, feel free to contribute :smiley:

Dietr1ch:
You are right, it seems that n/p for iterating is way too deep into emacs, so changing it may break many workflows ([relevant xkcd](http://xkcd.com/1172/)).

I'll review the bépo layer and start a new one. I might be in need of an automated way to find n/p usages, theoretically most of the changes can be done automatically, only requiring review on bind collisions, but an explicit remap might be good enough.

StreakyCobra:
I don't know if bepo will be really useful for this, as usages are listed manually and it doesn't deal with transient states for now.

Also, maybe before going too far into making it a layer, ask @syl20bnr if he is ready to merge this kind of changes.

Dietr1ch:
It seems that `n/N/p` is actually better. I'll get used to it eventually.

PR #7598, opened 2016-10-30
Buffer vim microstate

Dietr1ch:
#4778
Adds j/k bindings to move around buffers on the buffer-transient-state. Old bindings remain.
This adds some consistency for vim users that prefer j/k instead of n/p or n/N.

### DIFF
--- a/layers/+distributions/spacemacs-base/keybindings.el
+++ b/layers/+distributions/spacemacs-base/keybindings.el
@@ -460,8 +460,10 @@
   :title "Buffer Selection Transient State"
   :bindings
   ("n" next-buffer "next")
+  ("j" next-buffer "next")
   ("N" previous-buffer "previous")
   ("p" previous-buffer "previous")
+  ("k" previous-buffer "previous")
   ("K" spacemacs/kill-this-buffer "kill")
   ("q" nil "quit" :exit t))
 (spacemacs/set-leader-keys "b." 'spacemacs/buffer-transient-state/body)


layers/+lang/d/README.org
```diff
@@ -30,3 +31,6 @@ instructions]]. After successfully built DCD, you need to copy the binary
 | ~SPC m g r~ | find references to all symbol at point                |
 |-------------+-------------------------------------------------------|
 
+* Requirements
+	- [[https://github.com/Hackerpilot/DCD][DCD]] (completion)
+	- [[https://github.com/dlang/dub][DUB]] (dub projects)
```
nixmaniack: These changes should be part of another PR.

nixmaniack:
IMO, this would break the spacemacs convention(`n/N/p`) which is followed across transient states and may also start conflicting if we're using `j/k` in transients states already.

rgrinberg:
I don't think j/k are consistent with vim here either. So I would rather not have this. Perhaps this could be a preference for users though?

Dietr1ch:
I agree that using `n/j` and `n/N/k` for moving seems bloated, but having the bindings will only give you more choices.
While `j/k` are not used for `:bnext` and `:bprev` on vim, many vim-inspired bindings do use `j/k` for moving through options.
Otherwise, using `j/k` by mistake will only take you out of the transient state.


If you search for `"[jk]"` on spacemacs, many bindings appear. I wouldn't say that `n/Np` is really established over `j/k`. Preference of `n/p` over `j/k` seems to vary depending on which one caused less conflicts with the other bindings.


If you are really sure that `n/Np` is closer to be default on spacemacs I'll close this and eventually get used to it, but I kept trying to use `j/k` on this transient state and thought that I was not the only one.

nixmaniack:
> but having the bindings will only give you more choices.

IMHO, having set of conventions will be more beneficial than having more choices as number of keybindings are going to increase. 

[Here's the link](https://github.com/syl20bnr/spacemacs/blob/aa3a799c39afab721ab47a1d24294d6a753c9e54/doc/CONVENTIONS.org#n-and-n) to spacemacs' convention which mentions about usage of `n/N`.

May be this could be considered as a proposal to re-think about that convention then and move to using `j/k` in TS? I'm weighing towards `n/N/p` as it indicates `next/previous` (may be I'm biased as I'm used to it!) more clearly than `j/k` (which I think of as `up/down`).

Dietr1ch:
Yes, it's true that `j/k` suggests vertical movement, and that `n/N/p` does not, while also being a little vim-ish.  It seems that I'l have to get used to it to favor consistency, eventually it'll be easier.
Thanks for your time (=

PR #7603, opened 2016-10-30
Home buffer: improved framed notes

deb0ch:
Hey hey, I worked a little bit on the home buffer framed notes (like release notes and quickhelp):
- adjust frame width to content width
- define minimum and maximum width
- allow for inserting a caption at the bottom of the frame
- use it to add a note inviting the user to update his packages and dotfile on every release (fixes #7357)
- framed notes are centered (if `dotspacemacs-startup-buffer-responsive`)
- max width is adapted to the window width (if `dotspacemacs-startup-buffer-responsive`)
- **UPDATE** nice rendering of numbered and dotted lists (as the one in the latest release got squished by `fill-region`)

The way the toggling was managed also messed up with the responsiveness, so while I was at it I refactored it a bit to make it work nicely.

In the process I accidentally discovered the syntax checking in elisp and went a little haywire on the warnings and docstrings cleanup, hope it's not too much.

Before:
screenshot from 2016-11-05 01-15-15

After:
screenshot from 2016-11-05 01-14-42

PR #7604, opened 2016-10-31
Home buffer: center home buffer on fixed-width 80 columns when not responsive

deb0ch:
This PR is one additional commit to #7603.

Set the internal variable used to get the home buffer's window's current width for centering to a fixed width of 80 when not responsive.

So basically center the home buffer on a fixed width of 80 when `dotsmacemacs-startup-buffer-responsive` is `nil`.

This not only simplifies the way the responsiveness is handled, but also makes the home buffer look nice on a vt100 :smile:

screenshot from 2016-11-01 20-26-49

### DIFF
--- a/core/core-spacemacs-buffer.el
+++ b/core/core-spacemacs-buffer.el
@@ -25,8 +25,9 @@
 (defconst spacemacs-buffer-buttons-startup-lists-offset 25
   "Relative position between the home buffer buttons and startup lists.")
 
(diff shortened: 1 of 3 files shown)


deb0ch:
Travis fixed !

I forgot to mention it, but this is what it looks like when not centered or responsive:

screenshot from 2016-11-01 20-26-49

Let me know if you like :smile:

syl20bnr:
Thank you again :-)

PR #7608, opened 2016-10-31
Expanded theming README file with an example

smimish:
Expanded readme file, based on the forum questions and my own experience. Hope it will save someones time.


TheBB:

PR #7609, opened 2016-10-31
Fix Layout Transient Menu's docstring

### LINKED ISSUE #7597: Updating helm autocomplete menu for Layouts
I think helm autocomplete SPC l for the "L" should go into the "Actions" section and be renamed to "load layouts from file". This will make it consistent with "s/S" actions, which save it to the file.

This will fix 2 issues with it:
- The current wording is not very clear, it takes time to find it.
- The location should be under the Actions, where you would naturally look for it.

pawandubey:
Fixes #7597

Moved the `L` option to `Actions` and changed the doc string for it.

@theBB This works but the output is a bit wonky. Like below. I perhaps do not understand the formatting completely.

Wonk

### DIFF
--- a/layers/+spacemacs/spacemacs-layouts/packages.el
+++ b/layers/+spacemacs/spacemacs-layouts/packages.el
@@ -138,13 +138,14 @@
  [_C-0_,_C-9_]^^ nth/new layout               [_A_]^^   add all from layout
  [_<tab>_]^^^^   last layout                  [_d_]^^   close current layout
  [_b_]^^^^       buffer in layout             [_D_]^^   close other layout
- [_h_]^^^^       default layout               [_r_]^^   remove current buffer
- [_l_]^^^^       layout w/helm/ivy            [_R_]^^   rename current layout
- [_L_]^^^^       layouts in file              [_s_/_S_] save all layouts/save by names
- [_n_/_C-l_]^^   next layout                  [_t_]^^   show a buffer without adding it to current layout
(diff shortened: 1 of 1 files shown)


TheBB:
Yeah, you need to add some `^^` on the new lines to account for color tags that aren't there. If you look closely you see that the number of `^` and `_` per line is the same. `^` is used as a “do nothing” alignment tag so that the string can be aligned both in code and when displayed.

pawandubey:
@TheBB I already have `^^` on the last two lines, and I also don't see how the number of `_` and `^` are the same on every line.

TheBB:
I didn't say you didn't have any, I said you need to **add some**.

> I also don't see how the number of _ and ^ are the same on every line.

Well, just count. In the previous version every left column has six `_` or `^`. Whenever there is one keybinding (two `_`) there are four `^`. If there's two keybindings (four `_`) there are only two `^`. Since neither `_` nor `^` show up in the final rendered form, this will keep the table aligned (exactly six characters are omitted from every line).

Just substitute six spaces with `^` on the left column in the last two lines, you'll see.

pawandubey:
@TheBB Thanks for being patient with me :smile:  I have made the changes now and it works fine for me.

pawandubey:
@TheBB can this be merged?

TheBB:
Thanks for being patient with me too. :-P Cherry-picked in develop. You can safely delete your branch.

Issue #7611, opened 2016-11-01
automatic layer installation will corrupt layer declarations if the first one has variables


emmanueltouzery:
#### Description :octocat:
automatic layer installation will corrupt layer declarations if the first one has variables

#### Reproduction guide :beetle:
- in the configuration file have layers like that =>

```lisp
    dotspacemacs-configuration-layers
    '((auto-completion :variables auto-completion-enable-help-tooltip t)
       syntax-checking haskell javascript emmanuel git html
       markdown emacs-lisp shell version-control
       windows-scripts java typescript org racket)
```

- open a yaml file (no layer for it installed)
- accept the prompt to install the yaml layer

*Observed behaviour:* :eyes: :broken_heart:

now the configuration file looks like that =>

```lisp
   dotspacemacs-configuration-layers
   '((
      yamlauto-completion :variables auto-completion-enable-help-tooltip t)
      syntax-checking haskell javascript emmanuel git html
      markdown emacs-lisp shell version-control
      windows-scripts java typescript org racket)
```
      
in other words the auto-completion layer inclusion was botched by the new yaml
layer, presumably because of the use of variables.

*Expected behaviour:* :heart: :smile:

the configuration file should look like that =>

```lisp
   dotspacemacs-configuration-layers
   '(yaml
      (auto-completion :variables auto-completion-enable-help-tooltip t)
      syntax-checking haskell javascript emmanuel git html
      markdown emacs-lisp shell version-control
      windows-scripts java typescript org racket)
```

#### System Info :computer:
- OS: gnu/linux
- Emacs: 25.1.1
- Spacemacs: 0.200.4
- Spacemacs branch: master (rev. 6feb70d)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
((yamlauto-completion :variables auto-completion-enable-help-tooltip t)
 syntax-checking haskell javascript emmanuel git html markdown emacs-lisp shell version-control windows-scripts java typescript org racket)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND DBUS GCONF GSETTINGS NOTIFY ACL LIBSELINUX GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XWIDGETS

TheBB:
[...]

PR #7613, opened 2016-11-01
More robust dotspacemacs/add-layer

TheBB:
- Don’t greedily search until last paren, just the first one
- Better call on newlines

https://github.com/syl20bnr/spacemacs/issues/7611


emmanueltouzery:
i think this should go even in the 0.200 branch (and clearly on develop too).

TheBB:

PR #7615, opened 2016-11-01
version control: let the user pick on which side to display the diff

emmanueltouzery:
I prefer displaying the version control diff info on the left. I always overrode the spacemacs right default in my config file but it's breaking almost everytime I upgrade.

I think a PR to make also the side for the flycheck info to be configurable will follow now (I prefer that one on the right). (EDIT: for that second one, it's probably not needed. for that one you need only ` (custom-set-variables '(flycheck-indication-mode (quote right-fringe))`, and I think spacemacs doesn't overwrite it at startup, unlike the source-control setting)

### DIFF
--- a/layers/+source-control/version-control/README.org
+++ b/layers/+source-control/version-control/README.org
@@ -35,6 +35,13 @@ by setting =version-control-diff-tool=
                   version-control-diff-tool 'diff-hl)
 #+END_SRC
 
+You can choose the side on which the diff appears (by default it's the right side)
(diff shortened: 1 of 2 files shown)


bmag:

You forgot to actually define the variable in `config.el`, so I defined it in 6e4442e0dbbf20e96847b909ea8e0456b382f8c4 and renamed it to `version-control-diff-side` to better match `version-control-diff-tool`.

PR #2146, opened 2015-06-30
Fixed `electric-pair-skip` not working in racket mode

x-ji:
See https://github.com/greghendershott/racket-mode/issues/140 I feel this is the more natural behavior as a closing paren is already automatically inserted by default in Spacemacs. Therefore it makes more sense to make `electric-pair-skip` the default behavior instead of `racket-insert-closing-paren`

PR #7616, opened 2016-11-01
Restore smart closing paren behavior in racket-mode

donm:
Remove the shadowing keybindings for `)`, `]`, and `}` so the default
`racket-insert-closing` command is used again.

### DIFF
--- a/layers/+lang/racket/packages.el
+++ b/layers/+lang/racket/packages.el
@@ -111,9 +111,4 @@
         ;; Tests
         "tb" 'racket-test
         "tB" 'spacemacs/racket-test-with-coverage)
-      (define-key racket-mode-map (kbd "H-r") 'racket-run)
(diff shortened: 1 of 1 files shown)


ohspite:
Racket-mode has the command `racket-insert-closing`, which inserts whatever closing delimiter matches the opening of the current sexp.  Spacemacs uses smartparens by default, and `racket-insert-closing` does the right thing when smartparens is enabled.  That is, starting with point inside a closing parenthesis: `(call-proc|)`  pressing `]` results in no change except for point advancing: `(call-proc)|`.  

It seems like this didn't use to be the case, so the bindings for  `)`, `]`, and `}` were changed to `self-insert-command` by #2146.  The default racket-mode behavior is very useful and matches default behavior in DrRacket, so I would recommend removing the three lines added by #2146 so they are bound to their default commands in racket-mode.

TheBB:

PR #7619, opened 2016-11-01
Make org-edit-src keybindings more similar to keybindings in similar situations (for example in magit git-comit)

david-sawatzke:
This adds a keybinding for pressing `, ,` to exit the src edit buffer in org-mode.
This makes them more line with: https://github.com/syl20bnr/spacemacs/blob/master/doc/CONVENTIONS.org#confirm-and-abort , even though they don't apply

### DIFF
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -136,6 +136,7 @@
         (spacemacs/set-leader-keys-for-minor-mode 'org-src-mode
           "'" 'org-edit-src-exit
           "c" 'org-edit-src-exit
+          "," 'org-edit-src-exit
           "a" 'org-edit-src-abort
           "k" 'org-edit-src-abort))
 


LemmingAvalanche:
From your commit message:

> This makes sense so org-edit-src 

Shouldn’t that be “org-edit-src-code”?

LemmingAvalanche:
What if the source code that you are editing is Elisp? The keybinding `, ,` is
used in that layer for `lisp-state-toggle-lisp-state`.

Relevant part of `packages.el` of that layer:

``` emacs-lisp
(defun emacs-lisp/init-emacs-lisp ()
  (dolist (mode '(emacs-lisp-mode lisp-interaction-mode))
    (spacemacs/set-leader-keys-for-major-mode mode
      …
      ","  'lisp-state-toggle-lisp-state
      …
```

LemmingAvalanche:
It seems to me that your change is in line with the “`C-c C-c`”-convention:

<https://github.com/syl20bnr/spacemacs/blob/master/doc/CONVENTIONS.org#confirm-and-abort>

In particular, using `, ,` for “confirm” (and `, k` for “abort/discard”). The
document mentions `org-mode` notes as an example, and I would guess also
`org-edit-src-code` would apply. Given that, perhaps your commit message could
mention _this_ instead of the more specific reference to commit messages in
Magit?

I would like for someone more knowledgable to confirm or deny my suspicion, if
they would be so kind.

david-sawatzke:
The leader keybings for the minor mode seem to only be accessible with with the mode leader key and not with `SPC m`, so that wouldn't be a problem.
This keybinding from emacs-elisp mode is overwritten when using the mode leader key but not when using `SPC m`.
But that makes it not confirm to the keybindings mentioned in the conventions, which explicitly mention `SPC m ,`.
Anyway, the conventions aren't even applicable, as the default keybinding (in vanilla emacs) is `C-c '` and not `C-c C-c`

LemmingAvalanche:
I see. Thanks for the explanation!

bmag:

PR #7620, opened 2016-11-01
Display 'done' after a theme is loaded

Bost:
Theme (down)loading may take a minute so it is good to see some confirmation when it's over.


bmag:

PR #7622, opened 2016-11-01
add godoctor-mode for go layer

TinySong:

### DIFF
--- a/layers/+lang/go/packages.el
+++ b/layers/+lang/go/packages.el
@@ -12,6 +12,7 @@
         go-mode
         go-guru
         (go-rename :location local)
+        godoctor
         ))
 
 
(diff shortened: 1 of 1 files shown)


layers/+lang/go/packages.el
```diff
@@ -145,3 +146,12 @@
...
   (spacemacs/helm-gtags-define-keys-for-mode 'go-mode))
+
+(defun go/init-godoctor ()
+  (use-package godoctor
+    :init
+    (spacemacs/set-leader-keys-for-major-mode 'go-mode
+      "dr" 'godoctor-rename
+      "de" 'godoctor-extract
+      "dt" 'godoctor-toggle
+      "dd" 'godoctor-godoc)))
```
TheBB: Are these refactoring? If so, the refactoring binding is [`SPC m r`](https://github.com/syl20bnr/spacemacs/blob/master/doc/CONVENTIONS.org#refactoring)

TinySong: @TheBB ,great, good idea，I will fix it， thanks for the reply

Issue #4910, opened 2016-01-31
jump-to-last-layout binding doesn't work in terminal


james-w:
#### Description

I suspect this is a slightly more general problem, but I'm new, so not
sure how to debug further, or what all the terms are. I think the
description might actually be "<tab> not received by transient states in the terminal"

Basically the <tab> binding for spacemacs/jump-to-last-layout
doesn't work in a terminal.
#### Reproduction guide
- Start Emacs with -nw
- Start a second perspective layout
- Try and jump between layouts with SPC l <tab>

_Observed behaviour:_
Layout doesn't switch, main buffer receives the <tab> and
tries to do the normal binding for that (next button, autocomplete,
whatever)

_Expected behaviour:_
Jump to last layout

<tab> works elsewhere, e.g. SPC <tab> so I suspect this is
specific to transient states.
#### System Info
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.9
- Spacemacs branch: develop (rev. 42bc71d)
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:

``` elisp
(auto-completion better-defaults emacs-lisp git markdown python
                 (shell :variables shell-default-height 30 shell-default-position 'bottom)
                 spell-checking syntax-checking version-control)
```

I also tried disabling everything except (elisp shell and better-defaults) and
it still happened.

Thanks to Andre0991 on Gitter for helping me to track down how to
reproduce.

rgrinberg:
I've tried reproducing with emacs in the terminal and could not reproduce. Switching between new perspectives just fine.

Emacs 25.1/Linux/develop.

bmag:
@james-w In terminal, `TAB` and `C-i` are indistinguishable. What's the value of `dotspacemacs-distinguish-gui-tab`? Does `SPC l <tab>` work if you change its value in dotspacemacs and restart Emacs?

PR #7630, opened 2016-11-02
Fix terminal RET and TAB in layouts/workspaces transient state

donm:
* Make RET and TAB (C-m and C-i) perform the same actions as <return>
  and <tab> so the maps work in the terminal.
* Catch <return> in workspaces transient state so that it exits without
  executing a command (like already done for layouts).
* Add "workspace w/helm/ivy" to docstring, mirroring the layouts
  docstring.


ohspite:
Oh hey, look at that.  Fixes #4910.

TheBB:

Issue #7634, opened 2016-11-02
web-mode "File mode specification error"


hershal:
#### Description :octocat:
web-mode give a "File mode specification error" when visiting a web-mode file.

#### Reproduction guide :beetle:
- Start Emacs, making sure you have the `html` layer enabled
- Open a html file
- M-x revert-buffer

*Observed behaviour:* :eyes: :broken_heart:
I get a `File mode specification error: (void-function spacemacs/toggle-smartparens-off)` in the modeline.

*Expected behaviour:* :heart: :smile:
I shouldn't get this message.

#### System Info :computer:
- OS: darwin
- Emacs: 25.1.1
- Spacemacs: 0.200.4
- Spacemacs branch: develop-local (rev. db20ba5)
- Graphic display: t
- Distribution: spacemacs-base
- Editing style: emacs
- Completion: helm
- Layers:
```elisp
(helm
 (auto-completion :variables auto-completion-enable-sort-by-usage t auto-completion-enable-snippets-in-popup t)
 better-defaults restclient search-engine
 (shell :variables shell-enable-smart-eshell t)
 (spell-checking :variables spell-checking-enable-by-default nil)
 syntax-checking gtags
 (c-c   :variables c-c  -enable-clang-support t c-c  -default-mode-for-headers 'c  -mode)
 emacs-lisp markdown
 (org :variable org-enable-bootstrap-support t)
 html javascript latex swift typescript ruby yaml python git perforce version-control hb-smartparens hb-cmake hb-lisp hb-org hb-multiple-cursors hb-yasnippet hb-javascript hb-expand-region hb-web)
```
- System configuration features: JPEG RSVG NOTIFY ACL LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS

#### Notes
What happened to the `spacemacs/toggle-smartparens-off` function?

#### Backtrace :paw_prints:
Not relevant

PR #7635, opened 2016-11-02
layers/html/web-mode: replace smartparens toggle

hershal:
This resolves the `File mode specification error` when visiting a
web-mode file.

https://github.com/syl20bnr/spacemacs/issues/7634

### DIFF
--- a/layers/+lang/html/packages.el
+++ b/layers/+lang/html/packages.el
@@ -155,7 +155,7 @@
      'smartparens-mode)
    '(css-mode-hook scss-mode-hook sass-mode-hook less-css-mode-hook))
 
-  (add-hook 'web-mode-hook 'spacemacs/toggle-smartparens-off))
+  (add-hook 'web-mode-hook 'turn-off-smartparens-mode))
 
 (defun html/init-tagedit ()
   (use-package tagedit


layers/+lang/html/packages.el
```diff
@@ -155,7 +155,7 @@
      'smartparens-mode)
    '(css-mode-hook scss-mode-hook sass-mode-hook less-css-mode-hook))
 
-  (add-hook 'web-mode-hook 'spacemacs/toggle-smartparens-off))
+  (add-hook 'web-mode-hook 'turn-off-smartparens-mode))
```
d12frosted: I wonder why it's not in `:init` of `use-package web-mode`.

d12frosted: I mean in the first place 💃

hershal: Should it be? If so, I can move it there. How should layers handle cross-package dependencies like this? I've asked myself this in my personal configurations, where I would like to enable or disable some minor mode in a particular major mode. Generally, should the `add-hook` be in the major mode `post-init-<major-mode>` function, or in the minor mode's `post-init-<minor-mode>`?

bmag:
From the issue report I see that you use `spacemacs-base`, while Smartparens belongs to `spacemacs-editing`. This post-init runs only if you have Smarparens configured and assumes you got Smartparens from `spacemacs-editing` (where the toggle is defined), but it looks like you install Smartparens from a private layer (that doesn't define the toggle). I'm fine with this PR, but be aware there may be other places with similar assumptions as well.

layers/+lang/html/packages.el
bmag: It depends how the packages relate to each other. In this case, I think the current situation is fine. @d12frosted what do you think?

hershal:
@bmag Nice, I didn't know that layer existed! The call to `spacemacs|add-toggle` in `spacemacs-editing/init-smartparens` explains why I couldn't find the `spacemacs/toggle-smartparens-off` defun anywhere. I think I'd rather use the `spacemacs|add-toggle`'d version, so I'm closing this PR.

PR #7637, opened 2016-11-02
Add ensime

brakhane:
add /ensime to .gitignore. It's created by -surprise- ENSIME (scala layer) and prevents updates from within spacemacs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,6 +44,7 @@ python-*-docs-html
 /session.*
 /srecode-map.el
 /recentf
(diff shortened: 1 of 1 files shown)


d12frosted:
Is it some kind of cache? If it is, then it would be nicer to configure ensime to store it's cache under `(concat spacemacs-cache-directory "ensime/")`.

brakhane:
Yes, you are right, it's better to fix this than working around it.

PR #7642, opened 2016-11-03
Update README.org

FrancescElies:
Uses flycheck instead of flycheck-typescript-tslint


d12frosted:
Ok, since `flycheck-typescript-tslint` is now part of the `flycheck` itself. Thanks for updating it!

d12frosted:

PR #7648, opened 2016-11-03
Update ensime variable name

brakhane:
This is a follow up to PR #7637. Instead of working around the problem, I just fixed it :)

`user-emacs-ensime-directory` was renamed to `ensime-startup-dirname` on Jun 25th. Without the change, ENSIME would create it's cache into spacemacs/ensime, which is not what we want.

The disadvantage is that people will get the "looks like you run ENSIME for the first time" message again, but it can't be helped.

### DIFF
--- a/layers/+lang/scala/packages.el
+++ b/layers/+lang/scala/packages.el
@@ -38,7 +38,7 @@
       (add-to-list 'spacemacs-jump-handlers-scala-mode 'ensime-edit-definition))
     :config
     (progn
-      (setq user-emacs-ensime-directory ".cache/ensime")
+      (setq ensime-startup-dirname (expand-file-name "ensime" spacemacs-cache-directory))
 
       (evil-define-key 'insert ensime-mode-map
         (kbd ".") 'scala/completing-dot


layers/+lang/scala/packages.el
```diff
@@ -38,7 +38,7 @@
       (add-to-list 'spacemacs-jump-handlers-scala-mode 'ensime-edit-definition))
     :config
     (progn
-      (setq user-emacs-ensime-directory ".cache/ensime")
+      (setq ensime-startup-dirname (concat spacemacs-cache-directory "ensime"))
```
bmag: It's more robust to use `expand-file-name` instead of `concat`.

bmag:
> The disadvantage is that people will get the "looks like you run ENSIME for the first time" message again, but it can't be helped.

As someone who doesn't use Ensime, What are the implications of this for Ensime-users? What will the user need to do in this case?

brakhane:
> What will the user need to do in this case?

Nothing, the message will only appear once (it appears only when the directory did not exist at the start) and ENSIME might rebuild its server (not sure).

Users might just be confused that they see this message again, but I doubt it will result in more than a "huh? That's weird" reaction in the worst case :)

bmag:
Oh, well. I just hope it doesn't take a long time to rebuild the server.

brakhane:
It took 10s on my machine, so negligible.

layers/+lang/scala/packages.el
brakhane: Thanks, will change it.

d12frosted:

PR #7649, opened 2016-11-03
Fix bug in org layer

spradnyesh:
org layer documentation mentions inclusion of org-bullets package
but it was missing in the actual layer

### DIFF
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -36,6 +36,7 @@
             :toggle org-enable-github-support)
     (ox-reveal :toggle org-enable-reveal-js-support)
     persp-mode
+    org-bullets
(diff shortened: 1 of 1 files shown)


bmag:
`org-bullets` is provided by `spacemacs-org` layer. You can find this out by running `SPC h SPC` and searching for "org-bullets". IMO the correct way to fix this is to add a `layers.el` file that calls `configuration-layer/declare-layer` to enable the `spacemacs-org` layer. See `layers/+distributions/spacemacs/layers.el` for an example.

bmag:
Also, please link the issue where you reported the bug (you did open an issue, right?).

TheBB:
I went with @bmag's solution. Thank you, though.

PR #7650, opened 2016-11-04
Reinstate avy jump commands

scot0357:
Avy jump command are in the news01.org and the documentation
but the actual keybindings are not defined. This change defines them.


bmag:
Avy is part of `spacemacs-editing` layer, not `spacemacs-base`. These bindings are already defined in `spacemacs-editing/init-avy`.

PR #7654, opened 2016-11-04
Add ibuffer bindings for updating and filter-group motion

ohspite:
The `ibuffer-update` command is originally on `g` but is evilified to
the inconvenient `C-S-g`, so bind it to `r` (for "refresh").

### DIFF
--- a/layers/+emacs/ibuffer/packages.el
+++ b/layers/+emacs/ibuffer/packages.el
@@ -32,7 +32,11 @@
       (evil-ex-define-cmd "buffers" 'ibuffer))
     :config
     (evilified-state-evilify-map ibuffer-mode-map
-      :mode ibuffer-mode)))
+      :mode ibuffer-mode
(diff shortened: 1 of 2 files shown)


layers/+emacs/ibuffer/packages.el
```diff
@@ -32,7 +32,11 @@
       (evil-ex-define-cmd "buffers" 'ibuffer))
     :config
     (evilified-state-evilify-map ibuffer-mode-map
-      :mode ibuffer-mode)))
+      :mode ibuffer-mode
+      :bindings
+      "r" 'ibuffer-update
```
bmag: We usually bind `gr` for refresh in evilified buffers, so can you bind `gr` please?

ohspite: Done.  Thank you.

bmag:

Issue #7658, opened 2016-11-04
Selecting holy mode upon installation will leave emacs in evil mode until restart


quietnan:
#### Reproduction guide :beetle:
- Remove .spacemacs and clear all caches (I went as far as spinning up a fresh AWS instance)
- Start Emacs
- When prompted, choose holy mode over evil
- Change buffer to some file or *scratch*
- Start typing to realize you are in evil mode
- Restart emacs
- Change buffer to some file or *scratch*
- Start typing to realize that you are now in holy mode

*Observed behaviour:* :eyes: :broken_heart:
- See above. Evil mode after installation before restart

*Expected behaviour:* :heart: :smile:
- Never seeing evil mode if I choose holy upon installation

#### System Info :computer:
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.200.5
- Spacemacs branch: master (rev. 664ba6a)
- Graphic display: t
- Distribution: spacemacs
- Editing style: emacs
- Completion: ivy
- Layers:
```elisp
(ivy auto-completion emacs-lisp git
     (go :variables gofmt-command "goimports" go-use-gometalinter t)
     (latex :variables latex-enable-visual-line t latex-enable-auto-fill nil latex-enable-folding t)
     markdown python spell-checking syntax-checking)
```

bmag:
The README [says](https://github.com/syl20bnr/spacemacs#default-installation) to restart Emacs after the initial installation of packages. Not sure if it's a bug or just a quirk of the installation process.

TheBB:
You should probably do that anyway, however I was able to tweak the loading process for this to work: https://github.com/syl20bnr/spacemacs/pull/7659

Downside is you don't get the nice home buffer showing until after you have answered the initial questions.

LemmingAvalanche:
It seems that this is a duplicate
of <https://github.com/syl20bnr/spacemacs/issues/7596>.

PR #7659, opened 2016-11-04
Install dotfile earlier

TheBB:
Fixes https://github.com/syl20bnr/spacemacs/issues/7658

This makes the install wizard look worse, but gives you the editing style you want right out of the box.

I'm not sure which I prefer.


quietnan:
In my case having the wrong edit style after the installation lead me to (wrongly) conclude that the installation was broken and I spent the working day with plain emacs instead of spacemacs before I had time to look into it. So I would have preferred a slightly worse looking wizard.
How about just enforcing a restart after the installation is finished?

LemmingAvalanche:
I just deleted my `.emacs.d` and tried starting Emacs on your branch with no
`.spacemacs`. I think that completely-vanilla look is fine. It doesn’t provide
the same intrigue as the previous behaviour, but giving up some form in favour
of more predictable function might be worth it.

The buggy behaviour is a paper cut, in the sense that “switching it on and off
again” solves it, but it’s a bit subtle and non-obvious (even though it’s
documented), and it can lead to a lot of wasted time spent on “debugging”
it (which has happened at least on two occasions by now).

syl20bnr:
It was way to ugly to my taste ;-)
I hacked a solution in commit https://github.com/syl20bnr/spacemacs/commit/2c4cd4d
Actually it is not really a hack, we must reapply the preferences selected in the wizard at the end. Since the dotfile is just installed we know that `dotspacemacs/init` is side effect free as it does not contain any user code.

Thank you !

Issue #7646, opened 2016-11-03
Feature Request - Allow bookmarks to be created / edited using Ivy


gilbertw1:
The current [Spacemacs documentation](https://github.com/syl20bnr/spacemacs/blob/master/doc/DOCUMENTATION.org#bookmarks) regarding bookmarks is specific to Helm and is inaccurate when using Ivy. Using ```SPC f b``` to access ```bookmark-jump``` when using Ivy displays a completion buffer that only allows you to select an existing bookmark. While doing the same using helm allows you to create a new if you enter a new bookmark name, and provides several bookmark management abilities (edit, delete, etc.).

It would be nice to bring Ivy to feature parity with Helm in regards to managing bookmarks.

bmag:
Does the Ivy backend support setting a bookmark? With helm layer, `SPC f b` is bound to `helm-filtered-bookmarks` which allows both jumping and setting a bookmark. With ivy layer, `SPC f b` is bound to `bookmark-jump`, which doesn't allow setting a bookmark. We can, and probably should, bind `SPC f b` to `counsel-bookmark`, but that doesn't allow setting a bookmark either. Do you know if Ivy/Counsel has support for this?

gilbertw1:
Ah, looks like it doesn't.....I'll move this feature request there!

PR #7660, opened 2016-11-04
rebind 'SPC f b' to 'counsel-bookmark' when using ivy

gilbertw1:
Currently ```SPC f b``` is bound to ```bookmark-jump```, however a more appropriate choice would be ```counsel-bookmark```.

Currently both functions behave identically, but if abo-abo/swiper#758 is accepted then ```counsel-bookmark``` will be much closer to feature parity with the helm implementation (originally requested here: #7646).


bmag:

Issue #4688, opened 2016-01-19
org bullets mess up when switching themes


mankoell:
When I switch themes with `SPC T n` while beeing in an org file, some weird bullets appear in front of. The only way to get rid of them is to restart spacemacs. I dont know exactly if it has to do with `org-bullets`.

![bildschirmfoto vom 2016-01-19 17 08 52](https://cloud.githubusercontent.com/assets/1057526/12424249/8d7498d6-becf-11e5-9d93-0a7b3fbbd5b4.png)
![bildschirmfoto vom 2016-01-19 17 08 02](https://cloud.githubusercontent.com/assets/1057526/12424255/981229b6-becf-11e5-93ef-7a45047c3172.png)
![bildschirmfoto vom 2016-01-19 17 08 17](https://cloud.githubusercontent.com/assets/1057526/12424273/b048d886-becf-11e5-948c-5484ac422a38.png)

TheBB:
This is a limitation of `org-bullets-mode` that I'm not sure we can fix. It sets the foreground of hidden bullets according to the background of the current theme, but doesn't update them afterwards. If you re-enable org-mode after changing theme, it'll be fixed. I'm leaning towards saying this is an org-mode problem and you should report it there, and not here.

mankoell:
I'm sorry, I didn't know. I guess you mean it is a problem of `org-bullet`, not org-mode itself?

TheBB:
Sorry, I think I'm confused. :-D

I believe this is `org-indent-mode`, not `org-bullets-mode`. This mode is part of the org package proper. It has a lot of problems, this is only one of them.

TheBB:
Actually I'm not even sure if that's the case any more.

mankoell:
So how do I disable `org-bullets` then? I tried:
```
  (setq org-excluded-packages '(org-bullets))
```

pmagwene:
@mankoell 
You can disable org-bullets by adding it to this line in `.spacemacs`:

```
   dotspacemacs-excluded-packages '()
```

mankoell:
@pmagwene Thanks, this works great!

joonro:
I don't have a solution for this, but you can refresh the local setup to make it look right after a theme change instead of restarting emacs. For example, you can go to the `#+TITLE` line and then do `C-c C-c` and the annoying stars should be gone.

kconner:
[...]

PR #7667, opened 2016-11-06
Make org-mode restart on theme change.

mwillsey:
Solves problem where hidden org bullets would be the old background color.

### DIFF
--- a/layers/+spacemacs/spacemacs-org/packages.el
+++ b/layers/+spacemacs/spacemacs-org/packages.el
@@ -33,6 +33,11 @@
 ;; dummy init function to force installation of `org-plus-contrib'
 (defun spacemacs-org/init-org-plus-contrib ())
 
+(defun spacemacs-org/org-restart ()
+    "Restart org-mode, preserving the current visibility of headers"
+   (when (featurep 'org)
+     (org-save-outline-visibility 'use-markers (org-mode-restart))))
+
 (defun spacemacs-org/init-default-org-config ()
   (use-package org
     :commands (org-clock-out org-occur-in-agenda-files org-agenda-files)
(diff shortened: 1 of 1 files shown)


LemmingAvalanche:
> Solves problem where hidden org bullets would be the old background color.

Was this reported somewhere?

mwillsey:
#4688 mentions it, but it was closed for another reason. The problem is still present.

LemmingAvalanche:
If that issue is relevant to your commit, maybe you could reference it in your commit message?

mwillsey:
Okay, I changed it.

layers/+spacemacs/spacemacs-org/packages.el
```diff
@@ -59,6 +59,11 @@
                     (1 font-lock-comment-face prepend)
                     (2 font-lock-function-name-face)
                     (3 font-lock-comment-face prepend))))
+      ;; changing themes can leave artifacts in org mode, it should reload
+      (add-hook 'spacemacs-post-theme-change-hook
+                '(lambda ()
+                   (when (featurep 'org)
+                     (org-save-outline-visibility 'use-markers (org-mode-restart)))))
```
nixmaniack: Would be better to extract that to a function and then add it to hook. Almost all the functions added to hooks in spacemacs are named functions (maybe "Unwritten Rule") which helps if someone wants to remove it from the hook.

nixmaniack:
Why did you close? I think this is a nice fix for users who change themes a lot at runtime and use org-mode.

mwillsey:
Sorry! I was doing some housekeeping and didn't realize this wasn't
merged. Feel free to take the changed and submit your own PR.

 - Max

PR #7669, opened 2016-11-06
Python: Highlight wdb breakpoints as well as pdb/ipdb/pudb

koterpillar:
`wdb` is another Python debugger, its breakpoints should be highlighted as well as `pdb`, etc.

Listed debugger names in full instead of a clever `i?pu?db` because this helps greppability (i.e. you won't miss this occurrence when searching the repository for `pdb`).


TheBB:

PR #7670, opened 2016-11-06
Python: search for Pylint and Flake8 in the virtualenvs

koterpillar:
When a Python virtual environment is activated, search for the Flycheck checkers - Pylint and Flake8 - in that virtual environment.

### DIFF
--- a/layers/+lang/python/funcs.el
+++ b/layers/+lang/python/funcs.el
@@ -35,6 +35,15 @@
       (setq python-shell-interpreter-args "-i")
       (setq python-shell-interpreter "python"))))
 
+(defun spacemacs/python-setup-checkers (&rest args)
+  (let ((pylint (spacemacs/pyenv-executable-find "pylint"))
+        (flake8 (spacemacs/pyenv-executable-find "flake8")))
+    (when pylint
+      (flycheck-set-checker-executable "python-pylint" pylint))
+    (when flake8
+      (flycheck-set-checker-executable "python-flake8" flake8))
+    ))
+
 (defun spacemacs/python-toggle-breakpoint ()
   "Add a break point, highlight it."
   (interactive)
--- a/layers/+lang/python/packages.el
+++ b/layers/+lang/python/packages.el
@@ -214,7 +214,8 @@
           "Vw" 'pyvenv-workon))
       ;; setup shell correctly on environment switch
       (dolist (func '(pyvenv-activate pyvenv-deactivate pyvenv-workon))
-        (advice-add func :after 'spacemacs/python-setup-shell)))))
+        (advice-add func :after 'spacemacs/python-setup-shell)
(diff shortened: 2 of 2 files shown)


layers/+lang/python/funcs.el
```diff
@@ -35,6 +35,16 @@
       (setq python-shell-interpreter-args "-i")
       (setq python-shell-interpreter "python"))))
 
+(defun spacemacs/python-setup-checkers (&rest args)
+  (progn
```
bmag: `progn` inside of `defun` is redundant, please remove it.

layers/+lang/python/funcs.el
```diff
@@ -35,6 +35,16 @@
...
+    (if (spacemacs/pyenv-executable-find "pylint")
+        (flycheck-set-checker-executable "python-pylint"
+                                         (spacemacs/pyenv-executable-find "pylint")))
+    (if (spacemacs/pyenv-executable-find "flake8")
+        (flycheck-set-checker-executable "python-flake8"
+                                         (spacemacs/pyenv-executable-find "flake8")))
+    ))
```
bmag: 1. Please use a loop (`dolist` will probably suffice) to avoid code duplication.
2. Use `when` instead of `if` when you don't have an `else` clause.
3. (optional) use `let` to bind the result of `spacemacs/pyenv-executable-find` instead of calling it twice. What I mean:
```elisp
(let ((path (spacemacs/pyenv- ...)))
  (when path
    (flycheck-set- ... path)))
```

koterpillar:
Thanks for the comments, updated the definition.

koterpillar:
I did actually forget, but I don't think it makes sense for just 2 checkers either.

robsonpeixoto:
@koterpillar is possible to use the `isort` from virtualenv too?

koterpillar:
@robsonpeixoto I've checked https://github.com/immerrr/py-isort.el/blob/master/py-isort.el, they don't have it configurable. Given that `isort` isn't very configurable or dependent on the environment, does it make sense to do it?

robsonpeixoto:
@koterpillar if `py-isort.el` is not configurable, do not make sense make it configurable now. Thanks @koterpillar

cpaulik:
Just to make sure. Is this supposed to work only with `pyenv` environments or also with regular virtualenvs? The title confuses me at the moment.

koterpillar:
Both are supported.

cpaulik:
Are you sure. As far as I know `spacemacs/pyenv-executable-find` only works with `pyenv`. If `pyenv` is not installed it will also find e.g. the system version of `flake8`

koterpillar:
Yes, since virtualenv sets `$PATH`, and `spacemacs/pyenv-executable-find` defaults to searching that. I have tested it with a regular virtualenv setup via `virtualenvwrapper`.

TheBB:

PR #7672, opened 2016-11-06
finer configuration of line numbers activation

deb0ch:
Allow for controlling more precisely when line numbers are activated.

This change is 100% backward compatible: the old `nil`, `t` and `'relative` values work just like usual.

As an alternative to these values, the user can now choose to set `dotspacemacs-line-numbers` to a property list like:

``` elisp
(setq-default dotspacemacs-lines-numbers '(:relative nil
                                           :disabled-for dired-mode
                                                         doc-view-mode
                                                         markdown-mode
                                                         org-mode
                                                         pdf-view-mode
                                                         text-mode
                                           :size-limit-kb 1000))
```
This allows for precisely filtering based on major and file size.

Possible improvements would be:

- choose between `linum` and `nlinum` from there (remove `nlinum` layer, merge it into `spacemacs-base`)
- disable/enable in all derived modes
- add a default list of major modes where you never want line numbers **(?)**

**PS:** also made nlinum consistent with linum by reflecting ddbee36: now activating nlinum-relative also activates nlinum.

### DIFF
--- a/core/templates/.spacemacs.template
+++ b/core/templates/.spacemacs.template
@@ -252,8 +252,19 @@ values."
    ;; scrolling overrides the default behavior of Emacs which recenters point
    ;; when it reaches the top or bottom of the screen. (default t)
    dotspacemacs-smooth-scrolling t
-   ;; If non-nil line numbers are turned on in all `prog-mode' and `text-mode'
-   ;; derivatives. If set to `relative', also turns on relative line numbers.
+   ;; Control line numbers activation.
+   ;; If set to `t' or `relative' line numbers are turned on in all `prog-mode' and
+   ;; `text-mode' derivatives. If set to `relative', line numbers are relative.
+   ;; This variable can also be set to a property list for finer control:
+   ;; '(:relative nil
+   ;;   :disabled-for-modes dired-mode
+   ;;                       doc-view-mode
+   ;;                       markdown-mode
+   ;;                       org-mode
+   ;;                       pdf-view-mode
+   ;;                       text-mode
+   ;;   :size-limit-kb 1000)
+   ;; See Spacemacs ducumentation (SPC h SPC) for a list of available properties.
    ;; (default nil)
    dotspacemacs-line-numbers nil
    ;; Code folding method. Possible values are `evil' and `origami'.
--- a/layers/+distributions/spacemacs-base/funcs.el
+++ b/layers/+distributions/spacemacs-base/funcs.el
@@ -1014,3 +1014,25 @@ a split-side entry, its value must be usable as the SIDE argument for
                                             text-scale-mode-amount) 1)
                                   (if (car (window-margins))
                                       (car (window-margins)) 1)))))
+
+(defun spacemacs/enable-line-numbers-p ()
+  "Return non-nil if line numbers should be enabled for current buffer.
+Decision is based on `dotspacemacs-line-numbers'."
+  (and dotspacemacs-line-numbers
+       (not (string-match-p "\\*.*\\*" (buffer-name)))
+       (not (and (spacemacs/mplist-get dotspacemacs-line-numbers
+                                       :size-limit-kb)
+                 (> (buffer-size)
+                    (* 1000 (car (spacemacs/mplist-get dotspacemacs-line-numbers
+                                                       :size-limit-kb))))))
+       (or (and (spacemacs/mplist-get dotspacemacs-line-numbers
+                                      :enabled-for-modes)
+                (memq major-mode (spacemacs/mplist-get dotspacemacs-line-numbers
+                                                       :enabled-for-modes)))
+           (and (not (spacemacs/mplist-get dotspacemacs-line-numbers
+                                           :enabled-for-modes))
+                (spacemacs/mplist-get dotspacemacs-line-numbers
+                                      :disabled-for-modes)
+                (not (memq major-mode
+                           (spacemacs/mplist-get dotspacemacs-line-numbers
+                                                 :disabled-for-modes)))))))
--- a/layers/+distributions/spacemacs-base/packages.el
+++ b/layers/+distributions/spacemacs-base/packages.el
@@ -240,9 +240,20 @@
     :init (spacemacs/set-leader-keys "ji" 'imenu)))
 
 (defun spacemacs-base/init-linum ()
-  (when dotspacemacs-line-numbers
-    (add-hook 'prog-mode-hook 'linum-mode)
-    (add-hook 'text-mode-hook 'linum-mode))
+  (use-package linum
+    :config
+    (progn
+      (if (or (eq dotspacemacs-line-numbers t)
+              (eq dotspacemacs-line-numbers 'relative))
+          (progn
+            (add-hook 'prog-mode-hook 'linum-mode)
+            (add-hook 'text-mode-hook 'linum-mode)))
+      (defun linum-on ()
+        "Overwrite the original `linum-on' function with a more selective one."
+        (when (spacemacs/enable-line-numbers-p)
+          (linum-mode)))
+      (when dotspacemacs-line-numbers
+        (global-linum-mode))))
   (setq linum-format "%4d")
   (spacemacs|add-toggle line-numbers
     :mode linum-mode
--- a/layers/+misc/nlinum/funcs.el
+++ b/layers/+misc/nlinum/funcs.el
@@ -1,0 +1,14 @@
+;;; packages.el --- nlinum Layer packages File
+;;
+;; Copyright (c) 2012-2016 Sylvain Benner & Contributors
+;;
+;; Author: Thomas de Beauchêne <thomas.de.beauchene@gmail.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+(defun spacemacs/nlinum-maybe-on ()
+  (when (spacemacs/enable-line-numbers-p)
+    (nlinum-mode)))
--- a/layers/+misc/nlinum/funcs.el
+++ b/layers/+misc/nlinum/funcs.el
@@ -12,3 +12,9 @@
 (defun spacemacs/nlinum-maybe-on ()
   (when (spacemacs/enable-line-numbers-p)
     (nlinum-mode)))
+
+(defun spacemacs/nlinum-relative-toggle ()
+  (interactive)
+  (if (not (bound-and-true-p nlinum-relative-mode))
+      (nlinum-mode))
+  (nlinum-relative-toggle))
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -301,8 +301,20 @@ recenters point when it reaches the top or bottom of the
 screen.")
 
 (defvar dotspacemacs-line-numbers nil
-  "If non nil line numbers are turned on in all `prog-mode' and `text-mode'
-derivatives. If set to `relative', also turns on relative line numbers.")
+  "Control line numbers activation.
+If set to `t' or `relative' line numbers are turned on in all `prog-mode' and
+`text-mode' derivatives. If set to `relative', line numbers are relative.
+This variable can also be set to a property list for finer control:
+'(:relative nil
+  :disabled-for-modes dired-mode
+                      doc-view-mode
+                      markdown-mode
+                      org-mode
+                      pdf-view-mode
+                      text-mode
+  :size-limit-kb 1000)
+The property `:enabled-for-modes' takes priority over `:disabled-for-modes' and
+restricts line-number to the specified list of major-mode.")
 
 (defvar dotspacemacs-persistent-server nil
   "If non nil advises quit functions to keep server open when quitting.")
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -1122,6 +1122,43 @@ If it is set to =relative=, line numbers are show in a relative way:
 (setq-default dotspacemacs-line-numbers 'relative)
 #+END_SRC
 
+=dotspacemacs-line-numbers= can also be set to a property list for finer control
+over line numbers activation.
+
+Available properties:
+
+| Property              | Description                                                                                  |
+|-----------------------+----------------------------------------------------------------------------------------------|
+| =:disabled-for-modes= | list of major modes where line numbering is inhibited                                        |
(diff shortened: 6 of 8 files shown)


layers/+distributions/spacemacs-base/funcs.el
```diff
@@ -979,3 +979,17 @@ a split-side entry, its value must be usable as the SIDE argument for
...
+  (if dotspacemacs-line-numbers
+      (if (or (string-match-p "\\*.*\\*" (buffer-name))
+              (and (plist-get dotspacemacs-line-numbers :size-limit-kb)
+		   (> (buffer-size) (* 1000 (plist-get dotspacemacs-line-numbers :size-limit-kb)))))
+	  nil
+        (if (plist-get dotspacemacs-line-numbers :enabled-for)
+            (member major-mode (plist-get dotspacemacs-line-numbers :enabled-for))
+          (and (plist-get dotspacemacs-line-numbers :disabled-for)
+               (not (member major-mode (plist-get dotspacemacs-line-numbers :disabled-for))))))
+    nil))
```
bmag: This function has a somewhat hard-to-read structure:
```elisp
(if a
    (if (or b
            (and c
                 d))
        nil
      (if e
          f
        (and g
             (not h))))
  nil)
```
If I'm not mistaken, it is equivalent to this easier-to-read structure:
```elisp
(and a
     (not b)
     (not (and c d))
     (or (and e f)
         (and g (not h))))
```

layers/+distributions/spacemacs-base/funcs.el
```diff
@@ -979,3 +979,17 @@ a split-side entry, its value must be usable as the SIDE argument for
                                             text-scale-mode-amount) 1)
                                   (if (car (window-margins))
                                       (car (window-margins)) 1)))))
+
+(defun spacemacs/linum-enabled-p ()
```
bmag: Function name is not accurate. Better name: `spacemacs/enable-linum-p` (it's "enable" without a "d", not a typo)

layers/+misc/nlinum/funcs.el
```diff
@@ -0,0 +1,20 @@
...
+
+(defun spacemacs/nlinum-on ()
+  (when (spacemacs/linum-enabled-p)
+    (nlinum-mode 1)))
+
+(defun spacemacs/nlinum-relative-toggle ()
+  (interactive)
+  (if (not (bound-and-true-p nlinum-relative-mode))
+      (nlinum-mode))
+  (nlinum-relative-toggle))
```
bmag: Is `(nlinum-relative-toggle)` supposed to be inside the `if`?

layers/+misc/nlinum/funcs.el
```diff
@@ -0,0 +1,20 @@
...
+;; Author: Thomas de Beauchêne <thomas.de.beauchene@gmail.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+(defun spacemacs/nlinum-on ()
+  (when (spacemacs/linum-enabled-p)
+    (nlinum-mode 1)))
```
bmag: `(nlinum-mode)` is the same as `(nlinum-mode 1)`.

layers/+distributions/spacemacs-base/packages.el
```diff
@@ -240,9 +240,20 @@
...
+    (progn
+      (if (or (eq dotspacemacs-line-numbers t)
+              (eq dotspacemacs-line-numbers 'relative))
+          (progn
+            (add-hook 'prog-mode-hook 'linum-mode)
+            (add-hook 'text-mode-hook 'linum-mode)))
+      (defun linum-on ()
+        "Overwrite the original `linum-on' function with a more selective one."
+        (when (spacemacs/linum-enabled-p)
+          (linum-mode 1)))
```
bmag: `(linum-mode)` is the same as `(linum-mode 1)`.

layers/+distributions/spacemacs-base/packages.el
```diff
@@ -240,9 +240,20 @@
...
+      (when dotspacemacs-line-numbers
+        (global-linum-mode 1))))
```
bmag: `(global-linum-mode)` is the same as `(global-linum-mode 1)`.

layers/+distributions/spacemacs-base/funcs.el
```diff
@@ -979,3 +979,17 @@ a split-side entry, its value must be usable as the SIDE argument for
...
+  "Return non-nil if line numbers should be enabled for current buffer.
+Decision is based on `dotspacemacs-line-numbers'."
+  (if dotspacemacs-line-numbers
+      (if (or (string-match-p "\\*.*\\*" (buffer-name))
+              (and (plist-get dotspacemacs-line-numbers :size-limit-kb)
+		   (> (buffer-size) (* 1000 (plist-get dotspacemacs-line-numbers :size-limit-kb)))))
+	  nil
+        (if (plist-get dotspacemacs-line-numbers :enabled-for)
+            (member major-mode (plist-get dotspacemacs-line-numbers :enabled-for))
```
bmag: major-modes are symbols, so you can use `memq` instead of `member`

layers/+distributions/spacemacs-base/funcs.el
```diff
@@ -979,3 +979,17 @@ a split-side entry, its value must be usable as the SIDE argument for
...
+          (and (plist-get dotspacemacs-line-numbers :disabled-for)
+               (not (member major-mode (plist-get dotspacemacs-line-numbers :disabled-for))))))
```
bmag: memq

layers/+misc/nlinum/funcs.el
```diff
@@ -0,0 +1,20 @@
...
+;; Copyright (c) 2012-2016 Sylvain Benner & Contributors
+;;
+;; Author: Thomas de Beauchêne <thomas.de.beauchene@gmail.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+(defun spacemacs/nlinum-on ()
```
bmag: Misleading name, because it doesn't always enable `nlinum-mode`. Better name: `spacemacs/nlinum-maybe-on`. (there are plenty of "maybe" functions in the ecosystem, e.g. `ahs-mode-maybe` and `doc-view-mode-maybe`)

bmag:
Another two general comments:

The `:disabled-for` and `:enabled-for` keywords for specifying (dis)allowed major modes are nice, but might cause confusion with `:enabled-for` and `:disabled-for` from the layer list, because there they specify layers. I don't know if want `:enabled-for` and `:disabled-for` to refer to major modes in one context and layers in another context. It's worth consideration.

The code doesn't take into account inheritance of major modes. AFAIU there's no way to specify "enabled for prog-mode, but disabled for c-mode". In practice, I suspect it's a non-issue and not worth the trouble of supporting it. (I mean, is there any logical use-case that will require it?)

layers/+distributions/spacemacs-base/funcs.el
deb0ch: Agreed, now that I think of it it is confusing.

layers/+distributions/spacemacs-base/funcs.el
deb0ch: What is the difference ?

bmag: `memq` uses `eq` while `member` uses `equal`. See https://www.gnu.org/software/emacs/manual/html_node/elisp/Equality-Predicates.html and https://www.gnu.org/software/emacs/manual/html_node/elisp/Sets-And-Lists.html

deb0ch: Just read some docs about `member` using `equal` and `memq` using `eq` being stricter, but am still blurry about the implications.

layers/+misc/nlinum/funcs.el
deb0ch: Why not, also makes sense.
It was inspired from the `linum-on` function, used in `define-globalized-minor-mode` to define `global-linum-mode`, which purpose is similar.
But yeah, `maybe` might be more explicit.

layers/+distributions/spacemacs-base/packages.el
deb0ch: I guess, didn't know for sure so I called it exactly as it was called in the original `linum-on` function in `linum.el:116`

layers/+distributions/spacemacs-base/funcs.el
deb0ch: I guess which one is harder to read depends on how your brain is wired. Or maybe I'm too much of an imperative person, but it was easier for me to read and build it that way.

I still will try out your suggestion though. Even if I find it harder in your example perhaps will it be otherwise with the real stuff.

deb0ch:
Yep, thanks for your review !

I named it like that because it seemed to be consistent with the existing convention, and also because I couldn't think of other appropriate names.

Now that you say that it is indeed worth consideration, especially if the behaviour of the properties becomes different from the behaviour of the layers' properties (right now they are made to be the same).

As of the inheritance of major modes, that is what I was referring to by `handle derived modes`. I would like it to be handled as well, but found that I could leave it for further work and have this part integrated before.

I'm not sure of the use case as well, originally mine is just to disable linum in org-mode, markdown-mode and some others. Maybe there will be a text-mode derivative in which I will want them ? But yeah, supporting mode inheritance would indeed be more elegant.

layers/+misc/nlinum/funcs.el
deb0ch: Nope, it is supposed to be like that :smile_cat: 

You want to toggle the toggle every time it is called, but turn on line numbers only if they were not already on, right ?

layers/+distributions/spacemacs-base/packages.el
deb0ch: Just read about it a bit, and indeed it is only when called interactively that without argument it toggles the mode. So confusing :cold_sweat:

layers/+distributions/spacemacs-base/funcs.el
deb0ch: Ok, ok, I guess it *is* horrible to read when you did not write it yourself :laughing: 

I will change it for the `and` `or` form :ok_hand:

deb0ch:
Done changing what you requested :smile_cat: 

About your suggestion:
> specify "enabled for prog-mode, but disabled for c-mode"

I didn't think of that kind of combination in the first place, I think that it would indeed be nice to have it.

But it would modify the behaviour of `:enabled-for` and `:disabled-for` to make it differ from the layers' `:enabled-for` and `:disabled-for`, so in that case it might be worth renaming them.

Only when that change will happen might be in a long(ish) time, and I don't know what else to call them.

Any suggestion for new names ?

core/templates/.spacemacs.template
```diff
@@ -252,8 +252,20 @@ values."
...
+   ;; If set to `t' or `relative' line numbers are turned on in all `prog-mode' and
+   ;; `text-mode' derivatives. If set to `relative', line numbers are relative.
+   ;; This variable can also be set to a property list for finer control:
+   ;; '(:relative nil
+   ;; 	:disabled-for (dired-mode
+   ;; 	               doc-view-mode
+   ;; 	               markdown-mode
+   ;; 	               org-mode
+   ;; 	               pdf-view-mode
+   ;; 	               text-mode)
```
syl20bnr: You can use `spacemacs/mplist-get` instead of  `plist-get` like for the layer properties and then the list is not needed, the user car write:
 ```
:disabled-for dired-mode doc-view-mode ...
```

Comment (deb0ch) [inline core/templates/.spacemacs.template]: I can't help but wonder, what are the advantages of this ?

Is it just getting rid of the `()` or is there more ?

**Comment by deb0ch** [comment]
- [x] rename `:*abled-for` → `:*abled-for-modes`
- [x] use `spacemacs/mplist-get` to save on `()`
- [x] update the docs

All check !

**Comment by deb0ch** [comment]
Ping: What's the status on this ?

I think it complies with all of your reviews.

File: core/templates/.spacemacs.template [hunk 11]
```diff
@@ -252,8 +252,24 @@ values."
    ;; scrolling overrides the default behavior of Emacs which recenters point
    ;; when it reaches the top or bottom of the screen. (default t)
    dotspacemacs-smooth-scrolling t
+<<<<<<< afa2226051945b73c89fad46b56eb5fe6d4ece89
```
d12frosted: Whoa, @deb0ch 😸 🥇

deb0ch: OMG :scream_cat: You didn't see a thing. Where is my neuralyzer ? :dark_sunglasses:

deb0ch: I wonder why, but magit doesn't show these as an addition or deletion, but with the same color and face as the rest.
I must have missed because of something like that :smirk_cat:

d12frosted:
@deb0ch I am talking about [this](https://github.com/syl20bnr/spacemacs/blob/develop/doc/DOCUMENTATION.org#disabling-layer-services-in-other-layers).

deb0ch:
Oh ! Yes, originally I called it the same  to reflect that feature, but after our discussion with Sylvain in this thread we figured that it would be confusing to have the same names, so we called it differently and didn't come up with a better name.

If you have inspiration for a shorter one don't hesitate, else I think that this one can be fine.

deb0ch:
> I wonder if we can configure minor-mode disable options from one place.

I still don't understand this :confused:

deb0ch:
@syl20bnr @TheBB @d12frosted @bmag What is the status on this PR ?

syl20bnr:
Thank you ! 💜 

I made some modifications to the PR, first I dispatched the big condition into functions to improve readability, commit https://github.com/syl20bnr/spacemacs/commit/ddfb996
Then I had an error with my current config about `relative` being not a list, I added a condition function for backward compatibility in commit https://github.com/syl20bnr/spacemacs/commit/d5d80eb 

syl20bnr:
@deb0ch there was a bug, I fixed it in commit https://github.com/syl20bnr/spacemacs/commit/3f38ab0

deb0ch:
Thank you for taking the time to review, integrate and improve this PR :smiley_cat: 

The code is indeed much *much* better / more readable broke down into subfunctions like that.

Not sure what brought that bug, I would gladly test it for myself but I was facing a couple of issues when pulling develop this morning, so I'm pushing these tests to later :smile_cat:

syl20bnr:
@deb0ch I introduced the second bug, or more precisely I reintroduced the original bug, it is likely that the refactor I made was responsible in the first place.

PR #7676, opened 2016-11-07
Removed deprecated insert-string and replace by insert

ksjogo:
The `insert-string` functions is deprecated and to be replaced by `insert`.

### DIFF
--- a/core/tools/spacefmt/spacefmt.el
+++ b/core/tools/spacefmt/spacefmt.el
@@ -72,7 +72,7 @@
     (while (re-search-backward org-heading-regexp nil t))
     (open-line 3)
     (forward-line 1)
-    (insert-string toc-headline)))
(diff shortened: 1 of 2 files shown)


bmag:

eldamir:
@bmag Wouldn't something like this merit a hotfix? Master branch is broken. Cannot load spacemacs with a fresh install. That would probably hurt the adoption rate of spacemacs :frowning_face:

bmag:
Master branch is broken only on Emacs 26, which isn't a stable version and isn't even in feature freeze yet (or was there already a freeze and I missed the announcement?). It still works for all officially supported Emacs versions (24.4, 24.5, 25.1), and develop branch works also on Emacs 26. Personally I don't think it merits a hotfix, but anyway @syl20bnr is the one who decides on hotfixes, not me.

PR #7679, opened 2016-11-07
keyboard-layout: fix twittering-mode defun name

namdnguyen:
Twittering-mode defun in the keyboard-layout layer appears to have the old layer name of "kl". Changed to match layer name.

### DIFF
--- a/layers/+intl/keyboard-layout/packages.el
+++ b/layers/+intl/keyboard-layout/packages.el
@@ -466,7 +466,7 @@
       "k"
       "l")))
 
-(defun kl/pre-init-twittering-mode ()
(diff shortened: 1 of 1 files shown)


namdnguyen:
This fixes the warning message shown when spacemacs starts up.

    - package twittering-mode not initialized in layer keyboard-layout, you may 
      consider removing this package from the package list or use the :toggle keyword 
      instead of a `when' form.

namdnguyen:
Fixed in commit [635dde2](https://github.com/syl20bnr/spacemacs/commit/635dde29c839cc9083ba2c16dd35f6e18c5236d3). Closing.

PR #7680, opened 2016-11-07
bind helm-cscope-find-assignments-to-this-symbol

dubnde:
Add cscope layer binding `SPC m g =` to `helm-cscope-find-assignments-to-this-symbol`


TheBB:

PR #7692, opened 2016-11-08
Added extra information

dineshbhosale:
Added extra information for setting default tab width for go files.

### DIFF
--- a/layers/+lang/go/README.org
+++ b/layers/+lang/go/README.org
@@ -89,6 +89,12 @@ the layer variable =go-tab-width=, e.g.
   (go :variables go-tab-width 4)
 #+end_src
 
+Or add this in =dotspacemacs/user-config ()= in your =.spacemacs= file
+
+#+begin_src emacs-lisp
+  (setq go-tab-width 4)
+#+end_src
+
 ** Tests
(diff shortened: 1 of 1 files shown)


LemmingAvalanche:
I would squash those two commits if I were you.

Stebalien:
I don't really see how that extra bit of documentation is useful; the documentation already gives you a way to set the tab width. This is one of those cases where I feel that less is more (don't confuse users with options).

KamranMackey:
@LemmingAvalanche They don't need to. GitHub offers automatic squashing capabilities when you merge the PR.

KamranMackey:
@Stebalien Yeah, I agree.

LemmingAvalanche:
@KamranMackey The repository owner tends to cherry-pick commits from pull requests, though. Of course Sylvain could squash them after having cherry-picked both of them (if he were to cherry-pick both). Personally though I would prefer to only rewrite my own commits, and not others’.

bmag:
@KamranMackey we prefer simple pull requests to be squashed into a single commit, please. We usually don't use GitHub's button because we want more control over the process.

KamranMackey:
@bmag Ah, alrighty then. :)

PR #7695, opened 2016-11-09
non-restricted-buffer-list-* to SPC b B instead of SPC B b and change which-key names

deb0ch:
`SPC b B` is more intuitive for listing all buffers and was only previously used in `ibuffer` layer, only when activated (which I suspect is not the case for most user).

Picked `SPC b I` instead, as it is mnemonic for `ibuffer`.

I don't know what was the thought behind this, but `spacemacs-layouts/non-restricted-buffers-list-*` was alone in its `SPC B` prefix.

I also renamed buffers listing functions in `which-key`. 
This way it is clearer for the user: names like `helm-mini` are pretty obscure and kind of defeat the purpose of `which-key`. `spacemacs-layouts-non-restricted-buffer-list-blah` had also such a long name that it couldn't even be displayed.

The user can now choose between `list-buffers` or `global-list-buffers` for listing buffers.

### DIFF
--- a/layers/+distributions/spacemacs-bootstrap/packages.el
+++ b/layers/+distributions/spacemacs-bootstrap/packages.el
@@ -302,7 +302,9 @@
            ("spacemacs/toggle-holy-mode" . "emacs (holy-mode)")
            ("evil-lisp-state-\\(.+\\)" . "\\1")
            ("spacemacs/\\(.+\\)-transient-state/\\(.+\\)" . "\\2")
-           ("spacemacs/\\(.+\\)-transient-state/body" . "\\1-transient-state"))))
+           ("spacemacs/\\(.+\\)-transient-state/body" . "\\1-transient-state")
+           ("helm-mini\\|ivy-switch-buffer" . "list-buffers")
+           ("spacemacs-layouts/non-restricted-buffer-list-\\(helm\\|ivy\\)" . "global-list-buffers"))))
     (dolist (nd new-descriptions)
       ;; ensure the target matches the whole string
       (push (cons (cons nil (concat "\\`" (car nd) "\\'")) (cons nil (cdr nd)))
--- a/layers/+emacs/ibuffer/packages.el
+++ b/layers/+emacs/ibuffer/packages.el
@@ -20,7 +20,7 @@
     :defer t
     :init
     (progn
-      (spacemacs/set-leader-keys "bB" 'ibuffer)
+      (spacemacs/set-leader-keys "bI" 'ibuffer)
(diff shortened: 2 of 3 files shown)


layers/+distributions/spacemacs-bootstrap/packages.el
```diff
@@ -301,7 +301,9 @@
            ("spacemacs/toggle-holy-mode" . "emacs (holy-mode)")
            ("evil-lisp-state-\\(.+\\)" . "\\1")
            ("\\(.+\\)-transient-state/\\(.+\\)" . "\\2")
-           ("\\(.+\\)-transient-state/body" . "\\1-transient-state"))))
+           ("\\(.+\\)-transient-state/body" . "\\1-transient-state")
+           ("helm-mini\\|ivy-switch-buffer" . "list-buffers")
+           ("spacemacs-layouts/non-restricted-buffer-list-\\(helm\\|ivy\\)" . "global-list-buffers"))))
```
bmag: Not good names because there is actually  a command called "list-buffers" (or "buffers-list"). Maybe "switch buffer" is a better name.

deb0ch:
`list-buffers` was indeed confusing, I didn't know the existence of that command.

I'm proposing `b → buffers` and `B → global-buffers`, which I find very concise.

I think that `switch-buffers` and `global-switch-buffers` are longer and don't carry the meaning that you can use the command just to see a detailed list of opened buffers (Helm provides about as much information as `ibuffer` or `list-buffers`).

bmag:
I'm not sure if the proposed key bindings are better (will let @syl20bnr and @TheBB decide on that), but I like `b → buffers` and `B → global-buffers` :+1:

deb0ch:
Glad that you like `b → buffers` and `B → global-buffers` :smile_cat: 

It is only a proposal, but I find it way more instinctive, and it is one of the very basic things that everybody will always want to use so I think it is worth giving it the most obvious place.

The only unknown is what was originally expected to go under `SPC B` in the future.

But 90% of the `SPC b` bindings are buffer-local, and the only ones which could have global/non-global versions (`m → kill-other-buffers` and `C-d → kill-matching-buffers-rudely`) have their capitalized equivalent free, so if we ever want to introduce global/non-global versions we still can keep consistency this way and have only one prefix for buffer-related bindings.

So yes, waiting for @syl20bnr and @TheBB input: what do you think guys ? :smiley_cat:

deb0ch:
ping

deb0ch:
ping @syl20bnr @TheBB @d12frosted

TheBB:

PR #7701, opened 2016-11-11
Prefer .spacemacs.d/snippets over .emacs.d/private/snippets

aaronjensen:
The default snippet directory configuration when you use a dotspacemacs dir set
the primary snippet directory to `.emacs.d/private/snippets`. The primary
snippet directory is where snippets are saved by default. If a user has a
dotspacemacs dir, I believe it's safe to assume that that is where they want
their snippets unless they override with
`auto-completion-private-snippets-directory`. 

There are multiple ways to implement this. I chose to exclude
`.emacs.d/private/snippets` in this case. Another alternative would be to always
include it, but include it after the `.spacemacs.d/snippets` directory. I'm
happy to switch to that if you'd prefer it.

### DIFF
--- a/layers/+completion/auto-completion/packages.el
+++ b/layers/+completion/auto-completion/packages.el
@@ -195,11 +195,9 @@
       ;; configure snippet directories
       (let* ((spacemacs--auto-completion-dir
               (configuration-layer/get-layer-local-dir 'auto-completion))
-             (private-yas-dir (if auto-completion-private-snippets-directory
-                                  auto-completion-private-snippets-directory
-                                (concat
-                                 configuration-layer-private-directory
-                                 "snippets/")))
+             (emacs-directory-snippets-dir (concat
+                                          configuration-layer-private-directory
+                                          "snippets/"))
              (spacemacs-layer-snippets-dir (expand-file-name
                                       "snippets"
                                       spacemacs--auto-completion-dir))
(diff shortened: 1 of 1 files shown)


d12frosted:
Nice! In my opinion second solution is preferable, though I agree with your thinking about private snippets directory of users with spacemacs dir set.

aaronjensen:
Cool, updated. Old commit for posterity:

https://github.com/aaronjensen/spacemacs/commit/2a6cf92f14bd13f5dcd562f649f7b6349a00921e

layers/+completion/auto-completion/packages.el
```diff
@@ -208,14 +206,16 @@
...
           (push dotspacemacs-directory-snippets-dir yas-snippet-dirs))
         ;; arbitrary directories in `auto-completion-private-snippets-directory'
-        (when private-yas-dir
-          (if (listp private-yas-dir)
-              (setq yas-snippet-dirs (append yas-snippet-dirs private-yas-dir))
-            (push private-yas-dir yas-snippet-dirs))))
+        (when auto-completion-private-snippets-directory
+          (if (listp auto-completion-private-snippets-directory)
+              (setq yas-snippet-dirs (append yas-snippet-dirs auto-completion-private-snippets-directory))
+            (push auto-completion-private-snippets-directory yas-snippet-dirs))))
```
syl20bnr: We should switch `push` functions to `add-to-list` because I'm afraid that any config reloading with `SPC f e R` may append the same directories again and over again.

This remark is for all the `push` in this `use-package` sexp.

aaronjensen: Line 204 is `(setq yas-snippet-dirs nil)` so I don't think that holds here. I'm happy to switch it if you think it still necessary, but it seems unnecessary given this.

aaronjensen:
@syl20bnr @d12frosted ping

Pharserror:
bump

aaronjensen:
I rebased this, @syl20bnr please let me know if there's anything I should address. Thanks!

TheBB:
Works for me. Thanks! Cherry-picked in develop. You can safely delete your branch.

PR #7705, opened 2016-11-11
Add ripgrep to search tools in template

houli:
https://github.com/syl20bnr/spacemacs/commit/385c12743a6108936d88504c516ff23bbe1b0c61
added ripgrep support but didn't add documentation in the template or update the default value.


bmag:
Thank you :+1:

PR #7706, opened 2016-11-11
Add Doom themes to required dependencies for dash

chriskaukis:
Doom themes fail to load because it's unable to find dash. This change fixes the issue by adding them to the list of themes that require dash as a dependency in the core-themes-support.el.


syl20bnr:
Thank you ! 👍 

PR #7707, opened 2016-11-11
Remove git layer documentation to show submodule prompt

dubnde:
There is no such keybinding `SPC g >` in the source code.


TheBB:

Issue #7702, opened 2016-11-11
I can't open call find-file-at-point with <g f> in Compilation mode


FrancescElies:
#### Description :octocat:
I can't open call `find-file-at-point` with `<g f>` in `Compilation` mode

#### Reproduction guide :beetle:
- Start Emacs
- Open a python file with some tests
- Run some test with `<, t t>`
*Observed behaviour:* :eyes: :broken_heart:
A new buffer in compilation mode with the results of the test is opened.
But I can't call `find-file-at-point` typing `<g f>`

*Expected behaviour:* :heart: :smile:
I would love to call `find-file-at-point` typing `<g f>` in `Compilation` mode

#### System Info :computer:
- OS: darwin
- Emacs: 25.1.1
- Spacemacs: 0.200.3
- Spacemacs branch: master (rev. b3918e8)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
((auto-completion :variables auto-completion-enable-snippets-in-popup t auto-completion-enable-help-tooltip t auto-completion-enable-sort-by-usage t
                  (haskell :variables haskell-completion-backend 'intero))
 better-defaults deft
 (elfeed :variables rmh-elfeed-org-files
         (list "~/.private/feed-index.org")
         elfeed-search-filter "@1-week-ago  unread ")
 csv
 (c-c   :variables c-c  -default-mode-for-headers 'c  -mode c-c  -enable-clang-support t)
 command-log emacs-lisp elm emoji erc fasd
 (git :variables git-magit-status-fullscreen t git-enable-github-support t git-gutter-use-fringe t)
 github gnus gtags haskell html ipython-notebook javascript latex markdown
 (org :variables org-agenda-files
      (list "~/org/sprints/fred.org" "~/org/home.org"))
 osx
 (python :variables python-test-runner 'pytest)
 (ranger :variables ranger-cleanup-eagerly t ranger-cleanup-on-disable t ranger-ignored-extensions
         '("mkv" "iso" "mp4")
         ranger-max-preview-size 10 ranger-parent-depth 2 ranger-show-preview t)
 restclient ruby search-engine semantic slack
 (shell :variables shell-default-shell 'multi-term)
 speed-reading spell-checking syntax-checking typescript
 (version-control :variables version-control-global-margin t)
 yaml)
```
[...]

PR #7712, opened 2016-11-12
compilation-mode: maps `gf` to `find-file-at-point`

FrancescElies:
Hi,

This PR maybe solves https://github.com/syl20bnr/spacemacs/issues/7702

I found two ways of adding the keybinding, but not sure if this is the right thing and the right place.

```
;; option 1
  (define-key evil-motion-state-map "gf" 'find-file-at-point)
;; option 2
  (evil-define-key 'motion compilation-mode-map (kbd "gf") 'find-file-at-point)
```

Guidance would be highly appreciated.

All best

### DIFF
--- a/layers/+distributions/spacemacs-base/keybindings.el
+++ b/layers/+distributions/spacemacs-base/keybindings.el
@@ -234,6 +234,7 @@
   "cr" 'recompile
   "cd" 'spacemacs/close-compilation-window)
 (with-eval-after-load 'compile
+  (evil-define-key 'motion compilation-mode-map (kbd "gf") 'find-file-at-point)
   (define-key compilation-mode-map "r" 'recompile)
   (define-key compilation-mode-map "g" nil))
 ;; narrow & widen -------------------------------------------------------------


bmag:
`(kbd "gf")` returns `"gf"`, so there isn't any difference. In other cases there is a difference, for example to bind the return key you'd use `(kbd "RET")` but not `"RET"`.

bmag:
Also, we probably want to bind the keys in [spacemacs-base](https://github.com/syl20bnr/spacemacs/blob/58c8fd515a9a0ab9bde21cfb531a5541c353b77a/layers/+distributions/spacemacs-base/keybindings.el#L236), not spacemacs-bootstrap.

layers/+distributions/spacemacs-base/keybindings.el
```diff
@@ -235,6 +235,7 @@
   "cd" 'spacemacs/close-compilation-window)
 (with-eval-after-load 'compile
   (define-key compilation-mode-map "r" 'recompile)
+  (define-key evil-motion-state-map "gf" 'find-file-at-point)
   (define-key compilation-mode-map "g" nil))
```
FrancescElies: Just being curious, what is `(define-key compilation-mode-map "g" nil)` doing, is this really needed?

FrancescElies:
keybinding moved to spacemacs-base

layers/+distributions/spacemacs-base/keybindings.el
```diff
@@ -235,6 +235,7 @@
   "cd" 'spacemacs/close-compilation-window)
 (with-eval-after-load 'compile
   (define-key compilation-mode-map "r" 'recompile)
+  (define-key evil-motion-state-map "gf" 'find-file-at-point)
```
bmag: Should use evil-define-key like before

FrancescElies: Updated https://github.com/syl20bnr/spacemacs/pull/7712/commits/12517cf695043b49ad7cc3c489af37cb0c270ab6
Is this one fine?

bmag: Yes (but I didn't test it locally yet)

FrancescElies: I still could not understand the difference between both if any, if I remember correctly both did the same. 

Thanks for your help 👍

bmag:

Issue #6924, opened 2016-08-26
Form feed character in `insert-page-break` causes issue with `Inconsolata`


menelaos:
#### Description

With the `Inconsolata` font , the form feed character (`U+000C`) in `spacemacs-buffer/insert-page-break` causes the home buffer content to shift when the cursor is over the horizontal line. Sometimes this behaviour stops after using `Spacemacs` for a while but I haven't been able to reproduce this reliably.
Is the form feed character really necessary? A similar visual effect could be achieved with e.g. `U+2500`.
#### Reproduction guide
- Install `Inconsolata` font and set it as default font in `.spacemacs`:
  
  ```
    dotspacemacs-default-font '("Inconsolata"
                                :size 16
                                :weight normal
                                :width normal
                                :powerline-scale 1.1)
  ```
- Start Emacs
- Press `j` until the cursor is over the horizontal line

_Observed behaviour:_
Cursor jumps to the end of the line and consequently, the home buffer content shifts to the left.

_Expected behaviour:_
There should be no sudden jump when the cursor is over the horizontal line.
#### System Info
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.22
- Spacemacs branch: master (rev. 9f9faa4)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:

``` elisp
(auto-completion colors deft emacs-lisp eyebrowse git haskell markdown org purescript react
                 (shell :variables shell-default-shell 'eshell shell-default-height 30 shell-default-position 'full)
                 themes-megapack)
```

andreybratukhin:
I think, the horizontal line should not exceed the current buffer width. In this case the jump to the end of line will not affect the current buffer so much.

andreybratukhin:
The following code creates a page break line from '─' (box drawings light horizontal) symbol by filling it up to window width.

    (defun spacemacs-buffer/insert-page-break ()
      "Insert a page break line in spacemacs buffer."
      (spacemacs-buffer/append (concat "\n" (make-string (window-body-width) ?\u2500) "\n")))

PR #7717, opened 2016-11-13
Change page break line on *spacemacs* buffer

andreybratukhin:
Fix the issue #6924 .
Instead of inserting one ^L page break symbol it creates a page break line with
using \u2500 (box drawings light horizontal) symbol. It improves rendering
when an user navigates through a page break line.

### DIFF
--- a/core/core-spacemacs-buffer.el
+++ b/core/core-spacemacs-buffer.el
@@ -538,7 +538,7 @@ ARGS: format string arguments."
 
 (defun spacemacs-buffer/insert-page-break ()
   "Insert a page break line in spacemacs buffer."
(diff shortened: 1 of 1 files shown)


bmag:
link please: #6924

syl20bnr:
Thank you for the PR!

The rendering is not as nice as `^L` in both the GUI and the terminal, moreover in the terminal there is a `$` character added at the end of the lines.

`^L` is the standard way to insert a break line in Emacs with the page-break package, not sure what happens with Inconsolata but pressing `0` should put the buffer back into its original place which is a acceptable solution to me to keep the current page breaks :-)

TheBB:
Closing due to :point_up:

PR #7728, opened 2016-11-14
Fix typo in Agda's packages.el

nightuser:
`pre-init-golden-ratio` function was copy-pasted from Idris layer without necessary change applied. It causes a warning during startup.


d12frosted:

PR #7729, opened 2016-11-14
Resolve symlinks when warning about duplicate layers

bgamari:
This allows the user to, for instance, include the spacemacs tree as a
git submodule of their configuration repository without seeing
superfluous duplicate layer warnings.


TheBB:

PR #7733, opened 2016-11-14
Change `SPC b m` to switch to the messages buffer.

### LINKED ISSUE #7732: `SPC b m` should switch to Messages buffer, not kill other buffers.
Title says it all, `spacemacs/kill-other-buffers` doesn't fit the mnemonic, and also seems less useful than switching to the messages buffer.

I propose mapping `SPC b m` to a new function `spacemacs/switch-to-messages-buffer`. Perhaps `spacemacs/kill-other-buffers` could be moved to something like `SPC b D` (a kind of alternate delete) or `SPC b o` (`o` is used for "other" in some other mnemonics).

On a related note, `*Messages*` should probably be marked as a "useful" buffer, so `SPC TAB` will switch back to it if `SPC b m` switches to it.

mwillsey:
Resolves #7732.

### DIFF
--- a/layers/+distributions/spacemacs-base/funcs.el
+++ b/layers/+distributions/spacemacs-base/funcs.el
@@ -967,11 +967,6 @@ a split-side entry, its value must be usable as the SIDE argument for
                (fboundp dotspacemacs-scratch-mode))
       (funcall dotspacemacs-scratch-mode))))
 
-(defun spacemacs/switch-to-messages-buffer ()
-  "Switch to the `*Messages*' buffer."
-  (interactive)
-  (switch-to-buffer (messages-buffer)))
-
 (defun spacemacs/close-compilation-window ()
   "Close the window containing the '*compilation*' buffer."
   (interactive)
--- a/layers/+distributions/spacemacs-base/funcs.el
+++ b/layers/+distributions/spacemacs-base/funcs.el
@@ -967,6 +967,11 @@ a split-side entry, its value must be usable as the SIDE argument for
                (fboundp dotspacemacs-scratch-mode))
       (funcall dotspacemacs-scratch-mode))))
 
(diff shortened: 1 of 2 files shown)


nixmaniack:
There's also `view-echo-area-messages` bound to `C-h e` which opens messages buffer.

mwillsey:
Good call, it should probably use that instead.

bmag:
Here's the source of `view-echo-area-messages`: (it also moves the cursor to the end of the messages buffer)
```elisp
(defun view-echo-area-messages ()
  "View the log of recent echo-area messages: the `*Messages*' buffer.
The number of messages retained in that buffer
is specified by the variable `message-log-max'."
  (interactive)
  (with-current-buffer (messages-buffer)
    (goto-char (point-max))
    (display-buffer (current-buffer))))
```
We can implement `spacemacs/switch-to-messages-buffer` in the same way, but with `pop-to-buffer` or `switch-to-buffer-other-window` instead of `display-buffer`. @mwillsey is there a reason to use `switch-to-buffer` (opens in current window) instead of the other two options?

mwillsey:
I think the only argument for using `switch-to-buffer` is that it's consistent with the other buffer manipulation commands; most of them only deal with your current window. TBH I just copied the code for `spacemacs/switch-to-scratch-buffer`. 

I'm not tied to this behavior. I think the `view-echo-area-messages` may make more sense because you wouldn't edit that buffer, so I guess you don't really need to have focus on it.

bmag:
I'm also in favor of just using `view-echo-area-messages`, so I guess we all agree on this :smile:

syl20bnr:
Thank you !
I updated the documentation and moved `kill other buffers` to `SPC b C-D` (so it moved kill buffers by regexp to `SPC b C-D`), commit https://github.com/syl20bnr/spacemacs/commit/237b2a7

PR #7735, opened 2016-11-15
Enable projectile-rails keybindings with additional modes. Attempt to…

### LINKED ISSUE #2448: Projectile-rails keybindings need to work for haml and slim as well.
Currently projectile-rails keybindings do not work for slim and haml files.

minusoneman:
Fix #2448

### DIFF
--- a/layers/+frameworks/ruby-on-rails/packages.el
+++ b/layers/+frameworks/ruby-on-rails/packages.el
@@ -27,7 +27,7 @@
       (spacemacs|diminish projectile-rails-mode " ⇋" " RoR")
 
       ;; Find files
-      (dolist (mode '(ruby-mode enh-ruby-mode))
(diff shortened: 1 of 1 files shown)


syl20bnr:
There is a new pattern to do this "cleanly":

Example:
- declare a private config variable: https://github.com/syl20bnr/spacemacs/blob/develop/layers/+lang/java/config.el#L20-L21
- hook modes to it with use-package hook: https://github.com/syl20bnr/spacemacs/blob/develop/layers/+lang/scala/packages.el#L30-L32
- then use the config variable: https://github.com/syl20bnr/spacemacs/blob/develop/layers/+lang/java/packages.el#L175-L187

juanibiapina:
How about from `.gitignore`, `.rspec-local`, `Makefile` or any other file you can possible have in your rails directory?

asok:
> How about from .gitignore, .rspec-local, Makefile or any other file you can possible have in your rails directory?

This is covered by having `projectile-rails` be a global mode. This can be closed.

PR #7736, opened 2016-11-15
Make Semantic fast

tuhdo:
Limit the parsing scope to file, local and project. The choices are listed in semanticdb-find-default-throttle variable.

### DIFF
--- a/layers/+emacs/semantic/README.org
+++ b/layers/+emacs/semantic/README.org
@@ -38,6 +38,10 @@ To use this configuration layer, add it to your =~/.spacemacs=. You will need to
 add =semantic= to the existing =dotspacemacs-configuration-layers= list in this
 file.
 
+By default, Spacemacs sets Semantic to parse only file, local and project scope.
+For a different parsing scope, you can customize the variable
+=semanticdb-find-default-throttle=.
+
 * Key Bindings
 
 | Key Binding | Description                         |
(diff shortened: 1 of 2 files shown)


syl20bnr:
Thank you ! 👍 

braham-snyder:
This alone didn't quite work for me, due to a mode-local default for `emacs-lisp-mode`  in my `/usr/local/Cellar/emacs-plus/25.2/share/emacs/25.2/lisp/cedet/semantic/db-el.el.gz` that kept resetting my `semanticdb-find-default-throttle` to include `omniscience`:

```
(defvar-mode-local emacs-lisp-mode semanticdb-find-default-throttle
  '(project omniscience)
  "Search project files, then search this omniscience database.
It is not necessary to do system or recursive searching because of
the omniscience database.")
```

This resets the mode-local version to the global, for what it's worth:

```
(use-package semantic
  :config
  (setq-mode-local emacs-lisp-mode
                   semanticdb-find-default-throttle
                   (default-value 'semanticdb-find-default-throttle)))
```

which might or might not result in some missing `semantic` functionality, but I was really just looking for `stickyfunc-mode`, which this appears to suffice for.

Lastly, just for those curious, I found that snippet in `db-el.el.gz` from the `semantic` directory (`find-variable` `semanticdb-find-default-throttle` to find the directory) then `spacemacs/search-ag` with input `-uz semanticdb-find-default-throttle` (looks like both flags are necessary, weirdly, even though the file isn't an `.elc.gz`?).

quicknir:
Can this be re-opened, and @braham-snyder 's suggested change be merged into spacemacs proper somehow? Right now, it's still the case that by default in a vanilla spacemacs setup with only semantic added, performance in emacs-lisp mode is simply terrible. Surely this isn't ok?

PR #7742, opened 2016-11-15
add refactoring for go-mode

TinySong:
SPC m r for gorename move to SPC m R
SPC m r for refactoring

### DIFF
--- a/layers/+lang/go/packages.el
+++ b/layers/+lang/go/packages.el
@@ -12,6 +12,7 @@
         go-mode
         go-guru
         (go-rename :location local)
+        godoctor
         ))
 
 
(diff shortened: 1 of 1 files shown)


layers/+lang/go/packages.el
```diff
@@ -131,8 +132,8 @@
 (defun go/init-go-rename()
   (use-package go-rename
     :init
-    (spacemacs/declare-prefix-for-mode 'go-mode "mr" "rename")
```
db47h: Wouldn't it make sense to simply rename "rename" to "refactor" then add godoctor's shortcuts under it alongside with the current "go-rename"?

TheBB: I agree, renaming is a refactoring operation.

TheBB:

PR #7747, opened 2016-11-16
Documentation has outdated shortcut

rodonn:
The suggestions to use SPC ; SPC y is from the previous versions where SPC y was bound to avy-goto-line.
Now that command is bound to SPC j l


syl20bnr:
Thank you ! 👍 

PR #7748, opened 2016-11-16
Remove backtick from smartparens pairs for Clojure

ekroon:
Single backticks are used for namespaced quotes, there is no purpose for
automatically inserting double backticks.

### DIFF
--- a/layers/+lang/clojure/packages.el
+++ b/layers/+lang/clojure/packages.el
@@ -269,7 +269,8 @@
   (add-hook 'cider-repl-mode-hook
             (if dotspacemacs-smartparens-strict-mode
                 #'smartparens-strict-mode
-              #'smartparens-mode)))
+              #'smartparens-mode))
+  (sp-local-pair 'clojure-mode "`" nil :actions nil))
 
 (defun clojure/post-init-subword ()
   (add-hook 'cider-mode-hook 'subword-mode))


ekroon:
I am not completely sure this is the right place, but it works at my machine ™

ekroon:
I had no issues with this fix for the last 12 days (I have it applied at master). Is there anything I can do to get this merged?

layers/+lang/clojure/packages.el
```diff
@@ -269,7 +269,8 @@
   (add-hook 'cider-repl-mode-hook
             (if dotspacemacs-smartparens-strict-mode
                 #'smartparens-strict-mode
-              #'smartparens-mode)))
+              #'smartparens-mode))
+  (sp-local-pair 'clojure-mode "`" nil :actions nil))
```
d12frosted: Why is it under hook? Shouldn't it be just the next sexp after `add-hook`? :)

d12frosted: P. S. sorry for long delay.

ekroon: It is the next sexp after `add-hook` :). First bracket closes `if` and second closes `add-hook`, after that the `sp-local-pair`). Easier to check in full file, but `sp-local-pair` is second sexp in `clojure/post-init-smartparens`.

TheBB:

I moved your call to `sp-local-pair` into a `with-eval-after-load` clause.

Issue #7752, opened 2016-11-16
Can't create new <object> using Ivy when a match exists.


gilbertw1:
#### Description :octocat:
When using Ivy can't create any 'object' whose name matches part of an existing match in the minibuffer. This affects a number of different functions including layouts, creating files, renaming files, virtually anything that opens an ivy completion, yet also allows for a new non-match (:require-match nil) value.

For example if I were to try and create a new file using ```SPC f f``` called 'Module.scala' and had an existing file in the same directory called 'LongNameModule.scala', the ivy buffer would only show 'LongNameModule.scala' as a selectable match and provides no way to actually create 'Module.scala'. A similar problem would occur if I tried to create a layout 'emacs' using ```SPC b b``` and a layout named 'emacs-test' already existed.

Using helm either of these scenarios would show an exact match under a heading that says 'Create ...' that is selectable. 

This presents a pretty big hurdle when using the Ivy layer since spacemacs uses this pattern pretty pervasively and is the primary way to create files / layers / buffers / etc.

#### Reproduction guide :beetle:
- Start Emacs
- Press ```SPC f f```
- Attempt to create file 'Module.scala' in directory that contains 'LongNameModule.scala'

*Observed behaviour:* :eyes: :broken_heart:
No option to create the file 'Module.scala'

*Expected behaviour:* :heart: :smile:
'Module.scala' should be a selectable option in some way allowing me to create the file.

gilbertw1:
After a bit more digging, I found that ivy has ```ivy-immediate-done``` which is bound to ```C-M-j```. This will exit with the current input which will allow me to create something that doesn't show up in the match list. Maybe we can find some way expose this functionality in a more Spacemacs friendly way?

gilbertw1:
After using ```C-<return>``` to enter a non-match, I can say that I actually prefer this method to having to select it from the completion list in Helm.

duianto:
It's now possible to select the "Find File: " prompt with a partial name of a current match.

This capability was added by @fabacino with this PR:
ivy.el: Make prompt line selectable abo-abo/swiper#1059

Set the new `ivy-use-selectable-prompt variable` to `t`, and move up the selection `C-p` to the prompt above the first match.

PR #7753, opened 2016-11-16
rebind 'ivy-immediate-done' to 'C-<return>' in ivy minibuffer

gilbertw1:
This solves #7752 by allowing users to use arbitrary non-matching text as the result of an ivy completion even when a matching completion is available.

### DIFF
--- a/layers/+spacemacs/spacemacs-completion/packages.el
+++ b/layers/+spacemacs/spacemacs-completion/packages.el
@@ -180,6 +180,7 @@ Current Action: %s(ivy-action-name)
       ("f" ivy-toggle-case-fold)
       ("o" ivy-occur :exit t))
     (define-key ivy-minibuffer-map "\C-o" 'spacemacs/ivy-transient-state/body)
+    (define-key ivy-minibuffer-map (kbd "C-<return>") 'ivy-immediate-done)
     ))
 
 (defun spacemacs-completion/init-ido ()


nixmaniack:
As an ivy user, not sure if I'm in favour of this as it's already provided by `C-M-j`. It definitely is obscure and unintuitive if you haven't used `ivy` or followed its [documentation](http://oremacs.com/swiper/#key-bindings) but so are other bindings of it (`M-o`, `C-'`, `C-M-m`, `C-M-o` etc). 

But since it doesn't conflict with other ivy bindings it shouldn't be problematic. 

Let's see what other ivy users say.

gilbertw1:
I understand your position and I was a bit torn making this change.

For background sake, I've come to use Ivy directly as a result of using Spacemacs and as such used it only using keybindings I had picked up while using Spacemacs. I have no experience with Emacs outside of Spacemacs, but I feel like Spacemacs has a very well established feel to it and it does a great job keeping that consistent as it brings in many different packages and rarely (if ever) forces me to use emac-sy keybindings.

As a Spacemacs user ```C-M-j``` is an incredibly awkward keybinding and I can honestly say that there is no other keybinding that I use regularly in Spacemacs that requires me to use the alt key. While this wouldn't be a huge issue if it had a more obscure use case, this keybinding is required in many cases to create a file, layout, or buffer. Requiring regular use of such keybindings makes Ivy feel like it's not very well integrated into Spacemacs and makes it feel more like an odd ball.

bmag:
A [similar PR](https://github.com/abo-abo/swiper/pull/418) was proposed upstream and rejected because `C-<return>` translates to `C-j` inside a terminal, causing `C-<return>` to have different meanings between graphical and terminal Emacs. I'm reluctant to merge this change for the same reason. Maybe you can find an alternative binding?

Inviting @timofeytt, @madand  and @duianto to participate (related issue: #8715)

gilbertw1:
I seem to have completely missed that PR against ivy, that would definitely be a problem :)

I've taken to just rebinding this within my Spacemacs config files. It would be nice to have this rebound to something more friendly, but I have a tough time thinking of anything else that would make sense and won't conflict.

TheBB:
I'm closing this since there seems to be no consensus. Agree that we should document this somehow.

sdwolf:
The `C-M-j` keybinding is already documented on develop, see: https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Bcompletion/ivy/README.org.

Issue #5708, opened 2016-04-06
Add ensime java mode to java layer?


brcolow:
Currently it seems that the java layer has only eclim support - but there is an alternative - namely: ensime (which has a java mode "enjime").

More details here: http://ensime.github.io/editors/emacs/

I am not proficient enough with LISP, Emacs, or Spacemacs to try and implement this myself, but am opening it as an issue as I believe it would be a good addition to the java layer (and, more honestly, would love to try it myself!).

Thanks.

biocyberman:
This will be very nice. I am using ensime-based scala-mode and like it very much.

thang1thang2:
Funny that you mention this. I just went through the trouble of doing this around yesterday (I haven't made a layer, I just shoehorned the code into the `.spacemacs` file). I'll put up what I did to get it working here in case that's helpful.

Changed `dotspacemacs-additional-packages` to include `'(ensime, sbt-mode)`

This was placed in my `dotspacemacs/user-config()` function

```elisp
  (setq exec-path (append exec-path '("/usr/local/bin")))
  (setq exec-path (append exec-path '("/usr/local/sbin")))
  (setenv "PATH" (shell-command-to-string "/bin/bash -c 'echo -n $PATH'"))

  ;; this obviously isn't required for ensime but it is for eclim. I probably don't need these two lines
  (setq eclim-eclipse-dirs "/usr/lib/eclipse"
        eclim-executable "/usr/lib/eclipse/eclim")

  (use-package ensime
    :commands ensime ensime-mode)
  (add-hook 'java-mode-hook 'ensime-mode)
  (add-hook 'scala-mode-hook 'ensime-mode)

  (use-package sbt-mode
    :commands sbt-start sbt-command
    :config
    ;; WORKAROUND: https://github.com/ensime/emacs-sbt-mode/issues/31
    ;; allows using SPACE when in the minibuffer
    (substitute-key-definition
     'minibuffer-complete-word
     'self-insert-command
     minibuffer-local-completion-map))
```

Everything after this point is just installing sbt onto your computer, and setting up stuff outside of emacs. To finish the actual *layer* off would require key binding. I would suggest a variable for the java package such as `java-backend` with the parameters either `ecim` or `ensime` and then enable each setup as necessary.

[...]

PR #7755, opened 2016-11-16
Add Ensime support to java layer

hedning:
### Description
To add Ensime support for java a lot of the setup code in the scala layer is moved to a dependency layer, which both scala and java can depend through `layers.el`. This creates very little code duplication. See #5708

The active backend is controlled by a new variable, `java-backend`, which is set to `'eclim` by default.

I put the new layer in `layers/+lang/ensime` which might be sup-optimal. Is it better to put it under `layers/+misc` or `layers/+tools`?

#### Scala
The scala layer should work just as before, the only change there is renamed which-key prefixes, eg. scala/build -> build.

#### Java
The backends are at the moment mutually exclusive and only the selected backend will be installed. If eclim is used there should be no change.

The scala keybindings are reused for java. This is both easy and creates consistency for those who use both layers.

#### Issues
There's some issues with using Ensime with java, though most of this is upstream:
- function annotations are formatted in a weird way (eg. splitString[] split(String arg0))
- `ensime-expand-region-mark-syntactic-context` doesn't work and is added to `er/try-expand-list` when using the scala layer
-  refactoring isn't support yet

#### Further work
There's some things which should probably only be in the scala layer, and some more `sbt` support which should be shared. In addition some functions in `ensime/funcs.el` should instead be defined when ensime is loaded.

Edit: should now enable completions consistently


arichiardi:
Great i am going to test this soon :smile:

syl20bnr:
Excellent 👍 

I made a lot of changes to add support for multiple backend to java layer.
I moved Ensime configuration to the `java` layer so now `scala` layer depends on the `java`  layer instead of the `ensime` layer.

I tried to setup Ensime but failed. Can you test if there is any regression ? There is this thread to post the eventual regressions (or create a new issue or PR): https://github.com/syl20bnr/spacemacs/issues/7806

diego-alvarez-hs:
I really liked the idea of having `ensime` under `misc/tools`.

I personally don't like the `scala` layer depending on the `java` layer, 
now my machine has some java emacs packages that I don't need/want and I don't have plans to use them. 😢 

even scala/java have some similarities, I think we shouldn't put all `JVM` languages depending on the java layer

PR #7759, opened 2016-11-16
Add tide keybindings to tsx web mode

george-miller:
mode so we can properly edit/format the html that's inside. However, with this
the current function/variable. You also lose the ", g u" keybinding which
gives you all references of current function/variable. This PR adds
all the tide bindings that are in typescript mode into web mode while using the typescript layer.

NOTE: 
I must add ("gg" tide-jump-to-definition) to the webList because spacemacs-jump-handlers-web-mode does not exist.

### DIFF
--- a/layers/+email/mu4e/README.org
+++ b/layers/+email/mu4e/README.org
@@ -45,23 +45,23 @@ existing =dotspacemacs-configuration-layers= list in this file.
 
 | Keybinding | Command    |
 |------------+------------|
-| SPC a M    | Start mu4e |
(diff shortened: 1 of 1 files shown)


LemmingAvalanche:
That's a nice PR comment. Why not add it to the commit message as well?

george-miller:
Accidentally forced push to my branch when trying to update it to develop, lost my changes for now. I have the changes locally in another computer which I will get access to after vacation.  I will update the commit message and redo the PR once I get my changes back!

Issue #6583, opened 2016-07-14
Delete packages on startup rather than before restarting 


nixmaniack:
Deleting packages when doing "Update Packages" in running instance of emacs may render emacs unusable. 

Reference: https://github.com/syl20bnr/spacemacs/issues/4127#issuecomment-226769130

Created this issue for tracking with clear issue description so that someone can try to fix it.

cpaulik:
There were problems with this approach but I can not find the discussion at the moment. Maybe it is enough to make it even clearer that emacs has to be restarted. Or maybe change the Button to `Update and Restart`?

nixmaniack:
After package update it does mention that you have to restart emacs. But yes, making it mandatory to restart after updating packages might be a better option

CestDiego:
I think this is the way to go, I end up restarting anyway, there should be a direct action to restart when updating,

nixmaniack:
There's just one possibility of `restart` failing in case the `restart-emacs` package is updated in the process because it gets deleted.

syl20bnr:
I like this approach, but we have to do it manually, i.e. _before_ initializing `package.el`.

Stebalien:
> I like this approach

I assume "this" is "delete on start"?

---

Regardless, would you consider the following instead of simply deleting? It would be nice to have atomic updates that are automatically rolled back on failure.

1. When an update is requested, record the intended update (which packages are being updated) in a file (`update.el`).
2. Early after restarting (before initializing anything)...
3. If `update.el` exists:
    1. Load `update.el`.
    2. Move it to `partial-update.el`
    3. Move the to-be-updated packages to `~/.emacs.d/elpa-backup/`.
    4. Continue loading and perform the update.
    6. Delete `partial-update.el`.
    5. Delete `~/.emacs.d/elpa-backup/`.
4. If `partial-update.el` exists:
    1. Load it.
    2. Delete the packages listed in it from `~/.emacs.d/elpa` (roll back the update).
    3. Move all packages from `~/.emacs.d/elpa-backup/` to `~/.emacs.d/elpa/`.
    4. Delete `partial-update.el`.
    5. Delete `~/.emacs.d/elpa-backup/`.
    6. Load as usual.
    7. Inform the user that an update was aborted.
4. Otherwise, delete `~/.emacs.d/elpa-backup/` if it exists.

JonasDralle:
`restart-emacs` updated successfully. Now how to I restart 😆

[...]

PR #7760, opened 2016-11-16
Don't suggest `SPC q r` if restart-emacs is missing

kini:
When upgrading packages, Spacemacs backs up the old versions of the
packages and then tells the user to restart Emacs using `SPC q r` so
that the new version will be auto-installed.  But `SPC q r` doesn't
work if restart-emacs is one of the packages which is being upgraded,
and hence isn't currently installed.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -1668,8 +1668,10 @@ If called with a prefix argument ALWAYS-UPDATE, assume yes to update."
         (spacemacs-buffer/append
          (format "\n--> %s package(s) to be updated.\n" upgraded-count))
         (spacemacs-buffer/append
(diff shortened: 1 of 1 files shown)


TheBB:
IMO the best solution to this problem is https://github.com/syl20bnr/spacemacs/issues/6583

kini:
Indeed, that is a better solution, but I don't know how to implement that properly so I made this smaller change instead :)

syl20bnr:
Thank you !

I agree with @TheBB that we should implement #6583

duianto:
I don't know if something has changed since this was implemented, but `restart-emacs` was just updated and `(SPC q r)` is still visible:
>Emacs has to be restarted to actually install the new version of the packages (SPC q r).

Issue #7766, opened 2016-11-17
org-present and normal/evilified state


jupl:
#### Description :octocat:
After exiting `org-present`, when I press `ESC` while in normal state I enter evilified state.

#### Reproduction guide :beetle:
- Start Emacs (with `org` layer enabled)
- Open an `org` file
- Press `ESC` to assure still in normal state
- Start presenting with the `org-present` command
- Press `q` to exit presentation mode
- Press `ESC`

*Observed behaviour:* :eyes: :broken_heart:
Now in evilified state.

*Expected behaviour:* :heart: :smile:
Remain in normal state.

#### System Info :computer:
- OS: darwin
- Emacs: 25.1.3
- Spacemacs: 0.200.5
- Spacemacs branch: develop (rev. fea919c)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(clojure colors docker editorconfig emacs-lisp html javascript markdown org osx react restclient shell shell-scripts syntax-checking typescript vimscript yaml
         (auto-completion :variables auto-completion-return-key-behavior nil auto-completion-tab-key-behavior 'complete auto-completion-enable-help-tooltip t)
         (git :variables git-magit-status-fullscreen t)
         (version-control :variables version-control-diff-tool
                          (if myspacemacs--gui 'diff-hl 'git-gutter)
                          version-control-diff-side 'left))
```
- System configuration features: NOTIFY ACL LIBXML2 ZLIB TOOLKIT_SCROLL_BARS

TheBB:
I get evilified, not emacs…

jupl:
Ah yes. I had the wrong name. I don't use that state normally so I didn't know what the proper name. However, it does throw me way off because my keybinds change. And even if I enter `evil-normal-state` when I press `ESC` again i go back to evilified.

TheBB:
https://github.com/syl20bnr/spacemacs/pull/7767

smile13241324:
@jupl I have merged a fix for your issue #13088, would you mind checking whether this solved your issue and close the ticket if it does?

zane:
@smile13241324 I was having the same issue. Looks fixed to me. Thank you so much!

PR #7767, opened 2016-11-17
[WIP] Restore normal state map on evilified exit

TheBB:
This seems to fix https://github.com/syl20bnr/spacemacs/issues/7766 but I'm not sure if has any other repercussions.

### DIFF
--- a/layers/+distributions/spacemacs-base/local/evil-evilified-state/evil-evilified-state.el
+++ b/layers/+distributions/spacemacs-base/local/evil-evilified-state/evil-evilified-state.el
@@ -142,6 +142,8 @@ Needed to bypass keymaps set as text properties."
 
 (add-hook 'evil-evilified-state-entry-hook
           'evilified-state--evilified-state-on-entry)
+(add-hook 'evil-evilified-state-exit-hook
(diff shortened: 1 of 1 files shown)


aroig:
This commit breaks at least mu4e. When quitting a mu4e buffer it produces an error

    Wrong type argument keymapp nil

After that mu4e is unusable until I restart emacs.

aroig:
Ok, some more details about the problem with mu4e.

When I start mu4e, the `*mu4e-main*` buffer has `evilified-state--normal-state-map` bound locally to a non-nil value (that looks ok). The global value is `nil`.

Then I do a search, or something that opens a mu4e buffer. I close that buffer (with `q`), and things break. I need to restart emacs to use mu4e again.

I have put some `message` in the `evilified-state--restore-normal-state-map` hook. At the instant it break things, it is executed with current buffer `*mu4e-main*` but the variable `evilified-state--normal-state-map` value is nil. I assume this is the cause of the error and mu4e breakage.

I do not understand what is going on. Any ideas?

syl20bnr:
I reverted the commit, we need to find another fix to this issue. We can use this PR to work on it.

PR #7774, opened 2016-11-17
Use "non-nil" in .spacemacs.template

mineo:
Replace the mixture of "non nil" and "non-nil" with the latter. It seems to be
the form used by Emacs itself (for example in the docstring of `when`).

There's probably more of the same in other parts of spacemacs, but .spacemacs is where I noticed it.


syl20bnr:
Indeed, thank you for the correction, will try to respect it in the future !

PR #7789, opened 2016-11-20
sql: support for keywords auto capitalization

kepi:
SQL, by convention, uses upper-case keywords, although lower-case works just as
well. As humans, the separation between upper-case and lower-case helps scan and
parse the code much more quickly.

This PR add configurable support for minor mode `sqlup-mode` to SQL layer.

I would be glad to receive comments how to improve PR if there are any troubles with it.
I tried to follow conventions and check other layers about common ways. But this is
not only my first contribution to Spacemacs but I'm also really not the best in elisp :)

Design decisions:
----------------------
- sqlup is disabled by default. IMHO it would be best to enable it by default but I was afraid it isn't good idea to change default behavior for current users. **If you want it on by default** please [upvote this comment](https://github.com/syl20bnr/spacemacs/pull/7789#issuecomment-262094062).
- If you enable sqlup it will enable for both sql-mode and sql-interactive mode which is little different from default sqlup behavior. IMHO it is more comfortable and expected way from spacemacs layer.
- sqlup has empty blacklist by default. IMHO `name` is so common column name (it is also mentioned in sqlup documentation) that it deserves default blacklisting.
- I wasn't sure if it is ok to create another layer config variable for blacklist as it can be configured through customizable variable `sqlup-blacklist`. But it made more sense to me to define it when enabling capitalization. I guess that people who want sql layer fast will appreciate this.
- I added keybinding to code formatting section but let me know if it is better to move it to different part (I thought about higlighting before).

### DIFF
--- a/layers/+lang/sql/packages.el
+++ b/layers/+lang/sql/packages.el
@@ -9,7 +9,12 @@
 ;;
 ;;; License: GPLv3
 
-(setq sql-packages '(sql sql-indent))
+(setq sql-packages
+      '(
+        sql
+        sql-indent
+        (sqlup-mode :toggle sql-capitalize-keywords)
+        ))
 
 (defun sql/init-sql ()
   (use-package sql
@@ -117,3 +122,21 @@
 (defun sql/init-sql-indent ()
   (use-package sql-indent
     :defer t))
+
+(defun sql/init-sqlup-mode ()
+  (use-package sqlup-mode
+    :defer t
+    :init
+    (progn
+      (add-hook 'sql-mode-hook 'sqlup-mode)
+      (unless sql-capitalize-keywords-disable-interactive
+        (add-hook 'sql-interactive-mode-hook 'sqlup-mode))
+      (spacemacs/set-leader-keys-for-major-mode 'sql-mode
+        "=c" 'sqlup-capitalize-keywords-in-region))
+    :config
+    (progn
+      (spacemacs|diminish sqlup-mode)
+      (when sql-capitalize-keywords-blacklist
+        (if (listp sql-capitalize-keywords-blacklist)
+            (setq sqlup-blacklist (append sqlup-blacklist sql-capitalize-keywords-blacklist))
+          (add-to-list 'sqlup-blacklist sql-capitalize-keywords-blacklist))))))
--- a/layers/+lang/sql/README.org
+++ b/layers/+lang/sql/README.org
@@ -5,14 +5,17 @@
 * Table of Contents                                         :TOC_4_gh:noexport:
  - [[#description][Description]]
  - [[#install][Install]]
+   - [[#sql-keywords-capitalization][SQL Keywords Capitalization]]
+     - [[#sql-interactive-mode][SQL Interactive Mode]]
+     - [[#blacklisting-keywords][Blacklisting keywords]]
  - [[#key-bindings][Key bindings]]
    - [[#highlighting][Highlighting]]
    - [[#inferior-process-interactions-sqli][Inferior Process Interactions (SQLi)]]
      - [[#send-sql-queries-to-sqli][Send SQL queries to SQLi:]]
    - [[#sqli-buffer][SQLi buffer]]
+   - [[#code-formating][Code Formating]]
 
 * Description
-
 This layer adds key bindings and configuration for =sql-mode=, which manages
 interactive SQL buffers and highlights a wide range of SQL dialects.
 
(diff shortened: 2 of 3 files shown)


layers/+lang/sql/packages.el
```diff
@@ -117,3 +122,19 @@
...
   (use-package sql-indent
     :defer t))
+
+(defun sql/init-sqlup-mode ()
+  (use-package sqlup-mode
+    :defer t
+    :init
+    (progn
+      (add-hook 'sql-mode-hook 'sqlup-mode)
+      (when (and sql-capitalize-keywords (not sql-capitalize-keywords-disable-interactive))
```
nixmaniack: Since you're using `:toggle sql-capitalize-keywords`, first test is redundant. You can reduce the expression to `(unless sql-capitalize-keywords-disable-interactive ...)`

nixmaniack:
> I wasn't sure if it is ok to create another layer config variable for blacklist as it can be configured through customizable variable sqlup-blacklist. But it made more sense to me to define it when enabling capitalization. I guess that people who want sql layer fast will appreciate this.

IMO, if the package has already defined it, I don't see a point in redefining as layer variable. Mentioning in the doc "how" should be enough. 
Users can modify the variable when adding the layer in `dotspacemacs-configuration-layers`.
```elisp
              '((sql :variables
                     sql-capitalize-keywords t
                     sqlup-blacklist '("name")))))

```

> sqlup is disabled by default. IMHO it would be best to enable it by default

I would 👍 for this as sane default!

layers/+lang/sql/packages.el
kepi: Good point. I added toggle at the end of modifications and forgot about this. Thanks.

kepi:
> IMO, if the package has already defined it, I don't see a point in redefining as layer variable. Mentioning in the doc "how" should be enough.
> Users can modify the variable when adding the layer in dotspacemacs-configuration-layers.

I forgot to mention maybe more important decision here. I decided to move it to spacemacs variable at the end because I think that there may be different implementation for capitalization than `sql-mode` later.  There are at least 3 different solutions as of today. I tought that with this spacemacs variable we can switch capitalization "backends"/solutions without user noticing it and preserve same behavior.

Anyway if you think this isn't reason enough I have no problem with removing this part.

> I would 👍 for this as sane default! 

Should I change it and try to get more thumbups? I'm not sure how to decide this.

nixmaniack:
> I also added diminish for mode (T letter as it is most used for capitalization).

It's already taken and we also have a conflict! 

```
$ ag Ⓣ
layers/+emacs/typography/packages.el
30:      (spacemacs|diminish typo-mode " Ⓣ" " T"))

layers/+lang/html/packages.el
167:      (spacemacs|diminish tagedit-mode " Ⓣ" " T"))))
```

nixmaniack:
> Should I change it and try to get more thumbups? I'm not sure how to decide this.

A comment asking for thumbs-up for making it default should help but I doubt how many will read it, so may be edit your first post pointing to that comment - asking users to upvote.

kepi:
> > I also added diminish for mode (T letter as it is most used for capitalization).
> It's already taken and we also have a conflict!

I know, because of that conflict I thought its ok :) Changed to "U", didn't find that one. If you have better idea, let me know.

kepi:
Please upvote this comment if you think that SQL capitalization should be on by default in SQL layer.

nixmaniack:
Regarding modeline character, the idea in Spacemacs is that they serve the purpose of corresponding "toggle key" e.g. for `whitespace-cleanup` it is "W" and toggle key is `SPC t W`, for `auto-completion` it's `a` etc. 
With sqlup mode I don't think a toggle is required and occupying that precious little space and character in modeline doesn't have much benefit - the toggle layer variable is enough for me, IMO.

kepi:
> I don't think a toggle is required and occupying that precious little space and character in modeline

I agree. I didn't realize, that it is possible to diminish it all together. Updated.

> the toggle layer variable is enough for me, IMO.

I'll currently not add this as there is problem with toggle `sqlup-mode` currently, see Trevoke/sqlup-mode.el#63

Dietr1ch:
Capitalization is the poor man's highlighting. You are using emacs, so you shouldn't need nasty casing changes to distinguish between keywords, column names, strings, and so on.

I think however that this deserves a layer variable.

Trevoke:
Happy new year everyone!

@Dietr1ch I admit I'm biased because I'm the author's plugin -- but I've also been writing SQL for a while. There's nothing wrong with having capitalization *and* syntax highlighting.

Most of the SQL code you come across online is capitalized; at the very least, let this be, just like syntax highlighting, a matter of cosmetic preference :)

@kepi How's progress on this?

kepi:
@Trevoke slow unfortunately. I should be able to find some time this week but no promises.

TheBB:

I'm actually tempted to make this on by default, but I didn't do that… yet.

I did change so that the blacklist is actually a list though, and not optionally a string. Type consistency is nice.

Issue #7312, opened 2016-10-07
ocaml layer without opam


bschommer:
It would be nice if the ocaml layer could be used without opam. A quick check shows that the problem is currently the opam-init. I would suggest adding some variables as alternatives for specifying other paths for the location of the merlin, etc.

TheBB:
I'm not sure what you mean here. What problem? Is there some reason you can't just exclude the packages causing problems? That's the basic mechanism in Spacemacs for using only parts of a layer.

TheBB:
I mean, it looks like tuareg and utop needs opam, so if you exclude tuareg and utop, won't that work?

bschommer:
Sorry that for the short problem report. Actually the ocaml layer works however the auto-completion offered by merlin does not.

In the funcs.el file the opam-init function takes care that the directory containing merlin.el, etc. is added to the load-path by querying opam where it installed merlin. The merlin-mode itself locates the needed binary by either querying opam itself or letting the user set it via a variable. 

My simple workaround for this is to replace opam-init with:

    (defun spacemacs//init-ocaml-opam ()
      ;; (if (executable-find "opam")
      ;;     (let* ((output (shell-command-to-string
      ;;                     "opam config var share 2> /dev/null"))
      ;;            (share (when (< 0 (length output))
      ;;                     (substring output 0 -1))))
      ;;         (when share
      ;;           (setq opam-share share
      ;;                 opam-load-path (concat share "/emacs/site-lisp")))
      ;;         (add-to-list 'load-path opam-load-path))
      ;;   (spacemacs-buffer/warning
      ;;    (concat "Cannot find \"opam\" executable. "
      ;;            "The ocaml layer won't work properly.")))
      (setq merlin-command "<merlin-prefix>/bin/ocamlmerlin")
      (add-to-list 'load-path "<merlin-prefix>/share/emacs/site-lisp")
      )

where merlin-prefix is my merlin installation path.

TheBB:
Okay. Is there a reason you need to set `merlin-command` though? Spacemacs never sets it anywhere.

TheBB:
Let me clarify. The code you have commented out

* sets `opam-share`
* sets `opam-load-path`
* adds to `load-path`

While the code you have added

* sets `merlin-command`
* adds to `load-path`

Is there a reason for the discrepancies?

bschommer:
[...]

PR #7793, opened 2016-11-21
Ocaml layer init without opam.

bschommer:
Currently the ocaml layer requires opam in order to work. However
opam is only really required for an easy installation of the
required ocaml binaries. The only place where opam is used
instead is to add the merlin files of the merlin installed
via opam instead of the merlin files from melp.

This commit introduces a fallback if opam is not installed but
ocamlmerlin is installed, since the melpa package of merlin
is already installed through packages.

This should fix issue #7312.


bschommer:
Any problem with this PR? Is there anything missing?

TheBB:

PR #7802, opened 2016-11-22
[scala] Make ensime use stable version as unstable is for ensime devs

d1egoaz:
Make ensime use stable version as unstable is for ensime devs.

Ensime unstable is not intended for general use and will cause problems.
see https://github.com/ensime/ensime-emacs/issues/553

### DIFF
--- a/layers/+lang/scala/README.org
+++ b/layers/+lang/scala/README.org
@@ -38,6 +38,14 @@ To use this configuration layer, add it to your =~/.spacemacs=. You will need to
 add =scala= to the existing =dotspacemacs-configuration-layers= list in this
 file.
 
+Then, you'll need to modify your =~/.spacemacs= to use the recommended
+Ensime version (Stable). Please add the following lines to =dotspacemacs/user-init=:
+#+BEGIN_SRC emacs-lisp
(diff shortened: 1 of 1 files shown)


layers/+lang/scala/README.org
```diff
@@ -61,8 +69,8 @@ project's =project/build.properties=. For example,
    ensimeConfig= and =sbt ensimeConfigProject= at the shell.
 2. Run =M-x ensime= within Emacs to start an ENSIME session.
 
-Each Scala project uses a dedicated ENSIME session, so you only need to run =M-x
-ensime= once per project. Any Scala files you create or visit within the project
+Each Scala project uses a dedicated ENSIME session, so you only need to
+run =SPC SPC ensime= once per project. Any Scala files you create or visit within the project
```
LemmingAvalanche: Should be:

```
... run ~SPC SPC ensime~ ...
```

Instead, since it is a keybinding.

diego-alvarez-hs:
up

diego-alvarez-hs:
bump, @TheBB

TheBB:

I put it in the java layer readme too, since ensime can be found there as well.

PR #7803, opened 2016-11-22
Fix python-enable-yapf-format-on-save

hidaris:
It seems the original config doesn't work when the variable is true.


hidaris:
Oops..

PR #7807, opened 2016-11-22
Remove unnecessary parameter to cider-test-run-project-tests

mahinshaw:
The nil parameter was causing [`cider-test-run-project-tests`](https://github.com/clojure-emacs/cider/blob/master/cider-test.el#L634) to fail (not run).  Furthermore, the function us a 0-arity function.


mahinshaw:
This is fixed by 15e8670

d12frosted:
I am sorry, didn't see your PR. Thanks for noticing that I cherry-picked similar PR.

mahinshaw:
No problem

PR #7808, opened 2016-11-22
Use proper markup to preserve meaning of code

LemmingAvalanche:
Mark up code that is mentioned in the documentation of some readmes.

In the case of “long options”, like for example `--with-emacs`, this is not just
cosmetic.  On GitHub, Org files are apparently rendered in such a way that
strings like `--` in non-verbatim text (i.e. not verbatim-quoted nor
code-quoted) is transformed to `–` (EN DASH U+2013).  So the string:

    … --with-emacs option:

Will show up like this:

    … –with-emacs option:

Also mark up nearby not-marked-up code mentions.  But this pattern was what was
searched for, so this mostly changes the abovementioned kind of thing.


TheBB:

PR #7810, opened 2016-11-22
which-key: Update usage of replacement alists

justbur:
See https://github.com/justbur/emacs-which-key/commit/19186917eb2958c11a297b03b7963e02053959ec


bmag:

PR #7815, opened 2016-11-23
Remove sbt-hydra rename wrapper

jdnavarro:
It has been already renamed as expected upstream.


d12frosted:

PR #7817, opened 2016-11-23
Count of killed matching buffers

Bost:
Enables feedback messsages indication like e.g.:
    "5 buffer(s) killed"

Please consider involving the original author @gempesaw to the discussion.

### DIFF
--- a/layers/+distributions/spacemacs-base/funcs.el
+++ b/layers/+distributions/spacemacs-base/funcs.el
@@ -539,19 +539,22 @@ If the universal prefix argument is used then will the windows too."
       (insert "\n")
       (setq count (1- count)))))
 
-;; from https://github.com/gempesaw/dotemacs/blob/emacs/dg-defun.el
+;; from https://github.com/gempesaw/dotemacs/blob/emacs/dg-elisp/dg-defun.el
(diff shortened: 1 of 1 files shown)


deb0ch:
Does it actually display the number of killed buffers with your PR or is extra code needed ?

Bost:
No, it just **enables** anything like that. For example in my own .spacemacs I have:
``` 
(defun kill-all-magit-buffers ()
    (interactive)
    (let* ((counts (mapcar
                    'spacemacs/my-kill-matching-buffers-rudely
                    '("\*magit: .*" "\*magit-.*"))))
      (message (format "%d buffer(s) killed" (reduce #'+ counts)))))
```

where `spacemacs/my-kill-matching-buffers-rudely` is the same function as in this PR.

deb0ch:
(would be so awesome if you implemented it :smiley_cat:)

Bost:
Originally I wanted to use `defadvice` or `advice-add`. But I found https://www.emacswiki.org/emacs/AdviceVsHooks so what I did in the end is just a split of original `spacemacs/kill-matching-buffers-rudely` into two functions - one for the functionality itself and the other for the message.

syl20bnr:
Thank you ! 👍 

PR #7820, opened 2016-11-23
Fix code markup

LemmingAvalanche:
- One preparation commit which adds a newline, to avoid unrelated changes in the main commit
- Fix code markup in readmes


TheBB:

PR #7821, opened 2016-11-23
Add verbatim markup to paths

LemmingAvalanche:
By convention, filepaths are marked up with `=`.  Also add this markup to the
string `ViewLogMode`.


bmag:

PR #7829, opened 2016-11-24
Use `RET` instead of `return` for return key

LemmingAvalanche:
It is an Emacs-wide convention to use `RET` when referring to the act of
pressing the return key.  Spacemacs seems to follow that convention as well.  So
fix some Org files that were deviating from this convention.

In particular, replace usages of `<return>` and `return` with `RET`.

### DIFF
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -2078,8 +2078,8 @@ Basic =ido= operations can be done with ~Ctrl~ key:
 
 | Key Binding        | Description                                       |
 |--------------------+---------------------------------------------------|
-| ~C-<return>~       | open a =dired buffer=                             |
-| ~M-<return>~       | open a =dired buffer= in terminal                 |
+| ~C-RET~            | open a =dired buffer=                             |
+| ~M-RET~            | select the buffer or file named by the prompt     |
 | ~C-d~              | delete selected file (ask for confirmation)       |
 | ~C-h~              | go to parent directory                            |
 | ~C-j~              | select next file or directory                     |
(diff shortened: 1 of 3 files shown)


doc/DOCUMENTATION.org
```diff
@@ -2078,8 +2078,8 @@ Basic =ido= operations can be done with ~Ctrl~ key:
 
 | Key Binding        | Description                                       |
 |--------------------+---------------------------------------------------|
-| ~C-<return>~       | open a =dired buffer=                             |
-| ~M-<return>~       | open a =dired buffer= in terminal                 |
+| ~C-RET~            | open a =dired buffer=                             |
+| ~M-RET~            | open a =dired buffer= in terminal                 |
```
bmag: I'm not sure this is bound anymore. Can you check and change (or remove) the description as necessary?

LemmingAvalanche: You’re right. Both of these are undefined when I’m a `dired` buffer or trying to use them in the `*scratch*` buffer. I’m not able to find functions that refer to something like “open dired buffer in terminal” (or similar), so I’ll just delete these lines. Will amend.

bmag: Wait a sec. These are supposed to be defined in Ido (`ido-mode`), so you need to run `M-x ido-find-file` and from that prompt test the key bindings. When I checked it `C-RET` actually did work, I was only talking about `M-RET`.

LemmingAvalanche: Gotcha. Pressing `M-RET` while in `ido-find-file` executes `ido-select-text`:

```
M-RET (translated from <M-return>) runs the command ido-select-text, which is an
interactive compiled Lisp function in `ido.el'.

It is bound to <C-return>, M-RET.

(ido-select-text)

Select the buffer or file named by the prompt.
If no buffer or file exactly matching the prompt exists, maybe create a new one.
```

LemmingAvalanche:
@bmag are my changes OK?  Do you have any other suggestions?

syl20bnr:
Thank you ! 👍 

Issue #6251, opened 2016-06-07
Pasting in evil mode doesn't work (in develop)


krotton:
#### Description :octocat:
Pasting in evil mode (using `p` or `P`) doesn't work, failing with: `command-execute: Wrong type argument: commandp, spacemacs/paste-transient-state/evil-paste-after`. If I run `:evil-paste-after`, it works.

Workaround: using `⌘   v`

#### Reproduction guide :beetle:
- Start Emacs
- Open any buffer and insert a line of text
- Yank (`yy`) or delete (`dd`) the line of text
- Try pasting it using `p` or `P`

*Observed behaviour:* :eyes:
No pasted text and a message in the bottom bar saying: `command-execute: Wrong type argument: commandp, spacemacs/paste-transient-state/evil-paste-after`.

*Expected behaviour:* :broken_heart:
The copied/deleted line pasted below or above the current one.

#### System Info :computer:
- OS: darwin
- Emacs: 25.1.50.1
- Spacemacs: 0.105.20
- Spacemacs branch: develop (rev. ea77a5d)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers (note - the problem occurs even if all are disabled):
```elisp
(csv markdown emacs-lisp osx git org
     (scala :variables scala-auto-insert-asterisk-in-comments t)
     syntax-checking graphviz plantuml ktt-groovy)
```

#### Backtrace :paw_prints:
```
command-execute: Wrong type argument: commandp, spacemacs/paste-transient-state/evil-paste-after
command-execute: Wrong type argument: commandp, spacemacs/paste-transient-state/evil-paste-before
```

CharlesHD:
I'm behind the same commit, on gnu/linux . Cannot reproduce (aka it works normally for me), I have not the same layers.
Maybe you could try with a fresh install of spacemacs ?

krotton:
I'll try and report back.

krotton:
Indeed, reinstallation has fixed it. I don't understand why, but I think that closes the issue ;).

alexanderkjeldaas:
I have the same issue.  Switched from master to devel.

d12frosted:
@alexanderkjeldaas develop is stable enough to use on a daily basis 😸

toupeira:
Please reopen, I keep running into this every now and then, I think it started about 2 weeks ago. With latest develop I can reproduce it as follows:

- kill all running Emacs instances
[...]

PR #7831, opened 2016-11-24
Disable paste transient state by default

toupeira:
To match the default configuration template which [also disables it](https://github.com/syl20bnr/spacemacs/blob/a1516a600399715e86926f5ce3c07d62fd3730ec/core/templates/.spacemacs.template#L208).

See https://github.com/syl20bnr/spacemacs/issues/6251#issuecomment-262289770


bmag:
Thank you

PR #7841, opened 2016-11-25
Update LAYERS.org

jankei:
path fix


syl20bnr:
Thank you for the PR, unfortunately `LAYERS.org` is a generated file (I guess we should add a header to mention this) so merging this PR will be overwritten by the next documentation export. I close this PR as it is against a generated file, sorry for not mentioning this in the file directly.

PR #6435, opened 2016-06-26
Media layer with EMMS music manager

alejandroerickson:
Configured a fork of EMMS with extra features. Keys and leaders
documented, but note `SPC a SPC` to play/pause music. EMMS extra
features are being integrated into EMMS, so layer uses my fork in the
mean time. Also includes instructions and code for configuring OS X
media keys for EMMS.  Media layer README.org pasted below has full details, keys, etc.
# +TITLE: media layer
# +HTML_HEAD_EXTRA: <link rel="stylesheet" type="text/css" href="../css/readtheorg.css" />
# +CAPTION: logo
# The maximum height of the logo should be 200 pixels.

[[img/media.png]]
- Table of Contents                                        :TOC_4_org:noexport:
  - [[Description][Description]]
  - [[Install][Install]]
    - [[EMMS configuration ][EMMS configuration ]]
  - [[Configuring media keys][Configuring media keys]]
    - [[Setting the media keys on Mac OS X][Setting the media keys on Mac OS X]]
  - [[Key bindings][Key bindings]]
    - [[Evilified keybindings][Evilified keybindings]]
      - [[Interactive Playlists][Interactive Playlists]]
    - [[Mark Playlists ][Mark Playlists ]]
    - [[EMMS Browser ][EMMS Browser ]]
    - [[Track (tag) Information Editor ][Track (tag) Information Editor ]]
  - [[Issues][Issues]]
- Description
  This layer installs and configures media library management software EMMS.

https://www.gnu.org/software/emms/
- Install
  To use the media layer add it to your =~/.spacemacs=
# +begin_src emacs-lisp

  (setq-default dotspacemacs-configuration-layers '(media))
# +end_src

Optionally, if you have a collection of media files in a single directory tree,
you can set the root directory of that tree as EMMS's default.
# +begin_src emacs-lisp

```
(setq emms-source-file-default-directory "/path/to/music/")
```
# +end_src

When you first open the EMMS browser it will prompt you to add music. It then
builds a cache of track info, defaulting to the directory ~/.emacs.d/emms.
Before doing so install a track info reader to properly extract track
information, otherwise you will encounter errors and the track info will not be
read. We recommend you use
[TagLib](https://www.gnu.org/software/emms/manual/#Using-TagLib), the one
included with EMMS, by compiling and installing it to your path.
# +begin_src shell

$ cd
~/.emacs.d/elpa/emms-<version>/cd-here-and-make-emms-print-metadata-and-put-in-path/
$ make emms-print-metadata
[...]

PR #7844, opened 2016-11-26
Add emms dictionary and chinese-pyim dcache

LdBeth:
chinese-pyim will create a folder to save the frequently used words in .emac.d


bmag:
It's better to move those files/folders into `dotspacemacs-cache-directory` (`~/.emacs.d/.cache/`) than adding them to gitignore. As for EMMS, there's another PR: #6435

PR #7848, opened 2016-11-27
engine-mode: Fix url to search engine github repository

wigust:
Just a little fix of url to search engine github repository


TheBB:
Thank you. Unfortunately this isn't needed any more, the link was changed to point to the repo instead of a file in the repo.

PR #7849, opened 2016-11-27
ivy: add and bind spacemacs/counsel-find-library to "fel"

nixmaniack:


nixmaniack:
Closing in favor of built-in function being added in #8690.

PR #7865, opened 2016-11-28
Adds tide keybindings to tsx web mode

george-miller:
mode so we can properly edit/format the html that's inside. However, with this
the current function/variable. You also lose the ", g u" keybinding which
gives you all references of current function/variable. This PR adds
all the tide bindings that are in typescript mode into web mode while using the typescript layer.

NOTE:
I must add ("gg" tide-jump-to-definition) to the webList because spacemacs-jump-handlers-web-mode does not exist.

### DIFF
--- a/layers/+lang/typescript/packages.el
+++ b/layers/+lang/typescript/packages.el
@@ -54,13 +54,20 @@
         (interactive)
         (tide-jump-to-definition t))
 
-      (spacemacs/set-leader-keys-for-major-mode 'typescript-mode
-        "gb" 'tide-jump-back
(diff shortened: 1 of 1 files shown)


george-miller:
@LemmingAvalanche  Please rereview, this is a redo of https://github.com/syl20bnr/spacemacs/pull/7759

Thanks

LemmingAvalanche:
Looks good.  My comment was only on the commit message.  I can’t really comment on anything beyond that.

syl20bnr:
Thank you ! 👍 

Issue #7743, opened 2016-11-15
Command log mode does not start correctly


aeberts:
#### Description :octocat:

Command log mode does not start correctly

#### Reproduction guide :beetle:
- Start Emacs
- enable command log mode in user init
- restart emacs
- from normal mode type: ```SPC a L```

*Observed behaviour:* :eyes: :broken_heart:

-> new emacs buffers are created and spacemacs becomes unusable (see screenshot)

*Expected behaviour:* :heart: :smile:

-> command mode is enabled. 

#### System Info :computer:
- OS: darwin
- Emacs: 25.1.1
- Spacemacs: 0.200.5
- Spacemacs branch: master (rev. 664ba6a)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(osx
 (auto-completion :variables auto-completion-enable-snippets-in-popup t auto-completion-enable-help-tooltip t auto-completion-enable-sort-by-usage t)
 better-defaults emacs-lisp
 (org :variables org-enable-github-support t)
 git github markdown clojure python html erc
 (shell :variables shell-default-height 30 shell-default-position 'bottom)
 spell-checking syntax-checking version-control themes-megapack command-log)
```
- System configuration features: JPEG RSVG IMAGEMAGICK NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS

#### Backtrace :paw_prints:
```
<<BACKTRACE IF RELEVANT>>
```
![screen shot 2016-11-15 at 9 29 51 am](https://cloud.githubusercontent.com/assets/571576/20316528/126cfa4a-ab17-11e6-93bd-40f4ccd5794f.png)

TheBB:
I can't reproduce this. Please try from a base installation and provide better reproduction steps.

aeberts:
Just tried with a completely clean install.

```SPC a L``` appears to work fine in an *.el or *.clj buffer.

Error occurs when enabling command-log from the ```*spacemacs*``` buffer with ```SPC a L```

## Backtrace

```
Debugger entered--Lisp error: (wrong-type-argument wholenump -6)
  make-string(-6 32)
  (insert (make-string (floor (/ (- maxcol badge-size) 2)) 32))
  (progn (insert (make-string (floor (/ (- maxcol badge-size) 2)) 32)) (insert-image badge))
[...]

PR #7874, opened 2016-11-29
home-buffer: fix error on footer insertion when window is narrow

deb0ch:
Related to #7743.

There was an error thrown when the home buffer got too narrow, due to the calculation of the centering giving a negative number to `make-string`.

This bug has likely been there for a long time.


d12frosted:

PR #7894, opened 2016-12-01
Add evil keybinding "i a" for org-attach

smile13241324:
Add evil keybinding for org-attach to avoid emacs style `C-c C-a` to add attachments to org files.


bmag:

PR #7924, opened 2016-12-07
Add info about SSH key to github layer README

sid-kap:
Github clone repo failed (with an unhelpful error) because I didn't have an SSH key associated. (This is because magit calls `git clone` using the `git://` protocol, rather than http or https.) Creating an SSH key and adding it to my GitHub account fixed the problem.


sid-kap:
Can someone take a look at this?

syl20bnr:
Thank you !

PR #7925, opened 2016-12-07
Major colors layer enhancements and bug fixes

SteveJobzniak:
These are some pretty major enhancements which make the "colors" layer suck a _lot_ less. Please consider merging these quickly into the next release. Because the current layer as-provided in Spacemacs is _full_ of hardcoded nonsense which makes it nearly useless (compared to just manually adding a package and having full control).

**Fixed and added:**
- Keybinding documentation has been corrected to reflect earlier code changes that moved color toggles to `<SPC tC>`.
- The messy documentation has been rewritten to fix bugs, straighten out terminology, and to make people's heads spin less.
- Added the ability to set the default saturation and lightness of rainbow-identifiers for any themes that haven't been manually overriden, instead of being forced to live with what the earliest author of this layer liked.
- Improved resetting of saturation and lightness in the transient-state, so that it now resets to the user's own defaults.

**Future wishlist, because I don't have more time:**
- Enhance the transient-state resetting even further, so that it resets to the default that would be used for that theme.
- Add the ability to set theme colors via a variable in :variables. Although, this is pretty much a non-issue after my documentation fixes that describe how to currently achieve it.

### DIFF
--- a/layers/+distributions/spacemacs-bootstrap/packages.el
+++ b/layers/+distributions/spacemacs-bootstrap/packages.el
@@ -301,8 +301,8 @@
            ("spacemacs/toggle-hybrid-mode" . "hybrid (hybrid-mode)")
            ("spacemacs/toggle-holy-mode" . "emacs (holy-mode)")
            ("evil-lisp-state-\\(.+\\)" . "\\1")
-           ("\\(.+\\)-transient-state/\\(.+\\)" . "\\2")
-           ("\\(.+\\)-transient-state/body" . "\\1-transient-state"))))
+           ("spacemacs/\\(.+\\)-transient-state/\\(.+\\)" . "\\2")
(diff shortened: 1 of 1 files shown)


SteveJobzniak:
Knowing that it sometimes takes months for a merge, I've created a private version of this exact code and renamed it to "paintvars" to avoid the "Duplicate layer detected" error. So just in case anyone wants to use it right away together with their normal non-development Spacemacs, you can just do the following:

```bash
git clone https://github.com/SteveJobzniak/spacemacs-private-paintvars.git ~/.emacs.d/private/paintvars
```

Documentation, with the renamed variables (every setting is literally the same as "colors" but with the word replaced with "paintvars"):
https://github.com/SteveJobzniak/spacemacs-private-paintvars/blob/master/README.org

PR #7929, opened 2016-12-08
Use raw prefix argument instead numeric prefix argument

lululau:
With numeric prefix argument, the `arg` will always be true, and `spacemacs/sudo-edit` will always prompt to select a file


bmag:
> Thank you for contributing to Spacemacs!
>
> Before you submit this pull request, please ensure it is against the develop branch and not master.
>
> This message should be replaced with a description of your change.
>
> Thank you <3

Please replace the message with a description of your change. Why is this PR needed?

lululau:
@bmag Thanks! Comment updated.

bmag:
Ok, now it makes sense, thanks.

bmag:

PR #7933, opened 2016-12-09
Add keybindings for magit clone and init.

Stebalien:


nixmaniack:
Most of the magit commands are intentionally removed. See https://github.com/syl20bnr/spacemacs/commit/b92daec7cf44b418024b49076be397009ea47cda.

Stebalien:
I know but i couldn't find these two in any of the dispatch menus. Am i missing something?

nixmaniack:
I see. I don't know if they exist in popup with any binding. But both these operations are occasional occurrences for me so I just `SPC SPC magit-init/clone`. That can be one reason of removing them as I see `SPC g i` in that commit.

Stebalien:
I think they were removed simply because all those bindings were being removed. Anyways, while I don't use these commands all the time, there's really no reason not to map them (and I do clone repos quite often).

d12frosted:
Sorry for the spam. Tried to assign someone for a review 😸 Doesn't work for me 😕

syl20bnr:
Since these key bindings are not part of the magit dispatch menu it makes sense to add them back, but I'm very careful with adding magit keybindings directly under `SPC g`.

Thank you !

PR #7839, opened 2016-11-25
which-key: Implement combined select window keys

justbur:
This is @deb0ch's idea and uses the new which-key interface

PR #7937, opened 2016-12-09
which-key: remove spacemacs/ prefix from which-key description

nixmaniack:
Transient states are prefixed with `spacemacs/`, since it's a common prefix - remove it to save some precious screen space.


nixmaniack:
Fixed by #7839 which is merged in `develop`. Closing.

PR #7943, opened 2016-12-11
switch to scratch (or messages) buffer with prefix arg opens in other window

deb0ch:
Adding a prefix arg before calling `spacemacs/switch-to-scratch-buffer` (`SPC b s`) will use `switch-to-buffer-other-window` instead of `switch-to-buffer`.

I very often want to open a scratch buffer, but also want to keep displaying whatever I'm already in. This is a way of easing this process, just doing `SPC u SPC b s`.

### DIFF
--- a/layers/+distributions/spacemacs-base/funcs.el
+++ b/layers/+distributions/spacemacs-base/funcs.el
@@ -986,7 +986,8 @@ a split-side entry, its value must be usable as the SIDE argument for
     (pop-to-buffer buffer '(spacemacs//display-in-split (split-side . below)))))
 
 (defun spacemacs/switch-to-scratch-buffer (&optional arg)
-  "Switch to the `*scratch*' buffer. Create it first if needed."
+  "Switch to the `*scratch*' buffer, creating it first if needed.
+if prefix argument ARG is given, switch to it in an other, possibly new window."
   (interactive "P")
   (let ((exists (get-buffer "*scratch*")))
     (if arg
@@ -996,6 +997,16 @@ a split-side entry, its value must be usable as the SIDE argument for
                (not (eq major-mode dotspacemacs-scratch-mode))
                (fboundp dotspacemacs-scratch-mode))
       (funcall dotspacemacs-scratch-mode))))
+
+(defun spacemacs/switch-to-messages-buffer (&optional arg)
+  "Switch to the `*Messages*' buffer.
+if prefix argument ARG is given, switch to it in an other, possibly new window."
+  (interactive "P")
+  (with-current-buffer (messages-buffer)
+    (goto-char (point-max))
+    (if arg
+        (switch-to-buffer-other-window (current-buffer))
+      (switch-to-buffer (current-buffer)))))
 
 (defun spacemacs/close-compilation-window ()
   "Close the window containing the '*compilation*' buffer."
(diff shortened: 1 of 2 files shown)


deb0ch:
Trying to harmonize this with `SPC b m` (`view-echo-area-messages`).

The current `spacemacs/switch-to-scratch-buffer` uses `switch-to buffer` and `switch-to-buffer-other-window`, while `view-echo-area-messages` uses `display-buffer`.

Which is the preferable approach ?

deb0ch:
I made up my mind, I find that it will be more convenient and less surprising to use `switch-to-buffer` and `switch-to-buffer-other-window`.

They use `display-buffer` behind the scenes anyways.

bmag:
> They use `display-buffer` behind the scenes anyways.

In most cases, that's actually not true for `switch-to-buffer` (only true if `switch-to-buffer` can't show the buffer in the currently selected window).

> I made up my mind, I find that it will be more convenient and less surprising to use `switch-to-buffer` and `switch-to-buffer-other-window`.

I agree.

Did you read the implementation of `view-echo-area-messages`? It uses `with-current-buffer` and `messages-buffer` and is quite nice, I think we should use a similar pattern. One subtle difference is that `view-echo-area-messages` moves the point to the end of the buffer *before* displaying it. Here's the implementation: (from Emacs 25.1 source)
```elisp
(defun view-echo-area-messages ()
  "View the log of recent echo-area messages: the `*Messages*' buffer.
The number of messages retained in that buffer
is specified by the variable `message-log-max'."
  (interactive)
  (with-current-buffer (messages-buffer)
    (goto-char (point-max))
    (display-buffer (current-buffer))))
```

deb0ch:
Yes, I actually wrote the code while looking at this function.

I don't see why using `with-current-buffer` is better, and we are not sure that the `*Messages*` buffer will exist at all times or that we want to create it if it doesn't exist.

I also don't see why moving the point *before* displaying the buffer is better. Is it a matter of optimization ?

bmag:
> I don't see why using `with-current-buffer` is better, and we are not sure that the `*Messages*` buffer will exist at all times or that we want to create it if it doesn't exist.

We do want to create it if it doesn't exist, and we are sure it will exist at all times because `(messages-buffer)` will create it if necessary.

> I also don't see why moving the point *before* displaying the buffer is better. Is it a matter of optimization ?

It eliminates the chance for a visual flicker after the buffer is displayed and before the point is moved to the end. To be fair, I don't know if the display engine will try to display the frame changes before leaving the function anyway, so it might be a moot point.
Another point, using `with-current-buffer` makes explicit that `goto-char` works on the messages buffer, rather than implicitly relying on `switch-to-buffer[-other-window]` changing the current buffer.
On the other hand, `switch-to-buffer[-other-window]` shouldn't be wrapped in `with-current-buffer` (because we don't want to restore the current buffer).

> I wrote it this way so that the implementation is as close as possible to the implementation of `switch-to-scratch-buffer`.

I realize that, but we should at least use `messages-buffer` because it properly creates the messages buffer (also sets correct major mode).

deb0ch:
Yeah ok sure, all of your points make much sense, and now I see that indeed this way is a lot better :+1: 

Will change it asap.

deb0ch:
Done ! :construction_worker:

d12frosted:
Looks useful 😸 

PR #7944, opened 2016-12-11
bind `SPC V` to er/contract-region

deb0ch:
This binding is instinctive and consistent with a lot of existing vim-like bindings, like `n` for `next` and `N` for `previous` and the like.

I guess that some people, like me, might not like expand-region fast keys. And it just makes more sense to have this to reflect `SPC v`.


bmag:
> I guess that some people, like me, might not like expand-region fast keys. And it just makes more sense to have this to reflect SPC v.

Do you disable the fast keys somehow? Regardless, looks fine to me, although I find a typing sequence of e.g. `v v v V` (after the initial `SPC v`) is superior (faster, easier to type) to  `SPC v SPC v SPC v SPC V`.

I've also checked what happens when calling `er/contract-region` without a marked region: looks like it's a no-op, so no bugs there.

dvzubarev:
I think this PR conflicts with Vagrant prefix. `SPC V` is used there for managing boxes.
https://github.com/syl20bnr/spacemacs/tree/develop/layers/%2Btools/vagrant

deb0ch:
Indeed, nicely spotted :smile_cat: 

Maybe we could move Vagrant related bindings under `SPC a V` ?

I don't think Vagrant layer is as widely used as `expand-region` and Vagrand can be seen as an application, am I wrong ?

deb0ch:
@bmag I deactivate it by setting `expand-region-fast-keys-enabled` to `nil` in my dotfile.

I don't think it is conflicting much with evil bindings (although I still don't know them very well), but back when I was using Emacs bindings only I always found myself hitting a fast key when I actually just wanted to type something, and this was very frustrating.

deb0ch:
Ok, I moved Vagrant layer bindings from `SPC V` to `SPC a v` to fix the conflict.

I find that even without the conflict it would make more sense to have Vagrant under `applications`.

If we want to keep `SPC a v` for another more often used application I can also make it `SPC a V`, which also makes sense and is closer to the original binding.

bmag:
@deb0ch To me it sounds logical to move Vagrant to `SPC a v` (or `SPC a V`), but since that layer already exists for a long time, and since I'm not entirely sure we really want `SPC V` for `er/contract-region`, I'd rather have @syl20bnr or @TheBB look at it.

deb0ch:
Sure, makes sense to me.

So I want to point out that `SPC V` will especially useful to `hybrid-mode` or `holy-mode` users :smile_cat:

nixmaniack:
I use expand/contract region as @bmag [explained](https://github.com/syl20bnr/spacemacs/pull/7944#issuecomment-267263505).  Not really sure how it helps in `hybrid` mode. 🤔

nixmaniack:
> I find that even without the conflict it would make more sense to have Vagrant under applications.

Also Docker should go under applications, which is currently taking `SPC D` .

deb0ch:
When in hybrid or emacs modes you have no normal mode, so you often want to write right away after selecting a region.

If you have the fast keys, you often end up doing stuff that you didn't intend to in the first place and that's very frustrating.

nixmaniack:
Well, hybrid style does have `normal` state (hybrid = vim with insert state replaced by emacs state).

For emacs style, I didn't have problem with `M-m v` `v v v V`. So I'm yet to understand how this helps. But I don't use emacs style anymore so there might be issues recently or usage pattern I'm not aware of.

deb0ch:
When using expand region in Emacs mode or hybrid mode, you often want to select something to replace it by something else. 

If what you want to replace the selection with starts with a letter that is also a fast key, you end up accidentally doing stuff that you didn't want, and that's what's very frustrating,

deb0ch:
If it is not a good idea, I don't mind this closed.

Better to have it closed than being abandoned open in the PR backlog.

syl20bnr:
Would `SPC u SPC v` works for you ? It would translates to `M-m u M-m v` which is longer but contracting a region is more used to fix a bad expansion most of the time.

I would like to keep `SPC V` for something else.

Vagrant should be behind `SPC a` indeed.

deb0ch:
Yep, `SPC u SPC v` would definitely work, I didn't think of it !

What was bothering was more that there was no way of doing it that the convenience of whatever the shortcut.

About Vagrant, glad you agree. Should it be more `SPC a v` or `SPC a V` ? I just checked and both seem free.

syl20bnr:
I would choose `SPC a V` as I suspect a small percentage of the user base uses this layer (while I use Vagrant I generally wrap the useful commands for the project in a Makefile so the layer may be even less useful) and it reflects the current `V`.

deb0ch:
Ok, just investigated it and there is actually already `SPC u - SPC v` that could call contract-region, so I'm closing this.

I'm submitting the vagrant stuff in a separate PR.

bmag:
Vagrant is now bound at `SPC a V`, see #8533

PR #7945, opened 2016-12-12
Colors Layer: Major layer enhancements and bug fixes

SteveJobzniak:
Here are some major enhancements which make the "colors" layer suck a _lot_ less. Please consider merging these in time for the next release. Because the current layer as-provided in Spacemacs is _full_ of hardcoded nonsense and bugs which makes it nearly useless (compared to just manually adding a package and having full control).

**Fixed and added:**
- Keybinding documentation has been corrected to reflect earlier code changes that moved color toggles to `<SPC t C>`.
- The messy documentation has been rewritten to fix bugs, straighten out terminology, and to make people's heads spin less.
- Added the ability to set the default saturation and lightness of rainbow-identifiers for any themes that haven't been manually overriden, instead of being forced to live with what the earliest author of this layer liked.
- Improved resetting of saturation and lightness in the transient-state, so that it now resets to the user's own defaults.
- Added support for 9 new popular themes, and updated the defaults for some old definitions to much more closely match the "spirit" of each theme's individual intended brightness and saturation level.
- Major fix to actually render the user's desired per-theme rainbow-identifiers settings when switching theme. This solves the theme switching bug where you always got stuck with an ugly mishmash of old colors and a new theme.
- Faster method of redrawing buffer for transient state.
- Carefully tuned the default fallback colors used for any themes that lack per-theme specifications. These are new, statistically balanced colors suitable for 90% of both light and dark themes, so that users get a great out-of-the-box experience instead of seeing a clashing mishmash of colors. These enhanced defaults now allow _most_ users to enjoy colors on _all_ of their themes without having to write _any_ individual per-theme optimizations. Of course, like any averages, they are a compromise, but they are a _damn_ good compromise.
- Brand new, much slicker method for adding/refreshing personal per-theme settings. It's much less verbose and less of a headache to type now, so it encourages people to add their own settings.

**Future wishlist, because I don't have more time:**
- Enhance the transient-state resetting even further, so that it resets to the default that would be used for that theme.
- Add the ability to set theme colors via a variable in :variables. Although, this is pretty much a non-issue after my documentation fixes that describe how to correctly achieve it with the current system.

**Bonus for end-users:**

Knowing that it sometimes takes months for a merge, I've created a private version of this exact code and renamed it to "[paintvars](https://github.com/SteveJobzniak/spacemacs-private-paintvars)" to avoid "Duplicate layer" errors. So just in case anyone wants to use this massively enhanced "colors" layer right away together with their normal non-development Spacemacs, you can just do the following to put it in your private layers folder, which will allow you to use it in your config just like any normal Spacemacs layer:

```bash
# if using regular emacs folder:
git clone https://github.com/SteveJobzniak/spacemacs-private-paintvars.git ~/.emacs.d/private/paintvars

# use this instead, if using ~/.spacemacs.d folder:
git clone https://github.com/SteveJobzniak/spacemacs-private-paintvars.git ~/.spacemacs.d/layers/paintvars
```

Here's the documentation, with the renamed variables (everything is exactly the same as this new version of "colors", but with the word replaced with "paintvars"):
https://github.com/SteveJobzniak/spacemacs-private-paintvars/blob/master/README.org

**Edit:**

I'm leaving Github for a long trip around the world, and won't be able to hang around to rebase this in the future if it takes a long time to merge. So either merge as quickly as possible, or do any necessary future rebasing yourselves. ;-) Take care, everyone - and have a wonderful new year! 🌟


SteveJobzniak:
@syl20bnr Came back today to check the status. Rebased (2017-02-24) onto latest develop just to show that it's still up to date. This was the last time that I have time to rebase. This patch is merge-ready and is already used by several people via my fork. Enjoy.

TheBB:
Spectacular work. Thanks! Cherry-picked in develop. You can safely delete your branch.

My only complaint is that you don't ever, ever [use underscores as part of variable names](https://github.com/syl20bnr/spacemacs/commit/b02384520e49b1b6fb139b18008261e45e945b51) in lisp. :-)

SteveJobzniak:
@TheBB Hehe, thanks for finally merging it. What a relief. It was a lot of work and I was worried this pull request was going nowhere.

And yeah, uhm, I thought maybe my function parameter names could conflict with global variables of the same name, so I used underscores in that location. Now I know that it's safe to use any name with hyphens there instead. ;-)

I was just a few days into having migrated to Spacemacs and learning Lisp for the first time, and it made my head spin with its quirks. It was like writing an [abstract syntax tree](https://en.wikipedia.org/wiki/Abstract_syntax_tree) by hand. Cool in theory but I'm glad to be back to C instead. ^_^

PR #7952, opened 2016-12-12
Use helm ag file with interactive input

quicknir:
Currently, when using helm-ag on an individual file through spacemacs, instead of going directly into the helm prompt, it asks for the pattern upfront in the minibuffer and only shows helm after enter is hit. This isn't as nice and is not consistent with anything else.

Most likely this has not been noticed because most people use swoop, which I don't.

### DIFF
--- a/layers/+completion/helm/packages.el
+++ b/layers/+completion/helm/packages.el
@@ -183,7 +183,7 @@
       (defun spacemacs/helm-file-do-ag (&optional _)
         "Wrapper to execute `helm-ag-this-file.'"
         (interactive)
-        (helm-ag-this-file))
+        (helm-do-ag-this-file))
 
       (defun spacemacs/helm-file-do-ag-region-or-symbol ()
         "Search in current file with `ag' using a default input."


LemmingAvalanche:
What’s the “Change-Id” trailer related to?  Gerrit?

quicknir:
@LemmingAvalanche  Yes, my apologies, it's been removed.

LemmingAvalanche:
Oh don’t apologise, maybe it was useful to the commit for all I knew. :-)  It was more out of curiosity.

LemmingAvalanche:
Also I don’t know if you care about this but the committer name or mail is different from the author name or mail.

d12frosted:

But note that I am not sure why we use wrapper instead of `helm-do-ag-this-file`. Not sure about old `helm-ag-this-file`, but `helm-do-ag-this-file` is interactive. Is `spacemacs/helm-file-do-ag` called with arguments somewhere? Like when it's used as custom action for helm.

quicknir:
@d12frosted I'm not completely certain, but I think that maybe the smart searches are doing something "clever" involving reflection. E.g. you have the following code:

```
         (spacemacs//helm-do-search-find-tool "helm-file-do"
                                              dotspacemacs-search-tools
                                              default-inputp)))
```

The fact that "helm-file-do" is getting passed as a string literal, suggests to me that the smart search tool builds the name of the function that it's going to call. If it's going to do that, it would need exact consistency in the names of the functions, i.e. `spacemacs/helm-*-do-*`. Just a guess though.

PR #7959, opened 2016-12-13
add magnars/multiple-cursors.el's mc-lists.el to gitignore

deb0ch:
Even if Spacemacs doesn't support this package, it still puts that file in your `.emacs.d` if you use it as an additional package.

(and prevents users from updating too)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,6 +44,7 @@ python-*-docs-html
 /session.*
 /srecode-map.el
 /recentf
+.mc-lists.el
 
 # Private directory
 private/


d12frosted:
It's possible to configure that by changing the value of `mc/list-file`. I am honestly against such modifications to `.gitignore` file. I think it would be nice to configure that to be somewhere inside of `spacemacs-start-directory/.cache` folder. Probably using some `eval-after-load`. So that doesn't force us to really support this thing but shows that we respect mc users 😸

d12frosted:
The question is - where to put this?

deb0ch:
Why are you against adding it to the `.gitignore` ?

We could indeed modify the value of `mc/list-file`, but then several problems arise:

- In addition to the `eval-after-load`, we have to check that the user did not (or will not) change the default value.
- As you said, where do we put that code ?
- Will `multiple-cursors` handle the change and move the file by itself ?
- Will it not break users' current `mc-lists.el` file ?
- Whatever the solution, it will be surprising to the user if Spacemacs modifies the configuration of an officially unsuported package. Would not be very [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) (a word that I learnt here :smile_cat:).

From my point of view it would be better if this issue was addressed, and the `.gitignore` solution seems to me to be the simplest.

What downsides do you see in it ?

bmag:
See: #5628, https://github.com/magnars/multiple-cursors.el#unknown-commands

The usual solution is to set cache files in `spacemacs-cache-directory`, not add them to gitignore. Since multiple-cursors isn't configured by us, it doesn't happen for `mc/list-file`. For MC users, the user-side solution is to configure the variable in `user-init` (`user-config` may be fine as well, depending on user's config):
```elisp
(setq mc/list-file (expand-file-name ".mc-lists.el" spacemacs-cache-directory))
```

However @nixmaniack [comments](https://github.com/syl20bnr/spacemacs/pull/5628#issuecomment-203056905) that "there are packages which pull `mc` as dependency e.g. `js2-refactor`, `clj-refactor`", which IMO is a reason we should handle the update issue somehow. Although https://github.com/magnars/multiple-cursors.el/pull/216 is reported to be usable, it doesn't look like MC and Evil integration will be merged soon, and I can't find a better place to configure `mc/list-file`, so my opinion is that we should add `.mc-lists.el` to gitignore until we have a proper configuration for it.

I'd like to answer some of the questions here: (that weren't addressed to me, but oh well)

@d12frosted:
> I think it would be nice to configure that to be somewhere inside of spacemacs-start-directory/.cache folder. Probably using some eval-after-load. So that doesn't force us to really support this thing but shows that we respect mc users :smile_cat:
The question is - where to put this?

We don't need any `eval-after-load`, because a simple `setq` will do. The tougher question is indeed "where to put this". I don't have a good answer for this one, maybe a special case in `core`? I think the gitignore approach is the simplest one.

@deb0ch:
> Why are you against adding it to the `.gitignore`?

It's not the "correct" solution. It's a cache file, cache files belong in the cache directory, which is already ignored by gitignore. More to the point, adding the file to gitignore means that it will clatter the top-level Spacemacs directory (`.emacs.d`).

> Whatever the solution, it will be surprising to the user if Spacemacs modifies the configuration of an officially unsuported package.

Indeed.

> Will it not break users' current mc-lists.el file ?

No, the old file won't be modified, but it *will* be ignored.

> Will multiple-cursors handle the change and move the file by itself ?

Little chance of that. Users who already have an `mc/list-file` will need to move it themselves, unless they already configured it to be in the cache directory and happened to choose the same name as the hypothetical name that will hypothetically be configured by us.

deb0ch:
> No, the old file won't be modified, but it will be ignored.

That's what I meant by "breaking" :smile_cat: 

I get the point that such files should go into the proper directory defined by Sapcemacs conventions and that it would clatter the top-level directory, but heh this package is not configured by Spacemacs so my point of view is either it is fully configured and integrated (which, looking at @Bmag's link really should happen one day) or it should not be at all.

Plus this configuration, that would be hidden somewhere in the depths of Spacemacs, could be called "Spacemacs' black magic" or "Spacemacs doing things behind your back that you don't expect", and be reproached to the entire project.

So after all this I still feel like the `.gitignore` solution is the simplest and the best one.

d12frosted:
@deb0ch and @bmag 

Your reasoning makes sense. So yeah, let's add that file to `.gitignore` for now. I don't want to prevernt users from being able to update using built-in update system. But once we start supporting that package, we should remove it from `.gitignore`.

P. S. @bmag you said really well about general policy regarding `.gitignore` file.

P. P. S.  @deb0ch Thank you :+1: Cherry-picked into `develop`, you can safely delete your branch.

d12frosted:
Regarding transition from old cache file to new one. Once we have a proper integration for `multiple-cursors` we could check for cache file existence in default location and move it to proper one. So old users of `mc` would not notice the change. Hopefully.

syl20bnr:
@d12frosted @deb0ch @bmag 
Since mc is not handled by spacemacs yet it makes sense to put it in `.gitignore`.
But typically this file is not something volatile, I mean there is a high probability that a user wants to keep this file in its dot directory. So a better place for it could be in `private`. We'll see when we are at the point where we can support mc. I believe we should provide an additional mechanism on top of the `mc/list-file` to support functions of all packages configured by spacemacs (actually we already have this mechanism, they will surely by dispatched to layers in `post-init-multiple-cursors` functions).

Issue #7960, opened 2016-12-14
Undo PR #6981 and replace it with unfill package


ghost:
https://github.com/syl20bnr/spacemacs/pull/6981 added an old, buggy version of the code.

That's not the best version of the code... I suggest deleting this commit.

> Due to implementation details unfilling doesn't work when called twice via ~M-x~.

Here is the evolved version of the exact code, and it is much more advanced and handles `M-x` correctly:

https://github.com/purcell/unfill/blob/master/unfill.el

http://melpa.org/#/unfill

So undo this commit and merge a dependency on the unfill package instead. ;-)

d12frosted:
> buggy version of the code

Please be more specific.

> Here is the evolved version of the exact code, and it is much more advanced

Indeed, it's better than the version I added 😸 

> handles M-x correctly

How is that? Implementation details that don't like it work when called via `M-x` is `(eq last-command this-command)` piece which is used in both versions. And I just tested `unfill` - got expected result - it only fills when called from `M-x`.

d12frosted:
BTW, forgot to mention that we do not delete commits 😸 I've created PR that adds `unfill` package and removes that less useful function we introduced in #6981. Thanks for sharing `unfill`. I saw it before but for some reason didn't pay attention to it's superiority 😸

Miciah:
Since #7967 / https://github.com/syl20bnr/spacemacs/commit/d76c5a75d5f08ac4a832a067ed2fcd4860741aad has been merged, can this issue be labeled as "Fixed in develop"?

PR #7967, opened 2016-12-14
replace fill-or-unfill with unfill package

d12frosted:
Following #7960. It's better than previous version for the following reasons.

- It works well with regions (thanks to `deactivate-mark` set to `nil`).
- Exposes functions to unfill region and paragraph.
- Better second usage check by using `this-command`.


d12frosted:

PR #7974, opened 2016-12-16
prodigy-view-mode: go to file at point

FrancescElies:
This is a small feature request.
When running a process and having a look at it's output in prodigy (`prodigy-view`), I would love to be able to open the file at point with <gf> as in other modes.

What do you think?
Thanks for your time

### DIFF
--- a/layers/+tools/prodigy/packages.el
+++ b/layers/+tools/prodigy/packages.el
@@ -27,4 +27,6 @@
       "L" 'prodigy-start
       "d" 'prodigy-jump-dired
       "g" 'prodigy-jump-magit
-      "Y" 'prodigy-copy-cmd)))
+      "Y" 'prodigy-copy-cmd)
+    (evil-define-key 'motion prodigy-view-mode-map (kbd "gf") 'find-file-at-point)
+))


FrancescElies:
I also added revert buffer, would it be fine.
This is useful in case you want to add a new service without having to restart emacs

d12frosted:
Hey @FrancescElies,

Thanks for PR!

I am not sure about the part about `find-file-at-point`. Since this change looks useful not only for prodigy layer.

 But regarding your second commit - [you should document it in README file of prodigy layer](https://github.com/syl20bnr/spacemacs/blob/master/CONTRIBUTING.org#contributing-a-keybinding). 

And note that these two commits [would look better](https://github.com/syl20bnr/spacemacs/blob/master/CONTRIBUTING.org#pull-request) as two separate PRs.

FrancescElies:
Hi @d12frosted,
thanks for your comments I made a new PR for the revert buffer binding https://github.com/syl20bnr/spacemacs/pull/8032

FrancescElies:
With regards to `find-file-at-point`, maybe we would like do something similar as with compile-mode see https://github.com/syl20bnr/spacemacs/blob/develop/layers/+distributions/spacemacs-base/keybindings.el#L238

What do you think, shall this be placed in keybindings.el?

Best

d12frosted:
Thanks for splitting these :) Welp, actually I went ahead and just merged this (so it's now available in `develop`). This should not be in `keybindings.el` as it [has other purpose](https://github.com/syl20bnr/spacemacs/blob/develop/doc/LAYERS.org#keybindingsel).

FrancescElies:
Wonderful, (and thanks for the explanations:)

PR #7977, opened 2016-12-16
Add notmuch integration to org-mode.

Stebalien:


layers/+emacs/org/packages.el
```diff
@@ -82,6 +83,11 @@
   (spacemacs|use-package-add-hook mu4e
     :post-config (require 'org-mu4e nil 'noerror)))
 
+(defun org/pre-init-notmuch ()
+  ;; Load org-notmuch when notmuch is actually loaded
+  (spacemacs|use-package-add-hook notmuch
+    :post-config (require 'org-notmuch nil 'noerror)))
+
```
nixmaniack: 1. How will this work when spacemacs doesn't provide a `notmuch` layer and there isn't a `init-notmuch` in spacemacs layers? 

2. If spacemacs isn't providing notmuch - is it right to add this configuration here?

Stebalien: This just mimics the mu4e integration.

1. This doesn't actually depend on a notmuch layer (well, it shouldn't at least). Instead, this just casues `org-notmuch` to be loaded when and if the notmuch package gets loaded.
2. This is really org-mode functionality so it belongs in the `org` layer (IMO). Additionally, users *can* install the notmuch package without a layer so this is still useful for spacemacs users. Regardless, if all goes according to plan, this discussion should be moot rather soon :wink:.

Stebalien:
Superseded by #8560.

PR #7981, opened 2016-12-17
adds basic org mode lazy installation

JoshTGreenwood:
I'm not sure this is exactly right but it seems like it may be a good start.

I'd be happy to add any other org specific packages to the default lazy-install

### DIFF
--- a/layers/auto-layer.el
+++ b/layers/auto-layer.el
@@ -100,6 +100,9 @@
 (configuration-layer/lazy-install 'rust :extensions '("\\(\\.rs\\'\\)" rust-mode))
 (configuration-layer/lazy-install 'rust :extensions '("\\(\\.toml$\\)" toml-mode))
 
+;; org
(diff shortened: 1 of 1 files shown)


d12frosted:
Spacemacs [comes with two layers](https://github.com/syl20bnr/spacemacs/pull/5372) configuring org-mode. One of them is useful for reading org files and another one for editing them (with some other things). Since all Spacemacs documentation is written in `org-mode`, `spacemacs-org` is included to `spacemacs` distribution. But since `org` layer brings too many dependencies, it's not really good to force that layer for people that just opened documentation but don't use `org-mode`.

d12frosted:
Aside from this - all PR [must be made against develop](https://github.com/syl20bnr/spacemacs/blob/master/CONTRIBUTING.org#pull-request). In your cases that's not critical though.

JoshTGreenwood:
Thanks for the heads up about develop, sorry about that.

I didn't realize `spacemacs-org` was a thing, I think that is probably sufficient as far as lazy installation goes.  Closing this PR.

PR #7986, opened 2016-12-17
Add the "evil-smartparens" package to editing layer

jballanc:
The default implementation of smartparens is not aware of all the various "extra" editing functions in evil mode. For example, if the caret is at `|` in the following code:
```
(defun foo (do-thing |bar) (other-thing baz))
```
and a user hits "C" in normal mode, the result is:
```
(defun foo (do-thing |
```
leaving two unbalanced parens. It is much more likely that the user would want the result be:
```
(defun foo (do-thing |))
```
which is the result following this commit.

### DIFF
--- a/layers/+spacemacs/spacemacs-evil/packages.el
+++ b/layers/+spacemacs/spacemacs-evil/packages.el
@@ -25,6 +25,7 @@
         evil-matchit
         evil-numbers
         evil-search-highlight-persist
+        evil-smartparens
         evil-surround
         ;; Temporarily disabled, pending the resolution of
         ;; https://github.com/7696122/evil-terminal-cursor-changer/issues/8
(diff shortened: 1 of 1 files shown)


Stebalien:
That's an interesting package but... changing`C` to do anything but delete till the end of the line by default will piss off lots of users with muscle memory.

jballanc:
Would making it configurable with a layer variable be a reasonable alternative?

Stebalien:
It would probably have to be a `dotspacemacs` variable defaulting to off (like `dotspacemacs-remap-Y-to-y$` etc.).

bmag:
We already have an `evil-cleverparens` layer for `evil-cleverparens`.

jballanc:
Would it make sense to add `evil-smartparens` to the same layer and provide a layer variable to switch between the two?

syl20bnr:
Thank you! 💜 

My sincere apologies, this PR has not been merged in due time and has been replaced by a more recent PR (https://github.com/syl20bnr/spacemacs/pull/9899). I hope it won't prevent you from opening new PRs, don't hesitate to ping me with a PM on gitter if you have stalled PRs that you judge important and useful.

Also note that I'm move `evil-cleverparens` from its own layer to `spacemacs-evil` layer with a toggle. For now `evil-cleverparens` replaces `evil-smartparens` but we can change that if it is shown that `evil-smartparens` is a better package to accomplish this job.

jballanc:
Thanks for the updates! Just switched to the `develop` branch and it works great so far.

PR #7987, opened 2016-12-17
dumb-jump: Specify selector from completion framework used

deuill:
This commit adds a conditional which allows `dumb-jump` to use a selector matching the current completion framework used (`ivy` or `helm`), instead of the default popup.

This relates to work done for adding Helm support to `dumb-jump` here: https://github.com/jacktasia/dumb-jump/pull/96

The motivation behind this is two-fold: the fact that popups are known to be broken when used with conflicting packages, such as `fci-mode` for showing a column marker, and the need for better completions with the possiblity for fuzzy-matching.

This should hopefully lead to a more consistent experience as well.


deuill:
Note that this implicitly changes the default behaviour of showing a popup to showing whatever your completion framework provides. In the case where neither `helm` or `ivy` are used, the default popup is used.

d12frosted:

Issue #7995, opened 2016-12-19
Horizontal scrolling and more vertical scrolling in scrolling transient state


Somelauw:
I think the scrolling transient state should have a way to scroll horizontally and maybe more keys for scrolling vertically.

I propose:
SPC n j for scrolling down
SPC n k for scrolling up
SPC n h for scrolling left
SPC n l for scrolling right
SPC n 0 or SPC n H for scrolling to the complete left
SPC n $ or SPC n L for scrolling to the complete right
SPC n d or SPC n J for scrolling half a page up
SPC n u or SPC n K for scrolling half a page down
SPC n f for scrolling a page down
SPC n b for scrolling a page up

Somelauw:
Since it's been implemented on develop, I think this can be closed.

PR #7997, opened 2016-12-19
Add more bindings to scroll-transient-state

Somelauw:
Add more bindings to scroll-transient-state

As described in #7995. Once in scroll-transient-state more bindings are available. I haven't added new SPC n mappings for scrolling yet. Let me know if I should add these too or if perhaps there should be a generic binding that switches to scroll-transient-state/body.

Edit: Originally, I had 0 and $ bound to evil-scroll-left and evil-scroll-right (just like H and L). I have removed them, because these keys are probably expected to scroll to the very first and last column of the buffer.

### DIFF
--- a/layers/+distributions/spacemacs-base/keybindings.el
+++ b/layers/+distributions/spacemacs-base/keybindings.el
@@ -35,6 +35,7 @@
                                        ("kD"  "delete-backward")
                                        ("k`"  "hybrid")
                                        ("n"   "narrow/numbers")
+                                       ("N"   "navigation")
                                        ("p"   "projects")
                                        ("p$"  "projects/shell")
                                        ("q"   "quit")
--- a/layers/+distributions/spacemacs-bootstrap/packages.el
+++ b/layers/+distributions/spacemacs-bootstrap/packages.el
@@ -139,17 +139,50 @@
   ;; scrolling transient state
   (spacemacs|define-transient-state scroll
     :title "Scrolling Transient State"
+    :doc "
+ Buffer^^^^              Full page^^^^     Half page^^^^        Line/column^^^^
+ ──────^^^^───────────── ─────────^^^^──── ─────────^^^^─────── ───────────^^^^─────
+ [_<_/_>_] beginning/end [_f_/_b_] down/up [_J_/_K_] down/up    [_j_/_k_] down/up
+  ^ ^ ^ ^                 ^ ^ ^ ^          [_H_/_L_] left/right [_h_/_l_] left/right
+  ^ ^ ^ ^                 ^ ^ ^ ^          [_d_/_u_] down/up     ^ ^ ^ ^"
     :bindings
-    ("," evil-scroll-page-up "page up")
-    ("." evil-scroll-page-down "page down")
+    ;; buffer
+    ("<" evil-goto-first-line)
+    (">" evil-goto-line)
+    ;; full page
+    ("f" evil-scroll-page-down)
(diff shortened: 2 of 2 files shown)


deb0ch:
I like this PR, because the bindings are consistent with what you would expect of Evil motions :smile_cat: 

It would be very instinctive I think to activate the transient state on any of these bindings, like `SPC n j`, `SPC n $`, etc... When you do one scroll command, isn't it likely that you will want to do many to move around your buffer ?

It would conflict with the narrowing stuff though, but the `SPC N` prefix is free so we could move narrowing to that :smile_cat: 

But now that I think of it, why are the **s**crolling actions under `SPC n` in the first place ?

@syl20bnr It seems that you are the master of key bindings, any insights ? :smiley_cat:

bmag:
This looks nice. Can we put narrowing on `SPC n` and scrolling on `SPC N`? Reasons:
- narrowing is quite useful at times
- there are already easy bindings for most scrolling commands
- the TS lowers the inconvenience of having scrolling on `SPC N`

deb0ch:
I don't think it's a bad thing or that it should be changed, but the question remains: why are the scrolling actions under the letter `n` or `N` ?

bmag:
@deb0ch I don't know, maybe because these are "navigation" commands (and "s" is taken by "search")

deb0ch:
Oooooooh THAT'S the mnemonic :open_mouth: 

Thanks @bmag ! :smiley_cat: :+1:

Somelauw:
I have put "narrow" back to n and moved "navigation" to N now. I have also squashed the commits.
Should I keep the original "N,", "N.", "N<" and "N>" shortcuts?

bmag:
@Somelauw just tried it, and it's very convenient. I think we can improve it with:
- `<`, `>`: go to start/end of buffer (evil-goto-first-line`, `evil-goto-line`) (that will also resemble emacs style `M-<` and `M->`)
- `,`, `.`: they don't do any harm as they are, but maybe we can bind `,` to `evil-jump-backward` and `.` to `evil-jump-forward` (that will also resemble emacs style `M-.` and `M-,`)

If no one has any objections, I'll merge it once @Somelauw applies these changes.

Somelauw:
I have added the requested shortcuts. In addition, I have added a hydra-docstring for readability.

bmag:
Hmm.. for some reason `SPC N .` (spacemacs/scroll-transient-state/evil-jump-forward) doesn't move the cursor (`SPC N ,` works fine). Do you experience the same problem?

Somelauw:
> Hmm.. for some reason SPC N . (spacemacs/scroll-transient-state/evil-jump-forward) doesn't move the cursor (SPC N , works fine). Do you experience the same problem?

@bmag Thanks. I have noticed that both evil-jump-forward and evil-jump-backward behave differently in scroll-transient-state:

- evil-jump-backward (,)
Behaviour: goes back one position on its first invocation but subsequent key presses jump back and forth.
Expected: Pressing it multiple times goes further back in history

- evil-jump-forward (.)
Behaviour: Does nothing
Expected: Undoes evil-jump-backward

When using evil's built-in C-o and C-i key bindings, these commands behave as expected. I'm not sure what causes them to behave differently in scroll-transient-state.

bmag:
@Somelauw so lets drop the `,` and `.` bindings and get this merged?

Somelauw:
I have dropped them for now.

bmag:

PR #7998, opened 2016-12-19
layouts: Default free buffers to current perspective

### LINKED ISSUE #7931: persp-mode warning when submitting via org-projectile/capture
#### Description :octocat:
When finishing or aborting a buffer opened with `org-projectile/capture`,
persp-mode complains that the buffer being killed is not in the current
perspective. You have to hit something to continue.

- Start Emacs
- `SPC a o p`
- Type something
- `C-c C-c`

Warning from persp mode

capture-todos org 2016-12-08 12-38-16

Buffer closes w/o warning

#### System Info :computer:
- OS: darwin
- Emacs: 25.1.1
- Spacemacs: 0.200.5
- Spacemacs branch: develop (rev. 4289a49)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: helm
- Layers:
```elisp
(better-defaults spacemacs-layouts helm emacs-lisp markdown
                 (auto-completion :variables auto-completion-enable-sort-by-usage nil auto-completion-enable-snippets-in-popup nil auto-completion-private-snippets-directory
                                  (expand-file-name "snippets" dotspacemacs-directory))
                 (org :variables org-enable-github-support t org-enable-reveal-js-support t)
                 colors osx github javascript react deft ruby
                 (spell-checking :variables spell-checking-enable-by-default t enable-flyspell-auto-completion t)
                 ranger version-control rcirc tmux yaml docker elm restclient typescript evil-little-word auto-correct frame-geometry cleverparens-lispy flow)
```
- System configuration features: NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS

aaronjensen:
Fixes #7931

### DIFF
--- a/layers/+spacemacs/spacemacs-layouts/packages.el
+++ b/layers/+spacemacs/spacemacs-layouts/packages.el
@@ -118,7 +118,9 @@
             persp-reset-windows-on-nil-window-conf nil
             persp-set-last-persp-for-new-frames nil
             persp-save-dir spacemacs-layouts-directory
-            persp-set-ido-hooks t)
+            persp-set-ido-hooks t
+            persp-add-buffer-on-after-change-major-mode 'free
+            persp-kill-foreign-buffer-indirect-behaviour-override 'as-base-buffer)
 
       (defun spacemacs//activate-persp-mode ()
         "Always activate persp-mode, unless it is already active.


layers/+spacemacs/spacemacs-layouts/packages.el
```diff
@@ -118,7 +118,8 @@
             persp-reset-windows-on-nil-window-conf nil
             persp-set-last-persp-for-new-frames nil
             persp-save-dir spacemacs-layouts-directory
-            persp-set-ido-hooks t)
+            persp-set-ido-hooks t
+            persp-add-buffer-on-after-change-major-mode 'free)
```
aaronjensen: These were somewhat, but not entirely in alphabetical order. I didn't want to change them and have lots of unrelated changes in this change to mess w/ blame, but i'd be happy to alphabetize if you think it'd be good.

aaronjensen:
Let me know if I should filter out `*Help*` and/or any other buffers via https://github.com/syl20bnr/spacemacs/issues/7931#issuecomment-268081622

bmag:
Sorry for the delay in handling this PR. I've just tried it, but it doesn't seem to solve the problem it's supposed to solve. When I use `org-capture` (`SPC a o c`) or `projectile/org-capture` (`SPC a o p`), I still see the behavior decsribed in https://github.com/syl20bnr/spacemacs/issues/7931#issue-194438862. (a warning from `persp-mode` after `C-c C-c` in capture buffer)

aaronjensen:
Is your persp-mode up to date?

bmag:
Yes, it is

aaronjensen:
Indeed, this behavior seems to have broken. /cc @bad-ptr

aaronjensen:
fwiw, the buffer does appear to get added to the persp (`persp-buffer-in-persps` is set to the current perspective) but the warning still appears.

aaronjensen:
Ok, I tracked it down to this, though the persp is in `persp-buffer-in-persps`, the `(persp-buffers (get-current-persp))` does not return that buffer, so `persp-contain-buffer-p` returns `nil`. I don't know enough to know what's going on, but something seems to be not working properly, or maybe something else broke. 

```
(memq (persp-get-buffer-or-null buff-or-name) (persp-buffers persp))
```

Bad-ptr:
How can I reproduce this?
Suppose I cloned the persp-mode repo and visiting the persp-mode.el source. What should I do next?
`M-x package-install RET org-projectile RET`, then `M-x projectile-mode RET` I guess, what's next?)

aaronjensen:
org-projectile is a spacemacs specific thing, I think. One "easy" way to repro would be:

1. Clone spacemacs to .emacs.d
1. Checkout the `develop` branch
1. Run emacs
1. Select all of the default options (you can probably select holy mode in the first option if you're not used to vim bindings, just use `M-m` instead of `SPC` for any bindings below)
1. Wait for all packages to install
1. Hit `SPC f e d` (`M-m` instead of `SPC` if holy mode) to edit your `.spacemacs`
1. Add the following to `dotspacemacs-configuration-layers:

    ```
   '(
    ;; ...
     (org :variables
          org-project-file "TODOs.org")
    )
    ```

1. Restart emacs `SPC q r`
1. Let it install new packages
1. Restart emacs again `SPC q r`
1. Open a file in a project directory `SPC f f` to find one
1. Open a layout for that project `SPC p l` 
1. Hit `SPC a o p` to open org projectile capture
1. Hit `C-c C-k` to cancel, this should pop the warning
1. `(setq persp-add-buffer-on-after-change-major-mode 'free)`
1. Try again, still should pop the warning.

Bad-ptr:
Thanks for instructions. Looks like CAPTURE-* buffer is an *inderect* buffer.
I think that I can catch this case looking at the return value of the `buffer-base-buffer` function(google gives very little information about indirect buffers).

aaronjensen:
You're welcome, thanks for bearing w/ the 16 step repro 😄 

It looks like you're correct afaict about `buffer-base-buffer`. What would you do w/ this information? Not show warnings for it if it's indirect?

Bad-ptr:
Ok, the best solution on this issue that i can think of so far: https://github.com/Bad-ptr/persp-mode.el/commit/06d56333d738c57fa543e47e7eb1c4962bd14344
Setting the `persp-kill-foreign-buffer-indirect-behaviour-override` to `'as-base-buffer` must work for this case.

aaronjensen:
@Bad-ptr this works, thank you. For this particular use case, we no longer need `persp-add-buffer-on-after-change-major-mode 'free` either, though it seems like a good idea to keep this any way so as to avoid random instances of that warning.

aaronjensen:
@bmag this is ready to go now. The persp-mode update may not yet be on melpa, but it should be shortly.

aaronjensen:
This was fixed upstream, closing.

PR #8002, opened 2016-12-20
ielm: start ielm in insert mode

ralesi:
ielm: start ielm in insert mode


rgrinberg:
IMO, this isn't worth the inconsistency. I frequently start ielm and immediately paste something for example.  So it doesn't even save some people any keystrokes.

ralesi:
I see your point, but it looks like that this is defined in `evil-vars.el` anyway, so it is already defined.

PR #8005, opened 2016-12-20
recentf: fix referencing to cache directory when symlinked.

ralesi:
Otherwise will show up in recentf list


TheBB:

PR #8006, opened 2016-12-20
recentf: add let bound variable to stop org from contaminating the re…

ralesi:
When spacemacs-recentf-ignore is `t`, will not update recentf-list.


TheBB:
Can we do this instead?

`(let ((recentf-keep '((lambda (&rest args) nil)))) ....)`

That way we don't have to introduce another variable and and advice.

TheBB:
Or this:

Can we do this instead?

```elisp
(let ((recentf-exclude '((lambda (&rest args) t)))) ....)
```

I haven't tried but one of them feels like it should work.

TheBB:
After looking at recentf code I went with the latter. Thanks though!

ralesi:
We can just pop your code in front of each of those though and it will work.

PR #8007, opened 2016-12-20
undo-tree: evil bindings for visualizer and window improvements

ralesi:
Adds evil bindings to undo-tree.


TheBB:

I changed to use evilified instead. I wasn't able to get the selection mode to work, neither the bindings nor the actual mode. I don't understand its use, so I removed those (the original bindings are still present at least).

PR #8008, opened 2016-12-20
emacs-lisp: add hl-defined package.

ralesi:
Add hl-defined package.

### DIFF
--- a/layers/+lang/emacs-lisp/packages.el
+++ b/layers/+lang/emacs-lisp/packages.el
@@ -21,6 +21,7 @@
         ggtags
         helm-gtags
         (ielm :location built-in)
+        hl-defined
         macrostep
         semantic
         smartparens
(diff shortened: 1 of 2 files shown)


layers/+lang/emacs-lisp/packages.el
```diff
@@ -105,6 +106,25 @@
...
 
+(defun emacs-lisp/init-hl-defined ()
+  (use-package hl-defined
+    :commands hdefd-highlight-mode
+    :init
+    ;; (add-hook 'emacs-lisp-mode-hook 'hdefd-highlight-mode)
+    (spacemacs|add-toggle hdl-defined
+      :mode hdefd-highlight-mode
+      :documentation "Syntax highlight known defuns and variables."
+      :evil-leader "thd")
```
TheBB: I assume this binding should be in the emacs-lisp leader map instead?

PR #8009, opened 2016-12-20
bug: fix defun naming.

ralesi:
Someone was using `nameless` =)


TheBB:
This has been fixed in the meantime. Thanks though!

PR #8012, opened 2016-12-20
Bug (ivy) : define matcher desc, deprecated from package, add M-SPC

ralesi:
`ivy--matcher-desc` was deprecated, make spacemacs defun for identification.


TheBB:

PR #8015, opened 2016-12-20
Add :layers property to keep track of all layers with package init

ralesi:
Grants visibility to owners of layers that are currently not loaded.  Added benefit is that `-help` functions can now find `package.el` files for unused layers.


ralesi:
I might need some help in fixing the tests that are failing.

PR #8017, opened 2016-12-20
emacs-restart: make debugging commands accessible.

ralesi:
Add debugging commands to `emacs-restart`, using `SPC q t` and `SPC q T`.


TheBB:

PR #8019, opened 2016-12-20
Fix some issues with persistent server

ralesi:
* frame killer logic wasn't correct when server not running
* even when debugging, we want to kill spacemacs


TheBB:

PR #8021, opened 2016-12-20
Delete alternate actions is too invasive

ralesi:
Accidentally deleted a file this way, we need to implement a prompt before delete to prevent clobbering a file.


d12frosted:
I do understand your concerns about this function. But instead of commenting it, I would implement version with prompt or move binding to `D` (compared to lowercase `d`).

d12frosted:

I went ahead and implemented it here: https://github.com/syl20bnr/spacemacs/commit/3e60792136d02252288052d1a4498bc25f8d4a9f

PR #8022, opened 2016-12-20
Defer loading of flyspell-correct-helm, otherwise it just requires helm

ralesi:
Defer loading this package because it explicitly requires `helm`.

### DIFF
--- a/layers/+checkers/spell-checking/packages.el
+++ b/layers/+checkers/spell-checking/packages.el
@@ -78,7 +78,8 @@
   (use-package flyspell-correct-ivy))
 
 (defun spell-checking/init-flyspell-correct-helm ()
-  (use-package flyspell-correct-helm))
+  (use-package flyspell-correct-helm
+    :defer t))
 
 (defun spell-checking/init-flyspell-correct-popup ()
   (use-package flyspell-correct-popup))


d12frosted:
Not sure I understand the problem beneath. I mean, what is the problem if it requires `helm` since it does so only for `helm` users via 

```emacs-lisp
(flyspell-correct-helm :toggle (configuration-layer/layer-usedp 'helm))
```

I don't see `helm` being loaded and configured if use `ivy` layer. Or did I misunderstood your comment?

d12frosted:
Anyways I have cherry-picked it into `develop`, as it's not good to instantly load this package. But I have made several modifications in following commit: https://github.com/syl20bnr/spacemacs/commit/d3fd067a3d0c7f24eaaa2d38676cfbd135ba48f1.

- You can't just defer it as it configures `flyspell-correct`. 
- This should be made not only for `flyspell-correct-helm`, but for `ivy` and `popup` versions as well.

ralesi:
Boris, 


The issue is that helm itself defers 1 sec to finish spacemacs loading,  which this then changes by requiring helm. 


Rich






On Thu, Dec 22, 2016 at 6:38 AM -0700, "Boris" <notifications@github.com> wrote:












Closed #8022.

d12frosted:
> The issue is that helm itself defers 1 sec to finish spacemacs loading,  which this then changes by requiring helm.

Ah. Thanks for explanation :) Makes sense. And it should behave as desired even after my change.

PR #8023, opened 2016-12-20
Speedup: defer loading of helm-purpose.  Requires all of purpose.

ralesi:
`helm-purpose` requires `windows-purpose` directly, so defer loading of it.


bmag:
`window-purpose` isn't deferred in the first place, but `helm-purpose` requires `helm-buffers` so this PR does defer part of Helm. Still a good PR.

bmag:

PR #8024, opened 2016-12-20
Defer gmail-message, requires a lot of files.

ralesi:
Defer `gmail-message` to speedup startup.


d12frosted:
Nice find! But again I am worried if it's ever get loaded :) There are no rules that force it.

syl20bnr:
Thank you ! 👍 

PR #8026, opened 2016-12-20
Fix documentation about replace mode

maxnordlund:
There is in fact a replace mode in vanilla Vim, see http://vimdoc.sourceforge.net/htmldoc/insert.html#mode-replace and not exclusive to Evil.


syl20bnr:
Thank you !

PR #8034, opened 2016-12-22
Remove unused variable p2 and two progn

### LINKED ISSUE #8031: spacemacs/select-current-block, unused variable "p2", and possibly a superfluous "(progn"
#### Description :octocat:
I only have basic lisp knowledge so i might be wrong here, but it seems like a couple of lines in the `spacemacs/select-current-block` could be removed.

This is what it currently looks like:
```elisp
(defun spacemacs/select-current-block ()
  "Select the current block of text between blank lines."
  (interactive)
  (let (p1 p2)
    (progn
      (if (re-search-backward "\n[ \t]*\n" nil "move")
          (progn (re-search-forward "\n[ \t]*\n")
                 (setq p1 (point)))
        (setq p1 (point)))
      (if (re-search-forward "\n[ \t]*\n" nil "move")
          (progn (re-search-backward "\n[ \t]*\n")
                 (setq p2 (point)))
        (setq p2 (point))))
    (set-mark p1)))
```
It's defined here:
https://github.com/syl20bnr/spacemacs/blob/b6d52030b165fc5b74c95d2ade9f1ba6b5c15ab7/layers/%2Bdistributions/spacemacs-base/funcs.el#L767

The `p2` variable doesn't seem to be used, so it's initial definition `(let (p1 p2)` probably could be removed: `(let (p1)`.

If i understand it correctly, then a `progn` is used to call several commands from one place, like it does in both if statements, but it doesn't seem to be needed below the `let` variable definitions.

If the `p2` variable isn't needed, then both of it's assignment lines:
``` elisp
(setq p2 (point)))
        (setq p2 (point))))
```
could be removed, and the `progn` in the second if statement won't be needed either.

With the suggestions above, the function would look like this:
```elisp
(defun spacemacs/select-current-block ()
  "Select the current block of text between blank lines."
  (interactive)
  (let (p1)
    (if (re-search-backward "\n[ \t]*\n" nil "move")
        (progn (re-search-forward "\n[ \t]*\n")
               (setq p1 (point)))
      (setq p1 (point)))
    (if (re-search-forward "\n[ \t]*\n" nil "move")
        (re-search-backward "\n[ \t]*\n"))
    (set-mark p1)))
```

The `spacemacs/select-current-block` function is bound to left mouse button double clicking to the left of a text block on the line number margin <kbd>SPC t n</kbd>. That will select everything between the blank lines above and below. The function can also be called with the cursor in a text block from HELM M-x:
- press: <kbd>SPC SPC</kbd>
- type: `select cur`
- press <kbd>RET</kbd>

It seems to work the same way after removing the suggested lines and evaluating the function.

#### System Info :computer:
- OS: windows-nt
- Emacs: 25.1.1
- Spacemacs: 0.200.5
- Spacemacs branch: develop (rev. 5d73a50)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
‘‘‘elisp
(auto-completion autohotkey emacs-lisp git helm html org)
‘‘‘

duianto:
Resolves #8031
The variable p2 was defined and assigned, but never used, and two instances of progn could be removed as well.

### DIFF
--- a/layers/+distributions/spacemacs-base/funcs.el
+++ b/layers/+distributions/spacemacs-base/funcs.el
@@ -867,16 +867,12 @@ With negative N, comment out original line and use the absolute value."
 (defun spacemacs/select-current-block ()
   "Select the current block of text between blank lines."
   (interactive)
-  (let (p1 p2)
-    (progn
-      (if (re-search-backward "\n[ \t]*\n" nil "move")
-          (progn (re-search-forward "\n[ \t]*\n")
-                 (setq p1 (point)))
-        (setq p1 (point)))
-      (if (re-search-forward "\n[ \t]*\n" nil "move")
-          (progn (re-search-backward "\n[ \t]*\n")
-                 (setq p2 (point)))
-        (setq p2 (point))))
+  (let (p1)
+    (when (re-search-backward "\n[ \t]*\n" nil "move")
+      (re-search-forward "\n[ \t]*\n"))
+    (setq p1 (point))
+    (if (re-search-forward "\n[ \t]*\n" nil "move")
+        (re-search-backward "\n[ \t]*\n"))
     (set-mark p1)))
 
 ;; END linum mouse helpers


layers/+distributions/spacemacs-base/funcs.el
```diff
@@ -867,16 +867,13 @@ With negative N, comment out original line and use the absolute value."
...
-        (setq p1 (point)))
-      (if (re-search-forward "\n[ \t]*\n" nil "move")
-          (progn (re-search-backward "\n[ \t]*\n")
-                 (setq p2 (point)))
-        (setq p2 (point))))
+  (let (p1)
+    (if (re-search-backward "\n[ \t]*\n" nil "move")
+        (progn (re-search-forward "\n[ \t]*\n")
+               (setq p1 (point)))
+      (setq p1 (point)))
```
bmag: Since `(setq p1 (point))` happens anyway, we can move it outside the `if`, and turn the `if` into a `when`.

duianto: That's what first brought me to this issue, i thought that since it's in both the "if" and "else" clause then i only needed to be called once, but i couldn't get it to work. Now i see how it should be done.

Is this correct?
``` elisp
    (when (re-search-backward "\n[ \t]*\n" nil "move")
      (re-search-forward "\n[ \t]*\n"))
    (setq p1 (point))

```
It seems to work the same way after that improvement.

d12frosted: @duianto LGTM.

bmag: Yes, that's what I meant

duianto:
Damn, while trying to merge the new committed changes into one, and making mistakes, i then tried to delete the local branch but i deleted the remote branch haha. Will it get fixed if i recreate a branch with the same name?

deb0ch:
I guess you can try, worse case scenario you commit is still accessible https://github.com/syl20bnr/spacemacs/pull/8034/commits/cfad7de1a33e83eb76e31ebb1b4b941c92a2aa7b so no need to worry :stuck_out_tongue:

duianto:
It didn't work, possibly because i changed to the corrected code, and updated the commit description so it didn't get the same hash.

duianto:
I managed to find the original commit: cfad7de that deb0ch suggested, and made a branch with the same name, then force pushed it to my fork. That made the "Closed with unmerged commits" message appeared, does it mean that this PR can't be amended anymore?

duianto:
Hmm now that i pushed bmag's corrections, then the closed message disappeared, and it looks correct in the "reduce-code" branch on my fork, but here it still shows the original changes.

duianto:
Maybe i shouldn't have force pushed, below the comment box where this was written, there's a "Reopen and comment" button but it's grayed out and the mouse over tooltip says: "The reduce-code branch was force-pushed or recreated.", should i just have pushed with out the -F Force (--force) switch? When i updated a previous pull request, then it was suggested that i should use the -F Force switch, but that might have been to solve another issue.

duianto:
I drove into a ditch but it seems like i managed to get back to where i started 😅. Will it work to push the squashed commits without the -F Force (--force) switch?

duianto:
Experimented with push, both with or without force and force was needed to replace the previous commit with this squashed one. Now it seems to be correct.

d12frosted:

PR #8037, opened 2016-12-22
Gitignore emojis folder

joshmedeski:
The Emoji layer downloads image files in a root directory called `emojis`. This
folder doesn't need to be tracked in version control.

Issue #5023, opened 2016-02-10
Feature request: add emojify to emoji layer


jb55:
https://github.com/iqbalansari/emacs-emojify

maybe default off...

Stebalien:
Could this be reopened? Personally, I'd like this feature (defaulted to on).

iqbalansari:
Hi,

Is there something I can do, to help include emojify in Spacemacs?

Miciah:
#8042 added the emojify package to the emoji layer; is anything else needed to close out this issue?

PR #8042, opened 2016-12-23
setup emojify cache directory

d12frosted:
Following #8037. 

It turns out that `slack` package pulls `emojify` as it's dependency. I am not sure that it's a proper place to configure `emojify`. That's why I am opening PR 😸 

Any thoughts?


Stebalien:
It could be added to the emoji layer (#5023).

syl20bnr:
Thank you ! 👍 
I made the slack layer dependent on emoji layer and moved this config bit to emoji layer.

PR #8044, opened 2016-12-23
replace 9 buffer-to-window entries with one 1..9

### LINKED ISSUE #8038: Reduce the 9 buffer-to-window SPC b entries to one 1..9, like it was done with SPC window 0..9
Maybe someone is working on this already, but the <kbd>SPC b</kbd> entries: 
1 -> buffer-to-window-1
...
9 -> buffer-to-window-9

probably should be reduced to a single line that looks something like this: 
1..9 -> buffer-to-window 1..9

The same way as the <kbd>SPC</kbd> window entries:
0  → window 0
..
9  → window 9

recently were reduced to:
0..9 -> window 0..9

as mentioned in this issue: https://github.com/syl20bnr/spacemacs/issues/7582, in fact the <kbd>SPC b</kbd> entries were mentioned in that issue.

The buffer-to-window-(nr) entries seem to be defined here:
https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Bdistributions/spacemacs-base/keybindings.el#L135

I don't know how it should be done. So someone with more knowledge is going to have to do it.

duianto:
Resolves #8038
This replaces the <kbd>SPC b</kbd> buffer-to-window 1 to 9 entries, with just one 1..9 > buffer to window 1..9.

### DIFF
--- a/layers/+distributions/spacemacs-bootstrap/packages.el
+++ b/layers/+distributions/spacemacs-bootstrap/packages.el
@@ -312,6 +312,10 @@
         which-key-replacement-alist)
   (push '((nil . "select-window-[1-9]") . t) which-key-replacement-alist)
 
+  (push '(("\\(.*\\) 1" . "buffer-to-window-1") . ("\\1 1..9" . "buffer to window 1..9"))
+        which-key-replacement-alist)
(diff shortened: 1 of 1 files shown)


bmag:

PR #8046, opened 2016-12-24
add sort-lines arg, reverse func and key binding

### LINKED ISSUE #7978: spacemacs/sort-lines, optional argument, sort-reverse function and key binding
#### 1. A visual block selection doesn't sort the selected lines.
The whole buffer gets sorted when a <kbd>C-v</kbd> visual block selection is active.

#### Reason
`spacemacs/sort-lines` checks `(region-active-p)`, if it's: `t`, then the region gets sorted, but if it's: `nil` then the whole buffer gets sorted.

When a visual block selection is active then `(region-active-p)` returns: `nil`

#### Possible solution
Change `(region-active-p)` to `(evil-visual-state-p)`, because it seems to return: `t`, for these selection commands (maybe there are more selection commands):
<kbd>v</kbd> (evil-visual-char)
<kbd>S-v</kbd> (evil-visual-line)
<kbd>C-v</kbd> (evil-visual-block)
<kbd>C-SPC</kbd> (set-mark-command)
<kbd>C-x SPC</kbd> (rectangle-mark-mode)

#### Before:
```
  (let ((beg (if (region-active-p) (region-beginning) (point-min)))
        (end (if (region-active-p) (region-end) (point-max)))
```
#### After:
```
  (let ((beg (if (evil-visual-state-p) (region-beginning) (point-min)))
        (end (if (evil-visual-state-p) (region-end) (point-max))))
```
#### Possible `(evil-visual-state-p)` issue
Will `(evil-visual-state-p)` work for people that use the Emacs editing style?

If not then maybe `(evil-mode)` could be checked and if it's true, then check: `(evil-visual-state-p)` otherwise check `(region-active-p)`.

#### 2. There are two instances when the cursor doesn't get restored correctly.
- When a <kbd>S-v</kbd> line selection is active then the cursor ends up on the first selected line, even if the selection was made from top to bottom.

- When there's no active selection and the whole buffer gets sorted, then the cursor ends up on the buffers first line.

#### A. Add an optional argument to `spacemacs/sort-lines`
`spacemacs/sort-lines` doesn't accept an argument, it just passes `nil` to the `sort-lines` command. A non-nil argument makes it possible to sorts the lines in reverse order with for example: <kbd>SPC u SPC x l s</kbd>.

#### Before:
```
(defun spacemacs/sort-lines ()
  "Sort lines in region or current buffer"
...
(sort-lines nil beg end)
```
#### After:
```
(defun spacemacs/sort-lines (&optional arg)
  "Sort lines in a region or the current buffer.
A non-nil argument sorts in reverse order."
...
(sort-lines arg beg end)
```

#### B. Add a function and key binding that sorts in the opposite direction
```
(defun spacemacs/sort-lines-reverse (&optional arg)
  "Sort lines in reverse order, in a region or the current buffer.
A non-nil argument sorts in non-reverse order."
  (spacemacs/sort-lines (if arg nil "non-reverse")))
```

"non-reverse order" sounds a bit strange, maybe the function should be written with "ascending" and "descending":

```
(defun spacemacs/sort-lines-descending (&optional arg)
    "Sort lines in descending order, in a region or the current buffer.
A non-nil argument sorts in ascending order."
  (spacemacs/sort-lines (if arg nil "descending")))
```

It should probably be bound to: <kbd>SPC x l S</kbd>, since a lot of commands work in the opposite direction when the uppercase key is pressed.

If "ascending" and "descending" order are better choices, then maybe the `spacemacs/sort-lines` function also should be renamed to `spacemacs/sort-lines-ascending`. At least it's function comment will need to be updated.

#### I. Should a message be shown when the sorting is done?
`(message "Lines sorted%s" (if arg " in reverse order." "."))`

> Lines sorted.
> Lines sorted in reverse order.

or
`(message "Lines sorted in %s order." (if arg "descending" "ascending"))`

> Lines sorted in ascending order.
> Lines sorted in descending order.

or something better.

#### II. Should the lines be counted?
If the lines in the region or buffer are counted, then the number of sorted lines could be shown: "12 lines sorted in reverse (descending) order.". And if there's only one line then a message could be shown: "Sorting requires two or more lines."

#### System Info :computer:
- OS: windows-nt
- Emacs: 25.1.1
- Spacemacs: 0.200.5
- Spacemacs branch: develop (rev. 6288b28)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
‘‘‘elisp
(auto-completion autohotkey emacs-lisp git helm html org)
‘‘‘

duianto:
Resolves #7978

### DIFF
--- a/layers/+distributions/spacemacs-base/funcs.el
+++ b/layers/+distributions/spacemacs-base/funcs.el
@@ -828,12 +828,19 @@ With negative N, comment out original line and use the absolute value."
         (while (re-search-forward "^\\(.*\n\\)\\1+" end t)
           (replace-match "\\1"))))))
 
-(defun spacemacs/sort-lines ()
-  "Sort lines in region or current buffer"
+(defun spacemacs/sort-lines (&optional reverse)
+  "Sort lines in a region or the current buffer.
+A non-nil argument sorts in reverse order."
+  (interactive "P")
+  (let* ((region-active (or (region-active-p) (evil-visual-state-p)))
+         (beg (if region-active (region-beginning) (point-min)))
+         (end (if region-active (region-end) (point-max))))
+    (sort-lines reverse beg end)))
+
+(defun spacemacs/sort-lines-reverse ()
+  "Sort lines in reverse order, in a region or the current buffer."
   (interactive)
-  (let ((beg (if (region-active-p) (region-beginning) (point-min)))
-        (end (if (region-active-p) (region-end) (point-max))))
-    (sort-lines nil beg end)))
+  (spacemacs/sort-lines -1))
 
 ;; BEGIN linum mouse helpers
 
--- a/layers/+distributions/spacemacs-base/keybindings.el
+++ b/layers/+distributions/spacemacs-base/keybindings.el
@@ -436,6 +436,7 @@
   "xjr" 'set-justification-right
   "xld" 'spacemacs/duplicate-line-or-region
   "xls" 'spacemacs/sort-lines
+  "xlS" 'spacemacs/sort-lines-reverse
   "xlu" 'spacemacs/uniquify-lines
   "xtc" 'transpose-chars
   "xtl" 'transpose-lines


bmag:

PR #8048, opened 2016-12-24
add sort-lines-by-column, -reverse funcs and keys

duianto:
Add two functions and key bindings:
<kbd>SPC x l c</kbd> `spacemacs/sort-lines-by-column` with an optional reverse argument,
<kbd>SPC x l C</kbd> `spacemacs/sort-lines-by-column-reverse`

### DIFF
--- a/layers/+distributions/spacemacs-base/funcs.el
+++ b/layers/+distributions/spacemacs-base/funcs.el
@@ -835,6 +835,20 @@ With negative N, comment out original line and use the absolute value."
         (end (if (region-active-p) (region-end) (point-max))))
     (sort-lines nil beg end)))
 
+(defun spacemacs/sort-lines-by-column (&optional reverse)
+  "Sort lines by the selected column.
+A non-nil argument sorts in reverse order."
+  (interactive "P")
+  (let* ((region-active (or (region-active-p) (evil-visual-state-p)))
+         (beg (if region-active (region-beginning) (point-min)))
+         (end (if region-active (region-end) (point-max))))
+    (sort-columns reverse beg end)))
+
+(defun spacemacs/sort-lines-by-column-reverse ()
+  "Sort lines by the selected column in reverse order."
+  (interactive)
+  (spacemacs/sort-columns -1))
+
 ;; BEGIN linum mouse helpers
 
 (defvar spacemacs-linum-mdown-line nil
--- a/layers/+distributions/spacemacs-base/keybindings.el
+++ b/layers/+distributions/spacemacs-base/keybindings.el
@@ -434,6 +434,8 @@
   "xjl" 'set-justification-left
   "xjn" 'set-justification-none
   "xjr" 'set-justification-right
+  "xlc" 'spacemacs/sort-lines-by-column
+  "xlC" 'spacemacs/sort-lines-by-column-reverse
   "xld" 'spacemacs/duplicate-line-or-region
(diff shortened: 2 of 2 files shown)


bmag:

duianto:
I made a mistake in the `spacemacs/sort-lines-by-column-reverse` function:
the last line calls a function with the wrong name: `  (spacemacs/sort-columns -1))`

When one tries to sort lines by column in reverse then the messages buffer shows:
`spacemacs/sort-lines-by-column-reverse: Symbol’s function definition is void: spacemacs/sort-columns`

The correct function looks like this:
```elisp
(defun spacemacs/sort-lines-by-column-reverse ()
  "Sort lines by the selected column in reverse order."
  (interactive)
  (spacemacs/sort-lines-by-column -1))
```
From now on i'm going to have to test more thoroughly.

Can i still push the fix to this PR?

bmag:
You can make a new PR

bmag:
Actually nevermind, I committed the fix myself in https://github.com/syl20bnr/spacemacs/commit/d2168806b634deef82c2d6c6948f5bcd9d034ed0

syl20bnr:
@duianto @bmag How does this function work ? How one can select the column ? Does it need block selection ? Can you update the docstring of the function if necessary ?

duianto:
Your right there isn't enough information on how it works.
Before:
```
  "Sort lines by the selected column.
A non-nil argument sorts in reverse order."
```
Does this seem like a good explanation?
After:
```
  "Sort lines by the selected column.
A non-nil argument sorts in reverse order.

A column selection is made by drawing a vertical selection,
(the start and end of the selection are in the same column)
with either of these commands and key bindings:
evil-visual-char,    v
evil-visual-block,   C-v
set-mark-command,    C-SPC
rectangle-mark-mode, C-x SPC

It doesn't work with the line selection command:
evil-visual-line,    S-v"
```

#### Questions:
- Am i correct in assuming that the commands and keys description also should be applied to the `spacemacs/sort-lines-by-column-reverse` function. Someone will probably look at the reverse functions help documentation first, then they would have to search for the main function to find the description.

- Is it possible to insert the current keys that are bound to the selection commands? I don't know if a lot of people rebind them, but then the new keys would be shown next to the commands. 
Or is it fine to just "hard code" them?

nixmaniack:
> Is it possible to insert the current keys that are bound to the selection commands?

Yes, you can use `\\[your-command]` in function doc-string which would substitute key-binding for `some-command` when viewing doc.

syl20bnr:
@duianto 

> Does this seem like a good explanation?

Yes this is better. Not sure you need to include the key bindings though.
Mentioning visual block selection seems sufficient to me:

```
 "Sort lines by the selected column using visual block selection.
A non-nil argument sorts in reverse order."
```

duianto:
That simpler and to the point. 👍

#### Question
Is it understood that a "visual block selection" also means that it works in `evil-emacs-state`, with the `rectangle-mark-mode` <kbd>C-x SPC</kbd> command.

Or should "rectangle" also be mentioned like this:
```
  "Sort lines by the selected column, using a visual block/rectangle selection.
A non-nil argument sorts in reverse order."
```

syl20bnr:
> using a visual block/rectangle selection.

That's better 👍

duianto:
The previous PR has been updated with the new docstrings:
Update sort lines by column, docstrings, how to #8196

PR #8049, opened 2016-12-24
update uniquify-lines, uniquify a block selection

duianto:
An active evil-visual-block selection, uniquified the whole buffer, this
checks either region-active-p or evil-visual-state-p, now it sorts the
selected lines.

The comment was also updated with "a" before "region" and "the" before
"current buffer".


bmag:

PR #7940, opened 2016-12-11
use winum.el for window numbers

deb0ch:
[winum.el](https://github.com/deb0ch/winum.el) is a rewrite of [window-numbering.el](https://github.com/nschum/window-numbering.el) that heavily extends it.

Among other things, it allows for sharing numbers across multiple frames, allowing to switch windows more smoothly when multi-screen.

It also brings a lot of new features and adapts the package to how we use it in Spacemacs.

See https://github.com/nschum/window-numbering.el/issues/16.

This should be merged only after https://github.com/TheBB/spaceline/pull/124 to make it display nicely in the mode-line.

It is also [on its way to Melpa](https://github.com/melpa/melpa/pull/4451).

**EDIT:**
- [X] winum [merged in Melpa](https://github.com/melpa/melpa/pull/4451)
- [X] https://github.com/TheBB/spaceline/pull/124 merged in Spaceline

PR ready for merge :100:

PR #8056, opened 2016-12-26
fix neotree window number when using menu-bar-mode in terminal

deb0ch:
when using `menu-bar-mode` in terminal, `(window-at 0 0)` does not return the NeoTree window, where `frame-first-window` does.


deb0ch:
@bmag Closing this as it overlaps with https://github.com/syl20bnr/spacemacs/pull/7940

PR #8059, opened 2016-12-26
Fix a link on beginner's tutorial

lpenz:
Fix a 404 link in doc/BEGINNERS_TUTORIAL.org


Nikolaiam:
Ah, seems like I spotted the same thing in #8093 :) great! Might be we'll have to wait and see whether or not my PR is going to be accepted as ours would conflict if both are merged?

lpenz:
I think they would :)
It's a simple fix though, on your side - on my side there's no need for a fix, we'll just drop the PR.

d12frosted:

To be honest I like the full url much more 😸 As we prefer our users to read it from Emacs 😸

lpenz:
:smiley: good to know! :+1:

PR #8065, opened 2016-12-28
Fix ivy wgrep when used with rg

aaronjensen:
wgrep does not know how to handle column numbers, which the `--vimgrep` flag
adds.

To reproduce the original issue:

* `SPC /`
* Search for something that has results
* `C-c C-o`
* `C-x C-q`
* Make a change
* `C-c C-e`

Note that the changes fail to apply. wgrep does not understand the results because of the column numbers.

### DIFF
--- a/layers/+completion/helm/packages.el
+++ b/layers/+completion/helm/packages.el
@@ -245,7 +245,7 @@ Search for a search tool in the order provided by `dotspacemacs-search-tools'."
       (defun spacemacs/helm-files-do-rg (&optional dir)
         "Search in files with `rg'."
         (interactive)
-        (let ((helm-ag-base-command "rg --smart-case --no-heading --vimgrep"))
+        (let ((helm-ag-base-command "rg --smart-case --no-heading --color never"))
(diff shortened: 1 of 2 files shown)


joehillen:
Can someone please merge this?

aaronjensen:
@d12frosted ping

bmag:
See alternative: #8407

bmag:

Swoorup:
Please revert this change. Without `vimgrep` parameter, rg support is broken. 

I can search and list files through helm, but opening them, results in emacs trying to open the whole search result line as file.

bmag:
@Swoorup thanks for letting us know. Apparently, Ripgrep doesn't show line numbers by default when called from Spacemacs in Windows, but does show them in Linux (according to my limited testing). The fix was to pass the `-n` flag that enables line numbers explicitly.

@Swoorup please confirm that #8475 fixes the issue with helm
@aaronjensen @joehillen please test that #8475 doesn't break ivy-wgrep

PR #8066, opened 2016-12-28
Add binding for jump to def'n other window

quicknir:
Adds a binding for goto-defn-other-window at gD, normal binding is gd. gD is not taken and is a natural choice, layers often use ,gg and ,gG similarly.

### DIFF
--- a/layers/+distributions/spacemacs-bootstrap/packages.el
+++ b/layers/+distributions/spacemacs-bootstrap/packages.el
@@ -135,6 +135,7 @@
 
   (define-key evil-normal-state-map (kbd "K") 'spacemacs/evil-smart-doc-lookup)
   (define-key evil-normal-state-map (kbd "gd") 'spacemacs/jump-to-definition)
+  (define-key evil-normal-state-map (kbd "gD") 'spacemacs/jump-to-definition-other-window)
 
   ;; scrolling transient state
   (spacemacs|define-transient-state scroll


nixmaniack:
Isn't it already there?

deb0ch:
If it is already used similarly in other layers that might be a different story, but a convenient way would also be to use the prefix argument to have it opened in another window, like in https://github.com/syl20bnr/spacemacs/pull/7943 for instance.

The two solutions don't have to be mutually exclusive, but if they were it would also keep a key binding free.

I don't know which approach would be better, it is just a suggestion.

quicknir:
I don't have a strong feeling about the prefix argument thing, other than to say it's quite a bit less ergonomic IMHO to use a prefix argument than to capitalize. I don't think we should worry too much about keeping a key binding free; this is something that's both consistent and incredibly useful (I'd say my gd/gD distribution would be about 50-50), keeping the binding free for some other function that may or may not come along later does not seem like a strong argument (of course, if you have something specific in mind, that's different).

deb0ch:
Mmmhhhh I don't have a specific one in mind, your point totally makes sense, this binding is definitely useful and we shouldn't keep one free "just in case maybe in the future something could go there".

I still feel like a prefix arg would be useful, if you don't think it is enough for you to modify your PR maybe will I PR something for that in the future and let's see if that comes through :smile_cat:

d12frosted:
I use prefix arguments a lot and I think it would be nice to have version of `spacemacs/jump-to-definition` that accepts prefix argument. Though for vim editing style users new key binding might be preferable. 

Do you guys use prefix arguments with vim editing style? 😸

deb0ch:
> Do you guys use prefix arguments with vim editing style? :smile_cat:

Well, I do :smile_cat:

d12frosted:

Looks very useful for vim editing style. Just tried it 😸

quicknir:
@d12frosted Thanks!

PR #8067, opened 2016-12-28
Add manual completion function to ycmd layer

quicknir:
Adds a function that is useful for those who prefer for ycmd completion to be triggered manually, as well as some explanation in the docs.

### DIFF
--- a/layers/+tools/ycmd/README.org
+++ b/layers/+tools/ycmd/README.org
@@ -9,6 +9,7 @@
  - [[#configuration][Configuration]]
    - [[#activating-ycmd-in-a-major-mode][Activating ycmd in a major mode]]
    - [[#getting-the-compilation-flags][Getting the compilation flags]]
+ - [[#functions][Functions]]
  - [[#key-bindings][Key Bindings]]
 
 * Description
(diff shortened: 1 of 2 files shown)


deb0ch:
Hey, this looks great ! I've wanted sth like that for quite some time now, just been using other function almost like this but not quite 😸 

2 suggestions:

- It would be convenient to have a key binding for it, maybe under the major-mode prefix.
- I think that this function could be described in one short sentence in the docs. If the docs get too long, nobody will read it and it will be not practical for those who will.

😸

quicknir:
IMHO ycmd should not do anything with regards to major modes. ycmd offers completions for many languages, and some languages have many completers available, of which ycmd is only one. It's up to the language layer whether to choose to use ycmd, not use ycmd (e.g. python), use ycmd based on a user option, use ycmd only if ycmd package has been loaded, etc.

Actually right now the docs say that the layer adds keybindings, and adds ycmd to the company backends, for c/c++. Neither of these statements is correct, and the docs should be updated to reflect that. And this is a good thing, btw, the c/c++ layer should decide whether and under what circumstances either of these are enacted.

A binding could be added in the c/c++ layer, but again, not everyone is using ycmd, so it would need to switch based on usage in some way. To top it all off, spacemacs major modes put all their bindings under SPC m. Such a binding for manual completion, which is generally accessed from insert mode, doesn't seem useful.

I can't say I have ever seen a piece of software that I thought was over documented. The docs now are quite short, and I've answered many questions in the gitter about ycmd. Getting ycmd to work in practice is quite a bit harder than most other layers; everyone that I know has encountered their share of edge cases and issues with it. The more we can inform people in the docs, the better their chances of solving their own problems. That's why I use a few sentences to explain what this function is doing, instead of a one liner. The docs would have to be way, way longer, and poorly organized to boot, for what you're describing to be an issue.

deb0ch:
Major-mode leader or not, I just think that some key binding to use the function would be useful.

Code navigation using ycmd, gtags or other tools that also work for more languages is already under major-mode leader key, so that's why I thought of it for the key binding.

beta1440:
@quicknir it sounds like you could make a few pull requests to improve the `c-c++` documentation.

syl20bnr:
Thank you ! 👍 
Beware!! I also merged https://github.com/syl20bnr/spacemacs/pull/9829 which renamed the function but... I just realize that you are the author of both PRs 😂  Two commits for the price of one 😄 

PR #8068, opened 2016-12-28
Add ycmd-eldoc to ycmd layer

quicknir:
As the commit says, simply adding ycmd-eldoc package to the ycmd layer.

### DIFF
--- a/layers/+tools/ycmd/README.org
+++ b/layers/+tools/ycmd/README.org
@@ -13,7 +13,8 @@
 
 * Description
 
-This layer adds [[https://github.com/abingham/emacs-ycmd][emacs-ycmd]] support.
+This layer adds [[https://github.com/abingham/emacs-ycmd][emacs-ycmd]] support, including its integrations with company,
+flycheck, and eldoc.
 
 * Install
 ** Layer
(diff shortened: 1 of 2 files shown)


deb0ch:
Just tried it, and I end up with an error because the package `ycmd-eldoc` does not exist in the *elpa archives.

Is it like this that it is supposed to be installed ?

```
Package ycmd-eldoc is unavailable. Is the package name misspelled?
```

Everytime I tried to use `ycmd-eldoc-setup` in the past to get the same information as `ycmd-get-type` in the echo area with eldoc, it just didn't do anything, so I guess that I was missing something like here, the installation of a package.

Just that here that way of installing the package doesn't seem to work, at least for me.

bmag:
Yeah, looking at ycmd's [recipe](https://github.com/melpa/melpa/blob/master/recipes/ycmd), I see that `ycmd-eldoc.el` is part of the `ycmd` package, not a separate package.

quicknir:
My mistake, I'll fix this.

quicknir:
@deb0ch @bmag It should be fixed now.

Note that in C++ particularly, many many things are (strangely) undocumented. If you activate ycmd in the python layer, and hit `C-h`, you should see it.

deb0ch:
@d12frosted @quicknir does this work for you guys ? I don't know why, but my eldoc-mode is not working at all.

It did when I had `semantic` layer enabled, but never worked again when I deactivated it (using Emacs 24.5.1).

d12frosted:
@deb0ch `eldoc-mode` works well for me without `semantic` layer. I remember I had some problems with it though. 

---

@quicknir 

quicknir:
@d12frosted Thank you, twice today :-)

PR #5004, opened 2016-02-08
elm: fix flycheck not working on subdirectories

AlejandroCatalina:
As pointed in #3763, this should fix it.

Credits to @bharrisau

PR #8075, opened 2016-12-29
Fix elm to not switch to root directory

kwrooijen:
Issue: `default-directory` gets changed to the project's root when you open an Elm file. This is annoying because whenever you try perform an action in the current directory it will instead change to the project's root  (e.g. `M-x find-file`).

I don't know why this was implemented like this, there are probably some function that need to be executed from the project's root. However changing `default-directory` globally is a bad solution and should be scoped within the function instead.

EDIT: Just found this this PR https://github.com/syl20bnr/spacemacs/pull/5004/files

However flycheck still works for me so I'm not sure what the issue is.

### DIFF
--- a/layers/+lang/elm/funcs.el
+++ b/layers/+lang/elm/funcs.el
@@ -12,9 +12,6 @@
 
 ;; elm-mode
 
-(defun spacemacs//elm-find-root ()
-  (setq default-directory (elm--find-dependency-file-path)))
-
 (defun spacemacs/elm-compile-buffer-output ()
   (interactive)
   (let* ((fname (format "%s.js" (downcase (file-name-base (buffer-file-name))))))
--- a/layers/+lang/elm/packages.el
+++ b/layers/+lang/elm/packages.el
@@ -25,8 +25,7 @@
   (push 'company-elm company-backends-elm-mode))
 
 (defun elm/post-init-flycheck ()
-  (add-hook 'elm-mode-hook 'flycheck-mode)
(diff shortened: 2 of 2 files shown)


syl20bnr:
Thank you ! 👍 
Let's see if it triggers regressions.
I agree with you that changing `default-directory` is not a good fix.

niklas:
For documentation purposes:

I ran into a variation of that faulty behavior.

I had not set `elm-mode`'s `elm-package-json` correctly according to the name change that happened on the elm upgrade to `0.19` (`elm-package.json` was migrated to `elm.json`. After removing the legacy file during cleanup, this caused `elm--find-dependency-file-path` to skip my elm root and my project root, and only stopped at my home directory, where by some coincidence a spurious `elm-package.json` was lying around.
This caused `default-directory` to be set to `"~/"`, which in turn broke `projectile` and `magit` (and maybe other packages), the former offering me all files from my home directory on `SPC p f` and the latter not finding a git repository.
As a cherry on top, I only saw a `(wrong-type-argument arrayp nil)` in the messages buffer when opening any elm file, which I could trace with `(toggle-debug-on-error)` to `vc-git-mode-line-string` trying (and failing) to populate the status line with the git status of that file.

Full trace: https://pastebin.com/R2masjtq
related: jcollard/elm-mode#141

Fixed by customizing `elm-package-json` to `"elm.json"`.

Credits and thanks go to @Compro-Prasad for holding my hand in the Gitter chat.

PR #8076, opened 2016-12-29
finance: Add keybindings for ledger-reconcile-mode

jgertm:
This commit introduces some keybindings which should prove useful during account reconciliation.

### DIFF
--- a/layers/+tools/finance/packages.el
+++ b/layers/+tools/finance/packages.el
@@ -46,6 +46,7 @@
       (spacemacs/set-leader-keys-for-major-mode 'ledger-reconcile-mode
         "," 'ledger-reconcile-toggle
         "a" 'ledger-reconcile-add
+        "q" 'ledger-reconcile-quit
         "t" 'ledger-reconcile-change-target
         "RET" 'ledger-reconcile-finish)
       ;; temporary hack to work-around an issue with evil-define-key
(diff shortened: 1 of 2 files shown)


layers/+tools/finance/packages.el
```diff
@@ -43,6 +43,12 @@
          "t" 'ledger-insert-effective-date
          "y" 'ledger-set-year
          "RET" 'ledger-set-month)
+      (spacemacs/set-leader-keys-for-major-mode 'ledger-reconcile-mode
+        (or dotspacemacs-major-mode-leader-key ",") 'ledger-reconcile-toggle
```
d12frosted: I am not sure what you are doing here 😸 Since `dotspacemacs-major-mode-leader-key` is declared via `defvar` and it's default value is `","` I can't understand why the check is needed. You know, if someone sets `dotspacemacs-major-mode-leader-key` to `nil`, then it's not our problem 😸

jgertm: You are correct. I copied it from the ```org``` layer FWIW.

d12frosted: Oh... thanks for finding it 😸

d12frosted:

d12frosted:
P. S. forgot to tell you that I have squashed your commits.

d12frosted:
Ok, sorry for confusion. Actually that `or` expression made sense. I've reverted to your original code in 0d6c378.

PR #8080, opened 2016-12-30
org: bind org-refile from 'R' to 'r'

nixmaniack:
Why?
 - 'r' is free binding - not used anywhere for org related modes
 - refiling is quite common activity if you use org heavily, so moving it for
 easy access makes sense


TheBB:
Thanks. I had to do some conflict editing. Please check. Cherry-picked in develop. You can safely delete your branch.

nixmaniack:
Thanks! Working as expected.

PR #8081, opened 2016-12-30
Update README.org

kindlychung:
Give a clearer example for setting autocompletion backend.


TheBB:
I appreciate the effort but the example you give is malformed. Layers are declared multiple times and the last haskell entry is inside `(auto-completion ...)`. If you're still interested, please resubmit with a corrected version.

PR #6904, opened 2016-08-24
Don’t break asynchronous jumpers (fix #6898)

TheBB:

PR #8089, opened 2017-01-01
Fix issue with jump handler in omnisharp-mode where the handler is async

razzmatazz:
This fixes an issue where default jump handler, like dump-jump-go runs
in sync and omnisharp-go-to-definition is not able to actually go to
definition until it completes thus slowing down the completion process.

A similar issue (for C++) was fixed in syl20bnr/spacemacs#6904


d12frosted:

PR #8092, opened 2017-01-02
select neotree window with backtick key

### LINKED ISSUE #7826: [Improvement] Bind NeoTree to SPC `
When active, NeoTree window is currently bound to `SPC 0` which is on the far right of the keyboard and nowhere close to `SPC 1`.
In my opinion it could be bound to ``SPC ` `` which is just beside `1` and just feels much more natural to me as a newcomer.

duianto:
Resolves #7826

Bind \` (backtick) to the select-window-0 command, it selects
an open neotree window.

Hide the \` -> select-window-0 entry in the SPC panel.

Add \` to the SPC window entry: `,0..9 -> window 0..9.

Update the SPC w .  window manipulation transient state panel:
Select column
Add \` to  [`,0..9] window 0..9.

### DIFF
--- a/layers/+distributions/spacemacs-base/keybindings.el
+++ b/layers/+distributions/spacemacs-base/keybindings.el
@@ -531,18 +531,19 @@
 (spacemacs|define-transient-state window-manipulation
   :title "Window Manipulation Transient State"
   :doc (concat "
- Select^^^^              Move^^^^              Split^^                Resize^^                     Other^^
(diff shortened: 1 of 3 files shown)


duianto:
I just read about making atomic commits: https://www.freshconsulting.com/atomic-commits/, and realized that i shouldn't have tried to save one PR by combining the window manipulation transient state formatting with this neotree backtick PR. I'll remove the formatting and submit them as a separate PR.

duianto:
A recent commit to the develop branch is using the <kbd>SPC `</kbd> key binding, so this PR might not have a future.

PR #8096, opened 2017-01-02
Fix eldoc initialization in typescript layer.

Stebalien:


Stebalien:
@syl20bnr fixes a bug introduced in https://github.com/syl20bnr/spacemacs/commit/bdd4e6c23f2e8fcb3eb7b7955c5d1201bcf1263a

PR #8099, opened 2017-01-03
Update buffer transient state formatting

duianto:
Changed the buffer transient states formatting so that it matches
most of the other transient states.

Frontend
Reduced spacing:
- Between the keys and descriptions, from 2 to 1.
- Between both columns, from 4 to 2.

Backend
The first column didn't need the last two "^" (caret) characters.
Changed the last line so that "^^^^" starts below the first column.


duianto:
There's a discussion here: 
Update window manipulation transient state formatting #8102

if two dots `..` should be used to indicate a sequence, so this PR might need to be updated.

d12frosted:

PR #8101, opened 2017-01-03
Fix pylookup.py in python layer

nightuser:
Make `pylookup.py` more compatible with both Python 2 & 3. Module `htmllib` is deprecated since 2.6 and doesn't exist in 3.

Tested with Python 2.7.13 and 3.6.0.

### DIFF
--- a/layers/+lang/python/local/pylookup/pylookup.py
+++ b/layers/+lang/python/local/pylookup/pylookup.py
@@ -11,26 +11,32 @@ Blais.
 
 """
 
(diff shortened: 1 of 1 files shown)


syl20bnr:
Thank you for the PR.
Recent changes have been made to pylookup in develop, can you test if this PR is still needed ? If yes, can you rebase it against develop ? Thank you!!

syl20bnr:
@nightuser is this PR still needed ?

PR #8102, opened 2017-01-03
Update window manipulation transient state formatting

duianto:
Update the formatting of the SPC w .  window manipulation
transient state panel.

Select column
Change "[0-9] window N", to "[0..9] window 0..9",
so that it matches the SPC "0..9 -> window 0..9" and
SPC b "1..9 -> buffer to window 1..9" combined entries.

Split column
Add "i" to the end of "vert", then it'll have the same number
of characters as "horiz", and the text will be aligned.

Resize column
Shorten "horizontally" and "vertically" to "horiz" and "verti",
it reduces the panels width slightly.

Other column
Shorten the horizontal line below the title, so that it matches
the longest key description.

Move [q] quit to the bottom of the column.

All columns
Equalize the horizontal spacing between the columns,
from: 2, 2, 3, 3, to 2 between all columns.

### DIFF
--- a/layers/+distributions/spacemacs-base/keybindings.el
+++ b/layers/+distributions/spacemacs-base/keybindings.el
@@ -531,16 +531,17 @@
 (spacemacs|define-transient-state window-manipulation
   :title "Window Manipulation Transient State"
   :doc (concat "
- Select^^^^              Move^^^^              Split^^                Resize^^                     Other^^
- ──────^^^^───────────── ────^^^^───────────── ─────^^─────────────── ──────^^──────────────────── ─────^^──────────────────────────────
- [_j_/_k_] down/up       [_J_/_K_] down/up     [_s_] vertical         [_[_] shrink horizontally    [_q_] quit
- [_h_/_l_] left/right    [_H_/_L_] left/right  [_S_] vert & follow    [_]_] enlarge horizontally   [_u_] restore prev layout
- [_0_-_9_] window N      [_r_]^^   rotate fwd  [_v_] horizontal       [_{_] shrink vertically      [_U_] restore next layout
- [_w_]^^   other window  [_R_]^^   rotate bwd  [_V_] horiz & follow   [_}_] enlarge vertically     [_d_] close current
- [_o_]^^   other frame   ^^^^                  ^^                     ^^                           [_D_] close other"
+ Select^^^^               Move^^^^              Split^^               Resize^^             Other^^
+ ──────^^^^─────────────  ────^^^^────────────  ─────^^─────────────  ──────^^───────────  ─────^^──────────────────
+ [_j_/_k_]  down/up       [_J_/_K_] down/up     [_s_] vertical        [_[_] shrink horiz   [_u_] restore prev layout
+ [_h_/_l_]  left/right    [_H_/_L_] left/right  [_S_] verti & follow  [_]_] enlarge horiz  [_U_] restore next layout
+ [_0_.._9_] window 0..9   [_r_]^^   rotate fwd  [_v_] horizontal      [_{_] shrink verti   [_d_] close current
+ [_w_]^^    other window  [_R_]^^   rotate bwd  [_V_] horiz & follow  [_}_] enlarge verti  [_D_] close other
+ [_o_]^^    other frame   ^^^^                  ^^                    ^^                   "
                (if (configuration-layer/package-usedp 'golden-ratio)
-                   "\n ^^^^                    ^^^^                  ^^                     ^^                           [_g_] golden-ratio %`golden-ratio-mode"
-                 ""))
+                   "[_g_] golden-ratio %`golden-ratio-mode"
+                 "")
+               "\n ^^^^                     ^^^^                  ^^                    ^^                   [_q_] quit")
   :bindings
   ("q" nil :exit t)
   ("0" select-window-0)


layers/+distributions/spacemacs-base/keybindings.el
```diff
@@ -531,16 +531,17 @@
...
- [_j_/_k_] down/up       [_J_/_K_] down/up     [_s_] vertical         [_[_] shrink horizontally    [_q_] quit
- [_h_/_l_] left/right    [_H_/_L_] left/right  [_S_] vert & follow    [_]_] enlarge horizontally   [_u_] restore prev layout
- [_0_-_9_] window N      [_r_]^^   rotate fwd  [_v_] horizontal       [_{_] shrink vertically      [_U_] restore next layout
- [_w_]^^   other window  [_R_]^^   rotate bwd  [_V_] horiz & follow   [_}_] enlarge vertically     [_d_] close current
- [_o_]^^   other frame   ^^^^                  ^^                     ^^                           [_D_] close other"
+ Select^^^^               Move^^^^              Split^^               Resize^^             Other^^
+ ──────^^^^─────────────  ────^^^^────────────  ─────^^─────────────  ──────^^───────────  ─────^^──────────────────
+ [_j_/_k_]  down/up       [_J_/_K_] down/up     [_s_] vertical        [_[_] shrink horiz   [_u_] restore prev layout
+ [_h_/_l_]  left/right    [_H_/_L_] left/right  [_S_] verti & follow  [_]_] enlarge horiz  [_U_] restore next layout
+ [_0_.._9_] window 0..9   [_r_]^^   rotate fwd  [_v_] horizontal      [_{_] shrink verti   [_d_] close current
```
d12frosted: I think we should come up to a convention on how to show sequences in hydras. In some places we use `,` (e. g. `0,9`), while here we use `-` or `..`.

duianto: I did a quick check through the <kbd>SPC ?</kbd> `transient`  (HELM Descbinds) entries.
#### Transient states
A comma `,` is used in these transient states:
- <kbd>SPC b .</kbd> Buffer Transient State
- <kbd>SPC l</kbd> Layouts Transient State
- <kbd>SPC l w</kbd> Workspaces Transient State

Two dots `..` are used in the <kbd>SPC n =</kbd> Evil Numbers Transient State  (+ or - instead of = also opens it).

Currently the <kbd>SPC w .</kbd> Window Manipulation Transient State uses a dash `-`, but if this PR is applied then it will use two dots `..`.

#### Documentation
In documentation.org these two lines use 2 dots `..`:

> [1..9, 0]	switch to nth layout
> [C-1..C-9, C-0]	switch to nth layout and keep the transient state active

https://github.com/syl20bnr/spacemacs/blob/develop/doc/DOCUMENTATION.org#layout-key-bindings

And news01.org uses 3 dots `...`:
>Spacemacs layouts now restrict the scope of the buffer list opened with SPC b b. Use SPC b B to list all the buffers of all the layouts. Also 1,2,...,9,0 keys now select a layout and close the layout menu, use C-1,C-2,...,C-9,C-0 to switch to a layout and keep the menu opened.

https://github.com/syl20bnr/spacemacs/blob/develop/news/news01.org#key-bindings
it's the next to last paragraph.

A dash `-` or a comma `,` won't work there. 

I found a page that talks about two vs three dots.
http://www.grammarphobia.com/blog/2008/12/two-dots-or-three.html
It states that two dots are used to save space, and the OED (Oxford English Dictionary) is a lot shorter because of it.

#### Question
Should `..` be used in all hydras (transient states)? Then it would be consistent in both the transient states and these two places in the news and documentation.

#### Notes
I have a Buffer TS formatting PR pending, but i didn't notice the comma as a sequence indicator when i made it. If two dots `..` should be used then i can update that PR, and make PRs for the layouts and workspaces as well.

About the two documentation.org lines. 
Is there a reason why they are written with the `0` and `C-0` after `1..9` and `C-1..C-9`? 
In the Layouts and Workspaces Transient States they are written starting from 0 up to 9: `[0,9]` and `[C-0,C-9]`. That seems more logical.

duianto: I just noticed that the same unusual order: `0` after `1..9` etc., also occur in the Workspace key bindings documentation:
>[1..9, 0]	switch to nth workspace
[C-1..C-9, C-0]	switch to nth workspace and keep the transient state active

https://github.com/syl20bnr/spacemacs/blob/develop/doc/DOCUMENTATION.org#workspace-key-bindings

d12frosted: Sorry for late response.

Personally I prefer `..`. 😸

d12frosted:

It's been a while 😸

layers/+distributions/spacemacs-base/keybindings.el
duianto: No problem with the response time.

I agree that `..` works best for sequences, since `,` (comma) separates different key bindings and `-` (dash) is used between keys in key chords.

Now that i re-read my previous notes section comment, where the last line said: 
>In the Layouts and Workspaces Transient States they are written starting from `0` up to `9`: `[0,9]` and `[C-0,C-9]`. That seems more logical.

I see that it might have sounded like i meant that the `,` (comma) would be more logical as a sequence. But that was not my intention, i should have written the documentation suggestion as: 
`[0..9]` and `[C-0..C-9]` 
instead of them current being written like this: 
`[1..9, 0]` and `[C-1..C-9, C-0]`.

Thanks for also updating the buffer, layout and workspace transient states 👍

d12frosted: `0..9` looks better than `1..9, 0` 😸 

> Thanks for also updating the buffer, layout and workspace transient states

Just wanted to be consistent. If you find something I missed - just send PR 😄

PR #8104, opened 2017-01-04
mu4e: Create email with attachments from dired

wigust:
Attachments are mostly handled using the helm baloo interface below, but
sometimes we want to send files from a directory `C-c C-m C-a` ([source](https://github.com/malb/emacs.d/blob/48508219453ebb542e6fec370bb49fba9980c2a2/malb.org))

### DIFF
--- a/layers/+email/mu4e/packages.el
+++ b/layers/+email/mu4e/packages.el
@@ -58,7 +58,24 @@
 
       (when mu4e-account-alist
         (add-hook 'mu4e-compose-pre-hook 'mu4e/set-account)
-        (add-hook 'message-sent-hook 'mu4e/mail-account-reset)))))
+        (add-hook 'message-sent-hook 'mu4e/mail-account-reset))
+
+      (require 'gnus-dired)
+      ;; make the `gnus-dired-mail-buffers' function also work on
+      ;; message-mode derived modes, such as mu4e-compose-mode
+      (defun gnus-dired-mail-buffers ()
+	"Return a list of active message buffers."
+	(let (buffers)
+	  (save-current-buffer
+	    (dolist (buffer (buffer-list t))
+	      (set-buffer buffer)
+	      (when (and (derived-mode-p 'message-mode)
+			 (null message-sent-message-via))
+		(push (buffer-name buffer) buffers))))
+	  (nreverse buffers)))
+
+      (setq gnus-dired-mail-mode 'mu4e-user-agent)
+      (add-hook 'dired-mode-hook 'turn-on-gnus-dired-mode))))
 
 (defun mu4e/init-mu4e-alert ()
   (use-package mu4e-alert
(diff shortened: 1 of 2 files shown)


layers/+email/mu4e/packages.el
```diff
@@ -58,7 +58,27 @@
 
       (when mu4e-account-alist
         (add-hook 'mu4e-compose-pre-hook 'mu4e/set-account)
-        (add-hook 'message-sent-hook 'mu4e/mail-account-reset)))))
+        (add-hook 'message-sent-hook 'mu4e/mail-account-reset))))
+
+  ;; Source https://github.com/malb/emacs.d/blob/48508219453ebb542e6fec370bb49fba9980c2a2/malb.org
+  (add-hook 'dired-mode-hook 'turn-on-gnus-dired-mode)
+  (use-package gnus-dired
```
d12frosted: Shouldn't it declared in `mu4e-packages` list and then configured in `mu4e/init-gnus-dired`? Am I missing something? 😸

wigust:
@d12frosted I even found that adding `(add-hook 'dired-mode-hook 'turn-on-gnus-dired-mode)` is enough. Is it ok to put inside `mu4e/init-mu4e`? https://github.com/syl20bnr/spacemacs/pull/8104/commits/c23c83d39d6f76216c620ca91b59ffee04806c86

wigust:
I think it's a bad place though. Maybe put it mu4e documentation is enough?

d12frosted:
Usually if you setup 3rd party package in a layer, you have to [follow the rules][1].

P. S. sorry for the late response.

  [1]: https://github.com/syl20bnr/spacemacs/blob/develop/doc/LAYERS.org#packagesel

nixmaniack:
This is good. But you should add following.

```elisp
(require 'gnus-dired)
;; make the `gnus-dired-mail-buffers' function also work on
;; message-mode derived modes, such as mu4e-compose-mode
(defun gnus-dired-mail-buffers ()
  "Return a list of active message buffers."
  (let (buffers)
    (save-current-buffer
      (dolist (buffer (buffer-list t))
        (set-buffer buffer)
        (when (and (derived-mode-p 'message-mode)
                (null message-sent-message-via))
          (push (buffer-name buffer) buffers))))
    (nreverse buffers)))

(setq gnus-dired-mail-mode 'mu4e-user-agent)
(add-hook 'dired-mode-hook 'turn-on-gnus-dired-mode)
```

1. This allows to add files to currently composing mail buffer
2. In case there are no mail composing buffers, it will correctly start `mu4e-compose-new` instead of default gnus compose.

Source: http://www.djcbsoftware.nl/code/mu/mu4e/Attaching-files-with-dired.html

syl20bnr:
Your repository is gone but I applied your PR in your name in commit: https://github.com/syl20bnr/spacemacs/commit/9c585370

Thank you 👍

PR #8111, opened 2017-01-05
configure xkcd inside the :init progn clause

deb0ch:
It worked before and it still works now, I just feel like this is more The Right Way™ of doing it 

@d12frosted :smile_cat:


deb0ch:
Thanks @d12frosted for fixing xkcd in 2581058a49f1b5c0591c1126f52839c793af3103, btw :smiley_cat:

d12frosted:
Uh-oh. Initially I placed it under `:config`, but then I changed my mind and... well, thanks for noticing it!

d12frosted:

PR #8118, opened 2017-01-06
also provide align functions for {, }, [ and ] under SPC x a

deb0ch:
This feature was requested by a user in the gitter channel ([January 6, 2017 6:32 PM](https://gitter.im/syl20bnr/spacemacs?at=586fd4b15ffdeea72314b947)), and it makes sense to have it.

Also, as @jredville suggested ([January 6, 2017 6:55 PM](https://gitter.im/syl20bnr/spacemacs?at=586fda0961e516c157902664)), it could be nice to have an align function reading a character to align by from the minibuffer.

I haven't had time to implement it today, just let me know if it could be useful and I'll implement it.

Cheers :smile_cat:

### DIFF
--- a/layers/+distributions/spacemacs-base/funcs.el
+++ b/layers/+distributions/spacemacs-base/funcs.el
@@ -772,6 +772,10 @@ the right."
 (spacemacs|create-align-repeat-x "bar" "|")
 (spacemacs|create-align-repeat-x "left-paren" "(")
 (spacemacs|create-align-repeat-x "right-paren" ")" t)
+(spacemacs|create-align-repeat-x "left-curly-brace" "{")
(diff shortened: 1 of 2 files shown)


deb0ch:
**Update:** aligning by regex is already available under `SPC x a r`. The thing is that nobody knows it (and uses it) because it is under the cryptic name `align-repeat`.

I think that it would be great to change the displayed name to `align-regex` or something like that.

I think that it would be good to also simplify the displayed name of the other aligning functions: `align-repeat` is not very descriptive and adds some *WTF* effect to the discovery of these bindings.

Last thing: I still haven't been able to figure out what `align` and `align-current` are supposed to do, even reading their docstring. How the hell do they work ? How can we make that clearer ?

syl20bnr:
Thank you ! 👍 

PR #8121, opened 2017-01-06
Fixes #8242 -- Adding additional go test functionality

### LINKED ISSUE #8242: Go layer tests can result in spacemacs buffer hanging indefinitely.
#### Description :octocat:
Go tests resulting in infinite loops cause Spacemacs tab to hang and must be restarted.

- Start Emacs
- Attempt to test an infintie loop `for {}`
- Watch Spacemacs test handler tab hang indefinitely.

Spacemacs go test handling tab will hang while waiting for the go tests to run. There needs to be a `-timeout 10s` or some other simlar option.

The go test ought to timeout (or be able to time out via a config value). There's no necessary system info, so the following lines will be deleted.

See https://github.com/syl20bnr/spacemacs/pull/8121

GrappigPanda:
The gist of this change is the following: `go test` allows different commands to be passed in (like -race, which tests for race conditions). Some of these test arguments are highly desirable in day-to-day testing and development. So, to improve developer experience, the new `go-use-test-args` variable allows go developers to specify whatever they may want or need.

Sorry if this is seen as a drive-by PR!

ADD: 
 - layers/+lang/go/config.el Added new variable `go-use-test-args` to allow specifying
    additional arguments being passed to `go test.

CHANGE:
  - layers/+lang/go/packages.el Updated `go-run-tests` to automatically concat
    the new variable `go-use-test-args` to args passed to `go test`.

### DIFF
--- a/layers/+lang/go/README.org
+++ b/layers/+lang/go/README.org
@@ -98,6 +98,12 @@ function testing to work.
 the coverage buffer. See [[https://www.gnu.org/software/emacs/manual/html_node/elisp/Choosing-Window.html][=display-buffer=]] for a list of possible functions.
 The default value is =display-buffer-reuse-window=.
 
+
+To provide additional arguments to go test, specify =go-use-test-args=
+#+begin_src emac-lisp
+  (go :variables go-use-test-args "-race -timeout 10s"
+#-end_src
+
 ** Guru
 
 Go Oracle has been deprecated as of October 1, 2016, it's replacement is =go-guru=.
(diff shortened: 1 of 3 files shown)


layers/+lang/go/README.org
```diff
@@ -98,6 +98,12 @@ function testing to work.
 the coverage buffer. See [[https://www.gnu.org/software/emacs/manual/html_node/elisp/Choosing-Window.html][=display-buffer=]] for a list of possible functions.
 The default value is =display-buffer-reuse-window=.
 
+
+To provide additional arguments to go test, specify =go-use-test-args=
+#+begin_src emac-lisp
+  (go :variables go-use-test-args "-race -timeout 10s"
```
db47h: Maybe mention "-v" as a very commonly used argument.

TheBB:

I had to resolve some conflicts. Please check if I got it right.

PR #8122, opened 2017-01-06
Fix cider and cider-repl mode names in company configuration for clojure layer

ilya-babanov:
Fix `cider` and `cider-repl` mode names in company configuration.
Right now there is no auto-completion for both of this modes.
Looks like this is just a typo introduced by [this change.](https://github.com/syl20bnr/spacemacs/commit/74fdbb6795018a93a665fa2c417589cd4cbd51fc#diff-43a23dbec19deb02144c5310368c065eR277)

### DIFF
--- a/layers/+lang/clojure/packages.el
+++ b/layers/+lang/clojure/packages.el
@@ -278,8 +278,8 @@
   (spacemacs|add-company-backends
     :backends company-capf
     :modes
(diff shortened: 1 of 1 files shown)


hujianxin:
Can confirm this pr.

d12frosted:

PR #8130, opened 2017-01-07
Add missing "m key" to debugging and help base prefix

smile13241324:
Noticed that for the debugging and help base prefix the "m key" was missing in CONVENTIONS.org and added it.


syl20bnr:
Thank you !

Issue #7757, opened 2016-11-16
Make space-doc-mode available for any Org file


jlombera:
#### Description :octocat:
`space-doc-mode` is very convenient to view notes/doc, I use it to view my personal notes too, not just Spacemacs's.  However, this function is not auto-loaded for org-mode files, it is available only after some Spacemacs doc is visited (e.g. after `SPC h l ...`).  This means that I need to open a random Spacemacs doc before I can use `space-doc-mode` to view my documents.

It would be would that `spac-doc-mode` were available right away (auto-loaded?) for any Org document.  Furthermore, since it is such a great view mode (IMHO), it would be even better if it were mapped to a key binding by default (e.g. `SPC m v`).

#### System Info :computer:
- OS: gnu/linux
- Emacs: 25.1.1
- Spacemacs: 0.200.5
- Spacemacs branch: develop (rev. fea919c)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
(ivy auto-completion better-defaults emacs-lisp git markdown org spell-checking syntax-checking version-control)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GCONF GSETTINGS NOTIFY ACL LIBSELINUX GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XWIDGETS

TheBB:
I mean, all you need is to mark it as autoloadable.

`(autoload 'space-doc-mode "space-doc" nil 'interactive)`

Then you can add it to org-mode-hook as usual.

jflombera:
Thanks for the pointer, I will do that.  However, it would be good if this were provided by default in stock Spacemacs.

TheBB:
Yeah that's why I mentioned it, now you can PR. ;-)

jflombera:
@TheBB thank you for you promptly reply.

I am new to Emacs and am not so familiar (yet) to Emacs/Spacemacs architecture/design and elisp in general.  I guess this is a trivial change but it will take me perhaps a little more time.  Sure, I will do it, just have a little patience with me ;)

Beta1440:
Don't forget to close this issue.

blackdev1l:
[...]

PR #8135, opened 2017-01-08
Bind space-doc-mode to "SPC m v" in Org

nikolaiam:
Provide a way to toggle `space-doc-mode` in any org file with a keybinding as requested in #7757. Useful for reading documents in the same mode as Spacemacs documentation is opened in.


syl20bnr:
Thank you !
I moved the key binding to `SPC m V` since it can be pretty confusing if hit inadvertently.

Issue #8054, opened 2016-12-25
Org export issue: unable to resolve link "configuring-spacemacs"


ztlevi:
I got an issue when I try to export org to other formats. 
![](https://ww3.sinaimg.cn/large/006tNc79gw1fb32a09augj30wo03wdh6.jpg)
Does anybody know this issue?
Thanks.

d12frosted:
I just tried to export to `html` and everything worked as expected. Could you please provide more clear reproduction steps? Also, having output of `SPC h d s` would be helpful.

ztlevi:
This is my config: https://github.com/ztlevi/Spacemacs-config
#### System Info :computer:
- OS: darwin
- Emacs: 25.1.1
- Spacemacs: 0.200.5
- Spacemacs branch: develop (rev. 703733b)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
(php ivy better-defaults github ranger colors prodigy search-engine graphviz
     (syntax-checking :variables syntax-checking-enable-by-default nil syntax-checking-enable-tooltips nil)
     (spell-checking :variables spell-checking-enable-by-default nil)
     (vinegar :variables vinegar-reuse-dired-buffer t)
     (spacemacs-layouts :variables layouts-enable-autosave nil layouts-autosave-delay 300)
     (git :variables git-magit-status-fullscreen t magit-push-always-verify nil magit-save-repository-buffers 'dontask magit-revert-buffers 'silent magit-refs-show-commit-count 'all magit-revision-show-gravatars nil)
     (ibuffer :variables ibuffer-group-buffers-by 'projects)
     (auto-completion :variables auto-completion-enable-sort-by-usage t auto-completion-enable-snippets-in-popup t :disabled-for org markdown)
     (osx :variables osx-dictionary-dictionary-choice "Simplified Chinese - English")
     restclient
     (gtags :disabled-for clojure emacs-lisp javascript latex python shell-scripts)
     (shell :variables shell-default-shell 'eshell)
     docker deft markdown org shaders yaml react
     (python :variables python-test-runner
             '(nose pytest))
     (ruby :variables ruby-version-manager 'chruby)
     ruby-on-rails lua html javascript
     (typescript :variables typescript-fmt-on-save nil typescript-fmt-tool 'typescript-formatter)
     emacs-lisp
     (clojure :variables clojure-enable-fancify-symbols t)
     racket
[...]

PR #8136, opened 2017-01-08
Fix links in documentation which breaks org html export

smile13241324:
It seems that some of the documentation files cannot longer be exported to html via org. I had a look into this and found some wrongly declared internal and external links which I have corrected so that it is again possible to export each spacemacs documentation file.

However in #8054 @deb0ch has noticed that this is happening even in the automatically generated table of content so the root cause may still be a bug in org mode.

Anyway this merge request may at least kill the symptoms of the issue by adding "Custom_ID properties" as well as correcting links to relative paths where necessary in the documentation files.

### DIFF
--- a/CONTRIBUTING.org
+++ b/CONTRIBUTING.org
@@ -25,18 +25,24 @@ Thanks! :heart: :heart: :heart:
  - [[#credits][Credits]]
 
 * Asking for help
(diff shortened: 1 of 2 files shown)


syl20bnr:
Thank you ! 👍 

Issue #8127, opened 2017-01-07
Helm bindings break when switching to help window.


Somelauw:
#### Steps to reproduce
- Open helm-apropos (SPC h d d).
- Press TAB (helm-persistent-action). A help window should open.
- Press M-1 (select-window-1) to go to help window.
- Press M-2 (select-window-2) to go back to helm

#### Result
Helm bindings stop working

#### Expected
Helm bindings remain working

#### System Info :computer:
- OS: gnu/linux
- Emacs: 24.4.1
- Spacemacs: 0.200.5
- Spacemacs branch: master (rev. 664ba6a)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: helm
- Layers:
```elisp
((auto-completion :variables auto-completion-return-key-behavior 'complete auto-completion-tab-key-behavior 'cycle auto-completion-complete-with-key-sequence nil auto-completion-complete-with-key-sequence-delay 0.1 auto-completion-private-snippets-directory nil auto-completion-enable-snippets-in-popup t auto-completion-enable-help-tooltip 'manual)
 better-defaults
 (evil-cleverparens :variables evil-cleverparens-use-additional-movement-keys nil evil-cleverparens-complete-parens-in-yanked-region t)
 games git gtags spell-checking syntax-checking version-control)
```

I haven't been able to test this on the "develop" branch. If it's fixed upstream, feel free to close.

bmag:
That's because during a helm session, the selected window should be the minibuffer, not the results window. Instead of `M-2`, use `SPC w b` or `M-m w b` to go back to the minibuffer.

@deb0ch I noticed that `M-1`, `M-2`, etc. don't select a window in develop anymore, is that because you moved the bindings in winum,el to `C-x w [digit]`? I didn't realize window-numbering.el used to set them on its, I always assumed we configured those in Spacemacs.

deb0ch:
@bmag Exactly, I had to move the bindings to something else as it was conflicting with native Emacs bindings and Purcell wouldn't let it get into Melpa in that state.

The shadowed bindings were the prefix arguments (`M-0` to `M-9`), so it makes reasonable sense.

I wonder if we should get it back to `M-0` to `M-9` in Spacemacs :smirk_cat:

bmag:
I think we should. Can you make a PR please?

deb0ch:
Sure :construction_worker:

bmag:
[...]

PR #8139, opened 2017-01-09
restore emacs-style window selection bindings

deb0ch:
@bmag's [request](https://github.com/syl20bnr/spacemacs/issues/8127#issuecomment-271214642)

I couldn't choose if we should shadow the (crap) original bindings of the package or keep them, so I have opened 2 PRs (see #8140).

This PR keeps the original bindings on top.

Just pick one and close the other :smile_cat:


bmag:

PR #8140, opened 2017-01-09
restore emacs-style window selection bindings

deb0ch:
@bmag's [request](https://github.com/syl20bnr/spacemacs/issues/8127#issuecomment-271214642)

I couldn't choose if we should shadow the (crap) original bindings of the package or keep them, so I have opened 2 PRs (see #8139).

Just pick one and close the other :smile_cat:


bmag:
I chose to keep the original bindings as well (from 8139)

PR #8142, opened 2017-01-09
Make python-execute-file use python-shell-interpreter

swaroopch:
So that we can customize to use `python3` instead of `python`


TheBB:

PR #8138, opened 2017-01-09
Fix export-breaking org links in documentation

nikolaiam:
As discussed in #8054 and #8136, org doesn't always export links correctly if
the syntax deviates from the one given on the org mode website. This results
in broken links on github and/or spacemacs.org as in #6722, #8036, #8054, #7956
and possibly others.

This modifies all links not already fixed in #8136. Internal links are
handled with Custom IDs. External file links are formatted according to org
guidelines as well (see bottom), and if they point to a specific headline in an
external file, a custom ID is created as well.

Fixes #6722, #8036, #7916 and #7956 as well as other unreported broken
links. #8146 hides the custom ID drawers introduced here.

http://orgmode.org/manual/External-links.html#External-links
http://orgmode.org/manual/Internal-links.html#Internal-links

PR #8146, opened 2017-01-10
Hide PROPERTIES drawers in space-doc-mode

nikolaiam:
Hide `PROPERTIES` drawers in space-doc-mode to allow the use of `CUSTOM_ID` tags in documentation for correct link exporting, as discussed in #8054, #8136 and #8138. Complementary the latter two PR:s.


syl20bnr:
Thank you again 💜 

I updated the docs on the develop website and the properties show up in the exported HTML docs: 
http://develop.spacemacs.org/doc/DOCUMENTATION.html#states

Do you know how to fix this ?

nikolaiam:
Ah! I'm not familiar with the export process, but I guess it differs from `org-export-dispatch` somehow? I checked for this (anticipating this could happen :smile:) before doing the PR. Here's how I got them to work locally:
1. Add `#+HTML_HEAD: <link rel="stylesheet" type="text/css" href="../css/readtheorg.css" />` to the Org file headers for linking to the correct CSS
2. Export using `SPC m e e h h` (`org-export-dispatch`)

So one idea would be just using the `org-export-dispatch` process. The `#+HTML_HEAD` line could be removed in `space-doc-mode` similarly to the fix here. How does this sound?

PR #8147, opened 2017-01-11
throw a real error if emacs version is too old

deb0ch:
Many people (including me) were confused when trying to install Spacemacs on an outdated Emacs. A message was issued, but honestly nobody thinks of going to see the `*Messages*` buffer when a perfeclty normal vanilla Emacs starts up, especially not beginners.

This way, an error is visible and users can take corrective actions.

One more thing: I changed the concatenated string to a normal full string even if it makes it > 80 cols, as cutting it prevents it from being searched in the code base. This is a recommendation of the Linux Kernel coding style (line 86) to never break user-visible strings, which makes a lot of sense to me.

### DIFF
--- a/init.el
+++ b/init.el
@@ -20,9 +20,9 @@
 (defconst spacemacs-emacs-min-version   "24.4" "Minimal version of Emacs.")
 
 (if (not (version<= spacemacs-emacs-min-version emacs-version))
-    (message (concat "Your version of Emacs (%s) is too old. "
-                     "Spacemacs requires Emacs version %s or above.")
-             emacs-version spacemacs-emacs-min-version)
+    (error (concat "Your version of Emacs (%s) is too old. "
+                   "Spacemacs requires Emacs version %s or above.")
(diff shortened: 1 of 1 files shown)


syl20bnr:
You can keep the string on 2 lines please ;-)
The break is between two sentences so readability is maintained and more importantly the "easy grep" argument does not hold since we have variable interpolation here.

deb0ch:
Changed :expressionless: 

What do you think of the initial idea ?

Story behind it is that some time ago I wanted to make a coworker try Spacemacs, then couldn't figure out what was going on when only vanilla Emacs kept popping out. Quickly he didn't have time anymore and gave up altogether. Now I'm not even sure he'll even try it one day, and it's very frustrating.

Having this we would have figured it out right away and without ambiguity :smile_cat: 

And seeing the gitter chat, it also affects a lot of people.

deb0ch:
Ping: what's the status on this PR ? (@syl20bnr ?) :smile_cat:

bmag:

PR #8149, opened 2017-01-11
fix ranger with golden-ratio activated

deb0ch:
A user in the gitter chat pointed out that `ranger` was not working correctly along with `golden-ratio`.

After some investigations, we came up with that fix :smile_cat:

### DIFF
--- a/layers/+tools/ranger/packages.el
+++ b/layers/+tools/ranger/packages.el
@@ -13,6 +13,7 @@
 (setq ranger-packages
       '(
         (dired :location built-in)
+        golden-ratio
         ranger
         ))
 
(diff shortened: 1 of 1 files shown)


deb0ch:
Just tested it again yesterday and for some reason the post-init hook does not execute anymore (I'm fairly sure that it did when I submitted this PR).

Have to look into that.

d12frosted:
This is an oldie. 😢 Let me know once you fix it!

deb0ch:
Reported upstream, more details there: https://github.com/ralesi/ranger.el/issues/148

deb0ch:
Long story short, this fixes the issue as long as `purpose-mode` is not activated. I think some additional code in Ranger is needed to handle `emacs-purpose`.

syl20bnr:
Thank you !
I merged although `emacs-purpose` is on by default in develop and this fix does not work here.

PR #8155, opened 2017-01-13
transient-state: Improve handling of additional bindings

madand:
Add two new functions: `spacemacs/transient-state-register-add-bindings`
and `spacemacs/transient-state-register-remove-bindings` to prevent layer authors and end users from dealing with the underlying variables' subtleties.

Background

`spacemacs-%s-transient-state-add-bindings` and
`spacemacs-%s-transient-state-remove-bindings` are special variables that allow
to hook into transient state creation process to add or remove some bindings
before a Hydra is created. The problem is that these variables are not
existing by default and expected to be created on demand by the external code.

Currently layers that use this feature, set the variables directly with `setq`.
This works OK until some other layer wants to register its own additional bindings
for the same transient state. Obviously, depending on the load order, the last
layer's `setq` will win, while other layer's changes will be lost. In fact,
it has [happend](https://xkcd.com/1172/) to me with b515253026998ebbea83df6cb825108e3c788623, since one of my private layers also modified `spacemacs-layouts-transient-state-add-bindings`.

For a while I used the following ugly hack in `helm/packages.el`:
```emacs-lisp
 (defun helm/post-init-persp-mode ()
   (or (boundp 'spacemacs-layouts-transient-state-add-bindings)
       (setq spacemacs-layouts-transient-state-add-bindings '()))
   (dolist (binding '(("b" spacemacs/persp-helm-mini :exit t)
                      ("l" spacemacs/helm-perspectives :exit t)))
     (push binding spacemacs-layouts-transient-state-add-bindings)))
```
Finally, found some time to properly address the issue. So here is the PR.

### DIFF
--- a/core/core-transient-state.el
+++ b/core/core-transient-state.el
@@ -27,6 +27,16 @@
 holds the key bindings."
   (intern (format "spacemacs/%S-transient-state/heads" name)))
 
+(defun spacemacs//transient-state-add-bindings-name (name)
+  "Return the name of the transient state add-bindings variable which
+may hold the additional key bindings. The variable may be unbound."
+  (intern (format "spacemacs-%s-transient-state-add-bindings" name)))
+
+(defun spacemacs//transient-state-remove-bindings-name (name)
+  "Return the name of the transient state remove-bindings variable which
+may hold the keys to be removed. The variable may be unbound."
+  (intern (format "spacemacs-%s-transient-state-remove-bindings" name)))
+
 (defun spacemacs//transient-state-adjust-bindings (bindings to-remove to-add)
   (append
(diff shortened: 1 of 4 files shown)


TheBB:
Great work. Thanks! Cherry-picked in develop. You can safely delete your branch.

I changed to symbols: https://github.com/syl20bnr/spacemacs/commit/909d23f41231ece45198e0352effd9b8da6c7bab

It's lisp after all. ;-)

madand:
@TheBB yeah, symbols look better than strings for sure. Please also merge a follow-up PR #8949, which addresses changes for code, that was added to `develop` after this PR was submitted.

PR #8157, opened 2017-01-13
add phpplus layer for company-php package

xcwen:
https://github.com/xcwen/ac-php#usage-spacemacs
use company-php


syl20bnr:
Than you for the PR, why a new layer ? Could it be moved to php layer instead ?

xcwen:
because  not need          ggtags   helm-gtags in php layer

xcwen:
ok , i push it to php layer .

PR #8165, opened 2017-01-13
a realtime markdown preview using chrome

sukeyisme:

Thank you <3


ksjogo:
There already is the vmd markdown preview functionality in spacemacs. What is the advantage of using the chrome one?

sukeyisme:
@ksjogo Only to discover later that there is no need

syl20bnr:
Seems that `flymd` requires no setup, it may be a superior solution to `vmd`, do we reopen ?

layers/+tools/chrome/packages.el
```diff
@@ -26,3 +27,17 @@
...
+
+(defun chrome/init-flymd ()
+  (use-package flymd
+    :defer t
+    :init
+    (progn
+      (defun my-flymd-browser-function (url)
+        (let ((process-environment (browse-url-process-environment)))
+          (apply 'start-process
+                 (concat "google-chrome " url) nil
```
joehillen: What about `chromium` and `google-chrome-stable`?

sukeyisme: might as well

joehillen: Maybe add a var to override it.

sukeyisme: this is only buffer name , see  chrome-exec-path

joehillen: Ah, my mistake.

layers/+tools/chrome/config.el
```diff
@@ -0,0 +1,2 @@
+(defvar chrome-exec-path "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"
```
joehillen: As a Linux user, this displeases me.

sukeyisme: replace your chrome path

joehillen: It doesn't make a good default.

joehillen: A better solution would be to have the default be `nil` and then when it is `nil` search `PATH` for `google-chrome`, `google-chrome-stable`, `chromium`, and whatever you do to find programs on Mac.

sukeyisme: you are right

layers/+tools/chrome/packages.el
```diff
@@ -26,3 +27,26 @@
...
+                 "flymd" nil
+                 browser
+                 (list "--new-window" "--allow-file-access-from-files" url))))
+
+      (defun my-flymd-browser-function (url)
+               (cond
+                ((executable-find "chromium") (start-browser (executable-find "chromium") url))
+                ((executable-find "google-chrome") (start-browser (executable-find "goolge-chrome") url))
+                ((executable-find "google-chrome-stable") (start-browser (executable-find "goolge-chrome-stable") url))
+                (chrome-exec-path (start-browser chrome-exec-path url))
```
d12frosted: Since you allows users to specify `chrome-exec-path`, I think it would be better to prioritise it over `executable-find` when the value is not nil.

d12frosted:
Thank you :+1: Squashed and cherry-picked into `develop`, you can safely delete your branch. Also I have added some notes about `flymd` to `markdown` layer documentation.

Issue #8133, opened 2017-01-08
gnus-group-mode-map doesn't work on evilified-mode


LdBeth:
#### Description :octocat:
gnus-group-mode-map doesn't work on evilified-mode

#### Reproduction guide :beetle:
- Configure the dotspacemacs file with
```
   dotspacemacs-editing-style '(hybrid
                                :variables
                                hybrid-mode-enable-evilified-state t
                                hybrid-mode-enable-hjkl-bindings t
                                hybrid-mode-default-state 'normal)
```
or just `vim`, and you need `gnus` layer.

- Start Emacs
- `SPC a g` open gnus
- In the Group buffer, normal state, pressing `G`
*Observed behaviour:* :eyes: :broken_heart:
You can't call out gnus-group commands under normal state. And of course the cursor gets to the end of the buffer.

*Expected behaviour:* :heart: :smile:
It should display the which-key menu, maybe it is the problem of evil.

#### System Info :computer:
- OS: darwin
- Emacs: 25.1.1
- Spacemacs: 0.200.7
- Spacemacs branch: master (rev. c2774bc)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: helm
- Layers:
```elisp
(helm
 (auto-completion :variables auto-completion-enable-snippets-in-popup t)
 emacs-lisp
 (markdown :variables markdown-live-preview-engine 'vmd)
 html lua
 (latex :variables latex-enable-auto-fill t latex-enable-folding t)
 yaml sql
 (shell :variables shell-default-shell 'term shell-default-term-shell "/bin/zsh" shell-default-height 30 shell-default-position 'bottom)
 syntax-checking
 (osx :variables osx-use-dictionary-app nil)
 chinese emms emoji
 erc gnus maxima blog)
```
- System configuration features: JPEG RSVG NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS

bmag:
`G` key is overridden by `(evilified-state-evilify gnus-group-mode gnus-group-mode-map)` in `gnus` layer. We can probably bind any original `G` bindings to a different prefix, though I never tried Gnus so I can't say for sure.
I also see that the documentation for `gnus` layer says that `G R` is "Group Buffer - Add RSS feed", which apparently is wrong and needs to change.

LdBeth:
[...]

PR #8177, opened 2017-01-15
Bind Shawdowed bindings to Major mode prefix

LdBeth:
Since `G` and `g` are shadowed in evilified-stae, these functions are rebind to major mode prefix in gnus-group-mode.

Also, the functions of `SPC` are bind to `'`.


LdBeth:
#8133

LdBeth:
Close this pr because I want a more elegant approach

PR #4728, opened 2016-01-21
Use built-in web-mode pairing instead of smartparens

TheBB:
Fixes https://github.com/syl20bnr/spacemacs/issues/4499 and a host of questions on gitter.

The current solution is quite buggy. This will prevent stuff autoclosing of stuff like `(` in regular text in HTML, but I think it's a net win.

Maybe other regular web-mode users can give some feedback?

Ping @synic

PR #8178, opened 2017-01-15
Turn on auto parens in web mode

posva:
I realised this feature was turned off while writing `.vue` files. Using the web-mode is the (currently) best mode to work and since you write a lot of script, having the smartparens makes it much better:
web-mode-smartparens

### DIFF
--- a/layers/+lang/html/packages.el
+++ b/layers/+lang/html/packages.el
@@ -164,9 +164,7 @@
    (if dotspacemacs-smartparens-strict-mode
        'smartparens-strict-mode
      'smartparens-mode)
-   '(css-mode-hook scss-mode-hook sass-mode-hook less-css-mode-hook))
-
-  (add-hook 'web-mode-hook 'spacemacs/toggle-smartparens-off))
+   '(css-mode-hook scss-mode-hook sass-mode-hook less-css-mode-hook)))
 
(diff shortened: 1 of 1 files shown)


nixmaniack:
It was disabled for a reason I remember. See https://github.com/syl20bnr/spacemacs/pull/4728 for more info.

posva:
@nixmaniack Hey, thanks for that. I didn't back traced to the issue when I did `git blame`.
The problem was that typing `{{` then `space` would create an extra closing `}` like `{{ }}}`.
I tested that and works fine: it doesn't add any extra `}` 🎉

TheBB:
Still breaks it for me. Try opening a `.djhtml` file and type `{%` followed by space.

posva:
How does it break for you? I tested setting web mode on the scratch buffer and typing `{% ` yields `{% }`. I thought that was good enough 😄

TheBB:
Sorry, that's not good enough. Web-mode deals with tons of different templating engines, and `{% %}` isn't special in all of them. Try running `web-mode-set-engine` then `django` first, or open a `djhtml` file, which will automatically choose the django engine.

aaronjensen:
Just bit by this. Should web-mode be inserting closing parens? I’m not seeing that happen on typescript-tsx-mode

PR #8179, opened 2017-01-15
Go: fix void reference to go--position-bytes in go-rename

db47h:
Looks like this is an obsolete reference to a now missing function.


rugginoso:
Thanks for this :heart:

bmag:

PR #8180, opened 2017-01-15
GO: Improve go test output buffer behavior.

db47h:
The current implementation  runs tests in an async shell window. If any errors occur or any source file is otherwise referenced in the output, the user has to manually load the file/goto line. 

- Use compilation-start to get highlighting of compilation or runtime errors.
- Configure output window in popwin and window-purpose.
- Configurable buffer name.

Now <kbd>C</kbd>-<kbd>g</kbd> closes the output window, <kbd>C</kbd>-<kbd>x</kbd> <kbd>`</kbd> jumps to the first error. This will also be very handy for people who do not use flycheck.


syl20bnr:
Thank you ! 👍 

PR #8184, opened 2017-01-16
elm layer docs: adds note about indentation

JoshTGreenwood:
Indentation in the elm layer was confusing to me at first so I thought it would be a good idea to add a note.  I stole the section from elm-modes docs: https://github.com/jcollard/elm-mode#indentation


syl20bnr:
Thank you !

PR #8196, opened 2017-01-17
Update sort lines by column, docstrings, how to

duianto:
Add a description of how to sort lines by a selected column,
to the dostrings of the functions:
`sort-lines-by-column` and `sort-lines-by-column-reverse`.

### DIFF
--- a/layers/+distributions/spacemacs-base/funcs.el
+++ b/layers/+distributions/spacemacs-base/funcs.el
@@ -841,7 +841,8 @@ A non-nil argument sorts in reverse order."
   (spacemacs/sort-lines -1))
 
 (defun spacemacs/sort-lines-by-column (&optional reverse)
-  "Sort lines by the selected column.
+  "Sort lines by the selected column,
+using a visual block/rectangle selection.
 A non-nil argument sorts in reverse order."
   (interactive "P")
   (let* ((region-active (or (region-active-p) (evil-visual-state-p)))
@@ -850,7 +851,8 @@ A non-nil argument sorts in reverse order."
     (sort-columns reverse beg end)))
 
 (defun spacemacs/sort-lines-by-column-reverse ()
-  "Sort lines by the selected column in reverse order."
+  "Sort lines by the selected column in reverse order,
+using a visual block/rectangle selection."
   (interactive)
   (spacemacs/sort-lines-by-column -1))
 


duianto:
I discovered that this "how to" description, of making a vertical selection, doesn't work in the 
`evil-emacs-state`.

The Emacs `sort-columns` function that is used by the `spacemacs/sort-lines-by-column` function, doesn't sort the lines if the selections start and end columns are the same. 

It doesn't work in the `evil-emacs-state` <kbd>C-z</kbd>, when the cursor is moved down vertically, because the end of the selection is located to the left of the cursor.

But It does work in the `evil-visual-char` <kbd>v</kbd> or `evil-visual-block` <kbd>S-v</kbd> states, because the end of the selection is located to the right of the cursor. 

Here's an example:
1. start emacs
2. open a new buffer <kbd>SPC b N</kbd>
3. type:
```
67890
12345
```
4. move the cursor to the `8`
5. start a selection <kbd>C-SPC</kbd>
6. move the cursor to the `3` on the second line
7. check the points location by pressing <kbd>S-M-;</kbd>, the "Eval: " prompt appears in the minibuffer.
8. type: `(point)` and press <kbd>C-m</kbd> or <kbd>RET</kbd>

(after step 4, the `evil-emacs-state` and `evil-normal-state` states can be toggled by pressing <kbd>C-z</kbd>).

In the `evil-emacs-state` the `(point)` is located between the `2` and `3`, the point position is `9`, it's in the same column as the start of the selection. If one tries to sort the lines by column <kbd>M-m x l c</kbd>, then the lines don't get sorted, the cursor just jumps to the `6`.

In of the `evil-visual-char` or `...-block` states, the `(point)` is located to the right of the `3`, the point position is `10`, it's one column to the right of the selections starting column. Sorting by column <kbd>SPC x l c</kbd> sorts the lines, and the cursor jumps to the `1`.

#### I might have found a solution:
If the the last argument that's passed to the `(sort-columns REVERSE &optional BEG END)` function, is one column to the right of the column where the selection was started, but on the last line. Then it doesn't matter where in the last line the cursor is placed. The lines will get sorted, by the column where the selection was started, even in the `evil-emacs-state`.

I'll make a new issue with the possible column correction and some other suggestions to the `spacemacs/sort-lines-by-column` function.

I can update this PR with a better description, when it's clear how it should work.

The updated description could say something like this:
```
  "The selections first character is used as the column for 
sorting the lines that the selection is drawn over.
A non-nil argument sorts the lines in reverse order.

It only works on an active selection, with at least two lines.

All but one selection command and key binding can be used:
evil-visual-char,    \\[evil-visual-char]
evil-visual-block,   \\[evil-visual-block]
set-mark-command,    \\[set-mark-command]
rectangle-mark-mode, \\[rectangle-mark-mode]

It doesn't work with the line selection command:
evil-visual-line,    \\[evil-visual-line]"
```

syl20bnr:
Thank you for the PR, closing it as it has been superseeded by #8491

PR #8207, opened 2017-01-18
adds key binding eshell-mode

FrancescElies:
I would find it handy to be able to switch to dired with a shortcut in eshell mode

What do you think?


Stebalien:
`SPC f j`?

FrancescElies:
Yes absolutely right, that's the way was using it before. But I thought it would be useful to have it under `, d`
I'm easy, if you think it's completely unnecessary I'll close the PR

Stebalien:
I'd rather type an extra key than memorize two but that may just be me. Regardless, the function should probably be `dired-jump`, not `projectile-dired` (only works in projects).

FrancescElies:
Makes sense to me too, I'm fine with typing an extra key, closing

PR #8208, opened 2017-01-19
Reliable call to `ipython --version`

olivierverdier:
Problem: `ipython --version` returns a bogus escape character `ESC[?1034h`

This is due to the following behaviour of the readline library:
http://stackoverflow.com/q/15760712
The solution: unset the `TERM` shell variable before calling `ipython --version`

### DIFF
--- a/layers/+lang/python/funcs.el
+++ b/layers/+lang/python/funcs.el
@@ -28,7 +28,7 @@
 (defun spacemacs/python-setup-shell (&rest args)
   (if (spacemacs/pyenv-executable-find "ipython")
       (progn (setq python-shell-interpreter "ipython")
-             (if (version< (replace-regexp-in-string "\n$" "" (shell-command-to-string "ipython --version")) "5")
+             (if (version< (replace-regexp-in-string "\n$" "" (shell-command-to-string "unset TERM; ipython --version")) "5")
                  (setq python-shell-interpreter-args "-i")
                (setq python-shell-interpreter-args "--simple-prompt -i")))
     (progn


bmag:
> Problem: ipython --version returns a bogus escape character ESC[?1034h

Why do you say that? There's no escape character in my session: (text copied from IELM buffer)
```
ELISP> (shell-command-to-string "ipython --version")
"5.1.0\n"
ELISP> (shell-command-to-string "ipython --simple-prompt --version")
"5.1.0\n"
ELISP> (shell-command-to-string "echo $TERM")
"dumb\n"
```

olivierverdier:
There is indeed no escape character in IELM, or when running the command from `SPC-SPC`. This probably because the `TERM` variable is then correctly unset.

However, when calling `shell-command-to-string "ipython --version")` during emacs starting up, this causes a warning. The warning message is:
`Error (use-package): python :init: Invalid version syntax: '5.1.0
[?1034h'`

I traced it back to the import of `readline` in the call to `ipython --version`, in conjunction with the `TERM` shell variable.

I don't know why this `TERM` variable is taken into account, or why it is set this way. Perhaps this is because I start emacs from a terminal where `TERM` is set?

olivierverdier:
Ok, indeed, if I start emacs using `unset TERM; emacs`, then the problem disappears. You can close this PR if you want, as I might be the only one affected by this problem (something specific to my config, perhaps?).

Still, it is a bit worrying that `shell-command-to-string` uses the original `TERM` shell variable.

bmag:
> Still, it is a bit worrying that shell-command-to-string uses the original TERM shell variable.

Yes, there is a problem here. Can you open an issue with some reproduction steps, so we can better track this an issue?

Also, does using `ipython --simple-prompt --version` return the same escape character?

olivierverdier:
> does using ipython --simple-prompt --version return the same escape character?

Yes. Apparently, any call to `ipython` returns that character. This is due to the `readline` library.

> Can you open an issue with some reproduction steps, so we can better track this an issue?

Ok, I'll do that.

olivierverdier:
I've submitted the issue #8209 instead.

Issue #7563, opened 2016-10-27
Geolocation theme changer doesn't work on fresh start of Emacs


iamthad:
#### Description :octocat:

Geolocation theme changer doesn't work on fresh start of Emacs. 
It does work after <SPC> f e R to reload .spacemacs, however.

It looks like the dotspacemacs-themes variable isn't initialized when the theme changer tries to load.
#### Reproduction guide :beetle:
- Start Emacs

_Observed behaviour:_ :eyes: :broken_heart:
The following error appears in the Warnings buffer:

```
Error (use-package): theme-changer :config: Wrong type argument: arrayp, nil
```

and the theme does not change

_Expected behaviour:_ :heart: :smile:
If appropriate, the theme changes to the night theme, and no error message appears.
#### System Info :computer:
- OS: darwin
- Emacs: 25.1.1
- Spacemacs: 0.200.4
- Spacemacs branch: master (rev. 6feb70d)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:

``` elisp
(helm auto-completion c-c   emacs-lisp evernote
      (geolocation :variables geolocation-enable-automatic-theme-changer t)
      git gnus html imenu-list javascript latex markdown org pandoc puppet python semantic
      (shell :variables shell-default-height 30 shell-default-position 'bottom)
      shell-scripts spell-checking syntax-checking vagrant vim-empty-lines yaml)
```
- System configuration features: NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS

iamthad:
A crude hack that does appear to fix this issue is to add 

    :defer 1

to the use-package declaration in the geolocation/init-theme-changer function in layers/+tools/geolocation/packages.el

This silences the warning and makes the theme change, but it takes ~ 1 second due to the defer.

namdnguyen:
The defer option didn't seem to fix it for me, but `f e R` after starting up spacemacs.

namdnguyen:
I gave it another try. Turns out I had a couple of errors in my `user-init` section. Some lines belonged in `user-config`.

Theme-changer still didn't change the theme after I made this fix, but I tried the `:defer 1` method again, and it worked for me this time.

joshcass:
Yeah, the `:defer' has been working for me as well. It's a bit of an unfortunate hack but not a huge deal.

Novite5:
[...]

PR #8212, opened 2017-01-19
Defer loading of theme changer to work around #7563

iamthad:
Add `:defer 1` to `use-package` declaration for the geolocation theme changer to work around issue #7563.


idoo:
@syl20bnr can you please take a look?

TheBB:
This is a bit of a hack that I think we can do without. Wouldn't it be easier to add a function to `spacemacs-post-user-config-hook` to ensure it runs after everything else is done, and therefore `dotspacemacs-themes` is present?

iamthad:
Please see new proposed fix #9081.

PR #8170, opened 2017-01-14
Fixed a typo

LdBeth:

PR #8215, opened 2017-01-19
Fix #8192 correct source block variable name

### LINKED ISSUE #8192: flyspell suggestions don't seem to be working
(sorry for not using the bug report form--will do it next time)

I'm, windows 10, GNU Emacs 25.1.1 (i686-w64-mingw32), spacemacs master branch.

**I am not able to get spelling suggestions work. Presumably it should work out of the box. When I see a misspelled word, I can see a hand cursor, which suggests that the word is clickable, but I nothing happens when you click.**

It is using the right dictionaries and gives suggestion, but not in `flyspell-mode`:
image

image

```lisp
;;layer config
 spell-checking
     (spell-checking :variables
                     =enable-flyspell-auto-completion= t)
     syntax-checking

;;user
(defun dotspacemacs/user-config ()
  "Configuration function for user code.
This function is called at the very end of Spacemacs initialization after
layers configuratuion.
This is the place where mosst of your configurations should be done. Unless it is
explicitly specified that a variable should be set before a package is loaded,
you should place your code there."
  (setq ispell-program-name "c:/msys64/mingw64/bin/hunspell.exe")
  (setenv "LANG" "en_US")
  (setq ispell-chage-dictionary "en_US")
  (setq ispell-hunspell-dict-paths-alist
        '(("en_US" "c:/msys64/mingw64/share/hunspell/en_US.aff")
          ("ru_RU" "c:/msys64/mingw64/share/hunspell/ru_RU.aff")))
  )
```

I'm also wondering about what all those out of the box options in `ispell-change-dictionary` in helm are for. Does emacs come with ispell dictionaries pre-loaded or something? —cause I only installed two languages. If so, can emacs even use all these dictionaries once you set the `ispell-program-name` to hunspell?

(btw; it might be a good idea to just ship hunspell with the package for windows users—its pretty small, but it's a pain to figure out where and how to get it (msys2), and the config step is also difficult, because apperently on WIN you have to set  `(setenv "LANG" "YOUR_LANG")` unix-style, otherwise hunspell wont work at all [see here](http://emacs.stackexchange.com/questions/2171/what-options-are-there-for-writing-better-non-programming-text-in-emacs)

Thanks again for all your effort!

nixmaniack:
Ref: https://github.com/syl20bnr/spacemacs/issues/8192#issuecomment-273905702

### DIFF
--- a/layers/+checkers/spell-checking/README.org
+++ b/layers/+checkers/spell-checking/README.org
@@ -85,7 +85,7 @@ set the layer variable =enable-flyspell-auto-completion= to t:
 
 #+BEGIN_SRC emacs-lisp
 (setq-default dotspacemacs-configuration-layers
-  '((spell-checking :variables =enable-flyspell-auto-completion= t)))
+  '((spell-checking :variables enable-flyspell-auto-completion t)))
 #+END_SRC
 
 * Key Bindings


joehillen:
This is a dup of #8170

nixmaniack:
Ah I didn't see that.

joehillen:
I don't blame you. The other one is badly named. I think this one should be used and the other one closed instead.

Leave this open maybe someone will notice.

nixmaniack:
Closing this as #8253 has been merged. We can close #8170 and #8192. @d12frosted

PR #8217, opened 2017-01-20
Move installation info to beginners tutorial

nikolaiam:
Installation instructions are moved from README.md to the beginners tutorial with some clarifications in the text along the way. The user now has a single source for a comprehensive installation walkthrough which is accessible with `SPC h SPC` and will presumably be accessible from spacemacs.org (at least if https://github.com/syl20bnr/develop.spacemacs.org/pull/7 is accepted) unlike the README.md.


syl20bnr:
Thank you for the PR.

I don't think this is a good move. First if people can `SPC h SPC` in Emacs then they have already installed it with Spacemacs. Second it is a popular practice to put the installation instruction in the README.md which is the default file displayed on the repo page.

The real issue to fix here is to have an easy access to installation instruction form the website, this should be addressed on the develop website repo, not in Spacemacs: https://github.com/syl20bnr/develop.spacemacs.org

nikolaiam:
I think you're right, I'll close this one!

Issue #5106, opened 2016-02-17
Yanking and pasting in Haskell mode inserts spurious space at beginning of line


nmk:
In haskell-mode yanking and pasting a line with `yyp` indents the newly pasted line by one space. So pressing `yypyypyyp` on a line containing `foo` results in
```
foo
 foo
  foo
   foo
```
Any idea what might be going on there? I can not see this in other modes, so I assume it is solely haskell-mode related.

I am using the develop branch and the latest haskell-mode from MELPA.

kRITZCREEK:
this is actually really annoying...

StreakyCobra:
There are already problems with haskell indentation:

https://github.com/syl20bnr/spacemacs/issues/621
https://github.com/syl20bnr/spacemacs/issues/5010

I'm pretty sure it's an upstream issue, but it would be nice if someone can try on a stock Emacs with just `haskell-mode` enabled.

CC @d12frosted

fice-t:
Is this really an upstream issue? It appears that after pasting, `spacemacs/yank-advised-indent-function` gets called, which calls `indent-region`, which would produce this result.

This of course doesn't happen on my stock Emacs.

StreakyCobra:
> This of course doesn't happen on my stock Emacs.

So it's a problem on our side

syl20bnr:
Quick solution is indeed to add haskell-mode to `spacemacs-indent-sensitive-modes`.

Work around is to use the universal prefix argument when pasting to disable the indentation but I'm aware this is pretty inconvenient.

spencerturkel:
How does one add haskell-mode to spacemacs-indent-sensitive-modes?

rrottier:
I am having the same issue but with the stable branch (0.105.11)

robbyoconnor:
I may be wrong but: 

`(add-to-list spacemacs-indent-sensitive-modes 'haskell-mode)`

robbyoconnor:
PR coming

robbyoconnor:
@nmk @spencerturkel, @kRITZCREEK: pull my PR local

rpglover64:
@fice-t @StreakyCobra I consider this an upstream issue:

From https://github.com/haskell/haskell-mode/issues/796:
> rpglover64: I think this is a misfeature in `haskell-indentation-indent-region`.
> gracjan: Region indentation can be considered unimplemented feature. We do not have that working yet.

See also: https://github.com/haskell/haskell-mode/issues/1032

[...]

PR #8223, opened 2017-01-21
Add asm-mode to the spacemacs-indent-sensitive-modes list

smalukav:
I found bug in asm indention. It's similar to #5106, and fix also similar. Just add asm-mode to spacemacs-indent-sensitive-modes.


syl20bnr:
Thank you ! 👍 

PR #8226, opened 2017-01-21
Python send cell

krakapwa:
This commit emulates MATLAB's cell function in python-mode, i.e. execute a block of code placed within tags. This is more practical then python-shell-send-region since it selects the appropriate region automatically. There is of course no convention for which tag to use. In this commit, I am using double hash "##".

Example:

print("First line.")
##
print("Second line.")
print("Third line.")
print("Fourth line.")
##

Place cursor within ## and execute python-shell-send-cell


TheBB:
I don't believe `##` is an accepted standard, nor is there an accepted standard for such a thing. If I'm wrong, please let me know, but for the moment I think we don't need this. Sorry, and thank you for the PR.

PR #8228, opened 2017-01-22
Change the frame-killer mapping to `SPC q f`

Stebalien:
It was `SPC q z` but:

1. 'z' is not a mnemonic for "frame" (although it may be more familiar to vim users).
2. More importantly, it's *really hard* to type 'qz' on a qwerty keyboard.


syl20bnr:
Thank you ! 👍 

PR #8230, opened 2017-01-22
syntax-checking: Mention that flycheck is disabled in emacs-lisp-mode.

mineo:


edrex:
Looks good to me, 👍 .

I was confused by the existing text: When `flycheck-mode` wasn't being enabled in `protobuf-mode`, I checked if it was being enabled in `emacs-lisp-mode`, and when it wasn't, I concluded incorrectly that it wasn't being enabled for any mode.

d12frosted:

PR #8231, opened 2017-01-22
keyboard-layouts: Reword installation instructions

alcrene:
Remove some of the ambiguity in the installation instructions for
the layer keyboard-layouts.
As they were, one could misinterpret them to mean first adding
the layer, and *then* adding the given code somewhere else in 
the .spacemacs file.


syl20bnr:
Thank you for the PR!
You are right about the ambiguity and this is something we discussed in the past, we chose the current formulation even if it is still ambiguous, we believe this ambiguity should be fixed by a tutorial instead.
As this is the default formulation in *all* README.org files I prefer to close this PR and keep the same sentence everywhere.

PR #8232, opened 2017-01-22
Fix keybindings for ess layer

YiLiu6240:
Change keybindings for sending function to REPL to be in line with
[Spacemacs convention](https://github.com/syl20bnr/spacemacs/blob/develop/doc/CONVENTIONS.org#send-code)


syl20bnr:
Nice catch 💜 
Thank you !

PR #7027, opened 2016-09-05
Add language layer for Protocol Buffers.

amol-mandhane:
Resolves #6973

PR #8234, opened 2017-01-22
Add protobuf layer

edrex:
Based on #7027, but updated to omit the defunct flycheck-protobuf layer.

@amol-mandhane @zhexuany @d12frosted please take a look

### DIFF
--- a/layers/+lang/protobuf/README.org
+++ b/layers/+lang/protobuf/README.org
@@ -1,0 +1,23 @@
+#+TITLE: Protocol Buffers layer
+#+HTML_HEAD_EXTRA: <link rel="stylesheet" type="text/css" href="../../../css/readtheorg.css" />
+
+* Table of Contents                                        :TOC_4_org:noexport:
+ - [[Description][Description]]
+ - [[Features][Features]]
+ - [[Install][Install]]
+
+* Description
+Highlighting and syntax checking for [[https://developers.google.com/protocol-buffers/][Protocol Buffer]] files.
+
+Uses Google's [[https://github.com/google/protobuf/blob/master/editors/protobuf-mode.el][protobuf-mode]] with [[https://github.com/edvorg/flycheck-protobuf][flycheck-protobuf]].
+
+* Features
++ Syntax Highlighting
++ Syntax checking using Flycheck (=protoc= compiler must be available)
++ Correct Indentation and commenting
++ Quickly browse file contents using =M-x helm-imenu= and =M-x imenu=.
+
+* Install
+To use this configuration layer, add it to your =~/.spacemacs=. You will need to
+add =protobuf= to the existing =dotspacemacs-configuration-layers= list in this
+file.
--- a/layers/+lang/protobuf/packages.el
+++ b/layers/+lang/protobuf/packages.el
@@ -12,7 +12,6 @@
 (setq protobuf-packages
       '(
         protobuf-mode
-        flycheck-protobuf
         ))
 
 (defun protobuf/init-protobuf-mode ()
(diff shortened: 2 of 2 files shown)


edrex:
This still doesn't enable `flycheck-mode` automatically for `protobuf-mode` buffers, for me at least. I'm not sure if that's an issue in the `flycheck` package or this change.

zhexuany:
Hi, 
`flycheck-protobuf` has a prerequisite: 

you have to install `protoc` first. I think this is a good start point to investigate this.

edrex:
I've verified that flycheck is enabled automatically for protobuf buffers in stock emacs:

```
SPC q D
flycheck,protobuf-mode
M-: global-flycheck-mode <Enter>
<open a protobuf file and note that flycheck-mode is enabled for that buffer>
```

So I'm not sure why it's not being enabled in Spacemacs. Presence or absence of 

```
(defun protobuf/post-init-flycheck ()
  (spacemacs/add-flycheck-hook 'protobuf-mode))
```

seems to make no difference.

edrex:
@zhexuany thanks for replying. I can enable `flycheck-mode` manually for the buffer (eg using `SPC t s`) so I don't think the prerequisite is the issue. Can you verify that it's not being enabled automatically with this patch in your setup?

theckman:
@edrex I managed to get it to automatically load. But GitHub isn't letting me issue a pull request against your branch. 😒 

LMK if this works for you: https://github.com/theckman/spacemacs/commit/a4c5d95363e97ef22a7153a4787ed96f98a92560

edrex:
@theckman with your change it's automatically enabling the minor mode for me too. It seems that I was missing flycheck from protobuf-packages. I reset my branch to match yours. Thanks!

This is ready for review.

theckman:
@edrex I'm fairly ignorant to getting layers and things working with Spacemacs (and emacs in general), so I apologize if I'm missing something that should be very obvious. The one thing that seems to not work, and I'm not sure if we have the ability to make it work, is automatic completion around things with braces.

For example, if I'm writing `Go` which has a similar syntax to Proto3, the `}` in this snippet would get auto-added when I type `{`:

```Go
type Test struct {}
```

However, in the `protobuf-mode` files when I type `{` I don't get the same completion:

```protobuf
message Request {
```

edrex:
me too 😄 

The brace matching would need to be added to the upstream mode I think:
https://github.com/google/protobuf/blob/master/editors/protobuf-mode.el

Doesn't seem like a blocker for getting this in, right?

theckman:
Yeah, I was just starting to look there to see if maybe that was the case. Bit of a yak shave. 😛 

Nope! I do not think that should be considered a blocker at all.

wanchaol:
Thanks for the effort! When will this get merged?

rargulati:
I've been testing this on my local development environment for a few weeks now and it works great! Anything I can do here to help move this along?

edrex:
I think it's ready to go and just needs a signoff. Maintainers are always swamped because of the volume of pull requests. Not sure the best way to help.

edrex:
https://nolanlawson.com/2017/03/05/what-it-feels-like-to-be-an-open-source-maintainer/

rgrinberg:
Only @syl20bnr merges new layers. So we should ping him.

Hopefully, it doesn't take too long to merge this. It seems like a very small start for a new layer.

theckman:
It might be slightly inappropriate, but I tried to use Twitter to ping @syl20bnr. Based on the date of Sylvain's last tweet I'm not sure it'll get seen. But I'm hopeful that this will be a small enough PR that he won't need to spend too much personal time on reviewing it.

theckman:
With `0.200.8` out, it sounds like @syl20bnr may have some free time to :eyes: outstanding PRs:

* https://twitter.com/syl20bnr/status/849067404533563394

PR #8259, opened 2017-01-26
Clojure refactor: replace cycle coll with convert to coll

benedekfazekas:
Cycle collection type was recently removed from `clj-refactor` as the
feature was migrated and rewritten in `clojure-mode`. The new feature
lets the user convert a collection into a specific collection type
instead of cycling through them.

To added shortcuts correspond to the shortcuts of this feature in
`clojure-mode` and placed in the refactor submenu.


d12frosted:

benedekfazekas:
ty for the quick response

Issue #8246, opened 2017-01-25
Documentation for replacing text in several files using helm-ag is incorrect


dqdinh:
#### Description :octocat:
The documentation about using helm-ag to replace text in multiple files no longer works.
(http://spacemacs.org/doc/DOCUMENTATION.html#replacing-text-in-several-files)

In order to get this to work, you have to follow a new sequence of steps:
- Start Emacs
- `SPC /`
- search for a text pattern
- `C-z`
- select action: [f4]  Edit search results
- select text
- `SPC s e`
- modify text
- `C-c C-c`

#### Reproduction guide :beetle:
- Start Emacs
- `SPC /`
- search for a text pattern
- `C-c C-e`
- select text
- `SPC s e`
- modify text
- `C-c C-c`

*Observed behaviour:* :eyes: :broken_heart:
`C-c C-e` is undefined

*Expected behaviour:* :heart: :smile:
It should replace selected text across multiple files if you complete the sequence of commands.

#### System Info :computer:
- OS: darwin
- Emacs: 25.1.1
- Spacemacs: 0.200.7
- Spacemacs branch: master (rev. c2774bc)
- Graphic display: nil
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(asciidoc auto-completion better-defaults git helm html javascript markdown python ruby scala vinegar yaml)
```
- System configuration features: JPEG RSVG NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS MODULES

dvzubarev:
Why should we change documentation? Why not reporting upstream? I think it is a bug. Am I missing something? There is already issue (#8254) about it in spacemacs.

dvzubarev:
I can't reproduce this issue.
- Emacs: 25.1.2
- Spacemacs: 0.200.9
- Spacemacs branch: master (rev. 8e1af14)
- helm-20170414
- helm-ag-20170209

PR #8264, opened 2017-01-26
Update steps for "Replacing text in several files"

dqdinh:
In the current latest stable version of helm (2.5.0), steps for "Replacing text in several files" don't work. This updates the documentation and also adds a note about doing the same thing on older version of helm.

https://github.com/syl20bnr/spacemacs/issues/8246


syl20bnr:
Thank you for the PR!
As you know it has been fixed upstream so we don't need this PR anymore. Closing this PR.

PR #8265, opened 2017-01-27
`.python-version` can contain more than one version specified.

s0undt3ch:
https://github.com/yyuu/pyenv/blob/master/COMMANDS.md#pyenv-local-advanced

This assures that we only take into account the first one.


syl20bnr:
Thank you !

PR #8266, opened 2017-01-27
ui-visual: improve colorization of compilation buffer

aroig:
Lines containing a carriage return alone, followed by color escape sequences, are not colorized properly.

A carriage return in the compilation buffer is interpreted by erasing contents since the beginning of the line, but `compilation-filter-start` is not updated accordingly before calling the filter-hooks.

Workaround it by always colorizing a region from to the beginning of the line where `compilation-filter-start` points.


syl20bnr:
Thank you ! 👍 

PR #8272, opened 2017-01-29
Revert "Make pylookup.py compatible with Python 2 and 3"

Hoot215:
This reverts commit 9d1b5745541df5ac38d700d0032ba3828cb55cf1.

This commit was apparently not tested with Python 3, since it breaks pylookup entirely when "python" is linked to "python3" instead of "python2". Making it work with Python 3 would require a rewrite of the HTML parsing code, since html.parser in Python 3 handles tags and data differently than htmllib in Python 2. For now, we should just stick to Python 2.

### DIFF
--- a/layers/+lang/python/local/pylookup/pylookup.py
+++ b/layers/+lang/python/local/pylookup/pylookup.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 
 """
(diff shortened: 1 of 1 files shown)


syl20bnr:
Thank you for your PR. We recently fixed support for python 3 in develop branch, can you check if this works for you ?

Hoot215:
Works great. Thanks!

PR #8276, opened 2017-01-30
Fix ranger layer documentation

cyberxndr:
Updated entry for "show dotfiles at ranger startup" option.
Old variable had no effect (possibly deprecated)

Updated entry for "show previews literally" option.
Old variable had no effect (possibly mistype)


d12frosted:
Please do not remove your fork in the future. 

Thank you :+1: Merged into `develop`, you can safely delete your branch.

PR #8283, opened 2017-01-31
minor elisp convention fix

deb0ch:
Not much, just replacing some `if`s with `when`s that I spotted when looking through recent commits.


d12frosted:

PR #8295, opened 2017-02-02
haskell: disable electric indent for cabal-mode

bennofs:
Electric-indent mode in cabal-mode is annoying, as it messes up existing indentation.
For example, if you have:

```
  build-depends:
    base,
    foo,|
```

And press enter, then that ends up as:

```
  build-depends:
    base,
                 foo
                 |
```

(`|` is cursor)

This PR fixes that issue.

### DIFF
--- a/layers/+lang/haskell/funcs.el
+++ b/layers/+lang/haskell/funcs.el
@@ -81,6 +81,12 @@
   (evil-define-key '(insert normal) intero-mode-map
     (kbd "M-.") 'intero-goto-definition))
 
+(defun spacemacs-haskell//disable-electric-indent ()
+  "Disable electric indent mode if available"
+  ;; use only internal indentation system from haskell
+  (if (fboundp 'electric-indent-local-mode)
+      (electric-indent-local-mode -1)))
+
 
 ;; Intero functions
 
--- a/layers/+lang/haskell/packages.el
+++ b/layers/+lang/haskell/packages.el
@@ -92,6 +92,11 @@
       (add-hook 'haskell-cabal-mode-hook
                 'spacemacs//force-haskell-mode-loading)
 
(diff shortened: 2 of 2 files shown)


joehillen:
The convention is to list all the packages after the colon and align with a leading `,`

https://github.com/commercialhaskell/stack/blob/master/stack.cabal#L185

For example:

```cabal
build-depends: base
             , foo
             , bar
             |
```

bennofs:
@joehillen: well, but you can still do that even with electric indent mode disabled, as pressing enter will keep the current indent level. So if you are at:

```
build-depends: base
             , foo|
```

Then pressing enter will still result in:

```
build-depends: base
             , foo
             |
```

So you can still use that style just fine with electric indent disabled. But there's lots of packages in the wild that use this style:

```
build-depends:
  base,
  foo,
  bar
```

(Example: [lens](http://hackage.haskell.org/package/lens-4.15.1/lens.cabal or text or snap)

And that style is *completely* broken with electric indent on: you end up with:

```
build-depends:
  base,
  foo,
               | 
```

And now you have to delete all those spaces manually using backspace! So you need to press backspace like 10 times, which is really annoying.

So looking at this tradeoff, I conclude the following:

* with electric-indent mode **off**, you can use both styles just fine without major inconvenience.
* with electric-indent mode **on**, using the alternative style is very annoying.

joehillen:
Oh, I didn't realize spacemacs already disables electric indent for `*.hs` files, so this is just doing it for cabal files as well. I'm fine with that then.

bennofs:
What's the status of this?

TheBB:

I removed the dangling function: https://github.com/syl20bnr/spacemacs/commit/144c37955f5081d135ac0f39824b68eafdb5d31b

PR #8301, opened 2017-02-03
mu4e-compose: add some convenient major mode bindings

xificurC:
I'm missing some nice shortcuts for sending mail, attaching files etc. These bindings are similar to that of magit. Open to further discussion

### DIFF
--- a/layers/+email/mu4e/packages.el
+++ b/layers/+email/mu4e/packages.el
@@ -55,8 +55,9 @@
         "," 'message-send-and-exit
         "c" 'message-send-and-exit
         "k" 'message-kill-buffer
+        "a" 'message-kill-buffer
         "s" 'message-dont-send         ; saves as draft
-        "a" 'mml-attach-file)
+        "f" 'mml-attach-file)
 
       (setq mu4e-completing-read-function 'completing-read)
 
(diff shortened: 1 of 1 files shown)


fandag:
Even though putting "attach" on `a` is tempting, I think it would be better to follow [conventions](http://spacemacs.org/doc/CONVENTIONS.html#confirm-and-abort) have "abort" on `a`.

xificurC:
@fandag didn't know about those conventions, thanks. I changed the bindings to follow them.

TheBB:

Note that the `,,` convention is not just `dotspacemacs-major-mode-leader-key` followed by `,` but actually a double tap of `dotspacemacs-major-mode-leader-key`, so I made a change to reflect this: https://github.com/syl20bnr/spacemacs/commit/c7df45a0a9fbb52415796d6ad23bcfd1c297cab1 See e.g. https://github.com/syl20bnr/spacemacs/blob/develop/layers/+source-control/git/packages.el#L364 as an example for magit commit messages.

xificurC:
Thanks @TheBB , nice to know :) The bindings for `mu` are a bit of a struggle sometimes, e.g.

- in `mu4e-headers-mode` `p` moves back a message but `n` does not, although both `C-n` and `C-p` work
- in `mu4e-view-mode` `n` doesn't work and `p` does, *but* `C-n` works and `C-p` doesn't :D 

So if you're open to new changes I might flash up another PR

PR #8310, opened 2017-02-05
Fixes attempt to use non-existent function

alindeman:
Since https://github.com/dominikh/go-mode.el/commit/d13feb239b13b910ec4db19356f34c3801299407, `go--position-bytes` no longer exists.


bmag:
Closing this once since #8179 was merged.

Issue #8304, opened 2017-02-04
spacemacs won't work with hybrid-mode on develop


kain88-de:
#### Description :octocat:
Spacemacs won't start with hybrid mode. This happens after [this commit](https://github.com/syl20bnr/spacemacs/commit/c7e47de). Using the commit before everything works

#### Reproduction guide :beetle:
- enable hybrid mode
- Start Emacs

*Observed behaviour:* :eyes: :broken_heart:
hybrid-mode-use-evil-search-module is void 

*Expected behaviour:* :heart: :smile:
emacs starts fine

#### Backtrace :paw_prints:
```
Debugger entered--Lisp error: (void-variable hybrid-mode-use-evil-search-module)
  (and (eq (quote hybrid) style) hybrid-mode-use-evil-search-module)
  (or (eq (quote vim) style) (and (eq (quote hybrid) style) hybrid-mode-use-evil-search-module))
  (cond ((or (eq (quote vim) style) (and (eq (quote hybrid) style) hybrid-mode-use-evil-search-module)) (setq-default evil-search-module (quote evil-search))) (t (setq-default evil-search-module (quote isearch))))
  spacemacs/set-evil-search-module(hybrid)
  spacemacs-bootstrap/init-evil()
  funcall(spacemacs-bootstrap/init-evil)
  (let* ((pkg-name (eieio-oref pkg (quote :name))) (owner (car (eieio-oref pkg (quote :owners))))) (spacemacs-buffer/message (format "Configuring %S..." pkg-name)) (mapc (function (lambda (layer) (if (configuration-layer/layer-usedp layer) (progn (if (not ...) (spacemacs-buffer/message ...) (spacemacs-buffer/message ...) (condition-case err ... ...)))))) (eieio-oref pkg (quote :pre-layers))) (spacemacs-buffer/message (format "  -> init (%S)..." owner)) (funcall (intern (format "%S/init-%S" owner pkg-name))) (mapc (function (lambda (layer) (if (configuration-layer/layer-usedp layer) (progn (if (not ...) (spacemacs-buffer/message ...) (spacemacs-buffer/message ...) (condition-case err ... ...)))))) (eieio-oref pkg (quote :post-layers))))
  configuration-layer//configure-package([eieio-class-tag--cfgl-package evil nil (spacemacs-bootstrap) nil (ivy emacs-lisp) elpa t bootstrap nil t nil])
[...]

Issue #8305, opened 2017-02-04
"evil: enable evil-search search module in evil state" breaks init


fabianhjr:
Error
```
Warning (initialization): An error occurred while loading ‘/home/fabianhjr/.emacs.d/init.el’:

Symbol's value as variable is void: hybrid-mode-use-evil-search-module

To ensure normal operation, you should investigate and remove the
cause of the error in your initialization file.  Start Emacs with
the ‘--debug-init’ option to view a complete error backtrace.
```
Likely Change: https://github.com/syl20bnr/spacemacs/commit/c7e47de9383f7556553c531ff27e7654b3cba3db#diff-18cdada95f0eaec9a19a35c51c0dbe27R47

Debug
```
Debugger entered--Lisp error: (void-variable hybrid-mode-use-evil-search-module)
  (and (eq (quote hybrid) style) hybrid-mode-use-evil-search-module)
  (or (eq (quote vim) style) (and (eq (quote hybrid) style) hybrid-mode-use-evil-search-module))
  (cond ((or (eq (quote vim) style) (and (eq (quote hybrid) style) hybrid-mode-use-evil-search-module)) (setq-default evil-search-module (quote evil-search))) (t (setq-default evil-search-module (quote isearch))))
  spacemacs/set-evil-search-module(hybrid)
  spacemacs-bootstrap/init-evil()
  funcall(spacemacs-bootstrap/init-evil)
  (let* ((pkg-name (eieio-oref pkg (quote :name))) (owner (car (eieio-oref pkg (quote :owners))))) (spacemacs-buffer/message (format "Configuring %S..." pkg-name)) (mapc (function (lambda (layer) (if (configuration-layer/layer-usedp layer) (progn (if (not ...) (spacemacs-buffer/message ...) (spacemacs-buffer/message ...) (condition-case err ... ...)))))) (eieio-oref pkg (quote :pre-layers))) (spacemacs-buffer/message (format "  -> init (%S)..." owner)) (funcall (intern (format "%S/init-%S" owner pkg-name))) (mapc (function (lambda (layer) (if (configuration-layer/layer-usedp layer) (progn (if (not ...) (spacemacs-buffer/message ...) (spacemacs-buffer/message ...) (condition-case err ... ...)))))) (eieio-oref pkg (quote :post-layers))))
[...]

PR #8317, opened 2017-02-06
fix evil-search enabling in evil-state (c7e47de)

deb0ch:
There was an error at startup saying that hybrid-mode-use-evil-search-module was
void as a variable, apparently brought by c7e47de.


deb0ch:
Fixes #8304 #8305 and #8321, which are all duplicates :smile_cat:

kain88-de:
This PR fixes the problem for me. Thanks.

syl20bnr:
Thank you ! 👍 
Seems that the variable must be set before evil is required otherwise although the variable as the correct value, the real activated module is `isearch`. Attempted to fix it in https://github.com/syl20bnr/spacemacs/commit/eef1a87

PR #8320, opened 2017-02-06
store cider repl history in spacemacs cache

ryfow:
Configure Cider, so that the REPL history is saved between sessions.


TheBB:

PR #8326, opened 2017-02-07
Add on save usage section to elm layer

Baransu:
According to official elm-format documentation there is elm-format-on-save option. I've added it to layer documentation and introduced "On save usage" section which now contains info about elm format on save and imports sort on save.


syl20bnr:
Thank you ! 👍 

Baransu:
Thanks! 😊

Issue #8289, opened 2017-02-01
Lisp mode -> insert mode -> lisp mode


gitfoxi:
If you go from Lisp mode [SPC k .] to insert mode [i] then [ESC] it transitions you to normal mode. Why not go back to Lisp mode? Then you'd have to [ESC ESC] to go all the way back to normal mode but maybe that's what you want.

Issue #8288, opened 2017-02-01
Lisp mode leader keys


gitfoxi:
In Lisp mode [SPC k .] the leader SPC and mode-leader COMMA keys are unbound. Why not just have them work the same as in normal mode?

PR #8329, opened 2017-02-07
Lisp mode more like iedit-mode

gitfoxi:
#8288
#8289

1. Have insert state return to lisp state.
2. Make leader and mode-leader available in lisp state.


gitfoxi:
I did not mean for the chanelog.develop file to get in there. Github is playing games. Anyone know how to fix it?

gitfoxi:
Fixed #8485 and #8502

TheBB:
Would it make more sense to make this a PR to evil-lisp-state instead? It's @syl20bnr's package.

candeira:
Bump! I found this issue by having the same brainwave as the submitter. Is there any way I can help?

duianto:
I've applied this PR locally and fixed the conflicts.

It does restore `SPC` and `,` in `lisp-state`.

---

But when I enter lisp insert state then the mode line becomes blank.

It seems to only happen in some buffers, for example in `.spacemacs` and `evil-evilified-state.el`.
But the mode lines contents remain in `funcs.el` files (have tested two different `funcs.el` files) or in a new empty buffer `SPC b N n`.

#### A possible cause/solution
Adding back the `lisp-insert` color and appending a `1` to the color name seems to fix the blank mode line issue.:
```elisp
                                 ("lisp-insert" "HotPink1" (bar . 2))
```
https://github.com/syl20bnr/spacemacs/blob/11b258c71c35dec6d9228f01b68eae1987c9bca0/layers/%2Bdistributions/spacemacs-bootstrap/config.el#L79

The reason for adding a `1` to the color name, was because there doesn't seem to be a color with the name: `HotPink` in the color picker `SPC C l` (`helm-colors`)
only `HotPink` followed by the `1-4` numbers or with a space in between `Hot Pink`.

I chose `HotPink1` becuase it's the same color that the lisp state uses.
If the lisp state and lisp insert state colors should be different then any suggestions are welcome.

---

I'm not sure if:
```elisp
(evil-define-state lisp-insert
  "Replace insert state in `lisp state'."
  :tag " <El> "
  :enable (insert)
  :cursor (bar . 2)
  :message "-- LISP-INSERT --"
  :input-method t)
```
should be added in the file:
`~/.emacs.d/layers/+distributions/spacemacs-bootstrap/local/evil-evilified-state/evil-evilified-state.el`
since as the name suggests, everything in it is `evilified` related.

If it's better to extract it, then maybe it should be added to a new `lisp-state` file, or possibly somewhere else, I'm not sure where.

---

I pushed the conflict resolved commits to a branch on my fork if someone wants to test it:
https://github.com/duianto/spacemacs/tree/8329-lisp-mode-more-like-iedit-mode

I'm not sure if the commits should be squashed.

PR #8330, opened 2017-02-07
Adding keybinding to jump to currently open file in neotree

arjunurs:
Adding a keybinding to jump to the currently open file in neotree. The existing shortcut `SPC f t` requires toggling neotree to jump to current file.


bmag:
Thanks for suggesting this key binding. However, I think `fn` isn't a good mnemonic. I assume `n` is for "neotree", but that's in contrast to `ft` which stands for "tree" (I think), so it will be awkward to have use `n` for "neotree" when there's another key binding that also uses neotree. I prefer to use `fT` (for "tree", because it's a variant of opening a file tree), despite it being harder to reach. What do you think?

arjunurs:
@bmag Sounds good to me

bmag:

syl20bnr:
Nice and perfect choice of key bindings as capital letter often means to change focus to the other window involved in the action 💜

PR #8338, opened 2017-02-09
Org layer: Bind org-update-statistics-cookies

mkaito:
Add a binding for `org-update-statistics-cookies` as `, #`, which mimics the original `C-c #` binding.


TheBB:

PR #8339, opened 2017-02-10
Add current company backend indicator to mode line

amosbird:


amosbird:
patch to spaceline instead.

PR #8340, opened 2017-02-10
Add company backend spaceline segment

amosbird:


deb0ch:
Nice idea ! Seems useful. :smiley_cat:

I wonder if this should be more in Spacemacs or in Spaceline. What do you think @TheBB ?

TheBB:
This should be in Spaceline. I actually said that on gitter but I think @amosbird misunderstood me.

PR #8343, opened 2017-02-10
Set leader keys for projectile-rails minor mode instead of ruby's maj…

### LINKED ISSUE #7424: Rails comma keybindings not available when on a view file window
Working with rails on spacemacs, I found it would be incredibly useful to have rails comma keybindings enabled on view file windows. 
For example when implementing feature tests it is cumbersome to have to switch to any *.rb window in order to be able to `, t a` [rspec-verify-all - run all tests].

I am still learning how to customize spacemacs, but so far couldn't identify how to have those keybindings enabled on view file windows. 

Will gladly send a PR once I learn how to implement that. Meanwhile, any suggestions?

asok:
…or mode

Closes #7424

### DIFF
--- a/layers/+frameworks/ruby-on-rails/README.org
+++ b/layers/+frameworks/ruby-on-rails/README.org
@@ -30,60 +30,60 @@ file.
 * Key bindings
 ** Code Navigation
 
-| Key binding   | Description                                                     |
-|---------------+-----------------------------------------------------------------|
-| ~SPC m r f a~ | find localization file                                          |
-| ~SPC m r f c~ | find controller                                                 |
-| ~SPC m r f e~ | find environment file                                           |
-| ~SPC m r f f~ | find feature                                                    |
-| ~SPC m r f h~ | find helper                                                     |
-| ~SPC m r f i~ | find initializer                                                |
-| ~SPC m r f j~ | find javascript file                                            |
-| ~SPC m r f l~ | find library                                                    |
-| ~SPC m r f m~ | find model                                                      |
-| ~SPC m r f n~ | find migration                                                  |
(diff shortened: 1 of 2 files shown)


asok:
Actually this overwrites a keybinding for web-mode accessible via `,r`.
So it cannot be merged in like that. Maybe someone could suggest a solution? Move the web-mode's keybinding from `,r` to somewhere else? This does not sound like a good approach since the projectile-rails minor mode could be on in other major modes also. So it would need to take care of them also (yaml-mode, compilation-mode, magit-mode etc...)

syl20bnr:
We should move `ruby-on-rails` key bindings as `SPC m r ` is for refactoring. Since it is a web framework, we could put them under `SPC m f` ? This prefix is free, we have no convention on it yet.

syl20bnr:
I added the convention: https://github.com/syl20bnr/spacemacs/commit/90452329

asok:
I've updated the PR so the keybindings are start with the `f` letter. Now the experience is much, much better. You can reach the projectile-rails' keybindings from any buffer with the mode enabled.

asok:
@syl20bnr I'm using this for the last week and it works perfectly. I strongly advise for merging this in :)

sdwolfz:
@asok Can you please also add these new keybindings to the documentation?

asok:
@sdwolfz done

sdwolfz:
Awesome! Thanks @asok, I'll take a look and merge it this weekend.

sdwolfz:
Thank you :heart:!

I reverted your change from `SPC m` to `,` in the documentation since all the other layers use `SPC m`: 3873ed52fe28b5cb8ebb00a34761b4caf63bb597.

Issue #8297, opened 2017-02-02
TypeScript formatting in TSX files


victorandree:
#### Description :octocat:
Typescript formatter cannot handle TSX files.

* tide does not appear to honor `tsfmt.json` rules
* `typescript-formatter` mangles the output

#### Reproduction guide :beetle:
- Create a TypeScript project with the following files:

  - tsfmt.json
  
  ```json
  {
    "indentSize": 2,
    "tabSize": 2,
    "newLineCharacter": "\n",
    "convertTabsToSpaces": true
  }
  ```
  
  - tsconfig.json
  
  ```json
  {
    "compileOnSave": false,
    "compilerOptions": {
      "newLine": "LF"
    }
  }
  ```
  
  - src/MyComponent.tsx
  
  ```tsx
  import * as React from 'react';

  const MyComponent: React.SFC<any> = props => (
      <div className="hello">
          <span className="label">Hello</span>
          <span className="value">Value</span>
      </div>
  );

  export default MyComponent
  ```

- Attempt to format

  - Using `SPC m =` won't work because you're in Web mode for TSX files
  - Call `typescript/tsfmt-format-buffer` to use typescript-formatter
  - Call `tide-format` to use tide format directly
  
*Observed behaviour:* :eyes: :broken_heart:

- Trying to use `SPC m =` formats everything as HTML

- With `typecsript/tsfmt-format-buffer`:

  ```
  import * as React from 'react';

  const MyComponent: React.SFC<any> = props => (
    <div className="hello" >
    <span className="label"> Hello < /span>
      < span className= "value" > Value < /span>
        < /div>
  );

  export default MyComponent
  ```

- With `tide-format` I get no changes

*Expected behaviour:* :heart: :smile:

- I expect `tide-format` to honor my tsfmt.json
- I expect output from `tsfmt` to be correct. Calling `tsfmt src/index.tsx` gives me this output:

  ```
  import * as React from 'react';

  const MyComponent: React.SFC<any> = props => (
    <div className="hello">
      <span className="label">Hello</span>
      <span className="value">Value</span>
    </div>
  );

  export default MyComponent
  ```
  
  which is correct.

#### System Info :computer:
- OS: darwin
- Emacs: 25.1.1
- Spacemacs: 0.200.7
- Spacemacs branch: master (rev. c2774bc0)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(csv html javascript
[...]

PR #8344, opened 2017-02-11
Call tsfmt with extension of current buffer for TSX formatting

victorandree:
Addresses #8297 for the tsfmt backend. However, does not fix tide issue, and automatic major mode formatting (`SPC m =`) does not work in TSX files as they are opened in web mode.


TheBB:

PR #8346, opened 2017-02-11
compilation-finish-function: Set case-fold-search to nil

mineo:
The various string-match calls in this function were obviosly written with that
in mind.

This also fixes the case where the compilation buffer contains the output of
`cargo test`, which includes "0 failed". Previously, this was treated as an
error, now it's not.


TheBB:

PR #8353, opened 2017-02-12
Add shortcuts for missing latex folding functions

nashamri:
Just a few keybindings for the `latex` layer. All the current folding functions do not have a "clearout" counterpart. And I've added a "dwim" keybinding.
:smile:


bmag:
Thank you :+1: Cherry-picked into develop, you can safely delete your branch. I added a comment about autoloading in https://github.com/syl20bnr/spacemacs/commit/6f463c4d0e7e7902be7a06e2dd39d6113375232b.

PR #8357, opened 2017-02-13
fix error when nlinum is enabled and dotspacemacs-line-numbers is t

deb0ch:
I don't know if there was a bug report for this one, I was told in the gitter chat.


bmag:

PR #8362, opened 2017-02-14
Keybindings for alchemist-phoenix

### LINKED ISSUE #8327: Elixir: include alchemist-phoenix commands in SPC-m tree
It would be really helpful to have the phoenix related commands available from the SPC menu. For reference, here's a screenshot of what's available from the alchemist docs:

image

The list is available to browse [here](https://alchemist.readthedocs.io/en/latest/basic_usage/)

sdwolf:
Refs #8327

The choice of the `SPC m n` prefix for the keybindings was chosen to be similar with the `C-c a n` one. `SPC m p` seems to have been chosen to mean `project`. Also `n` can be seen as short for `nix` which can be seen as short for `phoenix` :smile:.

If this is not desirable I can switch it to `SPC m p` or `SPC m a p` (but this seems a bit to long), or a better suggestion from the audence.

### DIFF
--- a/layers/+lang/elixir/README.org
+++ b/layers/+lang/elixir/README.org
@@ -14,6 +14,7 @@
    - [[#help][Help]]
    - [[#mix][Mix]]
    - [[#project][Project]]
+   - [[#alchemist-phoenix][Alchemist Phoenix]]
    - [[#evaluation-in-place][Evaluation in place]]
    - [[#repl-interactions][REPL interactions]]
    - [[#tests][Tests]]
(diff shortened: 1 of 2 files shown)


natecox:
This seems to work fine, I haven't noticed any issues with any of the commands.

Regarding the prefix, my only comment is that it's slightly cumbersome to type `SPC m n` as m and n are so close together, resulting in a lack of flow. E.g., viewing models becomes `SPC m n m`.

I'm not sure there's a _better_ option, though, and this is functionally complete.

sdwolf:
How about the `,` (comma) which is an alias for `SPC m` (also customizable in `.spacemacs`)? Does that help you improve your flow?

natecox:
Sounds good to me, at least I can use two different fingers for it now.

sdwolf:
Closing this in favor of #7192

Issue #5896, opened 2016-04-26
coffee-script doestn't auto complete


huaoguo:
#### Description

CoffeeScript doesn't auto complete
#### Reproduction guide
- Start Emacs
- Open a .coffee file
- write code

_Observed behaviour:_
doesn't have auto completion support

_Expected behaviour:_
has auto completion support
#### System Info
- OS: darwin
- Emacs: 24.5.1
- Spacemacs: 0.105.19
- Spacemacs branch: master (rev. 3274e7d)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:

``` elisp
(auto-completion emacs-lisp javascript git markdown react org clojure yaml)
```
#### Backtrace

```
BACKTRACE IF RELEVANT
```

TheBB:
Yes, there's no company backend for coffeescript. If you find one, do let us know.

huaoguo:
When I M-x invoke company-mode in coffee file, it realy works. So I think company-mode has support it, we just need to enable it?

TheBB:
You're getting the basic autocompletion then, I believe, not semantic. Not sure how useful that is, since it's pretty straightforward and knows nothing of the language you're writing.

huaoguo:
If you write coffeescript,you will know it is useful.

syl20bnr:
We can add company to the coffeescript hook then.

huaoguo:
Cool @syl20bnr

JuanCaicedo:
On the latest develop, I see company-disabled with coffescript by default. How do I add it to the coffeescript hook like you mentioned @syl20bnr ? Or should it already be enabled by default and I'm seeing a bug?

sdwolf:
@JuanCaicedo let me know if #8365 fixes your problem.

To check out my branch execute the following commands:
```bash
cd ~/.emacs.d
git remote add sdwolf https://github.com/sdwolf/spacemacs.git
git fetch sdwolf feature/coffee-dumb-autocomplete
git checkout feature/coffee-dumb-autocomplete
```

JuanCaicedo:
@sdwolf Thanks for the step by step instructions! I'll comment on your PR

PR #8365, opened 2017-02-14
Adds autocomplete for coffeescript

sdwolf:
Refs #5896 

This is a naive take on activating company-mode for CoffeeScript.

Reviews welcome!

### DIFF
--- a/layers/+lang/javascript/packages.el
+++ b/layers/+lang/javascript/packages.el
@@ -12,6 +12,7 @@
 (setq javascript-packages
   '(
     coffee-mode
+    company
     (company-tern :toggle (configuration-layer/package-usedp 'company))
     evil-matchit
     flycheck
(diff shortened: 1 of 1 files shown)


JuanCaicedo:
This works great for me! I'll leave the reviewing to someone with more experience with the source code :)

layers/+lang/javascript/packages.el
```diff
@@ -37,7 +38,8 @@
       (add-hook 'coffee-mode-hook
                 '(lambda ()
                    (setq indent-line-function 'javascript/coffee-indent
-                         evil-shift-width coffee-tab-width))))))
+                         evil-shift-width coffee-tab-width)
+                   (company-mode 1))))))
```
TheBB: Please rewrite to use `spacemacs|add-company-backends`.

sdwolf:
I've used `company-capf` since coffeescript does not have a dedicated backend (inspired from the `idris` layer). 

The `company` package is already in the `javascript-packages` list, is there any other place I need to add it as a dependency?

TheBB:
If you've added the package to the package list, you need to do the initialization that depends on that package in the `javascript/post-init-company` function. Not in `javascript/init-coffee-mode`. Otherwise people who use the javascript layer but not autocompletion will find themselves confused as to why they're getting errors about no function being called `company-mode`.

TheBB:

Issue #8309, opened 2017-02-04
Smarter alignment start/end points.


esiegel:
[feature-request]

It would be nice to have `spacemacs/align-repeat` method auto select its start and end points if start and end are the same.  Currently, you have to select a region and then type `SPC x a ,` to align.  I imagine this would be like tabular where we could search up and down lines to find the valid range.

esiegel:
Was cherry picked into development, 5be474f and code/reviewed fixed in 5c4b7e6

PR #8367, opened 2017-02-14
Smarter default alignment with spacemacs/align-repeat (#8309)

esiegel:
Fixes issue #8309 

This commit modifies `spacemacs/align-repeat` to scan up and down lines in search of smarter start and end points when a region is not specified.  It is inspired from the vim plugin Tabular.

Example:

```
one: 1
two: 2
three: 3
four: 4
```

with `SPC x a :` becomes without the need of selecting the lines you want to align

```
one:   1
two:   2
three: 3
four:  4
```

WARNING: this is my first code in elisp, so I apologize if it is way off mark.

### DIFF
--- a/layers/+distributions/spacemacs-base/funcs.el
+++ b/layers/+distributions/spacemacs-base/funcs.el
@@ -740,7 +740,22 @@ the right."
                               (concat regexp ws-regexp)
                             (concat ws-regexp regexp)))
          (group (if justify-right -1 1)))
+
     (message "%S" complete-regexp)
+
+    (if (equalp start end)
+        (progn
+          (save-excursion
+            (while (and
+                    (string-match-p complete-regexp (thing-at-point 'line))
+                    (equalp 0 (forward-line -1)))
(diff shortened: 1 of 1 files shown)


TheBB:

I made some changes: https://github.com/syl20bnr/spacemacs/commit/5c4b7e62b16bd7b074e4a21a04d4d17679bdd2a7

- You can use `when` or `unless` which works as an `if` without an `else`. Then you don't need that lame `progn`.
- `=` is fine for comparing numbers, `equalp` is from `cl.el` which is bad for technical reasons.
- I used `(use-region-p)` as a more rigorous test whether the region was active
- I also used `(point-at-bol)` and `(point-at-eol)` instead of just `(point)` because during testing I got some weird results sometimes when point was left halfway through a line I expected to be in the range.

Also I removed that `(message ...)` which I have no idea why it was there.

esiegel:
Awesome, thanks for the thorough code review.  The changes look great.

PR #8370, opened 2017-02-15
Prevent `tmux-command` at GUI mode

gfZeng:


TheBB:

PR #8371, opened 2017-02-15
Fix markdown org table saving

sdwolf:
When creating a table in `markdown-mode` or `gfm-mode` (both using `orgtbl-mode` beind the scenes) the separator between the header and body use a `+` character to indicate column split:

```markdown
| This | Is | A   | Table |
|------+----+-----+-------|
| this | is | the | body  |
```

The markdown layer has a function to convert those `+` signs to `|` in order to produce valid markdown code but it was previously called after saving the buffer content to file, requiring saving once more to persist the change.

This patch ensures the change is done before saving to eliminate the need for a second save.


syl20bnr:
Nice catch :-)
Thank you ! 👍 

PR #8376, opened 2017-02-15
Add details for install instructions in README.md

soorho:
Compilation warnings - `Warning (bytecomp)` - cause no harm. We know this, but others may not. Indeed, many users thought the warnings are damn serious. One of them even spent seven hours trying to fix the warnings, "reinstalling emacs and spacemacs seven times." It took her seven hours, although obvious to us, to realize the warnings are not to be worried about. 

Luckily, we can prevent this from happening again by adding few sentences in `README.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@
 |
 <b><a href="CONTRIBUTING.org">contribute</a></b>
 |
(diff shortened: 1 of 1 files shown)


syl20bnr:
Thank you !

PR #8385, opened 2017-02-16
Add extension ".wl" for wolfram-mode

kenkangxgwe:
I just associated `.wl` with `wolfram-mode` in the `packages.el` of `extra-langs` layer.

### DIFF
--- a/layers/+lang/extra-langs/packages.el
+++ b/layers/+lang/extra-langs/packages.el
@@ -36,8 +36,8 @@
 (defun extra-langs/init-thrift ()
   (use-package thrift :defer t))
 
-;; no associated extension because conflicts with more common Objective-C, manually invoke for .m files.
 (defun extra-langs/init-wolfram-mode ()
   (use-package wolfram-mode
     :defer t
-    :interpreter "\\(Wolfram\\|Mathematica\\)Script\\( -script\\)?"))
+    :interpreter "\\(Wolfram\\|Mathematica\\)Script\\( -script\\)?"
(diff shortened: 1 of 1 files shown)


layers/+lang/extra-langs/packages.el
```diff
@@ -36,8 +36,8 @@
 (defun extra-langs/init-thrift ()
   (use-package thrift :defer t))
 
-;; no associated extension because conflicts with more common Objective-C, manually invoke for .m files.
```
bmag: modify this comment, but don't delete it entirely. The info about .m files is important

bmag:
Is it normal to associate `.wl` files with `wolfram-mode`? I couldn't find any files that use the `.wl` extension...

kenkangxgwe:
Hi @bmag , please just check the [Wolfram Language on Wikipedia](https://en.wikipedia.org/wiki/Wolfram_Language) and [Wolfram Language Scripts](https://reference.wolfram.com/language/tutorial/WolframLanguageScripts.html). You will see this filename extension. Unlike .m and .nb, .wl is preferred for it is distinguishable from MATLAB and Objective-C file, and more human readable than notebook format.

bmag:
Thank you :+1: Cherry-picked into develop, you can safely delete your branch. I also squashed the two commits into one commit, and broke the long comment line in two in https://github.com/syl20bnr/spacemacs/commit/f853b78148b4985bcc4e6c1232564406ee9c4466. Next time please remember to do so.

Issue #8074, opened 2016-12-29
void function org-link-types


sangjeedondrub:
how to fix the following error which occurs when opening an org file
```lisp
Debugger entered--Lisp error: (void-function org-link-types)
  (org-link-types)
  (regexp-opt (org-link-types))
```

d12frosted:
@sangjeedondrub it's great that you opened this issue without reading the message we posted for new issues. I am going to duplicate it in the end of my comment. Since you have already opened this issue, could you please share the output of `SPD h d s`? Also, have you updated all your packages? 

> If you are posting about a *bug* please close this issue and open a new one
using [SPC h I] or [M-m h I] (I = capital i) from within Spacemacs. If the
bug makes Spacemacs unusable, you can find the bug report template [here](https://github.com/syl20bnr/spacemacs/blob/develop/core/templates/REPORTING.template)
You can delete this message to start typing.

sangjeedondrub:
```
#### System Info :computer:
- OS: darwin
- Emacs: 25.1.1
- Spacemacs: 0.200.5
- Spacemacs branch: develop (rev. f5957eb)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
(ivy better-defaults github ranger colors prodigy search-engine graphviz
     (syntax-checking :variables syntax-checking-enable-by-default nil syntax-checking-enable-tooltips nil)
     (spell-checking :variables spell-checking-enable-by-default nil)
     (vinegar :variables vinegar-reuse-dired-buffer t)
     (spacemacs-layouts :variables layouts-enable-autosave nil layouts-autosave-delay 300)
     (git :variables git-magit-status-fullscreen t magit-push-always-verify nil magit-save-repository-buffers 'dontask magit-revert-buffers 'silent magit-refs-show-commit-count 'all magit-revision-show-gravatars nil)
     (ibuffer :variables ibuffer-group-buffers-by 'projects)
     (auto-completion :variables auto-completion-enable-sort-by-usage t auto-completion-enable-snippets-in-popup t :disabled-for org markdown)
     (osx :variables osx-dictionary-dictionary-choice "Simplified Chinese - English")
     restclient
     (gtags :disabled-for clojure emacs-lisp javascript latex python shell-scripts)
     (shell :variables shell-default-shell 'eshell)
     docker deft markdown org shaders yaml react
     (python :variables python-test-runner
             '(nose pytest))
[...]

PR #8399, opened 2017-02-18
Resolving org version conflicts: removed org-plus-contrib packages from org-layer

RockyRoad29:
I've tried to polish my proposal made on #8074 into a PR.
The modules are still available:

```elisp
(require 'ob)
(require 'org)
(require 'org-agenda)
(require 'org-mime)
```
run fine, but now no package is loaded from emacs built-in org-mode.

Here is the [tiny script : lpkg-explorer.el](https://gist.github.com/RockyRoad29/bd4ca6fdb41196a71662986f809e2b1c) that  I used to check that:

To be accurate, I tested it only (but thoroughly) on the master branch only.
When I thought I was ready for PR, I saw your message about develop branch !!! More work then !
On my poor-mans-testing-workbench, more issues arose on the develop branch , so I leave that part to you guys .

But here it is, ready to merge. About testing it, I kinda believe that it could reveal new issues in the rest of the code, that would have been tested against the shadowing built-in modules. 

Thanks to all contributors for this great piece of software and keep up the good job.

### DIFF
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -20,14 +20,15 @@
     htmlize
     mu4e
     ;; ob, org and org-agenda are installed by `org-plus-contrib'
-    (ob :location built-in)
-    (org :location built-in)
-    (org-agenda :location built-in)
-    (org-expiry :location built-in)
+    ;; which is loaded from core layers/+spacemacs/spacemacs-org/packages.el
+    ;; (ob :location built-in)
+    ;; (org :location built-in)
+    ;; (org-agenda :location built-in)
+    ;; (org-expiry :location built-in)
     (org-journal :toggle org-enable-org-journal-support)
     org-download
     ;; org-mime is installed by `org-plus-contrib'
-    (org-mime :location built-in)
+    ;; (org-mime :location built-in)
     org-pomodoro
     org-present
     (org-projectile :toggle (configuration-layer/package-usedp 'projectile))


RockyRoad29:
UPDATE
As I investigate further and read about org-plus-contrib and more issues with org-babel-evaluate, I realize that I may have missed some points. I've already spent a lot of time on this, read that many users complain on issues are related to this, and with a bit of help from you geeks, my time would not be lost and we might reach a satifactory solution faster.

bmag:
`ob`, `org`, `org-agenda`, etc. must be part of `org-packages` so Spacemacs will call their respective init functions during startup. Using `:location built-in` prevents Spacemacs from trying to install `ob`, `org`, `org-agenda`, etc. as separate Melpa packages, since they don't show up as separate packages in Melpa but are all installed as part of `org-plus-contrib` package.

I appreciate your efforts in this regard. However, commenting out these packages isn't the way to go, since it "turns off" some of our Org configuration. There are at least two possible reasons why old Org code could be loaded:
- during Emacs startup, a (unknown) part of Spacemacs loads (a part of) Org before we modify `load-path` and put `org-plus-contrib`'s path.
- when installing `org-plus-contrib`, parts of the built-in Org get byte compiled into `elc` files in `org-plust-contrib`'s directory. (IIRC there's some info about this in [emacs.stackexchange.com](https://emacs.stackexchange.com) or elsewhere)

Do you plan on investigating any of these possiblities?

RockyRoad29:
Hi bmag,

I greatly appreciate your prompt and accurate answer, and I will do my best
(also beware I'm not a native english speaker - french - so please don't take offence for poor language style).

I can see two investigation directions:
 
1.   Detect early shadowing packages, maybe simply with `#'locate-library `.
It could be done 
    a) first in tests (sorry I know nothing about elisp testing now, apart from a name, 'cask, maybe)
, basically aimed at devs, but also 
    b)  for end users, with suitable output (not frightening, easy to report), until the problem is acertained solved.

2. The second point could be avoided if problems are detected before compilation.
 
"Early" implies good understanding of spacemacs initialization process, which obviously I have not.

Would you have for me some tips for setting up an efficient-not-invasive workbench for emacs dev branch ?

RockyRoad29:
Ouf ! I have finished a polished issue report #8414 . 
I hope it will help

syl20bnr:
Thank for all the effort you put in this investigation 💜 Indeed this PR is not a valid solution and break Spacemacs loading mechanism so I close this PR.

PR #8406, opened 2017-02-19
add dart layer

averrin:
Config layer for dart. I have many problems with packages and keybindings and i commented some lines, which i think correct, but not works.

1. ~~It breaks if i use `recipe` for installing package~~
2. If i use `spacemacs/set-leader-keys-for-major-mode` it works only for first opened file.


Miciah:
Does this PR have a future? I see that https://github.com/averrin/dart-layer has a deprecation notice recommending using lsp-mode instead.

averrin:
Yeah, i think i can close it.

PR #8407, opened 2017-02-19
Don't use --vimgrep with ripgrep

agriffis:
See conversation with @mijoharas at
https://github.com/syl20bnr/spacemacs/issues/7370#issuecomment-280681995
and following conversation with @syohex at
https://github.com/syohex/emacs-helm-ag/issues/293

In a nutshell, emacs-helm-ag recommends --vimgrep to work around a bug
in ag. It doesn't apply to ripgrep.

I also filed a PR against emacs-helm-ag to update their README, see
https://github.com/syohex/emacs-helm-ag/pull/294

### DIFF
--- a/layers/+completion/helm/packages.el
+++ b/layers/+completion/helm/packages.el
@@ -246,7 +246,7 @@ Search for a search tool in the order provided by `dotspacemacs-search-tools'."
       (defun spacemacs/helm-files-do-rg (&optional dir)
         "Search in files with `rg'."
         (interactive)
-        (let ((helm-ag-base-command "rg --smart-case --no-heading --vimgrep"))
+        (let ((helm-ag-base-command "rg --smart-case --no-heading"))
           (helm-do-ag dir)))
 
       (defun spacemacs/helm-files-do-rg-region-or-symbol ()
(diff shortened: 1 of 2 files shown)


bmag:
Conflicts: #8065 

@aaronjensen in #8065 you replaced `--vimgrep` with `--color never`, while here @agriffis just removed `--vimgrep`. Is it necessary to disable colors for ivy-wgrep to work? I'd like to merge one of these PRs soon-ish.

aaronjensen:
I don't know if it's necessary, but it seems better to strip colors so you don't have to worry about parsing them, yeah? All of the others disable colors. It may be that rg is smart enough in most cases to not emit color, but I'd rather be explicit.

aaronjensen:
@bmag I updated my branch to fix helm-ag as well. Feel free to merge whichever you like.

aaronjensen:
Just tested quickly, it does not appear to be necessary to disable color for ivy and wgrep to work. Works on my machine, but I'd still suggest/prefer the explicitness for whatever it's worth.

bmag:
> I'd rather be explicit.

Me too, I just want to make sure removing colors from the command doesn't remove them from the Emacs results buffer as well. I did a quick test and it appears that removing colors from the command doesn't effect the colors in the buffer.

@agriffis I'm gonna merge #8065 instead of this one. Thanks for the additional info.

Issue #8415, opened 2017-02-20
 spacemacs/rename-current-buffer-file produces error


lislon:
#### Description :octocat:

spacemacs/rename-current-buffer-file produces error when called on a non-file buffer

#### Reproduction guide :beetle:
- Start Emacs
- create empty buffer <kbd>SPC b N</kbd>
- rename file <kbd>SPC f R</kbd>

*Observed behaviour:* :eyes: :broken_heart:
Error: 
```
let*: Wrong type argument: stringp, nil
```

*Expected behaviour:* :heart: :smile:
Human readable message:
```
Buffer ’untitled’ is not visiting a file!
```

#### System Info :computer:
- OS: gnu/linux
- Emacs: 25.1.1
- Spacemacs: 0.200.7
- Spacemacs branch: develop (rev. 0c8af9686)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(helm
 (auto-completion :disabled-for org markdown)
 better-defaults emacs-lisp common-lisp yaml search-engine php html lua c-c   shell-scripts git github playground colors
 (java :packages
       (not eclim))
 javascript python gnus sql php docker nginx dash
 (\,@
  (unless
      (eq system-type 'windows-nt)
    '(fasd spell-checking)))
 fasd spell-checking mu4e erc emoji
 (ranger :variables ranger-override-dired t)
 (markdown :variables markdown-live-preview-engine 'vmd)
 org my-org my-keys
 (shell :variables shell-default-height 30 shell-default-position 'bottom)
 version-control latex)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GCONF GSETTINGS NOTIFY ACL GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11

#### Backtrace :paw_prints:
```
<<BACKTRACE IF RELEVANT>>
```

PR #8416, opened 2017-02-20
spacemacs/rename-current-buffer-file fix on non-file buffer

lislon:
fix issue #8415

Moved `(file-name-directory filename)` under error-checking condition


bmag:

PR #8431, opened 2017-02-23
Add test for org-html export of documentation

smile13241324:
To avoid further issues with the publishing of documentations I have written a test to ensure that
all documentation files can be exported to html.

To avoid issues with loading the right org version after package installation I have also slightly
modified the test workflow to include a dedicated package installation target done prior to executing
any tests.

### DIFF
--- a/.travis-build.sh
+++ b/.travis-build.sh
@@ -11,7 +11,7 @@
 ##
 ## License: GPLv3
 
-tests=("core")
+tests=("core" "layers/+org/html-export")
 
 if [ $USER != "travis" ]; then
     echo "This script is not designed to run locally."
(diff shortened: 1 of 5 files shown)


smile13241324:
Seems like there is still an issue with the org installation. The tests fail with `(void-function org-link-types)` on the travis workers. I can also reproduce the issue when I run the .travis-build.sh locally however this only happens when org is freshly installed, when I do not delete my elpa directory and only let the test run we have the correct behaviour. So this looks like a separate bug/issue only happening right after installation.

My second question is how can I introduce utility functions for the test execution? In my case I needed a replacement for emacs 25.1.1's `directory-files-recursively`. I have first tried to add it to the `user-config` function in the `dotspacemacs.el` however this seems to not have been loaded. So I have ended up with adding this function to the beginning of my test file. But I am sure that this is not the right way to go.

Could someone lend me a hand with this?

smile13241324:
Found a workaround for the org-link-set-parameters issue. Changed the test workflow a bit to include a dedicated installation step before any test execution is done. This ensures that between installing packages and executing tests emacs is at least restarted once which ensures that the correct org mode version is loaded.

I have also succeeded in verifying all external links in that test.
For now local links are not verified, if it is needed it needs to be added later.

syl20bnr:
Thank you ! 💜 
I moved the tests to the `/doc` directory in commit https://github.com/syl20bnr/spacemacs/commit/62167f70

PR #8433, opened 2017-02-24
ipython notebook layer app command changes

bmabey:
This adds a new command to login to a notebook which is often required on newer installs of Jupyter these days. 

I also changed the command to open the notebook list from `n` to `o`. Better mnemonics was the motivation behind this change. I'm not sure if `n` was chosen to match another spacemacs convention but I couldn't find any reference to such a convention. I realize that this is change could annoy people use to the previous binding so I can remove this commit from the PR.

### DIFF
--- a/layers/+lang/ipython-notebook/packages.el
+++ b/layers/+lang/ipython-notebook/packages.el
@@ -14,10 +14,12 @@
 (defun ipython-notebook/init-ein ()
   (use-package ein
     :defer t
-    :commands ein:notebooklist-open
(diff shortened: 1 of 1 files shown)


bmabey:
I just rebased this PR so it should be ready to merge in... Is there anything else I can do to help speed this along?

bmabey:
I have some other improvements I'd like to make to this layer but am holding off until this gets merged in.

TheBB:

PR #8446, opened 2017-02-27
bibtex: add key bindings

et2010:
IMHO, the `org-ref-helm-insert-ref` and `org-ref-helm-insert-label` are both important commands of org-ref. This commit adds keybindings for them.


TheBB:
It looks like these bindings use helm directly, rather than going through something like `completing-read`. Therefore they need to be predicated on whether helm is actually being used, no?

PR #8449, opened 2017-02-28
revert bug introduced by commit eef1a87

LukeXuan:
Hi, caused by the commit https://github.com/syl20bnr/spacemacs/commit/eef1a87e98d8258c2a295566d954bb7b391942fb. A bug is introduced.

In the current develop branch the `spacemacs/set-evil-search-module` is called before the `hybrid-mode` is loaded, when the `hybrid-mode-use-evil-search-module` is unbounded, and thus always choose `'evil-search` as the search module. This pull request fixes this bug by reverting the change on that single file.

### DIFF
--- a/layers/+distributions/spacemacs-bootstrap/packages.el
+++ b/layers/+distributions/spacemacs-bootstrap/packages.el
@@ -41,10 +41,10 @@
     :override-mode-name spacemacs-leader-override-mode))
 
 (defun spacemacs-bootstrap/init-evil ()
-  ;; ensure that the search module is set at startup
-  ;; must be called before evil is required to really take effect.
-  (spacemacs/set-evil-search-module dotspacemacs-editing-style)
-  (add-hook 'spacemacs-editing-style-hook 'spacemacs/set-evil-search-module)
+  (with-eval-after-load 'hybrid-mode
+    ;; ensure that the search module is set at startup
+    (spacemacs/set-evil-search-module dotspacemacs-editing-style)
+    (add-hook 'spacemacs-editing-style-hook 'spacemacs/set-evil-search-module))
 
   ;; evil-mode is mandatory for Spacemacs to work properly
   ;; evil must be require explicitly, the autoload seems to not


bmag:
I'm using hybrid style I'm not experiencing any bug. If I don't specify `hybrid-mode-use-evil-search-module` I get `isearch` as the search module. If I specify it using `dotspacemacs-editing-style` I get `evil-search`:
```elisp
   dotspacemacs-editing-style '(hybrid :variables
                                       hybrid-mode-default-state 'normal
                                       hybrid-mode-enable-evilified-state t
                                       hybrid-mode-enable-hjkl-bindings nil
                                       hybrid-mode-use-evil-search-module t)
```

LukeXuan:
Hi, what I've got in the variable `evil-search-module` is indeed `isearch`. However, what I got by key-bindings for `/` is `evil-ex-search-forward`, this shouldn't be the case. It should be `isearch-forward` instead.

LukeXuan:
It seems that when the first time invoke `(spacemacs/set-evil-search-module dotspacemacs-editing-style)` in the `init-evil` the `hybrid-mode.el` is not loaded therefore `hybrid-mode-use-evil-search-module` is `void`, which leads to setting `'evil-search` as the search module.

I'm not quite familiar with elisp but I think there should be some problem in these codes.

LukeXuan:
Hi, @syl20bnr @bmag any update? Spacemacs these days is annoying me since I should manually patch this bug every time I update it.

bmag:
From what I see, `evil-search-forward` is the expected command for `isearch` module (when `hybrid-use-evil-search-module` is bound and equals `nil`), and `evil-ex-search-forward` is the expected command for `evil-search` module. There are only certain modes where `/` is bound to `isearch` independently of the `evil-search-module`'s value.

With the current code (without reverting eef1a87), I get:
- if `hybrid-mode-use-evil-search-module` is set to `nil` in `dotspacemacs-editing-style`: `evil-search-module` is `isearch` and `/` is bound to `evil-search-forward`, as expected.
- if `hybrid-mode-use-evil-search-module` is set to `t` in `dotspacemacs-editing-style`: `evil-search-module` is `evil-search` and `/` is bound to `evil-ex-search-forward`, as expected.
- if `hybrid-mode-use-evil-search-module` isn't set in `dotspacemacs-editing-style`: `evil-search-module` is `isearch` and `/` is bound to `evil-ex-search-forward`, NOT as expected.

After applying this PR, I get:
- if `hybrid-mode-use-evil-search-module` is set to `nil` in `dotspacemacs-editing-style`: `evil-search-module` is `isearch` and `/` is bound to `evil-search-forward`, as expected.
- if `hybrid-mode-use-evil-search-module` is set to `t` in `dotspacemacs-editing-style`: `evil-search-module` is `evil-search` and `/` is bound to `evil-search-forward`, NOT as expected. It should be `evil-search` and `evil-ex-search-forward`.
- if `hybrid-mode-use-evil-search-module` isn't set in `dotspacemacs-editing-style`: `evil-search-module` is `isearch` and `/` is bound to `evil-search-forward`, NOT as expected. the same as setting `hybrid-mode-use-evil-search-module` to `t` (`evil-search` and `evil-ex-search-forward`).

Bottom line, this PR doesn't solve the actual bug, so I'm not going to merge it. However, I believe I found the real source of the bug - `spacemacs/set-evil-search-module`. It only sets `evil-search-module`, while it should call `evil-select-search-module`. PR coming shortly.

bmag:
Correction: when `hybrid-mode-use-evil-search-module` isn't set in `dotspacemacs-editing-style`, then the default should be the same as setting it to `nil`.

bmag:
@LukeXuan can you confirm that #8501 fixes the issue?

PR #8461, opened 2017-03-02
Add enable/disable transparency functions

adamcavendish:
Add spacemacs/enable-transparency and spacemacs/disable-transparency.

For anyone who'd like to enable transparency at startup, just write '(spacemacs/enable-transparency)' to .spacemacs, and it'll work fine.

### DIFF
--- a/layers/+distributions/spacemacs-base/keybindings.el
+++ b/layers/+distributions/spacemacs-base/keybindings.el
@@ -671,7 +671,8 @@ If FRAME is nil, it defaults to the selected frame."
   "Increase transparency for FRAME.
 If FRAME is nil, it defaults to the selected frame."
   (interactive)
-  (let* ((current-alpha (car (frame-parameter frame 'alpha)))
+  (let* ((frame-alpha-pair (frame-parameter frame 'alpha))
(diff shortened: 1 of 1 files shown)


bmag:
These two functions look good. Are you willing to expand the PR a bit to achieve the following changes?
- `spacemacs/toggle-transparency` should use the new functions
- `spacemacs/increase-transparency` and `spacemacs/decrease-transparency` currently fail is `(frame-parameter nil 'alpha)` returns `nil`. They should treat the `nil` case the same as if current alpha is 100 (i.e. frame isn't transparent).

adamcavendish:
`spacemacs/increase-transparency` and `spacemacs/decrease-transparency` are fixed. I'd like to remain `spacemacs/toggle-transparency` as it is, cause it's now more efficient.

bmag:

I've made some changes to docstrings and code in https://github.com/syl20bnr/spacemacs/commit/165b4b425f45c8184553b47176cc02788bafad68. I've made `spacemacs/toggle-transparency` rely on th other functions in a way that adds minimal overhead, and anyway it is not a performance critical part.

PR #8465, opened 2017-03-05
Remove CUSTOM ID from documentation to fix #8216

### LINKED ISSUE #8216: CUSTOM_ID tags showing in develop.spacemacs.org
Reporting here since the org files used for generation are in this project.

every CUSTOM_ID tags from DOCUMENTATION.org (haven't checked the others) is showing on develop.spacemacs.org (for instance [here](http://develop.spacemacs.org/doc/DOCUMENTATION.html#replacing-text-with-iedit)) like on this screenshot:

image

smile13241324:
In the currently generated web pages for the documentation files we show internal custom id tags. This is caused by `spacemacs//org-heading-annotate-custom-id` which will be adding a duplicate `:PROPERTIES:` tag if one is already existing. 

To solve this I have simply removed the superfluous CUSTOM IDs from the effected documents. As extending the function to reliably detect those `PROPERTIES` drawers I found not really worth the trouble as I cannot find any reason to add those drawers except CUSTOM IDs.

### DIFF
--- a/CONTRIBUTING.org
+++ b/CONTRIBUTING.org
@@ -25,9 +25,6 @@ Thanks! :heart: :heart: :heart:
  - [[#credits][Credits]]
 
 * Asking for help
-  :PROPERTIES:
-:CUSTOM_ID: asking-for-help
-:END:
 If you want to ask an usage question, be sure to look first into some places as
 it may hold the answer:
 
@@ -40,9 +37,6 @@ discuss it with us :relaxed:. We will direct you to a solution, or ask you to
 open an issue if it is needed.
 
 * Reporting issues
-:PROPERTIES:
-:CUSTOM_ID: reporting-issues
-:END:
 Issues have to be reported on our [[https://github.com/syl20bnr/spacemacs/issues][issues tracker]]. Please:
 
 - Check that the issue has not already been reported.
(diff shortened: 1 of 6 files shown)


syl20bnr:
Thank you ! 👍 

nikolaiam:
I know I'm late on this as it was already merged, but doesn't this reintroduce broken links on spacemacs.org? The `CUSTOM_ID` tags were originally introduced to fix broken links when linking from one file of the documentation to another, because the export process could not parse the links correctly without the tags. Thus, if the export hasn't been altered in the meantime, I believe that by merging this PR the links (such as on line 122 of the FAQ) are broken again, with no obvious way to fix this. I suggest we reconsider the merging of this PR.

As a way of getting hyperlinks to work and hiding the `CUSTOM_ID` tags at the same time I proposed in #8146 using the default `org-export-dispatch` process to export the doc files. Another way would be to extend the current export function to detect the `PROPERTIES` drawers as proposed by @smile13241324. What do you think?

JAremko:
@nikolaiam  `spacemacs//org-heading-annotate-custom-id` adds `CUSTOM_ID`

JAremko:
But I think it worth a while to add `CUSTOM_ID` detection to the `spacemacs//org-heading-annotate-custom-id` function so it will not add extra nor override original  `CUSTOM_ID` - can be helpful for solving corner cases like same `CUSTOM_ID` in a document.

nikolaiam:
Exactly!

PR #8471, opened 2017-03-06
Fix broken links to missing README files in layers list to fix #8291

### LINKED ISSUE #8219: [latex layer] About use xelatex instead of pdflatex to build tex file
Hi, spacemacs latex layer use pdflatex to build tex file by default, but I use xelatex to do the same work(I am Chinese), I can set the variable `TeX-engine` provided by auctex to "xetex" to implement this, weather latex layer could provide a layer variable to custom this?

### LINKED ISSUE #8291: All three links on Distributions broken in LAYERS docs
http://spacemacs.org/layers/LAYERS.html#distributions

smile13241324:
Some of the layers do not supply a README.org file which caused invalid links in the layer list. To avoid this I have changed spacemacs//generate-layers-from-path to only add links to layers with a valid README.org file.

I have also removed an invalid check to exclude the non existing directory "distribution" from the layers list. I think that originally the folder "distributions" should have been excluded but this is not longer a feasible action as there is at least one layer with a valid README.org file in there today. So now we add links to all layers providing a README.org file independent of their category.

### DIFF
--- a/core/core-documentation.el
+++ b/core/core-documentation.el
@@ -35,14 +35,16 @@
          (categories (-filter
                       (lambda (p)
                         (eq 'category (configuration-layer//directory-type p)))
-                              all-subs)))
+                      all-subs)))
     (message "%S" layers)
     (dolist (l layers)
       (let ((layer-name (file-name-nondirectory l))
-            (target-path (concat (file-relative-name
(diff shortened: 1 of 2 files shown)


syl20bnr:
Thank you !

PR #8473, opened 2017-03-08
Remapping of twittering-mode was broken in keyboard-layout. Fix it.

CharlesHD:
I've just noticed that twittering-mode support for the keyboard-layout layer was broken because the syntax was out of date. This PR fix it.


syl20bnr:
Thank you ! 👍 

PR #8475, opened 2017-03-08
rg: explicitly request line numbers

bmag:
rg tries to be smart about printing line numbers in its output (if running inside a TTY or not). On Windows, sometimes (always?) it chooses not to print line number. However, we always want line numbers, so we need to pass -n flag.

fixes https://github.com/syl20bnr/spacemacs/pull/8065#issuecomment-285016670

### DIFF
--- a/layers/+completion/ivy/config.el
+++ b/layers/+completion/ivy/config.el
@@ -13,12 +13,15 @@
 ;; Variables
 
 (defvar spacemacs--counsel-commands
-  '(("rg" . "rg --smart-case --no-heading --color never %s %S .")
+  '(;; --line-number forces line numbers (disabled by default on windows)
+    ;; no --vimgrep because it adds column numbers that wgrep can't handle
+    ;; (see https://github.com/syl20bnr/spacemacs/pull/8065)
+    ("rg" . "rg --smart-case --no-heading --color never --line-number %s %S .")
     ("ag" . "ag --nocolor --nogroup %s %S .")
     ("pt" . "pt -e --nocolor --nogroup %s %S .")
     ("ack" . "ack --nocolor --nogroup %s %S .")
     ("grep" . "grep -nrP %s %S ."))
-  "Alist of search commands and their corresponding commands
+  "An alist of search commands and their corresponding commands
 with options to run in the shell.")
 
 (defvar spacemacs--counsel-search-max-path-length 30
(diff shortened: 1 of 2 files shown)


layers/+completion/ivy/config.el
```diff
@@ -13,7 +13,10 @@
 ;; Variables
 
 (defvar spacemacs--counsel-commands
-  '(("rg" . "rg --smart-case --no-heading --color never %s %S .")
+  '(;; -n forces line numbers (disabled by default on windows)
+    ;; no --vimgrep because it adds column numbers that wgrep can't handle
+    ;; (see https://github.com/syl20bnr/spacemacs/pull/8065)
+    ("rg" . "rg --smart-case --no-heading --color never -n %s %S .")
```
aaronjensen: I prefer long-form flags when not typing them at a terminal myself: `--line-number`

layers/+completion/ivy/config.el
```diff
@@ -13,7 +13,10 @@
...
     ("ag" . "ag --nocolor --nogroup %s %S .")
     ("pt" . "pt -e --nocolor --nogroup %s %S .")
     ("ack" . "ack --nocolor --nogroup %s %S .")
```
aaronjensen: ⬇️  There's a typo down here "Alist" if you want to fix that up while you're in here

bmag:
Option 1: add PRs path to your git remote, then you have local access to the PR branch. For example:
```ini
# file: .git/config
[remote "upstream"]
	url = https://github.com/syl20bnr/spacemacs.git
	fetch = +refs/heads/*:refs/remotes/upstream/*
	fetch = +refs/pull/*/head:refs/remotes/upstream/pr/*
```
The last line allows me to checkout PR branches as any git branch. The remote name might be different in your config (it's probably called `origin` and not `upstream`), so change accordingly.

Option 2: download the [raw diff](https://patch-diff.githubusercontent.com/raw/syl20bnr/spacemacs/pull/8475.diff) and apply it with command line (or other) tools.

Option 3: just make the change manually. If you're not using git, remember to save a backup so you can revert the changes later.

layers/+completion/ivy/config.el
bmag: ok, will change that

layers/+completion/ivy/config.el
bmag: Not a typo, actually. It really is an [alist](https://www.gnu.org/software/emacs/manual/html_node/elisp/Association-Lists.html). Should be "An alist" instead of "Alist", though.

aaronjensen: ah, hah. Makes sense, thanks.

Swoorup:
Cool cheers,
I can confirm, this works for me now.

bmag:
I changed `-n` to `--line-number` and fixed the grammar mistake, then merged.

PR #8476, opened 2017-03-08
added dollar to match only function name

nl5887:
the check.f and run functions use regular expressions, without the dollar sign it will run not only function _2, but also _21, etc.


syl20bnr:
Thank you ! 👍 

PR #8486, opened 2017-03-09
javascript: leverage js-doc Yasnippet integration if available

madand:
js-doc provides an alternative function for generating function doc comments
that makes use of Yasnippet to easily jump between the comment sections/tags.

This commit enables use of that function, if Yasnippet package is active.


TheBB:

PR #8490, opened 2017-03-10
search compile_commands.json in build subdir

amosbird:


TheBB:
Any other typical subdirectories we might want to look in? You could just add an optional argument to `FindNearest` that is a list of potential subdirectories (default value `[]`) and search through them one by one. I also tend to think that the root dir `.` should have the highest priority.

TheBB:
I'm proposing this https://github.com/TheBB/spacemacs/commit/d50b12d8863e0cf2f8b276c649d479917f69eb71

Can you check if it works as expected?

TheBB:
Cool, I'll merge it later today

amosbird:
@TheBB hi, we need to change the tail recursion part to `return FindNearest(parent, target, filename, subdirs)` .

PR #8501, opened 2017-03-11
Set evil-search-module correctly

bmag:
- Just setting `evil-search-module` is not enough. If Evil is loaded already, we must call `evil-select-search-module` as well to actually change the key-bindings to use the correct search module.
- If the user didn't set `hybrid-mode-use-evil-search-module` so it's not bound, we treat it as if it was set to `nil`, to match the documentation and the default value of `hybrid-mode-use-evil-search-module`.

Should fix the issue raised by @LukeXuan in #8449


LukeXuan:
Yes, I confirm the bug is fixed.

bmag:

PR #8503, opened 2017-03-12
Modify default eshell pcomplete interface to helm.

aijony:
Override the keybinding for tab in eshell from pcomplete to helm-esh-pcomplete.
This is a (subjectively) nicer interface.
The results should remain unchanged.

There are [still](https://github.com/syl20bnr/spacemacs/issues/8504) two auto-completion when eshell is running. However, instead of pcomplete and company it is now helm-esh-pcomlete and company. 

Thank you for viewing my pull request, please give feedback if you'de like. 😄


aijony:
Sorry, I modified the wrong, file

PR #8508, opened 2017-03-13
Fix and update eshell's clear features

### LINKED ISSUE #5419: eshell clear leaves a line at top
Entering clear leaves a line on top of eshell.

Reproducible by:
fresh install
enable shell layer
:eshell
type clear RET

image

When you press `RET`,

image

On the other hand, C-L, the equivalent command, works as intended.

System: OS X 10.11.3
Emacs: 24.5.1

### LINKED ISSUE #5424: eshell clear/ctrl-L in emacs 25 creates unwanted page of whitespace
#### Description

Both clear commands in emacs 25 are very strange; clear RET creates a page of whitespace then prompts you. Ctrl-L creates half a page of whitespace, then cycles through three different positions of prompts.
#### Reproduction guide

a) clear
- Start Emacs
- Start eshell
- type "clear" then RET

b) Ctrl-L
- Start Emacs
- Start eshell
- Produce some command output
- Ctrl-L

_Observed behaviour:_
a) clear

image

This is fresh load up of eshell in Emacs 25.

image

This is what happens after `clear RET`

b) Ctrl-L

image

Some command history.

image

image

image

Pressing Ctrl-L cycles between these three images.

_Expected behaviour:_

image

Both keystrokes should land in a cleared page.
#### System Info
- OS: darwin
- Emacs: 25.1.50.1
- Spacemacs: 0.105.13
- Spacemacs branch: master (rev. 2c3e6b4)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:

``` elisp
(emacs-lisp)
```

aijony:
Modify the default behavior for eshell's clear functionality.

 - Make eshell `clear RET` _similar_ to `C-l`
   - `clear RET` is eshell/clear
   - `C-l` is eshell-clear-stroke
 - Prevent `C-l` clearing cycle
 - Prevent `clear RET` inserting a page of white-space
 - Prevent duplicate insert lines from `clear RET`
 - Make eshell-clear-keystroke that is dependent on eshell/clear
   - Load both after eshell is initialized to prevent them from being overwritten

This fixes #5424, fixes #5419

Thank you for looking at my code.
Constructive critisicm is deeply appreciated. 
I'm sorry if I missed something.
I'm only a beginner. 😅

### DIFF
--- a/layers/+tools/shell/funcs.el
+++ b/layers/+tools/shell/funcs.el
@@ -125,15 +125,21 @@ is achieved by adding the relevant text properties."
   (when (configuration-layer/package-usedp 'semantic)
     (semantic-mode -1))
   ;; Caution! this will erase buffer's content at C-l
-  (define-key eshell-mode-map (kbd "C-l") 'eshell/clear)
+  (define-key eshell-mode-map (kbd "C-l") (funcall 'eshell-clear-keystroke))
   (define-key eshell-mode-map (kbd "C-d") 'eshell-delchar-or-maybe-eof))
 
-;; This is an eshell alias
-(defun eshell/clear ()
-  (interactive)
-  (let ((inhibit-read-only t))
-    (erase-buffer))
-  (eshell-send-input))
+(eval-after-load 'eshell
+  '(progn
+     ;; This is an eshell alias
+     (defun eshell/clear ()
+       (let ((inhibit-read-only t))
+         (erase-buffer)))
+     ;; This is a key-command
+     (defun eshell-clear-keystroke ()
+       (lambda ()
+         (interactive)
+         (eshell/clear)
+         (eshell-send-input)))))
 
 (defun spacemacs/helm-eshell-history ()
   "Correctly revert to insert state after selection."
--- a/layers/+tools/shell/funcs.el
+++ b/layers/+tools/shell/funcs.el
@@ -129,12 +129,13 @@ is achieved by adding the relevant text properties."
     (let ((inhibit-read-only t))
       (erase-buffer)))
   ;; This is a key-command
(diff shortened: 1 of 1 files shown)


bmag:
I don't really have time to review this properly, but just a few pointers:
- the call to `define-key` is wrong (I doubt it does what you think it does - currently it binds `C-l` to the return value of `(eshell-clear-keystroke)`.
- `eshell-clear-keystroke` doesn't match the naming conventions (`eshell/something` causes `something` to be a valid command in an Eshell buffer IIRC, not sure about `ehsell-something`) - it's not part of `eshell.el`
- `eval-after-load` is probably unnecessary here, there has to be a way to avoid it (on a side note,  `with-eval-after-load` is more convenient than `eval-after-load`)

aijony:
Thank you for the tips. I definately feel like I've been hacking my way through emacs, but some input was really helpful in leading me towards a bit more guided approach.

I think I understand the define-key a bit more. I don't know what led me to using lambda, but after I removed `eval-after-load` things became a lot easier. My previous attempt that I tried _way_ before using lambda suddenly worked. Killed two birds with one stone, I guess.

For the naming, I didn't realize that `eshell-*` meant it came from eshell.el file. I did however relize `eshell/*` makes it accessable via eshell, and that's why I didn't name it such. I wanted it to be a keycommand, not a shell command. Would using `spacemacs/eshell-clear-keystroke` work?

Thanks again for the feedback, it is very much is appreciated.

bmag:
> Would using `spacemacs/eshell-clear-keystroke` work?

I think it will.

syl20bnr:
Thank you ! 👍 

PR #8509, opened 2017-03-13
fix url jump error

TinySong:
fix url jump error in comments

### DIFF
--- a/layers/+completion/helm/packages.el
+++ b/layers/+completion/helm/packages.el
@@ -248,7 +248,7 @@ Search for a search tool in the order provided by `dotspacemacs-search-tools'."
         (interactive)
         ;; --line-number forces line numbers (disabled by default on windows)
         ;; no --vimgrep because it adds column numbers that wgrep can't handle
-        ;; (see https://github.com/syl20bnr/spacemacs/pull/8065)
+        ;; (see https://github.com/syl20bnr/spacemacs/pull/8065 )
         (let ((helm-ag-base-command "rg --smart-case --no-heading --color never --line-number"))
           (helm-do-ag dir)))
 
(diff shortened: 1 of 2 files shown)


bmag:
Not sure what the error is. Please explain the issue (or link to an existing issue).

sdwolf:
The closing paranthesis is considered part of the url. I confirm this bug and 👍 the fix.

bmag:
> ... is considered part ...

Considered by what? When using `SPC x o` it tries to open a URL that contains the closing parenthesis, or is it something else?

duianto:
Confirmed. With the cursor on the url, if one presses `SPC x o` then the default browser opens a github 404 page, and the address bar includes the closing parenthesis.

sdwolf:
The following link:
```emacs-lisp
;; (see https://github.com/syl20bnr/spacemacs/pull/8065)
```
Opens `https://github.com/syl20bnr/spacemacs/pull/8065)` in a browser, which is 404.

The fixed link:
```emacs-lisp
;; (see https://github.com/syl20bnr/spacemacs/pull/8065 )
```
Opens the expected `https://github.com/syl20bnr/spacemacs/pull/8065` in a browser.

bmag:
Possibly an upstream bug in [link-hint](https://github.com/noctuid/link-hint.el) (unless it's an intended behavior or a known issue). Can someone ask them please?

duianto:
It has been reported upstream: https://github.com/noctuid/link-hint.el/issues/15

noctuid:
Just to clarify, this is a bug in emacs' url libraries (see my post at the above link). Org mode doesn't have this problem though. I can look into alternatives, but I'm not personally familiar with rules related to parens in urls and am not the right person to solve this.

syl20bnr:
Thank you ! 👍 
I removed the parenthesis in an additional commit.

PR #8513, opened 2017-03-13
Remove "ph" 'helm-projectile from spacemacs-base

duianto:
Problem: In the Ivy layer, `SPC p` lists: h -> helm-projectile
but when it's called, then it shows:
`command-execute: Wrong type argument: commandp, helm-projectile`

Solution: Remove "ph" 'helm-projectile from: spacemacs-base/package.el
Because it's also defined in the Helm layer.


syl20bnr:
Thank you !

PR #8514, opened 2017-03-13
Combine 2 projectile spacemacs/set-leader-keys sections

duianto:
Problem: There are two projectile spacemacs/set-leader-keys sections,
next to each other.

Solution: Combine and sort the keys alphabetically with lower case
before upper case. A quick search showed that order in a lot of
other set-leader-key sections.


robbyoconnor:
@duianto  There are conflicts -- resolve them -- otherwise this looks good.

TheBB:

PR #8515, opened 2017-03-13
Remove "pr" 'projectile-recentf from ivy layer

duianto:
problem: "pr" 'projectile-recentf is defined in both the Ivy layer
and spacemacs-base/packages.el

solution: remove it from the Ivy layer.


syl20bnr:
Thank you !

PR #8516, opened 2017-03-13
Use `terraform fmt` by default in the terraform layer

Samangan:
Hey.

I wanted to allow terraform layer users to automatically `terraform fmt` when they saved. I updated the layer documentation to show examples of turning it off if they do not wish to use it.

The reason I decided to turn it on by default is because `terraform fmt` is both encouraged and modeled after `go fmt` which is automatically done before a save in the go layer. Please let me know if you disagree.


syl20bnr:
Thank you ! 👍 
I renamed the layer variable to `terraform-auto-format-on-save` and set the default value to `nil` as it feels more like an opt-in variable to me.

PR #8522, opened 2017-03-14
Rebind go-coverage to gC

### LINKED ISSUE #7306: go and ctags key binding conflict
go layer uses `SPC m g c` for `go-coverage` and gtags layer `SPC m g c`  for `helm-gtags-create-tags`  

Can we use `SPC m g C` for  `helm-gtags-create-tags` ? This is not a function that you would call often and it will help to avoid similar conflicts.

JAremko:
Fixes #7306


syl20bnr:
Thank you ! 👍 
I reverted to your initial proposal, see comment https://github.com/syl20bnr/spacemacs/issues/7306#issuecomment-290595482

PR #8523, opened 2017-03-15
Update which-key popup: key/command grouping

duianto:
Comment the code and group the keys/commands based on the
which-key section they appear in.

Reduce the width of the listed keys:
By removing spaces around the two dots `..` in the key sequences, and
by removing the modifier keys before the last number in a sequence.

Update the `renamed select-window-[1-9]` command name, to the new
`winum-select-window-[1-9]` name.

Combine keys that call the same command.

### DIFF
--- a/layers/+distributions/spacemacs-bootstrap/packages.el
+++ b/layers/+distributions/spacemacs-bootstrap/packages.el
@@ -342,13 +342,97 @@
       (push (cons (cons nil (concat "\\`" (car nd) "\\'")) (cons nil (cdr nd)))
             which-key-replacement-alist)))
 
-  (push '(("\\(.*\\) 0" . "select-window-0") . ("\\1 0..9" . "window 0..9"))
+  ;; Group together sequence and identical key entries in the which-key popup
+  ;; SPC h k- Top-level bindings
+  ;; Remove spaces around the two dots ".."
+  (push '(("\\(.*\\)1 .. 9" . "digit-argument") .
+          ("\\11..9" . "digit-argument"))
         which-key-replacement-alist)
-  (push '((nil . "select-window-[1-9]") . t) which-key-replacement-alist)
 
-  (push '(("\\(.*\\) 1" . "buffer-to-window-1") . ("\\1 1..9" . "buffer to window 1..9"))
+  ;; And remove the modifier key(s) before the last nr in the sequence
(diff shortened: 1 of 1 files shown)


layers/+distributions/spacemacs-bootstrap/packages.el
```diff
@@ -342,14 +342,40 @@
       (push (cons (cons nil (concat "\\`" (car nd) "\\'")) (cons nil (cdr nd)))
             which-key-replacement-alist)))
 
-  (push '(("\\(.*\\) 0" . "select-window-0") . ("\\1 0..9" . "window 0..9"))
+  ;; Group together sequence and identical key entries in the which-key popup.
+  ;; SPC- Spacemacs root
+  ;; rename "0 -> winum-select-window-0-or-10"
+  ;; to  "0..9 -> winum-select-window-0..9-or-10"
+  (push '(("\\(.*\\) 0" . "winum-select-window-0-or-10") .
```
duianto: Would it be better to move the `.` from the end of the line to the beginning of the next line?
There are 3 other occurrences below.

duianto:
I found more instances where the keys are listed with spaces around the two dots: `..` in the Top-level bindings `SPC h k`.

`1 ..9 -> digit-argument`,
`0 .. C-5 -> digit-argument`,
`7 .. C-9 -> digit-argument`,
and
`0 .. C-M-9 -> digit-argument`.

I'll try to remove the spaces around those keys as well.

duianto:
I'm able to remove the spaces around the two dots. But the key sequences became a little bit longer when the leading `C-` and `C-M-` appeared. 

They could be shortened slightly if they were change from this:
`C-0..C-5`, `C-7..C-9` and `C-M-0..C-M-9`
to this:
`C-0..5`, `C-7..9` and `C-M-0..9`

Is it easily understood what's meant or should `C-` and `C-M-` also be written before the last key in the sequence?

layers/+distributions/spacemacs-bootstrap/packages.el
duianto: I kept the `.` at the end, and added some additional improvements to this PR.

deb0ch:
Just tested, it works for me :thumbsup: :sparkles:

It looks very neat, in the code as much as in `which-key` :smiley_cat:

TheBB:

PR #8526, opened 2017-03-15
[enhancement] OSX Layer. add config variables for different OSX modif…

Chream:
…ier keys.

Added variables to set all different modifier keys. See the README for details.
The commit is also backwards compatible with the previous config and the variable
`osx-use-option-as-meta` will take precedence if set to nil or non-nil value (except if
set to 'deprecated which is the new default).

Although the defaults should be unchanged I am using an Norwegian keyboard so if
anyone disagrees with my defaults please let me know and I will fix that.

### DIFF
--- a/layers/+os/osx/README.org
+++ b/layers/+os/osx/README.org
@@ -27,27 +27,51 @@ encourage you to do so :)
 
 * Install
 ** Layer
+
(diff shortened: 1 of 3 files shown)


syl20bnr:
Thank you ! 👍 
Not that this PR as been modified by #8689 

Issue #6683, opened 2016-07-26
What's the deal with "useful" and "useless" buffers?


bmag:
Spacemacs has a concept of "useful" and "useless" buffers. The rules for usefuleness are defined and implemented by variables `spacemacs-useful-buffers-regexp`, `spacemacs-useless-buffers-regexp` and functions `spacemacs/useful-buffer-p`, `spacemacs/useless-buffer-p`. I'm trying to understand what is the rule-of-thumb to decide whether a buffer is useful.

I thought the rule-of-thumb is "if it's normal for the user to want to see this buffer, then it's useful". According to this logic, I plugged `spacemacs/useful-buffer-p` into `buffer-predicate` in commit fe60d0fc1e265756585836bd9f476e44fc7526d4, so functions like `next-buffer`, `other-buffer`, etc. skip useless buffers. The commit caused two unexpected issues (#6678, #6681), so the question is: am I wrong, and `spacemacs/useful-buffer-p` shouldn't be plugged into `buffer-predicate`? or am I right and it's the current imlpementation just doesn't implement the rules we want?

Usefulness rules, according to current implementation:
- comint-derived buffers are useful (read: all REPLs and shells, except those
  that don't derive from `comint-mode`)
- buffers whose name start with `*cider-repl`, `*ansi-term`, `*eshell`,
  `*terminal` and end with `*` (REPLs/shells that might not derive from `comint-mode`)
- the scratch buffer
- buffers whose name doesn't contain `*`

In short, currently useful buffers are files, directories, REPLs, shells and the scratch buffer. This excludes also buffers that the user interacts with, like notmuch buffers (see #6681), `*IBuffer*` and magit buffers.

For context, in the previous situation, useful/useless buffer distinction was only used by <kbd>SPC b n</kbd> and <kbd>SPC b p</kbd> which where bound to are own implementations of `next-buffer` and `previous-buffer`. My commit was discussed in #6574 and fixes #6159.

a13ph:
@bmag FYI this distinction is (at least originally) from vanilla Emacs and its packages. E,g. code for managing buffers which should [not] be garbage collected. So it's not Spacemacs initiated term.
That's just my understanding/memory, EmacsWiki might have more about that
_No comment on the substance of this issue yet (too tired even to read it, so pardon if my msg is irrelevant)._

TheBB:
Presumably the useless buffer regexp is too permissive. It should perhaps be `"\\`\\*.*\\*\\'"`?

[...]

Issue #5129, opened 2016-02-18
`SPC TAB` alternate-buffer-in-persp is broken


syl20bnr:
#### Description
`SPC TAB` to go to back and forth to last buffer is broken.

#### Reproduction guide
- Start Emacs
- `SPC f e i`
- `SPC w V`
- `SPC f f core/core-funcs.el RET`
- `SPC 1`
- `SPC f f core/core-debug.el RET`
- `SPC 2`
- `SPC TAB`

*Observed behaviour:*
The displayed buffer is `core-debug.el`

*Expected behaviour:*
The displayed buffer should be `init.el`

#### System Info
- OS: gnu/linux
- Emacs: 24.4.1
- Spacemacs: 0.105.9
- Spacemacs branch: develop (rev. 2fef64e)
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(ansible
 (auto-completion :variables auto-completion-return-key-behavior 'complete auto-completion-tab-key-behavior 'cycle auto-completion-complete-with-key-sequence "jk")
 c-c  
 (colors :variables colors-enable-rainbow-identifiers t)
 django elixir elm emacs-lisp emoji erlang
 (evil-snipe :variables evil-snipe-enable-alternate-f-and-t-behaviors t)
 (eyebrowse :variables eyebrowse-display-help nil)
 (git :variables git-use-magit-next t git-magit-status-fullscreen nil)
 github html
 (javascript :variables javascript-disable-tern-port-files t)
 markdown
 (org :variables org-enable-github-support t)
 perforce php python ranger semantic
 (shell :variables shell-default-shell 'eshell)
 syntax-checking spell-checking vagrant version-control windows-scripts)
```

cc @CestDiego

CestDiego:
Why is the expected buffer `core-debug.el` ? I had to rewrite the alternate
buffer function so it doesn't use the evil one because it did something not
in the way emacs is supposed to (by checking the buffer list- which is
convenient because persp-mode changes the buffer list)

On Thu, Feb 18, 2016 at 2:13 PM Sylvain Benner <notifications@github.com>
wrote:

> Description
>
> SPC TAB to go to back and forth to last buffer is broken.
> Reproduction guide
>
>    - Start Emacs
>    - SPC f e i
>    - SPC w V
>    - SPC f f core/core-funcs.el RET
>    - SPC 1
>    - SPC f f core/core-debug.el RET
>    - SPC 2
[...]

PR #8530, opened 2017-03-15
do not filter for alternating buffers

### LINKED ISSUE #7104: Cant switch to last buffer (SPC tab) to spacemacs home buffer
I just switched to the develop branch and found this is a thing. Is this intended?

deb0ch:
Hey hey !

Now that I'm done taking care of my responsive Spaceline PR I'm back into business, so I stopped deffering it and fixed #7104 (and felt a little stupid as it was so quick to do) !

@syl20bnr if I understood you right I'm sure you will like this one 😸 

I also removed the `spacemacs/alternate-buffer-in-persp` function that wasn't used or bound, and that we don't want anyways because it has that same behaviour that was so annoying of filtering out buffers.

### DIFF
--- a/core/core-funcs.el
+++ b/core/core-funcs.el
@@ -306,19 +306,13 @@ buffer."
   "Switch back and forth between current and last buffer in the
 current window."
   (interactive)
-  (let ((current-buffer (window-buffer window))
-        (buffer-predicate
-         (frame-parameter (window-frame window) 'buffer-predicate)))
-    ;; switch to first buffer previously shown in this window that matches
-    ;; frame-parameter `buffer-predicate'
+  (let ((current-buffer (window-buffer window)))
+    ;; if no window is found in the windows history, `switch-to-buffer' will
+    ;; default to calling `other-buffer'.
     (switch-to-buffer
-     (or (cl-find-if (lambda (buffer)
-                       (and (not (eq buffer current-buffer))
-                            (or (null buffer-predicate)
-                                (funcall buffer-predicate buffer))))
-                     (mapcar #'car (window-prev-buffers window)))
-         ;; `other-buffer' honors `buffer-predicate' so no need to filter
-         (other-buffer current-buffer t)))))
+     (cl-find-if (lambda (buffer)
+                   (not (eq buffer current-buffer)))
+                 (mapcar #'car (window-prev-buffers window))))))
 
 (defun spacemacs/alternate-window ()
   "Switch back and forth between current and last window in the
--- a/layers/+spacemacs/spacemacs-layouts/funcs.el
+++ b/layers/+spacemacs/spacemacs-layouts/funcs.el
@@ -46,13 +46,6 @@ Cancels autosave on exiting perspectives mode."
                        *persp-hash* 'non-existent))
     (persp-switch spacemacs--last-selected-layout)))
 
-(defun spacemacs/alternate-buffer-in-persp ()
-  "Switch back and forth between current and last buffer in the
-current perspective."
-  (interactive)
-  (with-persp-buffer-list ()
-                          (switch-to-buffer (other-buffer (current-buffer) t))))
-
 (defun spacemacs-layouts/non-restricted-buffer-list-helm ()
   (interactive)
   (let ((ido-make-buffer-list-hook (remove #'persp-restrict-ido-buffers ido-make-buffer-list-hook)))
--- a/layers/+spacemacs/spacemacs-layouts/packages.el
+++ b/layers/+spacemacs/spacemacs-layouts/packages.el
@@ -210,7 +210,6 @@
       (spacemacs/declare-prefix "B" "global-buffers")
       ;; Override SPC TAB to only change buffers in perspective
       (spacemacs/set-leader-keys
-        "TAB"  'spacemacs/alternate-buffer-in-persp
         "ba"   'persp-add-buffer
         "br"   'persp-remove-buffer))))
 


core/core-funcs.el
```diff
@@ -297,19 +297,13 @@ buffer."
...
     ;; switch to first buffer previously shown in this window that matches
     ;; frame-parameter `buffer-predicate'
     (switch-to-buffer
-     (or (cl-find-if (lambda (buffer)
-                       (and (not (eq buffer current-buffer))
-                            (or (null buffer-predicate)
-                                (funcall buffer-predicate buffer))))
-                     (mapcar #'car (window-prev-buffers window)))
-         ;; `other-buffer' honors `buffer-predicate' so no need to filter
-         (other-buffer current-buffer t)))))
```
bmag: `other-buffer` is used when there isn't any suitable buffer in `window-prev-buffers`. However, it's ok to remove it because it wasn't needed in the first place. If `cl-find-if` returns `nil`, then `switch-to-buffer` will default to calling `other-buffer` (because it got `nil` as its BUFFER argument). It's worth a comment in the code, though.

bmag:
Remember that `buffer-predicate` also tests if a buffer belongs to the current layout (perspective). By removing it, we allow `spacemacs/alternate-buffer` to show buffers that don't belong, for example Helm buffers. This is a bug that we had, but I can't find the exact report right now.

We really need to collect here an index of all relevant issues/PRs (including closed ones), to get the full context and make sure we don't cause regressions.

Also, I think we should improve the default useless/useful buffers and the relevant predicate functions (`spacemacs-useful-buffers-regexp`, `spacemacs-useless-buffers-regexp`, `spacemacs/useful-buffer-p`, `spacemacs/useless-buffer-p`). I'm quite sure we can mitigate most of the problem by only disabling specific buffers, instead of almost everything that isn't a file or a shell.

Stebalien:
See #6683.

deb0ch:
The helm bug was about navigating buffers, for instance with `SPC b n`, that would bring you to helm buffers that you usually don't want to see (I think I was one of the ones who reported that one). 

The point here is getting back to the last visited buffer whatever it was, so I think it needs no filtering at all (and if I understood him right [@syl20bnr agreed](https://github.com/syl20bnr/spacemacs/issues/7104#issuecomment-277888520) on that).

I've been using it since a little bit before I posted it and I haven't seen a Helm buffer since then, and I use it a lot.

core/core-funcs.el
deb0ch: Oh, I didn't even take that into account 😮 
In my thinking, if there was nothing in that list then, well, then there was nothing to get back to 😸

bmag:
> The point here is getting back to the last visited buffer whatever it was, so I think it needs no filtering at all

That probably works for all cases except when a user presses `SPC TAB` right after switching a layout (because one of the windows isn't deleted, so it retains its buffer history). We could just say we don't support this case and have it as a known issue.

Other than that I'm fine with this PR.

deb0ch:
I don't think that showing a window from another layout is really undesirable. Having it that way gives it the simplest logic: "show the last window that was here whatever it was".

Having it that simple makes it easy to anticipate and POLA-friendly: less neurons for the user to use on his editor, more on his code.

And even in the case where we would want a different window history per layout, I don't think that implementing it using this filter would work. Instead, I think that it would best work if it was `eyebrowse` (or `persp` ?) who killed and restored windows keeping their history instead of reusing them.

bmag:
> And even in the case where we would want a different window history per layout, I don't think that implementing it using this filter would work.

Neither do I, and I'm not sure if we could get it to work reliably either.

> Instead, I think that it would best work if it was eyebrowse (or persp ?) who killed and restored windows keeping their history instead of reusing them.

They use the built-in `window-state-get` and `window-state-put`, so it's more of an Emacs issue. And yes, it would work best if Emacs handled it.

> I don't think that showing a window from another layout is really undesirable. Having it that way gives it the simplest logic: "show the last window that was here whatever it was".
>
> Having it that simple makes it easy to anticipate and POLA-friendly: less neurons for the user to use on his editor, more on his code.

So basically, you're saying it's a feature, while I'm thinking of it as a known issue that isn't worth the effort and probably complex solution required to fix it. Anyway, both of us agree with merging this PR. (I'm waiting to see if there are any other comments before merging it, will probably merge it next Friday)

core/core-funcs.el
deb0ch: Done commenting :thumbsup:

deb0ch:
I definitely think it's a feature :smile_cat:

deb0ch:
> They use the built-in window-state-get and window-state-put, so it's more of an Emacs issue. And yes, it would work best if Emacs handled it.

Thanks for the pointer !

`persp-mode` and `eyebrowse` transformed me Emacs experience so deeply, they are so awesome, so well integrated into Spacemacs, but at the same time there are still such a tremendous amount of things to improve in them !

Maybe will I get to it one of these days :construction_worker:

Anyways, glad we agree :smile_cat:

bmag:

deuill:
I realize this has been merged already, but I feel this may be a regression, or at least does not match expectations:

Perspectives are designed to isolate buffers based on user assignment -- this is important, since it's up to the user to decide which buffers are relevant to them at that point in time. Hijacking this decision arguably makes perspectives less useful, as there are no more guarantees of isolation.

I see alternating between buffers with `SPC TAB` as a quicker alternative to `SPC b b RET` (which, at least with Helm, will switch between the last two buffers used in a perspective). While the semantics for the former are (presumably) less clear, there's no indication why the two would differ in any significant way. If anything, more actions (such as `evil-jump-forward` or `evil-jump-backward`) should be made to respect the isolation guarantees.

A simple test for me is: if I switch between projects in perspectives and jump back to a buffer that belongs to the other, does my perspective change? Does the buffer now belong to both perspectives? Currently, the answer to both is "no", which, I feel, breaks expectations. Jumping back to the previous buffer between perspectives makes sense, but you can already do that by simply switching to the relevant perspective with `SPC l <num>`.

Clarifying the semantics for all buffer-switching actions, with regards to perspectives, may help with resolving this one way or the other.

deb0ch:
Buffer isolation between perspectives is indeed desirable, but I don't think any of it should be implemented in a `SPC tab` command: it would just be a hacky piece of tape over a base that needs to be improved.

The problem you are describing rather lies within `persp-mode` itself. It is that, when switching perspectives, `persp-mode` reuses the same emacs windows, keeping and altering their history. Instead, it should keep a set of emacs windows in memory for each perspectives and kill and re-display them at every perspective switch.

This way there would be one history per window and perspective, the problem of isolation wouldn't arise anymore when switching perspectives, and `SPC tab` could keep being as simple as it is right now.

One other thing is that we wouldn't want to prevent `SPC tab` from showing a buffer that is not in the current perspective, because then it gets infuriating when you can't get back to a terminal, magit, dired, Messages or others buffers that were just displayed there but were not added automatically to the current perspective.

So I think that, as this way might not be the most optimal, it *is* the best we can get until `persp-mode` gets improved.

And, as I said earlier, I'm having an increasing urge to dive into `persp`, as I see more and more little things that I really want improved in there. Just waiting for my time to allow it, soon I'll have more free time :smile_cat:

syl20bnr:
Sorry to be late to the party but this PR actually introduces the regression mentioned by @deuill. 
The bug was reported here (by myself 😄 ): https://github.com/syl20bnr/spacemacs/issues/5129

Isolation of buffers is indeed desirable and in my book we break POLA if we break it without an explicit command that breaks it (for instance `SPC B b`). 

If it cannot be easily fixed in persp-mode then we have to fix it on our side.

syl20bnr:
> if I understood you right I'm sure you will like this one

@deb0ch Indeed this change makes me happy 💜  But we should keep the persp isolation if we can.

deb0ch:
I changed my mind a little bit on that matter, I no longer think that no perspective isolation is a good idea.

However, I think that the perspective isolation we need is about the window history, not the buffers themselves: if a buffer that doesn't belong to the persp was just there, for instance `*Magit`, `*Messages*`, `*term`, `*dired: ` or whatever, we want to get back to it when using `SPC tab`.

It is if the buffer was just there because of a perspective change that we don't want to get back to it, right ?

This is why I think that this problem should be fixed by modifying the way `persp-mode` switches perspectives and reuses existing windows when doing so. Changing perspectives shouldn't alter the perspective's window histories.

Does it make sense ?

deuill:
I think the following can be used as indicators for what the isolation guarantees should be:

  * Behaviour between perspectives, windows, and separate Emacs instances should be the same. Buffers that open within the same window and instance as the perspective should be considered to be isolated from other perspectives, like they would be between other windows and instances.
  * Only buffers that would otherwise appear in the <kbd>SPC</kbd><kbd>b</kbd><kbd>b</kbd> list of buffers should be considered as eligible. That means that deleted buffers shouldn't be switched to.
This is different from what, e.g. `evil-jump-backward` would do, since these commands explicitly track history. Not sure why <kbd>SPC</kbd><kbd>Tab</kbd> should work any different from the <kbd>Alt</kbd><kbd>Tab</kbd> scheme (or variations thereof) used in window managers, which don't typically revive closed windows.
  * Whether "special" buffers such as `*Messages*`, `*dired*` etc. should be added to the current perspective is probably a different matter entirely. I tend to think of these not as "useful" or not (which is the current distinction used in Spacemacs), but as transient (or not). 
I'm more likely to consider a `dired` buffer transient than, say, a `term` buffer. If these buffers are navigated to due to user action (and are in the same window), then they should probably be part of the perspective and thus eligible for switching to using <kbd>SPC</kbd><kbd>Tab</kbd>.

As always, IMHO, but the above definitely fits my use of perspectives.

PR #8532, opened 2017-03-15
Allow 'w','b','j' and 'k' movements inside evil-numbers-transient-state

mmajcher:
Those movements allow to change more than one number without repeating SPC n + / SPC n -.

'w' is very handy as evil-numbers take you automatically to the next number once you are past the changed one.

### DIFF
--- a/layers/+spacemacs/spacemacs-evil/packages.el
+++ b/layers/+spacemacs/spacemacs-evil/packages.el
@@ -187,11 +187,17 @@
     (progn
       (spacemacs|define-transient-state evil-numbers
         :title "Evil Numbers Transient State"
-        :doc "\n[_+_/_=_] increase number  [_-_] decrease  [0..9] prefix  [_q_] quit"
+        :doc "\n[_+_/_=_] increase number  [_-_] decrease  [0..9] prefix  [_w_/_b_] move forward/backwards  [_j_/_k_] go line up/down  [_q_] quit"
         :bindings
         ("+" evil-numbers/inc-at-pt)
         ("=" evil-numbers/inc-at-pt)
         ("-" evil-numbers/dec-at-pt)
+        ("_" evil-numbers/dec-at-pt)
+        ("b" evil-backward-word-begin)
+        ("w" evil-forward-word-begin)
+        ;; ^ best would be to move point to the next/prev number in buffer
+        ("j" evil-next-line)
+        ("k" evil-previous-line)
         ("q" nil :exit t))
       (spacemacs/set-leader-keys
         "n+" 'spacemacs/evil-numbers-transient-state/evil-numbers/inc-at-pt


Somelauw:
I like the idea, but 'b' and 'w' have the limit that they only work if the numbers are located next to each other. j and k have the same problem but vertically.

I think it would be better if there were key bindings in this micro-state that always bring you to the previous/next number, no matter where they are located.

mmajcher:
(WHEN NOT IN A REGION) When you do evil-numbers' 'inc-at-point' it executes its function evil-numbers/search-number. This uses re-search-forward to find "[[:digit:]]" with optional argument (BOUND) = (point-at-eol), so it searches up to the end of the current line. 

If given no BOUND it would search up to the end of buffer, but this function belongs to evil-number, not spacemacs and I don't know how to handle that.

The same function could be used to find previous number if re-search-forward would be replaced with re-search-backward.

TheBB:
We can steal that function and repurpose it for ourselves. It's a little bit of a challenge to write one that works the intended way though.

TheBB:
This can probably be done simpler, but it works for me.

    (defun next-number ()
      (interactive)
      (let ((regexp (rx (and (or line-start (not digit)))
                        (group (or (and "0x" (+ hex-digit))
                                   (and "0o" (+ (in "0-7")))
                                   (and "0b" (+ (in "01")))
                                   (+ digit)))))
            (old-point (point))
            (new-point (point-at-bol)))
        (save-excursion
          (goto-char new-point)
          (while (and (<= new-point old-point)
                      (re-search-forward regexp (point-max) t))
            (setq new-point (match-beginning 1))))
        (when (and new-point (> new-point old-point))
          (goto-char new-point))))

    (defun previous-number ()
      (interactive)
      (let ((regexp (rx (and (or line-start (not digit)))
                        (group (or (and "0x" (+ hex-digit))
                                   (and "0o" (+ (in "0-7")))
                                   (and "0b" (+ (in "01")))
                                   (+ digit)))))
            (old-point (point))
            (new-point (point-at-eol))
            (end-point (point)))
        (save-excursion
          (goto-char new-point)
          (while (and (>= end-point old-point)
                      (re-search-backward regexp (point-min) t))
            (setq new-point (match-beginning 1))
            (setq end-point (match-end 1)))
          (goto-char (point-at-bol))
          (while (and (< (point) end-point)
                      (re-search-forward regexp (point-at-eol) t)))
          (when (>= (point) end-point)
            (setq new-point (match-beginning 1))))
        (when (and new-point (< new-point old-point))
          (goto-char new-point))))

syl20bnr:
Thank you for the PR and sorry for the long delay, your repo does not exist anymore, can you resubmit a new PR ?

Closing this PR as it cannot be merged.

PR #8533, opened 2017-03-15
put vagrant bindings under `a` prefix

deb0ch:
:octocat:


bmag:
Thank you :+1: Cherry-picked into develop, you can safely delete your branch. I've updated the docs as well.

deb0ch:
Oooops, forgot about the docs, thanks for doing it :sweat_smile:

Issue #8537, opened 2017-03-16
Tame exec-path-from-shell


Stebalien:
The `exec-path-from-shell` package is a horrible hack that should only be used as a last resort.

Each time it imports an environment variable, it starts an interactive login shell which:

1. Can be really slow (logins are expected to be infrequent).
2. Can break systems that (reasonably) expect one login per session.
3. Breaks spectacularly when one does something like run screen in interactive shells. Note: This may be fixed when and if https://github.com/purcell/exec-path-from-shell/issues/17 is.

To mitigate these issues:

* [x] It should be possible to exclude the `exec-path-from-shell` package. Unfortunately, the go and rust layers expect `exec-path-from-shell` to always be available and crash on startup when it's excluded.
  - [x] `+lang/rust`
  - [x] `+lang/go`

* [ ] It shouldn't be enabled on anything but MacOS. Any not-totally-broken display manager on Linux, FreeBSD, etc. will properly setup the environment using the user's login shell.

* [x] Layers should avoid importing variables from the shell when already defined as this takes away control from the user (having one's environment silently overridden is mildly infuriating). Currently, the rust layer imports `RUST_SRC_PATH` even if already set.
  - [x] `+lang/rust`

* [x] Layers should avoid calling `exec-path-from-shell-copy-env` multiple times as it will run an interactive login shell each time (which is *slow*). The go layer currently calls it twice.
  - [x] `+lang/go`

bmag:
@Stebalien I tend to agree with you, and several people indeed had trouble caused by exec-path-from-shell because it runs an interactive login shell and they had a heavy configuration (IIRC there are several such issues in the bug tracker). I'd like to hear from someone who does need exec-path-from-shell (mainly macOS users I guess), to understand if we can safely disable it in non-macOS machines. Also, Rust and Go layers will need an alternative method of reading the environment variable (at least on non-macOS).

Stebalien:
> I'd like to hear from someone who does need exec-path-from-shell (mainly macOS users I guess), to understand if we can safely disable it in non-macOS machines.

Wouldn't you want to hear from more non-macos users (unless you're looking to remove it entirely). 

Display/Login Managers (I'm reasonably certain all of these work...):

* [x] `login` (the standard textual login manager always starts a login shell).
* [ ] GNOME 3/GDM
* [ ] KDE/KDM
* [ ] XFCE/XDM
* [ ] MATE
* [ ] LXWM/LXDM
[...]

PR #8543, opened 2017-03-18
Partially tame exec-path-from-shell

Stebalien:
1. Make it possible to exclude the package (fix the go and rust layers).
2. Import variables all at once (avoid spawning multiple login shells).
3. Import variables early (during the "pre" package loading stage).
4. Centralize the platform check by using a package toggle in the `spacemacs-base`.
5. Avoid importing already defined variables (except `PATH`/`MANPATH`).

Relates to #8537.

TODO: We can probably trim the platform list to macos only.

### DIFF
--- a/layers/+lang/go/packages.el
+++ b/layers/+lang/go/packages.el
@@ -18,6 +18,7 @@
                                              'flycheck)))
         ggtags
         helm-gtags
+        exec-path-from-shell
         go-eldoc
         go-mode
         go-guru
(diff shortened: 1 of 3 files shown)


layers/+lang/go/packages.el
```diff
@@ -37,12 +38,14 @@
...
 
-(defun go/init-go-mode()
-  (when (memq window-system '(mac ns x))
-    (dolist (var '("GOPATH" "GO15VENDOREXPERIMENT"))
-      (unless (getenv var)
-        (exec-path-from-shell-copy-env var))))
+(defun go/pre-init-exec-path-from-shell ()
+  (spacemacs|use-package-add-hook exec-path-from-shell
+    :pre-config
+    (dolist (var '("GOPATH" "GO15VENDOREXPERIMENT") exec-path-from-shell-variables)
```
Stebalien: For some reason, returning `nil` here stops the package configuration. Is that expected?

Stebalien: Looks like a bug. PR: #8544

bmag:
Haven't tested it, but it looks excellent. Any Go or Rust users (on macOS or Linux) tested this PR? (that's more a question to the general audience, not directly to @Stebalien)

bmag:

PR #8544, opened 2017-03-19
Always return `t` from use-package pre/post hooks.

Stebalien:
Hooks that don't do this will prevent any further configuration of the package.

Docs: https://github.com/jwiegley/use-package/blob/53bf803f1d3efc61653f94fe56ff30a72304861e/use-package.el#L125
Context: https://github.com/syl20bnr/spacemacs/pull/8543/files#r106792261


bmag:

syl20bnr:
@Stebalien Nice catch 👍

PR #8545, opened 2017-03-19
[WIP] Add pippel to Python layer

et2010:
Add pippel for Python package management.

> This package is an Emacs frontend for the Python package manager pip. As pippel also uses tabulated-list-mode, it provides a similiar package menu like package-list-packages.

Here is a screenshot:

image

I'm not so sure about the <kbd>SPC a l</kbd> keybinding. If you got a better idea, please let me know.

### DIFF
--- a/layers/+lang/python/packages.el
+++ b/layers/+lang/python/packages.el
@@ -27,6 +27,7 @@
     (nose :location local)
     org
     pip-requirements
+    pippel
     py-isort
     pyenv-mode
     (pylookup :location local)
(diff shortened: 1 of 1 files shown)


et2010:
Sorry for the mess, I created a branch on the wrong commit, how do I fix this?

syl20bnr:
Thank you ! 💜 

I moved the key bindings to major mode specific `SPC m P`. To call this outside of a python buffer use `SPC SPC pippel-package-list`. See commit https://github.com/syl20bnr/spacemacs/commit/eee2b4b8

PR #8548, opened 2017-03-19
Late expand completion backends in the `spacemacs|add-company-backends` macro

Stebalien:
Otherwise, the Haskell layer throws an eager expansion error because `layers/+lang/haskell/funcs.el`, which calls this macro, is loaded before `layers/+completion/auto-completion/config.el`, which defines `spacemacs-default-company-backends`.

Late expansion may have unintended consequences. Was there a reason this variable was eagerly expanded?


syl20bnr:
Nice find, this was not intended.
Thank you ! 💜 

PR #8554, opened 2017-03-20
Default enabel smartparen in SLIME REPL

Habens:
**smartparens** provide many useful features, is there any special reason disable the smartparens in SLIME REPL?

### DIFF
--- a/layers/+lang/common-lisp/packages.el
+++ b/layers/+lang/common-lisp/packages.el
@@ -54,11 +54,11 @@
       (setq slime-complete-symbol*-fancy t)
       (setq slime-complete-symbol-function 'slime-fuzzy-complete-symbol)
       ;; enabel smartparen in code buffer and SLIME REPL
-      ;; (add-hook 'slime-repl-mode-hook #'smartparens-strict-mode)
-      (defun slime/disable-smartparens ()
-        (smartparens-strict-mode -1)
-        (turn-off-smartparens-mode))
-      (add-hook 'slime-repl-mode-hook #'slime/disable-smartparens)
+      (add-hook 'slime-repl-mode-hook #'smartparens-strict-mode)
+      ;;(defun slime/disable-smartparens ()
+      ;;  (smartparens-strict-mode -1)
+      ;;  (turn-off-smartparens-mode))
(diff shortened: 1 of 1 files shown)


Chream:
Personally I dont like have SM on while in the repl. The reason is that if I am writing something directly in the repl with SM=on then hitting return will do an eval of the expression and not go to a new line.

Habens:
Hi  @Chream 
Yes, there is the problem in the repl. But I think there is only '(' will bring this issue, we shouldn't disable whole SM.
And there are two solutions for this issue if enable SM in the repl:
1st: before type '(', using 'C-q' to prevent autocomplete the ')'. This way works in any mode.
2nd: remove the '(' pair from SM in slime-repl-mode, than only if you manually type the ')' and hit return, that expression will be eval.

Chream:
@Habens 
I would not consider using C-q first an option. That slows down typing considerably. The second option would be ok. 

One thing to consider is that matching pairs in the repl can become extremely slow and laggy if the buffer has a lot of text. I would like to have SM on for the repl but every time I have tried to configure it I never end up satisfied.

dangom:
Why not adding the following to the config?
```elisp
(defvar slime-repl-enable-smartparens nil
  "If non nil smartparens is enabled on SLIME REPL.")
```

And then

```elisp
(defun slime/disable-smartparens ()
      (smartparens-strict-mode -1)
      (turn-off-smartparens-mode))

(when slime-repl-enable-smartparens
      (add-hook 'slime-repl-mode-hook #'slime/disable-smartparens))
```

So you can `(common-lisp :variables slime-repl-enable-smartparens t)`?


Seems like a much safer and cleaner thing to do than enforcing smartparens on all spacemacs users, or coming up with hacky workarounds.

syl20bnr:
I would go the @dangom's route.

Habens:
@dangom's idea is pretty awesome. 

BTW
1st: Just an remander only type once 'C-q' is enough, cause only outermost '(' and ')' will bring the expression eval
2nd: Below is an idea that remove "(" pair from SM in slime repl:
```
(defun slime/smartparens ()
	(sp-local-pair 'slime-repl-mode "(" nil :actions nil)
	(smartparens-strict-mode))
(add-hook 'slime-repl-mode-hook #'slime/smartparens)
```

PR #8555, opened 2017-03-21
Add evil-lion

et2010:
Add evil-lion for operator alignment. More details here: https://github.com/edkolev/evil-lion

### DIFF
--- a/layers/+spacemacs/spacemacs-evil/packages.el
+++ b/layers/+spacemacs/spacemacs-evil/packages.el
@@ -16,6 +16,7 @@
         evil-exchange
         evil-iedit-state
         evil-indent-plus
+        evil-lion
         evil-lisp-state
         ;; for testing purpose, contribute by reporting bugs and sending PRs
         ;; to https://github.com/gabesoft/evil-mc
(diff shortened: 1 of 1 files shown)


Somelauw:
I think evil-lion looks nice, because it acts like an operator.

Unfortunately, the gl binding conflicts with current Spacemacs conventions:
https://github.com/syl20bnr/spacemacs/blob/master/doc/CONVENTIONS.org#movement-in-normal-mode

By the way, Spacemacs already has some alignment bindings on SPC xa.

et2010:
Thanks for pointing that out, @Somelauw. I didn't notice it before.
Maybe we could add a toggle for evil-lion?

Somelauw:
Maybe put it on ga?

TheBB:
I think it would be ok to put a keybinding under `SPC x a`, no? I don't personally use alignment so often that this would be prohibitive, but maybe others do.

Seems it would suffice to set `evil-lion-left-align-key` and `evil-lion-right-align-key` to `nil` before loading the mode, and then adding leader bindings for `evil-lion-left` and `evil-lion-right`.

TheBB:
I made some changes at 529045630496faafa0706a0c486cb9cff7b184df reflecting my previous comment. Also added documentation.

peterhoeg:
What is the proper way to use evil-lion as an operator with spacemacs?

PR #8560, opened 2017-03-21
Load org-mode email integration (mu4e/notmuch) when org is loaded.

Stebalien:
Before, `org-mu4e` would be loaded when mu4e was. However, as an extension to
org mode, not mu4e, it should be loaded when org mode is.

This commit also adds similar support for notmuch.

This supersedes #7977.

### DIFF
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -18,7 +18,6 @@
     evil-surround
     gnuplot
     htmlize
-    mu4e
     ;; ob, org and org-agenda are installed by `org-plus-contrib'
     (ob :location built-in)
     (org :location built-in)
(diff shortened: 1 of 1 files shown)


TheBB:

syl20bnr:
There was some confusion here. `mu4e` should be part of the org layer, it feels it should be the other way around. And it was too :-)
Check: https://github.com/syl20bnr/spacemacs/blob/develop/layers/+email/mu4e/packages.el#L102-L106

So I removed the `org-mu4e` and `org-notmuch` from `org` layer and just added `org-notmuch` in the already existing `org-post-init` of `mu4e` layer.
commit: https://github.com/syl20bnr/spacemacs/commit/fcc8b9a3

Stebalien:
Ah. I didn't realize it was being loaded in two places. Now that you point it out, this definitely does go in the mu4e layer (like flycheck, helm, etc. integration). Sorry about that.

PR #8563, opened 2017-03-22
Switch off `spacemacs-purpose` if not enabled from .spacemacs

bixuanzju:
The README says to enable it, add `spacemacs-purpose` to the dotfile.

### DIFF
--- a/layers/+distributions/spacemacs/layers.el
+++ b/layers/+distributions/spacemacs/layers.el
@@ -17,7 +17,6 @@
                                       spacemacs-evil
                                       spacemacs-language
                                       spacemacs-misc
-                                      spacemacs-purpose
                                       spacemacs-ui
                                       spacemacs-ui-visual
                                       spacemacs-org))
--- a/layers/+spacemacs/spacemacs-purpose/README.org
+++ b/layers/+spacemacs/spacemacs-purpose/README.org
@@ -68,7 +68,7 @@ result is a better control over how buffers are displayed, since
 
 * Install
 To use this configuration layer, add it to your =~/.spacemacs=. You will need to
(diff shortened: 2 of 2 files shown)


nixmaniack:
I think the layer README was not updated when the `purpose` layer was made part of Spacemacs experience. So the documentation should be changed and not the other way around :).

bixuanzju:
Personally I don't like purpose. If this is part of Spacemacs, then at least let me know how to disable it.

nixmaniack:
Unaware of the clean way to disable full layer which are part of the spacemacs distribution. But `dotspacemacs-excluded-packages '(window-purpose)` comes to my mind as an easy fix. The hard way would be to go with `spacemacs-core` and build up from there.

bmag:
> I think the layer README was not updated when the `purpose` layer was made part of Spacemacs experience. So the documentation should be changed and not the other way around :).

Indeed.

bmag:
I'm closing this one, but feel free to submit a PR fixing the documentation.

bixuanzju:
@nixmaniack Well I need to also exclude `helm-purpose` among other packages used by `spacemacs-purpose`. That's not satisfactory.  Maybe we could provide, like `dotspacemacs-excluded-packages`, something like `dotspacemacs-excluded-layer`.

bixuanzju:
ok, found a way, at least feel regain the control. use `spacemacs-base`, then add layers I want.

nixmaniack:
> That's not satisfactory. Maybe we could provide, like `dotspacemacs-excluded-packages`, something like `dotspacemacs-excluded-layer`

I can totally understand and agree.

PR #8564, opened 2017-03-22
Adds the `neotree-quick-look` keybind

sdwolf:
Weirdly, the `'` keybind for `neotree-mode-map` was already present when
doing `M-x neotree-quick-look` with the cursor on the neotree window but
I could not find where exactly it was defined.

This commit officially adds it to the `neotree-mode-map` and the neotree
transient state.


syl20bnr:
Thank you ! 👍 

PR #8576, opened 2017-03-24
Make  @robbyoconnor happy again

JAremko:
Now Travis should detect and report misplaced white spaces.

Fixes #7456

**drop test commits**

example https://travis-ci.org/syl20bnr/spacemacs/jobs/214563770

### DIFF
--- a/.travis-build.sh
+++ b/.travis-build.sh
@@ -21,7 +21,7 @@ fi
 
 if  [ $TRAVIS_SECURE_ENV_VARS = false ] &&
     [ $TRAVIS_PULL_REQUEST != false ] &&
-    [ $TRAVIS_BRANCH = "master" ]; then
+    [ "$TRAVIS_BRANCH" = "master" ]; then
 
     	printf '=%.0s' {1..70}
     	printf "\n       し(*･∀･)／   Thanks for the contribution!  ＼(･∀･*)ノ\n"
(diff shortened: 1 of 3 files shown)


JAremko:
@robbyoconnor

JAremko:
ok seems to work as expected https://travis-ci.org/syl20bnr/spacemacs/builds/214555767

JAremko:
ok seems to work as expected https://travis-ci.org/syl20bnr/spacemacs/builds/214563765
But it needs to be battle tested. :shipit: 

@robbyoconnor  Are you happy now? :roll_eyes:

robbyoconnor:
you're awesome!

JAremko:
Hmm. It seems that Travis doesn't warn when a test with "Allowed Failures" fails. So we probably should keep whitespace test critical https://travis-ci.org/syl20bnr/spacemacs/builds/214566373 but probably not.

> you're awesome!  

@robbyoconnor  What a news...  :roll_eyes:

.travis-build.sh
```diff
@@ -32,6 +32,23 @@ if  [ $TRAVIS_SECURE_ENV_VARS = false ] &&
     	exit 1
 fi
 
+if [ "${FORMATTING}" = "space-test" ]; then
+	cd "${TRAVIS_BUILD_DIR}"
+	echo "Testing for trailing and all sorts of broken white spaces"
+	echo "TRAVIS_COMMIT_RANGE: ${TRAVIS_COMMIT_RANGE}"
+	first_commit=`echo ${TRAVIS_COMMIT_RANGE} | sed -r 's/\..*//'`
+	git reset -q "${first_commit}"
+##	git reset -q HEAD~1
```
JAremko: @d12frosted  looks like this line was a bit too much.  "in depth"  :smile: 

At the same time it's a pretty neat idea if you think about it. When someone opens a PR CI will fail if someone committed malformed code before this person so the PR author will have to fix it. Hmmmhmmm :smiling_imp:

.travis-build.sh
```diff
@@ -21,7 +21,7 @@ fi
 
 if  [ $TRAVIS_SECURE_ENV_VARS = false ] &&
     [ $TRAVIS_PULL_REQUEST != false ] &&
-    [ $TRAVIS_BRANCH = "master" ]; then
+    [ "$TRAVIS_BRANCH" = "master" ]; then
```
d12frosted: Oh, this looks like the right change anyways.

.travis-build.sh
d12frosted: Looks good. Force people to fix other peoples bad spaces! 😈

.travis-build.sh
JAremko: Yeah. In case if we gonna have weird branch names

.travis-build.sh
robbyoconnor: `SPC x d w` exists -- if they're too lazy.

syl20bnr:
Thank you ! 👍  
I like PR which makes people happy.
Cherry-picked only the first commit into develop branch, you can safely delete your branch.

PR #8582, opened 2017-03-26
Rewrite sudo-edit

dvzubarev:
Old version worked only for method 'ssh' and did not work at all if
an user was not specified (e.g. /ssh:hostname:/path). Since the symbol '@' was
required in a filename.

### DIFF
--- a/layers/+distributions/spacemacs-base/funcs.el
+++ b/layers/+distributions/spacemacs-base/funcs.el
@@ -400,23 +400,31 @@ FILENAME is deleted using `spacemacs/delete-file' function.."
 ;; from magnars
 (defun spacemacs/sudo-edit (&optional arg)
   (interactive "P")
+  (eval-when-compile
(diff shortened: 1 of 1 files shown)


Stebalien:
I've briefly tested this and it appears to work as advertised.

layers/+distributions/spacemacs-base/funcs.el
```diff
@@ -10,6 +10,8 @@
 ;;; License: GPLv3
 
 (require 'cl-lib)
+(eval-when-compile
+  (require 'tramp))
```
TheBB: Move this into the function that uses it, maybe? That way we don't load tramp on startup. (I know tramp is usually loaded on startup anyway, but…)

dvzubarev: Done

TheBB:

PR #8586, opened 2017-03-27
scala: fix flyspell setup

andrebeat:
This PR fixes Scala's layer flyspell setup which was calling an undefined function.

### DIFF
--- a/layers/+lang/scala/packages.el
+++ b/layers/+lang/scala/packages.el
@@ -54,7 +54,7 @@
 
 (defun scala/post-init-flyspell ()
   (spell-checking/add-flyspell-hook 'scala-mode)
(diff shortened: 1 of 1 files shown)


bmag:
I went ahead and merged this, since it seems like a typo. However, we can probably get rid of the following as well:
```elisp
(spell-checking/add-flyspell-hook 'scala-mode)
```
Because 1) it doesn't do anything, should be `scala-mode-hook` instead of `scala-mode`, and 2) `spacemacs//java-setup-ensime-flyspell` already activates `flyspell-mode`, so by hooking it into `scala-mode-hook` it's redundant to hook `flyspell-mode` directly as well.

Issue #7186, opened 2016-09-24
Replace github-browse-file with browse-at-remote ?


JAremko:
Consider replacing [github-browse-file](https://github.com/osener/github-browse-file) with [browse-at-remote](https://github.com/rmuslimov/browse-at-remote)  - the last one seems to have more features.

Does anyone have experience working with the both tools?

PR #8601, opened 2017-03-29
Add browse at remote

JAremko:
Solve #7186

Instead of removing  `github-browse-file` I simply rebinded it to `ghO` and made `browse-at-remote` `gho` (it seems to be better)


syl20bnr:
Thank you ! Nice Find 👍 
I removed `github-browse-file` and made `github` layer "depend" on `version-control` layer.

PR #8614, opened 2017-03-31
[typescript] Use tslint and tsserver from node_modules if available

jupl:
Because different projects can have specific versions of typescript and/or tslint, it makes sense to use node_modules if available to avoid version clashing.


jupl:
I'm going to close this in favor of another PR that uses [this](https://github.com/codesuki/add-node-modules-path), which can be used for React, JavaScript, and TypeScript layers.

jupl:
#8621

PR #8615, opened 2017-03-31
Specified which versions of Python are compatible with anaconda mode ...

edwargix:
…in README

In README for Python layer, there is now a notice regarding which versions of Python work for anaconda-mode

### DIFF
--- a/layers/+lang/python/README.org
+++ b/layers/+lang/python/README.org
@@ -55,6 +55,9 @@ file.
 
 ** Dependencies
 *** Auto-completion: Anaconda dependencies
+=anaconda-mode= currently only supports Python versions 2.6, 2.7, 3.3, and 3.4.
+Source: https://github.com/proofit404/anaconda-mode#supported-python-versions
+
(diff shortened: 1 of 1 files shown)


cpaulik:
Not sure that this is correct. Anaconda-mode uses jedi in the background which should also be compatible with 3.5 according to https://github.com/davidhalter/jedi#feature-support-and-caveats

cpaulik:
See also https://github.com/proofit404/anaconda-mode/issues/220

edwargix:
Thanks for your help! I think you're right about it working with 3.5. However on my Arch system I am having some trouble getting it to work with 3.6. I just assumed that the versions from https://github.com/proofit404/anaconda-mode#supported-python-versions were tested. Should I update the branch to refer to jedi's supported versions instead?

cpaulik:
I'm not even sure that we should maintain such a list. It will probably just end up being out of date. Mentioning that Anaconda mode uses jedi and linking to jedi might be better.

edwargix:
Yeah I guess that sounds right. The README already mentions jedi.

Issue #2477, opened 2015-07-30
Unable to disable org-bullets


honza:
Yes, I know I can change the bullets to other unicode characters but how can I disable this to get the normal ascii bullets?

sooheon:
Add `org-bullets` to `dotspacemacs-excluded-packages`.
```
(dotspacemacs-excluded-packages '(org-bullets))
```

honza:
Most excellent, thank you!

tommyjiang:
Thanks for providing the solution @sooheon

muusik:
Since org-bullets has been replaced by org-superstar, disabling bullets can now be achieved thus:
`(dotspacemacs-excluded-packages '(org-superstar))`

PR #8617, opened 2017-03-31
Add a hint about disabling org bullets

ShaneKilkelly:
There is no documented way to disable fancy bullets in org mode.

I eventually found what I was looking for in this issue: https://github.com/syl20bnr/spacemacs/issues/2477

So I've added a line to the `org` layer documentation about this.


syl20bnr:
Thank you !

PR #8620, opened 2017-03-31
[evil-mc] Disable paste transient state when multiple cursors are active

korayal:
As discussed at https://github.com/gabesoft/evil-mc/issues/53 evil-mc has no support for the 'paste transient state', hence we should fall back to the default version of paste when multiple cursors are active.

### DIFF
--- a/layers/+spacemacs/spacemacs-evil/packages.el
+++ b/layers/+spacemacs/spacemacs-evil/packages.el
@@ -108,19 +108,21 @@
     :defer t
     :init
     (progn
-      (defun spacemacs-evil/evil-mc-set-paste (enable-paste-transient-state)
-        (if (and enable-paste-transient-state dotspacemacs-enable-paste-transient-state)
-          (progn
-            (define-key evil-normal-state-map "p" 'spacemacs/paste-transient-state/evil-paste-after)
-            (define-key evil-normal-state-map "P" 'spacemacs/paste-transient-state/evil-paste-before))
-          (progn
-            (define-key evil-normal-state-map "p" 'evil-paste-after)
-            (define-key evil-normal-state-map "P" 'evil-paste-before))))
+      (defun spacemacs-evil/evil-mc-paste-after (&optional count &optional register)
+        (interactive "p")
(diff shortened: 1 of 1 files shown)


bmag:
Won't it be more robust to write a function that calls either `spacemacs/paste-transient-state/evil-paste-*` or `evil-paste-*` depending on evil-mc's state, and bind that to `p/P` instead of `spacemacs/paste-transient-state/evil-paste-*`?

korayal:
@bmag I'm sorry I couldn't get my mind around your suggestion, doesn't `spacemacs-evil/evil-mc-set-paste`  do that?

or, are you suggesting that I should use something other than `evil-mc-before-cursors-created` hooks?

korayal:
Am I doing something wrong here?

I am able to get the right command called on both cases, but paste doesn't work on all cursors.

``` elisp
      (defun spacemacs-evil/evil-mc-paste-after (&optional arg)
        (interactive "p")
        (if (and (eq (evil-mc-get-cursor-count) 1) dotspacemacs-enable-paste-transient-state)
            (spacemacs/paste-transient-state/evil-paste-after)
            (evil-paste-after)))

      (defun spacemacs-evil/evil-mc-paste-before (&optional arg)
        (interactive "p")
        (if (and (eq (evil-mc-get-cursor-count) 1) dotspacemacs-enable-paste-transient-state)
            (spacemacs/paste-transient-state/evil-paste-before)
            (evil-paste-before)))

      (define-key evil-normal-state-map "p" 'spacemacs-evil/evil-mc-paste-after)
      (define-key evil-normal-state-map "P" 'spacemacs-evil/evil-mc-paste-before)
```

korayal:
updated the PR, but it will also require the PR at evil-mc to be accepted before proceeding.

korayal:
If there is nothing else, the mentioned PR from evil-mc lib is merged, so this is also ready to go.

syl20bnr:
Thank you ! 👍 
**Warning** I renamed the functions in commit https://github.com/syl20bnr/spacemacs/commit/fed06956
Please adjust the function names accordingly upstream.

PR #8621, opened 2017-04-01
[javascript,html,react,typescript] Use add-node-modules-path to automatically find local executables

jupl:
Currently it is a struggle to use some JS related libraries as I need to rely on globally installing Node packages (with possible conflicts) when I have project local packages available. React layer has a solution for the `eslint` package, which is nice. However, instead of patching all the JS related layers/libraries, I discovered `add-node-modules-path`.


layers/+lang/typescript/packages.el
```diff
@@ -19,6 +20,17 @@
         web-mode
         ))
 
+(unless (or (configuration-layer/layer-usedp 'javascript)
+            (configuration-layer/layer-usedp 'html))
+  (defun typescript/init-add-node-modules-path ()
+    (use-package add-node-modules-path :defer t)))
```
aaronjensen: Unless I'm missing something, it seems like if you use all of these layers together (or even 2 of them), nothing will define an init. I don't know if spacemacs has a way to deal w/ this (multiple layers vying to own a package) but if not, then you probably at least want to decide on a priority order for who can init it, meaning that if it's if the order is javascript->typescript->html, for example, that javascript always inits, typescript doesn't if javascript is used, and html doesn't unless neither javascript nor typescript is used.

jupl: The first package that gets loaded gets the init. I had to write it this way when I was testing different combinations of these layers in different orders in `dotspacemacs-configuration-layers`.

aaronjensen: That's surprising to me, I would think that configuration-layer/layer-usedp would return true for any layer that was declared and load order would be irrelevant. Otherwise there would be layer load order requirements and... I don't believe there are. plus, looking at the actual code:

```lisp
  (configuration-layer//declare-used-layers dotspacemacs-configuration-layers)
  (configuration-layer//declare-used-packages configuration-layer--used-layers)
  ;; then load the functions and finally configure the layers
  (configuration-layer//load-layers-files configuration-layer--used-layers
                                          '("funcs.el"))
```

it looks like `configuration-layer--used-layers` is defined before layer files are loaded.

@TheBB @syl20bnr could you clarify please?

jupl: You're free to checkout my branch and experiment with it. Just make sure if you try a different approach you try different combinations of the 3 in `dotspacemacs-configuration-layers`.

aaronjensen: I just did a simple test and it looks like you're right. That's very surprising.

aaronjensen: Ok, so I think the issue is that spacemacs loads packages.el (maybe while it's defining which layers are used) but doesn't actually invoke anything in it until after they've all been loaded. 

This means that `configuration-layer/layer-usedp` is order dependent if used while packages.el is initially loading as you are here. If you use it within the packages var or within functions in your packages.el, it will not be order dependent.

This makes me wonder if add-node-modules-path should be its own layer that these 3 layers can declare dependency on. Or maybe there's some other way of them working out who should own the package. Or maybe the way you have it is totally "fine".

jupl: Those are good points. I'll make any changes from feedback from the authors if needed.

TheBB: We should probably not rely on this behaviour. I think a better idea would be to make a layer called something like `web-common` that owns this package, and the javascript and html layers can depend on it.

It's also not necessary to guard your `add-hook` calls like that.

jupl: @TheBB Thanks. Will update later today. Where should this `web-common`/etc. layer go? Should it stay in `+lang`?

TheBB: Node is fine with me too, and in that case it can be argued it would fit under `+tools`

jupl: @TheBB Updated.

jupl:
Is there anything on my part that can be done to help get this merged? I've been using this for several months with no issues as of yet.

TheBB:
Fair enough, I'll trust you. :-P

TheBB:

PR #8625, opened 2017-04-01
common-lisp: Add slime-asdf to slime-contribs

dschoepe:
Some slime commands such as `,load-system` do not work without `slime-asdf` being loaded. Since ASDF is the main system definition facility used by lisp programs, I think this functionality should be available by default.


syl20bnr:
Thank you !

PR #8629, opened 2017-04-02
Unbreak diff-mode

Stebalien:
diff-mode is for *editing* diffs. Defaulting to evilified state and binding j/k breaks this.

You can still use TAB/S-TAB to navigate through hunks.

### DIFF
--- a/layers/+source-control/version-control/packages.el
+++ b/layers/+source-control/version-control/packages.el
@@ -24,11 +24,7 @@
 
 (defun version-control/init-diff-mode ()
   (use-package diff-mode
-    :defer t
-    :config
-    (evilified-state-evilify diff-mode diff-mode-map
-      "j" 'diff-hunk-next
-      "k" 'diff-hunk-prev)))
+    :defer t))
 
 (defun version-control/init-diff-hl ()
   (use-package diff-hl


TheBB:
This has been merged, apparently.

braham-snyder:
absolutely +1 on this, but does this incidentally cause `save-some-buffers` calls from `spacemacs/prompt-kill-emacs` to show its `view-mode` diffs with `"q"` bound to `evil-record-macro` instead of `View-quit`? or is this just me? FWIW, here's my workaround:

```
(use-package view
  :config

  ;; This version wouldn't bind until after pressing q once per new view-mode
  ;; buffer (at least in the context of `save-some-buffers' called from
  ;; `spacemacs/prompt-kill-emacs') -- the first q instead invokes
  ;; `evil-record-macro'... ???
  ;;
  ;;     (evil-define-key '(normal motion) view-mode-map (kbd "q") #'View-quit)
  ;;
  ;; Instead, just fix it buffer-locally for now.
  ;; TODO PR/raise issue for Spacemacs? previously "q" was bound to same or
  ;; similar function because `diff-mode' was (incorrectly) evilified, which was
  ;; fixed by:
  ;; https://github.com/syl20bnr/spacemacs/pull/8629
  (defun brds/View-quit-shadow-record-macro ()
    (evil-local-set-key 'normal (kbd "q") #'View-quit)
  (add-hook 'view-mode-hook #'brds/View-quit-shadow-record-macro)
```

PR #8632, opened 2017-04-03
Add the org-brain layer

Tritlo:
This pull request adds a spacemacs layer for the excellent `org-brain` package by @Kungsgeten, including documentation and some additional key bindings. Should this be a separate layer, or would this be better at home as a modification of the `org` layer?


yssource:
Awesome! Can't wait!

mrkgnao:
/cc @syl20bnr 
I couldn't think of anyone else to ping: just pointing out in case it's passed under the radar. This would be a great addition.

TheBB:

I moved it to the org layer and added a condition on Emacs 25.

PR #8635, opened 2017-04-05
Work around bug in eshell's preoutput-filter code.

Stebalien:
Eshell doesn't call preoutput-filter functions in the context of the eshell buffer. This breaks the xterm color filtering when the eshell buffer is updated when the eshell buffer is not currently focused.

I intend to submit this fix upstream when I get a chance but an advice fixes it for now. The other fix is to move the xterm color filter to a output-filter function (instead of preoutput-filter) but those operate on strings that have already been inserted into the buffer.

### DIFF
--- a/layers/+tools/shell/packages.el
+++ b/layers/+tools/shell/packages.el
@@ -83,6 +83,13 @@
       (add-hook 'eshell-mode-hook 'spacemacs/disable-hl-line-mode))
     :config
     (progn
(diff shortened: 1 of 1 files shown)


syl20bnr:
Thank you ! 👍 

kini:
The `define-advice` macro doesn't seem to exist in Emacs 24.  Does this work on Emacs 24.4, which according to the spacemacs README is the earliest currently supported Emacs version?

Stebalien:
@kini sigh... #8831

PR #8636, opened 2017-04-05
adds prodigy-view-clear-buffer leader key binding

FrancescElies:
Hello,

When I am in a buffer with `Prodigy-view` I would like to have a key binding to clear the buffer.

This makes `,c` to clear the the buffer by calling `prodigy-view-clear-buffer`.

Do you think this is useful?

Thanks in advance

### DIFF
--- a/layers/+tools/prodigy/packages.el
+++ b/layers/+tools/prodigy/packages.el
@@ -16,6 +16,7 @@
     :init
     (spacemacs/set-leader-keys "aS" 'prodigy)
     :config
+    (evil-define-key 'motion prodigy-view-mode-map "c" 'prodigy-view-clear-buffer)
     (evilified-state-evilify prodigy-mode prodigy-mode-map
       "h" 'prodigy-first
       "j" 'prodigy-next


bmag:
There's definitely no harm in adding the key-binding. However, the binding should be part of the `evilified-state-evilify` call instead of a separate `spacemacs/set-leader-keys-for-major-mode`, unless you have a good reason otherwise. (this way `c` is bound instead of `, c`, doesn't seem to me that `c` has any use in this buffer currently)

FrancescElies:
I thought in the future `prodigy-view-buffer` might become an editable buffer, that's why I set the binding to `,c`.

I tried to set the binding with `(evilified-state-evilify prodigy-view-mode prodigy-view-mode-map "c" 'prodigy-view-clear-buffer)` but did not work.

`prodigy-view-mode` buffers are read only, I guess this has influence on it.

This command though `(evil-define-key 'motion prodigy-view-mode-map "c" 'prodigy-view-clear-buffer)` sets correctly the binding to `c`, shall I use this one?

I think I need to read more about evilified state, I don't quite understand why the first attempt did not work.

All best & thanks for your reply

FrancescElies:
I see the motion state is used for read only buffers, I updated the PR to use `(evil-define-key 'motion prodigy-view-mode-map "c" 'prodigy-view-clear-buffer)` I hope is fine

FrancescElies:
Not sure why but failing `FAILED  test-spacemacs-html-export` just for `EVM_EMACS=emacs-24.5-travis` the rest of environments seem to be fine.

syl20bnr:
Thank you !
I moved the key binding to the macro instead: https://github.com/syl20bnr/spacemacs/commit/22e40f4f

PR #8643, opened 2017-04-06
rcirc: Fix using multiple .authinfo lines for authentication.

wordempire:
The auth-source-search function has an implied :max parameter with value 1. In
the case of connecting to multiple irc servers, only the first line in .authinfo
was applied. Thus, we should pass a value for :max to auth-source-search that is
guaranteed to at least be able to fit all our credentials.

Considerations: one might of course have irc-related lines in `.authinfo` which are not configured in spacemacs. In such a case, it is possible that some relevant `.authinfo` lines will never be seen by spacemacs. It would also be possible to replace the `(length rcirc-server-alist)` by a reasonably high number, or perhaps expose the value passed to `:max` in `auth-source-search` as a layer-specific variable.


syl20bnr:
Thank you !

PR #8647, opened 2017-04-06
evilify ensime search

d1egoaz:


syl20bnr:
I'm curious about this being active in insert mode only. I would like to test it for myself but I don't have a working setup for ENSIME.

Can someone make a screencast of the feature so I can see how it fits the rest of Spacemacs ?

diego-alvarez-hs:
@syl20bnr you're right, there was no reason to do that just on evil insert mode, I've updated to work on insert/normal mode.

This is how it works:
1. Scala/Java mode
2. Start Ensime
3. `, /` to go to ensime-search-mode
4. Find for a class/type/object
5. `C-i` to insert full type name at point
6. or `Enter` to go to type source code
7. `C-j` `C-k` to move down/up on ensime search results

j2wxjy

syl20bnr:
Thank you !
I just realized that the ensime layer does not use the evilified macro `evilified-state-evilify-map`, I guess it would be good to move to using them at some point instead of changing the evil keymaps directly.

d1egoaz:
That would be better. 
Thank you

PR #8657, opened 2017-04-07
Add opencl-mode to extra-langs

eklitzke:
This adds a very basic OpenCL layer, for editing OpenCL kernels. [OpenCL](https://en.wikipedia.org/wiki/OpenCL) is a framework for writing high performance kernels for CPUs, GPUs, FPGAs, and so forth.

I considered adding this to the existing "shaders" layer, which adds GLSL support, but this is sufficiently different that I thought it deserved its own layer. The GLSL stuff is for people interested in writing graphics shaders for GPUs, whereas the OpenCL stuff is for people interested in writing compute kernels on GPUs.

Please let me know if you want me to change anything, or to merge this functionality into an existing mode!


eklitzke:
Or should I just put this into `extra-langs`?

PR #8659, opened 2017-04-07
Add GPU layer

eklitzke:
This adds a new layer for doing GPU computations. Right now this just sets up `opencl-mode` for OpenCL (the open compute language), and `cuda-mode` for Nvidia's CUDA language. These languages are both in the same problem domain: they both target GPUs as computation devices, therefore it's likely that a user would want both installed at the same time. Additionally, both of these packages are very lightweight.


syl20bnr:
Thank you ! 👍 
I merged `gpu` and `shaders` layers, see commits.

PR #8660, opened 2017-04-07
alphabetize the init functions in extra-langs

eklitzke:
This just alphabetizes the order of the init functions for `extra-langs`, to be consistent with the order they're defined in `extra-langs-packages` (which is itself alphabetized).


syl20bnr:
Thank you !

PR #8661, opened 2017-04-07
Add automatic formatting on save for C++ using ClangFormat

robbert-vdh:
This adds a variable to enable automatic formatting using `clang-format` when saving a C/C++ buffer. Most language layers have similar functionality, but right now the recommended way to use ClangFormat is to create a user defined key binding on the tab key. I'm not really sure if this is the proper way to go about doing this, so please correct me if there's a better way.

### DIFF
--- a/layers/+lang/c-c++/packages.el
+++ b/layers/+lang/c-c++/packages.el
@@ -61,7 +61,11 @@
 
 (defun c-c++/init-clang-format ()
   (use-package clang-format
-    :if c-c++-enable-clang-support))
+    :if c-c++-enable-clang-support
+    :init
+    (when c-c++-enable-clang-format-on-save
+      (spacemacs/add-to-hooks 'clang-format/format-on-save
+                              '(c-mode-hook c++-mode-hook)))))
 
 (defun c-c++/init-cmake-mode ()
   (use-package cmake-mode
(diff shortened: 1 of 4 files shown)


eklitzke:
I just happened to be browsing spacemacs pull requests, and I want this very much. I have similar logic in my `dotspacemacs/user-config`, and it would be great to have this in Spacemacs by default.

layers/+lang/c-c++/packages.el
```diff
@@ -61,7 +61,10 @@
 
 (defun c-c++/init-clang-format ()
   (use-package clang-format
-    :if c-c++-enable-clang-support))
+    :if c-c++-enable-clang-support
+    :init
+    (when c-c++-enable-clang-format-on-save
+      (add-hook 'c-mode-common-hook 'clang-format/format-on-save))))
```
dubnde: Should this add hook only for c/c++ mode as follows inline with rest of layer?

`(spacemacs/add-to-hooks 'clang-format/format-on-save '(c-mode-hook c++-mode-hook))`

robbert-vdh: I'd misunderstood `c-mode-common-hook`, as it also seems to run for Java and other C-like languages. Besides, consistency is always good. I'll change it in a minute!

syl20bnr:
Thank you ! 👍 
I modified it a bit to allow usage of dir-local variable: https://github.com/syl20bnr/spacemacs/commit/cc61e08b

PR #8673, opened 2017-04-10
Documents conflict resolution between ctags and etags

### LINKED ISSUE #8648: Instructions for solving conflict between ctags and etags
This is a problem I had to deal with and was quite frustrating.

When installing emacs from source, you also get a binary named `ctags` that is actually emacs's `etags` which replaces `ctags` inside `/usr/local/bin`. This breaks the `gtags` layer since their interface is different.

What I propose is a notice in the gtags layer `Readme.org` (maybe section 3.1) that you need to configure the emacs build script to change the name of the `ctags` binary it produces.

Here is an article that describes the fix: https://www.topbug.net/blog/2016/11/10/installing-emacs-from-source-avoid-the-conflict-of-ctags/

What I have in mind is something like:

---
3.1.3 Conflict between `ctags` and emacs's `etags`

If you installed `emacs` from source after `ctags`, your original `ctags` binary is probably replaced by emacs's `etags`. To get around this you will need to configure emacs as following before installing:
```
./configure --program-transform-name='s/^ctags$/ctags.emacs/'
```

To check if you have the correct version of ctags execute:
```
ctags --version | grep Exuberant
```
If yo do not get any output then it means you have the wrong `ctags` and must install it again.

---

I am ready to submit a PR if the idea is 👍

sdwolf:
Closes #8648.

### DIFF
--- a/layers/+tags/gtags/README.org
+++ b/layers/+tags/gtags/README.org
@@ -10,6 +10,7 @@
        - [[#install-recommended-dependencies][Install recommended dependencies]]
        - [[#install-with-recommended-features][Install with recommended features]]
        - [[#configure-your-environment-to-use-pygments-and-ctags][Configure your environment to use pygments and ctags]]
+     - [[#conflict-between-ctags-and-emacss-etags-binary][Conflict between =ctags= and emacs's =etags= binary]]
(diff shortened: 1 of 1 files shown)


syl20bnr:
Thank you ! 👍 

PR #8682, opened 2017-04-11
Add Habitica support for org mode

ksonney:
Cleaner merge request for adding habitica support to org-mode with habitrpg.el.

Key mappings work, the readme is fairly accurate, and you can get to the status screen with SPC-a-H to show your tasks and rewards and all that.


layers/+tools/habitrpg/packages.el
```diff
@@ -0,0 +1,29 @@
...
+
+(defconst habitrpg-packages
+  '(
+    (habitrpg :location (recipe
+                         :fetcher github
+                         :repo "ryjm/habitrpg.el"))))
+
+(defun habitrpg/init-habitrpg ()
+  (require 'org-habit)
+  (require 'habitrpg))
```
TheBB: Please change this to use lazy loading.

layers/+tools/habitrpg/keybindings.el
```diff
@@ -0,0 +1,3 @@
+;; Key Bindings for habitrpg.el
+(spacemacs/set-leader-keys "aH" 'habitrpg-status)
+(add-hook 'org-after-todo-state-change-hook 'habitrpg-add 'append)
```
TheBB: These should be in `packages.el` as part of the init function to which they belong.

layers/+tools/habitrpg/packages.el
```diff
@@ -0,0 +1,29 @@
...
+;; - If PACKAGE is not referenced by any other Spacemacs layer, define a
+;;   function `habitrpg/init-PACKAGE' to load and initialize the package.
+
+;;; Code:
+
+(defconst habitrpg-packages
+  '(
+    (habitrpg :location (recipe
+                         :fetcher github
+                         :repo "ryjm/habitrpg.el"))))
```
TheBB: What's the problem with https://github.com/abrochard/emacs-habitica? That package is on MELPA.

ksonney: That particular package doesn't support org-mode, and this one does.

ksonney: @TheBB until I can convince the upstream maintainer to  submit the package to MELPA, is that an absolute blocker? As noted above, the linked package does not support org-mode, nor is it as feature complete.

PR #8685, opened 2017-04-11
Add escape and M-SPC bindings to ivy

Somelauw:
Add escape and M-SPC bindings to ivy
- escape quits the minibuffer
- M-SPC enters transient state

See conventions at https://github.com/syl20bnr/spacemacs/blob/master/doc/CONVENTIONS.org

#### Issues to be resolved ####
- [x] Pressing escape in hydra-ivy exits minibuffer, but doesn't exit transient state
To reproduce press M-SPC, escape.

### DIFF
--- a/layers/+completion/ivy/packages.el
+++ b/layers/+completion/ivy/packages.el
@@ -164,6 +164,11 @@
        'counsel-recentf
        spacemacs--ivy-file-actions)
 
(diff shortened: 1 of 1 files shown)


syl20bnr:
Thank you ! 💜 
We need to make a README for the Ivy layer with all the bindings.

PR #7536, opened 2016-10-24
Added hide-others shortcut to osx layer

basveeling:
In emacs 25.1, the osx system default shortcut cmd+alt+h to hide other windows is broken. This commit fixes this (see #7464).

PR #8689, opened 2017-04-12
[enhancement] OSX Layer. change osx-command-as default for ⌘ to hyper instead of super

aculich:
This PR builds on #8526 by @Chream

The purpose of this PR is to set the default for ⌘ to `hyper` and to provide the
function `kbd-mac-command` to replace `kbd` when defining keybindings for ⌘ in
case someone decides to explicitly set osx-command-as to `super` or `alt`.

There are problems setting osx-command-as to `alt` and `super`,
so we use `hyper` as a default instead because, for example:
  - Using `alt`:   Command-x or Command-m inserts, respectively: × µ
  - Using `super`: Control-Command-f produces keycode: `<C-s-268632078>`
Setting to `hyper` seems to avoid both types of the above problems.
Also, while it is possible, it is not recommended to set to `meta`
since standard OSX shortcuts would overshadow important keys such
as M-x.

Two other small changes include:

  - Commenting out the code that defines `<C-s-268632078>` (C-s-f) since it is
    unnecessary if we use `hyper` as the default; and if we really want to use
    `super` then we should figure out how to solve the weird keycode issue.

  - add keybinding for ⌘\` (Command-backtick) to `other-window`.
    Emacs usually swallows this keystroke, so other-window basically restores
    the default behavior that most Mac OSX users would expect.

### DIFF
--- a/layers/+os/osx/README.org
+++ b/layers/+os/osx/README.org
@@ -15,7 +15,7 @@
 
 * Description
 Spacemacs is not just emacs+vim. It can have OSX keybindings too! This layer
-globally defines common OSX keybindings. ~⌘~ is set to ~super~ and ~⌥~ is set to
+globally defines common OSX keybindings. ~⌘~ is set to ~hyper~ and ~⌥~ is set to
 ~meta~. Aside from that, there's nothing much, really.
 
 While in =dired= this layer will try to use =gls= instead of =ls=.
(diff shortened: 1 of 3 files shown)


syl20bnr:
Thank you ! 👍 
I also merged two other PRs regarding OSX key bindings, I think they should be updated with this PR:
- https://github.com/syl20bnr/spacemacs/pull/8878
- https://github.com/syl20bnr/spacemacs/pull/7536

Can you look at them ?

PR #8695, opened 2017-04-12
Diminish the YAPF minor mode indicator

robbert-vdh:
Some default minor modes are still showing their full indicators, this adds an indicator for yapfify in the Python layer.


syl20bnr:
Thank you !
Circled unicode symbols are reserved for toggles on `SPC t` and `SPC T`. I removed the YAPF lighter completely since it is an opt-in variable, users should be aware that it is enabled.

robbert-vdh:
That makes more sense, thanks for you time!

PR #8699, opened 2017-04-13
README.md: fixed wrong flag for macOS

metalelf0:
In the macOS installation guide the build flag for the `emacs-plus` package is spelled `--with-natural-title-bars` instead of `--with-natural-title-bar`. The correct one is the latter.


syl20bnr:
Thank you ! The change was already in develop branch ;-)

PR #8707, opened 2017-04-13
A perl5 layer with material from perlnow and bioperl.

abelew:
a perl5 layer, theoretically my perl5 branch against your develop branch.


syl20bnr:
That's a lot of files, any possibility to use MELPA or fetch the files from somewhere else ? We don't put complex 3rd party source files in Spacemacs.

TheBB:
Closing this. Sorry, but this is far too complicated. We have a simpler perl5 layer now, please consider improving that instead. :-)

PR #8708, opened 2017-04-13
Update doc/LAYERS.org: Fix 2 Use-package titles/links

duianto:
```
problem:  The TOC link: Layers tips and tricks > Use-package
          navigates to: The Emacs loading process > Use-package
solution: Rename: Layers tips and tricks > Use-package
          to: Layers tips and tricks > Use-package init and config
```
The issue occurs in Emacs `SPC h r` `Tips on writing layers for Spacemacs` and on both of the pages: http://spacemacs.org/doc/LAYERS.html and http://develop.spacemacs.org/doc/LAYERS.html

But it works as expected here: https://github.com/syl20bnr/spacemacs/blob/develop/doc/LAYERS.org
This is just a guess, but maybe githubs org parser (is that the right word?), reads the id `#use-package-1` from `doc/LAYERS.org`'s TOC, and adds it as a href link to the second occurrence of the "Use package" heading: `Layers tips and tricks` > `Use package`.

This almost resolves syl20bnr/develop.spacemacs.org#10 , the updated doc/LAYERS.org just needs to generate a new doc/LAYERS.html

### DIFF
--- a/doc/LAYERS.org
+++ b/doc/LAYERS.org
@@ -21,7 +21,7 @@
  - [[#case-study-auto-completion][Case study: auto-completion]]
  - [[#layer-tips-and-tricks][Layer tips and tricks]]
    - [[#cross-dependencies][Cross-dependencies]]
-   - [[#use-package-1][Use-package]]
+   - [[#use-package-init-and-config][Use-package init and config]]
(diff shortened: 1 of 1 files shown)


syl20bnr:
Thank you ! 👍 

PR #8712, opened 2017-04-14
Update new-empty-buffer: split argument

duianto:
Updated the `spacemacs/new-empty-buffer` function, to accept a `split`
argument, that can have 4 values: `left`, `below`, `above` or `right`.

Added new functions and key bindings (`SPC b M-h`, -j, -k and -l) for
each direction.

#### Questions:
1. Function names
Current: `spacemacs/new-empty-buffer-left`, `-below`, etc.
Would it be better to include `split` in the function names: `spacemacs/new-empty-buffer-split-left`?

2. Key bindings
Current: `SPC b M-h`, `-j`, `-k` and `-l`.
Should they remain unchanged? Or would it be better to use `SPC b C-h`, `-j`, etc.?
I might be missing better key bindings, suggestions are welcome.

3. Buffer transient state `SPC b .`
Which of these three options should be chosen?
First: No `spacemacs/new-empty-buffer-(split-)-left`, etc., keys in the buffer TS.
Second: Keys in both the `SPC b` buffer which-key popup and in the buffer TS.
Third: Only keys in the buffer TS, that would make the `h`, `j`, `k` and `l` keys available.

4. Option to keep the cursor in the pre-split window
Would it be useful to have functions and key bindings that don't move the cursor from the active window? The key bindings could use the uppercase versions `M-H`, `M-J`, etc. If this is a useful feature, then the "default" function names in Q1 above, probably should be renamed to something like: `spacemacs/new-empty-buffer-(split-)left-and-switch`.

### DIFF
--- a/layers/+distributions/spacemacs-base/funcs.el
+++ b/layers/+distributions/spacemacs-base/funcs.el
@@ -491,11 +491,43 @@ If the universal prefix argument is used then will the windows too."
   (ediff-files (dotspacemacs/location)
                (concat dotspacemacs-template-directory ".spacemacs.template")))
 
-(defun spacemacs/new-empty-buffer ()
-  "Create a new buffer called untitled(<n>)"
+(defun spacemacs/new-empty-buffer (&optional split)
+  "Create a new buffer called untitled(<n>).
+A SPLIT argument with the value: `left',
+`below', `above' or `right', opens the new
+buffer in a split window."
   (interactive)
   (let ((newbuf (generate-new-buffer-name "untitled")))
+    (case split
+      ('left  (split-window-horizontally))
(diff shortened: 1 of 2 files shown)


syl20bnr:
Thank you ! 👍 
I moved all the key bindings for new empty buffer under ` SPC b N`, see commit: https://github.com/syl20bnr/spacemacs/commit/ff0bd88b

bmag:
I updated `spacemacs/new-empty-buffer` to pass a non-nil `force-same-window` argument to `switch-to-buffer`. This guarantees that `switch-to-buffer` displays the buffer in the new window. See https://github.com/syl20bnr/spacemacs/commit/4b92183cc0dd7fb1e0ec4291d341d457cccc1afc

PR #8716, opened 2017-04-14
[elixir] Update fn names - update to new names and remove extinct fns

swaroopch:
Some functions no longer existed, and some had newer names.

I updated the tex file but couldn't find instructions to regenerate the PDF.


syl20bnr:
Thank you !
I'm not sure how to update the PDF either...

PR #8718, opened 2017-04-15
Core: add support for building multi-file local packages with Quelpa

madand:
At the moment a Quelpa recipe like `(recipe :fetcher local)` is being translated to something like `(recipe :fetcher file :path "my-layer/local/my-pkg/my-pkg.el")`. So we can build simple single-file local packages.

This commit changes it to translate to the package directory instead of exact lisp file, so we can build multi-file local packages. Thus, the above example will be translated to `(recipe :fetcher file :path "my-layer/local/my-pkg")`.

Also, add the relevant info to LAYERS.org.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -633,10 +633,10 @@ If TOGGLEP is nil then `:toggle' parameter is ignored."
                (eq (plist-get (cdr location) :fetcher) 'local))
           (cond
            (layer (let ((path (expand-file-name
-                               (format "%s%s/%s.el"
+                               (format "%s%s"
                                        (configuration-layer/get-layer-local-dir
                                         layer-name)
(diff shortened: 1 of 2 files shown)


syl20bnr:
This is nice. I believe we don't need to define another keyword though, if we always translate to the `my-pkg` directory instead of `my-pkg.el` file it should work for all cases right ?

madand:
@syl20bnr yes, you are right. I overlooked the fact that Spacemacs' policy of having a directory per each local package eliminates the need for special handling of single-file packages. Thanks for stopping the unneeded complexity from going into the core.

I rewrote the commit, the documentation and edited the OP.
I see that 1 Travis check failed, but the error occurs in the place not changed by this PR, dunno what causing this.

syl20bnr:
Thank you ! 👍 

PR #8727, opened 2017-04-17
Haskell: use intero-goto-definition for jump handler if intero-mode is enabled.

MaskRay:
Haskell: use intero-goto-definition for jump handler if intero-mode is enabled.


syl20bnr:
Thank you ! 👍 

PR #8730, opened 2017-04-17
In M-x shell, set tab-width to 8 spaces

gregory-nisbet:
Certain programs like `ls` expect a `tab-width` of 8 characters. Spacemacs sets the default tab-width to 2. (https://github.com/syl20bnr/spacemacs/blob/bd7ef98e4c35fd87538dd2a81356cc83f5fd02f3/layers/%2Bdistributions/spacemacs-base/config.el#L79)

This is most noticeable under something like `M-x shell`, where `ls` will have the wrong formatting.

I think the cleanest solution would probably be to *not* set the tab-width to 2 in `spacemacs-base/config.el` and instead override it selectively for modes where it makes sense to do that.

### DIFF
--- a/init.el
+++ b/init.el
@@ -32,3 +32,11 @@
   (spacemacs/setup-startup-hook)
   (require 'server)
   (unless (server-running-p) (server-start)))
+
+(defun spacemacs//interactive-shell-set-tab-width ()
+  "set the tab width to 8 for interactive shell"
+  (setq tab-width 8))
+
+(advice-add 'shell :after
+            (lambda (&rest _)
+              (spacemacs//interactive-shell-set-tab-width)))
--- a/init.el
+++ b/init.el
@@ -33,10 +33,4 @@
   (require 'server)
   (unless (server-running-p) (server-start)))
 
-(defun spacemacs//interactive-shell-set-tab-width ()
-  "set the tab width to 8 for interactive shell"
-  (setq tab-width 8))
-
-(advice-add 'shell :after
-            (lambda (&rest _)
-              (spacemacs//interactive-shell-set-tab-width)))
+(advice-add 'shell :after (lambda (&rest _) (shell/init-shell)))
--- a/layers/+tools/shell/packages.el
+++ b/layers/+tools/shell/packages.el
@@ -181,6 +181,11 @@
 
 (defun shell/init-shell ()
   (spacemacs/register-repl 'shell 'shell)
+
+  (defun shell-interactive-shell-set-tab-width ()
+    "set the tab width to 8 for interactive shell"
+    (setq tab-width 8))
+
   (defun shell-comint-input-sender-hook ()
     "Check certain shell commands.
  Executes the appropriate behavior for certain commands."
@@ -202,7 +207,13 @@
              ;; Send other commands to the default handler.
              (t (comint-simple-send proc command))))))
   (add-hook 'shell-mode-hook 'shell-comint-input-sender-hook)
-  (add-hook 'shell-mode-hook 'spacemacs/disable-hl-line-mode))
+  (add-hook 'shell-mode-hook 'spacemacs/disable-hl-line-mode)
+
+  (advice-add 'shell :after
+              (lambda (&rest _)
+                (shell-interactive-shell-set-tab-width))))
+
+
 
 (defun shell/init-shell-pop ()
   (use-package shell-pop
--- a/layers/+tools/shell/packages.el
+++ b/layers/+tools/shell/packages.el
@@ -213,8 +213,6 @@
               (lambda (&rest _)
                 (shell-interactive-shell-set-tab-width))))
 
-
-
 (defun shell/init-shell-pop ()
   (use-package shell-pop
     :defer t


layers/+distributions/spacemacs-base/config.el
```diff
@@ -76,7 +76,7 @@ It runs `tabulated-list-revert-hook', then calls `tabulated-list-print'."
 
 ;; use only spaces and no tabs
 (setq-default indent-tabs-mode nil
-              tab-width 2)
```
robbyoconnor: NO! Do this in your own config.

robbyoconnor:
Thank you for your contribution, however why does this need to be done globally?

gregory-nisbet:
@robbyoconnor 

I'm not trying to be rude or to pollute the project with a personal notion of how many spaces a tab should be. I glanced through a few of the layers, and many of them seem to have their own independent variable for controlling the width of a tab... 

The problem that I'm encountering is that there are bits of functionality in emacs (like `M-x shell` that are "lower-level" than spacemacs and seem in some ways to "depend" on a tab being 8 spaces (depend might be too strong of a word)).

An example alternative (that seems even worse to me) is here: https://github.com/syl20bnr/spacemacs/pull/8730/commits/eafecd89eb23362ea6fabd1ca5ec37f83ae8a638

I've tried setting the `tab-width` to `8` in a few places in order to get the desired behavior in `M-x shell` and none of them seem to work.

I'd like `M-x shell` to look right out of the box. It took about a while to figure out how to get the behavior to go away, since add `(setq tab-width 8)` stuff to `dotspacemacs/user-init` and `dotspacemacs/user-config`  has no effect. Also, using `advice` as non-[end user] seems a bit, uh, non-ideal?

Idk.

robbyoconnor:
That came off far more abrasive than I intended -- sorry about that.. but yes tab width seems to be major-mode dependent..I'm just against this being here...at the very least allow it to be configurable.. don't just hard-code.

gregory-nisbet:
@robbyoconnor 

The major mode for `M-x shell` is just `shell-mode` which vanilla Emacs also uses for shell scripts. Spacemacs does not modify this mode. Spacemacs does have a layer for shell scripts, but that doesn't seem to get activated for interactive shells, somewhat confusingly.

The possible solutions I can think of are:

- Use the vanilla emacs default `tab-width` and selectively override it. (this idea has been rejected)
- advise `shell` and put the advice in some file somewhere in the Spacemacs source tree.
- add another hook to `shell-mode` and put that somewhere.
- add documentation in an FAQ or something explaining why `M-x shell` then `ls` looks weird, say that it's intentional, and maybe have a code snippet people can copy-paste.

jupl:
I use `editorconfig` which is nice because I have a file in `$HOME` which applies default settings to my files. https://github.com/editorconfig/editorconfig-emacs

gregory-nisbet:
@robbyoconnor 

Okay, this should be better... I guess. The change is pretty small and could really go anywhere.

I tried placing it in `layers/+distributions/spacemacs-base/funcs.el` and in the main `init.el`, but those caused one of the CI build+tests to fail. Not sure why, still trying to figure it out.

robbyoconnor:
Remember to rebase/squash when done.

gregory-nisbet:
done and cleaned up. Solves the immediate problem (setting the `tab-width` to 8 when starting an `M-x` shell without interfering with anything else.)

Also, `init.el` is not a fantastic spot for this, but I'm not sure a more appropriate place would be.

gregory-nisbet:
I think it doesn't happen with ansi-term because ansi-term is emulating a tty and its buffer doesn't contain literal `0x09`s.

gregory-nisbet:
I could move this into a layer maybe? I'm not sure where the possible homes for it are.

The first place I tried was the existing the `shell-script` layer, but that mode doesn't trigger unless you're looking at a shell script ... and not just any time that `shell-mode` is activated.

robbyoconnor:
Shell layer?

robbyoconnor:
Hmm...I mean ultimately...why not use ansi-term?

gregory-nisbet:
`M-x shell` is lightweight relative to ansi-term and doesn't emulate a terminal, so you can select a region and get the same bytes that the program wrote to stdout/stderr. It's also more responsive and lets you use the emacs for line editing rather than readline or libedit or what have you.

nixmaniack:
`init.el` isn't the right place for this configuration. Also I don't think advising is correct way to solve it. You already mentioned about hook approach which should be about right. 

Create a function and add hook. Relevant place to make these changes would be: https://github.com/syl20bnr/spacemacs/blob/f974b4c9b219f57f49c332c0ad9246ab67c2774e/layers/+tools/shell/packages.el#L182-L205.

If the value of the `tab-width` could be different as per user preference for this layer - then a layer variable should be provided.

gregory-nisbet:
@nixmaniack I tried to move as much as I could out of `init.el` ... unfortunately it doesn't seem to trigger unless the `shell` configuration layer is active ... and I still need some advice in `init.el`... grr.

nixmaniack:
@gregory-nisbet I'm not sure what exactly are you doing. But here's what I tried and it seemed to have worked.

```elisp
(defun my/set-shell-tab-width ()
  (setq-local tab-width 8))

(add-hook 'shell-mode-hook 'my/set-shell-tab-width)
```

Evaluating `tab-width` in `shell` buffer gives `8` before this it was `2`.

Also, I fail to reproduce your issue. I don't have any scrambled text alignment issues in my `M-x shell` even though I'm using default `tab-width` set by Spacemacs. Did you check whether your terminal escape sequences are messing around?

syl20bnr:
@gregory-nisbet Thank you for your PR. Indentation in Emacs is a pain point and it is not really obvious to setup indentation for a mode since each mode has its own variable(s).

For now I consider indentation as a user preference and so this PR falls into this category.

PR #8732, opened 2017-04-18
Remove trailing whitespace (again)

robbyoconnor:
This snuck in -- please type `SPC x d w` to prevent this!


robbyoconnor:
@syl20bnr, @bmag, @ThEBB, @d12frosted

robbyoconnor:
CI failing isn't my fault.

JAremko:
@robbyoconnor  hmm 24.5 keep on failing but 24.4 works.

JAremko:
@deb0ch  :roll_eyes:

robbyoconnor:
Again not my problem -- I literally changed org/markdown files

JAremko:
@deb0ch  :eyes:

robbyoconnor:
I use emacs 26

JAremko:
This https://github.com/syl20bnr/spacemacs/commit/62167f708d0d244ca98fe9e578f32b13434ca0c4 might be the reason.

deb0ch:
@JAremko I didn't really work on the subject, just reported a bug the the `html-export` repo (forgot the name) and the guy considered it "not a bug".

Here it is: https://github.com/snosov1/toc-org/issues/35

deb0ch:
Oh, I also reported an issue in Spacemacs concerning org tags showing in the docs: https://github.com/syl20bnr/spacemacs/issues/8216

smile13241324:
ups sorry didn't notice that, sorry for the inconvenience. Was my test killing the CI builds?

JAremko:
@smile13241324  I looks like the publishing function doesn't like https://github.com/syl20bnr/spacemacs/blob/develop/CONTRIBUTING.org and only with emacs 24.5  :thinking:

smile13241324:
Sounds interesting, I'll have a look this evening.

JAremko:
@smile13241324  I think I'll move doc test into separate test suite.  It takes too long and even package installation process is too verbose +  there is no reason to export all files and do it for each commit and each Emacs version.

smile13241324:
@JAremko you may be right, that test is actually not reliable enough if this fails randomly on certain emacs versions. Do we have a place to store tests we want to run once in a while but not for each and every commit?

JAremko:
@smile13241324  My current solution https://github.com/syl20bnr/spacemacs/pull/8736 also I'll try to make URL check work if possible (by retrieving only http headers)

smile13241324:
@JAremko I like your solution, this would still run the doc tests for every commit but we avoid those random problems with older emacs versions. Having emacs only retrieve the http headers would be very interesting, I have also tried to get that running but did not find a good elisp function to do so.

JAremko:
@smile13241324  We probably can exclude this test if no *.ORG files have changed. If it will run too long (in comparison to other tests)

JAremko:
Apparently Emacs 24.5 can't work without `CUSTOM_ID` but earlier versions and later ones are ok. I'll try to fix it. Also given how random it is, we probably need to run export test with all versions :smile:

robbyoconnor:
Can we move this discussion elsewhere? Thanks!

JAremko:
@robbyoconnor  discussion is over. We won :shipit:

syl20bnr:
Thank you ! :-D

PR #8747, opened 2017-04-20
Add spacefmt test to CI

JAremko:
Adding extra CI test that will offload some easy tasks to PR authors

### DIFF
--- a/.travis-build.sh
+++ b/.travis-build.sh
@@ -19,6 +19,7 @@ if [ $USER != "travis" ]; then
     exit 1
 fi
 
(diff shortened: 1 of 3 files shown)


JAremko:
@syl20bnr Hopefully nothing broken and hopefully after adding this test we won't need to run `spacefmt` anymore.   Well.. you'll need to run it once to fix the old bugs - the test only checks changed files. :smile:

syl20bnr:
Nice addition as always 💜 
Thank you !

PR #8753, opened 2017-04-20
Allow customizing if ENSIME should print types at point

jcazevedo:
According to the [documentation](https://github.com/syl20bnr/spacemacs/blob/dd8508482629c042718681c3fa2df07541a37028/layers/%2Blang/scala/README.org), the `scala-enable-eldoc` variable should be used to determine whether or not ENSIME should print types at point. This makes sure that the `spacemacs//java-setup-ensime-eldoc` hook is only added to `scala-mode-hook` if the variable is non-`nil`.

### DIFF
--- a/layers/+lang/scala/packages.el
+++ b/layers/+lang/scala/packages.el
@@ -24,7 +24,8 @@
     ))
 
 (defun scala/post-init-eldoc ()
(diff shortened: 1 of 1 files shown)


diego-alvarez-hs:
Please merge this @syl20bnr 
The latest java/scala ensime refactor broke this.
Thanks @jcazevedo

syl20bnr:
Done :-)
Thank you !

PR #8760, opened 2017-04-21
Fix ID collisions in html publish

JAremko:
Fixes syl20bnr/develop.spacemacs.org#10


JAremko:
aggregated into #8777

PR #8765, opened 2017-04-21
format spacefmt CI message

JAremko:
It was easy to overlook


JAremko:
@syl20bnr  I set `spacefmt` test to check only org files

JAremko:
aggregated into #8777

PR #8766, opened 2017-04-21
Add various slack keybindings

benjamreynolds:
The only readily available keybinding for joining channels in slack is for
joining public channels. This commit adds bindings for joining private
channels, (called groups in emacs-slack) as well as a keybinding for
slack-select-rooms which allows joining a channel, group, or DM.

This commit also rebinds the major mode `k` to slack-select-rooms, which puts it
more in line with how the related keybinding in the slack client works.


syl20bnr:
Thank you ! 👍 

PR #8773, opened 2017-04-22
Update README.md: links/rewrites/missing words/etc

duianto:
I also deleted the first empty line after the headers, as suggested in the conventions here: https://github.com/syl20bnr/spacemacs/blob/master/doc/CONVENTIONS.org#spacing-in-documentation

Should the mention of Ubuntu LTS 12.04 be removed? It's end of life date is april 28, 2017.
source: https://wiki.ubuntu.com/Releases

After reading this line:
>If the mode-line turns red then be sure to consult the FAQ.

I tried to find the direct link to that issue, but I was unable to find any reference to `mode-line`, `modeline` or `red` (at least not the color) in `FAQ.org`. 
So I'm not sure if that line should remain or be removed?

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,11 +3,11 @@
 ***
 <p align="center"><img src="/doc/img/title2.png" alt="Spacemacs"/></p>
 <p align="center">
-<b><a href="http://spacemacs.org/doc/DOCUMENTATION#orgheadline5">philosophy</a></b>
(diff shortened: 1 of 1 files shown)


syl20bnr:
Thank you ! 👍 
Nice ! I made some tweaks in commit https://github.com/syl20bnr/spacemacs/commit/d25d032e see commit message for explanation.

PR #8777, opened 2017-04-23
CI improvement DLC [RDY]

JAremko:
Automatically publish docs to develop.spacemacs.org on push to master via PR from unprivileged account.

*Example: https://github.com/syl20bnr/develop.spacemacs.org/pull/15*

PR #8787, opened 2017-04-24
remove automatic call to linum-setup when using git-gutter

gilbertw1:
This pull request removes the automatic invocation of ```git-gutter:linum-setup``` when using ```git-gutter``` with linum mode is enabled globally. As far as I can tell, this fundamentally breaks line numbers in spacemacs resulting in many duplicated line numbers. 

Additionally, ```git-gutter-fringe``` is automatically included and results in the icons being placed in two locations, the fringe and in the line numbers. As it stands there is no way to prevent linum-setup from being called and simply use git-gutter in the fringe.

I chose to remove the call entirely instead of adding an additional layer configuration variable to hide this call because the functionality appears fundamentally broken in spacemacs and an additional configuration value would be specific to ```git-gutter``` and I believe would only serve to confuse users.

### DIFF
--- a/layers/+source-control/version-control/packages.el
+++ b/layers/+source-control/version-control/packages.el
@@ -59,9 +59,6 @@
       (when (and (eq version-control-diff-tool 'git-gutter)
                  version-control-global-margin)
         (global-git-gutter-mode t))
-      ;; If you would like to use git-gutter.el and linum-mode
-      (if dotspacemacs-line-numbers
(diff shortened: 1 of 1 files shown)


JAremko:
@geksilla  ignore CI it's messed up.  Waiting for https://github.com/syl20bnr/spacemacs/pull/8777

syl20bnr:
Thank you !
Let's remove it and see what other users think.

Issue #5395, opened 2016-03-07
ESS layer help.start makes inferior ess read-only


izahn:
#### Description

Starting an R repl and opening the help system in a browser makes the inferior-ess buffer read-only.
#### Reproduction guide
- Start Emacs
- `SPC a'` `R <ret>` to start an R repl (requires ess layer)
- At the R prompt type `help.start()` to launch R help in a browser
- Click on a link in the browser, e.g., "Packages".

_Observed behaviour:_
In the inferior ess buffer I see

```
> help.start()
starting httpd help server ... done
If the browser launched by '/usr/bin/xdg-open' is already running, it
    is *not* restarted, and you must switch to its window.
Otherwise, be patient ...
> 
Making 'packages.html' ... done
```

(which is expected), but further input is impossible. Any attempt to type in the inferior ess buffer results in `Text is read-only` being echoed in the minibuffer.

_Expected behaviour:_
The R repl should be usable after starting the help system in R.
#### System Info
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.13
- Spacemacs branch: develop (rev. 895455d)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: ivy
- Layers:

``` elisp
(spacemacs-ivy auto-completion command-log emacs-lisp ess octave python sql latex git markdown org pdf-tools pandoc
               (shell :variables shell-default-height 35 shell-default-position 'bottom shell-default-shell 'multi-term)
               (mu4e :variables mu4e-headers-include-related t mu4e-headers-skip-duplicates t mu4e-headers-fields
                     '((:human-date . 12)
                       (:flags . 6)
                       (:mailing-list . 10)
                       (:from-or-to . 22)
                       (:thread-subject))))
(spacemacs-ivy auto-completion command-log emacs-lisp ess octave python sql latex git markdown org pdf-tools pandoc
               (shell :variables shell-default-height 35 shell-default-position 'bottom shell-default-shell 'multi-term)
               (mu4e :variables mu4e-headers-include-related t mu4e-headers-skip-duplicates t mu4e-headers-fields
                     '((:human-date . 12)
                       (:flags . 6)
                       (:mailing-list . 10)
                       (:from-or-to . 22)
                       (:thread-subject))))
```
#### Backtrace

No error is triggered, so no backtrace.

izahn:
I can also reproduce this on the latest released version of spacemacs, i.e.,

#### System Info
[...]

PR #8790, opened 2017-04-25
Solves #5395 "ESS layer help.start makes inferior ess read-only"

jackkamm:
This solves https://github.com/syl20bnr/spacemacs/issues/5395, which reported that calling `help.start()` in the R repl causes it to become read-only and unusable. The same issue causes a `shinyApp()` to be unkillable as I reported in the comments, and possibly causes other issues in the R repl. The issue is caused by `(setq comint-prompt-read-only t)` in the shell layer, but can be solved by adding the hooks from https://github.com/emacs-ess/ESS/issues/300#issuecomment-231314374


JAremko:
@jackkamm  ignore CI it's messed up.  Waiting for https://github.com/syl20bnr/spacemacs/pull/8777

syl20bnr:
Thank you ! 👍 

Issue #5910, opened 2016-04-27
Feature request: support realgud


jluttine:
Would it be possible/reasonable to add support for realgud?

https://www.emacswiki.org/emacs/DebuggingWithEmacs#toc3

https://github.com/rocky/emacs-dbgr

I guess it's a better debugging front-end for interacting with external debuggers than built-in(?) GUD.

jluttine:
Realgud would probably add support for ipdb (see #5907) among other things.

bmag:
I haven't tried realgud, but it sounds really cool :+1:

nixmaniack:
It's really nice, I have had it before Spacemacs, seems I lost it in migration to Spacemacs!

jluttine:
I'm trying to write a layer for realgud (my first layer ever). I'll write about my progress here so others can help/comment.

I wrote a simple (private) layer which uses realgud package. After activating the layer and restarting spacemacs, I got a loooong list of compiling errors, similar to those listed here: https://github.com/rocky/emacs-dbgr/issues/76

Is this something I should be worried about and should it be fixed or is it something normal?

jluttine:
I'm trying to load the library as told in [the realgud instructions](https://github.com/rocky/emacs-dbgr/wiki/How-to-Use#loading-and-unloading) (`M-x load-library realgud`), but I get error: `Cannot open load file: no such file or directory`. So, I suppose the installation didn't work. Now I'm stuck.

robbyoconnor:
Why not make a pull request ;)

bmag:
@jluttine can you upload the layer to github so we can see and comment? As @robbyoconnor suggested, you can make a pull request and mark it as WIP

jluttine:
Thanks for the suggestion. I made a pull request here: https://github.com/syl20bnr/spacemacs/pull/5915

JohnLunzer:
I realize the PR is stuck but just expressing my interest in this layer.

AndreaOrru:
+1 for this.

pedrovanzella:
This would be great too.

syl20bnr:
`realgud` is now part of the `c-c++` layer.

bkudria:
@syl20bnr Why that layer? `realgud` has support for many different debuggers: https://github.com/realgud/realgud/wiki/Debuggers-Supported

robbyoconnor:
yeah this has power beyond just c/c++...

yubrshen:
[...]

PR #8799, opened 2017-04-26
add debug layer

sukeyisme:
add realgud debuger for c++. related: #5910


dvzubarev:
Thank you, I've tested this PR and it worked quite well for c++. I would suggest to add `realgud-short-key-mode` to e.g. `,ds`. I found myself in need to enable this mode often (when I set breakpoints) or disable it for a while to fully use evil-mode in a buffer. 
Are you able to activate a backtrace buffer: `M-x realgud:window-bt` or `F` in a source file. It hangs for me, I can only recover  by spamming `C-g`.

sukeyisme:
@dvzubarev May be a bug. https://github.com/realgud/realgud/issues/155

syl20bnr:
Thank you ! 👍 
Since the layer is only configured for C and C++ for now I just moved the realgud configuration to the `c-c++` layer. See commit https://github.com/syl20bnr/spacemacs/commit/7d215cb0

PR #8801, opened 2017-04-26
swap-buffers-to-window: symbols fun def void

duianto:
```
problem:  target window doesn't get focus after buffer swap:
          'symbols function definition is void: select-window-by-number'
solution: missing 'winum-' before 'select-window-by-number'
```

#### Reproduction steps: 
1. Split a frame into 2 or more windows, with different buffers
2. Press `SPC u SPC b <target window nr>` to swap buffers between the current and the target window


syl20bnr:
Thank you ! 👍 

PR #8812, opened 2017-04-27
rename-current-buffer-file: non-file bfr sav or ren

duianto:
```
problem:  renaming a non-file buffer, only shows an error message
solution: ask if the buffer should be saved to a file or just renamed
```

Additional changes:
- Removed `(not ...` from the line:
`(if (not (and filename (file-exists-p filename)))`
by moving the non-file buffer section, below the file buffer section.

- Some unmodified lines (by this PR), had their indentation corrected, when the `evil-indent` paragraph key binding: `= i p`, was used.
- Added some newlines to keep the line lengths under 80 characters.


syl20bnr:
Thank you ! 💜 
Nice !

Issue #3222, opened 2015-10-01
ESS: not clear where to set (ess-toggle-underscore nil)


dewarrn1:
ESS's default behavior in R is to replace a typed underscore with R's assignment operator ```<-```.  Many users do not like this.  Setting (ess-toggle-underscore nil) should disable this behavior, but because Spacemacs lazy-loads ESS and the variable needs to be set after ESS loads (or at least after its init phase), it is not clear where to set the ess-toggle-underscore variable.

TheBB:
You can set it where ever. It's safe to set variables before a package is loaded so long as it's not explicitly changed by the loading code, which, as far as I can tell, is not the case here (that is, Spacemacs does *not* set this variable).

TheBB:
Although, for the record, this is how to do postpone code until something loads.

`(with-eval-after-load 'ess (setq ess-toggle-underscore nil))`

dewarrn1:
Thanks for the feedback.  For the record, I've tried a bunch of variations (including adding your recent suggestion to dotspacemacs/user-config), and I still can't suppress the <- for _ behavior.  I don't know if this is an ESS problem or a Spacemacs-plus-ESS problem.

TheBB:
First thing you can check is whether the value of `ess-toggle-underscore` actually *is* `nil` when that happens.

dewarrn1:
OK, and now is when I reveal my massive ignorance of Emacs: how do I check that value?

justbur:
`SPC hdv` allows you to see the docs and value of any variable.

dewarrn1:
OK, got it.  So, some new insight: I appear to have misinterpreted the ```(ess-toggle-underscore nil)``` suggestion I mentioned above.  Instead of setting a variable ```ess-toggle-underscore``` to ```nil```, what I need to do is run the ESS function ```ess-toggle-underscore``` with ```nil``` as the argument.  I tried ```(with-eval-after-load 'ess (ess-toggle-underscore nil))``` and a couple of variations, but that throws a "File mode specification error" for the function.

TheBB:
Okay, maybe you have to do that with an active ESS buffer. Try putting it in a hook

```lisp
(add-hook 'ess-mode-hook
          (lambda () (ess-toggle-underscore nil)))
```

dewarrn1:
That did it — you guys rock!  Thanks a bunch.

TheBB:
[...]

PR #8827, opened 2017-04-29
ESS layer option to turn off automatic replacement of underscores

jackkamm:
Added variable `ess-disable-underscore-assign` to `ess` layer to disable automatic replacement of underscore with `<-` assignment operator. Also replaced the  instructions in the Readme on how to manually do this with a hook.

Disabling the automatic replacement is a common thing to do by ESS users, and a variable to do this was previously requested in the thread https://github.com/syl20bnr/spacemacs/issues/3222, with a spacemacs maintainer responding that they wouldn't mind such an option, but no PR was ever submitted.

For now I leave the default behavior unchanged, i.e. `ess-disable-underscore-assign` is `nil` by default. However, IMO we should consider changing the default to `t`, the vanilla ESS behavior is from an earlier time when underscore was uncommon in R code, but now it is ubiquitous.

### DIFF
--- a/layers/+lang/ess/README.org
+++ b/layers/+lang/ess/README.org
@@ -51,19 +51,21 @@ Helpers for inspecting objects at point are available in R buffers only.
 | ~SPC m h t~ | view table using [ess-R-data-view][ess-R-data-view]                 |
 
 * Options
-=ess-smart-equals= is disabled by default. In order to enable it, set in your
-=~/.spacemacs=
+
+To turn off the automatic replacement of underscores by =<-=,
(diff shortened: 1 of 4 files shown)


jackkamm:
Crap, I tried to update my develop branch to be current and accidentally added commits to this PR >_> Also on second thought I think it would be best for me to first raise an issue to discuss changing the default behavior of ESS layer, and if people are in agreement, combine it with this PR. I will close this PR for now, and resubmit it on a clean branch, after I have solicited the feedback.

PR #8830, opened 2017-04-29
Add custom layout for mu4e buffers

benjamreynolds:
This commit adds a custom layout for mu4e related buffers. Four different major
modes are used in mu4e buffers based on their purpose, so a function is used to
add buffers in these modes to the new layout. This setup is largely based on
that of the ERC layers custom layout.


syl20bnr:
Thank you ! 👍 

PR #8831, opened 2017-04-29
Replace define-advice with advice-add in the shell layer

Stebalien:
`define-advice` requires Emacs 25.

Addresses: https://github.com/syl20bnr/spacemacs/pull/8635#issuecomment-297887734

Really, this should have a regression test. Unfortunately, I have no idea how to add such a thing to spacemacs.


syl20bnr:
Nice catch 👍 
Thank you !

PR #8835, opened 2017-05-01
Add a new keybinding for org-preview-latex-fragment

wrbs:
The keybinding I chose is `<space> m X`. `l`, `L` and `x` were taken already, and the X can be remembered by LaTe**X**.

I'm unsure if I put it in the right place in the docs but there didn't seem to be any other more relevant section.


syl20bnr:
Thank you !
Org mode key bindings were a mess, I reorganized them in commit https://github.com/syl20bnr/spacemacs/commit/4898644a
`org-preview-latex-fragment` has been moved to `SPC m t x`.

PR #8838, opened 2017-05-01
Add keybinding to Clojure layer for eval to comment

jr0cket:
Added a Spacemacs style keybinding for the function cider-eval-defun-to-comment

This function evaluates and expression and displays the result as a comment on the following line.

The CIDER keybinding is `C-C M-;` so the Spacemacs binding uses the `;` convention, which is also the general character for comments in Emacs.

As this is an evaluation function, the keybinding is placed under the evaluation part of the major mode menu.


syl20bnr:
Thank you ! 👍 
Nice choice of key binding.

PR #8860, opened 2017-05-06
helm/ivy: Limit rg results to 150 columns (fix emacs freeze)

aaronjensen:
Emacs doesn't handle long line lengths very well. If you accidentally rg a file
a file that has very long line lengths it can freeze emacs. This prevents rg
from returning any results with more than 150 characters. Instead, it will
indicate that there were results in that file but they were elided.

I could imagine this being configurable, but it'd probably be better to wait and see if someone requests it since it is possible to override it manually, at least with ivy.

### DIFF
--- a/layers/+completion/helm/packages.el
+++ b/layers/+completion/helm/packages.el
@@ -250,7 +250,7 @@ Search for a search tool in the order provided by `dotspacemacs-search-tools'."
         ;; --line-number forces line numbers (disabled by default on windows)
         ;; no --vimgrep because it adds column numbers that wgrep can't handle
         ;; see https://github.com/syl20bnr/spacemacs/pull/8065
-        (let ((helm-ag-base-command "rg --smart-case --no-heading --color never --line-number"))
+        (let ((helm-ag-base-command "rg --smart-case --no-heading --color never --line-number --max-columns 150"))
           (helm-do-ag dir)))
 
       (defun spacemacs/helm-files-do-rg-region-or-symbol ()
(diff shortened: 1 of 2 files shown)


syl20bnr:
150 characters is not a lot, what about 500+ ?

aaronjensen:
> 150 characters is not a lot, what about 500+ ?

I'm not sure of the threshold at which emacs gets slow. I'm sure that it varies from machine to machine and 150 is relatively safe. Also, abo-abo [has accepted 150](https://github.com/abo-abo/swiper/pull/988) into swiper (by the way, is there a reason that `counsel-rg` is not used?), and 150 should be more than enough for most normal programming files (it's common to limit line length between 80 and 100 characters).

That all said, I'm happy to bump it to whatever you think would be best and pull request it back to swiper if there's good reason to. My thinking on it was, set it to something "reasonable" and if people complained, bump it up.

Thoughts?

aaronjensen:
FWIW, I just tested w/ 500 characters and it was fine on my machine. I'm happy to bump this in both places. It would sufficiently cover lots of scenarios while still preventing the problematic ones (minified js files, for example)

@abo-abo do you have any preferences?

syl20bnr:
Emacs is slow with long lines, but I don't think that 500 characters is an issue for Emacs.

> a reason that counsel-rg is not used?

I guess that originally we had our own functions but if we can rely on upstream now code let's do it.
In the meantime I'll merge this PR, the value can be updated later, either by bumping the value or using councel-rg.

syl20bnr:
Thank you ! 👍 

PR #8861, opened 2017-05-07
error-transient-state: n/{Np} for next/prev error inst. of only n/p

braham-snyder:
Add <kbd>N</kbd> for `spacemacs/previous-error` to the `error-transient-state` keybindings (in addition to the existing keybinding <kbd>p</kbd>)


syl20bnr:
Thank you ! 👍 

PR #8864, opened 2017-05-08
Add X clipboard support layer for terminal emacs.

### LINKED ISSUE #4662: New layer: clipboard support on terminal for multiple systems
Right now, there doesn't seem to be clipboard support with spacemacs TUI on Linux, and from the gitter chat I get that it doesn't work on other systems too.

Maybe is a good idea to gather here together the relevant configs to make a layer with them, with some layer docs on external dependencies.

cweill:
Add X clipboard support layer to terminal emacs. 

The X clipboard allows a user to copy and paste content between different X windows, e.g. copying text from Chrome into a Terminal. Copy/pasting with the X clipboard it well supported in GUI Emacs, but not so well in terminal Emacs (i.e `emacs -nw` or `emacsclient -t`) without resorting to using the mouse, since terminal Emacs has no awareness of X. There are several incomplete Elisp solutions out that work for the most part, but may not have cross-platform support, or may fail over SSH with X forwarding or within a `tmux` session. 

This layer adds support for OSX, Linux, Windows, and Cygwin using the relevant binary on each system. For example on Linux, it uses `xsel` or `xclip` to interface with the clipboard, depending which one is available. It also adds support for ssh'ing into a different OS with X forwarding via `ssh -Y hostname`, and copy/pasting to and from a remote terminal Emacs. It also supports an edge case of continuing to work in an Emacs instance running inside a `tmux` session which may have been started by a different ssh session, which relies on explicitly reseting the `$DISPLAY` environment variable before calling the relevant binary. 

Yank code inspired by https://github.com/tmux-plugins/tmux-yank.

Fix #4662.

### DIFF
--- a/layers/+tools/xclipboard/README.org
+++ b/layers/+tools/xclipboard/README.org
@@ -1,0 +1,15 @@
+#+TITLE: xclipboard layer
+
+* Table of Contents                                        :TOC_4_gh:noexport:
+ - [[#description][Description]]
+ - [[#key-bindings][Key Bindings]]
+
+* Description
+This layer adds copy/paste support to the X-clipboard from the terminal.
+
+* Key Bindings
+
+| Key Binding | Description                                 |
+|-------------+---------------------------------------------|
+| ~SPC x y~   | Copy selection to clipboard                 |
+| ~SPC x p~   | Paste clipboard contents at cursor position |
--- a/layers/+tools/xclipboard/funcs.el
+++ b/layers/+tools/xclipboard/funcs.el
@@ -1,0 +1,96 @@
+;;; funcs.el --- xclipboard layer functions file for Spacemacs.
+;;
+;; Copyright (c) 2012-2017 Google Inc. & Contributors
+;;
+;; Author: Charles Weill <weill@google.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+(defun xclipboard/get-display ()
+  (shell-command-to-string "if [[ -n $TMUX ]]; then
+      export DISPLAY=$(tmux show-environment | grep -o '^DISPLAY.*$' | sed 's/DISPLAY=//')
+    fi
+    if [[ -z $DISPLAY ]]; then
+      export DISPLAY=:0
+    fi
+    printf $DISPLAY")
+  )
+
+(defun xclipboard/get-copy-command ()
+  (shell-command-to-string "command_exists() {
+    local command=\"$1\"
+    type \"$command\" >/dev/null 2>&1
+}
+
+  # Installing reattach-to-user-namespace is recommended on OS X
+  if command_exists \"pbcopy\"; then
+      if command_exists \"reattach-to-user-namespace\"; then
+          printf \"reattach-to-user-namespace pbcopy\"
+      else
+          printf \"pbcopy\"
+      fi
+  elif command_exists \"clip.exe\"; then # WSL clipboard command
+      printf \"clip.exe\"
+  elif command_exists \"xsel\"; then
+      printf \"xsel -ib\"
+  elif command_exists \"putclip\"; then # cygwin clipboard command
+      printf \"putclip\"
+  fi")
+  )
+
+(defun xclipboard/get-paste-command ()
+  (shell-command-to-string "command_exists() {
+    local command=\"$1\"
+    type \"$command\" >/dev/null 2>&1
+}
+
+# Installing reattach-to-user-namespace is recommended on OS X
+if command_exists \"pbpaste\"; then
+    if command_exists \"reattach-to-user-namespace\"; then
+        printf \"reattach-to-user-namespace pbpaste\"
+    else
+        printf \"pbpaste\"
+    fi
+elif command_exists \"paste.exe\"; then # WSL clipboard command
+    printf \"paste.exe\"
+elif command_exists \"xsel\"; then
+    printf \"xsel -ob\"
+elif command_exists \"getclip\"; then # cygwin clipboard command
+    printf \"getclip\"
+fi")
+  )
+
+(defun xclipboard/copy ()
+  "Copies selection to x-clipboard."
+  (interactive)
+  (if (display-graphic-p)
+    (progn
+      (message "Copied region to x-clipboard!")
+      (call-interactively 'clipboard-kill-ring-save)
+      )
+    (if (region-active-p)
+      (progn
+        (shell-command-on-region (region-beginning) (region-end) (format "DISPLAY=%s %s" (xclipboard/get-display) (xclipboard/get-copy-command)))
+        (message (format "Copied region to clipboard \"%s\"!" (xclipboard/get-display)))
+        (deactivate-mark)
+        )
+      (message "No region active; can't copy to clipboard!")
+      )
+    )
+  )
+
+(defun xclipboard/paste ()
+  "Pastes from x-clipboard."
+  (interactive)
+  (if (display-graphic-p)
+    (progn
+      (clipboard-yank)
+      (message "graphics active")
+      )
+    (insert (shell-command-to-string (format "DISPLAY=%s %s" (xclipboard/get-display) (xclipboard/get-paste-command))))
+    )
+  (message (format "Pasted from clipboard \"%s\"!" (xclipboard/get-display)))
+  )
--- a/layers/+tools/xclipboard/funcs.el
+++ b/layers/+tools/xclipboard/funcs.el
@@ -1,15 +1,16 @@
 ;;; funcs.el --- xclipboard layer functions file for Spacemacs.
 ;;
-;; Copyright (c) 2012-2017 Google Inc. & Contributors
+;; Copyright (c) 2012-2018 Sylvain Benner & Contributors
 ;;
-;; Author: Charles Weill <weill@google.com>
+;; Authors: Charles Weill <weill@google.com>
+;;          Google LLC.
 ;; URL: https://github.com/syl20bnr/spacemacs
 ;;
 ;; This file is not part of GNU Emacs.
 ;;
 ;;; License: GPLv3
 
-(defun xclipboard/get-display ()
+(defun spacemacs//xclipboard-get-display ()
   (shell-command-to-string "if [[ -n $TMUX ]]; then
       export DISPLAY=$(tmux show-environment | grep -o '^DISPLAY.*$' | sed 's/DISPLAY=//')
     fi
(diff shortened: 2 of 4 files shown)


sdwolf:
👍 

Any way to set the `interprogram-cut-function` and `interprogram-paste-function` so that normal yank and paste use the clipboard directly, just like [xclip.el](https://www.emacswiki.org/emacs/xclip.el)?

cweill:
Some people may want to keep their emacs registers separate from the X-clipboard, so I'm hesitant to add this code to the keybindings. I could include in the README instructions to optionally add the following to  `dotspacemacs/user-config`:

```elisp
(setq interprogram-cut-function 'xclipboard/yank)
(setq interprogram-paste-function 'xclipboard/paste)
```

sdwolf:
How about:

```emacs-lisp
(defun xclipboard/turn-on-interprogram ()
  (interactive)
  (setq interprogram-cut-function 'xclipboard/yank)
  (setq interprogram-paste-function 'xclipboard/paste))

(defun xclipboard/turn-off-interprogram ()
  (interactive)
  (setq interprogram-cut-function nil)
  (setq interprogram-paste-function nil))
```

So in your `user-config` you can just:

```emacs-lisp
(xclipboard/turn-on-interprogram)
```

cweill:
Sure. Which file should that code go in?

sdwolf:
I think in `layers/+tools/xclipboard/funcs.el` since it will be part of the public API.

cweill:
Done.

sdwolf:
Awesome! 🎉 
I will try it out tommorow at work and let you know if I find any issues.

cweill:
There is a known issue when trying to copy from a remote server to the OSX pasteboard. This can happen when you are on a MacBook, and you `ssh -Y hostname` into a remote machine. If you try to copy something from an emacs running on the remote host (or even `xsel` or `xclip`) . This is due to a [bug with XQuartz](https://bugs.freedesktop.org/show_bug.cgi?id=92650) which took me quite a while to track down.

layers/+tools/xclipboard/README.org
```diff
@@ -0,0 +1,15 @@
...
+
+* Description
+This layer adds copy/paste support to the X-clipboard from the terminal.
+
+* Key Bindings
+
+| Key Binding | Description                                 |
+|-------------+---------------------------------------------|
+| ~SPC o y~   | Copy selection to clipboard                 |
+| ~SPC o p~   | Paste clipboard contents at cursor position |
```
robbyoconnor: You can't use `SPC o` in official layers. Please choose something else.

layers/+tools/xclipboard/keybindings.el
```diff
@@ -0,0 +1,13 @@
...
+;;
+;; Author: Charles Weill <weill@.google.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+(evil-leader/set-key "o y" 'xclipboard/yank)
+(evil-leader/set-key "o p" 'xclipboard/paste)
```
robbyoconnor: The `SPC o` prefix is reserved for custom keybindings.

robbyoconnor:
See comments above - I'm shocked nobody caught this.

sdwolf:
`SPC x` seems to have the `p` and `y` keys available.

I've enabled `xclipboard/turn-on-interprogram` in my `user-config`

Problems found so far:
1. Pressing `x` in vim normal mode now does a paste.
2. I get a freeze when trying to yank with `SPC o y`

emacs --version
```
GNU Emacs 26.0.50
Copyright (C) 2017 Free Software Foundation, Inc.
GNU Emacs comes with ABSOLUTELY NO WARRANTY.
You may redistribute copies of GNU Emacs
under the terms of the GNU General Public License.
For more information about these matters, see the file named COPYING.
```
xclip -version
```
xclip version 0.13
Copyright (C) 2001-2008 Kim Saunders et al.
Distributed under the terms of the GNU GPL
```

uname -a
```
Linux CGI 4.9.24-1-MANJARO #1 SMP PREEMPT Fri Apr 21 08:32:28 UTC 2017 x86_64 GNU/Linux
```

layers/+tools/xclipboard/README.org
cweill: What would be a good replacement? How about `SPC y y/p`?

robbyoconnor: Read: https://github.com/syl20bnr/spacemacs/blob/develop/doc/CONVENTIONS.org

layers/+tools/xclipboard/keybindings.el
```diff
@@ -0,0 +1,13 @@
...
+;;
+;; Author: Charles Weill <weill@.google.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+(evil-leader/set-key "o y" 'xclipboard/copy)
+(evil-leader/set-key "o p" 'xclipboard/paste)
```
robbyoconnor: Read: https://github.com/syl20bnr/spacemacs/blob/develop/doc/CONVENTIONS.org

cweill:
@sdwolf - I addressed your comments in a few commits:

1. I fixed this by just removing the `'xclipboard/turn-[on,off]-interprogram` functions since I'm not sure yet how to implement them correctly. 
1. The freezing is due to an issue with `xclip`, so I've dropped support for it for now, since `xsel` is also available on Linux.

I could not reproduce the issue of `SPC x` having the p and y keys available.

I'd like some suggestions for the best key-bindings to use since `SPC o` is reserved. Any advice is welcome.

layers/+tools/xclipboard/README.org
robbyoconnor: I'd also just make the normal evil copy/paste commands work like this...and since we're dealing with text, use `SPC x`

robbyoconnor: Both `SPC x p` and `SPC x y` are free.

robbyoconnor: @cweill update this as well.

robbyoconnor:
Also rebase when you're done ;)

layers/+tools/xclipboard/README.org
cweill: Fixed to `SPC x y` and `SPC x p`.

As for making normal evil copy/paste commands work like this: I'm not familiar enough with elisp to know how to do that, since interprogram-copy-function behaves differently than expected as reported by @sdwolf.

cweill:
Squashed, though I guess we lost the comments.

layers/+tools/xclipboard/README.org
robbyoconnor: That is why I didn't explicitly approve this...I don't need to test this..as it works just fine for me....

cweill:
Do you mind if I add "Google Inc." to the "Copyright" section of the file headers, since it's a policy at Google for us to do so in our open source contributions?

sdwolf:
Actually I was sugesting `SPC x p` and `SPC x y` for the keybindings.

The interprogram functions were a nice to have, no problem if they are not included in this PR.

cweill:
I've updated the copyright section. Please let me know if it that's okay, otherwise I can revert it.

layers/+tools/xclipboard/packages.el
```diff
@@ -0,0 +1,12 @@
...
+;; Copyright (c) 2012-2017 Google Inc. & Contributors
+;;
+;; Author: Charles Weill <weill@google.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+(defconst xclipboard-packages '(xclipboard))
```
syl20bnr: You need to declare a function `xclipboard/init-xclipboard` otherwise Spacemacs won't consider this package as "used" so it won't install the package.
I guess that you have the package listed in `dotspacemacs-additional-packages` so on your system the package is still installed. See other layers for more info.

layers/+tools/xclipboard/packages.el
```diff
@@ -0,0 +1,12 @@
+;;; packages.el --- xclipboard layer packages file for Spacemacs.
+;;
+;; Copyright (c) 2012-2017 Google Inc. & Contributors
```
syl20bnr: This line should be:
```
;; Copyright (c) 2012-2017 Sylvain Benner & Contributors
```

You can add the google one in the `Author` field, you can rename it `Authors`, see https://github.com/syl20bnr/spacemacs/blob//layers/+distributions/spacemacs-base/local/hybrid-mode/hybrid-mode.el#L5-L6

layers/+tools/xclipboard/funcs.el
```diff
@@ -0,0 +1,96 @@
...
+;; Copyright (c) 2012-2017 Google Inc. & Contributors
+;;
+;; Author: Charles Weill <weill@google.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+(defun xclipboard/get-display ()
```
syl20bnr: We usually put all the functions in `funcs.el` under the `spacemacs/` prefix.
For instance this one should be renamed `spacemacs/xclipboard-get-display`.

layers/+tools/xclipboard/packages.el
cweill: Done.

alanxoc3:
Is this working? I just tried to install the "xclipboard" layer, and it couldn't be found. This is a useful layer for me!

robbyoconnor:
@alanxoc3 unless you pulled this layer locally -- it won't work.

alanxoc3:
Cool, I didn't know I could pull in pull requests locally. I got it installed, but it doesn't quite work for me. Maybe I'll check it out later.

robbyoconnor:
So basically layers are loaded from very specific locations. Unless you have it loaded -- it won't work.  I personally don't see a need for this as it works for me just fine.

cweill:
@robbyoconnor: will you be reviewing this PR?

robbyoconnor:
@cweill unfortunately not - I understand it's been a while, but be patient-- things don't move fast around here

cweill:
No problem. Whenever you get the chance. I'll try to be prompt in responding.

jesuspc:
That's a feature that I've needed for a long time. Thanks a lot!

I've just tried it out and behaves as expected. I work inside tmux on a Nixos with xsel as a server, and I connect to it from a mac OS High Sierra with x-forwarding.

It would be great if the layer also provided the option that was mentioned before to enable normal copy and paste commands (in evil mode yy, dd, p, etc.) to transparently use xclipboard.

syl20bnr:
Thank you ! 👍 

I made a local package to make it more idiomatic, see commit https://github.com/syl20bnr/spacemacs/commit/1befcad00

I just tested that the key-bindings are correctly installed.

PR #8870, opened 2017-05-09
Fix graphviz erroneous usage of default-tab-width in Emacs 26

tuhdo:
default-tab-width is obsolete since Emacs 23 and is removed in Emacs 26. Better
adapt it properly.


syl20bnr:
Thank you !
I removed the unnecessary `unless` guard.

PR #8878, opened 2017-05-10
In OSX, add key binding to use Cmd-1..9 for selecting window

liujoey:
In OSX it is very intuitive to select window by pressing Cmd-1..9. So I added these keybindings to the OS X layer.


syl20bnr:
Thank you !

PR #8881, opened 2017-05-10
Define SPC C c 'org-capture on :init

duianto:
```
problem:  SPC C c was defined in the org layers :config section, which
          runs when a .org file is loaded
solution: move the definition to the :init section, so that it loads on
          startup, like SPC a o c and C-c c
```


syl20bnr:
Thank you !

PR #1526, opened 2015-05-08
Delete or simplify Editorconfig layer

jupl:
Editorconfig package has been updated, meaning that the layer has been simplified. The PR is for if you still want the layer around, but I would just delete the layer instead now that `dotspacemacs-additional-packages` can be used to install the Editorconfig package.

PR #8883, opened 2017-05-11
Add editorconfig layer.

marshall-lee:
Sometime in the past editorconfig layer has been deleted.

Commit: 5887344.
Related PR: #1526.

Explanation was that layer was so simple that everything you need to get editorconfig to work is just add it to `dotspacemacs-additional-packages`.

However it's not quite true.

- `(editorconfig-mode t)` should be performed manually.
- it looks ugly in modeline.

### DIFF
--- a/layers/+tools/editorconfig/packages.el
+++ b/layers/+tools/editorconfig/packages.el
@@ -1,0 +1,21 @@
+;;; packages.el --- editorconfig Layer Packages File for Spacemacs
+;;
+;; Copyright (c) 2012-2017 Sylvain Benner & Contributors
+;;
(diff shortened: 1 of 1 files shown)


jupl:
Agreed. It used to be automatically enabled when loaded, but that got changed. You can probably grab the README and image I used for this package.

TheBB:

I moved it to spacemacs-editing, seems like a reasonable default package. Also, I decided to remove the MM lighter completely (capital E is for editing styles).

marshall-lee:
@TheBB thanks, it's a good news!

PR #8887, opened 2017-05-11
Add key binding for helm-gtags-find-symbol: '<SPC> g y'

Sodel-the-Vociferous:
There was a keybinding for helm-gtags-find-rtag, but no binding for helm-gtags-find-symbol. I added one, and also put the binding in the layer's README. :)


TheBB:

PR #8899, opened 2017-05-15
Add a layer for LanguageTool support

robbert-vdh:
This adds support for grammar checking using [LanguageTool](https://languagetool.org/). It tries to use the current flyspell language (the language set with `SPC S d`) when available. I couldn't think of a good navigation shortcut as the mnemonic ones were already in use, so I went with `[ a` and `] a` for the time being as they are easy to reach.


layers/+tools/languagetool/README.org
```diff
@@ -0,0 +1,66 @@
+#+TITLE: LanguageTool layer
+
+[[img/languagetool.png]]
```
JAremko: Replace it with `[[file:img/languagetool.png]]`

robbert-vdh: Ah, I was wondering what I was missing. Fixed it in the commit.

JAremko:
@robbert-vdh  test on emacs 24.4 seems to glitch so ignore it.

syl20bnr:
We can move this package to the `spell-checking` layer. If the term `spell-chrcking` is not appropriate anymore, I'm not against renaming it to something better.

Not behind a computer right now but `SPC S n` is not free ?

JAremko:
@syl20bnr  how about renaming it into` natural languages layer` ?
  Does anyone still remember https://github.com/syl20bnr/spacemacs/issues/5996 :smile:  ?

JAremko:
@syl20bnr 
> Not behind a computer right now but SPC S n is not free ?   
it's [`flyspell-goto-next-error`](https://github.com/syl20bnr/spacemacs/blob/bd7ef98e4c35fd87538dd2a81356cc83f5fd02f3/layers/%2Bcheckers/spell-checking/packages.el#L67)

robbert-vdh:
I agree that it would be nice to have a few different options for working with prose, though it would probably be better to keep them separate from regular spell checking. Flyspell only depends on either `aspell` or `hunspell`, which is probably already installed on the system, while other packages like `langtool` rely on their specific binaries being available. It might be an idea to combine these kinds of tools into a prose layer (like in #5996) and put the different packages behind flags.

shwsun:
Hi @robbert-vdh @JAremko @syl20bnr , any further update on this PR?

I think a grammar checking tool like Language Tool could be very help full if some Spacemacs users do writing heavily. `LangTool` checks the grammar and also gives suggestions to correct them which is a bit impressive to me.

jsmestad:
Yeah this works great for me as well.

gdkrmr:
I would love to see this as a layer!

c02y:
Any updates? Anything wrong with merging this PR?

duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/14dcebd4a30a1a18847adbaefa9df8d1e5c2bc4d

PR #8902, opened 2017-05-15
named django mode prefixes

BorisVerk:


TheBB:

PR #8906, opened 2017-05-16
fix `prodigy-view-clear-buffer` binding

FrancescElies:
#### Description :octocat:
Pressing `c` in `Prodigy-view` mode should clear the buffer instead of being closed.

Related Pull Request: https://github.com/syl20bnr/spacemacs/pull/8636

- Start Emacs
- `SPC a S`   | *Open the prodigy buffer*
- Select with the cursor any process you like
- `L`         | *Start prodigy*
- `H`         | *Display current process*
- `c`         | *Clear buffer*

- `c`         | Runs *View-leave* and closes the buffer

- `c`         | Should run *Clear buffer* `prodigy-view-clear-buffer`

Best & thanks in advance for your time

### DIFF
--- a/layers/+tools/prodigy/README.org
+++ b/layers/+tools/prodigy/README.org
@@ -34,24 +34,24 @@ You start prodigy with this:
 
 ** Navigate through it
 
-| Key Binding | Description                     |
-|-------------+---------------------------------|
-| ~c~         | Clear buffer                    |
-| ~h~         | First service                   |
-| ~j~         | Next service                    |
-| ~k~         | Previous service                |
-| ~l~         | Last service                    |
-| ~H~         | Display current process         |
-| ~J~         | Next service with status        |
-| ~K~         | Previous service with status    |
-| ~L~         | Start prodigy                   |
-| ~S~         | Stop prodigy                    |
-| ~r~         | restart prodigy                 |
-| ~R~         | revert buffer  (refresh list)   |
-| ~d~         | Jump to the dired of service    |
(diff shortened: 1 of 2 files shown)


TheBB:
I can't reproduce your problem. The keybinding works fine for me. From your code it seems you are inadvertently finding yourself in motion state rather than evil state. I don't use prodigy so I can't comment on whether that is intentional or not, but at any rate I can't merge a fix for a problem that may not exist.

FrancescElies:
Ok, fine, I will close this one for now and keep the changes for myself until I find out what is actually going on.

FrancescElies:
Hi @TheBB I thought a bit more about this, I am still a bit puzzled how this actually can work.

The main problem I see with the current definition (see below) is that the binding is being associated with `prodigy-mode`, as far as I can tell `prodigy-view-clear-buffer` should be used in `prodigy-view-mode` buffers.
```
(evilified-state-evilify prodigy-mode prodigy-mode-map
    "c" 'prodigy-view-clear-buffer ...)
```

With regards to motion state the prodigy package sets buffers `prodigy-view-mode` as read only, and if I understand correctly spacemacs sets the motion state for read-only buffers.
```
(evil-define-key 'motion prodigy-view-mode-map "c" 'prodigy-view-clear-buffer)
```

Does this make sense? please correct me if I'm wrong.
May I reopen this one?

Thanks

FrancescElies:
Sorry, ignore my previous message, somehow now I do not end up in evil motion state, not sure why, maybe because I updated the packages.

Thanks for your help

PR #8908, opened 2017-05-16
Add a key binding for magit-find-file

duianto:
```
problem:  currently, viewing a file at a specific branch/commit, requires one to
          type the magit-find-file command, at the M-x prompt
solution: SPC g f f is available and consistent with the SPC f f key binding
```


benjamreynolds:
Could you document the new keybinding in the README for the git layer?

duianto:
@benalbrecht Of course how could i have forgotten that, mybad, thanks for the reminder 👍

duianto:
I didn't change the second duplicate header link `#git-flow` to `#git-flow-1`, it must be an automatic feature that corrects duplicate header links, when the documentation files are modified.

TheBB:
I like it. 

PR #8909, opened 2017-05-16
Fix frame cycling in evil-unimpaired.

marshall-lee:
`raise-frame` does not set input focus on selected frame.


TheBB:
Indeed, much better. Thanks! Cherry-picked in develop. You can safely delete your branch.

marshall-lee:
Thanks!

PR #8911, opened 2017-05-16
Update git doc magit-clone and magit-init

duianto:


TheBB:

PR #8915, opened 2017-05-17
Mention the change of dotspacemacs-check-for-update in docs

xfq:
Its default value was changed to `nil` in 9d72b122459a45edc81511522f72c7ba58488e50.


TheBB:

PR #8920, opened 2017-05-17
Fix markdown table cleanup hook

sdwolf:
The org table cleanup must be executed at the `before-save-hook` for it to take effect on tables inside a markdown document. The commit that changed this already working behaviour was 4fd895a9170b45f5959e4971c1565561303f0fd4. 

In order to test this out you need to create a `test.md` file with the following contests:
```markdown
| A | B |
|---+---|
| a | b |
```

Replace one of the letters, let's say `A` with `C` and then save.

**Before this PR**: The `+` is not replaced by a `|`.
**After this PR**: The `+` is replaced by a `|`.

### DIFF
--- a/layers/+lang/markdown/packages.el
+++ b/layers/+lang/markdown/packages.el
@@ -64,7 +64,9 @@
     :config
     (progn
       (add-hook 'markdown-mode-hook 'orgtbl-mode)
-      (add-hook 'markdown-mode-hook 'spacemacs//cleanup-org-tables nil 'local)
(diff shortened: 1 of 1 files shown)


TheBB:

I think that with very few exceptions, hook functions should be named, so I made this change: https://github.com/syl20bnr/spacemacs/commit/3dde98821297f2fb51cfe2f9a668097f4313e996

sdwolf:
Indeed it looks better now.

PR #8926, opened 2017-05-18
Add note about separate ledger install.

legalcodes:
Finance layer depends upon a separate install of ledger, but our docs do not yet convey this to users.


TheBB:

PR #8930, opened 2017-05-19
Shell scripts: add new company-shell environment variable backend.

Alexander-Miller:
Simple change to make the shell scripts layer use company-shell's new environment variable backend.


TheBB:

PR #8931, opened 2017-05-19
Spacefmt update toc-org

JAremko:
<s>From the looking at https://travis-ci.org/syl20bnr/spacemacs/jobs/233872503 It appears that we have an outdated copy of toc-org</s>  it probably happens because spacefmt used build-in org.

- - - - -

This is temporary fix just to make it work properly. I'm planning to rewrite spacefmt on pure ELisp and wrap it into a package that will be an initial part of Contributing layer


JAremko:
Actually it will save time if I do it now instead of putting lipstick on a pig :smile:

JAremko:
Ok we simply update toc-org for now as a hotfix

JAremko:
ok now everything should be fixed.

JAremko:
`^M` :rage:

JAremko:
@syl20bnr  Can we crucify someone for this https://github.com/syl20bnr/spacemacs/pull/8931/commits/46cbe1874593c56d2b408027d4cd22a9bce03c18 Or it's against community guidelines?

JAremko:
@TheBB  Seems ready :roll_eyes:

TheBB:

PR #8936, opened 2017-05-20
Keep find-ert-test-buffer from moving point.

Alexander-Miller:
`spacemacs/ert-run-tests-buffer` has an annoying habit of moving point to somewhere else from where it is run. This change will fix it.

I moved the change one level down on the call stack to make for a more general solution - I don't know if `find-ert-test-buffer` is also used elsewhere - but its purpose is to only get a reference to a buffer and it's explicitly calling a -noselect function, so putting the change here should be fine.


TheBB:

PR #8942, opened 2017-05-22
Render Madoko document using markdown mode

hanw:
Madoko is a fast markdown processor with a special .mdk extension
from microsoft research. emacs should use markdown-mode to render
a .mdk file as markdown file.


robbyoconnor:
I recommend you close this and open a PR in a feature branch. This is because PRs tend to sit around here for awhile and you want ONLY your change in here.

hanw:
Close as suggested.

PR #8944, opened 2017-05-23
Render Madoko document using markdown mode

hanw:
Madoko is a fast markdown processor with a special .mdk extension
from microsoft research. emacs should use markdown-mode to render
a .mdk file as markdown file.


TheBB:

PR #8945, opened 2017-05-23
Fill the current filename as a suggestion of spacemacs/rename-current…

tddsg:
Hi,

When using the function `spacemacs/rename-current-buffer-file` (`SPC-f-R`), I realize that most of the time, I create a new file name by modifying the current file name. 

Would it make sense to change the current file name suggestion (which is the folder path containing the current file), to the new file name suggestion, which is the entire file path? 

I create a PR for this suggestion, and there are two reasons for this modification:

- Deleting some words/character, and then writing some new words/characters is normally faster than writing the entire new name (of course, when using Emacs strokes).

- If we write a new name from scratch, the fuzzy matching feature of Helm sometimes suggests existing file names that we don't want, so we have to move the cursor to the new file name beginning with `[?]` (in Helm window). That takes more effort to write a new name.

Please let me know your opinion about this PR.
Thanks!

### DIFF
--- a/layers/+distributions/spacemacs-base/funcs.el
+++ b/layers/+distributions/spacemacs-base/funcs.el
@@ -301,8 +301,7 @@ be saved to a file, or just renamed."
          (filename (buffer-file-name)))
     (if (and filename (file-exists-p filename))
         ;; the buffer is visiting a file
-        (let* ((dir (file-name-directory filename))
-               (new-name (read-file-name "New name: " dir)))
+        (let ((new-name (read-file-name "New name: " filename)))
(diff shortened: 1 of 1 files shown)


TheBB:
It's fine with me. I added a commit that makes the behaviour changeable with a prefix argument, your suggestion being the default: https://github.com/syl20bnr/spacemacs/commit/8f7afe5eec1a0a6c65ab156dc915049ebc0c539f

tddsg:
Thanks for the update! Yes, adding the prefix argument gives more flexibility!

PR #8949, opened 2017-05-23
ivy: Use new API to register additional transient state bindings

madand:
This is a follow-up commit after #8155 has been merged. It applies the same changes to code of the `ivy` layer. It is not part of #8155 because there was no `ivy` layer when that PR was submitted.


TheBB:

PR #8953, opened 2017-05-24
init-evil-mc: 1 cursor hide mode line text variable

duianto:
Problem:
Spacemacs rewrites the evil-mc-mode-line variable, to remove the evil-mc mode
line text, when there's only one cursor. The rewrite blocks the recent evil-mc
updates, that made the multi cursor, mode line text, more readable.

Solution:
Use the new evil-mc-one-cursor-show-mode-line-text variable, to only hide the
mode line text, when there's one cursor. This unblocks the current, and any
future updates, that the evil-mc package makes to the mode line text.

### DIFF
--- a/layers/+spacemacs/spacemacs-evil/packages.el
+++ b/layers/+spacemacs/spacemacs-evil/packages.el
@@ -111,12 +111,7 @@
       ;; evil-mc is not compatible with the paste transient state
       (define-key evil-normal-state-map "p" 'spacemacs/evil-mc-paste-after)
       (define-key evil-normal-state-map "P" 'spacemacs/evil-mc-paste-before)
-      ;; remove emc prefix when there is not multiple cursors
-      (setq evil-mc-mode-line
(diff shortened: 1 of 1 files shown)


TheBB:
Is this bad enough that it needs to go in release?

duianto:
I don't know. More people might have complained if it was a huge issue. It might be enough to merge it to develop for now.

This is what it currently looks like: https://github.com/gabesoft/evil-mc/issues/57#issuecomment-298026013

And with this PR it will look like the second screenshot here:
https://github.com/gabesoft/evil-mc/pull/64
no emc text when there's only one cursor,
yellow background (cursor color), when there's more than 1 cursor and they are resumed (unpaused),
and gray (inverse color) background with (paused) text, when there's more than 1 cursor and they are paused.

TheBB:
Okay. Thanks! Cherry-picked in develop. You can safely delete your branch.

PR #8958, opened 2017-05-24
Update colors layer readme: rainbow-mode screenshot

duianto:
Problem:
The current rainbow-mode screenshot, is from an older version. It shows white
text on light colors. This makes it hard to read the color codes. The old
screenshot hides the fact that issue has been solved.

Solution:
This updates the screenshot, to show it's current behaviour. Where the text
color is either white or black, depending on if the luminance is greater than
0.5 of 1.0.


TheBB:

PR #8969, opened 2017-05-26
[python] Prefer virtualenv python to compile file

swaroopch:
Otherwise, it falls back to system python which may not have the dependencies used in the imports in the file.


TheBB:
You should be able to use `(spacemacs/pyenv-executable-find python-shell-interpreter)` instead. That's what we use for all the other things.

swaroopch:
@TheBB Excellent point. Updated.

TheBB:

PR #8970, opened 2017-05-26
Remove SPARQL from extra-langs

houli:
Support for SPARQL was added with the semantic-web layer in https://github.com/syl20bnr/spacemacs/commit/f7210c30c2fe979eed6033becc14af9af892b431


TheBB:

PR #8971, opened 2017-05-26
[slack] Add key binding to add reaction

swaroopch:


TheBB:

PR #8973, opened 2017-05-27
Add tldr layer

et2010:
This layer adds integration of https://github.com/tldr-pages/tldr which is a collection of simplified and community-driven man pages.


fiveNinePlusR:
looks good to me. The only thing that requires some thought is if those key bindings are the right ones to use.

et2010:
I'm using the `SPC d m` keybinding for `tldr`. Is `SPC d t` better?

fiveNinePlusR:
I looked through the layers and it seems like that is not bound to anything. spc d seems to be mainly for the dash layer but could be renamed to documentation. I personally would suggest that you change the prefix for that leader to show "documentation" instead of dash(look for set-prefix for that). not sure where that would best be put though as it's currently not set with the dash layer. perhaps one of the maintainers knows a good place to put it so it works for both dash and tldr regardless of which layer is installed.

robbyoconnor:
I agree -- change the prefix from `SPC d` please.

et2010:
@fiveNinePlusR @robbyoconnor Thanks for the feedback! I've changed the prefix to 'documentation`.

robbyoconnor:
LGTM.

BadDecisionsAlex:
I want this <3
Thank you!

PR #8980, opened 2017-05-28
Pass: support for ivy and fix undefined dostpacemacs-use-ido

kepi:
Looks like pass is trying to use variable which isn't available for some time,
at least there are no other packages using it.

`completing-read` should be universal for all completing needs I hope.


TheBB:

PR #8982, opened 2017-05-28
themes: use package-activate to support dependencies

bennofs:
This fixes an issue where the solarized light theme wouldn't load early due to missing dependencies.

### DIFF
--- a/core/core-themes-support.el
+++ b/core/core-themes-support.el
@@ -218,15 +218,14 @@ THEME."
           ;; Load theme
           (unless (or (memq theme-name (custom-available-themes))
                       (eq 'default theme-name))
-            (let ((pkg-dir (spacemacs//get-theme-package-directory theme)))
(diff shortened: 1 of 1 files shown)


bennofs:
Some of the remaining code could probably be removed since `package-activate` does it already. Before I do that though, was there any reason not to use `package-activate` originally here?

TheBB:
The original reason was that the theme was one of the first things that was applied, even before the packages were downloaded. However, now I think that that has been made more robust, there's a fallback theme that is applied early on in the load process if the default theme fails to load, and at any rate, the packages are initialized before this, so I believe this is a safe change now.

bennofs:
@TheBB ok, I now removed all code that I think is unnecessary with the `package-activate` approach. The themes I tested are `solarized-light` and `moe-light`

TheBB:

PR #8983, opened 2017-05-28
nixos layer: defer loading of options

bennofs:
This improves startup time (loading the nixos options can take quite some time)


TheBB:

Issue #8992, opened 2017-05-30
key map is not restored after exit evil-evilified-state


enzochiau:
Use 0.200.9025.1.8 with Emacs 25.2  (vim mode)

When exit edebug-mode, it should back to `evil-normal-state` from `evil-evilified-state`, but the keymap is not restarted which the `ESC` is still binded to `evil-evilified-state`. 

Reproduce:
1. Instruct any elisp function for edebug
2. Trigger edebug-mode
3. quit edebug-mode (goes to `evil-normal-state`)
4. press `ESC` (goes to `evil-evilified-state`, which should stay at `evil-normal-state`)

I think the `evilified-state-restore-normal-state-keymap` should be added to the state exit hook.

Miciah:
Duplicate of #1874 and #7766.

PR #8993, opened 2017-05-30
ISSUE-8992 Fix normal-state keymap is not restored issue

enzochiau:
As reported in #8992, after exit `evilified-state`, the normal-state keymap is not restored.
This fix adds restore function to exit-hook.

### DIFF
--- a/layers/+distributions/spacemacs-base/local/evil-evilified-state/evil-evilified-state.el
+++ b/layers/+distributions/spacemacs-base/local/evil-evilified-state/evil-evilified-state.el
@@ -61,7 +61,7 @@
   :enable (emacs)
   :message "-- EVILIFIED BUFFER --"
   :cursor box
-  :exit-hook (list evilified-state--restore-normal-state-keymap))
+  :exit-hook (evilified-state--restore-normal-state-keymap))
 
 (bind-map spacemacs-default-map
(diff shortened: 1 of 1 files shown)


layers/+distributions/spacemacs-base/local/evil-evilified-state/evil-evilified-state.el
```diff
@@ -60,7 +60,8 @@
   :tag " <N'> "
   :enable (emacs)
   :message "-- EVILIFIED BUFFER --"
-  :cursor box)
+  :cursor box
+  :exit-hook (list evilified-state--restore-normal-state-keymap))
```
TheBB: You shouldn't have `list` here.

TheBB:
Yes, but `evil-define-state` is a macro, all its arguments are quoted, and this one is not evaluated. So the final hook you get has two functions, one of them is `list`.

enzochiau:
Thanks for the explanation. "list" is removed.

Miciah:
This change is similar to #7767, which was reverted because it broke mu4e.  Has anyone tested this change with mu4e?

enzochiau:
After almost 3 years, I still have this issue when exit from `edebug-mode`. I am not sure if the fix still break the `mu4e`.

PR #8994, opened 2017-05-30
Use a more consistent keybinding for flycheck-buffer

dindon-sournois:
Changed `flycheck-buffer` keybinding for `SPC e b` (was `SPC e e`) to be more consistent with `flyspell-buffer` keybinding (`SPC S b`).


TheBB:

PR #8997, opened 2017-05-30
Update keyboard layout image, gray border

duianto:
problem:
The keyboard layouts main image, is displayed as a black rectangle, when the
readme.org file, it's opened in Spacemacs. This seems to be because there's an
issue with the transparency when a .png is saved with indexed colors.

When the transparency issue is fixed, by re-saving it as a non-indexed
(grayscale or RGB) image. Then there still is an issue, with themes that have a
dark background, i.e. spacemacs-dark. The black keys become almost invisible.

solution:
Add a 1px gray border, around the keys.

#### This is what the different outcomes are: 
1. Current
2. Transparency fixed
3. This PR

current-transparency-fixed-after

The keys look a little bit blurrier when the keyboard layouts readme.org is viewed on github:

github-keyboard-layout-animation

but it's probably worth it to solve the black rectangle problem.


TheBB:

PR #8998, opened 2017-05-30
Add keybinding for calendar

Stebalien:
Bind "SPC aC" to `calendar`.

The `calendar` command displays a small popup calendar showing the previous,
current, and next month by default.

The primary drawback is that we will probably, eventually, add a full calendar
layer that may want to use the "SPC aC" keybinding.


TheBB:
`SPC aC` is used as a prefix for the slack layer.

https://github.com/syl20bnr/spacemacs/blob/develop/layers/+chat/slack/packages.el#L66-L71

Stebalien:
Damn. Any suggestions?

* `D` for date? (`d` is taken by dired).
* `T` for time? (`t` is taken by twitter).

PR #9007, opened 2017-06-01
counsel-find-file: Bind `find-file-other-window' to "j"

et2010:
By default, "M-o w" binding has already been used by ivy:

image

So we bind `find-file-other-window' to "M-o j" here.


TheBB:

PR #1417, opened 2015-04-28
Using worf is so much better

CestDiego:
[worf-mode](http://oremacs.com/worf/README.html) gives more vi-like
key bindings for org-mode :D evil-org on the other hand shadows some
long stablished vim keybindings :) I feel you should try it. there is
not really a conf for it so maybe just try it?

THIS DELETES EVIL-ORG

PR #9010, opened 2017-06-01
Remove evil-org for worf

TheBB:
This is a repurposing of https://github.com/syl20bnr/spacemacs/pull/1417, the oldest still-open pull request to Spacemacs. It removes evil-org and uses worf for much of its functionality. As a whole, worf has way higher quality and more functionality that we can bind in the future.

- `g{h,j,k,l}` still exist and work mostly the same way (a little bit more strict about levels)
- `g]` and `g[` added, they are pure next/previous operations that don't care about levels
- Change bindings for inserting headings to be more consistent, both with itself and evil.
  - Previous:
     - `o`: insert item below
     - `O`: insert heading below
     - `M-o`: insert sub-heading below
     - `M-t`: insert todo sub-heading below
     - `T`: insert todo heading below
  - New:
     - `o`: insert item below
     - `O`: insert item above
     - `M-o`: insert heading below
     - `M-O`: insert heading above
     - `M-t`: insert todo heading below
     - `M-T`: insert todo heading above
- Overriding `<`, `>` and `-` is gone. `M-h` and `M-l` cover the first two and `,-` the latter.
- Overriding `t` for `org-todo` is also gone, moved this to `,TT`.
- Add an org navigation transient state on `,n`.


TheBB:
Maybe put this on hold pending discussion in https://github.com/syl20bnr/spacemacs/pull/9011

Stebalien:
FWIW, I've applied this PR locally and find I much prefer worf over evil-org. evil-org to deviates from Evil/Vim in surprising and frustrating ways.

Stebalien:
Actually, things aren't quite as rosy as I thought. Most importantly, `[` and `]` are rebound to `worf-backward` and `worf-forward` so they can't be typed (WTF?).

Also, `o` (`insert-item-below`) can be very aggressive. For example, pressing `o` in either of the following states:

```org
* Item
  - A
█
```

```org
* Item
  - A

█
```

Opens a new item instead of opening a new line:

```org
* Item
  - A
  - |
```

(in the second case, the cursor goes *up*).

Stebalien:
Actually, on further inspection, I'm pretty sure worf is *not* intended to be used alongside evil. It adds vim-like features to org-mode but doesn't restrict these features to normal-mode (it doesn't integrate with evil at all).

TheBB:
Closing this for #9041

syl20bnr:
> On further inspection, I'm pretty sure worf is not intended to be used alongside evil. It adds vim-like features to org-mode but doesn't restrict these features to normal-mode (it doesn't integrate with evil at all).

Indeed this was my principal issue with worf (like with lispy), the implicit states (state changes depending on the position in the buffer) are the opposite philosophy of Vi. I believed I commented on the original PR but I guess that we discussed it on Gitter a few years ago instead.

syl20bnr:
@Stebalien thank you BTW ;-)

PR #3305, opened 2015-10-08
Make evil-org a local package

TheBB:

Issue #3294, opened 2015-10-08
org-capture (SPC o c) getting unbound in org mode


darkfeline:
My evil leader binding for `org-capture` on `SPC o c` is getting unbound in Org mode.

I'd appreciate any advice for debugging this, or if anyone else is running into this or can reproduce it.  It's a horrible heisenbug: it's fixed on a fresh start, but randomly `SPC o c` gets unset in Org mode.  I haven't found a way to reliably reproduce it.  I would update Spacemacs, update my packages, delete all package and reinstall them fresh, and this will keep happening randomly.

An errant timer?  Some delayed package loading?  Does Emacs/Spacemacs have anything that can help debug this?

EDIT: my binding of `org-agenda` on `SPC o a` is also getting unset.  I have bound a non-org command onto `SPC o s`, updated Spacemacs and packages, and will wait to see if I can pin this down.

robbyoconnor:
When reporting bugs: ALWAYS include`SPC hds`

if you are on `master`, try `develop` -- because if the bug does not exist there...this issue is moot.

robbyoconnor:
Also, if you are trying to get us to debug code, paste the code.

darkfeline:
#### System Info
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.0
- Spacemacs branch: master (rev. 4f3102c)
- Distribution: spacemacs
- Layers:
```elisp
((auto-completion :variables auto-completion-enable-sort-by-usage t)
 syntax-checking
 (shell :variables shell-default-shell 'shell shell-default-height 50)
 perspectives gtags erc org finance ibuffer
 (chrome :variables edit-server-new-frame nil)
 git version-control c-c++ emacs-lisp go html javascript latex lua markdown python ruby sql yaml mir-bbdb mir-elfeed mir-magit-annex mir-mu4e mir-plus mir-ranger mir-remind-to-diary mir-mpc)
```

Uh, I'm on `develop` though.  The "code", for what it's worth, is:

  (evil-leader/set-key "o c" 'org-capture)
  (evil-leader/set-key "o a" 'org-agenda)

TheBB:
Does it work if you exclude `evil-org`?

darkfeline:
@TheBB I'll try that, and great suggestion.

[...]

PR #9011, opened 2017-06-01
Update of `evil-org-mode`.

dschrempf:
Change `evil-org-mode` link to the new maintained GitHub repository.

Add the ability to process prefix arguments with `evil-org-eol-call`.

Use this ability to create list items with checkboxes with `clever-insert-item`.

I just saw that there is another pull request about `evil-org-mode` that asks for replacement with `worf`. I do not know `worf` but still provide this code in case it is needed.

Further, wouldn't it be better to use a submodule for `evil-org-mode` so that the spacemacs version does not diverge from the maintained one?

### DIFF
--- a/layers/+emacs/org/README.org
+++ b/layers/+emacs/org/README.org
@@ -40,7 +40,7 @@
 This layer enables [[http://orgmode.org/][org mode]] for Spacemacs.
 
 ** Features:
-- Vim inspired key bindings are provided by [[https://github.com/edwtjo/evil-org-mode][evil-org-mode]]
+- Vim inspired key bindings are provided by [[https://github.com/Somelauw/evil-org-mode][evil-org-mode]]
 - Nicer bullet via [[https://github.com/sabof/org-bullets][org-bullets]]
 - A [[https://cirillocompany.de/pages/pomodoro-technique][pomodoro method]] integration via [[https://github.com/lolownia/org-pomodoro][org-pomodoro]]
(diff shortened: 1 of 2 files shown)


TheBB:
> Further, wouldn't it be better to use a submodule for evil-org-mode so that the spacemacs version does not diverge from the maintained one?

If the idea was not to diverge from the official version, then we would be using the MELPA package. The reason we have a local copy is so that we *can* diverge, and the reason we wanted to diverge was essentially that evil-org was doing some very intrusive things, like making leader bindings that we had to restore. Eventually it just got easier to maintain a fork. See https://github.com/syl20bnr/spacemacs/issues/3294 and https://github.com/syl20bnr/spacemacs/pull/3305.

I see a lot of work has been done on evil-org lately by @Somelauw, which is great. I'll have to look at the current state of the package to see if we can't just remove our fork and go back to MELPA instead, or whether worf is still better. The original evil-org was quite frankly more trouble than it was worth.

Stebalien:
The default keybindings seem to have been tamed. `evil-org` has introduced ["key themes"](https://github.com/Somelauw/evil-org-mode/blob/master/doc/keythemes.org) for enabling/disabling sets of keybindings and the invasive ones appear to have been disabled by default.

TheBB:
Closing this for https://github.com/syl20bnr/spacemacs/pull/9041

Thanks though!

dschrempf:
Thanks for taking care of this!  Cheers.

PR #9014, opened 2017-06-01
[slack] Enable completion for slack-message-add-reaction (emoji)

swaroopch:
Tested with helm.


TheBB:
https://github.com/yuya373/emacs-slack/blob/master/slack-buffer.el#L60-L62

Why not use this??

swaroopch:
> https://github.com/yuya373/emacs-slack/blob/master/slack-buffer.el#L60-L62
> Why not use this?

Err, that is indeed better! Thanks for looking into it!

PR #9026, opened 2017-06-02
[slack] Add key binding for slack-message-remove-reaction

swaroopch:


TheBB:

PR #9027, opened 2017-06-02
Fix Swift REPL leader key bindings

sonelliot:
The `swift-mode` REPL bindings have been updated to use a different
naming convention (e.g. a prefix of `swift-mode:` in the CL style).
Typing `, s s` would trigger this error:

```
command-execute: Wrong type argument: commandp, swift-mode-run-repl
```

This change fixes up the command names so that they are usable again
from the leader key.


TheBB:

PR #9028, opened 2017-06-03
Update keyboard layout layer readme: text/images

duianto:
Text:
- Rewrite, add missing words, etc.
- Remove the begin and end verse block lines, but kept the verse block text,
  because Github doesn's show text in .org verse blocks.
- Change `_after_` to `*after*`, because Github doesn't show underscores in .org
  files.
- Add the Colemak, Dvorak and Workman layouts vim movement key bindings, and the
  key bindings for the lost (overwritten) keys.
- Fix the image link to the workman layout image, it was missing the square
  brackets.
- Link replaced: The workmanlayout.com domain name seems to have been lost.
  Google Chrome warns "Your connection is not private", and the advanced
  information states that the security certificate is from a "just host" domain.
  The most stable replacement link seems to be the workman section on the
  keyboard layout wikipedia page.
- Indicate the images that have been modified, as suggested by the CC BY-SA
  Attribution section.

Images:
- Add the dvorak simplified layout.
- Fix the programmer dvorak layout, the transparent key faces made it almost
  invisible when viewed in Spacemacs with a dark background.
- Rename images for type, consistency, and most common usage:
  use logo instead of layout for logo names, and use layout for all layout
  images, instead of keymap for some and layout for others. Layout is the most
  common name, wikipedia even redirects keymap to the keyboard layout page.
- Reduce their width to 640px, so that they fit within column 80.


TheBB:

The part about adding the layer to the list in dotspacemacs is part of the README.org template so I rejected that change. It's phrased in this (admittedly awkward) way because people keep misunderstanding it. If we want to change that we'll have to do it for all the layers.

Issue #9006, opened 2017-05-31
[Severe] OSX Layer Breaks Page Up/Down!


ghost:
#### Description :octocat:
Function + up/downarrow to do Page up/page down doesn't work AT ALL when the "osx" layer is enabled.

#### Reproduction guide :beetle:
- Enable the "osx" layer on latest develop
- Start Emacs on macOS (I'm using the Railwaycat port).
- Press fn+up/down (for page up/page down)

*Observed behaviour:* :eyes: :broken_heart:
The regular downarrow/uparrow gets triggered and just moves a single line.

*Expected behaviour:* :heart: :smile:
It should obviously move a whole page (since the keys are page up/page down). If I disable the "osx" layer, the page up/page down keys work again. This bug was added pretty recently (I just updated Spacemacs and ALL packages for the first time in a few months).

#### System Info :computer:
- OS: darwin
- Emacs: 25.2.1
- Spacemacs: 0.200.9
- Spacemacs branch: develop (rev. 0e0e90e5)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(asm c-c++ csv emacs-lisp html javascript lua markdown org php python shell-scripts sql vimscript windows-scripts helm
     (auto-completion :variables auto-completion-return-key-behavior 'complete auto-completion-tab-key-behavior 'cycle auto-completion-enable-help-tooltip nil auto-completion-enable-snippets-in-popup t auto-completion-private-snippets-directory
                      '("~/.spacemacs.d/snippets/" "~/.spacemacs.d/snippets-default/"))
     ibuffer imenu-list
     (git :variables git-magit-status-fullscreen t)
     (colors :variables colors-colorize-identifiers 'all colors-default-rainbow-identifiers-sat 50 colors-default-rainbow-identifiers-light 65)
     (shell :variables shell-default-shell 'ansi-term shell-default-term-shell "/bin/bash" shell-default-height 30 shell-default-position 'bottom)
     semantic syntax-checking)
```
- System configuration features: NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS MODULES

SteveJobzniak:
I found the exact bug commit:

https://github.com/syl20bnr/spacemacs/commit/5ff0fbe6d9caa07796d40d066a74e5230e46d33d

[...]

PR #9031, opened 2017-06-03
OSX: Fix the function keys not working correctly

fiveNinePlusR:
page up and page down were not working correctly and this allows the normal emacs handling of system function keys.

Fixes bug #9006

### DIFF
--- a/layers/+os/osx/config.el
+++ b/layers/+os/osx/config.el
@@ -35,10 +35,10 @@
    Default: `meta'.
    For backwards compatibility the variable `osx-use-option-as-meta'
    takes precedence is set to t.")
-(defvar osx-function-as 'none
+(defvar osx-function-as nil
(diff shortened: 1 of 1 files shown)


layers/+os/osx/config.el
```diff
@@ -35,10 +35,10 @@
    Default: `meta'.
    For backwards compatibility the variable `osx-use-option-as-meta'
    takes precedence is set to t.")
-(defvar osx-function-as 'none
+(defvar osx-function-as nil
   "Sets the key binding of the `FUCTION' key on OSX.
```
duianto: The missing "N" in `FUCTION` can probably also be fixed when the documentation is updated for this PR.
With that change, the first line in the doc-string becomes:
"Sets the key binding of the `FUNCTION' key on OSX.

sdwolfz:
Thank you :heart:!

PR #9033, opened 2017-06-04
Fixed bug so hy commands work when hy is installed on a virtualenv

dannyfreeman:
Hy is typically installed on a python virtual environment, not globally. When
that is the case, the init-hy-mode function doesn't set any leader keys for hy 
since it can't find the hy executable on the path. When a new virtual 
environment is activated, init-hy-mode is run again to see if it can find a hy
executable and setup the leader keys for hy.

### DIFF
--- a/layers/+lang/python/packages.el
+++ b/layers/+lang/python/packages.el
@@ -214,6 +214,7 @@
     :defer t
     :init
     (progn
+      (add-hook 'pyvenv-post-activate-hooks 'python/init-hy-mode)
       (pcase python-auto-set-local-pyvenv-virtualenv
         (`on-visit
          (spacemacs/add-to-hooks 'spacemacs//pyvenv-mode-set-local-virtualenv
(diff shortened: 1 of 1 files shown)


dannyfreeman:
Thanks for the feedback. Just to clarify, the keys need to be set in that different function as well. So at this point, the only thing the init-hy-mode function will be doing is setting up the hooks so that the keys can be set when the hy executable can be found. Does that sound right?

TheBB:
No, the keys can be set in the original function, just straightforward without checking for a hy executable or anything.

TheBB:

I made some changes in https://github.com/syl20bnr/spacemacs/commit/16f25488a90c68dbb6fc7d6b47876656daf1be78 and https://github.com/syl20bnr/spacemacs/commit/ed90f2b4ff652991599f9688a9469aabd09da941 bringing this solution in line with the existing solutions (for the python interpreter and syntax checkers). They should work the same way now. Let me know if I messed something up.

dannyfreeman:
I made a comment on #16f2548 about your changes. One of them does mess something up.

PR #9035, opened 2017-06-05
Fix typo

edwargix:
Fixed typo in [layers/+completion/helm/local/helm-spacemacs-help/helm-spacemacs-help.el](https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Bcompletion/helm/local/helm-spacemacs-help/helm-spacemacs-help.el). These are the only occurences of the function name.


TheBB:

PR #9041, opened 2017-06-06
Use evil-org from MELPA

TheBB:
Supplants https://github.com/syl20bnr/spacemacs/pull/1417, https://github.com/syl20bnr/spacemacs/pull/9010 and https://github.com/syl20bnr/spacemacs/pull/9011

Evil-org has been heavily overhauled and looks like a much better package now. It has good defaults and is much easier for the end-user to customize. Oh, and it's maintained.

I still removed that silly `t` binding though. :-)

cc @Stebalien @dschrempf


TheBB:
Seems ok, so I merged it.

ccaseiro:
Why remove the `t` binding?

TheBB:
My reasoning is

* single-shortcut keys are few in number and valuable
* the vim/evil defaults for those keys are designed for editable modes, and org is an editable mode
* practically nowhere else in Spacemacs are such keys overridden
* the benefit of making an exception in this case is dubious
* it's not too difficult to configure on a user level

For sure, many people will be toggling todo state more often than using the `t` jump operator while in org-mode, but I'm not convinced that this is worth violating POLA in org-mode. But of course if there is consensus that I'm wrong then I'm happy to be convinced otherwise.

Stebalien:
I'm strongly against remapping `t` as I do use it for navigation but, then again, I have `SPC` free so I just mapped that for toggling (I'm using CapsLock as my leader).

If you're going to be doing a lot of navigation/toggling, you might consider a hydra:


```elisp
    (defun org-toggle-element ()
      (interactive)
      (if (org-at-item-checkbox-p)
          (org-toggle-checkbox)
        (org-todo)))

    (defhydra hydra-org-nav (:exit nil :columns 5)
      "Org Navigate"

      ("h" (org-up-element) "Parent")
      ("j" (org-forward-element) "Next Sibling")
      ("k" (org-backward-element) "Previous Sibling")
      ("l" (org-down-element) "Child")

      ("t" (org-toggle-element) "Toggle")

      ("H" (org-promote-subtree) "Promote")
      ("J" (org-move-subtree-down) "Move Down")
      ("K" (org-move-subtree-up) "Move Up")
      ("L" (org-demote-subtree) "Demote")

      ("q" nil "quit" :exit t)

      ("d" (org-cut-subtree) "Cut Subtree")
      ("y" (org-copy-subtree) "Copy Subtree")
      ("p" (org-paste-subtree) "Copy Subtree"))

      "j" #'hydra-org-nav/body))
```

duianto:
@Stebalien The text "Copy Subtree" for last key "p", should probably be changed to "Paste Subtree".

PR #9004, opened 2017-05-31
.spacemacs template docstring revision

wygulmage:
I edited the docstrings of the main functions of the .spacemacs template for grammar and brevity. It makes more sense to me, so I figured I'd share it.

PR #9042, opened 2017-06-06
Revise .spacemacs docstrings

wygulmage:
Tighten up the docstrings for initialization and configuration functions. I realize that's pretty subjective, so please only use what makes sense.
Below is a break-down to avoid seeming _too_ arbitrary.

Rationale
---------
- Headings: We know they're functions, so we don't need to say 'X function'. 'Layer configuration' is called that in the other functions.

- Instructions: Say what I should do.

- Description of `dotspacemacs/init': 'very beginning' is the more common English idiom.

If I sound strident, it's just 'cause I'm trying to be terse.
This is all opinion so YMMV. I hope it helps.
______
This is a correction of PR #9004 (closed).

### DIFF
--- a/core/templates/.spacemacs.template
+++ b/core/templates/.spacemacs.template
@@ -3,9 +3,8 @@
 ;; It must be stored in your home directory.
 
 (defun dotspacemacs/layers ()
-  "Configuration Layers declaration.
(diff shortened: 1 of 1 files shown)


TheBB:

Issue #9037, opened 2017-06-05
How to enable auto-complete for sql layer


spacegoing:
<<REPLACE ALL UPPERCASE EXPRESSIONS AND DELETE THIS LINE, `C-c C-c` TO SUBMIT>>
#### Description :octocat:
I'm using SQL layer and having trouble to enable `auto-complete-mode` when `SQL` enabled.

Even after I enabled global mode by adding the following code under `user-config()`

#### Reproduction guide :beetle:
  under user-config() add:
  ;; Add company mode
  (require 'auto-complete)
  (global-auto-complete-mode t)

*Observed behaviour:* :eyes: :broken_heart:
The completion menu still don't show up.

*Expected behaviour:* :heart: :smile:
I don't require sql keywords showing up. Even simple text completion will be much helpful.

#### System Info :computer:
- OS: darwin
- Emacs: 25.2.1
- Spacemacs: 0.200.9
- Spacemacs branch: master (rev. 8e1af145)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(ruby yaml octave auto-completion better-defaults emacs-lisp git semantic github sql c-c++ python html javascript org markdown
      (markdown :variables markdown-live-preview-engine 'vmd)
      pandoc latex
      (latex :variables latex-enable-auto-fill t)
      (latex :variables latex-build-command "LaTeX")
      (latex :variables latex-enable-folding t)
      spell-checking syntax-checking)
```
- System configuration features: NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS MODULES

swaroopch:
1. Please paste output of `SPC h I` so that the community can assist you better.
2. Add the `auto-completion` layer?

spacegoing:
@swaroopch tks for your help! I've updated my question. Would you please show me how to "add the auto-completion layer"?

swaroopch:
It looks like you do have the auto-completion layer already installed as per your config.

Looking at https://github.com/syl20bnr/spacemacs/tree/develop/layers/%2Blang/sql - I'm not sure there is any auto-completion support for SQL. Maybe someone more familiar with the SQL layer can comment on it.

swaroopch:
Try `alt-/` for simple completion.

spacegoing:
@swaroopch Many thanks! `alt-/` did the work! 

[...]

PR #9045, opened 2017-06-07
Adds dumb autocomplete for sql

sdwolf:
Refs #9037.

I know about the existance of https://github.com/proofit404/company-edbi but I am not sure how exactly that would be integrated. So I'm adding the dumb autocomplete until someone has the willpower to try that one out.


TheBB:

Issue #8804, opened 2017-04-26
Perl 5 (Not Perl 6) Layer


CeleritasCelery:
I originally opened #7758, but is was mistakenly assumed to be a duplicate of #5019 which is for Perl 6. Perl 5 and Perl 6 are two different languages. So don't erroneously close this one. I am opening this issue so that those interested in a Perl 5 layer can input what they would like included. #8572 is an attempt at Perl 5 layer but is too messy to be merged in its current state. 

Some of the packages that I would like to include would be 
[PLSense](https://github.com/aki2o/emacs-plsense) - auto-completion
[dumb jump](https://github.com/jacktasia/dumb-jump) - jump to definition
[perl-flycheck](http://www.flycheck.org/en/latest/) - linting and syntax checking
[helm perldoc](https://github.com/syohex/emacs-helm-perldoc) - perldoc browsing
[Perlnow](https://github.com/doomvox/perlnow) - improved perl debugger
[PDE](https://github.com/wenbinye/emacs-pde) - useful perl utilities 

We could also include a fix for #480 

Once we have an idea of what we want, I can put together a PR. Hopeful once #8686 is resolved it can actually get merged.

vibrys:
I'm missing perl5 layer.
Looks like the progress on this one has been stopped for a while.

Could You please suggest any other perl5 usable layer that could be checkout as private one for a while?

CeleritasCelery:
I have a PR open for it @vibrys

TheBB:
Ay, sorry about closing that PR earlier. I must have had a serious brain fart.

vibrys:
@CeleritasCelery, this is great news! I just saw TheBB approved the changes, although these still didn't land into develop branch.

bersace:
Nice addition ! Can't wait for next release :D

CeleritasCelery:
Also for this interested, I have created a company backend for perl5 based the plsense server. You can find it [here](https://github.com/CeleritasCelery/company-plsense). I am going to also create a Spacemacs PR to add it to this layer.

PR #8572, opened 2017-03-23
layers/+lang/perl5

abelew:
perl5 layer using Joseph Brenner's perlnow:
http://obsidianrook.com/perlnow/
along with some minor mmm-mode tweaks for mason, and
bioperl-mode with some minor changes because it has not been maintained very much since 2008.
https://github.com/bioperl/bioperl-live/tree/master/ide/bioperl-mode

This does still use template as it is used by perlnow and I did not wish to mess with that too much.

Issue #480, opened 2015-01-25
cperl-mode + smartparens issue


alexrkopp:
Hello,

If you switch into cperl-mode and type the following:

```
if (0) {
```

It will be automatically changed to

```
if (0) {}}
```

Note the two closing curly braces. Switching into perl-mode and typing that same line yields proper behavior.

Any ideas?

alexrkopp:
Anyone have any idea what is causing this?

jasonblewis:
I see this behaviour too. I think its related to smartparens as I don't use spacemacs

spacebat:
I think its a conflict between cperl's electric braces and related options, and smartparens' auto pairing. I've disabled all of cperl's electric settings.

Fuco1:
It might very well be what @spacebat describes.  Smartparens is quite aggresive in its own enforcing rules, since most of the time the "electric" stuff is quite, well, idiotic.  The apparently random order of execution of hooks doesn't help one bit either... it is most recomended to turn off any electric stuff while using SP.

pasja:
My quick&dirty solution:
`(define-key cperl-mode-map "{" 'nil)`

TheBB:
Since Spacemacs doesn't have a layer for this mode, and since there's an easy workaround from @pasja, I am liable to close this. If the time comes when Spacemacs has official Perl support we can revisit.

For future visitors, the workaround should be (in `user-config`):

```elisp
(with-eval-after-load 'cperl-mode
  (define-key cperl-mode-map "{" nil))
```

benlind:
I recently had this problem. I have `electric-pair-mode` on with `cperl-mode`, and I was getting `{}}` when typing `{` on an empty line.

This solved my problem:

    (add-hook 'cperl-mode-hook (lambda () (local-unset-key (kbd "{"))))

PrgrmAtCeleritas:
This solution will allow `electric-pair-mode` to still work when smartparens is toggled off

```
(with-eval-after-load 'cperl-mode
  (add-hook 'smartparens-enabled-hook  (lambda () (define-key cperl-mode-map "{" nil)))
  (add-hook 'smartparens-disabled-hook  (lambda () (define-key cperl-mode-map "{" 'cperl-electric-lbrace))))
```

PR #9046, opened 2017-06-07
Perl5 layer

CeleritasCelery:
This is a basic perl5 layer for Spacemacs. This is in response to #8804 and address the bug in #480. THis layer adds flycheck support as well as setting better defaults and fixes some known bugs. There is currently an open PR for Perl5 at #8572 but this one should be used in favor of that one because 
- 5 files changed instead of 82. Most of changes in #8572 are unrelated to perl.
- only uses packages available on MELPA
- bioperl is a narrow use case and should not be included in the default layer
- no merge conflicts


layers/+lang/perl5/packages.el
```diff
@@ -0,0 +1,117 @@
...
+      ;; Use less horrible colors for cperl arrays and hashes
+      (set-face-attribute 'cperl-array-face nil :foreground  "#DD7D0A"    :background  'unspecified :weight 'unspecified)
+      (set-face-attribute 'cperl-hash-face nil  :foreground  "OrangeRed3" :background  'unspecified :weight 'unspecified)
+      (setq cperl-highlight-variables-indiscriminately t)
+
+      ;; default settings
+      (setq cperl-indent-level 4) ;; 4 spaces is the standard indentation
+      (setq cperl-close-paren-offset -4) ;; indent the closing paren back four spaces
+      (setq cperl-continued-statement-offset 4) ;; if a statement continues indent it to four spaces
+      (setq cperl-indent-parens-as-block t) ;; parentheses are indented with the block and not with scope
```
TheBB: Would be good to make these settings in `:init` so that they can at least be overridden more easily by the user.

layers/+lang/perl5/packages.el
```diff
@@ -0,0 +1,117 @@
...
+
+      ;; tab key will indent all marked code when pressed
+      (add-hook 'cperl-mode-hook
+                (lambda () (local-set-key (kbd "<tab>") 'indent-for-tab-command)))
```
TheBB: Is this really necessary?

layers/+lang/perl5/packages.el
```diff
@@ -0,0 +1,117 @@
...
+
+      (font-lock-add-keywords 'cperl-mode
+                              '(("\\_<const\\|croak\\_>" . font-lock-keyword-face)))
+      (font-lock-add-keywords 'cperl-mode
+                              '(("\\_<say\\|any\\_>" . cperl-nonoverridable-face))))))
+
+(defun perl5/post-init-smartparens ()
+  :config
```
TheBB: `:config` works in `use-package` calls, it's not something you can put just anywhere. Here you can just remove it.

layers/+lang/perl5/packages.el
```diff
@@ -0,0 +1,117 @@
...
+  ;; fix a bug with electric mode and smartparens https://github.com/syl20bnr/spacemacs/issues/480
+  (with-eval-after-load "cperl-mode"
+    (add-hook 'smartparens-enabled-hook  (lambda () (define-key cperl-mode-map "{" nil)))
+    (add-hook 'smartparens-disabled-hook  (lambda () (define-key cperl-mode-map "{" 'cperl-electric-lbrace)))))
+
+(defun perl5/post-init-flycheck ()
+  (spacemacs/add-flycheck-hook 'cperl-mode))
+
+(defun perl5/init-perl-completion ()
+  (use-package perl-completion
```
TheBB: I really, really recommend that we turn this package into a company source instead. Auto-complete-mode is sort of a dud package in Spacemacs, everything uses company, and the assumption is that everything should keep using it. I don't think it should be too hard to wrap it.

CeleritasCelery: I would love that but there is currently no Company backend for Perl. see [company-mode/company-mode#666](https://github.com/company-mode/company-mode/issues/666) where I am looking to create one. I can remove the autocompletion until a company backend is available if that is the preferred solution.

layers/+lang/perl5/packages.el
CeleritasCelery: I added this back in Spacemacs 0.105.11 because hitting tab key would not indent the code properly and default to some cperl function. However I removed the above code and everything seems to work fine on the latest version. I will go ahead and remove it.

layers/+lang/perl5/packages.el
TheBB: That might be good. I had a look at the package, and it's a little disheartening. It has received no updates since 2009 and has an undeclared dependency on MELPA (to the somewhat archaic `anything` package). So actually as it stands the autocompletion doesn't work as advertised. When patched up it does deliver completion candidates though.

CeleritasCelery: Okay I will go ahead and remove it. That way we will have a minimally viable layer. I am going to take some time in the next several weeks and see if I can move it over company. I will do a PR then.

Also it does not have be called manually. It will automatically provide completion candidates while typing. Despite its age and lack of maintenance is actually works quite well. Best solution I have found so far.

TheBB:

PR #9048, opened 2017-06-07
Add key bindings to spacemacs-base

beta1440:
Add key bindings for goto-line and zap-to-char

### DIFF
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -1950,9 +1950,11 @@ The ~SPC j~ prefix is for jumping, joining and splitting.
 | ~SPC j 0~   | go to the beginning of line (and set a mark at the previous location in the line) |
 | ~SPC j $~   | go to the end of line (and set a mark at the previous location in the line)       |
 | ~SPC j b~   | undo a jump (go back to previous location)                                        |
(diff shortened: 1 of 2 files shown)


Beta1440:
I realized that "Ng" might be a better key binding for goto-line

Beta1440:
May I ask which documentation you are referring to? From what I understand, the `spacemacs-base` layer doesn't have a README like other layers do.

TheBB:
DOCUMENTATION.org in the root directory serves as the documentation for all the `spacemacs-*` layers.

Beta1440:
I've added the keybindings to the DOCUMENTATION.org file

Beta1440:
I'm planning on adding more keybindings to spacemacs-base on this pull request so that I can squash them all together.

Beta1440:
I've done adding key bindings

TheBB:
Are these necessary for emacs editing style users? Most of the bindings feel quite superfluous from the POV of an evil user.

Beta1440:
It's true that most of these keys are basic.Emacs has bindings for most of them, but I'm adding these key bindings since it makes it easier for beginners to discover them. I want to reduce the need to remember obscure key bindings.

PR #9053, opened 2017-06-08
Fix lazy insert point

sdwolf:
A recent change in the `.spacemacs.template` made it so that the
`dotspacemacs-configuration-layers` variable name can be found inside a
comment right before the expression that sets it's value. This makes the
lazy insertion fail to detect the proper place to add lazy loaded
layers.

This fix solves the immediate problem but maybe a better solution can
be found.


TheBB:

Issue #8823, opened 2017-04-28
evil-paste-pop doesn't work, previous command was not an evil-paste


duianto:
#### Description :octocat:
Replacing the just pasted text (by pressing `p`) with a previous entry from the kill ring (by pressing `C-p`) doesn't work.

#### Reproduction guide :beetle:
1. Start Emacs
2. Open the `*scratch*` buffer `SPC b s`
3. Type: `two words`
4. Move the cursor to the first character `gg`
5. Yank both words: press `yww` and `yw`
6. Open a newline below: `o`
7. Exit insert state: `Esc`
8. Paste the last yanked text: `p`
9. Try to replace it with a previous item from the kill ring: `C-p`

*Observed behaviour:* :eyes: :broken_heart:
The `*Messages*` buffer shows:
`user-error: Previous command was not an evil-paste: spacemacs/evil-mc-paste-after`

*Expected behaviour:* :heart: :smile:
A previous entry in the kill ring should replace the just pasted text.

#### Possible cause
`evil-paste-pop` (`C-p`) expects that the `last-command` was any of these 3:
```elisp
  (unless (memq last-command
                '(evil-paste-after
                  evil-paste-before
                  evil-visual-paste))
```
https://github.com/emacs-evil/evil/blob/f0684c1f6644a8c4e0e5fce50a1a04796555eaf1/evil-common.el#L2598

But `p` is bound to: `spacemacs/evil-mc-paste-after`. And the same thing applies to: `]p` which is bound to `evil-unimpaired/paste-below`, it shows this message:
`user-error: Previous command was not an evil-paste: evil-unimpaired/paste-below`

#### When it does work
Replacing the last pasted text with `C-p` works, if there was an active selection when the text was pasted with either `p` or `]p`.

#### System Info :computer:
- OS: windows-nt
- Emacs: 25.1.1
- Spacemacs: 0.200.9
- Spacemacs branch: develop (rev. 177da959)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(autohotkey markdown html ranger emacs-lisp git github helm org syntax-checking version-control)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG SOUND DBUS NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS

MHOOO:
I can confirm this bug happening on the develop branch

[...]

PR #9056, opened 2017-06-08
Fix issue #8823: evil-paste-pop doesn't work `previous command was not an evil paste.`

nochiel:
Fix issue #8823.

[evil-common.el: evil-paste-pop](https://github.com/emacs-evil/evil/blob/master/evil-common.el#L2604), has the following check:

```
(unless (memq last-command
                '(evil-paste-after
                  evil-paste-before
                  evil-visual-paste))
    (user-error "Previous command was not an evil-paste: %s" last-command))
```
Calls to [spacemacs/evil-mc-paste-after](https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Bspacemacs/spacemacs-evil/funcs.el#L87) and [spacemacs/evil-mc-paste-before](https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Bspacemacs/spacemacs-evil/funcs.el#L94) thus break `evil-paste-pop`.  A more idiomatic way to create this functionality is to use [elisp advice](https://www.gnu.org/software/emacs/manual/html_node/elisp/Advising-Functions.html).

### DIFF
--- a/layers/+spacemacs/spacemacs-evil/funcs.el
+++ b/layers/+spacemacs/spacemacs-evil/funcs.el
@@ -85,15 +85,17 @@ Otherwise, revert to the default behavior (i.e. enable `evil-insert-state')."
            (eq (evil-mc-get-cursor-count) 1))))
 
 (defun spacemacs/evil-mc-paste-after (&optional count register)
-  "Disable paste transient state if there is more that 1 cursor."
+  "Disable paste transient state if there is more than 1 cursor."
   (interactive "p")
+  (setq this-command 'evil-paste-after)
   (if (spacemacs//paste-transient-state-p)
       (spacemacs/paste-transient-state/evil-paste-after)
     (evil-paste-after count (or register evil-this-register))))
 
 (defun spacemacs/evil-mc-paste-before (&optional count register)
-  "Disable paste transient state if there is more that 1 cursor."
+  "Disable paste transient state if there is more than 1 cursor."
   (interactive "p")
+  (setq this-command 'evil-paste-before)
   (if (spacemacs//paste-transient-state-p)
       (spacemacs/paste-transient-state/evil-paste-before)
     (evil-paste-before count (or register evil-this-register))))
--- a/layers/+spacemacs/spacemacs-evil/local/evil-unimpaired/evil-unimpaired.el
+++ b/layers/+spacemacs/spacemacs-evil/local/evil-unimpaired/evil-unimpaired.el
@@ -54,11 +54,13 @@
 
 (defun evil-unimpaired/paste-above ()
   (interactive)
+  (setq this-command 'evil-paste-after)
   (evil-insert-newline-above)
   (evil-paste-after 1))
 
 (defun evil-unimpaired/paste-below ()
   (interactive)
+  (setq this-command 'evil-paste-after)
   (evil-insert-newline-below)
   (evil-paste-after 1))
 


layers/+spacemacs/spacemacs-evil/funcs.el
```diff
@@ -84,16 +84,20 @@ Otherwise, revert to the default behavior (i.e. enable `evil-insert-state')."
...
            (eq (evil-mc-get-cursor-count) 1))))
 
-(defun spacemacs/evil-mc-paste-after (&optional count register)
-  "Disable paste transient state if there is more that 1 cursor."
-  (interactive "p")
-  (if (spacemacs//paste-transient-state-p)
-      (spacemacs/paste-transient-state/evil-paste-after)
-    (evil-paste-after count (or register evil-this-register))))
+(defun spacemacs/evil-mc-paste-after (fn &rest args)
+   "Disable paste transient state if there is more that 1 cursor."
```
duianto: The extra space before these lines should probably be removed. Two spaces seem to be default. At least when one auto indents the paragraph with `evil-indent`: `= i p`. The `spacemacs/evil-mc-paste-before` function only has two spaces, so they should probably match. Otherwise these lines will show up unexpectedly (for a git beginner), if the whole file is auto indented.

Additionally there's a small typo in the doc-string. 
"that" should be changed to "than" resulting in:
"Disable paste transient state if there is more than 1 cursor."
the same typo occurs in both functions.

Or should the typo be fixed in a separate PR?

layers/+spacemacs/spacemacs-evil/funcs.el
```diff
@@ -84,16 +84,20 @@ Otherwise, revert to the default behavior (i.e. enable `evil-insert-state')."
...
+   (interactive "p")
+   (if (spacemacs//paste-transient-state-p)
+       (spacemacs/paste-transient-state/evil-paste-after)
+     (apply fn (first args) (or (second args) evil-this-register) (last args))))
 
-(defun spacemacs/evil-mc-paste-before (&optional count register)
+(advice-add 'evil-paste-after  :around  #'spacemacs/evil-mc-paste-after)
```
duianto: There are two spaces before and after `:around`, is this just a style choice to make it more readable, or should there only be one? The same thing applies to the evil-paste-before advice below.
Searching through the .emacs.d folder found 5 other occurrences of `advice-add` ... `:around`, and they only have 1 space around them.

duianto:
`C-n` and `C-p` now cycles through the kill ring after pasting with `p` or `P` 👍.

It does however not work after pasting with `[p` `evil-unimpaired/paste-above` or `]p` `evil-unimpaired/paste-below`, but they might not have worked before this issue occurred. They probably need to be fixed in a separate PR.

There's also a hiccup when cycling through the kill ring, it seems like when it reaches the end or beginning of the kill ring, then nothing happens until a second `C-n` or `C-p` is pressed. But I'm guessing that this also is unrelated to this issue and might even need to be fixed in the evil package.

bmag:
> A more idiomatic way to create this functionality is to use elisp advice.

Using advices isn't idiomatic, it's more of a last resort. It might be merited here, I'm not sure.

The calls to `advice-add` should be in the :init block of `spacemacs-evil/init-evil-mc`, that way they don't apply when `evil-mc` is excluded.

Should use `funcall` with `count` and `register` instead of `apply` with `(first args)`, `(second args)` and `(last args)`. Function signatures should change accordingly from `(fn &rest args)` to `(fn &optional count register)`.

@duianto thanks for testing this

layers/+spacemacs/spacemacs-evil/funcs.el
bmag: > should the typo be fixed in a separate PR?

Either as part of this one or a separate PR, doesn't really matter.

nochiel:
> It does however not work after pasting with [p evil-unimpaired/paste-above or ]p evil-unimpaired/paste-below, but they might not have worked before this issue occurred. They probably need to be fixed in a separate PR.

A number of Evil extensions also break `evil-paste-pop` in the same way that `spacemacs-evil` does.  Thus far, I count:
1. `evil-unimpaired`
2. `evil-iedit`
3. `evil-mc`

The correct fix would seem to be to write a new `evil-yank-handler` and/or setq `evil-last-paste` with sane values after the paste wrapper or yank handler runs (see, for example, what [evil-cleverparens](https://github.com/luxbock/evil-cleverparens/blob/82c920ba04accfd31fa292e11c454d5112b4fd51/evil-cleverparens.el#L419) does).  Currently it's too easy to break `evil-paste-pop` with the result that almost every extension that wraps it breaks it.  Alternatively, all these evil extensions could be PR'd with changes to their paste functions to `defadvice` them instead of `defun`.  But @bmag's insistence that advice is *not* idiomatic is given some credence by the elisp documentation which says that advice can lead to hard-to-diagnose bugs. (I happen to love advice and use it rather liberally in my own fiddling.)

> There's also a hiccup when cycling through the kill ring, it seems like when it reaches the end or beginning of the kill ring, then nothing happens until a second C-n or C-p is pressed. But I'm guessing that this also is unrelated to this issue and might even need to be fixed in the evil package.

Initially, I assumed that this was due to  [current-kill](https://www.gnu.org/software/emacs/manual/html_node/eintr/Code-for-current_002dkill.html#Code-for-current_002dkill) (used by `evil-paste-pop`):
>> Rotate the yanking point by N places, and then return that kill.      
>> If N is zero and `interprogram-paste-function' is set to a function that returns a string or a list of strings, and if that function doesn't return nil, then that string (or list) is added  to the front of the kill ring and the string (or first string in the list) is returned as the latest kill.

But a cursory test seems to indicate that this isn't the case (`yank-pop` seems to do the same thing). More testing required.

nochiel:
> The calls to advice-add should be in the :init block of spacemacs-evil/init-evil-mc, that way they don't apply when evil-mc is excluded.

Done.

> Should use funcall with count and register instead of apply with (first args), (second args) and (last args). Function signatures should change accordingly from (fn &rest args) to (fn &optional count register).

Done

nochiel:
Closing this and creating a new PR because there's a better simpler fix: ddd5499bc1aaeeae50fcedb7420229ccc42cdf9b

PR #9057, opened 2017-06-09
[ipython-notebook] Change prefix from "ai" to "ay"

swaroopch:
Because "ai" is already being used for IRC (ERC)!

It's weird to say ipython-notebook stuff under "IRC"


TheBB:
`"an"` is used for deft though. :-/

Would be good to figure out once and for all what to do in such cases.

swaroopch:
Err, is there a catalogue of used key bindings? :)

swaroopch:
Switching to `ay`, I believe it is unused...

TheBB:

We still need to figure out how to fix this problem of colliding bindings in different layers, though.

PR #9067, opened 2017-06-10
Require avy in `spacemacs-navigation`

jonasws:
When using the `o` keybinding to open links with `ace`, Spacemacs would complain that the `avy-with` symbol's definition was void. This PR fixes that by requiring `avy` explicitly, as it starts to work later on, as then the `avy` package appears to have been loaded.


TheBB:

PR #9068, opened 2017-06-10
string-inflection: group which-keys, sort code/doc

duianto:
Group together the which-key entries that call the same command:
k and - becomes k,-
u and _ becomes u,_

Declare a prefix name for SPC x i, this changes the SPC x entry:
from: "i -> +prefix"
to: "i -> +inflection".

Capitalize the transient state title, so that it matches the other TS titles.

Sort the code and documentation key bindings alphabetically.


TheBB:

PR #9069, opened 2017-06-10
Fix a missing package name in mu4e documentation

nashamri:
There was a missing package name in mu4e that messed up the documentation in the **Maildirs extension** subsection.


TheBB:
@

PR #9073, opened 2017-06-11
Fix issue #8823: evil-paste-pop doesn't work `previous command was not an evil paste.`

nochiel:
@duianto, @bmag. A simpler fix than was [previously proposed](https://github.com/syl20bnr/spacemacs/pull/9056): Use `this-command` (ref. [emacs: command loop info](https://www.gnu.org/software/emacs/manual/html_node/elisp/Command-Loop-Info.html)) to ensure `evil-paste-pop` works correctly after `spacemacs/evil-mc-paste-{after,before}` and `evil-unimpaired/paste-{above,below}`.

### DIFF
--- a/layers/+spacemacs/spacemacs-evil/funcs.el
+++ b/layers/+spacemacs/spacemacs-evil/funcs.el
@@ -85,15 +85,17 @@ Otherwise, revert to the default behavior (i.e. enable `evil-insert-state')."
            (eq (evil-mc-get-cursor-count) 1))))
 
 (defun spacemacs/evil-mc-paste-after (&optional count register)
-  "Disable paste transient state if there is more that 1 cursor."
+  "Disable paste transient state if there is more than 1 cursor."
   (interactive "p")
+  (setq this-command 'evil-paste-after)
   (if (spacemacs//paste-transient-state-p)
       (spacemacs/paste-transient-state/evil-paste-after)
     (evil-paste-after count (or register evil-this-register))))
 
 (defun spacemacs/evil-mc-paste-before (&optional count register)
-  "Disable paste transient state if there is more that 1 cursor."
+  "Disable paste transient state if there is more than 1 cursor."
   (interactive "p")
+  (setq this-command 'evil-paste-before)
   (if (spacemacs//paste-transient-state-p)
       (spacemacs/paste-transient-state/evil-paste-before)
     (evil-paste-before count (or register evil-this-register))))
--- a/layers/+spacemacs/spacemacs-evil/local/evil-unimpaired/evil-unimpaired.el
+++ b/layers/+spacemacs/spacemacs-evil/local/evil-unimpaired/evil-unimpaired.el
@@ -54,11 +54,13 @@
 
 (defun evil-unimpaired/paste-above ()
   (interactive)
+  (setq this-command 'evil-paste-after)
   (evil-insert-newline-above)
   (evil-paste-after 1))
 
 (defun evil-unimpaired/paste-below ()
   (interactive)
+  (setq this-command 'evil-paste-after)
   (evil-insert-newline-below)
   (evil-paste-after 1))
 


sdwolf:
Note that you don't need to close PRs when you add commits to your branch. When you `git push` tne commits on github they automaticall appear on the PR associated to your branch.

nochiel:
@sdwolf Understood. (I got myself into a rookie tangle with git commits and couldn't figure out how to sort out the mess.)

duianto:
After pasting with the `spacemacs/evil-mc-paste-{after,before}` commands (`p`, `P`), then `C-n` and `C-p` cycles through the kill-ring as expected. :+1:

### evil-unimpaired paste
The `evil-unimpaired/paste-{above,below}` commands also work, when the functions are evaluated after starting Emacs.

But when Emacs is restarted then the evil-unimpaired elpa package:
`.emacs.d/elpa/evil-unimpaired-20170519.64238/evil-unimpaired.el`
takes precedence over the local Spacemacs version: 
`.emacs.d/layers/+spacemacs/spacemacs-evil/local/evil-unimpaired/evil-unimpaired.el` (which this PR modifies).

I tested this by adding a message:
`  (message "%s" "elpa: evil-unimpaired/paste-above")` and `-below`
in the elpa `evil-unimpaired/paste-{above,below}` functions, and this message:
`  (message "%s" "local: evil-unimpaired/paste-above")` and `-below`
in the local evil-unimpaired paste functions.

When Emacs is restarted and `[p` or `]p` is pressed, then the elpa message is shown.

Maybe this change needs be made in the evil-unimpaired package, but I'm unable to find where the package is hosted.

Does anyone know why there's a local and an elpa version?
I ran an ediff on both evil-unimpaired.el files and they are almost identical, except for this line in the elpa version:
> ;; Package-Version: 20170519.64238

### kill-ring cycling issue
While testing these changes. I ran into the issue where the kill ring `C-n` or `C-p` cycling, required a second press to change the pasted text (I mentioned it here: https://github.com/syl20bnr/spacemacs/pull/9056#issuecomment-307376604).

#### possible cause
I must have accidentally killed an already existing kill-ring entry. When I'm careful not to delete any text, then the `evil-paste-pop(-next)` commands (`C-{n,p}`), work as expected.

The `kill-ring` variable can contain two identical entries, but the `helm-show-kill-ring` (`SPC r y`) buffer only shows the same entry once. That's probably why I thought that `C-{n,p}` needed to be pressed twice, at the beginning/end of the kill-ring.

nochiel:
> @duianto But when Emacs is restarted then the evil-unimpaired elpa package takes precedence [...]

You need to remove the `evil-unimpaired-<stamp>` in your elpa cache. What happens is that quelpa (see the recipe in `spacemacs-evil/packages.el`) builds the version in `spacemaces-evil/local` and installs it in `.cache/elpa`. If you remove it, when you restart, spacemacs will build the updated version and install it. After that, the .cache/elpa version is loaded.

nochiel:
> @duianto: [...] The kill-ring variable can contain two identical entries, but the helm-show-kill-ring (SPC r y) buffer only shows the same entry once.

It's probably better to use `evil-show-registers` and look at the numbered registers or use `(describe-variable 'kill-ring)` with a clean kill ring as you cycle using `evil-paste-pop`. That should give you an accurate picture of any problem. Thus far, I'm not able to reproduce the problem described.

TheBB:

duianto:
@nochiel Okay, now I understand that the evil-unimpaired package is built from the file:
`.emacs.d/layers/+spacemacs/spacemacs-evil/local/evil-unimpaired/evil-unimpaired.el`
into the `.emacs.d/elpa/evil-unimpaired-0.1pre0.(date.time)` folder.

#### These are the steps I took, before understanding how it works
I first thought that by `.cache/elpa`, you meant the `.cache/quelpa/` folder, because it has an `evil-unimpaired` folder in it, with two `evil-unimpaired` files: `.el` and `.stamp`, you mentioned \<stamp\>, so it sounded like I was on the right track. (I now realise that you meant time stamp)

I first deleted that `.cache/quelpa/evil-unimpaired` folder (and restarted Emacs), but it didn't work. I then checked the `.emacs.d/elpa/evil-unimpaired-(date.time)` folder and saw the `.elc` file, and thought that, maybe it's being prioritised so I deleted it, and restarted, but no change. 

I then deleted the whole `.cache` folder, and still no change. So I reinstalled a fresh copy/config of Spacemacs, then modified the local `evil-unimpaired.el` with the `this-command` from this PR, and restarted, no change.

Now I noticed that the `.emacs.d/elpa/evil-unimpaired.(date.time)` folder name had changed after the fresh install:
from: evil-unimpaired-20170519.64238
to: evil-unimpaired-0.1pre0.20170612.141833

That made me re-read what you wrote and try to delete the `.emacs.d/elpa/evil-unimpaired(date.time)` folder. (The local `evil-unimpaired.el` was still modified). 

After restarting Emacs, now `C-n` and `C-p` cycles through the kill-ring after pasting with `[p` and `]p`. And when I check the elpa `evil-unimpaired` folder, it has the date and time from the last restart and the `.el` file contains `this-command`. 👍 :smile:

Issue #9064, opened 2017-06-09
Ranger readme missing key bindings


pbrn46:
#### Description :octocat:
README.org at https://github.com/syl20bnr/spacemacs/blob/master/layers/%2Btools/ranger/README.org is missing the "TAB" and "t" key.

Had to look at emacs ranger docs to find out: https://github.com/psamim/ranger.el/blob/21e24d098f4d081baf776ba4b8f4e941637fa142/README.md

#### Reproduction guide :beetle:
- In online docs @ github: https://github.com/syl20bnr/spacemacs/blob/master/layers/%2Btools/ranger/README.org

*Observed behaviour:* :eyes: :broken_heart:
Missing "TAB" and "t" key bindings in 

*Expected behaviour:* :heart: :smile:
Should have the bindings listed.

pbrn46:
Since the PR #9075 was cherry-picked into develop, I'm closing this issue.

PR #9075, opened 2017-06-11
Ranger Layer: Docs: Updated key bindings

pbrn46:
Key bindings updated describing file marking, based on the issue https://github.com/syl20bnr/spacemacs/issues/9064

Hope this helps all new people like me!

### DIFF
--- a/layers/+tools/ranger/README.org
+++ b/layers/+tools/ranger/README.org
@@ -111,35 +111,37 @@ To set the max files size (in MB), set the following parameter:
 
 * Key Bindings
 
-| Key Binding | Description                                          |
-|-------------+------------------------------------------------------|
-| ~SPC a r~   | launch ranger                                        |
-| ~SPC a d~   | deer (minimal ranger window in current directory)    |
-| ~C-p~       | (ranger) toggle ranger in dired buffer               |
-| ~j~         | (ranger) navigate down                               |
-| ~k~         | (ranger) navigate up                                 |
-| ~yy~        | (ranger) copy                                        |
-| ~pp~        | (ranger) paste                                       |
-| ~R~         | (ranger) rename                                      |
-| ~D~         | (ranger) delete                                      |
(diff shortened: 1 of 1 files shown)


duianto:
### Commit message
Something to think about for the next PR/commit, try to make the commit message more specific, there are many places "Updated key bindings" could be referring to.

Something like this is more descriptive: 
Ranger layer README.org: update key bindings

### Key bindings
The ranger package: https://github.com/ralesi/ranger.el documentation was updated recently, the key bindings were grouped into sections: basic, navigation, copy, rename/move or delete, marking, sorting, etc.

I'm not sure how many of the bindings should be listed in the Spacemacs layers documentation. It's probably enough to just list the most used key bindings.

Maybe a link could be added at the bottom, stating that more key bindings can be found in the ranger package documentation: https://github.com/ralesi/ranger.el

### Corrections to this PR
Here are a couple of key bindings that should be changed:
`| ~r~             | (ranger) revert buffer                               |`
r is undefined, change `~r~` to `~;g~`.

The single key entry:
`  | ~C-SPC~         | (ranger) mark current file                           |`
can be removed since it's part of the new entry:
`  | ~C-SPC~ / ~TAB~ | (ranger) mark current file                           |`

and
`| ~V~             | (ranger) visually select lines                       |`
can be removed, it only selects the current line, it doesn't seem to be able to expand the selection to the next/previous line either, so I don't know what it's used for. It has also been removed from the ranger documentation.

pbrn46:
Thanks for the info! If I reword the commit in my branch will it effectively update this PR, or should create a new commit? Same for the PR itself, is it usually better to create a PR or edit this title?

TheBB:
If you force-push your branch it will update this PR. That includes any changes whatsoever, be it changed commit messages or new commits. So feel free to change. (The title you'll have to change here on github.)

pbrn46:
@duianto I've added the corrections suggested to this PR.

TheBB:

Issue #8027, opened 2016-12-21
Symbol’s value as variable is void: ansible-doc-mode


nixmaniack:
#### Description :octocat:
Trying to get function binding for a key results in error `Symbol’s value as variable is void: ansible-doc-mode`. I have updated packages and updated spacemacs to latest develop.

#### Reproduction guide :beetle:
- Start Emacs
- Goto spacemacs init file with `SPC f e i`
- Describe binding with `SPC h d k` then `SPC /`

*Observed behaviour:* :eyes: :broken_heart:
Got error `Symbol’s value as variable is void: ansible-doc-mode`

*Expected behaviour:* :heart: :smile:
Show function under binding `SPC /`

#### System Info :computer:
- OS: darwin
- Emacs: 26.0.50.1
- Spacemacs: 0.200.5
- Spacemacs branch: develop (rev. 5d73a50)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: ivy
- Layers:
```elisp
(ivy restclient
     (auto-completion :variables auto-completion-enable-snippets-in-popup t auto-completion-enable-sort-by-usage t auto-completion-enable-help-tooltip nil auto-completion-private-snippets-directory "~/.spacemacs.d/snippets/" :disabled-for org erc)
     (semantic :disabled-for emacs-lisp python)
     better-defaults emacs-lisp
     (git :variables git-gutter-use-fringe t git-magit-status-fullscreen nil)
     github
     (markdown :variables markdown-live-preview-engine 'vmd)
     yaml shell rebox
     (spell-checking :variables spell-checking-enable-by-default nil spell-checking-enable-auto-dictionary t enable-flyspell-auto-completion nil)
     syntax-checking twitter
     (elfeed :variables rmh-elfeed-org-files
             (list "~/.emacs.d/private/elfeed.org"))
     lua
     (python :variables python-enable-yapf-format-on-save t python-test-runner
             '(nose pytest))
     ruby ruby-on-rails sml latex
     (clojure :variables clojure-enable-fancify-symbols t)
     (go :variables gofmt-command "goimports" go-tab-width 4)
     (gtags :disabled-for emacs-lisp python)
     shell-scripts html javascript chrome docker spacemacs-layouts
     (mu4e :variables mu4e-installation-path "/usr/local/share/emacs/site-lisp/mu/mu4e/" mu4e-enable-mode-line t mu4e-enable-notifications t)
     nix-mu4e colors deft finance
[...]

PR #9087, opened 2017-06-13
Fixes ansible-doc-mode error on `SPC h d k`

sdwolf:
I saw this being done for `org-mode` and implemented it here as well.

Refs #8027.


robbyoconnor:
The reason it is done with `org-mode` is that there is a bundled version of `org-mode` which is ancient - so we want to be sure which version of org-mode we're using.

robbyoconnor:
cc: @syl20bnr @TheBB

TheBB:

sdwolfz:
Apparently this fix was not enough, I'm now having this problem again... Looks like if I move the keybinding definition in `:config` everything works. 

Will submit a patch soon.

sdwolfz:
Pushed a3debe157032fc0999044eeb060d1e83107c87ee as a fix.

syl20bnr:
:+1: moving them in `:config` was a better solution.

PR #8356, opened 2017-02-13
Fixes keybindings for gfm-mode

sdwolf:
Markdown mode is actually made of two modes:
* `markdown-mode` for the canonical markdown
* `gfm-mode` for github flavoured markdown

This PR enables some of the keybindings available inside `markdown-mode` for the `gfm-mode` as well.

The way I enable `gfm-mode` is by adding this line inside my `dotspacemacs/user-config`
```emacs-lisp
(add-to-list 'auto-mode-alist '("\\.md\\'" . gfm-mode))
```

(Maybe I should create a PR to document this in the layer's Readme?)

There is a lot more configuration going on there that maybe should be enabled for `gfm-mode` as well but since I am a lisp beginner I am not sure of the consequences. Maybe someone with a bit more experience can take a look?

In the meantime, it does not hurt to port functionality to `gfm-mode` bit by bit.

Note that this will conflict with #2160 that extracts the keybinding in a function.

PR #9093, opened 2017-06-14
Makes all markdown keybindings available in gfm-mode

sdwolf:
Continuing on #8356, this PR makes all `markdown-mode` keybindings available for `gfm-mode`.

The only problem I encountered was a message with the following text:
```
Not enabling jit-lock: it does not work in indirect buffer [2 times]
```

This is related to:
```diff
-    (mmm-add-mode-ext-class 'markdown-mode nil class)))
+    (dolist (mode markdown--key-bindings-modes)
+      (mmm-add-mode-ext-class mode nil class))))
```

Not sure how to solve this yet...

I would also like to make the preffered markdown mode (either `markdown-mode` or `gfm-mode`) configurable, either from a layer `:variable` or from a `setq` inside the `user-config` but failed so far, any hints on how to do that?

### DIFF
--- a/layers/+lang/markdown/packages.el
+++ b/layers/+lang/markdown/packages.el
@@ -23,12 +23,16 @@
     ))
 
 (defun markdown/post-init-company ()
-  (spacemacs|add-company-backends :backends company-capf :modes markdown-mode))
+  (dolist (mode markdown--key-bindings-modes)
+    (spacemacs|add-company-backends
+      :backends company-capf
+      :modes mode)))
 
 (defun markdown/post-init-company-emoji ()
-  (spacemacs|add-company-backends
-    :backends company-emoji
-    :modes markdown-mode))
+  (dolist (mode markdown--key-bindings-modes)
+    (spacemacs|add-company-backends
+      :backends company-emoji
+      :modes mode)))
 
 (defun markdown/post-init-emoji-cheat-sheet-plus ()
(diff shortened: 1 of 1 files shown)


TheBB:
> any hints on how to do that?

```elisp
(push '("\\.md\\'" . (lambda () (if (should-use-md-mode) (markdown-mode) (gfm-mode)))) auto-mode-alist)
```

sdwolf:
@TheBB I was not able to implement it using the hint you gave me. I tried adding it in `markdown/init-markdown-mode` and even in a new `markdown/post-init-markdown-mode` but it would still use the default. I will give it another try later when my elisp skills are better.

Meanwhile, you can merge this PR in if you find it OK. I still don't know how to remove the `Not enabling jit-lock` warning. It seems harmless but I personally hate seeing it in the logs.

TheBB:

I don't know exactly what your code was like so can't comment on that. You need to make sure that your gfm binding comes before all others in *auto-mode-alist* though.

Not sure about using this list for company (it's probably called *markdown--key-bindings-modes* for a reason), but I'll roll with it for now.

Note that *spacemacs|add-company-backends* is a macro that doesn't evaluate its *:mode* argument so this fix is necessary: https://github.com/syl20bnr/spacemacs/commit/01d8ce76a14862e6b587ad19a91bde88913eb572

PR #9094, opened 2017-06-14
Replace obsolete which-key functions

duianto:
Problem:
Two which-key functions doc-strings, state that they are obsolete:

(which-key-declare-prefixes KEY-SEQUENCE REPLACEMENT &rest MORE)
This function is obsolete since 2016-10-05;
use ‘which-key-add-key-based-replacements’ instead.

(which-key-declare-prefixes-for-mode MODE KEY-SEQUENCE REPLACEMENT &rest MORE)
This function is obsolete since 2016-10-05;
use ‘which-key-add-major-mode-key-based-replacements’ instead.

Solution:
Replace the obsolete functions.

Searching with `SPC /` through the .emacs.d folder, didn't find any other
occurrences of these functions.


TheBB:

PR #9104, opened 2017-06-16
theming: robustness

TheBB:
When a team is applied for any reason, always disable all existing themes.

This is mostly a robustness patch, but it's necessary for `cycle-spacemacs-theme` to work properly if `default` is one of the themes.


TheBB:
Merged

PR #9108, opened 2017-06-18
Workspaces transient state: create workspace keys

duianto:
Problem:
Creating a new workspace from the workspaces transient state, requires that one
first looks for the next available workspace number, and then presses that
number key.

Solution:
Add the keys: `c` and `C` to create a new workspace. Lowercase exits the state
and uppercase keeps it open.

Notes:
This matches the behaviour of 0..9, which exits the state, and C-0..C-9, which
keeps it open.

This is the last key/command, from the C-c C-w eyebrowse key bindings, that was
missing from the workspaces transient state.


TheBB:

PR #9109, opened 2017-06-19
add android logcat layer.

kzjeef:
Add android logcat layer.

This layer is using logcat mode (https://github.com/dcolascione/logcat-mode),
this mode can do few thing that really helpful on debug and view android log


TheBB:

I moved it to the new `major-modes` layer since it was so small. Also removed the auto-mode-alist entry. Files called `.log` are much too common to be associated automatically with such a common file ending. Also fixed the deferred loading.

kzjeef:
you're right, that's better.

PR #9110, opened 2017-06-19
install orgit only when org package is used

### LINKED ISSUE #8890: org-plus-contrib install fails with spacemacs-base and no spacemacs-org layer
The `git` layer depends on `org-plus-contrib`, which lives on org elpa instead of the normal elpa. The information that `org-plus-contrib` lives on org elpa is in `spacemacs-org` layer.

When using `spacemacs-base` the `spacemacs-org` is not added. When one tries to install `org-plus-contrib` or anything that depends on `org-plus-contrib`, one is presented with the following (unhelpful) error:

```
An error occurred while installing orgit (error: (file-error http://elpa.gnu.org/packages/org-plus-contrib-20170502.tar Not found))
```

I think it would be good to add some documentation, or an extra check when installing orgit.

(I fixed it by just adding the `spacemacs-org` layer).

d12frosted:
Fixes #8890

### DIFF
--- a/layers/+source-control/git/packages.el
+++ b/layers/+source-control/git/packages.el
@@ -25,7 +25,7 @@
         magit-gitflow
         ;; not compatible with magit 2.1 at the time of release
         ;; magit-svn
-        orgit
+        (orgit :toggle (configuration-layer/package-usedp 'org))
         smeargle
         ))
 


TheBB:

PR #9111, opened 2017-06-19
Replace usage of `ensime-print-type-at-point` by `ensime-type-at-point`

jcazevedo:
This PR replaces the usage of `ensime-print-type-at-point` by `ensime-type-at-point` as the method name was recently changed in `ensime-emacs` (see ensime/ensime-emacs@8c7f659a170a2af248f912e56ec8ac25453fc867).

### DIFF
--- a/layers/+lang/java/funcs.el
+++ b/layers/+lang/java/funcs.el
@@ -93,7 +93,7 @@
   (setq-local eldoc-documentation-function
               (lambda ()
                 (when (ensime-connected-p)
-                  (ensime-print-type-at-point))))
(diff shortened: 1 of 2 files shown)


TheBB:
Hm, I know that we have advised people in the documentation to use the stable version of ensime. Maybe that means we need to defalias this function in the meantime so that it will continue to work for people who are on stable?

TheBB:

See https://github.com/syl20bnr/spacemacs/commit/a394e6b89443db8f13fe6ba9371415451a9be6b3

jcazevedo:
Thanks!

PR #9112, opened 2017-06-19
Implement helm-dir-search

jgertm:
I found myself wishing for this functionality.

 It behaves like the other helm-search wrappers, with the current directory as the scope of the search.


TheBB:
Is there a reason why you let-bind `default-directory` to `dir`, using that as an argument, instead of just calling with `default-directory` as a function argument directly?

TheBB:
IOW, why

```elisp
(let ((dir default-directory))
  (some-function dir))
```

instead of

```elisp
(some-function default-directory)
```

jgertm:
Mostly misguided aesthetic sensibility.

jgertm:
@TheBB ping

TheBB:
You neglected the ivy layer so I had to do that myself. That's why it took a while. https://github.com/syl20bnr/spacemacs/commit/ebc6ba015eef0184d9bb46dabd2052b004baf652

Issue #9016, opened 2017-06-02
Which function controls popup window in Spacemacs?


tddsg:
Hi,

I want to display the popup-window of Spacemacs by splitting the current window, but not the entire frame, like what was described in this [StackOverflow](https://emacs.stackexchange.com/questions/27711/how-to-make-pop-up-window-appear-in-the-divided-portion-of-the-frame) post (created by me before).

The solution to customize `popwin:popup-buffer` used to work with me a few months ago. However, with recent changes in Spacemacs, it doesn't work anymore. 

I guess Spacemacs change the way the popup-window is displayed. However, I couldn't find out how it can be done. Can you advise me which function currently control the popup-window of Spacemacs?

Thank you!

PR #9116, opened 2017-06-20
Customization which allows pupo to split the active window or the entire frame

tddsg:
Hi,

I add a customization to the `pupo` mode, which allows Spacemacs to split the popup-window either inside the active window or inside the entire frame.

This patch addresses an issue in [Stackoverflow ](https://emacs.stackexchange.com/questions/27711/how-to-make-pop-up-window-appear-in-the-divided-portion-of-the-frame) and also the issue https://github.com/syl20bnr/spacemacs/issues/9016 of Spacemacs.

In the patch, the variable `pupo-split-active-window` is set by default to `nil`, which doesn't change the default behavior of `pupo`.

If this variable is set to `t`, then `pupo` will split only the active window.

Please advise if the fix is useful?

Thanks!

### DIFF
--- a/layers/+spacemacs/spacemacs-purpose/local/spacemacs-purpose-popwin/spacemacs-purpose-popwin.el
+++ b/layers/+spacemacs/spacemacs-purpose/local/spacemacs-purpose-popwin/spacemacs-purpose-popwin.el
@@ -6,6 +6,12 @@
 
 (require 'window-purpose)
 
+(defcustom pupo-split-active-window nil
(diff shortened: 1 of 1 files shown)


TheBB:
cc @bmag does this make sense?

tddsg:
FYI, it can be useful when the `popwin` is used for some buffer like `compilation`.
You can see from the photo that only few space is occupied

spacmacs

bmag:
Haven't tested it, but looks very reasonable.

TheBB:

PR #9120, opened 2017-06-20
Fix disabled scrollbar remains

ZarsBranchkin:
When starting a new emacs instance, there is this ugly bar on the right side(spotted by @greenbigfrog):

Ugly scrollbar remains

Although it disappears on resizing the emacs frame, it's still an annoying visual bug.

Luckily it can be easily fixed by disabling scrollbar before any other UI elements.


TheBB:

PR #9123, opened 2017-06-20
Clojure: eval -> pretty-print

jgertm:
I find this behavior useful. Displays the result of evaluation in a separate buffer, pretty-printed according to Clojure rules.


TheBB:

PR #9126, opened 2017-06-21
Remove spurious quote and fix parinfer for .spacemacs

aaronjensen:
Fixes parinfer's parsing of .spacemacs

Without this, parinfer detects unbalanced parens (which is surely a bug in parinfer) but this works for us to get it working.


TheBB:

aaronjensen:
wow new record. Thanks @TheBB

PR #9133, opened 2017-06-23
typescript: allow to disable eldoc-mode

vegansk:
Sometimes eldoc-mode can be very slow, for example when using type inference on a complex data structures. This PR adds the ability to disable it.

### DIFF
--- a/layers/+lang/typescript/README.org
+++ b/layers/+lang/typescript/README.org
@@ -65,6 +65,15 @@ You can choose formatting tool:
 
 Default is 'tide .
 
+Sometimes eldoc-mode can be very slow, for example when using type inference on a complex data structures.
+You can disable it:
+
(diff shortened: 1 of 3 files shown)


TheBB:
Thanks, but the correct way to do this is to use `(typescript :packages (not eldoc))` in your layer list. Then `typescript/post-init-eldoc` will not run. You'll still get the eldoc-mode in `.tsx` files as per the `post-init-web-mode` function, but that code is just not good and should be changed entirely (as the comment says).

vegansk:
I tried to do this using `(typescript :packages (not eldoc))`, unfortunatly this doesn't work and I don't know why:
image

image

TheBB:
Well, do you have `eldoc-mode` in `typescript-mode-hook`?

PR #9136, opened 2017-06-23
Fix: Check dependencies for owned packages too

### LINKED ISSUE #9134: The :depends keyword does not take package ownership into account
#### Description :octocat:

The layer packages `:depends` keyword does not take package ownership into
account. That is, if a layer declares a package in `<layer>-packages` but
doesn't initialize it (doesn't own it), the `:depends` keyword will still treat
the package as if it were available.

For example, given:

`layers/foo/packages.el`:

```elisp
(setq foo-packages '(foo))

(defun foo/pre-init-foo ())
```

`layers/bar/packages.el`:

```elisp
(setq bar-packages '(bar :depends foo))

(defun bar/init-bar ())
```

`bar` *will* be installed (and `foo` along with it if the `bar` package actually
depends on `foo`).

#### System Info :computer:
- OS: gnu/linux
- Emacs: 26.0.50
- Spacemacs: 0.200.9
- Spacemacs branch: local (rev. c5ef8edd)
- Graphic display: t
- Editing style: vim
- Completion: ivy

TheBB:
Fixes #9134

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -1695,9 +1695,10 @@ wether the declared layer is an used one or not."
   "Returns true if PKG should be configured for LAYER.
 LAYER must not be the owner of PKG."
   (let* ((owner (configuration-layer/get-layer (car (oref pkg :owners))))
-         (disabled (oref owner :disabled-for))
-         (enabled (oref owner :enabled-for)))
-    (and (not (memq nil (mapcar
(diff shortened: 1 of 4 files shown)


robbyoconnor:
I pulled this locally and tested it and it seems to work and fix the issue I was seeing

syl20bnr:
Thank you ! 💜 

I added some tests to verify the new predicates code paths.

PR #9140, opened 2017-06-24
ycmd: python3 compatibility fix for global_conf.py

aroig:


TheBB:
I think more robust would be

```python
try:
    from itertools import izip as zip
except ImportError:
    pass
```

Then later on in the file use `zip` instead of `itertools.izip`.

aroig:
I force-pushed the new version

TheBB:

PR #9146, opened 2017-06-26
Perl5 keybindings and bug fix

CeleritasCelery:
- added keybindings for POD and HERE doc selection
- added keybindings for perldoc's
- added dump jump support
- fixed tab command. In my initial PR #9046 I had included this fix but then removed it believing it was not necessary. However I have found that the default `cperl-indent-command` is not sufficient and won't work properly when indenting blocks of unaligned code.


TheBB:

PR #9149, opened 2017-06-26
Ensure that .venv is a regular file before trying to source it.

eklitzke:
Until today, I wasn't using virtualenvwrapper.sh, and I was in the habit of creating a directory called `.venv` in my projects that contained my virtualenv. However, virtualenvwrapper.sh uses the convention that instead a file named `.venv` at the root of the directory should contain the name of the virtualenv to activate. This makes Spacemacs be more cautious, by checking that any `.venv` file it finds is actually a regular file before trying to read its contents. This will thus avoid a spurious error that I get currently from Spacemacs, which is like:
```
File mode specification error: (file-error Read error Is a directory /home/evan/code/foo/.venv)
```

### DIFF
--- a/layers/+lang/python/funcs.el
+++ b/layers/+lang/python/funcs.el
@@ -105,16 +105,15 @@
 (defun spacemacs//pyvenv-mode-set-local-virtualenv ()
   "Set pyvenv virtualenv from \".venv\" by looking in parent directories."
   (interactive)
-  (let ((root-path (locate-dominating-file default-directory
-                                           ".venv")))
-    (when root-path
-      (let* ((file-path (expand-file-name ".venv" root-path))
-             (virtualenv
-              (with-temp-buffer
-                (insert-file-contents-literally file-path)
-                (buffer-substring-no-properties (line-beginning-position)
-                                                (line-end-position)))))
-            (pyvenv-workon virtualenv)))))
+  (-when-let* ((root-path (locate-dominating-file default-directory ".venv"))
+               (file-path (expand-file-name ".venv" root-path))
+               (venv-is-file (file-regular-p file-path))  ; for side-effects
+               (virtualenv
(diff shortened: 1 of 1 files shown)


pareshmg:
I'm having the same issue, and I solved the issue as:

```
(defun spacemacs//pyvenv-mode-set-local-virtualenv ()
  "Set pyvenv virtualenv from \".venv\" by looking in parent directories."
  (interactive)
  (let ((root-path (locate-dominating-file default-directory
                                           ".venv")))
    (when root-path
      (let* ((file-path (expand-file-name ".venv" root-path)))
        (if (file-directory-p file-path)
            (pyvenv-activate file-path)
          (pyvenv-workon (with-temp-buffer
                           (insert-file-contents-literally file-path)
                           (buffer-substring-no-properties (line-beginning-position)
                                                           (line-end-position)))))))))


```

This checks if the .venv file is a directory or a file. If it is a file, it uses the spacemacs default. Otherwise, it directly calls pyvenv-activate.

Would this be the desired effect instead of the when-let option?

TheBB:
@eklitzke This is fine, but can you please rewrite so this doesn't depend on dash? E.g. with Emacs standard functions instead of `-when-let`. Spacemacs source code doesn't have dash as a dependency.

art-solopov:
I've worked around the issue by managing the virtualenvs semi-manually and disabling the hooks:

```lisp
  (remove-hook 'python-mode-hook 'spacemacs//pyvenv-mode-set-local-virtualenv)
  (remove-hook 'hy-mode-hook 'spacemacs//pyvenv-mode-set-local-virtualenv)

  (add-hook 'projectile-after-switch-project-hook
            (lambda ()
              (when (projectile-project-p)
                (activate-current-virtualenv (projectile-project-root)))
              )
            )
```

PR #9153, opened 2017-06-27
[+lang/shell-scripts] Add checkbashisms

Yamakaky:
I hope I didn't forget any modification to do.

### DIFF
--- a/layers/+lang/shell-scripts/README.org
+++ b/layers/+lang/shell-scripts/README.org
@@ -6,6 +6,7 @@
 - [[#description][Description]]
   - [[#features][Features]]
 - [[#install][Install]]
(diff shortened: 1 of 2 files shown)


sdwolfz:
This looks like a useful tool but I'm not sure if it's place is here. Maybe in a debian specific layer?

mineo:
checkbashisms is not only available on Debian, Arch Linux ([click](https://www.archlinux.org/packages/community/any/checkbashisms/)) and Homebrew ([click](https://formulae.brew.sh/formula/checkbashisms)) both carry packages for it.

syl20bnr:
Please make sure it does not break for user without the tool installed.

PR #9155, opened 2017-06-28
Prompt to save modified, user created buffers on quit

ZarsBranchkin:
By default, there is no way to save all user created buffers, so they can be easily lost when quitting emacs.

This can be solved by making use of the second optional argument for the `save-some-buffers` function and buffer local variable `buffer-offer-save`.

Thanks, @pcuinty, for bringing this problem up!


TheBB:

PR #9162, opened 2017-06-29
Actually configure SPC m b to switch between ERC buffers.

eklitzke:
The documentation for the ERC layer already says that `SPC m b` is supposed to switch between buffers, but it's not actually set up to do that. This change fixes the issue.


TheBB:
It's a rare case that the docs are ahead of the code for once. Thanks! Cherry-picked in develop. You can safely delete your branch.

PR #9163, opened 2017-06-30
Add option for controlling ERC inline image support

eklitzke:
This adds a new option `erc-enable-images` which can be used to control whether or not ERC displays inline images. The default value is `t`, to preserve the existing behavior. A user can set this value to `nil` to disable inline images.


bmag:
Thanks for the contribution, however this change is not needed. Any package can be disabled completely by adding it to `dotspacemacs-excluded-packages`. In addition, `erc-image`'s configuration in `erc` layer can also be disabled by using `(erc :packages (not erc-image))` in `dotspacemacs-configuration-layers`.

PR #9164, opened 2017-06-30
No-sync startup flag

ZarsBranchkin:
This adds a new startup flag `--no-sync`. It will force spacemacs to skip
package synchonization. This can be useful in cases when you're working under
poor or restrictive network.

Thanks, @zaript, for this idea!

### DIFF
--- a/core/core-command-line.el
+++ b/core/core-command-line.el
@@ -16,8 +16,8 @@
 (defvar spacemacs-insecure nil
   "If non-nil force Spacemacs to operate without secured protocols.")
 
-(defvar spacemacs-skip-sync nil
-  "If non-nil skip package synchronization on startup")
+(defvar spacemacs-sync-packages t
+  "If non-nil synchronize packages on startup")
 
 (defun spacemacs//parse-command-line (args)
   "Handle Spacemacs specific command line arguments.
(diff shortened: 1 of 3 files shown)


zaript:
This is very useful for careless users like me, who fiddle with configuration and restart emacs in an environment, where no packages can be downloaded. So I sit there waiting for 50+ package download to time out and can do nothing about that.

TheBB:
Looks ok to me. Can I ask you to invert the meaning of the variable? So that it's named in the positive rather than the negative. You can do that with the parameter to `configuration-layer/sync` too.

ZarsBranchkin:
Something like this, I guess? Should I also change the flag to `--skip-sync` or that can stay `--no-sync`?

TheBB:
Ah, what I meant was, name the variable something like `spacemacs-sync-packages`, not `spacemacs-DONT-sync-packages`, give it a default value of `t`, etc. You get the idea I'm sure.

TheBB:
The command line flag should have the same semantic meaning as it currently has, because syncing packages should be the default behaviour.

ZarsBranchkin:
This should be it then, made the variable and parameter positive.

syl20bnr:
Thank you ! 👍 

I made some important modifications as in its current state the PR triggered a void variable error if the parameter was not set:

- I renamed the function `configuration-layer/sync` to `configuration-layer/load` and removed the parameter in exchange of the global variable `spacemacs-sync-packages`.
- I renamed `--no-sync` to `--no-package-sync` which is more explicit

The effect is that `--no-package-sync` always prevent package sync which is the expected behavior I presume.

PR #9172, opened 2017-07-01
Use JSX header without breaking React

jamrizzi:
Using the header below is now deprecated in React.
```js
/** @jsx React.DOM */
```

However, there are situations where I have to use the *.js extension, and need to tell my text editor that the file has JSX.

I suggest switching to JSX mode if it detects a header starting with `@jsx`. Then I'll be able to do the following without breaking react.

```js
/** @jsx */


TheBB:

jamrizzi:
Dang, thanks so much :)

PR #9174, opened 2017-07-02
Add clip2org layer

leezu:
This pull request adds a layer for the clip2org.el package. Note that this package is not on melpa due to https://github.com/thamer/clip2org/pull/1 so it can't simply be installed by adding it to `dotspacemacs-additional-packages` (i.e. it seems not possible to add the `recipe:` to `dotspacemacs-additional-packages`?).

The proposed configuration is
```
  '((clip2org :variables
          clip2org-clippings-file "~/Dropbox/Books/My Clippings.txt")))
```

where `clip2org-clippings-file` is a variable defined by clip2org.el and not by a potential `config.el` file. Is this considered ok? It works at least..

### DIFF
--- a/layers/+tools/clip2org/README.org
+++ b/layers/+tools/clip2org/README.org
@@ -1,0 +1,23 @@
+#+TITLE: clip2org
+
+* Table of Contents                                        :TOC_4_gh:noexport:
+- [[#description][Description]]
+- [[#install][Install]]
+- [[#configuration][Configuration]]
+
+* Description
+This package converts Kindle’s “My Clippings.txt” to a format usable in Org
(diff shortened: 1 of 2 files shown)


TheBB:
The variable is fine. I'm more concerned that it's a little excessive to have a whole layer dedicated to this IMO rather niche package. It's a perfect example of something that should be in additional-packages and in fact there's no problem using `:location (recipe :fetcher github :repo "thamer/clip2org")`. I'm not sure why you are experiencing problems with that.

leezu:
@TheBB thanks. In that case I probably got the syntax for specifying the recipe wrong. I am closing the pull request.

PR #9177, opened 2017-07-03
Idris layer: ignore *.ibc files in helm-projectile

juxtin:
Idris projects end up with one .ibc file for every .idr (source) file, but they're not human readable so you always have to skip over them in projectile. This commit adds the suffix to `projectile-globally-ignored-file-suffixes` so they'll just be invisible.


TheBB:
I'm not so sure about this. Projectile reads project-specific ignores from *.projectile*, and in case of git projects it also obeys *.gitignore*. It seems a little blunt to **globally** ignore file formats by default.

sdwolfz:
@holguinj can you please confirm that @TheBB's suggestion (adding `*.ibc` to your .gitignore for example) acheves the desired effect?

PR #9185, opened 2017-07-04
Add Kubernetes layer

mattjbray:
This layer provides the [kubernetes-el](https://github.com/chrisbarrett/kubernetes-el#kubernetes-el) package, a magit-style interface to the Kubernetes command-line client.


nixmaniack:
IMO, Kubernetes shouldn't take top level keybinding. It should be moved to `applications` under `SPC a`.

Since we have Docker under `SPC D` it's debatable but since we have almost all apps/tools under `SPC a` it would be a better place.

mattjbray:
@nixmaniack Sounds reasonable. I took inspiration from the `docker` layer for `SPC K`.

`SPC a k` is taken by `paradox-list-packages`.

So does `SPC a K o` sound OK?

(There is only a single binding so it could be `SPC a K`. But I thought `SPC a K o` might be better in case bindings are added in the future.)

nixmaniack:
> (There is only a single binding so it could be SPC a K. But I thought SPC a K o might be better in case bindings are added in the future.)

`SPC a K` should be good if there's only command to start it. If there are more entrypoints then it would make sense for `SPC a K o` as you stated. Does it start magit like popup based operations menu? In that case I would go for `SPC a K`.

mattjbray:
> Does it start magit like popup based operations menu? In that case I would go for SPC a K.

@nixmaniack Yes it's exactly that. I've changed the keybinding to `SPC a K`.

layers/+tools/kubernetes/packages.el
```diff
@@ -0,0 +1,34 @@
...
+(defconst kubernetes-packages
+  '(
+    kubernetes
+    kubernetes-evil
+    )
+  )
+
+(defun kubernetes/init-kubernetes ()
+  (use-package kubernetes
+    :init
```
TheBB: Please defer loading of this package.

mattjbray: @TheBB Done. It works for me but I'm not 100% sure I've done it the right way.

robbyoconnor:
Fix formatting issues.

mattjbray:
@robbyoconnor Updated and rebased onto `develop`. Looks like the failing test is unrelated to this change.

robbyoconnor:
@mattjbray CI failures are not your fault.

robbyoconnor:
Please add a README.

tsutsarin:
bump
any work needed to merge this pr?

/cc @mattjbray, @robbyoconnor, @TheBB

smile13241324:

PR #9187, opened 2017-07-05
Navigate to project file without toggle

sdwolfz:
This is something that was bothering me. If a file was part of a project located in a different tree than the one displayed and I was navigating to it with `SPC f T` then neotree would show it's parent directory and not it's project. But `SPC p t` would close neotree and I'd have to repeat the action to open it's project.

If you don't like the change to the keybind functionality then please consider adding just the function definition so it can be easily configured in a `.spacemacs` file (let me know and I will update the PR).


sdwolfz:
Closing this as I am no longer interested in the functionality (and no longer using neotree).

syl20bnr:
😂

PR #9188, opened 2017-07-05
Add `SPC b u` keybinding for reopening killed buffers

deuill:
This is similar to the `Ctrl+Shift+T` keybinding found in major browsers, and helps when accidentally killing a buffer (i.e. fat-fingering `SPC b d` when meaning to press `SPC b s`).

Only buffers that resolve to files (and not ephemeral buffers) will be considered, and stored in a stack which is pushed to and popped from on buffer kill.


deuill:
Noted that the underlying stack is boundless, and will grow as needed. Not certain if memory use will be a problem, but I have not encountered any issues, not having restarted Emacs in weeks.

TheBB:
Thanks, this looks good. Could you

* move the `(add-hook ...)` call to *config.el*, and
* document the new binding in *DOCUMENTATION.org*

deuill:
Bah, completely forgot about documentation. Will do both.

TheBB:

PR #9190, opened 2017-07-05
ruby: change documentation lookup key binding

p-sdk:
`SPC m h h` is conventional key binding to show documentation for thing under
point. It is called by `spacemacs/evil-smart-doc-lookup`, so this change makes
`K` in normal state work as expected.


TheBB:

PR #9191, opened 2017-07-05
Fix docstring for spacemacs/evil-smart-doc-lookup

p-sdk:
Rephrase and replace (inaccurate) goto-definition with documentation lookup.


TheBB:

PR #9200, opened 2017-07-07
Fix Org MIME keybindings in gnus layer help.

dschrempf:


TheBB:

PR #9206, opened 2017-07-09
Adds GOROOT to list of variables copied from shell environment

alindeman:
From <https://golang.org/doc/install> "Installing to a custom location"

> The Go binary distributions assume they will be installed in /usr/local/go
> (or c:\Go under Windows), but it is possible to install the Go tools to a
> different location. In this case you must set the GOROOT environment variable
> to point to the directory in which it was installed.


d12frosted:
That makes sense.

PR #9208, opened 2017-07-10
add gtags support for c-c++ layer

chengyi:
add gtags support for c-c++ layer using `counsel-gtags`

### DIFF
--- a/layers/+lang/c-c++/packages.el
+++ b/layers/+lang/c-c++/packages.el
@@ -21,6 +21,7 @@
     flycheck
     gdb-mi
     ggtags
+    counsel-gtags
     helm-cscope
     helm-gtags
     realgud
@@ -174,3 +175,7 @@
     :post-init
     (dolist (mode '(c-mode c++-mode))
       (spacemacs/setup-helm-cscope mode))))
+
+(defun c-c++/post-init-counsel-gtags ()
+  (spacemacs/counsel-gtags-define-keys-for-mode 'c-mode)
+  (spacemacs/counsel-gtags-define-keys-for-mode 'c++-mode))
--- a/layers/+tags/gtags/funcs.el
+++ b/layers/+tags/gtags/funcs.el
@@ -56,6 +56,19 @@ Otherwise does nothing."
     "gy" 'helm-gtags-find-symbol
     "gu" 'helm-gtags-update-tags))
 
+(defun spacemacs/counsel-gtags-define-keys-for-mode (mode)
+  "Define counsel-gtags key binding for Mode"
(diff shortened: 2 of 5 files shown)


layers/+tags/gtags/packages.el
```diff
@@ -14,6 +14,7 @@
   '(
     ggtags
     (helm-gtags :requires helm)
+    (counsel-gtags :toggle (configuration-layer/layer-usedp 'ivy))
```
TheBB: Please use the new `:requires` keyword for this sort of thing. (See line above.)

layers/+lang/c-c++/packages.el
```diff
@@ -174,3 +175,7 @@
     :post-init
     (dolist (mode '(c-mode c++-mode))
       (spacemacs/setup-helm-cscope mode))))
+
+(defun c-c++/post-init-counsel-gtags ()
+  (spacemacs/counsel-gtags-define-keys-for-mode 'c-mode)
+  (spacemacs/counsel-gtags-define-keys-for-mode 'c++-mode))
```
TheBB: Is there any reason why you have done this only for C/C++ and not for all the other modes where `spacemacs/helm-gtags-define-keys-for-mode` is used currently?

layers/+tags/gtags/funcs.el
```diff
@@ -56,6 +56,19 @@ Otherwise does nothing."
...
+  (spacemacs/set-leader-keys-for-major-mode mode
+    "gC" 'counsel-gtags-create-tags
+    "gd" 'counsel-gtags-find-definition
+    "gr" 'counsel-gtags-find-reference
+    "gS" 'counsel-gtags-find-symbol
+    "gs" 'counsel-gtags-dwim
+    "gu" 'counsel-gtags-update-tags
+    "gf" 'counsel-gtags-find-file
+    "gn" 'counsel-gtags-go-forward
+    "gp" 'counsel-gtags-go-backward))
```
TheBB: The differences between this and the helm version should be documented in the README.

chengyi: OK, I'll modify the README. thanks.

layers/+lang/c-c++/packages.el
chengyi: @TheBB I'll add variable to `gtags` layer, so people can decide which mode that the want gtags enable.

beta1440:
Maybe we could have a `spacemacs/gtags-define-keys-for-mode (gtags-backend mode)` function where the `gtags-backend` arguement specifies whether to define keys for `helm-gtags` or `counsel-gtags`.

PR #9210, opened 2017-07-10
[ruby-layer] Adding rufo.el package for automatic formatting

dcluna:
This adds [rufo.el](https://github.com/danielma/rufo.el) to ruby modes for automatic code formatting upon save (like `gofmt`).

### DIFF
--- a/layers/+lang/ruby/packages.el
+++ b/layers/+lang/ruby/packages.el
@@ -14,6 +14,7 @@
         bundler
         chruby
         company
+        rufo
         (enh-ruby-mode :toggle ruby-enable-enh-ruby-mode)
         evil-matchit
         flycheck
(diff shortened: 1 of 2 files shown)


dcluna:
@robbyoconnor Can I get LGTM on this one? This is a simple integration with [rufo](https://github.com/ruby-formatter/rufo).

layers/+lang/ruby/packages.el
```diff
@@ -65,6 +66,13 @@
     :init (spacemacs/add-to-hooks 'chruby-use-corresponding
                                   '(ruby-mode-hook enh-ruby-mode-hook))))
 
+(defun ruby/init-rufo ()
+  (use-package rufo
+    :init (dolist (mode '(ruby-mode enh-ruby-mode))
+            (add-hook (intern (concat (symbol-name mode) "-hook")) 'rufo-minor-mode)
+            (spacemacs/set-leader-keys-for-major-mode mode
+              "=b" 'rufo-format))))
+
```
robbyoconnor: Why not add formatting for regions as well?

dcluna: AFAIK `rufo` works on the complete AST of the program, so it _may_ break something if we feed it to a narrowed region. Not sure if it's worth it. Also, the `rufo.el` package automatically formats on save, so there is no need for adding it to regions.

robbyoconnor: Another thing: Use `SPC m=` -- which is what is used elsewhere.

dcluna:
Closing this since I'm not using `rufo` anymore. Whoever wants to move this forward, feel free to use this.

PR #9216, opened 2017-07-12
Add missing which-key prefix names

duianto:
Unnamed Which-key prefixes:
`SPC b`, `N -> +prefix`
`SPC x`, `j -> +prefix`
have to be opened to see what kind of commands they contain.


duianto:
When the `version-control` layer is installed, then another unnamed prefix appears: `SPC g`, `h -> +prefix`, but it only contains one key/command: `o -> browse-at-remote`, so I'm not sure what prefix name it should have.

d12frosted:

PR #9219, opened 2017-07-12
Add customization for hiding original buffer

### LINKED ISSUE #9214: LaTeX compilation does not work in `centered-buffer-mode`
When editing LaTeX files in `centered-buffer-mode`, compilation gives an error

    ! I can't find file `<none>.tex'.
    <*> \input <none>.tex

This may be related to #8653, although I am using the `develop` branch.

Also, when `centered-buffer-mode` is activated, two buffers are pointing to the
same file (one is the non-centered buffer, the other is the centered buffer). In
my opinion (of course this is open for debate), `centered-buffer-mode` should
work like a toggle or switch (on or off, one buffer, no confusion).

Thanks a lot,
Dominik

JAremko:
fix #9214


d12frosted:
You should get something like https://stackoverflow.com/help/badges/37/disciplined

PR #9223, opened 2017-07-13
fix ipython args (issue #6580)

jinyangustc:
Since IPython 5.4+ and 6.0+, `rlipython` brings back the classic readline functionality. We should start ipython (4.2 < version < 5.4) with "--simple-prompt -i", and "-i" for other versions. I also add a section in README to give installation instruction.

### DIFF
--- a/layers/+lang/python/README.org
+++ b/layers/+lang/python/README.org
@@ -10,6 +10,7 @@
   - [[#dependencies][Dependencies]]
     - [[#auto-completion-anaconda-dependencies][Auto-completion: Anaconda dependencies]]
     - [[#syntax-checking][Syntax checking]]
+    - [[#rlipython][rlipython]]
   - [[#test-runner][Test runner]]
   - [[#automatic-buffer-formatting-on-save][Automatic buffer formatting on save]]
   - [[#autoflake][autoflake]]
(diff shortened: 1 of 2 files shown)


robbert-vdh:
Wouldn't this break the REPL for people who have IPython 5.4+ installed, but not rlipython?

jinyangustc:
@robbert-vdh In fact when I was at 5.4, the `--simple-prompt -i` doesn't work well, neither. According to issue #6580, it seems that `--simple-prompt -i` isn't fixed for all. Considering that the python layer recommends to install several dependencies, I think `rlipython` should be the right way to delivery the full experience.

I've also considered to add a `defvar`, but it may be a more intrusive solution. Any idea?

robbert-vdh:
I guess you could check whether the `rlipython` module is installed. A rather crude way of doing this would be to add `(string= "" (shell-command-to-string "python -c 'import rlipython'"))` to [line 32](https://github.com/syl20bnr/spacemacs/pull/9223/files#diff-5820ae173a30d31c1eab8d69d8d72ac5R32), though I'm sure there's a much better way to do this.

et2010:
I haven't tried this yet but you could squash the two commits into one.

jinyangustc:
@robbert-vdh Yeah, this approach is better. Thanks!

jinyangustc:
@robbert-vdh @et2010 I tried `(string= "" (shell-command-to-string "python -c 'import rlipython'"))` and it works, but I didn't come up with better solution. Calling shell commands seems to increase the start up time of the REPL (a little bit).

I agree that the PR shouldn't break other's config or make one pay for something he/she doesn't use. I close this PR unless a better solution is found.

If necessary, maybe we should add something related to the interpreter shell and args in the README, since there are dozens of issues related to the topic.

Thank you for your guidance and suggestions!

PR #9225, opened 2017-07-13
Add keybinding to toggle between a keyword & a string

jr0cket:
The function `clojure-toggle-keyword-string` will convert a string to a keyword, or keyword to a string.

The keybinding is added under the Clojure refactor menu, in the "cycle/clean/convert" section.

`SPC m r c k`

As this is a `clojure-mode` function, it is defined in the clojure-mode refactoring keybinding section of packages.el.

There are no conflicting keybindings from the clj-refactor package that is also part of the Clojure layer.
https://github.com/clojure-emacs/clj-refactor.el/wiki

### DIFF
--- a/layers/+lang/clojure/README.org
+++ b/layers/+lang/clojure/README.org
@@ -239,53 +239,54 @@ As this state works the same for all files, the documentation is in global
 
 *** Refactoring
 
-| Key Binding   | Description                 |
-|---------------+-----------------------------|
-| ~SPC m r ?~   | describe refactoring        |
-| ~SPC m r a d~ | add declaration             |
-| ~SPC m r a i~ | add import to ns            |
-| ~SPC m r a m~ | add missing libspec         |
-| ~SPC m r a p~ | add project dependency      |
-| ~SPC m r a r~ | add require to ns           |
-| ~SPC m r a u~ | add use to ns               |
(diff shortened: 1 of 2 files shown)


jr0cket:
The description of the new keybinding in the README.org file extends the org-table width, so I included the whole "Refactoring" table as the change.  Please let me know if this was not the correct thing to do and I will change the pull request.

Thanks.

bmag:
Since keywords start with `:`, have you considered <kbd>rc:</kbd> instead of <kbd>rck</kbd>? Seems a bit more mnemonic to me, but also a bit less convenient. Also fits well with the <kbd>rc#</kbd>, <kbd>rc[</kbd>, etc.

Changing the whole table as you did was indeed the correct thing.

jr0cket:
@bmag a great point about using `:` instead of `k`.  As other keybindings use the character then so should the keyword.

I will update this pull request and and change the keybinding to `rc:`

Thank you

jr0cket:
Updated this pull request to use a new keybinding, `SPC m r c :`, based on feedback on this pull request.

This new keybinding replaces the `SPC m r c k` keybinding and is in line with the other keybindings within this section of the menu.

Thanks to @bmag for the feedback.

bmag:

PR #9236, opened 2017-07-15
@Fix issue: SPC m : call current major mode commands

TinySong:


bmag:
Is this related to a specific issue? Please add some info about the issue, and link to relevant issue if one exists. There might really be a bug, but I can't tell at a glance.

bmag:
This PR has been replaced by #9265.

PR #7355, opened 2016-10-10
New debug layer

mfulz:
Thank you for contributing to Spacemacs!

Before you submit this pull request, please ensure it is against the develop branch and not master.

I've created a debug layer, which is using gud and can be extended to different debuggers for specific modes.
For now this layer took over ownership of gdb-mi and the c-c++ layer is changed to use this layer instead.
In my opinion this makes sense and it would be nice if this could be included.

Thank you <3

PR #9246, opened 2017-07-18
add debug layer

CeleritasCelery:
@sukeyisme originally created this layer with #8799. However his implementation only supported the `c-c++` layer, so @syl20bnr merged it into that layer instead of creating a new one. I have since refactored the code so that this functionality can be added to any language. I updated the `c-c++` layer to use the new method for adding debuggers and also added support for perl5. 

This should be its own layer since realgud supports many [languages](https://github.com/realgud/realgud/wiki/Debuggers-Supported).

I have added a method called `spacemacs|add-realgud-debugger` that can be called from any language layer to add support for their respective debuggers. If language layer maintainers want to add debugger support they can use this call in a `pre-init-realgud` function.


beta1440:
@CeleritasCelery I refactored the `c-c++` layer in commit 8e3a59dbfdaf5716db987d7a51ee6144b92bea66. I would appreciate it if you could rebase your changes on the current development branch.

CeleritasCelery:
@JAremko I remerged this commit and now the travis build is failing because of documentation issues (missing features with a colon). Am I honestly expected to go and fix everyone's documentation issues in order for my commit to be valid?

JAremko:
@CeleritasCelery  Not expected but it would be nice of you :wink: 
In this case it looks like only `layers/+lang/perl5/README.org` needs fixing. 

- - - - - - - -

But you can ignore it. I'll make this part of the test disabled until we sort out #9346

CeleritasCelery:
Thanks @JAremko. However it looks like a lot more then just one file needs to be fixed. I fixed the issue in the perl5 README as you suggested and now it is complaining about  `CONTRIBUTING.org`'s indentation. Before it was complaining about the `bibtex` layer. I don't feel that making all the required documentation fixes is in the scope of this PR. I feel that disabling these checks until develop is clean is the best course of action. 

On another note, is there anyway this could get merged (or at very least reviewed) @JAremko? As you can see from the number of 👍  this layers has generated a lot of interest and would be huge addition to Spacemacs. All the Travis build were passing before I remerged.

JAremko:
@CeleritasCelery  I made the FORMATTING test optional. Just rebase your PR :+1:

CeleritasCelery:
Thanks @JAremko! looks like it is good again.

JAremko:
@CeleritasCelery   `ci/circleci` is optional :wink:

CeleritasCelery:
@JAremko How do I disable it?

JAremko:
@CeleritasCelery  you can simply ignore it or add `[ci skip]` to the commit message

mfulz:
Would be really great, if that could be merged. My PR is nearly a year old and it looks like the realgud stuff is even better

CeleritasCelery:
@syl20bnr self assigned this, so imagine he will make it a priority when he gets back to Spacemacs.

CeleritasCelery:
@syl20bnr @JAremko could we get a merge or a review on this? This would be a great feature for the next release, especially as other language layers add support. It has generated a lot of interest.

robbert-vdh:
Any update on merging this? Being able to use a single debugger for every language would be a huge plus.

CeleritasCelery:
I think I may have to open a new PR.  This one seems to have fallen into the massive heap of forgotten Pull requests. The maintainers only look at the first few.

JAremko:
@CeleritasCelery  I think @syl20bnr  is busy with the new release. He self assigned to this PR so it should be merged. Eventually :wink:

myrgy:
What about adding one extra folder.
'+debug/realgud'. I'd like to add gdm-mi layer as well.

myrgy:
Nice hint from this PR:  #7355 - make debugger transient state

dvzubarev:
I created recenty [transient state](https://gist.github.com/dvzubarev/4cf3490e2a42850f067c6a94b619149a) for realgud. It may be useful for this PR.

CeleritasCelery:
I am going to leave this PR open but I am also going to create a new repo which has the debug layer. There has been a lot of interest in this and many user requests but no action has been taken in Spacemacs. If we make the layer it own repo, users can install and use that until/if this gets merged. 

@dvzubarev What does a transient state add that the shortkey state doesn't?

dvzubarev:
It is not convenient for me to use shortkey state. It overwrites some evil bindings that I used to use in the source code buffer, e.g. `e`, `s`. Also it has unusual keys:  `s` for -next, `i` for -step, `o` for finish, instead of `n`, `s`, `f`.

CeleritasCelery:
> Also it has unusual keys


Those are the keybinding mnemonics that Spacemacs [uses](https://github.com/syl20bnr/spacemacs/blob/master/doc/CONVENTIONS.org#debugging).  

What keybinding do you use to activate your transient state? I do like the idea of having the transient state help text.

dvzubarev:
I use these keybindings:
```
(spacemacs/set-leader-keys-for-major-mode mode
      "d." 'spacemacs/realgud-transient-state/body
      "de" 'my/realgud-eval-dwim
      "dJ" 'realgud:cmd-jump)
```

dvzubarev:
Actually, I think it is better to run this transient state automatically when entering new source buffer. It can be achived via this code:
```elisp
(add-hook 'realgud-short-key-mode-hook 'spacemacs/realgud-transient-state/body)
```

MaskRay:
Regarding `(evilified-state-evilify-map realgud:shortkey-mode-map`

Do you have some idea to set up keys for evaluating variables around the point? Something like avy-jump will be great.

I currently use my clumsy 1 to 9
https://github.com/MaskRay/Config/blob/master/home/.emacs.d/private/%2Bmy/my-code/packages.el#L248

CeleritasCelery:
It is up to @syl20bnr to merge new layers, so not much can be till then. In the meantime I moved the code to its own repo that people can use until it is merged. 
https://github.com/CeleritasCelery/debug-layer

ajholyoake:
@CeleritasCelery,

I've added the following to the packages.el in the python layer
```
(defun python/pre-init-realgud()
(dolist (mode '(anaconda-mode)) 
(spacemacs|add-realgud-debugger mode "trepan2")))
```
And tried adding both (separately)
``` (debug :variables debug-additional-debuggers '("trepan2"))```
and
``` debug``` to my .spacemacs.

trepan2 seems to be registered when I do M-x realgud:trepan2 (and will execute as it should), however SPC-m-d-d doesn't seem to be firing as it should do. As far as I can see the add-realgud-debugger mode should do this.

I'm fairly new to emacs, so am at a bit of a loss as to how to debug this - have I missed anything?

**Comment by CeleritasCelery** [comment]
Adding it to `debug-additional-debuggers` does not establish a key binding because it doesn’t know what major mode. Also IIRC, `anaconda-mode` is a minor mode not a major mode. 

If you want to add `trepan2` to python use this 
```elisp
(defun python/post-init-realgud ()
    (spacemacs/add-realgud-debugger ‘python-mode “trepan2”))
```

PR #9253, opened 2017-07-21
Fix iedit paste-replace, kill-ring cycling

duianto:
Problem:
Pressing `C-n` or `C-p` to cycle the kill ring, after pasting in iedit state,
shows the message: `user-error: Previous command was not an evil-paste:
evil-iedit-state/paste-replace`.

Solution:
Set the `this-command` variable to `evil-paste-before`, before calling
`evil-iedit-state/paste-replace`.

Issue:
Sometimes a cycled kill-ring element only appears by the main cursor. I haven't been able to figure out the cause, but more people investigating it will probably help.

It seems to work most of the time, so it's at least better than the current `previous command was not an evil-paste` message.

Advice function:
I haven't looked into how "advicing" a function works, but maybe that would be a simpler solution, than what I came up with.

### DIFF
--- a/layers/+spacemacs/spacemacs-evil/funcs.el
+++ b/layers/+spacemacs/spacemacs-evil/funcs.el
@@ -83,6 +83,13 @@ Otherwise, revert to the default behavior (i.e. enable `evil-insert-state')."
 
 ;; multiple-cursors
 
+(defun spacemacs/evil-iedit-state/paste-replace (count)
+  "Enable `C-n' and `C-p' for cycling the kill ring after pasting in
+iedit-state."
+  (interactive "P")
+  (setq this-command 'evil-paste-before)
+  (evil-iedit-state/paste-replace count))
+
 (defun spacemacs//paste-transient-state-p ()
   "Return non-nil if the paste transient state is enabled."
   (and dotspacemacs-enable-paste-transient-state
--- a/layers/+spacemacs/spacemacs-evil/packages.el
+++ b/layers/+spacemacs/spacemacs-evil/packages.el
@@ -172,7 +172,10 @@
         (kbd dotspacemacs-leader-key) spacemacs-default-map)
       (spacemacs//iedit-insert-state-hybrid dotspacemacs-editing-style)
       (add-hook 'spacemacs-editing-style-hook
-                #'spacemacs//iedit-insert-state-hybrid))))
+                #'spacemacs//iedit-insert-state-hybrid)
+      ;; enable kill ring cycling with `C-n' and `C-p' after pasting
+      (define-key evil-iedit-state-map "p"
+        'spacemacs/evil-iedit-state/paste-replace))))
 
 (defun spacemacs-evil/init-evil-indent-plus ()
   (use-package evil-indent-plus


d12frosted:
LGFM, but I am not using evil, so it's hard for me to justify this PR. I'll wait a little bit for more input 😸 

P. S. Thanks for the PR! ❤️

duianto:
The issue where the pasted text only appears at the cursors iedit occurrence, is unrelated to this PR. It happens on both the current master and develop branches.

But it works as expected if the cursors iedit occurrence is at the beginning of a line.

#### Reproduction steps.
1. it happens in any buffer, but open the scratch buffer `SPC b s`
2. copy and paste the following into Emacs:
```
some sample text
text some sample
sample text some

copy-me
```
3. copy the 5th line "copy-me", by pressing `yy`
4. put the cursor on the 2nd lines, 2nd word: "some"
5. start iedit `SPC s e`
6. paste `p`
**observed:**
```
 sample text
copy-me
text  sample
sample text 
```
All occurrences were deleted (this happens before `evil-iedit-state/paste-replace` calls `evil-paste-before`), but then "copy-me" only got inserted above the 2nd line, instead of at the occurrences.

"copy-me" getting inserted above the 2nd line, seems to be an `evil-paste-before` issue/feature, because when iedit mode is disabled, and `evil-paste-before` is called by pressing 'P' (uppercase). Then "copy-me" also gets inserted above the 2nd line). This is why "copy-me" wasn't inserted between the 2nd lines two words "text" and "sample".

**expected:**
"copy-me" (and a newline, which was copied with `yy`) should have been inserted at all occurrences, resulting in:
```
copy-me
 sample text
text copy-me
 sample
sample text copy-me
```
7. undo twice `uu`, to restore text from step 2
8. move to the first occurrence by pressing `N`
9. paste `p`
**observed and expected behaviour are the same:**
all occurrences have been replaced with "copy-me" (and the newline)


#### Possible cause:
It might have something to do with `yank-handler` and/or `(evil-yank-line-handler nil t)` being added to the kill ring element.

When copying the "copy-me" line with `yy` then the kill-ring element `SPC h d v` `kill-ring` looks like this:

On the develop branch:
```
Its value is (#("copy-me\n" 0 8
   (fontified t yank-handler
              (evil-yank-line-handler nil t))))
```
or:
```
Its value is (#("copy-me\n" 0 7
   (fontified t yank-handler
              (evil-yank-line-handler nil t))
   7 8
   (fontified t yank-handler
              (evil-yank-line-handler nil t))))
```
(I don't know what I did while testing these instructions, but suddenly another fontified section showed up in the kill-ring element, when "copy-me" was copied with: `yy`. I tried restarting Emacs but it still shows a second fontified section.)


This is what it looked like on the master branch:
```
Value: (#("copy-me\n" 0 4
   (fontified t yank-handler
              (evil-yank-line-handler nil t))
   4 5
   (fontified t yank-handler
              (evil-yank-line-handler nil t))
   5 7
   (fontified t yank-handler
              (evil-yank-line-handler nil t))
   7 8
   (yank-handler
    (evil-yank-line-handler nil t))))
```

#### When it works as expected
Make sure that there's an empty line below the "copy-me", and select the "copy-me" line before copying it: `0v$y`, now the kill-ring element looks like this:

On the develop branch:
```
Its value is (#("copy-me
" 0 8 (fontified t)))
```
or this:
```
Its value is (#("copy-me
" 0 7 (fontified t) 7 8 (fontified t)))
```

On the master branch:
```
Its value is (#("copy-me\n" 0 4
   (fontified t)
   4 5
   (fontified t)
   5 7
   (fontified t)
   7 8
   (fontified t)))
```
Pasting in iedit mode now replaces all occurrences as expected, even if the current occurrence isn't in the left most column.

Miciah:
@duianto, I made https://github.com/syl20bnr/evil-iedit-state/pull/24 to fix the problem you described in your last comment.

Miciah:
@duianto, this PR needs to be rebased, but does this PR plus https://github.com/syl20bnr/evil-iedit-state/pull/24 resolve all the issues mentioned above?

duianto:
@Miciah The `evil-iedit-state` `#24` PR, makes it possible to paste a line that's been copied with `yy` from any iedit occurrence.

Currently (without the iedit `#24` PR) pasting to all occurrences only works from an occurrence at the beginning of a line.

There's another issue where the iedit occurrences only stays selected when pasting from certain words in the sample text:
```
some sample text
text some sample
sample text some

copy-me
```

Currently the occurrences stay selected when pasting from the left most words on each line.
With Miciah's PR, the occurrences stay selected when pasting from the first occurrence (in the sample text from the words on the first line).
When pasting from the other occurrences, then they get deselected which prevents this `kill-ring` cycling PR from working.

I don't know if that made much sense, so I took some screenshots. They'll either make it more clear or more confusing. 😄 


without evil-iedit-state pr 24


with evil-iedit-state pr 24

sdwolfz:
Should I wait for https://github.com/syl20bnr/evil-iedit-state/pull/24 to get merged first? Or these two PRs are independent and can be merged at any time and in any order?

Miciah:
Each PR is independent, and I think each one strictly improves behavior.  I need to look more into most recent issue reported by duianto, but I think it is yet another issue to fix.

duianto:
Maybe this PR (that solves the kill-ring cycling issue), just is a patch to an underlying problem.

The current Spacemacs paste commands are:
https://github.com/syl20bnr/spacemacs/blob/ffc1617dd92b35ef4173a3f65fb0a4d5fffde714/layers/%2Bspacemacs/spacemacs-evil/funcs.el#L92
https://github.com/syl20bnr/spacemacs/blob/ffc1617dd92b35ef4173a3f65fb0a4d5fffde714/layers/%2Bspacemacs/spacemacs-evil/funcs.el#L100

and the keys are bound here:
https://github.com/syl20bnr/spacemacs/blob/ffc1617dd92b35ef4173a3f65fb0a4d5fffde714/layers/%2Bspacemacs/spacemacs-evil/packages.el#L214-L224

`evil-mc-mode` isn't enabled by default in Spacemacs. Maybe the evil-mc paste commands only should be bound when `evil-mc-mode` is enabled and there is more than one cursor.

This might be the commit that rebound the Spacemacs paste keys:
Disable paste transient state when multiple cursors are active https://github.com/syl20bnr/spacemacs/commit/58458f2d2abcc1211444c2060ab598f55e518da4

I seem to remember paste cycling working before the Spacemacs paste keys were rebound.

---
[evil-goggles-mode](https://github.com/edkolev/evil-goggles) has a similar issue where pasting in Spacemacs doesn't show the green flash: 


source: https://github.com/edkolev/evil-goggles#preview

It's because evil goggles is looking for `evil-paste-before` and `evil-paste-after`.
```
(defun evil-goggles--paste-advice (_ &optional register yank-handler)
  "Advice for `evil-paste-before' and `evil-paste-after'.
  ...
```
source: https://github.com/edkolev/evil-goggles/blob/d7876e6566ac82b7c3251a59651e7db6ab756589/evil-goggles.el#L535-L536

Calling `M-x evil-paste-after` shows the green flash.

syl20bnr:
I merged https://github.com/syl20bnr/evil-iedit-state/pull/24 and I'm going to release a new version of evil-iedit-state. My remarks below is with this fix.

Not sure about the original behavior but with `evil-mc` enabled pressing `C-n` and `C-p` now add cursors to the end of the next occurrences.

Given:

```
some sample text
text some sample
sample text some

copy-me
```

After copying `copy-me` and replacing `some`, stay in `iedit` state and press `C-n`, a new `evil-mc` cursor is created. What should be the expected behavior here ? I was not familiar with the kill ring cycling (does it make the paste transient state useless?).

duianto:
Confirmed, the `insert-as-yank` fix works. Pasting text with a line break now occurs at every instance.

I haven't really used the paste transient state (yet).

The default emacs-evil behavior after pasting (`p`) is:
```
 C-p			;; evil-paste-pop
 C-n			;; evil-paste-pop-next
```

But evil-mc which has been enabled by default after this PR was posted, creates new cursors:
```
 C-p                     ;; evil-mc-make-and-goto-prev-match
 C-n                     ;; evil-mc-make-and-goto-next-match
```

It seems unlikely that one would want to start a new evil-mc cursor while one is in `iedit-mode`.
Maybe `evil-mc-mode` could be disabled while `iedit-mode` is active.

Or should it be left as is for now? It's still possible to cycle the kill ring after a yank or evil-paste-{before, after} with `M-y` (`yank-pop` or `evil-paste-pop`).

---

This PR adds: `(setq this-command 'evil-paste-before)` to a Spacemacs version of the function:
`evil-iedit-state/paste-replace`

But it should probably be handled upstream in `evil-iedit-state`.

With both the `insert-as-yank` fix and this PRs kill ring cycle fix (`this-command`):
```elisp
  (defun evil-iedit-state/paste-replace (count)
    "Replace the selection with the yanked text."
    (interactive "P")
    (setq this-command 'evil-paste-before)
    (when kill-ring (iedit-delete-occurrences))
    (evil-paste-before count nil '(#'insert-as-yank)))
```
Pressing `M-y` after pasting in `evil-iedit-state/iedit-mode` (`SPC s e`), shows the error message:
>Error in post-command-hook (iedit-update-occurrences-2): (wrong-type-argument number-or-marker-p nil)

I'll open an issue upstream in evil-iedit-state.

This PR isn't relevant anymore, therefore I'll close it.

PR #9258, opened 2017-07-22
shell layer: set "terminal" purpose on shell buffers

bennofs:
Sets the `terminal` purpose on `eshell` and `term-mode` buffers.

### DIFF
--- a/layers/+tools/shell/packages.el
+++ b/layers/+tools/shell/packages.el
@@ -27,6 +27,7 @@
         (term :location built-in)
         xterm-color
         vi-tilde-fringe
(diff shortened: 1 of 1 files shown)


bmag:
@bennofs I'd like to make this change upstream at https://github.com/bmag/emacs-purpose. Can you submit a PR upstream please? (below [this line](https://github.com/bmag/emacs-purpose/blob/master/window-purpose-configuration.el#L299))

bennofs:
@bmag done in https://github.com/bmag/emacs-purpose/pull/109

PR #9259, opened 2017-07-22
Fix magit navigation in dvp layout.

JuanCaicedo:
This fixes the use of `h` to move up when in magit mode using `dvp` layout. I assume that other layouts might benefit from a similar fix, but I only fixed this one since it's the one I use. Glad to help with others though 😃 

### How to test
- Set layout to dvp
    ```lisp
   (dotspacemacs-configuration-layers
   '((keyboard-layout :variables
                      kl-layout 'dvp)))
    ```
- Open magit mode
- On develop, `h` will run `magit-dispatch-popup`
- On this branch, it will run `evil-next-line`


JuanCaicedo:
Hi there, I just wanted to ask what you think of this PR 😀 I would like to make other improvements for myself for using dvp, but don't want to send them upstream if you don't think they're helpful

JuanCaicedo:
Hi there, wanted to ask again what you think of this PR. Does it fix the problem in the way that you would imagine? @syl20bnr @StreakyCobra

StreakyCobra:
I'm not using Spacemacs often those days, and I haven't programmed in elisp for months 🙈 so I can not judge on the "technical" side of your solution anymore. But if there is something not working with `dvp` keyboard layout in magit, fixing it in the `:dvp` of `magit` seems to be the correct way indeed.

It's all what I can say :-) I'll let current Spacemacs collaborators review the PR. If the situation has not changed since last year, it can take a while for a PR to be reviewed/included.

PR #9262, opened 2017-07-22
Fix stuff

JAremko:


JAremko:
I think it will be better to replace all local org-mode links with GitHub links because people put all kind of stuff in them like fuzzy headline/text search that doesn't work at GitHub. It will not affect documentation viewing within Spacemacs because `space-doc-mode` translates such links back into local ones.

PR #9265, opened 2017-07-23
Fix smex major mode command

TinySong:
"SPC m :" shortcut should be major mode command, See smex README.org


d12frosted:
Indeed, it was broken several commits ago. 

P. S. In future, please squash your commits and include only relevant changes. E. g. in this case you also committed changes related to indentation.

PR #9270, opened 2017-07-23
fix misused function in OCaml layer

jinyangustc:
`merlin' is a package, not a layer.


d12frosted:

PR #9283, opened 2017-07-25
[WIP] Improve CI

JAremko:
Improve CI


JAremko:
So after all my test and trials with Travis CI stages and deploy...  
None of above help to streamline and/or speedup Spacemacs testing :disappointed:

PR #9286, opened 2017-07-26
Change kill-other-buffers from `b C-d` to `b D`. Mnemonics maintained!

aniketd:
I have been using 'SPC b m' for a long time to kill-other-buffers like scratch, home, those opened by magit and others...
But after that being changed recently to 'C-d' in #7733 I strongly felt that for such a common use case using combination-keys would be uncomfortable. There is no issue related to this change.

If it is mandatory for an issue to be present for all pull requests I will be happy to raise one.

I hope this change is welcome by members of the community.

### DIFF
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -2103,10 +2103,10 @@ Buffer manipulation commands (start with ~b~):
 | ~SPC TAB~       | switch to alternate buffer in the current window (switch back and forth) |
 | ~SPC b b~       | switch to a buffer                                                       |
 | ~SPC b d~       | kill the current buffer (does not delete the visited file)               |
+| ~SPC b D~       | kill other buffers                                                       |
 | ~SPC u SPC b d~ | kill the current buffer and window (does not delete the visited file)    |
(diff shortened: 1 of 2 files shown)


bmag:
`SPC b D` is already bound to `spacemacs/ace-kill-this-buffer`. I'm not sure `kill-other-buffers` is a common use case either.

aniketd:
@bmag I found `SPC b D` in the documentation (CHANGELOG, DOCUMENTATION, news/news01.org), but did not find it in code anywhere in the whole code base. As for `kill-other-buffers` I agree that it is not a common use case generally, but only I happen to use it on a daily basis, perhaps by being OCD'ed about open buffers... :smile:  ... Well, if `b D` is not good for this case, what alternative would you suggest for it? `b o`? ...any non-chording sequence?
Thanks for your feedback! :+1:

bmag:
@aniketd I think `SPC b C-d` is already a good choice for the default key-binding. Don't forget that `SPC o` is reserved for the user, so in this case my opinion is that binding `SPC o <something>` in your personal config is better than changing the default binding. Does it sound reasonable?

aniketd:
@bmag Yep! In that case, the possible mnemonics are already taken. I'll stick to using `C-d` and surely get used to it in a while... :smile:
Thanks @bmag ... Closing this one...

PR #9289, opened 2017-07-26
Adds some profiler keybinding under `SPC P`

sdwolf:
This is to make the profiler more accessible to people that experience bad performance. I chose `P` since it was unused and only added the ones I've used before. Feel free to suggest changes, or other commonly used ones and I will add them.

Also, I feel like this needs to be documented somewhere, any hints on where to put it?


d12frosted:
I really think that they should be documented somewhere (in DOCUMENTATION.org?) in order to make them even more ** accessible** 😸 

Other than this - I am OK with your PR. Though I am very bad at creating key bindings (if I was an author of Spacemacs it wouldn't have mnemonics as it's pillar).

nixmaniack:
Although profiler is useful, IMO it's not something that deserves top level binding. I consider profiler as "helpful" which can go under Spacemacs' "help" section under `SPC h`.

sdwolf:
@nixmaniack that's an interesting poin, I was also thinking of adding it under `SPC a` but both the `p` and 'P' namespace were taken.

The main audience for which I want to optimise this are the people new to spacemacs or unfamilliar with the profiler that run into problems, ask for help in the issue seciton and are asked to submit a profiler report. `SPC P` is faster to type, but `SPC h P` might fit in with the report issue `SPC h I`.

I'm gonna leave this as is for now to gather more oppinions.

CeleritasCelery:
I agree with @nixmaniack on this. Should be a help binding.

sdwolf:
I've updated the PR, I will look into adding some documentation now, in the meantine, if you can think of any other commonly used profiler functions that I can add here, please let me know.

PR #9293, opened 2017-07-27
make spacemacs/delete-file to actually work on directories

d12frosted:
It didn't really worked on directories, though documentation stated that it should! I couldn't live with such knowledge as a burden, so I tried to fix that. 😸 

P. S. Also fixed that part where it kills buffer before asking user. 

P. P. S. I have implemented helper function that might be used in future. By me at least.

### DIFF
--- a/layers/+distributions/spacemacs-base/funcs.el
+++ b/layers/+distributions/spacemacs-base/funcs.el
@@ -380,16 +380,22 @@ projectile cache when it's possible.
 When ASK-USER is non-nil, user will be asked to confirm file
 removal."
   (interactive "f")
-  (when (and filename (file-exists-p filename))
-    (let ((buffer (find-buffer-visiting filename)))
-      (when buffer
-        (kill-buffer buffer)))
-    (when (or (not ask-user)
-              (yes-or-no-p "Are you sure you want to delete this file? "))
-      (delete-file filename)
+  (let ((is-dir (file-directory-p filename)))
+    (when (and (file-exists-p filename)
+               (or (not ask-user)
+                   (yes-or-no-p
+                    (message  "Are you sure you want to delete this %s? "
+                              (if is-dir "directory" "file")))))
+      (spacemacs/kill-file-or-dir-buffers filename)
+      (if (file-directory-p filename)
+          (delete-directory filename t)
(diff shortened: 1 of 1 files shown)


bmag:
Wait, there really isn't a built-in or idiomatic way to delete a directory recursively and kill all the relevant buffers? How does Dired handle it?

layers/+distributions/spacemacs-base/funcs.el
```diff
@@ -380,16 +380,22 @@ projectile cache when it's possible.
...
-      (delete-file filename)
+  (let ((is-dir (file-directory-p filename)))
+    (when (and (file-exists-p filename)
+               (or (not ask-user)
+                   (yes-or-no-p
+                    (message  "Are you sure you want to delete this %s? "
+                              (if is-dir "directory" "file")))))
+      (spacemacs/kill-file-or-dir-buffers filename)
+      (if (file-directory-p filename)
+          (delete-directory filename t)
```
bmag: Shouldn't we pass also a non-nil `trash` argument?

d12frosted: No why? We don't move files to trash, we remove them, so it's consistent with a line after. 

Though we might add an argument for trashing.

bmag: Well, we delete the files either way. The question is whether to delete them permanently (as `rm` does) or put them in the trash can (as file managers do). If we move the files to the trash then the user can still restore them if he/she realizes it was a mistake to delete the files. I think that should be the default, and that's also the default behavior of `delete-directory` when called interactively (but only when called interactively).

Same comment goes for `delete-file` below, I just missed it before.

bmag: Lets add an optional argument in any case, and lets make it have the reverse meaning when `spacemacs/delete-file` is called interactively. (same as `delete-directory`)

Miciah:
Counsel upstream recently added a "delete" action that uses `dired-delete-file`: https://github.com/abo-abo/swiper/commit/e0293f23fa3d2d027403f826175b301d0d529975.  Do we still need `spacemacs/delete-file`?

Issue #8571, opened 2017-03-23
Not possible to create new file, if another file with similar name exists


abellmann:
If I want to create a new file in spacemacs, I usually use SPC f f or C-x C-w in a new buffer.

This fails to work, if a file with a similar name already exists, because the existing file is matched and opened..

example: I want to create the file test.org, but a file sometest.org already exists in the same directory. I have no way to create the file. (note: I am using ivy).

abellmann:
In evil mode, this can be done with ":e filename", but this is more a workaround than a solution. I think it should be possible to decide in ivy, if I want to create a new file or open an existing one.

I also found, that this is possible with C-M-j in ivy, but this is anything, but intuitive..

sdwolf:
I agree, just tried it out, not intuitive but `C-M-j` is the [official keybind](https://github.com/abo-abo/swiper/blob/ac944ead01871bea01e95b09f18459c83ef1eae2/ivy.el#L275). Maybe a README file for ivy could help by documenting it, currently there is no such file.

I could start writing one if it's OK with everybody.

d12frosted:
Oh right, this particular issue bogged me as well when I tried ivy for the first time. Thanks to official documentation I've found solution pretty fast. 

There is a [good place][1] for mentioning this key binding.

  [1]: https://github.com/syl20bnr/spacemacs/blob/develop/doc/DOCUMENTATION.org#ivy

Stebalien:
Maybe bind it to something like `S-RET`, `M-RET`, or `C-RET`. Also, I've opened an issue (https://github.com/abo-abo/swiper/issues/933) to discuss ways to make this more user friendly.

fabacino:
Please see https://github.com/abo-abo/swiper/issues/933#issuecomment-310638597 for a possible solution to this issue.

duianto:
Tested on the latest master and develop branches:
- press: `SPC SPC`, type: `set-variable` and press: `RET`
- type: `ivy-use-selectable-prompt` and press: `RET`
- type: `t` and press: `RET`
- press: `SPC f f`, type: the first letter(s) of a file/folder in the current folder. The first match gets selected.
- Move up the selection `C-p` to the "Find file: " prompt, and press: `RET`

A new buffer opens, with the same name as the letters that were typed at the "Find file: " prompt. 👍

d12frosted:
[...]

PR #9309, opened 2017-07-29
set ivy-use-selectable-prompt to t

d12frosted:
Following #8571.

---

I think it's a good default. It doesn't make anyone suffer, but most importantly, might save someone from struggle while renaming file to a new name that is a prefix of old name. Just an example 😸 

Let me know if you have any objections or suggestions.


d12frosted:

PR #9316, opened 2017-07-30
Add layer for `snoopy-mode`

jgertm:
`snoopy-mode` is quite useful while programming.

### DIFF
--- a/layers/+misc/snoopy/README.org
+++ b/layers/+misc/snoopy/README.org
@@ -1,0 +1,16 @@
+#+TITLE: snoopy layer
+
+* Table of Contents                                        :TOC_4_gh:noexport:
+- [[#description][Description]]
+- [[#install][Install]]
+
+* Description
+  This layer adds ~snoopy-mode~ as a minor mode in all programming modes.
+  ~snoopy-mode~ is a minor mode for number row unshifted character insertion in
+  Emacs. It can insert characters such as ~!~ and ~@~ without pressing the shift
+  key.
+
+* Install
+  To use this configuration layer, add it to your =~/.spacemacs=. You will need to
+  add =snoopy= to the existing =dotspacemacs-configuration-layers= list in this
+  file.
--- a/layers/+misc/snoopy/packages.el
+++ b/layers/+misc/snoopy/packages.el
@@ -1,0 +1,20 @@
+;;; packages.el --- snoopy layer packages file for Spacemacs.
+;;
+;; Copyright (c) 2012-2017 Sylvain Benner & Contributors
+;;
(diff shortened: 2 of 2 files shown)


d12frosted:
Hey @jgertm,

Thank you for your PR. I am not sure if it ever will be merged as a separate layer as it kind of violates [new layer contribution policy](https://github.com/syl20bnr/spacemacs/blob/master/CONTRIBUTING.org#contributing-a-layer).

> Layer with no associated configuration will be rejected. For instance a layer with just a package and a hook can be easily replaced by the usage of the variable `dotspacemacs-additional-packages`.

Though if it's really that useful, it might get into existing layer, like `spacemacs-editing`. 

I am going to tag this PR as `- Mailing list -` in hopes to gather more input from other Spacemacs users.

AdrieanKhisbe:
Could be usefull, but as @d12frosted point out it could be a good idea to put it in spacemacs-editing.

Also a binding to turn on and off the mode would be usefull I think

smile13241324:
Just found that old PR, is it still desirable to have it merged into develop? If it is I could incorporate it into spacemacs-editing as well as cherry pick @jgertm original PR to give due credit.

smile13241324:

Snoopy mode can be activated in the newest version of the dotfile together with a custom keyboard layout.

syl20bnr:
@smile13241324 Seems there is another package that could fit the bill and is better integrated with evil: https://github.com/wbolster/evil-swap-keys
I believe it can work in insert mode only so this is handy to still have numbers easily accessible in normal mode for the universal prefix.
Also in the dotfile we should refer to this feature as `swap-keys` instead of referencing `snoopy-mode` which does not say really what it does. Snoopy is the implementation detail of the feature to swap keys.
I believe we should only expose the number row swap feature and just call it `dostspacemacs-swap-number-row` and hide the advanced feature. The value of this variable could be the keyboard layout:
- `nil` is disabled
- `qwerty-us` 
- `qwerty-ca`
- `azerty-fr`
- etc...
And let people provide new layouts.

smile13241324:
@syl20bnr good point, I will have a look at evil-swap-keys and report here.
Renaming the existing variable to remove snoopy mode is no problem, I had no clue how to describe that in English in the first place :) and swap-keys is much clearer.

I'll check how to add new keyboard layouts and try to describe that in the dotfile description in `dotspacemacs-swap-number-row`.

smile13241324:
I have had a look at evil-swap-keys it looks more polished than snoopy, I will prepare a commit to replace snoopy and make the adjustments on the dotfile.

PR #9318, opened 2017-07-30
spacemacs-base: Add key bindings for projectile-edit-dir-locals

beta1440:
This key binding opens the projects `dir-locals.el` file in a buffer.


d12frosted:

PR #9319, opened 2017-07-31
Add snippet for bundle template

nickanderson:
This brings snippet support into the cfengine layer.


beta1440:
Why not add your snippet to the [yasnippets repo?](https://github.com/AndreaCrotti/yasnippet-snippets).

nickanderson:
I suppose I could do that. Thanks for the suggestion @beta1440.

PR #9320, opened 2017-07-31
c-c++: Refactor to reduce duplication

beta1440:
Instead of repeating actions for c-mode and c++-mode and their respective hooks,
let loop through each mode or hook to reduce duplication. This makes it easier to add new functionality for the c and c++ modes.


d12frosted:

PR #9324, opened 2017-07-31
Chinese layer: update package name due to upstream renaming

et2010:
`chinese-pyim` package of Chinese layer has been renamed to `pyim`. This will use the correct package and variable names to replace the deprecated ones.


d12frosted:

Issue #9115, opened 2017-06-20
[dev] `php` layer is not show code-completion any more.


wdy1184:
#### Description :octocat:
`php` layer in develop branch is not show code-completion any more.

#### Reproduction guide :beetle:
- Checkout `develop` branch 
- Start Emacs
- Open a php file
- Type some code

*Observed behaviour:* :eyes: :broken_heart:
There is no code completion below cursor.

*Expected behaviour:* :heart: :smile:
Show `company` completion below cursor when I type.

#### System Info :computer:
- OS: darwin
- Emacs: 25.2.1
- Spacemacs: 0.200.9
- Spacemacs branch: develop (rev. d8c00ffd)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: helm
- Layers:
```elisp
(ibuffer yaml vimscript sql html python go javascript emacs-lisp common-lisp helm git org markdown syntax-checking better-defaults imenu-list dash
         (chinese :variables chinese-enable-fcitx nil)
         (gtags :variables gtags-enable-by-default t)
         (auto-completion :variables auto-completion-enable-snippets-in-popup t auto-completion-enable-sort-by-usage t :disabled-for markdown org)
         (version-control :variables version-control-diff-tool 'diff-hl version-control-global-margin t version-control-diff-side 'left)
         php)
```
- System configuration features: JPEG IMAGEMAGICK NOTIFY ACL LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS

TheBB:
Works here if I trigger manually. Have you tried that?

wdy1184:
@TheBB How can I turn it on?

wdy1184:
I trigger by using `company-complete`, but it just show `function` completion.

beta1440:
@wdy1184 Do you have `company-mode` enabled when editing PHP files? Try enabling `company-mode` manually to see if that helps.

wdy1184:
@beta1440 Yes, I opened it, but I must open it every time I want to complete my code.

beta1440:
For some reason in the PHP layer [packages.el](https://github.com/syl20bnr/spacemacs/blob/develop/layers/+lang/php/packages.el#L74) file, `company-minimum-prefix-length` is set to 1000. You have to type at least 1000 characters to get autocompletion working. This variable shouldn't be set in the PHP layer if there's no compelling reason. Below is the relevant code snippet from the PHP layer `packages.el`.

[...]

PR #9327, opened 2017-07-31
php: Fix php-company autocompletion

beta1440:
Solve #9115.

The company-minimum-prefix-length variable shouldn't be set to 1000. Doing so
prevents the company completion popup from showing. Instead, let the
company-minimum-prefix-length stay at its default value.


d12frosted:

PR #7629, opened 2016-11-02
Add clang-format-buffer to c-c++ layer

psamim:
Bind `SPC m = in` c-c++ layer to clang-format-buffer (see syl20bnr#7628).

This aligns with the same key binding in other layers, such as `yapfify-buffer` in Python layer or `web-beautify-js` in Javascript layer.

Issue #7628, opened 2016-11-02
c-c++ Layer needs improvement


psamim:
Hi,

I am using Spacemacs for 6 months now, today I started using c-c++ layer for the first time. I feel it needs major improvements. It needs some major mode key bindings (`SPC m`) for

- Compile/debug of one file
- Compile/debug of project (using projectile)
- ~~clang-format (buffer/file) `(SPC m =)`~~ (fixed in develop branch)
- Refer to gtags layer in README

Thanks

-----------------------------------------------------

Edit:
- _syl20bnr on 19th December 2017_ Strikethrough "clang-format (buffer/file) `(SPC m =)`"

TheBB:
> Compile/debug of project (using projectile)

`SPC c c`

> Compile/debug of one file

In C/C++ this typically doesn't make sense.

> clang-format (buffer/file) (SPC m =)

PR welcome.

> Refer to gtags layer in README

PR welcome.

PR #9334, opened 2017-08-01
c-c++: Add clang-format functions and key bindings

beta1440:
Solve part of #7628. Additionally, #7629 is no longer needed.

Add support for formatting lines, functions, and buffers with clang-format. 
Each function produces a clear message on success.

The README was updated to reflect the usage of the new key bindings. References
to binding clang-format to the tab key were removed.

### DIFF
--- a/layers/+lang/c-c++/packages.el
+++ b/layers/+lang/c-c++/packages.el
@@ -65,8 +65,14 @@
   (use-package clang-format
     :if c-c++-enable-clang-support
     :init
-    (when c-c++-enable-clang-format-on-save
-      (spacemacs/add-to-hooks 'spacemacs/clang-format-on-save c-c++-mode-hooks))))
+    (progn
+      (when c-c++-enable-clang-format-on-save
+        (spacemacs/add-to-hooks 'spacemacs/clang-format-on-save c-c++-mode-hooks))
+      (dolist (mode c-c++-modes)
+        (spacemacs/declare-prefix-for-mode mode "m=" "format")
+        (spacemacs/set-leader-keys-for-major-mode mode
+          "==" 'spacemacs/clang-format-region-or-buffer
+          "=f" 'spacemacs/clang-format-function)))))
 
 (defun c-c++/init-cmake-ide ()
   (use-package cmake-ide)
--- a/layers/+lang/c-c++/README.org
+++ b/layers/+lang/c-c++/README.org
@@ -9,12 +9,13 @@
 - [[#install][Install]]
   - [[#layer][Layer]]
   - [[#default-mode-for-header-files][Default mode for header files]]
-  - [[#enable-clang-support][Enable Clang support]]
+  - [[#clang-configuration][Clang Configuration]]
     - [[#clang-format][clang-format]]
     - [[#company-clang-and-flycheck][Company-clang and flycheck]]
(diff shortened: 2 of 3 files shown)


beta1440:
@d12frosted is there someone who can review this pull request?

d12frosted:
I am totally ok with this PR. The only thing that I am not confident with is `spacemacs/clang-format-line-or-region`. Usually there are functions that work either on buffer or region and separate ones for lines. 

How do others feel about it?

myrgy:
when I'm using clang-format -- it's usually for the whole buffer or selection
I never faced one-line formatting case because I can just format whole document.
(actually we have code-style check - so if file formatter partially it will not pass the check)

beta1440:
@myrgy 
> UPD: my proposal if to change hotkeys a bit. Add SPC m f to format buffer.

According to [Spacemacs conventions](https://github.com/syl20bnr/spacemacs/blob/master/doc/CONVENTIONS.org#code-formatting), `SPC m =` is the preferred formatting key binding.

>UPD2: and distinct it from clang-mode, because ex. I'm using rtags (authentication, flycheck, refactoring) with clang-format.

Thats outside the scope of the pull request. Feel free to create an issue.

beta1440:
@robbyoconnor what do you think of these key bindings?

myrgy:
@beta1440 , in that case we can at least use some repeating key
for example `SPC m = = `
so make default case the as simple as possible.
What do you think?

beta1440:
@myrgy could you expand on what you mean by "making the default case as simple as possible"?

beta1440:
@JAremko or @bmag what are your thoughts on this pull request?

myrgy:
@beta1440 , 
in my case I usually format whole file. I'm using you MR locally (it's good (y) ), and one observation - I had to press
`, = b`
3 different keys
I would like to simplify that keychord - for example `, = =`
I'm not agains `= b`, my proposal is to make additional shortcut which will just format all code.

beta1440:
@myrgy that makes sense. I'll apply your changes.

beta1440:
I've updated the functions

layers/+lang/c-c++/packages.el
```diff
@@ -65,8 +65,13 @@
   (use-package clang-format
     :if c-c++-enable-clang-support
     :init
-    (when c-c++-enable-clang-format-on-save
-      (spacemacs/add-to-hooks 'spacemacs/clang-format-on-save c-c++-mode-hooks))))
+    (progn
+      (when c-c++-enable-clang-format-on-save
+        (spacemacs/add-to-hooks 'spacemacs/clang-format-on-save c-c++-mode-hooks))
+      (dolist (mode c-c++-modes)
```
myrgy: one more useful thing - otherwise spacemacs will show "prefix"
```(spacemacs/declare-prefix-for-mode mode "m=" "format")```

beta1440: Done

beta1440:
@d12frosted what are your thoughts on this pull request?

beta1440:
@syl20bnr please take a look at this pull request when you have time.

myrgy:
@syl20bnr , please, merge this PR. It's really useful, I'm using it every day!

syl20bnr:
Thank you ! 👍 

PR #9344, opened 2017-08-03
osx: add option to separate OSX pasteboard from kill-ring

### LINKED ISSUE #5750: Ability to yank / delete and *not* overwrite clipboard
Often I do things in this order ->
1. Copy text outside emacs
2. Switch back to emacs, delete some text 
3. Paste
Problem is I get back the deleted text since it overwrites the clipboard. 

I want to just keep the two things completely separate. vim commands/emacs will deal with only the ring buffer stuff, my clipboard is untouched. That way I can setup a explicit keybinding when I actually do want to copy over something to my clipboard like with c-c (win) or s-c (mac).

How can I do this? I tried in my `.spacemacs` file:
```
  (setq save-interprogram-paste-before-kill t)
  (setq x-select-enable-clipboard nil)
```

But it's not working for me. :(

### LINKED ISSUE #6533: Configurable option to keep the macOS and spacemacs clipboards separate
Feature request, as mentioned at the bottom of [spacemacs/layers/osx/README.org](https://github.com/syl20bnr/spacemacs/blob/a298a5a311dfe8b449b5f538740303f1ab3d5695/layers/osx/README.org)

- <kbd>command</kbd>+<kbd>c</kbd> and <kbd>command</kbd>+<kbd>v</kbd> would copy and paste to and from the system clipboard, respectively
- Keyboard sequences like `"+yy` and `"+p` would use the `+` register, connected to the system clipboard.
- Otherwise, `y` and `p` would copy/paste to/from the kill ring, separate from the system clipboard.

----
keywords:  macOS OS X OSX killring yank yankring copy paste buffer

sgeb:
This commit adds a new option `osx-use-separate-clipboards` to the `osx`
layer. When set to non nil, the layer treats the OSX pasteboard separate
from the spacemacs kill-ring. Default is `nil`, which is the same
behaviour as without this commit, hence not expecting any breaking
changes.

This commit also includes changes to the README file, in response to a
previously failing test (document must have a `Features:` section).

Fixes #6533
Fixes #5750

### DIFF
--- a/layers/+os/osx/keybindings.el
+++ b/layers/+os/osx/keybindings.el
@@ -63,20 +63,24 @@ other than the three sane values listed above."
                             (alt   . "A-")))))
         (when found (kbd (concat (cdr found) keys)))))
 
+    ;; Keybindings for copy/cut/paste
+    ;; Uses the + register to operate on the system pasteboard
+    (global-set-key (kbd-mac-command "v") (kbd "C-r +"))
+    (global-set-key (kbd-mac-command "c") (kbd "\"+y"))
+    (global-set-key (kbd-mac-command "x") (kbd "\"+d"))
+
     ;; Keybindings
     (global-set-key (kbd-mac-command "=") 'spacemacs/scale-up-font)
     (global-set-key (kbd-mac-command "-") 'spacemacs/scale-down-font)
     (global-set-key (kbd-mac-command "0") 'spacemacs/reset-font-size)
     (global-set-key (kbd-mac-command "q") 'save-buffers-kill-terminal)
-    (global-set-key (kbd-mac-command "v") 'yank)
-    (global-set-key (kbd-mac-command "c") 'evil-yank)
     (global-set-key (kbd-mac-command "a") 'mark-whole-buffer)
-    (global-set-key (kbd-mac-command "x") 'kill-region)
     (global-set-key (kbd-mac-command "w") 'delete-window)
     (global-set-key (kbd-mac-command "W") 'delete-frame)
     (global-set-key (kbd-mac-command "n") 'make-frame)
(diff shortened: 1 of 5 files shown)


JAremko:
The thing wants `Features:` list  https://travis-ci.org/syl20bnr/spacemacs/jobs/260455496#L747 (like in https://github.com/syl20bnr/spacemacs/blob/develop/core/templates/README.org.template)

d12frosted:
@JAremko this is out of scope of this PR 😸

JAremko:
@d12frosted  It's just an additional test in the `spacefmt` chain. When an org file gets changed in a PR it will be tested and any error will be reported(even if it was there before the PR). This is how we eventually got links and formatting in order, by incrementally fixing stuff. But may be it is too much extra works for PR authors :thinking:

d12frosted:
@JAremko yeah, I understand it 😸

sgeb:
No worries, I'll add a `Features:` entry to the README and update the PR. +1 to incrementally fixing stuff.

sgeb:
All green now. I took the opportunity to add missing pieces in the README and update the list of keybindings.

d12frosted:
@sgeb thanks! ❤️ 💯

layers/+os/osx/keybindings.el
```diff
@@ -63,20 +63,24 @@ other than the three sane values listed above."
                             (alt   . "A-")))))
         (when found (kbd (concat (cdr found) keys)))))
 
+    ;; Keybindings for copy/cut/paste
+    ;; Uses the + register to operate on the system pasteboard
+    (global-set-key (kbd-mac-command "v") (kbd "C-r +"))
```
aaronjensen: Will these work in hybrid or holy?

PR #9348, opened 2017-08-03
Python: Fix pyenv-executable-find in presence of pyenv-which-ext.

cpaulik:
Pyenv-which-ext falls back to the system executable if a executable is not found in the current environment but is installed on the system. This meant that it would always try to execute e.g. the system `ipython` if `ipython` was not installed in a certain environment. 

This PR should fix the problem since it only selects the pyenv executable if it is in the same path as the currently activated environment.


syl20bnr:
Thank you ! 👍 

PR #9353, opened 2017-08-03
core: mocker: Fall back on cl-flet instead of flet

beta1440:
The `flet` function has been deprecated and has been replaced with `cl-flet`.


beta1440:
Actually, this should be submitted to the [mocker upstream](https://github.com/sigma/mocker.el).

PR #9354, opened 2017-08-03
emacs-lisp: Speed up semantic

### LINKED ISSUE #1907: When enable semantic in emacs-lisp buffer, the completion is very slow
If I enable semantic layer or manually enable semantic mode, the company mode completion is very very slow when editing elisp code. It will try to parse all the elisp in Emacs installation  and ELPA directory.

And even in elisp comments, the company tries to use `company-capf` as the backend.
If I disable semantic layer, it will use `company-dabbrev` when completing comments.

@tuhdo 
Toughts?

delaanthonio:
Resolve #1907.

Emacs 25 sets `semanticdb-find-default-throttle` as a mode-local variable in
`emacs-lisp-mode` and includes the `omniscient` throttle which slows down indexing.
Let's override the mode-local variable so it doesn't include the `omniscient`
throttle.

See #7736 for additional context.

### DIFF
--- a/layers/+lang/emacs-lisp/packages.el
+++ b/layers/+lang/emacs-lisp/packages.el
@@ -201,7 +201,10 @@
 (defun emacs-lisp/post-init-semantic ()
   (add-hook 'emacs-lisp-mode-hook 'semantic-mode)
   (with-eval-after-load 'semantic
-    (semantic-default-elisp-setup)))
+    (semantic-default-elisp-setup)
+    (setq-mode-local emacs-lisp-mode
+                     semanticdb-find-default-throttle
+                     '(file local project unloaded system))))
 
 (defun emacs-lisp/post-init-srefactor ()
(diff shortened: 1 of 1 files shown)


syl20bnr:
What do we loose when we deactivate this throttle ? I found this https://www.gnu.org/software/emacs/manual/html_node/semantic/Search-Throttle.html but it does not tell what feature are enabled.

beta1440:
I'm not exactly sure what we lose when we deactivate this throttle either. Although I've noticed that Semantic no longer parses several Elisp files when I'm spelling out a word in Elisp-mode.

duianto:
The PR:
Improve the use of semantic #11511
is now on the develop branch.

The author stated that this PR might have been made obsolete:
>This pull request has the same effects as #9354, except that it works globally rather that only works in emacs-lisp-mode, and it's more clear what we have done and what we will lose.

Issue #9346, opened 2017-08-03
Crowd fixing documentation


JAremko:
We have [an average pile of documentation issues](https://gist.githubusercontent.com/JAremko/d8462b9640bee4b0b91c495e2a86e4e6/raw/aacbb9be308e4de7c238d8fe7da70c6951cc1f0a/res.txt)  and we need to fix them **YESTERDAY**

So how about.... We all gab some and fix them? :shipit:  

Proper [README.org example](https://github.com/syl20bnr/spacemacs/blob/develop/core/templates/README.org.template)
- - - - - - -

## *Check other people's lists*

- - - - - -

# thx! :heart:

benjamreynolds:
Fixing emoji layer...
```
File "~/.emacs.d/layers/+fun/emoji/README.org" doesn't have "** Features:" list in the top level "Description" headline
```
and slack layer...
```
Link "file:../../../doc/DOCUMENTATION.org::#layouts-and-workspaces" in the file "~/.emacs.d/layers/+chat/slack/README.org" should target the org file at GitHub (GitHub style anchors are supported)
File "~/.emacs.d/layers/+chat/slack/README.org" doesn't have "** Features:" list in the top level "Description" headline
```

benjamreynolds:
@JAremko I originally opened #9359 to fix a few more of these but it looks like a handful of them were fixed by you in 2ef230369acc2bf2b73eebc20652afb26ec42bf1. Can you re-run whatever test generates this output? (or is it documented somewhere?)

JAremko:
@benjamreynolds  I updated the list :+1: 

- - - - - -

> a handful of them were fixed by you in 2ef2303  

Sorry, I forgot that I had to fix some beforehand :disappointed:

benjamreynolds:
Grabbing another chunk of these...
```
Link "file:doc/FAQ.org" in the file "~/.emacs.d/CONTRIBUTING.org" should target the org file at GitHub (GitHub style anchors are supported)
Link "file:doc/DOCUMENTATION.org" in the file "~/.emacs.d/CONTRIBUTING.org" should target the org file at GitHub (GitHub style anchors are supported)
Link "file:doc/LAYERS.org" in the file "~/.emacs.d/CONTRIBUTING.org" should target the org file at GitHub (GitHub style anchors are supported)
[...]

PR #9356, opened 2017-08-03
Fix documentation issues with slack and emoji layers

benjamreynolds:
See #9346


JAremko:

PR #9359, opened 2017-08-04
Fix relative file links to github style in Docs

benjamreynolds:
See #9346


benjamreynolds:
Oops looks like these were already resolved in 2ef230369acc2bf2b73eebc20652afb26ec42bf1

PR #9361, opened 2017-08-04
core-spacemacs-buffer: Bind SPC to spacemacs-cmds

beta1440:


bmag:
The idea is that `SPC` in the home buffer will work in emacs-style the same as in vim-style and hybrid-style, right? So in the home buffer, the usual `SPC` menu is available also in Holy mode?

beta1440:
@bmag Exactly. `SPC` in `holy-mode` inserts a space. Since the `spacemacs-buffer` is read only, `SPC` doesn't do anything.

bmag:
On the one hand, having `SPC` work on the home buffer even for emacs-style users is fitting, because of the emphasis Spacemacs has on the space key. On the other hand, there are some downsides. First, we let users customize the leader key, so then ignoring their choice in the home buffer is unexpected. Second, why just in the home buffer and not in other special buffers? It's inconsistent.

All in all, I think the idea is cool, but still causes more harm than good. Besides, `M-m` is still there. Thanks for the contribution, but I'm closing this PR.

PR #9364, opened 2017-08-04
Remove R repl binding C-d to delete-char

jackkamm:
This removes the keybinding of `C-d` to `delete-char` within the R repl buffer (`inferior-ess-mode`).

Currently, `C-d` is bound to `delete-char` in `inferior-ess-mode-map`, which overrides the vim-style motion `evil-scroll-down` within the R repl buffer. The PR fixes this behavior.

### DIFF
--- a/layers/+lang/ess/packages.el
+++ b/layers/+lang/ess/packages.el
@@ -115,7 +115,8 @@
         "cn" 'ess-noweb-next-chunk))
     (define-key ess-mode-map (kbd "<s-return>") 'ess-eval-line)
     (define-key inferior-ess-mode-map (kbd "C-j") 'comint-next-input)
-    (define-key inferior-ess-mode-map (kbd "C-k") 'comint-previous-input)))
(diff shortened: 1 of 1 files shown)


beta1440:
What if someone wants to use the Emacs key bindings?

jackkamm:
Emacs key bindings still works (I use hybrid mode and C-d still deletes forward there).
This simply removes `C-d` from the `inferior-ess-mode-map` which overrides the other keymaps in `inferior-ess-mode`.
I do not know why ESS manually assigned `C-d` to `inferior-ess-mode-map` when `delete-char` work fine without this additional binding. The binding was added over 2 decades ago (sometime before 1997, I just went through the ESS github history) and is probably no longer needed. The exact line where the binding is added is in `ess-inf.el`:
```
    (define-key map "\C-d"     'delete-char)   ; EOF no good in S
```
which suggests it has something to do with S programming language. My understanding is that ESS is written to be compatible not only with R but also with S, SAS, and other related languages, however S (the parent language of R) is quite obsolete and I have never met anyone who actually uses it.

PR #9369, opened 2017-08-05
Fix links in BEGINNERS_TUTORIAL.org

planigan:
See #9346


JAremko:

PR #9370, opened 2017-08-05
add debug layer

CeleritasCelery:
@sukeyisme originally created this layer with #8799. However his implementation only supported the c-c++ layer, so @syl20bnr merged it into that layer instead of creating a new one. I have since refactored the code so that this functionality can be added to any language. I updated the `c-c++` layer to use the new method for adding debuggers and also added support for perl5.

This should be its own layer since realgud supports many languages.

I have added a method called `spacemacs|add-realgud-debugger` that can be called from any language layer to add support for their respective debuggers. If language layer maintainers want to add debugger support they can use this call in a `pre-init-realgud` function.


CeleritasCelery:
this was an attempt to clean up my other request

PR #9371, opened 2017-08-05
Add insert commands from sh-script package to major mode menu

smile13241324:
The shell-scripts layer provides the sh-script package which defines a lot of useful functions to insert standard script blocks like case statements or specific loops. However those are not bound to the major-mode menu and will probably be overlooked by all who do not read the mode description.

I have therefore added those to the major mode menu in the category `mi insert`. I have also added a binding to the "insert-shebang" function in there though a binding for this function is already available in `spc i !` I think it should also be mentioned in the major-mode menu as most users will only have a look there.


d12frosted:

Issue #9322, opened 2017-07-31
How to use helm swoop transient state's `edit occurrences` feature? (not working for now)


MilooHub:
I find `Helm Transient State`(press `M-SPC` while helm swoop buffer appear) of `Helm Swoop` a great way to deal with candidates matched by Helm Swoop command.

One could move cursor up/down to navigate different candidates easily.

But the `edit occurrences` feature(press `e` in Helm Transient State) never works for me. No matter how many times I pressed `e` trying to edit the candidates matched by my input pattern, Nothing Happens!

And I can not find detailed description on this **edit occurrences** feature with google.

So if any spacemacs guru could provide any information on this, it would be great help for sure.

Thanks in advance.

CeleritasCelery:
I don't know about the transient state but I can edit occurrences with `c-c c-e`

CeleritasCelery:
upon deeper investigation is looks like Spacemacs helm transient state only supports edits for helm-ag operations (`SPC s f`) I am going to open a PR to fix this.

MilooHub:
Thanks for digging deep into it! And it is absolutely great to finally see this feature actually works in current buffer and more general and broad context!

Thanks for the PR!

p.s. yeah `C-c C-e`(helm swoop occur/edit) works here as well, although helm transient state builtin edit should look like a much more convenient and productivity-boosting feature. And `M-SPC` then `e` to edit occurences is mentioned in Spacemacs official documentation on Helm Transient State section anyway. So I figured there would be more users expect this feature becoming available.

PR #9379, opened 2017-08-06
fixed helm transient state swoop edit

CeleritasCelery:
As mentioned in #9322 the `edit occurrences` command of the helm transient state is not calling `helm-swoop-edit` when using helm swoop. This fixes that.


d12frosted:

Issue #8655, opened 2017-04-07
Set variable `company-c-headers-path-system` failed


Voleking:
#### Description :octocat:

To fix issue #4493, I set `company-c-headers-path-system '"/usr/include/c++/4.2.1"` in `user-config`
But I fail to set this.

#### Reproduction guide :beetle:
- Start Emacs
- open an cpp file
- describe-variable `company-c-headers-path-system`

*Observed behaviour:* :eyes: :broken_heart:
```
Its value is ("/usr/include" "/usr/local/include")
Original value was 
("/usr/include/" "/usr/local/include/")

Local in buffer test.cpp; global value is 
"/usr/include/c++/4.2.1"
```

*Expected behaviour:* :heart: :smile:
Maybe?
```
Its value is ("/usr/include/c++/4.2.1")
Original value was 
("/usr/include/" "/usr/local/include/")

Local in buffer test.cpp; global value is 
"/usr/include/c++/4.2.1"
```

#### System Info :computer:
- OS: darwin
- Emacs: 25.2.1
- Spacemacs: 0.200.8
- Spacemacs branch: master (rev. bd7ef98e)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(helm osx chinese dash git github pandoc
      (c-c   :variables c-c  -default-mode-for-headers 'c  -mode c-c  -enable-clang-support t)
      emacs-lisp java
      (markdown :variables markdown-command "pandoc" markdown-live-preview-engine 'pandoc)
      org python latex
      (shell :variables shell-default-height 30 shell-default-position 'bottom)
      (syntax-checking :variables flycheck-delayed-errors-timeout 1000))
```
- System configuration features: NOTIFY ACL LIBXML2 ZLIB TOOLKIT_SCROLL_BARS

Voleking:
Well, everything work fine when I reinstall my macOS, so I think it's not spacemacs issue. I'd like to close it.

Voleking:
Sorry for my fault

bo-hong:
@Voleking Hi,I have the same issues with yours ,could you give me a hand? Company-C-Headers doesn't work...

Voleking:
@bo-hong Sorry for replying you late. Currently everything works fine on my computer.

- OS: darwin
- Emacs: 26.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 390462e36)
- Graphic display: t
[...]

Issue #4493, opened 2016-01-08
Some bugs or questions about C-C++ layer


jcppython:
Here is my configuration
![screenshot from 2016-01-08 18 26 36](https://cloud.githubusercontent.com/assets/5669903/12195856/676b8bc4-b635-11e5-8007-9f63e0b5d68f.png)

emacs-version
GNU Emacs 24.4.1 (x86_64-unknown-linux-gnu, GTK+ Version 3.10.8) 
1. The "SPC m" is undefined when I open a C/C++ header file. 
   In addition, the illustration of “Key Bindings” part has some wrongs in C-C++.
   https://github.com/fenxian/spacemacs/blob/master/layers/+lang/c-c++/README.org
2. It can't find the C++ headers. 
   ![screenshot from 2016-01-08 18 34 39](https://cloud.githubusercontent.com/assets/5669903/12196043/9c119444-b636-11e5-93a3-81727f22a8cc.png)
   I need to configure it?
3. An unfriendly tip as shown in following:
   ![screenshot from 2016-01-08 18 31 51](https://cloud.githubusercontent.com/assets/5669903/12195989/5c235264-b636-11e5-89d7-b042dda4bd42.png)
   
   And, the code is right. In other words, this tip shouldn't exist in here.
   In addition, how to open a header file? command or key binding?
4. Can you give me more information about how to scan a big C/C++ project with spacemacs?
   For  example, how to configure it for my project? 
   How to jump to the definition of references?

Thank you~

tuhdo:
1. It's strange. If you enabled `semantic` layer, at least `SPC m r` should be available to access `srefactor`'s menu.

2. For C++ header files, you must add your C++ header paths to the variable `company-c-headers-path-system` for it to complete. Something like this:

```elisp
(add-to-list 'company-c-headers-path-system "/usr/include/c++/4.8/" )
```

Because it's is system-specific, we cannot preconfigured it.

3. As for the error message, probably it cannot find the header file. You might try adding project include paths for your C++ using the variable `flycheck-clang-include-path`.

4. To jump around C++ source code, you can either use the `gtags` layer (after adding the layer, some `SPC m` key bindings for `gtags` should be available. Use the command `helm-gtags-create-tags` to generate a tag database at your project root. Then, you will be able to jump to definitions and find references.

fenxian:
1. 
[...]

PR #9381, opened 2017-08-06
c-c++: Improve the header include auto completion

fredrikbergstrand:
This change makes the c-c++ layer configure company-c-headers to search
for and list completion alternatives according to the priority list
found in 'man gcc'. Default system include paths are also fetched from
gcc's configuration instead of being hard coded to "/usr/include" and
"/usr/local/include". This also remedies the problem where the
C++ standard library headers does not show up as completion alternatives
without manual addition of include paths (see #4493, #8655).

### DIFF
--- a/layers/+lang/c-c++/funcs.el
+++ b/layers/+lang/c-c++/funcs.el
@@ -34,12 +34,13 @@
   "Load the flags from CC-FILE, one flag per line."
   (let ((invocation-dir (expand-file-name (file-name-directory cc-file)))
         (case-fold-search nil)
+        (include-regex "\\(-I\\|-isystem\\|-iquote\\|-idirafter\\)\\s-*\\(\\S-+\\)")
         compile-flags)
     (with-temp-buffer
       (insert-file-contents cc-file)
       ;; Replace relative paths with absolute paths (by @trishume)
       ;; (goto-char (point-min))
-      (while (re-search-forward "\\(-I\\|-isystem\n\\)\\(\\S-+\\)" nil t)
+      (while (re-search-forward include-regex nil t)
         (replace-match (format "%s%s" (match-string 1)
                                (expand-file-name (match-string 2)
                                                  invocation-dir))))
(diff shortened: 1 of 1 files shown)


layers/+lang/c-c++/funcs.el
```diff
@@ -53,6 +54,21 @@
                     (split-string (buffer-string) "\n" t))))
     compile-flags))
 
+(defun spacemacs//c-c++-get-standard-include-paths (lang)
+  (let* ((start "#include <...> search starts here:")
```
d12frosted: Please add doc string.

layers/+lang/c-c++/funcs.el
```diff
@@ -53,6 +54,21 @@
...
+         (end "End of search list.")
+         (gcc-tmplt "echo | gcc -x%s -E -v - 2>&1")
+         (sed-tmplt " | sed -n '/%s/,/%s/{/%s/b;/%s/b;p}' | sed -e 's/^ *//g'")
+         (template (concat gcc-tmplt sed-tmplt))
+         (inc-dirs-cmd (format template lang start end start end))
+         (inc-dirs (shell-command-to-string inc-dirs-cmd)))
+    (split-string inc-dirs "\n" t)))
+
+(defun spacemacs//filter-and-substring (flags filter-prefix substr-index)
+  (mapcar (lambda (f) (substring f substr-index))
```
d12frosted: Please add doc string.

layers/+lang/c-c++/funcs.el
```diff
@@ -53,6 +54,21 @@
                     (split-string (buffer-string) "\n" t))))
     compile-flags))
 
+(defun spacemacs//c-c++-get-standard-include-paths (lang)
+  (let* ((start "#include <...> search starts here:")
+         (end "End of search list.")
+         (gcc-tmplt "echo | gcc -x%s -E -v - 2>&1")
```
d12frosted: Does it work with all versions of `gcc`?

fredrikbergstrand: Looking into the GCC source, the lines that print the include directories were last changed in January 2003. GCC 3.2.2 was released February 2003. I don't know if that is sufficient to you. I was thinking that perhaps some sort of check should be done for the existence of GCC and sed on the system, and then make sure that the extracted lines really are absolute search paths. If one of these checks fail, a default of /usr[/local]/include could be returned.

syl20bnr:
Thank you ! 👍 

PR #9386, opened 2017-08-06
Add option to enable separate shells for each layout

pspencil:
I tested the functions in dotfile `user-config` but not after changing the files since I do not know how to create an isolated environment to test the change I made.

### DIFF
--- a/layers/+spacemacs/spacemacs-layouts/config.el
+++ b/layers/+spacemacs/spacemacs-layouts/config.el
@@ -34,6 +34,9 @@
 (defvar spacemacs--layouts-autosave-timer nil
   "Timer for layouts auto-save.")
 
+(defvar spacemacs-persp-unique-shell nil
+  "If true, the default shell for each layout will have the same 'index' as the layout.")
+
 (defvar spacemacs-generic-layout-names
   '(("zebra" "zucchini" "zen" "yellow" "yeti" "yard") ; grab-bag
     ("baboon" "banana" "blue")                        ; 2nd layout
--- a/layers/+spacemacs/spacemacs-layouts/config.el
+++ b/layers/+spacemacs/spacemacs-layouts/config.el
@@ -35,7 +35,7 @@
   "Timer for layouts auto-save.")
 
 (defvar spacemacs-persp-unique-shell nil
-  "If true, the default shell for each layout will have the same 'index' as the layout.")
+  "If true, each layout will have its own unique shell.")
 
 (defvar spacemacs-generic-layout-names
   '(("zebra" "zucchini" "zen" "yellow" "yeti" "yard") ; grab-bag
--- a/layers/+spacemacs/spacemacs-layouts/funcs.el
+++ b/layers/+spacemacs/spacemacs-layouts/funcs.el
@@ -619,5 +619,5 @@ FRAME defaults to the current frame."
           do (progn (setq shell-pop-last-shell-buffer-index i) (return)))))
 
 (defun spacemacs//set-persp-specific-shell-create (arg)
-  "Same as above. But when user creates new layouts, spacemacs/layout-switch-by-pos is called instead of persp-switch. Two functions are thus needed."
+  "Same as spacemacs//set-perp-specific-shell. But when user creates new layouts, spacemacs/layout-switch-by-pos is called instead of persp-switch. Two functions are thus needed."
   (setq shell-pop-last-shell-buffer-index (+ 1 arg)))
--- a/layers/+spacemacs/spacemacs-layouts/funcs.el
+++ b/layers/+spacemacs/spacemacs-layouts/funcs.el
@@ -608,3 +608,16 @@ FRAME defaults to the current frame."
                                         (eyebrowse--get 'last-slot frame))
                                   (get-frame-persp frame)
                                   frame))
+
+;; shell-pop integration
+
+(defun spacemacs//set-persp-specific-shell (arg)
+  "To be called after persp-switch. Make the default shell number to use equal to the layout number."
+  (progn
+    (loop for persp in (persp-names-current-frame-fast-ordered) and i from 1
+          when (string= persp arg)
+          do (progn (setq shell-pop-last-shell-buffer-index i) (return)))))
+
+(defun spacemacs//set-persp-specific-shell-create (arg)
+  "Same as above. But when user creates new layouts, spacemacs/layout-switch-by-pos is called instead of persp-switch. Two functions are thus needed."
+  (setq shell-pop-last-shell-buffer-index (+ 1 arg)))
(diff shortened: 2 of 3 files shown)


layers/+spacemacs/spacemacs-layouts/packages.el
```diff
@@ -209,7 +209,12 @@
       (spacemacs|define-custom-layout "@Spacemacs"
         :binding "e"
         :body
-        (spacemacs/find-dotfile)))
+        (spacemacs/find-dotfile))
+      (when (and spacemacs-persp-unique-shell
+               (configuration-layer/package-usedp 'shell-pop))
+        (progn
```
aaronjensen: `progn` is unnecessary here, `when` can take multiple forms

layers/+spacemacs/spacemacs-layouts/config.el
```diff
@@ -34,6 +34,9 @@
 (defvar spacemacs--layouts-autosave-timer nil
   "Timer for layouts auto-save.")
 
+(defvar spacemacs-persp-unique-shell nil
+  "If true, the default shell for each layout will have the same 'index' as the layout.")
```
aaronjensen: I would talk more about what it actually buys the user rather than its technical implementation. Something like "If true, each layout will have its own unique shell."

layers/+spacemacs/spacemacs-layouts/funcs.el
```diff
@@ -608,3 +608,16 @@ FRAME defaults to the current frame."
...
+

+;; shell-pop integration
+
+(defun spacemacs//set-persp-specific-shell (arg)
+  "To be called after persp-switch. Make the default shell number to use equal to the layout number."
+  (progn
+    (loop for persp in (persp-names-current-frame-fast-ordered) and i from 1
+          when (string= persp arg)
+          do (progn (setq shell-pop-last-shell-buffer-index i) (return)))))
```
aaronjensen: Could `pop-shell-last-shell-buffer-name` be used instead? It seems it would be better if the shells and persps could be tied by name rather than index.

layers/+spacemacs/spacemacs-layouts/funcs.el
```diff
@@ -608,3 +608,16 @@ FRAME defaults to the current frame."
...
+
+(defun spacemacs//set-persp-specific-shell-create (arg)
+  "Same as above. But when user creates new layouts, spacemacs/layout-switch-by-pos is called instead of persp-switch. Two functions are thus needed."
```
aaronjensen: I'd replace "Save as above" with a reference to the function or just repeat it directly.

layers/+spacemacs/spacemacs-layouts/funcs.el
Ps-Pencil: No. The shell-pop package only uses index to find buffers.

layers/+spacemacs/spacemacs-layouts/packages.el
Ps-Pencil: Thanks. Will fix.

layers/+spacemacs/spacemacs-layouts/config.el
Ps-Pencil: Make sense. Will fix.

layers/+spacemacs/spacemacs-layouts/funcs.el
Ps-Pencil: Ok.

layers/+spacemacs/spacemacs-layouts/funcs.el
aaronjensen: Ah, I misread the code, you are right. 

That said, I wonder if it would be better to update the `make-shell-pop-command` macro to call `shell-pop` with the appropriate index (which is looked up by way of some persp name map.

I'm assuming this still has the problem with deleting or reordering perspectives causing the shells to get mismatched, or does that work now?

Ps-Pencil: No. It doesn't work. The shell popped will be the current layout number. I don't know enough about emacs lisp to create a map and get/set values. But changing the `make-shell-pop-command` seems like a better idea. 

Btw, inside `make-shell-pop-command`, there is an `if` that says no popup if index equals 4 which I do not get.

layers/+spacemacs/spacemacs-layouts/packages.el
```diff
@@ -209,7 +209,11 @@
       (spacemacs|define-custom-layout "@Spacemacs"
         :binding "e"
         :body
-        (spacemacs/find-dotfile)))
+        (spacemacs/find-dotfile))
+      (when (and spacemacs-persp-unique-shell
+                 (configuration-layer/package-usedp 'shell-pop))
+        (advice-add 'persp-switch :after #'set-persp-specific-shell)
```
jonyamo: When I tried running this locally I kept getting an error that the `set-persp-specific-shell` definition is void, so I updated this to be `spacemacs//set-persp-specific-shell` and then it worked like a charm 👌 . This is a life saver by the way! Been looking for something like this for awhile... well done.

aaronjensen: @jonyamo fwiw, I do have that function. I'm on `develop` though, maybe you're on `master`?

aaronjensen:
@syl20bnr where is the right place to document this and/or allow for configuration? Would it go in `dotspacemacs/init`?

layers/+spacemacs/spacemacs-layouts/packages.el
jonyamo: @aaronjensen I'm on `develop` as well, but don't see that function defined anywhere... perhaps I have something amiss with my project. Would you mind pointing me to its definition?

aaronjensen: err, my bad. It's defined in this patch and it was changed. It looks like you're right, this line does need to change to `spacemacs//set-persp-specific-shell` cc @Ps-Pencil

CeleritasCelery:
FYI, I changed `make-shell-pop-command` so that it creates shells that are layout local. If I use it for a while and it works fine I may open a PR.

aaronjensen:
@CeleritasCelery did you ever open a PR? It'd be great to get this behavior merged in.

CeleritasCelery:
I have not yet, but I have been using it for the past couple of months and it works great. When I get some time I will open a PR.

CeleritasCelery:
I just opened a PR with this functionality in #10457 . If anyone would like to review it that would be great.

Compro-Prasad:
Any updates on this? Should this be closed?
@aaronjensen @CeleritasCelery @jonyamo @Ps-Pencil

CeleritasCelery:
I would close this in favor of #10457. But I may be a little biased 😉

duianto:
We'll close this PR, because the following PRs changes, have been applied to the `develop` branch:
Local shell pop #13556

PR #9387, opened 2017-08-06
Defer git-gutter loading

aaronjensen:
As it was, git-gutter and git-gutter+ get started as soon as emacs starts if
they're enabled. This added about 500ms to startup time on my machine, so this
will instead defer it until emacs has been idle for 5 seconds.

This could theoretically be an option, but I would guess that most would rather have startup time over an immediate gutter.

I'm also open to other ideas on how to defer its loading.

### DIFF
--- a/layers/+source-control/version-control/packages.el
+++ b/layers/+source-control/version-control/packages.el
@@ -54,7 +54,7 @@
       ;; If you enable global minor mode
       (when (and (eq version-control-diff-tool 'git-gutter)
                  version-control-global-margin)
-        (global-git-gutter-mode t))
+        (run-with-idle-timer 1 nil 'global-git-gutter-mode t))
       (setq git-gutter:update-interval 2
             git-gutter:modified-sign " "
             git-gutter:added-sign "+"
@@ -105,14 +105,14 @@
 
 (defun version-control/init-git-gutter+ ()
   (use-package git-gutter+
-    :commands (global-git-gutter+-mode git-gutter+-mode)
+    :commands (global-git-gutter+-mode git-gutter+-mode git-gutter+-refresh)
     :init
     (progn
       ;; If you enable global minor mode
       (when (and (eq version-control-diff-tool 'git-gutter+)
                  version-control-global-margin)
         (add-hook 'magit-pre-refresh-hook 'git-gutter+-refresh)
-        (global-git-gutter+-mode t))
+        (run-with-idle-timer 1 nil 'global-git-gutter+-mode t))
       (setq
        git-gutter+-modified-sign " "
        git-gutter+-added-sign "+"


d12frosted:
I feel your pain, but I am not sure about your solution. Ideally it should be enabled once you are editing file in git repo. Deferring by 5 seconds doesn't sound ideal. 

Let's wait for more input from other users.

aaronjensen:
I could reduce the time and/or make it an option. I personally do not rely on the gutter for any sort of decision making, it's merely a nice-to-have, so the delay is inconsequential for me.

aaronjensen:
Another option is a one-time `find-file-hook`. This would make this only beneficial for a `SPC q R` or a clean start, so I wouldn't consider that ideal for my needs either.

layers/+source-control/version-control/packages.el
```diff
@@ -105,14 +106,15 @@
 
 (defun version-control/init-git-gutter+ ()
   (use-package git-gutter+
-    :commands (global-git-gutter+-mode git-gutter+-mode)
+    :commands (global-git-gutter+-mode git-gutter+-mode git-gutter+-refresh)
```
bmag: Are there any benefits to adding an autoload for `git-gutter+-refresh`? I don't see any...

aaronjensen: Yes, without it, the `magit-pre-refresh-hook` below will fail if you open magit before git-gutter+ loads from the idle timer.

layers/+source-control/version-control/packages.el
```diff
@@ -54,7 +54,8 @@
       ;; If you enable global minor mode
       (when (and (eq version-control-diff-tool 'git-gutter)
                  version-control-global-margin)
-        (global-git-gutter-mode t))
+        (run-with-idle-timer 5 nil (lambda ()
```
bmag: Better pass a `:defer` to use-package declaration instead of calling `run-with-idle-timer` directly. It's cleaner.

layers/+source-control/version-control/packages.el
```diff
@@ -105,14 +106,15 @@
...
-    :commands (global-git-gutter+-mode git-gutter+-mode)
+    :commands (global-git-gutter+-mode git-gutter+-mode git-gutter+-refresh)
     :init
     (progn
       ;; If you enable global minor mode
       (when (and (eq version-control-diff-tool 'git-gutter+)
                  version-control-global-margin)
         (add-hook 'magit-pre-refresh-hook 'git-gutter+-refresh)
-        (global-git-gutter+-mode t))
+        (run-with-idle-timer 5 nil (lambda ()
```
bmag: See `:defer` comment above

layers/+source-control/version-control/packages.el
aaronjensen: And then move the `(global-git-gutter+-mode t)` to config? It'll still run init right away, right?

That seems like a good idea/use of use-package.

bmag:
At first it seems wrong, but I can't actually find anything wrong with deferring the gutter, so I'm ok with it. We already defer Helm by 1 second, so we probably should defer git-gutter by 1 second as well (not 5) so they are loaded together. (this way the user isn't interrupted in too many different times)

layers/+source-control/version-control/packages.el
bmag: I don't remember right now. If init runs right away then yes, move it to config.

layers/+source-control/version-control/packages.el
bmag: Oh, ok then.

layers/+source-control/version-control/packages.el
aaronjensen: Actually, this will cause git-gutter and git-gutter+ to load for everyone. I'd need to set the defer conditionally and I'm not sure of the impact of that if one were byte compiled that file (since use-package expands at compile time)

aaronjensen:
@bmag I updated the defer time to 1s, but did not use `:defer`. See my [comment](https://github.com/syl20bnr/spacemacs/pull/9387#discussion_r132719852) and let me know if you have any good ideas to work around that.

layers/+source-control/version-control/packages.el
aaronjensen: I suppose I could use `:if EXPR`, but that would change the behavior slightly (people could, if they wanted to, use `git-gutter` in some buffers and `git-gutter+` in others. If we were going to make it conditional, that could also be done in `version-control-packages`. Thoughts?

bmag: It's true that now people could use different diff tools for different buffers. I don't remember if that's intentional, but it might be. Ok, lets not use `:defer`, current situation is fine.

bmag:
@aaronjensen probably no better solution than current one. Just one last thing: we don't need a lambda to wrap `(some-mode t)`. Calling `some-mode` has the same effect:
```elisp
;; the following have the same effect:
(global-git-gutter-mode)
(funcall (lambda () (global-git-gutter-mode)))
```

aaronjensen:
Ah, I guess my only concern would be if they toggled it elsewhere in their config (say they didn't want the 1s defer 😄 ), because I believe `(global-git-gutter-mode)` toggles rather than sets truthy.

I'd be happy to switch it if you don't think toggling would be an issue.

aaronjensen:
@bmag just updated to make use of the `&rest ARGS` in `run-with-idle-timer`.

bmag:

I removed the additional `t` argument in the timers because it was not needed (see https://github.com/syl20bnr/spacemacs/commit/433ae7c30bc0884519018590942e516a4edbc14b)

aaronjensen:
@bmag I believe it is needed if the end user were to put `(global-git-gutter+-mode)` in their `user-config`. I believe your change would make it so that spacemacs would toggle it OFF after 1 second. As it was before, it would remain on. So, it was more correct before and causing no harm, unless I'm misunderstanding something.

This is the docstring:

```
global-git-gutter+-mode is an interactive autoloaded compiled Lisp function in
‘git-gutter+.el’.

(global-git-gutter+-mode &optional ARG)

Toggle Global Git-Gutter+ mode on or off.
With a prefix argument ARG, enable Global Git-Gutter+ mode if ARG is
positive, and disable it otherwise.  If called from Lisp, enable
the mode if ARG is omitted or nil, and toggle it if ARG is ‘toggle’.

Uses keymap ‘global-git-gutter+-mode-map’, which is not currently defined.
```

bmag:
>  your change would make it so that spacemacs would toggle it OFF after 1 second

is not true because:
> If called from Lisp, enable the mode if ARG is omitted or nil, and toggle it if ARG is ‘toggle’

My change didn't change the behavior.

aaronjensen:
ah! helps to read the whole doc string, thank you, sorry about the confusion.

PR #9399, opened 2017-08-10
adds global evil-leader mappings `aokr` for `org-resolve-clocks`

FrancescElies:
Hi,

I find `org-resolve-clocks` quite useful and  I would like to add a global evil-leader mapping for `aokr`.

Would you be fine with this addition?

Best wishes & thank in advance for your time


bmag:

PR #9406, opened 2017-08-11
Fix switch to home buffer moves prev buffers cursor

duianto:
Switching to the Spacemacs home buffer `SPC b h`, moves the previous buffers
cursor to the top.


bmag:
I never noticed that bug. I prefer that you put the `(if save-line ...)` block inside of the (`with-current-buffer ...)` block, instead of moving `switch-to-buffer`. The minor difference is that the buffer is scrolled before displaying it, instead of after.

bmag:
Thank you :+1: Cherry-picked into develop, you can safely delete your branch. @duianto I didn't get a notification that you pushed changes to the PR, so I didn't know you fixed it according to my comment, and that's why it took so long to merge.

duianto:
I should have posted a comment, that the PR had been updated. I'll be doing that from now on.

PR #8605, opened 2017-03-29
Fix keybinding to helm-cscope-find-calling-this-function

dubnde:
Typo which was incorrect before has now been fixed [upstream](https://github.com/alpha22jp/helm-cscope.el/commit/3cc7259ab4989f9f7ca039e703cdac14b907530a). Therefore the `cscope` layer needs this fix as it started throwing an error

PR #9411, opened 2017-08-11
Fix typo in cscope find-calling-this-function binding

mathewprokos:
Fixes a small typo in the cscope layer for the helm-cscope-find-calling-this-function binding


bmag:
Thanks, but this has already been fixed in develop, see #8605. By the way, PRs to Spacemacs must be made against develop branch, not master. That's why the tests are failing.

PR #9413, opened 2017-08-12
[Haskell] Fix Readme.org typos, missing words

duianto:


tonylotts:
This looks like an easy merge, and it's been over 2 months since submission.  Someone want to review?  @syl20bnr

duianto:
I fixed the merge conflict. The formatting test has been updated/added since I first made the PR, it found two extra newlines.

Compro-Prasad:
@duianto Still on it?

duianto:
Sorry for not checking back sooner.

I was focusing on to going through all of the [Systematic Program Design](https://www.youtube.com/channel/UC7dEjIUwSxSNcW4PqNRQW8w) videos.

@Miciah Thanks for the corrections, I hope I got them all.

Compro-Prasad:
@sdwolfz I think this should be a clean merge now.

layers/+lang/haskell/README.org
```diff
@@ -464,52 +468,53 @@ snippet into your =dotspacemacs/user-config= function.
...
-directory name is not the version of the cabal library installed by stack, but
-the version of cabal associated to the GHC version. This error can happen after
-upgrading cabal or cabal-install. To check if this is the problem, compare the
-path name of the build path used by flycheck using =flycheck-compile= and
-compare it to to the actual path in the =.stack-work= directory. If they are
-different you need to reinstall ghc using the command =stack setup
---upgrade-cabal=.
+However the version of the cabal library that's used by =stack= to generate the
+directory name, is not the version of the cabal library that's installed by
+=stack=. But it's the version of cabal that's associated to the GHC version. This
```
Miciah: I think these two sentence were clearer when written as one.  Whether you write it as one sentence or two, adding "rather" would help.  As one sentence, it would read, "However the version of the cabal library [...] is not [...] but rather [...]."  As two sentences, it would read, "However the version of the cabal library [...] is not [...].  Rather, it's [...]."

layers/+lang/haskell/README.org
```diff
@@ -464,52 +468,53 @@ snippet into your =dotspacemacs/user-config= function.
...
+error can happen after upgrading cabal or cabal-install. To check if this is the
+problem, compare the path name of the build path that's used by flycheck, using
+=flycheck-compile=, and compare it to the actual path, in the =.stack-work=
+directory. If they are different, then you'll need to reinstall ghc, using the
```
Miciah: That's a lot of commas, and most of them aren't necessary.  It would read better without the commas before the two instances of "using" and the comma after "actual path".

layers/+lang/haskell/README.org
```diff
@@ -464,52 +468,53 @@ snippet into your =dotspacemacs/user-config= function.
...
+the directory of the file that's being checked. This is normally not a problem,
+as all the paths are set properly, however it could be a problem if some
+template Haskell functions use relative paths (e.g. in Yesod scaffolded
+projects).
 
-Until it's fixed in flycheck the workaround is to the wrap the stack command to
-run all subcommands from the project root directory. You can do so with the
-following script:
+Until it's fixed in flycheck, the workaround is to wrap the =stack= command in
+order to run all subcommands from the projects root directory. You can do so
```
Miciah: Should be "project's" with an apostrophe.

layers/+lang/haskell/README.org
duianto: Would the single sentence read as:
However the version of the cabal library that's used by =stack= to generate the
directory name is not the version of the cabal library that's installed by
=stack=, but rather it's the version of cabal that's associated to the GHC
version.

With a comma after the second `=stack=`?

layers/+lang/haskell/README.org
duianto: It's probably better if you write the correct sentence, since I have a case of the "commas", and don't know which to delete or keep. 😄 

This is what I came up with:
This error can happen after upgrading cabal or cabal-install. To check
if this is the problem, compare the path name of the build path that's used by
flycheck using =flycheck-compile= and compare it to the actual path in the
=.stack-work= directory. If they are different then you'll need to reinstall ghc
using the command =stack setup --upgrade-cabal=.

Miciah: What you came up with looks good.  (I would keep the comma in "If they are different," but it's not necessarily needed.)

sdwolfz:
Thank you :heart:!

PR #9414, opened 2017-08-12
Add org-mode support for shell-scripts layer

NerdsvilleCEO:
This adds support for org-babel execution with the shell-scripts layer installed :)


NerdsvilleCEO:
Moved to https://github.com/syl20bnr/spacemacs/pull/9434

PR #9426, opened 2017-08-13
Remove obsolete purpose-conf name argument

duianto:
This message is shown when Spacemacs starts:
>../../Users/Username/AppData/Roaming/.emacs.d/layers/+spacemacs/spacemacs-purpose/local/spacemacs-purpose-popwin/spacemacs-purpose-popwin.el: Obsolete name arg "pupo" to constructor purpose-conf

The name argument, seems to only be needed for Emacs <= 24.3, according to this
line in the emacs-purpose package:
>;; the name arg ("purpose-x-code1") is necessary for Emacs 24.3 and older

source:
https://github.com/bmag/emacs-purpose/blob/00ddafcf4802e7430ca709769b888656a6eb421b/window-purpose-x.el#L71

The minimum version supported by Spacemacs is 24.4.

### DIFF
--- a/layers/+spacemacs/spacemacs-purpose/local/spacemacs-purpose-popwin/spacemacs-purpose-popwin.el
+++ b/layers/+spacemacs/spacemacs-purpose/local/spacemacs-purpose-popwin/spacemacs-purpose-popwin.el
@@ -185,8 +185,7 @@ Popwin's settings are taken from `popwin:special-display-config'."
                 (cond ((symbolp pattern) mode-purposes)
                       ((plist-get settings :regexp) regexp-purposes)
                       (t name-purposes))))
-    (purpose-conf "pupo"
-                  :mode-purposes mode-purposes
+    (purpose-conf :mode-purposes mode-purposes
                   :name-purposes name-purposes
                   :regexp-purposes regexp-purposes)))
 


JAremko:
@duianto   rebase your PR :wink:

syl20bnr:
Thank you ! 👍 

bmag:
We need to revert this change. The removed argument is mandatory in Emacs 24.5 and 24.4. Causes #9583

bmag:
I'll revert it later today/tomorrow

duianto:
Mybad, I misinterpreted the comment:
>;; the name arg ("purpose-x-code1") is necessary for Emacs 24.3 and older

to mean that the "name arg" only were necessary for 24.3 and older, but it probably means the 
`("purpose-x-code1")` string.

bmag:
I don't think you misinterpreted the comment, it's just that the comment is wrong. Sorry about that :-(

duianto:
Np, at least one problem was discovered by this PR 😄

bmag:
Reverted it now (see https://github.com/syl20bnr/spacemacs/commit/22deb279c35b494e27e2d88dc7d9a9cfd16f9415)

marienz:
Now that Spacemacs requires Emacs 25.1 (099bfcb0d926e9e476f14a15869e9bf97d4758eb), maybe the revert could be reverted?

It's not a big deal, but this is the only warning triggered by Spacemacs startup here. Would be nice to see it gone :)

PR #9428, opened 2017-08-14
Add the dotspacemacs-directory (~/.spacemacs.d) to the load-path

radious:
Users who choose to store they dotfile in ~/.spacemacs.d most likely store there also other files and want an ability to load them. Enabling it during startup makes the overall experience better.

### DIFF
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -35,6 +35,10 @@
 SPACEMACSDIR environment variable. If neither of these
 directories exist, this variable will be nil.")
 
+  ;; add dotspacemacs-directory to the load-path if present
+  (when dotspacemacs-directory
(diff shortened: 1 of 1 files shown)


beta1440:
From my experience, adding the `dotspacemacs-directory` to the load path raises a warning in Emacs since the `dotspacemacs-directory` contains an `init.el` file.

aaronjensen:
Yeah, I'm not a fan of this, I and others put lisp files in a separate directory or in personal layers. We should probably encourage that instead.

syl20bnr:
I agree with @aaronjensen, we can update the documentation about the dotdirectory to mention what this PR and make it an explicit action for the user to do. For instance add this load-path to `dotspacemacs-user-init` function.

radious:
Ok, that makes sense. Thanks for feedback!

PR #9430, opened 2017-08-14
Check if dotspacemacs-directory-snippets-dir exists before adding it

radious:
dotspacemacs-directory-snippets-dir contains a path to a directory with yas snippets in ~/.spacemacs.d (or arbitrary picked dotspacemacs direcotry). However, while it's not always present (user has to create it manually), it is always checked when yas is loaded what generates errors.

This diffs forces checking whether the assumed path exists before adding it to the yas-snippet-dirs.


syl20bnr:
Thank you ! 👍 
It was triggerring a `stringp` error for users not using the dotdirectory, I made a fix in commit https://github.com/syl20bnr/spacemacs/commit/b804aef6 please check everything is fine on your side.

PR #9444, opened 2017-08-16
Fix reference to cfgl-package-distant-p

coventry:
Ran into this when installing the OCaml layer. The function it's calling doesn't seem to be defined, but the commit it's from mentions what I've replaced it with, and fixed the failed layer install.


syl20bnr:
Nice catch !! 👍 
Thank you !

PR #9447, opened 2017-08-16
Add projectile-find-other-file keybinding

myrgy:
Useful shortcut from projectile to find other files. In c-c++ it's switch between .h/.cpp files


myrgy:
we have it in `SPC m g a/A`, since closed.

PR #9449, opened 2017-08-16
Clear both / and C-s search highlights with SPC s c

duianto:
When `evil-search-module` is set to:
evil-search:
Then `/` search highlights are cleared by pressing: `SPC s c`,
and `C-s` highlights are cleared with `:noh`.

isearch:
Then `/` highlights are cleared with `M-x evil-ex-nohighlight`,
and `C-s` highlights are cleared with both `SPC s c` and `:noh`.


syl20bnr:
Nice ! 👍 
Thank you !

Would be cool that have `:noh` to behave like `SPC s c` too :-)

duianto:
@syl20bnr fixed 😄
Clear both / and C-s search highlights with :noh #9497

PR #9454, opened 2017-08-17
BugFix: kill buffer shall return to *docker-containers*

FrancescElies:
Hi,

This PR fixes the bug described below, I'm not sure if this is the right place, I hope is fine.

All best & thanks
Francesc
 
----------------
#### Description :octocat:
After killing a log buffer from the docker layer I see the scratch buffer instead of the previous buffer.

- Start Emacs
- `SPC D c`: open docker containers buffer  =*docker-containers*= 
- `L`:       open log popup for a running container               
- `L`:       open the log                                         
- `SPC b d`: kill the buffer                                      

The `scratch` buffer is shown                        

I would like to see again the `*docker-containers*` buffer

### DIFF
--- a/layers/+tools/docker/packages.el
+++ b/layers/+tools/docker/packages.el
@@ -34,7 +34,8 @@
         "DP" 'docker-push
         "Dp" 'docker-pause
         "Dr" 'docker-restart
-        "Ds" 'docker-start)))
+        "Ds" 'docker-start)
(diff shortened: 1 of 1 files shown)


syl20bnr:
Thank you ! 👍 

@bmag @TheBB @d12frosted @JAremko We should think about refining the defaults for useful buffers. I sense we should consider almost all of them useful except for the obvious ones.

PR #9471, opened 2017-08-20
bm layer: fix readme.org

kccai:


JAremko:
Replace  `* Features` with `** Features` and it should be ok. Thx :wink:

JAremko:

PR #9474, opened 2017-08-20
spacemacs-base: Bind "SPC b l" to buffer-menu-other-window

henriquenj:
buffer-menu-other-window is a built-in function which opens a list of all buffers on the current session of Emacs. With the universal argument, it opens a list of buffers that are visiting a file.

It works much like "SPC g s": it opens the list on a new window, and can be dismissed with "q".

I've been using this binding for a couple of weeks with no issues. It seems the most natural place to put the command, and I was surprised that the keys were unused.

### DIFF
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -2092,6 +2092,8 @@ Buffer manipulation commands (start with ~b~):
 | ~SPC b C-D~     | kill buffers using a regular expression                                  |
 | ~SPC b e~       | erase the content of the buffer (ask for confirmation)                   |
 | ~SPC b h~       | open =*spacemacs*= home buffer                                           |
+| ~SPC b l~       | open the Buffer Menu with a list of all buffers                          |
+| ~SPC u SPC b l~ | open the Buffer Menu listing only the buffers that are visiting a file   |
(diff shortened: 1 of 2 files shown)


syl20bnr:
Thank you for the PR. I'm not sure we need this command though, in Spacemacs the native way to switch buffer is `SPC b b`. If you like `buffer-menu-other-window` you should take a look at the `ibuffer` layer.

henriquenj:
`SPC b b` and `buffer-menu-other-window` serve different purposes. `SPC b b` is indeed better suited for switching buffers, I use `buffer-menu-other-window` for taking an overall look at all the opened buffers, and allows me to kill a large amount of them all at once. I can also search their names/modes using ivy. So I think their functionality complete each other much like `dired` completes `SPC f f`.

That being said, I did take a look at `ibuffer` layer and it seems to do everything that `buffer-menu-other-window` does, but with more features. So now I'm confused if this binding is useful or not. Maybe we can use `SPC b l` as `buffer-menu-other-window` but replace it by `ibuffer` if they layer is installed? My goal is just to provide a spacemacs-style way to list all buffers.

henriquenj:
It seems there's no consensus about this binding, so I'll close this PR for now.

PR #9479, opened 2017-08-23
Install sed-mode with shell-scripts layer

eklitzke:
The `sed-mode` package provides a major mode for editing sed scripts, which is very useful for shell scripting.

### DIFF
--- a/layers/+lang/shell-scripts/packages.el
+++ b/layers/+lang/shell-scripts/packages.el
@@ -18,6 +18,7 @@
         ggtags
         helm-gtags
         insert-shebang
+        sed-mode
         (sh-script :location built-in)
         ))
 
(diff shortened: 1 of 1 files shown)


syl20bnr:
Thank you for the contribution, seems to be a useful major mode. Could you add how to use it in the README.org of the layer ? I looked at the documentation available of this mode in its source code and found nothing, I suppose it needs to be activated explicitly ? Anyway some docs would be welcome to help people to use this mode :-)

robbyoconnor:
Also -- you should rebase as you have conflicts.

eklitzke:
My motivation was just that if you have the `shell-scripts` layer enabled, editing a file named `foo.sed` should turn on sed syntax highlighting. There's no special functionality, it just makes `.sed` files be opened in the correct major mode. Sorry there is a conflict.

robbyoconnor:
LOL @eklitzke -- I meant merge conflict...

sdwolfz:
This looks like it could still be useful, @eklitzke do you want to revive this PR?

PR #9481, opened 2017-08-24
Conditionally enable ERC notifications via erc-enable-notifications

eklitzke:
Enable ERC notifications only when erc-enable-notifications is non-nil; the default configuration for the variable is t.


syl20bnr:
Thank you ! 👍 

PR #9485, opened 2017-08-24
[org] org-verbatim, not org-verbose

swaroopch:
I'm assuming this is a typo, because I see only verbatim in the orgmode
manual - http://orgmode.org/org.html#Emphasis-and-monospace


syl20bnr:
Thank you ! 👍 

PR #9495, opened 2017-08-27
Add unicode-fonts layer

aaronjensen:
Add support for the unicode-fonts package for better unicode font handling.

I'm not sure if this belongs in `+emacs` or `+intl`, let me know if you'd like me to move it.


syl20bnr:
Nice ! Thank you ! 👍 
I created a new category `+fonts` and moved the layer there, I imagine we may have other fonts related layers like ligatures.

Issue #8518, opened 2017-03-14
ex command :nohlsearch has no effect on evil-search-module highlights


liujoey:
#### Description :octocat:
evil-search-highlight-persist not working properly with evil-search-module

#### Reproduction guide :beetle:
- set editing style to hybird with these variables
```
   dotspacemacs-editing-style '(hybrid :variables
                                       hybrid-mode-enable-hjkl-bindings t
                                       hybrid-mode-use-evil-search-module t)
```
- Start Emacs
- Open any file and search with / command.
- remove highlight by using ex command `:noh`

*Observed behaviour:* :eyes: :broken_heart:
Highlights are not removed

*Expected behaviour:* :heart: :smile:
Highlights should be removed

#### System Info :computer:
- OS: darwin
- Emacs: 25.1.1
- Spacemacs: 0.200.7
- Spacemacs branch: develop (rev. e624e8f)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: helm
- Layers:
```elisp
(osx
 (auto-completion :variables auto-completion-enable-snippets-in-popup t auto-completion-enable-help-tooltip 'manual)
 better-defaults emacs-lisp imenu-list colors
 (mu4e :variables mu4e-enable-mode-line t mu4e-enable-notifications t)
 (version-control :variables version-control-diff-tool 'diff-hl)
 (shell :variables shell-default-shell 'ansi-term)
 markdown csv gtags yaml git html
 (go :variables go-use-gometalinter t)
 javascript typescript python django evil-commentary org syntax-checking restclient vinegar)
```
- System configuration features: JPEG RSVG IMAGEMAGICK NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS MODULES

liujoey:
I believe the issue is in this function:
```
(defun spacemacs-evil/init-evil-search-highlight-persist ()
  (use-package evil-search-highlight-persist
    :init
    (progn
      (global-evil-search-highlight-persist)
      ;; (set-face-attribute )
      (spacemacs/set-leader-keys "sc" 'spacemacs/evil-search-clear-highlight)
      (define-key evil-search-highlight-persist-map (kbd "C-x SPC") 'rectangle-mark-mode)
      (evil-ex-define-cmd "nohlsearch"
                          'evil-search-highlight-persist-remove-all)
      (spacemacs//adaptive-evil-highlight-persist-face)
[...]

PR #9497, opened 2017-08-28
Clear both / and C-s search highlights with :noh

duianto:
Clear both `/` and `C-s` search highlights with `:noh` and `:nohlsearch`,
and group the clear highlight commands next to each other.

---
This seems to resolve this issue:
ex command :nohlsearch has no effect on evil-search-module highlights #8518


syl20bnr:
Thank you ! 👍 

PR #9498, opened 2017-08-28
Add ivy support to the gtags layer

robbert-vdh:
I've added ivy support to the gtags layer through [counsel-gtags](https://github.com/syohex/emacs-counsel-gtags). It's missing a few minor features compared to `helm-gtags` but most of it is there.

There should probably be a single function in the future that calls all three functions (`spacemacs/counsel-gtags-define-keys-for-mode`, `spacemacs/ggtags-mode-enable` and `spacemacs/helm-gtags-define-keys-for-mode`) for us, but that's a job for another time.

### DIFF
--- a/layers/+lang/asm/packages.el
+++ b/layers/+lang/asm/packages.el
@@ -15,6 +15,7 @@
         asm-mode
         electric-indent-mode
         ggtags
+        counsel-gtags
         helm-gtags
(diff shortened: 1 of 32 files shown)


syl20bnr:
Thank you ! 💜 

Good idea for the simplification of `gtags` config 👍 

PR #9499, opened 2017-08-28
Add counsel-css as an ivy alternative to helm-css-scss

robbert-vdh:
I've added an ivy alternative to `helm-css-scss` to make it easier to navigate CSS though [counsel-css](https://github.com/hlissner/emacs-counsel-css). This also makes adds imenu support to CSS and Sass when using the ivy layer. Perhaps the `SPC m g h` keybinding could be changed to something more fitting, but I can't think of anything.


syl20bnr:
Thank you ! 👍 

PR #9503, opened 2017-08-28
Update Mac Install Instructions on README

ams340:
Brew linksapps is deprecated and emacs-plus does not work with
spotlight so suggesting to use emacs cask for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,19 +142,13 @@ commands in the terminal to install both Emacs and the default Source Code Pro
 font:
 
 ```sh
-brew tap d12frosted/emacs-plus
-brew install emacs-plus
-brew linkapps emacs-plus
+brew cask install emacs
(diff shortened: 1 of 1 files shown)


ams340:
Perhaps we should use emacs cask for now? 

screen shot 2017-08-25 at 3 11 53 pm

ranmacar:
I have tried both ways, and after cloning into .emacs.d and launching emacs nothing happens. Is there a way to trigger the installation manually, or see what is going wrong?

Thanks!

Mac os 10.13.2

duianto:
This PR was closed because, the current `develop` branch `macOS` > `Install emacs` documentation has a `Using cask` section with the same command as suggested here.
https://github.com/syl20bnr/spacemacs/blob/develop/README.md#install-emacs

PR #9507, opened 2017-08-28
Clojure mode: fix broken cider-inspect-prev-page binding

brianfay:
Hello!

I was trying to figure out what key to use to go backwards in the cider-inspector buffer (accessed with `, d v`)

I found that `N` was bound to `'cider-inspector-previous-page`, but that isn't an actual cider command so it was not working - the correct command is `'cider-inspector-prev-page`.

I bound this to both `N` and `p`; the `N` makes sense if you're using vim muscle memory, as if you're searching.

The `p` makes sense if you're using emacs muscle memory and you're thinking about using `C-n`, `C-p` to move to next and previous. Maybe the former is more in line with the spacemacs evil mode philosophy?


syl20bnr:
Thank you ! 👍 

PR #9509, opened 2017-08-29
Add docstring to align-repeat-x commands

troyp:
Update `spacemacs|create-align-repeat-x` macro to add a docstring to created
functions. The docstring refers the user to the macro, providing a link to
the source code.

Currently, calling `describe-function` on `spacemacs|align-repeat-` commands is not very helpful. It has no link to the definition, only to the top of the `funcs.el` file. Worse, if a user employs the macro in their init file to create custom align-repeat commands (I know I do), the help for those functions doesn't even identify the file where they were defined. There's no obvious way to fix this, but we can identify the defining macro itself, which is usually more important. Having the name of macro makes it easier to search for the definitions which call that macro, and the user can jump from the macro's documentation to its source code.

This commit adds the following docstring to each generated command.

    Created by `spacemacs|create-align-repeat-x'.

### DIFF
--- a/layers/+distributions/spacemacs-base/funcs.el
+++ b/layers/+distributions/spacemacs-base/funcs.el
@@ -882,11 +882,14 @@ the right."
                 nil))
 
 (defmacro spacemacs|create-align-repeat-x (name regexp &optional justify-right default-after)
-  (let ((new-func (intern (concat "spacemacs/align-repeat-" name))))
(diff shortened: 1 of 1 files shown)


duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/3c67ba94f38eec75a4c9aed72b752ef965164109

PR #9523, opened 2017-08-31
Format symbol highlight transient state

duianto:
Reduce the width of the listed key bindings by:
Moving the range and highlighted symbols counter to the right of the transient
state title.
Shorten some key description names.
Reduce three instances of "search" to just one.

And reorder the listed keys based on their search scope, small to large:
swoop, buffers, files, project.

---
before:
wider

after:
narrower


sdwolfz:
Thank you :heart:!

PR #9526, opened 2017-09-01
Rename ms in func ahs-ms-on-exit to ts

duianto:
`ms` is an abbreviation for a micro state, which was the previous name for a
transient state.


sdwolfz:
Thank you :heart:!

PR #9528, opened 2017-09-01
Rename ms in helm-nav face definition to ts

### LINKED ISSUE #8618: Open Helm Transient State: Invalid face reference... msg
#### Description
Invalid face reference messages appear in the messages buffer when the Helm Transient State is opened.

#### Steps to reproduce
1. Start Spacemacs with Helm completion
2. Open the messages buffer `SPC b m`
3. Open a helm buffer e.g. Helm M-x `SPC SPC`
4. Open the Helm Transient State `M-SPC`
(If latest message in the messages buffer has scrolled below the window, then the messages window can be scrolled up by pressing `C-M-v`, if it scrolled to far then scroll down again by pressing `S-C-M-v`)

#### Observed behaviour
The messages buffer shows: `Invalid face reference: minibuffer-prompt [7 times]`

The message seems to be repeating in the background, because if an action is performed that updates the messages buffer, e.g. by selecting the next/prev candidate with `j` or `k`, or by selecting another application so that Emacs loses focus, then one can see that the [* times] number increases by a larger amount depending on how long one waits.

#### Expected behaviour
The cause of the `Invalid face...` messages should be fixed.

#### Related: 
Sometimes the first message appears at the end of the previous message:
```
evil-line-move: End of bufferInvalid face reference: minibuffer-prompt
Invalid face reference: minibuffer-prompt [6 times]
```

or

```
QuitInvalid face reference: minibuffer-prompt
Invalid face reference: minibuffer-prompt [6 times]
```

While the Helm TS is open, if one presses `M-SPC`, then the messages buffer shows: 
`Invalid face reference: spacemacs-helm-navigation-ts-face [4 times]`
and the `Invalid face reference: minibuffer-prompt` message stops repeating.

#### Possible source of the messages:
This issue seems to be occurring in the `spacemacs//helm-navigation-ts-on-enter` function: https://github.com/syl20bnr/spacemacs/blob/2cbc1a6a6eeace22862921183db762244c1df4e2/layers/%2Bspacemacs/spacemacs-completion/funcs.el#L168

Because both of the `minibuffer-prompt` and `spacemacs-helm-navigation-ts-face` symbols are called by the `face-remap-add-relative` function.

#### Another function calling almost identical commands
I haven't and am not sure how to test this, but i suspect that the same `Invalid face reference: *` messages might be appearing when the `spacemacs//ido-navigation-ms-set-face` function is called:
https://github.com/syl20bnr/spacemacs/blob/2cbc1a6a6eeace22862921183db762244c1df4e2/layers/%2Bspacemacs/spacemacs-completion/funcs.el#L287

because the commands it calls are almost identical to the `spacemacs//helm-navigation-ts-on-enter` function.

#### System Info :computer:
- OS: windows-nt
- Emacs: 25.1.1
- Spacemacs: 0.200.7
- Spacemacs branch: develop (rev. 2cbc1a6a)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(autohotkey emacs-lisp git github helm version-control)
```

duianto:
The `spacemacs//helm-navigation-ts-set-face` function had been changed to call
the face `spacemacs-helm-navigation-ts-face` with `-ts-`, but the face
definition was still called `spacemacs-helm-navigation-ms-face` with `-ms-`.

---
Resolves #8618

### DIFF
--- a/layers/+spacemacs/spacemacs-completion/config.el
+++ b/layers/+spacemacs/spacemacs-completion/config.el
@@ -12,7 +12,7 @@
 
 ;; Helm
 
-(defface spacemacs-helm-navigation-ms-face
+(defface spacemacs-helm-navigation-ts-face
   `((t :background ,(face-attribute 'error :foreground)
        :foreground "black"))
   "Face for helm header when helm transient-state is activated."


sdwolfz:
Thank you :heart:!

PR #9529, opened 2017-09-02
Fixes search in helm for org files

agzam:
When you search with helm in Org files and the outline is collapsed
e.g.: `(setq org-startup-folded t)`, user may not see the results
when executing helm-persistent-action (`TAB` key)

This PR to fix that.

### DIFF
--- a/layers/+emacs/org/funcs.el
+++ b/layers/+emacs/org/funcs.el
@@ -44,3 +44,10 @@
 (defun spacemacs//evil-org-mode ()
   (evil-org-mode)
   (evil-normalize-keymaps))
+
+
+
+(defun spacemacs//org-reveal-on-helm-peristent-action ()
+  "reveal content for helm-persistent-action used in an Org file with folded outline"
+  (when (equalp major-mode 'org-mode)
+    (org-reveal)))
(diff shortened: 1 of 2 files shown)


bmag:
Thanks for your contribution, but IMO this PR is a no-go.

Revealing collapsed entries already happens for `helm-swoop` (`SPC s s`), `helm-grep-do-git-grep` and maybe others. You are probably trying to fix our "smart search" commands, such as `spacemacs/helm-project-smart-do-search` (`SPC /`) and `spacemacs/helm-files-smart-do-search` (`SPC s f`).

Those commands rely on `helm-do-ag` from [helm-ag](https://github.com/syohex/emacs-helm-ag) package, so this really is an upstream bug. The proper fix is to change upstream to use `helm-goto-char` and `helm-goto-line` instead of `goto-char` and `goto-line` (where it makes sense).

P.S. `major-mode` is a symbol, so check its equality with `eq`, not `equalp`

bmag:
See upstream issue and PR:
- https://github.com/syohex/emacs-helm-ag/issues/303
- https://github.com/syohex/emacs-helm-ag/pull/304

agzam:
@bmag no, I never expected this to be accepted right away, Spacemacs getting very popular (which is awesome) and other channels get very crowded. I knew it's an upstream issue, but I wasn't sure where to address it.

So I thought maybe I'll create a PR to start a conversation. Thank you for review and your elaborate answer.

bmag:
@agzam well, it worked 😄 Waiting for upstream to merge the fix now.

PR #9535, opened 2017-09-03
Fix rust layer call to `rustup`

Dietr1ch:
Bash compatibility was assumed, which made the shell command to fail on
fish and probably other shells too.

### DIFF
--- a/layers/+lang/rust/funcs.el
+++ b/layers/+lang/rust/funcs.el
@@ -27,7 +27,7 @@ using `cargo-process-run'."
   (let ((input-file-name (buffer-file-name))
         (output-file-name (concat temporary-file-directory (make-temp-name "rustbin"))))
     (compile
-     (format "rustc -o %s %s && %s"
+     (format "sh -c 'rustc -o %s %s  &&  %s'"
              (shell-quote-argument output-file-name)
              (shell-quote-argument input-file-name)
              (shell-quote-argument output-file-name)))))
--- a/layers/+lang/rust/funcs.el
+++ b/layers/+lang/rust/funcs.el
@@ -24,10 +24,24 @@ open a junk Rust file, type in some code and quickly run it.
 If you want to use third-party crates, create a a new project using `cargo-process-new' and run
 using `cargo-process-run'."
   (interactive)
-  (let ((input-file-name (buffer-file-name))
-        (output-file-name (concat temporary-file-directory (make-temp-name "rustbin"))))
+  (lexical-let ((input-file-name
+                  (buffer-file-name))
+                (output-file-name
+                  (concat
+                    temporary-file-directory
+                    (file-name-nondirectory (buffer-file-name))
+                    "-"
+                    (md5 (buffer-file-name)))))
+    (add-to-list 'compilation-finish-functions
+                 (lambda (buffer status)
+                   (if (string-match "finished" status)
+                     (shell-command
+                       (shell-quote-argument output-file-name)
+                       buffer)
+                     (message "Compilation failed with: %s" status))))
     (compile
-     (format "sh -c 'rustc -o %s %s  &&  %s'"
-             (shell-quote-argument output-file-name)
-             (shell-quote-argument input-file-name)
-             (shell-quote-argument output-file-name)))))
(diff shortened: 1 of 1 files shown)


Dietr1ch:
I don't know, but I'd bet for it.
I think there's no single commandline that runs " A && B" on arbitrary shells and systems.
Do you know how to do that `and` on elisp rather than relying on the shell command?

joehillen:
I think what you'd want is to check the result of:

```elisp
(compile
  (format "rustc -o %s %s"
          (shell-quote-argument output-file-name)
          (shell-quote-argument input-file-name)))
```

and run `(compile (shell-quote-argument output-file-name))` if it passes.

syl20bnr:
@Dietr1ch what about @joehillen comment ? Do you plan to integrate it ?

Dietr1ch:
I can, but I'm not sure how to test the output of that `compile`

I thought that it would be simply,
```
(and
 (compile
  (format "rustc -o %s %s"
          (shell-quote-argument output-file-name)
          (shell-quote-argument input-file-name)))
 (compile (shell-quote-argument output-file-name))
 )
```

I checked `compile` docs with `SPC h d f compile`, but they don't mention how to get the status code.

Dietr1ch:
It still needs some work, it's not catching compile errors :/

Dietr1ch:
There was a silly "\n" at the end of the status.., now it works

Miciah:
Just noticed this PR.  https://github.com/syl20bnr/spacemacs/pull/10610 is a different solution to the same problem (namely, the other PR runs the command using `sh -c`; simpler, not sure about portability of using `sh -c` though).  Which solution is preferred?

layers/+lang/rust/funcs.el
```diff
@@ -24,10 +24,28 @@ open a junk Rust file, type in some code and quickly run it.
...
+                    (md5 (buffer-file-name))))
+                )
+    (add-to-list 'compilation-finish-functions
+                 (lambda (buffer status)
+                   (if (string-match "finished" status)
+                     (shell-command
+                       (shell-quote-argument output-file-name)
+                       buffer
+                       )
+                     (message "Complation failed with: %s" status)
```
Miciah: Typo: s/Complation/Compilation/

Dietr1ch:
@Miciah, this started as the same `sh -c` hack, but windows compatibility was uncertain.
I consider moving to lisp is better because you can actually distinguish the build from the run.

I'll fix the hanging parentheses in the hopes that I get used to them at some point (maybe I just
need to learn how to really edit S-expressions on emacs rather than keeping my vi habits)

Dietr1ch:
The failure at `ci/circleci: core` is an infra failure,

> Starting container jare/spacemacs-circleci:emacs24
>   image cache not found on this host, downloading jare/spacemacs-circleci:emacs24
> Error response from daemon: manifest for jare/spacemacs-circleci:emacs24 not found

syl20bnr:
See comment: https://github.com/syl20bnr/spacemacs/pull/10610#issuecomment-397785479

Dietr1ch:
IIRC this issue was becaue using `&&` to chain commands doesn't work on fish (I think there's a recent hack/change to allow it).
Grabbing the right environment is nice as it ensures consistently with your shell, but it won't make the `&&` work.

The issue with using`sh -c "a && b"` was suspected incompatibility with windows, if it actually works it may be an ok solution whether we care or not about mixing the outputs or not.

syl20bnr:
What about using `shell-command-to-string` which should use the shell
pointed by `shell-file-name` ?

smile13241324:

duianto:
I haven't tested this updated `rust` function, but I noticed that it added a: `lexical-let`

The `cl` package was recently deprecated in Spacemacs, in favor of: `cl-lib`

It caused an issue with the `ivy` layer that used `lexical-let`.
It was fixed by this commit:
Remove uses of lexical-let, part of the deprecated cl package https://github.com/syl20bnr/spacemacs/commit/05fb56bfbfd43a38d0b4044796d2cff4cccac971

I'm just wondering if a similar issue might occur in the `rust` layer or at least with the usage of this function?

Issue #9476, opened 2017-08-22
Crowd fixing documentation [2]


JAremko:
We have [an average pile of documentation issues](https://gist.githubusercontent.com/JAremko/ec1a6d308eb1d26ceb77a95b315f463a/raw/c725cdcced491c23b1929635bc25c6ad38030c3f/links)  and we need to fix them **YESTERDAY**

So how about.... We all gab some? :shipit:  

Proper [README.org example](https://github.com/syl20bnr/spacemacs/blob/develop/core/templates/README.org.template)
- - - - - - -

## *Check other people's lists*

- - - - - -

# thx! :heart:

CeleritasCelery:
I will fix the convention of `Features` to `Features:`

duianto:
I started looking into fixing the bibtex layer link in the org layers readme, and discovered an issue with `%2B` in links.

I posted it as a separate issue: 
[org-mode link with %2B, becomes %252B #9484](https://github.com/syl20bnr/spacemacs/issues/9484)

CeleritasCelery:
One thing I noticed while fixing documentation was an inconsistency in the `Install` header. about half the layers have a sub header call `Layer` that includes installation instruction for the layer. The other half don't (which is how it is show in the template). I think it could be useful to standardize this.

In other words, you should only have the sub-header `Layer` if your `Install` header has multiple sub-headers. 
For example
```org
* Install
** Layer
** External Tool
``` 
would be okay.
But 
```org
* Install
** Layer
```
would not.

Just some food for thought @JAremko

JAremko:
I updated the links list.

JAremko:
updated list

benjamreynolds:
Grabbed a few more of these in #9525

smile13241324:
Fixed some in #9536, also noticed that the html generation for the plantuml layer seems to be broken, I have opened a separate issue for this [here](#9537)

duianto:
@smile13241324 The `here` link is pointing back to this issue.
`https://github.com/syl20bnr/spacemacs/issues/9476#9537`
the correct issue id `9537` became a fragment identifier `#` for this issue.

The correct url is:
`https://github.com/syl20bnr/spacemacs/issues/9537`
Generating plantuml layer documentation fails #9537

smile13241324:
@dubianto thanks fixed the link

smile13241324:
Fixed elisp, color, ibuffer and windows-scripts layer.

[...]

PR #9536, opened 2017-09-03
Add "features" block to doc of plantuml, shell and search-engine layer

smile13241324:
Added the missing "features" block to the plantuml, shell and search-engine layer documentation as requested in #9476.


JAremko:

PR #1486, opened 2015-05-04
Disable smartparens in SLIME REPL

tuhdo:
Some people prefer typing code in the SLIME REPL, and autoclosing parenthesis makes `RET` evaluate code instead of inserting newline. Disable this in SLIME REPL to remove the annoyance, though it's better to create a temp buffer, write code there then evaluate code written there.

PR #9540, opened 2017-09-05
Add configurable smartparens for common lisp layer

MatthewRock:
Currently, the Common Lisp layer disables smartparens by default. If user wishes to enable smartparens, he either has to add hook manually or to uncomment the proper line in the source code(and comment out the disabling part).

The commit I'm submitting adds a new variable, `common-lisp-enable-smartparens-mode`. The name is quite long, but I think that the only thing that I could drop is the "enable-" part.

The variable can be provided via :variables to enable smartparens mode. It stays turned off by default.

The usage guideline can be found in README.

Since the Install part of README now contains two separate instructions(for setting Lisp implementation and for setting smartparens mode), I've created a new headers for both.

### DIFF
--- a/layers/+lang/common-lisp/README.org
+++ b/layers/+lang/common-lisp/README.org
@@ -4,7 +4,10 @@
 
 * Table of Contents                                         :TOC_4_gh:noexport:
 - [[#description][Description]]
+  - [[#features][Features:]]
 - [[#install][Install]]
+  - [[#choosing-common-lisp-implementation][Choosing Common Lisp implementation]]
+  - [[#enabling-smartparens-strict-mode-or-smartparens-mode][Enabling smartparens-strict-mode or smartparens-mode]]
 - [[#key-bindings][Key Bindings]]
   - [[#working-with-lisp-files-barfage-slurpage--more][Working with lisp files (barfage, slurpage & more)]]
   - [[#leader][Leader]]
@@ -16,7 +19,11 @@
     - [[#macroexpansion][Macroexpansion]]
 
 * Description
-A Spacemacs layer providing Common Lisp support via [[https://github.com/slime/slime][SLIME]].
+** Features:
+   - Interactive REPL, debugger, inspector and more via [[https://github.com/slime/slime][SLIME]].
+   - Autocompletion via slime-company.
+   - Support for Common Lisp code snippets, user-made and pre-made, via common-lisp-snippets.
+   - Simple support for parenthesis organization via parinfer
 
 * Install
 If you have previously installed slime in any other way, it is recommended that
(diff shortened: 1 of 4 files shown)


MatthewRock:
I have added two commits(one was a failure) that fixed the "style" check of the readme by listing features.

joehillen:
@MatthewRock You should squash these commits.

MatthewRock:
@joehillen I have squashed the commits.

MatthewRock:
@joehillen just catching up, should I do something else before my work is merged, or are we waiting for the next release or something like that?

joehillen:
Unfortunately, I'm don't have commit access. If it were up to me, this and many other PR would be merged.

I have no idea why they being ignored. It's bad for the project.

MatthewRock:
This fixes #10001

CeleritasCelery:
While I see the reason for wanting more granular control, I don’t feel this is the right way to do it. If we took this approach to its logical conclusion we would be adding `smartparens-strict-mode` code to every language layer and it would just be a mess. What we should do is have a mechanism similar to `dotspacemacs-line-numbers` Where this can be controlled for all major modes from the `dotspacecs/init`.

MatthewRock:
@CeleritasCelery I agree. The main reason I did it this way was to incorporate the change I wanted(i.e. allow the user(me) to choose which smartparens option I want to use), but make it optional and leave the old default behavior(no smartparens). Plus, I thought it was a nice feature to be able to select the backend(since Lisp is parenthesis-heavy language, this feature is really important)

In the perspective it looks like a bad idea conceptually, though many other modes have smartparens by default so I didn't really notice. The main problem I had with this layer was that it forced no smartparens in SLIME and there wasn't easy way of changing it(hooks didn't work as they were triggered before turning smartparens off).

Plus, since the problem is localised, I don't know if the conceptual problem is that bad. If other modes don't need to control this, than maybe this small fix would do, though in a long run the global solution would be a better idea.

Are you suggesting that I should implement the solution like `dotspacemacs-line-numbers` instead? Also, how would the solution look like(the example user code)?

CeleritasCelery:
Few things here.
First I think we should look at why they are disabling smartparens in SLIME in the first place. If you look at the git blame we can see who added the code and (hopefully) why. If it was just a matter of personal preference then maybe we should just enable it in SLIME since that seems better. 

Also you could easily change this in your own config by adding `(remove-hook ‘slime-repl-mode-hook ‘slime/disable-smartparens)`.

As you mentioned, I am suggesting implement a solution like `dotspacemacs-line-numbers`. You can find the relavent code [here](https://github.com/syl20bnr/spacemacs/blob/bd7ef98e4c35fd87538dd2a81356cc83f5fd02f3/layers/%2Bdistributions/spacemacs-base/funcs.el#L1082)

OlegTheCat:
@CeleritasCelery @MatthewRock guys, any update on this issue? Having `smartparens` disabled by default and not being able to enable them with `(add-hook 'slime-repl-mode-hook 'smartparens-mode)` is definitely not great UX.

> If we took this approach to its logical conclusion we would be adding smartparens-strict-mode code to every language layer and it would just be a mess.

It seems that is already [being done](https://github.com/syl20bnr/spacemacs/search?utf8=%E2%9C%93&q=post-init-smartparens&type=) anyway :(

The only thing that I'd add to @MatthewRock solution is handling of `dotspacemacs-smartparens-strict-mode` variable, instead of using `common-lisp-enable-smartparens-mode` value

OlegTheCat:
>  First I think we should look at why they are disabling smartparens in SLIME in the first place. If you look at the git blame we can see who added the code and (hopefully) why. If it was just a matter of personal preference then maybe we should just enable it in SLIME since that seems better.

 @CeleritasCelery here's that PR #1486

MatthewRock:
@CeleritasCelery as you can see from the PR provided by @OlegTheCat , it seems like someone added it as a matter of personal taste.

I agree that enabling it by default would make it better, but I wouldn't want to make it hard for other users (I think this should be configurable by some variable).

Relying on dotspacemacs-smartparens-strict-mode is okay and I could implement that, but the problem is that this way CL files and REPL ale coupled; you either have the smartparens mode turned on or off. You can't have it on in files and off in the REPL. To get around that, you'd have to use similar methods to the one I've used.

As I expect smartparens mode to work in REPL, I don't care that much if other users would have to modify this, but at the same time I would prefer to provide as smooth transition as I can.

Also thanks @OlegTheCat for participation, especially finding the PR.

PR #9543, opened 2017-09-05
ivy: add ivy-rich display transformer for ivy-switch-buffer

a-j-f:
This change adds the ivy-rich package to the ivy layer and configures it by default (https://github.com/Yevgnen/ivy-rich)

As you can see from the package github page this adds very useful contextual information to ivy-switch-buffer.

### DIFF
--- a/layers/+completion/ivy/packages.el
+++ b/layers/+completion/ivy/packages.el
@@ -20,6 +20,7 @@
         helm-make
         ivy
         ivy-hydra
+        ivy-rich
         (ivy-spacemacs-help :location local)
         persp-mode
         projectile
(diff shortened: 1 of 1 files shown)


robbert-vdh:
There seem to be some problems when using layouts (perspectives). For example, here I've got a new layout with a single file and some open dired buffers. Only the file's buffer shows up, things like scratch and recent buffers are missing. These will only show after typing a pattern that matches none of the currently open file buffers. Everything seems to work correctly on the default layout though.

syl20bnr:
Thank you ! 💜 

Nice package, ivy users will love that.

zzjjzzgggg:
Can I opt to disable this feature? It is totally a mess in my emacs. 

See
default
Thanks!

PR #9547, opened 2017-09-06
Fix helm-bookmark initialization

cdlm:
The helm layer depended on `bookmark` directly instead of through `helm-bookmark`, and initialization was in the wrong place, resulting in *symbol's value as varable is void* errors about `helm-bookmark-map`.

### DIFF
--- a/layers/+completion/helm/packages.el
+++ b/layers/+completion/helm/packages.el
@@ -15,7 +15,6 @@
         auto-highlight-symbol
         helm
         helm-ag
-        helm-bookmark
         helm-descbinds
         helm-flx
         helm-make
--- a/layers/+completion/helm/packages.el
+++ b/layers/+completion/helm/packages.el
@@ -13,9 +13,9 @@
       '(
         ace-jump-helm-line
         auto-highlight-symbol
-        bookmark
         helm
         helm-ag
+        helm-bookmark
         helm-descbinds
         helm-flx
         helm-make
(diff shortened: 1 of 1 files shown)


cdlm:
This code looks dubious though… it doesn't seem like `helm-bookmark` has a binding, and the `C-e` `C-d` etc bindings do not work in the `helm-bookmark` buffer either… I'm confused as to how this was intended to work :|

duianto:
Describing a variable `C-h v` or function `C-h f`, and saving with `SPC f s`, works again after applying these changes.

But it seems to create issues with the `SPC f b` helm filtered bookmarks buffer.
Trying to create a new bookmark shows:
>bookmark-get-bookmark: Invalid bookmark test

in the messages buffer.

And `C-d` to delete and `C-e` to edit a bookmark, doesn't work either.

bmag:
This PR fixes the void-variable error, but at the cost of not running the post-init code at all. Changed in https://github.com/syl20bnr/spacemacs/commit/f221d7ed7308d029bb302757161ec51853c461b2 to use `with-eval-after-load` instead of a hook.

AFAICT the post-init code is for `bookmark` package instead of `helm-bookmark` so it depends on `bookmark` init from `spacemacs-base` layer, but if anyone thinks current situation isn't good you're welcome to open a new PR for that.

Anyway, the linked commit should fix the issue, so I'm closing the PR, thanks.

GaryHeath:
How do I, as a non-technical Spacemacs user, import this fix? After doing an "Update Packages" today I'm now unable to load a file into a cider repl (C-c C-l) due to “Symbol’s value as variable is void: helm-bookmark-map”.

cdlm:
I think you have to either wait for the next release, or switch to the develop branch.

duianto:
Adding `(require 'helm-bookmark)` to the user-config section of the `.spacemacs` file, is probably the simplest current solution on the master branch. As stated by bmag here: https://github.com/syl20bnr/spacemacs/issues/9549#issuecomment-327788403

PR #9555, opened 2017-09-07
faust layer: use faustine instead of faust-mode

magnetophon:
faustine depends on faust-mode, and extends it.


magnetophon:
I moved to doom emacs.

PR #9568, opened 2017-09-10
Restore evil-org-additional-bindings in insert mode

smile13241324:
Before some time the additional evil-org bindings were available in all three evil modes. However due to a bug in evil-org this seems to have been changed. The bug is fixed now and I have restored the old insert mode bindings by setting evil-org-use-additional-insert to true during package initialization.

### DIFF
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -55,6 +55,7 @@
     :init
     (progn
       (add-hook 'org-mode-hook 'spacemacs//evil-org-mode)
+      (setq evil-org-use-additional-insert t)
       (setq evil-org-key-theme `(textobjects
                                  navigation
                                  additional


smile13241324:
@d12frosted Sorry to disturb you but I found another one which is now waiting for 3 month to be merged. Is there anything I could do to speed things up a bit? 

This is a fix for the evil-org integration. There was a bug in evil which broke the insert mode bindings for meta left and meta right which where consequently disabled in the evil-org integration.

I made a commit at evil which was merged 3 month ago that fixed this issue. Then I made this PR to reactivate the insert bindings in the evil-org integration. Currently I am working with bindings I do in my dotfile but I think that many more users are missing those insert bindings still and probably end up with a local fix in their dotfile.

syl20bnr:
Thank you !

PR #9574, opened 2017-09-12
Added known-issues section to spell-checking layer

adamkruszewski:
Vim-empty-lines casues problems when using spell-checking inside org-mode buffer.
Just added information for others as the stack traces given doesn't indicate directly that vim-empty-lines messes with it (attached a screenshot with the stacktrace just to the record)

spell-check-flycheck-spacemacs-error

### DIFF
--- a/layers/+checkers/spell-checking/README.org
+++ b/layers/+checkers/spell-checking/README.org
@@ -2,16 +2,25 @@
 
 * Table of Contents                                         :TOC_4_gh:noexport:
 - [[#description][Description]]
+  - [[#features][Features:]]
 - [[#install][Install]]
   - [[#layer][Layer]]
   - [[#spell-checker-configuration][Spell Checker Configuration]]
   - [[#disabling-by-default][Disabling by default]]
   - [[#enabling-auto-dictionary-mode][Enabling auto-dictionary-mode]]
+  - [[#enabling-multi-dictionary-support-with-hunspell][Enabling multi-dictionary support with hunspell]]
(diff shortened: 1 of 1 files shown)


adamkruszewski:
Can't really figure out how to fix the optional test that fails. The message from build server reads:
```
!!!!!!!!!!!!!! TEST FAILED !!!!!!!!!!!!!!
File "/root/.emacs.d/layers/+checkers/spell-checking/README.org" doesn't have "Features:"(With a colon) list in the top level "Description" headline
See "https://github.com/syl20bnr/spacemacs/blob/develop/core/templates/README.org.template"
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```
It seems my modifications didn't affected the results but the file itself doesn't adhere to set standard. Maybe the standard changed after the file was created and that's the reason why it doesn't meet it?

duianto:
It just means that `readme.org` is missing a `- [[#features][Features:]]` entry in the TOC at the top and a
`** Features:` section in the `* Description` section below. Here's an example of a recent PR, that added three such sections:
https://github.com/syl20bnr/spacemacs/pull/9536/files

adamkruszewski:
It seems I was able to fix all "formatting" issues with the file along my changes and I have added short instructions how to enable multi-dictionary support for hunspell as well (as it might come in handy for someone).
Hopefully I have done everything alright this time :)

syl20bnr:
Thank you ! 👍 

PR #9575, opened 2017-09-12
Mitigation of enriched text remote code execution

reactormonk:
See http://seclists.org/oss-sec/2017/q3/422

... yes, it's against master.

### DIFF
--- a/init.el
+++ b/init.el
@@ -9,6 +9,12 @@
 ;;
 ;;; License: GPLv3
 
+;; Mitigation of GNU Emacs 25.2 enriched text remote code execution
+;; see http://seclists.org/oss-sec/2017/q3/422
+(eval-after-load "enriched"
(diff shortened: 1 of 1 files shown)


init.el
```diff
@@ -9,6 +9,12 @@
 ;;
 ;;; License: GPLv3
 
+;; Mitigation of GNU Emacs 25.2 enriched text remote code execution
+;; see http://seclists.org/oss-sec/2017/q3/422
+(eval-after-load "enriched"
```
benjamreynolds: Should this be wrapped in a conditional to check the current emacs version? [25.3](https://lists.gnu.org/archive/html/emacs-devel/2017-09/msg00211.html) was released recently and patched this vulnerability, so this patch should only be necessary for emacs versions lower than that.

PR #9586, opened 2017-09-14
The mu4e maildirs extension is now optional

JorisE:
I was wondering why #4892 never led to anything, but it seems that I never actually created a PR for this.


JorisE:
Formatting failed on a stray whiteline that I removed in the commit.

syl20bnr:
Thank you ! 👍 

This time it is in :-)

Issue #903, opened 2015-03-21
Add new use-package keywords `:evil` and `:holy`


syl20bnr:
use-package will support keyword extension in a near future. We will be able to add new keywords:
- `:evil-init` and `:evil-config`
- `:holy-init` and `:holy-config`

It will make the code easier to understand and navigate.

StreakyCobra:
@syl20bnr Any news on this?

TheBB:
Use-package already supports this so it's just up to us now.

CarlQLange:
I added a PR with these changes: #9587. I have no idea if this is still desired or useful, but hey! Maybe it is!

syl20bnr:
This request is out-date, @CarlQLange I'm sorry it stayed open for so long.

PR #9587, opened 2017-09-14
Add :evil-init, :holy-init, :evil-config and :holy-config to use-package keywords

CarlQLange:
This PR would close https://github.com/syl20bnr/spacemacs/issues/903.

Please check it out - I'm not very sure of the interface with use-package (do you need to place your keywords in a specific place in `use-package-keywords`?). The code also isn't very DRY or lispy - please forgive me and let me know what the best way to do this would be. I'm particularly unsure of the location of my changes.

Hopefully it'll simplify package inits/configs for some packages!

Basically, this will let you do:

```elisp
(defun helpful-layer/init-helpful()
    (use-package helpful
      :evil-init (evil-define-key whatever) ;more evil specific init here
      :holy-config (holy-something ;holy specific config
)))
```

or `:evil-config` or `:holy-config`.

I suppose this could become a bigger PR, changing layers and other `use-package` uses to use this paradigm, but I wanted to get feedback first! I personally don't even know if this is particularly useful, since the original issue hasn't been touched in a Long Time. Let me know! ♥️

### DIFF
--- a/layers/+distributions/spacemacs-bootstrap/packages.el
+++ b/layers/+distributions/spacemacs-bootstrap/packages.el
@@ -304,7 +304,58 @@
   (setq use-package-verbose init-file-debug
         ;; inject use-package hooks for easy customization of stock package
         ;; configuration
-        use-package-inject-hooks t))
+        use-package-inject-hooks t)
+
+
+  (add-to-list 'use-package-keywords :evil-init t)
+  (defalias 'use-package-normalize/:evil-init 'use-package-normalize-forms)
+  (defun use-package-handler/:evil-init (name-symbol keyword arg rest state)
+    (when (bound-and-true-p evil-mode)
+      (let ((body (use-package-process-keywords name rest state)))
+        (use-package-concat
+         ;; The user's initializations for evil-mode
+         (let ((init-body
+                (use-package-hook-injector (use-package-as-string name)
+                                           :init arg)))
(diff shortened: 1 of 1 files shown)


syl20bnr:
Thank you for the PR, I really appreciate you took the time to address this issue. I'm sorry but this issue should have been closed as it is out-dated. Spacemacs is now always using `evil-mode` behind the scene even in `holy-mode` as we use the abstraction of states of evil to define `holy-mode` (emacs state).

We started at some point to make specific configuration for evil and holy modes and it turned out to be a recipe for maintenance nightmare so I decided to keep `evil` always enabled and see the holy mode like another state (emacs-state) which replaces all the other evil states. It drastically simplified how we approach the configuration as we can just think about evil users while still having everything working for holy users.

Moreover it is important to be able to switch dynamically the editing style, it is a feature of Spacemacs I consider essential, so Vim people and Emacs people can do pair-programming using the same editor and switch the style if needed with `SPC t E` and everything stays consistent (evilified buffers, hjkl navigation etc...). We have a hook to be able to do this  `spacemacs-editing-style-hook` which is simpler that a `use-package`  keyword.

I'm closing this PR since I see no use of it in the current context.

Thank you again for the PR 💜

CarlQLange:
No problemo, @syl20bnr! Thanks for the explanation!

PR #9601, opened 2017-09-18
add bindings to navigate between errors in tex mode

olivierverdier:
Add two bindings to navigate back and forth between errors in tex mode.

Ideally, the binding for the next error should have been something like `SPC m e n`, but `SPC m e` is already taken for environment commands.

### DIFF
--- a/layers/+lang/latex/README.org
+++ b/layers/+lang/latex/README.org
@@ -128,6 +128,8 @@ By default, the underlying latex code is echoed in the echo area.
 | ~SPC m k~     | kill TeX job                               |
 | ~SPC m l~     | recenter output buffer                     |
 | ~SPC m m~     | insert LaTeX macro                         |
+| ~SPC m n~     | goto next error                            |
+| ~SPC m N~     | goto previous error                        |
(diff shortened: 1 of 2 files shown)


syl20bnr:
I see that `flycheck` is configure for the `LaTeX` so we may have `SPC e n` and `SPC e N` that work.

olivierverdier:
Yes, but strangely enough, flycheck does not find errors, it gives only style warnings (using tex-chk). I agree that making flycheck work with LaTeX would be the best solution.

olivierverdier:
Any news about this?

duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/7818fa3ccf8579c8192e471b61678c6c57db8e1f

The key binding comment: `;; M-m g` seems to have been a typo, because `M-m` is the 
`dotspacemacs-emacs-leader-key`, and `g` is the `git/version-control` prefix.

The comment was changed to: `;; M-g p` because it's bound to `TeX-previous-error` in a `.tex` file.

Issue #8134, opened 2017-01-08
how to make configuration-layer/package-usedp 'helm return true?


ghost:
#### Description :octocat:
I want to use package ```helm-gtags``` and ```helm-cscope``` without ```helm layer``` (I use ivy).

I add ```package helm``` into  ```dotspacemacs-additional-packages```, but package ```helm-gtags``` and ```helm-cscope``` still cann't be installed.

#### Reproduction guide :beetle:
- backup .emacs.d and .spacemacs.d
- git clone https://github.com/syl20bnr/spacemacs.git ~/.emacs.d -b develop
- add ```package helm``` into  ```dotspacemacs-additional-packages```
- Start Emacs
- cd .emacs.d/elpa 

*Observed behaviour:* :eyes: :broken_heart:
helm-gtags and helm-cscope not exist.

*Expected behaviour:* :heart: :smile:
helm-gtags and helm-cscope can be installed.

#### System Info :computer:
- OS: darwin
- Emacs: 25.1.1
- Spacemacs: 0.200.7
- Spacemacs branch: develop (rev. 3bbeae)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: ivy
- Layers: ```ivy emacs-lisp```

syl20bnr:
Seems like a bug to me, need to investigate.

chasinglogic:
I've submitted a PR with a fix #9617

hgijeon:
Is this resolved?

PR #9617, opened 2017-09-19
Issue #8134 Correctly return non-nil for dotspacemacs-additional-packages

chasinglogic:
`configuration-layer/package-used-p` would not return non-nil for any packages
added via `dotspacemacs-additional-packages`. It appears that the :requires slot
is unbound on such packages so I changed the call to oref-default from oref so
that the proper initform (`nil`) is returned instead.

I did try using some form of `(when (slot-boundp obj 'requires))` to only
perform the check when the slot is bound but for some reason slot-boundp threw
the same error that oref did. I will be investigating this further and
potentially submitting a bug upstream but `oref-default` seems like a fine
solution for this problem.

EDIT:
Putting issue #8134 in the description so it links I thought the subject line would.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -1447,7 +1447,7 @@ RNAME is the name symbol of another existing layer."
          (not (oref obj :excluded))
          (not (memq nil (mapcar
                          'configuration-layer/package-used-p
-                         (oref obj :requires)))))))
+                         (oref-default obj :requires)))))))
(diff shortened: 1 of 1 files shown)


tonylotts:
If you rebase onto develop that tests might pass.

chasinglogic:
Rebased onto current develop

chasinglogic:
The circleCI tests are still failing but I never get the results to load in the Web UI is it a permissions thing? I'm not familiar with CircleCI

PR #9624, opened 2017-09-21
prodigy: bind 'q' to quit-window in view mode

razzmatazz:
Update prodigy keybindings so 'q' quits the window in `prodigy-view-mode` so it matches the behaviour of 'q' keypress in parent `prodigy-mode` window. So pressing 'q' + 'q' from prodigy view mode closes both service log and parent prodigy service list windows.


syl20bnr:
Thank you ! 👍 

PR #9643, opened 2017-09-24
Improved emacs lisp layer

AdrieanKhisbe:
This PR improves the emacs-layer by:

- identifying `Cask` files as emacs-lisp file
- addign the overseer package, and bindings it's commands to run test to the <kbd>,t</kbd> prefix

### DIFF
--- a/layers/+lang/emacs-lisp/README.org
+++ b/layers/+lang/emacs-lisp/README.org
@@ -9,7 +9,9 @@
 - [[#auto-compile][Auto-compile]]
 - [[#working-with-lisp-files-barfage-slurpage--more][Working with lisp files (barfage, slurpage & more)]]
 - [[#debugging-elisp][Debugging Elisp]]
+- [[#nameless][Nameless]]
 - [[#key-bindings][Key bindings]]
+  - [[#additional-testing-functions-with-overseer][Additional testing functions with overseer]]
   - [[#additional-evaluation-functions][Additional evaluation functions]]
   - [[#format-code][Format code]]
   - [[#debugging][Debugging]]
@@ -25,6 +27,8 @@ always be in your dotfile, it is not recommended to uninstall it.
 - Support for specific lisp navigation styles via =emacs-lisp-mode=
 - Auto-compile via [[https://github.com/tarsius/auto-compile][auto-compile]] package
 - Debuggin via [[https://www.gnu.org/software/emacs/manual/html_node/elisp/Edebug.html#Edebug][edebug]]
+- Ert test runner with [[https://github.com/tonini/overseer.el][overseer]]
+- Nameless package prefix with [[https://github.com/Malabarba/Nameless][nameless]]
 
 * Install
 To use this configuration layer, add it to your =~/.spacemacs=. You will need to
@@ -93,6 +97,13 @@ function or press ~o~ to go out of it.
 
 7) Press ~a~ to stop debugging.
 
+* Nameless
+Nameless hides package namespaces in your emacs-lisp code, adn replace it by leading ~:~
+It can be toggled by ~SPC m~.
+
+To have it automatically on, you need to add ~(setq nameless-auto-mode t)~ in ~dotspacemacs/user-init~.
+Further configuration can be done with variable ~nameless-global-aliases~, cf [[https://github.com/Malabarba/Nameless#requiring-other-packages-as-aliases][original library documentation]]
+
 * Key bindings
 
 | Key Binding                | Description                                            |
@@ -112,6 +123,22 @@ function or press ~o~ to go out of it.
 | ~SPC m t b~                | run tests of current buffer                            |
 | ~SPC m t q~                | run =ert=                                              |
 | ~SPC m d m~                | open [[https://github.com/joddie/macrostep][macrostep]] transient-state                         |
+| ~SPC m :~                  | toggle nameless minor mode                             |
+
+** Additional testing functions with overseer
+Function related to test are present under the ~SPC m t~ prefix:
+
+| Key Binding | Description   |
+|-------------+---------------|
+| ~SPC m t a~ | overseer test |
+| ~SPC m t A~ | test debug    |
+| ~SPC m t t~ | run test      |
+| ~SPC m t b~ | test buffer   |
+| ~SPC m t f~ | test file     |
+| ~SPC m t g~ | test tags     |
+| ~SPC m t p~ | test prompt   |
+| ~SPC m t q~ | test quiet    |
+| ~SPC m t h~ | test help     |
 
 ** Additional evaluation functions
 If =smartparens= is used the following additional key bindings are available:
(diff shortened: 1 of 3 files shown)


layers/+lang/emacs-lisp/packages.el
```diff
@@ -175,6 +177,21 @@
       (spacemacs/set-leader-keys-for-major-mode 'emacs-lisp-mode
         "dm" 'spacemacs/macrostep-transient-state/body))))
 
+(defun emacs-lisp/init-overseer ()
+  (use-package overseer
+    :defer t
+    :init (spacemacs/set-leader-keys-for-major-mode 'emacs-lisp-mode
+            "oa" 'overseer-test
+            "ot" 'overseer-test-run-test
+            "ob" 'overseer-test-this-buffer
```
tko: I believe this should be `SPC m t b` -- similarly probably for (some) other keys as well. See more examples in [CONVENTIONS](https://github.com/syl20bnr/spacemacs/blob/master/doc/CONVENTIONS.org#tests)

AdrieanKhisbe: @tko do you mean as extra bindings?
I can add those for sure for the one that matches

dcluna: @AdrieanKhisbe `,o` is a prefix [reserved for users](https://github.com/syl20bnr/spacemacs/blob/master/doc/CONVENTIONS.org#user-prefix). Please follow @tko 's suggestion.

AdrieanKhisbe: My bad, i didn't know back then. Just read about it after

AdrieanKhisbe: And here it is. just ammened my commit

CeleritasCelery:
I wonder if it’s wise to have nameless mode enabled by default. *most* people will not expect that behavior. Also you should document these new features in the readme

nixmaniack:
I would prefer not enabling nameless by default.

AdrieanKhisbe:
@nixmaniack @CeleritasCelery 
Sure, I'll disable it. and document it of course

AdrieanKhisbe:
- Rebased
- non automatic trigger of nameless mode, behind a init switch `nameless-auto-mode`
- documentation of nameless and overseer

cc @nixmaniack @CeleritasCelery

layers/+lang/emacs-lisp/README.org
```diff
@@ -93,6 +97,13 @@ function or press ~o~ to go out of it.
 
 7) Press ~a~ to stop debugging.
 
+* Nameless
+Nameless hides package namespaces in your emacs-lisp code, adn replace it by leading ~:~
```
CeleritasCelery: adn -> and

CeleritasCelery: adn replace -> and replaces

layers/+lang/emacs-lisp/README.org
```diff
@@ -93,6 +97,13 @@ function or press ~o~ to go out of it.
...
+It can be toggled by ~SPC m~.
```
CeleritasCelery: `~SPC m~` -> `~SPC m :~`

layers/+lang/emacs-lisp/packages.el
```diff
@@ -175,6 +178,33 @@
       (spacemacs/set-leader-keys-for-major-mode 'emacs-lisp-mode
         "dm" 'spacemacs/macrostep-transient-state/body))))
 
+(defun emacs-lisp/init-nameless ()
+  (use-package nameless
+    :defer t
+    :init (progn
+            (when (bound-and-true-p nameless-auto-mode)
```
CeleritasCelery: The standard "Spacemacs" way to do this would be to add a variable to `config.el` which would be called `emacs-lisp-nameless-mode` (or something similar). That variable would be set to nil and a user could enable it with     
```elisp
'((emacs-lisp :variables emacs-lisp-nameless-mode t))
```
you could then check the value of it in `:init` and add the hook if necessary. The documentation would need to be updated to reflect this.

layers/+lang/emacs-lisp/README.org
```diff
@@ -93,6 +97,13 @@ function or press ~o~ to go out of it.
...
+
+To have it automatically on, you need to add ~(setq nameless-auto-mode t)~ in ~dotspacemacs/user-init~.
```
CeleritasCelery: see comment below about adding layer variables.

layers/+lang/emacs-lisp/packages.el
```diff
@@ -175,6 +178,33 @@
       (spacemacs/set-leader-keys-for-major-mode 'emacs-lisp-mode
         "dm" 'spacemacs/macrostep-transient-state/body))))
 
+(defun emacs-lisp/init-nameless ()
+  (use-package nameless
+    :defer t
+    :init (progn
```
CeleritasCelery: put the `progn` on its own line.

layers/+lang/emacs-lisp/README.org
AdrieanKhisbe: oups, the typo

AdrieanKhisbe:
Thanks for your feedback @CeleritasCelery, I'll apply your recommendation as soon as I can

AdrieanKhisbe:
which happened sooner than first expected

layers/+lang/emacs-lisp/README.org
```diff
@@ -93,6 +97,15 @@ function or press ~o~ to go out of it.
 
 7) Press ~a~ to stop debugging.
 
+* Nameless
+Nameless hides package namespaces in your emacs-lisp code, and replaces it by leading ~:~
+It can be toggled by ~SPC m :~.
+
+To have it automatically on, you need to define a layer variable:
+~(emacs-lisp :variables emacs-lisp-nameless-mode t)~ in ~.spacemacs~ ~dotspacemacs/layers~
```
nixmaniack: Spacemacs Document convention: `~` is used for <kbd>keybindings</kbd> and `=` is used for normal `code`.

AdrieanKhisbe: my bad

AdrieanKhisbe: @nixmaniack solved

syl20bnr:
Thank you ! 💜 

I made some opinionated modifications as follow:
- Change nameless prefix to > instead of default :
- Change toggle to SPC m >
- Set nameless separator to nil in order to make it work with any type or separator
- Add diminish unicode ⧁
- Define some default global aliases for Spacemacs source code
- Rename layer variable emacs-lisp-nameless-mode to emacs-lisp-hide-namespace-prefix
- Make variable nameless-current-name safe as a local variable for string values
- Set nameless prefix for core-configuration-layer.el

Commit: https://github.com/syl20bnr/spacemacs/commit/2ed3946d
 

PR #9647, opened 2017-09-25
Org babel keys

timor:
Added org-babel-* commands under sub-keymap "m v", along with a transient state that allows for easy source block navigation

### DIFF
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -171,6 +171,7 @@ Will work on both org-mode and any mode that accepts plain html."
                         ("mti" . "insert")
                         ("mtt" . "toggle")
                         ("mx" . "text")
+                        ("mb" . "src-blocks/babel")
                         ))
         (spacemacs/declare-prefix-for-mode 'org-mode (car prefix) (cdr prefix)))
       (spacemacs/set-leader-keys-for-major-mode 'org-mode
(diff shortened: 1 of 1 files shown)


timor:
I created the transient state mainly for ease of navigating src blocks, if there are any other repetitive actions for these, they can easily be added.

dvzubarev:
Why don't you use `m b` as a prefix? I think it is more mnemonic than `m v`.

Somelauw:
Probably based on the C-c C-v prefix used by Emacs org-mode.

joehillen:
A transient state for org-babel!? Yes please! 💯

timor:
Yes, I chose `m v`  because of the C-C C-v bindings. `m b` is also a good choice, how to decide? I suspect there are may more org bindings that have not been made available yet?

timor:
also, `m b` is already taken for `org-tree-to-indirect-buffer`

syl20bnr:
+1 for `SPC m b` prefix which fits nicely with `block` and `babel`.
`SPC m b` is now free, it may have been used before but after the reorganization of key bindings `org-tree-to-indirect-buffer` is now on `SPC m s b`.

timor:
Ok `m b` it is, then.  Which section of the Readme should be updated?

syl20bnr:
Thank you ! 💜 

syl20bnr:
@timor would be wonderful to add the key bindings in the README.org: https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Bemacs/org/README.org 👼

PR #9648, opened 2017-09-25
Enable purescript-decl-scan-mode when loading purescript-mode

bjarkevad:
This should fix `SPC j i` for PureScript

### DIFF
--- a/layers/+lang/purescript/packages.el
+++ b/layers/+lang/purescript/packages.el
@@ -33,6 +33,7 @@
     :init
     (progn
       (add-hook 'purescript-mode-hook 'turn-on-purescript-indentation)
+      (add-hook 'purescript-mode-hook 'purescript-decl-scan-mode)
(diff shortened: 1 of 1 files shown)


kRITZCREEK:
I'd prefer to not do it this way. With `psc-ide` we actually have a less "stringly" way of implementing this, by querying the ide server. Would you mind listing the feature set you care about in an issue on the psc-ide-emacs repo and maybe we can look at implementing the feature there?

bjarkevad:
I'm very interested in implementing this is a more robust way, as decl-scan does seem a bit hackish to me. 
The feature I'm interested in is primarily `SPC j i` for queries like this:

image

Another feature that may be valuable is `imenu-list`:
image

I'll open a corresponding issue in `psc-ide-emacs` :)

syl20bnr:
Thank you !
@kRITZCREEK Let us know if there is a better way to do it.
For now I cherry-picked it into develop branch.
@bjarkevad You can safely delete your branch.

PR #9658, opened 2017-09-27
shell: use toggle-truncate-lines-on in README

joehillen:
Fix docs for "Fish shell and ansi-term" and explicitly enable `truncate-lines`.

I was using the suggested hook as well as `(setq-default truncate-lines t)`,
so my fish shell was acting funny because `truncate-lines` was being disabled 
rather than enabled.


syl20bnr:
Thank you ! 👍 

Issue #9603, opened 2017-09-18
M-RET broken in org-mode


brabalan:
#### Description :octocat:
M-RET no longer bound to `org-meta-return` in org-mode. This is probably a consequence of the recent org-mode change of binding M-RET to `org-meta-return` instead of M-return.

#### Reproduction guide :beetle:
- Start Emacs
- Switch to Org Mode
- start a list
- type M-RET

*Observed behaviour:* :eyes: :broken_heart:
The menu for `M-m m` is shown.

*Expected behaviour:* :heart: :smile:
An entry is added to the list.

#### System Info :computer:
- OS: darwin
- Emacs: 25.3.1
- Spacemacs: 0.200.9
- Spacemacs branch: master (rev. 8e1af145)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
((auto-completion :variables auto-completion-enable-snippets-in-popup t company-minimum-prefix-length 4)
 bibtex
 (coq :variables proof-general-path "~/src/PG/generic/proof-site.el" proof-splash-enable nil)
 csv dash
 (elfeed :variables rmh-elfeed-org-files
         (list "~/.spacemacs.d/elfeedconfig.org")
         elfeed-use-curl t elfeed-sort-order 'ascending)
 emacs-lisp erc finance git gnus html ivy javascript
 (latex :variables latex-build-command "LatexMk" latex-enable-auto-fill t)
 markdown python ocaml org ruby
 (shell :variables shell-default-shell 'eshell)
 spell-checking themes-megapack
 (twitter :variables twittering-use-master-password t)
 yaml)
```
- System configuration features: NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS

duckdodger:
Came here to make this exact issue. I'm experiencing this problem too.

brabalan:
As additional information, here is what changed in org-mode.
Before:
``` emacs-lisp
  (org-defkey org-mode-map [(meta return)] 'org-meta-return)
```
After
``` emacs-lisp
  (org-defkey org-mode-map (kbd "M-RET") #'org-meta-return)
```

dindom999:
It drop me crazy. I spend a half day to find out reason, now pop out a menu...
![_001](https://user-images.githubusercontent.com/5274818/30676437-c3e9b110-9eb8-11e7-816f-f865b4fabcc3.png)

[...]

PR #9661, opened 2017-09-28
Bind org-meta-return for graphical display

darkfeline:
Org mode binds org-meta-return to M-RET, but in graphical mode
Spacemacs will use M-RET as the major mode leader.

https://github.com/syl20bnr/spacemacs/issues/9603

(I don't use Spacemacs any more, but I requested the Org mode change so I thought I should fix this downstream issue in Spacemacs.)

### DIFF
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -316,6 +316,12 @@ Will work on both org-mode and any mode that accepts plain html."
         (kbd (concat dotspacemacs-major-mode-emacs-leader-key " '"))
         'org-edit-src-exit)
 
+      ;; In graphical mode, the default value for
+      ;; `dotspacemacs-major-mode-emacs-leader-key', "C-M-m", will
+      ;; override org-mode's binding for "M-RET".
+      (when (display-graphic-p)
+        (define-key org-mode-map (kbd "M-<return>") 'org-meta-return))
+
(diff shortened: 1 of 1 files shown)


joehillen:
Thank you. I just happened to notice this problem yesterday.

syl20bnr:
What about the terminal ?
Does this work when emacs daemon is used ?

syl20bnr:
Thank you for the PR and thank you for using and contributing so much to Spacemacs 💜 
Also thank you to continue to contribute to it :-)

I added `M-RET` under the major mode leader key, not the best solution but `M-RET` is already taken in Spacemacs and in terminals `C-m` is `RET`,  so simply put this is a limitation of both the terminal and Spacemacs. We have to deal with it. 

So now `M-RET M-RET` is required to do what once was done with `M-RET`. Sad but there is no elegant solution to this problem.

HaoZeke:
This does not work for org-mode.
I get M-RET M-RET is not defined.

joehillen:
@HaoZeke It works for me. 

* `M-RET M-RET`
* `, M-RET`
* `SPC m M-RET`

All these behave the same for me.

Are you up to date with the develop branch?

HaoZeke:
Ah, no, my bad.. I was on the regular branch

PR #9671, opened 2017-09-29
Enable backward theme cycling and add dedicated transient-state

AdrieanKhisbe:
+ Enable to cycle backward using negative or universal Argument
+ Add a transient state bound to <kbd>SPC T N</kbd> to quickly navigate through the themes

Original Commit List
- update cycle-spacemacs-theme function to work backward with universal arg
- add a transient-state hydra to cycle through the modes
- move the transient-state definition in the +distribution spacemacs-base
- refactor using hydra syntax for expression as command
- modified cycle-theme to handle negative command argument
- add keybing for helm-themes in the transient-state

### DIFF
--- a/core/core-themes-support.el
+++ b/core/core-themes-support.el
@@ -281,20 +281,24 @@ THEME."
       (eval `(spacemacs|do-after-display-system-init
               (load-theme ',theme-name t))))))
 
-(defun spacemacs/cycle-spacemacs-theme ()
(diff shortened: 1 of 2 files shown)


syl20bnr:
Thank you again !  💜 
I removed the binding `SPC T N` and made `SPC T n` go to the next theme and start the transient state at the same time.
Commit https://github.com/syl20bnr/spacemacs/commit/037d1367

PR #9672, opened 2017-09-29
Improved buffer transient mode with extra bindings and new commands

AdrieanKhisbe:
- Add command to list buffer, burry them and also switch window
- Add left and right arrow keybinding to the previous/next buffer


syl20bnr:
Thank you ! 👍 
I changed the buffer transient state key bindings like this:
`b` -> list buffers as in `SPC b b`
`C-d` -> burry buffer (all all `b` bindings are already taken)
commit https://github.com/syl20bnr/spacemacs/commit/038100d1

Issue #9578, opened 2017-09-12
C/C++ layer incomplete documentation


wolfv6:
#### Description :octocat:

The "C/C++ layer" README on
https://github.com/syl20bnr/spacemacs/tree/master/layers/%2Blang/c-c%2B%2B#key-bindings
says:

    SPC m g a | open matching file (e.g. switch between .cpp and .h)

Would be nice if it also said, "this is a projectile function, so it only works from inside a project."

Also, the Spacemacs online help is missing "bound to SPC m g a".

#### Reproduction guide :beetle:
- Start Spacemacs
- M-x h d k
- Describe key: SPC m g a

*Observed behaviour:* :eyes: :broken_heart:
It is bound to C-c p a, M-RET g a, M-m m g a, and many ordinary text characters.

(projectile-find-other-file &optional FLEX-MATCHING)

*Expected behaviour:* :heart: :smile:
It is bound to SPC m g a

#### System Info :computer:
- OS: gnu/linux
- Emacs: 25.2.1
- Spacemacs: 0.200.9
- Spacemacs branch: master (rev. 8e1af145)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
(c-c++ emacs-lisp evil-snipe ivy org spell-checking)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND DBUS GCONF GSETTINGS NOTIFY ACL LIBSELINUX GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 MODULES XWIDGETS

CeleritasCelery:
You should report this upstream at projectile.

wolfv6:
@CeleritasCelery, thanks for your feedback.  I updated the issue to be more Spacemacs centric.  I don't think it's anything that projectile can fix.

wolfv6:
@CeleritasCelery that's right.

CeleritasCelery:
Go ahead and open a PR. This will be a great way for you to contribute to the project 😄 !

wolfv6:
OK.  I will submit a PR next year, after I learn more about editing Emacs.

CeleritasCelery:
This is just documentation, so no elisp required. Just Org mode.

smile13241324:
Tried to reproduce your issue on develop. 

[...]

PR #9678, opened 2017-10-01
Add remark for required project for "open matching file" commands

smile13241324:
In #9578 it was mentioned that the c++ layer documentation includes commands which only work when the files are part of a projectile project. 

I have added a remark to the commands in question that a project is necessary for those.


syl20bnr:
Thank you !

Issue #9285, opened 2017-07-26
Let's talk about treemacs and spacemacs


Alexander-Miller:
Remember [treemacs](https://github.com/Alexander-Miller/treemacs)? There's been a bit of talk about integrating it in spacemacs on the gitter a couple weeks ago, bit it quickly fizzled out since treemacs wasn't out on MELPA then. Well it's out now, and I'm currently busy cleaning up and writing tests for a big new feature - showing imenu (and potentially other) tags - so I'll have some free capacity soon. 

I like to think that my readme presents good enough a case for treemacs on its own (and I don't like attacking the competition to make a better case for myself), so if the interest on the side of spacemacs is still there I'd take a crack at integrating treemacs and spacemacs soon. I just need to know whether it should be done as an extra layer, or if we're going for a full replacement in the core. IMO the latter would be easier since then there would be no conflicts with e.g. winum-assign-func.

And of course if someone has an idea for icons I can use I'm all ears.

JAremko:
cc @syl20bnr

bmag:
I see there's a dependency on Emacs 25.1. Currently Spacemacs supports also 24.4 and 24.5, so we can't just replace Neotree with Treemacs in `spacemacs-navigation` layer (this is what counts as "full replacement in the core"). We have the tools to have both Neotree and Treemacs in the same layer, and only install and load one of them depending on Emacs version, but we probably don't want this complexity to be part of our default Spacemacs experience. (I'm not saying there's 0% chance, but personally I'm against it)

For now, I think the best way is to provide Treemacs as a separate optional layer (similar to imenu-list). I don't remember the exact semantics, but a layer in Spacemacs can cause other packages/layers to be excluded (meaning, they won't be installed). Treemacs can exclude Neotree by using `:excluded` in the layer's package list. I see that `spacemacs-navigation/init-winum` is going to set `winum-assign-func` even if Neotree is excluded - but this can be fixed.

Note that even a full replacement doesn't necessarily mean we get rid of Neotree - we could (and IMO should) just move it to an optional layer in this case. That means there's still a potential conflict with `winum-assign-func`. By the way, I'm not sure there's going to be any Treemacs-Neotree conflicts other than Winum integration and key-binding.

[...]

PR #9680, opened 2017-10-01
Add treemacs layer.

Alexander-Miller:
A layer for [treemacs](https://github.com/Alexander-Miller/treemacs), as discussed [here](https://github.com/syl20bnr/spacemacs/issues/9285).


fiveNinePlusR:
I would say this is a big win for spacemacs to get it merged into develop. Great package and nice integration. pretty low friction replacement for neotree.

JAremko:
Would be nice to have screenshot in the documentation (not too big one) :thinking:

syl20bnr:
And now we have the shadow mechanism to easily mark a layer to replace another.
I'll have time to merge some PRs at the end of the week, this PR will be part of the batch. I'll merge it myself as I have some reorganization to do with neotree as well.

Good job!

Alexander-Miller:
> And now we have the shadow mechanism to easily mark a layer to replace another.

Anything that I need to do here?

> Would be nice to have screenshot in the documentation

Added one to the description.

layers/+tools/treemacs/README.org
```diff
@@ -0,0 +1,106 @@
...
+
+#+BEGIN_SRC emacs-lisp
+  (treemacs
+   :variables
+   treemacs-use-follow-mode t
+   treemacs-use-filewatch-mode t
+   treemacs-use-collapsed-directories 3)
+#+END_SRC
+
+* Configuraion
```
nightscape: Typo

Alexander-Miller: Fixed it now.

syl20bnr:
Thank you ! 👍 
I moved the layer to `+filetree` directory and made some formatting fixes in README to make it more idiomatic.
I had to add `treemacs--window-number-ten` explicitly in `use-package` treemacs init otherwise I had breaking errors when helm is invoked before emacstree.

Nice job 💙 

Alexander-Miller:
Thanks for the help.

I looked around the new files and it looks like neotree's readme is listing features from the python layer. The linked image isnt there either. I'd put up a PR, but I dont know whats supposed to be in that list.

syl20bnr:
Thank you for noting this :-)
I can fix the list myself.

syl20bnr:
@Alexander-Miller I pushed a fix, thanks again !

duianto:
Does Treemacs in Spacemacs require python?
It's working fine with `emacs --quick`, `M-x package-initialize` and `M-x treemacs`.

In Spacemacs, this message is shown, the first time `SPC f t` is pressed, and every time `SPC f T` is pressed:
`apply: Searching for program: No such file or directory, python`

The treemacs window only shows the `*Roaming*` parent directory, when treemacs was called from `.spacemacs`, no directories or files are shown below the `*Roaming*` line.

These might be related to the python error:
- Trying to navigate with either the `vim` or `emacs` editing styles, doesn't work, emacs style reports:
`funcall-interactively: Buffer is read-only: #<buffer *Treemacs*>`
and with the vim style, the cursor moves left and right on the `*Roaming*` line, and up and down reports:
evil-line-move: Beginning of buffer
evil-line-move: End of buffer

- With the `vim` style, pressing `?` opens the `evil-ex-search-backward` "?" minibuffer prompt, instead of the hydra. With the `emacs` style, `?` shows:
`funcall-interactively: Buffer is read-only: #<buffer *Treemacs*>`

The following was only been tested in `emacs --quick`:
The hydra closes as soon as a key is pressed, ex: `n` to move to the next line. If it would stay open until it was closed manually by pressing `?` again, then one could try the key bindings while the hydra is open, without having to open it between each key press. That would also match the behaviour of the other spacemacs hydras.

#### System Info :computer:
- OS: windows-nt
- Emacs: 26.0.50
- Spacemacs: 0.200.9
- Spacemacs branch: develop (rev. 07014dee)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
((auto-completion :variables auto-completion-enable-help-tooltip 'manual auto-completion-enable-snippets-in-popup t)
 autohotkey emacs-lisp git helm markdown org spell-checking syntax-checking treemacs version-control)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG SOUND NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS MODULES

Alexander-Miller:
> Does Treemacs in Spacemacs require python?

The collapse directory feature does. Set the collapse-dirs value to 0 to disable it. I've already an update for the readme in the works, Ill add the part about python and adjust the default value accordingly.

Ill look at the other issues when I get home.

duianto:
That fixed it. 👍 
Navigation works, and `?` toggles the hydra, with both the emacs and vim editing style.

Maybe one solution could be to set it to `0`, if emacs is unable to find python.

Alexander-Miller:
> Maybe one solution could be to set it to 0, if emacs is unable to find python.

Opened a PR to do just that. Ill also put a more permanent hydra on my todo list.

robbert-vdh:
It would also be nice if the treemacs layer bound `SPC p t` instead of `SPC f p/P` to `treemacs-projectile-toggle`  so it's consistent with the existing neotree keybindings.

Alexander-Miller:
I tried just that at first, but it didnt work. pt was always bound to `projectile-find-tag`. Thats why I went with fp instead.

PR #9684, opened 2017-10-02
Prompt user to install react-mode if a .jsx file is opened

jonhermansen:
https://github.com/syl20bnr/spacemacs/tree/master/layers/%2Bframeworks/react#description

Automatic prompts for installing modes happened when I opened .js, .rb, and .py files. I thought... why not for JSX files too?

Also fixed a comment that said these modes are for "java".

I tested this change against master + develop branches. When I answer "y" to the prompt, react gets added to my dotspacemacs-configuration-layers in ~/.spacemacs


syl20bnr:
Thank you !

Issue #5570, opened 2016-03-23
layer github, gist list buffer couldn't use gist-list-reload


daleione:
I use vim mode, it said   (that binding is currently shadowed by another mode).
By the way, how to set the gist code title. When I use SPC g g r, It just use the filename as the title.

jsab:
The `g` binding is shadowed.

In order to be consistent with magit `gist-list-reload` the binding should probably be `g r`.

PR #9685, opened 2017-10-02
Add `g r` evilified binding to gist-list-mode

jsab:
Fix for https://github.com/syl20bnr/spacemacs/issues/5570

The emacs `g` binding get shadowed in hybrid and vim mode.

This creates an evilified binding for the same command using `g r` which is very consistent with magit.


syl20bnr:
Thank you ! 👍 

PR #9688, opened 2017-10-02
remove org table lighter in markdown mode

CeleritasCelery:
`orgtbl-mode` adds a rather large and unnecessary lighter to `markdown-mode` so we diminish it to keep the look clean.


syl20bnr:
Thank you !

PR #9694, opened 2017-10-04
Remove ess-R-object-popup from ess layer.

NGaffney:
As described in #9694, this package was removed from melpa for not having a GPL compatible license (it doesn't have any license at all) and is causing an error every time spacemacs starts.

A pull request myuhe/ess-R-object-popup.el#4 to add a license has been made in the original repository however it has sat for nearly a month unmerged, so I suggest removing the package from the layer in the meantime.

Alternatively it could be included directly into spacemacs as it's < 100 SLOC. Could there be licensing issues if we did that?

### DIFF
--- a/layers/+lang/ess/README.org
+++ b/layers/+lang/ess/README.org
@@ -43,11 +43,10 @@ Send code to inferior process with these commands:
 ** Helpers
 Helpers for inspecting objects at point are available in R buffers only.
 
-| Key Binding | Description                                                         |
-|-------------+---------------------------------------------------------------------|
-| ~SPC m h d~ | view data under point using [ess-R-data-view][ess-R-data-view]      |
-| ~SPC m h i~ | object introspection popup [ess-R-object-popup][ess-R-object-popup] |
-| ~SPC m h t~ | view table using [ess-R-data-view][ess-R-data-view]                 |
+| Key Binding | Description                                                    |
+|-------------+----------------------------------------------------------------|
+| ~SPC m h d~ | view data under point using [ess-R-data-view][ess-R-data-view] |
+| ~SPC m h t~ | view table using [ess-R-data-view][ess-R-data-view]            |
 
 * Options
 To turn off the automatic replacement of underscores by =<-=, set in your
--- a/layers/+lang/ess/packages.el
+++ b/layers/+lang/ess/packages.el
@@ -13,7 +13,6 @@
       '(
         ess
         ess-R-data-view
-        ess-R-object-popup
         (ess-smart-equals :toggle ess-enable-smart-equals)
         golden-ratio
         org))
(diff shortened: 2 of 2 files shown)


mathcass:
I think removing it from the `ess` layer makes sense unless the author can address the license issue. 

For others coming across this and wanting to suppress trying to download `ess-R-object-popup` each time Emacs starts up, you can add the package to the `dotspacemacs-excluded-packages` list within the `.spacemacs` file, like: 

```
   ;; A list of packages that will not be installed and loaded.
   dotspacemacs-excluded-packages
   '(
     ess-R-object-popup)
```

bmag:
> Alternatively it could be included directly into spacemacs as it's < 100 SLOC. Could there be licensing issues if we did that?

I'm pretty sure it's ok to write new code that does the same thing, as long as our implementation is different enough than the original (so no copy-pasting), but I'm not a lawyer.

The source file has a copyright [comment](https://github.com/myuhe/ess-R-object-popup.el/blob/7e1f601bfba72de0fda44d9c82f96028ecbb9948/ess-R-object-popup.el#L6) saying "all rights reserved", so maybe it wasn't ok to include in Melpa from the beginning, and I agree we should remove the package. I'll see if I can write a function to replace the package, but I'll need someone to help test it.

bmag:
And also, @NGaffney there is documentation (layer's readme) and some code (init function, key-binding) you didn't remove.

bmag:
Question: what if we use the command `ess-help` (alias for `ess-display-help-on-object`) instead of the popup? Any of you using ESS and can check it out?

gdkrmr:
I am no lawyer so this comes without any warranty at all, reading  this: 

https://softwareengineering.stackexchange.com/questions/276144/using-unlicensed-code#276147
and this:
https://help.github.com/articles/licensing-a-repository/
and this:
https://opensource.stackexchange.com/questions/1640/if-im-using-a-gpl-3-library-in-my-project-can-i-license-my-project-under-mit-l

My opinion is the following:
The default license is definitely "all rights reserved" but because it is on github, you can fork it and view it. I think you are not allowed to redistribute it but you could
1) make some code (i.e. a recipe) that clones the repository.
or 2) write new code that has the same functionality and publish that with a license. If you develop the code independently this shouldn't be a problem.

bmag:
@gdkrmr do you think `ess-display-help-on-object` is an adequate replacement? I'd rather we use it (if adequate) than re-write the popup code.

NGaffney:
Thanks @bmag, I've removed the init function, keybinding and the reference in the doc and updated the PR.

gdkrmr:
@bmag I always use `ess-display-help-on-object` anyways, because the popup only showed the very first part of a function and it mostly showed the help of the `default` version which is often not useful.

syl20bnr:
Thank you !

bmag:
The commit was cherry-picked before @NGaffney modified it. I added the missing parts in #9746 (and also unrelated missing sections in the layer's readme)

PR #9709, opened 2017-10-07
Make nose fail compilation when tests fail

wizmer:
While nose.run() returns a boolean, nose.main() returns 0 or 1 depending of the test results.
This number can be used to set the status of the `*compilation*` buffer.
So now if the tests fail, the `*compilation*` buffer will be in state `exited abnormally`


syl20bnr:
Thank you ! 💜 

PR #9714, opened 2017-10-09
Treemacs layer readme & config fixup

Alexander-Miller:
Minor readme fixes and making sure we dont run into trouble when theres no python. Its difficult to see in the diff -  the change to the keybind table is due to the final column being misaligned. 
Changes are small enough that I dont think its worth splitting.


duianto:
Tested the check for python line, and Treemacs work as expected when python isn't installed 👍

syl20bnr:
Thank you ! 👍 

PR #9716, opened 2017-10-09
No need to install evil-org for emacs style

### LINKED ISSUE #9713: evil-org must not be installed if editing style is emacs
#### Description :octocat:
Spacemacs attempts to install evil-org when when my editing style is **emacs**!

- Start Emacs
- Add `org` layer
- Save and restart

```
--> installing package: evil-org@org... [1/12]
An error occurred while installing evil-org (error: (file-error https://elpa.gnu.org/packages/org-plus-contrib-20171009.tar Not found))
```

evil-org must not be installed if my editing style is emacs

#### System Info :computer:
- OS: gnu/linux
- Emacs: 26.0.60
- Spacemacs: 0.200.9
- Spacemacs branch: develop (rev. 07014de)
- Graphic display: t
- Editing style: emacs
- Completion: helm
- Layers:
```elisp
(helm emacs-lisp neotree
      org)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GSETTINGS NOTIFY ACL LIBSELINUX GNUTLS LIBXML2 FREETYPE LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK2 X11 MODULES

sdwolfz:
Fixes #9713.

### DIFF
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -15,7 +15,7 @@
         company-emoji
         emoji-cheat-sheet-plus
         (evil-org :toggle (not (equal dotspacemacs-editing-style 'emacs)))
-        evil-surround
+        (evil-surround :toggle (not (equal dotspacemacs-editing-style 'emacs)))
         gnuplot
         htmlize
         ;; ob, org and org-agenda are installed by `org-plus-contrib'
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -14,7 +14,7 @@
         company
         company-emoji
         emoji-cheat-sheet-plus
-        evil-org
+        (evil-org :toggle (not (equal dotspacemacs-editing-style 'emacs)))
         evil-surround
         gnuplot
         htmlize


sdwolf:
@kaushalmodi I've implemented your second request as well.

kaushalmodi:
Thanks! Looks like this fixed the `evil-org` issue, but I get this warning:

> (Spacemacs) Warning: Ignoring :toggle for package evil-surround because layer org does not own it.

kaushalmodi:
Copying the original author (@TheBB) who added `evil-surround` in https://github.com/syl20bnr/spacemacs/commit/3fb38773b326db1d18a811e229bf9054c349b696.

sdwolf:
@kaushalmodi I've looked into this more and I see that this PR might not be a good idea... simply disabling every evil package for emacs edition mode is not something spacemacs was build with in mind. The `evil-surround` is also used with the `keyboard-layout` spacemacs layer, together with other `evil` related packages.

Maybe the best solution for you would be simply adding `evil-org` and `evil-surround` to `dotspacemacs-excluded-packages`, inside your `~/.spacemacs` file.

I'm afraid a better solution is beyond my capabilities right now.

kaushalmodi:
> simply disabling every evil package for emacs edition mode is not something spacemacs was build with in mind

Hopefully @syl20bnr can jump in to clarify this. I have often heard that Spacemacs is not just for vim users. 

I don't use Spacemacs, but in my attempt to support the ox-hugo package for a Spacemacs user, I installed Spacemacs with emacs bindings (which I am more comfortable in).  Then why should evil packages get installed, and then block my simple task by errors related to those packages that I will never use?

Having Spacemacs install utility packages by default which a user might use *at some point* is a different thing than installing packages that would **never** be used as they don't fit the user's habits (emacs vs vim).

ellakk:
Spacemacs guidelines states that a package should only be initialized in one place. The surround package is initialized in spacemacs/spacemacs-evil/packages.el:
`(defun spacemacs-evil/init-evil-surround ()`

There is no need to disable evil-surround in the org layer because it's a post-init block:
`(defun org/post-init-evil-surround ()`

This will not install any packages and the code in the post-init block will only run if evil-surround is enabled in the spacemacs-evil layer.

sdwolfz:
According to this [comment](https://github.com/syl20bnr/spacemacs/pull/9587#issuecomment-358545718) this PR is not a good idea. I'm going to close it for now. If evil packages still create problems then maybe someone with a bit more experience can drop by and help.

PR #9470, opened 2017-08-20
Remove themes that were deleted from Melpa

henriquenj:
This makes Spacemacs stop checking their existence every time it starts,
which performs a refresh of every package archive and makes startup
slower.

firebelly, niflheim and tronesque were all removed from MELPA because
they either don't have licenses or have licenses that are incompatible
with GPL3.

The relevant commits on MELPA are

melpa/melpa@cf92ce1a2bc92c73071d157e7e223982bbd3f818
melpa/melpa@c3366117f3399660509f62849a88dc8be56a1fef

If the themes are added again in the future, we can simply uncomment the
lines.

PR #9717, opened 2017-10-10
Update themes megapack

nickclasener:
I commented out the unavailable themes

### DIFF
--- a/layers/+themes/themes-megapack/packages.el
+++ b/layers/+themes/themes-megapack/packages.el
@@ -18,7 +18,7 @@
     apropospriate-theme
     anti-zenburn-theme
     ;; contains errors
-    ; badger-theme
+    ;; badger-theme
     badwolf-theme
     birds-of-paradise-plus-theme
     bubbleberry-theme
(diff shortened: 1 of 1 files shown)


duianto:
Thanks for submitting a PR, but:

source:
https://github.com/syl20bnr/spacemacs/blob/master/CONTRIBUTING.org#pull-request

4 of the themes have already been removed on the develop branch:
Remove themes that were deleted from Melpa #9470

`zonokai-theme` was removed 6 days ago: https://github.com/melpa/melpa/commit/5bf94592d6e06f849b434f8da1f14c9c2fe85c6e so that one is new.

Your right that themes that have been commented out probably should use two semicolons, however, themes that are commented out, might never get fixed/re-added to melpa, therefore it probably is better to remove them, and only add them back if/when they are forked/fixed and someone requests/submits a PR for them.

So if you want to submit a new PR, then make it against the develop branch, and remove the `zonokai-theme`, you could also remove these lines:
```
    ;; contains errors
    ; badger-theme
```
and
```
    ;; contains error
    ; tommyh-theme
```

The current fix on the master branch is to add the removed themes to the excluded packages list in your `.spacemacs`:
```elisp
   dotspacemacs-excluded-packages '(firebelly-theme
                                    niflheim-theme
                                    pastels-on-dark-theme
                                    tronesque-theme
                                    zonokai-theme)
```

syl20bnr:
Thank you !

PR #9718, opened 2017-10-10
Fix typo in layer shadowing documentation

benjamreynolds:


duianto:
There seems to also be a filename typo in the Shadowing sections, third paragraphs, first sentence:
>A layer can shadow other layers by calling in its config.el file the function configuration-layer/declare-shadow-relation.

it states that the `configuration-layer/declare-shadow-relation` function should be called in the layers `config.el` file. But both the `helm` and `ivy` layers call the function in `layers.el`.

https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Bcompletion/helm/layers.el#L12
https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Bcompletion/ivy/layers.el#L12

syl20bnr:
@duianto thank you for spotting this documentation bug, fixing it :-)

syl20bnr:
Thank you ! 👍 

PR #9721, opened 2017-10-11
Correct undo-tree buffer name in layer spacemacs-visual

winkey728:
There is a space in the buffer name of undo-tree. The popwin did not display right and the function `undo-tree-visualizer-quit` also seems not work well until I correct the buffer name.


syl20bnr:
Thank you ! 👍 

PR #9726, opened 2017-10-12
Adds keybinding for updating packages

sdwolf:
This is an action I do quite often since I like to have the latest packages (with the latest fixes) but I either have to type the name of the command inside `M-x` or navigate in the home buffer and press `C-m` on the `[Update Packages]` text.

I belive it's time to have a keybinding for this. I chose `SPC h` for it since it was the closest I can find to a spacemacs related prefix. Other that went trough my mind were: `SPC S` (for spacemacs) or `SPC E` (for emacs, or Editor).


syl20bnr:
Thank you ! 👍 
I moved the key binding under `SPC f e` prefix as it stands for `File Emacs` and we already have the dotfile here, so feel natural to me :-)

PR #9728, opened 2017-10-13
Fix name of window-purpose in readme

aaronjensen:
The name of the package is actually `window-purpose`


syl20bnr:
Thank you ! 👍 

PR #9734, opened 2017-10-15
spacemacs-evil-cursors: make them easy to configure

bmag:
Motivation: so layers with their own evil states (e.g. treemacs) can also
contain their own cursor configuration

Example usage: `(spacemacs/add-evil-cursor "treemacs" "RoyalBlue1" '(hbar . 0))`


bmag:
merged

PR #4578, opened 2016-01-12
Fix early creation of empty org directory

tboby:
Resolves #4145,

Commit message:
Move org-default-notes-file from ~/.notes to org-directory/notes.org
and remove forced creation of org-directory. User will be prompted
to create directory on completing capture.

PR #9750, opened 2017-10-18
fix org-default-notes-file precedence

### LINKED ISSUE #9748: org mode layer is stepping on my customized value for `org-default-notes-file`
<<REPLACE ALL UPPERCASE EXPRESSIONS AND DELETE THIS LINE, `C-c C-c` TO SUBMIT>>
#### Description :octocat:
<<DESCRIPTION OF THE PROBLEM>>
org mode layer is stepping on my customized value for `org-default-notes-file` leading to notes.org files all over my hd that are not added to my agendas
- Start Emacs
- <<REPRODUCTION STEPS>>
customize variable `org-default-notes-file` to something other than the default and save for future sessions

note that the value you set is in your `.spacemacs`

<<DESCRIPTION OF THE OBSERVED BEHAVIOUR>>
restart spacemacs navigate to `.spacemacs` to where the variable was set by customize

`SPC hdv <enter>` note that the value is "notes.org" and not what was set in customize

<<DESCRIPTION OF THE EXPECTED BEHAVIOUR>>
dont step on my customized variable or if you do, make it something actually useful. 

source of the problem https://github.com/syl20bnr/spacemacs/blob/bd7ef98e4c35fd87538dd2a81356cc83f5fd02f3/layers/%2Bemacs/org/packages.el#L297
#### System Info :computer:
- OS: darwin
- Emacs: 25.2.1
- Spacemacs: 0.200.9
- Spacemacs branch: master (rev. 8e1af145)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(scheme helm csv sql auto-completion clojure common-lisp
        (ruby :variables ruby-test-runner 'rspec ruby-version-manager 'rbenv)
        python javascript emacs-lisp markdown
        (org :variables org-enable-reveal-js-support t org-enable-org-journal-support t)
        nim syntax-checking version-control git osx shell ibuffer yaml html colors erc)
```
- System configuration features: NOTIFY ACL LIBXML2 ZLIB TOOLKIT_SCROLL_BARS

#### Backtrace :paw_prints:
```
<<BACKTRACE IF RELEVANT>>
```

CeleritasCelery:
Fixes #9748, which demonstrates `org-default-notes-file` is overriding user configuration because it set after org is loaded. Moved it to `:init` so that is can be overridden by the user.

### DIFF
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -103,6 +103,8 @@
                                           ".org-id-locations")
             org-publish-timestamp-directory (concat spacemacs-cache-directory
                                                     ".org-timestamps/")
+            org-directory "~/org" ;; needs to be defined for `org-default-notes-file'
+            org-default-notes-file (expand-file-name "notes.org" org-directory)
             org-log-done t
             org-startup-with-inline-images t
             org-image-actual-width nil
@@ -309,8 +311,6 @@ Will work on both org-mode and any mode that accepts plain html."
       (define-key global-map "\C-cc" 'org-capture))
     :config
     (progn
-      (setq org-default-notes-file "notes.org")
-
       ;; We add this key mapping because an Emacs user can change
       ;; `dotspacemacs-major-mode-emacs-leader-key' to `C-c' and the key binding
       ;; C-c ' is shadowed by `spacemacs/default-pop-shell', effectively making


bmag:
Why did you put the file in the cache dir? It doesn't seem to belong there.

CeleritasCelery:
Where would the unqualified “notes.org” go? `$HOME`? Current directory? 

If `.cache` is not the place, we should should at least add a qualifier for where ever it would go. Where do you think that should be?

bmag:
Current directory, I guess. I don't really use org-capture, do you? How about `(expand-file-name "notes.org" org-directory)` ?

CeleritasCelery:
From looking at the message of the [commit](https://github.com/syl20bnr/spacemacs/pull/4578/commits/f9b712e333e2e11a296619f88811dc0b389db6c5) that originally added this line, `org-directory/notes.org` was the intended location. I will update the PR.

bmag:
I tested it and I get an error because `org-directory` isn't defined yet in `:init`. (void-variable)
Easily fixable by setting `org-directory` to `"~/org"` (the default value) early in `:init`, with comment about why we need to set it.

CeleritasCelery:
Well that's embarrassing 😕 . I pushed a fix. And tested it this time 😄. Though looking at it, I see no reason that we can't just remove this code. Having `org-default-notes-dir` in `~/org/notes.org` vs `~/.notes.org` (the org default) seems just to be a matter of personal preference.

bmag:
Thank you :+1: Cherry-picked into develop, you can safely delete your branch. Yes, it's borderline personal preference, but I prefer this location (and so did the author of the link you did).

bcc32:
This PR has the unfortunate side-effect of causing org-default-notes-file to be bound but not `defcustom`'d before org-mode is loaded.  This means `custom-save-variables` (which happens whenever spacemacs installs packages, because of `package-selected-packages`) marks `org-default-notes-file` to be evaluated immediately on startup.  That breaks when `org-default-notes-file` is customized to an expression involving, say, `org-directory`:

```elisp
 '(org-default-notes-file (expand-file-name "refile.org" org-directory))
```

I think Spacemacs shouldn't change this variable from the org-mode default, and let users customize it instead if they wish.  The "correct" way to do this anyway would probably be to use `custom-theme-set-variables` with a `spacemacs` theme, so that the user customizations override Spacemacs's suggested defaults.

perryprog:
I can second this issue—this essentially makes it impossible to customize `org-default-notes-file` with `custom-set-variables`, which is a tad annoying (albeit workaround-able).

PR #9753, opened 2017-10-18
Make popwin handle google-translate buffers

ellakk:
Make popwin handle google-translate buffers. I figured it should go in the init-popwin block instead of the init-google-translate block since undo-tree which is also part of the default distribution is there already.


syl20bnr:
Thank you again !

PR #9754, opened 2017-10-18
Improve coffeescript support

ellakk:
Register REPL and add major-mode bindings.

Spacemacs got coffeescript support in the javascript layer but it does not add any keybindings and does not register the REPL.


bmag:

PR #9755, opened 2017-10-19
Switch back to buffer from Intero REPL

tonylotts:
Add ```, s S``` to switch back to the originating buffer, from the Intero REPL buffer.
Symmetric with haskell-interactive-mode.

### DIFF
--- a/layers/+lang/haskell/funcs.el
+++ b/layers/+lang/haskell/funcs.el
@@ -81,8 +81,8 @@
   (dolist (mode (cons 'haskell-cabal-mode haskell-modes))
     (spacemacs/set-leader-keys-for-major-mode mode
       "sc"  nil
-      "ss"  'haskell-intero/display-repl
-      "sS"  'haskell-intero/pop-to-repl))
+      "sS"  'haskell-intero/display-repl
+      "ss"  'haskell-intero/pop-to-repl))
 
   (dolist (mode (append haskell-modes '(haskell-cabal-mode intero-repl-mode)))
     (spacemacs/declare-prefix-for-mode mode "mi" "haskell/intero")
@@ -118,13 +118,13 @@
 
 (defun haskell-intero/display-repl (&optional prompt-options)
   (interactive "P")
-  (let ((buffer (intero-repl-buffer prompt-options)))
+  (let ((buffer (intero-repl-buffer prompt-options t)))
     (unless (get-buffer-window buffer 'visible)
       (display-buffer buffer))))
 
 (defun haskell-intero/pop-to-repl (&optional prompt-options)
   (interactive "P")
-  (pop-to-buffer (intero-repl-buffer prompt-options)))
+  (pop-to-buffer (intero-repl-buffer prompt-options t)))
 
 (defun haskell-intero//preserve-focus (f &rest args)
   (let ((buffer (current-buffer)))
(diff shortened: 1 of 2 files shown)


layers/+lang/haskell/funcs.el
```diff
@@ -124,7 +124,7 @@
 
 (defun haskell-intero/pop-to-repl (&optional prompt-options)
   (interactive "P")
-  (pop-to-buffer (intero-repl-buffer prompt-options)))
+  (pop-to-buffer (intero-repl-buffer prompt-options t)))
```
bmag: Shouldn't `haskell-intero/display-repl` receive the same treatment?

tonylotts: @bmag, I agree about ',ss'; as it's much easier to type, and (partially) consistent with the other language layers.

I haven't verified in-depth, but from memory, there is a bit of inconsistency of the switch to/from REPL functionality across languages.  I think it should be consistent.  A few languages layers bind ',ss' (Clojure, Racket).  Many layers bind ',si' to start the REPL.  I think in some cases ',si' this starts the REPL buffer in the background, in others it starts and switches to the REPL.

If you prefer, I can close this PR, and submit a PR that changes the keybinding for both Intero and haskell-interactive to ',ss'.

bmag: You can just update this PR by pushing more commits (normally we prefer that you also squash them, but if that's a problem then I can do it myself).

What about `haskell-intero/display-repl`? Shouldn't its call to `intero-repl-buffer` also be modified?

tonylotts: @bmag I applied the same treatment to `haskell-intero/display-repl`, and swapped ',ss' and ',sS'.

layers/+lang/haskell/packages.el
```diff
@@ -211,8 +214,8 @@
         "n"   'haskell-cabal-next-subsection
         "p"   'haskell-cabal-previous-subsection
         "sc"  'haskell-interactive-mode-clear
-        "ss"  'spacemacs/haskell-interactive-bring
-        "sS"  'haskell-interactive-switch
+        "sS"  'spacemacs/haskell-interactive-bring
+        "ss"  'haskell-interactive-switch
```
bmag: Not sure why you swapped these two. Better revert it.

layers/+lang/haskell/packages.el
```diff
@@ -148,8 +149,8 @@
 
           "sb"  'haskell-process-load-file
           "sc"  'haskell-interactive-mode-clear
-          "ss"  'spacemacs/haskell-interactive-bring
-          "sS"  'haskell-interactive-switch
+          "sS"  'spacemacs/haskell-interactive-bring
+          "ss"  'haskell-interactive-switch
```
bmag: Not sure why you swapped these two. Better revert it.

tonylotts:
Hey @bmag, as you mentioned, this is why I swapped `,ss` and `,sS`.
> I think haskell-interactive-switch-back and intero-repl-switch-back should both be bound to ,ss instead of ,sS.

Per your suggestion I bound switch to/from REPL to `,ss`.  I had a look at some other language layers for a consistent binding to show the REPL (without switching).  I didn't find any example in another layer, so I bound it `,sS`; which I think makes sense.  Hence `,ss`, and `,sS` are swapped.

Thanks for your input thus far.  I'm eager to have this merged.

bmag:
Thank you :+1: Cherry-picked into develop, you can safely delete your branch. Sorry for the delay, I'm not sure why I haven't merged this earlier.

PR #9759, opened 2017-10-20
Remove unnecessary comments in init.el

ReimuXMX:
Hello

Thanks for your work. Your work makes more people know and use Emacs. Good job~~

In order to prevent Emacs25  adding "(package-initialize)", we only need to create (package-initialize) anywhere in init.el. Do not need a separate line.

I'm sorry for my very poor English.

### DIFF
--- a/init.el
+++ b/init.el
@@ -10,7 +10,6 @@
 ;;; License: GPLv3
 
 ;; Without this comment emacs25 adds (package-initialize) here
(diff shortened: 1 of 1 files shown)


bmag:
We don't want to call `package-initialize` in `init.el`, and we don't want Emacs to add it. That's why the comment is there.

PR #9760, opened 2017-10-20
Change Frame delete bindings from c/C to d/D

duianto:
The Frame delete bindings SPC F c and C, are inconsistent with the delete window
and buffer bindings, that use: d and D

---

I don't know what `"Fd" 'dired-other-frame` should be changed to, so I just commented it out.

I can change it if someone has a suggestion.

### DIFF
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -2162,11 +2162,12 @@ Frame manipulation commands (start with ~F~):
 | Key Binding | Description                                         |
 |-------------+-----------------------------------------------------|
 | ~SPC F f~   | open a file in another frame                        |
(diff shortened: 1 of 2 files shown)


sdwolf:
Maybe `"FO" 'dired-other-frame` since `"Fo" 'other-frame` already exists.

duianto:
Done, I also added the missing entry in the documentation for the dired other frame binding.

bmag:
Thank you :+1: Cherry-picked into develop, you can safely delete your branch. "FO" isn't perfect, because "open <something> in other frame" is pretty generic. However, I couldn't find a better alternative ("FI", "FR", "FF"?) so we'll go with "FO".

PR #9761, opened 2017-10-20
Update themes transient state, t for helm-themes

duianto:
Add the key binding: t
to open helm-themes, without having to use the arrow up key.

Reorder/rename: cycles backward and cycles forward
to: next and previous

Next is listed first, because it's the most common action, and it matches the
order in other transient states.

Change the t argument in the call to spacemacs/cycle-spacemacs-theme, to the
symbol 'backward, to make it clear what the argument does.

Add documentation for the Themes Transient State bindings.

### DIFF
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -64,6 +64,7 @@
   - [[#color-themes][Color themes]]
     - [[#default-theme][Default theme]]
     - [[#choosing-themes][Choosing themes]]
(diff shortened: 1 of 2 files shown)


duianto:
Sorry about that.

I see now, that my limited test case of only having the two default `spacemacs-dark` and `spacemacs-light` themes in the `dotspacemacs-themes` list, hid the fact that cycling backwards actually was cycling forwards.

PR #9762, opened 2017-10-21
Update haskell setup-intero func

PierreR:
This change allows the use of white/blacklisting
see https://github.com/commercialhaskell/intero#whitelistingblacklisting-projects

### DIFF
--- a/layers/+lang/haskell/funcs.el
+++ b/layers/+lang/haskell/funcs.el
@@ -50,7 +50,7 @@
   (add-to-list 'company-backends-haskell-mode
                '(company-intero company-dabbrev-code company-yasnippet))
   (push 'intero-goto-definition spacemacs-jump-handlers)
-  (intero-mode)
+  (intero-global-mode 1)
   (dolist (mode haskell-modes)
     (spacemacs/set-leader-keys-for-major-mode mode
       "hi" 'intero-info


syl20bnr:
The repository is gone, I cannot merge this PR.

I looked at the Haskell and found that there is no isolation with the different backends. Backend should be able to be set by project using the local directory variables. An example of this mechanism can be found in Java layer: https://github.com/syl20bnr/spacemacs/tree/develop/layers/%2Blang/java#choosing-a-backend

If we correctly isolate the configuration of completion backends for Haskell then the `white/blacklisting` feature would not be necessary. 

I prefer to avoid to enable global modes in Spacemacs if possible as isolation is a very important property to maintain as much as possible.

d12frosted:
@syl20bnr I've prepared a PR for this: https://github.com/syl20bnr/spacemacs/pull/11550

duianto:
We'll close this because:
- The PR is against the `master` branch (they should be against the `develop` branch)
- The linked issue: intero-blacklist and intero-whitelist not working #364
seems to have instructions for configuring the `intero-blacklist`.
- The linked PR #11550 was superseded by this merged PR:
Haskell: add LSP support via HIE and lsp-haskell #11887

PR #9779, opened 2017-10-25
Make auto-completion triggers more friendly in "react" layer

ralph-cheng:
Auto completion menu triggers too frequently in react-mode, for example, the
completion menu will pop up right after user finishes typing a javascript
statement.

The root cause is that the 'company-minimum-prefix-length variable has been set
to 0 in "html" layer, which is used by "react" layer.

Overriding this variable to 2, as defined in "auto-completion" layer, will
resolve this issue.

Thank you <3


syl20bnr:
Thank you ! 👍 

Issue #9549, opened 2017-09-06
Describe variable fails due to missing definition of "helm-bookmark-map" on develop


smile13241324:
#### Description :octocat:
Since the latest update on develop every call of `SPC h d v` fails due to an undefined variable `helm-bookmark-map`.

#### Reproduction guide :beetle:
- Start Emacs
- Press `SPC h d v` and select a variable of your choice
- The error appears

*Observed behaviour:* :eyes: :broken_heart:
The symbol `helm-bookmark-map` is referenced but not defined which causes emacs to fail at displaying the variable description. A quick smart search showed me the file is referenced from the helm layer's packages.el only.

*Expected behaviour:* :heart: :smile:
Show the description

#### System Info :computer:
- OS: gnu/linux
- Emacs: 25.2.1
- Spacemacs: 0.200.9
- Spacemacs branch: develop (rev. 6d73df70)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(helm nim scheme
      (auto-completion :variables auto-completion-return-key-behavior 'complete auto-completion-tab-key-behavior 'cycle auto-completion-complete-with-key-sequence nil auto-completion-complete-with-key-sequence-delay 0.0 auto-completion-private-snippets-directory nil auto-completion-enable-snippets-in-popup nil auto-completion-enable-help-tooltip t auto-completion-enable-sort-by-usage t)
      better-defaults
      (clojure :variables clojure-enable-fancify-symbols t)
      gpu graphviz
      (plantuml :variables plantuml-jar-path "~/.plantuml/plantUml.jar" org-plantuml-jar-path "~/.plantuml/plantUml.jar")
      salt ibuffer emacs-lisp pdf-tools cscope gtags
      (org :variables org-want-todo-bindings t org-enable-github-support t org-enable-bootstrap-support t org-enable-reveal-js-support t)
      (shell :variables shell-enable-smart-eshell t shell-default-shell 'ansi-term shell-default-height 60 shell-default-position 'bottom shell-default-full-span nil shell-default-term-shell "/usr/bin/fish")
      syntax-checking version-control semantic themes-megapack
      (c-c++ :variables c-c++-default-mode-for-headers 'c++-mode c-c++-enable-clang-support t c-c++-enable-clang-format-on-save t)
      git github sml pass common-lisp
[...]

PR #9781, opened 2017-10-26
[master] add work-around for bookmark bug into helm layer

waisbrot:
This is against master and not develop.

I'd like to have a fix for #9549, which keeps biting me.

This PR writes the work-around, suggested by @bmag, into the Helm layer. This fixes the issue for me.

Please feel free to discard this PR and make the change from scratch if that's easier.


waisbrot:
Thanks for fixing!

PR #9784, opened 2017-10-27
Fix using eyebrowse without persp-mode

Stebalien:


duianto:
Formatting:
The last two hooks can probably be split into two lines, like the hooks above, to keep their lengths under 80.

syl20bnr:
Thank you ! 👍 

Issue #8663, opened 2017-04-07
Jump-to-definition could be improved


agzam:
#### Description :octocat:
Jump-to-definition is awful in Clojure and Clojurescript

#### Explanation
The way how Spacemacs handles `jump-to-definiton` (`g d` or `, g g`) with `core-jump.el` is nice, yet it does not work as one would expect. Currently my main language of use is Clojure, so I can only describe Clojure here, it may be applicable to other languages as well.

Clojure layer defines jump-handlers and adds `cider-find-var` to the list.

There is a couple of problems with this:

a) `cider-find-var` only works when there's a connected repl and miserably fails when there's no REPL 
b) `cider-find-var` prompts for a symbol to jump to, even when the cursor on a symbol, although `cider-prompt-for-symbol` is set to nil in the layer's packages.el. I'm not sure about this one - might be an upstream issue

#### Proposed solution

Maybe we could do something like this:

- If there's a connected REPL (cider jacked in) - then try to use `cider-find-var` (without it prompting for a symbol)
- when there's no repl - use `dumb-jump-go`. 

Can someone suggest maybe even better way? Thanks!

### Also

it seems that `g d` and `, g g` - strictly not the same. sometimes the first one uses `evil-goto-definition` and not the function listed in jump-handler list. Can someone elaborate on the differences and how to make (if they should be) the same thing?

#### System Info :computer:
- OS: darwin
- Emacs: 25.1.1
- Spacemacs: 0.200.9
- Spacemacs branch: develop (rev. 12eacab6)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
((org :variables org-enable-reveal-js-support t org-enable-bootstrap-support t org-enable-github-support t :packages
      (not org-present))
 csv yaml emacs-lisp javascript
 (html :packages
       (not pug-mode slim-mode))
 (markdown :variables markdown-live-preview-engine 'vmd)
 (shell-scripts :packages
                (not fish-mode))
 clojure lua
 (ruby :variables ruby-version-manager 'rvm)
 haskell
 (auto-completion :variables auto-completion-tab-key-behavior 'cycle auto-completion-enable-snippets-in-popup t auto-completion-enable-help-tooltip t auto-completion-enable-sort-by-usage t)
 syntax-checking spell-checking helm
 (shell :variables shell-enable-smart-eshell t comint-scroll-show-maximum-output nil comint-move-point-for-output nil)
 fasd
[...]

PR #9792, opened 2017-10-28
Improves jump-to-definition for Clojure modes

agzam:
### What does this PR do?

Improves/fixes jump-to-definition for Clojure layer

### Rationale

Jump to definition in Clojure modes doesn't work reliably.

- CIDER's default `cider-find-var` only works when CIDER is connected or jacked-in and even then it doesn't always work. 
  - it doesn't work in cljs, cljc files and cider-repl-mode
  - sometimes it doesn't correctly recognize symbol-at-point and prompts user to enter one

The idea is to use `cider-find-var` whenever possible. If it is not available, or if it fails: use the fallback option - `dumb-jump-go`, it is simple and it works reliably in most cases (but CIDER's fn still preferable).

### How to test?

- Open any Clojure/Clojurescript project
- try using `g d` and `, g g` 
- try in Clojure, Clojurescript and .cljc files if possible
- execute M-x `cider-jack-in`
- wait for CIDER to get connected to the REPL
- try using `g d` and `, g g` again

### DIFF
--- a/layers/+lang/clojure/funcs.el
+++ b/layers/+lang/clojure/funcs.el
@@ -170,3 +170,12 @@ If called with a prefix argument, uses the other-window instead."
   (when (memq dotspacemacs-editing-style '(hybrid vim))
     (evil-make-overriding-map cider--debug-mode-map 'normal)
     (evil-normalize-keymaps)))
+
+(defun spacemacs/clj-find-var ()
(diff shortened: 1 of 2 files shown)


agzam:
related: https://github.com/syl20bnr/spacemacs/issues/8663

eggsyntax:
I've been using this for a week, and it's been a nice improvement over the default behavior. I think it'd make a nice improvement to SM.

syl20bnr:
Thank you ! 💜 

Issue #9598, opened 2017-09-17
ox-twbs doesn't work automatically


amosbird:
#### Description :octocat:
ox-twbs isn't in load-path

#### Reproduction guide :beetle:
- Start Emacs
- Enable ox-twbs via `(org :variables org-enable-bootstrap-support t)`
- Open an org file
- Press `Ctrl-c Ctrl-e` and there isn't a twb entry

*Observed behaviour:* :eyes: :broken_heart:
twb entry appears in org-export-dispatch.

*Expected behaviour:* :heart: :smile:
twb entry doesn't appear in org-export-dispatch.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 25.3.1
- Spacemacs: 0.200.9
- Spacemacs branch: develop (rev. 804498b4)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
((mu4e :variables mu4e-installation-path "~/share/emacs/site-lisp/mu4e")
 blog-admin ipython-notebook pdf-tools bibtex latex perl5 go rust csv colors command-log javascript dash docker sql nginx vimscript html shell-scripts markdown yaml helm
 (auto-completion :variables auto-completion-return-key-behavior nil auto-completion-tab-key-behavior nil auto-completion-complete-with-key-sequence nil auto-completion-complete-with-key-sequence-delay 0.1 auto-completion-enable-sort-by-usage t auto-completion-private-snippets-directory nil)
 better-defaults
 (org :variables org-enable-bootstrap-support t)
 (ranger :variables ranger-override-dired t)
 emacs-lisp git
 (c-c++ :variables c-c++-default-mode-for-headers 'c++-mode)
 (gtags :variables gtags-enable-by-default nil)
 (python :variables python-test-runner
         '(nose pytest))
 ycmd syntax-checking ab-package)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GCONF GSETTINGS NOTIFY ACL GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11

DWShuo:
where is (org :variables org-enable-bootstrap-support t) placed? i just tried it on my install and it works

amosbird:
in `   dotspacemacs-configuration-layers `

smile13241324:
[...]

PR #9797, opened 2017-10-29
Fix loading issues with org ox extensions as described in #9598

smile13241324:
I have noticed that various ox packages were loaded by calling `spacemacs|use-package-add-hook org ...` but this did not work as intended because of the fact that packages within a layer are called alphabetically. This is causing `org` to load before `ox-...` packages could modify `use-package's post-config hook` to make org's post-config hook to require the respective ox package to load.  

I have modified the init function to load the `ox-...` packages after loading `ox`.

### DIFF
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -370,12 +370,12 @@ Will work on both org-mode and any mode that accepts plain html."
         "it" 'org-agenda-set-tags
         "sr" 'org-agenda-refile)
       (spacemacs|define-transient-state org-agenda
-      :title "Org-agenda transient state"
-      :on-enter (setq which-key-inhibit t)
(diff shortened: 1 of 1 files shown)


smile13241324:
@JAremko can it be that we have an issue with circle.ci since the new release 0.300? I am getting some very strange error messages on my branch. 

Also it seems that tests are run for now unsupported emacs versions lower than 25. Can you have a quick look at this?

JAremko:
Try rebasing @syl20bnr  fixed tests :thinking:

smile13241324:
@d12frosted Sorry to disturb you but I just noticed that this is now waiting for almost a month to be merged. Is there anything I could do to speed things up a bit? 

This is a bug fix for some of the ox-org export packages which are currently not usable from the org export menu. As they are quite popular I hoped to reduce irritations for new comers with this.

PR #9825, opened 2017-11-04
Fix bindings in imenu-list

bet4it:
According to bmag/imenu-list@9cbe6baa55c301ffc21c621ca16d1faea96540aa, we don't need to add binding to quit imenu-list window specifically, and we can add binding to refresh imenu-list window.

According to bmag/imenu-list@8f6538c68b45f678e13d517c6cb82246950aeecb, it is better to toggle imenu-list by `imenu-list-smart-toggle` rather than `imenu-list-minor-mode`.


syl20bnr:
Nice findings, thank you ! 👍 

PR #9828, opened 2017-11-05
improve recognition of react-mode

venikx:
Previously it was impossible for react-mode to recognise .js files, with
additional import statements from 'react'.

Example 1: import React, { Component } from 'react';
Example 2: import React, {
  Component
} from 'react';

The fix is a revert of changes introduced in the following commit:
12f823ba3c18d036ecfa8944775324842c3f3b55

Thank you <3

### DIFF
--- a/layers/+frameworks/react/packages.el
+++ b/layers/+frameworks/react/packages.el
@@ -64,5 +64,5 @@
   (add-to-list 'auto-mode-alist '("\\index.android.js\\'" . react-mode))
   (add-to-list 'auto-mode-alist '("\\index.ios.js\\'" . react-mode))
   (add-to-list 'magic-mode-alist '("/\\*\\* @jsx .*\\*/" . react-mode))
-  (add-to-list 'magic-mode-alist '("import\s+[^\s]+\s+from\s+['\"]react['\"]" . react-mode))
+  (add-to-list 'magic-mode-alist '("import React" . react-mode))
   (add-hook 'react-mode-hook 'spacemacs//setup-react-mode))


AnaRobynn:
First-timer on spacemacs, any idea what these failed builds mean?

Lyonsclay:
Just a naive guess but you have removed the regex that picks up the \from\ the list.

Can you add your line and not delete the line you are replacing?

AnaRobynn:
I'll try that one. Then it would cover both import anything from react and import react.

NerdsvilleCEO:
Agreed, let's keep the old regex in addition to this one, for that exact case of import * as React from react;

Or, another option would be to fix the regex :)

syl20bnr:
Is this PR up to date with last comments ?

AnaRobynn:
No it is not. I haven't found a way to target both cases that I would like to handle. You can reject the PR and I'll come back later if needed.

sdwolfz:
You can just leave the current one as is and add your new one next to it. It does not have to be a regex matching both cases. Leave them me 2 regexes, each for it's case.

AnaRobynn:
Can someone else adjust the PR?
I have recently started writing my own Emacs config and I would like to focus on that.

NerdsvilleCEO:
I can go ahead and do that

Compro-Prasad:
@NerdsvilleCEO Are you still on this? Any reference to a new PR you made?

duianto:
The line that's changed by this PR seems to have been removed by this commit:
[Add: use rjsx-mode for react framework](https://github.com/syl20bnr/spacemacs/commit/8616f07b86d99decf9e00d3beea02ccc74091856)

It might be handled by this line:
https://github.com/syl20bnr/spacemacs/blob/2cfcf54458844f707a99befb3890a390c1e41473/layers/%2Bframeworks/react/packages.el#L64

Therefore we'll close this.

PR #9829, opened 2017-11-05
Ycmd

quicknir:
Two small commits for the ycmd layer. The first removes a warning that is supposed to fire when a variable is unset. In my experience this code runs too early and always fires, even when the variable is set. The second adds a function that can be used to trigger semantic completion manually. This is a prerequisite for the cpp2 layer that I'm planning to push soon. On larger C++ projects semantic completion while acceptable isn't lightning fast, so it makes sense to have dabbrev completion trigger automaticalally and manually trigger semantic. Might be a good opportunity to start thinking about a convention, spacemacs wide, for manually triggering completion, for both evil and holy.


syl20bnr:
Thank you again ! 😄 

PR #9835, opened 2017-11-07
Fix elfeed selection bindings in visual state

jtsymon:
Some elfeed commands can operate on a number of entries, based on the active
selection. When selecting entries with evil (visual state), the bindings for
these commands are shadowed by visual state bindings, and can't be accessed.

To fix this, the affected elfeed bindings are explicitly defined for visual
state.


syl20bnr:
Thank you ! 👍 

PR #9804, opened 2017-10-30
Remove stray characters from output-dir argument

duianto:
The output-dir argument seems to have gotten some stray characters.

---

This was brought to my attention by @sdwolf in the chat:
https://gitter.im/syl20bnr/spacemacs?at=59f59330614889d47513b400

PR #9842, opened 2017-11-08
Fix typo

edwargix:
Fix typo in core/core-configuration-layer.el


duianto:
Thanks for the PR, but this time a PR already has been submitted for this typo:
Remove stray characters from output-dir argument #9804

Issue #9803, opened 2017-10-30
Starting/Updating develop Spacemacs eats memory


MD-ThorErik:
#### Description :octocat:
When I start Spacemacs after changing branches to develop, it hangs while updating and consumes an abnormal amount of memory (at least 2 GB). I even deleted .spacemacs and .emacs.d.
Opening Spacemacs from the master branch functions as expected.

#### Reproduction guide :beetle:
- Start Emacs

*Observed behaviour:* :eyes: :broken_heart:
The statusline shows:
Downloading stable ELPA repository: spacelpa... (please wait)
uncompressing publicsuffix.txt.gz...done
And spacemacs hangs, consuming memory.

*Expected behaviour:* :heart: :smile:
Spacemacs starts and updates normally.

System Info 💻
    OS: Arch Linux
    Emacs: 25.3.1
    Spacemacs: (0.300.1@25.3.1)
    Spacemacs branch: develop (d15772fadcb86937522c49cb0819fd36ba5184b6)
    Graphic display: t
    Distribution: spacemacs
    Editing style: vim
    Completion: helm
    Layers: Even with the default layers in .spacemacs happens

**Edit: Content of Messages buffer after runnining emacs with "ulimit -v 1200000; exec emacs --debug-init"**

> Loading /home/me/.emacs.d/core/core-load-paths.el (source)...done
Loading /home/me/.emacs.d/.lock...done
Loading /home/me/.spacemacs...done
(Spacemacs) Calling dotfile init...
(Spacemacs) Calling dotfile user init...
Setting the font...
(Spacemacs) Setting font "Source Code Pro"...
(Spacemacs) Banner: /home/me/.emacs.d/core/banners/img/spacemacs.png
PNG warning: Interlace handling should be turned on when using png_read_image
Open the quickhelp.
Contacting host: github.com:443
uncompressing publicsuffix.txt.gz...done
Memory exhausted--use M-x save-some-buffers then exit and restart Emacs
error in process filter: Memory exhausted--use M-x save-some-buffers then exit and restart Emacs [514 times]
error in process filter: apply: Variable binding depth exceeds max-specpdl-size
error in process filter: Variable binding depth exceeds max-specpdl-size
Variable binding depth exceeds max-specpdl-size
error in process filter: Memory exhausted--use M-x save-some-buffers then exit and restart Emacs [514 times]
error in process filter: apply: Variable binding depth exceeds max-specpdl-size
[...]

PR #9859, opened 2017-11-12
Fix excessive memory usage on first startup

jjlee:
I experienced this reproducibly when spacemacs reached the stage where it shows
the message in the mode line starting like this:

"Downloading stable ELPA repository: "

Without this change I was unable to start spacemacs and in fact had to reboot my
machine multiple times. Emacs 25.1.1, Ubuntu 17.04, no swap configured.

Obviously this needs cleaning up a bit...

Also fixed an extraneous 'kkj' which I assume is a typo (actually I thought this
was my typo but turns out to be syl20bnr himself :-)

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -2372,9 +2372,9 @@ The URL of the descriptor is patched to be the passed URL"
       (error "Opening output file: File already exists, %s" newname))
   (cond
    ((executable-find "wget")
-    (shell-command (concat "wget -O " (shell-quote-argument newname) (shell-quote-argument url))))
+    (shell-command (concat "wget -O " (shell-quote-argument newname) " " (shell-quote-argument url))))
    ((executable-find "curl")
-    (shell-command (concat "curl -o " (shell-quote-argument newname) (shell-quote-argument url))))
+    (shell-command (concat "curl -o " (shell-quote-argument newname) " " (shell-quote-argument url))))
    ;; (t (url-copy-file url newname 'ok-if-already-exists))))
    (t (message (concat "Neither curl nor wget found found; please install")))))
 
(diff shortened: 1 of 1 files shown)


jjlee:
Maybe related to #9803?

jjlee:
Er, to clarify, the excessive memory usage occurred when starting spacemacs for the first time after git cloning.

jjlee:
This fixed the problem for me with wget installed, but this is intended mostly to show the problem and a working solution just to illustrate what the problem is:

* It has not been carefully tested (I didn't even try the curl case for example, nor the fallback "you need to install curl/wget" case)
* I don't know if this works on Windows / Mac
* There's the question of what to do if no external url fetching program is available (curl, wget, ...), hence the commented-out code here
* I imagine my coding style is not up to scratch

Seems bad to fall back to the emacs native `url-copy-file`: I guess I think that having the potential for this bad behaviour, even for only some people, is worse than failing entirely because wget or curl is not installed.  My guess is people won't get as angry if they just have to install curl or wget as if their computer hangs and has to be rebooted :-)  I guess spacemacs could just not fetch the big elpa tarball archive if no external download tool is available, and rely on lazily fetching packages one-by-one in that case?

(I'm not angry :-)

NerdsvilleCEO:
@jjlee Nice catches, I am wondering if it makes sense to just give a message saying it couldn't download all of the elpa dependencies and asking to download cURL or wget if they aren't present... but this does indeed beg the question, how would this be handled in Windows... maybe this is why the fallback was there, I am unsure as I am myself new to spacemacs

jjlee:
Alternative possible hack: split the tarball into 10 parts (may not work: it's just my guess that might fix the most extreme memory usage).  Then, fall back to emacs-native url-copy-file if curl/wget isn't available.

PR #9861, opened 2017-11-12
document how to use a local version of a package

deb0ch:
I've had friction with this many times, even though I already had it working in the past.

Now I just got it working again thanks to TheBB helping me. I think it would be useful for other people who want to start patching or contributing to packages, to prevent the same "where the heck is this documented ??? - Oh, it's not..." kind of friction that I had.

### DIFF
--- a/core/templates/.spacemacs.template
+++ b/core/templates/.spacemacs.template
@@ -54,6 +54,9 @@ This function should only modify configuration layer settings."
    ;; wrapped in a layer. If you need some configuration for these
    ;; packages, then consider creating a layer. You can also put the
    ;; configuration in `dotspacemacs/user-config'.
+   ;; To use a local version of a package, use the `:location' property:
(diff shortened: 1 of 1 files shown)


timor:
cc @spacefrogg

spacefrogg:
This approach does not work for all packages. Especially themes seem to be loaded at a very early stage in the loading process. To get self-written themes to work, I had to use the following expression in the themes' `packages.el`:
```
(setq your-theme-packages
      '((your-theme :location (recipe :fetcher local))))
...
```
The consequence is, that packages defined like that are copied to the local elpa store, like packages from the Internet. I think, that dependencies are resolved, but haven't tested it.

kRITZCREEK:
Thank you, I'd maybe add that you need to explicitly require the added package, as the layer that usually loads it won't.

syl20bnr:
Thank you !

PR #9862, opened 2017-11-12
.spacemas.template: add newline after each configuration variable

deb0ch:
This allows to:

- navigate between the variables conveniently, using the '{' and '}' paragraph movements
- make it clearer wether description comment is for the variable below or above it

I have been using it like this ever since I started using Spacemacs and always use the paragraph movement keys to navigate between configuration variables.


wilhelmy:
Merry Spacemas :)

syl20bnr:
WOW what a good idea, never thought about that 👍 
Thank you !

PR #9863, opened 2017-11-13
Use python-mode for SCons script files

shanemikel:
The SCons build system uses Python build scripts conventionally named:
* SConstruct
* SConscript

This commit would make python-mode load by default for these files.


syl20bnr:
Thank you ! 👍 

Issue #9826, opened 2017-11-04
SPC-f-f (spacemacs/helm-find-files) gives error if invoked too quickly after startup


jackkamm:
#### Description :octocat:

SPC-f-f (spacemacs/helm-find-files) gives an error message if invoked
immediately on startup (within 1 second).

This error can be fixed by adding `helm-current-directory`
to the autoloaded commands when starting up helm.
I can submit a PR for this but wanted to first check whether
this is the appropriate solution, and if anyone else
experiences this error.

#### Reproduction guide :beetle:
- Start Emacs
- Immediately (within 1 second of startup) type "SPC f f"

*Observed behaviour:* :eyes: :broken_heart:
helm-find-files doesn't start, and gives an error message
```
expand-file-name: Symbol’s function definition is void: helm-current-directory
```

*Expected behaviour:* :heart: :smile:
helm-find-files starts correctly

#### System Info :computer:
- OS: gnu/linux
- Emacs: 25.3.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. a363f85d)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: helm
- Layers:
```elisp
(helm major-modes html csv yaml javascript markdown ranger bibtex latex octave
      (ess :variables ess-enable-smart-equals t)
      ipython-notebook python shell-scripts
      (shell :variables shell-default-shell 'multi-term)
      org emacs-lisp auto-completion git my-spacemacs-layer)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GCONF GSETTINGS NOTIFY ACL GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 MODULES

tavurth:
I have the same issue, it's also reproducible if you run `SPC-p-f` Projectile find-file.

tavurth:
I was able to fix this issue by putting:
```lisp
  ;; Fixes undefined functions at startup 
  (require 'helm)
  (require 'tramp)
```
In the `dotspacemacs/user-config` area.

mopsfelder:
@syl20bnr, I see this was labeled as 'Fixed in develop', but using branch develop at 7c9f612c4f4a1eda7c8a11194ff2bce21ca53ca0, the issue is still present.

[...]

PR #9869, opened 2017-11-13
fix autoloading spacemacs/helm-find-files #9826

jackkamm:
This is a proposed fix for https://github.com/syl20bnr/spacemacs/issues/9826

Despite `spacemacs/helm-find-files` being passed to `:commands`, use-package doesn't seem to autoload helm when it is called, causing the error. I think use-package needs the `:commands` values to actually be part of helm? Not an expert on use-package/autoloading.


syl20bnr:
Thank you ! 👍 

Issue #5483, opened 2016-03-14
How to set indent in xml to 4?


fuzihaofzh:
I have set tab equal to 4 spaces. But when I use xml and lua mode, it turn out to set the tab as 2 spaces. How can I solve that?

darfink:
In most cases each major-mode has mode-specific options for adjusting indentation:

1. [Web-mode](http://web-mode.org/) `(setq web-mode-markup-indent-offset 4)`
2. [Lua-mode](https://github.com/immerrr/lua-mode#customization): `(setq lua-indent-level 4)`

maplewizard:
Hi, @darfink, I tried as you said, and the user-config is as follows:
```
(defun dotspacemacs/user-config ()
  "Configuration function for user code.
 This function is called at the very end of Spacemacs initialization after
layers configuration. You are free to put any user code."
  ;;(xterm-mouse-mode 0)
  (global-hl-line-mode 0)
  ;;tab
  (setq-default tab-width 4)
  (setq web-mode-markup-indent-offset 4)
  (setq lua-indent-level 4)
  (setq nxml-child-indent 4 nxml-attribute-indent 4)
  (setq-default c-basic-offset 4)
  ;;neotree
  (global-set-key (kbd "<f9>") 'neotree-toggle)
  (global-set-key (kbd "C-j") 'avy-goto-word-or-subword-1)
  (spacemacs/toggle-automatic-symbol-highlight)
  )
```
But it still cannot work. How could that happen?

darfink:
Have you configured a custom major-mode for XML files or are you using the default? (`nxml-mode`).
Using a complete vanilla version of Spacemacs I achieve the desired behavior by using:
```
(defun dotspacemacs/user-config ()
  (setq nxml-child-indent 4))
```

TheBB:
Use `setq-default`, not `setq`. Then confirm that in the relevant buffer(s), the variables actually have the values that you expect.

maplewizard:
@TheBB @darfink 
 This is my layer config and it seems still not work. 
```
dotspacemacs-configuration-layers
   '(
     ;; ----------------------------------------------------------------
     ;; Example of useful layers you may want to use right away.
     ;; Uncomment some layer names and press <SPC f e R> (Vim style) or
     ;; <M-m f e R> (Emacs style) to install them.
     ;; ----------------------------------------------------------------
     osx
     python
     ipython
     c-c++
     markdown
     auto-completion
     better-defaults
[...]

PR #9878, opened 2017-11-16
html layer: use web-mode for `.xml` files

emanresusername:
open `xml` files in `web-mode` by default in `html` layer

that's the behavior i'd expect/prefer and [i'm not <sup>totally</sup> alone](https://github.com/syl20bnr/spacemacs/issues/5483#issuecomment-196277122) @TheBB


syl20bnr:
Thank you !

PR #9886, opened 2017-11-18
Add Troubleshooting section

NerdsvilleCEO:
This adds some basic troubleshooting for the php layer hopefully to save some other people a 6 month timesink like I had! :)

### DIFF
--- a/layers/+lang/php/README.org
+++ b/layers/+lang/php/README.org
@@ -37,7 +37,7 @@ file.
 
 * Troubleshooting
 Because of the way that the php-extras package works, there are a couple of simple
-initialization tasks which must occur to get the completion working as should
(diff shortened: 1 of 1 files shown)


layers/+lang/php/README.org
```diff
@@ -46,4 +46,4 @@ On any new project make sure to perform the following initialization tasks:
     touch .ac-php-conf.json
   #+END_SRC
 2. Inside of spacemacs run:
-  ac-php-remake-tags-all
+  = ac-php-remake-tags-all =
```
duianto: If the spaces inside the `=` signs are removed, then the `=` signs won't be visible. The same way as
`=dotspacemacs-configuration-layers=` is written in the Install section.

syl20bnr:
Thank you ! 👍 

PR #9897, opened 2017-11-22
Update README of Spotify layer

sunghyunzz:
As "searching for a new track" feature is only provided by `helm-spotify`, I added a description about Helm only key bindings. It would be less confusing for non Helm layer configured users.


sunghyunzz:
A formatting test is broken but it seems like original README file was not formatted properly. Please let me know if it would be better me to fix it.

syl20bnr:
Thank you ! 👍 

PR #9899, opened 2017-11-23
use evil-smartparens-mode for smartparens-strict mode

timor:
This is more dwim in the sense that the normal deleting commands now behave more
like paredit, i.e. region deletion is adjusted to always keep the pairs
balanced.

### DIFF
--- a/layers/+spacemacs/spacemacs-editing/packages.el
+++ b/layers/+spacemacs/spacemacs-editing/packages.el
@@ -25,6 +25,7 @@
         (origami :toggle (eq 'origami dotspacemacs-folding-method))
         password-generator
         smartparens
+        (evil-smartparens :toggle dotspacemacs-smartparens-strict-mode)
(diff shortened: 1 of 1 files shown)


syl20bnr:
Thank you ! 👍 

As we already had the `evil-cleverparens` layer and I decided to move it out of its own layer and use it instead of `evil-smartparens`. Read the commit message of commit https://github.com/syl20bnr/spacemacs/commit/2cb98447 for more information about this change.

If you can test wether `evil-cleverparens` does the job or not it would be fantastic. The idea is to have support for only one package for evil safe structural editing, we will support the best of them. For now we use the legacy `evil-cleverparens` but we can move to another one if it is shown to be better.

timor:
@syl20bnr I did not know about the cleverparens layer, thanks for working this in.  However, I really think this should be tightly coupled to smartparens strict mode if the whole idea of its strict mode is to prevent unbalancing.

PR #9910, opened 2017-11-24
fix typo in cmake-ide by moving a closing parenthesis

emacs18:
One closing parenthesis for use-package call is mis-placed.
The fix is trivial.  Just move it to the end of the sexp.

Without this fix, I get "cmake-ide-setup undefined" error.


myrgy:
thanks for that fix!

syl20bnr:
Thank you ! 👍 

PR #9911, opened 2017-11-24
core: add reference handler as per jump handler

MaskRay:
`xref-find-references` `rtags-find-references-at-point` `helm-gtags-find-rtag` can be potential customers of this feature.

"Definition handler" might be a more appropriate name for "jump handler".

If this PR is accepted, various language configuration files `layers/+lang/*/config.el` needs to include `(spacemacs|define-reference-handlers c++-mode)`

This function may need to added to `xref-prompt-for-identifier`. See https://github.com/emacs-lsp/lsp-mode/issues/194


MaskRay:
Ping ...

duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/c825035e48c047f907968813a4bf57cace05ac21

PR #9920, opened 2017-11-27
fix julia keybindings and nice prefixes

gdkrmr:
Issues I would like to resolve:

- [x] avoid `eval` not sure why `apply` is not working
- [x] `spacemacs/ess-start-repl` does not work for julia


gdkrmr:
@bmag is this stalled because of the unmarked checkbox?

bmag:
@gdkrmr not stalled, I just didn't see the PR. I'll see if I can find a way to avoid the `eval`, but it will be at least several days until I have time to do that. Maybe a `dolist` could work?

gdkrmr:
@bmag this is not performance critical, so I think that `eval` is ok here if it requires a weird hack to avoid. I tried `apply`, see [here](https://github.com/syl20bnr/spacemacs/pull/9920/files#diff-4f8089419795841827127fdb8303d489R121), but it did not work.

sdwolf:
I suggest using a `dolist` like here: https://github.com/syl20bnr/spacemacs/blob/249efa31dec4e6820199d698848ecbe7ff3f4e1f/layers/%2Blang/markdown/packages.el#L71

gdkrmr:
dolist works great!

bmag:

I moved the declare-prefixes into the dolist as well, and did some moving/squashing of the commits before cherry-picking them into develop. See https://github.com/syl20bnr/spacemacs/commit/99d575b9c822a2ded7b8aa2848f44a334abb7041, https://github.com/syl20bnr/spacemacs/commit/88a952fb3357f755687d8d6e94ae0b6575579605 and https://github.com/syl20bnr/spacemacs/commit/502a62a640e3f869feed8228c0fbb720190511fc.

FYI, when updating a PR with changes from the develop branch, you should use `git rebase` or `git pull --rebase` rather than `git merge`.

gdkrmr:
ok, thx

PR #9921, opened 2017-11-27
Make global-semantic-stickyfunc-mode configureable

rasztasd:
Hello!

I am using emacs's header line functionality for showing tabs (of buffers in project) and not for the semantic stickyfunc mode.

I want the global semantic sticky func to be configurable to show or to hide, because I want to use it for a different purpose (not for semtantic sticky func).

Thanks for any input.
Br, Daniel

### DIFF
--- a/layers/+emacs/semantic/config.el
+++ b/layers/+emacs/semantic/config.el
@@ -13,6 +13,9 @@
                                     "srecode-map.el"))
 (setq semanticdb-default-save-directory (concat spacemacs-cache-directory
                                                 "semanticdb/"))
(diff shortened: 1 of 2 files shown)


syl20bnr:
Thank you for the PR! 💜 
There is a built-in way in the layer system to do this, in your dotfile where you added `semantic` layer, replace it with a list and the `:packages`  keyword like this:

```elisp
(semantic :packages (not stickyfunc-enhance))
```

I pushed a fix in develop for this to work properly with `stickyfunc-enhance`.

Since it is already supported I close this PR.

PR #9924, opened 2017-11-28
Add keybinding for Rust cargo check

alienscience:
Cargo check is a faster alternative to compiling when
developing Rust code that checks syntax and types without
producing a binary. This commit adds a keybinding for Cargo check.

### DIFF
--- a/layers/+lang/rust/packages.el
+++ b/layers/+lang/rust/packages.el
@@ -41,6 +41,7 @@
         "cf" 'cargo-process-fmt
         "ci" 'cargo-process-init
         "cl" 'cargo-process-clippy
+        "ck" 'cargo-process-check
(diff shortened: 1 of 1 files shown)


vityafx:
Well, this is almost the same PR as [mine](#9944), but I also remap `cargo-process-build` to a proper key so there is much better use of `cargo-process-check` because it is mapped to `, c c` instead of `, c k` as in this PR. @alienscience what do you think?

alienscience:
Personally, I would be against changing the existing keys. If someone changed the keybinding for compile, I would be annoyed until I got the new version into my muscle memory. However, I don't feel strongly about it.

@vityafx if you want to go with your PR that is fine by me.

PR #9927, opened 2017-11-28
add notmuch layer

### LINKED ISSUE #2163: Request to add notmuch
Can we add a new layer for notmuch ?

jb55:
This patch adds a notmuch[1] layer to Spacemacs!

notmuch is a mail indexing system written in C. It provides an emacs
interface, and this patch adds Spacemacs support for this interface.

Thanks to @FrancescElies for starting this work, and working with me to
get it finished. We have been using the layer for awhile now and are
happy with the initial release!

Closes #2163

[1] https://notmuchmail.org/

Tested-by: Francesc Elies <elies@posteo.net>
Tested-by: William Casarin <jb55@jb55.com>
Signed-off-by: William Casarin <jb55@jb55.com>


FrancescElies:
@jb55 Thanks to you, for making this happen 👍

kalbasit:
@jb55 thank you for working on this, it's awesome to finally see it here. Do you know if it's possible to change the keyboard layout? I primarily use [Colemak](https://github.com/kalbasit/vim-colemak).

jb55:
Hmm I'm not sure what the precedent is for changing the keybindings for
users of different keyboard layouts. I'm not sure how to support this
use case...

kalbasit:
@jb55 I'm not really sure if spacemacs own keybindings (so Vim's standard bindings) are changeable, but if they are, I suggest declaring the keybindings following the same way it's done; For instance, https://github.com/syl20bnr/spacemacs/pull/9927/files#diff-8e21960feb14d71085ffc27f77eac1c0R46 replacing `"a"` with a variable, will allow me to change it, `a` is for entering visual mode for me.
In any case, we shouldn't block this PR because of this, I'll have to do my own homework on the feasibility of Colemak bindings for spacemacs itself, and move forward from there.

smile13241324:
@kalbasit I think there is a separate layer for changing the key layout used by spacemacs here http://spacemacs.org/layers/+intl/keyboard-layout/README.html.

Not sure it supports colemag but you should have a look.

kalbasit:
@smile13241324 oh awesome, thank you for the link. I'll check if it has Colemak, I'll add it if not. Mine is a bit more custom than what I have seen done with Colemak layout, I move the `hjkl` one key to the right to align more with the keyboard actual home row since I'm not bound to the `;` problem (of course for me it's `neio` :)).

jb55:
Updated to fix the remaining test issues 👍

fuxialexander:
You might be interested in my new [`counsel-notmuch`](https://github.com/fuxialexander/counsel-notmuch) package, which is inspired by `helm-notmuch`. I've use it for quite a while and currently try to submit it to MELPA.

jb55:
@fuxialexander oooo nice, I will try this out.

jackkamm:
When I try to call `notmuch-help` (by pressing `?`) I get an error message,
```
edmacro-format-keys: Wrong type argument: listp, 126
```
Does anyone else experience this? It seems to have something to do with the `evilified-state-evilify-map`, as when I comment these out it works.

jb55:
@jackkamm yeah it's a known issue, I haven't figured it out yet. If anyone has any ideas...

timor:
cc @spacefrogg

timor:
Nice!  Is there a reason not to pick the simpler a n prefix instead of a N?

FrancescElies:
> Nice! Is there a reason not to pick the simpler a n prefix instead of a N?

The [deft](https://github.com/syl20bnr/spacemacs/tree/master/layers/%2Btools/deft) layer is already using that prefix, but I would also prefer to use `n` prefix :)

syl20bnr:
Thank you ! 💜 
I let the shortcut as this `SPC a N` as `SPC a n` is already taken and I have a plan to better support applications in Spacemacs which will fix the key bindings collision (and mess :-)) we have under `SPC a`.

aijony:
@jb55 @jackkamm 
Regarding the `notmuch-help` issue, I have a "solution" in my personal config that I made a couple of months ago. The problem with evilifying map is that it grunks up the format that `notmuch-help`.
However, if the the remapping is done manually, it keeps the format nice and clean.
I think a problem with my solution is that it doesn't play well with holy-mode users.
A possibly more elegant hack would be to overwrite `notmuch-help` to play nicer with evil,
and that might even fit more into a pull request to the notmuch repo.

Anyways, here is a [gist](https://gist.github.com/aijony/74b8dd54eba739203884fb4879a765de) with my code. This fix only works if the notmuch layer's keybindings are commented out. I just updated the code to add the new keybindings in this package as well.

There was also an issue involving the text boxes in `notmuch-hello` and evil, so there is also a very un-elegant fix to that issue at the bottem.

jb55:
@aijony thanks! I put together an issue for this here: #10123

guillaumecherel:
Thanks for all the work! I didn't see the layer in the [layers list](http://spacemacs.org/layers/LAYERS.html#e-mail). Is it not ready to appear in the doc yet?

jb55:
> Thanks for all the work! I didn't see the layer in the [layers list](http://spacemacs.org/layers/LAYERS.html#e-mail). Is it not ready to appear in the doc yet?

Looks like an oversight, feel free to make a PR!

PR #9934, opened 2017-11-30
Write file with SPC f w.

dschrempf:
Add a spacemacs shortcut to save files with a different name (`write-file`).

### DIFF
--- a/layers/+distributions/spacemacs-base/keybindings.el
+++ b/layers/+distributions/spacemacs-base/keybindings.el
@@ -215,6 +215,7 @@
   "fvd" 'add-dir-local-variable
   "fvf" 'add-file-local-variable
   "fvp" 'add-file-local-variable-prop-line
+  "fw" 'write-file
   "fy" 'spacemacs/show-and-copy-buffer-filename)
 ;; frame ----------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


nixmaniack:
This isn't needed since we already have `write-file` wrapper under <kbd>SPC f c</kbd>.

dschrempf:
That's right. However, as far as I know the buffer is still pointing to the old destination then? So the file has to be reopened. Please correct me, if this is wrong! Thanks.

CeleritasCelery:
The buffer is pointing to the new file after `SPC f c`

dschrempf:
Ah OK, thank you. It's interesting that there are two functions that do apparently the same thing.

nixmaniack:
I don't know the exact reasoning of [writing a wrapper on `write-file`](https://github.com/syl20bnr/spacemacs/blob/b4b7d63e41ace986b05ee31c344e7da2f3db5b87/layers/%2Bdistributions/spacemacs-base/funcs.el#L918-L921) but it certainly helps to have a descriptive name in the keybindings pop-up. `copy-file` describes the purpose more prominently than `write-file`.

You can close the issue if you are satisfied with the current binding.

PR #9938, opened 2017-11-30
Fix documentation for speed-reading, restclient and pandoc layer

smile13241324:
Fixed missing features block for speed-reading, restclient and pandoc layer.
This is part of #9476.


JAremko:

PR #9944, opened 2017-12-02
Add shortcut for the `cargo check` command #9943

### LINKED ISSUE #9943: Add a shortcut for `cargo check` command in the `rust` layer.
There is a very useful command in the cargo  called `check`, which does not compile anything but simply checks the project for being able to compile at all. Can we map it to a shortcut? It would also be good in my opinion to move `cargo-process-build` to a `cb` shortcut.

```el
"cc"  'cargo-process-check
"cb"  'cargo-process-build
```

vityafx:
1. Adds a shortcut for the `cargo check` command as "cc".
2. Remaps a shortcut for `cargo build` command from "cc" to "cb".

Fixes #9943

### DIFF
--- a/layers/+lang/rust/packages.el
+++ b/layers/+lang/rust/packages.el
@@ -33,7 +33,8 @@
         "c." 'cargo-process-repeat
         "cC" 'cargo-process-clean
         "cX" 'cargo-process-run-example
-        "cc" 'cargo-process-build
+        "cc" 'cargo-process-check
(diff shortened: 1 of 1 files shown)


syl20bnr:
Thank you ! 👍 

As `SPC m c c` is part of the convention I moved the check to `SPC m c v` , see commit https://github.com/syl20bnr/spacemacs/commit/af625a94

PR #9949, opened 2017-12-04
Add evil-avy-goto-char-timer to leader jt

duianto:


syl20bnr:
Thank you ! 💜 

Seems so good that I moved it on `SPC j j` to replace the one char and two chars functions (removed `SPC j J`). Commit: https://github.com/syl20bnr/spacemacs/commit/4c231ae2

PR #9953, opened 2017-12-04
Support importmagic in the Python layer

signalpillar:
Problem
-------
If the symbol is not resolved due to a missing import there is no automatic way
in the Python layer to import it.

Solution
--------
Use 'importmagic' package.

Pressing `SPC m r f` at point that requires an import will suggest a list of
possible options.


signalpillar:
If virtual environment is used this solution requires installing of the python libs required in the venv itself - not always handy. I think we need in the future add this to anaconda server as there is no need to run another RPC server.

CeleritasCelery:
`importmagic` should be autoloaded (don’t use `:ensure t`)

syl20bnr:
Thank you ! 👍 

signalpillar:
Thank you

lukeorland:
It seems like this code addition might freeze Emacs startup: #10145

signalpillar:
@lukeorland look like it does. Can we roll it back, the feature I've added to the python layer?

PR #9960, opened 2017-12-05
Add go-tag for go-layer

brantou:
[go-tag](https://github.com/brantou/emacs-go-tag) is used to edit  field tags for golang struct fields，so I added it to the **refactoring** category.


syl20bnr:
Thank you ! 👍 

PR #9968, opened 2017-12-07
Fix OPAM share directory in OCaml layer

leviroth:
This commit uses process-lines instead of shell-command-to-string, as seen in ocaml/merlin#547.

This fixes an issue on Windows where the external command fails, resulting in a broken load-path.

### DIFF
--- a/layers/+lang/ocaml/funcs.el
+++ b/layers/+lang/ocaml/funcs.el
@@ -11,14 +11,22 @@
 
 (defun spacemacs//init-ocaml-opam ()
   (if (executable-find "opam")
-      (let* ((output (shell-command-to-string
-                      "opam config var share 2> /dev/null"))
-             (share (when (< 0 (length output))
-                      (substring output 0 -1))))
-          (when share
-            (setq opam-share share
-                  opam-load-path (concat share "/emacs/site-lisp")))
-          (add-to-list 'load-path opam-load-path))
+      (let ((share (string-trim-right
+                    (with-output-to-string
+                      (with-current-buffer
+                          standard-output
+                        (process-file
+                         shell-file-name nil '(t nil) nil shell-command-switch
+                         "opam config var share"))))))
+        (cond ((string= "" share)
+               (spacemacs-buffer/warning
+                "\"opam config var shar\" output empty string."))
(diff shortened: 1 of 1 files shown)


rgrinberg:
What happens when this command actually outputs something into stderr?

rgrinberg:
OK just tried it out and it doesn't seem so good.

```
(process-lines "bash" "-c" "echo 'stdout'; (>&2 echo 'stderr')")
;; gives
("stdout" "stderr")
```

leviroth:
Are you suggesting that I'm leaving out something essential from merlin's recommended config, or that the recommended config could be improved? Or something else I'm not understanding?

rgrinberg:
All Im saying that `opam config var share` will sometimes output stuff into stderr. So in the case where stderr does happen to print first, your code for obtaining the share path will fail. We need a way to ignore the stderr or only read the stdout to be on the safe side.

leviroth:
Right, I guess I was asking if you thought the `ignore-errors` from the merlin guide would do the trick (I'm not very experienced with elisp). I believe the change I just pushed will accomplish what you're looking for; let me know.

syl20bnr:
@rgrinberg is it OK to merge this ?

rgrinberg:
I don't think ignore-errors does anything here. The problem is that stderr needs to be handled correctly, and the old behavior relied on the shell to do that. Let me roll up my elisp sleeves and see what I can do.

rgrinberg:
OK here's how you can run the command and only get the stdout as output:

```
(string-trim-right
 (with-output-to-string
   (with-current-buffer
       standard-output
     (process-file shell-file-name nil '(t nil) nil shell-command-switch "opam config var share"))))
```

Hopefully this works well on windows as well, and we should also verify that the output we get is valid (non blank, path exists), and warn the user appropriately.

In an ideal world, we'd also print the stderr to messages or something. But this is likely good enough.

syl20bnr:
Thank you @rgrinberg.
@leviroth can you make the modification ?

leviroth:
Pushed.

syl20bnr:
Thank you ! 👍 

PR #10002, opened 2017-12-15
Add helpful layer

jdenen:
Adds layer for the [helpful](https://github.com/Wilfred/helpful) package.

### DIFF
--- a/layers/+emacs/helpful/README.org
+++ b/layers/+emacs/helpful/README.org
@@ -1,0 +1,26 @@
+#+TITLE: helpful layer
+
+* Table of Contents                                        :TOC_4_gh:noexport:
+- [[#description][Description]]
+  - [[#features][Features:]]
+- [[#install][Install[]]
+- [[#key-bindings][Key bindings]]
+
+* Description
+  This layer supports the [[https://github.com/Wilfred/helpful][helpful]] package, which extends the Emacs help buffer to include things like references and source code.
+** Features:
+   Better help buffers that include source code, callers, and prettier docstrings.
+
+* Install[
+To use this configuration layer, add it to your =~/.spacemacs=. You will need to
+add =helpful= to the existing =dotspacemacs-configuration-layers= list in this
+file.
+
+* Key bindings
+
+| Key Binding | Description                              |
+|-------------+------------------------------------------|
+| =SPC h h .= | Open helpful buffer for item under point |
+| =SPC h h f= | Open helpful buffer for function         |
+| =SPC h h k= | Open helpful buffer for key binding      |
+| =SPC h h v= | Open helpful buffer for variable         |
--- a/layers/+emacs/helpful/README.org
+++ b/layers/+emacs/helpful/README.org
@@ -3,15 +3,19 @@
 * Table of Contents                                        :TOC_4_gh:noexport:
 - [[#description][Description]]
   - [[#features][Features:]]
-- [[#install][Install[]]
+- [[#install][Install]]
 - [[#key-bindings][Key bindings]]
 
 * Description
   This layer supports the [[https://github.com/Wilfred/helpful][helpful]] package, which extends the Emacs help buffer to include things like references and source code.
-** Features:
-   Better help buffers that include source code, callers, and prettier docstrings.
 
-* Install[
+** Features:
+   - Better help buffers with [[https://github.com/Wilfred/helpful][helpful]]
+     - Source code
+     - Callers
+     - Prettier docstrings
+
+* Install
 To use this configuration layer, add it to your =~/.spacemacs=. You will need to
 add =helpful= to the existing =dotspacemacs-configuration-layers= list in this
 file.
--- a/layers/+emacs/helpful/packages.el
+++ b/layers/+emacs/helpful/packages.el
@@ -1,0 +1,31 @@
+;;; packages.el --- helpful layer packages file for Spacemacs.
+;;
+;; Copyright (c) 2012-2017 Sylvain Benner & Contributors
+;;
+;; Author: Johnson Denen <johnson@johnsons-macbook-pro.local>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+(defconst helpful-packages
+  '(helpful))
+
+(defun helpful/init-helpful ()
+  (use-package helpful
+    :defer t
+    :init
+    (progn
+      (global-set-key (kbd "C-h f") #'helpful-callable)
+      (global-set-key (kbd "C-h k") #'helpful-key)
+      (global-set-key (kbd "C-h v") #'helpful-variable)
+      (spacemacs/declare-prefix "hh" "helpful")
+      (spacemacs/set-leader-keys
+        "hh." #'helpful-at-point
+        "hhc" #'helpful-command
+        "hhf" #'helpful-callable
+        "hhk" #'helpful-key
+        "hhm" #'helpful-macro
+        "hhv" #'helpful-variable))))
+
+;;; packages.el ends here
--- a/layers/+emacs/helpful/packages.el
+++ b/layers/+emacs/helpful/packages.el
@@ -16,12 +16,9 @@
     :defer t
     :init
     (progn
-      (global-set-key (kbd "C-h f") #'helpful-callable)
-      (global-set-key (kbd "C-h k") #'helpful-key)
-      (global-set-key (kbd "C-h v") #'helpful-variable)
       (spacemacs/declare-prefix "hh" "helpful")
       (spacemacs/set-leader-keys
-        "hh." #'helpful-at-point
+        "hhh" #'helpful-at-point
         "hhc" #'helpful-command
         "hhf" #'helpful-callable
         "hhk" #'helpful-key


nixmaniack:
I recently started using this package and thought this could be a default functionality of spacemacs. If you can add some screenshots of the default help buffer and helpful help buffer that would be helpful to gather feedback from users on this thread.

layers/+emacs/helpful/packages.el
```diff
@@ -0,0 +1,50 @@
...
+  '(helpful))
+
+(defun helpful/init-helpful ()
+  (use-package helpful
+    :defer t
+    :init
+    (progn
+      (global-set-key (kbd "C-h f") #'helpful-callable)
+      (global-set-key (kbd "C-h k") #'helpful-key)
+      (global-set-key (kbd "C-h v") #'helpful-variable)
```
nixmaniack: Since this a layer, I don't think we should replace existing default help bindings because helpful bindings are provided as extra helpful functionality under  `SPC h h <key>`.  Just my opinion though.

layers/+emacs/helpful/packages.el
```diff
@@ -0,0 +1,50 @@
...
+      (spacemacs/declare-prefix "hh" "helpful")
+      (spacemacs/set-leader-keys
+        "hh." #'helpful-at-point
+        "hhf" #'helpful-callable
+        "hhk" #'helpful-key
+        "hhv" #'helpful-variable))))
```
nixmaniack: May as well add `helpful-macro` and `helpful-command` under `m` and `c`.

layers/+emacs/helpful/packages.el
```diff
@@ -0,0 +1,50 @@
...
+;; Briefly, each package to be installed or configured by this layer should be
+;; added to `helpful-packages'. Then, for each package PACKAGE:
+;;
+;; - If PACKAGE is not referenced by any other Spacemacs layer, define a
+;;   function `helpful/init-PACKAGE' to load and initialize the package.
+
+;; - Otherwise, PACKAGE is already referenced by another Spacemacs layer, so
+;;   define the functions `helpful/pre-init-PACKAGE' and/or
+;;   `helpful/post-init-PACKAGE' to customize the package as it is loaded.
+
```
nixmaniack: Above helpful message regarding layer usage should be removed.

layers/+emacs/helpful/packages.el
jdenen: I'd like to hear more feedback regarding this. I debated it myself, but I ultimately concluded that if you use the `helpful` layer, you probably want it to replace the default help buffers.

If you would want the default help buffers, they'll still exist under `SPC h d`. But I'm open to hear more opinions.

layers/+emacs/helpful/packages.el
jdenen: Thanks for this catch.

layers/+emacs/helpful/packages.el
CeleritasCelery: These keybidings should be under the Spacemacs leader key. That is how they are “supposed” to be accessed. Don’t mess with default bindings (especially since this is not how evil users would access it). I like @nixmaniack suggestion of using `SPC h h`, but if you feel like helpful should override the defaults they should be bound to `SPC h d`

nixmaniack:
I have recently also added following config locally.
```elisp
(evilified-state-evilify helpful-mode helpful-mode-map)
```
and for having helpful buffers popup at bottom instead of right side(opinionated change).
```elisp
(push '(helpful-mode :dedicated t :position bottom :stick t :noselect t :height 0.4) popwin:special-display-config)
```

layers/+emacs/helpful/packages.el
fiveNinePlusR: I have a custom function under `spc h h` that looks like it's very similar to what `spc h h .` does(without helpful though). I think this is a good thing to have, but the overrides should be a toggle option though.

nixmaniack: > but the overrides should be a toggle option though.

👍

layers/+emacs/helpful/README.org
```diff
@@ -0,0 +1,26 @@
...
+  - [[#features][Features:]]
+- [[#install][Install[]]
+- [[#key-bindings][Key bindings]]
+
+* Description
+  This layer supports the [[https://github.com/Wilfred/helpful][helpful]] package, which extends the Emacs help buffer to include things like references and source code.
+** Features:
+   Better help buffers that include source code, callers, and prettier docstrings.
+
+* Install[
```
fiveNinePlusR: trailing [

layers/+emacs/helpful/README.org
```diff
@@ -0,0 +1,26 @@
+#+TITLE: helpful layer
+
+* Table of Contents                                        :TOC_4_gh:noexport:
+- [[#description][Description]]
+  - [[#features][Features:]]
+- [[#install][Install[]]
```
fiveNinePlusR: install link broken and trailing : on the Features link

fiveNinePlusR:
Playing around with this package and I think it's neat. It had a problem with helpful-at-point where it says selecting deleted buffer but on a restart it seemed to work correctly. instead of `spc h h .` I think it should be the default action and set to `spc h h h` for ease of use. following the convention of most of the other double press actions like `spc f f`, `spc b b`, `spc s s`, and others.

layers/+emacs/helpful/README.org
jdenen: Nice catch. Thanks.

layers/+emacs/helpful/README.org
jdenen: Nice catch on the broken link. The `:` is there for a lot of layers (like Python, for example). It's because the header is `Features:` and the thing will always autocorrect the `:` back in on save.

layers/+emacs/helpful/packages.el
jdenen: Thanks for the feedback, all.

jdenen:
That's a good idea @fiveNinePlusR. I changed it. I think I was following the package's suggested keybindings with `.`, but `SPC h h h` feels more Spacemacs-y to me.

layers/+emacs/helpful/packages.el
```diff
@@ -0,0 +1,28 @@
...
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+(defconst helpful-packages
+  '(helpful))
+
+(defun helpful/init-helpful ()
+  (use-package helpful
+    :defer t
```
aaronjensen: `:defer t` can be removed now, [all are deferred by default](https://github.com/syl20bnr/spacemacs/commit/29c78ce841ee5a99bb4077a2862afc96c8136107).

layers/+emacs/helpful/README.org
```diff
@@ -0,0 +1,30 @@
...
+* Install
+To use this configuration layer, add it to your =~/.spacemacs=. You will need to
+add =helpful= to the existing =dotspacemacs-configuration-layers= list in this
+file.
+
+* Key bindings
+
+| Key Binding | Description                              |
+|-------------+------------------------------------------|
+| =SPC h h .= | Open helpful buffer for item under point |
```
aaronjensen: `h h h`

layers/+emacs/helpful/README.org
```diff
@@ -0,0 +1,30 @@
...
+| =SPC h h f= | Open helpful buffer for function         |
+| =SPC h h k= | Open helpful buffer for key binding      |
+| =SPC h h v= | Open helpful buffer for variable         |
```
aaronjensen: missing a few of the bindings too

Compro-Prasad:
@sdwolfz This is an awesome layer. Would be great to integrate it with spacemacs. Would be better if Emacs provided it sooner. But things are slower there.

syl20bnr:
Would it make sense to replace the describe key bindings by this ?
i.e. `SPC h d v` would become `helpful-variable`

If this can be a drop in replacement then we could move it to spacemacs distribution.

aaronjensen:
> Would it make sense to replace the describe key bindings by this ?

I'd prefer not, helpful is quite a bit slower in many situations (at least it was the last time I tried it)

syl20bnr:
OK, maybe we can map it to `SPC H` prefix so we can keep the same
bindings: like `SPC h d v` --> `SPC H d v`

I don't like `SPC h h h` :-)

Compro-Prasad:
Yep, it is a bit slower because it grabs the whole definition into the buffer. We can add a variable to disable the definition lookup if speed is a higher priority. Don't know if there is a way to do it in helpful itself.

syl20bnr:
Adding a variable to pimp up the help buffers can be a solution too.

Also instead of changing `SPC h` to `SPC H` we can do this:

`SPC h d v` --> `SPC h d V`

So we can keep the prefix `SPC H` for something else.

ztlevi:
Like nixmaniack mentioned, we probly need to set the evil's initial state.

Something like:
```
(evil-set-initial-state 'helpful-mode 'normal)
(evil-make-overriding-map helpful-mode-map 'normal)
```

And one more suggestion is we could put a global variable like `helpful-override-default-keybindings`. Unless the user sets it to `t`, we overrides the defaults.

beyondpie:
Hi, all, just wonder the progress of this merge?

layers/+emacs/helpful/packages.el
syl20bnr: For the record, we have reverted `:defer t` to be the default has it introduced lots of regression in the end and increased maintenance burden.

smile13241324:
Thank you for contributing to spacemacs :purple_heart:, I have merged your PR into develop.

PR #10003, opened 2017-12-15
Fix documentation for sphinx, confluence, helm and ivy layer

smile13241324:
Added the missing feature block to the sphinx, confluence, helm and ivy layer.
This is part of #9476.


JAremko:

PR #10020, opened 2017-12-21
Fix doc for cscope, geolocation, wakatime, prodigy and smex layer

smile13241324:
Added the missing "Features:" block to the above mentioned layers. 
This is part of #9476.


JAremko:

PR #10024, opened 2017-12-22
Add defvar python-save-before-test

wizmer:
If true, buffer is automatically saved when calling
python-test-* functions


syl20bnr:
Thank you ! 👍 

PR #10025, opened 2017-12-23
Add a default value for kbd-mac-command

braineo:
Support use case setting command to keys other than hyper super or alt. Like when use swaps option and command.


nixmaniack:
This PR needs a rebase.

braineo:
@nixmaniack Seems I cannot change target branch. Will open another clean one.

PR #10027, opened 2017-12-23
Update README.org with correct settings for username and nickname.

mapolone:
Looking into [documentation](https://www.gnu.org/software/emacs/manual/html_node/rcirc/Configuration.html), `nick` has to be set instead of `user`, otherwise once connected to server (Freenode, in my case), my local login username is used instead.


JAremko:
@mapolone  PRs are accepted only to develop branch.

PR #10033, opened 2017-12-26
fix cut-and-paste error by using correct google hook function name

emacs18:
Use google-make-newline-indent rather than google-set-c-style.


syl20bnr:
Thank you ! 👍 

Issue #9858, opened 2017-11-12
More useful multi selection main behavior for helm


quicknir:
When you open multiple files in helm, it either opens them all in the same window (so only one ends up showing), or it creates a bunch of horizontal/vertical splits. I don't find either of these behaviors very good. Typically I'm working with 2-4 windows open, and if I want to open two files at once in say `SPC p f`, like a .cpp and .h file, I want it to open in my existing windows.

So, I'm suggesting a customization to the main action (enter) for various helm windows: find file, mini, projectile, etc. With one candidate the behavior would be identical. But if you mark multiple candidates and hit enter, it starts placing the buffer/files in existing windows. It could either place the first marked candidate in the current window, and then cycle up and around (e.g. if current active window is 2 and you have 4 total, it will placed 4 marked files in 2,3,4,1), or simply start at 1. If you mark more things than you have windows, then it either ignores the extra (buffers) or opens the files but doesn't place them in any windows.

I have some code that accomplishes this (I implemented the start with 1 behavior because it's simpler):

```
  (require 'helm)
  (defun helm-open-buffers-in-windows (buffers)
    (if (= (length buffers) 1) (set-window-buffer nil (car buffers))
      (let ((cur-win 1)
            (num-windows (length (window-list))))
        (cl-loop for buffer in buffers
                 do (when (<= cur-win num-windows)
                      (set-window-buffer (winum-get-window-by-number cur-win) buffer)
                      (setq cur-win (+ cur-win 1)))))))

  (require 'helm-projectile)
  (defun helm-find-files-windows (candidate)
    (let* ((files (helm-marked-candidates))
           (buffers (mapcar 'find-file-noselect files)))
      (helm-open-buffers-in-windows buffers)))

  (helm-add-action-to-source "open in 1" 'helm-find-files-windows helm-source-projectile-files-list)
  (helm-add-action-to-source "open in 1" 'helm-find-files-windows helm-source-find-files)
  (define-key helm-projectile-find-file-map (kbd "RET")
    (lambda ()(interactive)
      (helm-exit-and-execute-action 'helm-find-files-windows)))
```

Obviously this code needs a lot of fixing up. But it works and isn't very complicated. I think this makes multi selection more powerful, and has a few great use cases. I think most of us probably have more than one window open most of the time, and this both makes opening multiple files much more convenient, and also allows us to switch around the buffers in all our windows at once. I find it much more useful than either of the current behaviors (burying all the files but one, or opening lots of splits).

[...]

PR #10057, opened 2017-12-31
Add better helm default multi action for files/buffers

quicknir:
As discussed in #9858 . Without this change, when you open `SPC b b`, selecting several buffers (with C-SPC) and hitting Enter would either case buffers to be opened in splits (which rarely looked good), or in some cases might simply try to assign all the buffers to the same window, effectively hiding all but one.

With this change, doing this will cause the following to happen:

1. The first selected buffer will go into the current active window.
2. The second buffer will go into the "next" window, in the spacemacs numbering.
3.  This continues in a modular fashion. So with 4 windows open and 2 active, opening 4 buffers causes 
 the first selected buffer to be placed in window 2, 3, 4, and 1 (note: the order depends on the order in which C-SPC was hit on them, nothing else).
4. If there are N windows open, only the first N buffers get assigned to windows. In the case of files, all the files get opened as buffers, but only the first N get assigned to windows still.

IMHO, this is much more useful default behavior, and actually lets you e.g. use `SPC b b` to rearrange all of your buffers in one go, let's you quickly and efficiently open a source file along with its test file, etc.

This probably needs further testing; I've tried to make sure it behaves correctly with treemacs. It works in `SPC b b` (both buffers and recent files), `SPC p f`, `SPC f f`, and probably more.

If helm were to be opened in a window instead of the mini-buffer (I seem to recall helm supports this), it would probably break; not sure if spacemacs supports that.

### DIFF
--- a/layers/+completion/helm/funcs.el
+++ b/layers/+completion/helm/funcs.el
@@ -187,6 +187,41 @@ Ensure that helm is required before calling FUNC."
     ((eq major-mode 'org-mode) 'helm-org-in-buffer-headings)
     (t 'helm-semantic-or-imenu))))
 
+(defun spacemacs//helm-open-buffers-in-windows (buffers)
+  "This function allows a different default action, on marking multiple
+candidate buffers/files for helm. By default, helm either opens all
+files/buffers in the same window, or creates splits. This function instead
+opens the buffers (or files) across different already-open windows. The first
+selected buffer is opened in the current window, the next is opened in the
+window with higher number, etc. This will make a loop around, so with 4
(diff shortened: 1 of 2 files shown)


syl20bnr:
😍 That's really great!
I wonder if we could have the default behavior with the universal prefix argument, if one does `C-u RET` then all the files open in the current window with the last selected candidate being the one that is displayed.

Thank you ! 💜 

syl20bnr:
If we do the `C-u RET` then it will be easier to document as the user has a choice, otherwise we just describe the behavior of `RET` and it is not as nice in the doc.

PR #10058, opened 2018-01-01
autocomplete: Make idle-delay configurable

bhipple:
Adds an `autocomplete-idle-delay` layer variable that gets passed to `company-idle-delay`
or `ac-delay`.

@Valloric's YouCompleteMe vim plugin inspires a run 'n gun style of heavy autocompletion use by streamlining the user interface enough that users can happily mash TAB for just about every word they type.  @nikital has recently added a "Company Tab and Go" mode (https://github.com/company-mode/company-mode/pull/706) that mimics this fantastic behavior, but to get the most out of it requires a fairly low idle delay (otherwise it's often faster to type 3-5 more characters instead of waiting 200ms for a completion that might not be there).  Conversely, users on slow computers may wish to increase the delay.

I spent a while messing with `auto-completion-complete-with-key-sequence-delay` before realizing that `company-idle-delay` is was what I was looking for, so I figured I'd send a PR and update docs to steer others in the right direction.

### DIFF
--- a/layers/+completion/auto-completion/README.org
+++ b/layers/+completion/auto-completion/README.org
@@ -2,6 +2,7 @@
 
 * Table of Contents                                         :TOC_4_gh:noexport:
 - [[#description][Description]]
+  - [[#features][Features:]]
 - [[#install][Install]]
 - [[#configuration][Configuration]]
   - [[#key-bindings][Key bindings]]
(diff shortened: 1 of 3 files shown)


bhipple:
Looks like I hit a grandfathered-in non-compliant readme format, so I also added a `Features:` section to bring it up to speed and get CI to pass.

nikital:
Looks good.

As a side note, I would separate this commit into two commits - one that adds the missing `Features:` section and one that adds the variable and updates documentation, but this is obviously not critical in any way...

syl20bnr:
Thank you ! 👍 

That's OK for the single commit but @nikital's advice is a good one to follow.

PR #10067, opened 2018-01-03
Add switch-to-help-buffer function and keybinding

duianto:
`SPC b H` is available, for opening or selecting the help buffer (if it
exists). This makes it much easier to open the help buffer, than having
to open the buffer list (`SPC b b`), then typing one or more characters,
to select the help buffer, and finally pressing `RET`.


syl20bnr:
Nice one! 
Thank you ! 👍 

Issue #9552, opened 2017-09-07
spacemacs/search-project-ag expends with wrong file path when open selected candidate


enzochiau:
#### Description :octocat:
spacemacs/search-project-ag expands with wrong file path when open selected candidate
After  checking counsel.el, I found the new release change variable "counsel--git-grep-dir" -> "counsel--git-dir".

I did the same thing in layers/+completion/ivy/funcs.el, it seems fix the problem. 

#### Reproduction guide :beetle:
- Start Emacs
- Open a file in a project but not at the project-root directory
- Start spacemacs/search-project-ag and search anything which can return some candidates which are not in the project-root directory
- hit "Enter" to select the candidate and you will see the issue

*Observed behaviour:* :eyes: :broken_heart:
If my current buffer is <my_project_root>/src/a.cpp and try to "ag" something in the file at <my_project_root>/include/b.hpp, when select the candidate, it try to open file "<my_project_dir>/src/include/b.hpp"

*Expected behaviour:* :heart: :smile:
It should open file "<my_project_root>/include/b.hpp"

#### System Info :computer:
- OS: gnu/linux
- Emacs: 25.1.8
- Spacemacs: 0.200.9
- Spacemacs branch: develop (rev. 6d73df70)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
(csv javascript markdown
     (auto-completion :variables auto-completion-enable-help-tooltip t auto-completion-enable-sort-by-usage t auto-completion-return-key-behavior 'complete auto-completion-tab-key-behavior 'cycle auto-completion-complete-with-key-sequence nil auto-completion-complete-with-key-sequence-delay 0.1 auto-completion-private-snippets-directory nil)
     (c-c++ :variables c-c++-default-mode-for-headers 'c++-mode c-c++-enable-clang-support t)
     cppplus emacs-lisp fasd
     (git :variables git-magit-status-fullscreen t global-git-commit-mode t)
     gitplus
     (gnus :variables gnus-secondary-select-methods
           '((nntp "news.gmane.org")))
     (go :variables gofmt-command "goimports" go-use-gometalinter t)
     goplus graphviz ivy ivy-plus
     (jabber :variables jabber-account-list
             '(("enze.chi@imc.com"
                (:password . "en$17AprImc")
                (:network-server . "vIMP01.imc.com")
                (:port . 5222)
                (:connection-type . starttls))))
[...]

PR #10070, opened 2018-01-04
Replace counsel--git-dir with (ivy-state-directory ivy-last)

jynnantonix:
abo-abo/swiper@4f57b5a9 removed the counsel--git-dir variable.  Code
that used that variable is now expected to use (ivy-state-directory
ivy-last) instead.  Update the code to reflect this change.

Without this change you get the same behavior as #9552.  I did some basic testing and it seems to have fixed the problem for me with ag.  I'm less confident about the grep version.

### DIFF
--- a/layers/+completion/ivy/funcs.el
+++ b/layers/+completion/ivy/funcs.el
@@ -72,7 +72,7 @@
         "Grep in the current directory for STRING."
         (if (< (length string) 3)
             (counsel-more-chars 3)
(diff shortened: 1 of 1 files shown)


kepi:
Works for me with `rg` search. Thanks! After I updated today spacemacs was unusable for me without this fix.

syl20bnr:
Nice one, thank you ! 👍 

PR #10077, opened 2018-01-06
Adds maximize-vertically function

agzam:
I have noticed there's `maximize-horizontally` but there isn't `maximize-vertically`. 
This PR adds it.

### DIFF
--- a/layers/+distributions/spacemacs-base/funcs.el
+++ b/layers/+distributions/spacemacs-base/funcs.el
@@ -124,6 +124,16 @@ automatically applied to."
     (while (condition-case nil (windmove-right) (error nil))
       (delete-window))))
 
+(defun spacemacs/maximize-vertically ()
+  "Delete all windows left or right of the current window."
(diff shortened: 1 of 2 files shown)


layers/+distributions/spacemacs-base/funcs.el
```diff
@@ -124,6 +124,16 @@ automatically applied to."
     (while (condition-case nil (windmove-right) (error nil))
       (delete-window))))
 
+(defun spacemacs/maximize-vertically ()
+  "Delete all windows left or right of the current window."
```
duianto: The doc-string needs to be updated to:
`"Delete all windows above and below the current window."`

The doc-string for the `spacemacs/maximize-horizontally` function, also should use `and` instead of `or`.
Resulting in:
`"Delete all windows to the left and right of the current window."`

syl20bnr:
Thank you ! 👍 

I made the modification from @duianto 

PR #10079, opened 2018-01-06
Fix the broken python repl invocation through SPC a '

pinetr2e:
SPC a ' <python> will end up with the following error without this change:
  'Wrong type argument: commandp, python-start-or-switch-repl'


syl20bnr:
Thank you ! 👍

I keep it open until the hotfix is released in master.

pinetr2e:
Thank you ~

PR #10082, opened 2018-01-06
Add a default value for kbd-mac-command

braineo:
Support use case setting command to keys other than hyper super or alt. Like when use swaps option and command.

@nixmaniack Closing the previous one and open a new PR with a new branch

### DIFF
--- a/layers/+os/osx/keybindings.el
+++ b/layers/+os/osx/keybindings.el
@@ -45,6 +45,9 @@
              (when (member key-value allowed-values)
                (setf (symbol-value internal-var) key-value))))
 
+  (when osx-swap-option-and-command
+    (cl-rotatef mac-command-modifier mac-option-modifier))
+
   (defun kbd-mac-command (keys)
     "Wraps `kbd' function with Mac OSX compatible Command-key (⌘).
 KEYS should be a string such as \"f\" which will be turned into values
@@ -52,8 +55,8 @@ such as \"H-f\", \"s-f\", or \"A-f\" depending on the value of
 `mac-commmand-modifier' which could be `hyper', `super', or `alt'.
 KEYS with a string of \"C-f\" are also valid and will be turned into
 values such as \"H-C-f\".
-Returns nil if `mac-command-modifier' is set to `none' or something
-other than the three sane values listed above."
+if `mac-command-modifier' is set to `none' or something  other than
+the three sane values listed above,  bind to `H-' by default"
     (let ((found (assoc mac-command-modifier
                         '((hyper . "H-")
                           (super . "s-")
--- a/layers/+os/osx/keybindings.el
+++ b/layers/+os/osx/keybindings.el
@@ -58,7 +58,9 @@ other than the three sane values listed above."
                         '((hyper . "H-")
                           (super . "s-")
                           (alt   . "A-")))))
-      (when found (kbd (concat (cdr found) keys)))))
+      (if found
+          (kbd (concat (cdr found) keys))
+        (kbd (concat "H-" keys)))))
 
   ;; Keybindings
   (global-set-key (kbd-mac-command "=") 'spacemacs/scale-up-font)
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2066,6 +2066,10 @@ Other:
   - Added ~M-s-h~ to hide other windows (thanks to Bas Veeling)
 - Replace pbcopy by osx-clipboard (thanks to Minh Nguyen-Hue)
 - Fixed key binding issue when Emacs is launched in daemon mode.
+- Added a default value =H-= (hyper) for =kbd-mac-command= if it's neither:
+  =hyper=, =super= or =alt= (thanks to Binbin Ye)
+- Added a layer variable =osx-swap-option-and-command= defaults to =nil=
+  (thanks to Binbin Ye)
 **** Pandoc
 - Fixed =spacemacs/run-pandoc= not to reset =pandoc--local-settings=
   (thanks to martian-f)
--- a/layers/+os/osx/README.org
+++ b/layers/+os/osx/README.org
@@ -47,13 +47,15 @@ The different modifier keys can be set as follows:
                      osx-function-as      nil
                      osx-right-command-as 'left
                      osx-right-option-as  'left
-                     osx-right-control-as 'left))
+                     osx-right-control-as 'left
+                     osx-swap-option-and-command nil))
 #+END_SRC
 
 These are also the default values. Setting the right modifier to =left=
 will equal the left modifier. Allowed values are: =super=, =meta=, =control=,
 =alt= and =nil=.
 Setting =nil= for modifiers will leave the left modifiers as emacs default.
+Setting =osx-swap-option-and-command= to =t= will swap =command= and =option= key.
 
 *** Use with non-US keyboard layouts
 If you need the ~⌥~ key to type common characters such as ={[]}~= which is usual
--- a/layers/+os/osx/config.el
+++ b/layers/+os/osx/config.el
@@ -60,6 +60,9 @@
 (defvar osx-use-dictionary-app t
   "If non nil use osx dictionary app instead of wordnet")
 
+(defvar osx-swap-option-and-command nil
+  "If non nil swap option key and command key")
+
 ;; Use the OS X Emoji font for Emoticons
 (when (fboundp 'set-fontset-font)
   (set-fontset-font "fontset-default"


syl20bnr:
Sorry I'm not sure I understand the use case, what do you need to do exactly ?

braineo:
The use case is user can set 

```
(osx :variables
          osx-option-as 'hyper
          osx-command-as 'meta)
```

in `dotspacemacs-configuration-layers` which meaning swapping command and option and it makes sense for users switching between linux and Mac as it keeps muscle memory working. But osx layer limit command to one of `hyper` `super` `alt`, and once user sets command as `meta` it will give an error.

syl20bnr:
Ok, I don't understand the fix though, the condition you modified is testing the variables `mac-command-modifier` which you don't seem to modify in your use case. I'm confused on this one 😕

syl20bnr:
What is the value of `mac-command-modifier` for you ?

braineo:
that is the magic done by this layer,

in `layers/+os/osx/keybindings.el`, it sets `mac-command-modifier` to `osx-command-as`

```
(setq modifier-keys '((osx-command-as       . mac-command-modifier)
```

braineo:
in my case `mac-command-modifier` is set to `meta`

syl20bnr:
Thank you for the explanation I understand know.
Instead of hardcoding `-H` maybe we could set the keys to `osx-option-as` with the same mechanic so we support other key bindings under `osx-option-as` or maybe this use-case does not make sense.
If it does not make sense then I guess we could create a layer variables `osx-swap-option-and-commands` and document what it does in the README.

braineo:
Right, `osx-swap-option-and-commands` sounds better. I'll take a look later.

braineo:
@syl20bnr Hi I implemented `osx-swap-option-and-command` in osx layer. Can you take a look at it? I'm not an expert at writing lisp that I do not know if this is the best practice. 

Cheers.

layers/+os/osx/keybindings.el
```diff
@@ -23,12 +23,12 @@
   ;; `Alist' linking the layer config variables to
   ;; the internal Emacs variables for the modifier keys.
   (setq modifier-keys '((osx-command-as       . mac-command-modifier)
-                        (osx-option-as        . mac-option-modifier)
-                        (osx-function-as      . mac-function-modifier)
-                        (osx-control-as       . mac-control-modifier)
-                        (osx-right-command-as . mac-right-command-modifier)
-                        (osx-right-option-as  . mac-right-option-modifier)
-                        (osx-right-control-as . mac-right-control-modifier)))
+                         (osx-option-as        . mac-option-modifier)
```
duianto: The indentation for these lines should probably be restored to default.
Otherwise they will appear in a future unrelated commit if someone for example indents the whole buffer by pressing: `=ig` (with the evil editing style).

layers/+os/osx/keybindings.el
```diff
@@ -41,24 +41,30 @@
 
   ;; Set internal variables according to the given config variables
   (cl-loop for (key-var . internal-var) in modifier-keys do
-           (let ((key-value (symbol-value key-var)))
-             (when (member key-value allowed-values)
-               (setf (symbol-value internal-var) key-value))))
+    (let ((key-value (symbol-value key-var)))
```
duianto: Restore default indentation.

layers/+os/osx/keybindings.el
```diff
@@ -41,24 +41,30 @@
...
+      (when (member key-value allowed-values)
+        (setf (symbol-value internal-var) key-value))))
 
   (defun kbd-mac-command (keys)
     "Wraps `kbd' function with Mac OSX compatible Command-key (⌘).
+     (when osx-swap-option-and-command
```
duianto: These two lines have no effect, because they are in the `kbd-mac-command` functions docstring.

Before your last force push: https://github.com/syl20bnr/spacemacs/commit/61d0d97bddf37608dd10a38fd84370ffb28b3cc6
you had these two lines above the function definition for: `  (defun kbd-mac-command (keys)`

That's probably where they should be placed, then they only will run once when the `osx/keybindings.el` file is evaluated.

Leave an empty line between them and the start of the following function definition.

layers/+os/osx/keybindings.el
```diff
@@ -41,24 +41,30 @@
...
+        (cl-rotatef mac-command-modifier mac-option-modifier))
+
 KEYS should be a string such as \"f\" which will be turned into values
 such as \"H-f\", \"s-f\", or \"A-f\" depending on the value of
 `mac-commmand-modifier' which could be `hyper', `super', or `alt'.
 KEYS with a string of \"C-f\" are also valid and will be turned into
 values such as \"H-C-f\".
 Returns nil if `mac-command-modifier' is set to `none' or something
 other than the three sane values listed above."
+
```
duianto: Remove this empty line, there shouldn't be an empty line between a functions dosctring and the function body.
```suggestion
```

layers/+os/osx/keybindings.el
```diff
@@ -41,24 +41,30 @@
...
     (let ((found (assoc mac-command-modifier
-                        '((hyper . "H-")
-                          (super . "s-")
-                          (alt   . "A-")))))
-      (when found (kbd (concat (cdr found) keys)))))
+                   '((hyper . "H-")
+                      (super . "s-")
+                      (alt   . "A-")))))
+      (when found (kbd (concat (cdr found) keys))
```
duianto: The proposed change is easier to read by putting the `when` condition on it's own line:
```elisp
      (when found
        (kbd (concat (cdr found) keys))
        (kbd (concat "H-" keys)))))
```
Only the last line will be returned by the `kbd-mac-command` function. Making the first `kbd` line obsolete.

Were you intending to do this?
```elisp
      (when found (kbd (concat "H-" keys)))))
```

syl20bnr said:
>Instead of hardcoding -H maybe we could set the keys to osx-option-as with the same mechanic so we support other key bindings under osx-option-as or maybe this use-case does not make sense ?
If it does not make sense then I guess we could create a layer variable osx-swap-option-and-commands and document what it does in the README.

Therefore it probably shouldn't be hard coded.

layers/+os/osx/keybindings.el
```diff
@@ -41,24 +41,30 @@
...
+        (kbd (concat "H-" keys)))))
+
```
duianto: Remove this empty line
```suggestion
```

layers/+os/osx/keybindings.el
```diff
@@ -75,9 +81,9 @@ other than the three sane values listed above."
   (global-set-key (kbd-mac-command "`") 'other-frame)
   (global-set-key (kbd-mac-command "z") 'undo-tree-undo)
   (global-set-key (kbd-mac-command "s")
-                  (lambda ()
-                    (interactive)
-                    (call-interactively (key-binding "\C-x\C-s"))))
+    (lambda ()
```
duianto: Restore default indentation.

layers/+os/osx/keybindings.el
braineo: It was a long time ago when I setup the config. When `mac-command-modifier` is not one of `hyper` `super` or `alt`, it ended up having not bindings at all. 

Probably we shouldn't hard code this part, or we at least provide some default value here?

layers/+os/osx/keybindings.el
braineo: i think the pared it gives me this indent. Fixed

layers/+os/osx/keybindings.el
braineo: hi @duianto, thanks for your time reviewing the PR. Really appreciate it. I fixed most of the comments while having one left. 

Actually I was indenting to do give it a default value, i changed `when` to `if`

duianto: I don't have a mac so i can't test it, these suggestions were just what I noticed.
Oh yeah, now I see it, `if` instead of `when` seems like the way to go. 👍

layers/+os/osx/keybindings.el
```diff
@@ -58,8 +61,9 @@ other than the three sane values listed above."
                         '((hyper . "H-")
                           (super . "s-")
                           (alt   . "A-")))))
-      (when found (kbd (concat (cdr found) keys)))))
-
+      (if found
+        (kbd (concat (cdr found) keys))
```
duianto: Could you fix the indentation here, by default the `then` statement is indented more than the `else` statement.
```elisp
      (if found
          (kbd (concat (cdr found) keys))
        (kbd (concat "H-" keys)))))
```

And add a newline between the if statement and the following comment:
>  ;; Keybindings

duianto:
Squash this PR into two commits:
- one for the default value
- and one for the swap variable

And add two entries to the end of the `osx` section of `changelog.develop`:
https://github.com/syl20bnr/spacemacs/blob/develop/CHANGELOG.develop#osx

Possibly like this (or something better):
```
- Added a default value for =kbd-mac-command= (thanks to Binbin Ye)
- Added a variable =osx-swap-option-and-command= (thanks to Binbin Ye)
```

braineo:
@duianto cheers

duianto:
Adding the default value to the changelog is a great addition 👍

duianto:
I just noticed that the docstring for `kbd-mac-command` needs to be updated.
Currently the last sentence says:
>Returns nil if \`mac-command-modifier' is set to `none' or something
other than the three sane values listed above."

It might also be helpful to show the default value for `kbd-mac-command` in the `changelog`.
Maybe something like this (as always if you come up with anything better feel free to change it):
```
- Added a the default value =H-= for =kbd-mac-command= if it's neither:
  =hyper=, =super= or =alt= (thanks to Binbin Ye)
```

And we try to keep the changelog entries below 80 chars. If it splits the thanks parentheses then it's moved to the next line.
```
- Added a layer variable =osx-swap-option-and-command= defaults to =nil=
  (thanks to Binbin Ye)
```

braineo:
a ha, docstring.

You are right. I have updated the docstring.

duianto:
Commits: https://github.com/syl20bnr/spacemacs/commit/da3c7ee1f706bf3e9afb03cfca6fc113d0f01250 and https://github.com/syl20bnr/spacemacs/commit/f644c59ab84370b14832810e184df04824ff6406

I did some Magit juggling to fit the right changelog.develop entry and changes with right the commit.
(it was fun 😄 )
I checked that nothing was lost, but just to make sure can you test that everything works as expected one last time now that they are live.

PR #10085, opened 2018-01-07
core: Fix lambda quoted with '

Miciah:
Delete the quote preceding the lambda in `spacemacs|add-toggle` to avoid the following warning:

    .emacs.d/core/core-toggle.el: (lambda (x) ...) quoted with ' rather than with #'

While `#'` would be more correct than `'`, there is no need to quote the lambda at all.


syl20bnr:
Nice catch! Thank you again ! 👍 

PR #10092, opened 2018-01-07
Fix #10091, which-key-delay affecting key presses after which-key has already been triggered.     

### LINKED ISSUE #10091: Which-key-delay affects all actions done within which-key, when it (arguably) shouldn't.
#### Description :octocat:
Which-key delay (dotspacemacs-which-key-delay, as set in the dotfile) applies both to the initial trigger of which-key, as well as subsequent keypresses to open folders in which-key.

- Start Emacs
- (Optional: Set dotspacemacs-which-key-delay to a higher value, say, 2 seconds)
- Press SPC, wait, which key triggers correctly after 2 seconds.
- Press another key (say, f)

- Nothing happens for 2 more seconds, before which-key finally changes to 'files'.

- Which-key responds instantally to subsequent keypresses once it has already opened, changing to the 'files' submenu.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 25.3.2
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 3b24f45)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
((c-c++ :variables c-c++-default-mode-for-headers 'c++-mode c-c++-enable-clang-support t)
 csharp csv elixir elm emacs-lisp go
 (haskell :variables haskell-completion-backend 'intero)
 html java javascript latex lua markdown python shell-scripts yaml vimscript
 (shell :variables shell-default-shell 'eshell shell-default-height 40 shell-default-position 'bottom)
 syntax-checking
 (auto-completion :variables auto-completion-return-key-behavior nil auto-completion-enable-snippets-in-popup t auto-completion-tab-key-behavior 'complete auto-completion-enable-sort-by-usage t :disabled-for org)
 colors git gtags ivy org osx
 (ranger :variables ranger-override-dired t)
 semantic)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GCONF GSETTINGS NOTIFY LIBSELINUX GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 MODULES

oisincar:
A fix for #10091. The default behavour for which-key is for 'which-key-idle-delay' to affect both the initial trigger and subsequent actions. By setting 'which-key-idle-secondary-delay' to something non-nil, that delay acts over the subsequent actions instead.
    
The which-key wiki recommends setting it to a 'non-zero value', as zero could cause issues, so instead set it to 0.01.

Again, I'm not sure if this is desired behavior, if so, please ignore (but maybe something should be added to the readme instead?).

### DIFF
--- a/.travis-build.sh
+++ b/.travis-build.sh
@@ -21,7 +21,7 @@ fi
 
 if  [ $TRAVIS_SECURE_ENV_VARS = false ] &&
     [ $TRAVIS_PULL_REQUEST != false ] &&
-    [ $TRAVIS_BRANCH = "master" ]; then
+    [ "$TRAVIS_BRANCH" = "master" ]; then
 
     	printf '=%.0s' {1..70}
     	printf "\n       し(*･∀･)／   Thanks for the contribution!  ＼(･∀･*)ノ\n"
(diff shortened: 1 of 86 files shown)


syl20bnr:
Apologies accepted 😛

PR #10093, opened 2018-01-07
Fix documentation for nim, fsharp, forth, dash and fasd layer

smile13241324:
Fixed the documentations for nim, fsharp, forth and fasd layer.

Also I have massively reworked the documentation of the dash layer to make clear
that this feature is not OS X specific and actually quite useful.

This is part of #9476.


syl20bnr:
Thank you ! 💜 💜 💜 

PR #10095, opened 2018-01-08
Add null-check for filename before calling file-name-extension func

CestDiego:
I was having errors when using org-reveal and the typescript layer.

This typescript function would try to run when enabling web-mode in a temp-buffer that had web-mode activated. 

I'm not sure if this can change the behavior of the typescript layer, but given that this will only affect the code when filename is `nil` I'm confident that this wouldn't be a problem.

Please feel free to tag the owner of the typescript layer


syl20bnr:
Thank you ! 👍 

PR #10098, opened 2018-01-08
purpose: Add chat and mail purposes

Miciah:
Map erc, jabber, rcirc, and slack buffers to the "chat" purpose.

Map gnus, mu4e, and notmuch buffers to the "mail" purpose.

---

I tag this PR as WIP because I have not tested erc, jabber, gnus, or mu4e, but notmuch and rcirc buffers seem to work properly, and I would like to get feedback on the idea and approach behind this code.  Does this seem like something that should be included in spacemacs?

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -55,10 +55,9 @@ the [[file:CHANGELOG.org][CHANGELOG.org]] file.
   correctly loads custom settings at the end of the loading process, fixing
   unwanted overriden custom settings by layers. See section =Custom variables=
   in =DOCUMENTATION.org=.
-- Introduction of =spacemacs-purpose= layer in charge of handling Emacs windows
-  assignation for actions opening new buffers. This layer brings more
-  consistency to windows creation by giving them a purpose (i.e. =edit=, =help=,
-  etc...). See [[https://github.com/bmag/emacs-purpose][emacs-purpose]] repository for more info. (thanks to bmag)
+- Introduction of =spacemacs-purpose= layer that assigns newly created buffers
+  to windows based on their inferred purpose (=edit=, =help=, =chat=, =mail=,
+  etc.).  See [[https://github.com/bmag/emacs-purpose][emacs-purpose]] repository for more info (thanks to bmag)
 - New distribution =spacemacs-docker= used to build the official docker image
   for the Spacemacs container (thanks to Eugene Yaremenko)
 - Added support for =ripgrep=. To enable it add =rg= to
--- a/layers/+chat/erc/packages.el
+++ b/layers/+chat/erc/packages.el
@@ -30,6 +30,7 @@
     erc-yt
     linum
     persp-mode
+    window-purpose
     ))
 
 (defun erc/post-init-company ()
@@ -246,3 +247,6 @@
               (erc/default-servers)
             (call-interactively 'erc)))))))
 
+(defun erc/pre-init-window-purpose ()
+  (spacemacs|use-package-add-hook window-purpose
+    :pre-config (add-to-list 'purpose-user-mode-purposes '(erc-mode . chat))))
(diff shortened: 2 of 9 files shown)


Miciah:
Amended to fix an error in the mu4e layer: replaced `mu4e/post-init-window-purpose` with `notmuch/post-init-window-purpose`.

Miciah:
Rebased.

syl20bnr:
Sounds to be a good idea, cc @bmag.

bmag:
Looks good. Maybe I would've added `purpose-compile-user-configuration` to `spacemacs-post-user-config-hook` instead of calling it directly in post-init, not sure.

@Miciah what tests did you run? Did you check that notmuch and rcirc windows open in expected locations and don't screw the window layout?

If erc, jabber, gnus, or mu4e have their own specialized window positioning routines then they might screw the window layout after this change. I don't use any of them, so I can't really test them either. We either need to wait a week or so to let interested people test this PR, or just merge it and let the testing happen in develop (not the nicest thing to do to our users).

Miciah:
Thanks!

Perhaps I should have explained from the outset, I didn't have (or didn't know I had) the problems that window-purpose solves, so I'm approaching it as a tinkerer, looking for ways to integrate it more usefully into spacemacs and take advantage of its capabilities.

As for when to invoke `purpose-compile-user-configuration`, it looks like `init.el` runs layers' hooks, then (via `spacemacs/setup-startup-hook`) `dotspacemacs/user-config`, and then `spacemacs-post-user-config-hook`.  It would be convenient to users to run `purpose-compile-user-configuration` after `user-config` in order to pick up any purpose-related configuration therein, and to obviate the need to invoke `purpose-compile-user-configuration` a second time if there is such configuration in the dotfile.

On the other hand, calling `purpose-compile-user-configuration` after `user-config` could be surprising or confusing to users, who might not expect it and who could erroneously learn to expect changes to `purpose-user-mode-purposes` etc. to take effect immediately.  Also, I like the way this PR keeps things contained in the spacemacs-purpose layer.  So I'm not sure either.

As for testing, `SPC r b`, `SPC r p`, and `SPC r P` work as expected.  My notmuch and rcirc buffers have the "mail" and "chat" indicators in their status lines, which makes me smile for some reason.

slack is having some unrelated problems; I've used it in the past, but now it fails to connect, and it has exhibited weird behavior in the past on startup and when reconnecting, which I need to diagnose and report or fix some time.

I don't generally have complicated window setups, but I did experiment with opening a notmuch-search buffer in one window, a notmuch-show buffer in another window, and a non-notmuch buffer in a third window.  Then I dedicated the search buffer (why isn't there a binding for `purpose-toggle-window-buffer-dedicated`?), after which, clicking or pressing Enter on a message in the search window opened the message in the other mail window, which is nifty and feels like a traditional 2-pane MUA layout.

I'm hoping someone will volunteer to test erc, jabber, gnus, and mu4e.  I could try erc and jabber, but I feel apprehensive about trying to set up gnus or mu4e, and I wouldn't necessarily know what behavior to expect or recognize as broken.

Miciah:
I pushed an update with the following changes:

* Rebased to resolve conflicts.
* Adapted to lazy loading of window-purpose.  Layers now add pre-config hooks using `spacemacs|use-package-add-hook window-purpose :preconfig` in the the layers' `pre-init-window-purpose` functions, and the spacemacs-purpose layer now calls `purpose-compile-user-configuration)` in the `:config` for window-purpose
* Added a changelog entry.

Miciah:
Rebased.

Compro-Prasad:
This PR is so old. Is this still relevant?

duianto:
The age of a PR might not matter if there hasn't been major changes to the section a PR targets.

This was updated Mar 3, so I'm sure it's still relevant, the conflicts just need to be resolved.

duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/461a36b981b9958ae4899dc23954e46fe3105d73

PR #10103, opened 2018-01-09
Move docker keys into `SPC a`

agzam:
Most of the tools we have are accessible through `SPC a`, maybe it makes sense to have docker-layer keys there as well?


nixmaniack:
Yes please! Having Docker under top level binding doesn't make sense when we have a dedicated section for apps/tools bindings.

syl20bnr:
💯 
Thank you !

PR #10112, opened 2018-01-10
Fix errors in rcirc and slack layers

Miciah:
Fix errors that were introduced into the rcirc and slack layers by commit fea035f413bb6022326b88a74376f7185c344507.

Change the header for the slack layer's `funcs.el` to refer to the correct layer.

Change the custom layout definition for rcirc to use the correct name and binding for the layout and the correct invocation for rcirc.


syl20bnr:
Oops, sorry!
Thank you ! 👍 

PR #10117, opened 2018-01-10
treemacs: only activate extended git-mode when python3 is installed.

Alexander-Miller:
Little addendum to the previous fix. The smarter git-mode needs python3, though @quicknir did want to take a look at the python script in question.


syl20bnr:
Ok let's merge it now.
Thank you ! 👍 

PR #10119, opened 2018-01-10
notmuch: fix package loading

jb55:
The notmuch refactor in d4b931e9355ad14843ba072038b9da864553781f broke loading of the notmuch package, this fixes that.

I'm not sure why this is necessary, but it seems to fix it.

### DIFF
--- a/layers/+email/notmuch/packages.el
+++ b/layers/+email/notmuch/packages.el
@@ -33,6 +33,7 @@
 (defun notmuch/init-notmuch ()
   (use-package notmuch
     :defer t
+    :commands notmuch
     :init
     (progn
       (spacemacs/declare-prefix "aN" "notmuch")


syl20bnr:
That's really weird, there is an autoload directive in the source code for the `notmuch` function. Let's add it back then.

syl20bnr:
Thank you ! 👍 

FrancescElies:
> The notmuch refactor in d4b931e broke loading of the notmuch package, this fixes that.
> 
> I'm not sure why this is necessary, but it seems to fix it

For me it seems to work both ways, with `:commands notmuch` and without it.
When I have weird problems I run `find ~/.emacs.d/ -name "*elc" -delete` and restart emacs, maybe that would help?

jb55:
> For me it seems to work both ways, with `:commands notmuch` and without it.
> When I have weird problems I run `find ~/.emacs.d/ -name "*elc" -delete` and restart emacs, maybe that would help?

oh this may have been my problem, I do remember having to wipe my
notmuch elc files before...

PR #10121, opened 2018-01-11
c-c++: only activate rtags keybindings when rtags-mode is active

lixinchin:
The keybindings to `rtags-select-other-window', `rtags-select' and
`rtags-bury-or-delete' should only take effect when `rtags-mode-map' is
active, instead of being activate for every buffer in evil normal state.


syl20bnr:
Thank you ! 👍 

PR #10122, opened 2018-01-11
Add YANG support layer.

choppsv1:
Add support for yang models. This layer uses yang-mode package for syntax highlighting and defines a flycheck checker using pyang for syntax checking.


duianto:
We'll close this since a Yang layer was added with the following PR:
IETF related layers and packages. #12033

Issue #10118, opened 2018-01-10
Buffer rendering broken in terminal


jsoo1:
#### Description :octocat:
Spaceline renders one row too high and breaks the rendering of the rest of the buffer in terminal.

#### Reproduction guide :beetle:
- *update* Open a tmux session.
- Start Emacs in tmux
- `SPC f e d` edit a file

*Observed behaviour:* :eyes: :broken_heart:
Weird buffer rendering and spaceline appears one row too high.

Note below that `;; cider-restart` is on only line 440.  Scrolling through the buffer updates lines.
![screenshot of weird rendering](https://files.gitter.im/syl20bnr/spacemacs/EDUQ/Screen-Shot-2018-01-10-at-12.17.08-PM.png)
*Expected behaviour:* :heart: :smile:
nice good usual rendering :smile:

* A little additional information 💡 
Using `(emacs-lisp :variables emacs-lisp-hide-namespace-prefix nil)` ~fixes~ helps with the issue.  The issue is worse in tmux.

#### System Info :computer:
- OS: darwin
- Emacs: 26.0.90
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 3cb9fea0f)
- Graphic display: nil
- Distribution: spacemacs
- Editing style: hybrid
- Completion: ivy
- Layers:
```elisp
((auto-completion
  (haskell :variables haskell-completion-backend 'dante))
 c-c++ clojure csv dash docker elm
 emacs-lisp
 emoji
 erlang evil-snipe fsharp
 (git :variables git-magit-status-fullscreen t)
 (gtags :variables gtags-enable-by-default t)
 html
 (ibuffer :variables ibuffer-group-buffers-by 'projects)
 idris ivy
 (java :variables java-backend 'ensime)
 (javascript :variables tern-command
             '("node" "/usr/local/bin/tern"))
 (latex :variables latex-build-command "LaTeX" latex-enable-auto-fill t latex-enable-folding t)
 lua major-modes
 (markdown :variables markdown-live-preview-engine 'vmd)
 nginx
 (org :variables org-enable-reveal-js-support t org-enable-github-support t org-projectile-file "TODOs.org")
 osx python ranger react restclient ruby rust
 (shell :variables shell-default-height 30 shell-default-position 'bottom)
 shell-scripts
 (slack :variables slack-spacemacs-layout-name "@Slack" slack-spacemacs-layout-binding "s")
[...]

PR #10138, opened 2018-01-11
Disable nameless by default

sdwolf:
A workaround to deal with #10118 until nameless works inside a terminal.

I honestly belive such a package should not be enabled by default. Also the configuration for this reads really weird... to enable it you set the hiding variable to true... positive, negative, positive... Better if it was just `emacs-lisp-enable-nameless`, default `nil` and let the experienced users set it to `t` if they want to. 

Since even beginners are expected to be able to edit an emacs lisp file regularly, the `.spacemacs`, this package adds unnecessary complexity and offputting behaviour.

### DIFF
--- a/layers/+lang/emacs-lisp/README.org
+++ b/layers/+lang/emacs-lisp/README.org
@@ -104,12 +104,15 @@ function or press ~o~ to go out of it.
 Nameless hides package namespaces in your emacs-lisp code, and replaces it by
 leading ~>~ It can be toggled with ~SPC m T n~.
 
-By default =nameless= is activated, to disable it set the layer variable
-=emacs-lisp-hide-namespace-prefix= to =nil=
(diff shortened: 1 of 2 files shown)


syl20bnr:
I don't understand why you say `positive, negatie, positive`, the name of the variable is `emacs-lisp-hide-namespace-prefix` not `emacs-lisp-enable-hiding-of-namespace-prefix`. Sounds correct to me :-D

Anyway you are right that it should not be not enabled by default.

syl20bnr:
Thank you ! 👍 

sdwolf:
I was reading it kind of like this: "To enable the nameless package you set the variable that hides the namespace prefix to true". Maybe it's because I am not a native English speaker, and maybe I was to tired from work last evening, but while searching for a fix I had trouble understanding the meaning of that `t` value I was setting.

PR #10139, opened 2018-01-11
Keybindings for kill buffer and window

sdwolf:
Adds keybindings for `kill-buffer-and-window` both under the `buffer` and `window` prefix. 

I've created two commits for easy cherry-picking in case only one is needed.

I presonally would prefer to have them both available for convenience, in case I already pressed either `SPC b` or `SPC w` with that intent, I should not need to cancel the current action and repeat. Currently the same behaviour takes to many keystrokes to execute.

### DIFF
--- a/layers/+distributions/spacemacs-base/keybindings.el
+++ b/layers/+distributions/spacemacs-base/keybindings.el
@@ -142,6 +142,7 @@
   "bR"    'spacemacs/safe-revert-buffer
   "bs"    'spacemacs/switch-to-scratch-buffer
   "bu"    'spacemacs/reopen-killed-buffer
+  "bx"    'kill-buffer-and-window
   "bY"    'spacemacs/copy-whole-buffer-to-clipboard
   "bw"    'read-only-mode)
 (dotimes (i 9)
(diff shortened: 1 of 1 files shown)


syl20bnr:
Indeed this is rather long for a key binding that does a combo and should save time :-)

Thank you ! 👍 

ryan27968:
It seems this feature never made its way into master for some reason? Was this an oversight or should I be running from develop instead?

sdwolfz:
This will be available on master with the new relase, you can track the progress of that on this branch: https://github.com/syl20bnr/spacemacs/commits/release_notes_0.300. Ballpark ETA ~2 weeks.

ryan27968:
Ah I see. It's strange to me that it would take a year for a simple feature to get into master but I've never managed a project of this scale so I can't comment. Is develop stable enough for everyday use? Is it equivalent to Debian testing for example?

sdwolfz:
Can't compare since I don't use debian testing, but develop is generally stable enough for normal use. Right now we're in a code freeze to prepare for the release so the current version is pretty much what you'll get, unless there's a serous issue we need to fix before releasing.

ryan27968:
I don't use debian testing either. Just thought it was a good analogue.Thanks for the help though.

Davidj361:
Is this forgotten?

sdwolfz:
This PR has ben merged a long time ago, been using the finctionality daily ever since 😄 . Not released to master though, only develop.

ayr-ton:
Any news about merging this to master? :D

sdwolfz:
@ayr-ton sorry, I'm as clueless and powerless as everybody else.

PR #10141, opened 2018-01-12
 spacemacs-modeline: colorize hybrid and inactive state for vim-powerline

nixmaniack:
Commit 1: Colorize `hybrid` state for `vim-powerline`
Commit 2: Colorize `inactive` state for `vim-powerline` (don't know why it was commented out - hence separate commit if this change isn't justified)

It was colored as `normal` state now with this change: 
screen shot 2018-01-12 at 6 57 00 pm


syl20bnr:
Nice !
Thank you again !

PR #10146, opened 2018-01-13
rcirc: Move defaults to :init to be overridable

Miciah:
Move `setq` forms in `init-rcirc` from the `:config` block to the `:init` block in order to enable users to override settings in `user-config`.


syl20bnr:
Thank you ! 👍 

Issue #10133, opened 2018-01-11
Magit navigation on dvorak keyboard-layout is not working as expected.


zer09:
#### Description :octocat:
4a34ef4 break the dvorak layout navigation for magit. The up and down was rebinded.

#### Reproduction guide :beetle:
- Pull the latest commit including the commit 4a34ef4
- Start Emacs
- Open project
- Open magit
- Then try to navigate.

*Observed) behaviour:* :eyes: :broken_heart:
The the navigation key ```t``` is just openning "I am not sure if it was ```tags```".
The the navigation key ```n``` is navigating downward.

*Expected behaviour:* :heart: :smile:
The navigation key ```t``` should navigate downward.
The navigation key ```n``` should navigate upward.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 25.3.1
- Spacemacs: 0.300.0
- Spacemacs branch: test (rev. 014ed731)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
((fasd)
 (imenu-list)
 (spell-checking :variables spell-checking-enable-by-default nil)
 (syntax-checking)
 (auto-completion :variables auto-completion-return-key-behavior 'complete auto-completion-tab-key-behavior 'cycle auto-completion-enable-snippets-in-popup t auto-completion-enable-help-tooltip t auto-completion-enable-sort-by-usage t)
 (ycmd)
 (rebox :variables rebox-enable-in-text-mode t)
 (ranger :variables ranger-enter-with-minus t ranger-show-hidden t ranger-max-preview-size 1)
 (colors)
 (c-c++ :variables c-c++-enable-clang-support t c-c++-enable-rtags-support t)
 (theming)
 (keyboard-layout :variables kl-layout 'dvorak)
 (git)
 (version-control :variables version-control-diff-tool 'git-gutter+ version-control-diff-side 'left)
 (systemd)
 (shell :variables shell-default-shell 'eshell)
 (protobuf)
 (go :variables go-use-gometalinter t gofmt-command "goimports" go-tab-width 4 flycheck-gometalinter-fast t flycheck-gometalinter-deadline "30s" flycheck-gometalinter-enable-linters
     '("testify" "safesql"))
 (html)
 (javascript :variables javascript-disable-tern-port-files nil)
 (python :variables python-enable-yapf-format-on-save t python-sort-imports-on-save t)
 (shell-scripts)
[...]

PR #10148, opened 2018-01-13
Fix bug: evil-magit not initailized inside 'keyboard-layout' layer

ACEMerlin:
The Issue: https://github.com/syl20bnr/spacemacs/issues/10133

### DIFF
--- a/layers/+intl/keyboard-layout/packages.el
+++ b/layers/+intl/keyboard-layout/packages.el
@@ -25,7 +25,7 @@
     helm
     imenu-list
     ivy
-    magit
+    evil-magit
     mu4e
     neotree
     org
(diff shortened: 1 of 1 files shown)


syl20bnr:
I think we should split the configuration into `magit` and `evil-magit`. With the current proposal people in holy mode won't get the correct keybindings as `evil-magit` is not initialized at all for them.

ACEMerlin:
not sure how to fix this. FYI evil-magit initialized 2 times inside git layer, seems a bit redundant.

nixmaniack:
@ACEMerlin Where do you see `evil-magit` initialized twice? Could you point to the code?

Electron-libre:
Don't understand why is this closed. It's not merged in develop and this issue is not fixed.

nixmaniack:
> not sure how to fix this 

perhaps because of this @Electron-libre. Help needed in short!

PR #10150, opened 2018-01-14
Make tags changed on notmuch delete action configurable

leezu:


jb55:
Notmuch already has notmuch-tagging-keys that we should probably be
using for this

syl20bnr:
Thank you ! 👍 

I renamed the variables to `notmuch-message-deleted-tags` since we don't add `spacemacs` in layer variable names usually. Also added a TODO with the remark of @jb55. See commit: https://github.com/syl20bnr/spacemacs/commit/5865be40

leezu:
@jb55 on the other hand notmuch-tagging-keys maps keys to tagging actions. It is not guaranteed that the "d" key always maps to the delete action. But anyways, it should be fine to take that assumption.

Btw, I have transferred the shortcut functions notmuch defines for the archive action for the spam and delete actions: https://github.com/leezu/dotfiles/commit/2db7f4d97bda999fa44a9f9d99b9354fed5e54e3
Let me know if you believe this can be helpful for the notmuch layer.

Issue #9798, opened 2017-10-30
Neotree/Treemacs opens as window number 1


piotrpalek:
#### Description :octocat:
Opening Neotree or Treemacs on the master or develop branch (tried it with clean installs removing .spacemacs and .emacs.d a few times) results in opening the tree view in window number 1. Normally the tree should be opened in window number 0.

#### Reproduction guide :beetle:
- Start Emacs
- Press SPC f t
- Look at window number of the filetree window

*Observed behaviour:* :eyes: :broken_heart:
The filetree opens in window number 1

*Expected behaviour:* :heart: :smile:
The filetree opens in window number 0

#### System Info :computer:
- OS: gnu/linux
- Emacs: 25.3.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. cc73871c)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(ruby python helm auto-completion better-defaults emacs-lisp typescript csv yaml lua git github markdown org
      (shell :variables shell-default-height 30 shell-default-position 'bottom)
      syntax-checking version-control javascript html erlang elixir react treemacs)
```
- System configuration features: XPM JPEG TIFF GIF PNG SOUND DBUS GSETTINGS NOTIFY LIBXML2 FREETYPE XFT ZLIB TOOLKIT_SCROLL_BARS GTK2 X11

piotrpalek:
Relevant discussion on gitter: https://gitter.im/syl20bnr/spacemacs?at=59f67bfe4ff065ac1890571a

safijari:
Gyahh, I just ran into this on a new machine and thought it was something weird with the configuration on the new machine.

I know this is new but can someone suggest a way to go back (old version of vimum or whatever?). It's breaking my brain right now.

duianto:
### Cause:
The emacs-winum package made the variable `winum-assign-func` obsolete on oct 19
https://github.com/deb0ch/emacs-winum/commit/bfef4ae6d0e0e5f1fb7e13e06298e4692e5b11a6#diff-44da362b85e63f9109555d7fdb343cbdR96

The obsolete variable is used:
in the neotree layer here:
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+filetree/neotree/packages.el#L114

and in the treemacs layer here:
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+filetree/treemacs/packages.el#L74

[...]

PR #10152, opened 2018-01-14
treemacs: bind M-0 to treemacs window selection.

Alexander-Miller:
Since treemacs supports multiple frames simply giving it a static number is no longer possible. It used to receive window 10, so now M-0 is rebound to directly select treemacs, such that there won't be any de facto change in the keybinds.
When this is merged I'll push a change to treemacs that will add its buffers to winum's ignore list.

This will be another fix for https://github.com/syl20bnr/spacemacs/issues/9798#

### DIFF
--- a/layers/+filetree/treemacs/packages.el
+++ b/layers/+filetree/treemacs/packages.el
@@ -76,4 +76,4 @@
 (defun treemacs/pre-init-winum ()
   (spacemacs|use-package-add-hook winum
     :post-config
-    (add-to-list 'winum-assign-functions #'treemacs--window-number-ten)))
+    (define-key winum-keymap (kbd "M-0") #'treemacs-select-window)))


syl20bnr:
Thank you ! 👍 

Maybe it should be `SPC f t`, and we move the deletion of the file tree window under `SPC f T` ?

Alexander-Miller:
t/T are already taken with toggle/open. A basic, commonly used feature like selecting treemacs should not require more than a single (combined) keystroke, or it'll degrade UX. At least it's not something I personally would want.
With the keymaps being as crowded as they are M-0 is an ideal candidate. Yes, it's technically overwriting a part of winum, but if you have a view with more than 9 windows it's probably time to spawn more frames, or get into eyebrowse.

PR #10155, opened 2018-01-14
Add evilification of mini-buffer, including helm

quicknir:
This PR should not be merged as is. It's more of a starting point to discuss what's possible. It introduces evil state to the mini buffer. It works very well, in ivy as well (from the little I tested it). In helm however there are a few issues. One issue in both helm and the minibuffer generally is that some important bindings get overwritten. In particular, C-k in helm, and C-r in the normal minibuffer which are both pretty crucial. I restore these, but there may be other important bindings we want to restore.

The more serious issue is that while everything works in helm, there's a pretty annoying visual glitch, where the cursor is shown both in the place you type, and at the bottom where the mini buffer usually is. Ideally, we would resolve this issue, or at least minimize it reasonably. Failing that, we could simply disable the evilifcation of helm, though that's a real shame considering it actually works quite well (visual glitchiness aside).

### DIFF
--- a/layers/+distributions/spacemacs-bootstrap/funcs.el
+++ b/layers/+distributions/spacemacs-bootstrap/funcs.el
@@ -174,6 +174,19 @@ Example: (evil-map visual \"<\" \"<gv\")"
     (cl-loop for (mode uni nouni) in spacemacs--diminished-minor-modes
              do (diminish mode (if unicodep uni nouni)))))
 
+(defun spacemacs//evil-minibuffer-insert ()
+  "Switch to insert state.
+This function is meant to be hooked in the minibuffer:
+  (add-hook 'minibuffer-setup-hook 'evil-collection-minibuffer-insert)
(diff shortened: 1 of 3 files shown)


dvzubarev:
Thank you, for this. I've used it for a while and I added  some bindings:
```
  (evil-define-key 'normal helm-map (kbd "C-z") 'helm-select-action)
  (evil-define-key 'insert helm-map (kbd "C-z") 'helm-select-action)

  (evil-define-key 'insert helm-map (kbd "C-p") 'helm-previous-line)
  (evil-define-key 'normal helm-map (kbd "C-p") 'helm-previous-line)
  (evil-define-key 'insert helm-map (kbd "C-n") 'helm-next-line)
  (evil-define-key 'normal helm-map (kbd "C-n") 'helm-next-line)
```

`C-z` is quite important, I think, since it is the only way to show possible actions (is there another way?). `C-n` and `C-p` are optional. I just got used to them and can't use `C-j` and `C-k` instead.

PR #10156, opened 2018-01-14
auto-completion: add yasnippet-snippets

jackkamm:
`yasnippet` is no longer bundled with any snippets as of https://github.com/joaotavora/yasnippet/commit/b1ca2197062340d4c9e78442ea38957a96f968d4. Snippets have all been moved to a separate package `yasnippet-snippets` (https://github.com/AndreaCrotti/yasnippet-snippets). This change means that all the snippets previously available with auto-completion layer are gone. This PR adds these snippets back by adding the `yasnippet-snippets` package to auto-completion layer.

Note `yasnippet-snippets` does not need to be initialized with `use-package`; simply installing it and putting it in the load-path is enough for `yasnippet` to find the snippets.

### DIFF
--- a/layers/+completion/auto-completion/packages.el
+++ b/layers/+completion/auto-completion/packages.el
@@ -21,6 +21,7 @@
         (helm-c-yasnippet :requires helm)
         hippie-exp
         yasnippet
+        yasnippet-snippets
         auto-yasnippet
         smartparens
         ))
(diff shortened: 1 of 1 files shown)


duianto:
So that's why I was getting these two messages:
>[yas] Check your `yas-snippet-dirs': c:/Users/Username/AppData/Roaming/.emacs.d/elpa/26.0/develop/yasnippet-20180111.1533/snippets is not a directory
[yas] Prepared just-in-time loading of snippets with some errors.  Check *Messages*.

Thanks for providing a solution.

syl20bnr:
Very nice ! Thank you ! 👍 

syl20bnr:
Oh I also removed the handling of this repository in the yasnippet configuration, see commit: https://github.com/syl20bnr/spacemacs/commit/091800b2

PR #10157, opened 2018-01-14
syntax-check: use correct error list faces when defining error levels with custom bitmaps

Alexander-Miller:
With this change the error levels in the flycheck list buffer (info/warn/error) no longer revert to using the generic button face.


syl20bnr:
Nice catch 💜 
Thank you again !

Issue #9096, opened 2017-06-14
key bindings ess mode


gdkrmr:
I propose the following keybinding for `ess-mode`, they are simply all the keybinding that are accessible by `C-x`, I am not sure about spacemacs conventions on this, but they don't clash. Using the key maps gives the additional advantage of having a nice description in `which-key`.

```elisp
(spacemacs/set-leader-keys-for-major-mode 'ess-mode
    ;; More REPL
    "z"  'ess-switch-to-inferior-or-script-buffer
    "ss" 'ess-switch-process
    "sf" 'ess-eval-function
    "sF" 'ess-eval-function-and-go
    ;; ESS doc map
    "d" 'ess-doc-map
    ;; "dd" 'ess-display-help-on-object
    ;; "de" 'ess-describe-object-at-point
    ;; "dm" 'ess-manual-lookup
    ;; "do" 'ess-display-demos
    ;; "dr" 'ess-reference-lookup
    ;; "dw" 'ess-help-web-search
    ;; "dv" 'ess-display-vignettes
    ;; "di" 'ess-display-index
    ;; "da" 'ess-display-help-apropos
    ;; ESS extra map
    "e" 'ess-extra-map
    ;; "e TAB" 'ess-install-library
    ;; "e/"    'ess-set-working-directory
    ;; "ed"    'ess-dump-object-into-edit-buffer
    ;; "ee"    'ess-execute
    ;; "ei"    'ess-install-library
    ;; "el"    'ess-load-library
    ;; "er"    'inferior-ess-reload
    ;; "es"    'ess-set-style
    ;; "et"    'ess-build-tags-for-directory
    ;; "ew"    'ess-execute-screen-options
    ;; R devtools
    "w" 'ess-r-package-dev-map
    ;; "w TAB" 'ess-r-devtools-install-package
    ;; "wa"    'ess-r-devtools-ask
    ;; "wc"    'ess-r-devtools-check-package
    ;; "wd"    'ess-r-devtools-document-package
    ;; "wi"    'ess-r-devtools-install-package
    ;; "wl"    'ess-r-devtools-load-package
    ;; "ws"    'ess-r-package-set-package
    ;; "wt"    'ess-r-devtools-test-package
    ;; "wu"    'ess-r-devtools-unload-package
    ;; "wr"    'ess-r-devtools-revdep-check-package
    ;; ess dev map
    "t" 'ess-dev-map
    ;; "t?" 'ess-tracebug-show-help
[...]

PR #10200, opened 2018-01-20
Revert ess REPL bindings to follow CONVENTIONS.org

jackkamm:
A couple of recent PRs (see https://github.com/syl20bnr/spacemacs/issues/9096) added many useful keybindings to ess layer, but also modified existing REPL keybindings so they no longer follow CONVENTIONS.org. This PR reverts the offending bindings to follow the spacemacs conventions.

More specifically, "evaluate buffer and switch focus to REPL" was moved from `SPC m s B` to `SPC m s b`, while "evaluate buffer without switching focus" was moved from `SPC m s b` to `SPC m e b`. Similarly, `SPC m s r/R` was moved to `SPC m e/s r`, etc.

In the discussion linked above there seems to be some confusion about whether functions like `ess-eval-buffer` should be for "Evaluation" or "REPL" (they have different bindings in CONVENTIONS.org). However, `ess-eval-buffer` is exactly equivalent to `python-shell-send-buffer` in python layer which is mapped to `SPC m s b`. Likewise for the other "evaluation" commands such as `ess-eval-region`, etc. So I think for the sake of maintaining consistency across layers we should revert these bindings.

### DIFF
--- a/layers/+lang/ess/packages.el
+++ b/layers/+lang/ess/packages.el
@@ -92,16 +92,16 @@
         "ss" 'ess-switch-to-inferior-or-script-buffer
         "sS" 'ess-switch-process
         ;; REPL
-        "sb" 'ess-eval-buffer-and-go
-        "eb" 'ess-eval-buffer
+        "sB" 'ess-eval-buffer-and-go
+        "sb" 'ess-eval-buffer
         "sd" 'ess-eval-region-or-line-and-step
         "sD" 'ess-eval-function-or-paragraph-and-step
-        "sl" 'ess-eval-line-and-go
-        "el" 'ess-eval-line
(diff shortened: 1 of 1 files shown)


syl20bnr:
Nice catch, you are right the bindings were totally wrong, thank you ! 💜 

Sorry for letting this merged.

timor:
I was wondering what exactly the semantics are here.  I don't see what the difference is between "evaluating" and "sending to REPL".  So far, the only instance where "evaluation" seems to be chosen I have come across in practice is in the lisp modes.  But there, stuff is actually "evaluated" BY "sending" it to the repl and collecting the result.  Since the `e` bindings are usually in muscle memory in those modes, it feels strange to have to switch to `s` for different language modes...

Issue #7889, opened 2016-12-01
Evilified org-agenda map overwrites `C-h`


jschaf:
org-agenda uses `h` for `org-agenda-holidays`.  The evilified maps needs `h` for cursor movement so `evilified-state-evilify-map` searches the following characters for an unused key sequence to use for `org-agenda-holidays`.

<kbd>h</kbd> -> <kbd>H</kbd> -> <kbd>C-h</kbd> -> <kbd>C-H</kbd>

<kbd>H</kbd> is also mapped to `org-agenda-holidays`, so <kbd>C-h</kbd> is used instead.

I think <kbd>C-h</kbd> is an important enough Emacs keybinding that it shouldn't be overwritten.  I propose that `h` be removed from org-agenda-map before evilifying it to prevent overwriting `C-h`.  `org-agenda-holiday` will still be available on `H`.

jschaf:
Found a cleaner solution, use the following line to remove the binding:

      (evil-define-key 'evilified org-agenda-mode-map "\C-h" nil)

Miciah:
Since #10207 / https://github.com/syl20bnr/spacemacs/commit/e2dd5b1e4d799e940898e29cedaf9d768d2fb9b1 has been merged, can this be labeled as "Fixed in develop"?

jschaf:
Yep, looks like this is fixed.

PR #10207, opened 2018-01-21
Org: don't rebind C-h in Agenda

agzam:
this tiny PR addresses #7889


syl20bnr:
Thank you ! 👍 

We should add this rule to the auto-evilification function, I added a comment in https://github.com/syl20bnr/spacemacs/commit/c3b2f189

PR #10208, opened 2018-01-21
Clojure connection browser: add key bindings

plexus:
The functionality of cider-connection-browser (e.g. "k" for "kill connection")
was being shadowed by evil's default key bindings ("k" for "previous line").
Instead this adds leader key (`SPC m` / `,`) mappings when inside the connection
browser (`cider-connections-buffer-mode`).

This is my first contribution to Spacemacs so not sure if I'm doing this The Right Way™. Feedback welcome! Thanks.

### DIFF
--- a/layers/+lang/clojure/README.org
+++ b/layers/+lang/clojure/README.org
@@ -31,6 +31,7 @@
     - [[#stacktrace-mode][stacktrace-mode]]
     - [[#inspector-mode][inspector-mode]]
     - [[#test-report-mode][test-report-mode]]
+    - [[#cider-connection-browser][cider-connection-browser]]
   - [[#sayid-buffers][Sayid Buffers]]
     - [[#sayid-mode][sayid-mode]]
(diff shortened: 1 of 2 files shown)


syl20bnr:
Why not using `evilified-state-evilify-map` instead ? Navigation with `hjkl` is not useful in this buffer ?
example: https://github.com/syl20bnr/spacemacs/blob/develop/layers/+lang/clojure/packages.el#L203-L212

Note that it will rebind kill-connection under `C-k` automatically.

syl20bnr:
BTW congrats on your first PR 💜

kommen:
I think this PR is obsolete as cider has revamped it connection management to use sesman and all the `cider-connection*` functions a marked obsolete:
https://github.com/clojure-emacs/cider/blob/db76168b776c490068b66f46858e48024cad56b3/cider-connection.el#L628-L636

plexus:
Sorry for dropping the ball on this. I'll see if I can come up with a better approach in the coming week or two.

PR #10209, opened 2018-01-21
enable flycheck for tuareg mode rather than merlin mode

rgrinberg:
merlin mode doesn't actually exist


syl20bnr:
Thank you ! 👍 

PR #10214, opened 2018-01-22
Adds magit-diff-buffer-file-popup key

agzam:
Sorry for a minuscule PR, this keybinding I think makes sense. ~I also tried to update layer's readme, but there's no section on `g f` keybindings. I wasn't sure where to add one~.

and while we're at it, can we please consider changing `g f h` to `g f l` maybe?

### DIFF
--- a/layers/+source-control/git/README.org
+++ b/layers/+source-control/git/README.org
@@ -106,7 +106,7 @@ Git commands (start with ~g~):
 |-------------+-----------------------------------------------------|
 | ~SPC g b~   | open a =magit= blame                                |
 | ~SPC g f f~ | view a file at a specific branch or commit          |
-| ~SPC g f h~ | show file commits history                           |
+| ~SPC g f l~ | commits log for current file                        |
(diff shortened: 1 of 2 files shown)


syl20bnr:
> can we please consider changing g f h to g f l maybe?

Good idea.

Be sure to update the documentation in corresponding `README.org` file when you submit/change key bindings ;-)

agzam:
@syl20bnr Done!

syl20bnr:
Nitpicking but can you split into 2 commits ? I would like to have the key binding change in its own commit to people can quickly spot it in the history or git blame. Or maybe you can put both changed on the first line of the commit if enough space for it.

agzam:
@syl20bnr I cannot agree with you more. Every single modified or new keybinding should be added with care and increased visibility.

agzam:
@syl20bnr this is ready to merge btw.

sdwolfz:
Thank you :heart:!

PR #10217, opened 2018-01-22
mu4e: Fix major-mode leader-key bindings

Miciah:
Instead of specifying `mu4e-foo-mode-hook` to `spacemacs/set-leader-keys-for-major-mode`, specify `mu4e-foo-mode`.


syl20bnr:
Outch, nice catch 😳 
Thank you ! 👍 

PR #10224, opened 2018-01-22
clojure: Enable safe structural editing for cljs and cljc

kommen:
`evil-lisp-safe-structural-editing-modes` was introduced in https://github.com/syl20bnr/spacemacs/commit/2cb98447afa3230d579d349b42e03c23017f43f5#diff-43a23dbec19deb02144c5310368c065e for `clojure-mode` only.

`clojurescript-mode` and `clojurec-mode` are variants of `clojure-mode` used for .cljs and .cljc files
respectively, which should always be balanced and thus have safe structural
editing enabled.

This is my first PR here. Feedback very welcome!

### DIFF
--- a/layers/+lang/clojure/packages.el
+++ b/layers/+lang/clojure/packages.el
@@ -313,7 +313,11 @@
 (defun clojure/pre-init-evil-cleverparens ()
   (spacemacs|use-package-add-hook evil-cleverparens
     :pre-init
-    (add-to-list 'evil-lisp-safe-structural-editing-modes 'clojure-mode)))
+    (dolist (m '(clojure-mode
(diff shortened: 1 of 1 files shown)


syl20bnr:
Congratulations on your first PR for Spacemacs 💜 

You can use a `dolist` form, example: https://github.com/syl20bnr/spacemacs/blob/develop/layers/+lang/clojure/packages.el#L46-L51

kommen:
@syl20bnr thanks for the feedback. I updated the commit as suggested and also added clojurex-mode.

syl20bnr:
Thank you ! 👍 

PR #10226, opened 2018-01-22
Make nimsuggest-doc-mode-map as overlay

cipharius:
Raise the `nimsuggest-doc-mode-map` to overlay map priority.

This is a special-mode map, which only appears in Nim quick documentation buffers and should have overlay bindings.

### DIFF
--- a/layers/+lang/nim/packages.el
+++ b/layers/+lang/nim/packages.el
@@ -25,6 +25,8 @@
       (push 'nimsuggest-find-definition spacemacs-jump-handlers-nim-mode))
     :config
     (progn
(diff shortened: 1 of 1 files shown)


syl20bnr:
Should we `evilify` this map instead ?

ZarsBranchkin:
Alright with me, wasn't sure of best method to evilify a map.

I looked into evilified state macros and it seems that they rely on mode being in evilified-state. Problem here is that `nimsuggest-show-doc` doesn't use a custom mode for the output buffer, but instead uses `rst-mode`. Not sure how to go about this.

PR #10241, opened 2018-01-24
Add +tools/lsp-python layer

casouri:
With ongoing support of lsp layer in #10211, I think I can contribute something to lsp-python layer.
This layer is depend on lsp layer. Because I really just a beginner, python-language-server needs to be installed manually by user using pip. But no other configuration is needed.

The functionality is not as versatile as python layer, but most of the daily functionalities are covered well. Autocompletion, jumping, executing script, doc, etc are all covered.

### DIFF
--- a/layers/+tools/lsp-python/README.org
+++ b/layers/+tools/lsp-python/README.org
@@ -1,0 +1,52 @@
+#+TITLE: lsp-python layer
+
+# The maximum height of the logo should be 200 pixels.
+[[img/lsp-python.png]]
+
+# TOC links should be GitHub style anchors.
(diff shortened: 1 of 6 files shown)


casouri:
Ok it seems the problem is readme.org, I tried to fix it.

casouri:
I kind of messed up... Let me see how can I recover this mess...

syl20bnr:
Thank you ! 👍 

I added LSP support directly to the python layer in commit https://github.com/syl20bnr/spacemacs/commit/81a931f4

I merged your PR but deleted the `lsp-python` layer in the above commit as it is not needed anymore.
See updated `README.org` of python layer to learn how to enable LSP.

Issue #9489, opened 2017-08-25
OCaml Layer:  Should be using flycheck for error checking


rgrinberg:
The OCaml layer is using merlin's own error reporting ability. Which is inferior to what https://github.com/flycheck/flycheck-ocaml provides. spacemacs should switch to use this package instead.

CeleritasCelery:
you should open a PR

jpathy:
To enable flycheck-mode you need to uncomment the 2 lines in ocaml-packages. The rest of the setup is already in place. You do need to enable syntax-checking layer and enable flycheck in files(SPC t s). 

I guess by inferior merlin error reporting, you mean it doesn't report errors on-fly, rather on save.

jeremyjh:
@jpathy I have syntax-checking enabled and use flycheck in many other layers. But un-commenting those package names in the ocaml layer does not make flycheck available. Instead I see this warning when Spacemacs starts:

```
            - package flycheck not initialized in layer ocaml, you may consider removing this
              package from the package list or use the :toggle keyword instead of a `when'
              form.      
```

Toggling syntax-checking does nothing.

jpathy:
@jeremyjh The problem seems to be : 
if you have syntax-checking package ordered after ocaml, then 
https://github.com/syl20bnr/spacemacs/blob/3208cc94f2bf3ea448805f0d66b003b2b2574022/layers/%2Blang/ocaml/packages.el#L34
prevents it from evaluating the necessary defs. You can change the order and it will download `flycheck-ocaml` and load it.

No other package seems to be doing this, so i removed the test. Also it seems like adding a minor-mode to `flycheck-global-modes` doesn't work(auto enable on buffer). So i had to change to `(spacemacs/enable-flycheck 'tuareg-mode)` .

jeremyjh:
@jpathy That was it - moving the syntax-checking earlier resolved it for me, thanks!

CeleritasCelery:
Feel free to close this if you feel that your question has been answered.

MarcusE1W:
Hi I am new to Sacemacs and am not quite sure about the result of the issue discussion. Is there an intention to updat the official ocaml layer or is the reccomendation to update some scripts manually ?

MarcusE1W:
Short update. I have followed the above instructions and it works fine.

Is there an intention to make this the default function, or maybe with a toggle?

In the meantime, if there is a layer update, will that overwrite the manual changes?

dagnachewa:
[...]

PR #10254, opened 2018-01-24
Ocaml layer: enable ocaml-flycheck package via variable

actionshrimp:
This adds a layer variable to the ocaml layer to use flycheck.

The support was already in there, but it seems from discussion here https://github.com/syl20bnr/spacemacs/issues/9489 the recommended way of enabling it was editing the layer file manually to enable the package. This just uses the variable as a package toggle.

I also enabled flycheck automatically for ocaml's tuareg-mode as the optional package is loaded.

### DIFF
--- a/layers/+lang/ocaml/packages.el
+++ b/layers/+lang/ocaml/packages.el
@@ -13,8 +13,8 @@
   '(
     ;; auto-complete
     company
-   ;; flycheck
-   ;; flycheck-ocaml
+    flycheck
+    flycheck-ocaml
     ggtags
     counsel-gtags
     helm-gtags
--- a/layers/+lang/ocaml/README.org
+++ b/layers/+lang/ocaml/README.org
@@ -7,6 +7,7 @@
   - [[#features][Features:]]
 - [[#install][Install]]
   - [[#layer][Layer]]
+  - [[#using-merlin-for-error-reporting][Using merlin for error reporting]]
   - [[#opam-packages][OPAM packages]]
 - [[#key-bindings][Key Bindings]]
   - [[#repl-utop][REPL (utop)]]
@@ -22,13 +23,23 @@ This is a very basic layer for editing ocaml files.
 - Syntax highlighting (major-mode) via [[https://github.com/ocaml/tuareg][tuareg-mode]]
 - Error reporting, completion and type display via [[https://github.com/ocaml/merlin][merlin]]
 - auto-completion with company mode via [[https://github.com/ocaml/merlin][merlin]]
-- syntax-checking via [[https://github.com/diml/utop][flycheck-ocaml]]
+- syntax-checking via [[https://github.com/ocaml/merlin][merlin]] or [https://github.com/flycheck/flycheck-ocaml][flycheck-ocaml]
 
 * Install
 ** Layer
 To use this configuration layer, add it to your =~/.spacemacs=. You will need to
 add =ocaml= to the existing =dotspacemacs-configuration-layers= list in this
 file.
+
+** Using merlin for error reporting
+By default [[https://github.com/flycheck/flycheck-ocaml][flycheck-ocaml]] is used for error reporting when the
+=syntax-highlighting= layer is also enabled as this is common throughout spacemacs.
+You can disable this and switch back to [[https://github.com/ocaml/merlin][merlin]]'s default error reporting method
+by adding [[https://github.com/flycheck/flycheck-ocaml][flycheck-ocaml]] to your excluded packages list in =.spacemacs=:
+
+#+BEGIN_SRC emacs-lisp
+dospacemacs-excluded-packages '(... flycheck-ocaml ... )
+#+END_SRC
 
 ** OPAM packages
 This layer requires some [[http://opam.ocaml.org][opam]] packages:


rgrinberg:
Do we really need this? I don't recall any other layer with flycheck support having such a variable? I'm pretty sure you can also just exclude flycheck-ocaml package to disable the support if you really want.

layers/+lang/ocaml/packages.el
```diff
@@ -41,6 +40,8 @@
       :defer t
       :init
       (progn
+        (add-hook 'tuareg-mode-hook 'flycheck-mode)
```
rgrinberg: I don't think this is necessary as we already have `(spacemacs/enable-flycheck 'tuareg-mode)`

actionshrimp: Ah, it didnt seem to kick in but I'd missed that - will take another look and see if I can clean it up.

actionshrimp:
Thanks for taking the time to take a look - the thing is that merlin mode (which is included) has it's own 'flycheck-esque' error reporting and error list which needs to be disabled for flycheck to work.

Perhaps it makes sense to use flycheck by default and allow switching back to the merlin error stuff by variable. But perhaps you're suggesting someone can just add flycheck-ocaml to their additional-packages list, in which case I'll change this to mention that in the README as it wasn't clear to me at first.

rgrinberg:
You're right about merlin's own attempt to reinvent this. Luckily, we can disable it by setting `(setq merlin-error-after-save nil)`. Which is what we do after enabling flycheck-ocaml.

actionshrimp:
Ah yep, sorry if I wasn't clear - I'm aware the layer does this stuff already I was just trying to make it more easy to turn in a way that didn't require manual edits of the layer source.

Turns out as flycheck wasnt in the layers package list (or rather it was, but was commented out) that `(spacemacs/enable-flycheck 'tuareg-mode)` wasn't kicking in, which I've now rectified and cleaned up.

I've also inverted the variable so that the current default behaviour (using merlin's error reporting) remains that way, but you can turn off merlin and 'revert' to the spacemacs standard of using flycheck for error reporting by setting the layer variable `ocaml-use-merlin-error-reporting` to `nil`.

actionshrimp:
Aware this is getting a bit spammy now, so long story short: this is a workaround for these lines being commented out in `master` and `develop`:

https://github.com/syl20bnr/spacemacs/blob/develop/layers/+lang/ocaml/packages.el#L16-L17

being commented out which perhaps they shouldnt be but I thought perhaps they were for historical reasons. As you say if they were uncommented the user could just exclude `flycheck-ocaml` to disable it and revert to merlin.

rgrinberg:
Hmm, I'm still not sure what problem is this variable solving.

* If the user doesn't list flycheck-ocaml in his packages, then merlin's own reporting should work.
* If the user lists flycheck-ocaml in his `dotspacemacs-additional-packages`, then spacemacs will correctly configure flycheck-ocaml and disable the merlin reporting.

We could argue about the default behavior here, but I don't see what option we are currently missing. Sorry if I'm being obtuse here.

actionshrimp:
Ah - the problem is that adding `flycheck-ocaml` to `dotspacemacs-additional-packages` doesn't have any effect as the layer doesn't run the `ocaml/...init...` functions unless the packages are listed in the layer's package list.

rgrinberg:
Dave Aitken <notifications@github.com> writes:

> Ah - the problem is that adding `flycheck-ocaml` to `dotspacemacs-additional-packages` doesn't have any effect as the layer doesn't run the `ocaml/...init...` functions unless the packages are listed in the layer's package list.

Hmm, well this mechanism works just fine for me. I have flycheck-ocaml
configured and working as expected and that's all I added to my config.

actionshrimp:
I had a chance to test this in a bit more detail and I can't replicate what you observe on `develop` (also tried on `master`) with a clean `.emacs.d` and `.spacemacs` with the only changes as in the diff below:

```
$ diff --git a/Users/dave/.emacs.d/core/templates/.spacemacs.template b/Users/dave/.spacemacs
index 80e47f6f..f108b08a 100644
--- a/Users/dave/.emacs.d/core/templates/.spacemacs.template
+++ b/Users/dave/.spacemacs
@@ -51,8 +51,9 @@ This function should only modify configuration layer settings."
      ;;        shell-default-height 30
      ;;        shell-default-position 'bottom)
      ;; spell-checking
-     ;; syntax-checking
+     syntax-checking
      ;; version-control
+     ocaml
      )

    ;; List of additional packages that will be installed without being
@@ -62,7 +63,7 @@ This function should only modify configuration layer settings."
    ;; To use a local version of a package, use the `:location' property:
    ;; '(your-package :location "~/path/to/your-package/")
    ;; Also include the dependencies as they will not be resolved automatically.
-   dotspacemacs-additional-packages '()
+   dotspacemacs-additional-packages '(flycheck flycheck-ocaml)

    ;; A list of packages that cannot be updated.
    dotspacemacs-frozen-packages '()
```

I am fairly convinced from poking my own private layers that the layer mechanism uses that package list to decide which init functions to run - for example in the [https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Blang/elm/packages.el][elm layer] they include both flycheck and the elm flycheck package to achieve a similar thing.

mattjbray:
I see the same as @actionshrimp (using latest `develop`). Adding `flycheck-ocaml` to `dotspacemacs-additional-packages` is not enough to trigger `ocaml/init-flycheck-ocaml`.

When I edit the `ocaml` layer to un-comment `flycheck-ocaml` in the `ocaml-packages` list, and restart Spacemacs in debug mode (`SPC q d`), then I see in the message log (`SPC w p m`):
```
(Spacemacs) flycheck-ocaml -> init (ocaml)...
```

So it does seem to me that some change in the layer is needed for users to enable `flycheck-ocaml` without having to edit the `ocaml` layer. I think a layer variable is perfect for this.

rgrinberg:
I still don't see how such a variable is desirable. The elm layer that was pointed out doesn't include such a variable. Nor any other layer as far as I can tell.

If a user syntax-checking on and has flycheck installed, it's unlikely that he wants to tweak a bunch of layer specific variables as well? Is it possible that we just make flycheck-ocaml the default?

actionshrimp:
Yeah that sounds good to me, just wasn't sure on the etiquette of swapping it over for ocaml merlin fans. I'll update the PR and put a note in the README that it can be disabled by adding to the excluded packages list which does seem to work when I get a sec, thanks.

actionshrimp:
Should all be up to date now.

syl20bnr:
Thanks to both of you for working on it ! 👍 

PR #10266, opened 2018-01-27
Fix broken package declaration for dockerfile-mode

smile13241324:
In the docker layer a package `dockerfile-mode` is loaded to provide features for editing `dockerfiles` however the `use-package` declaration did not load the right package, causing it to skip the `:config` part of the declaration.

This did not bind the `dockerfile-build-buffer` function.


syl20bnr:
Thank you ! 👍 

PR #10269, opened 2018-01-28
Increase gc-cons-threshold during startup

aaronjensen:
For me, this results in a drop from 9-10 seconds to 8-9 second startup.

Idea borrowed from https://github.com/hlissner/doom-emacs/wiki/FAQ#how-is-dooms-startup-so-fast

### DIFF
--- a/init.el
+++ b/init.el
@@ -12,9 +12,15 @@
 ;; Without this comment emacs25 adds (package-initialize) here
 ;; (package-initialize)
 
-;; Increase gc-cons-threshold, depending on your system you may set it back to a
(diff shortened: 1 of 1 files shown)


syl20bnr:
Thank you ! 👍 

To cleanup the init file and allow people to modify the values without forking Spacemacs (which should not be necessary) I created a new variable `dotspacemacs-gc-cons`, see commit https://github.com/syl20bnr/spacemacs/commit/72c89df9

aaronjensen:
Awesome, good idea. Thanks for the quick merge!

PR #10271, opened 2018-01-28
Fix doc for django, scala, sml, floobits, puppet and evernote layer

smile13241324:
Fixed the missing features block for the django, scala, sml, floobits, puppet and evernote layers.
This is part of #9476.


syl20bnr:
💜 💜 💜

PR #10278, opened 2018-01-30
Add defun spacemacs/python-test-last

wizmer:
It is a wrapper around the nosetests-again defun.
It will call the latest called nosetest function.
I've also added a binding for it: leader key + "t l" (mnemonic = test last)


syl20bnr:
Thank you ! 👍 

PR #10282, opened 2018-01-31
Fix rcirc README to ensure you can login

rhoml:
Fix typo on rcirc attribute configuration as per rcirc [docs](https://www.gnu.org/software/emacs/manual/html_node/rcirc/Configuration.html)


Miciah:
The change in the last commit looks correct, but your branch needs to be rebased onto the develop branch.

rhoml:
@Miciah Commit rebased should be good to go now

syl20bnr:
Thank you ! 👍 

PR #10287, opened 2018-02-01
fix: ycmd don't search for compilation DB in subfolders

myrgy:
ycmd don't search for compilation_commands.json in subfolders.


syl20bnr:
Thank you ! 👍 

PR #10289, opened 2018-02-01
Fix doc for graphviz, deft and finance layer

smile13241324:
Fixed the missing features block for the graphviz, deft and finance layer.
I have also revised the Plantuml layer a bit, mostly to make installing and configuring
plantuml easier.

This is part of #9476


JAremko:
Thank you ! +1

PR #10296, opened 2018-02-03
Wrap init in file-name-handler-alist nil to speed it up

aaronjensen:
This shaves about 0.7s off of my init, or 7.8%

See
https://www.reddit.com/r/emacs/comments/3kqt6e/2_easy_little_known_steps_to_speed_up_emacs_start/
and https://github.com/hlissner/doom-emacs/wiki/FAQ#how-is-dooms-startup-so-fast


aaronjensen:
I wonder if this will impact resuming layouts in some way? Where does that happen? Maybe we don't need to wrap that piece with the let.

aaronjensen:
Ah, it looks like there may be a delay, so it may actually wait until init is completely done, so this is probably fine. It'd be great if someone else could test this out.

aaronjensen:
ping @syl20bnr @TheBB @d12frosted @bmag @JAremko

syl20bnr:
I can confirm the boost on several on my installs, nice one 👍 
Thank you again !

PR #10297, opened 2018-02-03
Make the Emacs server optional

tadfisher:
Add `dotspacemacs-enable-server`, which if non-nil prevents Spacemacs from
calling `(server-start)` at init.

This is useful for cases in which automatically starting a server is undesirable.


syl20bnr:
Thank you ! 👍 

PR #10304, opened 2018-02-05
Fix doc for nginx, jr, nlinum and systemd layer

smile13241324:
Fixed the missing features block for nginx, jr, nlinum and systemd layers.
This is part of #9476.


JAremko:
Thank you ! :+1: 

PR #10305, opened 2018-02-05
Adding use-package-chords to bootstrap.

dangirsh:
This enables the :chords keyword in use-package definitions for defining key
chords (with key-chord.el). I tried several ways to avoid adding this to the
bootstrap layer, but none loaded the package early enough.

See https://github.com/jwiegley/use-package#use-package-chords


aaronjensen:
Can you give an example of when this would be used?

dangirsh:
The examples in the `use-package` docs are relevant for Spacemacs:

```
(use-package ace-jump-mode
  :chords (("jj" . ace-jump-char-mode)
           ("jk" . ace-jump-word-mode)
           ("jl" . ace-jump-line-mode)))
```

duianto:
Confirmed, this works.

I got it to work like this::
- Add the `ace-jump-mode` package to the additional packages list:
```elisp
   dotspacemacs-additional-packages '(ace-jump-mode)
```
- Add this to the user config section:
```elisp
    (use-package ace-jump-mode
      :chords (("jj" . ace-jump-char-mode)
               ("jk" . ace-jump-word-mode)
               ("jl" . ace-jump-line-mode)))

    (key-chord-mode 1)
```

It would be helpful to document these (or better) instructions on how to set it up.

Possibly in the `Commands` section: https://github.com/syl20bnr/spacemacs/blob/develop/doc/DOCUMENTATION.org#commands
Under a new heading between the `Vim key bindings` and `Reserved prefix command for user` sections, because the vim section has an entry for evil escape which also uses a key sequence.

A possible heading might be:
Key chords

It could start by saying something like this:
To bind a command to a key chord (a sequence of keys pressed quickly).

This PR also needs a `changelog.develop` entry in the `Distribution changes` section, at the end of the `Improvements` sub section.

A possibly descriptive entry could be:
> - Added support for =use-package-chords= (thanks to Dan Girshovich)

duianto:
The `Ace Jump` link in the sentence:
>For example, to add key chords for Ace Jump, you might have

leads to a specific commit:
https://github.com/winterTTr/ace-jump-mode/tree/8351e2df4fbbeb2a4003f2fb39f46d33803f3dac

Was that intentional? Or if it's just a typo then it probably should link to the main page.
https://github.com/winterTTr/ace-jump-mode

---

I haven't used `use-package` directly, is it also supposed to install the `ace-jump-mode` package?
```elisp
    (use-package ace-jump-mode
      :chords (("jj" . ace-jump-char-mode)
               ("jk" . ace-jump-word-mode)
               ("jl" . ace-jump-line-mode)))

    (use-package key-chord
      :config
      (key-chord-mode 1))
```

It doesn't do that for me. Nothing happens after adding the lines above to the user config section of my `.spacemacs` and restarting Spacemacs.

It gets installed and works as long as `ace-jump-mode` is in to the additional packages list.

If I'm not doing something wrong, and this behavior isn't a bug (aka it's working as expected).

Then there needs to be some instructions stating that `ace-jump-mode` needs to be added to the existing `dotspacemacs-additional-packages` variable in `.spacemacs`:
```elisp
   dotspacemacs-additional-packages '(ace-jump-mode)
```

An alternative could be to replace the examples with the equivalent `avy` commands since `avy` comes with Spacemacs:
- `evil-avy-goto-char-timer` (like `ace-jump-char-mode` except that it accepts any number of characters)
- `evil-avy-goto-word-or-subword-1`
- `evil-avy-goto-line`

But it might be better to have another location in the documentation where it's demonstrated how to use an external package.

---

And could you squash the commits.

aaronjensen:
@duianto this PR doesn't have anything to do with `ace-jump-mode` other than using it as an example, so I'd say that its actual usage and the particulars around `use-package` are out of scope for this PR. This is only about enabling the `:chords` keyword.

duianto:
Understood but I'm sure a beginner will copy the snippets to their user config section and ask why it isn't working or they might assume that they did something wrong and not try it again.

If we have a working example then they can start from there and modify it as they learn more.

Then it's probably better if we use a simpler example, so that it will work with a minimal setup.

My suggestion of using avy doesn't seem to work:
```elisp
    (use-package avy
      :chords (("jj" . evil-avy-goto-char-timer)
               ("jk" . evil-avy-goto-word-or-subword-1)
               ("jl" . evil-avy-goto-line)))

    (use-package key-chord
      :config
      (key-chord-mode 1))
```
Nothing happens when I press the chords (with this PR applied).

Any suggestions for a simpler example?

PR #10310, opened 2018-02-06
Configure rg max column number for search.

mijoharas:
Allow user configuration of max-column setting for searching by `ripgrep`.

Otherwise, `C-c C-e` (edit all) will swap a bunch of lines to be `[Omitted long line with n matches]`.

### DIFF
--- a/layers/+completion/helm/config.el
+++ b/layers/+completion/helm/config.el
@@ -37,6 +37,9 @@ values."
   in all non-asynchronous sources. If set to `source', preserve individual
   source settings. Else, disable fuzzy matching in all sources.")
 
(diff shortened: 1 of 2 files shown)


mijoharas:
(this allows user to set their own, or nil to be able to not have a `max-column` setting applied e.g.:

```
  (setq spacemacs-helm-rg-max-column-number nil)
```

mijoharas:
Does anyone have any opinions on this? would be nice to get it off my branch onto main, (global find and replace is broken without it for me).

syl20bnr:
Thank you ! 👍 

I bumped the default to 512 and added some doc in Helm README.org file.

mijoharas:
Thanks!

PR #10326, opened 2018-02-09
Add SPC w | binding for split-window-right

dp12:
Added to match the SPC w - binding for split-window-bottom.


sdwolfz:
Suggestion: put it next to `w\`. Also please add the keybinding to the documentation.

dp12:
Thanks, I just saw the conflicted keybinding as I was looking up the documentation.

PR #10332, opened 2018-02-11
shell: move eshell key bindings definition into a hook function

p-sdk:
Problem
---

`C-j` and `C-k` key bindings don't work in eshell buffers except the first one.

Solution
---

`eshell-mode-map` is buffer-local, so key bindings have to be defined in a hook
function.


syl20bnr:
Thank you again !

PR #10338, opened 2018-02-12
javascript: Add variable to disable tern warning at startup

edran:
This PR adds `javascript-disable-tern-warning` to the `javascript` layer, which allows users to disable the warning if they don't wish to install tern (so that there's no need to rely on the exclude-package hack).

This should be useful for instance to folks who use the javascript layer just for `json-mode`.

### DIFF
--- a/layers/+lang/javascript/README.org
+++ b/layers/+lang/javascript/README.org
@@ -85,6 +85,9 @@ layer:
   (javascript :variables tern-command '("node" "/path/to/tern/bin/tern"))
 #+END_SRC
 
+If you don't use tern and/or do not wish to install it, you can disable the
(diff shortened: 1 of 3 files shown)


syl20bnr:
Do you think it is possible to completely get rid of this annoying warning ? Maybe we can make it lazy and display it only when the use actually uses a tern related key binding ?

edran:
@syl20bnr Good question. Probably someone can do it, but I don't really use Tern and I don't have the time to figure out how / what to install to get this going.

NightMachinary:
I believe this tern warning comes even when Emacs does find tern. It seems it looks for tern before setting up exec-path.

PR #10357, opened 2018-02-17
Smarter upcase function for keyboard-layout layer.

marshall-lee:
In my custom layout I have to remap `";"` key.


syl20bnr:
Thank you fro the PR.
This function may be keyboard layout specific. Depending on the keyboard layout language the upper key may be different.

PR #10359, opened 2018-02-18
Add typescript-tsx-mode (derived from web-mode)

aaronjensen:
This makes it easier to configure things for tsx files. It borrows from the idea
in the react layer.

### DIFF
--- a/layers/+lang/typescript/config.el
+++ b/layers/+lang/typescript/config.el
@@ -21,3 +21,4 @@ Currently avaliable 'tide (default)
 and 'typescript-formatter .")
 
 (spacemacs|define-jump-handlers typescript-mode)
+(spacemacs|define-jump-handlers typescript-tsx-mode)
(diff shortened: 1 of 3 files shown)


syl20bnr:
Thank you ! 👍 

linchen2chris:
typescript-tsx-mode is missing in spacemacs now,

aaronjensen:
> typescript-tsx-mode is missing in spacemacs now,

Are you on `develop`? I don't think it's on `master` yet, and I still see it on `develop`

linchen2chris:
It works now. I am on develop branch, but forgot to add html layer 😅

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

aaronjensen:
Ah, I wonder then if the html layer should be depended on by the typescript layer now...

linchen2chris:
> Ah, I wonder then if the html layer should be depended on by the typescript layer now...

that would be better

PR #10367, opened 2018-02-19
Fix package initialization on Emacs 27

Miciah:
Before Emacs 27, the init file was responsible for initializing the package manager by calling `package-initialize`.  Emacs 27 changed the default behavior: It now calls `package-initialize` before loading the init file.  This behavior would prevent Spacemacs's own package initialization from running.  However, Emacs 27 also loads the "early init" file, `early-init.el`, before it initializes the package manager, and Spacemacs can use this early init file to prevent Emacs from initializing the package manager.  (See
<http://git.savannah.gnu.org/cgit/emacs.git/commit/?id=24acb31c04b4048b85311d794e600ecd7ce60d3b>.)

Earlier Emacs versions do not load the early init file and do not initialize the package manager before loading the init file, so `early-init.el` is neither needed nor loaded on those versions.

### DIFF
--- a/early-init.el
+++ b/early-init.el
@@ -1,0 +1,24 @@
+;;; early-init.el --- Spacemacs Early Init File
+;;
+;; Copyright (c) 2018 Sylvain Benner & Contributors
+;;
+;; Author: Miciah Dashiel Butler Masters <miciah.masters@gmail.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+;; Before Emacs 27, the init file was responsible for initializing the package
+;; manager by calling `package-initialize'. Emacs 27 changed the default
+;; behavior: It now calls `package-initialize' before loading the init file.
+;; This behavior would prevent Spacemacs's own package initialization from
+;; running. However, Emacs 27 also loads the "early init" file (this file)
+;; before it initializes the package manager, and Spacemacs can use this early
+;; init file to prevent Emacs from initializing the package manager. (See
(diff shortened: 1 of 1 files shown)


Miciah:
Note: I have not actually tested earlier Emacs versions with this change; any volunteers?

early-init.el
```diff
@@ -0,0 +1,13 @@
+;; Before Emacs 27, the init file was responsible for initializing the package
```
sdwolfz: You probably have to add the copyright notice at the top of this file:

```
;;; early-init.el --- Spacemacs Initialization File
;;
;; Copyright (c) 2012-2018 Sylvain Benner & Contributors
;;
;; Author: Your details here <=============================================================
;; URL: https://github.com/syl20bnr/spacemacs
;;
;; This file is not part of GNU Emacs.
;;
;;; License: GPLv3
```

See the `init.el` file for an example.

early-init.el
```diff
@@ -0,0 +1,13 @@
...
+;; manager by calling `package-initialize'.  Emacs 27 changed the default
```
sdwolfz: There are two spaces before the word `Emacs` on this line.

sdwolfz:
I've tested it by installing from source the following branches:
- emacs-25
- emacs-26
- master (with the mentioned commit)

There is no problem caused by the presence of `early-init.el`. But I also did not notice any problems by removing the `;; (package-initialize)` comment from `init.el`, not even on `emacs-25`.

early-init.el
Miciah: Good call.  Fixed.

early-init.el
Miciah: I'm a traditionalist!... Fixed.

Miciah:
Thanks for reviewing and testing!

syl20bnr:
Thank you to both of you for the PR and testing 👍 

PR #10396, opened 2018-02-27
Fix spacemacs/swiper-all-region-or-symbol function

perfectayush:
spacemacs/swiper-all-region-or-symbol function doesn't work properly. It breaks for search in other buffers. This is probably due to using old swiper api which has changed.


aaronjensen:
Confirmed that this fixes `spacemacs/swiper-all-region-or-symbol`

bmag:

PR #10406, opened 2018-02-28
Fix (void-variable hydra-ivy/keymap) issue

ezchi:
After change to by default using use-package-always-defer, the ivy layer
complain about (void-variable hydra-ivy/keymap) which should be caused the
variable is used before ivy package is loaded.

Add ":demand t" to ensure ivy package will be loaded first.


bmag:
Thank you, after looking at this PR I found a few more regressions in Ivy layer. I fixed all of them in https://github.com/syl20bnr/spacemacs/commit/f294d42ae553a1d8cdb30967935e1cf23d16e5c2, so this PR isn't needed anymore.

PR #10415, opened 2018-03-01
use-package-always-defer: fix regression in tmux

aaronjensen:


aaronjensen:
Great idea enabling `use-package-always-defer` @syl20bnr ... help force us to better habits. It also shaved some time off my startup 😄

bmag:

PR #10419, opened 2018-03-01
Add tide-fix binding to typescript layer

altschuler:
Adds the `tide-fix` function as `SPC r f` in typescript mode. It's an immensely useful function and should definitely be available.


layers/+lang/typescript/packages.el
```diff
@@ -75,6 +75,7 @@
                              "gu" tide-references
                              "hh" tide-documentation-at-point
                              "rr" tide-rename-symbol
+                             "rf" tide-fix
```
aaronjensen: I've had this bound to just "f" for a while. "r" is "rename" at the moment. You could change that to "refactor". I don't know how @syl20bnr feels about single letter commands, but I prefer just "f" because of how frequently I use it. Also, "rf" is a single finger combination which makes it harder to type.

altschuler: Good points. I would definitely rename "r" to "refactor", I didn't notice it was "rename" at the moment. I believe this is the common practice in other layers. As for the single finger point, this is true, but I'm not sure what key would be suited better while being descriptive. Maybe "x" or "i"?

aaronjensen: Any reason not to use "f"?

altschuler: Sorry, forgot to mention that. IMO it's nicer to group the refactorings, while I see the point of a quick binding. But "ri" is pretty fast and fluent, been using it for a bit now.

aaronjensen: Sounds good to me.

aaronjensen: I played w/ it a little more and I'd actually be good w/ either... rf is somewhat annoying to type, but not as much as something like rv. It may be worth it to keep the most obvious mnemonic.

syl20bnr:
Thank you ! 👍 

We already had similar functions for other languages, I added a convention to add all errors specific key bindings specfic to a major mode under `SPC m E` (`e` is already taken for `evaluation`). I did not choose `SPC m f` because it is already taken and I prefer the `error` semantic than the `fix` semantic.

Related commit: https://github.com/syl20bnr/spacemacs/commit/3731d0d3

Issue #10407, opened 2018-02-28
lsp layer  File mode specification error: (void-function lsp-ui-mode) - use-package changes?


cormacc:
This issue arises when opening any C file since a recent melpa package update.
Looks very similar to this https://github.com/syl20bnr/spacemacs/issues/10394, and can be resolved the same way (i.e. by adding :commands (lsp-ui-mode) to the use package macro invocation in lsp layer packages.el).

Haven't raised a PR yet, as this is starting to look like a systematic problem, either due to some upstream change in use-package, or something screwy with my system. At least one other individual @myrgy experienced the other issue though.

If someone can confirm it's not just me, I'll raise a PR.

System details below

#### System Info :computer:
- OS: gnu/linux
- Emacs: 25.3.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 4fe1600f)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:

```elisp
((auto-completion :variables auto-completion-enable-snippets-in-popup t auto-completion-enable-help-tooltip t)
 cquery my-code c-c++
 (cmake :variables)
 (colors :variables colors-enable-nyan-cat-progress-bar t nyan-bar-length 8)
 csv chrome dash deft emacs-lisp evil-snipe git
 (go :variables gofmt-command "goimports")
 haskell html
 (ibuffer :variables ibuffer-group-buffers-by 'projects)
 (javascript :variables node-add-modules-path t)
 (markdown :variables markdown-live-preview-engine 'vmd)
 nlinum notmuch octave
 (org :variables org-enable-hugo-support t org-enable-github-support t org-enable-reveal-js-support t)
 pandoc plantuml python racket
 (ranger :variables ranger-show-preview t ranger-cleanup-on-disable t ranger-show-dotfiles t)
 (ruby :variables ruby-test-runner 'rspec)
 rust search-engine
 (shell :variables shell-default-shell 'multi-term shell-default-term-shell "/bin/zsh" shell-default-height 30 shell-default-position 'bottom)
 shell-scripts spell-checking
 (syntax-checking :variables syntax-checking-enable-by-default nil)
 systemd treemacs version-control windows-scripts yaml)
```
[...]

PR #10420, opened 2018-03-02
lsp layer -- allow deferred loading of lsp-ui package (resolves issue #10407)

cormacc:
lsp layer was broken when use-package default changed to deferred loading

Resolved by adding a :command lsp-ui-mode parameter to the use package call.


cormacc:
Fixed upstream -- ignore this

PR #10421, opened 2018-03-02
use-package-always-defer: fix regressions in spacemacs-editing

dsedivec:
`clean-aindent-mode` was previously loaded unconditionally at init time.  This re-enables that behavior.


bmag:
Thank you, fixed this and a few others in https://github.com/syl20bnr/spacemacs/commit/f84ab9af8fb9de33ef3a4e044ade07cd082cdfc9.

Issue #1967, opened 2015-06-12
Should helm be deferred?


justbur:
I don't know about others, but the first command I want to run when I open spacemacs is almost always something that uses helm. Deferring helm introduces a small, but noticeable delay with that first command. It seems to me that if this is the common usage pattern helm should be loaded right away (despite the slightly longer load time). It was annoying me enough that I disabled the deferral mechanism for helm, and I think the experience is better.

Sorry for nitpicking.

syl20bnr:
You are right :-)

One of the explanation: in the emacs community we have an inferiority complex about loading time and helm is not very quick to load, it is even worth on a windows machine.

Another explanation is about marketing: while spacemacs is fairly optimized and the loading time is big then people will have a bad perception of it and we will see it on the screenshots.

We could make an exception for helm and simply remove the number of seconds displayed after emacs is loaded but I don't want to do that to keep a fast as possible startup time. Then the explicit metric tries to remove the biased perception by showing explicitly that even if it's not instantaneous "it's only 2 seconds" and often better in the terminal (I had sub-second loading time in the terminal with raw spacemacs). It also shows that we care about loading time and we try to optimize it as much as we can.

We can imagine a feature to disable the lazy-loading per package but it may be some work to make it slick and general enough.

justbur:
I never expected such a detailed answer :-), and I see your point. 

In that case, would changing `:defer t` to `:defer 1` (or some small number) at least bump helm up in priority so it doesn't get stuck behind other packages?

syl20bnr:
Oh `:defer n` does this ? `:defer t` packages are loaded after any `:defer n` ? Nice :-)

justbur:
I could be wrong, but here's what the documentation says 

> Defer loading of a package -- this is implied when using
>   `:commands', `:bind', `:bind*', `:mode' or `:interpreter'.
>   This can be an integer, to force loading after N seconds of
>   idle time, if the package has not already been loaded.

syl20bnr:
Ok so it is not really a priority queue, it is just a timer value :-( I don't think it has any value when the package is requested before the time out.

syl20bnr:
I have to test it, this is some kind of middle ground between user perception and marketing bullshit :-)

justbur:
Yes, not strictly speaking. I interpreted it as a hacky way to get helm to load before other packages.

justbur:
[...]

PR #10424, opened 2018-03-03
Call helm-mode when helm is initialized

bet4it:
If we don't call `helm-mode` when helm is initialized, helm's changing to functions like `completing-read-function` can't take effect before helm is lazy loaded. So if I call functions using `completing-read` like `describe-function`(<kbd>C-h f</kbd>), it won't be taken over with helm.
In commits prior to 29c78ce841ee5a99bb4077a2862afc96c8136107 ,helm was set with `:defer 1` , which means that helm will be initialized after 1 second.(#1967)
However, this way causes many confustion. Functions using `completing-read` will be taken over with helm only after emacs has idled for 1 second. #9826 comes with the same reason.
As `helm-mode` requires `helm-lib`, where `helm-current-directory` locates in, we don't need to specify it now like what #9869 do.
And I still wonder [Should helm be deferred?](https://github.com/syl20bnr/spacemacs/issues/1967)

### DIFF
--- a/layers/+completion/helm/packages.el
+++ b/layers/+completion/helm/packages.el
@@ -58,8 +58,6 @@
 
 (defun helm/init-helm ()
   (use-package helm
-    :commands (spacemacs/helm-find-files
-               helm-current-directory)
     :init
     (progn
       (add-hook 'helm-cleanup-hook #'spacemacs//helm-cleanup)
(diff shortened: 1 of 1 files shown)


syl20bnr:
Thank you for the PR. Helm takes a long time to load, what the impact on loading time when calling `helm-mode` ?

Is there a hook where we could initialize helm ? In the worst case could we advice temporarily (just for fist call, kind of trigger once advice) the completing read function to load Helm just in time ?

bet4it:
`helm-mode` will load only some parts of helm, some other parts like `helm-eshell` or `helm-color` will still be lazy loaded until they are first called.
On my computer, this change only delays the start time for less than 0.3 seconds. Most time I even can't tell the difference between them.
As spacemacs need more than 5 second to be fully loaded on my computer, I think the possible delay of this change is negligible. 
Helm are still be loaded before, but after 1 second because of `:defer 1`, most people will wait for this second although they don't know about this, while other people actioned with helm immediately get confustion like #9826. The real load time is base time + 1 second before, which will delay more time.

I think advice temporarily can solve the problem, but it's not necessary to do this.

syl20bnr:
Alright you convinced me, thank you ! 👍 

bet4it:
https://github.com/syl20bnr/spacemacs/commit/b0f77612e318b236822de940ecea7c8a43226f41#diff-fa2afbb00b304a76ad8b127262c68de4R62
We don't need it any more.

PR #10425, opened 2018-03-03
typescript: Fix eldoc

aaronjensen:
eldoc cannot be enabled until after tide-setup is run.


syl20bnr:
Thank you ! 👍 

I changed to a `use-package` hook instead so we can keep the isolation of eldoc config. Commit: https://github.com/syl20bnr/spacemacs/commit/baf154b0

aaronjensen:
Thanks for merging, but baf154b doesn't appear to actually work. It needs to run eldoc-mode in `typescript-mode-hook` after `tide-setup` is run. 

I tried doing:

```elisp
(spacemacs|use-package-add-hook tide :post-config (add-hook typescript-mode-hook 'eldoc-mode t))
```

which works, but not on the first buffer. This seems to work, however: https://github.com/syl20bnr/spacemacs/pull/10427

PR #10427, opened 2018-03-03
typescript: Fix eldoc hooks

aaronjensen:
Follow up to #10425 and baf154b

### DIFF
--- a/layers/+lang/typescript/packages.el
+++ b/layers/+lang/typescript/packages.el
@@ -30,7 +30,9 @@
     :modes typescript-mode typescript-tsx-mode))
 
 (defun typescript/pre-init-eldoc ()
-  (spacemacs|use-package-add-hook tide :post-config (eldoc-mode)))
+  (spacemacs|use-package-add-hook tide :post-init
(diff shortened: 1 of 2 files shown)


layers/+lang/typescript/packages.el
```diff
@@ -30,7 +30,9 @@
     :modes typescript-mode typescript-tsx-mode))
 
 (defun typescript/pre-init-eldoc ()
-  (spacemacs|use-package-add-hook tide :post-config (eldoc-mode)))
+  (spacemacs|use-package-add-hook tide :post-init
+                           (add-hook 'typescript-tsx-mode-hook 'eldoc-mode t)
+                           (add-hook 'typescript-mode-hook 'eldoc-mode t)))
```
aaronjensen: Can't use `spacemacs/add-to-hooks` here because we need to append. Should I add an optional append flag to `spacemacs/add-to-hooks`?

syl20bnr: Yes it would be a good addition, for now I merge your PR as it fixes my mistake.

syl20bnr:
Thank you ! 💯 

PR #10428, opened 2018-03-04
cc-mode: move c-toggle-auto-newline to hook

aaronjensen:
Without this, the first time cc-mode is required the current major-mode name
gets `//la` appended to it, which is very confusing. This will run
`c-toggle-auto-newline` when a c-mode is enabled, which seems to be more correct
any way.


syl20bnr:
Thank you again ! :-)

I made a named function to replace the lambda.

PR #10430, opened 2018-03-04
recentf : use recentf-expand-file-name to generate recentf-exclude

bet4it:
I don't know what problem #8005 has solved.
I think he has changed `recentf-filename-handlers`.
It's better to use recentf-expand-file-name to do it.


sdwolfz:
Thank you :heart:!

PR #10439, opened 2018-03-06
fix problem: c-c++-enable-clang-format-on-save not work

chenchuanyin:
fix problem: when c-c++-enable-clang-format-on-save is set in c-c++ layer, but it didn't work.

### DIFF
--- a/layers/+lang/c-c++/packages.el
+++ b/layers/+lang/c-c++/packages.el
@@ -71,6 +71,8 @@
     :if c-c++-enable-clang-support
     :init
     (progn
+      (when c-c++-enable-clang-format-on-save
+        (spacemacs/add-to-hooks 'spacemacs/clang-format-on-save c-c++-mode-hooks))
       (dolist (mode c-c++-modes)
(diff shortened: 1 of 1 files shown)


iromise:
hope it can be fixed soon....

Miciah:
This looks like a duplicate of #10847.

Compro-Prasad:
@syl20bnr @sdwolfz This PR and https://github.com/syl20bnr/spacemacs/pull/10847 seem to solve the same problem in a same way.

delaanthonio:
@eklitzke If you want auto-formatting to be a project local setting, then this change allows you to set `c-c++-enable-clang-format-on-save` as a project local variable. As you can see in the [c-c++ funcs.el](https://github.com/syl20bnr/spacemacs/blob/7c9f612c4f4a1eda7c8a11194ff2bce21ca53ca0/layers/%2Blang/c-c%2B%2B/funcs.el#L40), the files are only formatted on save when `c-c++-enable-clang-format-on-save` is true. 

@syl20bnr can you merge in this change when you have the time?

PR #10442, opened 2018-03-06
theming: Use custom-theme-set-faces

aaronjensen:
This prevents the customizations made from getting pushed to the customization section of init.el. It also obviates unsetting the customizations.


aaronjensen:
This appears to work better than the previous method, but maybe I'm missing something.

aaronjensen:
It seems like it actually overrides the theme's styling entirely, which isn't what I want. Don't merge this yet, please.

aaronjensen:
Going to close this out for now. I'm not sure of the best way to do this and it's not a huge deal.

PR #10453, opened 2018-03-08
Remove spacemacs//c-toggle-auto-newline

MaskRay:
This is an annoying option that is very hard to override on user's side.

The probably choices are:

```elisp
(add-hook 'after-change-major-mode-hook (lambda() (electric-indent-mode -1)))

;; private layer
(defun my-code/post-init--cc-mode () ... (delq 'spacemacs//c-toggle-auto-newline c-mode-common-hook)
```

Neither is obvious. The wrapper ` (c-toggle-auto-newline 1)` serves no purpose... and at least it should use 0

auto-newline is annoying because people spend more time editing existing code than adding consecutive lines. The auto added new line is oftentimes undesired rather than useful.

### DIFF
--- a/layers/+lang/c-c++/funcs.el
+++ b/layers/+lang/c-c++/funcs.el
@@ -9,10 +9,6 @@
 ;;
 ;;; License: GPLv3
 
-(defun spacemacs//c-toggle-auto-newline ()
-  "Toggle auto-newline."
-  (c-toggle-auto-newline 1))
-
 
 ;; clang
 
--- a/layers/+lang/c-c++/packages.el
+++ b/layers/+lang/c-c++/packages.el
@@ -43,8 +43,7 @@
     :init
     (progn
       (add-to-list 'auto-mode-alist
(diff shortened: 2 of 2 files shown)


eklitzke:
+1 for this change. Finding out what hook needs to be removed is a lot more difficult than adding a hook in the first place, and this is sufficiently annoying that it should not be enabled by default. If someone wants to turn it into a config option (default off) that's one thing, but setting this as the default doesn't make sense. I spend more time working on addressing feedback on my changes or making changes to existing code than I do writing long blocks of new code, and this setting interferes with that.

sdwolfz:
Burn it with :fire:! `auto-newline` has to go... it's incredibly frustrating and searching for a fix onlie gives no results.

sdwolfz:
I created a varibale that controlls wether this feature is enabled or not, default disabled. You can set it in your `.spacemacs` as a layer variable.

```emacs-lisp
(c-c++ 
  :variables
  c-c++-enable-auto-newline t)
```

Thank you :heart:!

Issue #10143, opened 2018-01-12
C-C++ layer cleanup


quicknir:
I'm glad to see that C-C++ layer has gotten some much needed TLC, especially rtags support. However, it's also become a bit of a mess with many packages supporting the same functionality, sometimes to no purpose. I've maintained my own fairly detailed C++ layer for a while, so I have some thoughts that I think could improve the layer, and also really reduce confusion for people using it.

First, is to eliminate redundant packages. In particular, there are multiple packages that all do the same exact thing as ycmd, but worse. ycmd is a really well maintained, polished project, and alternatives to error checking and completion that are a hundred lines of elisp calling to the clang binary are not going to match it. These packages (or, in some cases, they are parts of other packages like flycheck or company), are: flycheck-clang, company-clang, and company-clang-headers. I'm sure someone will argue in favor of them, but ycmd does all of what these do, and significantly better, and ycmd is far more performant on large projects. ycmd is very easy to install (it's even in some package managers). We should encourage people to use by far the best tool in the semantic completion/error checking space, simplifying the layer in the process.

Second, there are some other packages that I think can't really carry their weight:

 - srefactor: to have this work properly, you need significant additional setup so that semantic properly indexes your entire project. In addition, the semantic parser isn't very good and this isn't likely to work reliably on a modern codebase. You are much better off using rtags refactor; it only does renaming but it will actually do it reliably.
 - realgud: I've played around with this in the past, and it offers no functionality that gdb-mi doesn't already offer (it actually uses gdb-mi). It doesn't support all the windows that gdb-mi offers with multiple windows. We should only use gdb-mi, and work on making small improvements rather than bothering with realgud.
 - google style: with all respect, this really shouldn't be there. Most of us don't use google style. If you want to provide a layer variable that sets up a C style, with a default that also serves as an example, and that example happens to be google's style, I'm fine with that (and it would be very helpful, as a beginner it took me forever to figure out how to configure style).

Third, there are a lot of issues with the keybindings. All of the rtags stuff is bound under `g`, even if it makes no sense. `g` is supposed to be for "goto", things that navigate you in the code. rename-symbol, reparse buffer, fixits, etc, none of these are examples of going anywhere. Also, it is not necessary to bind every single rtags functions under the sun. Some of these are useless; `location stack backward` is already handled by `C-o`, for example. `rtags-goto-offset`. Etc. Let's look and see what's actually useful before binding it. 

[...]

PR #10454, opened 2018-03-08
c-c++: remove c-c++-enable-c++11

MaskRay:
This config variable does not make sense.

There are c++11 c++14 c++17 c++2a gnu++11 gnu++14 gnu++17 gnu++2a. `-std=c++11` will actually disable some features, for example, generic lambda.

The language option is also a per-project option, it should not be global.

https://github.com/syl20bnr/spacemacs/issues/10143

### DIFF
--- a/layers/+lang/c-c++/config.el
+++ b/layers/+lang/c-c++/config.el
@@ -39,9 +39,6 @@
   "If non-nil, automatically format code with ClangFormat on
   save. Clang support has to be enabled for this to work.")
 
-(defvar c-c++-enable-c++11 nil
-  "If non nil then c++11 related features will be enabled")
-
 (spacemacs|define-jump-handlers c++-mode)
 (spacemacs|define-jump-handlers c-mode)
 
--- a/layers/+lang/c-c++/packages.el
+++ b/layers/+lang/c-c++/packages.el
@@ -77,8 +77,6 @@
   (when c-c++-enable-clang-support
     (spacemacs|add-company-backends :backends company-clang
       :modes c-mode-common)
-    (when c-c++-enable-c++11
-      (setq company-clang-arguments '("-std=c++11")))
     (setq company-clang-prefix-guesser 'spacemacs/company-more-than-prefix-guesser)
     (spacemacs/add-to-hooks 'spacemacs/c-c++-load-clang-args c-c++-mode-hooks)))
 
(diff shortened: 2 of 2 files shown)


jackkamm:
Chiming in to add that the flycheck configuration done by `c-c++-enable-c++-11` actually has no effect -- in particular it sets `flycheck-clang-language-standard` globally but this variable only has effect when it is set buffer-locally (see  https://stackoverflow.com/a/30964293/3718509 and flycheck/flycheck#333).

So I think it a good idea to remove this configuration variable entirely, as done in this PR.

eklitzke:
This makes sense to me. I think it also makes sense to remove the `c-c++-default-mode-for-headers` config option for the same reason, as setting this globally doesn't make a lot of sense.

Compro-Prasad:
@syl20bnr This PR should be merged.

sdwolfz:
This change was introduced from the following PR: https://github.com/syl20bnr/spacemacs/pull/5654. Looking at it now, I think it was done in a rush and the intent was to have a local variable you can configure with any standard.

@MaskRay would you be willing to implement that requirement? Or do you want me to remove the current vairable and leave that for another PR?

MaskRay:
@sdwolfz Feel free to remove the variable :)

MaskRay:
Also note that GCC 6 and clang 6 default to `-std=gnu++14`. This fell out-of-fashion.

smile13241324:
@MaskRay would you mind rebasing master on your branch to fix CI? The error messages don't relate to your PR but it helps to get it merged when all the tests are successful.

MaskRay:
@smile13241324 Rebased. I'm not sure if this helps to be merged but I actually don't use it any longer.

sdwolfz:
Thank you :heart:!

tkunic:
I've been getting:

```
An error occurred while post-configuring company in layer c-c++ (error: (void-variable c-c++-enable-c++11))
```

whenever I start spacemacs for the past week. I suspect it has something to do with this change, but I'm not too familiar with debugging elisp. Is there a standard way for getting call stacks for errors like this one so I can make sure? Thanks!

sdwolfz:
Start emacs with the `--debug-init` flag and you will see a stacktrace of the error. If not, then do an unrestricted search in the `.emacs.d` directory for the variable. Also check your `.spacemacs` and your custom config file if you configured that to be in any other location.

svenpanne:
I have the same problem as @tkunic, and the reason is that a recent commit (https://github.com/syl20bnr/spacemacs/commit/445f6af93f6d5555bbda7608713bd4838c9b8274#diff-7e76133eb9f33f52192b46d5f5f0b0e3R86) references the already removed variable. :cry:  Simple fix: Just remove the buggy `when` clause (2 lines) from `+lang/c-c++/packages.el`.

sdwolfz:
Ups, that one slipped by, I always forget to escape the `+` when searching in a project. Should be fixed in 48447a6f1bf8b75d41aa3e650c87ed8b3889e294.

PR #10460, opened 2018-03-09
Make fcitx.el in Chinese layer work by default.

AmaiKinono:
- Make fcitx.el work by default. fcitx.el was not configured properly in original chinese layer.

- Corrected README.org. It is dbus interface (not fcitx-remote) that is needed for linux.

### DIFF
--- a/layers/+intl/chinese/packages.el
+++ b/layers/+intl/chinese/packages.el
@@ -32,7 +32,7 @@
     (setq fcitx-active-evil-states '(insert emacs hybrid))
     (fcitx-default-setup)
     (fcitx-prefix-keys-add "M-m" "C-M-m")
-    (if (eq system-type 'gnu/linux)
+    (if chinese-fcitx-use-dbus
         (setq fcitx-use-dbus t)))
   )
 
--- a/layers/+intl/chinese/packages.el
+++ b/layers/+intl/chinese/packages.el
@@ -27,7 +27,14 @@
 (defun chinese/init-fcitx ()
   (use-package fcitx
     :init
-    (fcitx-evil-turn-on)))
+    (fcitx-evil-turn-on)
+    :config
+    (setq fcitx-active-evil-states '(insert emacs hybrid))
+    (fcitx-default-setup)
+    (fcitx-prefix-keys-add "M-m" "C-M-m")
+    (if (eq system-type 'gnu/linux)
+        (setq fcitx-use-dbus t)))
+  )
 
 (defun chinese/init-chinese-wbim ()
   "Initialize chinese-wubi"
--- a/layers/+intl/chinese/packages.el
+++ b/layers/+intl/chinese/packages.el
@@ -32,7 +32,7 @@
     (setq fcitx-active-evil-states '(insert emacs hybrid))
     (fcitx-default-setup)
     (fcitx-prefix-keys-add "M-m" "C-M-m")
-    (if chinese-fcitx-use-dbus
+    (when chinese-fcitx-use-dbus
         (setq fcitx-use-dbus t)))
   )
 
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1103,6 +1103,11 @@ Other:
 - New packages:
   - Added package =chinese-conv= for conversion between simplified and
     traditional Chinese texts (Xiang Ji)
+- Improvements:
+  - Added variable =chinese-fcitx-use-dbus=, and fixed the misinformation about
+    dbus in README.org (thanks to AmaiKinono)
+- Fixes:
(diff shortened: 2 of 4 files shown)


layers/+intl/chinese/packages.el
```diff
@@ -26,7 +26,14 @@
...
   (use-package fcitx
     :init
-    (fcitx-evil-turn-on)))
+    (fcitx-evil-turn-on)
+    :config
+    (setq fcitx-active-evil-states '(insert emacs hybrid))
+    (fcitx-default-setup)
+    (fcitx-prefix-keys-add "M-m" "C-M-m")
+    (if (eq system-type 'gnu/linux)
+        (setq fcitx-use-dbus t)))
```
et2010: This option may lead to command lag on linux:

https://github.com/cute-jumper/fcitx.el/issues/30

AmaiKinono: I see. I made it an option by adding a layer variable. README.org is also updated.

layers/+intl/chinese/packages.el
```diff
@@ -26,7 +26,14 @@
 (defun chinese/init-fcitx ()
   (use-package fcitx
     :init
-    (fcitx-evil-turn-on)))
+    (fcitx-evil-turn-on)
+    :config
+    (setq fcitx-active-evil-states '(insert emacs hybrid))
+    (fcitx-default-setup)
+    (fcitx-prefix-keys-add "M-m" "C-M-m")
+    (if chinese-fcitx-use-dbus
```
duianto: When there's no else clause, then `if` should be changed to `when`.

duianto:
Update request:
- Rebase
- Add an entry to the end of the Chinese section of the changelog.develop file: https://github.com/syl20bnr/spacemacs/blob/develop/CHANGELOG.develop#chinese

AmaiKinono:
@duianto Hi, thanks for the code review. The requests were done, but since it's been so long, and I am really not familiar with git, so please see if anything goes wrong.

duianto:
It's fine if your not familiar with git, we can fix or suggest changes and try to help.

You'll learn more as you use git, I didn't know anything about it before starting to use Spacemacs a couple of years ago. If you haven't tried Magit I suggest you try it. It comes with the Spacemacs `git` layer and can be accessed with `SPC g s`. It's great.

### I made these changes:

### readme.org
Removed double space between `=fcitx=` and `on your machine.`
Added `the` between `use` and `dbus`.
#### Before:
>It's recommended to use dbus interface to speed up a little:

#### After:
>It's recommended to use the dbus interface to speed up a little:

#### Question 1:
What's being sped up? `fcitx` or the `dbus interface`?

If it's `fcitx`, then the sentence could read as:
>It's recommended to use the dbus interface to speed up `fcitx` a little:

If it's the `dbus interface`, then it could say:
>It's recommended to use the dbus interface to speed it up a little:

### packages.el
Since `:init` only has one command, it can be moved inline with `:init`.
```elisp
    :init (fcitx-evil-turn-on)
```

When `:config` (or `:init`) has multiple commands, then it's wrapped in a `progn`.
And the trailing parenthesis should be joined with the line above.
(An exception is at the top of the `packages.el` file where the packages are listed, that has the open and closing parentheses on their own lines.)
```elisp
    :config
    (progn
      (setq fcitx-active-evil-states '(insert emacs hybrid))
      (fcitx-default-setup)
      (fcitx-prefix-keys-add "M-m" "C-M-m")
      (when chinese-fcitx-use-dbus
        (setq fcitx-use-dbus t)))))
```


We prefer one commit per PR, so the four commits were squashed into one with this commit message:
>[chinese] Fix fcitx.el. And add a variable
>
>Make fcitx.el work by default, it was not configured properly.
>
>Added a layer variable to decide if dbus should be used or not:
>chinese-fcitx-use-dbus
>(default: nil)
>
>Corrected README.org:
>It's the dbus interface (not fcitx-remote) that is needed for linux.

#### Question 2:
Does the commit message match the meaning you had in mind?

I have these changes ready, but feel free to mention any changes/improvements you would like.

AmaiKinono:
> What's being sped up? fcitx or the dbus interface?

 It's fcitx.el that's being sped up. Setting `fcitx-use-dbus` makes fcitx.el use the functions with `-dbus` suffix, which should be faster in Linux.

> Does the commit message match the meaning you had in mind?

Yes they do.

> We prefer one commit per PR

I found that `git rebase` can squash multiple commits into one.Is the contributor suggested to do so?

> I have these changes ready

Thanks a lot!

duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/5b7b5f252643a44298f6161097fd80f73c95eeb0

PR #10461, opened 2018-03-09
Add various keycommands to markdown layer.

AmaiKinono:
Added some keycommands to markdown layer.

The original `SPC m t` (toggle) was moved to `SPC m T` because the convention says table commands should be under `SPC m t`.

Btw, I noticed that `M-RET` seems like calling `markdown-insert-list-item`. That's not right because `M-RET` is the major mode leader key in terminal. But I didn't figure out how it was binded...

### DIFF
--- a/layers/+lang/markdown/README.org
+++ b/layers/+lang/markdown/README.org
@@ -15,6 +15,7 @@
 - [[#key-bindings][Key bindings]]
   - [[#element-insertion][Element insertion]]
   - [[#element-removal][Element removal]]
+  - [[#table-manipulation][Table manipulation]]
   - [[#completion][Completion]]
   - [[#following-and-jumping][Following and Jumping]]
   - [[#indentation][Indentation]]
@@ -112,6 +113,7 @@ To generate a table of contents type on top of the buffer:
 | ~SPC m i I~ | insert reference image                                            |
 | ~SPC m i t~ | insert Table of Contents (toc)                                    |
 | ~SPC m x b~ | make region bold or insert bold                                   |
+| ~SPC m x B~ | insert gfm checkbox                                               |
 | ~SPC m x i~ | make region italic or insert italic                               |
 | ~SPC m x c~ | make region code or insert code                                   |
 | ~SPC m x C~ | make region code or insert code (Github Flavored Markdown format) |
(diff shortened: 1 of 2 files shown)


Somelauw:
In org-mode, `M-ret` was bound to `M-ret M-ret`.
Users can still disable the prefix key by 
```
dotspacemacs-emacs-leader-key nil
```
to get `M-ret` back for `markdown-insert-item`.

AmaiKinono:
@Somelauw Thanks. But what I mean is if we bind `markdown-insert-item` to a key chord, then the user should use both his major mode leader key *and* this keychord in terminal. We should rebind it to something like `M-RET M-RET`, but not disable the leader key. But I didn't figure out how it was bound, `markdown-insert-item` even doesn't appear in markdown layer, which is weird to me.

Somelauw:
> We should rebind it to something like M-RET M-RET, but not disable the leader key.

I agree, but if you are using Spacemacs in evil-mode, I recommend not using a leader key in insert mode. Just press `ESC` followed by `SPC`.

> But I didn't figure out how it was bound, markdown-insert-item even doesn't appear in markdown layer, which is weird to me.

`M-return` is probably bound in the markdown-mode, which is a major mode in Emacs. `M-m` is bound by Spacemacs. `M-return` only falls back to `M-m` if nothing is bound to `M-return` directly. An exception is made in terminals where `M-m` and `M-return` are the same key. If both are bound, the command bound to `M-m` is executed on both key presses.

To resolve, Spacemacs should probably bind both `M-m` and `M-return` in Emacs mode. Alternatively, one should use `M-m` as a leader key and `M-return` to insert a new item.

AmaiKinono:
@Somelauw Thank you! `M-RET` is indeed bound in `markdown-mode`, I see. I bound it to `SPC m M-RET` or `M-RET M-RET` so that terminal users can use it.

duianto:
Commits: https://github.com/syl20bnr/spacemacs/commit/0fcdf44b3482ff0441b440a3a9da1ac8f6e08192 https://github.com/syl20bnr/spacemacs/commit/234c7e7e10fcf199c14085ecf4056d7201009c0f https://github.com/syl20bnr/spacemacs/commit/4cb92263027b6ec7ffd49ce5bd5d2051e8fe8d21

`SPC m M-RET` and `, M-RET` works to insert a list item.
But `M-RET M-RET` doesn't work, there's an open issue about it:
Markdown: alt+enter (in holy mode) does not insert a new list item #10949

Issue #9795, opened 2017-10-28
Package wolfram-mode is unavailable


jackkamm:
#### Description :octocat:

On enabling `major-modes` layer in develop or `extra-langs` layer in master,  spacemacs will try to download and install `wolfram-mode` on startup, but is unable to find the package.

From googling, it appears that there used to be a `wolfram-mode` on melpa but it's no longer there (https://melpa.org/#/wolfram-mode).

Note, I personally don't need `wolfram-mode` and am adding it to my
`dotspacemacs-excluded-packages` until this issue is resolved
(either by the download being fixed or by the package being removed from the layer).

#### Reproduction guide :beetle:
- If on master, add `extra-langs` to `dotspacemacs-configuration-layers`
- If on develop, add `major-modes` to `dotspacemacs-configuration-layers`
- Start Emacs

*Observed behaviour:* :eyes: :broken_heart:
```
Found 1 new package(s) to install...
--> refreshing package archive: gnu... [3/3]
--> installing package: wolfram-mode@major-modes... [1/1]
Package wolfram-mode is unavailable. Is the package name misspelled?
```

*Expected behaviour:* :heart: :smile:
Spacemacs starts up without trying to download and install
a missing package.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 25.3.1
- Spacemacs: 0.200.9
- Spacemacs branch: develop (rev. e2a45e0c)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: helm
- Layers:
```elisp
(major-modes html csv yaml javascript markdown ranger bibtex latex octave
             (ess :variables ess-enable-smart-equals t)
             ipython-notebook python shell-scripts
             (shell :variables shell-default-shell 'multi-term)
             org emacs-lisp git helm my-spacemacs-layer)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GCONF GSETTINGS NOTIFY ACL GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 MODULES

duianto:
The `wolfram-mode` package was removed from MELPA 24 days ago:
[Remove recipes with malformed headers (see #5010)](https://github.com/melpa/melpa/commit/5bf94592d6e06f849b434f8da1f14c9c2fe85c6e)
[...]

PR #10464, opened 2018-03-10
Re-enable Wolfram-mode, since it is added back to Melpa.

kenkangxgwe:
Wolfram-mode was added back to Melpa, so let us enable it. See #9795.


CarlQLange:
+1, this would make me happy

sdwolfz:
Thank you :heart:!

PR #10466, opened 2018-03-11
Added dvorak right handed to keyboard layout layer

NightMachinary:
Added dvorak-right-handed to config.el and documentation of keyboard-layout layer.


duianto:
With this PR applied, when setting the configuration layer to:
```elisp
(keyboard-layout :variables kl-layout 'dvorak-right-handed)
```

A `*Warnings*` buffer opens on startup, with the message:
>Warning (initialization): An error occurred while loading ‘c:/Users/username/.emacs.d/init.el’:
>
>error: Key sequence w c c starts with non-prefix key w c
>
>To ensure normal operation, you should investigate and remove the
cause of the error in your initialization file.  Start Emacs with
the ‘--debug-init’ option to view a complete error backtrace.

The two handed `dvorak` layout does not show a warning on startup.
```elisp
(keyboard-layout :variables kl-layout 'dvorak)
```

NightMachinary:
Well, I don’t remember having this error when I wrote this PR (which was
ages ago), and with the treatment of PRs here in general, I don’t really
care to investigate it. Especially since I reached the conclusion that
learning right-handed dvorak was not worth the time.

duianto:
Then we'll close this, it can be reopened if someone requests it and a solution to the issue has been found.

PR #10471, opened 2018-03-12
Fix run-pandoc resetting settings

martian-f:
When using spacemacs/run-pandoc to open the pandoc hydra, pandoc-mode
is called each time, which will overwrite any previous changes the user
might have done to pandoc--local-settings.

Fix this by calling pandoc-mode only when it is not yet enabled.


sdwolfz:
Thank you :heart:! 

PR #10482, opened 2018-03-14
c-c++: fix setting flycheck to use c++11

jackkamm:
Under certain conditions the `c-c++` layer globally sets `flycheck-clang-language-standard` to "c++11", but this has no effect, as detailed in https://stackoverflow.com/a/30964293/3718509 and https://github.com/flycheck/flycheck/issues/333#issuecomment-34994898. This PR fixes the issue by setting `flycheck-clang-language-standard` locally via a hook, as shown in the linked examples.


MaskRay:
Can you also take a look at https://github.com/syl20bnr/spacemacs/pull/10454

c++11 c++14 c++17 c++2a

jackkamm:
@MaskRay good point,  the `c-c++-enable-c++11` is problematic as it can prevent using c++17, etc. So I withdraw this PR and support your PR to remove the variable.

PR #9317, opened 2017-07-30
Fix automatic syntax checking for cfengine

nickanderson:
The wrong mode for cfengine was specified for enabling flycheck. This properly
loads flycheck support when cfengine3 mode is active.

PR #10485, opened 2018-03-15
Flycheck is now properly enabled when loading cfengine3-mode

nickanderson:


nickanderson:
@zzamboni you might be interested in this small fix.

nickanderson:
This is a duplicate of #9317. Closing in favor of original.

Issue #9873, opened 2017-11-16
Usability issue related to "jump to definition"


sdwolfz:
Starting some time ago I noticed that when I try to jump to definition, I would get prompted to specify a `TAGS` file. This did not happen before, and although I just press enter and get the same result, this is an extra keypress I need to do every time I want to jump. 

It was an annoying thing but I kind of ignored it, I tought it was something wrong with my setup, or some glitch that might be fixed soon. But I noticed that other people had this problem as well, [here](https://www.reddit.com/r/emacs/comments/7cawkm/after_a_week_i_must_begrudgingly_admit_vscode_is/) is one example. I think I saw someone also mentioning it in an issue or PR here as well.

So I poked around a bit to find out why this was. First I inspected spacemacs jump handlers in a ruby file by doing:
```
M-; (pp spacemacs-jump-handlers)
```
Which gave me this result:
```emacs-lisp
(robe-jump evil-goto-definition dumb-jump-go)
```

I expected `robe-jump` to be the first one since it always asks to start a ruby console. I also expected `dumb-jump-go` to be a fallback. But I did not know what the deal was with `evil-goto-definition`, it seemed useless to have it since `dumb-jump` is already here.

Looks like `evil-goto-definition` is added [by this line](https://github.com/syl20bnr/spacemacs/blob/36c15b79f6d490a4733cfca3278c0c837fdc2cff/layers/%2Bdistributions/spacemacs-bootstrap/packages.el#L56). And `dumb-jump` is added at the end of the jump handler list [here](https://github.com/syl20bnr/spacemacs/blob/36c15b79f6d490a4733cfca3278c0c837fdc2cff/layers/%2Bspacemacs/spacemacs-misc/packages.el#L39).

Since this functionality was way older than the time I started getting the `TAGS` prompt, I looked inside `evil` for clues and it led me to [this line](https://github.com/emacs-evil/evil/blob/5337d2b3d2f739b4b5b4a73bc0e66363be6b71f9/evil-commands.el#L2677) which is the right age and everything made sense.

When jumping to definition, `xref-find-definitions` is called before `dubm-jump-go` and this is why the `TAGS` prompt appears. When no `TAGS` file is specified it just falls back to `dumb-jump`, which is the expected functionality, but with an extra step that's both confusing and annoying. 

< rant> 
No wonder people sell their soul to vscode, altough the jump to definition functionality works well in spacemacs, they get prompted to provide who knows what tags file like we're in the stone age or something. 
[...]

PR #10486, opened 2018-03-16
lsp-layer additions: configuration and building blocks for derived layers.

cormacc:
This adds some useful defaults to the `lsp` layer (in a new config.el), and adds some building block functions to facilitate development of derived layers.

It declares a number of major mode prefixes for typical classes of language server function, and provides the function `spacemacs/lsp-bind-keys-for-mode` which can be used by derived layers to provide a consistent set of keybindings common core language-server functions.

See README.org and the commit message for further details and a trail of the multiple amendments since initial submission.

A usage example can be found in the `lsp-c-c++` layer, in my branch of the spacemacs repo:
https://github.com/cormacc/spacemacs/tree/lsp-c-c%2B%2B-layer
N.B. The `lsp-c-c++` layer will not be the subject of a PR as is, but form the basis of a new PR against the `c-c++` layer after this gets merged. Please do not provide any feedback relating to it in the discussion thread relating to this PR.

This is partially derived from @MaskRay's personal config and has benefited from some feedback from him and quite a number of other contributors to this discussion.

### DIFF
--- a/layers/+tools/lsp/README.org
+++ b/layers/+tools/lsp/README.org
@@ -4,38 +4,111 @@
 - [[#description][Description]]
   - [[#features][Features:]]
 - [[#configuration][Configuration]]
+  - [[#derived-layers][Derived layers]]
+    - [[#spacemacslsp-bind-keys-for-mode-mode][=spacemacs/lsp-bind-keys-for-mode mode=]]
+  - [[#variables][Variables]]
+  - [[#diagnostics][Diagnostics]]
+- [[#future-additionsimprovements][Future additions/improvements]]
+  - [[#make-spacemacslsp-bind-keys-for-mode-bind-conditionally][Make =spacemacs/lsp-bind-keys-for-mode= bind conditionally]]
+- [[#references][References]]
 
 * Description
-This layer adds support for basic language server protocol packages speaking
-[[https://microsoft.github.io/language-server-protocol/specification][language server protocol]].
+  This layer adds support for basic language server protocol packages speaking
+  [[https://microsoft.github.io/language-server-protocol/specification][language server protocol]].
+
+  Different language servers may support the language server protocol to varying degrees
+  and they may also provide extensions; check the language server's website for
+  details.
+  =M-x lsp-capabilities= in a LSP buffer to list capabilities of the server.
 
 ** Features:
-- Cross references (definitions, references, document symbol, workspace symbol
-  search and others)
-- Workspace-wide symbol rename
-- Symbol highlighting
-- Flycheck
-- Completion with =company-lsp=
-- Signature help with eldoc
-- Symbol documentation in a child frame (=lsp-ui-doc=)
-
-Each language server may support the language server protocol in varying degree
-and they may also provide extensions, check the language server's website for
-details.
-=M-x lsp-capabilities= in a LSP buffer to list capabilities of the server.
+   - Cross references (definitions, references, document symbol, workspace symbol
+     search and others)
+   - Workspace-wide symbol rename
+   - Symbol highlighting
+   - Flycheck
+   - Completion with =company-lsp=
+   - Signature help with eldoc
+   - Symbol documentation in a child frame (=lsp-ui-doc=)
+   - Navigation using imenu
 
 * Configuration
-The LSP ecosystem is based on two packages: [[https://github.com/emacs-lsp/lsp-mode][lsp-mode]] and [[https://github.com/emacs-lsp/lsp-ui][lsp-ui]].
-Please check out their documentation.
+  The LSP ecosystem is based on two packages: [[https://github.com/emacs-lsp/lsp-mode][lsp-mode]] and [[https://github.com/emacs-lsp/lsp-ui][lsp-ui]].
+  Please check out their documentation.
 
-If you add =lsp-*-enable= to major mode hooks for auto initialization of
-language clients, customize =lsp-project-whitelist= =lsp-project-blacklist= to
-disable projects you don't want to enable LSP.
+  If you add =lsp-*-enable= to major mode hooks for auto initialization of
+  language clients, customize =lsp-project-whitelist= =lsp-project-blacklist= to
+  disable projects you don't want to enable LSP.
 
-If some features doe not work as intended, here is a common check list.
+** Derived layers
+   A number of elisp functions have been added to facilitate development of derived layers.
 
-- =M-x lsp-capabilities= If the LSP workspace is initialized correctly
-- =M-: xref-backend-functions= should be =(lsp--xref-backend)= for cross
-  references
-- =M-: completion-at-point-functions= should be =(lsp-completion-at-point)= for
-  completion
+*** =spacemacs/lsp-bind-keys-for-mode mode=
+    Binds keys to a number of lsp features useful for all/most modes for the given major mode.
+
+    The prefix conventions suggested in spacemacs CONVENTIONS.org have been observed, where appropriate.
+
+    The default bindings are listed below. Derived language server layers should extend this list.
+
+    | binding | function                                        |
+    |---------+-------------------------------------------------|
+    | ~m = b~ | format buffer (lsp)                             |
+    |---------+-------------------------------------------------|
+    | ~m g i~ | goto implementation                             |
+    | ~m g t~ | goto type-definition                            |
+    | ~m g a~ | goto viewport symbol (avy)                      |
+    | ~m g m~ | browse file symbols (lsp-ui-imenu)              |
+    | ~m g d~ | find definitions                                |
+    | ~m g r~ | find references                                 |
+    | ~m g s~ | find-workspace-symbol                           |
+    | ~m g p~ | jump prev (lsp-ui-peek stack - see Note 1)      |
+    | ~m g n~ | jump next (lsp-ui-peek stack - see Note 1)      |
+    | ~m g f~ | jump to flycheck error                          |
+    |---------+-------------------------------------------------|
+    | ~m h h~ | describe thing at point                         |
+    |---------+-------------------------------------------------|
+    | ~m l r~ | lsp-restart-workspace                           |
+    | ~m l a~ | execute code action                             |
+    |---------+-------------------------------------------------|
+    | ~m r r~ | rename                                          |
+    |---------+-------------------------------------------------|
+    | ~m T d~ | toggle documentation overlay                    |
+    | ~m T F~ | toggle documentation overlay function signature |
+    | ~m T s~ | toggle symbol info overlay                      |
+    | ~m T S~ | toggle symbol info overlay symbol name          |
+    | ~m T I~ | toggle symbol info overlay duplicates           |
+
+    Note 1: There is a window local jump list dedicated to cross references
+
+** Variables
+   A number of configuration variables have been exposed via the LSP layer =config.el=.
+   Sensible defaults have been provided, however they may all be overridden in your .spacemacs, or dynamically using the bindings added
+   under the derived mode t prefix by =(spacemacs/lsp-bind-keys-for-mode mode)=
+
+   | Variable name                   | Default | Description                                                                               |
+   |---------------------------------+---------+-------------------------------------------------------------------------------------------|
+   | =lsp-ui-remap-xref-keybindings= | nil     | When non-nil, xref keybindings remapped to lsp-ui-peek-find-{definition,references}       |
+   | =lsp-ui-doc-enable=             | t       | When non-nil, the documentation overlay is displayed                                      |
+   | =lsp-ui-doc-include-signature=  | nil     | When nil, signature omitted from lsp-ui-doc overlay (this is usually redundant)           |
+   | =lsp-ui-sideline-enable=        | t       | When non-nil, the symbol information overlay is displayed                                 |
+   | =lsp-ui-sideline-show-symbol=   | nil     | When non-nil, the symbol information overlay includes symbol name (redundant for c-modes) |
+   | =lsp-ui-peek-expand-by-default= | nil     | When non-nil, =lsp-ui-peek= expands file matches automatically (may degrade performance)  |
+
+** Diagnostics
+   If some features do not work as expected, here is a common check list.
+
+  - =M-x lsp-capabilities= If the LSP workspace is initialized correctly
+  - =M-: xref-backend-functions= should be =(lsp--xref-backend)= for cross
+    references
+  - =M-: completion-at-point-functions= should be =(lsp-completion-at-point)= for
+    completion
+
+* Future additions/improvements
+** Make =spacemacs/lsp-bind-keys-for-mode= bind conditionally
+   i.e. only bind keys if the language server supports the capability (queried using =lsp-capabilities=).
+   =lsp-capabilities= uses current buffer to determine the language server, so this would probably entail adding the bindings dynamically in
+   a mode hook.
+
+* References
+  - [[https://github.com/emacs-lsp/lsp-mode][lsp-mode repo]]
+  - [[https://github.com/emacs-lsp/lsp-ui][lsp-ui repo]]
--- a/layers/+tools/lsp/config.el
+++ b/layers/+tools/lsp/config.el
@@ -1,0 +1,19 @@
+;;; config.el --- Language Server Protocol Layer config file for Spacemacs
+;;
+;; Copyright (c) 2012-2018 Sylvain Benner & Contributors
+;;
+;; Author: Fangrui Song <i@maskray.me>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+;; ;; These all have toggles bound under 't' in spacemacs/lsp-define-keys-for-mode
+(defvar lsp-ui-remap-xref-keybindings nil "When non-nil, xref keybindings remapped to lsp-ui-peek-find-*")
+(defvar lsp-ui-peek-expand-by-default t "Expand lsp-ui-peek by default (may have performance implications)")
+(defvar lsp-ui-doc-enable t "Enable/disable lsp-ui-doc overlay")
+(defvar lsp-ui-doc-include-signature nil "When non-nil, type signature included in the lsp-ui-doc overlay")
+(defvar lsp-ui-sideline-enable t "Enable/disable lsp-ui-sideline overlay")
+(defvar lsp-ui-sideline-show-symbol nil "When non-nil, sideline includes symbol info (largely redundant for c modes)")  ; don't show symbol on the right of info
+(defvar lsp-ui-sideline-ignore-duplicate t "Ignore duplicates")
--- a/layers/+tools/lsp/funcs.el
+++ b/layers/+tools/lsp/funcs.el
@@ -1,4 +1,4 @@
-;;; packages.el --- Language Server Protocol functions File for Spacemacs
+;;; funcs.el --- Language Server Protocol Layer functions file for Spacemacs
 ;;
 ;; Copyright (c) 2012-2018 Sylvain Benner & Contributors
 ;;
@@ -46,3 +46,98 @@ https://github.com/emacs-lsp/lsp-javascript/issues/9#issuecomment-379515379"
       (all-completions (company-grab-symbol) candidates)))
   (make-local-variable 'company-transformers)
   (add-to-list 'company-transformers 'lsp-prefix-company-transformer))
+
+(defun spacemacs/lsp-bind-keys-for-mode (mode)
+  "Define key bindings for the specific MODE."
+  (spacemacs/declare-prefix-for-mode mode "m=" "format")
+  (spacemacs/declare-prefix-for-mode mode "mg" "goto")
+  (spacemacs/declare-prefix-for-mode mode "mh" "hierarchy")
+  (spacemacs/declare-prefix-for-mode mode "ml" "lsp/backend")
+  (spacemacs/declare-prefix-for-mode mode "mr" "refactor")
+  (spacemacs/declare-prefix-for-mode mode "mT" "toggle")
+
+  (spacemacs/set-leader-keys-for-major-mode mode
+    ;;format
+    "=b" #'lsp-format-buffer
+    ;;goto
+    "gi" #'lsp-goto-implementation
+    "gt" #'lsp-goto-type-definition
+    "ga" #'spacemacs/lsp-avy-document-symbol
+    "gm" #'lsp-ui-imenu
+    "gd" #'lsp-ui-peek-find-definitions
+    "gr" #'lsp-ui-peek-find-references
+    "gs" #'lsp-ui-peek-find-workspace-symbol
+    "gp" #'lsp-ui-peek-jump-backward
+    "gn" #'lsp-ui-peek-jump-forward
+    "gf" #'lsp-ui-flycheck-list
+    ;;hierarchy
+    "hh" #'lsp-describe-thing-at-point
+    ;;jump
+    ;;lsp/backend
+    "la" #'lsp-execute-code-action
+    "lr" #'lsp-restart-workspace
+    ;;refactor
+    "rr" #'lsp-rename
+    ;;toggles
+    "Td" #'lsp-ui-doc-mode
+    "Ts" #'lsp-ui-sideline-mode
+    "TF" #'spacemacs/lsp-ui-doc-func
+    "TS" #'spacemacs/lsp-ui-sideline-symb
+    "TI" #'spacemacs/lsp-ui-sideline-ignore-duplicate
+    )
+
+  )
+
+
+(defun spacemacs/lsp-ui-doc-func ()
+  "Toggle the function signature in the lsp-ui-doc overlay"
+  (interactive)
+  (setq lsp-ui-doc-include-signature (not lsp-ui-doc-include-signature)))
+
+(defun spacemacs/lsp-ui-sideline-symb ()
+  "Toggle the symbol in the lsp-ui-sideline overlay.
+(generally redundant in C modes)"
+  (interactive)
+  (setq lsp-ui-sideline-show-symbol (not lsp-ui-sideline-show-symbol)))
+
+(defun spacemacs/lsp-ui-sideline-ignore-duplicate ()
+  "Toggle ignore duplicates for lsp-ui-sideline overlay"
+  (interactive)
+  (setq lsp-ui-sideline-ignore-duplicate (not lsp-ui-sideline-ignore-duplicate)))
+
+;; Used for lsp-ui-peek-mode, but may be able to use some spacemacs fn. instead?
+(defun spacemacs/lsp-define-key (keymap key def &rest bindings)
+  "Define multiple key bindings with KEYMAP KEY DEF BINDINGS."
+  (interactive)
+  (while key
+    (define-key keymap (kbd key) def)
+    (setq key (pop bindings)
+      def (pop bindings))))
+
+;; From https://github.com/MaskRay/Config/blob/master/home/.config/doom/autoload/misc.el#L118
+;;;###autoload
+(defun spacemacs/lsp-avy-document-symbol ()
+  (interactive)
+  (let (ranges point0 point1 (line 0) (col 0) (w (selected-window)) candidates)
+    (save-excursion
+      (goto-char 1)
+      (dolist (loc
+               (lsp--send-request (lsp--make-request
+                                   "textDocument/documentSymbol"
+;;;;;; I added :all t in ccls to return all symbols in the document
+                                   `(:textDocument ,(lsp--text-document-identifier) :all t))))
+        (let ((range (->> loc (gethash "location") (gethash "range"))))
+          (-let* [((&hash "line" l0 "character" c0) (gethash "start" range))
+                  ((&hash "line" l1 "character" c1) (gethash "end" range))]
+            (when (or (< line l0) (and (= line l0) (<= col c0)))
+              (forward-line (- l0 line))
+              (forward-char c0)
+              (setq point0 (point))
+              (forward-line (- l1 l0))
+              (forward-char c1)
+              (setq point1 (point))
+              (setq line l1 col c1)
+              (push `((,point0 . ,point1) . ,w) candidates))))))
+    (avy-with avy-document-symbol
+      (avy--process candidates
+                    (avy--style-fn avy-style)))))
--- a/layers/+tools/lsp/packages.el
+++ b/layers/+tools/lsp/packages.el
@@ -1,4 +1,4 @@
-;;; packages.el --- Language Server Protocol packages File for Spacemacs
+;;; packages.el --- Language Server Protocol Layer packages file for Spacemacs
 ;;
 ;; Copyright (c) 2012-2018 Sylvain Benner & Contributors
 ;;
@@ -11,15 +11,26 @@
 
 (defconst lsp-packages
   '(
-    (company-lsp :requires company)
-    ;; `flycheck-lsp' does not exist so we defined it as built-in to avoid
-    ;; fetching it from ELPA repositories.
-    ;; this logical package serves to hook all flycheck related configuration
-    ;; for LSP.
-    (flycheck-lsp :requires flycheck :location built-in)
-    lsp-mode
-    lsp-ui
-    ))
+     lsp-mode
+     lsp-ui
+     (lsp-imenu :requires imenu :location built-in)
+     (lsp-ui-imenu :requires lsp-imenu :location built-in)
+     (company-lsp :requires company)
+     ;; `flycheck-lsp' does not exist so we defined it as built-in to avoid
+     ;; fetching it from ELPA repositories.
+     ;; this logical package serves to hook all flycheck related configuration
+     ;; for LSP.
+     (flycheck-lsp :requires flycheck :location built-in)
+     ))
+
+(defun lsp/init-lsp-mode ()
+  (use-package lsp-mode
+    :defer t
+    :config
+    (progn
+      (spacemacs|hide-lighter lsp-mode)
+      (evil-set-command-property 'lsp-goto-type-definition :jump t)
+      (evil-set-command-property 'lsp-goto-implementation :jump t))))
 
 (defun lsp/init-company-lsp ()
   (use-package company-lsp
@@ -28,19 +39,12 @@
     ;; Language servers have better idea filtering and sorting,
     ;; don't filter results on the client side.
     (setq company-transformers nil
-          company-lsp-async t
-          company-lsp-cache-candidates nil)))
+      company-lsp-async t
+      company-lsp-cache-candidates nil)))
 
 (defun lsp/init-flycheck-lsp ()
   ;; Disable lsp-flycheck.el in favor of lsp-ui-flycheck.el
   (setq lsp-enable-flycheck nil))
-
-(defun lsp/init-lsp-mode ()
-  (use-package lsp-mode
-    :defer t
-    :config
-    (progn
-      (spacemacs|hide-lighter lsp-mode))))
 
 (defun lsp/init-lsp-ui ()
   (use-package lsp-ui
@@ -49,5 +53,26 @@
     :config
     (progn
       (spacemacs//lsp-sync-peek-face)
-      (add-hook 'spacemacs-post-theme-change-hook
-                #'spacemacs//lsp-sync-peek-face))))
+      (add-hook 'spacemacs-post-theme-change-hook #'spacemacs//lsp-sync-peek-face)
+
+      (if lsp-ui-peek-expand-by-default
+        (setq lsp-ui-peek-expand-function (lambda (xs) (mapcar #'car xs))))
+
+      (if lsp-ui-remap-xref-keybindings
+        (progn (define-key lsp-ui-mode-map [remap xref-find-definitions] #'lsp-ui-peek-find-definitions)
+          (define-key lsp-ui-mode-map [remap xref-find-references] #'lsp-ui-peek-find-references)))
+
+      (spacemacs/lsp-define-key
+        lsp-ui-peek-mode-map
+        "h" #'lsp-ui-peek--select-prev-file
+        "j" #'lsp-ui-peek--select-next
+        "k" #'lsp-ui-peek--select-prev
+        "l" #'lsp-ui-peek--select-next-file
+        )
+      )))
+
+(defun lsp/init-lsp-imenu ()
+  (use-package lsp-imenu :defer t :init (add-hook 'lsp-after-open-hook #'lsp-enable-imenu)))
+
+(defun lsp/init-lsp-ui-imenu ()
+  (use-package lsp-ui-imenu :defer t :config (evil-make-overriding-map lsp-ui-imenu-mode-map)))
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -93,7 +93,10 @@ This file containes the change log for the next major version of Spacemacs.
 - Add support for multiple backends. Supported backends are: =megahnada=,
   =eclim= and =ensime=. The default backend is =meghanada=.
 - ~SPC m e e~ is now to fix error around point. Use ~SPC e~ prefix to navigate
-  between errorsl
+  between errors.
+**** LSP
+- Added function =spacemacs/lsp-bind-keys-for-mode= to provide a consistent set of core keybindings across LSP layers.
+- Added a number of =lsp-ui= configuration variables to the layer - see [[file:./layers/+tools/lsp/README.org][LSP layer README]] for details.
 **** Org
 - Add package =org-journal= (Nick Anderson)
 - Move clock related key bindings to ~SPC a o k~
--- a/layers/+spacemacs/spacemacs-editing/packages.el
+++ b/layers/+spacemacs/spacemacs-editing/packages.el
@@ -55,7 +55,7 @@
 (defun spacemacs-editing/init-avy ()
   (use-package avy
     :defer t
-    :commands (spacemacs/avy-open-url spacemacs/avy-goto-url avy-pop-mark)
+    :commands (spacemacs/avy-open-url spacemacs/avy-goto-url avy-pop-mark avy-with)
     :init
     (progn
       (setq avy-all-windows 'all-frames)


layers/+tools/lsp/funcs.el
```diff
@@ -30,3 +30,79 @@
...
+  (spacemacs/declare-prefix-for-mode mode "mg" "goto")
+  (spacemacs/declare-prefix-for-mode mode "ml" "lsp")
+  (spacemacs/declare-prefix-for-mode mode "mr" "refactor")
+  (spacemacs/declare-prefix-for-mode mode "mt" "toggle")
+
+  (spacemacs/set-leader-keys-for-major-mode mode
+    ;;Format
+    "==" #'spacemacs/lsp-format-buffer
+    ;;goto
+    "gi" #'imenu
```
MaskRay: `imenu` is already bound at `SPC s j`

layers/+tools/lsp/funcs.el
```diff
@@ -30,3 +30,79 @@
...
+    "gs" #'lsp-ui-find-workspace-symbol
+    "gf" #'spacemacs/lsp-ffap-no-prompt
```
MaskRay: ffap does not use LSP.

MaskRay:
Great job! Hope this can be merged soon.

layers/+tools/lsp/funcs.el
cormacc: Also removed. Originally included that a binding like this was appropriate in a general purpose lsp layer, but maybe there's a better idiomatic equivalent, i.e an lsp method that would jump to a header file when invoked on a c source #include, or the appropriate ruby source file when invoked on a  'require...' statement etc.?

cormacc:
Thanks Maskray -- amended the commit with your suggestions -- also some corrections to keybindings and file headers in line with CONVENTIONS.org.

You got any insight into the failing CI tests? They're all the same issue. Probably something daft I've done..

>Test test-declare-layers--distribution-layer-is-second condition:
>   (ert-test-failed
>     ((should
>       (eq 'spacemacs-base
>	    (second configuration-layer--used-layers)))
>      :form
>      (eq spacemacs-base spacemacs-defaults)
>      :value nil))

layers/+tools/lsp/funcs.el
cormacc: Removed...

MaskRay:
No idea why CI fails... all recent PRs fail..

Regarding cquery, if the point is on a `#include` directive, you can invoke textDocument/definition (xref-find-definition, lsp-ui-peek-find-definitions, ...) to jump to the included file.

Implementation https://github.com/cquery-project/cquery/blob/master/src/messages/text_document_definition.cc#L116
Other cross reference candies https://github.com/cquery-project/cquery/wiki/FAQ#definitions

sebastiencs:
I replied to @MaskRay in the comments, sorry for being late

MaskRay:
I cannot see your comments in the web  view but I remember I saw it somewhere...

For lsp-ui-peek performance issue, code bases of most languages wouldn't be great enough to have more than 1000 references and make Emacs slow. I think unfolding all files is a good default value.

sebastiencs:
@MaskRay The issue is not the number of references but the number of files to open.
For each file it will call `font-lock-ensure` to fontify the whole file. With 10 big files it can takes +5 seconds to just open the peek view.

bkchr:
@cormacc you are missing the `spacemacs/lsp-format-buffer` function. You declare a shortcut, but the function does not exist :)

cormacc:
@bkchr Thanks -- corrected in an amendment to the commit (stripped the incorrect 'spacemacs/' prefix, which I must have added at some stage while refactoring the various changes between this PR and the new cquery layer). I haven't actually been using this keybinding, as believe cquery (the only lsp server I'm using) only includes the functionality if built using the --use-clang-cxx option, which the aur cquery-git package doesn't do out of the box.

https://github.com/cquery-project/cquery/wiki/Formatting
https://github.com/cquery-project/cquery/wiki/Build-%28Waf%29

bkchr:
@cormacc Yeah thanks :) I'm using your changes with the rust language server and I wondered why it does not work. Today I finally got time to look into it :D

cormacc:
Saw a recent commit to lsp-mode added `lsp-goto-type-definition` and `lsp-goto-implementation` so added bindings for these. Moved all the symbol browser type bindings under 'j' (jump), as 'g' was getting fairly crowded. So loose rule is that anything with a single target should go under g, whereas those with multiple targets (lsp-ui-imenu, lsp-ui-peek bindings) goes under j.

Amended my personal `new-cquery-layer` branch in the same way.
https://github.com/cormacc/spacemacs/tree/new-cquery-layer

@bkchr I've also raised a new PR against lsp-mode that adds a capability check to the lsp-format-buffer call. If that gets merged, you'll see an appropriate signal if the language server doesn't support the functionality.

franksn:
@cormacc can you set default cquery-sem-highlight-method to nil? 
As people won't likely use it unless their theme support it, and the default spacemacs-theme hasn't supported it either.

cormacc:
@franksn Semantic highlighting appears to me to work using the standard spacemacs theme -- at least with mode set to either 'font-lock or 'overlay I get colour coded tokens etc. -- or have I set my expectations too low? :)

cormacc:
@shwsun I've a simple cquery layer ready for a PR if this one gets merged. See top post. You can just clone the linked branch if you want to check it out.

cormacc:
Rebased this and new-cquery-layer on 6905d9cafae6d5ab67827308fa7dfca64ddfd213

shwsun:
Hi @cormacc , do you use the lsp + cquery in your spacemacs?  I just cloned your branch and tested it on my machine. It doesn't seem to be working properly tho. (I only tested keybindings in either layer)

cormacc:
Hi @shwsun 

Yep I do -- not extensively the last few weeks as I've been dragged into some IT governance stuff around GDPR, but checked again there and it seems to be working fine on one machine running arch linux and another running osx. You sure you checked out the latest commit of the 'new-cquery-layer' branch? I had another 'cquery-layer' branch where I did some initial work, but then realised the changes were best split between lsp-layer and a new cquery layer. Just deleted that branch to reduce potential for confusion.

I've periodically been rebasing my lsp-layer-improvements branch on upstream develop, and then rebasing new-cquery-layer on that, which may be causing some pain.

My c config currently looks like this (though you can leave out the explicit loading of the lsp-layer if you're not customising):

```
       ;; lsp  and c-c++ layers implicitly loaded by cquery layer...
       (lsp :variables
         lsp-ui-peek-expand-by-default t)
       (cquery :variables
         cquery-sem-highlight-method 'overlay)
       cmake
```

You do of course need to have cquery available in your path, and have generated a compile_commands.json in your project root.

shwsun:
@cormacc   **IT WORKS!** I just tried a vanilla spacemacs and it works fine. So there might be something weird in my config file :(
Thank you very much for the effort!

In case someone else want to try it, you can use the combined branch I worked on (see [here](https://github.com/shwsun/spacemacs/tree/lsp-improve-and-cquery))
```sh
git clone https://github.com/shwsun/spacemacs.git -b lsp-improve-and-cquery ~/.emacs.d
```
You then need to add the config posted above and compile a `compile_commands.json`.

I do have some feedback tho:
* apparently there are bunch of lsp functionalities in the [cquery](https://github.com/cquery-project/cquery/wiki/Emacs), are they all integrated in the new lsp layer? And if not maybe we can work together to make it happen?
* it looks like cquery folks tried to have ivy and helm abstracted from some lsp functionalities, can we also have that in new lsp layer (i.e. make the layer smart enough to toggle `ivy-xref` or `helm-xref`)?
* can we enable company mode?

@syl20bnr Do you have some other feedback on this? I do think a better lsp layer will avoid other heavy lifting someone has to do for Rust or other languages.

torpak:
This is the best c++ editing environment ever! I can't wait for this to be merged!
Also, having the same LSP layer with the same (or similar) keyboard commands and UI-Elements for rust ( and other languages) would be awesome too.

myrgy:
@cormacc  Thanks a lot!

layers/+tools/lsp/packages.el
```diff
@@ -1,4 +1,4 @@
-;;; packages.el --- Language Server Protocol packages File for Spacemacs
+;;; packages.el --- lsp Layer packages File for Spacemacs
```
robbyoconnor: Please revert.

layers/+tools/lsp/funcs.el
```diff
@@ -1,4 +1,4 @@
-;;; packages.el --- Language Server Protocol functions File for Spacemacs
+;;; funcs.el --- lsp Layer funcs File for Spacemacs
```
robbyoconnor: Please revert -- keep the change to `funcs.el` -- but Language Server Protocol should be fully spelled out.

layers/+tools/lsp/config.el
```diff
@@ -0,0 +1,19 @@
+;;; config.el --- lsp Layer config File for Spacemacs
```
robbyoconnor: Language Server Protocol layer.

cormacc:
@shwsun 
Re. the keybinding warning -- can't reproduce it here. Seems to be the , r r (refactor/rename) binding, but I don't see any similar warning in my messages or lsp-cquery buffers when I try it?
Re. initial delay -- if you've copied my config with `lsp-ui-peek-expand-by-default t` it could be that -- see @sebastiencs comments earlier in this thread.
Re. portable dumper error, I must admit to not even knowing what the portable dumper is, though did see it referenced in some recent commits -- I'm clearly out of touch :)

@robbyoconnor Amended the file headers. Also rebased this PR on current development tip. Rebased my personal cquery layer branch as well.

shwsun:
@cormacc **I just rebased my testing branch and it is working fine after the changes.**

RE:
* **keybinding error:** that has something to do with my spacemacs config. I will figure it out myself.
* **portable dumper:** that is a brand new feature that allow you to start Emacs in 0.5 sec by using a snapshot. Since it is still experimental even in Emacs 27 I don't think we need to worry about it now. 

I give a +1 to merge this PR.

shwsun:
I think it is worth noting that the existing lsp backend for python could use some improvement too -- I don't think the lsp in python can `find-reference` and `find-definition` correctly right now? (see [a good pyls example in emacs](https://vxlabs.com/2018/06/08/python-language-server-with-emacs-and-lsp-mode/))

A better defined LSP layer could be a good place to standardize the functionalities should be supported for different languages: `Rust` and `Java` etc

CestDiego:
lsp ui is bringing these weird behavior to my emacs:
image

the thing to the right is part of the Emacs window aparently.

CestDiego:
Also. Do any of you also see that when using `gd` both Help and peek ui gets summoned? is there a way around this? other than using `,g`? 
image

cormacc:
Rebased this and new-cquery-layer on current dev tip d28990537fca5266095e442e3806bcb2edb7f32e
CircleCI builds failing again, but I'm guessing this is due to some higher level restructuring and will conveniently ignore until someone tells me otherwise :)

cormacc:
... though maybe don't pull the rebased branches for now, as an upstream change to lazy-load evil surround appears to have broken org-mode (nooooo!). I'll rebase again when that gets fixed.

https://github.com/syl20bnr/spacemacs/issues/10895

cormacc:
Rebased on f1f8db002af108f58c9a78c5286e8beca59af6de a few days ago -- i.e. after #10895 got fixed

cormacc:
This has been sitting here for quite a while. Is the issue that it's annoying to evalute in the absence of a layer using the new functionality, in which case I can replace it with a PR for the cquery layer incorporating these changes? Or just that the core maintainers are busy (there seems to be quite a lot of restructuring work going on, which I'm sure is taking up most of their bandwidth). Perhaps one of the maintainers could indicate one way or the other?

yyoncho:
Looks great! I have missed this pull request and I have created separate pull request for LSP Java - https://github.com/syl20bnr/spacemacs/pull/10820
Few additions:
"mhh" - `lsp-describe-thing-at-point` 
"mqr" - `lsp-workspace-restart`
"mea" - `lsp-execute-code-action`
I would suggest moving `lsp-rename` to `rrs` standing for Refactoring -> Rename Symbol. This will be the same mapping as Clojure layer and will provide option derived modes to add more `rename` operations, e. g. `Rename file` will be `rrf`.

myrgy:
Symbol rename is more frequent operation comparing to other refactor actions. I think we should have shorter shortcut for that. 
For example:
 rr - refactor symbol 
 ro ... - refactor other / rename file....etc
Otherwise we will have too much redundant keys just to reach 3-d level menu item.

yyoncho:
@myrgy Your point makes sense even I was using the same shortcut as the one in this pull request. When I was writing the LSP Java layer I realized that having 3 stroke shortcuts won't be future proof since potentially there could be a lot of refactoring(at this point JDT LSP server have more than 20). I am not insisting on getting in this direction but IMO this is a point to consider.

You may take a look at Clojure mappings: https://github.com/syl20bnr/spacemacs/tree/master/layers/%2Blang/clojure#refactoring

MaskRay:
How do you like the idea of `textDocument/documentSymbol` (with a tweak: displaying all symbols instead of just outline) combined with avy

```elisp
;;;###autoload
(defun +my/avy-document-symbol ()
  (interactive)
  (let (ranges point0 point1 (line 0) (col 0) (w (selected-window)) candidates)
    (save-excursion
      (goto-char 1)
      (dolist (loc
               (lsp--send-request (lsp--make-request
                                   "textDocument/documentSymbol"
;;;;;; I added :all t in ccls to return all symbols in the document
                                   `(:textDocument ,(lsp--text-document-identifier) :all t))))
        (let ((range (->> loc (gethash "location") (gethash "range"))))
          (-let* [((&hash "line" l0 "character" c0) (gethash "start" range))
                  ((&hash "line" l1 "character" c1) (gethash "end" range))]
            (when (or (< line l0) (and (= line l0) (<= col c0)))
              (forward-line (- l0 line))
              (forward-char c0)
              (setq point0 (point))
              (forward-line (- l1 l0))
              (forward-char c1)
              (setq point1 (point))
              (setq line l1 col c1)
              (push `((,point0 . ,point1) . ,w) candidates))))))
    (avy-with avy-document-symbol
      (avy--process candidates
                    (avy--style-fn avy-style)))))
```

[demo](https://ptpb.pw/JkjK.gif)

myrgy:
@MaskRay, looks awesomw!  is it possible to integrate with helm?

yyoncho:
@MaskRay maybe using `imenu-create-index-function` instead of calling directly `documentSymbol` and this function will work in any imenu supported buffer.

myrgy:
@yyoncho, thank you. I took a look. Still its not clear why find is inside refactoring because we have lsp+xref. And from my expirience symbol rename is more frequent operation. All other (file rename, code teansformations, etc ) should be in sub menu.

cormacc:
I've rebased the branch, added the bindings suggested by @yyoncho (though left the refactor/rename as is as per subsequent discussion). Also added the avy keyboard navigation function suggested by @maskray (bound under m j a) and updated the default lsp-ui-peek mode as per his suggestion.

The avy function required an update to avy initialisation in spacemacs-editing/packages.el to load avy when avy-with invoked.

Might take a look at using lsp--capability to only create bindings for functionality that's supported by the implementation. lsp-mode uses this to provide a message when the function is invoked, but cleaner not to create the binding in the first place. Does this sound sensible or otherwise? Would require the language server to be running when creating the bindings...

yyoncho:
>  Does this sound sensible or otherwise? Would require the language server to be running when creating the bindings...

Also, there might be servers with different capabilities running at the same time.

cormacc:
@CestDiego (Better late than never) your second issue looks like a manifestation of upstream issue #9873 -- you can work around it by adding the following to your dotspacemacs/user-config...
```
  (setq spacemacs-default-jump-handlers
    (remove 'evil-goto-definition spacemacs-default-jump-handlers))
```
No idea re. the first issue. What platform you using (os, emacs version, window manager etc.)? I'm running arch linux/i3 as a general rule. Have windows and osx machines I can test on, but don't use regularly.

cormacc:
@CestDiego I spoke to soon. The default-jump-handlers seem to be called in order for some reason, so the fix cited above doesn't work alone. I've just removed the call to spacemacs//setup-lsp-jump-handler, which means that m g g now works (without calling lsp-peek-ui-find-definitions, which is bound under m j d anyway).

I've also renamed the cquery layer to lsp-c-c++ and added a new config variable to enable/disable rainbow semantic highlighting (disabled by default)

layers/+tools/lsp/README.org
```diff
@@ -4,38 +4,99 @@
...
+    | ~m T d~ | toggle documentation overlay                    |
+    | ~m T F~ | toggle documentation overlay function signature |
+    | ~m T s~ | toggle symbol info overlay                      |
+    | ~m T S~ | toggle symbol info overlay symbol name          |
+    | ~m T I~ | toggle symbol info overlay duplicates           |
+
+** Variables
+   A number of configuration variables have been exposed via the LSP layer =config.el=.
+   Sensible defaults have been provided, however they may all be overridden in your .spacemacs, or dynamically using the bindings added
+   under the derived mode t prefix by =(spacemacs/lsp-define-keys-for-mode mode)=
```
JustAPerson: the actual function name is `spacemacs/lsp-bind-keys-for-mode`

yyoncho:
The following lines are needed to make `C-o` work with type definition/implementation.
```elisp
  (evil-set-command-property 'lsp-goto-type-definition :jump t)
  (evil-set-command-property 'lsp-goto-implementation :jump t)
```

layers/+tools/lsp/funcs.el
```diff
@@ -46,3 +46,98 @@ https://github.com/emacs-lsp/lsp-javascript/issues/9#issuecomment-379515379"
...
+    ;;hierarchy
+    "hh" #'lsp-describe-thing-at-point
+    ;;jump
+    "ja" #'spacemacs/lsp-avy-document-symbol
+    "ji" #'lsp-ui-imenu
+    "jd" #'lsp-ui-peek-find-definitions
+    "jr" #'lsp-ui-peek-find-references
+    "js" #'lsp-ui-peek-find-workspace-symbol
+    ;;quit
+    "qr" #'lsp-restart-workspace
```
myrgy: (spacemacs/declare-prefix-for-mode mode "mq" "lsp")

cormacc:
@myrgy Rebound that function under mlr and declared the prefix as 'lsp' as suggested
@yyoncho Added those two lines in the use-package config for lsp-mode, which may be (a) the wrong place and (b) a bad idea as I'm not sure what problem they're addressing :) Is C-o a holy mode or hybrid-mode binding? 

Also moved the lsp-c-c++ layer into this branch as I've spent too much time switching branches, syncing upstream, rebasing this on develop and the lsp-c-c++ layer on that, then realising I've made a formatting error in lsp layer README and starting again :) If this ever looks like getting merged and the maintainers want them split, I'll do it in a final commit amendment.

yyoncho:
@cormacc - `C-o` **evil-jump-back** . If command is not marked with 'jump **evil-jump-back** won't work. Here it is another example https://github.com/syl20bnr/spacemacs/blob/master/core/core-jump.el#L77

layers/+lang/lsp-c-c++/funcs.el
```diff
@@ -0,0 +1,71 @@
...
+      'write "textDocument/references"
+      (plist-put (lsp--text-document-position-params) :context
+        '(:role 16))))
+
+  )
+
+(defun cquery/define-keys-for-mode (mode)
+  "Define key bindings for the specific MODE."
+
+  (spacemacs/set-leader-keys-for-major-mode mode
```
myrgy: `cquery-freshen-index binding` is missed. This call is usedful in case of branch switch.

cormacc: Thanks @myrgy -- bound under `m l f`
Also added cquery-preprocess-file under `m l p`

Glaeqen:
I don't know what I'm doing wrong bo no new keybindings appear under `,` 😢

cormacc:
@Glaeqen some detail as to what you've done might facilitate a diagnosis :)
Did you add `lsp-c-c++` to your spacemacs configuration layers? The modifications in this PR won't rebind keys for any other lsp layer unless the layer is modified to call `spacemacs/lsp-bind-keys-for-mode`

Glaeqen:
@cormacc I've tried to debug (my first time with elisp code debugging). I get `Error (use-package): cquery/:config: Key sequence r r starts with non-prefix key r`. Apparently, it breaks in `spacemacs/lsp-bind-keys-for-mode mode` function in `packages.el:28` . 
Full backtrace: https://gist.github.com/Glaeqen/600552f96a953d17c0157cc1afebe3ce

Thank you for your effort in making this layer and helping out!

cormacc:
@Glaeqen I've only figured out enough elisp to get by myself, but that sounds like a keybinding conflict -- i.e. you may have a function bound to `m r` before we try to use it as a prefix. Either in your user config or a binding conflict with another layer, so I'd check the following:
0. Try help describe-bindings (`SPC h d b`) to see what function is bound under `m r` -- this should help identify which of the following steps to try first
1. See if you've defined any custom c/c++ mode bindings under `m r`
2. If you've enabled rtags or cscope support or anything like that for c-c++ mode, it's possible (or even likely) that it's causing the conflict. The lsp-c-c++ layer implicitly loads the c-c++ layer, so you can either comment out that entry in your spacemacs init altogether, or leave it in place and strip any references to cscope/rtags whatever. lsp/cquery is an alternative to rtags/cscope.
3. If neither of those suggestions gets you anywhere, try disabling most of your other layers temporarily and restarting emacs. Assuming that gets rid of the problem, you can re-introduce your other layers a few at a time to isolate the source of the conflict.
4. If none of that gets you anywhere, I'm out of ideas :)

Glaeqen:
@cormacc My setup is pretty default when it comes to spacemacs. I bisected layers, it appears `semantic` layer with srefactor binding at `, r` was an issue which conflicted with `, r r` rename lsp command from lsp layer.

Glaeqen:
Now, when I know what's up, I realize warning message was pretty straightforward. I'm dumb

Glaeqen:
I don't think that @syl20bnr will approve layer which conflicts with native spacemacs' layer - semantic. What do you think? @cormacc 
This PR hangs here for a while now, so I'm wondering how does its current situation look like. I think that having a separate lsp-c-c++ layer is awkward, when we have lsp *and* c-c++ layers in spacemacs already. Someone new might find it confusing. I'll try to fork it and try to do something cool maybe.

yyoncho:
@Glaeqen IMO semantic layer must be fixed since having "mr" is the standard for refactoring.

cormacc:
I don't use the semantic layer, so hadn't come across this. Moved the srefactor-refactor-at-point binding from `m r` to `m r p` for those who do enable it.

@Glaeqen the lsp layer is supposed to provide a common basis (and maybe consistent keybindings) for all language-specific layers with lsp-compliant backends. Most users shouldn't need to add it directly to their spacemacs config unless they want to customise some config option.

The c-c++ layer has been around a long time and there's an existing issue around some ongoing tidy-up efforts. See #10143. This layer builds on top of the c-c++ layer, but has been kept separate so as not to muddy the waters unnecessarily.

Glaeqen:
@cormacc Thanks for link to issue report. Keeping lsp-c-c++ separate seems reasonable choice then. At least until c-c++ layer situation clarifies.

cormacc:
@Glaeqen Re. refactoring I'd say it would be more useful to provide that feedback upstream, either against emacs-cquery or cquery itself, once you identify which is problematic. The lsp-c-c++ layer just adds a few relevant keybindings to spacemacs config. Similarly with ccls -- I may update the layer to support this as an alternate backend to cquery via the emacs-ccls package at some point if I get a bit of time -- wouldn't have expected it to work at all configured as the cquery executable for emacs-cquery.

Anyway, this is all getting a bit OT -- just added lsp-c-c++ to this PR as a usage example, as hard to evaluate the proposed lsp layer changes in isolation. Perhaps I should withdraw the PR altogether -- I'd welcome some feedback from some of the maintainers one way or the other. @syl20bnr ?

myrgy:
Hi,

I think it worth to add cquery directory  `cquery-cache-dir`

btw, I did modified you branch to support ccls - https://github.com/myrgy/fork-spacemacs/tree/ccls
please, let me know if you know how can we merge this 2 modes into one.

Thanks!

cormacc:
That's great @myrgy -- thank you kindly! I've merged the two in my local branch using a new lsp-c-c++-backend parameter, but probably won't push til next week, for two reasons:
1. CCLS is crashing when trying to index the project I'm currently working on (possibly due to the fact that it's embedded / PIC / xc16-gcc) and don't have any more time today to diagnose the problem properly.
2. I've done a fair bit to separate common and backend-specific configuration while merging, but could do more. A good bit of remaining backend specific config is still identical, just substituting 'ccls' for 'cquery' in a method call or variable name. Just need to figure out how to do this in elisp. It's pretty basic stuff, but just not that familiar with the syntax. An example below from `funcs.el`:

```

(defun lsp-c-c++//customise-lsp-ui-peek ()
  (if (eq lsp-c-c++-backend 'cquery)
    (cquery//customise-lsp-ui-peek) (ccls//customise-lsp-ui-peek))

  (defun lsp-c-c++/references-address ()
    (interactive)
    (lsp-ui-peek-find-custom
      'address "textDocument/references"
      (plist-put (lsp--text-document-position-params) :context
        '(:role 128))))

  (defun lsp-c-c++/references-read ()
    (interactive)
    (lsp-ui-peek-find-custom
      'read "textDocument/references"
      (plist-put (lsp--text-document-position-params) :context
        '(:role 8))))

  (defun lsp-c-c++/references-write ()
    (interactive)
    (lsp-ui-peek-find-custom
      'write "textDocument/references"
      (plist-put (lsp--text-document-position-params) :context
        '(:role 16)))))

(defun cquery//customise-lsp-ui-peek ()
  (defun lsp-c-c++/base () (interactive) (lsp-ui-peek-find-custom 'base "$cquery/base"))
  (defun lsp-c-c++/callers () (interactive) (lsp-ui-peek-find-custom 'callers "$cquery/callers"))
  (defun lsp-c-c++/derived () (interactive) (lsp-ui-peek-find-custom 'derived "$cquery/derived"))
  (defun lsp-c-c++/vars () (interactive) (lsp-ui-peek-find-custom 'vars "$cquery/vars"))
  (defun lsp-c-c++/random () (interactive) (lsp-ui-peek-find-custom 'random "$cquery/random")))

(defun ccls//customise-lsp-ui-peek ()
  (defun lsp-c-c++/base () (interactive) (lsp-ui-peek-find-custom 'base "$ccls/base"))
  (defun lsp-c-c++/callers () (interactive) (lsp-ui-peek-find-custom 'callers "$ccls/callers"))
  (defun lsp-c-c++/derived () (interactive) (lsp-ui-peek-find-custom 'derived "$ccls/derived"))
  (defun lsp-c-c++/vars () (interactive) (lsp-ui-peek-find-custom 'vars "$ccls/vars"))
  (defun lsp-c-c++/random () (interactive) (lsp-ui-peek-find-custom 'random "$ccls/random")))

```

cormacc:
@myrgy I've verified ccls works with a non-embedded project, and that the cquery functionality seems working as before with my current embedded project. Also updated the readme. Pushed if you want to check it out. Will raise my elisp game and eliminate some of the duplication mentioned in the previous post when I get a bit of time next week.

MaskRay:
@cormacc If you can give an example of non-confidential compile command line, I would like to know how ccls segfaults with the embedded / PIC / xc16-gcc setting!

`$ccls/base` no longer exists. `lsp-goto-implementation` is the replacement.

Personally, I use a ccls-specific extension (the `:flat` attribute in inheritanceHierarchy)

```
(defun ccls/bases ()
  (interactive)
  (lsp-ui-peek-find-custom 'base "$ccls/inheritanceHierarchy"
                           (append (lsp--text-document-position-params) '(:flat t :level 3))))
```

layers/+lang/lsp-c-c++/funcs.el
```diff
@@ -0,0 +1,131 @@
...
+
+(require 'cl-lib)
+(require 'subr-x)
+
+(defun lsp-c-c++//enable ()
+  (condition-case nil
+    (if (eq lsp-c-c++-backend 'cquery) (lsp-cquery-enable) (lsp-ccls-enable))
+    (user-error nil)))
+
+(defun cquery//enable ()
```
myrgy: This function might be unused, so it could be removed.

layers/+lang/lsp-c-c++/funcs.el
```diff
@@ -0,0 +1,131 @@
...
+
+(defun lsp-c-c++//bind-keys-for-mode (mode)
+  (spacemacs/set-leader-keys-for-major-mode mode
+    ;; format
+    "=c" 'spacemacs/clang-format-region-or-buffer
+    "=f" 'spacemacs/clang-format-function
+    ;; goto
+    "gf" 'find-file-at-point
+    "gF" 'ffap-other-window)
+  (ecase lsp-c-c++-backend
```
myrgy: I would propose to use ecase in other places. It looks much better that if cases.

myrgy:
Thank you @cormacc , it works for me!

cormacc:
@MaskRay Thanks for that suggestion re. bases. Re ccls indexing failure, I'm no longer seeing the error I was saying Friday (which I sadly didn't record -- vaguely remember something about a 'triple'. What I'm seeing now is a failure to find a header, following which indexing appears to stop? My .ccls-cache only contains entries for a subset of the files in the project. The directory containing the unfound header is included in the relevant entry in compile_commands.json (header is 'Neuromod.h', parent folder ending 'components/core/src', source file 'Usb.c'). Relevant excerpts from lsp-ccls stderr and compile_commands.json below. Though perhaps I should raise an issue against ccls or ccls-emacs rather than taking this PR too far OT. Let me know which you'd prefer and I'll move the discussion there.

lsp-ccls stderr
```
===-------------------------------------------------------------------------===
                         Miscellaneous Ungrouped Timers
===-------------------------------------------------------------------------===

   ---User Time---   --User+System--   ---Wall Time---  --- Name ---
   0.0000 (100.0%)   0.0000 (100.0%)   0.0000 (100.0%)  scan include paths
   0.0000 (100.0%)   0.0000 (100.0%)   0.0000 (100.0%)  Total

/home/cormacc/nmd/products/mbt/cd/firmware/src/Usb.h:24:10: fatal error: 'Neuromod.h' file not found
#include "Neuromod.h"
         ^~~~~~~~~~~~
In file included from /home/cormacc/nmd/products/mbt/cd/firmware/src/Usb.c:65:
/home/cormacc/nmd/products/mbt/cd/firmware/src/Usb.h:24:10: fatal error: 'Neuromod.h' file not found
#include "Neuromod.h"
         ^~~~~~~~~~~~

```

compile_commands.json entry
``` json
{
  "directory": "/home/cormacc/nmd/products/mbt/cd/build",
  "command": "/opt/microchip/xc16/v1.35/bin/xc16-gcc -D__33EP512GM710__ -D__dsPIC33E__ -I/home/cormacc/nmd/products/mbt/cd/bootloader/src -I/home/cormacc/nmd/products/mbt/cd/firmware/src -I/home/cormacc/nmd/products/mbt/cd/firmware/src/bist -I/home/cormacc/nmd/products/mbt/cd/neuromod/vendor/ceedling -I/home/cormacc/nmd/products/mbt/common/src -I/home/cormacc/nmd/products/mbt/common/test -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/core/src -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/core/test -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/core/support -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/bootloader/src -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/bootloader/test -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/bootloader/support -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/button/src -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/button/test -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/crypt/src -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/crypt/test -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/datetime/src -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/datetime/test -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/deprecated/src -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/deprecated/test -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/display/src -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/display/test -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/eventparser/src -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/eventparser/test -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/fileio/src -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/fileio/test -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/fileio/support -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/logging/src -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/packet/src -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/packet/test -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/packet/support -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/pal/src -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/pal/test -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/pal/support -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/pal/pic/src -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/pal/pic/test -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/pal/pic/support -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/scheduler/src -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/scheduler/test -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/scheduler/support -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/selftest/src -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/selftest/test -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/selftest/support -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/wave/src -I/home/cormacc/nmd/products/mbt/cd/neuromod/components/wave/test    -o CMakeFiles/cd.dir/firmware/src/Usb.c.obj   -c /home/cormacc/nmd/products/mbt/cd/firmware/src/Usb.c",
  "file": "/home/cormacc/nmd/products/mbt/cd/firmware/src/Usb.c"
},
```

cormacc:
@myrgy Incorporated your and @MaskRay's suggestions. Also tidied up some hasty crap left in Friday. Thank you both

cormacc:
Greatly reduced the duplication between configuration for cquery and ccls backends by introducing a couple of helper functions at the top of funcs.el

I've a couple of niggly bits annoying me -- see the excerpt from funcs.el below. The two commented out regions of =lsp-c-c++//common-config= are more elegant/concise, but result in the errors mentioned in the comments. Any ideas?

``` elisp
(defun lsp-c-c++//backend-string (prefix suffix) (concat prefix (symbol-name lsp-c-c++-backend) suffix))
(defun lsp-c-c++//backend-symbol (prefix suffix) (intern (lsp-c-c++//backend-string prefix suffix)))
(defun lsp-c-c++//call-backend-function (prefix suffix) (funcall (lsp-c-c++//backend-symbol prefix suffix)))
(defun lsp-c-c++//set-backend-symbol (prefix suffix value) (set (lsp-c-c++//backend-symbol prefix suffix) value))
(defun lsp-c-c++//set-backend-config (param prefix suffix) (when param (lsp-c-c++//set-backend-symbol prefix suffix param)))
(defun lsp-c-c++//set-backend-param (suffix) (lsp-c-c++//set-backend-config (intern (concat "lsp-c-c++" suffix)), nil, suffix))

(defun lsp-c-c++//enable ()
  (condition-case nil
  (lsp-c-c++//call-backend-function "lsp-" "-enable")
  (user-error nil)))

(defun lsp-c-c++//common-config ()
  (dolist (mode c-c++-modes)
    (spacemacs/lsp-bind-keys-for-mode mode)
    (lsp-c-c++//bind-keys-for-mode mode))
  (lsp-c-c++//customise-lsp-ui-peek)
  (setq-default flycheck-disabled-checkers '(c/c++-clang c/c++-gcc))

  (lsp-c-c++//set-backend-config lsp-c-c++-executable nil "-executable")
  (lsp-c-c++//set-backend-config lsp-c-c++-extra-init-params nil "-extra-init-params")
  (lsp-c-c++//set-backend-config lsp-c-c++-cache-dir nil "-cache-dir")
  (lsp-c-c++//set-backend-config lsp-c-c++-project-whitelist nil "-project-whitelist")
  (lsp-c-c++//set-backend-config lsp-c-c++-project-blacklist nil "-project-blacklist")
  (lsp-c-c++//set-backend-config lsp-c-c++-sem-highlight-method nil "-sem-highlight-method")
  ;; @todo -- This results in 'Error (use-package): cquery/:config: Symbol's function definition is void: \,
  ;; (lsp-c-c++//set-backend-param "-executable")
  ;; (lsp-c-c++//set-backend-param "-extra-init-params")
  ;; (lsp-c-c++//set-backend-param "-cache-dir")
  ;; (lsp-c-c++//set-backend-param "-project-whitelist")
  ;; (lsp-c-c++//set-backend-param "-project-blacklist")
  ;; (lsp-c-c++//set-backend-param "-sem-highlight-method")
  (when lsp-c-c++-sem-highlight-rainbow
    (lsp-c-c++//call-backend-function nil "-use-default-rainbow-sem-highlight"))

  ;;evil-record-macro keybinding clobbers q in cquery-tree-mode-map for some reason?
  ;; This doesn't work for some reason -- doing it per-backend in packages.el
  ;; Error (use-package): cquery/:config: Wrong type argument: keymapp, cquery-tree-mode-map
  ;; (evil-make-overriding-map (lsp-c-c++//backend-symbol nil "-tree-mode-map"))

  (evil-set-initial-state '(lsp-c-c++//backend-symbol nil "-tree-mode") 'emacs))

```

rcoacci:
Is there anything preventing this from being merged?  
Anything we can do to help?

sdwolfz:
This PR is a bit overwhelming due to it's size, both code and thread, but I feel like I can focus my effort on it and guide you to how to get it merged.

The main thing I want is to have the lsp functionality inside the current c-c++ layer, instead of having another layer.

Let me know if there are other changes or PRs I can focus on that will simplify your work on this one and I will get right to it (to the limits of my ability).

cormacc:
@sdwolfz That would be great. This PR contained just the lsp layer additions for most of its life -- I'd been maintaining the lsp-c-c++ layer in a branch of my own fork of the spacemacs repo, but added it recently as I was wondering whether the lack of a usage example was (one of) the reason(s) it wasn't getting any love.

The intent with the lsp-layer improvements was to provide a consistent set of of prefixes for core keybindings for all/most of the lsp layers for different languages. And a few of the other guys/gals working on lsp-based layers seem to think that's useful. So how does this sound as a plan of action...
1. Remove the lsp-c-c++ layer from this PR and work with you on getting the lsp layer changes merged
2. Integrate the lsp-c-c++ layer with the c-c++ layer and raise a new PR. This should (famous last words) be pretty straightforward as lsp-c-c++ is pretty orthogonal to c-c++, which it declares as a layer dependency anyway

This sound like a reasonable approach?

Re. 2 -- my original reason for creating a separate layer rather than taking this route originally were some responses to @MaskRay 's original suggestion of adding cquery to the c-c++ layer here: #10143 
@quicknir was working on a some tidy-up of the c-c++ layer at the time and appeared to have progressed quite a bit, but don't know what the current status of that work is.

sdwolfz:
@cormacc sounds like a good plan.

I recently merged the lsp support for Golang and Java so you can take a look there for hints on how to add your changes to the c-c++ layer.

Let me know when you are ready and I will start reviewing the code.

cormacc:
@sdwolfz OK - that latest commit amendment has stripped all the c-c++ stuff from this PR. So anyone who's actually using the lsp-c-c++ layer shouldn't pull this update :)

layers/+tools/lsp/funcs.el
```diff
@@ -46,3 +46,101 @@ https://github.com/emacs-lsp/lsp-javascript/issues/9#issuecomment-379515379"
...
+  (spacemacs/declare-prefix-for-mode mode "mj" "jump")
+  (spacemacs/declare-prefix-for-mode mode "ml" "lsp/backend")
+  (spacemacs/declare-prefix-for-mode mode "mr" "refactor")
+  (spacemacs/declare-prefix-for-mode mode "mT" "toggle")
+
+  (spacemacs/set-leader-keys-for-major-mode mode
+    ;;format
+    "=b" #'lsp-format-buffer
+    ;;execute
+    "ea" #'lsp-execute-code-action
```
sdwolfz: The `CONVENTIONS.org` file requires the `SPC e` prefix to contain code evaluation functionality. Is this what `lsp-execute-code-action` does? As far as I can tell it's a different thing.
Here's the definition of the `SPC m e` prefix:
```
Live evaluation of code is under the prefix ~SPC m e~.

| Key     | Description                                   |
|---------+-----------------------------------------------|
| ~m e $~ | put point at the end of the line and evaluate |
| ~m e b~ | evaluate buffer                               |
| ~m e e~ | evaluate last expression                      |
| ~m e f~ | evaluate function                             |
| ~m e l~ | evaluate line                                 |
| ~m e r~ | evaluate region                               |
```

layers/+tools/lsp/funcs.el
```diff
@@ -46,3 +46,101 @@ https://github.com/emacs-lsp/lsp-javascript/issues/9#issuecomment-379515379"
...
+    ;;goto
+    "gi" #'lsp-goto-implementation
+    "gt" #'lsp-goto-type-definition
+    ;;hierarchy
+    "hh" #'lsp-describe-thing-at-point
+    ;;jump
+    "ja" #'spacemacs/lsp-avy-document-symbol
```
sdwolfz: What is the difference between jump and goto? I feel like all the jump keybindings should go under goto.

layers/+tools/lsp/funcs.el
cormacc: Yep -- added that one at someone else's suggestion. Perhaps @yyoncho? Believe it does things like syntax corrections etc. if suggested by language server. Support seems to be pretty recent in cquery, though was able to execute a code action of some description during my brief time playing with ccls.

I've moved it under `m l e` for now, though open to other suggestions...

layers/+tools/lsp/funcs.el
cormacc: Yeah you're right. At some point I felt goto was in danger of getting overcrowded and made an artificial distinction between actions with only one possible target (goto) and those with multiple potential targets (jump - lsp-ui-peek). Moved all the `j` bindings back under `g` (had to rebind lsp-imenu under `m g m` as `m g i` was already bound to goto implementation).

cormacc:
Made requested changes and rebased on develop branch tip.

sdwolfz:
Thank you :heart:!

sdwolfz:
@cormacc It's better if you open up a new PR.

By the way it looks like you associated the commit with the `cormaccannon` user, is that intentional? There's no problem about it but you won't get credit for it in GitHub's statistics.

cormacc:
Grand -- raised #11203
No idea how I associated the commit with that other account. Had forgotten it existed (was empty), but I've deleted it now :)
Thanks for the help -- it'll be next week before I get back to the c-c++ layer -- house guests arriving tomorrow.

layers/+tools/lsp/funcs.el
yyoncho: Execute code action is applying quick fix, e. g. if you are on a string a potential code action is "Extract constant". Possible bindins are under "mrea", which will stand for Refactoring -> Execute Action.

layers/+tools/lsp/funcs.el
```diff
@@ -46,3 +46,98 @@ https://github.com/emacs-lsp/lsp-javascript/issues/9#issuecomment-379515379"
       (all-completions (company-grab-symbol) candidates)))
   (make-local-variable 'company-transformers)
   (add-to-list 'company-transformers 'lsp-prefix-company-transformer))
+
+(defun spacemacs/lsp-bind-keys-for-mode (mode)
+  "Define key bindings for the specific MODE."
+  (spacemacs/declare-prefix-for-mode mode "m=" "format")
+  (spacemacs/declare-prefix-for-mode mode "mg" "goto")
+  (spacemacs/declare-prefix-for-mode mode "mh" "hierarchy")
```
yyoncho: IMO this shouldn't be here since only small part of the LSP server provide hierarchy support. Also, as per  CONVENTIONS.org "mh" stands for help. I am voting to remove this line out of base layer and maybe use "mH" for hierarchy in C++ mode. @sdwolfz what do you think?

cormacc: I've no objections, but maybe move this discussion over to #11203 (which I'll re-title) as this PR is now closed

PR #10491, opened 2018-03-17
org: add `SPC m i n` binding for org-add-note

tko:
Add `SPC m i n` binding for `org-add-note`

Note: the note capturing buffer states `Finish with C-c C-c, or cancel with C-c C-k.` and while I think those should probably be bound to `, c` and `, k` or similar and the note updated accordingly, I don't quite now how to do that exactly.

Further note, `, ,` appears to do the same function as `C-c C-c` already.


eklitzke:
This makes sense to me.

sdwolfz:
Thank you :heart:!
Let's deal with the note buffer keybindings in another PR.

PR #10497, opened 2018-03-19
add eslint-fix to javascript layer, run on save in js2-mode

coldpour:
this change adds config to the javascript layer to run `eslint --fix` on your code before saving the file.

### DIFF
--- a/layers/+lang/javascript/packages.el
+++ b/layers/+lang/javascript/packages.el
@@ -15,6 +15,7 @@
         coffee-mode
         company
         (company-tern :requires company)
+        eslint-fix
         evil-matchit
         flycheck
         ggtags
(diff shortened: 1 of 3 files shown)


layers/+lang/javascript/packages.el
```diff
@@ -15,6 +15,7 @@
         coffee-mode
         company
         (company-tern :requires company)
+        eslint-fix
```
aaronjensen: this should be conditionally installed only if `javascript-eslint-fix-on-save`, I believe it's:

```elisp
(eslint-fix :toggle javascript-eslint-fix-on-save)
```

Then you can remove the `:if` in the use-package below.

coldpour: ah smart. thanks

coldpour:
@aaronjensen, thanks for the pointer to use [eslintd-fix](https://github.com/aaronjensen/eslintd-fix). I'll definitely give that a go.

This pr also has a [sibling that adds eslint-fix as it's own layer](https://github.com/syl20bnr/spacemacs/pull/10508). I'm not sure which is a better approach... to have the javascript layer set up stuff like this and flow-type with config vars or to have them be separate layers per tool. Do you have a preference?

aaronjensen:
Probably the way to do it would be to model it off of the `syntax-checking` and `auto-completion` layers. An `auto-fix` layer (or add it to `syntax-checking`) that configured as many languages as it could would probably be the most spacemacs way.

PR #10510, opened 2018-03-22
Add function and key binding for single column window layer.

mirubiri:
We have key bindings to create two and three column
window layers.

This commit makes available a key binding to return to
the default spacemacs window layer, the single column one.


sdwolfz:
Thank you :heart:! 

This is a really handy function. Great job! 
Any chance you would be interested in contributing a `w4` function that splits the screen into a 2x2 grid?

mirubiri:
@sdwolfz sure, I'll try, i'm slowly learning elisp :)

PR #10538, opened 2018-03-29
go: add support to generate go tests with go-gen-test

ccojocar:
Add support to generate go tests with [go-gen-test](https://github.com/s-kostyaev/go-gen-test).

### DIFF
--- a/layers/+lang/go/README.org
+++ b/layers/+lang/go/README.org
@@ -29,6 +29,7 @@ This layer adds extensive support for go to Spacemacs.
 - Edit struct field tag with [[https://github.com/fatih/gomodifytags][gomodifytags]]
 - Linting with flycheck's built-in checkers or flycheck-gometalinter
 - Coverage profile visualization
+- Test generation with [[https://github.com/s-kostyaev/go-gen-test][go-gen-test]]
 
 * Install
 ** Pre-requisites
@@ -41,6 +42,7 @@ You will need =gocode=, =gogetdoc=, =godef= and =godoctor=:
   go get -u -v golang.org/x/tools/cmd/gorename
   go get -u -v golang.org/x/tools/cmd/goimports
   go get -u -v github.com/zmb3/gogetdoc
+  go get -u -v github.com/cweill/gotests/...
 #+END_SRC
 
 If you wish to use =gometalinter= set the value of =go-use-gometalinter= to t:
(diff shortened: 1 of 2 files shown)


cosmincojocar:
@syl20bnr Please could you have a loot at this PR?

sdwolfz:
@cweill can I bother you to ask for reviews to this PR and the other ones opened by @ccojocar here (there should be 4). 

These look OK to me but I am not really familiar with Golang packages and would appreciate some more feedback from the community about them.

cweill:
@sdwolfz: Thanks for reaching out. Unfortunately, I don't have any experience writing emacs plugins for go tools. In general, installing go commandline tools is as simple as calling `go get <package>`. Then it's just a matter of having your editor configured to pass the right arguments to the commandline tools, which should already be available if $PATH is configured correctly (which it should be upon installing `go`). 

I think it would be better to reach out to the maintainer of a Go emacs package. 

Otherwise, these pull requests look good to me.

sdwolfz:
Glad to hear from you @cweill. I was interested in your feedback regarding the Golang tools themselves (since you're the author of one of them :smile_cat: ).

While your repo seems quite popular and recently updated, the rest look inactive in comparison. Do you have any concerns about them not surviving in the long run? I would not want to add some functionality only to remove it a year later because the tools themselves become obsolete... (maybe I'm overreacting and there's nothing to worry about :confused:).

ccojocar:
Just want to mention that [vscode-go](https://github.com/Microsoft/vscode-go) or vim-go are using as well these Go tools.

sdwolfz:
That gives me some peace of mind. I will go ahead and test them out and merge them soon. Thanks @ccojocar.

cweill:
@sdwolfz: Sorry for the late reply; lately I work only in Python and C++, and am no longer very active in the Go community. [gotests](https://github.com/cweill/gotests) is essentially in maintenance mode, though it still actively accepts PRs.

I would let someone else comment on the state of the other packages. As @ccojocar said, other IDEs use these packages in their Go plugin, so it may be good to include.

syl20bnr:
Thank you ! 👍 

I updated the conventions for test generation in commit https://github.com/syl20bnr/spacemacs/commit/43f2feb2b and update the key bindings of this PR accordingly in https://github.com/syl20bnr/spacemacs/commit/a0277695d

PR #10544, opened 2018-03-30
Revert "git-gutter: remove redundant arg for timer"

zhenlei:
It's not redundant arg. This change will disable pomodoro counter down timer display.

This reverts commit 433ae7c30bc0884519018590942e516a4edbc14b.


sdwolfz:
Can you please check if this problem still reproduces for you with the latest develop? I am able to see the pomodoro timer even with git gutter.

If you do happen to still have this problem, please add some steps I can take to reproduce. And it would be best if you made a separate branch for this PR since you seem to have more changes on this branch that are not related to the issue.

duianto:
@zhenlei This PR needs some cleanup, it even includes your `.spacemacs` file.

But before fixing it, test if the issue still occurs on the latest `develop` branch with updated packages.

PR #10572, opened 2018-04-08
LaTeX-mode -> latex-mode

ljli:
`LaTeX-mode` doesn't work for me, but `latex-mode` does.

### DIFF
--- a/layers/+lang/latex/packages.el
+++ b/layers/+lang/latex/packages.el
@@ -155,7 +155,7 @@
   (add-hook 'LaTeX-mode-hook 'evil-matchit-mode))
 
 (defun latex/post-init-flycheck ()
-  (spacemacs/enable-flycheck 'LaTeX-mode))
+  (spacemacs/enable-flycheck 'latex-mode))
 
 (defun latex/post-init-flyspell ()
   (spell-checking/add-flyspell-hook 'LaTeX-mode-hook))


sdwolfz:
@ljli can you please give a small example where this fails so I can try it out locally before merging? I am not familiar with latex mode at all :crying_cat_face: ...

ljli:
Sure, so the problem is just that the major-mode for latex is named `latex-mode` not `LaTeX-mode`. You can open any `.tex` file (can be empty) and inspect the variable `major-mode`. You can also see that the variable `flycheck-enabled-checkers` is nil, with the current code. If we give the correct mode name flycheck will be enabled and the variable will contain something like `'(tex-chktex)`.

ljli:
Or maybe that is just me, I'm using the standard latex layer, though. Confusingly there seems to be a `LaTeX-mode` and `latex-mode` the former belonging to the AUCTeX package and the later being part of emacs. The layer seems to use AUCTeX, though, and both mode names are referenced in the layer code e.g. packages.el. Color me confused. All I can say is that flycheck works for me with this commit and does not without. So indeed it probably good if someones investigates further. This tex code should give an error like `Solo document found`.
```
\documentclass{article}
begin{document}
\end{document}
```

smile13241324:
Tested this with the latest version on develop and the change seems to be not longer necessary. Whether I am selecting `latex-mode` or `LaTeX-mode` the result is always `LaTeX-mode` in my `.tex` file.

PR #10583, opened 2018-04-12
handle .venv created by pipenv correctly

twillis:
when pipenv is run as...

PIPENV_VENV_IN_PROJECT=1 pipenv ...

a virtualenv directory called `.venv` is installed in the directory as the
`Pipfile`

the current functionality is to look for a file and then tell pyenv to 'workon'
it's contents. doing the above will cause it to blow up.

this change checks to see if it's a directory or file and either calls

pyvenv-activate if directory

or

pyvenv-workon with the contents of the file if file(existing behaviour)

then end result, a virtualenv created local for the project by pipenv will get
activated automatically instead of blowing up like it does now

Thank you <3

### DIFF
--- a/layers/+lang/python/funcs.el
+++ b/layers/+lang/python/funcs.el
@@ -207,11 +207,15 @@ as the pyenv version then also return nil. This works around https://github.com/
     (when root-path
       (let* ((file-path (expand-file-name ".venv" root-path))
              (virtualenv
-              (with-temp-buffer
-                (insert-file-contents-literally file-path)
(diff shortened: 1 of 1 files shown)


twillis:
yeah I guess someone cooler than me added equivalent functionality to the python layer and knew how to get their contribution merged.

duianto:
I doubt it was someone cooler 😄 
The newer PR was just reviewed before this one.
I don't know why this fell through the cracks 😢

PR #10585, opened 2018-04-12
Fix helm layer broken with smex package

CeleritasCelery:
Currently if you add the package `smex` (not the layer) to Spacemacs it will cause `SPC SPC` to be unbound. We should be checking for the smex layer (not package) to be present because the layer defines the proper keybindings and the package does not.


sdwolfz:
Thank you :heart:!

PR #10599, opened 2018-04-17
add support for helm find files edit to helm-navigation transient state

CeleritasCelery:
We created a function to allow one to edit a directory from `helm-find-files` using `dired`. We should add that to the edit action (`e`) in the `helm-navigation` transient state.


sdwolfz:
Thank you :heart:! Took me a while to understand what this did but it was worth it.

PR #10611, opened 2018-04-22
Add variable rcirc-enable-rcirc-late-fix

Miciah:
#### rcirc-late-fix: Fix compiler warnings

Fix the following errors:

    rcirc-late-fix.el:49:12:Warning: ‘rcirc-print-hooks’ is an obsolete
        variable (as of 24.3); use ‘rcirc-print-functions’ instead.

    rcirc-late-fix.el:51:1:Warning: defface for ‘rcirc-late-fix-face’ fails
        to specify containing group

    In rcirc-late-fix-hook:
    rcirc-late-fix.el:79:82:Warning: (lambda (x) ...) quoted with ' rather
        than with #'

    In rcirc-late-fix-matching-buffer:
    rcirc-late-fix.el:85:34:Warning: function ‘find-if’ from cl package
        called at runtime

#### Add variable `rcirc-enable-late-fix` to rcirc layer

Add a configuration variable to enable the included rcirc-late-fix package in the rcirc layer.  The package is still disabled by default.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -115,6 +115,9 @@ This file containes the change log for the next major version of Spacemacs.
 - Make =org-projectile= integration compatible with the newest version.
 - Add package =helm-org-rifle=, with keybinding ~SPC a o r~ to rifle through
   files.
+**** Rcirc
+- New variable =rcirc-enable-late-fix= to enable or disable the included
+  =rcirc-late-fix= package (disabled by default).
 **** Scala
 - Move =ensime= to the =java= layer (Tor Hedin Bronner)
 **** Syntax-checking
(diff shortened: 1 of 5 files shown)


Miciah:
I updated the PR to add a changelog entry.  Anything else I can do to get this PR merged?

Miciah:
I updated the PR so that instead of `(require 'rcirc-late-fix)` in the config hook in `rcirc/init-rcirc`, it uses `(use-package rcirc-late-fix)` in a post-config hook added in `rcirc/init-rcirc-late-fix`.  I would like to make a similar change to the way rcirc-color is loaded in a subsequent PR if there are no objections.

Miciah:
Rebased.

Miciah:
Pushed a couple small simplifications for `rcirc-late-fix.el`: replace `mapc` + lambda with just a `dolist`, and replace `cl-find-if` + lambda with just a `cl-find`.

sdwolfz:
Thank you :heart:!

PR #10614, opened 2018-04-22
Fix ipython-notebook layer instructions

x-ji:
Now `pip install jupyter` is needed instead of `pip install ipython[notebook]`.
Otherwise `ob-ipython` package will not work. I encountered errors when trying to open org files after installing the layer.


sdwolfz:
Thank you :heart:!

PR #10624, opened 2018-04-25
spacemacs-modeline: compute height only after display init

bennofs:
Before the display system is initialized, we cannot reliably make font
measurements so the height will be incorrect. This lead to display artefacts
if emacs was started in daemon mode without a graphical interface and later a
graphical client was connected (so for example, if you do `emacs --daemon`
followed by `emacsclient -c`).


syl20bnr:
Thank you ! 👍 

Issue #8206, opened 2017-01-18
Replace react-mode with rjsx-mode?


aaronjensen:
[rjsx-mode](https://github.com/felipeochoa/rjsx-mode) has been working quite well for me instead of `web-mode`/`react-mode`. It derives from `js2-jsx-mode`, but has working indentation and syntax highlighting for jsx. 

What does everyone think of dropping the web-mode derived react-mode and replacing it w/ rjsx-mode? Personally, I just use rjsx-mode for *all* javascript.

d12frosted:
I have never tried `rjsx-mode` (and I don't do much react development), but `react-mode` and `web-mode` are both buggy whenever I use them for editing react files. Next time I'll try `rjsx-mode`.

Would be interesting to hear from react users on this matter.

myme:
I was thinking about filing this request myself. I've been using `rjsx-mode` a bit and it feels much closer to editing in `js2-mode`, obviously. Since it's so close to `js2-mode`, perhaps parameterizing the `javascript` layer with `rjsx`/`react` which pulls in that extra package could be an idea?

elvis-sik:
Flycheck with eslint isn't working out of the box with `rjsx-mode` for me.

elvis-sik:
Was able to get it working easily with `SPC e S`. I am currently having two problems with `rjsx-mode`. `Tab` in insert mode is never indenting and always bringing the autocompletion tooltip, even with no previous input.

Also, there is some new ES syntax not yet supported in `js2-mode`, and therefore also not in `rjsx-mode`. Though I don't see why this should stop Spacemacs from adopting it. Anyway, the specific syntax rule I am thinking about has already been reported to `js2-mode` and it seems they are working on it.

aaronjensen:
> Flycheck with eslint isn't working out of the box with rjsx-mode for me.

`(spacemacs/add-flycheck-hook 'rjsx-mode)`

> Tab in insert mode is never indenting and always bringing the autocompletion tooltip, even with no previous input.

Can't help you here. I don't experience this. I use company-mode with some custom stuff and my tab is bound to `indent-for-tab-command`.

> Anyway, the specific syntax rule I am thinking about has already been reported to js2-mode and it seems they are working on it.

What is the syntax out of curiosity?

elvis-sik:
This kind of method assignment:

```javascript
class Foo {
  bar = () => {
    return this;
  }
}
```

[...]

PR #10627, opened 2018-04-26
Use rjsx-mode for react framework

ztlevi:
Since `rjsx-mode` is generally considered better mode for react editing, I put the PR to replace the react-mode with `rjsx-mode`. `rjsx-mode` will also be enabled for files with =react= imported by using `magic-mode-alist`.


ztlevi:
Since we use this method to enable `yasnippet` in `react-mode` setup [here](https://github.com/ztlevi/spacemacs/blob/develop/layers/+frameworks/react/funcs.el#L21). This command might need to be added when init `yasnippet`.

ztlevi:
As for the failed test, I'm not sure if we need to change the second to the third. It seems like `spacemacs-defaults` is the second one here while the [test](https://github.com/syl20bnr/spacemacs/blob/develop/tests/core/core-configuration-layer-ftest.el#L36) assigns the `spacemacs-base` as the distribution. But I wasn't really change any base distributions in the commits.
```
configuration-layer--used-layers is a variable defined in ‘core-configuration-layer.el’.
Its value is
(spacemacs-bootstrap spacemacs-defaults spacemacs-base ...)

Documentation:
A non-sorted list of used layer names.
```

fiveNinePlusR:
I am curious why you removed company-tern... does completion still work with your changes?

ztlevi:
My bad. I totally moved to lsp now and I forget to put company-tern back. I am going to fix this and open a new PR for lsp.😂

ztlevi:
Just added `company-tern` with rebase.

ztlevi:
lsp support can be tested through this [PR](https://github.com/syl20bnr/spacemacs/pull/10628)

ztlevi:
Any thoughts on this PR. From my perspective, I prefer to put `rjsx-mode` in the Javascript layer because it's just one single packages and the `JSX` file editing is really common across Javascript development.

fiveNinePlusR:
I think it is probably worth using. are you aware of anything that `rjsx-mode` can't do compared to `react-mode`?

ztlevi:
Not really. There are a couple reason I pick `rjsx-mode`.
1. `web-mode` is not designed for JSX and I might need more tweaks compared with `rjsx-mode`
2. `rjsx-mode` support `flow-type`.
3. It's just kind of tricky to enable `lsp` for `web-mode` because that doesn't really work for `html`'s scripts.

ztlevi:
Another choice is to put `rjsx` just under `javascript` layer since it's a single package added for JSX file editing.

syl20bnr:
Yes, it could be put under the javascript layer :+1:

fiveNinePlusR:
I tested this pr out and the layer does not have lsp support when in rjsx mode. tested by calling `lsp-capabilities` and I get a message in the bottom saying this buffer does not support lsp mode. switching to js2-mode and doing the above works just fine. any insight as to why that wouldn't work?

syl20bnr:
Actually let's keep it in `+frameworks/react` for now.

We'll see later if another pattern emerges, for now `react` seems the
natural way to go.

Sylvain Benner <notifications@github.com> writes

> Yes, it could be put under the javascript layer :+1:
>

fiveNinePlusR:
@ztlevi indeed i did and it definitely works under the js2-mode

ztlevi:
@fiveNinePlusR It should be good to test now. Anyone who familiar with flow-type javascript is welcome to test since I'm not really using it.

`lsp-capibilities` result for `rjsx-mode`:
```
The server provides command execution support
The server provides code actions
The server provides completion support
The server provides project symbol support
The server provides file symbol support
The server provides references support
The server provides goto definition support
The server provides signature help support
The server provides hover support
Document sync method: None
```

image

axyz:
didn't try this PR yet, how does it behave with indentation?
as mentioned on https://github.com/syl20bnr/spacemacs/issues/8206 web-mode has in general more consistent indentation, with good prettier/eslint support won't be an issue nowadays, but not everybody will use them.

In general I'm currently trying out mixing webmode and rjsx-mode (waiting for https://github.com/felipeochoa/rjsx-mode/pull/66), maybe is something that can be tried out here as well. It still feel a bit buggy:
- weird copy paste
- not relevant completions polluting the list
- indentation issues
- no autoclose of jsx components

Let me know what's the state in these regards and if could benefit to use rjsx as a minor mode, in general I would also be more than happy to move to a more js specific layer as web-mode is working great on the parsing/indenting side but still feel a bit of a hack and lot of IDE features are not available without rjsx minor mode

axyz:
cool then, I think we can give it a try and eventually iterate over it. Ideally in the future everything could be included under js layer in order to also have a single place that could enable prettier/eslint/flow/etc... using autodiscovery

fiveNinePlusR:
>   1.     < inserts </> whenever it would start a new JSX node (and simply inserts < otherwise).
>   2.   > or C-d right before the slash in a self-closing tag automatically inserts a closing tag and places point inside the element

also, typing the tagname and pressing tab will expand it to the full double enclosing tag.

syl20bnr:
Thank you ! 👍 

I extracted both `tern` and `web-beautify` from javascript layer into their own layer. I also simplified the react layer where I could, see the commits following the merge of this PR.

Cherry-picked and squashed into develop branch, you can safely delete your branch.

PR #10628, opened 2018-04-26
Add lsp-javascript backend for javascript layer

ztlevi:
Added `lsp-javascript` for javascript layer. Tested `company`, `spacemacs-jump-handlers-js2-mode` for both tern and lsp.

PR #10632, opened 2018-04-28
Add LSP to jump handlers

ztlevi:
`spacemacs//set-lsp-key-bindings` function is defined in this [commit](https://github.com/syl20bnr/spacemacs/pull/10628/commits/1c94c234931940122635de18c3199affc4abb5b2).

I added the support for python layer. But it conflicts with `xref` somehow when we have multiple candidates.

Here is an example.
kapture 2018-04-27 at 23 54 42


syl20bnr:
And thank you one more ! 🎉 

PR #10643, opened 2018-05-03
Add kaolin themes to spacemacs-theme-name-to-package

wjlroe:
This helps with using kaolin themes, so that spacemacs doesn't try to install
the specific theme name from ELPA.

Kaolin themes are here: https://github.com/ogdenwebb/emacs-kaolin-themes


wjlroe:
Looks like these tests are failing on the develop branch already, so I'm not sure if I need to do anything to make them pass in this PR

PR #10649, opened 2018-05-05
Add keybinding to docker layer to build image without cache

smile13241324:
Made a small change at the docker layer to also provide keybindings for building the image without using the cache as it is sometimes necessary.

### DIFF
--- a/layers/+tools/docker/packages.el
+++ b/layers/+tools/docker/packages.el
@@ -54,4 +54,5 @@
       (spacemacs/declare-prefix-for-mode 'dockerfile-mode
         "mc" "compile")
       (spacemacs/set-leader-keys-for-major-mode 'dockerfile-mode
-        "cb" 'dockerfile-build-buffer))))
+        "cb" 'dockerfile-build-buffer
+        "cB" 'dockerfile-build-no-cache-buffer))))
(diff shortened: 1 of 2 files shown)


layers/+tools/docker/packages.el
```diff
@@ -54,4 +54,6 @@
       (spacemacs/declare-prefix-for-mode 'dockerfile-mode
         "mc" "compile")
       (spacemacs/set-leader-keys-for-major-mode 'dockerfile-mode
-        "cb" 'dockerfile-build-buffer))))
+        "cb" 'dockerfile-build-buffer)
+      (spacemacs/set-leader-keys-for-major-mode 'dockerfile-mode
```
sdwolfz: This can be merged with the one above:

```emacs-lisp
(spacemacs/set-leader-keys-for-major-mode 'dockerfile-mode
         "cb" 'dockerfile-build-buffer
         "cB" 'dockerfile-build-no-cache-buffer)
```

smile13241324: @sdwolfz yep, you are right, I have adjusted my code.

sdwolfz:
I tried it locally and it works as expected. 
A bit dissapointed that it sets the build context to the file's directory and you can't change that, but I guess that'll be a feature request to https://github.com/spotify/dockerfile-mode.

Thank you :heart:!

PR #10655, opened 2018-05-07
Add function 'spacemacs/last-error

wizmer:
Sometimes it is useful to go directly to the last error message in the buffer.
This function recursively calls 'spacemacs/next-error until the last error is
reached.
The binding "SPC e l" is associated with it

### DIFF
--- a/core/core-funcs.el
+++ b/core/core-funcs.el
@@ -350,6 +350,12 @@ is not visible. Otherwise delegates to regular Emacs next-error."
      ((eq 'flycheck sys) (call-interactively 'flycheck-next-error))
      ((eq 'emacs sys) (call-interactively 'next-error)))))
 
+(defun spacemacs/last-error ()
+  "Go to last flycheck or standard emacs error."
+  (interactive)
+  (spacemacs/next-error)
+  (spacemacs/last-error))
+
 (defun spacemacs/previous-error (&optional n reset)
   "Dispatch to flycheck or standard emacs error."
   (interactive "P")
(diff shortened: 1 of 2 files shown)


syl20bnr:
Thank you for the PR. `SPC e l` is already taken by `error list` from Flycheck as well as `SPC e L`. I suggest `SPC e $` as `$` in Vim is used to go to last.

wizmer:
To be honest, I don't think last-error would be very useful in the flycheck context. So maybe in the spirit of the dispatching between `flycheck-next-error` and `next-error`, `SPC e l` could dispatch between flycheck `error list` and `last-error` ?

syl20bnr:
It would be surprising to have different behaviors with the same key
binding :-(

sdwolfz:
Maybe it's to much of a hacky solution, but how about:

```emacs-lisp
(defun spacemacs/last-error ()
  "Go to last flycheck or standard emacs error."
  (interactive)
  (evil-goto-line)
  (spacemacs/previous-error))

(spacemacs/set-leader-keys
    "eG" 'spacemacs/last-error)
```

This navigates to the last line of the file and then to the previous error.

This could also work as a keybinding:
```emacs-lisp
(spacemacs/set-leader-keys
    "ez" 'spacemacs/last-error)
```

If all else fails you could do the following key combination to get to the last error without any code changes `SPC eLGkq` :smile_cat: .

smile13241324:
Thank you for contributing to spacemacs :purple_heart:. I have cherry picked your commit into develop.

I have used a subtly changed version of @sdwolfz suggestion to improve the function's performance. The original function would have run once for every error in the buffer.
I have also adjusted the keybinding to `ez` not to create conflicts with existing bindings.

Issue #10662, opened 2018-05-10
[typescript] tsserver no longer starts automatically when opening first file for a project


jupl:
#### Description :octocat:
tsserver for tide/lsp does not start automatically.

#### Reproduction guide :beetle:
- Start Emacs with the TypeScript layer
- Open a `.ts` file in a TS project.

*Observed behaviour:* :eyes: :broken_heart:
Tide/LSP is not running. I need to do `M-x tide-setup` manually (for Tide), `revert-buffer`, or open another file in the project.

*Expected behaviour:* :heart: :smile:
Tide/LSP should automatically start as in previous versions. (at least for Tide)

#### System Info :computer:
- OS: darwin
- Emacs: 25.3.3
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. aa5bc0373)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(helm auto-completion emacs-lisp git lsp neotree syntax-checking typescript version-control)
```
- System configuration features: NOTIFY ACL LIBXML2 ZLIB TOOLKIT_SCROLL_BARS

jupl:
Commit 59e0c56 (Add: lsp-typescript support) seemed to have broken it, as the commit just before that works as expected.

jupl:
@ztlevi Can you verify this? Appears you were the author for this.

ztlevi:
Could you try to `revert-buffer` after couple seconds just waiting for `tide` to set up properly?

I have the similar issue with the backend startup and it's due to the process is not async enough to enable the backend in Emacs after the backend is started. And a lot of time, I have to `revert-buffer` to make sure everything is running correctly.

I use `revert-buffer-no-confirm` cuz it's faster.
```
;; refresh buffer
(defun revert-buffer-no-confirm ()
  "Revert buffer without confirmation."
  (interactive)
  (save-buffer)
  (revert-buffer :ignore-auto :noconfirm))
```

ztlevi:
It works fine on my machine... Including `company` and `jump-handler`.
![image](https://user-images.githubusercontent.com/16655096/39897213-333fb4f2-5466-11e8-81fb-fd556ebbd7b7.png)

jupl:
[...]

PR #10664, opened 2018-05-11
Fix: remove buggy hooks in javascript and typescript

ztlevi:
There are several recursive hooks in Javascript and Typescript layer [causing backends does not start properly](https://github.com/syl20bnr/spacemacs/issues/10662#issuecomment-388172951). I removed those buggy hooks. And it should be good now.

Examples like:
```
(add-hook 'js2-mode-hook 'spacemacs//javascript-setup-backend))
```
While in `spacemacs//javascript-setup-backend`, it has `spacemacs//javascript-setup-tern` which hooks `tern-mode` to the `js2-mode`.
```
(defun spacemacs//javascript-setup-tern ()
  "Setup tern backend."
  (add-hook 'js2-mode-hook 'tern-mode)
  (progn
    (spacemacs|hide-lighter tern-mode)
    (when javascript-disable-tern-port-files
      (add-to-list 'tern-command "--no-port-file" 'append))
    (spacemacs//set-tern-key-bindings 'js2-mode)))
```

### DIFF
--- a/layers/+lang/typescript/packages.el
+++ b/layers/+lang/typescript/packages.el
@@ -25,8 +25,7 @@
   (add-hook 'typescript-tsx-mode-hook #'add-node-modules-path))
 
 (defun typescript/post-init-company ()
-  (add-hook 'typescript-mode-hook #'spacemacs//typescript-setup-company)
-  (add-hook 'typescript-tsx-mode-hook #'spacemacs//typescript-setup-company))
+  (spacemacs//typescript-setup-company))
 
 (defun typescript/pre-init-eldoc ()
   (spacemacs|use-package-add-hook tide :post-init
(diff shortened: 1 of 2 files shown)


layers/+lang/typescript/packages.el
```diff
@@ -84,8 +83,7 @@
     :defer t
     :init
     ;; setup javascript backend
```
Miloas: ;; setup typescript backend

jupl:
One of the side effects here is for the JavaScript layer is before if I don't have `tern` available it's a silent warning. After testing this every time Spacemacs starts up the home screen is interrupted with a full-screen *Warnings* buffer:

```
Error (use-package): js2-mode/:init: Symbol’s value as variable is void: tern-command
```

ztlevi:
@jupl This error can be turned off by using `:toggle` keyword and `find-executable` method but the warning might be necessary to remind users to install `tern`.

ztlevi:
@jupl Not sure if this is what you want. But that also needs the machine has package managers installed.
https://github.com/jwiegley/use-package#use-package-ensure-system-package

syl20bnr:
Thank you ! 👍 

PR #10668, opened 2018-05-11
Fix: use recipe for lsp-javascript-typescript

ztlevi:
Melpa pulled off some libraries from `lsp-javascript` ([link](https://github.com/melpa/melpa/commit/97b6ddeb34297aacfefdd4c227ad520b70a12bc6#commitcomment-28940048) here). However, the author doesn't really want to split these libraries ([link](https://github.com/emacs-lsp/lsp-javascript/issues/12#issuecomment-388491329) here).

I guess we need to use recipe to install `lsp-javascript` for now.

### DIFF
--- a/layers/+lang/javascript/packages.el
+++ b/layers/+lang/javascript/packages.el
@@ -30,7 +30,7 @@
         web-beautify
         skewer-mode
         livid-mode
-        (lsp-javascript-typescript :requires lsp-mode)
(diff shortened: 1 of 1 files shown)


syl20bnr:
Thank you ! 👍 

CestDiego:
This is not working anymore :(

syl20bnr:
Do we still have the issue ?

ztlevi:
I think they don’t have the issue anymore cuz I drop the recipe and use melpa to install. Not sure why they cannot use recipe. I really cannot reproduce the issue.

PR #10669, opened 2018-05-12
Add a 2x2 grid layout function shortcut

mirubiri:
@sdwolfz Here you got the window shortcut for the 2x2 grid ;)

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -559,6 +559,17 @@ in a split window to the right."
   (split-window-horizontally)
   (other-window 1))
 
+(defun spacemacs/layout-grid ()
+  " Set the layout to a 2x2 grid. "
+  (interactive)
+  (delete-other-windows)
+  (split-window-right)
+  (split-window-below)
+  (windmove-right)
+  (split-window-below)
+  (windmove-left)
+  (balance-windows))
+
 (defun spacemacs/layout-triple-columns ()
   " Set the layout to triple columns. "
   (interactive)
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -2069,7 +2069,8 @@ Windows manipulation commands (start with ~w~):
 | ~SPC w \vert~          | maximize window vertically                                                  |
 | ~SPC w 1~              | make 1 window layout                                                        |
 | ~SPC w 2~              | make 2 windows layout (split vertically)                                    |
-| ~SPC w 3~              | make 3 windows layout  (split vertically)                                   |
+| ~SPC w 3~              | make 3 windows layout (split vertically)                                    |
+| ~SPC w 4~              | make 4 windows layout (split in 2x2 grid)                                   |
 | ~SPC w b~              | force the focus back to the minibuffer                                      |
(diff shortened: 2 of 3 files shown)


sdwolfz:
Awesome! I'll take a look and merge it this evening.

layers/+spacemacs/spacemacs-defaults/funcs.el
```diff
@@ -559,6 +559,17 @@ in a split window to the right."
   (split-window-horizontally)
   (other-window 1))
 
+(defun spacemacs/layout-grid ()
```
mirubiri: I'm not sure if the name of the function should include a `2x2` or this is ok.

sdwolfz: It's OK as is, I don't realy think other grid layouts might be practical due to the small size of windows.

Sometimes I code with a 2x2 layout with an extra neotree open on the side and it becomes hard to read if I have to do many changes in just one file. But it's great for small one/two liners across multiple files.

mirubiri: Yup ok, that's my way of thinking too :)

bmag:
Hey, I saw this nice addition and I've got a small comment. It's meant only to share some knowledge, not as a request for changes (since @sdwolfz already said he'd look and merge it):

FWIW, nearly all window-splitting functions return the new window as their return value, and `split-window` can split any window (not just the currently selected window), so you could use it to write code that doesn't rely on `windmove` or on changing the selected window.

Just as a taste, this code splits the current window into a 2x2 windows as well:
```elisp
(split-window (split-window-right) nil 'below)  ; split selected window to the right, then split that window below
(split-window-below)
```

sdwolfz:
@bmag That's some really nice elisp!

I was also thinking about implementing a variable that controls whether the filetree (neotree/treemacs) window is deleted or not. Anyway I can't wait to get home and play around with this.

mirubiri:
@bmag Nice! I don't mind to change the code it that works!

bmag:
> a variable that controls whether the filetree (neotree/treemacs) window is deleted or not

Maybe utilize the prefix argument for that? So `M-x layout-grid` doesn't delete neotree/treemacs but `C-u M-x layout-grid` deletes also neotree/treemacs.
(in Spacemacs it's `SPC SPC` and `SPC u` instead of `M-x` and `C-u`, I used the latter for the sake of being clear)

Also, you probably know that normally neotree/treemacs prevent their window from being deleted by `delete-other-windows`. However, If they are well behaved, then
```elisp
(let ((ignore-window-parameters t))
  (delete-other-windows))
```
will override the protections and delete also the neotree/treemacs window.

bmag:
> currently all the layout functions defined here remove the neotree window.

I haven't used them in a while, maybe I'm out of the loop...

fiveNinePlusR:
> currently all the layout functions defined here remove the neotree window.

They might do that now but should they do that moving forward? I am for leaving the file tree up as the standard behavior as default because that's more of a protected class of windows imo. Same with the imenu list on the right. having a prefix arg close everything would be a good compromise or maybe a even a toggle option?

sdwolfz:
Thanks everyone for the help and insights :heart:!
Keeping the filetree alive seems to be a bit tricky... I'll open up a PR with a possible solution and we can continue the discussion there.

PR #10677, opened 2018-05-14
Add recommended brew method to install emacs

athersharif:
Seems like `brew linkapps emacs-plus` doesn't work anymore. This PR updates the README with the updated recommended method from brew in order to install emacs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,8 +166,9 @@ This installs a pre-built package from https://emacsformacosx.com/
 ```
 brew tap railwaycat/emacsmacport
 brew install emacs-mac
-copy and paste the ln from the install instructions if brew linkapps is deprecated
 ```
+
+Copy and paste the symlink (`ln`) from the install instructions if `brew linkapps` is deprecated.
 
 ##### Other ways
 
--- a/README.md
+++ b/README.md
@@ -148,25 +148,46 @@ Emacs has graphical support.
 This might also be true for other linux distributions.
 
 ### macOS
-The recommended way of installing Emacs on macOS is using [Homebrew][]. It's a
-package manager for macOS. Once Homebrew is installed, run the following
-commands in the terminal to install both Emacs and the default Source Code Pro
-font:
+
+#### Install emacs
+
+##### Cask Homebrew Installation
+
+Homebrew now recommends to use the cask version with the following message: "Please try the Cask for a better-supported Cocoa version". To install the cask version:
+
+```
+brew cask install emacs
+```
+
+This installs a pre-built package from https://emacsformacosx.com/
+
+##### Using emacs-mac
+
+```
+brew tap railwaycat/emacsmacport
+brew install emacs-mac
+copy and paste the ln from the install instructions if brew linkapps is deprecated
+```
+
+##### Other ways
+
+If you're not comfortable with the ways mentioned above, then [EmacsWiki](https://www.emacswiki.org/emacs/EmacsForMacOS#toc12) lists down a few ways to install Emacs for Mac OS. 
+
+#### Install Source Code Pro font
+Once Emacs is installed, run the following commands in the terminal to install the default Source Code Pro font:
 
 ```sh
-brew tap d12frosted/emacs-plus
-brew install emacs-plus
-brew linkapps emacs-plus
 brew tap caskroom/fonts
 brew cask install font-source-code-pro
+```
+
+#### Install Spacemacs
+
+```
 git clone https://github.com/syl20bnr/spacemacs ~/.emacs.d
 ```
 
 **Notes:**
-The proposed `emacs-plus` tap is identical to the `emacs` formulae, it just
-builds GNU Emacs with support for several features by default, including the
-Spacemacs icon. See [emacs-plus][] for more information.
-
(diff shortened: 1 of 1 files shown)


fiveNinePlusR:
this needs to be rebased on develop. there are way too many commits here. 

as to your commit... why is this cask now the suggested way? i think emacs-mac is a better formula all around as it lets you install with a bunch of different versions and options.

fiveNinePlusR:
I think the wiki is out of date. the better way to install the gui emacs for cocoa is through `emacs-mac` from homebrew proper instead of the binary blobs that have fewer options.

syl20bnr:
But `emacs-plus` is still the superior/more comprehensive way to install
Emacs from `brew` no ? Or this is not the case anymore ?

No sure we should confuse people with another brew way to install Emacs
just because `brew linkapps emacs-plus` does not work anymore. Can we
just remove `brew linkapps emacs-plus` from the snippet and tell people
how to replace it or does it work only with `brew cask install emacs` ?

athersharif:
@fiveNinePlusR @syl20bnr i'm fine either way but here are my two cents:

i think what makes spacemacs so awesome is that it's a one stop shop and has the documentation to start a newbie from start to finish. imo, removing the linkapps with something else would make things a little bit complex. how would you guys think about just linking to https://www.emacswiki.org/emacs/EmacsForMacOS#toc12 for emacs installation so users can pick whatever they're comfortable with, and may be adding your preferred choice from that list?

thoughts?

athersharif:
@syl20bnr updated the PR. let me know if you guys want me to change it more. :)

fiveNinePlusR:
looks good to me. I think the best one is the emacs-mac tap, but that's a tap and is (very) slightly more complex to install from homebrew. (one extra step). Linking to that page at that part gives the user all the info they really need.

htc502:
As a user, I think this edition is not good enough. It makes me more confused:
1. I am redirected to another 'long list' (probably longer than spacemacs readme.) which I think is not quite user friendly.  
2. That wikipage also uses   'brew linkapps emacs' to 'finally link it to your Applications folder'. As I've tested, brew linkapps is already deprecated. This is the problem that the RP want to deal with. 

If you guys think emacs-mac is a good way to go, probably you can link the user to that specific part or just copy that part over here. 
To me, the following command saved me for linking the app to Application:
'ln -s /usr/local/Cellar/emacs/24.5/Emacs.app /Applications'
Not sure we can do this in readme or not.
Hope this is helpful. Thank you.

athersharif:
@htc502 i can totally see the confusion there. and you're right, this might be too big a documentation to link to and the wiki not supporting navigation by id doesn't help either. before we had the conversation above, i was initially in the favor of the following:

```
Cask homebrew installation
Homebrew now recommends to use the cask version with the following message: “Please try the Cask for a better-supported Cocoa version”

To install the cask version:

brew cask install emacs
This installs a pre-built package from https://emacsformacosx.com/
```

also, the symlink you posted above is for `24.5`. this version would have to be kept like `[version]` or something similar if we were to add this to the readme, imo.

@fiveNinePlusR @syl20bnr - thoughts? are we inclining towards `emacs-plus` or `emacs-mac`?

athersharif:
@fiveNinePlusR thanks for the great suggestion. really like it :).

i've updated the PR. thoughts?

fiveNinePlusR:
looks good to me. might want to downcase the ln statement though.

athersharif:
ah, sorry about that. updated.

syl20bnr:
What are the advantages of cask over the `emacs-plus` tap ?

I have the feeling that `emacs-plus` is more complete than cask.

See the `emacs-plus` readme here: https://github.com/d12frosted/homebrew-emacs-plus

I was more for keeping `emacs-plus` in the README and just add a
reference to emacsformacosx

For `emacs-mac` it was the promoted way to install Emacs but starting
with Emacs 25 we had lots of report of performance issues with it.

cc @d12frosted

athersharif:
@syl20bnr - added the emacs-plus instructions in there as well. also, probably should keep an eye on https://github.com/d12frosted/homebrew-emacs-plus/issues/41#issuecomment-311361453 and update the readme when this change is made in the future.

athersharif:
also, @syl20bnr @fiveNinePlusR - any idea why the Emacs24 build is failing?

syl20bnr:
What's the error ?

Also develop branch support Emacs 25+ only.

syl20bnr:
Nice work 👍 
Thank you !

aaronjensen:
Until 26.1 is released (which should be soon) this change will encourage people to install 25.3 which lacks the vfork patch which means significantly slower magit. It's too bad, but hopefully 26.1 will be out soon.

dijonkitchen:
When is the `develop` branch going to be merged into `master` for these new Readme changes? It's pretty confusing for a new user unless you've looked across a few issues.

tkothe:
Just checking whether this will be merged to `master` soon as I also just came across the issue of the lack of the `brew linkapps` command. Cheers!

cjoshmartin:
Brew suggested that I should run `brew link emacs-plus` and it work it!

PR #10678, opened 2018-05-14
Window layout with custom delete

sdwolfz:
Continuing the discussion from #10669.

This adds a variable `spacemacs-layout-delete-function` that controlls which function is used to delete other windows when calling one of the following:  
- `spacemacs/layout-grid`
- `spacemacs/layout-triple-columns`
- `spacemacs/layout-double-columns`
- `spacemacs/layout-single-column`

The idea is to keep the filetree when executing these functions but I'm not sure how to implement this. What I found is that treemacs has `treemacs-delete-other-windows` which does exactly this, so setting:
```emacs-lisp
(setq spacemacs-layout-delete-function 'treemacs-delete-other-windows)
```
In your `dotspacemacs/user-config` is a quick fix... but it does not work for netoree, and it only handles the treemacs window, meaning any other window that might be of interest is deleted.

Please review this PR and throw ideas at me :smile_cat:!

/cc everybody from #10669 @mirubiri @bmag @fiveNinePlusR

### DIFF
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -2067,10 +2067,6 @@ Windows manipulation commands (start with ~w~):
 | ~SPC w =~              | balance split windows                                                       |
 | ~SPC w _~              | maximize window horizontally                                                |
 | ~SPC w \vert~          | maximize window vertically                                                  |
-| ~SPC w 1~              | make 1 window layout                                                        |
-| ~SPC w 2~              | make 2 windows layout (split vertically)                                    |
-| ~SPC w 3~              | make 3 windows layout (split vertically)                                    |
-| ~SPC w 4~              | make 4 windows layout (split in 2x2 grid)                                   |
 | ~SPC w b~              | force the focus back to the minibuffer                                      |
 | ~SPC w c~              | maximize/minimize a window and center it                                    |
 | ~SPC w C~              | maximize/minimize a window and center it using [[https://github.com/abo-abo/ace-window][ace-window]]                   |
@@ -2105,6 +2101,50 @@ Windows manipulation commands (start with ~w~):
 | ~SPC w w~              | cycle and focus between windows                                             |
 | ~SPC w W~              | select window using [[https://github.com/abo-abo/ace-window][ace-window]]                                              |
 | ~SPC w x~              | delete a window and its current buffer (does not delete the file)           |
+
+Split the current window into multiple ones, deleting all others using the
+following commands:
+
+| Key Binding | Description                               |
+|-------------+-------------------------------------------|
+| ~SPC w 1~   | make 1 window layout                      |
+| ~SPC w 2~   | make 2 windows layout (split vertically)  |
+| ~SPC w 3~   | make 3 windows layout (split vertically)  |
+| ~SPC w 4~   | make 4 windows layout (split in 2x2 grid) |
+
+By default, the commands above ignore some windows, like the filetrees (treemacs
+and neotree) but you can use a prefix argument to force delete them.
+
+| Key Binding     | Description                                                 |
+|-----------------+-------------------------------------------------------------|
+| ~SPC u SPC w 1~ | make 1 window layout with force delete                      |
+| ~SPC u SPC w 2~ | make 2 windows layout (split vertically) with force delete  |
+| ~SPC u SPC w 3~ | make 3 windows layout (split vertically) with force delete  |
+| ~SPC u SPC w 4~ | make 4 windows layout (split in 2x2 grid) with force delete |
+
+You can also configure which windows get ignored when executing the above
+commands by adding a prefix to the =spacemacs-window-split-ignore-prefixes=
+list:
+
+#+BEGIN_SRC emacs-lisp
+(add-to-list 'spacemacs-window-split-ignore-prefixes "My Favourite Window")
+#+END_SRC
+
+Or if you want a certain ignored window to always be deleted you can remote it
+from the list:
+
+#+BEGIN_SRC emacs-lisp
+(setq spacemacs-window-split-ignore-prefixes
+  (remove treemacs--buffer-name-prefix spacemacs-window-split-ignore-prefixes))
+#+END_SRC
+
+And if you want even more control over the way windows are removed you can
+define your own deletion function:
+
+#+BEGIN_SRC emacs-lisp
+(defun my-delete-other-windows () (delete-other-windows))
+(setq spacemacs-window-split-delete-function 'my-delete-other-windows)
+#+END_SRC
 
 **** Window manipulation transient state
 A convenient window manipulation transient state allows performing most of the
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -559,30 +559,148 @@ in a split window to the right."
   (split-window-horizontally)
   (other-window 1))
 
-(defun spacemacs/layout-grid ()
-  " Set the layout to a 2x2 grid. "
-  (interactive)
-  (delete-other-windows)
-  (split-window (split-window-right) nil 'below)
-  (split-window-below))
+
+;; Window Split
 
-(defun spacemacs/layout-triple-columns ()
-  " Set the layout to triple columns. "
-  (interactive)
-  (delete-other-windows)
-  (dotimes (i 2) (split-window-right))
+(defun spacemacs--window-split-splittable-windows ()
+  (seq-remove
+    (lambda (window)
+      ;; TODO: find a way to identify unsplittable side windows reliably!
+      nil)
+    (spacemacs--window-split-non-ignored-windows)))
+
+(defun spacemacs--window-split-non-ignored-windows ()
+  "Determines the list of windows to be deleted."
+  (seq-filter
+    (lambda (window)
+      (let* ((name (buffer-name (window-buffer window)))
+              (prefixes-matching
+                (seq-filter
+                  (lambda (prefix) (string-prefix-p prefix name))
+                  spacemacs-window-split-ignore-prefixes)))
+        (not prefixes-matching)))
+    (window-list (selected-frame))))
+
+(defun spacemacs/window-split-default-delete ()
+  "Deletes other windows, except a list of excluded ones."
+  (if spacemacs-window-split-ignore-prefixes
+      (let* ((deletable (spacemacs--window-split-non-ignored-windows))
+              (splittable (spacemacs--window-split-splittable-windows)))
+        (when splittable
+          (let* ((selected (car splittable))
+                  (to-delete (delq selected deletable)))
+            (select-window selected)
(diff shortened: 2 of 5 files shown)


fiveNinePlusR:
you could copy the function from treemacs and modify it to look for a few special patterns.

... I did that and this is what I came up with... seems to work pretty well except it doesn't check if you are invoking that function from inside a treemacs or neotree window before deleting the other windows it doesn't break anything in treemacs but it also won't work correctly to split the window as the treemacs window is not divisible.

```elisp 
(defvar spacemacs--layouts-save-filetree-onupdate t
  "When this variable is set to t, the layout functions will not close
   the filetree windows when updating layout")

(defun spacemacs--delete-other-windows ()
  "Same as `delete-other-windows', but will not delete the filetree windows when `spacemacs--layouts-save-filetree-onupdate` is t."
  (interactive)
  (let ((w (selected-window)))
    (--each (window-list (selected-frame))
      (unless (or (eq it w)
                  (if (and spacemacs--layouts-save-filetree-onupdate (boundp 'treemacs--buffer-name-prefix))
                      (->> it (window-buffer) (buffer-name) (s-starts-with? treemacs--buffer-name-prefix)))
                  (if (and spacemacs--layouts-save-filetree-onupdate (boundp 'neo-buffer-name))
                      (->> it (window-buffer) (buffer-name) (s-starts-with? neo-buffer-name))))
        (delete-window it)))))

 ```

**Comment by fiveNinePlusR** [comment]
One thing that this could be extended to is have a list of buffer names that you iterate through and check against and that would be general enough to just add-to-list "somebufferprefix" that you don't want to delete.

**Comment by sdwolfz** [comment]
@fiveNinePlusR That snippet of code is exactly what I needed to get started, thanks!

@Miciah I read that `no-delete-other-windows` was introduced in emacs 26. But I will keep it in mind + the heuristics you suggested for future improvements.

**Comment by sdwolfz** [comment]
Note to self: a future improvement could be made to open the last N buffers (that have a file associated) inside the window splits instead of having the same buffer opened in all of them.

**Comment by fiveNinePlusR** [comment]
> Note to self: a future improvement could be made to open the last N buffers (that have a file associated) inside the window splits instead of having the same buffer opened in all of them.

That would be much more useful than showing the same buffer, yeah.

**Comment by sdwolfz** [comment]
Finished implementing all the features I wanted. I'll leave this here over the weekend in case someone wants to review/test it, and I'll refine the documentation.

Thank you @fiveNinePlusR @Miciah for jumping in to help!

**Comment by fiveNinePlusR** [comment]
I really like it! super useful and feels natural.

File: layers/+spacemacs/spacemacs-defaults/funcs.el [hunk 1]
```diff
@@ -559,30 +559,134 @@ in a split window to the right."
...
+  "Set the layout to a 2x2 grid.
+
+Uses the funcion defined in `spacemacs-window-split-delete-function' as a means to
+remove windows.
+
+When called with a prefix argument, it uses `delete-other-windows' as a means
+to remove windows, regardless of the value in `spacemacs-window-split-delete-function'."
+  (interactive "P")
+  (if purge
+    (delete-other-windows)
```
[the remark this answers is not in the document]
sdwolfz: Hmm... whis is how evil lisp state does formatting.

sdwolfz:
Thank you @bmag, I fixed some of the items pointed by you in the review, will do the rest a bit later.

Tested with emacs 25.3 in termux android and emacs 27.

layers/+spacemacs/spacemacs-defaults/funcs.el
```diff
@@ -559,30 +559,134 @@ in a split window to the right."
...
-  (split-window (split-window-right) nil 'below)
-  (split-window-below))
+

+;; Window Split
+
+(defun spacemacs/window-split-delete-function ()
+  "Deletes other windows, except a list of excluded ones."
+  (if spacemacs-window-split-ignore-prefixes
+    (let* ((current-window (selected-window))
```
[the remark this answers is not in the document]
sdwolfz: Took me a while to understand that `thread-last` was not reated to emacs 26 threads :smile_cat:. Will keep in mind for next time when I'll have code like that. I refactored that code not to need such functionality.

sdwolfz:
The only thing I don't like right now... when treemacs is the only window and I try to split it, it gives me: `split-window: Cannot split side window or parent of side window`. But I can split it horizontally (by accident) when doing the 2x2 grid... I'll have to see how this can be handled.

layers/+spacemacs/spacemacs-defaults/funcs.el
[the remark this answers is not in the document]
sdwolfz: From the documentation I see that passing `t` to `delete-other-windows` should have the same effect. I preffer passing `t` over dynamic binding.

sdwolfz:
Looks like there is no "fits all" solution for splitting a side window, so I will display a message saying there is no main window available to split.

Ready for another round of reviews :bowing_man:!

doc/DOCUMENTATION.org
```diff
@@ -2106,6 +2106,15 @@ Windows manipulation commands (start with ~w~):
 | ~SPC w W~              | select window using [[https://github.com/abo-abo/ace-window][ace-window]]                                              |
 | ~SPC w x~              | delete a window and its current buffer (does not delete the file)           |
 
+You can use a prefix argument with some keybindings to make them more potent:
```
syl20bnr: What does potent mean exactly ? What additional power do they get ? :-)

sdwolfz: I king of ignored the documentation until the last moment. Will rewrite this.

layers/+spacemacs/spacemacs-defaults/funcs.el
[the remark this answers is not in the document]
sdwolfz: I was wrong about passing `t` to `delete-other-windows`, misread the documentation.

layers/+spacemacs/spacemacs-defaults/funcs.el
```diff
@@ -559,30 +559,138 @@ in a split window to the right."
...
+(defun spacemacs/window-split-grid (&optional purge)
+  "Set the layout to a 2x2 grid.
+
+Uses the funcion defined in `spacemacs-window-split-delete-function' as a means to
+remove windows.
+
+When called with a prefix argument, it uses `delete-other-windows' as a means
+to remove windows, regardless of the value in `spacemacs-window-split-delete-function'."
+  (interactive "P")
+  (let ((ignore-window-parameters t))
```
bmag: You really want to bind `ignore-window-parameters` only for `delete-other-windows`. Really possible that `spacemacs-window-split-delete-function` wants respect window paramters.
Same goes for the other occurrences.

layers/+spacemacs/spacemacs-defaults/funcs.el
```diff
@@ -559,30 +559,138 @@ in a split window to the right."
...
+    (window-list (selected-frame))))
+
+(defun spacemacs/window-split-default-delete ()
+  "Deletes other windows, except a list of excluded ones."
+  (if spacemacs-window-split-ignore-prefixes
+      (let* ((to-delete (spacemacs--window-split-non-ignored-windows)))
+        (when to-delete
+          (select-window (car to-delete))
+          (dolist (window (cdr to-delete)) (delete-window window))))
+    (delete-other-windows)))
```
bmag: IIRC if selected window is a side window, this throws an error about not being able to make a side window the only window, or something like this.
Function `window-at-side-p` checks if a window is a side windows.

layers/+spacemacs/spacemacs-defaults/funcs.el
```diff
@@ -559,30 +559,138 @@ in a split window to the right."
...
+                  spacemacs-window-split-ignore-prefixes)))
+        (not prefixes-matching)))
+    (window-list (selected-frame))))
+
+(defun spacemacs/window-split-default-delete ()
+  "Deletes other windows, except a list of excluded ones."
+  (if spacemacs-window-split-ignore-prefixes
+      (let* ((to-delete (spacemacs--window-split-non-ignored-windows)))
+        (when to-delete
+          (select-window (car to-delete))
```
bmag: If `(car to-delete)` is a side window, could face errors as noted in comment below. Maybe select the first window from `to-delete` that isn't a side window?

bmag:
> when treemacs is the only window and I try to split it, it gives me: `split-window: Cannot split side window or parent of side window`

Treemacs doesn't create itself as a side window anymore, so this is odd behavior. Maybe you're using an old version of treemacs.

> But I can split it horizontally (by accident) when doing the 2x2 grid

In `window.el` there's a `window--check` function that is called from several places (including `split-window`, `delete-window`, `delete-other-windows`), and can remove the "side window" status from all side windows if "something is not right" (it checks for consistency between window layout and "side window" status of windows).
In Emacs 25.3 the change is done silently, in 26(rc1) it also throws an error, and in 27(devel) I don't know.

Look at my latest comments, the problems maybe are caused by the split function not accounting for side windows.

layers/+spacemacs/spacemacs-defaults/funcs.el
sdwolfz: I can not make it work with `window-at-side-p`. Looks like this chekcs if a window is located at one of the sides, but does not actually have anything to do with the `window-side` window parameter...

sdwolfz:
I do have the latest version of treemacs. No idea why I get that error (emacs 27). Also I am unable to find a reliable way to tell if a window can be split or not, neither can I tell if a window is a side window or not.

Some more digging into splitting a side window... aparently when the treemacs window is the only window, it can not be split using `(split-window-right)` but it can be split using `(split-window (selected-window) nil 'right)` (I found this in the treemacs source code). I can not do the same for neotree though, it can not be split when it's the only window, not even with `(split-window (selected-window) nil 'right)`, but It can be split if it's not the only window.

This has the potentiall to get verry messy trying to handle all these edge cases... I suggest merging this in without handling side windows, and in case someone opens a bug report with an actual use case then I'll look into it further.

fiveNinePlusR:
emacs already does the wrong thing in a lot of cases anyway so I say it's worth merging now with a todo and a new issue that references this one.

bmag:
> This has the potentiall to get verry messy trying to handle all these edge cases... I suggest merging this in without handling side windows, and in case someone opens a bug report with an actual use case then I'll look into it further.

Fine by me. Although if you judge `spacemacs/window-split-default-delete` as it is now as too problematic/complicated, consider setting `delete-other-windows` as the default delete function, and maybe treemacs layer can then set the default to `treemacs-delete-other-windows`.

sdwolfz:
It's not problematic as it is now, it just does not satisfy my inner desire to create perfectly working software :smile:. I'll have to spend some more time digging trough emacs's guts to figure a way of handling unsplittable windows, but until then this should be a strict improvement of the current functionality.

sdwolfz:
Merged this into develop. If someone reports problems with the functionality I'll jump right in to help.

PR #10681, opened 2018-05-15
Load evil-escape only if not in holy-mode

### LINKED ISSUE #10667: `evil-escape` should not be loaded in `emacs` editing style
#### Description :octocat:
`evil-escape` should not be loaded in `emacs` editing style

- Set `dotspacemacs-editing-style` to `'emacs` in `.spacemacs`
- Start Emacs
- Type `fd` rapidly

The typed letters disappear.

Editing works as in vanilla Emacs, and letters are written.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 25.2.2
- Spacemacs: 0.200.9
- Spacemacs branch: master (rev. 8e1af145)
- Graphic display: t
- Editing style: emacs
- Completion: ivy
- Layers:
```elisp
(ocaml asciidoc ruby go python html nginx rust haskell julia ansible yaml ivy emacs-lisp git markdown theming org version-control)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GSETTINGS NOTIFY ACL LIBSELINUX GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11

smile13241324:
Fix #10667 by not loading `evil-escape` when editing style is neither `vim` nor `hybrid`.

### DIFF
--- a/layers/+spacemacs/spacemacs-evil/funcs.el
+++ b/layers/+spacemacs/spacemacs-evil/funcs.el
@@ -41,6 +41,15 @@ Otherwise, revert to the default behavior (i.e. enable `evil-insert-state')."
     (evil-put-property 'evil-state-properties 'iedit-insert
                        :enable states)))
 
+(defun spacemacs//evil-escape-deactivate-in-holy-mode(style)
+  "Deactivate `evil-escape' if STYLE is emacs otherwise enable it."
+  (cond
+   ((or (eq 'vim style)
+        (eq 'hybrid style))
+    (evil-escape-mode t))
+   (t
+    (evil-escape-mode -1))))
+
 
 ;; evil-search-highlight-persist
 
--- a/layers/+spacemacs/spacemacs-evil/packages.el
+++ b/layers/+spacemacs/spacemacs-evil/packages.el
@@ -96,8 +96,10 @@
 
 (defun spacemacs-evil/init-evil-escape ()
   (use-package evil-escape
-    :init (evil-escape-mode)
-    :config (spacemacs|hide-lighter evil-escape-mode)))
+    :init (progn
+            (spacemacs|hide-lighter evil-escape-mode)
+            (spacemacs//evil-escape-deactivate-in-holy-mode dotspacemacs-editing-style)
+            (add-hook 'spacemacs-editing-style-hook #'spacemacs//evil-escape-deactivate-in-holy-mode))))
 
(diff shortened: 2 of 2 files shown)


sdwolfz:
Hmm... I'm not sure about this, how does this play along with `C-z` for switching between editing styles?

smile13241324:
I have played a bit around with `C-z`. When my `dotspacemacs-editing-style` is `vim` or `hybrid` I can still switch to `emacs` style with `C-z`. When I enter `fd` I escape from `emacs` style and I am back in `vim` style. So this still works as expected.

When my `dotspacemacs-editing-style` is `emacs`, `C-z` does not change to `vim` but stays in `emacs` style. This however seems to be independent of my change. When it would work `evil-escape` would stay deactivated. However even when we would reload all affected packages the value for `dotspacemacs-editing-style` would still be `emacs` thus we would load non of the `evil` specific packages. So in this case deactivating `evil-escape` should not create any new problems.

syl20bnr:
Spacemacs allows to dynamically switch between editing style so we have
to retain the functionalities even if the style are not used.

It means that we should always install `evil-escape` but disable it when
switching the editing style. This can be done by creating a function
that takes one argument which is the style and hook this function to
`spacemacs-editing-style-hook`

You can find an example here:
- function definition: https://github.com/syl20bnr/spacemacs/blob/develop/layers/+distributions/spacemacs-bootstrap/funcs.el#L75-L90
- hook: https://github.com/syl20bnr/spacemacs/blob/develop/layers/+distributions/spacemacs-bootstrap/packages.el#L60-L61

Note that after setting up the hook we directly call the function, this
is to do the first initial setup at startup with the style of the user.

sdwolfz:
@syl20bnr I like that approach :+1: . This also means the package is always installed, no matter what editing style is used, so switching styles does not trigger package deletion/download.

smile13241324:
Ok I'll adjust my code in the next view days.

sdwolfz:
Thank you :heart:!

syl20bnr:
@smile13241324 Nice!

I simplified the code a bit in https://github.com/syl20bnr/spacemacs/commit/1374cd615
Also the lighter must be hidden in :config instead of :init

PR #10703, opened 2018-05-19
org: add key binding for org-reveal

p-sdk:


syl20bnr:
What does org-reveal exactly ? I feel it should not take a precious
top level `SPC m r` major mode prefix and it should be under antoher
prefix where we can group similar key bindings.

Compro-Prasad:
It might be useful. But I can't find a way to use it.

p-sdk:
The command shows the current entry, its ancestors and siblings. It can be used
to show the context when the buffer contents are folded after jumping to the
entry from an agenda or a link.

https://www.gnu.org/software/emacs/manual/html_node/org/Global-and-local-cycling.html

Originally it's bound to `C-c C-r`. Other key binding may be `SPC m s R`, but
it's two key strokes more than `SPC m r`.

syl20bnr:
mmmh I tried it and as @Compro-Prasad I don't know how to use it
effectively, 9 out of 10 times it does nothing. I'm a bit confused :-(

Compro-Prasad:
@p-sdk I am a org-mode user and I don't think this feature is ground breaking or important to share with the community. If you have mastered it then you can use it personally. In future if it becomes more useful or the demand rises then it will definitely be considered. But for now I don't think this is a feature worth to share with us. No offence but you can take down this PR so that we can reduce the burden on @syl20bnr .

PR #9550, opened 2017-09-06
New powerline colour space fix for osx

CarlQLange:
https://github.com/syl20bnr/spacemacs/issues/4426

The old fix for this modeline issue was to change the emacs colour space, but that kinda sucked because it would change all of the colours in the app, and everything would look a little bit off. This new fix, added to powerline in this commit: https://github.com/milkypostman/powerline/commit/8a246902e86a0c59015bb897a9c59be9729ef5c4 , instead changes the separator image colours to Apple RGB.

Sorry, this is a bad commit message, [I'm literally in a tent](https://toughsoles.ie)

PR #10704, opened 2018-05-19
Fix modeline colors on emacs-mac port

p-sdk:
In 665c09b modeline separators colors have been fixed for cocoa emacs, but it has broken colors on emacs-mac port. This commit sets `powerline-image-apple-rgb` to `t` only if `window-system` is `ns`, as suggested in https://github.com/syl20bnr/spacemacs/pull/9550#issuecomment-365807703


syl20bnr:
Thank you again ! 👍 

PR #10715, opened 2018-05-21
Add 'company-irony' to c-c++ layer for auto-completion

farva:
Add the clang-based auto-completion service 'Irony' and it's 'company'
backend. In order to work properly, 'company-c-headers' is now added
to the 'company-backends' in the 'port-init' phase, so to come before
'company-irony' in the list.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -109,5 +109,6 @@ This file containes the change log for the next major version of Spacemacs.
 **** C-C++
 - Add possible value =no-completion= to =c-c++-enable-rtags-support= flag.
   This adds the option to opt-out of =company-rtags= while enabling Rtags.
+- Add support for =irony= with =c-c++-enable-irony-support=.
 *** Various improvements
 - Various documentation improvements (thanks Carl Lange, Diego Berrocal, Wieland Hoffmann)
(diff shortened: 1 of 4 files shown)


Compro-Prasad:
Is it necessary? We already have so many backends for auto completion.

farva:
@Compro-Prasad , I can only say that it was necessary for my work environment, since `irony` works best for my projects. And since it might be useful to others, I added it to the public `c-c++` layer.

syl20bnr:
We need to modernize the c-c++ layer like we did recently for javascipt.
Right know this layer is all over the place.

As I did for javascript, I'll allocate some time to work on the c-c++
layer.

Please, keep this PR open, I'll merge it and adapt it.

thyeme:
Thanks for the layer configuration with irony!
When i use auto-completion with c-c++ layer i have issues, that irony fixes.
So far i have added irony in the dotspacemacs/user-config section, with hooks to the c-c++ mode.

This is much more elegant. I'll try to use it and integrate in my emacs configuration
Thank you

thyeme:
I still encounter an issue though..

When Im working on c++ files, sometimes irony mode just stops working, apparently randomly. 

I have no idea what triggers its disabling. I use multiple windows on my buffer with different .C .cxx .h files. I dont know if this could have an effect.. Does anyone have the same problem?

PR #10721, opened 2018-05-21
Add ubuntu section in experimental doc (for pdumper)

shwsun:
Documentation for getting pdumper work with Ubuntu. I am using ubuntu 17.10 but I think it should work for most of the up-to-date Ubuntu. It would be great if someone can add the labels for it.

Happy to take suggestions etc. 

Additionally, perf improvement:
        `vanila: 5.263 sec -> 1.509 sec`

### DIFF
--- a/EXPERIMENTAL.org
+++ b/EXPERIMENTAL.org
@@ -60,7 +60,7 @@ In the root directory of your freshly cloned Emacs repository and with the
 #+END_SRC
 
 **** Ubuntu
-/Note: This tested on Ubuntu 17.10./
+/Note: This is tested on Ubuntu 17.10./
 
 In the root directory of your freshly cloned Emacs repository and with the
 =pdumper= branch as the current branch:
@@ -189,5 +189,3 @@ command:
 #+BEGIN_SRC shell
   ./emacs --dump-file=/Users/sylvain/.emacs.d/.cache/dumps/spacemacs.pdmp &
 #+END_SRC
-
-
--- a/EXPERIMENTAL.org
+++ b/EXPERIMENTAL.org
@@ -60,6 +60,8 @@ In the root directory of your freshly cloned Emacs repository and with the
 #+END_SRC
 
 **** Ubuntu
+/Note: This tested on Ubuntu 17.10./
+
 In the root directory of your freshly cloned Emacs repository and with the
 =pdumper= branch as the current branch:
 
--- a/EXPERIMENTAL.org
+++ b/EXPERIMENTAL.org
@@ -80,7 +80,7 @@ need to install the following packages:
    sudo apt-get install libtiff5-dev libgif-dev libpng-dev libxpm-dev libgtk-3-dev libgnutls28-dev
 #+END_SRC
 
-*** Install Emacs 
+*** Install Emacs
 After the compiling finished successfully, you will need to install Emacs.
 
 #+BEGIN_SRC shell
--- a/EXPERIMENTAL.org
+++ b/EXPERIMENTAL.org
@@ -7,6 +7,7 @@
     - [[#simple-patch-of-emacs-source-code][Simple patch of Emacs source code]]
     - [[#compile-emacs-from-source][Compile Emacs from source]]
       - [[#macos][MacOS]]
+      - [[#ubuntu][Ubuntu]]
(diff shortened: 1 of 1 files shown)


EXPERIMENTAL.org
```diff
@@ -52,11 +53,28 @@ your OS).
...
   ./autogen.sh
   ./configure --with-ns --with-dbus --with-gnutls --with-imagemagick --with-rsvg  --with-mailutils --with-xml2 --with-modules
 #+END_SRC
 
+If you have never compile Emacs from source on your machine then you probably
+need to install the following packages:
+
+#+BEGIN_SRC shell
+   sudo apt-get install build-essential automake texinfo libjpeg-dev libncurses5-dev
+   sudo apt-get install libtiff5-dev libgif-dev libpng-dev libxpm-dev libgtk-3-dev libgnutls28-dev
```
bmag: Better command to install almost all the necessary libraries:
```
apt-get build-dep emacs
```
I don't know exactly which packages aren't covered by it, though.

syl20bnr:
@shwsun do you have an SSD ?

1.5sec seems a lot. How many packages do you have?

Im my case I have ~380 packages which take approx. 10sec with
conventional loading. It takes ~0.5s with a dump.

shwsun:
@syl20bnr I have a NVMe SSD. Without pdumper it took around 6.08 sec for 395 packages. What did you config to make it to 0.5 sec? Did you just use the default config in the `.spacemacs` ?

syl20bnr:
Oh yeah, I did not think about that, this is surely your additional
setup in the dotspacemacs/user-init and dotspacemacs/user-config
functions.

Mine are almost empty.

If you have some `(require ...)` or `use-package` not defered in your
`user-config` function you can put them in `dotspacemacs/user-load`
functions and the required libraries will be part of the dump.

EXPERIMENTAL.org
shwsun: I don't think that actually works. I assume  the command  will look for dependencies of deb package `emacs` right? That will point to Emacs 25.

bmag: right, but it still covers most of the dependencies IIRC

shwsun:
@syl20bnr Hot darn! I have tones of stuff in `user-config` and moving them to `user-load` gives me a 0.687 sec for 395 packages now.

shwsun:
@bmag @syl20bnr Do you think we could provide a snap package of pdump enabled emacs for linux users? I know that Ubuntu try to get people use snap instead of deb recently. And it looks like snap package actually cross distro (see https://snapcraft.io/)

syl20bnr:
@JAremko is thinking about it and has done lots of research on snap
appimage, flatpack etc...
I even believe he is experimenting with it.

EXPERIMENTAL.org
shwsun: Last time I checked, the current `emacs` in Ubuntu's repo is Emacs 25. I am not sure building dependencies for Emacs 25 would be very helpful since this is Emacs 27.

shwsun:
@syl20bnr I tried to polish the workflow a little bit. Let me know your thoughts.

EXPERIMENTAL.org
bmag: I don't have an Ubuntu machine myself to check exactly what libraries are installed, but I do expect it to be very helpful for Emacs 27 as well. Anyway, I'm not going to insist on changing the instructions. If you believe current instructions are good enough, lets roll with it.

EXPERIMENTAL.org
```diff
@@ -57,6 +59,36 @@ In the root directory of your freshly cloned Emacs repository and with the
...
+
+If you have never compile Emacs from source on your machine then you probably
+need to install the following packages:
+
+#+BEGIN_SRC shell
+   sudo apt-get install build-essential automake texinfo libjpeg-dev libncurses5-dev
+   sudo apt-get install libtiff5-dev libgif-dev libpng-dev libxpm-dev libgtk-3-dev libgnutls28-dev
+#+END_SRC
+
+*** Install Emacs
```
JAremko: trailing white-space

shwsun:
@JAremko Thx!

PR #10730, opened 2018-05-23
Enable command history navigation in the cider REPL in an insert mode

radious:
This is the same navigation that is used in normal mode. This behavior matches the
one from the elisp REPL.


sdwolfz:
Thank you :heart:!

PR #10636, opened 2018-05-01
cmake: fix cmake-ide mode hook function name

TheJJ:
`cmake-ide` has changed the hook name:
https://github.com/atilaneves/cmake-ide/commit/26605efce2c866fcf507bf03869228e695294721#diff-5718e36b5fe0a5af87d4af78e6eaa88dR247

PR #10736, opened 2018-05-23
Replace cmake-ide--mode-hook with cide--mode-hook

Compro-Prasad:
The cmake-ide package update breaks the configuration because of the change in
`cmake-ide--mode-hook` function name.


sdwolfz:
I ended up merging #10636 since it contains the exact same change but was opened first. But I took your PR as a confirmation that the change was OK.

Thank you! :heart: 
You can safely delete your branch!

PR #10745, opened 2018-05-25
Add key bindings for org-feed-update-all and org-feed-goto-inbox

seemyvest:
Add global and layer key bindings for `C-c C-x g` and `C-c C-x G` (org-feed-update-all & org-feed-goto-inbox, respectively).


Compro-Prasad:
Seems like a nice addition @syl20bnr . It is part of `org-plus-contrib` thus a default org mode function and the keybinding is nice too. See for yourself to confirm.

seemyvest:
I don't mind either way, both make sense to me.

syl20bnr:
We try to not put key bindings under `C-c`.
The only place where I agree to put them globally is under `SPC a o`
which is application/org.

So `SPC a o f u` for instance.

syl20bnr:
Thank you ! 👍 

PR #10746, opened 2018-05-25
Revert "Move yas-activate-extra-mode setup to javascript layer"

aaronjensen:
This reverts commit e4eca01e827b68447604e0a7a0b7163662334e29.

This made js snippets unavailable in typescript and react modes. They need to be
activated in a hook when the mode starts, not after yasnippet is loaded.

### DIFF
--- a/layers/+lang/javascript/packages.el
+++ b/layers/+lang/javascript/packages.el
@@ -30,9 +30,7 @@
                            :repo "emacs-lsp/lsp-javascript"))
         skewer-mode
         tern
-        web-beautify
-        yasnippet
-        ))
+        web-beautify))
 
 (defun javascript/post-init-add-node-modules-path ()
   (spacemacs/add-to-hooks #'add-node-modules-path '(css-mode-hook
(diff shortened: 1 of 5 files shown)


layers/+lang/javascript/packages.el
```diff
@@ -30,9 +30,7 @@
                            :repo "emacs-lsp/lsp-javascript"))
         skewer-mode
         tern
-        web-beautify
-        yasnippet
```
sdwolfz: Doesen't this remove snippet functionality from the `javascript` layer? (for example when developing for nodejs without typescript/react).

aaronjensen: It shouldn't, no. Snippets are enabled via the `auto-completion` layer. All of the stuff that this does only has to do with enabling js-mode's snippets in other, non-js-mode-derived modes. yasnippet enables js-mode snippets in js-mode derived modes automatically.

syl20bnr:
Thank you ! 👍 

PR #10748, opened 2018-05-25
add all yasnippet-snippets

moisesroj0:
I add all the code of yasnippet-snippets


moisesroj0:
I don't understand how to solve some of the formatting problems,  you can accept the commit and solve it.

**layers/+completion/auto-completion/local/snippets/html-mode/form:1: trailing whitespace.**

don't show in red the problem.

sdwolfz:
@moisesroj0 what is the problem you're trying to solve in this PR? 

Yous hould have all those snippets by default when you use the `auto-completion` layer. Also you can have them by adding `yasnippet-snippets` in your `dotspacemacs-additional-packages` list.

moisesroj0:
I have the `auto-completion` and `yasnippet` enabled `M-m t a` but when I want to see the snippets `M-m i s` there is none snippets.

I have the same config in fedora 28 than the one I use in fedora 27 (I save the config in Mega)
but in fedora 28 yasnippet don't works.
I see the path of the snippets and It's empty.
`~/.emacs.d/elpa/yasnippet-xxxxx/snippets`
`~/.emacs.d/layers/auto-completion/snippets`
`~/.emacs.d/private/snippets`
And I know how this works, actually you can see my profile and I have the repo of yassnipet, and I work and that creating and improving some snippets.
https://github.com/moisesroj0/yasnippet-snippets/commits?author=moisesroj0
And I use spacemacs, since 2 years ago.

So... I was thinking if have yasnippet installed and don't works (because the folder it's empty) I will add the snippets in `layers/+completion/auto-completion/local/snippets/` And works.
I could put the snippets in.
`~/.emacs.d/private/snippets`
If that work better for you.
But I have not Idea why yasnippet snippets folder it's deleted in fedora 28.

moisesroj0:
In short, If I install yasnippet it download the version:
`yasnippet-snippets-20180503.657` with the snippets folder. but when restart the folder it's deleted and just 
`yasnippet-20180513.1004` keeps, but this don't have inside the snippets folder. just the `.el` files.

Compro-Prasad:
So, you get the point. It is provided by the package. If you include it into spacemacs itself then you will have to update it separately. Spacemacs has code for setting up packages, not including them. Because it will be hard to synchronize if something in the package changes.
Your problem is that you can't get the package `yasnippet-snippets*`. Try solving this problem instead. Try updating spacemacs or tweak your config.

moisesroj0:
I will close the P_Request, and solve it by downloading the yasnippet snippets official repo, and merge the `snippets` folder to `.emacs.d/elpa/yasnippet-XXXX`. Normally that is the way this works right?, Or however I know how to solve it. So just will Close the P_Request.

Compro-Prasad:
For me nothing is required. It is almost a zero config thing. I have literally nothing in my config to include the snippets separately. As I pointed out, try a fresh install and switch to develop to use it. Try to see anything in your configuration that removes yasnippet-snippets dependency. There is also a `dotspacemacs-additional-packages` variable which you can set to prevent the removal of the package.

moisesroj0:
`dotspacemacs-additional-packages '(yasnippet-snippets)` works, thanks. But I don't know why it deletes the folder. I have the std installation, all what I do it's remove the replace the files of emacs by, spacemacs. and enable:
```
     (c-c++ :variables
            c-c++-default-mode-for-headers 'c++-mode
            c-c++-enable-clang-support t)
     auto-completion
     helm
     emacs-lisp
     syntax-checking
```
there is no reason to delate the folder. however, it works with `dotspacemacs-additional-packages '(yasnippet-snippets)` thanks.

PR #10756, opened 2018-05-27
filetree: Update treemacs layer.

Alexander-Miller:
Some basic treemacs housekeeping. You already caught some of the new v2 changes, but there's more to be done:

 * Added some useful bits to the feature list.
 * Added a keybind for `find-tag`
 * Updated list of key binds
 * Removed obsolete treemacs-projectile code

### DIFF
--- a/layers/+filetree/treemacs/README.org
+++ b/layers/+filetree/treemacs/README.org
@@ -25,8 +25,13 @@ readme]]. In a short summary Treemacs offers the following:
 - Simple and powerful navigation and ability to detail exactly how and where a
   file should be opened
 - Good looking png icons.
+- Display of multiple file trees organized as projects residing in a workpsace.
 - Ability to show tags contained in files. Tags are provided by imenu, so
   nearly every filetype is supported.
+- Mouse interface for single and double left clicks in line with modern GUI standards.
+  (Clicking on an icon will also display a file's tags)
+- Locational awareness: commands like ~find-file~ or ~magit-status~ will use the location
+  of the node at point (with $HOME as fallback)
 - Optionally fontifying files based on their git status.
 - Optionally collapsing single-dir-child directories into one.
 - Doing both asynchronously for an imperceptible performance cost.
@@ -84,10 +89,9 @@ Default value is 3 (or 0 when python is not installed).
 | ~M-0~       | Selected the treemacs window. Actually selects window #10, which is always assigned to treemacs.                                               |
 | ~SPC 0~     | Open treemacs in the current directory. When not visiting a file use $HOME as fallback. With a prefix arg manually select the root instead.    |
 | ~SPC f B~   | Find and select a bookmark. If it cannot be found rebuild view with the bookmark's location as root. Also open the bookmark with a prefix arg. |
-| ~SPC f p~   | Same as ~SPC f t~, but when creating a new buffer open it in the current project's root.                                                       |
-| ~SPC f P~   | Open treemacs in the current project's root. With a prefix arg manually select the project instead.                                            |
 | ~SPC f t~   | Hide/show existing treemacs buffer. Create one for the current directory if no buffer exists.                                                  |
 | ~SPC f T~   | Manually focus the treemacs view on the currently selected file. Not needed when =treemacs-follow-mode= is enabled.                            |
+| ~SPC f M-t~ | Manually focus the treemacs view on the currently selected file and tag.                                                                       |
 
 ** Inside Treemacs
 Treemacs will use either ~j/k~ or ~n/p~ to go to the next/previous line,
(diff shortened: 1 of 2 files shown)


sdwolfz:
Thank you :heart:!

The `SPC f M-t` keybinding feels a bit weird to me, how about in a future PR we change `SPC f t` to mean filetree toggle (as it is now) and `SPC f T` to be a prefix dedicated to the filetree, where all the other goodies reside, this way `treemacs-find-tag` could have an easyer to type `SPC f T t` keybinding (since you don't have to release `Shift` if you don't have `T` bound to anything and no need to press `Meta`).

Alexander-Miller:
> we change SPC f t to mean filetree toggle (as it is now)

Not sure what you mean here. SPC f t is a toggler, as `treemacs-toggle` was in v1. Or do you mean it should also toggle neotree when it's used instead of treemacs?

> SPC f T to be a prefix dedicated to the filetree, where all the other goodies reside

The M-t bind came about because find-file used to be SPC f C-t, so I figured it's close enough. A prefix sounds alright to me. It's `fTT` for find-file and `fTt` for find-tag then? And if you want to move all the goodies there what about bookmark, should that be moved to `fTb`?

sdwolfz:
`SPC f t` should remain as it is now, I meant toggle as in showing/hiding the filetree (neotree has the same behaviour).

For the `SPC f T` keybindings I'm thinking about:
- `b` treemacs-bookmark
- `f` treemacs-find-file
- `t` treemacs-find-tag

Maybe also:
- `d` treemacs-delete-other-window

And others I don't know about yet... I would not want any of them to be uppercase letters, it's easy to type `SPC f T` and foget your finger is still pressing `Shift` while pressing your last key, and you would get the wrong function called by mistake.

Alexander-Miller:
Makes sense. I'll put this on my todo list.

syl20bnr:
I'm not satisfied with the current key bindings, at least I think that
`SPC f t` should not be a toggle. It should just open the file tree and
focus it and we could remove it from `SPC 0`, i.e. it would be ignored
by winum.

Then `treemacs-find-file` could be `SPC f t` and the file tree actions
would be behind `SPC f T`.  `SPC f t` could be another DWIM function.

Other possibility is to put the DWIM function that show/focus the file
tree under `SPC f t t` and then everything would be under `SPC f t`.

Alexander-Miller:
> and we could remove it from `SPC 0`, i.e. it would be ignored by winum.

That's already the case. `SPC 0` and `M-0` call `treemacs-select-window` which select the treemacs window regardless of winum and will also summon the window if it is not visible or does not exist.

> `SPC f t` could be another DWIM function.

find-file is no replacement for `treemacs`. find-file will ask you for the file to find if the current buffer is not visiting a file and I imagine people will also want toggle treemacs in scratch and magit buffers without that popup. find-file will also complain and show you a no-can-do pulse if you call it on a file that's outside the projects in your workspace. 

I think we need to keep both `treemacs` and select-window around and separate. The former is a toggler, it shows the file tree or makes it go away. The latter will *always* select it. You need some quick select keybind if you want to quickly navigate in the filetree and keeping it on M-0 keeps it consistent with the rest of the winum-based window-selection.

>  and then everything would be under `SPC f t`.

That's ok, too. A bit verbose, but it avoids modifier keys. How about this then?


| Key     | Function                 |
|---------|--------------------------|
| `f t t` | `treemacs`               |
| `f t f` | `treemacs-find-file`     |
| `f t a` | `treemacs-find-tag`      |
| `f t b` | `treemacs-bookmark`      |
| `M-0`   | `treemacs-select-window` |

syl20bnr:
I forgot to say that the toggle is not needed because we can use the DWIN function that show/focus treemacs and press q. In practice it feels like `SPC f t q` but it is first `SPC f t` to focus then `q` in the treemacs window to close it.

I feel confusing/not needed the current distinction between toggle and focus. Whenever a user wants to go to the file tree it should be only one binding. I’m the one who made the distinction between toggle and focus and it was a mistake, I did not really use neotree in the end but I’m slowly starting to use treemacs more.

Also I really want to make treemacs completely ignored by wining, even if it is technically the case already, ``SPC 0` makes it feel it is not. This is also a legacy of the design I made for neotree and it is a bad one.

Alexander-Miller:
> we can use the DWIN function that show/focus treemacs and press q.

That's also my own workflow, but I figured not everyone will agree, so `treemacs` and `select-window` are made separate.

>  ``SPC 0` makes it feel it is not.

Maybe not, but it is fast and simple. The last thing I would want is have to press some 4 letter combo like SPC f t t just to select the filetree. That's fine for one-off actions, but if treemacs becomes part of your workflow you'll see that selecting it is something you do *very often*. So if you want a different keybind it should be something equally small. They keymap is full mostly, but there's still some in places like `C-+` or `M-+`.

fiveNinePlusR:
I was hesitant when `spc f t` just selected the treemacs window but now I really appreciate it. I strongly disagree that spacemacs should put the bindings under a prefix like `spc f t` though. If you really want to prefix those other functions it should go under `spc f T` and leave `spc f t` as the `treemacs` function call(that just opens the file tree or selects it if it's open already). The extra keypress over time would be annoying and goes against the ergonomic goal of spacemacs.(`spc f t t`)  

`spc f t` then `q` is fine to close the window and having one main shortcut to get to the filetree is great. (`spc f t`)

Alexander-Miller:
> like `SPC b b`, `SPC f f` etc

Well, both examples have you hit the same key twice, which certainly helps.  `SPC t t` looks to be free as well, so it's also an option.

Of course there's really no need for me to your

syl20bnr:
Unfortunately `SPC t t` will be used to give us another layer of
toggles.

`SPC f t` is not as nice as double `SPC t t`, I should have compared it
to `SPC f s`.

If we really want a 1-key focus/show shortcut and we can find something
else than `SPC 0`, maybe `SPC [` ? Then if we ever add a useful sidebar
to the right we could bind it to `SPC ]`. Not sure how well it plays
with different keyboard layouts though.

fiveNinePlusR:
`imenu-list-minor-mode` is a somewhat permanent right hand sidebar if you want something for `spc ]` not sure how many other things there are that pop up a right hand sidebar that lasts a while though.

syl20bnr:
Indeed, we could do that.

Issue #10237, opened 2018-01-23
Ido keybinds that is still working and needs to be fixed


longtomjr:
There is still issues regarding people running into ido when using emacs keybinds. The purpose of this post is to just document the ido keybindings that is still running rampant.  

### Ido keybindings in develop:
 - `C-x d` - `ido-dired`
 - `C-x C-w` - `ido-write-file`
 - `C-x k` - `ido-kill-buffer`
 - `C-x 4` - `switch-to-buffer-other-window` and in turn `ido-switch-buffer-other-window`
 - `C-x 4 f` and `C-x 4 C-f` - `ido-find-file-other-window`
 - `C-x C-v` - `ido-find-alternate-file`
 - a number of keybinds for `ido-find-file-other-frame` including `SPC F f`

What the best way might be to fix it is up in the air. Remove the bindings completely (to dissalow use), to show in the minibuffer what the spacemacs binding is for this feature (to encourage changing to spacemacs style), or fix the bindings to call the correct helm command. Please post if you run into any ido keybinds still, and I will add them to the list, I will also be going through ido commands and see what they are bind to since `SPC h d b` does not list the correct keybinds

For more context see #1126

quicknir:
Discussed briefly in gitter; @longtomjr suggested I post my thoughts here:

Generally, spacemacs does not bind, or even change bindings, in any of the native emacs prefixes (C-c, C-x, C-w...). All three suggested solutions involve messing around with those bindings. Maybe we could add a section to the FAQ (https://github.com/syl20bnr/spacemacs/blob/master/doc/FAQ.org), "I'm coming from regular emacs. How do I do X idiomatically?". 

Another possibility would be to have a "spacemacs tutor mode". It could basically operate by adding advice before functions that "shouldn't" really be used in spacemacs. The function would basically error out, telling the user what function to use instead. However, this is more work, I'm not sure who really wants to do it (it's pretty easy to do though).

As far as the solutions go:

- ido-dired: just use `SPC f f` and hit enter on a directory
- ido-write-file `SPC f c` to copy a file
- ido-kill-buffer `SPC b b` and then `M-S-d` to delete whatever you want (can kill multiple in helm via marking)
- switch-to-buffer-other-window and... `SPC b b` and then `C-c o` (in helm)
- ido-find-file-other-window `SPC f f` and `C-c o` as above
- ido-find-alternate-file `, g a`; this is major mode dependent
[...]

PR #10772, opened 2018-05-29
Work around ido mode remapping spacemacs keys

Miciah:
#### Work around ido-mode remapping Spacemacs keys

Add aliases for commands that ido-mode remaps that we want to bind to Spacemacs keys, and use these aliases in the Spacemacs key bindings so that these bindings invoke the original commands rather than the ido commands.

See <https://github.com/syl20bnr/spacemacs/issues/10237#issuecomment-359861570>.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -79,6 +79,7 @@ This file containes the change log for the next major version of Spacemacs.
 - Use Helm or Ivy (if one of these is enabled) for ~SPC a d~ (=dired=),
   ~SPC F f~ (=find-file-other-frame=), ~SPC F b~ (=switch-to-buffer-other-frame=),
   ~SPC F B~ (=display-buffer-other-frame=), and ~SPC F O~ (=dired-other-frame=).
+- New key binding ~SPC f i~ for =insert-file=.
 *** Layer changes
 **** GitHub
 - Add new package =magithub=
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -76,6 +76,9 @@ This file containes the change log for the next major version of Spacemacs.
   - Avoid importing already defined variables (except `PATH`/`MANPATH`).
 - Move all scrolling key bindings from ~SPC n~ to ~SPC N~ (Somelauw)
 - Overhaul the scroll transient state on ~SPC N~ prefix (Somelauw)
+- Use Helm or Ivy (if one of these is enabled) for ~SPC a d~ (=dired=),
+  ~SPC F f~ (=find-file-other-frame=), ~SPC F b~ (=switch-to-buffer-other-frame=),
+  ~SPC F B~ (=display-buffer-other-frame=), and ~SPC F O~ (=dired-other-frame=).
 *** Layer changes
 **** GitHub
 - Add new package =magithub=
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -98,6 +98,7 @@ automatically applied to."
 (defalias 'spacemacs/dired 'dired)
 (defalias 'spacemacs/dired-other-frame 'dired-other-frame)
 (defalias 'spacemacs/switch-to-buffer-other-frame 'switch-to-buffer-other-frame)
+(defalias 'spacemacs/insert-file 'insert-file)
 (defalias 'spacemacs/display-buffer-other-frame 'display-buffer-other-frame)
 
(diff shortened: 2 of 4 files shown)


layers/+completion/helm/funcs.el
```diff
@@ -513,6 +513,23 @@ Removes the automatic guessing of the initial value based on thing at point. "
...
     (helm-find-files-1 input)))
 
+;; ido-mode remaps some commands to ido counterparts.  We want default Emacs key
+;; bindings (those under C-x) to use ido, but we want to use the original
+;; commands in Spacemacs key bindings (those under M-m or SPC) so that they use
+;; `read-file-name-function', `completing-read-function',
+;; `completion-in-region-function', etc. configured by Helm or Ivy etc.  The
+;; following aliases allow us to bind Spacemacs keys to the original commands.
+
+(defalias 'spacemacs/find-file-other-frame 'find-find-other-frame)
```
bmag: `find-find-other-frame` --> `find-file-other-frame`

layers/+completion/helm/funcs.el
```diff
@@ -513,6 +513,23 @@ Removes the automatic guessing of the initial value based on thing at point. "
...
+
+(defalias 'spacemacs/dired-other-frame 'dired-other-frame)
+
+(defalias 'spacemacs/switch-to-buffer-other-frame 'switch-to-buffer-other-frame)
+
+(defalias 'spacemacs/insert-files 'insert-file)
```
bmag: `spacemacs/insert-files` --> `spacemacs/insert-file`

layers/+completion/helm/funcs.el
```diff
@@ -513,6 +513,23 @@ Removes the automatic guessing of the initial value based on thing at point. "
     (set-text-properties 0 (length input) nil input)
     (helm-find-files-1 input)))
 
+;; ido-mode remaps some commands to ido counterparts.  We want default Emacs key
+;; bindings (those under C-x) to use ido, but we want to use the original
+;; commands in Spacemacs key bindings (those under M-m or SPC) so that they use
+;; `read-file-name-function', `completing-read-function',
+;; `completion-in-region-function', etc. configured by Helm or Ivy etc.  The
+;; following aliases allow us to bind Spacemacs keys to the original commands.
+
```
bmag: I think we can get rid of the empty lines after the comment and between every `defalias`, they don't really add value this time and it will be nice to make this code a bit more compact.

layers/+completion/helm/funcs.el
Miciah: Sorry! That's embarrassing—and weird because I thought I tested every one of these bindings...

layers/+completion/helm/funcs.el
Miciah: My reasoning for naming it `spacemacs/insert-files` was that helm allows selecting multiple files.  I did find that helm only really inserts one file, but that looks like a defect in helm (I'll look into fixing it upstream later).

layers/+completion/helm/funcs.el
Miciah: Fixed.

syl20bnr:
Sorry, I don't understand this PR. What issues does it fix exactly ? In
which context ? Even with the referenced issue I don't see the relation
with this PR.

bmag:
> What issues does it fix?

Some commands use ido prompt instead of helm/ivy. For example, <kbd>SPC F f</kbd> (we bind it to `find-file-other-frame`) uses ido and not helm/ivy, because `ido-mode` remaps `find-file-other-frame` to `ido-find-file-other-frame`. Binding to aliases circumvents ido's rebinding, thus fixing the issue.

Another example is <kbd>SPC a d</kbd> (`dired`), which is missing from this PR. @Miciah can you add it please?

Ideally I think we can remove our ido configuration, and not enable `ido-mode` at all, but until that happens we can merge this PR. What do you think @syl20bnr?

Miciah:
Thanks, @bmag! I've added a `spacemacs/dired` for `SPC a d`.  @syl20bnr, sorry for not explaining more clearly initially; I hope bmag's explanation clears things up.

syl20bnr:
@bmag Thank you for the explanation.

We can deprecate ido configuration indeed but this might not be enough,
some packages may require it and I guess we would still need this
work-around in the end.

@Miciah my bad, this explanation is clear to me now.

Miciah:
I pushed an update that adds changelog entries.

syl20bnr:
Thank you ! 👍 

PR #10773, opened 2018-05-29
spacemacs/integrate-evil-search also dupdates evil-ex-search-direction

thanhvg:
This PR is to have spacemacs/integrate-evil-search update evil-ex-search-direction to have consistency in behavior when using both spacemacs `*`,  `#` commands and evil `/` , `?` commands

Currently evil-ex-search-direction isn't updated when using `*` and `#` 
When spacemacs starts up evil-ex-search-direction is nil, do `*` or `#` on a buffer to have some occurrences. Now escape, hit `n` you will get to the next occurrence and it gets stuck there, hitting `n` won't move to the next one. You have to use `?` or `/` first to initialize evil-ex-search-direction, then it will be ok. Also because of this, search directions in spacemacs are evil are not the same when alternating these commands


syl20bnr:
Great Fix, thank you ! 💜 

PR #10782, opened 2018-05-31
add transpose sexps keybinding

gdkrmr:
the transpose keybindings do not contain a binding for `'transpose-sexps`, the original emacs keybinding `C-M-t` is still available, but on most linux desktops this will be bound to launch a terminal.

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/keybindings.el
+++ b/layers/+spacemacs/spacemacs-defaults/keybindings.el
@@ -492,10 +492,10 @@
   "xlS" 'spacemacs/sort-lines-reverse
   "xlu" 'spacemacs/uniquify-lines
   "xtc" 'transpose-chars
+  "xte" 'transpose-sexps
   "xtl" 'transpose-lines
   "xtp" 'transpose-paragraphs
   "xts" 'transpose-sentences
-  "xte" 'transpose-sexps
   "xtw" 'transpose-words
   "xU"  'upcase-region
   "xu"  'downcase-region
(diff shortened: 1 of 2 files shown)


layers/+spacemacs/spacemacs-defaults/keybindings.el
```diff
@@ -495,6 +495,7 @@
   "xtl" 'transpose-lines
   "xtp" 'transpose-paragraphs
   "xts" 'transpose-sentences
+  "xte" 'transpose-sexps
```
sdwolfz: Please put this in alphabetical order based on the keybinding. Also this needs to be added to the documentation.

gdkrmr: should be ok now.

sdwolfz:
Thank you :heart:!

gdkrmr:
great, thanks!

syl20bnr:
Note that transposing sexps is available in lisp state with `SPC k t`

gdkrmr:
I did not know that, in the helm buffer it showed up as `C-M-t`.

syl20bnr:
As it is from an Evil state it cannot be seen by `SPC ?` or `SPC SPC`
unfortunately :-(

To see it you have to explore with which key after typing `SPC k` or
read the documentation about the lisp state.

gdkrmr:
I really like some of the lisp state function, especially making parenthesis swallow arguments is very useful. For some reason I never used the `transpose-sexp` function from there.

PR #10788, opened 2018-06-01
Checks if tern layer is enabled and tern executable exists before using it

sdwolfz:
This fixes errors that appear when navigating between javascript files when using the default javascript layer setup without having the `tern` layer nor the `tern` executable installed.

While the problem is fixed I don't feel 100% satisfied with this solution since these functions are executed every time a javascript file is opened and `(locate-file "tern" exec-path)` will be called every time... but if you have the `tern` layer enabled without the tern executable installed then you probably deserve the extra warnings.

@syl20bnr @ztlevi Please take a look and advise me if you know a better way!


syl20bnr:
Thank you ! :-)

PR #10791, opened 2018-06-02
markdown: fix M-h, M-j, M-k, M-l bindings.

### LINKED ISSUE #10640: "M-l" shortcuts overwrite lowercase behavior (default in emacs) with markdown-demote
May consider a different key binding, the code line:
https://github.com/syl20bnr/spacemacs/blob/bd7ef98e4c35fd87538dd2a81356cc83f5fd02f3/layers/%2Blang/markdown/packages.el#L155

Similarly, 

`M-l` overwrites the default behavior of cutting a paragraph,

https://github.com/syl20bnr/spacemacs/blob/bd7ef98e4c35fd87538dd2a81356cc83f5fd02f3/layers/%2Blang/markdown/packages.el#L154

kccai:
These key bindings are very useful in emacs-editing mode and should not be
overwrited.

This fixes #10640.

### DIFF
--- a/layers/+lang/markdown/README.org
+++ b/layers/+lang/markdown/README.org
@@ -188,12 +188,12 @@ To generate a table of contents type on top of the buffer:
 
 ** Promotion, Demotion
 
-| Key Binding | Description        |
-|-------------+--------------------|
-| ~M-k~       | markdown-move-up   |
-| ~M-j~       | markdown-move-down |
-| ~M-h~       | markdown-promote   |
-| ~M-l~       | markdown-demote    |
+| Evil  | Holy        | Command            |
+|-------+-------------+--------------------|
+| ~M-k~ | ~C-c C-x u~ | markdown-move-up   |
+| ~M-j~ | ~C-c C-x d~ | markdown-move-down |
+| ~M-h~ | ~C-c C-x l~ | markdown-promote   |
+| ~M-l~ | ~C-c C-x r~ | markdown-demote    |
 
 ** Toggles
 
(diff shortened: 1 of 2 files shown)


sdwolfz:
This sounds like a good idea to me, emacs keybindings for these already exist under `C-c C-x` and although I like the convenience these `M-` keybindings bring we should keep the emacs specific functionality intact as much as possible.

I'm going to merge this in a few days unless somebody has any objections.

layers/+lang/markdown/README.org
```diff
@@ -187,6 +187,7 @@ To generate a table of contents type on top of the buffer:
 | ~SPC m P~   | previous link      |
 
 ** Promotion, Demotion
+These keybindings only work in evil-normal and evil-insert modes.
```
sdwolfz: Do you think we should specify here that they are avaiable under `C-c C-x` for holy mode? These are set by the markdown mode so we don't have control over when they change but I feel it's unlikely it will.

mroh69:
What about hybrid mode?
Maybe these keys should only be in evil insert mode, if its not hybrid mode?

sdwolfz:
@mroh69 Hybrid does not use insert mode, you enter in emacs mode instead so it achieves the behaviour you describe by default.

syl20bnr:
Don't use a lambda, prefer a named function.

We need to be able to hot swap the editing style so `M-l` and co must be
reverted to vanilla key bindings when switching from Vim to Emacs style.

Also hybrid style comes with variables to enable hjkl navigation.

I think the best is that I merge this PR and I will adapt it to fit all
the requirements.

syl20bnr:
Thank you ! 👍 

I adapted it so it works in all situation (especially when switching style on the fly) see commit https://github.com/syl20bnr/spacemacs/commit/834c39c2f

PR #10792, opened 2018-06-02
Multiple ways to copy file path

sdwolfz:
Adds a new prefix `SPC f y` which replaces the current `spacemacs/show-and-copy-buffer-filename` keybinding.

The new keybindings will allow you to copy an absolute / relative file path, with added line and column numbers.

I still have some work and documentation to do, so it's a WIP at the moment. Also it would be nice if the spaceline PR was merged first so that toggling column indexing would show up correctly: https://github.com/TheBB/spaceline/pull/194.

/cc @agzam since we discusses a bit about these at work :wink:.

### DIFF
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -1227,6 +1227,7 @@ and ~T~):
 | ~SPC t L~   | toggle visual lines                                               |
 | ~SPC t n~   | toggle line numbers                                               |
 | ~SPC t v~   | toggle smooth scrolling                                           |
+| ~SPC t z~   | toggle 0/1 based column indexing                                  |
 
 | Key Binding | Description                                                      |
 |-------------+------------------------------------------------------------------|
@@ -2366,7 +2367,10 @@ Files manipulation commands (start with ~f~):
 | ~SPC f v d~ | add a directory variable                                                                               |
 | ~SPC f v f~ | add a local variable to the current file                                                               |
 | ~SPC f v p~ | add a local variable to the first line of the current file                                             |
-| ~SPC f y~   | show and copy current file absolute path in the minibuffer                                             |
+| ~SPC f y c~ | show and copy current file absolute path with line and column number in the minibuffer                 |
+| ~SPC f y d~ | show and copy current directory absolute path in the minibuffer                                        |
+| ~SPC f y l~ | show and copy current file absolute path with line number in the minibuffer                            |
+| ~SPC f y y~ | show and copy current file absolute path in the minibuffer                                             |
 
 **** Frame manipulation key bindings
 Frame manipulation commands (start with ~F~):
@@ -3255,39 +3259,43 @@ To search in a project see [[#searching-in-a-project][project searching]].
 
 =projectile= commands start with p:
 
-| Key Binding | Description                                             |
-|-------------+---------------------------------------------------------|
-| ~SPC p '​~   | open a shell in project's root (with the =shell= layer) |
-| ~SPC p !~   | run shell command in project's root                     |
-| ~SPC p &~   | run async shell command in project's root               |
-| ~SPC p %~   | replace a regexp                                        |
-| ~SPC p a~   | toggle between implementation and test                  |
-| ~SPC p b~   | switch to project buffer                                |
-| ~SPC p c~   | compile project using =projectile=                      |
-| ~SPC p d~   | find directory                                          |
-| ~SPC p D~   | open project root in =dired=                            |
-| ~SPC p e~   | edit dir-locals.el                                      |
-| ~SPC p f~   | find file                                               |
-| ~SPC p F~   | find file based on path around point                    |
-| ~SPC p g~   | find tags                                               |
-| ~SPC p G~   | regenerate the project's =etags= / =gtags=              |
-| ~SPC p h~   | find file                                               |
-| ~SPC p I~   | invalidate the projectile cache                         |
-| ~SPC p k~   | kill all project buffers                                |
-| ~SPC p o~   | run =multi-occur=                                       |
-| ~SPC p p~   | switch project                                          |
-| ~SPC p r~   | open a recent file                                      |
-| ~SPC p R~   | replace a string                                        |
-| ~SPC p t~   | open =NeoTree= in =projectile= root                     |
-| ~SPC p T~   | test project                                            |
-| ~SPC p v~   | open project root in =vc-dir= or =magit=                |
-| ~SPC /~     | search in project with the best search tool available   |
-| ~SPC s p~   | see [[#searching-in-a-project][searching in a project]]                              |
-| ~SPC s a p~ | run =ag=                                                |
-| ~SPC s g p~ | run =grep=                                              |
-| ~SPC s k p~ | run =ack=                                               |
-| ~SPC s t p~ | run =pt=                                                |
-| ~SPC s r p~ | run =rg=                                                |
+| Key Binding | Description                                                                              |
+|-------------+------------------------------------------------------------------------------------------|
+| ~SPC p '​~   | open a shell in project's root (with the =shell= layer)                                  |
+| ~SPC p !~   | run shell command in project's root                                                      |
+| ~SPC p &~   | run async shell command in project's root                                                |
+| ~SPC p %~   | replace a regexp                                                                         |
+| ~SPC p a~   | toggle between implementation and test                                                   |
+| ~SPC p b~   | switch to project buffer                                                                 |
+| ~SPC p c~   | compile project using =projectile=                                                       |
+| ~SPC p d~   | find directory                                                                           |
+| ~SPC p D~   | open project root in =dired=                                                             |
+| ~SPC p e~   | edit dir-locals.el                                                                       |
+| ~SPC p f~   | find file                                                                                |
+| ~SPC p F~   | find file based on path around point                                                     |
+| ~SPC p g~   | find tags                                                                                |
+| ~SPC p G~   | regenerate the project's =etags= / =gtags=                                               |
+| ~SPC p h~   | find file                                                                                |
+| ~SPC p I~   | invalidate the projectile cache                                                          |
+| ~SPC p k~   | kill all project buffers                                                                 |
+| ~SPC p o~   | run =multi-occur=                                                                        |
+| ~SPC p p~   | switch project                                                                           |
+| ~SPC p r~   | open a recent file                                                                       |
+| ~SPC p R~   | replace a string                                                                         |
+| ~SPC p t~   | open =NeoTree= in =projectile= root                                                      |
+| ~SPC p T~   | test project                                                                             |
+| ~SPC p v~   | open project root in =vc-dir= or =magit=                                                 |
+| ~SPC /~     | search in project with the best search tool available                                    |
+| ~SPC s p~   | see [[#searching-in-a-project][searching in a project]]                                                               |
+| ~SPC s a p~ | run =ag=                                                                                 |
+| ~SPC s g p~ | run =grep=                                                                               |
+| ~SPC s k p~ | run =ack=                                                                                |
+| ~SPC s t p~ | run =pt=                                                                                 |
+| ~SPC s r p~ | run =rg=                                                                                 |
+| ~SPC f y C~ | show and copy current file path relaitve to the project root with line and column number |
+| ~SPC f y D~ | show and copy current directory path relaitve to the project root                        |
+| ~SPC f y L~ | show and copy current file path relaitve to the project root with line number            |
+| ~SPC f y Y~ | show and copy current file path relaitve to the project root                             |
 
 *Note for Windows Users*: To enable fast indexing the GNU ~find~ or
 Cygwin ~find~ must be in your ~PATH~.
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -479,15 +479,96 @@ If the universal prefix argument is used then will the windows too."
             (if dedicated "no longer " "")
             (buffer-name))))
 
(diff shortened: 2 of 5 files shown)


henriquenj:
I will give my feedback since I use this keybind a lot.

I think it would be much less disruptive to my muscle memory if the original functionally be retained on the `SPC f y y` keybind. From what I notice it now lies on `SPC f y p`. But that's of course just me. Does anybody else agree?

I also noticed that `SPC f y p` does not work on magit buffers and on dired buffers, it fails with `WARNING: Current buffer is not attached to a file!`, while the previous implementation works. Is there a way to retain this behaviour?

All that said, I _did_ miss all these variations of the `SPC f y` command, so thanks for the work!

sdwolfz:
I chose `fyp` as a mnemonic for `file yank path`, but `fyy` also makes a lot of sense since it's easyer to type and is in the same class as the vim yank keybinding `yy`. Not sure at the moment, maybe I'll keep both :confused:.

About the magit and dired buffers, I did not know it works with them as well... I'll have to take a look to see what's happening, thanks for bringing it up!

layers/+spacemacs/spacemacs-project/packages.el
```diff
@@ -54,6 +54,11 @@
             projectile-known-projects-file (concat spacemacs-cache-directory
                                                    "projectile-bookmarks.eld"))
       (spacemacs/set-leader-keys
+        ;; File path
+        "fyL" 'spacemacs/projectile-show-and-copy-file-relative-path-with-line-number
+        "fyc" 'spacemacs/show-and-copy-file-absolute-path-with-line-and-column-number
```
Miciah: Seems like `layers/+spacemacs/spacemacs-defaults/keybindings.el` should have this binding, and `layers/+spacemacs/spacemacs-project/packages.el` should have the `spacemacs/projectile-show-and-copy-file-relative-path` binding.

layers/+spacemacs/spacemacs-defaults/keybindings.el
```diff
@@ -359,6 +362,11 @@
   :mode font-lock-mode
   :documentation "Toggle syntax highlighting."
   :evil-leader "ths")
+(spacemacs|add-toggle spacemacs/toggle-column-indexing
+  :on (spacemacs/toggle-column-indexing)
+  :off (spacemacs/toggle-column-indexing)
+  :evil-leader "tz")
```
Miciah: I think you want something like the following:
```emacs-lisp
(spacemacs|add-toggle column-indexing
  :status column-number-indicator-zero-based
  :on (setq column-number-indicator-zero-based t)
  :off (setq column-number-indicator-zero-based nil))
```
...Wait, what does this have to do with the rest of the PR?

layers/+spacemacs/spacemacs-defaults/funcs.el
```diff
@@ -477,15 +477,55 @@ If the universal prefix argument is used then will the windows too."
...
+

+;; Copy file path
+(defun spacemacs/show-and-copy-file-absolute-path ()
+  "Show and yank the full path file name in the minibuffer."
   (interactive)
-  ;; list-buffers-directory is the variable set in dired buffers
-  (let ((file-name (or (buffer-file-name) list-buffers-directory)))
+  (let ((file-name (buffer-file-name)))
     (if file-name
```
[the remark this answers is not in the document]
syl20bnr: We could use `if-let`, Emacs 25 is the minimum Emacs version in develop branch, we try to not rely on `dash` though.

syl20bnr:
I like it, agree to put/duplicate `SPC f y y` for the legacy function.

sdwolfz:
@Miciah Thanks for the input, I'll give you some answeres, although I would preffer that for PRs marked as WIP people would not review the implementation details and focus on the general idea / overall funcionality.  There are a lot of obvious mistakes I made while prototyping this that I am aware of, but not yet gotten the time to fix.

layers/+spacemacs/spacemacs-defaults/keybindings.el
sdwolfz: Emacs 26 introduced this functionality, and I want my functions to respect the setting. So in case you figured out you did not copy the path with the appropriate column number you can use this toggle to fix it up.

layers/+spacemacs/spacemacs-defaults/funcs.el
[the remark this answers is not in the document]
sdwolfz: I just read that `if-let` is obsolete in emacs 26, and they added `if-let*` to replace it... but it's not available in older versions. So I guess `-if-let` would be the cross version approach.

sdwolfz:
@syl20bnr I've chosen to keep only `SPC f y y` in the end. Although `SPC f y p` sounds better, it's pretty uncomfortable to type (based on my experience for the last weeks).

henriquenj:
Thanks for the work @sdwolfz . I did some test locally. Everything that I use is now working under `SPC f y y`. The new `SPC f y Y` is really nice and works wonderful with dired as well.

The only thing I found odd is that calling `SPC f y C` yelds an error (`file-relative-name: Wrong type argument: arrayp, nil`) on dired and magit, maybe it should print a message saying that the buffer is not visiting a file?

IMO we should not make the command different between dired/magit and an ordinary file buffer. Maybe it's just a bad habit, but I like to think of `SPC f y` "as show me _where_ is the buffer that I'm visiting on the filesystem". If we split the functions I'd have to think the _kind_ of buffer I am before issuing the command. It seems easier to be concerned with the type of path I want (project? absolute? etc) and not bother with the kind of buffer I'm visiting. If it's a directory, we get a path directory, if it's magit, a project.

layers/+spacemacs/spacemacs-defaults/funcs.el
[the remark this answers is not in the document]
Miciah: > I just read that if-let is obsolete in emacs 26, and they added if-let* to replace it

Where did you read that? It is true that at one point, `if-let` was marked obsolete ([commit](http://git.savannah.gnu.org/cgit/emacs.git/commit/?id=4612b2a2b37026bef5a9b8e92878a15dabb9b261)), but doing so caused the sort of compatibility problem that you are talking about, and so the change was reverted ([commit](http://git.savannah.gnu.org/cgit/emacs.git/commit/?id=af4697faa1f5b643f63a9ea61aa205a4c1432e23)).  These changes both happened prior to the Emacs 26.1 release, and [the Emacs 26.1 NEWS file](http://git.savannah.gnu.org/cgit/emacs.git/tree/etc/NEWS?h=emacs-26.1) does not say that `if-let` is obsolete.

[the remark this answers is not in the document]
sdwolfz: I read about it here: https://github.com/clojure-emacs/cider/issues/2130.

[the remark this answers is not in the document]
Miciah: Ah, that discussion took place during the time in between the two commits I cited.  As far as I know, `if-let` was never marked obsolete in a released version of Emacs.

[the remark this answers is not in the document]
sdwolfz: That's good news. I'll switch to it.

sdwolfz:
I've added `SPC f y d` and `SPC f y D` to display the directory path, which work for dired and magit as well. I don't like the idea of displaying the directory path in `SPC f y y`. It should only handle files. Otherwise the behaviour gets compilcated and unpredictable.

henriquenj:
@sdwolfz I understand. Would you agree to a `SPC b f` command that works as I described? Like "show me where in the filesystem this buffer is".

I'm not asking to do it on this PR, just brainstorming.

sdwolfz:
@henriquenj sure, please open a feature request issue describing the behavior you want. Add in as much detail as possible, and reference this PR in it.

sdwolfz:
I got the functionality to a point I think it's ready for review. Thanks everybody for the early feedback :heart:! Let me know if there is anything not right. Otherwise I will merge in the weekend.

doc/DOCUMENTATION.org
```diff
@@ -3251,39 +3255,43 @@ To search in a project see [[#searching-in-a-project][project searching]].
...
+| ~SPC s p~   | see [[#searching-in-a-project][searching in a project]]                                                             |
+| ~SPC s a p~ | run =ag=                                                                               |
+| ~SPC s g p~ | run =grep=                                                                             |
+| ~SPC s k p~ | run =ack=                                                                              |
+| ~SPC s t p~ | run =pt=                                                                               |
+| ~SPC s r p~ | run =rg=                                                                               |
+| ~SPC f y C~ | show and copy current file relative path with line and column number in the minibuffer |
+| ~SPC f y D~ | show and copy current directory relative path in the minibuffer                        |
+| ~SPC f y L~ | show and copy current file relative path with line number in the minibuffer            |
+| ~SPC f y Y~ | show and copy current file relative path in the minibuffer                             |
```
Miciah: Initially, I thought these bindings should be documented with the absolute-path counterparts.  Maybe clarify "relative to the project root"?

layers/+spacemacs/spacemacs-defaults/funcs.el
```diff
@@ -479,15 +479,61 @@ If the universal prefix argument is used then will the windows too."
...
+(defun spacemacs/copy-directory-path ()
+  "Copy and show the directory path of the current buffer.
+If the buffer is not visiting a file, use the `list-buffers-directory'
+variable as a fallback to display the directory, useful in buffers like the
+ones created by `magit' and `dired'."
+  (interactive)
+  (if-let (directory-name (or
+                            (if-let (file-name (buffer-file-name))
+                              (file-name-directory file-name))
+                            list-buffers-directory))
```
Miciah: The `or` seems unnecessary.  How about:

```emacs-lisp
(if-let (directory-name (if-let (file-name (buffer-file-name))
￼                            (file-name-directory file-name)
                          list-buffers-directory))
```

layers/+spacemacs/spacemacs-defaults/funcs.el
```diff
@@ -479,15 +479,61 @@ If the universal prefix argument is used then will the windows too."
...
+    (message "%s" (kill-new (file-truename directory-name)))
```
Miciah: Usually the then form is indented two spaces more than the else form.  Might be worth doing an auto-indent on this file.

layers/+spacemacs/spacemacs-defaults/funcs.el
```diff
@@ -479,15 +479,61 @@ If the universal prefix argument is used then will the windows too."
...
+(defun spacemacs/copy-file-path-with-line-column ()
+  "Copy and show the file path of the current buffer, including line and column number.
+
+This function respects the value of the `column-number-indicator-zero-based'
+variable."
+  (interactive)
+  (if-let (file-name (buffer-file-name))
+    (let* ((current-file-path (file-truename file-name))
+            (current-line-number (number-to-string (line-number-at-pos)))
+            (current-column-numnber
```
Miciah: Typo in `current-column-numnber`—but you were consistent about it, so it does not make the code incorrect.

layers/+spacemacs/spacemacs-defaults/funcs.el
```diff
@@ -479,15 +479,61 @@ If the universal prefix argument is used then will the windows too."
...
+              (number-to-string (if (and
+                                      (boundp column-number-indicator-zero-based)
+                                      (not column-number-indicator-zero-based))
```
Miciah: Could you `defvar` it to avoid the need for `boundp`?

Miciah:
You could factor out some helper functions and user `user-error` to make the code shorter and less repetitive, something like the following (this is for relative paths; should be similar for absolute paths):
```emacs-lisp
(defun spacemacs//relative-file-path ()
  (if-let (file-name (buffer-file-name))
      (file-relative-name (file-truename (buffer-file-name))
                          (projectile-project-root))
    (user-error "Current buffer is not visiting a file")))

(defun spacemacs//relative-file-path-with-line ()
  (concat (spacemacs//relative-file-path)
          ":"
          (number-to-string (line-number-at-pos))))

(defun spacemacs//relative-file-path-with-line-column ()
  (concat (spacemacs//relative-file-path-with-line)
          ":"
          (number-to-string (if column-number-indicator-zero-based
                                (current-column)
                              (1+ (current-column))))))

(defun spacemacs/copy-relative-file-path ()
  (message "%s" (kill-new (spacemacs//relative-file-path))))

(defun spacemacs/copy-relative-file-path-with-line ()
  (message "%s" (kill-new (spacemacs//relative-file-path-with-line))))

(defun spacemacs/copy-relative-file-path-with-line-column ()
  (message "%s" (kill-new (spacemacs//relative-file-path-with-line-column))))
```
What do you think?

layers/+spacemacs/spacemacs-defaults/funcs.el
sdwolfz: I'm using `evil-lisp-state` from time to time and it reindents everything like this. Maybe there's a setting that controls this behavior, I'll try to see if I can get it to indent properly by default.

layers/+spacemacs/spacemacs-defaults/funcs.el
sdwolfz: I'm still wrapping my head around using `if-let` properly. I wish I had it in other languages as well, seems like it compacts a lot of branching code.

sdwolfz:
@Miciah I've addressed your review in the lastest commit.

The functions now use `when-let` to check if a path exists, because I don't like using errors for this type of logic. Also I left the `column-number-indicator-zero-based` unbound since it feels wrong to me to define this if the feature does not exist.

sdwolfz:
@JAremko this should be a magit/magithub functionality, it's less tied to the file itself and more related to the vcs and central service used.

sdwolfz:
Merged into develop. Thanks everyone for the feedback :heart:!

syl20bnr:
Nice ! 👍

PR #10803, opened 2018-06-04
Fix typo in spacemacs/plist-get docstring

benreyn:


benreyn:
@syl20bnr FWIW, its not super clear to me from the doc strings how this function differs from the old `spacemacs/mplist-get`. Is the goal to migrate all usage to the new function and deprecate the old? Is there any functional difference between the two?

Miciah:
Looks like `spacemacs/mplist-get` returns all values after the keyword and `spacemacs/plist-get` returns only the first value after the keyword.  Presumably the "m" stands for "multiple", and `spacemacs/mplist-get` says "Get the values" (plural) whereas `spacemacs/plist-get` says "Get the value" (singular), but the distinction could be more explicitly stated.

benreyn:
Huh.... I was under the impression the `mplist` stood for "modified plist" 🤷‍♂️

Miciah:
Sorry, you're right of course.

syl20bnr:
You are all right. Naming could be improved indeed:

```
spacemacs/mplist-get-values
spacemacs/mplist-get-single-value
```

Also `spacemacs/mplist-get-single-value` could be written in term of
`spacemacs/mplist-get-values` by just returning the car value of it.

PR welcome :-)

In the meantime I'll merge this one.

syl20bnr:
Thank you ! 👍 

PR #10814, opened 2018-06-05
auto-completion: use counsel-company to show completion candidates

ellakk:
Trying to get the ivy layer up to par with helm.

### DIFF
--- a/layers/+completion/auto-completion/packages.el
+++ b/layers/+completion/auto-completion/packages.el
@@ -156,13 +156,14 @@
       (spacemacs/set-leader-keys "is" 'spacemacs/helm-yas)
       (setq helm-c-yas-space-match-any-greedy t))))
 
-(defun auto-completion/init-helm-company ()
-  (use-package helm-company
-    :if (configuration-layer/package-used-p 'company)
-    :defer t
-    :init
-    (with-eval-after-load 'company
+(defun auto-completion/pre-init-helm-company ()
+  (spacemacs|use-package-add-hook company
+    :post-config
+    (use-package helm-company
+      :defer t
+      :init
       (define-key company-active-map (kbd "C-/") 'helm-company))))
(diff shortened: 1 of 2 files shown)


layers/+completion/auto-completion/packages.el
```diff
@@ -128,6 +129,11 @@
                                             "company-statistics-cache.el"))
       (add-hook 'company-mode-hook 'company-statistics-mode))))
 
+(defun auto-completion/post-init-counsel ()
```
syl20bnr: Use a `pre-init` function instead to setup use-package hooks, using a `post-init` does not guarantee the hook will be setup before the `:config` part of the targeted  package(in this case company) is executed (for instance the package is not lazy loaded).

Said differently, all `pre-init` functions are executed before the `init` functions so we are 100% guaranteed that the hooks are installed before the `init` function is executed.

syl20bnr: You could update `auto-completion/init-helm-company`with this better idiom as well :-)

syl20bnr: Oh to update `auto-completion/init-helm-company` you'll be able to create a `pre-init` function as well as an empty `init` function. `init` function is mandatory, if there is no `init` function for a package then it is considered unused. Also the layer defining the `init` function of a package is called the owner of this package.

This rule with `init` function may be simplified at some point :-)

syl20bnr: You can see an example of this here: https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Blang/elixir/packages.el#L172

ellakk: Ok, i pushed your requested changes.

ellakk:
Updated, think i got it right.

syl20bnr:
Thank you ! 👍 

PR #10821, opened 2018-06-06
Counsel jump in buffer and fix missing imenu declaration

ellakk:
Hello, this works the same as spacemacs/helm-jump-in-buffer but for counsel/ivy.

I also noticed that imenu was missing from the package declaration in the ivy layer so i fixed that. The ivy/post-init-imenu code was never running and the only reason "ji" called counsel-imenu was because counsel-mode remaps imenu to counsel-imenu.


bmag:

PR #10829, opened 2018-06-07
Fix several documentation errata

dmolina:
My small contribution to the documentation:

- global documentation in README.org: 'packags' => 'packages'.
- pdf layer: updated layer name in the documentation.

The first one has been for a long time, but the second one is very recent
(due to rename of the layer).

Thanks for your work!

Updated: Initially it gave me an error, but using several smaller commit there was no error.


syl20bnr:
Thank you !
Cherry-picked and squashed into develop branch, you can safely delete your branch.

Issue #10825, opened 2018-06-07
layer go: add support for golangci-lint


utrack:
Go layer has support for gometalinter, but it has no switch to use golangci-lint (https://github.com/golangci/golangci-lint) instead.
Flycheck already has package `flycheck-golangci-lint`.

panchoh:
This issue was fixed a long time ago in `develop` (see referenced PR above), so I think it should be closed.

duianto:
@panchoh Your right, thanks.

dbriemann:
Sorry to hijack this but I think my question is related to this feature. I configured my spacemacs to use the `golangci-lint`. Now I have the problem, that compiler errors are not correctly shown. If I have no compile errors, the linter works correctly and shows me the linter errors/warnings on the correct lines.

Do I need to set the `go-build` linter before the `golangci-lint` linter? Is that possible? Or is there another recommended way to highlight build errors correctly before using the linter? If I set the linter to `lsp-ui`, build errors are also shown correctly, but is worse than `golangci-lint`.

p.s.
I am on the current develop branch.

seriousben:
I see the same behavior as @dbriemann

seriousben:
Using an official way of installing `golangci-lint` (https://github.com/golangci/golangci-lint#install) seems to work for me now.

```
$ golangci-lint version
golangci-lint has version 1.20.1 built from 849044b on 2019-10-15T19:11:27Z
```

PR #10832, opened 2018-06-07
Add golangci-lint support for go layer

ljupchokotev:
This PR adds support for [golangci-lint](https://github.com/golangci/golangci-lint) with [flycheck-golangci-lint](https://github.com/weijiangan/flycheck-golangci-lint).

Fixes this issue (request): #10825

### DIFF
--- a/layers/+lang/go/README.org
+++ b/layers/+lang/go/README.org
@@ -53,6 +53,8 @@ to get all the goodies of this layer:
   go get -u github.com/josharian/impl
 #+END_SRC
 
+If you with to use a linters aggregator tool, you can enable =gometalinter= or =golangci-lint=.
+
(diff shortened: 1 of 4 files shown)


syl20bnr:
I need more time to merge this one, thanks for your patience.

Miciah:
Instead of having `go-use-gometalinter` and `go-use-golangci-lint` variables, would it make more sense to have a single variable, e.g., `go-flycheck-checker`, with possible values `default`,  `golangci-lint`, or `gometalinter`?

ljupchokotev:
@Miciah 
You might want to use both golangci-lang and gometalinter, for example using different linters from both of them for some reason.

sdwolfz:
Thank you :heart:!

I'll go ahead and merge this as is, and if we discover that having them separate is not desirable then we can improve the layer with @Miciah's suggestion.

PR #10835, opened 2018-06-07
Fix missing package declarations in different layers

ellakk:
Yesterday i found that imenu was missing the the package declarations in ivy by mistake when i was trying to improve jump-in-buffer with counsel so i submitted a pull request. Then i realized that there might be more errors like this so i wrote a Python script that went thru all the layers looking for missing package declarations and it found these 4.
I don't use these layers myself but i think it should improve the editing experience for those that does.
Also, as a byproduct the script also found some wonky-ness in the restructuredtext layer that i will submit a pull request for later.

### DIFF
--- a/layers/+lang/asm/packages.el
+++ b/layers/+lang/asm/packages.el
@@ -13,6 +13,7 @@
       '(
         ;; package names go here
         asm-mode
+        company
         electric-indent-mode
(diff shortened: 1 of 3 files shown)


syl20bnr:
Very nice finding ! 👍 

Issue #10779, opened 2018-05-30
ansi-term scrolling


jvillasante:
After updating to Emacs 26.1 I can't scroll on ansi-term.
In normal mode `j`, `k` don't seem to work anymore.

Any thoughts on this?

jvillasante:
I don't know if #10844 is merged or not, but on latest develop this issue has not been fixed yet.

akirakyle:
[In Emacs 26.1 the behavior of `term-char-mode` was changed](https://emacs.stackexchange.com/questions/44270/previous-line-function-in-term-char-mode), specifically `term-char-mode-point-at-process-mark` was introduced which keeps the point always on the current shell line. As a result any movement commands would be undone, even the built in `previous-line` which is what evil calls. So to go back to the previous behavior, just set `(setq term-char-mode-point-at-process-mark nil)`

The solution in #10844 breaks some things like pasting and doesn't work with my workflow of toggling between emacs state and normal mode.

duianto:
The following PR is now on the `develop` branch:
shell: allow evil normal-mode movement in terminal buffer #12591

PR #10844, opened 2018-06-09
shell: add support for vim/hybrid styles in term/ansi-term buffers.

ellakk:
I was trying out the vim editing style today and noticed that the cursor is stuck in normal/visual mode in term buffers, you can't move around and you cant edit/copy anything. This solves it. 
This commit also ensured the cursor goes to the prompt if the cursor enters insert/hybrid mode from a different line than the prompt line. 
Fixes:  #10779.

### DIFF
--- a/layers/+tools/shell/funcs.el
+++ b/layers/+tools/shell/funcs.el
@@ -182,3 +182,29 @@ is achieved by adding the relevant text properties."
   "Send tab in term mode."
   (interactive)
   (term-send-raw-string "\t"))
+
+(defun spacemacs//term-enable-line-mode ()
+  "Enable `term-line-mode' when in `term-mode' buffer."
+  (when (eq major-mode 'term-mode)
+    (term-line-mode)))
+
+(defun spacemacs//term-enable-char-mode-maby-goto-prompt ()
+  "Enable `term-char-mode' when in a `term-mode' buffer.
+
+Will also put cursor on prompt position if called when cursor is not on the same
+line as prompt."
+  (when (eq major-mode 'term-mode)
+    (when (get-buffer-process (current-buffer))
+      (term-char-mode)
+      (if (not (eq (line-number-at-pos (point))
+                   (line-number-at-pos (marker-position (term-process-mark)))))
+          (goto-char (term-process-mark))))))
+
+(defun spacemacs//init-term-vim-hybrid ()
+  "Enable term support for vim and hybrid editing-styles."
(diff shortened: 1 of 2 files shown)


giuscri:
Inside a project, a shell is no longer spawned with `SPC '`. Instead, current window is split and the same buffer is shown twice.

This is what's printed in `*Messages*`:
```
term-char-mode: Wrong type argument: processp, nil
Error in post-command-hook (evil-mode-check-buffers): (wrong-type-argument processp nil)
```

ellakk:
Well that sucks, I'm gonna have a look at it when i get home or tomorrow. Did this happen with the same setup you posted in #10956. I've been using it on emacs 26.1 and seems to work just fine but ill investigate.

OkIdeal0:
I had the same issue as @giuscri. The problem seems to be that `(term-char-mode)` is called before the process is properly set up. Changing `term-enable-char-mode-maby-goto-prompt` to
```
(defun spacemacs//term-enable-char-mode-maby-goto-prompt ()
     (when (eq major-mode 'term-mode)
       (when (get-buffer-process (current-buffer))
         (term-char-mode)
         (if (not (eq (line-number-at-pos (point))
                      (line-number-at-pos (marker-position (term-process-mark)))))
             (goto-char (term-process-mark))))))
```
works for me. This is the approach taken by [evil-collections](https://github.com/emacs-evil/evil-collection/blob/master/evil-collection-term.el).

ellakk:
Thanks @OkIdeal0, updated pr.

giuscri:
It seems to work.

Yet, when I copy from a buffer using `y` and I use paste to a terminal using `p` the following happens
output
Basically pasting deletes the start of the line when in Normal mode. It reappears when you're back into Insert mode.

layers/+tools/shell/funcs.el
```diff
@@ -182,3 +182,29 @@ is achieved by adding the relevant text properties."
...
+
+(defun spacemacs//term-enable-char-mode-maby-goto-prompt ()
+  "Enable `term-char-mode' when in a `term-mode' buffer.
+
+Will also put cursor on prompt position if called when cursor is not on the same
+line as prompt."
+  (when (eq major-mode 'term-mode)
+    (when (get-buffer-process (current-buffer))
+      (term-char-mode)
+      (if (not (eq (line-number-at-pos (point))
```
brandonwillard: Why not use `term-goto-process-mark-maybe`?

PR #10855, opened 2018-06-11
Add LSP backend to go layer

ljupchokotev:
This PR adds LSP backend to the go layer.

### DIFF
--- a/layers/+lang/go/packages.el
+++ b/layers/+lang/go/packages.el
@@ -11,6 +11,7 @@
 
 (setq go-packages
       '(
+        company
         (company-go :requires company)
         counsel-gtags
         flycheck
@@ -28,6 +29,10 @@
         go-rename
         go-tag
         godoctor
+        (lsp-go
+         :requires lsp-mode
+         :location (recipe :fetcher github
+                           :repo "emacs-lsp/lsp-go"))
         popwin
         ))
 
@@ -38,6 +43,13 @@
             :backends company-go
             :modes go-mode
             :variables company-go-show-annotation t)))
+
+(defun go/init-lsp-go ()
+  (use-package lsp-go
+    :commands lsp-go-enable))
+
+(defun go/post-init-company ()
+  (add-hook 'go-mode-local-vars-hook #'spacemacs//go-setup-company))
 
 (defun go/post-init-counsel-gtags ()
   (spacemacs/counsel-gtags-define-keys-for-mode 'go-mode))
@@ -108,7 +120,12 @@
     (progn
       ;; get go packages much faster
       (setq go-packages-function 'spacemacs/go-packages-gopkgs)
-      (add-hook 'go-mode-hook 'spacemacs//go-set-tab-width))
+      (add-hook 'go-mode-hook 'spacemacs//go-set-tab-width)
+      (add-hook 'go-mode-local-vars-hook
+                #'spacemacs//go-setup-backend)
+      (dolist (value '(lsp default))
+        (add-to-list 'safe-local-variable-values
+                     (cons 'go-backend value))))
     :config
     (progn
       (when go-format-before-save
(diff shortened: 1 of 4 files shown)


robbyoconnor:
Rebase.

ljupchokotev:
Done.

sdwolfz:
Tried this out locally and it seems to work properly. Only 2 minor (non-blocking) nitpicks I have:

1. If you don't do `go get -u github.com/sourcegraph/go-langserver` then `lsp-go-enable` fails with a cryptic `Debugger entered--Lisp error: (wrong-type-argument stringp nil)`, but I guess this should be fixed upstream.
2. The name of the "normal" backend is `default` which does not sound right to me since part of the lsp related functionality is provided by `go get` packages and their functionality is implemented in `go-mode`. How about specifying the backend name as `go-mode`?

bittopaz:
Tribes out this locally, but I can not get lsp-flycheck to work…

ljupchokotev:
@sdwolfz I'll change `default` to `go-code`.

@bittopaz It seems like there is some problem with flycheck, it doesn't seem to work for me in the `develop` branch either. Can you confirm this? I will investigate more and fix it so lsp-flycheck works when lsp is enabled.

bittopaz:
@ljupchokotev Thanks for the reply. 

I managed to make it partially work with my private fork of [lsp-go](https://github.com/bittopaz/lsp-go), which turn on the `-diagnostics`. But now only the errors would show by flycheck.  However, for those warnings, such as `Exported functions should have comment` are NOT shown anymore. I think that might be related with some linters.

layers/+lang/go/packages.el
```diff
@@ -39,6 +44,13 @@
             :modes go-mode
             :variables company-go-show-annotation t)))
 
+(defun go/init-lsp-go ()
+  (use-package lsp-go
+    :commands lsp-go-enable))
+
```
bittopaz: Do you think we need a line `    :config (spacemacs//setup-lsp-jump-handler 'go-mode)))` here. 

I not familiar with elisp, but I got line from python layer.

ljupchokotev: I am not sure about that, I did try that locally but it doesn't seem to do anything different. Someone with more experience should be able to comment on this.

ljupchokotev:
@bittopaz If you just do `SPC SPC flycheck-select-checker` and pick `gometalinter`, all the errors should show up from gometalinter as they should with lsp-ui-flycheck. I will try to fix this in the PR.

bittopaz:
@ljupchokotev I could make it work with your suggestion. But the checker is so so slow...

And that is the reason why I turn to LSP, just to make all these checker faster, just as fast as VSCode-go is very enough for me.

ljupchokotev:
@bittopaz You should only enable the fast checkers from gometalinter (check the go layer README). I have another PR (#10832) for golangci-lint support which claims to be faster then gometalinter, you can try that as well.

bittopaz:
OK, I'll try it. Thanks @ljupchokotev.

sdwolfz:
@bittopaz Let me know if all your concerns were addressed, after which I will press the merge button 😃!

ljupchokotev:
@sdwolfz There is still the problem with flycheck. If you use the lsp backend, gometalinter doesn't get activated, so there is no linting available. I looked at it briefly but couldn't figure out why that was happening.

ljupchokotev:
@sdwolfz Seems to me that the problem with flycheck is present in the other layers with LSP backends. Can we merge this and fix the flycheck problems in another PR?

sdwolfz:
Yes, I was thinking the same thing. If these changes do not affect the functionality when using `go-mode` backend then it can be fixed in a future PR.

I will give it another go (pun intended) this evening and merge it.

sdwolfz:
Thank you :heart:!

PR #10859, opened 2018-06-12
Fix breakage of `M-x` caused by Helm lazy load changes.

bleggett:
Fix issue where `M-x` no longer brings up `Helm M-x`, caused by helm lazy load changes: a417e3ca856a2a6c885521d35ae1ce72513de978

These 3 lines were removed in that commit, but do not seem to have anything to do with lazy-loading.

AFAIK putting this back should not break lazy-loading, and fixes the issue for me.


syl20bnr:
Thank you ! 👍 

Sorry I thought the deferred key bindings was the same thing but it is not: https://github.com/syl20bnr/spacemacs/pull/10859/files#diff-fa2afbb00b304a76ad8b127262c68de4R124

Issue #10290, opened 2018-02-01
ivy-occur is no longer editable in evil


aaronjensen:
#### Description :octocat:
ivy-occur is no longer editable

#### Reproduction guide :beetle:
- Start Emacs
- Go to a project directory
- `SPC /`
- Search for something
- `C-c C-e`
- Try to edit the buffer (`c w` for example)

If you disable evil-mode, you can edit it

*Observed behaviour:* :eyes: :broken_heart:
You can't edit the buffer, it uses the normal occur map which doesn't have any
change commands in evil.

*Expected behaviour:* :heart: :smile:
You can edit the buffer

#### System Info :computer:
- OS: darwin
- Emacs: 26.0.91
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 64cf9f56)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: ivy
- Layers:
```elisp
(octave nginx better-defaults spacemacs-layouts ivy
        (markdown :variables markdown-live-preview-engine 'vmd)
        (syntax-checking :variables syntax-checking-enable-tooltips t)
        (auto-completion :variables auto-completion-enable-sort-by-usage t auto-completion-enable-snippets-in-popup nil spacemacs-default-company-backends
                         '(company-capf
                           (company-dabbrev-code company-keywords)
                           company-files company-dabbrev))
        git dash pdf-tools html
        (org :variables org-enable-github-support t org-enable-reveal-js-support t org-want-todo-bindings t)
        colors
        (osx :variables osx-command-as 'super)
        github
        (shell :variables shell-default-shell 'ansi-term shell-default-height 30 shell-default-position 'bottom)
        (spell-checking :variables spell-checking-enable-by-default nil)
        version-control tmux yaml docker restclient lua
        (unicode-fonts :variables unicode-fonts-force-multi-color-on-mac t)
        emoji aj-elixir aj-emacs-lisp aj-javascript aj-term aj-typescript aj-ruby auto-correct cleverparens-lispy dotenv flow match-indent shift-number)
```
- System configuration features: JPEG RSVG IMAGEMAGICK NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS MODULES LCMS2

aaronjensen:
ping @abo-abo just in case you have an idea if anything has changed in this area.

aaronjensen:
[...]

PR #10873, opened 2018-06-13
use evil normal mode as default for ivy-occur-grep-mode

ztlevi:
Recent updates break default using `evil-normal-state` for `ivy-occur-grep-mode`. In order to use `evil-normal-state` after invoking `spacemacs//counsel-edit (C-c C-e)` under `ivy-minibuffer`, this commit hooks it.

I tried `evil-make-overriding-map` for `ivy-occur-grep-mode` but it's still ended with `evil-motion-state`.

### DIFF
--- a/layers/+completion/ivy/packages.el
+++ b/layers/+completion/ivy/packages.el
@@ -202,6 +202,7 @@
       (global-set-key (kbd "C-c C-r") 'ivy-resume)
       (global-set-key (kbd "<f6>") 'ivy-resume)
       ;; Occur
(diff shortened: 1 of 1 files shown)


Miciah:
Is this the same issue as #10290? Does `(evil-set-initial-state 'ivy-occur-grep-mode 'normal)` work?

ztlevi:
@Miciah  Yes! I just rebased my commit to use `evil-set-initial-state`. Thanks.

bmag:

PR #10880, opened 2018-06-14
Load spaceline-all-the-icons in neotree's post-init.

whirm:
Otherwise it will fail with 
```
Error (use-package): spaceline-config/:config: Symbol’s function definition is void: spaceline-all-the-icons--setup-neotree
```
when enabled.


whirm:
Are all this failures caused by my patch?

whirm:
Closing as an alternative fix has been pushed to `devel`.

syl20bnr:
Thank you or the PR. Sorry that this has be fixed without merging your PR. I hope it won't prevent from submitting new PRs.

whirm:
It won't. Also the other fix was actually nicer than mine :)

PR #10884, opened 2018-06-14
[WIP][CORE] synchronize the environment variables from the default shell

fiveNinePlusR:
hopefully fixes the env caching problem by spawning a shell and synchronizing the environment variables from that.

### DIFF
--- a/layers/+distributions/spacemacs-bootstrap/funcs.el
+++ b/layers/+distributions/spacemacs-bootstrap/funcs.el
@@ -185,42 +185,3 @@ Example: (evil-map visual \"<\" \"<gv\")"
   "Custom hint documentation format for keys."
   (format (format "[%%%ds] %%%ds" key-width (- -1 doc-width))
           key doc))
-
-
-
-;; exec-path-from-shell
-
-(defun spacemacs//initialize-exec-path-from-shell (&optional force)
-  "Initialize exec-path and cache its value.
-Load from cache file if cache file exists and FORCE is nil."
-  (when (display-graphic-p)
-    (when force (delete-file spacemacs-env-vars-file))
-    (if (file-exists-p spacemacs-env-vars-file)
-        (load spacemacs-env-vars-file nil (not init-file-debug))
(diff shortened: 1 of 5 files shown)


bleggett:
This is generally and historically a macOS problem, so it probably makes sense to disable it by default on other platforms.

Also, it is only a problem with **vanilla** Emacs macOS builds, the main Emacs macOS porter [maintains a modified macOS build](https://bitbucket.org/mituharu/emacs-mac/overview) that patches the env var inheritance problem, so even for macOS I would like an option to disable this and not take the startup hit.

fiveNinePlusR:
it is a problem in the main version that the spacemacs manual says to install though. It's not present on the emac-mac port but it's definitely a bug/doesn't import the shell env with emacs-plus.

bleggett:
@fiveNinePlusR As long as I can disable it with no ill effects I'm fine with it. I know we want things to work with vanilla Emacs because that's what we want most people to use, but if vanilla Emacs is borked and I'm running a variant that's not I don't want to take the hit.

fiveNinePlusR:
you should be able to put spacemacs-environment into your excluded packages and it should work fine. if @syl20bnr wants to we can add in a variable to disable it... something like `(defvar spacemacs-import-env t "import all environment values from your default shell")`

bleggett:
Tested this PR on macOS/emacs-mac 26.1, I cannot detect or measure any significant perf hit, so I am fine with this as a non-toggleable workaround for vanilla Emacs.

ljupchokotev:
Tested this PR with the `go` layer and I didn't find any issues. `GOPATH` and `GOROOT` are set, which means all the go tools work.

I didn't notice any performance hit either.

syl20bnr:
Thank you ! 💜 

I made it asynchronous and added some dotfile variables for better control over it, see commits: https://github.com/syl20bnr/spacemacs/commit/9ee832955 and  https://github.com/syl20bnr/spacemacs/commit/d28990537

PR #10887, opened 2018-06-14
Additional documentation for spacemacs/toggle-current-window-dedication

axelson:
Add additional user documentation to the `spacemacs/toggle-current-window-dedication` function. The additional documentation is meant to aid discovery of the meaning of this function (via `SPC h d f spacemacs/toggle-current-window-dedication`) for users that are not yet completely familiar with Emacs.

Description is based on content from: http://www.gnu.org/software/emacs/manual/html_node/elisp/Dedicated-Windows.html


syl20bnr:
Thank you !

axelson:
Awesome, thanks!

PR #10888, opened 2018-06-14
spacemacs-navigation: remove duplicate code

ellakk:
I think this code was forgotten about during a refactoring. It exists and is in use in spacemacs-visual layer and zoom-frm is initialized there.


syl20bnr:
Thank you ! Nice catch 👍 

Issue #10865, opened 2018-06-13
Error (use-package): flycheck/:init: Symbol’s value as variable is void: flycheck-error-list-mode-map


Electron-libre:
#### Description :octocat:

Error when initializing flycheck package.

```
Error (use-package): flycheck/:init: Symbol’s value as variable is void: flycheck-error-list-mode-map
```

#### Reproduction guide :beetle:
- Start Emacs

*Observed behaviour:* :eyes: :broken_heart:

The warning window report an error

```
Error (use-package): flycheck/:init: Symbol’s value as variable is void: flycheck-error-list-mode-map
```

*Expected behaviour:* :heart: :smile:

No error in warning window.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 26.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 8754b2450)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(helm
 (auto-completion :variables auto-completion-enable-help-tooltip t auto-completion-enable-snippets-in-popup t auto-completion-enable-sort-by-usage t)
 rust better-defaults emacs-lisp git markdown org
 (shell :variables shell-default-height 30 shell-default-position 'bottom)
 spell-checking
 (syntax-checking :variables syntax-checking-enable-tooltips nil ispell-dictionary "english")
 csv javascript nginx docker elixir elm html ruby ruby-on-rails
 (haskell :variables haskell-enable-hindent-style "chris-done" haskell-completion-backend 'ghc-mode)
 sql shell-scripts yaml themes-megapack
 (keyboard-layout :variables kl-layout 'bepo)
 erc ansible floobits fsharp neotree)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GSETTINGS NOTIFY ACL GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 MODULES THREADS LIBSYSTEMD LCMS2

CharlesHD:
[...]

PR #10893, opened 2018-06-15
Keyboard-Layout : after-config hooks for lazy-loaded packages

CharlesHD:
Fix for #10865 .
New lazy-loaded packages using spacemacs|add-transient-hook introduced in bd316f4c308931a787236e8b7e4844f25e551a03 have their setup at `:config` phase instead of `:init` phase. That breaks some keyboard-layout expectation of keymap defined at `:init` time. 
This PR update the two impacted packages in keyboard-layout i'm aware of : evil-surround and flycheck.


syl20bnr:
Thank you ! 👍 

PR #10901, opened 2018-06-15
Add copy-as-format layer

rynffoll:
Added layer for integration with [copy-as-format](https://github.com/sshaw/copy-as-format).


Compro-Prasad:
Seems like an useful layer.

sdwolfz:
Thank you :heart:!

Issue #10897, opened 2018-06-15
Develop branch does not properly read environment variables.


dangom:
#### Description :octocat:
Develop branch does not properly read environment variables.

#### Reproduction guide :beetle:
- Start Emacs
- Run (getenv "PATH") and cry

*Observed behaviour:* :eyes: :broken_heart:
PATH only contains "/usr/bin:/bin:/usr/sbin:/sbin"

*Expected behaviour:* :heart: :smile:
PATH contains everything I defined in .bash_profile and .profile.local.

#### System Info :computer:
- OS: darwin
- Emacs: 26.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 1faf4d3ee)
- Graphic display: t
- Distribution: spacemacs-base
- Editing style: emacs
- Completion: helm
- Layers:
```elisp
(lua spacemacs-completion spacemacs-defaults spacemacs-layouts spacemacs-editing spacemacs-navigation spacemacs-modeline spacemacs-project spacemacs-purpose spacemacs-org spacemacs-visual helm
     (auto-completion :variables auto-completion-return-key-behavior nil auto-completion-tab-key-behavior 'complete auto-completion-complete-with-key-sequence "jk" auto-completion-complete-with-key-sequence-delay 0.25 auto-completion-enable-sort-by-usage t company-mode-completion-cancel-keywords
                      '("do" "then" "case" "begin" "ipython" "<ip"))
     better-defaults emacs-lisp git markdown
     (org :variables org-enable-reveal-js-support t org-enable-bootstrap-support t org-enable-org-journal-support t)
     (shell :variables shell-enable-smart-eshell t shell-default-height 30 shell-default-position 'bottom)
     spell-checking syntax-checking version-control javascript
     (python :variables python-fill-column 79 python-sort-imports-on-save t python-tab-width 4)
     ipython-notebook shell-scripts csv common-lisp html
     (c-c++ :variables c-c++-default-mode-for-headers 'c++-mode c-c++-enable-clang-support t)
     (mu4e :variables mu4e-installation-path "/usr/local/share/emacs/site-lisp/mu" mu4e-contexts t mu4e-enable-notifications nil)
     ibuffer dash yaml major-modes
     (latex :variables latex-build-command "LatexMk")
     bibtex pdf
     (scimax :variables scimax-org-return-dwim t))
```
[...]

PR #10905, opened 2018-06-16
[osx] use init instead of post-init for exec-path-from-shell.

Chream:
Small error where a post-init function was defined when the layer actually owns the package.

#10897  might be relevant.


smile13241324:

syl20bnr:
I reverted this PR since Spacemacs does not use `exec-path-from-shell` anymore.
See https://develop.spacemacs.org/doc/DOCUMENTATION.html#default-behavior

You can use this PR thread to mention any issue you may have with the `.spacemacs.env` file, I'll monitor this thread for a bit. 

Thank you for the contribution.

PR #10907, opened 2018-06-17
typescript: Fix jump handling with multiple backends

aaronjensen:
Because the setup functions run after local vars are set, setting jump handlers needs to happen directly. Otherwise setting them only applies to the next typescript file that is opened, which may or may not even have the same backend.

### DIFF
--- a/layers/+lang/typescript/funcs.el
+++ b/layers/+lang/typescript/funcs.el
@@ -40,10 +40,6 @@
       (kbd "C-k") 'tide-find-previous-reference
       (kbd "C-j") 'tide-find-next-reference
       (kbd "C-l") 'tide-goto-reference)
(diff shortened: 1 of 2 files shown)


FieryCod:
@syl20bnr @aaronjensen Is there a possibility to merge this?

Compro-Prasad:
@aaronjensen Can you give the reproduction steps for what this PR about?

aaronjensen:
1. open spacemacs
2. open a typescript file
3. attempt to `gd` on a symbol

expected:

it goes to the definition

actual:

it prompts for tag file because the jump handlers are not yet registered.


If you then `M-x typescript-mode` and try again, it will work as expected.

aaronjensen:
Okay, this is now properly tested and does not break lsp mode. Once again ready to merge.

sdwolfz:
Thank you :heart:!

PR #10908, opened 2018-06-17
shell: enable prompt autojump when using smart-eshell

ellakk:
Does what the old comment in the code suggested:
`
;; Idea: maybe we could make auto-jump smarter and jump only if point is not on a prompt line
`


smile13241324:

PR #10919, opened 2018-06-19
Fix javascript layer enables lsp-mode even when using tern

sei40kr:
Fix the issue javascript layer enables `lsp-mode` even if the backend is tern.

### DIFF
--- a/layers/+lang/javascript/packages.el
+++ b/layers/+lang/javascript/packages.el
@@ -170,8 +170,6 @@
 (defun javascript/init-lsp-javascript-typescript ()
   (use-package lsp-javascript-typescript
     :commands lsp-javascript-typescript-enable
-    :defer t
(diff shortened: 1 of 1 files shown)


sei40kr:
`lsp-ui-mode` selects a flycheck checker `lsp-ui` and it disables other flycheck checkers.

sdwolfz:
I merged https://github.com/syl20bnr/spacemacs/pull/10795 which also removed the `:init`. Is your removal of `deffer: t` still needed?

ztlevi:
I've already fixed this in both Javascript layer and python layer. Both are already merged into develop branch. From my perspective, `defer t` is needed since `js2-mode-local-vars-hook` has `lsp-javascript-typescript-enable` function that will enable everything until we open a javascript buffer.

sdwolfz:
OK I will close this. Thank you both for helping out :heart:!

PR #10921, opened 2018-06-19
diff-mode: define key bindings with leader key for vim editing style

miyabinomuratic:
This patch defines key bindings with leader key instead of evilified state to
make it easier for people those who edit in vim style to use diff-mode for both
of editing and reviewing files/buffers.

There are two use cases for diff-mode:

(1) to manually edit diff in a patch file
(2) to review diff that is generated by `vc-diff' or similar commands

The evilified state is useful for the case (2), but is confusing for the
case (1). Usually, users of vim editing style expect that they are in normal
state when a new buffer is created. However, when evilified state is used for
the case (1), some keys insert their character in a buffer, which let the users
think they are in insert state and keep hitting keys to edit the buffer, while
other keys invoke commands or change contents of buffer, which let them think
they are in normal state, so finally the user get lost what state they are in if
they don't know evilified state is used for diff-mode.

The changeset 58d521af5 "Unbreak diff-mode", originally written on Apr 2 2017,
tried to avoid the confusion by removing the evilified state configuration.
However, it is overwritten by the other and older changeset 8009e1bf5 "evilify
vc-* buffers", which is written on Feb 22 2016, for some reason.

This patch respects both of the use cases listed above, thus defines key
bindings with leader key instead of just removing the evilified state
configuration.

### DIFF
--- a/layers/+source-control/version-control/README.org
+++ b/layers/+source-control/version-control/README.org
@@ -10,7 +10,7 @@
 - [[#key-bindings][Key Bindings]]
   - [[#vc-directory-buffer-commands][VC Directory buffer commands]]
   - [[#commit-message-editing-buffer][Commit message editing buffer]]
-  - [[#diff-buffer][Diff buffer]]
+  - [[#diff-mode][Diff mode]]
   - [[#log-view-buffer][Log view buffer]]
   - [[#annotation-buffer][Annotation buffer]]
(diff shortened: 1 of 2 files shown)


syl20bnr:
Thank you ! 👍 

Nice investigation!

miyabinomuratic:
Great. Thank you!

Issue #1369, opened 2015-04-25
Spacemacs gets stuck for a few seconds and hogs CPU(100% usage) from time to time


x-ji:
~~Now https://github.com/syl20bnr/spacemacs/issues/1300 is solved, Emacs no longer eats up memory and totally freezes.~~ (Update: https://github.com/syl20bnr/spacemacs/issues/1300 now happened again as well...) However, I still encounter random situations where Emacs just ramps up its CPU usage all the way to 100% from time to time, and gets stuck for at least 5 seconds. It happens almost every 30 seconds now. 

The first screenshot is when CPU usage peaks. Subsequently it drops rapidly to only about 1% and Emacs responds normally again.

![screenshot 2015-04-25 17 11 58](https://cloud.githubusercontent.com/assets/2999209/7332543/6dea4a1a-eb76-11e4-9900-5c8d12a04b8d.png)
![screenshot 2015-04-25 17 12 04](https://cloud.githubusercontent.com/assets/2999209/7332544/72b90914-eb76-11e4-9d63-8fcd4430631e.png)
![screenshot 2015-04-25 17 12 09](https://cloud.githubusercontent.com/assets/2999209/7332545/74cea006-eb76-11e4-971e-d9f0940e540b.png)

When this happens I'm not doing anything except for navigating a file using Vim bindings. It's weird because I've erased all the variables set via `Custom` but it doesn't help.

Also, pressing `SPC` and mouse-clicking on Emacs could both also induce the issue.

I think other users mentioned similar situations in both https://github.com/syl20bnr/spacemacs/issues/1300#issuecomment-95020725 and https://github.com/syl20bnr/spacemacs/issues/868#issuecomment-83773149

x-ji:
Deleted both `.emacs.d` and `.spacemacs` and did a fresh reinstall again... This time didn't add any private layer and things seem to be working. Let me just suppose for now it's induced by some package not playing well with Spacemacs.

danielwuz:
FYI, I recently ran into an issue of noticeable stuck when doing copy and paste. I eventually was able to find out it's related to `pbcopy` in `osx` layer, not sure why that happened, still need to look further into this problem.

x-ji:
Now it happened again... I now suppose it's an issue of `semanticdb`, because it only occurred after Spacemacs asked me to create `semanticdb` folder and I clicked yes.

tuhdo:
[...]

PR #10922, opened 2018-06-19
Delete duplicates in history & remove limit

Compro-Prasad:
- Removing duplicates is very useful

- Considering that everyone might have varying length of history it is better to preserve it and stop complaining about any limits.

- Saving the kill ring is also beneficial. See https://github.com/syl20bnr/spacemacs/pull/10922#issuecomment-400726974 how to enable it.

Please suggest whether to prioritize kill ring or the history limit.

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/packages.el
+++ b/layers/+spacemacs/spacemacs-defaults/packages.el
@@ -297,17 +297,15 @@
   (use-package savehist
     :init
     (progn
-      ;; Minibuffer history
-      (setq savehist-file (concat spacemacs-cache-directory "savehist")
-            enable-recursive-minibuffers t ; Allow commands in minibuffers
-            history-length 1000
-            savehist-additional-variables '(mark-ring
-                                            global-mark-ring
-                                            search-ring
-                                            regexp-search-ring
-                                            extended-command-history)
-            savehist-autosave-interval 60)
-      (savehist-mode t))))
+      (setq enable-recursive-minibuffers t) ; Allow commands in minibuffers
+      (when spacemacs-minibuffer-history-length
+        ;; Minibuffer history
+        (setq savehist-file                 (concat spacemacs-cache-directory "savehist")
+              history-length                spacemacs-minibuffer-history-length
+              history-delete-duplicates     t
+              savehist-additional-variables spacemacs-minibuffer-history-objects
+              savehist-autosave-interval    spacemacs-minibuffer-history-save-interval)
+        (savehist-mode t)))))
 
 (defun spacemacs-defaults/init-saveplace ()
   (use-package saveplace
--- a/layers/+spacemacs/spacemacs-defaults/config.el
+++ b/layers/+spacemacs/spacemacs-defaults/config.el
@@ -31,6 +31,24 @@
   "Regexp used to define buffers that are useful despite matching
 `spacemacs-useless-buffers-regexp'.")
 
+;; Variables to control saving of minibuffer history
+(defvar spacemacs-minibuffer-history-length t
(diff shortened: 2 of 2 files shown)


layers/+spacemacs/spacemacs-defaults/packages.el
```diff
@@ -300,8 +300,10 @@
       ;; Minibuffer history
       (setq savehist-file (concat spacemacs-cache-directory "savehist")
             enable-recursive-minibuffers t ; Allow commands in minibuffers
-            history-length 1000
+            history-length t
+            history-delete-duplicates t
             savehist-additional-variables '(mark-ring
+                                            kill-ring
```
sdwolfz: This might be a bad idea if you're copy-pasting passwords...

Compro-Prasad: So, you suggest removing this functionality?

sdwolfz: Not necessarily. You could have a toggle for this, off by default, with a a setting and a warning documented.

Compro-Prasad: Sure. Will update this.

Compro-Prasad: I currently don't know how to add this kind of setting as it requires more logic here and there(by seeing implementation of variables in `dotspacemacs/init`). Can I just remove `kill-ring`?

Compro-Prasad: I can do it. It will take some time though.

sdwolfz: There's no rush. I can come back to this PR and give you some pointers in case you are stuck.

Compro-Prasad: Should I force push a single commit everytime?

sdwolfz: Personally I prefer a single commit, but if it's easier you can push intermediary commits and squash them in the end.

Compro-Prasad: I am defining the variables for spacemacs-default layer and not in the core. Is that okay?

sdwolfz: Not sure, please push your changes and I will comment on the line of code once I see it (and have more context).

Compro-Prasad:
Updated it!

Compro-Prasad:
I can use:
```
(spacemacs-defaults :variables
                         spacemacs-minibuffer-history-length 1024
                         spacemacs-minibuffer-history-objects '(kill-ring mark-ring global-mark-ring search-ring regexp-search-ring extended-command-history))
```
to set the variables.

Compro-Prasad:
Why should this PR be merged?
Because it works.

Why should this PR not be merged?
Please let me know why.

What improvements can be made?
These variables can be moved to core from spacemacs-defaults.

sdwolfz:
This looks OK to me, I just need a little time (and energy) to try it out locally before merging.

Do you think you can find a proper place in the documentation to write about these new vars, also adding instructions on how to add the kill-ring there, with a (big) warning about passwords being saved?

syl20bnr:
I’m fine with this although I don’t like we duplicate Emacs variables... Can we keep the vanilla variables instead or does our variables provide a better/easier interface ?

Also keep in mind that we once disabled this because some users got performance issue: https://github.com/syl20bnr/spacemacs/issues/1369

Compro-Prasad:
Will be back when its needed.

Issue #10870, opened 2018-06-13
develop branch, typescript layer: jump-to-type-def not defined


bschwehn:
#### Description :octocat:
In a .ts file, using the typescript layer, when I press `,gt` I get the error

[userdir]/.emacs.d/elpa/26.1/develop/tide-20180608.239/tide.elc failed to define function typescript/jump-to-type-def

I can see that in commit 20bafdbb923 - 2017-14-01 the function "typescript/jump-to-type-def" was removed in typescript/packages.el. in the develop branch. It is still there in the master branch. 

Not sure how this should now work after the change, but when I simply re-add the function to file .emacs.d/layers/+lang/typescript/packages.el:

```elisp
(defun typescript/jump-to-type-def()
  (interactive)
  (tide-jump-to-definition t))
```

the short cut works again as expected. I assume however, that the change was intentional and this is not a valid fix. 

#### Reproduction guide :beetle:
- Start Emacs
- open a typescript file with typescript layer enabled
- go to a class name
- press `,gt` (or `SPC m g t`)

*Observed behaviour:* :eyes: :broken_heart:
- observe error in *Messages* buffer:
[userdir]/.emacs.d/elpa/26.1/develop/tide-20180608.239/tide.elc failed to define function typescript/jump-to-type-def

*Expected behaviour:* :heart: :smile:
tide-jump-to-definition is called and navigation to definition happens

#### System Info :computer:
- OS: windows-nt
- Emacs: 26.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 8754b24)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(auto-completion better-defaults emacs-lisp git org javascript html syntax-checking semantic treemacs yaml typescript)
```
-

PR #10926, opened 2018-06-20
typescript: re-map keybinding to provide jump to def

roychoo:
Hi,

As mentioned on this issue https://github.com/syl20bnr/spacemacs/issues/10870,

it seems to me the function is removed, not sure if the change was to use particular tide function directly.

 one question i have is since ```tide-jump-to-definition``` handles type too, should we just use ```SPC m g g``` and remove ```SPC m g t```

### DIFF
--- a/layers/+lang/typescript/funcs.el
+++ b/layers/+lang/typescript/funcs.el
@@ -153,3 +153,7 @@
 
 (defun spacemacs/typescript-yasnippet-setup ()
   (yas-activate-extra-mode 'js-mode))
+
+(defun spacemacs/typescript-jump-to-type-def ()
+  (interactive)
+  (tide-jump-to-definition t))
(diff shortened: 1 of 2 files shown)


sdwolfz:
From the function description here: https://github.com/ananthakumaran/tide/blob/master/tide.el#L728 I see that it should jump to type definition only when executed with a prefix. So `SPC m g g` should be `tide-jump-to-definition` and `SPC m g t` should be `tide-jump-to-definition t`.

I think `typescript/jump-to-type-def` was accidentally removed and the intent was to define it outside of `typescript/init-tide` in `layers/+lang/typescript/funcs.el`. Do you think you can add it back and implement both keybindings (don't forget to document them as well)?

roychoo:
@sdwolfz i see what you mean and yes i think you are right.

i had made the changes. regarding the documentation, it was already documented in the README.
https://github.com/syl20bnr/spacemacs/tree/develop/layers/%2Blang/typescript#typescript-major-mode

Unless i misunderstood what you mean.

sdwolfz:
Thank you :heart:! 

My mistake, I thought the documentation was removed with the previous changes.

roychoo:
Awesome and thanks @sdwolfz

PR #10945, opened 2018-06-23
add syntax checking for jsonnet with flycheck

robbyoconnor:
Add syntax checking for jsonnet.


robbyoconnor:
I had a PR #10931 -- that I somehow borked...so...this reopens it.

PR #10947, opened 2018-06-24
Adds `*Process List*` to `popwin:special-display-config`

simendsjo:
`list-processes` would not focus the newly created window as it was missing from popwin.


bmag:

Issue #10906, opened 2018-06-16
Env vars are ignored when using GUI in macOS


Andre0991:
When using GUI Emacs in macOS,`(shell-command-to-string "env")` ~displays my env vars defined in `.bashrc`~, but not the ones in `.bash_profile`.

My config is [here](https://github.com/Andre0991/dotfiles/blob/master/.spacemacs).

#### System Info :computer:
- OS: darwin
- Emacs: 25.3.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 1abce5ec1)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: ivy
- Layers:
```elisp
(clojure common-lisp emacs-lisp erc git github html ivy javascript latex markdown org osx plantuml shell-scripts scala spacemacs-layouts syntax-checking themes-megapack theming version-control
         (shell :variables shell-default-shell 'eshell)
         (elfeed :variables rmh-elfeed-org-files
                 (list "~/Dropbox/backup/emacs/elfeed-feeds.org"))
         (c-c++ :variables c-c++-enable-clang-support t)
         (auto-completion :variables auto-completion-return-key-behavior 'complete auto-completion-tab-key-behavior 'cycle auto-completion-enable-help-tooltip nil auto-completion-enable-sort-by-usage t auto-completion-enable-snippets-in-popup nil :disabled-for latex)
         (spell-checking :variables spell-checking-enable-by-default nil enable-flyspell-auto-completion t)
         andre-common-lisp andre-elfeed andre-erc andre-eww andre-ivy nov)
```
- System configuration features: JPEG RSVG IMAGEMAGICK NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS MODULES

Also, I use `bash` from Homebrew. My default shell is `/usr/local/bin/bash` (correct according to `env`).

Related to https://github.com/syl20bnr/spacemacs/issues/10897.

Andre0991:
Other interesting thing to notice is that Homebrew added  `/usr/local/bin` to `/etc/paths`, which ends up added to PATH, but Spacemacs is not picking it up, so I can't use Homebrew commands from Emacs.

Andre0991:
I just updated to the latest develop version (`d3c594512d4e683edfa604053c424aa88c41b240`),  seems that even the vars from `.bashrc` are not being read.

[...]

PR #10951, opened 2018-06-24
Revert changes to PATH setup on develop to fix breakage and start over

Blaisorblade:
Hi all! Thanks for Spacemacs, it's really cool! But recent changes to PATH loading (including #10884) do not actually load the correct PATH because `bash -c env` is essentially NEVER the correct command to call, due to intricacies of shell startup — people tested #10884, but nobody describes their setup or even confirms the need to fix their Emacs PATH. So I can't find evidence the documented testing exercised interesting scenarios (see **Questions** below).

Due to the intricacies of shell startup and the amount of shells to support, I'm convinced we must start from https://github.com/purcell/exec-path-from-shell and its customization options.

So, while this PR is meant as a basis for discussion, I can't avoid reverting all the relevant changes from @syl20bnr and @fiveNinePlusR. We should probably allow again users to disable `exec-path-from-shell`, as @blegget suggested in #10884 (but I'm not sure where that would go).

If Spacemacs has different requirements we *might* be able to simplify the code a bit, but this must be done much more carefully. For starters, before introducing risky optimizations somebody should demonstrate any speedup, and they should be tested in enough different scenarios that *need* the PATH setting.

## Details

That's because `bash -c env` doesn't give the right info (it only gives the default PATH, here `PATH=/usr/bin:/bin`) — you'd need at least `bash -l -c env` (which accounts for PATH settings in, for instance, `~/.profile`, `~/.bash_profile`, `/etc/paths` and `/etc/paths.d`). Still, that doesn't account for all scenarios accounted by `exec-path-from-shell`.

Calling `bash -c env` in your terminal might seem to work, since there PATH *is* already initialized — what worked for me was adding these lines to `crontab -e` and checking my local mail in `/var/mail/$USERNAME` with an editor, tho YMMV.

```
*/1 * * * * /usr/local/bin/bash -c env
*/1 * * * * /usr/local/bin/bash -l -c env
```

Hence, those changes completely break loading the agda layer, but also code based on `call-process` (and relatives) in `.spacemacs`.

## Why was the new code faster? Is there space for speedups?

It's not clear why you'd save time while still starting a shell. Skipping the initialization done by login shells saves some time, but defeats the point of this code. `exec-path-from-shell` runs *two* shells to check for broken setups, but there's a setting to disable that (and the second shell could be run asynchronously), and it uses `printf`, but I don't see 

Doing this asynchronously of course saves time, but why not use that with `exec-path-from-shell`?
And you need to wait for the loading to complete before loading layers that run programs during setup (e.g. agda). But if *that* is the point, you can just call `exec-path-from-shell` asynchronously.

## Could we revert less?

To find out the problem, I tried testing the intermediate states to do a smaller change, but I failed to find one that worked. Even c12b72090ca000ba1261f13ec69d9fb986be6b16 failed, tho it still builds on `exec-path-from-shell`. Apparently, that caching only saved `exec-path`, but Emacs requires also setting `PATH` (and docs require setting `MANPATH`) — nobody called `(spacemacs/copy-env-list '("PATH" "MANPATH"))`.

## Questions

What's the setup for @fiveNinePlusR, @blegget, @ljupchokotev, @syl20bnr? @blegget confirmed his PATH is setup correctly, does anybody else have a setup that *breaks* without `exec-path-from-shell`? For context, my Spacemacs startup is plenty fast already, so I'm not sure I can observe the speedup. I'd also fine with a `.spacemacs` setting to load `exec-path-from-shell` properly, even if it defaults to off — but the current code just doesn't work.

## Merging

I'd merge this as-is while figuring out a proper fix — but if anybody wants to try out optimizations, feel free to. But any followup changes should please go on this branch to be tested before merging it, please.

While this is not my project and how to merge is up to you, I'd also like to actually merge this PR as such (feel free to squash-merge) instead of bypassing GitHub's tracking, otherwise people debugging breakage from here will struggle piecing the history together as much as I did. And this code is *hard* to get right and test on all scenarios, it'd be nice people *will* need to check this PR again.

Finding all the relevant commits and discussion was very tricky *because* the relevant commits were cherry-picked, and then fixes were intermingled without a feature branch.


fiveNinePlusR:
Basically, `exec-path-from-shell` fundamentally does the exact same thing that the env parsing code does except it does not do it in one single call to `env` but with a bunch of calls to `printf "$SOME_VAR"` (and a new shell is spawned each time unless grouped up into one call) 

The fix is to either tell people to put the required path and other non-interactive required variables in the .profile, .bash_profile or .zshenv files instead. Those are supposed to be used by non-interactive shells and are the correct place to put those variables or to call the shell with `-i` and `-l` or `-d` depending on the shell.

The reason that `exec-path-from-shell` works is that they use `-i` as well as `-l`. there is also a check for if the shell is tchs or chs and it supplies a `-d` to it instead. 

I don't think that this code is really all that tricky to warrant an extra package that is slow. We can just add `-i` to the shell commands and call it a day.

> What's the setup for @fiveNinePlusR

My setup is a `.zshenv` file that sets all my path and common shell variables.

The other way this could be resolved is to have people explicitly set the path and any other required variables using `setenv` inside the `dotspacemacs/init` function or maybe later.  Having the environment inherited is nice to have though.

anton-trunov:
> The fix is to either tell people to put the required path and other non-interactive required variables in the .profile, .bash_profile or .zshenv files instead.

@fiveNinePlusR I keep my variables in `.zshenv` but still the `develop` branch does not work for me (I'm on macOS 10.13.5).
@Blaisorblade Thanks for this PR!

Blaisorblade:
@fiveNinePlusR Thanks for explaining the reasoning behind your change, I was just going off the PR comments and commit messages. I'm not yet convinced tho.

> We can just add -i to the shell commands and call it a day.

No: in general, one could study if that is actually robust make a PR with that change, wait for enough people to test it, and then merge it.

In particular, that change just doesn't help.

But your suggestion doesn't square with the amount of issues https://github.com/purcell/exec-path-from-shell/issues has, so I'd rather push the maintenance burden upstream as far as possible.

> The fix is to either tell people to put the required path and other non-interactive required variables in the .profile, .bash_profile or .zshenv files instead.

But my vars *are* in `.profile`, as they're supposed to be, that's just *not* loaded by `bash -c env`, nor by `bash -c -i env` (I just double-checked). If anything, the quickfix would be to add `-l` for sh-like shells — but in that package those options are all customizable because users have very different setups, and once you readd the customization you've readded half of the code.

> with a bunch of calls to printf "$SOME_VAR" (and a new shell is spawned each time unless grouped up into one call)

Multiple calls are indeed be inefficient, but most users only need one call for PATH and MANPATH — users of Go and Rust are an exception, so that doesn't yet explain the change.

I don't know of any problem using `env`, but if it *is* a valid change it can be done upstream (or by forking that package). However, this was already proposed in https://github.com/purcell/exec-path-from-shell/issues/23 and rejected, since `env`'s output cannot be parsed reliably.

> My setup is a .zshenv file that sets all my path and common shell variables.

Oh, that explains why you don't need to pay attention to `-l` here. It's cool zsh has that feature, tho I suspect you'd still need `-l` for `zsh` (never used it, not sure). But what if you want to set up environments just for login shells? `PATH=/some/path:$PATH` is wasteful to run twice, so it seems it'd go in `.zprofile`. Or do people write a smarter PATH extension function?
Anyway, `exec-path-from-shell` uses `-l` for zsh too, correctly, and some user have setup bugs such that they prefer to have it off (https://github.com/purcell/exec-path-from-shell/issues/40).

Blaisorblade:
@anton-trunov I'm not sure why the current code would fail for you, but based on https://github.com/purcell/exec-path-from-shell/issues/23, do some of your vars contain `=` or `\n` or other special chars?

syl20bnr:
Nice PR, thank you for taking the time to do it and for the explanation and dig in, this is highly appreciated.

`exec-path-from-shell` is really slow as it can fetch env vars one at a time but more importantly I think that all this "getting env. vars from shell` is not reliable enough and feels more like an heuristic than proper initialization, this is mostly because shell environments initialization is complicated and there are too many variables to take into account.

So now Spacemacs has its own `.env` file with proper key bindings to edit it and reload it. For convenience Spacemacs will try to initialize this file with your shell environment the first time. A message is displayed about it when this is the case.

More info in message: https://github.com/syl20bnr/spacemacs/issues/10906#issuecomment-399852319

CarlQLange:
I was wondering why my `eshell` stopped working right! Great discussion, very lucid

anton-trunov:
@Blaisorblade This is my `.zshenv`:
```shell
typeset -U path

path=(
  $HOME/prj/coq/{coq-tools,coq-bench}
  $HOME/.cargo/bin
  $HOME/.local/bin
  $HOME/.cabal/bin
  /usr/local/{bin,sbin}
  /usr/{bin,sbin}
  /{bin,sbin}
  $path
)
```

I does not contain any weird characters (and `$HOME` is `/Users/antontrunov`), so I have no idea why it did not work.

Blaisorblade:
Since this PR was rejected for a different plan, closing.

Issue #2324, opened 2015-07-14
ERC notifications not working correctly on OS X


gunnarhelgason:
I've enabled the [erc layer](https://github.com/syl20bnr/spacemacs/tree/master/contrib/!irc/erc) in .spacemacs and installed the terminal-notifier gem. However, instead of receiving notifications I get the following error in emacs: `error in process filter: peculiar error: "Emacs not compiled with dbus support"`

If I add `(require 'erc-terminal-notifier)` to my .spacemacs, the notifications are delivered to the OS X notification center, but the dbus error persists.

The documentation for the erc layer does not specify that you have to do anything special (other than installing terminal-notifier) to get notifications to work on OS X. Is this a bug, or have I misunderstood something? I assume that at least the dbus error is a bug as dbus isn't available on OS X AFAIK.

CestDiego:
Sorry cannot reproduce ;_; don't have osx. :3 maybe one of the osx users can help.

cmccloud:
I was having a similar error. I don't make use of notifications in my own erc configuration, but the error seems to be coming from this block of code:
```
      (defun erc-global-notify (match-type nick message)
        "Notify when a message is recieved."
        (notifications-notify
         :title nick
         :body message
         :app-icon "/home/io/.emacs.d/assets/spacemacs.svg"
         :urgency 'low))

      (add-hook 'erc-text-matched-hook 'erc-global-notify)
```
notifications-notify requires d-bus. 
I fixed it by adding the following line to my post-erc configuration:

```
(remove-hook 'erc-text-matched-hook 'erc-global-notify)
```
@gunnarhelgason if you try this out and it works for you let me know and I'll look into making a change to the erc package layer to only add that hook if it detects dbus support.

asheshambasta:
@cmccloud sorry if this is a stale issue and I'm bringing it up again – but I can still reproduce this even with 
```
(remove-hook 'erc-text-matched-hook 'erc-global-notify)
``` 
At the end of my init file.

Miciah:
Are you on macOS? What do you get if you open the eval prompt (`M-:`) and enter `(bound-and-true-p dbus-compiled-version)`?

PR #10954, opened 2018-06-25
Clojure layer: fix function usage in clojure layer according to cider changes

andrestylianos:
Clojure layer: fix function usage in clojure layer according to cider changes

The function `cider-current-repl-buffer` does not exist anymore in cider after PR [#2324](https://github.com/clojure-emacs/cider/pull/2324) was merged.
It broke the usage of functions in the clojure layer that used `spacemacs//cider-eval-in-repl-no-focus`, which sends forms to be evaluated in the repl buffer without focusing on it.
The fix is to use `cider-current-repl` instead of `cider-current-repl-buffer`


sdwolfz:
Thank you :heart:! 

PR #10960, opened 2018-06-25
enable eldoc for rust/racer

ggarlic:
enable eldoc-mode for rust layer with racer

according to  https://github.com/racer-rust/emacs-racer , eldoc should be added to racer-mode-hook，not rust-mode-hook 

before commit e246ac1957, it did add to rust-mode-hook, the wrong way

### DIFF
--- a/layers/+lang/rust/packages.el
+++ b/layers/+lang/rust/packages.el
@@ -14,7 +14,6 @@
     cargo
     company
     counsel-gtags
-    eldoc
     racer
     flycheck
     (flycheck-rust :requires flycheck)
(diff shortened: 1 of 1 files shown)


layers/+lang/rust/packages.el
```diff
@@ -14,6 +14,7 @@
     cargo
     company
     counsel-gtags
+    eldoc
```
bmag: Unnecessary here, please remove.

ggarlic: done

please check it

bmag:

PR #10965, opened 2018-06-26
Add a binding to return from definition in intero mode

magthe:
Currently there is only a binding for `jump-to-definition` (via the spacemacs jump handlers), but there is no binding for returning. It turns out that Intero's jump-to-definition function uses the `xref` cross-referencing functions, so adding a return-from-definition is as simple as adding a binding for `xref-pop-marker-stack`.


sdwolfz:
Please arrange the keybinding in alphabetical order, and also add it to the documentation.

magthe:
@sdwolfz Done, and I also rebased it onto the latest _develop_ while at it.

sdwolfz:
Thank you :heart:!

Issue #3103, opened 2015-09-22
Consistent eval functions in Vim style


syl20bnr:
To be consistent with the eval functions behavior in elisp we have to advice the eval functions of other lisp modes to be able to eval the expressions when the point is on their closing parens in Vim style
- [ ] clojure

vkz:
this also makes behaviour of `expand-region` inconsistent:
```clojure
(println "hello world")
```
with point on last `)` invoking `er/expand-region` it will initially select `"hello-world"`. Not sure what's the right approach here.

emidln:
I started working some on this last week. I'd like to introduce an evil operator motion for eval that is generic and can be hooked into with whatever mode's particular eval-region or send-region. This gives us a nice way to compose generic sexp-based evaluation functions around standard emacs, evil, and smartparens functionaltiy that will apply equally to scheme, elisp, common lisp, clojure, hy, pixie, etc. Further, most of this is also useful (in particular something like eval-paragraph or eval-line) in any language with an interpreter (Python, Ruby, Javascript, etc). I bind `SPC m e o` to be the generic-eval-region operator, and then I can define other things in terms of generic eval region. What's more, users can use this motion for precision evaluation using any evil motions

Right now, I have this snippet in my startup file:

```elisp
(defun spacemacs//emacs-lisp-eval-region-and-print (beg end)
  (interactive)
  (let* ((raw-result (with-output-to-string
                       (eval-region beg end standard-output)))
         (result (when (>= (length raw-result) 1)
                   (substring raw-result 1 -1))))
    (when result
      (cider--display-interactive-eval-result result end))))

(defvar spacemacs--generic-eval-region-alist
  '((clojure-mode . cider-eval-region)
    (emacs-lisp-mode . spacemacs//emacs-lisp-eval-region-and-print)
    (python-mode . python-shell-send-region)))

(defun spacemacs//get-eval-region-for-current-mode ()
  (interactive)
  (catch 'break
    (dolist (test spacemacs--generic-eval-region-alist)
      (let ((mode (car test))
            (val (cdr test)))
        (when (and (symbolp mode)
                   (derived-mode-p mode)
                   (symbolp val))
          (throw 'break val)))))) 

(defun spacemacs//generic-eval-region (beg end)
  (interactive)
  (let ((f (spacemacs//get-eval-region-for-current-mode)))
    (when (functionp f)
      (funcall f beg end))))

[...]

PR #10966, opened 2018-06-27
[WIP] Generic and Consistent eval/send across modes

emidln:
First, this is a Work In Progress PR intended to gauge whether this feature is still wanted and how I should go about it. This work started life based on #3103 and my gist in that thread. 

This provides two primary functions which abstract the idea of evaling and sending a region to a repl or process. This is done to deduplicate mode-specific eval and send functions. The idea is that instead of building a special `your-mode-eval-region`, `your-mode-eval-buffer`, `your-mode-eval-until-end-of-line`, `your-model-eval-defun`, we can simply implement a function like:

```elisp
(defun your-mode-eval-region (beg end)
   "Evals for your-mode"
   (interactive)
   .....)
```

And then hook it onto spacemacs--generic-eval-region-alist like so:

```elisp
(add-to-list 'spacemacs--generic-eval-region-alist '(your-mode . your-mode-eval-region))
```

In exchange, you get:

* An evil operator for evaling.
* Functions for evaling buffers, defuns, paragraphs, active marked regions, current line, etc.
* Optional highlighting and inline-results via eval-sexp-fu.
* Consistency behavior and capabilities across modes. 

This code makes a distinction between `eval`ing and `send`ing. Eval is when you can talk to a repl or process and get results back. Send is when you're limited to sending text to some other buffer which is hopefully running a REPL. 

This WIP is intended to discover:

* if this is wanted (I've been using my own generic-eval.el for several years now, so I know *I like it*.)
* how to break apart this into reviewable chunks
* what documentation needs to be created
* how badly written this code is (emacs lisp isn't something I use a ton)

So far, this WIP does not cover:

* mode/layer-specific modifications to use these new commands

My intention with not including changes to layers is to be able to shake out the usefulness of this to the community at large before beginning negotiations to rip out a ton of custom, mode-specific eval/send functions.

### DIFF
--- a/core/core-generic-ops.el
+++ b/core/core-generic-ops.el
@@ -1,0 +1,144 @@
+(require 'evil)
+
+;; Evaling is for modes who can send text to a repl and receive a response
+
+(defvar spacemacs--generic-eval-region-fallback-to-send t
+  "When truthy, fallback to using spacemacs--generic-send-region-alist if no eval func exists for mode")
+
+(defvar spacemacs--generic-eval-region-alist '()
+  "Alist mapping mode to a 2-arity function which evals from beg to end for the mode")
+
+(defun spacemacs//get-eval-region-for-current-mode ()
+  (interactive)
+  (catch 'break
+    (dolist (test spacemacs--generic-eval-region-alist)
+      (let ((mode (car test))
+            (val (cdr test)))
+        (when (and (symbolp mode)
+                   (derived-mode-p mode)
+                   (symbolp val))
+          (throw 'break val))))))
+
+
+(defun spacemacs//generic-eval-region (beg end)
+  (interactive)
+  (let* ((f (or (spacemacs//get-eval-region-for-current-mode)
+                (when spacemacs--generic-eval-region-fallback-to-send
+                  (spacemacs//get-send-region-for-current-mode)))))
+    (when (functionp f)
+      (funcall f beg end))))
+
+(defvar spacemacs--generic-eval-region-want-flash t)
+(when (and spacemacs--generic-eval-want-flash
+           (require 'eval-sexp-fu nil 'noerror))
+  (define-eval-sexp-fu-flash-command spacemacs//generic-eval-region
+    (eval-sexp-fu-flash (cons beg end))))
+
+(evil-define-operator spacemacs//generic-evil-eval-operator (beg end)
+  (spacemacs//generic-eval-region beg end))
+
+(defun spacemacs//generic-eval-line ()
+  "Evals the current line as though you did (kbd \"ma^ SPC meo $`a\")"
+  (interactive)
+  ;; we don't use execute-kbd-macro here, because popping the mark kills the overlay
+  (save-excursion
+    (let* ((beg (progn (beginning-of-line) (point)))
+           (end (progn (end-of-line) (point))))
+      (spacemacs//generic-eval-region beg end))))
+
+(defun spacemacs//generic-eval-paragraph ()
+  "Evals the current paragraph as though you did (kbd \"ma SPC meo ap`a\")"
+  (interactive)
+  ;; we don't use execute-kbd-macro here, because popping the mark kills the overlay
+  (save-excursion
+    (let* ((graph (evil-inner-paragraph))
+           (beg (first graph))
+           (end (second graph)))
+      (spacemacs//generic-eval-region beg end))))
+
+(defun spacemacs//generic-eval-active-region-or-paragraph ()
+  "Evals the current active region or paragraph"
+  (interactive)
+  (if (use-region-p)
+      (spacemacs//generic-eval-region (region-beginning)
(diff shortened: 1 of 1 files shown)


jb55:
Concept ACK 👍

timor:
The distinction between eval and send is a bit of a mystery to me.  I came across this when I started to work on factor-mode.  Is there a difference between the two from the user perspective?  I always thought that eval really means evaluating in the running emacs image, which obviously limits this to emacs-lisp code, and also is really the only one which can return anything, right?  For all other modes, aren't we "sending" text to a foreign repl one way or the other, and always doing that for the side effects, thus not caring about a return value?

As for having an abstraction for different modes, I like the idea very much.

tko:
While I have only a tiny bit first hand experience with implementing send-to-repl I believe conceptually this has potential to be a fairly good quality of life improvement. I can understand the unwillingness to go about rewriting _all_ the existing layers, but I would suggest considering rewriting _one_ to make it concretely clear how much simpler the layer specific code is when all the boring bits get delegated to the generic code.

core/core-generic-ops.el
```diff
@@ -0,0 +1,144 @@
+(require 'evil)
```
syl20bnr: While Spacemacs is based on evil we don't require it in `core` we can put the `evil` dependent part in an `with-eval-after-load`.

core/core-generic-ops.el
```diff
@@ -0,0 +1,144 @@
...
+;; Evaling is for modes who can send text to a repl and receive a response
+
+(defvar spacemacs--generic-eval-region-fallback-to-send t
+  "When truthy, fallback to using spacemacs--generic-send-region-alist if no eval func exists for mode")
+
+(defvar spacemacs--generic-eval-region-alist '()
+  "Alist mapping mode to a 2-arity function which evals from beg to end for the mode")
+
+(defun spacemacs//get-eval-region-for-current-mode ()
+  (interactive)
```
syl20bnr: Interactive function should use `/` instead of `//` in their names. `//` is for implementation details functions.
Also we try to document all interactive function with a docstring ;-)

core/core-generic-ops.el
```diff
@@ -0,0 +1,144 @@
...
+
+(defun spacemacs//generic-eval-region (beg end)
+  (interactive)
+  (let* ((f (or (spacemacs//get-eval-region-for-current-mode)
+                (when spacemacs--generic-eval-region-fallback-to-send
+                  (spacemacs//get-send-region-for-current-mode)))))
+    (when (functionp f)
+      (funcall f beg end))))
+
+(defvar spacemacs--generic-eval-region-want-flash t)
```
syl20bnr: Regroup all the `defvar` of this file at the top of the file.

syl20bnr:
> if this is wanted

yes it is :-)

Would be interesting to include somehow the key bindings of the conventions as well so we don't duplicate them all over the layers with the risk to loose consistency: https://github.com/syl20bnr/spacemacs/blob/develop/doc/CONVENTIONS.org#L159-L195

syl20bnr:
Thinking about it, this could be moved to a layer `spacemacs-repl` with a local package and we can use layer dependencies to bring it automatically when needed, i.e. Python layer `layers.el` file would declare `spacemacs-repl` as a dependency:

```elisp
(configuration-layer/declare-layers '(spacemacs-repl))
```

And then we use post-init functions to setup the `alist` variables correctly. Actually it would be easy for you to create a stand-alone package for this. Maybe a package like this already exists ? Maybe not we all the feature you have...

rgemulla:
This is interesting and potentially very useful. 

I had [developed a macro](https://github.com/rgemulla/spacemacs-layers/blob/master/%2Blang/pythonp/README.org) which automatically creates a number of commands that send things to the repl for Python, which I use in everyday work with Spacemacs. Perhaps this can serve as an inspiration for additional features / usecases. 

Example bindings below (similar bindings are created for sending regions, code blocks, statements, ...).

| ,ef | Evaluate current function and keep the point where it is                                                  |
| ,eF | Evaluate current function, keep the point where it is, and switch focus to the Python shell               |
| ,sf | Evaluate current function and move point to first subsequent statement                                    |
| ,sF | Evaluate current function, move point to first subsequent statement, and switch focus to the Python shell |

| ,td | Toggle whether to show the shell buffer after sending something to it       |
| ,ti | Toggle whether to show code fragments sent to the shell in the shell buffer |
| ,to | Toggle whether to show shell outputs in the echo area                       |

smile13241324:
This could still be useful, though I think we will need to invest some effort into it before this will be fully working.
Right now I am not yet sure how evaluation in non lisp languages should work, basically the repl process is pretty dump but if it is doable we should give it a try.

smile13241324:
After carefully thinking about this I think this is far too old to be useful anymore. If this is desirable we should make a complete redesign instead.

PR #10976, opened 2018-06-30
Fix handling of keymap property when it's symbol valued

rgrinberg:
The previous code would only handle the case when the property was a keymap. But
it can also be a symbol representing a keymap.

### DIFF
--- a/layers/+distributions/spacemacs-bootstrap/local/evil-evilified-state/evil-evilified-state.el
+++ b/layers/+distributions/spacemacs-bootstrap/local/evil-evilified-state/evil-evilified-state.el
@@ -80,7 +80,10 @@
 Needed to bypass keymaps set as text properties."
   (unless (bound-and-true-p isearch-mode)
     (when (memq evil-state '(evilified visual))
-      (let* ((map (get-char-property (point) 'keymap))
+      (let* ((map-or-symbol (get-char-property (point) 'keymap))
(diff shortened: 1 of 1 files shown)


sdwolfz:
Could you please give an error example or steps to reproduce this. I would like to try it out locally before merging it.

rgrinberg:
Codruț Constantin Gușoi <notifications@github.com> writes:

> Could you please give an error example or steps to reproduce this. I would like to try it out locally before merging it.

Sure. In my config, I get a symbol (compilation-button-map) when evaluating:

(get-char-property (point) 'keymap)

When the cursor is on top of some grep result.

Checking the documentation here:
https://www.gnu.org/software/emacs/manual/html_node/elisp/Special-Properties.html
I confirmed that symbol's are indeed valid values for this property.

Without my fix, I get the following error:

Error in pre-command-hook (evilified-state--pre-command-hook): (wrong-type-argument listp compilation-button-map)

If you'd like to reproduce this, try adding this:

(with-eval-after-load 'ivy
  (evilified-state-evilify-map ivy-occur-grep-mode-map
    :mode ivy-occur-grep-mode
    :bindings
    "dd" #'ivy-occur-delete-candidate
    "ga" #'ivy-occur-read-action
    "gr" #'ivy-occur-revert-buffer))

This should trigger the evilified check induce the error.

sdwolfz:
Thank you :heart:!

PR #10978, opened 2018-06-30
Change order of evaluating template directories

fuzzycode:
The user defined template directory was never taken into account even though it was set and the default directory was always used.


sdwolfz:
Closing this as it duplicates #10978. But thank for bringing this up, let's continue the discussion there.

PR #10982, opened 2018-07-02
Use fundanmental-mode to open minified file

ztlevi:
This commit aims to solve the freezing problem when Emacs editing minified files. It's very simple by just checking the first line's length.

And it's disabled for `text-mode` files.

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -451,7 +451,7 @@ FILENAME is deleted using `spacemacs/delete-file' function.."
   (and
    (not (member (file-name-extension (buffer-file-name))
                 '("org" "md" "markdown" "txt" "rtf")))
-   (> (spacemacs/get-nth-line-length 1) 500)))
+   (> (spacemacs/get-nth-line-length 1) 1000)))
 
 (defun spacemacs/delete-window (&optional arg)
   "Delete the current window.
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -451,7 +451,8 @@ FILENAME is deleted using `spacemacs/delete-file' function.."
   (and
    (not (member (file-name-extension (buffer-file-name))
                 '("org" "md" "markdown" "txt" "rtf")))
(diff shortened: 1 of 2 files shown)


layers/+spacemacs/spacemacs-defaults/funcs.el
```diff
@@ -436,6 +436,23 @@ FILENAME is deleted using `spacemacs/delete-file' function.."
...
+        (- (line-end-position)
+           (line-beginning-position)))))
+
+;; check if the file has been minified
+;; disable for text-mode files
+(defun spacemacs/check-minified-file ()
+  (and
+   (not (member (file-name-extension (buffer-file-name))
+                '("org" "md" "markdown" "txt" "rtf")))
+   (> (spacemacs/get-nth-line-length 1) 500)))
```
syl20bnr: Can we up this number a bit? 500 seems low to me.

ztlevi: How much do you want?

ztlevi:
I made an update of the function see [this line](https://github.com/syl20bnr/spacemacs/pull/10982/commits/7cdce32fde9f4f82579bdf398ba791e7dde4ed33#diff-8a41e260024d225e124323deab81aaf3R455). But I'm not sure how can we break the loop as soon as we find a long line and just return `t`.

syl20bnr:
Performance wise this new function is still fine ? Do you know the complexity of the function `line-end-position` ?

ztlevi:
Hi, I pushed a new commit and the modified function checks the first 20 lines. This prevents minified files with comments in the top get passed.

For the question, it's a built-in function. See:

```
line-end-position is a built-in function in ‘src/editfns.c’.

(line-end-position &optional N)

Return the character position of the last character on the current line.
With argument N not nil or 1, move forward N - 1 lines first.
If scan reaches end of buffer, return that position.

This function ignores text display directionality; it returns the
position of the last character in logical order, i.e. the largest
character position on the line.

This function constrains the returned position to the current field
unless that would be on a different line than the original,
unconstrained result.  If N is nil or 1, and a rear-sticky field ends
at point, the scan stops as soon as it starts.  To ignore field
boundaries bind ‘inhibit-field-text-motion’ to t.

This function does not move point.
```

As for the performance, I think if we put the collector in the `spacemacs/get-nth-line-length`, and return collected lines length, it should be better.

PR #10985, opened 2018-07-02
org: add warning against manually loading org

smile13241324:
Added a warning to the Readme.md of "org" to make more clear that for working with it, the layer should be installed instead of the package be manually loaded.

I have also tried to give a hint to affected users to install the layer when they encounter various missing key bindings while using "org"


sdwolfz:
Thank you :heart:!

PR #11012, opened 2018-07-08
Add go-benchmark function, binding and documentation.

joshua-mcintosh:
(Fair warning, I'm an emacs newbie/convert from vim, so this is more or less the first emacs func I've written.)

Adds a function / keybinding for "go test -bench ." with optional
numeric argument for indicating a count.

### DIFF
--- a/layers/+lang/go/packages.el
+++ b/layers/+lang/go/packages.el
@@ -129,6 +129,7 @@
         "ia" 'go-import-add
         "ig" 'go-goto-imports
         "ir" 'go-remove-unused-imports
+        "tb" 'spacemacs/go-run-package-benchmarks
         "tP" 'spacemacs/go-run-package-tests-nested
         "tp" 'spacemacs/go-run-package-tests
         "ts" 'spacemacs/go-run-test-current-suite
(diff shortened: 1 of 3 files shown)


sdwolfz:
@ljupchokotev may I ask you to provide your opinion on this PR as well?

ljupchokotev:
@sdwolfz 
I tested this locally on a simple project with 1 benchmark defined, running the benchmarks, with or without the numeric argument, works well. It's fine to be merged as it is.

layers/+lang/go/packages.el
```diff
@@ -129,6 +129,7 @@
         "ia" 'go-import-add
         "ig" 'go-goto-imports
         "ir" 'go-remove-unused-imports
+        "tb" 'spacemacs/go-run-package-benchmarks
```
sdwolfz: The `CONVENTIONS.org` file requires that `tb` is reserved for running all the tests in the buffer. For benchmarks a new keybinding must be used. Please look at how other layers define a similar functionality.

duianto:
@joshua-mcintosh 
The key binding needs to be changed.
And there's a conflict in `README.org`.

Issue #6499, opened 2016-07-05
error: spacemacs/search-project-auto-region-or-symbol


pA0k:
#### Reproduction guide :beetle:
- Start Emacs
- M-m *

_Observed behaviour:_ :eyes: :broken_heart:

_Expected behaviour:_ :heart: :smile:
#### System Info :computer:
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.21
- Spacemacs branch: develop (rev. a650877)
- Graphic display: t
- Distribution: spacemacs
- Editing style: emacs
- Completion: ivy
- Layers:

``` elisp
(ivy auto-completion better-defaults shell-scripts emacs-lisp
     (python :variables python-test-runner 'pytest python-enable-yapf-format-on-save t)
     (ruby :variables ruby-enable-enh-ruby-mode t ruby-test-runner 'rspec)
     git html markdown org
     (shell :variables shell-default-height 30 shell-default-position 'bottom shell-protect-eshell-prompt t)
     (spell-checking :variables spell-checking-enable-by-default nil)
     syntax-checking version-control chrome)
```
#### Backtrace :paw_prints:

```
```

I got an error message.

```
rxt-string->pcre: Wrong type argument: arrayp, nil
```

pA0k:
I found out why error occured.
It happend running command at blank line.

d12frosted:
@pA0k thanks for the report. Didn't see such behaviour before but I can confirm it.

Miciah:
#11013 should prevent the error.

PR #11013, opened 2018-07-08
ivy: Fix error on nil input with counsel search

Miciah:
### ivy: Fix error on nil input with counsel search

Fix issue #6499: `SPC m` (`spacemacs/search-project-auto-region-or-symbol`) on a blank line causes an error:

    rxt-string->pcre: Wrong type argument: arrayp, nil

* `layers/+completion/ivy/funcs.el` (`spacemacs/counsel-search`): Check whether `initial-input` is `nil` in order to avoid calling `rxt-quote-pcre` on a nil value.


sdwolfz:
Thank you :heart:!

PR #11016, opened 2018-07-09
scala/java: Remove broken ensime key bindings and fix yanking types

bjarkevad:
A bunch of functions were removed from https://github.com/ensime/ensime-emacs/ but Spacemacs still has keybindings for those. This PR removes those. 
Included is a fix to `spacemacs/ensime-yank-type-at-point`.

### DIFF
--- a/layers/+lang/java/funcs.el
+++ b/layers/+lang/java/funcs.el
@@ -169,12 +169,12 @@
 (defun spacemacs/ensime-yank-type-at-point ()
   "Yank to kill ring and print short type name at point to the minibuffer."
   (interactive)
-  (ensime-type-at-point t nil))
+  (ensime-type-at-point '(4)))
 
 (defun spacemacs/ensime-yank-type-at-point-full-name ()
   "Yank to kill ring and print full type name at point to the minibuffer."
   (interactive)
-  (ensime-type-at-point t t))
+  (ensime-type-at-point '(4) t))
 
 
 ;; eclim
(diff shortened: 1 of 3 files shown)


layers/+lang/java/funcs.el
```diff
@@ -169,12 +169,12 @@
 (defun spacemacs/ensime-yank-type-at-point ()
   "Yank to kill ring and print short type name at point to the minibuffer."
   (interactive)
-  (ensime-type-at-point t nil))
+  (ensime-type-at-point '(4)))
```
sdwolfz: This is... weird, but it's present in the definition of `ensime-type-at-point`. I'm not sure what it's supposed to mean.

layers/+lang/java/packages.el
```diff
@@ -198,7 +198,6 @@
           "br"     'ensime-sbt-do-run
 
           "ct"     'ensime-typecheck-current-buffer
-          "cT"     'ensime-typecheck-all
```
sdwolfz: Are these also removed from the stable release? The recommendation is to pin ensime to stable melpa and the functionality might still be there... I'm not sure how to handle this, or even if we should handle it.

layers/+lang/java/funcs.el
bjarkevad: `'(4)` is basically the universal arg `C-u`. You could probably do something like `(setq current-prefix-arg '(4))`, but I thought it was clearer to just pass it along.

layers/+lang/java/packages.el
bjarkevad: Yep, there are all removed from the package on stable melpa.

bjarkevad:
I meant to remove the docs, thanks for reminding me! Updated PR.

sdwolfz:
Thank you :heart:!

PR #11017, opened 2018-07-09
Fix the projectile layer

FieryCod:
The projectile layer is broken with the newest projectile-mode. This PR addresses the issue with projectile-global-mode variable not found.


FieryCod:
Thank you! That is the honour for me to help the spacemacs/emacs community!

PR #11018, opened 2018-07-10
Fix and rename column-indexing toggle

Miciah:
The column-indexing toggle did not have a `:status` property.  As a result, the toggle command would enable the toggle but would not disable it.  In addition, the absence of `:status` prevented `-off` and `-on` commands from being defined.

This commit adds a `:status` property so that the toggle works properly and `-off` and `-on` commands are defined.

This commit also renames the toggle to `zero-based-column-indexing` in order to make its purpose clearer.

* `layers/+spacemacs/spacemacs-defaults/keybindings.el` (`column-indexing`): Add `:status` and rename to `zero-based-column-indexing`.

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/keybindings.el
+++ b/layers/+spacemacs/spacemacs-defaults/keybindings.el
@@ -366,10 +366,11 @@
   :mode font-lock-mode
   :documentation "Toggle syntax highlighting."
   :evil-leader "ths")
-(spacemacs|add-toggle column-indexing
-  :documentation "Toggle column indexing starting at 1."
-  :on (setq column-number-indicator-zero-based nil)
-  :off (setq column-number-indicator-zero-based t)
+(spacemacs|add-toggle zero-based-column-indexing
+  :documentation "Toggle column indexing starting at 0 versus 1."
+  :status (bound-and-true-p column-number-indicator-zero-based)
+  :on (setq column-number-indicator-zero-based t)
+  :off (setq column-number-indicator-zero-based nil)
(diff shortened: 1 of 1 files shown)


Miciah:
On a related topic, is it too late to rename `column-indexing` to `positive-column-indexing` or similar? The toggle's doc text reads, "Toggle column-indexing on and off", and the toggle prints "column-indexing disabled" and "column-indexing enabled", which is confusing.

Miciah:
(Oh, I made a funny typo in the branch name!)

sdwolfz:
:fox_face: 

It's not to late, I did not know what to put there so I just went with `column-indexing`. But I don't like `positive-column-indexing` since positive does not really suggest this. Maybe it's better to go with the spirit of the `column-number-indicator-zero-based` variable and call it `zero-based-indexing`, inverting the values for on/off, what do you think?

Miciah:
Yeah, I think making a toggle's name consistent with the underling variable's name is a good guideline.  Is `zero-based-column-indexing` all right?

sdwolfz:
Yes `zero-based-column-indexing` sounds perfect!

Miciah:
Updated.

sdwolfz:
Thank you :heart:!
I added some more documentation and values for `:on-message`/`:off-message` to clarify what's going on when you press `SPC t z` in e2c02505488496f6e08f215135ef1ba9ae2e77c6:
```
Column indexing starts at 0 (current column is 64)
Column indexing starts at 1 (current column is 65)
```

Issue #6416, opened 2016-06-24
"helm-file-smart-do-search" complains "Wrong type argument: stringp, nil"


AlohaWorld:
#### Description

When invoke "helm-file-smart-do-search" with a current unsaved buffer, emacs will complain
"helm-basename: Wrong type argument: stringp, nil"
#### Reproduction guide
- Start Emacs
- M-x list-packages
- M-m s f
  Error messages appear.
#### System Info
- OS: windows-nt
- Emacs: 24.5.1
- Spacemacs: 0.105.21
- Spacemacs branch: cyd (rev. 0283f64)
- Graphic display: t
- Distribution: spacemacs
- Editing style: emacs
- Completion: helm
- Layers:

``` elisp
((auto-completion :variables auto-completion-return-key-behavior 'complete auto-completion-tab-key-behavior 'cycle auto-completion-enable-snippets-in-popup t auto-completion-complete-with-key-sequence nil auto-completion-complete-with-key-sequence-delay 0.1 auto-completion-private-snippets-directory
                  (concat
                   (car dotspacemacs-configuration-layer-path)
                   "snippets/")
                  auto-completion-enable-sort-by-usage t)
 better-defaults emacs-lisp git org
 (shell :variables shell-default-height 30 shell-default-position 'bottom)
 spell-checking syntax-checking themes-megapack chinese my-commons my-cygwin my-buffer)
```

TheBB:
Can you upgrade your packages and provide a backtrace (`SPC t D`)

AlohaWorld:
Upgraded all the packages.

**System Info**
#### System Info
- OS: windows-nt
- Emacs: 24.5.1
- Spacemacs: 0.105.21
- Spacemacs branch: cyd (rev. 0283f64)
- Graphic display: t
- Distribution: spacemacs
- Editing style: emacs
- Completion: helm
- Layers:
```elisp
((auto-completion :variables auto-completion-return-key-behavior 'complete auto-completion-tab-key-behavior 'cycle auto-completion-enable-snippets-in-popup t auto-completion-complete-with-key-sequence nil auto-completion-complete-with-key-sequence-delay 0.1 auto-completion-private-snippets-directory
                  (concat
                   (car dotspacemacs-configuration-layer-path)
                   "snippets/")
                  auto-completion-enable-sort-by-usage t)
 better-defaults emacs-lisp git org
 (shell :variables shell-default-height 30 shell-default-position 'bottom)
[...]

PR #11020, opened 2018-07-11
helm: Fix error from SPC s f on non-file buffers

Miciah:
Fix issue #6416: `SPC s f` (`spacemacs/helm-files-smart-do-search`) in a buffer that is not visiting a file causes an error:

    Wrong type argument: stringp, nil

* `layers/+completion/helm/funcs.el` (`spacemacs//helm-do-grep-region-or-symbol`): Check whether `preselection` is `nil` in order to avoid calling `helm-basename` on a nil value.


sdwolfz:
Thank you :heart:!

PR #11025, opened 2018-07-12
Fix key bindings for wgrep

jjlee:
The problem was not with the existing bindings, but rather:

1. `C-x C-e` (`spacemacs//counsel-edit` -- from an ivy completion session) would
land you in an `ivy-occur` buffer with wgrep enabled but in motion state.  This
meant the spacemacs key bindings for wgrep were not available.  In particular,
the usual routine is:

* (enter ivy completion session, e.g. with `SPC s a P`)
* `C-c C-e`
* edit the buffer as needed
* `, ,` (`wgrep-finish-edit`)
* `, s` (`wgrep-save-all-buffers`)

2. That last binding (`, s`) was not there.

One issue is that `, s` is available before `wgrep-finish-edit` is called, which is
confusing because I don't think it has any useful function at that point in
time.

### DIFF
--- a/layers/+completion/ivy/funcs.el
+++ b/layers/+completion/ivy/funcs.el
@@ -99,7 +99,7 @@
 (defun spacemacs//counsel-edit ()
   "Edit the current search results in a buffer using wgrep."
   (interactive)
-  (run-with-idle-timer 0 nil 'ivy-wgrep-change-to-wgrep-mode)
+  (run-with-idle-timer 0 nil 'spacemacs/ivy-wgrep-change-to-wgrep-mode)
   (ivy-occur))
 
 (defun spacemacs//gne-init-counsel ()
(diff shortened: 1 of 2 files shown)


Compro-Prasad:
> One issue is that , s is available before wgrep-finish-edit is called, which is
> confusing because I don't think it has any useful function at that point in
> time.

I think its fine if you want to save without leaving the edit state.

So, I think this is ready?

Compro-Prasad:
I don't use ivy in Spacemacs. @sdwolfz Can you see if this PR works?

sdwolfz:
Thank you :heart:! 

I made a small change, see: 9d2a1085eed17d18b17df89cb9aeeb9c5c937794. The new `'spacemacs/ivy-wgrep-change-to-wgrep-mode` function needs to be marked as `(interactive)` otherwise you can not call it after the last step you described. There might be a more subtle underlying issue that's causing all the trouble but this fix is OK for now. If there are other problems in this are in the future then maybe it will require a more in-depth analysis.

PR #11026, opened 2018-07-12
c-c++: Update SPC g Y command

Miciah:
Upstream renamed `rtags-cycle-overlays-on-screen` to `rtags-cycle-through-diagnostics`.  As a result, `SPC m g Y` causes the following error:

    Wrong type argument: commandp, rtags-cycle-overlays-on-screen

This commit updates the corresponding key binding to the new command name in order to prevent the error.

https://github.com/Andersbakken/rtags/commit/3011f4e787775497d482b16d2f51fb98a5fb3ddc

* `layers/+lang/c-c++/packages.el` (`c-c++/init-rtags`): Change `rtags-cycle-overlays-on-screen` to `rtags-cycle-through-diagnostics`.


Compro-Prasad:
I am not using rtags but this seems an important fix for those who are using rtags. @bmag look into this.

PR #11027, opened 2018-07-13
repair search highlighting (spacemacs/counsel-search)

ivasonn:
Repair search highlighting (spacemacs/counsel-search) with ivy--regex-plus by replaced it with spacemacs/ivy--regex-plus

bug with search highlighting when using extra arguments for searching using ivy--regex-plus

after fixing

### DIFF
--- a/layers/+completion/ivy/README.org
+++ b/layers/+completion/ivy/README.org
@@ -47,6 +47,10 @@ You can customize ivy with the following variables:
 - =ivy-height= The height of the minibuffer. The Spacemacs default is 15.
 - =ivy-use-selectable-prompt= When non-nil, make the prompt line selectable like
   a candidate. The Spacemacs default value is =t=.
+- =ivy-re-builders-alist= An alist of regex building functions for each collection function.
+  See ivy documentation for possible choices.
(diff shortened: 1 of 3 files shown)


ivasonn:
With current `spacemacs/counsel-search` we can't write "foo -- bar" to search this phrase, because it will translated to `foo "bar" .` and ag (for example) will try to search foo in bar directory. if user search for `-Glang -- "foo -- bar"` the end of ag command will be right. If user want to search `"foo -- bar"` he should use `foo.--.bar` or `foo \-\- bar`. 
I know about this bug. But i don't have enough experience in elisp to fix it correctly. Sorry.

bmag:
Thanks, I'll fixup the documentation a bit and merge today or tomorrow.

ivasonn:
Thanks to you too

bmag:

I squashed your commits into a single commit and re-worded the commit message into a valid form - see https://github.com/syl20bnr/spacemacs/commit/d2dab66724711aba77454f677771b59566bd6208.
I fixed the docs in https://github.com/syl20bnr/spacemacs/commit/e109777f8b8c5e9b9a697ad1004c1521b09ccd68.

PR #11030, opened 2018-07-13
Added support to customise bold headings of org levels and priority

is-r-not-j:
Spacemacs theme makes headings of org items level 1 and 2 inherit bold weight without apparent way to customise it.
The same goes for agenda (priority) items, which are bold.

This commit adds two custom variables:

- spacemacs-theme-org-bold
	controls if level 1 and 2 items should be bold or not
	default: t

- spacemacs-theme-org-priority-bold
	controls if items in the agenda should be bold
	default: t

### DIFF
--- a/core/libs/spacemacs-theme/spacemacs-common.el
+++ b/core/libs/spacemacs-theme/spacemacs-common.el
@@ -70,6 +70,16 @@ to 'auto, tags may not be properly aligned. "
   :type 'boolean
   :group 'spacemacs-theme)
 
+(defcustom spacemacs-theme-org-bold t
+  "Inherit text bold for org headings"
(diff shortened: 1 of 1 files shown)


bmag:
Hey, thanks for the effort, but this should be proposed upstream at https://github.com/nashamri/spacemacs-theme. The maintainer is pretty cool and your changes seem good, so I'm pretty sure it would be easy to merge this upstream.

We ship a copy of spacemacs-theme because it's the default theme and we want it to be available even if the user can't access the internet to install themes, but we don't really want to diverge from upstream, so I'm closing this PR. Our copy is updated periodically with upstream changes, so after the changes are made upstream they will eventually arrive here as well.

PR #11031, opened 2018-07-13
Update obsolete `cider-refresh` alias to `cider-ns-refresh`

Andre0991:
`cider-refresh` is an obsolete alias:

https://github.com/clojure-emacs/cider/blob/8b8309d9702a98dbfb1804de3d643b36b188057b/cider-ns.el#L204


FieryCod:
@syl20bnr May we please merge this one?

Issue #9271, opened 2017-07-23
Continuosly getting ‘invalid face reference: t‘ message in the message buffer


yalu:
#### Description :octocat:
Continuosly getting ‘invalid face reference: t‘ message in the message buffer

#### Reproduction guide :beetle:
- Start Emacs with clojure layer turned on and (setq clojure-enable-fancify-symbols t) in spacemacs configuration.
- open any clj or cljs file
- switch to message buffer or have both buffers open side-by-side to observe the messages.

*Observed behaviour:* :eyes: :broken_heart:
Continuosly getting ‘invalid face reference: t‘ message in the message buffer

*Expected behaviour:* :heart: :smile:
No such messages

If I disable ‘clojure-enable-fancify-symbols messages go away.

#### System Info :computer:
- OS: darwin
- Emacs: 25.2.1
- Spacemacs: 0.200.9
- Spacemacs branch: nil (rev. 9825606c)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: helm
- Layers:
```elisp
(nginx php sql
       (auto-completion :variables auto-completion-return-key-behavior nil auto-completion-tab-key-behavior 'cycle :disabled-for org erc)
       (org :variables org-enable-github-support t)
       (shell :variables shell-default-shell 'eshell shell-default-height 30 shell-default-position 'bottom)
       (spell-checking :variables spell-checking-enable-by-default nil)
       (version-control :variables version-control-diff-tool 'diff-hl)
       (evil-snipe :variables evil-snipe-enable-alternate-f-and-t-behaviors t)
       (typography :variables typography-enable-typographic-editing t)
       (colors :variables colors-enable-rainbow-identifiers nil colors-enable-nyan-cat-progress-bar
               (display-graphic-p))
       better-defaults ranger emacs-lisp common-lisp erc git markdown syntax-checking evil-cleverparens vinegar extra-langs semantic tmux gtags fasd clojure parinfer html java
       (javascript :variables javascript-disable-tern-port-files nil)
       react python ruby shell-scripts yaml speedtype theming)
```
- System configuration features: JPEG RSVG IMAGEMAGICK NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS MODULES

neolee:
Same here. Toggling on clojure-enable-fancify-symbols causes continuously getting ‘invalid face reference: t‘ message in the message buffer.

[...]

PR #11032, opened 2018-07-14
clojure: Fix face name of fancified symbols

Miciah:
### clojure: Fix face name of fancified symbols

Fix issue #9271: Setting `clojure-enable-fancify-symbols` to `t` causes the following to the be printed repeatedly in the message buffer:

    invalid face reference: t

The problem is that the `clojure/fancify-symbols` function adds font-lock keywords with highlight forms containing face-name expressions that evaluate to the return value of `compose-region`, which apparently is `t`; this value is interpreted as a face name.

The solution to this problem is that the face-name expressions should evaluate to `nil`, which will be interpreted as an empty list of properties instead of a face name.

* `layers/+lang/clojure/funcs.el` (`clojure/fancify-symbols`): Define keywords using face-name expressions that evaluate to `nil`.

### clojure: Delete bogus escaping in font-lock regexps

* `layers/+lang/clojure/funcs.el` (`clojure/fancify-symbols`): Delete bogus backslashes.

### DIFF
--- a/layers/+lang/clojure/funcs.el
+++ b/layers/+lang/clojure/funcs.el
@@ -13,15 +13,15 @@
   "Pretty symbols for Clojure's anonymous functions and sets,
    like (λ [a] (+ a 5)), ƒ(+ % 5), and ∈{2 4 6}."
   (font-lock-add-keywords mode
-    `(("(\\(fn\\)[\[[:space:]]"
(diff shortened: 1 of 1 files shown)


syl20bnr:
Thank you ! 👍 

PR #11034, opened 2018-07-14
Fix maximizing/restoring a window when neotree is active.

### LINKED ISSUE #8107: Window maximize bug with Neotree
#### Description :octocat:
Window mazimize doesn't toggle windows back to their previous state before being maximized when there is a side window with NeoTree.
- Start Emacs
- Open a file
- Split window (either with SPACE w / or SPACE w -) 
- SPACE f t (to open the file tree)
- SPACE 1 (to go to the first window)
- SPACE w m (to maximize first window)
- SPACE w m (to toggle back to split windows)
One window remains in maximized state along with the file tree side window.
Two windows should appear as split windows along with the Neotree side window.  The windows should toggle back to their previous state before being maximized.
#### System Info :computer:
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.200.5
- Spacemacs branch: master (rev. 664ba6a)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(python markdown ranger c-c++ auto-completion ycmd evil-commentary
        (c-c++ :variables c-c++-enable-clang-support t)
        emacs-lisp git org ibuffer pdf-tools tmux
        (shell :variables shell-default-shell 'term shell-default-height 40 shell-default-position 'bottom)
        version-control themes-megapack)
```

#### Backtrace :paw_prints:
```
<<BACKTRACE IF RELEVANT>>
```

simon-bourne:
This fixes #8107. It's worth noting this fix only has an effect in emacs >= 26, although it doesn't break anything in earlier versions. In emacs < 26, maximize buffer closes the neotree window. Looks like this is because `no-delete-other-windows` is not available in emacs < 26.

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -102,16 +102,19 @@ automatically applied to."
           (message "Indented buffer.")))
       (whitespace-cleanup))))
 
-;; from https://gist.github.com/3402786
 (defun spacemacs/toggle-maximize-buffer ()
   "Maximize buffer"
   (interactive)
-  (if (and (= 1 (length (window-list)))
-           (assoc ?_ register-alist))
-      (jump-to-register ?_)
+  (let ((starting-layout (current-window-configuration))
+        (window-count (length (window-list))))
     (progn
-      (window-configuration-to-register ?_)
-      (delete-other-windows))))
+      (funcall spacemacs-window-split-delete-function)
+      ;; Were we already maximized?
+      ;; We use window count in case window order has changed
+      (if (= window-count (length (window-list)))
+          (let ((saved-layout (frame-parameter nil 'spacemacs--saved-layout)))
+            (when saved-layout (set-window-configuration saved-layout)))
+        (set-frame-parameter nil 'spacemacs--saved-layout starting-layout)))))
 
 ;; https://tsdh.wordpress.com/2007/03/28/deleting-windows-vertically-or-horizontally/
 (defun spacemacs/maximize-horizontally ()
@@ -1459,4 +1462,3 @@ Decision is based on `dotspacemacs-line-numbers'."
           enabled-for-parent            ; mode is one of default allowed modes
           disabled-for-modes
           (not disabled-for-parent)))))
-
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -65,6 +65,7 @@ This file containes the change log for the next major version of Spacemacs.
   the =*Messages*= buffer in another window (thanks to deb0ch)
 *** Core changes
 - 39667d3 | * Don't suggest `SPC q r` if restart-emacs is missing (Keshav Kini)
+- toggle-maximize-window now works with neotree, treemacs and frames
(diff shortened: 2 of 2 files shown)


layers/+spacemacs/spacemacs-defaults/funcs.el
```diff
@@ -106,9 +106,14 @@ automatically applied to."
...
   (interactive)
-  (if (and (= 1 (length (window-list)))
-           (assoc ?_ register-alist))
-      (jump-to-register ?_)
+  (if (let ((window-count (length (window-list))))
+        (and
+         (or ( = 1 window-count)
+             (and
+              ( = 2 window-count)
+              (neo-global--window-exists-p)))
```
syl20bnr: neotree is not always installed so we cannot expect to have this function always defined.

sdwolfz:
@simon-bourne were you able to follow up on the feedback provided above?

simon-bourne:
I'm currently looking at whether there's a more general way to detect if we're in the only non side window. Progress will be a bit slow as I'm learning elisp/emacs and don't have a lot of time to devote to this.

simon-bourne:
I've re done this to use `spacemacs-window-split-delete-function` and correctly handle frames. It also works with treemacs and neotree.

duianto:
Confirmed this PR solves the issue:
Window maximize bug with Neotree #8107

Since this PR was posted. The following PR:
Keeps previous position on toggle-maximize-buffer #11480
has been applied.

It wraps the toggle maximize buffer expressions with a `save-excursion`.

But that issue doesn't occur with this PR, because the dosctring for `current-window-configuration` says:
>This describes the number of windows, their sizes and current buffers,
and for each displayed buffer, where display starts, and the position of
point.  An exception is made for point in the current buffer:
its value is -not- saved.

Specifically the point isn't saved in the current buffer. Therefore it can't be restored incorrectly, and the `save-excursion` command isn't needed.

#### Frame issue
Is this the frame issue?

Without this PR applied.
- Split the current frame `SPC w s`
- Maximize the selected window `SPC w m`
- Create a new frame `SPC F n`
- Toggle maximize buffer `SPC w m`

#### Observed behavior
The other frame was unmaximized.

It doesn't occur with this PR applied.

#### Possible changes to this PR.

#### Docstring
The docstring would be more descriptive if it was changed:
From: `Maximize buffer`
To: `Toggle Un/Maximize buffer.`

#### Remove progn
The `progn` isn't needed, because the expressions in a `let` are called sequentially.

#### Commit subject, body message

I made the commit subject shorter (The suggestion in a lot of places seem to be <=50 chars).
And I added two lines describing what the commit fixes.

```
toggle-maximize-window support filetrees and frames

Enable unmaximizing when a filetree (neotree or treemacs) is open.
Only unmaximize the current frame.
```

#### Emacs <26

If there are some issues with `<26` versions, could one solution be to use the current version for `<26` and this PRs solution for `>=26`?

jeremyjh:
@simon-bourne Your patch fixes this issue for me, which I found very frustrating. Do you have plans to revisit this and get it merged?

PR #11035, opened 2018-07-14
Fix counsel-search

et2010:
With recent ivy updates, counsel-search is broken. This should fix it.

### DIFF
--- a/layers/+completion/ivy/funcs.el
+++ b/layers/+completion/ivy/funcs.el
@@ -71,7 +71,7 @@
       (lambda (string &optional _pred &rest _unused)
         "Grep in the current directory for STRING."
         (if (< (length string) 3)
-            (counsel-more-chars 3)
+            (counsel-more-chars)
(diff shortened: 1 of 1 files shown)


layers/+completion/ivy/funcs.el
```diff
@@ -71,7 +71,7 @@
       (lambda (string &optional _pred &rest _unused)
         "Grep in the current directory for STRING."
         (if (< (length string) 3)
```
whatacold: I think it would be better to remove the condition of `if` too.

et2010: I don't think so. There is an "else" part in the conditional unless we don't need that also.

bmag:
Thank you, I chose a slightly different fix - see https://github.com/syl20bnr/spacemacs/commit/e63a9c05bd8e6c1bc905e4c28ceb5aa0d48d5b47

Issue #9070, opened 2017-06-11
spacemacs/python-execute-file-focus don't change focus


Voleking:
#### Description :octocat:
`spacemacs/python-execute-file-focus` don't change focus

#### Reproduction guide :beetle:
- Start Emacs
- Open a python file
- run `spacemacs/python-execute-file-focus` function (,-c-C)

*Observed behaviour:* :eyes: :broken_heart:
didn't switch to the shell buffer

*Expected behaviour:* :heart: :smile:
switch to the shell buffer in `insert state'

#### System Info :computer:
- OS: darwin
- Emacs: 25.2.1
- Spacemacs: 0.200.9
- Spacemacs branch: develop (rev. 87545755)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(helm osx
      (chinese :variables chinese-enable-fcitx t)
      dash git github pandoc
      (c-c   :variables c-c  -default-mode-for-headers 'c  -mode c-c  -enable-clang-support nil)
      emacs-lisp html java
      (markdown :variables markdown-command "pandoc -t html5 -f markdown -s -S --mathjax --highlight-style=pygments --toc --toc-depth 3 --template github.html5 --css html/css/github.css" markdown-live-preview-engine 'pandoc)
      (org :variables org-agenda-files
           '("~/Documents/Org")
           org-directory '"~/Documents/Org")
      (python :variables python-enable-yapf-format-on-save t)
      (auto-completion :variables auto-completion-enable-help-tooltip t auto-completion-enable-help-tooltip 'manual auto-completion-enable-snippets-in-popup t auto-completion-enable-sort-by-usage t :disabled-for org markdown)
      (shell :variables shell-default-shell 'multi-term shell-default-height 80 shell-default-position 'bottom)
      (syntax-checking :variables flycheck-delayed-errors-timeout 500)
      (wakatime :variables wakatime-api-key "2ccf4cb6-4369-40b5-9165-718666b8bb32" wakatime-cli-path "/usr/local/bin/wakatime")
      cfg)
```
- System configuration features: NOTIFY ACL LIBXML2 ZLIB TOOLKIT_SCROLL_BARS

Miciah:
Does #11038 resolve the issue?

PR #11038, opened 2018-07-15
python: Initialize inferior-python-mode variables

Miciah:
Fix <kbd>SPC m c c</kbd> (`spacemacs/python-execute-file`) and <kbd>SPC m c C</kbd> (`spacemacs/python-execute-file-focus`), which were causing the following errors:

    save-current-buffer: Symbol’s value as variable is void: python-shell--interpreter
    error in process filter: python-shell-comint-end-of-output-p: Wrong type argument: arrayp, nil
    error in process filter: Wrong type argument: arrayp, nil

This commit also fixes issue #9070: `spacemacs/python-execute-file-focus` does not change focus.

The problem is that `spacemacs/python-execute-file-focus` invokes `spacemacs/python-execute-file`, `spacemacs/python-execute-file` invokes `inferior-python-mode`, and `inferior-python-mode` uses the variables `python-shell--interpreter` and `python-shell--interpreter-args`, which the python package declares but does not initialize.

To resolve this problem, this commit initializes both variables to `nil`.

* `layers/+lang/python/config.el`: Initialize `python-shell--interpreter` and `python-shell--interpreter-args` to `nil`.

### DIFF
--- a/layers/+lang/python/config.el
+++ b/layers/+lang/python/config.el
@@ -51,3 +51,8 @@ Possible values are `on-visit', `on-project-switch' or `nil'.")
 
 (defvar spacemacs--python-pyvenv-modes nil
   "List of major modes where to add pyvenv support.")
+
+;; inferior-python-mode needs these variables to be defined.  The python
+;; package declares them but does not initialize them.
+(defvar python-shell--interpreter nil)
+(defvar python-shell--interpreter-args nil)


bmag:
I think a better solution is to `defvar` both variables to nil in config.el:
```elisp
(defvar python-shell--interpreter nil)
(defvar python-shell--interpreter-args nil)
```

Miciah:
Thanks! Updated.

bmag:
Commit message needs to be updated as well

Miciah:
I did update the commit message.  Did I miss something?

bmag:
It's fine now, looks different than when I commented. (maybe I looked in my local copy of the PR before fetching newest changes? not sure)

bmag:

PR #11040, opened 2018-07-15
Add keybindings for scrolling in epub layer

Andre0991:
Add `d` and `u` for scrolling down and up, just like the `pdf` layer.

Note that `up` is actually represented by `nov-scroll-down` and vice-versa (following `pdf` layer convention).

### DIFF
--- a/layers/+readers/epub/packages.el
+++ b/layers/+readers/epub/packages.el
@@ -23,6 +23,8 @@
       (kbd "L") 'nov-next-document
       (kbd "[") 'nov-previous-document
       (kbd "]") 'nov-next-document
+      (kbd "d") 'nov-scroll-up
+      (kbd "u") 'nov-scroll-down
       (kbd "gm") 'nov-display-metadata
       (kbd "gr") 'nov-render-document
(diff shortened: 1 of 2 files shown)


layers/+readers/epub/packages.el
```diff
@@ -23,6 +23,8 @@
       (kbd "L") 'nov-next-document
       (kbd "[") 'nov-previous-document
       (kbd "]") 'nov-next-document
+      (kbd "d") 'nov-scroll-up
+      (kbd "u") 'nov-scroll-down
```
sdwolfz: It seems like these keybindings are the other way around.

Andre0991: It's because `nov-scroll-up` is actually `down` in Spacemacs conventions (following the `pdf` layer).

`nov-scroll-up` is like `C-d` with `evil`; `nov-scroll-down` is like `C-u`.

sdwolfz: Thanks for clarifying that.

sdwolfz:
Thank you :heart:!

PR #11041, opened 2018-07-15
Use lexical binding in dotfile template

Andre0991:
Lexical binding is easier to reason about and it's generally recommended for new code. It's faster too.

Dynamic scoping can lead to tricky situations. Example: https://emacs.stackexchange.com/questions/10394/scope-in-lambda

This probably won't break existing code. Most people don't even know the default is dynamic binding.

### DIFF
--- a/core/templates/.spacemacs.template
+++ b/core/templates/.spacemacs.template
@@ -1,5 +1,4 @@
-;; -*- lexical-binding: t -*-
-;; -*- mode: emacs-lisp -*-
+;; -*- mode: emacs-lisp; lexical-binding: t -*-
 ;; This file is loaded by Spacemacs at startup.
 ;; It must be stored in your home directory.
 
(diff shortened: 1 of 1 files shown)


sdwolfz:
I'll try to run with this for a while in my personal setup and see if I encounter any problems. Seems safe since the template only sets variables and defines functions... but still...

Also, could this be merged with the `mode: emacs-lisp` line, or is it better to have it separately?

syl20bnr:
While I agree that we should set it to true I fear the potential nasty side effects which might happen in some user's config.

> Also, could this be merged with the mode: emacs-lisp line, or is it better to have it separately?

+1 Please merge them on the same line.

Andre0991:
> +1 Please merge them on the same line.

Done. This is actually necessary, otherwise only the first line is considered.

> I haven't noticed any problems in my usage with this so far. Will merge this next week and if any problems arise I will be quick to revert.
 
👍

sdwolfz:
Thank you :heart:!

Issue #9477, opened 2017-08-22
spacemacs/smart-closing-parenthesis closes more parenthesis than are actually opened


kassick:
#### Description :octocat:
Closing a parenthesis right beside a question-mark, exclamation, space, #, among others, causes spacemacs to insert two closing parenthesis instead of one if dotspacemacs-smart-closing-parenthesis is non-nil

The issue does not occur with with dotspacemacs-smart-closing-parenthesis set to nil (default setting)

smart-closing-parenthesis itselt only inserts one parenthesis, but sp-up-exp ends up inserting another one before (insert ?\)) is called.

Tested with a clean spacemacs install, same issue.

#### Reproduction guide :beetle:
- Set dotspacemacs-smart-closing-parenthesis to t
- Start Emacs
- Open some text file (e.g. org-mode) that won't auto enable smartparens-mode
- start typing something such as

  this is some (test within parenthesis!

- Press ")" (closing parenthesis)

*Observed behaviour:* :eyes: :broken_heart:

two parenthesis are inserted and the resulting text is

```
  this is some (test within parenthesis)!)
```

*Expected behaviour:* :heart: :smile:

Exactly one parenthesis is inserted and the resulting text would be

```
  this is some (test within parenthesis!)
```

#### System Info :computer:
- OS: gnu/linux
- Emacs: 25.2.1
- Spacemacs: 0.200.9
- Spacemacs branch: master (rev. 8e1af145)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(javascript helm
            (auto-completion :variables auto-completion-enable-snippets-in-popup t auto-completion-enable-help-tooltip 'manual)
            emacs-lisp git ranger html markdown org
            (shell :variables shell-default-height 30 shell-default-position 'bottom)
            (syntax-checking :variables syntax-checking-enable-by-default t)
            (spell-checking :variables spell-checking-enable-by-default nil spell-checking-enable-auto-dictionary t)
            ycmd python c-c   fsharp csharp java antlr graphviz
            (latex :variables latex-enable-folding t latex-enable-auto-fill nil)
            markdown pandoc pdf-tools shell-scripts semantic
            (ibuffer :variables ibuffer-group-buffers-by 'projects)
            themes-megapack theming kzk-visual-editing kzk-helm kzk-company)
```
[...]

PR #11042, opened 2018-07-15
Fix smart-closing-parenthesis extra parenthesis

Miciah:
Fix issue #9477: `spacemacs/smart-closing-parenthesis` sometimes adds an extra closing parenthesis.

`spacemacs/smart-closing-parenthesis` may call `sp-up-sexp` twice: first to find the position of an existing closing delimiter or the position at which one would be inserted, and second to insert a delimiter if necessary or else move point after the existing closing delimiter.

The problem is that each call to `sp-up-sexp` may insert a delimiter, but `spacemacs/smart-closing-parenthesis` is supposed to insert at most one.

The solution is to undo any edit performed by the first call to `sp-up-sexp`.

* `layers/+spacemacs/spacemacs-editing/funcs.el`: Undo any edit performed by the first call to `sp-up-sexp`.

---

@StreakyCobra, does this change look correct?

### DIFF
--- a/layers/+spacemacs/spacemacs-editing/funcs.el
+++ b/layers/+spacemacs/spacemacs-editing/funcs.el
@@ -22,14 +22,37 @@
   (indent-according-to-mode))
 
 (defun spacemacs/smart-closing-parenthesis ()
+  "Insert a closing pair delimiter or move point past existing delimiter.
+
+If the expression at point is already balanced and there is a
+closing delimiter for that expression on the current line, move
+point forward past the closing delimiter.
+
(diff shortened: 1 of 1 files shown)


Miciah:
Latest push adds a docstring.  I realized that my original solution did not work properly in some situations (it did not insert the correct delimiter in some cases), so I reworked it to set `sp-navigate-close-if-unbalanced` to `t` for both calls to `sp-up-sexp` and undo any changes that the first invocation makes.  I believe this approach is more correct, but the code is also a bit more gnarly now.

syl20bnr:
Thank you ! 👍 

PR #11047, opened 2018-07-17
Fix overflow of line in page-break-lines-mode when using line numbers

Compro-Prasad:
This is picked up from

https://github.com/purcell/page-break-lines/blob/fd3b7e38ad8747cd009ead7ef1bb150849ccc693/page-break-lines.el

Do check the full diff as it contains many changes about versions(its 0?).

@bmag , @syl20bnr

### DIFF
--- a/core/libs/page-break-lines.el
+++ b/core/libs/page-break-lines.el
@@ -1,11 +1,11 @@
-;;; page-break-lines.el --- Display ugly ^L page breaks as tidy horizontal lines
+;;; page-break-lines.el --- Display ^L page breaks as tidy horizontal lines
 
 ;; Copyright (C) 2012-2015 Steve Purcell
 
 ;; Author: Steve Purcell <steve@sanityinc.com>
 ;; URL: https://github.com/purcell/page-break-lines
-;; Package-Version: 20160109.1813
-;; Package-X-Original-Version: DEV
+;; Package-Version: 0
+;; Package-Requires: ((emacs "24.4"))
 ;; Keywords: convenience, faces
 
 ;; This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


sdwolfz:
Wow, I always (wrongly) assumed the `^L` display as a line was done by emacs by default. I tried it out and it does fix the issue with line numbers, but looks like it still has problems with the git gutter. Anyway it's an external package so these issues should be fixed upstream. I'll merge in a few days if nobody else does it before I can.

Compro-Prasad:
Its preferable to use the package itself. It may fix changes that may come up in future. This way we don't need to modify multiple times as the actual package changes.

Compro-Prasad:
The CI is fixed. Wonderful. Looks premium now.

syl20bnr:
The reason why it is in core is because we need it in the home buffer, so this is just aesthetic reason :-) I guess we can live with updating it once in a while. Nice PR 👍

core/libs/page-break-lines.el
```diff
@@ -1,11 +1,11 @@
-;;; page-break-lines.el --- Display ugly ^L page breaks as tidy horizontal lines
```
syl20bnr: ahah, funny change :-)

purcell:
> Its preferable to use the package itself. It may fix changes that may come up in future. This way we don't need to modify multiple times as the actual package changes.

Yeah, I agree - it's unclear why this package would need to be vendored here. The vendoring led to @Compro-Prasad filing an issue upstream for something that had already been resolved.

sdwolfz:
I think it was vendored since the functionality was needed even before packages were installed.

sdwolfz:
Thank you :heart:!

Issue #11037, opened 2018-07-15
Window resizing horizontally when using pop shells


mhaffner:
#### Description :octocat:
Let's say I have two windows open (side by side) and use `SPC '` from window 1 to open a pop shell. The pop shell opens below window 1 with a height of 35 (as expected; see my layers), but window 1 (and also this newly created window 3) are shrunk in half horizontally.

This is ocurring on my main machine, which is Manjaro i3 with Emacs 25.3 and a bunch of layers, and also this example which is a clean install of Manjaro i3, Emacs 26.1, and Spacemacs; only the shell variables are modified.

#### Reproduction guide :beetle:
- Start Emacs
- `SPC w /` to split the window
- `SPC '` to open a pop shell

*Observed behaviour:* :eyes: :broken_heart:
Windows 1 and 3 are shrunk (looks like in half) horizontally, and thus window 2 is much larger.

*Expected behaviour:* :heart: :smile:
Window 3 opens below window 1 but nothing else happens. Window 2 should not resize at all.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 26.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. e109777f8)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(helm emacs-lisp neotree
      (shell :variables shell-default-shell 'eshell shell-default-height 35 shell-default-full-span nil shell-default-position 'bottom))
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GSETTINGS NOTIFY ACL GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 MODULES THREADS LIBSYSTEMD LCMS2

#### Backtrace :paw_prints:
There's no error so no backtrace, I guess? I'd be curious about debugging this on my own to figure out why the windows are resizing horizontally, so tips on that would be appreciated.

Miciah:
This issue is caused by #1023 and should be resolved by #11051.

mhaffner:
@Miciah Woot. Thanks for working on this and for referencing the root issue. Through that I was able to create a workaround until the PR gets merged.

PR #11051, opened 2018-07-18
shell: Adjust popup shell width only for left/right

Miciah:
Fix issue #11037: Do not adjust the width of a popup shell's window unless its position is left or right.

* `layers/+tools/shell/funcs.el`: Check that `shell-pop-window-position` is `'left` or `'right` before adjusting the window's width.


sdwolfz:
Thank you :heart:!

PR #11052, opened 2018-07-18
Add the missing keybindings in ruby-on-rails layer

asok:
@sdwolfz those commands were missing in the spacemacs key map.

### DIFF
--- a/layers/+frameworks/ruby-on-rails/packages.el
+++ b/layers/+frameworks/ruby-on-rails/packages.el
@@ -26,6 +26,7 @@
       ;; Find files
       (spacemacs/set-leader-keys-for-minor-mode 'projectile-rails-mode
         "ffa" 'projectile-rails-find-locale
+        "ffb" 'projectile-rails-find-job
         "ffc" 'projectile-rails-find-controller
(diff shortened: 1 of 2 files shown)


layers/+frameworks/ruby-on-rails/packages.el
```diff
@@ -44,6 +44,8 @@
         "ffv" 'projectile-rails-find-view
         "ffy" 'projectile-rails-find-layout
         "ff@" 'projectile-rails-find-mailer
+        "ffw" 'projectile-rails-find-webpack
+        "ffb" 'projectile-rails-find-job
```
sdwolfz: This should be ordered alphabetically under `"ffa" 'projectile-rails-find-locale`.

asok: Done.

sdwolfz: You should do it in the documentation as well.

asok: Double done.

sdwolfz:
Thank you :heart:!

PR #11057, opened 2018-07-18
auto-completion: make company-backends-($major-mode) user-definable.

Alexander-Miller:
The functions created in `spacemacs|add-company-backends` are put into major mode hooks, so there's no good way to overwrite them. This small change will make it possible to define your own major-mode based backend lists.

I am also pretty sure that this was already possible a while ago, but recently changed. It's that or I've been doing it wrong for a long time.

### DIFF
--- a/layers/+completion/auto-completion/funcs.el
+++ b/layers/+completion/auto-completion/funcs.el
@@ -94,14 +94,15 @@ Available PROPS:
         ;; add backends
         (dolist (backend backends)
           (push `(add-to-list ',raw-backends-var-name ',backend) result))
+        ;; define mode-local backends with `defvar' so they can be overwritten
+        (if auto-completion-enable-snippets-in-popup
+            (push
+             `(defvar ,backends-var-name (mapcar 'spacemacs//show-snippets-in-company ,raw-backends-var-name))
+             result)
+          (push `(defvar ,backends-var-name ,raw-backends-var-name) result))
         ;; define initialization hook function
         (push `(defun ,init-func-name ()
                 ,(format "Initialize company for %S." mode)
-                (if auto-completion-enable-snippets-in-popup
-                    (setq ,backends-var-name
-                          (mapcar 'spacemacs//show-snippets-in-company
-                                  ,raw-backends-var-name))
-                  (setq ,backends-var-name ,raw-backends-var-name))
                 (set (make-variable-buffer-local 'auto-completion-front-end)
                      'company)
                 (set (make-variable-buffer-local 'company-backends)


Compro-Prasad:
Moving `(defvar ,backends-var-name` outside is better IMO.

Alexander-Miller:
Done.

syl20bnr:
It won't work as expected as we already `defvar` the variable here: https://github.com/syl20bnr/spacemacs/blob/develop/layers/+completion/auto-completion/funcs.el#L92 I'm pretty sure we would have regression, but we have no tests for this 👼 

You are right that there is a regression for your use-case introduced maybe a few month ago here: https://github.com/syl20bnr/spacemacs/commit/40f28c7d1

What if you set `company-backends-($major-mode)-raw` variable instead ?

Alexander-Miller:
> What if you set company-backends-($major-mode)-raw variable instead ?

It would work for *me*, but I'd have to wonder about the design goal in this matter - should there be a general means for users to decide what major-mode gets what backends or should this be reserved for the kind of power-user types who know how company works and can make sense of that massive macro?

Not that the solution for the first approach needs to incur a large refactoring. I think we can do with giving `backends-raw` a better fitting name and an appropriate doc-string.

syl20bnr:
>  should this be reserved for the kind of power-user types who know how company works

I believe so but I may miss a common use case for users. What is your use case of this customization ?

We can create a function that manipulates the not user-friendly named variable `-raw`.

Alexander-Miller:
> What is your use case of this customization ?

Keeping backends separate, as is done right now, means there are situations where backends further in the back of the list won't be asked for completions, even though they could supply them. For example you cannot complete `~/` in elisp mode with the current setup.

That is why I prefer to use one combined backend, using a main backend like capf, supported by helpers like dabbrev-code, files and yasnippet.

I suppose some people might also prefer (not) to include backends like dabbrev, keywords or etags in their setup.

> We can create a function that manipulates the not user-friendly named variable -raw.

Sounds good to me.

syl20bnr:
> For example you cannot complete ~/ in elisp mode with the current setup.

Looks like a bug, if we loose completion functionality with the spacemacs setup this is not good.

Alexander-Miller:
Not really, I think. It's been years since I've had that conversation with Dmitry, so I cannot remember all the details, but basically it's working as intended, since, depending on *how* capf refuses to complete, all the other backends after it, like files, won't even be queried, and if you don't like that you should combine your backends, like I did.

There's also the question of whether some users would want to include backends like dabbrev in their general completions, which is really just a matter of taste.

Also note that my example only applies to the unquoted text. Most of your paths in elisp will be quoted strings in quotes and completing `"~/"` does work.

Alexander-Miller:
I have since moved to a 100% custom config, so let's shelve this.

PR #11059, opened 2018-07-19
Fix compilation finish hook

Miciah:
#### Fix compilation finish hook

Use `compilation-finish-functions` instead of `compilation-finish-function`.

Spacemacs assigns an anonymous function to the `compilation-finish-function` variable, which was deprecated in Emacs 22.1 and removed in Emacs 27.1 in favor of the `compilation-finish-functions` hook.  This commit deletes this assignment, replaces the anonymous function with a named function, and adds the newly named function to `compilation-finish-functions`.

* `layers/+spacemacs/spacemacs-defaults/funcs.el` (`compilation-finish-function`): Delete assignment of an anonymous function.
(`spacemacs/compilation-finish-function`): New defun based on the old anonymous function.
* `layers/+spacemacs/spacemacs-defaults/config.el` (`compilation-finish-functions`): Add `spacemacs/compilation-finish-function`.

#### Avoid hardcoded key binding in compilation message

* `layers/+spacemacs/spacemacs-defaults/funcs.el` (`spacemacs/compilation-finish-function`): Use `substitute-command-keys` to replace the hardcoded key binding in the message that is printed when there are errors.

#### Delete redundant "compilation ok" message

The compile command already prints "Compilation finished" if there are no errors.

* `layers/+spacemacs/spacemacs-defaults/funcs.el` (`spacemacs/compilation-finish-function`): Delete "compilation ok" message.

---

Really, the compilation finish hook does not seem especially useful—all it adds is printing the key binding for `spacemacs/next-error`, and it prevents the echo line from showing the default output.  For example, my `*Messages*` buffer shows the following after I enter `:make false`:

    Compilation exited abnormally with code 1
    There were errors. Use SPC e n to visit.

Because Spacemacs prints its "There were errors" message immediately after Emacs prints its "Compilation exited abnormally" message, the latter is obscured.

Do we want to keep this hook at all?

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/config.el
+++ b/layers/+spacemacs/spacemacs-defaults/config.el
@@ -63,6 +63,9 @@ It runs `tabulated-list-revert-hook', then calls `tabulated-list-print'."
 ;; Scroll compilation to first error or end
 (setq compilation-scroll-output 'first-error)
 
+;; Print the key binding for spacemacs/next-error on errors.
+(add-hook 'compilation-finish-functions #'spacemacs/compilation-finish-function)
+
 ;; Don't try to ping things that look like domain names
 (setq ffap-machine-p-known 'reject)
 
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -1217,7 +1217,9 @@ In case of errors, show the key binding for spacemacs/next-error."
             (string-match "FAILED" (buffer-string)))
 
         ;; there were errors
-        (message "There were errors. SPC-e-n to visit.")
+        (message "%s"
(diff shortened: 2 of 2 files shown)


Compro-Prasad:
> Do we want to keep this hook at all?

No from my side.

syl20bnr:
> No from my side.

+1

Miciah:
#11185 is now in contention for the merge.  May the better PR win!

Compro-Prasad:
I prefer the other PR but it needs @syl20bnr 's approval.

sdwolfz:
From the feedback gathered so far, I'll go with #11185, unless anybody objects.

sdwolfz:
I went ahead and merged #11185 since nobody had any objections. Thank you for your effort :heart:!

PR #11068, opened 2018-07-20
add guard to check winum is loaded to fix #10446

### LINKED ISSUE #10446: Symbols's function definition is void: winum--window-list
#### Description :octocat:
When I use Ctrl-x Ctrl-f to open a file, it wouldn't show the new buffer, only opened in the background, and shows "Symbols's function definition is void: winum--window-list" in the minibuffer, this bug also occurs when I use Ctrl-x b to switch buffer

- Start Emacs
- At home page, Ctrl-x Ctrl-f to open a file / Ctrl-x b to switch buffer

Showed "Symbols's function definition is void: winum--window-list" in the minibuffer

Show me the new buffer

#### SYSTEM_INFO
- OS: GNU/Linux kernel 4.15
- Emacs Version: 27(Latest from Emacs master branch)
- Spacemacs: 0.300
- Spacemacs branch: develop(HEAD 2e8727ae)
- Graphic display: t

#### Backtrace
```
Symbols's function definition is void: winum--window-list
This function is called in `.emacs/layers/+completion/helm/funcs.el`
```

dangirsh:
Add guard around the spacemacs/helm-find-files-windows binding, which implicitly depends on winum. 

Fixes #10446

### DIFF
--- a/layers/+completion/helm/packages.el
+++ b/layers/+completion/helm/packages.el
@@ -319,8 +319,9 @@
       (defalias
         'spacemacs/helm-project-do-grep-region-or-symbol
         'helm-projectile-grep))
-    :config (define-key helm-projectile-find-file-map
-              (kbd "RET") 'spacemacs/helm-find-files-windows)))
+    :config (when (configuration-layer/package-used-p 'winum)
+              (define-key helm-projectile-find-file-map
+                (kbd "RET") 'spacemacs/helm-find-files-windows))))
 
 (defun helm/init-helm-spacemacs-help ()
   (use-package helm-spacemacs-help


layers/+completion/helm/packages.el
```diff
@@ -319,8 +319,9 @@
       (defalias
         'spacemacs/helm-project-do-grep-region-or-symbol
         'helm-projectile-grep))
-    :config (define-key helm-projectile-find-file-map
-              (kbd "RET") 'spacemacs/helm-find-files-windows)))
+    :config (when (configuration-layer/package-used-p 'winum)
```
Compro-Prasad: I think it should be wrapped around (with-eval-after-load 'winum ... ) and not (when ... ).

dangirsh: I'm just following what @syl20bnr did in e5d8603 . The original issue was that this layer can be loaded without winum (like with `(helm :packages (not winum))`), so the approach in this change seems most appropriate.

Issue #10760, opened 2018-05-28
Magithub sections not added to Magit unless it's explicitly loaded first


marienz:
#### Description :octocat:
Magithub sections are not added to Magit unless it's explicitly loaded first (either explicitly or by calling some autoloaded function from it).

#### Reproduction guide :beetle:
- Start Spacemacs (develop branch) with the git and github layers enabled
- Open a Magit status buffer (`SPC f e i SPC g s` opens one for Spacemacs)
- Kill the status buffer
- `(require 'magithub)`
- Open the same Magit status buffer again

*Observed behaviour:* :eyes: :broken_heart:
Magit status buffer does not have Magithub's sections added to it the first time, but does the second time.

*Expected behaviour:* :heart: :smile:
Magit behaves the same way both times.

I tried removing the `:defer t` from `use-package magithub` (to make it more like the [Magithub quickstart](https://github.com/vermiculus/magithub#quick-start)). That seems to work: it doesn't load magit when I start Spacemacs, but it does load Magithub along with Magit. But I don't know if that's the correct Spacemacsy way of doing it :)

Macro-expanding the `use-package` call as Spacemacs has it confirms the `magithub-feature-autoinject` call ends up in `eval-after-load 'magithub`, which defeats its purpose.

I did not see the same problem on one of my other systems, where I had not updated packages for a while. I suspect `use-package`'s behavior changed.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 26.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. b68bd33a1)
- Graphic display: t
- Distribution: spacemacs
- Editing style: emacs
- Completion: ivy
- Layers:
```elisp
(ivy git github emacs-lisp)
```
- System configuration features: XAW3D XPM JPEG TIFF PNG RSVG SOUND DBUS NOTIFY ACL GNUTLS LIBXML2 FREETYPE XFT ZLIB TOOLKIT_SCROLL_BARS LUCID X11 THREADS LIBSYSTEMD LCMS2

Miciah:
I think deleting `:defer t` so that Magithub loads when Magit loads is the best solution.  `magithub-feature-autoinject` is nontrivial—it hooks MagitHub into Magit in several ways (adding an action to a popup, adding a binding to one of Magit's key maps, and hooking into the status buffer)—so trying to do anything more nuanced would be much pain for little gain.

#11071 makes this change.

marienz:
[...]

PR #11071, opened 2018-07-22
Fix loading of Magithub

Miciah:
Fix issue #10760: Magithub sections are not added to Magit.

The problem is that the Spacemacs init function's `use-package` form for Magithub specifies both `:after magit` and `:defer t`. It should specify only the former so that Magithub will be loaded when Magit is loaded.

* `layers/+source-control/github/packages.el`: (`github/init-magithub`) Delete `:defer` argument.

### DIFF
--- a/layers/+source-control/github/packages.el
+++ b/layers/+source-control/github/packages.el
@@ -81,7 +81,6 @@
 
 (defun github/init-magithub ()
   (use-package magithub
-    :defer t
     :after magit
     :init
     (setq magithub-dir (concat spacemacs-cache-directory "magithub/"))


sdwolfz:
I'm not sure It's a good idea to have this functionality by default, when I tried it out `SPC g s` would get stuck on fetching all the issues from this repo.

Miciah:
The user can toggle Magithub features per repository using `SPC g m H C`, or we can put Magithub into "offline mode" by default: 
```emacs-lisp
(magit-set "false" "--global" "magithub.online")
```
Should we set the global default setting?

sdwolfz:
Hmm... how about a configuration variable that sets it online, being offline by default? This way it can be controlled from the `.spacemacs` config.

Miciah:
I hesitate to avoid adding a new configuration variable if we can avoid it, and `magithub.online` is already a sort of configuration variable, albeit one that is stored in the Git configuration.  We could set it to "false" iff it does not have a setting:
```emacs-lisp
(when (null (magit-get "--global" "magithub.online"))
  (magit-set "false" "--global" "magithub.online"))
```
Then we could document how to override this default using `git config` or `magit-set`.  Would that be acceptable?

syl20bnr:
what does `:after` do exactly ? what is the expanded form ?

Miciah:
Using `:after` without `:defer` will cause magithub to be loaded when magit is loaded.  Here is the expanded form:
```emacs-lisp
(progn
    (defvar use-package--warning672
      #'(lambda
          (keyword err)
          (let
              ((msg
                (format "%s/%s: %s" 'magithub keyword
                        (error-message-string err))))
            (display-warning 'use-package msg :error))))
    (condition-case-unless-debug err
        (eval-after-load 'magit
          '(progn
             (condition-case-unless-debug err
                 (when
                     (run-hook-with-args-until-failure 'use-package--magithub--pre-init-hook)
                   (setq magithub-dir
                         (concat spacemacs-cache-directory "magithub/"))
                   (run-hooks 'use-package--magithub--post-init-hook))
               (error
                (funcall use-package--warning672 :init err)))
             (if
                 (not
                  (require 'magithub nil t))
                 (display-warning 'use-package
                                  (format "Cannot load %s" 'magithub)
                                  :error)
               (condition-case-unless-debug err
                   (progn
                     (when
                         (run-hook-with-args-until-failure 'use-package--magithub--pre-config-hook)
                       (progn
                         (magithub-feature-autoinject t)
                         (define-key magit-status-mode-map "@" #'magithub-dispatch-popup))
                       (run-hooks 'use-package--magithub--post-config-hook))
                     t)
                 (error
                  (funcall use-package--warning672 :config err))))))
      (error
       (funcall use-package--warning672 :catch err))))
```

syl20bnr:
Thank you, I'm fine with using `:after` keyword then 👍

syl20bnr:
Thank you ! 👍 

Miciah:
It's more about how annoying the surprise is.

* When Magithub is in offline mode, it adds a colorful and informative line to the Magit status window.

* When Magithub is in online mode, it adds a list of issues, a list of PRs, and a delay while it fetches this information.  This delay ranges from several seconds for a project with few issues and PRs (e.g., about 10 seconds for the Evil repository) to several minutes for a project with many issues and PRs (e.g., about 5 minutes for the Kubernetes repository; I recall that Spacemacs took a few minutes when I tried it before, but right now it isn't working at all for some reason).  During this time, Emacs is unusable.

Either one may be surprising to a user who is seeing Magithub for the first time, but online mode has great potential to enrage users.

sdwolfz:
Using this:

```
(when (null (magit-get "--global" "magithub.online"))
  (magit-set "false" "--global" "magithub.online"))
```
Does not seem to work, magithub is still online and it takes a lot of time to do a `SPC g s`.
I've excluded `magithub` right now from my installed packages so I can still use the normal functionality. Not sure what would be a good solution for this... but it's definitely not letting it as it is right now.

Miciah:
Do you already have `magithub.online = false` in your `.gitconfig` file? What does `(magit-set "false" "--global" "magithub.online")` return?

The `when` form is only supposed to change the setting if no value is set, the idea being that it will set a default if the first time the github layer is initialized but will not touch the setting after that, or if the user has already configured the setting.  To change the setting when it has already been set before, use just the `magit-get` form.

sdwolfz:
I've used your setting and `(magit-get "--global" "magithub.online")` returns `"false"`, the problem still persists.

Somebody opened an issue related to this: https://github.com/syl20bnr/spacemacs/issues/11176. Let's discuss there since it's easier to track that way.

PR #11086, opened 2018-07-27
fix: spacemacs/pyenv-executable-find support multi pyenv versions

yetone:
Because of the `pyenv` support setting multiple versions, so `string-match` will always failed. And the `spacemacs/pyenv-executable-find` will always return `nil`.

```sh
➜ pyenv version-name
2.7.14:3.6.4
```


syl20bnr:
Thank you !  👍 

PR #11092, opened 2018-07-28
Develop

rilerez:
Aaaa sorry I meant to push this to develop.

I escape the filename before calling python through bash in the function `spacemacs/python-execute-file`. A filename with spaces, for example, broke the previous `spacemacs/python-execute-file`. Calling `shell-quote-argument` before passing to the shell fixes this problem.

See this superuser.com question for reference https://superuser.com/questions/159412/escaping-spaces-in-emacs-perl


Compro-Prasad:
@rilerez Are you aware of this? Why haven't you closed it yet?

syl20bnr:
Closing in favor of https://github.com/syl20bnr/spacemacs/pull/11093

PR #1109, opened 2015-04-11
fix org toc

cpaulik:

PR #11093, opened 2018-07-28
Fix #1109: Escape argument for shell spacemacs/python-execute-file

rilerez:
Escape the filename before calling python on the filename in the shell in the
function `spacemacs/python-execute-file` by calling `shell-quote-argument`.

Thank you <3


syl20bnr:
Thank you ! 👍 

PR #11098, opened 2018-07-28
WIP Fix idris file loading.

### LINKED ISSUE #11097: Bug in idris layer: Wrong type argument when loading file.
#### Description :octocat:

error in process filter: idris-filename-to-load: Wrong type argument: stringp, nil
error in process filter: Wrong type argument: stringp, nil

This causes long delays in opening the holes buffer.

- Load an idris file with `, r`

See the wrong type argument error in messages. Holes buffer takes a few seconds to load.

Fast holes buffer pops up with no error.

#### System Info :computer:
- OS: darwin
- Emacs: 26.1
- Spacemacs: 0.300.0
- Spacemacs branch: my-develop (rev. 79bbbe270)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: ivy
- Layers:
```elisp
((auto-completion
  (haskell :variables haskell-completion-backend 'dante))
 c-c++ clojure coq csv dash docker
 (elm :variables elm-format-command "elm-format")
 (emacs-lisp :variables emacs-lisp-hide-namespace-prefix nil)
 emoji erlang evil-snipe fsharp
 (git :variables git-magit-status-fullscreen t)
 graphviz
 (gtags :variables gtags-enable-by-default t)
 (haskell :variables haskell-process-type 'stack-ghci haskell-enable-hindent-style "fundamental")
 html
 (ibuffer :variables ibuffer-group-buffers-by 'projects)
 idris ivy
 (java :variables java-backend 'ensime)
 (javascript :variables tern-command
             '("node" "/usr/local/bin/tern"))
 (latex :variables latex-build-command "LaTeX" latex-enable-auto-fill t latex-enable-folding t)
 lsp
 (markdown :variables markdown-live-preview-engine 'vmd)
 nginx nixos
 (org :variables org-enable-reveal-js-support t org-enable-github-support t org-projectile-file "TODOs.org")
 osx
 (python :variables python-backend 'lsp python-enable-yapf-format-on-save t)
 restclient rust
 shell-scripts
 spotify sql swift syntax-checking systemd tmux treemacs twitter version-control vimscript vinegar windows-scripts yaml)
```
- System configuration features: NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS MODULES THREADS

#### Backtrace :paw_prints:
```
Debugger entered--Lisp error: (wrong-type-argument stringp nil)
  file-name-directory(nil)
  idris-filename-to-load()
  idris-update-loaded-region(nil)
  #f(compiled-function (result) #<bytecode 0x415eb15d>)(nil)
  #f(compiled-function (result) #<bytecode 0x415eb191>)(nil)
  funcall(#f(compiled-function (result) #<bytecode 0x415eb191>) nil)
  (progn (set-buffer (symbol-value G11)) (funcall (symbol-value G12) result))
  (if (symbol-value G12) (progn (set-buffer (symbol-value G11)) (funcall (symbol-value G12) result)))
  (let* ((--cl-rest-- rand-8) (result (if (= (length --cl-rest--) 1) (car --cl-rest--) (signal 'wrong-number-of-arguments (list nil (length --cl-rest--)))))) (if (symbol-value G12) (progn (set-buffer (symbol-value  ;; ... elided
```
### Additional information
Setting `:noselect t` for the idris buffers in `idris/pre-init-popwin` fixes the issue. I reported upstream in https://github.com/idris-hackers/idris-mode/issues/423 .

jsoo1:
Don't select idris popwins to avoid idris loading errors.

  - Fixes #11097

### DIFF
--- a/layers/+lang/idris/packages.el
+++ b/layers/+lang/idris/packages.el
@@ -124,9 +124,9 @@
 (defun idris/pre-init-popwin ()
   (spacemacs|use-package-add-hook popwin
     :post-config
-    (push '("*idris-notes*" :dedicated t :position bottom :stick t :noselect nil :height 0.4)
+    (push '("*idris-notes*" :dedicated t :position bottom :stick t :noselect t :height 0.4)
           popwin:special-display-config)
-    (push '("*idris-holes*" :dedicated t :position bottom :stick t :noselect nil :height 0.4)
+    (push '("*idris-holes*" :dedicated t :position bottom :stick t :noselect t :height 0.4)
           popwin:special-display-config)
-    (push '("*idris-info*" :dedicated t :position bottom :stick t :noselect nil :height 0.4)
+    (push '("*idris-info*" :dedicated t :position bottom :stick t :noselect t :height 0.4)
           popwin:special-display-config)))


Miciah:
From your comments in the linked issue and upstream issue, it looks like the issue lies with idris-mode's `idris-filename-to-load` function, which uses the current buffer instead of `idris-currently-loaded-buffer`.  Shouldn't the issue be fixed in idris-mode?

jsoo1:
idris-mode also provides hooks for successful loading of a file (`idris-load-file-success-hook`). I am not familiar with popwin enough, but can we hook into that rather than use raw popwins?

Miciah:
All popwin does here is identify windows that idris-mode creates, adjust those windows' sizes and positions, and prevent those windows from getting automatically selected.  Popwin does so using `display-buffer-alist`, which is part of Emacs.

Maybe we can use hooks provided by idris-mode, but then we're foregoing general functionality that popwin provides using standard Emacs mechanisms, and re-implementing it ad hoc using mechanisms specific to idris-mode.  If the issue can be resolved by fixing idris-mode's assumptions about window focus, then that seems preferable.  I suggest at least waiting for the idris-mode maintainers' response.

On the other hand, it may be that the issue is annoying enough to warrant applying an immediate short-term workaround in Spacemacs.  Long-term, I hope the issue would still be resolved upstream.

Disclaimer: I'm not a maintainer of Spacemacs, popwin, or idris-mode.

jsoo1:
Neither am I. The issue is fairly annoying and I think warrants a workaround. Also, I am not sure about the UI decision to select the popup window anyways. Is there a spacemacs convention for selecting the popwin when creating one? I'm not sure I love the experience when doing idris work.  I put in https://github.com/idris-hackers/idris-mode/pull/483 to maybe address the issue in idris mode, but I think the general user experience issue of immediately focusing on the window should be addressed too.

syl20bnr:
> Is there a spacemacs convention for selecting the popwin when creating one? 

Not really, but we have a recurrent scheme which is to have the lower case key with no focus and the corresponding upper case key to focus.

For instance: `SPC w v` vs `SPC w V`, `SPC e l` vs `SPC e L`.

If it can be applied here this would be ideal. I guess that in all cases the popup should not take the focus so we can decide for ourselves to focus or not afterwards.

jsoo1:
Thanks. I'll take a look at using the lower/upper case to select the popup.

sdwolfz:
@jsoo1 have you reached any conclusion related to this PR? Would you like to wait for the issue to be fixed upstream or have this PR merged in the meantime?

jsoo1:
I have been meaning to work on this but became very busy. I would like to follow up on using the capital keybinding for focus. My upstream PR is not ready yet either. Let’s keep it open and I think I can work on it this week.

duianto:
I added a prefix `WIP` to the title, to make easy to see that its in progress. Just change it back when it's ready.

PR #11107, opened 2018-07-30
Open javascript files in js2 mode

deb0ch:
The javascript layer configures a lot of things for js2-mode, but as it doesn't change the default mode for js files none of it is effective when just activating the layer. And as it is not documented, some research is necessary to benefit the features, more or less painful according to your proficiency with (spac)emacs, that more or less users will have the courage to go through.

I also included `.jsx` files, as it is a common extension for javascript files when using React. It is the recommended extension in the Airbnb coding style, the seemingly most widely adopted coding style among the Javascript community.

### DIFF
--- a/layers/+lang/javascript/packages.el
+++ b/layers/+lang/javascript/packages.el
@@ -75,7 +75,7 @@
 (defun javascript/init-js2-mode ()
   (use-package js2-mode
     :defer t
-    :mode "\\.m?js\\'"
+    :mode (("\\.m?js\\'"  . js2-mode) ("\\.jsx\\'" . js2-mode))
     :init
     (progn
       (add-hook 'js2-mode-local-vars-hook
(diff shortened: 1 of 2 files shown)


layers/+lang/javascript/packages.el
```diff
@@ -86,6 +86,9 @@
                      (cons 'javascript-backend value))))
     :config
     (progn
+      ;; file extensions
+      (push '("\\.js\\'"  . js2-mode) auto-mode-alist)
+      (push '("\\.jsx\\'" . js2-mode) auto-mode-alist)
```
syl20bnr: I would prefer using `add-to-list` instead of push here to prevent stacking this on top of the list with each `SPC f e R`. Also I would put these in `:init` instead of  `:config`, also check use-package's `:mode` (example here: https://github.com/syl20bnr/spacemacs/blob/develop/layers/+readers/pdf/packages.el#L18).

deb0ch: A better, an even better and an even-even-better solution 😸

deb0ch:
Updated 👍:sparkles:

sdwolfz:
Thank you :heart:!

PR #11115, opened 2018-08-01
spacemacs-defaults: add a keybind for `find-alternate-file`

miyabinomuratic:
The function is useful when you mistakenly open a wrong file and want to open a correct one while closing the wrong one. The function is also sometimes recommended to "reload" a file. So let's add a shortcut for the function.

Note: The function is bound to `C-x C-v` by default, but I would like to have it in the leader key menu to make it discoverable and easier to type.

### DIFF
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -2369,6 +2369,7 @@ Files manipulation commands (start with ~f~):
 
 | Key Binding | Description                                                                                            |
 |-------------+--------------------------------------------------------------------------------------------------------|
+| ~SPC f a~   | open a file and replace the current buffer with the new file                                           |
 | ~SPC f b~   | go to file bookmarks                                                                                   |
 | ~SPC f c~   | copy current file to a different location                                                              |
 | ~SPC f C d~ | convert file from unix to dos encoding                                                                 |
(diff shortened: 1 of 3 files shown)


syl20bnr:
mmmmh I find the term used in this function ambiguous, `alternate` brings the idea of succession like:

> A alternates with B

means B replaces A and A replaces B etc...

So to me the first thing that comes to mind with `alternate` is alternating between a file and its counterpart like a code file and its associated test file, or implementation file with its declaration file etc...

I suggest we alias this function to something more explicit like `spacemacs/find-file-and-replace-buffer`.

Miciah:
Using an alias for the Spacemacs key binding should also cause it to use Helm or Ivy instead of Ido (related: #10772).

miyabinomuratic:
Thank you for your comments. I'm OK with the suggested name and using an alias. I am going to write a patch.

Should I squash the old commit and the new one on my side?

sdwolfz:
@miyabinomuratic yes, squashing them would be great.

Compro-Prasad:
@miyabinomuratic I am not a bot. This PR needs an update and I want to be subscribed to it :laughing:

miyabinomuratic:
Hi. I updated the patch with the suggestions, squashed into one commit and rebased it onto the latest develop branch.

Can you review the new patch please?

sdwolfz:
Thank you :heart:!

miyabinomuratic:
Thank you for accepting the PR :-)

PR #11118, opened 2018-08-02
html: fix indenting for emacs editing style

ellakk:
When using Emacs style you use tab to indent. This binding in the html-layer make's tab either expand text or do nothing when editing html or jsx (React) code. I think using C-j (the default) for emmet is fine for Emacs style and the html-layer even links to the emmet documentation that says that C-j is used for expansion.


syl20bnr:
Instead of `fix indenting for emacs editing style` can you change the commit message that better convey the change ? Imagine a user that wonders why the tab key behavior has changed, they should be able to spot the change quickly.

```
html: remove emmet TAB key binding for Emacs style

Allow to use TAB for indenting as usual.
```

ellakk:
I'm sorry, fixed it.

sdwolfz:
Thank you :heart:!

PR #11123, opened 2018-08-03
Make spacemacs-light/dark support pdf-view-midnight-minor-mode

yanghaoxie:
The `pdf-tools` package has a "midnight mode" with bright text on dark background (`pdf-view-midnight-minor-mode`). Using `spacemacs-light/dark` colors makes it look nicer.

### DIFF
--- a/core/libs/spacemacs-theme/spacemacs-common.el
+++ b/core/libs/spacemacs-theme/spacemacs-common.el
@@ -915,7 +915,10 @@ to 'auto, tags may not be properly aligned. "
 
     (custom-theme-set-variables
      theme-name
-     `(ansi-color-names-vector [,bg4 ,red ,green ,yellow ,blue ,magenta ,cyan ,base]))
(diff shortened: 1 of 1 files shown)


sdwolfz:
@yanghaoxie this change should be made in the upstream repository first: https://github.com/nashamri/spacemacs-theme and after that we will take the changes and apply them here.

yanghaoxie:
Thanks for point out that!
I saw the source code of `spacemacs-theme` in `.emacs.d/core/lib`, which makes me think the spacemacs-theme is bounded to spacemacs.
I will try to create a pull request there.

sdwolfz:
It is vendored so we have it as a theme initially before packages are being downloaded. Maybe we should make it more clear where to update the theme in the package itself.

sdwolfz:
@yanghaoxie have you managed to propose this upstream? I will close this PR since the theme will be updated with all the changes from upstream at once.

yanghaoxie:
Just propose to the upstream.
Sorry for the delay.
Thank you for close this PR!

PR #11135, opened 2018-08-05
go-mode set-tab-width conflict with editorconfig

lawrsp:
when i use editorconfig, I found it not work for go-mode
and then I found it because the go layer has a function spacemacs//go-set-tab-width
I think it should give me a way to choose to use the config in .editorconfig, so I add condition to the function
then I can define the layer variable to use editorconfig like this:
```elisp
(go :variables
    go-tab-width nil)
```

### DIFF
--- a/layers/+lang/go/funcs.el
+++ b/layers/+lang/go/funcs.el
@@ -11,7 +11,8 @@
 
 (defun spacemacs//go-set-tab-width ()
   "Set the tab width."
-  (setq-local tab-width go-tab-width))
+  (when go-tab-width
+    (setq-local tab-width go-tab-width)))
 
 (defun spacemacs//go-enable-gometalinter ()
    "Enable `flycheck-gometalinter' and disable overlapping `flycheck' linters."
--- a/layers/+lang/go/README.org
+++ b/layers/+lang/go/README.org
@@ -119,6 +119,13 @@ the layer variable =go-tab-width=, e.g.
   (go :variables go-tab-width 4)
 #+end_src
 
+If you're using =.editorconfig= in your project, set the value to nil to avoid
+confliction, e.g.
+
+#+begin_src emacs-lisp
+  (go :variables go-tab-width nil)
+#+end_src
+
 ** Tests
 If you're using =gocheck= in your project you can use the
 =go-use-gocheck-for-testing= variable to enable suite testing and to get single


layers/+lang/go/funcs.el
```diff
@@ -66,3 +67,4 @@
 (defun spacemacs/go-packages-gopkgs ()
   "Return a list of all Go packages, using `gopkgs'."
   (sort (process-lines "gopkgs") #'string<))
+
```
robbyoconnor: Remove this. No blank lines at the end of the file.

lawrsp: OK, done. sorry for that

robbyoconnor: No problem at all :)

lawrsp:
I tried to add a `go install after save` hook  for my issue [https://github.com/syl20bnr/spacemacs/issues/11139](url)
I just pushed to my fork, and now I found it appeared here automatically, 
maybe some one could check it and decide if it has value

Miciah:
A GitHub PR is a request to merge one branch into another branch (in this case your "develop" branch into the upstream "develop" branch), and since you pushed new commits to the branch from which you are requesting the merge, those new commits were added to this PR.  See the [Beginner's Guide to Contributing a Pull Request to Spacemacs](https://github.com/syl20bnr/spacemacs/wiki/Beginner's-Guide-to-Contributing-a-Pull-Request-to-Spacemacs), specifically the section [Creating a branch for your changes](https://github.com/syl20bnr/spacemacs/wiki/Beginner's-Guide-to-Contributing-a-Pull-Request-to-Spacemacs#creating-a-branch-for-your-changes).

layers/+lang/go/funcs.el
```diff
@@ -11,7 +11,8 @@
 
 (defun spacemacs//go-set-tab-width ()
   "Set the tab width."
-  (setq-local tab-width go-tab-width))
+  (if go-tab-width
+      (setq-local tab-width go-tab-width)))
```
Miciah: When you do not have an else clause, it is a good idea to use `(when a b)` instead of `(if a b)`.  Besides stylistic preference, it is all too easy for someone to change your `(if a b)` to `(if a b c)` (if a is true, then evaluate and return b, else evaluate and return c) when the intention was `(if a (progn b c))` (if a is true, then evaluate b and then evaluate and return c).

lawrsp: OK, updated, thank you ! 
Also thank you for the PR problem, I have revert my branch

lawrsp:
@Miciah  thank you 
I revert my branch, the extra patches gone, 
The only problem is the wrong reference to another issue, 
I edited my message , it is not work
I deleted it, is still not work

sdwolfz:
Thank you :heart:! 

lawrsp:
:)

Issue #11139, opened 2018-08-06
[go layer] auto completion did not work some times


lawrsp:
gocode works only with local package  and packages installed in pkg.
But usually, we have more than one package writing and they are not installed , then gocode will not work if you want a completion over these your own packages
I found 2 solutions:
1. add "-source" to company-go args
2. run go install after change the package code

I tried, and found solution-1  is too slowly, so I think 2 is better
The problem is , is there a quick way to run go install after we change the file ? I think 'auto after save' is a good choice.
Because I can not write elisp code,  is there any one can help ?

lawrsp:
use lsp-mode + gopls instead

PR #11140, opened 2018-08-07
go-layer add install after save

lawrsp:
for my issue #11139

### DIFF
--- a/layers/+lang/go/README.org
+++ b/layers/+lang/go/README.org
@@ -199,6 +199,18 @@ You have a few options to ensure you always get up to date suggestions:
   for you.
 - You can configure your task runner to run the =go install ./...= command on every
   file change.
+- Set =go-install-after-save= to have spacemacs run =go install --buildmode=archive= for
+  current pacakge after you saved a file:
+  #+begin_src emacs-lisp
+    (go :variables go-install-after-save t)
+  #+end_src
+  this behavier is asynchronous and slicent as default, you can choose the synchronous way to
+  get some result hint showed by message:
+  #+begin_src emacs-lisp
+    (go :variables go-install-after-save 'sync)
+  #+end_src
+  it also creates a buffer named `*GoInstall*` to record the command output details, both in
+  synchronous and asynchronous.
 
 ** LSP backend
 To enable the LSP backend, ensure that the =lsp= layer is enabled, and set the
--- a/layers/+lang/go/config.el
+++ b/layers/+lang/go/config.el
@@ -44,3 +44,6 @@ If `nil' then `go-mode' is the default backend unless `lsp' layer is used.")
 
 (defvar go-run-args ""
   "Additional arguments to by supplied to `go run` during runtime.")
+
+(defvar go-install-after-save nil
+  "Run `go install --buildmode=archive` after save. Possible values are `t' `sync' `nil'.")
--- a/layers/+lang/go/funcs.el
+++ b/layers/+lang/go/funcs.el
@@ -139,3 +139,17 @@
   "Set the tab width."
   (when go-tab-width
     (setq-local tab-width go-tab-width)))
+
+(defun spacemacs/go-install-after-save ()
+  "Install current package to pkg."
+  (interactive)
+  (when (eq major-mode 'go-mode)
+    (let ((errbuf (get-buffer-create "*GoInstall*"))
+          (install-args (list "install" "--buildmode=archive")))
+      (with-current-buffer errbuf
+        (erase-buffer))
+      (if (eq go-install-after-save 'sync)
+          (if (zerop (apply #'call-process "go" nil errbuf nil install-args))
+              (message "Applied go install")
+            (message "Apply go install failed"))
+        (apply #'start-process "go-install" errbuf "go" install-args)))))
--- a/layers/+lang/go/packages.el
+++ b/layers/+lang/go/packages.el
@@ -128,6 +128,8 @@
     (progn
       (when go-format-before-save
         (add-hook 'before-save-hook 'gofmt-before-save))
+      (when go-install-after-save
+        (add-hook 'after-save-hook 'spacemacs/go-install-after-save))
       (spacemacs/declare-prefix-for-mode 'go-mode "me" "playground")
       (spacemacs/declare-prefix-for-mode 'go-mode "mg" "goto")
       (spacemacs/declare-prefix-for-mode 'go-mode "mh" "help")


robbyoconnor:
You need to update the docs as well.

lawrsp:
@robbyoconnor
Ok, do you mean the file `README.org` in go layer dir?

robbyoconnor:
Yes.

robbyoconnor:
Also, pay attention to how it's formatted.

lawrsp:
done

lawrsp:
I saw this commit 09618503918f85ff9cb74821cce9f6d695399819 said, use gogetdoc then need not install the package on changed
I tried as the doc said, but it does not work for me 
do you tried that ?
@robbyoconnor

sdwolfz:
I would like someone that is more familiar than me with the golang echosystem to confirm if this is a good idea or not.

Also, this has the potentiall to block for a long period of time, do you think you can rewrite this to use an asynchronous process instead?

ljupchokotev:
Having an option to run go install automatically instead of doing it yourself manually is convenient. I am fine with having this in the go layer, but it should be rewritten as an asynchronous process before merging it.

lawrsp:
@sdwolfz 
I agree with you
but it is not simple for me to write lisp code
Could you give me some tips where can I find some examples with asynchronous process ?

lawrsp:
OK
I find out the `start-process` call, and changed it to  asynchronous.

but there is a defect with asynchronous:
 It cannot give the result of the command
Sometimes, I add some functions  in one package, save, then jump to another package where use it, and found it cannot complete the code , so I have to jump back to check the error, 
this happens especially when I use `go generate` tools  and the save point is not in the file really effected changed.

So I keep the sync way as an option:
```
install-after-save 'sync
```

And, I checked the action used here (`go install --buildmode=archive`):
it just compile the single one package independently, and it would not compile main package
as usually a package is not very big, I think the time it cost would be acceptable most of the time

layers/+lang/go/README.org
```diff
@@ -228,6 +228,12 @@ You have a few options to ensure you always get up to date suggestions:
   ./...= for you.
 - You can configure your task runner to run =go install ./...= command on every
   file change.
+- Set =go-install-after-save= to have spacemacs run =go install --buildmode=archive= for
+  current pacakge after you saved a file:
+
+  #+begin_src emacs-lisp
+  (go :variables go-install-after-save t)
```
sdwolfz: Please document all the possible values of that variable both here and in the config file.

layers/+lang/go/funcs.el
```diff
@@ -117,3 +117,15 @@
...
+  "Install current package to pkg."
+  (interactive)
+  (when (eq major-mode 'go-mode)
+    (let ((errbuf (get-buffer-create "*GoInstall*"))
+          (install-args (list "install" "--buildmode=archive")))
+      (if (eq go-install-after-save 'sync)
+          (if (zerop (apply #'call-process "go" nil errbuf nil install-args))
+              (message "Applied go install")
+            (message "Apply go install failed")))
+      (apply #'start-process "go-install" errbuf "go" install-args))))
```
sdwolfz: Shouldn't this `apply` be inside the `if` above? As I understand it setting the `go-install-after-save` to `sync` will execute the install command once synchronously and the second time asynchronously.

lawrsp: no, it is not necessary to run install twice, just choose sync or async

sdwolfz: Then `(apply #'start-process...` needs to be inside the `if`. Also instead of an if I would like to have a `cond` that checks for `eq go-install-after-save 'sync` and `eq go-install-after-save t` (or `async`) since there are 3 possible values for that variable, `nil` included.

lawrsp: I got it, you mean this if: 
```
      (if (eq go-install-after-save 'sync)
```
fixed.

and about the checks:
I think it should check nil before call this function
so, when called, it will do async as default, or sync as configured.
I don't know if double check is conversions, should I check nil here?

sdwolfz: I guess since the code is simple enough then the second nil check is not needed. It should be OK as it is right now.

layers/+lang/go/config.el
```diff
@@ -36,3 +36,6 @@
 
 (defvar go-use-test-args ""
   "Additional arguments to be supplied to `go test` during runtime.")
+
+(defvar go-install-after-save nil
+  "Run `go install --buildmode=archive` after save.")
```
sdwolfz: Please add the possible values to this docstring as well.

sdwolfz:
@ljupchokotev if everything looks good to you and if you approve this PR then I will merge it.

layers/+lang/go/config.el
lawrsp: ok, fixed

ljupchokotev:
@sdwolfz @lawrsp 
I tried this and it seems like I get different results if I run `go install --buildmode=archive` in the terminal in the currently opened project and if Spacemacs does it for me. Seems like when running it in Spacemacs it doesn't look into vendor packages and complains that it can't find the dependencies, but if I run it in the terminal it succeeds. Am I missing something here?

Also, if the asynchronous option is used, it should show a message if the process failed instead of not showing anything. In the synchronous option, the GoInstall buffer should be cleared each time go install runs after saving because it gets cluttered after the first time.

There is more to this as well. Currently, gocode doesn't work with vendored dependencies and it doesn't work with Go 1.11 at all and there are people involved in fixing it which should fix the original issue i.e. autocompletion doesn't work. I am not sure what should be done with this PR at this stage.

lawrsp:
@ljupchokotev 
I don't use vendor , so you may help to check it. maybe the ` .spacemacs.env `? Or if you use any auto install tools , give some informations how they works ?

I don't think async install should show the errors:
Because we use `gofmt` to detective  errors in the current buffer before save. If go install prompts , they would conflict. So, these error messages are only useful when the errors *does not* occur in *current* buffer. But if there are errors in other files, it may means you would edit them after then, unless you just used some generation tools, which I think is not a common use case.
So, to keep it simple,  there has a  sync method which only tell you there are errors, if you want your completion works fine, you could start a flycheck of buffer, go to check the GoInstall buffer to get the error file or. Or if you doesn't  finish the current package editing, just ignore it and continue to edit.
These cases are what I get when use it. I think you could give your use case why it should show error messages, then we could see.

And I agree with you it should clear the buffer, I will fix it.

Although gocode has these problems, there are many people use it and works fine. And this PR  only compares to the other ways and try to make an simple out-of-box way.

lawrsp:
@ljupchokotev @sdwolfz 
And I carefully checked showing the error message when gofmt runs successfully, I found it may good if provided in some cases. 
And I submit an issue to go-mode.el dominikh/go-mode.el#257 to check if this can be provided there, that will make it simple to implement to co-work with gofmt.

lawrsp:
@sdwolfz @ljupchokotev 
update:
I did some digging, and found `vscode` uses  `buildOnSave` flag to run `go build -i` to update the 
dependencies of the target, that will rebuild the imported packages which source are modified. Microsoft/vscode-go#673
Microsoft/vscode-go#338
Microsoft/vscode-go#953

And then I found the same mechanism in `flycheck`. There is a flag `flycheck-go-build-install-deps` , if it is enabled , it will add `-i ` to the `go-build` check. The `go-build` checker needs go >= 1.6

As in the threads they talked, `go install` would be faster because of its build cache, and `go build -i` would give build errors as a linter.

I suggest us using this option for a while to test if it is good, it just need a document update.
The `install` option can wait if you still think there are some questions

lawrsp:
we don't need gocode any more. lsp + gopls   is better

PR #11148, opened 2018-08-08
Support optional set-namespace argument to Clojure layer's `, s B`

emidln:
The spacemacs analogue of cider-load-buffer-and-switch-to-repl-buffer
doesn't accept an argument to set the namespace while you are switching
to the repl buffer. This adds support for an argument, invokable with
the universal argument, for setting up the namespace as well. This will
eliminate an extra `(in-ns ...)` or `, s n` call as long as you invoke
`, s B` prefixed by the universal argument: `SPC u , s B`.

### DIFF
--- a/layers/+lang/clojure/funcs.el
+++ b/layers/+lang/clojure/funcs.el
@@ -94,12 +94,12 @@ the focus."
   (cider-insert-ns-form-in-repl t)
   (evil-insert-state))
 
-(defun spacemacs/cider-send-buffer-in-repl-and-focus ()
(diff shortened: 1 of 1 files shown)


sdwolfz:
Can you please share some steps to reproduce your issue? I am not familiar with the functionality as I don't use clojure and would like to try it out locally before merging.

smile13241324:
I like clojure :), lets take that as an oportunity to revisit this layer and see whether we do not yet have that functionality. If not I will rebase this PR.

smile13241324:

Issue #11152, opened 2018-08-09
Counsel projectile (wrong-type-argument arrayp nil) error


FieryCod:
#### Description :octocat:
Counsel projectile stopped working with the newest version of packages.

#### Reproduction guide :beetle:
- Start Emacs
- SPC p f (or any of counsel-projectile command)

*Observed behaviour:* :eyes: :broken_heart:
Every counsel projectile command not work

*Expected behaviour:* :heart: :smile:
Counsel projectile commands should work

#### System Info :computer:
- OS: darwin
- Emacs: 26.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 9af8e1564)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
((auto-completion :variables auto-completion-enable-snippets-in-popup nil auto-completion-return-key-behavior nil auto-completion-enable-help-tooltip t auto-completion-enable-sort-by-usage t auto-completion-tab-key-behavior 'complete spacemacs-default-company-backends
                  '(company-keywords company-files company-dabbrev-code company-dabbrev company-css company-web-html))
 (python :variables python-backend 'anaconda)
 (c-c++ :variables c-c++-default-mode-for-headers 'c++-mode c-c++-enable-clang-support t)
 (git :variables git-magit-status-fullscreen t)
 (version-control :variables version-control-diff-tool 'diff-hl version-control-global-margin t)
 (org :variables org-enable-github-support t)
 (shell :variables shell-default-height 35 shell-default-position 'bottom shell-default-term-shell "/bin/bash")
 colors markdown osx docker php javascript evil-commentary csv better-defaults clojure go html github emacs-lisp syntax-checking neotree yaml parinfer plantuml ruby sql typescript gtags ivy)
```
- System configuration features: JPEG RSVG IMAGEMAGICK NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS MODULES THREADS LCMS2

#### Backtrace :paw_prints:
```
Debugger entered--Lisp error: (wrong-type-argument arrayp nil)
  define-key((keymap) nil counsel-projectile-command-map)
[...]

PR #11153, opened 2018-08-09
Fix 'projectile-keymap-prefix for counsel-projectile

FieryCod:
The newest commit in projectile breaks compatibility of projectile-counsel because of fact that it set's the projectile-keymap-prefix to nil. Projectile-counsel uses that variable, therefore, it needs to be set.

This PR addresses the issue #11152

### DIFF
--- a/layers/+spacemacs/spacemacs-project/config.el
+++ b/layers/+spacemacs/spacemacs-project/config.el
@@ -1,0 +1,21 @@
+;;; config.el --- Projectile Layer Configuration File for Spacemacs
+;;
+;; Copyright (c) 2012-2018 Sylvain Benner & Contributors
+;;
+;; Author: Sylvain Benner <sylvain.benner@gmail.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+
+;; Layer Variables
+
+(defvar projectile-keymap-prefix (kbd "C-c C-p")
+  "Default projectile-keymap-prefix.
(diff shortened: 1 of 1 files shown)


layers/+completion/ivy/config.el
```diff
@@ -47,6 +47,11 @@ than this amount.")
     ("r" spacemacs/rename-file "rename file"))
   "Default ivy actions for files.")
 
+(defvar projectile-keymap-prefix (kbd "C-c C-p")
```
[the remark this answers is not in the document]
FieryCod: @sdwolfz Thank you very much for your comment. Did a fix. Please check whether it's ok now.

layers/+spacemacs/spacemacs-project/config.el
```diff
@@ -0,0 +1,21 @@
...
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+

+;; Layer Variables
+
+(defvar projectile-keymap-prefix (kbd "C-c C-p")
```
aaronjensen: This should be a `setq` I would think. That said, I'd think that something like https://github.com/ericdanan/counsel-projectile/pull/92 is the right fix. Then the decision as to what the projectile keymap prefix is in spacemacs (if any) is a separate decision. Maybe it should be `C-c C-p`, but I don't know how @syl20bnr thinks about things like this.

aaronjensen: @seagle0128 could you elaborate? I don't see a commit in projectile or spacemacs setting the prefix to `C-c p`, but maybe I'm missing something?

PR #11154, opened 2018-08-09
core: Add support for lazy installation of the systemd layer

mineo:
The list of supported file extensions is taken from
https://github.com/holomorph/systemd-mode/blob/401d71c2dd24e424216ae5e4275c830f2a9c6b0c/systemd.el#L391-L392.

### DIFF
--- a/layers/auto-layer.el
+++ b/layers/auto-layer.el
@@ -74,6 +74,7 @@
 (configuration-layer/lazy-install 'sml :extensions '("\\(\\.s\\(ml\\|ig\\)\\'\\|\\.\\(sml\\|sig\\)\\'\\)" sml-mode))
 (configuration-layer/lazy-install 'sql :extensions '("\\(\\.sql\\'\\)" sql-mode))
 (configuration-layer/lazy-install 'swift :extensions '("\\(\\.swift\\'\\)" swift-mode))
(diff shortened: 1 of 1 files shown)


layers/auto-layer.el
```diff
@@ -74,6 +74,7 @@
 (configuration-layer/lazy-install 'sml :extensions '("\\(\\.s\\(ml\\|ig\\)\\'\\|\\.\\(sml\\|sig\\)\\'\\)" sml-mode))
 (configuration-layer/lazy-install 'sql :extensions '("\\(\\.sql\\'\\)" sql-mode))
 (configuration-layer/lazy-install 'swift :extensions '("\\(\\.swift\\'\\)" swift-mode))
+(configuration-layer/lazy-install 'systemd :extensions'("\\(\\.\\(nspawn\\|automount\\|busname\\|mount\\|service\\|slice\\|socket\\|swap\\|target\\|timer\\|link\\|netdev\\|network\\)\\)" systemd-mode))
```
sdwolfz: This regexp matches strings inside a filename, it should match it only at the end of a filename.

mineo: You're right, it was missing the `\\'`. Fixed.

sdwolfz:
Thank you :heart:!

PR #11162, opened 2018-08-11
emacs-lisp: Add company-elisp to company backends

delaanthonio:
`company-elisp` generally has more helpful suggestions than `company-capf`. The suggestions come much faster than semantic.

### DIFF
--- a/layers/+lang/emacs-lisp/packages.el
+++ b/layers/+lang/emacs-lisp/packages.el
@@ -54,7 +54,7 @@
           (lisp-indent-line))))))
 
 (defun emacs-lisp/post-init-company ()
-  (spacemacs|add-company-backends :backends company-capf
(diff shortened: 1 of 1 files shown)


Compro-Prasad:
I think I already get the elisp completion. What does this PR do?

sdwolfz:
I am not familiar with the differences between the two, hopefully the elisp comunity can jump in and give some insight into this :smile:.

yqrashawn:
> ;; In newer versions of Emacs, company-capf is used instead.

[company-elisp.el line 25](https://github.com/company-mode/company-mode/blob/c95a6b41d621de4253b77e512aa61fc0e75acddc/company-elisp.el#L25)

User can have fuzzy match with company-capf and company-flx. `company-elisp` will break this. I don't think that's what we want.

PR #11163, opened 2018-08-11
Fix Elfeed not saving on quit

DiGitHubCap:
This binding is redundant since Elfeed already binds q to elfeed-search-quit-window.

### DIFF
--- a/layers/+readers/elfeed/packages.el
+++ b/layers/+readers/elfeed/packages.el
@@ -31,14 +31,12 @@
         "gR" 'elfeed-search-update--force
         "gu" 'elfeed-unjam
         "o"  'elfeed-load-opml
-        "q"  'quit-window
         "w"  'elfeed-web-start
         "W"  'elfeed-web-stop)
       (evilified-state-evilify-map elfeed-show-mode-map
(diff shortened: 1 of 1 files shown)


Miciah:
Using the package's own quit function generally seems like a good idea.  Any reason not to make the same change to `elfeed-show-mode-map` (which by default binds <kbd>q</kbd> to `elfeed-kill-window`)?

DiGitHubCap:
Not that I can see. I've just changed it.

sdwolfz:
I would preffer having `elfeed-kill-window` explicitly bound to `q` in the config, since it's also documented in the `README`. Having it be changed under the hood is unlikely but it might be confising to se a binding documented but not appearing in the code.

sdwolfz:
Thank you :heart:!

I don't feel to strong about having the keybinding explicit or not. This looks like an improvement and there is no need to delay it further.

DiGitHubCap:
Sorry, I must have missed that first email! As long as you're okay with it, I guess it's fine. It also seems unlikely to me that the binding would change in Elfeed, but if it ever does, I can just update the layer (assuming I'm still using Elfeed then).

Issue #11073, opened 2018-07-22
Flycheck disabled by default


jorgemorais:
#### Description :octocat:

`syntax-checking` layer is enabled; `syntax-checking-enable-by-default` is t; but
Flycheck starts disabled

#### Reproduction guide :beetle:

- Enable layer `syntax-checking` in `dotspacemacs-configuration-layers`
- Restart Emacs
- Open a Bash script
- SPC e v

*Observed behaviour:* :eyes: :broken_heart:

Flycheck is disabled.  I have reported this with my full Spacemacs config, but I have also reproduced it with an almost-vanilla config.

*Expected behaviour:* :heart: :smile:

Flycheck should be enabled by default, according to the layer's `README.org`.

Workaround: set `flycheck-global-modes` to t.  Unfortunately this needs to be
done in every Emacs session; just setting it in `.spacemacs` is not enough,
because layer code overrides it – which is probably a bug in itself.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 26.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. d9eecb760)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: helm
- Layers:
```elisp
((chrome :packages
         (not gmail-message-mode))
 csv erc helm
 (auto-completion :variables auto-completion-enable-help-tooltip t auto-completion-private-snippets-directory
                  (concat spacemacs-private-directory "shared_snippets")
                  auto-completion-enable-sort-by-usage t)
 (elfeed :variables rmh-elfeed-org-files
         `(,(concat spacemacs-private-directory "elfeed.org")))
 emacs-lisp git gnus json markdown neotree
 (notmuch :variables notmuch-message-deleted-tags
          '("+deleted"))
 org
 (shell :variables shell-default-shell 'shell)
 php shell-scripts spell-checking syntax-checking)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND DBUS GSETTINGS NOTIFY ACL GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 THREADS

Miciah:
I could not reproduce the issue.

What is the value of `flycheck-global-modes`? For me, it is `(json-mode php-mode sh-mode)`.  Each enabled layer that uses Flycheck should add itself to this list.

[...]

PR #11169, opened 2018-08-12
doc: revise docs for auto-complete and syntax-checking layers

smile13241324:
Syntax-checking and auto-complete layers should more directly communicate that their function depends on the installation of individual language layers to avoid misunderstandings with new users who only install syntax-checking or auto-complete and install the packages manually.

This is part of #11073.


sdwolfz:
Thank you :heart:!

PR #11178, opened 2018-08-15
search-engine: add ecosia and make each engine use https

smile13241324:
Added "Ecosia - the search engine to plant trees" to search-engine layer. Also I have updated the existing definitions to use HTTPS instead of HTTP when supported by the search provider.


sdwolfz:
Thank you :heart:!

PR #11183, opened 2018-08-16
Add support for golangci-linter to go layer

winmillwill:
Mostly just a copy-paste of the support for gometalinter using these things:

https://github.com/golangci/golangci-lint
https://github.com/weijiangan/flycheck-golangci-lint


winmillwill:
Agree.

PR #11185, opened 2018-08-17
Delete compilation finish hook

Miciah:
Before this commit, Spacemacs assigned an anonymous function to the `compilation-finish-function` variable.  This variable was deprecated in Emacs 22.1 and removed in Emacs 27.1 in favor of the
`compilation-finish-functions` hook.  Consequently, the assignment has no practical effect in Emacs 27.1.

Although we could add the Spacemacs function to the new `compilation-finish-functions` hook, the function serves little purpose; it prints a generic message when compilation succeeds:

    compilation ok.

and it prints the key binding for `spacemacs/next-error` when compilation fails:

    There were errors. Use SPC e n to visit.

However Emacs already prints messages to indicate success or failure, respectively:

    Compilation finished

or (for example)

    Compilation exited abnormally with code 1

Because the Spacemacs function prints its "There were errors" message immediately after Emacs prints its "Compilation exited abnormally" message, the latter was obscured.  Because the exit code in Emacs's message could be useful and the added value of the Spacemacs function's message is minimal, we can simply delete the Spacemacs function.

* `layers/+spacemacs/spacemacs-defaults/funcs.el` (`compilation-finish-function`): Delete the assignment and anonymous function.

---

This PR supersedes #11059.

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -1228,23 +1228,6 @@ using a visual block/rectangle selection."
 
 ;; END linum mouse helpers
 
-;; From http://xugx2007.blogspot.ca/2007/06/benjamin-rutts-emacs-c-development-tips.html
-(setq compilation-finish-function
-      (lambda (buf str)
-
-        (let ((case-fold-search nil))
-          (if (or (string-match "exited abnormally" str)
-                  (string-match "FAILED" (buffer-string)))
-
-              ;; there were errors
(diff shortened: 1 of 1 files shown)


Compro-Prasad:
Then how to suggest the keybind? Isn't it useful?

sdwolfz:
Honestly I don't think we should suggest a keybinding here. I'm going to merge this PR unless someone has any objections.

sdwolfz:
Thank you :heart:!

Waited for two weeks and it seems no one is objecting so I'll merge this. In case someone complains we'll see how we handle it.

PR #11194, opened 2018-08-21
Clojure layer: make sayid optional

### LINKED ISSUE #11146: Remove sayid from clojure layer [development]
With recent changes to CIDER and nREPL, sayid no longer works. The 0.300 development branch currently installs it and injects it on `cider-jack-in` by default, causing a number of issues for unsuspecting users of the Clojure layer.

I have raised an issue on the github project, however the last commit to sayid was over a year ago: https://github.com/bpiel/sayid.

It's safe to say that it's at risk for being abandoned and, IMO, should be removed from the defaults that the Clojure layer comes with.

plexus:
The sayid debugger and profiler for Clojure currently has compatibility issues with other packages. It's also a package that only a small percentage of users will use, while having it present slows down Clojure startup for everyone.

The author has agreed to address the compatibility issues, so this will still be useful for people in the future, but it shouldn't be installed by default.

Rather than fully remove it as suggested in #11146, this PR makes it optional using the same pattern used in other layers.

Closes #11146

### DIFF
--- a/layers/+lang/clojure/config.el
+++ b/layers/+lang/clojure/config.el
@@ -18,4 +18,7 @@
 (spacemacs|define-jump-handlers cider-repl-mode)
 
 (defvar clojure-enable-fancify-symbols nil
-  "If non nil the `fancify-symbols' function is enabled.")
+  "If non-nil, the `fancify-symbols' function is enabled.")
+
+(defvar clojure-enable-sayid nil
+  "If non-nil, the Sayid Clojure debugger is enabled.")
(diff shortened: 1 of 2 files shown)


sdwolfz:
Thank you :heart:! 

bbatsov:
@plexus Mind doing this for `clj-refactor` as well? Based on the numerous bug reports I've received over the years it seems that most Spacemacs users don't even know what clj-refactor is, but are often bit by it being out of sync with CIDER. I'd rather disable it by default, so only people who know what they're doing have to deal with the potential problems that might arise if it's enabled.

PR #11200, opened 2018-08-23
Advise rename-buffer when persp-mode activated

Miciah:
Before this commit, `spacemacs-layouts/init-eyebrowse` advised `rename-buffer` with `spacemacs//fixup-window-config`.  However, this advice causes problems if persp-mode is not activated when a function tries to rename a buffer.

In particular, Spacemacs's CI uses Emacs's batch mode to run `spacemacs/publish-doc`, which uses `rename-buffer`.  In batch mode, persp-mode is not activated, and consequently `spacemacs/publish-doc` was failing:

    Publishing file /home/travis/.emacs.d/news/news01.org using ‘org-html-publish-to-html’
    Skipping check for new version (reason: dotfile)
    Wrong type argument: hash-table-p, nil
    Warning (emacs): recentf mode: Wrong type argument: hash-table-p, nil
    spacemacs/publish-doc failed

In order to avoid problems, this commit moves the advising of `rename-buffer` from `spacemacs-layouts/init-eyebrowse` to `spacemacs//activate-persp-mode`.

* `layers/+spacemacs/spacemacs-layouts/packages.el` (`spacemacs-layouts/init-eyebrowse`): Move advice from here...
* `layers/+spacemacs/spacemacs-layouts/funcs.el` (`spacemacs//activate-persp-mode`): ...to here.

### DIFF
--- a/layers/+spacemacs/spacemacs-layouts/funcs.el
+++ b/layers/+spacemacs/spacemacs-layouts/funcs.el
@@ -17,7 +17,16 @@
  (e.g. don't re-activate during `dotspacemacs/sync-configuration-layers' -
  see issues #5925 and #3875)"
   (unless (bound-and-true-p persp-mode)
-    (persp-mode)))
+    (persp-mode)
+    ;; eyebrowse's advice for rename-buffer only updates workspace window
+    ;; configurations that are stored in frame properties, but Spacemacs's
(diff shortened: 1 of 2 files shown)


sdwolfz:
Thank you :heart:!

PR #11203, opened 2018-08-23
lsp-layer core keybinding amendments

cormacc:
`m g a` conventionally used to switch between c/c++ source and header files, but bound to a keyboard symbol navigation function in lsp-layer. Moved lsp-layer binding under `m g k`

### DIFF
--- a/layers/+tools/lsp/README.org
+++ b/layers/+tools/lsp/README.org
@@ -6,6 +6,8 @@
 - [[#configuration][Configuration]]
   - [[#derived-layers][Derived layers]]
     - [[#spacemacslsp-bind-keys-for-mode-mode][=spacemacs/lsp-bind-keys-for-mode mode=]]
+      - [[#declared-prefixes][Declared prefixes]]
+      - [[#default-keybindings][Default keybindings]]
   - [[#variables][Variables]]
   - [[#diagnostics][Diagnostics]]
 - [[#future-additionsimprovements][Future additions/improvements]]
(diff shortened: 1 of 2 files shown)


cormacc:
@yyoncho suggested migrating hierarchy-related keybindings from `m h ..` to `m H ..` on closed PR #10486 as `m h ..` conventionally help-related.

Also moving the prefix declaration from lsp-layer to c-c++ layer. I'd be inclined to move `m h .. ` to `m H ..` as suggested but to leave the prefix declaration in the lsp layer as it's applicable to more than just c++...

yyoncho:
Make sense to me.

cormacc:
@yyoncho has also suggested moving the `execute-code-action` binding from `m l  a` to `m r e a`.

I'm not 100% convinced of this for two reasons:
1. Code actions are not necessarily refactorings -- at least in my limited experience. What I've seen offered by `ccls` are clang FixIt suggestions (https://clang.llvm.org/diagnostics.html), which I'd consider corrections rather than refactorings. Although a code action could also be a refactoring, depending on the -- it seems like a broad umbrella.
2. Regardless of the specificity argument, 'm r e a' is a little long-winded, and I'd have thought the `m r e` sequence a better fit for `extract` type refactorings. 

That's not to say that `m l a` is the right binding, just that `m r e a` / `m r a` seems narrow. Maybe bind twice :)

yyoncho:
My main concern about having 3 keys like mra is that some languages like java have a lot refactorings. For example, mrai is "add import" in lsp-java layer. The list of java refactorings could go to 40+ and if the base layer is taking mra/mrr upper layers will be forced to use non mnemonical (long) shortcuts.

cormacc:
Hierarchy relates to structure -- a subclass is lower in the class hierarchy than its base class etc.

The single binding you mention under `h` in the lsp layer fits within the `help` category.
Some bindings in the `lsp-c-c++` layer used `h` in the `hierarchy` sense -- now bound under `H` instead -- see below:

``` elisp
    ;; hierarchy
    "Hb" #'c-c++/base ;;Replace this with lsp-goto-implementation in lsp-layer?
    "Hd" #'c-c++/derived
    "Hc" #'c-c++/call-hierarchy
    "HC" #'c-c++/call-hierarchy-inv
    "Hi" #'c-c++/inheritance-hierarchy
    "HI" #'c-c++/inheritance-hierarchy-inv
    "Hm" #'c-c++/member-hierarchy
```

I don't really use any of these bindings myself, as it's pretty much all embedded C I do. Open to suggestions for an alternative to `H`...

yyoncho:
An alternative approach is to use binding under "g" since hierarchy is kind of similar to references (e. g. "ghc").

sdwolfz:
I like the idea of putting it under `, g h`, but I don't think we're going to have many thing s to put under `, h` in the first place, so maybe it's reasonable to leave them there. The current java layer has them like that already.

Let's change the prefix name to `help/hierarchy`.

cormacc:
Commit amended as per discussion. Also documented the declared prefixes in layer `README.org`.

yyoncho:
I have some considerations related to `goto` functionality. The `lsp` layer bind the navigation functions that are coming from `lsp-mode` in favor of `lsp-ui` mode one(e. g. `lsp-goto-implementation` vs `lsp-ui-peek-find-implementation` . I personally prefer the `lsp-mode` native one since they are faster and provide an option to filter the results. I suggest either providing a way to chose which one to use(e. g. having flag `lsp-layer-use-lsp-ui`) or have separate bindings for `lsp-mode` one. In `lsp-java` layer I am using "mgi" for lsp-mode one and "mgI" for lsp-ui one. What do you think?

cormacc:
Re. goto keybindings, I've added the missing binding for `lsp-ui-peek-find-implementation` under `mgl` as you suggest. I had thought about what you say about `lsp-mode` vs `lsp-ui-peek` bindings a while back -- that's what prompted me to initially bind some of the more ui-heavy goto functions under a separate prefix (`j` -- which had been `p` for peek at an earlier point). But in retrospect that was a fairly arbitrary distinction on the basis of implementation-specifics rather than function...

Think that (i.e. the `m g l` keybinding) may be the less confusing option for now.

yyoncho:
@cormacc just one minor nit it is I(capital i) not l(lowercase L) .
Here it is the full list for `lsp-java` - https://github.com/syl20bnr/spacemacs/tree/develop/layers/+lang/java#lsp-java-1 . I will sync up them once your PR is in. The convention used in `lsp-java` is problematic too since typically the capital letter means "act on thing under cursor". As a side note, I liked the previous approach(I saw it in some of the previous revisions and I thought that this version is in). - bind `lsp-ui` under a separate shortcut.

sdwolfz:
Thank you :heart:!

cormacc:
@yyoncho Saw your 'l' vs capitalised 'i' message too late there I'm afraid. Re. separate prefix for lsp-ui-peek bindings, eliminated that based on feedback from @sdwolfz during the merge of the early PR.

Could re-introduce it as a configurable option, e.g. 
``` elisp
(lsp :variables lsp-ui-alternate-keybindings t)
```
and then bind under `m p` or `m j` (as in 'peek' or 'jump')?

Anyway I've started on the c-c++ layer lsp backend integration, so won't be raising any more PRs against lsp layer before that's done (I'm sure @sdwolfz will be glad to hear)

yyoncho:
@cormacc no problems, I have pending request against `lsp-mode` for multiroot support, once it is merged I will do a change against `lsp` layers so I could revisit the keybindings.

cormacc:
Great!

PR #11208, opened 2018-08-25
Adds j/k  to zoom transitions

agzam:
This is small but meaningful change. I think it makes sense.

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/keybindings.el
+++ b/layers/+spacemacs/spacemacs-defaults/keybindings.el
@@ -716,13 +716,16 @@ otherwise it is scaled down."
 
 (spacemacs|define-transient-state scale-font
   :title "Font Scaling Transient State"
-  :doc "\n[_+_/_=_] scale up [_-_] scale down [_0_] reset font [_q_] quit"
+  :doc "\n[_+_/_=_/_k_] scale up [_-_/_j_] scale down [_0_] reset font [_q_] quit"
   :bindings
   ("+" spacemacs/scale-up-font)
+  ("k" spacemacs/scale-up-font)
   ("=" spacemacs/scale-up-font)
   ("-" spacemacs/scale-down-font)
+  ("j" spacemacs/scale-down-font)
   ("0" spacemacs/reset-font-size)
   ("q" nil :exit t))
+
 (spacemacs/set-leader-keys "zx" 'spacemacs/scale-font-transient-state/body)
 
 ;; end of Text Manipulation Transient State
@@ -781,11 +784,13 @@ If FRAME is nil, it defaults to the selected frame."
 
 (spacemacs|define-transient-state scale-transparency
   :title "Frame Transparency Transient State"
-  :doc "\n[_+_/_=_] increase transparency [_-_] decrease [_T_] toggle [_q_] quit"
+  :doc "\n[_+_/_=_/_k_] increase transparency [_-_/_j_] decrease [_T_] toggle [_q_] quit"
   :bindings
   ("+" spacemacs/increase-transparency)
+  ("k" spacemacs/increase-transparency)
   ("=" spacemacs/increase-transparency)
   ("-" spacemacs/decrease-transparency)
+  ("j" spacemacs/decrease-transparency)
   ("T" spacemacs/toggle-transparency)
   ("q" nil :exit t))
 (spacemacs/set-leader-keys "TT"
--- a/layers/+spacemacs/spacemacs-visual/packages.el
+++ b/layers/+spacemacs/spacemacs-visual/packages.el
@@ -93,12 +93,16 @@
       (spacemacs|define-transient-state zoom-frm
         :title "Zoom Frame Transient State"
         :doc "
-[_+_/_=_] zoom frame in [_-_] zoom frame out [_0_] reset zoom [_q_] quit"
+[_+_/_=_/_k_] zoom frame in [_-_/_j_] zoom frame out [_0_] reset zoom [_m_] max frame [_f_] fullscreen [_q_] quit"
         :bindings
         ("+" spacemacs/zoom-frm-in)
+        ("k" spacemacs/zoom-frm-in)
         ("=" spacemacs/zoom-frm-in)
         ("-" spacemacs/zoom-frm-out)
+        ("j" spacemacs/zoom-frm-out)
         ("0" spacemacs/zoom-frm-unzoom)
+        ("f" spacemacs/toggle-frame-fullscreen-non-native)
+        ("m" spacemacs/toggle-maximize-frame)
         ("q" nil :exit t))
       (spacemacs/set-leader-keys "zf" 'spacemacs/zoom-frm-transient-state/body)
 


robbyoconnor:
Code duplication is bad.

agzam:
@robbyoconnor yes, it repeats the same function calls in favor to duplicate `+/-` to `j/k`.

I think it makes sense in transients to use home row buttons (wherever that's possible). In this specific transient I don't think it makes sense to break existing keybindings. If there is a more idiomatic way to bind same function to different keys, please show me - I'd be happy to make adjustments. Otherwise, I think this small code duplication is harmless and pros of convenience outweigh the cons.

agzam:
@robbyoconnor here's an example of similar code:

https://github.com/syl20bnr/spacemacs/blob/develop/layers/+spacemacs/spacemacs-layouts/packages.el#L186

see? I can say: "I have literally done it by the book"

robbyoconnor:
I think you could combine the keybinding into a one-liner -- my elisp is weak but I've seen it done elsewhere in spacemacs

jr0cket:
@robbyoconnor having looked through the Emacs wiki on keymaps, I have not found a way to define multiple key sequences to a single function, to create the one-liner you are suggesting.  The extensive documentation did suggest that may be possible is to use one keymap to define another although this failed when I tested the following code changes:
```
:bindings
  ("+" spacemacs/scale-up-font)
  ("j" "+")
```
or 
```
  :bindings
  ("+" spacemacs/scale-up-font)
  ("j" (kbd "+"))
```

If anyone knows how to create multiple key sequences for one function all within one readable key binding expression, I would appreciate knowing how this is done.

The convention of simply adding a second or third keybinding is quite common in Spacemacs.  It is especially used when you have convienience keybinings, such as `=` for =`+` or `N` for `p`.

Having hunted through the spacemacs code base I have not found any examples of a one-liner for multiple key sequences.  As I am not sure what that looks like so tricky to search for.

Both the files included in this PR already had keybindings that call the same function (`+` `=`) to avoid the need to press shift key when zooming.  

Another extensive example of multiple keybindings is in the spacemacs defaults layer:
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+spacemacs/spacemacs-defaults/keybindings.el#L656-L671

Other examples I have found include:

https://github.com/syl20bnr/spacemacs/blob/develop/layers/+distributions/spacemacs-bootstrap/packages.el#L152-L155
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+spacemacs/spacemacs-layouts/packages.el#L69-L81
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+spacemacs/spacemacs-layouts/packages.el#L182-L201
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+spacemacs/spacemacs-completion/packages.el#L90-L103
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+spacemacs/spacemacs-completion/packages.el#L163-L174
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+spacemacs/spacemacs-completion/packages.el#L256-L267
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+spacemacs/spacemacs-navigation/packages.el#L135-L136
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+spacemacs/spacemacs-editing/packages.el#L258-L262
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+spacemacs/spacemacs-editing/packages.el#L398-L401
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+spacemacs/spacemacs-visual/packages.el#L98-L99
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+spacemacs/spacemacs-defaults/keybindings.el#L162-L163
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+spacemacs/spacemacs-defaults/keybindings.el#L171-L172
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+spacemacs/spacemacs-defaults/keybindings.el#L197-L198
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+spacemacs/spacemacs-defaults/keybindings.el#L559-L564
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+spacemacs/spacemacs-defaults/keybindings.el#L787-L790
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+spacemacs/spacemacs-evil/packages.el#L303-L309
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+source-control/version-control/packages.el#L284-L285
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+source-control/git/packages.el#L100-L102

I would apprecate your thoughts on this investigation and how to proceed next.

Thank you
John.
cc @agzam

agzam:
oh wow @jr0cket, it seems you have done more work on this tiny PR than I ever thought it'd require. Can you think of any other transients where having home-row buttons would make sense?

jr0cket:
@agzam 
It makes sense for the zoom based transient states and there is `spacemacs|define-transient-state scale-transparency`.  This uses the `+`, `=` and `-` bindings for scaling and adding `j` and `k` would make sense here.
```(spacemacs|define-transient-state scale-transparency
  :bindings
  ("+" spacemacs/increase-transparency)
  ("=" spacemacs/increase-transparency)
  ("-" spacemacs/decrease-transparency)
```
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+spacemacs/spacemacs-defaults/keybindings.el#L787-L790

I dont think it makes sense to add `jk` to the transient states that use `n`, `N`, `p` for getting the next an previous thing (buffer, symbol, hunk, etc). 

I also considered `spacemacs|define-transient-state evil-numbers`, however, you enter that transient state with `-` or `+` so it feels less natural to switch to `j` and `k` as your fingers are already way from the center.
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+spacemacs/spacemacs-evil/packages.el#L303-L309

sdwolfz:
These changes make sense to me, but let's discuss other transient states when the opportunity arises.

One more change I would ask for: have `_` decrease just like `-` does (complementary to `=` and `+`). 

Emacs has the `C-x C-+` functionality these transient states were inspired from and it's also missing the `_` keybinding which makes forgetting to release shift a pain while using that, I was meaning to add this to the core myself but I have not gotten around to sending the patch yet (and signing the copyright documents).

sdwolfz:
Tried it out for a bit. It feels weird to have `j` scale up since `j` moves a line down, I was expecting the size to decrease with it, similar for `k`.

I think these should be reversed.

agzam:
but, but it's like: "bring it down to me" kind of way, no? Oh, oh, maybe we can have a var `zoom-type-australian` or something? ;)

sdwolfz:
YES! `zoom-type-australian` would be amazing! :australia:

sdwolfz:
It took me a while to get used to the different scrolling behavior on macbooks compared to normal laptops as well. I always assumed the scroll was like dragging the scroll bar and not dragging the page. Maybe there's a similar thing happening here, I'm not viewing it as an interaction with the window but with the controls. 

Anyway I would like to hear more feedback related to this from other people.

layers/+spacemacs/spacemacs-defaults/keybindings.el
```diff
@@ -714,13 +714,16 @@ otherwise it is scaled down."
...
 (spacemacs|define-transient-state scale-font
   :title "Font Scaling Transient State"
-  :doc "\n[_+_/_=_] scale up [_-_] scale down [_0_] reset font [_q_] quit"
+  :doc "\n[_+_/_=_/_j_] scale up [_-_/_k_] scale down [_0_] reset font [_q_] quit"
   :bindings
   ("+" spacemacs/scale-up-font)
+  ("j" spacemacs/scale-up-font)
   ("=" spacemacs/scale-up-font)
   ("-" spacemacs/scale-down-font)
+  ("k" spacemacs/scale-down-font)
```
robbyoconnor:
```elisp
(dolist (key '("+"
                   "j"))
       (spacemacs/scale-up-font)) 
```

And similar for scale-down as well :)

robbyoconnor:
I wonder if we could use `(dolist)` to DRY this up a bit?

Miciah:
For what it's worth, in [feh](https://feh.finalrewind.org/), <kbd>Up</kbd> zooms in, and <kbd>Down</kbd> zooms out.  In Firefox, Chrome, Evince, and eog, mouse-wheel up zooms in, and mouse-wheel down zooms out.

agzam:
Can this be merged?

duianto:
@agzam Do these changes look ok?

I pushed the commit to my fork as a preview: https://github.com/duianto/spacemacs/commit/708f4661bf92aa7f317c34510a7d7a83178822db

#### Updated:
- changelog.develop
- commit
  - subject: `Add j and k to transient states`
(since it isn't zoom TS only)
  - message: included the key bindings
- Zoom Frame Transient State
Arranged the keys as two columns instead of one long line, now it doesn't extend past the right edge of a default emacs frame size (80 chars)

#### Before

before


#### After

after

#### Added documentation:
- A new transparency section under `Gui Elements > Frame > Transparency`
(also made the heading `Frame Title` a sub section `Gui Elements > Frame > Title`)
https://github.com/duianto/spacemacs/blob/testing-j/k-keys-in-transient-states/doc/DOCUMENTATION.org#transparency
- Text scale bindings
https://github.com/duianto/spacemacs/blob/testing-j/k-keys-in-transient-states/doc/DOCUMENTATION.org#text
- Zoom Frame bindings
https://github.com/duianto/spacemacs/blob/testing-j/k-keys-in-transient-states/doc/DOCUMENTATION.org#frame-1

The text scale and zoom frame bindings are comma separated so that only one description line is needed.

jr0cket:
Looks great to me, thanks for your work on this

agzam:
Very nice 👌

duianto:

agzam:
@duianto I have noticed that `(spacemacs/zoom-frm-in)` works differently. Unlike `zoom-frm-in` from zoom-frm.el - it screws up the frame when it is in fullscreen mode.
I had these piece of code in my config:
```
(spacemacs/transient-state-register-add-bindings 'zoom-frm
  '(("k" zoom-frm-in)
    ("j" zoom-frm-out)
    ("r" zoom-frm-unzoom)
    ("f" spacemacs/toggle-frame-fullscreen-non-native)
    ("m" spacemacs/toggle-maximize-frame)
    ("n" (ag/move-frame-one-display "North"))
    ("p" (ag/move-frame-one-display "South"))))
```

which I thought about removing, after this PR gets merged. I forgot that I had to use zoom-frm functions instead of spacemacs fns. Do you think we need to fix `(spacemacs/zoom-frm-in)` and ``(spacemacs/zoom-frm-out)`?

duianto:
Yeah there are issues with `spacemacs/zoom-frm-in` and `-out`.

It causes flickering because it first calls `zoom-frm-*`, then tries to restore the frame size. And if the window is maximized then it first toggles it unmaximized, tries to restore the frame size and maximizes again. This doesn't seem to be necessary because `zoom-frm-*` seems to handle it without un/maximizing.

There's also a windows 10 issue where almost every application has a 8px invisible border that extends past the application windows left, right and bottom edges. This isn't handled by the Spacemacs zoom function, so the frame shrinks when zooming in or out.

This is might be why it keeps increasing in size when zooming with a fullscreen frame.

In Ubuntu 18.04.02, Spacemacs zooming in decreases the frame size (height more than width), zooming out seems to keep the size.

The Spacemacs zoom functions docstrings say:
`"zoom in frame, but keep the same pixel size"`

So it seems like it tries to keep the frame size the same but just zoom it's contents, like font scaling `SPC z x`, but for the whole frame instead of just the current window.

`zoom-frm-in` / `-out` also in/decreases the frame size.

zoom-frm doesn't reset the mode-line height so the separators can end up looking like this:


source: https://github.com/syl20bnr/spacemacs/blob/develop/doc/DOCUMENTATION.org#separator-scale

#### Question
I'm not sure what the expected behavior should be. 

To zoom a frames contents, with or without resizing the frame?.

#### Possible future features

Maybe the font scaling TS could have a scope toggle for scaling the frame contents without resizing the frame:
- window (current behavior)
- frame
- or all frames (`zoom-frm.el` has functions for `zoom-all-frames-in` and `zoom-all-frames-out` (currently they resize the frame, but maybe they can be modified to only zoom the contents)).

And maybe there could be a separate frame manipulation TS, move/resize (in steps of 1, 10, 100px), it could have presets.

agzam:
oh wow. it seems it's a bit messier than I thought (is there a separate issue ticket?). I don't use Windows these days also I set `powerline-default-separator` to nil - don't like them much. I wonder if tweaking `frame-resize-pixelwise` makes any difference for the problems you mentioned?

duianto:
Enabling `frame-resize-pixelwise` doesn't seem to change the behavior.

There doesn't seem to be a current issue about it. I'll open one.

duianto:
Issue posted: Spacemacs zoom frame, shrinks, flickers #12343

PR #11211, opened 2018-08-25
python-mode: fix indent offset guessing for tab indentation

bennofs:
Since spacemacs changes the python tab width, we need to do the indent guessing after setting
the tab width.

### DIFF
--- a/layers/+lang/python/funcs.el
+++ b/layers/+lang/python/funcs.el
@@ -91,6 +91,11 @@ when this mode is enabled since the minibuffer is cleared all the time."
   (setq mode-name "Python"
         tab-width python-tab-width
         fill-column python-fill-column)
+
+  ;; since we changed the tab-width we need to manually cally python-indent-guess-inden-offset herr
+  (when python-spacemacs-indent-guess
+    (python-indent-guess-indent-offset))
+
   (when (version< emacs-version "24.5")
     ;; auto-indent on colon doesn't work well with if statement
     ;; should be fixed in 24.5 and above
--- a/layers/+lang/python/funcs.el
+++ b/layers/+lang/python/funcs.el
@@ -92,7 +92,7 @@ when this mode is enabled since the minibuffer is cleared all the time."
         tab-width python-tab-width
         fill-column python-fill-column)
 
-  ;; since we changed the tab-width we need to manually cally python-indent-guess-inden-offset herr
+  ;; since we changed the tab-width we need to manually call python-indent-guess-indent-offset here
   (when python-spacemacs-indent-guess
     (python-indent-guess-indent-offset))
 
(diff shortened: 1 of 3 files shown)


layers/+lang/python/funcs.el
```diff
@@ -91,6 +91,11 @@ when this mode is enabled since the minibuffer is cleared all the time."
   (setq mode-name "Python"
         tab-width python-tab-width
         fill-column python-fill-column)
+
+  ;; since we changed the tab-width we need to manually cally python-indent-guess-inden-offset herr
```
Compro-Prasad: Fix:
1. `cally` -> `call`
2. `herr` -> `here`

layers/+lang/python/packages.el
```diff
@@ -334,6 +334,9 @@
         "sR" 'spacemacs/python-shell-send-region-switch
         "sr" 'python-shell-send-region)
 
+      ;; we call python-indent-guess-indent-offset ourselves so set to nil to avoid duplicate work
```
Compro-Prasad: Replace this comment with something similar to:
```elisp
;; Set `python-indent-guess-indent-offset' to `nil' to prevent guessing `python-indent-offset'
```

layers/+lang/python/funcs.el
```diff
@@ -91,6 +91,11 @@ when this mode is enabled since the minibuffer is cleared all the time."
...
+  (when python-spacemacs-indent-guess
+    (python-indent-guess-indent-offset))
```
Compro-Prasad: Does this run whenever an user opens a new Python buffer? I guess not.

layers/+lang/python/packages.el
```diff
@@ -334,6 +334,9 @@
...
+      (setq-default python-indent-guess-indent-offset nil)
```
Compro-Prasad: Can't this be:
```
(setq-default python-indent-guess-indent-offset python-spacemacs-indent-guess)
```
?

Compro-Prasad: We could then probably cut out manually calling the function.

layers/+lang/python/funcs.el
bennofs: it does run whenever you open a buffer (it's part of `python-mode-hook`)

layers/+lang/python/packages.el
bennofs: No, this exactly the bug that is fixed by this pr: since `python-default` changes `tab-width` and runs after `python-mode-hook` (so after `python-mode` has already initialized itself), setting `python-indent-guess-indent-offset` to `t` will be buggy: since it runs with the wrong `tab-width` it calculates a wrong value (it needs to run after we set the tab-width).

bennofs:
@Compro-Prasad thanks for the review, fixed the parts that needed changing

Compro-Prasad:
@sdwolfz I think this is ready to merge.

sdwolfz:
@bennofs could you please give me some steps to reproduce the problem, I want to try it out locally but I'm not familiar with the issue.

sdwolfz:
Thank you :heart:!

Issue #11207, opened 2018-08-24
startup warning ad-disable-advice: rename-buffer is not advised


gnzlbg:
#### Description :octocat:
startup warning ad-disable-advice: rename-buffer is not advised

#### Reproduction guide :beetle:
- Start Emacs

*Observed behaviour:* :eyes: :broken_heart:

I am getting the following message on startup:

```
ad-disable-advice: ad-disable-advice: ‘rename-buffer’ is not advised
```

I'm on latest develop using emacs 26.

*Expected behaviour:* :heart: :smile:
no warning

#### System Info :computer:
- OS: darwin
- Emacs: 26.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. c122eb6a0)
- Graphic display: t
- Distribution: spacemacs
- Editing style: emacs
- Completion: helm
- Layers:
```elisp
(html helm auto-completion better-defaults emacs-lisp
      (git :variables git-gutter-use-fringe t git-enable-github-support t)
      markdown
      (shell :variables shell-default-height 30 shell-default-position 'bottom)
      spell-checking syntax-checking version-control javascript python yaml
      (c-c++)
      common-lisp latex osx shell-scripts rust)
```
- System configuration features: JPEG RSVG IMAGEMAGICK NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS MODULES THREADS LCMS2

#### Backtrace :paw_prints:
```
<<BACKTRACE IF RELEVANT>>
```

Miciah:
This is strange.  Disabling the advice when it is already disabled does not cause any error (at least it did not when I tried), so it looks like in your case, the advice does not exist at all.  However, the eyebrowse package advises `rename-buffer` when the package is loaded, and Spacemacs always installs the eyebrowse package and should be initializing it before it activates persp-mode (which is what tries to disable the advice).

@gnzlbg, can you check a few things?

1. What version of the eyebrowse package is installed? Use <kbd>M-m h d p</kbd> `eyebrowse` <kbd>RET</kbd> to check.
2. What advice does the `rename-buffer` function have? Use <kbd>M-m h d f</kbd> `rename-buffer` <kbd>RET</kbd> and look for the ":around advice:" lines in the help buffer to check.
[...]

PR #11213, opened 2018-08-25
Only disable eyebrowse advice if it exists

Miciah:
Fix issue #11207.

Only try to disable eyebrowse's advice on rename-buffer if we can find that advice.  If eyebrowse is older than version 0.7.6 (2017-11-22), the advice will not exist, and attempting to delete it will cause an error:

    ad-disable-advice: rename-buffer is not advised

* `layers/+spacemacs/spacemacs-layouts/funcs.el` (`spacemacs//activate-persp-mode`): Only disable the `eyebrowse-fixup-window-configs` advice for `rename-buffer` if the advice exists.


sdwolfz:
Thank you :heart:!

PR #11214, opened 2018-08-25
[mu4e] kill mu4e layout on app exit

agzam:
Kills mu4e layout when user quits mu4e app


sdwolfz:
@agzam I am not familiar with the functionality, could you please tell me why you think this should be enabled by default?

agzam:
when user activates mu4e's custom layout - `SPC l o m` and then closes mu4e app - `M-x mu4e-quit` - it leaves an empty layout. Totally unnecessary.

sdwolfz:
Thank you :heart:!

Issue #10896, opened 2018-06-15
Can not copy paste because pbcopy was disabled on Mac OS X layer


andresgutgon:
#### Description :octocat:
[This change](https://github.com/syl20bnr/spacemacs/commit/386a8fa79d0c8564daab0a539b66b52fe5d808ef) commented `pbcopy` on Mac OS X layer

#### Reproduction guide :beetle:
- Start Emacs on Terminal
- copy a line with `yy` and go to other app or terminal. The text is not in your clipboard

*Observed behaviour:* :eyes: :broken_heart:
-

*Expected behaviour:* :heart: :smile:
Copy & paste text from inside spacemac  with `yy` to have in Mac OS X clipboard

#### Backtrace :paw_prints:
-

I hope is enough info. What a did as temporal fix was [revert the commit](https://github.com/syl20bnr/spacemacs/commit/386a8fa79d0c8564daab0a539b66b52fe5d808ef)

cc @syl20bnr

syl20bnr:
You need to add it yourself in your dotfile. For now it introduces unbearable latency with Emacs 26.1 and daemon mode.

andresgutgon:
Can we let this issue opened? I think current behaviour is buggy. It's obscure that copy & paste doesn't work

bleggett:
This is another issue that is [fixed in Mitsuharu's mac-enhanced port](https://bitbucket.org/mituharu/emacs-mac/).

aaronjensen:
I use this:

```
(defun aj/copy-from-osx ()
  (shell-command-to-string "pbpaste"))

(defun aj/paste-to-osx (text &optional push)
  (let ((process-connection-type nil))
    (let ((proc (start-process "pbcopy" nil "pbcopy")))
      (process-send-string proc text)
      (process-send-eof proc))))

(defun aj/select-text (text)
  (if (display-graphic-p)
      (gui-select-text text)
    (aj/paste-to-osx text)))

(defun aj/selection-value ()
  (if (display-graphic-p)
      (gui-selection-value)
    (aj/copy-from-osx)))

(setq interprogram-cut-function 'aj/select-text
      interprogram-paste-function 'aj/selection-value)
```

There is very little latency, which could have to do with how optimized my shell boot time is.

> This is another issue that is fixed in Mitsuharu's mac-enhanced port without the need for pbcopy.el

Unfortunately, that also breaks the ability to have tty and gui emacs share a server 😢

ngoue:
[...]

PR #11215, opened 2018-08-26
Replace pbcopy by osx-clipboard

suonlight:
Issue: https://github.com/syl20bnr/spacemacs/issues/10896#issuecomment-415563095

Why:

* When using emacs >= 26, pbcopy has latency actions in terminal.

* Replace pbcopy by osx-clipboard

### DIFF
--- a/layers/+os/osx/packages.el
+++ b/layers/+os/osx/packages.el
@@ -16,9 +16,7 @@
         launchctl
         (osx-dictionary :toggle osx-use-dictionary-app)
         osx-trash
-        ;; disabled because it introduces input latency with some
-        ;; actions when using emacs -daemon and opening a GUI client
-        ;; pbcopy
+        osx-clipboard
         reveal-in-osx-finder
         term
         ))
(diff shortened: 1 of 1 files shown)


aaronjensen:
Looks okay to me, though It's going to make me disable it because it's not properly handling the case of one daemon serving terminal and gui clients. To do so, something like https://github.com/syl20bnr/spacemacs/issues/10896#issuecomment-415563095 should be done, which would actually allow me to remove that custom code from my dotfiles and would be great.

Oh, and I'd remove that check for the emacs major version unless `osx-clipboard` has compatibility issues with <26.

sdwolfz:
@suonlight will you be willing to incorporate @aaronjensen's suggested code as well?

aaronjensen:
@suonlight this looks good to me. The only thing I might suggest is making those lambdas proper functions, but it's probably not necessary.

theckman:
@duianto I saw you cherry-picked some open PRs in to develop the other day. I've also applied this patch on top of `develop` and can confirm that this results in Spacemacs working correctly on my system in both TUI an GUI mode. Without this patch TUI definitely is no-op, and I've some issues with GUI mode.

Any chance you could also `cherry-pick` this one? It's the final issue that has me maintaining my own `develop` branch so that I have a fully functioning editor.

duianto:

PR #11221, opened 2018-08-27
Fix pyenv using system Python

aeberspaecher:
This PR tries to fix the invalid Python executables being found on systems that have pyenv installed, but use system Python instead of a Python installed in a virtual environment.

Maybe @cpaulik (who wrote the code being changed) or @vv111y or @kccai (who commented on the isue hopefully fixed here in commit #0c878ae) can comment on this PR.


sdwolfz:
I am not sure about this since I don't use python. Preferably some people can test this out before I can merge it.

vv111y:
Just tried it out and it worked

yetone:
I am very sorry!

PR #11224, opened 2018-08-28
Fix fasd layer keybindings

agzam:
keybinging conflict introduced in
https://github.com/syl20bnr/spacemacs/commit/c2e377c902ba5cbed9ef6be47c3fa7f45a8bb846
made fasd layer unusable.

This PR proposes a fix.

### DIFF
--- a/layers/+tools/fasd/README.org
+++ b/layers/+tools/fasd/README.org
@@ -36,6 +36,6 @@ On OS X, it can be installed via [[https://github.com/Homebrew/legacy-homebrew][
 
 | Key Binding | Description                        |
 |-------------+------------------------------------|
-| ~SPC f a s~ | find a file or directory with fasd |
-| ~SPC f a d~ | find a directory with fasd         |
-| ~SPC f a f~ | find a file with fasd              |
+| ~SPC a f f~ | find a file or directory with fasd |
+| ~SPC a f d~ | find a directory with fasd         |
+| ~SPC a f s~ | find a file with fasd              |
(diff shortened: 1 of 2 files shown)


sdwolfz:
The conflict is my mistake, sorry about that. I wanted to suggest moving `spacemacs/find-file-and-replace-buffer` to `SPC f A` as a quick fix, but your change makes more sense since `fasd` is an external tool, it should live under the applications prefix (just hope this does not overwrite something else).

This PR should have a change-log entry as well since it's a big one for the people used to the old keybindings.

sdwolfz:
Actually, `SPC a f` seems to be taken by `elfeed`.

agzam:
oh well... it's a mess now. That's why I am very opposed for any new keybindings to be added. Every single new keybinding requires careful thought and consideration. What do we do now?

sdwolfz:
Let's go with my `SPC f A` workaround. I will get to it once I arrive home.

sdwolfz:
Pushed 816099aa2b5c4c9237da68e11fdc3abc47bc851b with the fix. Thanks for helping out.

sdwolfz:
Oh sorry about that, do you think you can come up with a different location for it? Both `SPC a f` and `SPC a F` are taken.

agzam:
Yeah, it seems we're running out of available options :) Spacemacs does so **much** now. I don't have any good suggestions to what to use, no.

sdwolfz:
There seems to be an `A` prefix for `other applications` according to `layers/+spacemacs/spacemacs-defaults/keybindings.el:18`. I guess it can be moved there. Do you find that to have good enough UX to fit in with your normal usage?

PR #11239, opened 2018-08-30
[spacemacs-docker] clean .#recentf

JAremko:
Looks like the lock-file of recentf needs to be deleted after dependency installation.


sdwolfz:
Would `(setq create-lockfiles nil)` help in this case?

JAremko:
@sdwolfz  I think it's more reliable to simply remove it

Issue #9328, opened 2017-07-31
Support for org-trello


ageekymonk:
Can Org-trello support be added in org layer? It could be a feature toggle as it will not be used by everyone who uses org mode.

d12frosted:
Once we had it and [then it got removed](https://github.com/syl20bnr/spacemacs/commit/5544451946c48e66bb35b3b3f0f2bff2611b2a86).

Miciah:
The referenced commit that removed org-trello (https://github.com/syl20bnr/spacemacs/commit/5544451946c48e66bb35b3b3f0f2bff2611b2a86) was made before the `:toggle` option for packages was added to the layers system.  Any reason not to restore org-trello with a `:toggle` qualifier?

sdwolfz:
I have no problem with that, we already have stuff like `org-enable-hugo-support` or `org-enable-org-journal-support`. You can make a PR that adds it like those do, with a `org-enable-trello-support` variable.

magthe:
I found `org-trello` today, which led me here. Would something like this work?

https://github.com/magthe/spacemacs/commit/8b77c2b34731e064953f644912838407905db709

sdwolfz:
@magthe seems right, you might want to try deffering that package, or adding in some keybindings. Open up a PR and maybe someone else in the community will be able to suggest further improvements.

CodeStar8:
Have you tried installing org-trello for emacs only? Since Spacemacs runs on top of emacs

CodeStar8:
I realized that Spacemacs is deleting the emacs package. This other github ticket talks about it...

https://github.com/syl20bnr/spacemacs/issues/9269

undergroundquizscene:
Should this issue be closed, as `org-trello` is already included in `develop` via magthe's PR above?

PR #11244, opened 2018-08-31
Re-add org-trello

magthe:
After finding issue #9328 I thought I'd try to add _org-trello_ in the "right" way.

- I'm using the really rather cool `:toggle` option of declaring dependencies.
- The loading should be deferred.
- There's a rather minimal set of spacemacs-style key bindings at the moment, but it should be enough to get started (I think)

Comments and suggestions are most welcome!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,6 +50,7 @@ python-*-docs-html
 nov-places
 workspace
 eclipse.jdt.ls
+.trello
 
 # Private directory
 private/
--- a/layers/+emacs/org/README.org
+++ b/layers/+emacs/org/README.org
@@ -15,6 +15,7 @@
   - [[#revealjs-support][Reveal.js support]]
   - [[#org-journal-support][Org-journal support]]
   - [[#hugo-support][Hugo support]]
+  - [[#trello-support][Trello support]]
   - [[#different-bullets][Different bullets]]
   - [[#project-support][Project support]]
   - [[#org-brain-support][Org-brain support]]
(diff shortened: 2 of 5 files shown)


magthe:
I believe that addresses all review comments so far. Anything else, or does it look good enough now?

sdwolfz:
Thank you :heart:!

I squashed your commits into one: d29c7832868e072a73166440437ca88c4a268724 and made a change to load `org-trello` after `org` otherwise the keybindings were not created.

syl20bnr:
We cannot use `SPC m o` prefix for this, it is reserved for the user. See conventions: https://github.com/syl20bnr/spacemacs/blob/develop/doc/CONVENTIONS.org#user-prefix

We need a prefix for Org integrations, `SPC m a` and `SPC m A` for `application` are already taken. `SPC m s` for `service` is already taken. `SPC i` for `integration` already taken ... :-)

Maybe `SPC m m` for `more` ? and then we can use `t` for Trello. Could be a conventions to add a second layer of major modes specific key bindings.

LucaMarconato:
Hi, I tried adding `org-trello` in the layers (I am using the `develop` branch) but at the startup I get this in the `*Message*` buffer. Any idea why?

```
(Spacemacs) Warning: Unknown layer org-trello declared in dotfile.
(Spacemacs) --> deleting request-deferred... [1/1]
Package ‘request-deferred-20160419.2305’ deleted.
```

PR #11249, opened 2018-08-31
Bind treemacs-projectile

ellakk:
Adds binding for treemacs-projectile. I also removed some unneeded pound signs i found while editing the treemacs package.el file. I had to remove the `:after` block because it inhibited the `:init` block from running, it should not matter since there is only one interactive command in the `treemacs-projectile.el` file and it's autoloaded.

Initially i created a custom function that would either add the current projectile project to treemacs or toggle treemacs if it's already added, kinda like the neotree version of the "pt" command does but changed my mind and went with the `treemacs-projectile` command instead. The difference is that in `treemacs-projectile` you have to select what project you want to add from all projectile projects and it does not toggle treemacs, it's only for adding new projects to treemacs. If you want me to submit a pull-request with the custom version and close this one then tell me so.

### DIFF
--- a/layers/+filetree/treemacs/funcs.el
+++ b/layers/+filetree/treemacs/funcs.el
@@ -1,0 +1,22 @@
+;;; packages.el --- Neotree Layer functions File
+;;
+;; Copyright (c) 2012-2018 Sylvain Benner & Contributors
+;;
+;; Author: Sylvain Benner <sylvain.benner@gmail.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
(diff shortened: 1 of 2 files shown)


sdwolfz:
Thank you :heart:!
Cherry-picked your first commit into develop.

For the second one, I would very much prefer it having the same functionality as neotree, if it's not too complicated to achieve it.

ellakk:
@sdwolfz Updated pr with said function, can you cherry-pick it and close this pr?

sdwolfz:
Thank you :heart:!
Cherry-picked the second commit into develop branch as well, you can safely delete your branch.

PR #11251, opened 2018-09-01
[CI] Add linters

sdwolfz:
This is the beginning for adding automatic checks. My idea is to have a place in the CI pipeline where we execute `make ci` and everything else flows as a result of that. Currently `editorconfig` is the only check executed, and it does not have full coverage, but this will be fixed in the future.

- `nyan-mode` layer had `CR` removed from line endings.
- some makefiles were fixed with tabs.

@JAremko Let me know what you think.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -101,6 +101,22 @@ jobs:
     working_directory: ~/.emacs.d
     <<: *spacemacs_steps
 
+  "Linters (required)":
+    docker:
+      - image: jare/spacemacs-circleci:latest
+    working_directory: ~/.emacs.d
+    steps:
+      - checkout
+      - run:
+          name: Install npm
+          command: apt-get update && apt-get install -y --no-install-recommends npm
+      - run:
+          name: Install eclint
+          command: npm install -g eclint
+      - run:
+          name: Run Linters
+          command: make ci
+
   # Emacs snapshot
   "core Emacs snapshot (optional)":
     docker:
@@ -149,6 +165,9 @@ workflows:
           requires:
             - "Validate PR"
       - "spacemacs dist. Emacs25 (required)":
+          requires:
+            - "Validate PR"
+      - "Linters (required)":
           requires:
             - "Validate PR"
       - "core Emacs snapshot (optional)":
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,61 @@
+root = true
+
+# -------------------------------------------------------------------------------
+# Global config, to be enabled when everything is fixed.
+# -------------------------------------------------------------------------------
+# [*]
+# charset                  = utf-8
+# end_of_line              = lf
+# indent_size              = 2
+# indent_style             = space
+# insert_final_newline     = true
+# tab_width                = 2
+# trim_trailing_whitespace = true
+
+# -------------------------------------------------------------------------------
+# Emacs Lisp
+# -------------------------------------------------------------------------------
+[{core,layers,tests}/**.el]
+charset                  = utf-8
+end_of_line              = lf
+# indent_size              = 2
+# indent_style             = space
+insert_final_newline     = true
+# tab_width                = 2
+trim_trailing_whitespace = true
+
+# -------------------------------------------------------------------------------
+# Make
+# -------------------------------------------------------------------------------
+[**/{Makefile,**.mk}]
+charset                  = utf-8
+end_of_line              = lf
+indent_size              = 2
+indent_style             = tab
+insert_final_newline     = true
+tab_width                = 2
+trim_trailing_whitespace = true
+
+# -------------------------------------------------------------------------------
+# Markdown
+# -------------------------------------------------------------------------------
+[**.md]
+charset                  = utf-8
+end_of_line              = lf
+# indent_size              = 2
+indent_style             = space
+insert_final_newline     = true
+tab_width                = 2
+trim_trailing_whitespace = true
+
+# -------------------------------------------------------------------------------
+# Org
+# -------------------------------------------------------------------------------
+[{core,doc,layers,news,tests}/**.org]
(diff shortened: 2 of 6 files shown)


JAremko:
@sdwolfz  You can stick it into https://github.com/syl20bnr/spacemacs/blob/develop/.travisci/test.sh. To fail the test you just need to exit with non-zero code. Fails will be reported only on branch updates (not PRs) so we can test new (well.. tests) internally before subjugating PR authors to them.

JAremko:
Have anyone tried [elisp-lint](https://github.com/gonewest818/elisp-lint) ?  Can we use it  in CI?
I think, in a long run, we'll need something like [package-lint](https://github.com/purcell/package-lint) for `packages.el` files.

sdwolfz:
Haven't tried it yet, but it's one of the things I will check next.

JAremko:
@sdwolfz  I'm a bit not sure about putting make Makefile in the root folder.  Mb it will be better to put it into https://github.com/syl20bnr/spacemacs/tree/develop/tests under something like `/lint/`  ?

JAremko:
This is how we run "make based" test:

Circle CI has [config.yml](https://github.com/syl20bnr/spacemacs/blob/develop/.circleci/config.yml) file with:
``` yaml
  steps_core_steps: &steps_core_steps
    steps:
      - checkout
      - run:
          name: core tests installation
          command:
            ./.circleci/test core installation
      - run:
          name: core unit tests
          command:
            ./.circleci/test core unit_tests
      - run:
          name: core functional tests
          command:
            ./.circleci/test core func_tests
```

this think says call  [./.circleci/test](https://github.com/syl20bnr/spacemacs/blob/develop/.circleci/test) with arguments where the first argument is the dir folder sub-directory (in the case above it is [core](https://github.com/syl20bnr/spacemacs/tree/develop/tests/core) the second one is the target  from [spacemacs.mk](https://github.com/syl20bnr/spacemacs/blob/develop/spacemacs.mk) (it is shared file)

sdwolfz:
I have big plans for that root Makefile. My intent is to make it the main `executable` for development and testing related commands so that I can ask contributors to execute a command in their spacemacs dir before submitting.

TravisCI seems like the wrong place to put these, I would rather have them with the rest of the CircleCI jobs. Do you think you can have `eclint` pre-installed in any of the docker images?

JAremko:
@sdwolfz  I added you as a collaborator to https://github.com/JAremko/spacemacs-circleci you can add anything you need :shipit:

sdwolfz:
Thanks, I'll see what I can do.

JAremko:
@sdwolfz  After pushing something  you can see the build status here https://hub.docker.com/r/jare/spacemacs-circleci/builds/

sdwolfz:
Looks like it's failing properly :tada:: https://circleci.com/gh/syl20bnr/spacemacs/14945?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Please let me know if you have any concerned with my current approach.

JAremko:
@sdwolfz  :+1:   How about splitting "Run Linters" into sub commands. Like "Run editorconfig linter" "Run ... linter"  It will be easier to tell which one is failing.

sdwolfz:
Good point. I will do that when I will add the next linter.

syl20bnr:
Can we but the makefile in a sub-directory somewhere, I would like to not add a makefile at the root of spacemacs which can be missleading about how the project works, i.e. Spacemacs does not need anything external, any outside process to be used.

We should put everything `ci` related in the `test` sub-folder except for conventional dot directories. Even `spacemacs.mk` should go there.

JAremko:
@sdwolfz  I'll nuke current PR format checker (in CircleCI) since the linters will detect misplaced white-spaces and all its other features will be covered by auto-fixer on branch updates.

 - - - - -

Mb we should merge this?

sdwolfz:
Yes I'll focus my effort this weekend to get this done and merged. But if you have any improvements of your own you would like to make yourself you can hijack this PR and push to this branch.

JAremko:
@sdwolfz  Consider adding  [checkdoc.el](https://github.com/emacs-mirror/emacs/blob/master/lisp/emacs-lisp/checkdoc.el) it will report missing or malformed dock-strings. [Sesman uses it](https://github.com/vspinu/sesman/blob/master/targets/checkdoc.el#L2)

JAremko:
Also worth mentioning. https://github.com/emacs-elsa/Elsa Mb it will mature enough for us to use it before this PR get merged :shipit:

sdwolfz:
With Elsa I failed to get it to run on a simple elisp file so I'm going to wait for a while.
With line length, I am not sure yet, but `eclint` seems to support that (unoficially) so I might integrate it that way.
Checkdoc seems interesting, I will keep it in mind for later.

Anyway I did not manage to keep my promise about doing it over that weekend... hopefully I can get something done this month.

@syl20bnr related to the `Makefile`, I intend to use it as a quick run tasks tool, my purpose in the end is to have commands like `make ci` and `make test` that contributors can run from the repository root before submitting PRs. This is to make the process easier, and a makefile seems like the simplest thing to add and I would like to keep it in the root. It will be properly documented so there should be no confusion about it's purpose. Do you intent to do anything else with it?

Here is an example of how I use it in a personal project: https://gitlab.com/sdwolfz/dotfiles/blob/master/Makefile

syl20bnr:
No problem to do a Makefile. Let’s put it in a sub-directory as it is already the case for the tests. Also we can have several Makefiles if that makes sense, i.e. one in `ci`, others in `tests`, another in `dev` etc... just not at the root of the project because Spacemacs is designed to be used without any Makefile which is an important distinction with other distributions.

sdwolfz:
I'll close this for now as I will not be able to work on this in the near future.

PR #11277, opened 2018-09-06
dap-layer implementation

yyoncho:
- I had some issues related to finding the corresponding shortcuts for each of
the commands and I am looking for suggestions for improvements.

- updated the lsp-java readme since the server installation is now automatic. If
you already have jdt server installation, please run lsp-java-update-server
which will download the related dap-mode related stuff.

- Make sure that you have latest melpa versions of lsp-java, lsp-ui and dap-mode
since several issues related to the java integration were fixed lately.

- dap-mode is pretty young package and it is not extensively tested

- dap-hydra which will be used as a transient state for the breakpoints and
debugging, in general, is not synched with the shortcuts in the layer. I will sync
them once there is agreement on the shortcuts configuration. Also, I have not
updated lsp-java README.org with the corresponding shortcuts for the same
reason.

- The debug integration with the python lsp layer will be implemented in
separate PR.

- There will be few more functions that are going to be added to the dap-mode at
some point and they would require shortcuts too:
* suspend/kill thread I guess - "m d t s" "m d t k"?
* watches - "m d l w"
* exception breakpoints - I have to figure out how it will look like on emacs
  side.
* goto point "m d H"?

### DIFF
--- a/layers/+lang/java/packages.el
+++ b/layers/+lang/java/packages.el
@@ -29,7 +29,7 @@
         maven-test-mode
         (meghanada :toggle (not (version< emacs-version "25.1")))
         mvn
-        (lsp-java :requires lsp-mode lsp-ui company-lsp)
+        (lsp-java :requires lsp-mode lsp-ui company-lsp dap-mode)
         org
         ))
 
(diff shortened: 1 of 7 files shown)


layers/+tools/dap/funcs.el
```diff
@@ -0,0 +1,66 @@
...
+    "dIr" #'dap-ui-inspect-region
+    "dIt" #'dap-ui-inspect-thing-at-point
+    ;; breakpoints
+    "dbb" #'dap-breakpoint-toggle
+    "dbc" #'dap-breakpoint-condition
+    "dbl" #'dap-breakpoint-log-message
+    "dbh" #'dap-breakpoint-hit-condition
+    "dba" #'dap-breakpoint-add
+    "dbd" #'dap-breakpoint-delete
+    "dB"  #'dap-breakpoint-delete-all
```
sdwolfz: I would suggest `dbD` for this one.

layers/+tools/dap/funcs.el
```diff
@@ -0,0 +1,66 @@
...
+    ;; repl
+    "d'"  #'dap-ui-repl
+    ;; windows
+    "dlo" #'dap-go-to-output-buffer
```
sdwolfz: Why have you chosen `dl` for this one, since you mention windows I would assume `w` would be more fitting?

layers/+tools/dap/funcs.el
```diff
@@ -0,0 +1,66 @@
...
+  (spacemacs/declare-prefix-for-mode mode "mdI" "inspect")
+  (spacemacs/declare-prefix-for-mode mode "mde" "eval")
+
+  (spacemacs/set-leader-keys-for-major-mode mode
+    ;; debuging/running
+    "ddd" #'dap-debug
+    "ddl" #'dap-debug-last
+    "ddr" #'dap-debug-recent
+    ;; stepping
+    "dc" #'dap-continue
```
sdwolfz: I am not yet convinced all these should be without another name-space. It does make typing them easier, but then again we have the hydra...

layers/+lang/java/packages.el
```diff
@@ -462,7 +462,12 @@
         "rem" 'lsp-java-extract-method
         "cc"  'lsp-java-build-project
         "an"  'lsp-java-actionable-notifications
-        "="   'lsp-format-buffer)
+        "="   'lsp-format-buffer
+
+        ;; dap-mode
+        "xj"   'dap-java-debug
```
sdwolfz: This `x` feels like a weird prefix. Why have you chosen this one?

layers/+lang/java/packages.el
```diff
@@ -462,7 +462,12 @@
...
+        "tt"   'dap-java-debug-test-method
```
sdwolfz: This is used in other modes as "Run only the current test at cursor position" and I would like to keep it that way here as well. Can you think of another key binding?

layers/+tools/dap/funcs.el
yyoncho: Make sense

layers/+tools/dap/funcs.el
yyoncho: There is no valid reason - I wanted to avoid as much as possible the standard navigation keys so the hydra could work and also I started with `mdls` which stands for `list session` and then I started using `l` as a prefix for going to the DAP windows.

layers/+lang/java/packages.el
yyoncho: I used `x` since it is available and wont be missed in the hydra - e. g. you will be able to navigate readonly with hydra on. Also it comes from `eXecute`.

layers/+lang/java/packages.el
yyoncho: What do you think about adding no-debug prefix for the command so `SPCu` `SPCmtt` would start without debugging?

yyoncho:
@sdwolfz  I have applied the comments without pending questions. 

I still have some consideration about putting everything behind "md" since DAP mode will work with LSP layer which has predefined shortcuts so the chances for collisions are not very high. I believe that LSP/DAP(probably BSP in the future) are meant to work in sync.

Also, I am thinking about the way language layers are organized. It will be much easier to have `enabled-languages` in lsp mode itself instead of spreading those cond statements in every lsp enabled layer. We are doing a refactoring in `lsp-mode` to provide single entry point and I wonder how this improvement will work in `spacemacs`.

layers/+tools/dap/README.org
```diff
@@ -0,0 +1,102 @@
...
+  - [[#features][Features:]]
+- [[#configuration][Configuration]]
+  - [[#derived-layers][Derived layers]]
+    - [[#declared-prefixes][Declared prefixes]]
+    - [[#default-keybindings][Default keybindings]]
+- [[#references][References]]
+
+* Description
+  Experimental integrated visual debugger using [[https://code.visualstudio.com/docs/extensionAPI/api-debugging][Debug Adapter Protocol]].
+
```
sdwolfz: Please add a `Layer Instalation` section to this readme.

layers/+tools/dap/funcs.el
sdwolfz: I don't think the hydra should contain toggles for windows anyway and `debug windows` is not a good description for an `l` key. What I suggest is either use `w` for this or change the description to say `layout`, although I would advise more on going with `w`.

layers/+tools/dap/README.org
```diff
@@ -0,0 +1,102 @@
...
+    |-------------+---------------------------------|
+    | ~m d b b~   | toggle a breakpoint             |
+    | ~m d b c~   | change breakpoint condition     |
+    | ~m d b l~   | change breakpoint log condition |
+    | ~m d b h~   | change breakpoint hit count     |
+    | ~m d b a~   | add a breakpoint                |
+    | ~m d b d~   | delete a breakpoint             |
+    | ~m d b D~   | clear all breakpoints           |
+    |-------------+---------------------------------|
+    | ~m d '~     | Run debug REPL                  |
```
sdwolfz: This line does not display properly, you need to change it to:

```suggestion
    | ~m d '_~     | Run debug REPL                  |
```

layers/+lang/java/packages.el
sdwolfz: I think this should be under the `debug` prefix as well. Unless there is a use-case for it being outside that I'm missing.

layers/+lang/java/packages.el
sdwolfz: I would rather this keybinding live under the `debug` prefix as well, maybe `, d t t` in case you want to put more keybindings next to it.

layers/+tools/dap/README.org
```diff
@@ -0,0 +1,102 @@
...
+   specific debug adapter and the language specific bindings(e. g. debug
+   application, debug test, etc). The package declares function
+   ~spacemacs/dap-bind-keys-for-mode~ which should be used by the derived layers
+   to configure the shortcuts.
+*** Declared prefixes
+    The following prefixes have been declared:
+
+    | prefix  | name              |
+    |---------+-------------------|
+    | ~m d~   | debug             |
```
sdwolfz: Please add `SPC` for each of these keybindings in the first column of the table. We have this everywhere else and it makes searching easier.

sdwolfz:
Also please rebase with the latest develop as there are some conflicts.

yyoncho:
@sdwolfz I have addressed your comments.

sdwolfz:
Thank you for your patience and awesome contributions :heart:! This functionality is simply amazing!

Issue #10846, opened 2018-06-10
PATCH: add Epub export to Org layer


tidux:
This patch to layers/+emacs/org/packages.el is tested and working.

```
30a31
>     ox-epub
525a527,529
> 
> (defun org/init-ox-epub ()
>   (spacemacs|use-package-add-hook org :post-config (require 'ox-epub)))
```

I can resubmit this as a PR if requested.

robbyoconnor:
Submit as a pull request...against `develop`

syl20bnr:
Anyone wants to submit a PR for this patch ?

TinySong:
should we add some shortcut  for org-epub-export-to-epub ?

duianto:
The PR:
Fixes #10846, added epub support for Org layer. #11926
Has been cherry-picked into the develop branch.

duianto:
Hopefully it worked this time 😄 
Thank you for contributing to Spacemacs!
The PR has been cherry-picked into the `develop` branch, you can safely delete your branch.

PR #11280, opened 2018-09-06
Add ox-epub package for exporting org -> epub

TinySong:
Refs: #10846

### DIFF
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -38,6 +38,7 @@
         persp-mode
         (ox-hugo :toggle org-enable-hugo-support)
         (org-trello :toggle org-enable-trello-support)
+        ox-epub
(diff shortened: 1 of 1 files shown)


sdwolfz:
It feels to me like we should have this under a toggle as well, what do you think?

vishvanath45:
@sdwolfz  Hi, I want to work on this issue, what kind of toggle you want to have for this feature as it will just show up in export dispatcher.

sdwolfz:
`(org-epub :toggle org-enable-epub-support)` in the package list and a variable `org-enable-epub-support` defaulting to nil.

duianto:
Thanks for posting this PR, but it's been superseded by:
Fixes #10846, added epub support for Org layer. #11926
which also added documentation and a toggle variable.

PR #11283, opened 2018-09-06
Reenable lsp mode lighter

yyoncho:
- The lsp modeline is used to provide information about the server status so it
should not be be hidden.

### DIFF
--- a/layers/+tools/lsp/packages.el
+++ b/layers/+tools/lsp/packages.el
@@ -28,7 +28,6 @@
     :defer t
     :config
     (progn
-      (spacemacs|hide-lighter lsp-mode)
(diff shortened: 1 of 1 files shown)


robbyoconnor:
Why was it disabled?

yyoncho:
@robbyoconnor I believe that it was disabled under the impression that the lighter is not used but it is only taking space since not all of the languages are using it to display that info.

robbyoconnor:
Ah okay, makes sense.

Compro-Prasad:
@sdwolfz This is good to go.

sdwolfz:
Thank you :heart:!

PR #11291, opened 2018-09-08
 Add `multi-term` to `shell-default-shell` doc

mjkramer:
Previously, `multi-term` was not listed as an available option for `shell-default-shell`.


robbyoconnor:
I recommend closing this and making a topic branch.

mjkramer:
Will do!

robbyoconnor:
It's not a HUGE issue but it's good practice nonetheless :)

PR #11294, opened 2018-09-08
org layer: add helm-org-rifle, bind to SPC m r

siddharthist:
This adds [helm-org-rifle](https://github.com/alphapapa/helm-org-rifle), a nice tool for navigating org files.


layers/+emacs/org/packages.el
```diff
@@ -17,6 +17,7 @@
         evil-org
         evil-surround
         gnuplot
+        helm-org-rifle
```
d12frosted: It should be available only for `helm` users I guess.

E.g. something like

```
(helm-org-rifle :toggle (configuration-layer/layer-used-p 'helm))
```

d12frosted:

P. S. I've made a respective change 😸

siddharthist:
Awesome, thanks so much for your help!

PR #11296, opened 2018-09-09
Add `multi-term' to `shell-default-shell' doc

mjkramer:
`multi-term` has long been an option for `shell-default-shell` but was missing from the doc string.


Compro-Prasad:
@sdwolfz This too looks good!

sdwolfz:
Thank you :heart:!

PR #11298, opened 2018-09-09
Add ivy-pass and use melpa for counsel-notmuch

ellakk:
Fetch ivy-notmuch from Melpa and add ivy-pass to the pass-layer, it provides the same functionality as helm-pass does.


sdwolfz:
Thank you :heart:!

PR #11299, opened 2018-09-09
Remove which-key replacement: `SPC 0` and `M-0`

duianto:
Problem:
The previous which-key replacement for both `SPC 0` and `M-0`,
renamed: `winum-select-window-0-or-10` to: `select window 1 or 10`,
but the replacement gets overwritten by the filetree layers that show:
neotree: `neotree-show`
treemacs: `treemacs-select-window`

Solution:
Remove the previous which-key replacement, it's not needed anymore.

### DIFF
--- a/layers/+distributions/spacemacs-bootstrap/packages.el
+++ b/layers/+distributions/spacemacs-bootstrap/packages.el
@@ -371,12 +371,6 @@
           ("\\1C-M-0..9" . "digit-argument"))
         which-key-replacement-alist)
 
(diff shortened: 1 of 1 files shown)


Miciah:
Seems like we should keep the which-key replacement for people who use the spacemacs-base distribution or who explicitly disable neotree using `dotspacemacs-excluded-packages`.

duianto:
I didn't think about that. Thanks for enlightening me. I'll close this since it still has a use.

Issue #10369, opened 2018-02-20
What's #'?


re-mx:
Hello, I am an Emacs rookie.

Some spacemacs function names have `#'` in front of them。 but more spacemacs functions don't have it.
eg:
with `#'`:
```
(add-hook 'kill-buffer-hook #'spacemacs//add-buffer-to-killed-list)
```

without `#'`:
```
  (add-hook 'react-mode-hook 'spacemacs/js-doc-require)
```

I search for a moment, it seems that this symbol is used to indicate that followed by a function.
So why not all spacemacs functions take this symbol?

What is any special for those function with `#'`?

Sorry for my poor English.

Thank you very much!

CeleritasCelery:
http://endlessparentheses.com/get-in-the-habit-of-using-sharp-quote.html

re-mx:
I have read articles.

But I don't understand is why not all use?

PR #11300, opened 2018-09-09
[Filetree layers] bind C-x w 0 to open the filtree

duianto:
problem:
`C-x w 0` is bound to `winum-select-window-0-or-10`, this differs from the
`SPC 0` and `M-0` bindings that open the current filetree: neotree or treemacs

solution:
Bind `C-x w 0` in both filetree layers to match the other bindings.

### DIFF
--- a/layers/+filetree/neotree/packages.el
+++ b/layers/+filetree/neotree/packages.el
@@ -118,5 +118,7 @@ Navigation^^^^             Actions^^         Visual actions/config^^^
     :post-config
     ;; window 0 is reserved for file trees
     (spacemacs/set-leader-keys "0" #'neotree-show)
-    (define-key winum-keymap (kbd "M-0") #'neotree-show)
+    (define-key winum-keymap [remap winum-select-window-0-or-10] #'neotree-show)
+    (push '((nil . "winum-select-window-0-or-10") . (nil . "neotree-show"))
+          which-key-replacement-alist)
     (add-to-list 'winum-assign-functions #'spacemacs//winum-neotree-assign-func)))
--- a/layers/+filetree/treemacs/packages.el
+++ b/layers/+filetree/treemacs/packages.el
@@ -72,4 +72,6 @@
     (progn
       ;; window 0 is reserved for file trees
       (spacemacs/set-leader-keys "0" 'treemacs-select-window)
-      (define-key winum-keymap (kbd "M-0") 'treemacs-select-window))))
+      (define-key winum-keymap [remap winum-select-window-0-or-10] 'treemacs-select-window)
+      (push '((nil . "winum-select-window-0-or-10") . (nil . "treemacs-select-window"))
+            which-key-replacement-alist))))


Miciah:
How about using command remapping?  For example, we could use the following for NeoTree:
```emacs-lisp
    (define-key winum-keymap [remap winum-select-window-0-or-10] #'neotree-show)
```
And similar for treemacs.

duianto:
That remapping works, both `M-0` and `C-x w 0` opens the installed filetree.

But the which-key entries for those two bindings, now show the which-key replacement text:
`select window 0 or 10`, instead of the filetree commands: `neotree-show` or `treemacs-select-window`.

When the keys are bound individually then the filetree commands are shown for both bindings.

One solution that works is to add a which-key replacement after the remapping. It renames the description for the keys to the filetree command.
```
(which-key-add-key-based-replacements
  "C-x w 0" "neotree-show"
  "M-0"     "neotree-show")
```
It also seems possible to just replace the which-key descriptions.
Source: the second example here: 
https://github.com/justbur/emacs-which-key#key-and-description-replacement
```
(push '((nil . "left") . (nil . "lft")) which-key-replacement-alist)
```
it seems like this should work:
```
(push '((nil . "select window 0 or 10") . (nil . "neotree-show"))
      which-key-replacement-alist)
```
but It doesn't. I tried adding it after the remapping, or in the `neotree/init-neotree` functions `:init` or `:config` sections, but the the which-key description for `M-0` and `C-x w 0` still show:
`select window 0 or 10`


#### Small side note:
The pound symbols in the neotree layer, doesn't seem to be needed, according to this commit:
treemacs: remove unneeded pound symbols https://github.com/syl20bnr/spacemacs/commit/3784b752b67941649e6d7dd82c6fbca971d78223

In that case, can they also be removed from the neotree layer with this PR, or is it better to open a separate PR?

duianto:
> It seems that rather than replacing the replacement ("select window 0 or 10"), we need to override the original ("winum-select-window-0-or-10"):

Doh! of course 😄 

I'll leave the `#'` on the neotree layer, it does seem like a good idea to sharp quote symbols that are function names.

There was a question about it in february: What's #'? #10369
and the http://endlessparentheses.com/get-in-the-habit-of-using-sharp-quote.html article was suggested there as well.

The emacs-lisp-style-guide also suggest the same (the 4th point):
https://github.com/bbatsov/emacs-lisp-style-guide#functions

And I noticed that the treemacs author uses `#'`: https://github.com/syl20bnr/spacemacs/commit/403b246212c3b86f1927f0f509cc7556deb6312c#diff-7ed0a08795fa2a3ce23b0025c377d39fR47

Maybe the commit:
treemacs: remove unneeded pound symbols https://github.com/syl20bnr/spacemacs/commit/3784b752b67941649e6d7dd82c6fbca971d78223
should be reverted?

@ellakk it's probably best to not remove anymore pound signs that you come across.

Miciah:
Yeah, unless we are referencing a function before it (or its autoload) is defined, it makes sense to use `#'`.

duianto:
Two of the Treemacs lines extend past 80 chars:
```
      (define-key winum-keymap [remap winum-select-window-0-or-10] 'treemacs-select-window)
      (push '((nil . "winum-select-window-0-or-10") . (nil . "treemacs-select-window"))
            which-key-replacement-alist))))
```

and I noticed another line further up with the same problem
```
      (add-to-list 'spacemacs-window-split-ignore-prefixes treemacs--buffer-name-prefix))))
```

I'll open a separate PR that adds line breaks. When this PR is merged.

duianto:
This PR is now on the develop branch.
Commit: https://github.com/syl20bnr/spacemacs/commit/8a4bf9c655744081eebb030f587777c061f224e5

Thanks @Miciah for coming up with a better solution.

PR #11303, opened 2018-09-09
[treemacs] give names to which-key +prefix entries

duianto:
problem:
Which-key subgroups get the default name `+prefix`. The subgroup has to be
opened to find out which keys it contains.

solution:
Add more descriptive names:
```
c      treemacs-create
o      treemacs-visit-node
oa     treemacs-visit-node-ace
t      treemacs-toggles
y      treemacs-copy
C-p    treemacs-projects
C-p c  treemacs-projects-collapse
```


Compro-Prasad:
@sdwolfz This should work.

sdwolfz:
Thank you :heart:!

PR #11307, opened 2018-09-10
layers/+emacs/org: more clock keybindings

siddharthist:
Just some more keybindings!

Also, the current documentation mentions
```
 evil-org-recompute-clocks 
```
which actually isn't bound to that key.

### DIFF
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -177,10 +177,20 @@ Will work on both org-mode and any mode that accepts plain html."
       (spacemacs/set-leader-keys-for-major-mode 'org-mode
         "'" 'org-edit-special
         "c" 'org-capture
+
+        ;; Clock
+        ;; These keybindings should match those under the "aoC" prefix (below)
         "Cc" 'org-clock-cancel
+        "Cd" 'org-clock-display
+        "Ce" 'org-evaluate-time-range
+        "Cg" 'org-clock-goto
         "Ci" 'org-clock-in
+        "CI" 'org-clock-in-last
+        "Cj" 'org-clock-jump-to-current-clock
         "Co" 'org-clock-out
+        "CR" 'org-clock-report
         "Cr" 'org-resolve-clocks
+
         "dd" 'org-deadline
         "ds" 'org-schedule
         "dt" 'org-time-stamp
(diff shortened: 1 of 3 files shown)


layers/+emacs/org/packages.el
```diff
@@ -175,10 +175,18 @@ Will work on both org-mode and any mode that accepts plain html."
...
         "'" 'org-edit-special
         "c" 'org-capture
+
+        ;; Clock
+        ;; These keybindings should match those under the "aok" prefix (below)
         "Cc" 'org-clock-cancel
-        "Ci" 'org-clock-in
+        "Cd" 'org-clock-display
+        "Cg" 'org-clock-goto
+        "Ci" 'org-clock-in-last
```
sdwolfz: I see you replaced `org-clock-in` with `org-clock-in-last`, is this intentional?

siddharthist: I'm glad you pointed this out. I'm of two minds here: it more closely matches the `aok`-prefix keybindings (which I think is positive), but removes a common action from the keybindings. What do you think about the following change, under both prefixes:
- `CI`: `org-clock-in-last`
- `ci`: `org-clock-in`

sdwolfz: Yes it makes sense to have them match and I like the solution.

siddharthist: In fact @sdwolfz what would you think about moving `aok` to `aoC` for more consistency as well?

siddharthist: I've made both changes: `aok -> aoC` and `ci`/`cI`. I've also updated the `CHANGELOG` to be accurate.

smile13241324:
@siddharthist I have just seen that the formatting test fails, would you mind rebasing develop on your branch and push again? Actually this test should make sure that the doc files are all properly formatted and its a bad sign if this fails.

siddharthist:
@smile13241324 I'll rebase, though it doesn't seem like the failure is related to my changes:
```
TESTING DOCUMENTATION FORMATTING:
================================================================================
Cannot open load file: No such file or directory, /opt/spacedoc/emacs-tools/docfmt/run.el
Documentation formatting script failed
Exited with code 2
```

smile13241324:
@JAremko can it be that we have a problem with `spacemacs-base dist` tests in the CI environment?
Just reviewed this PR and noticed that even after rebasing develop the test fails during installation of some package.

JAremko:
@smile13241324  Looks like it's melpa again... I'll rerun it.

siddharthist:
Thanks for the reviews, everyone!

siddharthist:
ping @sdwolfz

sdwolfz:
Thank you :heart:!

PR #11313, opened 2018-09-11
Add godoc installation into go layer README

sei40kr:
go-mode uses multiple tools for doc view, and `godoc-at-point` (`, hh`) uses `godoc` only, so added godoc installation into README.


Compro-Prasad:
@sdwolfz This should be a clean merge.

duianto:

PR #11320, opened 2018-09-14
wip: Use a lsp server for dockerfile-mode

sei40kr:
# Changes

* Use `lsp-dockerfile` in dockerfile-mode
* Enable company
* Enable flycheck

# TODOs

* [x] ~Update README~
* [x] ~Add a key binding to format buffer~

### DIFF
--- a/layers/+tools/docker/README.org
+++ b/layers/+tools/docker/README.org
@@ -6,6 +6,8 @@
 - [[#description][Description]]
   - [[#features][Features:]]
 - [[#install][Install]]
+  - [[#lsp][LSP]]
+- [[#options][Options]]
 - [[#usage][Usage]]
   - [[#tramp-access-to-docker-containers][TRAMP access to docker containers]]
 - [[#key-bindings][Key bindings]]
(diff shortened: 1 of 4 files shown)


sei40kr:
Should we add `+lang/dockerfile` layer? Current `docker` layer contains a syntax highlighting for Dockerfile and a Docker client using helm.

smile13241324:
Not sure what the others think about it, but in my opinion we should try to keep the layers organised this is just have one Docker layer providing all the docker functionality which includes dockerfile syntax highlighting. Otherwise we may end up with a lot of micro layers.

Miciah:
I enable the Docker layer for docker-tramp and dockerfile-mode.  I do not use the docker package; however, installing the package does not break anything, so why put the burden of explicitly enabling a proliferation of layers on the people who do want the packages?

It does occur to me that the layer could check for the `docker` executable and enable the docker package only if the executable were found.  If that seems like a good idea, I can work on a PR to do that.

The situation with LSP is even better.  All of the enabled language layers that support LSP install LSP-related packages if, and only if, the LSP layer is enabled.

layers/+tools/docker/README.org
```diff
@@ -28,6 +30,28 @@ file.
...
 For syntax checking the external library =hadolint= is also necessary.
 
+** LSP
+This layer can be used with [[https://github.com/rcjsuen/dockerfile-language-server-nodejs][=dockerfile-language-server-nodejs=]] and emacs
+[[https://github.com/emacs-lsp/lsp-mode][=lsp-mode=]] to provide richer, IDE-like capabilities.
+To use this layer with lsp, you must do the following:
+
+1. Add =lsp= to =dotspacemacs-configuration-layers=.
+2. Install =dockerfile-language-server-nodejs= by running
+   =npm i -g dockerfile-language-server=
```
dscole: I believe you meant
`npm i -g dockerfile-language-server-nodejs`
Because the original didn't work for me.

sei40kr: Thanks! I'll fix it.

smile13241324:
@Miciah I like your idea of only enabling docker package when the docker executable can be found. You could control that via the `:toggle` keyword like in the python layer.

sei40kr:
I made this wip again because it doesn't support the latest lsp.

sei40kr:
I'm closing this since `lsp-dockerfile` doesn't work with the latest lsp.

Issue #11229, opened 2018-08-28
Feature request: Layer specific purpose configuration


smile13241324:
Currently `Purpose` comes with a minimal config from it's own layer. The default config can then be adjusted on a per user basis.

I find this problematic as many layers have specific kinds of buffers which require config changes on a per layer basis i.e. adding `Anaconda buffer` to a non `general` category for `python layer`. This requires users to maintain a rather big general purpose config or to use a minimal configuration instead.

Wouldn't it be better to have each layer provide a purpose config for its specific buffers via the post-init mechanic? Purpose-layer could expose a wider default list of purposes including `help` or `compile` and the layers could map the buffers respectively.

Would this be a good idea? Or am I getting too far into personal config territory here?

sdwolfz:
This area is beyond my knowledge, I summon @syl20bnr and @bmag to answer your questions.

bmag:
> Would this be a good idea? Or am I getting too far into personal config territory here?

This could be a good idea, depending on how far we take it. On the one hand, giving Anaconda's documentation buffer a `Help` purpose is pretty basic and reasonable, although it may not be the best example since other help buffers are configured as popup windows (by adding them to `popwin:special-display-config`, see `spacemacs-visual/init-popwin` for example). On the other hand, giving all Python buffers a `python` purpose is a bit too much, since then it wouldn't share the same purpose with other editable buffers - for example C and Python buffers have different purposes, which could be wrong for most users.

> Wouldn't it be better to have each layer provide a purpose config for its specific buffers via the post-init mechanic?

I think it is good approach for how layers add their own purposes, and quite easy to implement as well. See below for proposed implementation.

> Purpose-layer could expose a wider default list of purposes including `help` or `compile` and the layers could map the buffers respectively.

It isn't necessary. Layers can use purposes without pre-defining the purposes elsewhere. As long as different layers use the same purpose for buffers that have similar purposes, we should be fine. There should be a list of purposes used in Spacemacs, likely in `spacemacs-purpose/README.org`, but we can build it organically and gradually when the need arises. One caveat to remember is that purposes shouldn't have the same name as existing functions (e.g. `help` is no good, but `Help` is ok), otherwise their display can't be customized in `purpose-special-action-sequences` (due to implementation details).

----

Proposed implementation:
[...]

PR #11324, opened 2018-09-14
Add python layer specific purpose config

smile13241324:
I have added a python-layer specific purpose config.
Also I have added two macros to spacemacs-purpose to make configuring easier.

This is part of #11229

### DIFF
--- a/layers/+lang/python/packages.el
+++ b/layers/+lang/python/packages.el
@@ -10,41 +10,41 @@
 ;;; License: GPLv3
 
 (setq python-packages
-  '(
-    company
-    counsel-gtags
-    cython-mode
-    eldoc
-    evil-matchit
-    flycheck
-    ggtags
-    helm-cscope
-    helm-gtags
-    (helm-pydoc :requires helm)
-    importmagic
-    live-py-mode
-    (nose :location local)
-    org
-    pip-requirements
-    pipenv
-    pippel
-    py-isort
-    pyenv-mode
-    (pylookup :location local)
-    pytest
-    (python :location built-in)
-    pyvenv
-    semantic
-    smartparens
-    stickyfunc-enhance
-    xcscope
-    yapfify
-    ;; packages for anaconda backend
-    anaconda-mode
-    (company-anaconda :requires company)
-    ;; packages for lsp backend
-    (lsp-python :requires lsp-mode)
-    ))
+      '(
+        company
+        counsel-gtags
+        cython-mode
+        eldoc
+        evil-matchit
+        flycheck
+        ggtags
+        helm-cscope
+        helm-gtags
+        (helm-pydoc :requires helm)
+        importmagic
+        live-py-mode
+        (nose :location local)
+        org
+        pip-requirements
+        pipenv
+        pippel
+        py-isort
+        pyenv-mode
+        (pylookup :location local)
+        pytest
+        (python :location built-in)
+        pyvenv
+        semantic
+        smartparens
+        stickyfunc-enhance
+        window-purpose
+        xcscope
+        yapfify
+        ;; packages for anaconda backend
+        anaconda-mode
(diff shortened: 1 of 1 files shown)


layers/+lang/python/packages.el
```diff
@@ -105,6 +105,15 @@
         "hh" 'anaconda-mode-show-doc
         "gu" 'anaconda-mode-find-references))))
 
+(defun python/post-init-window-purpose ()
+  (add-hook 'python-mode-hook (lambda ()
+                                (spacemacs|window-purpose-define-config
+                                 python (:name-purposes '(("*compilation*" . compile))
```
bmag: The compilation buffer isn't "owned" by the python layer, so this is a bad place to configure its purpose. Also, it's better to configure by mode than by name when possible, so in this case use `compilation-mode` instead of `"*compilation*"`.

Note that `compilation-mode` already has a `search` purpose (a bad name for a purpose, but that's an upstream issue), but we can override it.

layers/+lang/python/packages.el
```diff
@@ -105,6 +105,15 @@
...
+                                                        :regexp-purposes '(("^\*Pydoc" . help)
+                                                                           ("^\*Pylookup" . help)
+                                                                           ("^\*Anaconda" . help)
+                                                                           ("^\*live-py" . compile)))))))
```
bmag: `compile` and `help` are not good purpose names, because there are also functions with the same names (this causes problems if configuring `purpose-special-action-sequences`). Use `Compile` and `Help` instead

bmag:
I think we shouldn't use `<layer>-configure-purposes-p` variables. Instead, users can rely on the `:packages` keyword in their layers list. For example, python layer can be listed as:
```elisp
(python :packages (not window-purpose)) ;; not sure about exact syntax, but something like this
```

About `spacemacs|window-purpose-define-config-p`, I think it's an unnecessary abstraction, and it will easier to understand the code if it used window-purpose's API directly.

Regarding the hook at `python-mode-hook`, that's just wrong. The purpose configuration is defined globally, and setting the layer's purpose configuration should only happen once. After removing the macros, it would be easy to replace it with:
```elisp
(with-eval-after-load 'window-purpose
  (purpose-set-extension-configuration
    :python-layer
    (purpose-conf ...)))
```

Instead of macros and their docstrings, we should rely on documentation in spacemacs-purpose layer's README.org or in LAYERS.org. LAYERS.org used to have a "How do I ... Idiomatically" section (it's still there in master branch), which we can resurrect and add a subsection for how a layer can configure its own purposes. I should have time to write the documentation this week - it won't be much different than what we discussed in the mentioned issue.

layers/+lang/python/packages.el
smile13241324: Ups, didn't thought about the compilation buffer's owner status. 
Ok I will use modes instead of names.

layers/+lang/python/packages.el
smile13241324: Hmm you are right, you have mentioned that in the original post ... I will fix it.

smile13241324:
Too bad, I was really proud of these macros, to be honest. But you are right I will remove the layers variable and replace `spacemacs|window-purpose-define-config` with your proposed code.

However I still have a question, why do you use `with-eval-after-load` in this context?
Wouldn't it suffice to run
``` elisp
(defun python/post-init-window-purpose ()
  (purpose-set-extension-configuration
    :python-layer
    (purpose-conf ...)))
```
in this case? I would have expected that this is only executed after window-purpose is properly loaded.

layers/+lang/python/packages.el
smile13241324: Done

smile13241324:
I have made the required changes @bmag would you mind to check again?

smile13241324:
I think this is to old now to be of use

PR #11328, opened 2018-09-15
bibtex: cleanup layer and add ivy support

ellakk:
Was initially just going to add ivy support but did some small cleaning of the layer also:
biblio-/core is pulled in as dependencies by multiple packages in this layer already and their package declarations was empty and undocumented in the readme so i removed them. Users can access the functionality biblio offers from helm/ivy-bibtex. helm-bibtex was downloaded in this layer but given no keybinding so i bound helm-bibtex/ivy-bibtex to SPC mm since it is used to manage your bibtex library.

### DIFF
--- a/layers/+lang/bibtex/packages.el
+++ b/layers/+lang/bibtex/packages.el
@@ -13,14 +13,19 @@
       '(
         auctex
         (helm-bibtex :requires helm)
+        (ivy-bibtex :requires ivy)
         markdown-mode
         org
         org-ref
         ))
 
 (defun bibtex/post-init-auctex ()
-  (spacemacs/set-leader-keys-for-major-mode 'latex-mode
-    "ic" 'org-ref-helm-insert-cite-link))
+  (cond ((configuration-layer/layer-used-p 'helm)
+         (spacemacs/set-leader-keys-for-major-mode 'latex-mode
+           "ic" 'org-ref-helm-insert-cite-link))
+        ((configuration-layer/layer-used-p 'ivy)
+         (spacemacs/set-leader-keys-for-major-mode 'latex-mode
+           "ic" 'org-ref-ivy-insert-cite-link))))
 
 (defun bibtex/init-helm-bibtex ()
   (use-package helm-bibtex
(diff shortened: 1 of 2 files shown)


leezu:
I'm not sure if the keybinding is useful, as the main motivation of `helm-bibtex` / `ivy-bibtex` is to provide easy access to the bibliography from any other buffer. The keybinding proposed here is only active when in viewing a bibtex file.

smile13241324:
This is pretty old now but could still be useful. I am to revise bibtex layer anyway so lets check whether these changes can be incorporated. If it is useful I will cherry pick this PR.

layers/+lang/bibtex/packages.el
```diff
@@ -12,27 +12,57 @@
...
+        ((configuration-layer/layer-used-p 'ivy)
+         (spacemacs/set-leader-keys-for-major-mode 'org-mode
+           "ic" 'org-ref-ivy-insert-cite-link))))
 
 (defun bibtex/init-org-ref ()
   (use-package org-ref
     :defer t
     :commands (org-ref-bibtex-next-entry
                org-ref-bibtex-previous-entry
+               org-ref-ivy-insert-cite-link
```
smile13241324, line 65: Must also be done for `org-ref-helm-insert-cite-link`

layers/+lang/bibtex/packages.el
```diff
@@ -46,6 +76,8 @@
                pubmed-insert-bibtex-from-pmid)
     :init
     (progn
+      (add-hook 'org-mode-hook (lambda () (require 'org-ref)))
```
smile13241324, line 79: Do not require org-ref here, autoloads must take care to load it.

layers/+lang/bibtex/packages.el
```diff
@@ -72,16 +104,8 @@
         "lA" 'arxiv-get-pdf-add-bibtex-entry
         "ld" 'doi-utils-add-bibtex-entry-from-doi
         "li" 'isbn-to-bibtex
-        "lp" 'pubmed-insert-bibtex-from-pmid))))
-
-(defun bibtex/pre-init-org-ref ()
-  (add-hook 'org-mode-hook (lambda () (require 'org-ref))))
-
-(defun bibtex/post-init-markdown-mode ()
```
smile13241324, line 80: Why is markdown inserts removed here?

smile13241324:

PR #11340, opened 2018-09-19
 Check for universal argument before paste

bet4it:
We should use `(interactive "*P")` in `spacemacs/evil-mc-paste-before` and `spacemacs/evil-mc-paste-after` to get the raw prefix argument.

`evil-paste-before` and `evil-paste-after` don't support plain prefix argument, so if we press <kbd>SPC u p</kbd> in `spacemacs`, we want to paste one time without indenting.
But there is already a meaning of <kbd>C-u C-y</kbd> in `emacs`, which is _put point at beginning, and mark at end_ when yanking. So it's ambiguous when we trigger <kbd>C-u C-y</kbd> in `spacemacs`.We may want to yank with the point at beginning or just yank without indenting.
We do both of them before, which looks like a misuse. After this pull request, <kbd>C-u C-y</kbd> will only yank without indenting. There may be some `emacs` users confuse with it, but if they don't work on `spacemacs-indent-sensitive-modes`, they should also be confused before.
And as there is only one `universal-argument` in `emacs`, we couldn't do something like _paste four times with auto indent disabled temporarily_. So maybe we should find another way to temporarily disable auto indent?

By the way, as `defadvice` is obsolete, I replace the use of `spacemacs|advise-commands` with `advice-add`.

### DIFF
--- a/layers/+spacemacs/spacemacs-evil/funcs.el
+++ b/layers/+spacemacs/spacemacs-evil/funcs.el
@@ -61,7 +61,6 @@ Otherwise, revert to the default behavior (i.e. enable `evil-insert-state')."
       (evil-escape-mode t)
     (evil-escape-mode -1)))
 
-
 (defun spacemacs/linum-relative-toggle ()
   (interactive)
   (if (not (bound-and-true-p linum-relative-mode))
--- a/layers/+lang/swift/packages.el
+++ b/layers/+lang/swift/packages.el
@@ -25,14 +25,14 @@
     :defer t
     :init
     (progn
-      (spacemacs|advise-commands "store-initial-buffer-name"
-                                 (swift-mode-run-repl) around
-       "Store current buffer bane in bufffer local variable,
+      (defun spacemacs//swift-store-initial-buffer-name (func &rest args)
+        "Store current buffer bane in bufffer local variable,
 before activiting or switching to REPL."
-       (let ((initial-buffer (current-buffer)))
-         ad-do-it
-         (with-current-buffer swift-repl-buffer
-           (setq swift-repl-mode-previous-buffer initial-buffer))))
+        (let ((initial-buffer (current-buffer)))
+          (apply func args)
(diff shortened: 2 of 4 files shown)


layers/+spacemacs/spacemacs-evil/funcs.el
```diff
@@ -61,7 +61,6 @@ Otherwise, revert to the default behavior (i.e. enable `evil-insert-state')."
       (evil-escape-mode t)
     (evil-escape-mode -1)))
 
-
```
duianto: This is a form-feed character, it's used in several places, for example in the Spacemacs buffer, and even 2 other times in this same file `spacemacs-evil/funcs.el`.

They just look like empty lines in the github diff, but they look like this in Spacemacs:

emacs_2019-04-24_16-44-28

In other words, they should not be removed.

duianto:
Another PR also suggests adding a universal argument to paste:
Fix pasting with universal argument #11858

How do this PR compare to that one?

bet4it:
What has done in #11858 has already been included in this pull request.
And this pull request adds two extra changes:
1. Fix the errors related to universal argument when people enable `multiple-cursors` layer.
2. Rewrite `spacemacs|advise-commands` with `advice-add`, as `defadvice` is obsoleted.

layers/+spacemacs/spacemacs-evil/funcs.el
bet4it: Yes, I know it.
But in this file, all the other form-feed characters are placed right before a comment, and I think it means start of a new section. But this one doesn't like this, so I wonder whether we should delete it.

duianto: You might be right, it seems logical to only add one before a comment/section. So it's probably fine to remove this one.

duianto:
Ok then we'll prefer this one over #11858

duianto:

PR #11353, opened 2018-09-20
layers: git: doc: fx,rw Repository list

Anton-Latukha:
1. `magit-repository-directories` no longer handles directory-only entries, only special form _(DIRECTORY . DEPTH)_ is accepted:

```elisp
(defcustom magit-repository-directories nil
  "List of directories that are or contain Git repositories.

Each element has the form (DIRECTORY . DEPTH).  DIRECTORY has
to be a directory or a directory file-name, a string.  DEPTH,
an integer, specifies the maximum depth to look for Git
repositories.  If it is 0, then only add DIRECTORY itself."
  :package-version '(magit . "2.8.0")
  :group 'magit-essentials
:type '(repeat (cons directory (integer :tag "Depth"))))
```
(From https://github.com/magit/magit/blob/59e3891fde37b2b0937b54ffdd685703614360ad/lisp/magit-repos.el#L40-L49)

Additional info:
* Relevant doc issue that is solved upstream: https://github.com/magit/magit/issues/3475
* Magit manual: https://magit.vc/manual/magit.html#Status-Buffer

2. `Magit auto-complete` and `Repository list` doc entries had descriptions of same configuration processes. Now  after some rewrite - `Magit auto-complete` points to relevant info in `Repository list`.

### DIFF
--- a/layers/+source-control/git/README.org
+++ b/layers/+source-control/git/README.org
@@ -59,13 +59,10 @@ To display the =magit status= buffer in fullscreen set the variable
 Magit auto-complete feature is enabled. For this feature to work best you
 have to setup your Git repository directory in your =dotspacemacs/user-config=
 function, this is the folder where you keep all your git-controlled projects
-(the path should end up with a ~/~ to respect Emacs conventions):
 
 #+BEGIN_SRC emacs-lisp
   (setq magit-repository-directories '("~/repos/"))
 #+END_SRC
-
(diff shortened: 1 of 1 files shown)


Anton-Latukha:
https://github.com/magit/magit/blob/59e3891fde37b2b0937b54ffdd685703614360ad/lisp/magit-repos.el#L40-L49

Anton-Latukha:
@syl20bnr
Notifying on clean PR.
People a couple of times already were bothered figuring this out.
This is straight-forward sync of the `magit-repository-directories` documentation with magit project.

Thank you.

PR #11361, opened 2018-09-22
Integrate helm-cider in clojure layer

Nimamoh:


kommen:
How does this PR affect people who use ivy instead of helm with Spacemacs?

0atman:
This is an important fix, as currently this package tries to get installed every startup, which means the package lists are refreshed every startup, which blocks the UI.
Thank you! +1

sdwolfz:
I've merged #11839 for the next release. Please rebase this if you want the helm package, and it will be part of the next release.

Nimamoh:
Updated the documentation and the clojure layer to take #11839 in account.
This PR now uses helm-cider when helm is used.
To answer @kommen . When using ivy, regular cider-apropos and cider-cheatsheet are used, I just tested that.

smile13241324:
I'll check whether this is still relevant and merge it if it is.

smile13241324:

Issue #11343, opened 2018-09-19
json/prettier reformatting SyntaxError in orgmode BEGIN_SRC block


tko:
#### Description :octocat:
json/prettier reformatting doesn't work in orgmode BEGIN_SRC block, instead prettier complains about "SyntaxError"

The same content when edited in a .json file reformats without problem.

#### Reproduction guide :beetle:
- Start Emacs
- Create test.org file containing
  ```org
  #+BEGIN_SRC json
  {"type":
   1}
  #+END_SRC
  ```
- Move within the `BEGIN_SRC` block and press `, '` (org-edit-special)
- Press `, =` to reformat the JSON in the buffer

*Observed behaviour:* :eyes: :broken_heart:
```
prettier errors:
[error] stdin: SyntaxError: Unexpected token, expected ";" (1:8)
[error] > 1 | {"type":
[error]     |        ^
[error]   2 |  1}
[error]   3 | 
```

*Expected behaviour:* :heart: :smile:
The buffer content is pretty-printed, something like
```org
#+BEGIN_SRC json
{
  "type: 1
}
#+END_SRC
```

#### System Info :computer:
- OS: darwin
- Emacs: 26.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 9865e7731)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(helm emacs-lisp neotree
      (javascript :variables javascript-fmt-tool 'prettier)
      (json :variables json-fmt-tool 'prettier))
```
- System configuration features: NOTIFY ACL LIBXML2 ZLIB TOOLKIT_SCROLL_BARS THREADS

Miciah:
In your example `BEGIN_SRC` blocks, the example json is missing a double-quote: `"type:` should be `"type":`.  However, from prettier's error message, it looks like you actually tested with `"type":`.  So just to be clear, are you getting the reported error message with `"type":`?

tko:
I'm testing with valid json, so with `"type":` .. not sure what happened there.

tko:
I'm seeing the same error if just I pipe json through `prettier`. Apparently for json you need `prettier --parser=json`. There's `prettier-js-args` variable that can pass arguments to prettier, I guess that should be prefilled for json.

tko:
[...]

PR #11364, opened 2018-09-23
json: tell prettier content is to be parsed as json

tko:
In json-mode tell prettier the content is always to be parsed as json.
Fixes formatting in orgmode `#+BEGIN_SRC json` blocks as prettier
otherwise expects content as javascript.

Fixes: #11343

### DIFF
--- a/layers/+lang/json/packages.el
+++ b/layers/+lang/json/packages.el
@@ -57,8 +57,9 @@
       "hp" 'jsons-print-path)))
 
 (defun json/pre-init-prettier-js ()
-  (if (eq json-fmt-tool 'prettier)
-      (add-to-list 'spacemacs--prettier-modes 'json-mode)))
+  (when (eq json-fmt-tool 'prettier)
+      (add-to-list 'spacemacs--prettier-modes 'json-mode)
+      (add-hook 'json-mode-hook #'spacemacs/json-setup-prettier)))
 
 (defun json/pre-init-web-beautify ()
   (if (eq json-fmt-tool 'web-beautify)
--- a/layers/+lang/json/funcs.el
+++ b/layers/+lang/json/funcs.el
@@ -32,3 +32,8 @@ If ARG is a numerical prefix argument then specify the indentation level."
     (if (equal start end)
         (save-excursion (json-reformat-region (point-min) (point-max)))
       (json-reformat-region start end))))
+
+(defun spacemacs/json-setup-prettier ()
+  "Tell prettier the content is to be parsed as JSON regardless of any file
+extensions."
+  (setq-local prettier-js-args '("--parser=json")))


layers/+lang/json/packages.el
```diff
@@ -32,6 +32,9 @@
   (use-package json-mode
     :defer t))
 
+(defun json/post-init-json-mode ()
```
d12frosted: How this function is being called?

tko: Isn't it the same layer calling conventions magic that calls the `{layer}/pre-init-{pkg}` and `{layer}/init-{pkg}` packages?

tko:
I see the options being
1. each major mode layer knows about prettier
2. prettier layer knows about each major mode / layer

for prettier the latter would be more convenient sure, but then it'd be leaking into major modes/layers and seems like major mode layer is meant to own everything about major mode. Haven't thought about too much, just went this way as that's the way it was already being setup.

I left the hook function always implemented and check for the `json-fmt-tool` there so the variable can be change on runtime and change the behavior dynamically without requiring restart. No strong feelings though.

layers/+lang/json/packages.el
d12frosted: @tko you're right. Sorry, I was confused. Then why this `add-hook` call is not inside of `json/init-json-mode`?

d12frosted:
@tko what you say makes sense to me. Let's leave these configurations in `json` layer.

Also, I've just took a closer look at this layer and noticed that there are other handlers for `json-fmt-tool`. But they are implemented differently. For example,

```emacs-lisp
(defun json/pre-init-web-beautify ()
  (if (eq json-fmt-tool 'web-beautify)
      (add-to-list 'spacemacs--web-beautify-modes
                   (cons 'json-mode 'web-beautify-js))))
```

To be honest, I like this approach a little bit more as it clearly/explicitly defines the dependencies.

On the other hand, I like that your approach allows one to switch `json-fmt-tool` on the fly 😸 

BTW, there are already configurations for `prettier`. Should not you just move your `add-hook` call to `json/pre-init-prettier-js`?

layers/+lang/json/packages.el
syl20bnr: Actually this should be the other way around. This is the `post-init-prettier` that setup the hook. Indeed `json-init` does not know anything about `prettier` being used or not but `post-init-prettier` knows it implicitly as it is called only if `prettier` is used (i.e. the `prettier` layer is used).

d12frosted: Yes, exactly 😸 Just like `web-beautify` is configured in the `json` layer.

tko:
Updated to do things in `pre-init-prettier-js` within the scope of the existing conditional. The finer points between using package :toggles, pre-init (like web-beautify) or post-init (like add-node-modules-path) escape me.

d12frosted:

Resulting commit is 895308aec.

PR #11365, opened 2018-09-23
Prettier support for markdown

ztlevi:
Since prettier is great for markdown as well, I added it to markdown layers.

Personally, I really like `prettier-js-mode`. It just a mode bound with `after-save-hook` but really saves me a lot of time, and I believe it's always nice for `C-s/s-s` masters. So maybe it's time to add `prettier-js-mode` to spacemacs.

### DIFF
--- a/layers/+lang/markdown/config.el
+++ b/layers/+lang/markdown/config.el
@@ -37,3 +37,6 @@ generate mmm classes.")
 
 (defvar markdown--key-bindings-modes '(markdown-mode gfm-mode)
   "Modes using markdown key bindings.")
+
+(defvar markdown-fmt-tool 'prettier
+  "The formatter to format a JavaScript file.")
--- a/layers/+lang/markdown/README.org
+++ b/layers/+lang/markdown/README.org
@@ -8,6 +8,7 @@
   - [[#bibtex][BibTeX]]
 - [[#install][Install]]
 - [[#configuration][Configuration]]
+  - [[#formatting][Formatting]]
   - [[#live-preview][Live preview]]
   - [[#automatic-mmm-mode-generation][Automatic MMM-Mode Generation]]
 - [[#usage][Usage]]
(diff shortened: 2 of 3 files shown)


WarFox:
Thank you for doing this. I think this is a good idea.

If you let me know the expected behaviour, I shall test this in my setup

ztlevi:
Yes, it is expected. 

But what I mentioned is we probably need to add some options for users to autoformat on save.

Something like this in my config:
```
(spacemacs/add-to-hooks 'prettier-js-mode '(js2-mode-hook
                                                  typescript-mode-hook
                                                  typescript-tsx-mode-hook
                                                  rjsx-mode-hook
                                                  json-mode-hook
                                                  css-mode-hook
                                                  markdown-mode-hook
                                                  gfm-mode-hook))
```

We can activate autoformat by setting `prettier-format-on-save`. Or setting `prettier-js-mode` for different languages' on-save option. In this case, we might have `js/typescript/react/json/css/markdown-format-on-save` variables.

WarFox:
`(spacemacs/add-to-hooks 'prettier-js-mode )` does it for me. Do we need to set `prettier-js-mode` for different languages?

layers/+lang/markdown/config.el
```diff
@@ -37,3 +37,6 @@ generate mmm classes.")
 
 (defvar markdown--key-bindings-modes '(markdown-mode gfm-mode)
   "Modes using markdown key bindings.")
+
+(defvar markdown-fmt-tool 'prettier
```
d12frosted: AFAIK, you need to install `prettier` using `npm`. If so, it is a bad default as not every `markdown` user uses `npm`. And what happens if you don't have `prettier` in your `PATH`? Do you see errors when you edit/save markdown buffers?

layers/+lang/markdown/config.el
```diff
@@ -37,3 +37,6 @@ generate mmm classes.")
...
+  "The formatter to format a JavaScript file.")
```
d12frosted: Shouldn't it be `The formatter to format a markdown file.`?

PR #11383, opened 2018-10-03
`spacemacs-theme'; no overrides.

dschrempf:
Do not override customizable settings from `spacemacs-theme', since those should
be set on a per user basis in the configuration file.


layers/+distributions/spacemacs-bootstrap/packages.el
```diff
@@ -543,7 +543,4 @@
 
 (defun spacemacs-bootstrap/init-spacemacs-theme ()
   (use-package spacemacs-theme
-    :defer t
-    :init (setq spacemacs-theme-keyword-italic t
-                spacemacs-theme-comment-bg t
-                spacemacs-theme-org-height t)))
+    :defer t))
```
robbyoconnor: You can override these at will...

dschrempf: That's correct. However, I had trouble setting those variables.

When having them in `spacemacs/user-init`, they are not honored upon startup by the theme because they are set too late (at least that is what happened to me, maybe because I am starting an Emacs server during boot).

When having them in `spacemacs/init` they are overridden by the lines that this pull request removes.

See also https://github.com/nashamri/spacemacs-theme/issues/122

robbyoconnor: I see...nevermind :)

duianto:

PR #11384, opened 2018-10-03
Make Treemacs resizable by default

deb0ch:
The treemacs window is not resizable by default and it is very annoying when having long file names and/or a deep directory structure.

I don't understand why the width of this window is locked by default, I think it is more sensible to have it unlocked.

### DIFF
--- a/layers/+filetree/treemacs/README.org
+++ b/layers/+filetree/treemacs/README.org
@@ -10,6 +10,7 @@
   - [[#follow-mode][Follow mode]]
   - [[#file-watch][File watch]]
   - [[#collapsed-directories][Collapsed directories]]
+  - [[#locking-width][Locking width]]
 - [[#key-bindings][Key Bindings]]
   - [[#global][Global]]
   - [[#inside-treemacs][Inside Treemacs]]
@@ -53,7 +54,7 @@ layer variable =treemacs-use-follow-mode= to non-nil.
     (treemacs :variables treemacs-use-follow-mode t)))
 #+END_SRC
 
-Default value is =t=.
+Default is =t=.
 
 ** File watch
 To automatically refresh the Treemacs buffer when there is a change in the
@@ -65,7 +66,7 @@ part of the file system shown by treemacs set the layer variable
     (treemacs :variables treemacs-use-filewatch-mode t)))
 #+END_SRC
 
-Default value is =t=.
+Default is =t=.
 
 ** Collapsed directories
 *This feature requires python to be installed*.
@@ -79,7 +80,19 @@ layer variable =treemacs-use-collapsed-directories= to a positive number.
     (treemacs :variables treemacs-use-collapsed-directories 3)))
 #+END_SRC
 
-Default value is 3 (or 0 when python is not installed).
+Default is 3 (or 0 when python is not installed).
+
+** Locking width
+To have the width of the treemacs window locked by default, meaning not manually
+resizable, set the variable =treemacs-lock-width= to non-nil. It will still be
+resizable through Treemacs commands and keybindings.
+
+#+BEGIN_SRC emacs-lisp
+  (setq-default dotspacemacs-configuration-layers '(
+    (treemacs :variables treemacs-lock-width t)))
+#+END_SRC
+
+Default is =nil=.
 
 * Key Bindings
 ** Global
--- a/layers/+filetree/treemacs/config.el
+++ b/layers/+filetree/treemacs/config.el
@@ -29,3 +29,6 @@ There are 2 possible values:
 1) simple, which highlights only files based on their git status, and is
    slightly faster
 2) extended, which highlights both files and directories, but requires python")
+
+(defvar treemacs-lock-width nil
+  "When non-nil the treemacs window will not be manually resizable by default.")
--- a/layers/+filetree/treemacs/packages.el
+++ b/layers/+filetree/treemacs/packages.el
@@ -67,7 +67,8 @@
         (treemacs-filewatch-mode t))
       (when (memq treemacs-use-git-mode '(simple extended))
         (treemacs-git-mode treemacs-use-git-mode))
-      (add-to-list 'spacemacs-window-split-ignore-prefixes treemacs--buffer-name-prefix))))
+      (add-to-list 'spacemacs-window-split-ignore-prefixes
+                   treemacs--buffer-name-prefix))))
 
 (defun treemacs/init-treemacs-evil ()
   (use-package treemacs-evil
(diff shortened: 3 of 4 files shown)


deb0ch:
CircleCI tests failure seems to not be related to this PR:

```
(Spacemacs) --> installing package: overseer@emacs-lisp... [15/15]
Cannot open load file: No such file or directory, bind-map
../../../../spacemacs.mk:39: recipe for target 'func_tests' failed
make: *** [func_tests] Error 255
Exited with code 2
```

layers/+filetree/treemacs/packages.el
```diff
@@ -41,6 +41,10 @@
             treemacs-never-persist nil
             treemacs-goto-tag-strategy 'refetch-index
             treemacs-collapse-dirs treemacs-use-collapsed-directories)
+      (add-hook 'treemacs-mode-hook
+                (lambda ()
+                  (when treemacs--width-is-locked
+                    (treemacs-toggle-fixed-width))))
```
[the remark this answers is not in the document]
d12frosted: Ah, sorry. I just tried to use `treemacs` and it turns out that I misunderstood the meaning of non-fixed width. It doesn't change the width automatically based on content, but rather allows the user to manually change the width of `treemacs` buffer.

Then my previous comment is not that critical, but I still think that it is a good idea to have an ability to enable fixed-width with a variable in `.spacemacs`. Plus, we already have several similar variables (e.g. `treemacs-use-follow-mode`).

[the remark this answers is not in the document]
deb0ch: Your second comment is accurate. Makes sense, will do. I will make it a layer variable since it is part of the treemacs layer.

It's just that in the first place I'm having a lot of troubles imagining for what reason anybody would want it locked. But at the same time, it being implemented that way means that at least one person in the world has a reason for that so a configuration variable makes sense.

deb0ch:
Updated, configuration variable added :thumbsup::sparkles:

layers/+filetree/treemacs/README.org
```diff
@@ -53,7 +54,7 @@ layer variable =treemacs-use-follow-mode= to non-nil.
     (treemacs :variables treemacs-use-follow-mode t)))
 #+END_SRC
 
-Default value is =t=.
+Default is =t=.
```
d12frosted: To be honest, I don't really care about this particular change. But I see both styles are being used in Spacemacs, so I don't even know what to do about it. 😸

layers/+filetree/treemacs/README.org
```diff
@@ -79,7 +80,20 @@ layer variable =treemacs-use-collapsed-directories= to a positive number.
     (treemacs :variables treemacs-use-collapsed-directories 3)))
 #+END_SRC
 
-Default value is 3 (or 0 when python is not installed).
+Default is 3 (or 0 when python is not installed).
+
+** Locking width
+To have the width of the treemacs window locked by default, meaning not manually
+resizable.
+Note that it will still be resizable using =treemacs-toggle-fixed-width= or
```
d12frosted: I think you should place an empty line here,

layers/+filetree/treemacs/README.org
```diff
@@ -79,7 +80,20 @@ layer variable =treemacs-use-collapsed-directories= to a positive number.
     (treemacs :variables treemacs-use-collapsed-directories 3)))
 #+END_SRC
 
-Default value is 3 (or 0 when python is not installed).
+Default is 3 (or 0 when python is not installed).
+
+** Locking width
+To have the width of the treemacs window locked by default, meaning not manually
```
d12frosted: Sorry for nitpicking, but I don't understand this description. If you wanted to copy the description style of `treemacs-use-filewatch-mode`, then you missed the part where it says what to do in order to get the desired behaviour. Like this:

> To have the width of the treemacs window locked by default (meaning the inability to resize treemacs window) set the layer variable `treemacs-lock-width` to non-nil.

layers/+filetree/treemacs/config.el
```diff
@@ -29,3 +29,6 @@ There are 2 possible values:
 1) simple, which highlights only files based on their git status, and is
    slightly faster
 2) extended, which highlights both files and directories, but requires python")
+
+(defvar treemacs-lock-width nil
+  "Make the treemacs window not manually resizable by default.")
```
d12frosted: Here I would use more popular doc style - 'If non-nil blah-blah'/'When non-nil blah-blah'. Just like other 'boolean' variables are documented.

layers/+filetree/treemacs/packages.el
```diff
@@ -64,7 +69,8 @@
         (treemacs-filewatch-mode t))
       (when (memq treemacs-use-git-mode '(simple extended))
         (treemacs-git-mode treemacs-use-git-mode))
-      (add-to-list 'spacemacs-window-split-ignore-prefixes treemacs--buffer-name-prefix))))
+      (add-to-list 'spacemacs-window-split-ignore-prefixes
```
d12frosted: 💯 boy scout rule!

layers/+filetree/treemacs/packages.el
```diff
@@ -41,6 +41,11 @@
             treemacs-never-persist nil
             treemacs-goto-tag-strategy 'refetch-index
             treemacs-collapse-dirs treemacs-use-collapsed-directories)
+      (unless treemacs-lock-width
+        (add-hook 'treemacs-mode-hook
+                  (lambda ()
```
d12frosted: Just a thought. Maybe lambda is not the best thing to place into a hook. But it's minor.

layers/+filetree/treemacs/README.org
deb0ch: Less words, same meaning. And as you said, both styles are already present.

layers/+filetree/treemacs/packages.el
deb0ch: Would you prefer a named function or do you have something else in mind ?

layers/+filetree/treemacs/README.org
deb0ch: You mean between the two sentences ?

layers/+filetree/treemacs/README.org
deb0ch: Rewriting :thumbsup::sparkles:

deb0ch:
Updated ✨

d12frosted:
@deb0ch thank you! I will take a look son 😸

d12frosted:

I didn't want to torture with other requests, so I've created a follow up PR (#11423). Maybe we don't want it at all, but still 😸

deb0ch:
Yay ! Some colleagues of mine were really bitching about it when I showed them my editor, made me want to dig into it 😸 

That and me trying without success to read some long file names 😓

PR #11386, opened 2018-10-03
Added more tips for avoiding pitfalls when building on windows.

Kethku:
I ran into some issues when building the pdumper build of emacs on windows and added some hints to the expiremental.org file to help people avoid them in the future.


d12frosted:

PR #11391, opened 2018-10-04
Enable colors in magit-log

deb0ch:
Colors makes it so much easier to visually differentiate and track the ascii lines representing the commit histories of the different branches 🌈


d12frosted:

duianto:
Magit has removed the `magit-log-arguments` variable. https://github.com/magit/magit/commit/4641aa07e0b1f4654fceb139a36b160d594d7846

But since the line from this PR still was present in Spacemacs, it caused these issues:
Error (use-package): magit/:config: Symbol’s value as variable is void: magit-log-arguments #12193

The fastest solution was to just remove the variable assignment.

@deb0ch Since you added the feature in this PR, maybe you would like to try and figure out how to add back the `--color` argument?

PR #11393, opened 2018-10-05
update to flyspell-correct v0.5

d12frosted:
`flyspell-correct-word-generic` and `flyspell-correct-previous-word-generic` are
deprecated and replaced by `flyspell-correct-at-point` and
`flyspell-correct-wrapper` respectively.

`flyspell-correct-wrapper` has a so-called 'rapid' mode, where you can correct
multiple words in a single run.

### DIFF
--- a/layers/+checkers/spell-checking/README.org
+++ b/layers/+checkers/spell-checking/README.org
@@ -120,13 +120,14 @@ set the layer variable =enable-flyspell-auto-completion= to t:
 
 * Key Bindings
 
-| Key Binding | Description              |
-|-------------+--------------------------|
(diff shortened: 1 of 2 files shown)


costa:
@d12frosted, @sdwolfz, guys, master/stable is broken wrt this (`.../.emacs.d/elpa/flyspell-correct-20200601.944/flyspell-correct.elc failed to define function flyspell-correct-previous-word-generic`). I personally wouldn't like to switch to `develop` just for this, I'm not even sure how to do it from the user interface (I've kept to the user side of Emacs since I'm on Spacemacs, ~5 years now, which is just fantastic by my standards).

PR #11396, opened 2018-10-05
Override read-only when capturing clear in shell

AlexanderWingard:
When shell layer intercepts the clear commands and tries to replace it with
erase-buffer, this fails with 'Text is read-only'.
Work around this by using inhibit-read-only.


sdwolfz:
Thank you :heart:! 

PR #11400, opened 2018-10-07
Adds a convenient keybinding for correcting word at point

agzam:
Adds a `SPC x x` keybinding supplemental to `z=`

## Rationale: 

- `z=` is not a very convenient binding for quickly spell-checking word at point
- having an easily triggered action under `SPC` seems to be more efficient

*To be honest this comes from my personal experience. I am a sloppy typist, I need to correct my mistakes all the time. One could say: "just add it to your own config", which I did. It still makes annoyingly difficult for me when I am pairing with someone at work and working on their computer.*

### DIFF
--- a/layers/+checkers/spell-checking/README.org
+++ b/layers/+checkers/spell-checking/README.org
@@ -128,6 +128,7 @@ set the layer variable =enable-flyspell-auto-completion= to t:
 | ~SPC S d~       | change dictionary                      |
 | ~SPC S n~       | flyspell goto next error               |
 | ~SPC t S~       | toggle flyspell                        |
+| ~SPC x x~       | correct word at point                  |
 
 ** Spell Checking Transient-state
 
--- a/layers/+checkers/spell-checking/keybindings.el
+++ b/layers/+checkers/spell-checking/keybindings.el
@@ -1,0 +1,13 @@
+;;; keybindigs.el --- Spell Checking Layer keybindigs File for Spacemacs
+;;
+;; Copyright (c) 2012-2018 Sylvain Benner & Contributors
+;;
+;; Author: Sylvain Benner <sylvain.benner@gmail.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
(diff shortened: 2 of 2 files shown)


agzam:
Oh... interesting... I didn't not know about `SPC S c`. I started digging more - found your post https://d12frosted.io/posts/2016-05-09-flyspell-correct-intro.html. 

- It is strange that `z=` doesn't work for you, it's should be bound to `ispell-word`

- `SPC S c` - somehow only works in text, markdown, org etc. For me (unlike `ispell-word`) it doesn't work for example in elisp-mode buffer ¯\_(ツ)_/¯

- Also I discovered `C-$`-`flyspell-correct-word-before-point`. That thing sometimes works, sometimes throws an exception. And when it works it shows word choices in a popup, even though `flyspell-correct-interface` is set to `'flyspell-correct-helm'`

Anyway, thank you for your suggestion. I'm gonna close this.

d12frosted:
That blog post is a little bit old 😸 Some major changes were introduced in `v0.5`, but the main idea holds. (though I am going to update that post)

Regarding `SPC S c` not working in some of the modes - you need to enable either `flyspell-mode` (for text buffers) or `flyspell-prog-mode` for programming modes (like `emacs-lisp-mode`). The good thing is that first one is given out of the box when you enable `spell-checking` layer and the second part is given you by settings `spell-checking-enable-by-default` variable to `t` in yours`.spacemacs` file. Give it a try 😸 

`flyspell-correct-word-before-point` is not from `flyspell-correct` package, that's why it doesn't work as you expect. You should use one of the following functions:

- `flyspell-correct-wrapper` (think of it as of `flyspell-correct-dwim`) - in most cases you want to use it.
- `flyspell-correct-previous` - corrects misspelt word at the point or a previous visible one.
- `flyspell-correct-next` - corrects misspelt word at the point or a next visible one.
- `flyspell-correct-at-point` - corrects misspelt word at the point.

All of them use `flyspell-correct-interface`.

> Anyway, thank you for your suggestion.

No problem ❤️

PR #11402, opened 2018-10-07
Make Colemak layout more ergonomic

allentiak:


allentiak:
I think I have added the wrong commits. I will correct this.

PR #11412, opened 2018-10-09
Fix: add prefix to org-mode clock options after recent change.

dmolina:
The commit #ad8eb85 is great, but with SPC a o C there was not name, it said only "prefix" instead of "clock". I have detected it and fixed it with this small change.


dmolina:
The previous commit update which introduce the fixed error was ad8eb85960d05184371cd7c0d477cefff6c12dbf, it was wrongly referred.

d12frosted:

PR #11415, opened 2018-10-10
Fix global toggles for aggressive-indent and smartparens

mjkramer:
Previously the global toggles acted on the local instead of global mode.


d12frosted:
My congratulations on your first contribution 😸 

Issue #5799, opened 2016-04-15
The `navigation' part of `visual-line-navigation' does not take effect until a state transition


et2010:
#### Description
When using ```SPC t L``` to toggle visual line navigation for the first time, visual-line-mode is turned on but navigation between visual lines is not available until a state transition.

#### Reproduction guide
- Start Emacs
- Open any file with long lines
- Hit ```SPC t L```
- Navigate between lines using ```j``` or ```k```

*Observed behaviour:*
Visual-line-mode is turned on, but cusor only moves by hard lines. After a state transition, eg. switch to insert state and switch back, the visual line navigation feature is back. 

*Expected behaviour:*
The visutal line navigation should work without having to do a state transition first.

#### System Info
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.105.18
- Spacemacs branch: nil (rev. 5ee970f)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: helm
- Layers:
```elisp
((auto-completion :disabled-for org)
 better-defaults emacs-lisp evil-snipe font-lock-studio git github latex markdown org citation gtd window-purpose
 (han :variables han-enable-youdao-dict t han-enable-fcitx t han-org-line-spacing 0.2)
 pdf-tools python
 (shell :variables shell-default-height 30 shell-default-position 'bottom)
 shell-scripts smex
 (spell-checking :variables spell-checking-enable-by-default nil)
 syntax-checking version-control)
```

#### Related
This issue happens after recent commit ec3bbd41769, and it is related to https://bitbucket.org/lyro/evil/issues/301/evil-define-key-for-minor-mode-does-not

TheBB:
I guess we have to trigger `(evil-normalize-keymaps)` in that toggle, then.

TheBB:
#5834

et2010:
Thanks, it works like a charm!

PR #11417, opened 2018-10-11
Add toggle visual-line-navigation-globally

mjkramer:
Given that there's apparently [some interest](https://emacs.stackexchange.com/questions/41798/spacemacs-visual-line-navigation-not-working/44853) in having a global toggle for `visual-line-navigation`, I've thrown one together. Under my testing it seems to work as expected. There are a few questions lingering, though:

- I'm not sure why/whether the `evil-escape` logic is (still?) necessary. `evil-escape-motion-state-shadowed-func` doesn't seem to exist.
- Given that the bindings are being injected directly into `visual-line-mode`, it seems unnecessary to bind them every time the (global or local) mode is toggled, but I could be overlooking some edge case (involving evil-escape?)
- I'm not sure why `evil-normalize-keymaps` is necessary, and I don't think it makes sense to call it for the global case (since `evil-normalize-keymaps` just makes `evil-mode-map-alist` buffer-local), but for the sake of code simplicity I've left it there for both cases. It doesn't seem to hurt.

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/keybindings.el
+++ b/layers/+spacemacs/spacemacs-defaults/keybindings.el
@@ -306,26 +306,40 @@
   :off (toggle-truncate-lines -1)
   :documentation "Truncate long lines (no wrap)."
   :evil-leader "tl")
+(defun spacemacs//init-visual-line-keys ()
+  (evil-define-minor-mode-key 'motion 'visual-line-mode "j" 'evil-next-visual-line)
+  (evil-define-minor-mode-key 'motion 'visual-line-mode "k" 'evil-previous-visual-line)
+  (when (bound-and-true-p evil-escape-mode)
+    (evil-escape-mode -1)
+    (setq evil-escape-motion-state-shadowed-func nil)
+    (evil-define-minor-mode-key 'motion 'visual-line-mode "j" 'evil-next-visual-line)
+    (evil-define-minor-mode-key 'motion 'visual-line-mode "k" 'evil-previous-visual-line)
+    (evil-escape-mode))
+  (evil-normalize-keymaps))
 (spacemacs|add-toggle visual-line-navigation
   :status visual-line-mode
   :on
   (progn
     (visual-line-mode)
-    (evil-define-minor-mode-key 'motion 'visual-line-mode "j" 'evil-next-visual-line)
-    (evil-define-minor-mode-key 'motion 'visual-line-mode "k" 'evil-previous-visual-line)
-    (when (bound-and-true-p evil-escape-mode)
-      (evil-escape-mode -1)
-      (setq evil-escape-motion-state-shadowed-func nil)
-      (evil-define-minor-mode-key 'motion 'visual-line-mode "j" 'evil-next-visual-line)
-      (evil-define-minor-mode-key 'motion 'visual-line-mode "k" 'evil-previous-visual-line)
-      (evil-escape-mode))
-    (evil-normalize-keymaps))
+    (spacemacs//init-visual-line-keys))
   :off
   (progn
     (visual-line-mode -1)
     (evil-normalize-keymaps))
   :documentation "Move point according to visual lines."
   :evil-leader "tL")
+(spacemacs|add-toggle visual-line-navigation-globally
+  :status global-visual-line-mode
+  :on
+  (progn
+    (global-visual-line-mode)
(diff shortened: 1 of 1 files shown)


dongcarl:
ACK

duianto:
#### `evil-escape-motion-state-shadowed-func`

Confirmed the `evil-escape-motion-state-shadowed-func` variable or anything named `shadow` are missing from [`evil-escape` package.](https://github.com/syl20bnr/evil-escape)

The variable was introduced in Spacemacs with this commit (Aug 13, 2015):
Sync visual-line-mode and evil-visual-xxx functions https://github.com/syl20bnr/spacemacs/commit/9a37c63991191ef1957056774f6b087e2910de5a

But `evil-escape` was rewritten in version 3.0 (Sep 2, 2015), which removed any shadow related variables:
Rewrite of evil-escape (v3.0) https://github.com/syl20bnr/evil-escape/commit/1271af99ac537a3e38d7ce5445e8d12e35e302b0

Therefore `(setq evil-escape-motion-state-shadowed-func nil)` can be removed.


#### `evil-escape`

The `evil-escape-mode` calls might only have been used with the shadow variable.

The visual line navigation seems to work the same without them.

So they can probably be removed as well.

#### `evil-normalize-keymaps`

`evil-normalize-keymaps` was added to solve this issue:

The \`navigation' part of \`visual-line-navigation' does not take effect until a state transition #5799

The same or a similar issue still occurs with this PR when multiple windows are open and visual line navigation globally is enabled.

Visual line navigation only works in the current window.

This issue can be reproduced like this:
- open a new buffer: `SPC b N n`
- insert a lorem ipsum paragraph: `SPC i l p`
- open a new buffer below: `SPC b N j`
- insert a lorem ipsum paragraph: `SPC i l p`
- enable visual line navigation globally: `SPC t C-l`
(move up `k` and down `j`, the cursor moves by visual line)
- select the window above: `SPC w k`
- move up `k` and down `j`, the cursor does not move by visual line.
(`k` and `j` are still bound to `evil-previous-line` and `evil-next-line`, instead of to the expected `evil-previous-visual-line` and `evil-next-visual-line`)

Pressing `Esc` in the second window seems to cause a state transition,and now `k` and `j` are bound to the correct visual line commands.

Maybe one solution could be to call `evil-normalize-keymaps` in all open windows.

#### About the key binding `SPC t C-L`

The current Spacemacs key binding `SPC t L` for toggling visual line navigation in the current buffer uses uppercase L.

And this PR also uses the upper case L with Ctrl: `SPC t C-L`.

Currently Spacemacs doesn't have a binding for Ctrl with lower case L: `SPC t C-l`

But Emacs converts the upper case binding to lower case.
source:
this comment: https://stackoverflow.com/a/22155466
from this page: https://stackoverflow.com/questions/22154688/how-to-get-emacs-key-bindings-to-recognize-capital-letters

Therefore both the upper (`SPC t C-L`) and lower (`SPC t C-L`) case bindings work.

### An issue

It's easier to press the lower case binding `SPC t C-l`, therefore people will get used to using the lower case binding for toggling visual line navigation globally.

But in the future if a command is bound to the lower case binding `SPC t C-l`, then this visual line navigation globally key binding `SPC t C-L` will only work with the upper case binding.

#### There seems to be at least two options:

#### Consistency:

Using upper case L in the key binding for toggling visual line navigation both in the current window or globally would be more consistent.

To only use the uppercase binding for toggling it globally, then this PR would have to change the key binding to: `SPC t C-S-l`

#### Convenience (short or long term, depending on if/when the lower case binding gets a command):

If it's fine that the key binding might change in the future.

Then the slightly simpler key binding `SPC t C-l` can be used, and this PR's key binding `SPC t C-L` won't need to be changed.

duianto:
The PRs changes have been cherry picked into the `develop` branch. https://github.com/syl20bnr/spacemacs/commit/e4c8cef41265721af9ef67db00345a76256d1cd2

Changes:
- Extracted the call to `(evil-normalize-keymaps)` from the `init-visual-line-keys` function and added it back into the `visual-line-navigation` (and `-globally`) `:on` sections.
This makes it easier to see that the normalize function is called in both the `:on` and `:off` sections, if it needs to be changed in the future.
- Moved the `init-visual-line-keys` function to: `funcs.el`
- Chose the key binding: `"t C-S-l"` to match the non global key binding that also requires shift to be held down: `"tL"`

PR #11419, opened 2018-10-11
add instructions for setting shell for multi-term

### LINKED ISSUE #5818: Make shell-default-term-shell affect multi-term
Currently `shell-default-term-shell` only affects the default shell in `term` and `ansi-term`, not `multi-term` (which is determined by `shell-file-name`).  I think it would be more consistent to make `multi-term` affected by `shell-default-term-shell` as well.

This is especially apparent if one is using a dumb shell for `shell-file-name` and a smarter one for `*-term`.

d12frosted:
Fixes #5818


d12frosted:

Issue #11388, opened 2018-10-04
default shell cant set hight properties and position.


brorbw:
#### Description :octocat:
Using the system default shell
```
   '(shell :variables
           shell-default-shell 'shell
           shell-default-height 30
           shell-default-position 'top)
```
Ignores height and position. Using any of the other shells works fine, but the colors are all broken. Using the built in is the only shell that have proper colors. Im using zsh as the system default shell.
I can drag the size with the mouse but it cannot be set 

#### Reproduction guide :beetle:
- Start Emacs
- Set the default shell to be the system default
- Open a shell window, `[SPC ']`

*Observed behaviour:* :eyes: :broken_heart:
The shell pops op where ever and at half the size of frame its in. 

*Expected behaviour:* :heart: :smile:
Expected the shell to open in the top and with a height of 30.

#### System Info :computer:
- OS: darwin
- Emacs: 26.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 9865e7731)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
((javascript :variables javascript-disable-tern-port-files nil)
 yaml octave html python swift rust scala latex ivy
 (auto-completion :variables auto-completion-enable-snippets-in-popup t auto-completion-enable-help-tooltip t)
 better-defaults emacs-lisp git markdown
 (c-c++ :variables c-c++-enable-clang-support t)
 (spell-checking :variables enable-flyspell-auto-completion t)
 (shell :variables shell-default-shell 'shell shell-default-height 30 shell-default-position 'top)
 xkcd
 (org :variables org-enable-reveal-js-support t org-projectile-file '"~/org/TODOs.org")
 themes-megapack
 (wakatime :variables wakatime-api-key "<<REMOVED KEY>>" wakatime-cli-path "/usr/local/bin/wakatime"))
```
- System configuration features: JPEG RSVG IMAGEMAGICK NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS MODULES THREADS LCMS2

#### Backtrace :paw_prints:
```
<<BACKTRACE IF RELEVANT>>
```

cipharius:
Try switching to develop branch, the issue should be fixed now.

brorbw:
I am

Issue #7691, opened 2016-11-08
`SPC a s i` messes up layout


kini:
#### Description :octocat:
`SPC a s i` messes up layout

#### Reproduction guide :beetle:
1. Start Emacs
2. Do `SPC w / SPC a s i`
3. Do `ESC SPC a s i`

*Observed behaviour:* :eyes: :broken_heart:

After step 2, the left window is focused and contains the shell
buffer.  The bottom window and right window contain the Spacemacs
splash screen buffer.

After step 3, the frame is split between a large top window and a
small bottom window, both containing the Spacemacs splash screen
buffer.

*Expected behaviour:* :heart: :smile:

After step 2, the bottom window is focused and contains the shell
buffer.  The left and right windows contain the Spacemacs splash
screen buffer.

After step 3, the frame is split between equally sized left and right
windows, both containing the Spacemacs splash screen buffer.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 25.1.1
- Spacemacs: 0.200.5
- Spacemacs branch: develop (rev. 2c4cd4d)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(yaml html python helm better-defaults emacs-lisp git github markdown org
      (shell :variables shell-default-height 30 shell-default-position 'bottom)
      (spell-checking :variables spell-checking-enable-by-default nil)
      syntax-checking version-control rust)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GSETTINGS NOTIFY GNUTLS FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11

TheBB:
This is easily reproducible outside Spacemacs, therefore please take it upstream.

kini:
Thanks.  Reported as kyagi/shell-pop-el#51.

PR #11422, opened 2018-10-11
Fix incorrect shell-pop window splitting for 'shell' option

cipharius:
Currently `shell` command, when switching to non-existing buffer, will
create a new window automatically. This behavior does not play will with
`shell-pop` package, as it already prepares window for shell to use.

This behavior can be prevented by creating a new buffer, before calling
the `shell` command.

Related issues: 
- #11388
- #7691


d12frosted:

PR #11423, opened 2018-10-11
allow dynamic width lock setup for treemacs

d12frosted:
Following #11384

Right now treemacs-lock-width variable affects width lock only during treemacs
initialisation. Any later modifications have no effect on treemacs behaviour.
This commit changes it by moving evaluation of treemacs-lock-width to the
treemacs-mode-hook.

CC @deb0ch

### DIFF
--- a/layers/+filetree/treemacs/funcs.el
+++ b/layers/+filetree/treemacs/funcs.el
@@ -21,8 +21,9 @@
       (treemacs-do-add-project-to-workspace path name)
       (treemacs-select-window))))
 
-(defun spacemacs/treemacs-toggle-locked-width-off ()
-  "Unlock the manual resizing of the treemacs window."
+(defun spacemacs/treemacs-setup-width-lock ()
+  "Setup the width lock of treemacs buffer based on
+`treemacs-lock-width'."
   (interactive)
-  (when treemacs--width-is-locked
+  (when (xor treemacs--width-is-locked treemacs-lock-width)
     (treemacs-toggle-fixed-width)))
--- a/layers/+filetree/treemacs/packages.el
+++ b/layers/+filetree/treemacs/packages.el
@@ -41,9 +41,8 @@
             treemacs-never-persist nil
             treemacs-goto-tag-strategy 'refetch-index
             treemacs-collapse-dirs treemacs-use-collapsed-directories)
-      (unless treemacs-lock-width
-        (add-hook 'treemacs-mode-hook
-                  'spacemacs/treemacs-toggle-locked-width-off))
+      (add-hook 'treemacs-mode-hook
+                #'spacemacs/treemacs-setup-width-lock)
       (spacemacs/set-leader-keys
         "ft"    'treemacs
         "fB"    'treemacs-bookmark


layers/+filetree/treemacs/funcs.el
```diff
@@ -21,8 +21,9 @@
...
       (treemacs-select-window))))
 
-(defun spacemacs/treemacs-toggle-locked-width-off ()
-  "Unlock the manual resizing of the treemacs window."
+(defun spacemacs/treemacs-setup-width-lock ()
+  "Setup the width lock of treemacs buffer based on
+`treemacs-lock-width'."
   (interactive)
-  (when treemacs--width-is-locked
+  (unless (eq treemacs--width-is-locked treemacs-lock-width)
```
deb0ch: Mmmmmmh this will make the docs non-accurate, since a simply non-nil value won't always do it. It should be

```
(unless treemacs-locked-width
  (when treemacs--width-is-locked
    (treemacs-toggle-fixed-width)))
```
It may be longer, but at least it makes clear what the intent is.

deb0ch:
I guess it makes more sense putting everything in the hook and always adding it, but I still can't put my finger on which situation it may help.

deb0ch:
Oh yes, or simply evaluating some lisp somehow to set that variable

d12frosted:
> Oh yes, or simply evaluating some lisp somehow to set that variable 

This is exactly what I did to see if it works 😸 I am not sure how often people would like to change that (especially since you can `tw` in `treemacs` buffer). But making it like this feels more Emacs-ish or so 😸

layers/+filetree/treemacs/funcs.el
d12frosted: Good point, there might be some shit instead of `t`. I'd use it like this:

```
(unless (eq (not treemacs--width-is-locked) (not treemacs-lock-width))
  ..)
```

I don't know how to compare two 'boolean' 'objects' 😭 😸

deb0ch: Just, simply 
```
(unless treemacs-locked-width
  (when treemacs--width-is-locked
    (treemacs-toggle-fixed-width)))
```
`Unless they share the same thruthyness` is clever but I think it also hides the intent behind it. It makes it hard to read because you also have to be clever (or think hard) to read it.

deb0ch: Just, simply 
```
(unless treemacs-locked-width
  (when treemacs--width-is-locked
    (treemacs-toggle-fixed-width)))
```
`Unless they share the same thruthyness` is clever but I think it also hides the intent behind it. It makes it hard to read because you also have to be clever (or think hard) to read it.

d12frosted: It holds true iff set `treemacs-locked-width` to `nil` and treemacs is currently locked. But my point was about the situation when treemacs is not locked and the user wants to lock it. I know that by default treemacs is locked, but I would not place my stake on implicit* default value from third-party 😸 One day they change behaviour and you have to come back and change this code. Being bullet-proof against changes outside of the project is a good thing I think :D 

* implicit here means that we do not use any real variable here, we just /know/ that it holds 😸

deb0ch: I see what you mean, did'nt think of it like that.

`(unless (eq (not treemacs--width-is-locked) (not treemacs-lock-width)))` would be equivalent to `(when (xor treemacs-lock-width treemacs--width-is-locked))` if I'm not mistaken.

d12frosted:

deb0ch:
Yay ! That looks definitely better than my original PR

sdwolfz:
I get the following error related to this PR:

```
Debugger entered--Lisp error: (void-function xor)
  (xor treemacs--width-is-locked treemacs-lock-width)
  (if (xor treemacs--width-is-locked treemacs-lock-width) (progn (treemacs-toggle-fixed-width)))
  spacemacs/treemacs-setup-width-lock()
  run-hooks(change-major-mode-after-body-hook special-mode-hook treemacs-mode-hook)
  apply(run-hooks (change-major-mode-after-body-hook special-mode-hook treemacs-mode-hook))
  run-mode-hooks(treemacs-mode-hook)
  treemacs-mode()
  treemacs--init()
  treemacs()
  funcall-interactively(treemacs)
  call-interactively(treemacs nil nil)
  command-execute(treemacs)
```

d12frosted:
Interesting. Maybe I load `array.el` somewhere early. Ok, I am going to switch to the first variant :) 

@sdwolfz thank you for letting me know about this problem. I've pushed a fix: https://github.com/syl20bnr/spacemacs/commit/4024b60757e105a389def9bb58f028f3d4a44aa0

PR #11424, opened 2018-10-11
use a global key binding for helm-org-rifle

### LINKED ISSUE #11420: Change keybinding for helm-org-rifle
there was addition of helm-org-rifle in ff9fd1bcc3c74c2bdc04e9fe0241f7769aa0166d

Shortcut was set to `SPC m r` which is used during capture to refile captured buffer. I think this is missleading and shouldn't be used in that way. What about `SPC m R`?

d12frosted:
Fixes #11420

### DIFF
--- a/layers/+emacs/org/README.org
+++ b/layers/+emacs/org/README.org
@@ -682,7 +682,7 @@ org-present must be activated explicitly by typing: ~SPC SPC org-present~
 
 | Key Binding | Description                                |
 |-------------+--------------------------------------------|
-| ~SPC m r~   | Search org files for keywords and headings |
+| ~SPC a o r~ | Search org files for keywords and headings |
 
 ** Org-projectile
 
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -76,8 +76,7 @@
 (defun org/init-helm-org-rifle ()
   (use-package helm-org-rifle
     :defer t
-    :init (spacemacs/set-leader-keys-for-major-mode 'org-mode
-            "r" 'helm-org-rifle)))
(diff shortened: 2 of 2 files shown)


kepi:
there is one more mention (in changelog develop) with key binding, should be change to probably

d12frosted:

syl20bnr:
@d12frosted  Is it possible to replace `SPC a o /` with `helm-rifle` as well when `helm` is used ? `SPC a o /` is org occur in agenda files.

If possible I would like to change `SPC a o r` for something agnostic of being `helm-org-rifle` behind the scene. I think we can replace `SPC a o s` with a `search` prefix and put everything behind. I then would see `SPC a o /` to some kind of `helm-org-rifler-projectile`.

kepi:
`SPC a o s` as prefix seems little long to me. Anyway `SPC a o /` for projectile-kind of thing doesn't seam right to me, or is it? I expect project related bidnings in `SPC p` and anything in `SPC a o` as global thing without relation to projects. To be more specific, I need my org behave globaly as I have org files in different places. It would be bad if it would search only in current project.

I didn't try `helm-org-rifle` yet as I'm ivy user so I'm not sure if it is 1:1 (or better) replacement for occur, but if it behaves differently, I don't think we should be changing bindings user are used too.

alphapapa:
> I didn't try helm-org-rifle yet as I'm ivy user so I'm not sure if it is 1:1 (or better) replacement for occur

It's very different.  See the animation in the readme.  :)

If I may, I suggest that you bind at least these two commands:

* `helm-org-rifle`, which searches all open Org buffers
* `helm-org-rifle-agenda-files`, which searches all `org-agenda-files`, even if they aren't open yet

PR #11426, opened 2018-10-12
Fix jedi requirements on python documentation

jaysonsantos:
Hi there, right now on a fresh install anaconda-mode will not work with the give instructions because jedi>=0.12 breaks import compatibility. So I just added the right version that needs to be installed in order to make it work. I will try to send a PR to anaconda-mode to support newer jedi versions.

```python
Traceback (most recent call last):
  File "<string>", line 4, in <module>
  File "/Users/jayson/.emacs.d/.cache/anaconda-mode/0.1.6/anaconda_mode-0.1.6-py3.5.egg/anaconda_mode.py", line 17, in <module>
    from jedi import Script, NotFoundError
```


jaysonsantos:
I've opened it to master, I will change it for develop.

PR #11427, opened 2018-10-12
 Fix jedi requirements on python documentation

jaysonsantos:
Right now on a fresh install anaconda-mode will not work with the give instructions because jedi>=0.12 breaks import compatibility.

So I just added the right version that needs to be installed in order to make it work.
I will try to send a PR to anaconda-mode to support newer jedi versions.

### DIFF
--- a/layers/+lang/python/README.org
+++ b/layers/+lang/python/README.org
@@ -94,7 +94,7 @@ opening a python buffer:
 To fix this, install the =anaconda-mode= [[https://github.com/proofit404/anaconda-mode/wiki][anaconda-deps]] by hand:
 
 #+BEGIN_SRC sh
-  pip install --upgrade "jedi>=0.9.0" "json-rpc>=1.8.1" "service_factory>=0.1.5"
+  pip install --upgrade "jedi>=0.9.0,<0.11.0" "json-rpc>=1.8.1" "service_factory>=0.1.5"
 #+END_SRC
 
 If you encounter problems with Jedi 1.0 consider downgrading to 0.9.0. See [[https://github.com/davidhalter/jedi/issues/873][this


layers/+lang/python/README.org
```diff
@@ -94,7 +94,7 @@ opening a python buffer:
 To fix this, install the =anaconda-mode= [[https://github.com/proofit404/anaconda-mode/wiki][anaconda-deps]] by hand:
 
 #+BEGIN_SRC sh
-  pip install --upgrade "jedi>=0.9.0" "json-rpc>=1.8.1" "service_factory>=0.1.5"
+  pip install --upgrade "jedi>=0.9.0,<0.11.0" "json-rpc>=1.8.1" "service_factory>=0.1.5"
```
d12frosted: I actually see that `anaconda-mode` requires `jedi >= 0.13.0`. [Source](https://github.com/proofit404/anaconda-mode/commit/150390968bb0c5180026932d27255483d690e00f). Am I missing something?

jaysonsantos: Could be because locally my anaconda-mode is on 0.1.6 and there they talked about this problem https://github.com/proofit404/anaconda-mode/tree/v0.1.6.
Checking M-x package-list-packages it shows this version `anaconda-mode      20161121.1137 ` installed. I think I misunderstood and thought that spacemacs would update them every time I ran M-m f e R

d12frosted: No. Please try upgrading your packages and try again 😸 If it works, I would actually update your PR with a new version of `jedi`.

jaysonsantos: I did this now `cd ~/.emacs.d && git pull`, installed jedi with `pip install -U jedi` which installed `jedi-0.13.1`, ran `M-x configuration-layer/update-packages` and it seems to work normally.

jaysonsantos: Thank you anyway!

PR #11431, opened 2018-10-12
Clojure enhancements

kommen:
1. ~~add a keybinding to open the cider repl history buffer in an evil way~~ (see https://github.com/syl20bnr/spacemacs/pull/11431#issuecomment-431571168)
2. evilfy cider-repl-history-mode-map
3. allow to send input to a cider repl in normal mode with RET

### DIFF
--- a/layers/+lang/clojure/packages.el
+++ b/layers/+lang/clojure/packages.el
@@ -170,7 +170,8 @@
 
       ;; cider-repl-mode only
       (spacemacs/set-leader-keys-for-major-mode 'cider-repl-mode
-        "," 'cider-repl-handle-shortcut)
+        "," 'cider-repl-handle-shortcut
+        "P" 'cider-repl-history)
       (spacemacs/set-leader-keys-for-major-mode 'cider-clojure-interaction-mode
         "ep" 'cider-eval-print-last-sexp))
     :config
--- a/layers/+lang/clojure/packages.el
+++ b/layers/+lang/clojure/packages.el
@@ -207,6 +207,8 @@
       (evilified-state-evilify cider-docview-mode cider-docview-mode-map
         (kbd "q") 'cider-popup-buffer-quit)
 
+      (evilified-state-evilify cider-repl-history-mode cider-repl-history-mode-map)
+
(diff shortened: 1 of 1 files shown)


kommen:
@d12frosted I discussed with @jr0cket on the Clojurians Slack Spacemacs channel about putting `cider-repl-history` onto `SPC s h` instead of `, P`. I used `, P` since cider has it on [`C-c M-p`](https://github.com/clojure-emacs/cider/blob/6d6aa42ecf5bad9110cc0a781e7786b87812b624/cider-repl.el#L1621), but I can see making since having it under `SPC s` since there all the repl commands are. Any opinions on that?

jr0cket:
We definately need keybindings for Vim Normal mode as it feels very inconsistent changing into vim insert mode just to call the repl history with `,`.  Its also easy to get confused between `,` in insert and normal mode.  Having to calle `SPC SPC cider-repl-history` is not great either.

I suggested `, s h` as it fits in with the mnemonic menu approach:  major-mode > cider > history.  

Although P is used in the cider package itself, several of the keybindings have changed to conform to the mnemonic approch.  If there were other history related keybindings already using P then it would make sense, but I havent seen this convention in Spacemacs.

kommen:
I've removed the `, P` key for now seems it seems this needs some more discussion. This still leaves the current  way of getting to the history via the cider repl shortcuts, which I just discovered now myself:
in a repl buffer in vim insert mode: `,` then select history
in a repl buffer in vim normal mode: `, ,` then select history

So I'm not sure if we even should add custom keybindings.

@jr0cket I've also addressed the issue you ran into that RET in vim normal mode didn't evaluate: I had I typo on the patch. 

@d12frosted Since this PR doesn't introduce any new keybindings now, I think no changes to the README are needed. The RET keybinding just makes it behave as it should have been and RET is not documented for vim input mode in the repl as well.

d12frosted:
@kommen makes sense. I've tried to search for any other layer that defines key bindings for navigating the REPL history and didn't find one. But since REPL usually has it's own mode, I don't see a reason to add key bindings under `, s`. 

Anyway, I've cherry-picked this commit into `develop`, you can safely delete your branch. Thank you 🎉

PR #11433, opened 2018-10-13
Add 'eshell to default-pop-shell

simendsjo:


simendsjo:
Looks like it's been fixed already

PR #11437, opened 2018-10-13
[TravisCI] Add documentation autofixer

JAremko:
I wonder what are the chances that it works from the first try.


JAremko:
:crossed_fingers:

PR #11435, opened 2018-10-13
Add keybings to rename the file and organize imports in typescript

Kjir:
Add two key bindings in typescript major mode:

- Rename the current file (`tide-rename-file`)
- Organize the imports (`tide-organize-imports`)

These operations are generic and useful enough to deserve to be set by default.

PR #11441, opened 2018-10-13
Fix the rename file and organize import key binding for typescript

Kjir:
Fix the keybindings introduced in #11435 which were based on the configuration present in `master` rather than the one in `develop`.


Kjir:
I missed it too... Oh well!

PR #11461, opened 2018-10-16
add actions to open funcs.el and layers.el from layers help

d12frosted:
I constantly miss these actions. Time to fix my miss.

P. S. I could not find the place where these are documented. Maybe they are not documented at all 🙀


d12frosted:
Rule of thumb - don't change documentation, so all checks pass.

d12frosted:

PR #11467, opened 2018-10-17
Descriptive rename function file variables

duianto:
problem:
the variable names: `dir`, `name`, and `filename` leads to confusion if they are
the old or new names

solution:
add the prefix `old-` to match `new-`,
use (`old-` `new-`) + `short-name` for just the `file.ext` (the `short-` idea
came from the `spacemacs/rename-file` function that's defined above this one),
and `old-full-path` instead of `old-filename` makes it clear that it's not just
the files name and extension.

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -311,32 +311,35 @@ be saved to a file, or just renamed.
 If called without a prefix argument, the prompt is
 initialized with the current filename."
   (interactive "P")
-  (let* ((name (buffer-name))
-         (filename (buffer-file-name)))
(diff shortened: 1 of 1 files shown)


duianto:
Maybe this PR should be on hold, because I just submitted two PR's that use the current variable names:
Update file renaming, handle same new and old name #11473
Update renaming, move and/or rename messages #11475

I didn't want to assume that this would get accepted first (if at all).

If they are accepted first then I'll update this PR's variables. Or the other way around, if this is accepted first, then I can update the variables in the other two PR's.

Would there have been a better way to handle situations like this, where one PR depends on another?

smile13241324:
Merged. I have adjusted the variable names to be inline with the two related PRs #11473 #11475.

PR #11468, opened 2018-10-17
Fix missing newline warning

duianto:
problem:
`'face 'minibuffer-prompt` and `new-name` are shown with an orange color,
and `describe-char` shows that they have the face: `font-lock-warning-face`

solution:
adding a newline before them removes the orange warning color


d12frosted:

Issue #9266, opened 2017-07-23
Giving directory as a rename target uses directory as buffer-name


shanavas786:
#### Description :octocat:
Giving directory as a rename target uses directory as buffer-name 

#### Reproduction guide :beetle:
- Start Emacs
- Visit a file say `/tmp/a.txt`
- `SPC f R` and give directory name as target say `/tmp2`

*Observed behaviour:* :eyes: :broken_heart:
- file `a.txt` is moved to `/tmp2/a.txt`
- buffer name is '/tmp2/' and it visits `/tmp2/a.txt`
- similar behaviour is observed in `spacemacs/rename-file` where buffer visits target directory

*Expected behaviour:* :heart: :smile:
- file `a.txt` is moved to `/tmp2/a.txt`
- buffer name is 'a.txt' and it visits `/tmp2/a.txt`

#### System Info :computer:
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.200.9
- Spacemacs branch: master (rev. 8e1af145)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: helm
- Layers:
```elisp
(ansible graphviz auto-completion c-c   elfeed csv emacs-lisp erc
         git gtags haskell helm html java javascript latex markdown org php python
         (ruby :variables ruby-version-manager 'rvm)
         ruby-on-rails rust shell shell-scripts spell-checking sql syntax-checking typescript version-control xkcd yaml)
```

#### Backtrace :paw_prints:
```

```

d12frosted:
I observe slightly different behaviour.

File `a.txt` visits `tmp/a.txt` instead of `tmp2/a.txt`. Also I see very funny message: `File ’’ successfully renamed to ’tmp2’`.

To be honest I am not sure that `rename-file` should accept directories. But there should separate function for directories that

- renames directory
- fixes all buffers visiting files in that directory
- invalidates projectile cache
- updates recentf list

What do you think?

shanavas786:
On July 27, 2017 10:47:11 PM GMT+03:00, Boris <notifications@github.com> wrote:
>I observe slightly different behaviour.
>
>File `a.txt` visits `tmp/a.txt` instead of `tmp2/a.txt`. Also I see
>very funny message: `File ’’ successfully renamed to ’tmp2’`.
>

Let me check it once more.

>To be honest I am not sure that `rename-file` should accept
>directories. 

[...]

PR #11470, opened 2018-10-18
support directories in spacemacs/rename-file

d12frosted:
1. Fix `short-name` for directories. Previously it was empty string.
2. Remove redundant check for existing buffer:
   1. `(get-buffer new-name)` always return `nil`, because `new-name` is a path
      to file, not a file name.
   2. `(find-file new-name)` handles this situation.
   3. `dired-rename-file` handles this situation as well.
3. Use `dired-rename-file` to do the dirty stuff.
   1. Renames files and directories.
   2. All associated buffers are updated. If you rename a directory, all buffers
      that were visiting files from that directory now visit files from new
      directory.
4. Properly handle recentf list update for directories - all files from the
   renamed directory are removed from recentf list and the same files but from
   new directory are added back.

Fixes https://github.com/syl20bnr/spacemacs/issues/9266

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -275,32 +275,47 @@ Dedicated (locked) windows are left untouched."
 
 When NEW-FILENAME is not specified, asks user for a new name.
 
-Also renames associated buffer (if any exists), invalidates
-projectile cache when it's possible and update recentf list."
+Also renames associated buffers (if any exists), invalidates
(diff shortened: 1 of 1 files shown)


d12frosted:
Just a side note - at first I thought about separating functions for renaming files and directories, but then I found that `dired-rename-file` does so much love that it's easy to implement one function for two purposes.

Also, note that `dired` comes from the same layer where `spacemacs/rename-file` is defined. So we should be good.

d12frosted:

Resulting commit is 69a1e056f.

PR #11471, opened 2018-10-18
Update delete file messages

duianto:
problems:
the delete file prompt doesn't show which file it's about to delete,
and when the deletion request is canceled, then the question remains in the
minibuffer (the text disappears when an action updates the minibuffer, like
moving the cursor, but it might be confusing to still be seeing the question
after it's been answered)

solution:
show the file name in the delete file prompt, and show a canceled message


sdwolfz:
Thank you :heart:!

PR #11473, opened 2018-10-18
Update file renaming, handle same new and old name

duianto:
problem:
renaming a file to it's current name, results in the message:
"File 'current-name' successfully renamed to 'current-name'"

solution:
when the new and old names are the same, show a rename failed message, and call
the rename function again

---
### Issue:
This method of showing the file renaming failed message with the `(sit-for 2)` command isn't optimal, because the helm rename buffer closes, then the failed message is shown in the minibuffer, and finally the helm rename buffer opens again.

Without the `sit-for` command, the helm rename buffer stays open after pressing `RET`, but the failed rename message only gets shown in the minibuffer for a split second.

Is it possible to show the failed rename message in the helm rename buffer?

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -342,6 +342,10 @@ initialized with the current directory instead of filename."
                (new-name (read-file-name "New name: " (if arg dir filename))))
           (cond ((get-buffer new-name)
                  (error "A buffer named '%s' already exists!" new-name))
+                ((string-equal new-name filename)
+                 (spacemacs/show-hide-helm-or-ivy-prompt-msg
+                  "Rename failed! Same new and old name" 1.5)
+                 (spacemacs/rename-current-buffer-file))
                 (t
                  (let ((dir (file-name-directory new-name)))
                    (when (and (not (file-exists-p dir))
@@ -388,6 +392,36 @@ initialized with the current directory instead of filename."
                             name new-name)))
                 ;; ?\a = C-g, ?\e = Esc and C-[
                 ((memq key '(?\a ?\e)) (keyboard-quit))))))))
+
+(defun spacemacs/show-hide-helm-or-ivy-prompt-msg (msg sec)
+  "Show a MSG at the helm or ivy prompt for SEC.
+With Helm, remember the path, then restore it after SEC.
+With Ivy, the path isn't editable, just remove the MSG after SEC."
+  (run-at-time
+   0 nil
+   #'(lambda (msg sec)
+       (let* ((prev-prompt-contents
+               (buffer-substring (line-beginning-position)
+                                 (line-end-position)))
+              (prev-prompt-contents-p
+               (not (string= prev-prompt-contents "")))
+              (helmp (fboundp 'helm-mode)))
+         (when prev-prompt-contents-p
+           (delete-region (line-beginning-position)
+                          (line-end-position)))
+         (insert (propertize msg 'face 'warning))
+         ;; stop checking for candidates
+         ;; and update the helm prompt
+         (when helmp (helm-suspend-update t))
+         (sit-for sec)
+         (delete-region (line-beginning-position)
+                        (line-end-position))
+         (when prev-prompt-contents-p
+           (insert prev-prompt-contents)
+           ;; start checking for candidates
+           ;; and update the helm prompt
+           (when helmp (helm-suspend-update nil)))))
+   msg sec))
 
 (defun spacemacs/delete-file (filename &optional ask-user)
   "Remove specified file or directory.
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -951,6 +951,8 @@ Other:
   - Improved Window Manipulation Transient state (thanks to yuhan0)
   - Exclude which-key from layer sync powerline restore (thanks to Doug Wilson)
   - Fix doom-modeline in the messages buffer (thanks to duianto)
+  - Fixed =spacemacs/rename-current-buffer-file= handle same new and old name
+    (thanks to duianto)
 *** Layer changes and fixes
 **** Ansible
 - Improvements:


layers/+spacemacs/spacemacs-defaults/funcs.el
```diff
@@ -319,6 +319,14 @@ initialized with the current filename."
                (new-name (read-file-name "New name: " (if arg dir filename))))
           (cond ((get-buffer new-name)
                  (error "A buffer named '%s' already exists!" new-name))
+                ((string-equal new-name filename)
+                 (message
+                  (propertize
+                   (format "File renaming failed! Same new and old name: '%s'"
+                           name)
+                   'face 'warning))
+                 (sit-for 2)
```
d12frosted: Doesn't look optimal to me as well. But it works.

I see the described problem with `ivy` as well.

layers/+spacemacs/spacemacs-defaults/funcs.el
```diff
@@ -319,6 +319,14 @@ initialized with the current filename."
...
+                 (spacemacs/rename-current-buffer-file))
```
d12frosted: Do we really need to call this function again? And if yes, maybe we should pass `arg` to it?

duianto: The reason for calling it again was to stay at the rename prompt, and let us type a new name directly, without having to open the rename prompt manually again.

If by `arg` you mean the current optional argument that's named `arg`, then that only seems to determine if just the `dir` (path without the file and extension name) or the `filename` (full path) should be shown after "New name: " at the rename prompt.
```elisp
               (new-name (read-file-name "New name: " (if arg dir filename))))
```

Or if you meant that we could add a second optional argument, then that seems like a possible solution.

It makes it possible call the rename function with `nil` for `arg` and for example:
```text
"Rename failed! Same new and old name; "
```
(maybe there's a shorter name that says the same thing?) as the second argument.

That results in the rename prompt changing from:
```text
New name: c:/Users/username/temp/file.txt"
```
to
```text
Rename failed! Same new and old name; New name: c:/Users/username/temp/file.txt"
```
when `RET` is pressed.

It would have been great to be able to show the rename failed text with one of the faces: warning (orange) or error (red), but it seems like the `read-file-name` function doesn't show the color of a propertized string.

One issue is if the Emacs frame or helm buffer has a narrow width, then the start of the rename prompts text scrolls off to the left.

---

So there's at least three outcomes when pressing `RET` after trying to rename to the current name:
- Close the rename helm prompt, show the rename failed message in the minibuffer for 2 seconds then open the rename prompt again.
- Close the rename prompt, show the rename failed message in the minibuffer.
- (With a second optional argument), keep the rename prompt open but change its text.

duianto:
>Do we really need to call this function again? And if yes, maybe we should pass arg to it?

I see now what you meant by passing `arg` to the rename function, the short filename (file.ext) gets removed.

There probably are pros and cons for both options. But it's probably easier to modify (ex: add a version number to) an existing file name than having to retype/complete it first and then modify it.

---

The reason why the helm buffer closes while the rename failed message is shown, is because the `sit-for` command performs a redisplay, this shows that the helm buffer has closed.
source: http://www.gnu.org/software/emacs/manual/html_node/elisp/Waiting.html

But there's a `sleep-for` command that doesn't update the display. It makes the helm buffer stay visibly open while the fail message is shown. And when the rename command is called, then it seems like the helm buffer never closed.

### Solution 1

```elisp
                ((string-equal new-name filename)
                 (message (propertize "Rename failed! Same new and old name"
                                      'face 'warning))
                 (sleep-for 2)
                 (spacemacs/rename-current-buffer-file))
```

helm-minibuffer-message

ivy-minibuffer-message


#### Issue: Fail message visibility in Helm

The helm prompt (where one looks while typing) is about a third of the way up from the bottom of the Emacs frame, but the failed message is shown in the minibuffer at the bottom.

At least on larger screens with Emacs maximized. The orange (warning face) message might be overlooked, since it's below the line with the helm buffers title that has a bright yellow background color.

### Solution 2

Another solution is to show the failed message at the helm/ivy prompt (and then restore the helm prompts previous directory/file path, ivys path isn't directly editable):

```elisp
                ((string-equal new-name filename)
                 (run-at-time
                  0 nil
                  #'(lambda ()
                      (let* ((prev-prompt-contents
                              (buffer-substring
                               (line-beginning-position) (line-end-position)))
                             (prev-prompt-contents-p
                              (not (string= prev-prompt-contents ""))))
                        (when prev-prompt-contents-p
                          (delete-region
                           (line-beginning-position) (line-end-position)))
                        (insert
                         (propertize
                          "Rename failed! Same new and old name"
                          'face 'warning))
                        ;; stop checking for candidates
                        ;; and update the helm prompt
                        (when helm-mode (helm-suspend-update t))
                        (sit-for 2)
                        (delete-region
                         (line-beginning-position) (line-end-position))
                        (when prev-prompt-contents-p
                          (insert prev-prompt-contents)
                          ;; start checking for candidates
                          ;; and update the helm prompt
                          (when helm-mode (helm-suspend-update nil))))))
                 (spacemacs/rename-current-buffer-file))
```

helm-prompt-message

ivy-prompt-message


### Issue: `~/` vs full path

While testing the two solutions in Ubuntu 18.04, sometimes the string comparison didn't work:
```elisp
(string-equal new-name filename)
```

because:
```elisp
         (filename (buffer-file-name)))
```
returned: `/home/username/temp/a.txt`

while:
```elisp
               (new-name (read-file-name "New name: " (if arg dir filename))))
```
returned: `~/temp/a.txt`

This seems to be caused by helm not being loaded automatically for some commands right after Emacs has been started.

There's an open issue about it:
Helm doesn't load automatically for some commands #11479

The current workaround is to open another helm buffer, ex: Helm M-x (`SPC SPC`). Then the `read-file-name` command in the rename function returns the full path until Emacs is restarted.

If/when the helm loading issue is fixed, then maybe the path issue won't come up?

Or would it be better to expand the `new-name` in the string comparison:
```elisp
((string-equal (expand-file-name new-name) filename)
```

Or directly from the `read-file-name` command:
```elisp
               (new-name
                (expand-file-name
                 (read-file-name "New name: " (if arg dir filename)))))
```


The issue doesn't occur with Ivy, because
```elisp
(read-file-name "New name: ")
```
shows the prompt as `New name: ~/temp/` but when confirming with `RET`, then it returns the full path `/home/username/temp/`.

### Windows specific issue

The issue this PR is fixing seems to be Windows specific, because in Ubuntu 18.04:

```elisp
(rename-file "/home/username/temp/a.txt" "/home/username/temp/a.txt" 1)
```

shows a prompt:
>File /home/username/temp/a.txt already exists; rename to it anyway? (y or n)

But in windows 10:

```elisp
(rename-file "c:/temp/a.txt" "c:/temp/a.txt" 1)
```

returns: `nil`

This means that in Windows the rest of the rename commands continue to run and the last command shows the message:
>File ’a.txt’ successfully renamed to ’a.txt’

#### System Info :computer:
- OS: windows-nt
- Emacs: 26.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 69a89cc41)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
(autohotkey emacs-lisp git html ivy markdown multiple-cursors org treemacs version-control)
```


#### System Info :computer:
- OS: gnu/linux
- Emacs: 26.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 69a89cc41)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(helm emacs-lisp git multiple-cursors neotree version-control)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GSETTINGS NOTIFY LIBSELINUX GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 MODULES THREADS XWIDGETS LIBSYSTEMD LCMS2

duianto:
I chose solution 2: https://github.com/syl20bnr/spacemacs/pull/11473#issuecomment-443422070





This is probably ready. It works in both `helm` and `ivy` in both windows and ubuntu.

But if someone can test it and point out anything that might need to be fixed/improved, then it can be cherry-picked.

smile13241324:
Merged

PR #11475, opened 2018-10-18
Update renaming, move and/or rename messages

duianto:
problem:
the rename command shows the same message:
"File 'old-file-name' successfully renamed to 'new-file-name'"
(without the directory path) for all three file operations:
move, rename and both move & rename

solution:
show messages that fit the file operation,
and align the from and to paths, to make it easier to see how the paths changed.

---
#### Question
Does the `cond` need a default `t` condition?
I don't know what to put there, because the only case I can think of right now that ends up there, is if one tries to rename to the current file name again, but I have a separate PR that handles that issue:
Update file renaming, handle same new and old name #11473 

If it's accepted, then that issue shouldn't reach this `cond`.

Without that same name PR applied then nothing seems to happen when one tries to rename to the same name with this PR applied.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -953,6 +953,8 @@ Other:
   - Fix doom-modeline in the messages buffer (thanks to duianto)
   - Fix void function error when recentf-save-list is undefined
     (thanks to Compro-Prasad)
+  - Fixed =spacemacs/rename-current-buffer-file=, separate messages for
+    move & rename, just move and just rename (thanks to duianto)
 *** Layer changes and fixes
 **** Ansible
 - Improvements:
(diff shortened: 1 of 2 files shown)


duianto:
The result of applying this PR:
#### Start file: `c:/a/1`

1. renamed file `1` to file `2`
2. moved file from `c:/a/2` to `c:/b/2`
3. moved and renamed `c:/b/2` to `c:/a/1`
4. rename from `1` to the same name as the old name `1`

### Before:
1. File ’1’ successfully renamed to ’2’
2. File ’2’ successfully renamed to ’2’
3. File ’2’ successfully renamed to ’1’
4. File ’1’ successfully renamed to ’1’

### After:
1.
```
File Renamed
From: 1
To:   2
```
2.
```
File Moved
From: c:/a/2
To:   c:/b/2
```
3.
```
File Moved & Renamed
From: c:/b/2
To:   c:/a/1
```
4. no message

Nr 4 is fixed by this PR:
Update file renaming, handle same new and old name #11473

smile13241324:
Merged.

PR #11480, opened 2018-10-18
Keeps previous position on toggle-maximize-buffer

agzam:
Spacemacs doesn't keep the current position on toggle maximize. This PR suggests a dumb and simple fix.

to repro the issue:
- create multiple splits, 
- maximize the window, 
- move cursor around
- toggle maximize-buffer

observed behavior: cursor is not where it was


sdwolfz:
Thank you :heart:!

PR #11481, opened 2018-10-19
Add clipboard layer to integrate with clipboard managers.

xuhdev:
Also, I'm wondering whether we should move contents from `xclipboard` layer to this layer, so we can have a "clipboard enhancement" layer.

### DIFF
--- a/layers/+tools/clipboard/README.org
+++ b/layers/+tools/clipboard/README.org
@@ -1,0 +1,20 @@
+#+TITLE: clipboard layer
+
+* Table of Contents                     :TOC_4_gh:noexport:
+- [[#description][Description]]
+  - [[#features][Features:]]
+- [[#key-bindings][Key Bindings]]
+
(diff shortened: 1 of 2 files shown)


sdwolfz:
I'm not sure this layer should exist by itself given that the configuration is minimal and only works with ivy. But it might make sense to live with the rest of the clipboard code although I would prefer it be behind a feature flag and disabled by default.

xuhdev:
@sdwolfz OK. Are you suggesting merging this layer and `xclipboard`, and name the merged layer as `clipboard`?

xuhdev:
And why disabled by default?

xuhdev:
Close in favor of #12510

sdwolfz:
Hello @xuhdev , sorry for not replying for so long, I lost track of this PR. I'm glad you and @duianto managed to sort this thing out ❤️.

PR #11483, opened 2018-10-19
`anaconda-mode-go-back` is replaced with `xref-pop-marker-stack`

yssource:
`anaconda-mode-go-back` is now not supported in anaconda-mode. 
And it is replaced with `xref-pop-marker-stack`


bmag:

PR #11484, opened 2018-10-19
Fixed haskell documentation about REPL

serch037:
The haskell layer documentation indicates the following commands
>
SPC m s s | show the REPL without switching to it
-- | --

SPC m s S | show and switch to the REPL
-- | --

However, they respectively call `haskell-interactive-switch` and `haskell-interactive-bring`, which is the opposite of what the documentation states. 

This was due to a change in `layers/+lang/haskell/funcs.el` and ` layers/+lang/haskell/packages.el` in commit 4c4a1a4e13bd191b04a0f34c64edb3208bf564b7  which did not update the documentation.


d12frosted:

Resulting commit is 2dc7102c9.

PR #11490, opened 2018-10-20
python3.7 comes with a builtin breakpoint() function (PEP 553)

verhovsky:
https://www.python.org/dev/peps/pep-0553/

If you're using Python 3.7 or above, calling `breakpoint()` opens a pdb shell. 

The correct way to do this would be to end with `(t "breakpoint()")))` as the default case and to to check if your python version is <3.7 before then

### DIFF
--- a/layers/+lang/python/funcs.el
+++ b/layers/+lang/python/funcs.el
@@ -109,6 +109,7 @@ when this mode is enabled since the minibuffer is cleared all the time."
 (defun spacemacs/python-annotate-pdb ()
   "Highlight break point lines."
   (interactive)
+  (highlight-lines-matching-regexp "breakpoint()")
   (highlight-lines-matching-regexp "import \\(pdb\\|ipdb\\|pudb\\|wdb\\)")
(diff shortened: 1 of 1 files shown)


layers/+lang/python/funcs.el
```diff
@@ -168,6 +169,8 @@ as the pyenv version then also return nil. This works around https://github.com/
                      ((spacemacs/pyenv-executable-find "pudb") "import pudb; pudb.set_trace()")
                      ((spacemacs/pyenv-executable-find "ipdb3") "import ipdb; ipdb.set_trace()")
                      ((spacemacs/pyenv-executable-find "pudb3") "import pudb; pudb.set_trace()")
+                     ((spacemacs/pyenv-executable-find "python3.7") "breakpoint()")
```
d12frosted: I agree that ideally, we should just use it as default for python of version ≥ 3.7. But let's see how it works out and collect some feedback?

d12frosted:

d12frosted:
BTW, congratulations on your first contribution 🎉

PR #11505, opened 2018-10-25
Set magit buffers as useless

deb0ch:
Magit creates a buffer for almost every possible action that you can do with it. Since recently, these buffers get added to the current perspective, which I find a good and sensible thing. But then, browsing your perspective with `<SPC> b n` or `<SPC> b p` gets tedious as you have to pass all the magit buffers.

I find that setting Magit's buffers as useless makes sense, since these buffers are accessed through Magit's commands (and they get refreshed that way).

### DIFF
--- a/layers/+source-control/git/packages.el
+++ b/layers/+source-control/git/packages.el
@@ -137,6 +137,8 @@
     :defer (spacemacs/defer)
     :init
     (progn
+      (push "magit: .*" spacemacs-useless-buffers-regexp)
+      (push "magit-.*: .*"  spacemacs-useless-buffers-regexp)
       (spacemacs|require 'magit)
       (setq magit-completing-read-function
             (if (configuration-layer/layer-used-p 'ivy)


layers/+source-control/git/packages.el
```diff
@@ -137,6 +137,7 @@
     :defer (spacemacs/defer)
     :init
     (progn
+      (push "magit: .*" spacemacs-useless-buffers-regexp)
```
sdwolfz: 💯 👌

layers/+source-control/git/packages.el
```diff
@@ -137,6 +137,8 @@
...
+      (push "magit-.*"  spacemacs-useless-buffers-regexp)
```
sdwolfz: This ends up filtering valid buffer names like `magit-tutorial.org`. Please add the entire `magit-.*?: .*` regex.

I wish magit would add `*` at the beginning of their buffers so that they would be easyer to filter.

deb0ch: Good call, updated 👍:sparkles:

sdwolfz:
Thank you :heart:!

deb0ch:
That was unexpectedly quick, thank you !

sdwolfz:
A simple fix for a problem that I've personally experienced and could easiliy test out is verry easy to merge. 

I'm slower at merging stuff that have a lot of background context or deal with technologies that I am not confirtable with.

posborne:
At least for me, I've found this change to be fairly disruptive to normal usage as the disappearing status buffer, in particular, when switching away from it can be extremely disruptive and something that has quickly become an annoyance requiring a patch for myself and a colleague that recently moved to develop.

For instance, consider this very common usage pattern:

* Open up magit-status to prepare a set of commits
* Stage some files, create a commit
* Write commit message

Previously, context would be returned to the magit status buffer so I can perform other git operations (maybe stage some more files for another commit, push the changes to a remote, etc.) but due to the buffer being marked as useless (and maybe some other changes) but the buffer is now gone and I end up in some other buffer and I need to pull up magit-status again.

lebensterben:
@posborne 
You can make a PR that adds a new layer config variable to control this behavior.

deb0ch:
Hmmm this was a long time ago ! :open_mouth: 

I understand about the status buffer it could be disruptive if you rely on such a workflow, however strangely I feel like I have similar workflow but never noticed the annoyance for some reason. However I have always noticed the pain of looking for your couple of buffers drowned among the ton of buffers that Magit creates and uses.

Would it make sense to make an exception for the `status` buffer ? :thinking: 

On the one hand an arbitrary exception adds some "black magic" logic that you wouldn't want surprising you, but on the other hand it is true that the `status` buffer has a special role as it is used as Magit's entrypoint and homepage and is usually the first one that you want to see, so why not keep it handy :thinking:

deb0ch:
After testing I don't think that this filtering is the source of your problem, my `spacemacs-useless-buffers-regexp` is `("magit-.*: .*" "magit: .*")` and after editing a commit my Spacemacs indeed brings me back to the `status` buffer.

But this idea of the `status` buffer being an exception still remains valid from my point of view :thinking:

PR #11506, opened 2018-10-25
alternate-buffer without purpose

deb0ch:
`<SPC> tab` is supposed to alternate to the previous buffer that was displayed in the current window, although that won't happen if you are using `purpose` and the previous buffer had a different purpose: it will get sent to another window.

`<SPC> tab` should, as its docs states, be very simple:
```
Switch back and forth between current and last buffer in the
current window.
```
Making it ignore `purpose` makes it match its documentation again, and gain back its very simple behavior 😺

### DIFF
--- a/core/core-funcs.el
+++ b/core/core-funcs.el
@@ -306,7 +306,8 @@ current window."
     (switch-to-buffer
      (cl-find-if (lambda (buffer)
                    (not (eq buffer current-buffer)))
-                 (mapcar #'car (window-prev-buffers window))))))
+                 (mapcar #'car (window-prev-buffers window)))
+     nil t)))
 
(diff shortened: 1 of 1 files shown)


bmag:
1. Did you try passing the optional `force` argument to `switch-to-buffer` instead of using `without-purpose`? It's a more robust and simple fix. 

2. You want to use `flet` or `letf`. `let` doesn't do what you want here.

deb0ch:
1. No I didn't, didn't even know about its existence 😅 
2. didn't know that either. But it works ?? (in both cases)

deb0ch:
Updated, the `force` argument is such a better solution ! 😻❤️👍✨

bmag:

PR #11513, opened 2018-10-26
Added send line to the sql layer

benbot:
added `SPC m s l` and `SPC m s L` to sql mode for sending specific lines to the SQLi buffer


sdwolfz:
Also it would be nice if you squashed your commits into 1 since the change is relatively small.

layers/+lang/sql/packages.el
```diff
@@ -114,6 +121,8 @@
         ;; command around point, which is what you probably want.
         "sf" 'sql-send-paragraph
         "sF" 'spacemacs/sql-send-paragraph-and-focus
+        "sl" 'sql-send-line-and-next
```
d12frosted: Since the user can configure `sql-pop-to-buffer-after-send-region`, shouldn't we make a wrapper for `sql-send-line-and-next` that doesn't focus the buffer no matter the value of `sql-pop-to-buffer-after-send-region`?

WDYT?

benbot: I don't think I understand what you're asking. the `SPC m s l` won't focus the SQLi buffer, it'll just move the cursor to the next line.

d12frosted: If user sets `sql-pop-to-buffer-after-send-region` to `t` in his/her configurations, then `SPC m s l` will focus the buffer. So behaviour will not be different from `SPC m s L`.

sdwolfz:
@d12frosted feel free to merge this once you confirm all your comments were addressed.

duianto:
- Update changelog.develop, list the new key bindings
https://github.com/syl20bnr/spacemacs/blob/develop/CHANGELOG.develop#sql
- Resolve the conflicts

duianto:
The Box at the bottom of this PR, below "All checks have passed"
says:
>This branch cannot be rebased due to conflicts
Rebasing the commits of this branch on top of the base branch cannot be performed automatically due to conflicts encountered while reapplying the individual commits from the head branch.

It's probably because a recent commit modified the sql layer.

And can you squash the commit: Merge branch 'develop' into develop

duianto:
Ok, it seems like it just appears for the collaborators, and it means that we can't use the 
`Rebase and merge` button in the github ui in this state.

Try rebasing, do you get any conflicts?

duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/d1934f8e30b730340a69df76e7433989c2f6df95

PR #11520, opened 2018-10-29
fix bm as bm-load-and-restore removed in upstream

braineo:
Use official recommended way to init bm package.


d12frosted:

Resulting commit is ff86c724b.

d12frosted:
And by the way, congratulations on your first contribution! 🎉

PR #11530, opened 2018-10-30
Add `z` binding to transient states to recenter buffer

yuhan0:
By analogy with the "zz", "zb", "zt" vim / evil bindings, this adds an
additional key binding "z" for `recenter-top-bottom` to the following
transient states which navigate around the buffer in large jumps:
- auto-symbol-highlight
- error
- buffer
- vcs
- org-babel

This allows for repositioning of the buffer for visibility without having to
exit the transient state.

Some updates are also made to the documentation of other transient states (98bd369)

### DIFF
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -133,7 +133,6 @@
       - [[#buffers-manipulation-key-bindings][Buffers manipulation key bindings]]
       - [[#create-a-new-empty-buffer][Create a new empty buffer]]
       - [[#buffers-manipulation-transient-state][Buffers manipulation transient state]]
-      - [[#special-buffers-usefuluseless-buffers][Special Buffers (useful/useless buffers)]]
       - [[#files-manipulations-key-bindings][Files manipulations key bindings]]
       - [[#frame-manipulation-key-bindings][Frame manipulation key bindings]]
       - [[#emacs-and-spacemacs-files][Emacs and Spacemacs files]]
@@ -189,6 +188,8 @@
   - [[#managing-projects][Managing projects]]
   - [[#registers][Registers]]
   - [[#errors-handling][Errors handling]]
+    - [[#error-transient-state][Error transient state]]
+    - [[#custom-fringe-bitmaps][Custom fringe bitmaps]]
   - [[#compiling][Compiling]]
 - [[#editorconfig][EditorConfig]]
 - [[#emacs-server][Emacs Server]]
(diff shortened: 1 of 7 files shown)


doc/DOCUMENTATION.org
```diff
@@ -1676,34 +1678,32 @@ selected layout.
...
+| ~N~ or ~p~ or ~C-h~ | previous layout in list                                    |
+| ~o~                 | open a custom layout                                       |
+| ~r~                 | remove current buffer from layout                          |
+| ~R~                 | rename current layout                                      |
+| ~s~                 | save layouts                                               |
+| ~t~                 | display a buffer without adding it to the current layout   |
+| ~w~                 | workspaces transient state (needs eyebrowse layer enabled) |
+| ~x~                 | kill current layout with its buffers                       |
+| ~X~                 | kill other layouts with their buffers                      |
+| Any other key       | leave the layout transient state                           |
```
sdwolfz: Instead of adding this line to all the transient state keybinding description I would prefer if we removed it from the existing ones. This functionality should be implied by the fact that it's a transient state.

layers/+source-control/version-control/README.org
```diff
@@ -180,26 +180,27 @@ A transient buffer is also defined, start it with ~SPC m .~ or ~, .~
...
+| ~U~         | Unstage file                 |
+| ~d~         | Repo diff popup              |
+| ~D~         | Show diffs of unstaged hunks |
+| ~c~         | Commit with popup            |
+| ~C~         | Commit                       |
+| ~P~         | Push repo with popup         |
+| ~f~         | Fetch for repo with popup    |
+| ~F~         | Pull repo with popup         |
+| ~l~         | Show repo log                |
+| ~z~         | Recenter buffer in window    |
```
sdwolfz: :ok_hand:

doc/DOCUMENTATION.org
yuhan0: Do you mean the line
`| Any other key       | leave the layout transient state                           | `?
 It might be a small technicality but isn't it possible to have transient states which don't exit on any other key, by setting the `:foreign-keys` prop? (I'm not sure if there are any in the default Spacemacs layers)

yuhan0: Ok, edited the rest of the doc to remove those references. Let me know if/how I should squash the commits :)

yuhan0:
(resolved a documentation conflict and did a force push, no other changes made)

sdwolfz:
@yuhan0 yes please squash all commits into one.

yuhan0:
Hi any update on this? I've squashed the commits on both this and the other clj-bindings PR, thanks :)

sdwolfz:
Thank you :heart:!

PR #11546, opened 2018-11-04
[python] update jedi instructions

d12frosted:
Following https://github.com/proofit404/anaconda-mode/commit/150390968bb0c5180026932d27255483d690e00f

### DIFF
--- a/layers/+lang/python/README.org
+++ b/layers/+lang/python/README.org
@@ -94,10 +94,10 @@ opening a python buffer:
 To fix this, install the =anaconda-mode= [[https://github.com/proofit404/anaconda-mode/wiki][anaconda-deps]] by hand:
 
 #+BEGIN_SRC sh
-  pip install --upgrade "jedi>=0.9.0" "json-rpc>=1.8.1" "service_factory>=0.1.5"
+  pip install --upgrade "jedi>=0.13.0" "json-rpc>=1.8.1" "service_factory>=0.1.5"
 #+END_SRC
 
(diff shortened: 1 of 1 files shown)


bet4it:
After proofit404/anaconda-mode#327, `anaconda-mode` won't check for `jedi` which is installed globally, but will install it in `~/.emacs.d/.cache/anaconda-mode/0.1.13`, so the `pip install` command has no effects with `anaconda-mode`.
We should delete these workarounds with `anaconda-mode` now.

d12frosted:
@bet4it cool! thank you for sharing this information!

d12frosted:
@bet4it wait, I still see [this](https://github.com/proofit404/anaconda-mode/blob/master/anaconda-mode.el#L150). If I understand it correctly, if you haven't installed `jedi >= 0.13`, then you'll get an assert.

bet4it:
proofit404/anaconda-mode#327 was created to resolve the situation that we have to do some workarounds like the README described.

Sure, even if we download packages locally, we still need to import them and it's better to check its version to avoid something unexpected.
I think what have been done in `install_deps` is same with we run `pip install` manually, except it will install them locally. There is no exception handling in this function, so if there are any errors during the installing, the following expressions won't be executed. So the `assert` will only be executed when the installing is success, and I can't image any situation that `jedi` still can't satisfy the requirement that we have to install them manually now.

I think it's better for us to notice user to take a look at the output in `*anaconda-mode*` buffer when `anaconda-mode` can work.

bet4it:
OMG, there is one situation that we should update `jedi` manually: https://github.com/proofit404/anaconda-mode/pull/327#issuecomment-436956933
I think we can wait `anaconda-mode` for a solution to it.

smile13241324:

PR #11551, opened 2018-11-05
fix: lsp-ui-peek-find-definitions is an async jump handler

yetone:
To avoid multiple peek popup situation:

image


yetone:
I don't known why this ci railed -, -

sdwolfz:
Your PR contains extra commits which are now merged in develop, I suggest you rebase and only keep your one-line chage.

Issue #8171, opened 2017-01-14
spacemacs/default-pop-shell conflict with window-purpose


deb0ch:
#### Description :octocat:
When having a window layout with one `edit` purposed window and a `general` purposed window side by side, `SPC '` (`spacemacs/default-pop-shell`) has a strange behaviour when used from the `edit` window.

@bmag, I feel like this one is for you :smile_cat: 

#### Reproduction guide :beetle:
- Start Emacs
- open a code buffer and its corresponding `magit-status` buffer:
![screenshot from 2017-01-14 18-46-52](https://cloud.githubusercontent.com/assets/6248219/21956879/4eacc89e-da8a-11e6-81c7-5839b3d96113.png)
- select the code window
- `SPC '` (`spacemacs/default-pop-shell`)

*Observed behaviour:* :eyes: :broken_heart:
![screenshot from 2017-01-14 18-46-38](https://cloud.githubusercontent.com/assets/6248219/21956890/8530fc14-da8a-11e6-9461-adb378bbb40c.png)

*Expected behaviour:* :heart: :smile:
![screenshot from 2017-01-14 18-51-42](https://cloud.githubusercontent.com/assets/6248219/21956891/8f262230-da8a-11e6-9587-b15e666a0c27.png)

Note that this does not happen when the `magit` window is selected.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 24.5.1
- Spacemacs: 0.200.7
- Spacemacs branch: develop (rev. dc36181)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: helm
- Layers:
```elisp
((auto-completion :variables :disabled-for markdown org erc)
 (better-defaults :variables better-defaults-move-to-beginning-of-code-first nil better-defaults-move-to-end-of-code-first t)
 (c-c++ :variables c-c++-default-mode-for-headers 'c++-mode c-c++-enable-clang-support nil)
 (colors :variables colors-enable-nyan-cat-progress-bar nil)
 common-lisp csv emacs-lisp
 (erc :packages
      (not erc-social-graph)
      :variables erc-server-list
      '(("jinan.parrot.biz" :ssl t :port "7000" :nick "ThomdB" :full-name "Thomas de Beauchene")))
 git github gtags helm markdown multiple-cursors org pdf-tools
 (python :variables python-enable-yapf-format-on-save t python-sort-imports-on-save t)
 ranger
[...]

PR #11552, opened 2018-11-05
Add support for libvterm

deb0ch:
Add support for https://github.com/akermu/emacs-libvterm in Spacemacs.

I also made some fixes that are related to the shell layer but not to vterm and would be cherry-pickable without the vterm support, for instance disabling `global-center-cursor-mode` in terminals as it made them flicker.

### DIFF
--- a/layers/+misc/multiple-cursors/config.el
+++ b/layers/+misc/multiple-cursors/config.el
@@ -12,4 +12,5 @@
 ;;; License: GPLv3
 
 (defvar multiple-cursors-backend 'evil-mc
-  "The package used as the backend for multiple cursors functionality.")
+  "The package used as the backend for multiple cursors functionality.
+Possible values are `mc' or `evil-mc'.")
--- a/layers/+tools/shell/README.org
+++ b/layers/+tools/shell/README.org
@@ -8,6 +8,15 @@
 - [[#description][Description]]
   - [[#features][Features:]]
 - [[#install][Install]]
+  - [[#install-vterm][Install vterm]]
+    - [[#check-that-your-emacs-supports-dynamic-modules][Check that your Emacs supports dynamic modules]]
+    - [[#install-cmake-311-or-higher][Install CMake 3.11 or higher]]
+    - [[#install-libtool][Install libtool]]
+      - [[#ubuntu][Ubuntu]]
+    - [[#install-libvterm][Install libvterm]]
+      - [[#macos][MacOS]]
+      - [[#ubuntu-1][Ubuntu]]
+      - [[#windows][Windows]]
 - [[#configuration][Configuration]]
   - [[#default-shell][Default shell]]
   - [[#default-shell-position-width-and-height][Default shell position, width, and height]]
@@ -34,23 +43,67 @@ To use this configuration layer, add it to your =~/.spacemacs=. You will need to
 add =shell= to the existing =dotspacemacs-configuration-layers= list in this
 file.
 
+** Install vterm
+=vterm= is the latest addition to Emacs' set of terminal emulators and the only
+one to be implemented in C, leveraging =libvterm=. It is the only one in Emacs
+at the moment to be as fast as a standalone terminal with full support for
+=ncurses=, =vim=, =htop= and the likes.
+
+On its first run, =vterm= will automatically compile its dynamic library, for
+which dependencies are needed. For more details, head to the [[https://github.com/akermu/emacs-libvterm][official docs]].
+
+*** Check that your Emacs supports dynamic modules
+You can check if your Emacs supports loading dynamic libraries by checking if
+the =system-configuration-features= variable contains the string =MODULES=. If
+not, you need to get a version of Emacs that supports it or compile it from
+source supplying the =./configure --with-module= option at configure time.
+
+*** Install CMake 3.11 or higher
+
+*** Install libtool
+If the =libtool= command does not exist in your system (usually in
+=/usr/bin/libtool=), you need to install it:
+**** Ubuntu
+#+BEGIN_SRC shell
+sudo apt install libtool-bin
+#+END_SRC
+
+*** Install libvterm
+**** MacOS
+#+BEGIN_SRC shell
+brew install libvterm
+#+END_SRC
+
+**** Ubuntu
+#+BEGIN_SRC shell
+sudo apt install libvterm-dev
+#+END_SRC
+
+In other distros, find it and install it in your favorite package manager.
+
+**** Windows
+Not supported at the moment, [[https://github.com/akermu/emacs-libvterm/issues/12][but possibly coming up]].
+
 * Configuration
+
 ** Default shell
-Emacs supports three types of shell:
-- the Emacs shell
+Emacs supports five types of shells/terminals:
+- the Emacs shell (eshell)
 - the inferior shell
 - the terminal emulator
 - the ANSI terminal emulator
+- the vterm terminal emulator based on the C library libvterm
 
 You can find a quick introductions to them [[https://www.masteringemacs.org/article/running-shells-in-emacs-overview][here]].
 
 To define the default shell you can set the layer variable =shell-default-shell=
 to the following variables:
-- =eshell=
+- =eshell= (default on Windows)
 - =shell=
 - =term=
-- =ansi-term=
+- =ansi-term= (default on Linux/macOS)
 - =multi-term=
+- =vterm=
 
 #+BEGIN_SRC emacs-lisp
   (setq-default dotspacemacs-configuration-layers
@@ -224,6 +277,7 @@ Some advanced configuration is setup for =eshell= in this layer:
 | ~SPC a s T~ | Open, close or go to a =term=                              |
 | ~TAB~       | browse completion with =helm=                              |
 | ~SPC m H~   | browse history with =helm= (works in =eshell= and =shell=) |
+| ~SPC a s v~ | Open, close or go to a =vterm=                             |
 | ~C-j~       | next item in history                                       |
 | ~C-k~       | previous item in history                                   |
 
--- a/layers/+tools/shell/funcs.el
+++ b/layers/+tools/shell/funcs.el
@@ -208,3 +208,12 @@ is achieved by adding the relevant text properties."
   (interactive)
   (switch-to-buffer "*shell*")
   (shell "*shell*"))
+
+;; https://stackoverflow.com/questions/6837511/automatically-disable-a-global-minor-mode-for-a-specific-major-mode
+(defun spacemacs//inhibit-global-centered-cursor-mode ()
+  "Counter-act `global-centered-cursor-mode'."
+  (add-hook 'after-change-major-mode-hook
+            (lambda ()
+              (centered-cursor-mode 0))
+            :append
+            :local))
--- a/layers/+tools/shell/funcs.el
+++ b/layers/+tools/shell/funcs.el
@@ -57,28 +57,28 @@
                                        100)
                                     (window-width)))))
 
-(defmacro make-shell-pop-command (func &optional shell)
+(defmacro make-shell-pop-command (name func &optional shell)
   "Create a function to open a shell via the function FUNC.
 SHELL is the SHELL function to use (i.e. when FUNC represents a terminal)."
-  (let* ((name (symbol-name func)))
-    `(defun ,(intern (concat "spacemacs/shell-pop-" name)) (index)
-       ,(format (concat "Toggle a popup window with `%S'.\n"
-                        "Multiple shells can be opened with a numerical prefix "
-                        "argument. Using the universal prefix argument will "
-                        "open the shell in the current buffer instead of a "
-                        "popup buffer.") func)
-       (interactive "P")
-       (require 'shell-pop)
-       (if (equal '(4) index)
-           ;; no popup
-           (,func ,shell)
-         (shell-pop--set-shell-type
-          'shell-pop-shell-type
-          (backquote (,name
-                      ,(concat "*" name "*")
-                      (lambda nil (,func ,shell)))))
-         (shell-pop index)
-         (spacemacs/resize-shell-to-desired-width)))))
+  `(defun ,(intern (concat "spacemacs/shell-pop-" name)) (index)
+     ,(format (concat "Toggle a popup window with `%S'.\n"
+                      "Multiple shells can be opened with a numerical prefix "
+                      "argument. Using the universal prefix argument will "
+                      "open the shell in the current buffer instead of a "
+                      "popup buffer.")
+              func)
+     (interactive "P")
+     (require 'shell-pop)
+     (if (equal '(4) index)
+         ;; no popup
+         (,func ,shell)
+       (shell-pop--set-shell-type
+        'shell-pop-shell-type
+        (backquote (,name
+                    ,(concat "*" name "*")
+                    (lambda nil (,func ,shell)))))
+       (shell-pop index)
+       (spacemacs/resize-shell-to-desired-width))))
 
 (defun projectile-multi-term-in-root ()
   "Invoke `multi-term' in the project's root."
--- a/layers/+tools/shell/packages.el
+++ b/layers/+tools/shell/packages.el
@@ -32,7 +32,9 @@
 
 (defun shell/init-comint ()
   (setq comint-prompt-read-only t)
-  (add-hook 'comint-mode-hook 'spacemacs/disable-hl-line-mode))
+  (add-hook 'comint-mode-hook 'spacemacs/disable-hl-line-mode)
+  (with-eval-after-load 'centered-cursor-mode
+    (add-hook 'comint-mode-hook 'spacemacs//inhibit-global-centered-cursor-mode)))
 
 (defun shell/pre-init-company ()
   ;; support in eshell
@@ -81,7 +83,9 @@
       (autoload 'eshell-delchar-or-maybe-eof "em-rebind")
 
       (add-hook 'eshell-mode-hook 'spacemacs//init-eshell)
-      (add-hook 'eshell-mode-hook 'spacemacs/disable-hl-line-mode))
+      (add-hook 'eshell-mode-hook 'spacemacs/disable-hl-line-mode)
+      (with-eval-after-load 'centered-cursor-mode
+        (add-hook 'eshell-mode-hook 'spacemacs//inhibit-global-centered-cursor-mode)))
     :config
     (progn
 
@@ -214,7 +218,9 @@
              ;; Send other commands to the default handler.
              (t (comint-simple-send proc command))))))
   (add-hook 'shell-mode-hook 'shell-comint-input-sender-hook)
-  (add-hook 'shell-mode-hook 'spacemacs/disable-hl-line-mode))
+  (add-hook 'shell-mode-hook 'spacemacs/disable-hl-line-mode)
+  (with-eval-after-load 'centered-cursor-mode
+    (add-hook 'shell-mode-hook 'spacemacs//inhibit-global-centered-cursor-mode)))
 
 (defun shell/init-shell-pop ()
   (use-package shell-pop
@@ -270,7 +276,9 @@
     (kbd "C-k") 'term-send-up
     (kbd "C-j") 'term-send-down)
 
-  (add-hook 'term-mode-hook 'spacemacs/disable-hl-line-mode))
+  (add-hook 'term-mode-hook 'spacemacs/disable-hl-line-mode)
+  (with-eval-after-load 'centered-cursor-mode
+    (add-hook 'term-mode-hook 'spacemacs//inhibit-global-centered-cursor-mode)))
 
 (defun shell/init-xterm-color ()
   (use-package xterm-color
--- a/layers/+tools/shell/packages.el
+++ b/layers/+tools/shell/packages.el
@@ -258,24 +258,26 @@
     "Send tab in term mode."
     (interactive)
     (term-send-raw-string "\t"))
-  ;; hack to fix pasting issue, the paste transient-state won't
-  ;; work in term
-  (evil-define-key 'normal term-raw-map "p" 'term-paste)
-  (evil-define-key 'normal term-raw-map (kbd "<mouse-2>") 'term-mouse-paste)
-  (evil-define-key 'insert term-raw-map (kbd "<mouse-2>") 'term-mouse-paste)
-  (evil-define-key 'normal term-raw-map (kbd "<mouse-3>") 'term-mouse-paste)
-  (evil-define-key 'insert term-raw-map (kbd "<mouse-3>") 'term-mouse-paste)
-  (evil-define-key 'insert term-raw-map (kbd "C-c C-d") 'term-send-eof)
-  (evil-define-key 'insert term-raw-map (kbd "C-c C-z") 'term-stop-subjob)
-  (evil-define-key 'insert term-raw-map (kbd "<tab>") 'term-send-tab)
 
   (when (eq dotspacemacs-editing-style 'vim)
     (evil-define-key 'insert term-raw-map
       (kbd "C-k") 'term-send-up
       (kbd "C-j") 'term-send-down))
+
+  (evil-define-key 'insert term-raw-map
+    (kbd "<mouse-2>") 'term-mouse-paste
+    (kbd "<mouse-3>") 'term-mouse-paste
+    (kbd "C-c C-d") 'term-send-eof
+    (kbd "C-c C-z") 'term-stop-subjob
+    (kbd "<tab>") 'term-send-tab)
+
   (evil-define-key 'normal term-raw-map
+    (kbd "<mouse-2>") 'term-mouse-paste
+    (kbd "<mouse-3>") 'term-mouse-paste
     (kbd "C-k") 'term-send-up
-    (kbd "C-j") 'term-send-down)
+    (kbd "C-j") 'term-send-down
+    ;; hack to fix pasting issue, the paste transient-state won't work in term
+    "p" 'term-paste)
 
   (add-hook 'term-mode-hook 'spacemacs/disable-hl-line-mode)
   (with-eval-after-load 'centered-cursor-mode
--- a/layers/+tools/shell/packages.el
+++ b/layers/+tools/shell/packages.el
@@ -231,7 +231,8 @@
       (setq shell-pop-window-position shell-default-position
             shell-pop-window-size     shell-default-height
             shell-pop-term-shell      shell-default-term-shell
-            shell-pop-full-span       shell-default-full-span)
+            shell-pop-full-span       shell-default-full-span
+            shell-pop-in-after-hook   #'evil-insert-state)
       (make-shell-pop-command "eshell" eshell)
       (make-shell-pop-command "term" term shell-pop-term-shell)
       (make-shell-pop-command "ansi-term" ansi-term shell-pop-term-shell)
--- a/layers/+tools/shell/packages.el
+++ b/layers/+tools/shell/packages.el
@@ -28,6 +28,7 @@
         (term :location built-in)
         xterm-color
         vi-tilde-fringe
+        (vterm :toggle (not (spacemacs/system-is-mswindows)))
         ))
 
 (defun shell/init-comint ()
@@ -231,14 +232,13 @@
             shell-pop-window-size     shell-default-height
             shell-pop-term-shell      shell-default-term-shell
             shell-pop-full-span       shell-default-full-span)
-      (make-shell-pop-command eshell)
-      (make-shell-pop-command term shell-pop-term-shell)
-      (make-shell-pop-command ansi-term shell-pop-term-shell)
-      (make-shell-pop-command inferior-shell)
-      (make-shell-pop-command multiterm)
+      (make-shell-pop-command "eshell" eshell)
+      (make-shell-pop-command "term" term shell-pop-term-shell)
+      (make-shell-pop-command "ansi-term" ansi-term shell-pop-term-shell)
+      (make-shell-pop-command "inferior-shell" inferior-shell)
+      (make-shell-pop-command "multiterm" multiterm)
 
       (add-hook 'term-mode-hook 'ansi-term-handle-close)
-      (add-hook 'term-mode-hook (lambda () (linum-mode -1)))
 
       (spacemacs/set-leader-keys
         "'"   'spacemacs/default-pop-shell
@@ -296,3 +296,32 @@
                             eshell-mode-hook
                             shell-mode-hook
                             term-mode-hook)))
+
+(defun shell/init-vterm ()
+  (use-package vterm
+    :defer t
+    :commands (vterm vterm-other-window)
+
+    :init
+    (progn
+      (make-shell-pop-command "vterm" vterm)
+      (spacemacs/set-leader-keys "asv" 'spacemacs/shell-pop-vterm)
+      (spacemacs/register-repl 'vterm 'vterm))
+
+    :config
+    (progn
+      (evil-define-key 'normal vterm-mode-map
+        [escape] 'vterm--self-insert
+        [return] 'vterm--self-insert)
+
+      (add-hook 'vterm-mode-hook 'spacemacs/disable-hl-line-mode)
+
+      (with-eval-after-load 'centered-cursor-mode
+        (add-hook 'vterm-mode-hook 'spacemacs//inhibit-global-centered-cursor-mode))
+
+      (with-eval-after-load 'window-purpose
+        (purpose-set-extension-configuration
+         :vterm
+         (purpose-conf "vterm"
+                       :mode-purposes
+                       '((vterm-mode . terminal))))))))
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2467,6 +2467,7 @@ Other:
 - Disabled =semantic-idle-summary-mode= when =gtags= or =lsp= layer is enabled
   (thanks to bet4it)
 **** Shell
+- Improved the integration of centered-cursor-mode (thanks to deb0ch)
 - Fixed xterm colors filtering bug in =eshell= when eshell buffers are updated
   and are not focused (thanks to Steven Allen)
 - Fixed =eshell= clear behavior (thanks to Aidan Nyquist):
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2467,6 +2467,7 @@ Other:
 - Disabled =semantic-idle-summary-mode= when =gtags= or =lsp= layer is enabled
   (thanks to bet4it)
 **** Shell
+- Added support for =vterm= (thanks to deb0ch)
 - Improved the integration of centered-cursor-mode (thanks to deb0ch)
 - Fixed xterm colors filtering bug in =eshell= when eshell buffers are updated
   and are not focused (thanks to Steven Allen)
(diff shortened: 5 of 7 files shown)


deb0ch:
**Update**: the latest commit fixes the startup error when `vterm` was not added to `dotspacemacs-additional-packages`. 

Not sure it is the best way to go, but at least the PR is functional.

Compro-Prasad:
@deb0ch I don't think it is a good addition to spacemacs as it lacks scrolling behaviour.

deb0ch:
@Compro-Prasad You actually can scroll in `vterm` in at least two ways:

1. Now you can scroll natively with `C-v` and `M-v`. It _was_ lacking native scrolling behaviour until  almost two weeks ago when https://github.com/akermu/emacs-libvterm/pull/21 was merged. Only the mouse events are not bound yet (but who cares for the mouse anyways).
2. As we are using `evil`, you can scroll back anywhere you like in your vterm buffer using the usual cursor movement bindings.

I want to outline that I would like this merged as an _experimental_ feature, the same way `evil-mc` is.  This way more people, especially from Spacemacs, can start playing around with this astoundingly promising package and perhaps start contributing to make it even better.

For now it is already very usable, more so than others like `multi-term` or `ansi-term` to my taste. What a pleasure to be able to run stuff like `vim` or `htop` from inside Emacs 😺

bmag:
@deb0ch I'm curious about the idea, but you must include some installation instructions. I'd like to try vterm, but I don't have prior knowledge of it, so how do I install and use this feature? At minimum it looks like I need to compile some C code by myself. Add some explanation and installation instructions to `shell` layer's readme.

There's more to review about this PR, but I prefer to understand how to use (and test) it first. The readme here is at least as important as the technical Elisp parts.

deb0ch:
Updated with some docs :thumbsup::sparkles:

deb0ch:
@bmag Updated again ! You don't need to worry about compiling anything anymore, as it is now done automatically 😁 

When https://github.com/akermu/emacs-libvterm/pull/35 the compilation will even be completely transparent, all you will see will be a slight 1-2s delay when opening your `vterm` terminal for the first time.

haormj:
Hi @deb0ch, I want to try, but I don't know how to do it, is there any plan to merge in develop branch? :-)

layers/+tools/shell/README.org
```diff
@@ -30,14 +38,51 @@ This layer configures the various shells available in Emacs.
 To use this configuration layer, add it to your =~/.spacemacs=. You will need to
 add =shell= to the existing =dotspacemacs-configuration-layers= list in this
 file.
+** Install vterm (EXPERIMENTAL)
```
duianto: Add an empty line.
```suggestion

** Install vterm (EXPERIMENTAL)
```

layers/+tools/shell/README.org
```diff
@@ -30,14 +38,51 @@ This layer configures the various shells available in Emacs.
...
+=vterm= is the latest addition to Emacs' set of terminal emulators and the only
+one to be implemented in C, leveraging =libvterm=. It is the only one in Emacs at
+the moment to be as fast as a standalone terminal and to support =ncurses=, =vim=,
+=top=, =htop= and the likes.
+*** Install libvterm
```
duianto: Add an empty line.
```suggestion

*** Install libvterm
```

layers/+tools/shell/README.org
```diff
@@ -30,14 +38,51 @@ This layer configures the various shells available in Emacs.
...
+In Ubtuntu:
+
+#+BEGIN_SRC shell
+sudo apt-get install libvterm
+#+END_SRC
+
+In other distros, find it and install it in your favorite package manager.
+
+**** Windows
+Not supported at the moment, but [[https://github.com/akermu/emacs-libvterm/issues/12][but possibly coming up]].
```
duianto: Remove duplicate "but".
```suggestion
Not supported at the moment, [[https://github.com/akermu/emacs-libvterm/issues/12][but possibly coming up]].
```

layers/+tools/shell/README.org
```diff
@@ -30,14 +38,51 @@ This layer configures the various shells available in Emacs.
...
+
+*** Clone emacs-libvterm
+#+BEGIN_SRC shell
+  git clone https://github.com/akermu/emacs-libvterm.git ~/.emacs.d/private/local/emacs-libvterm
+#+END_SRC
+
+*** Add it to your additional packages
```
duianto: This section seems to be outdated since it's on Melpa now. https://melpa.org/#/vterm

layers/+tools/shell/README.org
```diff
@@ -30,14 +38,51 @@ This layer configures the various shells available in Emacs.
...
+  (setq-default dotspacemacs-additional-packages
+    '((vterm :location "~/.emacs.d/private/local/emacs-libvterm")))
+#+END_SRC
 
 * Configuration
 ** Default shell
-Emacs supports three types of shell:
-- the Emacs shell
+Emacs supports five types of shells/terminals:
+- the Emacs shell eshell
```
duianto: Maybe there should be a parentheses around `eshell`.
```suggestion
- the Emacs shell (eshell)
```

layers/+tools/shell/README.org
```diff
@@ -233,3 +279,13 @@ in the current buffer instead of a popup.
 | Key Binding        | Description                                |
 |--------------------+--------------------------------------------|
 | ~SPC m H~ or ~M-l~ | shell commands history using a helm buffer |
+
+
+
+
+** Vterm (EXPERIMENTAL)
```
duianto: Reduce the four empty lines above to one.
Github only shows them as one.
https://github.com/syl20bnr/spacemacs/blob/f4a8ac0f7c9244b90a3a64a5a380eee0442caf96/layers/%2Btools/shell/README.org

layers/+tools/shell/funcs.el
```diff
@@ -56,28 +56,27 @@
...
-                      ,(concat "*" name "*")
-                      (lambda nil (,func ,shell)))))
-         (shell-pop index)
-         (spacemacs/resize-shell-to-desired-width)))))
+  `(defun ,(intern (concat "spacemacs/shell-pop-" name)) (index)
+     ,(format (concat "Toggle a popup window with `%S'.\n"
+                      "Multiple shells can be opened with a numerical prefix "
+                      "argument. Using the universal prefix argument will "
+                      "open the shell in the current buffer instead of a "
+                      "popup buffer.") func)
```
duianto: This was not introduced by this PR, but we might as well fix it at the same time.

Move `func` to the line below to remove the `font-lock-warning-face` (orange color):
With the cursor on `func` and calling `describe-char` (`SPC h d c`), the help buffer says:
>There are text properties here:
  face                 font-lock-warning-face
  fontified            nil
  help-echo            "Hidden behind deeper element; move to another line?"

```suggestion
                      "popup buffer.")
              func)
```

layers/+tools/shell/packages.el
```diff
@@ -29,9 +29,16 @@
         vi-tilde-fringe
         ))
 
+;; [EXPERIMENTAL] As the vterm package does not exist in any of Emacs' package
```
duianto: Since it's on melpa, then this might not be needed.

deb0ch:
Hey hey, I will update all that as soon as I have some free time to work on it 😸 

I saw a lot of movement upstream since I wrote this integration, I will take a serious look at what needs to be changed, fix conflicts, melpa-ize it and all 👍

layers/+tools/shell/funcs.el
deb0ch: Done 👍 ✨

deb0ch:
I just could get around to it and finished updating the code ! 🎉 

I'm still having some failed tests on Emacs 25, 26 and Snapshot that I'm having troubles understanding 🤔 (the first error only happens for Emacs Snapshot):

```
FUNCTIONAL TESTS FOR CORE
=================================================================
Running 17 tests (2019-08-29 18:53:49+0000, selector `t')
   passed   1/17  assertion-library-should-work (0.000345 sec)
Package cl is deprecated
   passed   2/17  test-declare-layers--base-layer-is-third-for-base-distribution (0.062676 sec)
   passed   3/17  test-declare-layers--bootstrap-layer-always-first (0.107298 sec)
   passed   4/17  test-declare-layers--defaults-layer-is-second-for-base-distribution (0.063489 sec)
   passed   5/17  test-git-fetch-tags--command-is-working (0.329175 sec)
   passed   6/17  test-git-fetch-tags--input-is-not-a-remote (0.004315 sec)
   passed   7/17  test-git-has-remote--command-is-working (0.001988 sec)
   passed   8/17  test-git-has-remote--input-is-not-a-remote (0.002191 sec)
   passed   9/17  test-render-framed-text--file-caption-and-padding (0.000701 sec)
   passed  10/17  test-render-framed-text--msg-no-width-no-caption-no-padding (0.015346 sec)
   passed  11/17  test-render-framed-text--msg-several-paragraphs (0.000298 sec)
   passed  12/17  test-render-framed-text--msg-short-text (0.000127 sec)
   passed  13/17  test-render-framed-text--msg-width-caption-and-padding (0.000210 sec)
   passed  14/17  test-render-framed-text--msg-width-caption-no-padding (0.000186 sec)
   passed  15/17  test-render-framed-text--msg-width-no-caption-no-padding (0.000181 sec)
Test test-stable-elpa-verify-archive--verification-failed backtrace:
  signal(ert-test-failed (((should (and (null (configuration-layer//st
  ert-fail(((should (and (null (configuration-layer//stable-elpa-verif
  (if (unwind-protect (setq value-65 (and (null (configuration-layer//
  (let (form-description-66) (if (unwind-protect (setq value-65 (and (
  (let ((value-65 (gensym "ert-form-evaluation-aborted-"))) (let (form
  (progn (fset 'message 'ignore) (fset 'configuration-layer//error vne
  (unwind-protect (progn (fset 'message 'ignore) (fset 'configuration-
  (let* ((vnew #'(lambda nil (concat spacemacs-test-directory "core/da
  (let (verification-error) (let* ((vnew #'(lambda nil (concat spacema
  (lambda nil (let (verification-error) (let* ((vnew #'(lambda nil (co
  ert--run-test-internal(#s(ert--test-execution-info :test #s(ert-test
  ert-run-test(#s(ert-test :name test-stable-elpa-verify-archive--veri
  ert-run-or-rerun-test(#s(ert--stats :selector t :tests [... ... ... 
  ert-run-tests(t #f(compiled-function (event-type &rest event-args) #
  ert-run-tests-batch(nil)
  ert-run-tests-batch-and-exit()
  command-line-1(("-l" "ert" "-l" "/root/.emacs.d/core/core-versions.e
  command-line()
  normal-top-level()
Test test-stable-elpa-verify-archive--verification-failed condition:
    (ert-test-failed
     ((should
       (and
	(null ...)
	(string-match-p "^Verification failed!.*" verification-error)))
      :form
      (and
       (null
	(configuration-layer//stable-elpa-verify-archive))
       (string-match-p "^Verification failed!.*" verification-error))
      :value nil))
   FAILED  16/17  test-stable-elpa-verify-archive--verification-failed (0.081515 sec)
Test test-stable-elpa-verify-archive--verification-ok backtrace:
  signal(ert-test-failed (((should (equal t (configuration-layer//stab
  ert-fail(((should (equal t (configuration-layer//stable-elpa-verify-
  (if (unwind-protect (setq value-62 (apply fn-60 args-61)) (setq form
  (let (form-description-64) (if (unwind-protect (setq value-62 (apply
  (let ((value-62 'ert-form-evaluation-aborted-63)) (let (form-descrip
  (let* ((fn-60 #'equal) (args-61 (condition-case err (let ((signal-ho
  (progn (fset 'message 'ignore) (fset 'configuration-layer//error vne
  (unwind-protect (progn (fset 'message 'ignore) (fset 'configuration-
  (let* ((vnew #'(lambda nil (concat spacemacs-test-directory "core/da
  (lambda nil (let* ((vnew #'(lambda nil (concat spacemacs-test-direct
  ert--run-test-internal(#s(ert--test-execution-info :test #s(ert-test
  ert-run-test(#s(ert-test :name test-stable-elpa-verify-archive--veri
  ert-run-or-rerun-test(#s(ert--stats :selector t :tests [... ... ... 
  ert-run-tests(t #f(compiled-function (event-type &rest event-args) #
  ert-run-tests-batch(nil)
  ert-run-tests-batch-and-exit()
  command-line-1(("-l" "ert" "-l" "/root/.emacs.d/core/core-versions.e
  command-line()
  normal-top-level()
Test test-stable-elpa-verify-archive--verification-ok condition:
    (ert-test-failed
     ((should
       (equal t
	      (configuration-layer//stable-elpa-verify-archive)))
      :form
      (equal t nil)
      :value nil :explanation
      (different-atoms t nil)))
   FAILED  17/17  test-stable-elpa-verify-archive--verification-ok (0.013287 sec)

Ran 17 tests, 15 results as expected, 2 unexpected (2019-08-29 18:53:50+0000, 1.111115 sec)

2 unexpected results:
   FAILED  test-stable-elpa-verify-archive--verification-failed
   FAILED  test-stable-elpa-verify-archive--verification-ok
```

duianto:
The tests are unrelated to PRs, I'm guessing that it's caused by the work being done on the spacelpa verification.

deb0ch:
I just cleaned up, reordered and fixed the description of the commits 👍 ✨ 

Now all of them are meaningful and they are best reviewed separately to understand what's going on.

duianto:
Commit message:
There's a trailing "j" in the url of the commit:
[shell] add support for vterm
Support for https://github.com/akermu/emacs-libvtermj

Readme.org:
- >**** GNU/Linux
In Ubtuntu:

Typo: Ubtuntu -> Ubuntu


- The linux instruction might need some additional info:
When I try: `sudo apt-get install libvterm` in Ubuntu:
>Reading package lists... Done
Building dependency tree       
Reading state information... Done
E: Unable to locate package libvterm

layers/+tools/shell/README.org
```diff
@@ -51,10 +82,11 @@ to the following variables:
 - =term=
 - =ansi-term=
 - =multi-term=
+- =vterm=
 
 #+BEGIN_SRC emacs-lisp
   (setq-default dotspacemacs-configuration-layers
-    '((shell :variables shell-default-shell 'eshell)))
+    '((shell :variables shell-default-shell 'vterm)))
```
duianto:
```suggestion
    '((shell :variables shell-default-shell 'eshell)))
```
It might be best to leave this as `eshell` since it's probably the best supported shell.
And could you add `(default)` after `- =eshell=` in the list of possible values for `shell-default-shell`

That's what was suggested for the multiple-cursors layer.
https://github.com/syl20bnr/spacemacs/tree/develop/layers/+misc/multiple-cursors#configuration

deb0ch: Haha, I am so enthusiastic about this terminal that I wanted to promote it as much as I could 😆 

After using it in Spacemacs for now over 11 month, I am convinced that it is far superior to all the other options in Emacs.

deb0ch: After checking I am not sure that we should mention `(default)` beside `eshell`, since it is only the default on Windows:

```elisp
(defvar shell-default-shell (if (eq window-system 'w32)
                                'eshell
                              'ansi-term)
  "Default shell to use in Spacemacs. Possible values are `eshell' (default),
`shell', `term', `ansi-term', `multi-term' and `vterm'.")
```

duianto: 😄 I'll ask the maintainers what they think.

deb0ch:
Fixed, fixed, fixed and fixed ! 👍 ✨ 

I took all your comments into account.

layers/+tools/shell/README.org
deb0ch: I wrote it as `(default on Windows)` and `(default on macOS)`, I don't think such trivial detail is worth bothering anyone who is not already following this PR.

duianto: I don't have much experience with shell (so you have a noob reviewing this 😄), I didn't know that the defaults were different.

That's good, It might be helpful for someone reading the readme, to know what the default value of `shell-default-shell` is if it isn't defined.

duianto:
I installed cmake with `sudo apt-get install cmake` but by default Ubuntu installed a to old version:
>CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
  CMake 3.11 or higher is required.  You are running version 3.10.2
>
>-- Configuring incomplete, errors occurred!
make: *** No targets specified and no makefile found.  Stop.

The instructions probably need to mention how to get the latest (or at least v3.11).

duianto:
The same thing happened without Spacemacs. With just Emacs and `vterm`. `M-x vterm` started showing messages in the minibuffer, I think at least some of them showed "compiling", but then Emacs closed.

deb0ch:
What version of Ubuntu are you running ?

I installed Emacs 26 with Spacemacs (develop) and my `vterm` branch on a fresh virtualized Ubuntu 18.04 AND at an other occasion a 19.04 without problems. My only issues were when I had to compile from source Emacs, CMake and libvterm for an old Ubuntu 14.04.

Maybe is it a bug that was introduced recently upstream, in the `emacs-libvterm` project ?

duianto:
Ubuntu 18.04.3 LTS

duianto:
https://github.com/akermu/emacs-libvterm
says:
>Emacs with module support. You can check that, by verifying that module-file-suffix is not nil.

emacs-snapshot shows:
`emacs-version`
>GNU Emacs 27.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.22.30) of 2019-08-30

`C-h v module-file-suffix`
>module-file-suffix is a variable defined in ‘C source code’.
Its value is ".so"

I also tried 26.3 by installing it from Ubuntu Software, but `M-x vterm` says:
>byte-code: VTerm needs module support. Please compile your Emacs with the --with-modules option!

layers/+tools/shell/funcs.el
```diff
@@ -208,3 +208,12 @@ is achieved by adding the relevant text properties."
   (interactive)
   (switch-to-buffer "*shell*")
   (shell "*shell*"))
+
+;; https://stackoverflow.com/questions/6837511/automatically-disable-a-global-minor-mode-for-a-specific-major-mode
+(defun spacemacs//inhibit-global-centered-cursor-mode ()
+  "Counter-act `global-centered-cursor-mode'."
```
bmag: If possible, better to hook locally into `centered-cursor-mode-hook`.

bmag: Also If possible, better to get upstream to use a smarter TURN-ON argument in call to `define-global-minor-mode'` Usually a `centered-cursor-mode-on` function, that checks if the current major-mode should have the minor-mode enabled (e.g. by consulting a blacklist/whitelist of major-modes) before activating, is used as the TURN-ON argument. For an example see `global-company-mode`, `company-mode-on` and `company-global-modes`.

Note these upstream changes are good as future plans, and they are not a requirement for this PR. We will merge it even without them.

layers/+tools/shell/README.org
```diff
@@ -34,23 +40,51 @@ To use this configuration layer, add it to your =~/.spacemacs=. You will need to
 add =shell= to the existing =dotspacemacs-configuration-layers= list in this
 file.
 
+** Install vterm
+=vterm= is the latest addition to Emacs' set of terminal emulators and the only
+one to be implemented in C, leveraging =libvterm=. It is the only one in Emacs at
+the moment to be as fast as a standalone terminal and to support =ncurses=, =vim=,
+=top=, =htop= and the likes.
```
bmag: Add list of dependencies (as of today), plus link to emacs-libvterm dependency documentation. Include the --with-modules requirement and how to check if Emacs was compiled with --with-modules (variable `system-configuration-options` contains "--with-modules" or variable `system-configuration-features` contains "MODULES").

bmag: Try to explain what emacs-libvterm tries to do when it insalls itself, which happens the first time the user tries to open vterm (compile emacs-libvterm.c, download stuff? etc.).

layers/+tools/shell/funcs.el
```diff
@@ -57,28 +57,28 @@
                                        100)
                                     (window-width)))))
 
-(defmacro make-shell-pop-command (func &optional shell)
+(defmacro make-shell-pop-command (name func &optional shell)
```
bmag: I like this change, it's a bit cleaner this way :+1:

layers/+tools/shell/packages.el
```diff
@@ -291,3 +290,23 @@
...
+  (use-package vterm
+    :defer t
+    :commands (vterm vterm-other-window)
+    :init
+    (progn
+      (make-shell-pop-command "vterm" spacemacs/open-vterm)
+      (spacemacs/set-leader-keys "asv" 'spacemacs/shell-pop-vterm)
+      (spacemacs/register-repl 'vterm 'vterm))
+    :config
+    (progn
```
bmag: is there a reason for these bindings?

layers/+tools/shell/packages.el
```diff
@@ -291,3 +290,23 @@
                             eshell-mode-hook
                             shell-mode-hook
                             term-mode-hook)))
+
+(defun shell/init-vterm ()
+  (use-package vterm
+    :defer t
+    :commands (vterm vterm-other-window)
+    :init
+    (progn
```
bmag: can't we just use `(make-...-command "vterm" vterm)` ? It seems to work.

layers/+tools/shell/funcs.el
```diff
@@ -209,6 +209,10 @@ is achieved by adding the relevant text properties."
   (switch-to-buffer "*shell*")
   (shell "*shell*"))
 
+(defun spacemacs/open-vterm (&optional ARG)
```
bmag: `spacemacs/open-vterm` might be unnecessary - we may be able to use `vterm` function directly.

layers/+tools/shell/packages.el
```diff
@@ -263,10 +263,9 @@
   (evil-define-key 'insert term-raw-map (kbd "C-c C-z") 'term-stop-subjob)
   (evil-define-key 'insert term-raw-map (kbd "<tab>") 'term-send-tab)
 
-  (when (eq dotspacemacs-editing-style 'vim)
-    (evil-define-key 'insert term-raw-map
-      (kbd "C-k") 'term-send-up
-      (kbd "C-j") 'term-send-down))
+  (evil-define-key 'insert term-raw-map
```
bmag: C-k and C-j need to depend on `dotspacemacs-editing-style` and `hybrid-style-enable-hjkl-bindings`.

layers/+tools/shell/packages.el
```diff
@@ -226,7 +226,8 @@
       (setq shell-pop-window-position shell-default-position
             shell-pop-window-size     shell-default-height
             shell-pop-term-shell      shell-default-term-shell
-            shell-pop-full-span       shell-default-full-span)
+            shell-pop-full-span       shell-default-full-span
+            shell-pop-in-after-hook   (lambda () (call-interactively #'evil-insert)))
```
bmag: Shouldn't we hook `evil-insert-state` instead of this lambda?

layers/+tools/shell/packages.el
```diff
@@ -309,4 +309,12 @@
...
       (add-hook 'vterm-mode-hook 'spacemacs/disable-hl-line-mode)
-      (add-hook 'vterm-mode-hook 'spacemacs//inhibit-global-centered-cursor-mode))))
+      (add-hook 'vterm-mode-hook 'spacemacs//inhibit-global-centered-cursor-mode)
+
+      (purpose-set-extension-configuration
+       :vterm
+       (purpose-conf "vterm"
+                     :mode-purposes
+                     '((vterm-mode . terminal))))
+      (add-to-list 'purpose-special-action-sequences
```
bmag: Changing `purpose-special-action-sequences` like this is an odd choice, what problem does it try to solve? There probably is a better solution. (I don't notice any problem if I remove this `add-to-list` call)

bmag:
@deb0ch I tried to make a separate review for each commit. Overall an excellent contribution, most of my comments are about documentation and peripheral issues.

Forgot to add this comment to one of the reviews: because vterm can't be installed on some systems, and it has a semi-complicated installation process, it should be opt-in. Add a `shell-enable-vterm-support` layer variable as a toggle for vterm, then in the packages list:
```elisp
(vterm :toggle shell-enable-vterm-support)
```
Obviously need to mention the variable in the install process, as well.

I believe most (if not all) of the installation issues will be independent of Spacemacs, and will be people struggling to install the build dependencies. (if Ubuntu 18.04 doesn't include cmake >= 3.11 it would be difficult for many people, but we can't do anything about it)

Can't wait to merge this :smile: 

P.S. the installation process is much smoother than last time I tried it

Also P.S. my system is openSUSE Tumbleweed (a rolling release), so getting up-to-date dependencies wasn't a problem.

bmag:
@duianto is it possible your problem is because you're testing on 27.0.50? Might not be a problem with 26.3 or 26.2.

duianto:
@bmag It's not at all impossible, the build is from today so maybe a change broke something.

layers/+tools/shell/funcs.el
deb0ch: Your approach would require to disable a minor mode from its own hook, which when I tried yielded an infinite recursion. Disabling from the global minor mode hook did not work either, surely because of the order in which hooks are run.

If you take a look at the Stackoverflow question in the comment, they tried to disable it from the incriminated major modes and it did not work because globalized minor modes use the `after-major-mode-change-hook`, which is run _after_ the major mode's hook. This is why what the guy did first, which was local into `centered-cursor-mode`, was not working.

Funny thing, they were even discussing precisely this mode, `centered-cursor-mode`.

All in all, the answer is clear, detailed, references well its sources and _works_, so I think we should stick to it.

layers/+tools/shell/packages.el
deb0ch: Maybe it changed, IIRC it did not at the time I first wrote this. I will test again and correct if it works.

layers/+tools/shell/funcs.el
bmag: Ok, lets keep it as is, even though I wonder if they tried:
```elisp
(lambda ()
  (when centered-cursor-mode
    (centered-cursor-mode 0)))
```

layers/+tools/shell/packages.el
deb0ch: It is a way of being able to use <RET> or <ESC> in normal mode, without having to go to insert mode. I found it convenient at the time as I was using a lot of interactive command line utilities, but now I haven't used it for a long time so I guess it is not _that_ important.

duianto:
@bmag You were correct. After compiling Emacs 26.3.50 (two irrelevant commits from 26.3), `SPC '` now opens a `*vterm-1*` buffer without any issues.

layers/+tools/shell/funcs.el
deb0ch: Tried again, it worked, it's fixed ✨

layers/+tools/shell/packages.el
deb0ch: Probably 😅

layers/+tools/shell/packages.el
deb0ch: Tested, it worked, it's fixed :sparkles:

layers/+tools/shell/funcs.el
deb0ch: Thanks ✨

deb0ch:
@duianto Good news, I was suspecting that it was unrelated to this PR. I guess either Emacs or emacs-libvterm will have to deal with this upstream.

@bmag Thanks for the detailed review ! About your last comment, I don't see how a `shell-enable-vterm-support` variable would make `vterm` more of an opt-in, beside adding an extra step to enable it and complicating the process. Already people would have to manually set the `shell-default-shell` to `vterm`

This PR is getting crowded, it is starting to be hard to get an overview 😅

layers/+tools/shell/packages.el
deb0ch: Removed the extra bindings ✨

duianto:
bmag knows what he's doing, I'm just randomly poking things with a stick to see if they poke back.

layers/+tools/shell/packages.el
deb0ch: Done

layers/+tools/shell/README.org
deb0ch: Done ✨

layers/+tools/shell/packages.el
deb0ch: We need the scroll bindings to be bound somehow, else they are not accessible from insert or hybrid state, which is inconvenient when using a terminal, and especially they are not accessible at all to non-evil users.

Maybe they could be bound to `C-u` and `C-d`, as it is a standard ? Like in `screen`, `less`, `vim` and other ?

layers/+tools/shell/packages.el
deb0ch: Don't remember, will look into it shortly. But now that you point it out it seems odd to me too to have put in `vterm` config this line which seems unrelated to `vterm` 🤔

deb0ch: Indeed it worked the same without this line, so I removed it for now. Why did I add this line ? 🤔 

I guess this is why I should have described _why_ I did it in the `set purpose` commit message. Now I just squashed this commit into the larger `add support for vterm` one, it seemed to belong there.

deb0ch: Oh, I think it was related to https://github.com/syl20bnr/spacemacs/issues/8171

deb0ch:
@bmag Hey, just finished taking into account all of your comments ! 👌 ✨ 

I just left out 2 of them, for which I expressed my concerns and waiting for your response 👂

layers/+tools/shell/packages.el
bmag: For scrolling there is C-v and M-v, but there is definitely something missing for non-Evil users. I noticed that <S-prior> (shift+page up), <S-next> (shift+page down), C-c C-p and C-c C-n aren't bound, and I couldn't guess any key binding for scrolling by one line. However, it is an upstream issue - adding bindings there will add them for all users, including Spacemacs users.

Besides, C-j and C-k are really useful in terminals (especially C-k), so overriding them is already questionable.

bmag:
@deb0ch the two concerns are about C-k/C-j and shell-enable-vterm-support, right?

The main reason I want `shell-enable-vterm-support` is because vterm's installation process has an increased risk of failing. I want to avoid the situation where we get many issue reports from users who failed to use vterm because they weren't aware of the external dependencies they need to install. I suppose we could omit the variable, and add it only if/after my hypothetical scenario really happens. Hmm, I think both approaches are ok, so I leave the decision to you.

I added a separate comment about C-k/C-j, but short answer is the original bindings are useful in a terminal and ultimately the issue should be solved upstream.

I'll review the rest of the changes later, but right now I'm on a computer where I excluded `centered-cursor-mode` package and got a "function's definition is void: centered-cursor-mode", so we need to add a `with-eval-after-load` in relevant places.

P.S. really is getting crouded here, the page loads noticeably slower than other PRs :stuck_out_tongue:

layers/+misc/multiple-cursors/config.el
```diff
@@ -12,4 +12,5 @@
 ;;; License: GPLv3
 
 (defvar multiple-cursors-backend 'evil-mc
-  "The package used as the backend for multiple cursors functionality.")
+  "The package used as the backend for multiple cursors functionality.
+Possible values are `mc' or `evil-mc'.")
```
bmag: Not relevant to this PR, probably added by mistake.

layers/+tools/shell/packages.el
```diff
@@ -28,11 +28,13 @@
         (term :location built-in)
         xterm-color
         vi-tilde-fringe
+        (vterm :toggle (not (eq window-system 'w32)))
```
bmag: Use `spacemacs/system-is-mswindows`, which consults `system-type` instead of `window-system`.

layers/+tools/shell/packages.el
deb0ch: Fair enough, I will put this line back how it was

layers/+tools/shell/packages.el
deb0ch: Should I also make the change where I found the example I based myself off of, in [shell/config.el](https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Btools/shell/config.el#L22) ?

layers/+misc/multiple-cursors/config.el
deb0ch: I added it while at it, I agree it is not related to this PR and adding a bit of noise but I didn't feel like opening a new PR just for that. I'll remove it if you prefer.

deb0ch:
@bmag I'm extremely glad that you are open to give it a try without the extra variable, thank you for that 😃 I think it is worth it, that we won't have many people opening issues and even in that case that they would be closable easily and promptly 😁

I will have an extra pass over all this to add `with-eval-after-load` everywhere relevant, your comment is precious as I wouldn't have thought of it myself.

layers/+tools/shell/packages.el
bmag: yes please

layers/+misc/multiple-cursors/config.el
bmag: Ok, in that case leave it here, no need for a new PR.

deb0ch:
@bmag I `eval-after-load`-ed everything I could that made sense, one strage thing to me is that I had absolutely no error and everything functional when I excluded `window-purpose` without `eval-after-load`-ing its configuration 🤔 

I still `eval-after-load`-ed it just in case  ¯\\\_(ツ)_/¯

layers/+tools/shell/packages.el
```diff
@@ -213,8 +218,7 @@
               (funcall 'man command))
              ;; Send other commands to the default handler.
              (t (comint-simple-send proc command))))))
-  (add-hook 'shell-mode-hook 'shell-comint-input-sender-hook)
-  (add-hook 'shell-mode-hook 'spacemacs/disable-hl-line-mode))
```
deb0ch: Ooops I don't think removing this was intentional 😅

layers/+tools/shell/README.org
```diff
@@ -250,3 +305,10 @@ in the current buffer instead of a popup.
 | Key binding        | Description                                       |
 |--------------------+---------------------------------------------------|
 | ~SPC m H~ or ~M-l~ | shell commands history using a helm or ivy buffer |
+
+** Vterm
```
deb0ch: I need to check that these bindings are still correct and sufficient

layers/+tools/shell/README.org
```diff
@@ -8,6 +8,15 @@
 - [[#description][Description]]
   - [[#features][Features:]]
```
deb0ch: Just noticed the typo here

Compro-Prasad:
I just tried this out. It works smoothly. I just added `vterm` in layer list and I was good to go by just doing `M-x vterm`.

bmag:
Ok, finally cherry-picked into develop. I edited one of your commits to replace `centered-cursor-mode` in the changelog with `=centered-cursor-mode=`, which is the correct formatting. The rest of the commit is untouched (so committer, dates, etc. have been kept).

The CI complains that the "Install CMake 3.11 or higher" heading doesn't have a body (i.e. instructions on how to install CMake). Left it empty for now, but some basic instructions can be added in a future PR.

Thanks everyone and especially @deb0ch.

deb0ch:
Yay ! :punch:

PR #11553, opened 2018-11-06
fix: lsp-ui-peek-find-definitions is an async jump handler

yetone:
To avoid multiple peek popup situation:

image


sdwolfz:
Thank you :heart:!

PR #11556, opened 2018-11-06
[TravisCI] Make sure that we deploy only from HEAD

JAremko:
Protecting from re-running old builds.

### DIFF
--- a/.travisci/pub_html.sh
+++ b/.travisci/pub_html.sh
@@ -30,6 +30,11 @@ else
     exit 0
 fi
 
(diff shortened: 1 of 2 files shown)


.travisci/pub_html.sh
```diff
@@ -30,6 +30,11 @@ else
     exit 0
 fi
 
+if [ `git rev-list HEAD...origin/$TRAVIS_BRANCH --count` != 0 ]; then
```
d12frosted: Oh, nice. I forgot about `--count` option. This is so helpful for one of my scripts.

sdwolfz: The `--left-right` flag is really useful as well.

PR #11558, opened 2018-11-06
Replace centered-buffer-mode with writeroom-mode

JAremko:
Seems like [`writeroom-mode`](https://github.com/joostkremers/writeroom-mode) covers it all.

Eventually we should add transient state for it.


syl20bnr:
Nice, remove quite a lot of code  👍

PR #11561, opened 2018-11-07
Update 'neo-vc-integration' example in DOCUMENTATION.org

wende:
The `neo-vc-integration` takes a list not a single element. The example is misleading

PS and disclaimer: Since it doesn't require any actual code change I base it on `master`. Let me know if that should be changed


sdwolfz:
@wende Yes you should open all PRs against the develop branch.

sdwolfz:
It's part of neotree's API so you should address the problem upstream if you really want to tidy it up.

PR #11572, opened 2018-11-09

emacspace:


sdwolfz:
@JAremko OK so I made it stop removing the `#+END_SRC` by also specifying the language, but it still insists on the weird indentation.

JAremko:
@sdwolfz  It means that org-mode parses doesn't recognize begin blocks - this is why everything is aligned this way.  But at the same time it recognizes end blocks - but they are unmatched so it throws them away. I think :thinking:  

Btw you can figure out what bot will do to your PR by running this https://github.com/JAremko/spacetools

JAremko:
I'll fix it. What  a fun little  :beetle:  :hugs:

PR #11586, opened 2018-11-12
ipython-notebook: remove spurious leader key

rocha:
There is an incorrect leader key being set in the ipython-notebook layer. The transient state leader key had already being define, and in the correct place.


rocha:
@CestDiego FYI

PR #10759, opened 2018-05-27
sql: Add default jump handlers for SQL layer

deuill:
Recent versions of `dumb-jump` have support for jumping to definitions for generic SQL buffers, and
support in Spacemacs requires that we enable these in the `sql` layer.

The full effect of this will be enabled once this PR makes it into `dumb-jump`: https://github.com/jacktasia/dumb-jump/pull/221

PR #11592, opened 2018-11-13
sql: Add default jump handlers for SQL layer

deuill:
Recent versions of `dumb-jump` have support for jumping to definitions for generic SQL buffers, and
support in Spacemacs requires that we enable these in the `sql` layer.

### DIFF
--- a/layers/+lang/sql/config.el
+++ b/layers/+lang/sql/config.el
@@ -9,6 +9,8 @@
 ;;
 ;;; License: GPLv3
 
+(spacemacs|define-jump-handlers sql-mode)
+
 (defvar sql-capitalize-keywords nil
   "Capitalize keywords in SQL mode.")
 


deuill:
This is a retry of #10759, which was supposedly pulled but never appeared on the `develop` branch.

deuill:
Is it possible this is looked at or closed? The change is still very relevant, but has languished for quite some time (May 2018 if we're going by the original PR)...

sdwolfz:
We're currently in a code freeze in order to release `0.300.0` and all code that is not a direct blocker for the release will not be merged in until it's over. Unfortunately I don't have an ETA for you as all my prior estimations were not accurate either.

deuill:
Thanks for the quick reply and the tag, just making sure it's on or off a queue. I know there's a lot on your plate right now!

duianto:

Issue #10293, opened 2018-02-03
pipenv directory .venv and pyvenv file .venv conflict


rgb-24bit:
#### Description :octocat:
In `Python`, an error occurs when there is a folder `.venv` in the root of my project.

#### Reproduction guide :beetle:
- Start Emacs
- There is a .venv folder and open a python script

*Observed behaviour:* :eyes: :broken_heart:
- error:
```elisp
File mode specification error: (file- error Opening input file Permission denied project/.venv)
```
- Can not be automatically completed

*Expected behaviour:* :heart: :smile:
- no error

#### System Info :computer:
- OS: windows-nt
- Emacs: 25.1.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. daed52b0)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
(graphviz octave yaml ivy auto-completion better-defaults emacs-lisp markdown org windows-scripts javascript sql html latex plantuml python
          (c-c++ :variables c-c++-enable-google-style t c-c++-default-mode-for-headers 'c-mode)
          my-csharp)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG SOUND NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS

#### Backtrace :paw_prints:
```
<<BACKTRACE IF RELEVANT>>
```

twillis:
this is going to be especially problematic since pipenv is apparently the official way to do these kinds of things.  :/

twillis:
fwiw redefining this function in your `dotspacemacs/user-config` like below seems to get around it.

```
(defun spacemacs//pyvenv-mode-set-local-virtualenv ()
    "Set pyvenv virtualenv from \".venv\" by looking in parent directories. handle directory or file"
    (interactive)
    (let ((root-path (locate-dominating-file default-directory
                       ".venv")))
      (when root-path
        (let* ((file-path (expand-file-name ".venv" root-path))
                (virtualenv
                  (if (file-directory-p file-path)
                    file-path
                    (with-temp-buffer
                      (insert-file-contents-literally file-path)
                      (buffer-substring-no-properties (line-beginning-position)
                        (line-end-position))))))
          (if (file-directory-p virtualenv)
            (pyvenv-activate virtualenv)
            (pyvenv-workon virtualenv))))))
```

svenpanne:
[...]

Issue #11145, opened 2018-08-08
Python-mode, Company not enabled when .venv in directory


saskenuba:
Today I started using `pipenv` with the whole virtualenv inside the project directory using `PIPENV_VENV_IN_PROJECT=true`, because previously I had some issues with pipenv emacs extension not finding my project. All this started after I upgraded my python to 3.7.

Now, every time I go into a file on the project, the following message appears in my `*Messages*` buffer:
`File mode specification error: (file-error Read error Is a directory /home/martin/Documentos/Programming/Python/Projetos/simple-flask-blog/.venv)`

This doesn't occur when accessing a standalone `.py` file, where company-mode gets initiated with the appropriate backends.

Any guess on what is happening?

rgb-24bit:
There is a compromise solution here. #10293

Issue #10798, opened 2018-06-03
`python-layer` does not recognise `.venv/` virtualenv subdirectory created by pipenv


skebanga:
#### Description :octocat:
`python-layer` does not recognise `.venv/` virtualenv subdirectory created by pipenv

#### Reproduction guide :beetle:
- set `PIPENV_VENV_IN_PROJECT=1` in environment
- create virtualenv using `pipenv`
- start Emacs
- open file in project

*Observed behaviour:* :eyes: :broken_heart:
virtualenv does not automatically activate
```
File mode specification error: (file-error Read error Is a directory /src/proj/.venv)
```

*Expected behaviour:* :heart: :smile:
Recognise that `.venv/` is a virtualenv subdirectory

#### System Info :computer:
- OS: gnu/linux
- Emacs: 25.2.2
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. b68bd33a1)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(helm emacs-lisp html javascript go lua octave csv git markdown neotree
      (shell :variables shell-default-height 30 shell-default-position 'bottom shell-default-shell 'eshell)
      imenu-list version-control
      (c-c++ :variables c-c++-enable-clang-support t c-c++-default-mode-for-headers 'c++-mode)
      (ranger :variables ranger-show-literal nil)
      python yaml lori-cpp
      (smart-tabs :variables smart-tabs-default-insinuations
                  '(c c++))
      protobuf)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GSETTINGS NOTIFY ACL LIBSELINUX GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11

#### Backtrace :paw_prints:
```
<<BACKTRACE IF RELEVANT>>
```

iafilatov:
This is because `.venv` is supposed to be a file that contains the path to your virtual env. This isn't configurable at the moment. As a workaround you can try using a different name for the dir, i.e. `_venv` or `.virtualenv`. Then put `/path/to/your/_venv` in `.venv`. Relative path also works.

skebanga:
pipenv doesn't support creating a `.venv` file, it only supports creating a `.venv/` directory.

iafilatov:
[...]

PR #11596, opened 2018-11-14
Fix conflict between pipenv directory .venv and pyvenv file .venv

rgb-24bit:
Using `Pipenv` might generate a directory named `.venv` in the project root, which conflicts with `pyvenv`'s `.venv` file.

This is the solution given by the code in #10293.

Issues that may be resolved: #10293, #11145,#10798


Trapez:
Now we can't use pipenv without setting PIPENV_VENV_IN_PROJECT. A project folder without a .venv folder doesn't get activated anymore.

PR #11597, opened 2018-11-14
Introduce key binding `sC` for `helm-swoop-clear-cache`

krizex:
helm-swoop cannot detect the buffer has changed when calling git checkout to
change the file content, so helm-swoop won't clear the cache, we have to clear
the cache manually by evaluating `helm-swoop--clear-cache`. As it is
inconvenient, so I introduce a new key binding to to this.

Signed-off-by: Yang Qian <yang.qian@citrix.com>

This is a workaround for the issue of helm-swoop: https://github.com/ShingoFukuyama/helm-swoop/issues/112

### DIFF
--- a/layers/+completion/helm/packages.el
+++ b/layers/+completion/helm/packages.el
@@ -615,8 +615,15 @@ Search for a search tool in the order provided by `dotspacemacs-search-tools'."
                      (if thing thing ""))))))
           (call-interactively 'helm-swoop)))
 
+      (defun spacemacs/helm-swoop-clear-cache ()
+        "Call `helm-swoop--clear-cache' to clear the cache"
+        (interactive)
+        (helm-swoop--clear-cache)
+        (message "helm-swoop cache cleaned."))
+
       (spacemacs/set-leader-keys
         "ss"    'helm-swoop
+        "sC"    'spacemacs/helm-swoop-clear-cache
         "sS"    'spacemacs/helm-swoop-region-or-symbol
         "s C-s" 'helm-multi-swoop-all)
       (defadvice helm-swoop (before add-evil-jump activate)


duianto:
Is it possible to add the clear cache key binding to the helm-swoop buffer?

It seems like the [helm-swoop](https://github.com/ShingoFukuyama/helm-swoop) author hasn't interacted with the repo for a year, so maybe it needs to be added in Spacemacs, possibly in the `helm/init-helm-swoop` function:
https://github.com/syl20bnr/spacemacs/blob/5d02b1c4bba2ce5b08c20731598d411c09a8acc1/layers/%2Bcompletion/helm/packages.el#L350

It should probably clear the cache and refresh (or reopen) the helm-swoop buffer.

Then it won't require three steps after opening the helm-swoop buffer and noticing the issue:
- Close the buffer
- Press the clear cache key binding
- Open the helm-swoop buffer again

Instead it would just require:
- Press the refresh/clear cache key binding

krizex:
@duianto If you want to clear the helm cache when you located in helm-swoop buffer, you need press some key combination to enter a new state which accept helm command. That means when you in a helm-swoop buffer you cannot clear the buffer by a <SPC> lead command because <SPC> will be typed into the buffer but not activate a new state.
It looks too complicated to add the new state. I'd prefer just add a simple key binding to clear the cache as is.

duianto:

duianto:
Now that I think of it, would it be better to change the key binding to `SPC s M-s`, then all swoop related commands would use `s` in the search/symbol section?
- `s` -> `helm-swoop`
- `S` -> `helm-swoop-region-or-symbol`
- `C-s` -> `helm-multi-swoop-all`

And should the clear cache key binding also open helm swoop?

The clear cache message can still be shown for a second or two before opening helm-swoop with:
`(sit-for <seconds>)`

krizex:
For <SPC> s s/S/C-s, all of them aim to trigger a search by helm-swoop. However, the <SPC> s C is not intend to open the search buffer but just clear the cache. I'd like to make the command simple and make the user explicitly knows what he is doing. If we change the key binding to `<SPC> s M-s`, it will induce the user to always search by this instead of `<SPC> s s` because this key binding always return the correct result, however it will be inefficient because we cannot take advantage of the cache.

PR #11606, opened 2018-11-15
Introduce 'defaults special value in dotspacemacs-line-numbers

Asgavar:
While I was struggling to make `display-line-numbers-mode` trigger in `tuareg-mode`, which does not derive from neither `prog-mode` nor `text-mode`, I realized that such a thing could be useful. Please let me know what you think.

### DIFF
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -1331,6 +1331,13 @@ Enable *line numbers* everywhere, even in non-prog-mode and non-text-mode buffer
   (setq-default dotspacemacs-line-numbers '(:enabled-for-modes 'all))
 #+END_SRC
 
+Enable *line numbers* in tuareg-mode, but keep the enabled-by-default ones:
+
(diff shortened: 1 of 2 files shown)


bmag:
You didn't actually explain why this is useful, or what problem this PR solves. Have you tried setting `dotspacemacs-line-numbers` to:
```elisp
(:enabled-for-modes prog-mode text-mode tuareg-mode)
```
Maybe we should add a special `defaults` value, so you could set it to:
```elisp
(:enabled-for-modes defaults tuareg-mode)
```

Asgavar:
Well, that's not really a _problem_, since what you suggested achieves exactly the same thing, I just find it neater not having to re-specify the defaults, event if they are not subject to change in the foreseeable future.

But yeah, `defaults` as a special value seems more intuitive than my cryptic `force-something`, I'll send a patch with this tomorrow.

Asgavar:
Done!

PR #11610, opened 2018-11-18
git: Fix doc on magit-repository-directories

panchoh:
The layer doc implied that the value of `magit-repository-directories` should be
a directory, when in fact it should be a *list* of directories.  Also, each
entry should specify the maximum depth to be scanned for git repos.  See the
official documentation¹ for more details.

¹: https://magit.vc/manual/magit.html#Status-Buffer

Apropos, the link to that documentation was not accurate, so I've fixed it as
well.


panchoh:
I've closed it because PR #11353  was already in the queue with a fix for it.

PR #11618, opened 2018-11-20
Move lexical-binding to first line

duianto:
problem:
Some files write the lexical binding to the right of the first line, that makes
it seem like it's just part of the layer description.

https://www.emacswiki.org/emacs/LexicalBinding states that:
> enable lexical binding in your elisp library, put a line like this at the top
> of the file:
> ;; -*- lexical-binding: t -*-

which seems to indicate that it should be written by itself.

And possibly the newest layer multiple-cursors writes the lexical binding on the
first line.

solution:
Move the lexical-binding to the first line.

---

#### Semicolon
Three files:
```
core/core-debug.el
core/libs/load-env-vars.el
layers/+lang/java/local/flycheck-eclim/flycheck-eclim.el
```
had a semicolon after `t` like this:
```
;; -*- lexical-binding: t; -*-
```
but it seemed like a mistake so I removed the semicolons.

#### React layer
The react layer enabled the lexical-binding in 3 out of the 4 files:
```
config.el
funcs.el
packages.el
```
I added it to the 4th file `layers.el` as well, for consistency.

### DIFF
--- a/core/core-debug.el
+++ b/core/core-debug.el
@@ -1,4 +1,6 @@
-;;; core-debug.el --- Spacemacs Core File  -*- lexical-binding: t; -*-
+;; -*- lexical-binding: t -*-
+;;
+;;; core-debug.el --- Spacemacs Core File
 ;;
 ;; Copyright (c) 2012-2018 Sylvain Benner & Contributors
 ;;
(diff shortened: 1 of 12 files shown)


duianto:
Nevermind, it seems like it isn't unusual to put it on the first line, for example Magit does it. Therefore I'll close this.

PR #11619, opened 2018-11-20
Unify packages list parentheses style

duianto:
problem:
some layer packages lists have the open and closing parentheses on the same line
as the first and last listed package, but most seem to have them on a separate
lines.

solution:
put the open and close parentheses on separate lines, except for lists with only
a single package, they are written on the same line as the variable name and
parentheses.

fix the lists indentation


sdwolfz:
Thank you :heart:!

Issue #10174, opened 2018-01-17
Dvorak layout is resetting navigation key, when entered to normal mode from insert mode.


zer09:
#### Description :octocat:
The ```s``` will reset its binding to ```substitute``` command, on dvorak
layout, while editing the ```.spacemacs``` config file.

Now you will lose the ```navigate right key```, and the ```s``` and ```l``` key
is now binded to ```substitute``` command.

Possibly the issue is introduced by this commit 2cb9844.

#### Edit:
It look like that not only the config file is affected but the ```Emacs lisp``` lang.

#### Reproduction guide :beetle:
- Start Emacs using the dvorak layout.
```
(keyboard-layout :variables
                      kl-layout 'dvorak)
```
- Open your config file ```SPC f e d```
- Try to navigate to check if it's working using ```h t n s```
- Enter to insert mode.
- Returning back to normal mode, now the ```s``` navigation is now binded to
```substitute```

*Observed behaviour:* :eyes: :broken_heart:
The navigation key ```s``` is now binded to ```substitute``` command.

*Expected behaviour:* :heart: :smile:
The navigation key ```s``` will should navigate you to the right direction.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 25.3.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 0f646b88)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
((fasd)
 (imenu-list)
 (spell-checking :variables spell-checking-enable-by-default nil)
 (syntax-checking)
 (auto-completion :variables auto-completion-return-key-behavior 'complete auto-completion-tab-key-behavior 'cycle auto-completion-enable-snippets-in-popup t auto-completion-enable-help-tooltip t auto-completion-enable-sort-by-usage t)
 (ycmd)
 (rebox :variables rebox-enable-in-text-mode t)
 (ranger :variables ranger-enter-with-minus t ranger-show-hidden t ranger-max-preview-size 1)
 (colors)
 (c-c++ :variables c-c++-enable-clang-support t c-c++-enable-rtags-support t)
 (theming)
 (keyboard-layout :variables kl-layout 'dvorak)
 (git)
 (version-control :variables version-control-diff-tool 'git-gutter+ version-control-diff-side 'left)
 (systemd)
 (shell :variables shell-default-shell 'eshell)
 (protobuf)
[...]

PR #11625, opened 2018-11-22
keyboard-layout: add evil-cleverparens remapping

CharlesHD:
When you use structurally safe editing, cleverparens is called and provides a different keymap that isn't remapped by keyboard-layout machinery. This problem is stated in #10174 . 

This PR opens the remapping for evil-cleverparens. As I do not use it myself, I think I didn't remap many of important movement bindings. Currently it provides a basic support for HJKL. 

Ideas of keys remapping are welcomed.


smile13241324:

PR #11626, opened 2018-11-22
Core lsp minor mode keybindings / migration to lsp.el

cormacc:
Replaced opt-in use of `spacemacs/lsp-bind-keys-for-mode` function with
keybindings for the lsp minor  mode, which should be enabled for all derived
layers.

Updated to reflect upstream changes in lsp-mode package (lsp.el). Much of the configuration now done upstream, so lsp layer simplified.

c-c++ layer updated to reflect changes. Updates to other layers from yyoncho pending.

### DIFF
--- a/layers/+lang/c-c++/config.el
+++ b/layers/+lang/c-c++/config.el
@@ -11,6 +11,9 @@
 
 ;; variables
 
+(spacemacs|define-jump-handlers c++-mode)
+(spacemacs|define-jump-handlers c-mode)
+
 (defconst c-c++-modes '(c-mode c++-mode)
   "Primary major modes of the `c-c++' layer.")
 
@@ -45,9 +48,6 @@
   "If non-nil, automatically format code with ClangFormat on
   save. Clang support has to be enabled for this to work.")
 
-(spacemacs|define-jump-handlers c++-mode)
-(spacemacs|define-jump-handlers c-mode)
-
 (defvar c-c++-default-mode-for-headers 'c-mode
   "Default mode to open header files. Can be `c-mode' or `c++-mode'.")
 
@@ -77,17 +77,10 @@ if no pattern matches the project root, lsp-c-c++ will be initialized.")
   "When non-nil, use rainbow semantic highlighting")
 
 (defvar c-c++-lsp-initialization-options nil
-  "Extra initialisation parameters to pass to the ccls backend. See
+  "Extra initialisation parameters to pass to the lsp backend. See
 https://github.com/MaskRay/ccls/blob/master/src/config.hh
-for details. N.B. cquery still uses the deprecated extra-init-params (below)")
+for details. N.B. This is remapped to cquery-extra-init-params when using cquery backend")
 
-;; I've left cquery/ccls -extra-init-params separate for now, as one has defaults while the other doesn't
-;; Just to facilitate switching between the two easily
-(defvar c-c++-lsp-extra-init-params '(:cacheFormat "msgpack")
-  "Extra initialisation parameters to pass to the backend. See
-https://github.com/cquery-project/cquery/blob/master/src/config.h or
-for details.")
-
-(defvar c-c++-lsp-extra-args nil
+(defvar c-c++-lsp-args nil
   "Extra args to pass to the backend. E.g. to log to file.
-https://github.com/MaskRay/ccls/wiki/Emacs for details")
+https://github.com/MaskRay/ccls/wiki/Emacs for details. N.B. this is remapped to cquery-extra-args when using cquery backend")
--- a/layers/+lang/c-c++/config.el
+++ b/layers/+lang/c-c++/config.el
@@ -76,10 +76,18 @@ if no pattern matches the project root, lsp-c-c++ will be initialized.")
 (defvar c-c++-lsp-sem-highlight-rainbow nil
   "When non-nil, use rainbow semantic highlighting")
 
+(defvar c-c++-lsp-initialization-options nil
+  "Extra initialisation parameters to pass to the ccls backend. See
+https://github.com/MaskRay/ccls/blob/master/src/config.hh
+for details. N.B. cquery still uses the deprecated extra-init-params (below)")
+
 ;; I've left cquery/ccls -extra-init-params separate for now, as one has defaults while the other doesn't
 ;; Just to facilitate switching between the two easily
 (defvar c-c++-lsp-extra-init-params '(:cacheFormat "msgpack")
   "Extra initialisation parameters to pass to the backend. See
 https://github.com/cquery-project/cquery/blob/master/src/config.h or
-https://github.com/MaskRay/ccls/blob/master/src/config.h
 for details.")
+
+(defvar c-c++-lsp-extra-args nil
+  "Extra args to pass to the backend. E.g. to log to file.
+https://github.com/MaskRay/ccls/wiki/Emacs for details")
--- a/layers/+lang/c-c++/funcs.el
+++ b/layers/+lang/c-c++/funcs.el
@@ -225,28 +225,27 @@ and the arguments for flyckeck-clang based on a project-specific text file."
   (spacemacs//c-c++-lsp-set-config (intern (concat "c-c++-lsp-" suffix)) nil (concat "-" suffix)))
 ;; -- END helper functions for common configuration of cquery and ccls backends
 
-
-(defun spacemacs//c-c++-lsp-enable ()
-  "Enable the LSP backend specified by the `c-c++-backend' configuration variable."
-    (progn (condition-case nil
-             (spacemacs//c-c++-lsp-call-function "lsp-" "-enable")
-             (user-error nil))))
-
 (defun spacemacs//c-c++-lsp-config ()
   "Configure the LSP backend specified by the `c-c++-backend' configuration variable."
     (progn
-      (spacemacs//c-c++-lsp-setup-company)
       (spacemacs//c-c++-lsp-define-extensions)
       (spacemacs//c-c++-lsp-wrap-functions)
       (setq-default flycheck-disabled-checkers '(c/c++-clang c/c++-gcc))
+
+      (dolist (param '("executable" "extra-init-params" "initialization-options" "extra-args" "project-whitelist" "project-blacklist" "sem-highlight-method"))
+        (spacemacs//c-c++-lsp-apply-config param))
 
       (if (eq c-c++-lsp-cache-dir nil)
         (progn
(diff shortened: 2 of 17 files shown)


cormacc:
The minor mode keybindings work fine. Can't figure out how to declare the prefixes though, without an explicit call per-mode to `(spacemacs/lsp-declare-prefixes-for-mode mode)`

cormacc:
@yyoncho this also reverts my rebasing error with your modelighter change

yyoncho:
@cormacc thanks.

Did you try to add the prefixes adding hook on `lsp-mode-hook` and using current `major-mode`?

cormacc:
@yyoncho @sdwolfz If going this automatic route, should we also be calling `spacemacs//setup-lsp-jump-handler` in the `lsp-after-open-hook` along with the call to `spacemacs//lsp-declare-prefixes-for-mode`, rather then requiring each lsp language layer to call it explicitly?

yyoncho:
IMO having as much as possible in `lsp` layer is better since it will remove the copy/paste from the language specific layers. When single entry point is implemented in `lsp-mode` we will be able to remove the call to `lsp-{language}-enable`.

cormacc:
@yyoncho I've amended the commit. Stripped out a fair bit of config in the lsp layer that now seems to be handled automagically by lsp.el -- great stuff. You might take a look when you get the chance to make sure I haven't done anything daft.
Also, disabled deferred loading of the cquery/ccls packages, as not sure how to to go about it using the new mechanism (where we're adding 'lsp to the mode hook rather than 'lsp-<cquery/ccls>-enable). Any ideas there?

yyoncho:
Something like ":defer t :after lsp"? You did a good job, no coments from my side. I will send a pr against your branch.

layers/+lang/c-c++/funcs.el
```diff
@@ -312,10 +303,9 @@ and the arguments for flyckeck-clang based on a project-specific text file."
     "hb" 'base) ;;Replace this with lsp-goto-implementation in lsp-layer?
 
   (when (eq c-c++-backend 'lsp-ccls)
-    (spacemacs/set-leader-keys-for-major-mode mode
-      "bR" 'ccls-reload)
+    ;; (spacemacs/set-leader-keys-for-major-mode mode
     (spacemacs/lsp-bind-extensions-for-mode mode "c-c++"
-      "mc" 'member-classes
+      "mc" 'member-types
```
gessen: Perhaps it should be "mt" to match its name?

```suggestion
      "mt" 'member-types
```

layers/+lang/c-c++/funcs.el
```diff
@@ -225,28 +225,27 @@ and the arguments for flyckeck-clang based on a project-specific text file."
...
           (setq c-c++-lsp-cache-dir (file-truename(concat "~/.emacs.d/.cache/" (symbol-name c-c++-backend))))
           (message (concat "c-c++: No c-c++-lsp-cache-dir specified: defaulting to " c-c++-lsp-cache-dir))))
 
-      (dolist (param '("executable" "extra-init-params" "cache-dir" "project-whitelist" "project-blacklist" "sem-highlight-method"))
-        (spacemacs//c-c++-lsp-apply-config param))
+      (ecase c-c++-backend
+        ('lsp-cquery (setq cquery-cache-dir c-c++-lsp-cache-dir))
+        ('lsp-ccls (setq ccls-initialization-options
+                     (if ccls-initialization-options
+                       (cons ccls-initialization-options `(:cacheDirectory ,c-c++-lsp-cache-dir))
```
gessen: I'm not sure `cons` is a good choice here. If ccls-initialization-options is a list like `(:a "1" :b "2")` then it would create something like `((:a "1 :b "2") :cacheDirectory "<path>")` which isn't correct

```suggestion
                       (append ccls-initialization-options `(:cacheDirectory ,c-c++-lsp-cache-dir))
```

layers/+lang/c-c++/funcs.el
```diff
@@ -312,10 +303,9 @@ and the arguments for flyckeck-clang based on a project-specific text file."
     "hb" 'base) ;;Replace this with lsp-goto-implementation in lsp-layer?
 
   (when (eq c-c++-backend 'lsp-ccls)
-    (spacemacs/set-leader-keys-for-major-mode mode
-      "bR" 'ccls-reload)
+    ;; (spacemacs/set-leader-keys-for-major-mode mode
     (spacemacs/lsp-bind-extensions-for-mode mode "c-c++"
-      "mc" 'member-classes
```
gessen: Also, you could add ccls-specific `"hd" 'derived` to match `"hb" 'base` function. Just need to add `(spacemacs/lsp-define-extensions "c-c++" 'derived "$ccls/inheritance" '(:derived t))` where custom extensions are defined

layers/+lang/c-c++/funcs.el
```diff
@@ -225,28 +225,27 @@ and the arguments for flyckeck-clang based on a project-specific text file."
...
-
 (defun spacemacs//c-c++-lsp-config ()
   "Configure the LSP backend specified by the `c-c++-backend' configuration variable."
     (progn
-      (spacemacs//c-c++-lsp-setup-company)
       (spacemacs//c-c++-lsp-define-extensions)
       (spacemacs//c-c++-lsp-wrap-functions)
       (setq-default flycheck-disabled-checkers '(c/c++-clang c/c++-gcc))
 
+      (dolist (param '("executable" "extra-init-params" "project-whitelist" "project-blacklist" "sem-highlight-method"))
```
gessen: I think two additional variables should be mapped from lsp to ccls/cquery here. Firstly, `initialization-options` to mirror `extra-init-params` which ccls doesn't use anymore. Secondly, `extra-args` to allow, for instance, adding logging from spacemacs configuration instead of creating a wrapper script that would add additional arguments. If you would add them though, they need to be defined in config.el

```suggestion
      (dolist (param '("executable" "extra-args" "extra-init-params" "initialization-options" "project-whitelist" "project-blacklist" "sem-highlight-method"))
```

layers/+lang/c-c++/funcs.el
cormacc: Corrected -- I need to learn a bit of lisp properly at some point :)

layers/+lang/c-c++/funcs.el
cormacc: Also done

gessen:
👍

MaskRay:
The old interface of lsp-mode has been deleted. lsp in spacemacs is broken at the point: `M-x lsp-*-enable` will no longer work. It will be good to prioritize merging this PR.

sdwolfz:
@cormacc is there any more work you need to do for this PR? Otherwise I'll jump in and merge it.

yyoncho:
@sdwolfz  we need to port java/javascript/python I am updating them now.

yyoncho:
@cormacc can you include this CL as well - https://github.com/yyoncho/spacemacs/commit/53734b8357a3a50277b0586ba0ef008d6653d12b

layers/+lang/c-c++/config.el
```diff
@@ -76,10 +76,18 @@ if no pattern matches the project root, lsp-c-c++ will be initialized.")
...
+
 ;; I've left cquery/ccls -extra-init-params separate for now, as one has defaults while the other doesn't
 ;; Just to facilitate switching between the two easily
 (defvar c-c++-lsp-extra-init-params '(:cacheFormat "msgpack")
   "Extra initialisation parameters to pass to the backend. See
 https://github.com/cquery-project/cquery/blob/master/src/config.h or
-https://github.com/MaskRay/ccls/blob/master/src/config.h
 for details.")
+
+(defvar c-c++-lsp-extra-args nil
```
gessen: Add c-c++-lsp-args

cormacc:
@gessen - resolved. All these different names for similar configuration parameters seem to me to be getting out of hand, so I've flattened them a bit (see excerpt from commit message below).
@sdwolfz Didn't amend the commit this time as that would have been an amendment to @yyoncho's additions, which are in unrelated layers. This means this PR comprises 3 commits -- apologies, but reckoned 3 reasonably discrete commits was preferable to 2 less coherent? I'm away on a work trip  for the next couple of days, so unlikely to get time to make any more changes myself til the weekend. Hopefully there won't be any more needed 

c-c++-lsp-initialization-options:
-> ccls-initialization-options
-> cquery-extra-init-params (cacheFormat set to "msgpack" by default)

c-c++-lsp-args:
-> ccls-args
-> cquery-extra-args

c-c++-lsp-cache-dir:
-> ccls: Appended to initialization-options
-> cquery-cache-dir

sdwolfz:
@cormacc It's OK. I'll take a look as soon as I have some time and if it's working well enough I'll merge it.

layers/+tools/lsp/packages.el
```diff
@@ -13,48 +13,29 @@
...
     (progn
-      (spacemacs|hide-lighter lsp-mode)
+      (require 'lsp-clients)
+      (setq lsp-prefer-flymake nil)
+      (spacemacs/lsp-bind-keys)
+      (add-hook 'lsp-after-open-hook (lambda ()
+                                       "Setup xref jump handler and declare keybinding prefixes"
+                                       (spacemacs//setup-lsp-jump-handler major-mode)
+                                       (spacemacs//lsp-declare-prefixes-for-mode major-mode)))
       (evil-set-command-property 'lsp-goto-type-definition :jump t)
```
MaskRay: `evil-set-command-property` has been added to lsp-mode.

cormacc: Thanks @MaskRay -- eliminated the duplication from packages.el

yyoncho:
@beyondpie

layers/+lang/c-c++/packages.el
```diff
@@ -259,28 +259,23 @@
       (spacemacs/set-leader-keys-for-major-mode mode "gi" 'cscope-index-files))))
 
 ;; BEGIN LSP BACKEND PACKAGES
-
 ;; See also https://github.com/cquery-project/cquery/wiki/Emacs
 (defun c-c++/init-cquery ()
   (use-package cquery
     :if (eq c-c++-backend 'lsp-cquery)
-    :defer t
-    :commands lsp-cquery-enable
```
beyondpie: @yyoncho 
Thanks for your remind. I also change my local spacemacs  config here to lsp, add (require 'cquery) in my user config. I plan to submit a PR at the moment, but feel quite happy that @MaskRay correct it so soon.

MaskRay:
@sdwolfz I believe this patch is in a good shape. Merging it addresses much hassle.

sdwolfz:
Thank you :heart:!

PR #11636, opened 2018-11-25
Update info-find-file to add third argument found in emacs 27

emacs18:
Emacs 27 (not yet release version found on master branch of emacs repo)
 added 3rd optional argument to Info-find-file.
However Info-find-file found in info+ local package has only two.
This causes failures in how I use info mode.
Given that the new argument is benign, and Info-find-file is pretty much identical
to Info-find-file found in emacs, this change should be very safe change which
makes the package work for emacs 27 as well as older emacs.

Should I try to get this patch applied to Drew's (the author of info+) version first, 
then try to have that version propagated to spacemacs?


emacs18:
Hmm.  I must have done something wrong with this PR.
I only made few lines of change in one file, i.e., info+.el.
Let me close this one and see if I can submit a new PR.

PR #11637, opened 2018-11-25
ported Info-find-file from emacs 27 by adding third argument

emacs18:
Emacs 27 (not yet release version found on master branch of emacs repo)
 added 3rd optional argument to Info-find-file.
However Info-find-file found in info+ local package has only two.
This causes failures in how I use info mode.
Given that the new argument is benign, and Info-find-file is pretty much identical
to Info-find-file found in emacs, this change should be very safe change which
makes the package work for emacs 27 as well as older emacs.

Should I try to get this patch applied to Drew's (the author of info+) version first, 
then try to have that version propagated to spacemacs?


sdwolfz:
@syl20bnr you might be able to advise here.

sdwolfz:
Actually, I'm going to look into this and merge it as part of the next release...

sdwolfz:
Thank you :heart:!
Commit: a46b784e85911afe8bd6648ca55c080ffae1081c

PR #11643, opened 2018-11-26
Fix sort in insert-lazy-install-configuration

Miciah:
Commit 1c4f685b13ab606f7ad71299547d22cf3eec6cc2 replaced `configuration-layer--layers` (a list of `cfgl-layer` objects) with `configuration-layer--used-layers` (a list of symbols).  It also changed the `configuration-layer/insert-lazy-install-configuration` function to use the new variable, but it did not change the sort predicate.  Consequently, the function started failing:

    Wrong type argument: listp, spacemacs-purpose

In addition, because `sort` is destructive, the function also deleted values from `configuration-layer--used-layers`.

 This commit fixes the sort predicate so that the function returns the correct value, and it copies the list before sorting to avoid modifying `configuration-layer--used-layers`.

* `core/core-configuration-layer.el` (`configuration-layer/insert-lazy-install-configuration`): Use `string<` as the sort predicate, and give `sort` a copy of the list.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -489,6 +489,8 @@ Other:
     to Sylvain Benner)
   - New function =configuration-layer/load-file= to load files silently (thanks
     to Sylvain Benner)
+  - Fix configuration-layer/insert-lazy-install-configuration, which has been
+    broken since v0.200.0.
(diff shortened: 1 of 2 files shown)


Miciah:
Yikes! I even called out in the commit message that `sort` was destructive.  Thanks! I have updated the PR.

smile13241324:

PR #11648, opened 2018-11-26
Set default shell on Pylookup makefile

### LINKED ISSUE #11647: Pylookup makefile breaks when 'fish' is the default shell
piping is different on 'fish', so when `$SHELL` is set to `fish` on this makefile, it breaks.

https://github.com/syl20bnr/spacemacs/blob/fc2d220b5eab09bfe1ee5cd61a777091f520af08/layers/%2Blang/python/local/pylookup/makefile#L1

korayal:
This should fix #11647


smile13241324:

PR #11658, opened 2018-11-28
ivy: remove evil and helm-make from ivy layer

robert-zaremba:
The `evil` and `helm-make` are not necessary for `ivy` layer, especially if someone is using `emacs` style. This dependencies (evil and helm-make) are included anyway if one is using evil and helm layers.

### DIFF
--- a/layers/+completion/ivy/packages.el
+++ b/layers/+completion/ivy/packages.el
@@ -15,7 +15,9 @@
         bookmark
         counsel
         counsel-projectile
+	(evil :toggle (dotspacemacs-editing-style 'evil))
         flx
+	(helm :toggle (configuration-layer/package-used-p 'helm-make))
         imenu
         ivy
         ivy-hydra
@@ -26,7 +28,7 @@
         persp-mode
         projectile
         recentf
-        smex
+        (smex :toggle (configuration-layer/package-used-p 'smex))
         swiper
         wgrep
         ))
--- a/layers/+completion/ivy/packages.el
+++ b/layers/+completion/ivy/packages.el
@@ -15,9 +15,7 @@
         bookmark
         counsel
         counsel-projectile
-        evil
         flx
-        helm-make
         imenu
         ivy
         ivy-hydra


yuhan0:
This would cause the post-init configuration functions defined below not to be executed at all - might be better to use a `:toggle` keyword to check for dotspacemacs-editing-style and the helm layer. 
It also looks like `smex` should be removed from this layer's packages? Ivy doesn't depend on it or vice versa, and there's already a separate smex layer for Spacemacs.

robert-zaremba:
Hi @yuhan0 , thanks for reply. Do we need toggle for `smex` and `help`? If they will need to be required in other layers, than what's the sense of adding it here? I'm just try to better understand the logic.
Anyway, I've added a toggle. Could you please check if everything is correct?

yuhan0:
Hey thanks - I've actually never used helm-make before, but it looks like it actually runs fine just with ivy as a backend?  
It might be better to just leave that alone or have someone who does use it chime in here.. If the issue is helm being installed as a dependency, I *believe* helm doesn't get loaded on startup so it won't affect load times, and you can always exclude it in `dotspacemacs-excluded-packages`.  

As for smex I don't see any reason it should be included in the ivy layer at all, maybe it was an oversight?  

I'm not a maintainer here btw, but they seem to be quite busy recently so I thought I'd help with a few suggestions 😅

robert-zaremba:
No idea. I'm not using `helm` so I would prefer to stream down the big dependencies (like helm).

yuhan0:
My bad, I just found out that `counsel-M-x` requires smex in order to sort commands in order of history - perhaps some documentation of this would be good.

robert-zaremba:
Right, I reverted the smex :toggle change

robert-zaremba:
Any chance to merge this PR?

Compro-Prasad:
@sdwolfz This seems like a well reviewed PR by @yuhan0 . Thus, it would be great to have it merged.

smile13241324:
I have checked your PR but when I read the package.el the removed packages are used for global bindings which I would be very careful to remove.

* we do not want to load different packages depending on the input mode to keep the entire system manageable therefore loading evil only for vim style would be against the advice of @syl20bnr.
* helm-make is used for make integrations but is using `ivy` as frontend so this is useful in an `ivy` context and should not be removed.
* smex has been moved into a separate layer `smex` which owns the package and adds the respective keybindings. The package should not longer be initialised in the ivy layer. If smex is used by one of the other packages a requirement on the smex layer must be declared or the package should only be loaded if smex layer is available.

robert-zaremba:
Right. Thanks for checking. I'm going to remove this branch.

PR #10927, opened 2018-06-20
WIP: Add new layer flow-type

sei40kr:

PR #11659, opened 2018-11-28
add lsp-javascript-flow to javascript-backends

robert-zaremba:
Adding LSP JavaScript Flow backend to the `javascript` layer.
This is highly anticipated feature and was requested many time.

### DIFF
--- a/layers/+lang/javascript/README.org
+++ b/layers/+lang/javascript/README.org
@@ -12,13 +12,15 @@
   - [[#choosing-a-formatter][Choosing a formatter]]
 - [[#backends][Backends]]
   - [[#tern][Tern]]
-  - [[#language-server-protocol][Language Server Protocol]]
+  - [[#language-server-protocol-typescript][Language Server Protocol: typescript]]
+  - [[#language-server-protocol-flow][Language Server Protocol: flow]]
 - [[#configuration][Configuration]]
   - [[#indentation][Indentation]]
   - [[#repl][REPL]]
   - [[#node-modules][Node Modules]]
 - [[#key-bindings][Key Bindings]]
   - [[#js2-mode][js2-mode]]
+  - [[#lsp-flow][lsp-flow]]
   - [[#folding-js2-mode][Folding (js2-mode)]]
   - [[#refactoring-js2-refactor][Refactoring (js2-refactor)]]
     - [[#documentation-js-doc][Documentation (js-doc)]]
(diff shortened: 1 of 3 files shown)


sei40kr:
https://github.com/syl20bnr/spacemacs/pull/10927/files
I was working on Flow support (and I've completely forgot about that).
As written in the PR, Flow support has many problems to solve. Most difficult one is syntax highlighting.
`flow-js2-mode` works but it's not perfect.

sei40kr:
I hope my old PR helps you a lot. There implements more Flow features.
I did not use LSP for Flow, because `lsp-mode` uses a same server in same major mode.

robert-zaremba:
Hi @sei40kr 
The solution I propose here works out of the box: syntax highlighting, error checking, type hints, jump-to-definition ...
Not sure what do you mean with: "lsp-mode uses a same server in same major mode." ?

yyoncho:
This pr is no longer relevant with the latest lsp mode since it has several breaking changes. Take a look at #11626

sei40kr:
@robert-zaremba 
Sorry for my poor English.
AFAIK, flow-minor-mode doesn't provide syntax highlighting.
And flow-bin overlaps flow-language-server, so it shouldn't be invoked together.

robert-zaremba:
@yyoncho  Thanks for the link. AFAIS your PR doesn't touch the JavaScript layer. 

I've rebased my updates on the latest develop (which include your changes). The `lsp-ui` stopped working, but all other functionality works (flow, type-checks, jumps, etc...). Could you advise what should I change to make it working again? Also - do you think I should rebind the `C-c C-j` to `M-.`?

Compro-Prasad:
> Also - do you think I should rebind the `C-c C-j` to `M-.`?

Yes, `M-.` is the standard keybinding in Emacs for `jump-to-definition` related stuff. But in most of jumps `M-,` helps in jumping back to the previous location. So, if `C-c C-j` makes use of the marker stack in Emacs then `M-.` is a good switch but if it doesn't then try defining a function that does the same.

smile13241324:
@robert-zaremba @yyoncho  can you tell is this PR still relevant for the javascript layer, especially for LSP support? From the last post of yyoncho I read that the changes would be obsolete compared with current javascript layer can you describe a bit more what your PR does and how it fits in with the javascript backend architecture applied in the javascript layer as of today?

If it is still relevant, and fits in with the existing backends in the layer aka not breaking anything, please rebase to develop and fix the conflicts. I will then happily review and merge it.

yyoncho:
@smile13241324 emacs-lsp/lsp-mode#603 was later reverted so I think that we will need some code in spacemacs but I think that the current approach is wrong. AFAIK you should be able to use typescript/flow-language server in one project at the same time and the server will be selected from lsp-mode based on the presence of the flow header. So, I think that there should be only one backend - `lsp` and spacemacs should let lsp-mode select the correct language server.

robert-zaremba:
I'm using this patch and it works well. Let me test tomorrow if it's needed or not.

robert-zaremba:
I've checked the latest version.
Indeed, `lsp` supports the type exploration. However the following features are missing:

* using `flow` for jumps (`*goto*`)
* flow syntax handling (eg through `flow-js2-mode`)

I'm going to decline this PR. Currently I'm using the following configuration to support `flow` in spacemacs:

```
;; in dotspacemacs-configuration-layers
(javascript :variables javascript-backend 'lsp )

;; in dotspacemacs/user-config
  (add-hook 'js2-mode-hook
            (lambda ()
              ;(flow-js2-mode)
              (define-key js2-mode-map (kbd "C-c C-j") #'flow-minor-jump-to-definition)))
```

I've added `flow-js2-mode` to `dotspacemacs-additional-packages`.

One idea would be to add `flow-js2-mode` to the `javascript` layer end enable it by default. What do you think?

PR #11662, opened 2018-11-29
Mu4e add 'gu' key binding to go to urls.

dschrempf:
This pull request adds a binding `gu` in `mue-view-mode` to follow URLs. This is usually bound to `g` in mu4e, but evilification removes the original map.


smile13241324:

PR #11663, opened 2018-11-29
Add symbol-overlay package to spacemacs/navigation layer

kenkangxgwe:
Defines a transient state for symbol-overlay, and bind it to <kbd>SPC s o</kbd> and <kbd>SPC s O</kbd>. See [symbol-overlay](https://github.com/wolray/symbol-overlay). This package allows multiple symbols highlighted by using the built-in overlay function.

### DIFF
--- a/layers/+spacemacs/spacemacs-navigation/packages.el
+++ b/layers/+spacemacs/spacemacs-navigation/packages.el
@@ -23,6 +23,7 @@
         paradox
         restart-emacs
         (smooth-scrolling :location built-in)
+        symbol-overlay
         winum))
 
 (defun spacemacs-navigation/init-ace-link ()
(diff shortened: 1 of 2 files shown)


duianto:
#### Issue:

Example buffer:
```
abc
def
abc
def
abc
def
```

With cursor on the first line `abc`
Pressing `SPC s o` does add overlay colors to each instance of `abc`,
but it doesn't open a Transient state and the messages buffer shows:
>abc: 1/3 (symbol-overlay-face-3)
let*: Wrong type argument: number-or-marker-p, "\\_<abc\\_>"

Calling: `M-x` `symbol-overlay-put` `RET` seems to work as expected, it adds the overlay color and only shows:
>abc: 1/3 (symbol-overlay-face-2)

And removing the overlays `SPC s O` also seem to work as expected.

#### System Info :computer:
- OS: windows-nt
- Emacs: 26.2
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 03cbc6c51) (This PR was applied to 3d11456f5115d970b334906870581711a4535760)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(autohotkey emacs-lisp git helm multiple-cursors org spell-checking treemacs version-control)
```

layers/+spacemacs/spacemacs-navigation/packages.el
```diff
@@ -369,6 +370,48 @@
     :documentation "Smooth scrolling."
     :evil-leader "tv"))
 
+(defun spacemacs-navigation/init-symbol-overlay ()
+  (use-package symbol-overlay
+    :init
+    (progn
+      (setq spacemacs--symbol-overlay-transient-state-doc "
```
duianto: To align both the transient state string and how it appears when the TS is opened, two caret characters (`^^`) need to be added for every additional key that's missing, to match the entry with the most keys in a column.

For example the current second column:
```
[_N_/_p_]
[_o_]
[_r_]
```
should look like this:
```
[_N_/_p_]
[_o_]^^
[_r_]^^
```

Since `q` is bound to quit, it should be listed as well.
Several other Transient states have two spaces between the columns.

With these changes the transient state looks like this:
```
      (setq spacemacs--symbol-overlay-transient-state-doc "
%s
 [_n_] next    [_N_/_p_] prev       [_d_] def            [_f_/_b_] switch  [_t_] scope
 [_e_] echo    [_o_]^^   unoverlay  [_O_] unoverlay all  [_c_]^^   copy    [_z_] center
 [_s_] search  [_r_]^^   replace    [_R_] rename         ^^^^              [_q_] quit")
```

Before:

before

After:

after

kenkangxgwe: Hi @duianto . Thanks for the review. I am not that familiar with hydra. I will refine this as you pointed out.

kenkangxgwe:
Yes, symbol-overlay just changed the definition of `symbol-overlay-get-list` last month, as you may see here https://github.com/wolray/symbol-overlay/commit/ad77d6cd1e48adc7345503ca86f3e81877033703. I will fix it soon.

kenkangxgwe:
@duianto , just force-pushed a new commit. Feel free to test and send me any feedback.

duianto:

PR #11664, opened 2018-11-29
Defer orgit package

yuhan0:
Just a simple change to optimize startup times - without deferring it was adding 6-7 seconds to the total startup because of `(require 'org)` and `(require 'magit)`


syl20bnr:
Nice catch, Thank you ! 💯 

PR #11667, opened 2018-11-30
Fix typo in mu4e documentation

lesteve:
I don't think the leading quote was intentional but maybe I am missing something.


smile13241324:

Issue #8414, opened 2017-02-20
Builtin Org mode (8.x) breaking Spacemacs Org mode (9.x)


RockyRoad29:
#### Description :octocat:
Builtin Org mode (8.x) breaking Spacemacs Org mode (9.x)

Some org packages are loaded from `org-plus-contrib` ,
which currently depends on org version 9.0.x ,
while some other are loaded from the built-in emacs,
which is usually 8.x .

This causes many incompatibilities due e.g. to changes
of variable to functions, like 
[`org-link-types`](https://emacs.stackexchange.com/questions/29471/error-message-symbols-function-definition-is-void-org-link-types-when-i-open) ( [change](https://lists.gnu.org/archive/html/emacs-orgmode/2016-07/txtJiS_iys87F.txt) ),
or [`org-babel-check-confirm-evaluate`](http://orgmode.org/Changes.html#org184e052)

#### Reproduction guide :beetle:
- Install a fresh spacemacs from git master branch (0.200.7)
- Evaluate lisp expressions, e.g. in ielm or `*scratch*` in *lisp interaction mode*.

```elisp
; These are ok:
(org-version)
"9.0.5"
(locate-library "org")
"/home/mich/.emacs.d/elpa/org-plus-contrib-20170210/org.elc"

; These are not:
(find-lisp-object-file-name #'org-version t)
"/usr/share/emacs/25.1/lisp/org/org.el"
(functionp 'org-link-types)
nil
(find-lisp-object-file-name 'org-link-types 'defvar)
"/usr/share/emacs/25.1/lisp/org/org.el"
(functionp 'org-babel-check-confirm-evaluate)
nil
```

Yes, it looks like `(org-version)` is lying !!!

However, the responsability of loading the libs properly comes to Spacemacs
I think.

I think that `#'find-lisp-object-file-name` is what is used 
in `#'describe-function` but it didn't work as expected for
functions from `'ob`. Anyway there is some discrepancy around.

*Observed behaviour:* :eyes: :broken_heart:

Some actions in org-mode fail and you'd see messages like:

    File mode specification error: (void-function org-link-types)
    

    org-element--set-regexps: Symbol’s function definition is void: org-link-types
 
(see more on #8074') .

Sometimes, I managed to get rid of these `org-link-types` errors, but **then** (with some recompilation voodoo) I found that source block evaluation was failing (package `ob` : org-babel):

    Evaluation of this emacs-lisp code block is disabled.

(see more on #7641)

[...]

Issue #7641, opened 2016-11-03
Org-Mode Evaluation of code disabled


atrifex:
I am trying to run code as a part of org-mode's code blocks. However, it keeps saying that evaluation of code-blocks is disabled. I am not sure where to go to turn this option on. Please let me know. Thank you in advance.

TheBB:
http://orgmode.org/manual/index.html

Atrifex:
I actually looked through this page.

I am currently doing the following in my user-config()

`  (with-eval-after-load 'org
    (require 'ob-python)
    (org-babel-do-load-languages
     'org-babel-load-languages
     '((C . t)
       (python . t)))
  )

LemmingAvalanche:
I was able to get Python to work. 

In my `dotspacemacs/user-init`:

``` elisp
(require 'ob-python)
(org-babel-do-load-languages
  'org-babel-load-languages
  '((python . t)))
```

So I didn’t use the `with-eval-after-load` bit. 

I also included the Python layer in `dotspacemacs/layers`.

Then I executed `org-babel-execute-src-block` while point was inside the source
block below: 

```
#+BEGIN_SRC python
import time
print("Hello, today's date is %s" % time.ctime())
print('Two plus two is')
return 2 + 2
#+END_SRC

#+RESULTS:
: 4
```

I get prompted about whether I want to execute the code each time I execute that
function.

⁂

Could you try the above? Then try with C if that works.

Atrifex:
It did help with that. However, for some reason the bullets just appear as *, **, ***. This is not a big deal however cuz i am just manually able to turn them on with little to no effort. Thank.

LemmingAvalanche:
@TheBB it seems that this can be closed.

agzam:
Let's re-open this issue, something definitely not right, in org 9.0 you have to load explicitly for every language you want to use?

bmag:
@agzam it seems the issue has been resolved for @Atrifex so I closed it. I don't use org-babel and I'm not aware of the differences with Org 9.0, can you explain the bug using the usual template please?

agzam:
@TheBB it was kinda expected - things to break due to org-mode 9.0. Maybe we need a separate issue or maybe each one should be treated individually - it's up to you guys, but there are couple of things that appear to be broken at the moment: 

- at startup:

     Error (use-package): org post-:config hook: Unknown keyword: :export-block

[...]

PR #11679, opened 2018-12-03
Recompiling Elpa: don't skip missing .elc files

RockyRoad29:
#### Addressing several issues arising after emacs/spacemacs upgrade

This could be a follow up of issues like #8414 , #7641 and may help with several
others where recompilation is involved.

A commonly proposed solution is to delete all byte-compiled modules to force
their recompilation, but it may not work as expected.

- Upgrade emacs or spacemacs.

Mostly issues related with org-contrib packages,
e.g. source code blocks evaluation (org-babel).

Recompilation skips many packages, and version conflicts still cause issues.

A smooth transistion towards a better world for everyone ?
#### About `spacemacs/recompile-elpa`

Given the docstring of
[`byte-recompile-directory`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Compilation-Functions.html#Compilation-Functions#index-byte_002drecompile_002ddirectory-1161
"link to gnu manual") :

    (byte-recompile-directory DIRECTORY &optional ARG FORCE)

> Recompile every ‘.el’ file in DIRECTORY that needs recompilation.
> This happens when a ‘.elc’ file exists but is older than the ‘.el’ file.
> Files in subdirectories of DIRECTORY are processed also.
>
> If the ‘.elc’ file does not exist, normally this function **does not**
> compile the corresponding ‘.el’ file.  However, if the prefix argument
> ARG is 0, that means do compile all those files.  A nonzero
> ARG means ask the user, for each such ‘.el’ file, whether to
> compile it.  A nonzero ARG also means ask about each subdirectory
> before scanning it.
>
> If the third argument FORCE is non-nil, recompile every ‘.el’ file
> that already has a ‘.elc’ file.

I think that `spacemacs/recompile-elpa` could be rewritten like this:

```elisp
      (defun spacemacs/recompile-elpa (arg)
        "Compile or recompile packages in elpa directory, if needed, that is
    if the corresponding .elc file is either missing or outdated.

      If ARG is non-nil, also recompile every `.el' file, regardless of date.

      Useful if you switch Emacs versions."
        (interactive "P")
        (byte-recompile-directory package-user-dir 0 arg))
```

Currently, it just calls

```elisp
  (byte-recompile-directory package-user-dir nil t)
```

which means that *missing .elc files won't be rebuilt*, but existing
ones will be recompiled over and over each time (time-consuming for often no improvement).

### DIFF
--- a/core/core-funcs.el
+++ b/core/core-funcs.el
@@ -258,11 +258,18 @@ result, incrementing passed-tests and total-tests."
      (concat "Hidden Mode Line Mode enabled.  "
              "Use M-x hidden-mode-line-mode to make the mode-line appear."))))
 
-(defun spacemacs/recompile-elpa ()
-  "Recompile packages in elpa directory. Useful if you switch
-Emacs versions."
-  (interactive)
-  (byte-recompile-directory package-user-dir nil t))
+;; https://github.com/syl20bnr/spacemacs/issues/8414
+(defun spacemacs/recompile-elpa (arg)
+  "Compile or recompile packages in elpa directory, if needed, that is
+    if the corresponding .elc file is either missing or outdated.
+
+      If ARG is non-nil, also recompile every `.el' file, regardless of date.
+
+      Useful if you switch Emacs versions."
+  (interactive "P")
+  ;; First argument must be 0 (not nil) to get missing .elc files rebuilt.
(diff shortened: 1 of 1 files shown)


RockyRoad29:
Thank you @syl20bnr <3 <3 <3 spacemacs

Issue #9025, opened 2017-06-02
"SPC sf" no longer works


synic:
#### Description :octocat:

`SPC sf` no longer works.

#### Reproduction guide :beetle:
- Start Emacs
- Open a file, type `SPC sf`, choose a directory you'd like to search in, and
  type some search terms that should be present in some of the files in the
  directory. 
  
Seems to be any recent version of spacemacs, I can observe the behavior even on
the web based spacemacs on http://spacemacs.org.  

This used to work, so is a regression somewhere.

*Observed behaviour:* :eyes: :broken_heart:
Search results are blank.

*Expected behaviour:* :heart: :smile:
Search results are search results.

#### System Info :computer:
- OS: darwin
- Emacs: 25.2.1
- Spacemacs: 0.200.9
- Spacemacs branch: master (rev. 8e1af145)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(csv lua nginx sql auto-completion emacs-lisp git yaml javascript html markdown extra-langs python org osx python ipython-notebook django syntax-checking spell-checking colors spacemacs-layouts vimscript helm
     (version-control :variables version-control-diff-tool 'diff-hl)
     (c-c   :variables c-c  -enable-clang-support t)
     (shell :variables shell-default-height 30 shell-default-position 'bottom))
```
- System configuration features: NOTIFY ACL LIBXML2 ZLIB TOOLKIT_SCROLL_BARS

JAremko:
@synic  I seems to work as intended [spacemacs/helm-files-smart-do-search ](https://github.com/syl20bnr/spacemacs/blob/bd7ef98e4c35fd87538dd2a81356cc83f5fd02f3/layers/%2Bcompletion/helm/packages.el#L246)
```
Search in opened buffers using `dotspacemacs-search-tools'.
Search for a search tool in the order provided by `dotspacemacs-search-tools'
If DEFAULT-INPUTP is non nil then the current region or symbol at point
are used as default input.
```
It successfully finds patterns in open buffers.

- - - - - -

I will mark it as `as expected :P` until proven otherwise :smile:

synic:
@JAremko Then what is this supposed to be doing differently?  https://github.com/syl20bnr/spacemacs/blob/bd7ef98e4c35fd87538dd2a81356cc83f5fd02f3/layers/%2Bcompletion/helm/packages.el#L298

`SPC sb` should be search open buffers.

[...]

PR #11685, opened 2018-12-04
Add a note on grep prerequisite in the README

vnctaing:
As mentioned in https://github.com/syl20bnr/spacemacs/issues/9025, this should fix the issue on  search on macos and help future people running into the issue


duianto:

Issue #11676, opened 2018-12-02
', g g' broken in Haskell with Intero mode


magthe:
#### Description :octocat:
When using `intero-mode` for Haskell `, g g` results in GTAGS-based goto definition instead of the one coming from `intero-mode`.

#### Reproduction guide :beetle:
- Configure the Haskell layer to use Intero
- Start Emacs
- Open a Haskell file
- Move to a function call site and press `, g g`

*Observed behaviour:* :eyes: :broken_heart:
A notice about GTAGS missing is shown, and then I'm offered to tell Emacs where to find the GTAGS file.

*Expected behaviour:* :heart: :smile:
That I'm taken to the function definition.

#### The changeset introducing the broken behaviour

Using `git bisect` I've found that `, g g` worked find before the change 152d41e7cf907482fcea7781a5e6337fd8a07752. Unfortunately I don't really understand what that change does, so I can't really provide a patch myself.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 26.1
- Spacemacs: 0.300.0
- Spacemacs branch: nil (rev. 152d41e7c)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
((auto-completion :variables auto-completion-enable-help-tooltip t auto-completion-enable-snippets-in-popup t auto-completion-enable-sort-by-usage t)
 dash docker evil-commentary evil-snipe git github gtags
 (helm :variables helm-no-header t)
 mu4e restclient syntax-checking
 (clojure :variables clojure-enable-sayid t clojure-enable-clj-refactor t)
 emacs-lisp
 (haskell :variables haskell-completion-backend 'intero)
 shell-scripts html json latex markdown
 (org :variables org-enable-trello-support t)
 pandoc terraform yaml archlinux clipboard-fix)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GSETTINGS NOTIFY ACL GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 MODULES THREADS LIBSYSTEMD LCMS2

zilongshanren:
@magthe You could try my patch, i also come across this issue and now it works at my side.

magthe:
@zilongshanren thanks!

Almost the same thing is in #11562.

magthe:
[...]

PR #11562, opened 2018-11-07
Fix jump handlers conflict for haskell and intero

CthulhuDen:
`intero-mode` is a minor mode which is usually active alongside `haskell-mode`. In code as it was previously in my installation hook from define-jump-handlers for haskell-mode ran *after* hook set up for intero-mode so `spacemacs-jump-handlers` got it's final value from `spacemacs-jump-handlers-haskell-mode`, so it did not contain `intero-goto-definition`.

Instead of defining separate `spacemacs-jump-handlers-intero-mode` via macro I modify `spacemacs-jump-handlers-haskell-mode` when initializing intero.

PR #11699, opened 2018-12-09
fix haskell intero goto definition

zilongshanren:
fix issue: https://github.com/syl20bnr/spacemacs/issues/11676


magthe:
I think this might be a duplicate of #11562 .

zilongshanren:
close due to duplication

Issue #11689, opened 2018-12-05
[feature-request] support for xah-fly-keys to have modal support for emacs editing


nipunarora:
Currently, spacemacs does not have modal mode support similar to `xah-fly-keys` for emacs editing mode (admittedly vi already has one). Could we get support for `xah-fly-keys` as a layer in spacemacs?

gcentauri:
I have opened a PR #11701  to add this as a layer.  Feel free to comment on it if you have opinions on how to integrate another modal map.

PR #11701, opened 2018-12-09
Add: xah-fly-keys layer for alternative modal keymap

gcentauri:
This adds a layer for the `xah-fly-keys` package to address #11689 .  I also use this keymap and turned my personal config into a layer to share.  I could use some help refining it, and making sure which-key hints don't get funky, but this should add mostly plug-and-play integration for those who are using `emacs` editing mode and want to try modal editing without going to the dark side of `evil`,

### DIFF
--- a/layers/+misc/xah-fly-keys/packages.el
+++ b/layers/+misc/xah-fly-keys/packages.el
@@ -1,0 +1,49 @@
+;;; packages.el --- xah-fly-keys layer packages file for Spacemacs.
+;;
+;; Copyright (c) 2012-2017 Sylvain Benner & Contributors
+;;
+;; Author: Grant Shangreaux <shshoshin@protonmail.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+;;; Code:
+
+(defconst xah-fly-keys-packages
+  '(
+    xah-fly-keys
+    magit
+    )
+  )
+
+
+(defun xah-fly-keys/init-xah-fly-keys ()
+  (use-package xah-fly-keys
+    ;; :defer t
+    :diminish (xah-fly-keys . "∑")
+    :init
+    (setq xah-fly-use-control-key nil)
+
+    :config
+    (xah-fly-keys-set-layout xah-fly-keys-layout)
+    (message "activating xah-fly-keys")
+    (add-hook 'xah-fly-command-mode-activate-hook 'xah-fly-keys-command-mode-on)
+    (add-hook 'xah-fly-insert-mode-activate-hook  'xah-fly-keys-insert-mode-on)
+    (define-key xah-fly-key-map (kbd "M-SPC") 'xah-fly-command-mode-activate)
+    (xah-fly-keys 1)
+    )
+  )
+
+(defun xah-fly-keys/pre-init-magit ()
+  "Switch to insert mode in magit buffers and popups."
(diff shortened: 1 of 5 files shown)


layers/+misc/xah-fly-keys/packages.el
```diff
@@ -0,0 +1,49 @@
...
+  '(
+    xah-fly-keys
+    magit
+    )
+  )
+
+
+(defun xah-fly-keys/init-xah-fly-keys ()
+  (use-package xah-fly-keys
+    ;; :defer t
```
gcentauri: I'm not sure that `xah-fly-keys` has autoloads set up.  If i tried defering this, I was having issues getting the mode to be active on startup.

duianto:
There's also an issue with returning from insert mode to command mode by pressing `M-SPC` in Ubuntu, because a menu appears in the top left corner:
Minimize
Maximize
Move
...

I don't know if Ubuntu Emacs users disable the OS binding `M-SPC` or how it's handled, but currently this `xah-fly-keys` layer only returns to command mode from insert mode with the `Home` key in Ubuntu.

In windows both `M-SPC` and `Home` works.

`F8` doesn't seem to do anything in either Windows or Ubuntu.

`F8` works without Spacemacs in both Windows and Ubuntu.

duianto:
Maybe the installation guide should be rewritten to state which variables should be changed before enabling the `xah-fly-keys` layer. I enabled the layer before changing the editing style to `'emacs`. It took a little while to be able to change the variables in a half evil and half xah-fly-keys mode :smile:

Here's a possible rewrite:

#### Before enabling the `xah-fly-keys` layer

Change these existing `.spacemacs` variables:
- Change the `dotspacemacs-editing-style` variable to `'emacs`
- Prevent the Spacemacs `SPC` bindings from interfering with `xah-fly-keys`, by setting the
`dotspacemacs-leader-key` variable to `""` (an empty string).
- To use the `xah-fly-keys` cursor state colors, set the 
`dotspacemacs-colorize-cursor-according-to-state` variable to `nil`

And finally add `xah-fly-keys` to the `dotspacemacs-configuration-layers` list, and restart Spacemacs.

gcentauri:
Ah, yes. Thanks for pointing those things out.  `M-SPC` is primarily used on my work machine, I have the same issue on Arch using GNOME. I should have some time soon to update some docs and add more customization features if keymaps are conflicting with the window manager.

I was trying to figure out how `evil` is documented in Spacemacs, but clearly most of the keybindings are documented in the evil package.  I have been hitting some other issues with keybinding conflicts with the `common-lisp` layer or maybe `slime`, so I'll try to sort those out too.

gcentauri:
I'm not sure when I will get to it, but I realized that it would be good to integrate some more Spacemacs features into this keymap.  For example, in xah-fly-keys command mode, `n` is bound to `isearch`.  It seems like mapping it to either `helm-swoop` or Ivy's `swiper` would be nice to have.

prnsml:
@duianto 
Yes, looks like alt+space is captured by Window Manager, at least in Xubuntu I had to disable/clear "Window operation menu" action in XFCE Window Manager:
window manager

gcentauri:
I've shifted work back to a repository for a private layer. There's a few other users there, I will plan to update this commit when we get some things smoothed out (including docs).

I'm wondering how deep into keybinding documentation this README should go?  I will take a look at the `evil` documentation in spacemacs.  If anyone here has opinions, please let me know.

Also, I was considering including some more images of the keymap from Xah's documentation if additional size is not an issue.

PR #11702, opened 2018-12-10
Add a format binding for RuboCop format

sei40kr:
Bind `SPC m =` ( `, =` ) to `rubocop-autocorrect-current-file`.

### DIFF
--- a/layers/+lang/ruby/README.org
+++ b/layers/+lang/ruby/README.org
@@ -142,6 +142,7 @@ directory local variables.
 
 | Key binding   | Description                                          |
 |---------------+------------------------------------------------------|
+| ~SPC m = r~   | Runs RuboCop on the currently visited file           |
 | ~SPC m r r f~ | Runs RuboCop on the currently visited file           |
 | ~SPC m r r F~ | Runs auto-correct on the currently visited file      |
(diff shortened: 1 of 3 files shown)


robbyoconnor:
Why is this needed? `,rrF`does the same thing.

sei40kr:
@robbyoconnor 
Because we need a common key binding for formatting in any major mode.

cormacc:
@sei40kr Might consider binding under `, = r` instead? The lsp layer binds `, = b` to lsp-format-buffer, and I think there's been some work towards adding an lsp backend to the ruby layer. Using `,=` as a prefix rather than keybinding facilitates grouping multiple formatting options (e.g. format buffer, format selection, format using rubocop, lsp-backend etc.)

sei40kr:
@cormacc 
Oh I didn't notice, I'll fix it.

sei40kr:
Should we use `rubocopfmt.el` instead?
`rubocop-autocorrect` shows corrected errors on every format, so it's uncomfortable.
https://github.com/jimeh/rubocopfmt.el

sei40kr:
https://github.com/syl20bnr/spacemacs/pull/12283

PR #11703, opened 2018-12-10
wip: Enable emmet-mode in typescript-tsx-mode

sei40kr:
Enable Emmet in `typescript-tsx-mode`.

Emmet is used in `html-mode`, `react-mode`, and `typescript-tsx-mode`. (use-package in `html` layer)
Perhaps we should move `emmet-mode` to a separated layer.


sei40kr:
I noticed `typescript-tsx-mode` is derived from `web-mode` so calling `emmet-mode` implicitly is unnecessary.
Still `emmet-expand-jsx-className?` is needed. I'll fix this PR.

sei40kr:
https://github.com/syl20bnr/spacemacs/pull/12288

Issue #11479, opened 2018-10-18
Helm doesn't load automatically for some commands


duianto:
### Description

Helm doesn't load automatically for some commands, like rename `SPC f R` or when checking out a branch in magit `bb`. But if a helm buffer is opened with for example `SPC SPC` then the rename and checkout branch commands open in a helm buffer until the next time Emacs is restarted.

### Steps to reproduce

1. Start Emacs, and before opening a helm buffer like `SPC SPC`
2. Open `.spacemacs` by pressing `SPC f e d`
3. Open a rename prompt by pressing `SPC f R`

### Observed behavior

The minibuffer shows the rename prompt:

```text
New name: ~/.spacemacs
```

Helm doesn't seem to have been loaded yet, because if a helm buffer is opened with for example `SPC SPC` then `SPC f R` opens in a helm buffer, until the next time Emacs is restarted.

### Expected behavior

Helm should probably load when a command that uses the helm buffer is called.

#### System Info :computer:
- OS: windows-nt
- Emacs: 26.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 7ccddbcec)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(auto-completion autohotkey emacs-lisp git helm html markdown org treemacs version-control)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG SOUND NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS THREADS LCMS2

sdwolfz:
I think this is the result of having helm deffered, @syl20bnr is the best person tho handle this as I remember this change was done to improve startup time.

duianto:
This issue has been fixed by this PR:
Load helm before read-file-name and completing-read #11877

PR #11705, opened 2018-12-10
Fix helm loading for layouts transient state

deb0ch:
Calling the transient state and its functions (particularly `SPC l L` for loading from file) doesn't load helm, so until helm is loaded by something else the default Emacs method is used (I don't remember how it's called).

This makes that helm will always be used for opening layouts from file and all the rest inside the transient state, may there be more currently or in the future.

### DIFF
--- a/layers/+spacemacs/spacemacs-layouts/packages.el
+++ b/layers/+spacemacs/spacemacs-layouts/packages.el
@@ -103,6 +103,7 @@
 
 
 (defun spacemacs-layouts/post-init-helm ()
+  (spacemacs||set-helm-key "l" spacemacs/layouts-transient-state/body)
   (spacemacs/set-leader-keys
     "bB" 'spacemacs-layouts/non-restricted-buffer-list-helm
     "pl" 'spacemacs/helm-persp-switch-project))


deb0ch:
Now that I think of it, should I update the changelog at the same time ? I think it's the new policy after #11741 but not sure

sdwolfz:
I will write proper documentation on adding stuff to the changelog after finishing with #11741 to make sure it's done as part of the PR or as part of the cherry-pick.

sdwolfz:
I'll have to elaborate on that a bit more, define some sections, add an example, and prepare a skeleton for the next release's changelog.

duianto:
The same issue occurs with  `SPC f R` (`spacemacs/rename-current-buffer-file`), as described in issue:
Helm doesn't load automatically for some commands #11479

Instead of fixing them individually (there probably are more), maybe a more universal solution could be to advice `read-file-name` to load helm the first time `read-file-name` is called.

This works for both `SPC l L` and `SPC f R`:
```elisp
(defun load-helm-before-read-file-name
  (prompt &optional dir default-filename mustmatch initial predicate)
  (require 'helm)
  (helm/init-helm)
  (advice-remove 'read-file-name #'load-helm-before-read-file-name))

(advice-add 'read-file-name :before #'load-helm-before-read-file-name)
```

The same issue occurs when trying to checkout a branch `bb` or popping a stash `zp` in Magit before helm has loaded, but Magit uses `magit-completing-read` instead of `read-file-name`.

This works for both `bb` and `zp`:
```elisp
    (defun load-helm-before-magit-completing-read
        (prompt collection &optional
                predicate require-match initial-input hist def fallback)
      (require 'helm)
      (helm/init-helm)
      (advice-remove
       'magit-completing-read #'load-helm-before-magit-completing-read))

    (advice-add 'magit-completing-read :before #'load-helm-before-magit-completing-read)
```

sdwolfz:
@syl20bnr please take a look at @duianto's suggestion as you were the one that lazy loaded helm a long time ago in order to speed up startup. Do you see any reason why this won't work?

deb0ch:
I don't like advising in general but under this light I don't see any real, factual drawback in this case. If Helm itself is specific about these functions and no others then instinctively I'd say it makes more sense

duianto:
@sdwolfz Done:
Load helm before read-file-name and completing-read #11877

PR #11706, opened 2018-12-10
Fix flyspell correction on spell-checking layer

bltavares:
Using the `spell-checking` layer mini-mode `Space S .` would cause an error

```
Wrong type argument: commandp -- flyspell-correct-previous-word-generic
```

According to the [README](https://github.com/d12frosted/flyspell-correct#deprecations-in-v05), it has since moved the command to a new name, which is not available anymore.

This commit changes the command used to point to the new name to make the transient mode work again.

### DIFF
--- a/layers/+checkers/spell-checking/packages.el
+++ b/layers/+checkers/spell-checking/packages.el
@@ -60,7 +60,7 @@ Spell Commands^^             Other
         ("b" flyspell-buffer)
         ("d" spell-checking/change-dictionary)
         ("n" flyspell-goto-next-error)
-        ("c" flyspell-correct-previous-word-generic)
+        ("c" flyspell-correct-previous)
         ("Q" flyspell-mode :exit t)
         ("q" nil :exit t)
         ("t" spacemacs/toggle-spelling-checking))


d12frosted:
Hey, thank you for PR!

That's funny that I forgot to fix the rename here. 

I would actually suggest using `flyspell-correct-wrapper` instead. It has much more functionality.

bltavares:
Good point, I didn't know.

The good thing of `flyspell-correct-wrapper` is that it uses the `ivy` (in my case).
But I've also noticed that it does not give you the option to save the current word into the dictionary.

If `flyspell-correct-wrapper` gets implemented, it would be nice to have a new command on the transient state to save the word.

d12frosted:
> The good thing of flyspell-correct-wrapper is that it uses the ivy (in my case).

Actually, all of them should use `flyspell-correct-interface`. If you are using `flyspell-correct-ivy` package (e.g. `ivy` completion), it should be `ivy` for all completion functions from `flyspell-correct` package.

> But I've also noticed that it does not give you the option to save the current word into the dictionary.

It gets you a several commands. You just need to `M-o` to get the hint about available commands. E.g. `M-o k` means `skip`.

So I am not sure that you need a transient state here.

bltavares:
Oh, I didn't knew about `M-o`. I've noticed that the other function displays better shortcuts on the bar under it.

Regarding the transient mode. Is the suggestion to remove it, or to adopt the `flyspell-correct-wrapper` on it? I didn't quite understand, sorry.

duianto:
@d12frosted What's the next step for this PR?

d12frosted:
@duianto nothing changed - it's better to use `flyspell-correct-wrapper`. :smile_cat:

duianto:
`flyspell-correct-wrapper` behaves differently depending on how many universal arguments it receives:
>(flyspell-correct-wrapper ARG)

>Correct spelling error in a dwim fashion based on ARG.

>- One SPC u enables rapid mode.
>- Two SPC u’s changes direction of spelling
>  errors search.
>- Three SPC u’s changes direction of spelling
>  errors search and enables rapid mode.

But this PR changes the transient state binding: `SPC S . c`

Currently `SPC u` closes the transient state.

`C-u` works, but `SPC u` is the universal argument key binding with the vim editing style.

I tried adding:
```
        ("SPC u" universal-argument)
```
but it doesn't accumulate, it only inserts one `C-u`, which enters rapid mode.

#### Question 1
Might it be better to add separate bindings for each wrapper behavior?
possibly:
- c correct previous word
- C correct next word
- r rapid mode correct previous word
- R rapid mode correct next word

(maybe they can be written shorter?)

That would make them easily discoverable, at least beginners won't know that one can call functions with universal arguments, let alone multiple times.


#### Question 2
If `SPC u` is preferred, then how does one make `SPC u` accumulate in a transient state?

#### Side note:
How does one exit flyspell rapid mode?

I had to press `C-g` or `Esc` 10-20 times until it had checked all previous typos/duplicates :smile:

d12frosted:
Indeed, the wrapper works differently based on the universal argument. It perfectly works with Emacs editing style :smile_cat: Though I never checked it with Vim editing style.

In terms of discoverability, I agree with you - having more key bindings is nice. Though it kind of contradict the idea of the prefix :smile_cat: 

Not sure how to answer your second question.

Regarding the `C-g` for rapid mode - I don't know why it doesn't work for you. When I am in rapid mode - hitting `C-g` break the flow.

duianto:
#### This happens when I try to exit Rapid mode in Spacemacs
- Open a new buffer `SPC b N n`
- Type: `oen otw thre`
- Spellcheck the buffer `SPC S b`
- With the cursor on the last word `thre`
- Call `flyspell-correct-wrapper` with one universal argument: `SPC u SPC S c`

A Helm Flyspell buffer opens with two source headings:
- Suggestions for "otw" in dictionary "Default"
- Options
  - save "otw"
  - accept (session) "otw"
  - accept (buffer) "otw"
  - skip "otw"

Pressing `C-g` shows suggestions for "oen",
`C-g` again, suggestions for "thre",
`C-g` again, now the Helm Flyspell buffer closes (probably because there are no more words to check).

The same thing happens if the editing style is set to `emacs`. And 
`C-u M-m S c` is called.

#### Without Spacemacs, `C-g` cancels Rapid mode

Starting `Emacs -q` and initializing the Spacemacs packages directory
```
(setq package-user-dir "~/.emacs.d/elpa/26.2/develop/")
(package-initialize)
```

New buffer `C-x b` `test` `RET`
type: `oen otw thre`
`M-x` `flyspell-buffer` `RET`
`C-u` `M-x` `flyspell-correct-wrapper`

The minibuffer shows:
>Correcting 'thre'

Pressing `C-g` shows `Quit`

#### A possible clue:
In Spacemacs `C-h l` (`view-lossage`) shows:
`C-g [helm-keyboard-quit]`

Without Spacemacs it shows:
`C-g [abort-recursive-edit]`

duianto:
`SPC u` does work multiple times, if they are pressed before opening the Spell Checking Transient State
`SPC S .` and pressing `c` correct word (`flyspell-correct-wrapper`).

Since `C-u` is the default Emacs universal argument binding, and it also works in the TS, then it's probably fine to leave it like this for now, making `SPC u` work in a TS can be a possible future project.

Canceling rapid mode is more of a flyspell-correct / Spacemacs issue unrelated to this PR.

And any additional transient state bindings will have to be a separate issue request or PR.

duianto:

d12frosted:
Wow. Interesting. Thank you for the investigation.

> Canceling rapid mode is more of a flyspell-correct / Spacemacs issue unrelated to this PR.

Totally agree on that!

PR #11723, opened 2018-12-15
Julia layer compatibility with new lsp-mode

gdkrmr:
There are some breaking changes in `lsp-mode` which are being fixed with this PR for the `julia` layer. For this to work you need to 
- install latest master of `lsp-mode` (or simply the `lsp` layer), melpa stable of `lsp-mode` still seems to be on an older version.
- `lsp-julia` from https://github.com/non-Jedi/lsp-julia/pull/5
- A Julia 1.0 compatible version of `LanguageServer.jl`, which is not yet released.

Not sure if this should be merged right away or after the official releases of `lsp-mode` and `LanguageServer.jl`


gdkrmr:
Everything works, but I still get the following error in the message buffer repeatedly:

```
Error running timer: (wrong-type-argument symbolp (quote flycheck-mode))
```

gdkrmr:
```elisp
Debugger entered--Lisp error: (wrong-type-argument symbolp (quote flycheck-mode))
  boundp((quote flycheck-mode))
  lsp-ui-sideline--diagnostics()
  lsp-ui-sideline--run()
  #f(compiled-function () #<bytecode 0x24794e5>)()
  apply(#f(compiled-function () #<bytecode 0x24794e5>) nil)
  timer-event-handler([t 0 0 200000 nil #f(compiled-function () #<bytecode 0x24794e5>) nil idle 0])
```

gdkrmr:
`LanguageServer.jl` `v0.5.0` has been released.

Compro-Prasad:
@gdkrmr I think the backtrace points to some bug in `lsp-ui`. Can you confirm that by disabling `lsp-ui`? If yes then you can report the bug to `lsp-ui` repo.

gdkrmr:
The `lsp-ui` related error has disappeared on its own as it seems, https://github.com/emacs-lsp/lsp-ui/issues/225.

gdkrmr:
until non-Jedi/lsp-julia#5 gets merged you can either manually use that branch or add the following to your `dotspacemacs/user-config`:
```elisp
  (with-eval-after-load 'lsp-julia
    (defcustom lsp-julia-default-environment "~/.julia/environments/v1.0"
      "The path to the default environment."
      :type 'string
      :group 'lsp-julia)
    (defun lsp-julia--get-root ()
      (let ((dir (locate-dominating-file default-directory "Project.toml")))
        (if dir (expand-file-name dir)
          (expand-file-name lsp-julia-default-environment))))
    (defun lsp-julia--rls-command ()
      `(,lsp-julia-command
        ,@lsp-julia-flags
        ,(concat "-e using LanguageServer, Sockets, SymbolServer;"
                 " server = LanguageServer.LanguageServerInstance("
                 " stdin, stdout, false,"
                 " \"" (lsp-julia--get-root) "\","
                 " \"\", Dict());"
                 " server.runlinter = true;"
                 " run(server);")))
    (lsp-register-client
     (make-lsp-client :new-connection (lsp-stdio-connection 'lsp-julia--rls-command)
                      :major-modes '(julia-mode)
                      :server-id 'julia-ls))
    )

```

sdwolfz:
@gdkrmr Does it make sense to add that code snippet in the layer until https://github.com/non-Jedi/lsp-julia/pull/5 is merged? I would not want to release this being broken, but then again the old version was broken as well so there's not really much else I can do about it.

gdkrmr:
I would just wait until non-Jedi/lsp-julia#5 gets merged which would be soon.

PR #11729, opened 2018-12-18
[treemacs] Use deferred git-mode by default.

Alexander-Miller:
Deferred git-mode adds a 0.5s delay to fontification, so treemacs will almost never have to wait for a git process to finish.


sdwolfz:
Thank you :heart:!

PR #11732, opened 2018-12-19
Fix #9950: Hook added to disable hl-todo-mode in org files

### LINKED ISSUE #9950: Inconsistent highlighting in org mode
#### Description :octocat:

I'm seeing strange highlighting in my .org files. The same string (e.g. `DONE` or `TODO` is getting highlighted in different ways, based on rules I can't figure out. It also seems to be stateful in some way, because I can delete a line and then re-add it, and the highlighting will change.

You can see an example below of what it looks like:

different highlighting

 (I'm so far not able to create a reliable MWE - perhaps with the info below you can provide some pointers towards one?)

`DONE` text shows up in 2 different styles, one bright and one pale.

I ran SPC U SPC SPC what-cursor-position for both the brightly-colored and the paler version, and got the folloring output. The main difference seems to be that in the bright colored version the face is listed as `org-done`, whereas in the paler one it is `(:inherit hl-todo :foreground "#afd8af")`.

##### Bright "DONE"

```
             position: 5088 of 8193 (62%), column: 5
            character: N (displayed as N) (codepoint 78, #o116, #x4e)
    preferred charset: ascii (ASCII (ISO646 IRV))
code point in charset: 0x4E
               script: latin
               syntax: w 	which means: word
             category: .:Base, L:Left-to-right (strong), U:Uppercase, a:ASCII, l:Latin, r:Roman
             to input: type "C-x 8 RET 4e" or "C-x 8 RET LATIN CAPITAL LETTER N"
          buffer code: #x4E
            file code: #x4E (encoded by coding system utf-8-unix)
              display: by this font (glyph code)
    xft:-ADBO-Source Code Pro-bold-normal-normal-*-18-*-*-*-m-0-iso10646-1 (#x0F)

Character code properties: customize what to show
  name: LATIN CAPITAL LETTER N
  general-category: Lu (Letter, Uppercase)
  decomposition: (78) ('N')

There are text properties here:
  face                 org-done
  fontified            t
  org-category         "todo"
```

##### Pale "DONE"

```
             position: 5183 of 8193 (63%), column: 5
            character: N (displayed as N) (codepoint 78, #o116, #x4e)
    preferred charset: ascii (ASCII (ISO646 IRV))
code point in charset: 0x4E
               script: latin
               syntax: w 	which means: word
             category: .:Base, L:Left-to-right (strong), U:Uppercase, a:ASCII, l:Latin, r:Roman
             to input: type "C-x 8 RET 4e" or "C-x 8 RET LATIN CAPITAL LETTER N"
          buffer code: #x4E
            file code: #x4E (encoded by coding system utf-8-unix)
              display: by this font (glyph code)
    xft:-ADBO-Source Code Pro-bold-normal-normal-*-18-*-*-*-m-0-iso10646-1 (#x0F)

Character code properties: customize what to show
  name: LATIN CAPITAL LETTER N
  general-category: Lu (Letter, Uppercase)
  decomposition: (78) ('N')

There are text properties here:
  face                 (:inherit hl-todo :foreground "#afd8af")
  fontified            t
  org-category         "todo"
```

`DONE` text is highlighted the same in both cases.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 25.3.1
- Spacemacs: 0.200.10
- Spacemacs branch: master (rev. 4bb4cb46)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(helm emacs-lisp org)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GCONF GSETTINGS NOTIFY ACL GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 MODULES

samspills:
I was getting inconsistent highlighting in org-mode and found issue #9950. A
comment in the hl-todo package says that org-mode shouldn't be getting included
because org defines its own todo states, but this isn't the case for
me ([comment for reference](https://github.com/tarsius/hl-todo/blob/master/hl-todo.el#L69-L71)). I couldn't figure out how to get hl-todo to activate only `modes that are text-mode but are not org-mode`. So instead I left the hl-todo activiation alone and added a hook to disable the mode in org files (+ a variable to turn the disable hook off if people want).

The added hook will disable hl-todo-mode so that it doesn't clobber the `org-todo-keyword-faces` but hl-todo faces will still be active in source code blocks inside of an org file.

### DIFF
--- a/layers/+emacs/org/config.el
+++ b/layers/+emacs/org/config.el
@@ -37,3 +37,6 @@ used.")
 
 (defvar org-enable-trello-support nil
   "If non-nil org-trello is configured")
+
+(defvar org-enable-hl-todo-mode nil
+  "If non-nil hl-todo is enabled in org buffers")
--- a/layers/+emacs/org/config.el
+++ b/layers/+emacs/org/config.el
@@ -37,6 +37,3 @@ used.")
 
 (defvar org-enable-trello-support nil
   "If non-nil org-trello is configured")
-
-(defvar org-enable-hl-todo-mode nil
-  "If non-nil hl-todo is enabled in org buffers")
--- a/layers/+emacs/org/funcs.el
+++ b/layers/+emacs/org/funcs.el
@@ -67,9 +67,3 @@
 (defun spacemacs/org-trello-push-card ()
   (interactive)
   (org-trello-sync-card))
-
-
-
-(defun spacemacs//hl-todo-org-mode ()
-  (unless org-enable-hl-todo-mode
-    (hl-todo-mode -1)))
--- a/layers/+emacs/org/funcs.el
+++ b/layers/+emacs/org/funcs.el
@@ -67,3 +67,9 @@
 (defun spacemacs/org-trello-push-card ()
   (interactive)
   (org-trello-sync-card))
+
+
+
+(defun spacemacs//hl-todo-org-mode ()
+  (unless org-enable-hl-todo-mode
+    (hl-todo-mode -1)))
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -18,7 +18,6 @@
         evil-surround
         gnuplot
         (helm-org-rifle :toggle (configuration-layer/layer-used-p 'helm))
-        hl-todo
         htmlize
         ;; ob, org and org-agenda are installed by `org-plus-contrib'
         (ob :location built-in)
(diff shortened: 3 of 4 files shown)


yuhan0:
From the hl-todo source you linked to, it looks like `global-hl-todo-mode` already has its own mechanism for dealing with org-mode. Perhaps a better fix would be to turn on the global mode instead of manually adding it to prog-mode-hook and text-mode-hook

samspills:
Oh.... Yeah that looks a lot simpler 😅

samspills:
@yuhan0 Just saw your edit, I wouldn't think the definition is necessary because that's already the default defined in hl-todo.el. Is it a convention to add variables like that to spacemacs layers to future proof against future changes in upstream packages?

yuhan0:
Hmm I'm not sure about any convention actually – it's probably fine to leave it out in this case, maybe with a small comment explaining the default behaviour.

duianto:
The conflict needs to be resolved.

And could you add an entry to the org section of changelog.develop:
https://github.com/syl20bnr/spacemacs/blob/develop/CHANGELOG.develop#org-1

PR #11756, opened 2018-12-27
ivy: Fix counsel search

aaronjensen:
This function was renamed upstream:
https://github.com/abo-abo/swiper/commit/58bf1b94c8346491b906aa306f5ed734be67310c

If we should support both versions, please let me know the best way to do that.

### DIFF
--- a/layers/+completion/ivy/funcs.el
+++ b/layers/+completion/ivy/funcs.el
@@ -79,7 +79,7 @@
                                (prog1 (pop split)
                                  (setq string (mapconcat #'identity split " -- "))))
                            ""))
-                   (regex (counsel-unquote-regex-parens
+                   (regex (counsel--elisp-to-pcre
                            (setq ivy--old-re
                                  (ivy--regex string)))))
               (setq spacemacs--counsel-search-cmd (format base-cmd args regex))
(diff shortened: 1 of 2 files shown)


robbyoconnor:
@sdwolfz

sdwolfz:
I think we should as we have a snapshot in https://github.com/syl20bnr/spacelpa from almost two months ago that we intend to use for 0.300 (and not sure if we're going to change that, it was done mainly to ensure magit's compatibility).

aaronjensen:
@sdwolfz so should only `develop` have this then? or should we add a backwards compatible polyfill?

aaronjensen:
Should I also wrap in an `(unless (fboundp 'counsel-unquote-regex-parens) ...` ?

yuhan0:
Shouldn't that be the other way around? Target the latest version of counsel but handle back-compatibility via  

```elisp
(unless (fboundp 'counsel--elisp-to-pcre)
  (defalias 'counsel--elisp-to-pcre 'counsel-unquote-regex-parens))
```

aaronjensen:
@jmromer that's going the wrong way, isn't it? I'll update this pr.

aaronjensen:
@sdwolfz this should handle it for both versions.

sdwolfz:
Thank you :heart:!

PR #11760, opened 2018-12-27
Include magit/forge SQLite db file in .gitignore

Asgavar:
Where by `forge` I mean: https://github.com/magit/forge

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ eshell/history
 .emacs.desktop.lock
 eshell/alias
 eshell/lastdir
+forge-database.sqlite
 url/*
 my-org/
 org-files/


nixmaniack:
As of now magit/forge isn't included in spacemacs. Once it's included it'd be better to setup `forge-database-file` inside `~/.emacs.d/.cache`(as per general convention of spacemacs package configs) than add it to `.gitignore`.

sdwolfz:
Thank you :heart:! 

syl20bnr:
Ok already done actually. I think this is safe to revert as it should not be necessary.

syl20bnr:
Seems that `forge` creates the file in `.emacs.d` when installing...

Miciah:
> Seems that `forge` creates the file in `.emacs.d` when installing...

Do we need to set `forge-database-file` earlier? Maybe in `layers/+source-control/github/config.el`?

PR #11772, opened 2018-12-30
Update change log (+893)

yuhan0:
Summarizing: https://github.com/syl20bnr/spacemacs/commits/develop?before=69a89cc41243cfb189238a2297fde4498d7fcf6d+893
Starting with: 25fa9b0

New pointer: https://github.com/syl20bnr/spacemacs/commits/develop?before=69a89cc41243cfb189238a2297fde4498d7fcf6d+858
Start with: d2c3b84

Note: I added a new "Regression fixes" section under Various changes for regression bug fixes which don't seem to fit anywhere else. 
Also started to add new entries to Documentation Fixes section on a new line each, which should make for easier merging.

PR #11779, opened 2018-12-31
Update CHANGELOG before=69a89cc412+788

Compro-Prasad:
Page: https://github.com/syl20bnr/spacemacs/commits/develop?before=69a89cc41243cfb189238a2297fde4498d7fcf6d+788


CHANGELOG.develop
```diff
@@ -299,6 +301,13 @@ sane way, here is the complet list of changed key bindings:
   - init function of all packages
   - post-init functions for all packages  (thanks to Sylvain Benner)
 - Backported battery fix for macOS (thanks to Sylvain Benner)
+- Defer packages by default using =use-package-always-defer= (thanks to Sylvain
+  Benner)
+- Regressions caused by removing =:defer t=. Fixed it by adding =:demand t=.
```
yuhan0: Btw, I added a "various regression fixes" section at the bottom in my previous PR (not yet merged):
https://github.com/syl20bnr/spacemacs/pull/11772/files#diff-f3337695f47c8cf4b3d417aa7971aa6bR1144
 
I think makes more sense for non user-facing regression fixes like this to go there? Also since they're quite hard to describe

Compro-Prasad: So, they should be removed?

yuhan0: I was thinking you could move it to the same place under the Various improvements section:

```
- Various regression fixes (thanks to bmag)
```

so it'll be merged accordingly :)

Miciah: I believe the following guideline applies:

> If you notice that a commit builds on top of an existing one you can group those together.

Sylvain made the change to defer by default, and bmag built on that change by adding some necessary overrides.

Where this gets confusing is that Sylvain subsequently reverted the change, changing the default for `use-package-always-defer` back to `nil`, which means that these two change log entries may ultimately be unnecessary.  In light of that, and since Sylvain and and bmag are already listed under "Various code style improvements", I believe that these two entries related to deferred loading can be omitted entirely.

Miciah:
> Don't forget to add your name in the `Release notes summarized` section if you wish to.

CHANGELOG.develop
Compro-Prasad: I am also curious about the lazy loading things. They are done all over the history IMO. So, should they be put in a specific section?

Issue #8849, opened 2017-05-04
user-config is loaded only after files listed on command line are opened


bgamari:
One annoyance I've noticed is that my `auto-mode-alist` configuration that I set from `spacemacs/user-config` doesn't apply to files opened via the command line (e.g. starting `emacs file.c` from the shell). I dug into this a bit this morning and found that `find-file-hook` is run before `spacemacs/user-config` is run when files are given on the command line.

It seeems that either the documentation should recommend against putting this sort of configuration in `spacemacs/user-config` or `user-config` should be made to run earlier in initialization.

LdBeth:
As long as you know what you need, just put the related configs to `dotspacemacs/user-init`.
Spacemacs documentation has the following descriptions about the two configs:
>- `dotspacemacs/user-init` is called immediately after `dotspacemacs/init`,
  before layer configuration. This function is mostly useful for variables
  that need to be set before packages are loaded.
>- `dotspacemacs/user-config` is called at the very end of Spacemacs
  initialization after layers configuration. This is the place where most of
  your configurations should be done. Unless it is explicitly specified that a
  variable should be set before a package is loaded, you should place your code
  here.

You are not prohibited to use `user-init`, but most configs should be put in `users-config` so they can override the behavior of Spacemacs rather than be overridden.

next-mad-hatter:
I'd expect that opening a file for editing would result in same editor state regardless of the way I opened it, i.e. argue it is a bug.

I'm afraid with my current spacemacs knowledge I won't be able to change this for now, but think it definietly deserves a mention in the docs until such time as someone can.

PR #11781, opened 2019-01-01
A note about user-config and files supplied via command line arguments

next-mad-hatter:
I'd argue that user-config getting executed after files have been opened is not expected behaviour and should be mentioned in the docs more prominently (took me some time to stumble over #8849).


smile13241324:

PR #11792, opened 2019-01-03
Enable experimental support for lsp within the php layer

DanielRichtmann:
Adds php LSP support.

Inspired by the python implementation.

Completion and jump to definition are working in first tests.

Still TODO:

- Further testing

Since this is my first work on `spacemacs` and also coding in `elisp` I'm grateful for all comments.

### DIFF
--- a/layers/+lang/php/README.org
+++ b/layers/+lang/php/README.org
@@ -8,6 +8,10 @@
 - [[#description][Description]]
   - [[#features][Features:]]
 - [[#install][Install]]
+  - [[#layer][Layer]]
+  - [[#setting-the-backend][Setting the backend]]
+- [[#backend][Backend]]
+  - [[#language-server-protocol][Language Server Protocol]]
 - [[#key-bindings][Key bindings]]
 - [[#usage][Usage]]
 
@@ -20,14 +24,37 @@ This layer adds PHP language support to Spacemacs.
 - Complete and jump to define with [[https://github.com/xcwen/ac-php][company-php]]
 - Run tests with PHPUnit
 - Reformat code with PHP CBF
+- Support for the [[https://langserver.org/][Language Server Protocol]] (experimental)
 
 The =gtags= layer is recommended to benefit from better =eldoc= and
 =helm-gtags=.
 
 * Install
+** Layer
 To use this configuration layer, add it to your =~/.spacemacs=. You will need to
 add =php= to the existing =dotspacemacs-configuration-layers= list in this
 file.
+
+** Setting the backend
+To activate the backend set the layer variable =php-backend=:
+
+#+BEGIN_SRC elisp
+  (php :variables php-backend 'lsp)
+#+END_SRC
+
+For spacemacs to use lsp you also need to add =lsp= to the
+=dotspacemacs-configuration-layers= in your =~/.spacemacs= file.
+
+* Backend
+** Language Server Protocol
+You have to install the php language server:
+
+#+BEGIN_SRC sh
+  composer require felixfbecker/language-server
+  composer run-script --working-dir=vendor/felixfbecker/language-server parse-stubs
+#+END_SRC
+
++You can find further information on the project's [[https://github.com/felixfbecker/php-language-server%20][GitHub page]].
 
 * Key bindings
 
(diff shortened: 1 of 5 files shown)


layers/+lang/php/funcs.el
```diff
@@ -0,0 +1,42 @@
...
+      (lsp)
+    (message "`lsp' layer is not installed, please add `lsp' layer to your dotfile.")))
+
+(defun spacemacs//php-setup-company ()
+  "Conditionally setup company based on backend."
+  (message "%s setting up company" php-backend)
+  (pcase php-backend
+    (`lsp (spacemacs//php-setup-lsp-company))))
+
+(defun spacemacs//php-setup-lsp-company ()
```
yyoncho: `lsp` automatically configures company if it is present so this is not needed. Not sure whether the other layers are updated.

DanielRichtmann: I now removed the company-related stuff, how does it look to you now?

Please also note, that I'm explicitly keeping the commits here for documentation purposes. I'm going to squash these once this PR is ready to be cherry-picked.

robbyoconnor:
You need to update the `CHANGELOG.develop` file :)

layers/+lang/php/README.org
```diff
@@ -18,15 +22,38 @@ This layer adds PHP language support to Spacemacs.
...
+#+END_SRC
+
+For spacemacs to use lsp you also need to add =lsp= to the
+=dotspacemacs-configuration-layers= in your =~/.spacemacs= file.
+
+* Backend
+** Language Server Protocol
+You have to install the php language server:
+
+#+BEGIN_SRC sh
```
yyoncho: Based on the users' feedback https://github.com/bmewburn/vscode-intelephense was recommended as the better server(and it is easier to install). IMO this should be mentioned here.

DanielRichtmann: Ok, thanks for the info. I'll play around with it once I find the time. Maybe we can use it instead of `php-language-server` then or let the user toggle between the two like in the c++ layer.

yyoncho: > let the user toggle between the two like in the c++ layer.

IMO better to use lsp-mode facilities to switch the server. C++ case is a bit different since there are a lot of functions that are language specific and also they require additional Emacs package to work while the php servers are bundled in lsp-mode.

DanielRichtmann: I've now tested intelephense and I agree that it's the better server. I updated the README and the changelog.

deuill:
Hi, is this still planned for merging at some point, or does it need any work done to push it over the line?

DanielRichtmann:
I've been using this in conjunction with `intelephense` for quite some time now and am happy with it in general. My test case is the [Shopware](https://github.com/shopware/shopware) code base. Here, references/definitions work nicely, but autocompletion isn't really great in the sense that it only completes php builtin functions. I didn't look into that any further due to lack of time but 
it may be a merit of Shopware, so some further testing may be needed. Glad about any pointers/help.

robbyoconnor:
A reminder that you don't have to wait for PRs to be merged and can pull
them in and use them as-is..

smile13241324:
Hi guys,

I would like to merge this however before I can do this someone need to have a test with this and see whether the layer works as expected with the LSP servers mentioned. 

So if someone has some time and is working regularly with PHP it would really speed things up if a test could be done and some results be posted here.

Thanks
@deuill @robbyoconnor

deuill:
I've been testing this with `intelephense` against a fairly large code-base (~3000 files, 400000 lines of code) -- it works very well, with no perceivable lag and accurate completions (not sure why @DanielRichtmann was having issues with his completions).

I'm sticking with this and will report any issues I find, but I suspect any issues will be down to `lsp-mode` itself. I'm pretty happy myself.

smile13241324:
Thank you for contributing to spacemacs :purple_heart:, I have squashed your commits and cherry-picked the result into develop. You can safely remove your branch.

Thank you very much for the test @deuill.

Btw is the usage section behind the key bindings in the readme.org of the layer still necessary when the LSP backend is used? Sounded like something from the non LSP backend...

DanielRichtmann:
I'm happy that this made it in :) Yes I guess that this section is only valid for the non LSP backend.

PR #11797, opened 2019-01-05
Update Symbol Highlight TS: Reorder keys

duianto:
What: Move "[z] recenter" to end of first row.
Why:  Top row (red keys) will not exit, bottom row (blue keys) exits.

---

Before: 

symbol highlight transient state before

After:

symbol highlight transient state after


duianto:

Issue #11788, opened 2019-01-02
org-archive-subtree does not work


dmbfm:
#### Description :octocat:
org-archive-subtree does not work.

#### Reproduction guide :beetle:
- Start Emacs
- Open an .org file
- Go to a headline and try to archive it (<C-c C-x C-a> or by calling
  the org-archive-subtree function)

*Observed behaviour:* :eyes: :broken_heart:
I get the following message: 'invalid function: org-preserve-variables' and the
headline is not archive. Neither the org file is changed nor a archive file is
created.

*Expected behaviour:* :heart: :smile:
The headline should be archived.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 26.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 2c28c7d21)
- Graphic display: t
- Distribution: spacemacs
- Editing style: emacs
- Completion: helm
- Layers:
```elisp
nil```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND DBUS GSETTINGS NOTIFY ACL LIBSELINUX GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 MODULES THREADS XWIDGETS LCMS2

dmbfm:
It also does not work on master. Works fine on vanilla emacs.

glogiotatidis:
I get the same error while refiling issues

samspills:
I also have this issue. But I've noticed a few other issues with the org layer: 
1. `, i t` now raises an error when setting tags (I think it needs to switch to using `org-set-tags-command` in new org release)
2. pressing tab no longer expands `<s`, `<q` etc. into the appropriate block

I'm 80% sure these issues are also related to the new org release. In that case, should a single issue be created to track the updates that need to be made to the org layer? I have a branch with `org-set-tags-command`, but the fix for tab completion and and archiving might be beyond my beginner abilities (I haven't had time to look into these problems yet though).

thanhvg:
> pressing tab no longer expands `<s`, `<q` etc. into the appropriate block

Org 9.2 has dropped this feature, you need org-tempo to re-enable it
https://orgmode.org/Changes.html#org1b5e967

I can't reproduce the org-archive-subtree issue though. I'm on dev branch and packages are up to date

samspills:
[...]

PR #11804, opened 2019-01-06
Updates org-layer key bindings for Org 9.2 release

samspills:
## Changes
This PR addresses my comment in issues #11788. 

The new Org 9.2 release seems to have changed the functionality of `org-set-tags`, reserving it only for lisp calls, and added a new interactive function `org-set-tags-command`. The method of interacting with org easy-templates was also changed. 

I updated the `, i t` binding to use `org-set-tags-command`, and I also added a new `, i x` binding for template expansion. (The old method of template expansion could be enabled with org-tempo, but I like the new style so I didn't look it into adding that to the layer).

### DIFF
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -306,6 +306,7 @@ Will work on both org-mode and any mode that accepts plain html."
         ;; attachments
         "A" 'org-attach
         ;; insertion
+        "ib" 'org-insert-structure-template
         "id" 'org-insert-drawer
         "ie" 'org-set-effort
         "if" 'org-footnote-new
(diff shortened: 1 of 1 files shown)


boothead:
Nice one @samspills I just came to submit an issue for this :-)

samspills:
@yuhan0 `i x` doesn't feel arbitrary to me because I still think of it as template expansion, it's just that now I choose the template to expand through a menu instead of needing to remember the shortcut. That said I'm not attached to the binding and will switch if you (or anyone else) feels strongly!

yuhan0:
No strong feelings on this either, perhaps one of the maintainers can decide? Just a comment that `x` generally stands for "text" elsewhere in Spacemacs, and I didn't realise that it meant "expand" here before reading your post.

samspills:
I didn't think about the confusion with `x` usually meaning text, that's a great point. I'll switch to the `, i b` binding :)

sdwolfz:
Thank you :heart:! 
Commit: 6d9dbcf6bbc18931a2e0efab0884d3d22993bd18

samspills:
Thanks :D

PR #11808, opened 2019-01-07
Add ibuffer square brackets filter group motion

duianto:
Match the magit buffer section/sibling navigation keys (bound in evil-magit) and
the evil-collections motion section also suggests using [ and ] for navigation
between sections.

And list the default ibuffer filter group motion bindings:
Forward: TAB and M-n
Backward: S-TAB and M-p


smile13241324:

PR #11848, opened 2019-01-15
[vinegar] Support custom dired hide details mode by default.

wallevy:
I'm using vinegar. It simplifies `dired` and is useful. But the dired details is hidden by default, and there is no config can custom dired-hide-details-mode. So I add a variable `vinegar-dired-hide-details` to support that. The default value is `t`, it will not break the previous behavior, and you can set it as below:

```emacs-lisp
(vinegar :variables vinegar-dired-hide-details nil)
```


smile13241324:

PR #11851, opened 2019-01-15
[treemacs] Make sure treemacs' buffers are ignored by winum.

Alexander-Miller:
There's a bit of a race condition going on, so that *sometimes* winum's ignored buffers are using their default value (winum's own spacemacs config defines them with setq). It seems to depend on when treemacs is loaded exactly. This change duplicates a small piece of compatibility code from treemacs.

I also silenced the message from the treemacs-toggle-fixed-width. I don't think that change is worth is own PR.


sdwolfz:
Thank you :heart:!
Commit: ea96c525e95c8c5e03f74c1fd48a8cba5c29a6b1

Issue #7378, opened 2016-10-13
Add keybindings for smerge-mode


rpglover64:
When using magit to resolve a merge conflict `smerge-mode` opens automatically for me; however, all the keybindings are under <kbd>C-c ^</kbd>. I believe it would be better to also have them under <kbd>SPC m</kbd> or <kbd>SPC g</kbd> or even <kbd>SPC g m</kbd>.
#### System Info :computer:
- OS: darwin
- Emacs: 24.5.1
- Spacemacs: 0.200.0
- Spacemacs branch: master (rev. e6357a5)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:

``` elisp
(auto-completion better-defaults haskell emacs-lisp git ivy emacs-lisp org agda deft spell-checking
                 (syntax-checking syntax-checking-enable-tooltips 'nil))
```

bmag:
For those who wonder, those bindings are:
```
C-c ^ RET       smerge-keep-current
C-c ^ =         Prefix Command
C-c ^ C         smerge-combine-with-next
C-c ^ E         smerge-ediff
C-c ^ R         smerge-refine
C-c ^ a         smerge-keep-all
C-c ^ b         smerge-keep-base
C-c ^ m         smerge-keep-mine
C-c ^ n         smerge-next
C-c ^ o         smerge-keep-other
C-c ^ p         smerge-prev
C-c ^ r         smerge-resolve

C-c ^ = <       smerge-diff-base-mine
C-c ^ = =       smerge-diff-mine-other
C-c ^ = >       smerge-diff-base-other
```
Since the bindings are mnemonic, we can just use the same keys under a different prefix. However, <kbd>SPC g s</kbd> is already taken by `magit-status`, and <kbd>SPC g m</kbd> is taken by `magit-dispatch-popup`. Maybe we can use <kbd>SPC g r</kbd> ("r" for "resolve").

We'd probably want a a transient-state <kbd>.</kbd> (that's a dot). We can also bind <kbd>j</kbd> to `smerge-next` and <kbd>k</kbd> to `smerge-previous`.

I think this issue is ready for work, for anyone who wants to implement it. What do you think @rpglover64?

rpglover64:
Sounds good! Unfortunately I don't think I can commit to working on it.

One more thought: it might be useful to start in the transient state (e.g. by an `smerge-mode-hook`).

perfectayush:
[...]

PR #11855, opened 2019-01-16
Extend smerge transient-state with diff keybindings

dwang20151005:
I often use these operations to ask how mine and other have diverged:

=< smerge-diff-base-mine
== smerge-diff-mine-other
=> smerge-diff-base-other

It's awkward to use their global keybindings under C-c ^ because that exits the
smerge transient-state.

Judging from issue GH-7378 there was no principled reason to omit them, only
that the author did not use them himself.

### DIFF
--- a/layers/+source-control/version-control/packages.el
+++ b/layers/+source-control/version-control/packages.el
@@ -275,9 +275,9 @@
  [_n_]^^    next hunk       [_b_] keep base          [_u_] undo
  [_N_/_p_]  prev hunk       [_m_] keep mine          [_r_] refine
  [_j_/_k_]  move up/down    [_a_] keep all           [_q_] quit
- ^^^^                       [_o_] keep other
- ^^^^                       [_c_] keep current
- ^^^^                       [_C_] combine with next"
+ ^^^^                       [_o_] keep other         [_=<_] diff base to mine
+ ^^^^                       [_c_] keep current       [_==_] diff mine to other
+ ^^^^                       [_C_] combine with next  [_=>_] diff base to other"
         :bindings
         ("n" smerge-next)
         ("p" smerge-prev)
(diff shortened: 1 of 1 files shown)


dwang20151005:
4 unexpected results:
       FAILED  test-package-install-org--package-desc-name-org-plus-contrib
       FAILED  test-package-install-org--package-desc-name-org-plus-contrib-already-installed
       FAILED  test-package-install-org--symbol-name-org-plus-contrib
       FAILED  test-package-install-org--symbol-name-org-plus-contrib-already-installed

Is this acceptable?  I haven't previously used CircleCI or emacs unit testing before, so I don't know how I broke these or even how to check that these used to pass.

yuhan0:
Why the "=" prefix though? Binding to `<`, `=` and `>` would be shorter. 
And it seems that the `keep-mine` ,`keep-other` commands have been obsoleted in favour of `keep-lower` and `keep-upper` - see https://github.com/alphapapa/unpackaged.el#hydra for reference.

dwang20151005:
# Regarding the choice of key sequence:

I wanted to follow the existing global keybindings so that people who already knew the global keybindings could easily learn the transient-state keybindings.  This appears to have been the convention so far, with only one or two exceptions.

I also like the grouping of the diff commands under `=`, which I find helps me remember them.

# Regarding the obsoleting

Aha!  This is why I am failing the integration test: 25.3, which I use, only has `keep-mine` and `keep-other` whereas 26.1 only has `keep-upper` and `keep-lower`.  I'll have to figure out some way to check which functions are available when I define the transient-state.  Thank you.

dwang20151005:
I'm not sure what changed, but now all tests pass.  Could someone review this PR?

dwang20151005:
Rebased again.  All tests pass.

This is 11 lines of diff and it's been sitting for over two months.  What's going on?

duianto:

dwang20151005:
Thanks!

PR #11858, opened 2019-01-17
Fix pasting with universal argument

### LINKED ISSUE #4219: How to stop auto indent?
I want to copy a pice of python code into emacs. But emacs makes it automatically to indent to make my code a mess. How can I solve the problem?

### LINKED ISSUE #8897: Can't disable auto-indent when pasting
`SPC u p` should paste without indent, but fails since at least an year.

> Spc-u 4 p actually does paste four times without autoindention. Fun.

The relevant issue is #4219, but is mislabeled as question/forum/Python, which might be the reason it's open since a couple of years. There's even a proposed fix at https://github.com/syl20bnr/spacemacs/issues/4219#issuecomment-246074883.

pritambaral:
The auto-indent advice around evil-paste-{before,after} calls the advised funcs
without removing the universal argument. This results in errors like:

    ad-Advice-evil-paste-before: Wrong type argument: number-or-marker-p, (4)

and, ultimately, in the paste failing.

This checks for the universal argument *before* calling the advised funcs,
adjusting the args to remove the universal argument.

Fixes #4219
Fixes #8897

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -1318,15 +1318,17 @@ Compare them on count first,and in case of tie sort them alphabetically."
  "If current mode is not one of spacemacs-indent-sensitive-modes
  indent yanked text (with universal arg don't indent)."
  (evil-start-undo-step)
- ad-do-it
- (if (and (not (equal '(4) (ad-get-arg 0)))
-          (not (member major-mode spacemacs-indent-sensitive-modes))
-          (or (derived-mode-p 'prog-mode)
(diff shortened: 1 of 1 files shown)


kommen:
Does this make https://github.com/syl20bnr/spacemacs/pull/11340 obsolete?

duianto:
The following PR is now on the `develop` branch:
Check for universal argument before paste #11340

It superseded this one with two additional features:
- Fix the errors related to universal argument when people enable multiple-cursors layer.
- Rewrite spacemacs|advise-commands with advice-add, as defadvice is obsoleted.

source: https://github.com/syl20bnr/spacemacs/pull/11340#issuecomment-486655845

PR #11870, opened 2019-01-19
[treemacs] Make treemacs the default file browser.

Alexander-Miller:
A fancy new default file tree, hopefully in time for the new release. 
Don't know if I've caught everything, I made my edits wherever I found references to neotree.

FYI @sdwolfz

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -7,28 +7,28 @@ When a release will be drafted, the notes will be copied at the top of
 the [[file:CHANGELOG.org][CHANGELOG.org]] file.
 
 #+BEGIN_EXAMPLE
-* Release 0.400.x
-** 0.400.0
-*** Breaking Changes
-**** Major
-**** Key bindings
-**** Layers
-***** Layer 1
-***** Layer 2
-***** Layer 3
-*** Hot new features
-*** Dotfile changes
-*** Core changes
-*** Distribution changes
-*** Layers
-**** New layers
-**** Layer 1
-**** Layer 2
-**** Layer 3
-*** Various improvements
-**** Code syle and organization
-**** Documentation and website
-**** Testing and CI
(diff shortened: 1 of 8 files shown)


Antiarchitect:
Guys, neotree was able to easily open a file in a numbered window by l-1 l-2 ... Can treemacs do it without any configuration? It would be nice Treemacs can do this by default.

Alexander-Miller:
oaa, oah & oav let you open files with ace-window.

JuanCaicedo:
Hi y'all! 😄 I'm finding that my navigation keys break after this change. It appears that `j`/`k` are still bound to the correct things, however since I'm using dvorak, I would expect those keys to be `h`/`t`. 

Is this by mistake (and those keys should instead by using something from the keyboard layer)? Or is the plan to only support the original evil binding, and users of other keyboards should remap things themselves?

Let me know if there is a better place to ask this question 👍

duianto:
@JuanCaicedo There doesn't seem to be a Treemacs config in the `keyboard-layout` layer yet.

Maybe you could fix that 👍 

A `treemacs` entry probably needs to be added here:
https://github.com/syl20bnr/spacemacs/blob/4b195ddfc9228256361e0b264fe974aa86ed51a8/layers/%2Bintl/keyboard-layout/packages.el#L12

The `keyboard-layout/pre-init-neotree` function is defined like this:
https://github.com/syl20bnr/spacemacs/blob/4b195ddfc9228256361e0b264fe974aa86ed51a8/layers/%2Bintl/keyboard-layout/packages.el#L432

it could probably be copied.

And here's how Treemacs defines the `j`/`k`, etc. keys:
https://github.com/Alexander-Miller/treemacs/blob/de1ba5217ca4c41fc36d405832b749b6fffd2b82/src/extra/treemacs-evil.el#L49

Issue #10663, opened 2018-05-10
diff-hl does not work correctly after syncing configuration


jvalkeejarvi:
#### Description :octocat:
diff-hl-highlight function is set to nil when running dotspacemacs/sync-configuration-layers

#### Reproduction guide :beetle:
- Start Emacs
- Set version-control layer variables as shown in System Info/Layers section below
- Run dotspacemacs/sync-configuration-layers

*Observed behaviour:* :eyes: :broken_heart:
Error of void function is shown and diff-hl does not display changed/added/deleted lines

*Expected behaviour:* :heart: :smile:
No errors and diff-hl works correctly

#### System Info :computer:
- OS: gnu/linux
- Emacs: 25.3.2
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 7596a156)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
(html javascript ivy auto-completion emacs-lisp git
      (version-control :variables version-control-diff-tool 'diff-hl version-control-diff-side 'left))
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GCONF GSETTINGS NOTIFY LIBSELINUX GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 MODULES

#### Backtrace :paw_prints:
```
Debugger entered--Lisp error: (void-function nil)
  nil(#<overlay from 20136 to 20136 in .spacemacs> insert top)
  diff-hl-add-highlighting(insert top)
  diff-hl-update()
  diff-hl-margin-mode(-1)
  (progn (setq diff-hl-side (if (eq version-control-diff-side (quote left)) (quote left) (quote right))) (diff-hl-margin-mode -1))
  (if init (progn (setq diff-hl-side (if (eq version-control-diff-side (quote left)) (quote left) (quote right))) (diff-hl-margin-mode -1)) (setq spacemacs--after-display-system-init-list (cons (function (lambda nil (setq diff-hl-side (if (eq version-control-diff-side ...) (quote left) (quote right))) (diff-hl-margin-mode -1))) spacemacs--after-display-system-init-list)))
[...]

PR #11871, opened 2019-01-20
Fix diff-hl-highlight function is set to nil

Voleking:
`diff-hl-margin-mode` is turned off after `display-system-init`, but if `diff-hl-margin-mode` close without opening, it will make `diff-hl-highlight function` set to nil, which cause this issue #10663.

This is my first pull request so feel free to tell if something in this issue should be improved.


sdwolfz:
Thank you :heart:! 
Commit: a8567cbed34f5ba72cf18836e90a320d2f1b7a0b.

PR #11877, opened 2019-01-20
Load helm before read-file-name and completing-read

duianto:
Fixes helm not being loaded before calls to:
read-file-name and completing-read.

These are some commands that call either of those two functions:
spacemacs/rename-current-buffer-file (SPC f R)

In a Treemacs window:
treemacs-add-project-to-workspace (C-p a)

In a Magit buffer:
magit-checkout (b b)

Thanks Miciah for a more elegant solution.

This commit also reverts:
Fix helm loading for layouts transient state #11705
because it's not needed anymore.

And the previously commented out transient hook minibuffer-setup-hook (it
doesn't seem to ever have been used) is removed because:
- It loads helm after a command is called that uses helm (instead of before)
- (spacemacs|hide-lighter helm-mode) has previously been moved to the
helm/init-helm :config section.

### DIFF
--- a/layers/+completion/helm/packages.el
+++ b/layers/+completion/helm/packages.el
@@ -63,11 +63,12 @@
     :defer (spacemacs/defer)
     :init
     (progn
-      ;; (spacemacs|add-transient-hook minibuffer-setup-hook
-      ;;   (lambda ()
-      ;;     (require 'helm-mode)
-      ;;     (spacemacs|hide-lighter helm-mode))
-      ;;   lazy-load-helm)
+      (spacemacs|add-transient-hook read-file-name
+        (lambda (&rest _args) (require 'helm))
+        lazy-load-helm-for-read-file-name)
+      (spacemacs|add-transient-hook completing-read
+        (lambda (&rest _args) (require 'helm))
+        lazy-load-helm-for-completing-read)
       (add-hook 'helm-cleanup-hook #'spacemacs//helm-cleanup)
       ;; key bindings
       ;; Use helm to provide :ls, unless ibuffer is used
--- a/layers/+spacemacs/spacemacs-layouts/packages.el
+++ b/layers/+spacemacs/spacemacs-layouts/packages.el
@@ -103,7 +103,6 @@
 
 
 (defun spacemacs-layouts/post-init-helm ()
-  (spacemacs||set-helm-key "l" spacemacs/layouts-transient-state/body)
   (spacemacs/set-leader-keys
     "bB" 'spacemacs-layouts/non-restricted-buffer-list-helm
     "pl" 'spacemacs/helm-persp-switch-project))


duianto:
Should both advices be removed when one of the `load-helm-before-*` functions are called?
The other one won't be needed since `helm/init-helm` already has been loaded.

sdwolfz:
I think there is a concept for this implemented in the core, take a look at transient hooks: https://github.com/syl20bnr/spacemacs/blob/develop/core/core-hooks.el#L40

duianto:
I've been reading about hooks and trying to figure out how to use: `spacemacs|add-transient-hook` to solve this issue, but I haven't figured out what hook to use as the first argument.

I found a usage of `spacemacs|add-transient-hook` in the `:init` section of `helm/init-helm`:
https://github.com/syl20bnr/spacemacs/blob/a46b784e85911afe8bd6648ca55c080ffae1081c/layers/%2Bcompletion/helm/packages.el#L66-L70

but they have been commented out in this commit:
Truly lazy load helm based on key bindings https://github.com/syl20bnr/spacemacs/commit/e24be2186adcbf1bafafd3ca2ea4e31d8919c760

When they are uncommented and Spacemacs is restarted. 

Then the first time `SPC f R` is called, it opens without helm but the second+ times it opens with helm.

deb0ch:
I agree that removing both advices makes more sense.

I think @duianto 's solution is simple and tackles the root problem. Using the primitive that you're talking about @duianto would only tackle the problem for transient states, _if_ it was working.

duianto:
Ok, I see now that, sdwolfz meant that something similar could probably be implemented.

I know even less about hooks than advicing, so I'll add the function for removing both advice, since it's the current working version.

If we come up with a better solution then we can switch to that.

bet4it:
`helm/init-helm` is called when `spacemacs` starts so we don't need to call it again.
A `require` call does nothing if the library is already loaded, so we can don't remove the advices. That's also what `spacemacs||set-helm-key` has done.

This pull request can be squashed into following code:
```elisp
(advice-add 'read-file-name :before                 
            #'(lambda (&rest args) (require 'helm)))
(advice-add 'completing-read :before                
            #'(lambda (&rest args) (require 'helm)))
```

duianto:
Confirmed, that works with the 3 commands above.

deb0ch:
On one hand there is more code for less runtime bloat, that is removing the advices from the advices.

On the other hand there is less code to maintain but we keep the advices forever, with the overhead of a few function calls and the empty call to `'require` that we assume to be neglectable.

bet4it:
@deb0ch All the keys hooked by `spacemacs||set-helm-key` are in the same situation, you can use <kbd>C-h k SPC f f</kbd> to see which function it is bound to, and have a look at 044237139b66058322e3824e7219f5741a0dd0ed.

I don't like this way, but it is already used by `spacemacs`.

Miciah:
> I've been reading about hooks and trying to figure out how to use: `spacemacs|add-transient-hook` to solve this issue, but I haven't figured out what hook to use as the first argument.

The following works for me:
```diff
modified   layers/+completion/helm/packages.el
@@ -68,6 +68,15 @@
       ;;     (require 'helm-mode)
       ;;     (spacemacs|hide-lighter helm-mode))
       ;;   lazy-load-helm)
+      (spacemacs|add-transient-hook read-file-name
+        (lambda (&rest _args) (require 'helm))
+        lazy-load-helm-for-read-file-name)
+      (spacemacs|add-transient-hook completing-read
+        (lambda (&rest _args) (require 'helm))
+        lazy-load-helm-for-completing-read)
+      (spacemacs|add-transient-hook minibuffer-setup-hook
+        (lambda (&rest _args) (require 'helm))
+        lazy-load-helm-for-minibuffer)
       (add-hook 'helm-cleanup-hook #'spacemacs//helm-cleanup)
       ;; key bindings
       ;; Use helm to provide :ls, unless ibuffer is used
```

> I found a usage of `spacemacs|add-transient-hook` in the `:init` section of `helm/init-helm`:
> [spacemacs/layers/+completion/helm/packages.el](https://github.com/syl20bnr/spacemacs/blob/a46b784e85911afe8bd6648ca55c080ffae1081c/layers/%2Bcompletion/helm/packages.el#L66-L70)
> 
> Lines 66 to 70 in [a46b784](/syl20bnr/spacemacs/commit/a46b784e85911afe8bd6648ca55c080ffae1081c)
> 
>  ;; (spacemacs|add-transient-hook minibuffer-setup-hook 
>  ;;   (lambda () 
>  ;;     (require 'helm-mode) 
>  ;;     (spacemacs|hide-lighter helm-mode)) 
>  ;;   lazy-load-helm) 
> but they have been commented out in this commit:
> Truly lazy load helm based on key bindings [e24be21](https://github.com/syl20bnr/spacemacs/commit/e24be2186adcbf1bafafd3ca2ea4e31d8919c760)
> 
> When they are uncommented and Spacemacs is restarted.
> 
> Then the first time `SPC f R` is called, it opens without helm but the second+ times it opens with helm.

The commented code adds a hook for `minibuffer-setup-hook`, but I found that a transient hook is needed for `read-file-name` or `completing-read` in order to auto-load Helm for <kbd>SPC f R</kbd>.

duianto:
I'll also remove the currently commented out lines:
```
      ;; (spacemacs|add-transient-hook minibuffer-setup-hook
      ;;   (lambda ()
      ;;     (require 'helm-mode)
      ;;     (spacemacs|hide-lighter helm-mode))
      ;;   lazy-load-helm)
```

because miciah's solution includes the `minibuffer-setup-hook`, and 
`(spacemacs|hide-lighter helm-mode)` has been moved to the `helm/init-helm` `:config` section.

deb0ch:
@sdwolfz Maybe this should be included in the 0.300 release and replace my limited fix of #11705 ?

@duianto Maybe this PR should also remove my limited fix of #11705

duianto:
@deb0ch Done.

bet4it:
@Miciah Should we add the hook for `minibuffer-setup-hook`? According to e24be2186adcbf1bafafd3ca2ea4e31d8919c760, this hook was never used before.

Miciah:
@bet4it, I don't know whether anything needs `minibuffer-setup-hook`.  I did not need it to get <kbd>SPC f R</kbd> to work.

duianto:
As I mentioned in the previous comment https://github.com/syl20bnr/spacemacs/pull/11877#issuecomment-456443543

>When they are uncommented and Spacemacs is restarted.
>
>Then the first time SPC f R is called, it opens without helm but the second+ times it opens with helm.

"they" is referring to the commented out transient hook `minibuffer-setup-hook`.

The same thing happens when Miciah's version of the transient hook `minibuffer-setup-hook` is called (without the transient hooks `read-file-name` and `completing-read`).

Since the goal is to load helm before a command is called that uses helm. 

I'll remove the `minibuffer-setup-hook`.

duianto:

Issue #11794, opened 2019-01-04
Background processes prevent `cargo` CLI commands from running


gnzlbg:
Background processes in the Rust layer prevent cargo commands from running by taking the lock in the project's directory. 

When I try to build my project, e.g., using `cargo test`, very very often `cargo test` is stuck at

```
cargo test
Blocking waiting for file lock on build directory
```

**forever** and I have to close all emacs buffers (or just restart emacs) from that project to unlock it. 

This is horrible, that this is the default behavior of the Rust layer is even worse, and AFAICT there is no way to disable this behavior in the Rust layer.

Issue #11623, opened 2018-11-21
Disable any libraries that "locks" Rust projects in the rust-layer


gnzlbg:
When editing Rust projects, when I try to build them, I often get: 

> Blocking waiting for file lock on build directory

And have to wait up to minutes until the build starts or the tests are run. The workaround, close spacemacs, then the build starts instantaneously. 

My suspicion, some library being used by the Rust layer is locking the project directory, maybe by executing some commands in the background. These libraries should release these locks instantaneously, and not hold to them for minutes without user input.

Blocking developer workflow for seconds or minutes is probably not worth whatever value these libraries might be trying to add.

gnzlbg:
> Turn off syntax checking in the -t menu.

How does one do that in the configuration file ?

Issue #11622, opened 2018-11-21
Disable "parentheses" matching / highlighting in rust layer


gnzlbg:
Whatever the library being used in the rust layer to match / highlight parentheses, it freezes spacemacs for seconds way too often, making it IMO unusable.

Reproducer: Try removing or adding new braces to any big file containing lots of blocks, like the unix module in the libc crate: https://github.com/rust-lang/libc/blob/master/src/unix/mod.rs

"Highlighting parentheses" is not an acceptable feature if it costs 10 second freezes.

gnzlbg:
> What's your hardware?

Macbook Air mid-2012 (8GB RAM, 1.8 Ghz Dual Core i5)

jrmiller82:
I think a 6-year old laptop is more your problem. I think these problems go away with a more powerful processor. Have you tried on a newer computer or desktop to see if the problem goes away?

gnzlbg:
Just to triple check, are you really suggesting that a ~2Ghz dual core with 8 Gb of RAM is not powerful enough to edit files containing 1000 lines of code ?

jrmiller82:
Yes. Spacemacs (emacs) isn't terribly efficient, and Rust (in any editor with code completion) is going to be running CPU-heavy background processes. It's not how many lines of code are in the specific file; it's how hard to compile your entire crate in rust. I have a 2013 MBPro that struggles to keep up with Rust development in spacemacs, and that likely has way more oomph than a 2012 Air.

jrmiller82:
There. It's likely blocking trying to call racer.

gnzlbg:
Hmm, that makes sense. I have the `racer` emacs package disabled like this:

```elisp
   ;; A list of packages that will not be installed and loaded.
   dotspacemacs-excluded-packages '(evil-mc racer)
```

Should I enable the package and install racer ?

PR #11882, opened 2019-01-21
Fix rust layer performance

### LINKED ISSUE #11241: mismatching parents make my rust workflow freeze for seconds
Say I have `foo<F>` and I delete the last `<` into `foo<F`, then the parents get "highlighted" and my emacs freezes for 5 seconds. The same happens when working with macros and getting mismatching `{` or `(`, etc. What do I need to disable to work around this and why is this enabled by default ?

gnzlbg:
This closes #11241 #11622 #11623 #11794 by reverting all changes that automatically enable broken and/or extremely poorly performing functionality in the Rust layer by default. 

The main things this disables are:

* auto-completion: Racer is a program that must be optionally installed by the user, and so is the Rust Language Server. Racer is slow on medium size projects, and gives incorrect autocompletions. The RLS crashes a lot, consumes a lot of memory, and is pretty much unusable on non-tiny projects. Enabling any of these by default, even if they are installed, as well as forcing the user to choose between one or another made spacemacs unusably slow for some of the projects i work on. 

    Those users who want to use them should opt-into these in their `.spacemacs` configuration files. In particular, given how often they crash, and how slow they can get, it doesn't really make sense to enable these unless their query always happen asynchronously in a different cancelable process that doesn't block the main editor thread and that cannot become a zombie to prevent dozens of zombie instances of the RLS consuming ~1Gb of RAM. 

* cargo, flycheck, and all other similar utilities: these are broken by design. Building a project and executing tests in Rust requires invoking `cargo build`, `cargo test`, or some other `cargo` sub-command. These won't run if some other process takes the lock of the project directory. Having these enabled in emacs running in the background prevents me from building my projects, running tests, etc. forcing me to kill all emacs rust buffers to release the lock. This makes for a horrible development experience. Those who want it should, again, opt into it, but a better design for this should make sure that the lock is not taken often, and when it is taken, then only for very brief periods of time and released immediately. 

* smartparens: completely freezes the editor for 5 seconds on tiny files O(1000k LOC) trying to match parentheses when editing code inside large-ish O(100LOC) Rust macros. The proposed solution (https://github.com/Fuco1/smartparens/issues/897) appears to be putting an upper bound of 1 second on the freezes. Freezing the editor for 1 or more seconds just to highlight parentheses in colors is not worth it. Users that want this should opt into this, and honestly, this is another computation that should run asynchronously instead of blocking the UI thread. 

I'm not against adding support for all of these things as opt-ins, but all new functionality appears to be added to this layer as "enabled by default without an option to disable it". Users of the rust-layer should not have to pay for things they don't or can't (e.g. because the tools don't cope well with project size) use.

### DIFF
--- a/layers/+lang/rust/config.el
+++ b/layers/+lang/rust/config.el
@@ -13,5 +13,5 @@
 
 (spacemacs|define-jump-handlers rust-mode)
 
-(defvar rust-backend 'racer
-  "The backend to use for completion. Possible values are `lsp' `racer'.")
+;; (defvar rust-backend 'racer
+;;   "The backend to use for completion. Possible values are `lsp' `racer'.")
--- a/layers/+lang/rust/funcs.el
+++ b/layers/+lang/rust/funcs.el
@@ -9,64 +9,64 @@
 ;;
 ;;; License: GPLv3
 
-(defun spacemacs/racer-describe ()
-  "Show a *Racer Help* buffer for the function or type at point.
-If `help-window-select' is non-nil, also select the help window."
-  (interactive)
-  (let ((window (racer-describe)))
-    (when help-window-select
-      (select-window window))))
+;; (defun spacemacs/racer-describe ()
+;;   "Show a *Racer Help* buffer for the function or type at point.
+;; If `help-window-select' is non-nil, also select the help window."
+;;   (interactive)
+;;   (let ((window (racer-describe)))
+;;     (when help-window-select
+;;       (select-window window))))
 
-(defun spacemacs/rust-quick-run ()
-  "Quickly run a Rust file using rustc.
-Meant for a quick-prototype flow only - use `spacemacs/open-junk-file' to
-open a junk Rust file, type in some code and quickly run it.
-If you want to use third-party crates, create a new project using `cargo-process-new' and run
-using `cargo-process-run'."
-  (interactive)
-  (let ((input-file-name (buffer-file-name))
-        (output-file-name (concat temporary-file-directory (make-temp-name "rustbin"))))
-    (compile
-     (format "rustc -o %s %s && %s"
-             (shell-quote-argument output-file-name)
-             (shell-quote-argument input-file-name)
-             (shell-quote-argument output-file-name)))))
+;; (defun spacemacs/rust-quick-run ()
+;;   "Quickly run a Rust file using rustc.
+;; Meant for a quick-prototype flow only - use `spacemacs/open-junk-file' to
+;; open a junk Rust file, type in some code and quickly run it.
+;; If you want to use third-party crates, create a new project using `cargo-process-new' and run
+;; using `cargo-process-run'."
+;;   (interactive)
+;;   (let ((input-file-name (buffer-file-name))
+;;         (output-file-name (concat temporary-file-directory (make-temp-name "rustbin"))))
+;;     (compile
+;;      (format "rustc -o %s %s && %s"
+;;              (shell-quote-argument output-file-name)
+;;              (shell-quote-argument input-file-name)
+;;              (shell-quote-argument output-file-name)))))
 
-(defun spacemacs//rust-setup-lsp ()
-  "Setup lsp backend"
-  (if (configuration-layer/layer-used-p 'lsp)
-      (lsp)
-    (message "`lsp' layer is not installed, please add `lsp' layer to your dotfile.")))
+;; (defun spacemacs//rust-setup-lsp ()
+;;  "Setup lsp backend"
+;;  (if (configuration-layer/layer-used-p 'lsp)
+;;      (lsp)
+;;    (message "`lsp' layer is not installed, please add `lsp' layer to your dotfile.")))
 
-(defun spacemacs//rust-setup-racer ()
-  "Setup racer backend"
-  (progn
-    (racer-mode)))
+;; (defun spacemacs//rust-setup-racer ()
+;;   "Setup racer backend"
+;;   (progn
+;;     (racer-mode)))
 
-(defun spacemacs//rust-setup-backend ()
-  "Conditionally setup rust backend."
-  (pcase rust-backend
-    (`racer (spacemacs//rust-setup-racer))
-    (`lsp (spacemacs//rust-setup-lsp))))
+;; (defun spacemacs//rust-setup-backend ()
+;;   "Conditionally setup rust backend."
+;;   (pcase rust-backend
+;;     (`racer (spacemacs//rust-setup-racer))
+;;     (`lsp (spacemacs//rust-setup-lsp))))
 
-(defun spacemacs//rust-setup-lsp-company ()
-  "Setup lsp auto-completion."
-  (if (configuration-layer/layer-used-p 'lsp)
-      (progn
-        (spacemacs|add-company-backends
-          :backends company-lsp
-          :modes rust-mode))
-    (message "`lsp' layer is not installed, please add `lsp' layer to your dotfile.")))
+;; (defun spacemacs//rust-setup-lsp-company ()
+;;   "Setup lsp auto-completion."
+;;   (if (configuration-layer/layer-used-p 'lsp)
+;;       (progn
+;;         (spacemacs|add-company-backends
+;;           :backends company-lsp
+;;           :modes rust-mode))
+;;     (message "`lsp' layer is not installed, please add `lsp' layer to your dotfile.")))
 
-(defun spacemacs//rust-setup-racer-company ()
-  "Setup racer auto-completion."
-        (spacemacs|add-company-backends
-          :backends company-capf
-          :modes rust-mode
-          :variables company-tooltip-align-annotations t))
+;; (defun spacemacs//rust-setup-racer-company ()
+;;   "Setup racer auto-completion."
+;;         (spacemacs|add-company-backends
+;;           :backends company-capf
+;;           :modes rust-mode
+;;           :variables company-tooltip-align-annotations t))
 
-(defun spacemacs//rust-setup-company ()
-  "Conditionally setup company based on backend."
-  (pcase rust-backend
-    (`racer (spacemacs//rust-setup-racer-company))
-    (`lsp (spacemacs//rust-setup-lsp-company))))
+;; (defun spacemacs//rust-setup-company ()
+;;   "Conditionally setup company based on backend."
+;;   (pcase rust-backend
+;;     (`racer (spacemacs//rust-setup-racer-company))
+;;     (`lsp (spacemacs//rust-setup-lsp-company))))
--- a/layers/+lang/rust/packages.el
+++ b/layers/+lang/rust/packages.el
@@ -11,93 +11,94 @@
 
 (setq rust-packages
       '(
-        cargo
-        company
+        ; cargo
+        ; company
         counsel-gtags
-        racer
-        flycheck
-        (flycheck-rust :requires flycheck)
+        ; racer
+        ; flycheck
+        ; (flycheck-rust :requires flycheck)
         ggtags
         helm-gtags
         rust-mode
-        smartparens
+        ; smartparens
         toml-mode
         ))
 
-(defun rust/init-cargo ()
-  (use-package cargo
-    :defer t
-    :init
-    (progn
-      (spacemacs/declare-prefix-for-mode 'rust-mode "mc" "cargo")
-      (spacemacs/set-leader-keys-for-major-mode 'rust-mode
-        "c." 'cargo-process-repeat
-        "cC" 'cargo-process-clean
-        "cX" 'cargo-process-run-example
-        "cc" 'cargo-process-build
-        "cd" 'cargo-process-doc
-        "cD" 'cargo-process-doc-open
-        "ce" 'cargo-process-bench
-        "cf" 'cargo-process-fmt
-        "ci" 'cargo-process-init
-        "cl" 'cargo-process-clippy
-        "cn" 'cargo-process-new
-        "co" 'cargo-process-current-file-tests
-        "cs" 'cargo-process-search
-        "ct" 'cargo-process-current-test
-        "cu" 'cargo-process-update
-        "cx" 'cargo-process-run
-        "cv" 'cargo-process-check
-        "t" 'cargo-process-test))))
+;; (defun rust/init-cargo ()
+;;   (use-package cargo
+;;     :defer t
+;;     :init
+;;     (progn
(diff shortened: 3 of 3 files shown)


sdwolfz:
@gnzlbg you raise some valid concerns here...
@jamartin9 can you please take a look?

Since this breaks functionality so bad, I think I'll merge this in for the next release and schedule restoring the code in the following bugfix release.

jamartin9:
@sdwolfz I will write up my thoughts on this.

I believe most of what is desired in this pr can be accomplished with little to no code change.

1. auto-complete - To disable all completion backends for the rust layer use `(rust :variables rust-backend nil)` in the layer config.
This will make it so no company backends are added and racer/rls/lsp will not be ran.
If a specific company backend for rust-mode only is desired then that can be done in your user config or as a another option to the the backend variable.

2. syntax-checking - There are many ways to disable flycheck or specific checkers.
It can be disabled by default with `(syntax-checking :variables syntax-checking-enable-by-default nil)` in the layer config.
It can be disabled by dir-locals.
It can be toggled off.
Add rust rust-clippy rust-cargo to excluded checkers
Disable flycheck for rust-mode via global list.

3. smartparens - exclude/configure the package to your preference. The only modification done in the rust layer is for lifetime specifiers. Similair to flycheck-mode this package can be configured/disabled many different ways.

All of these changes could be potentially added as an toggle/config option for easier access.

To address some your other raised issues:

1. 
> Racer is slow on medium size projects, and gives incorrect autocompletions. The RLS crashes a lot, consumes a lot of memory, and is pretty much unusable on non-tiny projects.

If there are issues with wrong/bad behavior (incorrect autocompleteions/crashing) I suggest reporting them upstream to be fixed.

Racer/LSP - 
The rust layer used racer by default when lsp was added.
RLS uses racer atm for better auto completion.
Both are very well known in the rust community (rls repo is from the rust-lang team) and is used/standardized in many projects despite being somewhat immature.
It would be nicer if users didn't have to install racer/rls for autocompletion but that is a seperate issue.
Given that using either RLS/RACER is optional any alternative autocompleteion utilities for rust code could be a solution too.

2. 
> cargo, flycheck, and all other similar utilities: these are broken by design.  

I disagree as synchronicity was probably not part of the design but the packages/utilities can be excluded.
However locking the project directory could be an issue. Without more information/steps to reproduce I can't suggest what is holding the lock/why.

Suggestions:
1. Dont merge this in its current state. Commenting out mostly configurable/optional code would be a large regression of functionality ( no autocomplete, syntax-checking or bindings to package manager). I suggest altering the docs/defaults/toggles to support a more minimalistic desired version. 
Alternatively since this removes so much not using the rust-layer and cherry picking the desired snippets ( basically toml-mode, rust-mode and ggtags ) into your own layer/user config should be easy.

2. Ensure up to date dependencies ( Racer, RLS, Emacs Packages, Cargo).
Many the issues listed could be due to old/broken/misconfigured packages ( ie old RUST_SRC_PATH for racer completion, old crashing version, etc)
So starting with a clean machine/config could help isolate the issues as would the logs from the processes in question. (using rustup may help with ensuring compatability with the src rls etc)

3. Define acceptable response times for generalized hardware specifications.
Performance is a very subjective thing without profiling information and specifications on acceptability.
Rust is not known to have the fastest compile times and an async stack would be nice but blocking sometime is mostly inevitable atm imo.
Removing optional functionality from the rust layer because it is 'too slow' is way too fuzzy a measure imo.

An example project that showcases issues with config/specs would be ideal for profiling/debugging/troubleshooting. 
I haven't had many of the issues mentioned but I am running a decent workstation. I will try on my very low end pc later to see if anything becomes obvious. 

I like the notion of opt-in behaviors but what the defaults should be is a conventions question. I can try to assist in achieving the desired default configuration while preserving functionality if it is deemed necessary to change the rust-layers defaults.

gnzlbg:
> All of these changes could be potentially added as an toggle/config option for easier access.

This sounds good to me. 

> If there are issues with wrong/bad behavior (incorrect autocompleteions/crashing) I suggest reporting them upstream to be fixed.

Upstream knows, the timeline to fix these issues was set for ~2021, but that was before the lead contributor quit Mozilla this week. Who knows when these tools will be fast and reliable.

> Without more information/steps to reproduce I can't suggest what is holding the lock/why.

I asked on gitter and others were also having the problem with emacs holding the lock. The time the lock was hold varied significantly, from "never happens", to 1-2 seconds, to forever. I've turned debug on errors on emacs by default, so that the next time it happens I might be able to get some more information.

It would be really helpful to have a channel / forum for spacemacs Rust users somewhere. I can ask for one in the official Rust Discord or Zulip channels if there is interest. 

> Ensure up to date dependencies 

I only use rust nightly and I update it often. I don't have neither racer nor the RLS installed. So all my dependencies have always been up-to-date.

> An example project that showcases issues with config/specs would be ideal for [...] Commenting out mostly configurable/optional code would be a large regression of functionality

I disagree, this PR implements a critical feature in spacemacs: the ability to edit text. 

See, with company mode, auto-completion, flycheck, etc. enabled, I cannot edit text. I type 3-4 characters, emacs freezes for 10 seconds, I type 3-4 chars again, emacs freezes for 10 seconds, etc. So what value do these tools add if they prevent me from editing Rust files? None, they actually remove all value in using spacemacs as a text editor since they prevent it to do the single thing that it should be able to do: edit text.

I've mentioned projects in which this happen in some of the issues I've opened, and the best answer I've gotten [is that my 6 year old dual core 2Ghz 8Gb RAM laptop](https://github.com/syl20bnr/spacemacs/issues/11622#issuecomment-447004994) is just not powerful enough to edit 1000 LOC files with spacemacs and that I should buy a new one or use a different editor. 

Turns out that 1-1.5 years ago, spacemacs was blazing fast, and with this PR, it is blazing fast again.

Spacemacs getting slow did not happen overnight, it has been a continuous process over a not so long period of time that continued until spacemacs couldn't do the only thing it is supposed to always be able to do: edit text.

> Don't merge this in its current state. 

I don't have time to materialize a PR that makes all broken / experimental components disabled by default and adds options to manually enable them in the Rust layer. Mainly because I don't want to use these features myself unless they preserve spacemacs ability to edit text, which they currently don't. 

I'd be ok with not merging this if such a PR materializes and gets merged instead. But I also be ok with merging this, and with those who want to keep using experimental features to add workarounds in their configs to do so if they don't have time to properly add them as opt-in to the Rust layer.

> I like the notion of opt-in behaviors but what the defaults should be is a conventions question. 

The current convention seems to be "let's enable all new features by default", this is how we've ended with me having to submit a PR that comments out 90% of the code in the spacemacs layer. 

Honestly, the Rust layer should be following the opposite convention: "no new features are enabled by default". I know spacemacs Rust layer might not have many contributors, and that requiring all features to be opt-in might turn some contributors down, but to enable a feature by default for everybody there should be a significant amount of proof that this feature does not impact spacemacs most critical task: being able to edit text. 

This is why I think a channel on Rust's Discord / Zulip for Rust spacemacs users would be a good idea. It is easy to reach many Rust users there, and before enabling a feature by default one could ask there everybody to try it and gather feedback.

The issues in this repo are not tagged, so it is almost impossible to tell how many issues are open about the Rust layer, but searching for Rust reveals 216 issues open. Such a communication channel might be a great way to attract contributors that can help bringing this number down.

jamartin9:
@gnzlbg 
Thank you for the feedback. It helps improve perspective a lot.

I think I have replicated one of the bugs you mention.
Spacemacs freezes for several seconds in the src/unix/mod.rs of the libc create line 364-365 when deleting a brace.
RLS crashes around the same spot. rls crashes with a stack overflow.
It seems to do so regarless of smartparens toggle (SPC t p).
It seems to do so with the autocomplete backend set to nil too.
It seems to do so with syntax checking off.

I am using nightly-2019-01-19-x86_64-unknown-linux-gnu rls nightly. racer built with same version.

I will dig more with the debugger when I have time. 

My low end pc is a Inspiron-11 with 2GB of RAM and Intel Celeror dual core @2GHZ. It works well enough usually so that is what im testing with. (my phone runs spacemacs too but rust isn't viable there).

It is a little noticable on my workstation i7-5960x @4GHz.

The feezing while typing you describe might be related to the autocomplete issue from gitter?

I will update when I have more results and try to address a few more of your points if I can.

gnzlbg:
Thank you for digging into this @jamartin9 I really appreciate it, I might have some time this weekend to re-enable the default rust layer locally and see if the workaround from gitter solves some of the issues for me. 

I experience these delays particularly with macros are involved. For example, in the `libc` crate or in the `slice-deque` crate, some functions have `#[cfg(...)]` or `#[inline(...)]`. When I delete, say, the closing `]`, emacs often freezes for a long time and becomes unresponsive. In the slice-deque crate (https://github.com/gnzlbg/slice_deque) I can also reproduce this by going into any function and writing an `if true {` (without a closing `}`), emacs becomes really unresponsive until I close the `}`.

jamartin9:
@gnzlbg 
I only had a little bit of time for digging but changing the variable in the spacemacs config helped immensely ( defaults to 'all ):
```
dotspacemacs-highlight-delimiters nil
```
I will dig some more when I have more time.

sdwolfz:
We'll look into this as part of a bugfix release but right now commenting out everything does look to drastic.

jamartin9:
Sounds good to me.

Hopefully something like rust-analyzer will be in a state to improve the auto-completetion by then. 

I will continue to try and look into poor performance with the defaults when  I have spare time and update my post above with any findings.

Miciah:
Does https://github.com/racer-rust/emacs-racer/commit/780074b861c3327c3830e72d81b71a4f7321122f alleviate the performance problem with auto-completion for you?

johnwalker:
No, the performance in emacs still isn't great.

Does `(syntax-checking :variables syntax-checking-enable-by-default nil)` work? I don't have that symbol defined. I'd love to be able to disable the extensions so that I can run tests and builds in the meantime.

gnzlbg:
So I tried again the default rust layer on this file (https://github.com/gnzlbg/slice_deque/blob/master/src/lib.rs) and trying to write something like:

```rust
if foo { 
   ...
```

anywhere in the file, without a closing `}` makes emacs unusably slow. After typing a character in the `...`, the character can take up to 10 seconds to appear on screen.

BigBeautifulBro:
Just commenting out the offending functionality is a bad way to address this issue. I think the better way to address this issue for the short term would be to document potential performance issues in the rust layer and how to turn off or on functionality that affects performance.

A main draw for spacemacs for me is that it packages useful integration and functionality out of the box and that you can then customize that. Users shouldn't have to be expected to look into the rust layer source files to figure out how to regain this functionality.

gnzlbg:
FWIW I did an spacemacs upgrade at the weekend, and ended up having to apply this patch again.

For some reason the rust layer was expecting racer to be installed, and trying to use it without checking, which created hundreds of file descriptors that brought spacemacs up to a halt (it couldn't write to any buffer anymore, it couldn't quit because it couldn't save the cache, etc.).

I tried disabling it but you can't do that apparently, you can only switch to use lsp instead of racer... which is pretty much as bad for any not-tiny codebase.. So I ended up disabling the whole stuff.

My main use of spacemacs is to program in Rust, but if I have to disable the Rust layer to do that, then what's the point ?

gnzlbg:
>  I think the better way to address this issue for the short term would be to document potential performance issues in the rust layer and how to turn off or on functionality that affects performance.

I'll document this as, if you want to use Rust, use something else, or do not enable the Rust layer.

> A main draw for spacemacs for me is that it packages useful integration and functionality out of the box and that you can then customize that. Users shouldn't have to be expected to look into the rust layer source files to figure out how to regain this functionality.

That's only useful if the quality of the layers is high. If you have a layer that's very poor quality, then having that layer isn't useful at all.

matthewjberger:
Please do not merge this PR. The layer itself is opt-in as-is right now and can be configured with a single line to use either racer, the rls, or rust-analyzer. I've been using this to program rust projects almost every day since @jamartin9 's original pr adding support for the rls came out and I've had zero issues with the functionality. Actively removing it because a single user is having very specific and very solvable issues would be very drastic and highly unnecessary. The quality of the layer itself is high.

gnzlbg:
FWIW I'm working on large Rust projects, and the layer has only gotten slower and slower over time.

> The layer itself is opt-in as-is right now and can be configured with a single line to use either racer, the rls, or rust-analyzer. I

The fact that it requires either of these, given that none of these is known to work on even medium-sized projects, and that does not handle the absence of any of these gracefully, trying to spawn thousands of processes when none of these are available, running out of file-descriptors, etc. Doesn't really sound like "high-quality" to me, and at least for my case, it makes spacemacs unsuitable for professional Rust work. The sad part is that the layer used to work just fine a couple of years ago.

allentiak:
@gnzlbg @duianto Just curious... Who's call is this PR being merged (or not)? What is needed for a decision regarding this to be taking? Maybe there is something I can help with...

duianto:
Another collaborator was part of the discussion.
@sdwolfz do you have insight into the state of this PR?

sdwolfz:
Sorry, I lost track of this PR and could not convince a Rust + Spacemacs user colleague to take a look and help debug.

As for reverting the change, I still think we should instead try to pinpoint the cause and fix it. I need to familiarize myself with the Rust ecosystem first and try it out locally, I'll see if I have some time over the weekend.

smile13241324:
Basically disabling all kinds of IDE features due to performance issues is not a good way to handle this. The default should reflect what people today expect from an IDE and this includes auto-completion as well as syntax-checking. If this is too slow for the respective node either the tools must be optimised or certain expensive settings must be disabled.

Wild guesses include:
* Don't auto trigger autocompletion (layer setting in auto-completion layer)
* Disable smartparens (dotfile)
* Disable flycheck either globally by disabling syntax checking layer for Rust only by removing the mode from `flycheck-global-modes` (user config)

Given that this is open since quite some time and the solution is clearly not acceptable I would like to close this rather than have it stay open forever.

With this being said I am absolutely for analysing and fixing those performance issues, just disabling everything by default is no solution at all instead disable/don't use the features which are not working for you. For performance issues, individual tickets should be opened with us or upstream depending on where time is lost. 

Still when you see issues with LSP the chance is very high that the issue is rather with the server implementation than with emacs given that there a general client is used.

lebensterben:
> Still when you see issues with LSP the chance is very high that the issue is rather with the server implementation than with emacs given that there a general client is used.

true, RLS is slow.

Given that rust-analyzer will replacing RLS, this PR is less relevant.

Issue #10520, opened 2018-03-24
Add haskell-ide-engine support


IvanMalison:
https://github.com/haskell/haskell-ide-engine and https://github.com/emacs-lsp/lsp-haskell seem to have finally reached a level of stability/ease of use that they should probably be added as an option for haskell support in spacemacs.

I made a PR to add instructions to the README for installing with emacs here https://github.com/haskell/haskell-ide-engine/pull/502

siddharthist:
I've begin work on this, my branch is available here: https://github.com/syl20bnr/spacemacs/compare/develop...siddharthist:lsp-hie?expand=1

I'm running into the following issue, which has halted my progress: https://github.com/emacs-lsp/lsp-haskell/issues/28

duianto:
The PR:
Haskell: add LSP support via HIE and lsp-haskell #11887
is now on the develop branch.

There's a comment: https://github.com/syl20bnr/spacemacs/pull/11887#issuecomment-462172870 on that PR, that states that it might fix this issue, can someone confirm?

demokritos:
I can confirm that lsp works via hie.
But it seems that line breaks are missing in the information box for symbols,
![spacemacs_lsp_hie](https://user-images.githubusercontent.com/3399860/56075663-23327b80-5e01-11e9-914e-131ab8ce0bbc.png)
Also often the box is too small to show the content.
I know that this is spacemacs's problem, but I don't know who concerns.

Compro-Prasad:
Maybe @yyoncho

yyoncho:
@demokritos I think this is lsp-ui feature, you may open lsp-ui issue so we could work on improving it.

razvan-panda:
Has this been implemented?

duianto:
@razvan-panda This PR has been merged into the `develop` branch:
Haskell: add LSP support via HIE and lsp-haskell #11887

There has not been a `master` branch release with it yet, but it's in the works.

PR #11550, opened 2018-11-04
[haskell] support dynamic completion backend

d12frosted:

PR #11887, opened 2019-01-22
Haskell: add LSP support via HIE and lsp-haskell

michaelpj:
This allows the use of the `lsp` layer as a backend for the Haskell layer along with an installation of `hie`.

The first commits are from https://github.com/syl20bnr/spacemacs/pull/11550, which performs some useful preparatory refactorings. All I did was fix the merge conflict (indentation), and get rid of the form-feed characters as requested in the review.

After that it's mostly a matter of just adding the appropriate additional hook.

The existing setup is now a little inconsistent, since we talk about the *completion* backend for the layer, whereas the `lsp` backend provides a bunch of other stuff too. However, I'm not planning to try and change that here.

### DIFF
--- a/layers/+lang/haskell/funcs.el
+++ b/layers/+lang/haskell/funcs.el
@@ -22,7 +22,58 @@
     (haskell-navigate-imports)
     (haskell-mode-format-imports)))
 
-;; Dante Functions
+
+;; Completion setup functions
+
+(defun spacemacs-haskell//setup-backend ()
+  "Conditionally setup haskell backend."
+  (pcase haskell-completion-backend
+    (`ghci (spacemacs-haskell//setup-ghci))
+    (`intero (spacemacs-haskell//setup-intero))
+    (`dante (spacemacs-haskell//setup-dante))
+    (`ghc-mod (spacemacs-haskell//setup-ghc-mod))))
+
+(defun spacemacs-haskell//setup-company ()
+  "Conditionally setup haskell backend."
+  (pcase haskell-completion-backend
+    (`ghci (spacemacs-haskell//setup-ghci-company))
+    (`intero (spacemacs-haskell//setup-intero-company))
+    (`dante (spacemacs-haskell//setup-dante-company))
+    (`ghc-mod (spacemacs-haskell//setup-ghc-mod-company))))
+
+
+;; ghci functions
+
+(defun spacemacs-haskell//setup-ghci ()
+  (interactive-haskell-mode))
+
+(defun spacemacs-haskell//setup-ghci-company ()
+  (spacemacs|add-company-backends
+    :backends (company-ghci company-dabbrev-code company-yasnippet)
+    :modes haskell-mode))
+
+
+;; ghc-mod functions
+
+(defun spacemacs-haskell//setup-ghc-mod ()
+  (ghc-init))
+
+(defun spacemacs-haskell//setup-ghc-mod-company ()
+  (spacemacs|add-company-backends
+    :backends (company-ghc company-dabbrev-code company-yasnippet)
+    :modes haskell-mode))
+
+
+;; Dante functions
+
+(defun spacemacs-haskell//setup-dante ()
+  (dante-mode)
+  (add-to-list 'spacemacs-jump-handlers-haskell-mode 'xref-find-definitions))
+
+(defun spacemacs-haskell//setup-dante-company ()
+  (spacemacs|add-company-backends
+    :backends (dante-company company-dabbrev-code company-yasnippet)
+    :modes haskell-mode))
 
 (defun spacemacs-haskell//dante-insert-type ()
   (interactive)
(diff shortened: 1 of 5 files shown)


siddharthist:
Looks like this would fix #10520

siddharthist:
Currently, it doesn't seem that the code in this PR warns the user if they forget to enable the `lsp` layer. You can find an example of how to do that here: https://github.com/syl20bnr/spacemacs/compare/develop...siddharthist:lsp-hie?expand=1#diff-87459216e7fd9ab9f3876d68e6737019R30

siddharthist:
When starting to edit a file (in a normal Cabal-based project), I get `ClassesC.hs is not part of any project. Select action:` with no "actions" provided. If I hit `TAB` (and it is not obvious that this is what I should do), I get a popup suggesting a few options. I selected the option to import the correct directory as the project root, but the server immediately exits. Checking the message buffer, I see
```
LSP :: hie has exited (finished)
Server hie:4675 status:starting exited with status exit. Do you want to restart it? (y or n) y
LSP :: Restarting LSP in buffer ClassesC.hs
Reading Haskell configuration failed with exit code 1 and output:
ghc: can't find a package database at $packageConfDir
```
which is the same issue I had when trying to set this up myself: https://github.com/emacs-lsp/lsp-haskell/issues/28

Any suggestions?

michaelpj:
I think the lack of displayed options is https://github.com/syl20bnr/spacemacs/issues/11888, which I also hit.

No idea about the other error I'm afraid - that seems like it's an issue with HIE rather than this layer?

siddharthist:
I actually think it's an issue with installing GHC via Nix... I found the string `$packageConfDir` in `nixpkgs`. Rats! Well, thanks for taking the time to set this up! I look forward to using it.

michaelpj:
Hmmm, I have it working fine inside a cabal2nix generated shell etc, so not sure what's up.

siddharthist:
@michaelpj How did you install HIE? I have that installed via my nix-shell as well, maybe that's the issue:
```nix  
  buildInputs = [
    (gpkgs.haskellPackages.ghcWithPackages (hpkgs: with hpkgs; [
      hie
    ] ++ this.buildInputs ++ this.propagatedBuildInputs))
```

michaelpj:
I used the hie-nix overlay and installed it outside my shell. Honestly, I don't really know why there's an overlay or what the difference is to what you did - I'd be interested if you figure it out!

siddharthist:
Well, after running the installation command from the `hie-nix` README:
```bash
$ nix-env -iA hies -f https://github.com/domenkozar/hie-nix/tarball/master
```
I get what seems like the same error: 
```
LSP :: Got error while processing diagnostics: can't find a package database at $packageConfDir
```
So it doesn't appear to be caused by the method of installing `hie`...

michaelpj:
Here's a toy project of mine that I've been using this on: https://github.com/michaelpj/abstract-machine-debuggers

Do you experience the same problem there? i.e. run `nix-shell; emacs src/Types.hs` see if it works.

michaelpj:
It's been several months, is there anything I can do to get this merged?

layers/+lang/haskell/funcs.el
```diff
@@ -22,15 +22,98 @@
...
+  (if (configuration-layer/layer-used-p 'lsp)
+      (progn
+        ;; The functionality we require from this is not an autoload, but rather some
+        ;; top-level code that registers a LSP server type. So we need to load it
+        ;; directly and can't rely on it being autoloaded.
+        (require 'lsp-haskell)
+        (lsp))
+    (message "`lsp' layer is not installed, please add `lsp' layer to your dotfile.")))
+
+(defun spacemacs-haskell//setup-lsp-company ()
```
yyoncho: no longer needed in latest lsp-mode, it will automatically configure lsp-mode

michaelpj: It looks like all the other layers still do this - are we sure to get the newest lsp-mode?

yyoncho: These are pretty much leftovers, there are PRs for fixing some of them (like https://github.com/syl20bnr/spacemacs/pulls/DanielRichtmann ). The version that introduced autoconfiguration is the same as the version that introduced `lsp` function.

michaelpj: Okay, I'll try and update this to do the right thing then :+1:

michaelpj: Done, seems to work!

michaelpj:
Rebased on develop, changelog added, tested with the latest version of lsp-mode.

michaelpj:
Also: I've been using this daily since I opened this PR without problems.

duianto:

michaelpj:
Thanks!

duianto:
@michaelpj I reverted the removal of the form-feed characters, because I was informed that they are used in multiple places, for example in the `layers/+lang`, `funcs.el` files (ex: `javascript` and `java`) and in the Spacemacs home buffer.

The discussion here:
https://github.com/syl20bnr/spacemacs/pull/11550#discussion_r230598804
Wasn't a request to remove them, it seems to have just been a rediscovery that the form-feed character just appears as an empty line in the github diffs.

This is just a note for the future, someone else that's searching for `form-feed`, might read this and see that they shouldn't be removed.

PR #11890, opened 2019-01-22
TravisCI build docs/site from HEAD

JAremko:
Should solve reverting  in documentation formatting commits. Should it? :thinking:


JAremko:
:sleeping:

PR #11898, opened 2019-01-24
Patch 1

jlumpe:
I'm hoping someone with more experience can weigh in on this, but just adding these two lines to my `dotspacemacs/user-config` didn't do anything on startup, only after reloading with `SPC f e R`. It works after wrapping it in `with-eval-after-load`.

### DIFF
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -974,8 +974,9 @@ If you are not convinced, then here is the snippet to revert back to the default
 =Vim + vim-surround= setup (add it to your =dotspacemacs/user-config=):
 
 #+BEGIN_SRC emacs-lisp
+(with-eval-after-load 'evil-surround
   (evil-define-key 'visual evil-surround-mode-map "s" 'evil-substitute)
-  (evil-define-key 'visual evil-surround-mode-map "S" 'evil-surround-region)
+  (evil-define-key 'visual evil-surround-mode-map "S" 'evil-surround-region))
 #+END_SRC
 
 * Evil plugins


jlumpe:
> > I'm hoping someone with more experience can weigh in on this
> 
> I don't think the above part in your commit adds any value.

I don't really know nearly enough about Emac's configuration system to know if this is specific to my own setup or not, so I don't know if this is actually a proper or required fix for everyone. It would be more appropriate to file this as an issue but the proposed changes are so simple I just went ahead and made the PR to save someone the work in case it gets accepted.

yuhan0:
Good documentation changes definitely add value! But in this case it seems like the root problem can be solved by moving the key binding definitions in the layer use-package declaration from the `:config` to the `:init` section. 

@jlumpe Can you try changing the function `spacemacs-evil/init-evil-surround` to the following and see if your config still requires wrapping in a `with-eval-after-load` form?
```emacs-lisp
(defun spacemacs-evil/init-evil-surround ()
  (use-package evil-surround
    :defer t
    :init
    (progn
      ;; `s' for surround instead of `substitute'
      ;; see motivation for this change in the documentation
      (evil-define-key 'visual evil-surround-mode-map "s" 'evil-surround-region)
      (evil-define-key 'visual evil-surround-mode-map "S" 'evil-substitute)
      (spacemacs|add-transient-hook evil-visual-state-entry-hook
        (lambda () (require 'evil-surround))
        lazy-load-evil-surround)
      (spacemacs|add-transient-hook evil-operator-state-entry-hook
        (lambda () (require 'evil-surround))
        lazy-load-evil-surround-2))
    :config
    (global-evil-surround-mode 1)))
```

smile13241324:
@jlumpe thanks for your PR, however as far as I can see from the linked issues the `with-eval-after-load` wrapping is only necessary due to a symptom of a different issue which is being addressed in #11965.

I think we should wait till the fix is merged than to adapt the documentation.

duianto:
The PR:
Don't lazy load keybindings for evil surround #11965
is now on the develop branch.
That makes the current documentation correct, and this PR won't be needed.

PR #11899, opened 2019-01-26
[fix] org-clock-jump-to-current-clock

FrancescElies:
After orgmode v9.1.12 `org-clock-resolve-clock` is no longer an
`interactive` function, this change broke bindings like `SPC a o C j`.

This pr fixes that

---
Edit 2019-03-16:
corrected error in the description, replaced `org-clock-resolve-clock` with `org-clock-jump-to-current-clock`

### DIFF
--- a/layers/+emacs/org/funcs.el
+++ b/layers/+emacs/org/funcs.el
@@ -67,3 +67,8 @@
 (defun spacemacs/org-trello-push-card ()
   (interactive)
   (org-trello-sync-card))
+
+(defun spacemacs/org-clock-jump-to-current-clock ()
+  (interactive)
+  (autoload #'org-clock-jump-to-current-clock "org-clock")
+  (org-clock-jump-to-current-clock))
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -194,7 +194,7 @@ Will work on both org-mode and any mode that accepts plain html."
         "Cg" 'org-clock-goto
         "Ci" 'org-clock-in
         "CI" 'org-clock-in-last
-        "Cj" 'org-clock-jump-to-current-clock
+        "Cj" 'spacemacs/org-clock-jump-to-current-clock
         "Co" 'org-clock-out
         "CR" 'org-clock-report
         "Cr" 'org-resolve-clocks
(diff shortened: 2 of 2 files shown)


sdwolfz:
I get the following error:
```
Debugger entered--Lisp error: (void-function org-clock-jump-to-current-clock)
  (org-clock-jump-to-current-clock)
  spacemacs/org-clock-jump-to-current-clock()
  funcall-interactively(spacemacs/org-clock-jump-to-current-clock)
  call-interactively(spacemacs/org-clock-jump-to-current-clock nil nil)
  command-execute(spacemacs/org-clock-jump-to-current-clock)
```

Having:
```
Org mode version 9.2 (9.2-43-gf9a8cc-elpaplus @ /home/sdwolf/.emacs.d/elpa/27.0/develop/org-plus-contrib-20190128/)
```

FrancescElies:
> I get the following error:
> 
> ```
> Debugger entered--Lisp error: (void-function org-clock-jump-to-current-clock)
>   (org-clock-jump-to-current-clock)
>   spacemacs/org-clock-jump-to-current-clock()
>   funcall-interactively(spacemacs/org-clock-jump-to-current-clock)
>   call-interactively(spacemacs/org-clock-jump-to-current-clock nil nil)
>   command-execute(spacemacs/org-clock-jump-to-current-clock)
> ```

That's right, this happens before calling `org-resolve-clocks`, which loads `org-clock` via autoload.
After having called `org-resolve-clocks` is loaded too.

I added a load inside the function, is that the correct way to address this problem?
https://github.com/syl20bnr/spacemacs/blob/04c4ce7886fe258b5440477b57a629d6769d40af/layers/+emacs/org/funcs.el#L71-L74

layers/+emacs/org/funcs.el
```diff
@@ -67,3 +67,8 @@
 (defun spacemacs/org-trello-push-card ()
   (interactive)
   (org-trello-sync-card))
+
+(defun spacemacs/org-clock-jump-to-current-clock ()
+  (interactive)
+  (load "org-clock")
```
smile13241324: In cases where you encounter issues with not yet autoloaded functions the proper approach is to only add the missing autoload rather than to load the entire file.

This should look like this:
```suggestion
   (autoload #'org-clock-jump-to-current-clock "org-clock")
```

FrancescElies: pr updated, thanks for the suggestion!

smile13241324:
The PR has been cherry-picked into the develop branch, you can safely delete your branch

PR #11901, opened 2019-01-26
Uncouple version from banner in spacemacs buffer

duianto:
The line in the Spacemacs home buffer that shows the Spacemacs version, Emacs
version and the Spacemacs distribution, for example: "0.300.0@26.1 (spacemacs)"
were tied to the banner. The line disappeared when the banner was hidden.


duianto:
The banner can be hidden by changing:
```
```
to
```
   dotspacemacs-startup-banner nil
```
in `.spacemacs`

smile13241324:

PR #11905, opened 2019-01-27
Add transient state for centered buffer

JAremko:
Simple transient state for shrinking and enlarging centered buffer.

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/keybindings.el
+++ b/layers/+spacemacs/spacemacs-defaults/keybindings.el
@@ -63,6 +63,7 @@
                                        ("T"   "UI toggles/themes")
                                        ("C-t" "other toggles")
                                        ("w"   "windows")
+                                       ("wc"  "centered")
                                        ("wp"  "popup")
                                        ("x"   "text")
                                        ("xa"  "align")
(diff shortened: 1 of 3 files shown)


yuhan0:
Could I suggest keeping the original 2-letter bindings for the toggles and using `SPC w >` and `SPC w <` instead as direct entry points into the transient state actions? 
I've been using that in my personal config for some time now - width adjustments usually take several keypresses, and `SPC w >>>>>` definitely seems more ergonomic than `SPC w c . ]]]]`

yuhan0:
Hmm, I didn't think about the `w` prefix being "crowded", I guess that would seem a little overwhelming to newcomers for the tradeoff of having shorter overall bindings. 

Was about to comment how `SPC w [` , `]`, `{` `}` all already put you straight into the window TS for the same reason, but then realised that was only in my personal config as well! 

I find that case really useful and literally use it daily, thinking of opening a PR but do you think it will be accepted?

```
;; spacemacs-defaults/keybindings.el

(spacemacs/set-leader-keys
  "w." 'spacemacs/window-manipulation-transient-state/body
  "w[" 'spacemacs/window-manipulation-transient-state/spacemacs/shrink-window-horizontally
  "w]" 'spacemacs/window-manipulation-transient-state/spacemacs/enlarge-window-horizontally
  "w{" 'spacemacs/window-manipulation-transient-state/spacemacs/shrink-window
  "w}" 'spacemacs/window-manipulation-transient-state/spacemacs/enlarge-window)
```

JAremko:
@yuhan0  On one hand it makes sense since you're extremely unlikely to be satisfied with a singe shrink; on the other hand,  you can press `.` to repeat the last command + it may be slightly too advanced behavior for the defaults. But I don't see it as surprising (in a bad way)  for beginners :thinking:

@sdwolfz  @syl20bnr any input :thinking:

yuhan0:
Thanks! I went ahead and made a PR: #11908 with several other improvements as well, feel free to continue the discussion there :)

PR #11906, opened 2019-01-27
Add labeled prefix for dash layer

jmromer:
Currently the "d" prefix in the which-key menu is displayed as "prefix". This patch replaces that with "docs".


smile13241324:
Thank you for contributing to Spacemacs!:+1:

PR #11908, opened 2019-01-27
Improvements to Window Manipulation Transient state

yuhan0:
- Adds `SPC w [`, `SPC w ]`, `SPC w {`, `SPC w }` leader key bindings for easier entry into transient state resizing

- Allow window transient state to be collapsed via "?" for more accurate presentation of layouts during the TS.

- Add `m`, `|` and `_` bindings to the transient state for consistency with the `SPC w` bindings

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/keybindings.el
+++ b/layers/+spacemacs/spacemacs-defaults/keybindings.el
@@ -628,22 +628,46 @@ respond to this toggle."
   (interactive "p")
   (enlarge-window delta t))
 
-(spacemacs|define-transient-state window-manipulation
-  :title "Window Manipulation Transient State"
-  :doc (concat "
+(defun spacemacs//window-manipulation-ts-toggle-hint ()
+  "Toggle the full hint docstring for the window manipulation transient-state."
+  (interactive)
+  (setq spacemacs--ts-full-hint-toggle
+        (logxor spacemacs--ts-full-hint-toggle 1)))
+
+(defun spacemacs//window-manipulation-ts-hint ()
+  "Return a condensed/full hint for the window manipulation transient state"
+  (concat
+   " "
+   (if (equal 1 spacemacs--ts-full-hint-toggle)
+       spacemacs--window-manipulation-ts-full-hint
+     (concat spacemacs--window-manipulation-ts-minified-hint
+             "  ([" (propertize "?" 'face 'hydra-face-red) "] help)"))))
+
+(spacemacs|transient-state-format-hint window-manipulation
+  spacemacs--window-manipulation-ts-minified-hint "
+Select: _w_ _h_ _j_ _k_ _l_ _0_.._9_ Move: _H_ _J_ _K_ _L_ _r_ _R_ Split: _s_ _v_ Resize: _[_ _]_ _{_ _}_")
+
+(spacemacs|transient-state-format-hint window-manipulation
+  spacemacs--window-manipulation-ts-full-hint
+  (format "\n [_?_] toggle help
  Select^^^^               Move^^^^              Split^^               Resize^^             Other^^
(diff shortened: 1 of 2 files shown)


yuhan0:
Here's a screenshot showing the changes in expanded state:
expanded
Minifed state:
minified

layers/+spacemacs/spacemacs-defaults/keybindings.el
```diff
@@ -628,22 +628,45 @@ respond to this toggle."
...
- [_o_]^^    other frame   ^^^^                  ^^                    ^^                   "
-               (if (configuration-layer/package-used-p 'golden-ratio)
-                   "[_g_] golden-ratio %`golden-ratio-mode"
-                 "")
-               "\n ^^^^                     ^^^^                  ^^                    ^^                   [_q_] quit")
+  :hint-is-doc t
+  :dynamic-hint (spacemacs//window-manipulation-ts-hint)
   :bindings
   ("q" nil :exit t)
+  ("?" spacemacs//window-manipulation-ts-toggle-hint)
```
sdwolfz: 👏

duianto:
`[?] help` isn't listed in the expanded state.

Is it enough to only show it in the minified state since it opens first and one has to press `?` to show the help? Or should all the transient state keys be listed?

Another solution could be to add it to the right of the title, and since the help popup already is open, maybe it could say something like this:
`Window Manipulation Transient State ([?] hide help)`
or
`Window Manipulation Transient State ([?] toggle help)`

duianto:
Some commands abbreviate transient state as `-ts-`, so maybe `Window Manipulation TS`, or if transient state doesn't need to be mentioned, Then it could just say `Window Manipulation`.

sdwolfz:
Can we have "TS" for the minified and "Transient State" for the other? if not "TS" for both should be enough.

duianto:
The parentheses around `([?] help)` might not be needed either, that would save 2 chars 😄

duianto:
The help key could be shown the same way as the other keys in the minified state:
```
Window Manipulation TS  Help: ? Select: w h ...
```

yuhan0:
Yeah I personally feel that "Transient State" is a little redundant, but it's in line with all the other transient state titles. It doesn't seem like the hydra titles can be changed dynamically, although there's surely a hack around it. 

@duianto I took the formatting from the current Layout and Workspaces transient states where they also use parens around `([?] help)`.

yuhan0:
Does this look alright for the expanded state?
image

duianto:
This can be the new template, and we'll fix the others based on this. They would probably also benefit from having a minified state.

yuhan0:
Layouts and Workspaces are the ones which *do* currently have a minified state, and I adapted most of the code structure from them. No strong opinions either way, up to the maintainers to make a call :)

> @sdwolfz 
> Can we have "TS" for the minified and "Transient State" for the other?

I figured out an ugly hack to change the title:
```emacs-lisp
  (setf (cadr spacemacs/window-manipulation-transient-state/hint)
        (concat (substring (cadr spacemacs/window-manipulation-transient-state/hint) 0 20)
                (propertize (if (zerop spacemacs--ts-full-hint-toggle) "Transient State" "TS")
                            'face 'spacemacs-transient-state-title-face)))
```

This can be stuck inside `spacemacs//window-manipulation-ts-toggle-hint` or the :on-enter prop of the TS, but I don't think messing with the hydra's internal representation is such a good idea.

sdwolfz:
I don't really think other TS would benefit from this. Window manipulation fits perfectly since you see exactly what the end result will be in real time.

@yuhan0 😄 let's skip it if it's a hack and go with `TS` for both.

yuhan0:
Alright, just pushed the changes :)

duianto:
> I don't really think other TS would benefit from this

👍

duianto:

PR #11913, opened 2019-01-28
python mode / lsp : Added support for Microsoft Python Language Server

cormacc:
Added support for MS Python Language Server as alternate backend.
Options are now `anaconda`, `lsp` or `lsp-ms`.

See layer README for configuration info.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2044,7 +2044,11 @@ Other:
 - Use =python-mode= for SCons script files (thanks to shanemikel)
 - Added LSP support, which can be used by enabling the =lsp= layer and setting
   the =python= layer's =python-backend= variable to =lsp=
-  (thanks to Yuan Fu and Sylvain Benner)
+  (thanks to Yuan Fu and Sylvain Benner).
+- The LSP backend can use either the =pyls= (default) server or the =mspyls=
+  (Microsoft) implementation, which may be selected by setting the
+  =python-lsp-server= layer variable to =mspyls=
+  (thanks to Cormac Cannon).
 - Added ~SPC m t l~ key binding to re-run the last test command
   (thanks to Benoit Coste).
 - Added support for breakpoints for the =trepan3k= python debugger
--- a/layers/+lang/python/packages.el
+++ b/layers/+lang/python/packages.el
@@ -43,6 +43,8 @@
         ;; packages for anaconda backend
         anaconda-mode
         (company-anaconda :requires company)
+        ;; packages for Microsoft LSP backend
+        (lsp-python-ms :requires lsp-mode)
         ))
 
 (defun python/init-anaconda-mode ()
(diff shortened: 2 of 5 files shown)


layers/+lang/python/packages.el
```diff
@@ -407,3 +408,20 @@ fix this issue."
...
+    :if (eq python-backend 'lsp-ms)
+    :ensure nil
+    :config
+
+    (if python-lsp-git-root
+      ;; Use dev version of language server checked out from github
+      (progn
+        (setq lsp-python-ms-dir
+          (expand-file-name (concat python-lsp-git-root "/output/bin/Release/")))
+        (message ( concat "lsp-python-ms: Using version at " lsp-python-ms-dir)))
```
yyoncho: more common is to use %s with message instead of concat.

cormacc: Thanks @yyoncho -- amended/pushed

cormacc:
I should get a chance to address the feedback tomorrow at some stage. @robbyoconnor re. antipattern/specifying language server, do you mean to revert to `anaconda` and `lsp` as allowed values for `backend`, and add another config variable `language-server`, values `default`, and 'microsoft' (or `nil` and `microsoft`)?

yyoncho:
> values `default`, and 'microsoft' (or `nil` and `microsoft`)?

IMO better names are the lsp-mode server ids. And just a heads up - at some point, we(lsp-mode team) will provide and option to manage the server priorities using variables so ideally, this configuration should disappear.

robbyoconnor:
@cormacc what I'm saying is specify the language server to use, not provide a separate backend if we want to change the language server that we use. So yes, I am saying to keep `anaconda` and `lsp` and specify the language server to use.

cormacc:
Incorporated review feedback and rebased

CHANGELOG.develop
```diff
@@ -2030,7 +2030,10 @@ Other:
 - Use =python-mode= for SCons script files (thanks to shanemikel)
 - Added LSP support, which can be used by enabling the =lsp= layer and setting
   the =python= layer's =python-backend= variable to =lsp=
-  (thanks to Yuan Fu and Sylvain Benner)
+  (thanks to Yuan Fu and Sylvain Benner). The LSP backend can use either the
```
duianto: Fill issue. Start on a newline below the previous thanks parentheses. With a leading `-` (dash).

layers/+lang/python/packages.el
```diff
@@ -420,3 +422,20 @@ fix this issue."
                  (eq 'yapf python-formatter))
         (add-hook 'python-mode-hook 'yapf-mode)))
     :config (spacemacs|hide-lighter yapf-mode)))
+
+;;https://github.com/andrew-christianson/lsp-python-ms
+;;Not on MELPA
```
duianto: Should this comment be removed?
@yyoncho Mentioned in a previous comment that it's on melpa now.

cormacc: Yep - had updated the layer to use the melpa version, but forgotten to strip the comment. Thanks

CHANGELOG.develop
cormacc: Had been trying to keep the python LSP stuff in a single changelog entry, but split now.

duianto:

PR #11919, opened 2019-01-29
add devdocs tool

JAremko:
A nice tool to have as a backup.  

- - - - - -

[dash layer](https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Breaders/dash/packages.el) adds keybindings under "SPC d"  without naming the prefix. So I  defined it more generally  `("d"   "documentation")`


robbyoconnor:
Yay for @JAremko and being awesome :tada:

JAremko:
For now I'll add this package, `devdocs-lookup` seems to be better (more mode -> docs mappings) but it's not on MELPA https://github.com/skeeto/devdocs-lookup/issues/2

PR #11921, opened 2019-01-29
Changelog.develop: Inline key bindings to list

duianto:
It's easier to read the key bindings as a list.


sdwolfz:
Thank you :heart:!

PR #11926, opened 2019-01-30
Fixes #10846, added epub support for Org layer.

### LINKED ISSUE #10846: PATCH: add Epub export to Org layer

vishvanath45:
Toggle variable named `org-enable-epub-support` lets user enable/disable epub
support.

This is an improvement upon existing PR #11280.

### DIFF
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -39,6 +39,7 @@
         persp-mode
         (ox-hugo :toggle org-enable-hugo-support)
         (org-trello :toggle org-enable-trello-support)
+        (ox-epub :toggle org-enable-epub-support)
         ))
 
 (defun org/post-init-company ()
@@ -715,3 +716,6 @@ Headline^^            Visit entry^^               Filter^^                    Da
         "mtdc" 'spacemacs/org-trello-pull-card
         "mtub" 'spacemacs/org-trello-push-buffer
         "mtuc" 'spacemacs/org-trello-push-card))))
+
+(defun org/init-ox-epub ()
+  (spacemacs|use-package-add-hook org :post-config (require 'ox-epub)))
--- a/layers/+emacs/org/README.org
+++ b/layers/+emacs/org/README.org
@@ -20,6 +20,7 @@
   - [[#project-support][Project support]]
   - [[#org-brain-support][Org-brain support]]
   - [[#mode-line-support][Mode line support]]
+  - [[#epub-support][Epub support]]
 - [[#key-bindings][Key bindings]]
   - [[#starting-org-mode][Starting org-mode]]
   - [[#toggles][Toggles]]
@@ -288,6 +289,16 @@ To permanently enable mode line display of org clock, add this snippet to your
 
 #+BEGIN_SRC elisp
   (setq spaceline-org-clock-p t)
+#+END_SRC
+
+** Epub support
+To install the Org exporter [[https://github.com/ofosos/ox-epub][ox-epub]] that generates e-book file format [[https://en.wikipedia.org/wiki/EPUB][epub]], set
+the variable =org-enable-epub-support= to =t=.
+
+#+BEGIN_SRC emacs-lisp
+  (setq-default dotspacemacs--configuration-layers
+                '((org :variables
+                       org-enable-epub-support t)))
 #+END_SRC
 
 * Key bindings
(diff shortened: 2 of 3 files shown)


vishvanath45:
Do I need to change `CHANGELOG.develop`? 
If Yes, will it be added in new release section?

sdwolfz:
This will be merged as part of a future release. We'll create a new section in the `CHANGELOG.develop` file after which you can write a line there and we'll merge this.

Compro-Prasad:
@duianto This seems to fix a simple issue.

duianto:

magthe:
It seems I _have_ to set this variable in order to avoid the following error on startup:

>Warning (initialization): An error occurred while loading ‘/home/magnus/.emacs.d/init.el’:
>
>Symbol's value as variable is void: org-enable-epub-support
>
>To ensure normal operation, you should investigate and remove the cause of the error in your initialization file.  Start Emacs with the ‘--debug-init’ option to view a complete error backtrace.

Setting it like this

```
     (org :variables
          org-enable-epub-support nil)
```

fixes it... is there a default missing?

duianto:
I reverted it.

duianto:
There's another PR that adds epub support to org:
Add ox-epub package for exporting org -> epub #11280
When it's applied, then the `ox-epub` package gets downloaded at restart.

The epub package wasn't downloaded with this PR.

Does the other PR do the same thing as this one?

vishvanath45:
@duianto I have made the required fixes in `layers/+emacs/org/config.el`.
Now the default value will be `nil`

duianto:
@vishvanath45 There's no breaking issue on restart now. 👍 

But the ox-epub package doesn't get downloaded on startup with this PR.

It is downloaded on startup with this PR: Add ox-epub package for exporting org -> epub #11280

That PR has this line in `org/init-ox-epub`
```
(defun org/init-ox-epub ()
  (spacemacs|use-package-add-hook org :post-config (require 'ox-epub))
  )
```
This PR uses:
```
(defun org/init-ox-epub ()
  (use-package ox-epub
    :after ox))
```

I'm not well versed in either hooks or use-package to know exactly what the difference is.

Otherwise this PR is more complete than the other one.

vishvanath45:
@duianto  Fix package installation at startup. 
This is my first time, I couldn't do it in single go. :)

duianto:
Ok now I understand what's going on, your first solution:
```
(defun org/init-ox-epub ()
  (use-package ox-epub
    :after ox))
```
was fine.

The package wasn't installed since the toggle is `nil` on startup:
```
(ox-epub :toggle org-enable-epub-support)
```
but when I added:
```
     (org :variables
          org-enable-epub-support t)
```
to the `dotspacemacs-configuration-layers` list, then the package was installed on startup (and probably when the configuration is synced `SPC f e R` as well).

So both solutions work. I don't know if one is preferred over the other.

But it seems to be fine now.

If you prefer the use-package solution over:
(spacemacs|use-package-add-hook org :post-config (require 'ox-epub))

Then you can change it back again, otherwise It seems ready to be added.

duianto:
There's a rebase issue, and it helps if the PR author adds a `changelog.develop` entry, but I can fix that before pushing. It's just something to think about for your next PR. 😄

layers/+emacs/org/packages.el
```diff
@@ -715,3 +716,6 @@ Headline^^            Visit entry^^               Filter^^                    Da
         "mtdc" 'spacemacs/org-trello-pull-card
         "mtub" 'spacemacs/org-trello-push-buffer
         "mtuc" 'spacemacs/org-trello-push-card))))
+
+(defun org/init-ox-epub ()
+  (spacemacs|use-package-add-hook org :post-config (require 'ox-epub)))
```
yuhan0: Following the convention of the other org packages, this should be defined in two functions:
```elisp
(defun org/pre-init-ox-epub ()
  (spacemacs|use-package-add-hook org :post-config (require 'ox-epub)))
(defun org/init-ox-epub ())
```

(I'm not sure of the rationale why it's done this way, it seems pretty convoluted compared to `:after org` in the use-package declaration.)

duianto: They should probably also be moved up before `(defun org/pre-init-ox-twbs ()`
so that the ox functions are groupped together.

duianto:

PR #11933, opened 2019-02-02
Replace company-capf with company-nimsuggest in nim-mode.

jjzmajic:
As described here: https://github.com/nim-lang/nim-mode/issues/159,
company-capf makes nim-mode unusable. I am replacing it with
company-nimsuggest as it is much more responsive. The reason I elected not to
leave company-capf in as a backup is that it would somehow still make it to the
car of company-backends, regardless of position at assignment, thus becoming the
default completion backend. Given that no functionality is lost this way, and
that nimsuggest is now part of the core nim installation
(https://github.com/nim-lang/nimsuggest), I do not think anything has been
sacrificed.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1556,6 +1556,8 @@ Other:
 **** Nim
 - Add key binding ~SPC m h h~ to show symbol documentation (thanks to Valts
   Liepiņš)
+- Replace =company-capf= with =company-nimsuggest= in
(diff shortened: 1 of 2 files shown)


sdwolfz:
Thank you :heart:! 

logc:
Thank you very much for this PR @jjzmajic ! I have been wanting to learn Nim for some time now, and Spacemacs freezing on any Nim file was a major drawback for me ...

:100: 🥇 🎉

PR #11942, opened 2019-02-04
Make `spacemacs/nim-compile-run` work with current buffer file.

jjzmajic:
Naming the main file `main.nim` is a `C` convention, not a `Nim` convention. In
fact, `Nimble` defaults to naming the main file after the project at
initialization. I will soon add a function that parses the Nimble configuration
for project entry point and compiles that. Until then, this is a solid
compromise between flexibility and reliability. I also added a doc-string to the
function.


smile13241324:

Issue #11896, opened 2019-01-24
Using S for evil-surround instead of s


hoyon:
#### Description :octocat:
I want to use capital S instead of lowercase s to activate evil-surround. I've read the documentation regarding this and have added the snippet
```
(evil-define-key 'visual evil-surround-mode-map "s" 'evil-substitute)
(evil-define-key 'visual evil-surround-mode-map "S" 'evil-surround-region)
```
to the end of my `dotspacemacs/user-config` function, but it has no effect. However if I then source the two lines after emacs has loaded, it then has the desired effect.

I understand the rationale behind the switch, however I also regularly use vim and the difference in keybinding is very annoying.

#### Reproduction guide :beetle:
- Start Emacs
- Add above snippet to `dotspacemacs/user-config`
- Press capital S in visual mode

*Observed behaviour:* :eyes: :broken_heart:
Substitutes characters

*Expected behaviour:* :heart: :smile:
Starts evil-surround

#### System Info :computer:
- OS: gnu/linux
- Emacs: 27.0.50
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 990290e24)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(sql lua erlang csv python themes-megapack helm
     (auto-completion :variables auto-completion-tab-key-behavior 'complete auto-completion-return-key-behavior nil)
     neotree emacs-lisp git github markdown org yaml html javascript typescript
     (haskell :variables haskell-completion-backend 'intero)
     (elm :variables elm-format-on-save t elm-sort-imports-on-save t)
     vimscript latex sml
     (c-c++ :variables c-c++-enable-clang-support t c-c++-enable-c++11 t)
     ycmd docker version-control syntax-checking
     (spell-checking :variables spell-checking-enable-by-default nil)
     deft epub
     react rust elixir phoenix ruby ruby-on-rails
     (shell :variables shell-default-shell 'eshell shell-enable-smart-eshell t shell-default-height 30 shell-default-position 'bottom)
     (ranger :variables ranger-show-preview t)
     restclient
     (sql :variables sql-capitalize-keywords t sql-capitalize-keywords-blacklist
          '("name")))
```
[...]

PR #11944, opened 2019-02-04
Fix docs for changing vim surround keybinding

hoyon:
Currently the snippet in the docs to change the vim-surround keybindings to be like in standard vim does not work. 

This PR fixes the doc to include a snippet which does work.

See #11896

### DIFF
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -974,8 +974,10 @@ If you are not convinced, then here is the snippet to revert back to the default
 =Vim + vim-surround= setup (add it to your =dotspacemacs/user-config=):
 
 #+BEGIN_SRC emacs-lisp
-  (evil-define-key 'visual evil-surround-mode-map "s" 'evil-substitute)
-  (evil-define-key 'visual evil-surround-mode-map "S" 'evil-surround-region)
+  (spacemacs|use-package-add-hook evil-surround
+    :post-config
+    (evil-define-key 'visual evil-surround-mode-map "s" 'evil-substitute)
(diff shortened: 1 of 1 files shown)


yuhan0:
I mentioned this on a previous PR #11898 , the root cause seems to be that the key bindings in `spacemacs-evil/init-evil-surround` should be defined in the `:init` section and not `:config`. Could you try editing the function to the following and see if the original snippet works?

```emacs-lisp
(defun spacemacs-evil/init-evil-surround ()
  (use-package evil-surround
    :defer t
    :init
    (progn
      ;; `s' for surround instead of `substitute'
      ;; see motivation for this change in the documentation
      (evil-define-key 'visual evil-surround-mode-map "s" 'evil-surround-region)
      (evil-define-key 'visual evil-surround-mode-map "S" 'evil-substitute)
      (spacemacs|add-transient-hook evil-visual-state-entry-hook
        (lambda () (require 'evil-surround))
        lazy-load-evil-surround)
      (spacemacs|add-transient-hook evil-operator-state-entry-hook
        (lambda () (require 'evil-surround))
        lazy-load-evil-surround-2))
    :config
    (global-evil-surround-mode 1)))
```

hoyon:
Sorry I didn't see that PR, being so poorly named. 

Yes making that change has made the original snippet work.

It seems the snippet was broken in 03786aa94a334aa1a2a15f95300751a6074dec84 when surround was changed to be lazy loaded.

I'll change this PR to correct the root bug instead of changing the snippet.

hoyon:
Closing in favour of #11965

PR #11947, opened 2019-02-05
Fix react layer: make evil matchit work and minor keybinding fix

thanhvg:
With evil-matchit, `%` will work on jsx tags too
Also correct rjsx-rename-tag-at-point keybinding otherwise it will be overridden by `lsp`


smile13241324:

PR #11948, opened 2019-02-06
Add line text object using `evil-textobj-line`.

jjzmajic:
As there are already two packages implementing this in Vim, and one in Emacs, I
figured this would be a nice addition. The sequence of `0 v g _` gets rather
tiring for selecting an inner line, and is just as useful (if not more) than the
buffer text object.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -432,6 +432,7 @@ Other:
     Kmit')
   - More robust =dotspacemacs/add-layer= implementation (thanks to Eivind Fonn)
   - Improved issue reporting (thanks to Eugene Yaremenko)
+  - Add line text object using =evil-textobj-line= (thanks to Uroš Perišić)
 
 - Fixes:
   - Fix loading of themes having dependencies (thanks to Benno Fünfstück)
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -1831,6 +1831,7 @@ Additional text objects are defined in Spacemacs:
 |---------+----------------------------|
 | ~a~     | an argument                |
 | ~g~     | the entire buffer          |
+| ~l~     | a line                     |
 | ~$~     | text between =$=           |
 | ~*~     | text between =*=           |
 | ~8~     | text between =/*= and =*/= |
--- a/layers/+spacemacs/spacemacs-evil/packages.el
+++ b/layers/+spacemacs/spacemacs-evil/packages.el
@@ -30,6 +30,7 @@
         ;; https://github.com/7696122/evil-terminal-cursor-changer/issues/8
         ;; evil-terminal-cursor-changer
         evil-tutor
+        evil-textobj-line
         (evil-unimpaired :location (recipe :fetcher local))
         evil-visual-mark-mode
         evil-visualstar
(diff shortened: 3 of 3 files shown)


duianto:
PR tested with `c` (change), `d` (delete), and `v` (select) and it works as expected:
- `il` targets the lines first non-blank char to the last non-blank char,
- `al` targets the lines first char to the last char (including whitespace at the start or end)

#### System Info :computer:
- OS: windows-nt
- Emacs: 26.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 66f7b9dda)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(autohotkey colors emacs-lisp git helm html ibuffer markdown multiple-cursors org spell-checking treemacs version-control)
```

duianto:

jjzmajic:
@duianto, I really don't mind (seriously), just curious as to what was different about your cherry-picking strategy this time as you're listes as both the `Committer` and the `Author` unlike my other commits.

I swear to God I'm not being pagro, I'm actually working on a platform that will need to rely on the author always being reflected, and I seem to have stumbled onto an edge case here that I should definitely handle.

duianto:
It was a double mistake on my part. Sorry about that.

Here's how it happened.

When a PR doesn't have a changelog, (which yours does, my first mistake was missing that, probably because I was on autopilot while doing other stuff. Not good!) 
then I download it as a patch in this example: https://github.com/syl20bnr/spacemacs/pull/11948.patch
and apply the patch to my local develop branch (which keeps the author in the commit message).
In this case the patch failed twice, so I though, hmm it's a good feature I guess I'll add the changes manually. (My second mistake, this sets me as the author. Not good x2!)

I then add the change log, (for PRs that need it (not this one)) and if needed make the title clearer (ex add the layer as [`layername`] to the front of the title.
I then push it to origin/develop.

If you depend on the Author on commits then we definitely need to fix it. Shall I revert it and we'll reopen this and rebase and merge it with the github button? Or is there some other way this could be solved?

jjzmajic:
@duianto that would be great, but if it's too much trouble just leave it as is. A missed commit or two shouldn't cause too much trouble as I'm using Spacemacs as a proof of concept rn, and I doubt anyone will go through the trouble of actually checking.

What's more important is that I figure out how to programmatically detect this in the future. It seems like step two is not something I can address (although if I'm wrong, please let me know, I'll definitely need input from project maintainers once this goes live), but I'm guessing step one happens with some frequency elsewhere as well.

So just to clarify, if step one went as planned, I would still be able to get the author from the `Author:` field? Or would I have to parse some other part of the commit message?

PS: I'll keep you posted on how the platform is going. Right now I can't share much, but once our beta rolls around, I know at least a couple of us would love to have you (Spacemacs), so I'll be in touch on Gitter ASAP.

jjzmajic:
@duianto also, don't worry about being on autopilot. From what I see you've been going through PRs like mad. You really seem to be breathing new life into Spacemacs, and I'm nothing but grateful. Keep up the good work!

duianto:
When a patch is applied (`git am`) without issues, then the Author field shows the PR authors name.

I have added `Check the commits Author Name!` to my `Before Pushing` checklist. So this will not happen again.

I don't know if anyone else is using my workflow.

PR #11949, opened 2019-02-06
Add support to start external terminal emulators in context.

jjzmajic:
Integrated terminal emulators/integrated shells get the job done for
quick-and-dirty shell commands, but a lot of people take great care to tweak
their terminal workflow exactly to their liking. This makes it possible to
switch to that workflow seamlessly, by opening their terminal emulator of
choosing in the current directory or at the project root, and even supports
working over `ssh` using `tramp`.


smile13241324:

Issue #12, opened 2014-10-03
rainbow-mode does not install properly on Windows


syl20bnr:
Produces a strange error:

```
package "rainbow-mode-" not available
```

Note the trailing dash.
Does not occur on Mac OS X (in XQuartz).

For now `rainbow-mode` is disabled.

PR #11958, opened 2019-02-09
new layer: outshine (with outorg support)

### LINKED ISSUE #5258: Request for outshine mode
It would be wonderful to have a package for [the outshine package](https://github.com/tj64/outshine) to be able to have workflows like [this](https://www.youtube.com/watch?v=kxFvpJPetaw&feature=youtu.be).
If I try to add manually `outshine` to the list of packages I don't get fontification or any functionality.

langston-barrett:
Fixes  #5258

### DIFF
--- a/layers/+emacs/outshine/README.org
+++ b/layers/+emacs/outshine/README.org
@@ -1,0 +1,39 @@
+#+TITLE: outshine layer
+
+* Table of Contents                                       :TOC_4_gh:noexport:
+- [[#description][Description]]
+  - [[#features][Features:]]
+- [[#install][Install]]
+- [[#key-bindings][Key bindings]]
+
+* Description
+This layer adds support for [[https://github.com/alphapapa/outshine][outshine]] and [[https://github.com/alphapapa/outorg][outorg]]. They will be enabled in all
+programming modes.
+
+** Features:
+- Navigate through code buffers via headings like you do with org buffers
+- Edit comments under outline headings in separate org-mode buffers
+
+* Install
+To use this configuration layer, add it to your =~/.spacemacs=. You will need to
+add =outshine= to the existing =dotspacemacs-configuration-layers= list in this
+file.
+
+* Key bindings
+
+| Key Binding   | Description                                                |
+|---------------+------------------------------------------------------------|
+| ~SPC a O S~   | Show the contents of all outline sections                  |
+| ~SPC a O .~   | Structural navigation/editing transient state              |
+| ~SPC a O g u~ | Navigate up an outline heading                             |
+| ~SPC a O g n~ | Navigate to the next outline heading                       |
+| ~SPC a O g j~ | Navigate to the next outline heading at the same level     |
+| ~SPC a O g j~ | Navigate to the previous outline heading at the same level |
+| ~SPC a O i h~ | Insert an outline heading                                  |
+| ~SPC a O I~   | imenu of outshine headlines                                |
+| ~SPC a O J~   | Move the current outline subtree down                      |
+| ~SPC a O K~   | Move the current outline subtree up                        |
+| ~SPC a O >~   | Demote the current outline subtree                         |
+| ~SPC a O <~   | Promote the current outline subtree                        |
+| ~SPC a O e~   | Edit the current heading as org                            |
+| ~SPC m c~     | Save the edits back to the original file                   |
--- a/layers/+emacs/outshine/packages.el
+++ b/layers/+emacs/outshine/packages.el
@@ -1,0 +1,78 @@
+;;; packages.el --- outshine layer packages file for Spacemacs.
+;;
+;; Copyright (c) 2012-2019 Sylvain Benner & Contributors
+;;
+;; Author: Langston Barrett <langston.barrett@gmail.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+;;; Commentary:
+
+;;; Code:
+
+(defconst outshine-packages
+  '(outshine
+    outorg))
+
+(defun outshine/init-outshine ()
+  (use-package outshine
+    :defer t
+    :init
+    (progn
+      (add-hook 'prog-mode-hook          'outline-minor-mode)
+      (add-hook 'outline-minor-mode-hook 'outshine-mode))
+    :config
+    (progn
+       (spacemacs/declare-prefix "aO" "outline")
+       (spacemacs/declare-prefix "aOg" "goto")
+       (spacemacs/declare-prefix "aOi" "insert")
+       (spacemacs/set-leader-keys
+         "aO."  'spacemacs/outshine-transient-state/body
+         "aOS"  'outline-show-all
+         "aOgu" 'outline-up-heading
+         "aOgn" 'outline-next-heading
+         "aOgj" 'outline-forward-same-level
+         "aOgk" 'outline-backward-same-level
+         "aOih" 'outline-insert-heading
+         "aOI"  'outshine-imenu
+         "aOK"  'outline-move-subtree-up
+         "aOJ"  'outline-move-subtree-down
+         "aO>"  'outline-demote
+         "aO<"  'outline-promote))
+       (spacemacs|define-transient-state outshine
+         :title "Outshine Transient State"
+         :doc "
+     Navigate headings^^^^      Move subtrees^^^^               Other^^
+     ─────────────────^^^^────  ─────────────^^^^─────────────  ─────^^─────────────
+     [_j_/_k_] down/up          [_J_/_K_] move subtree down/up  [_q_] quit
+     [_n_/_N_] next/up heading  [_>_/_<_] demote/promote        [_i_] insert heading
+     [_I_] heading imenu"
+         :bindings
+         ("q" nil :exit t)
+         ("i" outline-insert-heading :exit t)
+         ("I" outshine-imenu :exit t)
+         ;; Navigate headings
+         ("n" outline-next-heading)
+         ("N" outline-up-heading)
+         ("j" outline-forward-same-level)
+         ("k" outline-backward-same-level)
+         ;; Move headings
+         ("J" outline-move-subtree-down)
+         ("K" outline-move-subtree-up)
+         ;; Move headings
+         (">" outline-demote)
+         ("<" outline-promote))))
+
+(defun outshine/init-outorg ()
+  (use-package outorg
+    :defer t
+    :config
+    (progn
+      (spacemacs/declare-prefix "aO" "outorg")
+      (spacemacs/set-leader-keys
+        "aOe"  'outorg-edit-as-org))))
+
+;;; packages.el ends here
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -359,6 +359,7 @@ sane way, here is the complete list of changed key bindings
 - ietf (thanks to Christian Hopps and Robby O'Connor)
 - multiple-cursors (thanks to Codruț Constantin Gușoi and Thomas de Beauchêne)
 - parinfer (thanks to Tianshu Shi)
+- outshine (thanks to Langston Barrett)
 **** Readers
 - epub (thanks to Jeremy Dormitzer and André Peric Tavares)
 **** Spacemacs
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -353,6 +353,7 @@ Will work on both org-mode and any mode that accepts plain html."
         "ao#" 'org-agenda-list-stuck-projects
         "ao/" 'org-occur-in-agenda-files
         "aoa" 'org-agenda-list
+        "aoo" 'org-agenda
         "aoc" 'org-capture
         "aoe" 'org-store-agenda-views
         "aofi" 'org-feed-goto-inbox
@@ -370,7 +371,6 @@ Will work on both org-mode and any mode that accepts plain html."
 
         "aol" 'org-store-link
         "aom" 'org-tags-view
-        "aoo" 'org-agenda
         "aos" 'org-search-view
         "aot" 'org-todo-list
         ;; SPC C- capture/colors
--- a/layers/+emacs/outshine/layers.el
+++ b/layers/+emacs/outshine/layers.el
@@ -1,0 +1,12 @@
+;;; layers.el --- Org layers File for Spacemacs
+;;
+;; Copyright (c) 2012-2019 Sylvain Benner & Contributors
+;;
+;; Author: Langston Barrett <langston.barrett@gmail.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+(configuration-layer/declare-layers '(org))


layers/+emacs/outshine/packages.el
```diff
@@ -0,0 +1,65 @@
...
+    :defer t
+    :init
+    (progn
+      (add-hook 'prog-mode-hook          'outline-minor-mode)
+      (add-hook 'outline-minor-mode-hook 'outshine-mode))
+    :config
+    (progn
+       (spacemacs/declare-prefix "aoo" "outline")
+       (spacemacs/set-leader-keys
+         "aooS"  'outline-show-all
```
yuhan0: These bindings seem quite unwieldy - I don't think there's any other layer in Spacemacs which defines such long sequences of keys, the maximum I've seen is 3 characters long. 

In my local outline-related config I have these bound to the `SPC n` prefix for outli[n]e, I wonder if that's a reasonable default to have? 

```elisp
(spacemacs/set-leader-keys
  ;; visibility
  "nA" 'outline-show-all
  "nB" 'outline-show-branches
  "nS" 'outline-show-subtree
  "n TAB" 'outshine-cycle-buffer

  ;; Structural edits
  "nJ" 'outline-move-subtree-down
  "nK" 'outline-move-subtree-up
  "n<" 'outline-promote ; C-u for single heading
  "n>" 'outline-demote
  "nv" 'outline-mark-subtree
  "n RET" 'outshine-insert-heading

  ;; Outorg/Navi commands
  "ne" 'outorg-edit-as-org
  "nT" 'outshine-todo
  "nN" 'outshine-navi

  ;; Narrowing
  "nn" 'outshine-narrow-to-subtree
  "nw" 'narrow-or-widen-dwim
  "nW" 'widen
)
```

(my outline navigation commands are in a separate hydra with entry points "gj" and "gk", which may be even more of a personal preference)

layers/+emacs/outshine/README.org
```diff
@@ -0,0 +1,45 @@
...
+| ~SPC a o o g n~ | Navigate to the next outline heading                       |
+| ~SPC a o o g j~ | Navigate to the next outline heading at the same level     |
+| ~SPC a o o g j~ | Navigate to the previous outline heading at the same level |
+| ~SPC a o o i h~ | Insert an outline heading                                  |
+| ~SPC a o o J~   | Move the current outline subtree down                      |
+| ~SPC a o o K~   | Move the current outline subtree up                        |
+| ~SPC a o o >~   | Demote the current outline subtree                         |
+| ~SPC a o o <~   | Promote the current outline subtree                        |
+| ~SPC a o O e~   | Edit the current heading as org                            |
+| ~SPC m O q~     | Save the edits back to the original file                   |
```
timor: First of all thanks for the PR, I am already using it.

I think the command to save the edits back should be `SPC m c` to be consistent with other layers.

layers/+emacs/outshine/packages.el
siddharthist: `SPC n` is bound by default. I couldn't think of an appropriate prefix which is shorter, though I'd be happy to find one. `SPC a o` is already used for some org-related stuff, so I thought it seemed appropriate. 

For navigation, a transient state as you suggest might be nice.

siddharthist:
Okay, I made the keybindings one key shorter (and am happy to hear other suggestions). I also added a transient state for navigation and structural edits.

siddharthist:
I'm not really sure what's making the CI build fail.

duianto:
They are unrelated to this PR.

About the bind-map issue:
>Cannot open load file: No such file or directory, bind-map

JAremko said that it's a melpa issue.

>

About the failed org messages:
>   FAILED  test-package-install-org--package-desc-name-org-plus-contrib
   FAILED  test-package-install-org--package-desc-name-org-plus-contrib-already-installed
   FAILED  test-package-install-org--symbol-name-org-plus-contrib
   FAILED  test-package-install-org--symbol-name-org-plus-contrib-already-installed

sdwolfz said:
>That job probably needs to be re-run, sometime packages fail to install for some reason

layers/+emacs/outshine/packages.el
timor: I repeatedly find myself using the org keybindings on outorg headlines from muscle memory, only to receive a message like `<C-return> is undefined` when I want to create a new headline.

Also, I think the `SPC a` prefix is more appropriate for "applications" embedded in emacs, rather than modes.
I would argue that activating outshine/outorg is a conscious decision, thus justifying more "direct" bindings for its usage.

Would it be possible/feasible to provide the default org headline editing commands on the minor mode map, somehow with a low priority as not to disturb any major mode bindings?

timor:
Sorry if this outside the scope of this layer, but:

There is one minor annoyance in combination with `counsel-imenu` (which is called on `SPC j i` by default): All the headline entries show up as `Functions: ...`

As far as I understand the code, `counsel-imenu` makes the assumption that an item is a function definition if it has not been categorized explicitly.  I suspect the intention is to provide a sensible default for very simple imenu indexers.  Even if there are categorized entries, it will still collect all uncategorized entries into the "functions" category though.

One approach to fix this could be to change the category in outshine's `imenu-generic-expression` from nil to e.g. "Headlines", however then the entries don't have top-level semantics anymore.

I think the problem lies more in the facts that a) `counsel-imenu` has above-mentioned default behavior, and b) outshine provides it's own imenu-interface `outshine-imenu`, which tackles that problem in a different approach.

I think it would be nice if `SPC j i` could Do The Right Thing™.

siddharthist:
@timor if you have a suggestion for how to fix the behavior, I'm happy to do so. I could also add outshine's imenu as a keybinding to this layer

layers/+emacs/outshine/README.org
siddharthist: Ah, I fixed this in my commits but need to update the documentation.

timor:
> I could also add outshine's imenu as a keybinding to this layer

Maybe that is a good idea!  The ideal solution would probably require changes in multiple other packages simultaneously...

siddharthist:
Okay, I fixed the keybindings in the documentation and added keybindings for `outshine-imenu` and the transient state.

timor:
Are there any plans of integrating this?  I am using outorg on a daily basis, so it would be nice to have this upstream.

layers/+emacs/outshine/README.org
```diff
@@ -0,0 +1,47 @@
...
+| ~SPC a O g j~ | Navigate to the previous outline heading at the same level |
+| ~SPC a O i h~ | Insert an outline heading                                  |
+| ~SPC a O I~   | imenu of outshine headlines                                |
+| ~SPC a O J~   | Move the current outline subtree down                      |
+| ~SPC a O K~   | Move the current outline subtree up                        |
+| ~SPC a O >~   | Demote the current outline subtree                         |
+| ~SPC a O <~   | Promote the current outline subtree                        |
+| ~SPC a O e~   | Edit the current heading as org                            |
+| ~SPC m c~     | Save the edits back to the original file                   |
+|               |                                                            |
```
duianto:
```suggestion
```
Remove the empty key binding table row.

And if the default key binding `M-#` is acceptable for saving and closing the buffer: `outorg-edit-buffer`
(more info in the review at the bottom of `packages.el`)
then change `~SPC m c~` to `~M-#~`.

layers/+emacs/outshine/README.org
```diff
@@ -0,0 +1,47 @@
...
+
+# Use GitHub URLs if you wish to link a Spacemacs documentation file or its heading.
+# Examples:
+# [[https://github.com/syl20bnr/spacemacs/blob/master/doc/VIMUSERS.org#sessions]]
+# [[https://github.com/syl20bnr/spacemacs/blob/master/layers/%2Bfun/emoji/README.org][Link to Emoji layer README.org]]
+# If space-doc-mode is enabled, Spacemacs will open a local copy of the linked file.
```
duianto:
```suggestion
```
Remove the layer setup instructions.

layers/+emacs/outshine/packages.el
```diff
@@ -0,0 +1,91 @@
...
+;; Briefly, each package to be installed or configured by this layer should be
+;; added to `outshine-packages'. Then, for each package PACKAGE:
+;;
+;; - If PACKAGE is not referenced by any other Spacemacs layer, define a
+;;   function `outshine/init-PACKAGE' to load and initialize the package.
+
+;; - Otherwise, PACKAGE is already referenced by another Spacemacs layer, so
+;;   define the functions `outshine/pre-init-PACKAGE' and/or
+;;   `outshine/post-init-PACKAGE' to customize the package as it is loaded.
+
```
duianto: Remove the layer setup instructions.

layers/+emacs/outshine/packages.el
```diff
@@ -0,0 +1,91 @@
...
+(defun outshine/init-outshine ()
+  (use-package outshine
+    :defer t
+    :init
+    (progn
+      (add-hook 'prog-mode-hook          'outline-minor-mode)
+      (add-hook 'outline-minor-mode-hook 'outshine-mode))
+    :config
+    (progn
+       (spacemacs/declare-prefix "aO" "outline")
```
duianto:
```suggestion
       (spacemacs/declare-prefix "aO" "outline")
       (spacemacs/declare-prefix "aOg" "goto")
       (spacemacs/declare-prefix "aOi" "insert")
```
Add two prefixes: `goto` and `insert`

layers/+emacs/outshine/packages.el
```diff
@@ -0,0 +1,91 @@
...
+         "aO>"  'outline-demote
+         "aO<"  'outline-promote))
+       (spacemacs|define-transient-state outshine
+         :title "Outshine Transient State"
+         :doc "
+     Navigate headings^^^^      Move subtrees^^                 Other^^
+     ─────────────────^^^^──    ─────────────^^───────────────  ─────^^───
+     [_j_/_k_] down/up          [_J_/_K_] move subtree down/up  [_q_] quit
+     [_n_/_N_] next/up heading  [_>_/_<_] demote/promote        [_i_] insert heading
+     [_I_] heading imenu"
```
duianto:
```suggestion
Navigate headings^^^^      Move subtrees^^^^               Other^^
─────────────────^^^^────  ─────────────^^^^─────────────  ─────^^─────────────
[_j_/_k_] down/up          [_J_/_K_] move subtree down/up  [_q_] quit
[_n_/_N_] next/up heading  [_>_/_<_] demote/promote        [_i_] insert heading
[_I_]^^   heading imenu"
```
Fixed the Transient state alignment.
The `^` (caret) characters have to match the number of underscores in each column.

layers/+emacs/outshine/packages.el
```diff
@@ -0,0 +1,91 @@
...
+         :bindings
+         ("q" nil :exit t)
+         ("i" outline-insert-heading :exit t)
+         ("I" outshine-imenu :exit t)
+         ;; Navigate headings
+         ;; Navigate headings
```
duianto:
```suggestion
```
Remove a duplicate line.

layers/+emacs/outshine/packages.el
```diff
@@ -0,0 +1,91 @@
...
+         ;; Move headings
+         (">" outline-demote)
+         ("<" outline-promote))))
+
+(defun outshine/init-outorg ()
+  (use-package outorg
+    :defer t
+    :config
+    (progn
+      (spacemacs/declare-prefix "aoO" "outorg")
```
duianto:
```suggestion
      (spacemacs/declare-prefix "aO" "outorg")
```
The prefix was still referring to the PRs initial key binding prefix.

layers/+emacs/outshine/packages.el
```diff
@@ -0,0 +1,91 @@
...
+      (spacemacs/set-leader-keys
+        "aOe"  'outorg-edit-as-org)
```
duianto:
```suggestion
        "aOe"  'outorg-edit-as-org))))
```
Remove the following org mode key binding:
```
      (spacemacs/declare-prefix-for-mode 'org-mode "mO" "outorg")
        "c" 'outorg-copy-edits-and-exit))))
```
Because it takes over the (possibly often used) `org-mode` binding:
`SPC m c` (or `, c` in evil mode) which calls `org-capture`.

Maybe it's enough to use the default key binding that's listed on the first line of the buffer: `*outorg-edit-buffer*`
that's opened with: `SPC a O e`
>[ packages.el ] Exit with M-#

duianto:
When calling: `SPC a O e` `outorg-edit-as-org`

Then a warnings buffer opens with the message:
>Warning (emacs): ‘outshine-hook-function’ has been deprecated, use ‘outshine-mode’

But it seems to be working because a new buffer opens: `*outorg-edit-buffer*`

There's an open PR upstream for fixing it:
Update outshine-hook-function to outshine-mode #12
https://github.com/alphapapa/outorg/pull/12

langston-barrett:
@duianto Thanks for your detailed review! I believe I've addressed all your comments.

duianto:
What's the correct `SPC a O` key binding prefix name?
`outline`, `outorg` or `outshine`?

Most of the commands start with the prefix `outline`, only one command starts with `outorg` and one with `outshine`, but the layer is called `outshine`.

Currently in the `outshine/init-outshine` function, it first declares it as:
```elisp
       (spacemacs/declare-prefix "aO" "outline")
```
https://github.com/syl20bnr/spacemacs/pull/11958/files#diff-188ebecaf0a0a4887eab7d8b861e1c3eR29

Then in the `outshine/init-outorg` function, it declares it as:
```elisp
      (spacemacs/declare-prefix "aO" "outorg")
```
https://github.com/syl20bnr/spacemacs/pull/11958/files#diff-188ebecaf0a0a4887eab7d8b861e1c3eR74

Spacemacs seems to use the last declared name `outorg`.

---

The last listed key binding says:
`| ~SPC m c~     | Save the edits back to the original file                   |`
https://github.com/syl20bnr/spacemacs/pull/11958/files#diff-02431386c8eb3d82b14a34da90c5aaf2R39

but `SPC m c` is probably used for something else in each major mode.

Is there a specific key binding for saving the edits to the original file or does it work to just save the buffer with `SPC f s`?

If there is such a command then the key binding should be renamed, or if there isn't then that line should be removed.

---

Outshine Transient State formatting:
- To much indentation before each line, the window and buffer transient states only have one leading space
- Missing `^^`  after `[_I_]` so that the `heading imenu` text was aligned correctly.
- Moved `[_q_] quit` as the last item in the Other column, to match the other TS's

#### Before
```
     Navigate headings^^^^      Move subtrees^^^^               Other^^
     ─────────────────^^^^────  ─────────────^^^^─────────────  ─────^^─────────────
     [_j_/_k_] down/up          [_J_/_K_] move subtree down/up  [_q_] quit
     [_n_/_N_] next/up heading  [_>_/_<_] demote/promote        [_i_] insert heading
     [_I_] heading imenu"
```

#### After
```
 Navigate headings^^^^      Move subtrees^^^^               Other^^
 ─────────────────^^^^────  ─────────────^^^^─────────────  ─────^^─────────────
 [_j_/_k_] down/up          [_J_/_K_] move subtree down/up  [_i_] insert heading
 [_n_/_N_] next/up heading  [_>_/_<_] demote/promote        [_q_] quit
 [_I_]^^   heading imenu"
```

duianto:
Add a tags line below the title at the the top of readme.org

Tags are used to group the layers in: https://github.com/syl20bnr/spacemacs/blob/develop/layers/LAYERS.org

Since this layer is located in `layers/+emacs`, just like the `org` layer, then the tag probably should be the same as in the org layer:
```
#+TAGS: emacs|layer
```
https://raw.githubusercontent.com/syl20bnr/spacemacs/develop/layers/%2Bemacs/org/README.org

---

In `packages.el` these lines can be removed:
```
;;; Commentary:

;;; Code:
```

rybern:
Just wanted to say that I would find this PR really useful, it would be great if it could be merged.

layers/+emacs/outshine/README.org
```diff
@@ -0,0 +1,39 @@
...
+| ~SPC a O g j~ | Navigate to the next outline heading at the same level     |
+| ~SPC a O g j~ | Navigate to the previous outline heading at the same level |
+| ~SPC a O i h~ | Insert an outline heading                                  |
+| ~SPC a O I~   | imenu of outshine headlines                                |
+| ~SPC a O J~   | Move the current outline subtree down                      |
+| ~SPC a O K~   | Move the current outline subtree up                        |
+| ~SPC a O >~   | Demote the current outline subtree                         |
+| ~SPC a O <~   | Promote the current outline subtree                        |
+| ~SPC a O e~   | Edit the current heading as org                            |
+| ~SPC m c~     | Save the edits back to the original file                   |
```
rybern, line 39: I have the layer installed. The 'SPC m c' keybind doesn't appear to work, and I don't see where it was mapped in the diff

duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/af6b1479f359533948517a85d455189c890be1ae


**Modifications:**
- Changed the `SPC a O` prefix name to: `out(line/org/shine)`
- Hid the `outshine-mode` and `outline-minor-mode` mode line indicators:

emacs_2020-08-04_12-56-01

- Changed the readme entry for `SPC m c` to: `SPC a O c`
and bound it to the command: `outorg-copy-edits-and-exit`

PR #11959, opened 2019-02-09
org layer: add support for org-sticky-header-mode

siddharthist:
This helps you keep track of what header you're editing in longer org files


smile13241324:

PR #11961, opened 2019-02-10
fix hook value to enable evil-matchit in js layer

thanhvg:
hook value should be js2-mode-hook instead of js2-mode
Before fix  `%` is bound to `evil-jump-item`
After fix `%` is bound to `evilmi-jump-items`


duianto:

PR #11965, opened 2019-02-11
Don't lazy load keybindings for evil surround

hoyon:
The snippet to swap the keybindings for evil-surround in visual mode in the documentation [here](https://github.com/syl20bnr/spacemacs/blob/develop/doc/DOCUMENTATION.org#the-vim-surround-case) does not work. I previously made a PR to change the snippets in #11944, but as pointed out by @yuhan0, the root cause is that the keybindings are lazy loaded, and hence cannot be overridden in the user-config. 

With this PR, the snippet in the documentation works again.

### DIFF
--- a/layers/+spacemacs/spacemacs-evil/packages.el
+++ b/layers/+spacemacs/spacemacs-evil/packages.el
@@ -297,6 +297,10 @@
     :defer t
     :init
     (progn
+      ;; `s' for surround instead of `substitute'
+      ;; see motivation for this change in the documentation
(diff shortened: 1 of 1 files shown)


syl20bnr:
Thank you for the PR. Seems accurate :+1:
Can you revert your doc change as well in this PR?

Cool avatar BTW :-)

hoyon:
Do you mean the change in #11944? I don't believe that was merged so as it stands the documentation is still wrong and with this PR it will be correct again.

duianto:

PR #11967, opened 2019-02-11
Cleanup which-key names

duianto:
Moved:
- ("ai" "irc") to both the `erc` and `rcirc` layers. They don't have to be
  renamed until the layers are loaded.

- ("ay" "ipython notebook") to the `ipython-notebook` layer.

- ("p$" "projects/shell")
  ("'"  "open shell")
  ("as" "shells")
  to the `shell` layer.

- The commands with keybindings from spacemacs-bootstrap/packages.el to
  spacemacs-defaults/keybindings.el.

Removed:
- ("gd" "diff") it seems to be an old group name, there's no `SPC g d` group at
  the moment.

- ("Re" "elisp") and ("Rp" "pcre") because they have moved to:
  ("xr" "regular expressions")
  ("xre" "elisp")
  ("xrp" "pcre")

- ("xm" "move") seems to be an old move text group, the current keybindings are
  `SPC x J` and `SPC x K` which opens the Move Text Transient State.

- ("b" "persp-buffers") because `SPC b` is also renamed as ("b" "buffers") which
  is more general and not persp-mode specific.


syl20bnr:
Excellent cleaning, I love it!
Thank you! 👍 

PR #11968, opened 2019-02-11
Fix key bindings in ruby-on-rails layer

sei40kr:
Add missing prefixes.
Remove rake because it also exists in ruby layer.

### DIFF
--- a/layers/+frameworks/ruby-on-rails/packages.el
+++ b/layers/+frameworks/ruby-on-rails/packages.el
@@ -73,7 +73,10 @@
         "fRx" 'projectile-rails-extract-region)
 
       (dolist (mode '(ruby-mode enh-ruby-mode))
-        (spacemacs/declare-prefix-for-mode mode "mf" "rails/rubocop")
+        (spacemacs/declare-prefix-for-mode mode "mf" "rails")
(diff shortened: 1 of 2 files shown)


layers/+frameworks/ruby-on-rails/packages.el
```diff
@@ -64,7 +64,6 @@
         "fgz" 'projectile-rails-goto-spec-helper
         "fg." 'projectile-rails-goto-file-at-point
         ;; Rails external commands
-        "f:" 'projectile-rails-rake
```
smile13241324: I have compared the commands `projectile-rails-rake` and `rake` and noticed that the first is outputting to a `projectile-rake` specific buffer and the latter is not.

I am no ruby developer so I do not know if this makes any difference, could you quickly test this?
If they are the same the keybinding can be removed however you also need to remove the documentation for the keybinding in `ruby-on-rails` layer.

sei40kr-work: Thanks, I didn't notice. I'm making this wip not to be merged.

sei40kr: @smile13241324 
Since this PR aims to just add missing prefixes, I decided not to remove this. Thank you.

smile13241324:

PR #11969, opened 2019-02-11
Ivy: Add `SPC h d a` counsel-apropos

duianto:


yuhan0:
I think `counsel-apropos` would be the ivy equivalent here - it indexes all symbols in variable and function namespaces, not only interactive commands as `apropos-command`. (Not sure what the Helm command does)

Also I had this bound to `SPC h d a` not realising that there was a Helm equivalent, wouldn't that be a more mnemonic binding? I think of it as "describe all" or "[__a__]propos"

duianto:
### Command

Your right `counsel-apropos` is the correct command.

My confusion came from the fact that the ivy `M-x` prompt adds a `^` (caret) to the beginning of the prompt, so it searches for commands starting with `apropos`.

`apropos-command` appeared first with the keybinding `C-h a` and when it was pressed then the correct 
`34618 Search for symbol (word list or regexp):` prompt appeared.

Now when the `^` is erased from the `M-x` prompt, then `apropos` finds both `apropos-command` and 
`counsel-apropos`, and both are listed with the same keybinding `C-h a`.

`C-h k` `C-h a` shows the help info for `counsel-apropos`.

And `C-h f` `apropos-command` says:
>Its keys are remapped to ‘counsel-apropos’. Without this remapping, it would be
bound to C-h a, \<f1> a, \<help> a, \<menu-bar> \<help-menu> \<search-documentation>
\<find-commands-by-name>.

### Keybinding

This seems to be the commit that added `helm-apropos` under `SPC h d d`.
[Add `helm-apropos` on `SPC h d d` ](https://github.com/syl20bnr/spacemacs/commit/ad448d66832163b387d711e915e04c6b73e5505f)

Your right again 😄 

`SPC h d a` is a better mnemonic, for describe all.

I thought that it seemed strange when I first saw `SPC h d d` for `helm-apropos` since `a` is available and apropos starts with an `a`.

I'll open a PR for changing the `helm-apropos` binding to `SPC h d a` as well.

syl20bnr:
Oops, need to be rebased after I merged the previous one or maybe this is due to all the recent changes in the change logs.

duianto:
Np, fixed.
I'm guessing that the issue occurred because both commits changed the same line and had the same parent.

syl20bnr:
Np, merging.

syl20bnr:
Thank you ! 👍 

PR #11979, opened 2019-02-15
[typescript] feat: add lsp as a formatting tool

danielpza:
Add lsp as a formatting tool from lsp layer

### DIFF
--- a/layers/+lang/typescript/config.el
+++ b/layers/+lang/typescript/config.el
@@ -16,7 +16,7 @@
 
 (defvar typescript-fmt-tool 'tide
   "The name of the tool to be used for TypeScript source code formatting.
-Currently avaliable 'tide (default), 'typescript-formatter and 'prettier.")
+Currently available 'tide (default), 'typescript-formatter, 'prettier and 'lsp.")
 
 (defvar typescript-backend 'tide
   "The backend to use for IDE features. Possible values are `tide'
--- a/layers/+lang/typescript/README.org
+++ b/layers/+lang/typescript/README.org
@@ -49,7 +49,7 @@ If you need formatting on save:
                 typescript-fmt-on-save t)))
 #+END_SRC
 
-You can choose formatting tool:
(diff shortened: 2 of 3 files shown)


layers/+lang/typescript/config.el
```diff
@@ -16,7 +16,7 @@
 
 (defvar typescript-fmt-tool 'tide
   "The name of the tool to be used for TypeScript source code formatting.
-Currently avaliable 'tide (default), 'typescript-formatter and 'prettier.")
+Currently avaliable 'tide (default), 'typescript-formatter, 'prettier and 'lsp.")
```
tko: Might want to spell that as "available" while at it.

danielpa9708:
The only requirements are having the lsp layer, and using lsp as the backend for typescript, but yeah I agree the documentation should be updated, I had trouble setting up prettier, I was actually considering improving the documentation in these areas later.

thanhvg:
I'm not using typescript layer at all so I might be wrong but I think your pr won't do anything. Just adding `lsp` to `typescript-formatter` won't make it work. `spacemacs/typescript-format` needs change also

By default, if you enable `lsp` as backend, you will have `spc m = b` bound to `lsp-format-buffer` automatically. This applies on any layer that invokes `lsp`. You don't have to do anything

Other problem I found with this current ts layer in dev branch is `spc m =` is overshadowed by `lsp` I think we will need other PR similar to this one https://github.com/syl20bnr/spacemacs/commit/f63009ae2073f89c4b9c4500328ebbdb1b8d7fdb

danielpa9708:
@thanhvg I agree that the keybindings are wrong with the `lsp` layer, but I think it should be addressed in other PR, I was using the auto format option so it worked for me.
Also I think the fix should be in the lsp layer, it shouldn't just add the `spc m = b` keybinding because it breaks several layers including `typescript` (and `json`?). I think the convention was `spc m =` for formatting. The other layers should be the one handling the formatting and deciding to use or not `lsp` as the formatter.

Issue #11978, opened 2019-02-15
auto-evilification broken on latest magit


axelson:
None of the normal spacemacs magit keybindings are working on the latest magit. I'm pretty sure that it's due to this PR that landed today: https://github.com/magit/magit/pull/3728

It changes from [magit-popup](https://github.com/magit/magit-popup) to [transient](https://github.com/magit/transient).

bmag:
We use evil-magit for those keybindings, but it isn't updated for Transient yet, see upstream issue https://github.com/emacs-evil/evil-magit/issues/61.

In the meantime you can add `evil-magit` to `dotspacemacs-excluded-packages` in your dotspacemacs file and live with the non-evil bindings, or rollback the package update to return to an older version of Magit.

As a side note, please read our ["How to make a great bug report"](https://github.com/syl20bnr/spacemacs/wiki/Debugging#how-to-make-a-great-bug-report). If the bug wasn't so obvious, it would have been hard to figure out the problem ("broken" and "not working" only tells me there's a problem, but I usually need to know the symptoms of  the problem). Kudos for the relevant links, though.

axelson:
Thanks, noted for next time!

yuhan0:
Looks like Doom just pushed a commit to pin the magit version in the meantime: https://github.com/hlissner/doom-emacs/commit/33ef90343e156f1123d6616a00ff7458018ab8f8, perhaps Spacemacs should do the same? It's a pretty big breaking change and the fixes to evil-magit seem to be non-trivial and might take a while to sort out.

(PS. thanks for the issue labels, that really needs to be a more common practice around here!)

bmag:
@yuhan0 I thought about pinning in a similar way, but I believe we'd also need to pin Magit's several dependencies to avoid other possible issues. I'd rather fix evil-magit than do that legwork, frankly.

Silex:
I think https://github.com/syl20bnr/spacemacs/blob/master/layers/%2Bsource-control/git/packages.el needs to be modified as well.

duianto:
Searching `SPC /` for `magit.*popup` in `.emacs.d` (spacemacs develop branch) finds these entries:
```

Search at ~/.emacs.d/
CHANGELOG.develop:1219:22:  - ~SPC g f d~ for =magit-diff-buffer-file-popup= (thanks to Ag Ibragimov)
CHANGELOG.develop:1227:11:- Change =magit-svn-popup= keybinding to =~= (thanks to hornuda)
[...]

PR #11983, opened 2019-02-15
Adapt for Magit's move to Transient

bmag:
Magit now uses Transient instead of magit-popup, and we must adapt. Also had to replace the recently removed `magit-diff-buffer-file-popup` with `magit-diff-buffer-file` (similar but doesn't offer a popup before displaying the diff).

Note some upstream packages haven't adapted to this change, but when they do we might need similar fixes:
- magithub
- magit-gitflow
- magit-svn

Related: https://github.com/syl20bnr/spacemacs/issues/11978


duianto:
Could `magit-diff` be the new equivalent command?

This is what the `SPC g f d` (`magit-diff-buffer-file-popup`) popup looked like before:

emacs_2019-02-15_19-37-01

This is what `SPC g f d` (`magit-diff-buffer-file`) looks like now:

emacs_2019-02-15_19-32-26


This is what `M-x` `magit-diff` looks like now:

emacs_2019-02-15_19-54-46

And pressing `d` (for Dwim), shows a prompt: `Diff for range (magit-transient-PR):`
Pressing `RET` shows the same buffer as `SPC g f d` (second screenshot) but with a shorter buffer name `magit-diff: .emacs.d`.

duianto:
Or if `magit-diff` isn't correct, then maybe `magit-file-dispatch` (`C-c M-g`) works better.

The popup looks different:

emacs_2019-02-15_20-19-07

but pressing `d` shows the same buffer as in the second screenshot above, with the long buffer-name (that includes the current file name).

Pressing `D` from that popup shows the same popup as in the third screenshot above.

syl20bnr:
Thank you ! 💯

domenzain:
Related: https://github.com/syl20bnr/evil-escape/issues/90
Although magit is useable, with the latest `develop`, `evil-escape` fails to escape from these transient windows.

PR #11984, opened 2019-02-15
Add ranger-mode to golden-ratio-exclude-modes

siddharthist:
Ranger manages window sizing in a way that conflicts messily with golden-ratio


duianto:
Could you rebase on top of develop, the file with the `golden-ratio-exclude-modes` list has been moved.

And could you add a `changelog.develop` entry to the end of the `Distribution changes`, `Fixed:` section, the fastest way to find it is probably to search for `Layer changes and fixes`. Then add the entry on the line above that heading:

PR #11985, opened 2019-02-16
wip: Add tabnine company backend

sei40kr:
https://tabnine.com
> TabNine is the all-language autocompleter. It uses machine learning to provide responsive, reliable, and relevant suggestions.

* [x] Provide a layer option to use tabnine or not
* [x] ~Fix and refactor c/c++ layer lsp setup~
* [x] Reduce amount of completions provided by company-tabnine
* [x] ~Remove some company backends that do similar completions from default backends~
* [x] Update doc (English is not my native tongue, so it's helpful if you review my doc update and correct it)
* [x] Fix an issue when multiple backends are added at once

**Bugs**
* cquery
* ccls
* emacs-racer


sei40kr:
It almost works, but there're still bugs.
It won't work with cquery, ccls, because they don't use `spacemacs|add-company-backends`.
And emacs-racer don't use company.

layers/+completion/auto-completion/funcs.el
```diff
@@ -93,7 +93,14 @@ Available PROPS:
                  ,(format "Company backend list for %S." mode)) result)
         ;; add backends
         (dolist (backend backends)
-          (push `(add-to-list ',raw-backends-var-name ',backend) result))
+          ;; add tabnine backend if enabled
```
tko: This looks overly complicated. Would this work instead?
```elisp
(dolist (backend backends)
  (push `(add-to-list ',raw-backends-var-name ',backend) result))
(when auto-completion-enable-tabnine
  (push `(add-to-list ',raw-backends-var-name 'company-tabnine) result))
```
(Mind you I'm pretty unfamiliar with elisp.)

layers/+completion/auto-completion/packages.el
```diff
@@ -130,6 +131,13 @@
                                             "company-statistics-cache.el"))
       (add-hook 'company-mode-hook 'company-statistics-mode))))
 
+(defun auto-completion/init-company-tabnine ()
+  (use-package company-tabnine
+    :if auto-completion-enable-tabnine
+    :defer t
+    :config
+    (setq company-tabnine-max-num-results 3)))
```
tko: This seems unnecessary, unless the defaults are horrible I think they should be left as-is. And if the defaults do need adjusting I think that magic number `3` should instead be a variable, one that is used to limit other completion sources to behave consistently.

tko:
It seems the backend wants external binary to be installed in `~/TabNine` (`company-tabnine-binaries-folder`) -- I think in spacemacs that should perhaps be something like `(concat spacemacs-cache-directory "ac-tabnine/")` by default instead. It's a `defcustom` variable and I don't know what would be the right way to change its (default) value from within the layer.

sei40kr:
@tko Many thanks.

tko:
Oh, and I noticed that having tabnine enabled gave much _worse_ completion options for elisp (while editing .spacemacs or auto-completion funcs.el)

I'm only speculating, but I've a nagging feeling elisp might be special enough for emacs / other backends be able to provide better completions than tabnine, and because of that one might want to have some way to dictate for which modes tabnine should (not) be installed :-/

sei40kr:
@tko
>gave much worse completion options for elisp

I agree that.
git-complete is more useful for auto-programming since it auto-closes parenthesis.
https://github.com/zk-phi/git-complete

CarlQLange:
I don't know that just because completions were worse for elisp that this shouldn't be merged regardless! It's an option: for some users, for some use cases, it might work better, and for some it might not. Perhaps this could be reopened and looked at again?

CarlQLange:
Having merged this locally, it seems like it works really well! Well done @sei40kr :)

layers/+completion/auto-completion/README.org
```diff
@@ -83,6 +85,21 @@ The default configuration of the layer is:
 ~jk~ is a good candidate for =auto-completion-complete-with-key-sequence= if
 you don't use it already.
 
+** TabNine
+[[https://tabnine.com/][TabNine]] is the all-language autocompleter. It uses machine learning to provide
+responsive, reliable, and relevant suggestions.
+To enable TabNine backend, set the variable
+=auto-completion-enable-sort-by-usage= to =t=.
```
pepeiborra: What does `auto-completion-enable-sort-by-usage` have to do with TabNine ?

dit7ya:
Sorry for highjacking this thread. I want to know if anyone has a good implementation for using company-tabnine along with other backbends. 

To be more specific, I'm trying to use company-tabnine with company-org-roam which basically provides completions of text as linked filenames that already exist in the org-directory. 

I have not yet found a configuration that makes these two play well. I always lose some tabnine feature or the other. My knowledge of how company works is limited. Here's a reddit thread about my problem in detail, without any satisfactory solution. 

https://www.reddit.com/r/spacemacs/comments/fnnkry/how_to_set_up_multiple_company_backends_in/
Sorry once again for using this thread, I'll move it to an issue if required.

PR #11988, opened 2019-02-16
Add Prolog layer.

newres:
Aiming to add a layer for the Prolog language. The layer is setup to work with SWI-Prolog by default, but can be customized for a number of other Prologs.

### DIFF
--- a/layers/+lang/prolog/README.org
+++ b/layers/+lang/prolog/README.org
@@ -1,0 +1,77 @@
+#+TITLE: prolog layer
+
+* Table of Contents                                       :TOC_4_gh:noexport:
+- [[#description][Description]]
+  - [[#features][Features:]]
+- [[#install][Install]]
+- [[#key-bindings][Key bindings]]
+  - [[#consulting][Consulting]]
+  - [[#compiling][Compiling]]
+  - [[#formatting][Formatting]]
+  - [[#inserting][Inserting]]
+  - [[#help][Help]]
+  - [[#evaluating][Evaluating]]
+
+* Description
+This layer adds support for Prolog using the bundled Prolog mode for Emacs. In addition it also adds ediprolog support for better interaction with SWI-Prolog. 
+
+** Features:
+  - Designed for SWI-Prolog as a default, but can be used with other Prologs that Prolog mode supports. 
+  - Interactive consulting and compiling.
+  - Auto-formatting.
+  - Apropos and help lookup.
+
+    
+* Install
+To use this configuration layer, add it to your =~/.spacemacs=. You will need to
+add =prolog= to the existing =dotspacemacs-configuration-layers= list in this
+file.
+
+* Key bindings
+  
+** Consulting
+
+| Key Binding | Description       |
+|-------------+-------------------|
+| ~SPC m s b~ | Consult Buffer    |
+| ~SPC m s f~ | Consult File      |
+| ~SPC m s p~ | Consult Predicate |
+| ~SPC m s r~ | Consult Region    |
+
+** Compiling
+
+| Key Binding | Description       |
+|-------------+-------------------|
+| ~SPC m c b~ | Compile Buffer    |
+| ~SPC m c f~ | Compile File      |
+| ~SPC m c p~ | Compile Predicate |
+| ~SPC m c r~ | Compile Region    |
+| ~SPC m c c~ | Compile File      |
+
+** Formatting 
+
+| Key Binding | Description   |
+|-------------+---------------|
+| ~SPC m =~   | Indent Buffer |
+
+** Inserting
+
+| Key Binding | Description                                        |
+|-------------+----------------------------------------------------|
+| ~SPC m i m~ | Insert a Modeline for Module Specification         |
+| ~SPC m i n~ | Insert Newline and the Name for the Current Clause |
+| ~SPC m i p~ | Insert Predicate Template for the Current Clause   |
+| ~SPC m i s~ | Insert Predicate Spec                              |
+
+** Help
+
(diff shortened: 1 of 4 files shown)


layers/+lang/prolog/README.org
```diff
@@ -0,0 +1,92 @@
...
+
+    
+* Install
+To use this configuration layer, add it to your =~/.spacemacs=. You will need to
+add =prolog= to the existing =dotspacemacs-configuration-layers= list in this
+file.
+
+Note that the layer currently uses the Prolog mode that is bundled with Emacs. 
+If you want to use the separately maintained [[https://bruda.ca/_media/emacs/prolog.el][Prolog Mode by Stefan D. Bruda]]:
+- Place =prolog.el= in the =local/prolog= directory relative the layer directory.
```
smile13241324: I think requesting users to manually copy a package into a local folder is not a solution we can live with.
You should rather declare the package you wish to load in your `package.el` and provide an init function.

You could declare a layer variable and load the package only if it is set.
```
;; Would look like this
(setq auto-completion-packages
      '(
        company
        (company-quickhelp :toggle auto-completion-enable-help-tooltip)
        ))
```

layers/+lang/prolog/README.org
```diff
@@ -0,0 +1,92 @@
...
+- When defining the =prolog-packages= for the layer, change =(prolog :location built-in)=
+ to =(prolog :location local)=.
+
+* Key bindings
+  
+** Leader
```
smile13241324: Please remove the `Leader` headline and move current third level headlines to the second level please.

layers/+lang/prolog/README.org
```diff
@@ -0,0 +1,92 @@
...
+| ~SPC m h a~ | Prolog Apropos for a Given String     |
+| ~SPC m h p~ | Online Help for the Atom under Cursor |
+
+*** Evaluating 
+
+| Key Binding | Description                               |
+|-------------+-------------------------------------------|
+| ~SPC m e e~ | Context Sensitive SWI-Prolog Interaction. |
+
+# Use GitHub URLs if you wish to link a Spacemacs documentation file or its heading.
```
smile13241324: This is a comment block and should not remain in your layer file.

layers/+lang/prolog/packages.el
```diff
@@ -0,0 +1,74 @@
...
+;; Author: Newres Al Haider <newrescode@gmail.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+(defconst prolog-packages
+  '(
+    (prolog :location built-in)
```
smile13241324: As mentioned above please add the packages the user can choose from here, and toggle with a config variable. This is how the other layers handle excluding package choices.

layers/+lang/prolog/README.org
newres: Using an external prolog-mode, as opposed to the built-in one, was an option I documented in the README that I tested previously. From using the layer I never had the need to use it over the prolog mode that comes as a default with emacs, which is the one currently integrated by default into the layer. I will remove the reference to the external layer from the documentation, as it functions just as well using the built-in prolog mode.

smile13241324: Fine with me.

layers/+lang/prolog/README.org
smile13241324: Just checked your code and the block is still in there. Could you please double check?

newres: My apologies, the block should be removed now.

layers/+lang/prolog/packages.el
```diff
@@ -0,0 +1,74 @@
...
+        (spacemacs/set-leader-keys-for-major-mode 'prolog-mode
+          ;;Key Mappings
+          ;;Consulting
+          "sb" 'prolog-consult-buffer
+          "sf" 'prolog-consult-file
+          "sp" 'prolog-consult-predicate
+          "sr" 'prolog-consult-region
+          ;;Compiling
+          "cb" 'prolog-compile-buffer
+          "cf" 'prolog-compile-file
```
smile13241324: Duplicate keybinding for `cf`, please only keep `cc`.

layers/+lang/prolog/README.org
```diff
@@ -0,0 +1,77 @@
...
+
+* Description
+This layer adds support for Prolog using the bundled Prolog mode for Emacs. In addition it also adds ediprolog support for better interaction with SWI-Prolog. 
+
+** Features:
+  - Designed for SWI-Prolog as a default, but can be used with other Prologs that Prolog mode supports. 
+  - Interactive consulting and compiling.
+  - Auto-formatting.
+  - Apropos and help lookup.
+
```
smile13241324: Please remove one of the two blank lines here.

layers/+lang/prolog/README.org
```diff
@@ -0,0 +1,77 @@
...
+| ~SPC m s f~ | Consult File      |
+| ~SPC m s p~ | Consult Predicate |
+| ~SPC m s r~ | Consult Region    |
+
+** Compiling
+
+| Key Binding | Description       |
+|-------------+-------------------|
+| ~SPC m c b~ | Compile Buffer    |
+| ~SPC m c f~ | Compile File      |
```
smile13241324: Please remove this line, we should only keep `cc` binding.

smile13241324:

newres:
My apologies for not getting back on the changes sooner. Thank you for taking the time to review and merge the code!

PR #11992, opened 2019-02-18
Add init function for transient package

yuhan0:
Moves the history.el, values.el and levels.el files into the .emacs.d/.cache directory.


syl20bnr:
Looking at it.

syl20bnr:
Thank you ! 👍 

yuhan0:
Thanks! Just realised I forgot to add a `:defer t` in the use-package form for startup time optimization..

PR #11993, opened 2019-02-18
Transient and magit-svn fixes

bmag:
- magit-svn-popup renamed to magit-svn (more fallout of Transient)
- more reliable placement of Transient cache files
- also defers Transient


syl20bnr:
:+1:

PR #11996, opened 2019-02-19
Replace lsp-capabilities keybinding with lsp-describe-session

Technius:
A change in `lsp-mode` replaced `lsp-capabilities` with `lsp-describe-session`. This PR updates the corresponding `lsp` layer keybinding with the newer function and changes the keybinding from `SPC m b c` to `SPC m b d`.

Edit: Fixed typo in description

### DIFF
--- a/layers/+tools/lsp/README.org
+++ b/layers/+tools/lsp/README.org
@@ -22,7 +22,7 @@ This layer adds support for basic language server protocol packages speaking
 Different language servers may support the language server protocol to varying degrees
 and they may also provide extensions; check the language server’s website for
 details.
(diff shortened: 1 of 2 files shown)


yyoncho:
You have to update the README.org as well.

Technius:
I just updated the documentation and rebased. Let me know if I missed anything!

duianto:
@Technius The PR is missing an entry in `changelog.develop`, there's no `LSP` section at the moment, so you'll have to add it after `Lua` here:
https://github.com/syl20bnr/spacemacs/blob/develop/CHANGELOG.develop#lua

PR #12002, opened 2019-02-19
improves spacemacs-jump-handlers for clojure modes

agzam:
makes jump handlers in clojure modes to work correctly for: `, g g` `, g G` `gd` and `gD` 

previously it would annoyingly often fall-back to evil-goto-definition and ignore dumb-jump handler, which made opening definition in other-window very difficult.

It would be nice if `spacemacs/jump-to-definition` could take a prefix argument and pass it to a handler, but it doesn't. That way instead of requiring a separate keybinding could be possible to `SPC u gd` (for example to force result to be displayed in other window). However `,gG` and `gD` to open definition in other window works and I guess it is more consistent with other modes.


agzam:
Any news on this? Is anything else missing? Can this be merged?

duianto:

PR #12014, opened 2019-02-26
c++ ccls[lsp] - rename cacheDirectory cacheFormat to cache.{directory,format}

myrgy:
Align c++ ccls options passing with ccls changes (https://github.com/MaskRay/ccls/commit/2fc2d1b16ca239f4d56ced02cbdb885ef345dce1) 
>> rename cacheDirectory cacheFormat to cache.{directory,format}


Compro-Prasad:
There is also `:cacheFormat`.

myrgy:
`cacheFormat` is used only by cquery. In case of ccls binary format is used by default. And there is not cacheFormat usage for ccls. So I assume that my fix is correct.

smile13241324:

PR #12019, opened 2019-02-27
Add missing key prefixes for java-mode, gradle-mode

sei40kr:
lsp-java backend config is actually broken.
This PR fixes it and adds other missing prefixes.

### DIFF
--- a/layers/+lang/java/packages.el
+++ b/layers/+lang/java/packages.el
@@ -317,12 +317,12 @@
       (when (configuration-layer/package-used-p 'groovy-mode)
         (add-hook 'groovy-mode-hook 'gradle-mode)
         (spacemacs/declare-prefix-for-mode 'groovy-mode "ml" "gradle")
-        (spacemacs/declare-prefix-for-mode 'groovy-mode "mc" "compile")
+        (spacemacs/declare-prefix-for-mode 'groovy-mode "mlc" "compile")
         (spacemacs/declare-prefix-for-mode 'groovy-mode "mlt" "tests"))
       (when (configuration-layer/package-used-p 'java-mode)
         (add-hook 'java-mode-hook 'gradle-mode)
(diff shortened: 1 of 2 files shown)


layers/+lang/java/packages.el
```diff
@@ -438,34 +438,41 @@
       (dolist (prefix '(("mc" . "compile")
                         ("mg" . "goto")
                         ("mr" . "refactor")
-                        ("mq" . "lsp")))
```
sei40kr: This loop seems to be meaningless.

layers/+lang/java/packages.el
```diff
@@ -317,12 +317,12 @@
       (when (configuration-layer/package-used-p 'groovy-mode)
         (add-hook 'groovy-mode-hook 'gradle-mode)
         (spacemacs/declare-prefix-for-mode 'groovy-mode "ml" "gradle")
-        (spacemacs/declare-prefix-for-mode 'groovy-mode "mc" "compile")
+        (spacemacs/declare-prefix-for-mode 'groovy-mode "mlc" "compile")
         (spacemacs/declare-prefix-for-mode 'groovy-mode "mlt" "tests"))
       (when (configuration-layer/package-used-p 'java-mode)
         (add-hook 'java-mode-hook 'gradle-mode)
         (spacemacs/declare-prefix-for-mode 'java-mode "ml" "gradle")
-        (spacemacs/declare-prefix-for-mode 'groovy-mode "mc" "compile")
```
smile13241324: Just noticed that you have removed the prefix for `mc` when you were fixing the wrong mode here. However this menu is used for general compile commands which are independent of the used build system like `meghanada-compile-project` and `lsp-java-build-project`.

Please add it back:
```suggestion
       (spacemacs/declare-prefix-for-mode 'java-mode "mc" "compile")
```

smile13241324:

PR #12028, opened 2019-03-02
Hide single cursor evil-mc lighter "emc" globally

duianto:
---
### Question:

There's already an entry in the `changelog.develop`:
>  - Hide multiple cursor text when there's only one cursor (thanks to duianto)

It's referring to when the variable was first added:
init-evil-mc: 1 cursor hide mode line text variable https://github.com/syl20bnr/spacemacs/commit/aada9702dc76042b2cd2c7a02da5d961fa1e0109

Is that message enough for both the previous commit and for this PR or should there be a separate entry for this PR as well? It could say something like this:
>  - Hide single cursor =evil-mc= lighter "emc" globally (thanks to duianto)

### Possible cause
The single cursor lighter "emc" might have appeared again when the variable was moved from `:init` to `:config` in:
```
(use-package evil-mc
```

In this commit:
Encapsulates multiple cursors functionality in a layer https://github.com/syl20bnr/spacemacs/commit/5c6057226ea031a43b1a815d2ab5e08c6cb671ba

That seams to have caused the `evil-mc-one-cursor-show-mode-line-text` variable to only be set locally in the `*spacemacs*` buffer.

### DIFF
--- a/layers/+misc/multiple-cursors/packages.el
+++ b/layers/+misc/multiple-cursors/packages.el
@@ -20,7 +20,7 @@
   (use-package evil-mc
     :config
     (progn
-      (setq evil-mc-one-cursor-show-mode-line-text nil)
+      (setq-default evil-mc-one-cursor-show-mode-line-text nil)
       (when (or (spacemacs/system-is-mac) (spacemacs/system-is-mswindows))
         (setq evil-mc-enable-bar-cursor nil))
 


duianto:

Instead of adding a second line to the changelog. The current entry was made more descriptive, there are multiple multiple cursors packages, now it specifies `evil-mc`.

Compro-Prasad:
@duianto Are you maintaining the CHANGELOG by yourself because many PR authors don't change it?

duianto:
I'm adding an entry if the PR doesn't have one, it helps if the PR author has added one, but it's probably not a common practice on other repositories for the PR author to add entries to the changelog. 

And there would be some back and forth waiting for a repsonse if there are questions etc. It's just easier to add it before pushing.

The circlci checker now tests if there's a changelog.develop modification for any future PRs, but it's marked as optional, the important thing is that PR is submitted, if someone wants to add a changelog entry that's a bonus.

Compro-Prasad:
Great. So, how do you do it? By adding a new commit for each PR or something else? Is it easy to do for every PR?

duianto:
I just add an entry to `changelog.develop`, then `ce` (commit extend) it in magit, which adds it to the current commit.

A recent commit was kept unsquashed, in that case I added a separate commit with just the changelog entry.

It's not hard. These are the steps I take when I'm going to push a PR.
1. Download the PRs .patch to a local file.
2. git am apply it to my the current Spacemacs branch (`develop`)
(these two steps are automated with Autohotkey, which is a keyboard/mouse automation scripting tool)
3. Open the commit `RET`
4. Copy the PR title, and the Author lines (CopyQ handles multiple clipboard entries)
5. Open `changelog.develop`.
6. Find the location where the entry should go.
7. Paste both the PR title and Author lines
8. Add (thanks to Author) to the end of the PR title
9. Sometimes I fix the wording in the title.
10. If the commit title needs to be modified, press `ca` (commit amend) and fix it. 
If the title is fine, press `ce` (commit extend).
Then the changelog entry has been added to the current commit.
11. Check if there's been any new commits to the develop branch.
12. Push it to develop.
13. Add a thank you comment and close the PR.
14. Do a little chair dance 😆

So I guess there are a couple of steps. But it's not difficult.

If the PR has a changelog entry, is squashed, and needs no modifications, then the Github `rebase and merge` button, adds the commit to the develop branch.

PR #12035, opened 2019-03-04
clojure: add missing argument to cider-test-run-project-tests

TimoFreiberg:
Both `cider-test-run-loaded-tests` and `cider-test-run-project-tests` have 1 mandatory argument, so I added a `nil` argument

(see https://github.com/clojure-emacs/cider/blob/master/cider-test.el#L714 and https://github.com/clojure-emacs/cider/blob/master/cider-test.el#L721)

### DIFF
--- a/layers/+lang/clojure/funcs.el
+++ b/layers/+lang/clojure/funcs.el
@@ -123,19 +123,19 @@ the focus."
   "Run namespace test."
   (interactive)
   (cider-load-buffer)
-  (cider-test-run-ns-tests nil))
+  (call-interactively #'cider-test-run-ns-tests))
 
 (defun spacemacs/cider-test-run-loaded-tests ()
   "Run loaded tests."
   (interactive)
   (cider-load-buffer)
-  (cider-test-run-loaded-tests))
+  (call-interactively #'cider-test-run-loaded-tests))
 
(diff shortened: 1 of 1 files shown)


Compro-Prasad:
But, by default with no args it should be nil. Right?

TimoFreiberg:
I'm not an expert in how the `(interactive)` arguments work exactly, but it looks like my fix is necessary (or something about my setup is very broken)

Evaling `(spacemacs/cider-test-run-project-tests)` without my changes results in the following backtrace:
```
Debugger entered--Lisp error: (wrong-number-of-arguments (1 . 1) 0)
  cider-test-run-project-tests()
  spacemacs/cider-test-run-project-tests()
  eval((spacemacs/cider-test-run-project-tests) nil)
  elisp--eval-last-sexp(nil)
```

With my changes it works.

yuhan0:
Probably better to use `call-interactively` (in case the interactive form changes in the future)
```
(call-interactively 'cider-test-run-loaded-tests)
```

TimoFreiberg:
> Probably better to use `call-interactively` (in case the interactive form changes in the future)
> 
> ```
> (call-interactively 'cider-test-run-loaded-tests)
> ```

Sounds good!

duianto:

PR #12042, opened 2019-03-07
factor-layer: Misc fixes for factor version 0.98

timor:
Since factor 0.98, `fuel-edit-word` has been removed, since
`fuel-edit-word-at-point` has the same behavior when point is not on a symbol.


duianto:

PR #12043, opened 2019-03-07
Stop setting lsp-prefer-flymake to nil in LSP layer initialization

### LINKED ISSUE #11680: layers/+lang/go: golangci-lint output not visible when lsp backend is used
#### Description :octocat:
An interaction between the Go layer using LSP for code completion results in the `golangci-lint` linter not running, or at a minimum the results not being rendered in the UI. I'm able to reproduce this issue in my local project when I have:

```elisp
(go :variables
    go-tab-width 4
    go-backend 'lsp
    gofmt-command "goimports"
    go-use-golangci-lint t)
```

If I comment out the `go-backend 'lsp` line so that the default is used, everything works as-expected.

- Start Emacs
- Open a file in `.go` that has linting issues

- Lint errors are not rendered in the UI

Not seeing any clear errors in the `*Messages*` buffer:
```
Saving file /Users/theckman/go/src/stash.corp.netflix.com/core/halp2/test/mocks/mockstore/store.go...
Wrote /var/folders/f5/zjcsdkrx2bxdm32zzqfg5yp80000gn/T/gofmtho43XA.go
Calling gofmt: goimports (-srcdir /Users/theckman/go/src/stash.corp.netflix.com/core/halp2/test/mocks/mockstore/store.go -w /var/folders/f5/zjcsdkrx2bxdm32zzqfg5yp80000gn/T/gofmtho43XA.go)
Buffer is already gofmted
Wrote /Users/theckman/go/src/stash.corp.netflix.com/core/halp2/test/mocks/mockstore/store.go
```

- Lint errors should be visible in the UI.

#### System Info :computer:
- OS: darwin
- Emacs: 26.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 4912b54f2)
- Graphic display: t
- Distribution: spacemacs
- Editing style: emacs
- Completion: helm
- Layers: go lsp

```elisp
(auto-completion lsp better-defaults dash git github helm nginx osx spell-checking syntax-checking version-control html markdown nginx org yaml emacs-lisp groovy sql docker protobuf rust typescript
                 (go :variables go-tab-width 4 go-backend 'lsp gofmt-command "goimports" go-use-golangci-lint t)
                 java javascript python
                 (ruby :variables ruby-enable-enh-ruby-mode t)
```

#### Backtrace :paw_prints:
N/A

theckman:
This updates the packages.el file in the LSP layer to not set
`lsp-prefer-flymake` to `nil`. This is because when the value is `nil`,
`lsp-mode` configures `lsp-ui` as the `flycheck-checker` instead of leaving it
`nil`. This causes issues with linters not being ran / errors not being
reported.

This was specifically observed when using the `lsp` backend for the Go layer,
and opening a Go source file directly, in a new Emacs GUI process, launched from
the command line [1]. It was causing the user's configuration to not be seen
when `lsp-mode` was configured.

It may make sense to change this to be an entry in the `config.el` file, and
have it default to `:none` instead of `nil`. However, it seems to work with just
removing this line.

[1] https://github.com/emacs-lsp/lsp-ui/issues/256

Fixes #11680

Signed-off-by: Tim Heckman <t@heckman.io>

### DIFF
--- a/layers/+tools/lsp/packages.el
+++ b/layers/+tools/lsp/packages.el
@@ -22,7 +22,6 @@
     :config
     (progn
       (require 'lsp-clients)
-      (setq lsp-prefer-flymake nil)
       (spacemacs/lsp-bind-keys)
       (add-hook 'lsp-after-open-hook (lambda ()
                                        "Setup xref jump handler and declare keybinding prefixes"


theckman:
@sdwolfz I see you were the last to merge a change related to this file. I'd really appreciate your :eyes: when you have a moment.

thanhvg:
I also hate intrusive lsp-ui flycheck but I think if you remove  `(setq lsp-prefer-flymake nil)`  `lsp-prefer-flymake` will be t by default and  flymake will show up together with flycheck on mode line. Change in lsp layer will have unwanted effect on all other layers using it.

IMO it's better for each lsp consuming layer to set a local var of `lsp-prefer-flymake`. For example for go layer at `funcs.el`
```
(defun spacemacs//go-setup-backend-lsp ()
  "Setup lsp backend"
  (if (configuration-layer/layer-used-p 'lsp)
      (lsp)
    (message "`lsp' layer is not installed, please add `lsp' layer to your dotfile.")))
```
You set a local `lsp-prefer-flymake` to `:none`, which means it won't do anything with flycheck or flymake. So the change is:

```
(defun spacemacs//go-setup-backend-lsp ()
  "Setup lsp backend"
  (if (configuration-layer/layer-used-p 'lsp)
      (progn
        ;; error checking from lsp sucks, turn it off
        (setq-local lsp-prefer-flymake :none)
        (lsp))
    (message "`lsp' layer is not installed, please add `lsp' layer to your dotfile.")))
```
I used this on all my custom js-based layers to have eslint working, so far so good. Just add my approach here for discussion and learning. 

Cheer

theckman:
It seems there are a few options, yeah.

- Accept this as-is, and deal with any possible fallout
- Add the snippet suggested above in the Go layer, playing whack-a-mole with any other layers that also exhibit issues
- Add a variable to the Spacemacs LSP layer to make it possible to also set this value to whatever you want for everything, since the issue seems to be related to the order by which different things are evaluated.

thanhvg:
For me, lsp error checking for python and c++ is super awesome but it's garbage on js. So I think the preference should be set per language layer not in `lsp` itself. If we turn off flycheck on lsp layer then on language layer we need to change to provide option to turn it on anyway.

For your case, did you try this with your `dotspacemacs/user-config` something like this:
```
(defun my//turn-off-lsp-any-fly()
   (setq-local lsp-prefer-flymake :none))

(add-hook 'go-mode-local-vars-hook  #'my//turn-off-lsp-any-fly)
```
I'm not an expert so take it easy on me if it won't work :)

theckman:
@thanhvg I hadn't tried that, but I'd given similar things a try. I'm in the same boat, and have been _mostly_ firing from the hip on this one. 😄

So I dropped that in to my config and started a fresh GUI Emacs session, opening a Go file directly. Linters aren't showing like before I made this change in my local `.emacs.d` directory.

Thank you for taking the time to throw some suggestions my way. 👍

theckman:
I presented an alternative solution in #12047.

duianto:
@theckman The PR:
Set lsp-prefer-flymake to :none when Go layer is using LSP #12047
has been cherry-picked into the develop branch.

You mentioned that 12047 was an alternative solution, so If it made this PR obsolete then you can close it.

PR #12047, opened 2019-03-09
Set lsp-prefer-flymake to :none when Go layer is using LSP

### LINKED ISSUE #11680: layers/+lang/go: golangci-lint output not visible when lsp backend is used

theckman:
When switching the Go layer to use the LSP backend with `golangci-lint`, I
noticed that none of my linter errors were being rendered with my source code.
Digging in further, I eventually learned it was due to `lsp-prefer-flymake`
being set to `nil` and not `:none`.

This change updates the Go layer, when using LSP and golangci-lint, to set
`lsp-prefer-flymake` to `:none`. If this is not set, the golangci-lint errors
will not be reported.

This is an alternative to #12043.

Fixes #11680

Signed-off-by: Tim Heckman <t@heckman.io>

@sdwolfz @thanhvg @yyoncho

### DIFF
--- a/layers/+lang/go/funcs.el
+++ b/layers/+lang/go/funcs.el
@@ -37,7 +37,13 @@
 (defun spacemacs//go-setup-backend-lsp ()
   "Setup lsp backend"
   (if (configuration-layer/layer-used-p 'lsp)
-      (lsp)
+      (progn
+        ;; without setting lsp-prefer-flymake to :none
+        ;; golangci-lint errors won't be reported
+        (when go-use-golangci-lint
+          (message "[go] Setting lsp-prefer-flymake :none to enable golangci-lint support.")
+          (setq-local lsp-prefer-flymake :none))
+        (lsp))
     (message "`lsp' layer is not installed, please add `lsp' layer to your dotfile.")))
 
 (defun spacemacs//go-setup-company-lsp ()


layers/+lang/go/funcs.el
```diff
@@ -37,7 +37,9 @@
 (defun spacemacs//go-setup-backend-lsp ()
   "Setup lsp backend"
   (if (configuration-layer/layer-used-p 'lsp)
-      (lsp)
+      (progn
+        (setq-local lsp-prefer-flymake :none)
```
yyoncho: This one should not be hardcoded since if someone wants to use lsp-ui it wont be possble.

theckman: @yyoncho I think my latest commit should resolve your concern. Provide the ability to set the option as a layer variable, and have its value default to a safe value for the Go layer (`:none`).

layers/+lang/go/config.el
```diff
@@ -36,3 +36,6 @@
 
 (defvar go-use-test-args ""
   "Additional arguments to be supplied to `go test` during runtime.")
+
+(defvar go-lsp-prefer-flymake :none
```
yyoncho: IMO variable should be called go-use-lsp-ui-flychecker or something like that and should be boolean. I think you should preserve the current default behaviour - imagine that you are go layer user and you have used lsp-ui for flychecker and now after update spacemacs and you will end up with no flychecker enabled. Also, it will be consistent with the default behaviour of the other layers that use language server. 

You have to document the variable in the README.org.

theckman: @yyoncho I definitely appreciate your point. Although, in my testing it didn't break other common Spacemacs Go configs. It also fixed an issue where the layer wasn't fully configuring things properly based on what it was doing that.

Considering that, maybe instead I'll remove the configuration completely, and if you are using `lsp` and `golangci-lint` it uses `:none`. Otherwise it does nothing. The reason I think this is appropriate is that currently turning on `golangci-lint` when in `lsp` mode will silently fail.

yyoncho: Make sense, this will cover my concerns.

layers/+lang/go/funcs.el
```diff
@@ -37,7 +37,14 @@
 (defun spacemacs//go-setup-backend-lsp ()
   "Setup lsp backend"
   (if (configuration-layer/layer-used-p 'lsp)
-      (lsp)
+      (progn
+        ;; without setting lsp-prefer-flymake to :none
+        ;; golangci-lint errors won't be reported
+        (if go-use-golangci-lint
```
yyoncho: nit: you could use `(when ...` and remove the `progn`.

theckman: 👍

layers/+lang/go/config.el
theckman: 👍

thanhvg:
This is very nice. We can apply this approach to other language/framework layers using `lsp` also

chrispickard:
are there any outstanding issues with this? I'd like to see this merged, if possible. I've pulled this patch over and it's working for me

duianto:

PR #12052, opened 2019-03-11
ignore /transient

bostonaholic:
Adding `/transient` to the ignore list. It seems to have been created and added by magit when adding a dependency on `magit-transient` (https://github.com/magit/transient) with this commit https://github.com/magit/magit/commit/f3b7f7ca7dc17609c61d2528c0762292b14431ee.

`transient-history-file` can be user-customized to put this data in another place if that is a better solution.

The contents of my `./transient/history.el` is:

```
((magit-blame
  ("-w"))
 (magit-branch nil)
 (magit-commit nil)
 (magit-dispatch nil)
 (magit-pull nil)
 (magit-push nil
             ("--force-with-lease"))
 (magit-reset nil))
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,7 @@ python-*-docs-html
 /session.*
 /srecode-map.el
 /recentf
+/transient
 .mc-lists.el
 nov-places
 workspace


danjin250:
nice! Look forward to this being accepted

yuhan0:
Transient files should already be placed in .emacs.d/.cache folder, as of https://github.com/syl20bnr/spacemacs/commit/d325c5027c324f2e36a1dbbb41616da90d4abdb9. 

Perhaps your files were left over from an intermediate stage, it should be safe to just delete them.

bostonaholic:
Thanks @yuhan0. I'll delete it and close this for now. If I see it again, I'll reopen.

PR #12063, opened 2019-03-14
python: add diminish for importmagic

loysollivier:
importmangic does not have a diminish character.
Add diminish characters to make the modeline nice(r).

Signed-off-by: Loys Ollivier <loys.ollivier@gmail.com>

--
Not that I don't like importmagic but it is stressing my eyes out to see it written in the modeline when the others have nice symbols to represent them. This should restore the world to a fair and equal atmosphere.


duianto:

PR #12069, opened 2019-03-17
Increase compatibility with guix

bryan-fer:
This PR fixes issues that arise when installing spacemacs system-wide.

I am working on a spacemacs package for GNU Guix. The package works by installing the spacemacs repository to a read-only system-wide directory and loading it from site-init.el. There are 4 bugs related to this setup which are fixed in this pull request and described in more detail below/in the commit message (ordering of issue list corresponds to ordering of changes in the diff). The second and third bugs have straightforward fixes that I expect to be non-controversial. The first and fourth changes have potential issues which are discussed in their respective sections below. If you would like me to split this into multiple PRs I will, but I think it makes sense to keep the changes together since they are all related to the same scenario.

1. Variable elpa-root-directory (core-configuration-layer.el)
spacemacs tries to install elpa packages to spacemacs-start-directory. On guix spacemacs-start-directory is a store location, so we set it to user-emacs-dir. This is the location that it gets installed to on vanilla emacs.

It seems like the spacemacs code base tries to refrain from referring directly to emacs variable like user-emacs-dir, so I'm not sure if this is an appropriate solution. However, the other option I see would be to create some new variable like spacemacs-writable-directory which seems like a bigger change, and I'm trying to keep this PR as minimal as possible.

2. Template generation (core-dotspacemacs.el)
When spacemacs-install runs on first boot it copies a template file from spacemacs-start-directory which causes it to be read-only. This causes an error when spacemacs tries to edit the template based on user input. We inhibit read-only after opening the template file. This works because the file is written into dotspacemacs-directory, not to the location it was opened from.

The diff here is mostly whitespace changes. The only real change is to wrap the relevant section in the let statement at the top of the diff.

3. Variable spacemacs-start-directory
spacemacs assumes that spacemacs is installed into `$HOME/.emacs.d`. We need to update it to point to the location in the guix store. Since the variable is defined in spacemacs/core/core-load-paths.el we set it to the second directory above `load-file-name`

This change seems reasonable in the sense that it maintains behavior on current setups and allows for a greater diversity of setups without adding significant complication to the code. However, it may cause issues on setups that symlink ~/.emacs to /unexpected/path/to/spacemacs/init.el.

4. Quelpa
Some quelpa functions do not work on work on spacemacs-for-guix out of the box. When they copy files from the store the files are left as read-only. The only change is to set file permissions after copying a file or directory into Quelpa's build directory, ~/.emacs.d/.cache/quelpa/build

Setting the file modes seems safe based on my useage, but I'm not super familiar with Quelpa's build process so it would be good if someone knowledgeable in Quelpa looked over the changes to make sure. I set files to `rw-r--r--` and directories to `rwxr-xr-x`.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -120,7 +120,7 @@ subdirectory of ROOT is used."
   "Hook executed at the end of configuration loading.")
 
 (defconst configuration-layer--elpa-root-directory
-  (concat spacemacs-start-directory "elpa/")
+  (concat user-emacs-directory "elpa/")
   "Spacemacs ELPA root directory.")
 
 (defconst configuration-layer--rollback-root-directory
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -637,19 +637,20 @@ If ARG is non nil then Ask questions to the user before installing the dotfile."
     (with-current-buffer (find-file-noselect
                           (concat dotspacemacs-template-directory
                                   ".spacemacs.template"))
-      (dolist (p preferences)
-        (goto-char (point-min))
-        (re-search-forward (car p))
-        (replace-match (cadr p)))
-      (let ((install
-             (if (file-exists-p dotspacemacs-filepath)
-                 (y-or-n-p
-                  (format "%s already exists. Do you want to overwrite it ? "
-                          dotspacemacs-filepath)) t)))
-        (when install
-          (write-file dotspacemacs-filepath)
-          (message "%s has been installed." dotspacemacs-filepath)
-          t))))
+      (let ((inhibit-read-only t))
+        (dolist (p preferences)
+          (goto-char (point-min))
+          (re-search-forward (car p))
+          (replace-match (cadr p)))
+        (let ((install
+               (if (file-exists-p dotspacemacs-filepath)
+                   (y-or-n-p
+                    (format "%s already exists. Do you want to overwrite it ? "
+                            dotspacemacs-filepath)) t)))
+          (when install
+            (write-file dotspacemacs-filepath)
+            (message "%s has been installed." dotspacemacs-filepath)
+            t)))))
   (dotspacemacs/load-file)
   ;; force new wizard values to be applied
   (dotspacemacs/init))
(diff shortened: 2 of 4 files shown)


pjhenning:
Hello, any update on when this might get merged?

SaffronSnail:
Hi pjhenning. I haven't heard anything from the devs, so I don't really have an answer to your question. However, if you're interested in using spacemacs on guix right now then you can either use spacemacs normally (with this repository cloned to your .emacs.d), or you can use it as a guix package. The guix package doesn't do anything different than cloning the repository into .emacs.d, it's just nice if you prefer your system to be organized in a guixy style. To install spacemacs as a guix package download this file:

https://gitlab.com/saffronsnail/experimental-guix-packages/blob/master/gnu/packages/spacemacs.scm

The bottom of the file defines a package, emacs-spacemacs, which will install a version of emacs that loads spacemacs in site-start. I find this to be a more satisfying way to use spacemacs, but it requires building emacs locally which may take a long time depending on your machine. I recommend downloading the file instead of adding that repository as a channel because that repository is not stable. However, I've been using that version of the package for a while without having any issues.

Alternatively, you can install the package spacemacs-rolling-release and load init.el from the store in your emacs init file (I assume this works but I haven't tested it; if you try this method and run into any problems please file a bug at https://gitlab.com/saffronsnail/spacemacs-for-guix).

rationalis:
Hi @SaffronSnail it looks like your gitlab repos no longer exist or have been made private. I'd like to know if you made any further changes since this PR, and if so, would you mind sharing them? Also, do/did you still use Spacemacs's built-in package management or did you find a solution for using Guix to handle it instead? Thanks.

SaffronSnail:
They moved... I'm not actively using Spacemacs anymore but the most recent version of the patches are attached. There are comments at the top of each diff.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

Mic92:
cc @timor has done something similar for NixOS: https://github.com/timor/timor-overlay/blob/master/pkgs/spacemacs/default.nix

Mic92:
I think the code lives here: https://github.com/timor/spacemacs/commits/nixos-adjustments-dev-test

smile13241324:
I think its save to say that nothing will happen here anymore given the age of this ticket.

Therefore I am closing this now.

Mic92:
This is how this was solved in doom-emacs with nix: https://github.com/vlaci/nix-doom-emacs

Issue #11935, opened 2019-02-02
Request: Use org-re-reveal instead of stale org-reveal


jmhammond:
[Melpa now includes](https://github.com/melpa/melpa/pull/5989#issuecomment-459890924) org-re-reveal, which is an actively developed fork of org-reveal. 

Org-reveal has not been updated in a very long time, and currently breaks org 9.2. [Many](https://github.com/yjwen/org-reveal/pull/361) [pull](https://github.com/yjwen/org-reveal/pull/362) [requests](https://github.com/yjwen/org-reveal/pull/364) have been made to fix it, but none have been acknowledged.

[Org-re-reveal](https://gitlab.com/oer/org-re-reveal) was created to address these concerns, and Jens Lechtenbörger has gotten the project included in Melpa. The variable `org-enable-reveal-js-support` should cause spacemacs to include org-re-reveal.

ezchi:
It will be good if there's a way to override the package receipt defined in the layers. Then we don't need to do pull request for every single broken package.

dangom:
No need to wait for PRs.

1. set `org-enable-reveal-js-support` to `nil`
2. Add `org-re-reveal` to additional packages
3. `(use-package org-re-reveal :after org)`

lesteve:
I think this issue can be closed since it has been fixed in the `develop` branch. #12072 has been cherry-picked in the commit 9e3340b8c389aa36954293bf34c9cc3e65ad59e7.

duianto:
Thanks for the heads up.
I think the issues that are labeled with `fixed in develop` will be closed when they are merged into the master branch.

tko:
FWIW it seems there's been bunch of recent activity in the original https://github.com/yjwen/org-reveal repo, though I've no idea how the two compare.

PR #12072, opened 2019-03-18
Use org-re-reveal, since it works

magthe:
Adressing #11935

Just a naive `s/ox-reveal/org-re-reveal`, but it seems to work.


duianto:

PR #12081, opened 2019-03-20
core/core-spacemacs-buffer: move up macro definition

timor:
The macro `spacemacs-buffer||notes-adapt-caption-to-width` was used before it was defined.


duianto:

PR #12093, opened 2019-03-22
Improve purescript keybinding discoverability & ergonomics.

Jason94:
Added prefix titles and removed some useless nesting to the keybindings of the purescript layer.

I'm pretty new to Spacemacs, and even newer to customizing layers, so let me know if there's a better way to declare prefixes for major-mode keybindings.

### DIFF
--- a/layers/+lang/purescript/packages.el
+++ b/layers/+lang/purescript/packages.el
@@ -39,7 +39,7 @@
     (progn
       (add-hook 'purescript-mode-hook 'turn-on-purescript-indentation)
       (add-hook 'purescript-mode-hook 'purescript-decl-scan-mode)
-      (spacemacs/declare-prefix "mg" "jump")
+      (spacemacs/declare-prefix "mg" "goto")
       (spacemacs/declare-prefix "mi" "imports")
       (spacemacs/set-leader-keys-for-major-mode 'purescript-mode
         "i="  'purescript-mode-format-imports
@@ -54,7 +54,7 @@
     (progn
       (spacemacs/register-repl 'psci 'psci "purescript")
       (add-hook 'purescript-mode-hook 'inferior-psci-mode)
-      (spacemacs/declare-prefix "ms" "psci")
+      (spacemacs/declare-prefix "ms" "repl")
       (spacemacs/set-leader-keys-for-major-mode 'purescript-mode
         "'"  'psci
         "sb" 'psci/load-current-file!
@@ -74,7 +74,7 @@
       (customize-set-variable 'psc-ide-rebuild-on-save purescript-enable-rebuild-on-save)
 
       (add-to-list 'spacemacs-jump-handlers-purescript-mode 'psc-ide-goto-definition)
-      (spacemacs/declare-prefix "mmi" "imports")
+      (spacemacs/declare-prefix "mmi" "insert/import")
       (spacemacs/set-leader-keys-for-major-mode 'purescript-mode
         "mt"  'psc-ide-add-clause
         "mc"  'psc-ide-case-split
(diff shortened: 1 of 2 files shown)


layers/+lang/purescript/packages.el
```diff
@@ -71,17 +74,18 @@
       (customize-set-variable 'psc-ide-rebuild-on-save purescript-enable-rebuild-on-save)
 
       (add-to-list 'spacemacs-jump-handlers-purescript-mode 'psc-ide-goto-definition)
+      (spacemacs/declare-prefix "mmi" "imports")
```
yuhan0: Was this left in by mistake? There's no "mmi" prefix

layers/+lang/purescript/packages.el
```diff
@@ -52,6 +54,7 @@
     (progn
       (spacemacs/register-repl 'psci 'psci "purescript")
       (add-hook 'purescript-mode-hook 'inferior-psci-mode)
+      (spacemacs/declare-prefix "ms" "psci")
```
yuhan0: "ms" is the "repl" prefix by convention across other layers, even though the actual backends are diffferent

layers/+lang/purescript/packages.el
```diff
@@ -39,6 +39,8 @@
     (progn
       (add-hook 'purescript-mode-hook 'turn-on-purescript-indentation)
       (add-hook 'purescript-mode-hook 'purescript-decl-scan-mode)
+      (spacemacs/declare-prefix "mg" "jump")
```
yuhan0: "mg" is generally "goto" in other layers, I think you missed out on "mh" = "help" too

layers/+lang/purescript/packages.el
```diff
@@ -71,17 +74,18 @@
       (customize-set-variable 'psc-ide-rebuild-on-save purescript-enable-rebuild-on-save)
 
       (add-to-list 'spacemacs-jump-handlers-purescript-mode 'psc-ide-goto-definition)
+      (spacemacs/declare-prefix "mmi" "imports")
       (spacemacs/set-leader-keys-for-major-mode 'purescript-mode
         "mt"  'psc-ide-add-clause
-        "mcs" 'psc-ide-case-split
+        "mc"  'psc-ide-case-split
```
yuhan0: Make sure to update the purescript/README.org to keep it in sync with these changes :)

layers/+lang/purescript/packages.el
Jason94: We don't have any 'help' commands in the purescript layer at the moment. Should I define it anyway in case we add some?

Jason94: There was one (`psc-ide-show-type`), but I moved it from `mht` to `mt` in this commit. I could also move it back, and give `mh` the 'help' prefix.

yuhan0: Oh right, I didn't notice that got shifted as well. There's a "mt" prefix = "tests" convention in other language layers although I'm not sure how applicable that is in Purescript. 
http://spacemacs.org/doc/CONVENTIONS#tests

It would also be helpful to have a `"mhh"` binding for "K" in evil normal state to bring up language-specific documentation, again not sure what that maps to in the Purescript packages.

Jason94: Ah, thanks for the link. I moved "mt" back to "mht" because that seems like the pretty standard location for that command. 

The purescript compiler does have a documentation generator ala javadoc or something, but I don't believe the emacs plugins have any way to show that.

layers/+lang/purescript/packages.el
Jason94: Done. Also reorganized the table of keybindings so the nesting is a little more apparent.

duianto:
Thanks for contributing!

Issue #3203, opened 2015-09-30
Find a solution to the indent level pain point


syl20bnr:
Is there any package to address this paint point ?
Do we implement something from scratch in Spacemacs ?
I guess an alist `mode . indent-level` would be the most convenient.

TheBB:
Editorconfig has [this](https://github.com/editorconfig/editorconfig-emacs/blob/master/editorconfig.el#L68-L109) alist, which is fairly complete. I uesd it in my [evil-shift-width](https://github.com/TheBB/spacemacs-layers/blob/master/evil-shift-width/config.el#L6-L35) layer to set that varible depending on the current mode's indentation settings. I think this is very close to a solution, but of course we can't get something that is both general and able to cover the advanced settings in a mode such as Haskell.

TheBB:
I don't know how the UI for such a feature would look. I want to stress that we make it possible to opt-out of using this in certain modes, because some of them have very thorough and detailed indentation settings that we might not want to mess with (unless asked).

syl20bnr:
I wonder if we could use the layer abstraction instead of the package. I believe I see this proposition pop out in the chat. Would make sense to have `(layer . indent-level)`.

syl20bnr:
Raising it as high priority because I can see in google that it is a very recurrent paint point for users.

aaronjensen:
I came across https://github.com/jscheid/dtrt-indent which seems to work well and is pretty slick. It could likely be extended to support most major modes.

a13ph:
Indentation style wars are second only to modal vs non-modal editing arguments. Well, maybe block style is up there too, but I've never heard anyone asking about that one.

And even if some projects use some special convention for indentation, most people probably have some preferred setting (their default for personal files etc), which can be asked about once easily - or be presented in homepage? toggles?... config file, I guess? Since many similar non-toggles go there. 

Having specific settings for particular layers/mode/languages in one place would be nice and would probably require just finding all those disparate variables, so that'd be great to add. Not sure how hard it would be to make possible project-specific overrides though, but that does seem like up EditorConfig alley

andschwa:
For what it's worth, [dtrt-indent](https://github.com/jscheid/dtrt-indent) now supports setting generic indentation variables as well, implemented in jscheid/dtrt-indent#23, mostly so I could automatically set `evil-shift-width`. It's very extensible and the maintainer is great.

DanielFGray:
[...]

Issue #7939, opened 2016-12-11
Proposed layer/package: quicksetindent


ghost:
One of Emacs' greatest strengths, the per-mode indentation settings, is also a great weakness. It's complicated to set up unified indentation across all of your filetypes. Newcomers have to scour the net for old, partial lists of what variables to use for configuring each mode, and where and how to put those configurations. And they end up with partial configurations where only a few of their modes are affected.

Spacemacs is Emacs for the masses. So perhaps I can help this situation.

I compiled an up-to-date list of all variables, and the great thing about these is that they're all pretty standardized by now. New modes usually stick to using one of these. So this list should stay universal for a long time.

I then put it all into a package which lets you easily set the indentation variables either globally or locally per-buffer, with a single command.

So somebody who wants a global default of 4 spaces, no tabs, would just add the following three lines to their dotspacemacs user-config:

```emacs-lisp
  (require 'quicksetindent)
  (quicksetindent/set-indentation t 4)
  (setq-default php-mode-coding-style 'pear)
```

The first line wouldn't be necessary if this was made into a layer or package. And the last line is only necessary for people who work in php-mode (which always applies a "style" at startup, which sets the indentation; the source code description for my package has more info).

I placed all of this into the public domain, so anyone is welcome to do whatever you want with it: Hack it apart, put it into a layer (to make its usage in Spacemacs even easier), submit it to MELPA, or do anything else you feel like. ;-)

Enjoy:
https://github.com/SteveJobzniak/emacs-quicksetindent/blob/master/quicksetindent.el

syl20bnr:
After thinking about it I believe that the best option for the integration of this great list of defaults is to turn it into a layer. Several advantages to do so:

1) we leverage all the mechanics provided by layers, especially `:disabled-for`/`:enabled-for` keywords which allow to include or exclude the effect of a layer onto the others. Also `post/pre-init` functions will be handy to handle the specific cases. 
2) we can introduce layer variables to name the settings.
3) we can document it in a layer README accessible easily with `SPC h SPC`
 
There are several way to do it, more or less "spacemacsy".

1) The first one, less spacemacsy is to keep the current structure of @SteveJobzniak package.
2) The second one, more spacemacsy is to define the defaults in `post-init` function for each language.

I would like to do a mix of both actually.

[...]

PR #12101, opened 2019-03-27
Add dtrt-indent layer for automatic indentation detection

keevindoherty:
This layer is a simple wrapper around the [dtrt-indent](https://github.com/jscheid/dtrt-indent) Emacs package. It automatically detects the indentation style in a file and sets defaults to match.

This solution is based on [this suggestion](https://github.com/syl20bnr/spacemacs/issues/3203#issuecomment-264175032).

This is essentially just a package with an added hook, though, and as per the guidelines:
> Layer with no associated configuration will be rejected. For instance a layer with just a package and a hook can be easily replaced by the usage of the variable dotspacemacs-additional-packages.

That said, this has been an open issue since 2015 (see #3203), so I thought maybe this could be helpful to people.

### DIFF
--- a/layers/+misc/dtrt-indent/README.org
+++ b/layers/+misc/dtrt-indent/README.org
@@ -1,0 +1,24 @@
+#+TITLE: dtrt-indent layer
+
+Spacemacs layer wrapping the Emacs package [[https://github.com/jscheid/dtrt-indent][dtrt-indent]].
+This package enables automatic detection and switching of indentation style to match the current file.
+See the [[https://github.com/jscheid/dtrt-indent][dtrt-indent]] repo for more details.
+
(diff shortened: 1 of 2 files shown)


Compro-Prasad:
A name like `detect-indent` is more easy to understand.

keevindoherty:
Also related: #7939

duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/bf78bd4ec5474bcfc4c7df1597cd0088e7d1d3c4

### Changes
- Removed the package location recipe, because it caused an error on startup.
And the package is available on melpa. https://melpa.org/#/dtrt-indent
- Hid the mode line indicator
- Removed the new layer template instructions.
- Added a changelog.develop entry

PR #12122, opened 2019-04-04
add go-run-args

ezchi:
When executing the main function which requires command line arguments, user can
set `go-run-args` to pass command line arguments to compiled binary.

The example below demonstrates how to pass command line arguments by setting
`go-run-args` as file local variable:

package main

import (
	"fmt"
	"os"
	"strconv"
)

func main() {
	//Atoi converts a string to an int
	fmt.Println("Arguments:", os.Args)
	a, _ := strconv.Atoi(os.Args[1])
	b, _ := strconv.Atoi(os.Args[2])

	result := sum(a, b)
	fmt.Printf("The sum of %d and %d is %d\n", a, b, result)
}

func sum(a, b int) int {
	return a + b
}

// Local Variables:
// go-run-args: "10 5"
// End:

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -804,6 +804,7 @@ Other:
     (thanks to Boris Buliga)
   - Cleanup =which-key= by removing obsolete entries and adding some missing
     ones (thans to duianto)
+  - Add =go-run-args= to pass command line arguments to `go run`
 - Fixed:
   - Fixed ~h~ key binding in compilation and grep buffers (thanks to Sylvain
     Benner)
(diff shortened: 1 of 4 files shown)


ezchi:
New commit includes:

- update README.org by adding description for `go-run-args` variable
- fix merge conflict
- update changelog by adding commit summary

I noticed some checks failed and have no idea why my commit could cause those tests failed.

smile13241324:
@JAremko hmm I have some very strange results from CI here which seem to be unrelated to the changes done. Would you mind to take a look?

JAremko:
@smile13241324 Looks like usual dependency install problem.  :thinking:

smile13241324:

PR #12130, opened 2019-04-05
Improve & apply conventions to multi-term keybinds

emil-r-petersen:
Hi,

I think this change will be positive, as it adheres to the Navigation conventions when users are in hybrid or vim mode.

It also adds shortcuts for term-line-mode and term-char-mode, which are handy for navigating terminal output to both editing styles.

### DIFF
--- a/layers/+tools/shell/README.org
+++ b/layers/+tools/shell/README.org
@@ -221,12 +221,14 @@ in the current buffer instead of a popup.
 
 ** Multi-term
 
-| Key binding | Description                  |
-|-------------+------------------------------|
-| ~SPC m c~   | create a new multi-term      |
-| ~SPC m n~   | go to next multi-term        |
-| ~SPC m p~   | go to previous multi-term    |
-| ~SPC p $ t~ | run multi-term shell in root |
+| Key binding            | Description                    |
+|------------------------+--------------------------------|
+| ~SPC m c~              | create a new multi-term        |
+| ~SPC m C~              | switch multi-term char mode    |
+| ~SPC m l~              | switch multi-term to line mode |
+| ~SPC m n~              | go to next multi-term          |
+| ~SPC m N~ or ~SPC m p~ | go to previous multi-term      |
+| ~SPC p $ t~            | run multi-term shell in root   |
 
 ** Eshell
 
--- a/layers/+tools/shell/packages.el
+++ b/layers/+tools/shell/packages.el
@@ -165,8 +165,11 @@
       ;; multi-term commands to create terminals and move through them.
       (spacemacs/set-leader-keys-for-major-mode 'term-mode
         "c" 'multi-term
-        "p" 'multi-term-prev
-        "n" 'multi-term-next))))
+        "C" 'term-char-mode
+        "l" 'term-line-mode
+        "n" 'multi-term-next
+        "N" 'multi-term-prev
+        "p" 'multi-term-prev))))
 
 (defun shell/pre-init-org ()
   (spacemacs|use-package-add-hook org
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2008,6 +2008,10 @@ Other:
 - Wrap 'shell' command to start in current buffer (thanks to Valts Liepiņš)
 - Fixed shell popup broken on macOS (thanks to Daniel Rivas Perez)
 - Fixed broken leader key binding for inferior shell (thanks to Valts Liepiņš)
+- Added =term-mode= bindings (Thanks to Emil Petersen):
+  - ~SPC m C~ switch multi-term to char mode
+  - ~SPC m l~ switch multi-term to line mode
+  - ~SPC m N~ go to previous multi-term
 **** Shell Scripts
 - Added new company-shell environment variable backend (thanks to
   Alexander-Miller)


emil-r-petersen:
Looking at the test output, it seems to me like it is failing due to a dependency install failure.

I will try force pushing again to see if it reruns the tests.

duianto:
The issue is not on your end, a maintainer said that:
>This usually happens when Emacs packages can't be retrieved (temporary )

emil-r-petersen:
@duianto Do you approve of the changes? I'm not sure what is required on my end to get a merge and/or reasoned rejection.

layers/+tools/shell/packages.el
```diff
@@ -163,10 +163,21 @@
...
       ;; multi-term commands to create terminals and move through them.
-      (spacemacs/set-leader-keys-for-major-mode 'term-mode
-        "c" 'multi-term
-        "p" 'multi-term-prev
-        "n" 'multi-term-next))))
+      (if (member dotspacemacs-editing-style '(vim hybrid))
+          (progn
+            (spacemacs/set-leader-keys-for-major-mode 'term-mode
+              "c" 'multi-term
+              "N" 'multi-term-prev
```
sdwolfz: No need to split this based on editing style, you can have both `N` and `p` bound to `multi-term-prev` for all 3.

layers/+tools/shell/README.org
```diff
@@ -221,12 +223,26 @@ in the current buffer instead of a popup.
...
 ** Multi-term
 
-| Key binding | Description                  |
-|-------------+------------------------------|
-| ~SPC m c~   | create a new multi-term      |
-| ~SPC m n~   | go to next multi-term        |
-| ~SPC m p~   | go to previous multi-term    |
-| ~SPC p $ t~ | run multi-term shell in root |
+
+*** Hybrid & Vim mode
```
duianto: Remove the separate modes, just list the bindings individually with these two next to each other:
```
| ~SPC m N~   | go to previous multi-term      |
| ~SPC m p~   | go to previous multi-term      |
```

Under the `** Multi-term` heading.

duianto:
And add a `changelog.develop` entry to the end of the shell section:

emil-r-petersen:
@sdwolfz @duianto I addressed your comments - Thank you for reviewing the PR.

emil-r-petersen:
@duianto I believe you will find the PR in a good state now.

layers/+tools/shell/README.org
```diff
@@ -221,12 +221,14 @@ in the current buffer instead of a popup.
...
-| ~SPC m p~   | go to previous multi-term    |
-| ~SPC p $ t~ | run multi-term shell in root |
+| Key binding            | Description                    |
+|------------------------+--------------------------------|
+| ~SPC m c~              | create a new multi-term        |
+| ~SPC m n~              | go to next multi-term          |
+| ~SPC m N~ or ~SPC m p~ | go to previous multi-term      |
+| ~SPC p $ t~            | run multi-term shell in root   |
+| ~SPC m l~              | switch multi-term to line mode |
+| ~SPC m C               | switch multi-term char mode    |
```
duianto: Missing `~` (tilde) after `~SPC m C`. The corrected line looks like this:
```
| ~SPC m C~               | switch multi-term char mode    |
```

It was part of the previous commits, but I also missed it.

The issue was discovered by clicking on the `View file` button (which leads to: https://github.com/syl20bnr/spacemacs/blob/5ccb08b3ed2221624a599c244bf4ead7607327fa/layers/%2Btools/shell/README.org) near the top right corner of the diff container for the file ` layers/+tools/shell/README.org`.

On the `Files changed` tab (https://github.com/syl20bnr/spacemacs/pull/12130/files)

chrome_2019-04-14_12-31-12

layers/+tools/shell/packages.el
```diff
@@ -165,6 +165,9 @@
       ;; multi-term commands to create terminals and move through them.
       (spacemacs/set-leader-keys-for-major-mode 'term-mode
         "c" 'multi-term
+        "C" 'term-char-mode
+        "l" 'term-line-mode
+        "N" 'multi-term-prev
         "p" 'multi-term-prev
         "n" 'multi-term-next))))
```
duianto: Sort the keys alphabetically like in the readme: with `n` before `N`, which results in the three last lines looking like this:
```
        "n" 'multi-term-next
        "N" 'multi-term-prev
        "p" 'multi-term-prev))))
```

emil-r-petersen:
@duianto - I sorted all the changes alphabetically, and made them appear in the same order across files.

I do wonder why they are listed as "SPC m <k>" when the way I invoke the menu inside of multi-term is M-RET. Even assuming this is a shortcut for "the mode menu" or similar, I found it strange that it was "SPC M" as I saw it noted elsewhere in the documentation that this was reserved for users... Just thinking aloud, there's probably an answer.

While I have your intention, what's the perceived value of having the list items alphabet sorted? Just curious.

duianto:
The convention is to list key bindings starting with `SPC` when they are accessible from the Spacemacs root.

`SPC m` is the major mode prefix. It's also accessible with just `,` (comma) with the Evil (vim) editing style.
https://github.com/syl20bnr/spacemacs/blob/develop/doc/CONVENTIONS.org#major-mode-prefix

It seems like `M-RET` is another prefix for the major mode bindings. At least in some modes, for example `org-mode`.

`SPC o` and `SPC m o` are prefixes that are reserved for user bindings.
https://github.com/syl20bnr/spacemacs/blob/develop/doc/CONVENTIONS.org#user-prefix

The other layers have listed the keys in alphabetical order since I started using Spacemacs.

And here's a rule stating that keys should be listed in alphabetical order, with lower before uppercase:
https://github.com/syl20bnr/spacemacs/blob/develop/doc/CONVENTIONS.org#evilified-buffers

It's also easier to see if a key binding is talken if they are in alphabetical order.

duianto:

emil-r-petersen:
@duianto Thank you for your time to review my PR and answer my questions - It is much appreciated.

PR #12137, opened 2019-04-07
Add eval-thing-at-point functions to Common Lisp Layer

Isimoro:
(This is the same pull request as #11738; I've messed the old one up, accidentally pushed the wrong branch. 
Sorry for the mess won't happen again.)

Add functions equivalent to the custom eval-functions of the elisp layer to the
common-lisp layer.

The functions are copied from the elisp layer only `eval-last-sexp' was replaced
with its slime equivalent `slime-eval-last-expression'.

Keybindings are chosen as they are in the elisp layer.

### DIFF
--- a/layers/+lang/common-lisp/funcs.el
+++ b/layers/+lang/common-lisp/funcs.el
@@ -32,3 +32,50 @@
   (interactive)
   (move-end-of-line 1)
   (slime-eval-last-expression))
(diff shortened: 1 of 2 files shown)


smile13241324:

I have changed your commit to include a changelog entry, next time please enter one yourself.

PR #12146, opened 2019-04-10
Add `flycheck-package' to `emacs-lip' layer for writing Emacs packages.

jjzmajic:
For those with MELPA on their mind. It's easier to work when packaging errors
are reported on the fly, and we don't even have `package-lint` integration for
occasional checkups yet.

It adds no unnecessary verbosity as it is only triggered by `Package-Requires`
and `Package-Version` headers.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1198,6 +1198,7 @@ Other:
 - Disabled nameless by default (thanks to Codruț Constantin Gușoi, Sylvain
   Benner)
 - Activate nameless only in GUI (thanks to Sylvain Benner)
+- Add =flycheck-package= for package metadata linting (thanks to Uroš Perišić)
 **** Emoji
 - Added support for Emoji fonts on macOS and Linux (thanks to CodeFalling )
 - Setup =emojify= cache directory (thanks to Boris Buliga)
--- a/layers/+lang/emacs-lisp/README.org
+++ b/layers/+lang/emacs-lisp/README.org
@@ -25,6 +25,7 @@ always be in your dotfile, it is not recommended to uninstall it.
 ** Features:
 - Auto-completion using company
 - Linting using flycheck integration
+- Linting package file metadata using [[https://github.com/purcell/flycheck-package][flycheck-package]]
 - Repl support via =IELM=
 - Support for specific lisp navigation styles via =emacs-lisp-mode=
 - Auto-compile via [[https://github.com/tarsius/auto-compile][auto-compile]] package
--- a/layers/+lang/emacs-lisp/packages.el
+++ b/layers/+lang/emacs-lisp/packages.el
@@ -22,6 +22,7 @@
         evil-cleverparens
         eval-sexp-fu
         flycheck
+        flycheck-package
         ggtags
         counsel-gtags
         helm-gtags
@@ -251,6 +252,12 @@
   ;; Make flycheck recognize packages in loadpath
   ;; i.e (require 'company) will not give an error now
   (setq flycheck-emacs-lisp-load-path 'inherit))
+
+(defun emacs-lisp/init-flycheck-package ()
+  (use-package flycheck-package
+    :defer t
+    :init (with-eval-after-load 'flycheck
+            (flycheck-pos-tip-mode))))
 
 (defun emacs-lisp/post-init-counsel-gtags ()
   (spacemacs/counsel-gtags-define-keys-for-mode 'emacs-lisp-mode))


Compro-Prasad:
Don't you think we should also add some documentation to README?

jjzmajic:
@Compro-Prasad fair point mate. Tonight isn't that great for me, but i'll add a commit tomorrow. @duianto  Should I make it a `fixup` or leave it as 2 commits?

duianto:
The convention is to squash commits.

jjzmajic:
My mistake, still switching from terminal to magit. Should be good now!

smile13241324:
Thank you for contributing to Spacemacs!:+1:

khjcph:
This seems to interfere with the 'syntax-checking' layer. Putting
```
```
in .spacemacs now has no effect.

jjzmajic:
I can't reproduce this, so I can't be sure, but this doesn't seem to have anything to do with the changes I introduced. I suspect the following is how it's supposed to work in `syntax-checking`'s `packages.el`:
```
(defun emacs-lisp/init-flycheck-package ()
  (use-package flycheck-package
    :defer t
    :init (with-eval-after-load 'flycheck
            (flycheck-pos-tip-mode))))
```
So it only loads the package if the variable is set. But for some reason, the `emacs-lisp` layer calls the function directly like so:
```
(defun emacs-lisp/init-flycheck-package ()
  (use-package flycheck-package
    :defer t
    :init (with-eval-after-load 'flycheck
            (flycheck-pos-tip-mode))))
```

Not checking the variable or loading the package. So if it happens to be loaded (which it likely is) you're screwed.

Thoughts/comments on anything I'm missing before I go in trying to fix this?

jjzmajic:
@khjcph any reason to believe it was these changes that broke it? if so i could really use a bit more info

khjcph:
Yeah, the `syntax-checking` layer defines:
```
(defun syntax-checking/init-flycheck-pos-tip ()
  (use-package flycheck-pos-tip
    :if syntax-checking-enable-tooltips
    :defer t
    :init
      (flycheck-pos-tip-mode))))
```
However, it seems that the `emacs-lisp` layer enables `flycheck-pos-tip-mode` nomatter whether the `syntax-checking-enable-tooltips` variable is set to `nil` or not. If I comment out,
```
(defun emacs-lisp/init-flycheck-package ()
  (use-package flycheck-package
    :defer t
    :init (with-eval-after-load 'flycheck
            (flycheck-pos-tip-mode))))
```
in the `emacs-lisp` layers `packages.el`, everything seems to work as expected(at least for the `syntax-checking` layer)

khjcph:
I'd be happy to help out if there is anything else I can do?

khjcph:
Is there a reason that the `emacs-lisp` layer uses `flycheck` explicitly, instead of relying on the `syntax-checking` layer?

smile13241324:
Yes, I agree `flycheck-pos-tip-mode` is clearly owned by the syntax checking layer, therefore `emacs-lisp layer` should not try to activate it. When `syntax-checking` is loaded it takes care that `flycheck-pos-tip` will be activated after flycheck is loaded if not deactivated via layer variables.

So I would assume the line
```
 :init (with-eval-after-load 'flycheck
            (flycheck-pos-tip-mode))
```
is not necessary as it would already by covered by the syntax checking layer during layer initialisation phase.

@khjcph if you have the time you could try this out and post a PR if it works.

duianto:
I don't know if it's relevant but when cloning a fresh copy of Spacemacs and after installing the packages, then a warnings buffer opens with this message:
>Error (use-package): flycheck-package/:init: Symbol’s function definition is void: flycheck-pos-tip-mode

After restarting the message doesn't reappear.

smile13241324:
@khjcph do you mean this block 
```
(defun emacs-lisp/post-init-flycheck ()
  ;; Don't activate flycheck by default in elisp
  ;; because of too much false warnings
  ;; (spacemacs/enable-flycheck 'emacs-lisp-mode)

  ;; Make flycheck recognize packages in loadpath
  ;; i.e (require 'company) will not give an error now
  (setq flycheck-emacs-lisp-load-path 'inherit))
```
the layer is only defining a function to be loaded after the initialisation of flycheck, however the layer has no init function for flycheck therefore it is not owning it and is not directly called.

Flycheck package is directly loaded in emacs package as it is emacs specific where flycheck is used by many different languages.

jjzmajic:
@duianto should this be a separate issue then? I don't see the connection with this PR

smile13241324:
@khjcph If you want to have a look, you could play a bit around with deactivating the init function, if not I will have a look at this later.

duianto:
@jjzmajic The variable `flycheck-pos-tip-mode` was mentioned here, and in the post-package install error message. So I thought that someone here might know if that also might have been caused by the issue being discussed here.

jjzmajic:
@duianto, don't get me wrong, I don't mind, I'll try to help out tonight as well if I find the time. Just thought it might be useful to separate it out for clarity as it seems like a broader concern others might have insight on. But if we resolve this here I'm a happy camper too

khjcph:
I created a pull request here: #12165.

smile13241324:
Sorry @khjcph just pushed a fix, anyway thanks for the effort.

aaronjensen:
AFAICT, this is never enabled. This should be the use package decl:

```el
  (use-package flycheck-package
    :hook (emacs-lisp-mode . flycheck-package-setup))
```

jjzmajic:
Dammit, I really botched this one. I even remember thinking don't forget to initialize the bloody thing. Given that it has already been merged *and* edited, should I do this on a separate PR to avoid a mess?

aaronjensen:
No worries at all! I’ve got a PR up already https://github.com/syl20bnr/spacemacs/issues/12196

PR #9689, opened 2017-10-02
colors layer: use nyan-mode from melpa

robbyoconnor:
We no longer need to use a local package for [nyan-mode](https://github.com/TeMPOraL/nyan-mode) now that it is [on melpa](https://melpa.org/#/nyan-mode).

PR #12153, opened 2019-04-10
Actually download `nyan-mode' from MELPA.

jjzmajic:
Why was this set to `:location local` in the first place? I know it's a useless
feature, it just bugged me.


sdwolfz:
I think there was an incompatibility that required a fork until the changes were upstreamed, don't know it's status though.

jjzmajic:
@sdwolfz I see. It worked when I tested it just now though

robbyoconnor:
See discussion here: https://github.com/syl20bnr/spacemacs/pull/9689

robbyoconnor:
Your PR fails to remove the local package as well. So this is woefully incomplete.

smile13241324:
I have just compared the version on melpa with the fork from @syl20bnr and noticed that the changes done for the spacemacs specific version are not included in the version on melpa.

I am not sure if providing a PR to upstream would help. I have checked the project pulse of [https://github.com/TeMPOraL/nyan-mode/pulse/monthly](the original repo) and there had been no commit since over a month.

robbyoconnor:
See the discussion on my closed PR.

robbyoconnor:
See this comment from @syl20bnr https://github.com/syl20bnr/spacemacs/pull/9689#issuecomment-336713331: 

> Is this in the MELPA repo ?
> 
> -** Fork modifications
> -- the starting nyan cat image has no rainbow behind it,
> -- the starting rainbow image has zigzags,
> -- the animation loop is not endless and occurs only when nyan cat moves.

jjzmajic:
Should I close this then? Or do you guys think we should should try to solve this upstream again?

robbyoconnor:
@jjzmajic I would close this

Issue #41, opened 2014-10-26
Cannot insert `f` in ex command mode


syl20bnr:
The `fd` quest continues, surprisingly enough I never noticed that `f` does not work while entering an ex command.

PR #12159, opened 2019-04-12
wip: Disable evil-escape-mode in evil-mc-mode

sei40kr:
Disable `evil-escape-mode` in `evil-mc-mode` since it doesn't work.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1627,6 +1627,7 @@ Other:
 **** Multiple Cursors
 - Disabled Spacemacs paste transient state when pasting to avoid pasting issue
   when there are more than one cursor (thanks to braham-snyder)
+- Disabled =evil-escape= minor mode in =evil-mc-mode=.
 **** Neotree
 - Made neotree an optional instead of a default layer
 - Move neotree to its own layer in new +filetree folder (thanks to Sylvain
--- a/layers/+misc/multiple-cursors/packages.el
+++ b/layers/+misc/multiple-cursors/packages.el
@@ -24,6 +24,11 @@
       (when (or (spacemacs/system-is-mac) (spacemacs/system-is-mswindows))
         (setq evil-mc-enable-bar-cursor nil))
 
+      ;; evil-mc is not compatible with evil-escape
+      (add-hook 'evil-mc-mode-hook
+                #'(lambda ()
+                    (add-to-list 'evil-mc-incompatible-minor-modes
+                                 'evil-escape-mode)))
       ;; evil-mc is not compatible with the paste transient state
       (define-key evil-normal-state-map "p" 'spacemacs/evil-mc-paste-after)
       (define-key evil-normal-state-map "P" 'spacemacs/evil-mc-paste-before)


yuhan0:
What about it doesn't work? I use evil-escape and evil-mc together all the time and it works fine for me.

antonalekseev:
It does work for me, but not in a practical way. It actually escapes on key sequence, but the first character of the sequence (`k` in my `kj` case) ends in insert mode. Maybe timeouts are involved somehow.

evil-mc-and-evil-escape-interplay

#### System Info :computer:
- OS: darwin
- Emacs: 26.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 4b195ddfc)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
(aaa-lispyville ansible
                (auto-completion :variables spacemacs-default-company-backends
                                 '(company-yasnippet
                                   (company-gtags company-etags company-keywords)
                                   company-files company-dabbrev)
                                 auto-completion-return-key-behavior 'complete auto-completion-tab-key-behavior 'cycle auto-completion-enable-snippets-in-popup t auto-completion-enable-help-tooltip 'manual auto-completion-enable-sort-by-usage t)
                common-lisp csv
                (dash :variables helm-dash-docset-newpath
                      (concat
                       (getenv "HOME")
                       "/Library/Application Support/Dash/DocSets/"))
                emacs-lisp fzf
                (gtags :variables gtags-enable-by-default nil)
                git github
                (go :variables go-tab-width 4 godoc-at-point-function 'godoc-gogetdoc go-use-gometalinter t)
                html aaa-ipython-notebook ivy
                (json :variables json-fmt-tool 'prettier)
                (lsp :variables lsp-ui-sideline-enable nil lsp-ui-doc-enable nil lsp-ui-flycheck-enable nil lsp-auto-guess-root t)
                (multiple-cursors :variables multiple-cursors-backend 'evil-mc)
                neotree
                (javascript :variables javascript-fmt-tool 'prettier javascript-backend 'lsp js2-basic-offset 2 js-indent-level 2)
                (org :variables org-confirm-babel-evaluate nil org-bullets-bullet-list
                     '("■" "◆" "▲" "▶"))
                (python :variables python-backend 'lsp python-test-runner 'pytest)
                racket restclient
                (shell :variables shell-default-shell 'eshell shell-default-term-shell "zsh")
                theming
                yaml)
```

duianto:
I'm seeing this issue when I try to exit insert state with a quick `fd`.
- Open the scratch buffer
- Type:
```
abc
abc
```
- With the cursor on the second lines `c`, press `C-p` (which calls `(evil-mc-make-and-goto-prev-match)`), it leaves a second cursor at the end of the second line, and the current cursor moves to the end of the first line.
- Enter insert state: `i`
- Press `fd` quickly.

#### Observed

```
abc
abfc
```
The cursor on the first line moves back to the `b` (expected).

An `f` was inserted between `b` and `c` on the second line (unexpected), 
and the second cursor is on the `c` (unexpected).

#### System Info :computer:
- OS: windows-nt
- Emacs: 26.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 12defb470)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(c-c++ elfeed helm emacs-lisp git multiple-cursors treemacs version-control)
```

antonalekseev:
Related issues https://github.com/gabesoft/evil-mc/issues/27, https://github.com/gabesoft/evil-mc/issues/41.

yuhan0:
Ah, I see - I must have quite a non-conventional setup of evil-escape then, because I bind the `evil-escape` command directly to a key, instead of relying on the two-character+auto-delete method which I found to be quite a hack and didn't work in many circumstances. 

In any case it's probably something to be solved upstream instead of in Spacemacs?

smile13241324:
Just read the discussion, can someone tell is this now to be merged or has somebody reported to upstream?

duianto:
There might be a WIP solution in the second related issue that antonalekseev posted:
[Evil-Escape compatibility #41](https://github.com/gabesoft/evil-mc/issues/41)

But the last comment was from 2016 so it seems to have stalled.

A possibly dangerous thing can currently happen when evil-escape is enable.

If someone has two evil-mc cursors and one is off screen, then `fd` (`evil-excape`) seems to have worked as expected at the visible cursor, but it added an `f` at the other cursor and the cursors are now out of sync.

Therefore it's probably best to disable it until evil-mc adds support for it.

duianto:
@sei40kr 
evil-mc-mode is enabled in both prog and text modes:
https://github.com/syl20bnr/spacemacs/blob/2342f76cf15d2845f8b38c56c23ebd7f4c83478b/layers/%2Bmisc/multiple-cursors/packages.el#L35-L36

If this PR disables `evil-escape` when `evil-mc-mode` is active, then it will be disabled in a lot of buffers. It should only be disabled if `evil-mc-mode` is active and there are two or more cursors.

lebensterben:
@duianto 

> @sei40kr
> evil-mc-mode is enabled in both prog and text modes:
> 
> https://github.com/syl20bnr/spacemacs/blob/2342f76cf15d2845f8b38c56c23ebd7f4c83478b/layers/%2Bmisc/multiple-cursors/packages.el#L35-L36
> 
> If this PR disables `evil-escape` when `evil-mc-mode` is active, then it will be disabled in a lot of buffers. It should only be disabled if `evil-mc-mode` is active and there are two or more cursors.

This would disable evil-escape when evil-mc has more than one cursor.

https://github.com/syl20bnr/spacemacs/issues/14765#issuecomment-839166851

What do you think?

duianto:
Yeah, it's probably best to disable `evil-escape` while there are multiple `evil-mc` cursors,
until a better solution is found.

Because there is at least one case where it seems like it worked as expected.

If only one instance of a variable is visible,
and one adds multiple cursors on every instance, with `grm`.

When `fd` is pressed quickly to exit insert state.

Then it did escape to normal state, and the variable looks as expected,
but every other instance has an extra `f` or `fd`.

lebensterben:
> Would it be possible to add a message that `evil-escape` is disabled,
> 
> with a link to the upstream issue, when someone tries to press `fd` quickly?
> 
> 

Sure. The evil-escape-inhibit function can be anything.

If evil-mc-state is non-nil, just call `(message ..)`, which should return non-nil. Or otherwise we can return a `t` after that. The non-nil return value will disable evil-escape as expected.

When evil-mc-state is nil, nothing happens and evil-mc is not disabled.

duianto:
Here's a possible WIP fix, by pausing the cursors before the `f`'s are inserted at the other cursors:
```elisp
(defun custom/evil-mc-evil-escape-fix ()
  (when (evil-mc-has-cursors-p)
    (evil-mc-pause-cursors)
    (run-with-idle-timer
     0 nil '(lambda ()
              (evil-mc-resume-cursors)))))

(advice-add 'evil-escape-func :before 'custom/evil-mc-evil-escape-fix)
;; (advice-remove 'evil-escape-func 'custom/evil-mc-evil-escape-fix)
```

Issue: The cursors aren't aligned vertically.

I thought that I should post the progress so far,
since it seems to almost work as expected.

gdkrmr:
@duianto Thanks! I gave it a quick try and it seems to work. I will test it in my setup!

duianto:
This moves back the fake cursors, to keep the cursors aligned, except for in two cases.
```elisp
  (defun custom/evil-mc-evil-escape-fix ()
    (when (evil-mc-has-cursors-p)
      (evil-mc-pause-cursors)
      (run-with-idle-timer
       0 nil '(lambda ()
                (evil-mc-resume-cursors)
                (let ((evil-mc-command '((:name . evil-backward-char))))
                  (evil-mc-execute-for-all))))))
```
With evil-mc cursors on the three instances of `abc`:
```
ab[c]
def ab[c]
ab[c]
```
`i` enters insert state (only the real cursor changes to a vertical bar)
```
ab|c
def ab[c]
ab[c]
```
`fd` quickly, moves all cursors back one character.
```
a[b]c
def a[b]c
a[b]c
```

### Issue 1 of 2
If insert state is entered on the lines last character, by pressing `a` (`evil-append`).
```
abc|
def abc[]
abc[]
```
`fd` results in:
```
ab[c]
def a[b]c
a[b]c
```
It seems like the fake cursors only looked like they were after the last character.

### Issue 2 of 2
When the real cursor is on the first column.
```
|abc
def [a]bc
[a]bc
```
`fd` moves back the fake cursors that are not on the first column.
```
[a]bc
def[ ]abc
[a]bc
```

### Notes
The docstring for: `evil-mc-execute-for-all` says:
>Execute the current command, stored at ‘evil-mc-command’, for all fake cursors.

When there are multiple evil-mc cursors.
and `h` (`evil-backward-char`) is pressed.
Then `evil-mc-execute-for-all` is called in the `post-command-hook`.

And `evil-mc-command` contains:
```elisp
((nil) (:keys . [104]) (:keys-count) (:undo-list-pointer-post nil (nil rear-nonsticky nil 8 . 9) (
 . -12) (8 . 13) 5 nil (nil rear-nonsticky nil 4 . 5) (
 . -8) (4 . 9) 3 nil (1 . 4) (t . 0)) (:keys-post-raw . [104]) (:keys-post . [104]) (:last-input . 104) (:evil-state-end . normal) (:undo-list-pointer-pre nil (nil rear-nonsticky nil 8 . 9) (
 . -12) (8 . 13) 5 nil (nil rear-nonsticky nil 4 . 5) (
 . -8) (4 . 9) 3 nil (1 . 4) (t . 0)) (:evil-state-begin . normal) (:keys-pre-with-count nil evil-backward-char [104] nil) (:keys-pre . [104]) (:name . evil-backward-char))
```

And it works to move the back the fake cursors when evaluating:
(newlines have been added to show each list on it's own line, for readability)
```elisp
(let ((evil-mc-command '((nil)
                         (:keys . [104])
                         (:keys-count)
                         (:undo-list-pointer-post nil (nil rear-nonsticky nil 8 . 9) ( . -12) (8 . 13) 5 nil (nil rear-nonsticky nil 4 . 5) ( . -8) (4 . 9) 3 nil (1 . 4) (t . 0))
                         (:keys-post-raw . [104])
                         (:keys-post . [104])
                         (:last-input . 104)
                         (:evil-state-end . normal)
                         (:undo-list-pointer-pre nil (nil rear-nonsticky nil 8 . 9) ( . -12) (8 . 13) 5 nil (nil rear-nonsticky nil 4 . 5) ( . -8) (4 . 9) 3 nil (1 . 4) (t . 0))
                         (:evil-state-begin . normal)
                         (:keys-pre-with-count nil evil-backward-char [104] nil)
                         (:keys-pre . [104])
                         (:name . evil-backward-char))))
  (evil-mc-execute-for-all))
```

But it seems to work to reduce `evil-mc-command` to:
```elisp
(let ((evil-mc-command '((:name . evil-backward-char))))
  (evil-mc-execute-for-all))
```

I'll ask upstream if the other lines are important.

duianto:
The second issue isn't limited to this "solution".
It also happens without this advice and without Spacemacs.
I'll report it upstream.

duianto:
This seems to work:
```elisp
  (setq evil-mc-custom-known-commands
        '((custom/evil-mc-evil-escape-move-back-fake-cursors
           (:default . evil-mc-execute-default-call))))

  (defun custom/evil-mc-evil-escape-move-back-fake-cursors ()
    "Move the fake cursors to the left once,
unless they already are at the beginning of the line."
    (unless (bolp) (backward-char)))

  (defun custom/evil-mc-evil-escape-fix ()
    "Prevent the first evil-escape-key-sequence key (default: f),
from being typed at all of the fake cursors.
And move back the fake cursors when the real insert state cursor is at the end
of a line."
    (when (evil-mc-has-cursors-p)
      (evil-mc-pause-cursors)
      (run-with-idle-timer
       0 nil '(lambda ()
                (evil-mc-resume-cursors)
                (let ((evil-mc-command '((:name . custom/evil-mc-evil-escape-move-back-fake-cursors))))
                  (evil-mc-execute-for-all))))))

  (advice-add 'evil-escape-func :before 'custom/evil-mc-evil-escape-fix)
  ;; (advice-remove 'evil-escape-func 'custom/evil-mc-evil-escape-fix)
```

It handles issue 1 of 2 above: https://github.com/syl20bnr/spacemacs/pull/12159#issuecomment-859995710
Pressing `fd` (`evil-escape`) with the real insert state cursor at the end of a line.
Moves back the fake cursors so that all cursors are aligned.

Issue 2 of 2 might not have an easy/desired fix.
The expected vim behavior when exiting insert state to normal state is that the cursor moves back.
This causes the evil-mc cursors to become unaligned, when the real or a fake cursor is at the beginning of a line.

Issue 2 of 2 been reported upstream.
`esc` unaligns the cursors, when real cursor on first column https://github.com/gabesoft/evil-mc/issues/124

PR #12176, opened 2019-04-14
[lsp] Sync lsp layer with latest lsp mode functionality

yyoncho:
- provide keybindings for the new lsp-mode functionality
 - use lsp-treemacs and helm-lsp for errors and workspace symbols

### DIFF
--- a/layers/+tools/lsp/funcs.el
+++ b/layers/+tools/lsp/funcs.el
@@ -38,36 +38,44 @@ https://github.com/emacs-lsp/lsp-javascript/issues/9#issuecomment-379515379"
       (spacemacs//lsp-bind-peek-navigation-functions "G")))
 
   (spacemacs/set-leader-keys-for-minor-mode 'lsp-mode
-    ;;format
+    ;; format
     "=b" #'lsp-format-buffer
-    ;;goto
+    "=r" #'lsp-format-region
+    ;; goto
     "gt" #'lsp-find-type-definition
     "gk" #'spacemacs/lsp-avy-goto-word
     "gK" #'spacemacs/lsp-avy-goto-symbol
-    "ge" #'lsp-ui-flycheck-list
     "gM" #'lsp-ui-imenu
-    ;;help
+    ;; help
     "hh" #'lsp-describe-thing-at-point
-    ;;jump
-    ;;backend
+    ;; jump
+    ;; backend
     "ba" #'lsp-execute-code-action
     "bd" #'lsp-describe-session
     "br" #'lsp-restart-workspace
-    ;;refactor
+    "bs" #'lsp-shutdown-workspace
+    ;; refactor
     "rr" #'lsp-rename
-    ;;toggles
+    ;; toggles
     "Td" #'lsp-ui-doc-mode
     "Ts" #'lsp-ui-sideline-mode
     "TF" #'spacemacs/lsp-ui-doc-func
     "TS" #'spacemacs/lsp-ui-sideline-symb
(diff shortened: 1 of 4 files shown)


layers/+tools/lsp/packages.el
```diff
@@ -11,10 +11,12 @@
...
   '(
-     lsp-mode
-     lsp-ui
-     (company-lsp :requires company)
-     ))
+    lsp-mode
+    lsp-ui
+    (company-lsp :requires company)
+    helm-lsp
+    lsp-treemacs
```
sdwolfz: I think this one should be `(lsp-treemacs :requires treemacs)`.

layers/+tools/lsp/funcs.el
```diff
@@ -38,36 +38,44 @@ https://github.com/emacs-lsp/lsp-javascript/issues/9#issuecomment-379515379"
...
+    ;; toggles
     "Td" #'lsp-ui-doc-mode
     "Ts" #'lsp-ui-sideline-mode
     "TF" #'spacemacs/lsp-ui-doc-func
     "TS" #'spacemacs/lsp-ui-sideline-symb
-    "TI" #'spacemacs/lsp-ui-sideline-ignore-duplicate))
+    "TI" #'spacemacs/lsp-ui-sideline-ignore-duplicate
+    "Tl" #'lsp-lens-mode
+    ;; folders
+    "fs" #'lsp-workspace-folders-switch
```
sdwolfz: These keybindings conflict with `Web frameworks` conventions (`SPC m f`). Can you think of any other prefix that works?

yyoncho: `SPC m b f a` or `SPC m w f a` or `SPC m F a`?

layers/+tools/lsp/packages.el
yyoncho: Is this really needed? I tested with setting `neotree` as a file explorer and it seems to be working fine. lsp-treemacs uses treemacs only as a tree renderer which doesn't seem to interfere with file browser functionality. I think that is different from the company case since company-lsp is directly related to completion framework. What do you think? It will be a pity if we have this functionality only if the user has selected treemacs as a file browser.

sdwolfz: OK, if it's not conflicting with other file explorers then all is good! :+1:

layers/+tools/lsp/funcs.el
sdwolfz: `SPC m F` should work as there is only on [wrongly] defined keybinding for Haskell that conflicts with it (which should be under `SPC m =` as it's a formatting one).

yyoncho: Force pushed a version with the new keybings.

smile13241324:
Thanks for this awesome update :)

I have added a prefix description for the new "F" bindings and fixed a small issue in the documentation before merging.

Btw: is it normal that the bindings for `SPC m g h` and `SPC m g m` are not bound?
Just noticed that I cannot find them in the python layer.

yyoncho:
> Btw: is it normal that the bindings for `SPC m g h` and `SPC m g m` are not bound?
> Just noticed that I cannot find them in the python layer.

Python does not provide that functionality (e. g. call hierarchies, etc). If I recall correctly, back then when lsp layer was introduced it was decided to keep this prefixes in lsp layer (not in cpp one)  although not all of the languages will use this prefixes. At this point, the methods are not part of the spec and only a few language servers have them(I think ccls/cquery) and the methods are going to be added to the spec soon so more and more servers will provide this functionality.

smile13241324:
Thanks for the info

PR #12177, opened 2019-04-14
Use dap mode layer for python debugging

yyoncho:
- I had to remove the "db" keybinding manually before setting the dap bindings.
  An alternative approach is to use a different binding for it.


smile13241324:

PR #12184, opened 2019-04-16
Bind ~SPC f F~ to =helm-find= for recursive directory search.

jjzmajic:
I realize this may be controversial, but as I understand it (correct me if I'm
wrong), all `SPC f F` does differently compared to `SPC f f` is call the vanilla
version of `helm-find-files` that does take `thing-at-point` into account. But I
presume we removed this feature in `spacemacs/helm-find-files` for a reason.

Also, it makes sense as a *forte* version of `helm-find-files` when you can't
find something in your current directory and decide to drill deeper. This
complies with the general tone of uppercase/lowercase bindings in Spacemacs, and
provides more utility than te subtle distinction between different
implementations of `helm-find-files`.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1411,6 +1411,7 @@ Other:
   removed upstream.
 - Load helm before read-file-name and completing-read (thanks duianto and thanks
   Miciah for a more elegant solution)
+- Bind ~SPC f F~ to =helm-find= for recursive directory search.
 **** HTML
 - Added impatient-mode under ~SPC m i~ (thanks to geo7)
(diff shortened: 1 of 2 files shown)


jjzmajic:
The build seems to be failing due to an I/O error:
```
(Spacemacs) Error: 
An error occurred while installing evil (error: (file-error https://melpa.org/packages/goto-chg-20190110.2114.el Bad Request))
```
Or could this have somehow been caused by my edits? Would it be possible to rerun it?

jjzmajic:
Oh, just noticed the `lazy-helm` wrappers. Should wrap `helm-find` in it too?

smile13241324:

PR #12185, opened 2019-04-16
Fixes #10762 Added support for user labels in LaTeX exports from org

### LINKED ISSUE #10762: Feature request: Support for CUSTOM_ID in org-mode
See #10731 for further reference on why this feature should be included.

Compro-Prasad:


Compro-Prasad:
@smile13241324 Simple fix

smile13241324:

I have read a bit in the org documentation and this change could have some side effects. However I agree with you in that without this change many of the more interesting org features would not work so lets merge it.

PR #12190, opened 2019-04-16
Added instruction on installing importmagic and epc.

Trapez:
Importmagic and epc isn't really optional if error messages are to be avoided.


duianto:

PR #12175, opened 2019-04-14
[java] init backends conditionally

yyoncho:
- otherwise, there are shortcut collisions

PR #12195, opened 2019-04-18
Moved Ensime support to the Scala layer, removed Java support

teozkr:
03e6bc0691456a9e39a1944bc04e65a6a404078c (#12175) broke ENSIME for Scala users who don't use it for Java. That's a problem since "ENJIME" (the Java support) was removed from the upstream about a year ago (ensime/ensime-server#1943). This PR removes all `java-mode` integration but enables it unconditionally for `scala-mode`. It also removes the `java--ensime-modes` machinery (since ENSIME is now only useful for `scala-mode` and its derivatives anyway).


sdwolfz:
@syl20bnr ☝️ should we discuss about this?

Maybe leave java with only the `lsp` and `meghanada` backends and remove the hacky ones?

smile13241324:
I would vote for moving ENSIME to scala layer especially as the ENSIME project dropped the java integration in favor of LSP. I think that everybody doing java development with spacemacs will use LSP anyway as this growing faster than everything else.

robbyoconnor:
I'd use LSP over any other backends for anything really. It's just better. It's one of the great contributions Microsoft made to the development communities.

Ziemin:
I would also move towards incorporating lsp wherever possible, especially that now we have very strong alternative to ensime- Metals, which is a provided through lsp-scala package. Moreover, as opposed to ensime it's actively developed.

- [Metals configuration for Emacs](https://scalameta.org/metals/docs/editors/emacs.html)

teozkr:
Metals has improved a lot lately, but it's still not on par with Ensime. For example, it's still missing implicit expansion and go-to-Scaladoc integration.

smile13241324:
Just noticed that there is another PR working on integrating `metals`, I have directed them to this discussion and asked them to adapt their PR to the decision taken here.

kryptt:
I'm in favor of reducing the surface area for ensime. It has not been updated since mid 2018, what I can do is remove the leave ensime while removing java layer dependency in my pull request.

syl20bnr:
> Maybe leave java with only the lsp and meghanada backends and remove the hacky ones?

:+1:

smile13241324:

I have also made some separate commits to drop eclim support and make LSP the default for the java layer. As far as I see everything should be working again.

Independent of this change I have noticed that the flycheck integration seems not to be valid anymore. When I check with meghanada the checker is disabled due to a missing config file though no config file is requested in the readme. For lsp the task for flycheck is delegated to `lsp-ui` which reports some strange `next-checkers` especially some javascript ones.

Both do not report basic syntax errors like `Sysem.out.println()` but do report stuff like `Sy sem.out.println()`. Is this normal with the java layer?

PR #12205, opened 2019-04-20
[org] Add agenda transient keybinding for org-agenda-show-clocking-issues

bhipple:
As documented upstream [1], the `org-agenda-show-clocking-issues` view can be
used to show overlapping clocks, gaps, etc. on the agenda, and is bound by
default to `v c`. This commit adds a keybinding it in the org-agenda transient mode.

This may not be the best place for it, because it isn't a proper toggle;
upstream says to clear the issues flags by toggling logbook mode itself and does
not provide a clocking issues toggle directly.

[1] https://orgmode.org/manual/Agenda-commands.html#Agenda-commands

### DIFF
--- a/layers/+emacs/org/README.org
+++ b/layers/+emacs/org/README.org
@@ -659,6 +659,7 @@ are listed bellow.
 | ~tl~        | log                 | org-agenda-log-mode               |
 | ~ta~        | archive             | org-agenda-archives-mode          |
 | ~tr~        | clock report        | org-agenda-clockreport-mode       |
+| ~tc~        | clocking issues     | org-agenda-show-clocking-issues   |
 | ~td~        | diaries             | org-agenda-toggle-diary           |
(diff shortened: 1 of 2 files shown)


smile13241324:
Thank you for contributing to spacemacs, I have cherry-picked your commit into develop, you can safely remove your branch.

Just for the records, I agree with you that putting this under `t` for toggle is not too nice, however I also see no other place for this. In a perfect world upstream would provide a toggle for this. But as it is right now the feature is useful so its nice to have it there though it is definitely not polished i.e. you can duplicate the error indicators by executing `org-agenda-show-clocking-issues` twice.

All in all I think its better to have it bound with a spacemacs binding instead of forcing users to use the original org bindings for this feature. Lets see if someone will complain.

bhipple:
Thanks for the reasoned and carefully considered response, @smile13241324!

Issue #9255, opened 2017-07-22
`spacemacs/error-delegate` does not detect flycheck when `org-mode` is open in background


JJJollyjim:
#### Description :octocat:
`spacemacs/next-error` does not detect flycheck when org-mode is open

#### Reproduction guide :beetle:
- Start Emacs
- Open an org file (including a spacemacs help file)
- Open a code file with functional flycheck checking
- Attempt to navigate around errors in the code buffer with `SPC e n` / `SPC e p`

*Observed behaviour:* :eyes: :broken_heart:
`spacemacs/error-delegate` detects that `'emacs` checking is in use rather than `'flycheck`, as `(next-error-find-buffer)` returns the org buffer

*Expected behaviour:* :heart: :smile:
Should be able to navigate flycheck errors in the code buffer, as is possible with no org buffer open in the background

#### System Info :computer:
- OS: gnu/linux
- Emacs: 25.2.1
- Spacemacs: 0.200.9
- Spacemacs branch: master (rev. 8e1af145)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(php yaml ruby sql typescript auto-completion c-c   emacs-lisp git markdown
     (latex :variables latex-build-command "LaTeX" latex-enable-auto-fill t)
     bibtex org
     (shell :variables shell-default-height 30 shell-default-position 'bottom)
     (spell-checking :variables ispell-dictionary "en_GB")
     syntax-checking colors emoji games javascript rust html elixir python java nixos elm haskell extra-langs systemd)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GCONF GSETTINGS NOTIFY ACL GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 MODULES

#### Backtrace :paw_prints:
```

```

Issue #6639, opened 2016-07-20
`spacemacs/next-error` cannot find flycheck error


bixuanzju:
#### Description :octocat:
`spacemacs/next-error` cannot find flycheck error

#### Reproduction guide :beetle:
- Start Emacs
- Create a file within which flycheck is enabled and an error is indicated
- `SPC e n` and see!

The root cause is that in `spacemacs//error-delegate` we use `next-error-find-buffer` to find a proper error buffer, but that function would signal an error if no such buffer is found, and immediately return to the Emacs editor command loop. 

Instead, I think we would like to return `nil` in such case. Simply wrap `next-error-find-buffer` with `ignore-errors` would solve the problem. I can make PR if you guys think that's correct.

*Observed behaviour:* :eyes: :broken_heart:
`SPC e n` gives "No buffers contain error message locations"

*Expected behaviour:* :heart: :smile:
Go to the next error found by `flycheck`

#### System Info :computer:
- OS: darwin
- Emacs: 25.0.95.1
- Spacemacs: 0.105.21
- Spacemacs branch: develop (rev. 9773ad6)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(helm
 (auto-completion :variables auto-completion-enable-sort-by-usage t)
 emacs-lisp
 (git :variables git-magit-status-fullscreen t)
 markdown org
 (shell :variables shell-default-height 30 shell-default-position 'bottom)
 spell-checking
 (syntax-checking :variables syntax-checking-enable-tooltips nil)
 (version-control :variables version-control-diff-tool 'diff-hl)
 osx idris
 (haskell :variables haskell-completion-backend 'intero haskell-enable-hindent-style "johan-tibell")
 (latex :variables latex-build-command "LatexMk")
 (evil-snipe :variables evil-snipe-enable-alternate-f-and-t-behaviors t)
 ocaml agda lua sml yaml
 (c-c   :variables c-c  -enable-clang-support t)
 html proof-general)
```

#### Backtrace :paw_prints:
```
Debugger entered--Lisp error: (error "No buffers contain error message locations")
  signal(error ("No buffers contain error message locations"))
  error("No buffers contain error message locations")
  next-error-find-buffer()
  (let ((buf (next-error-find-buffer))) (not (and buf (get-buffer-window buf))))
[...]

Issue #8078, opened 2016-12-30
Javascript layer: js2 and flycheck conflict on `next-error-function`


lbonn:
#### Description :octocat:

I'd like to delegate syntax checking to flycheck in javascript mode.

When the layer is enabled, js2 conflicts with flycheck and makes it hard to cycle through errors.
This happens even if js2 is configured to silence its warnings (`js2-mode-show-parse-errors` and `js2-mode-show-strict-warnings` set to `nil`).

Apparently, in my setup, js2 sets `next-error-function` after flycheck. So, this works for me (in `dotspacemacs/user-config`):

```
(setq js2-mode-show-parse-errors nil)
(setq js2-mode-show-strict-warnings nil)
(add-hook 'js2-init-hook
          '(lambda ()
              (setq next-error-function 'flycheck-next-error)
              )
          )
```

I've been trying spacemacs (and emacs) for about a week, so I don't know if this should go into the javascript layer like this...

#### Reproduction guide :beetle:
- Start Emacs
- Disable js2 errors
- Open a javascript file with errors
- Look at `spacemacs/goto-flycheck-error-list`
- Try to cycle with `]q`

*Observed behaviour:* :eyes: :broken_heart:

The error jumps do not match with the flycheck list.

*Expected behaviour:* :heart: :smile:

`spacemacs/next-error` should use `flycheck-next-error`.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 25.1.1
- Spacemacs: 0.200.5
- Spacemacs branch: nil (rev. 664ba6ab)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: helm
- Layers:
```elisp
(helm auto-completion better-defaults git
      (shell :variables shell-default-height 30 shell-default-position 'bottom)
      (spell-checking :variables spell-checking-enable-by-default nil)
      syntax-checking version-control org evil-commentary evil-snipe vinegar c-c++ emacs-lisp haskell html javascript markdown python yaml)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GCONF GSETTINGS NOTIFY ACL GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11

duianto:
The following PR is now on the `develop` branch:
javascript, typescript and react layers: linter option, next-error-function fix #12207

PR #12208, opened 2019-04-21
[core] fix spacemacs/error-delegate

thanhvg:
Determine flycheck or emacs error backend using next-error-function instead of
next-error-find-buffer

Currently `spacemacs/error-delegate` uses `next-error-find-buffer` to determine error backends

``` elisp
(defun spacemacs/error-delegate ()
  "Decide which error API to delegate to.
Delegates to flycheck if it is enabled and the next-error buffer
is not visible. Otherwise delegates to regular Emacs next-error."
  (if (and (bound-and-true-p flycheck-mode)
           (let ((buf (ignore-errors (next-error-find-buffer))))
             (not (and buf (get-buffer-window buf)))))
      'flycheck
    'emacs))
```
The problem is that `next-error-find-buffer` seems to look for all opened buffers not the current buffer of interest. This leads to false error call when you also have a org-mode buffer opened in another window. Because of org-mode's design `next-error-find-buffer always returns positive on org-mode buffer like this one https://github.com/syl20bnr/spacemacs/issues/9255

I propose to use the local var next-error-function instead, for a buffer if this var is set then the buffer 
errors are handled by emacs, otherwise try flycheck. The change would be:
```elisp
(defun spacemacs/error-delegate ()
  "Decide which error API to delegate to.
Delegates to flycheck if it is enabled and the next-error buffer
is not visible. Otherwise delegates to regular Emacs next-error."
  (if (and (bound-and-true-p flycheck-mode)
            (not next-error-function))
      'flycheck
    'emacs))

```

Previous discussion: https://github.com/syl20bnr/spacemacs/issues/6639

`next-error-function` doc string
```
  Automatically becomes buffer-local when set.
  This variable may be risky if used as a file-local variable.

Documentation:
Function to use to find the next error in the current buffer.
The function is called with 2 parameters:
ARG is an integer specifying by how many errors to move.
RESET is a boolean which, if non-nil, says to go back to the beginning
of the errors before moving.
Major modes providing compile-like functionality should set this variable
to indicate to ‘next-error’ that this is a candidate buffer and how
to navigate in it.

```

### DIFF
--- a/core/core-funcs.el
+++ b/core/core-funcs.el
@@ -340,8 +340,7 @@ current frame."
 Delegates to flycheck if it is enabled and the next-error buffer
 is not visible. Otherwise delegates to regular Emacs next-error."
   (if (and (bound-and-true-p flycheck-mode)
-           (let ((buf (ignore-errors (next-error-find-buffer))))
-             (not (and buf (get-buffer-window buf)))))
+           (not next-error-function))
       'flycheck
     'emacs))
 


syl20bnr:
Does this fix introduces a different behavior than the current one ? If `org-mode` always returns `t` maybe we can ignore them ?

thanhvg:
> Does this fix introduces a different behavior than the current one ? If `org-mode` always returns `t` maybe we can ignore them ?

It will do the job because `next-error-function` checks only current buffer of interest (say a javasctipt file user is working on) while `next-error-find-buffer` scans all opened buffers which is not right.

A quick test is having 2 windows opened side by side (`spc w 2), one window with a js file and other with an org file.

syl20bnr:
Do you think it can have a different behavior if you have errors in multiple buffers after a compilation and you want to cycle between all the compilation errors ?

thanhvg:
> Do you think it can have a different behavior if you have errors in multiple buffers after a compilation and you want to cycle between all the compilation errors ?

I never make a compilation call inside spacemacs so I cannot be 100% sure of this case. If someone can test this case that will be the ultimate answer. 

Here is my speculation on this fix when an error compilation buffer is involved: If `flycheck` is not enabled for the coding mode, then the behavior is the same. If `flycheck` enabled, then compilation error buffer will not be considered unless you are viewing it and invoke `spacemacs/next-error, spacemacs/previous-errror` on it, it this case just do a `` c-x  ` `` which is bound to `next-error`

I find that 'next-error` sometimes is used not for error report. `org-mode` and `occur`
buffer  are heavy users of this function but not for error report. I don't think using `next-error-find-buffer` is the right way.

smile13241324:
@syl20bnr thanks for your answer, I will test and merge this.

@thanhvg can you provide a test case for this feature please? I have just opened two buffers in two different frames, one is org and the other is elisp with flycheck and used `SPC e n` to call `spacemacs/next-error`. The system worked correctly, this is it moved my cursor from the first to the last error in my elisp buffer. Doing the same in org mode has no effect.

To be honest right now I do not see anything go wrong here. With a concrete example of where the function misbehave I might be able to understand and test this properly.

tko:
I may have just witnessed something similar to this, though possibly the exact opposite. 

flycheck-eslint was disabled for jsx buffer. I enabled it, got errors in flycheck buffer, `spacemacs/next-error` falsely claims "No errors" whereas `flycheck-next-error` jumps to the next error appropriately.

I haven't tried this PR to see if it fixes my problem, sorry.

Possibly related https://github.com/syl20bnr/spacemacs/issues/8078

thanhvg:
Yeah, this is hard to explain without an example. I will produce one this weekend when I have time

thanhvg:
@smile13241324 I created an example repo at https://github.com/thanhvg/example for your consideration. It's a python one. 

There are screencasts also. The setup and explain is in TODOs.org 

Basically org buffer will make the call next error to 'next-error instead of  'flycheck-next-error

smile13241324:
@thanhvg thanks for your detailed explanation :+1: 

Thank you for contributing :1st_place_medal:

Miciah:
@smile13241324, can https://github.com/syl20bnr/spacemacs/issues/9255 now be labeled as "Fixed in develop"?

duianto:
@Miciah @smile13241324 I added the `Fixed in develop` label.

PR #12226, opened 2019-04-24
Fix typo in diagnostic message.

bschommer:
The warning that opam config did output an empty string/did not work missed an e in share.


duianto:

PR #12233, opened 2019-04-26
Typescript: make linter configurable between tslint/eslint

TheAncientGoat:
As tslint is being deprecated (see palantir/tslint#4534), it would be a good idea to support linting with eslint in typescript mode, this PR enables that.

### DIFF
--- a/layers/+lang/typescript/README.org
+++ b/layers/+lang/typescript/README.org
@@ -26,7 +26,7 @@ This layer adds support for TypeScript and TSX editing.
 - Eldoc-mode
 - Documentation at point
 - Auto complete
-- Flycheck with linter
+- Flycheck with either tslint or eslint
 - Jump to definition, Jump to type definition
 - Find occurrences (Imenu-mode)
 - Rename symbol
@@ -61,14 +61,36 @@ You can choose formatting tool:
 
 Default is ='tide=.
 
+You can choose either tslint (default) or eslint for linting:
+
+#+BEGIN_SRC emacs-lisp
+  (setq-default dotspacemacs-configuration-layers '(
+    (typescript :variables
+                typescript-linter 'tslint)))
+#+END_SRC
+
+
 ** Pre-requisites
 You will need =node.js v0.12.0= or greater.
 
-If you want linting run:
+If you want linting with tslint run:
 
 #+BEGIN_SRC shell
   npm install -g typescript tslint
 #+END_SRC
+
+If you want linting with eslint run:
+
+#+BEGIN_SRC shell
+  npm install eslint
+#+END_SRC
+
+We need to use the project-local eslint installation in order to pick up plugins
+and presets installed locally.
+
+Ensure that the project directory =node_modules/.bin= is added to the buffer 
+local =exec_path= (see the Javascript layer [[file:~/.emacs.d/layers/+lang/javascript/README.org][README]] documentation for more 
+details).
 
 If you want to use typescript-formatter for formatting run:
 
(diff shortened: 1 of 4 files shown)


layers/+lang/typescript/config.el
```diff
@@ -22,5 +22,8 @@ Currently avaliable 'tide (default), 'typescript-formatter and 'prettier.")
   "The backend to use for IDE features. Possible values are `tide'
 +and `lsp'.")
 
+(defvar typescript-linter 'tslint
+  "Height in percents for the shell window.")
```
smile13241324: You doc string is wrong, please fix it.

layers/+lang/typescript/README.org
```diff
@@ -1,4 +1,5 @@
 #+TITLE: TypeScript layer
+#+TAGS: layer
```
smile13241324: Should be `#+TAGS: layer|uncategorized`

TheAncientGoat:
Thanks for the review @smile13241324  - update got lost in my notifications, will fix / update over the weekend

layers/+lang/typescript/packages.el
```diff
@@ -42,8 +42,19 @@
   (spacemacs/enable-flycheck 'typescript-tsx-mode)
   (with-eval-after-load 'tide
     (with-eval-after-load 'flycheck
-      (flycheck-add-mode 'typescript-tide 'typescript-tsx-mode)
-      (flycheck-add-mode 'typescript-tslint 'typescript-tsx-mode))))
+      (if (eq typescript-linter `tslint)
```
tko: Would probably be better to use `cond` and check for explicit `'eslint` and `'tslint` values instead (and complain if it's anything else)

layers/+lang/typescript/config.el
TheAncientGoat: Fixed

layers/+lang/typescript/README.org
TheAncientGoat: Fixed

layers/+lang/typescript/packages.el
TheAncientGoat: Added the cond check - just writes to the message buffer, not sure if there is a better practice way to do it.

TheAncientGoat:
@smile13241324 Fixed the requested changes, please take a look

layers/+lang/typescript/packages.el
smile13241324: I think this is ok so far, of what I have seen in other layers i.e. ruby if something non matching is given an message is written to the log and not an error is raised.

TheAncientGoat:
@smile13241324  - Hmm, yes, seems like it has an issue when used with lsp-mode, which overrides the flycheck checker. Will see if I can fix

TheAncientGoat:
@smile13241324 Added the hook when using the lsp backend; it works but I think there may be a better way using tsserver plugins as mentioned here: https://github.com/theia-ide/typescript-language-server/issues/58; however, it seems like there isn't one for eslint at this point in time (https://github.com/typescript-eslint/typescript-eslint/issues/254 / )

TheAncientGoat:
Does eslint work for a plain JavaScript project in your setup? You need a
.eslintrc file for flycheck to enable it.

smile13241324:
Hi @TheAncientGoat,

I have given it another try and this time I have added the respective .eslint config file. This fixed my issue. I have fixed the conflicts and merged your PR.

You can safely remove your branch. Thank you for contributing to spacemacs :1st_place_medal:

PR #12234, opened 2019-04-26
Add scala-lsp and metals support to scala

kryptt:
It appears that the ensime project is loosing a lot of steam recently and the scalameta organization is working hard on advancing the metals lsp server. 

This pull request discards ensime in favor of metals...


Compro-Prasad:
IMO it would be better to keep an option for both of these. People can switch to either one by just changing the value of a variable.

kryptt:
I still would like to polish this a bit, but wanted first to see if generally people would agree with the direction of this PR

beyondpie:
@kryptt 
do hope that keep an option for ensime.
I use it for a quite long time, and I love ensime for Scala though ensime is old now. I find lsp mode is not so stable in Emacs currently.

kryptt:
One of the things I noticed while attempting to have them both simultaneously, is that ensime eagerly hooks itself up in scala mode. Let me look into figuring out how to remove those hooks...

smile13241324:
I think this discussion is related to #12195, and should be continued there.

Basically there is a java and a scala layer, the idea would be to move ensime to the scala layer rather than to the java layer. We would like to keep the old systems but would nevertheless encourage to use LSP server solutions were possible.

So I think that integrating the metal LSP server and if possible a DAP compatible extension for scala would be great however I am not sure whether this should be in the scala or in the java layer.

I think this discussion should be continued in #12195 and depending on the outcome be adapted.

kryptt:
@smile13241324 What do you mean by a DAP compatible extension?

Compro-Prasad:
Debugger Adapter Protocol

yyoncho:
@smile13241324 

> We would like to keep the old systems but would nevertheless encourage to use LSP server solutions were possible.

IMO we should encourage the better backend no matter whether it is lsp or anything else. I would propose the following change: When the user hasn't specified explicitly backend spacemacs will pop-up a dialog to which will ask the user to select a backend. Each of the backends might have a number of stars from 0 to 5(optional) + a note which backend is recommended(optional). 

5* - language X supported better than the other editors (e. g. Cider)
4* - very good support (e. g. tide)
3* - good support
...etc.

(spacemacs is opinionated after all)

kryptt:
@yyoncho Do we do this type of thing elsewhere?

Right now, this patch has a default as well as the ability to disable both backends (with notes in the README)... similar to how the c/c++ backends work...

yyoncho:
@kryptt first the maintainers of spacemacs must agree with that approach. IMO It is out of the scope of your PR. I think  that we may create a function which is like this:

```(spacemacs-ask-for-backend 'layer-backed-variable ((backend1 . true) (backend2 . nil)...))```

Which can be used by all layers with more than one backend. The function will inspect whether layer-backed-variable is populated and if it is not populated it will ask the user to select backend and initialize the variable.

**Comment by fabianhjr** [comment]
Hey, thanks a lot @kryptt, I cherrypicked you commit into my ~/.emacs.d and have been using metals 0.7.0 delightfully. :3

**Comment by yyoncho** [comment]
@kryptt  lsp-scala is now deprecated and now you should switch to the built in lsp-metals.el.

**Comment by kryptt** [comment]
@yyoncho I will add it as a third option and make it the default

**Comment by yyoncho** [comment]
IMO it will be better to remove lsp-scala at all since the package will be removed from melpa at some point.

File: layers/+lang/scala/packages.el [hunk 1]
```diff
@@ -52,7 +51,8 @@
   ;; Don't use scala checker if ensime mode is active, since it provides
   ;; better error checking.
   (with-eval-after-load 'flycheck
-    (add-hook 'ensime-mode-hook 'spacemacs//scala-disable-flycheck-scala)))
+    (add-hook 'ensime-mode-hook 'spacemacs//scala-disable-flycheck-scala)
+    (add-hook 'lsp-scala-)))
```
yyoncho: this line is not completed.

kryptt: Thanks, missed it

kryptt:
@yyoncho also addressed the issue you mentioned in gitter

layers/+lang/scala/packages.el
```diff
@@ -81,6 +86,20 @@
...
+      (unless (spacemacs//scala-backend-ensime)
+        (progn
+          (fmakunbound 'ensime)
+          (remove-hook 'after-change-functions 'ensime-after-change-function)
+          (remove-hook 'window-configuration-change-hook 'ensime-show-left-margin-hook)))
+
+      ;; Auto start metals
+      (if (spacemacs//scala-backend-metals)
+          (add-hook 'scala-mode-hook #'lsp)
+        (with-eval-after-load 'lsp
```
yyoncho: this is not needed - lsp layer is setting lsp-prefer-make to nil

kryptt: tweaked

layers/+lang/scala/funcs.el
```diff
@@ -17,6 +17,14 @@
 (defun spacemacs//scala-disable-flycheck-scala ()
   (push 'scala flycheck-disabled-checkers))
 
+(defun spacemacs//scala-backend-ensime ()
+  "Return true if the selected backend is ensime"
```
duianto:
```suggestion
  "Return t if the selected backend is ensime"
```
>In contexts where a truth value is expected, any non-nil value is considered to be true. However, t is the preferred way to represent the truth value true.

https://www.gnu.org/software/emacs/manual/html_node/elisp/nil-and-t.html

Another way to write it seems to be `"Return non-nil ...`, that seems to be the most used out of:
`t`, `true` or `non-nil` in Spacemacs.
These were the number of results when searching with `SPC /` (`HELM Ag: Using rg`)
```
"return.t\b
11

"return.true\b
4

"return.non-nil\b
29
```

The same Emacs manual conventions page also shows the return value as `non-nil`:
>— Function: booleanp object
Return non-nil if object is one of the two canonical boolean values: t or nil.

https://www.gnu.org/software/emacs/manual/html_node/elisp/nil-and-t.html

But it seems like `"Return t ...` probably would be better, because that's what the function returns. And an elisp beginner might expect `"Return non-nil ...` to return `non-nil`.

layers/+lang/scala/funcs.el
```diff
@@ -17,6 +17,14 @@
...
+  (eq scala-backend 'scala-ensime))
+
+(defun spacemacs//scala-backend-metals ()
```
duianto:
```suggestion
(defun spacemacs//scala-backend-metals-p ()
```

layers/+lang/scala/funcs.el
```diff
@@ -17,6 +17,14 @@
...
+  "Return true if the selected backend is metals"
```
duianto:
```suggestion
  "Return t if the selected backend is metals"
```

layers/+lang/scala/packages.el
```diff
@@ -25,7 +22,7 @@
         ))
 
 (defun scala/post-init-eldoc ()
-  (when scala-enable-eldoc
+  (when (and scala-enable-eldoc (spacemacs//scala-backend-ensime))
```
duianto:
```suggestion
  (when (and scala-enable-eldoc (spacemacs//scala-backend-ensime-p))
```

layers/+lang/scala/packages.el
```diff
@@ -35,10 +32,11 @@
 (defun scala/post-init-ensime ()
   (use-package ensime
     :defer t
+    :if (spacemacs//scala-backend-ensime)
```
duianto:
```suggestion
    :if (spacemacs//scala-backend-ensime-p)
```

layers/+lang/scala/packages.el
```diff
@@ -81,6 +86,18 @@
         (add-to-list 'completion-ignored-extensions ext)))
     :config
     (progn
+
+      ;; Ensure only one of metals and ensime is loaded
+      (unless (spacemacs//scala-backend-ensime)
```
duianto:
```suggestion
      (unless (spacemacs//scala-backend-ensime-p)
```

layers/+lang/scala/packages.el
```diff
@@ -68,6 +66,13 @@
 (defun scala/init-sbt-mode ()
   (use-package sbt-mode
     :defer t
+    :config
+    ;; WORKAROUND: https://github.com/ensime/emacs-sbt-mode/issues/31
+    ;; allows using SPACE when in the minibuffer
```
duianto:
```suggestion
    ;; allows for using SPACE in the minibuffer
```

layers/+lang/scala/packages.el
```diff
@@ -81,6 +86,18 @@
         (add-to-list 'completion-ignored-extensions ext)))
     :config
     (progn
+
+      ;; Ensure only one of metals and ensime is loaded
+      (unless (spacemacs//scala-backend-ensime)
+        (progn
+          (fmakunbound 'ensime)
+          (remove-hook 'after-change-functions 'ensime-after-change-function)
+          (remove-hook 'window-configuration-change-hook 'ensime-show-left-margin-hook)))
```
duianto: Add a line break to keep this line under 80 chars.
```suggestion
          (remove-hook 'window-configuration-change-hook
                       'ensime-show-left-margin-hook)))
```

layers/+lang/scala/packages.el
```diff
@@ -81,6 +86,18 @@
...
+
+      ;; Auto start metals
+      (if (spacemacs//scala-backend-metals)
```
duianto:
```suggestion
      (if (spacemacs//scala-backend-metals-p)
```

dmytro-altsyvanovych-exa:
Hi all,
Is there some work to merge this PR?

smile13241324:
@dmytro-altsyvanovych-exa hmm, as far as I can see there is still some unresolved discussions the PR author should address before this can be merged.

From the discussion so far it looks like the packages in question are now deprecated, I would still not remove them yet. However there is no harm in adding more backends to the layer.

So from us currently nobody is working on this actively for we are waiting for @kryptt.

layers/+lang/scala/packages.el
```diff
@@ -81,6 +88,15 @@
         (add-to-list 'completion-ignored-extensions ext)))
     :config
     (progn
+
+      ;; Ensure only one of metals and ensime is loaded
+      (unless (spacemacs//scala-backend-ensime)
+        (progn
+	  (fmakunbound 'ensime)
+	  (remove-hook 'after-change-functions 'ensime-after-change-function)
+	  (remove-hook 'window-configuration-change-hook 'ensime-show-left-margin-hook)))
```
duianto: Add a line break to keep this line under 80 chars.
```suggestion
	  (remove-hook 'window-configuration-change-hook
                 'ensime-show-left-margin-hook)))
```

layers/+lang/scala/packages.el
```diff
@@ -81,6 +88,15 @@
...
+
```
duianto: Remove the extra empty line.
```suggestion
```

layers/+lang/scala/packages.el
```diff
@@ -139,6 +155,18 @@ If it's part of a left arrow (`<-'),replace it with the unicode arrow."
             scala-indent:default-run-on-strategy
             scala-indent:operator-strategy))))
 
+(defun scala/init-lsp-scala ()
+  (use-package lsp-scala
+    :after scala-mode
+    :demand t
+    :if (spacemacs//scala-backend-metals)
```
duianto:
```suggestion
    :if (spacemacs//scala-backend-metals-p)
```

syl20bnr:
Once this PR is merged and Metals backend is confirmed to work as expected in Spacemacs then we will be able to remove Ensime backend entirely.

kryptt:
Sorry for not checking this issue recently.
@duianto I addressed all your comments. 

It works for me quite well, I did notice disabling the auto-complete layer gives you a warning about yasnippet but I didn't find a good way to address this...

teozkr:
@syl20bnr Metals still (AFAIK, from what I found in the changelogs) isn't a complete replacement for Ensime since it doesn't show resolved implicits.

duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/e2d9e32e92e49a777b4a486f65b950c0948d0c91

#### Additions
Added prefix names for:
- `SPC m b` as `sbt`
- `SPC m g` as `goto`

colin4124:
@kryptt As this point out https://github.com/emacs-lsp/lsp-mode/issues/973#issuecomment-523372005  

It should not use lsp-scala any more.

kryptt:
Oh, good to know thanks @colin4124

colin4124:
@kryptt So would you remove `lsp-scala` from `pacakges.el` ?

PR #12256, opened 2019-05-05
Fix void function error when recentf-save-list is undefined

Compro-Prasad:
This happens because there is an idle timer which runs `recentf-save-list` every
600 secs.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -951,6 +951,8 @@ Other:
   - Improved Window Manipulation Transient state (thanks to yuhan0)
   - Exclude which-key from layer sync powerline restore (thanks to Doug Wilson)
   - Fix doom-modeline in the messages buffer (thanks to duianto)
+  - Fix void function error when recentf-save-list is undefined (thanks to
+    Compro-Prasad)
 *** Layer changes and fixes
 **** Ansible
 - Improvements:
--- a/layers/+spacemacs/spacemacs-defaults/packages.el
+++ b/layers/+spacemacs/spacemacs-defaults/packages.el
@@ -274,6 +274,7 @@
 (defun spacemacs-defaults/init-recentf ()
   (use-package recentf
     :defer (spacemacs/defer)
+    :commands (recentf-save-list)
     :init
     (progn
       (spacemacs|require 'recentf)


Compro-Prasad:
Where should the CHANGELOG entry go?

duianto:
at the end of the `Fixed:` section. (the line above the header: **Layer changes and fixes**)

Compro-Prasad:
Done

duianto:
Does it happen the first time the timer runs for you?

I'm not seeing a void recentf-save-list message when the timer runs.

In both windows 10 and ubuntu 18.04 the messages buffer just shows:
>Saving file c:/Users/username/.emacs.d/.cache/recentf...
Wrote c:/Users/username/.emacs.d/.cache/recentf

(and the ubuntu equivalent path)

I don't know if this is a correct test but in the function:

https://github.com/syl20bnr/spacemacs/blob/a385f17efb145678ee391d1eb6b192dc2fa80175/layers/%2Bspacemacs/spacemacs-defaults/packages.el#L274

In the section:
```
  (use-package recentf
```

When this log message is added to the `:init` section:
```
      (message ":init fboundp recentf-save-list: %s" (fboundp 'recentf-save-list))
```
And this in the `:config` section:
```
      (message ":config fboundp recentf-save-list: %s" (fboundp 'recentf-save-list))
```

Then the current output is (without this PR applied):
>:init fboundp recentf-save-list: nil
:config fboundp recentf-save-list: t


With this PR applied:
>:init fboundp recentf-save-list: t
:config fboundp recentf-save-list: t

Since the timer runs after 10 minutes, shouldn't `recentf-save-list` be bound, since it already is bound when the `:config` section runs?

Compro-Prasad:
It only shows up when you don't trigger anything related to `recentf`. To test it just open spacemacs and wait for 10-15 mins. Do something else and switch back. The minibuffer will have the error.

duianto:
It doesn't seem to happen for me, these are the entries in the messages buffer. The last two appeared after the timer ran.
```
Loading c:/Users/username/.spacemacs...done
PNG warning: Interlace handling should be turned on when using png_read_image
Starting new Ispell process c:/tools/cygwin/bin/aspell.exe with default dictionary...
Spacemacs is ready.
Loading c:/Users/username/.emacs.d/.cache/recentf...done
Skipping check for new version (reason: dotfile)
Saving file c:/Users/username/.emacs.d/.cache/recentf...
Wrote c:/Users/username/.emacs.d/.cache/recentf
```

#### System Info :computer:
- OS: windows-nt
- Emacs: 26.2
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. a385f17ef)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(autohotkey emacs-lisp git helm multiple-cursors org spell-checking syntax-checking neotree version-control)
```

Compro-Prasad:
For me the whole `*Messages*` at startup is:
```
Loading /home/compro/.spacemacs.d/init.el (source)...done
c-c++: No c-c++-lsp-cache-dir specified: defaulting to /home/compro/.ihsec/spacemacs/.cache/lsp-ccls
‘dap’ layer is not installed, please add ‘dap’ layer to your dotfile.
(Spacemacs) Layer list has not changed since last time. Skipping dumping process!
Starting new Ispell process /usr/bin/aspell with default dictionary...
Spacemacs is ready.
Skipping check for new version (reason: dotfile)
Waiting for git... [2 times]
Error running timer ‘recentf-save-list’: (void-function recentf-save-list) [2 times]
```

duianto:
I wonder if it could be related to:
>Loading c:/Users/username/.emacs.d/.cache/recentf...done

It loads on startup for me in both Windows and Ubuntu.

Maybe that also binds recentf-save-list for some reason.

Compro-Prasad:
Which line in Spacemacs conig causes the loading to start?

duianto:
`(recentf-save-list)` seems to be called when the recent files list is added to the spacemacs home buffer.

In the function: `spacemacs-buffer//do-insert-startupify-lists`
https://github.com/syl20bnr/spacemacs/blob/8fb3be891db35e8c1af32ee2b8fbd4d26a712e89/core/core-spacemacs-buffer.el#L892

`(recentf-mode)` is defined in Emacs:
https://github.com/emacs-mirror/emacs/blob/31c60dfbd8541c9f1b1bc8127dde85e5d5af51b5/lisp/recentf.el#L1358

it calls `(recentf-save-list)` if `recentf-mode` is `nil`.

duianto:
Reproduction achievement unlocked!
>Error running timer ‘recentf-save-list’: (void-function recentf-save-list)

Removing the startup lists in `.spacemacs` causes the issue.

From:
```
                                (projects . 7))
```
To:
```
   dotspacemacs-startup-lists '()
```

I'm testing the PR now.

Compro-Prasad:
My config has the following:
```
   dotspacemacs-startup-lists '((bookmarks . 7)
                                (projects . 5))
```

duianto:
Yeah that's why, `(recentf-mode)` isn't called when `recents` isn't part of the startup list.

Confirmed this PR solves the void issue.

duianto:

PR #12267, opened 2019-05-07
[better-defaults] Remap `M-\` to `cycle-spacing`

bb2020:
`cycle-spacing` combines three whitespace commands into a single binding. It is described in [EmacsWiki](https://www.emacswiki.org/emacs/EmacsNiftyTricks).


duianto:
I wonder if it might even be better to change `M-SPC` (without the `better-defaults` layer) to `cycle-spacing`, since it wouldn't change the current behavior the first time `M-SPC` is pressed, but any additional presses would cycle through the three spacing amounts.

Don't change this PR yet, I'll ask the maintainers what they think.

bb2020:
Good idea. I totally forgot about `M-SPC` as it is bound to Spotlight on macOS. Also `M-\` is hard to press with regional keyboards. `M-SPC` is the way to go :ok_hand:

duianto:
The following PR is now on the `develop` branch:
Remap `M-SPC` and `SPC-x-d-SPC` to `cycle-spacing` #12268

It made this PR obsolete.

PR #12271, opened 2019-05-08
[shell] fix eshell-z home pollution

bb2020:
`eshell-z-freq-dir-hash-table-file-name` has the value of `~/.z` and writes into user home directory. It should be moved into spacemacs cache.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2129,6 +2129,8 @@ Other:
   - ~SPC m C~ switch multi-term to char mode
   - ~SPC m l~ switch multi-term to line mode
   - ~SPC m N~ go to previous multi-term
+- Moved =eshell-z-freq-dir-hash-table-file-name= into cache dir
+  (thanks to bb2020)
 **** Shell Scripts
 - Added new company-shell environment variable backend
   (thanks to Alexander-Miller)
--- a/layers/+tools/shell/packages.el
+++ b/layers/+tools/shell/packages.el
@@ -135,8 +135,11 @@
   (use-package eshell-z
     :defer t
     :init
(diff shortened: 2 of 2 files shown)


duianto:
`init-eshell` uses `.cache/eshell`:
https://github.com/syl20bnr/spacemacs/blob/54f2c85c1890f066155e73f9e4a9e5cac23bc978/layers/%2Btools/shell/packages.el#L75

Maybe `eshell-z` also should use the `.cache/eshell` directory.
Or if `eshell` and `eshell-z` related files should be kept separate, then maybe it could use `.cache/eshell-z`.

That would make it clear where the `.z` file comes from. I had no idea until I read this PR 😄

bb2020:
Yeah, `.z` is very confusing. At first, I thought it is some kind of virus. `.cache/eshell-z` seems better as `.cache/eshell` is default emacs eshell directory.

duianto:
While testing this I realized that it's probably simpler to use `.cache/eshell/.z`, then Emacs will take care of creating and checking that the directory exists.

The lines should be wrapped so that they don't extend past 80 chars:
```
      (setq eshell-z-freq-dir-hash-table-file-name
            (concat spacemacs-cache-directory "eshell/.z"))
```

Can you add an entry to `changelog.develop`, at the end of the `shell` section:
```
- Moved =eshell-z-freq-dir-hash-table-file-name= into cache dir
  (thanks to bb2020)
```

(The thanks parentheses would have been split in the middle to keep the line under 80 chars, therefore the whole parentheses is moved to the second line.)

bb2020:
Done

duianto:

PR #12273, opened 2019-05-08
fix finding find on windows

barcharcraz:
The projectile configuration would search for find.exe on windows and if found, enable alien mode for indexing. Unfortunately windows ships with find.exe that is NOT a find implementation (it's like grep by bad). This change makes spacemacs also check if it found find in the windows directory, and if so ignore it.


barcharcraz:
Actually it may be reasonable to just not do this detection since fd exists.

smile13241324:

Issue #78, opened 2014-11-05
Use Helm histories for eshell/comint/minibuffer


tuhdo:
History using Helm is provided by Helm:
- helm-eshell-history: 

``` lisp
(require 'helm-eshell)

(add-hook 'eshell-mode-hook
          #'(lambda ()
              (define-key eshell-mode-map (kbd "M-l")  'helm-eshell-history)))
```

![helm-eshell-history](https://cloud.githubusercontent.com/assets/4818719/4912996/1f653544-64a8-11e4-9147-96a7762cc372.gif)
- helm-mini-buffer-history: Display a list of all previous input entered into minibuffer.

``` lisp
(define-key minibuffer-local-map (kbd "C-c C-l") 'helm-minibuffer-history)
```
- helm-comint-input-ring: similar to `helm-eshell-history` but is used for `M-x shell`.

``` lisp
(define-key shell-mode-map (kbd "C-c C-l") 'helm-comint-input-ring)
```

PR #12274, opened 2019-05-08
[shell] Enable `helm-esh-pcomplete` for eshell tab completion

bb2020:
I am using `helm-esh-pcomplete` over a year and it seems stable. I think it is not broken anymore and ready for prime time.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2132,6 +2132,7 @@ Other:
   - ~SPC m N~ go to previous multi-term
 - Moved =eshell-z-freq-dir-hash-table-file-name= into cache dir
   (thanks to bb2020)
+- Enabled ~TAB~ completion in =eshell= with =Helm= (thanks to bb2020)
 **** Shell Scripts
 - Added new company-shell environment variable backend
   (thanks to Alexander-Miller)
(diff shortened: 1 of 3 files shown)


duianto:
It seems to be working:

helm-esh-pcomplete

The `<tab>` key binding for calling `helm-esh-pcomplete` might never have been enabled in Spacemacs. 

The commented out line seems to have been added in this commit (Dec 12, 2014): 
Fixes #78 Use helm histories for eshell/comint https://github.com/syl20bnr/spacemacs/commit/79302c3ca63d7dfe89e3d10a14696fed6e2ac5a1

duianto:
List `TAB` in the Eshell key binding documentation:
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+tools/shell/README.org#eshell-1

And add a changelog.develop entry:

duianto:
Currently it looks like this when `TAB` is pressed:

emacs_2019-05-08_23-54-53

I wonder if there are users that might be used to or prefer to use the current behavior.

Should there be a variable to enable or disable `helm-esh-pcomplete`?

bb2020:
As long as Emacs is concerned, probably :grin:
This command should achieve that. However it does not work.
`(setq-default dotspacemacs-configuration-layers
    '(org git shell
      (helm :disabled-for shell)))`

duianto:

PR #12277, opened 2019-05-09
notmuch: add note about the notmuch layout

jb55:
Adding this in case people aren't aware it exists. Pretty much can't use the notmuch layer without it.


duianto:
Maybe it would be better to add it as a key binding table, then it would match how the other key bindings are shown.

Possibly something like this:
```org
*** Notmuch Layout

| Key binding | Command                        |
|-------------+--------------------------------|
| ~SPC l o n~ | Open the notmuch custom layout |
```

duianto:
I totally missed that it's already listed in the Global bindings section 😄 
`SPC l o n | Start notmuch in a custom layout, “@Notmuch”`

https://github.com/syl20bnr/spacemacs/blob/develop/layers/+email/notmuch/README.org#global-bindings

jb55:
oops I must have missed that as well. thanks!

PR #12283, opened 2019-05-10
Add rubocopfmt to format Ruby buffers

sei40kr:
Added rubocopfmt for Ruby layer.
`rubocop-autocorrect-current-file` in `rubocop` shows a result buffer on every format, so it's uncomfortable to use it as a formatter.


sei40kr:
I avoided to use `SPC m = b` to format buffer, because it is overridden by lsp, and there're some working for Ruby LSP backend, I assigned `SPC m = r` for RuboCop format.

duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/80ff9053a3545f6d7dbad5fd6446ff7b1bc8442c

PR #12285, opened 2019-05-10
[WIP] Add npm-and-yarn to node layer

sei40kr:
Enable to install a Node package in Spacemacs.

# New Key Bindings

| Key binding | Description                                              |
|:-------------|:----------------------------------------------------------|
| `SPC m n i` | install and save a package using npm                     |
| `SPC m n I` | install and save a package to `devDependencies` via npm  |
| `SPC m y a` | install and save a package using yarn                    |
| `SPC m y A` | install and save a package to `devDependencies` via yarn |

### DIFF
--- a/layers/+tools/node/local/npm-and-yarn/npm-and-yarn.el
+++ b/layers/+tools/node/local/npm-and-yarn/npm-and-yarn.el
@@ -1,0 +1,54 @@
+;; -*- lexical-binding: t -*-
+
+;;; npm-and-yarn.el
+
+;; Author:   Seong Yong-ju <sei40kr@gmail.com>
+;; Created:  May 10 2019
+;; Version:  0.1.0
+;; Keywords: node, npm, yarn
+
+;;; Commentary:
+
+;;; Code:
+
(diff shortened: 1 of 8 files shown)


layers/+tools/node/local/npm-and-yarn/npm-and-yarn.el
```diff
@@ -0,0 +1,54 @@
+;; -*- lexical-binding: t -*-
+
+;;; npm-and-yarn.el
+
```
smile13241324: Please add a standard Melpa package header, otherwise we cannot accept it into emacs.
Do you also intent to upload the package to Melpa? If it is so I would prefer to replace the locale version once it is uploaded.

robbyoconnor:
You have a conflict.

duianto:
Let's close this PR, since it probably won't be updated, the author said this in another PR: 
https://github.com/syl20bnr/spacemacs/pull/12970#issuecomment-586555829
>I'm not using Spacemacs now and I have no motivation to update this. Sorry.

PR #12288, opened 2019-05-11
Configure emmet-mode for typescript-tsx-mode

sei40kr:
In `typescript-tsx-mode` buffer, `emmet-mode` is enabled (derived from `web-mode`), but it was not configured for TSX syntax.


duianto:

PR #12290, opened 2019-05-11
[shell] Add ivy completion

bb2020:
Added Eshell completion with Ivy. It is similar to existing Helm completion.

### DIFF
--- a/layers/+tools/shell/funcs.el
+++ b/layers/+tools/shell/funcs.el
@@ -174,6 +174,23 @@ is achieved by adding the relevant text properties."
   (define-key eshell-mode-map
     (kbd "M-l") 'spacemacs/helm-eshell-history))
 
+(defun spacemacs/ivy-eshell-history ()
+  (interactive)
+  (counsel-esh-history)
+  (evil-insert-state))
+
+(defun spacemacs/pcomplete-std-complete ()
+  (interactive)
+  (pcomplete-std-complete)
+  (evil-insert-state))
+
+(defun spacemacs/init-ivy-eshell ()
+  "Initialize ivy-eshell."
+  (spacemacs/set-leader-keys-for-major-mode 'eshell-mode
+    "H" #'spacemacs/ivy-eshell-history)
+  (define-key eshell-mode-map (kbd "M-l") #'spacemacs/ivy-eshell-history)
+  (define-key eshell-mode-map (kbd "<tab>") #'spacemacs/pcomplete-std-complete))
+
 (defun term-send-tab ()
   "Send tab in term mode."
   (interactive)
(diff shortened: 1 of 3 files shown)


layers/+tools/shell/funcs.el
```diff
@@ -174,6 +174,20 @@ is achieved by adding the relevant text properties."
...
+  ;; (evil-insert-state)
+  )
+
+(defun spacemacs/init-ivy-eshell ()
+  "Initialize ivy-eshell."
+  (spacemacs/set-leader-keys-for-major-mode 'eshell-mode
+    "H" #'spacemacs/ivy-eshell-history)
+  (define-key eshell-mode-map (kbd "M-l") #'spacemacs/ivy-eshell-history)
+  (define-key eshell-mode-map (kbd "<tab>")
+    (lambda () (interactive) (pcomplete-std-complete))))
```
duianto: Maybe this should be a named function, like `spacemacs/ivy-eshell-history`. 

At least one case where a named function would be helpful is in the ` C-h l` (lower case L) (`view-lossage`) buffer.

With a lamda the lossage entry looks like this:
> \<tab> [anonymous-command]

With a named function, for example like this:
```
(defun spacemacs/pcomplete-std-complete ()
  (interactive)
  (pcomplete-std-complete))
```

And in `spacemacs/init-ivy-eshell`
```
  (define-key eshell-mode-map (kbd "<tab>") #'spacemacs/pcomplete-std-complete)
```

Then the entry is more descriptive:
> \<tab> [spacemacs/pcomplete-std-complete]

bb2020: Definitely. I rarely use `view-lossage` and often overlook what "lossage" shows.

layers/+tools/shell/funcs.el
```diff
@@ -174,6 +174,23 @@ is achieved by adding the relevant text properties."
   (define-key eshell-mode-map
     (kbd "M-l") 'spacemacs/helm-eshell-history))
 
+(defun spacemacs/ivy-eshell-history ()
+  (interactive)
+  (counsel-esh-history)
+  ;; (evil-insert-state)
```
duianto: Maybe this should be uncommented or removed.

`RET` doesn't run the command in normal state.

But it might be useful to insert a command from the history (for example a long path) and edit it in normal state before entering insert state and pressing `RET`.

bb2020: Right, now I tested with Evil mode. `M-l` may switch to insert state while `TAB` stays at normal state as the command name would be edited right away to add arguments. However this may be confusing. I think both should switch to insert state, similar to Helm completion.

duianto:

PR #12295, opened 2019-05-12
[ivy] add variable ret-visits-directory

bb2020:
https://github.com/emacs-helm/helm/issues/1480

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1583,6 +1583,7 @@ Other:
 - Fixed counsel search errors due to function rename (thanks to Aaron Jensen)
 - Display results count properly in prompt (thanks to yuhan0)
 - Fixed typos in spacemacs/search-dir* bindings (thanks to yuhan0)
+- Added =ivy-ret-visits-directory= variable (thanks to bb2020)
 **** Imenu-list
 - Fixed key bindings (thanks to bet4it):
   - ~r~ inside imenu-list buffer to refresh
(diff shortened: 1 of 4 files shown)


duianto:
Does the readme need to mention that the variable (maybe it applies to all the ivy layer variables), need to be set in the `dotspacemacs-configuration-layers`:
`(ivy :variables ivy-ret-visits-directory t)`
or in the user-init section:
`(setq ivy-ret-visits-directory t)`

It doesn't work in the `user-config` section.

`M-m f f`
partially typed path (full path: `~/temp/`):
`~/tem`
and pressing `RET` and `C-j`.

`(ivy :variables ivy-ret-visits-directory t)`
or
user-init
    `(setq ivy-ret-visits-directory t)`

`C-h l` shows:
```
<return> [ivy-alt-done]
C-j [ivy-done]
```

user-config
    `(setq ivy-ret-visits-directory t)`
or if the variable isn't set and it's `nil`

`C-h l` shows:
```
<return> [ivy-done]
C-j [ivy-alt-done]
```

Or is it assumed that the variables should be applied in the configuration layers or user init sections?

bb2020:
This is kinda expected because the variable does not belong to ivy package. Other variables, for example `ivy-wrap`, comes directly from the package. However `ivy-ret-visits-directory` is a spacemacs customization on top of ivy. If it has not been activated during layer config, then it can't be activated anymore.

duianto:

PR #12300, opened 2019-05-13
Remove line-feed characters

robbyoconnor:
Remove line-feeds from the elisp files -- I noticed the ^L all over the place

Used the following to do it:

``` shell
find -type f -name *.el -not -path "./.cache/*" -not -path "./elpa/*" -exec sed -i 's/\f$//' {} \; -print
find -type f -name *.el -not -path "./.cache/*" -not -path "./elpa/*" -exec sed -i 's/\r$//' {} \; -print
find -type f -name *.el -not -path "./.cache/*"  -not -path "./elpa/*" -not -path "./tests/*" -exec sed -i 's/[[:space:]]*$//'  {} \; -print
```

### DIFF
--- a/core/core-debug.el
+++ b/core/core-debug.el
@@ -15,7 +15,7 @@
   (let ((msg (format "Spacemacs v.%s" spacemacs-version)))
     (message msg) (kill-new msg)))
 
-
+
 ;; startup debug
 
 (require 'profiler)
@@ -149,7 +149,7 @@ seconds to load")
   ;; Keep debug-on-error on for stuff that is lazily loaded
   (add-hook 'after-init-hook (lambda () (setq debug-on-error t))))
 
-
+
 ;; Report issue
 
 (defun spacemacs//describe-system-info-string ()
--- a/core/core-funcs.el
+++ b/core/core-funcs.el
@@ -331,7 +331,7 @@ current frame."
   (let ((comint-buffer-maximum-size 0))
     (comint-truncate-buffer)))
 
-
+
 ;; Generalized next-error system ("gne")
 
 (defun spacemacs/error-delegate ()
--- a/core/libs/package-build.el
+++ b/core/libs/package-build.el
@@ -1325,7 +1325,7 @@ Returns the archive entry for the package."
                                   " "))
       (package-build--url-copy-file badge-url badge-filename t))))
 
-
+
 ;;; Helpers for recipe authors
 
 (defvar package-build-minor-mode-map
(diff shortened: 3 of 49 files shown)


robbyoconnor:
I also removed carriage returns as well -- nyan-mode is affected by this

robbyoconnor:
cc @duianto @smile13241324 @JAremko @sdwolfz @syl20bnr

duianto:
There was a recent commit (https://github.com/syl20bnr/spacemacs/commit/bd91a0b2b8872d5fe34e30fd155e85a2c3efc257) that removed the form-feed characters from the haskell layer.

syl20bnr asked in the maintainer chat, why they were removed, since they are used in a lot of places, including in the Spacemacs home buffer.

Therefore I reverted the removal. https://github.com/syl20bnr/spacemacs/commit/e726a70685938a94b0e23800453feb14f4d16ba9

But syl20bnr also asked if they were removed because they were causing an issue.

In this case the removal seems to just have been a misunderstanding of this discussion about the `^L` characters: https://github.com/syl20bnr/spacemacs/pull/11550#discussion_r230598804

They were just talking about them but never requesting their removal as the removal author thought.

I'm guessing that they aren't critical to keep if they are causing issues.

What's the issue with `nyan-mode`?

robbyoconnor:
Do you want me to close this @duianto? 

@syl20bnr -- your call - I just scratched an itch I had...

robbyoconnor:
This also removed trailing whitespace.

duianto:
The `^L` characters are probably there on purpose to separate different sections.

There are also keys/commands for jumping by page `C-x [` (`backward-page`) and `C-x ]` (`forward-page`).

Is the `nyan-mode` package pulled from another other repository or is it handled locally in Spacemacs?

I'm guessing locally. Then you can decide if you want to modify this PR or open a new one.

robbyoconnor:
I'll wait until @syl20bnr chimes in -- I'm fine with this sitting.

duianto:
You can do both.

`nyan-mode.el` can be extracted to a new PR (it should probably be a separate PR since it's unrelated to the form-feed character removal).

Then that can be fixed while you wait for a response to this PR.

syl20bnr:
Thank you for the PR and to wait for me on this one.
There is form feeds everywhere in Emacs and we even have them in the home buffer, so let's keep them ;-)
If there is other associated fixes like whitespaces you can open a new PR with them.

robbyoconnor:
Perfectly okay -- closing this.

PR #12301, opened 2019-05-13
notmuch: fix performance issues with light theme

jb55:
box borders for unread items seems to cause performance issues with
the light theme.

This is also a strange default anyways...


duianto:
The https://github.com/nashamri/spacemacs-theme is handled by it's author. Then the author opens a PR here with any updates. So you'll have to open a PR over there.

It also helps to post before and after screenshots.

jb55:
good to know, thx

Issue #12168, opened 2019-04-12
Org-journal: no shortcut for "open today’s journal without creating a new entry"


fladd:
There probably should be a shortcut for this. Currently the only shortcuts are for creating a new entry and searching through the journal.

Miciah:
Use `org-journal-new-entry` with a prefix: <kbd>SPC u</kbd> <kbd>SPC a o j j</kbd>.

PR #12302, opened 2019-05-14
[org] Document how to open today's journal

Miciah:
Fix issue #12168 by documenting that the <kbd>SPC u</kbd> <kbd>SPC a o j j</kbd> key binding can be used to open today's journal without adding a new entry.

* `layers/+emacs/org/README.org`: Add key binding.

### DIFF
--- a/layers/+emacs/org/README.org
+++ b/layers/+emacs/org/README.org
@@ -722,10 +722,11 @@ org-present must be activated explicitly by typing: ~SPC SPC org-present~
 
 ** Org-journal
 
-| Key binding   | Description            |
-|---------------+------------------------|
-| ~SPC a o j j~ | New journal entry      |
-| ~SPC a o j s~ | Search journal entries |
+| Key binding         | Description                                     |
+|---------------------+-------------------------------------------------|
+| ~SPC a o j j~       | New journal entry                               |
+| ~SPC u SPC a o j j~ | Open today's journal without adding a new entry |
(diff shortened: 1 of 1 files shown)


duianto:
It's probably good to document that the command has a secondary behavior.

But it would be more discoverable from inside Spacemacs, if it had a dedicated key binding.

Possibly: `SPC a o j o`

A quick test shows that this seems to work:
```
    (defun spacemacs/org-journal-open ()
      (interactive)
      (org-journal-new-entry t))

    (spacemacs/set-leader-keys
      "aojo" 'spacemacs/org-journal-open)
```

duianto:
There's also an external `spacemacs-journal` layer (archived):
https://github.com/ruibinx/spacemacs-journal

The suggested key binding for opening todays journal is:
`SPC a j v` for  `view journal of today`

But maybe there's a risk that `v` as in view, could be interpreted as opening it in `view-mode` (read-only)?

#### Sidenote
The `spacemacs-journal` layer might have been created by the author of [org-journal](https://github.com/bastibe/org-journal), because the link from the `Can I use org-journal with Spacemacs?` section:
https://github.com/bastibe/org-journal#can-i-use-org-journal-with-spacemacs

Shows the url to spacemacs-journal as: https://github.com/borgnix/spacemacs-journal
It just redirects to the `ruibinx` users repository.

smile13241324:

PR #12312, opened 2019-05-16
Update dash layer to reflect dash-docs merge

gilbertw1:
This pull request updates the `dash` layer to reflect recent changes to `helm-dash` and `counsel-dash`. Primarily the separation of `helm-dash`'s core functionality into a separate package `dash-docs`, which removed the dependency between `counsel-dash` and `helm-dash`.

This update primarily handles updating variable names from the `helm-dash-*` variant to the common `dash-docs-*` variant.

A few notes:
* I've added an alias for `helm-dash-docset-newpath` to `dash-docs-docset-newpath` to keep existing users' configuration valid.
* I removed the `counsel-dash-at-point` as this function was added to `counsel-dash` during the transition.

Thanks!

### DIFF
--- a/layers/+readers/dash/README.org
+++ b/layers/+readers/dash/README.org
@@ -50,13 +50,13 @@ The result will be displayed in the offline browser’s UI.
 
 However having to leave emacs to have a look at the search results may be a bit awkward.
 To help with this it is also possible to integrate the search results directly in =helm= or =ivy=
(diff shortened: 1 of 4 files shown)


bhipple:
This looks nice, but it needs to be rebased onto the latest `develop` to fixup merge conflicts.

PR #12316, opened 2019-05-17
Add org-cut-subtree key-bind to org layer

clwgg:
This binds `org-cut-subtree` to `SPC m s d`, similar in spirit to other 'delete'
bindings in evil-mode emacs.


duianto:

PR #12317, opened 2019-05-17
Workspaces TS: Add single win workspace key

duianto:
Updated the Workspaces Transient State `SPC l w`.

Added a new key: `[s/S] single win workspace` to the Actions column.

It creates a new workspace with a single window, that shows the
Spacemacs home buffer.

Moved the existing key: `[c/C] create workspace` to the Actions column.


smile13241324:
Thank you for contributing to spacemacs :purple_heart:, I have cherry-picked your commit into develop. You can safely remove your branch.

PR #12329, opened 2019-05-18
Add keybindings for `check-parens` and `goto-last-change`

bb2020:
Both commands don't have keybindings. `check-parens` is built-in emacs function. `goto-last-change` comes from Evil and assigned as a top-level binding to jump rapidly through edits.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -982,6 +982,9 @@ Other:
     - =spacemacs-theme-comment-bg=
     - =spacemacs-theme-org-height=
       (thanks to Dominik Schrempf)
+  - Key bindings (thanks to bb2020)
+    - ~SPC-j-C~ to =check-parens=
+    - ~SPC-,~ to =goto-last-change=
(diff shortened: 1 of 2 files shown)


duianto:
`goto-last-change` is currently bound to `g ;` and there's also
`goto-last-change-reverse` which is bound to `g ,`.

But the `g` prefix might not be available for Emacs editing style users, so it would be good to add another binding for them.

Would it be helpful to make a Goto Last Transient State?
`SPC ,` could both open the transient state and move to the last change the first time it's pressed.

Maybe the TS could bind `,` (comma) for last change and `.` (period) for last change reverse.

bb2020:
Yes, `goto-last-change` is not available for Emacs mode. TS would be good. Maybe we should open an issue for TS and look into it later.

Squashed two commits by the way.

duianto:
Yeah it can be a possible future feature.

smile13241324:

PR #12334, opened 2019-05-19
Do not delete window on closing term

demokritos:
When term or ansi-term mode buffer is closing, its window is also deleted. This
somewhat surprises the user. In addition, this does not applied to multi-term
mode and eshell mode. Users who split windows on purpose wouldn't like this
behavior. And since pop shell mode exists, users have an option.

### DIFF
--- a/layers/+tools/shell/README.org
+++ b/layers/+tools/shell/README.org
@@ -17,6 +17,7 @@
   - [[#enable-em-smart-in-eshell][Enable em-smart in Eshell]]
   - [[#protect-your-eshell-prompt][Protect your Eshell prompt]]
   - [[#fish-shell-and-ansi-term][Fish shell and ansi-term]]
+  - [[#close-window-with-terminal][Close window with terminal]]
 - [[#eshell][Eshell]]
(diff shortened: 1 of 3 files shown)


smile13241324:
@demokritos I have just had a look at your PR and noticed that CI is failing completely with wrong base, are you sure that your PR is based on `develop`?

demokritos:
I think so. I just rebased to develop, but a check fails.

smile13241324:
@demokritos thanks for rebasing. 

I agree that having two terminals have a different closing behaviour is bad. However I would prefer to make the default behaviour configurable not to produce surprises with old configs.

So could you please:
- Add a new layer variable `close-windows-with-terminal`.
- If this variable is nil (default) do not close the window when the terminal is stopped.
- When it is set, close the window when the terminal is closed.
- Document this in the readme with an example of how to set it in the dotfile.

When this is done I would gladly merge your PR.

demokritos:
I have modified the PR as you suggested. Please review it.
Thanks.

smile13241324:

PR #12339, opened 2019-05-20
Add transient state keys _, j, k

duianto:
Thanks sdwolfz for the suggestion to add underscore to the transient states,
then the shift key doesn't have to be released if it was held down to press `+`.

#### Added key bindings:
Evil numbers:
`SPC n _` decrease number under point
in the transient state:
`j` decrease number under point
`k` increase number under point

`SPC T T _` decrease transparency
`SPC z x _` scale down font
`SPC z f _` zoom out frame

#### Update documentation:
Zoom frame:
`q` quit transient state

#### Evil numbers:
`SPC n =` increase number under point
`SPC n _` decrease number under point
`0..9` add a number prefix argument
`q` quit transient state

Fixed a typo and made a small rewrite of the tip about using a prefix argument.

#### Font scaling:
Added `_` scale down
Reordered `k` assignment to match the order in the transient state

#### Frame transparency:
Added `_` decrease
Reordered `k` assignment to match the order in the transient state

#### Evil numbers:
Added `k` increase
Added `_` and `j` decrease

#### Zoom frame:
Added `_` zoom frame out
Reordered `k` assignment to match the order in the transient state

#### Which-key entries for Evil numbers `SPC n`:
Group together keys that call the same command:
From: `+` and `=`
To:   `+,=`

From: `-` and `_`
To:   `-,_`

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -782,6 +782,12 @@ Other:
     - ~q~ exits the transient state
       (thanks to kenkangxgwe)
   - Added =link-hint-copy-link= to ~SPC x y~ (thanks to William Casarin)
+  - Added key bindings (thanks to duianto):
+    - Evil numbers ~SPC n _~ decrease number at point
+      and in the transient state ~j~ and ~k~
+    - Font scaling ~SPC z x _~ scale down
+    - Frame transparency ~SPC T T _~ decrease transparency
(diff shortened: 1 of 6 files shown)


smile13241324:
Cherry picked into develop :+1:, you can safely remove your branch.

PR #12341, opened 2019-05-21
[dap] add dap-edge and dap-node for javascript debugging

winsphinx:
Since dap-mode has Microsoft-edge and native-nodejs, adding them to js layer.

[emacs-lsp/dap-mode](https://github.com/emacs-lsp/dap-mode#javascript-1)

### DIFF
--- a/layers/+lang/javascript/funcs.el
+++ b/layers/+lang/javascript/funcs.el
@@ -38,6 +38,8 @@
       (progn
         (require 'dap-firefox)
         (require 'dap-chrome)
+        (require 'dap-edge)
+        (require 'dap-node)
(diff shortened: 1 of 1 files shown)


yyoncho:
IMO we should have separate variables to enable each dap backend otherwise users will see a message that they havent installed a particular dap backend even if they are not interested in it(e. g. if you are on linux you do not want edge adapter). Something like:

```
(when javascript-dap-chrome
 (require 'dap-chrome))
```

By default, we may have chrome and node adapter enabled. In the future, dap will install automatically the required backend when you use it for the first time but for now having these additional variables look like a good approach to avoid dap-mode limitations.

PR #12349, opened 2019-05-22
[search-engine] Add `spacemacs/search-engine-default` function

bb2020:
This PR adds `spacemacs/search-engine-default` function into `search-engine` layer and binds it to `SPC .`. It looks like the binding is free. This is very similar to `spacemacs/default-pop-shell` from `shell` layer which is bound to `SPC '`. Default search function reads current line if there is not an active region. Otherwise it just reads the region. This is good for quick web searches with a simple binding. Line searching is great for searching error messages easily (for example, a line from compile log).

The variable `search-engine-default-entry` configures the engine function to call.
`(setq-default dotspacemacs-configuration-layers
'((search-engine :variables search-engine-default-entry 'wikipedia)))`

Also I've added google-ncr source to google search without region redirect. This may not be beneficial for everyone. Waiting for suggestions for that should I remove it and put it into my config.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2145,6 +2145,8 @@ Other:
 **** Scheme
 - Added missing =parinfer= package declaration (thanks to Kalle Lindqvist)
 - Update install docs for Chicken 5 changes (thanks to Grant Shangreaux)
+**** Search Engine
+- Added =search-engine-default= function (thanks to bb2020)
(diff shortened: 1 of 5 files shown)


bb2020:
I've re-discovered [google-this](https://github.com/Malabarba/emacs-google-this) package and started using it again. Looks like it almost does the same job so I'm closing this for now.

PR #12350, opened 2019-05-22
Update Helm Transient State, show g and G keys

duianto:
Revealed already bound navigation keys:
- `g` go to first candidate
- `G` go to last candidate

Moved `[q] quit` from bottom left to bottom right, to match other TSs.
Reduced spacing between keys and descriptions from two to one space.


smile13241324:
Thank you for contributing to Spacemacs :purple_heart:, I have cherry picked your commit to develop. You can safely remove your branch.

PR #12355, opened 2019-05-23
Allow middle mouse click pasting in term mode

yangsheng6810:
Middle mouse click does not paste into term mode, i.e. term, ansi-term, multi-term. This fixes the behaviour.


smile13241324:
Thank you for contributing to Spacemacs :purple_heart:, I have cherry picked your commit into develop. You can safely delete your branch.

PR #12361, opened 2019-05-24
[multiple-cursors] Add magnars's multiple-cursors backend

bb2020:
`multiple-cursors` is amazing and a must have for spacemacs.

### DIFF
--- a/layers/+misc/multiple-cursors/README.org
+++ b/layers/+misc/multiple-cursors/README.org
@@ -20,12 +20,16 @@ add =multiple-cursors= to the existing =dotspacemacs-configuration-layers= list 
 file.
 
 * Configuration
-Currently the only supported backend is =evil-mc=, but more backends will be
-available in the future.
+Currently supported backends are:
+- [[https://github.com/gabesoft/evil-mc][evil-mc]] (default)
+- [[https://github.com/magnars/multiple-cursors.el][mc]]
+
+To set your choice of backend, configure =multiple-cursors-backend= variable of
+the layer.
 
 #+BEGIN_SRC emacs-lisp
   (setq-default dotspacemacs-configuration-layers '(
-    (multiple-cursors :variables multiple-cursors-backend 'evil-mc))
+    (multiple-cursors :variables multiple-cursors-backend 'mc))
 #+END_SRC
 
 * Key bindings
@@ -60,3 +64,44 @@ For easy navigation you also have the following:
 | ~C-t~       | evil-mc-skip-and-goto-next-match   |
 | ~C-M-j~     | evil-mc-make-cursor-move-next-line |
 | ~C-M-k~     | evil-mc-make-cursor-move-prev-line |
+
+** =multiple-cursors=
+The =multiple-cursors= backend provides the following key bindings to
+insert new cursors:
+
+| Key binding | Description                     |
+|-------------+---------------------------------|
+| ~SPC s m m~ | mc/mark-more-like-this-extended |
+| ~SPC s m r~ | mc/edit-lines                   |
+| ~SPC s m b~ | mc/mark-all-like-this           |
+| ~SPC s m a~ | mc/mark-all-dwim                |
+
+These special bindings manipulate text under cursors:
+
+| Key binding   | Description                   |
+|---------------+-------------------------------|
+| ~SPC s m s s~ | mc/sort-regions               |
+| ~SPC s m s r~ | set-rectangular-region-anchor |
+| ~SPC s m s m~ | mc/mark-sgml-tag-pair         |
+| ~SPC s m s n~ | mc/insert-numbers             |
+| ~SPC s m s l~ | mc/insert-letters             |
+| ~SPC s m s t~ | mc/reverse-regions            |
+
+* Notes
+** =multiple-cursors=
+Some commands executed during =multiple-cursors= enabled may lead Emacs to go frenzy.
+Commands like window manipulation will be executed multiple times if =mc= is active.
+Not all Emacs commands may be compatible with =mc=.
+
+To run interactive ~M-x~ commands with =mc=, run the command first. This will result
+with application of the command to the leading cursor. Then press =C-:= to apply the
+command to consequtive cursors. If this is not the case or you want a command to be
+executed only once, configure the =mc/cmds-to-run-once= variable of the layer like
+in the following example.
+
+#+BEGIN_SRC emacs-lisp
+  (setq-default dotspacemacs-configuration-layers '(
+    (multiple-cursors :variables
+                      multiple-cursors-backend 'mc
+                      mc/cmds-to-run-once '(upcase-region))))
+#+END_SRC
--- a/layers/+misc/multiple-cursors/packages.el
+++ b/layers/+misc/multiple-cursors/packages.el
@@ -13,8 +13,8 @@
 
 (setq multiple-cursors-packages
       '(
-        evil-mc
-        ))
+        (evil-mc :toggle (eq multiple-cursors-backend 'evil-mc))
+        (multiple-cursors :toggle (eq multiple-cursors-backend 'mc))))
 
 (defun multiple-cursors/init-evil-mc ()
   (use-package evil-mc
@@ -34,3 +34,28 @@
           (define-key state-map (car keybinding) (cdr keybinding))))
       (add-hook 'prog-mode-hook 'turn-on-evil-mc-mode)
       (add-hook 'text-mode-hook 'turn-on-evil-mc-mode))))
+
+(defun multiple-cursors/init-multiple-cursors ()
+  (use-package multiple-cursors
+    :defer t
+    :init
+    (progn
+      (spacemacs/declare-prefix "sm" "multiple-cursors")
+      (spacemacs/declare-prefix "sms" "specials")
+      (spacemacs/set-leader-keys
+        "smm" 'mc/mark-more-like-this-extended
+        "smr" 'mc/edit-lines
+        "smb" 'mc/mark-all-like-this
+        "sma" 'mc/mark-all-dwim
+        "smss" 'mc/sort-regions
+        "smsr" 'set-rectangular-region-anchor
+        "smsm" 'mc/mark-sgml-tag-pair
+        "smsn" 'mc/insert-numbers
+        "smsl" 'mc/insert-letters
+        "smst" 'mc/reverse-regions)
+      (setq mc/always-run-for-all t)
+      (with-eval-after-load 'multiple-cursors-core
+        (add-to-list 'mc/cmds-to-run-once 'helm-M-x)
+        (add-to-list 'mc/cmds-to-run-once 'counsel-M-x)
+        (add-to-list 'mc/cmds-to-run-once 'spacemacs/default-pop-shell)
+        ))))
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1773,6 +1773,7 @@ Other:
 **** Multiple Cursors
 - Disabled Spacemacs paste transient state when pasting to avoid pasting issue
   when there are more than one cursor (thanks to braham-snyder)
+- Added magnars's multiple-cursors package as a new backend (thanks to bb2020)
 **** Neotree
 - Made neotree an optional instead of a default layer
 - Move neotree to its own layer in new +filetree folder


layers/+misc/multiple-cursors/README.org
```diff
@@ -20,8 +20,7 @@ add =multiple-cursors= to the existing =dotspacemacs-configuration-layers= list
 file.
 
 * Configuration
-Currently the only supported backend is =evil-mc=, but more backends will be
-available in the future.
+Currently supported backends are =evil-mc= (default) and =mc=.
```
duianto: Maybe the text `evil-mc` and `mc` could link to the packages:
https://github.com/gabesoft/evil-mc
https://github.com/magnars/multiple-cursors.el

Should we add parentheses after `mc`  to indicate that it stands for `(multiple cursors)` or possibly even 
`(magnars multiple-cursors)` so as not to confuse it with the layer name.

layers/+misc/multiple-cursors/README.org
```diff
@@ -60,3 +59,43 @@ For easy navigation you also have the following:
 | ~C-t~       | evil-mc-skip-and-goto-next-match   |
 | ~C-M-j~     | evil-mc-make-cursor-move-next-line |
 | ~C-M-k~     | evil-mc-make-cursor-move-prev-line |
+|             |                                    |
```
duianto: Replace the last empty table row with an empty line, to leave a space between the key binding table and the following header.

layers/+misc/multiple-cursors/README.org
```diff
@@ -60,3 +59,43 @@ For easy navigation you also have the following:
...
+
+| Key binding | Description                   |
+|-------------+-------------------------------|
+| ~g r s s~   | mc/sort-regions               |
+| ~g r s r~   | set-rectangular-region-anchor |
+| ~g r s m~   | mc/mark-sgml-tag-pair         |
+| ~g r s n~   | mc/insert-numbers             |
+| ~g r s l~   | mc/insert-letters             |
+| ~g r s t~   | mc/reverse-regions            |
+* Notes
```
duianto: Add a newline to leave a space between the key binding table and the following header.

layers/+misc/multiple-cursors/packages.el
```diff
@@ -34,3 +34,25 @@
           (define-key state-map (car keybinding) (cdr keybinding))))
       (add-hook 'prog-mode-hook 'turn-on-evil-mc-mode)
       (add-hook 'text-mode-hook 'turn-on-evil-mc-mode))))
+
+(defun multiple-cursors/init-multiple-cursors ()
+  (use-package multiple-cursors
+    :defer t
```
duianto: Should `:defert t` be removed?
The `evil-mc` use-package section doesn't have it.

And the `multiple-cursors` package might need to be loaded for prog and text mode with hooks like `evil-mc`:
```
(add-hook 'prog-mode-hook 'turn-on-evil-mc-mode)
(add-hook 'text-mode-hook 'turn-on-evil-mc-mode)
```

layers/+misc/multiple-cursors/packages.el
```diff
@@ -34,3 +34,25 @@
...
+    :init
+    (progn
+      (spacemacs/set-leader-keys "grr" 'mc/mark-more-like-this-extended)
```
duianto: This binds the keys to `SPC g r r` (but they are probably overwritten by the `git` layer, because it's still bound to the expected command: `spacemacs/smerge-transient-state/body`

`evil-mc` seems to bind the keys to an `evil-mc-key-map` variable:
https://github.com/gabesoft/evil-mc/blob/master/evil-mc.el#L79

layers/+misc/multiple-cursors/README.org
```diff
@@ -60,3 +59,43 @@ For easy navigation you also have the following:
...
+Not all Emacs commands may be compatible with =mc=.
+
+To run interactive ~M-x~ commands with =mc=, run the command first. This will result
+with application of the command to the leading cursor. Then press =C-:= to apply the
+command to consequtive cursors. If this is not the case or you want a command to be
+executed only once, configure the =mc/cmds-to-run-once= variable of the layer like
+in the following example.
+
+#+BEGIN_SRC emacs-lisp
+  (setq-default dotspacemacs-configuration-layers '(
```
duianto: Maybe the first line should be excluded:
`(setq-default dotspacemacs-configuration-layers '(`
(and it's closing parenthesis)

Because there has been questions in the gitter chat about people trying to add the whole code block to their user-config section and running into issues.

There could possibly be a mention above that the code segment, that it should be added inside the 
`dotspacemacs-configuration-layers` list, near the top of `.spacemacs`.

And the indentation of the command should probably match how `elisp-mode` indents them:
```
(multiple-cursors :variables
                  multiple-cursors-backend 'mc
                  mc/cmds-to-run-once '(upcase-region))
```

layers/+misc/multiple-cursors/README.org
bb2020: I think links are enough.

layers/+misc/multiple-cursors/README.org
bb2020: Done

layers/+misc/multiple-cursors/README.org
bb2020: Done

layers/+misc/multiple-cursors/packages.el
bb2020: `:defer t` should remain I suppose. I don't know why `evil-mc` does not have it. Probably it is initialized with other evil stuff and deferring it would have no effect.

`use-package` documentation says:
`If you aren't using :commands, :bind, :bind*, :bind-keymap, :bind-keymap*, :mode, :interpreter, or :hook (all of which imply :defer; see the docstring for use-package for a brief description of each), you can still defer loading with the :defer keyword ...`

`mc` is auto-loaded when it is first invoked with key binding.

layers/+misc/multiple-cursors/README.org
bb2020: Fixed the indentation.

layers/+misc/multiple-cursors/packages.el
bb2020: This is a problem. They are overwritten by `source-control` layer. Actually my advice is to bind `mc` under `SPC x m`. `SPC g` is mostly used by source control layers. However if you say that they should be under `SPC g r` (as evil-mc resides here), I will update the bindings.

layers/+misc/multiple-cursors/packages.el
duianto: Got it, thanks for the info.

layers/+misc/multiple-cursors/packages.el
duianto: The `evil-mc` bindings are under `gr` without the `SPC` prefix. But since they are evil editing style specific they aren't an option for emacs editing style users.

They should not be under `SPC g r`, since as you say `SPC g` is for `git/versions-control` keys.

Sorry for not being clear, I meant that:
` (spacemacs/set-leader-keys "grr" 'mc/mark-more-like-this-extended)`
binds them to `SPC g r r`, which we shouldn't do.

`iedit` is bound to `SPC s e` and `SPC s m` seems to be available, would be better to try and group them together. Or is there a benefit to have them under `SPC x m`?

bb2020: The thing is `SPC x` is about text manipulation and so does `mc`. `SPC s` is the searching prefix. However since `iedit` is under `SPC s`, this makes `SPC m` a good candidate.

layers/+misc/multiple-cursors/README.org
duianto: Yeah. If `dotspacemacs-configuration-layers` in the code blocks it's going to be changed, then it should be changed in separate PR that changes it everywhere.

layers/+misc/multiple-cursors/packages.el
duianto: `SPC s` is called `search/symbol`, so maybe `iedit` is there because multiple cursors usually change multiple functions/symbols/variables/etc.

Where do you see that `SPC x m` is called `move`?

bb2020: It seems is no longer a prefix. It used to be. https://github.com/syl20bnr/spacemacs/blob/c7a103a772d808101d7635ec10f292ab9202d9ee/layers/%2Bdistributions/spacemacs-base/keybindings.el#L64

bb2020:
`mc (default)` seems apparent. I've added a description line.

layers/+misc/multiple-cursors/packages.el
duianto: It seems to have been removed from the develop branch.
https://github.com/syl20bnr/spacemacs/blob/b4907d0540bae38588ed50fad063883c76c41867/layers/%2Bspacemacs/spacemacs-defaults/keybindings.el#L81-L88

bb2020: I've changed the bindings for now. Someone may want to test it.

layers/+misc/multiple-cursors/README.org
```diff
@@ -60,3 +64,44 @@ For easy navigation you also have the following:
...
 | ~C-M-j~     | evil-mc-make-cursor-move-next-line |
 | ~C-M-k~     | evil-mc-make-cursor-move-prev-line |
+
+** =multiple-cursors=
+The =multiple-cursors= backend provides the following key bindings to
+insert new cursors:
+
+| Key binding | Description                     |
+|-------------+---------------------------------|
+| ~s m m~     | mc/mark-more-like-this-extended |
```
duianto: The convention is to include `SPC` in key bindings:
`SPC s m m`
etc...

>To keep things readable only mention the prefix SPC when documenting key bindings, you don’t need to mention M-m.

https://github.com/syl20bnr/spacemacs/blob/c7a103a772d808101d7635ec10f292ab9202d9ee/doc/CONVENTIONS.org#writing-documentation

duianto: Since you've added `SPC` I say leave it, then it at least matches the `c-c++` layer.

bb2020: Done. Since I use Emacs editing style, my brain automatically casts `g r` to `SPC g r` 😄

layers/+misc/multiple-cursors/packages.el
```diff
@@ -34,3 +34,27 @@
...
       (add-hook 'text-mode-hook 'turn-on-evil-mc-mode))))
+
+(defun multiple-cursors/init-multiple-cursors ()
+  (use-package multiple-cursors
+    :defer t
+    :init
+    (progn
+      (spacemacs/declare-prefix "sm" "multiple-cursors")
+      (spacemacs/declare-prefix "sms" "specials")
+      (spacemacs/set-leader-keys "smm" 'mc/mark-more-like-this-extended)
```
duianto: The key bindings can be bound under a single `(spacemacs/set-leader-keys`:

https://github.com/syl20bnr/spacemacs/blob/b4907d0540bae38588ed50fad063883c76c41867/layers/%2Bchat/erc/packages.el#L50-L54

duianto:
I did a quick test and it seems to be working with the emacs editing style.

But there are issues with the evil editing style.
`mc` doesn't seem to have evil support, so it's understandable that there are issues.

The cursors get out of alignment just by exiting insert state.

And I managed to get into a state where char `v` and line `V` selections stopped working, in both evil and emacs state `C-z`, even when multiple cursors was disabled. (I don't remember the reproduction steps :/)

The issue was just limited to that buffer. It was an unsaved buffer `SPC b N n`, so I block selected `C-v` the buffer, copied it's contents, deleted it and pasted the contents in a new buffer and the selection worked again. If it was a file then it might have been enough to just close and reopen the file,

Is there a benefit to binding `mc/mark-all-like-this` to `SPC s m a`?
It requires a region.

`mc/mark-all-like-this-dwim` uses the region if it's active, otherwise it uses the symbol.
It has several behaviors.
>  "Tries to guess what you want to mark all of.
Can be pressed multiple times to increase selection.
With prefix, it behaves the same as original `mc/mark-all-like-this'"

https://github.com/magnars/multiple-cursors.el/blob/6a7c3c0853e3fe9e4b8e5985dbed8fd4075f33ff/mc-mark-more.el#L541

There's also a `mc/mark-all-dwim` function.
>  "Tries even harder to guess what you want to mark all of.
If the region is active and spans multiple lines, it will behave
as if `mc/mark-all-in-region'. With the prefix ARG, it will call
`mc/edit-lines' instead.
If the region is inactive or on a single line, it will behave like
`mc/mark-all-like-this-dwim'."

https://github.com/magnars/multiple-cursors.el/blob/6a7c3c0853e3fe9e4b8e5985dbed8fd4075f33ff/mc-mark-more.el#L570

I don't know which is better, but the current one might be a bit limiting.

bb2020:
Hmm, `mc/mark-all-dwim` seems like a wiser choice since it replaces `mc/mark-all-in-region`, `mc/mark-all-like-this-dwim` and `mc/mark-all-like-this-in-defun`.

`mc/mark-all-like-this-dwim` prints an error message sometimes if there is not an active region marked. For example, pressing `SPC s m a` on an open parenthesis (elisp), prints:
`mc/mark-all-like-this-dwim: Wrong type argument: integer-or-marker-p, nil`
Nevertheless I've tested it now and looks very convenient. This is probably by design because it looks for a symbol.

`mc/mark-all-like-this` should stay because this is the only buffer-wise variation. I've now bound it to `SPC s m A`. Otherwise multiple presses to `SPC s m a` is required to work on entire buffer which is daunting.

duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/d68c6311d66ca036a3efdf42c632d6154ee39256

I added the key bindings to the changelog.develop and sorted the readme key bindings alphabetically.

PR #12373, opened 2019-05-26
Improve point handling in perltidy functions

jimpudar:
This will move the point and content of the window close to where you were
before you ran perltidy. Of course if perltidy adds or removes a significant
amount of characters / lines as part of tidying, the point will be moved by that
amount. However in practice this I've found this to be close enough.


smile13241324:
Thank you for contributing to Spacemacs :purple_heart:, I have cherry picked your commit into develop. You can safely delete your branch.

jimpudar:
@smile13241324 Thanks! 🎉

PR #12383, opened 2019-05-28
Honor custom major-mode in spacemacs/new-empty-buffer

### LINKED ISSUE #12382: Spacemacs ignores custom major-mode for new buffers
#### Description :octocat:
Spacemacs ignores custom major-mode for new buffers

- Start Emacs
- `M-m h d v major-mode RET`
- `RET` on the "customize" link to enter Customize, to change the default
- change the value to "text-mode" (without quotes)
- `RET` on the State button, choose option `1` (set and save for future sessions)

Then:

- `M-m b N n` to create a new buffer

New buffer opens in Fundamental mode, regardless of the setting.

New buffer opens in Text mode, as per the customized default major-mode.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 27.0.50
- Spacemacs: 0.300.0
- Spacemacs branch: fix-12358 (rev. 5e4c04b04)
- Graphic display: t
- Distribution: spacemacs
- Editing style: emacs
- Completion: helm
- Layers:
```elisp
(racket python helm
        (auto-completion :variables auto-completion-enable-snippets-in-popup t auto-completion-return-key-behavior 'complete auto-completion-tab-key-behavior 'cycle auto-completion-complete-with-key-sequence nil auto-completion-complete-with-key-sequence-delay 0.1 auto-completion-private-snippets-directory nil)
        emacs-lisp multiple-cursors treemacs git latex markdown
        (shell :variables shell-default-width 40 shell-default-position 'right)
        spell-checking syntax-checking themes-megapack pdf)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG SOUND GPM DBUS GSETTINGS GLIB NOTIFY INOTIFY ACL LIBSELINUX GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XDBE XIM MODULES THREADS XWIDGETS LIBSYSTEMD PDUMPER LCMS2 GMP

*Notes*

I noticed `M-x switch-to-buffer` (by default bound to `C-x b`), then entering a name for the new buffer (as suggested [here](https://stackoverflow.com/questions/25791605/emacs-how-do-i-create-a-new-empty-buffer-whenever-creating-a-new-frame) for plain Emacs), indeed honors the custom major-mode setting, but `generate-new-buffer` does not. Spacemacs uses the latter in `M-m b N n`, as can be seen from function `spacemacs/new-empty-buffer` in file `.emacs.d/layers/+spacemacs/spacemacs-defaults/funcs.el`.

Note also that at least in Emacs 27.0.50, major-mode becomes buffer-local if set outside Customize. To set the default value, Customize must be used. (This note comes from `M-m h d v major-mode RET`.)

Technologicat:
Here's one possibility how to make Spacemacs honor a Customized default for major-mode when creating a new buffer with `SPC b N n`.

No idea if this is the best or correct approach, but `default-value` should at least retrieve the global value (according to [this](https://stackoverflow.com/questions/3806423/how-can-i-get-a-variables-initial-value-in-elisp)), and funcalling (the function cell of) that symbol obviously runs the mode-setting code. By my (manual) testing it seems the mode takes effect only in the new buffer, as desired.

Fixes #12382.

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -647,9 +647,10 @@ variable."
 
 (defun spacemacs/new-empty-buffer (&optional split)
   "Create a new buffer called untitled(<n>).
-A SPLIT argument with the value: `left',
-`below', `above' or `right', opens the new
-buffer in a split window."
+A SPLIT argument with the value: `left', `below', `above' or `right',
+opens the new buffer in a split window.
+If the variable `dotspacemacs-new-empty-buffer-major-mode' has been set,
+then apply that major mode to the new buffer."
   (interactive)
   (let ((newbuf (generate-new-buffer "untitled")))
     (case split
@@ -660,7 +661,9 @@ buffer in a split window."
       ('frame (select-frame (make-frame))))
     ;; Prompt to save on `save-some-buffers' with positive PRED
     (with-current-buffer newbuf
-      (setq-local buffer-offer-save t))
+      (setq-local buffer-offer-save t)
+      (when dotspacemacs-new-empty-buffer-major-mode
+        (funcall dotspacemacs-new-empty-buffer-major-mode)))
     ;; pass non-nil force-same-window to prevent `switch-to-buffer' from
     ;; displaying buffer in another window
     (switch-to-buffer newbuf nil 'force-same-window)))
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -477,6 +477,8 @@ Other:
   - Added line text object using =evil-textobj-line= (thanks to Uroš Perišić)
   - Added more =kaolin-themes= (thanks to ogdenwebb)
   - Calling ~spacemacs/recompile-elpa~ with an argument nukes all *.elc files (thanks to Ag Ibragimov)
+  - Allow customizing default major mode for new empty buffers in ~SPC b N n~,
+    see =dotspacemacs-new-empty-buffer-major-mode= (thanks to Juha Jeronen)
 - Fixed:
   - Avoid non-idempotent use of push in init code (thanks to Miciah Masters)
   - Moved Spacemacs startup progress bar to =core-progress-bar.el=, removed
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -414,6 +414,9 @@ List sizes may be nil, in which case
   "Run `spacemacs/prettify-org-buffer' when
 visiting README.org files of Spacemacs.")
 
+(defvar dotspacemacs-new-empty-buffer-major-mode nil
+  "Set the major mode for a new empty buffer.")
+
 (defun dotspacemacs//prettify-spacemacs-docs ()
   "Run `spacemacs/prettify-org-buffer' if `buffer-file-name'
 has `spacemacs-start-directory'"
--- a/core/templates/.spacemacs.template
+++ b/core/templates/.spacemacs.template
@@ -176,6 +176,11 @@ It should only modify the values of Spacemacs settings."
    ;; True if the home buffer should respond to resize events. (default t)
    dotspacemacs-startup-buffer-responsive t
 
+   ;; Default major mode for a new empty buffer. Possible values are mode
+   ;; names such as `text-mode'; and `nil' to use Fundamental mode.
+   ;; (default `text-mode')
+   dotspacemacs-new-empty-buffer-major-mode 'text-mode
+
    ;; Default major mode of the scratch buffer (default `text-mode')
    dotspacemacs-scratch-mode 'text-mode
 


layers/+spacemacs/spacemacs-defaults/funcs.el
```diff
@@ -660,7 +660,9 @@ buffer in a split window."
       ('frame (select-frame (make-frame))))
     ;; Prompt to save on `save-some-buffers' with positive PRED
     (with-current-buffer newbuf
-      (setq-local buffer-offer-save t))
+      (progn
```
duianto: The `progn` isn't needed here, `with-current-buffer` runs all functions in it's body.

One place where a `progn` is needed is in an if statement:
```elisp
(if this-is-non-nil
    (run-this)
  (otherwise-run-this))
```
that works as expected, if `this-is-non-nil` is non-nil, then `(run-this)` runs, otherwise `(otherwise-run-this)` runs.
```elisp
(if this-is-non-nil
    (run-this)
  (and-run-this)
  (otherwise-run-this))
```
Here only `(run-this)` runs if `this-is-non-nil` is non-nil.

The `if` functions docstring says:
>THEN must be one expression, but ELSE... can be zero or more expressions.

Therefore if `this-is-non-nil` is `nil` then both `(and-run-this)` and `(otherwise-run-this)` runs.

To run the first two commands when `this-is-non-nil` is non-nil, we need a `progn`
```elisp
(if this-is-non-nil
    (progn
      (run-this)
      (and-run-this))
  (otherwise-run-this))
```

layers/+spacemacs/spacemacs-defaults/funcs.el
```diff
@@ -660,7 +660,9 @@ buffer in a split window."
...
+        (setq-local buffer-offer-save t)
+        (funcall (default-value 'major-mode))))
```
duianto: Instead of having to set the default value of the `major-mode` in `customize`.

It's probably better to define a separate variable for the new-empty-buffer function.

Above the function definition: 
https://github.com/syl20bnr/spacemacs/blob/46100767091a73f5744cc45fc2cacb56778329a9/layers/%2Bspacemacs/spacemacs-defaults/funcs.el#L648

Add a variable definition:
```elisp
(defvar spacemacs-new-empty-buffer-major-mode nil
  "Set the major mode for a new empty buffer.")
```

Then replace:
```elisp
        (funcall (default-value 'major-mode))))
```
with:
```elisp
      (when spacemacs-new-empty-buffer-major-mode
        (funcall spacemacs-new-empty-buffer-major-mode)))
```

And update the `spacemacs/new-empty-buffer` functions docstring to:
```text
  "Create a new buffer called untitled(<n>).
A SPLIT argument with the value: `left', `below', `above' or `right',
opens the new buffer in a split window.
If the variable `spacemacs-new-empty-buffer-major-mode' has been set,
then apply that major mode to the new buffer."
```

Now the major mode can be set from the `dotspacemacs/user-config` section of `.spacemacs`, with:
```elisp
    (setq spacemacs-new-empty-buffer-major-mode 'text-mode)
```

layers/+spacemacs/spacemacs-defaults/funcs.el
Technologicat: Ok, it seems I missed the `&rest` in the documentation of `with-current-buffer`... `<f1> f with-current-buffer RET` said `body`, not `body...`, so at a quick glance I thought it needs a `progn`. :)

layers/+spacemacs/spacemacs-defaults/funcs.el
Technologicat: The reason I thought to use `major-mode` is that the it would use the same mechanism as standard Emacs (without Spacemacs), but if a variable for a user-chosen function is preferred, I'll change it.

I'll update the PR.

Technologicat:
PR updated.

Needed a rebase and force-push, followed the instructions.

Compro-Prasad:
Update `CHANGELOG.develop`. The CI is failing because of that.

Technologicat:
Changelog updated.

(It was listed as optional, so I didn't originally bother, but if it's recommended, I'll start doing that.)

layers/+spacemacs/spacemacs-defaults/funcs.el
```diff
@@ -645,11 +645,13 @@ variable."
   (ediff-files (dotspacemacs/location)
                (concat dotspacemacs-template-directory ".spacemacs.template")))
 
+(defvar spacemacs-new-empty-buffer-major-mode nil "Set the major mode for a new empty buffer.")
```
duianto: Formatting suggestion:
The other variables in the same file, shows the docstring on the line below the variable definition. This also keeps the line under 80 chars. And they leave an empty line between the variable and the following function.

For example:
https://github.com/syl20bnr/spacemacs/blob/ad5dfc1ad5c3d08266141487397a83810fa7fa7f/layers/%2Bspacemacs/spacemacs-defaults/funcs.el#L910-L913

Technologicat: Formatting suggestion applied.

Technologicat:
@smile13241324 , @duianto : Having the variable in the dotfile would ease discoverability, and being able to easily choose the default mode of new buffers feels at least to me an important convenience customization, so this would be nice to have, if you feel that's ok.

But if we want to do that, it seems to me that for consistency, the variable should be renamed as `dotspacemacs-new-empty-buffer-major-mode` (note the **dot** prefix), and defvar'd in `~/.emacs.d/core/core-dotspacemacs.el` (not directly in `~/.emacs.d/layers/+spacemacs/spacemacs-defaults/funcs.el`), to follow the conventions of the existing variables in `.spacemacs.template`.

What should we do? Shall I change this?

Technologicat:
Meant also to say, thanks for reviewing, and once we have a decision, I'll update the PR and squash the commits.

duianto:
Nice catch.
Yes, when adding a variable to `.spacemacs.template` then it should match the other variable names and their definition location.

Technologicat:
PR updated. Anything else before I squash the commits and submit the final version?

 - new var `dotspacemacs-new-empty-buffer-major-mode`
 - define it also in `.spacemacs.template`

smile13241324:
@Technologicat looks good. After the commits are squashed this should be merged. 
@duianto do you see something else?

duianto:
When the commits are squashed then it's good to go!

With `text-mode` as the default, now the new buffer also has `evil-mc-mode` is enabled by default 👍

Technologicat:
I... see? What is... `M-m h d a evil mc mode RET`... aha!

Commits squashed for great justice.

(For any other newcomers to magit following along at home, [this short guide](http://www.howardism.org/Technical/Emacs/magit-squashing.html) was useful.)

duianto:
`evil-mc` is part of the `multiple-cursors` layer: https://github.com/syl20bnr/spacemacs/tree/8c2eb8a06ee736d65b8a516d0ad90e838ac73af9/layers/%2Bmisc/multiple-cursors

Here's the repository for the package:
https://github.com/gabesoft/evil-mc

smile13241324:
Thank you for contributing to spacemacs :purple_heart:. 
I have cherry-picked your commit into develop. You can safely remove your branch.

Technologicat:
Thanks! Branch deleted.

Yeah, I knew about the `multiple-cursors` layer, but hadn't configured it yet. `evil-mc` didn't ring a bell, though I should have guessed. :)

The layer seems rather useful, especially since it supports also the `mc` backend for those of us that prefer the `emacs` editing mode. Thanks for the preconfigured keybindings!

PR #12388, opened 2019-05-29
[dash] Fix dash//activate-package-docsets when using counsel-dash

madand:
When counsel-dash is installed, but helm-dash isn't, startup error
(void-function dash-docs-installed-docsets) will be triggered.

Since common functionality of both counsel-dash and helm-dash was refactored
into dash-docs, a standalone package, the layer code was updated to use the
variables and functions from that package. This also fixed the mentioned error.


duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/a281e6979f65e31607500cdc3a54ad50ebf2568e

Issue #12392, opened 2019-05-30
Bibtex layer (org-ref): Emacs freezes when loading an Org file containing :ignore: in a heading


aschaap:
#### Description :octocat:
Spacemacs (`develop` branch) with the `org` and `bibtex` layers (the latter for `org-ref`) causes Emacs to freeze when loading an Org file containing:
```org
    * Title page                                                         :ignore:
```
This is fixed by removing the contents of the `bibtex/pre-init-org-ref` function in the `bibtex` layer's `packages.el`, though I am unsure what other effects this change has because I am a fairly new Emacs and Spacemacs user without much knowledge of its internals or Elisp in general.

#### Reproduction guide :beetle:
- Ensure you are on Spacemacs' `develop` branch
- Ensure the `org` and `bibtex` layers are enabled in `.spacemacs` (I also have `ocaml`, `git`, `helm` and `spell-checking`, but these seem irrelevant)
- Open an Org file with Emacs that contains a top-level heading ending in `:ignore:`

*Observed behaviour:* :eyes: :broken_heart:
Emacs appears to freeze; sometimes one can move around a few seconds before it does so. It maxes out a single core (100% in `top`) until a `killall emacs25` stops it.

*Expected behaviour:* :heart: :smile:
Any Org file loads quickly and the user can subsequently edit it.

#### System Info :computer:
- OS: Ubuntu 16.04 LTS
- Emacs: 25.3.2
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 39ce04c)
- Graphic display: t
- Distribution: spacemacs
- Editing style: emacs
- Completion: helm
- Layers:
```elisp
(ocaml bibtex git helm org spell-checking)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GCONF GSETTINGS NOTIFY LIBSELINUX GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 MODULES

#### Backtrace :paw_prints:
Sorry, I read https://github.com/syl20bnr/spacemacs/wiki/Debugging but I'm not sure how to generate a backtrace with Emacs when it freezes upon opening a file.

duianto:
I'm unable to reproduce a freeze in either Ubuntu or Windows.

On the latest develop branch with updated packages.

[...]

PR #12393, opened 2019-05-30
Removing line causing Emacs to freeze on Org files with :ignore:; fix…

aschaap:
…es #12392


duianto:
We'll close this. Since it seems to be related to the Emacs version:
>Turns out it is an Emacs version problem. Just tested it with Emacs 26.1 on Gentoo and it seems perfectly fine.

source: https://github.com/syl20bnr/spacemacs/issues/12392#issuecomment-497673561

If this becomes relevant again, then it can be reopened.

PR #12409, opened 2019-06-05
Fix mismatched brackets in markdown README

liuchong:
Added a missing close bracket and removed the useless line `dotspacemacs-configuration-layers '(`

### DIFF
--- a/layers/+lang/markdown/README.org
+++ b/layers/+lang/markdown/README.org
@@ -56,8 +56,7 @@ To use =vmd= (Github-flavored live preview) instead set the value of the
 variable =markdown-live-preview-engine= to =vmd=:
 
 #+BEGIN_SRC emacs-lisp
-  dotspacemacs-configuration-layers '(
-    (markdown :variables markdown-live-preview-engine 'vmd))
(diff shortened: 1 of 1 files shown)


duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/803d4e720227251e73f687462d54047f849a3f9f

Your right if `  dotspacemacs-configuration-layers '(` is removed then there's a missing closing parenthesis, otherwise there are two missing parentheses.

I think all layers use `  dotspacemacs-configuration-layers '(`, or
`(setq-default dotspacemacs-configuration-layers` in the layer variable snippets. So if they are going to be removed, then all of them probably should be removed.

I've messaged the maintainers about it. So let's just fix the parentheses for now.

I forgot to change the commit message body, so it still says that it removed the configuration layers lines.

liuchong:
Just fix the parentheses is ok 😃 

I've removed the line `dotspacemacs-configuration-layers '(` because I've found think many other layers do not have it, such as `javascript`, `go` in +lang, but yes we can keep it.

PR #12414, opened 2019-06-06
Add new key bindings for indirect buffer functions

antonalekseev:
This one adds Spacemacs-flavoured bindings for [indirect buffer](https://www.gnu.org/software/emacs/manual/html_node/emacs/Indirect-Buffers.html) functions: `make-indirect-buffer`, `clone-indirect-buffer`, `clone-indirect-buffer-other-window` under `SPC b i` prefix.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -790,6 +790,10 @@ Other:
   - Added =link-hint-copy-link= to ~SPC x y~ (thanks to William Casarin)
   - Added =evil-unimpaired= navigation keys prefixed by ~[~ and ~]~ to the
     Spacemacs home buffer (thanks to Sorawee Porncharoenwase)
+  - Added key bindings for indirect buffer functions under ~SPC b c~:
+    - ~n~ calls =make-indirect-buffer=
+    - ~c~ calls =clone-indirect-buffer=
+    - ~w~ calls =clone-indirect-buffer-other-window=
 - Improvements:
   - Rewrote window layout functions for ~SPC w 1~, ~SPC w 2~, ~SPC w 3~, and
     ~SPC w 4~ (thanks to Codruț Constantin Gușoi):
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -2345,6 +2345,9 @@ Buffer manipulation commands (start with ~b~):
 |-----------------+----------------------------------------------------------------------------------------|
 | ~SPC TAB~       | switch to alternate buffer in the current window (switch back and forth)               |
 | ~SPC b b~       | switch to a buffer                                                                     |
+| ~SPC b c n~     | create an indirect buffer                                                              |
+| ~SPC b c c~     | create an indirect buffer that is clone of current buffer                              |
+| ~SPC b c w~     | create an indirect buffer that is clone of current buffer, and open it in other window |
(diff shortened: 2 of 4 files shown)


duianto:
`SPC b i` is used by the `imenu-list` layer:
https://github.com/syl20bnr/spacemacs/tree/develop/layers/+tools/imenu-list

and `SPC b I` (upper case i) is used by the `ibuffer` layer:
https://github.com/syl20bnr/spacemacs/tree/develop/layers/+emacs/ibuffer

Could `SPC b c` be an alternative?

antonalekseev:
> Could `SPC b c` be an alternative?

By all means. Sorry, I should've paid attention while grepping sources. It turns out it's rather crowded there. Just have pushed updated version.

duianto:
There's nothing to be sorry about, you suggested a key binding with a good mnemonic 👍 , it turned out that it was taken. But we might have found a good alternative.

A PR doesn't have to be thoroughly researched. It can be a starting point, then other's can join in with suggestions on how to improve it.

While testing this I noticed that `SPC b c w` which calls `clone-indirect-buffer-other-window` opens the clone in the current window, even if there's another open window.

Something in Spacemacs seems to keep it in the current window.

Without Spacemacs, for example when starting `emacs -Q` (no configurations are loaded).

Then `clone-indirect-buffer-other-window` works as expected and opens the clone in the other window (splitting a single window).

#### System Info :computer:
- OS: windows-nt
- Emacs: 26.2
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 1ec2e319c) (this PR applied on 37dcecbec9220e9359e27741640be67040bf9ab3)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(autohotkey emacs-lisp git helm javascript markdown multiple-cursors org spell-checking syntax-checking treemacs version-control)
```

antonalekseev:
> Something in Spacemacs seems to keep it in the current window.

It's `purpose-mode`. Somehow related issue is https://github.com/bmag/emacs-purpose/issues/122, but there authors of `purpose-mode` insist that it's bug in `org-tree-to-indirect-buffer`, while `clone-indirect-buffer-other-window` unequivocally sets `pop-up-windows` to `t` which IMO should be respected.

Following code sorts out issue
```
(purpose-add-user-purposes :regexps '(("<[0-9]>$" . indirect)))
(purpose-compile-user-configuration)
```
but it's user configuration we should better not interfere with.

Therefore our options are: report the issue upstream and wait for resolution, or inject aforementioned configuration in `purpose--default-regexp-purposes` during `spacemacs-purpose` initialisation, as "suggested" in `purpose-mode` [docs](https://github.com/bmag/emacs-purpose/wiki/Purpose-Configuration#default-settings):
> The default settings are hard-coded, and are not meant to be changed (though you could if you try hard).

antonalekseev:
> Following code sorts out the issue

Unfortunately, it's not the case. This way `clone-indirect-buffer` also opens new buffer in other window.

antonalekseev:
Just've come up with solution. The only thing of my concern is whether I've chosen the right spot to place `defalias`.

duianto:

If someone with more knowledge than us discovers our "horrible mistake" 😆, Then it can be moved to the correct place. All we know is that this solves the issue. I'm sure it's fine since it's `window-purpose` related.

PR #12423, opened 2019-06-11

emacspace:


AmaiKinono:
This is not done right. Typically spaces are added between CJK characters and latin characters / numbers, so the original is right. See https://github.com/sparanoid/chinese-copywriting-guidelines/blob/master/README.en-US.md#Spacing

duianto:
Thanks for posting a correction.

I thought it seemed strange that the bot joined them, I'll inform the maintainers that the bot needs to be fixed.

JAremko:
@AmaiKinono  @duianto  We're trying to match 1 to 1 the GitHub's headline links "convention" 
As you can see [current TOC links don't work](https://github.com/syl20bnr/spacemacs/tree/develop/layers/%2Bintl/chinese#configure-the-default-input-method%20%E9%85%8D%E7%BD%AE%E9%BB%98%E8%AE%A4%E4%B8%AD%E6%96%87%E8%BE%93%E5%85%A5%E6%B3%95)

@snosov1 :thinking:

duianto:
@JAremko Your right the links are currently broken.

The link that isn't changed by this PR works:
`[[#configure-chinese-conv-配置简繁转换][Configure =chinese-conv= (配置简繁转换)]]`

It has a dash between the latin and chinese characters.

Possibly because the word in the title before the chinese characters is marked as org mode verbatim (surrounding `=` signs), that might add a dash automatically.

Let's see if the toc-org author has a solution.

JAremko:
@duianto You can see exactly what link we have to generate by click at any headline (in GitHub preview) -anything else won't work.

duianto:
I misspoke, not a space added to the link, but to the title:
#### Before:
Configure the Default Input Method(配置默认中文输入法)

#### After
Configure the Default Input Method (配置默认中文输入法)

(A space before the open parenthesis)

That's how the title is written in the working toc link:
Configure chinese-conv (配置简繁转换)

The parentheses are removed in the generated link, that's why there is no space between the latin and chinese characters.

The other spaces in the titles are converted to dashes.

Would that work or am I misunderstanding something?

JAremko:
@duianto  Should work  :thinking:

duianto:
There are also two instances of regular text without a space before the open parenthesis containing chinese characters.
>The default Chinese input method is Chinese-pyim, if you are a Wubi(五笔) user, you could set the variable chinese-default-input-method to wubi:

source:
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+intl/chinese/README.org#configure-the-default-input-method%E9%85%8D%E7%BD%AE%E9%BB%98%E8%AE%A4%E4%B8%AD%E6%96%87%E8%BE%93%E5%85%A5%E6%B3%95

>The first time you use Chinese-pyim, you need to install the lexicon(词库) file.

source:
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+intl/chinese/README.org#configure-the-chinese-pyim-input-method%E9%85%8D%E7%BD%AE%E4%B8%AD%E6%96%87%E6%8B%BC%E9%9F%B3%E8%BE%93%E5%85%A5%E6%B3%95

@AmaiKinono Since only one instance has a space before the open parenthesis. 
>Configure chinese-conv (配置简繁转换)
source:
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+intl/chinese/README.org#configure-chinese-conv-%E9%85%8D%E7%BD%AE%E7%AE%80%E7%B9%81%E8%BD%AC%E6%8D%A2

Would it break another convention to add a space before all non-code open parenthesis?

duianto:
There seems to also be two different types of closing parenthesis in the chinese readme:
    - [[#enable-youdao有道-dictionary激活有道字典][Enable YouDao(有道) Dictionary(激活有道字典)]]
    - [[#set-monospaced-font-size设置等宽字体][Set monospaced font size(设置等宽字体）]]

The closing parenthesis at the end of the:
- first description looks like a "normal" parenthesis
- second description looks like it has a space after it, but it's a single character.

The second one seems to be the only one, it should probably be changed to match the other once.

snosov1:
Not sure if I need to anything, but here's what I've tried:

1. 1 out of 5 links with Chinese characters works alright (`Configure chinese-conv (配置简繁转换)`)
2. The ones that don't work have a space in the link for some reason
3. When I try it on My Machine(tm) - the link appears correct and there's no undesired spaces (maybe updating to the latest version would solve it? Maybe you're using something other than UTF-8 and it causes problems?)

Please, let me know if you think it's really a `toc-org`'s issue. At this point it doesn't seem like it.

duianto:
I don't know where the spaces come from. I didn't touch the toc links before pushing the commit:
[chinese] Fix fcitx.el. And add a variable https://github.com/syl20bnr/spacemacs/commit/5b7b5f252643a44298f6161097fd80f73c95eeb0

But the spaces seem to have been introduced there.

You mentioned `UTF-8`. I'm not quite sure what it is, but it seems to have something to do with character encoding (https://en.wikipedia.org/wiki/UTF-8).

In for example a Magit buffer, the right side of the mode line shows `utf-8`. But when the file `.emacs.d/layers/+intl/chinese/README.org` is opened, then the mode line shows `dos`.

Could that have something to do with it?

I'm using windows 10, maybe that's why it shows `dos`?

---

I tried to replicate the changes I made to the PR:
Make fcitx.el in Chinese layer work by default. #10460

By applying the PRs patch:
https://patch-diff.githubusercontent.com/raw/syl20bnr/spacemacs/pull/10460.patch
on top of the same commit:
3d4e6fc0d * documentation formatting: Sun Jun  9 19:54:33 UTC 2019
that it was applied to on the current Spacemacs `develop` branch.

But the toc links do not get a space in them.

---

The reverse happens on the current develop branch when `chinese/readme.org` is modified (for example by adding a newline anywhere in the document) and saved, then the toc link spaces are removed. (probably expected since as we've seen here, spaces don't work in links).

JAremko:
Can we just merge it :wink:

duianto:
Yes, I was just about to suggest that we merge this since it's holding up other documentation changes. When AmaiKinono responds, we can decide what the next steps are.

AmaiKinono:
> Would it break another convention to add a space before all non-code open parenthesis?

It depends. In this document it's totally fine to do this.

There are two kinds of parenthesis. You guys are probably more familiar with the half-width "()", but the full-width "（）" is typically used in CJK environment (and we have other full-width punctuations like the period "。"). The idea is to:

- have spaces between full-width and half-width characters so it doesn't look like crowded together. e.g.
 
  This is bad:
  ```
  中文(Chinese)中文
  ```

  This is good:
  ```
  中文 (Chinese) 中文
  ```

- have no extra spaces around full-width punctuation since there are enough blank in itself. e.g.

  This is bad:
  ```
  Chinese （中文） Chinese
  ```

  This is good:
  ```
  Chinese（中文）Chinese
  ```

These are conventions developed loosely by CJK users on the internet, with no standardized documentations about them (thought there are drafts like http://w3c.github.io/clreq/, and https://www.w3.org/TR/jlreq/). I would say it's fine to ignore the conventions if you are not willing to go down the rabbit hole.

> The convention about a space between latin and chinese characters is only broken in the backend links, which aren't visible to the users.

In my opinion this is totally fine since the goal of the conventions is for the text to be visually beautiful. The only visible problems are in the headings, like it should be this:

```
Configure the Default Input Method (配置默认中文输入法)
```

or

```
Configure the Default Input Method（配置默认中文输入法）
```

not this

```
Configure the Default Input Method(配置默认中文输入法)
```

And in this heading, the full-width and half-width parenthesis are mixed, which is not good:

```
*** Set monospaced font size(设置等宽字体）
```

duianto:
That's interesting, I didn't know about full-width parentheses.

But we can't use full-width parentheses here, because that's the issue, the toc links remove the parentheses from the titles.

Therefore we'll fix all of the titles by adding a space and half-width parentheses around the chinese characters. Then the toc links will replace the space with a dash, and they won't combine the latin and chinese characters.

Here's an example with the first title:

### Before:
#### Title:
```org
*** Configure the Default Input Method(配置默认中文输入法)
```
#### Toc link:
```org
    - [[#configure-the-default-input-method配置默认中文输入法][Configure the Default Input Method(配置默认中文输入法)]]
```
### After:
#### Title:
```org
*** Configure the Default Input Method (配置默认中文输入法)
```
#### Toc link:
```org
    - [[#configure-the-default-input-method-配置默认中文输入法][Configure the Default Input Method (配置默认中文输入法)]]
```

#### Should the other instances of chinese characters in parentheses also have spaces in front of them?

Three links in the `Features` section:

Before | After
-|-
Pinyin(拼音) | Pinyin (拼音)
Wubi(五笔) | Wubi (五笔)
Youdao(有道) Dictionary | Youdao (有道) Dictionary

A word in the `*** Configure the Default Input Method (配置默认中文输入法)` section:

Before | After
-|-
Wubi(五笔) | Wubi (五笔)

A word in the `*** Configure the =Chinese-pyim= Input Method (配置中文拼音输入法)` section:

Before | After
-|-
lexicon(词库) | lexicon (词库)

AmaiKinono:
> Should the other instances of chinese characters in parentheses also have spaces in front of them?

Yes. And all the examples you give is done right!

duianto:
They have now been fixed: https://github.com/syl20bnr/spacemacs/commit/ec01b45744b441d09a4c65e3b0a2b62c239a964b
https://github.com/syl20bnr/spacemacs/tree/develop/layers/%2Bintl/chinese

PR #12425, opened 2019-06-12
Fix lazy-loading of ruby-test-mode

amfranz:
This change fixes a syntax mistake in the `use-package` declaration of `ruby-test-mode` that causes the package to be loaded immediately, instead of deferring the load until the first ruby file is visited, as is intended.


Compro-Prasad:
Add an entry to `CHANGELOG.develop`.

amfranz:
> Add an entry to `CHANGELOG.develop`.

Of course. I added it into the "ruby layer" section.

duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/ddf7459a0253674cc931596235736fb5b37f394b

PR #12428, opened 2019-06-12
[Clojure] Add keybinding for cider-interrupt

jr0cket:
Interrupt longer running evaluations without having to kill or reset the REPL connection.

Calls the `cider-interrupt` function, which is not currently mapped to a keybinding.

Placed in the evaluate section, as it is specific to the currently running evaluation.

Tested by evaluating the Clojure expression:

` (range 10000000)`

Then used `, e i` to interrupt the evaluation before it completed.


duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/bb3437481827b9e2bc2328ee127388a3496a560c

PR #12431, opened 2019-06-12
[defaults] Don't use large file check in buffers already in fundamental-mode.

Alexander-Miller:
A trivial change inspired by me suddenly being asked to open elfeed's index literally every time I first started it..


Alexander-Miller:
> Some checks were not successful

First one's a connection issue:

> Warning (spacemacs): 
> Error connection time out for melpa repository!

For the second one I don't think this change is significant enough to warrant a changelog entry.

duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/2e14b4a88ce09bfe5526b883b638b4920955a9b2

sdwolfz said that there should be a changelog entry, so I added this:
>  - Excluded =fundamental-mode= from =spacemacs/check-large-file=
    (thanks to Alexander Miller)

PR #12435, opened 2019-06-13
Fix copied dir not in minibuffer, clipboard disabled

### LINKED ISSUE #12426: `spacemacs/copy-file-path` no longer prints file path to minibuffer
#### Description :octocat:

`SPC f y y` used to print the file path which was very useful for finding out where you were (as suggested in this issue https://github.com/syl20bnr/spacemacs/issues/2140).

This stopped working, and it seems that `kill-new` inside the spacemacs/copy-file-path function no longer returns the killed string and instead seems to return 0, which is what gets printed (perhaps due to an emacs version upgrade)?

I was seeing this on emacs 26.1_1 (from homebrew), so upgraded to 26.2 to see if the issue persisted and it does.

- Start Emacs
- navigate to any file
- press `SPC f y y`

The string "0" appears in the minibuffer.

The file path should appear in the minibuffer.

#### System Info :computer:
- OS: darwin
- Emacs: 26.2
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 51b04beb2)
- Graphic display: nil
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
((osx :variables osx-right-command-as 'meta)
 tmux my-evil auto-completion helm emacs-lisp rust treemacs git markdown org
 spell-checking javascript
 (node :variables node-add-modules-path t)
 shell-scripts html sql
 (elm :variables elm-format-on-save t elm-sort-imports-on-save t)
 yaml ruby
 (python :variables python-backend 'anaconda)
 ocaml ipython-notebook docker markdown terraform my-clojure my-evil-lisp my-sql my-js-json my-haskell my-python lsp reason imandra xclipboard)
```
- System configuration features: NOTIFY ACL GNUTLS LIBXML2 ZLIB MODULES THREADS

#### Backtrace :paw_prints:
```

```

duianto:
Fixed the minibuffer showing: `nil`
when `select-enable-clipboard` is `nil` and the current files
path/dir/file/line/column/etc. is copied with the keys under
the prefix: `SPC f y`

Resolves #12426

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1000,6 +1000,8 @@ Other:
   - Key bindings (thanks to bb2020)
     - ~SPC-j-C~ to =check-parens=
     - ~SPC-,~ to =goto-last-change=
+  - Fixed copied dir/path not shown in minibuffer when =select-enable-clipboard=
+    is =nil= (thanks to duianto)
 *** Layer changes and fixes
 **** Alda
 - Key bindings:
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -583,21 +583,27 @@ variable as a fallback to display the directory, useful in buffers like the
 ones created by `magit' and `dired'."
   (interactive)
   (if-let (directory-path (spacemacs--directory-path))
-      (message "%s" (kill-new directory-path))
+      (progn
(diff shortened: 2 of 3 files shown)


smile13241324:
Thank you for contributing to spacemacs :purple_heart:, cherry picked into develop.
You can safely remove your branch.

Issue #12351, opened 2019-05-23
Is there any plan to support clangd as lsp-backend for c-c++ mode?


idears-org:
I can find some workaround method in #12084 , but I think it's better to add a feature to support lsp-clangd in c-c++-backend.

Is there any plan to do this?

auscompgeek:
I hacked up some clangd support in https://github.com/auscompgeek/spacemacs/tree/clangd. Far from perfect, but mostly gets the job done for now.

yyoncho:
@auscompgeek IMO it looks ok and you could open a PR.

idears-org:
@auscompgeek Awesome job, waiting for your pull request to be merged.

yyoncho:
this is fixed

PR #12443, opened 2019-06-15
c-c++: Add clangd support

auscompgeek:
Closes: #12351

### DIFF
--- a/layers/+lang/c-c++/config.el
+++ b/layers/+lang/c-c++/config.el
@@ -20,7 +20,7 @@
 (defconst c-c++-mode-hooks '(c-mode-hook c++-mode-hook)
   "Primary hooks of the `c-c++' layer.")
 
-(defconst c-c++-lsp-backends '(lsp-cquery lsp-ccls)
+(defconst c-c++-lsp-backends '(lsp-cquery lsp-ccls lsp-clangd)
   "Language Server Protocol (LSP) backends supported by the `c-c++' layer.")
 
 (defvar c-c++-backend nil
--- a/layers/+lang/c-c++/funcs.el
+++ b/layers/+lang/c-c++/funcs.el
@@ -213,7 +213,8 @@ is non-nil."
   "Return a string representation of the LSP backend specified by the `c-c++-backend' configuration variable, without the `lsp-' prefix."
   (ecase c-c++-backend
     ('lsp-ccls "ccls")
-    ('lsp-cquery "cquery")))
+    ('lsp-cquery "cquery")
+    ('lsp-clangd "clangd")))
 
 (defun spacemacs//c-c++-lsp-string (prefix suffix)
   (concat prefix (spacemacs//c-c++-lsp-backend) suffix))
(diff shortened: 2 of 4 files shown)


kvaneesh:
One issue I have found with a similar approach is handling the lsp-extensions. Some of the command within ` SPC m g` gives me the error like:

command-execute: Wrong type argument: commandp, c-c++/find-refs-write
command-execute: Wrong type argument: commandp, c-c++/find-callers

auscompgeek:
@kvaneesh Yup, this will also need to not bind the keys for the cquery/ccls LSP extensions.

cormacc:
@auscompgeek Finally got around to doing this myself yesterday - didn't realise you'd most of the work done already until I came to raise a PR :)  My implementation is pretty similar, but tackles the keybinding issue. Also separates the lsp-server selection and backend selection (i.e. you now set `c-c++-backend` to `nil`, `rtags` or `lsp`, then set `c-c++-lsp-server` to `clangd`, `ccls` or `cquery` -- `clangd` default). 

Want to pool our efforts or something? My branch is here:
https://github.com/cormacc/spacemacs/tree/c-lsp-clangd

N.B. it branched from another open PR (minor update to lsp layer #12695 )

auscompgeek:
> Also separates the lsp-server selection and backend selection (i.e. you now set c-c++-backend to nil, rtags or lsp, then set c-c++-lsp-server to clangd, ccls or cquery -- clangd default).

I'd be wary of possibly breaking people's existing configs here.

It'd definitely be good to consolidate our efforts. Your changes to packages.el look much better than mine. We still need to define a keybinding for clangd's switch source header extension though, which looks to be missing from yours altogether.

cormacc:
Re. clangd switch source/header -- I hadn't seen the clangd extensions page -- good call. Some of those other extensions look nice too, though probably a bit more work in putting them to use. Do you reckon the clangd switch should replace or complement the existing goto-alternate-file binding?

Re. breaking existing configs -- I'm afraid I'm guilty of creating the current `c-c++-backend` situation. I adopted a similar approach in a later PR against the python layer and @sdwolfz called it an anti-pattern -- on reflection I was inclined to agree with him and changed it to something like what I've introduced in my current branch of the `c-c++` layer.  Considering none of the lsp additions to c/c++ have been part of a spacemacs release yet, it may be acceptable? Perhaps one of the maintainers might weigh in...

cormacc:
@auscompgeek Did you manage to use that textDocument/switchSourceHeader binding? I'm having difficulties with it (and figuring out if/what parameters it needs). Pushed some changes to my branch (which was pretty profoundly broken, but now seems less so) if you fancy taking a look.

bhipple:
FWIW, one of the best things about spacemacs is that it's reasonably opinionated with a sane/good set of defaults for everything that mostly just works.  The C/C++ layer is a bit of an exception, in that it seems to try to support just about everything, and (at least last time I've played with it) ends up being a little confusing and gives the impression that it doesn't work particularly well.  I see a lot of commentary on various isssues, r/emacs, and other communities trying to figure out what the "best" way to setup Spacemacs for C++ is, so if you think one set of tools has momentum, such as the clangd lsp backend, I think most users would be happy to absorb some develop branch breakage in order to get a more polished / opinionated experience.

cormacc:
I've fixed some additional clangd configuration options that weren't getting passed through (based on feedback from @kvaneesh) Think my branch is more or less there now, except that the clangd switch source/header bit, which I'm not calling correctly. To be honest, I'm perfectly happy with the generic projectile goto alternate file implementation and not inclined to spend more time on clangd-specific switch source header (particularly as I'll probably continue to use the ccls server anyway). @auscompgeek will I go ahead and raise a new PR from my branch, or would you rather proceed some other way?

auscompgeek:
Go for it - I'm not personally attached to this PR :)

cormacc:
Thanks @auscompgeek -- didn't want to tread on anyone's toes ;)

Issue #11141, opened 2018-08-07
Can we add `visual` to `dotspacemacs-line-numbers`?


yanghaoxie:
Since emacs 26 has `display-line-numbers-mode`, which support `visual` option. Can we add this option to `dotspacemacs-line-numbers`. This option play well with folding code, and a detail use case can be found here #6536.

NQNStudios:
I think I ran into the same issue with Spacemacs and Emacs 25: https://emacs.stackexchange.com/questions/34679/making-relative-line-numbers-behave-well-with-folds

duianto:
The following PR is now on the `develop` branch:
Add support for visual line numbers #12450

PR #12450, opened 2019-06-18
Add support for visual line numbers

jcaw:
Emacs 26 added built-in support for line numbers, relative line numbers, and
[visual line numbers](https://www.gnu.org/software/emacs/manual/html_node/emacs/Display-Custom.html). Spacemacs supports only absolute and relative, but there is
**no way to access the visual mode.** It's hard to get around this, since Spacemacs
abstracts line numbers. 

Arguably, `visual` is much more useful than `relative` as a display type. Visual
line numbers are like relative line numbers, but only lines that are actually
showing are counted. This means:

  1. Hidden lines are not counted. If a large amount of text is folded, the line
     numbers won't jump from "10" to "546". This is particularly useful in
     buffers like `magit-status`, where a large amount of information is folded
     by default.

  2. Lines that are wrapped are counted as multiple lines, since they're being
     displayed as multiple lines in the editor. Under `relative`, the entire 
     wrapped line is numbered once, meaning numbers skip one or more 
     lines. `visual` labels them all.

With standard relative line numbers, you can't actually navigate using the line
numbers in the sidebar as soon as folded or wrapped lines are introduced, 
since the numbers don't correspond to movements. The actual number of 
lines between your cursor and the target position is different to the number 
displayed. Since this is one of the main use cases for relative line numbers, 
it undermines the mode.

`visual` fixes that problem. Every line that's being displayed is labelled.
Numbers always correspond to the actual number of lines you'd need to 
navigate to reach the target.

This pull request extends Spacemacs' line number interface to provide visual 
line number support.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -875,6 +875,7 @@ Other:
     - ~m~ for toggle maximize
     - ~f~ for toggle fullscreen
     (thanks to Ag Ibragimov)
+  - Added support for visual line numbers (thanks to jcaw)
 - Fixed:
   - Fixed ~h~ key binding in compilation and grep buffers
     (thanks to Sylvain Benner)
--- a/core/templates/.spacemacs.template
+++ b/core/templates/.spacemacs.template
@@ -340,10 +340,14 @@ It should only modify the values of Spacemacs settings."
    dotspacemacs-smooth-scrolling t
 
    ;; Control line numbers activation.
-   ;; If set to `t' or `relative' line numbers are turned on in all `prog-mode' and
-   ;; `text-mode' derivatives. If set to `relative', line numbers are relative.
+   ;; If set to `t', `relative' or `visual' then line numbers are enabled in all
+   ;; `prog-mode' and `text-mode' derivatives. If set to `relative', line
+   ;; numbers are relative. If set to `visual', line numbers are also relative,
+   ;; but lines are only visual lines are counted. For example, folded lines
+   ;; will not be counted and wrapped lines are counted as multiple lines.
    ;; This variable can also be set to a property list for finer control:
    ;; '(:relative nil
+   ;;   :visual nil
    ;;   :disabled-for-modes dired-mode
    ;;                       doc-view-mode
    ;;                       markdown-mode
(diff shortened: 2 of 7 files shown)


jcaw:
Closes #11141

Compro-Prasad:
I think you should also change the spacemacs template.

duianto:
This seems to be working fine. 👍 

### Update request:
#### .spacemacs.template comment
The comment for the `dotspacemacs-line-numbers` variable in the `.spacemacs.template` needs to be updated to mention `visual`.

https://github.com/syl20bnr/spacemacs/blob/f644c59ab84370b14832810e184df04824ff6406/core/templates/.spacemacs.template#L347

Here's a possible rewrite, but any improvements are always welcome.
#### Before
```elisp
   ;; Control line numbers activation.
   ;; If set to `t' or `relative' line numbers are turned on in all `prog-mode' and
   ;; `text-mode' derivatives. If set to `relative', line numbers are relative.
   ;; This variable can also be set to a property list for finer control:
   ;; '(:relative nil
   ;;   :disabled-for-modes dired-mode
   ;;                       doc-view-mode
   ;;                       markdown-mode
   ;;                       org-mode
   ;;                       pdf-view-mode
   ;;                       text-mode
   ;;   :size-limit-kb 1000)
   ;; (default nil)
   dotspacemacs-line-numbers nil
```
#### After

```elisp
   ;; Control line numbers activation.
   ;; If set to `t', `relative' or `visual' then line numbers are enabled in all
   ;; `prog-mode' and `text-mode' derivatives. If set to `relative', then line
   ;; numbers are relative. If set to `visual', then line numbers are also
   ;; relative, but lines are separated by visual state - a wrapped line, for
   ;; example, will be treated as more than one line. This option takes
   ;; precedence over `relative'.
   ;; This variable can also be set to a property list for finer control:
   ;; '(:relative nil
   ;;   :visual nil
   ;;   :disabled-for-modes dired-mode
   ;;                       doc-view-mode
   ;;                       markdown-mode
   ;;                       org-mode
   ;;                       pdf-view-mode
   ;;                       text-mode
   ;;   :size-limit-kb 1000)
   ;; (default nil)
   dotspacemacs-line-numbers nil
```

#### changelog.develop
Add an entry to:
In the `Distribution changes` section,
at the end of the `Improvements:` sub section (the line above the `Fixed:` sub section header)
```
  - Added support for visual line numbers (thanks to jcaw)
```

jcaw:
The suggestion seem good. I will implement them today.

jcaw:
Force-pushed the updates. Everything is in one commit.

duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/d4cca74854f0bde9d725146ae358fb101c604c2f

PR #12452, opened 2019-06-19
[treemacs] Load projectile key binding with treemacs

duianto:
The `C-c C-p p` (`treemacs-projectile`) key binding wasn't loading automatically
with Treemacs, until the command had been called manually.


smile13241324:
Thank you for contributing to Spacemacs :purple_heart:, I have cherry-picked your commit to develop. You can safely remove your branch.

PR #12463, opened 2019-06-23
Emacs Info layer

agzam:
# Emacs info layer

### DIFF
--- a/layers/+emacs/emacs-info/README.org
+++ b/layers/+emacs/emacs-info/README.org
@@ -1,0 +1,10 @@
+#+TITLE: Emasc Info Imrprovements
+
+#+TAGS: emacs|layer
+
+* Description
+  This layer improves navigation for Emacs Info mode
+** Features:
+   - TODO: list features
+* Key bindings
+  - TODO: list keys
--- a/layers/+emacs/emacs-info/packages.el
+++ b/layers/+emacs/emacs-info/packages.el
@@ -1,0 +1,59 @@
+;;; packages.el --- Org Layer packages File for Spacemacs
+;;
+;; Copyright (c) 2012-2018 Sylvain Benner & Contributors
+;;
+;; Author: Sylvain Benner <sylvain.benner@gmail.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+(defconst emacs-info-packages
+  '((info :location built-in)))
+
+(defun emacs-info/init-info ()
+  (use-package info
+    :defer t
+    :init
+    (progn
+      (spacemacs|define-transient-state emacs-info
+        :title "Emacs Info transient state"
+        :foreign-keys run
+        :doc
+"
+ [_q_] quit "
+        :bindings
+        ("n" Info-next)
+        ("p" Info-prev)
+        ("e" Info-scroll-up)
+        ("y" Info-scroll-down)
+        ("<tab>" Info-next-reference)
+        ("S-<tab>" Info-prev-reference)
+        ("h" Info-help)
+        ("l" Info-history-back)
+        ("r" Info-history-forward)
+        ("d" Info-directory)
+        ("u" Info-up)
+        ("L" Info-history)
+        ("s" Info-search)
+        ("S" Info-search-case-sensitively)
+        ("i" Info-index)
+        ("I" Info-virtual-index)
+        ("a" info-apropos)
+        ("m" Info-menu)
+        ("gd" Info-goto-node)
+        ("gm" Info-menu)
+        ("gt" Info-top-node)
+        ("gT" Info-toc)
+        ("gf" Info-follow-reference)
+        ("gj" Info-next)
+        ("gk" Info-prev)
+
+        ("q" nil :exit t)
+        :config
+        (evilified-state-evilify-map Info-mode-map
+          :mode Info-mode
+          :bindings
+          (kbd "M-SPC" 'spacemacs/emacs-info-transient-state/body)))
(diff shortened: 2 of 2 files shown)


agzam:
Hey friends, I think we need better navigation for Emacs Info. I have started this, but I need help with categorizing keys and adding them - most importantly I don't want to make wrong assumptions of what the good key for any given function would be. Any thoughts?

duianto:
The evil collection has an info file:
https://github.com/emacs-evil/evil-collection/blob/master/evil-collection-info.el

It might have some useful suggestions.

layers/+emacs/emacs-info/packages.el
```diff
@@ -0,0 +1,59 @@
+;;; packages.el --- Org Layer packages File for Spacemacs
+;;
+;; Copyright (c) 2012-2018 Sylvain Benner & Contributors
+;;
+;; Author: Sylvain Benner <sylvain.benner@gmail.com>
```
smile13241324: You should be mentioning yourself as author

layers/+emacs/emacs-info/packages.el
```diff
@@ -0,0 +1,59 @@
...
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+(defconst emacs-info-packages
```
smile13241324: This should be a setq instead
```suggestion
(setq emacs-info-packages
```

layers/+emacs/emacs-info/packages.el
```diff
@@ -0,0 +1,59 @@
...
+        ("gj" Info-next)
+        ("gk" Info-prev)
+
+        ("q" nil :exit t)
+        :config
+        (evilified-state-evilify-map Info-mode-map
+          :mode Info-mode
+          :bindings
+          (kbd "M-SPC" 'spacemacs/emacs-info-transient-state/body)))
+      (define-key Info-mode-map (kbd "M-SPC") 'spacemacs/emacs-info-transient-state/body))))
```
smile13241324: You should also bind your transient state to `s-M-SPC` for all where `M-SPC` is bound by the OS.
```elisp
      (define-key helm-map (kbd "M-SPC")
        'spacemacs/helm-navigation-transient-state/body)
      (define-key helm-map (kbd "s-M-SPC")
        'spacemacs/helm-navigation-transient-state/body)
```

layers/+emacs/emacs-info/README.org
```diff
@@ -0,0 +1,10 @@
+#+TITLE: Emasc Info Imrprovements
+
+#+TAGS: emacs|layer
+
+* Description
+  This layer improves navigation for Emacs Info mode
+** Features:
```
smile13241324: Please add a blank line between headlines and the last line of the previous paragraph.
Also add a list of provided features please.

I would suggest:
```suggestion
** Features:
- Intuitive =transient state=
```

layers/+emacs/emacs-info/README.org
```diff
@@ -0,0 +1,10 @@
...
+   - TODO: list features
+* Key bindings
+  - TODO: list keys
```
smile13241324: You should document the key bindings in your layer please refer to the `helm-layer` for a good example of how to document transient key bindings.

layers/+emacs/emacs-info/README.org
```diff
@@ -0,0 +1,10 @@
+#+TITLE: Emasc Info Imrprovements
+
```
smile13241324: You are missing a table of content which is necessary for generating HTML documents from org.
```suggestion
* Table of Contents                     :TOC_5_gh:noexport:
- [[#description][Description]]
  - [[#features][Features:]]
- [[#install][Install]]
- [[#key-bindings][Key bindings]]
  - [[#transient-state][Transient state]]
```

layers/+emacs/emacs-info/README.org
```diff
@@ -0,0 +1,10 @@
...
+#+TAGS: emacs|layer
+
+* Description
+  This layer improves navigation for Emacs Info mode
+** Features:
+   - TODO: list features
+* Key bindings
```
smile13241324: You are missing a mandatory Install block
```suggestion
* Install
To use this configuration layer, add it to your =~/.spacemacs=. You will need to
add =emacs-info= to the existing =dotspacemacs-configuration-layers= list in this
file.
```
After this the key bindings section should follow. Have a look at the suggested table of content for the right order and level please.

layers/+emacs/emacs-info/packages.el
agzam: What? No, Org-agenda transient is bound to M-SPC. There's some other transient that also binds to M-SPC (I forgot which one).

smile13241324: I don't agree there is a difference between being OS agnostic and just being inflexible. A modern emacs distribution should not rely on default keybindings which 90% of users have to manually free from there OS.

Anyway that's just my personal opinion which in fact doesn't matter in this discussion. 

As to your comment, I have double checked the current version of `CONVENTIONS.org` which all spacemacs PRs shall follow. There in the section of `Transient-state` you will find the following paragraph:

```org
* Key bindings conventions
** Reserved prefix
...
...
*** Transient-state
Whenever possible a transient-state should be enabled with ~M-SPC~ and ~s-M-SPC~. We
need the latter bindings on OS X since ~M-SPC~ is used by the OS for spotlight.

For instance transient-states dedicated to special buffers like =helm= or =ido=
buffers are good candidates to be put on ~M-SPC~ and ~s-M-SPC~.

It is recommended to add ~q~ to leave the transient-state.
```
In this paragraph you can find that the requested binding of `s-M-SPC` for the transiant-state is required by the conventions and must be added. You are right not all layers are yet following the latest version of `CONVENTION.org` but this will change in time.

PR #12474, opened 2019-06-25
fix 12455 lsp layer

### LINKED ISSUE #12455: excessive delays inside which-key--maybe-replace when using lsp layer
#### Description :octocat:

I have observed performance degradation in which-key after a day's worth of development with spacemacs. Specifically, pressing the leader key and waiting for the menu to appear takes longer and longer.

Using the emacs profiler, I determined that almost all cpu time was being spent inside `which-key--maybe-replace`. Investigating further, it seems that the delays are due to growth of the `which-key-replacement-alist` variable, and this is in turn due to the `lsp` layer.

In the lsp layer, there is a call to `(spacemacs//lsp-declare-prefixes-for-mode major-mode)` as a hook attached to `lsp-after-open-hook`. This adds to the `which-key-replacement-alist` variable. Growth of this variable is correlated with the delays in showing the spacemacs menu.

- Start Emacs
- Open a file which uses LSP mode. I believe this could be any major-mode, but I am using c++-mode.
- Use `SPC b R` to revert this buffer
- Repeat the above step many times, or just open other files in the same LSP workspace
- Press leader key SPC and wait for menu to appear

The more files opened, or the more times each buffer is reverted, the longer it takes to the menu to appear. This is observed to be correlated with the increasing value of `(length (assoc major-mode which-key-replacement-alist))`.

The menu appears instantateously. Also the value of `(length (assoc major-mode which-key-replacement-alist))` is stable.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 26.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 37dcecbec)
- Graphic display: nil
- Distribution: spacemacs
- Editing style: emacs
- Completion: helm
- Layers:
```elisp
(systemd javascript html sql auto-completion better-defaults
         (c-c++ :variables c-c++-backend 'lsp-ccls c-c++-lsp-executable
                (expand-file-name "~/hack/extern/ccls/Release/ccls"))
         clojure dap
         (dash :variables helm-dash-docset-newpath
               (expand-file-name "~/.local/share/Zeal/Zeal/docsets"))
         emacs-lisp ess git helm markdown
         (org :variables org-enable-reveal-js-support t)
         python shell-scripts syntax-checking version-control)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GSETTINGS NOTIFY ACL LIBSELINUX GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 THREADS LIBSYSTEMD LCMS2

thanhvg:
fix #12455

add an internal variable for `lsp` layer `lsp-layer--active-mode-list`
to keep track of active major modes using lsp,
`spacemacs//lsp-declare-prefixes-for-mode` will check this variable
to decide whether to add prefix to which-key or not


duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/653a38bce52f1a69f60cb32cb4ea0b4cb89d0ea5

@RichardDRJ Thanks for testing this.

PR #12484, opened 2019-06-30
[core] Restore point in spacemacs/alternate-buffer

Miciah:
* `core/core-funcs.el` (`spacemacs/alternate-buffer`): Instead of using `switch-to-buffer`, use `set-window-buffer-start-and-point`, specifying the previous window start position and point if known.  Replace use of `cl-find-if` with `cl-find`.
* `CHANGELOG.develop`: Add entry for change to `spacemacs/alternate-buffer`.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -559,6 +559,7 @@ Other:
     (thanks to Alexander Miller)
   - Added missing force argument to the =spacemacs//init-spacemacs-env= call in
     the function =spacemacs/load-spacemacs-env= (thanks to sergeiz2)
+  - Fixed =spacemacs/alternate-buffer= to restore point.
 - Other:
   - New function =configuration-layer/message= to display message in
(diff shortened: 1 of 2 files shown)


duianto:
Confirmed.

- open `.spacemacs`: `SPC f e d`
- move the cursor to the bottom of the buffer `G`
- split the window: `SPC w s`
- move the cursor to the top of the buffer `gg`
- switch to the previous buffer and back: `SPC TAB` `SPC TAB`

#### Current behavior

The top window now has the cursor at the bottom of the buffer, in the same place as the cursor the bottom window.

#### With this PR applied

The cursor stays in the position it had before switching the buffers.
It's even possible to switch to another buffer in both windows, and it doesn't matter in what order the windows are switched back to the `.spacemacs` buffer, the cursor position remains where it was.

---

Commit: https://github.com/syl20bnr/spacemacs/commit/ab893e3af0211f552d8fe7cec93744dfbf664311

Issue #12188, opened 2019-04-16
spacemacs/enter-ahs-backward doesn't work if word starts with underscore


sid-kap:
#### Description :octocat:
spacemacs/enter-ahs-backward doesn't work if word starts with underscore

#### Reproduction guide :beetle:
- Start Emacs
- In evil mode, move the cursor over a word that starts with an underscore (e.g. `_hi`) and press <kbd>#</kbd>

*Observed behaviour:* :eyes: :broken_heart:
The word doesn't get highlighted in the darker color, and <kbd>n</kbd> and <kbd>p</kbd> don't do anything

*Expected behaviour:* :heart: :smile:
The word should get highlighted in the darker color, and <kbd>n</kbd> and <kbd>p</kbd> should go to the other instances of the word in the file

#### System Info :computer:
- OS: darwin
- Emacs: 26.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 8b3a30f24)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(yaml go javascript markdown
      (python :variables python-backend 'lsp)
      helm auto-completion better-defaults emacs-lisp git github neotree lsp groovy latex)
```
- System configuration features: NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS MODULES THREADS

thanhvg:
By default evil/emacs doesn't take underscore as word, you can find the fix at
 
http://spacemacs.org/doc/FAQ.html#include-underscores-in-word-motions

sid-kap:
Thanks!

PR #12487, opened 2019-06-30
update symbol regex when searching

ralexstokes:
There is a bug when searching for symbols that start with `_` on the transition out of the transient highlight mode and into regular text mode (searching w/ `n` in evil's normal mode).

This PR updates the searching regexp recalled upon `n` after `*` to produce the correct behavior.

The update mirrors the logic in the `auto-hightlight-symbol` package's function `ahs-search-symbol` (without setting a regexp group around the symbol) as `*` seems to do the correct thing.


duianto:
I'm not sure how to test the issue. Do you have some reproduction steps?

thanhvg:
I am not sure we need this. I think this problem has been addressed in the FAQ
http://spacemacs.org/doc/FAQ.html#include-underscores-in-word-motions
Also check out this closed https://github.com/syl20bnr/spacemacs/issues/12188

ralexstokes:
@thanhvg  yes, this seems to get the behavior i'm after working, thanks!

PR #12495, opened 2019-07-03
Add org-pomodoro keybinding in org-journal-mode.

klochowicz:
Add a missing keybinding for org-pomodoro when editing journal entries.

### DIFF
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -647,6 +647,8 @@ Headline^^            Visit entry^^               Filter^^                    Da
         (setq org-pomodoro-audio-player "/usr/bin/afplay"))
       (spacemacs/set-leader-keys-for-major-mode 'org-mode
         "Cp" 'org-pomodoro)
+      (spacemacs/set-leader-keys-for-major-mode 'org-journal-mode
+        "Cp" 'org-pomodoro)
       (spacemacs/set-leader-keys-for-major-mode 'org-agenda-mode
         "Cp" 'org-pomodoro))))
 


duianto:
`SPC m C p` seems to already be bound to `org-pomodoro` in a org journal buffer.

Maybe it's because of the line above the proposed one?
```elisp
        "Cp" 'org-pomodoro)
```

I tested it like this:
```elisp
(org :variables org-enable-org-journal-support t)
```
- open a new journal file `SPC a o j j`
- press `SPC m C p`

Now a pomodoro countdown timer appeared on the mode line.

I might be missing a case where it doesn't work without this specific org journal mode binding?

Or is there another benefit to specifying the key explicitly?

klochowicz:
Strange....

On my machine, after creating new journal file ( SPC a o j j ), the buffer is opened in org-journal major mode. The keybinding I tried to add by default results in following message:

"SPC m C p is undefined"

If I change the major mode manually (e.g. by using helm) to 'org-mode', then the aforementioned keybinding works, but then I lose other journal bindings. 
So the other benefit would be ensuring that this keybinding works even if the major mode is still 'org-journal'.

Are you sure you are in the 'org-journal' mode? Are other keybindings specific to this mode available? e.g. SPC m p (for going to previous journal entry?)

duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/d6c8df3d00b62f5626d34a243a7d2b1c79b0e8d7

Your right `SPC m C p` is undefined when I tried it again, maybe it was a loading order issue or something. I did open a new buffer and switched it to org-mode to see if `org-pomodore` was available there, and it is. But I'm unable to reproduce it appearing in the journal buffer again without this PR.

I didn't enable `org-mode` in the org-journal buffer. Anyway, it works ever time now with this PR, thanks again for submitting the fix. 👍

PR #12496, opened 2019-07-03
feat(clojure-lang): add `cider-eval-sexp-at-point`

dijonkitchen:
Adds a useful CIDER command that doesn't have an idiomatic Spacemacs command.

Ref:
https://cider.readthedocs.io/en/latest/interactive_programming/#using-cider-mode

### DIFF
--- a/layers/+lang/clojure/packages.el
+++ b/layers/+lang/clojure/packages.el
@@ -83,6 +83,7 @@
             "eb" 'cider-eval-buffer
             "ee" 'cider-eval-last-sexp
             "ef" 'cider-eval-defun-at-point
+            "ev" 'cider-eval-sexp-at-point
             "ei" 'cider-interrupt
             "em" 'cider-macroexpand-1
             "eM" 'cider-macroexpand-all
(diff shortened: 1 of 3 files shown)


layers/+lang/clojure/packages.el
```diff
@@ -83,6 +83,7 @@
             "eb" 'cider-eval-buffer
             "ee" 'cider-eval-last-sexp
             "ef" 'cider-eval-defun-at-point
+            "ev" 'cider-eval-sexp-at-point
```
duianto: Sort this alphabetically, like the other key bindings, that makes it easier to find which keys are used/available, without having to look through the whole list.

And add an entry for it in the clojure readme > evaluation section:
https://github.com/syl20bnr/spacemacs/tree/develop/layers/+lang/clojure#evaluation

duianto:
Squash the commits into one.

layers/+lang/clojure/packages.el
dijonkitchen: I actually tried to keep it sorted by the CIDER function so all the "eval" related ones are grouped together. Is that OK or is it better to sort by the command? 

I'll squash commits and add the docs as you note though.

duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/28adf374fee6739ee7ed95bbb43e3b1a68701977

I sorted it alphabetical by key binding to match the rest of the file.

PR #12502, opened 2019-07-05
Add j/k to evil-numbers transient state

jmromer:
A common use case for vim's / evil-mode's `inc-at-point`/`dec-at-point` is to change numbers across several lines, as in numbered lists. This patch adds `next-line` / `previous-line` to `evil-numbers-transient-state` to enable incrementing across lines without having to exist and restart the transient state.

#### Before

demo-before

#### After

demo-after

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -897,6 +897,7 @@ Other:
     - ~G~ go to last candidate
     (thanks to duianto)
   - Added support for visual line numbers (thanks to jcaw)
+  - Updated evil-numbers transient state to run foreign keys (thanks to Jake Romer)
 - Fixed:
   - Fixed ~h~ key binding in compilation and grep buffers
     (thanks to Sylvain Benner)
--- a/layers/+spacemacs/spacemacs-evil/packages.el
+++ b/layers/+spacemacs/spacemacs-evil/packages.el
@@ -283,6 +283,7 @@
         :title "Evil Numbers Transient State"
         :doc
         "\n[_+_/_=_] increase number  [_-_] decrease  [0..9] prefix  [_q_] quit"
+        :foreign-keys run
         :bindings
         ("+" evil-numbers/inc-at-pt)
         ("=" evil-numbers/inc-at-pt)


duianto:
I noticed that the `Code Fold Transient State` (`SPC z .`) seems to allow for using any of the keys that aren't bound in the transient state. But it doesn't bind the keys individually.

https://github.com/syl20bnr/spacemacs/blob/28adf374fee6739ee7ed95bbb43e3b1a68701977/layers/%2Bdistributions/spacemacs-bootstrap/packages.el#L230-L240

The code fold TS defines a property: `:foreign-keys run`
The docstring for `spacemacs|define-transient-state` says this about it:
https://github.com/syl20bnr/spacemacs/blob/28adf374fee6739ee7ed95bbb43e3b1a68701977/core/core-transient-state.el#L143-L148

I just tried adding the property to the `Evil Numbers Transient State`
```elisp
      (spacemacs|define-transient-state evil-numbers
        :title "Evil Numbers Transient State"
        :doc
        "\n[_+_/_=_] increase number  [_-_] decrease  [0..9] prefix  [_q_] quit"
        :foreign-keys run
        :bindings
        ("+" evil-numbers/inc-at-pt)
        ("=" evil-numbers/inc-at-pt)
        ("-" evil-numbers/dec-at-pt)
        ("q" nil :exit t))
```
And it seems to work quite nicely.

duianto:
Squash the commits to one.
And move the changelog.develop entry to the:
in the `Improvements` subsection.

jmromer:
I'm aware of `:foreign-keys run`, but I think that might create confusion—while test-driving that approach I found myself navigating away without explicitly quitting the transient state. 

I also can't think of other foreign keys (i.e., aside from navigating up/down) one would want to run from within the transient state. 

If there are other use cases I'm missing though, I agree it would make sense to just add `:foreign-keys run`. Let me know what you think.

duianto:
It seems like it would be useful to also be able to move to the left and right if numbers aren't aligned vertically.

For example when changing numbers that are comma separated: `[1, 2, 3, 4, 5]`
Then it can be useful to jump to the next/previous number with `W` and `B`.

jmromer:
Good call thanks -- updated in 48396ce

duianto:

PR #12503, opened 2019-07-05
core: Ugly fix for set-default-font

derkling:
Since my elisp knowledge is pretty limited, I'm not really sure why but it turns out this small change makes `dotspacemacs-default-font` working as expected.

Either by:
 1) using `spacemacs-buffer/message` to print `fontspec`
    or
 2) using `message` to print font
is not working, unless emacs is started with `--debug-init`.

The change above is the only (nonsense) combination which makes fonts settings working in my setup. With that change, I can successfully both change font (among the installed ones) and its size. Without that change, whatever settings I try the font is not set.

So, the patch is ugly because it "fixes" an issue without really pointing out what's the issue. Nevertheless, I hope the above change can help more knowledgable folks to pinpoint the real problem.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -95,6 +95,7 @@ the [[file:CHANGELOG.org][CHANGELOG.org]] file.
 - The property =:powerline-scale= of variable =dotspacemacs-default-font= has
   been removed and replaced by the property =:separator-scale= used in the new
   dotfile variable =dotspacemacs-mode-line-theme=.
+- Fixup =dotspacemacs-default-font- to properly set the configured font.  
 **** Layers
 ***** Spacemacs distribution layers
 - Key bindings:
--- a/core/core-fonts-support.el
+++ b/core/core-fonts-support.el
@@ -34,7 +34,7 @@ The return value is nil if no font was found, truthy otherwise."
                             (spacemacs/mplist-remove props :powerline-scale)
                             :powerline-offset))
                (fontspec (apply 'font-spec :name font font-props)))
-          (spacemacs-buffer/message "Setting font \"%s\"..." font)
+          (message "Using font: [%s]" fontspec)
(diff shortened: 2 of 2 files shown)


derkling:
Verified this patch on a different system and... it does not always work.
I've got one time spacemacs coming up with the correct font but after a couple of restarts it does not change the font again. There is something really odd going on with `set-default-font`.

Here is my current setting:
```
dotspacemacs-default-font '("Ubuntu Mono"
                               :size 8
                               :weight normal
                               :width normal
                               :powerline-scale 1.1)
```

with the setting above I cannot always get the font I want. While, by running `menu-set-font` it just pick up that same font and size correctly.

smile13241324:
@derkling I have just double checked the code you have changed and to be honest this should not have any effect on changing the font at all. Basically `spacemacs-buffer/message` is only our way for coding a `log.log_when_in_debug` function.

```elisp
(defun spacemacs-buffer/message (msg &rest args)
  "Display MSG in *Messages* prepended with '(Spacemacs)'.
The message is displayed only if `init-file-debug' is non nil.
ARGS: format string arguments."
  (when init-file-debug
    (message "(Spacemacs) %s" (apply 'format msg args))))
```
It can still be that there is an issue with your emacs though. May I ask on which OS you are? I would assume Ubuntu judging from the used font but please confirm. Also I would like to have some more information about the setup you are using i.e. are you starting a vanilla spacemacs session or do you connect to a remote spacemacs-server with emacs-client? And last but not least are you using an official build from your OS or a local build from master?

With the current level of information it is impossible for us to get any insight into your problem.

Btw: The best way to provide these information is to use the official issue template from within spacemacs with `SPC SPC spacemacs/report-issue` from there provide the `System Info` block please.

smile13241324:
Closed due to very long in activity and no real chance of getting the proposed changes be merged.

PR #12504, opened 2019-07-05
Enables eldoc for anaconda-mode

vikigenius:
Enables eldoc in anaconda mode which was currently not working, because lsp already sets it up on it's own i think ?


duianto:
Confirmed.
The `pcase` docstring says:
>CASES is a list of elements of the form (PATTERN CODE...).

But it's only getting: `(spacemacs//python-setup-anaconda-eldoc)`

Commit: https://github.com/syl20bnr/spacemacs/commit/3ed24436cd325ca2f897e583437b476072149127

PR #12505, opened 2019-07-06
Fixes flycheck support in jsonnet files

vikigenius:
jsonnet syntax checking does not seem to work in jsonnet files, this fixes the issue

### DIFF
--- a/layers/+lang/jsonnet/packages.el
+++ b/layers/+lang/jsonnet/packages.el
@@ -15,7 +15,7 @@
         flycheck
         ))
 
-(defun javascript/post-init-flycheck ()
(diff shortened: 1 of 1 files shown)


duianto:
Confirmed.
When the `jsonnet` layer is enabled. Then this warning message appears on startup:
>Warnings:
    - package flycheck not initialized in layer jsonnet, you may consider removing
      this package from the package list or use the :toggle keyword instead of a
      `when' form.

The warning disappears when this PR is applied.

Commit: https://github.com/syl20bnr/spacemacs/commit/5def0eaae50853ba3ddb44159f1e9e1de5c6659f

PR #12516, opened 2019-07-09
Go linters

robert-zaremba:
`layers/+lang/go`: refactore the extra flycheck linter setup.

It doesn't make sense to have two variables: `go-use-gometalinter` and `go-use-golangci-lint` as only one of them can be in use.


duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/d6420953803217a7170d52a4466aa8520a55c37d

robert-zaremba:
Thanks

PR #12520, opened 2019-07-11
[Clojure] Keybinding require Clojure repl utils in current namespace

jr0cket:
Although the Clojure repl utils (clojure.repl) are including in the `user`
namespace in Clojure projects, they are not available in your application
namespace.

The existing CIDER function `cider-repl-require-repl-utils` adds the Clojure
repl utils into the current namespace, so you can use functions such as `doc`
and `source`.

This is often a source of confusion for those new to Clojure / CIDER.


jr0cket:
My bad.  I forgot to rebase before pushing.  I will fix asap

duianto:

PR #12522, opened 2019-07-11
Feature sql improvements

mijoharas:
add sql-connect function as it allows you to select from saved sql connections.

NOTE:

I have some things in my private spacemacs and I'm wondering if I should contribute it along with this Pr:
it would define a predifined place for people to put passwords for specific databases that they connect to. What do you think? should I roll it in and write some docs or not (don't worry, I'd change the function name!)
```
  ;; sql stuff
  (defun mike-load-sql-passwords ()
      (interactive)
      (load-library "~/sql-credentials.el.gpg")
      (setq sql-connection-alist (delete-dups (append sql-connection-alist mike-sql-credentials)))
    )

  (spacemacs/set-leader-keys-for-major-mode 'sql-mode
    "op" 'mike-load-sql-passwords)
```

### DIFF
--- a/layers/+lang/sql/README.org
+++ b/layers/+lang/sql/README.org
@@ -104,12 +104,13 @@ auto-indent by setting the variable =sql-auto-indent= to =nil=.
 
 ** Inferior Process Interactions (SQLi)
 
-| Key binding | Description                 |
-|-------------+-----------------------------|
-| ~SPC m b b~ | show the SQLi buffer name   |
-| ~SPC m b s~ | set the SQLi buffer         |
-| ~SPC m l a~ | List all objects            |
-| ~SPC m l t~ | list all objects in a table |
+| Key binding | Description                                          |
+|-------------+------------------------------------------------------|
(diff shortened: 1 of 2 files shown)


duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/ae7139612bcc6343e6720d4f8ecfae126375ff89

### Changes made to this PR
- Squashed the commits
- Added a changelog.develop entry in the Sql section
- Sorted both the documentation and code key binding alphabetically.

### About the Passwords PR
Open a new PR, and mention `@smile13241324` in the PR.

Smile said this when I asked about your proposed new PR:
>I am working a lot with sql and normally my projects get their credentials from environment variables or other upstream sources, i.e. kubernetes config files. For local development I normally have my db configured to only accept connections from localhost so at least I have never had the need to have a local credential store for my sql databases. But there may be other workflows so why not let him add a function for this, as long as it interfaces with gpg or any other encrypted storage this could save some keystrokes for some users.

If you have thoughts/questions about it, then the best place to discuss them probably are in the new PR.

PR #12525, opened 2019-07-12
[Clojure] keybinding spacemacs/clj-find-var

jr0cket:
Update the existing `SPC m g g` keybinding to use the command
`spacemacs/clj-find-var`.  

This makes finding a function definition a much
better experience as you dont have to have the REPL running to find a
definition, but when it is you can use a CIDER specific function.

`spacemacs/clj-find-var` is a wrapper that calls `cider-find-var` if the REPL is
running, otherwise it uses `dump-jump-go`.

Wrapper added in #9792

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1198,6 +1198,8 @@ Other:
     (thanks to John Stevenson)
   - ~SPC m s j s~ updated to call =cider-jack-in-cljs= rather than old alias
     =cider-jack-in-clojurescript= which is deprecated
+  - ~SPC m g g~ changed to =spacemacs/clj-find-var=, a wrapper for
+    =cider-find-var= if REPL running, otherwise =dumb-jump-go=
     (thanks to John Stevenson)
 - Fixes:
   - Remove `cider.nrepl/cider-middleware` in lein quick start setting
--- a/layers/+lang/clojure/README.org
+++ b/layers/+lang/clojure/README.org
@@ -263,16 +263,16 @@ As this state works the same for all files, the documentation is in global
 
 *** Goto
 
-| Key binding | Description      |
-|-------------+------------------|
-| ~SPC m g b~ | go back          |
-| ~SPC m g C~ | browse classpath |
-| ~SPC m g g~ | goto var         |
-| ~SPC m g e~ | goto error       |
-| ~SPC m g n~ | goto namespace   |
-| ~SPC m g r~ | goto resource    |
-| ~SPC m g s~ | browse spec      |
-| ~SPC m g S~ | browse all specs |
+| Key binding | Description                                  |
+|-------------+----------------------------------------------|
+| ~SPC m g b~ | go back                                      |
+| ~SPC m g C~ | browse classpath                             |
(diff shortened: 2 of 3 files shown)


duianto:
There's a Spacemacs and Helm lazy loading issue when `,gg` is pressed before helm has loaded.

I'll cherry-pick this PR anyway because it works fine when Helm has been loaded, for example after pressing `SPC SPC` to open the `HELM M-x` buffer.

### Issue (unrelated to this PR)
When `,gg` is pressed on for example `defn` or `vec` near the top of this example clojure file:
https://github.com/clojurebook/ClojureProgramming/blob/master/ch03-game-of-life/src/com/clojurebook/collections/life.clj
before helm has been loaded, then this message appears:
>dumb-jump-prompt-user-for-choice: Symbol’s function definition is void: helm-make-source

### Failed attempt at a solution

A similar issue has been solved using a transient-hook for the commands:
`completing-read`
`completion-at-point`
and
`read-file-name`

The PR's where the solutions were discussed can be found here:
Load helm before read-file-name and completing-read #11877
[helm] Load helm before completion-at-point #12399

I tried adding what seems to be the source function:
`dumb-jump-prompt-user-for-choice` as a transient hook:
```elisp
      (spacemacs|add-transient-hook dumb-jump-prompt-user-for-choice
        (lambda (&rest _args) (require 'helm))
        lazy-load-helm-for-dumb-jump-prompt-user-for-choice)
```
below the other helm transient hooks:
https://github.com/syl20bnr/spacemacs/blob/034ee0c3119815c9b17d6a6c8ab4f8ec8b905376/layers/%2Bcompletion/helm/packages.el#L74

but it doesn't seem to work.

jr0cket:
Looking back at my notes I fixed the issue with `dumb-jump-go` by simply deleting the `dumb-jump.elc` file from the dump-jump package and running `SPC SPC spacemacs/recompile-elpa`

duianto:

duianto:
It does work when `dumb-jump.elc` is deleted. Haven't tried rebuilding the file or elpa.

But the issue returns when the package is installed after deleting the `elpa/dumb-jump-*` directory and restarting Spacemacs.

So we need to find a way to load helm before the correct command that needs it.

duianto:
I'm now getting another helm void message before opening a helm buffer:
>dumb-jump-prompt-user-for-choice: Symbol’s function definition is void: helm-build-sync-source

And I don't know what's changed but I'm getting an slightly different error message after opening a helm buffer as well:
>dumb-jump-prompt-user-for-choice: Invalid function: helm-build-sync-source

I found this issue on the dumb-jump repo that mentions the second error:
dumb-jump-prompt-user-for-choice: Invalid function: helm-build-sync-source #153

But it doesn't mention Spacemacs, so I might not be Spacemacs related.

There's a suggestion on that issue to reinstall helm, so I moved all helm packages out of the elpa directory:
`C:\Users\username\.emacs.d\elpa\26.2\develop\`

After restarting most of the helm packages were reinstalled (some of them probably are specific to some layers that I don't have enabled at the moment).

It didn't help, so I tried removing and reinstalling dumb-jump, but both error messages still appear.

PR #12549, opened 2019-07-21
stop pulling in treemacs/helm when their respective layers aren't enabled

Stebalien:


duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/c0763a32405d1b119038f6a24b089daeeacb59e2

yyoncho:
@duianto @syl20bnr we are planning to port the gui controls in dap-mode to use treemacs as a tree renderer. This means that dap-mode will have to have a dependency on the treemacs *layer* which technically does not exist. Can you take a look and coment whether adding `:require treemacs` is really needed when treemacs is used as a library.

syl20bnr:
@yyoncho internal dependencies are installed automatically by `package.el` so we don't need to put `:requires treemacs` in the layer.

PR #12561, opened 2019-07-25
Add new var treemacs-enable-magit to enable treemacs-magit

eklitzke:
This adds a new var `treemacs-enable-magit` to enable the `treemacs-magit` package, which lets treemacs know when files are staged/unstaged using magit. I think it would be better to automatically enable this if a user has both the `treemacs` and `magit` layers enabled (rather than using a config variable) but I'm not sure how to do this; if someone wants to give me a pointer I can try to rework the logic.

The `treemacs-magit` package is described in the upstream treemacs README: https://github.com/Alexander-Miller/treemacs

The `treemacs-magit` package is available in MELPA: https://melpa.org/#/treemacs-magit

### DIFF
--- a/layers/+filetree/treemacs/config.el
+++ b/layers/+filetree/treemacs/config.el
@@ -34,3 +34,6 @@ There are 2 possible values:
 
 (defvar treemacs-lock-width nil
   "When non-nil the treemacs window will not be manually resizable by default.")
+
+(defvar treemacs-enable-magit nil
+  "When non-nil, enable integration with magit.")
--- a/layers/+filetree/treemacs/packages.el
+++ b/layers/+filetree/treemacs/packages.el
@@ -14,6 +14,7 @@
     golden-ratio
     treemacs
     (treemacs-evil :toggle (memq dotspacemacs-editing-style '(vim hybrid)))
+    (treemacs-magit :toggle treemacs-enable-magit)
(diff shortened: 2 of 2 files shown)


duianto:
This seems to only install `treemacs-magit` if both `magit` and `treemacs` are used:

```elisp
    (treemacs-magit :toggle (and (configuration-layer/package-used-p 'magit)
                                 (configuration-layer/package-used-p 'treemacs)))
```

smile13241324:
@eklitzke I agree having treemacs-layer load treemacs-magit if magit and treemacs are available is the right way to go, much more understandable than a new config variable.

Was the hint from duianto enough for you to make the change or do you require some help?

syl20bnr:
Agree. Use the `:require` keyword in the `treemacs` layer.
You can find an example here: https://github.com/syl20bnr/spacemacs/blob/develop/layers/+lang/python/packages.el#L24

This way the `magit` integration for `treemacs` will be enabled only if the `magit` package is effectively used.

syl20bnr:
Better example: https://github.com/syl20bnr/spacemacs/blob/develop/layers/+tools/lsp/packages.el#L18

smile13241324:
I have made the necessary changes to load `treemacs-magit` when `magit` is loaded.

PR #12567, opened 2019-07-27
Correct function-format keybinding in EL layer

adlawren:
It appears that the illustrated keybinding isn't defined in the layer; I think [SPC m = d](https://github.com/syl20bnr/spacemacs/blob/master/layers/%2Blang/emacs-lisp/packages.el#L213) was intended.

### DIFF
--- a/layers/+lang/emacs-lisp/README.org
+++ b/layers/+lang/emacs-lisp/README.org
@@ -207,7 +207,7 @@ The [[https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Bemacs/semanti
 | Key binding | Description             |
 |-------------+-------------------------|
 | ~SPC m = b~ | format current buffer   |
-| ~SPC m = f~ | format current function |
+| ~SPC m = d~ | format current function |
 | ~SPC m = o~ | format all on one line  |
 | ~SPC m = s~ | format current sexp     |
 


duianto:
Your right that `SPC m = d` is bound:
https://github.com/syl20bnr/spacemacs/blob/a7e39b910fd64320d766f4a70052133b588457ba/layers/%2Blang/emacs-lisp/packages.el#L290

but the readme shows `SPC m = f`.

The readme `Format code` section:
https://github.com/syl20bnr/spacemacs/blob/d04ec0b93623d2dadcbc1ec2f3fa4427d05829cf/layers/%2Blang/emacs-lisp/README.org#format-code
says:
>The semantic layer should be installed for these key bindings to become active.

the text `semantic` links to the layer with the same name:
https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Bemacs/semantic/README.org

and the `semantic` layers readme links to the `semantic-refactor` package:
https://github.com/tuhdo/semantic-refactor

which suggests `d` as the last key in the key binding for the same command:
```
    (global-set-key (kbd "M-RET d") 'srefactor-lisp-format-defun)
```
https://github.com/tuhdo/semantic-refactor#installation

Therefore the currently bound key binding `SPC m = d` matches the upstream `semantic-refactor` package key binding.

But three other `emacs-lisp` layer key bindings that are referring to functions use `f` as the final key.
- `SPC m e f` | evaluation current function
- `SPC m d f` | on a defun symbol toggle on the instrumentalisation of the function
- `SPC m d F` | on a defun symbol toggle off the instrumentalisation of the function

The question is if all the function related key bindings should be consistent throughout the `emacs-lisp` layer or if the suggested key binding from the upstream `semantic-refactor` package should take precedence.

smile13241324:
Thank you for contributing to spacemacs :purple_heart:, I have cherry picked your PR into develop.

@duianto I have merged this PR for the proposed changes adjust the documentation to the real binding situation and I think `=d` is more idiomatic to lispers then `=f` aka `defun` is more common then `functions` in this context.

For the various bindings against `f` I think that the different parent bindings provide different contexts so I would keep them unharmonized.

PR #12568, opened 2019-07-28
Layout layer: counsel-projectile: requires fix

thanhvg:
Guard `:requires` against `ivy` not `projectile`.` projectile` is already owned by
buil-in `spacemacs-project` layer, so it's always `t`.

This prevents install of `ivy`, `counsel`, `swiper` and `counsel-projectile`
packages regardless of `helm` layer.


duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/7bc23b895fe700911f49cb5fa8c2ee11c3164f04

PR #12570, opened 2019-07-29
[doc] Clarify line wrapping vs truncation

### LINKED ISSUE #7962: spacemacs/toggle-truncate-lines-off enables truncation, and vice versa
`(spacemacs/toggle-truncate-lines-off)` = enable truncation
`(spacemacs/toggle-truncate-lines-on)` = disable truncation

No idea why. keybindings.el looks okay:

```emacs-lisp
(spacemacs|add-toggle truncate-lines
  :status truncate-lines
  :on (toggle-truncate-lines)
  :off (toggle-truncate-lines -1)
  :documentation "Truncate long lines (no wrap)."
  :evil-leader "tl")
```

Edit: I know why... seems like it is a bug in the upstream package...

`(toggle-truncate-lines -1)` outputs the message "disabling" and enables it. And `(toggle-truncate-lines 1)` outputs the message "enabling" and disables it.

Edit 2: Here is the code from simple.el, it seems this was a built-in emacs function and that Spacemacs is possibly misusing it:

```emacs-lisp
(defun toggle-truncate-lines (&optional arg)
  "Toggle truncating of long lines for the current buffer.
When truncating is off, long lines are folded.
With prefix argument ARG, truncate long lines if ARG is positive,
otherwise fold them.  Note that in side-by-side windows, this
command has no effect if `truncate-partial-width-windows' is
non-nil."
  (interactive "P")
  (setq truncate-lines
	(if (null arg)
	    (not truncate-lines)
	  (> (prefix-numeric-value arg) 0)))
  (force-mode-line-update)
  (unless truncate-lines
    (let ((buffer (current-buffer)))
      (walk-windows (lambda (window)
		      (if (eq buffer (window-buffer window))
			  (set-window-hscroll window 0)))
		    nil t)))
  (message "Truncate long lines %s"
	   (if truncate-lines "enabled" "disabled")))
```

Edit 3: May also be a bug in my port of Emacs for Mac (the from by d12frosted). Because it seems truncation is setting a truncate-lines property and just reading it...

Edit 4: Ehh, wait... "When truncating is off, long lines are folded." - Now I see, I think I thought about this backwards. I thought truncation meant wrapping the lines (truncating them to make them shorter). But it turns out that truncating off = "Do word-wrapping to display the ENTIRE line without cutting it off at the right and showing a 'right-arrow for more' icon". God damn that's confusing. Whoever designed that function should be slapped. ;) Because "truncating" is mentally the same as wrapping (truncating a line early to wrap it). So wrapping should be truncating == on.

I will leave this ticket open to let you consider if there's something Spacemacs can do to improve this situation and fix the confusing terminology, perhaps by clearer documentation in the toggle-wrapper you've added `"Truncate long lines (no wrap)."` is not clear enough until I `wrapped` my head around this weird function. Anyway, close this ticket at will.

duianto:
Resolves #7962

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/keybindings.el
+++ b/layers/+spacemacs/spacemacs-defaults/keybindings.el
@@ -332,7 +332,7 @@
   :status truncate-lines
   :on (toggle-truncate-lines)
   :off (toggle-truncate-lines -1)
-  :documentation "Truncate long lines (no wrap)."
+  :documentation "Toggle between line wrapping or truncation (no wrap)."
   :evil-leader "tl")
 (spacemacs|add-toggle visual-line-navigation
   :status visual-line-mode


smile13241324:
Merged, however I am not sure if your changed doc string is shown somewhere, on mouse-hover I only see `Toggle truncate-lines on and off` same on `spc h d k spc t l`

duianto:
Issue #7962 mentioned the entry:
https://github.com/syl20bnr/spacemacs/blob/4bcd3402fec8e9ca3e4f0b99d029644990c73a37/layers/%2Bspacemacs/spacemacs-defaults/keybindings.el#L331-L336

It's shown in the `SPC h SPC` `Toggles` section (search for `trunc`).

But your right, the confusing documentation remains for:
- When the mouse is over the `SPC t` which-key entry `l -> truncate-lines`
- And in the help buffer `SPC h d k SPC t l`

Seem to be set as the docstring for the automatically generated `spacemacs/toggle-truncate-lines` function in `spacemacs|add-toggle`, based on the `name` (first argument) that the toggle function receives.
https://github.com/syl20bnr/spacemacs/blob/4bcd3402fec8e9ca3e4f0b99d029644990c73a37/core/core-toggle.el#L106

One solution might be to add the toggle section documentation to the docstring.
```elisp
,(format "Toggle %s on and off.%s"
                  (symbol-name name)
                  (if doc (concat "\n\n" doc) "")
```
That results in:
#### Mouse over:
`Toggle truncate-lines on and off.  Toggle between line wrapping or truncation (no wrap).`

#### Docstring:
> SPC t l runs the command spacemacs/toggle-truncate-lines (found in
> evil-normal-state-local-map), which is an interactive Lisp function in
> ‘../../keybindings.el’.
>
> It is bound to SPC t l, M-m t l.
>
> (spacemacs/toggle-truncate-lines)
>
>Toggle truncate-lines on and off.
>
>Toggle between line wrapping or truncation (no wrap).

### Issue:
Currently the add-toggle `:documentation` string for smartparens in the `SPC h SPC` toggles section says:
>(+) Smartparens: Enable smartparens. (SPC t p)

And that works, the `(+)` or `(-)` matches the description, if it is enabled or disabled.

But if the proposed change is applied, then the mouse over text for `SPC t` `p -> smartparens` will say:
`Toggle smartparens on and off.  Enable smartparens.`

and the docstring for `SPC t p` says:
> It is bound to SPC t p, M-m t p.
> 
> (spacemacs/toggle-smartparens)
> 
> Toggle smartparens on and off.
> 
> Enable smartparens.

even though smartparens already is enabled.

smile13241324:
Thanks for the explanation, yeah I think it would really make sense adding the docstring to the mouse over documentation. I am afraid that not many people will search for toggle in `SPC h SPC`.

duianto:
I opened a PR:
Add toggle key docs to docstring and which-key mouseover #12590

PR #12572, opened 2019-07-30
Remove Unicode triple dot in .spacemacs file in favor of ASCII one

nshadov:
Remove Unicode triple dot from .spacemacs file in favor of ASCII one. 
Unicode character in comment breaks initialization with message:

"*temp file*" default encoding cannot encode

It's was tricky to debug because of breaking in intermediary "temp file"
and regular error message was unable to point back to original file or
"temp file" contents.


duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/0e0319968db33a0846252d4cc42b96e1f0d46617

PR #12581, opened 2019-08-01
spacemacs-layout layer: Make commands to work within layouts

thanhvg:
spacemacs-layout layer: limit commmands to work within layouts

Add options to restrict some functions to current layout's buffers. Default is off.

Thanks smile13241324 and duianto for suggestions and fixing typos

### DIFF
--- a/layers/+spacemacs/spacemacs-layouts/README.org
+++ b/layers/+spacemacs/spacemacs-layouts/README.org
@@ -5,6 +5,7 @@
 * Table of Contents                     :TOC_5_gh:noexport:
 - [[#description][Description]]
   - [[#features][Features:]]
+- [[#limit-functions-to-the-current-layouts-buffers][Limit functions to the current layout's buffers]]
 
 * Description
 This layer adds support for distinct layouts/workspaces to Spacemacs.
@@ -12,3 +13,35 @@ This layer adds support for distinct layouts/workspaces to Spacemacs.
 ** Features:
 - Support for distinct layouts via =eyebrowse=
 - Integration with =helm= and =ivy= to search for buffers within layouts
+
+* Limit functions to the current layout's buffers
+Besides =helm-mini= and =ivy-switch-buffer= that are aware of a layouts buffers.
+This layer provides variable ~spacemacs-layouts-advice-list~ holding a list of
+functions that need to be limited to current layout's buffers. They will be
+wrapped around by persp-mode ~with-persp-buffer-list~. 
+
+Default value of ~spacemacs-layouts-advice-list~ is
+~'(spacemacs/window-split-double-columns spacemacs/window-split-triple-columns
+spacemacs/window-split-grid)~.
+
+Another variable  ~spacemacs-layouts-restrict-spc-tab~ when set to ~t~ will 
+limit ~spacemacs/alternate-buffer~ (~SPC TAB~) to current layout's buffers.
+
+Default value of ~spacemacs-layouts-restrict-spc-tab~ is nil
+
+If you want to modify them, edit the ~dotspacemacs-configuration-layers~
+section, for example:
+ 
+ #+begin_example elisp
+  (spacemacs-layouts :variables
+                     spacemacs-layouts-restrict-spc-tab t
+                     spacemacs-layouts-advice-list '(spacemacs/window-split-double-columns
+                                                     spacemacs/window-split-triple-columns
+                                                     spacemacs/window-split-grid
+                                                     helm-do-ag-buffers))
+ #+end_example
+
+Note that ~spacemacs-layouts-restrict-spc-tab~ can be toggle on the fly with
+~setq~ or can be set in ~dotspacemacs/user-config~, while
+~spacemacs-layouts-advice-list~ can be changed only at
+~dotspacemacs-configuration-layers~ section
--- a/layers/+spacemacs/spacemacs-layouts/config.el
+++ b/layers/+spacemacs/spacemacs-layouts/config.el
@@ -55,3 +55,11 @@ layout, the 4th for the 4th, and so on until the 10th (aka layout
 number 0). The first list is sepcial - it is a grab-bag for names
 in case none of the regular names can be used for a new layout.")
 
+(defvar spacemacs-layouts-advice-list
+  '(spacemacs/window-split-double-columns
+    spacemacs/window-split-triple-columns
+    spacemacs/window-split-grid)
+  "list of functions to be wrapped by `with-persp-buffer-list'")
+
+(defvar spacemacs-layouts-restrict-spc-tab nil
+  "If `t' then `SPC-TAB' will be limited to the current layouts buffers.")
--- a/layers/+spacemacs/spacemacs-layouts/funcs.el
+++ b/layers/+spacemacs/spacemacs-layouts/funcs.el
@@ -80,6 +80,9 @@ Cancels autosave on exiting perspectives mode."
   (let ((ivy-ignore-buffers (remove #'spacemacs//layout-not-contains-buffer-p ivy-ignore-buffers)))
     (ivy-switch-buffer)))
 
+(defun spacemacs-layouts//advice-with-persp-buffer-list (orig-fun &rest args)
+  "Advice to provide perp buffer list."
+  (with-persp-buffer-list () (apply orig-fun args)))
 
 ;; Persp transient-state
 
@@ -289,7 +292,7 @@ Available PROPS:
                                        ,binding ,already-defined? ,name)
              (setq spacemacs--custom-layout-alist
                    (delete (assoc ,binding spacemacs--custom-layout-alist)
-                     spacemacs--custom-layout-alist))
+                           spacemacs--custom-layout-alist))
              (push '(,binding . ,name) spacemacs--custom-layout-alist))
          (push '(,binding . ,name) spacemacs--custom-layout-alist)))))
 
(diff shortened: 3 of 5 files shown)


layers/+spacemacs/spacemacs-layouts/README.org
```diff
@@ -5,10 +5,61 @@
...
   - [[#features][Features:]]
+- [[#make-funtions-work-within-layouts-buffers-only][Make funtions work within layout's buffers only]]
 
 * Description
 This layer adds support for distinct layouts/workspaces to Spacemacs.
 
 ** Features:
 - Support for distinct layouts via =eyebrowse=
 - Integration with =helm= and =ivy= to search for buffers within layouts
+* Make funtions work within layout's buffers only
```
smile13241324: Please add an empty line between the features paragraph and your new headline.

layers/+spacemacs/spacemacs-layouts/README.org
```diff
@@ -5,10 +5,61 @@
...
+ #+begin_example
+  next-buffer
+  previous-buffer
+  spacemacs/window-split-double-columns
+  spacemacs/window-split-single-column
+  spacemacs/window-split-triple-columns
+  spacemacs/window-split-grid
+ #+end_example
+
+ They are wrapped around by persp-mode ~with-persp-buffer-list~. The varialbe
```
smile13241324: small typo varialbe should be variable

layers/+spacemacs/spacemacs-layouts/README.org
```diff
@@ -5,10 +5,61 @@
...
+ ~spacemacs-layouts-advice-list~ holds the list of these functions. Basically
+ you can try adding functions that call ~buffer-list~. For example you can add
+ ~helm-do-ag-buffers~ to the list if you want it to search for buffers of the
+ current layout only not all the buffers accross all layouts.
+ 
+ Function ~spacemacs/alternate-buffer(SPC Tab)~ is also made to be work within
+ layouts. If you want the old behavior set ~spacemacs-layouts-restrict-spc-tab~
+ to nil.
+ 
+ To cusmomize, if you don't want any of these changes add this to your
```
smile13241324: small typo cusmomize should be customize

layers/+spacemacs/spacemacs-layouts/README.org
```diff
@@ -5,10 +5,61 @@
...
+- [[#make-funtions-work-within-layouts-buffers-only][Make funtions work within layout's buffers only]]
 
 * Description
 This layer adds support for distinct layouts/workspaces to Spacemacs.
 
 ** Features:
 - Support for distinct layouts via =eyebrowse=
 - Integration with =helm= and =ivy= to search for buffers within layouts
+* Make funtions work within layout's buffers only
+ Besides =helm-min= and =ivy= that are aware of layout's buffers the follwing
```
smile13241324: Small typo, follwing should be following.

thanhvg:
@smile13241324 thanks for your review, I fixed the typos

layers/+spacemacs/spacemacs-layouts/README.org
```diff
@@ -5,10 +5,53 @@
...
+ 
+ To customize, if you don't want any of these changes add this to your
+ ~dotspacemacs/layers~
+ #+begin_example elisp
+ (spacemacs-layouts :variables
+                    spacemacs-layouts-restrict-spc-tab nil
+                    spacemacs-layouts-advice-list '())
+ #+end_example
+ 
+ if you enjoy this feature and want to add ~helm-do-ag-buffers~ to it:
```
duianto: This line was moved up, below the line:
>current layout and not through all the buffers across all layouts.
```suggestion
```

On line 35 in this review, but it might change when the review changes are applied, line wrapping etc.

layers/+spacemacs/spacemacs-layouts/README.org
```diff
@@ -5,10 +5,53 @@
...
+ #+begin_example elisp
```
duianto: Move up this example block, below the line:
>If you enjoy this feature and want to add \~helm-do-ag-buffers\~ to it:

In this review it's on line 35, but it might change when the review changes are applied.

layers/+spacemacs/spacemacs-layouts/funcs.el
```diff
@@ -80,6 +80,9 @@ Cancels autosave on exiting perspectives mode."
   (let ((ivy-ignore-buffers (remove #'spacemacs//layout-not-contains-buffer-p ivy-ignore-buffers)))
     (ivy-switch-buffer)))
 
+(defun spacemacs-layouts//advice-with-persp-buffer-list (orig-fun &rest args)
+  "advice to provide perp buffer list"
```
duianto:
```suggestion
  "Advice to provide perp buffer list."
```

layers/+spacemacs/spacemacs-layouts/config.el
```diff
@@ -55,3 +55,14 @@ layout, the 4th for the 4th, and so on until the 10th (aka layout
...
+  '(next-buffer
+    previous-buffer
+    spacemacs/window-split-double-columns
+    spacemacs/window-split-single-column
+    spacemacs/window-split-triple-columns
+    spacemacs/window-split-grid)
+  "list of functions to be wrapped by `with-persp-buffer-list'")
+
+(defvar spacemacs-layouts-restrict-spc-tab t
+  "if `t' then SPC-tab will respect perspective")
```
duianto:
```suggestion
  "If `t' then `SPC-TAB' will be limited to the current layouts buffers.")
```

duianto:
`spacemacs-layouts-restrict-spc-tab` should probably be set to `nil` by default.

One case where it would be undesired behavior to limit it to a layouts buffers, is if a buffer is opened without adding it to the current layout `SPC l t`.

With the variable set to `t`.
And a "temporary" buffer is opened `SPC l t`.
Now `SPC TAB` switches from the "temporary" buffer to a buffer in the current layout.
And `SPC TAB` again switches to another buffer in the current layout. Or if the current layout only has one buffer then nothing happens when `SPC TAB` is pressed.

### Questions:
#### Q1
Is there a case where `next-buffer` and `previous-buffer` navigates to a buffer in another layout?

For me `SPC b n`, `SPC b p` or `SPC b .` (buffer selection transient state, `n` or `p`) is limited to the current layouts buffers. Both with and without this PR.

#### Q2
I'm also not clear how `spacemacs/window-split-single-column` should interact with other buffers since it only seems to delete the other windows.

With the Emacs frame split into two windows, and the selected window shows a buffer that's not part of the current layout (`SPC l t`).
Now `SPC w 1` removes the other window (which seems to be the expected behavior).

It seems to behave the same with or without this PR.

thanhvg:
@duianto your Q1 and Q2 are valid. I should remove them. The behavior you described when `spacemacs-layouts-restrict-spc-tab` is `t` is exactly what I want.

However I think @smile13241324 is right I should turn off this feature by default. Let users change 
the options when they opt in. Default setting will be as follows in `config.el`:
``` elisp
(defvar spacemacs-layouts-advice-list
  '()
  "list of functions to be wrapped by `with-persp-buffer-list'")

(defvar spacemacs-layouts-restrict-spc-tab nil
  "if `t' then SPC-tab will respect perspective")
```

Let me know what you all think and I'll make the change.

Also thank you all for going through all the silly typos I made, I really appreciated it.

smile13241324:
Ok so far from my side. @duianto could you also have a short look about it?

duianto:
#### readme.org
The functions list and the spc-tab variable should probably be two different sections since they are independent of each other.

#### Backend code
I would say limit the split functions to the current layouts buffers as was initially suggested.

Because the current behavior seems to add random buffers from the `buffer-list` (I'm not sure which order they are added, maybe based on the last buffer added to the list).

It's better to show the scratch buffer if the current layout doesn't have enough buffers to fill the windows.

But the `spacemacs-layouts-restrict-spc-tab` variable should probably be disabled by default, because it's docstring says:
>Switch back and forth between current and last buffer in the
current window.

Otherwise `SPC-TAB` might seem to be broken when nothing happens if the current layout only has one buffer.

When the variable has been enabled manually then it might be easier to realize why nothing happens.

While writing this, I realized that a warning message should be shown when `SPC-TAB` is pressed and the current layout only has one buffer.

The message could say something like:
>Can't switch buffers, because the current layout only has one. Add another buffer to this layout or disable `spacemacs-layouts-restrict-spc-tab` to switch to another layouts buffer.

It's probably to long, but it's just a first draft, any suggestions/rewrites are welcome.

Maybe the variable doesn't have to be mentioned in the message, but it should be described in the `spacemacs/alternate-buffer` docstring.

thanhvg:
Change: 
- default is limit the split functions to the current layouts buffers
- tab switch is off by default, when on will throw a message if no window found

Unexpected change in source code: some files (funcs.el) has indentation changed since i ran `SPC j =` on them.

duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/0e9866efcee9915f3418cec11371d02ee619ced2

Indentation changes are fine if a whole buffer is formatted.

Changes:
- Added an entry to the `spacemacs/alternate-buffer` docstring about the 
`spacemacs-layouts-restrict-spc-tab` variable.
- Renamed `spacemacs-layouts-advice-list` to be more descriptive:
`spacemacs-layouts-restricted-functions`
- Made some corrections/rewrites to Readme.org:
  - Removed:
`They will be wrapped around by persp-mode ~with-persp-buffer-list~.`
the description of how it works can probably be left out (but if it's important then we can add it back).
  - Added a separate section for the `spacemacs-layouts-restrict-spc-tab` variable
- Added some line breaks to keep the lines under 80 chars
- Added entries to changelog.develop

PR #12589, opened 2019-08-04
lazy load lsp-python-ms

yqu212:
Enabling lsp-python-ms on startup will slow down the spacemacs very much since its direct result is the Invalidation of the lazy-loading of python and many other related packages. So it is reasonable to lazy load the lsp-python-ms.

I have tested this change. lsp-python-ms loads correctly both on dump and undumped mode.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2292,6 +2292,7 @@ Other:
 - Added support for 'black' formatter (thanks to Mike Macpherson)
 - Enabled =eldoc= for =anaconda-mode= (thanks to Vikash Balasubramanian)
 - Various fixes for =lsp-python-ms= setup (thanks to Trapez Breen)
+- Added lazy loading of =lsp-python-ms= (thanks to Ying Qu)
 **** Racket
 - Restore smart closing paren behavior in racket-mode (thanks to Don March)
 - Updated racket logo (thanks to Vityou)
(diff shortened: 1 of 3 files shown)


cormacc:
Thanks @yqu212 - I should have thought of this when raising the earlier PR.

Re. declaring the dependency on the lsp layer -- rather than having to explicitly enable lsp in your .spacemacs, you could require it in the python layer by adding a new `layers.el` looking something like this....
```
(when (and (boundp 'python-backend)
           (eq python-backend 'lsp))
  (configuration-layer/declare-layer 'lsp))
```
This is the idiom used in the c/c++ layer and could be considered preferable -- if you're setting python-backend to 'lsp you clearly intend to use the lsp layer, so redundant to have to enable it separately. However that's just my opinion - so maybe wait for @sdwolfz or one of the other maintainers to chime in before giving it any credence :)

yqu212:
Thank you for your response. I have refractored this PR inspired by your
comment. Now, lsp-python-ms loads just after lsp has been loaded in
spacemacs//python-setup-lsp funcion, meaning that checking dependency on lsp
layer is not necessary.

Python layer defaultly asks user to enable lsp layer manually. See function
spacemacs//python-setup-lsp-company. I agree with you that it is redundant to
enable lsp layer separately. It is a good idea to deal with the dependency on
lsp layer automatically. However, it is just a another PR. You can have a try.

cormacc:
Thanks @yqu212 -- @bhipple has beaten me to it (see #12691)

smile13241324:

PR #12590, opened 2019-08-04
Add toggle key docs to docstring and which-key mouseover

duianto:
Added spacemacs|add-toggle key binding documentation strings:
- To the toggle keys function docstring
- And to it's which-key mouseover minibuffer text


smile13241324:
Thank you for contributing to spacemacs :purple_heart:, you can safely remove your branch.

PR #12595, opened 2019-08-05
helm: update helm-xref to set the correct xref-show-xrefs-function

njunxuan:
This pull requests updates the helm-xref `use-package` definition to set the correct `x-ref-show-xrefs-function` based on the emacs version. helm-xref currently provides two functions and the user is expected to use the correct one based on their emacs version (as 27 uses a different handler for xrefs). Due to lazy loading, the `xref-show-xrefs-function` is not set correctly and must be manually handled in the `use-package` definition.

### DIFF
--- a/layers/+completion/helm/packages.el
+++ b/layers/+completion/helm/packages.el
@@ -396,7 +396,7 @@
 
 (defun helm/init-helm-xref ()
   (use-package helm-xref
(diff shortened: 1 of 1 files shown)


duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/6b2f849efdcbfa3f7d0ce937ca054b07a0854776

PR #12594, opened 2019-08-05
Clean .spacemacs.template: Remove dotspacemacs-verbose-loading

yqu212:
dotspacemacs-verbose-loading has been removed in [commit 1623ff6044
](https://github.com/syl20bnr/spacemacs/commit/1623ff60449918298a2c3a5dd0dbd3fd34d0b02b).

PR #12596, opened 2019-08-06
Update CHANGELOG.develop

yqu212:
I forgot to update CHANGELOG.develop in #12594 to reflect the change. Sorry.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -429,6 +429,8 @@ Other:
   - Removed obsolete =:powerline-scale= property from
     =dotspacemacs-default-font= variable
     (thanks to Muneeb Shaikh and Sylvain Benner)
+  - Removed =dotspacemacs-verbose-loading= variable
+    (thanks to Sylvain Benner and Ying Qu)
 *** Core changes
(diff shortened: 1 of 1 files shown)


duianto:
A changelog.develop entry was added to the commit when your PR was cherry picked: https://github.com/syl20bnr/spacemacs/commit/53a0dd85bf001a1a364dd99a22de34b1a58a1fa8

But your correct that it should be in the `Dotfile changes` section, I'll move it there.

yqu212:
@duianto Please use my full name. Thank you.

duianto:
We use the full name from the commits `From:` field, which can be viewed on the second line here:
https://patch-diff.githubusercontent.com/raw/syl20bnr/spacemacs/pull/12596.patch

If you update your gitconfig to show your full name then we won't make the same mistake again.

duianto:
There are two places where it can be set:
- Globally in the `.gitconfig` file in your home directory.
- Locally in the `CONFIG` file in a repositories `.git` directory.

yqu212:
It's my fault. I have set my full name in the repository.

Issue #12598, opened 2019-08-08
Can't install spacemacs and M is undefined


lassiter:
Getting this error from clean install:
```
--> installing package: paradox@spacemacs-ui... [71/90]
Package spinner is unavailable. Is the package name misspelled?

An error occurred while installing paradox (error: (error Package ‘spinner-1.7.3’ is unavailable))
--> installing package: undo-tree@spacemacs-editing... [83/90]
Package undo-tree is unavailable. Is the package name misspelled?
--> installing package: ws-butler@spacemacs-editing... [90/90]
```

Tried following instructions here to install: https://elpa.gnu.org/packages/undo-tree.html

M is undefined in the editor so can't run `M-x package-install RET undo-tree RET`

tuxayo:
Thanks I was just blocked by that :)

So even for the long run, the solution will be to use `emacs --insecure` ?

duianto:
This issue has a workaround that seems to fix installation issues:
Failed to download 'gnu' archive. Bad request. (Solution included) #12535

PR #12599, opened 2019-08-08
Update README.md to inform new users about resolution for failed package installs.

lassiter:
Users like myself run into errors like explained in #12598. This PR is to add a resolution that fixed my error installing packages.


smile13241324:
Thank you for contributing to spacemacs :purple_heart:. I have cherry-picked your branch into develop. You can safely delete your branch.

PR #12199, opened 2019-04-19
Adding Ruby LSP Support (Continuing PR-#11593)

billy1kaplan:
Trying to continue existing work on adding LSP backend to Ruby layer. Influenced heavily by the python lsp backend.

I'm not sure if it's best to start a new PR, but I wasn't sure how to continue in the existing one.

As far as I can tell, the open issues were:

1. Solve the key binding conflict with rubocop.
2. Figure out if documentation can be displayed without needing to execute yard gems.
3. Document it in the README.md.
4. Integrate with rvm.

Based on trying this out, I think #2 is crossed off since the documentation appeared to be working.

PR #12609, opened 2019-08-09
Remove stray ruby file.

mijoharas:
Looks like this was added in https://github.com/syl20bnr/spacemacs/pull/12199
I noticed it as I did a pull, I think it was a test file that was included by accident. 

There is no reference to `test.rb` anywhere, and the ruby code is very generic hello-world-y type code.


duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/3df5b43d28b4ce00c0bde72e0b1d51ab1065a494

PR #12611, opened 2019-08-10
Clojure clj kondo

robertluo:
Add flycheck-clj-kondo which is an optional linter package to clojure layer.

### DIFF
--- a/layers/+lang/clojure/README.org
+++ b/layers/+lang/clojure/README.org
@@ -18,6 +18,8 @@
 - [[#usage][Usage]]
   - [[#cheatsheet][Cheatsheet]]
   - [[#linting][Linting]]
+    - [[#squiggly-linter][Squiggly linter]]
+    - [[#clj-kondo-linter][clj-kondo linter]]
   - [[#structuraly-safe-editing][Structuraly safe editing]]
 - [[#key-bindings][Key bindings]]
   - [[#working-with-clojure-files-barfage-slurpage--more][Working with clojure files (barfage, slurpage & more)]]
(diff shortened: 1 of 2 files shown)


robertluo:
> I have made a short review of your PR and would like to request some changes:
> 
> * squash your commits please, you should only have one
OK.
> * I think the idea of clojure-enable-linters is to activate all available clojure linters, please load all linters when it is on the flycheck checkers should then check whether they can activate or not.
`clojure-enable-linters` sounds like the idea, but actually it binds to squiggly linter, while the linter I proposed to add is of different idea, and I do not think most users want to enable them at the same time. I thought to rename the variable, but considering backward compatibility leave it as is.

> * Please mention in the README whether another binary object must be installed to have the respective checker work and add a short example how to install these i.e. with lein.
I will update the doc to include this, actually follow the link the package itself has detailed documentation, I am not sure if it is good to duplicate it. 

> 
> Please check the requests and mention me when you are ready for another review.

smile13241324:
Hmm ok, I have thought that the linter would complement each others if they don't then I would prefer to have changed the existing activate linter variable to rather provide a choice of which of the two linters to use i.e. clojure-enable-linters with 'squigly 'kondo or nil.

About the more detailed installation instructions, if the package is very detailed about this you can also include a link to the ticket, but then you should mention that detailed installation instructions have to be obtained from there.

kommen:
I was also going to add clj-condo to the clojure layer when I found this PR. Anything I can help getting this merged?

>  and I do not think most users want to enable them at the same time

I agree with @robertluo, most users probably only want one linter.

smile13241324:
@kommen thanks for asking :). Properly yes when you have a look at my last comment I agree to only enable one linter at a time. However if it is this way with clojure I would like to have the linter selectable via a variable rather than have two binary variables only resulting in loading one thing.

Anyway this is just a small issue nothing worth delaying this any further. @kommen  if you could find the time to test this pr and post your results I would be happy to merge it.

smile13241324:
@robertluo please squash your commits for this PR. If everything works I can merge it. The variable refactoring can be done in a separate PR.

kommen:
I've tested this branch it it works fine. Thank you @robertluo!

I do have one request:
It would make it easer if in the readme users would be instructed where to place the configuration to enable clj-kondo. For example in `Pretty Symbols` it is explicitly said it has to go into `dotspacemacs/user-config`

If I understood correctly, the linter config has to go into `dotspacemacs/user-init`, correct?

robertluo:
Thanks @kommen ! I followed the previous documentation of `flycheck-clojure`, only added a similar section for `clj-kondo` value.

IMO teaching users where to put a value of variable is so common that a general document is better than every variable author repeat it hundreds of times.

kommen:
I see, it is very inconsistent across layers. I guess @smile13241324 should make the call. Functionality wise this PR is 👍  from my side. Thanks again!

smile13241324:
Thank you for contributing to spacemacs :purple_heart:. I have cherry picked your commit into spacemacs, you can safely delete your branch.

jr0cket:
@robertluo Thank you for the contribution, its very much appreciated.

There are a few minor issues I suggest dealing with in future PRs

Naming: 
Using `[Clojure]` at the start of the commit message is very helpful way to inform people that this is for the Clojure layer when they are scanning the commit log.  There are lots of changes and layers in Spacemacs, so quick clues like this help alot (and also help your PR get looked at sooner)

CHANGELOG.develop
An entry in the change log should be provided with all PR's (I am surprised the Spacemacs maintainer accepted this PR without one, as this is one reason we haven't had a master release in 18 months)

Documentation
If you need to specify settings to be added, then it should be clear which section of the `.spacemacs` file they are to go in.  Otherwise people will raise issues about something not working when its just incorrectly configured.

Issue #380, opened 2015-01-06
`*` doesn't work in text mode


valrus:
Repro:

Open a text mode buffer, insert text
```
blah
blah
```
Put point on the first `blah` and hit `*`

Result (traceback):
```elisp
Debugger entered--Lisp error: (wrong-type-argument stringp none)
  propertize(none face ahs-plugin-defalt-face)
  ahs-decorated-current-plugin-name()
  ahs-stat()
  (let* ((i 0) (overlay-count (length ahs-overlay-list)) (overlay (format "%s" (nth i ahs-overlay-list))) (current-overlay (format "%s" ahs-current-overlay)) (st (ahs-stat)) (plighter (ahs-current-plugin-prop (quote lighter))) (plugin (format " <%s> " (cond ((string= plighter "HS") "D") ((string= plighter "HSA") "B") ((string= plighter "HSD") "F")))) (propplugin (propertize plugin (quote face) (list (quote :foreground) "#ffffff" (quote :background) (face-attribute (quote ahs-plugin-defalt-face) :foreground))))) (while (not (string= overlay current-overlay)) (setq i (1+ i)) (setq overlay (format "%s" (nth i ahs-overlay-list)))) (let* ((x/y (format "[%s/%s]" (- overlay-count i) overlay-count)) (propx/y (propertize x/y (quote face) ahs-plugin-whole-buffer-face)) (hidden (if (< 0 (- overlay-count (nth 4 st))) "*" "")) (prophidden (propertize hidden (quote face) (quote (:weight bold))))) (echo "%s %s%s press (n/N) to navigate, (e) to edit, (r) to change range or (R) for reset" propplugin propx/y prophidden)))
  spacemacs/auto-highlight-symbol-overlay-map()
  ad-Advice-ahs-forward(#[nil "\300\301\302\"\207" [ahs-select ahs-forward-p t] 3 ("c:/Users/valrus/AppData/Roaming/.emacs.d/elpa/auto-highlight-symbol-20130313.243/auto-highlight-symbol.elc" . 36082) nil])
  apply(ad-Advice-ahs-forward #[nil "\300\301\302\"\207" [ahs-select ahs-forward-p t] 3 ("c:/Users/valrus/AppData/Roaming/.emacs.d/elpa/auto-highlight-symbol-20130313.243/auto-highlight-symbol.elc" . 36082) nil] nil)
  ahs-forward()
[...]

PR #12614, opened 2019-08-11
Fix treemacs-project-toggle (SPC p t)

Trapez:
Seems to be caused by projectile-project-root returning nil.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2581,6 +2581,8 @@ Other:
   ~C-x w 0~, ~SPC 0~ and ~M-0~ call the same command (thanks to duianto, Miciah)
 - Loaded =treemacs-projectile= key binding ~C-c C-p p~ with Treemacs
   (thanks to duianto)
+- Handle ~SPC p t~ when =projectile-project-root= is nil (thanks to Trapez
+  Breen)
 **** Typescript
 - Call tsfmt with extension of current buffer for TSX formatting
   (thanks to Victor Andrée)
--- a/layers/+filetree/treemacs/funcs.el
+++ b/layers/+filetree/treemacs/funcs.el
@@ -14,7 +14,7 @@
   (interactive)
   (if (eq (treemacs-current-visibility) 'visible)
       (delete-window (treemacs-get-local-window))
-    (let ((path (projectile-project-root))
+    (let ((path (projectile-ensure-project (projectile-project-root)))
           (name (projectile-project-name)))
       (unless (treemacs-current-workspace)
         (treemacs--find-workspace))


duianto:
### Issue:
If `SPC p t` is pressed in a non-project buffer, for example: `.spacemacs`, before helm has been loaded (this occurs right after starting Spacemacs, before opening a helm buffer), then this error message is shown:
>user-error: Please install helm from https://github.com/emacs-helm/helm


And after helm has loaded, then `SPC p t` from a non-project buffer opens a helm projectile buffer.
But when a project is chosen, then Treemacs opens with the selected project, but it's name is a dash: `-`


The non-project buffer error message has been discussed and solved upstream:
Add an error message to: do add project to workspace #380
https://github.com/Alexander-Miller/treemacs/issues/380

The new Treemacs function: `treemacs-add-and-display-current-project`
shows an error message if the current buffer isn't part of a project:
>[Treemacs] Not in a project.

### A possible solution
I've been working on a possible solution that uses built-in Treemacs functions:
```elisp
(defun spacemacs/treemacs-project-toggle ()
  "Toggle Treemacs closed/open and add the current project root.
If Treemacs is:
- Open: Close Treemacs.
- Closed: Open Treemacs:
  - If the current buffer isn't part of a project: Show an error message.
  - If the current buffers project root:
    - Exists in Treemacs: Select the projects root.
    - Doesn't exist in Treemacs: Add the project and select the current file."
  (interactive)
  (treemacs-with-toggle (treemacs-add-and-display-current-project)))
```

This requires adding `treemacs-with-toggle` to the `treemacs/init-treemacs` function in `treemacs/packages.el`
```elisp
(defun treemacs/init-treemacs ()
  (use-package treemacs
    :commands (treemacs-select-window
               treemacs-with-toggle
               treemacs--window-number-ten)
...
```

### Possible improvements
I hadn't created a PR for it yet because there's a slight inconvenience with the current `SPC p t` behavior.

If Treemacs is open and one navigates to a file in a project that isn't listed in Treemacs, and one want to add that project to Treemacs, then:
`SPC p t` detects that Treemacs is open and closes Treemacs.
`SPC p t` again opens Treemacs and adds the new project.

It seems like one should be able to add the current project while still keeping Treemacs open.

And with this PR's suggestion of opening Helm Projectile if there's no project, that might be better than just showing an error message.

smile13241324:
Nice fix :+1: , I have cherry-picked your commit into spacemacs, you can safely remove your branch.

duianto:
@smile13241324 I'm guessing that you pulled the PR before I posted the comment. This introduces two issues. They are not bad, but they might cause some confusion.

smile13241324:
Sorry @duianto I have misread your comment, I have thought that the issues has been already solved by the author. As my test worked well I have just merged it.

Thanks for pointing this out, I will try to fix the issues shortly.

duianto:
No problem, at least the helm issue is a side effect of Spacemacs lazy loading helm, so it's not encountered with everyday use.

PR #12616, opened 2019-08-11
update org-layer: helm-org is now a separate package

simonpintarelli:
helm-org has been moved to a separate package. Update org-layer accordingly.


smile13241324:
Thanks for the quick fix :+1:. I have cherry-picked your commit into develop, you can safely remove your branch.

PR #12625, opened 2019-08-16
Switch recreated messages buffer to evil-normal-state

duianto:
If the messages buffer is deleted.
When it's recreated, then it opens in evil-evilified-state
where the evil keys `w` and `b` are undefined.


duianto:
I updated this to only switch to `normal` state if the messages buffer is evilified:
```elisp
    (when (evil-evilified-state-p)
      (evil-normal-state))
```
The initial suggestion was to switch `evil-normal-state` if the editing style is `vim` or `hybrid`, but then it would have tried to switch every time the messages buffer was opened.
```elisp
    (when (memq dotspacemacs-editing-style '(vim hybrid))
      (evil-normal-state))
```

smile13241324:
Thank you for contributing to spacemacs :purple_heart:, cherry-picked into develop.

PR #12634, opened 2019-08-18
[golang] Update README with a new gocode link

signalpillar:
# Problem

The recommended `github.com/nsf/gocode` doesn't work properly and causes error on completion

```
Company: An error occurred in auto-begin
Company: backend company-go error "Wrong type argument: stringp, nil" with args (candidates )
```

It happens because the project is not maintained any more.

> VERY IMPORTANT: this project is not maintained anymore, look for alternatives or forks if you need Go autocompletion tool

> IMPORTANT: consider switching to https://github.com/mdempsky/gocode if you have problems starting with Go version 1.10, due to changes in binary packages architecture (introduction of package cache) I'm not going to adjust gocode for it for quite some time. There is a higher chance that fork under the given link will have some solution to the problem sooner or later.

# Solution
Update documentation to mention a recommended fork of `gocode`.


duianto:
Thanks for the suggestion, but in this case the gocode url has already been updated on the `develop` branch.
https://github.com/syl20bnr/spacemacs/tree/develop/layers/+lang/go#pre-requisites

A note for your next pull request. They should be submitted against the `develop` branch: 

https://github.com/syl20bnr/spacemacs/blob/master/CONTRIBUTING.org#pull-request

signalpillar:
@duianto thank you, for some reason I thought that I successfully switched to `develop`.
Closing it.

PR #12638, opened 2019-08-19
Make spacemacs theme handling support ewal themes (second try)

smile13241324:
Ewal themes customise spacemacs-theme-custom-colors which
changes the appearance of all spacemacs-theme based
themes. To avoid permanent changes on the theme and
still allow custom color themes we need to save
the original setting after having run user-config.
This we then restore everytime we switch a theme.

This restores 5f308b0306d44b20d200ee2bc490540e1c9ee149

### DIFF
--- a/core/core-spacemacs.el
+++ b/core/core-spacemacs.el
@@ -221,6 +221,10 @@ Note: the hooked function is not executed when in dumped mode."
                              "Calling dotfile user config...")
      (dotspacemacs|call-func dotspacemacs/emacs-custom-settings
                              "Calling dotfile Emacs custom settings...")
+     (defconst spacemacs-original-spacemacs-custom-colors (when (boundp 'spacemacs-theme-custom-colors) spacemacs-theme-custom-colors)
+       "Value of spacemacs-theme-custom-colors as it was after running user-config.
+  Needs to be saved as some themes customise this variable and we can otherwise not restore the original setting.")
+
      ;; don't write custom settings into the dotfile before loading them,
      ;; otherwise https://github.com/syl20bnr/spacemacs/issues/10504 happens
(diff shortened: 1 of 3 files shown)


duianto:
There's no variable is void message on startup now 👍 

But an issue still persists, when switching to an `ewal-spacemacs-*` theme in HELM Themes `SPC T s`, and when switching back to `spacemacs-dark` or `spacemacs-light` then the colors from the ewal themes remain, the light theme still has dark colors.

There seems to be a workaround to restore the spacemacs colors, but it might just work when the spacemacs dark and light themes are the only themes in `dotspacemacs-themes`. Then it's possible to cycle through the themes `SPC T n` and the spacemacs themes will appear with the correct colors.

But if for example an ewal theme is in the `dotspacemacs-themes` list, then the ewal colors persist when the themes are cycled.

jjzmajic:
I tweaked the package some to address this issue. Hopefully it makes it easier for you guys. https://github.com/syl20bnr/spacemacs/pull/12330 is now closed though due to reverting the problematic commit. Should I open a new one? The code would be the exact same, as the edits are on my end

smile13241324:
@jjzmajic will I don't think we need another PR, I have seen that the package has been updated on Melpa. I'll make a quick check and see how it goes, maybe we do not need this workaround anymore.

smile13241324:
@jjzmajic looks like there are still some issues with your ewal-spacemacs-* themes after your last changes on Melpa. I have just tried to apply `ewal-spacemacs-modern` with my fix and I get:
```
Error was: (error Invalid face underline :style wave :color nil)
```
After this restoring other themes fail too. Can you have another look?

smile13241324:
Ewal themes support has been rolled back, so there is no need for this fix anymore.

PR #12649, opened 2019-08-22
Update Japanese layer: fix README.org

kenjimyzk:
Because some packages used in Japanese layer are updated,  I mainly fix README.org.


smile13241324:
I have cherry-picked your PR into develop and squashed the commits myself.
Thank you for contributing you can safely remove your branch.

Issue #12636, opened 2019-08-19
Can't use kivy-mode in extra-langs layer


PicoSushi:
I'm new to Kivy, and I'm looking for community-made configuration for kivy, which works well with Spacemacs.

NOTE: There's official kivy-mode implementation.
https://github.com/kivy/kivy/blob/master/kivy/tools/highlight/kivy-mode.el

duianto:
Here's an example of how a recent layer was implemented for the `zig` language.
Maybe you could try to implement something similar.
https://github.com/syl20bnr/spacemacs/commit/a500662a690ad64545f3ec67418735af0b283173#diff-9d38db4925f6b821ab6070a998cb22c1

PicoSushi:
Thanks for comment! I've created kivy layer as your instruction ( https://github.com/PicoSushi/spacemacs/tree/64e7ac90440175de95b698fb60744234c236ed95 ), but I found there's already some implementation about kivy. 
https://github.com/syl20bnr/spacemacs/pull/4649/files

Then, how can I use this layer? I added `extra-langs` to `dotspacemacs-configuration-layers` but `*dotfiles-test-results*` caused **FAIL** on `extra-langs`.

duianto:
The `extra-langs` layer has been removed on the `develop` branch, in this commit:
https://github.com/syl20bnr/spacemacs/commit/28823177097ba59d2b654b5f980746a755d93c66#diff-d82172a9eeeb0ca4a15e5c7d192ffb06

The packages including `kivy-mode` were moved to a layer called `major-mode`:
https://github.com/syl20bnr/spacemacs/tree/develop/layers/%2Blang/major-modes

In this PR:
[BREAKING CHANGE] delete extra-langs layer #10727

The author describes the major-modes layer as:
>it's mostly just major-modes that didn't really belong in a layer..or are too basic...

---
Your PR suggestion doesn't add more functionality (yet), but It's probably better to have a dedicated layer for a language. The layers readme describes what Kivy is and what the layers feature is.

If you post a pull request with a `kivy` layer, then it also should remove every mention of kivy from the `major-modes` layer:

One instance in the README.org:
https://github.com/syl20bnr/spacemacs/blame/5d01efbb6d94fe68a9eed032ac88f647d0247cc6/layers/%2Blang/major-modes/README.org#L22

And two instances in packages.el:
[...]

PR #12657, opened 2019-08-25
Move kivy-mode from major-mode to independent layer

PicoSushi:
Currently, kivy-mode is included in major-mode layer, but this causes installing unnecessary packages.

I'd like to use kivy-mode as a independent layer to use kivy-mode singlealone and create other features in the future, so created this PR.

ref: https://github.com/syl20bnr/spacemacs/issues/12636


duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/57baa4ba4a13c83f6132011ee14d149083c7cea3

PR #12669, opened 2019-08-27
[gnus]: map `gnus-group-get-new-news' to `g r'

hexagonal-sun:
Since the `g` key is remapped in an evilified state for the gnus group buffer, use the combination `g r` to check for new news. This is similarly done in magit.

### DIFF
--- a/layers/+email/gnus/README.org
+++ b/layers/+email/gnus/README.org
@@ -116,6 +116,7 @@ Basic and Spacemacs specific key bindings can be found in the following table.
 | ~SPC a g~   | Starts Gnus                                         |
 | ~m~         | New Message                                         |
 | ~G R~       | Group Buffer - Add RSS feed                         |
+| ~g r~       | Group Buffer - Check for new mail                   |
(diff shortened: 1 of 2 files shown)


duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/f088eb077c30d02544844dbb1eb14fdc10993fb5

I noticed that the `G` prefix didn't work evil state (it does work in Emacs state), so I added it as well.
I haven't used `gnus` before. Could you test that it's working as expected so I didn't miss something.

hexagonal-sun:
> Maybe that key binding should be rebound so that it works in both Emacs and Evil state.

Good idea, I'll post a patch.

duianto:
After thinking about it some more, it's probably better to add major mode key bindings, under the prefixes: `SPC m`, `C-M-m` or `,` (comma).
Then Emacs editing style users won't notice a difference.

Here's an example of how keys are bound in the org layer:
https://github.com/syl20bnr/spacemacs/blob/58d5bb92044673e49a23ba9e69cd635c9ba1c774/layers/%2Bemacs/org/packages.el#L195

PR #12670, opened 2019-08-28
ansible: simplify filename matching re

abitrolly:


abitrolly:
#12671 is against `develop`.

PR #12671, opened 2019-08-28
 ansible: simplify filename matching re

abitrolly:


duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/c5c63f48ca70240d86dacb5953b88053af0b4256

Changes:
`\\` double backslashes were added before the vertical bars and parentheses, because they need to be escaped in Elisp strings. This can be seen in the previous string:
`".*\\(main\.yml\\|site\.yml\\|encrypted\.yml\\|roles/.+\.yml\\|group_vars/.+\\|host_vars/.+\\)")`

It seems like one backslash is enough in for example the minibuffer.
>You need to double-escape ("\\") regexp metacharacters in strings in elisp code, but you single-escape them when entering regexps at the minibuffer.

https://sites.google.com/site/steveyegge2/effective-emacs#item9

That article was linked from:
https://stackoverflow.com/questions/538842/emacs-regular-expressions-in-lisp-need-to-be-double-escaped-why

abitrolly:
Emacs needs to import [raw strings](https://stackoverflow.com/questions/2081640/what-exactly-do-u-and-r-string-flags-do-and-what-are-raw-string-literals) from Python. :D

duianto:
Spacemacs includes https://github.com/joddie/pcre2el
```elisp
(pcre-to-elisp ".*((main|site|encrypted|roles/.+)\.yml|group_vars/.+|host_vars/.+)")
```
output:
```
".*\\(\\(main\\|site\\|encrypted\\|roles/.+\\).yml\\|group_vars/.+\\|host_vars/.+\\)"
```

It matches what we ended up with from this PR:
https://github.com/syl20bnr/spacemacs/blob/1299f262ca0ba24c24d7b8855f1c829c321e243b/layers/%2Btools/ansible/config.el#L21-L22

It even removed an unnecessary `\` backslash in front of `.yml`. I now noticed that Spacemacs colors that backslash orange:

emacs_2019-09-05_08-11-36

When describing the character by placing the cursor on it and pressing `SPC h d c` or calling 
`M-x describe-char`. Then the bottom of the `*Help*` buffer says:
```
There are text properties here:
  face                 (font-lock-warning-face font-lock-string-face)
  fontified            nil
  help-echo            "This \\ has no effect"
```
A `main.yml` file still opens in `ansible-mode` without the slash, so I opened a PR to remove it.
[ansible] Fix filename re, remove no effect backslash #12698

Emacs also comes with `rx.el`:
https://github.com/emacs-mirror/emacs/blob/master/lisp/emacs-lisp/rx.el
```elisp
(rx ".*((main|site|encrypted|roles/.+)\.yml|group_vars/.+|host_vars/.+)")
```
output:
```
"\\.\\*((main|site|encrypted|roles/\\.\\+)\\.yml|group_vars/\\.\\+|host_vars/\\.\\+)"
```
and
```elisp
(rx-to-string ".*((main|site|encrypted|roles/.+)\.yml|group_vars/.+|host_vars/.+)")
```
output:
```
"\\(?:\\.\\*((main|site|encrypted|roles/\\.\\+)\\.yml|group_vars/\\.\\+|host_vars/\\.\\+)\\)"
```

But the output escapes `.` periods and `+` signs, but not `|` vertical bar (pipe) characters.
I tested those strings and the same `main.yml` file as above doesn't open in `ansible-mode`.

I probably don't understand how `rx` works or what they can be used for (Ihaven't read up on it), but it might be useful to know that these commands are available.

duianto:
smile13241324 corrected me in the PR:
[ansible] Fix filename regexp, escape period #12698

That removing the `\` before `.yml` makes it also match `main-yml`.
The correct way to escape a period in elisp regexp is with double backslashes.

PR #12678, opened 2019-08-30
python layer - corrected keybinding collision (format buffer)

cormacc:
`spacemacs/python-format-buffer` was bound to '='
Subsequently clobbered by lsp layer declaration of = as prefix with subsidiary
keybindings.
Resolved by moving `spacemacs/python-format-buffer` binding to '=='

Opportunistically added 'mspyls' dir to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,6 +55,7 @@ eclipse.jdt.ls
 .lsp-session-*
 .trello
 .extension
+mspyls/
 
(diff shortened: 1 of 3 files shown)


smile13241324:
Thank you for contributing to spacemacs :purple_heart:. I have merged your PR, you can safely remove your branch.

PR #12686, opened 2019-09-01
[gnus] add `O` prefix for gnus-group-group-map

hexagonal-sun:
Since the `O` key is not used in the `group-group-mode-map` use it as a prefix
for the `group-group-group-map` in evilifed state.


duianto:
Commit: https://github.com/syl20bnr/spacemacs/commit/4914df0108c6582752f082e5ec4efbcfc04fb1bc

The gnus layers readme was also updated to include the new key binding.
https://github.com/syl20bnr/spacemacs/tree/develop/layers/%2Bemail/gnus

PR #12688, opened 2019-09-02
[clojure] Added support for the Clojure Joker linter to the Clojure layer.

didibus:
The variable clojure-enable-linters can now be set to two additional
options:

1. 'joker

This will pull down flycheck-joker and add Joker as a checker for
Clojure modes.

2. 'clj-kondo+joker

clj-kondo is already a linter option in the layer, but in most setups,
running clj-kondo and joker together make the most sense, as they catch
different errors, and having them both run does not slow down editing.
This options does exactly that, it adds clj-kondo as te primary linter
for clojure modes, but setups joker as the next checker for it.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1219,6 +1219,7 @@ Other:
   - Added =sayid= debugger (thanks to Daniel Manila and Arne Brasseur)
   - Added =flycheck-clojure= linters (thanks to Eugene Yaremenko)
   - Added =clj-kondo= to Clojure linters (thanks to Luo Tian and John Stevenson)
+  - Added =flycheck-joker= to Clojure linters (thanks to didibus)
 - Improvements:
   - Store cider REPL history in spacemacs cache (thanks to Ryan Fowler)
   - Remove backtick from smartparens pairs for Clojure
@@ -1234,6 +1235,7 @@ Other:
     (thanks to Dieter Komendera)
   - Improved spacemacs-jump-handlers (thanks to Ag Ibragimov)
   - Autoscroll to end of repl when sending buffer content (thanks to Vitaly Banchenko)
+  - Added ability to use multiple linters together (thanks to didibus)
 - Key bindings:
   - ~SPC m e ;~ to eval sexp and show result as comment
     (thanks to John Stevenson)
--- a/layers/+lang/clojure/README.org
+++ b/layers/+lang/clojure/README.org
@@ -13,7 +13,9 @@
   - [[#enabling-sayid-or-clj-refactor][Enabling sayid or clj-refactor]]
   - [[#enabling-automatic-linting][Enabling Automatic Linting]]
     - [[#enable-clj-kondo-linter][Enable clj-kondo linter]]
+    - [[#enable-joker-linter][Enable joker linter]]
     - [[#enable-squiggly-linter][Enable Squiggly linter]]
+    - [[#enable-multiple-linters][Enable multiple linters]]
   - [[#starting-clojure-manually-outside-of-emacs][Starting Clojure manually (outside of Emacs)]]
     - [[#quick-start-with-boot][Quick Start with boot]]
     - [[#quick-start-with-lein][Quick Start with lein]]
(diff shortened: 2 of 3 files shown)


didibus:
Failing tests seem unrelated to my change. Not sure they passed prior to it.

mpanarin:
maybe its worth to just make the `clojure-enable-linters` a list of linters to activate?

Because going your way complicates mixing an matching of the linters, imo

didibus:
That's what I had done at first. The issue is I couldn't find a way to generically make it work with flycheck.

But now that I'm saying this, maybe I have an idea to make it work. I'll try and report back.

Also open to suggestions.

didibus:
Alright, I managed to do it. The logic in `clojure/post-init-flycheck` is pretty long for it, but it works well and fast from me trying it out.

Now the user can mix and match linters freely in any order they want, by just specifying a list of them. This can be extended easily in the future if more linters are added as well.

didibus:
Looks like one new test failed, but error seems unrelated to my change again.

didibus:
Alright, turned out I wasn't using next-checkers the way it is supposed too. I fixed that, so now all checkers work without issue. User can chain whatever combination they want in any order.

The only thing I did not handle that I can think off is a specific edge case with flycheck-clojure. Because with my logic, you can only specify one checker per linter, if flycheck-clojure is configured to not have eastwood checker run, then its other checkers won't run either. Not sure what to do about it. Its a small edge case, and I suspect most people will either not use squiggly, or use it with eastwood, or use it by itself.

mpanarin:
Right when I was writing on your `let*` usage :)

This does look fine to me. Code review only, as I am not a Clojure dev or user.

didibus:
@smile13241324 I'm kind of new to the github workflow, so I squashed and had to do a force-push to overwrite history with the squashed commit instead, but it appears that broke the link to see you requested changes. So I didn't get too see them. Was it just the squash, or was there something else?

smile13241324:
@didibus no worries I was just asking for the squash.

didibus:
Sorry for the delay on this, life got in the way.

I have addressed the issue with configuring only a single linter without using a list such as in the reported configuration:

```
(clojure :variables
              clojure-enable-fancify-symbols t
              clojure-enable-sayid t
              clojure-enable-clj-refactor t
              clojure-enable-linters 'squiggly)
```

The other reported configuration is not a valid configuration:

```
(clojure :variables
              clojure-enable-fancify-symbols t
              clojure-enable-sayid t
              clojure-enable-clj-refactor t
              clojure-enable-linters '(clj-kondo, joker))
```

The presence of the comma in the list makes this an incorrect elisp list, and is not a valid configuration.

smile13241324:
Thank you for contributing to spacemacs :+1:, I have cherry picked your PR into develop. You can safely remove your branch.

jr0cket:
Thanks for this change, its very useful to have and means that the unofficial clojure-lint layer is no longer need to use joker.
 
A small point, but I don't fully agree that running clj-kondo and joker together are the recommended approach.  They may catch a few more issues, however, you will have more false positives.  I especially found this with Joker and was the main reason I stopped using it. 

When running both linters you may also experience duplication of warning.  The author of clj-kondo wrote a basic script around joker to turn of some conflicting features to reduce warnings

```
#!/usr/bin/env bash

## turn off some joker features, so I can test clj-kondo better

/usr/local/bin/joker "$@" 2>&1 \
    | grep -v "unused parameter" \
    | grep -v "unused binding" \
    | grep -v "unused namespace" \
    | grep -v "unsupported binding form" \
    | grep -iv "unable to resolve symbol" \
    | grep -v "duplicate require" \
    | sed -e 's/$/ (joker)/'
exit 0
```

I think this should just be elaborated on in the documentation so people are making a more informed choice.  If I get chance this week I will suggest an update in a PR and link it to this PR.

Happy to discuss in the #spacemacs channel of clojurians slack community or in this PR if appropriate.

Thanks again.

didibus:
I definitely would recommend at least having one of clj-kondo or joker on. Anyone trying the Clojure layer without it is doing themselves a disservice.

I don't know that both are necessarily needed. But it is still true that their union provide more coverage. But you will need to double configure them for false positives, as they don't share a configuration.

I think it's a great idea to expand on the trade offs of using them both, versus only using one of them.

`clj-kondo` might be the better one to use if you pick only one, mostly because the project is more active and dedicated to linting. That said, joker has nicer error messages in my experience and can catch some things that I don't know when clj-kondo will, such as minimal type errors related to core functions. 

I also am not a fan of the temporary files left behind by clj-kondo. Joker doesn't leave any trace as such. 

All this could better be explained in the doc I agree.

As for the recommend default, that's hard to say. I choose to use both personally, so I obviously recommend that to others. I think where I feel it is critical is to at least recommend joker or clj-kondo be used as a linter. Honestly, I was tempted to have their binaries be automatically downloaded by Emacs as well, and set the layer to use linting by default 😋

PR #12691, opened 2019-09-03
[python] declare lsp layer dependency when python-backend is 'lsp

bhipple:
Previously, if a user set `python-backend` to `'lsp`, it would not do anything
unless the user also manually added the `lsp` layer to the spacemacs `init.el`
file. This commit updates the `python` layer to correctly add the dependency
when needed, like the `c-c++` layer does.

### DIFF
--- a/layers/+lang/python/layers.el
+++ b/layers/+lang/python/layers.el
@@ -1,0 +1,14 @@
+;;; layers.el --- Python Layer declarations File for Spacemacs
+;;
+;; Copyright (c) 2012-2019 Sylvain Benner & Contributors
+;;
+;; Author: Sylvain Benner <sylvain.benner@gmail.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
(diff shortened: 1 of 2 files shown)


layers/+lang/python/layers.el
```diff
@@ -0,0 +1,14 @@
...
+;; Author: Sylvain Benner <sylvain.benner@gmail.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+(when (and (boundp 'python-backend)
+           (eq python-backend 'lsp))
+  (configuration-layer/declare-layer 'lsp))
```
bhipple: This is essentially the same thing the `c-c++` layer does: conditionally depend on `lsp` when the user has enabled support for it.

https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Blang/c-c%2B%2B/layers.el

smile13241324:
Thank you for contributing to spacemacs :purple_heart:, I have cherry picked your commit to develop. You can safely remove your branch.

PR #12704, opened 2019-09-06
Fix formatting in Latex layer README.org

jonasws:
This PR fixes a typo in the Latex layer README.org file, which caused the preview at http://develop.spacemacs.org/layers/+lang/latex/README.html to be off.


jonasws:
Thanks for the heads up! I figured it had to do with alignment within the `org-mode` syntax definitions, but maybe not. I'll look into having it reported as a bug to GitHub!

PR #12707, opened 2019-09-07
Keybinds for ruby-send-buffer and ruby-send-buffer-and-go

lestercheung:
💓 spacemacs! Thanks for creating and keeping it alive!
Here is the keybinds for ruby-send-buffer and ruby-send-buffer-and-go ;-)

### DIFF
--- a/layers/+lang/ruby/packages.el
+++ b/layers/+lang/ruby/packages.el
@@ -178,6 +178,8 @@
           "sf" 'ruby-send-definition
           "sF" 'ruby-send-definition-and-go
           "si" 'robe-start
+          "sb" 'ruby-send-buffer
+          "sB" 'ruby-send-buffer-and-go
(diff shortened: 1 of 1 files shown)


duianto:
They are already defined three lines above:
https://github.com/syl20bnr/spacemacs/pull/12707/files#diff-a2dff97dda4ab7bcf6d47c892dec213cR176

lestercheung:
> They are already defined three lines above:
> https://github.com/syl20bnr/spacemacs/pull/12707/files#diff-a2dff97dda4ab7bcf6d47c892dec213cR176

Oops how embarrassing! I didn't see that in master (which I use) and didn't see that in develop!

Sorry for wasting your time!

duianto:
It's not a problem at all, "stuff" happens 😄

PR #12708, opened 2019-09-08
update julia-layer

gdkrmr:
Updated julia-layer, more functionality and keybindings.

### DIFF
--- a/layers/+lang/julia/funcs.el
+++ b/layers/+lang/julia/funcs.el
@@ -9,6 +9,24 @@
 ;;
 ;;; License: GPLv3
 
+(defun spacemacs//julia-hash-to-alist (hash)
+  "Convert a `hash-table' to an `alist' for the use in a helm buffer."
+  (let (res)
+    (maphash (lambda (key value)
+               (push `(,key . ,value) res))
+             hash)
+    res))
+
+(when (configuration-layer/layer-used-p 'helm)
+  (defun spacemacs//julia-helm-math-insert()
+    "Insert a utf8 symbol from `julia-latexsubs'"
+    (interactive)
+    (helm :sources (helm-build-sync-source "test"
+                     :candidates (spacemacs//julia-hash-to-alist julia-latexsubs)
+                     :fuzzy-match t
+                     :action (lambda (candidate) (insert candidate)))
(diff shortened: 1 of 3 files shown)


layers/+lang/julia/funcs.el
```diff
@@ -9,6 +9,24 @@
 ;;
 ;;; License: GPLv3
 
+(defun spacemacs//julia-hash-to-alist (hash)
```
gdkrmr: not sure if this is the optimal way to do this, also asked on stackoverflow: https://emacs.stackexchange.com/questions/52552/helm-source-from-a-hash-table

smile13241324:
Thank you for contributing to spacemacs :purple_heart:, I have cherry picked your commit into spacemacs. You can safely remove your branch.

layers/+lang/julia/funcs.el
gdkrmr: no opinions on this?

smile13241324: I am not sure if there is a better way to do this. In any case you will need to read all elements. I have checked the elisp reference manual but there is no specific function. Then I have checked https://github.com/Wilfred/ht.el/blob/master/ht.el which is hash table library for elisp.

They do:
```elisp
(defmacro ht-amap (form table)
  "Anaphoric version of `ht-map'.
For every key-value pair in TABLE, evaluate FORM with the
variables KEY and VALUE bound."
  `(ht-map (lambda (key value) ,form) ,table))

(defun ht->alist (table)
  "Return a list of two-element lists '(key . value) from TABLE.
Note that hash tables are unordered, so this cannot be an exact
inverse of `ht<-alist'.  The following is not guaranteed:
\(let ((data '((a . b) (c . d))))
  (equalp data
          (ht->alist (ht<-alist data))))"
  (ht-amap (cons key value) table))
```
Which is exactly what you have done.

I think it would be a minor improvement to use ht-alist instead which is already loaded in core/libs/ht.el. For this however I would like to ask for a separate PR.

gdkrmr: Thanks for taking a look a this! I may want to edit the function a bit, so that the stuff showed in the helm buffer looks a bit nicer, this will be easier if I keep my own function around.

PR #12710, opened 2019-09-09
Fix hybrid-mode error when dumping (fixes #11923)

### LINKED ISSUE #11923: pdumper crash on setting dotspacemacs-editing-style to 'hybrid in dotfile
- Just follow the normal steps as given in the https://github.com/syl20bnr/spacemacs/blob/develop/EXPERIMENTAL.org file. [You don't need to `git checkout pdumper`]
- Then set `dotspacemacs-editing-style` to `'hybrid` in your dotfile accessible using `SPC f e d`
- Modify layer list as you like so that Spacemacs starts dumping after a restart.
- Restart Spacemacs.
- Let the process in `*spacemacs-dumper*` complete.
- Close spacemacs
- Open again using `emacs-27.0.50 --dump-file ~/.emacs.d/.cache/dumps/spacemacs.pdmp`

This will crash Emacs without even launching.

OS: Antergos (Arch Linux)

madand:
Hybrid mode is a local package in the `spacemacs-bootstrap` layer. Pdumping process exits with error because some other layers (e.g. `auto-completion`) refer to the `hybrid-style-enable-hjkl-bindings`  variable, which is not properly autoloaded. 

In this PR I add autoload cookies to all (just in case) hybrid mode custom settings and adjust its definition to be build by Quelpa.

fixes #11923

### DIFF
--- a/layers/+distributions/spacemacs-bootstrap/local/hybrid-mode/hybrid-mode.el
+++ b/layers/+distributions/spacemacs-bootstrap/local/hybrid-mode/hybrid-mode.el
@@ -31,7 +31,7 @@
 
 (require 'evil)
 
-
+;;;###autoload
 (defcustom hybrid-style-default-state
   (spacemacs|dotspacemacs-backward-compatibility
    hybrid-mode-default-state normal)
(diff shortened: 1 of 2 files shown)


smile13241324:
Thank you for contributing to spacemacs :purple_heart:, I have cherry-picked your commit into develop. You can safely remove your branch.

syl20bnr:
wow, that’s nice. I thought I fixed this with a work-around 2 months ago. @madand you still had a crash ?

madand:
@syl20bnr yeah it still crashed. seems like you have fixed another issue with your workaround, so this one had a chance to pop up :)

Issue #12628, opened 2019-08-16
[Org layer] Error complaining (wrong-type-argument keymapp nil)


a0919610611:
#### Description :octocat:
some error in org layer after update

also have tried reinstalling all packages

error is shown only when enabling org layer

#### Reproduction guide :beetle:
- Start Emacs (emacs --debug-init)
- Error pop up

*Observed behaviour:* :eyes: :broken_heart:

*Expected behaviour:* :heart: :smile:

System Info 💻
OS: mac OSX 10.11.6
Emacs: 27.0
Spacemacs: 0.300.0
Spacemacs branch: develop (rev. 0ca502823)
Graphic display: t
Distribution: spacemacs
Editing style: vim
Completion: helm
Layers: 

#### Backtrace :paw_prints:
```
Debugger entered--Lisp error: (wrong-type-argument keymapp nil)
  define-key(nil [menu-bar Tools Helm] ("Helm" keymap "Helm"))
  easy-menu-get-map(nil ("Tools" "Helm") nil)
  easy-menu-add-item(nil ("Tools" "Helm") ("Org" ["Org headlines in org agenda files" helm-org-agenda-files-headings t] ["Org headlines in buffer" helm-org-in-buffer-headings t]) "Elpa")
  eval-buffer(#<buffer  *load*-423864> nil "/Users/fuyu0425/.emacs.d/elpa/27.0/develop/helm-or..." nil t)  ; Reading at buffer position 487
  load-with-code-conversion("/Users/fuyu0425/.emacs.d/elpa/27.0/develop/helm-or..." "/Users/fuyu0425/.emacs.d/elpa/27.0/develop/helm-or..." nil t)
  load("/Users/fuyu0425/.emacs.d/elpa/27.0/develop/helm-or..." nil t)
  package--activate-autoloads-and-load-path(#s(package-desc :name helm-org :version (20190813 604) :summary "Helm for org headlines and keywords completion" :reqs ((helm (3 3)) (emacs (24 4))) :kind nil :archive nil :dir "/Users/fuyu0425/.emacs.d/elpa/27.0/develop/helm-or..." :extras ((:url . "https://github.com/emacs-helm/helm-org") (:maintainer "Thierry Volpiatto" . "thierry.volpiatto@gmail.com") (:authors ("Thierry Volpiatto" . "thierry.volpiatto@gmail.com")) (:commit . "78b1620884bfe004cc68e2eedc7333b5b9d75b85")) :signed nil))
[...]

PR #12713, opened 2019-09-10
eval (require 'helm-config) after activating helm to define helm keymap

emacs18:
This is to assure that helm keymap is defined to prevent errors while activating helm-org.

Without this change following error can be seen while emacs starts up:

```
Debugger entered--Lisp error: (wrong-type-argument keymapp nil)
  define-key(nil [menu-bar Tools Helm] ("Helm" keymap "Helm"))
  easy-menu-get-map(nil ("Tools" "Helm") nil)
```

This is because helm keymap is not defined when helm is activated.
When helm-org is then activated, then it calls easy-menuj-get-map which fails
due to helm keymap not being defined.

### DIFF
--- a/layers/+completion/helm/packages.el
+++ b/layers/+completion/helm/packages.el
@@ -251,6 +251,11 @@
         (kbd "gr") 'helm-ag--update-save-results
         (kbd "q") 'quit-window))))
 
+(defun helm/pre-init-helm ()
+  ;; This is to define helm keymap which is needed by easy-menu-add-item called
+  ;; within helm-org-autoloads.el while activating helm-org.
+  (require 'helm-config))
+
 (defun helm/init-helm-descbinds ()
   (use-package helm-descbinds
     :defer (spacemacs/defer)


emacs18:
The author of helm said this

```
The right thing to do is (require 'helm-config) which among other things load all the helm autoloads. Doing this is very fast, all the heavy helm libraries are not loaded, only the autoloads.
Once this done, you can use use-package for helm-org or whatever helm files.
```

at https://github.com/emacs-helm/helm-org/issues/5

emacs18:
Well, the change seem to work when I tested it ...

It make take some time for me to look into the test failures.

duianto:
This has been fixed upstream as discussed in the helm-org issue: https://github.com/emacs-helm/helm-org/issues/5

The issue was tracked in the Spacemacs repository here:
[Org layer] Error complaining (wrong-type-argument keymapp nil) #12628

Make sure that your packages are up to date.

emacs18:
Thanks @duianto for your response.  I double checked that all my packages are fully up to date.  The problem I ran into I don't think can be fixed by any change in helm-org.el other than to remove autoload cookie for the easymenu command.

The problem I run into is within configuration-layer//configure-packages-2 where all packages are activated.  When helm-org package is activated only helm-org-autoloads.el is loaded, not helm-org.el.  Hence I don't see how adding lines of code within helm-org.el can impact the problem.  What I'm trying to do with the patch I submitted is to eval (require 'helm-config) prior to loading helm-org-autoloads.el.

duianto:
My understanding of the problem is that `helm-easymenu` wasn't loaded before `easy-menu-add-item` was called.

The fix: https://github.com/emacs-helm/helm-org/commit/542dda7bc9a3b9dfb439e4f8a1e5f60cfb6cc256 included `(require 'helm-easymenu)` in the autoload of 
`easy-menu-add-item` by wrapping both in a `progn`.

That inserted `(require 'helm-easymenu)` into:
`.emacs.d/elpa/26.3/develop/helm-org-20190819.617/helm-org-autoloads.el`

Do you have `(require 'helm-easymenu)` in your `helm-org-autoloads.el`?

emacs18:
I'm sorry.  Thierry made two changes on Aug 19 for helm-org.  For some unknown reason my git clone of helm-org did not pick up the second change!  I need to examine my build script which was supposed to have updated all my git clones.

With this revelation, I of course agree that this issue is fixed.
Thanks again and my apologies.  

PS: I build all packages locally from source rather than using internet package archives.
I've been doing this for years.  Looks like there is a bug in my build script somewhere.

duianto:
No problem.
Could you post an update with the source of your issue in the `helm-org` issue.
The author seems to have been left with the impression that the issue was caused by Spacemacs.

PR #12714, opened 2019-09-10
Add the transient/ directory to gitignore

jnahorny:
While packages are being updated, a directory 'transient' is created, and used as a temp working directory. If those exist, the 'Update Spacemacs' process fails, because `.emacs.d` has local modifications. Adding this directory and it's contents to `.gitignore` fixes that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,6 +56,7 @@ eclipse.jdt.ls
 .trello
 .extension
 mspyls/
+transient/
 
 # Private directory
 private/


jnahorny:
My first PR here. Not sure why 4 tests failed. Could somebody help me here? Thx!

mpanarin:
1st one is optional (CHANGELOG.develop)

others are failing on the upstream as well, so don't worry

smile13241324:
Thank you for contributing to spacemacs :purple_heart: I have cherry-picked your PR into develop.

PR #12287, opened 2019-05-11
spacemacs-layouts: Fix adding buffers to new persp

Miciah:
Delete the hook that https://github.com/syl20bnr/spacemacs/pull/11642 added to `persp-created-functions`, and change `spacemacs/ivy-spacemacs-layouts`, `spacemacs/helm-perspectives`, `spacemacs/helm-persp-switch-project`, and `spacemacs/ivy-persp-switch-project` to achieve the intended goal of adding the desired buffers after creating a new perspective.

Change `spacemacs/helm-persp-switch-project` and `spacemacs/ivy-persp-switch-project` as follows: If the user selects a project but then quits without selecting a file or buffer, the new perspective is now immediately killed.  Otherwise, if the perspective did not already exist, any buffers that belong to the selected project are added to the perspective.

Add the following actions to `spacemacs/ivy-spacemacs-layouts` and `spacemacs/helm-perspectives`:

* Create a new perspective with the Spacemacs home buffer (default action).
* Create a new perspective with the buffers that belong to the current buffer's project.
* Create a new perspective with the buffers that belong to the current perspective (i.e., make a copy of the current perspective).

This commit resolves the problem reported in https://github.com/syl20bnr/spacemacs/commit/9fcf8c898dd1069ef362f1ede320e85065358ac6#commitcomment-33343455.

This commit also fixes a problem with the default action for `spacemacs/ivy-spacemacs-layouts` and `spacemacs/helm-perspectives`.  According to https://github.com/syl20bnr/spacemacs/pull/4077, the default action is supposed to display the home buffer if the action creates a new perspective, which was determined by checking whether the perspective was missing from the list of perspectives _before_ switching.  However, commit eb7ca651fe72241a5784c289543f73b6c273faf4 changed this logic so it was checking whether the perspective was missing from the list of perspectives _after_ switching.  This commit restores the correct logic.

* `CHANGELOG.develop`: Update.
* `layers/+completion/ivy/funcs.el` (`spacemacs/ivy-spacemacs-layouts`): Add "Copy Current Layout" and "Create Project Layout" actions to the docstring.  Use `spacemacs//create-persp-with-home-buffer` for the default action.
* `layers/+completion/ivy/packages.el` (`ivy/post-init-persp-mode`): Add actions for `spacemacs//create-persp-with-current-project-buffers` and `persp-copy`.
[...]

PR #12715, opened 2019-09-10
layout layer: add more persistant actions to switch layout

thanhvg:
Most of the time when I switch project layout (`SPC p l`) I want to see recent file list instead of project file list.

I added two more persistent commands: `helm-projectile-recent` and `spacemacs/helm-project-smart-do-search`. As standard helm persistent actions they are accessible via (`C-z`) or functions keys: `F2` and `F3` when `SPC p l` is invoked.

One concern is I turn on `lexcial-binding` for this file `layers/+spacemacs/spacemacs-layouts/funcs.el`. If necessary I can refactor the code to work without the need of lexcial binding.

Thanks

### DIFF
--- a/layers/+spacemacs/spacemacs-layouts/funcs.el
+++ b/layers/+spacemacs/spacemacs-layouts/funcs.el
@@ -1,4 +1,4 @@
-;;; funcs.el --- Spacemacs Layouts Layer functions File
+;;; funcs.el --- Spacemacs Layouts Layer functions File -*- lexical-binding: t; -*-
 ;;
 ;; Copyright (c) 2012-2018 Sylvain Benner & Contributors
 ;;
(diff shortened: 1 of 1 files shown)


duianto:
I doubt it's more comprehensive, I just tested that the actions work as expected, which they do 👍 

If you have the time and interest, then it would be helpful to also add the actions to the ivy layer.
Here's how a PR added some ivy actions:
https://github.com/syl20bnr/spacemacs/pull/12287/files

`ivy` actions can be accessed by pressing `M-o` at a prompt.

I don't have an informed opinion about `lexical-bindings`. (I just read a wiki page about it)
https://www.emacswiki.org/emacs/DynamicBindingVsLexicalBinding

thanhvg:
Sorry I know nothing about `ivy` so I can't do anything about it and I am not sure if `ivy` users need this feature at all. Maybe a separate PR for `ivy` by someone else is needed.

duianto:
The PR has been cherry picked into the `develop` branch. https://github.com/syl20bnr/spacemacs/commit/d6bbc257d0170c1191eb2ce943a7515a54c2b432

Your right, this PR works so there's no need to complicate it by adding ivy actions. It was just something I thought of while testing.

Some other Spacemacs layers also use lexical-bindings so I picked this.

PR #12721, opened 2019-09-11
[core] Allow text-files to be referenced as startup banner.

fzuellich:
I added a new cond-step to allow a user to reference custom text-banners as startup-banner.

Before it was only possible to reference a custom image file for example:
 `(setq dotspacemacs-startup-banner "~/my-banners/custom.png")`)

With this patch it is possible to reference text files ending on .txt:
`(setq dotspacemacs-startup-banner "~/my-banners/custom.txt")`

I had a look at the `tests` folder, but couldn't find any tests related to loading a banner. If I have overlooked anything, please let me know!

### DIFF
--- a/core/core-spacemacs-buffer.el
+++ b/core/core-spacemacs-buffer.el
@@ -179,6 +179,14 @@ Cate special text banner can de reachable via `998', `cat' or `random*'.
              (spacemacs-buffer/warning (format "could not find banner %s"
                                                dotspacemacs-startup-banner))
              (spacemacs-buffer//get-banner-path 1)))
+          ((and dotspacemacs-startup-banner
+                (string= "txt" (intern (file-name-extension
+                                        dotspacemacs-startup-banner))))
+           (if (file-exists-p dotspacemacs-startup-banner)
+               dotspacemacs-startup-banner
+               (spacemacs-buffer/warning (format "could not find banner %s"
+                                                 dotspacemacs-startup-banner))
+             (spacemacs-buffer//get-banner-path 1)))
           (t (spacemacs-buffer//get-banner-path 1)))))
 
 (defvar spacemacs-buffer--random-banner nil


mpanarin:
I am not sure why it is limited to txt only.

While I understand that other types can be ugly, but that is on shoulder of the end user.
Why not simply allow any files and render images, everything else as plain text?

fzuellich:
Even though it is still possible with this patch (just harder): what happens if the user loads an application file by accident? Why load extensive org-mode content if it's not going to be highlighted? What happens if I load HTML, CSV or bash files? Should a bash script be executed and the result be used as content?

Personally I find the banner, on top of the screen, an unusual position to display information about a roadmap or tasks. To some extent that is even implemented for the homebuffer already. Just like it is for recent files and projects.

Perhaps a better approach would be, to replace the existing Spacemacs homebuffer. This keeps the standard distribution simple and coherent (Spacemacs already supports text banners, just not from a custom path).

I'm not entirely convinced, to display just about any file content as banner. But I also couldn't find any guidelines in the docs, to aid the decision. I would be interested in a few more opinions, especially from a maintainer?! :)

smile13241324:
Not sure about this, technically it would be possible to display literally any file in the home buffer but I wonder whether anybody would load anything different than text or images. Especially I think it would be highly surprising if not dangerous to execute a bash script by loading the home buffer.

I would stick with @mpanarin suggestion to load everything as text except of images. For more flexible solutions I would rather like to have a separate buffer for this in a future PR.

duianto:
There seems to be an issue with how the line that shows the Spacemacs and Emacs version in the top right corner is added.

It overwrites the first line of the text file.

For example with a text file containing:
```
a
b
c
```
The Spacemacs home buffer looks like this:

emacs_2019-11-18_23-49-15

That might be why the banners start with an empty line.
https://github.com/syl20bnr/spacemacs/tree/develop/core/banners

duianto:
There's problems with the error checking.

If the banner ends in `.txt`, but the file doesn't exist, then Spacemacs stops loading after it tries to load the banner.
>Wrong type argument: stringp, (could not find banner c:/Users/username/Temp/test-banne.txt)

And only the Emacs key bindings work.

If the banner doesn't end in `.txt`, ex: `test-banner.tx`,
then there is no warning message that the banner wasn't found. It just loads the default banner:
https://github.com/syl20bnr/spacemacs/blob/develop/core/banners/001-banner.txt

fzuellich:
Hey @duianto,
thanks for your feedback!

> 
> 
> There seems to be an issue with how the line that shows the Spacemacs and Emacs version in the top right corner is added.
> 
> It overwrites the first line of the text file.
> [...]
> That might be why the banners start with an empty line.
> https://github.com/syl20bnr/spacemacs/tree/develop/core/banners

I believe this is an issue that was introduced before. I'm interested in fixing it, though I'm not sure my lisp/emacs knowledge will suffice. Other than that, I believe this fix should go into a seperate commit and PR? What do you think?
 
> 
> There's problems with the error checking.
> 
> If the banner ends in `.txt`, but the file doesn't exist, then Spacemacs stops loading after it tries to load the banner.
> 
> > Wrong type argument: stringp, (could not find banner c:/Users/username/Temp/test-banne.txt)
> 
> And only the Emacs key bindings work.

Fixed it (have to update the PR though).

> If the banner doesn't end in `.txt`, ex: `test-banner.tx`,
> then there is no warning message that the banner wasn't found. It just loads the default banner:
> https://github.com/syl20bnr/spacemacs/blob/develop/core/banners/001-banner.txt

I believe in this case the default behavior is activated, which just loads a hard coded banner. Not sure if this should go in the same PR? Definitely fixable though. :)

duianto:
> I believe this is an issue that was introduced before. I'm interested in fixing it, though I'm not sure my lisp/emacs knowledge will suffice. Other than that, I believe this fix should go into a seperate commit and PR? What do you think?

Your right, the version numbers replacing the first line is unrelated to this PR.
It's better to fix it in a separate PR.

And providing a banner file name that doesn't end in `.txt`, is also unrelated to this PR.
It's expected behavior to open the default case.

fzuellich:
Updated PR to fix issue with referencing non existing txt file. Startup continues without breaking vi keybindings. Instead an error message is displayed and the default banner is rendered.

duianto:
There's some code duplication both in this PR and in the code above that checks for an image banner.

For example this line: `(and dotspacemacs-startup-banner`
checks if the banner is non-nil, even though the functions first line already checked for it:
`(when dotspacemacs-startup-banner` therefore the second check will always be `t`.

And I noticed that if an empty string is provided, then Spacemacs breaks on startup with the message:
>Wrong type argument: stringp, nil

because of this line:
```
                (image-type-available-p (intern (file-name-extension
                                                 dotspacemacs-startup-banner)))
```
Calling `file-name-extension` with an empty string returns `nil`, and calling `intern` with `nil` returns that error message.

So I refactored the function like this:
```elisp
(defun spacemacs-buffer//choose-banner ()
  "Return the full path of a banner based on the dotfile value."
  (when dotspacemacs-startup-banner
    (cond ((eq 'official dotspacemacs-startup-banner)
           (if (and (display-graphic-p) (image-type-available-p 'png))
               spacemacs-banner-official-png
             (spacemacs-buffer//get-banner-path 1)))
          ((eq 'random dotspacemacs-startup-banner)
           (spacemacs-buffer//choose-random-text-banner))
          ((eq 'random* dotspacemacs-startup-banner)
           (spacemacs-buffer//choose-random-text-banner t))
          ((eq 'doge dotspacemacs-startup-banner)
           (spacemacs-buffer//get-banner-path 999))
          ((eq 'doge-inverted dotspacemacs-startup-banner)
           (spacemacs-buffer//get-banner-path 997))
          ((eq 'cat dotspacemacs-startup-banner)
           (spacemacs-buffer//get-banner-path 998))
          ((integerp dotspacemacs-startup-banner)
           (spacemacs-buffer//get-banner-path dotspacemacs-startup-banner))
          ((and
            ;; check that the banner isn't an empty string, because
            ;; calling file-exists-p on an empty string returns: t
            (not (string-empty-p dotspacemacs-startup-banner))
            (file-exists-p dotspacemacs-startup-banner)
            (or (and
                 ;; gui emacs
                 (display-graphic-p)
                 ;; image banner
                 (image-type-available-p
                  (intern (file-name-extension
                           dotspacemacs-startup-banner))))
                ;; txt banner
                (string= "txt" (intern (file-name-extension
                                        dotspacemacs-startup-banner)))))
           dotspacemacs-startup-banner)
          ((string-empty-p dotspacemacs-startup-banner)
           (spacemacs-buffer/warning
            (concat "The banner path is an empty string." "\n"
                    "To remove the banner set dotspacemacs-startup-banner to: nil"))
           (spacemacs-buffer//get-banner-path 1))
          (t (spacemacs-buffer/warning (format "Could not find banner %s"
                                               dotspacemacs-startup-banner))
             (spacemacs-buffer//get-banner-path 1)))))
```

I tested them with these banners:
```
   ;; dotspacemacs-startup-banner 'nil
   ;; dotspacemacs-startup-banner "c:/Users/username/Temp/temp.png"
   ;; dotspacemacs-startup-banner "c:/Users/username/Temp/test-banner.txt"
   ;; dotspacemacs-startup-banner "c:/Users/username/Temp/test-banne.txt"
   ;; dotspacemacs-startup-banner "c:/Users/username/Temp/test-banner.tx"
   ;; dotspacemacs-startup-banner ""
```
And it seems to work, but could you also test if they work for you (change the paths to match your local files). You can toggle commenting on a line with: `SPC ; ;`

If Spacemacs breaks, then you can:
- Open `.spacemacs` with `C-x C-f`
- Navigate to `~/.spacemacs`
- Search `C-s` for `banner`
- Navigate to the next line `C-n`, or to the previous line `C-p`
- Toggle commenting on the current banner line that caused the issue with `C-x C-;`
- And toggle another line that should work, for example `'official` or `nil`.
- Save with `C-x C-s`
- And quit Emacs with `C-x C-c`

You'll have to start Emacs manually, I don't think the `restart-emacs` package is loaded in a case like this. It's probably possible to load the package and restart using it. But it's faster to just close Emacs and start it manually.

fzuellich:
@duianto Thank you for your refactoring! 

**Tests**

I had a look at some test cases. Here is what I found:

* It can't load jpgs (but gifs), no error seems to be produced. Perhaps it's my platform.
* If I define something like `dotspacemacs-startup-banner 't` or `dotspacemacs-startup-banner 'doesntexist` then it crashes on startup (Wrong type argument: stringp, doesntexist)
* `'()` what I understand to be an empty list, doesn't display a banner nor an error message.
* `""` the empty string triggers the display of the message you added
* I created a file 'test.tx'. The default banner is displayed with an error message "Could not find banner". I would expect an error message indicating: file exists, but we wont load it.

**Refactoring**

Just an idea how we might refactor this further: provide a function for the mapping of default banners and a keyword, basically what is this code right now:

```
((eq 'official dotspacemacs-startup-banner)
           (if (and (display-graphic-p) (image-type-available-p 'png))
               spacemacs-banner-official-png
             (spacemacs-buffer//get-banner-path 1)))
          ((eq 'random dotspacemacs-startup-banner)
           (spacemacs-buffer//choose-random-text-banner))
          ((eq 'random* dotspacemacs-startup-banner)
           (spacemacs-buffer//choose-random-text-banner t))
          ((eq 'doge dotspacemacs-startup-banner)
           (spacemacs-buffer//get-banner-path 999))
          ((eq 'doge-inverted dotspacemacs-startup-banner)
           (spacemacs-buffer//get-banner-path 997))
          ((eq 'cat dotspacemacs-startup-banner)
           (spacemacs-buffer//get-banner-path 998))
```
Provide a function to handle integers.

Have one function figuring out user defined files.

Perhaps it's easier to combine three separate functions and provide a fallback in case nothing meaningful is produced? I'm reluctant to add new functions to the file, as I'm not sure about the coding conventions related to the creation of functions.

Let me know what you think.

duianto:
>It can't load jpgs (but gifs)

I'm seeing the same thing:
A `.gif` image works.
But not a `.jpg` image, however I'm seeing a warning message:
>- Could not find banner c:/Users/username/Temp/temp.jpg

It works to copy the `.jpg` and rename it to `.jpeg`.
And I'm able to open the `.jpg` image manually from a dired buffer.

The `image-type-available-p` function might not recognize `.jpg` images.
It's docstring says:
>Return non-nil if image type TYPE is available.
Image types are symbols like ‘xbm’ or ‘jpeg’.

With the startup banners value as:
>dotspacemacs-startup-banner is a variable defined in ‘core-dotspacemacs.el’.
Its value is "c:/Users/username/Temp/temp.jpg"

This returns: `"jpg"`
```elisp
(file-name-extension dotspacemacs-startup-banner)
```

And this returns: `jpg` (a symbol)
```elisp
(intern (file-name-extension dotspacemacs-startup-banner))
```

But this returns: `nil`
```elisp
(image-type-available-p
 (intern (file-name-extension
          dotspacemacs-startup-banner)))
```

The Emacs manual states that:
>Emacs is compiled to recognize JPEG, PNG, GIF, TIFF, and RSVG images also, but displaying these image types require external DLLs which are not bundled with Emacs

https://www.gnu.org/software/emacs/manual/html_node/efaq-w32/Image-support.html

My Emacs `bin` directory has .`dll` files starting with `lib` for:
`png`, `jpeg`, `gif`, but not for `jpg`, so that might have something to do with why the banner isn't shown. Although since `.jpg` files can be opened manually, one would think that they would work as banners as well.

The last line of the system info `SPC h d s` shows:
>- System configuration features: XPM JPEG TIFF GIF PNG RSVG SOUND NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS THREADS LCMS2

---

>If I define something like dotspacemacs-startup-banner 't or dotspacemacs-startup-banner 'doesntexist then it crashes on startup (Wrong type argument: stringp, doesntexist)

Confirmed

---

`'()` is an empty list and it's the same as `nil`, therefore no banner is shown.

---

>I created a file 'test.tx'. The default banner is displayed with an error message "Could not find banner". I would expect an error message indicating: file exists, but we wont load it.

We could check for a supported file format.
And if the file extension isn't supported then we could show a message stating something like:
`'file.ext' is not a supported extension for a banner.`

Maybe with an additional line listing the supported file types.

---

Yeah it sounds great to refactor it more.

It's fine to create any type of function that solves a problem. I don't know if there are any specific conventions, but I can provide any suggestions I notice. Hopefully if someone else sees something, they'll also provide some feedback.

duianto:
I think I found a solution to the first line being removed.
Fix home buffer version and banner #13000

PR #12730, opened 2019-09-14
Add a variable to disable elfeed-goodies

dromaludaire:
By default, the elfeed layer enables elfeed-goodies.

This commit adds a variable that controls this behavior, so that people can easily disable elfeed-goodies if they don't like it.


smile13241324:

bhipple:
Thanks for this contribution!  I was doing the same thing locally and can now cleanup my init.el with the standard feature :+1:

PR #12732, opened 2019-09-17
[clojure] add c-k c-j bindings company dropdown menu in repl

thanhvg:
Make `c-j` and `c-k` bind to down and up in company dropdown menu when it shows up in clojure repl.

Currently users must use tab key to select a completion in company dropdown menu in the clojure repl (`, s s`).


smile13241324:
Thank you for this great fix, I was stumbling over the same issue some time ago but couldn't find the reason for.

I have cherry-picked your commit into develop, you can safely remove your branch.

Issue #12531, opened 2019-07-16
In ivy-switch-buffer, `<C-k>` kills the selected buffer 


demokritos:
#### Description :octocat:
In ivy-switch-buffer, `<C-k>` kills the selected buffer,
while `<C-k>` is supposed to move cursor.

#### Reproduction guide :beetle:
- Start Emacs
- Type `<s-b>` or `M-x ivy-switch-buffer <return>`
- Type `<C-j>` several times
- Type `<C-k>`

*Observed behaviour:* :eyes: :broken_heart:
The selected buffer is killed and disappears in the buffer list.

*Expected behaviour:* :heart: :smile:
The selection cursor should only move up.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 27.0.50
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. e9afae228)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(rust systemd javascript html yaml lsp
      (haskell :variables haskell-completion-backend 'lsp haskell-process-type 'stack-ghci)
      helm
      (auto-completion :variables auto-completion-return-key-behavior nil auto-completion-tab-key-behavior 'complete)
      better-defaults emacs-lisp git markdown treemacs neotree org ibuffer
      (shell :variables shell-default-height 30 shell-default-position 'bottom shell-default-term-shell "/bin/zsh")
      spell-checking syntax-checking unicode-fonts gnus react c-c++ julia racket scheme shell-scripts latex prolog games emoji spotify twitter themes-megapack exwm)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND DBUS GLIB NOTIFY INOTIFY ACL LIBSELINUX GNUTLS LIBXML2 FREETYPE LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XDBE XIM MODULES THREADS XWIDGETS LIBSYSTEMD JSON PDUMPER LCMS2 GMP

#### Backtrace :paw_prints:
```
```

duianto:
Confirmed.

The key binding does load on startup:
https://github.com/syl20bnr/spacemacs/blob/e9afae2285220157c985d97b3c56d3b00cbfb72b/layers/%2Bspacemacs/spacemacs-completion/funcs.el#L199

But in the `SPC b b` (`ivy-switch-buffer`),
[...]

PR #12734, opened 2019-09-17
Fix C-k ivy-previous-line in ivy-switch-buffer

duianto:
It applies to the editing styles:
- hybrid with the variable hybrid-style-enable-hjkl-bindings set to t
- vim

Moved the default key binding for ivy-switch-buffer-kill:
From: C-k
To:   C-M-k

Thanks for the suggestion: asok

---
Fixes: #12531


smile13241324:

Issue #12742, opened 2019-09-18
Python layer attempts to start lsp language server before setting it up


lsp-ableton:
#### Description :octocat:
It appears that the python layer is attempting to start the language server before it's finished setting it up.

#### Reproduction guide :beetle:
- Start Emacs with `python-lsp-server` set to `mspyls`
- Open a python project

*Observed behaviour:* :eyes: :broken_heart:
I see the following in the *Messages* buffer, and the language server doesn't start:
`LSP :: No LSP server for python-mode(check *lsp-log*).~`
`lsp-python-ms: Using version at ‘/home/lane/src/python-language-server/output/bin/Release/’`

The next time I try to open the same project, it works.

*Expected behaviour:* :heart: :smile:

The project opens using mspyls on the first try.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 27.0.50
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. d6bbc257d)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
(ivy ranger better-defaults emacs-lisp
     (c-c++ :variables c-c++-enable-clang-format-on-save t c-c++-backend 'lsp-ccls)
     lsp dap docker nginx semantic lispy
     (multiple-cursors :variables multiple-cursors-backend 'evil-mc)
     javascript react coffeescript clojure html ruby-on-rails
     (python :variables python-backend 'lsp python-lsp-server 'mspyls python-lsp-git-root "~/src/python-language-server")
     restclient yaml sql systemd colors
     (auto-completion :variables auto-completion-enable-snippets-in-popup t)
     git
     (markdown :variables markdown-live-preview-engine 'vmd)
     (shell :variables shell-default-shell 'shell)
     org spell-checking syntax-checking version-control)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG SOUND GPM DBUS GSETTINGS GLIB NOTIFY INOTIFY ACL LIBSELINUX GNUTLS LIBXML2 FREETYPE HARFBUZZ M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XDBE XIM MODULES THREADS XWIDGETS LIBSYSTEMD PDUMPER LCMS2 GMP

lsp-ableton:
Should be fixed in #12743

smile13241324:
[...]

PR #12743, opened 2019-09-18
[python] Don't start the lsp server until after mspyls is optionally required

lsp-ableton:
This addresses [this issue](https://github.com/syl20bnr/spacemacs/issues/12742#issue-495302554) that I ran into. It is a simple change to the python layer to make sure that the correct language server is set up before starting lsp.


smile13241324:

PR #12745, opened 2019-09-19
[treemacs] Turn off treemacs' minor modes when they are configured to be disabled.

Alexander-Miller:
Treemacs' minor modes will activate by default, so when their config variables are nil they must be turned off explicitly.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2684,6 +2684,8 @@ Other:
   (thanks to duianto)
 - Handle ~SPC p t~ when =projectile-project-root= is nil (thanks to Trapez
   Breen)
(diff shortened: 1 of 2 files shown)


Alexander-Miller:
> FAILED  test-stable-elpa-verify-archive--verification-ok

That one shouldn't be on me.

duianto:
Correct, it's unrelated to any PR.
It happens because of the work being done on the verification of the stable elpa repository for the next release.

duianto:
The PR has been cherry picked into the `develop` branch. https://github.com/syl20bnr/spacemacs/commit/1fae31b56df202b3f17aaf5bcdbaa597551f66e5

PR #12748, opened 2019-09-20
[Clojure] fix repl-buffer autoscroll when send forms.

qdzo:
When send form/region/buffer to repl without focus - repl buffer is
not scrolling and result is not visible.

Added scrolling to the end of the repl-buffer after eval.


smile13241324:
Thank you for contributing to spacemacs, :purple_heart:, cherry picked your commit to develop.

Issue #10537, opened 2018-03-29
How to show the context after finding the file and text using (spacemacs/helm-project-do-rg)


zhixing2017:
I often use the command `(spacemacs/helm-project-do-rg)`. 

But there is a little problem. Every time I run the command, I must run `(org－reveal)` to unfold the context.

Is there a simple way to open the context after finding the target?

Thank you.

Here is the information of PC.

Windows 7, 32 bit
GNU Emacs 25.3.1 (i686-w64-mingw32)
Org mode version 9.1.9
Spacemacs 0.300.0

Miciah:
How about the following?
```emacs-lisp
(defun spacemacs/org-reveal (&rest _args)
  (when (derived-mode-p 'org-mode)
    (org-reveal)))
(advice-add 'helm-ag--find-file-action :after #'spacemacs/org-reveal)
```
(I would use `org-mode-hook`, but hooks run before `helm-ag--find-file-action` has positioned point.)

Miciah:
This issue looks like a duplicate of https://github.com/syl20bnr/spacemacs/pull/9529, which has been reported upstream: https://github.com/syohex/emacs-helm-ag/issues/303, with propose fix: https://github.com/syohex/emacs-helm-ag/pull/304.

Meanwhile, I have been using the workaround I suggested in my previous comment.

zhixing2017:
The code you provided works. Thanks very much.

Miciah:
@fleimgruber, I don't understand what you mean by "`spacemacs/org-reveal` is not there anymore".  The workaround defines the `spacemacs/org-reveal` function and then uses `advise-add` to run `spacemacs/org-reveal` after helm-ag jumps to the target.  I have the following in my `dotspacemacs/user-config` definition, and it still works for me:

```emacs-lisp
(use-package org
    :defer t
    :config
    (defun spacemacs/org-reveal (&rest _args)
      (when (derived-mode-p 'org-mode)
        (org-reveal)))
    (advice-add 'helm-ag--find-file-action :after #'spacemacs/org-reveal))
```

fleimgruber:
Well this is embarassing, sorry for the noise, just ignore my comment please. And thanks for the shove.

duianto:
The following PRs changes have been cherry-picked into the develop branch:
Unfold an org line from helm-ag and magit #12750

PR #12750, opened 2019-09-20
Unfold an org line from helm-ag and magit

duianto:
Unfold org headings to a target line when opening a .org file from:
helm-ag, magit diff or blame buffers

---

The advice function and helm-ag advice was initially suggested by: Miciah
https://github.com/syl20bnr/spacemacs/issues/10537#issuecomment-406505520

fixes: #10537

There's an upstream pending PR that fixes the helm-ag issue
Use helm-goto-char to show match and reveal outlines https://github.com/syohex/emacs-helm-ag/pull/304
but the repository was last updated two and a half years ago.

If it's accepted then the helm-ag advice can be removed.

#### Notes
In the github diff it looks like there's three empty lines above the `spacemacs/org-reveal-advice` function, but there's a line feed `^L` character on the middle line.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1594,6 +1594,8 @@ Other:
 - Fixed magit conflicting with golden-ratio (thanks Eric Zhang)
 - Hide with-editor mode-line indicator: =WE= (thanks to duianto)
 - Fixed magit blame in org buffers (thanks to Compro Prasad)
+- Unfolded org headings to a target line when a .org file is opened from a Magit
+  diff or blame buffer (thanks to duianto and Miciah Masters)
 **** Gnus
 - Key bindings:
   - Added ~g r~ for =gnus-group-get-new-news= (thanks to Matthew Leach)
@@ -1749,6 +1751,8 @@ Other:
     (thanks to Codruț Constantin Gușoi)
   - Fixed =spacemacs//helm-open-buffers-in-windows= when opening more files than
     available windows (thanks to Juha Jeronen)
+  - Unfolded org headings to a target line when a .org file is opened from
+    Helm-ag (thanks to duianto and Miciah Masters)
 **** HTML
 - Added impatient-mode under ~SPC m i~ (thanks to geo7)
 - Added =counsel-css= as an =ivy= alternative to =helm-css-scss=
--- a/layers/+emacs/org/funcs.el
+++ b/layers/+emacs/org/funcs.el
@@ -71,3 +71,14 @@
 (defun spacemacs/org-clock-jump-to-current-clock ()
   (interactive)
   (org-clock-jump-to-current-clock))
+
+
+
+(defun spacemacs/org-reveal-advice (&rest _args)
+  "Unfold the org headings for a target line.
+This can be used to advice functions that might open .org files.
+
+For example: To unfold from a magit diff buffer, evaluate the following:
+(advice-add 'magit-diff-visit-file :after #'spacemacs/org-reveal-advice)"
+  (when (derived-mode-p 'org-mode)
+    (org-reveal)))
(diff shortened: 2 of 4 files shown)


duianto:
### To test the issue with opening an org file from `helm-ag`:
- `SPC s f` choose the path to `~/.emacs.d`
- search for: `thus`
- find and open `RET` the candidate (it was the second result for me):
```
.\COMMUNITY.org:77:- Pull request are also issues, thus issue guidelines also apply to pull
```
#### Before this PR:
If the `COMMUNITY.org` file was unfolded, then the cursor appears on the expected word `thus`.
Fold the org file by pressing `S-TAB`, then repeat the steps above to see the issue that this PR solves.

Now the cursor is on the first of the three dots at the end of the heading:
```
* Moderation...
```
Pressing `TAB` to try and unfold the heading, only one word appears:
```
* Moderation...
  requests....
```
Moving the cursor to the left once doesn't seem to do anything, but moving it left a second time moves the cursor to the second `o` in `Moderation...`
Pressing `TAB` now unfolds the `Moderation` section, but its sub sections are still folded.

The word `thus` is in a sub section so it still requires additional unfolding or searching for the target word.

#### After this PR:
Only the headings for the target word `thus` are unfolded and the cursor is at the start of the target word `thus`.


### To test the issue with opening an org file from `magit`:
- `SPC g s` choose the path to `~/.emacs.d`
- open the magit-log: `ll` (l = lower case L)
- move the cursor to the commit:
```
cf15de9f2 [java] Make lsp the default backend and drop eclim support
```
- open it: `RET`
- move the cursor to the line:
```
+  - [[#lsp-java][LSP Java]]
```
- press: `RET` to open the .org file

#### Before this PR:
The cursor is on the first of the three dots at the end of the line:
```
* Table of Contents                     :TOC_5_gh:noexport:...
```

Pressing `TAB` unfolds strangely:
```
* Table of Contents                     :TOC_5_gh:noexport:...  - Lsp Java
   - Installation
```

#### After this PR:
The `Table of Contents` heading is unfolded and the cursor is on the line:
```
  - [[#lsp-java][LSP Java]]
```

duianto:
#### Update:
In the Magit test above:
Pressing `RET` which calls `magit-diff-visit-file`, opens the file in a read-only buffer. (the commit hash is surrounded by `~` and it's appended to the buffer name).
It didn't unfold to the target line without this PR, but it unfolds correctly with this PR.

But in the test above when `C-RET` is pressed which calls: `magit-diff-visit-worktree-file`, then the target file opens normally, meaning that it's editable.
`C-RET` unfolded to the correct line both with and without this PR. 

But I discovered an instance where `C-RET` didn't unfold all the way to the target line, every time. Possibly because the line is under three levels of sub headings.

The line in question is in the commit:
`1646a246d * Update DOCUMENTATION.org`

the line:
`  - Updated DOCUMENTATION.org to fix the example for how to change the separator style`

I updated the PR to also advice the `magit-diff-visit-worktree-file` command and now the org headings seems to unfold correctly every time when `C-RET` is pressed.

duianto:
I found another case where an org file opens folded.

When jumping backwards in a Magit blame buffer: `SPC g b` .

### Reproduction steps
- open the file: `SPC f f` `~/.emacs.d/CONTRIBUTING.org`
- search for: `/` `write commit` `RET`
- open a Magit blame buffer: `SPC g b`
- move backwards in history by pressing: `b`

#### Before
The cursor is on the first period of the folded heading:
`Contributing code...`

#### After
The headings are unfolded:
```
Contributing code
General contribution guidelines
Commit messages
```
and the cursor is at the beginning of the line starting with `Write commit messages...`

The PR has been updated to also handle opening `.org` files from a Magit blame buffer.

smile13241324:
Merged :)

PR #12751, opened 2019-09-21
Fixes #12747 - Org mode supports commit browsing through magit blame

### LINKED ISSUE #12747: RET doesn't show the current commit when blaming a .org file
#### Description :octocat:
`RET` doesn't open the current commit (`magit-show-commit`) when blaming `SPC g b` a `.org` file.
Instead `RET` calls: `org-open-at-point`.

- `SPC f f`
- choose an org file, for example in the Spacemacs repository:  `/.emacs.d/community.org`
- `SPC g b`
- `SPC h d k RET` shows:
>RET (translated from <return>) runs the command org-open-at-point, which is an
interactive compiled Lisp function in ‘org.el’.

It works as expected in non org files, for example:
- `SPC f f` `/.emacs.d/.gitignore`
- `SPC g b`
- `SPC h d k RET` shows:
>RET (translated from <return>) runs the command magit-show-commit (found in
magit-blame-read-only-mode-map), which is an interactive autoloaded compiled
Lisp function in ‘magit-diff.el’.

#### System Info :computer:
- OS: windows-nt
- Emacs: 26.3
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 1fae31b56)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(autohotkey emacs-lisp git helm multiple-cursors org spell-checking syntax-checking treemacs version-control)
```

Compro-Prasad:
See https://github.com/syl20bnr/spacemacs/issues/12747 for further details.

### DIFF
--- a/layers/+source-control/git/packages.el
+++ b/layers/+source-control/git/packages.el
@@ -223,7 +223,9 @@ Press [_b_] again to blame further in the history, [_q_] to go up or quit."
       (when git-magit-status-fullscreen
         (setq magit-display-buffer-function
               'magit-display-buffer-fullframe-status-v1))
-      (spacemacs|hide-lighter with-editor-mode))))
+      (spacemacs|hide-lighter with-editor-mode)
+      ;; Workaround for #12747 - org-mode
+      (evil-define-key 'normal magit-blame-read-only-mode-map (kbd "RET") 'magit-show-commit))))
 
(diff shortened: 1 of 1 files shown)


smile13241324:

PR #12752, opened 2019-09-21
Do not use tabs in inferior python mode

tsoernes:
The problem with using tabs is that if you paste code from anywhere, which is likely to use
spaces which is the PEP standard, and then modify that pasted code, you'll
end up with a mix of tabs and spaces, which the python interpreter throws
an error at.

Also:
Add keybindings for inferior python mode (clearing the repl; restarting the repl; changing virtualenv)


tsoernes:
These two commit should probably be in different PRs, but if they are both good then no bother

smile13241324:

PR #12755, opened 2019-09-22
[python] fix IPython shell setup on windows path with whitespaces

nmdanny:
On systems where IPython executable path contains whitespace
(e.g, if IPython is under Local Appdata and your Windows username has
whitespace), the shell command was invalid, quoting the executable is required.


smile13241324:
Thank you for contributing to spacemacs, I have cherry picked your commit into develop.

PR #12758, opened 2019-09-22
Lsp-scala is deprecated in favor of lsp itself

kryptt:
The lsp-scala package is no longer required and has been moved into lsp proper, we should reflect this change.


smile13241324:

PR #12764, opened 2019-09-24
[gitignore] ignode bbdb file

hexagonal-sun:
Ignore the BBDB file as by default it is stored in ~/.emacs.d.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,6 +57,7 @@ eclipse.jdt.ls
 .extension
 mspyls/
 transient/
+bbdb
 
 # Private directory
 private/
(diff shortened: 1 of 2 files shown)


Miciah:
Would it be better to add something like `(setq-default bbdb-file (expand-file-name "bbdb" spacemacs-private-directory))` to `layers/+spacemacs/spacemacs-defaults/config.el`?

smile13241324:

Not sure if it is feasible to add a default for `bbdb-file` in `spacemacs-default` as the package is not part of any layer yet. For now I would run with the gitignore change, we can still change it if this package becomes part of an official layer.

PR #12768, opened 2019-09-24
[clojure] bind c-ret to cider-repl-newline-and-indent in repl

thanhvg:
For evil insert mode

### DIFF
--- a/layers/+lang/clojure/packages.el
+++ b/layers/+lang/clojure/packages.el
@@ -235,6 +235,9 @@
           (kbd "C-j") 'cider-repl-next-input
           (kbd "C-k") 'cider-repl-previous-input))
 
+      (evil-define-key 'insert cider-repl-mode-map
+        (kbd "<C-return>") 'cider-repl-newline-and-indent)
+
       (when clojure-enable-fancify-symbols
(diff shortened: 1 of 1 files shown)


didibus:
What's the current behavior of C-<return> and <return> ?

didibus:
Hum, I'm wondering, in my own config, I have it that RET is always `cider-repl-newline-and-indent` and C-RET is `cider-repl-return`.

I guess keeping RET intact, and just adding C-RET might be the less intrusive change. I'm just wondering what defaults are best here, since REPL is something people use all the time.

thanhvg:
Default binding for cider in Spacemacs is
````
(kbd "RET") 'cider-repl-return)
 ````
`c-ret` is unbound. 

Your own custom config is the reverse.

duianto:
The PRs changes have been cherry picked into the `develop` branch. https://github.com/syl20bnr/spacemacs/commit/18715b872bd5574f12feca11cd9af04a6652a4ec

didibus:
Ya, I think the default you chose makes more sense as a default for everyone. I would probably had applied it to the `cider-repl-mode-map` as well, so its consistent for Holy mode users as well.

PR #12782, opened 2019-10-01
[gnus] add gnus perspective

hexagonal-sun:
Configure the gnus layer to add a gnus perspective, similar to the erc layout.

### DIFF
--- a/layers/+email/gnus/config.el
+++ b/layers/+email/gnus/config.el
@@ -1,0 +1,16 @@
+;;; config.el --- gnus Layer configuration File for Spacemacs
+;;
+;; Copyright (c) 2019 Matthew Leach & Contributors
+;;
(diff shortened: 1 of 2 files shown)


duianto:
The PRs changes have been cherry picked into the `develop` branch. https://github.com/syl20bnr/spacemacs/commit/d3cb959e36580a11cf2b17c47ba72886668fecd0

#### Modification

- Changed the docstrings from `micro-state` to `transient state`, because they have been renamed in `0.200.0 (2016/10/02)`:
>Micro states are now called transient states, the macro spacemacs|define-micro-state is deprecated, it is replaced by the new hydra powered macro spacemacs|define-transient-state.

https://github.com/syl20bnr/spacemacs/blob/develop/CHANGELOG.org#important-breaking-changes

PR #12792, opened 2019-10-04
[auto-completion] List feature pkg company-quickhelp

duianto:
The company-quickhelp package wasn't mentioned in the auto-completion readme.

And the variable name auto-completion-enable-help-tooltip that enables the
package doesn't mention the package name either (which is fine).

This just makes it somewhat easier to see/find that the company-quickhelp
package is related to tooltips and that it's part of the auto-completion layer.


smile13241324:
Merged :)

PR #12795, opened 2019-10-05
Bound counsel-git to gff and move magit-find-file to gfF.

xuhdev:
counsel-git find a file in the current git repository. "gff" is
currently bound to magit-find-file. While counsel-git seemingly performs the
same functionality as magit-find-file, they are substantially different:
magit-find-file is not intended to find a file, contrary to what the
name indicates. See magit/magit#3967 for
explanation. This is why we need another key binding for counsel-git.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1577,6 +1577,7 @@ Other:
   - Get git permalink for a region (thanks to Roman Gonzalez):
     - ~SPC g l p~ opens browser with region permalink URL
     - ~SPC g l P~ adds region permalink URL to clipboard
+  - Moved =magit-find-file= from ~SPC g f f~ to ~SPC g f F~ (thanks to Hong Xu)
 - Added feature to toggle =evil-magit= based on editing style
   (thanks to Muneeb Shaikh)
 - Added =use-package= for deferred loading of evil-magit
@@ -1794,6 +1795,7 @@ Other:
   - Added ~SPC h d F~ to list faces (thanks to Muneeb Shaikh)
   - Bind =find-file-other-window= to ~j~ (thanks to James Wang)
   - Added =counsel-mark-ring= to ~rm~ (thanks to bmag)
+  - Added =counsel-git= to ~SPC g f f~ (thanks to Hong Xu)
 - Limit =ripgrep= results to 150 columns (thanks to Aaron Jensen)
 - Use new API to register additional transient state bindings
   (thanks to Andriy Kmit')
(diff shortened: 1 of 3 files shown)


xuhdev:
@duianto

duianto:
If `magit-find-file` isn't intended for finding files. Is there an equivalent command for opening a file in another branch?

I've found it useful, for example when looking at a `master` branch file while the `develop` branch is checked out. Maybe there's another way to accomplish the same thing?

I don't think it's to late to swap them. More frequently used commands should require fewer key presses.

Opening a file in a branch seems like an alternate behavior, therefore it's probably better to have it under `SPC g f F`.

It would also be easier to remember and expect the following behavior.
- `SPC f f` finds a file in the current directory.
- `SPC g f f` finds a file in the current project (git/version control).

Could you add entries in `changelog.develop`. At the end of the `Key bindings:` sub section in 
`Layer changes and fixes/ivy`.
https://github.com/syl20bnr/spacemacs/blob/develop/CHANGELOG.develop#ivy

It could say something like this:
```
  - Moved =magit-find-file= from ~SPC g f f~ to ~SPC g f F~ (thanks to Hong Xu)
  - Added =counsel-git= to ~SPC g f f~ (thanks to Hong Xu)
```

And rewrite the outdated commit title and message.

xuhdev:
`magit-find-file` is intended for opening a file in another branch, but this will be read-only and is very different from the expectation of the regular `find-file`. It is indeed useful, but I think it's less useful than a `counsel-git`, which limits the scope to the working directory but confers full editing power. I also agree with you on the easiness to remember the key bindings.

I will make the suggested changes.

xuhdev:
@duianto Updated

duianto:
The PRs changes have been cherry picked into the `develop` branch. https://github.com/syl20bnr/spacemacs/commit/e37b70cfa3e5e6bee57bba0db9fe053e5b47b4c0

Nice catch. Your right the magit-find-file changelog entry goes in the git section.

Miciah:
I use <kbd>SPC g f f</kbd> (`magit-find-file`) to look at a file from another branch, and I use <kbd>SPC p f</kbd> (`helm-projectile-find-file` / `counsel-projectile-find-file`) to find a file from the current Git repository.  Does `counsel-git` provide any benefit over `projectile-find-file`?

xuhdev:
@Miciah `counsel-projectle-find-file` seems to find any file in in the project, while `counsel-git` only finds files indexed by git.

PR #12798, opened 2019-10-05
[ivy] Add counsel colors and faces key bindings

duianto:
`SPC C e` calls `counsel-colors-emacs`
`SPC C f` calls `counsel-faces`
`SPC C w` calls `counsel-colors-web`


smile13241324:
Merged :)

duianto:
Your right we should document them.

The equivalent in helm is `helm-colors` which is bound to `SPC C l` (l = lower case L).
It has two sections:
- Colors, with different color names for the same color and the hex value
- Faces

`l` doesn't seem very mnemonic for colors. But I can't come up with a better one at the moment.
`SPC C c` is taken by `org-capture`.

I'll add them their own completion layer readme.

duianto:
I opened a PR:
Document helm and ivy color key bindings #12850

PR #12802, opened 2019-10-07
shell: Tell vterm to respect `shell-default-term-shell` setting

br3athein:
Effectively, this forces `vterm` to spawn a shell set in a `shell-default-term-shell` variable i/o default value from "shell-file-name" - just like other shells from `shell-pop` do.


smile13241324:

PR #12809, opened 2019-10-09
Make react layer respect js fmt on save

tgroshon:
The react layer relies on the javascript layer and uses multiple values
from it including definition of a formatter. However, the react layer
does not respect the javascript layer setting for formatting a file on
save.

This patch configures the react layer to add a before save hook that
calls the javascript layer formatting function when the
`javascript-fmt-on-save` variable is enabled.


duianto:
The PRs changes have been cherry picked into the `develop` branch. https://github.com/syl20bnr/spacemacs/commit/59c95e7da222be701200087ca221b86813d832cf

PR #12810, opened 2019-10-09
[javascript] Document layer var: javascript-fmt-on-save

duianto:
Also mention prettier as the other formatter in the features section.
And wrap some long lines.


smile13241324:
Merged :)

PR #12817, opened 2019-10-11
Python jump to definitions

rakanalh:
Hello,

When trying to jump to definition, i was being asked for the "TAGS" file, though my virtualenv was active. I looked at the configuration and the line added was missing.

### DIFF
--- a/layers/+lang/python/packages.el
+++ b/layers/+lang/python/packages.el
@@ -58,7 +58,8 @@
         "hh" 'anaconda-mode-show-doc
         "ga" 'anaconda-mode-find-assignments
         "gb" 'xref-pop-marker-stack
-        "gu" 'anaconda-mode-find-references)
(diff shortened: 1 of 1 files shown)


duianto:
The PRs changes have been cherry picked into the `develop` branch. https://github.com/syl20bnr/spacemacs/commit/9bafd3f56193edd0ef7babee2f3bf73a50c8e6ae

duianto:
This PR was reverted by the following PR:
Fix python backend setup #12819
>anaconda-mode-find-definitions is added to spacemacs-jump-handlers-python-mode later when configuring the package, which is then used by jump-to-definition, which is bound to exactly same shortcut SPC m g g .

That PR also fixed the python-backend checking.
Can you check the latest develop branch, maybe your issue was related to that.

rakanalh:
Tried latest develop, it works.

Thanks

PR #12819, opened 2019-10-12
Fix python backend setup

rakyi:
This reverts commit 9bafd3f. `anaconda-mode-find-definitions` is added to `spacemacs-jump-handlers-python-mode` later when configuring the package, which is then used by `jump-to-definition`, which is bound to exactly same shortcut `SPC m g g` .

ac30247811d28e86c4b33d6150d678d5f5824330 broke the backend setup in case when `python-backend` was bound to `nil` since there were still some places that checked the variable directly.

I also moved the leader keys setup from `:init` to `:config` because e.g. `anaconda-mode-find-references` was being overridden to `helm-gtags-update-tags`.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2401,8 +2401,6 @@ Other:
 - Fix lazy loading of =lsp-python-ms= (thanks to lsp-ableton)
 - Fix =Ipython= path on windows (thanks to Daniel K)
 - Make =inferior-python-mode= do not use tabs (thanks to tsoernes)
-- Added =anaconda-mode-find-definition= for ~SPC m g g~
-  (thanks to Rakan Alhneiti)
 **** Racket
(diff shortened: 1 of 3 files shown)


rakyi:
I suppose another option to fix the key bind clash would be to bind `anaconda-mode-find-references` to e.g. `SPC m g U` instead of `SPC m g u`.

duianto:
https://github.com/syl20bnr/spacemacs/commit/c95534ace5a98a53460b620f1c5c8b61cfabbc18
https://github.com/syl20bnr/spacemacs/commit/1fec4ec135382e51da94991fc92406a92d7c150e

The issue the author of the reverted PR: Python jump to definitions #12817
was trying to solve might have been related to the broken python-backend checking.
I'll ask him to check if this solved that issue.

PR #12825, opened 2019-10-13
[shell] Simplify loading logic for eshell-z

Miciah:
* `layers/+tools/shell/packages.el` (`shell/init-eshell-z`): Replace use of `with-eval-after-load` by using use-package's `:after` keyword to load eshell-z when eshell is loaded.


duianto:
The PRs changes have been cherry picked into the `develop` branch. https://github.com/syl20bnr/spacemacs/commit/7db8815ee595c20fa80f61041ae1e52b65e0de9a

PR #12827, opened 2019-10-13
Update references to macOS

Miciah:
Apple renamed "Mac OS X" to "OS X" in 2012 and then to "macOS" in 2016.  Update references to use the current name.

#### Various documentation copy-edits

Follow up the changes in the previous commit with some minor improvements to formatting, grammar, spelling, and wording.

### DIFF
--- a/EXPERIMENTAL.org
+++ b/EXPERIMENTAL.org
@@ -7,7 +7,7 @@
     - [[#clone-emacs][Clone Emacs]]
     - [[#simple-patch-of-emacs-source-code][Simple patch of Emacs source code]]
     - [[#compile-emacs-from-source][Compile Emacs from source]]
-      - [[#macos][MacOS]]
+      - [[#macos][macOS]]
       - [[#ubuntu][Ubuntu]]
       - [[#windows][Windows]]
     - [[#install-emacs][Install Emacs]]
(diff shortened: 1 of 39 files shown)


duianto:
https://github.com/syl20bnr/spacemacs/commit/552fd5953caf9826648429d26b31d6fcf17db341, https://github.com/syl20bnr/spacemacs/commit/97ba5561429836078fde5eb2db8d200b42e14504

A minor correction was applied to the Treemacs key binding table. The change from `Mac` to `macOS`, only increased that lines length by two characters, therefore `TAB` needed to be pressed with the cursor in the table to realign the whole tables right edge.

Miciah:
> A minor correction was applied to the Treemacs key binding table. The change from `Mac` to `macOS`, only increased that lines length by two characters, therefore `TAB` needed to be pressed with the cursor in the table to realign the whole tables right edge.

I wanted to keep the diff small, and I thought @emacspace would take care of reformatting tables.  Is that not so?  There is also a table in the racket layer that will need to be adjusted.

duianto:
You were way ahead of me 😄
I missed the one in the racket layer.

Hmm, the doc bot usually opens a new PR shortly after a commit is merged and it contains something the that the bot can fix.

I'll ask if the bot can fix key binding table alignments or if it's down at the moment.

PR #12830, opened 2019-10-14
Add conda layer (to work with anaconda environments)

zjp:
This is a simple commit based on the existing ipython-notebook layer that provides a way to interact with conda.el through Spacemacs's UI

### DIFF
--- a/layers/+lang/conda/README.org
+++ b/layers/+lang/conda/README.org
@@ -1,0 +1,22 @@
+#+TITLE: Conda Layer
+
+#+TAGS: layer|programming|util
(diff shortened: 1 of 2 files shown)


smile13241324:

By the way please notice that I had to revise your layer to make it spacemacs conform, please try to take care of below points for your next commit:
- Your README was missing a table of content and feature block
- I have changed your bindings to rather extent the bindings of python-layer than to bind them as mode independent bindings modelled after the `django` layer.

zjp:
Thanks!

PR #12844, opened 2019-10-18
tags-table-mode should be in spacemacs-large-file-modes-list

xuhdev:
It is light-weight and it is common that tags are large.


duianto:
The PRs changes have been cherry picked into the `develop` branch. https://github.com/syl20bnr/spacemacs/commit/301d867affa63792b65a09644f1fb8ce9cb4ea62

PR #12845, opened 2019-10-18
Fix C-k in company with tooltip

### LINKED ISSUE #2974: Completion: C-k not working consistently
`C-j` and `C-k` are bound to select completion candidates. But after the company quickhelp window popped up `C-k` is no longer working correctly. It is only producing `^K` in the buffer. 

Disabling `company-quickhelp` is a partial workaround but the same problem happens after pressing `C-d` to get to the documentation.

https://github.com/expez/company-quickhelp/issues/17 suggested a possible solution of putting the keybindings into the `company-completion-finished/cancelled/started-hook`'s but this did not work for me.

duianto:
Got the idea for the fix from this comment:
https://github.com/expez/company-quickhelp/issues/17#issuecomment-526783169

I simplified it to only un/define C-k, since C-j worked without the fix.

Fixes #2974

### DIFF
--- a/layers/+completion/auto-completion/funcs.el
+++ b/layers/+completion/auto-completion/funcs.el
@@ -294,8 +294,14 @@ MODE parameter must match the :modes values used in the call to
       (define-key map (kbd "C-j") 'company-select-next)
       (define-key map (kbd "C-k") 'company-select-previous)
       (define-key map (kbd "C-l") 'company-complete-selection))
-    (when (require 'company-quickhelp nil 'noerror)
-      (evil-define-key 'insert company-quickhelp-mode-map (kbd "C-k") 'company-select-previous)))
+    ;; Fix company-quickhelp Evil C-k
+    (defun spacemacs//set-C-k-company-select-previous (&rest args)
+      (define-key evil-insert-state-map (kbd "C-k") 'company-select-previous))
+    (defun spacemacs//restore-C-k-evil-insert-digraph (&rest args)
+      (define-key evil-insert-state-map (kbd "C-k") 'evil-insert-digraph))
+    (add-hook 'company-completion-started-hook 'spacemacs//set-C-k-company-select-previous)
+    (add-hook 'company-completion-finished-hook 'spacemacs//restore-C-k-evil-insert-digraph)
+    (add-hook 'company-completion-cancelled-hook 'spacemacs//restore-C-k-evil-insert-digraph))
    (t
     (let ((map company-active-map))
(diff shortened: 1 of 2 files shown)


smile13241324:
Merged :+1:

layers/+completion/auto-completion/funcs.el
```diff
@@ -294,8 +294,14 @@ MODE parameter must match the :modes values used in the call to
       (define-key map (kbd "C-j") 'company-select-next)
       (define-key map (kbd "C-k") 'company-select-previous)
       (define-key map (kbd "C-l") 'company-complete-selection))
-    (when (require 'company-quickhelp nil 'noerror)
-      (evil-define-key 'insert company-quickhelp-mode-map (kbd "C-k") 'company-select-previous)))
+    ;; Fix company-quickhelp Evil C-k
+    (defun spacemacs//set-C-k-company-select-previous (&rest args)
+      (define-key evil-insert-state-map (kbd "C-k") 'company-select-previous))
+    (defun spacemacs//restore-C-k-evil-insert-digraph (&rest args)
+      (define-key evil-insert-state-map (kbd "C-k") 'evil-insert-digraph))
```
yzprofile: that's really a stupid change...

duianto: @yzprofile Do you have a suggestion for a better solution?

This is the issue it's trying to fix:
Completion: C-k not working consistently #2974

yzprofile: I mean do not override `C-k` in `company-completion-finished-hook` and `company-completion-cancelled-hook ` or make it as optional.

duianto: Ok, that's more descriptive. A tip for future comments, add more details in the first comment:
>This is stupid...
Because...

😄 

#### This is what I'm seeing with:
`(auto-completion :variables auto-completion-enable-help-tooltip t)`

#### Without this change.
- On an empty line in `.spacemacs`
- Type `dot`

Now a drop down list (company-mode) opens with suggestions.

Before a tooltip has appeared
- `C-j` moves down in the list
- `C-k` moves up in the list

When the tooltip has appeared, then `C-k` inserts `?` (it called `evil-insert-digraph`).
It waits for two key presses, for example: if one types `12` then it inserts the character: ½

#### With this change
`C-k` continues to navigate up in the list after the tooltip has been shown.
To insert a digraph character with `C-k`, the drop down list has to be closed first with `C-g`.

--- 

Could you explain what your observing and expecting.
And include your system info (`SPC h d s` copies it to your clipboard).

PR #12850, opened 2019-10-19
Document helm and ivy color key bindings

duianto:
Helm:
`SPC C l`   calls `helm-colors`
`SPC h d F` calls `spacemacs/helm-faces`

Ivy:
`SPC C e`   calls `counsel-colors-emacs`
`SPC C f`   calls `counsel-colors-faces`
`SPC C w`   calls `counsel-colors-web`
`SPC h d F` calls `counsel-describe-face`

Thanks Miciah for the `SPC h d F` suggestion.

### DIFF
--- a/layers/+completion/helm/README.org
+++ b/layers/+completion/helm/README.org
@@ -19,6 +19,7 @@
   - [[#hjkl-navigation][hjkl navigation]]
   - [[#transient-state][Transient state]]
   - [[#bookmarks][Bookmarks]]
+  - [[#colorsfaces][Colors/Faces]]
   - [[#c-z-and-tab-switch][C-z and Tab switch]]
   - [[#helm-focus][Helm focus]]
   - [[#helm-swoop][Helm-swoop]]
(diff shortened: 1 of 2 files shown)


layers/+completion/helm/README.org
```diff
@@ -180,6 +181,13 @@ In the =helm-bookmarks= buffer:
 | ~C-f~       | toggle filename location                     |
 | ~C-o~       | open the selected bookmark in another window |
 
+** Colors
+~SPC C l~ opens =helm-colors=.
```
Miciah: There is also a key binding <kbd>SPC h d F</kbd> for `helm-faces`, which might warrant mention here.

duianto: Your right, thanks for the suggestion.

The same key binding with the `ivy` layer calls: `counsel-describe-face`
It doesn't show the faces with colors, but opening a face shows the same Emacs help buffer with the face properties. So I'll add that as well to the Ivy layer.

Maybe the heading could be called `Colors/Faces`.

smile13241324:
Merged

PR #12855, opened 2019-10-24
Support relative virtualenv path in .venv file for Python layer

Namburgesas:
Specifying a relative path in a .venv file did not work properly because it was relative to the current directory. Relative path should be relative to the detected .venv file. 

Manually tested all four cases:
1. .venv virtualenv directory in project root
2. .venv text file containing absolute path to virtualenv directory
3. .venv text file containing path to virtualenv directory relative to project root
4. .venv text file containing path to virtualenv directory relative to `pyvenv-workon-home()` which is ~/.virtualenvs/ by default (See https://github.com/jorgenschaefer/pyvenv/blob/master/pyvenv.el)

Also updated README.org for python layer.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2423,6 +2423,7 @@ Other:
   (thanks to Boris Verhovsky)
 - Fixed conflict between pipenv directory .venv and pyvenv file .venv
   (thanks to rgb-24bit)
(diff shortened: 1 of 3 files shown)


Namburgesas:
Thanks @smile13241324 , just squashed. Sorry about the delay. I missed the email notification.

smile13241324:

Namburgesas:
Awesome, thanks!!!

PR #12863, opened 2019-10-25
Set counsel-find-file-at-point to t by default.

xuhdev:
This enables better auto completion of =counsel-find-file=.


duianto:
The PRs changes have been cherry picked into the `develop` branch. https://github.com/syl20bnr/spacemacs/commit/cff6b47e6fd63dbc99e757b8e64b72c4bd5e8807

Modifications:
- Moved the changelog.develop entry from distribution changes to the `Ivy` section, since it's a change in the `ivy` completion layer.
- Added line breaks to keep the lines below 80 chars.

Issue #12838, opened 2019-10-15
search functionality appears to be broken in ivy.


8i8:
On a fresh install when using ivy, on both the master and the develop branch a similar error is occurring: All of the search realated functions seem to be having an issue, Spc s g P for example does not pick up the token at point and launches an error when characters are entered.
The following error is produced : Error in post-command-hook (ivy--queue-exhibit): (file-missing "Setting current directory" "No such file or directory" "tempFilename")

Here is an debug-on-error output created when using search-project-auto-region-or-symbol : 

Debugger entered--Lisp error: (wrong-type-argument stringp nil)
  string-match("\\`\\(.*?\\):\\([0-9]+\\):\\(.*\\)\\'" nil)
  counsel-git-grep-action(nil)
  ivy-call()
  ivy-read("grep from [/home/iain/go/src/pranava/]: " (lambda (&rest --cl-rest--) "Grep in the current directory for STRING." (apply (function (lambda (G0 string &optional _pred &rest _unused) (or (ivy-more-chars) (let* ((default-directory (ivy-state-directory ivy-last)) (args (if (string-match-p " -- " string) (let ((split (split-string string " -- "))) (prog1 (car-safe (prog1 split (setq split (cdr split)))) (setq string (mapconcat (function identity) split " -- ")))) "")) (regex (counsel--elisp-to-pcre (setq ivy--old-re (ivy--regex string))))) (setq spacemacs--counsel-search-cmd (format (symbol-value G0) args regex)) (spacemacs//counsel-async-command spacemacs--counsel-search-cmd) nil)))) '--base-cmd-- --cl-rest--)) :initial-input "Command" :dynamic-collection t :history counsel-git-grep-history :action counsel-git-grep-action :caller spacemacs/counsel-search :keymap (keymap (3 keymap (5 . spacemacs//counsel-edit)) (f3 . spacemacs//counsel-save-in-buffer)) :unwind (lambda nil (counsel-delete-process) (swiper--cleanup)))
[...]

PR #12865, opened 2019-10-26
Ivy layer bugfixes for #12838

Caribpa:
Hi!

These are some bugfixes for #12838

Thanks!

### DIFF
--- a/layers/+completion/ivy/funcs.el
+++ b/layers/+completion/ivy/funcs.el
@@ -262,11 +262,14 @@ that directory."
   (spacemacs/counsel-search dotspacemacs-search-tools
                             nil spacemacs-docs-directory))
 
-(defun spacemacs//counsel-occur ()
+(defun spacemacs//counsel-occur (&optional candidates)
   "Generate a custom occur buffer for `counsel-git-grep'."
   (ivy-occur-grep-mode)
   (setq default-directory (ivy-state-directory ivy-last))
-  (let ((cands ivy--old-cands))
+  (let ((cands (if (boundp 'candidates)
+                   candidates
+                 ivy--old-cands))
+        (inhibit-read-only t))
     ;; Need precise number of header lines for `wgrep' to work.
     (insert (format "-*- mode:grep; default-directory: %S -*-\n\n\n"
                     default-directory))
@@ -274,7 +277,7 @@ that directory."
     (ivy--occur-insert-lines
      (mapcar
       (lambda (cand) (concat "./" cand))
-      ivy--old-cands))))
+      cands))))
 
 (defun spacemacs/counsel-up-directory-no-error ()
   "`counsel-up-directory' ignoring errors."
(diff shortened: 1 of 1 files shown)


layers/+completion/ivy/funcs.el
```diff
@@ -254,27 +254,30 @@ that directory."
                    (buffer-substring-no-properties
                     (region-beginning) (region-end))
                  (thing-at-point 'symbol t))))
-    (counsel-git-grep nil input)))
+    (counsel-git-grep input)))
```
xuhdev: I'm not sure whether this function is useful at all. Thing at point can be better guessed by running `counsel-git-grep` followed by `M-n`.

duianto: `M-n` which calls `ivy-previous-history-element`, only cycles backwards through the prompts history. If one hasn't manually typed in the symbol under the point then it won't be listed.

This PR fixes the issue where `SPC s g P` behaved the same as `SPC s g p` and didn't insert the symbol under the point.

duianto:
https://github.com/syl20bnr/spacemacs/commit/cb99bf9f390abbdcd4f2760297037fa1752bf9c6
https://github.com/syl20bnr/spacemacs/commit/0b0203e03f7bb67c7ce10fd08677f9d4a528d7a3

layers/+completion/ivy/funcs.el
xuhdev: @duianto `M-n` calls the next (not previous) element in history. When the next element does not exist, it tries to predict what the user wants to input, in this case, what the thing at point is. See https://oremacs.com/2016/02/15/counsel-ffap/

duianto: Your right, `M-n` does pick up the symbol under the cursor.
I did not know that. Thanks for the explanation.

I'm used to navigating backwards through the history at prompts, therefore I assumed that you meant `M-p`, but I still managed to type the correct key binding: `M-n` 😄 

Still, `SPC s g P` was broken, but it's fixed now.

layers/+completion/ivy/funcs.el
```diff
@@ -254,27 +254,30 @@ that directory."
...
   (spacemacs/counsel-search dotspacemacs-search-tools
                             nil spacemacs-docs-directory))
 
-(defun spacemacs//counsel-occur ()
+(defun spacemacs//counsel-occur (&optional candidates)
   "Generate a custom occur buffer for `counsel-git-grep'."
   (ivy-occur-grep-mode)
   (setq default-directory (ivy-state-directory ivy-last))
-  (let ((cands ivy--old-cands))
+  (let ((cands (if (boundp 'candidates)
```
yuhan0: I see this PR has already been merged but this should just be `(or candidates ivy--old-cands)`, the symbol itself is always bound in this context.

duianto: Thanks for the suggestion. Fixed: https://github.com/syl20bnr/spacemacs/commit/b837c1d39705b9e03bd9922da249e4c154023a92

PR #12866, opened 2019-10-26
Keep ~C-h~ binding in =counsel-find-map= only for hjkl navigation

xuhdev:
It has already been bound to <kbd>C-backspace</kbd> and <kbd>C-delete</kbd>. Such an
additional binding is redundant (if hjkl navigation is not used) and makes the original <kbd>C-h</kbd> binding
(help) unavailable.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1870,6 +1870,8 @@ Other:
   of =counsel-find-file= (thanks to Hong Xu)
 - Fixed broken =ivy-occur= (thanks to Carlos Ibáñez)
 - Fixed =counsel-git-grep= with input ~SPC s g P~ (thanks to Carlos Ibáñez)
+- Keep ~C-h~ binding in =counsel-find-map= only for hjkl navigation (thanks to
+  Hong Xu)
 **** Imenu-list
 - Changed ~SPC b i~ to ~SPC b t~ for =imenu= tree view
   (thansk to Sylvain Benner)
--- a/layers/+completion/ivy/packages.el
+++ b/layers/+completion/ivy/packages.el
@@ -131,7 +131,6 @@
        'counsel-find-file
        spacemacs--ivy-file-actions)
 
-      (define-key counsel-find-file-map (kbd "C-h") 'counsel-up-directory)
       (define-key read-expression-map (kbd "C-r") 'counsel-minibuffer-history)
       ;; remaps built-in commands that have a counsel replacement
       (counsel-mode 1)
--- a/layers/+spacemacs/spacemacs-completion/funcs.el
+++ b/layers/+spacemacs/spacemacs-completion/funcs.el
@@ -199,6 +199,7 @@ See https://github.com/syl20bnr/spacemacs/issues/3700"
                        ivy-switch-buffer-map))
       (define-key map (kbd "C-j") 'ivy-next-line)
       (define-key map (kbd "C-k") 'ivy-previous-line))
+    (define-key counsel-find-file-map (kbd "C-h") 'counsel-up-directory)
     (define-key ivy-minibuffer-map (kbd "C-h") (kbd "DEL"))
     ;; Move C-h to C-S-h
     (define-key ivy-minibuffer-map (kbd "C-S-h") help-map)


duianto:
It's seems more comfortable and vim navigation like to:
- complete the selected directory (forward) with: `C-l` 
- and move to the parent directory (backward) with: `C-h` 

since `C-j` and `C-k` moves down and up in the list.

The same behavior is observed with `helm`.

`C-backspace` requires some pinky stretching.

At least for me `C-delete` calls `ivy-kill-word` which deletes the word after the cursor.

With this PR applied, `C-h` doesn't get restored to the default Emacs help behavior.

`C-h l` (`view-lossage`) shows:
```
 C-h [anonymous-command]
 [ivy-backward-delete-char]
```

And after typing some characters at the `SPC f f` prompt, then `C-h` erases them one at a time, but if there are no typed characters then `C-h` moves up in the directory structure.

#### Slight side note
One behavior that's missing or not enabled if it already exists, is that in `ivy` when moving to the parent directory, then the cursor just moves to the first line: `./`.

In `helm` the previous directory is selected when moving up to it's parent directory.

Are you aware if ivy has such a feature?

xuhdev:
I agree that your whole key map scheme may be what one wants, but as far I can see, our observed behavior seems different: `C-l` is mapped to `recenter-top-bottom`, `C-j` is mapped to `ivy-alt-done`, and `C-k` is mapped to `ivy-kill-line`. Even backspace, if the last character is `/`, would move one directory level up, which might not be too stretching. After removing `C-h` mapping, it also comes back to help for me. Please let me know how you would like to proceed.

This is the original commit that adds `C-h` but it also moves the help mapping, which for some reason got lost: 1d28c96faea89610e4ae0dc735f72a329c44c5a5

The fact that the cursor moves to `./` seems reasonable to me, because when one goes up one level, it is more likely that the user would want to select the parent dir than the previous dir (which can be selected by staying in the previous dir and press Enter). Do you have a use case? Maybe I can help you from there.

duianto:
Your probably using the `hybrid` editing style. I'm getting the same commands as you, when I just tried it.

I'm using the `vim` editing style.

#### Side note:
The `hybrid` editing style has `hybrid-style-enable-hjkl-bindings` variable, when it's set to `t` then  `hybrid` mode navigates the same as `vim`.

---

The behavior I saw when the cursor ends up at `./` when moving to the parent directory was caused by commenting out `C-h` as `counsel-up-directory` with the `vim` editing style.

Without this PR then `counsel-up-directory` behaves as I expected, the child directory is selected when navigating to the parent.

In addition to `helm` having that behavior, I just noticied that `^` in a dired buffer also selects the child directory.

It's probably more helpful to select the child directory than the top of the list `./`, because it's easier to jump to the top of the list with `M-S-<` than it is to navigate down to find the child directory again.

One example is if one is searching for a file/image but don't know/remember it's name (therefore can't search by file). One just knows that it's in one of a couple of adjacent directories.

---

The commit that added `C-h` for `counsel-up-directory` also added `C-S-h` for `help-map`.
The same key binding in `helm`, just calls `describe-key`.
So it might be on purpose to try and keep the key bindings similar where ever possible.

xuhdev:
OK. How about binding `C-h` only if in vim editing style? This would be more consistent. `C-S-h` has been lost, but I'm not sure whether it was intentional or merely an accident.

I think the idiomatic way to go back down to the previous directory in `counsel-find-file` is to call `counsel-down-directory`. It looks like it encourages a workflow that is different from helm or dired.

duianto:
Your right, `C-h` for `counsel-up-directory` should be limited to the editing styles:
- `hybrid` style with `hybrid-style-enable-hjkl-bindings` set to `t`.
- `vim`

Because the normal `hybrid` style seems to use the same key bindings as `emacs` style in prompts, completion buffers, etc.

This seems to work:
```elisp
      (when (or (eq 'vim dotspacemacs-editing-style)
                (and (eq 'hybrid dotspacemacs-editing-style)
                     hybrid-style-enable-hjkl-bindings))
        (define-key counsel-find-file-map (kbd "C-h") 'counsel-up-directory))
```
I tried all four editing style variants and it seems to work as expected.

xuhdev:
Sounds good. Should this be defined as a function? It looks like there may be a lot of occasions that this code snippet may be called.

```elisp
            (or (eq 'vim dotspacemacs-editing-style)
                (and (eq 'hybrid dotspacemacs-editing-style)
                     hybrid-style-enable-hjkl-bindings))
```

duianto:
I just found that there already is a function that defines ivy hjkl bindings:
https://github.com/syl20bnr/spacemacs/blob/57721d8e768c4cd4761d9a2b36a83390456d82ac/layers/%2Bspacemacs/spacemacs-completion/funcs.el#L192-L212

```elisp
(define-key counsel-find-file-map (kbd "C-h") 'counsel-up-directory)
```
should probably be moved into that functions vim/hybrid with hjkl section.

xuhdev:
@smile13241324 @duianto Sorry for the delay. I've updated the PR now.

duianto:
The PRs changes have been cherry picked into the `develop` branch. https://github.com/syl20bnr/spacemacs/commit/383da46e38712ef6d08ef3e1e96d418bdf2182db

Changes
- Changed `counsel-find-map` to `counsel-find-file-map`

Issue #11329, opened 2018-09-16
Sudo-edit does not work on remote files


roblabla:
#### Description :octocat:
sudo-edit does not work on remote (TRAMP) files. It crashes with Wrong number of arguments.

#### Reproduction guide :beetle:
- Start Emacs
- Open a remote file (for instance, on ssh)
- Call spacemacs/sudo-edit (SPC f E)

*Observed behaviour:* :eyes: :broken_heart:
Wrong number of arguments (6 . 7), 5

*Expected behaviour:* :heart: :smile:
A new buffer is opened with the file in sudo.

#### System Info :computer:
- OS: darwin
- Emacs: 26.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 9865e7731)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
(yaml rust ivy auto-completion better-defaults emacs-lisp git neotree
      (shell :variables shell-default-shell 'eshell shell-enable-smart-eshell t shell-default-height 30 shell-default-position 'bottom)
      syntax-checking version-control)
```
- System configuration features: JPEG NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS THREADS

#### Backtrace :paw_prints:
```
Debugger entered--Lisp error: (wrong-number-of-arguments (6 . 7) 5)
  tramp-make-tramp-file-name(#("sshx" 0 4 (escaped t)) #("roblabla" 0 8 (escaped t)) #("roblab.la" 0 9 (escaped t)) nil nil)
  (let* ((new-hop (tramp-make-tramp-file-name parsed-method parsed-user parsed-host nil parsed-hop)) (new-hop (substring new-hop 1 -1)) (new-hop (concat new-hop "|")) (new-fname (tramp-make-tramp-file-name "sudo" "root" parsed-host parsed-localname new-hop))) new-fname)
[...]

PR #12867, opened 2019-10-27
[spacemacs/sudo-edit] Fix sudo-edit on TRAMP & [spacemacs/counsel-search] Fix wrong initial directory

Caribpa:
Hi!

Here's a pull request fixing #11329

Btw, the indentation was changed in the second tramp-make-tramp-file-name for consistency with the first tramp-make-tramp-file-name.

### DIFF
--- a/layers/+completion/ivy/funcs.el
+++ b/layers/+completion/ivy/funcs.el
@@ -169,10 +169,10 @@ that directory."
                                   spacemacs--counsel-search-max-path-length)
                                (length default-directory))))))
     (cond ((string= tool "ag")
-           (counsel-ag initial-input initial-directory nil
+           (counsel-ag initial-input default-directory nil
                        (format "ag from [%s]: " display-directory)))
           ((string= tool "rg")
-           (counsel-rg initial-input initial-directory nil
+           (counsel-rg initial-input default-directory nil
                        (format "rg from [%s]: " display-directory)))
           (t
            (ivy-read
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -512,19 +512,30 @@ FILENAME is deleted using `spacemacs/delete-file' function.."
        (with-parsed-tramp-file-name fname parsed
          (when (equal parsed-user "root")
            (error "Already root!"))
-         (let* ((new-hop (tramp-make-tramp-file-name parsed-method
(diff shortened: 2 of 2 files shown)


layers/+completion/ivy/funcs.el
```diff
@@ -169,10 +169,10 @@ that directory."
                                   spacemacs--counsel-search-max-path-length)
                                (length default-directory))))))
     (cond ((string= tool "ag")
-           (counsel-ag initial-input initial-directory nil
+           (counsel-ag initial-input default-directory nil
                        (format "ag from [%s]: " display-directory)))
           ((string= tool "rg")
-           (counsel-rg initial-input initial-directory nil
+           (counsel-rg initial-input default-directory nil
```
xuhdev: I think this is right: `default-directory` is the one that user selects

xuhdev: @duianto Any plan to get this merged :) I'm a bit anxious about this bug I introduced

duianto: Since the search fix was applied to what seems to be an unrelated PR. I asked the maintainers if they had any suggestions for how this should be handled. But I haven't heard back. They are busy.

Do you know anything about TRAMP and could you test if that commit works as expected?

Caribpa: My bad about the mess with the commit on unrelated PR. I can create separate branches with proper PR if that'd help.

Regarding TRAMP, I use it for work everyday, and even for accessing my home router (it has a `busybox` `ash` shell). I also use `lsp` with `ccls` to code on C and C++ remotely. When I started using Spacemacs I realized that `SPC f E` was broken for my use case (mostly for editing root config files on the company's server), so I fixed it out of necessity and convenience.

Anyways, I can say that this was only tested with `ssh`, `scp`, and `rsync` protocols, but the logic behind this fix should work for every properly configured TRAMP protocol. Worst case scenario a `Protocol not supported for multi-hopping` message will appear in the minibuffer.

duianto:
https://github.com/syl20bnr/spacemacs/commit/b5727f77994e2c91d780fa6176aa35615d6d61dd
https://github.com/syl20bnr/spacemacs/commit/6ca42efe7c7c304a934b1917ecaeec9cc87cbb82
https://github.com/syl20bnr/spacemacs/commit/a2fbb3d94af080a76a65cc3d71558b8bd81bfe5b

PR #12868, opened 2019-10-27
[helm] Add helm-ls-git-ls to SPC g f f

duianto:
The HELM Lsgit buffer has three sections:
- Git status (files: modified, deleted, etc.)
- Buffers in git project
- Git files


smile13241324:
Merged

PR #12712, opened 2019-09-09
[c-c++] Added clangd support. Separated lsp server selection from backend.

cormacc:
Replaces: #12443
Closes: #12351 

Added support for `clangd` (to the existing `ccls` and `cquery`) as lsp backends to the c-c++ layer.
Updated the backend selection mechanism -- now aligned with the approach used in the python layer, suggested by @sdwolfz during an earlier PR review.

Specifically:
The existing mechanism is to set the `'c-c++-backend` layer variable to one of `nil`, `'rtags`, `'lsp-ccls` or `'lsp-cquery`.
The new/proposed mechanism is to set `'c-c++-backend` to one of `nil`, `'rtags` or `'lsp` and (optionally) set the `'c-c++-lsp-server` layer variable to one of `'clangd`, `'ccls` or `'cquery`. If unspecified, the server selection defaults to `'clangd`.

Re. default backend selection, I personally prefer/use `'ccls`, but opted to leave `'clangd` as the default, as it's the default supported by the upstream `lsp-mode` package and (arguably) easier to install. Use of `'cquery` is no longer recommended, as the project seems to have been abandoned, however support has been left in place for existing users with a functional installation.

Also opportunistically tidied up a few bits of configuration, including eliminating some company-related config that's now done automagically upstream by the `lsp-mode` package.

Thanks to @kvaneesh for some early feedback, and @auscompgeek for accomodating this PR. And @yyoncho and @maskray for all their work on `lsp-mode` and `ccls` respectively.

N.B. there's a clangd-specific find-other-file extension to the LSP which I haven't bound, as...
a) I haven't got it to work (clearly not reading the spec / calling it correctly).
b) I'm perfectly happy with projectile-find-other-file, and using ccls rather than clangd anyway
If anyone else is interested in taking a look while reviewing this PR, it's wrapped in interactive defuns: `c-c++/find-clangd-other-file` and `c-c++/peek-clangd-other-file`. If you do get it to work, it could be bound under `,go`, `,Go` OR replace the existing `,ga` keybinding.

PR #12869, opened 2019-10-28
[c-c++] Layer refactor

syl20bnr:
@cormacc @emacs18 @robertodr @yyoncho @henriquenj

This refactor aligns the C/C++ layer with the others regarding multiple backends definition.

The base commit for this PR is the one by @cormacc in this PR: https://github.com/syl20bnr/spacemacs/pull/12712 so most of my comments start from there.

The main mechanism used in this PR (and other layers) is the hooks `c-mode-local-vars-hook` and `c++-mode-local-vars-hook` via a set of functions in `funcs.el` at the top of the file. These hooks allow to evaluate code right after evaluating the file `.dir-locals.el` so users can define different backends for different projects. We have some big limitation right now though, I'll explain at the end of this message.

Backends are now completely independent which means they don't share configuration, it is better to have code duplication here to retain proper isolation and make the code easier to follow. Each backend has a unique name so each LSP server has its own backend value: `lsp-clangd`, `lsp-ccls` and `lsp-cquery`. It makes sense to do that because now we don't have shared configuration. It will also make it easier to remove a backend if we need to.

I removed some layer variables in `config.el` that were redundant, we try to avoid to duplicate variables of packages and prefer users to directly set the package variables instead whenever it is possible. But sometimes it is good to give some additional visibility to a feature by making a layer variable so it clearly stands out or just to mention them in the `README.org`. An example is the `executable` variable for LSP servers, users can directly set the original package variable if they want to, we can add a section in the `README.org` to mention it though.

I moved the `ycmd` layer back to C/C++ layer as I don't see how `ycmd` is going to be useful for other languages given the growing popularity of LSP and it was easier to make it a C/C++ backend directly in the layer.

I tested `lsp-clangd`, `lsp-ccls` and `rtags` and they seem to work correctly though I was not able to have proper `compile_commands.json` file so there were lots of error with the std namespace etc...
The projects I tested with:
- Catch2
- aria2
- json
- googletest

The big limitation is regarding key bindings conflicts. This is a not a big deal when using different LSP backends for different projects, say one project uses `lsp-clangd` and the other `lsp-ccsl`. But this is a problem when defining `rtags` for a project and `lsp-xxxx` for another project in the same Emacs session. I doubt lots of users will hit this use case, the work around would be to start two different Emacs sessions. We will at some point make a big refactor of the key bindings management in Spacemacs and it will solve this use case.


thanhvg:
I am very excited about this pr. C++ layer regression in dev branch has prevented me to to update spacemacs. 

I have a question about the usage of `:requires` in `packages.el`. I think `:requires` works with list or one item.
So when you `:requires` more than one packages you need to put them in a list

This will not work, only `flycheck` will be considered for `:requires`:
``` elisp
         (flycheck-rtags :requires flycheck rtags)
```
This will work:
``` elisp
         (flycheck-rtags :requires `(flycheck rtags))
```

I might be wrong so it's a  chance for me to learn here anyway. I found the `:requires` spec in `core-configuration-layer.el` file. function `configuration-layer/make-package` and class `cfgl-package`

Even spacemacs unit test also has `:requires` as a list.

yyoncho:
> The big limitation is regarding key bindings conflicts.

IMO we could have backend-specific minor mode and use it for the backend-specific shortcuts.

layers/+lang/c-c++/funcs.el
```diff
@@ -12,151 +12,335 @@
...
+      "gF" 'ffap-other-window
+      ;; hierarchy
+      "ghc" 'ccls-call-hierarchy
+      "ghC" 'spacemacs/c-c++-lsp-ccls-call-hierarchy-inv
+      "ghi" 'ccls-inheritance-hierarchy
+      "ghI" 'spacemacs/c-c++-lsp-ccls-inheritance-hierarchy-inv
+      ;; members
+      "gmh" 'ccls-member-hierarchy)
+
+    ;;Replace this with lsp-goto-implementation in lsp-layer?
```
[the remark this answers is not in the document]
cormacc: `lsp-goto-implementation` is a single function in the lsp-mode package. Strictly speaking (just looked up the source), it's an alias to `lsp-find-implementation`. Didn't exist when I originally added support for ccls/cquery - and pretty sure that comment is now orphaned / redundant -- i.e. it related to a ccls/cquery specific binding from an earlier pass that I stripped out after adding the binding to the generic lsp-mode version that this layer now inherits (under `,gi`)

layers/+lang/c-c++/funcs.el
```diff
@@ -12,151 +12,335 @@
...
+      "gF" 'ffap-other-window
+      ;; hierarchy
+      "ghc" 'cquery-call-hierarchy
+      "ghC" 'spacemacs/c-c++-lsp-cquery-call-hierarchy-inv
+      "ghi" 'cquery-inheritance-hierarchy
+      "ghI" 'spacemacs/c-c++-lsp-cquery-inheritance-hierarchy-inv
+      ;; members
+      "gmh" 'cquery-member-hierarchy)
+
+    ;;Replace this with lsp-goto-implementation in lsp-layer?
```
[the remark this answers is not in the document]
cormacc: As above

cormacc:
Looks great, though minding my daughter at the minute and not on a dev machine to actually test it. Have some C work to do tomorrow so will pull the changes and test in the morning.

henriquenj:
Thanks for the PR. I have tested with some big repos using clangd on macos and everything seems to be working.

cormacc:
Looking good here based on some quick and dirty evaluation of three lsp backends. Will stick with lsp-ccls for the rest of the day and report back if I encounter any issues. My use is restricted (c/embedded), so won't be exercising the c++-specific/OO-related bindings.

layers/+lang/c-c++/README.org
```diff
@@ -9,197 +9,201 @@
...
-                         c-c++-backend 'lsp-ccls
-                         c-c++-lsp-executable (file-truename "~/dev/cpp/ccls/Release/ccls")
-                         c-c++-lsp-sem-highlight-rainbow t)))
-#+END_SRC
+ #+BEGIN_SRC emacs-lisp
+   (setq-default dotspacemacs-configuration-layers
+                 '((c-c++ :variables
+                          c-c++-adopt-subprojects t
+                          c-c++-backend 'lsp-ccls
+                          c-c++-lsp-executable (file-truename "~/dev/cpp/ccls/Release/ccls")
```
gdkrmr: This doesn't work. The readme should probably mention somewhere that you have to set `lsp-clients-clangd-executable ` for clangd, I am not sure about the other backends.

henriquenj: I use clangd and I didn't have to set `lsp-clients-clangd-executable`. Are you sure your clangd is available on the PATH? What does `M-: (executable-find "clangd")` returns?

layers/+lang/c-c++/packages.el
```diff
@@ -182,75 +186,45 @@
...
   (dolist (mode c-c++-modes)
     (spacemacs/add-realgud-debugger mode "gdb")))
 
 (defun c-c++/post-init-semantic ()
-  (spacemacs/add-to-hooks 'semantic-mode c-c++-mode-hooks)
-  (when (or (configuration-layer/layer-used-p 'gtags)
-            (configuration-layer/layer-used-p 'lsp))
-    (spacemacs/add-to-hooks 'spacemacs//disable-semantic-idle-summary-mode c-c++-mode-hooks t)))
+  (add-hook 'c-mode-local-vars-hook #'spacemacs//c-c++-setup-semantic)
+  (add-hook 'c++-mode-local-vars-hook #'spacemacs//c-c++-setup-semantic))
```
gessen: I encountered an error here saying spacemacs//c-c++-setup-semantic is a void-function. I can't find this function defined anywhere, am I missing something?

layers/+lang/c-c++/funcs.el
```diff
@@ -12,151 +12,335 @@
...
+
+(defun spacemacs//c-c++-setup-lsp-ccls ()
+  "Setup LSP ccls."
+  (require 'ccls)
+  ;; configuration
+  (let ((cache-dir
+         (if (null c-c++-lsp-cache-dir)
+             (concat spacemacs-cache-directory "lsp-ccls")
+           (concat (file-truename (file-name-as-directory c-c++-lsp-cache-dir))
+                   "lsp-ccls"))))
```
gessen: ccls doesn't require `cache-dir` to be defined as it uses `.ccls-cache` default directory, similar to `.clangd`. With this scheme it is impossible to set it to a default value by not defining it. I'd frankly let a user set it through `ccls-initialization-options` instead as only cquery requires this setting and clangd does not even allow to change it.

layers/+lang/c-c++/funcs.el
```diff
@@ -12,151 +12,335 @@
...
+    (append ccls-initialization-options
+            `(:cache (:directory ,cache-dir))))
```
gessen: This function only returns a new list but doesn't apply it to `ccls-initialization-options`

layers/+lang/c-c++/README.org
gdkrmr: maybe this wasn't clear: If clangd is not in your path you have to set `lsp-clients-clangd-executable`, `c-c++-lsp-executable` does **not** work for this purpose. `clangd` is relatively new and people with older distros will have to download clang manually and put it somewhere in order to be able to use `clangd`.

gdkrmr:
I think this is very cool, thanks for the hard work!! With `clangd` it works perfectly for me! What is missing for this to get merged?

layers/+lang/c-c++/funcs.el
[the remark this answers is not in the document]
syl20bnr: Thanks for the explanation, I removed both of them

layers/+lang/c-c++/README.org
syl20bnr: Added a section to the README.org titled `Path to LSP server executables`

layers/+lang/c-c++/packages.el
syl20bnr: Thank you for catching this, I added semantic config at the end and forgot to add the dispatch function.
Should be fixed.

layers/+lang/c-c++/funcs.el
syl20bnr: I made `c-c++-lsp-cache-directory` a `cquery` only variable by renaming it `c-c++-lsp-cquery-cache-directory`. I updated the README.org file to add the information you mentioned.
Tell me if this sounds good to you now.

layers/+lang/c-c++/funcs.el
syl20bnr: I removed this as per your suggestion and added a mention of `ccls-initialization-options` in the README.org.

syl20bnr:
> IMO we could have backend-specific minor mode and use it for the backend-specific shortcuts.

@yyoncho you are right. We are heading into this direction as we have plan to refactor the key bindings management in Spacemacs and improve them quite a lot.

syl20bnr:
Tried to address all the code comments, let me know if that's ok.

syl20bnr:
> Even spacemacs unit test also has :requires as a list

@thanhvg indeed, let me give it a 👀

syl20bnr:
@thanhvg Good catch, it was indeed a bug. I added some tests to make sure it was a bug. I fixed the places where we forgot to use a list. 

Turns out that in the code we have some kind of convention to allow a single symbol without being a list, but for multiple values a list must be passed. IOW only one value is allowed after a keyword so either it is a symbol, either it is a list. Here is in the code an example of such "hack" to allow single symbol whereas the spec allows only list: https://github.com/syl20bnr/spacemacs/blob/develop/core/core-configuration-layer.el#L827

Bad news is that I'm not sure we can easily detect this and then log a warning that some values are ignored. 

Thank you for catching this 💯

layers/+lang/c-c++/funcs.el
gessen: Looks good!

thanhvg:
It works on mine with `ccls` backend. Here is my settings:

``` elisp
     (c-c++ :variables
            c-c++-adopt-subprojects t  
            c-c++-backend 'lsp-ccls
            ccls-executable (file-truename "~/git/ccls/Release/ccls")
            c-c++-lsp-sem-highlight-rainbow t)
```

I didn't enable `dap-mode` it's way over my head :). 

I suggest we defer loading of `ccls` and `cquery` packages.

``` elisp
(defun c-c++/init-cquery ()
  (use-package cquery
    :defer t))

(defun c-c++/init-ccls ()
  (use-package ccls
    :defer t))
```

Without :defer these packages will  too soon trigger `lsp` and then `yasnipet` and all other `yasnipet` support packages such as `yasnippet-snippets` etc. The end result is `auto-completion/init-yasnippet` which is supposed to run first now runs too late. Our snippets become empty for all layers not just this `c-c++` (`SPC i s` shows empty list or just a few). Btw you must have latest `lsp` package to be able to notice this problem.

@yyoncho pushed a fix recently for this problem (06cf8e384215156f070ccfc1ec3fc983e5eb3509) but I don't think it works for this case.

We can move some of the logic in `auto-completion/init-yasnippet` to `auto-completion/pre-init-yasnippet` but in this case I find adding :defer will work just fine.

Other than that just merge it please :+1: .

syl20bnr:
Cherry-picked into develop branch 🎉 , I close this PR.

@thanhvg I defered the packages, thank you for the great feeback.
I encountered the behavior you described while updating my packages and restarting afterward even with the defer for ccls and cquery, unfortunately I cannot replicate it now and Emacs launches just fine.

duianto:
There was a report in the main gitter chat (and I'm seeing the same thing) about this error message:
>File local-variables error: (file-error Cannot open load file No such file or directory lsp-ui-flycheck)

source: https://gitter.im/syl20bnr/spacemacs?at=5dcd6007e75b2d5a19cfd02b

When the `c-c++` layer is enabled without the `lsp` layer.

And there was a request for a more descriptive error message.
https://gitter.im/syl20bnr/spacemacs?at=5dcd6649c26e8923c4237d22

Can someone with more experience when it comes to adding a missing layer message, either open a PR or suggest a solution?

cormacc:
Hmmm - the report linked in gitter indicates that the backend was specified as 'lsp-ccls, but that the 'lsp layer wasn't enabled explicitly.

That's how I'm setup, and don't see this issue, either on my main linux machine running emacs 27 or on a windows test box running emacs 26.x. Both are working off the latest commit on develop, and have had packages updated today.

Any more details of your own config with the issue?

duianto:
I enabled the `c-c++` layer like this:
```
(c-c++ :variables c-c++-backend 'lsp-ccls)
```
without having the `lsp` layer enabled.

And when the following example file is opened, then the messages buffer shows:
>File local-variables error: (file-missing Cannot open load file No such file or directory lsp-ui-flycheck)

`hello.cpp`
```
# include <iostream>

using namespace std;

int main ( int argc, char *argv[] );

//****************************************************************************80

int main ( int argc, char *argv[] )

//****************************************************************************80
//
//  Purpose:
//
//    MAIN is the main program for HELLO.
//
//  Discussion:
//
//    HELLO is a very simple C++ program.
//
//  Licensing:
//
//    This code is distributed under the GNU LGPL license.
//
//  Modified:
//
//    13 March 2004
//
//  Author:
//
//    John Burkardt
//
{
  cout << "\n";
  cout << "HELLO:\n";
  cout << "  C++ version.\n";
  cout << "\n";
  cout << "  Hello, world!\n";

  return 0;
}
```

<details><summary>
System Info
</summary>
<pre>
#### System Info :computer:
- OS: windows-nt
- Emacs: 26.3
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. eaa76b7ee)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
(autohotkey auto-completion
            (c-c++ :variables c-c++-backend 'lsp-ccls)
            elixir emacs-lisp git ivy markdown multiple-cursors
                 '("~/org/notes.org"))
            python spell-checking syntax-checking treemacs version-control)
```
</pre>
</details>

cormacc:
The fact that it disappears when you explicitly enable the lsp layer suggests some sort of bug in dependency resolution when a layer is included via layers.el rather than explicitly, but strange that I can't reproduce locally. Also, the conditional logic in c-c++/layers.el looks fine to me, so if this IS the issue, it may be occurring elsewhere in spacemacs dependency resolution rather than in the layer itself (but take that with plenty of salt / handwaving).

I've only verified this properly this morning BTW, and only on Arch linux / emacs 27 -- was clearly text blind yesterday, as config file had explicitly included lsp layer, but commenting this line and testing with an empty folder containing your example  worked fine for me -- no errors. Will try to take another look on the windows box later if I get a chance, but under deadline pressure at the moment.

lsp-ui-flycheck.el is part of the lsp-ui package. Have you browsed your .emacs.d/elpa directories to verify that lsp-ui package is NOT present when lsp layer is enabled implicitly via the c-c++ layer? Another giveaway would be seeing packages installed / removed when you toggle explicitly enabling the layer in your spacemacs config, but leave the lsp-ccls backend specified for the c-c++ layer.

duianto:
I have the variable: `dotspacemacs-install-packages` in `.spacemacs` set to: `'used-but-keep-unused`
so that packages won't be removed all the time when I switch layers while testing issues/prs.

When I set it to: `'used-only`
with the `lsp` layer is commented out and the `c-c++` layer defined as:
`(c-c++ :variables c-c++-backend 'lsp-ccls)`

When Spacemacs is restarted then these 5 lsp packages were removed:
>(Spacemacs) --> deleting lsp-ui... [96/221]
Package ‘lsp-ui-20191030.1350’ deleted.
(Spacemacs) --> deleting lsp-treemacs... [97/221]
Package ‘lsp-treemacs-20191105.1838’ deleted.
(Spacemacs) --> deleting lsp-python-ms... [98/221]
Package ‘lsp-python-ms-20191103.1832’ deleted.
(Spacemacs) --> deleting lsp-mode... [99/221]
Package ‘lsp-mode-20191107.333’ deleted.
(Spacemacs) --> deleting lsp-julia... [100/221]
Package ‘lsp-julia-20191011.1005’ deleted.

(including 216 other packages)



#### Additional notes:
I don't know if it's related but I haven't installed `clangd` or `ccls`.

When the `lsp` layer is enabled, and the `hello.cpp` file is opened, then the minibuffer shows:
>LSP :: No LSP server for c++-mode(check *lsp-log*).

and the lsp-log buffer shows:
>Command "clangd" is not present on the path.
Command "ccls" is not present on the path.

Don't worry about it now. Focus on the deadline it's more important.

PR #12877, opened 2019-10-29
vterm: M-n/M-p for next/previous command

agzam:
Adds a few useful keybindings for vterm

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2653,6 +2653,8 @@ Other:
 - Enabled mouse based pasting into term shells (thanks to Sheng Yang)
 - Set =term-char-mode-point-at-process-mark= to =nil= to allow evil normal-mode
   movement in term shells on emacs 26.1
+- vterm keybindings added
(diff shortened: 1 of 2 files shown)


duianto:
They should probably also work in the Emacs and Hybrid states and possibly even in Evil `normal` state?

```elisp
      (define-key vterm-mode-map (kbd "M-p") 'vterm-send-up)
      (define-key vterm-mode-map (kbd "M-n") 'vterm-send-down)
```
This seems to enable it in all three.

agzam:
@duianto I have added suggested keybindings.

duianto:
The PRs changes have been cherry picked into the `develop` branch. https://github.com/syl20bnr/spacemacs/commit/08fdb02d7d80e6adfeeff74286fc75a6ab51e268

`M-n`, `M-p` and `M-/` didn't need to be duplicated. They work in all four states Evil normal/insert, Emacs and Hybrid when they are defined with just `define-key`.

I also added `C-y` to `vterm-yank` in Evil insert state, because it called `evil-copy-from-above`, which doesn't seem to useful in a terminal.

PR #12878, opened 2019-10-29
vterm: make use of hippie-expand key

agzam:


duianto:
The same as in the `M-n/M-p` PR.
```elisp
(define-key vterm-mode-map (kbd "M-/") 'vterm-send-tab)
```
this seems to enable it in Emacs, Hybrid and even in Evil normal state.

agzam:
closing in favor of https://github.com/syl20bnr/spacemacs/pull/12877

PR #12883, opened 2019-11-02
lsp layer must depend on yasnippet

yyoncho:
- if this line is missing yasnippets might not be loaded because lsp-mode might
load the snippets before the auto-completion layer kick in.


duianto:
The PRs changes have been cherry picked into the `develop` branch. https://github.com/syl20bnr/spacemacs/commit/06cf8e384215156f070ccfc1ec3fc983e5eb3509

PR #12886, opened 2019-11-04
[julia] Limit evil-surround redefinitions to julia-mode

duianto:
```
problem:  The julia layer redefines evil-surround `b` globally to:
          begin <x> end
solution: Limit the redefinition to julia-mode buffers.
```
---

The issue seems to only occur the first time a file is opened.

For example if the scratch buffer is opened: `SPC b s`
and one types: `abc`
and surrounds it with the evil-surround "key" `b`: `ysiwb`
result: `begin abc end`
expected: `(abc)`

If one now opens `.spacemacs`: `SPC f e d`
then the steps above results in: `(abc)`

The reverse also happens, if `.spacemacs` is opened first, and then the scratch buffer.
then `.spacemacs` shows: `begin abc end`
and the scratch buffer: `(abc)`

After `b` has started surrounding with parentheses, then it also happens in `.jl` julia-mode buffers.
So it wasn't working as expected.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1974,6 +1974,8 @@ Other:
 - Fixes:
   - Loaded =lsp-julia= from melpa (thanks to Guido Kraemer)
   - Fixed =julia-mode-local-vars-hook= (thanks to Guido Kraemer)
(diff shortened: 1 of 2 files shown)


smile13241324:
Merged

PR #12887, opened 2019-11-04
 [python] fixes spacemacs/python-remove-unused-imports

FrancescElies:
problem: `spacemacs/python-remove-unused-imports` has multiple docstrings. Any call to this function fails.

This pr fixes that.


FrancescElies:
closing, sorry for the noise.

My problem was a corrupted venv.

PR #12885, opened 2019-11-04
Optimize =spacemacs/counsel-search= for =ag= and =rg=

xuhdev:
Use =counsel-ag= and =counsel-rg= when possible. They have more
optimizations, such as switching directory, the ability to turn on and off casefold, etc. Counsel will also carry
future improvement with these two functions, and spacemacs will
automatically benefit from it.

PR #12890, opened 2019-11-06
Use string= instead of eq when comparing strings.

xuhdev:
This is a follow-up of #12885, in which I used the wrong operator.

Sorry about the error


duianto:
The PRs changes have been cherry picked into the `develop` branch. https://github.com/syl20bnr/spacemacs/commit/5cd73edd1d1fb23d6d652973c9566eab1cc0b754

PR #12892, opened 2019-11-06
[core] [helm] add action to recompile an elpa package

thanhvg:
How to use: `SCP h p` select package then `F3` or `C-z` and choose `Recompile`

Motivation: Some packages when compiling need other packages to be loaded first.
The pacakage maintainers should make sure of this requirement but sommetimes
they don't. It also doesn't help when Spacemacs is lazy loading and compiles
updated pacakges on startup. So the manual fix for this problem is recompile the
package once Spacemacs has fully loaded. This requires user go to the package
install location to delete elc files and then do a `spacemacs/recompile-elpa`.
This commit will do that chore for them.

This will make fixing the problem with `org-plus-contrib` or `dumb-jump` update
every now and then easier.

### DIFF
--- a/core/core-funcs.el
+++ b/core/core-funcs.el
@@ -278,6 +278,15 @@ result, incrementing passed-tests and total-tests."
      (directory-files-recursively user-emacs-directory "\\.elc$" t)))
   (byte-recompile-directory package-user-dir 0 arg))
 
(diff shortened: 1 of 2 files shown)


smile13241324:
Thank you for contributing to spacemacs :purple_heart:, I have cherry-picked your commit into develop.

PR #12893, opened 2019-11-06
Transient states wording etc

duianto:
- Removed the word `Manipulation` from the transient state titles:
  - Buffer Manipulation Transient State and Window Manipulation TS
  - Expanded the TS abbreviation to Transient State
- Documented the Buffer Transient State key binding:
  `q` to quit the transient state
- Capitalized `Transient State` in the titles
- Window Transient State:
  - Added a line break after the title in the folded hint.
    This allows for more keys to be visible horizontally.
  - Moved up the [?] help key after the title, so that
    it's in the same place in both the folded and full
    transient state.

Renamed window transient state related variables and functions,
from: `*window-manipulation*`
to:   `*window*`.

---

I left it as two commits, so that the variable/function renaming commit can be reverted if any related issues are discovered. Although it's probably fine. I tested all the transient state key bindings and they still work as expected.

### DIFF
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -131,13 +131,13 @@
       - [[#joining-and-splitting][Joining and splitting]]
     - [[#window-manipulation][Window manipulation]]
       - [[#window-manipulation-key-bindings][Window manipulation key bindings]]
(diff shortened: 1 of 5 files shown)


smile13241324:
I have rebased your two commits onto develop. Lets see what happens.

PR #12906, opened 2019-11-09
fix warning: void definition (dap-bind-keys-for-mode)

Whiothes:
There is no `dap-bind-keys-for-mode` definition in dap-mode , so it always appears warning message


duianto:
The `c-c++` layer is being refactored in this PR:
[c-c++] Layer refactor #12869

The lines that are checking if the `dap` layer is used, are being removed:
https://github.com/syl20bnr/spacemacs/pull/12869/files#diff-b63521ba1c6916fb05ad60eabead50e2L287

syl20bnr:
Refactored PR has been pushed to develop branch.
Does it fix this issue ?

Whiothes:
Yes, it does, thanks a lot

PR #12909, opened 2019-11-09
gitignore: ignore org-journal.cache file

bhipple:
The org-journal package from the org layer is writing this cache file to more
quickly compute the next journal date. While it's configurable via
`org-journal-cache-file`, by default it goes into the `user-emacs-directory` and
results in a dirty git status on spacemacs.


smile13241324:

PR #12913, opened 2019-11-10
Add lsp backend support for Dockerfile

sei40kr:
# Description

Add lsp backend support for Dockerfile.

https://github.com/syl20bnr/spacemacs/pull/11320

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1446,6 +1446,7 @@ Other:
   - Added ~SPC a D m~ for =docker-machines= (thanks to Ramz)
 - Fixed: broken package declaration for dockerfile-mode
   (thanks to Maximilian Wolff)
+- Added LSP support
 **** Dotnet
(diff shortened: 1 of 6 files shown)


nixmaniack:
Do you have any reasoning why this needs to be a separate layer? 

IMO having a separate layer just for `Dockerfile` doesn't make much sense when the functionality could be part of the `docker` layer.

sdwolfz:
I agree with @nixmaniack, Dockerfiles have no meaning without docker so they should be bundled together.

sei40kr:
@nixmaniack @sdwolfz 
I've updated the PR!

sei40kr:
@nixmaniack @smile13241324 
Thank you for your kind review. I updated the PR, please review again.

smile13241324:
Thank you for contributing to spacemacs :purple_heart:, I have cherry picked your PR into develop.

PR #12929, opened 2019-11-12
Add LSP backend support for R

sei40kr:
# Description

Add LSP backend support for R.

11/15 FYI I've updated the PR and rename `ess-backend` to `ess-r-backend`.

### DIFF
--- a/layers/+lang/ess/config.el
+++ b/layers/+lang/ess/config.el
@@ -9,7 +9,12 @@
 ;;
 ;;; License: GPLv3
 
+(spacemacs|define-jump-handlers ess-r-mode)
+
 ;; Variables
+
+(defvar ess-r-backend 'lsp
+  "The backend to use for IDE features. Possible values are `ess' and `lsp'.")
 
 (defvar ess-assign-key nil
   "Call `ess-insert-assign'.")
--- a/layers/+lang/ess/funcs.el
+++ b/layers/+lang/ess/funcs.el
@@ -9,6 +9,91 @@
 ;;
 ;;; License: GPLv3
 
+
+;; R
+
+(defun spacemacs//ess-r-backend ()
+  "Returns selected backend."
+  (if ess-r-backend
+      ess-r-backend
+    (cond ((configuration-layer/layer-used-p 'lsp) 'lsp)
+          (t 'ess))))
+
+(defun spacemacs//ess-may-setup-r-lsp ()
+  "Conditionally setup LSP based on backend."
+  (when (eq (spacemacs//ess-r-backend) 'lsp)
+    (spacemacs//ess-setup-r-lsp)))
+
+(defun spacemacs//ess-setup-r-lsp ()
+  "Setup LSP backend."
+  (if (configuration-layer/layer-used-p 'lsp)
+      (lsp)
+    (message "`lsp' layer is not installed, please add `lsp' layer to your dotfile.")))
+
+
+;; Key Bindings
+
+(defun spacemacs//ess-bind-keys-for-mode (mode)
+  "Bind the keys in MODE."
+  (spacemacs/declare-prefix-for-mode mode "md" "debug")
+  (spacemacs/declare-prefix-for-mode mode "mD" "devtools")
+  (spacemacs/declare-prefix-for-mode mode "mDc" "check")
+  (spacemacs/declare-prefix-for-mode mode "mE" "extra")
+  (spacemacs/declare-prefix-for-mode mode "mh" "help")
+  (spacemacs/set-leader-keys-for-major-mode mode
+    "h" 'ess-doc-map              ;; help
+    "d" 'ess-dev-map              ;; debug
+    "D" 'ess-r-package-dev-map    ;; devtools
+    "E" 'ess-extra-map            ;; extra
+    ))
+
+(defun spacemacs//ess-bind-repl-keys-for-mode (mode)
+  "Set the REPL keys in MODE."
+  (spacemacs/declare-prefix-for-mode mode "ms" "repl")
+  (spacemacs/set-leader-keys-for-major-mode mode
+    ","  #'ess-eval-region-or-function-or-paragraph-and-step
+    "'"  #'spacemacs/ess-start-repl
+    "si" #'spacemacs/ess-start-repl
+    "ss" #'ess-switch-to-inferior-or-script-buffer
+    "sS" #'ess-switch-process
+    "sB" #'ess-eval-buffer-and-go
+    "sb" #'ess-eval-buffer
+    "sd" #'ess-eval-region-or-line-and-step
+    "sD" #'ess-eval-function-or-paragraph-and-step
+    "sL" #'ess-eval-line-and-go
+    "sl" #'ess-eval-line
+    "sR" #'ess-eval-region-and-go
+    "sr" #'ess-eval-region
+    "sF" #'ess-eval-function-and-go
+    "sf" #'ess-eval-function))
+
+(defun spacemacs/ess-bind-keys-for-julia ()
+  (spacemacs//ess-bind-keys-for-mode 'ess-julia-mode)
+  (spacemacs//ess-bind-repl-keys-for-mode 'ess-julia-mode))
+
+(defun spacemacs/ess-bind-keys-for-r ()
+  (when ess-assign-key
+    (define-key ess-r-mode-map ess-assign-key #'ess-insert-assign))
+
+  (spacemacs//ess-bind-keys-for-mode 'ess-r-mode)
+  (spacemacs//ess-bind-repl-keys-for-mode 'ess-r-mode))
+
+(defun spacemacs/ess-bind-keys-for-inferior ()
+  (define-key inferior-ess-mode-map (kbd "C-j") #'comint-next-input)
+  (define-key inferior-ess-mode-map (kbd "C-k") #'comint-previous-input)
+  (when ess-assign-key
+    (define-key inferior-ess-r-mode-map ess-assign-key #'ess-insert-assign))
+
+  (spacemacs/declare-prefix-for-mode 'inferior-ess-mode "ms" "repl")
+  (spacemacs/declare-prefix-for-mode 'inferior-ess-mode "me" "eval")
+  (spacemacs/declare-prefix-for-mode 'inferior-ess-mode "mg" "xref")
+  (spacemacs/set-leader-keys-for-major-mode 'inferior-ess-mode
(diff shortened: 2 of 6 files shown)


gdkrmr:
Thanks for this! I tried it and it works, except, that `lsp-mode` interferes with the normal autocompletion.

Also I have the following wishes:

- Add short cuts for the capabilities (https://github.com/REditorSupport/languageserver#services-implemented).

Because it often provides more noise than useful information and consumes tons of cpu:
- Add a short cut to toggle the language server on a per buffer basis (I propose `SPC m E t`, AFAIK there is no convention for this, maybe some of the devs may have a better proposal).
- Disable it by default.

sei40kr:
I raised a PR for lsp-mode to add R language support.
https://github.com/emacs-lsp/lsp-mode/pull/1182

sei40kr:
@gdkrmr 
> Add a short cut to toggle the language server on a per buffer basis (I propose SPC m E t, AFAIK there is no convention for this, maybe some of the devs may have a better proposal).
I tried to implement it.
But lsp-mode changes the company backends and flycheck checker.
It can't be reverted after disabling `lsp-mode`, `lsp-ui-mode`, `lsp--managed-mode`.

gdkrmr:
do you get useful autocompletion with this?

sei40kr:
@gdkrmr 
lsp-r-completion
works great for me.

gdkrmr:
I get no autocompletion at all, the only config I have is:
```
     (ess :variables
          ess-style 'RStudio
          ess-use-flymake nil
          ess-assign-key "\M--")
```

sei40kr:
@gdkrmr 
You need set `ess-r-backend` to `'lsp`
```
(ess :variables
    ess-r-backend 'lsp
    ess-style 'RStudio
    ess-use-flymake nil
    ess-assign-key "\M--")
```

gdkrmr:
ok thanks, I also had to update `lsp-mode`, now it works. I will give it a spin. Maybe there should be a backend `'ess` that uses the R process buffer for autocompletion, not sure how much work that would be.

gdkrmr:
`lsp-mode` overwrites som existing keybindings: 
- `SPC m r r` was used for `inferior-ess-reload`. I think the best would be to move the entire `+extra` group away from `SPC m r`, not sure what would be good prefix, maybe some of the devs can give a suggestion.
- The other case I found was `SPC m h h` which is overwritten but is not a big deal, because this is defined as `SPC m h d` anyway. This should probably stay under `SPC m h` because it all has do do with finding help.

sei40kr:
@gdkrmr 

> * Do you suggest using this as a complement to ess or without ess?

`ess` means auto-completion backend using R process? If so, I suggest use LSP without it.

> * The help buffer integrates much better with spacemacs than the ess help buffers. Is there a way to search all symbols via a helm buffer, as in ess-display-help-on-object?

`helm-lsp-workspace-symbol` should work, but it somehow didn't work.
https://github.com/REditorSupport/languageserver
AFAIS, `workspaceSymbolProvider` is implemented...

> * Does lsp-mode provide interaction with repls? I know they have an interface for debuggers.

No.

> * Is there a way to not use the lintr part or at least run it only when you save a file, this makes my fan spin constantly on larger projects and is not good for battery life. Also many of the default suggestions of lintr are not really useful or plain wrong.

```elisp
(setq lsp-ui-flycheck-live-reporting nil
          flycheck-check-syntax-automatically '(save))
```

> * I miss an automatic completion of directory names inside strings.

I agree. Perhaps we would be better to use `company-lsp` with `company-files` together, like Path Autocomplete in VSCode:
https://marketplace.visualstudio.com/items?itemName=ionutvmi.path-autocomplete

```elisp
(add-hook 'ess-r-mode-hook #'(lambda ()
  (setq company-backends '(company-lsp company-files))))
```

sei40kr:
## Remaining Tasks

* [ ] Disable R backend completions when using LSP
* [ ] Resolve key binding conflicts

gdkrmr:
Isn't the R backend completion already disabled?

smile13241324:
Setting company-files should already be done, by the autocompletion layer. See below list of backends which are bound for all modes:

```elisp
(defvar spacemacs-default-company-backends
  '((company-semantic company-dabbrev-code company-gtags company-etags company-keywords)
    company-files company-dabbrev)
  "The list of default company backends used by spacemacs.
This variable is used to configure mode-specific company backends in spacemacs.
Backends in this list will always be active in these modes, as well as any
backends added by individual spacemacs layers.")
```
In there you should see company-files so directory completion should already be working.

layers/+lang/ess/config.el
```diff
@@ -9,7 +9,12 @@
 ;;
 ;;; License: GPLv3
 
+(spacemacs|define-jump-handlers ess-r-mode)
+
 ;; Variables
 
+(defvar ess-r-backend nil
```
smile13241324: nil is bad default for ess-r-backend please set `LSP` instead.

layers/+lang/ess/funcs.el
```diff
@@ -9,6 +9,42 @@
...
+(defun spacemacs//ess-r-setup-company-ess ()
+  "Setup ESS company."
+  (spacemacs|add-company-backends
+    :backends (company-R-library company-R-args company-R-objects :separate)
+    :modes ess-r-mode inferior-ess-r-mode))
+
+

+;; REPL
+
```
smile13241324: Please add a REPL declaration to the layer similiar to what has been done in the packages.el in the python layer.
```elisp
      (spacemacs/register-repl 'python
                               'spacemacs/python-start-or-switch-repl "python")
```

gdkrmr:
> Setting company-files should already be done, by the autocompletion layer.

I don't get directory/file completion, how do I use `company-files`? Ideally it should complete automatically, when you are typing a string.

smile13241324:
@gdkrmr that doesn't look right in general this should be a list of lists of which each list gets a request for completion, when a request can be satisfied a value is returned, if not the next backend is called in turn. 

If this is set to a single value literally the completion system is broken.

Basically when running:
```elisp
  (spacemacs|add-company-backends
    :backends (company-files company-capf)
    :modes inferior-python-mode
    :variables
    company-minimum-prefix-length 0
    company-idle-delay 0.5)
```
This sets company-files company-capf as well as all standard backends for the `inferior-python-mode`. If only this is set `company-diag` should show you something like this:
```elisp
company-backends: ((company-lsp :with company-yasnippet)
 (company-semantic company-dabbrev-code company-gtags company-etags company-keywords :with company-yasnippet)
 (company-files :with company-yasnippet)
 (company-dabbrev :with company-yasnippet))
```
Not sure why this is not working on this branch. If this is far enough to have the WIP flag removed I can have a look into this if you like.

gdkrmr:
Current issues for me:
- `company-backends` is still `(company-lsp)`. 
- All ess keybindings under `SPC m` are gone, which makes interaction with the REPL impossible.

sei40kr:
@gdkrmr @smile13241324 
I updated this PR:
1. Key binding conflicts
-> Resolved. Some leader keys in ESS has been changed.
2. File path completion
-> This should not be handled our side. LSP should provide it.
-> for an instance, TypeScript language server already does it.

Please review.

sei40kr:
@smile13241324 

This is not related to this PR, but I found 2 problems:
- `ess-first-continued-statement-offset` and `ess-continue-statement-offset` are now obsolete.
- ESS predefined keymap contains redundant key bindings.

gdkrmr:
I tried it out a bit and I like the new keybindings for R.
There are still some issues I encountered:
- The lsp keymap is gone (i cannot turn of the documentation overlay, no renaming of symbols)
- `K` doesn't show help on symbols any more, I think this is a `spacemacs` convention

layers/+lang/ess/funcs.el
gdkrmr: https://github.com/sei40kr/spacemacs/blob/ffee1c58d560c6037325de4788c2b6ed1931ec61/layers/%2Blang/ess/packages.el#L75

This is currently done here, should this be moved?

gdkrmr:
Please don't forget this PR, from my side the only open question is adding `K` and the lsp-keymap.

smile13241324:
@sei40kr Hi there, sorry I am bit busy at the moment, I will try to have a look at your PR this or next weekend. Sorry for the delay.

smile13241324:
@gdkrmr I agree with your comment about not having the repl ask for a project path everytime. However in my opinion this does not need to be in the intial layer contribution.

Lets just keep things separate and open a new PR for the path improvement.

smile13241324:
@sei40kr found and fixed a small issue with the company backends. Right now it was always using the bindings for non LSP, as for ess-r-mode backends have been set unconditionally. 

Normally I would vote to not set a backend in the lsp case which falls back to `company-lsp` however this seems to also omit the default backends including file completion. So for now I have set `company-lsp` manually as a backend in case of lsp omitting the non lsp ones. This adds lsp completion as well as the default backend including file completion.

smile13241324:

biocyberman:
@sei40kr Since this is still in the develop branch I would like to ask here if it is OK. `lintr` is giving too many messages while I navigate around my code, which was written before enabling `LSP`. It creates lagging for code navigation, especially when I  work on a remote server. Is there a way to disable some linting level or only show when I ask for it (preferably with a keybinding)?

gdkrmr:
You can tune the warnings that `lintr`gives by modifying `flycheck-lintr-linters`.

pietrodito:
@smile13241324 Why does this PR is closed? You said:

> Thank you for contributing to spacemacs purple_heart, I have cherry picked your commit into develop.

smile13241324:
@pietrodito we do not always auto merge a PR, which we only do for trivial stuff, the rest is cherry picked manually. In this case one of us pushes the commit and close the PR. In the other case github does the merge and the PR is put into merged state.

pietrodito:
@smile13241324 thanks for the answer but I do not get how it works. I have pulled this PR on my local repo, merged it by myself and I am happy with the autocompletion. But, I cannot have the same behavior when I used the bare develop branch after you have cherry picked the PR. I am probably missing something here...

smile13241324:
Not sure where your issue comes from, this could either be a setup thing or something is broken.

Here are the original commits:
539e95422 Fix R lsp company setup
ee5a5e8b7 Add LSP backend support for R

I have noticed that my code which was manually loading company-lsp is gone now. I think there had been a commit for removing these as lsp mode itself should set it up. Could still be that this change is the reason for what you are seing. @gdkrmr can you confirm that autocompletion with R lsp is broken?

PR #12930, opened 2019-11-12
Add dired evil search next/prev key bindings

duianto:
The search next/previous commands are different
because of the `evil-search-module` values:
`vim`    = `evil-search`
`hybrid` = `isearch`

- With the `vim` editing style in `normal` state:
  - `n` `evil-ex-search-next`
  - `N` `evil-ex-search-previous`
- With the `hybrid` editing style in `normal` state:
  - `n` `evil-search-next`
  - `N` `evil-search-previous`

---

`N` was bound to `dired-man`:
>dired-man is an interactive compiled Lisp function in ‘dired-x.el’.
> 
> It is bound to N.
> 
> (dired-man)
> 
> Run ‘man’ on this file.

I don't know if/where it could be rebound to.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -847,6 +847,13 @@ Other:
   - Added visual-line-navigation keys (thanks to duianto)
     - ~up~ =evil-previous-visual-line=
     - ~down~ =evil-next-visual-line=
+  - Added =dired= evil search next/prev key bindings (thanks to duianto):
+    - With the =vim= editing style in =normal= state:
+      - ~n~ =evil-ex-search-next=
+      - ~N~ =evil-ex-search-previous=
+    - With the =hybrid= editing style in =normal= state:
+      - ~n~ =evil-search-next=
+      - ~N~ =evil-search-previous=
 - Improvements:
   - Rewrote window layout functions for ~SPC w 1~, ~SPC w 2~, ~SPC w 3~, and
     ~SPC w 4~ (thanks to Codruț Constantin Gușoi):
--- a/layers/+spacemacs/spacemacs-defaults/packages.el
+++ b/layers/+spacemacs/spacemacs-defaults/packages.el
@@ -75,7 +75,16 @@
     "ad" 'spacemacs/dired
     "fj" 'dired-jump
     "jd" 'dired-jump
-    "jD" 'dired-jump-other-window))
+    "jD" 'dired-jump-other-window)
+  ;; The search next/previous commands are different
+  ;; because of the `evil-search-module' values:
+  ;; vim = evil-search, hybrid = isearch
+  (when (eq 'vim dotspacemacs-editing-style)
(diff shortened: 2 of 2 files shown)


layers/+spacemacs/spacemacs-defaults/packages.el
```diff
@@ -75,7 +75,13 @@
...
     "fj" 'dired-jump
     "jd" 'dired-jump
-    "jD" 'dired-jump-other-window))
+    "jD" 'dired-jump-other-window)
+  (when (eq 'vim dotspacemacs-editing-style)
+    (evil-define-key 'normal dired-mode-map (kbd "n") 'evil-ex-search-next)
+    (evil-define-key 'normal dired-mode-map (kbd "N") 'evil-ex-search-previous))
+  (when (eq 'hybrid dotspacemacs-editing-style)
+    (evil-define-key 'normal dired-mode-map (kbd "n") 'evil-search-next)
+    (evil-define-key 'normal dired-mode-map (kbd "N") 'evil-search-previous)))
```
nixmaniack: How about something like following?
```elisp
(when (memq dotspacemacs-editing-style '(hybrid vim))
    (evil-define-key 'normal dired-mode-map (kbd "n") 'evil-ex-search-next)
    (evil-define-key 'normal dired-mode-map (kbd "N") 'evil-ex-search-previous))
```

CHANGELOG.develop
```diff
@@ -847,6 +847,13 @@ Other:
   - Added visual-line-navigation keys (thanks to duianto)
     - ~up~ =evil-previous-visual-line=
     - ~down~ =evil-next-visual-line=
+  - Added =dired= evil search next/prev key bindings (thanks to duianto):
+    - With the =vim= editing style in =normal= state:
+      - ~n~ =evil-ex-search-next=
+      - ~N~ =evil-ex-search-previous=
+    - With the =hybrid= editing style in =normal= state:
```
nixmaniack: This could be clubbed in one block.
```org
With the =vim=/=hybrid= editing style in =normal= state:
```

layers/+spacemacs/spacemacs-defaults/packages.el
duianto: I should have specified the reason for the different commands.

The `evil-search-module` values are different for both editing styles:
- `vim` = `evil-search`
- `hybrid` = `isearch`

With `isearch` the command `evil-ex-search-next` returns:
>Search failed: "No search pattern"

And it doesn't work the other way around, `evil-search-next` with the `vim` editing style returns:
>"'vim": pattern not found

(`'vim` was the last search pattern when the `evil-search-module` value was `isearch` (hybrid))

I'll add a comment about the `evil-search-module` difference above the key bindings, 
and to the commit message, to make it clear why they use different commands.

nixmaniack: I had missed the reasoning. This clarifies.

smile13241324:

PR #12910, opened 2019-11-09
[multiple-cursors] Add evil-mc keys, cursors from selection

duianto:
Added evil-mc make cursors from selection key bindings:
`grI` calls `evil-mc-make-cursor-in-visual-selection-beg`
`grA` calls `evil-mc-make-cursor-in-visual-selection-end`

PR #12934, opened 2019-11-13
Improve multiple-cursors layer

sei40kr:
* Fix evil-mc-mode won't start automatically
* Add evil-mc key bindings to `evil-mc-mode`, not global keymap
* Undo all cursors by escape in normal state

May conflict with https://github.com/syl20bnr/spacemacs/pull/12910

### DIFF
--- a/layers/+misc/multiple-cursors/packages.el
+++ b/layers/+misc/multiple-cursors/packages.el
@@ -18,6 +18,10 @@
 
 (defun multiple-cursors/init-evil-mc ()
   (use-package evil-mc
+    :init
+    (progn
+      (add-hook 'prog-mode-hook 'turn-on-evil-mc-mode)
+      (add-hook 'text-mode-hook 'turn-on-evil-mc-mode))
     :config
     (progn
       (setq-default evil-mc-one-cursor-show-mode-line-text nil)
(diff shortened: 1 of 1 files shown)


duianto:
Thanks for the suggestions, it's easier to exit from multiple evil-mc cursors with `Esc` 👍 

One thing that might be surprising is that the cursors also exit if one just want to cancel (`Esc`) out of a started operation: `c`, `d` or `y`

Maybe `evil-iedit-state` (`SPC s e`) can be used as a template for finding a solution, because `Esc` cancels a started operation, before exiting the `evil-iedit-state` when `Esc` is pressed a second time.

---

There's an unrelated upstream issue with trying to cancel from `c` or `d` with `C-g`. The messages buffer gets flooded a lot of commands, which causes navigation in the messages buffer to be slow.

It also happens without Spacemacs, so I'll open an issue upstream.

sei40kr:
@duianto Thank you for kindly review.

I tried:
```elisp
(unless (eq evil-state 'operator)
  (evil-mc-undo-all-cursors))
```
This works for `d`, but still buggy for `y` and `c`, so I decided to that commit from this PR.

duianto:
The PRs changes have been cherry picked into the `develop` branch. https://github.com/syl20bnr/spacemacs/commit/6556cfc80d4ba070c7385f44f932fbe2b97c0663

leezu:
Hi, this breaks magit (https://github.com/syl20bnr/spacemacs/issues/12961). Please revert.

duianto:
>Fix evil-mc-mode won't start automatically

@sei40kr Where did evil-mc-mode not start automatically?

@leezu Let's continue the discussion about the magit issue in #12961

sei40kr:
@leezu @duianto 
https://github.com/syl20bnr/spacemacs/pull/12969
Sorry I created a reverting PR.

duianto:
There's nothing to be sorry about. I'm just wondering where you saw `evil-mc-mode` not working from the `:config` section, and it started working from the `:init` section?

duianto:
@sei40kr The `evil-mc` package has been updated to not flood the messages buffer with error messages when canceling an evil operation.

You can open a new PR to cancel the evil-mc cursors with `Esc` if you want.

The benefit of removing the cursors with `Esc` probably outweighs the unexpected behavior of the cursors being removed when canceling an evil operation.

If it's used by more people, then someone else might be able to figure out how to fix it.

sei40kr:
@duianto 
> There's nothing to be sorry about. I'm just wondering where you saw `evil-mc-mode` not working from the `:config` section, and it started working from the `:init` section?

I tried to move them back to `:init`, and then I confirmed it works, so probably I had an wrong idea.

duianto:
There might actually have been cases where evil-mc didn't load automatically.

The issue:
`P` wants to paste in magit buffer #12961
was caused when `(global-evil-mc-mode 1)` was called in the user-config section.

Hopefully the loading issue was fixed with this PR.

PR #12937, opened 2019-11-13
Don't assume the previous binding was evil-insert-digraph

lovrolu:
Restore the temporarily replaced keybinding with the one that was active at the time of replacement.

This kept rebinding `C-k` which I have unbound in my config using `(define-key evil-insert-state-map (kbd "C-k") nil)` whenever Company completion popped up.


duianto:
The PRs changes have been cherry picked into the `develop` branch. https://github.com/syl20bnr/spacemacs/commit/bc22a259e87ed6ff3a1fcdb83dd27ffacad77a15

PR #12942, opened 2019-11-14
Gnus application start

equwal:
Add keybindings for Gnus startup, since it is common to need to start it in "slave" mode (for more than one Gnus to run simultaneously) or "unplugged" mode (for offline queueing of mail to send later). Also adds a binding in message mode for FLAME ON, in RFC1855.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -648,6 +648,13 @@ Other:
   - Added =symbol-overlay= to the =spacemacs-navigation= layer
     (thanks to kenkangxgwe)
 - Key bindings:
+  - GNUs keybindings now include *unplugged* and *slave* variants. These are for
+    offline usage and having multiple gnus running, respectively.
+    - ~SPC a g g~ gnus (normal). Not slave, plugged.
+    - ~SPC a g o~ slave & unplugged.
+    - ~SPC a g u~ unplugged (but not slave).
+    - ~SPC a g s~ gnus slave (but plugged).
+  - GNUs layer now includes the message mode insert binding: ~SPC m i F~ for FLAME ON, following [[https://tools.ietf.org/html/rfc1855][RFC 1855]].
   - New evil text objects =«=, =｢=, =‘= and =“=.
   - Improved buffer transient state with extra bindings and new commands:
     - Added ~<right>~ for next-buffer
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -648,14 +648,7 @@ Other:
   - Added =symbol-overlay= to the =spacemacs-navigation= layer
     (thanks to kenkangxgwe)
 - Key bindings:
-  - GNUs keybindings now include *unplugged* and *slave* variants. These are for
-    offline usage and having multiple gnus running, respectively.
-    - ~SPC a g g~ gnus (normal). Not slave, plugged.
-    - ~SPC a g o~ slave & unplugged.
-    - ~SPC a g u~ unplugged (but not slave).
-    - ~SPC a g s~ gnus slave (but plugged).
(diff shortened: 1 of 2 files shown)


equwal:
The test need to be reran by someone with write permission.

layers/+email/gnus/packages.el
```diff
@@ -30,7 +30,16 @@
     :defer t
     :commands gnus
     :init
-    (spacemacs/set-leader-keys "ag" 'gnus)
+    (progn (spacemaccs/declare-prefix "ag" "+gnus" "Gnus newsreader")
```
duianto:
```suggestion
    (progn (spacemacs/declare-prefix "ag" "gnus" "Gnus newsreader")
```
Typo: double `c`
This results in an error message on startup:
>Error (use-package): gnus/:init: Symbol’s function definition is void: spacemaccs/declare-prefix

The `+` before `gnus`, results in two plus signs `++gnus` in the which-key popup.
Which-key adds the `+` automatically for prefixes.

CHANGELOG.develop
```diff
@@ -648,6 +648,13 @@ Other:
   - Added =symbol-overlay= to the =spacemacs-navigation= layer
     (thanks to kenkangxgwe)
 - Key bindings:
+  - GNUs keybindings now include *unplugged* and *slave* variants. These are for
```
duianto: Moved the changelog entries to the GNUS section:
https://github.com/syl20bnr/spacemacs/blob/develop/CHANGELOG.develop#gnus

And made some minor rewrites:
```
  - Added *unplugged* and *slave* variants. These are for offline usage and
    for having multiple gnus running, respectively.
    - ~SPC a g g~ gnus (normal). Not slave, plugged.
    - ~SPC a g o~ slave & unplugged.
    - ~SPC a g u~ unplugged (but not slave).
    - ~SPC a g s~ gnus slave (but plugged).
      (thanks to Spenser Truex)
  - Added message mode insert binding: ~SPC m i F~ for FLAME ON, following
    [[https://tools.ietf.org/html/rfc1855][RFC 1855]]. (thanks to Spenser Truex)
```

layers/+email/gnus/packages.el
```diff
@@ -71,6 +80,16 @@
...
       (require 'browse-url)
       (require 'nnrss)
+      (defun spacemacs/gnus-flame-on ()
+        "Most important email function, for RFC1855 compliance."
+        ;; https://tools.ietf.org/html/rfc1855
+        (interactive)
+        (insert "FLAME ON:
+")
+        (insert "FLAME OFF
+")
```
duianto: It looks a bit off to have the close quote at the beginning of the next line.
It might be better to write the newline characters as `\n`.

```elisp
        (insert "FLAME ON:\n")
        (insert "FLAME OFF\n")
```
Should there also be a semicolon after `OFF`?

equwal: I think no, here is the [relevant page of RFC1855](https://tools.ietf.org/html/rfc1855#page-4).

equwal: > It looks a bit off to have the close quote at the beginning of the next line.
> It might be better to write the newline characters as `\n`.
> 
> ```emacs-lisp
>         (insert "FLAME ON:\n")
>         (insert "FLAME OFF\n")
> ```
Done. Good feedback once again.

CHANGELOG.develop
equwal: These are moved in the next commit.

duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/5d1cf35e48343fe6def1412c62694d73b4edaf17

equwal:
Huzzah! Thank you for taking the time to review my code.

PR #12943, opened 2019-11-15
An attempt to bring some order in the "jump back from definition" mess

magthe:
These changesets only affects the layers for languages I use regularly.

I've noticed that while "SPC m g g" (or ", g g") seems to _always_ jump to the definition of the symbol under the cursor, there is a lack of a common key combination to jump back.

Taking the `lsp` layer as inspiration I've used "SPC m g p" for jumping back:

- haskell was "SPC m g b" (in Intero only)
- elisp was missing
- clojure was "SPC m g b"
- python was "SPC m g b"

I am aware that this will impact users of the `gtags` layer, however since `lsp` already used "gp" I assume that it's all right. I guess the argument would be that `gtags` is more of a backup and if the language-specific layer has direct support for jumping to definition then that's preferred over `gtags`.

### DIFF
--- a/layers/+lang/emacs-lisp/packages.el
+++ b/layers/+lang/emacs-lisp/packages.el
@@ -136,6 +136,8 @@
       (add-hook 'emacs-lisp-mode-hook 'elisp-slime-nav-mode)
       (dolist (mode '(emacs-lisp-mode lisp-interaction-mode))
         (spacemacs/declare-prefix-for-mode mode "mg" "find-symbol")
+        (spacemacs/set-leader-keys-for-major-mode mode
+          "gb" 'xref-pop-marker-stack)
         (spacemacs/declare-prefix-for-mode mode "mh" "help")
         (spacemacs/set-leader-keys-for-major-mode mode
           "hh" 'elisp-slime-nav-describe-elisp-thing-at-point)
--- a/layers/+lang/haskell/packages.el
+++ b/layers/+lang/haskell/packages.el
@@ -140,6 +140,7 @@
     (progn
       (dolist (mode haskell-modes)
         (spacemacs/set-leader-keys-for-major-mode mode
+          "gb" 'xref-pop-marker-stack
           "ht" 'dante-type-at
           "hT" 'spacemacs-haskell//dante-insert-type
           "hi" 'dante-info
--- a/layers/+tools/lsp/funcs.el
+++ b/layers/+tools/lsp/funcs.el
@@ -83,7 +83,7 @@
     (concat prefix-char "d") #'xref-find-definitions
     (concat prefix-char "r") #'xref-find-references
     (concat prefix-char "e") #'lsp-treemacs-errors-list
-    (concat prefix-char "p") #'xref-pop-marker-stack)
+    (concat prefix-char "b") #'xref-pop-marker-stack)
   (if (configuration-layer/package-usedp 'helm)
       (spacemacs/set-leader-keys-for-minor-mode 'lsp-mode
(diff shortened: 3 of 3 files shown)


duianto:
It seems like most layers use `"gb"` for `jump-back` or `pop-*-stack`
```

"gb"
./layers/+distributions/spacemacs-bootstrap/packages.el:111:  (evil-define-key 'motion help-mode-map (kbd  "gb") 'help-go-back)
./layers/+lang/c-c++/funcs.el:297:             "gB" 'rtags-show-rtags-buffer
./layers/+lang/clojure/packages.el:107:        "gb" 'cider-pop-back
./layers/+lang/common-lisp/packages.el:117:    "gb" 'slime-pop-find-definition-stack
./layers/+lang/d/packages.el:39:               "gb" 'company-dcd-goto-def-pop-marker
./layers/+lang/elixir/packages.el:113:         "gb" 'alchemist-goto-jump-back
./layers/+lang/haskell/packages.el:110:        "gb" 'xref-pop-marker-stack
./layers/+lang/html/packages.el:216:           "gb" 'web-mode-element-beginning
./layers/+lang/java/packages.el:146:           "gb" 'meghanada-back-jump
./layers/+source-control/git/packages.el:166:  "gb"  'spacemacs/git-blame-micro-state
./layers/+lang/nim/packages.el:52:             "gb" 'pop-tag-mark
./layers/+lang/ocaml/packages.el:104:          "gb" 'merlin-pop-stack
./layers/+lang/python/packages.el:63:          "gb" 'xref-pop-marker-stack
./layers/+lang/reasonml/packages.el:71:        "gb" 'merlin-pop-stack
./layers/+lang/scheme/packages.el:59:          "gb" 'geiser-pop-symbol-stack
./layers/+lang/typescript/packages.el:103:     "gb" tide-jump-back


"gp"
./layers/+lang/c-c++/funcs.el:309:            "gp" 'rtags-set-current-project
./layers/+lang/html/packages.el:218:          "gp" 'web-mode-element-parent
./layers/+tags/gtags/funcs.el:47:             "gp" 'counsel-gtags-go-backward
./layers/+tags/gtags/funcs.el:94:             "gp" 'helm-gtags-previous-history
./layers/+lang/scala/packages.el:95:          "gp"     'ensime-pop-find-definition-stack
./layers/+tools/transmission/packages.el:28:  "gp" 'transmission-peers
./layers/+tools/transmission/packages.el:45:  "gp" 'transmission-peers
./layers/+tools/transmission/packages.el:58:  "gp" 'transmission-peers
```

magthe:
I don't really care which is used in the end, I'd really just like there to be a combo in the layers where it's lacking and ideally the same one in all layers (at least the ones I use :smiling_imp:).

@duianto does your comment mean that if I change to using "gb" then the PR will be accepted? In that case I'll be happy to make the change, of course.

thanhvg:
Won't `c-o` work for you?

sdwolfz:
@magthe  I would have to say 'gb' is the convention I remember as well. Your PR should be alright if you change to that key binding.

@thabhvg 'C-o' is a workaround, and not a solution.

magthe:
@thanhvg `C-o` is a completely different beast, AFAIU.

magthe:
@sdwolfz changes switching to `gb` are coming up.

smile13241324:
Thank you for contributing to spacemacs :purple_heart:, I have squashed and cherry-picked your commit into develop.

duianto:
>does your comment mean that if I change to using "gb" then the PR will be accepted?

It was just an observation. And as you said, it's better to try to use the same key bindings in all layers for consistency.

PR #12957, opened 2019-11-17
notmuch improvements, key bindings

jb55:
These patches fix an annoyance when viewing github patches, and adds a view missing keybinds to notmuch-tree mode

### DIFF
--- a/layers/+email/notmuch/funcs.el
+++ b/layers/+email/notmuch/funcs.el
@@ -105,7 +105,7 @@ messages in the current thread"
          (browse-url url)))
       (diff-mode)
       (view-mode 1)
-      (pop-to-buffer (current-buffer)))))
+      (pop-to-buffer-same-window (current-buffer)))))
 
 (defun spacemacs/notmuch-show-open-github-patch ()
   "Open patch from GitHub email."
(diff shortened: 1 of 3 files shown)


jb55:
> I have noticed you have removed the notmuch evil keybinding hack, did you test that this is not necessary anymore?

yeah perhaps I don't understand what the point of it was, we should be using the standard N and P bindings from notmuch.

layers/+email/notmuch/funcs.el
```diff
@@ -105,7 +105,7 @@ messages in the current thread"
          (browse-url url)))
       (diff-mode)
       (view-mode 1)
-      (pop-to-buffer (current-buffer)))))
+      (switch-to-buffer (current-buffer)))))
```
Miciah: Would using the `pop-to-buffer-same-window` function achieve the desired effect?

jb55: I think this would be better..

jb55:
> I don't see any reason to squash my commits

perhaps I'll squash the documentation changes into their respective commits

smile13241324:
Thank you for contributing to spacemacs 💜, I have rebased your commits onto develop.

PR #12958, opened 2019-11-18
Fix handling of java-backend

yyoncho:
- Without this line the lsp-java package won't be enabled when java-backend is
nil and lsp layer is present


duianto:
The PRs changes have been cherry picked into the `develop` branch. https://github.com/syl20bnr/spacemacs/commit/38838cce9a70c7c21ada542d8e7c778369bb0d0c

PR #12960, opened 2019-11-19
[treemacs] Fix use of treemacs-use-collapsed-directories.

Alexander-Miller:
Treemacs sets its collapse-dirs value at load time, so the spacemacs layer
has to override it the :config section of treemacs' package declaration.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2822,6 +2822,8 @@ Other:
 - Made sure treemacs' minor modes will be disabled when they are configured
   to be off.
 - Add =tag= as an accepted value for option =treemacs-use-follow-mode=.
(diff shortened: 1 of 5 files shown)


Alexander-Miller:
I packed it in there when I didn't know that `:variables` can set any value and I wanted it to be like the use-minor-mode declarations. You're right that everything is in treemacs already - I'll change the PR to just update the readme.

Alexander-Miller:
Update pushed. `use-collapsed-directories` is removed and documented as such. I also updated the terminology - flattening is a better fit for what's happening and won't be confused with expaning/collapsing nodes in the file tree.

I also went ahead and got rid of that width locking message.

smile13241324:
Thank your for contributing to spacemacs :purple_heart:, I have squashed your commits and cherry picked them into develop.

Issue #12933, opened 2019-11-13
[Feature Request] Implement spacemacs|define-fmt-tool


sei40kr:
I want something like `spacemacs|define-fmt-tool`.

```lisp
(spacemacs|define-fmt-tool
    :id javascript-eslint-fix
    :command "eslint --stdin --fix"
)

(spacemacs/javascript-eslint-fix-format)
```

If we implement this, we will be able to support various formatters.

Issue #11704, opened 2018-12-10
json-reformat key binding is overridden


sei40kr:
As far as I see the source code, it seems that `json-reformat` key binding is overridden by `web-beautify` and `prettier`.

**To those who fix this problem:**
Please refactor json formatter.
https://github.com/syl20bnr/spacemacs/pull/11594

sei40kr:
Now json-reformat overrides Prettier's key bindings.

Issue #11580, opened 2018-11-11
[Feature Request] Add fixjson support


sei40kr:
I want `fixjson` in Spacemacs. There're no packages for Emacs yet.

`fixjson` corrects syntax errors automatically, and format whole the buffer.
The official repo has a GIF that shows how it works.
https://github.com/rhysd/fixjson

jrmiller82:
If there's no package for emacs yet, then I think your issue here may be a bit premature.

sei40kr:
https://github.com/syl20bnr/spacemacs/pull/12970

PR #12970, opened 2019-11-21
[core] Improve code formatting

sei40kr:
## Related Issues

- https://github.com/syl20bnr/spacemacs/issues/12933
- https://github.com/syl20bnr/spacemacs/issues/11580
-> supported in this PR
- https://github.com/syl20bnr/spacemacs/issues/11704
-> fixed in this PR

## Description

Add a new macro `spacemacs|define-fmt-tool` to make it easy for developers to add new formatters. There are also example usages for javascript and json layers.

e.g.
```lisp
(spacemacs|define-fmt-tool
  :id json-prettier
  :name "prettier"
  :program "prettier"
  :args (lambda () `("--stdin" "--stdin-filepath" ,buffer-file-name)))

(spacemacs/json-prettier-format)

(spacemacs|define-fmt-tool
  :id json-web-beautify
  :name "web-beautify"
  :program "js-beautify"
  :args ("-f" "-"))

(spacemacs/json-web-beautify-format)
```

It would be helpful if there're unnatural expressions in docstring since English is not my native tongue.

BTW  didn't use [reformatter](https://github.com/purcell/reformatter.el) because it also generates the unwanted `format-region` command even if the formatter don't support it.

### DIFF
--- a/core/core-format.el
+++ b/core/core-format.el
@@ -1,0 +1,146 @@
+;;; core-format.el --- Spacemacs Core File
+;;
+;; Copyright (c) 2012-2019 Sylvain Benner & Contributors
+;;
+;; Author: Seong Yong-ju <sei40kr@gmail.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+(defun spacemacs-format//format-goto-line (line)
+  "Move cursor to line LINE."
+  (goto-char (point-min))
+  (forward-line (1- line)))
+
+(defun spacemacs-format//format-apply-rcs-patch (patch-buffer)
+  "Apply an RCS-formatted diff from PATCH-BUFFER to the current buffer."
+  (let ((target-buffer (current-buffer))
+        ;; Relative offset between buffer line numbers and line numbers
+        ;; in patch.
+        ;;
+        ;; Line numbers in the patch are based on the source file, so
+        ;; we have to keep an offset when making changes to the
+        ;; buffer.
+        ;;
+        ;; Appending lines decrements the offset (possibly making it
+        ;; negative), deleting lines increments it. This order
+        ;; simplifies the forward-line invocations.
+        (line-offset 0))
+    (save-excursion
+      (with-current-buffer patch-buffer
+        (goto-char (point-min))
+        (while (not (eobp))
+          (unless (looking-at "^\\([ad]\\)\\([0-9]+\\) \\([0-9]+\\)")
+            (error "Invalid rcs patch or internal error in spacemacs-format//format-apply-rcs-patch"))
+          (forward-line)
+          (let ((action (match-string 1))
+                (from (string-to-number (match-string 2)))
+                (len  (string-to-number (match-string 3))))
+            (cond
+             ((equal action "a")
+              (let ((start (point)))
+                (forward-line len)
+                (let ((text (buffer-substring start (point))))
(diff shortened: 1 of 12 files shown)


sei40kr:
## Remaining Tasks

* [ ] Add documentation
* [ ] Prevent lsp layer from overriding format bindings
* [ ] Show error when LSP format tool is specified and lsp layer is not present
* [ ] Support the format tools won't support stdin (e.g. ESLint)

core/core-format.el
```diff
@@ -0,0 +1,146 @@
...
+A list of arguments or a function which evaluates to a list of arguments."
+  (let* ((id      (spacemacs/mplist-get-value props :id))
+         (name    (spacemacs/mplist-get-value props :name))
+         (program (spacemacs/mplist-get-value props :program))
+         (args    (spacemacs/mplist-get-value props :args))
+         (format-func-name (intern (format "spacemacs/%S-format" id))))
+    ;; define format function
+    `(defun ,format-func-name ()
+       (interactive)
+       (let* (,(if (functionp args) `(args (,args)) `(args ',args)))
```
smile13241324: I think your macro contains a memory anomaly, this is the variable `args` should be an uninterned symbol to avoid that a possible local variable `args` in the formating function is overwritten by the one in your macro.

See https://www.gnu.org/software/emacs/manual/html_node/elisp/Surprising-Local-Vars.html#Surprising-Local-Vars for a detailed explanation of the problem and how to resolve it.

layers/+lang/javascript/funcs.el
```diff
@@ -161,13 +167,45 @@
...
+    (call-interactively 'spacemacs/javascript-prettier-eslint-format))
+
    (t (error (concat "%s isn't valid javascript-fmt-tool value."
-                     " It should be 'web-beutify or 'prettier.")
-                     (symbol-name javascript-fmt-tool)))))
+                     " It should be 'web-beutify, 'eslint_d, 'prettier,"
+                     " 'prettier-eslint.")
+             (symbol-name javascript-fmt-tool)))))
+
+(spacemacs|define-fmt-tool
```
smile13241324: I think we should not setup formatting tools if they are not used. I think rather we should test for the selected formatter and only setup this one, similiar as to how company backends are handled.

layers/+lang/json/funcs.el
```diff
@@ -33,7 +53,32 @@ If ARG is a numerical prefix argument then specify the indentation level."
         (save-excursion (json-reformat-region (point-min) (point-max)))
       (json-reformat-region start end))))
 
-(defun spacemacs/json-setup-prettier ()
-  "Tell prettier the content is to be parsed as JSON regardless of any file
-extensions."
-  (setq-local prettier-js-args '("--parser=json")))
+(spacemacs|define-fmt-tool
```
smile13241324: Same here, formatter tools should only be created if really used.

sei40kr:
@smile13241324 
Let me close this because I'm not using Spacemacs now and I have no motivation to update this. Sorry.

layers/+lang/javascript/funcs.el
sei40kr: @smile13241324 
For this, I think it's preferable to leave it as it is so that the users can choose a different formatter for each project.
Of course, it would be great if we can define the formatters lazily, but I was not enough skilled to implement it.

smile13241324:
@sei40kr 
Sorry to hear that. Nevertheless thank you for all the high quality PRs you have made and have fun with the IDE solution you have choosen to run with now.

PR #12971, opened 2019-11-21
Window manipulation: Transient state includes ace-window movement

fleimgruber:
Adds `ace-window` movement to the window manipulation transient state.


smile13241324:
Thank you for contributing to spacemacs 💜, I have cherry picked your commit into develop. You can safely remove your branch.

PR #12976, opened 2019-11-22
Fix react layer fmt on save

tgroshon:
Fixes a function naming bug in the react layer so it will again start formatting files on save if the appropriate config variable is set.


smile13241324:

PR #12981, opened 2019-11-23
rust: fix LSP/DAP integration so it doesn't depend on java layer

tko:
Fixes error with rust files when java layer isn't loaded:
```
File local-variables error: (void-function spacemacs//java-setup-lsp-dap)
```


smile13241324:

PR #12990, opened 2019-11-25
Add function for changing font features.

### LINKED ISSUE #12966: Modeline bug when changing font size.
Changing a font size after loading Spacemacs messes up the mode line, since it is not able to resize itself dynamically. So upsizing your font requires restarting Emacs (and it should not).

A workaround is to scale it up to the largest font size you might use in a session (expanding the mode-line's height), then downsize after load.

image

# Reproduction
```emacs-lisp
;;; Run this after loading your Spacemacs
(set-face-attribute 'default nil :height (* 24 10))
```
# Workaround
```emacs-lisp
;;; Put this in a config.el somewhere in your layers.
(set-face-attribute 'default nil :height (* 24 10))
;;; Make sure that this thing runs after the mode line loads (eg. by keybinding it).
(set-face-attribute 'default nil :height (* 12 10))
```
**Edit: improved workaround**
Use:
```emacs-lisp
(defun spacemacs/set-face-attribute (face frame &rest attributes)
  "Set the FACE (eg. 'default) for the FRAME (eg. nil). See `set-face-attribute'.
ATTRIBUTES should be valid entries in `set-face-attribute'."
  (apply #'set-face-attribute face frame attributes)
  (powerline-reset))

;; point * 10
;; Fix bug with mode line: Scale it up before we scale down the font.
(spacemacs/set-face-attribute 'default nil :family "Hack"
                              :height (* 24 10))
```

equwal:
Adds the function suggested in #12966, and resolves #12966. Other refresh functions may be added in the future as needed, currently only the mode line needs refreshing.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -454,6 +454,8 @@ Other:
     (thanks to Muneeb Shaikh and Sylvain Benner)
 *** Core changes
 - Improvements:
+  - Function =spacemacs/set-face-attribute= sets fonts and resets anything that
+    relies on them (currently just the powerline mode line).
   - Display time spent in =user-config= in home buffer (thanks to Sylvain Benner)
   - Added dotfile function =dotspacemacs/user-load= (thanks to Sylvain Benner)
   - Added dotfile function =dotspacemacs/user-env= (thanks to Sylvain Benner)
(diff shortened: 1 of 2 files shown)


equwal:
I'm not sure this will go anywhere at this point.

PR #13000, opened 2019-11-29
Fix home buffer version and banner

duianto:
- Removed the first line deletion.
This required the banners to have an empty first line.

- Removed the empty first line from the banners.

- Made sure that there is a single empty line between:
  - The Version and Banner.
  - Banner and Buttons ([?] [Homepage] ...).
  - Version and Buttons (when the banner is hidden).

- Reordered the backend version insertion function call
to match the frontend Spacemacs buffer order.
  - version
  - banner
  - buttons

Before it added:
- the banner
- then the buffers first line was removed
- then the version was added to the first line
- then the buttons were added at the bottom of the buffer
(there are more things added after the buttons,
but they haven't been changed by this commit).

- Removed the version insertion call in:
core/core-spacemacs-buffer.el

The changes above caused the version to appear twice
in the Spacemacs home buffer on startup.

I'm not sure if it has/had other uses.

This seems to be where it was first added to:
core-configuration-layer.el:
core: restore default mode line in home buffer
77161bd

It was slightly modified in this commit:
core: defer distro insertion in home buffer
cd50d9c

Here it was joined with the banner:
Fix version injection in home buffer
  Don't inject version if banner is nil
a764eb4

core: condensed versions into one string in left-hand side
  spacemacs-version@emacs-version (distribution)
627e934

- I had not seen that the version was joined with the
banner on purpose until now.

It seemed useful to still be able to see the versions
if one just wanted to hide the banner.

Therefore I opened this PR (it's been cherry-picked):
Uncouple version from banner in spacemacs buffer #11901
#11901

If there's a good reason to combine them again,
then the changes can be reverted.

---

### Notes:
In the Github diff it looks like the banners 997 and 999 had some characters on the first line removed. But I checked and on the current develop branch the cursor doesn't move past the first column on the first line.


smile13241324:
Merged

PR #13004, opened 2019-12-01
[csharp] add support for the alternative `lsp' backend

razzmatazz:
This add support for the lsp package-based backend to `csharp` layer. A variable can be set when importing a layer to use an alternative (lsp) backend, in which case `lsp' layer is pulled in and omnisharp-emacs is not used.

I could not find a way to declare optional package dependency, like I did it for the `lsp` layer using  `configuration-layer/declare-layer-dependencies` – which would be nice to do to avoid pulling in the `omnisharp-emacs` package when using `lsp` backend.

### DIFF
--- a/layers/+lang/csharp/README.org
+++ b/layers/+lang/csharp/README.org
@@ -7,10 +7,11 @@
 * Table of Contents                     :TOC_5_gh:noexport:
 - [[#description][Description]]
   - [[#features][Features:]]
-- [[#packages-included][Packages Included]]
+- [[#backends][Backends]]
 - [[#install][Install]]
+  - [[#omnisharp-backend][Omnisharp backend]]
 - [[#caveats][Caveats]]
-- [[#key-bindings][Key bindings]]
+- [[#key-bindings-for-the-omnisharp-backend][Key bindings for the omnisharp backend]]
   - [[#navigation][Navigation]]
   - [[#helpers-documentation-info][Helpers (documentation, info)]]
   - [[#refactoring][Refactoring]]
@@ -18,11 +19,9 @@
   - [[#tests][Tests]]
 
 * Description
-This layer adds support for C# language using the [[https://github.com/OmniSharp/omnisharp-roslyn][omnisharp-roslyn]] language
-server and corresponding [[https://github.com/OmniSharp/omnisharp-emacs][omnisharp-emacs]] package.
-
-Please report any issues encountered to [[https://github.com/OmniSharp/omnisharp-emacs/issues][omnisharp-emacs issue page on github]].
-PRs are welcome too!
+This layer adds support for the C# language using the [[https://github.com/OmniSharp/omnisharp-roslyn][omnisharp-roslyn]]
+language server with either the [[https://github.com/OmniSharp/omnisharp-emacs][omnisharp-emacs]]
+or the [[https://github.com/emacs-lsp/lsp-mode][lsp-mode]] packages.
 
 ** Features:
 - Syntax checking with flycheck (when =syntax-checking= layer is used)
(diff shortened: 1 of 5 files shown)


layers/+lang/csharp/README.org
```diff
@@ -31,32 +30,51 @@ PRs are welcome too!
...
+the =csharp-backend= variable when configuring the layer:
+ - =omnisharp= (the default), or
+ - =lsp= (testing)
+
+To use the lsp backend, add `csharp' layer with =csharp-backend= variable set to 'lsp:
+#+BEGIN_SRC elisp
+  (csharp :variables csharp-backend 'lsp)
+#+END_SRC
+
+** omnisharp backend
```
smile13241324:
```suggestion
** Omnisharp backend
```

layers/+lang/csharp/config.el
```diff
@@ -1,4 +1,4 @@
-;;; packages.el --- C# Layer configuration File for Spacemacs
+;;; packages.el --- csharp Layer configuration File for Spacemacs
```
smile13241324: Please be consistent with the name either use C# everywhere or csharp

layers/+lang/csharp/funcs.el
```diff
@@ -0,0 +1,150 @@
...
+  (when (configuration-layer/layer-used-p 'lsp)
+    (progn
+      (spacemacs|add-company-backends
+        :backends company-lsp
+        :modes csharp-mode
+        :append-hooks nil
+        :call-hooks t)
+      (company-mode))))
+
+(defun spacemacs//csharp-configure-lsp ()
```
smile13241324: Please remove this function

layers/+lang/csharp/funcs.el
```diff
@@ -0,0 +1,150 @@
...
+  "Conditionally setup company based on backend."
+  (pcase (spacemacs//csharp-backend)
+    (`omnisharp (spacemacs//csharp-setup-omnisharp-company))
+    (`lsp (spacemacs//csharp-setup-lsp-company))))
+
+(defun spacemacs//csharp-configure ()
+  "Conditionally configure csharp layer based on backend."
+  (pcase (spacemacs//csharp-backend)
+    (`omnisharp (spacemacs//csharp-configure-omnisharp))
+    (`lsp (spacemacs//csharp-configure-lsp))))
```
smile13241324: Please remove this line, this call is unnecessary

layers/+lang/csharp/funcs.el
```diff
@@ -0,0 +1,150 @@
...
+;; lsp
+
+(defun spacemacs//csharp-setup-lsp ()
+  "Setup lsp backend."
+  (when (configuration-layer/layer-used-p 'lsp)
+    ;; Load omnisharp-mode with csharp-mode,
+    ;; this should start the omnisharp server automatically
+    (add-hook 'csharp-mode-hook #'lsp)))
+
+(defun spacemacs//csharp-setup-lsp-company ()
```
smile13241324: This is not needed when `(lsp)` is run it will automatically setup lsp-company for you, in the more recent versions. You only have to make sure that company-mode is activated.

layers/+lang/csharp/funcs.el
```diff
@@ -0,0 +1,150 @@
...
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+

+;; backend
+
+(defun spacemacs//csharp-backend ()
```
smile13241324: This function is not necessary just directly reference `csharp-backend`, the point is lsp layer is always loaded when `csharp-backend` is `lsp` this makes this getter superfluous.

layers/+lang/csharp/funcs.el
```diff
@@ -0,0 +1,150 @@
...
+      )
+    (spacemacs|hide-lighter omnisharp-mode)))
+
+

+;; lsp
+
+(defun spacemacs//csharp-setup-lsp ()
+  "Setup lsp backend."
+  (when (configuration-layer/layer-used-p 'lsp)
```
smile13241324: Please remove all tests against the usage of lsp layer in your code. You request lsp to be loaded in your layer file so you can be sure that it is available at the point of loading. So all these tests will be true.

layers/+lang/csharp/README.org
razzmatazz: fixed

layers/+lang/csharp/config.el
razzmatazz: I changed it on all files to `C#`

layers/+lang/csharp/funcs.el
razzmatazz: ok

layers/+lang/csharp/funcs.el
razzmatazz: ok

layers/+lang/csharp/funcs.el
razzmatazz: ok

layers/+lang/csharp/funcs.el
razzmatazz: I do believe it is useful, actually. Since csharp-backend is a nil'able variable and we want to default to *some* backend when resolving it our code.

layers/+lang/csharp/funcs.el
razzmatazz: OK

razzmatazz:
Hey @smile13241324 , thanks for a the review. I think I made all the changes you requested, except for maybe the `defun spacemacs//csharp-backend` part, where I think it is actually useful to have a "getter" for `csharp-backend` variable which may contain nil and which I cannot pcase against w/o adding a default case in each place.

razzmatazz:
@smile13241324 I thought a bit more about your comment on `csharp-backend` and rewrote my changes to the layer so that:
 - it is now possible to have no backend set (by setting `csharp-backend` to nil via `:variables`)
 - `spacemacs//csharp-backend` is no longer needed and I `pcase` directly agains `csharp-backend` as you have recommended

smile13241324:

PR #13005, opened 2019-12-01
pass 'result' as argument since lexical-binding is enabled

emacs18:
This patch would not be needed if dynamic binding is active.
However lexical binding is on for the file. 
I verified that the function being modified is called only within the file.


emacs18:
I think the fix will not work.  It appears that elisp function arguments don't work via "pass by reference" as in python.  I'll have to refresh my memory on how to do this in elisp.

PR #13007, opened 2019-12-02
Set VIRTUAL_ENV environment variable when switching to new pyenv

jg-made:
This environment variable is used by python LSP to determine
the pyenv virtual environment to use. When switching to a different
project using projectile 'switch-to-project', it should only be
necessary to restart the LSP server but a restart will only
work if this VIRTUAL_ENV changes value too.


smile13241324:

PR #13008, opened 2019-12-02
Fix typos

bwignall:
Should be non-semantic.

Uses https://en.wikipedia.org/wiki/Wikipedia:Lists_of_common_misspellings/For_machines to find likely typos.


duianto:
I checked the first four typos, and they seem to have been fixed on the `develop` branch already, so you'll have to compare which of them are still typos on the `develop` branch.

bwignall:
Apologies for that. I will compare against develop and submit a new PR if I find anything there. Thanks for your help and explanations.

PR #13012, opened 2019-12-03
use dynamic binding for "result" variable

emacs18:
Restore dynamic binding to one variable so that "result" variable can be accessed by spacemacs/window-state-walk-windows-1

Using defvar with just variable name is documented at
https://www.gnu.org/software/emacs/manual/html_node/elisp/Using-Lexical-Binding.html#Using-Lexical-Binding

### DIFF
--- a/layers/+spacemacs/spacemacs-layouts/funcs.el
+++ b/layers/+spacemacs/spacemacs-layouts/funcs.el
@@ -596,6 +596,7 @@ WINDOW is the representation of a window in a window-state object."
   "Execute FN once for each window in STATE and make a list of the results.
 FN is a function to execute.
 STATE is a window-state object."
+  (defvar result) ;; use dynamic binding
   (let (result)
     (spacemacs/window-state-walk-windows-1 (cdr state) fn)
     result))


layers/+spacemacs/spacemacs-layouts/funcs.el
```diff
@@ -596,6 +596,7 @@ WINDOW is the representation of a window in a window-state object."
   "Execute FN once for each window in STATE and make a list of the results.
 FN is a function to execute.
 STATE is a window-state object."
+  (defvar result) ;; use dynamic binding
   (let (result)
     (spacemacs/window-state-walk-windows-1 (cdr state) fn)
     result))
```
smile13241324: Wouldn't the same result be reached by changing the let part to:
```suggestion
     (let (result)
         (spacemacs/window-state-walk-windows-1 (cdr state) fn))
```
without having to introduce dynamic binding again?

smile13241324 (changes requested):
Hi @emacs18,

I have reviewed your code and I am not yet sure whether this function could not be fixed without having to introduce dynamic binding again. Please see my questions below.

BR
Max

layers/+spacemacs/spacemacs-layouts/funcs.el
emacs18: I don't follow what change you are suggesting.
Can you provide few more lines of context so that I can understand your proposed change?

emacs18: In case it wasn't clear, dynamic linking is to deal with line 593 which is the last line of spacemacs/window-state-walk-windows-1, i.e., the called function updates value of 'result' free variable.

smile13241324: Sorry @emacs18 I was a bit in a hurry when I did the review so here is the long version of my review.

I think I have grasped your idea, you are reestabling dynamic binding for `result` as it is set within `spacemacs/window-state-walk-windows-1` and currently is lost due to
```elisp
(defun spacemacs/window-state-walk-windows (state fn)
  "Execute FN once for each window in STATE and make a list of the results.
FN is a function to execute.
STATE is a window-state object."
  (let (result)
    (spacemacs/window-state-walk-windows-1 (cdr state) fn) ---> sets result but is lost
    result))  ---> result is nil
```
However dynamic binding introduces lots of issues within emacs so I was taking a long look at `spacemacs/window-state-walk-windows-1` and noticed that this function returns `result`. So my idea was to use
```elisp
(defun spacemacs/window-state-walk-windows (state fn)
  "Execute FN once for each window in STATE and make a list of the results.
FN is a function to execute.
STATE is a window-state object."
    (spacemacs/window-state-walk-windows-1 (cdr state) fn))  ---> returns result
```
instead. Could you give this a try? If it doesn't work we can still make result dynamically bound.

emacs18: spacemacs/window-state-walk-windows-1 calls itself recursively in a loop.
Each call updates 'result'.  Can you explain how this is supposed to be handled by your proposed change which I still don't understand?

emacs18: I don't understand your reluctance with dynamic binding.  Emacs still has many thousands of dynamic variables.  The use of 'result' is is really trivial and limited use of dynamic binding.  So can you explain why you think dynamic binding is any issue here?

emacs18: Following is  away to rewrite this to not use dynamic binding.
I don't see why this is any better.

```elisp
(defun spacemacs/window-state-walk-windows-1 (window fn)
  "Helper function for `spacemacs/window-state-walk-windows'."
  ;; WINDOW is a misleading name. WINDOW is a list that can represent a window,
  ;; or a concatenation of several windows. window-state objects are weird.
  (let ((child-windows
         (-filter #'spacemacs/window-state-window-p window))
        (bare-window
         ;; if WINDOW contains more than one window, take only the first window
         (--take-while (not (spacemacs/window-state-window-p it))
                       window))
        result)
    (--each child-windows
      (push (spacemacs/window-state-walk-windows-1 it fn) result))
    (push (funcall fn bare-window) result)))

(defun spacemacs/window-state-walk-windows (state fn)
  "Execute FN once for each window in STATE and make a list of the results.
FN is a function to execute.
STATE is a window-state object."
  (spacemacs/window-state-walk-windows-1 (cdr state) fn))
```

smile13241324:
@emacs18 thanks for checking, ok lets run with the dynamic variable binding for now. Thank you for contributing to spacemacs :purple_heart:, I have cherry picked your commit into develop.

Issue #13011, opened 2019-12-02
Symbol’s function definition is void: global-semantic-stickyfunc-mode


rakanalh:
#### Description :octocat:
if: Symbol’s function definition is void: global-semantic-stickyfunc-mode

#### Reproduction guide :beetle:
- Start Emacs
- Open python file
- Press SPC T C-s 

*Observed behaviour:* :eyes: :broken_heart:
- global-sematic-stickyfunc-mode enabled

*Expected behaviour:* :heart: :smile:
- The error in the title

#### System Info :computer:
- OS: gnu/linux
- Emacs: 26.3
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. f34ca06b0)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
(sql yaml lua javascript python rust solidity auto-completion better-defaults emacs-lisp git ivy markdown multiple-cursors org
     (shell :variables shell-default-height 30 shell-default-position 'bottom)
     syntax-checking treemacs version-control mu4e)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GSETTINGS GLIB NOTIFY ACL GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XDBE XIM MODULES THREADS LIBSYSTEMD LCMS2

#### Backtrace :paw_prints:
```
Debugger entered--Lisp error: (void-function global-semantic-stickyfunc-mode)
  (global-semantic-stickyfunc-mode)
  (if (spacemacs/toggle-semantic-stickyfunc-globally-status) (progn (global-semantic-stickyfunc-mode -1) (if (called-interactively-p (quote any)) (progn (message "semantic-stickyfunc-globally disabled.")))) (global-semantic-stickyfunc-mode) (if (called-interactively-p (quote any)) (progn (message "semantic-stickyfunc-globally enabled."))))
[...]

PR #13013, opened 2019-12-03
[semantic] Moved toggle keys to the layer

duianto:
`SPC T S` and `SPC T C-S` showed void variable
semantic-stickyfunc messages when the semantic
layer wasn't installed.

Fixes: #13011


smile13241324:
Merged

PR #13023, opened 2019-12-04
fix(+lang/typescript): correct readme typos

arjansingh:
Fixed a typo in the Typescript layer README.

### DIFF
--- a/layers/+lang/typescript/README.org
+++ b/layers/+lang/typescript/README.org
@@ -171,7 +171,7 @@ then set the variable =typescript-lsp-linter= to =nil=.
 | Key binding | Description                                                  |
 |-------------+--------------------------------------------------------------|
 | ~SPC m =~   | reformat the buffer                                          |
-| ~SPC m E d~ | add =tslitn-disable-next-line= at point                      |
+| ~SPC m E d~ | add =tslint:disable-next-line= at point                      |
 | ~SPC m E e~ | fix thing at point                                           |
 | ~SPC m g b~ | jump back                                                    |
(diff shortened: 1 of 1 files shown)


duianto:
The full command name seems to be longer:
https://github.com/syl20bnr/spacemacs/blob/42fc02bd480494dab5e8bfa25cd9bc15cf05fc6c/layers/%2Blang/typescript/packages.el#L109

Should the description be updated to show the full name, or do you happen to know if there is a reason to only mention the end of the command?

arjansingh:
So I read the README as intent on referring to (`tslint:disable-next-line`)[https://palantir.github.io/tslint/usage/rule-flags/], which is a specific flag that you can use to disable lint checking. That's why I referenced it instead of the actual `tide` command that it's running.

I did actually forget that the correct syntax is `tslint:disable-next-line` instead of `tslint-disable-next-line` so I will correct my PR to fix that.

If you still prefer, I'm happy to reference the actual package command. Let me know and i'll change that as well.

duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/4e9d62b80d672b3f562b315a56f855768367226e

You're correct, it adds a flag.

The `tide` package says this about the command:
>`M-x tide-add-tslint-disable-next-line` If the point is on one or more tslint errors, add a `tslint:disable-next-line` flag on the previous line to silence the errors. Or, if a flag already exists on the previous line, modify the flag to silence the errors.

https://github.com/ananthakumaran/tide#commands

I wasn't familiar with ts flags, therefore it looked like only part of the package command was referenced.

PR #13027, opened 2019-12-05
Fix typos

bwignall:
A reworking of https://github.com/syl20bnr/spacemacs/pull/13008 , from "develop" instead of "master".

Should be non-semantic.

Uses https://en.wikipedia.org/wiki/Wikipedia:Lists_of_common_misspellings/For_machines to find likely typos.


smile13241324:

PR #13032, opened 2019-12-07
Use lsp-disabled-clients instead of lsp-enabled-clients

yyoncho:
- lsp-mode supports defining remote versions of each clients(e. g. clangd which
works over tramp). Before this change the language server will be disabled
because it has different server-id.


smile13241324:

PR #13033, opened 2019-12-07
[html][vue] Improve Transient States

duianto:
Added a full/minified transient state toggle.

Grouped backend transient state key bindings by column.

Moved the html/init-web-mode :init section before the :config section.

---

### Before
before

#### Minified
after minified

#### Full
after full

### Notes
The transient state key bindings in both the `html` and `vue` layers are the same.
It's just the variables, functions and transient state title that use different names `web` vs `vue`.

It's redundant and any key binding modifications will have to be changed in two places (four when including the transient state listings for the keys).

It's probably possible to make a single source for both layers but since this seems to be working it might be better to implement this change now and refactor it in the future.


smile13241324:
Merged

PR #13034, opened 2019-12-08
Fix SPC-TAB in spacemacs-base distribution

duianto:
Problem:
The variable `spacemacs-layouts-restrict-spc-tab` is defined in:
`~/.emacs.d/layers/+spacemacs/spacemacs-layouts/config.el`
But the `spacemacs-layouts` layer isn't loaded by default in `spacemacs-base`.

Solution:
Check if the variable is bound.


smile13241324:
Merged

Issue #10041, opened 2017-12-28
c-c++ layer on develop c-c++-enable-clang-format-on-save is t, but not working


Jackson-soft:
![image](https://user-images.githubusercontent.com/3928114/34410565-278d238c-ec0c-11e7-8298-72b1b1719d9b.png)

Jackson-soft:
The problem lies in this function ：

(defun spacemacs//clang-format-on-save ()
  "Format the current buffer with clang-format on save when
`c-c++-enable-clang-format-on-save' is non-nil."
  (when c-c++-enable-clang-format-on-save
    (spacemacs/clang-format-buffer)))

Invoked a non-existent command "spacemacs/clang-format-buffer",and This command should be called
"clang-format-buffer"

myrgy:
@delaanthonio, could you please take a look?

delaanthonio:
@Jackson-soft Thanks for catching this error. I've submitted a pull request fixing this issue.

syl20bnr:
@delaanthonio Thank you!

Should be fixed in develop.

rdeterre:
Hi, I'm trying this with spacemacs' development branch (commit d4017b1c) and `c-c++-enable-clang-format-on-save` does not seem to have any effect.

The C/C++ layer is loaded with clang support:

```
     (c-c++ :variables
            c-c++-default-mode-for-headers 'c++-mode
            c-c++-enable-clang-support 't)
```
Also, `c-c++-enable-clang-format-on-save` is set to `t` in my `dotspacemacs/user-config()`.

When I open a C++ file, the `before-save-hook` does not contain "enable-clang-format-on-save". Its value on my machine is `(delete-trailing-whitespace spacemacs//python-sort-imports)`, because the python layer is active and `dotspacemacs-whitespace-cleanup` is set to `trailing`.

I tried removing both the Python layer and the `dotspacemacs-whitespace-cleanup` but `before-save-hook` becomes empty.

Is there something I'm not doing right ?

Thanks

vegansk:
My `before-save-hook` doesn't contain `spacemacs//clang-format-on-save`  too:

```
before-save-hook’s value is
(tide-format-before-save spacemacs//python-sort-imports gofmt-before-save)
Original value was nil
```

crazymousethief:
@delaanthonio Haven't merged into the develop branch yet? Format on save is a very useful feature, I can't wait😘

delaanthonio:
[...]

PR #13036, opened 2019-12-08
Fix clang-format-on-save

olishmollie:
Howdy!

Currently, `c-c++/init-clang-format` adds `spacemacs//c-c++-setup-format` to c-c++-mode-hooks. This has the effect of only automatically formatting c/c++ files on save _the second time_ I open a buffer, or I refresh the major mode with SPC-h-M.

In other words, the c-c++ layer adds a hook to add a hook when Spacemacs enters a c-c++-mode. This commit simply sets up clang format directly when initializing the package.

I'm not sure if this issue is affecting other Spacemacs users, but #10041 is still open, and I hope this can help everyone.

Thanks!

### DIFF
--- a/layers/+lang/c-c++/packages.el
+++ b/layers/+lang/c-c++/packages.el
@@ -78,8 +78,7 @@
 
 (defun c-c++/init-clang-format ()
   (use-package clang-format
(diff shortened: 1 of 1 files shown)


smile13241324:

PR #13037, opened 2019-12-09
Update funcs.el

JoyConX:
should  add  disable go-staticcheck


smile13241324:

PR #13039, opened 2019-12-09
[CI] Use fresh gpg key with Emacs25 tests

JAremko:
Hopefully it will make all tests green.


JAremko:
There is a better way :shipit:

PR #13047, opened 2019-12-10
Automatically choose the `lsp' as python-formatter when lsp-layer is enabled

sunlin7:
Hi,
The python-backend will automatically switch to lsp when lsp-layer is enabled, but the python-formatter dosen't present similar behavior. 
This PR will make python-formatter automatically switch to lsp when lsp-layer is enabled.

@smile13241324 Could you please review this PR? Thank you.


smile13241324:

PR #13049, opened 2019-12-10
[scala] Update metals binary version in README

nhurden:
Update the version of metals referenced in the metals installation instructions, so that an old version of the metals binary doesn't get installed by accident.


smile13241324:

PR #13050, opened 2019-12-10
[ivy] Add `SPC h m`: search available man pages

madand:
This keybinding is documented in DOCUMENTATION.org, yet was missing in ivy layer.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1971,7 +1971,7 @@ Other:
 - Added =recentf= alt actions to refresh and delete items (thanks to Rich Alesi)
 - Added =projectile= alternate actions to invalidate cache (thanks to Rich Alesi)
 - Key bindings:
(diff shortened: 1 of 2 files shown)


duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/239f60a65f16f0ab233627ced84756556c4f68e2

madand:
Lol, there was no need to expand 2 lines into 5 in the CHANGELOG note because I am Andriy Kmit. That's why I injected this fix into my previous ivy fixes, not to clutter already super-big changelog.

duianto:
It looked like two different people until your comment 😄 

I would have listed the key bindings on separate lines anyway, because I added the commands that the keys call. It might be helpful for anyone that reads the changelog and wonders what commands the new key bindings call.

madand:
This totally makes sense. I will also document commands behind keybindings in further commits.

PR #13061, opened 2019-12-14
removed unnecessary custom location for counsel-css package

emacs18:
Removed custom package URL, because the custom URL is the same as the default URL in melpa, i.e., 
https://github.com/melpa/melpa/blob/master/recipes/counsel-css

Hence there is no need to specify a custom URL.


smile13241324:

PR #13062, opened 2019-12-14
[CI] add changelog regex

JAremko:
Forgot to add it :grimacing:


JAremko:
Strange. It worked in my fork :thinking:

JAremko:
Sooo it might actually not work for PRs from forks witch makes it kinda useless :thinking:

PR #13059, opened 2019-12-14
fix for #12989 to use cl-lib instead of obsolete cl

emacs18:
Replaced (require 'cl) with (require 'cl-lib) and made other necessary changes to support this change.

This is to fix PR #12989.

Issue #12989, opened 2019-11-25
`cl` Package Deprecated


equwal:
The `cl` package is deprecated (use `cl-lib`); this results in a message warning on startup. Here are all the core files which `(require 'cl)` on the current (2019-11-24) dev branch:
```plain
core/core-configuration-layer.el
core/libs/ht.el
core/libs/mocker.el
core/tools/spacefmt/spacefmt.el
layers/+chat/erc/local/erc-sasl/erc-sasl.el
layers/+chat/erc/local/erc-tex/erc-tex.el
layers/+completion/helm/local/helm-spacemacs-help/helm-spacemacs-help.el
layers/+completion/ivy/local/ivy-spacemacs-help/ivy-spacemacs-help.el
layers/+distributions/spacemacs-base/local/help-fns+/help-fns+.el
layers/+distributions/spacemacs-bootstrap/packages.el
layers/+spacemacs/spacemacs-ui/local/info+/info+.el
layers/+spacemacs/spacemacs-ui-visual/local/zoom-frm/frame-cmds.el
layers/+spacemacs/spacemacs-ui-visual/local/zoom-frm/frame-fns.el
```

duianto:
The first one: `core/core-configuration-layer.el` seems to already be using `cl-lib`
https://github.com/syl20bnr/spacemacs/blob/cab1e69585f7ad47613e45d7123ebfe64bba0a7e/core/core-configuration-layer.el#L13

I'm unable to find an instance of `'cl` in that file.

But here's some more files that also use `'cl`:
```
./layers\+emacs\org\local\org-async-init.el:9:(require 'cl)
+distributions\spacemacs-docker\deps-install\lib\deps-install-helpers.el:12:(require 'cl)
./layers\+misc\ietf\local\irfc\irfc.el:336:(eval-when-compile (require 'cl))
./layers\+lang\python\local\nose\nose.el:56:(require 'cl) ;; for "reduce"
./layers\+lang\python\local\pylookup\pylookup.el:30:  (require 'cl)
```

It seems like functions/macros have been renamed in `cl-lib to use the `cl-` prefix, so maybe we would have to find and rename them.

source: https://emacs.stackexchange.com/questions/48109/require-cl-or-require-cl-lib

equwal:
> It seems like functions/macros have been renamed in `cl-lib to use the `cl-` prefix, so maybe we would have to find and rename them.
> 

Yes, this is the purpose of the deprecation.

[...]

PR #13067, opened 2019-12-14
Do not use the deprecated lexical-let.

equwal:
My autocompletions cannot load, since these two files do not `(require 'cl)` here are some cross-references:
`cl` deprecated issue: https://github.com/syl20bnr/spacemacs/issues/12989
Commit switching to `cl-lib`: https://github.com/syl20bnr/spacemacs/commit/da8063437052943f749bc0f14c5ef5a7536aeaac
That pull request: https://github.com/syl20bnr/spacemacs/pull/13059

These are the only places where `lexical-let` gets used. It looks like `let` works lexically unless the variable is already declared specially somewhere else. [Here are the docs on that](https://www.gnu.org/software/emacs/manual/html_node/elisp/Using-Lexical-Binding.html).

I don't think there are any more spurious uses of `cl` functions in the core.
Here is the magic regex I used:
```bash
rg '\((flet|cl-macro-environment|cl-macroexpand-all|macroexpand-all|cl-not-hash-table|cl-builtin-remhash|cl-builtin-clrhash|cl-builtin-maphash|cl-map-keymap|cl-copy-tree|copy-tree|cl-gethash|cl-puthash|cl-remhash|cl-clrhash|cl-maphash|cl-make-hash-table|cl-hash-table-p|cl-hash-table-count|cl-map-intervals|cl--map-intervals|cl-map-extents|cl--map-overlays|cl-set-getf|cl--set-getf|cl-struct-setf-expander|lexical-let|lexical-let\*|labels|define-setf-expander|defsetf|define-modify-macro)'
```

### DIFF
--- a/layers/+completion/auto-completion/funcs.el
+++ b/layers/+completion/auto-completion/funcs.el
@@ -1,4 +1,4 @@
-;;; funcs.el --- Auto-completion functions File
+;;; funcs.el --- Auto-completion functions File -*- lexical-binding: t; -*-
 ;;
 ;; Copyright (c) 2012-2018 Sylvain Benner & Contributors
 ;;
@@ -295,7 +295,7 @@ MODE parameter must match the :modes values used in the call to
       (define-key map (kbd "C-k") 'company-select-previous)
       (define-key map (kbd "C-l") 'company-complete-selection))
     ;; Fix company-quickhelp Evil C-k
-    (lexical-let ((prev nil))
+    (let ((prev nil))
       (defun spacemacs//set-C-k-company-select-previous (&rest args)
         (setf prev (lookup-key evil-insert-state-map (kbd "C-k")))
         (define-key evil-insert-state-map (kbd "C-k") 'company-select-previous))
--- a/layers/+completion/ivy/funcs.el
+++ b/layers/+completion/ivy/funcs.el
@@ -1,4 +1,4 @@
-;;; funcs.el --- Ivy Layer functions File for Spacemacs
+;;; funcs.el --- Ivy Layer functions File for Spacemacs -*- lexical-binding: t; -*-
 ;;
 ;; Copyright (c) 2012-2019 Sylvain Benner & Contributors
 ;;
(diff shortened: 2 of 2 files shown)


madand:
This issue raises a bigger question: should Spacemacs consistently use lexical binding across its codebase (with `;;; -*- lexical-binding: t -*-`), unless code in a file explicitly relies on non-lexical behavior. I believe templates for new layer generation should also include `lexical-binding: t` at least in `funcs.el`.

Ping @syl20bnr

madand:
@equwal you cannot just replace `lexical-let` with `let`, you also need to enable lexical binding for the whole file with the special comment `-*- lexical-binding: t -*-` on the very first line of a file. For example see: https://github.com/syl20bnr/spacemacs/blob/26b8fe0c317915b622825877eb5e5bdae88fb2b2/core/core-debug.el#L1

equwal:
> @equwal you cannot just replace `lexical-let` with `let`, you also need to enable lexical binding for the whole file with the special comment `-*- lexical-binding: t -*-` on the very first line of a file. For example see:
> 
> https://github.com/syl20bnr/spacemacs/blob/26b8fe0c317915b622825877eb5e5bdae88fb2b2/core/core-debug.el#L1

Good point, I just force pushed that @madand.

layers/+completion/auto-completion/funcs.el
```diff
@@ -295,7 +295,7 @@ MODE parameter must match the :modes values used in the call to
       (define-key map (kbd "C-k") 'company-select-previous)
       (define-key map (kbd "C-l") 'company-complete-selection))
     ;; Fix company-quickhelp Evil C-k
-    (lexical-let ((prev nil))
+    (let ((prev nil))
```
madand: You need to add `-*- lexical-binding: t; -*-` to the first line of this file as well.

equwal: @madand indeed I should have done that, and have done it.

duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/05fb56bfbfd43a38d0b4044796d2cff4cccac971

PR #13068, opened 2019-12-14
[ivy] Fix visual selection expansion across the buffer on `SPC s S` and `SPC s B`

madand:
When one invokes `spacemacs/swiper-region-or-symbol` (`SPC s S`) or
`spacemacs/swiper-all-region-or-symbol` (`SPC s B`) with active visual selection
(transient-mark-mode), the selection expands over the buffer while one jumps
around with swiper. Thus I believe it is appropriate to deactivate visual
selection within the above command.

Since the mentioned commands both use identical code for getting the text to
operate on, I moved that code into a separate function.

On the following screenshot one can see what happened when I visually select symbol `ivy-read`, then `SPC s S` and then , in swiper, just select the next candidate:
spacemacs-swiper-visual-bug


smile13241324:

PR #13069, opened 2019-12-15
[ivy] `SPC h m`: fallback to `woman` on Windows

madand:
`man` command displays output from the external program with the same name,
so it will generally fail to work on Windows.

`woman` is a pure elisp parser for manpages, so it will work on Windows. But
`woman` lacks colorizing of manpages, so we only use it as a fallback.


smile13241324:

PR #13072, opened 2019-12-15
Add LSP backend support for YAML

sei40kr:
Added LSP backend support for YAML.

Code completions for GitHub Workflow:
lsp-yaml

Please also see:
- [yaml-language-server](https://github.com/redhat-developer/yaml-language-server)
- [lsp-yaml.el in lsp-mode](https://github.com/emacs-lsp/lsp-mode/blob/master/lsp-yaml.el)
- [JSON Schema Store](http://schemastore.org/json)


smile13241324:
Works like a charm, can even format my cockroachdb configs, which is a real step forward from the broken emacs vanilla yaml mode. Thanks for contributing this @sei40kr 👍

PR #13075, opened 2019-12-15
Fix CI builds by reverting a breaking change to `mocker.el`

madand:
The change was introduced by #13059. mocker.el is a large piece of legacy code
that depends on deprecated functionality in EIEIO and cannot be trivially
converted to use `cl-lib` by search and replace.


smile13241324:
Merged

PR #13087, opened 2019-12-17
Enhance lsp support for dart layer

itome:
## Description
`dart_language_server` is now obsolete and no longer supported by `lsp-mode`. Instead of that, `dart_analysis_server` which is built in official lsp server, is used by default.
And dart_analysis_server covers all features in [dart-server](https://github.com/bradyt/dart-server).
So I make `dart-server` disabled when `lsp` feature used.

## Changes
- Refactor code.
- Update README.
- add `dart-backend` configuration (`lsp` or `analyzer`)

### DIFF
--- a/layers/+lang/dart/README.org
+++ b/layers/+lang/dart/README.org
@@ -8,12 +8,12 @@
 - [[#description][Description]]
   - [[#features][Features:]]
 - [[#install][Install]]
-  - [[#layer][Layer]]
-  - [[#dart-sdk-location][Dart SDK location]]
+  - [[#choosing-a-backend][Choosing a backend]]
+- [[#backends][Backends]]
+  - [[#dart-analyzer][Dart analyzer]]
     - [[#flutter-integration][Flutter integration]]
+    - [[#format-on-save][Format on save]]
   - [[#lsp][LSP]]
-- [[#configuration][Configuration]]
-  - [[#format-on-save][Format on save]]
 - [[#key-bindings][Key bindings]]
   - [[#normal-mode][Normal mode]]
   - [[#insert-mode][Insert mode]]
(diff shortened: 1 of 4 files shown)


smile13241324:
@itome thank you for contributing to spacemacs, I have merged your PR and revised the documentation and code. However I am not sure that I have got everything right, can you double check the version on develop?

My main questions are about which features of the layer are "analyzer" specific, from my current understanding it should be all features except of what lsp layer provides. This includes flutter support as well as the "formatting on save" feature or I am mistaking there.

Also when I setup everything correctly for lsp I am getting always "invalid file path" directly from LSP though `lsp-dart-sdk-dir` is correctly set to `~/Downloads/dart-sdk/`. Can you double check?

itome:
@smile13241324 Thank you for approving. I'll check it soon.

itome:
@smile13241324 
## TL;DR
`lsp` and `analyzer` now works separately.

The history of dart backend is a little bit complicated. 
This layer originally depends on [dart-server](https://github.com/bradyt/dart-server) and lsp-mode with [dart_language_server](https://github.com/natebosch/dart_language_server). They both use `dart_analyzer` which is built into dart sdk. Because `dart_language_server`(third party `lsp` wrapper 
 for `dart_analyzer`) is slow, original implementation uses `lsp` for only completion and cover other features with `dart-server`.

After this layer created. Dart sdk officially support `lsp` with `dart_analyzer --lsp` command, and `dart_language_server` stopped maintenance. `lsp-mode` follow that change and use `dart_analyzer --lsp` as dart lsp backend. `dart_analyzer` lsp support have full features and sufficient performance.

For the above reasons, `dart-server` and `lsp-mode` is almost same features. The only difference between them is that `dart-server` doesn't support completion.

In current develop. The two backends (`lsp-mode` and `dart-server`) doesn't work at the same time. If you set `dart-backend 'lsp` `dart-server` is disabled and if you set `dart-backend 'analyzer` `lsp-mode` is disabled.

"formatting on save" feature in README is for `analyzer`, if you want to use same feature with `lsp` you can set `lsp-enable-on-type-formatting t`. And both backends works well with flutter (I tested with flutter project).

itome:
@smile13241324 I checked lsp on recent develop branch and it works well. Would you confirm `~/Downloads/dart-sdk/` exists?

smile13241324:
@itome thanks for the long answer :), yep the folders are existing on my side however I have a lot of layers in my dotfile so maybe its a side effect of my config, thanks for double checking.

I will add a note for the `lsp-enable-on-type-formatting` variable in the README.

PR #13088, opened 2019-12-17
Un evilify org present mode

duianto:
Un-evilified org-present-mode

Because [escape] was bound to evil-evilified-state
after exiting org-present-mode.

The arrow keys switched to the next/prev page before this PR,
but this solution seems to require them to be defined explicitly.

Resolves: #7766

---

### Reproduction steps
- `example.org`:
```org
* one
a b c d e
* two
1 2 3 4 5
* three
I II III III IV
```
- `M-x org-present`
- `q` exits `org-present-mode`
- `ESC`

#### Before
`ESC` has been rebound to `evil-evilified-state`.
It changes `evil-state` to `evilified` where for example:
`w`, `b` and `u` insert the characters.

#### After
`ESC` didn't change, it still calls: `evil-force-normal-state`,
and the evil bindings `w`, `b`, `u`, etc. work as expected.


smile13241324:
Merged

PR #13092, opened 2019-12-18
Convert `case' to `cl-case' to fix cl-lib related errors

nikital:
Broken since PR #13059

I tested the modified functions with Emacs 26.3 on Linux (so the macOS change is untested but it should work...).

Thanks :)

### DIFF
--- a/layers/+completion/auto-completion/packages.el
+++ b/layers/+completion/auto-completion/packages.el
@@ -212,7 +212,7 @@
             company-box-doc-enable nil)
       (add-hook 'company-box-selection-hook
                 (lambda (selection frame) (company-box-doc--hide frame)))
-      (case auto-completion-enable-help-tooltip
+      (cl-case auto-completion-enable-help-tooltip
         ('manual (define-key company-active-map
                    (kbd "M-h") #'company-box-doc-manually))
         ('t (setq company-box-doc-enable t))))))
--- a/layers/+completion/ivy/funcs.el
+++ b/layers/+completion/ivy/funcs.el
@@ -337,7 +337,7 @@ To prevent this error we just wrap `describe-mode' to defeat the
           (user-error "C-c C-c can do nothing useful at this location"))
     (let* ((context (org-element-context))
            (type (org-element-type context)))
-      (case type
+      (cl-case type
         ;; When at a link, act according to the parent instead.
         (link (setq context (org-element-property :parent context))
               (setq type (org-element-type context)))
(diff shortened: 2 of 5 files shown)


duianto:
There's also a usage of `case` in the Spacemacs local version of: `zoom-frm.el`
https://github.com/syl20bnr/spacemacs/blob/6f029cc0573b8f42a05dea26be92617d5d5282cb/layers/%2Bspacemacs/spacemacs-visual/local/zoom-frm/zoom-frm.el#L376

Should it be changed upstream?
The package still supports versions down to Emacs 20.x.
https://github.com/syl20bnr/spacemacs/blob/6f029cc0573b8f42a05dea26be92617d5d5282cb/layers/%2Bspacemacs/spacemacs-visual/local/zoom-frm/zoom-frm.el#L17

The upstream version https://www.emacswiki.org/emacs/zoom-frm.el
- Mentions `cl-lib`:
```
;; Features that might be required by this library:
;;
;;   `avoid', `backquote', `bytecomp', `cconv', `cl-lib',
;;   `frame-fns', `macroexp', `misc-fns', `strings', `thingatpt',
;;   `thingatpt+'.
```
- And it has this command:
```
(eval-when-compile (require 'cl)) ;; case
```

The local Spacemacs version does not have that line, because it's been added in a more recent version.
The local Spacemacs version is:
```
;; Last-Updated: Thu Jan  1 11:24:13 2015 (-0800)
;;           By: dradams
;;     Update #: 322
```
Upstream: https://www.emacswiki.org/emacs/zoom-frm.el
```
;; Last-Updated: Tue Nov 19 13:15:54 2019 (-0800)
;;           By: dradams
;;     Update #: 359
```

This commit:
converted cl to cl-lib https://github.com/syl20bnr/spacemacs/commit/da8063437052943f749bc0f14c5ef5a7536aeaac

has already changed the local Spacemacs version of `frame-cmds.el`.
Which isn't up to date either.
```
;; Last-Updated: Thu Jan  1 10:44:52 2015 (-0800)
;;           By: dradams
;;     Update #: 3036
```
Upstream: https://www.emacswiki.org/emacs/frame-cmds.el
```
;; Last-Updated: Fri Dec 13 17:41:12 2019 (-0800)
;;           By: dradams
;;     Update #: 3181
```

One issue with changing the local versions is that when they are updated, then there's a risk that these deprecation fixes will be lost.

But it's better to fix the local version now.

nikital:
I didn't modify zoom-frm.el is because I noticed that `zoom-in/out`, which is the function which has the `case`, doesn't even seem to be evaluated in Spacemacs.
I started with a empty spacemacs-base config on Emacs 26.3, added spacemacs-visual layer and confirmed that `zoom-in/out` isn't even evaluated.
Therefore I didn't want to touch the local version since there seemed to be no reason to diverge more from upstream.

duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/aaa11a6823e0980177f8df1769f085faa5d82f31

PR #13093, opened 2019-12-18
shell: Wrap terminal-here in use-package

nikital:
Introduced in PR #11949, terminal-here wasn't wrapped in `use-package`. This made it impossible to use `spacemacs|use-package-add-hook` to customize it.

Thanks!


smile13241324:

Issue #13083, opened 2019-12-16
Keys not remapped by keyboard-layout after da80634


PiOverFour:
#### Description :octocat:
Using the keyboard-layout layer, the following error message appears on startup for 14 packages (evil, helm, magit, etc.):
`(Spacemacs) Error: 
An error occurred while pre-configuring <package> in layer keyboard-layout (error: (void-function typep))`

The keys are thus not remapped.

The bug was introduced by commit da8063437052943f749bc0f14c5ef5a7536aeaac, which renamed some functions. Proposed fix:

``` diff
diff --git a/layers/+intl/keyboard-layout/funcs.el b/layers/+intl/keyboard-layout/funcs.el
index 5be0cc76e..1de781f24 100644
--- a/layers/+intl/keyboard-layout/funcs.el
+++ b/layers/+intl/keyboard-layout/funcs.el
@@ -57,7 +57,7 @@ then `c' will be defined to the old `a' function, not to `b'."
   "Replace recursively all occurrences of `elem' by `repl' in the
 list `lst'."
   (declare (indent 0))
-  (if (typep lst 'list)
+  (if (cl-typep lst 'list)
       (let* ((body-position (cl-position elem lst)))
         (if body-position
             ;; The element is in the list, replace it
@@ -134,7 +134,7 @@ evil states, except insert."
 (defun kl/correct-keys (map &rest keys)
   (declare (indent 1))
   (let ((bindings (mapcan #'kl//guess-rebindings keys)))
-    (kl//remap-key-as map (remove-if #'null bindings))))
+    (kl//remap-key-as map (cl-remove-if #'null bindings))))
 
 (defun kl/evil-correct-keys (state map &rest keys)
   (declare (indent 2))

```

#### Reproduction guide :beetle:
- Enable keyboard-layout layer in dotspacemacs-configuration-layers
- Start Emacs
- Error occurs

*Observed behaviour:* :eyes: :broken_heart:
The keys are not remapped for these packages:
evil, evil-evilified-state, ace-window, avy, company, evil-escape, evil-magit, evil-surround, eyebrowse, helm, magit, org, org-agenda, undo-tree

*Expected behaviour:* :heart: :smile:
They should.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 26.3
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. c6d764e04)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
[...]

PR #13095, opened 2019-12-18
Fix keys not remapped by keyboard-layout

PiOverFour:
This fixes the issue #13083.
keyboard-layout was broken by commit da80634, which replaced cl by cl-lib. Some function calls weren’t replaced. There may be old calls remaining in other layers.


duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/473d019b47d75ee7cba25528bbf14fb9acbf0aa4

Issue #9631, opened 2017-09-21
cannot choose from helm find-file actions just after emacs is started.


vibrys:
#### Description :octocat:
cannot choose from helm find-file actions just after emacs is started.
I'm running fresh spacemacs (not spacemacs-base).

#### Reproduction guide :beetle:
- Start Emacs
- SPC f f  (helm-find-files)
- display actions for current file (does not matter which file is currently chosen). You'll see all the actions list in separate emacs buffer
- press 's' key

*Observed behaviour:* :eyes: :broken_heart:
following error message is displayed:
Helm issued errors: helm-match-from-candidates in source `Actions': wrong-type-argument (stringp nil)

*Expected behaviour:* :heart: :smile:
actions' list should be narrowed down to those which have 's' in their name.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 25.3.1
- Spacemacs: 0.200.9
- Spacemacs branch: nil (rev. ca28bde)
- Graphic display: t
- Distribution: spacemacs
- Editing style: emacs
- Completion: helm
- Layers:
```elisp
(helm emacs-lisp)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG SOUND GPM DBUS GCONF GSETTINGS NOTIFY GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK2 X11

#### Backtrace :paw_prints:
I'm able to display the backtrace, but unfortunatelly cannot copy it, because it dissapears when I'm continuing to press any keys (helm is still in action).

vibrys:
for the sake of completeness, this problem DOES NOT persist with vanilla: emacs + helm.

regards,
Mat

bmag:
Can't repro with latest Spacemacs and Helm. "display actions for current file" means pressing `C-z`?

duianto:
Confirmed on both the latest master and develop branches with all packages updated.

This seems to be a duplicate of this issue:
spacemacs/helm-find-files (SPC f f) not correctly autoloading #7325

The next to last comment on that issue might be describing the source or a clue to what's going on.

That issue explains the behaviour I'm seeing. If the helm find files, helm action buffer, is opened as the first helm buffer, then the issue occurs. But as soon as another helm buffer has been opened. For example helm mini `SPC b b`, then the helm find files, helm action buffer, is able to narrow down the results based on the typed pattern.

[...]

PR #13107, opened 2019-12-20
Fix searching in the helm-find-file actions

duianto:
Fixes: #9631

---

If it's overkill to mention the error message it fixes and where the function is used in it's docstring, then I can remove it.


smile13241324:
Merged

PR #13113, opened 2019-12-22
Frame title setting doesn't need to depend on environment (gui/cli)

mphojele:
remove `display-graphic-p` from conditional that sets `frame-title-format`. Take this example:
- you run `emacs -daemon`, it loads everything and `display-graphic-p` at the time returns `nil`
- you run `emacslient -c ` which launches a gui emacs frame but now your `frame-title` is not set
  because it was dependent on `display-graphic-p` returning a true value.
We certainly do not want this. Moreover, the are no dire effects to setting the `frame-title` in a `CLI`

### DIFF
--- a/core/core-spacemacs.el
+++ b/core/core-spacemacs.el
@@ -103,7 +103,7 @@ the final step of executing code in `emacs-startup-hook'.")
                                     dotspacemacs-editing-style))
   (configuration-layer/initialize)
   ;; frame title init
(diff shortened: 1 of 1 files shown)


duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/08d19389cce3d906ef50e9c9d0e6155194d47dd2

Issue #13111, opened 2019-12-21
Rust quick run broken on develop


kidpollo:
#### Description :octocat:
Rust quick run does not produce output

#### Reproduction guide :beetle:
- Start Emacs
- create hello.rs as https://doc.rust-lang.org/rust-by-example/hello.html
- [SPC q]

*Observed behaviour:* :eyes: :broken_heart:
Compilation works but program does not get run

*Expected behaviour:* :heart: :smile:
"Hello World" output

#### System Info :computer:
- OS: gnu/linux
- Emacs: 26.3
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. ffccb1742)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(csv graphviz rust spell-checking gtags helm treemacs html github
     (ruby :variables ruby-version-manager 'rvm)
     yaml javascript better-defaults osx colors org syntax-checking emacs-lisp
     (auto-completion :variables auto-completion-enable-help-tooltip t auto-completion-enable-sort-by-usage t)
     (shell :variables shell-default-shell 'multi-term shell-default-term-shell "zsh")
     (version-control :variables version-control-global-margin t version-control-diff-tool 'git-gutter+ version-control-diff-side 'left)
     (git :variables git-gutter-use-fringe t)
     (clojure :variables clojure-enable-fancify-symbols t clojure-enable-linters 'clj-kondo)
     markdown
     (spell-checking :variables spell-checking-enable-by-default nil)
     sql)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GSETTINGS GLIB NOTIFY ACL GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XDBE XIM MODULES THREADS LIBSYSTEMD LCMS2

#### Backtrace :paw_prints:
```
<<BACKTRACE IF RELEVANT>>
```

gcentauri:
@kidpollo - if you pull recent changes, i think this should work for you

duianto:
In the reproduction steps, the third step says: `[SPC q]`

Have you rebound it to a custom command, or was it a typo?

Because by default `SPC q` shows the Spacemacs quit key bindings for the commands:
[...]

PR #13116, opened 2019-12-23
[rust] Fix: broken rust-quick-run command

gcentauri:
removing the `lexical-let` caused the intended closure to be invalid. the
filename being compiled was nil when the compilation-finish-function was
run.

this adds a special var to hold the temp file being compiled instead, and
tries to narrow down the compilation-finish-function to only run when the
compilation buffer includes the "rustc -o /tmp" regexp

attempts to address #13111 which seems to have been caused by beff6ec9149cc4954a41bcb5cbf4d02c8af8103f

### DIFF
--- a/layers/+lang/rust/funcs.el
+++ b/layers/+lang/rust/funcs.el
@@ -82,6 +82,23 @@ If `help-window-select' is non-nil, also select the help window."
 
 ;; Misc
 
+(defvar spacemacs/rust-quick-run-tmp-file nil
+  "Stores filename for the rust-quick-run function")
+
+(defun spacemacs//rust-quick-run-generate-tmp-file-name (input-file-name)
+  (concat temporary-file-directory
+          (file-name-nondirectory (buffer-file-name))
+          "-"
+          (md5 (buffer-file-name))))
+
+(defun spacemacs//rust-quick-run-compilation-finish-function (buffer status)
+  (if (and (string-match "finished" status)
+           (with-current-buffer buffer
+             (string-match (concat "rustc -o " temporary-file-directory) (buffer-string))))
+      (progn
+        (newline)
+        (shell-command (shell-quote-argument spacemacs/rust-quick-run-tmp-file) t))))
+
 (defun spacemacs/rust-quick-run ()
   "Quickly run a Rust file using rustc.
 Meant for a quick-prototype flow only - use `spacemacs/open-junk-file' to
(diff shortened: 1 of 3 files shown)


gcentauri:
I'm not sure if its better or worse to introduce a new special variable to store the temp file name used for this, but i've done a bit of manual testing and this seems to work fine, as well as not interfere with some other compilation modes.  since this feature adds a function to run when *any* compilation command finishes, it had been causing strange behavior like in `M-x rgrep` you'd see a message like "Compilation failed with "finished""

layers/+lang/rust/funcs.el
```diff
@@ -82,24 +82,33 @@ If `help-window-select' is non-nil, also select the help window."
...
+  "Stores filename for the rust-quick-run function")
+
+(defun spacemacs//rust-quick-run-generate-tmp-file-name (input-file-name)
+  (concat temporary-file-directory
+          (file-name-nondirectory (buffer-file-name))
+          "-"
+          (md5 (buffer-file-name))))
+
+(defun spacemacs//rust-quick-run-compilation-finish-function (buffer status)
+  (if (and (with-current-buffer buffer (string-match "rustc -o /tmp" (buffer-string)))
```
gcentauri: i guess the `"/tmp"` part isn't going to work, since it is being populated by `temporary-file-directory` :(

smile13241324:

Btw, I agree with you having a global temporary variable is not idiomatic as far as I see it the proper way would be to finally set lexical binding for the layer files. However this is no easy change and must be separately tested before.

I have also tried an alternative approach with buffer local variables to avoid setting the compilation finish function globally however as `compile` switches to a global `*compilation*` buffer this was also not working. For now I have added some code to try to remove the compilation function when it did run.

However there must be a more idomatic approach for this. @duianto do you have encountered a similiar problem yet? The current code removes the symptom but it is definitely not clean. Do you have a seen a better example for an issue like that?

gcentauri:
@smile13241324 - i just had a thought that maybe instead of adding a function to the global `compilation-finish-functions`, the contents of the compilation buffer could be copied into another buffer and combined with the output of running the compiled code?  I had an issue trying to write results directly into the compilation buffer (it was read-only).  

I guess I'm not exactly sure the desired behavior of this `rust-quick-run` function, but without using the `&&` operator (which was removed for being shell specific) I'm not sure how to include the running of the compiled code in the compilation buffer.

smile13241324:
@gcentauri well adding output to the compilation buffer should already been working currently the `compilation-finish-function` just runs the binary output and adds the result to the buffer. I am not sure that copying everything into another buffer would be a proper solution the compilation buffer normally is pretty good in formatting highlighting the compilation output. But there are other layers i.e. the go layer which must have the same issue. As it looks like nobody else is using these `compilation-finish-functions` so I have the feeling we doing something wrong or at least overly complicated here.

duianto:
@smile13241324 I don't know much about the dynamic/lexical stuff.

My comment https://github.com/syl20bnr/spacemacs/pull/9535#issuecomment-566969900 about there being a possible issue with using `lexical-let` in the `rust` layer, was just based on the fact that `lexical-let` was removed in the `auto-completion` and `ivy` layers.
Do not use the deprecated lexical-let. #13067

PR #13121, opened 2019-12-24
[go] Update build instructions for golangci-lint

panchoh:
See #10825.

Try to match more closely the build process followed upstream:

- Force module build mode, which builds the latest stable release (not master),
and takes care of module replacements.

- Disable cgo.

- The -trimpath argument to go get requires go >= 1.13, which is also needed
because go <= 1.12 has a bug that causes problems with golangci-lint², as
explained in the golangci-lint installation instructions².

Note that the only thing that is missing to exactly match the upstream build
process is the addition of some -X variables, which only affect the data
reported by golangci-lint version.

¹: https://github.com/golang/go/issues/29612
²: https://github.com/golangci/golangci-lint#install

Thanks to @dbriemann and @seriousben for reporting and clarifying this issue!

### DIFF
--- a/layers/+lang/go/README.org
+++ b/layers/+lang/go/README.org
@@ -76,6 +76,7 @@ to get all the goodies of this layer:
 
 #+BEGIN_SRC sh
   GO111MODULE=on go get -v golang.org/x/tools/gopls@latest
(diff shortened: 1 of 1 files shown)


duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/1fa28b345c70743c6ebfd26c9de57f0af15a4988

PR #13122, opened 2019-12-25
[core] Toggleable hint for the Scrolling Transient State (~SPC N~)

madand:
By default the full hint is shown, which is consistent with the behavior before
this commit.


duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/37fdd1ef87f739f27c548e19a78c3a5b0d51a0ef

#### Changes:
It's not a big deal but it might be helpful to know in the future:

Moved the changelog entry from:
*** Breaking Changes
**** Layers
***** Spacemacs distribution layers
- Key bindings:

To:
*** Distribution changes
- Improvements:

since the changed files are in the `layers/+distributions` section, and this isn't a breaking change, (hopefully 😄)

madand:
Note taken. Thanks!

PR #13124, opened 2019-12-26
[spacemacs-layouts] add details of restricted functions

jr0cket:
Detail how to use the layer variables to make use of restricted functions and
limit their scope to within a layout.

Documentation copied from the following commit, which did not get included along with the code
https://github.com/syl20bnr/spacemacs/tree/0e9866efcee9915f3418cec11371d02ee619ced2/layers/%2Bspacemacs/spacemacs-layouts

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1090,6 +1090,7 @@ Other:
                spacemacs/window-split-triple-columns
                spacemacs/window-split-grid)
     (thanks to Thanh Vuong)
+  - Added documentation to use a =spacemacs-layouts-restricted-functions= variables (John Stevenson)
   - Added a link to =spacemacs|create-align-repeat-x= in the docstring for
      =align-repeat-= commands (thanks to Troy Pracy)
   - Improved TS formatting: workspaces & layouts (thanks to duianto)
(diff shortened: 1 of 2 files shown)


jr0cket:
I initially forgot to add an entry in the CHANGELOG,develop.  This has now been squashed and pushed.

duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/b0d15851d9b99a9aff05c2944f6b576b1a1fde7e

duianto:
I forgot that magit strips brackets from the commit name by default, when applying a patch.
I've saved the setting to keep it, but I must have reinstalled since then. My bad.

jr0cket:
Looking at the Spacemacs commit history, it seems most of the commits do not have the layer name in Square brackets.  I didn't realise it maybe because of magit.  Will have to look up that feature.  

Thanks for your help.

duianto:
They aren't required, but it's probably helpful when there are leading brackets in the commit subject, with the layer or section (ex: `[docs]`). Then it's easy to see which sections the commit targets.

You did everything right, It was my mistake for not remembering to turn off the bracket removal after reinstalling Spacemacs. I wrote it down and hopefully I'll remember to check the notes during the next reinstall.

Thanks again for improving the documentation.

PR #13131, opened 2019-12-27
[rust] remove compilation buffer string match on rust-quick-run

gcentauri:
this removes the (now unnecessary) check for the string `(concat "rustc -o " temporary-file-directory)` in the compilation finish function.  it should make this work on Windows now @duianto

### DIFF
--- a/layers/+lang/rust/funcs.el
+++ b/layers/+lang/rust/funcs.el
@@ -93,9 +93,7 @@ If `help-window-select' is non-nil, also select the help window."
 
 (defun spacemacs//rust-quick-run-compilation-finish-function (buffer status)
   (setq compilation-finish-functions (delete 'spacemacs//rust-quick-run-compilation-finish-function compilation-finish-functions))
-  (if (and (string-match "finished" status)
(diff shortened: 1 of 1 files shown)


duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/2e30b9bb09850646635dec437fe81ab9073aff57

Confirmed, now the result is shown in the compilation buffer when calling `spacemacs/rust-quick-run` (`SPC m q`) from the `hello.rs` file from the issue:
Rust quick run broken on develop #13111 
```
-*- mode: compilation; default-directory: "~/Temp/rust/" -*-
Compilation started at Sat Dec 28 15:29:19

rustc -o "c:/Users/username/AppData/Local/Temp/hello.rs-ef295451ba920ffe13695ec92417bbd5" "c:/Users/username/Temp/rust/hello.rs"

Compilation finished at Sat Dec 28 15:29:20

Hello World!
```

#### Change(s)
The `if` was changed to a `when`, since there isn't an else statement.
That also eliminated the need for the `progn`.

PR #13134, opened 2019-12-29
[java][lsp] Minor cleanup in lsp-java

yyoncho:
- Fixed setting dap-java shortcuts - ATM the shortcuts were not enabled due to
the fact that dap-mode configuration was overriding them since it was running
after this piece of code
- removed configuration code about lsp-java since it was either about
missing properties or about setting defaults already set by lsp-mode/lsp-java
(e. g. lsp-eldoc-render-all is now nil by default)

### DIFF
--- a/layers/+lang/java/funcs.el
+++ b/layers/+lang/java/funcs.el
@@ -133,7 +133,15 @@
 
 (defun spacemacs//java-setup-lsp-dap ()
   "Setup DAP integration."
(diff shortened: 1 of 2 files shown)


duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/3771759ddb59ba426ed90c6e9622fba14a22b00f

Issue #13123, opened 2019-12-25
let*: Symbol’s function definition is void: helm-org-get-candidates


duianto:
#### Description :octocat:
The `helm-org` package has removed the function: `helm-org-get-candidates`

In this PR:
Use dynamic completion emacs-helm/helm-org#10

Spacemacs still uses it here:
https://github.com/syl20bnr/spacemacs/blob/37fdd1ef87f739f27c548e19a78c3a5b0d51a0ef/layers/%2Bcompletion/helm/local/helm-spacemacs-help/helm-spacemacs-faq.el#L64

This issue was reported by `mphojele` in the gitter chat:
https://gitter.im/syl20bnr/spacemacs?at=5e03b519d2dadb3893793ff3

#### Reproduction guide :beetle:

- `SPC h f`

*Observed behaviour:* :eyes: :broken_heart:
`let*: Symbol’s function definition is void: helm-org-get-candidates`

*Expected behaviour:* :heart: :smile:
The headings in FAQ.org should be shown in a helm buffer:

![emacs_2019-12-25_22-21-36](https://user-images.githubusercontent.com/13420573/71450754-4707d880-2768-11ea-9c3d-2c5c75adaa57.png)

#### Notes :notebook:
The screenshot was taken with a previous version of helm-org: `helm-org-20190818.521`
from before the PR emacs-helm/helm-org#10 was merged.

#### System Info :computer:
- OS: windows-nt
- Emacs: 26.3
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 37fdd1ef8)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(autohotkey auto-completion command-log emacs-lisp git helm imenu-list markdown multiple-cursors
            (org :variables org-agenda-files
                 '("c:/Users/username/org/notes.org"))
            (shell :variables shell-default-shell 'eshell)
            spell-checking syntax-checking treemacs version-control
            (xclipboard :variables xclipboard-enable-cliphist t))
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG SOUND NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS THREADS LCMS2

duianto:
There's a pending draft PR:
Fix helm-spacemacs-help-faq, SPC h f #13136

smile13241324:
[...]

PR #13136, opened 2019-12-30
Fix helm-spacemacs-help-faq, SPC h f

duianto:
The helm-org-get-candidates function was removed upstream.

Reversed the helm headings list, to match the order of the
questions in FAQ.org, so that the Common questions are
listed first.

Removed the : (colon) after the heading: How do I
in FAQ.org, to match the other parent headings.

Resolves: #13123

---

### Question
This PR proposes two solutions:
- `Version 1` uses the private function: `helm-org--get-candidates-in-file`
- `Version 2` uses `helm-org-build-sources`
This version uses a somewhat convoluted way to drill down to the FAQ.org headings:
`     (dolist (c (aref (aref (cdadar cands) 2) 0))`
Maybe there's a simpler method for accessing the headings.

Is one version better than the other?
Both versions seem to work the same otherwise.

### Issue
With either version, when a helm candidate is chosen, then the cursor always goes to the end of the parent heading, ex: `Common`, `How do I` or `Windows`
instead of to the expected target question.

I don't know how to get it to jump to the marker for the question heading instead of the parent heading.

As noted in the comments before version 1.
While testing, if the version 2 function is evaluated, then the version 1 function won't work until Spacemacs has been restarted, because version 1 adds the prefix `FAQ.org:` before each heading line (internally not in the helm buffer).
After the version 2 function has been evaluated, then the filename prefix won't appear, therefore the string-match regexp won't find the `:` (colon) before the headings.

There probably is a regexp that handles either case, or the regexp could probably be improved, any suggestions for improvements are always welcome.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1258,6 +1258,7 @@ Other:
   - Fixed searching in the =helm-find-file= actions (thanks to duianto)
   - Fixed =evilified-state=, mapped ~C-w~ to =evil-window-map=
     (thanks to Muneeb Shaikh)
+  - Fixed ~SPC h f~ =helm-spacemacs-help-faq= (thanks to duianto)
 *** Layer changes and fixes
 **** Agda
 - Fixes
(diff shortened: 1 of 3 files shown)


syl20bnr:
I would go with version 2 as `helm-org-build-sources` is the new public method to build candidates.
`(aref (aref (cdadar cands) 2) 0)` can be extracted in its own function to better document what it does.

After looking at the changes they are also caching implementation added around the same time to I bet caching is messing with the results and it might explain you observe such a  weird behavior.

syl20bnr:
💯 
Feel free to merge when rebased.

Issue #12593, opened 2019-08-05
[Clojure] Add Cider connect and Session management keybindings to layer


jr0cket:
We now have more cider-connect-* commands and recent session management commands via sesman that do not have keybindings in the Clojure layer.

Please suggest keybindings you would like to see, with respect to the mnemonic approach used by Spacemacs.

From [docs.cider.mx](https://docs.cider.mx/cider/usage/managing_connections.html) we have the following connection commands available

`C-c C-x c j` to call `cider-connect-clj`
`C-c C-x c s` to call `cider-connect-cljs`
`C-c C-x c m` to call `cider-connect-clj&cljs`

Add new REPLs to the current session with

`C-c C-x s j cider-connect-sibling-clj`
`C-c C-x s s cider-connect-sibling-cljs`

Session life-cycle management commands live on the Sesman map (C-c C-s)

`C-c C-s s sesman-start`
`C-c C-s r sesman-restart`
`C-c C-s q sesman-quit`

Discussion forked from the abandoned PR #12061

yuhan0:
Here are the sesman related bindings from my local config, never quite got around to making a PR:
```diff
+            "sI" 'sesman-info
+            "slb" 'sesman-link-with-buffer
+            "sld" 'sesman-link-with-directory
+            "slp" 'sesman-link-with-project
+            "slu" 'sesman-unlink
+            "sm" 'sesman-browser
+            "sQ" 'sesman-quit
+            "sS" 'sesman-start
-            "sX" 'cider-restart
+            "sX" 'sesman-restart
```

Rationale: `sesman-start`, `-quit` and `-restart` commands are like more general variants of the `cider-*` ones, hence the uppercase.
 E.g. for jacking-in / connecting / siblings I usually go for `sesman-start` and interactively choose from the dropdown, rather than memorize 8 different bindings. 
`sesman-link-*` commands are also very useful for evaluating forms in external libraries.

On a related note, why was the "f" in "sjf" chosen for `cider-jack-in-clj&cljs`? Cider defaults use "m" which seems equally arbitrary.

jr0cket:
@yuhan0 thanks, this keybindings are very useful for our discussions.  I will add some and take them for a test drive myself this week

`s j f` was chosen as the mnemonic (s)cider (j)jack-in (f)fullstack - as fullstack is a fairly common term for combining front end (ClojureScript) and backend (Clojure).  Although I appreciate that you can also do ClojureScript on the backend with node.js.

[...]

PR #13140, opened 2019-12-31
[clojure] Add sesman keybindings

jr0cket:
Add Spacemacs keybindings to the Clojure layer for all the sesman functions.
https://docs.cider.mx/cider/0.23/usage/managing_connections.html

Sesman has been a part of CIDER for many versions now and provides session
management for nREPL connections (in fact any connections).

Several keybindings relating to the use of a specific repl been migrated the
`mR` sub-menu, providing a cleaner separation between repl connection and
session management and interacting with a specific repl.

**Please read issue #12593 which discusses alternatives to this PR**

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1453,6 +1453,54 @@ Other:
   - Removed ~SPC m h g~ for =grimoire= as it is deprecated (thanks to Sam Hedin)
   - ~C-return~ to =cider-repl-newline-and-indent= in REPL evil insert state
     (thanks to Gia Thanh Vuong)
+  - Moved evaluation sent to repl keybindings to ~SPC m e s~ menu
+    ~esb~ 'spacemacs/cider-send-buffer-in-repl-and-focus
+    ~esc~ 'cider-repl-clear-buffer
+    ~esC~ 'cider-find-and-clear-repl-output
+    ~ese~ 'spacemacs/cider-send-last-sexp-to-repl
+    ~esE~ 'spacemacs/cider-send-last-sexp-to-repl-focus
+    ~esf~ 'spacemacs/cider-send-function-to-repl
+    ~esF~ 'spacemacs/cider-send-function-to-repl-focus
+    ~esn~ 'spacemacs/cider-send-ns-form-to-repl
+    ~esN~ 'spacemacs/cider-send-ns-form-to-repl-focus
+    ~eso~ 'cider-repl-switch-to-other
+    ~esr~ 'spacemacs/cider-send-region-to-repl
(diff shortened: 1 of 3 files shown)


jr0cket:
I have updated the keybinding changes based on feedback in the Clojurians slack channel.  The change should have very minimal impact to esablished workflows, whist adding more features of CIDER.

I still need to update the documentation and changelog.

jr0cket:
Positive feedback for these revised changes have been received from the Clojure community, so have updated the documentation and changelog entries

duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/e47494229b965014e7ed8ecb308fa6e710508a12

syl20bnr:
@jr0cket This is a great PR but I need to revert it as it breaks the conventions: https://github.com/syl20bnr/spacemacs/blob/develop/doc/CONVENTIONS.org#repls

I suggest reopening a PR and managing sessions under `SPC m S` or another available prefix that does not break convention.

jr0cket:
@syl20bnr ah, finally I understand what the `s` means.  Sorry, although I knew there were conventions I never realized there was a document that explained them, its very helpful.  Thanks

Although I am now confused about what the keybindings should be now, as it raises more questions than it answers. 

The `s` convention does seems to be a little overloaded in the Clojure layer, as there have long been keybindings to start/restart/clear/quit the REPL.  These don't seem to fit into a menu where the focus is sending code to the REPL.  There are a lot of commands used manage the REPL connection in Clojure it seems, and lots more to send code to the REPL buffer, and quite a few to evaluate (some of which have been asked about but no one added yet).

In Clojure its far less common to have the REPL buffer on display and the buffer is not displayed by default.  Instead evaluation is done in the source code buffer, using the keybindings under `SPC m e`.  So it did seem to make sense to put all the keybindings that evaluate code in one place.   But I appreciate this breaks the convention.

If it fits convention, then I can keep `SPC m e` for source code buffer evaluation, `SPC m s` for sending code to the REPL buffer.

I am confused where start/stop/restart/clear/quit the REPL make sense.  I checked the conversions doc for suggestions, but didnt see anything about sessions or starting / stopping external repl processes.  

Leaving the `SPC m s` menu as it is after the roll-back seems to break convention too, because it has REPL connection management in there too.  Or am I missing something?

What about moving the start/stop/restart/clear/quit REPL keybindings to `SPC m c` (for connections) and keeping all the send-code-to REPL commands in `SPC m s`?  Would this be okay?

I'll make a new PR so people can try the new set of keybindings and see what feedback I get from the Clojure community as well.

Thank you.

PR #13146, opened 2020-01-02
[helm] Fix rg interaction with its config file

### LINKED ISSUE #10953: ripgrep / rg stops working with Helm / helm-ag
#### Description :octocat:
`ripgrep` stops working with `helm`

screen shot 2018-06-25 at 4 16 17 pm

**Putting `--` in the pattern removes the error**

screen shot 2018-06-25 at 4 13 14 pm

`helm-ag` works if I uninstall `ripgrep` but it is using `ag`.

- Install `ripgrep`
- Start Emacs
- Open files in directory (projectile)
- Do `[SPC /]` to invoke `Helm Ag: Using rg`
- type some text/pattern

```
error: The following required arguments were not provided:
    <PATTERN>
```

searching using `rg`

#### System Info :computer:
- OS: darwin
- Emacs: 26.1
- Spacemacs: 0.300.0
- Spacemacs branch: dev (rev. 81c2eb8e4) *this is based on develop*
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(ansible better-defaults django docker emacs-lisp git github html javascript markdown nginx
         (org :variables org-confirm-babel-evaluate nil)
         osx search-engine slack spotify sql theming vimscript yaml
         (auto-completion :variables auto-completion-return-key-behavior nil auto-completion-tab-key-behavior 'complete auto-completion-complete-with-key-sequence
                          (kbd "jk"))
         (helm :variables helm-ag-base-command "rg --vimgrep --no-heading")
         (python :variables python-test-runner 'pytest)
         (restclient :variables restclient-use-org t)
         (shell :variables shell-default-shell 'multi-term shell-default-term-shell "/bin/zsh" shell-default-height 50 shell-default-position 'top)
         (version-control :variables version-control-global-margin t))
```
- System configuration features: NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS MODULES THREADS

#### Backtrace :paw_prints:
```
N/A
```

#### Ripgrep
```
❯ rg --version
ripgrep 0.8.1 (rev c8e9f25b85)
+SIMD -AVX
```

gessen:
If a user uses ripgrep persistent configuration file that sets options
helm overrides the argument parser for ripgrep might incorrectly parse
the command line used by helm.

The problem is encountered whenever an argument option occurs multiple
times and the last one is of this form: '--arg value' or '-a v'. If that
last argument is followed by the positional argument (input) then its
value will be treated as the input resulting in error due to too many
positional arguments.

In other worlds, this would fail to be parsed correctly:
    rg --max-columns=10 --max-columns 20 input
    rg --max-columns 10 --max-columns 20 input

While this works as expected:
    rg --max-columns=10 --max-columns=20 input
    rg --max-columns 10 --max-columns=20 input

The solution would be to use arguments in the form of '--arg=value'
which will never be treated as a positional argument.

Fixes #10953

### DIFF
--- a/layers/+completion/helm/funcs.el
+++ b/layers/+completion/helm/funcs.el
@@ -164,9 +164,9 @@ If DEFAULT-INPUTP is non nil then the current region or symbol at point
   ;; --line-number forces line numbers (disabled by default on windows)
   ;; no --vimgrep because it adds column numbers that wgrep can't handle
   ;; see https://github.com/syl20bnr/spacemacs/pull/8065
-  (let* ((root-helm-ag-base-command "rg --smart-case --no-heading --color never --line-number")
+  (let* ((root-helm-ag-base-command "rg --smart-case --no-heading --color=never --line-number")
          (helm-ag-base-command (if spacemacs-helm-rg-max-column-number
-                                   (concat root-helm-ag-base-command " --max-columns " (number-to-string spacemacs-helm-rg-max-column-number))
+                                   (concat root-helm-ag-base-command " --max-columns=" (number-to-string spacemacs-helm-rg-max-column-number))
                                  root-helm-ag-base-command)))
     (helm-do-ag dir)))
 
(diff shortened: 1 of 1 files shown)


smile13241324:

PR #13147, opened 2020-01-02
Fix erc-log transient state - change it to be in erc-view-log-mode

mphojele:
With the fix I also added a utility function to open a erc log file/buffer from
a erc chat/channel buffer.


smile13241324:

PR #13152, opened 2020-01-04
Add var for custom xclipboard command

ricardolee:
There is a problem when I use clip.exe for utf8 character, add custom command for copy.


duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/591ab10be2de30035f55a54aab250d1fe3f31e7c

PR #7208, opened 2016-09-28
Fixes #7119 - Add support for Facebook Flow language

jdelStrother:
--------------------------------
**👉  There is a bounty source opened on this issue, click on the badge for more info**

[Bountysource](https://www.bountysource.com/issues/37889526-some-request-support-facebook-flow-language?utm_source=37889526&utm_medium=shield&utm_campaign=ISSUE_BADGE)

------------------------------

Feature list in order to complete this PR and unlcok the bounty:

**Auto-completion**
- [ ] The layer should activate `company-flow` whenever the buffer begins with either of these lines:
    - `// @flow`
    - `/* @flow */`

[Author note: It's currently enabled in all of react-mode and js2-mode, and I've not noticed any ill-effects in files where flow isn't present.]

**Syntax Highlighting**
- [ ] The layer should improve syntax highlighting for `flow`. Right now it's mostly ok in `react-mode` and quite bad in `javascript-mode` (for instance, use of generics `Map<K, V>` breaks the syntax highlighting). There may be some inspiration in `emacs-flow-jsx`, `react-mode` and `typescript-mode`.

**Documentation**
- [x] The type of current word should be rendered with `eldoc` -  `flow type-at-pos` should give the type.
- [x] bonus points for highlighted `eldoc` like in `typescript`.

**Code Navigation**
- [x] Using `flow get-def` and `flow find-module` should enable amazing jump to definition functionality. I'd love to have them binded to <kbd>g</kbd> or <kbd>d</kbd> or at least local leader.

-----------------------------

*Original message*

Hi there - I've cobbled this together from the various flow-related things I've been using. There seems to be some interest in having a proper spacemacs flow layer (#7119), so thought I'd throw this out for discussion.

There's a few bits that might be somewhat targeted towards my own setup - for example, I chain the flow flychecker onto the eslint checker, but conceivably others might want to chain it onto jshint instead.   How much of that would you want to see customizable before merging this?

-----------------------------

EDIT: [@syl20bnr] changed title and message body to reflect the bounty source attached to it.

PR #13157, opened 2020-01-05
Remove flow layer, suggest javascript + LSP

jdelStrother:
Following discussion in https://github.com/syl20bnr/spacemacs/pull/7208, we think we should remove the flow-type layer, and instead encourage people to use LSP in the regular javascript layer.

Any suggestions that might improve things documentation-wise?


smile13241324:

Issue #13160, opened 2020-01-06
Can't disable auto-indent when pasting after emacs-evil patch


pszynk:
This bug was present before #8897 and was fixed on develop #12240

But, after recent changes to emacs-evil (issue emacs-evil/evil#1199 and pr emacs-evil/evil#1222) 
`SPC u p` no longer pastes without indent, but rather pastes  4 times (which is an expected behavior from the look of emacs-evil creators)

This is rather troublesome. I'd argue that pasting 4 times after `SPC u p` makes little sense but it is consistent with standard emacs behavior. Still it would be nice to have this option of pasting without auto-indent. 

I would like to contribute and propose some fixes (I'm not to experienced in emacs/elisp).
I see two ways to approach this:
* abandon the idea of raw paste triggered by universal argument `SPC u`, maybe it's not the best idea to confuse it with the option of pasting n-times. Use a different key combination for raw paste i.e. `_p`
* check if universal argument `<= 0`, if so assume that user wants a raw paste. Pasting 0 or less times makes no sense either way. `SPC u - p` or `M-- p` would do raw paste.

pszynk:
Actually added a pr #13161 so `SPC u -<n> p` pastes `n` times without auto-indent. This allows for multiple non-auto-indent and doesn't clash to much with intended use of universal argument (at least as I see it)

smile13241324:
@pszynk your PR has been merged, can you double check whether this issue is still present and close this bug if it is solved please.

pszynk:
@smile13241324 yep, sorry for the hanging bug, closing.

PR #13161, opened 2020-01-06
Fix bug: disable paste auto-indent

pszynk:
Fix bug:  #13160
Using `SPC u p` for raw paste stopped working.
Allow for `SPC u -<n> p` to paste `n` not auto-indented lines.

(Fixes in documentation included)


duianto:
This doesn't seem to work, and it breaks pasting `p` in `.spacemacs`.

### In the `*scratch*` buffer (`text-mode`) (`SPC b s`)
#### Before and after applying this PR:
`p` pastes without taking the previous lines indentation into account. Probably because the scratch buffer is in `text-mode` which doesn't seem to auto indent.

`SPC u p` shows the following message:
> apply: Wrong type argument: number-or-marker-p, (4)

Copying `y` a whole line, by selecting it with `V` (uppercase), and trying to paste it with `SPC u p` in the `scratch` buffer, shows the message:
>evil-yank-line-handler: Wrong type argument: wholenump, (4)

I'm not sure but it might be an upstream evil issue.

#### After applying the PR:
`SPC u - p` shows the message:
>and: Wrong type argument: number-or-marker-p, -

### In `.spacemacs` (`emacs-lisp-mode`) (`SPC f e d`)
#### Before applying this PR:
`p` pastes and indents based on the indentation on the lines above.
`SPC u p` pastes without the above lines indentation (as it's written in the `kill-ring`).

#### After applying this PR:
`p` shows the message:
>and: Wrong type argument: number-or-marker-p, nil

`SPC u p` shows the same message as before applying the PR:
>and: Wrong type argument: number-or-marker-p, (4)

`SPC u - p` shows the message:
>and: Wrong type argument: number-or-marker-p, -

pszynk:
@duianto are you sure you tested this with the newest version of `emacs-evil` package? Because I would expect this kind of behavior when using pre emacs-evil/evil#1222.

One thing that I'm sure will happen is using `SPC u -<n> p` in mode without auto-indent option. In this case `-<n>` is just forwarded to `evil-paste-*` which gives an error (expected behavior). In those scenarios we could convert `-<n>` to `<n>` or to `1` to suppress the error. Not sure what is the best option.

duianto:
I update packages every day, so I should be on a evil commit later than that evil PR:
emacs-evil/evil#1222

The previous version I tested with in the previous comment was: `evil-20200103.1756`

I just updated and there was a new update now: `evil-20200113.1232`

I found an older evil version: `evil-20191219.1249`
It's from 25 days ago, that Evil PR was merged 22 days ago on Dec 22, 2019.

I'm still seeing the same behavior as in the previous test, both with and without the PR applied in both of these OSs:
- PopOS 19.10 (I just tested with the latest evil version)
- Windows 1903 (I tested with both the latest `evil-20200103.1756` and the older one: `evil-20191219.1249` from 25 days ago)

I didn't have any paste related user configurations, but I commented out everything in `dostpacemacs-user-config` anyway, to make sure that it wasn't a user config issue.

#### System Info :computer:

<details><summary>
PopOS 19.10
</summary>
<pre>
#### System Info :computer:
- OS: gnu/linux
- Emacs: 26.3
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. aa6a44684)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(auto-completion command-log emacs-lisp git helm javascript markdown multiple-cursors
                      '("~/org/notes.org"))
                 python
                 spell-checking syntax-checking treemacs version-control)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GSETTINGS GLIB NOTIFY ACL LIBSELINUX GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XDBE XIM THREADS LIBSYSTEMD LCMS2
</pre>
</details>

<details><summary>
Windows 1903
</summary>
<pre>
#### System Info :computer:
- OS: windows-nt
- Emacs: 26.3
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 3d88564bc)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(autohotkey auto-completion command-log emacs-lisp git helm html imenu-list javascript markdown multiple-cursors
                 '("c:/Users/username/org/notes.org"))
            python
            (shell :variables shell-default-shell 'eshell)
            spell-checking syntax-checking treemacs version-control)
```
</pre>
</details>

duianto:
Ok, I think I found the source. When commenting out the `mulitiple-cursors` layer.

### In the scratch buffer with or without the PR applied
`p`  pastes once
`SPC u p` pastes four times
`SPC u - p` doesn't paste and there's no message in the messages buffer. (probably as expected)

### In `.spacemacs`
#### Without the PR applied
`p` pastes with auto-indent
`SPC u p` pastes four times
`SPC u - p` doesn't paste and there's no message in the messages buffer. (probably as expected)

#### With the PR applied
`p` pastes with auto-indent
`SPC u p` pastes four times
`SPC u - p` pastes without auto-indent

pszynk:
Thanks @duianto, good catch! 
I've never used `multiple-cursors` layer. Tomorrow I'm gonna check out what's going on.

pszynk:
Well it looks like emacs-evil/evil#1222 broke `multiple-cursors` layer. `spacemacs/evil-mc-paste-*` functions are implemented with `(interactive "*P")`, while `evil-paste-*` from `emacs-evil/evil` is now declared with `(interactive "p<x>")`.

Combination `SPC u p` gives an error because `evil-paste-after` is called by `evil-mc-paste-after` with a list argument `(4)` but it expects an integer. `multiple-cursors` layer needs a fix.

BTW could someone explain to me is there any difference between `(interactive "p")` and `(interactive "p<x>")`? Is  `<x>` doing sth or is it just a naming convention?

bet4it:
This solution is OK, but `yank` also uses `(interactive "*P")`, so we can't handle them in a unified way.
Maybe it's better to suggest `evil` to use `(interactive "*P")` as well?

PS: the definition of `<x>` is in https://github.com/emacs-evil/evil/blob/d14e9e2cac539f388a1f4699eb3ffab0834aa3e4/evil-types.el#L332-L334.

pszynk:
All this only seems to complicate things. `yank` and `yank-pop` utilizes prefix argument in different ways. Adding a common functionality by interpreting a prefix argument value to decide whether to auto-format or not is asking for trouble. 

How about this approach:
1. Add a toggle option that switches on and off the auto-format by default for all this functions.
2. Add key bindings for `evil-paste-before` and `evil-paste-after` (for example `=P` and `=p`) that have an opposite behavior to the currently toggled one. I.e. if auto-format is switched on then `=p` pastes without auto-format, if auto-format is switched of then `=p` pasted with auto format.
3. Similarly for `yank` and `yank-pop`, but I'm not sure what key combination to propose, I'm not a vanilla emacs user.

duianto:
syl20bnr said that we should use the universal argument `SPC u p` for pasting once without indentation.

That does seem more resonable. If someone wants to paste four times then they can add a digit argument.

pszynk:
well that's not consistent with the usage of universal argument in `evil` (`evil-paste-*`) nor with `emacs`  (`yank*`) itself, so it brings only problems and confusion.

bet4it:
Solved in upstream. We may close this PR now.

duianto:
Confirmed.

With bet4it's upstream fix, it seems like the previous Spacemacs behavior has been restored. I don't remember what `SPC u p` did in `text-mode` or `fundamental-mode` buffers, but it might be working as expected.

`SPC u p` pastes:
- Without indentation in buffers that handle indentation (ex. `emacs-lisp-mode`).
- Four times in for example a `text-mode` or `fundamental-mode` buffer (see below).

In an `emacs-lisp-mode` buffer:
```elisp
;; source: copy both lines
abc
  def

;; target, with the cursor on the empty line below: (and t
(and t

     )

;; paste with: p
(and t

     abc
     def
     )

;; paste with: SPC u p
(and t

abc
  def
     )
```

In a `text-mode` or `fundamental-mode` buffer:
```text
;; source: copy both lines
abc
  def

;; target, with the cursor on the empty line below: (and t
(and t

     )

;; paste with: p
(and t

abc
  def
     )

;; paste with: SPC u p
(and t

abc
  def
abc
  def
abc
  def
abc
  def
     )
```

syl20bnr:
That's great news 👍 
@pszynk can you confirm and close the PR ?

Thank you for taking care of this.

pszynk:
Yep, sorry for the delay. I'm closing this PR. Still, I believe this inconsistency with evil can be a problem in the future. Thanks to all for helping with this issue.

PR #13164, opened 2020-01-07
[nixos] add toggle for company-nixos-options

bhipple:
On some Nix setups, company completion via the Nix repl process can be extremely
slow and hang emacs. This commit adds an option toggle to disable it in the
`nixos` layer. The default behavior is unchanged.

Also updates the comment and license headers to be consistent with other layers.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2380,6 +2380,8 @@ Other:
   function (thanks to Profpatsch)
 - Associate nix-mode with .nix files (thanks to jpathy)
 - Updated layer banner (thanks to kalium)
+- Added option =nixos-enable-company= to toggle company auto-completion (thanks
+  to bhipple)
 **** Nim
 - Added key binding ~SPC m h h~ to show symbol documentation
   (thanks to Valts Liepiņš)
--- a/layers/+os/nixos/README.org
+++ b/layers/+os/nixos/README.org
@@ -26,6 +26,14 @@ To use this configuration layer, add it to your =~/.spacemacs=. You will need to
 add =nixos= to the existing =dotspacemacs-configuration-layers= list in this
 file.
 
+* Configuration
+Toggle whether =company-nixos-options= completion is enabled (defaults to true).
+
+#+BEGIN_SRC emacs-lisp
+(setq-default dotspacemacs-configuration-layers
+              '((nixos :variables nixos-enable-company t)))
+#+END_SRC
+
 * Key bindings
 ** NixOS Options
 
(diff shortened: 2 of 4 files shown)


bhipple:
CC @duianto any chance you could take a look?  This should be a backwards compatible new feature with no change to default behavior; most of the diff in the PR is just formatting/license/standardization churn to try to keep the nixos layer code up-to-date with best practices.

duianto:
I don't have `nixos` to test with, but it seems fine.

Commit(s): https://github.com/syl20bnr/spacemacs/commit/d4a27babf67d0c702f33c52ca02cb30cb69ff5fa

syl20bnr:
@bhipple Can you test this ? https://github.com/syl20bnr/spacemacs/blob/develop/doc/DOCUMENTATION.org#disabling-layer-services-in-other-layers

Something like:

```elisp
     (auto-completion :disabled-for nixos)
```

If this works as expected, we can remove the option but keep an entry in the layer to warn user about along with the example above.

bhipple:
Sorry for the slow response; I've been on vacation. That does seem like a much cleaner, more consistent, and eaiser to maintain way of doing things. I'm having trouble getting it working, though; I think this is because the `nixos` layer specifically initializes `company`, so it always gets dragged in?  Looking at the examples of `org` and `git`, they don't seem to declare a dependency on the `auto-completion` layer, so I'm not sure how this works; any pointers would be appreciated.  This is what I'm trying:

```diff
diff --git a/layers/+os/nixos/packages.el b/layers/+os/nixos/packages.el
index acfb993d0..922548f2e 100644
--- a/layers/+os/nixos/packages.el
+++ b/layers/+os/nixos/packages.el
@@ -18,17 +18,15 @@
         nixos-options))

 (defun nixos/post-init-company ()
-  (when nixos-enable-company
     (let ((backends '(company-capf)))
       (when (configuration-layer/package-used-p 'company-nixos-options)
         (add-to-list 'backends 'company-nixos-options t))
       (eval `(spacemacs|add-company-backends
                :backends ,backends
-               :modes nix-mode)))))
+               :modes nix-mode))))

 (defun nixos/init-company-nixos-options ()
  (use-package company-nixos-options
-   :if nixos-enable-company
    :defer t))

 (defun nixos/init-helm-nixos-options ()
```
(plus the `auto-completion :disabled-for nixos` in my `init.el`).

BTW, this doesn't actually require NixOS to use/test; it just requires installing Nix on any Linux or Mac system:
https://nixos.org/nix/download.html

syl20bnr:
It is not really recommended to use `(global-company-mode t)`, it would be better to not use it and open an issue if auto-completion is missing somewhere so it can be configured explicitly in Spacemacs. This could be removed from documentation from my point of view, Spacemacs goes in great length to maintain proper isolation between packages configuration and using the company globally breaks that (we also lose control about when company should be used). Moreover if it breaks the functionality described above it should be considered as a dirty hack and it should not be in the documentation.

PR #13166, opened 2020-01-07
Fix void destructuring-bind after cl pkg deprecation

duianto:
`destructuring-bind` is a macro in the deprecated package: `cl`

`SPC-TAB` calls: `spacemacs/alternate-buffer`
In the `spacemacs-base` distribution, it shows the message:
>spacemacs/alternate-buffer: Symbol’s function definition is void: destructuring-bind

Also renamed three instances of `destructuring-bind` in the `rirc` layer.


smile13241324:
Merged

PR #13169, opened 2020-01-08
Spacemacs buffer in other window with prefix arg

duianto:
This makes `SPC u SPC b h` behave the same for the Spacemacs buffer, as it does for the scratch and messages buffers:
- `SPC u SPC b s` scratch buffer
- `SPC u SPC b m` messages buffer


smile13241324:
Merged :)

PR #13175, opened 2020-01-09
[persp] Bind `C-d` to project perspective and dired

magthe:
This change adds a key binding for having `Spc p l` jump to the project in _dired_.

`Ctl-d` is the binding for the corresponding action when using `Spc p p`.

### DIFF
--- a/layers/+spacemacs/spacemacs-layouts/funcs.el
+++ b/layers/+spacemacs/spacemacs-layouts/funcs.el
@@ -518,9 +518,17 @@ Run PROJECT-ACTION on project."
                projectile-known-projects))
      :fuzzy-match helm-projectile-fuzzy-match
      :mode-line helm-read-file-name-mode-line-string
+     :keymap (let ((map (make-sparse-keymap)))
+               (define-key map
+                 (kbd "C-d") #'(lambda () (interactive)
+                                 (helm-exit-and-execute-action
+                                  (lambda (project)
+                                    (spacemacs||switch-project-persp project
+                                      (dired project))))))
+               map)
      :action `(("Switch to Project Perspective" .
                 spacemacs//helm-persp-switch-project-action)
-               ("Switch to Project Perspective and Open Dired" .
+               ("Switch to Project Perspective and Open Dired `C-d'" .
                 ,(spacemacs//helm-persp-switch-project-action-maker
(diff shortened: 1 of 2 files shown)


nixmaniack:
This is good. Should be added to `ivy` as well though.

magthe:
I'm not using `ivy`, so I'm probably not the best person to add it there :worried:

nixmaniack:
@magthe If you can check and merge https://github.com/magthe/spacemacs/pull/1.

duianto:
With `helm`, `SPC p l` `C-d` works.

In `ivy`, `SPC p l` `M-o` shows the available actions:

emacs_2020-01-19_10-59-53

But when `C-d` is pressed then it shows:
>C-d is not bound

nixmaniack:
@duianto I forgot to update here. I had opened an issue at upstream to fix that issue. 
> `ivy-set-actions` doesn't seem to support `(kbd "C-d")` hence using evaluated
value.

The issue is fixed, let me update the PR. 

Also, regarding your opinion of using `d`, I was in the same camp but given that we need to keep bindings consistent in Spacemacs, I used `C-d`.

duianto:
I'm not sure how close the consistency rule should be followed for the completion actions.
For example:
- In `ivy`, copy the current candidate is `w` in the ivy actions `M-o`.
- In `helm` `C-c C-y` copies the current candidate. it doesn't have a separate actions "state", so single keys will just type at the prompt.

Since the ivy actions accepts single keys it seems more convenient to just press `d`, instead of `M-d`.

nixmaniack:
That makes sense. I'll update to use `d` then.

duianto:
If the only change that's need is to replace the current `^D` character with `d` in the line:
```elisp
             '(("" spacemacs/ivy-switch-project-open-dired "dired")))))
```
Then I can do it and apply this PR to the develop branch.
I have the PR loaded locally.

nixmaniack:
> Then I can do it and apply this PR to the develop branch.

@duianto Yes. You can do it that way. I'm fine with that.

duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/3496f0c1f2d74ca915e6eb320a53f79a171b4fe1

PR #13176, opened 2020-01-11
Fix centered-cursor-mode partially breaks Scrolling Transient State

madand:
When one activates Scrolling Transient State (`SPC N`) in a buffer with
`centered-cursor-mode` active, the keys `j` and `k` do nothing. This commit
disables `centered-cursor-mode` for the duration of Scrolling Transient State.


duianto:
This doesn't seem to work.

With `center-point` (`center-cursor-mode`) enabled: `SPC t -`

Nothing happens when `j` or `k` are pressed in the `SPC N j` scrolling transient state.

The `spacemacs//scroll-ts-on-enter` function seems to never get called.

I replaced the two lines:
```elisp
    :on-enter (spacemacs//scroll-ts-on-enter)
    :on-exit (spacemacs//scroll-ts-on-exit)
```
with:
```elisp
    :on-enter (progn
                (message ":on-enter SPACEMACS//SCROLL-TS-ON-ENTER")
                (spacemacs//scroll-ts-on-enter))
    :on-exit (progn
               (message ":on-exit SPACEMACS//SCROLL-TS-ON-EXIT")
               (spacemacs//scroll-ts-on-exit))
```
Nothing appears in the messages buffer when the scroll TS is opened,
with for example: `SPC N j`

But when the scroll TS is closed with `q` or `C-g`, then the exit message appears:
>:on-exit SPACEMACS//SCROLL-TS-ON-EXIT

I have tested with all of the scroll keys under the `SPC N` prefix, but none of them show the on-enter message.

---

It seems like there might be a risk of interpreting the variable name:
`spacemacs--scroll-ts-centered-cursor-active-p`
as if `center-cursor-mode` is active when the variable is `t`.

A more descriptive name might be something like:
`spacemacs--scroll-ts-on-exit-restore-centered-cursor`
`spacemacs--scroll-ts-restore-prev-centered-cursor-state`
`spacemacs--scroll-ts-on-enter-centered-cursor-was-active`
or something better.

madand:
I renamed the variable to `spacemacs--scroll-ts-on-enter-centered-cursor-was-active` and already force-pushed the change.

The other issue, regarding the `:on-enter` hook not being called, is a more tricky one. 

Everything worked for me when I made the PR because I activate the transient state via a custom binding to `spacemacs/scroll-transient-state/body` and this way `:on-enter` hook is called as expected. But, as @duianto has discovered, the `:on-enter` hook is ignored when transient state is activated via other sub-commands like `spacemacs/scroll-transient-state/evil-scroll-line-down` (`SPC N j`).

It feels like a bug in the transient state machinery and the `:on-enter` hook must be called regardless of how the transient state was activated. @justbur could you please give us your opinion as the original author of the transient state macro?

duianto:
>It feels like a bug in the transient state machinery

👍

madand:
I went on and investigated the issue further.

Spacemacs Transient State machinery is actually a (rather thin) wrapper around [hydra](https://github.com/abo-abo/hydra). The contents of the `:on-enter` property are being passed to hydra as its `:pre-body` hook, which is being called only when a `hydra-something/body` command is called. Hydra also provides a `:pre` hook, which is being called on _every_ hydra command invocation (basically every key press while the transient state is active). At the moment, hydra provides no hook that is being called only once, when the hydra session is started, regardless of the way it was started (by `hydra-something/body` or `hydra-something/sub-command`).

We can hack the solution on the Transient State side by utilizing both of the `:pre-body` and `:pre` hydra hooks. But I feel like the new hook might be also useful for other hydra users, so I will try to push the change upstream before hacking the Spacemacs-specific solution.

madand:
The upstream PR abo-abo/hydra#368 still awaits its fate. I've just pinged the maintainer.

madand:
Hi, @lebensterben. I am still willing to resolve this issue, if the upstream decides to merge my PR.

lebensterben:
@madand 
Great. That's also why I removed the `stale` label to prevent it from being closed the the bot.

smile13241324:

PR #13177, opened 2020-01-11
Add SPC x d l delete-blank-lines

duianto:
Bind `SPC x d l` to `delete-blank-lines` (default: `C-x C-o`)

If the cursor is on a blank line:
- with blank lines above or below, then they are reduced to only one.
- without blank lines above or below, then the blank line is removed.


smile13241324:
Nice function 👍, merged

PR #13181, opened 2020-01-13
Improve rainbow-identifiers experience for Elisp and CL

madand:
Make `rainbow-identifiers` not colorize special operators and macros, so they
always visually stand out. Rationale behind this change is that special
operators and macros in Lisp may be considered "syntax" elements, so it makes
sense to have them visually distinguished at all times.

lisp-rainbow-identifiers


duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/e1ad03551370643d82a4390c97802722e0cf3b0b

Issue #8671, opened 2017-04-10
access ghci history


amilkov1:
Is there a way to see/access your ghci history in the standard REPL that comes as part of the Haskell layer? (Essentially the equivalent of Ctrl-r in a shell or even just up arrow key to view recently submitted commands)

mrkgnao:
@amilkov1 does `C-k` work (or `C-c M-o` in Emacs mode, or maybe `M-n`/`M-p`)?

amilkov1:
`C-k` yields a `Text is read-only : #<buffer myfile*>
`M-n` `M-p` and `C-c M-o` don't do anything

ccressent:
up arrow, down arrow, `M-n`, `M-p`, `C-j` and `C-k` work as expected for me.
No special configuration necessary. I just `, s S` from a Haskell buffer to get a REPL and it all works. I'm using stack and intero if that makes a difference. Here is my Haskell layer configuration:

```emacs-lisp
(haskell :variables
         haskell-completion-backend 'intero
         haskell-enable-hindent-style "johan-tibell")
```

amilkov1:
Yup. Doesn't seem to work with `ghci` as the designated backend (just tried it) but intero does the trick

duianto:
The following PRs changes have been applied to the `develop` branch:
Add C-j, C-k, C-l to ghci repl in haskell layer #13184

PR #13184, opened 2020-01-14
Add C-j, C-k, C-l to ghci repl in haskell layer

Garmelon:
There was an issue on this previously (#8671), but it seems not to have been fixed.


duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/099ead9b6a031ca2265fc3b941577a7025264abd

#### Changes
The documentation entries:
```
| ~C-j~       | switch to previous history item                 |
| ~C-k~       | switch to next history item                     |
```
had the previous and next words on the opposite key bindings.
It was corrected.

PR #13188, opened 2020-01-16
Add persp-mode integration in treemacs layer

sei40kr:
## Description

Treemacs now supports Perspective as the Treemacs buffer scope.

[treemacs-persp] https://github.com/Alexander-Miller/treemacs/blob/master/src/extra/treemacs-persp.el
[discussion] https://github.com/Alexander-Miller/treemacs/issues/222

To set Perspective as the Treemacs buffer scope, execute `treemacs-select-scope-type` and choose `Perspective`.


nixmaniack:
@sei40kr Is there a way to define the default scope type via a variable instead of invoking and choosing?
I would like to have the default scope set to `Perspectives`(maybe a default in Spacemacs?).

duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/58279f89ead4dceb1db509048cf5d13b4c9d841e

Issue #13167, opened 2020-01-08
Missing command helm-dir-do-grep


mcfarljm:
#### Description :octocat:
Missing command `helm-dir-do-grep`.

It looks like it is almost just a matter of a missing definition of `spacemacs/helm-dir-do-grep` in completion/helm/funcs.el, although for some reason the function that this would call, `helm-files-do-grep`, does not accept a `dir` argument like the others.  Probably an easy fix.

#### Reproduction guide :beetle:
- Start Emacs on a system that does not have the advanced search utilities (e.g., ag) installed
- Run SPC-s-d

*Observed behaviour:* :eyes: :broken_heart:
`Wrong type argument: commandp, helm-dir-do-grep`

*Expected behaviour:* :heart: :smile:
A searching function called `helm-dir-do-grep` gets called

#### System Info :computer:
- OS: gnu/linux
- Emacs: 26.3
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 7636811b7)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: helm
- Layers:
```elisp
(emacs-lisp git helm html markdown multiple-cursors org treemacs)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GSETTINGS GLIB NOTIFY ACL GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XDBE XIM MODULES THREADS LIBSYSTEMD LCMS2

bbolvig:
same here

emacs18:
It looks like following gets the job done.

```elisp
diff --git a/layers/+completion/helm/funcs.el b/layers/+completion/helm/funcs.el
index 382384515..0fda42ae1 100644
--- a/layers/+completion/helm/funcs.el
+++ b/layers/+completion/helm/funcs.el
@@ -209,6 +209,11 @@ If DEFAULT-INPUTP is non nil then the current region or symbol at point
   (interactive)
   (spacemacs/helm-files-do-ack default-directory))
 
+(defun spacemacs/helm-dir-do-grep ()
+  "Search in current directory with `grep'."
+  (interactive)
+  (spacemacs//helm-do-grep-region-or-symbol (list default-directory) nil))
+
 (defun spacemacs/helm-dir-do-ack-region-or-symbol ()
[...]

PR #13194, opened 2020-01-18
created spacemacs/helm-dir-do-grep to fix issue #13167

emacs18:
This is used if someone does not have ag, pt, rgrep, etc.


duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/278be410108e07ee77f8dd700c778e19bb7976cf

PR #13195, opened 2020-01-19
rust: Guides to use rust-analyzer; fix link

lebensterben:
- Add instruction to choose `rust-analyzer` instead of `rls` as LSP server
- Clarify that even when `rls` is chosen as the backend, user still needs `Racer` for code completion
- Replace the obsolete link of `lsp-rust` with `lsp-mode`


duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/d7e9a803fdda7d594c6abeffbc222baf8c5a3884

Issue #13197, opened 2020-01-22
C-w C-w with Agenda List


Matelasse:
#### Description :octocat:
C-w C-w doesnt work when you have Agenda List open

#### Reproduction guide :beetle:
Start spacemacs and open some orgfile, also open agenda-list in a split window (M-x org-agenda-list). Now you want to move between the windows using C-w C-w, but it doesnt work. Emacs C-x o does work though.
*Observed behaviour:* :eyes: :broken_heart:
C-w C-w doesnt move between windows. C-x o does.

*Expected behaviour:* :heart: :smile:
C-w C-w should move between windows, like C-x.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 26.3
- Spacemacs: 0.200.13
- Spacemacs branch: master (rev. 26b8fe0c3)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm

duianto:
Confirmed, the `evil-state` in the agenda buffer is `evilified` where `C-w` isn't bound, therefore it seems to use the default Emacs command from the `global-map`, `kill-region`, but it has no effect since the agenda buffer is in read-only state. It doesn't seem like an appropriate command for the agenda buffer anyway.

A fix is to bind:
```elisp
      (kbd "C-w") 'evil-window-map
```
in the `org/init-org-agenda` functions:
```elisp
    :config
    (evilified-state-evilify-map org-agenda-mode-map
``` 
section.

But this would only fix it in the agenda buffer, I suspect that the same issue is occurring in most/all evilified buffers. So it would probably better to have a more "global" solution.

However I'll open a PR that fixes it in this case. If someone comes up with a general solution then it can be used instead.

nixmaniack:
> I suspect that the same issue is occurring in most/all evilified buffers. 

I have experienced this. 

> If someone comes up with a general solution then it can be used instead.

@duianto Here 👇https://github.com/syl20bnr/spacemacs/blob/3496f0c1f2d74ca915e6eb320a53f79a171b4fe1/layers/+distributions/spacemacs-bootstrap/local/evil-evilified-state/evil-evilified-state.el#L149-L152

Patch with this 👇? 
```elisp
(define-key evil-evilified-state-map (kbd "C-w") 'evil-window-map)
```

duianto:
That seems to be the right solution 👍 
@nixmaniack if want to open a new PR then I can test and merge it.

nixmaniack:
[...]

PR #13199, opened 2020-01-22
[evilified-state] Map C-w to evil-window-map

nixmaniack:
Fix https://github.com/syl20bnr/spacemacs/issues/13197

Motivation: https://github.com/syl20bnr/spacemacs/issues/13197#issuecomment-577051372
> the evil-state in the agenda buffer is evilified where C-w isn't bound, therefore it seems to use the default Emacs command from the global-map, kill-region, but it has no effect since the agenda buffer is in read-only state. It doesn't seem like an appropriate command for the agenda buffer anyway.


duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/0ab1e45fe468b63197779e8d0f68c97489344fd7

nixmaniack:
@duianto Thanks for updating changelog. I missed it.

duianto:
Np, I think it's marked as optional. The important thing is to suggest a fix/improvement.

PR #13205, opened 2020-01-24
Fix: exiting org-present presentation always removes inline images

Kazark:
Spacemacs configures `org-present` to inline images while presenting. That much
makes sense. But it has a hook to turn them off unconditionally when exiting
presentation mode. This causes a problem if you just want images to show in and
out of presentation mode.

# Steps to Reproduce
1. Create an Org file
2. Add a link to a file
3. Set `#+STARTUP: inlineimages`
4. `M-x org-present`
5. Exit presentation mode

# Expected
Because of `#+STARTUP: inlineimages` images are still inlined.

# Actual
The inline images have been removed upon exiting the presentation mode.

# Fix
Check `org-startup-with-inline-images` on exiting presentation mode and only
remove the inline images if it is false.

# Testing
To ensure that this did not introduce a regression of the converse behavior, I
also tested with `#+STARTUP: noinlineimages`, and it still turned off the images
when exiting the presentation.

### DIFF
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -692,7 +692,8 @@ Headline^^            Visit entry^^               Filter^^                    Da
       (defun spacemacs//org-present-end ()
         "Terminate `org-present' mode"
         (org-present-small)
-        (org-remove-inline-images)
+        (if (not org-startup-with-inline-images)
+            (org-remove-inline-images))
         (org-present-show-cursor)
         (org-present-read-write))
       (add-hook 'org-present-mode-hook 'spacemacs//org-present-start)


Kazark:
I see there is a check for `CHANGELOG` changes. Should I update the changelog for this?

duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/4ae7be5404bdf1bd914a027432a46df801b6ca49

The test for a changelog entry is marked as optional, but it can be helpful if the PR authors also add a changelog entry.

I added it for this commit.

In this case the changed file: `layers/+emacs/org/packages.el `
is in the `org` layer, therefore the changelog entry was added to the section:
```
*** Layer changes and fixes
**** Org
```

Some layers have one or more of these sub sections:
`Packages`, `Key bindings`, `Improvements`, `Fixed` and `Other`
Then the changelog entry can be added to the relevant sub section.

But in this case the `org` section only has a `Packages` and  `Key bindings` section near the top, therefore this was added to the end, on the line before the next layer: `Osx`

If it's unclear where it should be added, then it's fine to add it to the end of the `changelog.develop` file and mention in the PR that it wasn't clear where it should be added and a maintainer will move it.

Kazark:
@duianto Thanks for your explanation of the changelog, and for merging.

PR #13207, opened 2020-01-25
[purescript] Add to spacemacs-indent-sensitive-modes

kayhide:
Add `purescript-mode` to `spacemacs-indent-sensitive-modes` .

The syntax of PureScript is similar to that of Haskell, which is also registered as an indent sensitive mode.

Without this configuration, auto-indentation is triggered every time when pasting some chunk of code.

### DIFF
--- a/layers/+lang/purescript/packages.el
+++ b/layers/+lang/purescript/packages.el
@@ -37,6 +37,7 @@
     :defer t
     :init
     (progn
+      (add-to-list 'spacemacs-indent-sensitive-modes 'purescript-mode)
       (add-hook 'purescript-mode-hook 'turn-on-purescript-indentation)
       (add-hook 'purescript-mode-hook 'purescript-decl-scan-mode)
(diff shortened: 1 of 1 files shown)


duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/08442790cc7c7748b48d0d68b11e8ee74e566d5f

kayhide:
Okay, that sounds good.
Thanks!

bet4it:
Maybe it's better to add it here?
https://github.com/syl20bnr/spacemacs/blob/08442790cc7c7748b48d0d68b11e8ee74e566d5f/layers/%2Bspacemacs/spacemacs-defaults/funcs.el#L61-L76

duianto:
@bet4it thanks for the suggestion. At first glance it does seem like it would be better to define all modes at the same place. But I'll check with the maintainers if they have a preference.

Issue #13180, opened 2020-01-12
Some packages in .emacs.d/private/local are not found when loaded via dotspacemacs-additional-packages


dalanicolai:
#### Description :octocat:
Some packaged in `.emacs.d/private/local` directory are not found, while other packages are found. For example I have three files in the /private/local directory, [helm-recoll.el](https://github.com/emacs-helm/helm-recoll), [helm-tracker.el](https://github.com/gdkrmr/helm-tracker) and [ob-racket.el](https://github.com/xchrishawk/ob-racket), after adding the three packages to `dotspacemacs-additional-packages` the helm-recoll package is loaded correctly while my spacemacs complains that the helm-tracker and ob-racket packages are unavailable and asks if the package names are misspelled.

If the files are loaded interactively with `load-file` then all packages load correctly.

#### Reproduction guide :beetle:
- Start Emacs
Place the files [helm-tracker.el](https://github.com/gdkrmr/helm-tracker), [ob-racket.el](https://github.com/xchrishawk/ob-racket) and [helm-recoll.el](https://github.com/emacs-helm/helm-recoll) in `.emacs.d/private/local` directory, add `helm-recoll`, `helm-tracker` and `ob-racket` to `dotspacemacs-additional-packages`, then start spacemacs.

*Observed behaviour:* :eyes: :broken_heart:
At startup spacemacs messages that the packages helm-tracker and ob-racket are not available. And asks if their names are misspeled. On the other hand, the helm-recoll is loaded correctly. 
*Expected behaviour:* :heart: :smile:
Expected behavior is that all packages load correctly.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 26.3
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. d4a27babf)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(javascript scheme racket html octave python hy jupyter shell-scripts pdf latex prodigy
            (orgwiki :variables org-wiki-location "~/Sync/org/wiki")
            eww themes-megapack auto-completion better-defaults emacs-lisp git helm lsp markdown multiple-cursors
            (org :variables org-enable-org-journal-support t)
            shell spell-checking syntax-checking
            (treemacs :variables treemacs-use-follow-mode 'tag))
```
[...]

Issue #8421, opened 2017-02-22
How to add a package? (Beginner question)


feigaoxyz:
I have just started using Spacemacs and I am sorry if this is a stupid question.

The problem is that I want to use this [package](https://github.com/dpo/ampl-mode/blob/master/emacs/ampl-mode.el) for AMPL language support.

1. I have compiled the `ampl-mode.el` by `SPC SPC byte-compile-file` to generate the `ampl-mode.elc`
2. I put the two files in `~/.emacs.d/private/local/ampl-mode` folder
3. I have put/update the following in my `~/.spacemacs`

```
   dotspacemacs-additional-packages '(ampl-mode)
```

4. I open a `ex.ampl` file and try to activate `ampl-mode` with `SPC SPC ampl-mode`. Emacs reports "Cannot open load file: No such file or directory, ampl-mode".

I have also tried to put the files directly under the `.emacs.d/private/local`, but not working either.

What did I do wrong?

nixmaniack:
You need to load this package to be able to use it. `dotspacemacs-additional-packages` will only make sure that package is downloaded by spacemacs (and won't get deleted on next spacemacs run).

You can add following to your `user-config` section in `.spacemacs`to load the package. 
```elisp
(require 'ampl-mode)
```
In case you want to customize this package before using it, I would recommend `use-package`.

feigaoxyz:
thank you very much! `(require 'ampl-mode)` is working! I will also look into `use-package`

Issue #2278, opened 2015-07-07
Private Additional Packages possible?


nugend:
Hi, I've defined a syntax highlighting package for a language I use. It doesn't need to do anything fancy or have any additional configuration with other modes. My global defaults are perfectly fine with it.

Is there a way to use dotspacemacs-additional-packages to make spacemacs load it up? What I'm trying right now is putting the my-mode.el file into ~/.emacs.d/private and specify dotspacemacs-additional-packages '(my-mode)

That doesn't seem to work. Not *quite* sure where I'm going wrong.

jasoncyu:
i think `dotspacemacs-additional-packages` is for packages in a repo like melpa. Can you just `(load-file ...)` and `(require ...)` your package in `dotspacemacs/config`?

nugend:
That works, but would it make sense for dotspacemacs-additional-packages to also look for packages in private? I'd like to eventually make this available through MELPA (there isn't a good one already out there), but I figured it'd make sense to test it locally first.

Sorry if I'm not getting it. I'm pretty new to spacemacs and getting a hang of the emacs ecosystem.

endrebak:
I'm not exactly certain what you are trying to do, but I suspect `M-x configuration-layer/create-layer <RET>` is what you want.

That is what I have used when trying out packages/layers I've created myself. See https://github.com/syl20bnr/spacemacs/blob/master/doc/DOCUMENTATION.org#configuration-layers for more info.

pigmej:
create a layer and add your package as extension. (extension.el and extension folder).

endrebak:
A simpler way would be to set the "dont delete orphan packages bool" in ´.spacemacs´and then just install your desired package from paradox.

pigmej:
@endrebak you can do it already, just define `dotspacemacs-additional-packages`

The problem is with custom el packages that you have on disk.

endrebak:
Thanks, had never used it myself. Now I better understand what it is for.

StreakyCobra:
It is possible to add additional packages… well, in `dotspacemacs-additional-packages`. To use a local package, it is possible to use there the syntax presented in [LAYERS.org#packages.el](https://github.com/syl20bnr/spacemacs/blob/develop/doc/LAYERS.org#packagesel) or, as already explained, to use `(load-file …)` and `(require …)`.

This issue having been answered, and as there was no activity last 4 months, I'm closing it. Feel free to open a new issue if you have troubles with all this :relaxed:

[...]

PR #13209, opened 2020-01-25
Add recipes for additional packages README in private/local dir

dalanicolai:
Add detailed info for how to use additional packages as the (online) documentation for how to use single local packages is lacking. See issues #13180,  #8421 and #2278

Add detailed info for how to use additional packages as the (online) documentation for how to use single local packages is lacking. See issues #13180,  #8421 and #2278

Thank you <3

### DIFF
--- a/private/local/README.md
+++ b/private/local/README.md
@@ -2,19 +2,27 @@
 
 The content of this directory is ignored by Git.
 
-This is the place to store the local packages that you define in
-the `dotspacemacs-additional-packages` variable of your dotfile.
+This is the place to store the local packages that you define in the
+`dotspacemacs-additional-packages` variable of your dotfile.
 
-Additional packages can be added using the same recipe as for [adding packages to layers](https://develop.spacemacs.org/doc/LAYERS.html#packagesel) i.e.:
+Additional packages can be added using the same recipe as for [adding packages
+to layers](https://develop.spacemacs.org/doc/LAYERS.html#packagesel) i.e.:
 
-- if the package is on (M)ELPA simply add the package name to the list
+- For a local package:
+  - load the file explicitly, using the full path to the file, by placing a
+  `(require '~/.emacs.d/private/local/package-name)` within the body of the
+  `dotspacemacs/user-config` function of your dotspacemacs file.
+  - Alternatively create a directory with the name of the package in the
+  .emacs.d/private/local directory, and add that directory to the load-path
+  variable by adding `(some-package :location local)` to the list
+  `dotspacemacs-additional-packages` within the `dotspacemacs/layers` function
+  of your dotspacemacs file. After placing your package file into this
+  package-directory the file can be loaded, without requiring the full path, by
+  placing a `(require 'package-name)` within the body of the
+  `dotspacemacs/user-config` function of your dotspacemacs file.
 
-- for a local package add `(some-package :location local)` to the list and load or
-  require it explicitly in the user/config section. The first part only adds the
-  package directory (i.e. .emacs.d/private/local/package-dir/) to the load path
-  so that the .el file should be placed inside the package directory. If the
-  package provides a feature it can be loaded with require, otherwise it must be
-  loaded using load.
+- If the package is on (M)ELPA simply add the package name to the list
+  `dotspacemacs-additional-packages` in your dotspacemacs file
 
 - for a package hosted on github the recipe for github packages can be used i.e. add
 
@@ -24,5 +32,4 @@ Additional packages can be added using the same recipe as for [adding packages t
                              :repo "some/repo"))
 ```
 
-to the list. This will fetch the package on every startup, this can be
-prevented by installing the package locally as described in the previous point.
+to the list `dotspacemacs-additional-packages` in your dotspacemacs file.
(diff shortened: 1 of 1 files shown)


duianto:
Here are some possible minor changes:
 ```diff
diff --git a/private/local/README.md b/private/local/README.md
index 2277a43af..52463da5b 100644
--- a/private/local/README.md
+++ b/private/local/README.md
@@ -7,22 +7,23 @@ the `dotspacemacs-additional-packages` variable of your dotfile.
 
 Additional packages can be added using the same recipe as for [adding packages to layers](https://develop.spacemacs.org/doc/LAYERS.html#packagesel) i.e.:
 
-- if the package is on (M)ELPA simply add the package name to the list
+- If the package is on (M)ELPA simply add the package name to the list.
 
-- for a local package add `(some-package :location local)` to the list and load or
-  require it explicitly in the user/config section. The first part only adds the
-  package directory (i.e. .emacs.d/private/local/package-dir/) to the load path
-  so that the .el file should be placed inside the package directory. If the
-  package provides a feature it can be loaded with require, otherwise it must be
-  loaded using load.
+- For a local package, add `(some-package :location local)` to the list and load
+  or require it explicitly in the `dotspacemacs/user-config` section. The first
+  part only adds the package directory to the load path:
+  i.e. `.emacs.d/private/local/package-dir/`
+  Therefore the `.el` file should be placed inside the package directory. If the
+  package provides a feature, then it can be loaded with require. Otherwise it
+  must be loaded using `load-file`.
 
-- for a package hosted on github the recipe for github packages can be used i.e. add
+- For packages that are hosted on GitHub, use the recipe for GitHub packages by
+  adding the following to the `dotspacemacs-additional-packages` list:
 ```
 ```diff
-(some-package :location (recipe
-:fetcher github
-                             :repo "some/repo"))
+(some-package :location (recipe :fetcher github
+                                :repo "some/repo"))
 ```
 ```diff
-to the list. This will fetch the package on every startup, this can be
-prevented by installing the package locally as described in the previous point.
+This will fetch the package on every startup, this can be prevented by
+installing the package locally as described in the previous point.
```

#### Questions
- The sentence:
>If the package provides a feature it can be loaded with require, otherwise it must be loaded using load.

Is `using load` referring to `using load-file`?
If yes then it might be clearer to spell it out.

>If the package provides a feature, then it can be loaded with require. Otherwise it must be loaded using `load-file`.

- The last paragraph doesn't seem to be correct, or maybe I'm missing something?
>This will fetch the package on every startup, this can be prevented by
installing the package locally as described in the previous point.

When I install a package using the GitHub recipe:
```elisp
     (evil-repeat-motion :location (recipe :fetcher github
                                           :repo yyoncho/evil-repeat-motion))
```
Then the package only gets installed once, it doesn't try to install it again on the next restart.

dalanicolai:
Thanks for the review and sorry for the late reply. Except for the change from `load` to `load-file`, I agree with all the suggested modifications.

About your first question, I guess `load` and `load-file` can both be used. However, the important point is that is can be done by adding it to the body of the dotspacemacs/user-config function. Referring to load-file might suggest that is should be done interactvely. But I think it should be stated more explicitly that the `load` should be placed also in the body of the dotspacemacs/user-config function.

About the loading every time on startup, it seems that you are right that it is installed only once. However I remember that the org-server had some problems the day before I created this pull request, and because of that starting spacemacs took very long as the initialization was hanging on trying to reaching that server, even though I was just connecting to the emacsclients (if I remember well not sure about this... btw I mentioned the slow startup on gitter where Grant Shangreaux confirmed that he also experienced this issue). Anyway, I concluded that fetching packages form servers was not an ideal approach although fetching from github might be handled differently than the fetching of org-packages.

duianto:
`The first part` should probably be clarified what it's referring to.

And the mention that load and require should be declared in the `dotspacemacs/user-config` section is mentioned twice. It's enough to mention it once.

duianto:
Maybe the `README.md` could say something like the text below the horizontal line:


This was written in the list item starting with `for a local package add:`
>The first part only adds the package directory (i.e.
`.emacs.d/private/local/package-dir/`) to the load path. The .el file should be
placed inside this package directory.

But I'm not sure what `The first part` is referring to.

---

# Private directory for local packages

The content of this directory is ignored by Git.

This is the place to store the local packages that you define in the
`dotspacemacs-additional-packages` variable of your dotfile.

Additional packages can be added using the same recipe as for [adding packages to layers](https://develop.spacemacs.org/doc/LAYERS.html#packagesel) i.e.:

- if the package is on (M)ELPA, then simply add the package name to the
  `dotspacemacs-additional-packages` list.

- for a local package add:
```
(some-package :location local)
```
to the `dotspacemacs-additional-packages` list.

Then in the `dotspacemacs/user-config` function in your dotfile:
- If the package provides a feature, add:
```
`(require 'package-name)`
```
- otherwise load the package with:
```
(load "package-name")
```

- for a package that's hosted on GitHub, use the recipe for GitHub packages,
  i.e. add the following to the `dotspacemacs-additional-packages` list:
```
(some-package :location (recipe
                         :fetcher github
                         :repo "some/repo"))
```

smile13241324:
As wished closed in favor of #13381.

PR #13211, opened 2020-01-25
[nixos] Update recommended method for disabling nixos-company-options

bhipple:
Per discussion in #13164, we have a standard, consistent technique for doing
this across any layer. Provided that `global-company-mode` is not set, this
works as expected to disable company without requiring custom variables.


layers/+os/nixos/layers.el
```diff
@@ -0,0 +1,12 @@
...
+;; Copyright (c) 2015-2018 Sylvain Benner & Contributors
+;;
+;; Author: Sylvain Benner <sylvain.benner@gmail.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+(configuration-layer/declare-layer-dependencies '(auto-completion))
```
smile13241324: This should not be necessary, normally users can disable auto completion as well as syntax checking globally by not loading the respective configuration layers. In your case you are enforcing `auto-completion` globally for all layers when `nixos layer` is loaded.

Please remove the layer reference. In the case of it not being present post init for company and company-nixos-options will not be loaded, so this should not require changes in the package.el code.

bhipple:
Thanks for the careful review and the tip.  I've removed the layer reference and confirmed that it does indeed omit company from loading.

While touching this file I've tidied up the whitespace/indent to be consistently 2 spaces, but if you prefer I can minify the diff to just the functional bits.

smile13241324:

PR #13212, opened 2020-01-26
[auto-completion] Update documentation on global-company-mode

bhipple:
Reword it to note that this is strongly discouraged, and include a link to the
layer opt-out mechanism that will be broken if this is enabled.

See comment here for details:
https://github.com/syl20bnr/spacemacs/pull/13164#issuecomment-578410362


smile13241324:

PR #13213, opened 2020-01-27
[haskell] Fix REPL stuck snippets

duianto:
The snippets seem to have been accidentally changed in this commit:
https://github.com/syl20bnr/spacemacs/commit/a2de9a63afcdab6fcb0a0c2bdedb35a91556394b#diff-75e0388987cdf040d82762426e40e5a5R434

where other instances of checking for a layer, were changed to check for a
package.

It doesn't work to check if the haskell package is used in the user config
section, because the haskell package hasn't loaded yet.

Reverting it to check if the haskell layer is used works.


smile13241324:
Merged 👍

PR #13215, opened 2020-01-27
[Clojure] Sesman and missing eval / format keybindings

jr0cket:
Added keybindings for the Sesman REPL session management commands to the Clojure
layer, following the Spacemacs keybinding conventions.
https://docs.cider.mx/cider/0.23/usage/managing_connections.html

Sesman has been a part of CIDER for many versions now and provides session
management for nREPL connections (in fact any connections).

Sesman keybinds are placed in the `SPC m c` menu, providing a cleaner separation
between sending code to the repl and session management.

See issue #12593 and PR #13140 which covers the background for the design of
these additional keybindings.

Formatting the region and defun keybinds have been added, along with formatting
for edn files.

Evaluation up to point keybinding has also been added.

Documentation in the README has been updated to complete the description of the
existing Clojure layer, as well as including details specific to the new
keybindings.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1464,6 +1464,44 @@ Other:
   - Removed ~SPC m h g~ for =grimoire= as it is deprecated (thanks to Sam Hedin)
   - ~C-return~ to =cider-repl-newline-and-indent= in REPL evil insert state
     (thanks to Gia Thanh Vuong)
+  - changed clear repl buffer of evaluation to match terminal clear key
+    ~SPC s s l~ 'cider-repl-clear-buffer
+    ~SPC s s L~ 'cider-find-and-clear-repl-output
+    (thanks to John Stevenson)
+  - Add sesman session management keybindings to ~SPC m m~
+    ~SPC m m b~ 'sesman-browser
+    ~SPC m m i~ 'sesman-info
+    ~SPC m m g~ 'sesman-goto
+    ~SPC m m l b~ 'sesman-link-buffer
+    ~SPC m m l d~ 'sesman-link-directory
+    ~SPC m m l p~ 'sesman-link-project
+    ~SPC m m l u~ 'sesman-unlink
+    ~SPC m m q q~ 'sesman-quit
+    ~SPC m m q r~ 'sesman-restart
+    ~SPC m m s~ 'sesman-start
+    ~SPC m m S j~ 'cider-connect-sibling-clj
+    ~SPC m m S s~ 'cider-connect-sibling-cljs
+    (thanks to John Stevenson)
+  - updated repl shortcut to use sesman-start wrapper, selecting any repl type
+    ~SPC m '~ 'sesman-start
+  - changed toggle between source and repl to match key for toggle between code
+    and test
+    ~SPC m s a~ (if (eq m 'cider-repl-mode)
+                     'cider-switch-to-last-clojure-buffer
+                   'cider-switch-to-repl-buffer)
+    (thanks to John Stevenson)
+  - added formatting command not previously included
+    ~SPC m = =~ 'cider-format-buffer
+    ~SPC m = e b~ 'cider-format-edn-last-sexp
+    ~SPC m = e e~ 'cider-format-edn-last-sexp
+    ~SPC m = e r~ 'cider-format-edn-region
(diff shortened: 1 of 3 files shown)


CHANGELOG.develop
```diff
@@ -1463,6 +1463,42 @@ Other:
...
+    ~SPC m '~ 'sesman-start
+  - changed toggle between source and repl to match key for toggle between code
+    and test
+    ~SPC m s a~ (if (eq m 'cider-repl-mode)
+                     'cider-switch-to-last-clojure-buffer
+                   'cider-switch-to-repl-buffer)
+    (thanks to John Stevenson)
+  - added formatting command not previously included
+    ~SPC m f d~ 'cider-format-defun
+    ~SPC m f e b~ 'cider-format-buffer
```
yyoncho: formatting should be:

`,==` - format buffer
`,=r` - format region.

jr0cket: Oh yes, something else that has not been consistent for a long time.  I will update these keybindings as well.  Thanks for spotting that.

https://github.com/syl20bnr/spacemacs/blob/develop/doc/CONVENTIONS.org#code-formatting

John.

jr0cket: Formatting menu updated to use the Spacemacs convention, `=` instead of `f`

layers/+lang/clojure/README.org
```diff
@@ -335,6 +440,32 @@ As this state works the same for all files, the documentation is in global
...
+
+| Key binding   | Description                                                      |
+|---------------+------------------------------------------------------------------|
+| ~SPC m c b~   | browse all REPL session (sesman-browser)                         |
+| ~SPC m c i~   | current REPL information, ~SPC u~ for all sessions (sesman-info) |
+| ~SPC m c g~   | go to most relevant REPL session (sesman-goto)                   |
+| ~SPC m c l b~ | link buffer to REPL session (seman-link-with-buffer)             |
+| ~SPC m c l d~ | link directory to REPL session (seman-link-with-directory)       |
+| ~SPC m c l p~ | link project to REPL session (seman-link-with-project)           |
+| ~SPC m c l u~ | unlink from REPL session (seman-unlink)                          |
```
rakyi: Typos: seman -> sesman.

jr0cket: Thanks for spotting that...  now fixed.

jr0cket:
Thanks to everyone who has reviewed this pull request and made it better.  All you efforts are
very much appreciated.

Everyone I've reached out to in the Clojure community seems happy with these keybinding updates, so it would be great to get them merged over the weekend if the maintainers have time.  If there is anything you would like me to change, please let me know.  I appreciate this change is a little bigger than normal.

syl20bnr:
@jr0cket 

That's a great PR and thank you for taking time to update the key bindings.
Unfortunately `SPC m c` is reserved for compilation purpose :-)

I propose to use `SPC m m` for `session Management`, which gives `, m` with the shortcut. No need to change the PR I can update it, just tell me if this would be OK to use `SPC m m`. 

Thanks again for the great contribution.

jr0cket:
@syl20bnr sure, `SPC m m` is okay.

There is no explicit compilation step in Clojure, code is compiled as it is evaluated.  So there would never be any compilation functions.  However, I do understand that if coming from another layer that does have compilation, then using `SPC m c` for a different purpose would be confusing.

Thank you.

jr0cket:
I've pushed an update to the PR with `SPC m m` as the sesman keybindings rather than `SPC m c`. I wanted to test the keybindings out anyway, just to make sure my fingers are happy :)

xificurC:
Adding keybindings is one thing, moving everything around is just annoying. I start a session with `, '`, and instead of starting a REPL I get a prompt. Then I hit `, s s` to switch to the REPL buffer, binding doesn't exist. So I try to find the name of the command that used to be there, find it under `, s a` (why?). Well whatever. I hit `, s x` and again, no such binding. Now I can't guess the name of the command so I have to use magit's time machine to find the offending commit, find the name of the previous command and find it under `, e n`.

Please please please think about backward compatibility when making changes. You are breaking people's muscle memories. If you want to bind switch-to-repl to `, s a` to be compatible with other major modes I understand, but for the old-timers you can also keep the old bindings around.

PR #13217, opened 2020-01-28
Add layer music/pianobar

cyberthal:
Configuration instructions are more detailed than usual, because it requires unusual effort to piece together otherwise.

This should save users a lot of keystrokes and/or configuration fiddling.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -361,6 +361,8 @@ sane way, here is the complete list of changed key bindings
 - ietf (thanks to Christian Hopps and Robby O'Connor)
 - multiple-cursors (thanks to Codruț Constantin Gușoi and Thomas de Beauchêne)
 - parinfer (thanks to Tianshu Shi)
+**** Music
+- pianobar (thanks to Leo Littlebook)
 **** Readers
 - epub (thanks to Jeremy Dormitzer and André Peric Tavares)
 **** Spacemacs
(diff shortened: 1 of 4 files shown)


syl20bnr:
Nice one. Should rename the layer `pandora` since `pianobar` seems to be a fancy name for a Pandora package :-)

cyberthal:
I cannot recommend this change, because it will generate significant confusion.

The package is named pianobar, not pandora. The pianobar.el package runs a pianobar application that interacts with Pandora. pianobar.el only interacts with the pianobar application, not directly with Pandora. All the Emacs functions and variables are prefixed with pianobar.

Thus it would be quite confusing to break convention by naming the layer pandora. Users would need to know where the layer ends and the package begins, which is non-obvious.

cyberthal:
I checked the spotify layer to verify my reasoning. It fundamentally differs from the pianobar layer. The spotify layer interacts directly with Spotify API. It does so through addon packages for helm and counsel, both of which are named -spotify. Thus "spotify" is the consistent name to use throughout.

The same convention applied to the pianobar layer rules that everything be named "pianobar", not "Pandora". Nothing is named Pandora. There are other Pandora clients, and the pianobar layer doesn't work with them.

syl20bnr:
> The package is named pianobar, not pandora. The pianobar.el package runs a pianobar application that interacts with Pandora. pianobar.el only interacts with the pianobar application, not directly with Pandora. All the Emacs functions and variables are prefixed with pianobar.

I see, makes sense.

syl20bnr:
Thank you ! 👍

PR #13220, opened 2020-01-28
treemacs: fix default sorting

michaelpj:
I would expect the default sort order in treemacs to be alphabetically
ascending, and indeed until recently it appeared to be so (it has recently 
started being the opposite for me!). But the
setting in spacemacs is `alphabetic-desc`, and has been ever since the
layer was introduced. How did this work?

The answer is https://github.com/Alexander-Miller/treemacs/pull/577.
Until recently the alphabetical ascending/descending orders were swapped
accidentally. So we need to now (more logically) set the sort order to
`alphabetic-asc`.

After that, we are setting a number of variables to their default value (which stops us from following upstream, which would have saved us from this bug), or setting non-existent variables (which does nothing), so I removed all of those.

### DIFF
--- a/layers/+filetree/treemacs/packages.el
+++ b/layers/+filetree/treemacs/packages.el
@@ -43,7 +43,7 @@
             treemacs-silent-refresh nil
             treemacs-indentation 2
             treemacs-change-root-without-asking nil
-            treemacs-sorting 'alphabetic-desc
+            treemacs-sorting 'alphabetic-asc
             treemacs-show-hidden-files t
             treemacs-never-persist nil
             treemacs-goto-tag-strategy 'refetch-index)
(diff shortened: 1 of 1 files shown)


duianto:
Commit(s):
treemacs: fix default sorting https://github.com/syl20bnr/spacemacs/commit/c8bc98a48ece3c04fb6198e6a8ade99fbb28c4e4
treemacs: don't set variables to their default https://github.com/syl20bnr/spacemacs/commit/6d0120c417a55ce21e650f9b91b8edd891ff328a
treemacs: don't set non-existent options https://github.com/syl20bnr/spacemacs/commit/fed397cea9fb0be32792ebac2464d8d32234db38
Update changelog.develop https://github.com/syl20bnr/spacemacs/commit/c2e8a98641a7ad653befe61d3544f6b4f3e8c847

PR #13228, opened 2020-01-30
Replace org-toggle-latex-fragment with org-latex-preview

tshu-w:
`org-toggle-latex-fragment` is an obsolete command (as of Org 9.3); use `org-latex-preview` instead.


smile13241324:
Thank you for contributing to spacemacs 💜 I have cherry picked your commit into develop. You can safely delete your branch.

PR #13236, opened 2020-02-02
[lsp] set jump handler per lsp-managed buffer

Stebalien:
Previously, this would try to add the lsp jump handler to the mode's jump handler list every time lsp started managing a buffer of that mode. Now, we only use the lsp jump handler for buffers managed by lsp.


smile13241324:

PR #13238, opened 2020-02-03
[base] Add cl- prefix to letf and letf*

duianto:
The `cl` package has been deprecated.

It had aliases for `cl-` prefixed commands without the `cl-` prefix.
Ex: `letf` or `letf*` instead of `cl-letf` or `cl-letf*`

On the `spacemacs-base` distribution, with the `ivy` layer.
When one tries to search with `SPC /`
(which calls `spacemacs/counsel-search`)

Then this error message is shown:
>spacemacs/counsel-search: Symbol’s function definition is void: letf*

It doesn't happen with the `helm` layer,
because it's search commands already has the `cl-` prefix.

There are also three instances of: `letf`
in the `spacemacs-editing-visual` layers functions:
- `spacemacs/toggle-centered-buffer` (`SPC w c c`)
- `spacemacs/toggle-distraction-free` (`SPC w c C`)
- `spacemacs/centered-buffer-transient-state` (`SPC w c .`)

Without the `cl-` prefix they show the error message:
>Symbol’s function definition is void: letf

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1266,6 +1266,7 @@ Other:
   - Fixed =evilified-state=, mapped ~C-w~ to =evil-window-map=
     (thanks to Muneeb Shaikh)
   - Fixed ~SPC h f~ =helm-spacemacs-help-faq= (thanks to duianto)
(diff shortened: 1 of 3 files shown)


smile13241324:
Merged

PR #13242, opened 2020-02-05
fix: ignore nil values in dotspacemacs-configuration-layers

agzam:
otherwise it throws warnings when layers introduced imperatively, e.g.:

    `(
      ,(when (eq system-type 'darwin) 'osx)
      ,(when (eq system-type 'gnu/linux) 'exwm))


duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/7f92a4388d3a3b7b928f25a088a5a46e8a4dfce1

PR #13246, opened 2020-02-06
Layout transient state: make jump-to-last-layout exit state when done

thanhvg:
Make layout transient state quit on `SPC l tab`

### DIFF
--- a/layers/+spacemacs/spacemacs-layouts/packages.el
+++ b/layers/+spacemacs/spacemacs-layouts/packages.el
@@ -180,7 +180,7 @@
         ("C-8" spacemacs/persp-switch-to-8)
         ("C-9" spacemacs/persp-switch-to-9)
         ("C-0" spacemacs/persp-switch-to-0)
-        ("<tab>" spacemacs/jump-to-last-layout)
+        ("<tab>" spacemacs/jump-to-last-layout :exit t)
         ("<return>" nil :exit t)
         ("TAB" spacemacs/jump-to-last-layout)
         ("RET" nil :exit t)


nixmaniack:
May I ask motivation for this change?

There's a [workflow which might break](https://xkcd.com/1172/) wherein a user invokes layout transient state to peek-and-compare between two layouts. `SPC l <tab> <tab> [pause] <tab> <tab> ...`. It might just be me but I have used that workflow! Although with new change that can be adjusted to `SPC l <tab>  SPC l <tab> [pause] ...`. 

I'd like to know the reason and let's wait for other users' opinion before we merge this.

thanhvg:
Yeah this change won't be good for your workflow at all :(. You can use `spc l` then `n` will be just your current `tab` to go forward. `p` or `N` to go backward.

For me `tab` is like switch back to what ever last thing without thinking what it is, like `spc tab`.

duianto:
It seems like `SPC l TAB` should close the TS for consistency and convenience. Because that's what happens when `SPC l h` is pressed to open the default layout.

It would feel the same as when switching to the:
- last buffer: `SPC TAB`
- last window: `SPC w TAB`

No further action needs to be taken.

Like thanhvg said, peeking at layouts forwards can be accomplished with `SPC l n` and one can also peek backwards with `SPC l p` or `SPC l N`.

@nixmaniack what do you think?

nixmaniack:
`SPC l h` would be a special case to deterministically go to home layout - so exiting there makes sense I think.

`SPC w TAB` isn't part of TS but if we add that to `SPC w .` then we need to think what it should do - `exit` or `not`. 

> Like thanhvg said, peeking at layouts forwards can be accomplished with `SPC l n` and one can also peek backwards with `SPC l p` or `SPC l N`.

`SPC l TAB` is useful for peek-and-compare **recently** visited layout, hence `n/p/N` aren't much useful if I'm switching between let's say layout 1 and 4 out of 6 layouts. 

There are workspaces(`SPC l w`) as well which would need similar change if we decide to `exit` out of TS when switching to last workspace.

_I'm fine with this change if it brings consistency across (buffers/layouts/workspaces/windows/frames) navigational transient states which could potentially become a Spacemacs convention._ 

Convention along the following lines: 

- select with numbers/alphabet `[0-9a-z]`
- **select with numbers/alphabet `C-[0-9a-z]` no-exit** (This is currently being followed)
- select first `^`
- select last `$`
- select next `n`
- select previous `p/N` (exit)
- select recently/last visited `TAB`
- **select recently/last visited `C-TAB` no-exit**
...
...

So maybe, a normal command if preceded with `C-`, it doesn't exit the transient state.

duianto:
>SPC l TAB is useful for peek-and-compare recently visited layout, hence n/p/N aren't much useful if I'm switching between let's say layout 1 and 4 out of 6 layouts.

Your right.

>select first ^
select last $

Those might be useful additions.
A recent issue or PR mentioned using more than 9 layouts. They might help with navigating between them.

>select previous p/N (exit)

`(exit)` at the end seems like a typo?

>select recently/last visited C-TAB no-exit

Interesting, that would match the current `0..9` vs `C-0..C-9` bindings.

So if we choose to exit with `SPC l TAB` then that could be a solution.

nixmaniack:
> (exit) at the end seems like a typo?

Yes, Typo!

smile13241324:

PR #13247, opened 2020-02-06
toggle for doom-modeline package

thanhvg:
Even though we have a check if doom is actually set in `dotspacemacs-mode-line-theme` in doom mode line `init` function. `doom-modeline` still adds a bunch functions to hook `after-change-functions` regardless of the value of mode line theme name. So some parts `doom-modeline` still run even when users don't use it. 

Hence we need to put a toggle on the declaration so `doom-modeline` only gets installed when it is actually in use.


smile13241324:

PR #13254, opened 2020-02-10
Python - Change self compiled mspyls directory according to system type

honuonhval:
Microsoft's python-language-server compiles (publish) to different directories depending on the system used.


smile13241324:

PR #13260, opened 2020-02-11
Rename c++-enable-organize-includes-on-save to c-c++-enable-organize-includes-on-save

eklitzke:
The logic that includes and enables the `cpp-auto-include` package is guarded by the variable `c++-enable-organize-includes-on-save`. This name kind of makes sense because it's only enabled in `c++-mode`, but on the other hand it doesn't make sense because naming the config variable and related functions with a leading `c++-` instead of `c-c++-` breaks the naming convention used by the layer.

This is kind of a subjective matter, but I think using a common naming convention for variables and functions within the layer is important and the variable/functions should thus be named accordingly. The `cpp-auto-include` stuff is still only enabled in `c++-mode` with this change, only the names of things have changed.

This also fixes a typo about the variable in CHANGELOG.develop (which should be fixed even if this change is rejected).

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1337,7 +1337,7 @@ Other:
     (thanks to Alexander Dalshov)
 - New packages:
   - =cpp-auto-include= Insert and delete C++ header files automatically:
-    - Added variable: =++-enable-organize-includes-on-save=
+    - Added variable: =c-c++-enable-organize-includes-on-save=
     - Added key binding: ~SPC m r i~ organize includes
 - Improvements:
   - Added =lsp= support using either =clangd=, =ccls= or =cquery= backends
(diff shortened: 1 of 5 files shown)


smile13241324:
Hmmm this change could be controversial, independent of the question whether we use the same prefix for each variable, or try to communicate to which part of the language that variable refers to using prefixes, just changing the variable would surely break existing configs.

I would vote for at least an alias for the old name. Anyway lets see what @syl20bnr says.

smile13241324:
This PR should be kept alive, keeping vars consistently named is important. Lets do this, anyway I will add an alias not to break old configs though.

smile13241324 (approved):
Looks good

smile13241324:

PR #13262, opened 2020-02-11
Coq layer: toggle the electric terminator setting.

kindaro:
Make a new binding for the Coq layer.

This binding toggles the electric terminator setting of Proof General.

Strategically, I propose we add all kinds of settings in the same place.

### DIFF
--- a/layers/+lang/coq/README.org
+++ b/layers/+lang/coq/README.org
@@ -127,3 +127,9 @@ Note the last two are regular =company-coq= bindings, left alone since they are
 most useful in insert mode. The full =company-coq= tutorial showcasing all
 available =company-coq= key bindings can be accessed at any time using =SPC SPC
 company-coq-tutorial=.
+
+** Options
+
+| Key binding     | Description                 |
+|-----------------+-----------------------------|
+| ~SPC m T e~     | Toggle electric terminator. |
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1564,6 +1564,9 @@ Other:
   - Fixed ~SPC m e~ key bindings to behave like in Emacs Lisp
     (thanks to Boris Avdeev)
   - Fixed initialization of =counsel-gtags= (thanks to Sylvain Benner)
+**** Coq
+- Key bindings:
+  - ~SPC m T e~ Toggle electric terminator (thanks to Ignat Insarov)
 **** Cscope
 - Key bindings:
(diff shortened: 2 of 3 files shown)


layers/+lang/coq/README.org
```diff
@@ -123,6 +123,12 @@ The mnemonic for =a= is "ask".
 | ~SPC m i t~     | Insert a tactic                                                 |
 | ~SPC m i T~     | Insert a tactical                                               |
 
+** Options
+
+| Key binding     | Description                 |
+|-----------------+-----------------------------|
+| ~SPC m o .~     | Toggle electric terminator. |
```
smile13241324: Please change the keybinding to `SPC m T t`.
`SPC m T` is reserved for major mode specific toggles, that's also the place where all toggles for the layer should be kept.
`.` is also reserved, the conventions document says it needs to start a transient state i.e. `SPC w .` starting the window transient state.

See below the areas of CONVENTIONS.org I am refering to:

> *** Transient-state
Transient states are generally put behin the dot key `.`, for instance the
windows manipulation transient state in `SPC w .`.
When it is not possible to use the `.` (like in =helm= buffers) transient states
should be enabled with `M-SPC` and `s-M-SPC`. We need the latter bindings on
macOS since `M-SPC` is used by the OS for spotlight.
It is recommended to make sure that `q` allows to leave the transient-state.

> ** Toggles
Global toggles are under `SPC t`, `SPC T` and `SPC C-t`
Major mode toggles are only under `SPC m T`

kindaro:
@smile13241324  Thank you.

* I fixed the key bindings to align with the convention.
* But I am not sure what to add to the change log, seeing how the `Coq` layer is itself listed as a new feature. I think that entry subsumes my small addition. Please advise me how to proceed about this.

duianto:
The `Coq` layer doesn't have it's own section in `changelog.develop` yet, so it will have to be added in the section: `Layer changes and fixes`
between `Common Lisp` and `Cscope`.

It could say something like this:
```
**** Coq
- Key bindings:
  - ~SPC m T e~ Toggle electric terminator (thanks to Ignat Insarov)
```
If we add the changelog entry then use the name in the commits `From:` field, in the `thanks to` parentheses, but when a PR author writes the `changelog.develop` entry, then they can chose to use their github name or something else.

kindaro:
@duianto  Thank you. I edited as you suggest.

smile13241324:

PR #13277, opened 2020-02-15
[eshell] Protect prompt only

Stebalien:
If the user enters text while a command is running and the command does not consume the text, eshell will insert this text after it displays the prompt. Unfortunately, that happens after this function was called so this text ends up getting protected along with the prompt text.

This change protects up until the end of the last output (the end of the prompt text), leaving buffered input unprotected.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -3027,6 +3027,8 @@ Other:
   movement in term shells on emacs 26.1
 - Stopped the cursor from jumping to =point-max= when entering insert state from
   a multi line (thanks to Steven Allen)
+- Stopped input buffered while the previous command was running while protecting
+  the prompt (thanks to Steven Allen).
 **** Shell Scripts
 - Added new company-shell environment variable backend
   (thanks to Alexander-Miller)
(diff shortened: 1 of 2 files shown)


Stebalien:
>  Btw: please also add a line to changelog.develop this saves me and the other maintainers precious time we can use for reviewing more contributions.

Will do. I figured only larger changes would go in the changelog.

> However after your change when I enter echo "asdf" before or after the sleep command.

Could you expand on this (exact sequence of events)? And do you have "smart prompt" enabled?

The one thing I _am_ noticing is that, if I hit "enter" on a _previous_ prompt, that text gets input as read-only (but I'm not sure if that's related to this patch).

Stebalien:
Changelog entry added.

smile13241324:
Here is my test procedure:
* Before your patch
   * Start eshell
   * Enter `echo "asdf" return`
   * The shell shows "asdf" and works as expected.

* After your patch with smart eshell enabled
    * Start eshell
    * Enter `echo "asdf" return`
    * The shell shows "asdf" followed by "Text is read only".

* After your patch without smart eshell enabled
    * Start eshell
    * Enter `echo "asdf" return`
    * The shell shows "asdf" and works as expected.

This seems to be caused by smart eshell @Stebalien.

smile13241324:
Given that this PR is already pretty old and we have better alternatives to  `eshell` now a days like `vterm` I think it does not make time to invest more time into this.

PR #13279, opened 2020-02-17
[nlinum] Show deprecation (Emacs 26.1+) warning on startup

duianto:
The warning:
nlinum layer is deprecated for Emacs 26.1 and above
wasn't shown until `SPC h SPC` was pressed.

Because the variable: `configuration-layer--used-layers`
wasn't populated before the warning check occurred in
`nlinum/packages.el`.

The solution (suggested by syl20bnr) was to move the
check to a new file: `nlinum/config.el`.


smile13241324:
Merged :)

PR #13284, opened 2020-02-18
Enhance `sqlfmt` support in SQL layer.

0x6362:
Add sqlfmt-region command; bindings; customize group and new setting for
error buffer handling.

`sqlfmt` was only being applied against the whole buffer, and did not allow
invocation against a specific region in the buffer. `sqlfmt` customize options
were not included under the `sql` group. It was not possible to control if
`sqlfmt` reused an existing error buffer or generated new ones.

- Added `sqlfmt-region` interactive command and binding under `sql-mode` keymap
  - `sqlfmt-buffer` calls into `sqlfmt-region` with `(point-min)` and
    `(point-max`)

- Added `sqlfmt` customize group as child to `sql` customize group

- Added `sqlfmt-reuse-error-buffer` option

- Updated `sql` layer documentation to reflect changes

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -3094,6 +3094,8 @@ Other:
 - Added key binding (thanks to Michael Hauser-Raspe):
   ~SPC m b c~ connect to a SQLi buffer from your saved buffer list
 - Added ~SPC m g~ prefix: =goto= (thanks to duianto)
+- Added regional ~sqlfmt~ invocation, error handling, and formatting bindings
+  (thanks to 0x6362)
 **** Spell-checking
 - Added spell checking transient state (thanks to Francesc Elies Henar)
 - Update to flyspell-correct v0.5 (thanks to Boris Buliga)
--- a/layers/+lang/sql/README.org
+++ b/layers/+lang/sql/README.org
@@ -8,6 +8,7 @@
 - [[#description][Description]]
   - [[#features][Features:]]
 - [[#install][Install]]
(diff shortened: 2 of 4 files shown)


0x6362:
> I have reviewed your PR and found some issues you should have a look at.
The new keybindings conflict with the spacemacs conventions, please adjust them.

I have:
- changed `mf` prefix to `m=`
- removed the top level `m=` binding for `sqlfmt-buffer`

> When I try to call one of the new format functions I am always getting apply: Searching for program: No such file or directory, sqlfmt. Please also have a look at this.

I have updated the layer `README` to reflect the need for an external `sqlfmt` binary and linked to its source. Please let me know if this isn't adequate - I had followed the pattern set by in the layer are using but recognise some others, like `lsp` sometimes pull in dependencies themselves.

I reworded and squashed commits as appropriate to better match other items in history.

0x6362:
@smile13241324 apologies for the churn / force pushes, but this should be good for second review when you have a moment - I've made the requested changes but am happy to try and improve further.

smile13241324:
Hmmm looks good so far. I am just installing the external sql formatter and will give it a short test before merging.

Btw, is it really necessary to build cockroachdb locally to get the formatter installed? On my old Linux box this runs for almost 1h.

0x6362:
No - there are pre-built binaries for linux and darwin available that I was able to use, I think so long as you are on a libc distro and not alpine or something it should be possible to skip the CockroachDB compile, which I suspect is not easy!

smile13241324:

PR #13290, opened 2020-02-20
Add option to hide version string in Spacemacs buffer.

zjp:
This just adds a var to hide the Spacemacs@Emacs(spacemacs) version string at the top right of the dashboard buffer, for those that like minimalism.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -436,6 +436,7 @@ sane way, here is the complete list of changed key bindings
     will obfuscate the current buffer acting like a screen saver.
     (thanks to Eugene Yaremenko and Sylvain Benner)
   - New variable =dotspacemacs-undecorated-at-startup= (thanks to bb2020)
+  - New variable =dotspacemacs-startup-buffer-show-version= (thanks to Zach Pearson)
 - Removed Variables:
   - Removed unused variable =dotspacemacs-verbose-loading= from
     =.spacemacs.template= (thanks to Ying Qu)
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -149,6 +149,10 @@ banner, `random' chooses a random text banner in `core/banners'
 directory. A string value must be a path to a .PNG file.
 If the value is nil then no banner is displayed.")
 
+(defvar dotspacemacs-startup-buffer-show-version t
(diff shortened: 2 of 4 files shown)


duianto:
Maybe it should be documented somewhere, that one should add:
```elisp
      (setq dotspacemacs-startup-buffer-show-version nil)
```
to the `dotspacemacs/user-init` section.

Otherwise the version is shown on startup until the user-config section loads.

And the Spacemacs logo looks a little squished against the top of the Emacs titlebar:

emacs_2020-02-20_15-17-51

It might help to add a newline on the first line when the version isn't shown:
```elisp
          (if dotspacemacs-startup-buffer-show-version
              (spacemacs-buffer//insert-version)
            (let ((inhibit-read-only t))
              (insert "\n")))
```

emacs_2020-02-20_15-26-03

smile13241324:
I would vote for a new setting in `dotspacemacs/layers` then we could take care that it is used at the right location. Having to set it manually in my `user-init` is pretty error prone.

zjp:
@duianto Can you try putting it in dotspacemacs/init instead, with the rest of the dotspacemacs variables? With it there (where I intended it to go), I never see the version string. 

@smile13241324 Can you expand on this? I had originally imagined this as a new setting for the default `.spacemacs`; however, I'm open to changes.

duianto:
@zjp smile probably meant the same place as you, in the `dotspacemacs/init` section.

It seems like it should be added before the variable: `dotspacemacs-startup-banner`

You can check this PR:
Honor custom major-mode in spacemacs/new-empty-buffer
https://github.com/syl20bnr/spacemacs/pull/12383/files

for how to add the entry in the `.spacemacs.template`.

For example write a short comment about what the variable does and include it's default value.

smile13241324:
Jep, @duianto is right that's exactly what I have meant. Sorry for being a bit short on my comment @zjp.

zjp:
I've made the proposed change to spacing -- I agree that it looks better. 

Also, I've added an entry to the `.spacemacs.template`

If anything else is missing let me know! Thanks for the feedback :smile:

smile13241324:
Thank for contributing to spacemacs :purple_heart:, I have cherry picked your commit into develop.

Issue #13270, opened 2020-02-14
The '-key behaves as the d-key instead of invoking pdf-view-jump-to-register


nhmacuk:
#### Description :octocat:
In pdf-view, the command jump-to-register seems to be bound to the  ```'``` key. However, when pressed, it performs the same action as pressing the ```d``` key, namely going down to the bottom of the page.
What I tried so far:
```
;; trying to fix ' to invoke "pdf-view-jump-to-register"
    ;;(with-eval-after-load 'pdf-view
    ;;  (evil-define-key 'evilified pdf-view-map (kbd "'") 'pdf-view-jump-to-register))
    ;(defun mp-add-pdf-view-keys ()
    ;  (local-set-key (kbd "'") 'pdf-view-jump-to-register))
    ;(add-hook 'pdf-view-hook 'mp-add-pdf-view-keys)

  ;; http://blog.lujun9972.win/emacs-document/blog/2019/03/14/evil-guide/
  ;;(evil-make-overriding-map pdf-view-mode-map 'normal)
  ;;(evil-define-key 'normal pdf-view-mode-map "'" 'pdf-view-jump-to-register)

    ;;   (add-hook 'pdf-view-mode-hook
    ;;     (lambda ()
    ;;      (local-set-key (quote [']) (quote pdf-view-jump-to-register))))
    (global-set-key (kbd "s-M-'") 'pdf-view-jump-to-register)
```
The last line works, but the keybinding is not optimal. 

#### Reproduction guide :beetle:
- Start Emacs
- open a pdf-file in pdf-view
- mark a page using ```m```.
- try visit the mark using ```d```.

*Observed behaviour:* :eyes: :broken_heart:
The key ```'``` to invoke the behavior bound to the key ```d``` (an```d``` works as expected/normal).

*Expected behaviour:* :heart: :smile:
The key ```'``` to invoke "pdf-view-jump-to-register".

#### System Info :computer:
- OS: gnu/linux
- Emacs: 26.3
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. a69323fb3)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: helm
- Layers:
```elisp
(ruby gnus
      (plantuml :variables plantuml-jar-path "~/plantuml.jar")
      dap exwm pdf lsp racket html haskell latex helm
[...]

PR #13291, opened 2020-02-20
[pdf] Fix pdf-view-jump-to-register

duianto:
pdf-view-mode being evilified caused:
`'` (single quote) to be bound to: `image-scroll-up`

Fixes: #13270


smile13241324:
Merged :)

Issue #13261, opened 2020-02-11
[React] RJSX mode still using LSP linter even though `javascript-lsp-linter` is disabled


br0p0p:
#### Description :octocat:
The combination of the `react` layer with LSP via the `javascript` layer doesn't work with `flycheck`.

#### Reproduction guide :beetle:
- Start Emacs
- Set the following in `dotspacemacs-configuration-layers`:
    ```
    react
    (javascript :variables
                    javascript-backend 'lsp
                    javascript-lsp-linter nil
                    node-add-modules-path t)
    ```

*Observed behaviour:* :eyes: :broken_heart:
RJSX mode still uses the LSP linter and doesn't use Flycheck. I tried both `(spacemacs/enable-flycheck 'rjsx-mode)` and `(add-hook 'rjsx-mode-hook 'flycheck-mode)` and neither corrected the problem.

*Expected behaviour:* :heart: :smile:
RJSX mode should use the Flycheck linter (ESlint) instead of the LSP linter.

#### System Info :computer:
- OS: darwin
- Emacs: 26.3
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. c9c5637ac)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(auto-completion better-defaults emacs-lisp git helm lsp markdown multiple-cursors org
                 (shell :variables shell-default-height 30 shell-default-position 'bottom)
                 syntax-checking treemacs version-control csv
                 (go :variables go-tab-width 4 go-backend 'lsp go-format-before-save t go-use-golangci-lint t godoc-at-point-function 'godoc-gogetdoc)
                 gtags html
                 (javascript :variables javascript-fmt-on-save t javascript-fmt-tool 'prettier javascript-import-tool 'import-js javascript-backend 'lsp javascript-lsp-linter nil js-indent-level 4 js2-basic-offset 4 js2-include-node-externs t node-add-modules-path t)
                 (json :variables json-fmt-tool 'prettier json-fmt-on-save t)
                 lsp
                 (php :variables php-backend 'lsp)
                 sql yaml react
                 (typescript :variables typescript-backend 'tide typescript-fmt-on-save t))
```
- System configuration features: NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS MODULES THREADS

br0p0p:
[...]

PR #13293, opened 2020-02-21
Replace obsolete lsp-prefer-flymake with lsp-diagnostic-package

duianto:
Fixes: #13261

#### lsp layer
The previous default value of `lsp-prefer-flymake` in Spacemacs was: `nil`
which meant use `flycheck`.

The upstream default value of the new variable `lsp-diagnostic-package` is: `:auto`
which means: 
>Pick flycheck if present and fallback to flymake

Therefore the local definition was removed, so that it's handled upstream.

#### lsp layer readme
Instead of renaming and updating the description for the variable in the lsp/readme variable table:
https://github.com/syl20bnr/spacemacs/tree/develop/layers/%2Btools/lsp#variables
that line was removed, since it only seems to be referring to the lsp layer variables in `config.el`
>A number of configuration variables have been exposed via the LSP layer config.el.

#### scala layer
The `scala` layer set the previous variable `lsp-prefer-flymake` to `nil`, this line was removed, because the new variables default behavior is the same.

#### changelog.develop
There was a previous entry in the changelog.develop from when the now obsolete variable was added. That line was removed. The authors name was included as `originally added by` in this new changelog entry.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1894,7 +1894,7 @@ Other:
     one =go-linter= (thanks to Robert Zaremba)
 - Improvements:
   - Added =LSP= support (thanks to Lupco Kotev)
-  - Set =lsp-prefer-flymake= to =:none= when =Go= layer is using =LSP=
+  - Set =lsp-diagnostic-package= to =:none= when =Go= layer is using =LSP=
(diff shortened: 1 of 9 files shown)


smile13241324:
Merged :)

PR #13298, opened 2020-02-23

emacspace:


doc/DOCUMENTATION.org
```diff
@@ -3407,32 +3407,16 @@ and ~@~ in normal state.
 
 To start recording a keyboard macro to a register type:
 
-#+begin_comment
```
JAremko: should be `#+BEGIN_EXAMPLE` and `#+END_EXAMPLE` since we are trying to keep the number of blog times to the minimum. I think we currently have `SRC`, `EXAMPLE` and `CENTER` :thinking:

duianto: I've pushed an update with your suggestion of wrapping the key bindings with `~` (tilde) characters.

PR #13299, opened 2020-02-24
Add documentation on the difference in regex syntax between Vim and Emacs

x-ji:
When I first started using Emacs I was surprised by some behaviors I encountered during the `/` search in `evil-mode`. They were caused by the differences between Emacs regex and Vim regex. I wanted to point them out to other newcomers so that they would not be confused.


smile13241324:

PR #13300, opened 2020-02-25
Add spacebind macro and tests for it

JAremko:
Testing new macro for binding keys. The pros are:
  - It's pretty much self descriptive.
  - Enforces definition of prefixes for which-key
  - Much more compact and readable than a bunch of function calls
  - [Integrated with use-package](https://github.com/syl20bnr/spacemacs/pull/13300/files#diff-b2ea4bf2528d69e613b7735aba542594R99)
  - Declarative and can be statically analyzed so:
    - We'll be able to catch key binding collisions in CI
    - Pull out data for  generating and keeping in sync key binding tables in README.org files.

### DIFF
--- a/core/core-spacebind.el
+++ b/core/core-spacebind.el
@@ -1,0 +1,275 @@
+;;; core-spacebind.el --- Spacemacs Core File
+;;
+;; Copyright (c) 2012-2020 Sylvain Benner & Contributors
+;;
+;; Author: Eugene "JAremko" Yaremenko <w3techplayground@gmail.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+(require 'core-keybindings)
+
+;; TODO: Make this configurable.
+(define-inline spacemacs/major-mode-prefix ()
+  "Get current prefix for major modes."
+  "m")
+
+(define-inline spacemacs/leader-key ()
+  "Returns `dotspacemacs-leader-key'"
+  dotspacemacs-leader-key)
+
+(defun spacebind//strip-docstring (binding-form)
+  "Remove second element of BINDING-FORM if it is a string."
+  (if (stringp (cadr binding-form))
+      (cons (car binding-form) (cddr binding-form))
+    binding-form))
+
+(defun spacebind//nosp (str)
+  "Remove all white-spaces from STR."
+  (replace-regexp-in-string " +" "" str))
+
+(defun spacemacs/add-key-based-replacements-for-minor-mode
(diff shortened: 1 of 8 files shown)


JAremko:
> It doesn't work to bind a key using a lambda function:
> 
> ```emacs-lisp
>       (spacebind
>        :global
>        (("y" "section y"
>          ("t" '(lambda () (interactive) (message "pressed t")) "t message"))))
> ```

I want to keep those forms code free so they will be passable(by static analyzer) and  easy to read.  At least for now.

> Then the `o` prefix string argument seems to be ignored, because the Spacemacs name `user bindings` remains unchanged.
> 
> It's not limited to this PR, the same thing has been observed before, when defining which-key prefixes, once a prefix has been set, then it doesn't update during the current session. A restart is required to see a new prefix name.

Yeah, `which-key` grabs the first defined prefix. But if you set `(setq which-key-enable-extended-define-key t)` it will probably grab the last one, not sure.
The idea behind `spacebind` macro is that we'll gradually replace  all the bindings we have with it so the static analyzer (that CI will run) will parse those binding forms and report all the collisions.

JAremko:
@duianto  thanks for the help :+1:

duianto:
> I want to keep those forms code free so they will be passable(by static analyzer) and easy to read. At least for now.

👍 

`(setq which-key-enable-extended-define-key t)` doesn't seem to help. But it's a low priority issue.
It's better to get the backend working first before tweaking the frontend stuff.

syl20bnr:
> Should it be supported, or is it better to encourage the use of named functions?

Interactive anonymous functions feel weird to me, if it is interactive then it better need to have a proper name to be `M-x`. I would prefer we don't allow such construction, ideally with a clear message.

@JAremko That's great work and it opens the door to maybe using another back-end in the future to handle the key bindings. Lots of potential regarding documentation as well etc...

I wonder about `which-key` performance once we got everything migrated to spacebind though, maybe we'll need to optimize it. For sure Spacemacs will be a great benchmark for it, we have literally thousands of key bindings.

TBH I prefer macros with explicit keywords even if it means to have a lot keywords but I understand the beauty of the simplicity here, if we can keep it as simple then it can work.

JAremko:
@syl20bnr 

> I wonder about which-key performance once we got everything migrated to spacebind though, maybe we'll need to optimize it. For sure Spacemacs will be a great benchmark for it, we have literally thousands of key bindings.

Worst-case scenario we can hack `which-key-key-replacement-alist` to be buffer-local and fill it with only global + current major/minor mode specific stuff :thinking:

syl20bnr:
Ready to be merged ?

PR #13305, opened 2020-02-26
[feature] tide layer

thanhvg:
Extract `tide-backend` from `typescript` layer to a separate layer named `tide`, so `javascript` and `react` layers can also use `tsserver` provided by this layer as backend sever.

With this commit, there will be three backends for `react` and `javacscript`: `tern`, `lsp` and the new one `tide`. `Typescript` layer still has the same two backends: `lsp` and `tide`.

Motivation: 
We have `lsp` backend which brings `Spacemacs` to the same level as `vsode` and I think `lsp` is the future. However, the performance with `lsp` for JS development can be sometimes sluggish.

Two popular JS lsp servers: `javascript-typescript-stdio` and `typescript-language-server` are missing features or somewhat abandoned.  To my understanding, `vscode` is also cheating when working with JS. It doesn't use any `lsp` server but it talks straight to `tsserver` just like `tide` does. The result is that the performance is better and all the features are available. 

This commit allows `Spacemacs` to have the same access to `tsserver` for JS dev as an alternative choice until the `lsp` project clears all these problems.

### DIFF
--- a/layers/+frameworks/react/packages.el
+++ b/layers/+frameworks/react/packages.el
@@ -22,6 +22,7 @@
     rjsx-mode
     smartparens
     tern
+    tide
     web-beautify
     yasnippet
     ))
@@ -100,6 +101,9 @@
 
 (defun react/post-init-tern ()
   (add-to-list 'tern--key-bindings-modes 'rjsx-mode))
+
+(defun react/post-init-tide ()
+  (add-to-list 'tide-managed-modes 'rjsx-mode))
 
 (defun react/pre-init-web-beautify ()
   (when (eq javascript-fmt-tool 'web-beautify)
--- a/layers/+lang/javascript/packages.el
+++ b/layers/+lang/javascript/packages.el
@@ -31,6 +31,7 @@
         prettier-js
         skewer-mode
         tern
+        tide
         web-beautify
         ))
 
@@ -87,7 +88,7 @@
       (add-hook 'js2-mode-local-vars-hook #'spacemacs//javascript-setup-backend)
       (add-hook 'js2-mode-local-vars-hook #'spacemacs//javascript-setup-next-error-fn)
       ;; safe values for backend to be used in directory file variables
-      (dolist (value '(lsp tern))
+      (dolist (value '(lsp tern tide))
         (add-to-list 'safe-local-variable-values
                      (cons 'javascript-backend value))))
     :config
@@ -260,6 +261,9 @@
 
 (defun javascript/post-init-tern ()
   (add-to-list 'tern--key-bindings-modes 'js2-mode))
+
+(defun javascript/post-init-tide ()
+  (add-to-list 'tide-managed-modes 'js2-mode))
 
 (defun javascript/pre-init-web-beautify ()
   (when (eq javascript-fmt-tool 'web-beautify)
--- a/layers/+lang/typescript/README.org
+++ b/layers/+lang/typescript/README.org
@@ -176,13 +176,14 @@ then set the variable =typescript-lsp-linter= to =nil=.
 | ~SPC m g b~ | jump back                                                    |
 | ~SPC m g g~ | jump to entity's definition                                  |
 | ~SPC m g t~ | jump to entity's type definition                             |
-| ~SPC m g u~ | references                                                   |
+| ~SPC m g r~ | references                                                   |
 | ~SPC m h h~ | documentation at point                                       |
+| ~SPC m p~   | send selected region or current buffer to the web playground |
 | ~SPC m r i~ | organize imports                                             |
 | ~SPC m r r~ | rename symbol                                                |
 | ~SPC m r f~ | rename file                                                  |
-| ~SPC m s p~ | send selected region or current buffer to the web playground |
-| ~SPC m s r~ | restart server                                               |
+| ~SPC m S r~ | restart server                                               |
+| ~SPC m S j~ | create a barebone =jsconfig.json= at project root            |
 
 ** Reference Major Mode
 
--- a/layers/+lang/typescript/packages.el
+++ b/layers/+lang/typescript/packages.el
@@ -17,10 +17,10 @@
         emmet-mode
         flycheck
         smartparens
-        tide
         typescript-mode
         import-js
         web-mode
+        tide
         yasnippet
         ))
 
@@ -84,50 +84,6 @@
     (spacemacs/add-to-hooks #'smartparens-mode '(typescript-mode-hook
                                                  typescript-tsx-mode-hook))))
 
-(defun typescript/init-tide ()
-  (use-package tide
-    :defer t
-    :commands (typescript/jump-to-type-def)
-    :config
-    (progn
-      (spacemacs/declare-prefix-for-mode 'typescript-mode "mE" "errors")
-      (spacemacs/declare-prefix-for-mode 'typescript-tsx-mode "mE" "errors")
-      (spacemacs/declare-prefix-for-mode 'typescript-mode "mg" "goto")
-      (spacemacs/declare-prefix-for-mode 'typescript-tsx-mode "mg" "goto")
-      (spacemacs/declare-prefix-for-mode 'typescript-mode "mh" "help")
-      (spacemacs/declare-prefix-for-mode 'typescript-tsx-mode "mh" "help")
-      (spacemacs/declare-prefix-for-mode 'typescript-mode "mn" "name")
-      (spacemacs/declare-prefix-for-mode 'typescript-tsx-mode "mn" "name")
-      (spacemacs/declare-prefix-for-mode 'typescript-mode "mr" "refactor")
-      (spacemacs/declare-prefix-for-mode 'typescript-tsx-mode "mr" "refactor")
-      (spacemacs/declare-prefix-for-mode 'typescript-mode "mS" "server")
-      (spacemacs/declare-prefix-for-mode 'typescript-tsx-mode "mS" "server")
-      (spacemacs/declare-prefix-for-mode 'typescript-mode "ms" "send")
-      (spacemacs/declare-prefix-for-mode 'typescript-tsx-mode "ms" "send")
-
-      (setq keybindingList '("Ee" tide-fix
-                             "Ed" tide-add-tslint-disable-next-line
-                             "gb" tide-jump-back
-                             "gg" tide-jump-to-definition
-                             "gt" spacemacs/typescript-jump-to-type-def
-                             "gu" tide-references
-                             "hh" tide-documentation-at-point
-                             "ri" tide-organize-imports
-                             "rr" tide-rename-symbol
-                             "rf" tide-rename-file
-                             "sr" tide-restart-server)
-            typescriptList (cons 'typescript-mode keybindingList)
-            typescriptTsxList (cons 'typescript-tsx-mode
-                                    (cons "gg" (cons 'tide-jump-to-definition
-                                                     keybindingList ))))
-      (apply 'spacemacs/set-leader-keys-for-major-mode typescriptList)
-      (apply 'spacemacs/set-leader-keys-for-major-mode typescriptTsxList)))
-
-  (add-to-list 'spacemacs-jump-handlers-typescript-tsx-mode
-               '(tide-jump-to-definition :async t))
-  (add-to-list 'spacemacs-jump-handlers-typescript-mode
-               '(tide-jump-to-definition :async t)))
-
 (defun typescript/post-init-web-mode ()
   (define-derived-mode typescript-tsx-mode web-mode "TypeScript-tsx")
   (add-to-list 'auto-mode-alist '("\\.tsx\\'" . typescript-tsx-mode))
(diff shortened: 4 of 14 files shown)


layers/+lang/typescript/README.org
```diff
@@ -170,24 +170,25 @@ then set the variable =typescript-lsp-linter= to =nil=.
...
-| ~SPC m g g~ | jump to entity's definition                                  |
-| ~SPC m g t~ | jump to entity's type definition                             |
-| ~SPC m g u~ | references                                                   |
-| ~SPC m h h~ | documentation at point                                       |
-| ~SPC m r i~ | organize imports                                             |
-| ~SPC m r r~ | rename symbol                                                |
-| ~SPC m r f~ | rename file                                                  |
-| ~SPC m s p~ | send selected region or current buffer to the web playground |
-| ~SPC m s r~ | restart server                                               |
+| ~SPC m = t~   | reformat the buffer                                          |
```
smile13241324: I think this binding is wrong, it should rather be `SPC m =` independent of who does the formating.

layers/+lang/typescript/README.org
```diff
@@ -170,24 +170,25 @@ then set the variable =typescript-lsp-linter= to =nil=.
...
+| ~SPC m S j~   | create a barebone =jsconfig.json= at project root              |
 
 ** Reference Major Mode
 
 | Key binding | Description             |
 |-------------+-------------------------|
-| ~C-j~       | find previous reference |
-| ~C-k~       | find next reference     |
-| ~C-l~       | goto reference          |
+| ~C-j~         | find previous reference |
```
smile13241324: Here I think the formating is wrong

layers/+lang/typescript/packages.el
```diff
@@ -84,49 +84,49 @@
...
-                                    (cons "gg" (cons 'tide-jump-to-definition
-                                                     keybindingList ))))
-      (apply 'spacemacs/set-leader-keys-for-major-mode typescriptList)
-      (apply 'spacemacs/set-leader-keys-for-major-mode typescriptTsxList)))
-
-  (add-to-list 'spacemacs-jump-handlers-typescript-tsx-mode
-               '(tide-jump-to-definition :async t))
-  (add-to-list 'spacemacs-jump-handlers-typescript-mode
-               '(tide-jump-to-definition :async t)))
+;; (defun typescript/init-tide ()
```
smile13241324: Please remove this code, it is obsolete now. If someone wants to see the history he can still check with git.

layers/+tools/tide/README.org
```diff
@@ -0,0 +1,92 @@
...
+#+begin_src sh
+  npm install -g tsserver
+#+end_src
+
+Next enable this layer in your =~/.spacemacs=. You will need to add =tide= to the
+existing =dotspacemacs-configuration-layers= list in this file. Then set the
+backend variables for JavaScript layer, React layer and TypeScript layer to =tide=
+in =dotspacemacs-configuration-layers=
+
+#+begin_src elsip
```
smile13241324: here is a type it should be elisp but it is elsip.

layers/+tools/tide/README.org
```diff
@@ -0,0 +1,92 @@
...
+Both =javascript-backend= and =typescript-backend= can be set per project.
+
+When using this layer, make sure [[http://www.typescriptlang.org/docs/handbook/tsconfig-json.html][tsconfig.json]] or [[https://code.visualstudio.com/docs/languages/jsconfig][jsconfig.json]] is present in
+the root folder of the project.
+
+Example from emacs tide package: For JavaScript projects you may want to create
+=jsconfig.json= in the root folder of your project. =jsconfig.json= is =tsconfig.json=
+with allowjs attribute set to true.
+
+#+begin_src
```
smile13241324: Here a language is missing.

layers/+tools/tide/README.org
```diff
@@ -0,0 +1,92 @@
...
+    }
+  }
+#+end_src
+
+Without this file =tsserver= will pick up current folder as project root.
+* Key bindings
+** Major Mode
+| Key binding | Description                                     |
+|-------------+-------------------------------------------------|
+| ~SPC m = t~   | format buffer                                   |
```
smile13241324: Same here with the binding, it should be `SPC m =` independent of the module which does the formating.

layers/+tools/tide/packages.el
```diff
@@ -0,0 +1,28 @@
...
+  (use-package tide
+    :defer t
+    :commands (typescript/jump-to-type-def)
+    :config
+    (progn
+      (spacemacs//tide-setup-prefix)
+      (spacemacs//tide-setup-bindings)
+      (spacemacs//tide-setup-jump-handle))))
+
+;;; packages.el ends here
```
smile13241324: No package ends here message in layer files please.

layers/+tools/tide/packages.el
```diff
@@ -0,0 +1,28 @@
...
+;; Copyright (c) 2012-2020 Sylvain Benner & Contributors
+;;
+;; Author: Thanh Vuong <thanhvg@gmail.com>
+;; URL: https://github.com/thanhvg
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+;;; Code:
```
smile13241324: No ;;; Code block in layer files please.

layers/+tools/tide/packages.el
```diff
@@ -0,0 +1,28 @@
...
+(defconst tide-packages
+  '(tide)
+  "The list of Lisp packages required by the tide layer.")
+
+(defun tide/init-tide ()
+  (use-package tide
+    :defer t
+    :commands (typescript/jump-to-type-def)
+    :config
+    (progn
```
smile13241324: This progn is not necessary, please remove.

layers/+tools/tide/funcs.el
```diff
@@ -0,0 +1,99 @@
...
+    "ri" #'tide-organize-imports
+    "rr" #'tide-rename-symbol
+    "rf" #'tide-rename-file
+    "Sr" #'tide-restart-server
+    "Sj" #'spacemacs//tide-create-jsconfig-file))
+
+(defun spacemacs//tide-setup ()
+  "Setup tide backend.
+Must be called by a layer using tide."
+  (progn
```
smile13241324: This progn is not necessary, please remove.

thanhvg:
Hi Max thanks for quick response, 
I changed two key bindings in `typescript` layer which may make some users uncomfortable.
1) `, =` to `= t`: this is to conform with `JavaScript` layer and `lsp` layer. Otherwise format key bindings from `lsp` layer will be overridden.
2) `, s p` to `,  p`, to make space for repl functions https://github.com/emacs-typescript/ts-comint in future PR.

layers/+lang/typescript/packages.el
```diff
@@ -156,16 +112,17 @@
       :config
       (progn
         (when typescript-fmt-on-save
-          (add-hook 'typescript-mode-hook 'spacemacs/typescript-fmt-before-save-hook))
+          (d-hook 'typescript-mode-hook 'spacemacs/typescript-fmt-before-save-hook))
```
smile13241324: This looks like a typo.
```suggestion
          (add-hook 'typescript-mode-hook 'spacemacs/typescript-fmt-before-save-hook))
```

layers/+tools/tide/packages.el
smile13241324: Please remove the line this should not be in a layer.

layers/+tools/tide/packages.el
smile13241324: Please remove this should not be in a layer

layers/+tools/tide/funcs.el
smile13241324: We want to keep the progns in the use-package declaration, however here it is really not necessary and should be removed.

smile13241324:
Hi @thanhvg,
thanks for the background information, I think we should bind the formatting conditionally depending on whether LSP is active or not.
* Bind to `, =` for non lsp
* Bind to `, = =` for lsp bindings

We did something like this already for another layer.
This should prevent any conflicts with existing layers but still don't complicate things for the classic workflow.

To the second binding I would agree immediately as `, s p` is non conventional anyway.

thanhvg:
I have made change in accordance with your request, especially keep the current ` m =`.  Thanks for the fix and constructive feedback.

layers/+tools/tide/funcs.el
```diff
@@ -0,0 +1,85 @@
...
+    (spacemacs/declare-prefix-for-mode mode "mE" "errors")
+    (spacemacs/declare-prefix-for-mode mode "mg" "goto")
+    (spacemacs/declare-prefix-for-mode mode "mh" "help")
+    (spacemacs/declare-prefix-for-mode mode "mn" "name")
+    (spacemacs/declare-prefix-for-mode mode "mr" "refactor")
+    (spacemacs/declare-prefix-for-mode mode "mS" "server")))
+
+(defun spacemacs//tide-setup-bindings ()
+  "Define keys bindings for `tide-mode'"
+  (spacemacs/set-leader-keys-for-minor-mode 'tide-mode
```
smile13241324: Are you sure that binding these is enough for the minor mode only? I would have assumed these would need to be bind for the major modes instead.

layers/+tools/tide/funcs.el
```diff
@@ -0,0 +1,85 @@
...
+(defun spacemacs//tide--list-to-string (list)
+  "Convert LIST to string."
+  (cl-reduce (lambda (x y) (concat x " " (symbol-name y)))
+             (cdr list)
+             :initial-value (format "%s" (car list) )))
+
+(defun spacemacs//tide-setup-company (&rest modes)
+  "Setup tide company for MODES.
+Must be called by a layer using tide."
+  (eval `(spacemacs|add-company-backends
```
smile13241324: The eval here is not necessary, just run:
```elisp
(spacemacs|add-company-backends
           :backends company-tide
           :modes @modes)
```
This also removed the append-hooks nil and call-hooks t flags as they prevent that the default backends are bound which are necessary for i.e. file completion.

layers/+lang/typescript/packages.el
```diff
@@ -159,13 +115,14 @@
...
+          "p" 'spacemacs/typescript-open-region-in-playground)))))
 
 (defun typescript/pre-init-import-js ()
   (if (eq javascript-import-tool 'import-js)
       (progn
         (add-to-list 'spacemacs--import-js-modes (cons 'typescript-mode 'typescript-mode-hook))
         (add-to-list 'spacemacs--import-js-modes (cons 'typescript-tsx-mode 'typescript-tsx-mode-hook)))))
+
+(defun typescript/post-init-tide ()
+  (add-to-list 'tide-managed-modes 'typescript-mode)
```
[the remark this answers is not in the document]
thanhvg: `This tide` layer's implementation is almost the same as `tern-layer`. I don't like the current `lsp` layer's implementation, it's too complicated. I can add a check for `javascript-backend/typescript-backend` in this `post-init` function but I don't want to because that would defeat the ability to switch backend on the fly by setting dir local variables or `M-x: (setq javascrpipt-backend 'lsp/tide/tern)` without the need to change `.spacemacs` and restart. 

For side effects: changes made for the modes (js2-mode, typescript-modde...) this way are only applied if `tide-mode` is loaded. Think of its effects as local buffer wise. So let's say your typescript has tide as backend, and javascript has lsp as backend, that doesn't mean both `tide-mode` and `lsp-mode` are both loaded on js and ts buffers, in fact they are local buffer scope. 

Of course this is my belief combined with trial error testing so feel free to dispute :)

layers/+lang/javascript/packages.el
```diff
@@ -261,6 +262,9 @@
 (defun javascript/post-init-tern ()
   (add-to-list 'tern--key-bindings-modes 'js2-mode))
 
+(defun javascript/post-init-tide ()
+  (add-to-list 'tide-managed-modes 'js2-mode))
```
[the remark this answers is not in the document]
thanhvg: Please see above

layers/+frameworks/react/packages.el
```diff
@@ -101,6 +102,9 @@
 (defun react/post-init-tern ()
   (add-to-list 'tern--key-bindings-modes 'rjsx-mode))
 
+(defun react/post-init-tide ()
+  (add-to-list 'tide-managed-modes 'rjsx-mode))
```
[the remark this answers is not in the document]
thanhvg: same as above

layers/+tools/tide/funcs.el
thanhvg: No, to my knowledge it's impossible that way passing symbol (quoted value) to  macro from function. I tried many times, otherwise you must hardcode the major mode symbol as `lsp`  is doing to enable `company-mode` . `eval` is a must or we need to write another macro to wrap marco `spacemacs|add-company-backends`. Again please check `tern-layer`.

I don't understand the part about file completion, for my js setup with tide, I have 
`C-h v`: `company-backends`

`(company-tide (company-dabbrev-code company-keywords) company-files company-dabbrev)`

layers/+tools/tide/funcs.el
thanhvg: Yeah I think it's ok, this how `lsp` layer adds its key bnidings

smile13241324:
Thanks for the fast answer if you have tested this its all right, I have
just seen that while I was running your code in my head.

Nevermind ignore my previous comment if it still works I will merge it next
time I am on the project again.

smile13241324:
Thank you for contributing to spacemacs, :purple_heart:, I have cherry picked your commit into develop.

PR #13321, opened 2020-02-27
[julia] Remove warning about LanguageServer.jl not working in version 1.0

dmolina:
Remove warning indicating that LanguageServer.jl is not working in Julia version 1.0, because it was updated and that problem version was fixed.


smile13241324:

Issue #1598, opened 2015-05-15
Daily tip on SpaceMacs homescreen


geo7:
In chat with tuhdo and TheBB subject of a daily tip came up. 

I leave gitter open a lot of the time as there is often a nice key command / function mentioned at some point that I can use. It might be nice to have something similar (non intrusively) on the homepage as well. This could (I presume) be linked to any relevant documentation.

tuhdo:
@syl20bnr Prelude [implemented this already](https://github.com/bbatsov/prelude/blob/30fea6d9d6546e2cf055b91335758cdaad9038ac/core/prelude-core.el#L375), so we can use it as an example. Prelude shows the tip [5 seconds after init](https://github.com/bbatsov/prelude/blob/00b48933bd9107273943fa7e278d05c7dc6840b0/init.el#L131).

geo7:
@tuhdo are their tips fixed or can the community add to them?

tuhdo:
Prelude's tips are fixed. But they can add more useful ones with PR.

syl20bnr:
This is an excellent idea. Like prelude we can take the `defvar` approach with a new layer variable called `layer-tips`. Then the engine should display randomly a tip for used layers only.
Tips can fit nicely with the current toggles we have on the homepage.

But for this section I see a necessary dotfile variable: `dotspacemacs-show-tips`. The idea is that the display of the tips is only driven by this variable so we don't have to remember the toggle state of the homepage button.

In all case a release notes that has not been viewed by the user has priority oven the display.

If I sum up the sections
- quickhelp:
  - shown by explicit usage of the toggle button
  - not persistent accros session
- tips:
  - shown by looking at the `dotspacemacs-show-tips` variable
  - can be toggled with the button - not persistent accross session (must use the dotfile variable)
- release notes:
  - driven by the current spacemacs version number and stored value in `.cache` folder
  - can be toggled with the button - persistent toggle state accross session
   - when it should be displayed it takes precedence over any other sections.

syl20bnr:
If we limit the tips to only short sentences then it could be a section independent from quickhelp and release-notes. We could display the tip on top of these action just beneath the button bar.

srid:
@BrianHicks created a tips site: http://spacemacs.brianthicks.com/

I like the mini animated gif demonstration. It would be nice if `<layer>-tips` contained them, to be rendered inside emacs as animated gifs.

BrianHicks:
[...]

PR #13331, opened 2020-02-28
[#1598] [mu4e] Fix purpose labels for mu4e buffers

korayal:
When purpose mode is enabled, Spacemacs fails to assign correct mu4e buffers to
correct windows when we're in headers view and select an e-mail from the list.

Since `*mu4e-headers*` and `*mu4e-view*` buffers have the same purpose
name (`mail`), and `*mu4e-loading*` buffer is either `fundamental` (or with a
recent version of upstream mu4e `mu4e-loading-mode`) which has the purpose
`general`.

So what happens is, when we select an e-mail from the list, e-mail opens up at
the headers' window, and loading opens up in the lower window which should have
displayed the actual email content.

With this commit, there will be two purpose names which will prevent this issue
from happening.


smile13241324:

PR #13332, opened 2020-02-28
Update scheme layer docs to note the need for srfi-18

carlf:
According to the docs for Geiser, chicken 5 requires srfi-18. Without it, I found I could not eval into my REPL. Add docs indicating this requirement.


carlf:
I'm guessing the Emacs25 test fails everywhere as this is just a doc change. Should I update the changelog for a change like this? It doesn't look like changes this small get added to the changelog but I'm happy to add that if needed.

smile13241324:

PR #13336, opened 2020-03-01
Bring ipython-notebook up to date.

dickmao:
For tidiness reasons, removed some bindings like `gj`, `gk`, `M-j`, `M-k`.

### DIFF
--- a/layers/+lang/ipython-notebook/packages.el
+++ b/layers/+lang/ipython-notebook/packages.el
@@ -9,214 +9,64 @@
 ;;
 ;;; License: GPLv3
 
-(setq ipython-notebook-packages
-      '(
-        company
-        ein
-        ob-ipython
-        websocket
-        request
-        request-deferred
-        dash
-        s
-        skewer-mode
-        ))
-
-(defun ipython-notebook/init-websocket ()
-  (use-package websocket
-    :defer t
-    :init))
-
-(defun ipython-notebook/init-request-deferred ()
-  (use-package request-deferred
-    :defer t
-    :init))
-
-(defun ipython-notebook/init-dash ()
-  (use-package dash
-    :defer t
-    :init))
-
-(defun ipython-notebook/init-s ()
-  (use-package s
-    :defer t
-    :init))
-
-(defun ipython-notebook/post-init-skewer-mode ())
-
-(defun ipython-notebook/post-init-request ())
-
-(defun ipython-notebook/post-init-company ())
-
-(defun ipython-notebook/init-ein-notebook ()
-  (use-package ein-notebook
-    :defer t
-    :init))
-
-(defun ipython-notebook/init-ein-subpackages ()
-  (use-package ein-subpackages
-    :defer t
-    :init))
+(setq ipython-notebook-packages '(ein ob-ipython))
 
 (defun ipython-notebook/init-ein ()
   (use-package ein
-    :defer t
-    :commands (ein:notebooklist-open ein:notebooklist-login ein:run ein:stop)
     :init
-    (progn
-      (spacemacs/set-leader-keys
-        "ayl" 'ein:notebooklist-login
-        "ayo" 'ein:notebooklist-open
-        "ayr" 'ein:run
-        "ays" 'ein:stop)
-      (spacemacs/declare-prefix "ay" "ipython notebook")
-      (with-eval-after-load 'ein-notebooklist
-        (evilified-state-evilify-map ein:notebooklist-mode-map
-          :mode ein:notebooklist-mode
-          :bindings
-          (kbd "o") 'spacemacs/ace-buffer-links)
-        (define-key ein:notebooklist-mode-map "o" 'spacemacs/ace-buffer-links)))
+    (spacemacs/set-leader-keys "ayl" 'ein:login
+                               "ayr" 'ein:run
+                               "ays" 'ein:stop)
+    (spacemacs/declare-prefix "ay" "ipython notebook")
     :config
-    (progn
-      (defun spacemacs/ein:worksheet-merge-cell-next ()
-        (interactive)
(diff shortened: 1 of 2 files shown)


smile13241324:

korayal:
@smile13241324 @dickmao are there any reason for the removal of bindings like `gj`, `gk`, `M-j` and `M-k`? I am not the person that added those, but both bindings were pretty useful at `normal` and `insert` modes respectively.

layers/+lang/ipython-notebook/packages.el
```diff
@@ -9,214 +9,62 @@
...
- Operations on Cells^^^^^^            On Worksheets^^^^              Other
- ----------------------------^^^^^^   ------------------------^^^^   ----------------------------------^^^^
- [_k_/_j_]^^     select prev/next     [_h_/_l_]   select prev/next   [_t_]^^         toggle output
- [_K_/_J_]^^     move up/down         [_H_/_L_]   move left/right    [_C-l_/_C-S-l_] clear/clear all output
- [_C-k_/_C-j_]^^ merge above/below    [_1_.._9_]  open [1st..last]   [_C-o_]^^       open console
- [_O_/_o_]^^     insert above/below   [_+_/_-_]   create/delete      [_C-s_/_C-r_]   save/rename notebook
- [_y_/_p_/_d_]   copy/paste           ^^^^                           [_x_]^^         close notebook
- [_u_]^^^^       change type          ^^^^                           [_q_]^^         quit transient-state
- [_RET_]^^^^     execute"
-        :evil-leader-for-mode (ein:notebook-mode . ".")
```
korayal: missed these, which causes users to lose the `,.` bindings for the transient mode and it's respective documentation minibuffer at the bottom.

korayal:
I started getting `list: Symbol’s value as variable is void: ein:%notebook%` error after this update when I do `SPC a o l` (**ein:login**) to load the notebok list.

The only way that I could get to the notebook list was via first `SPC a o r` (**ein:run**) and cancel it, and then I could do `SPC a o l` and it works.

layers/+lang/ipython-notebook/packages.el
dickmao: Indeed I missed `:evil-leader-for-mode`, but intentionally omitted the docstring.
I readded `:evil-leader-for-mode` and noticed the transient menu is unusable without the docstring.
So my conclusion is to eliminate the transient menu altogether since it cannot exist without a manually curated docstring, the manual upkeep of which I am opposed.

#13348

dickmao:
> @smile13241324 @dickmao are there any reason for the removal of bindings like `gj`, `gk`, `M-j` and `M-k`? I am not the person that added those, but both bindings were pretty useful at `normal` and `insert` modes respectively.

Proliferation of keybindings is a personal concern.  When that proliferation also increases the number of slocs to maintain in a module highly susceptible to bitrot, that further incentives removal.

I am not convinced M-j and M-k are useful.  Moving cells is not generally done in notebooks, and there are already bindings for that rare action.

dickmao:
a

> I started getting `list: Symbol’s value as variable is void: ein:%notebook%` error after this update when I do `SPC a o l` (**ein:login**) to load the notebok list.
> 
> The only way that I could get to the notebook list was via first `SPC a o r` (**ein:run**) and cancel it, and then I could do `SPC a o l` and it works.

I could not reproduce.  I also needed to use `a y l` instead of `a o l` so perhaps we're on completely different versions.

korayal:
> I could not reproduce. I also needed to use a y l instead of a o l so perhaps we're on completely different versions.

I mistakenly typed the wrong bindings, we're on the same versions unless you're not using the latest `develop` branch.

korayal:
> I am not convinced M-j and M-k are useful. Moving cells is not generally done in notebooks, and there are already bindings for that rare action.

I can agree that M-j/k is not that much useful. But `gj` & `gk` is extremely useful, since I can move between cells without moving my hand from the default touch-typing position. right now the only alternative you're leaving is `,j` and `,k` and that's not much of a touch-typing friendly alternative compared to `gj` / `gk` (prefixing a binding with g is a commonly used method in evil-normal-mode)

dickmao:
> I mistakenly typed the wrong bindings, we're on the same versions unless you're not using the latest `develop` branch.

Yup, it's an upstream bug.  I couldn't reproduce it earlier because I only typed `SPC a y l` after bootstrapping spacemacs, and the bug only manifests in a post-bootstrapping spacemacs.

Fixed https://github.com/millejoh/emacs-ipython-notebook/commit/160a0aa885524516d7b62f5136a6f469d0e5d936

dickmao:
> But `gj` & `gk` is extremely useful

Okay, 7f7189e reinstates `gj` and `gk`.  I maintain my stance on the docstring, however -- I agree it rarely changes but I would prefer it be dynamically generated than hardcoded.  This would take me at least half a day to do tastefully.

If you reinstate the hardcoded docstring, I would not intercede as you seem to be the biggest (only?) user of EIN in the spacemacs world.  I would merely frown on it, and move on.

korayal:
> If you reinstate the hardcoded docstring, I would not intercede as you seem to be the biggest (only?) user of EIN in the spacemacs world. I would merely frown on it, and move on.

I'm not sure that I'm the only one, but they'll probably visit the repo when their transient-mode/or bindings disappear. I don't plan on making changes on your MR, I'm just stating my preferences as a user of Spacemacs for the last 5 years. I can just go ahead and re-instate those changes locally if I want.

dickmao:
"Transient states" appear to be unrelated to magit's transient library, which I am more familiar with.
I read the top Google hit https://practicalli.github.io/spacemacs/why-spacemacs/transient-state-menus.html , and its UI is too slick for this aging programmer from the last century.

As I say, feel free to reinstate (you can use the `bindings` variable instead of repeating all twenty or so keybindings), but I'll not be the one to do it.

dickmao:
> they'll probably visit the repo when their transient-mode/or bindings disappear.

I put back the transient menu and the docstring 445910b #13348

PR #13338, opened 2020-03-02
[#1598] Fix missing mu4e-modes error

korayal:
The previous PR ( #13331 ) failed to handle everywhere `mu4e-modes` list is being used. So this will bring it back, but still use the added lists for `purpose` setup.


smile13241324:
Thank you for contributing to spacemacs :purple_heart:, I have merged your fix.

PR #13341, opened 2020-03-03
[mu4e] Fix for value of `mu4e-modes` being void

magthe:
It looks like `mu4e-modes` was split into two variables, `mu4e-list-modes` and `mu4e-view-modes` without the use of the variable being adjusted.

This change silences the startup error

```
Error (use-package): helm-mu/:init: Symbol’s value as variable is void: mu4e-modes
```

### DIFF
--- a/layers/+email/mu4e/packages.el
+++ b/layers/+email/mu4e/packages.el
@@ -137,7 +137,7 @@
 (defun mu4e/init-helm-mu ()
   (use-package helm-mu
     :defer t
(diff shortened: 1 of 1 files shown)


jperras:
I'm also seeing this error on startup.

nixmaniack:
Duplicate of https://github.com/syl20bnr/spacemacs/pull/13338?

smile13241324:
Sorry duplicate of #13338

PR #13348, opened 2020-03-05
[+lang/ipython-notebook] Further reduce maintenance

dickmao:
~Remove transient menu that requires expansive docstring.~
Reinstate transient menu with autogen docstring.
Remove unmaintained ob-ipython in favor of maintained ob-ein.
Reinstate `gk` and `gj`.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2111,6 +2111,7 @@ Other:
   - Added ~SPC a y s~ =ein:stop= (thanks to Zach Pearson)
   - Added package dependencies for =ein= (thanks to ft)
   - Fixed keybindings per removal of =ein:notebook-multilang-mode=
+  - DRY and tidiness update by primary EIN developer
 **** Ivy
 - Added =recentf= alt actions to refresh and delete items (thanks to Rich Alesi)
 - Added =projectile= alternate actions to invalidate cache (thanks to Rich Alesi)
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2112,7 +2112,7 @@ Other:
   - Added package dependencies for =ein= (thanks to ft)
   - Fixed keybindings per removal of =ein:notebook-multilang-mode=
   - DRY and tidiness update by primary EIN developer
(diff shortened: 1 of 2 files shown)


korayal:
I think the transient mode is an extremely useful tool to browse/execute through a notebook with a single key press. And that docstring doesn't change much, in fact it may have never changed since it's last updated by @syl20bnr & @justbur  [~4 years ago](https://github.com/syl20bnr/spacemacs/commit/7a0933f4af081ff0543e5f0e521c19d028481f3f)

smile13241324:
I also don't like removing a complete transient state just for maintenance reasons.
This is a core feature of spacemacs, if it breaks due to refactoring the community will fix it.

smile13241324:
@dickmao you mentioned you have replaced ob-ipython with ob-ein, but I see no setup code for it. Doesn't it need some extra actions?

dickmao:
> @dickmao you mentioned you have replaced ob-ipython with ob-ein, but I see no setup code for it. Doesn't it need some extra actions?

`ob-ein` is integrated into `ein`.  There is no separate package.

dickmao:
> I also don't like removing a complete transient state just for maintenance reasons.
> This is a core feature of spacemacs, if it breaks due to refactoring the community will fix it.

445910b I restored the transient state.

smile13241324:
Merged, thank you for contributing to spacemacs :purple_heart:.
@korayal can you double check whether the layer is functioning again for you?

If not please open an issue or a new PR for this.

korayal:
@dickmao @smile13241324 it does work, thanks! 

I see that the documentation is generated automatically. Thanks for the hard work, but I think it misses the actual purpose of the transient mode's doc view which is easy visual accessibility to necessary bindings. And I'm not sure that it is easy to automate, or that it's worth the effort since it's visual curation requires manual intervention.

ie. here is a view for **buffer transient mode**:

image

And this is what I'm seeing as the ein's transient mode:

image

dickmao:
I concede the previous handcrafted docstring was better.  If you wish to restore it, I will not intercede.

PR #13349, opened 2020-03-05
Add new function: spacemacs/kill-back-to-indentation

bb2020:
I could not find any similar function in Spacemacs. Credits and information can be found in the comment.


smile13241324:

PR #13353, opened 2020-03-06
Fix reversed search direction after highlighting

juanedi:
To reproduce the problem:

- Start on a text buffer with the following text
```
aaa
aaa
aaa
aaa
```
- Place the cursor on the second line and enter symbol highlight transient state (`SPC s h`)
- Exit the transient state (by pressing `ESC` for example)
- Press `n`

At this point, I would expect the third line to be highlighted, but the cursor goes to the first one. Repeatedly pressing `n` after that makes it evident that we're stuck searching backwards.

I couldn't trace the purpose of that `nil` being set there, so may be missing something but TBH I don't see any reason for it being there instead of `t`. This change seems to be working fine for me but I'd still appreciate input on how to test for regressions!


smile13241324:

PR #13365, opened 2020-03-11
[spacemacs-language] improve google-translate user expirence

sunlin7:
Hi @smile13241324 
The layer for google-translate using "en" as source language and using "fr" as target language, it's maybe confused someone who first use the translation feature. 
So I change the source language to "auto" and target language to "nil", that will prompt user select the target language.
User can set the target language by press `SPACE x g l`, or `SPACE x g L` to change both source/target language for google-translate.


smile13241324:

sunlin7:
Thank you more.

PR #13366, opened 2020-03-11
Add Verb layer

federicotdn:
Hi!

This PR adds a new layer to use a package that I wrote to organize and send HTTP requests from Emacs (an HTTP client based on Org mode). Here's the link to the repo: https://github.com/federicotdn/verb.

I am not a very experienced Evil user, so if anyone has suggestions on better keybindings, please feel free to comment here.

### DIFF
--- a/layers/+emacs/org/README.org
+++ b/layers/+emacs/org/README.org
@@ -25,6 +25,7 @@
   - [[#sticky-header-support][Sticky header support]]
   - [[#epub-support][Epub support]]
   - [[#jira-support][Jira support]]
+  - [[#verb-support][Verb support]]
 - [[#key-bindings][Key bindings]]
   - [[#starting-org-mode][Starting org-mode]]
   - [[#toggles][Toggles]]
@@ -52,6 +53,7 @@
     - [[#org-mode-bindings][org-mode bindings]]
     - [[#visualization-bindings][Visualization bindings]]
   - [[#org-jira][Org-jira]]
+  - [[#verb][Verb]]
 
 * Description
 This layer enables [[http://orgmode.org/][org mode]] for Spacemacs.
@@ -334,6 +336,15 @@ you connect, add your authentication credentials to =~/.authinfo.gpg= or
 
 #+BEGIN_SRC authinfo
   machine yourcompany.atlassian.net login you@example.com password yourPassword port 443
+#+END_SRC
+
+** Verb support
+To install [[https://github.com/federicotdn/verb][Verb]], an HTTP client based on Org mode, set the
+=org-enable-verb-support= variable to =t=:
+
+#+BEGIN_SRC emacs-lisp
+  (setq-default dotspacemacs-configuration-layers
+   '((org :variables org-enable-verb-support t)))
 #+END_SRC
 
 * Key bindings
@@ -834,3 +845,34 @@ are available.
 | ~SPC a o J s g~ | Get subtasks                                     |
 | ~SPC a o J c u~ | Update the comment at point or add a new comment |
 | ~SPC a o J t j~ | Convert the TODO item at point to a Jira ticket  |
+
+** Verb
+*** verb-mode bindings
+
+| Key Binding | Description                                    |
+|-------------+------------------------------------------------|
+| ~SPC m r r~ | =verb-send-request-on-point-other-window-stay= |
+| ~SPC m r s~ | =verb-send-request-on-point-other-window=      |
+| ~SPC m r m~ | =verb-send-request-on-point-no-window=         |
+| ~SPC m r f~ | =verb-send-request-on-point=                   |
+| ~SPC m r k~ | =verb-kill-all-response-buffers=               |
+| ~SPC m r e~ | =verb-export-request-on-point=                 |
+| ~SPC m r u~ | =verb-export-request-on-point-curl=            |
+| ~SPC m r b~ | =verb-export-request-on-point-verb=            |
+| ~SPC m r n~ | =verb-export-request-on-point-human=           |
+| ~SPC m r v~ | =verb-set-var=                                 |
+
+*** verb-response-body-mode bindings
+
+| Key Binding | Description                            |
+|-------------+----------------------------------------|
+| ~SPC m r r~ | =verb-toggle-show-headers=             |
+| ~SPC m r k~ | =verb-kill-response-buffer-and-window= |
+| ~SPC m r f~ | =verb-re-send-request=                 |
+
+*** verb-response-headers-mode bindings
+
+| Key Binding | Description                   |
+|-------------+-------------------------------|
+| ~SPC m r q~ | =verb-kill-buffer-and-window= |
+
--- a/layers/+emacs/org/README.org
+++ b/layers/+emacs/org/README.org
@@ -849,30 +849,29 @@ are available.
 ** Verb
 *** verb-mode bindings
 
-| Key Binding | Description                                    |
-|-------------+------------------------------------------------|
-| ~SPC m r r~ | =verb-send-request-on-point-other-window-stay= |
-| ~SPC m r s~ | =verb-send-request-on-point-other-window=      |
-| ~SPC m r m~ | =verb-send-request-on-point-no-window=         |
-| ~SPC m r f~ | =verb-send-request-on-point=                   |
-| ~SPC m r k~ | =verb-kill-all-response-buffers=               |
-| ~SPC m r e~ | =verb-export-request-on-point=                 |
-| ~SPC m r u~ | =verb-export-request-on-point-curl=            |
-| ~SPC m r b~ | =verb-export-request-on-point-verb=            |
-| ~SPC m r n~ | =verb-export-request-on-point-human=           |
-| ~SPC m r v~ | =verb-set-var=                                 |
+| Key Binding | Description                                                  |
+|-------------+--------------------------------------------------------------|
+| ~SPC m r r~ | Send request on point in another window, but don't select it |
+| ~SPC m r s~ | Send request on point in another window                      |
+| ~SPC m r m~ | Send request on point, show result status in minibuffer      |
+| ~SPC m r f~ | Send request on point in selected window                     |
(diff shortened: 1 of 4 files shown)


nixmaniack:
This looks nice 👍.

IMO, since the package/layer depends on/extends `org`, it might make sense to provide this as a feature of `org` layer and enable it via a layer variable such as `org-enable-verb-support`.

If the intent is to have this as a separate layer then you should add `layers.el` with following content to define dependency.
```elisp
(configuration-layer/declare-layer-dependencies '(org))
```

layers/+tools/verb/README.org
```diff
@@ -0,0 +1,68 @@
+#+TITLE: Verb layer
+
+file:img/logo.png
+
+# TOC links should be GitHub style anchors.
```
smile13241324: Remove comment this is only for guidance during layer creation

layers/+tools/verb/packages.el
```diff
@@ -0,0 +1,47 @@
+;;; packages.el --- Verb layer packages file for Spacemacs.
+;;
+;; Copyright (c) 2012-2017 Sylvain Benner & Contributors
```
smile13241324: Should be 2012-2020

layers/+tools/verb/packages.el
```diff
@@ -0,0 +1,47 @@
...
+;;
+;; Author: Federico Tedin <federicotedin@gmail.com>
+;; URL: https://github.com/federicotdn/verb
```
smile13241324: The link should be pointing to spacemacs ;; URL: https://github.com/syl20bnr/spacemacs

layers/+tools/verb/packages.el
```diff
@@ -0,0 +1,47 @@
...
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+;;; Commentary:
+
+;; Layer for Verb, an Emacs package to organize and send HTTP
+;; requests based on Org mode.
+
+;;; Code:
```
smile13241324: Please no code blocks in layer files

layers/+tools/verb/packages.el
```diff
@@ -0,0 +1,47 @@
...
+
+(defconst verb-packages '(verb)
+  "List of Lisp packages required by Verb.")
+
+(defun verb/init-verb ()
+  (use-package verb
```
smile13241324: Please defer loading if possible

layers/+tools/verb/packages.el
```diff
@@ -0,0 +1,47 @@
...
+			 (spacemacs/set-leader-keys-for-minor-mode
+			  'verb-response-body-mode
+			  "rr" #'verb-toggle-show-headers
+			  "rk" #'verb-kill-response-buffer-and-window
+			  "rf" #'verb-re-send-request)
+			 (spacemacs/set-leader-keys-for-minor-mode
+			  'verb-response-headers-mode
+			  "rq" #'verb-kill-buffer-and-window))))
+
+;;; packages.el ends here
```
smile13241324: No package.el ends here in a layer please

layers/+tools/verb/README.org
```diff
@@ -0,0 +1,68 @@
+#+TITLE: Verb layer
+
```
smile13241324: Add proper tags declaration, in your case it should be I assume #+TAGS: layer|tool

federicotdn:
Hey @nixmaniack and @smile13241324, thanks for the feedback. Just to clarify, does this mean my package should live entirely insde `layers/+emacs/org/packages.el` and `layers/+emacs/org/config.el`, or does it need its own `verb` directory with README and the other files? I can see that there are other packages that depend on Org, such as `org-trello` or `org-jira` that don't have their own layer directories.

nixmaniack:
> does this mean my package should live entirely insde `layers/+emacs/org/packages.el` and `layers/+emacs/org/config.el`

Yes. 

> does it need its own `verb` directory with README and the other files?

No. It should be similar to those other extensions that you mentioned (`org-trello`, `org-jira`). The readme content should be just another heading in org layer readme.

smile13241324:
@federicotdn just for the record your package should live in the org layer folder and should not have its own layer files.

federicotdn:
@smile13241324 no problem, thanks for the feedback. I will get back to this in the next days 👍🏻

federicotdn:
I've updated the PR! I wasn't able to find a way to defer the package from loading. Normally (without `use-package`) it would get autoloaded when the user uses any command from the `verb-command-map`.

layers/+emacs/org/README.org
```diff
@@ -834,3 +845,34 @@ are available.
...
+| ~SPC m r s~ | =verb-send-request-on-point-other-window=      |
+| ~SPC m r m~ | =verb-send-request-on-point-no-window=         |
+| ~SPC m r f~ | =verb-send-request-on-point=                   |
+| ~SPC m r k~ | =verb-kill-all-response-buffers=               |
+| ~SPC m r e~ | =verb-export-request-on-point=                 |
+| ~SPC m r u~ | =verb-export-request-on-point-curl=            |
+| ~SPC m r b~ | =verb-export-request-on-point-verb=            |
+| ~SPC m r n~ | =verb-export-request-on-point-human=           |
+| ~SPC m r v~ | =verb-set-var=                                 |
+
```
nixmaniack: In general, description would be preferred instead of just function names.

layers/+emacs/org/packages.el
```diff
@@ -810,3 +811,29 @@ Headline^^            Visit entry^^               Filter^^                    Da
     :defer t
     :init
     (add-hook 'org-mode-hook 'org-sticky-header-mode)))
+
+(defun org/init-verb ()
+  (use-package verb
+               :after org
+               :config
```
nixmaniack: Use `:defer t` to defer the loading.
Keybindings should generally go inside `:init`. Registering keybindings adds entrypoints to the package but doesn't load the package. 
I'm not sure if `:after` is required but keep it if removing it is causing an issue.

federicotdn:
@nixmaniack Thanks for the feedback! Changing `config` to `init` did the trick.

nixmaniack:
@federicotdn Looks good to me 👍. 

I looked at your verb package and it also provides `org-babel` integration. It'd be good if you can add support for that feature as well. 

You can take a look at [other configurations](https://github.com/syl20bnr/spacemacs/blob/bb8ec23ec1794fb94f1fb4c4459af20fa3620c9c/layers/+lang/clojure/packages.el#L423-L426) for adding org-babel support.
e.g.
```elisp
(spacemacs|use-package-add-hook org
    :post-config (add-to-list 'org-babel-load-languages '(verb . t)))
```

federicotdn:
Good idea! I have tried adding the following at the bottom of `+emacs/org/packages.el`:
```elisp
(defun org/pre-init-verb ()
  (spacemacs|use-package-add-hook org
    :post-config (add-to-list 'org-babel-load-languages '(verb . t))))
```

but the problem is that now, my package is getting eagerly loaded. Any idea of how to fix this? I could put that instead in the `:config` section of the `use-package` call, but that would mean that the package would need to be loaded (by using one of the keybindings) before any Babel block could be executed (I believe).

nixmaniack:
> I could put that instead in the `:config` section of the `use-package` call, but that would mean that the package would need to be loaded (by using one of the keybindings) before any Babel block could be executed (I believe).

Will having autoload to your `org-babel-execute:verb` would help? Might be going against convention of how we add the org-babel support but see if that fixes. 

There's more [documentation](https://develop.spacemacs.org/doc/LAYERS.html) on layer and how the loading works which might help. I can take a look later in the week if above suggestion doesn't work or maybe maintainers can suggest.

federicotdn:
I've done a little research because I was curious about if adding `autoload` would work or not. The first thing I noticed is that none of the `org-babel-execute:` functions included in Emacs are autoloaded.

I also noticed that `org-babel-do-load-languages` [immediately requires](https://github.com/bzg/org-mode/blob/master/lisp/org.el#L206) the packages listed inside the `org-babel-load-languages` list. For this reason I believe autoload won't help us in this case.

In the `packages.el` file for the Org layer, in the `use-package ob` section, `org-babel-do-load-languages` is being set up to be called as soon as an Org file is opened. For that reason, Verb is being loaded right after an Org file is opened (when I said "eagerly loaded" in my previous comment, I was referring to this).

The ideal thing would be for Verb to be loaded only when the Babel block is executed, but I can't see how `org-babel-do-load-languages` will allow us to do that.

nixmaniack:
> For that reason, Verb is being loaded right after an Org file is opened (when I said "eagerly loaded" in my previous comment, I was referring to this).

Thanks for doing that research. I understand it now.

> The ideal thing would be for Verb to be loaded only when the Babel block is executed, but I can't see how `org-babel-do-load-languages` will allow us to do that.

Hmm. Given how org-babel is doing its setup — I'm not sure if we have an option of deferred loading the way you expected. And, I think that's a deliberate choice that's made at upstream, but it does allow a user to explicitly enable/disable languages of choice to be available when writing org-document using the pair `(<lang> . <state>)`. I would have definitely preferred even that part to be deferrable but we can skip that for now, IMO.

federicotdn:
No problem. I've pushed a new commit, but for some reason CircleCI can't find the base branch for my PR 🤔

layers/+emacs/org/packages.el
```diff
@@ -45,6 +45,7 @@
         (ox-jira :toggle org-enable-jira-support)
         (org-trello :toggle org-enable-trello-support)
         (org-sticky-header :toggle org-enable-sticky-header)
+	(verb :toggle org-enable-verb-support)
```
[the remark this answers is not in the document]
smile13241324: Is fixed in the latest version

layers/+emacs/org/packages.el
smile13241324: package is now properly defered.

smile13241324:

federicotdn:
Awesome!! Thanks @smile13241324 and @nixmaniack.

PR #13376, opened 2020-03-14
[themes][keybindings] Use apt completion framework for loading theme

nixmaniack:
Theme transient state key-binding `t` is broken if `ivy` completion framework is
used. The binding unconditionally invokes `helm-themes`.

This change adds a new function which takes care of invoking correct
theme-loader function based on completion framework currently in use.

The old binding used `helm-themes` but the new function uses already defined
`spacemacs/helm-themes` which removes the number of candidates limit.


duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/bb8ec23ec1794fb94f1fb4c4459af20fa3620c9c

PR #13377, opened 2020-03-14
replace custom ivy hydra-menu by original hydra-menu

dalanicolai:
Replaced the custom ivy hydra menu with the original (and updated) ivy hydra menu. Additionally checked parts of the documentation that required modification and applied them accordingly (no modifications were needed, only some additions)

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2146,6 +2146,8 @@ Other:
 - Limit =ripgrep= results to 150 columns (thanks to Aaron Jensen)
 - Use new API to register additional transient state bindings
   (thanks to Andriy Kmit')
+- Replaced custom hydra-menu by original and updated hydra-menu, introducing
+  marking feature (thanks to Daniel Nicolai)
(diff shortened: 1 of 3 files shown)


nixmaniack:
👍 For this change, my only gripe is the missing `s-M-SPC`–an alternate for `M-SPC`–according the [convention](https://github.com/syl20bnr/spacemacs/blob/cb48ec74c1f401bd2945760799633c0e81e69088/doc/CONVENTIONS.org#transient-state). We should, ideally, add them.

smile13241324:

@dalanicolai next time please take care that your PR has only one commit. To do so either force push on your PR branch or make single commits and squash them in the end.

dalanicolai:
Yeah I am learning git... and sometimes some things happen that I do not understand yet... anyway I will check things again and next time think about the squash

j-martin:
I suspect this PR broke the completion if not using the `ivy` layer. https://github.com/syl20bnr/spacemacs/issues/13395

PR #13383, opened 2020-03-16
[Helm] Pick up the symbol at point when do smart search with "rg"

sunlin7:
The `spacemacs/helm-project-smart-do-search' will pick up the symbol at point
when searching with "grep" backend, but pick up nothing when searching with "rg"
backend, here we make the "rg" backend has similar behavior as default backend.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2022,8 +2022,8 @@ Other:
   - Added guard to check if =winum= is loaded (thanks to Dan Girshovich)
   - Updated =helm-xref= to set the correct =xref-show-xrefs-function= for
     Emacs 27 (thanks to Junxuan)
-  - Make "rg" backend pick up symbol at point for 
-    =spacemacs/helm-project-smart-do-search= (thanks to Lin Sun)
+  - Made the spacemacs/helm-files-do-rg pickup sympol at point to follow 
+    the behavior of spacemacs/helm-files-do-grep (thanks to Lin Sun)
 - Key bindings:
   - Added Key bindings for directory search:
     - ~SPC s d for =spacemacs/helm-dir-smart-do-search=
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2022,6 +2022,8 @@ Other:
   - Added guard to check if =winum= is loaded (thanks to Dan Girshovich)
   - Updated =helm-xref= to set the correct =xref-show-xrefs-function= for
     Emacs 27 (thanks to Junxuan)
+  - Make "rg" backend pick up symbol at point for 
+    =spacemacs/helm-project-smart-do-search= (thanks to Lin Sun)
 - Key bindings:
   - Added Key bindings for directory search:
     - ~SPC s d for =spacemacs/helm-dir-smart-do-search=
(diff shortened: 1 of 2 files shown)


sunlin7:
@smile13241324 Could you please review this change? Thank you.

smile13241324:
I am not sure whether prompting the symbol at point to your smart search is not a controversial change. Let me discuss this with the other maintainers first.

sunlin7:
Okay, look forward your good news. The change is try to make the  "rg" backend behavior to follow the default backend (grep/ack).

nixmaniack:
I'm not sure why this change is required.

If you check [here](https://github.com/syl20bnr/spacemacs/blob/2049817b1281dbda1286771a99498aa6ea4696bc/doc/DOCUMENTATION.org#searching-in-a-project), it'll give you a bit more idea.

In brief, `SPC /` or `SPC s p` which is bound to  `spacemacs/helm-project-smart-do-search` allows you to `search project` with an input prompt.
And, `SPC *` or `SPC s P` which is bound to `spacemacs/helm-project-smart-do-search-region-or-symbol` allows you to search project `with symbol/word under point`.

This all pieces together with auto tool selection mechanism based on the availability of the `search tool` (rg, ag etc. defined with `dotspacemacs-search-tools`) on the system. So, changing the function will essentially break `SPC /` and functions dependent on it. 

Another disclaimer, I use `ivy` instead of `helm` but the underlying mechanism remains same as far as I understand. So, unless `SPC *` related functionality is broken in `helm`, this change is not required. 

TL;DR: What the OP wants is already available with `SPC *` or `SPC s P` or if you need specific tool based search e.g. `rg` use `SPC r P`.

sunlin7:
@nixmaniack Thank you for your explanation. This change aim to make the "rg" backend follow the  beharior of default "grep" backend for completion with heml.
With the "helm" layer enabled, the function `spacemacs/helm-project-smart-do-search'  will pick up the symbol at point with default backend while it pick up nothing when there is "rg" in the path, it's confused me quite a while for same spacemacs configurationes on two machines, one has "rg" but another didn't.
And the change acturly for helm, it made the function `spacemacs/helm-files-do-rg' act like `spacemacs/helm-files-do-grep', didn't affect ivy.

nixmaniack:
As explained in the previous comment,
`SPC /` **shouldn't** fill in the symbol at point
`SPC *` **should** fill in the symbol at point 
the search tool doesn't matter, the functionality/convention should be adhered to.

If having `grep` as search tool causes the helm prompt to fill in the symbol at point for `SPC /`, then that's a bug with "helm smart search with grep" and we should be fix that. 

Could you tell us which keybinding you're invoking and what is the expected behaviour?

sunlin7:
@nixmaniack Okay. Close this PR as innocent. Thank you.
It's `M-m /`, `spacemacs/helm-project-smart-do-search`.

nixmaniack:
You may want to create an issue for the incorrect behaviour of helm grep.

smile13241324:
I am not sure why helm-grep adds symbol under point to the search input. I have double checked the code and it clearly gives "nil" to helm-grep-1. Maybe there is an issue with helm-grep? I will need to investigate a bit further for this.

PR #13389, opened 2020-03-18
fixed emacs major mode leader key binding

### LINKED ISSUE #13374: M-RET no longer org mode prefix in GUI
#### Description :octocat:
M-RET no longer org mode prefix in GUI, it will add headline.

`SPC h d k` shows `<M-return> runs the command org-meta-return (found in org-mode-map), which is an
interactive compiled Lisp function in ‘org.el’.
`

 I tried both emacs-mac and emacs-plus, neither of them works. But it works as expected in terminal.

- Start Emacs
- create org file, press M-RET

- Insert new headline

- Show mode commands

#### System Info :computer:
- OS: darwin
- Emacs: 26.3
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 710aa947b)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
(html markdown javascript
      (python :variables python-backend 'anaconda)
      ivy
      (auto-completion :variables auto-completion-enable-snippets-in-popup nil auto-completion-enable-help-tooltip t auto-completion-enable-sort-by-usage t)
      better-defaults git treemacs
      (org :variables org-enable-hugo-support t org-startup-indented t org-enable-sticky-header t)
      osx
      (shell :variables shell-default-shell 'vterm shell-default-term-shell "/bin/zsh" shell-default-height 30 shell-default-position 'bottom)
      spell-checking syntax-checking)
```
- System configuration features: NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS MODULES THREADS

#### Backtrace :paw_prints:
```
<<BACKTRACE IF RELEVANT>>
```

emacs18:
Allow M-RET key to work as major mode emacs leader key for both GUI and terminal modes.
This should fix #13374.

The current value of "C-M-m" as leader key works only in terminal mode, but not in GUI mode.
That is why we set this to "<M-return>" for GUI mode, and to "C-M-m" for terminal mode.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -475,6 +475,9 @@ Other:
   - Removed obsolete =:powerline-scale= property from
     =dotspacemacs-default-font= variable
     (thanks to Muneeb Shaikh and Sylvain Benner)
+  - Modified =dotspacemacs-major-mode-leader-key= so that =M-RET= key works in
+    org-mode as well as in terminal and GUI modes. =C-M-m= still works in
+    terminal mode, but no longer in GUI mode. (emacs18)
 *** Core changes
 - Improvements:
   - Display time spent in =user-config= in home buffer (thanks to Sylvain Benner)
--- a/core/templates/.spacemacs.template
+++ b/core/templates/.spacemacs.template
@@ -234,8 +234,10 @@ It should only modify the values of Spacemacs settings."
    dotspacemacs-major-mode-leader-key ","
 
    ;; Major mode leader key accessible in `emacs state' and `insert state'.
-   ;; (default "C-M-m")
-   dotspacemacs-major-mode-emacs-leader-key "C-M-m"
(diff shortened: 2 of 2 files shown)


smile13241324:

A short remark when you change/add a new variable to the template you also need to adapt it in `core-dotspacemacs.el`. See 6a745993e360a9e45146d8dd9e1661b7e1a9e914 for details.

aguadopd:
@emacs18 This is fixing the problem for evil mode but not for holy mode #9603 , right?

smile13241324:
@aguadopd quite the contrary actually the only changed variable is the one used by Holy and Hybrid mode, evil mode never had that issue as the major mode leader key there is ",".

aguadopd:
Thanks @smile13241324. For some unknown reason it was not working for me then, but just started from a clean spacemacs.d file and it works as expected.

PR #13390, opened 2020-03-19
Fix spacemacs/open-junk-file

mjkramer:
I've encountered three "problems" with `spacemacs/open-junk-file`:

1. If the junk directory doesn't exist, it will *not* be automatically created. Instead, you'll get prompted to create it when you save the junk file. It's not a huge inconvenience, just slightly annoying, but it runs contrary to the behavior of `open-junk-file`, which *does* create the directory. This PR makes `spacemacs/open-junk-file` behave consistently to `open-junk-file`.

2. If you're using Ivy, and you call `spacemacs/open-junk-file` from a Dired buffer, you'll be prompted to create a file within the Dired directory, instead of in the junk directory. This PR uses a one-line hack to force Counsel to use the correct directory.

3. If the junk file is an Org file, and your junk folder is the default (i.e. `.cache/junk`), `spacemacs/prettify-org-buffer` will be called (activating `space-doc-mode` etc.), since the path lies within `spacemacs-start-directory`. Personally I never want space-doc-mode to be active in my personal Org files. This PR adds the variable `dotspacemacs--pretty-ignore-subdirs` with the default value `'(".cache/junk")`, and modifies `dotspacemacs//prettify-spacemacs-docs` accordingly.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -636,6 +636,8 @@ Other:
   - Ignore nils in dotspacemacs-configuration-layers (thanks to Ag)
   - Fixed redundant package version checking during update
     (thanks to aaronjensen)
+  - Fixed unexpected behavior in =spacemacs/open-junk-file= (thanks to Matt
+    Kramer)
 - Other:
   - New function =configuration-layer/message= to display message in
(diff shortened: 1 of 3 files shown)


smile13241324:

PR #13392, opened 2020-03-20
Fix mode name, and remove unmaintainable code

lcos77:
The **ipython notebook** layer in its current state doesn't load its minor mode keybindings in my spacemacs. And I imagine this might be the case for everyone else, because if I'm not mistaken, the ipython notebook pacakge on which this layer depends, has been changing a lot recently. For this reason, I am proposing the following changes:

-  Previously, the code had references to _ein:notebook-mode_ which doesn't seem to work (anymore), so I changed these references to _ein:notebook_. 

- The previous code had a convoluted function that produced the doc-string and that was tedious to read that depended on some other smaller functions, though not enough of them to make the code sufficiently modular, and therefore, it wasn't very maintainable. Besides, when I tried the previous version with the name of the mode fixed and the convoluted function, spacemacs hung for a long time -- it's a mystery whether spacemacs would have continued normally if I had waited long enough. Therefore I replaced the doc-string with a hard coded one. 

- I added some keybindings from  _tkamat_'s fork to the code for I didn't see a good reason not to.

- I removed the cl- prefix from "second" and "mapcan", because cl is a deprecated package. However, I'm afraid the names without the prefix might end up being unrecognized by a previous version of emacs.

### DIFF
--- a/layers/+lang/ipython-notebook/packages.el
+++ b/layers/+lang/ipython-notebook/packages.el
@@ -75,14 +75,14 @@
                         ("fs" ein:notebook-save-notebook-command))))
         (apply #'spacemacs/set-leader-keys-for-minor-mode
                (quote ein:notebook)
-               (mapcan
+               (cl-mapcan
                 (lambda (bind)
-                  (if (fboundp (second bind))
+                  (if (fboundp (cl-second bind))
                       bind
                     (prog1 nil
                       (display-warning
                        'warn (format "ipython-notebook/init-ein: undefined %s"
-                                     (second bind))))))
+                                     (cl-second bind))))))
                 (copy-tree bindings)))
         (eval (append '(spacemacs|define-transient-state
 			                   ipython-notebook
--- a/layers/+lang/ipython-notebook/packages.el
+++ b/layers/+lang/ipython-notebook/packages.el
@@ -15,166 +15,89 @@
   (use-package ein
     :init
     (spacemacs/set-leader-keys "ayl" 'ein:login
-                               "ayr" 'ein:run
-                               "ays" 'ein:stop)
+      "ayr" 'ein:run
+      "ays" 'ein:stop)
     (spacemacs/declare-prefix "ay" "ipython notebook")
     :config
-    (require 'seq)
-    (require 'subr-x)
     (mapc
      (lambda (mode)
        (evil-define-minor-mode-key
-        mode 'ein:notebook-mode
-        (kbd "<C-return>") 'ein:worksheet-execute-cell-km
-        (kbd "<S-return>") 'ein:worksheet-execute-cell-and-goto-next-km))
+         mode 'ein:notebook
+         (kbd "<C-return>") 'ein:worksheet-execute-cell-km
+         (kbd "<S-return>") 'ein:worksheet-execute-cell-and-goto-next-km))
      '(insert hybrid normal))
     (evil-define-minor-mode-key
-     'normal 'ein:notebook-mode
-     "gj" 'ein:worksheet-goto-next-input-km
-     "gk" 'ein:worksheet-goto-prev-input-km)
+      'normal 'ein:notebook
+      "gj" 'ein:worksheet-goto-next-input-km
+      "gk" 'ein:worksheet-goto-prev-input-km)
     (with-eval-after-load 'ein-notebook
-      (evil-define-key nil ein:notebooklist-mode-map "o" 'spacemacs/ace-buffer-links)
-      (let ((bindings '(("j" ein:worksheet-goto-next-input-km)
-                        ("k" ein:worksheet-goto-prev-input-km)
-                        ("J" ein:worksheet-move-cell-down-km)
-                        ("K" ein:worksheet-move-cell-up-km)
-                        ("e" ein:worksheet-toggle-output-km)
-                        ("d" ein:worksheet-kill-cell-km)
-                        ("y" ein:worksheet-copy-cell-km)
-                        ("p" ein:worksheet-yank-cell-km)
-                        ("m" ein:worksheet-merge-cell-km)
-                        ("s" ein:worksheet-split-cell-at-point-km)
-                        ("o" ein:worksheet-insert-cell-below-km)
-                        ("O" ein:worksheet-insert-cell-above-km)
-                        ("t" ein:worksheet-toggle-cell-type-km)
-                        ("RET" ein:worksheet-execute-cell-km)
-                        ("l" ein:worksheet-clear-output-km)
-                        ("L" ein:worksheet-clear-all-output-km)
-                        ("C-s" ein:notebook-save-notebook-command-km)
-                        ("C-r" ein:notebook-rename-command-km)
-                        ("x" ein:notebook-close-km)
-                        ("z" ein:notebook-kernel-interrupt-command-km))))
+      (let ((bindings '(("y" ein:worksheet-copy-cell)
+                        ("p" ein:worksheet-yank-cell)
+                        ("d" ein:worksheet-kill-cell)
+                        ("h" ein:notebook-worksheet-open-prev-or-last)
+                        ("i" ein:worksheet-insert-cell-below)
+                        ("I" ein:worksheet-insert-cell-above)
+                        ("j" ein:worksheet-goto-next-input)
+                        ("k" ein:worksheet-goto-prev-input)
+                        ("l" ein:notebook-worksheet-open-next-or-first)
+                        ("H" ein:notebook-worksheet-move-prev)
+                        ("J" ein:worksheet-move-cell-down)
+                        ("K" ein:worksheet-move-cell-up)
+                        ("L" ein:notebook-worksheet-move-next)
+                        ("t" ein:worksheet-toggle-output)
+                        ("R" ein:worksheet-rename-sheet)
+                        ("RET" ein:worksheet-execute-cell-and-goto-next)
+                        ;; Output
+                        ("C-l" ein:worksheet-clear-output)
+                        ("C-S-l" ein:worksheet-clear-all-output)
+                        ;;Console
+                        ;; ("C-o" ein:console-open)
+                        ;; Merge cells
+                        ("C-k" ein:worksheet-merge-cell)
+                        ;;("C-j" spacemacs/ein:worksheet-merge-cell-next)
+                        ("s" ein:worksheet-split-cell-at-point)
+                        ;; Notebook
+                        ("C-s" ein:notebook-save-notebook-command)
+                        ("C-r" ein:notebook-rename-command)
+                        ("1" ein:notebook-worksheet-open-1th)
+                        ("2" ein:notebook-worksheet-open-2th)
+                        ("3" ein:notebook-worksheet-open-3th)
+                        ("4" ein:notebook-worksheet-open-4th)
+                        ("5" ein:notebook-worksheet-open-5th)
+                        ("6" ein:notebook-worksheet-open-6th)
+                        ("7" ein:notebook-worksheet-open-7th)
+                        ("8" ein:notebook-worksheet-open-8th)
+                        ("9" ein:notebook-worksheet-open-last)
+                        ("+" ein:notebook-worksheet-insert-next)
+                        ("-" ein:notebook-worksheet-delete)
+                        ("x" ein:notebook-close)
+                        ("u" ein:worksheet-change-cell-type)
+                        ("fs" ein:notebook-save-notebook-command))))
         (apply #'spacemacs/set-leader-keys-for-minor-mode
-               (quote ein:notebook-mode)
-               (cl-mapcan
+               (quote ein:notebook)
+               (mapcan
                 (lambda (bind)
-                  (if (fboundp (cl-second bind))
+                  (if (fboundp (second bind))
                       bind
                     (prog1 nil
                       (display-warning
                        'warn (format "ipython-notebook/init-ein: undefined %s"
-                                     (cl-second bind))))))
+                                     (second bind))))))
                 (copy-tree bindings)))
-	(eval (append '(spacemacs|define-transient-state
-			ipython-notebook
-			:title "iPython Notebook Transient State"
-			:evil-leader-for-mode (ein:notebook-mode . ".")
-			:bindings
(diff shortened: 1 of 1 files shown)


mjkramer:
Nice. But if I understand correctly, `cl-lib` is not being deprecated, `cl` is. The functions with the `cl-` prefix belong to `cl-lib`, while the unprefixed functions belong to `cl`. So `cl-mapcan` and `cl-second` should remain as-is.

lcos77:
I googled it, and you're right, I had them backwards, thanks for pointing it out!

smile13241324:

dickmao:
Hi, thanks for coming to ipython-notebook layer's rescue.  It perturbs me that `evil-define-minor-mode-key` on `ein:notebook-mode` causes `evil-surround-mode-map` to usurp `ein:notebook-mode-map` in the `minor-mode-map-alist` order of precedence (thus causing the buggy behavior that this PR fixes).

Changing `ein:notebook-mode` to `ein:notebook` nicely preserves `ein:notebook-mode-map` at the head of `minor-mode-map-alist` but has the side effect of creating a brand new keymap called `ein:notebook`, which I must say is a little untidy and possibly unintentional.

It would be nice if someone well-versed in evil-surround or evil-define-key could augment the `ein:notebook-mode-map` without inadvertently creating a whole other keymap.

lcos77:
Yes, it does add one element to the minor-mode-map-alist, and ein-notebook.el. It looks like the ein:notebook is an obsolete alias though it still works because of the ein:notebook--worksheet-render function. I tried renaming the instances of ein:notebook to ein:notebook-mode and it also works with the added benefit of not having the extra element in the minor-mode-map-alist. However, I'm not sure why it didn't hung as a result of the ipython-notebook/transient-doc as it did when the keybindings were named for the ein:notebook, yet, of course, the key bindings weren't bound (Besides, as enabled minor mode it's called just Ein:Notebook, though I imagine it's mostly because it was parsed that way). If anything it seems that perhaps the ipython-notebook/transient-doc was the problem since after all, the only update to the ein-notebook after the fix, was made the 28th of March, and it wasn't related to ein:notebook--worksheet-render function. It was perhaps the only problem, even for the ein:notebook-mode case (maybe it hung in the background heh).  Can you people confirm this?

dickmao:
l> It looks like the ein:notebook is an obsolete alias though
l> it still works because of the ein:notebook--worksheet-render function.

I fail to see how ein:notebook is anything but a random string.  Changing the
argument of `evil-define-minor-mode-key` to "foobar" I suspect would have worked.

l> tried renaming the instances of ein:notebook to ein:notebook-mode and it
l> also works with the added benefit of not having the extra element in the
l> minor-mode-map-alist.

You mean leaving it as it was?  That did not work for me insofar as
evil-surround-map ended up frontrunning ein:notebook-mode-map in the
minor-mode-map-alist, causing a binding like `C-c C-c` to reference
python-mode instead of ein:notebook-mode.  `C-RET` however did work because
presumably python-mode doesn't fight for it.

l> However, I'm not sure why it didn't hung as a result
l> of the ipython-notebook/transient-doc as it did when the keybindings were

Using the spacemacs commit right before yours (d7868da^), I could not get it
to hang.  But it would seem the consensus is to kill 
ipython-notebook/transient-doc, so figuring out the hanging is moot.

lcos77:
I think I see what you mean, and it seems like we do share the same problem. I was only paying attention to the evil keybindings that do work for all major modes in an ipython notebook worksheet, but you're right that with ein:notebook-mode the default keybindings are overshadowed by the python major mode (On the other hand, they don't even appear in the fundamental mode, yet they work perfectly in the markdown mode.)

dickmao:
> What I meant was that d7868da hung only with the mode name changed to
> "ein:notebook"

Ah, well, I find it rather unfortunate that my dynamically generated docstring
was a casualty of your workaround of changing ein:notebook-mode to
ein:notebook, which as we're both aware was a hack.

> whether this could give us any hint

I don't think the hanging (which, if I understood you, only happens when you
changed ein:notebook-mode to ein:notebook) has anything to do with the
original problem of the wrong keybindings.

> minor-mode-map-alist variable in my spacemacs, evil-surround-map doesn't
> appear at all

That is because you're using d7868da, and I am using the commit previous to
that d7868da^, which is the only sensible commit to use in addressing the
original problem.  Once we apply your workaround in d7868da, all bets are off.

> it's used at least once in ein-notebook.el, and a few times outside that file.

I see.  You must be referencing the `define-obsolete-variable-alias` of "ein:notebook".  I must say I'm unsurprised by the "tweak-and-guess" or
"hack-until-you're happy" approach to debugging spacemacs layers, which is a
natural outcome of "community" members rushing to paper over a more deepseated
deficiency in often superficial ways.

I believe this could all be resolved by changing `evil-define-minor-mode-key`
to `evil-define-key` which I've tested to work in
https://github.com/dickmao/spacemacs/commit/3f783cc43fd80dee6ad881c7fa260ca60d1a3ce6.

Like you, I must plead ignorance about the details of 
evil-mode (as I am a vanilla emacs user).  To reiterate, I concede the
dynamically generated docstring was disliked by some, but it's unfortunate the
final impetus to remove it wasn't so much that dislike but its hampering of a
not-well-understood workaround.

dickmao:
On an empathetic note, I too am guilty of wanting to summarily delete a bunch of code to facilitate maintenance, but upon [reasonable reproach](https://github.com/syl20bnr/spacemacs/pull/13348#issuecomment-596101033) (I will generally [dismiss reproach](https://github.com/millejoh/emacs-ipython-notebook/pull/594#issuecomment-539247190) redolent of preservationist obstinacy), I made a weekend effort to incorporate it.

lcos77:
>but upon reasonable reproach (I will generally dismiss reproach redolent of preservationist obstinacy), I made a weekend effort to incorporate it.

You did a lot of effort and deleting all that code from your previous PR might have felt like I invalidated all of your hard work. As a user, I wanted to fix the problems with the iptyhon notebook layer, so I read the documentation on layers and made some effort to read all of your code and while I could grasp most of it through the documentation that I found online, I didn't understood where did certain variables came from (that appeared to be bound to only one scope, yet that wasn't the case). The function that dynamically generated the docstring seemed:

-  too large (and in this sense tedious, at least for someone without experience in elisp)
- not too easy to test
- like it could be made more modular
- that it perhaps didn't belong to the ipython layer but to some utility function that could be used by many other layers. 

Because otherwise, it just seems like the handcrafted docstring is more parsimonious and even easier to maintain since even the auto-generated docstring might require maintenance, since after all there are many ways of doing it. (Though well I suppose that most could agree with something like the ein:notebook-worksheet- prefix, ein:notebook prefix, and ein:worksheet- correspond to worksheets, notebook files, and cells respectively, or could even be passed as arguments to the docstring generating function) 

Either way, I see it might have been unjust that by consensus my contribution (that deleted your autogen docstring function) was committed instead of through debate or through a way that would have felt more validating of your hard work.

> That is because you're using d7868da, and I am using the commit previous to
that d7868da^, which is the only sensible commit to use in addressing the
original problem. 

Well I tried it with d7868da and the modified version of d7868da (with the mode name changed to ein:notebook), but in both cases I couldn't find the evil-surround in the minor-mode-map-alist

> "hack-until-you're happy" approach to debugging spacemacs layers

I think this approach is convenient when the code isn't working, because this way, one might at least have a makeshift fix that works for me and all users (that can later be fixed more rationally). Still, I think eventually one would prefer  well-founded code over a simple hack.

> I believe this could all be resolved by changing `evil-define-minor-mode-key`
to `evil-define-key` which I've tested to work in
dickmao@3f783cc.

I tested your fix, and it's not working for the transient mode, which appears to be a result of an inconsistency betwen the ein:notebook of spacemacs/set-leader-keys-for-minor-mode and the ein:notebook-mode of the :evil-leader-for-mode:
- If both of them are consistent, the transient state works well. 
- If both of them are ein:notebook-mode, then the emacs keybindings are overshadowed by the other modes.
- If at least one of them is ein:notebook, then like before there is the extra ein:notebook in the minor-mode-alist. (which applies to this partially inconsistent fix)

In this sense, for your fix, we're more like in a catch 22.

layers/+lang/ipython-notebook/packages.el
```diff
@@ -15,47 +15,66 @@
...
-                        ("O" ein:worksheet-insert-cell-above-km)
-                        ("t" ein:worksheet-toggle-cell-type-km)
-                        ("RET" ein:worksheet-execute-cell-km)
-                        ("l" ein:worksheet-clear-output-km)
-                        ("L" ein:worksheet-clear-all-output-km)
-                        ("C-s" ein:notebook-save-notebook-command-km)
-                        ("C-r" ein:notebook-rename-command-km)
-                        ("x" ein:notebook-close-km)
-                        ("z" ein:notebook-kernel-interrupt-command-km))))
+      (let ((bindings '(("y" ein:worksheet-copy-cell)
```
dickmao: One mustn't remove the `-km` suffixes.  Those are important.

I don't know who "tkamat" is, but many of the commands like "ein:notebook-worksheet-open-prev-or-last", "ein:notebook-worksheet-move-prev", and "ein:notebook-worksheet-open-1th" are suspect, and upstream has deprecated.

layers/+lang/ipython-notebook/packages.el
```diff
@@ -65,116 +84,20 @@
                        'warn (format "ipython-notebook/init-ein: undefined %s"
                                      (cl-second bind))))))
                 (copy-tree bindings)))
-	(eval (append '(spacemacs|define-transient-state
```
dickmao: I see now why you needed "tkamat"'s bindings.  It's because spacemacs has some clever check to ensure the hardcoded docstring is "mated" to the bindings.  I am sympathetic to the hardcoded docstring, but as the primary EIN developer, the bindings from 72ea049 are highly advisable, if not non-negotiable, especially as the `-km` is important.

dickmao:
Sigh.  It's a wonder how anyone get any work done with emacs, much less spacemacs which add a whole other layer of fear, uncertainty and doubt also known as "too many cooks in the kitchen."

At the core is how one preserves all our evilified bindings and still have the `ein:notebook-mode` minor mode override the `python-mode` major mode as naturally happens in vanilla emacs.  Once `spacemacs/set-leader-keys-for-minor-mode` gets into the picture, the "shadowing" of the major mode doesn't happen, and things like C-c C-c defer to python-mode instead of ein:notebook-mode.  I even tried explicitly removing `ein:notebook-mode` from the `evil-mode-map-alist`, but the shadowing never occurs.  I am sure some spacemacs "old hand" knows how to do this.

Between this commit and 72ea049, I prefer the latter, not only because I wrote the latter, but I fully understand what's going on there.  This PR essentially works "by accident."  The arbitrary choice of `ein:notebook` is not a bonafide minor mode.  And as described in my review, tkamat's bindings are problematic.

On the one hand, I appreciate spacemacs for bringing a whole new audience of vimmers to emacs and EIN.  On the other hand, its highly decentralized development model (too many cooks in the kitchen) prevents me from ensuring a uniform EIN experience.  Well, as long as you guys can make it work for yourselves, it's probably not my place to object.

layers/+lang/ipython-notebook/packages.el
lcos77: >One mustn't remove the -km suffixes. Those are important.

Wow, it does solve the problem for me. Adding the -km bindings and being consistent in using ein:notebook-mode for the four cases, seems to solve it (I don't have neither the extra fuction in the minor-mode-map-alist, the emacs keybindings work well, and so does the evil keybindings!) 

>I don't know who "tkamat" is,

A user repository I found through Google :p 

> but many of the commands like "ein:notebook-worksheet-open-prev-or-last", "ein:notebook-worksheet-move-prev", and "ein:notebook-worksheet-open-1th" are suspect, and upstream has deprecated

It's a shame they were deprecated, It does seem convenient to be able to open several worksheets

layers/+lang/ipython-notebook/packages.el
lcos77: Yes but even then it does work for other keybindings too

lcos77:
>Sigh. It's a wonder how anyone get any work done with emacs, much less spacemacs which add a whole other layer of fear, uncertainty and doubt also known as "too many cooks in the kitchen."

Yeah like a decentralized bazaar haha. Well, I guess it does make it quite prone to errors and hanging. I was recently reading an essay called "designer as designed", that deals with these ideas about conceptual integrity in design, especially of software and poetry. I wonder if it could inspire some community to better organize itself.

>At the core is how one preserves all our evilified bindings and still have the ein:notebook-mode minor mode override the python-mode major mode as naturally happens in vanilla emacs.
 [...]  I am sure some spacemacs "old hand" knows how to do this.

I wonder, I kind of would like to have that knowledge now that I've been spending time reading part of the spacemacs code.
 

> This PR essentially works "by accident." The arbitrary choice of ein:notebook is not a bonafide minor mode.

Yes it was by accident, but maybe it works because both the keybindings were deprecated and so was the ein:notebook, though I can't really tell just from that. 

Anyway, does it work for you with the current fixes, the -km suffix, and the four substitutions from ein:notebook to ein:notebook-mode? If it does will you make a PR about it, or do I make one, or what do you suggest instead?

dickmao:
>  does it work for you with the current fixes, the -km suffix, and the four substitutions from ein:notebook to ein:notebook-mode?

Yes, but it's really not the route I prefer.

I have submitted a PR with a fair amount of commentary in #13439.

lcos77:
> Yes, but it's really not the route I prefer.

I do see you reverted the removal of the automated generation of the docstring
It's fine I guess, though it would be nice if it could become a spacemacs utility function instead

> I have submitted a PR with a fair amount of commentary in #13439.

Very well, and it does work!

dickmao:
Yes, the autogen is unanimously disliked by every comment I've seen.

As [I've previous stated](https://github.com/syl20bnr/spacemacs/pull/13348#issuecomment-596430032), I would not intercede if someone wants to construct a hardcoded docstring for the updated bindings, and remove the too-complicated autogen.  In its defense, it did save me having to construct a new docstring for the non-tkamat bindings (which I do insist on), and one day it might save you some time too!

lcos77:
> and one day it might save you some time too!

We'll see!! :p

Issue #13395, opened 2020-03-21
Develop - completion ivy-hydra required even if not using hydra


j-martin:
The support for ivy was modified here, but now forces the user to install `ivy-hydra` regardless if it used or not: https://github.com/syl20bnr/spacemacs/pull/13377/files#diff-570753425f7f850ae31f63faa65fbe33R139

It broke my setup after pulling from develop today. I am on bb8ec23ec1794fb94f1fb4c4459af20fa3620c9c

I worked around the issue by adding it to my `dotspacemacs-additional-packages`.

nixmaniack:
The [functionality](https://github.com/abo-abo/swiper/blob/64f05f4735bba8b708bc12cfc2cbfb7fb7706787/ivy.el#L451) is part of `ivy` package.
The recent changes decided to use the upstream version instead of maintaining spacemacs version. 

Also, `ivy-hydra` was part of spacemacs `ivy` layer for a long time – just not used where it was deemed I think. 

It'd be better if you can describe how it broke your setup.

Finally, if I look back at the changes of that PR - I think we can get rid of that explicit `require` block since the function is [autoloaded](https://github.com/abo-abo/swiper/blob/64f05f4735bba8b708bc12cfc2cbfb7fb7706787/ivy.el#L467-L468). @dalanicolai Did you face any issues if explicit require was not there?

nixmaniack:
But again, removal of that `require` won't bring you the functionality under `C-o` if you're using `helm` layer and installing `ivy` as an extra package. In essence, spacemacs considers these two completion frameworks mutually exclusive(or at least one shadowing the other) as far as I know hence that would be a non-issue.

dalanicolai:
@nixmaniack Indeed removing that require block does not break ivy or ivy-hydra functionality. I just uncommented it when I removed  the custom ivy-hydra menu (that I removed in this PR), as it seemed it had been there before that custom menu was implemented. Anyway... I will just comment it out again, as it was before this PR which should fix this issue.

j-martin:
@nixmaniack I should have been clearer. I am using `helm` (the default completion system) and not the `ivy`. When loading emacs/spacemacs, I get a missing `ivy-hydra` error because of the `(require 'ivy-hydra)` that was added. Because the error occurs fairly early in the loading process, it leaves spacemacs in a barely usable state.

[...]

PR #13397, opened 2020-03-21
fix issue #13395

dalanicolai:
Ivy-hydra is autoloaded so the require statement here is unnecessary. Because this require (commented out) statement was here before I uncommented it in the last PR (#13377), I leave it here (commented out) as it was before that PR.


smile13241324:
Thanks for this fix :+1:, I was some days of and didn't notice the issue.

PR #13404, opened 2020-03-23
[helpful] Bindings and documentation improvement

madand:
New bindings:
`gr` - `helpful-update`, reload helpful buffer.
`SPC m q` - `helpful-kill-buffers`, kill all helpful buffers.

Move `o` binding definition into separate post-init function, so the binding
is defined only if the `link-hint` package is actually installed.

Add documentation of the added bindings.


smile13241324:

PR #13421, opened 2020-03-27
[clojure] Pretty print evaluation keybindings

jr0cket:
Show results of evaluating Clojure code in the same format they would be written
in the source code.  This is especially useful for results that are
collections and nested collections.

Pretty print results as a comment.


smile13241324:

PR #13427, opened 2020-03-28
Replace org-ref-helm-insert-cite-link with org-ref-insert-link

tshu-w:
replace `org-ref-helm-insert-cite-link-with` with a more general function `org-ref-insert-link`, incase someone prefers others completion library as `org-ref-completion-library` instead of `org-ref-helm-bibtex`.


smile13241324:

PR #13430, opened 2020-03-29
Added some tips for macOS use of the pdf layer

NightMachinary:


layers/+readers/pdf/README.org
```diff
@@ -49,6 +49,24 @@ If you use a Debian based system such as Ubuntu, you can skip this step, as
 =pdf-tools= installation asks whether you want it to try to download
 automatically all dependencies using =apt-get=.
 
+On macOS, the default instructions might not work. There is a brew tap at dunn/emacs that contains pdf-tools, but it did not work for me. I used the following successfully. Note that your PATH should be set correctly in Emacs, and that pdf-tools does not work in terminal (GUI-only).
+
+#+BEGIN_SRC
```
smile13241324: Org source blocks must always set a language. See CI error:
`error: Language not specified in "#+BEGIN_SRC <language>"`

layers/+readers/pdf/README.org
```diff
@@ -49,6 +49,24 @@ If you use a Debian based system such as Ubuntu, you can skip this step, as
...
+brew install poppler automake
+#+END_SRC
+
+Then in emacs run the following:
+
+#+BEGIN_SRC
+(use-package pdf-tools
```
smile13241324: Can you explain why you require a `use-package` declaration? Is it just for setting the variable? This could have been set much more easier during the layer declaration i.e. `(pdf variables: pdf-info-epdfinfo-program)`.

layers/+readers/pdf/README.org
```diff
@@ -95,6 +113,7 @@ Due to =image-mode=’s behaviour, the =pdf-tools= buffer might go back to the
 first page after switching windows. If this happens, you can quickly go back to
 the previous page in history with =pdf-history-backward= (~``~ in Evil state and
 ~B~ in Emacs state.)
+On some machines, pdf-tools can be quite slow (especially with a line-number mode active). You might want to try alternatives such as zathura.
```
smile13241324: A short link or explanation how to install `zathura` and why it is better/faster would be useful in the docs.

tshu-w:
The default instructions work on my side. (macOS 10.15.4 emacs-mac 26.3)

layers/+readers/pdf/README.org
NightMachinary: Truth to tell, I copied this off stackoverflow, and don’t know if it’s necessary. (But it certainly worked for me.)  I figured people can improve on it if they wish. I personally found pdf-tools to be glacial, and did not investigate it further. I’ll incorporate your suggestion as an alternative.

layers/+readers/pdf/README.org
NightMachinary: Zathura is not an Emacs package though. Should I add some explanation/links?
Zathura is basically a specialized pdf viewer with a vim-like interface. It’s fast. And it works pretty much out of the box. It’s available on both macOS and linux, though I doubt Windows support.

smile13241324: I think in this case it is enough to say to use an external PDF reader though it somehow defeats the purpose, I mean the idea is to do that with emacs otherwise I could always use the native PDF reader of the respective OS, I think all big three have one now a days.

smile13241324:
I think this can be closed as the current default seems to be working on Mac. Feel free to reopen if this is not solved and this change is still relevant.

PR #13432, opened 2020-03-29
[chrome] Add link to more details about chrome jquery workaround in docs

bhipple:
We probably want to update the `markdown` layer to be able to do this directly
with `firefox`; having to kill and restart all of your `chrome` processes before
using `flymd` is a little clunky! It's nice that we have the upstream workaround
inlined in spacemacs, though.


smile13241324:

smile13241324:
You are right, the markdown preview features should rather be moved to the markdown-layer. However there is already at least one browser backend implemented, but I am not sure which one it is using. I will have a look in the next days.

PR #13438, opened 2020-03-31
Fix bug with keyboard-layout layer an bépo layout

Phundrak:
When the bépo layout is set from the `keyboard-layout` layer, the message `(wrong-type-argument keymapp nil)` shows up. After some digging, it turns out Spacemacs tries to set shortcuts for `nil` `map` variables in this function.

This commit ensures it is impossible to try to set a shortcut for an inexisting keymap.

However, I think some investigation should be done as to why Spacemacs tries to set keyboard shortcut for empty keymaps.

This commit has been tested with Emacs 27.0.90 with the latest Spacemacs develop commit at the time of my own commit.


smile13241324:

PR #13439, opened 2020-03-31
[ipython-notebook] bind-map.el counterfeits ein:notebook-mode

dickmao:
`bind-map` replaces on the `minor-mode-map-alist` the original
`ein:notebook-mode-map` with the evilified map.

This might be intentional as having active both the
original keymap and the evilified keymap might incur fear, uncertainty and doubt.

In EIN's case, yes, we really want the original keymap, e.g., `C-c C-c`
to execute the present cell, to work.  Otherwise `C-c C-c` defaults to running a python
shell via the `python-mode-map`, which is really not what we want.

As discussed at length in #13392, the primary developer of EIN insists
on updating the spacemacs bindings to avoid several-iterations stale commands of
dubious functionality (in particular, multi-worksheet commands).

As the transient state docstring must be "mated" to the updated bindings, he prefers
not having to manually curate it, and exhorts the return of the
automated docstring.  He doesn't insist on it, though, like the above.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2136,6 +2136,7 @@ Other:
   - Fixed keybindings per removal of =ein:notebook-multilang-mode=
   - DRY and tidiness update by primary EIN developer
   - Dynamically generate transient-state docstring.  Remove ob-ipython (Ibid.)
+  - Avoid counterfeiting of ein:notebook-mode-map (Ibid.)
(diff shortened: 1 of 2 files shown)


smile13241324:
Thank you for contributing to spacemacs :purple_heart:, I have cherry picked your commits into develop.

PR #13441, opened 2020-04-01
add and document org-journal scheduled(-view) shortcuts

dalanicolai:
`org-journal-new-scheduled-entry` and `org-journal-schedule-view` are both basic org-journal features.  `SPC a o j t` adds `t` for TODO under the org-journal prefix (`t` because it adds a TODO). 

Besides providing these shortcuts, their (self-)documentation exposes this functionality to the user.


smile13241324:

PR #13454, opened 2020-04-03
[imenu-list] Added keybinding to focus the imenu-list sidebar, creating if necessary.

cormacc:
The imenu-list layer currently defines a single global binding, which toggles the imenu tree sidebar.
With only a single buffer open, it's easy to refocus the sidebar using `spc w l` (the created window doesn't get an index for easy navigation), however this gets annoying with multiple buffers open. I've added a new function to focus the sidebar, creating it if it hasn't already been created.

New function bound under `spc b i`.
Existing toggle binding moved under UI toggle prefix: `spc T i` (previously `spc b t`). As this seems more consistent with the project binding mnemonics...

Bindings now using spacebind macro (via use-package keyword mechanism). Because it's shiny and new.


smile13241324:

PR #13455, opened 2020-04-04
Removed company-lsp in favour of company-capf

yyoncho:
- Also removed some cquery artefacts which accidentally ended up in this PR
instead of in #13446


smile13241324:

TheZano:
woah, now how can i get the similar effect of fuzzy match from the variable `company-lsp-cache-candidates auto`

PR #13458, opened 2020-04-05
Added dap inialization for php layer. Added dap using manual to php l…

programmador:
Added ability to use dap for php layer as I promised in gitter.

Used `with-eval-after-load` for a source path mapping configuration example, tentatively checking that everything is actually working.

Sorry, I still didn't add any automation for conditional running of `dap-php-setup` command.
Added a temporary manual of when and how to run it manually instead.

### DIFF
--- a/layers/+lang/php/funcs.el
+++ b/layers/+lang/php/funcs.el
@@ -13,3 +13,14 @@
   "Conditionally setup php backend."
   (pcase php-backend
     (`lsp (lsp))))
+
+(defun spacemacs//php-setup-dap ()
+  "Conditionally setup elixir DAP integration."
+  ;; currently DAP is only available using LSP
+  (pcase php-backend
+    (`lsp (spacemacs//php-setup-lsp-dap))))
+
+
+(defun spacemacs//php-setup-lsp-dap ()
+  "Setup DAP integration."
+  (require 'dap-php))
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2762,6 +2762,7 @@ Other:
   - Added =refactoring= and =debugging= key bindings (thanks to Tyoma Kostyuk)
 - Improvements:
   - Enabled =evil-matchit-mode= (thanks to duianto)
+  - Added debugger integration for LSP backends via =dap= layer (thanks to Alexander Konotop)
 - Fixes:
   - Added company-php (thanks to jim and Eivind Fonn)
   - Fixed php-company autocompletion (thanks to Dela Anthonio)
--- a/layers/+lang/php/README.org
+++ b/layers/+lang/php/README.org
@@ -10,16 +10,20 @@
 - [[#install][Install]]
   - [[#layer][Layer]]
   - [[#backends][Backends]]
-    - [[#intelephense][intelephense]]
-    - [[#php-language-server][php-language-server]]
+    - [[#lsp][LSP]]
+      - [[#intelephense][intelephense]]
+      - [[#php-language-server][php-language-server]]
+      - [[#debugging][Debugging]]
     - [[#ac-php-core][ac-php-core]]
       - [[#setup][Setup]]
       - [[#refactoring][Refactoring]]
+      - [[#debugging-1][Debugging]]
 - [[#key-bindings][Key bindings]]
   - [[#general][General]]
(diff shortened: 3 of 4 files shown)


layers/+lang/php/funcs.el
```diff
@@ -13,3 +13,14 @@
   "Conditionally setup php backend."
   (pcase php-backend
     (`lsp (lsp))))
+
+(defun spacemacs//php-setup-dap ()
+  "Conditionally setup elixir DAP integration."
+  ;; currently DAP is only available using LSP
+  (pcase (spacemacs//php-backend)
```
duianto, line 20: Is this supposed to be:
```suggestion
  (pcase php-backend
```
Then it would match the way the backend is checked in the `spacemacs//php-setup-backend` function on line 14 above?

Because I'm not finding any definition of a function that's called: `spacemacs//php-backend`

programmador: Fixed in 491d0a4fdd256c466717e723e69b735ca4c92a74 buuuuut...
Actually it was working fine before a fix.

Silly but if You ask me what exactly was working - unfortunately I don't have an answer. I've copypasted the code from python layer and it seems like python and php layers don't follow same conventions so 'spacemacs//' prefix is odd for php..

So actually I don't know whether this code is needed at all if dap-php works fine both before and after fix. Anyway for python it probably does smth if exists. Similar code exists in other layers too AFAIR.

programmador:
I've also updated changelog for CI to stop blaiming. Please tell if I've written smth wrong there.

layers/+lang/php/funcs.el
duianto: The function name conventions are:
https://github.com/syl20bnr/spacemacs/blob/master/doc/CONVENTIONS.org#spacemacs-core-and-layer
- spacemacs/xxx is an interactive function called xxx
- spacemacs//xxx is a private function called xxx (implementation details)
- spacemacs|xxx is a macro called xxx

The changelog is in the right place and it looks fine, it's just missing the thanks parentheses:
`(thanks to <your name>)`, I know it's strange to thank yourself 😄 but that's the convention, and it helps when the PR author already has added it.

The `spacemacs//php-setup-dap` function isn't used anywhere, so currently it has no effect.

Could you squash the commits, it's one commit per PR (unless it's a complex PR).

programmador: Thank's for a link to conventions, it will be useful.
I'll re-check and fix everything and squash commits.

programmador:
@duianto Please review commit c883cfc7de6df48d111d6c1fde38694465493fe8
I've changed
`(add-hook 'php-mode-local-vars-hook #'spacemacs//php-setup-lsp-dap))`
to
`(add-hook 'php-mode-local-vars-hook #'spacemacs//php-setup-dap))`
That's why You didn't find any calls to `spacemacs//php-setup-dap`.
I had to call a 'generic' function which first checked whether we are using LSP backend and only in case of positive result called an 'lsp' function.

So before the fix it looks like dap could be initialized for non-LSP backend which is undesired.
But it's not the most interesting. Actually I've made this mistake probably because I was copypasting different fragments from python and java layers. And actually python calls the right function:
https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Blang/python/packages.el#L124
and java calls the wrong one (as I did in PHP before fix):
https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Blang/java/packages.el#L35
though java layer has the 'generic' function
https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Blang/java/funcs.el#L31
You also won't find any calls to the 'generic' one across the whole Spacemacs.
Looks like a bug, isn't it?

BTW I have no idea why every single lang layer avoids enabling dap for non-lsp backends. Presumably it really won't work this way.

smile13241324:
Well dap is a specific capability an lsp server may support or not. Therefore activating it without having lsp activated does not make much sense.

You are right with the java layer, if dap layer is in your dotfile but you are happening to use a different backend for java i.e. Meghanada dap will be loaded and most probably just fail. I will make a fix in the next view days.

Anyway please squash all your commits and double check the CI results. If you are ready for another review please request one via the github api. This allows us to see directly at the beginning of an PR whether it is already taken.

smile13241324:

PR #13462, opened 2020-04-08
[julia] Fix inconsistencies in backend configuration

mjkramer:
Just wanted to fix a few rough edges in the Julia layer with respect to backend configuration:

- The README (correctly) says to enable LSP by setting `julia-backend` to `'lsp`, but the example snippet suggests (incorrectly) to set `julia-mode-enable-lsp` to `t`. In reality, `julia-mode-enable-lsp` is never actually used by the layer (despite being `defvar`'d). This PR gets rid of `julia-mode-enable-lsp` and gives `julia-backend` a docstring.

- Eliminate trivial `spacemacs//julia-backend` function.

- In `julia-mode-local-vars-hook`, put `spacemacs//julia-setup-backend` instead of `spacemacs//julia-setup-lsp`. The latter will still be called by the former if `julia-backend` is `'lsp`. With this change, if we ever add another backend, it will be properly configured on buffer init (and also, now we don't call `spacemacs//julia-setup-lsp` when we're not using LSP).

- Make the docstring of `spacemacs//julia-setup-buffer` more vague, in case we want to add additional, non-ESS/LSP-related stuff here in the future.

### DIFF
--- a/layers/+lang/julia/README.org
+++ b/layers/+lang/julia/README.org
@@ -43,7 +43,7 @@ for installation, and then install this layer with:
   (setq-default
    dotspacemacs-configuration-layers
    '(
-     (julia :variables julia-mode-enable-lsp t)))
+     (julia :variables julia-backend 'lsp)))
 #+END_SRC
 
 =LanguageServer.jl= tends to have a very long startup time. In the worst case,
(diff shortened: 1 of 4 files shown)


duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/d95d41f55ba074128b3ed8f0eedcdc3c31b52e42

PR #13463, opened 2020-04-09
[treemacs] Complete treemacs + persp-mode integration.

Alexander-Miller:
Finished treemacs-persp integration by actually changing the scope type - I don't think that option was available when treemacs-persp was first added to spacemacs.

Not sure about making it opt-in, since the previous setup did attempt to activate it by default, but given that perspective-scoping was never actually enabled until now I thought the default should stay as is.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -3248,6 +3248,7 @@ Other:
 - Deleted default values to track upstream behavior
   (thanks to Michael Peyton Jones)
 - Removed obsolete options (thanks to Michael Peyton Jones)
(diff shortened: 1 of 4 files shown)


nixmaniack:
I, for one, would definitely suggest this to be a default. Since, we're still on `develop` and `treemacs` isn't part of released `master`, pushing this sane default would make sense—IMHO.

duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/3a1bfc2a91a9b4e685249eea82a94537f58db489

PR #13471, opened 2020-04-10
add js log key bindings to react layer

coldpour:
add key bindings for log functions from the javascript layer to the react layer since they are oh so useful


smile13241324:

PR #13479, opened 2020-04-11
Modified Clojure layer's major leader binding sl to be smarter and fi…

didibus:
…nd the associated repl when in a non cider repl buffer and clear it, instead of throwing an exception, which is what it currently did.

In Clojure major mode, `sl` used to be set to a command that would clear the cider repl buffer, but it only worked if you were currently in the cider repl buffer. If you were in another buffer, it would throw an exception.

I changes it so that if you are not in the cider buffer, it will now try to find the cider buffer using the associated cider buffer info of the buffer you are in, and clear the cider repl buffer.


smile13241324:

PR #13482, opened 2020-04-12
[Python][LSP] Fix the issue autocomplete does't work in simple python file or python project with lsp enabled when using lsp as python-backend

c02y:
When using lsp as python-backend:

1. When I'm editing simple python file without lsp enabled, company doesn't work at all, company-mode is even not enabled, 

2. When I'm editing Python project with lsp enabled, company doens't work as expected, it echoes error message.


c02y:
Sorry about this PR, maybe I just don't know how to use lsp for python, I cannot get any auto completion for a simple python script, I'll just use anaconda as python-backend.

PR #13483, opened 2020-04-12
add pdf-mode keybindings to scroll through history

dalanicolai:
Added to pdf-view and pdf-view transient-state `[` and `]` keys to scroll through backward/forward through history. These keys were still free in both states while all other logical alternatives were already taken. Until now there was no key-binding to scroll forward through history, while no history key-bindings were presented at all in the transient-state.

### DIFF
--- a/layers/+readers/pdf/README.org
+++ b/layers/+readers/pdf/README.org
@@ -101,45 +101,45 @@ If you use Emacs editing style, check the key bindings at the [[https://github.c
 
 ** PDF View
 
-| *Key binding*      | *Description*                               |
-|------------------+-------------------------------------------|
-| *Navigation*       |                                           |
-|------------------+-------------------------------------------|
+| *Key binding*        | *Description*                             |
+|----------------------+-------------------------------------------|
+| *Navigation*         |                                           |
+|----------------------+-------------------------------------------|
 | ~M-SPC~ or ~s-M-SPC~ | pdf-tools transient state                 |
-| ~0/$~              | Left/right full scroll                    |
-| ~J~                | Move to next page                         |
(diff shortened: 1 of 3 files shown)


layers/+readers/pdf/README.org
```diff
@@ -101,43 +101,45 @@ If you use Emacs editing style, check the key bindings at the [[https://github.c
 
 ** PDF View
 
-| *Key binding*        | *Description*                             |
-|----------------------+-------------------------------------------|
-| *Navigation*         |                                           |
-|----------------------+-------------------------------------------|
+| *Key binding*      | *Description*                               |
```
smile13241324, line 104: The headlines are not longer correctly aligned. Please align the headlines like below:
```
|----------------------+-------------------------------------------|
| *Navigation*         |                                           |
|----------------------+-------------------------------------------|
```

dalanicolai: Hi, of course I would like to align the table correctly but this is how org-mode aligns them. Of course I just used TAB to let org-mode align the table automatically, so on my system this file shows like shown in the image below
image
Indeed when I switch to text mode it looks like the table in the text file.
When I align the table in the text file and switch to org-mode then the table looks like in the following image
image
But I guess the alignment is not very important as the table looks fine here on github anyway.
If my org-mode does the alignment wrong then let me know (and also let me know if you know a solution). Also, I am fine to push the manually text-mode aligned table if prefered

smile13241324:
Please push the manually aligned tables. The docs should not be
autoformated normally this produces lots of errors during the various
conversions (html, markdown).

smile13241324:

PR #13490, opened 2020-04-16
[dap] Provide global keybindings for dap-mode

yyoncho:
- added the same bindings you normally have under `SPC m d` under `SPC d`
similar to what Debug layer does. Most of the dap-bindings are not content
sensitive(e. g. rerun last debug session) and it doesn't make sense to restrict
them only to particular major modes.
- Moved `spacemacs/browse-docs-online-at-point` from `SPC d b` to `SPC h b d` to
avoid collisions.


smile13241324:

PR #13492, opened 2020-04-16
[visual] Pin dap debug buffers to the bottom

yyoncho:
- reuse the bottom window for all windows that are in compilation mode and in
dap-server-log-mode. Without this change, you will get two windows on top of each
other when you do `SPC c c` and then `SPC m d d d`


smile13241324:

PR #13495, opened 2020-04-16
[defaults] Added shortcut to switch to last compilation buffer

yyoncho:


CHANGELOG.develop
```diff
@@ -1012,6 +1012,7 @@ Other:
   - Added ~SPC x d l~ =delete-blank-lines= (thanks to duianto)
   - Added keybindings for Easy PG built in Emacs application
    (thanks to John Stevenson)
+  - Added ~SPC c b~ to switch to last compilation buffer(thanks to Ivan Yonchovski).
```
sdwolfz, line 1015: How about making it mimic the behaviour of `SPC b p` and `SPC b n`, had have two functions that havigate back and fourth?

yyoncho: Or maybe better to emulate `SPC TAB` and have one `SPC c b` to go back to the original buffer when invoked in compilation buffer? 

In both cases, what do you think should be the "go back implementation"? `other-window`?

sdwolfz: Actually never mind, the more I think about my suggestion the dumber it sounds. Your change should be good enough.

smile13241324:

PR #13498, opened 2020-04-16
[wakatime] readme, note api key in bug reports

choffee:
In the Wakatime layer add a line to the readme to highlight the fact that having the api key in the main layer config will submit it as part of a bug report. There is probably a better way to achieve this but at least this highlights it for now.

### DIFF
--- a/layers/+web-services/wakatime/README.org
+++ b/layers/+web-services/wakatime/README.org
@@ -62,6 +62,9 @@ Example:
                  ))
 #+END_SRC
 
+*** Note about api key and bug submission
+If you use the built in spacemace bug sumission tool it will include the snippit of code above, including your api key, don't forget to remove it before submitting.
+
 ** API Keys
 After this go to your wakatime account and have your API key handy
 [[https://wakatime.com/settings/account?apikey=true]].


sdwolfz:
I've had success with using envchain to store my secrets in my gnome-keyring and fetching them on demand, see and example here: https://www.sdwolfz.pro/snippets/how-to-use-restclient

Not sure how good it is here as it might block of the keyring is locked... but you can try it out by defining a similar function and putting it in your config.

As for a Spacemacs specific fix... IDK... we can probably tell people not to keep secrets in their config and show alternatives, but they are OS/person specific.

layers/+web-services/wakatime/README.org
```diff
@@ -62,6 +62,9 @@ Example:
                  ))
 #+END_SRC
 
+*** Note about api key and bug submission
```
smile13241324, line 65: Your new headline should be mentioned in the toc. Also there are some typos in your text you should fix.
```suggestion
If you use the built in Spacemacs bug submission tool it will include all of your layer configs including your secret api key. Please remove it before submitting.

Even better don't store your key directly in the layer config. Instead put the following code in your user-config which is set after your layer config:
#+BEGIN_SRC emacs-lisp
  (setq wakatime-api-key "your-api-key")
#+END_SRC

robbyoconnor:
Not a problem if you use a custom.el file.

smile13241324:
Thank you for contributing to spacemacs 💜, I have cherry picked your PR into develop and revised it myself.

PR #13504, opened 2020-04-18
[emacs-lisp] Added emr integration

yyoncho:
see https://github.com/Wilfred/emacs-refactor

- I got my inspiration for the keybindings from the clojure layer

### DIFF
--- a/layers/+lang/emacs-lisp/README.org
+++ b/layers/+lang/emacs-lisp/README.org
@@ -19,6 +19,7 @@
   - [[#additional-evaluation-functions][Additional evaluation functions]]
   - [[#format-code][Format code]]
   - [[#debugging][Debugging]]
+  - [[#refactoring-with-emr][Refactoring with emr]]
 
 * Description
 This layer gathers all the configuration related to emacs-lisp. This should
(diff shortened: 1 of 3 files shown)


layers/+lang/emacs-lisp/README.org
```diff
@@ -229,3 +230,22 @@ In =debugger-mode= (=Debugger= is displayed in major mode segment of the mode
 line)
 
 TODO
+
+** Refactoring with emr
+
+| Key binding | Description                  |
+|-------------+------------------------------|
+| ~SPC r f e~ | implement function           |
```
sdwolfz, line 238: Since these are major mode specific, they go under `SPC m r` and not `SPC r`

yyoncho: Fixed here and in changelog, thanks.

smile13241324:
Woa thats a great addition to the elisp layer, I wonder whether something similiar could be used for the other layers too ...

yyoncho:
> I wonder whether something similiar could be used for the other layers too ...

I guess here we should depend on what lsp servers could provide...

smile13241324:
Jep you are right :)

PR #13505, opened 2020-04-18
Update README.md with current osx app link method

librariandad:
In the macOS section, replace the deprecated `brew linkapps` command with the command currently specified in the emacs-plus homebrew package to create a link to the Emacs.app file in /Applications


librariandad:
My apologies.  I will go read the guidelines again.

PR #13507, opened 2020-04-20
[auto-completion] Set company-minimum-prefix-length to 1 per upstream recommendation

bhipple:
In the `emacs-lsp` plugin, it recommends setting this to 1:
https://github.com/emacs-lsp/lsp-mode/#recomended-settings-for-lsp-mode-related-packages

Having set it to 1 myself and used it for a day or two, it feels much more
useful, as the autocompletion popping up earlier reminds you to use it more!

### DIFF
--- a/layers/+completion/auto-completion/packages.el
+++ b/layers/+completion/auto-completion/packages.el
@@ -90,7 +90,7 @@
     :init
     (progn
       (setq company-idle-delay auto-completion-idle-delay
-            company-minimum-prefix-length 2
+            company-minimum-prefix-length 1
             company-require-match nil
             company-dabbrev-ignore-case nil
             company-dabbrev-downcase nil)


smile13241324:
I am not sure about this, if we really would go with LSP's recommendation we would have to go all the way, this would include:
* `(setq company-minimum-prefix-length 1 company-idle-delay 0.0)` resulting in auto completion to fire immediatelly when you type.
* `(setq read-process-output-max (* 1024 1024))` ;; 1mb --> increase the amount of data to be read from a foreign process.
* `(setq lsp-prefer-capf t)` --> use company capf instead of company lsp as capf caches results automatically. This setting makes lsp-mode prefer capf if both are available.

These changes would very much affect the default behaviour of emacs and would create serious performance issues with users running non lsp backends or lsp backend with emacs < 27 as there is no native support for json parsing available causing a lot of elisp parsing to occure in the background which is expensive.


Seriously I would like to get an opinion from @syl20bnr about this, especially as I would like to set these values only for lsp-mode and not affect the classic modes if possible.

bhipple:
Sounds reasonable, no terrible rush on this one.

As of c3f13d03910e6ae55e5c617ce7c7bd9cb9861d70, it appears spacemacs prefers `company-capf` anyways, so I think the remainder are the delays and `read-process-output-max` settings.  

We already have a configurable idle-delay value exposed at the layer, so maybe the right answer here is to just make this parameter configurable too.

smile13241324:
You may be right, lets make the company-minimum-prefix-length configurable too on the layer.
We should also set `read-process-output-max` in the core, this should not have any negative side effects.

smile13241324:
@bhipple I have finished the changes. You should now be able to make the necessary configurations via `auto-complete` layer variables. In addition the value of `read-process-output-max` can be set via a new dotfile variable which is only used for emacs 27 and above.

Can you double check with emacs 27 whether you notice any difference/issues? If all is working well I would close this ticket.

bhipple:
Thanks @smile13241324!  I've updated my spacemacs and everything appears to be working perfectly.

PR #13513, opened 2020-04-22
[javascript] proper hook setting for js-doc

thanhvg:
move stray execution in funcs into package init


smile13241324:

PR #13516, opened 2020-04-23
Update README.org of clojure layer to remove redundant `linting via...`

liuchong:


jr0cket:
Ah yes, the description feature list has two lines about Linting that are almost identical.  This is the right one to remove.  Thanks.

smile13241324:

Issue #13252, opened 2020-02-09
Org Mode consistency graph not rendering on Agenda View


ssbothwell:
#### Description :octocat:
Org Mode consistency graph not rendering on Agenda View

#### Reproduction guide :beetle:
1 Erase .spacemacs file to start with a fresh config.
2 Add `org` layer to new .spacemacs file.
3 Reload config.
4 Open an org file with habits in it. Add this file to the agenda list.

*Observed behaviour:* :eyes: :broken_heart:
5. Open Org Agenda Week View.
6. Witness that the consistency graph is being rendered correctly for habits.
7. Restart emacs.
8. Open Org Agenda Week View.
9. Consistency graphs are no longer being rendered. :(

*Expected behaviour:* :heart: :smile:
I would expect the consistency graphs to be rendered.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 26.3
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. f94fea920)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(emacs-lisp helm multiple-cursors org treemacs)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG SOUND DBUS GSETTINGS GLIB NOTIFY LIBSELINUX GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XDBE XIM MODULES THREADS LIBSYSTEMD

#### Backtrace :paw_prints:
```
<<BACKTRACE IF RELEVANT>>
```

smile13241324:
@ssbothwell I have merged the fix from @klochowicz. Please check whether this solves your issue and close the ticket accordingly.

ssbothwell:
I just pulled the latest from develop and restarted emacs. I am not seeing any diffference.

klochowicz:
I'm sorry to hear that. I don't recall doing anything more than this fix my issue which sounded similar to yours (but I'm on emacs 28.0.50, maybe that made the difference). Would you mind checking if org-habit module is enabled in your config (print result of the `org-modules` variable)?

ssbothwell:
`space h d v` `org-modules`:
```
org-modules is a variable defined in ‘org.el’.
Its value is
[...]

Issue #12003, opened 2019-02-19
What is the most proper way to add org modules?


Anton-Latukha:
#### Description :octocat:
Spacemacs has a specially crafted structure for Org-mode.

I want to clarify a question to solve for me and others in a doc/PR about this.

Officially by `org-mode` - additional (especially official) extensions to `org-mode` recommended to be installed by extending the `org-modules` list.
I asked a question on Gitter, and recieved some responses, as also found my own, but nothing really solid.
I what to clarify what is the most proper way to add official org modules to extend `org-mode` from the `org` layer.

---

Also good to have some input to maybe open other issue/PR. Default `org-modules` list has many modules enabled by default:
```elisp
(org-w3m org-bbdb org-bibtex org-docview org-gnus org-info org-irc org-mhe org-rmail org-eww)
```
And most of them probably not used by many if not majority people that use Spacemacs. Browsers AFAIK in Spacemacs not in great shape.
Probably it can be right to allow people to populate `org-modules` from the ground-up.

#### Reproduction guide :beetle:
- Start Emacs
Wonder what is the proper way to add `org-habit` to extend `org` layer.

*Observed behaviour:* :eyes: :broken_heart:
Declare installation of `org-habit` after the initialization of `org` layer.

*Expected behaviour:* :heart: :smile:
Something beautiful. Like:
```elisp
(org :variables
  org-modules '(
    org-habit
    org-capture
    org-id
    org-protocol
    org-brain
    )
)
```
#### System Info :computer:
- OS: gnu/linux
- Emacs: 26.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 95422ae5d)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm

agzam:
Modules have to be extended, not in prescriptive manner like you suggesting using `:variables`. Simply doing something like this should work `(add-to-list 'org-modules 'org-tempo t)`.

Anton-Latukha:
Yes, currently I do a number of (add-to-list ...) expressions.
It does not seem like the best way to do it.
1. On every module - Emacs runs an `add-to-list` expression that adds a single element to the beginning of the list.
2. Official `org-mode` documentation states to add new entries to the end of the list. So if respect that - `nreverse` -> `add-to-list` -> `nreverse` it back. For every element.  

I did:
```elisp
require 'org
(nconc org-modules
       '(
         org-capture
         org-habit
         org-id
[...]

PR #13522, opened 2020-04-25
[org] Enable org-habit module.

klochowicz:
Enable missing org-module.

fixes the open bug: 
https://github.com/syl20bnr/spacemacs/issues/13252

Last year's open discussion that touches similar topic:
https://github.com/syl20bnr/spacemacs/issues/12003

I believe that enabling it is a reasonable default, if org-agenda shows "Habit" enabled in the bottom of its UI and `org-habit` variables and functions are visible for the user.


smile13241324:

PR #13527, opened 2020-04-26
[fasd] Use helm-fasd in fasd layer when helm is present

yyoncho:
- Underestimated but a very handy package.


smile13241324:

PR #13529, opened 2020-04-27
[ruby] Add instruction on disabling the automatic insertion of encoding comment

x-ji:
`ruby-mode` by default inserts an encoding comment at the top of a `.rb` file, which triggers [an error from RoboCop](https://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/Cop/Style/Encoding) on Ruby 2.0+. [There is a variable to disable it](https://stackoverflow.com/questions/6453955/how-do-i-prevent-emacs-from-adding-coding-information-in-the-first-line). The instruction for doing so is added in the README for the Ruby layer.


smile13241324:

PR #13542, opened 2020-05-02
[treemacs] load treemacs-icons-dired-mode earlier

duianto:
Opening a dired buffer from the Treemacs buffer,
right after starting Spacemacs, shows the error message:
>treemacs-icons-dired--display: Wrong type argument: char-or-string-p, nil

Loading `treemacs-icons-dired-mode` from the `dired-load-hook`,
instead of from the `dired-mode-hook`, seems to fix it.

The `dired-load-hook` docstring says:
>Run after loading Dired.

The `dired-mode-hook` docstring says:
>Run at the very end of ‘dired-mode’.

---

A dired buffer can be opened from the Treemacs buffer:
- By pressing `SPC a d` and choosing a directory.
- Or with the cursor on a Treemacs projects directory and pressing: `oo`

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -3296,6 +3296,7 @@ Other:
   variable: =treemacs-use-scope-type= with the possible values:
   ='Frames= (default) or ='Perspectives= (thanks to Alexander Miller)
 - Added integration with =treemacs-icons-dired= package.
+- Loaded =treemacs-icons-dired-mode= earlier (thanks to duianto)
 **** Typescript
 - Key bindings:
   - ~SPC m E e~ is for =tide-fix= which was originally added as ~SPC m r f~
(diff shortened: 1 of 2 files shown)


Alexander-Miller:
I've independently run into the same thing in https://github.com/Alexander-Miller/treemacs/issues/480 - pity I only looked at the PRs here after doing my own debugging and coming up with my own fix :/

Anyway you should also remove the quote from the hook list, otherwise it will expand to this:

```Elisp
(add-hook 'quote-hook (function treemacs-icons-dired))
(add-hook 'dired-load-hook (function treemacs-icons-dired-mode))
```

emacs18:
I learned more about this issue afterwards.  I think following explains all the issues and what we need to do to solve this for good:
https://github.com/syl20bnr/spacemacs/issues/13620

PR #13547, opened 2020-05-03
Fixed agda layer commands

guilhermehas:
With newer commands, it is equal to C-c C-? commands

Thank you <3

### DIFF
--- a/layers/+lang/agda/packages.el
+++ b/layers/+lang/agda/packages.el
@@ -64,7 +64,7 @@
           ","   'agda2-goal-and-context
           "="   'agda2-show-constraints
           "SPC" 'agda2-give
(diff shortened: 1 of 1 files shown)


smile13241324:
@guilhermehas I have checked your commit and the change is ok, nevertheless your PR is against master which we cannot accept. Please make a new PR against develop and this can be merged.

Btw, while you are at it, it would be great if you could describe a bit more why the commands have been changed, is it like to be consistent and use the same as for emacs style or is there some change of functionality involved? Anyway please give a short description in the new PR.

guilhermehas:
In my agda version, 2.6.1. There are newer commands.
If you try to install Agda mode in this version. You can see all the changes.

PR #13558, opened 2020-05-06
[idris] Add to spacemacs-indent-sensitive-modes

kayhide:
Since `idris` is indent-sensitive like Haskell, Python, Agda or PureScript, it should be added to `spacemacs-indent-sensitive-modes`.

Without this configuration, spacemacs inserts indentation when pasting a fragment of codes, which is not desirable.


smile13241324:

PR #13565, opened 2020-05-07
[org-mode] [verb] fix binding for verb-response-headers-mode

thanhvg:
for verb-response-headers-mode It's a major mode not a minor one.

lazy keybinding for  verb-response-body-mode

Without this change `C-h k` will crash throwing verb-response-headers-mode
variable is void.

### DIFF
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -819,7 +819,6 @@ Headline^^            Visit entry^^               Filter^^                    Da
   (use-package verb
     :defer t
     :init
-    (progn
       (spacemacs/set-leader-keys-for-major-mode
         'org-mode
         "rf" #'verb-send-request-on-point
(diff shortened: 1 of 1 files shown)


lawrsp:
there is still has the problem before opened an org-mode file

thanhvg:
> there is still has the problem before opened an org-mode file

Try again I add change to account for the lazy loading

lawrsp:
it works now, 
maybe we could just delete that key binding, 
since verb has a simple 'q' key to quit that verb-response-headers-mode window

layers/+emacs/org/packages.el
```diff
@@ -831,12 +830,14 @@ Headline^^            Visit entry^^               Filter^^                    Da
...
         "rv" #'verb-set-var)
+    :config
+    (progn
       (spacemacs/set-leader-keys-for-minor-mode
         'verb-response-body-mode
         "rr" #'verb-toggle-show-headers
         "rk" #'verb-kill-response-buffer-and-window
         "rf" #'verb-re-send-request)
-      (spacemacs/set-leader-keys-for-minor-mode
+      (spacemacs/set-leader-keys-for-major-mode
```
federicotdn, line 840: My mistake, thanks for noticing this.

smile13241324:

PR #13573, opened 2020-05-10
[clojure] Change leader key to fix warning

v2okimochi:
I changed key sequence of `cider-eval-print-last-sexp` from `ep` to `epl`.
This PR fixes this warning when reloading spacemacs configuration:

```
Error (use-package): cider/:init: Key sequence e p ; starts with non-prefix key e p
```

If anyone have the better way than this PR to resolve above warning, I would like to agree it.

Thanks for Spacemacs which is a wonderful distribution.


smile13241324:

PR #13577, opened 2020-05-12
Update latest libvterm installation guide

weihua-lu:
Update latest guide from emacs-libvterm.

Install ubuntu(18.04) libvterm-dev will cause compilation error.

Emacs-libvterm shall download latest libvterm automatically if not installed.


smile13241324:

PR #13607, opened 2020-05-21
Allow config syntatx-checking-use-standard-error-navigation.

mijoharas:
When you select `next-error` (which defaults to `spacemacs/next-error`) it will either call `flycheck-next-error` or emacs `next-error`.

With this variable set, when flycheck mode is enabled, the next-error function will first check for flycheck errors, then check for compilation buffer errors. (Personally i think this should be the default, but I'm preserving behaviour).

### DIFF
--- a/layers/+checkers/syntax-checking/config.el
+++ b/layers/+checkers/syntax-checking/config.el
@@ -20,5 +20,8 @@
 (defvar syntax-checking-use-original-bitmaps nil
   "If non-nil, use the original bitmaps from flycheck.")
 
+(defvar syntax-checking-use-standard-error-navigation nil
+  "If non-nil hook into emacs standard error navigation")
+
 ;; Command Prefixes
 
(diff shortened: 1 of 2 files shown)


mijoharas:
Hmm... the changelog CI step is optional, seems a small edit, so not sure if I should put anything in there?

pietrodito:
I have the same issue on my PR but the CHANGELOG file seems not to be used since 2017...

mijoharas:
@pietrodito have a look at `CHANGELOG.develop`. I think the just move the sections over to `CHANGELOG.org` when a new release is shipped to `master`

smile13241324:
Adding a line to changelog.develop would be great. Contribution.org mentions how to add your change there however we do not enforce it right now. When we merge a PR it depends on how much time we have whether we do the extra work on adding a line to changelog.develop for the PR author or just leave it.

Still if you have the time, please consider adding a line there.

smile13241324:

PR #13609, opened 2020-05-21
Tune org-babel transient state.

pietrodito:
When navigating source code blocks, it is now possible to execute several of
them without exiting the transient state.

Thank you <3


smile13241324:

PR #13611, opened 2020-05-22
Added mustache-mode to CFEngine layer

nickanderson:
Mustache is the preferred templating in modern versions of CFEngine and it's
nice to have syntax highlighting when working with templates.

Now, when the CFEngine layer is in use, files ending in .mustache will be opened
in mustache-mode.


smile13241324:

nickanderson:
Thank you!

PR #13613, opened 2020-05-23
fixed typo by changing dired-load to dired-mode

emacs18:
Fixed a typo to fix treemacs-icons-dired-mode to work, i.e. icons are displayed within dired-mode.

With this fix, treemacs-icons-dired-mode just work first time.  Without this fix, no iconds are shown in dired mode.

I think it was just a typo to use dired-load instead of dired-mode.

First the intent is to enable treemacs-icons-dired-mode whenever dired-mode is setup.  Hence dired-mode-hook is the right place to do that.

Second dired-load is an internal function used within dired code to implement dired-do-load command which is bound to L within dired mode.  Thus dired-load assumes that current mode is dired-mode, extracts the name of the file at current cursor, then calls (load) on that file to load an emacs-lisp file.  Thus dired-load cannot be the correct symbol.

### DIFF
--- a/layers/+filetree/treemacs/packages.el
+++ b/layers/+filetree/treemacs/packages.el
@@ -97,7 +97,7 @@
 
 (defun treemacs/init-treemacs-icons-dired ()
   (use-package treemacs-icons-dired
-    :hook (dired-load . treemacs-icons-dired-mode)))
+    :hook (dired-mode . treemacs-icons-dired-mode)))
 
 (defun treemacs/pre-init-winum ()
   (spacemacs|use-package-add-hook winum


smile13241324 (approved):
Hmmm you are right, `dired-load` looks really strange.
I have tried that out locally but it did not change anything for me, but it also did not break anything.

Anyway as this fixes issues for others it should be merged soon.

smile13241324:

duianto:
It wasn't a typo, it was an attempt at fixing the error message:
>treemacs-icons-dired--display: Wrong type argument: char-or-string-p, nil

That occurred the first time a dired buffer was opened from Treemacs.

I used the `dired-load-hook` because when comparing their docstrings:

The `dired-load-hook` docstring says:
>Run after loading Dired.

The `dired-mode-hook` docstring says:
>Run at the very end of ‘dired-mode’.

That made it seem like the issue was caused because `dired-mode-hook` loaded to late.

Here's the PR:
[treemacs] load treemacs-icons-dired-mode earlier #13542

---

It did fix the error message, but this PR is probably correct because emacs18 knows more about how Emacs works and the error message doesn't occur anymore.

Possibly because of this recent Treemacs commit:
[Dired] Select an icon set for all existing dired buffers on minor-mode activation.
https://github.com/Alexander-Miller/treemacs/commit/1ff0a8f0d3ce771d71c200e39b75d895acf35ec2

When that line is commented out and this `fix typo` PR is applied.
```elisp
(treemacs--select-icon-set)
```

Then the following error message is shown:
>let*: Wrong type argument: char-or-string-p, nil

The first time a dired buffer is opened:
- Start Spacemacs
- Open Treemacs `SPC f t`
- Move the cursor to a directory
- Open it by pressing: `oo`

There's no error message the second time `oo` is pressed.
The dired buffer opens and shows the icons.

When the `(treemacs--select-icon-set)` line is uncommented (the current Treemacs state), then the dired buffer opens without any errors the first time.

@Alexander-Miller For future reference, my attempt at fixing the error message:
>treemacs-icons-dired--display: Wrong type argument: char-or-string-p, nil

using the `dired-load-hook`:
[treemacs] load treemacs-icons-dired-mode earlier #13542
doesn't seem to have been the correct solution.

But everything is working now.
With this PR applied and the Treemacs commit:
[Dired] Select an icon set for all existing dired buffers on minor-mode activation.
https://github.com/Alexander-Miller/treemacs/commit/1ff0a8f0d3ce771d71c200e39b75d895acf35ec2

Alexander-Miller:
>  doesn't seem to have been the correct solution.

It used to - when I was independently working on the same issue I came up with the same solution before finding your PR. All's well that ends well :man_shrugging: .

PR #13617, opened 2020-05-25
[doc] Remove duplicate keyboard shortcuts

zhd173:
delete `ta`,`tf`,`th`,`tw` duplicate key binding.


smile13241324:
Hi @zhd173,

Can you further explain why you think these bindings are duplicate? From what I see these are valid bindings in the transient state/tree and  should not be removed from the documentation.

smile13241324:

PR #13619, opened 2020-05-25
[keyboard-layout] evil-lisp-state

PiOverFour:
Hi, this patch provides keyboard mappings for the `evil-lisp-state` package. 
It may be related to the unmerged #11625, but this is a different mode altogether.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2389,6 +2389,7 @@ Other:
 - Fixed key remapping after =cl= package deprecation (thanks to Damien Picard)
 - Fixed =bepo= layout, commented out broken evil-window :common bindings
   (thanks to CharlesHD)
+- Added evil-lisp-state mapping (thanks to Damien Picard)
 **** Kivy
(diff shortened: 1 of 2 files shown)


PiOverFour:
Hi @smile13241324, thanks for reviewing this PR. I’m afraid I don’t understand what you mean by “where you need to dynamically reset `evil-lisp-state` bindings”.

I don’t know if you are familiar with the `keyboard-layout` layer, but its purpose is to rebind the `hjkl` keys for users of keyboard layouts different from qwerty, such as dvorak, colemak and bépo, so that whichever keys are at those four positions act as movement. So there is no key conflict per se, but without that modification, movement is almost impossible.
The PR only affects `evil-lisp-state-map`, and it is pretty much the same logic as other functions in this file—see for instance `keyboard-layout/pre-init-neotree`.

smile13241324:
Hi @PiOverFour,

Thanks for your kind explanation, looks like I did mix this up with core-keybindings.el. Now with your explanation this looks like a straight forward fix.

PR #13621, opened 2020-05-25
[latex] Fix jump handlers

mjkramer:
Two changes:

1. Manually add `spacemacs//init-jump-handlers-latex-mode` to `LaTeX-mode-hook`. Otherwise it doesn't get run, because `spacemacs|define-jump-handlers` puts it in `latex-mode-hook`, which is only run by the built-in `latex-mode`, not by AUCTeX. (Note that passing `LaTeX-mode` instead of `latex-mode` to `spacemacs|define-jump-handlers` does NOT work; see the commit message for further explanation.) Up to this point, there were no custom jump handlers for LaTeX, so it didn't matter, but see next change! 

2. Make `dumb-jump-go` the primary jump handler for LaTeX. Otherwise, the default is used, which prioritizes `evil-goto-definition` over `dumb-jump-go`. Dumb Jump tends to Just Work, while `evil-goto-definition` doesn't handle LaTeX very well, at least not without a TAGS table.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2407,6 +2407,7 @@ Other:
 - Fixed README to say auto-fill on by default. (thanks to Max Willsey)
 - Added ConTeXt mode to the key bindings (thanks to ft)
 - Fixed ~SPC m ;~ binding (thanks to Tianshu Wang)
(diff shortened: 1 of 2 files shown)


smile13241324:

PR #13629, opened 2020-05-28
Turn off ability to convert to unicode arrows which are deprecated in Scala

Kazark:
As of Scala 2.13, Unicode arrows are deprecated:

* https://github.com/scala/scala/pull/7540
* https://github.com/scala/scala-dev/issues/585
* https://github.com/scala/bug/issues/11210

Using one will give a deprecation warnings like so:

> The unicode arrow `⇒` is deprecated, use `=>` instead. If you still wish to
> display it as one character, consider using a font with programming ligatures
> such as Fira Code.

As such the Scala layer's version slick capability to replace ASCII arrows with
Unicode ones is no longer useful, and I have removed it.

Based on my tests it doesn't seem that there is a need for a more graceful way
to deprecate this: i.e. nothing fails if there is extra junk in `:variables`.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -3084,6 +3084,8 @@ Other:
 - Disabled =sbt-supershell= in =sbt-mode= (thanks to Rodolfo Hansen)
 - Enable new DAP and lsp-treemacs integration via metals
   (thanks to Rodolfo Hansen)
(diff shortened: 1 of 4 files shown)


duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/f5fddc4256399ff3cd479221f4a7adbe2db17dc3

PR #13639, opened 2020-06-04
[org] replace org-bullets with org-superstar

thanhvg:
org-bullets has been deprecated


smile13241324:

PR #13641, opened 2020-06-04
Move folding bindings to evil motion state map

### LINKED ISSUE #13640: Enable folding in non editable buffers
Folding with `za` via origami is useful not only in editable buffers.

E.g. I'd like to use in my `*exunit-compilation*` buffer to inspect nested data structures when debugging tests. Folding makes it much easier.

Unfortunately, folding keybindings are defined in `spacemacs-editing` layer using `evil-normal-state-map`. Pressing `za` tells me, they are unbound. We could change it to `evil-motion-state-map` instead.

Would you find it a good idea to move folding to `evil-motion-state-map` and enable it globally even for non-editable buffers?

The change was small, so I've already made a PR. In case you don't like it - feel free to reject it :)

tomekowal:
Closes #13640

Moves origami bindings from `evil-normal-state-map` to `evil-motion-state-map`.
It enables folding in non-editable buffers like compilation buffers.

### DIFF
--- a/layers/+spacemacs/spacemacs-editing/packages.el
+++ b/layers/+spacemacs/spacemacs-editing/packages.el
@@ -223,19 +223,19 @@
     :init
     (progn
       (global-origami-mode)
-      (define-key evil-normal-state-map "za" 'origami-forward-toggle-node)
-      (define-key evil-normal-state-map "zc" 'origami-close-node)
-      (define-key evil-normal-state-map "zC" 'origami-close-node-recursively)
-      (define-key evil-normal-state-map "zO" 'origami-open-node-recursively)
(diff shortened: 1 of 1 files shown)


smile13241324:

duianto:
This change seems to have broken origami folding in non evil motion state buffers:
Some origami folding bindings broken #13683

tomekowal:
True, evil looks first in evil-normal-mode-state-map and if it does not find the key binding, it falls back to evil-motion-state-map, so I assumed it will work for both modes. I didn't take into account that those bindings are already set in normal mode.
Two solutions for this are:
- set the mapping for both modes
- unset the mapping for evil-normal-mode-state-map so that fallback mechanism works correctly.

PR #13645, opened 2020-06-05
[layout] fix persp-mode integration with helm v3.3

thanhvg:
Helm v3.3 no longer uses ido to get buffer list. So the change perp-mode makes
to ido won't affect helm anymore. Hence helm buffer list commands such as
helm-mini and helm-buffer-list will shows all buffers regardless of current
layout. Also `SPC l l` crashes with new helm.

This PR fixes them. We use the new helm variable helm-buffer-list-reorder-fn
to filter buffers. Because we compose this value users are still able to
customize this variable.


smile13241324:

PR #13651, opened 2020-06-06
[core][keybindings] add select recent buffer transisient state

thanhvg:
Inspired by layout `SPC l`. Implement the same one for recent
buffers by order of most recent display or selection. The state is bound to `SPC .`

The buffer list is responsive, it is cut off to fit one line it it is too long.
Press number to switch to buffer. `b` to show the buffer list.

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/keybindings.el
+++ b/layers/+spacemacs/spacemacs-defaults/keybindings.el
@@ -1014,3 +1014,73 @@ If FRAME is nil, it defaults to the selected frame."
   'spacemacs/scale-transparency-transient-state/spacemacs/toggle-transparency)
 
 ;; end of Transparency Transient State
+
+;; select buffer transient state
+(defun helm-buffer-p (buffer)
+  (string-prefix-p "*helm"
+                   (buffer-name buffer)))
+
+(defun spacemacs//get-recent-buffers ()
+  (seq-filter  (lambda (buff)
+                 (and
+                  (funcall (frame-parameter nil 'buffer-predicate) buff)
+                  (not (helm-buffer-p buff))))
+               (buffer-list)))
+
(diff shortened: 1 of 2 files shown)


layers/+spacemacs/spacemacs-defaults/keybindings.el
```diff
@@ -1014,3 +1014,73 @@ If FRAME is nil, it defaults to the selected frame."
...
+  ("1" spacemacs/switch-to-buff-1 :exit t)
+  ("2" spacemacs/switch-to-buff-2 :exit t)
+  ("3" spacemacs/switch-to-buff-3 :exit t)
+  ("4" spacemacs/switch-to-buff-4 :exit t)
+  ("5" spacemacs/switch-to-buff-5 :exit t)
+  ("6" spacemacs/switch-to-buff-6 :exit t)
+  ("7" spacemacs/switch-to-buff-7 :exit t)
+  ("8" spacemacs/switch-to-buff-8 :exit t)
+  ("9" spacemacs/switch-to-buff-9 :exit t)
+  ("0" spacemacs/switch-to-buff-0 :exit t))
```
sdwolfz, line 1083: I thin you need to remove `:exit t` from `switch-to-buff-*`, otherwise there is no point in having a transient state for this. The whole idea of transient states is to be able to execute multiple namespaced actions without typing the prefix again.

thanhvg: Thanks for the feedback, I have to admit that I now find my PR not very useful. I thought a single line of buffer list would help me jump back to a buffer faster. But it is actually slower than `spc b b`. It seems that my eyes prefer to scan content vertically than horizontally. 

I can bind `C-#` to to do `switch-to-buff-*` without `:exit t` like what `spc l .` does but again it doesn't provide much improvement than `[b`or `]b`

PR #13656, opened 2020-06-08
[which-key] fixup for change to help text for "buffer-to-window-N" so…

skebanga:
… which-key-replacement-alist works again

Fixup for regression introduced in commit #727d45e69 where the which-key text was changed from the function name (`buffer-to-window-1` etc) to a string "Move buffer to window 1".

The associated change to the `which-key-replacement-alist` wasn't made, meaning the previous behaviour of collapsing the 1..9 options of `buffer-to-window-N`  into a single entry was lost.

This change restores that behaviour (while keeping the next string)


smile13241324:

PR #13660, opened 2020-06-09
fix: org-store-link doesn't work unless you load mu4e

agzam:
When mu4e layer is present, org-store-link doesn't work immediately when you
start Emacs, unless you explicitly load mu4e.

It'd display "Please load mu4e before mu4e-org" message and do nothing.


smile13241324:

PR #13661, opened 2020-06-09
[Scala] Add lsp-metals

bjarkevad:
The Scala layer doesn't currently work with lsp, this PR fixes that.

Having an explicit dependency on lsp-treemacs for post-init-lsp-treemacs is no
longer necessary in the Scala layer, as lsp-metals superseeds it.
Removed lsp-metals-treeview-enable as it no longer exists.


bjarkevad:
Updated changelog

smile13241324:

PR #13665, opened 2020-06-10
[layouts] update switch worskpace key bindings

jr0cket:
Emacs 27.1 has tab bar functionality built in and over-rides the existing
keybindings for `eyebrowse-next-window-config` and
`eyebrowse-previous-window-config`

Change keybindings to `g>` and `g>` so they are no longer over-ridden.

Sorry, had to correct an error in the docs, pushed an updated commit.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1035,6 +1035,11 @@ Other:
     (thanks to John Stevenson)
   - Added ~SPC c b~ to switch to last compilation buffer
     (thanks to Ivan Yonchovski)
+  - Fixed eyebrowse-next/previous-window-config commands key bindings,
+    over-ridden by tab-bar-switch functions in Emacs 27.1 onward
+   ~g>~ eyebrowse-next-window-config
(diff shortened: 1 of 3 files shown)


jr0cket:
Open to other suggestions as I appreciate changing key bindings is not the smoothest option.

I could not figure out where the new code was for the tab bar key bindings.  There didn't seem to be anything in the `emacs.d` directory.

Would putting the original eyebrowse keybindings in evil normal mode map, rather than evil motion mode map work?

Not particularly confident about the differences between the evil maps, except that switching to a motion mode will show the eyebrowse key bindings again.

Also couldn't find a way to switch explicitly to motion state without calling `SPC SPC evil-motion-mode`.

jr0cket:
With hindsight I am closing this PR, as I dont really like changing keybindings if I can help it.

I will raise an issue to see if there is a better approach to managing workspaces now that tabs have been introduced into Emacs by default from Emacs 27 onward.

I will add a custom keybinding in dotspacemacs/user-config to cycle workspaces for now.

PR #13672, opened 2020-06-13
set initial evil state of ccls-tree-mode to be emacs

emacs18:
Both emacs and vim modes define j and k keys as down and up.

However emacs mode provides added feature of popping up matching source code buffer on left as one traverses within the tree mode buffer on right.

Thus it appears to me that emacs should be the default mode for almost everyone.

### DIFF
--- a/layers/+lang/c-c++/funcs.el
+++ b/layers/+lang/c-c++/funcs.el
@@ -160,7 +160,7 @@
      "mf" 'member-functions
      "mv" 'member-vars))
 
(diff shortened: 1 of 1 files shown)


smile13241324:
Hmmm strange I have noticed that this was disabled in d536b36d154708d195d903fe8538fb501c3c43b8 possibly because it is unusual to use emacs mode statically in a layer.

Given that `j` and `k` are probably bound I will merge this. If necessary we can later add a specific binding for this mode map.

smile13241324 (approved):

PR #13674, opened 2020-06-14
[keyboard-layout] Add ediff remapping

iv-nn:
Add `ediff` support to the `keyboard-layout` layer.

It works well using the bepo layout and should work with the other layouts but is untested with those.

`ediff-mode-map` is only defined once ediff is started, that's why I used `ediff-startup-hook` instead of `spacemacs|use-package-add-hook` or `with-eval-after-load` as I read in this issue: noctuid/general.el#59 but there may be another way. I'm new to emacs and spacemacs and this is my first time writing elisp so I'm not entirely sure.

See also this issue: https://github.com/emacs-evil/evil-collection/issues/196#issuecomment-419681661


iv-nn:
Done :)

smile13241324:

Issue #13681, opened 2020-06-18
--timed-requires broken on emacs 27 and emacs 28


yqu212:
If I start with emacs --timed-requires --debug-init, I will get 

> error: Recursive load, /usr/share/emacs/27.0.91/lisp/calendar/time-date.elc, /usr/share/emacs/27.0.91/lisp/emacs-lisp/debug.elc, /usr/share/emacs/27.0.91/lisp/calendar/time-date.elc, /usr/share/emacs/27.0.91/lisp/calendar/time-date.elc, /usr/share/emacs/27.0.91/lisp/calendar/time-date.elc, /usr/share/emacs/27.0.91/lisp/calendar/time-date.elc, /home/yqu212/.emacs.d/init.el

--timed-requires works well on emacs 26.

PR #13685, opened 2020-06-19
[core] fix broken --timed-requires and --adv-timers under emacs 27 and above

yqu212:
This fixs #13681.

time-date.el requires cl-lib under Emacs 27 and above. Then the around advice
spacemacs//timed-require and spacemacs//load-timer of require will cause
recursive load under emacs 27 and above. The load sequences is: requiring cl-lib
uses autoload function time-since which loads time-date.el and then time-date.el
requires cl-lib.

The solution to break this recursive load sequence is to load time-date.el
before adding the around advice spacemacs//timed-require and
spacemacs//load-timer to require.


smile13241324:

yqu212:
OK.

PR #13691, opened 2020-06-20
[terraform] Add LSP support for terraform-lsp

v2okimochi:
I added LSP support using lsp-mode at terraform layer.
We can use it through layer configuration:

```emacs-lisp
(terraform :variables
           terraform-backend 'lsp)
```

### DIFF
--- a/layers/+tools/terraform/packages.el
+++ b/layers/+tools/terraform/packages.el
@@ -12,8 +12,6 @@
 (setq terraform-packages
       '(
         (company-terraform :requires company)
-        (lsp-terraform :requires lsp-mode
-                       :toggle (eq terraform-backend 'lsp))
         terraform-mode
         ))
 
@@ -35,6 +33,8 @@
 (defun terraform/init-terraform-mode ()
   (use-package terraform-mode
     :defer t
+    :init (add-hook 'terraform-mode-hook
+                    'spacemacs//terraform-setup-backend)
     :config (when terraform-auto-format-on-save
               (add-hook 'terraform-mode-hook
                         'terraform-format-on-save-mode))))
--- a/layers/+tools/terraform/packages.el
+++ b/layers/+tools/terraform/packages.el
@@ -1,6 +1,6 @@
 ;;; packages.el --- terraform Layer packages File for Spacemacs
 ;;
-;; Copyright (c) 2012-2018 Sylvain Benner & Contributors
+;; Copyright (c) 2012-2020 Sylvain Benner & Contributors
 ;;
 ;; Author: Brian Hicks <brian@brianthicks.com>
 ;; URL: https://github.com/syl20bnr/spacemacs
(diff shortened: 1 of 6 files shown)


layers/+tools/terraform/packages.el
```diff
@@ -12,6 +12,8 @@
 (setq terraform-packages
       '(
         (company-terraform :requires company)
+        (lsp-terraform :requires lsp-mode
```
smile13241324, line 15: This package cannot be installed, instead it is part of `lsp-mode`.
Please remove it from here and instead modify `terraform/init-terraform-mode` to setup the mode hook to startup lsp.

See the following excerpt from the python layer:
```elisp
(defun python/init-python ()
  (use-package python
    :defer t
    :mode (("SConstruct\\'" . python-mode) ("SConscript\\'" . python-mode))
    :init
    (progn
      (spacemacs/register-repl 'python
                               'spacemacs/python-start-or-switch-repl "python")
      (spacemacs//bind-python-repl-keys)
      (spacemacs/add-to-hook 'python-mode-hook
                             '(spacemacs//python-setup-backend
                               spacemacs//python-default))
```

Please also make sure that the default case, lsp not bound, is working as in the other layers, normally the behaviour should be when the variable is `nil` lsp is used if lsp layer is loaded otherwise the non lsp solution should be loaded.
This can be overwritten by setting the backend in the dotfile. There is normally a getter function for this `(spacemacs//layer-backend)` which must be used to obtain the backend in the layer.

layers/+tools/terraform/packages.el
```diff
@@ -1,6 +1,6 @@
 ;;; packages.el --- terraform Layer packages File for Spacemacs
```
smile13241324, line 1: Please also add a layer dependency on lsp in layers.el when the backend is set to lsp, to automatically load lsp when someone has set the backend respectively. Again please refer to the python layer for a good example how to do this.

v2okimochi:
@smile13241324 Thanks for your reviews and excerpts. I modified code to fix issues. I will squash my commits after LGTM.

I intend that if `terraform-backend` is `'lsp` then setup lsp, else do nothing.

I would like your re-review (and your advice) because I may have made some mistakes. I don't understand well about some relation of process.
(for example, is it correct that I added dependancy `dap` at `layers.el` of terraform?)

Best regards,
v2okimochi

smile13241324:
Your PR looks ok so far. 

I have made some smaller changes nevertheless to streamline the code a bit and make it work the same as the other lsp supporting layers. One of the bigger changes is that I have reworked the package loading by moving the company setup code to a `post-init-company` function to inject the necessary setup code. This has to happen only once in the process.

As to requiring dap, in this case lsp would be a better choice, dap mode is like lsp++ and adds graphical debugging support, like you may be used from visual studio, but it requires additional capabilities from the lsp server.

If you are interested in the loading process of layers, which works like a framework in that it calls functions in a certain order and takes care to look for certain namings of functions, you should check `Tips on writing layers for Spacemacs`, directly in spacemacs via `spc h spc`.

You can also have a look at my commit if you want to see exactly what I did change 5dcf7ae8e.

smile13241324:

PR #13695, opened 2020-06-21
[haskell] Remove hayoo from Haskell layer keybindings

magthe:
The Hayoo site hasn't been available for quite a while now, and it doesn't seem to be coming back any time soon. There isn't much point in having a keybinding for it, I think.


smile13241324:

PR #13696, opened 2020-06-21
[haskell] Add keybinding for Hoogle that works with LSP

magthe:
The shortcut ", h h" gets bound to `lsp-describe-thing-at-point` by the LSP
layer. Having a shortcut for Hoogle is still very useful, so bind it to ", h g"
when LSP is activated.

*N.B.* This change doesn't yet contain any logic to limit the change to only when the LSP layer is active. I'd like to add that, but don't know how. I'm hoping some kind soul will let me know how to achieve that :)

### DIFF
--- a/layers/+lang/haskell/packages.el
+++ b/layers/+lang/haskell/packages.el
@@ -244,9 +244,6 @@
 
       (dolist (mode haskell-modes)
         (spacemacs/set-leader-keys-for-major-mode mode
-          "gi"  'haskell-navigate-imports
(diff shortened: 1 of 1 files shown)


magthe:
I hope I understand what you meant in your comment, @smile13241324 , and that my latest version comes close to what you intended.

I threw in `hH` on a whim, I think it makes some sense to have `hh`/`hH` (non-LSP) and `hg`/`hG` (LSP) for Hoogle lookups. Let me know what you think.

smile13241324:

Issue #13702, opened 2020-06-22
Shorten org file names when displaying org agenda on splash screen


greole:
When displaying the org todo and agenda list on the splash screen long file names tend to clutter the splash screen, especially if the org files are store in a folder like `~/Dropbox/gtd/project1/foo.org` etc. Hence it would be great to have an option to shorten the file names to `foo.org`. Ideally an option to keep long filenames could be implemented to distinguish  `~/Dropbox/gtd/project1/inbox.org` and   `~/Dropbox/gtd/project2/inbox.org` if needed.

smile13241324:
Merged your code and revised it a bit, please close this issue if it is ok @greole

greole:
Thank you for merging!

PR #13703, opened 2020-06-22
use file-name-nondirectory to shorten filename displayed on splash screen for org agenda

greole:
Use `file-name-nondirectory` to shorten filename to enhance readability #13702

### DIFF
--- a/core/core-spacemacs-buffer.el
+++ b/core/core-spacemacs-buffer.el
@@ -851,7 +851,7 @@ LIST: list of `org-agenda' entries in the todo list."
                            :button-suffix ""
                            :format "%[%t%]"
                            (format "%s %s %s"
-                                   (abbreviate-file-name
+                                   (file-name-nondirectory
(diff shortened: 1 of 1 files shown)


smile13241324:
I am not sure I like that change, the point is yes it does shorten the names, however in the current implementation it would possibly break the use case of having more than identically named files in multiple directories. I think to be able to merge this the relating function must be tweaked to detect if the abbreviation would result in duplicate file names and length the names consequently to have them unique again.

If I am mistaken and this is already happening feel free to set me right :).

greole:
I think some sort of automatic detection could be super confusing, but how about a flag/variable?

smile13241324:
Deal, lets add a new dotfile variable to control this.

greole:
Ok, i added the variable `spacemacs-org-agenda-display-filename-only` which is `nil` by default. If set to true only the filename is displayed and the path is omitted.

smile13241324:

PR #13711, opened 2020-06-24
Fix incf causing an error message

mosic:
This was the only place where `incf` was used instead of `cl-incf`, the PR fixes this error:

> Symbol’s function definition is void: incf


smile13241324:

PR #13714, opened 2020-06-25
[spacemacs-layouts] Fix return value of `rename-buffer`

bcc32:
`spacemacs//fixup-window-configs` is added as an `:around` advice for
`rename-buffer`.  It should preserve `rename-buffer`'s return value,
which is the new name of the buffer.

This pull request makes it do so.


smile13241324:

PR #13718, opened 2020-06-25
[SQL] Add lsp support

v2okimochi:
I added lsp support for [sqls](https://github.com/lighttiger2505/sqls) whiches configuration was added to lsp-mode. ( https://github.com/emacs-lsp/lsp-mode/commit/26eccfa561dc19e162a7504eefa61c83645bcaa2 )

If anyone have more better way to implement correct intends of the supporting than my code, I would like to agree it. I'm anxious about setting of lsp and company.

Thanks for great distribution Spacemacs.

### DIFF
--- a/layers/+lang/sql/funcs.el
+++ b/layers/+lang/sql/funcs.el
@@ -1,6 +1,6 @@
 ;;; funcs.el --- SQL Layer functions File for Spacemacs
 ;;
-;; Copyright (c) 2012-2018 Sylvain Benner & Contributors
+;; Copyright (c) 2012-2020 Sylvain Benner & Contributors
 ;;
 ;; Author: Sylvain Benner <sylvain.benner@gmail.com>
 ;; URL: https://github.com/syl20bnr/spacemacs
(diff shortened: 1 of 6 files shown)


layers/+lang/sql/funcs.el
```diff
@@ -16,3 +16,45 @@
...
+            :modes sql-mode))))
+
+(defun spacemacs//sql-setup-backend ()
+  "Conditionally setup sql backend."
+  (pcase (spacemacs//sql-backend)
+    (`lsp (spacemacs//sql-setup-lsp))))
+
+;; lsp
+
+(defun spacemacs//sql-setup-lsp ()
```
smile13241324, line 47: Do we need to require lsp-sqls directly? I would expect that this is loaded by lsp automatically.
Do we need to setup the sql variables for every buffer aka are they buffer local? If not I would prefer to move them to sql package init.

v2okimochi:
@smile13241324 Thanks for your reviews. I modified my codes.
I would like your re-review if there is any misunderstanding.

smile13241324:

PR #13724, opened 2020-06-26
[lsp] remove requirement for yasnippet

skebanga:
lsp-mode does not require yasnippet in order to function

from [the source](https://github.com/emacs-lsp/lsp-mode/blob/master/lsp-mode.el#L60):

```
(require 'yasnippet nil t)
```

This just removes the `:requires` clause from the lsp layer's packages list


smile13241324:

PR #13726, opened 2020-06-28
fix: removes nils from layers list

agzam:
Spacemacs allows to define layers declaratively, like so:

    `(,(when (eq system-type 'darwin) 'osx))

The problem - in Linux that would add a nil element into the list, which then
makes it unable to run `dotspacemacs/sync-configuration-layers <SPC f e R>`, the
tests won't pass.


duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/41276119eae0750d4bb1e8a3ac119da48e52fb3b

Issue #13503, opened 2020-04-18
Refactor SPC m a key bindings - running out of keys to use


practicalli-john:
As discussed on #13496 there are a large number of applications added to Spacemacs as layers and sensible mnemonic keybindings are hard to come by and there are some conflicts already.

The proposal is to group applications by their categories, based on the directory they are contained in under layers
- chat
- email
- fun 
- music  (already exists)
- readers
- tools
- security
- web-services

Here is what the new keybindings would be based on categories.  This shows the new keybindings, the path and name of the layer and the current keybindings
```
 ~SPC a c j~ +chat/jabber ~SPC a j~
 ~SPC a c i~ +chat/erc  ~SPC a i~  ---> No conflict: different sub level bindings
 ~SPC a c i~ +chat/rcirc  ~SPC a i~ --> No conflict: different sub level bindings
 ~SPC a c s~ +chat/slack ~SPC a C~
 ~SPC a e g~ +email/gnus  ~SPC a g~
 ~SPC a e m~ +email/mu4e ~SPC a M~
 ~SPC a e n~ +email/notmuch  ~SPC a N~
 ~SPC a f e~ +fun/emoji ~SPC a E~
 ~SPC a f g~ +fun/games ~SPC a G~
 ~SPC a f x~ +fun/xkcd ~SPC a x~
 ~SPC a m~ +music/pianobar ~SPC a m~ --> No conflict: different sub level bindings
 ~SPC a m~ +music/spotify ~SPC a m~  --> No conflict: different sub level bindings
 ~SPC a r z~ +readers/dash ~SPC a z~
 ~SPC a r n~ +readers/deft ~SPC a n~
 ~SPC a r f~ +readers/elfeed  ~SPC a f~
 ~SPC a r s~ +readers/speed-reading ~SPC a R~
 ~SPC a t b~ +tools/bm ~SPC a b~
 ~SPC a t c~ +tools/chrome ~SPC a F~
 ~SPC a t d~ +tools/docker ~SPC a D~
 ~SPC a t g~ +tools/geolocation  ~SPC a w~ --> Duplicate: move
 ~SPC a t g~ +tools/geolocation  ~SPC a W~ --> Duplicate: remove
 ~SPC a t j~ +lang/ipython-notebook (move layer to tools, rename to Jupiter Notebook)
 ~SPC a t k~ +tools/kubernetes ~SPC a K~
 ~SPC a t l~ +tools/command-log ~SPC a L~
[...]

PR #13728, opened 2020-06-28
[applications] refactor key bindings to layer categories

jr0cket:
refactor key bindings for applications to provide additional room for
applications and use lower case characters.

Move calc-dispatch to `SPC a *`

Relates to #13503

cc/
@smile13241324

### DIFF
--- a/layers/+readers/elfeed/README.org
+++ b/layers/+readers/elfeed/README.org
@@ -109,7 +109,7 @@ In this case you can disable these by setting =elfeed-enable-goodies= to =nil=.
 
 | Key binding | Description  |
 |-------------+--------------|
-| ~SPC a f~   | start elfeed |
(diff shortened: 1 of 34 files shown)


layers/+readers/elfeed/README.org
```diff
@@ -109,7 +109,7 @@ In this case you can disable these by setting =elfeed-enable-goodies= to =nil=.
 
 | Key binding | Description  |
 |-------------+--------------|
-| ~SPC a f~   | start elfeed |
+| ~SPC a r f~ | start elfeed |
```
smile13241324, line 112: I think this should be `SPC a r e` instead

jr0cket: Yes, good catch.  Fixed in latest push.

smile13241324:

PR #13737, opened 2020-07-01
Fix spacelpa public key file name

klao:
This fixes the issue that users have to import the public key manually
to pass spacelpa archive verification. See #12652


smile13241324:

@syl20bnr I hope this merge is ok, if not feel free to revert, I have not worked much yet with spacelpa so I could be wrong with this.

PR #13739, opened 2020-07-01
Use the standard elpa gpg homedir

### LINKED ISSUE #13738: Set gpg homedir for spacelpa archive verification
Adding spacemacs' public key to check spacelpa archives to the user's default gpg keyring is probably the wrong thing to do. Instead it should use the elpa keyring under `~/.emacs.d/elpa/gnupg`.

klao:
This PR makes it so spacemacs uses the same gpg keyring as elpa (under `~/.emacs.d/elpa/gnupg`), instead of using the user's default gpg keyring to verify spacelpa archives.

Closes #13738

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -2687,6 +2687,10 @@ continue with the stable ELPA repository installation."
           (error (setq verification-err
                        (format "GnuPGP doesn't seem to be available. %s"
                                (cdr error)))))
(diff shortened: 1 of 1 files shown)


klao:
I have updated this PR with a better snippet from `package`. It uses the appropriate `package` variable, not baking in the `~/.emacs.d/elpa/gnupg` path.

smile13241324:

PR #13740, opened 2020-07-01
[sql] fix regession by #13718

thanhvg:
remove `sql-mode` which is not a package move its init to init-sql

Otherwise spacemacs will complain about not able to install `sql-mode` package on startup.


duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/4928a9f55df0f3afe76e503fd0f858deadccd50e

PR #13745, opened 2020-07-03
Add ctp files to html language

AdeAttwood:
When editing Cakephp template files there is not highlighting. This adds the
`.ctp` files to the html language using web-mode.


smile13241324:

PR #13751, opened 2020-07-06
Allow multichar major mode leader key for exiting magit commit edits

tuh8888:
I use `DEL` as my major-mode leader key and found that the magit bindings during commit message edits weren't recognizing it because it was more than one character (like the default `,`. I think anyone who uses a major mode leader key like `DEL`, `RET` or even `SPC` will appreciate the consistency with org-capture, mu-compose, etc.

The fix just involves wrapping the `mm-key` variable in `kbd` to convert it to a form that can be concatenated and interpreted as expected by `evil-define-key`


duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/ce5f278a2aa283172c99176d1ed06d8e258c2d99

PR #13753, opened 2020-07-07
[agda] Change outdated agsy call

citrusmunch:
Use the current `agda2-autoOne` command to attempt to auto-prove
instead of the previous `agda2-auto`.

### DIFF
--- a/layers/+lang/agda/packages.el
+++ b/layers/+lang/agda/packages.el
@@ -78,7 +78,7 @@
           ","   'agda2-goal-and-context
           "="   'agda2-show-constraints
           "SPC" 'agda2-give
-          "a"   agda2-auto
+          "a"   'agda2-autoOne
           "c"   'agda2-make-case
           "d"   'agda2-infer-type-maybe-toplevel
           "e"   'agda2-show-context


duianto:
The `agda2-auto` command that's called in the Spacemacs `agda` layer is based on the agda version:
https://github.com/syl20bnr/spacemacs/blob/ce5f278a2aa283172c99176d1ed06d8e258c2d99/layers/%2Blang/agda/packages.el#L40-L42

Here's the commit that added the version check: https://github.com/syl20bnr/spacemacs/commit/f2a6872afb390f56096768315b760ab014271601

> Adapt agda2-auto call to agda 2.6.0 changes
> 
> * It's now agda2-auto-maybe-all. If you are not currently in a hole
> it tries to `auto` all holes by default.
> * Find and set an agda-version variable, to keep compatibility with
> previous agda versions.

So maybe it's working as expected?

Or did you encounter a case where it tries to call the old command `agda2-auto`?

citrusmunch:
Oops! You're right, that was really short sighted and hasty of me.
It works as expected, but I didn't see the `develop` branch already had the changes I wanted while testing on the main branch.

Thanks for the attention and sorry for the trouble! Will close the issue.

duianto:
It's no trouble at all, it's totally understandable since the variable that contains the version based command has the same name as the obsolete command.

PR #13755, opened 2020-07-07
[clojure] refactor namespace key bindings

jr0cket:
Add key bindings to refactor namespace forms for existing functions in clojure-mode

"ran" 'clojure-insert-ns-form
"raN" 'clojure-insert-ns-form-at-point
"rsn" 'clojure-sort-ns


jr0cket:
I had planned on adding `SPC m r u n` for the command `clojure-update-ns`, however, that function has an outstanding bug https://github.com/clojure-emacs/clojure-mode/issues/480 .  If that bug is fixed, I'll submit a new PR to add the key binding.

smile13241324:

PR #13758, opened 2020-07-09
[osx] Don't initialize osx-dictionary on non-mac systems

bcc32:
Restore the `SPC x w d` key binding for define-word-at-point, from
spacemacs-language.

### DIFF
--- a/layers/+os/osx/config.el
+++ b/layers/+os/osx/config.el
@@ -57,7 +57,7 @@
    Possible values are `super' `meta' `hyper' `alt' `left' `none'.
    Default: `left'.")
 
-(defvar osx-use-dictionary-app t
+(defvar osx-use-dictionary-app (spacemacs/system-is-mac)
   "Use the macOS dictionary app instead of Wordnet.")
 
 (defvar osx-swap-option-and-command nil


duianto:
#### Issue
With this PR applied and the `osx` layer enabled in Windows:
>`SPC x w d` is undefined


#### Possible cause
The variable `osx-use-dictionary-app` is set to `t` when the `osx` layer is enabled,
which causes this line to disable the `define-word` package/configs:
https://github.com/syl20bnr/spacemacs/blob/f3f0e3a98d431b38ca60e7c43bd1f2deb749e575/layers/%2Bspacemacs/spacemacs-language/packages.el#L12-L14

#### A possible solution

One solution might be to set the `osx-use-dictionary-app` variable:

https://github.com/syl20bnr/spacemacs/blob/f3f0e3a98d431b38ca60e7c43bd1f2deb749e575/layers/%2Bos/osx/config.el#L60

to: `(spacemacs/system-is-mac)`

```elisp
(defvar osx-use-dictionary-app (spacemacs/system-is-mac)
```

Then the current develop branch line that this PR changes, could remain unchanged:
```elisp
    :if osx-use-dictionary-app
```

bcc32:
Thanks, that workaround sounds good to me!

smile13241324:

PR #13764, opened 2020-07-12
Restore attribute and variable names in theming layer docs

milserk:
The names for the example attributes and the optional variables in the theming layer's documentation were lost in a previous commit. This commit restores those names.


smile13241324:

PR #13768, opened 2020-07-13
Remove gradle-mode to stop it breaking bindings of other major modes

### LINKED ISSUE #13750: Gradle bindings infect bindings of any subsequently used major-mode
#### Description :octocat:
After using Java mode, Gradle keybindings infect or even clobber keybindings of any subsequently used major-mode.

- Prerequisites: `java` layer installed, as well as one other layer (for this example, we will use `agda`, but this problem is not specific to the Agda layer)
- Start Emacs
- Open a Java file
- Open an Agda file
- Press `,` or `SPC m` to see the major mode menu

The critical Agda major mode keybinding of `l` as `agda2-load` has been replaced with a Gradle menu! (Not sure if this whether this is more frustrating, or more hilarious.) Again, though, Agda is only a particularly poignant example; it seems that at this point Gradle has permanently infected all major mode menus.

The Agda mode (or whatever major mode is pertinent) keybindings are not polluted by those from the Java layer.

#### System Info :computer:
- OS: darwin
- Emacs: 26.3
- Spacemacs: 0.300.0
- Spacemacs branch: keith (rev. 6729eb62a)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(agda auto-completion c-c++ emacs-lisp games git github graphviz
      helm html idris java javascript jsonnet markdown org osx
      (scala :variables scala-backend 'scala-metals scala-enable-eldoc t scala-auto-start-backend t)
      scheme slack spell-checking sql syntax-checking version-control yaml)
```
- System configuration features: NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS MODULES THREADS LCMS2

#### Backtrace :paw_prints:
No actual exception is surfaced.

Kazark:
After using Java mode, Gradle keybindings infect or even clobber keybindings of any subsequently used major-mode. For example, the critical Agda major mode keybinding of `l` as `agda2-load` gets with a Gradle menu!

The cause, found by @duianto, is that `emacs-gradle-mode` sets itself to a [global](https://github.com/jacobono/emacs-gradle-mode/blob/e4d665d5784ecda7ddfba015f07c69be3cfc45f2/gradle-mode.el#L176-L183) minor mode. The docs for `define-minor-mode` say about `:global:`:

> If non-nil specifies that the minor mode is not meant to be
> buffer-local, so don't make the variable MODE buffer-local.
> By default, the mode is buffer-local.

I don't know why `gradle-mode` is doing this; presumably there is some reason or need for it. But the author of that package hasn't been on GitHub since 2017, and the last update of the package itself is from early 2015. To seal the deal, `gradle-mode` hasn't been working anyway (at least, I've never got it to work). I've been running Gradle by launching a terminal within Emacs.

Fixes #13750.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2325,7 +2325,7 @@ Other:
   - Added support for multiple backends. Supported backends are: =meghanada=,
     =eclim=. The default backend is =meghanada=. (thanks to Ivan Yonchovski)
   - Added =ENSIME= jump handlers (thanks to Joao Azevedo)
-  - Improved =maven= and =gradle= support (thanks to Sylvain Benner)
+  - Improved =maven= (thanks to Sylvain Benner)
   - Made =java= layer depend on =groovy= layer (thanks to Sylvain Benner)
   - Added =org-babel= support (thanks to Michael Rohleder)
   - Remove broken =ENSIME= key bindings (thanks to Bjarke Vad Andersen)
@@ -2337,6 +2337,7 @@ Other:
   - Added /** */ smartparens pair (thanks to Ivan Yonchovski)
   - Added the prefix name "actionable" to ~SPC m a~ (thanks to duianto)
 - Fixes:
+  - Removed =gradle-mode= which is unsupported and was breaking other major modes
   - Replace usage of =ensime-print-type-at-point= by =ensime-type-at-point=
     (thanks to Joao Azevedo)
   - Added defalias for =ensime-type-at-point= for compatibility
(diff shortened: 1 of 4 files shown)


Kazark:
@smile13241324 @duianto ℹ️

Kazark:
It's making me log in to CircleCI to GitHub authentication, but that is asking for too many permissions (in particular, it wants to be able to see the org from my employer, which I do not have the right to give it access to). How do I find out what I broke on all these Emacs versions?

smile13241324:
Nevermind there were some issues in CI lately, It has been fixed already but is not yet on your branch. For the small change you did it should not affect the results.

smile13241324:

PR #13770, opened 2020-07-13
[spell-checking] Add keybinding for flyspell-region

be11ng:
`flyspell-region` should have its own keybinding for times when just a comment
needs to be spellchecked and the like.


duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/39de9ff0af3d1e52c045d739a96a83c1718bce09

PR #13777, opened 2020-07-16
visual: apply ansi color only on compilation major mode

loysollivier:
Some modes have compilation-mode as parent mode but the ansi coloring
scheme is breaking their highlighting.
i.e. https://github.com/Wilfred/ag.el/issues/124

Fix this by applying ansi-colors only when the major mode is
compilation-mode.


smile13241324:

PR #13783, opened 2020-07-17
Change browse-at-remote to "SPC go" sicne it is not github specific

sdwolfz:
Recently discovered `browse-at-remote` and it seems to not be tied to github, it handles many other websites, so this change makes the key binding simpler.


smile13241324:

Issue #13763, opened 2020-07-12
vim migration guide: what's the way to convert keybindings which result in multiple actions?


peey:
The [migration guide](https://github.com/syl20bnr/spacemacs/blob/master/doc/VIMUSERS.org#the-spacemacs-file) mentions how you can map a key to a function:

```
(define-key evil-normal-state-map (kbd "j") 'evil-next-visual-line)
```

But how can I map multi-action keybindings? e.g.

```
" https://vim.fandom.com/wiki/Insert_newline_without_entering_insert_mode
nmap <S-Enter> O<Esc>j
```

Here we map `Shift-Enter` to a 3 actions: `O` then `Esc` then `j`

duianto:
A macro can be bound like this:
```elisp
(define-key evil-normal-state-map (kbd "S-<return>") (kbd "O <escape> j"))
```
source: https://emacs.stackexchange.com/questions/22031/configuring-multiple-key-bindings

That macro causes the cursor to move to the beginning of the current line.

If you want to keep the cursor in the same column, then you can use an existing command that comes with Spacemacs:
```elisp
(define-key evil-normal-state-map (kbd "S-<return>") 'evil-unimpaired/insert-space-above)
```

There's also a command to add a space below: `evil-unimpaired/insert-space-below`

Spacemacs binds them by default to:
- `[ SPC` calls `evil-unimpaired/insert-space-above`
- `] SPC` calls `evil-unimpaired/insert-space-below`


If you want to unbind a key binding, then you can bind it to `nil`.
```elisp
(define-key evil-normal-state-map (kbd "S-<return>") nil)
```

peey:
Thanks! This solves my problem.

I'm wondering if this information should be added to the migration guide. Feel free to keep the issue open if you think so, otherwise you may close it.

duianto:
A PR is pending with a documentation update:
[doc] Document how to bind keys to a macro (one or more keys) #13785

PR #13785, opened 2020-07-18
[doc] Document how to bind keys to a macro (one or more keys)

duianto:
Related discussion:
vim migration guide: what's the way to convert keybindings which result in multiple actions? #13763


smile13241324:

PR #13788, opened 2020-07-19
Improve slack layer keybindings

wang-d:
1. Make slack-mode bindings available in slack-thread-message-buffer-mode
2. Bind : to slack-insert-emoji

Let me know if you would prefer I split these into separate patches.


smile13241324:

PR #13790, opened 2020-07-22
[layers/+lang/javascript] Enriched README with linter configuration details

Martinsos:
Hi!
I was just setting up Spacemacs to work with StandardJS, and I spent quite some time until I figured out that I need to turn off js2-mode's warnings -> I was constantly thinking that flycheck is using the wrong linter or that something else weird is happening.

Therefore, I created this PR where I added a note about turning of js2-mode warnings in case you are using custom linter via flycheck.

I also added information about StandardJS next to ESLint and JSHint since it is so popular, and finally I ended up reorganizing a little bit (creating special section for linting, moving part about installing importJS to section about importJS).

I hope this PR helps!

### DIFF
--- a/layers/+lang/javascript/README.org
+++ b/layers/+lang/javascript/README.org
@@ -8,6 +8,7 @@
 - [[#description][Description]]
   - [[#features][Features:]]
 - [[#install][Install]]
+  - [[#error-checking--linting][Error checking / linting]]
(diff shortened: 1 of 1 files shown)


duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/684976c993a2e18a1a55b5551795c19a700f10d9

PR #13802, opened 2020-07-30
[layers/+lang/haskell] Updated README with more detail about Dante + misc

Martinsos:
First of all thanks for this awesome layer and docs :)!

I was just recently revisiting haskell layer because I wanted to improve my haskell experience in spacemacs, and I found that some things in README maybe could be better explained (especially comparison of backends) so I am creating this PR!

I also have few additional questions:
1. Why are we calling these backends "completion backends"? Most of them are doing more than just completion, they also do syntax checking, navigation, auto-fixing of errors. Is that leftover from before?
2. Why is Dante not a default backend? I get that company-ghci was chosen as default because it is lightweight and was guaranteed to work and Dante was probably not in the game yet then, but I believe Dante offers the same guarantees or working out of the box while it also gives much more and better features. I would say currently Dante and HIE (HSL) are the best options, with Dante being logical default choice since there is no setting up to do.
3. Should we remove Intero? It is not actively maintained any more.
4. We could also remove company-ghci I think, it seems to me it can't really compete with other backends?
5. While README is really detailed and offers a lot of info on a lot of options, what I found difficult when starting with it is deciding what to use. There are many backends, flycheck and flymake, and many terms, I got confused with all the information relatively quickly. I am wondering, could we reshape README to start with kind of TLDR section that describes in simple worlds how to simply set up the layer in a way that works for 90% cases, and then the rest of README can cover all the other options and alternatives? Kind of like opinionated quick start guide, and then comes the real stuff for those that want to know more. I imagine if Dante would be default backend, this quick start guide would be pretty simple, but at least it would be clear that hey, this is enough to get stuff working, you don't have to read everything else if you just want to get started.

I would be happy helping creating additional PRs if any of the questions above results with the agreement to try doing something in those directions.

### DIFF
--- a/layers/+lang/haskell/README.org
+++ b/layers/+lang/haskell/README.org
@@ -12,7 +12,7 @@
   - [[#dependencies][Dependencies]]
   - [[#setup-path][Setup PATH]]
   - [[#choosing-a-backend][Choosing a backend]]
-    - [[#company-ghci][=company-ghci=]]
+    - [[#ghci-aka-company-ghci][=ghci= (aka =company-ghci)=]]
     - [[#intero][=intero=]]
     - [[#dante][=dante=]]
     - [[#ghc-mod][=ghc-mod=]]
@@ -61,8 +61,8 @@ This layer adds support for the [[https://www.haskell.org/][Haskell]] language.
 
 ** Features:
 - syntax highlighting for [[https://github.com/haskell/haskell-mode][haskell source]], [[https://github.com/haskell/haskell-mode][cabal files]], [[https://github.com/bgamari/cmm-mode][C-- source]],
-- auto-completion with one of the selected backends (=intero=, =dante=, =ghci= or
-  =ghc-mod=).
+- auto-completion with one of the selected backends (=intero=, =dante=, =ghci=,
+  =ghc-mod= or =lsp=).
 
 *This layer is under construction, it needs your contributions and bug reports.*
 
@@ -79,8 +79,8 @@ This layer requires some [[https://www.haskell.org/cabal/][cabal]] packages:
 - =stylish-haskell= (optional for =haskell-mode=)
 - =hasktags= (optional)
 - =hoogle= (optional for =haskell-mode= and =helm-hoogle=)
-- =ghc-mod= (optional for completion)
-- =intero= (optional for completion)
+- =ghc-mod= (optional for completion if using =ghc-mod= as completion backend)
+- =intero= (optional for completion if using =intero= as completion backend)
 
 To install them, use the following command (or the =stack= equivalent):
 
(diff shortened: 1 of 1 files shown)


layers/+lang/haskell/README.org
```diff
@@ -373,26 +381,26 @@ This top-level command is prefixed by ~SPC m~:
 | ~SPC m g b~ | return from definition |
 
 * Syntax checking
-At the moment there are four components, which can check the syntax and indicate
+At the moment there are four components, which can indicate
 errors and warnings in the code. Those components are:
-- flycheck
+- dante, intero (via flycheck)
```
Martinsos, line 386: I am not sure about this line, does this make sense what I wrote here? Flycheck is frontend for syntax checker, and syntax checkers are dante or intero. But maybe somebody else also? Is it wrong to phrase it this way?

smile13241324: Nope perfectly fine this way :+1:

layers/+lang/haskell/README.org
```diff
@@ -104,20 +104,26 @@ For information about setting up =$PATH=, check out the corresponding section in
 the FAQ (~SPC h SPC $PATH RET~).
 
 ** Choosing a backend
-To choose a default backend set the layer variable =haskell-completion-backend=:
+To choose a backend set the layer variable =haskell-completion-backend=:
 
-#+BEGIN_SRC elisp
-  (haskell :variables haskell-completion-backend 'ghci)
+#+BEGIN_SRC emacs-lisp
+  (setq-default dotspacemacs-configuration-layers
```
smile13241324, line 110: This line is wrong, the layer declaration for haskell is not part of the auto-completion declaration. Here I think the old version is more clear.

layers/+lang/haskell/README.org
```diff
@@ -104,20 +104,26 @@ For information about setting up =$PATH=, check out the corresponding section in
...
 Note that in order to enable completion, you'll have to enable the
 =auto-completion= layer as well.
 
-Alternatively the =lsp= backend will be automatically chosen if the layer =lsp=
-is used and you did not specify any value for =haskell-completion-backend=.
+Supported values for =haskell-completion-backend= are =intero=, =dante=,
+=ghci=, =ghc-mod=, and =lsp=.
+
+If you haven't specified any value for =haskell-completion-backend=,
+=ghci= (aka =company-ghci=) will used as default backend, unless
```
smile13241324, line 122: Small typo, 

```
...If you haven't specified any value for =haskell-completion-backend=,
=ghci= (aka =company-ghci=) will be used as default backend, unless...
```

smile13241324:
Here are some answers to your additional questions:
1. => This is historic. Other layers shall name these backends.
2. => I think this is also historic. Lets fix that, could you make a PR to define "dante" as the default backend? I think it doesn't make sense to use an outdated backend as default still.
3. => If it is not longer maintained I think the right thing to do would be to remove it from the layer. Care to make a PR?
4. => Same for company-ghci
5. => Absolutely, I think the Haskell layer was always a bit confusing also there is this statement of a work in progress which should also be removed. Ok basically the correct document structure would be to have an "install" section which just shows how the layer is declared, followed by a "Configuration" section of which the first choice is "Choosing a backend" which defines a primary backend which should be "dante" and alternative backends with additional setup procedures. It should give easy to understand information when to choose which. A good example would be the "groovy" layer.

If there are configurations which are independent of the chosen backend it should follow on the same level as the "Choosing a backend" headline. For LSP if there are multiple servers to choose from the differences should be described below the "LSP" headline which is below the "Choosing a backend" headline.

layers/+lang/haskell/README.org
Martinsos: I was also confused with this at first, but I believe it is not wrong -> we are just listing layers in dotspacemacs-configuration-layers, first we list auto-completion, then haskell, and haskell has variables. So auto-completion and haskell are here separate elements of the list, haskell is not part of auto-completion declaration. Although, visually it can be deceiving.

I actually copy-pasted this from the old code, line 138, and just changed `intero` into `dante`, so I didn't really change much, I just moved it.

There is sentence below explaining that auto-completion layer is needed for completion to work, this is from old code and I left as is, and I believe that is why auto-completion was listed in the code example.

I do however understand the confusion here hm! What I did not is move the sentence about `auto-completion` layer being required to the top of the section, since it is a requirement and we don't want somebody to miss it, and then we continue with the simple code example that shows how to set haskell backend. Let me know if this seems better.

smile13241324:
Yep lets follow this plan. Creating an issue for it will not hurt maybe we find someone with good reasons to keep them.

layers/+lang/haskell/README.org
smile13241324: Yep sounds better. The point is that we should never demand syntax checking or auto completion layer this should be a pure user choice. The hooks make sure that the respective setup code is skipped.

Having a hint in the Readme that these are required for certain features does not hurt.The documentation however already mentiones their plugin nature so I would expect that most users are aware of it.

smile13241324 (approved):
Looks also good

Martinsos:
Great, I opened the issues (#13869 , #13868 and #13867) and will continue from those!

PR #13803, opened 2020-07-30
[layers/+lang/haskell] Prolong flycheck popup duration in Haskell layer.

Martinsos:
While recently using haskell layer, I realized that flycheck popups dissapear after 5 seconds, which was making my development really hard -> error messages in haskell are usually pretty complex and lenghty, and trying to read them in 5 seconds is in most cases impossible, so I would repeatedly have to move and return my cursor on the position of error to see the message again and again.

After some digging and help from people on gitter, I learned that by default flycheck shows popup for 5 seconds (this default is coming from flycheck-pos-tip), and that duration of popup can be modified with `flycheck-pos-tip-timeout` variable. I put it to 20 seconds since I believe that is fairly long, we could even put it to something longer. I am not even sure if this popup timeout is useful at all to be honest.

This is my first commit and I am not very proficient with elisp so I put it is most obvious place and I can confirm that it works, however I don't know if there is a better location to put this command, I am hoping you will tell me!

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -3314,6 +3314,7 @@ Other:
 - Fixed =goto-flycheck-error-list= (thanks to Thanh Vuong)
 - Added lsp support for sqls
   - Added layer variable =sql-lsp-sqls-workspace-config-path= to setting workspace configuration
+- Added layer variable =syntax-checking-auto-hide-tooltips= (thanks to Martin Sosic)
 **** Swift
 - Update Swift REPL key bindings (thanks to Elliot Bulmer)
 **** Systemd
--- a/layers/+checkers/syntax-checking/README.org
+++ b/layers/+checkers/syntax-checking/README.org
@@ -14,6 +14,7 @@
   - [[#enable-flycheck-globally][Enable flycheck globally]]
   - [[#enable-support-for-traditional-error-navigation][Enable support for traditional error navigation]]
   - [[#bitmaps][Bitmaps]]
+  - [[#auto-hide-tooltips][Auto hide tooltips]]
 - [[#key-bindings][Key bindings]]
 
 * Description
@@ -91,6 +92,19 @@ variable =syntax-checking-use-original-bitmaps= to =t=:
 #+BEGIN_SRC emacs-lisp
   (setq-default dotspacemacs-configuration-layers
     '((syntax-checking :variables syntax-checking-use-original-bitmaps t)))
+#+END_SRC
+
+
+** Auto hide tooltips
+You can set time in seconds after which tooltips are automatically hidden by setting
+the variable =syntax-checking-auto-hide-tooltips= to a positive number of seconds.
+
+Default value of =syntax-checking-auto-hide-tooltips= is =nil=, and in that case tooltips
+are kept open until the cursor is moved.
+
+#+BEGIN_SRC emacs-lisp
+  (setq-default dotspacemacs-configuration-layers
+    '((syntax-checking :variables syntax-checking-auto-hide-tooltips 10)))
 #+END_SRC
 
 * Key bindings
--- a/layers/+checkers/syntax-checking/config.el
+++ b/layers/+checkers/syntax-checking/config.el
@@ -14,6 +14,9 @@
 (defvar syntax-checking-enable-tooltips t
   "If non nil some feedback are displayed in tooltips.")
 
+(defvar syntax-checking-auto-hide-tooltips nil
+  "If non-nil and positive number, auto hide tooltips after number of seconds.")
+
 (defvar syntax-checking-enable-by-default t
   "Enable syntax-checking by default.")
 
--- a/layers/+checkers/syntax-checking/packages.el
+++ b/layers/+checkers/syntax-checking/packages.el
@@ -102,7 +102,8 @@
     :defer t
     :init
     (with-eval-after-load 'flycheck
-      (flycheck-pos-tip-mode))))
+      (flycheck-pos-tip-mode)
+      (setq flycheck-pos-tip-timeout (or syntax-checking-auto-hide-tooltips 0)))))
 
 (defun syntax-checking/pre-init-popwin ()
   (spacemacs|use-package-add-hook popwin


duianto:
Setting `(setq flycheck-pos-tip-timeout 20)` like that, will probably change it in every mode.

Let's make a `syntax-checking` layer variable instead.

Remove that variable assignment `(setq flycheck-pos-tip-timeout 20)`

Add these lines:
```elisp
(defvar syntax-checking-auto-hide-tooltips 5
  "Seconds until the tooltip auto hides (default 5).
If nil, keep tooltips open until the cursor is moved.")
```

before this `syntax-checking-enable-tooltips` variable declaration:
https://github.com/syl20bnr/spacemacs/blob/a76bda6a54a79d8e2ab9daddf79eb176a3179e8c/layers/%2Bcheckers/syntax-checking/config.el#L14

Then in the `syntax-checking/init-flycheck-pos-tip` function:
https://github.com/syl20bnr/spacemacs/blob/a76bda6a54a79d8e2ab9daddf79eb176a3179e8c/layers/%2Bcheckers/syntax-checking/packages.el#L99

after the `flycheck-pos-tip-mode` activation.

Add this variable assignment:
```elisp
(setq flycheck-pos-tip-timeout (or syntax-checking-auto-hide-tooltips -1))
```

The function ends up looking like this:
```elisp
(defun syntax-checking/init-flycheck-pos-tip ()
  (use-package flycheck-pos-tip
    :if syntax-checking-enable-tooltips
    :defer t
    :init
      (flycheck-pos-tip-mode)
      (setq flycheck-pos-tip-timeout (or syntax-checking-auto-hide-tooltips -1)))))
```

That will make it possible to change the tooltip duration by adding the new variable to the
`dotspacemacs-configuration-layers` list near the top of the `.spacemacs` file, like this:

Keep the tooltip open until the cursor is moved:
```elisp
     (syntax-checking :variables syntax-checking-auto-hide-tooltips nil)
```
Or set it to a specific number of seconds:
```elisp
     (syntax-checking :variables syntax-checking-auto-hide-tooltips 20)
```

duianto:
Also document the new layer variable in `syntax-checking/readme.org`

After the `Enabling/Disabling tooltips` section
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+checkers/syntax-checking/README.org#enablingdisabling-tooltips

Add this or something more descriptive:
```org
** Auto hide tooltips
By default tooltips are hidden after 5 seconds.

The auto hide time can be changed by setting the variable
=syntax-checking-auto-hide-tooltips= to a number.

Or to keep the tooltips open until the cursor is moved, set the variable to =nil=.

#+BEGIN_SRC emacs-lisp
    '((syntax-checking :variables syntax-checking-auto-hide-tooltips nil)))
#+END_SRC
```

smile13241324:
Any news here, @Martinsos? Do you wish to try to implement duiantos idea, or should I take over that change?

Martinsos:
Sorry for not replying sooner! I would like to implement this and other
issues/PRs I opened, I was slow with responses because I am on summer
vacation, but I will be back in couple of days and will get to it then!

Martinsos:
@duianto and @smile13241324 thank you for very detailed instructions, since I am beginner with spacemacs they helped a lot and helped me learn more about how spacemacs works.

I implemented your suggestions, with only difference being using `0` instead of `-1` for `flycheck-pos-tip-timeout` when we want tooltips to not auto hide, since in `pos-tip` documentation they said any non-positive number works, and I felt 0 is more meaningful for this purpose than -1.

I believe this is it, but I do still have questions:
1. While this enables spacemacs users to easily set the timeout of tooltips, it still does not solve the problem I originally wanted to solve, which was that when using haskell layer to write haskell code, default timeout of tooltips (which is 5 seconds) makes no sense. So while user can now easily modify that, I still think it would be great if haskell layer took care of that immediately instead of everybody having to figure out on their own that they should configure it themselves. Is there a way in haskell layer to configure `syntax-checking-auto-hide-tooltips` just for itself, without affecting other language layers?
2. In general, why do we have 5 seconds as default value? Why do we even need the timeout, in practice? I can't see the purpose of it. Should be set it by default to 0 or at least some bigger value (10, 20 seconds) and users can configure it if needed? Maybe we could run some kind of poll to see if people find timeout useful?

smile13241324 (approved):
Code looks good so far, just the default I would change to nil so that the popups do not longer autohide

smile13241324:
@Martinsos while I agree that a default timeout should be set to 0, as I would also expect a tooltip not to disappear automatically, this is still unrelated to the language used. In spacemacs all auto complete and syntax checking decisions are centralised so that these behave consistent in all languages, therefore this setting cannot be set on a layer basis.

@duianto do you see any reasons why not to set the default to "0" hence let the popup not disappear automatically?

smile13241324:
Btw @Martinsos can you add your change to the changelog.develop please, we hope to use this for documenting the main changes when we finally have another release.

duianto:
>why do we have 5 seconds as default value?

I just used the default value that's defined upstream in flycheck-pos-tip:
https://github.com/flycheck/flycheck-pos-tip/blob/dc57beac0e59669926ad720c7af38b27c3a30467/flycheck-pos-tip.el#L56

---

>do you see any reasons why not to set the default to "0" hence let the popup not disappear automatically?

There seems to be some reports requesting both behaviors:
- Tool-tip not disappearing https://github.com/flycheck/flycheck-pos-tip/issues/6
How to set a timeout? https://github.com/flycheck/flycheck-pos-tip/issues/9
- Disable tooltip timeout by default https://github.com/flycheck/flycheck-pos-tip/issues/30
flycheck tooltip disappears https://github.com/chris-martin/home/issues/44

I don't have a preference either way.

But since you mentioned that lsp mode doesn't have a timeout for the popups, then it might be more consistent if both types of tooltips (lsp and flycheck) behaved the same.

smile13241324:
Yep, I have not yet noticed that, you are right yes LSP has no popup hence there is no timeout the information is projected into the right side of the screen as long as your cursor is on the warning. So lets set the default to no popup, then the people who wish to have it autohide can set it accordingly, and the others who do not want to see any popup and rather see it in the mini buffer, eldoc style, can disable tool-tips all together.

Martinsos:
@smile13241324 and @duianto sounds good, I am also in favor of not auto-hiding the tooltips by default. I added commit where I set `syntax-checking-auto-hide-tooltips` to `nil` by default.

@smile13241324 regarding changelog.develop, I already added the change there, is that it?

duianto:
The readme needs to be updated to reflect that the tooltips aren't hidden automatically by default.

And can you squash the commits, the recommendation is to limit PRs to one per PR, unless the changes are complex.

Martinsos:
@duianto ah I forgot about the README! Ok updated, and I squashed the commits! I normally squash commits when working on my code, but I do it when PR is done so I don't mess up with the comments on github because it can happen they become outdated otherwise and it becomes harder to track the conversation.

duianto:
I'm observing the issue described here:
Tool-tip not disappearing https://github.com/flycheck/flycheck-pos-tip/issues/6

The tooltip is displayed on top of all applications when switching to an application that's on top of Emacs.
- Move the cursor to an error so that the tooltip appears

emacs_2020-08-25_22-40-32

- Switch to another application (Alt + Tab)
the tooltip remains on top of the application that one switched to

chrome_2020-08-25_22-40-33


That issues was fixed for the issue author when it was reported in 2015.

Maybe something's changed since then.

Martinsos:
@duianto true, I can also replicate this! Never noticed it before.
I don't think we should change behavior to cover for this bug though, if somebody has this problem they can set the timeout themselves, if that helps.

PR #13804, opened 2020-07-30
[layers/+lang/haskell] Make flycheck run hlint checker after dante checker.

Martinsos:
Currently, when dante is used as backend, hlint is not used at all in flycheck.

Therefore, in this PR I added hlint checker as next checker after dante checker.

I have to admit I wasn't sure at all where to put the code in question. I can confirm that it works currently, but I am considering this first commit more of a first try and I hoping somebody will me guide better where to put it exactly and how to implement this properly.

### DIFF
--- a/layers/+lang/haskell/packages.el
+++ b/layers/+lang/haskell/packages.el
@@ -138,6 +138,8 @@
     :defer t
     :config
     (progn
+      (add-hook 'dante-mode-hook
+                (lambda () (flycheck-add-next-checker 'haskell-dante '(info . haskell-hlint))))
       (dolist (mode haskell-modes)
         (spacemacs/set-leader-keys-for-major-mode mode
           "gb" 'xref-pop-marker-stack
--- a/layers/+lang/haskell/packages.el
+++ b/layers/+lang/haskell/packages.el
@@ -138,8 +138,6 @@
     :defer t
     :config
     (progn
-      (add-hook 'dante-mode-hook
-                (lambda () (flycheck-add-next-checker 'haskell-dante '(info . haskell-hlint))))
       (dolist (mode haskell-modes)
         (spacemacs/set-leader-keys-for-major-mode mode
           "gb" 'xref-pop-marker-stack
@@ -162,7 +160,11 @@
       (spacemacs/set-leader-keys-for-major-mode mode "hf" 'helm-hoogle))))
 
 (defun haskell/post-init-flycheck ()
-  (spacemacs/enable-flycheck 'haskell-mode))
+  (progn
+    (add-hook 'dante-mode-hook
+              (lambda () (flycheck-add-next-checker 'haskell-dante '(warning . haskell-hlint))))
+    (spacemacs/enable-flycheck 'haskell-mode))
+  )
 
 (defun haskell/init-flycheck-haskell ()
   (use-package flycheck-haskell


layers/+lang/haskell/packages.el
```diff
@@ -138,6 +138,8 @@
     :defer t
     :config
     (progn
+      (add-hook 'dante-mode-hook
```
smile13241324, line 141: I think this is not the right spot, you are using functions from flycheck though it could be that flycheck is not installed at all. The only safe place for this is in the init function of a flycheck specific extension or a post-init-flycheck block.

Here is an example from the go layer:
```elisp
(defun go/post-init-flycheck ()
   ;;; Do your specific activation code, runs once during init
  (spacemacs/enable-flycheck 'go-mode) )
```
Btw, I have noticed you would call haskell-hlint only if only "info" level messages from the "haskell-dante" are available. Is that what you intended? I would expect that hlint is called when the error levels drop to "warnings" instead.

Martinsos: Hm, unfortunately I don't have much experience with elisp, nor with how elisp code is organized, so I find it difficult to figure out how to organize things correctly. But I hope this is an opportunity to learn more.

Ok, I moved the code into haskell/post-init-flycheck for now. I guess other location could have been init-flycheck-haskell? How do you decide where to put this, what is the difference?

This is probably caused by my inexperience with elisp, but I can't figure out where are haskell/post-init-flycheck and haskell/init-flycheck-haskell used? I tried searching project-wide for them but I couldn't find any usage. On the other hand I noticed other layers have similar functions. So I am guessing they are defined by some naming convention and then some other part of spacemacs pick them up? When in situation like this, how do you go about figuring this stuff out, should I be looking into spacemacs docs, flycheck docs, where do I search?

Regarding "info" -> makes sense, it also seems more reasonable to me to use "warning" -> warnings means it could be ignored, and we want linting to work even if there is warning from compiler that we are ignoring. I used "info" because I copied that line of code from Dante repo README https://github.com/jyp/dante. I wonder why they used it though.

smile13241324: As you already have guessed these post-*-config functions are handlers called during the loading process. These are only called if the respective layer is loaded.

The init function for flycheck just registeres a new checker the layer has still to activate flycheck for its modes if it makes sense.

Here is a handy document describing the loading process:
https://github.com/syl20bnr/spacemacs/blob/develop/doc/LAYERS.org

As to why you cannot find them this is due to the generic nature of lisp these functions are called dynamically during the loading process in the spacemacs core.

smile13241324 (approved):
Looks ok so far

layers/+lang/haskell/packages.el
Martinsos: Awesome, I see everything is documented here! I will make sure to check this before proceeding, so I have better understanding of how layers work. Thanks!

Martinsos:
Reading more about layers in spacemacs, I read that each function can only assume the existence of one package (mentioned [here](https://github.com/syl20bnr/spacemacs/blob/develop/doc/LAYERS.org#localize-your-configuration)).

However, we are doing something with `dante` in `post-init-flycheck` function, so now it cares both about dante and flycheck -> about two packages. Maybe we should have used smth like `configuration-layer/package-usedp` to check that `dante` is used? Or we don't care because add-hook will not fail if dante is missing (or maybe it will fail I am not sure)?

smile13241324:
Hmmmm basically I would expect it to not fail for we always install "dante" in that layer so the hook should be defined. I am not sure if `add-hook` will fail if dante is excluded, if this is the case we should check whether dante is installed with `configuration-layer/package-usedp`.

Normally the post-*-package mechanism should protect us from this, but it fails when multiple packages are needed, unfortunatelly we have that situation often i.e. json layer assumes that json-mode is available though users may have it manually excluded. I normally tend to not check if the package is unconditionally excluded, however in this case checking would be sensible if add-hook leads to an error message that is.

Martinsos:
Got it, I read that users can manually exclude a certain package, so in
that case it would make sense that we check for it. However, from what I
also read, add-hook should not fail (instead it should create hook even if
it does not exist), so I don't think in this case we have to do anything.
But I was interested in this case where we have multiple packages, which
you explained now. Ok, makes sense, I would leave it as it is then!

smile13241324:
Yeah emacs and spacemacs in particular is a complicated beast sometimes :).

PR #13806, opened 2020-07-31
Add more slack keybindings

wang-d:
Two patches:

1. alphabetize existing keybindings
2. add keybindings for `slack-all-threads` and `slack-all-unreads`

I tested by trying each of the keybindings.

Note that `slack-all-unreads` is buggy for me, which might be an argument for not publicizing it with toplevel keybindings.  But it mostly works, and the bug is unrelated to the keybindings (I can reproduce via `M-x`). I'd like to bind the function now and hope for a fix from upstream later.


wang-d:
Fixed.

I did the alphabetization in the same commit now that I know it's not a controversial change.  Let me know if this whole PR is small enough to be squashed, despite having a few different changes in it.

smile13241324:
thanks @wang-d would be enough to be squashed, however I can do this via the github web interface now so no need for further actions from your side.

Issue #13596, opened 2020-05-18
ess layer: defvaralias obsolete var produces a warning with emacs 27


pietrodito:
#### Description :octocat:
With emacs 27 warning about `ess-continued-statement-offset` when opening code file. I cannot reproduce this warning with vanilla emacs and ess package only installed.

#### Reproduction guide :beetle:
- Start Emacs
- Open `.R` file with code which needs to be indented. 

*Observed behaviour:* :eyes: :broken_heart:
Warning (defvaralias): Overwriting value of ‘ess-continued-statement-offset’ by aliasing to ‘ess-offset-continued’

*Expected behaviour:* :heart: :smile:
No warnings

#### System Info :computer:
- OS: gnu/linux
- Emacs: 27.0.91
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. bf9d9d196)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: helm
- Layers:
```elisp
(helm emacs-lisp ess)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG SOUND DBUS GSETTINGS GLIB NOTIFY INOTIFY GNUTLS LIBXML2 FREETYPE HARFBUZZ M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XDBE XIM MODULES THREADS PDUMPER GMP

#### Backtrace :paw_prints:
```
```

beyondpie:
Hi, I guest the problem from the line below.

https://github.com/beyondpie/spacemacs/blob/920f568d819d813956a07a72328ae18df6ce95c6/layers/%2Blang/ess/packages.el#L75

And I just comment `ess-continued-statement-offset`, which should be OK then.

smile13241324:
Merged, please have another look @pietrodito.

PR #13826, opened 2020-08-08
[ESS] Fix: remove obsolete variables in ESS

beyondpie:
- use ess-offset-continued to replace the obsolete variables (from ESS 15.04):
  ess-first-continued-statement-offset and ess-continued-statement-offset

- remove ess-expression-offset: no such variable now


beyondpie:
#13596

PR #13833, opened 2020-08-11
Better UX for opening and copying links

sdwolfz:
Not really sure if I should keep `xA` and `xm`. Those key bindings might be more useful for something else.


smile13241324:
I would not worry about the `xA` and `xm` bindings, we can still change them if we need them for something else.

kepi:
Can I ask why this is better UX and why it was merged?

I belive changing bindings which many of us remembered is always bad and shouldn't be done if it can be avoided. I'm talking especially about `x o` which is now not only behaving differently than before but to open links I now have to use one more key (shift).

If we are talking about UX - more used bindings should be easier to type. IMHO open link at point is used less the open link.

Same for copy link and copy link at point.. changed too. Can we swap them so less harm is done?

PR #13844, opened 2020-08-14
add indium repl/debugging option to javascript layer

wminshew:
Adding an 'indium option to the javascript layer for debugging/repl. Haven't contributed before so am sure there's plenty of good feedback. Tried to follow the conventions as much as possible, but the distinction between `indium-interaction-mode` and `indium-debugger-mode` made it a little challenging

### DIFF
--- a/layers/+lang/javascript/packages.el
+++ b/layers/+lang/javascript/packages.el
@@ -19,6 +19,7 @@
     flycheck
     ggtags
     helm-gtags
+    indium
     imenu
     impatient-mode
     import-js
@@ -259,6 +260,38 @@
           "sr" 'spacemacs/skewer-eval-region
           "sR" 'spacemacs/skewer-eval-region-and-focus
           "ss" 'skewer-repl)))))
+
+(defun javascript/init-indium ()
+  (when (eq javascript-repl 'indium)
+    (use-package indium
+      :defer t
+      :init
+      (progn
+        (require 'indium)
+        (add-hook 'js2-mode-hook 'indium-interaction-mode)
+        ;; prefixes
+        (spacemacs/declare-prefix-for-mode 'indium-interaction-mode "mdi" "debugger")
+        (spacemacs/declare-prefix-for-mode 'indium-debugger-mode "mdd" "debugger")
+        ;; keybindings
+        (spacemacs/set-leader-keys-for-minor-mode 'indium-interaction-mode
+          "dil" 'indium-launch
+          "dic" 'indium-connect
+          "die" 'indium-eval-defun
+          "dii" 'indium-inspect-expression
+          "dir" 'indium-repl-switch-from-buffer
+          "did" 'indium-switch-to-debugger
+          "dib" 'indium-toggle-breakpoint
+          "diB" 'indium-remove-all-breakpoints-from-buffer)
+        (spacemacs/set-leader-keys-for-minor-mode 'indium-debugger-mode
+          "ddp" 'indium-debugger-previous-frame
+          "ddn" 'indium-debugger-next-frame
+          "ddn" 'indium-debugger-next-frame
+          "ddi" 'indium-debugger-step-into
+          "dds" 'indium-debugger-step-over
+          "ddo" 'indium-debugger-step-out
+          "ddv" 'indium-debugger-evaluate
+          "ddl" 'indium-debugger-locals
+          "ddc" 'indium-debugger-resume)))))
 
 (defun javascript/post-init-tern ()
   (add-to-list 'tern--key-bindings-modes 'js2-mode))


layers/+lang/javascript/packages.el
```diff
@@ -260,6 +261,38 @@
           "sR" 'spacemacs/skewer-eval-region-and-focus
           "ss" 'skewer-repl)))))
 
+(defun javascript/init-indium ()
+  (when (eq javascript-repl 'indium)
+    (use-package indium
+      :defer t
+      :init
+      (progn
+        (require 'indium)
```
smile13241324, line 270: You should not require a package manually, instead this should be done by use-package and the autoload mechanism in emacs.

layers/+lang/javascript/packages.el
```diff
@@ -260,6 +261,38 @@
           "sR" 'spacemacs/skewer-eval-region-and-focus
           "ss" 'skewer-repl)))))
 
+(defun javascript/init-indium ()
+  (when (eq javascript-repl 'indium)
```
smile13241324, line 265: Good idea, but slightly wrong place, please move it into the `use-package if` directive.

layers/+lang/javascript/packages.el
```diff
@@ -260,6 +261,38 @@
...
+    (use-package indium
+      :defer t
+      :init
+      (progn
+        (require 'indium)
+        (add-hook 'js2-mode-hook 'indium-interaction-mode)
+        ;; prefixes
+        (spacemacs/declare-prefix-for-mode 'indium-interaction-mode "mdi" "debugger")
```
smile13241324, line 273: Do you really want to set a prefix for `mdi` I would have expected to have a prefix for `md` instead for debugging.
Also you need to make sure not to bind this when `dap` is activated i.e.
```elisp
(pcase (spacemacs//javascript-backend)
    (`lsp (yourBindings)))
```

layers/+lang/javascript/packages.el
```diff
@@ -260,6 +261,38 @@
...
+        (spacemacs/declare-prefix-for-mode 'indium-debugger-mode "mdd" "debugger")
```
smile13241324, line 274: Same here

layers/+lang/javascript/packages.el
```diff
@@ -260,6 +261,38 @@
...
+        ;; keybindings
+        (spacemacs/set-leader-keys-for-minor-mode 'indium-interaction-mode
```
smile13241324, line 276: These are repl bindings and must be bound under the prefix `s` as to the conventions. Also there are certain standard bindings which must be set for a repl. Please have a look at how the `nodejs` repl is set up in this layer for a good example.

layers/+lang/javascript/packages.el
```diff
@@ -260,6 +261,38 @@
...
+          "dil" 'indium-launch
+          "dic" 'indium-connect
+          "die" 'indium-eval-defun
+          "dii" 'indium-inspect-expression
+          "dir" 'indium-repl-switch-from-buffer
+          "did" 'indium-switch-to-debugger
+          "dib" 'indium-toggle-breakpoint
+          "diB" 'indium-remove-all-breakpoints-from-buffer)
+        (spacemacs/set-leader-keys-for-minor-mode 'indium-debugger-mode
```
smile13241324, line 285: This is the debugger which is ok to be bound here, the question is wouldn't it be nice to also have this minor mode activated when `js2-mode` is started given that the non lsp backends here define no debugging solution?

Of course you need to avoid binding these in case of `lsp` as in this layer `dap` is loaded which is its own debugging server solution.

layers/+lang/javascript/packages.el
```diff
@@ -19,6 +19,7 @@
     flycheck
     ggtags
     helm-gtags
+    indium
```
smile13241324, line 22: You need to document the additional repl choice in the README. In addition you need to extend the documentation for the layer variables to show that the new setting of `indium` is supported.

smile13241324 (changes requested):
Hi @wminshew,

I have just reviewed your PR and found some stuff we should get fixed first before this can be merged.
Please have a look at my comments and see whether you can fix them.

Any questions, just drop me a line.
Otherwise please ping me when you are ready for another review.

BR
Max

wminshew:
thanks for the feedback! will re-ping when ready for review

smile13241324:
Just stopped by to check the progress on this PR, @wminshew are you planning to continue working on this PR?
If not could you drop me a short line, then I could take over to get this merged.

wminshew:
Hey -- thanks for following up, ended up moving away from using javascript debuggers & so forgot about this PR. Not doing much with javascript (or indium) anymore so unlikely to return to this PR in the near term--please do takeover if this is something you think would be valuable to add!

smile13241324:
As we already have full support for a graphical debugger via LSP (DAP) I think I am just closing this for now.

Issue #13821, opened 2020-08-06
Pasting into cells of table.el breaks table


WinogradSchema:
#### Description :octocat:
Pasting into cells of table.el

#### Reproduction guide :beetle:
- Start Emacs
- Open org buffer
- org-table-create-with-table.el (, t N)
- 4x7
- Open table editor with C-c C-'
- Paste a large chunk of text
- Try C-c C-c to adjust

*Observed behaviour:* :eyes: :broken_heart:

Table no longer adjusts its cell width. Nearby cells no longer detected. Odd
characters appearing.

*Expected behaviour:* :heart: :smile:

C-c C-c would adjust table cell on paste similar to how it does on manually
filling cells

#### System Info :computer:
- OS: gnu/linux
- Emacs: 26.3
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 04b80d610)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(table-manipulation
 (chinese :variables chinese-enable-youdao-dict t)
 pdf lsp gtags search-engine python latex csv html java django c-c++ javascript kotlin w3m emms auto-completion better-defaults emacs-lisp git helm lsp markdown multiple-cursors
 (org :variables org-enable-github-support t org-projectile-file "TODOs.org" org-enable-org-journal-support t org-journal-dir "~/Dropbox/Org/Shared/journal/" org-journal-file-format "%Y%m%d.org" org-image-actual-width 450 org-latex-pdf-process
      '("xelatex -interaction nonstopmode %f" "xelatex -interaction nonstopmode %f")
      org-hide-emphasis-markers t org-journal-enable-agenda-integration t)
 (shell :variables shell-default-height 30 shell-default-position 'bottom)
 spell-checking syntax-checking version-control treemacs)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND DBUS GSETTINGS GLIB NOTIFY ACL LIBSELINUX GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XDBE XIM MODULES THREADS XWIDGETS LCMS2

#### Backtrace :paw_prints:
```
Not relevant
```

Burkean:
[...]

PR #13845, opened 2020-08-15
Fix org table evil paste

duianto:
Paste using *table--cell-yank, in org-src-mode.

Fixes: Pasting into cells of table.el breaks table #13821

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2815,6 +2815,7 @@ Other:
   (thanks to Tianshu Wang)
 - Enabled =org-habit= module
   (thanks to Mariusz Klochowicz)
+- Fixed evil paste in a =org-src-mode= table (thanks to duianto)
 **** Osx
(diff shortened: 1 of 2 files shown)


duianto:
This solution won't work, because `org-src-mode` buffers aren't limited to editing org tables.

Pasting fails when it calls `*table--cell-yank` in for example a python org source buffer.

Which can be opened by pressing `,'` (`org-edit-special`) in a python source block:
```org
#+begin_src python
  def f():
    print('test')
#+end_src
```

Is it possible to detect that the `org-src-mode` buffer is editing a table?

duianto:
I think I found a solution.

Now `p` only pastes with: `*table--cell-yank` when:
- the `org-src-mode` is `t`
- and the text property `table-cell` is `t` at the cursor

smile13241324 (approved):
Nice fix :+1:

Issue #12403, opened 2019-06-04
exec-path ordering is inconsistent with .spacemacs.env


leezu:
#### Description :octocat:
For example, when putting
/path/to/my/home/.pyenv/bin/:/path/to/my/home/.pyenv/shims/ at the front of the
PATH in .spacemacs.env, they appear in the middle of the exec-path variable.
/usr/bin appears before
/path/to/my/home/.pyenv/bin/:/path/to/my/home/.pyenv/shims/ in exec-path, thus
takes precedence.

#### Reproduction guide :beetle:
- Configure PATH in .spacemacs.env
- Start Emacs

*Observed behaviour:* :eyes: :broken_heart:
- exec-path ordering is inconsistent with .spacemacs.env

*Expected behaviour:* :heart: :smile:
- exec-path ordering is the same as in .spacemacs.env

#### System Info :computer:
- OS: gnu/linux
- Emacs: 26.2
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. fc3677131)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(helm
 (auto-completion :variables auto-completion-enable-sort-by-usage t auto-completion-enable-help-tooltip t)
 (spell-checking :variables spell-checking-enable-by-default nil enable-flyspell-auto-completion t)
 vinegar my-ui multiple-cursors neotree bibtex
 (org :variables org-projectile-file "~/org/projectile.org" org-want-todo-bindings t org-agenda-span 1 org-agenda-start-on-weekday nil)
 (clip2org :variables clip2org-clippings-file "~/Dropbox/Books/My Clippings.txt")
 my-org
 (deft :variables deft-directory "~/org/deft")
 (my-mail :variables notmuch-delete-tags
          '("+deleted" "-inbox" "-autoinbox")
          notmuch-spam-tags
          '("+spam" "-inbox" "-autoinbox")
          notmuch-kill-tags
          '("+killed" "-inbox" "-autoinbox"))
 finance dash sphinx cscope emacs-lisp package-lint git github
 (latex :variables latex-enable-folding t latex-build-command "LatexMk")
 restructuredtext markdown pandoc syntax-checking
 (ess :variables ess-disable-underscore-assign t)
 (python :variables python-enable-yapf-format-on-save nil)
 java go yaml lua semantic
[...]

Issue #12396, opened 2019-05-31
Broken path settings after upgrade to the develop branch 


abhi18av:
#### Description :octocat:

Spacemacs should behave the very same way wrt path variables in `master` and `develop`.

#### Reproduction guide :beetle:
- Start Emacs
- Open any python file

*Observed behaviour:* :eyes: :broken_heart:
- Spacemacs only detects the pre-installed `python2` by apple and the `anaconda3` version.

This was fine in the `master` branch but as soon as I have switched to the `develop` branch the path settings seem to have become broken, even though everything is correct in the `.spacemacs.env` file.

*Expected behaviour:* :heart: :smile:

Spacemacs should behave the very same way wrt path variables in `master` and `develop`.

#### System Info :computer:
- OS: darwin
- Emacs: 26.2
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. a281e6979)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(alda
 (auto-completion
  (haskell :variables haskell-completion-backend 'intero haskell-process-type 'stack-ghci haskell-enable-hindent-style "johan-tibell"))
 (auto-completion :variables auto-completion-enable-sort-by-usage t auto-completion-enable-help-tooltip t auto-completion-tab-key-behavior 'cycle auto-completion-private-snippets-directory "~/.spacemacs.d/snippets/" auto-completion-enable-snippets-in-popup t :disabled-for org)
 copy-as-format
 (clojure :variables clojure-enable-fancify-symbols t clojure-enable-sayid t clojure-enable-clj-refactor t)
 csv deft emoji evil-snipe github git
 (helm :variables helm-enable-auto-resize nil)
 html hy imenu-list
 (ibuffer :variables ibuffer-group-buffers-by 'projects)
 javascript json
 (multiple-cursors :variables multiple-cursors-backend 'evil-mc)
 (markdown :variables markdown-live-preview-engine 'vmd)
 (org :variables org-enable-github-support t org-enable-bootstrap-support t org-enable-org-journal-support t org-enable-reveal-js-support t org-enable-hugo-support t)
 parinfer
 (python :variables python-enable-yapf-format-on-save t python-sort-imports-on-save t python-test-runner
         '(pytest nose))
 prodigy react rust ranger shell-scripts spacemacs-org spacemacs-evil
[...]

PR #13858, opened 2020-08-19
[core] Fix append path from .spacemacs.env

mirubiri:
Fixes the issue described in #12403 and potentially #12396 #10906.

The function `load-env-vars-set-env` is appending to `exec-path` the PATH values found in `.spacemacs.env`. 

This means that the directories defined in `exec-path` will be searched first instead of the ones we have defined in `.spacemacs.env` PATH variable. 

Usually `exec-path` already includes `/bin;/usr/bin;/usr/local/bin`, so it's not possible to look for binaries in the directories specified in our custom PATH variable. This is needed when working with shims, which is something we usually found in tools like `asdf`,`rbenv` or `brew`..

This PR preprends the `.spacemacs.env` PATH variable to `exec-path` to avoid this issues.


smile13241324 (approved):
Strange, actually this should have appended at the end of `exec-path` already given the `append` argument. When I check the content of `exec-path` on my linux system it also seems to be in the right order.

Anyway your code is much more functional and if it fixes issues on other oses it should definitely be merged.

core/libs/load-env-vars.el
```diff
@@ -87,12 +87,10 @@
...
         (if (string-equal "PATH" key)
             (let ((paths (split-string value path-separator)))
-              (dolist (p paths)
-                (add-to-list 'exec-path (directory-file-name
-                                         (subst-char-in-string ?\\ ?/ p))
-                             'append))
-              (setenv "PATH" (mapconcat convert-to-os-path exec-path
-                                        path-separator)))
+              (setq exec-path (append
+                               (mapcar (lambda (path) (subst-char-in-string ?\\ ?/ path)) paths)
```
smile13241324, line 91: Ups, just found a possible issue, the original code did also call `directory-file-name` on the paths. Please restore that part of the logic otherwise we may risk issues on certain systems.

smile13241324 (changes requested):
Looks like I was too fast, I think I have found a possible issue, please check my comment.
As this is affecting the core some manual testing should happen on this PR before it is merged.

core/libs/load-env-vars.el
mirubiri: Upss, restored!

mirubiri:
> Strange, actually this should have appended at the end of exec-path already given the append argument. When I check the content of exec-path on my linux system it also seems to be in the right order.

It is appended at the end, but the problem is that `exec-path` might already contain paths for binaries like `usr/bin`. I think it's interesting to `prepend` the PATH var defined in `.spacemacs.env` to `exec-path` instead. 

For example I always want that the first section of my PATH to be `~/.asdf/shims` to look for binaries, but it doesn't have any effect if `exec-path` already includes `usr/bin`. If I'm allowed to prepend my PATH to `exec-path` problem solved and my configurations will take effect. This should be valuable for any system I guess.

core/core-env.el
```diff
@@ -63,7 +63,7 @@ current contents of the file will be overwritten."
           "# regexps in `spacemacs-ignored-environment-variables'. If you add any\n"
           "# duplicate settings for a variable, only the first setting is effective.\n"
           "# PATH is a special case: all PATH settings are read, each non-duplicate\n"
-          "# directory entry is appended to the `exec-path' variable, and then PATH is\n"
+          "# directory entry is prepended to the `exec-path' variable, and then PATH is\n"
```
mirubiri, line 66: @smile13241324 I have also updated the generated documentation here to reflect this change

mirubiri:
@smile13241324 Are you launching it from the console? These are my values when launching it in osx with emacs-mac port.

**Withouth this PR:**

Original: (/usr/local/bin /usr/bin /bin /usr/sbin /sbin /usr/local/Cellar/emacs-mac/emacs-27.1-mac-8.0/libexec/emacs/27.1/x86_64-apple-darwin19.6.0)

Modified: (/usr/local/bin /usr/bin /bin /usr/sbin /sbin /usr/local/Cellar/emacs-mac/emacs-27.1-mac-8.0/libexec/emacs/27.1/x86_64-apple-darwin19.6.0 /Users/alex/.asdf/shims /usr/local/opt/asdf/bin /Users/alex/bin /usr/local/sbin /Users/alex/Library/Python/3.7/bin)

**With this PR:**

Original: (/usr/local/bin /usr/bin /bin /usr/sbin /sbin /usr/local/Cellar/emacs-mac/emacs-27.1-mac-8.0/libexec/emacs/27.1/x86_64-apple-darwin19.6.0)

Modified: (/Users/alex/.asdf/shims /usr/local/opt/asdf/bin /usr/local/bin /usr/bin /bin /usr/sbin /sbin /Users/alex/bin /usr/local/bin /usr/local/sbin /Users/alex/Library/Python/3.7/bin /usr/local/bin /usr/bin /bin /usr/sbin /sbin /usr/local/Cellar/emacs-mac/emacs-27.1-mac-8.0/libexec/emacs/27.1/x86_64-apple-darwin19.6.0)

PATH=/Users/alex/.asdf/shims:/usr/local/opt/asdf/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Users/alex/bin:/usr/local/bin:/usr/local/sbin:/Users/alex/Library/Python/3.7/bin


As you can see, starting my path starts with `/Users/alex/.asdf/shims`, without this PR my PATH is appended to the end of exec-path and with this PR, my PATH is prepended to exec-path.

I'm going to remove make some changes to remove the duplicates

mirubiri:
Duplicates removed 🤞

smile13241324:
Thanks 👍 , I'll have another go in the next view days.

smile13241324 (approved):

smile13241324:
Looks good now, and the code is still much clearer than before. 

The order of paths does not change for me though but this may be system related. I am not starting emacs from the console I am running it from latest Gnome on Manjaro. Anyway if it solves issues on one Os and does not hurt the others it is worth to be merged.

PR #13860, opened 2020-08-19
[lsp] Remove mention of `lsp-clients`

fosskers:
This PR fixes the recent `lsp` layer breakage.

PR #13861, opened 2020-08-19
removed lsp-clients.el in lsp-mode package - fixes #13857

### LINKED ISSUE #13857: lsp-clients module is gone
Looks like it was recently [refactored](https://github.com/emacs-lsp/lsp-mode/pull/2084).
`develop` branch errors on startup after a package update.

Local modification that simply replaces `(require 'lsp-clients)` with `(require 'lsp)` seems to work.

stkain:
Hello, 

lsp-mode removed lsp-clients.el which contained a couple of (require '...)
The consequence is that now starting spacemacs results in an error due to
(require 'lsp-clients)  in packages.el 

and a not fully functional lsp-mode (major mode commands are missing)

my pull request removes this require statement and replaces it with the require
statements in the now non-existing lsp-clients.el.

So the sequence of events on loading layer lsp is the same as before.

Thank you <3

### DIFF
--- a/layers/+tools/lsp/packages.el
+++ b/layers/+tools/lsp/packages.el
@@ -23,7 +23,11 @@
     :defer t
     :config
     (progn
-      (require 'lsp-clients)
+      (require 'lsp-mode)
+      (require 'dash)
+      (require 'dash-functional)
+      (require 'rx)
+      (require 'cl-lib)
       (spacemacs/lsp-bind-keys)
       (setq lsp-prefer-capf t)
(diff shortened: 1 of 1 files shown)


duianto:
Another PR was posted before this one.
[lsp] Remove mention of `lsp-clients` #13860

It changed the require to `lsp` instead of `lsp-clients`.

But it seems like it worked to just remove that require line all together.
That PR has been merged into the `develop` branch, therefore we'll close this one.

PR #13874, opened 2020-08-23
Move the Rust RON support to my own package (please read the description, the title may need context)

enderger:
# Moving RON Mode to my own package
Hi, it's me again. A few weeks ago, I added Nabero's ron-mode package to the Spacemacs Rust layer. Since then, I've been working on my own package which adds support for RON to EMACS for a few reasons:

1. Nabero's package is not on MELPA. My new package has finished going through their review process and is now a part of that repository.

2. Nabero's package is a fork of a dead project. My new implementation is built from scratch, which is nessicary for a few purposes (most notably the licensing part, which is discussed in point 3).

3. Nabero's package appears to be unlicensed. This may be problematic in a project as large as Spacemacs (see https://choosealicense.com/no-permission/, GitHub's TOS prevent its use from becoming illegal but it is best to use a properly licensed project). My new package is licensed under the BSD 2-Clause License.

4. Nabero's package has issues disabled, preventing us from getting in contact with him were the package to break or to get it ready. Mine has issues enabled (or at least, Fossil SCM's ticket system since issues is a Git thing).

---
In this PR, I switch Nabero's package out for my own varient, which addresses all above points. The source code is available at https://chiselapp.com/user/Hutzdog/repository/ron-mode/home.

### DIFF
--- a/layers/+lang/rust/packages.el
+++ b/layers/+lang/rust/packages.el
@@ -19,7 +19,7 @@
     (flycheck-rust :requires flycheck)
     ggtags
     helm-gtags
-    (ron-mode :location (recipe :fetcher github :repo "nabero/ron-mode"))
+    ron-mode
     racer
     rust-mode
     smartparens


sdwolfz (approved):

sdwolfz:
Related to the licensing, it's my understanding that all code written in Emacs Lisp is by default GPL3 licensed because Emacs itself is GPL3, there is no other implementation that can run it, and any packages use Emacs internal code directly, so no dynamic linking is involved. So the old package was not unlicensed, and yours is most likely GLP3 too.

:point_up: not an issue, just a FYI.

enderger:
The BSD License is "GPL Compatible", a requirement listed in MELPA. Also, as the code is technically not redistribuitng EMACS, it is my understanding that it is not a derivative work and hence is not forcibly GPL licensed.

PR #13875, opened 2020-08-24
[lua] Remove require lsp-clients

duianto:
lsp-clients have been removed upstream as discussed in:
lsp-clients module is gone #13857


smile13241324:

PR #13876, opened 2020-08-24
[lua] Ignore EmmyLua-LS-all.jar in .emacs.d

duianto:
The Lua readme: https://github.com/syl20bnr/spacemacs/tree/develop/layers/+lang/lua
and lsp-mode instructions: https://github.com/emacs-lsp/lsp-mode/wiki/Install-EmmyLua-Language-server
suggest adding `EmmyLua-LS-all.jar` to `~/.emacs.d`

That makes the file appear as an untracked file in magit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,7 @@ eclipse.jdt.ls
 edts/
 elnode/
 elpa/
+EmmyLua-LS-all.jar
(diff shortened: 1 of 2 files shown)


smile13241324 (approved):
:+1:

Issue #13868, opened 2020-08-20
[layers/+lang/haskell] Remove Intero and company-ghci backends


Martinsos:
As discussed here https://github.com/syl20bnr/spacemacs/pull/13802#event-3672622287 with @smile13241324, company-ghci is not competitive with other backends any more, while intero is not maintained any more, so both of them should be removed.
This should be done after issue https://github.com/syl20bnr/spacemacs/issues/13867 has been resolved, because company-ghci is currently default backend so we want to first switch that to Dante.
I plan to tackle this in the following week or so.

smile13241324:
This has been merged today so this ticket is done.

Issue #13867, opened 2020-08-20
[layers/+lang/haskell] Make Dante default backend


Martinsos:
As discussed here https://github.com/syl20bnr/spacemacs/pull/13802#event-3672622287 with @smile13241324, company-ghci (current default) is not so relevant anymore while Dante works very well and out of the box in many cases, so we want to have Dante as default backend.
I plan to tackle this in the following week or so.

smile13241324:
This has been done today, so this ticket can be closed.

PR #13877, opened 2020-08-25
[layers/+lang/haskell] Removed intero, ghc-mod and company-ghci backends, set dante as default backend.

Martinsos:
@duianto and @smile13241324 : here is another PR for Haskell layer with changes that we discussed. Solves issues #13867 and #13868.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -174,6 +174,8 @@ the [[file:CHANGELOG.org][CHANGELOG.org]] file.
   - Swapped key bindings for better usability (thanks to Tony Lotts):
     - ~SPC m s s~ for =haskell-interactive-switch=
     - ~SPC m s S~ for =spacemacs/haskell-interactive-bring=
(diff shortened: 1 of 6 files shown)


michaelpj:
`ghc-mod` is also very obsolete, and I think could be removed.

Martinsos:
@michaelpj true, thanks for pointing this out. On their repo they say: `Please note that using ghc-mod as a user facing tool for IDE/Editor integration is no longer supported or maintained. We are in the process of refocusing our efforts around haskell-ide-engine as the main user/IDE facing program.`. So I agree we should remove it also, I will add that to this PR.

Martinsos:
All right, should be it now! Intero, company-ghci and ghc-mod removed, dante is default.

smile13241324 (approved):
Woah the entire layer looks much clearer now.

Martinsos:
Ah yes, the best PR is the PR that removes a lot of code :)!

PR #13883, opened 2020-08-28
Set org-log-done to 'time instead of t

bcc32:
Setting this variable to 'time is equivalent, while setting to t is
deprecated.


smile13241324 (approved):
:+1:

PR #13891, opened 2020-08-29
[ipython] Fix elisp whitespace

bhipple:
This runs the default spacemacs elisp formatter on buffer save, with no
functional changes. I have further changes I'd like to integrate, but first
want to sweep this large but no-op diff forward.

### DIFF
--- a/layers/+tools/ipython-notebook/packages.el
+++ b/layers/+tools/ipython-notebook/packages.el
@@ -21,15 +21,15 @@
     :config
     (with-eval-after-load 'ein-notebook
       (add-hook 'ein:notebook-mode-hook
-		(lambda ()
-		  (add-to-list 'minor-mode-overriding-map-alist
-			       `(ein:notebook-mode . ,ein:notebook-mode-map))))
+    (lambda ()
+      (add-to-list 'minor-mode-overriding-map-alist
+             `(ein:notebook-mode . ,ein:notebook-mode-map))))
       (mapc
        (lambda (mode)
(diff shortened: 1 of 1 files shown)


smile13241324:

smile13241324:
Btw not sure what formatter you have used, on my emacs 27 I have got a totally different result, I am using `(evil-indent (point-min) (point-max))`

bhipple:
I'm not sure either. I just ran `:w` with the spacemacs setting to cleanup all whitespace.

smile13241324:
Thats not the standard formatting, if you want to change elisp files please apply formatting with a function like below:
``` elisp
  ;; Format file on save
  (defun format-for-filetype ()
    "Run generic format function if not a mode specific one is available"
    (let ((filetypes '("elm" "hs" "c" "cpp" "h" "hpp" "py" "pyc" "robot" "tf" "go" "yml" "yaml")))
      (if (not (member (file-name-extension (buffer-file-name)) filetypes))
          (save-excursion
            (evil-indent (point-min) (point-max))))))
  (add-hook 'before-save-hook 'format-for-filetype)

```
or much more easier and cleaner just activate `parinfer` in your dotfile via the parinfer layer. This is a modern approach for working with lisp structures not working like moving a cursor within a tree but rather setting the paranthesis, depending on the indentation like in python. This forces your files to be properly formatted.

Independent of which approach you are using you should apply:
``` elisp
(evil-indent (point-min) (point-max))
```
before you commit an elisp buffer.

Issue #13869, opened 2020-08-20
[layers/+lang/haskell] Improve README (reorganize and polish)


Martinsos:
As discussed here https://github.com/syl20bnr/spacemacs/pull/13802#event-3672622287 with @smile13241324, there are some ways we could improve README to make it easier to start with Haskell layer.

Ideally, we would handle this after tackling issue https://github.com/syl20bnr/spacemacs/issues/13868, to make sure those changes are well documented.

Short summary of what we should do:
 - "install" section which just shows how the layer is declared.
 - "Configuration" section after "install" section of which the first choice is "Choosing a backend" which defines a primary backend which should be "dante" and alternative backends with additional setup procedures. It should give easy to understand information when to choose which. A good example would be the "groovy" layer. If there are configurations which are independent of the chosen backend it should follow on the same level as the "Choosing a backend" headline. For LSP if there are multiple servers to choose from the differences should be described below the "LSP" headline which is below the "Choosing a backend" headline.

I plan to tackle this in the following week or so.

Martinsos:
Note for myself: If not yet in README, add information/FAQ on how to fix Dante misbehaving regarding tests, if that is happening.

smile13241324:
This has been merged by now, I am putting this ticket into done state.

PR #13902, opened 2020-09-01
Improved Haskell language layer README.

Martinsos:
Improved Haskell README as discussed here https://github.com/syl20bnr/spacemacs/issues/13869 @smile13241324 .

There are a few things I was wondering about:
1. groovy layer README does not mention that user should enable auto-complete or syntax-checking layer. I wonder why is that so?
2. Syntax-checking is still a standalone section. I was considering if I can make it part of Configuration, but it really goes beyond just configuration, it explains how things work, so I decided to leave it as it is for now.
3. I wanted to make it very obvious what the minimal setup is. I am not sure if I managed to do that. Right now, I think new user needs to read Description, which is short, Install, which is straight forward, and then Configuration, and they are ready to go. Key bindings and especially Syntax checking and FAQ are kind of like bonuses and are not required to get started. But is that clear to the user? So I was thinking, maybe we could have a "QuickStart" section at the beginning that would summarize only the critical information from Description, Install and Configuration? But I gave up on that because it still felt like a lot of duplication. I might experiment more with this if I manage to get my brother to use Spacemacs :D, I can observe him then and see where he will get stuck. I am driving this whole point because I felt overwhelmed when I first started reading Haskell README, but now I know it very well so I can't remember exactly what I was confused about.

EDIT: Actually, I just added "Quick start" section which I think could solve the problem I described above under point number three -> it very quickly instructs user how to move around the README and what to look at first. Let me know what you think about it, we can remove it if it is too much. Btw in this Quick start section I tried using internal links to headers in Org, and they worked in emacs, but they don't work when README is read via Github, so I gave up on them and just bolded the names of sections/headings.

### DIFF
--- a/layers/+lang/haskell/README.org
+++ b/layers/+lang/haskell/README.org
@@ -7,13 +7,17 @@
 * Table of Contents                     :TOC_5_gh:noexport:
 - [[#description][Description]]
   - [[#features][Features:]]
+- [[#quick-start-and-how-to-use-this-readme][Quick start (and how to use this README)]]
 - [[#install][Install]]
   - [[#layer][Layer]]
   - [[#dependencies][Dependencies]]
   - [[#setup-path][Setup PATH]]
+- [[#configuration][Configuration]]
   - [[#choosing-a-backend][Choosing a backend]]
     - [[#dante][=dante=]]
     - [[#lsp][=lsp=]]
+      - [[#hie][=hie=]]
+      - [[#hls][=hls=]]
   - [[#optional-extras][Optional extras]]
     - [[#structured-haskell-mode][structured-haskell-mode]]
     - [[#hindent][hindent]]
@@ -28,10 +32,8 @@
 - [[#syntax-checking][Syntax checking]]
   - [[#flycheck][Flycheck]]
   - [[#hlint][HLint]]
-  - [[#interactive-haskell-mode][Interactive haskell-mode]]
-  - [[#flymake][Flymake]]
-  - [[#troubleshooting][Troubleshooting]]
 - [[#faq][FAQ]]
+  - [[#dante-reports-missinghidden-imports-for-test-files][Dante reports missing/hidden imports for test files.]]
   - [[#the-repl-doesnt-work][The REPL doesn't work]]
   - [[#the-repl-is-stuck][The REPL is stuck]]
   - [[#indentation-doesnt-reset-when-pressing-return-after-an-empty-line][Indentation doesn't reset when pressing return after an empty line]]
@@ -48,10 +50,20 @@
 This layer adds support for the [[https://www.haskell.org/][Haskell]] language.
 
 ** Features:
-- syntax highlighting for [[https://github.com/haskell/haskell-mode][haskell source]], [[https://github.com/haskell/haskell-mode][cabal files]], [[https://github.com/bgamari/cmm-mode][C-- source]],
-- auto-completion with one of the selected backends (=dante= or =lsp=).
+- syntax highlighting for [[https://github.com/haskell/haskell-mode][haskell source]], [[https://github.com/haskell/haskell-mode][cabal files]], [[https://github.com/bgamari/cmm-mode][C-- source]]
+- auto-completion and syntax-checking with one of the selected backends (=dante= or =lsp=).
 
-*This layer is under construction, it needs your contributions and bug reports.*
+* Quick start (and how to use this README)
+- Follow instructions in *Install* section to correctly install the layer.
+- Ensure that you have =auto-completion= and =syntax-checking= layers enabled.
+- Set backend to =dante= (check *Configuration* section for details). =dante= is default already, but if =lsp= layer is enabled, you will have to set it explicitly.
+- You are ready to go! Open any Haskell project and enjoy syntax-checking, auto-completion and more.
+
+After that, check the rest of the README to:
+- Learn about more powerful (but more complicated to set up) backend: =lsp= with =hie= / =hls=: Check *Configuration* -> *lsp*.
+- Learn about all the functionalities and key bindings: Check *Key bindings* and *Configuration* -> *Optional extras*.
+- Learn about the details of how syntax-checking works and how it can be tweaked: Check *Syntax checking*.
+- Find solutions to common problems in FAQ: Check *FAQ*.
 
 * Install
 ** Layer
@@ -85,13 +97,16 @@ packages. If you are using =cabal= it should be =~/.cabal/bin= or
 =~/Library/Haskell/bin= (for 'Haskell for Mac' users). If you are using =stack=
 then it should be =~/.local/bin=.
 
-For information about setting up =$PATH=, check out the corresponding section in
+For more information about setting up =$PATH=, check out the corresponding section in
 the FAQ (~SPC h SPC $PATH RET~).
 
+* Configuration
 ** Choosing a backend
-First, ensure that you have =auto-completion= layer enabled.
+Language backend is the core component of a language layer - it has the responsibility of compiling/parsing the actual code and reporting errors, warnings, suggesting fixes, auto-completions and more.
 
-To choose a haskell backend, set the haskell layer variable =haskell-completion-backend=:
+To get the most out of the language backend, you will want to ensure that you have =auto-completion= and =syntax-checking= layers enabled.
+
+Then, to choose a haskell backend, set the haskell layer variable =haskell-completion-backend=:
 
 #+BEGIN_SRC emacs-lisp
   (haskell :variables haskell-completion-backend 'dante)
@@ -99,7 +114,7 @@ To choose a haskell backend, set the haskell layer variable =haskell-completion-
 
 Supported values for =haskell-completion-backend= are =dante= and =lsp=.
 
-If you haven't specified any value for =haskell-completion-backend=,
+If you don't specify any value for =haskell-completion-backend=,
 =dante= will be used as default backend, unless the layer =lsp= is enabled,
 in which case =lsp= is used as default backend.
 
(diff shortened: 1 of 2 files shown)


smile13241324:
Thanks for the hard work, I will have a look.

As to 1. normally the layers do not tell that you need to have syntax checking layer or auto complete layer installed, there is a general user manual which we expect new comers to read first there the whole system of syntax checking layer and auto complete layer is described in detail.

The individual layers should not repeat information already written down in a general document, especially for maintenance reasons.

Martinsos:
> Thanks for the hard work, I will have a look.
> 
> As to 1. normally the layers do not tell that you need to have syntax checking layer or auto complete layer installed, there is a general user manual which we expect new comers to read first there the whole system of syntax checking layer and auto complete layer is described in detail.
> 
> The individual layers should not repeat information already written down in a general document, especially for maintenance reasons.

Makes sense, the more we duplicate information the harder it is to keep it in sync.

Still, I can say from my experience: I went through most of the spacemacs docs, and still I couldn't even remember there was syntax-checking or auto-complete layer once I got to read Haskell layer README! So it is probably useful to mention at least some things, maybe those that don't change quickly. But I will let you make the final decision on that!

layers/+lang/haskell/README.org
```diff
@@ -116,28 +134,48 @@ Backend can be chosen on a per project basis using directory local variables
...
-
 *** =lsp=
-=lsp= requires an appropriate installation of =hie=, the official Haskell language server, to work properly.
+[[https://microsoft.github.io/language-server-protocol][Language Server Protocol]] is a standard for implementing language backends.
+
+In Haskell layer, you can use a backend that implements Language Server Protocol for Haskell by specifying =lsp= as backend
+and then installing concrete backend implementation, of which there are two available at the moment: =hie= and =hls=.
+
+Enabling the =lsp= backend requires the =lsp= layer to be enabled, and provides access to
+all the additional =lsp-mode= key bindings. =lsp= uses =lsp-ui= to show the errors/warnings instead of =flycheck=.
```
smile13241324, line 157: Well the line about LSP not using flycheck is not right, actually there is a specific checker registered within the flycheck system which nowadays is called `lsp` not `lsp-ui` anymore (latest version). However that's still the flycheck system so the same setup instructions apply.

To elaborate a bit more, the traditional emacs system is to register multiple checkers with flycheck which then run sequentially. This is normally slow and requires a lot of setup within emacs. Lsp-mode in contrary passes the job of maintaining linter threads to the external LSP server allowing them to run in parallel. For the emacs user this is visible by that there is only one checker registered called `lsp`. Normally this is ok, however if the server is not well enough designed to be able to run multiple linters some layers have tried to register a second checker after `lsp` this however is very difficult to get right and again is very slow. The right approach in this case is to open a PR for the LSP server to support that specific linter.

layers/+lang/haskell/README.org
```diff
@@ -280,13 +318,14 @@ Refactor commands are prefixed by ~SPC m r~:
 Only some of the HLint suggestions can be applied.
 
 * Syntax checking
-At the moment there are three components, which can indicate
+There are multiple components that can indicate
```
smile13241324, line 321: I am not sure I like this section. 

Basically it should list all available checkers, which are all for flycheck. This however should not list haskell-mode interactive as this is your repl, which may also have syntax checking in place but is not strictly working on your code files.

So my suggestion would be to remove haskell-mode interactive from this section as well as remove the (via .*) block.

layers/+lang/haskell/README.org
```diff
@@ -325,6 +364,12 @@ Flycheck can fail silently for miscellaneous reasons. See the [[#faq][FAQ]]
 for troubleshooting.
```
smile13241324, line 364: We should remove the section about `flymake` this is horribly outdated.
In addition if we keep this troubleshooting subject, we should move it to the FAQ

layers/+lang/haskell/README.org
```diff
@@ -307,7 +346,7 @@ if it doesn't work, then you can change it with ~SPC e s~.
 the file) but bad coding style and code smell. The HLint checker is called
 *after* the flycheck GHC checker.
 
-HLint can be configured via .hlint.yaml.
+HLint can be configured per project via .hlint.yaml (check Hlint docs for more details).
 
 ** Interactive haskell-mode
 Finally, interactive haskell-mode (~SPC m s b~) also displays errors. These
```
smile13241324, line 352: Can you double check whether this is still true? Basically standard emacs next error and previous error should work in this buffer too.

layers/+lang/haskell/README.org
```diff
@@ -116,28 +134,48 @@ Backend can be chosen on a per project basis using directory local variables
...
-=lsp= requires an appropriate installation of =hie=, the official Haskell language server, to work properly.
+[[https://microsoft.github.io/language-server-protocol][Language Server Protocol]] is a standard for implementing language backends.
+
+In Haskell layer, you can use a backend that implements Language Server Protocol for Haskell by specifying =lsp= as backend
+and then installing concrete backend implementation, of which there are two available at the moment: =hie= and =hls=.
+
+Enabling the =lsp= backend requires the =lsp= layer to be enabled, and provides access to
+all the additional =lsp-mode= key bindings. =lsp= uses =lsp-ui= to show the errors/warnings instead of =flycheck=. 
+
+**** [[https://github.com/haskell/haskell-ide-engine][Haskell Ide Engine]] (=hie=)
```
smile13241324, line 159: Can you verify that links as headlines work in org and markdown? I want to avoid issues with the document conversion beforehand.

smile13241324 (changes requested):
@Martinsos all in all this is a decent rewrite of most of the layer documentation.

I have given it a good read and found one or two points you should check and some smaller ones you should change, before we can merge this into develop.

layers/+lang/haskell/README.org
Martinsos: Thanks for detailed explanation! But from what I read / understood, LSP errors don't appear in flycheck's error list? Or do they then? I would try it myself but I don't have HIE/HLS set up at the moment, might get to it during the weekend but certainly not sooner.

layers/+lang/haskell/README.org
Martinsos: Agreed, I didn't know much about flymake so I didn't want to touch it at the moment. Troubleshooting part -> it does not seem to bring any new information so I will just remove it.

layers/+lang/haskell/README.org
Martinsos: Well I have to admit I don't feel like I have enough knowledge to rewrite this section on my own, and I am not sure I 100% got what you are saying.

Are you proposing that we remove the first part completely, starting with "There are multiple" and ending with "- haskell-mode interactive"? And then that we also remove "** Interactive haskell-mode" section?

My thoughts:
1. I agree this section should be updated :D.
2. If we remove interactive-haskell mode section, are we losing any information? I don't see any other place mentioning error navigation shortcuts that are mentioned here (=haskell-goto-next-error= (~M-n~) and the other one), should we document those somewhere else? Under keybindings probably, but I am not sure if that should go under haskell mode or REPL hm. I could investigate that further, read more about haskell-mode, see if they are valid there also and so on.
3. If we remove that first section and interactive haskell section, not much is left in the Syntax checking section, right? Flycheck section talks about flycheck, which is something that could be looked up in syntax-checking layer docs. It mentions different checkers, but that feels more like implementation detail, why would use need to know explicitly about this? And finally it mentions HLint, but again it does not say anything that can't be looked up in Hlint documentation. So maybe just removing the whole section would make sense? Or making it much much shorter -> but what would you leave in it then?

layers/+lang/haskell/README.org
Martinsos: Hmmm intersting, so I tried it out and what I got is that haskell-goto-prev-error does not work (says No further notes from Haskell compiler) while standard emacs next error and previous error work!
Also, haskell-goto-prev-error does not respond to M-n in my case, it seems like it is not connected to that key binding.

So maybe we should just remove this section completely, the one with haskell interactive mode, as you mentioned?

layers/+lang/haskell/README.org
Martinsos: I can confirm that they work!

layers/+lang/haskell/README.org
smile13241324: Errors from lsp are registered as flycheck errors, as are all errors you see in any buffer (only if it is using very obscure or veeeerrryyy old code for this).

layers/+lang/haskell/README.org
smile13241324: Good point, no idea yet, lets keep it as it is for now, we can later revise it again.

layers/+lang/haskell/README.org
smile13241324: This is already happening, for the spacemacs websites, especially for develop.spacemacs.org, for this org is converted to markdown. You can check if this still works by running `spacemacs/publish-doc` in `core/core-documentation.el`. This should not fail for the haskell layer.

smile13241324:
Just had another look, I think now only two points are open before we can merge it. Please check the unresolved discussion threads and the we can finally merge it.

layers/+lang/haskell/README.org
Martinsos: Ok makes sense! I actually just moved links to the text, instead of headers, I think it is just eaiser that way. I did run spacemacs/publish-doc from core/core-documentation.el, but after I moved the links from headers, and it generated a lot of stuff and I haven't seen any errors reported. However maybe I missed some error there was a lot of logs, and I wasn't sure where docs got generated so I didn't inspect them. Anyway this should solve it!

layers/+lang/haskell/README.org
Martinsos: Ok, just to confirm -> I left syntax checking section as it was, but removed interactive haskell.

Martinsos:
Ok this should be it! I squashed all commits and took care of all unresolved comments. I also undid resolution of one previously resolved comment, just to confirm that I understood right, which leaves us at 3 unersolved comments right now. If those are all ok, you can merge.

smile13241324 (approved):

smile13241324:

PR #13912, opened 2020-09-04
Make find-init-file compatible with profile switcher

agzam:
<SPC f e i> should work with Emacs profile switcher plexus/chemacs

--- 

Hey folks, I recently tried using https://github.com/plexus/chemacs and noticed that `<SPC f e i>` opens chemacs's init.el instead of Spacemacs' init.el.

This tiny change fixes the problem for me, but maybe I'm not alone who wants this behavior, what do you think?

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -729,12 +729,11 @@ variable."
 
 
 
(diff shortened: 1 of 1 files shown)


smile13241324 (approved):
Looks cool, going to the emacs based dir and looking there for the init.el looks pretty equivalent to me, lets hope we are breaking some obscure configs with this.

Issue #13524, opened 2020-04-25
Company-capf does not show up in php file


phenix-factory:
#### Description :octocat:
Company-capf does not show up in php file.
I m using lsp backend and intelephense.

My full spacemacs config is on this repo : https://gitlab.com/phenix-factory/config-spacemacs

#### Reproduction guide :beetle:
- Start Emacs
- Open a php file
- Type some already existing variable

*Observed behaviour:* :eyes: :broken_heart:
Company does'nt provide completion at all.
But I can for it via company-capf command

*Expected behaviour:* :heart: :smile:
company working

#### System Info :computer:
- OS: gnu/linux
- Emacs: 26.3
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 1f6e39ea8)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
((auto-completion :variables auto-completion-return-key-behavior 'complete auto-completion-tab-key-behavior 'cycle auto-completion-enable-snippets-in-popup t auto-completion-enable-help-tooltip t auto-completion-enable-sort-by-usage t auto-completion-complete-with-key-sequence-delay 0.1 auto-completion-idle-delay 0.2)
 helm emacs-lisp
 (git :variables git-magit-status-fullscreen t)
 markdown csv
 (org :variables org-enable-reveal-js-support t)
 treemacs
 (erc :variables erc-server-list
      '(("irc.freenode.net" :port "6667" :ssl t :nick "phenix_spip")))
 (shell :variables shell-default-shell 'multi-term shell-default-height 30 shell-default-position 'bottom multi-term-program "/usr/bin/zsh")
 (spell-checking :variables spell-checking-enable-by-default nil spell-checking-enable-auto-dictionary nil enable-flyspell-auto-completion t)
 syntax-checking shell-scripts
 (php :variables php-backend 'lsp)
 restclient
 (sql :variables sql-capitalize-keywords t)
 (html :variables css-enable-lsp t less-enable-lsp t scss-enable-lsp t html-enable-lsp t web-fmt-tool 'prettier)
 lsp dap
 (javascript :variables javascript-fmt-tool 'prettier javascript-import-tool 'import-js javascript-backend 'lsp)
 (vue :variables vue-backend 'lsp)
 prettier import-js
 (node :variables node-add-modules-path t)
 lua imenu-list yaml finance colors dash multiple-cursors unicode-fonts
[...]

PR #13916, opened 2020-09-05
Disable php-extras when the php layer backend is lsp

AdeAttwood:
This conflicts with `company-capf` and you don't get any company completions
from the lsp-mode

Fix issue #13524


smile13241324 (approved):
Good catch, will be merged soon.

PR #13920, opened 2020-09-06
Update spacemacs-org README on bullet library

fabrik42:
Hi there! This is just a small fix in the documentation.

In https://github.com/syl20bnr/spacemacs/commit/145126875731e8ee38770b2adf709805f23672f7 org-bullets has been replaced with org-superstar, but the Readme of the `spacemacs-org` layer was not updated.

This commit mentions the correct library that is used to customize bullets.


smile13241324 (approved):
:+1:

Issue #6972, opened 2016-08-31
FiraCode “Ligature” integration


Profpatsch:
[FiraCode](https://github.com/tonsky/FiraCode/) is a nice derivation from Fira Mono that adds many ligatures for a nice code look. It uses Unicode ligatures, but Emacs sadly doesn’t support them (yet). I’d like to add a layer to Spacemacs to support that.

There is a [page about integration into Emacs via font-lock](https://github.com/tonsky/FiraCode/wiki/Setting-up-Emacs), and the second example is also [how Hasklig can be made to work with Emacs](https://gist.github.com/Profpatsch/4a05b5ab26d4c09e4b7e5bf735f714f0). I noticed integrating too many ligatures that way makes font rendering very slow.

Profpatsch:
Probably not, no. But it looks awesome.

GregorySchwartz:
I'm not sure if that's the right solution here, as it looks like prettify-symbols-mode alters the number of characters per row, while the ligature rendering preserves that property. For instance, using FiraCode in Konsole would make "->" become an arrow but still take up two characters. Similarly, "::", ">>=", et al. conserve that many characters, keeping the spacing correct. However, if prettify-symbols-mode can keep the character count, then that works too.

hyiltiz:
So to install and use Fira code with spacemacs, does one follow the instructions for emacs from the fira code project site [1], or there is already some mechanism in spacemacs that can use fira code so that the user only needs to "helm xselect font" the Fira Code Retina? I would very much like to use Fira Code in Spacemacs.

[1] https://github.com/tonsky/FiraCode/wiki/Setting-up-Emacs

Profpatsch:
I haven’t yet looked at it any further, maybe the regex-method works, but I suspect it to severely slow down every kind of text display. (It’s essentially running a couple dozen matches on every substring in each buffer…)

Profpatsch:
News!
https://ekaschalk.github.io/post/prettify-mode/

It seems to work with firacode now. If anyone wants to take a look: go for it!

Profpatsch:
Another update: Hasklig [seems to have moved to the FiraCode mechanism with 1.0 as well](https://github.com/i-tu/Hasklig/releases/tag/v1.0-beta). So getting this to work would give us two fonts to choose from.

GregorySchwartz:
[...]

PR #13921, opened 2020-09-06
Add support for font ligatures in Emacs 27 +

AdeAttwood:
Ligature support is provided by https://github.com/mickeynp/ligature.el

Ref: #6972


layers/+fonts/ligatures/packages.el
```diff
@@ -0,0 +1,23 @@
...
+    '((ligature :location (recipe :fetcher github :repo "mickeynp/ligature.el"))))
+  (defconst ligatures-packages '()))
+
+(defun ligatures/init-ligature ()
+  "Initialise the ligatures"
+  (dolist (mode ligature-modes)
+    (ligature-set-ligatures mode ligature-set))
+  (global-ligature-mode t))
+
+;;; packages.el ends here
```
smile13241324, line 23: No package statements please

smile13241324 (changes requested):
Hi @AdeAttwood,

I have double checked your contribution. In principle I like it however I have two issues we need to solve before this is to be merged:
* In your code you activate `ligature-mode` globally, in your documentation however you mention that users can configure modes where it should be activated, can you double check that the configuration as described in the README really works? I would have expected that ligatures are on for all modes independent of the modes I configure.
* I don't think that this feature justifies its own layer. Rather I would like to have this integrated into the `unicode-fonts` layer. There it could be activated with a configuration variable.

AdeAttwood:
@smile13241324 thanks for the feedback.

I have moved the ligature code out of its own layer and moved it under the unicode-fonts

I have checked and the `unicode-fonts-ligature-modes` and `unicode-fonts-enable-ligatures` work as described in the README. I have tested in Emacs 26.3 and 27.1

smile13241324 (approved):
Looks fine now, will be merged soon.

smile13241324:

Issue #13923, opened 2020-09-07
Spacemacs FAQ won't open within Spacemacs


NicholasKirchner:
#### Description :octocat:
Cannot open the Spacemacs FAQ within Spacemacs

#### Reproduction guide :beetle:
- Start Emacs
- Type 'SPC h f'

*Observed behaviour:* :eyes: :broken_heart:
- Error message: "Cannot open load file: No such file or directory, helm-org"
- No FAQ gets loaded

*Expected behaviour:* :heart: :smile:
FAQ should load in a buffer.

#### System Info :computer:
- OS: darwin
- Emacs: 27.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 59f72709b)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(go helm emacs-lisp)
```
- System configuration features: NOTIFY KQUEUE ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS MODULES THREADS JSON PDUMPER

duianto:
The `helm-spacemacs-help-faq` function which is called with `SPC h f` seems to depend on the `org` layer, where `helm-org` is defined.
https://github.com/syl20bnr/spacemacs/blob/59f72709b8c2863bb9182095fbdbc71eba56040f/layers/%2Bemacs/org/packages.el#L20
https://github.com/syl20bnr/spacemacs/blob/59f72709b8c2863bb9182095fbdbc71eba56040f/layers/%2Bemacs/org/packages.el#L88-L91

The pending PR #13924 does fix the error.

But it causes a warning on startup when the `org` layer is enabled:
>                      Warnings:
>                          - More than one init function found for package helm-org. Previous owner was helm,
>                            replacing it with layer org.      

I'm not sure how this should be handled, I'll ask if the other maintainers have any ideas.

duianto:
The following PRs changes have been applied to the `develop` branch:
Add the helm-org package so that helm-spacemacs-faq works #13924

The `helm-org` definition was removed from the `org` layer. It's now only defined in the `helm` layer.

PR #13924, opened 2020-09-07
Add the helm-org package so that helm-spacemacs-faq works

NicholasKirchner:
This is a fix for issue #13923

### DIFF
--- a/layers/+completion/helm/packages.el
+++ b/layers/+completion/helm/packages.el
@@ -21,6 +21,7 @@
         (helm-ls-git :require git)
         helm-make
         helm-mode-manager
+        helm-org
         helm-projectile
         helm-swoop
         helm-themes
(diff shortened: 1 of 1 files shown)


duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/516da8dfbd922c90e2c53ba926851701959e5543

There was a warning with this PR when the `org` layer was enabled:
>                      Warnings:
>                          - More than one init function found for package helm-org. Previous owner was helm,
>                            replacing it with layer org.

Because both the `helm` and `org` layers were defining a init-helm-org function.

JAremko (Spacemacs maintainer) pointed out that the `org` layer works without `helm-org`, but `helm-org` doesn't work without `helm`, therefore the `helm-org` definition was removed from the `org` layer.

PR #13928, opened 2020-09-07
[ipython] Fix bindinding for RET in transient state

bhipple:
In the documentation, it says this performs execute cell and goto next, while
the code was just executing the cell.

This desync could be fixed in either direction, but executing and going to next
feels more ergonomic while in the transient state, since we're jumping around
running rather than editing in that mode, so I've fixed it to do that.


smile13241324 (approved):
:+1:

PR #13929, opened 2020-09-08
[layer/+lang/c-c++] Introduction of shortcut for rtags-symbol-type

fperies:
Introduction of <SPC/M-m> g t shortcut for rtags-symbol-type.
Replacement of previous <SPC/M-m> g t into <SPC/M-m> g u for rtags-dependency-tree


smile13241324 (approved):
I expect that looking up a type is happening more often than displaying the dependency tree, though I am not sure that `u` is really a memonic key type.

I'll merge that for now, but next time please describe in more detail the rational behind a proposed keyword change.

fperies:
> I expect that looking up a type is happening more often than displaying the dependency tree, though I am not sure that `u` is really a memonic key type.

I agree.
`t` makes sense for type - there is no particular reason for `u` : this is where I've found first location available.
This key mapping likely would need some rework someday.

Thanks for the merge.

PR #13935, opened 2020-09-10
[tests] skip spacelpa test on emacs26+win

JAremko:
I feel dirty now X_X


JAremko:
Also we have some kind of locking heisenbug while installing from local mirror. Guess, I'll disable locks during package installation if it gets out of hand.

PR #13938, opened 2020-09-11
[fasd] Fix bug in fasd-find-directory-only

bcc32:
`fasd-find-file` incorrectly interprets a prefix argument of value 1
as equivalent to nil ([issue][0]).

Work around this bug by just passing another positive number instead.
We choose 2.

[0]: https://framagit.org/steckerhalter/emacs-fasd/-/issues/14


smile13241324 (approved):
Lol, nice workaround :)

PR #13941, opened 2020-09-11
[tests] Catching heisenbug

JAremko:
Let's see if it stops those dependency installation freezes and if so come up with a better fix :thinking:


JAremko:
I'll rerun tests 10 times or so. And if they won't bug out then it probably helped...

JAremko:
Looks like it helped. I'll keep the fix for now. In the current form it will only help with test(since I injected the code into testing .lock file).  @syl20bnr  can we disable async-bytecomp for good (at least when installing from a local repo)? Or maybe there is a better way?

PR #13942, opened 2020-09-12
[Python] Add python-shell-send-line function

beyondpie:
Add the python-shell-send-line function as in the other language layer, such as ESS. And bind it also as the same key.

### DIFF
--- a/layers/+lang/python/funcs.el
+++ b/layers/+lang/python/funcs.el
@@ -440,6 +440,15 @@ Bind formatter to '==' for LSP and '='for all other backends."
   (let ((python-mode-hook nil))
     (python-shell-send-region start end)))
 
+(defun spacemacs/python-shell-send-line ()
+	"Send the current line to shell"
+	(interactive)
+	(let* ((python-mode-hook nil)
+				 (start (point-at-bol))
+				 (end (point-at-eol)))
+		(python-shell-send-region start end))
+	)
+
 (defun spacemacs/python-start-or-switch-repl ()
   "Start and/or switch to the REPL."
   (interactive)
--- a/layers/+lang/python/funcs.el
+++ b/layers/+lang/python/funcs.el
@@ -443,10 +443,10 @@ Bind formatter to '==' for LSP and '='for all other backends."
 (defun spacemacs/python-shell-send-line ()
 	"Send the current line to shell"
 	(interactive)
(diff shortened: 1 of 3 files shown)


beyondpie:
I am not familiar with Elisp. So I just use the ESS layer and python layer as a reference, and add this function. 
- should I use `let` or `let*`?  I use `let*` as used in ESS. 
- for the `interactive` part, should I add an "r" argument? I see most of the functions in the Python layer, they use this. But in ESS, not. I choose not to add this.

layers/+lang/python/packages.el
```diff
@@ -354,7 +354,8 @@
         "sf" 'spacemacs/python-shell-send-defun
         "si" 'spacemacs/python-start-or-switch-repl
         "sR" 'spacemacs/python-shell-send-region-switch
-        "sr" 'spacemacs/python-shell-send-region)
+        "sr" 'spacemacs/python-shell-send-region
+				"sl" 'spacemacs/python-shell-send-line)
```
Compro-Prasad, line 358: This line is over indented.

layers/+lang/python/funcs.el
```diff
@@ -440,6 +440,15 @@ Bind formatter to '==' for LSP and '='for all other backends."
   (let ((python-mode-hook nil))
     (python-shell-send-region start end)))
 
+(defun spacemacs/python-shell-send-line ()
+	"Send the current line to shell"
+	(interactive)
+	(let* ((python-mode-hook nil)
+				 (start (point-at-bol))
```
Compro-Prasad, line 447: This line is over indented.

layers/+lang/python/funcs.el
```diff
@@ -440,6 +440,15 @@ Bind formatter to '==' for LSP and '='for all other backends."
...
+				 (end (point-at-eol)))
```
Compro-Prasad, line 448: This line is over indented.

Compro-Prasad:
`let*` would work but `let` would be better.

beyondpie:
@Compro-Prasad 
Thanks for your comments! I've updated the files.

Compro-Prasad:
Add some docs if possible. It would be helpful for starters.

beyondpie:
> Add some docs if possible. It would be helpful for starters.

Done.

smile13241324 (approved):

PR #13945, opened 2020-09-15
Add `spacemacs/python-shell-send-statement' for sending statement

sunlin7:
Hi @smile13241324 ,

The Emacs-27 had a new function `python-shell-send-statement` for sending statement under cursor to python shell, I add it to Spacemacs and compatible with Emacs < 27. Please review it. Thanks.

B.R.
Lin

### DIFF
--- a/layers/+lang/python/packages.el
+++ b/layers/+lang/python/packages.el
@@ -351,6 +351,8 @@
         "ri" 'spacemacs/python-remove-unused-imports
         "sB" 'spacemacs/python-shell-send-buffer-switch
         "sb" 'spacemacs/python-shell-send-buffer
+        "sE" 'spacemacs/python-shell-send-statement-switch
+        "se" 'spacemacs/python-shell-send-statement
         "sF" 'spacemacs/python-shell-send-defun-switch
         "sf" 'spacemacs/python-shell-send-defun
         "si" 'spacemacs/python-start-or-switch-repl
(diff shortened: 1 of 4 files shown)


layers/+lang/python/packages.el
```diff
@@ -351,6 +351,8 @@
         "ri" 'spacemacs/python-remove-unused-imports
         "sB" 'spacemacs/python-shell-send-buffer-switch
         "sb" 'spacemacs/python-shell-send-buffer
+        "sE" 'spacemacs/python-shell-send-statement-switch
```
smile13241324, line 354: This binding should also be documented.

smile13241324 (changes requested):
Nice addition, but one of the two new bindings is not documented.
When the documentation is fixed it will be merged.

layers/+lang/python/packages.el
sunlin7: Hi @smile13241324 , Yes, you're right, I add it into the Readme.org now. 
Please review again.
Thank you

sunlin7: https://github.com/syl20bnr/spacemacs/commit/deb13cd2faf726d275b86b9622cca4f413970b1f#diff-069ce66d5b3d471d2ab7f214fab50ae7R446

smile13241324 (approved):
Looks good, thanks for fixing :+1:

sunlin7:
@smile13241324 thank you.

PR #13950, opened 2020-09-16
[pass] Add support for auth-source-pass and password-store-otp.

whirm:
----

#


smile13241324 (changes requested):
The code is ok, however the documentation is missing.
Please document the new keybindings for `password-store-otp`.
Also at least a basic introduction or a link to more information about how to use `auth-source-pass` would be nice.

smile13241324 (approved):

Issue #11938, opened 2019-02-03
`Error callling diff` message when syncing configuration layers: SPC f e R


duianto:
#### Description :octocat:

A `Error callling diff` error message is shown when the configuration layers are synced `SPC f e R`.
(Yes with 3 l's)

#### Reproduction guide :beetle:

- Start Spacemacs
-  Press: `SPC f e R` (which calls: `dotspacemacs/sync-configuration-layers`)

*Observed behaviour:* :eyes: :broken_heart:

Everything seems to be reloaded as expected. But the messages buffer shows:
```
Error callling git diff:

```
The minibuffer has the height of two lines (the second line is empty), so it seems like a newline is added after the error message.

The messages buffer shows:
```
Loading c:/Users/username/.spacemacs...done
Loading c:/Users/username/.spacemacs...done
Loading c:/Users/username/.spacemacs...done
Done.
Error callling git diff:

```

*Expected behaviour:* :heart: :smile:

The error message should probably not be there.

#### Notes :notebook:

The error message is defined in the `git-gutter-plus` package.
In the function:
```elisp
(defun git-gutter+-diff (curfile)
  (let ((args (git-gutter+-diff-args curfile))
        (file (buffer-file-name))) ;; for tramp
    (with-temp-buffer
      (if (git-gutter+-insert-git-output args file)
          (progn (goto-char (point-min))
                 (let ((diff-header (git-gutter+-get-diff-header))
                       (diffinfos   (git-gutter+-get-diffinfos)))
                   (list diff-header diffinfos)))
        (message "Error callling git diff:\n%s" (buffer-string))
        nil))))
```
source: https://github.com/nonsequitur/git-gutter-plus/blob/b7726997806d9a2da9fe84ff00ecf21d62b6f975/git-gutter%2B.el#L199

There's an open issue from the git gutter plus repository that mentions that error message:
Getting 'Error callling diff" errors with latest magit #41
https://github.com/nonsequitur/git-gutter-plus/issues/41

It just states that:
>Apart from the spelling error in "callling", this seems to break everything.
package version:
magit-20180409.1612

#### Possible source ℹ️ 

Spacemacs uses `git-gutter+` in the `version-control` layer:
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+source-control/version-control/packages.el#L202

This might be the command with the issue:
`        (add-hook 'magit-pre-refresh-hook 'git-gutter+-refresh)`

because the function: `git-gutter+-refresh`
[...]

PR #13958, opened 2020-09-17
Fix git-gutter+ refresh on Magit refresh

claritystorm:
Fixed `git-gutter+` refresh when a buffer is staged or unstaged in Magit.

### DIFF
--- a/layers/+source-control/version-control/funcs.el
+++ b/layers/+source-control/version-control/funcs.el
@@ -152,3 +152,7 @@ the number of conflicts detected by `smerge-mode'."
   (interactive)
   (setq spacemacs--smerge-ts-full-hint-toggle
         (not spacemacs--smerge-ts-full-hint-toggle)))
+
+(when (configuration-layer/package-used-p 'git-gutter+)
+  (defun spacemacs//git-gutter+-refresh-in-all-buffers ()
+    (git-gutter+-in-all-buffers (when git-gutter+-mode (git-gutter+-refresh)))))
--- a/layers/+source-control/version-control/packages.el
+++ b/layers/+source-control/version-control/packages.el
@@ -205,10 +205,8 @@
     (progn
       ;; If you enable global minor mode
       (when version-control-global-margin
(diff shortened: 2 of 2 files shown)


smile13241324 (approved):
Looks ok, will be merged soon

smile13241324:

smile13241324:
I had to revert your PR, it has negative side effects when I stage a changed file and then commit. In this case I see the following error in the message buffer:
```
Error callling git diff:
```
Can you have another look?

smile13241324 (changes requested):
Please see my previous commit for the side effects

duianto:
I've got an open issue about that error message here: 
`Error callling diff` message when syncing configuration layers: SPC f e R #11938

There's also an open issue about it on the `git-gutter-plus` repo, but it doesn't say much:
Getting 'Error callling diff" errors with latest magit · Issue https://github.com/nonsequitur/git-gutter-plus/issues/41

So that error might already have been there occurring in some specific situations, this PR just made it even more obvious.

claritystorm:
This error is reproduced by calling `git-gutter+-refresh` from a buffer visiting `.spacemacs`.

smile13241324 (approved):
Looks ok, when this is a known issue not coming from your change, I will merge it again.

smile13241324:
Just retested now the issue is gone, thanks for committing again @claritystorm.

MHOOO:
For anyone who stumples accross performance issues when staging:
I have serious lag when staging changes in the magit status buffer (1-2s when pressing "s"). Removing `spacemacs//git-gutter+-refresh-in-all-buffers` from `magit-pre-refresh-hook` fixed that for me.

smile13241324:
@MHOOO well looks like the issue is the number of buffers open in your emacs session. In the current implementation all of these are read which could take some seconds depending on the number of open buffers.

Maybe we should define an configurable max number of buffers to run through. Or change the implementation not to have to go through all open buffers for this.

I think this would justify an issue to discuss a possible solution.

PR #13960, opened 2020-09-17
[org] support rich insert of code from other buffers (org-rich-yank)

Kazark:
See the animation in the package's [README](https://github.com/unhammer/org-rich-yank/blob/master/README.org). If necessary, I can put this package behind a switch like others are. However it's a small package (133 sloc) which integrates well with what's there so I didn't default to that.

I like that Spacemacs provides a very rich, full-featured experience via layers and I think this fills a niche. I use this package myself regularly when I am studying and analyzing a problem, and want to snip pieces of code out of various files in a code base and put them into one place with commentary, TODOs, &c.


Kazark:
Is there any way to see why some checks failed?

smile13241324 (approved):
Great, contribution :+1:, will be merged soon.

PR #13962, opened 2020-09-17
[core] add keybinding hdT for describe-text-properties

Kazark:
This is to bundle it with other describe functions. The hope is to make the help -> describe menu more complete.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -518,6 +518,8 @@ Other:
     terminal mode, but no longer in GUI mode. (emacs18)
 *** Core changes
 - Improvements:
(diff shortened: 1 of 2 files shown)


Kazark:
Is there any way to see why some checks failed?

smile13241324:
Unfortunately not without accessing circleCI directly. Currently we have a lot of issues with installing dependencies reliably. So as only an install job fails it is highly unlikely that this comes from your change.

smile13241324 (approved):
Looks ok, just noticed that the documentation is missing for the new binding but I can add it while I am merging it.

PR #13949, opened 2020-09-15
[core] evilify Custom-mode if dotspacemacs-editing-style is 'vim

Kazark:
This has been requested many times but never implemented.

Fixes #6699
Fixes #5668
Fixes #7583
Fixes #13309

The comments describe the rational behind the bindings. Please let me know if you disagree with my choices at all as I am open to changing; this was my attempt to put something together that didn't entirely disregard the default bindings, while making a lot of sense from an Evil perspective.

PR #13963, opened 2020-09-18
[core] Fix evil keybindings when a link is selected in Custom-mode

Kazark:
The links use a separate keymap bound directly to the widget, circumventing the existing evil keybindings.

This fixes the [bug](https://github.com/syl20bnr/spacemacs/pull/13949#issuecomment-694685304) found by @duianto.

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/packages.el
+++ b/layers/+spacemacs/spacemacs-defaults/packages.el
@@ -102,8 +102,14 @@
   (evil-define-key 'normal cus-edit-mode-map (kbd "q") 'Custom-buffer-done)
   (evil-define-key 'normal cus-edit-mode-map (kbd "u") 'Custom-goto-parent)
   ;; `RET' does not work well in the search field. Fix:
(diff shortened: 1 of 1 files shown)


duianto:
`SPC` and `n` works now 👍 

However I noticed that both before and after applying this PR, 
`q` and `u` only work on links to quit and goto parent.

I haven't used customize a lot so I might be missing something, 
but I'm not finding a key map that starts with: `cus-edit` or even `cus-`
in the `SPC h d K` HELM Describe Keymap buffer.

Kazark:
@duianto thanks; I fixed that problem too, in 4b51430.

The problem that is inscrutable to me is: if you enter text into the search bar, and then navigate way, things are in a messed-up state... unless you go to some other buffer and then come back, and then things are okay. ???

I don't think that a regression in this PR though, but a case I didn't test at first.

smile13241324:
Now it works.

duianto:
Confirmed, thanks.

PR #13964, opened 2020-09-18
Fix up markdown mode specifiers

bcc32:
Simplify the regexps, and make sure they are anchored.


smile13241324 (approved):
Nicer than before, will be merged soon.

PR #13968, opened 2020-09-21
[core] add `spacemacs-scratch-mode-hook`

rayw000:
### What Changed

Add hook `spacemacs-scratch-mode-hook` which runs on buffer `*scratch*` after applying `dotspacemacs-scratch-mode`.

### Why

Sometimes we may need to do something to buffer `*scratch*` after it is loaded, for example enabling a minor mode which only makes sense in `*scratch*` buffer. 

I tried to do this by creating a custom layer, but layers are loaded before https://github.com/syl20bnr/spacemacs/blob/dd94a13a48b40562a76c4aca1073796619077d0f/core/core-spacemacs.el#L233

That means my buffer specified enabled minor modes are turned off after this line.

That's why I'd like to add a `hook` to run after `dotspacemacs-scratch-mode` was enabled, no matter what the actually `major-mode` is.

Or, if you have some way better to achieve this, please let me know. 😃

Thanks!

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -595,6 +595,8 @@ Other:
     (thanks to duianto)
   - Add =emacs-version= to the default value for
     =dotspacemacs-emacs-dumper-dump-file= (thanks to Evan Klitzke)
+  - Add =spacemacs-scratch-mode-hook= to run functions
+    after =spacemacs-scratch-mode= is applied to =*scratch*= buffer (thanks to rayw000)
 - Fixes:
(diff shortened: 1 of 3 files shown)


smile13241324 (approved):
Took me a while till I came around to reviewing this. After playing a bit around with it I would now agree that this change is indeed useful and should be merged.

smile13241324:

rayw000:
Thanks a lot! And then I could open another PR to add a scratch buffer config layer which depends this one later. 😄

PR #13969, opened 2020-09-21
Add C-RET keybinding for ivy-alt-done

dalanicolai:
Currently there is only the option to use `ivy-alt-done` by rebinding `ivy-done` to `C-j` with setting `ivy-ret-visits-directory` to `t`.
In ivy, `ivy-alt-done` is bound to `C-j` which of course in Spacemacs is used for selecting the next entry (in ivy minibuffer). However, `C-RET` is still free, so this PR binds `ivy-alt-done` to `C-RET`.

I prefer `ivy-alt-done` over `TAB-TAB` to browse into directories. Also, this setting enables to go into directories with Ivy (instead of opening the directory in dired) when using e.g the `fasd` package.

### DIFF
--- a/layers/+completion/ivy/README.org
+++ b/layers/+completion/ivy/README.org
@@ -77,17 +77,18 @@ In case you don't want to read everything, at least familiarise with
 
 Some useful key bindings are presented in the following table.
 
-| Key binding | Description                                                                                          |
-|-------------+------------------------------------------------------------------------------------------------------|
-| ~RET~       | call default action on current candidate                                                             |
-| ~M-RET~     | the same as ~RET~ but doesn't close completion minibuffer                                            |
-| ~C-M-j~     | use current input immediately (this can be used to create a new file in Find File)                   |
-| ~TAB~       | complete partially                                                                                   |
-| ~M-o~       | show the list of valid actions on current candidate (then press any of described keys to execute it) |
-| ~C-M-o~     | the same as ~M-o~ but doesn't close completion minibuffer                                            |
-| ~C-'~       | use avy to quickly select completion on current page (sometimes faster than using arrows)            |
-| ~<ESC>~     | close minibuffer                                                                                     |
-| ~C-M-k~     | kill buffer (in =ivy-switch-buffer= (~SPC b b~))                                                     |
+| Key binding | Description                                                                                                                                           |
+|-------------+-------------------------------------------------------------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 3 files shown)


layers/+completion/ivy/README.org
```diff
@@ -77,17 +77,18 @@ In case you don't want to read everything, at least familiarise with
...
-| ~C-M-j~     | use current input immediately (this can be used to create a new file in Find File)                   |
-| ~TAB~       | complete partially                                                                                   |
-| ~M-o~       | show the list of valid actions on current candidate (then press any of described keys to execute it) |
-| ~C-M-o~     | the same as ~M-o~ but doesn't close completion minibuffer                                            |
-| ~C-'~       | use avy to quickly select completion on current page (sometimes faster than using arrows)            |
-| ~<ESC>~     | close minibuffer                                                                                     |
-| ~C-M-k~     | kill buffer (in =ivy-switch-buffer= (~SPC b b~))                                                     |
+| Key binding | Description                                                                                                                                         |
+|-------------+-----------------------------------------------------------------------------------------------------------------------------------------------------|
```
smile13241324, line 81: This looks like the formatting went completely crazy, please double check

smile13241324 (changes requested):
I think your change is reasonable, though I have found some formatting issues in the README which should be fixed before this can be merged.

layers/+completion/ivy/README.org
dalanicolai: Okay, so this generally happens when I use `org-cycle` (i.e. `TAB` key) to automatically format the table. So I should always correct the formatting manually in text-mode? Or do you have some advice for how to do it automatically?

smile13241324: Nope, I write the documentation in vanilla vim, as my spacemacs heavily reformats buffers before saving, however as the documentation is also used as basis for generating html and markdown we have to follow certain formatting rules.

dalanicolai: Ah thanks! Will use vanilla Emacs for that then :+1:

smile13241324:
Fixed the conflicts and reformatted the table after the merge. 

layers/+completion/ivy/README.org
duianto: Pressing `TAB` in a key binding table has worked for me in Spacemacs,
to automatically fit the table to the longest description line.

dalanicolai: Seems like with a new/clean dotfile it works correctly indeed.

PR #13972, opened 2020-09-22
Adding magit-delta option for magit

dcluna:
This PR enables [magit-delta](https://github.com/dandavison/magit-delta) with the new `git-enable-magit-delta-plugin` option.

For example usage, one could write this in their `.spacemacs`:

```emacs-lisp
  '( 
    ;; rest of packages here
    (git :variables git-enable-magit-delta-plugin t))
```

### DIFF
--- a/layers/+source-control/git/packages.el
+++ b/layers/+source-control/git/packages.el
@@ -25,6 +25,7 @@
         (helm-git-grep :requires helm)
         (helm-gitignore :requires helm)
         magit
+        (magit-delta :toggle git-enable-magit-delta-plugin)
         magit-gitflow
         magit-section
         magit-svn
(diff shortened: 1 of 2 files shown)


layers/+source-control/git/packages.el
```diff
@@ -25,6 +25,7 @@
         (helm-git-grep :requires helm)
         (helm-gitignore :requires helm)
         magit
+        (magit-delta :if git-enable-magit-delta-plugin)
```
smile13241324, line 28: This will not prevent installing this package, the keyword you need is `:toggle` like in `(dash-at-point :toggle (spacemacs/system-is-mac))`. Just tested this and without me setting the new variable the package has been installed.

smile13241324 (changes requested):
Your package declaration is not correct, as it is right now the package will always be installed. Please check my comment for details.

layers/+source-control/git/packages.el
dcluna: Hi @smile13241324 , I hope to have addressed your comments. Would you mind reviewing and merging/cherry-picking when you have a chance?

Thanks,

smile13241324 (approved):
Looks good, I'll give it a quick test and merge when everything is ok.

smile13241324:

PR #13973, opened 2020-09-22
Don't let magit-gitflow mode overwrite `C-f` key binding.

rayw000:
In `*magit*` buffer, minor mode `magit-flow` overwrites key `C-f` from `forward-char` to `magit-gitflow-popup`.

It may be confusing that `C-a` `C-e` `C-b` are all cursor motion except `C-f`.

This PR makes `magit-gitflow-popup` only bound to `%`.

Ref:
https://github.com/jtatarik/magit-gitflow/issues/11
https://github.com/jtatarik/magit-gitflow/blob/cc41b561ec6eea947fe9a176349fb4f771ed865b/magit-gitflow.el#L43-L48


smile13241324 (approved):
Good addition, will be merged soon.

PR #13976, opened 2020-09-23
Add instruction how to use ranger by default

dalanicolai:
The ranger layer does not describe yet how to actually use ranger by default, while also the [instructions for the original ranger package](https://github.com/ralesi/ranger.el#setting-as-default-directory-handler) don't work (i.e. setting `ranger-override-dired` via customize does not work in Spacemacs).

I created a feature branch but forgot to check it out (still need to get used to magit). I hope it is fine that I create this PR now directly (and be more careful/wiser next time).

### DIFF
--- a/layers/+tools/ranger/README.org
+++ b/layers/+tools/ranger/README.org
@@ -18,11 +18,17 @@ To use this configuration layer, add it to your =~/.spacemacs=. You will need to
 add =ranger= to the existing =dotspacemacs-configuration-layers= list in this
 file.
 
-To default with preview enabled when entering ranger:
+To use ranger/deer by default set ~ranger-override-dired~ to ~ranger/deer~ like
+shown in the example below (setting this via customize as explained by the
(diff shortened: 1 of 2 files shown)


layers/+tools/ranger/README.org
```diff
@@ -18,11 +18,19 @@ To use this configuration layer, add it to your =~/.spacemacs=. You will need to
 add =ranger= to the existing =dotspacemacs-configuration-layers= list in this
 file.
 
-To default with preview enabled when entering ranger:
+To
```
smile13241324, line 21: This does not look right

smile13241324 (changes requested):
I like the addition to the documentation, however I have found a row which doesn't look like intended. Please double check and mention me when I should review again.

dalanicolai:
@smile13241324 Sorry for that, I cleared it up now, so I guess it can be merged.

smile13241324 (approved):
Looks good now, will be merged soon.

PR #13977, opened 2020-09-23
literate-haskell-mode has been renamed to haskell-literate-mode

michaelpj:
See https://github.com/haskell/haskell-mode/commit/7ccb1ab0cc460bbc84e0e4b35719839f2a9dd598.

We should keep the old assignment around for a bit to support people on
old versions of `haskell-mode`.


smile13241324 (approved):
Thanks for finding and fixing this :+1:

PR #13980, opened 2020-09-24
[core] Enable pdumper by default if the Emacs build supports it

eklitzke:
This enables the portable dumper by default if the Emacs instance supports it. If the user is running an older Emacs version, or an Emacs 27 build that was compiled without pdumper support, the Spacemacs configuration will continue to work as before without attempting to pdump. I think this is warranted because:
 * Emacs 27 is now the current stable Emacs major version
 * The portable dumper is no longer considered experimental, it's enabled by default when building Emacs 27+

Besides updating the default definition of `dotspacemacs-enable-emacs-pdumper` to `t`, I've modified `spacemacs/emacs-with-pdumper-set-p` to also check that `dump-emacs-portable` is a bound function (which will only be the case on an Emacs build that supports pdumping).

Of course, Spacemacs users with an old `.spacemacs` file may continue to have the value of `dotspacemacs-enable-emacs-pdumper` set as `nil` until they diff/sync with the upstream template, but that's expected (and desired).

**NOTE:** I'm not sure to do with the file `EXPERIMENTAL.org`, which has the legacy instructions for building Emacs with pdumper support. It arguably could just be removed, but it also has some still relevant instructions about using the pdumper mode (e.g. aliasing emacs to `emacs --dump-file=path/to/spacemacs.dump`). Maybe we need to remove this file and update the README?

### DIFF
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -58,10 +58,9 @@ exists. Otherwise, fallback to ~/.spacemacs"))
 `+distributions'. For now available distributions are `spacemacs-base'
 or `spacemacs'.")
 
-(defvar dotspacemacs-enable-emacs-pdumper nil
-  "If non-nil then enable support for the portable dumper. You'll need
-to compile Emacs 27 from source following the instructions in file
-EXPERIMENTAL.org at to root of the git repository.")
+(defvar dotspacemacs-enable-emacs-pdumper t
+  "If non-nil then enable support for the portable dumper if this Emacs build
+supports portable dumping (requires Emacs 27 or later).")
 
 (defvar dotspacemacs-emacs-pdumper-executable-file "emacs"
   "File path pointing to emacs 27 or later executable.")
--- a/core/core-dumper.el
+++ b/core/core-dumper.el
@@ -100,6 +100,7 @@ the end of the loading of the dump file."
 (defun spacemacs/emacs-with-pdumper-set-p ()
   "Return non-nil if a portable dumper capable emacs executable is set."
   (and dotspacemacs-enable-emacs-pdumper
+       (fboundp 'dump-emacs-portable)
        (file-exists-p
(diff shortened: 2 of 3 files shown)


emacs18:
Is the following in `EXPERIMENTAL.org` still valid?

**Note: This seems to not be needed anymore. For reference we still keep it in
the guide.**

We need to increase the number of `remembered_data` slots in `src/pdumper.c`, we
double the number of slots by replacing 32 with 64:

    static struct
    {
      void *mem;
      int sz;
    } remembered_data[64];

emacs18:
That is, is the patch really not needed anymore?

robbyoconnor:
There's still a bug where the home buffer gets wonky

smile13241324:
I am not sure that this is a good idea, though I have never used pdumper as I have just upgraded to emacs 27 may this not break older loading behaviour? Maybe @syl20bnr can help here, I think he is using pdumper in his config.

emacs18:
I used pdumper apart from spacemac, i.e., I dumped an image without using spacemacs, then I used it with or without spacemacs.  When used with spacemacs my emacs startup time went down from 7 seconds to 2 seconds.

However I have never been able to successfully create dumped image following spacemacs instructions.  Even after I changed ~/.emacs.d symlink to point to spacemacs directory, I utterly failed to get pdumper to create an image.

Also now that emacs may be getting GccEmacs merged from feature branch to master branch in the near future, I'm not sure how this can impact pdumper feature.  I don't know how the two may interact.  So I would also be concerned if pdumper were to be enabled by default.  It would not matter to me too much since I already use my own git branch off of develop branch, i.e., I would just disable pdumper on my branch.

robbyoconnor:
In addition to everything, pdumper still has odd regressions.

smile13241324:
I am sorry but given from the response in the ticket I think it is not a good idea to set pdumper as default just yet. When it is more widely used and more stable we can talk about this again.

PR #13981, opened 2020-09-24
[vim][new layer] add evil-easymotion layer

thanhvg:
layer evil-easymotion for https://github.com/PythonNut/evil-easymotion
bind prefix to `gs`

copy pasta from Doom project https://github.com/hlissner/doom-emacs

### DIFF
--- a/layers/+vim/evil-easymotion/README.org
+++ b/layers/+vim/evil-easymotion/README.org
@@ -1,0 +1,27 @@
+#+TITLE: evil-better-jumper layer
+#+TAGS: layer|vim
+
+* Table of Contents                                       :TOC_4_gh:noexport:
+- [[#description][Description]]
+  - [[#features][Features:]]
+- [[#install][Install]]
+- [[#key-bindings][Key bindings]]
+
+* Description
+This layer adds support for evil-easymotion. It is a port from [[https://github.com/hlissner/doom-emacs][Doom project]].
+
+** Features:
+=evil-easymotion= comes with predefined easymotions for all basic motions that
+evil defines by default, and provides key bindings for these using the
+standard evil keys in the built-in keymap =evilem-map=.
+
+* Install
+To use this configuration layer, add it to your =~/.spacemacs=. You will need to
+add =evil-easymotion= to the existing =dotspacemacs-configuration-layers=
+list in this file.
+
+* Key bindings
+
+| Key Binding | Description       |
+|-------------+-------------------|
+| ~gs~        | easymotion prefix |
--- a/layers/+vim/evil-easymotion/packages.el
+++ b/layers/+vim/evil-easymotion/packages.el
@@ -1,0 +1,43 @@
+;;; packages.el --- evil-easymotion layer packages file for Spacemacs.
+;;
+;; Copyright (c) 2012-2020 Sylvain Benner & Contributors
+;;
+;; Author: Thanh Vuong <thanhvg@gmail.com>
+;; URL: https://github.com/thanhvg
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+;;; Commentary:
+
+;; Port from Doom project modules/editor/evil/config.el
+
+;;; Code:
+
+(defconst evil-easymotion-packages
+  '(evil-easymotion))
+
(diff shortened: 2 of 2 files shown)


duianto:
This has a lot of nice motion commands.

One limitation seems to be that magit uses `gs` for `magit-jump-to-staged`, so these won't work for jumping around in the magit buffer. But it's probably outweighed by the easy key binding for jumping around everywhere else.

We should define a named function for:
```elisp
    (define-key evilem-map (kbd "SPC") (lambda (&rest _)
                                   (interactive)
                                   (let ((current-prefix-arg t)) (evil-avy-goto-char-timer))))
```
so that it describes what the command does instead of just listing: `SPC -> lambda` in the which-key popup.

Since it seems to limit the motions to the current buffer, maybe it could be called: `buffer-evil-avy-goto-char-timer` or something better.

### Prefixes
Add a prefix names for `gs`, `gs[` and `gs]`:
`backward section` and `forward section` would match the order that the words appear in the commands:
`gs[`
```
[ -> evilem-motion-backward-section-begin
] -> evilem-motion-backward-section-end
```
`gs]`
```
[ -> evilem-motion-forward-section-end
] -> evilem-motion-forward-section-begin
```

Or would it be more descriptive to list `section` first?

Could there be a risk of confusion if someone tries to remember the function name (possibly by calling the command directly) and the prefix word order is opposite to the command name.

What could the prefix name be for: `gsg`
It contains these keys and commands:
```
e -> evilem-motion-backward-word-end
j -> evilem-motion-next-visual-line
k -> evilem-motion-previous-visual-line
E -> evilem-motion-backward-WORD-end
```

This seems to work for adding the prefix names:
```elisp
  (which-key-add-key-based-replacements
    "gs"  "evil-easymotion"
    "gs[" "section backward"
    "gs]" "section forward")
```

### changelog
Add a changelog.develop entry:
```diff
1 file changed, 2 insertions(+)
CHANGELOG.develop | 2 ++

modified   CHANGELOG.develop
@@ -435,6 +435,8 @@ sane way, here is the complete list of changed key bindings
 - transmission (thanks to Eugene Yaremenko)
 - web-beautify (thanks to Sylvain Benner and Juuso Valkeejärvi)
 - xclipboard (thanks to Charles Weill, Sylvain Benner, and Alex Palaistras)
+**** Vim
+- evil-easymotion (thanks to Thanh Vuong)
 **** WEB services
 - confluence
 *** Renamed layers
```

smile13241324:
I really like the proposed changes, I am a heavy vim user myself, however I wonder, if we move this into a separate layer we:
   - introduce another 1 package layer 
   - as well as risk that most vim user will never see this, if they don't look for specific improved bindings

Therefore I would propose to move this into the spacemacs-editing layer and have a new setting in the dotfile for this. Especially if this has no side effects I would activate it by default as long as the editing style is not `emacs`. We could add the key binding documentation in the general doc file then.

@duianto what's your opinion?
@thanhvg would you be willing to do that change? No harm when not, in this case I would just merge your layer and refactor it myself then.

thanhvg:
Hi @smile13241324 I agree with you suggestion and I also think @duianto's suggestions should be included. But sorry I lost the will to do it so if you or anyone can take over this PR that would be great.

Thanks

duianto:
>Therefore I would propose to move this into the spacemacs-editing layer and have a new setting in the dotfile for this.

👍

smile13241324:
Thank you for contributing to spacemacs 💜, I have cherry picked your commit into develop and did the migration into the core.
In addition I have fixed the missing which key bindings.

@duianto we need to be very careful with `which-key-add-key-based-replacements` it comes with a ton of side effects, i.e. binding a prefix for `gr` binds it also for magit obscuring the description for `magit-refresh`. A much better alternative is the new keybased api `which-key-add-keymap-based-replacements` which makes such conflicts very unlikely. In addition it is much faster than the old one as no texts must be compared at run time.

The only side effect I have still seen is that `magit-mode-map` seems to be inheriting from the general `evil-motion-state-map` which I have no clue why. My final fix was now to manually remove the inherited bindings from `magit-mode-map` after its config is done. I have also fixed the missing prefixes for `gf` and `gp` in `magit-status-mode-map`.

Now all the which-key bindings should be available

Issue #12767, opened 2019-09-24
[protobuf] Auto-detect protobuf files and suggest to install the protobuf layer


Asalle:
#### Description :octocat:
Auto-detect protobuf files and suggest to install the protobuf layer

#### Reproduction guide :beetle:
- Start Emacs
- Open a .proto file

*Observed behaviour:* :eyes: :broken_heart:
- See it opened in plan text

*Expected behaviour:* :heart: :smile:
- See a message that asks if I want to install the protobuf layer

#### System Info :computer:
- OS: gnu/linux
- Emacs: 26.2
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 9f32750f6)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(javascript markdown
            (go :variables go-format-before-save t go-use-golangci-lint t go-backend 'lsp godoc-at-point-function 'godoc-gogetdoc)
            helm auto-completion git org c-c++
            (shell :variables shell-default-height 30 shell-default-position 'bottom shell-default-shell 'ansi-term)
            better-defaults syntax-checking lsp)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND DBUS GSETTINGS GLIB NOTIFY ACL LIBSELINUX GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XDBE XIM MODULES THREADS XWIDGETS LCMS2

duianto:
The following PRs changes have been applied to the `develop` branch:
Lazy install protobuf layer for .proto files #13983

PR #13983, opened 2020-09-25
Lazy install protobuf layer for .proto files

duianto:
Resolves: [protobuf] Auto-detect protobuf files and suggest to install the protobuf layer #12767


smile13241324 (approved):
:+1: , did not know we had a file for this :)

PR #13984, opened 2020-09-25
Improving the tabs key bindings

sdwolfz:
There's a problem with the Control + Shift keys like `C-{`on terminal, Shift is ignored and we end up typing `C-[` instead. This PR will try to define better keys for tab manipulation.

Ideas welcome.

### DIFF
--- a/layers/+emacs/tabs/packages.el
+++ b/layers/+emacs/tabs/packages.el
@@ -31,9 +31,17 @@
         (centaur-tabs-group-by-projectile-project)
       (centaur-tabs-group-buffer-groups))
     (centaur-tabs-mode t)
+
+    (spacemacs/declare-prefix "bt" "tabs")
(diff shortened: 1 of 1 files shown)


WarFox:
Thanks @sdwolfz 

On GUI:
`M-{` and `M-}` is working as expected in GUI. However I noticed in `org-mode` buffers it is navitaging through headings instead. I was going backwards and forwards using `M-{` and `M-}`, when it reached an `org-mode` buffer it stopped moving tabs and started going through headings instead.

Non GUI:
Works on `i-term` (with org-mode exception)
Works on `i-term + tmux` (with org-mode-exception)

Doesn't work in `alacritty`
Doesn't work in `alacritty + tmux`

sdwolfz:
Closing this in favor of https://github.com/syl20bnr/spacemacs/pull/13995, let's continue the discussion there.

PR #13991, opened 2020-09-28
[clojure] key binding for cider eval list around point

jr0cket:
New command added to CIDER in version 0.26 to evaluate the list around the
current point, different to evaluate sexp around point which also evaluates
symbols.

This command is useful for evaluating nested expressions without having to jump to the end of the expression.  It will evaluate function calls as well as evaluate collections as functions (lists, maps, vectors, sets).

https://github.com/clojure-emacs/cider/pull/2881

Thank you <3


duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/f2c075b67304a90e5224eb93172dc249ec06d901

PR #14000, opened 2020-09-30
Replace spacemacs/swiper with better native swiper functions

dalanicolai:
Spacemacs uses spacemacs/swiper functions that use the less sophisticated
spacemacs//counsel-current-region-or-symbol instead of ivy's native
ivy-thing-at-point function. This commit removes those spacemacs functions and
rebinds the shortcuts to their better native equivalents.

Again forgot to make a fix branch because I forgot because I am always fighting
with git. Anyway, again I will try to be more alert next time.


smile13241324 (approved):
Looks ok, will be merged soon.

PR #14003, opened 2020-10-01
Update README.org

kimchichef:
evilifcation -> evilification


robbyoconnor:
@syl20bnr @duianto mark this as invalid so they don't get credit for this to pad their Hacktoberfest PR count

PR #14006, opened 2020-10-02
[org] Add major mode keybindings for org-jira

klochowicz:
org-jira is typically used from org major mode - new keybindings add a more
convenient way of invoking its functions.

The new org-jira submenu is available under "mj" - using the same pattern as
org-trello.

The existing keybindings (<spc> a o J ....) are inconvenient to use due to uppercase "J" and involving more keypresses.

PS I have considered using "j" instead of "mj", but I didn't want to potentially hog up keybinding that might be use for other extensions, e.g. org-journal.

### DIFF
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -644,7 +644,28 @@ Headline^^            Visit entry^^               Filter^^                    Da
         "aoJsc" 'org-jira-create-subtask
         "aoJsg" 'org-jira-get-subtasks
         "aoJcu" 'org-jira-update-comment
-        "aoJtj" 'org-jira-todo-to-jira))))
+        "aoJtj" 'org-jira-todo-to-jira)
+      (spacemacs/declare-prefix-for-mode 'org-mode "mmj" "jira")
+      (spacemacs/declare-prefix-for-mode 'org-mode "mmjp" "projects")
+      (spacemacs/declare-prefix-for-mode 'org-mode "mmji" "issues")
+      (spacemacs/declare-prefix-for-mode 'org-mode "mmjs" "subtasks")
+      (spacemacs/declare-prefix-for-mode 'org-mode "mmjc" "comments")
+      (spacemacs/declare-prefix-for-mode 'org-mode "mmjt" "todos")
+      (spacemacs/set-leader-keys-for-major-mode 'org-mode
+        "mjpg" 'org-jira-get-projects
+        "mjib" 'org-jira-browse-issue
+        "mjig" 'org-jira-get-issues
+        "mjih" 'org-jira-get-issues-headonly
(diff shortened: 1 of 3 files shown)


layers/+emacs/org/packages.el
```diff
@@ -621,6 +621,7 @@ Headline^^            Visit entry^^               Filter^^                    Da
 
 (defun org/init-org-jira ()
   (use-package org-jira
+    :after org
```
duianto, line 624: This causes the `SPC a o J` key bindings not to load on startup.

And it won't be needed for the new key bindings org mode key bindings, because this:
`(spacemacs/set-leader-keys-for-major-mode 'org-mode`
probably means that the key bindings will be defined when `org-mode` loads.

layers/+emacs/org/packages.el
```diff
@@ -644,7 +645,30 @@ Headline^^            Visit entry^^               Filter^^                    Da
...
+        "mjiu" 'org-jira-update-issue
+        "mjiw" 'org-jira-progress-issue
+        "mjir" 'org-jira-refresh-issue
+        "mjic" 'org-jira-create-issue
+        "mjiy" 'org-jira-copy-current-issue-key
+        "mjsc" 'org-jira-create-subtask
+        "mjsg" 'org-jira-get-subtasks
+        "mjcu" 'org-jira-update-comment
+        "mjtj" 'org-jira-todo-to-jira)
+
```
duianto, line 670: Remove the empty line.

And:
>Place all trailing parentheses on a single line instead of distinct lines.
```
;; good; single line
(when something
  (something-else))

;; bad; distinct lines
(when something
  (something-else)
)
```
source: https://github.com/bbatsov/emacs-lisp-style-guide

duianto:
Since the key bindings are the same except for the prefixes.

Instead of listing the same commands twice in the readme key binding table:
https://github.com/syl20bnr/spacemacs/tree/develop/layers/+emacs/org#org-jira

One solution could be to specify both prefixes before the key binding table.
Like this (or something better):

[Key binding prefixes]:
- `SPC a o J` (anywhere)
- `SPC m m j` (in an org-mode buffer)

Then the key bindings could be listed as :

|Key binding|Description|
|-|-|
|[prefix] p g|Get project list|
...

And add a changelog.develop entry.

layers/+emacs/org/packages.el
klochowicz: fixed. thanks for pointing these out and linking the style guide -  I'm still very much a beginner with lisp...

klochowicz:
I've actioned all the suggestions from above, including specifying the prefix before the keybindings table.

duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/1227cbd75d95c3aa1a8ec25797cc9d5ffbb12ff3

PR #14009, opened 2020-10-03
remove package `mercury-mode` which no more exist

0kaguya:
`mercury-mode` had been replaced by `metal-mercury-mode`. Now `mercury-mode` can't be found from all sources so it needs to be removed.


smile13241324 (approved):
Looks ok, will be merged soon

PR #14010, opened 2020-10-03
Bind `SPC s l` to ivy-resume

mineo:
Helm already has spacemacs/resume-last-search-buffer bound to it and is in my
muscle memory. spacemacs/resume-last-search-buffer doesn't work with ivy, so
just use ivy-resume.

### DIFF
--- a/layers/+completion/ivy/packages.el
+++ b/layers/+completion/ivy/packages.el
@@ -208,6 +208,7 @@
         "Cw" 'counsel-colors-web
         "fr" 'counsel-recentf
         "rl" 'ivy-resume
+        "sl" 'ivy-resume
         "bb" 'ivy-switch-buffer)
       ;; Moved C-k to C-M-k
       (define-key ivy-switch-buffer-map (kbd "C-M-k") 'ivy-switch-buffer-kill))


duianto:
Commit(s) https://github.com/syl20bnr/spacemacs/commit/b606d82dfb63a07b9d300a6edb8325f6613067cc:

I thought that since Spacemacs has a custom function for opening the previous helm search buffer.

https://github.com/syl20bnr/spacemacs/blob/c6c949156ac7f4eddf8c96d099787c0f869ad158/layers/%2Bcompletion/helm/funcs.el#L497

Then maybe something similar could be created for the ivy layer.

I found this suggestion:
https://emacs.stackexchange.com/questions/55775/how-can-i-resume-swiper-isearch-with-the-next-line-selected
```elisp
(defun my-swiper-isearch-again ()
  "Start swiper-isearch with the last thing searched for."
  (interactive)
  (swiper-isearch (car swiper-history)))
(global-set-key (kbd "s-g") 'my-swiper-isearch-again)
```
But it just repeats the last swiper search term in the current buffer.

This might be a starting point for reopening a previous search prompt:
`(counsel-rg (car counsel-git-grep-history) "~/.emacs.d")`

But then the previous search directory might need to be remembered,
(maybe there's already a variable that stores it, I didn't check).

But then I saw the docstring for `ivy-resume`
>(ivy-resume &optional SESSION)
>
>Resume the last completion session, or SESSION if non-nil.
With a prefix arg, try to restore a recorded completion session,
if one exists.

`SPC u SPC r l` (with this PR: `SPC u SPC s l`),
lists the previous ivy sessions:

emacs_2020-10-10_05-15-39

It's probably best to use the upstream method for finding a previous ivy session.

PR #14014, opened 2020-10-04
line-number funcs under SPC t n

Bost:
Before:
before

After - new prefix:
after-prefix

After - content of the new prefix:
after-prefix-content


smile13241324 (approved):
Nice improvement of these bindings, I especially like that they are more mnemonic with the new subgroup. Keeping the old symbols around is also a good idea.

PR #14015, opened 2020-10-04
Handle buffer move/select/swap to last win nr +1

duianto:
problem:
The following error messages are shown.

When trying to move, select or swap the current buffer,
to a window nr greater than the total number of open windows.

For example:
With only one window open.
`SPC b 2`
>ad-Advice-select-window: Wrong type argument: window-live-p, nil

In the Buffer Transient State (`SPC b .`):
- `C-2` (`spacemacs/buffer-transient-state/winum-select-window-2`)
>Wrong type argument: window-valid-p, #<window 12>

- `2` (`spacemacs/buffer-transient-state/move-buffer-window-no-follow-2`)
>Wrong type argument: window-live-p, #<window 14>

- `M-2` (`spacemacs/buffer-transient-state/swap-buffer-window-no-follow-2`)
>Wrong type argument: stringp, nil

Solution:
Show descriptive message, if the target window nr is greater than the number of
open windows. And ignore the LV buffer (hydra).


smile13241324 (approved):
Looks good, conflict is just in the changelog, I can fix it during the merge.

PR #14031, opened 2020-10-11
Fix fsharp-mode autocomplete not working

niontrix:
Fsharp-mode has recently undergone some major changes. FSAutocomplete is
no longer distributed with Fsharp-mode. Instead Eglot is used as a language
server that pulls in FSAutocomplete. To get auto-completion working
again eglot-fsharp has to be required and an instance of Eglot has to be
started once an F# file is opened.

I'm not quite sure if it would be better if Eglot would start automatically when fsharp-mode is activated?!
But as it is it works for now.

### DIFF
--- a/layers/+lang/fsharp/README.org
+++ b/layers/+lang/fsharp/README.org
@@ -28,11 +28,12 @@ file.
 
 * Key bindings
 
-| Key binding | Description                |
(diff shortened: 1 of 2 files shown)


niontrix:
Will squash two commits into one before reopening pull request

PR #14032, opened 2020-10-11
Fix fsharp-mode autocomplete not working

niontrix:
Fsharp-mode has recently undergone some major changes. FSAutocomplete is
no longer distributed with Fsharp-mode. Instead Eglot is used a language
server that pulls in FSAutocomplete. To get auto-completion working
again eglot-fsharp has to be required and an instance of Eglot has to be
started once an F# file is opened.

I'm not quite sure if it would be better to start Eglot automatically when fsharp-mode is activated.
But as it is now it works.

### DIFF
--- a/layers/+lang/fsharp/README.org
+++ b/layers/+lang/fsharp/README.org
@@ -16,7 +16,7 @@ This layer adds support for F# language using [[https://github.com/fsharp/fsharp
 
 ** Features:
 - Syntax highlighting
-- Code completion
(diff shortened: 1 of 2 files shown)


smile13241324 (approved):
Looks ok, using eglot actually just means there is an lsp server in place, we should consider adding eglot as alternative for lsp in time. Anyway for now this layer only supports eglot.

smile13241324:

PR #14034, opened 2020-10-12
Port fix and enabled TLS encryption

kn07:
Just updated the port and set tls encryption.


smile13241324:
Sorry no PRs to master, I am closing this now.

PR #14035, opened 2020-10-12
Port fix and enabled TLS encryption

kn07:
Just updated the port and set tls encryption.


smile13241324 (approved):
Good idea, we should show how to make a tls encrypted connection in the samples where ever we can.

PR #14037, opened 2020-10-13
Set start directory

emacs18:
`spacemacs-start-directory` has to be set to only one value, so why not set it within `init.el`?

If spacemacs is "embedded" so that spacemacs is not the same as `user-emacs-directory`, then the user can simply load spacemacs' `init.el` file directly without having to set `spacemacs-start-directory` first, e.g.,
```elisp
(load "path/to/spacemacs/init.el")
```
This is exactly what is done within `dump-init.el`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,6 @@ the [full installation instructions](#install) for other options.
     - [Default installation](#default-installation)
     - [Alternative installations](#alternative-installations)
         - [Modify HOME environment variable](#modify-home-environment-variable)
-        - [Modify spacemacs-start-directory variable](#modify-spacemacs-start-directory-variable)
     - [Spacemacs logo](#spacemacs-logo)
 - [Update](#update)
     - [Automatic update (on master branch)](#automatic-update-on-master-branch)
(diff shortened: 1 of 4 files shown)


smile13241324:
I am not entirely sure about possible side effects of this change, care to have a look @syl20bnr?

emacs18:
I thought of this after realizing that the variable has to be set to one and only one value, i.e., where `init.el` is located.
If this is indeed the case, then the variable should not even be exposed to users, and it should be just an internal variable.

README.md
```diff
@@ -388,16 +387,6 @@ HOME=~/spacemacs emacs
...
-### Modify spacemacs-start-directory variable
-This solution is better suited to "embed" Spacemacs into your own configuration.
-If you've cloned Spacemacs into `~/.emacs.d/spacemacs/`, then drop the following
-lines in the `~/.emacs.d/init.el` file:
-
-```elisp
-(setq spacemacs-start-directory "~/.emacs.d/spacemacs/")
-(load-file (concat spacemacs-start-directory "init.el"))
-```
-
```
smile13241324, line 400: Setting `spacemacs-start-directory` in `init.el` should be ok, however I think that we still need to document how this would be setup. In this case we would need to mention that it is required to load the spacemacs init.el for this to work.

smile13241324 (changes requested):
Your change should be ok, except of removing the documentation, here I would prefer to still have an example how to do such a setup for users not being familiar with the emacs world yet.

core/core-load-paths.el
```diff
@@ -18,7 +18,7 @@
 ;; paths
 (defvar spacemacs-start-directory
   user-emacs-directory
-  "Spacemacs start directory.")
+  "Spacemacs start directory.  This is set within `init.el'.")
```
lebensterben, line 21: I'm a bit lost here.

`spacemacs-start-directory` is set in dotspacemacs. But in `init` file

emacs18:
I think I'm just going to close this out.  I don't remember the details.  In any case I have adopted using chemacs2 package to solve the problem I wanted to solve which was to allow me to decide whether to use spaceamcs or doom or any other configuration when I launch emacs.

smile13241324:
Ok, I'll close it.

PR #14038, opened 2020-10-13
Fix counsel find-file functions to use Spacemacs large file check

dalanicolai:
By default Ivy uses `find-file` which uses Emacs `abort-if-file-too-large`
function to check/warn before opening large files. That function however does
not take into account Spacemacs its `spacemacs-large-file-modes-list`. Instead
ivy should use `find-file-noselect` to skip Emacs it large file warning and let
it to Spacemacs its `spacemacs/check-large-file` function. Additionally [Emacs
tips on
comments](https://www.gnu.org/software/emacs/manual/html_node/elisp/Comment-Tips.html)
recommends to use 3 semicolons for comments that function as headings.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2349,6 +2349,8 @@ Other:
 - Fix visual selection expansion across the buffer on `SPC s S` and `SPC s B`
   (thanks to Andriy Kmit)
 - Fixed =cl= package deprecated =letf*= (thanks to duianto)
+- Fixed counsel find-file functions to use Spacemacs check for opening large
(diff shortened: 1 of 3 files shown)


dalanicolai:
For example, while `pdf-view-mode` is in the `spacemacs-large-file-modes-list`, Spacemacs will ask for confirmation for opening the large file (literally) when opening a pdf larger than `large-file-warning-threshold` using Ivy.

smile13241324 (approved):
Nice fix, will be merged soon.

smile13241324:

PR #14042, opened 2020-10-14
[ivy] Rebind C-k to C-M-k in ivy-reverse-i-search

duianto:
This enables navigating up in the `ivy-reverse-i-search` list with `C-k`.

The same rebinding from `C-k` to `C-M-k` to kill a list entry,
is also done in the `ivy-switch-buffer`.


smile13241324 (approved):
Looks good.

PR #14043, opened 2020-10-14
Only set undo system if fun. exists.

khjcph:
The PR fixes an issue where spacemacs is in a bad state after upgrading. I suppose what happens is that `evil-set-undo-system` is invoked before `evil` has been upgraded.


real-or-random:
I can confirm that this fixes the bad state after upgrading.

duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/dad4406f70e8d84f6e0eb8dcc1398ddcb05bbaee

PR #14046, opened 2020-10-14
[layers/+tools/languagetool] Allow HTTP server instead of binary

rpls:
The `langtool` package can also work against a running server instead of a LanguageTool binary/classpath. Accordingly, the `spacemacs//languagetool-detect` function should check whether the server variables are set. I added a check and a description in the `README.md`.


smile13241324 (approved):
Nice addition to the layer, will be merged soon.

smile13241324:

PR #14047, opened 2020-10-14
[editing] Integrate evil-vimish-fold via dotspacemacs-folding-method

Kazark:
Not confident that I have plugged this in everywhere that I should have. However
I tested locally and it seems to be working. Feedback coveted.

### DIFF
--- a/layers/+spacemacs/spacemacs-editing/packages.el
+++ b/layers/+spacemacs/spacemacs-editing/packages.el
@@ -31,6 +31,8 @@
         string-inflection
         undo-tree
         uuidgen
+        (vimish-fold :toggle (eq 'vimish dotspacemacs-folding-method))
+        (evil-vimish-fold :toggle (eq 'vimish dotspacemacs-folding-method))
         ws-butler))
 
 ;; Initialization of packages
@@ -275,6 +277,18 @@
 ;; Note: The key binding for the fold transient state is defined in
 ;; evil config
 
+(defun spacemacs-editing/init-vimish-fold ()
+  (use-package vimish-fold
+    :ensure
+    :after evil))
+
+(defun spacemacs-editing/init-evil-vimish-fold ()
+  (use-package evil-vimish-fold
+    :ensure
+    :after vimish-fold
+    :init
+    (setq evil-vimish-fold-target-modes '(prog-mode conf-mode text-mode))
+    :config (global-evil-vimish-fold-mode)))
 
 (defun spacemacs-editing/init-password-generator ()
   (use-package password-generator
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -1310,7 +1310,7 @@
   =dotspacemacs-command-key=. This variable represents the key to press after
   the leader key to execute =M-x=. Default value is ~SPC~.
 - New variable =dotspacemacs-folding-method= to allow choosing between
-  different code folding methods. Currently supported are =evil= and =origami=.
+  different code folding methods. Currently supported are =evil=, =origami= and =vimish=.
   Default value is =evil=. (thanks to ralesi)
 - New variable =dotspacemacs-ex-substitute-global=, if non-nil then
   the behavior of the =g= flag in =:substitute= ex-command is inverted.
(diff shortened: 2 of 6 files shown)


layers/+spacemacs/spacemacs-editing/packages.el
```diff
@@ -31,6 +31,8 @@
         string-inflection
         undo-tree
         uuidgen
+        (vimish-fold :toggle (eq 'vimish dotspacemacs-folding-method))
+        (evil-vimish-fold :toggle (eq 'vimish dotspacemacs-folding-method))
```
robbyoconnor, line 35: use `when`.

robbyoconnor (changes requested):

layers/+spacemacs/spacemacs-editing/packages.el
robbyoconnor: https://www.gnu.org/software/emacs/manual/html_node/elisp/Conditionals.html

Kazark: @robbyoconnor like this?

```elisp
(when (eq 'vimish dotspacemacs-folding-method) vimish-fold evil-vimish-fold)
```

Hm, that can't be quite right.

Kazark: Do I need to use a quasiquote as well?

```elisp
`(
  ...
  @,(when (eq 'vimish dotspacemacs-folding-method) '(vimish-fold evil-vimish-fold))
  ...
  )
```

robbyoconnor: Check other layers

smile13241324: I don't think we need a `when` statement here, as `when` and `if` take a condition and then either execute an action or choose between two alternative actions. In this case this is already covered by the `:toggle` label which will take the condition.

smile13241324 (approved):
Looks good so far, I think you have found all the places to change.

smile13241324:

ECon87:
New to emacs so probably an error on my part. I am on the develop branch but I can't seem to get this to work.

I am working on an R file; I enclose a snippet of code witin "# {{{" and "# }}}", however nothing gets folded after I press "z c".

In python only the line underneath gets folded (instead of the whole snippet of code).

Thanks.

lebensterben:
@ECon87 Can you post a minimum example R script so I can try to reproduce the bug?

ECon87:
> @ECon87 Can you post a minimum example R script so I can try to reproduce the bug?

```
library(dplyr)

# ==================
# Directory Paths
# ==================

# {{{

Root0Dir <- "some/path/"
TabFigDir <- paste0(Root0Dir, "Tables_Figures/")
RootDir <- paste0(Root0Dir,"Empirical/")
DataDir <- paste0(RootDir, "Data/")


# }}}

```

The above snippet should it. When I press "z c" while it folds the area above "# {{{", and not below. If I remove the comment, then "zc" does not fold anything.

Thanks for the response.

lebensterben:
I can confirm this.
But I'm not sure this is bug.
Neither can Vim fold it.

Do you know any particular editor that folds them?

ECon87:
I can fold it in Vim. But just to make sure we are talking about identical items.

I have the following code in my vimrc

```
autocmd BufNewFile,BufRead *.r,*.R :set foldmethod=marker
```

Then, when I press "z R" (or z o, or z r or z a and so). The fold works as expected.

lebensterben:
@ECon87 

https://github.com/alexmurray/evil-vimish-fold

This may work. In particular it folds {{{ }}}.

ECon87:
I might be wrong, but isn't this already working underneath the spacemacs fold functions?

Kazark:
@ECon87

1. What is `dotspacemacs-folding-method` set to in your `.spacemacs`?
2. Do you have this problem in other major modes?

ECon87:
Hey @Kazark,

1. It's set to 'evil

2. In Python too. I am only using R and Python to be honest, and i am new to emacs.

Kazark:
@ECon87 this will definitely not work with it set to `'evil`. This is a bit confusing, but there is no folding package for Emacs that provides the full Vim functionality out of the box, so what this PR did was it gave the `'vimish` option, which uses the package that @lebensterben linked to. Trying setting `dotspacemacs-folding-method 'vimish` and reloading your configuration (`SPC f e R`; you may also need to restart Spacemacs, not sure; `SPC q r`), then let us know if it's still not working.

ECon87:
@Kazark that worked like a charm. Thank you.

I checked the merged code above, and just noticed that the possible values for the `dotspacemacs-folding-method` are `evil, origami, and vimish`.  My `.spacemacs` only listed `evil and origami` as possible values.

For anyone else on spacemacs develop branch, you might need to update via `cd .emacs.d; pull --rebase`.

Kazark:
@ECon87 you may need to learn about `SPC f e D` (diff your `.spacemacs` with the newest version).

PR #14070, opened 2020-10-20
[git] Implement performance hack for MacOS from Magit Manual

Kazark:
I indeed observed the slowness and this helped. It may not be necessary for every old version of Mac OS but should be harmless on them all.

### DIFF
--- a/layers/+source-control/git/packages.el
+++ b/layers/+source-control/git/packages.el
@@ -234,6 +234,9 @@
       ;; whitespace
       (define-key magit-status-mode-map (kbd "C-S-w")
         'spacemacs/magit-toggle-whitespace)
+      ;; https://magit.vc/manual/magit/MacOS-Performance.html
+      (when (eq system-type 'darwin)
+        (setq magit-git-executable (s-trim (shell-command-to-string "which git"))))
       ;; full screen magit-status
       (when git-magit-status-fullscreen
         (setq magit-display-buffer-function
--- a/layers/+source-control/git/packages.el
+++ b/layers/+source-control/git/packages.el
@@ -238,9 +238,8 @@
       ;; But modified according Tommi Komulainen's advice: "...going through
       ;; shell raises an eyebrow, and in the odd edge case of not having git
       ;; setting the executable to empty string(?) feels slightly wrong."
-      (when (eq system-type 'darwin)
-        (when-let ((git (executable-find "git")))
-          (setq magit-git-executable git)))
+      (when-let ((git (executable-find "git")))
+        (setq magit-git-executable git))
       ;; full screen magit-status
       (when git-magit-status-fullscreen
         (setq magit-display-buffer-function
--- a/layers/+source-control/git/packages.el
+++ b/layers/+source-control/git/packages.el
@@ -235,8 +235,12 @@
       (define-key magit-status-mode-map (kbd "C-S-w")
         'spacemacs/magit-toggle-whitespace)
       ;; https://magit.vc/manual/magit/MacOS-Performance.html
+      ;; But modified according Tommi Komulainen's advice: "...going through
+      ;; shell raises an eyebrow, and in the odd edge case of not having git
+      ;; setting the executable to empty string(?) feels slightly wrong."
       (when (eq system-type 'darwin)
-        (setq magit-git-executable (s-trim (shell-command-to-string "which git"))))
+        (when-let ((git (executable-find "git")))
+          (setq magit-git-executable git)))
       ;; full screen magit-status
       (when git-magit-status-fullscreen
         (setq magit-display-buffer-function
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1990,6 +1990,8 @@ Other:
   - Moved =magit-find-file= to view a file at a specific branch or commit
     from ~SPC g f f~ to ~SPC g f F~ (thanks to duianto and Hong Xu)
   - Unbind ~C-f~ from =magit-gitflow-popup= (thanks to rayw000)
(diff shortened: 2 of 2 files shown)


layers/+source-control/git/packages.el
```diff
@@ -234,6 +234,9 @@
       ;; whitespace
       (define-key magit-status-mode-map (kbd "C-S-w")
         'spacemacs/magit-toggle-whitespace)
+      ;; https://magit.vc/manual/magit/MacOS-Performance.html
+      (when (eq system-type 'darwin)
+        (setq magit-git-executable (s-trim (shell-command-to-string "which git"))))
```
tko, line 239: I'm probably nitpicking but going through shell raises an eyebrow, and in the odd edge case of not having git setting the executable to empty string(?) feels slightly wrong. I would've thought something like this, but I'm not sure if deferring to shell would actually be preferrable
```elisp
(when-let ((git (executable-find "git")))
  (setq magit-git-executable git))
```
I suppose one could argue this should be safe on all OSs and there might be some benefit for being consistent, but :shrug:

Kazark:
Thanks @tko. I did leave the `(when (eq system-type 'darwin) ...)` as a marker for the reason why the code exists at all.

layers/+source-control/git/packages.el
smile13241324: I tend to agree to @tko, this code should be fairly save for all OSes so for the sake of consistency please remove the `only darwin when` please.

Kazark: @tko @smile13241324 alright, removed the condition in ad941de

Kazark:
The "Validate PR" build is broken. I wonder why.

smile13241324 (approved):
Looks good now.

smile13241324:
Looks like you have merged develop onto your branch which should have been a rebase, that's why the PR check fails. I will squash your commits anyway so it does not hurt.

smile13241324:

Kazark:
Ah, rebase, don't merge. Got it---thanks.

bcc32:
FYI, this PR breaks the use of magit over TRAMP when the path to git is different on the local and remote machines.  For example, on my mac, magit-git-executable gets set to `/usr/local/bin/git` (installed via homebrew), but this breaks when I try to access files remotely on my NixOS server using TRAMP, where git is at `/run/current-system/sw/bin/git`.

Also, as it happens, on macOS Big Sur (11.1), I see no speed difference in `magit-status` with and without this patch.  I think, at the very least, this code should be re-guarded behind an OS check.  IMO, it seems weird to lock magit into using whatever git was found in path at startup.

Kazark:
@bcc32 Hm, maybe I should just revert this PR entirely. I didn't understand that there were conditions under which it could be undesirable.

bcc32:
@Kazark Since you observed the slowness without this patch, I would be sad to lose it entirely.  Maybe it could be opt-in behind a configuration option, or simply a suggestion in the git layer docs?

Kazark:
@bcc32 is there a way to detect the presence of TRAMP, and disable it in that case?

bcc32:
The definition of `magit-git-executable` makes me think there's no current way to do this:

```
(defcustom magit-git-executable
  ;; Git might be installed in a different location on a remote, so
  ;; it is better not to use the full path to the executable, except
  ;; on Window where we would otherwise end up using one of the
  ;; wrappers "cmd/git.exe" or "cmd/git.cmd", which are much slower
  ;; than using "bin/git.exe" directly.
```

This suggests to me that, on Windows, magit across TRAMP just doesn't work by default due to lack of magit support for changing `magit-git-executable` dynamically. :(

Kazark:
@bcc32 I meant: detect if TRAMP is installed/used in that Emacs at the time that I set `magit-git-executable`, and just don't set it. I don't use TRAMP, and I don't know much about it.

bcc32:
Ah, my bad.  Since TRAMP is included with Emacs, it's always installed.

Kazark:
What if I reguarded the code, with a specific check for Mac OS version? Based on your results, perhaps whatever the root problem here is fixed in Big Sur? (and I wonder how I would determine that...)

bcc32:
That sounds fine to me.  I'm not really sure which versions it would be appropriate to enable the patch for.  The PR that originally introduced this suggestion in the magit docs is magit/magit#4069, which links to some reddit threads with discussions.

lebensterben:
@Kazark 
In my opinion this should be reported to upstream.
First, the optimisation trick comes from magit's manual, so its impact goes beyond spacemacs users.
Second, if macOS should be treated specially (as opposed to other OS), it should be done so in magit itself.

Kazark:
@lebensterben excellent point. Will raise it on the issue that @bcc32 linked to.

jjgriego:
Just gonna leave a breadcrumb for anyone else who gets here and needs a fix, since it was a little nonobvious, sorry for resurrecting a zombie PR. 

I fixed the TRAMP issue for myself locally by adding to `spacemacs/user-config`:

`(eval-after-load 'magit '(setq magit-git-executable "git"))`

I can understand why we might want to leave this in until magit can patch since an error ("where's your git?!") is at least a little bit more to work with when debugging than mysterious slowness.

bcc32:
> I can understand why we might want to leave this in until magit can patch since an error ("where's your git?!") is at least a little bit more to work with when debugging than mysterious slowness.

I agree in principle, although I didn't actually experience an error.  Instead, when I pressed `SPC g s` in a repository over TRAMP, I was prompted for the git repo root.  (But I'm already in a git repo!).  So it is confusing and nonobvious either way.

PR #14071, opened 2020-10-20
Add meghanada, eln-cache and vimish-fold to .gitignore

Kazark:
* meghanada is an LSP client for Java
* eln-cache (gccemacs) is for the native-comp branch of Emacs 28.0
* vimish-fold is there because said package has persistent folds

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,3 +91,6 @@ tags
 .tags
 !tags/
 gtags.files
+eln-cache/
+meghanada/
+vimish-fold/


.gitignore
```diff
@@ -91,3 +91,7 @@ tags
 .tags
 !tags/
 gtags.files
+
+eln-cache/
+meghanada/
+vimish-fold/
```
sdwolfz, line 97: Make sure you add a newline at the end of file, otherwise :+1:

Kazark: @sdwolfz fixed

PR #14072, opened 2020-10-20
[org] SPC m s N -> SPC m s w to match mnemonics of SPC n w

Kazark:
I chose to go this way rather than the other way because:

* Whenever you can avoid holding down a modifier key, great!
* It seems like the more specific should yield to the more general

This may be a controversial change, because it could break some folks' finger
memory. However I consider this a very unpleasant inconsistency that inhibits
the development of finger memory, so I thought I would at least suggest it as
the default for Spacemacs.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2809,6 +2809,8 @@ files (thanks to Daniel Nicolai)
   - Key binding changes for archiving (thanks to Ag Ibragimov):
     - Added ~SPC m s a~ to toggle archive tag for subtree
     - Added ~SPC m s A~ to archive subtree (previously ~SPC m s a~)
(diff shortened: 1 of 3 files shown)


smile13241324 (approved):
I like this change, out of the following reasons:
  - `, s w` is more mnemonic than `, s N`which is subtree widen
  - `SPC n w` is already bound following the mnemonic above, so it would be fit to adjust all bindings to match it.
  - the upper case modifiers should actually be used to generate a more specific version of the lower case version of the binding like it is shown in `SPC s d` and `SPC s D`. The current binding violates that principle.

Therefore I think it is right to adjust the bindings in that case.

smile13241324:

Issue #2157, opened 2015-07-01
Helm-swoop keyboard arrow keys broken when helm-full-frame is set to true


dtreiter:
_Steps to reproduce_:
- Open spacemacs
- Set helm-full-frame to t
- Do helm-swoop
- Try to use the up and down arrow keys to move between candidates

_Expected behavior_: Should move the selection up and down one line at a time.
_Current behavior_: Behaves unpredictably.

I'm not sure if this is an issue with spacemacs or with helm/helm-swoop. Please correct me if this should be opened on helm or helm-swoop.

StreakyCobra:
Can anyone reproduce this with a raw emacs? To know if it should be reported upstream or solved here?

dlebrecht:
differently than reported above, but probably the same bug. 

If i have in my .spacemacs, 
`(setq same-window-regexps '("."))`

helm-swoop get stuck in the first and last entries, using all available shortcuts. Removing that line solves the issue.

tgroshon:
I'm having this same issue since upgrading to `0.200.10`.  It doesn't occur in my `0.200.9` install on my other computer.

davhanks:
Same issue since upgrading to `0.200.10`. Cannot go up or down with arrows or other move commands.

ngoue:
Also having this issue in `0.200.10@25.2.1`.  Maybe related, swoop now defaults to full-screen mode too...

tgroshon:
Also, my `helm-full-frame` is set to `nil`. So it doesn't appear to be related to this problem.

duianto:
After setting `helm-full-frame` to `t`, and opening `helm-swoop`, then the first time one moves the selection, with either `C-j`, `C-n` or `<down>` (or the move up equivalent bindings), then the selection moves one line, as expected. 

But after the first time, then moving down doesn't move the selection, and moving up, moves the selection by two lines at a time.

It's as if the selection actually is on the line above the current visibly selected line.

Checking the key binding `S-C-h <down>` shows that it's bound to the expected command: `helm-next-line`, the same applies for the `<up>` binding which calls `helm-previous-line`.

I tried to replicate it without Spacemacs, by starting Emacs with the `--quick` argument, and pointing the `package-user-dir` variable, to the develop branchs' packages directory:
`(setq package-user-dir "~/.emacs.d/elpa/26.0/develop/")`
`M-x package-initialize`
`S-M-; (setq helm-full-frame t)`
`M-x helm-swoop`

[...]

PR #14075, opened 2020-10-21
[helm-swoop] ignore `helm-full-frame' and `pop-up-windows' settings

surgeboris:
because current implementation does not support full-frame usage:
see https://github.com/syl20bnr/spacemacs/issues/2157


duianto:
Commit(s): https://github.com/syl20bnr/spacemacs/commit/3e033e39f656868f8f83bcb903690dec1757fc08

PR #14077, opened 2020-10-21
Modify documentation to reflect that Emacs server is off by default

dmusican:
Pull request #10297 made the Emacs server optional, so that it does
not start up unless the init file is modified to enable it. The
documentation, however, indicates that the opposite is true. This
fix updates the documentation to be consistent with the actual
behavior.

Thank you <3


smile13241324 (approved):
Good catch, will be merged soon.

smile13241324:

PR #14082, opened 2020-10-22
Specify configuration layers info pages and add cross links

dalanicolai:
Searching for information about configuration layers is confusing because both
the overview and development info pages have the same title (try it with
google). Specifying those info pages title names and adding cross links makes
the docs friendlier.

NOTE!!! I have inserted relative links. But I'm afraid they will not work in the website, however I as I can not build the doc pages, I can not test the links. Maybe you could direct me to instructions about how to make 'internal' links (between different pages) in the Spacemacs docs.

### DIFF
--- a/.ci/spacedoc-cfg.edn
+++ b/.ci/spacedoc-cfg.edn
@@ -1,4 +1,14 @@
-{:spacetools.spacedoc.config/layers-org-query
+{:spacetools.spacedoc.config/layers-org-title-text "Spacemacs layers list"
+
+ :spacetools.spacedoc.config/layers-org-description
+ "\nTHIS FILE IS AUTO-GENERATED!
+Don't edit it directly. See [[https://github.com/syl20bnr/spacemacs/blob/develop/CONTRIBUTING.org#readmeorg-tags][\"README.org  tags\" section of CONTRIBUTING.org for the instructions]].
+
+This is an overview of Spacemacs configuration layers. For information about
+configuration layer development see [[https://develop.spacemacs.org/doc/LAYERS.html][Configuration layers development]].
+"
+
+ :spacetools.spacedoc.config/layers-org-query
  {"layer" ["chat"
            "checker"
            "completion"
--- a/doc/LAYERS.org
+++ b/doc/LAYERS.org
@@ -1,4 +1,4 @@
-#+TITLE: Configuration layers
+#+TITLE: Configuration layers development
 
 * Table of Contents                     :TOC_5_gh:noexport:
 - [[#introduction][Introduction]]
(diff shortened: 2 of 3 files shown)


smile13241324:
I am not sure that this change will help, when you check the remarks in layers/layers.org you will notice that this file is auto generated so the text you have added will most likely be overwritten with the next bot run.

About the relative links maybe @JAremko can help you with this, I am not entirely sure how this file is build and what to do to change its headline. However I would expect that the change would rather need to be made in the template than in the generated file.

JAremko:
Changes to `doc/LAYERS.org` should work. And @smile13241324  right  `layers/LAYERS.org` will get overridden. We can set title for  `layers/LAYERS.org` to something like "layers list" to make it even less confusing.

You can change the title by adding `:spacetools.spacedoc.config/layers-org-title-text "new title"` and the text by `:spacetools.spacedoc.config/layers-org-description "new text"` into
 https://github.com/syl20bnr/spacemacs/blob/develop/.ci/spacedoc-cfg.edn#L83

current value for description text is
```
"This file is auto-generated!\nDon't edit it directly.\nSee [[https://github.com/syl20bnr/spacemacs/blob/develop/CONTRIBUTING.org#readmeorg-tags][\"README.org tags\" section of CONTRIBUTING.org for the instructions]].
```

JAremko:
also you can test it locally with docker

``` sh
docker run --rm  -v <spacemacs_root_dir>:/tmp/docs/ 
                 -v <config_file>:/opt/spacetools/spacedoc-cfg.edn 
                 jare/spacetools docfmt /tmp/docs/`
```

it will reformat all docs and generate new layers.org.

smile13241324 (changes requested):
@dalanicolai please check JAremko's comment and do the changes for the generated layers.org file at the specified location, also please check the generation locally.

dalanicolai:
I was trying to reformat the docs but while many README's are exported successfully still, despite that I synced with latest develop branch, the process stops with the following error:

!!!!!!!!!!!!!!!! ERROR !!!!!!!!!!!!!!!!
current-buffer: INSTALL.org<2>
current-file: /tmp/docs/layers/+emacs/org/local/gnuplot/INSTALL.org
error: Link "Installing without configure or make" has unknown type "fuzzy"
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
Unfortunately I do not know what that means. Any suggestion?

JAremko:
@dalanicolai  you're probably using old Spacemacs revision.  I don't see file `gnuplot` dir here https://github.com/syl20bnr/spacemacs/tree/develop/layers/%2Bemacs/org/local

try running  bot over up to date develop fork.

JAremko:
also you can temporary delete INSTALL.org

dalanicolai:
Thanks for the nice instructions. So I totally forgot that I had manually installed gnuplot and did not notice it was in the local directory and I did not understand the docker command completely.

Anyway, I fixed the PR now and tested the result. I thought "Spacemacs layers list" to be a nice title, as this corresponds more or less were I would google for intuitively. But of course, otherwise better suggestions are welcome also.

JAremko:
Good job :+1:

smile13241324 (approved):
Looks good

smile13241324:

PR #14083, opened 2020-10-22
[core] provide keybindings for using narrow with an indirect buffer

### LINKED ISSUE #12550: Change default behavior of narrow-to-defun
I think it would be better to change the `narrow-to-defun` behavior to first use `clone-indirect-buffer` then use `narrow-to-defun`. That way you can divide your screen to narrow on different areas of the same file. I think this would improve workflow, so that instead of having to type
1. `SPC w /` `(split-window-right)`
2. `SPC b N i` `(clone-indirect-buffer)` 
3. `SPC n f` `(narrow-to-defun)`

You could instead simplify this to:

1. `SPC w /` 
2. `SPC n f`  

This adds functionality and retains the functionality found in the default behavior (for those who have no need of narrowing multiple regions).

Kazark:
This is useful if you want multiple narrowing in the same base buffer.

Fixes #12550 with proposed bindings by @duianto.


duianto:
#### Narrowing to a page `SPC n P`
Seems to work as expected.

A page line can be inserted by pressing `C-r` which calls `quoted-insert`,
followed by `C-l` (l = lower case L).

The docstring for `C-r` says:
>Read next input character and insert it.
This is useful for inserting control characters.
With argument, insert ARG copies of the character.

http://www.gnu.org/software/emacs/manual/html_node/emacs/Pages.html

Adding two or more page lines, and narrowing to a page `SPC n P`,
narrows the indirect buffer to the page with the cursor (between two page lines).

### Issues

#### Narrowing to a region `SPC n R`

Does create a new buffer with `<2>` appended to the buffer name.

But it shows this error message:
>Error in post-command-hook (evil-visual-post-command): (error "Marker points into wrong buffer" #<marker at 27903 in .spacemacs<2>>)

And most likely as a result of the error, the selected lines face (color) is stuck on the text in the new buffer.

#### Narrowing to a function `SPC n F`

With the cursor in this example function:
```elisp
  (defun my/function ()
    "add 1 and 2"
    (+ 1 2))
```
Creates a new buffer with the `<2>` post-fix name.

But the buffer isn't narrowed, and the line with the cursor scrolls to the bottom of the buffer.

There doesn't seem to be any error messages.

---

This was tested in `.spacemacs`.

<details><summary>
Windows 10, System Info (Click to expand)
</summary>
<pre>
#### System Info :computer:
- OS: windows-nt
- Emacs: 27.1
- Spacemacs: 0.300.0
- Spacemacs branch: nil (rev. nil)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
(autohotkey emacs-lisp
            (erlang :variables erlang-enable-lsp-mode t)
            git ivy javascript lsp multiple-cursors
                 '("c:/Users/username/org/")
                 org-enable-jira-support t org-enable-roam-support t)
            python racket rust shell spell-checking syntax-checking version-control treemacs
            (vue :variables vue-backend 'lsp)
            yaml)
```

</pre>
</details>

Kazark:
@duianto 🤔 Hm, that's troubling; thanks for testing. I tested too and did not run into errors. Do you get the same results if you just run the the component commands in sequence yourself?

duianto:
>Do you get the same results if you just run the the component commands in sequence yourself?

#### `(call-interactively 'clone-indirect-buffer)`

Creates the indirect buffer.
But it also scrolls the line with the cursor to the bottom of the buffer.

This message was shown: `#<buffer .spacemacs<2>>`
But that probably just means that it returned the indirect buffer.

#### `(call-interactively 'narrow-to-defun)`

Doesn't narrow as expected, to just show the closest function to the cursor:
```elisp
  (defun my/function ()
    "add 1 and 2"
    (+ 1 2))
```
in `.spacemacs` it narrows to the outer function: `dotspacemacs/user-config`
where the above function was defined.

Maybe `narrow-to-defun` looks for the closest function that doesn't have any indentation.

#### `(call-interactively 'narrow-to-region)`

This seems to narrow the buffer to the correct region.

Evil visual blocks `C-v`, are just treated as a normal selection. The buffer text between the characters in the top left to the bottom right selection block corners, are narrowed, even the text outside the block.

But it's expected, that feature probably hasn't been implemented yet in evil, since a visual block is evil specific, and narrow-to-region is a built in function in Emacs.

#### (spacemacs/narrow-to-region-indirect-buffer)

I also tried defining:
```elisp
  (defun spacemacs/narrow-to-region-indirect-buffer ()
    (interactive)
    (call-interactively 'clone-indirect-buffer)
    (call-interactively 'narrow-to-region)
    (message "Region narrowed to an indirect buffer"))
```
and evaluating: `S-M-;` `(spacemacs/narrow-to-region-indirect-buffer)`
in Manjaro, and it shows the same error:
>Error in post-command-hook (evil-visual-post-command): (error "Marker points into wrong buffer" #<marker at 27875 in .spacemacs<2>>)

Since you didn't see any errors, are you using the `emacs` editing style?

Kazark:
> it also scrolls the line with the cursor to the bottom of the buffer.

Hm, that seems odd to me.

I do run Manjaro on my home machine but I was doing this on my work (Mac OS) machine because that is where the motivating need came up. Perhaps it's not the OS but the major mode that makes the difference. I will retest in a Lisp buffer (was using Scala).

Thanks for the further test. I will try to track this down.

Kazark:
I tested this in an ELisp buffer and it worked just fine. Will test on Manjaro.

Kazark:
> Since you didn't see any errors, are you using the emacs editing style?

No, I am quite ignorant of Emacs bindings; 10 years of Vim in my background before coming to Emacs. Utterly helpless with `-q`. 😁

duianto:
The error message mentions `evil-visual-post-command` and that `Marker points into wrong buffer`.
>Error in post-command-hook (evil-visual-post-command): (error "Marker points into wrong buffer" #<marker at 27875 in .spacemacs<2>>)

So it might be related to this expression, in the `evil-visual-post-command` function:
https://github.com/emacs-evil/evil/blob/4dc63903d9688e2ce838a220b0e24d8f14a64c12/evil-states.el#L358
```elisp
      (setq evil-visual-x-select-timer
            (run-with-idle-timer evil-visual-x-select-timeout nil
                                 #'evil-visual-update-x-selection
                                 (current-buffer)))
```
It's the only place where `buffer` is mentioned in that function.

But I'm not quite sure how it works.

If we can't figure out of to fix the evil visual selection

Then another solution might be to deselect before cloning and narrowing.

This seems to work:
```elisp
  (defun spacemacs/narrow-to-region-indirect-buffer ()
    (interactive)
    (deactivate-mark)
    (call-interactively 'clone-indirect-buffer)
    (call-interactively 'narrow-to-region)
    (message "Region narrowed to an indirect buffer"))
```

The first thing that one does in the narrowed buffer, probably is to deselect anyway.

duianto:
Add the new key bindings to both the:
- Region narrowing section of the documentation:
https://github.com/syl20bnr/spacemacs/blob/develop/doc/DOCUMENTATION.org#region-narrowing
- And in changelog.develop (someone reading the changelog might be interested in what the new key bindings are)

Kazark:
Oh!! You were in visual select mode when you issued these commands! I can now reproduce the problem. I was issuing them from normal mode.

Kazark:
@duianto thanks for all the help with this. The problem should be fixed; I wrapped the creation of the indirect buffer in `(deactivate-mark) ... (activate-mark)`. This made things work for `SPC n R` as well and should follow the principle of least surprise by keeping the selection while avoiding the error you encountered.

I've also updated the documentation. I need to get better about remembering and locating the docs.

Kazark:
I'm learning (slowly)---remembered to rebase when I saw _Validate PR_ was red.

duianto:
>should follow the principle of least surprise by keeping the selection

Good point.

But our current solution:
- still has a surprise, the selection is deactivated in the original buffer.
- activates the visual selection, even if it wasn't active before narrowing.

This seems to work, to only activate it if it was active:
```elisp
(defun spacemacs/clone-indirect-buffer-de-activate-mark ()
  "This is a workaround for the evil visual state error message:
Error in post-command-hook (evil-visual-post-command):
(error \"Marker points into wrong buffer\" #<marker at 27875 in .spacemacs<2>>)"
  (let ((region-was-active (region-active-p)))
    (when region-was-active (deactivate-mark))
    (call-interactively 'clone-indirect-buffer)
    (when region-was-active (activate-mark))))

(defun spacemacs/narrow-to-defun-indirect-buffer ()
  (interactive)
  (spacemacs/clone-indirect-buffer-de-activate-mark)
  (call-interactively 'narrow-to-defun)
  (message "Function narrowed to an indirect buffer"))

(defun spacemacs/narrow-to-page-indirect-buffer ()
  (interactive)
  (spacemacs/clone-indirect-buffer-de-activate-mark)
  (call-interactively 'narrow-to-page)
  (message "Page narrowed to an indirect buffer"))

(defun spacemacs/narrow-to-region-indirect-buffer ()
  (interactive)
  (spacemacs/clone-indirect-buffer-de-activate-mark)
  (call-interactively 'narrow-to-region)
  (message "Region narrowed to an indirect buffer"))
```

---

There's an issue with the evil visual block `C-v`.
The characters in the visual block in the indirect buffer are stuck with the selected face.

And if the selection is canceled (`ESC`) in the indirect buffer, if one returns to the original buffer,
either by closing (`SPC b d`) the indirect buffer or by just switching to the original buffer (`SPC TAB`),
then the following error message is shown:
>Error in post-command-hook (evil-visual-post-command): (wrong-type-argument number-or-marker-p nil)

and the cursor is in `evil-state`: `visual` but there is no visible selection.

---

It might be easier to read the changelog key bindings if they are shown as a list:
```
  - Provide keybindings for using narrow with an indirect buffer, for when you
    want narrowing multiple times in the same base buffer:
    - ~SPC n F~ narrow to function
    - ~SPC n P~ narrow to page
    - ~SPC n R~ narrow to region
    (thanks to Keith Pinson and duianto)
```

---

The documentation key binding table can be resized to fit the longest line,
by pressing `TAB` with the cursor in the table.
```
| Key binding | Description                                          |
|-------------+------------------------------------------------------|
| ~SPC n f~   | narrow the buffer to the current function            |
| ~SPC n p~   | narrow the buffer to the visible page                |
| ~SPC n r~   | narrow the buffer to the selected text               |
| ~SPC n w~   | widen, i.e. show the whole buffer again              |
| ~SPC n F~   | narrow to the current function in an indirect buffer |
| ~SPC n P~   | narrow to the visible page in an indirect buffer     |
| ~SPC n R~   | narrow to the selected text in an indirect buffer    |
```

Kazark:
@duianto I think I have the table already aligned; I was using `, t a` and for me it looks like this: 
image

Kazark:
@duianto okay, I obviously need to move slower and test my stuff more thoroughly! I was really only testing my happy path, so I should have been trying harder to break it. I've incorporated your fix and simply blocked the use of visual block mode, as explained in the commit message of 4ef55b1. Hopefully this time I haven't missed anything/messed anything up 😬

duianto:
> I obviously need to move slower and test my stuff more thoroughly!

We can only test what we can think of at the time, it's fine to submit a PR that just solves the authors specific problem.

It might be easier for PR reviewers that haven't been working on solving the problem, to notice other use cases.

https://en.wikipedia.org/wiki/Linus%27s_law
>In software development, Linus's law is the assertion that "given enough eyeballs, all bugs are shallow".

---

I'm not sure why the key binding table is aligned for you.

There are two extra spaces after each key binding:
- in the GitHub diff:
https://github.com/syl20bnr/spacemacs/pull/14083/files#diff-fdda7ac8fccc885924f7f9874eb2184524099bb7e6a7f29a4876abc9fa1b897d

- and for me with this PR applied:
```
| Key binding | Description                                          |
|-------------+------------------------------------------------------|
| ~SPC n f~     | narrow the buffer to the current function            |
| ~SPC n p~     | narrow the buffer to the visible page                |
| ~SPC n r~     | narrow the buffer to the selected text               |
| ~SPC n w~     | widen, i.e. show the whole buffer again              |
| ~SPC n F~     | narrow to the current function in an indirect buffer |
| ~SPC n P~     | narrow to the visible page in an indirect buffer     |
| ~SPC n R~     | narrow to the selected text in an indirect buffer    |
```

With the cursor in the table, it becomes aligned for me when I press either:
-  `, t a                  ;; org-table-align`
- `<tab>                  ;; org-cycle`

---

Currently nothing happens when a evil visual block selection is active and one tries to narrow to an indirect buffer.

Let's add a message about it not being supported, fell free to rewrite the message if you come up with something better/shorter (it applies in general, if you would like to improve the de/activating mark function):
```diff
1 file changed, 6 insertions(+), 4 deletions(-)
layers/+spacemacs/spacemacs-defaults/funcs.el | 10 ++++++----

modified   layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -1745,10 +1745,12 @@ to work incorrectly. `x' is the string name of the entity being narrowed to."
   ;; workaround we did for visual line mode; this usecase however seems like an
   ;; edgecase at best, so let's patch it if we find out it's needed; otherwise
   ;; let's not hold up the base functionality anymore.
-  (when (not (and (eq evil-state 'visual) (eq evil-visual-selection 'block)))
-    (spacemacs/clone-indirect-buffer-de-activate-mark)
-    (call-interactively f)
-    (message (format "%s narrowed to an indirect buffer" x))))
+    (if (and (eq evil-state 'visual) (eq evil-visual-selection 'block))
+       (message (concat "An evil visual block selection is not supported,"
+                        " when narrowing to an indirect buffer"))
+     (spacemacs/clone-indirect-buffer-de-activate-mark)
+     (call-interactively f)
+     (message (format "%s narrowed to an indirect buffer" x))))
 
 (defun spacemacs/narrow-to-defun-indirect-buffer ()
   (interactive)
```

---

I noticed another issue:
The evil search `/` or `?` highlights get stuck, if there are active highlights when a buffer is narrowed.

Because the variable: `evil-ex-active-highlights-alist`
is `nil` in the indirect buffer.

In the original buffer, the variable contains:
```
evil-ex-active-highlights-alist is a variable defined in ‘evil-vars.el’.
Its value is
((evil-ex-search .
                 [evil-ex-search
                  ("narrow" t t)
                  evil-ex-lazy-highlight #<window 46 on funcs.el> nil nil nil nil
                  (#<overlay from 69934 to 69940 in funcs.el> #<overlay from 69789 to 69795 in funcs.el> #<overlay from 69321 to 69327 in funcs.el> #<overlay from 69126 to 69132 in funcs.el> #<overlay from 69067 to 69073 in funcs.el>)]))
Local in buffer funcs.el; global value is nil
```

Should we clear the highlights before narrowing, if the highlight variable is non-nil?

This seems to work:
```elisp
  (when evil-ex-active-highlights-alist
    (spacemacs/evil-search-clear-highlight))
```

It might be easy enough to press: `n` or `N` to re-highlight.

---

I also tested the highlight clearing and the evil-visual-selection 'block checking:
- on both the `spacemacs` and `spacemacs-base` distributions
- with both the `emacs` and `vim` editing styles

I didn't get any errors, so it seems to be working as expected.

---

Maybe it would be more descriptive to change the parameters: `f` and `x`
Possibly to something like:
```
narrow-function
narrow-target-string
```
or
```
narrowing-function
narrowing-target-string
```

Kazark:
> I'm not sure why the key binding table is aligned for you.

Perhaps this is the culprit (from my `.spacemacs`):

```elisp
(setq org-hide-emphasis-markers t)
```

Kazark:
> Maybe it would be more descriptive to change the parameters: f and x

You can see I am a Haskeller, used to getting my information from the types rather than the variable names. 😁 Since this is ELisp though I can totally see that renaming them could be helpful.

Kazark:
Alright, I think I've dealt with your latest string of comments. Let me know if there's anything else. 😁

duianto:
Commit(s):
603d7fbe0bde223700f762204d037493f4a178bd
5b4b6af8d352fc38050f7c7be283b10928e46bd9
155490afaf2f13e98afab5190554ae39af62fe2a
82974b454e204af1c18d9b222ebc6c3580c1040d
61fc17c6eefc4a2d1e0f1dfd5c60013b1dc4873b

**Modifications:**
Fixed the formatting of the if clause, according to:
https://github.com/bbatsov/emacs-lisp-style-guide#source-code-layout--organization
>Note that the “then” clause of an if form is a special argument, indent it by 4 spaces.
```elisp
;; good
(if something
    then-clause
  else-clause)

;; bad
(if something
  then-clause
  else-clause)
```

---

>Perhaps this is the culprit (from my .spacemacs):
>
>(setq org-hide-emphasis-markers t)

Yes, that seems to be the cause, it hides the `~` tilde characters around the key bindings, which makes it look like it's aligned.

Kazark:
Thanks for the explanation of the if formatting. I'd seen that style before and assumed it was wrong, hitherto only aware of naive "format by S-expression".

> it hides the ~ tilde characters around the key bindings, which makes it look like it's aligned.

Not only that, but Org actually considers it aligned; i.e. when I tell Org to align it, it does nothing.

Issue #13943, opened 2020-09-14
How should we respond to the archiving of Helm?


Kazark:
Background: https://github.com/emacs-helm/helm/issues/2386

Helm is the default in Spacemacs, so this is something we will need to respond to. Should we:

* use a fork? I haven't heard of it being forked yet, but surely it will be.
* Make Ivy the default? I've never tried Ivy and don't know anything about this option.
* Add support for [Selectrum](https://github.com/raxod502/selectrum)?

Thought I should start a discussion.

Kazark:
https://github.com/raxod502/ctrlf is a replacement for `helm-swoop`.

jb55:
> Make Ivy the default? I've never tried Ivy and don't know anything about this option

I switched to ivy in spacemacs awhile ago and it works fine. Seems like the most reasonable option.

Kazark:
@jb55 are there any features you miss?

smile13241324:
Don't know if this is the right moment to do so, I still expect that someone will adopt this package. 

Personally I don't think that `Selectrum` is an alternative, it follows the same design principle `ivy` does this is less features, be fast and simple. That's normally not the design expected by people using helm, rather something among the line of provide a lot of choices to do exactly what I want, be allowed to be slow but comprehensive.

Lets try to gather alternative comprehensive search and narrowing frameworks for emacs.
Maybe we find a replacement for helm after all.

Btw that doesn't mean we could not support `Selectrum` but it sounds pretty undistinguishable from standard `ivy` from my perspective.

Kazark:
@smile13241324 Selectrum's README links to another of comparable projects.

syl20bnr:
This is big and sad news. I'm a Helm user myself.
For sure Helm is a great package but it is also huge, I really hope someone can take over 🤞 

We won't phase it out any time soon though but it is always good to see the alternatives.
Fortunately we have very good support for Ivy too. 

Let's see what happens following this news.

Kazark:
Given that Helm is back, in some capacity, I'm thinking there's nothing for us to do at this time.

PR #14088, opened 2020-10-25
Minor changes on `helm` layer

lebensterben:
It's been advised by the upstream of `helm-ag` to set
`helm-ag-success-exit-status` to `'(0 2)`.

This PR adopted this suggestion.

ref: https://github.com/emacsorphanage/helm-ag#helm-agel-with-other-searching-tools

---

For `helm-ls-git` package, it's advised to set use `magit`as a bette alternative to
`vcr-dir` as the underlying command for getting `git status`.

This PR adopted this suggestion when `magit` package is in use.

Ref: https://github.com/emacs-helm/helm-ls-git/blob/4da1a53f2f0a078ee2e896a914a1b19c0bf1d5ed/helm-ls-git.el#L184-L189

Signed-off-by: Lucius Hu <lebensterben@users.noreply.github.com>

### DIFF
--- a/layers/+completion/helm/funcs.el
+++ b/layers/+completion/helm/funcs.el
@@ -297,7 +297,8 @@ If DEFAULT-INPUTP is non nil then the current region or symbol at point
   ;; --line-number forces line numbers (disabled by default on windows)
   ;; no --vimgrep because it adds column numbers that wgrep can't handle
   ;; see https://github.com/syl20bnr/spacemacs/pull/8065
-  (let ((helm-ag-base-command "rg --smart-case --no-heading --color=never --line-number --max-columns=150"))
+  (let ((helm-ag-base-command "rg --smart-case --no-heading --color=never --line-number --max-columns=150")
+        (helm-ag-success-exit-status '(0 2)))
     (helm-do-ag-buffers)))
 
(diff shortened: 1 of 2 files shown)


lebensterben:
I also suggest to move the functionality of `helm-git-ls` package to `git` layer, in a future PR.

This is because it registers a key binding with prefix `SPC g`, which fits well into `git` layer's configuration and documentation.

---

Also, this package is not actively maintained anymore. 

It mainly serves two functionalities. First it provides a view of repository status. Second it opens a helm buffer for filtering files in current repository. These two functionalities are well-served by `magit-status` and `projectile`.

We should also think about "marking" it as obsolete and remove this at 0.4x release.

AFAIK, there's no guide/convention on how to deprecating a package/feature. Is this correct?

smile13241324 (approved):
These changes look ok.

smile13241324:
About moving this package to the git layer, I don't think its such a good idea, basically the helm and ivy layers are excluding each other providing bindings for various other layers. If we move the respective packages in the single layers we would have to keep two different implementations in the same layers, i.e. helm and ivy git integrations.

This follows the same idea as the auto-complete layer and the language layers. If we would merge them we end up with two very complex layers which we should avoid.

About the helm depreciation discussion please refer to #13943.

Issue #14086, opened 2020-10-23
Spacemacs ranger layer: use ranger with ranger-enter-with-minus option


jacksondm33:
Currently, if set, the ranger-enter-with-minus option binds "-" to open deer in evil normal mode. It would be nice to have the option to choose between ranger or deer for this. If there is a specific reason for this being set to deer only, then is there any other simple way to access ranger without manually configuring a keybinding ('SPC a t r r' is annoyingly long)?

duianto:
Here's the commit where the `ranger-enter-with-minus` variable was added:
ranger: add jump other window bindings, minus to enter
https://github.com/syl20bnr/spacemacs/commit/47f90684c99d5b3f7e33fab7c60470b27e95710f

The variable docstring says:
>"Option to enter `deer' when `-' is pressed.  Idea from `vim-vinegar'."

When the variable is set to `t` (which it is by default), then the `-` key is bound like this:
```elisp
  (define-key evil-normal-state-map (kbd "-") 'deer)
```
If you want it to open ranger now, then you can add the following to the `dotspacemacs/user-config` section in your `.spacemacs` file:
```elisp
  (define-key evil-normal-state-map (kbd "-") 'ranger)
```

A PR has been opened which adds `'ranger` as an option to the `ranger-enter-with-minus` variable.
[ranger] Add 'ranger as an option to ranger-enter-with-minus #14096

duianto:
The following PRs changes have been applied to the `develop` branch:
[ranger] Add 'ranger as an option to ranger-enter-with-minus #14096

PR #14096, opened 2020-10-29
[ranger] Add 'ranger as an option to ranger-enter-with-minus

duianto:
Added `'ranger` as an option to the variable: `ranger-enter-with-minus`
Changed the default value from: `t` to: `'deer`.
Bound the key to `evil-motion-state-map` instead of `evil-normal-state-map`
so that `-` also works from the Spacemacs home buffer.

Fixes:
Spacemacs ranger layer: use ranger with ranger-enter-with-minus option
#14086


smile13241324 (approved):
:+1:

PR #14097, opened 2020-10-30
Add layer scratch into layers/+emacs/scratch 

rayw000:
### What Changed

Add layer scratch which makes your `*scratch*` buffer persistent and unkillable.

### Why
It makes sense if you want to save your drafts in `*scratch*` buffer temporarily.
Things you write down in `*scratch*` buffer will be automatically saved and restored.

### Packages Used
https://github.com/Fanael/persistent-scratch
https://github.com/EricCrosson/unkillable-scratch

### Note
This layer requests `spacemacs-scratch-mode-hook` introduced in PR https://github.com/syl20bnr/spacemacs/pull/13968

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -345,6 +345,7 @@ sane way, here is the complete list of changed key bindings
 - helpful (thanks to Johnson Denen, Andriy Kmit)
 - tabs (thanks to Deepu Mohan Puthrote)
 - verb (thanks to Federico Tedin)
+- scratch (thanks to rayw000)
 **** Email
 - notmuch (thanks to Francesc Elies Henar, Leonard Lausen, Willian Casarin,
   Kalle Lindqvist)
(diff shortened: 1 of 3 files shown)


smile13241324:
I like the idea of a persistent scratch buffer though I would not put it into a separate layer, instead I will merge it in the core letting users activate it via their dotfile.

I think most of your code should be ok as it is. The only thing I have seen is that your default bindings are in emacs style, for this we definitely require evil bindings too. But thats something I can fix while I am integrating it.

rayw000:
Thanks! And I'd like to do this job if you don't mind.😀

smile13241324:
Feel free to give it a try :+1:, lets add it into spacemacs-editing layer and install the packages if set in the dotfile. The dotfile should also document the features and be off by default.

rayw000:
Due to PR #14115, this one could be closed.

PR #14107, opened 2020-11-02
[org] Activate insert state after some org-insert-* commands

madand:
Having to manually press i after using any of these commands is annoying to say
the least. This commit affects the following commands:
- org-insert-drawer
- org-insert-heading
- org-insert-item
- org-insert-structure-template

Change to org-insert-heading also affects other commands that use it
internally (e.g. org-insert-subheading).


smile13241324 (approved):
Good idea, will be merged soon.

smile13241324:

PR #14108, opened 2020-11-02
add org-roam tag-add and tag-delete +prefix keybinding

snuffop:
Add key bindings for org-roam-tag-add and org-roam-tag-delete


snuffop:
Forgot documentation and had snafu with updating this PR so created additional closed this one.

PR #14120, opened 2020-11-08
Add grip-mode to github layer

dalanicolai:
Grip-mode renders a preview of github README files locally, so you can check how they look before pushing to github.
I figured that more people would like it, and that it would be great to have this as standard functionality in the github layer.
I've added it under keybindings `SPC g h p` (toggle grip mode)

### DIFF
--- a/layers/+source-control/github/README.org
+++ b/layers/+source-control/github/README.org
@@ -11,10 +11,12 @@
   - [[#layer][Layer]]
   - [[#git-configuration][Git configuration]]
   - [[#forge][Forge]]
+  - [[#grip-mode][grip-mode]]
 - [[#key-bindings][Key bindings]]
   - [[#forge-1][Forge]]
   - [[#gistel][gist.el]]
   - [[#clone-repositories][Clone repositories]]
+  - [[#grip-mode-1][Grip-mode]]
 
 * Description
 This layers adds support for [[http://github.com][GitHub]].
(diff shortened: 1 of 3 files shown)


smile13241324 (changes requested):
Can you explain a bit more about the preference of =grip-mode= vs =vmd-mode=? We can have both but right now they look pretty identical. However both should be located in the `markdown` layer and be bound to a mode specific binding for previewing in addition to a org mode binding possibly under `SPC m T m` similar to what we have with previewing Latex.

dalanicolai:
First I did not know about vmd-mode. But now that I have looked at it, I have noticed a few things:
1. vmd-mode only supports markdown preview, although probably quite straightforward, a function to preview org-mode files has not been included yet.
2. vmd-mode is installed using `npm` while grip is installed using `pip`. In this respect I think maybe both should be supported, so that users can use the package they prefer (personally I am not very familiar with npm, while pip I know quite well)

A third point is that, although `grip-mode` renders markdown or org files, for me it is more logical that this functionality belongs to "github tools/layer" and not really to "markdown/org tools/layers". There is no real no need to have the markdown layer or org layer installed for writing github README's, a user could just use `grip-mode` directly. On the other hand, I somewhat agree with you too, but if it is only added to markdown/org layers then I think that at least the github preview functionality should be mentioned/referred to from the github layer.

Overall, I think users are welcome to add vmd-mode functionality to the org layer and refer to it from the github layer. But I think the proposal in this PR is a quite small and welcome addition (only a small file and a single keybinding). Also I find it quite intuitive to add it under the `SPC g h`, but I agree that markdown/org major-mode keybindings would be nice too. Although in that case it might be more intuitive to create a prefix `preview` under `SPC m p` and also move the latex preview there...

layers/+source-control/github/README.org
```diff
@@ -55,6 +58,10 @@ dotfile:
   (dotspacemacs-additional-packages '((forge :toggle t)))
 #+END_SRC
 
+** grip-mode
+Grip-mode [[https://github.com/seagle0128/grip-mode#prerequisite][requires python and the python package grip]] to be installed on the
+system. Grip can usually be installed with the command =pip3 install grip=.
```
smile13241324, line 63: This should definitely refer to "pip" not "pip3". "pip3" is only available on debian based systems.

smile13241324 (approved):
I think you are right, the current proposal is straight forward and will be helpful for most people for now. We still should try not to duplicate functionality but we can revise this later.

layers/+source-control/github/README.org
smile13241324: Never mind I will change it myself

PR #14128, opened 2020-11-12
Add support for viewing latex fragment in org files

FieryCod:
I found that there is no keybinding for previewing latex fragments on demand, therefore I had to revert buffer over and over again.


lebensterben:
should this be a Toggle ( under SPC m T ) ?

FieryCod:
@lebensterben I don't think so due to the fact that the function will be called multiple times in buffer to expand latex fragments. It's not like one global switch.

lebensterben:
I think you need to update the key binding section in layer README.org

smile13241324 (changes requested):
There is already a binding for this, its under `SPC m T x` and will toggle the preview of either the latex fragment under point or all fragments in the document.

I think this PR is then obsolete.

PR #14130, opened 2020-11-12
Add tR hotkey for org-table-recalculate-buffer-tables

iarebatman:
Add tR hotkey for org-table-recalculate-buffer-tables. Without this
hotkey, there isn't an easy way to recalculate all of the calculated
fields within a table at once.

The tr hotkey for org-table-recalculate is only for recalculating a
single field within the table.

I used this page from the manual as a reference:
https://orgmode.org/manual/Updating-the-table.html#Updating-the-table

### DIFF
--- a/layers/+emacs/org/README.org
+++ b/layers/+emacs/org/README.org
@@ -473,36 +473,37 @@ are also available.
 
 ** Tables
 
-| Key binding   | Description                                                                |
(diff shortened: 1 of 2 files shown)


smile13241324 (approved):
Good addition to the existing functionality, will be merged soon.

Issue #14102, opened 2020-10-31
Section 5 (on debugging) of emacs-lisp layer doc is faulty 


dalanicolai:
#### Description :octocat: The section on emacs-lisp debugging in the emacs-lisp
layer doc
[here](https://develop.spacemacs.org/layers/+lang/emacs-lisp/README.html#debugging-elisp)
seems faulty.

First it says to use `, '` withing the source block to open another
emacs-lisp-mode buffer, while `, '` is bound to `IELM`. Later it says to put
cursor on the parenthesis and press `, e e`. But in evil mode, following this
step only evaluates the "spacemacs" string. Furthermore if someone just
copy-pasted the code to a file then it is not even possible to put the cursor after
the parenthesis for pressing `, '`, as there is no whitespace there for placing
the cursor, and also pressing `A` is not working because the buffer is in
read-only mode.

Proceeding evaluating with `, e f` (instead of `, e e`, and then pressing `s` to
next step up seems to not do what is meant in the doc (it places the cursor
behind the sexp, instead of going to some opening parentheses). And then also
`i` does not do what is meant.

It would be great if a more experienced developer, could fix that section. I
would have liked to do it, but I do not understand what is meant.

#### Reproduction guide :beetle:
- Start Emacs
- follow the steps
[here](https://develop.spacemacs.org/layers/+lang/emacs-lisp/README.html#debugging-elisp)
using evil-mode

#### System Info :computer:
- OS: gnu/linux
- Emacs: 27.1
- Spacemacs: 0.300.0
- Spacemacs branch: specify_configuration_layer_info_and_add_cross_links (rev. 58b1d7046)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
(centaur-tabs auto-completion emacs-lisp common-lisp clojure rust python lua jupyter
              (git :variables magit-repository-directories
                   '(("~/git" . 1)))
              (ivy :variables ivy-enable-advanced-buffer-information t)
              fasd
              (ranger :variables ranger-override-dired 'ranger ranger-show-preview t)
              lsp pdf djvu
[...]

PR #14132, opened 2020-11-14
Add instructions to open emacs-lisp layer README

dalanicolai:
in response to issue #14102

### DIFF
--- a/layers/+lang/emacs-lisp/README.org
+++ b/layers/+lang/emacs-lisp/README.org
@@ -66,10 +66,16 @@ As this state works the same for all files, the documentation is in global
 [[https://github.com/syl20bnr/spacemacs/blob/master/doc/DOCUMENTATION.org#lisp-key-bindings][DOCUMENTATION.org]]. In general, use ~SPC k~ to interact with the lisp-state.
 
 * Debugging Elisp
-Here is an interactive quick start to debug Emacs Lisp.
+Here is an interactive quick start to debug Emacs Lisp from an =emacs-lisp-mode= buffer.
 
-First you need to read this file in Emacs in order to have the following code
-block to be interactively opened as Emacs Lisp.
+To follow this tutorial open the code of the following source block in an
+=emacs-lisp-mode= buffer. One way to do this is to first find this emacs-lisp
+layer documentation file in Spacemacs using ~SPC h l emacs-lisp RET~. This will
+open the current file in =space-doc-mode=. Toggle off =space-doc-mode= using =,
+T V= so that Spacemacs will switch to =org-mode=. Now you can place the cursor
+somewhere inside the code block and press ~, '~ to open the code in an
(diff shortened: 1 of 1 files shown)


lebensterben:
I think the change is misleading.
The layer documentation isn’t talking about how to debug the example code block per se, but about debugging elisp files in general.
So the original instruction of opening the script is correct, but the new instruction "SPC h l emacs-lisp RET" would open the layer doc and thus incorrect in general cases.

layers/+lang/emacs-lisp/README.org
```diff
@@ -68,8 +68,9 @@ As this state works the same for all files, the documentation is in global
 * Debugging Elisp
 Here is an interactive quick start to debug Emacs Lisp.
 
-First you need to read this file in Emacs in order to have the following code
-block to be interactively opened as Emacs Lisp.
+First you need to read this file in Spacemacs (`SPC h l emacs-lisp RET` and then
```
smile13241324, line 71: Please don't intermix markdown and org formatting, this is don't use backticks instead use `~` for text formatting. This will be translated correctly for display in github.

layers/+lang/emacs-lisp/README.org
```diff
@@ -68,8 +68,9 @@ As this state works the same for all files, the documentation is in global
...
+toggle off with `, T V`) in order to have the following code block to be
```
smile13241324, line 72: I think to make stuff clearer you also need to explain what your single commands do.

smile13241324 (changes requested):
I like the additions in general. The new description talks about the small emacs lisp debugging tutorial so refering to this file is ok in my opinion given that we have some statement saying that these instructions are specific for this tutorial case and are not the default way to debug lisp of course.

I have found two issues were markdown formatting has been intermixed with org file. Please change these to org formatting.

dalanicolai:
> I think the change is misleading.
> The layer documentation isn’t talking about how to debug the example code block per se, but about debugging elisp files in general.
> So the original instruction of opening the script is correct, but the new instruction "SPC h l emacs-lisp RET" would open the layer doc and thus incorrect in general cases.

The original instructions say, read THIS FILE in Emacs, and subsequently it continues instructing how to open the code in an emacs-lisp-mode buffer from the source code block. So the original documentation is definitely talking about the example code block. However it forgot to instruct users where to find/how to open the file. I have only added instruction/keybindings for that.

dalanicolai:
Of course indeed I only added the keybinding instructions where I mistakenly used markdown syntax unconsciously as it gets confusing sometimes to work on github, gitter and spacemacs docs at the same time.

Anyway, I agree with you that the instructions can be improved a little more, and so I have updated the PR accordingly. Let me know if you have any other suggestions...

smile13241324 (approved):
Cool, that's more than expected, well done @dalanicolai :+1:

smile13241324:

PR #14135, opened 2020-11-14
Move forge from github to git

fabianhjr:
Supports:
- Github (Both .com and enterprise on premises)
- Gitlab (Both .com and self-hosting)
- gogs / gitea

https://github.com/magit/forge/

### DIFF
--- a/layers/+source-control/git/README.org
+++ b/layers/+source-control/git/README.org
@@ -15,6 +15,7 @@
   - [[#global-git-commit-mode][Global git commit mode]]
   - [[#git][Git]]
   - [[#git-flow][Git-Flow]]
+  - [[#forge][Forge]]
   - [[#org-integration][Org integration]]
 - [[#working-with-git][Working with Git]]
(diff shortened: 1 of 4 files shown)


smile13241324 (changes requested):
I have double checked your changes and noticed that the package you have integrated is already owned by the github layer. If it is able to handle gitlab too, please move it from there to the git layer.

Also remember that just adding a package to the package.el file does not load or install it, it always requires a init function, see doc/LAYERS.org for details.

fabianhjr:
Thanks for the heads up, will take a look.

fabianhjr:
Hi @smile13241324, I amended the commit to be a move of forge form github into the git layer.

smile13241324 (approved):
Looks good :+1: , will be merged soon.

smile13241324:

fabianhjr:
Nice, thanks @smile13241324 UwU

PR #14136, opened 2020-11-15
Warn user they are reading develop docs

merimna:
The intent of this PR is to let users know they are reading documentation that might not apply to their version of spacemacs. The default install is based on master, but the docs for both master and develop show up in search engine results.

Currently the only way to see that you are reading the develop docs is by looking at the subdomain. You'd also need to know that a non-develop website exists, which is not obvious. I'm hoping this will help new users have a smooth experience when trying to figure out how spacemacs works.

This PR is my first shot at adding in a warning banner for users. I'm still not 100% sure I understand how the website is generated, but my current guess is

1. The org docs are converted into HTML-like sdn files via spacetools
1. spacetools are used to convert the sdn files _back_ into org docs
1. and then the code here turns the HTML-ready org files into actual HTML?

#### Questions

1. Is this actually the code used to generate the website?
1. Can/Should we only show this if we are not on the `master` branch? Doing so would make the eventual merge of develop easier. I'm not sure what branch information is available at HTML doc build time.

#### Sample search engine results
image

#### Mockup of the warning
image


smile13241324:
@syl20bnr I think this would be a question to you, how we should proceed with the Website.

JAremko:
Currently specetools only responsible for reformatting documentation files and creating layers.org file. `core-documentation.el`  responsible for creating .html files from .org files.

JAremko:
This is how we call it https://github.com/syl20bnr/spacemacs/blob/develop/.travisci/pub_prep.sh#L86

merimna:
Ah, perfect! I got it to export locally by

1. Cloning the repo to a directory named `.emacs.d`
1. Running `HOME=<parent dir of .emacs.d from above> emacs -batch -l init.el -l core/core-documentation.el -f spacemacs/publish-doc -Q`

It looks pretty good with the `develop` CSS files. The link has two behaviors:

1. Javascript enabled, will redirect to the same path but on `www.spacemacs.org`
1. Javascript disabled, will redirect to `www.spacemacs.org/doc/DEVELOPMENT.html`

#### GIF of functionality
output

merimna:
Gonna see if there's anything I can do about only adding this for non-master branches, and getting fancy by replacing `develop` with the actual branch name.

JAremko:
This PR also need re-base.

smile13241324:
I think this PR is obsolete with the latest changes from #14151. Now, with the develop branch being the primary one we do not longer have to warn users from using it.

merimna:
Agreed that this isn’t necessary if we’re getting rid of `develop.spacemacs.org`. If the domain is staying though, and master can drift away from develop, then I think this is still useful. 

How would y’all feel about me taking a stab at generating a site that can handle multiple versions? Something with a drop down selector to switch between tags or a specific list of branches.

lebensterben:
Even though the `develop` branch became the default, we still can benefit from a the warning because by definition `develop` branch has no guarantee of stable interface.
That means users should be well informed that, anything they see in the doc could be be outdated any time. Features, UI, and key-bindings, may change any time without notice.

smile13241324:
Just read the messages on the maintainer chat and it looks like my previous comment was wrong, the develop branch as primary branch is just a workaround, when the feature "stable-elpa" will be ready the master will be set as primary branch and there will be a regular release.

Ok with this being said, this PR is not obsolete, please go ahead @merimna your help would be very much appreciated.

altsalt:
I hope that this issue (and the related conversation about which branch is canonical) is resolved soon. Copying over my thoughts from an open issue, as an example of why this is important to make clear.

> Wow, while reading through the documentation and considering switching to Spacemacs, I came across a broken link to one of the layers. Just dropped by to submit a quick PR and my pre-search found this issue. Glad I didn't put in too much effort, but to answer your question @blablablerg , yes, people absolutely use the main branch, since that's what anyone unfamiliar with the project is directed to.

_Originally posted by @altsalt in https://github.com/syl20bnr/spacemacs/issues/14420#issuecomment-826141878_

lebensterben:
@JAremko 
I've rebased this PR.

smile13241324:
@syl20bnr @JAremko @duianto @lebensterben @sdwolfz 

I think its time to make a decision what to do with our stale master. Either we should have the next release, with or without spacelpa, or we should be consistent and change the websites to point to the develop branch, clearly communicating that we switched to a rolling release and do not longer provide stable ones.

Right now the communication is unclear causing a lot of confusions on new users.

Also we should make a final decision about this PR. I would like to avoid having PRs wait forever, rather we should give a clear feedback whether it will be merged or not. On the long run this is much more motivating for contributors than the current approach of waiting till a PR is stale.

lebensterben:
This PR should not conflict with versioning.

If we want we can easily change the warning message, from indicating it's develop branch, to indicating it's a certain release.

duianto:
>I think its time to make a decision what to do with our stale master. 

>we should be consistent and change the website

👍

lebensterben:
I still believe we should merge this PR.
It's quite simple. And we can remove it easily in the future without breaking anything.

smile13241324 (approved):
I really like this small change, I think we have waited long enough for news about the master branch. Lets have this merged.

PR #14138, opened 2020-11-15
Fix major-mode set key prefix documentation

dalanicolai:
The documentation for setting a prefix for major-mode keybinding is wrong.
Although it would be logical when using `spacemacs/declare-prefix-for-mode` to omit the "m" (for `SPC m` or `,`) in the keybinding definition, it currently does not work that way. Omitting the "m" results in a prefix named "prefix" instead of the desired prefix name. Because all (correct functioning) layers (e.g. org and pdf layer) define the prefix including the "m", it is probably best to just correct the documentation (instead of correcting the `spacemacs/declare-prefix-for-mode` function definition)


smile13241324 (approved):
Great catch! Will be merged soon.

PR #14141, opened 2020-11-17
[core][tide][lsp] improve spacemacs/set-leader-keys-for-minor-mode and apply it to tide and lsp layers

thanhvg:
There are three commits in this PR. The first one is the most important, the other two are demonstration of the improvement.

## Commit 1
This commit added add a new function defun `spacemacs/declare-prefix-for-minor-mode` and improved `spacemacs/set-leader-keys-for-minor-mode`.

`which-key` package recently introduced a new api `which-key-add-keymap-based-replacements` which improves performance and allows prefix and namings to be stored directly in keymap. This is a great improvement.

With this new api we now make change to `spacemacs/declare-prefix-for-minor-mode` to manage prefix also. For example:
``` elisp
  (spacemacs/set-leader-keys-for-minor-mode 'lsp-mode
    "=" "format"
    "=b" #'lsp-format-buffer)
````
Before we had to use another api to bind prefix `spacemacs/declare-prefix-for-mode` which only works on major-mode. As `lsp-mode` is a minor mode this api causes a lot of problems to `which-key` performance. An example is https://github.com/syl20bnr/spacemacs/issues/12455 which led to my hack in https://github.com/syl20bnr/spacemacs/pull/12474.

The improved `spacemacs/set-leader-keys-for-minor-mode` will take care of both prefix and key naming for the minor mode. This will allows us to have a better set up for dynamic minor modes such as `lsp-mode`, `tide-mode` etc.

Also another api is created to make prefix for minor mode: `spacemacs/declare-prefix-for-minor-mode`.

Usage:
```` elisp
(spacemacs/declare-prefix-for-minor-mode 'tide-mode "E" "errors")"
````
## Commit 2 
Apply improved api to tide layer

## Commit 3 
Apply improved api to lsp layer

### DIFF
--- a/core/core-keybindings.el
+++ b/core/core-keybindings.el
@@ -166,12 +166,28 @@ they are in `spacemacs/set-leader-keys'."
 `dotspacemacs-major-mode-emacs-leader-key' for the minor-mode
 MODE. MODE should be a quoted symbol corresponding to a valid
 minor mode. The rest of the arguments are treated exactly like
-they are in `spacemacs/set-leader-keys'."
+they are in `spacemacs/set-leader-keys'. If DEF is string, then
+it is treated as a prefix not a command."
   (let* ((map (intern (format "spacemacs-%s-map" mode))))
     (when (spacemacs//init-leader-mode-map mode map t)
-      (while key
-        (define-key (symbol-value map) (kbd key) def)
-        (setq key (pop bindings) def (pop bindings))))))
+      (let ((map-value (symbol-value map)))
+        (while key
+          (if (stringp def)
+              (which-key-add-keymap-based-replacements map-value key def)
+            (define-key map-value (kbd key) def))
+          (setq key (pop bindings) def (pop bindings)))))))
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

**Comment by smile13241324** [comment]
@JAremko this sound interesting, I wonder however if this doesn't clash with the new spacebind api, care to have a look?

**Comment by JAremko** [comment]
Seems like it doesn't override old functionality so there shouldn't be any problem with that.
Not sure how useful it will be tbh **when** we migrate to spacebind but shouldn't harm.

**Review by smile13241324 — APPROVED** [review]
Woa, this is pretty cool, I will merge it right away.

**Comment by smile13241324** [comment]

**Comment by Claude-Ray** [comment]
Hi, I really appreciate this PR. But it left a warning on `spacemacs//tide-setup-jump-handle`
```
Error (use-package): tide/:config: Symbol’s value as variable is void: tide-managed-modes
```
Hope you can notice this. Thanks!

thanhvg:
@Claude-Ray thanks for letting me know, I just added a new PR for this fix. 

I'm sorry for this, I have my own fork of Spacemacs and I missed this tide part when I ported some of its features back here in this PR.

PR #14146, opened 2020-11-18
[core] add z$ as Evil keybinding for set-selective-display

Kazark:
This is a feature whose equivalent I had looked for previously and not found. I bumped into it in [this excellent article](https://karthinks.com/software/batteries-included-with-emacs/) which is a survey of "batteries included" parts of Emacs. He mentions Spacemacs (in a positive light) in the article, but the article is about what you can do _without_ distributions like Spacemacs, Doom, or Prelude. As such, this seems like an opportunity to make Spacemacs even better---i.e. by making these relatively unknown batteries that are already present in Emacs more discoverable in Spacemacs.

I put it in the folding menu `z` because that fits its usecase. I used `$` because:
* the out-of-the-box Emacs keybindings is `C-x $`.
* the menu is already fairly saturated
* I didn't know what else to base the mnemonic on because
  * I'm not aware of a Vim equivalent
  * The name of the Emacs function itself is nondescriptive of its behavior

I couldn't find any keybinding documentation table to add it to. Please let me know if I missed something.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -521,6 +521,10 @@ Other:
     terminal mode, but no longer in GUI mode. (emacs18)
 *** Core changes
 - Improvements:
+  - Make =set-selective-display=, which is a kind of built-in Emacs quick-n-dirty
+    whole-file folding mechanism for indentation-based syntaxes, more
+    discoverable for Evil users by placing it in the folding menu as ~z f~ (thanks
+    to Keith Pinson)
   - Provide keybindings for using narrow with an indirect buffer for when you
     want narrowing multiple times in the same base buffer:
     - ~SPC n F~ for narrow to function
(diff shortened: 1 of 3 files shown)


lebensterben:
Since you've brought this up, I think we should instead use a simple wrapper function of "set-selective-display".

It's explained [here](https://www.emacswiki.org/emacs/HideShow#toc5) and first proposed by [jao](http://emacs.wordpress.com/2007/01/16/quick-and-dirty-code-folding/).

Specifically, we can define this function
```elisp
(defun toggle-selective-display (column)
  (interactive "P")
  (set-selective-display
   (or column
       (unless selective-display
         (1+ (current-column))))))
```
And bind it to "z $". If you'd like, you can name it as "toggle-folding-by-indentation"

Kazark:
@lebensterben interesting, I'll give that a spin.

Kazark:
@lebensterben Thanks, it's a definite improvement. I've put those changes in if you want to take 'em for a spin.

smile13241324 (approved):
Cool, indeed this makes spacemacs even better.

gondalez:
I just tried this after seeing it come in from `develop`. Very handy - thank you!

PR #14152, opened 2020-11-20
[clojure] add cider-inspect key bindings

jr0cket:
Add cider inspect key bindings missing from Spacemacs


smile13241324 (approved):
Looks good

PR #14156, opened 2020-11-22
Add dired-quick-sort layer

xuhdev:
This layer helps quickly sorting Dired buffers in various ways, such as
time, size, etc.
I think the underlying package is useful and worth a place in spacemacs.
It has also been promoted by Pragmatic Emacs before (albeit long time
ago): http://pragmaticemacs.com/emacs/speedy-sorting-in-dired-with-dired-quick-sort/

Disclaimer: I'm the author of dired-quick-sort package

### DIFF
--- a/layers/+spacemacs/spacemacs-editing/README.org
+++ b/layers/+spacemacs/spacemacs-editing/README.org
@@ -29,3 +29,4 @@ This layer adds packages to improve editing with Spacemacs.
 - Support for conversion between Emacs regexps and PCRE regexps.
 - Support for persistent scratch via =persistent-scratch=.
 - Support for unkillable scratch via =unkillable-scratch=.
+- In =dired-mode=, press ~Shift S~ to select sorting.
--- a/layers/+spacemacs/spacemacs-editing/packages.el
+++ b/layers/+spacemacs/spacemacs-editing/packages.el
@@ -14,6 +14,7 @@
         avy
         (bracketed-paste :toggle (version<= emacs-version "25.0.92"))
         (clean-aindent-mode :toggle dotspacemacs-use-clean-aindent-mode)
+        dired-quick-sort
         editorconfig
         eval-sexp-fu
         expand-region
@@ -98,6 +99,12 @@
     (progn
       (clean-aindent-mode)
       (add-hook 'prog-mode-hook 'spacemacs//put-clean-aindent-last t))))
+
+(defun spacemacs-editing/init-dired-quick-sort ()
+  (use-package dired-quick-sort
+    :defer t
+    :init
+    (dired-quick-sort-setup)))
 
 (defun spacemacs-editing/init-editorconfig ()
   (use-package editorconfig


smile13241324 (changes requested):
I like the package however introducing a single package layer for it will create a lot of clutter which I want to avoid especially as dired is always included so I see no reason why somebody would not want to have these bindings in place.

Can you move your package to `spacemacs-editing` layer instead?

xuhdev:
@smile13241324 Good point, updated

smile13241324 (approved):
Looks good

smile13241324:

winsphinx:
image
This causes this warning, and how can I solve it?

xuhdev:
@winsphinx To suppress this warning, set `ls-lisp-use-insert-directory-program` to t. I've updated this warning message to the latest version. Otherwise you can either leave it there (in which case the package will do nothing) or disable the package.

joebalancio:
@xuhdev I'm getting a similar message on startup after pulling the latest from develop.

image

SadSock:
> @xuhdev I'm getting a similar message on startup after pulling the latest from develop.
> 
> image

I also get this message on startup.

macos 11.0.1

xuhdev:
If you don't have GNU ls installed, the only way to suppress it is to disable the extension. @smile13241324 Do you think it makes sense to add a warning suppression option for this package to spacemacs? If so, I'm struggling with coming up with a reasonable default.

duianto:
The warning:
```
Warning (dired-quick-sort): `ls-lisp-use-insert-directory-program'
is nil. The package `dired-quick-sort' will not work and thus is not set up by
`dired-quick-sort-setup'. Set it to t to suppress this warning.
```
appears in Windows, because the variable is set to `nil` in Windows.
https://github.com/emacs-mirror/emacs/blob/3af9e84ff59811734dcbb5d55e04e1fdb7051e77/lisp/ls-lisp.el#L194-L201
```elisp
(defcustom ls-lisp-use-insert-directory-program
  (not (memq system-type '(ms-dos windows-nt)))
  "Non-nil causes ls-lisp to revert back to using `insert-directory-program'.
This is useful on platforms where ls-lisp is dumped into Emacs, such as
Microsoft Windows, but you would still like to use a program to list
the contents of a directory."
  :type 'boolean
  :group 'ls-lisp)
```

Should we set the variable: `ls-lisp-use-insert-directory-program`
it to `t` in Windows?

Possibly like this:
```diff
1 file changed, 4 insertions(+), 1 deletion(-)
layers/+spacemacs/spacemacs-editing/packages.el | 5 ++++-

modified   layers/+spacemacs/spacemacs-editing/packages.el
@@ -105,7 +105,10 @@
   (use-package dired-quick-sort
     :defer t
     :init
-    (dired-quick-sort-setup)))
+    (progn
+      (when (memq system-type '(ms-dos windows-nt))
+        (setq ls-lisp-use-insert-directory-program t))
+      (dired-quick-sort-setup))))
 
 (defun spacemacs-editing/init-editorconfig ()
   (use-package editorconfig
```

I don't have a second computer to test what happens on a clean Windows installation.

But for me `dired-quick-sort` works.
`S-s` shows the hydra and the sorting options work as expected.

The variable `insert-directory-program` has the value `ls`.

`dired-quick-sort` checks for `GNU` here:
https://gitlab.com/xuhdev/dired-quick-sort/-/blob/9741bae65b5ca077a81589dcc6b3ca3c12f0f76f/dired-quick-sort.el#L193-194
```elisp
           (call-process insert-directory-program nil t nil "--version")
           (string-match-p "GNU" (buffer-string))
```

For me, the expression:
```elisp
(call-process insert-directory-program nil t nil "--version")
```
returns:
```
ls (GNU coreutils) 8.26
Packaged by Cygwin (8.26-2)
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Richard M. Stallman and David MacKenzie.
```

and
```
(string-match-p "GNU" (buffer-string))
```
returns:
```
4 (#o4, #x4, ?\C-d)
```

xuhdev:
I added an option `dired-quick-sort-suppress-setup-warning` to suppress these warnings: https://gitlab.com/xuhdev/dired-quick-sort/-/commit/e22a0c26a7726032eb6674976e3e22ea7de30a5a

We can consider setting it to `t` by default, but it is not ideal as it would lose the point of having a warning in the first place (i.e., system environment won't let this extension work). What do you think?

duianto:
The `dired-quick-sort-suppress-setup-warning` variable seems to have the opposite behavior.

It's `nil` by default, but the warning isn't shown on startup now.

The warning returns on startup, when the variable is enabled in `dotspacemacs/user-init`:
```
(setq dired-quick-sort-suppress-setup-warning t)
```

### A possible solution:
Change `when` to `unless`:
```elisp
(when dired-quick-sort-suppress-setup-warning
```
```elisp
(unless dired-quick-sort-suppress-setup-warning
```

---

It might be helpful to add a `Requirements` section in the package readme. 
Currently the `GNU ls` requirement is mentioned in the middle of the `Configuration` section.

---

Your right, disabling the warning would silently disable the package for anyone that doesn't meet the requirements.

Showing a Warnings buffer on startup when the package won't work, does bring it to peoples attention, but it's quite an interruption, just to inform people that they won't be able to sort more effectively a dired buffer.

Might it be better to load the package when a dired buffer is opened, instead of on Emacs startup?

And would it be enough to show the warning in the minibuffer (it would also appear in the messages buffer)?

xuhdev:
@duianto Thanks, I fixed the issue regarding the new configuration (https://github.com/xuhdev/dired-quick-sort/commit/bd113308b0e2445a5256af6fda2e7966e0e5061d) and added a requirement section to README (https://github.com/xuhdev/dired-quick-sort/commit/56d25d81d7bf09db47969a0e7a54524b0678c1b5).

> And would it be enough to show the warning in the minibuffer (it would also appear in the messages buffer)?

Currently it uses `display-warning`: https://github.com/xuhdev/dired-quick-sort/blob/bd113308b0e2445a5256af6fda2e7966e0e5061d/dired-quick-sort.el#L194

I can change it to `message`, and let `dired-quick-sort-suppress-setup-warning` control whether it should be `message` or `display-warning`. Then, spacemacs can set `dired-quick-sort-suppress-setup-warning` to t by default (`message` is used). Would this be a satisfying solution from spacemacs' perspective?

duianto:
Yes, it would be less of an interruption to have an option to show the message in the minibuffer.

Spacemacs has a transient hook function, that removes the hook when it runs, so we can set it to only show the message the first time a dired buffer is opened.
```elisp
(defun spacemacs-editing/init-dired-quick-sort ()
  (use-package dired-quick-sort
    :defer t
    :init
    (spacemacs|add-transient-hook dired-mode-hook
      (lambda () (dired-quick-sort-setup)))))
```
I changed `display-warning` to `message` and it looks like this.

emacs_2020-12-14_15-17-53

xuhdev:
@duianto Thanks! Please see #14214 for the update

duianto:
The `evil-collection-dired` bindings were added recently:
[evil-collection] support for dired #14333

It replaced the default dired-quick-sort hydra binding: `S`
to: `S` `dired-do-symlink`

But it also restored the default evil behavior of: `s` `evil-substitute`

Since it isn't useful in the dired buffer since it's: read-only

A PR is pending to bind the `dired-quick-sort` hydra to: `s`
Bind s to the dired-quick-sort hydra #14415

PR #14158, opened 2020-11-23
Revert helm-ag and helm-grep search results window binding of RET to default

kini:
Currently 'helm-ag-mode-jump-other-window is bound both to RET and to C-o (its
original default binding), while 'helm-ag-mode-jump isn't bound at all (its
original default binding is RET).  This commit removes our customization so that
both 'helm-ag-mode-jump-other-window and 'helm-ag-mode-jump have bindings in the
mode.

### DIFF
--- a/layers/+completion/helm/packages.el
+++ b/layers/+completion/helm/packages.el
@@ -211,7 +211,6 @@
 
       ;; evilify the helm-grep buffer
       (evilified-state-evilify helm-grep-mode helm-grep-mode-map
-        (kbd "RET") 'helm-grep-mode-jump-other-window
         (kbd "q") 'quit-window)
 
       (spacemacs/set-leader-keys
@@ -273,7 +272,6 @@
       (advice-add 'helm-ag--save-results :after 'spacemacs//gne-init-helm-ag)
       (evil-define-key 'normal helm-ag-map "SPC" spacemacs-default-map)
       (evilified-state-evilify helm-ag-mode helm-ag-mode-map
-        (kbd "RET") 'helm-ag-mode-jump-other-window
         (kbd "gr") 'helm-ag--update-save-results
         (kbd "q") 'quit-window))))
 


smile13241324 (changes requested):
I don't agree with your change, when you have a look at helm/packages.el you will notice that for all helm based functions the `*-other-window` version is called with `RET`. Also `C-o` is definitely unbound for `Evil` users.

Do you have a specific use-case where the current binding pose a problem?

kini:
> I don't agree with your change, when you have a look at helm/packages.el you will notice that for all helm based functions the `*-other-window` version is called with `RET`. Also `C-o` is definitely unbound for `Evil` users.
> 
> Do you have a specific use-case where the current binding pose a problem?

Hmm. `C-o` is not unbound for me and I am an Evil user. This is for the keymap that's used in the `*helm ag results*` buffer that appears when you do (e.g.) `SPC /`, search for something, and hit `<f3>`. I don't think I've configured anything differently that would affect this. Can you check again whether `C-o` is unbound for Evil users?

As for the other helm based functions, I only see one other `RET` binding to a `*-other-window` function, which is the binding of `helm-grep-mode-jump-other-window`, i.e. the analogue of `helm-ag-mode-jump-other-window` when using `grep` instead of `ag`. The `helm-grep` upstream package has analogous bindings to the `helm-ag` upstream package (see [here](https://github.com/emacs-helm/helm/blob/5ad6c83681fda7bfd5a745fedecaa924630253dd/helm-grep.el#L346-L347)) so I suppose it's more consistent to change that binding as well. I'll add that change to my commit, too.

The specific use case is that I hit `<f3>` during a search, and then when I have the search results buffer open, there's no key binding for `'helm-ag-mode-jump`. I could just add a key binding in my `.spacemacs`, sure, but since the upstream package provides a binding for that function already and Spacemacs deletes it for no apparent reason, I figured it's better to change Spacemacs in this case.

smile13241324:
Thanks for the explanation I will check

smile13241324 (approved):
Now I see your point, in this case you are right, the default for Return should be `jump-window`, the default binding `C-o` for `jump-other-window` should be enough for most people.

smile13241324:

PR #14159, opened 2020-11-23
[clojure] add namespace eval and reload key bindings

jr0cket:
Add key binding to evaluate the current namespace form (very useful when
starting CIDER as it loads in clojure.core to enable the cider-clojure-docs
functions to work)

Add keybinding to reload all namespaces into the REPL process

Adjusted key bindings to be lower case under `en` sub-menu.


smile13241324 (approved):
:+1:

Issue #14103, opened 2020-10-31
React layer does not inherit keybindings from javascript layer


dvzubarev:
I have added both layers to `dotspacemacs-configuration-layers` :
```
(javascript :variables javascript-repl 'nodejs
                 javascript-backend 'lsp
                 javascript-lsp-linter nil
                 js2-mode-show-strict-warnings nil)
 react
```
If open `temp.js`, than there is keybinding for `js2-mode-toggle-warnings-and-errors` - `,w`.
When opening `temp.jsx` there is no such keybinding (like many other keybindings that are defined in javascript layer).
`Rjsx-mode` is derived from `js2-mode`, so I would expect that all keybindings should be inherited. 
Emacs 27.1, Spacemacs latest develop

duianto:
The `javascript` layer keys are bound like this:
https://github.com/syl20bnr/spacemacs/blob/28a4baeea4abc2c55f728355ba2666e77ed78118/layers/%2Blang/javascript/packages.el#L105-L112

which creates a new keymap: `spacemacs-%s-map`

There are two key maps called:
```
spacemacs-js2-mode-map
spacemacs-rjsx-mode-map
```
but it seems like `.js` files use the lsp map:
```
spacemacs-lsp-mode-map-prefix
```

Because in a `.js` file, `SPC ?` shows:
```
,             spacemacs-lsp-mode-map-prefix
```
and in a `.jsx` file, `SPC ?` shows:
```
,             spacemacs-rjsx-mode-map-prefix
```

The other maintainers suggested that it probably would be simplest to bind the keys explicitly in the `react` layer as well, and document them in the `react` layers readme.

https://github.com/syl20bnr/spacemacs/tree/develop/layers/%2Bframeworks/react

Could you open a pull request?

dvzubarev:
> but it seems like .js files use the lsp map:

But lsp keybindings are available in `.jsx` files, so I guess `.jsx` files use the lsp map too.

> The other maintainers suggested that it probably would be simplest to bind the keys explicitly in the react layer

There are many keybindings for js2-mode, see `init-js2-refactor`, `init-skewer-mode`, `init-nodejs-repl` and maybe more in the future. I don't think it is reasonable to copy all these keybindings to react layer.

duianto:
Your right.

A possible fix is pending:
[react] Set layer keymaps: javascript as parent to react #14163

duianto:
The PR has been merged.
You can close this issue if it's working as expected.

PR #14163, opened 2020-11-23
[react] Set layer keymaps: javascript as parent to react

duianto:
Declared key binding prefixes.
And updated the react layer documentation:
- Mentioned the additional key bindings in the javascript layer
- Added a link to the javascript layer

Fixes: 
React layer does not inherit keybindings from javascript layer #14103

Before:
before

After:
after


smile13241324 (approved):
:+1:

PR #14173, opened 2020-11-27
[git] fix magit-mode-map void value

thanhvg:
Minor fix for regression from https://github.com/syl20bnr/spacemacs/commit/be908875cb2d7f5f7210c19101aed01c747bb7df

use-package is lazy so post-init-magit will not see magit-mode-map value

### DIFF
--- a/layers/+source-control/git/packages.el
+++ b/layers/+source-control/git/packages.el
@@ -148,14 +148,6 @@
     (spacemacs/set-leader-keys
       "gfi" 'gitignore-templates-new-file)))
 
-(defun git/post-init-magit ()
-  ;; Remove inherited bindings from evil-mc and evil-easymotion
-  (which-key-add-keymap-based-replacements magit-mode-map
-    "<normal-state> g r" nil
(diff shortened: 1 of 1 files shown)


smile13241324:
Thanks but this should be fixed on develop already, could you have a small test and confirm @thanhvg?
Btw, I had it this way in my first version but strangly enough it did not work on my small linux box then, the current form seems to work there, but is failing on my bigger windows box in a docker container.

I hope the form done here https://github.com/syl20bnr/spacemacs/commit/be908875cb2d7f5f7210c19101aed01c747bb7df will work on all systems.

thanhvg:
@smile13241324 yeah I don't get the error anymore after rebasing with develop but I don't use `evil-magit` in my git layer so I can't see the intentional bindings. I think it should world for dev branch. 

I think the error you got with the other linux box is probably due to  the fact that `evil-magit` has been deprecated recently. It now  belongs to `evil-collection`.

I'm closing this PR.

Thanks.

PR #14176, opened 2020-11-29
[core] Fix package update restart message

duianto:
### Problem:
When the restart-emacs package is being updated.

The update packages restart message, still shows the (SPC q r) key binding even
though it doesn't work.

### Cause:
The member function compared a string against a list of symbols.

### A possible improvement:
Explain that the key binding won't work this time.

### Before:
Emacs has to be restarted to actually install the new version of the packages (SPC q r).

### After:
Emacs has to be restarted to actually install the new version of the packages
(SPC q r) won't work this time, because the restart-emacs package is being updated.


smile13241324 (approved):
Looks good.

Issue #14142, opened 2020-11-17
ui/font is too small in default .spacemacs even with all prerequisites  


subhajit-halder:
In the default installation the ui scale/font size is too small. (Even with all prerequisites)
So I found this;

> Emacs will set the font size in pixels if you specify an integer (13) but in points if you specify a real (13.0)

[reddit source](https://www.reddit.com/r/spacemacs/comments/gi7t3r/hi_font_or_resolution_way_too_small_how_can_i_fix/fqe26p4?utm_source=share&utm_medium=web2x&context=3)

I didn't see this mentioned in the documentation, can you please add this. And in the default configuration file can we have this as well. Or is this the intended behaviour?

I have attached two ss one with font size: 14 and other with size: 14.0
![size: 14](https://user-images.githubusercontent.com/54535412/99350562-b7f63880-28c4-11eb-8d8a-5d726aaebecd.png)

![size:  14.0](https://user-images.githubusercontent.com/54535412/99350616-da885180-28c4-11eb-9166-5fdb817d1b6a.png)

jr0cket:
Checking the `core/templates/.spacemacs.template` it seems the `master` branch is using an integer value, `13`.  The `develop` branch uses a real number of `13.0`, so this is probably fixed in develop if I understand correctly.

https://github.com/syl20bnr/spacemacs/blob/master/core/templates/.spacemacs.template#L135-L139

duianto:
The font size, integer vs floating-point distinction is mentioned in the `develop` branch documentation.
https://github.com/syl20bnr/spacemacs/blob/develop/doc/DOCUMENTATION.org#font
>:size The font size- either a non-negative integer that specifies the pixel size, or a floating-point number that specifies the point size.

A PR has been opened, to also mention it in the `.spacemacs` font variables comment.
[core] Update dotfile, font :size comment: int vs float #14183

lebensterben:
It's worth noting that both point-size and pixel-size plays an role in font rendering

https://www.x.org/releases/X11R7.6/doc/xorg-docs/specs/XLFD/xlfd.html

> PIXEL_SIZE usually is used by X client applications that need to query fonts according to device-dependent size, regardless of the point size or vertical resolution for which the font was designed.

> POINT_SIZE and RESOLUTION_Y are used by X clients to query fonts according to device-independent size to maintain constant text size on the display regardless of the PIXEL_SIZE used for the font.

[...]

PR #14183, opened 2020-12-03
[core] Update dotfile, font :size comment: int vs float

duianto:
Add a comment about the font :size difference between:
integer (pixel size)
floating-point (point size)

Point size is recommended, because it's device independent.

source:
https://www.x.org/releases/X11R7.6/doc/xorg-docs/specs/XLFD/xlfd.html

Thanks libensterben for the info about point size being device independent.
#14142 (comment)

The font size comment was requested in:
ui/font is too small in default .spacemacs even with all prerequisites #14142


smile13241324 (approved):
Great comment, this will make the different meanings much more clear.

Issue #2828, opened 2015-08-30
Improve Info reading experience


xfq:
I'm aware of #902, but using Info mode in Spacemacs is still a PITA.  For example:
- I can't use `l` to go back to the last node I was at.
- `b` goes to beginning of node instead of the beginning of the previous word, and I can't use `g g` to go to the first line (of the node).
- Visual mode can't be entered with `v`.
- `n` doesn't repeat the last search.

sooheon:
Yep, this is one of those cases where the modal keys defined by info-mode and evil really clash, and it's unclear which is the best way to reconcile them. FWIW, I think that `l` behaviour is replicated by `C-o`, but then `C-i` is something different, and there isn't any kind of mnemonic symmetry like `l` and `r` or `C-i` and `C-o`. It's just a complete mixup right now.

I actually think the modal keys of info-mode are well thought out, and prefer just learning them and minimizing vi keys, but others may disagree. Just getting used to emacs keys isn't bad (i.e. `M-f M-b` for your second point, and `C-spc` for selection`).

felipe-barros:
In my experience I always switch to Holy when browsing Info files. Maybe there could be some sort of warning for the user that in this mode Evil isn't "supported"? Like a red-something in the modeline?

sooheon:
@felipe-barros yeah, just going to holy-mode and doing it the emacs way does seem more "pure" and edge-case free, for things like info-mode or magit. But the vinegar layer here, for example, is a case of evil integration done right, imo.

felipe-barros:
@sooheon although after switching to Spacemacs from Vim I see myself more and more inclined to using regular Emacs keybindings everywhere and Vim only for text editing (using Hybrid for now) I'm not against Evil integration at all. But I agree with you that with Magit/Info, default means better, at least for now. I hope it changes somewhere in the future, but for now, I think the only thing missing is a nice warning "hey, this is unexplored land, be careful and if you need help, press h for emacs intro" or something :)

myrjola:
What about not evilifying Info-mode at all? I would like to have it modal. In-buffer navigation and visual-mode selections is most comfortable in the normal state. Navigating the Info hierarchy can be done effectively with the standard info-mode keybindings. There was probably some clashes between info-mode and the insert-state, so maybe emacs-state should be used instead of insert state.

PR #14188, opened 2020-12-05
In navigation layer, Bind J/K to scroll up/down in Info-mode

dalanicolai:
Spacemacs lacks a keybinding alternative to the most natural way of scrolling
Info pages (i.e. SPC) in vanilla Emacs.
This commit adds J/K to scroll most naturally through info pages.
Currently, in Info-mode, a keybinding for J is not defined while K is bound to
evil-lookup.

Issue #2828 already addresses the inconsistent experience, and in my opinion this
can and should be improved as navigating Info pages is a very crucial part of
using Emacs.

Personally I have bound J/K to scroll page up/down in buffers/pdf/djvu/doc-view,
which I inherited from using the zathura pdf reader, and I think this is a better
default than the default vim alternatives.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -317,6 +317,9 @@ sane way, here is the complete list of changed key bindings
   - Changed =projectile-rails= key binding prefix from ~SPC m r~ to ~SPC m f~
     to avoid conflicts with key bindings in the =web-mode= layer
     (thanks to Adam Sokolnicki)
+***** Spacemacs-navigation
+- Key bindings:
+  - Add ~J/K~ to scroll up/down (or next/previous node) in Info-mode
 ***** Shaders
 - =shaders= layer has been moved to =gpu= layer.
 ***** Syntax checking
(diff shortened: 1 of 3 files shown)


layers/+spacemacs/spacemacs-navigation/README.org
```diff
@@ -21,3 +21,13 @@ This layer adds general navigation functions to all supported layers.
...
 - Shortcuts for easily switching between windows
+
+  
+* Key bindings
+
+** Info-mode
+
+| Key binding | Description                |
+|-------------+----------------------------|
+| ~J~         | Scroll up (or next node)   |
```
smile13241324, line 32: I think this should rather be moved into the documentation instead.

smile13241324 (changes requested):
I like the new bindings for info-mode however I think the bindings must be moved into the general documentation.

layers/+spacemacs/spacemacs-navigation/README.org
dalanicolai: Ah did not realize that. So I have updated the PR.

UNRELATED: wouldn't [this new github feature](https://github.blog/changelog/2020-12-08-github-discussions-now-available-as-a-public-beta/) be handy for Spacemacs development?

smile13241324: Maybe lets discuss it in a separate issue 👍

smile13241324 (approved):

smile13241324:

Issue #14178, opened 2020-11-30
Bad ligature sequence breaks spacemacs init


korayal:
After the latest package updates, I started getting this error:

```
Warning (initialization): An error occurred while loading ‘/home/korayal/.emacs.d/init.el’:

error: Ligature ‘\’ must be 2 characters or longer

To ensure normal operation, you should investigate and remove the
cause of the error in your initialization file.  Start Emacs with
the ‘--debug-init’ option to view a complete error backtrace. Disable showing Disable logging
```

Removing it fixes the issue, but I'm not sure if we should remove it, or replace it with `"\\\\"`. I do not know what that ligature corresponds to.
``` diff
layers/+fonts/unicode-fonts/config.el
@@ -31,7 +31,7 @@ macOS port version.")
                                      "<$" "<=" "<>" "<-" "<<" "<+" "</" "#{" "#[" "#:" "#=" "#!"
                                      "##" "#(" "#?" "#_" "%%" ".=" ".-" ".." ".?" "+>" "++" "?:"
                                      "?=" "?." "??" ";;" "/*" "/=" "/>" "//" "__" "~~" "(*" "*)"
-                                     "\\" "://")
+                                     "://")
   "List of ligatures you would like enabled")
 
 (defvar unicode-fonts-ligature-modes '(t)
```

smile13241324:
This should be fixed today, can you retry and report back?

korayal:
it's fixed! thanks.

Issue #13967, opened 2020-09-20
Ligatures interact badly with header-line-format, hanging emacs


bcc32:
#### Description :octocat:
Enabling unicode-font-ligatures for all modes causes bad interactions
with magit and hangs emacs.

In particular, setting `header-line-format` to some string that would
contain a ligature, in a major-mode selected by
`unicode-fonts-ligature-modes`, appears to immediately cause the emacs
process to hang.

#### Reproduction guide :beetle:
- Enable unicode-fonts layer with `unicode-fonts-enable-ligatures` set to t
- Start Emacs
- Navigate to some git repo
- Press `SPC g s`
- Press `l L` (`magit-log-branches`)

*Observed behaviour:* :eyes: :broken_heart:
Emacs hangs, using 100% CPU, as soon as `magit-set-header-line-format`
is called.

*Expected behaviour:* :heart: :smile:
Magit shows log of all local branches.

#### System Info :computer:
- OS: darwin
- Emacs: 27.1
- Spacemacs: 0.300.0
- Spacemacs branch: unicode-fonts-layer-docs (rev. 691dc6e98)
- Graphic display: t
- Distribution: spacemacs-base
- Editing style: hybrid
- Completion: ivy
- Layers:
```elisp
(spell-checking
 (syntax-checking :variables syntax-checking-use-original-bitmaps t)
 (auto-completion :variables auto-completion-idle-delay 1.0)
 (ivy :variables ivy-enable-advanced-buffer-information t)
 helpful ibuffer
 (org :variables org-projectile-file "TODO.org" org-want-todo-bindings t)
 (unicode-fonts :variables unicode-fonts-enable-ligatures t unicode-fonts-ligature-modes
                '(prog-mode text-mode magit-log-mode))
 autohotkey c-c++ csv nil emacs-lisp nil nil html
 (javascript :variables javascript-fmt-tool 'prettier javascript-fmt-on-save t)
 json markdown ocaml perl5 python ruby rust scheme shell-scripts sql
 (typescript :variables typescript-fmt-tool 'prettier typescript-fmt-on-save t)
 yaml nixos osx git
 (version-control :variables version-control-diff-tool nil)
 spacemacs-completion spacemacs-layouts spacemacs-editing spacemacs-editing-visual spacemacs-evil spacemacs-language spacemacs-misc spacemacs-modeline spacemacs-navigation spacemacs-org spacemacs-project spacemacs-purpose spacemacs-visual command-log fasd finance
[...]

PR #14190, opened 2020-12-05
Update ligature to disable some chars and only enable in prog-mode

AdeAttwood:
When ligatures are enabled globally emacs gets pined at 100% CPU. In #13967
there were some ligatures that were highlighted but enabling this only in
prog-mode so emacs dose not get pined at 100% when just enabling with
(unicode-fonts-enable-ligatures t). In #14178 it was highlighted that "\\" had
an issue in the `init.el` so this has been removed

- "//" Issues with doom mode line
- ".." "..." Issues with helm find file
- "--" Issues with magit
- "\\" Issue with init.el


lebensterben:
The upstream package support per-mode ligature sets.
I thinks it's better to have `unicode-fonts-ligature-modes-alist` instead.

So the default value is then `((prog-mode . unicode-fonts-ligature-set))`

AdeAttwood:
If you have `unicode-fonts-ligature-modes-alist` would you not have to have a map of all modes to ligatures

lebensterben:
> If you have `unicode-fonts-ligature-modes-alist` would you not have to have a map of all modes to ligatures

The upstream recommends to set these

https://github.com/mickeynp/ligature.el/blob/c830b9d74dcf4ff08e6f19cc631d924ce47e2600/ligature.el#L71-L83

which includes a generic setting for `prog-mode`.

Specific major modes, if have its own ligature settings, should always take priority over the generic one. I haven't test this because ligature breaks `lsp-ui`. But this should to be an expected behavior

> Entering a programming language mode runs the custom Lisp functions specified in the hook variable prog-mode-hook, followed by those specified in the mode’s own mode hook (see Major Modes)

AdeAttwood:
I am leaning towards getting this merged to fix the immediate issues of #14178 so emacs will load.

Then we can address the issue of setting different ligature sets per mode in a separate feature. Currently, the `unicode-fonts` layer dose not support this. I will be happy to work on it but may need some guidance on the lisp. I had to look up what the `alist` was and the dot in `((prog-mode . unicode-fonts-ligature-set))` did.

My main questions around the "setting different ligature sets per modes" is what will the config look like. I think we should keep the single toggle of `unicode-fonts-enable-ligatures` but this can be discussed in the issues if it gets split out.

lebensterben:
@AdeAttwood Agree

smile13241324 (approved):
Fast workaround, lets merge it and do the details later.

smile13241324:
I have merged the workaround and adjusted the documentation to state the new default setting. 

Lets have a separate issue for the ligature alist discussion. This could be done in a similar fashion how company settings are done, this is the individual layers define a `pre-init-ligature-mode` function where the alist is extended with mode specific settings. Same thing is done for the `purpose-config`.

PR #14193, opened 2020-12-07
Fix fsharp-mode auto-completion unreliable

niontrix:
First off I want to thank @smile13241324 for taking the time to make lsp
working with the fsharp layer!

Unfortunately the eglot configuration for automatically starting a server instance when
opening a file is not quite right. It tries to start the server twice which leads
to more problems down the line. Also the company-backend is not properly setup
so auto-completion does not work reliably when using lsp or eglot.

I also found an upstream issue in fsharp-mode which causes problems with
company-mode. I already created a [pull request](https://github.com/fsharp/emacs-fsharp-mode/pull/255) to fix that and hope that it will be
merged soon.

### DIFF
--- a/layers/+lang/fsharp/funcs.el
+++ b/layers/+lang/fsharp/funcs.el
@@ -19,20 +19,16 @@
 
 (defun spacemacs//fsharp-setup-company ()
   "Conditionally setup company based on backend."
-  (pcase (spacemacs//fsharp-backend)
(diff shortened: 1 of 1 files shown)


smile13241324 (approved):
Thanks for double checking, I will merge your fix soon.

niontrix:
I'm glad I could help :)

PR #14197, opened 2020-12-09
[markdown] fix: mmm clobbering comment-start in Markdown Git commit

### LINKED ISSUE #14195: Syntax highlighting for comments in Magit commit buffers with custom core.commentChar
#### Description :octocat:

I set `git-commit-major-mode` to be `markdown-mode`, and to keep the `#` comment
lines that Git inserts from getting formatted like markdown headers, I switched
away from `#` as Git's comment delimiter. In my `.gitconfig`:

```ini
[core]
	commentChar = !
```

The commit buffer gets highlighted as Markdown, but the Git commit syntax
highlighting is lost. Thinking this may not already be supported or may require
special configuration, I asked a
[question](https://emacs.stackexchange.com/q/60915/19069) about it on the Emacs
StackExchange, and Jonas Bernoulli, maintainer of Magit, suggested:

> Try setting the comment-start variable using git-commit-setup-hook.

However, that did not fix the problem; when I did that, it made no difference.
Also, when I set commit buffers to have no major mode, it properly highlights
the comments that use `!`.

Jonas then recommended:

> Try to reproduce with `emacs -Q`, see [magit.vc/manual/magit/Debugging-Tools.html](https://magit.vc/manual/magit/Debugging-Tools.html)

However, I don't really know how to use Emacs from scratch; to date I have been
Spacemacs-only so I get a little lost in bare Emacs mode. After I said so, he
[responded](https://emacs.stackexchange.com/a/62123/19069):

> This is supposed to already work. Figuring out why it doesn't work for you
> will require a debugging effort on your part and I am afraid I won't be able
> to help you any more with that because I do not know enough about Spacemacs to
> do so. I suggest you ask for help from some other Spacemacs users.

I had a coworker who has built his own Emacs configuration test it out, and he
got the same result as Jonas: it works. So it looks like this is a Spacemacs
bug.

- Set `core.commentChar` to `!` (or any other symbol that is not part of Markdown's syntax)
- Start Emacs
- `(setq git-commit-major-mode 'markdown-mode)`
- Enter a Git repo in Emacs
- Make a change
- Start to commit the change with Magit

Git's commit lines at the bottom of the buffer are kept from being formatted
like headers by the switch to `!`, but these lines are no longer
syntax-highlighted according to the syntax of a Git commit; they show as they
would in a plain Markdown file, as plain text.

To see markdown mode meshed with the an understanding of Git commits, so that
_above_ the comment lines would be highlighted with Markdown, and the comment
lines themselves would be highlighted.

#### System Info :computer:
- OS: darwin
- Emacs: 28.0.50 (gccemacs)
- Spacemacs: 0.300.0
- Spacemacs branch: keith (rev. 86cbb2777) (*Note*: this branch is barely divergent from the standard Spacemacs `develop`, I just use it so I can be dogfooding my own PRs before they hit upstream; currently I have none outstanding and the differences are minute)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(spell-checking syntax-checking
                (auto-completion :variables auto-completion-enable-snippets-in-popup t)
                helm
                (org :variables org-enable-verb-support t)
                (unicode-fonts :variables unicode-fonts-ligature-modes
                               '(prog-mode))
                emoji games agda c-c++ clojure emacs-lisp fsharp graphviz
                html idris java javascript jsonnet kotlin markdown ocaml
                (scala :variables scala-backend 'scala-metals scala-enable-eldoc t scala-auto-start-backend t)
                scheme sql typescript yaml
                (multiple-cursors :variables multiple-cursors-backend 'evil-mc)
                osx git github version-control
                (colors :variables colors-colorize-identifiers 'variables)
```
- System configuration features: JPEG TIFF PNG NOTIFY KQUEUE ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS MODULES NATIVE_COMP THREADS XWIDGETS JSON PDUMPER LCMS2

#### Backtrace :paw_prints:

N/A

Kazark:
Fixes #14195; see there for details.

I'm open to a higher-fidelity way to check that it is performing a Git commit.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2592,6 +2592,8 @@ files (thanks to Daniel Nicolai)
 - Improvements:
   - Evilfied =dap= debug windows
 **** Markdown
+- Disable =mmm-mode= in Git commit buffers, in order to allow the use of Markdown
+  mode when committing (thanks to Keith Pinson)
 - New layer variable =markdown-mmm-auto-modes= which is a list of language names
   or lists of language and mode names that are supported in source blocks, you
   can add you own modes, see layer README.org for more info
--- a/layers/+lang/markdown/funcs.el
+++ b/layers/+lang/markdown/funcs.el
@@ -10,8 +10,9 @@
 ;;; License: GPLv3
 
 (defun spacemacs/activate-mmm-mode ()
+  (unless git-commit-mode
   ;; Enable `mmm-mode'.
-  (mmm-mode 1))
+    (mmm-mode 1)))
 
 ;; stolen from http://stackoverflow.com/a/26297700
 ;; makes markdown tables saner via orgtbl-mode


lebensterben:
I will take a look later today

lebensterben:
@Kazark You probably can check the value of `git-commit-mode` instead

Kazark:
@lebensterben Thanks. I didn't know there were mode variables in addition to mode functions. That's exactly what I needed. Updated.

smile13241324:
That's a strange bug 🤔 . Anyway thanks for fixing @Kazark for now I also see no more elegant solution to solve this. I have tested this a bit to make sure that `mmm-mode` is unaffected for other buffers.

In future we could try to use `mmm-never-modes` to reach the same.

smile13241324:

PR #14199, opened 2020-12-10
[ivy] improve key bindings

thanhvg:
bind `C-SPC` to  `ivy-call-and-recenter`
this is the closest thing to helm follow mode in ivy world

unbind `g` key and map `U` for
`for ivy-occur-mode-map` and `ivy-occur-grep-mode-map`

### DIFF
--- a/layers/+completion/ivy/packages.el
+++ b/layers/+completion/ivy/packages.el
@@ -228,6 +228,7 @@
       (define-key ivy-minibuffer-map [escape] 'minibuffer-keyboard-quit)
       (define-key ivy-minibuffer-map (kbd "M-SPC") 'hydra-ivy/body)
       (define-key ivy-minibuffer-map (kbd "C-<return>") #'ivy-alt-done)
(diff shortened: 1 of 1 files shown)


lebensterben:
ibus and fcitx by default use C+Space to toggle input method. Though many people use Emacs's built-in input method, I still imagine this new binding would be inconvenient to some users.

thanhvg:
`C-spc` is standard key binding for mark set in vanilla emacs.  Default ibus key bindings for my gnome desktop is `winkey-spc`. Helm allows you to select multiple candidates with `C-spc` and act on them. Afaik ivy can't do that, so `C-spc` is safe to use for this case. Also I stole this binding from Doom.

lebensterben:
>`C-spc` is standard key binding for mark set in vanilla emacs.  Default ibus key bindings for my gnome desktop is `winkey-spc`. Helm allows you to select multiple candidates with `C-spc` and act on them. Afaik ivy can't do that, so `C-spc` is safe to use for this case. Also I stole this binding from Doom. 

@thanhvg thx

smile13241324 (approved):
Looks good, thanks for contributing 💜

PR #14200, opened 2020-12-10
[spacemacs-evil] add evil-collection

thanhvg:
It is impossible to ignore `evil-collection` anymore. `evil-magit` has been
deprecated recently and moved to `evil-collection`.

It will save time and effort for both Spacemacs and Evil to share and contribute
to `evil-collection` imo.

However I strongly prefer Spacemacs binding scheme over evil-collection's one.
We should only pick what we need from `evil-collection`.

This PR add mechanism to embrace `evil-collection` and apply it to shell layer (`vterm`)
and git layer.

Note: I still have a bug with git rebase binding. `C-j` and `C-k` are note bound to move commits. Must use `C-n` and C-p` instead. Not sure if it is our bug or `evil-collection's

### DIFF
--- a/layers/+distributions/spacemacs-bootstrap/packages.el
+++ b/layers/+distributions/spacemacs-bootstrap/packages.el
@@ -64,6 +64,8 @@
   ;; evil-mode is mandatory for Spacemacs to work properly
   ;; evil must be require explicitly, the autoload seems to not
   ;; work properly sometimes.
+  ;; `evil-collection' wants this value
+  (setq evil-want-keybinding nil)
   (require 'evil)
   (evil-mode 1)
 
--- a/layers/+source-control/git/packages.el
+++ b/layers/+source-control/git/packages.el
@@ -11,7 +11,7 @@
 
 (setq git-packages
       '(
-        evil-magit
+        evil-collection
         fill-column-indicator
         ;; forge requires a C compiler on Windows so we disable
         ;; it by default on Windows.
@@ -43,26 +43,8 @@
     :post-config
     (add-to-list 'golden-ratio-exclude-buffer-names " *transient*")))
 
-(defun git/pre-init-evil-magit ()
-  (spacemacs|use-package-add-hook magit
-    :post-config
-    (when (spacemacs//support-evilified-buffer-p dotspacemacs-editing-style)
-      (evil-magit-init))
-    (evil-define-key 'motion magit-mode-map
-      (kbd dotspacemacs-leader-key) spacemacs-default-map)
-    ;; Remove inherited bindings from evil-mc and evil-easymotion
-    ;; do this after the config to make sure the keymap is available
-    (which-key-add-keymap-based-replacements magit-mode-map
-      "<normal-state> g r" nil
-      "<visual-state> g r" nil
-      "<normal-state> g s" nil
-      "<visual-state> g s" nil)))
-
-(defun git/init-evil-magit ()
-  (use-package evil-magit
-    :defer t
-    :init (add-hook 'spacemacs-editing-style-hook
-                    'spacemacs//magit-evil-magit-bindings)))
+(defun git/pre-init-evil-collection ()
+  (add-to-list 'spacemacs-evil-collection-allowed-list 'magit))
 
 (defun git/post-init-fill-column-indicator ()
   (add-hook 'git-commit-mode-hook 'fci-mode))
--- a/layers/+spacemacs/spacemacs-evil/README.org
+++ b/layers/+spacemacs/spacemacs-evil/README.org
@@ -5,6 +5,8 @@
 * Table of Contents                     :TOC_5_gh:noexport:
 - [[#description][Description]]
   - [[#features][Features:]]
+- [[#install][Install]]
+- [[#note-about-evil-collection][Note about evil-collection]]
 
 * Description
 This layer adds various adjustments to packages to create an evilified experience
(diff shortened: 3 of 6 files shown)


duianto:
I'm unable to reproduce the git rebase issue.
- `r` rebase
- `i` interactively
- move the cursor down a couple of commits
- `C-c C-c` rebase on the commit

Now both with and without this PR applied:
```
 C-j                    ;; electric-newline-and-maybe-indent
 C-k                    ;; git-rebase-kill-line
 C-n                    ;; evil-paste-pop-next
 C-p                    ;; evil-paste-pop
```

The `# Commands:` list in the `git-rebase-todo` buffer shows:
```
# M-k      move the commit at point up
# M-j      move the commit at point down
```
and they work as expected.

<details><summary>
Windows 10, System Info (Click to expand)
</summary>
<pre>
#### System Info :computer:
- OS: windows-nt
- Emacs: 27.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. aa87b010a)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(autohotkey emacs-lisp git helm multiple-cursors
                 '("c:/Users/username/org/")
                 org-enable-jira-support t org-enable-roam-support t)
            shell spell-checking syntax-checking treemacs version-control)
```

</pre>
</details>

---

The removal of the which-key prefixes for `evil-mc` and `evil-easymotion` should be added back, because when they are removed, then the which-key popup for the prefix: `g`
shows:
- `r -> evil-mc` but it calls the expected `magit-refresh`
- `s -> evil-easymotion` but it calls the expected `magit-jump-to-staged`

and we probably should remove the now unused function:
`spacemacs//magit-evil-magit-bindings`

```diff
2 files changed, 8 insertions(+), 10 deletions(-)
layers/+source-control/git/funcs.el    | 9 ---------
layers/+source-control/git/packages.el | 9 ++++++++-

modified   layers/+source-control/git/funcs.el
@@ -68,15 +68,6 @@
       (and (eq style 'hybrid)
            hybrid-style-enable-evilified-state)))
 
-(defun spacemacs//magit-evil-magit-bindings (style)
-  "Set `evil-magit' bindings for the given editing STYLE."
-  (cond
-   ((spacemacs//support-evilified-buffer-p style)
-    (evil-magit-init))
-   (t
-    (when (featurep 'evil-magit)
-      (evil-magit-revert)))))
-
 
 ;; git blame transient state
 
modified   layers/+source-control/git/packages.el
@@ -253,7 +253,14 @@
       (evil-define-key 'normal magit-section-mode-map (kbd "M-6") 'spacemacs/winum-select-window-6)
       (evil-define-key 'normal magit-section-mode-map (kbd "M-7") 'spacemacs/winum-select-window-7)
       (evil-define-key 'normal magit-section-mode-map (kbd "M-8") 'spacemacs/winum-select-window-8)
-      (evil-define-key 'normal magit-section-mode-map (kbd "M-9") 'spacemacs/winum-select-window-9))))
+      (evil-define-key 'normal magit-section-mode-map (kbd "M-9") 'spacemacs/winum-select-window-9)
+      ;; Remove inherited bindings from evil-mc and evil-easymotion
+      ;; do this after the config to make sure the keymap is available
+      (which-key-add-keymap-based-replacements magit-mode-map
+        "<normal-state> g r" nil
+        "<visual-state> g r" nil
+        "<normal-state> g s" nil
+        "<visual-state> g s" nil))))
 
 (defun git/init-magit-delta ()
   (use-package magit-delta

```

duianto:
`evil-magit` is also:
- Mentioned in the git readme:
https://github.com/syl20bnr/spacemacs/blame/develop/layers/%2Bsource-control/git/README.org#L163
>Spacemacs uses [[https://github.com/justbur/evil-magit][evil-magit]] for key bindings in magit buffers (unless your editing
style is set to emacs, in which case you get the default magit bindings), which
are the standard magit key bindings with some minimal changes to make them
comfortable for evil users.

Maybe it's enough to change the url and description to evil-collection.
- Used in the `keyboard-layout` layer:
https://github.com/syl20bnr/spacemacs/blob/7ffcae0360af1529346f19bcfe4da4b0a06bc450/layers/%2Bintl/keyboard-layout/packages.el#L25
https://github.com/syl20bnr/spacemacs/blob/7ffcae0360af1529346f19bcfe4da4b0a06bc450/layers/%2Bintl/keyboard-layout/packages.el#L281-L297

robbyoconnor:
I pulled this in

mpe:
I had some custom `evil-define-key`s, they needed updating from `evil-magit-state` to `evil-collection-magit-state`.

emacs18:
Aside from `evil-magit`, it appears that `evil-collection` also includes the following packages currently used by spacemacs:

- evil-ediff
- evil-indent
- evil-unimpaired

Perhaps these packages should also be replaced with `evil-collection`.

robbyoconnor:
evil-unimpaired is amazing

lebensterben:
Please review and accept this PR since `evil-magit` is no longer available on Melpa.

smile13241324:
I am afraid I will have no time before the second of February, due to some urgent schedules in my primary job. Maybe @duianto @JAremko or @sdwolfz can chip in if this is urgent.

sdwolfz:
I say merge it, if @JAremko and @duianto are :+1: with this too I can press the big green button :smile:

eklitzke:
I tested this branch locally and it works for me. However, I do have the `which-key` issue @duianto reported in https://github.com/syl20bnr/spacemacs/pull/14200#issuecomment-766440865 where I'm seeing the wrong `which-key` information (and therefore the wrong information in the helm minibuffer) when using `g r` and `g s` in magit. This is a minor issue though, it can be fixed in another PR if it can't be updated here.

sdwolfz:
@eklitzke yea, sounds like something that can be fixed later.

Atrus7:
By the way, this regressed my dired-mode behavior. Suddenly my RET key tried to perform evil-ret instead of dired-find-file. After debugging a bit I rolled back just before the patch, and it's working again.

emacs18:
It is not just RET key, but dired-mode key binding is completely different!
This is a deal breaker for me.  I cannot find a variable that I can tweak or a function I can call to go back to the old key bindings.
For now I'm just going to revert this change in my own repository.
The fact that MELPA does not have evil-magit does not impact me, since I use straight.el package manager which still supports evil-magit just fine.

duianto:
The following PR has been merged:
[evil-collection] support for dired #14333

Now `RET` calls `(dired-find-file)`.

duianto:
Here's the list of `evil-collection-dired` keys:
https://github.com/emacs-evil/evil-collection/blob/a36f742c215d233ab383e91933b916a54c9c4e56/modes/dired/evil-collection-dired.el#L41

Spacemacs used to have a mix of evil and default dired bindings.

It's understandable that people have gotten used to them, 
but it was unexpected when for example:
`gg` didn't jump to the top, the first `g` refreshed the buffer.

Now `gr` refreshes the buffer, the same key binding as in magit.

jojojames:
In addition to dired, you may wish to look at https://github.com/emacs-evil/evil/blob/master/evil-keybindings.el to see a list of modes that were probably bound for spacemacs before and aren't anymore.

Issue #14192, opened 2020-12-06
latexmk not working properly in LaTeX mode


wang1zhen:
I have enabled latex layer in my config file and specified 'LatexMk' as the build command:
```
(latex :variables
            latex-backend 'company-auctex
            latex-build-command "LatexMk"
            latex-enable-auto-fill t
            latex-enable-folding t
            latex-enable-magic nil
            ;; magic-latex-enable-block-highlight t
            ;; magic-latex-enable-block-align t
            ;; magic-latex-enable-pretty-symbols nil
            ;; magic-latex-enable-suscript nil
            ;; magic-latex-enable-inline-image nil
            )
```
And here goes my test latex file (with Chinese characters, so I wished to build this file with xelatex)

```latex
\documentclass[a4paper]{article}
\usepackage{ctex}
\usepackage{amsmath}
\usepackage{lipsum}

\author{test}

\begin{document}

\tableofcontents

\section{formulas}

\subsection{Newton's Law}

\begin{equation}
  F=ma
  \label{eq1}
\end{equation}

As mentioned above, equation \ref{eq1} is Newton's Law.

\section{中文}

式\ref{eq1}是牛顿第二定律。

\subsection{lorem-ipsum}

\lipsum[1]

\end{document}
```

I have created a .latexmkrc file to use xelatex instead of pdflatex while building the doc:

```
$pdflatex = 'pdflatex -file-line-error -hold-on-error -interaction=nonstopmode -synctex=1 %O %S';
$xelatex = 'xelatex -file-line-error -hold-on-error -interaction=nonstopmode -no-pdf -synctex=1 %O %S';
$xdvipdfmx = "xdvipdfmx -E -o %D %O %S";
$pdf_mode = 5;
$out_dir = './LaTeX.out';
```

If I run `latexmk main.tex' in command line, everything works fine, and this is the beginning of the log file:
```
This is XeTeX, Version 3.14159265-2.6-0.999991 (TeX Live 2019/Debian) (preloaded format=xelatex 2020.12.4)  6 DEC 2020 16:37
entering extended mode
 restricted \write18 enabled.
 file:line:error style messages enabled.
 %&-line parsing enabled.
**main.tex
(./main.tex
LaTeX2e <2020-02-02> patch level 2
L3 programming layer <2020-02-14>
(/usr/share/texlive/texmf-dist/tex/latex/base/article.cls
```

[...]

PR #14204, opened 2020-12-11
LaTeX: improved i18n support and documentation

lebensterben:
- Added a new layer config variable `latex-build-engine`, a thin wrapper of
  `TeX-engine` (This addresses #14192) :
  - For better support of non-european character set rendering, `xetex` is
    chosen when it's found on PATH, and when eithe `chinese` or `japanese`
    layer is enabled
- Added a new layer config variable `latex-enable-preview`:
  - When non-nil, `auto-revert-mode` is added to `doc-view-mode-hook`
  - Previously this is described in documentation but requires users to set it
    manually
- Documentation:
  - Added a section that lists are layer config variables
  - Improved all documentation examples, s.t. they are valid elisp forms to be
    evaluated directly
  - Improved the section on backend:
    - Emphsized that LSP layer is always preferred
    - Provided information on installing the language server
  - Added the default value for layer config variable, `latex-nofil-env`


smile13241324 (approved):
Good job! I'll be merging this right now. 👍

PR #14206, opened 2020-12-13
dumb-jump: Replace obsolete interface with xref-backend

lebensterben:
- Add `dumb-jump-xref-activate` hook to `xref-backend-functions`, which
  add a fallback option, i.e. `dumb-jump`, when no better alternatives
  is available
- Removed the obsolete interface `dumb-jump-go` from the following layers:
  - clojure
  - latex
  - vue
- The global binding `SPC j q` requires the obsolete function
  `dumb-jump-quick-look` and is therefore removed

### DIFF
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -2228,7 +2228,6 @@ The ~SPC j~ prefix is for jumping, joining and splitting.
 | ~SPC j j~   | jump to one or more characters in the buffer (works as an evil motion)            |
 | ~SPC j k~   | jump to next line and indent it using auto-indent rules                           |
 | ~SPC j l~   | jump to a line with avy (works as an evil motion)                                 |
-| ~SPC j q~   | show the dumb-jump quick look tooltip                                             |
 | ~SPC j u~   | jump to a URL in the current buffer                                               |
 | ~SPC j U~   | select a URL in the current buffer and follow it                                  |
 | ~SPC j v~   | jump to the definition/declaration of an Emacs Lisp variable                      |
--- a/layers/+frameworks/vue/funcs.el
+++ b/layers/+frameworks/vue/funcs.el
@@ -38,7 +38,7 @@
 
 ;; dumb
 
-(defun spacemacs//vue-setup-dumb-imenu ()
+(defun spacemacs//vue-setup-dumb ()
   (setq imenu-generic-expression '(("html" "^<template>$" 0)
                                    ("js" "^<script>$" 0)
                                    ("js" "^\\s-*\\(data\\).*()\\s-?{" 1)
@@ -52,10 +52,6 @@
                                    ("js" "^\\s-*\\(props\\):\\s-?{" 1)
                                    ("css" "^<css>$" 0))
         imenu-create-index-function #'imenu-default-create-index-function))
-
-(defun spacemacs//vue-setup-dumb ()
-  (add-to-list 'spacemacs-jump-handlers-vue-mode 'dumb-jump-go)
-  (spacemacs//vue-setup-dumb-imenu))
 
 (defun spacemacs//vue-setup-dumb-company ()
   (spacemacs|add-company-backends :backends (company-web-html company-css company-files company-dabbrev)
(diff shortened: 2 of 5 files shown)


lebensterben:
References
==========

The old interface is obsolete now:

https://github.com/jacktasia/dumb-jump/blob/ff9fc9360d39f5e07c1f480f8b0656b49606781b/dumb-jump.el#L2992-L3012

To use the new xref backend:

https://github.com/jacktasia/dumb-jump/blob/ff9fc9360d39f5e07c1f480f8b0656b49606781b/dumb-jump.el#L34-L36

thanhvg:
- What happens when other smart xref handlers such as `lsp`, `tide`  etc come in? Will this xref binding let those handlers take precedence over dumb jump?

-  Warning about obsolete `dumb-jump-quick-look` can be suppressed with recent update I think so we should keep `spc j q`

lebensterben:
>- What happens when other smart xref handlers such as `lsp`, `tide`  etc come in? Will this xref binding let those handlers take precedence over dumb jump? ...

@thanhvg

`(add-hook 'xref-backend-functions #'dumb-jump-xref-activate)`

Add the jump handler to the end of the backend-function list. So it behaves as the fallback option when no other better alternative is available.

All warnings can be suppressed actually. But dumb jump is not very often used in most prog-mode, I accidentally come across it when working on another  commit. My main motivation for pushing this radical change is that people are not likely to know this interface is obsolete. And when it's finally removed, it will break workflows for certain people.

tshu-w:
> Add the jump handler to the end of the backend-function list. So it behaves as the fallback option when no other better alternative is available.

@lebensterben  In fact, it seems that lsp turns xref-backend-functions into a buffer-local variable, making it impossible to use dumb-jump via xref. For this case, we need something like `(add-hook 'xref-backend-functions #'dumb-jump-xref-activate 5 t)`

Screen Shot 2020-12-15 at 9 47 44 AM

lebensterben:
>> Add the jump handler to the end of the backend-function list. So it behaves as the fallback option when no other better alternative is available. ...

@tshu-w xref-backend-functions is always a buffer local variable.

I will do some testing.
As far as I know, only Clojure, LaTeX, and Vue, uses dumb-jump. It's also possible to add dumb-jump after the LSP layer is loaded.

tshu-w:
> xref-backend-functions is always a buffer local variable.

yes, you are right. But what I'm trying to say is that the current setup doesn't allow it as a fallback option.

smile13241324:
I think removing the individual bindings in the vue, clojure and latex layer are pretty much ok, especially as `dump-jump-go` was a default jump handler in the first place so it should have been available for each mode already.

For the binding of `SPC j q`, is there a xref specific alternative for this binding? If not I think we should remove it. Otherwise we risk again that this feature breaks later on.

We should also do the same with the default jump handler binding, which at the current state has simply be removed, this is lets have the default xref function for looking up symbols be bound, then with dump-jump properly configured this would restore `dump-jump` fallback bindings for `SPC gg` and `SPC gG`.

About the `xref-backend-functions` var binding, when the binding is done globally this should be triggered automatically if no local symbol satisfies the function requirements, this is due to `add-hook` adding `t` as last position to the list which causes the global bindings to be executed.

``` elisp
;;; Excerpt from add-hook help:
The optional fourth argument, LOCAL, if non-nil, says to modify
the hook's buffer-local value rather than its global value.
This makes the hook buffer-local, and it makes t a member of the
buffer-local value.  That acts as a flag to run the hook
functions of the global value as well as in the local value.
```

lebensterben:
@smile13241324 
Anything I need to do to push forward this PR?

smile13241324:
Sorry for my late reply @lebensterben but I am currently going through some hard times at work and will most likely not be able to work on Spacemacs this week. 

Maybe @duianto, @JAremko, @sdwolfz or @syl20bnr can jump in for me.

duianto:
I don't know much about dumb-jump but since smile didn't seem to have any requests for changes, and you've got high quality PRs so let's merge this.

duianto:

PR #14211, opened 2020-12-14
Make valign-mode opt-out in org-mode-hook

Claude-Ray:
valign-mode is nice, but it could affect the performance when scaling in a big org buffer.

Add an opt-out for people who want to toggle valign-mode manually.


lebensterben:
Naming it as `org-enable-valign` seems better to me.

Claude-Ray:
Good idea, it's done.

smile13241324 (approved):
Good point, lets keep people a chance to opt-out.

Issue #13913, opened 2020-09-04
Mu4e Org support is broken


enderger:
#### Description :octocat:
The Org support for mu4e appears to be broken

#### Reproduction guide :beetle:
- Start Emacs with the mu4e and org layers installed
- Launch mu4e
- Compose an email
You should notice the lack of support for Org syntax despite 

*Observed behaviour:* :eyes: :broken_heart:
No org support is found in mu4e messages

*Expected behaviour:* :heart: :smile:
Org syntax can be used in mu4e

#### System Info :computer:
- OS: gnu/linux (Artix)
- Emacs: 27.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 3ba5234e8)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(haskell rust html helm auto-completion emacs-lisp racket clojure git java python
         (org :variables org-enable-trello-support t org-latex-preview-ltxpng-directory \.ltximg)
         (shell :variables shell-default-height 30 shell-default-position 'bottom)
         (spell-checking :variables spell-checking-enable-auto-dictionary t =enable-flyspell-auto-completion= t)
         syntax-checking perl6 gnus
         (mu4e :variables mu4e-installation-path "/usr/share/emacs/site-lisp" mu4e-use-maildirs-extension t mu4e-enable-mode-line t)
         version-control)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG CAIRO SOUND GPM DBUS GSETTINGS GLIB NOTIFY INOTIFY ACL GNUTLS LIBXML2 FREETYPE HARFBUZZ M17N_FLT LIBOTF ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XDBE XIM MODULES THREADS LIBSYSTEMD JSON PDUMPER LCMS2 GMP

smile13241324:
Not sure what the issue is, the layer correctly requires the org related mu4e packages. I expect this to be a load order issue or an issue between mu4e org package and latest version of org.

Would be great if someone using mu4e can tackle this. I am using plain old thunderbird.

codetrick:
There are two separate org - mu4e integration features distributed with mu4e, one `org-mu4e.el`, the other `mu4e-org.el`. 

[...]

PR #14212, opened 2020-12-14
Improving org-mu4e integration

codetrick:
This pull request addresses issue #13913.

Changes:
1. Added two configuration variables for the mu4e layer to control mu4e-org integration.
2. Optionally configures mu4e compose mode to support Org-mode syntax.
3. Improves lazy-loading of relevant features. E.g. loading org-mode does not automatically trigger loading of mu4e.

### DIFF
--- a/layers/+email/mu4e/packages.el
+++ b/layers/+email/mu4e/packages.el
@@ -50,7 +50,8 @@
       (setq mu4e-completing-read-function 'completing-read
             mu4e-use-fancy-chars 't
             mu4e-view-show-images 't
-            message-kill-buffer-on-exit 't)
+            message-kill-buffer-on-exit 't
+            mu4e-org-support nil)
       (let ((dir "~/Downloads"))
         (when (file-directory-p dir)
           (setq mu4e-attachment-dir dir))))
@@ -150,10 +151,20 @@ mu4e-use-maildirs-extension-load to be evaluated after mu4e has been loaded."
     :init (with-eval-after-load 'mu4e (mu4e-maildirs-extension-load))))
 
 (defun mu4e/pre-init-org ()
-  ;; load mu4e-org when org is actually loaded
-  (with-eval-after-load 'org
-    (require 'mu4e nil 'noerror)
-    (require 'mu4e-org nil 'noerror)))
+  (if mu4e-org-link-support
+      (with-eval-after-load 'org
+        (require 'mu4e-org)
+        ;; We require mu4e due to an existing bug https://github.com/djcb/mu/issues/1829
+        ;; Note that this bug prevents lazy-loading.
+        (require 'mu4e-meta)
+        (if (version<= mu4e-mu-version "1.4.13")
+            (require 'mu4e))))
+  (if mu4e-org-compose-support
+      (progn
+        (spacemacs/set-leader-keys-for-major-mode 'mu4e-compose-mode
+          "o" 'org-mu4e-compose-org-mode)
+        (autoload 'org-mu4e-compose-org-mode "org-mu4e")
+        )))
 
 (defun mu4e/post-init-window-purpose ()
   (let ((modes))
--- a/layers/+email/mu4e/README.org
+++ b/layers/+email/mu4e/README.org
@@ -10,6 +10,7 @@
   - [[#global-bindings][Global bindings]]
   - [[#headers-mode][Headers mode]]
   - [[#view-mode][View mode]]
+  - [[#compose-mode][Compose mode]]
 - [[#configuration][Configuration]]
   - [[#maildirs-extension][Maildirs extension]]
   - [[#multiple-accounts][Multiple Accounts]]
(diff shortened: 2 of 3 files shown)


codetrick:
I have a warning "package org not initialized in layer mu4e". The dependency on org-mode here is optional. Is it better to keep it this way, or remove `org` from the list of packages, or move the `(use-package mu4e-org ...` to `mu4e/pre-init-org`? The two features loaded here belong to the `mu4e` package, not `org` though...

lebensterben:
spacemacs has function 
`configuration-layer/package-usedp`

codetrick:
Ok did as suggested.

However I found there's a bug related to the lazy-loading of `mu4e-org` (point 3 in the first comment above) from the upstream (https://github.com/djcb/mu/issues/1829), which is fixed in their master branch, but not released yet. This means in order to not break `org-store-link`, `mu4e` still needs to be loaded when `org` is loaded, if `mu4e-org-link-support` is set. I put in some comments so it will be clear why mu4e is required, and hopefully someone will see it and remove it whenever the bugfix is released in the future.

smile13241324:
@codetrick seems to be fixed upstream, care to have a look and revise your PR if necessary?

codetrick:
> @codetrick seems to be fixed upstream, care to have a look and revise your PR if necessary?

Is it okay to remove the workaround if the upstream fix is not yet released though? Note that `mu4e` is not distributed through elpa or melpa, but usually bundled with `mu` through Linux distribution package managers, so the rollout will be slower. Removing the workaround will **break** the org-mu4e integration.

lebensterben:
>> @codetrick seems to be fixed upstream, care to have a look and revise your PR if necessary? ...

@codetrick Then you can add a test on Emacs version, so it only applies to old (including the  current) releases.

codetrick:
> > > @codetrick seems to be fixed upstream, care to have a look and revise your PR if necessary? ...
> 
> @codetrick Then you can add a test on Emacs version, so it only applies to old (including the current) releases.

Thanks for the advice. I think this should work now.

layers/+email/mu4e/packages.el
```diff
@@ -50,7 +49,8 @@
       (setq mu4e-completing-read-function 'completing-read
             mu4e-use-fancy-chars 't
             mu4e-view-show-images 't
-            message-kill-buffer-on-exit 't)
+            message-kill-buffer-on-exit 't
+            mu4e-org-support nil)
```
smile13241324, line 53: Can you tell why you are disabling `mu4e-org-support` by default here?

layers/+email/mu4e/packages.el
```diff
@@ -122,7 +122,27 @@
                 (push (buffer-name buffer) buffers))))
           (nreverse buffers)))
       (setq gnus-dired-mail-mode 'mu4e-user-agent)
-      (add-hook 'dired-mode-hook 'turn-on-gnus-dired-mode))))
+      (add-hook 'dired-mode-hook 'turn-on-gnus-dired-mode)))
+
+  (if (configuration-layer/package-usedp 'org)
```
smile13241324, line 127: This must be refactored, you cannot simply inline multiple use-package calls. Instead these must be declared on the top-level with their respective condition in their `:if` construct or refactored to do their respective require steps in the `pre-init-org`.

Your test against the mu4e version is ok and should not require additional use-package declarations.

layers/+email/mu4e/packages.el
```diff
@@ -15,7 +15,6 @@
     mu4e-alert
     mu4e-maildirs-extension
     (helm-mu :requires helm)
-    org
```
smile13241324, line 18: You will require this otherwise you will not be able to modify org parameters.

smile13241324 (changes requested):
I think you need to refactor the implementation for loading the org extensions before this can be merged. Please check my individual comments.

layers/+email/mu4e/packages.el
codetrick: Setting this variable causes `(require 'mu4e-org)` (see [src](https://github.com/djcb/mu/blob/master/mu4e/mu4e.el#L40)) which causes `(require 'org)`.

One may not want to load org right when mu4e is loaded. In fact the only reason I can think of for triggering loading org from mu4e is when you want to write in org-mode syntax in the message buffer, which is handled by the `org-mu4e` feature, loaded explicitly via `org-mu4e-compose-org-mode`.

codetrick: On the other hand, because `mu4e-org` merely enables capturing mu4e messages and views as org links, it could simply be loaded with org (`:after org`).

layers/+email/mu4e/packages.el
codetrick: I used `use-package` instead of `(autoload 'org-mu4e-compose-org-mode ...)` or `(with-eval-after-load 'org ...)` because I thought it is a nicer wrapper that achieves the same functionality. 

I am pretty new to both emacs and spacemacs. Could you kindly tell me if there's something extra that `use-package` does which is undesired here, or this is more of a convention rule within spacemacs?

layers/+email/mu4e/packages.el
smile13241324: Ok, this sounds right, lets keep this deactivated otherwise we break lazy loading of org.

layers/+email/mu4e/packages.el
smile13241324: The big difference is that use-package implies a package installation, this however is controlled on the layer's top level by the mechanism of init-package-name. When you call use-package in an inner condition you are actually overwritting the layer system i.e. if someone did exclude the package of `org-mu4e` in his dotfile the layer would not trigger the package installation, in your case use-package would simply auto-install it if the right variable would be set as part of the layers loading process.

smile13241324 (approved):

smile13241324:

PR #14216, opened 2020-12-16
Add ace-link-man keybinding to navigation layer

dalanicolai:
Initially I added an ace-link-man keybinding to Man-mode. However, `ace-link-man` (as well as `ace-link`) only activates links to other man pages, but not url's. Therefore I changed it to `link-hint-open-link` in woman-mode as with this combination all links are activated correctly (using this command in man-mode also does not activate url's). An example man page that contains links to other man pages as well as url's is the ocrodjvu man page.

### DIFF
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -2174,9 +2174,9 @@ selecting an avy candidate.
 Similar to =avy=, [[https://github.com/abo-abo/ace-link][ace-link]] allows one to jump to any link in
 =help-mode= and =info-mode=.
 
-| Key binding | Description                                           |
-|-------------+-------------------------------------------------------|
-| ~o~         | initiate ace link mode in =help-mode= and =info-mode= |
+| Key binding | Description                                                 |
+|-------------+-------------------------------------------------------------|
+| ~o~         | initiate ace link mode in =help-=, =info-= and =woman-mode= |
 
 *** Unimpaired bindings
 Spacemacs comes with a built-in port of [[https://github.com/tpope/vim-unimpaired][tpope's vim-unimpaired]].
--- a/layers/+spacemacs/spacemacs-navigation/packages.el
+++ b/layers/+spacemacs/spacemacs-navigation/packages.el
@@ -34,12 +34,14 @@
     (progn
       (define-key spacemacs-buffer-mode-map "o" 'spacemacs/ace-buffer-links)
       (with-eval-after-load 'info
-        (define-key Info-mode-map "o" 'ace-link-info))
+        (define-key Info-mode-map "o" 'ace-link))
       (with-eval-after-load 'help-mode
-        (define-key help-mode-map "o" 'ace-link-help))
+        (define-key help-mode-map "o" 'ace-link))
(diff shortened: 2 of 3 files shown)


doc/DOCUMENTATION.org
```diff
@@ -2174,9 +2174,9 @@ selecting an avy candidate.
 Similar to =avy=, [[https://github.com/abo-abo/ace-link][ace-link]] allows one to jump to any link in
 =help-mode= and =info-mode=.
 
-| Key binding | Description                                           |
-|-------------+-------------------------------------------------------|
-| ~o~         | initiate ace link mode in =help-mode= and =info-mode= |
+| Key binding | Description                                               |
```
smile13241324, line 2177: This table looks misaligned.

layers/+spacemacs/spacemacs-navigation/packages.el
```diff
@@ -34,12 +34,14 @@
     (progn
       (define-key spacemacs-buffer-mode-map "o" 'spacemacs/ace-buffer-links)
       (with-eval-after-load 'info
-        (define-key Info-mode-map "o" 'ace-link-info))
```
smile13241324, line 37: Can you explain why you have replaced the mode specific functions with `ace-link`?

smile13241324 (changes requested):
Hi @dalanicolai,

I have just checked your PR and found a small documentation issue and some code changes I have questions about the reasons for the change.

Please have a look at my comments.

layers/+spacemacs/spacemacs-navigation/packages.el
dalanicolai: Ah I forgot to set them back to original. I've read the docstring in `ace-link-man` and it says this function will open a link in `man` buffer. Obviously, when a link in a man page is e.g. some url then you do not want to open that link in a man buffer. (indeed the action function in `ace-link-man` calls actually `#'man-follow`) So I just experimented if maybe using `ace-link` would highlight all links instead of only man page links.

Anyway, the docstring for ace-link says it will just run the link function for the current major mode, so it also calls `ace-link-man` (and similar for the other modes). I am not sure if it is necessary to call the mode specific functions again. Maybe at some point ace-link will get some more powerful functionality than the mode specific functions (i.e. opening different style links in their corresponding mode buffers).

I even think that possibly `link-hint-open-link` will also work better for the other buffers. But that should be tested first then.

So in the end, you can decide if you prefer to keep the mode specific or the general `ace-link` functions.

doc/DOCUMENTATION.org
dalanicolai: fixed

layers/+spacemacs/spacemacs-navigation/packages.el
smile13241324: I will give it a quick test and change it back to the mode specific one if necessary. If I see no issues using the more generic function is ok with me.

smile13241324 (approved):
Looks good so far, I will give it a quick test and merge it when all is ok.

smile13241324:

PR #14220, opened 2020-12-18
Add convenient scroll keybindings to epub layer

dalanicolai:
Similar to the keybindings added to the info layer in PR #14188, I here propose to add the same
keybindings to the epub layer. These J/K keybindings are currently undefined in the epub
layer. These keybindings are used in the Zathura document reader, they are very
convenient and I guess they feel very natural and are preferred over d/u by most people.


smile13241324 (approved):
:+1:  good idea, will be merged soon.

PR #14224, opened 2020-12-20
[base] Fix void helm-use-fuzzy

duianto:
problem:
The spacemacs-completion layer defines the variable:
helm-use-fuzzy

But the spacemacs-base distribution doesn't use the
spacemacs-completion layer by default.

helm-use-fuzzy is used in the helm layer.

When a helm buffer is opened, for example: `SPC SPC`
Then a warnings buffer opens:
>Error (use-package): helm/:config: Symbol’s value as variable is void: helm-use-fuzzy

solution:
Make sure that helm-use-fuzzy is bound and true.


smile13241324 (approved):
:+1:

PR #14234, opened 2020-12-23
[git] Restore magit transient args between sessions

duianto:
problem:
The Magit w (apply patches) arguments are not remembered between sessions.

cause:
Magit reads the transient files:
history.el, levels.el, and values.el
before Spacemacs has redirected them to the .cache directory.

solution:
Redirect the files before magit reads them.


smile13241324 (approved):
:+1:

PR #14235, opened 2020-12-23
[base] Fix evil undo

duianto:
problem:
Pressing: u
in the spacemacs-base distribution, shows:
evil-undo: Symbol’s function definition is void: undo-tree-undo

cause:
The evil undo system is setup as: undo-tree
but undo-tree is loaded in the spacemacs-editing layer,
and the layer isn't used by default in the spacemacs-base distribution.

solution:
Set the evil undo system to undo-tree when the undo-tree package is used.

Now if undo-tree isn't setup, then:
- u   uses the default Emacs undo command.
- C-r says: Customize ‘evil-undo-system’ for redo functionality.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -714,6 +714,7 @@ Other:
     (thanks to Ag Ibragimov)
   - Fixed duplicate quick help and release notes in home buffer
     (thanks to mark30247)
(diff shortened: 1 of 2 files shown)


smile13241324 (approved):
:+1:

martinxyz:
Big thank you. This was driving me crazy, I wasn't able to figure out the cause.

PR #14238, opened 2020-12-25
Configure org-roam to lazy load

codetrick:
Set up autoload for a few org-roam commands via use-package.
I also replaced a number of obsolete functions (org-roam-dailies-...).


smile13241324 (approved):
Cool, will be merged immediately :+1:

Issue #9505, opened 2017-08-28
Are you interested in yet another package manager?


raxod502:
I've created a new package manager [`straight.el`](https://github.com/raxod502/straight.el) as an alternative to `package.el`/Quelpa/Cask/Pallet/el-get/Borg. It appears that some people like it enough to mess around with hacking it into Spacemacs: https://github.com/raxod502/straight.el/issues/128.

Would you be interested in a pull request that adds an option to Spacemacs for using `straight.el` instead of `package.el` + Quelpa, much like the existing options for Ivy vs. Helm and Evil vs. One True Keybindings?

JAremko:
@raxod502  Does it work with Windows ?

cc @syl20bnr

raxod502:
No, see https://github.com/raxod502/straight.el/issues/124.

syl20bnr:
@raxod502 That's funny you opened this issue, we talked about your package last month. I'm considering it to back the layer system and to replace the current (not released) approach to create a `spacelpa` repository to freeze the version of packages for master branch. `staight.el` would be a far lighter solution than maintaining a whole elpa repository.

> Would you be interested in a pull request that adds an option to Spacemacs for using straight.el instead of package.el + Quelpa

Yes of course but the current layer system relies a lot on `package.el`!
The layer system code is in `core/configuration-layer.el: https://github.com/syl20bnr/spacemacs/blob/develop/core/core-configuration-layer.el

> much like the existing options for Ivy vs. Helm
As it is in `core` we don't need to support several alternatives, we would simply replace the current backend by `straight.el`.

A few questions:

1) What are the performance of `straight.el` vs. `package.el` when installing several hundreds of packages ? 

2) Do you use `package.el` behind the scene or are you reimplementing some `package.el` feature ?

3) What's your plan for the future of this project ? Do you consider to publish in Gnu ELPA ?

syl20bnr:
Also like @JAremko said, we need `windows` support ;-)

raxod502:
> 1. What are the performance of straight.el vs. package.el when installing several hundreds of packages ?

Installing packages is more or less the same speed as `package.el` in my experience. You might think that doing full Git clones would be slower—and it is, for big repos like Org—but for the most part there's no observable difference since the deciding factor is network latency. (Since `package.el` relies on a central server, it *could* speed this up by using bulk calls… but it doesn't.)

If you'd like me to do an official benchmark, I can. It would be pretty trivial.

[...]

PR #14240, opened 2020-12-25
use straight package manager instead of built-in package.el

emacs18:
https://github.com/syl20bnr/spacemacs/issues/9505 discussed using https://github.com/raxod502/straight.el package manager.

This PR is to explore how to do that.  Once this PR is applied to develop git branch, and emacs started, it will use `straight` to install most packages.  The key changes are
- setup `straight` package manager in early-init.el
- install `use-package` using straight, then set `straight-use-package-by-default` to `t` so that subsequent use of `use-package` will end up using `straight`
- advise `package-install` also to use `straight-use-package` as opposed to what it does by default which is to download the package tarball, then install it
- add `:straight nil` within some `use-package` forms which `straight` does not know how to deal with, e.g., spacemacs built-in packages such as `hybrid-mode`

This seems to work for the most part.  Some packages are installed via `package.el` such as most packages built into spacemacs.  This is not meant to be checked in.  Rather  I hope this to help discuss how spacemacs can support `straight` in addition to `package.el` so that users can choose.


lebensterben:
+1
Among all benefits off straight.el, the most important one for me is that it downloads package as git repository, which makes contribution much easier. This also provides an opportunity to roll back a particular package easily with standard Git operations, should user ever encounter a bug in an package update.

emacs18:
I started using `straight` for my spacemacs.  It seems to be working out for now.  As shown below only 7 packages were installed by `package.el`, and 352 packages installed by `straight` from 329 repositories.  `site` branch has my own changes on top of `develop` branch as detailed at https://github.com/emacs18/spacemacs fork.

```
$ cd ~/repos/my/spacemacs
$ find elpa -type d
elpa
elpa/gnupg
elpa/gnupg/private-keys-v1.d
elpa/27.1
elpa/27.1/site
elpa/27.1/site/archives
elpa/27.1/site/archives/gnu
elpa/27.1/site/archives/melpa
elpa/27.1/site/archives/org
elpa/27.1/site/evil-unimpaired-20200927.133103
elpa/27.1/site/s-20200330.1354
elpa/27.1/site/evil-20201107.1030
elpa/27.1/site/hybrid-mode-20201018.152504
elpa/27.1/site/f-20191110.557
elpa/27.1/site/dash-20201223.955
elpa/27.1/site/goto-chg-20201101.229
$ ls straight/build | wc -l
352
$ ls straight/repos | wc -l
329
$ 
```

smile13241324:
I am still a bit reluctant to globally replace package.el however I think this would definitely qualify for an option in the dotfile. Especially as having repos for packages may ease contributing to packages. Though I think this is something only advanced users will want to have.

early-init.el
```diff
@@ -22,3 +22,73 @@
...
+
+;; Install use-package first and setup so that use-package will use straight.el
+;; rather than package.el by default.
+(straight-use-package 'use-package)
+
+;; Following is from https://github.com/raxod502/straight.el/issues/352
+;; This can be fixed when org 9.5 is released where org and contrib will be separate packages.
+(straight-use-package '(org-plus-contrib :includes (org)))
+
+;; These are needed after removing copies from core/libs.
```
JAremko, line 66: Heaving some libs built in preferable for:
- Making Spacemacs some kind of working  even if network is down or upstream has a bug.
- Speeds up install/update.
- At this stage the Spacemacs package source configs aren't available. 
- We already have an semi-auto-update system for libs https://github.com/syl20bnr/spacemacs/pull/14285 so them going stale no longer a problem.
-  Some of the libs are forks. For example, upstream version of `load-env-vars` won't work on Windows.

Also this change is out of scope of the PR.

JAremko:
Blockers for this would be:
- it should  work with Spacelpa
-   And [testing packages archive](https://github.com/JAremko/testelpa-develop/releases)  - daily build  with all the packages used by tests.  CircleCI PR code uses it to speed the tests. We have a lot of PR test so we need this system to work or the tests will take ages to complete.   
-  Fresh install with `straight.el` used to take a lot of time and the package dir was huge.  How bad/good is it now?
-  Heaving  `straight.el` as an option would be much nicer than the only way to use Spacemacs. But then how to configure it... Something like starting usual install from `package.el` and then promoting the user like we do with editing style could work :thinking: 
-  @syl20bnr  has to approve change this big.

progfolio:
FYI:

```emacs-lisp
 (straight-use-package '(org-plus-contrib :includes (org)))
 ```
 There is no `:includes` keyword for straight.el. It was mentioned as an idea here:
 
 https://github.com/raxod502/straight.el/issues/352#issuecomment-460069774
 
 And seems to have proliferated via copy+paste despite not actually being implemented.
 The latest default org-plus-contrib recipe should do the right thing until Org splits "contrib" into its own repository.

emacs18:
I would like to remind folks that this PR is not intended to be checked in.  It is meant only to point out what changes would be needed to get going with straight.el.  I have two primary goals for using straight.el.

- Specify the version that I use for every package similar to how doom-emacs specifies (or pins) git version for most packages.
- Use the same packages for all emacs configurations such as spacemacs, doom, scimax, etc.
 
For now I can select spacemacs, doom, scimax, or purcell init files on emacs startup.  The config name is specified on command line.  Regardless of which one I use, the same packages installed by straight.el are used.  My hope is to setup emacs configuration in my company network so that any one of my hundreds or thousands of emacs using colleagues can use these common setups without having to install any package or even require internet access.

raxod502:
As the original author of `straight.el`, happy to advise as needed here!

> Fresh install with straight.el used to take a lot of time and the package dir was huge. How bad/good is it now?

Installation speed has been increased by using shallow clones, but https://github.com/raxod502/straight.el/issues/149 and/or https://github.com/raxod502/straight.el/issues/510 have not yet been implemented.

dalanicolai:
Maybe I should drop this comment elsewhere (as a separate issue), as this is not really directly related to `straight.el`. But I read above that @smile13241324 and @lebensterben mention that `straight.el` would allow for using git repos. But we can already use local packages (i.a. git repos directly) when using the `:location local` syntax, either from a layer's `packages.el` or from `dotspacemacs-additional-packages`. So couldn't the issue for contributing to packages get solved by letting users "override" the `packages.el` declaration from the dotspacemacs file?

Of course in this way, those git repo's do not get automatically updated (while I guess straight.el probably can also update the git repos automatically), but at least it would be nice for making contributions to packages easier (we should be able to switch between local installed to quelpa installed packages from the dotfile). (Or is this maybe somehow possible already?)

dalanicolai:
To elaborate a little more, I have found that when adding some package, that is already installed from a layer, to the list of `dotspacemacs-additional-packages` with the `:location local` syntax, will change the load path for the package (from the `elpa` directory to the `layer/local/package-name` directory). But then the `layer-name/init-package-name` function in `packages.el` does not find it anymore. So it would be great if adding the "local package" to `dotspacemacs-additional-packages`, would also make that layer/init function search for the package in the  `layer/local/package-name` directory, or even better in the `.emacs.d/private/local` directory. Would that be possible?

syl20bnr:
Great effort to integrate with `straight.el`. 

Unfortunately I'm not anymore in favor to use this package manager.

Supporting two different package managers looks like a maintenance hell and not something that would really benefit the project. :-(

The main reason I changed my mind is that we are going to leverage spacelpa and we will take a more emacsy approach to package pinning.
Another reason is that `straight.el` is kind of a beast with lots of code and lots of unneeded feature.

I don't feel that relying on such big package which in turn is not relying on official `package.el` by design is a good idea in the long run. From the start the project while very ambitious is not meant to be contributed back to GNU Elpa or integrate well with its ecosystem. I learnt to be careful with such projects.

I largely prefer the approach of Quelpa which acts as a layer on top of `package.el` and leverage MELPA (that does the same). Our integration of Quelpa already gives some flexibility brought by `straight.el` and `el-get`.

I hear you that it has convenient feature like easy fork of a given package etc... This is something we could add ourselves later maybe with a different approach. 

We can re-open the discussion if/when `straight.el` is adopted and supported by the Emacs developers. 

This being said, @emacs18 I'm interested to know about your use cases for `straight.el` and see what we could do to support them if they can be useful for others.

syl20bnr:
Reading about @dalanicolai comments, this is definitely a better and smarter way to provide such capability.

Autoloads can indeed be an issue for packages with an overridden location, I guess that in this case the configuration prodived by Spacemacs can break. This is also a current issue for plain local packages, we don't support autoloads. We could kill two birds with one stone if we find a way to bring autoload capability to local packages. Mhhhh we could take inspiration on `straight.el` 🤔 

@emacs18 apart from loading packages from a local git repo, what kind of feature do you need from `straight.el` ?

lebensterben:
@syl20bnr 

Image we add a `:patches` keyword which allows us to specify a list of local (git) patches files which would be applied after the upstream package is downloaded, but before compiled. This allows both spacemacs developers and users an easy way to add local modifications without forking an entire package.

This will greatly helps us to reduce the size of spacemacs as well.

emacs18:
@syl20bnr These are the features of straight.el that I'm interested in.
- ability to "freeze" versions where the git revisions of each and every package is saved away into a file so that the exact state can be restored later on should the need arise.
- the fact that all `*.el` files that are located via `C-h f` or `C-h v` are in their respective git repos so that should any change need to be made, then I can simply create a local branch and check in the change.  So it is very hassle free to tweak things.
-  share package installation across different startup configurations of spacemacs, doom-emacs, scimax, etc.  I should expand this point further, but it may require too many words, so I'll stop here.

dalanicolai:
Reading the latest comments and to keep things in one place, I am just mentioning that current [instructions for overriding layer packages](https://develop.spacemacs.org/doc/DOCUMENTATION.html#overriding-a-layer-package) have been added to the [FAQs](https://develop.spacemacs.org/doc/FAQ.html#how-to-override-a-layer-package). (#14491)

lebensterben:
@dalanicolai 
That's a full override.
Comparatively the patching approach I described has the additional property of failing-early, which is very desirable.

dalanicolai:
@lebensterben I agree. Just I thought it would be handy to refer to those instructions and mention the PR in this thread. But I explicitly wrote "current instructions". I don't have a strong opinion about it because, until now, I do not need (or don't know how to use) more advanced package management. But I'm sure you will decide on a great design and would be happy to learn from and use it.

syl20bnr:
> ability to "freeze" versions where the git revisions of each and every package is saved away into a file so that the exact state can be restored later on should the need arise.

This will be handled by Spacelpa. Having our own ELPA repository will bring some advantages I'll talk about later.

> the fact that all *.el files that are located via C-h f or C-h v are in their respective git repos so that should any change need to be made, then I can simply create a local branch and check in the change. So it is very hassle free to tweak things.

We might be able to achieve this with the system in place. It would be an opt-in per package though, and I feel this is better this way than the other way around. I see two main things that need to be done:
- generate autoloads
- load from and point to the cloned Git repositories of development packages

> Imagine we add a :patches keyword which allows us to specify a list of local (git) patch files which would be applied after the upstream package is downloaded, but before compiled. 

Patches can be hard to handle as they need to resolve without conflicts to be applied. Better to allow people to contribute directly in the context of a git repo and have the complete git commands arsenal at their disposal.

edrex:
> Unfortunately I'm not anymore in favor to use this package manager.

I hope feelings will loosen up on this. I observe this issue driving an effective fork of the community. Honestly, doom does not have the excellent polish and consistency of spacemacs, but I'm driven to it as I migrate from Arch to NixOS because straight.el fits in with the way the rest of the software on my system is updated so much better (eg, https://github.com/vlaci/nix-doom-emacs). I think a lot of folks jumping to doom would come back to spacemacs if it learned to update via straight.el. If that's not in the cards, maybe an external project can streamline using spacemacs layers from straight? Seems like folks are already doing that. Fragmentation is always a little sad when the community is already stretched so thing. Anyway I wish you all the best!

thanhvg:
I tried this branch and it worked, great job @emacs18 it showed that many still care about Spacemacs and see its values. However it took ages to install with about a gig of extra data. Probably that's why emacs Doom has to outsource this job to a standalone cli app to improve the speed.

I have both emacs 27 and 28 (the native-compilation branch) running well on the same spacemacs but I can see the difficulty in making they work together with `straight.el` without a loader like `cheesemacs`.

Anyway I like to remind @syl20bnr that you have been promising Spacelpa for years. When I first I saw your announcement on gitter about Spacelpa, emacs 25 was the bleeding edge verion, Doom was still a name for a video game. Since then I got to learn elisp then git to use Spacemacs, emacs saw one or two major releases. Still the feature is no where to be seen. Please don't take this as if I were blaming you for anything, I am a happy Spacemacs user, and I grateful that you made it. But to be blunt I don't buy your Spacelpa argument at all.

syl20bnr:
@edrex I'm not familiar with Nix and I have a few questions.

Why is this so important to have layers understood by Nix, Spacemacs has its own package manager already ? If Spacemacs provides a reproducible state with Spacemacs repo + `.spacemacs` is this not enough ?
Spacelpa will solve the issue of reproducible state as a stable version of Spacemacs will be bound to a frozen ELPA. Any change in Spacemacs lock file + `.spacemacs` could be a Nix version maybe ? I'm speculating I surely miss the basic concepts of Nix.

syl20bnr:
> Anyway I like to remind @syl20bnr that you have been promising Spacelpa for years. 

You are right and I'm not getting the shit done on this front. I need people to remind me of this. No excuse I have to get back to it and finishing it.

JAremko:
@syl20bnr  We might do something like this to generate weekly spacelpa for develop https://github.com/JAremko/testelpa-develop

JAremko:
Yeah and instead of messing with signs we can make our bot create a list of tagged valid hashes of spacelpa-develop releases. Then a end user can fetch one of them by tag or hash.

syl20bnr:
@JAremko let's start a discussion about this and ping me in it.

edrex:
> @edrex I'm not familiar with Nix and I have a few questions.

I am pretty new to Nix so I hope what I say is at least basically accurate.
 
> Why is this so important to have layers understood by Nix, Spacemacs has its own package manager already ? If Spacemacs provides a reproducible state with Spacemacs repo + `.spacemacs` is this not enough ?

Nix doesn't necessarily need to understand the individual layers, but it wants to be able to verify the hashes of the contents of all the build inputs.

> Spacelpa will solve the issue of reproducible state as a stable version of Spacemacs will be bound to a frozen ELPA. Any change in Spacemacs lock file + `.spacemacs` could be a Nix version maybe ? I'm speculating I surely miss the basic concepts of Nix.

If the entire Spacelpa repo is usually downloaded as a tarball, the Nix expression could do the download ahead of time and verify it using a provided SHA. Similar for a git ref. This would make managing Spacemacs "in the Nix store" much more tractable vs the current situation where individual packages are fetched.

An aside: for more granular inputs there are tools for translating the dependency trees of various language/application specific package managers into equivalent nix expressions (eg cabal2nix). This is likely to be done with straight.el eventually ([1](https://github.com/raxod502/straight.el/issues/105#issuecomment-557836799))

If part of what spacelpa is intended to provide is a compile cache for gccEmacs (seems like a good idea as more folks are jumping over), trying to track that cache via the nix store seems wrong. I'm not sure if it would be considered OK to just let Spacemacs fetch the cached compilations. It's beyond my current understanding of Nix. Maybe a much more experienced Nix dev who uses Spacemacs eg @Mic92 could chime in. AFAIK https://github.com/nix-community/emacs-overlay provides a daily compile cache of MELPA packages, so maybe that could be leveraged.

Mic92:
I am actually using doom-emacs because it has better integration into nix: https://github.com/vlaci/nix-doom-emacs
Also the fact that doom-emacs is pinning packages for me makes it a lot more usable to me,
since they make sure that configuration still works after updating packages.

lebensterben:
How would it be less usable to you as we don't pin package version?

Mic92:
If upstream or me don't update spacemacs fast enough to keep up with the emacs packages, than spacemacs provided configuration could be no longer valid. So it is a bit of a gamble if it still works. For doom-emacs this is atomic as package updates is done by doom-emacs itself. This is different when using `Spacelpa` so, which I did not look at as it was not ready at the time I left.

lebensterben:
In my use case I update Spacemacs several times a week and update emacs packages every time I upgrade Spacemacs.

I cannot recall any case where I found some packages stopped working... There were a few times I saw the community reports some bugs but usually it's resolved very quickly.

Also, you can freely checkout any older commit of Spacemacs and rollback package updates.

So pinning package is really of no utility to me.

emacs18:
My experience is different.  I read  only a small fraction of spacemacs issues, but I recall discussing several problems reported that were due to lack of proper version control of spacemacs and packages.  Sure if you update spacemacs and reinstall all packages at the same time, then chances of running into problems are very small.  However some folks update spacemacs only without updating all packages, or update packages without updating spacemacs.  Such updating of only spacemacs or packages has caused a variety of problems.  I personally ran into such issues several times.

It is nearly impossible to run into this type of problems with doom-emacs due to package version pinning.  With spacemacs everyone is rolling the dice whenever packages are installed, because package versions at melpa are constantly updated.  You never know which version will be installed when you install spacemacs or update packages after installing.  Also every time you update spacemacs you are almost forced to update all packages if you don't want to run into problems due to version mismatches.  I doubt that many folks reinstall all packages after each spacemacs update.

lebensterben:
To be clear, you don't need to reinstall packages after you update Spacemacs.

You only need to update existing package.

If you really want, it's easy to make a new function to update Spacemacs and emacs packages at the same time.

Pinning package is a no-go for me. My system package manager is on rolling release + auto update. The security risk of using old packages always outweighs the marginal gain of stability, if it exists.

smile13241324:

Issue #13073, opened 2019-12-15
Reloading config with fullscreen-at-startup t (in macOS) toggles fullscreen mode


tshu-w:
#### Description :octocat:
If I reload config on macOS with `dotspacemacs-fullscreen-at-startup t`, it make frame fullscreen.

#### Reproduction guide :beetle:
- Start Emacs
- toggle fullscreen mode
- spc-f-e-R

*Observed behaviour:* :eyes: :broken_heart:
config reload and fullscreen frame

*Expected behaviour:* :heart: :smile:
config reload without fullscreen frame

#### System Info :computer:
- OS: darwin
- Emacs: 26.3
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 6649e78f3)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(helm osx
      (chinese :variables chinese-enable-fcitx t)
      dash git
      (version-control :variables version-control-diff-tool 'diff-hl)
      github
      (pandoc :variables pandoc-data-dir "~/.spacemacs.d/pandoc-mode")
      lsp dap
      (c-c++ :variables c-c++-default-mode-for-headers 'c++-mode c-c++-enable-clang-support nil c-c++-enable-google-style nil)
      emacs-lisp
      (python :variables python-backend 'anaconda python-sort-imports-on-save t python-formatter 'black python-format-on-save t blacken-fast-unsafe t python-fill-column 99)
      (markdown :variables markdown-command "pandoc -t html5 -f markdown+smart --mathjax --highlight-style=pygments --toc --toc-depth 3 --template github.html5 --css html/css/github.css" markdown-live-preview-engine 'pandoc)
      (org :variables org-agenda-files
           '("~/Documents/Org")
           org-directory '"~/Documents/Org" org-projectile-file "TODOs.org" org-want-todo-bindings t org-enable-hugo-support t org-enable-reveal-js-support nil org-enable-org-journal-support t org-journal-dir "~/Documents/Org/Journals/")
      (latex :variables latex-enable-auto-fill nil latex-enable-folding t latex-enable-magic nil)
      (bibtex :variables org-ref-default-bibliography
              '("~/Dropbox/Documents/Zotero/refs.bib")
              org-ref-pdf-directory "~/Dropbox/Documents/Zotero/" org-ref-bibliography-notes "~/Dropbox/Documents/Zotero/notes.org")
[...]

PR #14241, opened 2020-12-25
[defaults] Prevent fullscreen on config reload

duianto:
problem:
Reloading the configuration: SPC f e R
after changing the variable: dotspacemacs-fullscreen-at-startup
to: t
changes Emacs into fullscreen mode.

solution:
Only check the variable and change to fullscreen mode on startup,
before Spacemacs has been initialized.

Fixes: Reloading config with fullscreen-at-startup t (in macOS) toggles fullscreen mode #13073

Additional change:
`if` was changed to `when`, because there is no else statement.


smile13241324 (approved):
:+1:

PR #14242, opened 2020-12-26
Make ipython notebook work with company-mode

senghoo:
By using the Jupyter notebook api, the ipython-notebook layer can be auto-completed by using the company-mode

### DIFF
--- a/layers/+tools/ipython-notebook/README.org
+++ b/layers/+tools/ipython-notebook/README.org
@@ -11,6 +11,7 @@
   - [[#deleting-visual-regions-dont-work-find-out-why][Deleting visual regions don't work, find out why]]
 - [[#install][Install]]
   - [[#layer][Layer]]
+  - [[#choosing-a-backend][Choosing a backend]]
   - [[#dependencies][Dependencies]]
   - [[#what-needs-to-be-run][What needs to be run]]
 - [[#using-the-ipython-notebook][Using the IPython notebook]]
@@ -34,6 +35,7 @@ lots of informative stuff.
 ** Features:
 - Key bindings available through transient-state or leader key
 - Lazy-loading
+- Auto-completion
 
 * List of TODOS
 This is a WIP, feel free to collaborate.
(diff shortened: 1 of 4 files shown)


robbyoconnor:
I do wonder if this should be hidden behind a toggle?

robbyoconnor (changes requested):
You should update the docs as well..

robbyoconnor (approved):
Mergey merge please @duianto

robbyoconnor:
Nice work!

duianto (commented):
Squash the commits into one.

Note 1/2:
I don't know anything about ipython notebook, or jupyter.

I installed it (https://jupyter.org/install) with:
`pip install jupyterlab`

started it with:
`jupyter-lab`

That opened a JupyterLab page in a browser.

In Spacemacs, I pressed: `SPC a t i r` (which calls: `ein:run`)
It opened a buffer called: `*ein:notebooklist http://127.0.0.1:8889*`

So something seems to be working, I see no errors either with or without this PR applied. So this seems to be good to go when the minor suggestions have been resolved.

Note 2/2:
I noticed that the Spacemacs readme for the `ipython-notebook` layer:
https://github.com/syl20bnr/spacemacs/tree/develop/layers/%2Btools/ipython-notebook

States the installation instructions as:
`pip install jupyter`

But the https://jupyter.org/install instructions say:
`pip install jupyterlab`

Is this the new/current/recommended command?

The readme also mentions the key binding: `SPC a t i o`
which is supposed to call: `ein:notebooklist-open`
But I don't have that key binding.

Could someone with some ipython-notebook, and jupyter knowledge, open a PR that updates the layer readme so that it's up to date?

robbyoconnor:
> Is this the new/current/recommended command?

Both are valid.

senghoo:
@duianto  I merged your suggestions and the commites into one. and I will continue to improve the this layer in the future PR.

duianto:

PR #14243, opened 2020-12-26
[docs] Elaborate on fetchers to include git

bhipple:
Although the docs show an example, but the paragraph below seems to imply that a
package can't be fetched directly off git. Update to make it more clear that it can.

### DIFF
--- a/doc/LAYERS.org
+++ b/doc/LAYERS.org
@@ -335,7 +335,8 @@ as follows:
 #+END_SRC
 
 The =:location= attribute specifies where the package may be found. Spacemacs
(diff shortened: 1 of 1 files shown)


lebensterben:
`git packages` sounds strange.
Something like `remote packages hosted on Git repositories` sounds better to me.

smile13241324 (approved):
:+1:

smile13241324:
I have also made a small commit and incorporated the suggestions from @lebensterben.

bhipple:
Thanks @smile13241324!  I agree the rewording is an improvement, ty for including that.

PR #14247, opened 2020-12-30
[treemacs] Check for git and python 3 upstream

duianto:
problem:
The Treemacs layer is unable to find python 3 in Windows.

cause:
The Treemacs layer looks for "python3", but the executable is called
"python.exe" in Windows.

solution:
Let Treemacs check for python 3 upstream, it has a better method for finding the
python 3 executable in Windows.


smile13241324 (approved):
:+1:

PR #14248, opened 2020-12-31
Add search-from ivy action to fasd layer

dalanicolai:
If I had known about browsing history in Ivy with `C-r` before
 [I modified the fasd package to implement full ivy support](https://framagit.org/steckerhalter/emacs-fasd/-/commit/c4c04873fd0c8e916186f38a75cd4ab7f8b7ad59),
then I guess I would not have modified it. Anyway, now that "full Ivy support"
has been merged into the fasd package, I could add this 'search-from' action
command to `fasd-find-file` which triggers Spacemacs 'search-auto'. Because fasd
only finds dirs that have been visited before, it probably does not add much to
the `C-r` functionality, but maybe it does (you decide?). Anyway, I hope you can
help decide if this is useful for merging.

So with this PR you can use fasd to find some directory, then use `M-o` to trigger
`search-auto` to search from that directory (I got tired of navigating to often searched directories).

If you think this PR is useful enough then I am happy to 'git amend' documentation
 and add the change to the changelog.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1972,6 +1972,8 @@ Other:
 - Support running graphical Factor listener instances from Spacemacs
 - Support reloading factor-mode/fuel-mode code when connecting to different
   Factor versions
+**** FASD
+- Add ~search from~ ivy action (thanks to Daniel Nicolai)
 **** Finance
 - Remove key-bindings pointing to removed commands (thanks to Alexander Baier)
 - Added =evil-ledger= package (thanks to Alexander Miller)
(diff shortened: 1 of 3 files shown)


smile13241324:
I think this is still useful, I am not a `fasd` user but after you have entered it in the package I think it would be a waste not to bind it here @dalanicolai.

smile13241324 (changes requested):
The code is ok, but the documentation is missing, when added I would merge it.

dalanicolai:
I thought I will not extend the docs for this almost redundant feature, but I extended it now.

smile13241324:
Manually merged.

PR #14249, opened 2020-12-31
Add consistent org-agenda transient-state keybinding

dalanicolai:
I do not see any reason why the keybinding to enter transient-state in
 org-agenda-mode should not be on `, .`. Therefore I propose here to
make that keybinding consistent with the keybindings for entering
transient states in most/all other modes.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -277,8 +277,6 @@ sane way, here is the complete list of changed key bindings
   - ~SPC m s r~ org-refile
   - ~SPC m s s~ org-sparse-tree
   - ~SPC m x o~ org-open-at-point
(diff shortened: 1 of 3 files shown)


smile13241324 (approved):
Big 👍 , this was really missing here.

For the record this is also ok with the latest version of the key binding conventions:
>*** Transient-state
>Transient states are generally put behin the dot key ~.~, for instance the
>windows manipulation transient state in ~SPC w .~.
>
>When it is not possible to use the ~.~ (like in =helm= buffers) transient states
>should be enabled with ~M-SPC~ and ~s-M-SPC~. We need the latter bindings on
>macOS since ~M-SPC~ is used by the OS for spotlight.
>
>It is recommended to make sure that ~q~ allows to leave the transient-state.

smile13241324:

PR #14252, opened 2021-01-01
Support prettier formatter for Ruby layer

sidraval:
This PR adds support for [prettier/plugin-ruby](https://github.com/prettier/plugin-ruby) in spacemacs' Ruby layer. Prettier can be configured to run on buffer save, or run ad-hoc via a keybinding. I tried to follow the implementation used by other layers for autoformatters, e.g. the javascript layer & typescript layer.

### DIFF
--- a/layers/+lang/ruby/layers.el
+++ b/layers/+lang/ruby/layers.el
@@ -11,4 +11,7 @@
 
 (when (and (boundp 'ruby-backend)
            (eq ruby-backend 'lsp))
-  (configuration-layer/declare-layer-dependencies '(dap prettier)))
+  (configuration-layer/declare-layer-dependencies '(dap)))
+
+(when (boundp 'ruby-prettier-on-save)
+  (configuration-layer/declare-layer-dependencies '(prettier)))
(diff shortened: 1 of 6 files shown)


layers/+lang/ruby/layers.el
```diff
@@ -11,4 +11,4 @@
 
 (when (and (boundp 'ruby-backend)
            (eq ruby-backend 'lsp))
-  (configuration-layer/declare-layer-dependencies '(dap)))
+  (configuration-layer/declare-layer-dependencies '(dap prettier)))
```
smile13241324, line 14: Hmmm thats not right dude, dap is an auxiliary layer for lsp server integration, however you want to have prettier required independently of lsp-mode, instead you need to check for your configuration variable and only require this layer if prettier is actually used.

smile13241324 (changes requested):
I have reviewed your PR and found a small issue with your layer dependency please have a look at my more detail comment.

The rest looks pretty ok.

layers/+lang/ruby/layers.el
sidraval: Ah of course, thanks. Updated.

smile13241324 (approved):
Looks good

Issue #14250, opened 2020-12-31
`K` runs `org-agenda-capture` instead of hiding habits in org-agenda 


dbankmann:
Hitting `K` in `org-agenda` with active `org-habits` should toggle habits, but instead opens `org-agenda-capture`.

`SPC h d k K` shows
```K runs the command org-agenda-capture, which is an interactive compiled Lisp
function in ‘org-agenda.el’.

It is bound to K, <evilified-state> K, <menu-bar> <Agenda> <Capture with cursor
date>.
```
`org-agenda.el` only shows a keymapping from "k" to org-agenda-capture.

`SPC h d k k` shows
```k runs the command org-agenda-previous-line, which is an interactive compiled
Lisp function in ‘org-agenda.el’.

It is bound to k, <up>, C-p, p, <evilified-state> k.
```

#### System Info :computer:
- OS: gnu/linux
- Emacs: 27.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. d545dcbdb)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(graphviz yaml csv csv html helm auto-completion better-defaults emacs-lisp git kotlin fasd copy-as-format markdown org c-c++ finance
          (rcirc :variables rcirc-enable-znc-support t)
          (shell :variables shell-default-height 30 shell-default-position 'bottom)
          (spell-checking :variables spell-checking-enable-auto-dictionary nil)
          treemacs version-control
          (lsp :variables lsp-ui-sideline-show-hover nil lsp-ui-doc-enable nil lsp-ui-sideline-update-mode 'point)
          (python :variables python-backend 'lsp python-lsp-server 'pyls python-test-runner 'pytest python-formatter 'yapf python-format-on-save t python-sort-imports-on-save t)
          sphinx
          (mu4e :variables mu4e-use-maildirs-extension nil mu4e-enable-mode-line t))
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG CAIRO SOUND GPM DBUS GSETTINGS GLIB NOTIFY INOTIFY ACL GNUTLS LIBXML2 FREETYPE HARFBUZZ M17N_FLT LIBOTF ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XDBE XIM MODULES THREADS LIBSYSTEMD JSON PDUMPER LCMS2 GMP

PR #14256, opened 2021-01-03
Fix issue #14250 by unshadowing 'K'

dbankmann:
Unshadows 'K' in agenda-mode-map such that 'K' still maps to `org-habit-toggle-display-in-agenda`


smile13241324 (approved):
Hmmm by doing this we win the original binding from `K` however we will also lose the original binding on `k` 🤔. 

Lets give it a try, right now I have no better idea how to handle this situation without introducing additional confusing evil only bindings.

dbankmann:
As far as I understand, the original binding on `k` must have been `org-agenda-capture`, which is mapped to `, c`  now anyways.

smile13241324:
Yep thats true most people won't miss it I presume.

PR #14258, opened 2021-01-04
[base] Show a winum required message

duianto:
### problem:
key bindings that use the `winum` package,
don't work in the `spacemacs-base` distribution.

for example in the buffer transient state: `SPC b .`
`{number}`, `C-{number}` or `M-{number}`

shows messages like these:
>Wrong type argument: commandp, winum-select-window-1
Symbol’s function definition is void: winum-get-window-by-number

### cause:
the `winum` package is loaded in the `spacemacs-navigation` layer,
but the layer isn't used in the `spacemacs-base` distribution.

### solution:
show the message:
>This command requires the winum package,
winum is part of the spacemacs-navigation layer.

### Additional changes:
The functions:
```
spacemacs/move-buffer-to-window
spacemacs/swap-buffers-to-window
```
are only defined when the `winum` package is used,
because they use `winum` functions.

And the `neotree` and `treemacs` layers, remapping of:
`winum-select-window-0-or-10`
is limited to when the `winum` package is used.

There were no errors before these conditions were applied,
but it might be better to only load them when they are used.

### Note:
I'm unable to test if the `osx` layer keys work, but they are probably fine, 
since the same command name change was applied to the keys in
the buffer and window transient states, and in magit buffers.


smile13241324 (approved):
👍

PR #14259, opened 2021-01-04
[core] support recents-by-project in dotspacemacs-startup-lists

Kazark:
`dotspacemacs-startup-lists` by default shows a number of recent files and projects as two separate lists. If I've been working with a lot of files in one project, then all the recent files are from one project, even if I set `recents` to a large amount like 24. This change allows me to see the recent files by project. Suppose, for example, I have a `vegetables` project and a `fruit` project, and set `dotspacemacs-startup-lists` to `(recents-by-file . (2 . 3))`. In the home buffer I will see something like:

    ~/vegetables
        lettuce.el
        squash.el
        tomatoes.el
    ~/fruit
        apple.py
        orange.py
        banana.py

Even though only a subpath is displayed for each file, the click functionality still works---i.e. the link still has the full path under the covers.

I originally asked a [question](https://emacs.stackexchange.com/q/62524/19069) on Emacs StackExchange to see if there were any pointers or if this was already a solved problem. After several days of receiving no answers, and having a little time to poke at it, I figured I'd implement it myself.

What this does not cover: mixing recent files totally outside projects into this list. Today they are just filtered out. That is a usecase I didn't need so I figured that could be done in a subsequent pass if somebody wanted it.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -529,6 +529,9 @@ Other:
     terminal mode, but no longer in GUI mode. (emacs18)
 *** Core changes
 - Improvements:
+  - Support =recents-by-project= in =dotspacemacs-startup-lists=, which allows
+    combining the existing =recents= and =projects= into one coherent view (thanks
+    to Keith Pinson)
   - Make =set-selective-display=, which is a kind of built-in Emacs quick-n-dirty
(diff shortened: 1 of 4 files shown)


smile13241324 (approved):
Cool feature 👍, will be merged soon.

PR #14260, opened 2021-01-04
[emacs-lisp] make `SPC m h h` and jump handlers work in `ielm`

Kazark:
Like in other Emacs Lisp modes.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -134,7 +134,7 @@ the [[file:CHANGELOG.org][CHANGELOG.org]] file.
   (thanks to Cormac Cannon)
 - Removed cquery.
 ***** EPUB
-- Add ~J/K~ for scrolling down/up (thanks to Daniel Nicolai) 
+- Add ~J/K~ for scrolling down/up (thanks to Daniel Nicolai)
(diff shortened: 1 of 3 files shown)


lebensterben:
-1 for `g d`
Normally REPL buffers are opened as a side window, that users may want it to be persistent.
With `g d`, the ielm buffer is replaced by a buffer of the source file, which in my view is quite opposite to how REPLs are used.

Kazark:
@lebensterben perhaps I mistitled the PR. The `g d` binding was already in place, and just didn't work. What I did was wired in the jump system so you can use any binding you want (e.g. `SPC m g D` in order to not replace the window)---or not use it at all, of course. I mentioned `g d` because that's what's in my muscle memory; that's what I've been frustrated before by the lack of in `ielm`.

lebensterben:
oops since it never works I thought it's not added.

Kazark:
@lebensterben I expunged the misnomer (despite the fact that this will make `g d` work).

smile13241324 (approved):
Nice fix, will be merged soon. 
Yes that `g d` binding is not nice however we should change it in a separate PR.

PR #14261, opened 2021-01-05
helm: Better integration with search tools and etc

lebensterben:
This commit includes multiple changes to `helm` layer. Aside from documentation
and reformatting, nontrivial changes are listed below.

- Changed search tool integration interface.
  - Currently, supported search TOOLs are `rg`, `ag`, `pt`, `ack` and `grep`.
    And the following search functions are implemented:
    - `spacemacs/helm-buffers-do-<TOOL>`: Search within all opened buffers.
    - `spacemacs/helm-dir-do-<TOOL>`: Search within current directory.
    - `spacemacs/helm-files-do-<TOOL>`: Search within a specified directory.
    - `spacemacs/helm-file-do-<TOOL>`: Search within current file/buffer.
    - `spacemacs/helm-project-do-<TOOL>`: Search within current project.
    - For each search SCOPE, i.e. one of BUFFERS, FILES, FILE, PROJECT, there is
      a variant that choose the first search tool found,
      `spacemacs/helm-<SCOPE>-smart-do-search`.
    - For every combinations of search TOOL and search SCOPE, there is a variant
      that uses the symbol at point or current selection as default input,
      `spacemacs/helm-<SCOPE>-<TOOL>-region-or-symbol`.
  - This commit renamed those functions from
    `spacemacs/helm-<SCOPE>-<smart-do-search|do-TOOL>[-with-region-or-symbol]`
    to `spacemacs/helm-<smart-search|TOOL>-<SCOPE>[-with-input]`.
    - The new names are shorter than the old ones, and uses `SVO` order instead
      of the less common `SOV` order, where subject is `helm`, verb is the
      search tool, and object is the scope.
  - This commit also added missing functionalities, so that feature parity is
    achieved for all supported search tools
    - Currently only `ag` and `grep` supports searching within the current
      file/buffer. This commits adds this functionality to all search tools.
  - The new search functions behaves **differently** from the old ones:
    - `spacemacs/helm-<smart-search|TOOL>-buffer[-with-input]`: Search in
      current buffer. When prefixed with universal argument, search within all
      opened buffers.
    - `spacemacs/helm-<smart-search|TOOL>-dir[-with-input]`: Search in current
      directory. When prefixed with universal argument, prompt for a directory
      and search within it.
    - `spacemacs/helm-<smart-search|TOOL>-project[-with-input]`: Search in
      current project. When prefixed with universal argument, prompt for string
      input as additional command-line arguments passed to the search tool.
    - A comparison of the interfaces is attached in the end of this commit
      message.
  - ALL aforementioned search interfaces are combined into a monolithic
    dispatcher function, `spacemacs-helm-search-dispatcher`.
    - The dispatcher itself is byte compiled. Key bindings are registered for
      thin wrappers of the dispatcher.
  - Proper error handling is implemented (the first two are new additions from
    this commit):
    - Reports an error if no search tool is found when calling one of the
      smart search function.
    - Reports an error if the search tool is not found when calling one of the
      search functions which has a specified search tool.
    - Reports an error if the current directory is not a project directory
      when calling one of the project-scope search function.
- Added filtering when adding per-search-tool key bindings.
  - Currently regardless of the value of `dotspacemacs-search-tools`, all
    supported search tools have their search functions declared and key bindings
    registered.
  - This commit still declares all search functions for all supported search
    tools, but only register key bindings for those listed in
    `dotspacemacs-search-tools`. E.g., if its value is `'(rg)`, then only those
    under `SPC s r` are added.
  - In other words, users can still invoke all supported search functions via
    `M-x` interface, but only those of the search tools they specified would
    have key bindings.
  - Since this commit also achieved feature parity across all supported search
    tools, in the future we might not add any search-tool-specific at all.

Comparison of the old and new interfaces, using smart search without default
input as examples:

|Scope                        |New Interface|Old Interface                     |
|-----------------------------|-------------|----------------------------------|
|All opened buffers           |`SPC s b`    |`SPC s b`                         |
|Current buffer               |`C-u SPC s b`|Only available for `ag` and `grep`|
|Current directory            |`SPC s d`    |`SPC s d`                         |
|Specified directory          |`C-u SPC s d`|`SPC s f`                         |
|Current project              |`SPC s p`    |`SPC s p`                         |
|Current project w/ extra args|`C-u SPC s p`|Not supported                     |

Signed-off-by: Lucius Hu <lebensterben@users.noreply.github.com>

### DIFF
--- a/layers/+completion/helm/funcs.el
+++ b/layers/+completion/helm/funcs.el
@@ -7,12 +7,17 @@
 ;;
 ;; This file is not part of GNU Emacs.
 ;;
-;;; License: GPLv3
+;; License: GPLv3
 
-
+;;; Commentary:
 
-(defvar spacemacs--helm-popwin-mode nil
-  "Temp variable to store `popwin-mode''s value.")
+;;; Code:
+
+(require 'core-funcs)
+(require 'dash)
+(require 'dired)
+
+;;;; Helper functions
 
 (defun spacemacs//helm-cleanup ()
   "Cleanup some helm related states when quitting."
@@ -32,15 +37,14 @@
     (popwin-mode -1)))
 
 (defun spacemacs//helm-restore-display ()
+  "Not documented."
   ;; we must enable popwin-mode first then restore `display-buffer-alist'
   ;; Otherwise, popwin keeps adding up its own buffers to
   ;; `display-buffer-alist' and could slow down Emacs as the list grows
-  (when spacemacs--helm-popwin-mode
-    (popwin-mode))
+  (when spacemacs--helm-popwin-mode (popwin-mode))
   (setq display-buffer-alist spacemacs-display-buffer-alist))
 
-
-;; REPLs integration
+;;;; REPLs Integration
 
 (defun helm-available-repls ()
   "Show all the repls available."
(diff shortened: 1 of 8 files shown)


lebensterben:
I will add description to CHANGELOG.develop and layer.org if the above changes are accepted.

lebensterben:
I've updated it to address feedback from the community (discussion from gitter channel).
The key bindings for search in current buffer and all opened buffers are switched:

    |Scope                        |New Interface|Old Interface                     |
    |-----------------------------|-------------|----------------------------------|
    |All opened buffers           |`SPC s b`    |`SPC s b`                         |
    |Current buffer               |`C-u SPC s b`|Only available for `ag` and `grep`|
    |Current directory            |`SPC s d`    |`SPC s d`                         |
    |Specified directory          |`C-u SPC s d`|`SPC s f`                         |
    |Current project              |`SPC s p`    |`SPC s p`                         |
    |Current project w/ extra args|`C-u SPC s p`|Not supported                     |

henriquenj:
Why do we now require `C-u` for searching the specified directory? I use `SPC s f` a lot and now it will be extra typing for me. As far as I am aware there's nothing else binded to `SPC s f`. Of course I don't mind getting used to new binding if I see a point for it.

I use ivy but I am assuming the changes will ported to ivy to keep consistency.

lebensterben:
>Why do we now require `C-u` for searching the specified directory? I use `SPC s f` a lot and now it will be extra typing for me. As far as I am aware there's nothing else binded to `SPC s f`. Of course I don't mind getting used to new binding if I see a point for it. ...

@henriquenj 
I can just switch all prefixed and non-prefixed bindings.
The idea is the difference between SPC s f and SPC s d are superficial, but the underlying functions have confusing names (search directory SPC s d and search files SPC s f)

henriquenj:
Thanks for the explanation. I think I will try with the new bindings and see how it goes. They are similar indeed.

smile13241324:
@henriquenj how is it going with the new set of changes? If you did not encounter major issues I would like to review and merge this PR. Would be great if you could give me some hints how trying out this branch did work out for you (saves me the time for testing :))

henriquenj:
I applied the patch on top of ba8f5414c9b316225063d4b2f6fd4437934f53d2 on a new clone (there's a merge conflict after that, not sure why it is not flagged on github). Since I don't use helm daily, I suggest someone else with more experience to also review it. I found some problems so I suggest more testing before we can merge.

Every time I open spacemacs now I have a buffer called `*Compile-Log*` with these warnings:

```
Warning (bytecomp): reference to free variable ‘helm-ff-transformer-show-only-basename’
Warning (bytecomp): reference to free variable ‘projectile-known-projects’
Warning (bytecomp): reference to free variable ‘helm-projectile-fuzzy-match’
Warning (bytecomp): reference to free variable ‘helm-read-file-name-mode-line-string’
Warning (bytecomp): reference to free variable ‘spacemacs-helm-rg-max-column-number’
Warning (bytecomp): reference to free variable ‘helm-ff-history’
```
The buffer is always created when I start emacs again. The messages appear even if I do not install the helm layer (like if I am using ivy).

`SPC s d` seems to be ok, but I can't get `SPC u SPC s d` to work. When I use it after selecting the directory and start typing I get this error:
```
Error running timer: (error "Candidates function ‘#[0 \\301 \\210\\302\\300!\\207 [(/home/henrique/Desktop/emacs-testing-home/.emacs.d) helm-ag--do-ag-set-command helm-ag--do-ag-candidate-process] 2]’ should run a process")
```
I tried to get a backtrace using debug-on-error but if I do that emacs become irresponsive and I couldn't copy the backtrace :(.

Also the bindings are not the same in ivy. Should we change ivy as well in the same PR? In my opinion they should be the same but of course that depends if someone is willing to work on the ivy implementation.

lebensterben:
I have made some changes but I haven't finished them yet since I'm caught up by another project.

I will revisit this PR shortly.

smile13241324:
The general idea is approved @lebensterben, still we need proper testing before this can be merged, I will give it a go this evening and see how it behaves with helm. If this is all ok, I think we should try to get the same level of integration for ivy but for this I would prefer to have a separate PR.

layers/+completion/helm/funcs.el
```diff
@@ -7,12 +7,17 @@
...

+;;; Commentary:
 
-(defvar spacemacs--helm-popwin-mode nil
-  "Temp variable to store `popwin-mode''s value.")
+;;; Code:
+
+(require 'core-funcs)
+(require 'dash)
+(require 'dired)
```
smile13241324, line 18: This is dangerous, in my case "dash" is not installed and subsequently Spacemacs stops booting.

layers/+completion/helm/funcs.el
```diff
@@ -55,465 +59,266 @@
...
+                                       (ask-dir-name)
+                                     default-directory)))
+                      ("project" (list #'helm-do-ag
+                                       (if prefix-p
+                                           (ask-project-dir)
+                                         (current-project-root)))))))
+         (helm-ag-wrapper func default-input-p args)))
+      (_ (user-error "Not a valid search tool %s" search-tool)))))
+
+(byte-compile 'spacemacs//helm-search-dispatcher)
```
smile13241324, line 233: This should be removed, otherwise you compile each time the file is loaded. Instead we should rely on the ability to auto-compile all spacemacs files, this is a setting in your dotfile.

layers/+completion/helm/funcs.el
```diff
@@ -523,7 +328,11 @@ Removes the automatic guessing of the initial value based on thing at point. "
     (set-text-properties 0 (length input) nil input)
     (helm-find-files-1 input)))
 
- ;; Key bindings
+(byte-compile 'spacemacs/helm-find-files)
```
smile13241324, line 331: This should be removed, otherwise you compile each time the file is loaded. Instead we should rely on the ability to auto-compile all spacemacs files, this is a setting in your dotfile.

layers/+completion/helm/packages.el
```diff
@@ -216,57 +216,37 @@
...
-        "srp" 'spacemacs/helm-project-do-rg
-        "srP" 'spacemacs/helm-project-do-rg-region-or-symbol
-        "stp" 'spacemacs/helm-project-do-pt
-        "stP" 'spacemacs/helm-project-do-pt-region-or-symbol))
+        ;; buffer(s) scope
+        "sb"  'spacemacs/helm-smart-search-buffer
+        "sB"  'spacemacs/helm-smart-search-buffer-with-input
+        ;; directory scope
+        "sd"  'spacemacs/helm-smart-search-dir
+        "sD"  'spacemacs/helm-smart-search-dir-with-input
```
smile13241324, line 224: These functions cannot be found, I think this is a direct side effect of the not executed "--each" block but nevertheless please double check.

layers/+completion/helm/funcs.el
```diff
@@ -55,465 +59,266 @@
...
+(--each
+    (-table-flat
+     'list
+     '(spacemacs||helm-search-command)
+     '(("buffer" "search in a specified buffers") ; spec, i.e. (scope doc)
+       ("dir" "search in a specified directory")
+       ("project" "search in a specified project"))
+     '("rg" "ag" "pt" "ack" "grep" nil)           ; tool
+     '(t nil))                                    ; default-input-p
+  (eval it))
```
smile13241324, line 277: This must be refactored not to require dash.

smile13241324 (changes requested):
Just finished my first round of testing, primarily I hit an issue with the dependency on dash which is not directly included in Spacemacs.

smile13241324:

PR #14262, opened 2021-01-05
fix page-break-lines-mode alias definition

dalanicolai:
This definition requires 3 arguments. Having only two can cause Emacs a
wrong-number-of-arguments error (was reported on gitter). So this PR just
 adds the date when the orginal functions were removed and the alias was
 created. Just a minor fix.

### DIFF
--- a/core/libs/page-break-lines.el
+++ b/core/libs/page-break-lines.el
@@ -106,7 +106,7 @@ horizontal line of `page-break-lines-char' characters."
   (page-break-lines--update-display-tables))
 
 ;;;###autoload
-(define-obsolete-function-alias 'turn-on-page-break-lines-mode 'page-break-lines-mode)
+(define-obsolete-function-alias 'turn-on-page-break-lines-mode 'page-break-lines-mode "2018-07-24")
 
 (dolist (hook '(window-configuration-change-hook
                 window-size-change-functions


lebensterben:
We probably should remove this line entirely, see https://github.com/purcell/page-break-lines/commit/69caea070379f3324c530e96e06625c3cd097cb9

lebensterben:
The larger problem is, we need a two-stage loading process, where in the first stage we load package that constitutes the core library instead of maintaining their source codes manually...

dalanicolai:
Completely removing it is ven better. I just was not sure if the old name was still used somewhere...

lebensterben:
it seems that this file is indeed just a copy of the upstream.
in the past we have manually updated too. for example, https://github.com/syl20bnr/spacemacs/commit/fcb79bf55b6cc6b1dc2f96d08cfe1ee7d74ebf16#diff-675c7e805802d3c8c184084adac902fa2bc0aa22ca4309a5e81786bfbbebbe59

smile13241324:
Yes these core dependencies are an open architectural issue in spacemacs, yyoncho did some work to have an auto-update in place for these packages (auto-update bot) however it looks like we are not longer compatible with some of the libraries main branch.

Current state is we need to go through those dependencies and see to which we are not longer compatible and either change our side or push PRs to upstream.

For now this is a valid hotfix until we have a proper solution.

PR #14263, opened 2021-01-05
[markdown] fontify code blocks natively: on by default

Kazark:
I realized this is on by default in Spacemacs for Org. Arguably Spacemacs should have a somewhat consistent default experience between Org and Markdown.

If anyone disagrees, it's no big deal to me. I wasn't even going to suggest upstreaming it until I saw this in `layers/+emacs/org/packages.el`:

```emacs-lisp
(defun org/init-org ()
  (use-package org
    :defer ...
    :commands ...
    :init
    (progn
      ...
      (setq ...
            org-src-fontify-natively t
            ...
```

Though it would be nice for look & feel of Spacemacs out of the box.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -134,7 +134,7 @@ the [[file:CHANGELOG.org][CHANGELOG.org]] file.
   (thanks to Cormac Cannon)
 - Removed cquery.
 ***** EPUB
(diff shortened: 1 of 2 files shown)


smile13241324 (approved):
Thats an absolutely sensible change @Kazark, Spacemacs should be consistent between org and markdown. Lets merge it.

PR #14264, opened 2021-01-05
Add org-journal-open-current-journal-file keybinding

dalanicolai:
This is such an elementary org-journal command that it should have a
keybinding.

### DIFF
--- a/layers/+emacs/org/README.org
+++ b/layers/+emacs/org/README.org
@@ -781,6 +781,7 @@ org-present must be activated explicitly by typing: ~SPC SPC org-present~
 
 | Key binding         | Description                                     |
 |---------------------+-------------------------------------------------|
+| ~SPC a o j f~       | Visit journal file                              |
 | ~SPC a o j j~       | New journal entry                               |
 | ~SPC u SPC a o j j~ | Open today's journal without adding a new entry |
 | ~SPC a o j s~       | Search journal entries                          |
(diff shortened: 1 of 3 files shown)


layers/+emacs/org/README.org
```diff
@@ -781,6 +781,7 @@ org-present must be activated explicitly by typing: ~SPC SPC org-present~
 
 | Key binding         | Description                                     |
 |---------------------+-------------------------------------------------|
+| ~SPC a o f~         | Visit journal file                              |
```
Kazark, line 784: Isn't the bind `SPC a o j f`, per the code?

Kazark (changes requested):

layers/+emacs/org/README.org
dalanicolai: Sorry, what do you mean with 'per the code'?

dalanicolai: Ah yes... that's sloppy

Kazark (approved):
LGTM (but I'm just a contributor; I don't have merge permissions)

dalanicolai:
Anyway thanks for notifying

smile13241324 (approved):
Good binding, will be merged soon.

smile13241324:
@Kazark thanks for testing, actually its a great help if someone else from the community already did some testing and gives a short summary in the PR, by this we can focus on the technical part and do not need to invest time into testing. So your help is very much appreciated.

PR #14266, opened 2021-01-06
Add variables for configuration of startup of the @Org perspective.

dbankmann:
Added two variables 
`org-persp-startup-org-file` and `org-persp-startup-with-agenda` initialized to `nil` with no changes to the current behavior.
These variables can be set by the user to initialize the @Org perspective with a specified file and/or a (custom) agenda view.

Docs will be updated, once the PR is accepted.

### DIFF
--- a/layers/+emacs/org/README.org
+++ b/layers/+emacs/org/README.org
@@ -27,6 +27,7 @@
   - [[#epub-support][Epub support]]
   - [[#jira-support][Jira support]]
   - [[#verb-support][Verb support]]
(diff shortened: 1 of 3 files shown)


smile13241324 (changes requested):
I like your change, its definitely an improvement to have the default agenda file properly configurable in the layer especially if we define a custom layout anyway.

Please add the missing README changes here, I will then give it another review and merge it if ok.

dbankmann:
Done.

smile13241324 (approved):
Looks good.

PR #12417, opened 2019-06-09
[core] Environment variable caching fixes

Miciah:
Fix several problems with environment variable caching.

First, if a shell printed extra output besides the command output, `spacemacs//init-spacemacs-env` included the extra output in `.spacemacs.env`.  For example, a login shell could print a motd, which would be erroneously included in `.spacemacs.env`.  To avoid this, `spacemacs//init-spacemacs-env` now redirects the command output to a temporary file and then reads the file.

Second, `spacemacs//init-spacemacs-env` sorted lines using the entire "name=value" string for each line whereas the correct behavior is to sort only on the "name" part.  To fix this, `spacemacs//init-spacemacs-env` now uses `sort-regexp-fields` with an appropriate regexp to match the "name" part and ignore the "value" part.

Third, although `load-env-vars-set-env` added all `PATH` settings in `.spacemacs.env` to `exec-path`, it set `PATH` to the first `PATH` setting in `.spacemacs.env`.  Now `load-env-vars-set-env` sets `PATH` from `exec-path` so that it will reflect that combined value.

Fourth, `load-env-vars-set-env` set `exec-path` and `PATH` without consideration to the operating system's directory separator.  Now `load-env-vars-set-env` converts backslashes to forward slashes when adding path entries to `exec-path` and converts forward slashes to backslashes when setting `PATH` on platforms that use backslashes.

Fifth, `load-env-vars-set-env` now normalizes `exec-path` by deleting trailing slashes from path entries, converting drive letters in path entries to lower-case, and deleting duplicate path entries.

Sixth, Spacemacs no longer uses `exec-path-from-shell`, but a changelog entry said that it did.  This commit corrects the changelog entry, deletes another outdated entry, and merges two entries that covered the same functionality.

Finally, this commit corrects several typos and tries to improve wording in several docstrings, messages, and comments.

* `CHANGELOG.develop`: Delete mention of `exec-path-from-shell`, which Spacemacs no longer uses by default.  Delete mention of `dotspacemacs-import-env-vars-from-shell` and `dotspacemacs-import-env-vars-shell-file-name`, which no longer exist.  Merge entries about synchronizing environment variables.
* `core/core-env.el` (`spacemacs-ignored-environment-variables`): Fix typos and mention that the strings are regexps that are matched against the names of environment variables.
[...]

PR #14274, opened 2021-01-08
[bot] Auto-update

emacspace:

### DIFF
--- a/core/libs/dash.el
+++ b/core/libs/dash.el
@@ -1,12 +1,14 @@
 ;;; dash.el --- A modern list library for Emacs  -*- lexical-binding: t -*-
 
-;; Copyright (C) 2012-2016 Free Software Foundation, Inc.
+;; Copyright (C) 2012-2021 Free Software Foundation, Inc.
 
 ;; Author: Magnar Sveen <magnars@gmail.com>
 ;; Version: 2.17.0
-;; Keywords: lists
+;; Package-Requires: ((emacs "24"))
+;; Keywords: extensions, lisp
+;; Homepage: https://github.com/magnars/dash.el
 
-;; This program is free software; you can redistribute it and/or modify
+;; This program is free software: you can redistribute it and/or modify
 ;; it under the terms of the GNU General Public License as published by
 ;; the Free Software Foundation, either version 3 of the License, or
 ;; (at your option) any later version.
@@ -17,19 +19,13 @@
 ;; GNU General Public License for more details.
 
 ;; You should have received a copy of the GNU General Public License
-;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+;; along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 ;;; Commentary:
 
-;; A modern list api for Emacs.
+;; A modern list API for Emacs.
 ;;
-;; See documentation on https://github.com/magnars/dash.el#functions
-;;
-;; **Please note** The lexical binding in this file is not utilised at the
-;; moment. We will take full advantage of lexical binding in an upcoming 3.0
-;; release of Dash. In the meantime, we've added the pragma to avoid a bug that
-;; you can read more about in https://github.com/magnars/dash.el/issues/130.
-;;
+;; See its overview at https://github.com/magnars/dash.el#functions.
 
 ;;; Code:
 
@@ -40,21 +36,10 @@
     (require 'cl)))
 
 (defgroup dash ()
-  "Customize group for dash.el"
+  "Customize group for Dash, a modern list library."
+  :group 'extensions
   :group 'lisp
   :prefix "dash-")
-
-(defun dash--enable-fontlock (symbol value)
-  (when value
-    (dash-enable-font-lock))
-  (set-default symbol value))
-
-(defcustom dash-enable-fontlock nil
-  "If non-nil, enable fontification of dash functions, macros and
-special values."
-  :type 'boolean
-  :set 'dash--enable-fontlock
-  :group 'dash)
 
 (defmacro !cons (car cdr)
   "Destructive: Set CDR to the cons of CAR and CDR."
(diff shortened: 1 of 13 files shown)


JAremko:
@duianto  @smile13241324  looks like latest version of some built-in libraries aren't compatible with Spacemacs. We need to figure out which ones. :thinking: 

This list handles auto update https://github.com/syl20bnr/spacemacs/blob/develop/.ci/built_in_manifest

Worst case scenario, we can apply custom patch after each auto-update.

smile13241324:
Jup, lets see why they are not compatible maybe we can change our side to be working with their main branch or at least factor out the spacemacs specific changes....

duianto:
There seems to only be an issue in Windows 10, with: `core/libs/load-env-vars.el`
It's fine in Manjaro.

#### Cause
The upstream function:
https://github.com/diasjorge/emacs-load-env-vars/blob/3808520efaf9492033f6e11a9bffd68eabf02a0f/load-env-vars.el#L74
```elisp
(defun load-env-vars-set-env (env-vars)
  "Set envariable variables from key value lists from ENV-VARS."
  (dolist (element env-vars)
    (let ((key (car element)) (value (cadr element)))
      (setenv key value))))
```

Was modified in this PR:
[[core] Environment variable caching fixes by Miciah #12417](https://github.com/syl20bnr/spacemacs/pull/12417)

https://github.com/syl20bnr/spacemacs/blob/bd1631bdce8625f4f56efc10c4bf342e60080e0e/core/libs/load-env-vars.el#L77-L94

With the upstream function, in Windows 10.
When I try to update the packages, then it shows:
>condition-case: Failed to checkout ‘font-lock+’: ‘Command ’(c:/tools/cygwin/bin/timeout.exe -k 60 600 git remote show -n origin)’ exited with non-zero status 127: /usr/bin/timeout: failed to run command 'git': No such file or directory

JAremko:
```
 FAILED  test-declare-layers--base-layer-is-third-for-base-distribution
 FAILED  test-declare-layers--bootstrap-layer-always-first
 FAILED  test-declare-layers--defaults-layer-is-second-for-base-distribution
```

I doubt that it's env related. Could be ht.el :thinking:

JAremko:
I'll update packages one by one and see when tests break.

JAremko:
Seems like it's `mocker.el` so I exclude it and env thingy from the auto-update list and we'll see if it works.

JAremko:
Looks like it helped.

But now we have 2 problems: 

- windows-latest/snapshot seems to fail now - it's not that bad since we don't guarantee that Spacemacs works with  snapshot builds.  But it would be helpful if we can figure out what's broken and report it upstream.

- We need to figure out what's up with the mocker.  It doesn't look that different https://www.diffchecker.com/4wBnjYbZ but those 3 tests  are failing with the latest version :thinking:

JAremko:
Ok mocker should work now https://github.com/syl20bnr/spacemacs/pull/14280

JAremko:
@TheBB  Looks like Evil might have compatibility issue with the Emacs snapshots (on windows for some reason)

(define-obsolete-function-alias 'evil-ex-shell-command-completion-at-point 'comint-completion-at-point)  


``` elisp
 #f(compiled-function (obsolete-name current-name when &optional docstring) "Set OBSOLETE-NAME's function definition to CURRENT-NAME and mark it obsolete.\n\n(define-obsolete-function-alias \\='old-fun \\='new-fun \"22.1\" \"old-fun's doc.\")\n\nis equivalent to the following two lines of code:\n\n(defalias \\='old-fun \\='new-fun \"old-fun's doc.\")\n(make-obsolete \\='old-fun \\='new-fun \"22.1\")\n\nWHEN should be a string indicating when the function was first\nmade obsolete, for example a date or a release number.\n\nSee the docstrings of `defalias' and `make-obsolete' for more details." #<bytecode -0xc3f2ab0a40cd5b1>)('evil-ex-shell-command-completion-at-point 'comint-completion-at-point)
  (define-obsolete-function-alias 'evil-ex-shell-command-completion-at-point 'comint-completion-at-point)
  eval-buffer(#<buffer  *load*-480978> nil "c:/Users/runneradmin/.emacs.d/elpa/28.0/evil-20201..." nil t)  ; Reading at buffer position 22912
  load-with-code-conversion("c:/Users/runneradmin/.emacs.d/elpa/28.0/evil-20201..." "c:/Users/runneradmin/.emacs.d/elpa/28.0/evil-20201..." nil t)
```

`(define-obsolete-function-alias 'evil-ex-shell-command-completion-at-point 'comint-completion-at-point)` is a suspect.

We had similar issue with another package  https://github.com/syl20bnr/spacemacs/pull/14262/files#diff-675c7e805802d3c8c184084adac902fa2bc0aa22ca4309a5e81786bfbbebbe59R109

robbyoconnor:
Snapshot has been failing for a while for me

smile13241324:
Is a failing snapshot build really an issue for us? We cannot guarantee that Spacemacs works on any of them anyway. In my opinion we should only worry about supporting finished releases.

Can I help with this ticket or do you want to run some more integration tests against it @JAremko?

JAremko:
@smile13241324  Snapshot tests are optional :+1:

JAremko:
Also tests seems to fix themselves.  Most likely someone realized that it's pretty stupid to break Emacs install for making last argument to obsolete annotation allays required

robbyoconnor:
This scares me -- how is this happening?

robbyoconnor:
> Is a failing snapshot build really an issue for us? We cannot guarantee that Spacemacs works on any of them anyway. In my opinion we should only worry about supporting finished releases.

@smile13241324 -- I never ever expect spacemacs to always work on Emacs Snapshot -- things break and I usually expect it from time to time -- I am after all using unreleased software that is built from the repository itself.

JAremko:
> This scares me -- how is this happening?

Well.. I can make an uneducated guess that Windows snapshot got built with different flags or something that caused  this. I don't think it was intended behavior.

JAremko:
Actually it seems to be fixed in the package https://github.com/emacs-evil/evil/commit/19cc5f8eef8bfffdec8082b604c7129782acb332  a while ago.  I guess we got the new version only now :thinking:

robbyoconnor:
> Well.. I can make an uneducated guess that Windows snapshot got built with different flags or something that caused this. I don't think it was intended behavior.

@JAremko I never used Emacs on windows lol -- I've used Mint for as long as I've been using emacs.

PR #14278, opened 2021-01-08
Add toggles for specific persistent which-key keymaps

dalanicolai:
Sometimes it is handy to show keymaps persistently with which-key. For example,
to show navigation keys in Info-mode, gnus, eww etc. This PR implements
handy toggles and documentation for that.

Of course this is just a quick way to implement some kind of alternative hydra transient states.

Also I am always happy about feedback/improvements/discussions, my PR's are merely proposals.


smile13241324 (approved):
I think this is a good idea, especially for new users which do not yet remember the key bindings a persistent which key bar may just be what helps them learn the key bindings. Especially as in default config which-key only opens after a delay.

PR #14284, opened 2021-01-10
fixed package name from org-super-star to org-superstar

emacs18:
This simply fixes what seems to be a typo in the package name where extra dash is removed between "super" and "star".

The correct name of the package is org-superstar, not org-super-star.
For example MELPA lists only org-superstar, not org-super-star.
I'm puzzled why this typo was not noticed before.

### DIFF
--- a/layers/+spacemacs/spacemacs-org/packages.el
+++ b/layers/+spacemacs/spacemacs-org/packages.el
@@ -64,7 +64,7 @@
       (evil-define-key 'normal org-mode-map (kbd "RET") 'org-open-at-point)))))
 
 (defun spacemacs-org/init-org-superstar ()
-  (use-package org-super-star
+  (use-package org-superstar
     :defer t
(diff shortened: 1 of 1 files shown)


smile13241324 (approved):
Another typo? Strange we have a lot of these lately. Maybe there was a historic version using this? Anyway better have it fixed now.

Kazark:
I thought superstar was working. I fail to understand how this wouldn't just make it completely broken.

Kazark:
Yeah, perhaps like Maximillian said, it was historically named that? I use Org every day and was under the impression that this was where I was getting my schnazzy bullets (because my schnazzy bullet styling did change when we switched to superstar).

emacs18:
This patch was submitted because it was required to get things to work when I used straight.el package manager rather than package.el.
I am still puzzled as to how package.el seemed to have worked despite the typo in package name.

PR #14287, opened 2021-01-11
Implement C-(S)-tab buffer switch in Ivy layer

dalanicolai:
These keybindings offer a nearly equivalent experience to the common C-tab buffer switching behavior. To me this is still the superior way for switching between previous buffers (more than only one like `SPC-tab`). Also, I think many other users like this behavior, see:
https://stackoverflow.com/questions/4146574/how-to-properly-configure-ctrl-tab-in-emacs
https://stackoverflow.com/questions/4133181/emacs-tab-between-buffers
https://www.reddit.com/r/emacs/comments/9z4faf/ctrltab_buffer_switching/
 (the only downside is that new users maybe will not use `SPC b b` when they should, but I guess they will).

I am not sure if the keystring used in the last defined keybinding is universal, but this one is working for my keyboard.

As far as I know these bindings do not interfere with any other bindings (i.e. they were all still undefined).

It is just a proposal... no need to merge it if you think it does not fit.

The table was outlined using `org-cycle` and `emacs -Q`.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2354,6 +2354,7 @@ Other:
   - Bound =SPC s l= to =ivy-resume= (thanks to Wieland Hoffmann)
   - Rebind =C-k= to =C-M-k= in =ivy-reverse-i-search= (=C-r= at a prompt)
     (thanks to duianto)
+  - Add =C-(S)-tab= keybindings for buffer switching (thanks to Daniel Nicolai)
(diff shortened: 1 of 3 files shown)


lebensterben:
I think this is a nice addition.

smile13241324 (approved):
Yep this should be merged, I think we have the belated fix for helm already merged.

smile13241324:
Merged manually.

PR #14288, opened 2021-01-11
[helpful] Add missing global key bindings

sr105:
- In addition to the spacemacs leader keys, also replace the default emacs
keybindings under `C-h`

- Fix a documentation copy-paste mistake for `SPC h d v`


sr105:
Should I have done it this way instead? I used `global-set-key` because I saw it in another layer.

```emacs-lisp
    :bind
    ([remap describe-key] . helpful-key)
    ([remap describe-function] . helpful-callable)
    ([remap describe-variable] . helpful-variable)
```

sr105:
The remap bindings work, but they don't update the function name in which-key. It still says `describe-key`, for example, but it really calls `helpful-key` after `C-h k`.

smile13241324 (approved):
Looks ok, thanks for adding

PR #14289, opened 2021-01-11
[scala] remove long-deprecated Ensime integration

Kazark:
According to its website, the Ensime project has been shut down since 2019. We have been supporting both Ensime and Metals for [a year and a half](https://github.com/syl20bnr/spacemacs/pull/12234#issuecomment-524916394). Ensime's GitHub repositories, including its Emacs integration, are archived. Meanwhile, Metals has been developed actively, regularly releasing new versions. It has stabilized significantly, and supports an increasingly full set of features. Scala 3 is just around the corner, and the community is poised to make the transition smoothly and relatively quickly. Metals supports it already, whereas of course Ensime does not and never will. In fact, Scala 2 has had several import minor versions released since Ensime died. Now that it's 2021, it's time to cut the baggage we are carrying around for Ensime.


layers/+lang/scala/layers.el
```diff
@@ -9,7 +9,7 @@
 ;;
 ;;; License: GPLv3
 
-(configuration-layer/declare-layer-dependencies '(java))
+(configuration-layer/declare-layer-dependencies '(java lsp))
```
smile13241324, line 12: This is not necessary, `dap` layer already includes lsp

layers/+lang/scala/config.el
```diff
@@ -11,24 +11,21 @@
...
+(defconst scala-backends '(scala-metals)
   "Backend server implementation to enable advanced IDE language features")
 
-(defvar scala-backend 'scala-ensime
+(defvar scala-backend 'scala-metals
   "Backend used to trigger IDE language features.
-`scala-ensime' or `scala-metals' are currently supported")
+Only `scala-metals' is currently supported.")
 
 (defvar scala-auto-start-backend nil
```
smile13241324, line 27: This variable must be removed too, it was only used by ensime, lsp servers are always auto-started.

smile13241324 (changes requested):
I like cutting this old piece of code however I have found some smaller issues which must be fixed before we can merge this.

* Some variables are now obsolete and should be removed also
* As far as I see there is no REPL support available anymore, I don't think metals has one, so the feature must be removed
* Have you made a test on your local config that metals still correctly loads?

layers/+lang/scala/config.el
Kazark: That's why I love deleting. I always learn something. 😁

layers/+lang/scala/layers.el
Kazark: It was necessary for me. I did not get either `dap` or `lsp` keybindings in my major mode map for Scala until I added this.

Kazark:
>  Have you made a test on your local config that Metals still correctly loads?

Yes, and Scala is what I write at work so I will notice fast if something is wrong.

Kazark:
@smile13241324 thanks for your corrections. I have rebased and made the revisions you requested, with the exception of removing `lsp` as a dependency, which was necessary to get this working for me. Let me know if I still missed something. 😁

Kazark:
Also, I have retested and my Scala environment is still working correctly.

smile13241324:
I will have a look

layers/+lang/scala/layers.el
smile13241324: Strange, anyway requiring it here too does not do any harm, so lets keep it.

smile13241324 (approved):

smile13241324:
Now its cool, lets merge it :)

hxx0215:
Since ensime seems relaunched. Would you consider to recover it?
[ensime-the-next-generation](https://contributors.scala-lang.org/t/ensime-the-next-generation/5474)

PR #14290, opened 2021-01-11
[scala] provide an easy way to a small SBT buffer at frame bottom

Kazark:
This feels like the natural place to put the SBT buffer, though we could also extend this with other values in the future. I defaulted this behavior to off in order to not impact folks' current setup.

The window needs to be dedicated because of its strange shape. Most other buffers (e.g. the Magit status buffer) will not look pleasant in it.


smile13241324 (approved):
Looks good

PR #14291, opened 2021-01-11
Add Emacs Application Framework (EAF) layer

dalanicolai:
Here is a layer for the Emacs Application framework (EAF). I think it is a very nice package, mainly because it adds a full-fledged browser to Spacemacs but it adds many other nice features too.

The layer modifies some keybindings, so that it nicely supports a Spacemacs workflow.

A lot more can be added but I think the fundaments are here.

I mainly create this PR to get feedback, although it might be ready for merging already.
It comes with a README containing the necessary documentation.
Any improvements are welcome.


layers/+tools/eaf/packages.el
```diff
@@ -0,0 +1,219 @@
...
+                    :files ("*")))))
+
+(defun eaf/init-eaf ()
+  (use-package eaf
+    ;; (eaf-bind-key clear_focus "C-f" eaf-browser-keybinding)
+    ;; (define-key eaf-mode-map (kbd "<escape>") 'eaf-proxy-clear_focus)
+    :defer t
+
+    :init
+    (spacemacs/declare-prefix "aa" "application-framework")
```
smile13241324, line 25: I think these bindings should rather go under application/tools/web services, there should already be a prefix bound for this in the web search layer.

layers/+tools/eaf/packages.el
```diff
@@ -0,0 +1,219 @@
...
+                  )))
+
+    (define-key key-translation-map (kbd ",")
+      (lambda (prompt)
+        (if (derived-mode-p 'eaf-mode)
+            (if (eaf-call "call_function" eaf--buffer-id "is_focus")
+                (kbd ",")
+              (kbd "C-,")))))))
+
+;;; packages.el ends here
```
smile13241324, line 219: No package end declaration in layer files please

layers/+tools/eaf/README.org
```diff
@@ -0,0 +1,89 @@
...
+  based distributions. Users of Arch (based) distributions can run the script
+  manually in their terminal. If the script does not work for you then [[https://github.com/manateelazycat/emacs-application-framework#dependency-list][install
+  the required dependencies manually]].
+
+* Usage
+
+*** Configuring keybindings
+  This layer implements some tricks that are required to make EAF behave as
+  desired for Spacemacs. Therefore it is recommended to keep the default
+  (major-mode-)leader keybindings. Also due to these tricks, ~SPC SPC~ does not
```
smile13241324, line 50: Can't we get this into the layer already? I have to say adding a layer which does not support leader key mechanisms would behave unexpectedly for most users I assume.

smile13241324 (changes requested):
Just double checked your PR and found some issues you need to have a look at before this can be merged. 

Btw: You don't have to change LAYERS.org this is auto generated anyway.

layers/+tools/eaf/README.org
dalanicolai: So I think my text in the documentation was not clear enough. I hope my change improved its meaning.
The problem is that EAF keybindings interfere with evil keybindings. Therefore the EAF developers created an eaf-evil.el package, but I found it not compatible enough with Spacemacs. So still using their eaf-evil.el package, I added some modifications to "emulate" Spacemacs behavior. So for the user it works exactly how they expect from Spacemacs (except for the `SPC SPC` but for that they can just use `M-x`. EAF comes with custom eaf keybinding functions that somehow commincate the keybindings to the python part of EAF. You can read more about my modifications [here](https://github.com/manateelazycat/emacs-application-framework/pull/500) and [here](https://github.com/manateelazycat/emacs-application-framework/issues/498). So making EAF work the Spacemacs way is the strong point of this layer. Getting the keybindings right is not trivial.

dalanicolai:
Thanks for your quick and useful feedback. I have implemented the changes you suggested/requested. I mentioned already that a lot can be added. So by implementing the layer, users can use it and provide feedback. I think the layer is working good enough for implementation (its browser really is quite nice).

layers/+tools/eaf/README.org
```diff
@@ -0,0 +1,89 @@
...
+  - Image viewer
+  - See [[https://github.com/manateelazycat/emacs-application-framework#launch-eaf-applications][EAF documentation]] for many more features
+
+* Install
+  To use this configuration layer, add it to your =~/.spacemacs=. You will need to
+  add =eaf= to the existing =dotspacemacs-configuration-layers= list in this
+  file.
+
+  Subsequently install the required dependencies with =SPC SPC
+  install-eaf-dependencies=. [fm:: This command only works for Ubuntu and Fedora
```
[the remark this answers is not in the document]
dalanicolai: This should have been a footnote i.e `[fn;;`. Probably, if Spacemacs uses the eamcs org -> html exporter, they are supported. But maybe the footnote wasn't so nice anyway. So I replaced it with normal text.

layers/+tools/eaf/packages.el
dalanicolai: Initially I updated the keybindings to this prefix, however, without the web search layer installed I still had to declare the prefix myself. On the other hand, I did not like the combination (too complicated). So no i moved everything under `aa` (and `aab`) i.e. `application`-`application-framework(-browser)` prefixes, as the browser is an application-framework application in the end anyway. (and typing `aa` is justnicer than typing `aw`).

dalanicolai (commented):
Actually I see that I wrote [here](https://github.com/syl20bnr/spacemacs/pull/14291#discussion_r555405703) that `SPC SPC` was not working, but is already working fine for a while now.
This layer is really quite nice, also comes with a nice pdf-viewer, and it is working more or less perfectly how Spacemacs users would expect.

dalanicolai:
I see the commit includes some "deleted files" from the `.emacs.d/private` folder. I don't know why this suddenly got included in the commit. Anyway, you can ignore these of course.

I have my private folder symlinked to a private folder in a some cloud synced directory. I did not yet figure out how to deal with that in git properly (maybe I have now).

So now I created a `personal` branch where I deleted the private folder and commited it. Then I created the symlink, and again committed it. I guess the issues with the symlinked private folder will be gone now.

If you have a similar situation, maybe you have some better suggestion for how to deal with it?

layers/+tools/eaf/README.org
```diff
@@ -0,0 +1,93 @@
...
+  Subsequently install the required dependencies with =SPC SPC
+  eaf-install-dependencies=. This command only works for Ubuntu and Fedora based
+  distributions. Users of Arch (based) distributions can run the script manually
+  in their terminal. If the script does not work for you then
+  [[https://github.com/manateelazycat/emacs-application-framework#dependency-list][install the required dependencies manually]].
+
+  To make EAF your default browser within Emacs (handy for e.g. [[https://github.com/sshaw/git-link][git-link]]), add
+  the following line to you user-config:
+ #+begin_src emacs-lisp :tangle yes
+(setq browse-url-browser-function 'eaf-open-browser)
```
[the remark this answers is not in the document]
smile13241324: Nope, I also think we do not need that alias.

smile13241324:
To be honest I have no private files in spacemacs-git repo, instead I have one in my home directory. However private should actually be ignored by git due to a rule in .gitignore 🤔 I could later have a look at this if you like, maybe the git ignore statement is wrong.

layers/+tools/eaf/packages.el
smile13241324: aa would be fine with me, however we should then also change the tag, as this layer would then do much more than just web-service related bindings i.e. I have seen functions accessing cameras for example.

I'll look up a proper tag when I am merging it.

smile13241324 (approved):
Looks ok, this layer definitely needs some more polish but this is something for a different PR. For now I find it already very useful.

I have noticed some smaller typos in the docs which I will fix during the merge.

smile13241324:
Manually merged your commit and revised the layer. Thats a good starting point, but we will definitely have to invest some more time for this layer to work fully, i.e. I have tried some random functions and got a lot of commandp errors which imply the function is not interactive.

Also it would be cool to combine the search targets maintained in the web search layer to be the default choices for the eaf-browser instead of starting a new collection there.

dalanicolai:
Great! For sure a lot can be done and improved about the layer but indeed I thought a starting point will help a lot as users can help discover it and provide feedback.

I discovered that the latest version of the EAF does not seem to work here, probably already from when they switched from dbus to epc. I discovered it earlier already, but I had not enough time to look more into it. However, so I guess the current version is working for you?
I will check if I can make epc work here. Otherwise we could maybe add some logic to see if a working epc is available and if not then in that case fix EAF to commit 133b827cc2ed1f43b0a8f92e8048a0728bd9e2c5.

Also I have another question which is why you wrap the code after the `use-package` keywords in a `progn` form?

Again thanks for all the work and investment!

dalanicolai:
@smile13241324 Does this PR work for you? I found the `epc` error was simply solved by installing the emacs `epc` package, but even after installing that, the layer does not work as expected . For example, I get errors when just pressing `SPC`. Maybe you can test it (just open the browser with e.g. `SPC aabs` and then try to hit `SPC`) and if it does not work, then we could fix eaf to commit `133b827cc2ed1f43b0a8f92e8048a0728bd9e2c5` with which this layer should be working fine.

dalanicolai:
Well I am working on a fix and the major part is easy to fix. Just I do not know how to get the `major-mode-leader-key` emulation back. But I have asked the EAF developers for help

smile13241324:
To be honest I did not bother installing the executable, I have just checked the lisp part @dalanicolai.

PR #14296, opened 2021-01-14
[org] Update org-enable-valign

duianto:
Document the variable: `org-enable-valign`
and move both the variable definition and the `valign` package setup to the `org` layer.

---

I asked the maintainers if the variable also should end in `-support`, like most of the other org layer variables.

But JAremko pointed out that:
>adding -support hints at some kind of integration
and this is not integration
>
>I would name it something like `org-force-table-alignment`
name like this hints at potential incontestability.
or `org-enable-forced-table-alignment`
kinda long-ish

JAremko also suggested that:
>maybe author of the mode can introduce option to alight only small tables ?
or like give end user a choice with integer variable or something

and:
>he also can add table annotation that controls alignment

If someone with more knowledge about valign agrees, then the variable can be changed. And feel free to suggest the other improvements upstream.

### DIFF
--- a/layers/+emacs/org/README.org
+++ b/layers/+emacs/org/README.org
@@ -26,6 +26,7 @@
   - [[#sticky-header-support][Sticky header support]]
   - [[#epub-support][Epub support]]
   - [[#jira-support][Jira support]]
+  - [[#valign-support][Valign support]]
   - [[#verb-support][Verb support]]
 - [[#key-bindings][Key bindings]]
   - [[#starting-org-mode][Starting org-mode]]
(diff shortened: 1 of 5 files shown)


smile13241324 (approved):
:+1: this is much more maintainable, I wonder why I have merged it into `spacemacs-org` in the first place.

lebensterben:
In the current `-support` section, some of them are actually `-integration` of external services/programs, and some are `-extension` of the ORG layer.

We can rename them in a future PR.

smile13241324:
Manually merged.

PR #14297, opened 2021-01-15
Add org-contacts to org layer

dalanicolai:
Org-contacts is a handy contacts management system. It can be used to manage
(email) addresses (compatible with gnus, mu4e, notmuch etc.), birthdays and
more. It is much simpler than bbdb/ebdb and probably powerful enough for most
users.

### DIFF
--- a/layers/+emacs/org/README.org
+++ b/layers/+emacs/org/README.org
@@ -15,6 +15,8 @@
   - [[#twitter-bootstrap-support][Twitter Bootstrap support]]
   - [[#gnuplot-support][Gnuplot support]]
   - [[#revealjs-support][Reveal.js support]]
+  - [[#org-contacts-support][Org-contacts support]]
+    - [[#v-card-importexport][V-Card import/export]]
   - [[#org-journal-support][Org-journal support]]
   - [[#hugo-support][Hugo support]]
   - [[#trello-support][Trello support]]
@@ -154,6 +156,42 @@ Alternatively, add the following line to each =.org= file you want to process:
 #+BEGIN_EXAMPLE
   #+REVEAL_ROOT: https://cdn.jsdelivr.net/npm/reveal.js@3.8.0
 #+END_EXAMPLE
+
+** Org-contacts support
+[[https://github.com/tkf/org-mode/blob/master/contrib/lisp/org-contacts.el][org-contacts]] is a simple contacts management system.
+
+To install org-contacts, set the variable =org-enable-org-contacts-support= to
+=t=. Optionally, also set the variable =org-contacts-files= and add a capture
+template. The value of the =org-contacts-files= variable should be a list with
+filenames to use as contact sources. If set to =nil= (default) then all your Org
+files will be used. The first file in the =org-contacts-files= list can be
+visited with the keyboard shortcut ~SPC a o C f~.
+#+BEGIN_SRC emacs-lisp
+  (setq-default dotspacemacs-configuration-layers
+                '((org :variables org-enable-org-contacts-support t
+                       org=contact-file '("file1.org" "file2.org")
+                       ("c" "Contacts" entry (file "~/Org/contacts.org")
+                        "* %(org-contacts-template-name)
+  :PROPERTIES:
+  :EMAIL: %(org-contacts-template-email)
+  :END:"))))
+#+END_SRC
+A more elaborate capture template can be found in the =org-contacts.el= file.
+
+Contacts can include the :BIRTHDAY: keyword. To include the birthdays in your
+org-agenda add ~%%(org-contacts-anniversaries)~ to one of your contacts files.
+It is important that this is inserted after a heading an that it has no
+preceding whitespace. So probably the best way is to add
+#+begin_src emacs-lisp
+  * Birthdays
+  %%(org-contacts-anniversaries)
+#+end_src
+to the beginning or the end of one of your contacts files.
+
+*** V-Card import/export
+Importing/exporting contacts can be done via the `org-vcard-import/export`
+commands provided by the [[https://github.com/flexibeast/org-vcard][org-vcard]] package. This package gets installed
+automatically after enabling org contacts support.
 
 ** Org-journal support
 [[https://github.com/bastibe/org-journal][org-journal]] is a simple journal management system that:
--- a/layers/+emacs/org/funcs.el
+++ b/layers/+emacs/org/funcs.el
@@ -12,6 +12,19 @@
 ;; Autoload space-doc-mode
 (autoload 'space-doc-mode "space-doc" nil 'interactive)
 
+(defun org-clocks-prefix ()
+  (if org-enable-org-contacts-support
+      "clocks/contacts"
+    "clocks"))
+
+(defun org-contacts-find-file ()
+  (interactive)
+  (if org-contacts-files
+      (find-file (car org-contacts-files))
+    (message "No specific org-contacts-files defined. Org-contacts uses all org files.")))
+
+
+
 (defun org-projectile/capture (&optional arg)
   (interactive "P")
   (if arg
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -19,10 +19,13 @@
         gnuplot
         (helm-org-rifle :toggle (configuration-layer/layer-used-p 'helm))
         htmlize
-        ;; ob, org and org-agenda are installed by `org-plus-contrib'
+        ;; ob, org, org-agenda and org-contacts are installed by `org-plus-contrib'
         (ob :location built-in)
         (org :location built-in)
         (org-agenda :location built-in)
+        (org-contacts :location built-in
+                      :toggle org-enable-org-contacts-support)
+        (org-vcard)
         (org-brain :toggle (version<= "25" emacs-version))
         (org-expiry :location built-in)
         (org-journal :toggle org-enable-org-journal-support)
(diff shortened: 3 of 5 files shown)


layers/+emacs/org/packages.el
```diff
@@ -563,6 +568,8 @@ Headline^^            Visit entry^^               Filter^^                    Da
         ("." org-agenda-goto-today)
         ("gd" org-agenda-goto-date)))
     :config
+    (when org-enable-org-contacts-support
```
Kazark, line 571: I think that the proper way to do this is to put it in the package list at the top of the file with a `:toggle`.

Kazark (commented):
I like the idea of a really full-featured org setup.

dalanicolai:
First of all thanks for reviewing. Maybe I understand you wrong, but did you see that the org-contacts package is declared in the package-list with a toggle already? But that did not work, it does not load the package. So I additionally have to load the package. For that I added the `init-org-contacts` function, but as it is deferred, it will only load after calling an autoload function. Often the agenda requires the package already  before that, so I load `org-contacts` also when the agenda is loaded on the condition that `org-enable-org-contacts-support` is `t`. So I load it like it is explained in the last example of [this use-package doc section](https://github.com/jwiegley/use-package#getting-started).

Kazark:
Ah, I'm careless! Sounds like you are ahead of me on this one. I just hadn't seen something like this before. I can't say I understand why the load with toggle doesn't just work.

layers/+emacs/org/README.org
```diff
@@ -15,6 +15,8 @@
   - [[#twitter-bootstrap-support][Twitter Bootstrap support]]
   - [[#gnuplot-support][Gnuplot support]]
   - [[#revealjs-support][Reveal.js support]]
+  - [[#org-contacts-supported][Org-contacts supported]]
+- [[#birthdays][Birthdays]]
```
Kazark, line 19: This birthdays line looks wrong to me

layers/+emacs/org/packages.el
```diff
@@ -184,7 +186,7 @@ Will work on both org-mode and any mode that accepts plain html."
 
       (dolist (prefix '(
                         ("mb" . "babel")
-                        ("mC" . "clocks")
+                        ("mC" . "clocks/contacts")
```
Kazark, line 189: Would it make any sense to add `/contacts` conditionally?

dalanicolai:
Good one... I am not sure, but I would say it makes sense indeed. I will
await some third person's opinion, and then I will be happy to update it. I
am not sure about the shortcut anyway. Better suggestions are welcome, but
probably the current one makes the most sense.

layers/+emacs/org/funcs.el
```diff
@@ -12,6 +12,14 @@
 ;; Autoload space-doc-mode
 (autoload 'space-doc-mode "space-doc" nil 'interactive)
 
+(defun org-contacts-find-file ()
+  (interactive)
+  (if org-contacts-files
+      (find-file (car org-contacts-files))
+    (message "No specific org-contacts-files defined. Org-contacts uses all org files.")))
+
```
lebensterben, line 20: Too many blank lines.

layers/+emacs/org/packages.el
lebensterben: I agree. This should be added only when `org-contacts` is enabled.

lebensterben (changes requested):
Please see my comments.

dalanicolai:
https://github.com/syl20bnr/spacemacs/pull/14297#discussion_r564182868 that comment should be solved now. (I don't know how to reply there, git(hub) is still a mystery sometimes)

layers/+emacs/org/funcs.el
```diff
@@ -12,6 +12,19 @@
 ;; Autoload space-doc-mode
 (autoload 'space-doc-mode "space-doc" nil 'interactive)
 
+(defun org-clocks-prefix ()
```
Kazark, line 15: 👍

layers/+emacs/org/README.org
```diff
@@ -155,6 +157,46 @@ Alternatively, add the following line to each =.org= file you want to process:
   #+REVEAL_ROOT: https://cdn.jsdelivr.net/npm/reveal.js@3.8.0
 #+END_EXAMPLE
 
+** Org-contacts supported
+[[https://github.com/tkf/org-mode/blob/master/contrib/lisp/org-contacts.el][org-contacts]] is a simple contacts managemant system.
```
smile13241324, line 161: I think I have spotted a small typo:
```suggestion
[[https://github.com/tkf/org-mode/blob/master/contrib/lisp/org-contacts.el][org-contacts]] is a simple contacts management system.
```

layers/+emacs/org/README.org
```diff
@@ -155,6 +157,46 @@ Alternatively, add the following line to each =.org= file you want to process:
...
+org-agenda add ~%%(org-contacts-anniversaries)~ to one of your contacts files.
+It is important that this is inserted after a heading an that it has no
+preceding whitespace. So probably the best way is to add
+#+begin_src emacs-lisp :tangle yes
+* Birthdays
+%%(org-contacts-anniversaries)
+#+end_src
+to the beginning or the end of one of your contacts files.
+
+The [[https://github.com/flexibeast/org-vcard][org-vcard]] package may be used to import/export vcards. This package can be
```
smile13241324, line 192: When you mention `org-vcard` I think it would be more useful to directly integrate it into the layer basically you have already with the instructions below ;)

smile13241324 (changes requested):
Nice idea, I am also a big fan of integrating everything we have into the org layer.
Therefore I would like to ask you to also integrate `vcard` support if possible.

I have also found a small typo, the rest looks good.

dalanicolai:
I agree with you, and I have added it to the PR (and removed a few more mistakes in the docs).  Additionally, I have sent a [PR](https://github.com/flexibeast/org-vcard/pull/33#issue-568372142) for a small fix to org-vcard because it errored when I tested it. I hope that PR will get merged soon also.

layers/+emacs/org/README.org
```diff
@@ -155,6 +157,42 @@ Alternatively, add the following line to each =.org= file you want to process:
...
+filenames to use as contact sources. If set to =nil= (default) then all your Org
+files will be used. The first file in the =org-contacts-files= list can be
+visited with the keyboard shortcut ~SPC a o C f~.
+#+BEGIN_SRC emacs-lisp
+  (setq-default dotspacemacs-configuration-layers
+                '((org :variables org-enable-org-contacts-support t
+                       org=contact-file '("file1.org" "file2.org")
+                       ("c" "Contacts" entry (file "~/Org/contacts.org")
+                        "* %(org-contacts-template-name)
+  :PROPERTIES:
```
lebensterben, line 175: Is the indentation correct here?

smile13241324:
I have merged your PR however I have noticed a lot of issues during testing, which I have fixed during the merge.
Can you double check the fixed version?

Issues I have solved:
   - org-contacts-find-file did not fail correctly if `org-contacts-files` was nil
   - New bindings for `org-contacts-find-file` were introduced independent of whether contact support was activated
   - `org-vcard` was installed even when no contact support was requested

In addition I have refactored the way the package is loaded to avoid cros referencing too much from the org-init function.
Can you give me a hint why we need to actively require org-contacts? I have checked the package and it seems to have autoloading implemented. I have kept eager loading for now but we should definitely double check.

Also I am struggling a bit to find out what this package actually adds to org-agenda, my agenda looks the same independent of whether I have loaded this package or not. Maybe some more hints to its capabilities should be made in the org readme.

dalanicolai:
@smile13241324 

Thanks for fixing things. Your commit looks good...

About point three, I could not find a good explanation for how the :toggle keyword works. (Also I think I could not find any documentation about the `:location built-in` option). So I only added the `:after org-contacts` in its use-package form, but I did not realize it would get installed in any case. Thanks for notifying, I will know it for the next time.

About actively requiring org-contacts. I do this because, without it, when I try to open my agenda immediately after startup I get the warning:
```
Bad sexp at line 2 in /home/dalanicolai/Sync/org/contacts.org: (let ((entry ) (date '(2 1 2021))) (org-contacts-anniversaries))
```
i.e. org-agenda requires org-contacts to know what to do with the (org-contacts-anniversaries) (see also [here](https://github.com/syl20bnr/spacemacs/pull/14297#issuecomment-761171655). Which also answers your question about what org-contacts adds to the agenda; with org-contacts you can add anniversaries/birthdays to your contacts and these will be shown in the agenda.
Of course org-contacts is also for contact management, so it also works as a `lightweight` (e.g. it uses a simple textfile instead of some less accessible data-structure to save contacts) alternative for managing mail contacts. Various major Emacs email packages can use it for address management/completion.

I will add these hints to the README after my breakfast:)

Thanks again for your great work!

KjartanOli:
@smile13241324 I just pulled your commit, and I get the "Bad sexp" error that @dalanicolai mentioned if I don't manually eval
```
(use-package org-contacts)
```
before opening the agenda.

dalanicolai:
@KjartanOli @smile13241324 Interesting... I also just now pulled the commit, and it seems to work fine here

dalanicolai:
I get the error too indeed (I thought I did not have it, but that was because I did not pull/merge correctly)

smile13241324:
Sorry guys, not sure why this is not working on your side, unfortunately I cannot properly reproduce, so I did the best I could that is I have analysed the loading order in the original PR and noticed that the package needs to be required before the config of org-agenda. Currently it would run after org-agenda.

So I have now made this change, please report here whether you still have this issue.

Second I have tried to copy paste the example code from the README which does not look right, @dalanicolai can you double check? I think the part about the custom org template is not correct.

fosskers:
Something here still isn't quite working as advertised. I have the following in my config:
```elisp
...
          org-contacts-files '("/home/colin/sync/contacts.org")
          ("c" "Contacts" entry (file "/home/colin/sync/contacts.org")
           "* %(org-contacts-template-name)
:PROPERTIES:
:EMAIL: %(org-contacts-template-email)
:END:"))
```
But upon startup I'm warned:
```
(Spacemacs) Warning: Missing value for variable (c Contacts entry (file /home/colin/sync/contacts.org) * %(org-contacts-template-name)
:PROPERTIES:
:EMAIL: %(org-contacts-template-email)
:END:) !
```

dalanicolai:
@smile13241324 @fosskers Yes I was just asked to have a look at it in the last comment. So I will send a PR now with the fix (just

The change from `post-config` to `pre-config` did not fix the "Bad sexp" error. I have also tested the `post-init` and `pre-init` variants, none of them work.

smile13241324:
I have added the original use-package declaration back, lets see if this resolves the issue. I have also merged the document fixes from @dalanicolai.

fosskers:
I'm still getting the `Warning: Missing value for variable...` warning that I pasted above. I honestly can't see what's wrong with the Elisp I have there.

dalanicolai:
@smile13241324 Ah I think a little error slipped in while merging manually merging the documentation fix [here](https://github.com/syl20bnr/spacemacs/pull/14354/commits/fae6a8c69a2cea8193ea154e87dbfc7e554f6d32).
The commit there shows that the :PROPERTIES: block keyword ends with a :END: keyword. However, when I pull the latest develop branch and visit the org-layer README (@fosskers you can use `SPC h l` for that), then I find the the :END: keyword was not included in the merge...

smile13241324:
Looks like this is an issue with `dotspacemacs//prettify-spacemacs-docs`. When I manually activate org-mode after having opened the buffer the text is correctly displayed.

More information about this can be found here #14354

PR #14298, opened 2021-01-15
Fix #14213 remove dash and ht from core libs

### LINKED ISSUE #14213: [perf] Spacemacs should make sure that it loads dash bytecompiled.
Without doing this the performance of spacemacs is seriously hurt. As a result, the dash functions run 2+ times slower than they would run if they are used in vanilla and they are on the critical path of many packages. 

The same most likely applies to `ht`

thanhvg:
- delete dash and ht in core
- reimplement all ht functions we need in `spacemacs-ht.el` and use them
- add dash to spacemacs-bootstrap


smile13241324:
As far as I understood the plan is actually to auto-update these libraries and compile them automatically later instead of copying functions we use right now. @JAremko can you have a look here?

JAremko:
Yeah we  keep synced versions of them.  Adding more code that we'll have to maintain isn't good.

thanhvg:
Honestly I would rather spacemacs didn't even use `ht` and `dash` at all. Those `ht` functions spacemacs are using are very simple. If you take  a look at the their implementation in this PR  which is a copy pasta the gain they give over vanilla is just to make sure that they are in line functions. All these functions don't need dash at all. 

When I was aware of this ugly dependency I found out that for `dash` we will have two versions of dash one in core lib by spacemacs and one  brought in by third party packages like `lsp`. It easy to verify,  Do a `c-h f` and pick a dash function like `--if-let`, emacs will show it is defined some where from  somewhere like `~/.emacs.d/elpa/28.0/develop/dash-20200803.1520/dash.el` not `~.emacs.d/core/lib.

JAremko:
I don't have strong feelings about moving dash (although we have to make sure that Spacemacs doesn't use it before dash becomes available) 

But copy-pasting copyrighted code from ht.el doesn't look like a good idea in general.  Also ht.el seems to be often updated. For example, they  switched from `defsubst` to `define-inline` recently.  I would assume they benchmark and tune the code and it will have to get updated if something big like native-compilation gets released.

JAremko:
Dash also needs to be removed from https://github.com/syl20bnr/spacemacs/blob/develop/.ci/built_in_manifest list

thanhvg:
Copy pasta is not good but saying it is because of copyright doesn't sound right to me. With that reason, isn't syncing dash and ht  from upstream to our repo also copy pasta? It's GPL I can reuse as long as I admit it and publish it. 

My point on ht is that our dependency on it is pretty light. We use only 6 out 29 functions ht provides. And these 6 functions are very basic. For example here is one function:

``` elisp
(define-inline ht-get (table key &optional default)
  "Look up KEY in TABLE, and return the matching value.
If KEY isn't present, return DEFAULT (nil if not specified)."
  (declare (side-effect-free t))
  (inline-quote
   (gethash ,key ,table ,default)))
```

I would just rather use `gethash` directly instead because the speed gain for us would be negligible. I suggest that spacemacs will stay away from using ht and dash going forward and this PR is the first step in that direction. 

There is no way to remove dash without removing ht. ht requires dash for some of its functions that spacemacs never uses. If ht is removed then dash can be put into bootstrap because we only use dash after bootstrap as far as I understand.

I looked into your solution to compile core libs which is brilliant and has much broader scope than this PR however I'm afraid we would create more problems that way. Now we have to manage the elc files and make sure they are up to date and it's not free. The speed we gain in elc would be wasted by the code to manage the sync. Also does it fix the problem of double versions of dash or ht in both core lib and epla? Will it be  compatible with the native compile feature with in emacs 28? Moreover why do we have to take more responsibilities while there are  already solutions such as elpa/package.el or straight.el?

Sorry if I'm sounding like a dick here but I really want to protect my bet on spacemacs. I've been managing my own fork of spacemacs on top of dev branch for more than a year without any problem, enjoying free improvements from the community. But I'm afraid the new direction on dev branch will make it harder for me to continue this free ride.

Thanks

JAremko:
> Copy pasta is not good but saying it is because of copyright doesn't sound right to me. With that reason, isn't syncing dash and ht from upstream to our repo also copy pasta? It's GPL I can reuse as long as I admit it and publish it.

The problem here is that `spacemacs-ht` isn't transformative (or what is the legal term) it's basically a fork. So you should include original header + add  extra cred to it (if I remember correctly) I wouldn't bother with a small personal project but.... Spacemacs is a totally different case.  If you really want to make this into a fork you should figure out what we should put into the header  :thinking: 

> I looked into your solution to compile core libs which is brilliant and has much broader scope than this PR however I'm afraid we would create more problems that way. 

When native compilation arrives I think we'll have to compile some of the core Spacemacs files anyway - to stay competitive :wink:  so with the lib compilation I'm started to lay the ground for it.  If we keep `auto-compile-on-load-mode` enabled then we won't have any problem with keeping .elc up to date. But it  has noticeable performance penalty (albeit only during the first load and if something really heavy starts like `org-mode`) so I hacked code that will clean .elc files when git revision changes. This should cover most of the cases and be pretty reliable.  I'm also looking into enabling `auto-compile-on-load-mode` after Spacemacs startup or make it configurable with `.spacemacs` as a potential solution if we'll have problems with stale .elc files.

thanhvg:
Ok, if I update the header of `spacemacs-ht` with layers' words and move it to `lib/forks` will this PR get accepted? 

Also your byte compile solution of core libs should take into account of Emacs version too I think. Emacs solves this problem by loading packages in particular paths, for example emacs 27 will load those in `~/.emacs.d/elpa/27.1/` emacs 28 will load `~/.emacs.d/elpa/28.0/`. 

I am hopping between 27 and 28 all the time on the same spacemacs code base. So to avoid any conflict I set `spacemacs--compiled-files` to nil which basically nullifies the byte compile feat.

JAremko:
> Ok, if I update the header of spacemacs-ht with layers' words and move it to lib/forks will this PR get accepted?

I think header like this should do:

```
;;; spacemacs-ht.el --- hash table util  -*- lexical-binding: t; -*-
;; Copyright (C) 2013 Wilfred Hughes - original implementation
;; Copyright (C) 2021 Sylvain Benner & Contributors - fork
;;
;; Author: Thanh Vuong
;; URL: https://github.com/thanhvg
;;
;; This file is not part of GNU Emacs.
;;
;;; License: GPLv3
;;; Commentary:  This is a fork of ht.el by Wilfred Hughes
;;  https://github.com/Wilfred/ht.el/blob/master/ht.el

;;; Code:
```

Also remove `ht.el` and `dash.el` from   https://github.com/syl20bnr/spacemacs/blob/develop/.ci/built_in_manifest
and it's should be ok.

JAremko:
``
Also your byte compile solution of core libs should take into account of Emacs version too I think. Emacs solves this problem by loading packages in particular paths, for example emacs 27 will load those in ~/.emacs.d/elpa/27.1/ emacs 28 will load ~/.emacs.d/elpa/28.0/.

I am hopping between 27 and 28 all the time on the same spacemacs code base. So to avoid any conflict I set spacemacs--compiled-files to nil which basically nullifies the byte compile feat.
```

Hmm. I think I'll add .spacemacs var that disables compilation + make sure that if Emacs version changes .elc files are recompiled.

thanhvg:
Update:
- moved spacemacs-ht to forks folder and updated its header, added to compilation list
- removed dash and ht from compilation list and from build manifest.

JAremko (approved):

JAremko:
@smile13241324  Seems legit.  Lets merge ?

smile13241324:
Yep, I'll merge it right away.

smile13241324:
Merged manually, thanks for the hard work @thanhvg and @JAremko.

PR #14299, opened 2021-01-16
[core] Shorten package update restart message

duianto:
problem:
When the Emacs frame has the default window width (80 characters).

Then the current restart to update message:
Emacs has to be restarted to actually install the new version of the packages (SPC q r)

is cut off at the Emacs frames right edge:
Emacs has to be restarted to actually install the new version of the packages (S

solution:
Shortening the message:
Restart Emacs to install the updated packages. (SPC q r)

says the same thing, and keeps it within 80 characters.


Kazark (approved):

smile13241324 (approved):
Looks good

PR #14301, opened 2021-01-18
Core comp

JAremko:
Byte compile some built-in libs.


core/core-compilation.el
```diff
@@ -0,0 +1,35 @@
...
+    "core/libs/package-build-badges.el"
+    "core/libs/package-build.el"
+    "core/libs/package-recipe-mode.el"
+    "core/libs/page-break-lines.el"
+    "core/libs/quelpa.el"
+    "core/libs/spinner.el")
+  "List of Spacemacs files that should be compiled.
+File paths are relative to the `user-emacs-directory'.")
+
+(defun spacemacs//ensure-compilation (files)
```
JAremko, line 28: I'll add some of the spacemacs core stuff here. It will require a lot of compile time stubs if done async so I'll do it in the main Spacemacs thread (instance) and use the fact that most of the core libs are loaded at this point.

core/core-compilation.el
```diff
@@ -0,0 +1,35 @@
...
+  "Make sure that the .el FILES are compiled."
+  (dolist (file files)
+    (let ((fbn (file-name-sans-extension (file-truename file))))
+     (unless (file-exists-p (concat fbn ".elc"))
+       (byte-compile-file (concat fbn ".el"))))))
```
JAremko, line 33: I'll add some of the Spacemacs core files to the compilation list. It will require a lot of compile time stubs if done async so I'll do it in the main Spacemacs thread (instance) and use the fact that most of the core libs are loaded at this point.

core/core-compilation.el
```diff
@@ -0,0 +1,35 @@
...
+    "core/libs/page-break-lines.el"
+    "core/libs/quelpa.el"
+    "core/libs/spinner.el")
+  "List of Spacemacs files that should be compiled.
+File paths are relative to the `user-emacs-directory'.")
+
+(defun spacemacs//ensure-compilation (files)
+  "Make sure that the .el FILES are compiled."
+  (dolist (file files)
+    (let ((fbn (file-name-sans-extension (file-truename file))))
```
JAremko, line 31: Should rename the variable to "fbp" for OCD sake.

core/core-compilation.el
```diff
@@ -0,0 +1,35 @@
...
+    "core/libs/package-build.el"
+    "core/libs/package-recipe-mode.el"
+    "core/libs/page-break-lines.el"
+    "core/libs/quelpa.el"
+    "core/libs/spinner.el")
+  "List of Spacemacs files that should be compiled.
+File paths are relative to the `user-emacs-directory'.")
+
+(defun spacemacs//ensure-compilation (files)
+  "Make sure that the .el FILES are compiled."
```
JAremko, line 29: Hmm those aren't necessary files .el extension.  Just Elisp files.

JAremko:
Ok. I'll fix typos and stuff in the next revision. Lest's see if it works at all.

core/core-compilation.el
```diff
@@ -0,0 +1,35 @@
...
+;; Copyright (c) 2012-2021 Sylvain Benner & Contributors
+;;
+;; Author: Eugene "JAremko" Yaremenko <w3techplayground@gmail.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+(defconst spacemacs-compiled-files
```
JAremko, line 12: Should also mention that `packed.el` and `auto-compile.el` are omitted on purpose.   They are loaded before we can check if they are up to date.

PR #14307, opened 2021-01-20
[Core] *core-spacemacs.el: fix default-path for compiling built-in libs

sunlin7:
Compiling core libs should based on `spacemacs-start-directory` instead of `user-emacs-directory`.


sunlin7:
Hi @JAremko , could you review and apply this patch please? Thank you.

JAremko:
Thanks :+1:

sunlin7:
Response so quick! :rocket: 
Thank you.

PR #14310, opened 2021-01-20
[c-c++] Add layer variable for DAP adapter selection

domenzain:
Require the `dap-mode` adapter to the official LLDB DAP backend (`lldb-vscode`)
by default.

### DIFF
--- a/layers/+lang/c-c++/README.org
+++ b/layers/+lang/c-c++/README.org
@@ -24,6 +24,10 @@
         - [[#example-dotspacemacs-configuration-layers-entry][Example dotspacemacs-configuration-layers entry]]
       - [[#completion][Completion]]
       - [[#debugger-dap-integration][Debugger (dap integration)]]
+        - [[#adapter-selection][Adapter selection]]
+        - [[#microsoft-adapter-installation][Microsoft adapter installation]]
+        - [[#codefreak-adapter-installation][CodeFreak adapter installation]]
+        - [[#lldb-adapter-installation][LLDB adapter installation]]
     - [[#rtags][rtags]]
       - [[#external-dependencies-1][External dependencies]]
       - [[#configuration-1][Configuration]]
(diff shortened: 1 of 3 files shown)


domenzain:
@yyoncho, I found this backend the easiest to install given lldb-vscode is in the apt package for lldb.
Is there a good reason to prefer lldb-gdb-dap over lldb-dap for use with lldb?

yyoncho:
@domenzain I think the best one is cpptools - see https://emacs-lsp.github.io/lsp-mode/tutorials/CPP-guide/

domenzain:
Thanks for the reply, I will have a look and maybe make that the c-c++ layer default as well as make it user-configurable on layer selection.

layers/+lang/c-c++/README.org
```diff
@@ -205,10 +209,39 @@ have been disabled in favour of server,
...
+To install the debug adapter you may run =M-x dap-cpptools-setup= when you are
+on Linux.
+
+***** CodeFreak adapter installation
 To install the debug adapter you may run =M-x dap-gdb-lldb-setup= when you are
 on Linux or download it manually from [[https://marketplace.visualstudio.com/items?itemName=webfreak.debug][Native Debug]] and adjust the
 =dap-gdb-lldb-path= variable.
 
+***** LLDB adapter installation
+Install the =lldb-vscode= program. It is a part of the =lldb= apt package.
```
lebensterben, line 240: This only applies to Ubuntu or any distro that uses `apt`.
A more generic description could be:
```text
Install the =lldb-vscode= program. It usually comes with `lldb` from your distribution's package repository.
Alternatively, ....
```

layers/+lang/c-c++/funcs.el
```diff
@@ -165,7 +165,9 @@
 
 (defun spacemacs//c-c++-setup-lsp-dap ()
   "Setup DAP integration."
-  (require 'dap-gdb-lldb))
+  (mapcar (lambda (dap-mode)
```
lebensterben, line 168: `dap-mode` is a minor mode (and also a buffer-local variable).
To avoid any possible confusion and increase readability, I suggest to rename the argument, for example just name it `x`.
In fact, the entire lambda can be written in a point-free style so you don't need to worry about the naming.

Also `mapc` is more appropriate here because we only want the side-effect.

```elisp
(mapc #'require c-c++-dap-adapters)
```

lebensterben (changes requested):

layers/+lang/c-c++/README.org
domenzain: Done. Thanks!

layers/+lang/c-c++/funcs.el
domenzain: Done. Thanks!

domenzain:
The last force push was rebased onto develop and includes a minor change to the documentation (accidental Markdown \`lldb\` to =lldb=)

smile13241324 (approved):
Good change will be merged soon.

PR #14315, opened 2021-01-23
emacs-lisp: resets to evil-normal after edebug session

agzam:
This one annoyed me for quite some time. Finally, I got too tired of it.

Basically, right after edebug session Spacemacs gets stuck in evil-evilified-state. 

To reproduce:

- instrument any function `M-x spacemacs/edebug-instrument-defun-on`
- run the function, edebug session starts
- quit the session, or step through the function, whatever (it doesn't matter)
- after edebug session keys are in evilified state (where not all normal evil keys work as expected)
  - e.g. you press `<escape>` - cursor turns yellow (should be orange)


duianto:

Issue #14314, opened 2021-01-22
Separators of `lsp-headerline-breadcrumb-mode` are garbled.


rayw000:
I didn't see any fonts initialization in both `all-the-icons` and `lsp-headerline`. Separators go garbled on my spacemacs due to font missing like this.

![image](https://user-images.githubusercontent.com/69779107/105439778-67260600-5ca0-11eb-9b87-0e69a5734e39.png)

After running `all-the-icons-install-fonts`, things look fine.

![image](https://user-images.githubusercontent.com/69779107/105440148-0cd97500-5ca1-11eb-95a4-0a9a9aca8fa1.png)

I'm not sure where I can put `all-the-icons-install-fonts` gracefully without any side effects.

Would someone here give me a hint?

lebensterben:
what do you mean by "without any side effects"?
if you don't want any side effects, why do you put this line of code anywhere in the first place?

rayw000:
@lebensterben 
> what do you mean by "without any side effects"?
> if you don't want any side effects, why do you put this line of code anywhere in the first place?

I mean it may be better if we can run `all-the-icons-install-fonts` automatically to avoid seeing garbled separators, but I'm not sure if it should be run at post-install stage of `all-the-icons`, or at initialization stage of `lsp-headerline`. Neither of them is perfect.

May be we should just document it to let users know they need to install fonts to make `lsp-headerline-breadcrumb` work as expected. Do you agree?

PR #14318, opened 2021-01-25
Document: prompt to run `all-the-icons-install-fonts`

rayw000:
Prompt to run `all-the-icons-install-fonts` in `layers/+tools/lsp/README.org` if both `all-the-icons` and `lsp-headerline` are enabled.

Fix: #14314

### DIFF
--- a/layers/+tools/lsp/README.org
+++ b/layers/+tools/lsp/README.org
@@ -157,6 +157,9 @@ To display the current project, current file, and document symbols,
                 '((lsp :variables lsp-headerline-breadcrumb-segments '(project file symbols))))
 #+END_SRC
 
+You may need to run ~all-the-icons-install-fonts~ if you have ~all-the-icons~ package installed,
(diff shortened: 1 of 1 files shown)


layers/+tools/lsp/README.org
```diff
@@ -157,6 +157,9 @@ To display the current project, current file, and document symbols,
                 '((lsp :variables lsp-headerline-breadcrumb-segments '(project file symbols))))
 #+END_SRC
 
+You may need to run =all-the-icons-install-fonts= if you have `all-the-icons` package installed,
```
Kazark, line 160: Did you mean to use backticks here? Since this is an Org buffer, perhaps you meant tildes?

rayw000: Oh, my fault..
Updated! Thanks for your review! @Kazark

Kazark (approved):

smile13241324 (approved):
Nice catch. For now this is a good solution. We can later think about an auto installation.

PR #14319, opened 2021-01-25
Only clean Spacemacs' .elc files on revision change

JAremko:
It looks like running `auto-compile.el` check  on  every load is quite expensive.
Especially if you do something like "emacs README.org"  - `org-mode` takes much longer time to fully load  with `auto-compile-on-load-mode` enabled.
So instead I'll see if we can simply clean up our .elc files on revision change.  `auto-compile.el` still handles recompilation of .el buffers so all should be fine :thinking:


core/core-release-management.el
```diff
@@ -348,6 +348,20 @@ Example: (1 42 3) = 1 042 003"
...
 
+(defun spacemacs//revision-update ()
+  "Update saved revision value and trigger `spacemacs-revision--changed-hook'."
+  (let ((proc-buffer "spacemacs//revision-update:git-get-current-rev"))
+    (when (eq 0 (process-file "git" nil proc-buffer nil "rev-parse" "HEAD"))
+      (with-current-buffer proc-buffer
+        (goto-char 1)
+        (setq spacemacs-revision--last (current-word))
+        (kill-buffer proc-buffer))))
+  (with-temp-file spacemacs-revision--file
```
JAremko, line 359: Should use Spacemacs'es dump function instead.

JAremko:
This should be a pretty reliable way to clean stale .elc files :thinking:

Issue #14230, opened 2020-12-22
Improve keybindings on spacemacs buffer (home buffer)


rommeswi:
#### Description :octocat:
The keyboard shortcuts for "recent", "projects", etc. are inconvenient. Navigating the spacemacs buffer is not streamlined.

#### Reproduction guide :beetle:
- Start Emacs.
- Press 'r'.
- Cursor jumps to first recent file.
- Press 'r' again.

*Observed behaviour:* :eyes: :broken_heart:
Cursor does not move to the second recent file after pressing 'r' the second time.

*Expected behaviour:* :heart: :smile:
Pressing r should consecutively cycle through all recent files.
Pressing R should cycle backwards through all recent files.

Similar behavior is expected for projects, agenda files, etc.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 26.3
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 62979df31)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
(ivy emacs-lisp
     (evil-snipe :variables evil-snipe-enable-alternate-f-and-t-behaviors t)
     git bibtex latex pdf mu4e multiple-cursors org spell-checking treemacs)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GSETTINGS GLIB NOTIFY ACL LIBSELINUX GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XDBE XIM THREADS LIBSYSTEMD LCMS2

blablablerg:
Ha I didn't know this, nice! But in evil mode j/k works to go up and down. r goes to the first line. Seems fine to me this way?

rommeswi:
I know about the j/k shortcut (Tab also works). I guess it has become just a pet peeve of me to get annoyed every time I have to press another button after "r" to jump to the second line when (just as with searches and "n"/"N") I want to press "r" a second time. In a perfect world, I would even like to cycle through more than just the number of documents displayed. So if I've set my config to display 3 documents, pressing "r" a fourth time would scroll the whole list upwards and display the next document...and....I guess I'm dreaming now....

duianto:
Here's a possible solution to cycle through the lists.
(have only tested them with Recent Files and Projects)

[...]

Issue #12519, opened 2019-07-11
Is it able to show the index of the recent files/projects in spacemacs home page?


c02y:
It is more convenient to display the index of the recent files/projects in spacemacs home page (**spacemacs** buffer) like space-vim/SpaceVim, and I can just hit the index character in the home page to open the file/project directly.

![vim](https://user-images.githubusercontent.com/2135996/61017879-65029c00-a3c7-11e9-8e1c-8d1605fcc860.png)

duianto:
There is a similar feature already.
Pressing `o` calls `spacemacs/ace-buffer-links` which shows one or two letters at the start of every link, pressing those letters opens the selected link.

c02y:
Didn't know that, thanks.

The index displayed automatically in spacemacs buffer is more convenient, and I don't have to type another few keys to open the file like using `o`(`spacemacs/ace-buffer-links`)

duianto:
Agreed, it would be simpler to have a visible single target key.

duianto:
A PR has been opened with a possible solution:
Add home buffer recent files numbers #14326

duianto:
The PR has been merged.
@c02y You can close this if it works as expected now.

lebensterben:
@duianto 
However, if you choose to display recent files by projects, the indices are not shown.

duianto:
How should the numbers be listed when the recent files by project is used?

Listing them from top to bottom is likely to run out of numbers before the last project, when there are several projects.

Another "tricky" case is when both the recent files and recent files by project are shown.

duianto:
I just remember that you had a suggestion here:
https://github.com/syl20bnr/spacemacs/pull/14326#issuecomment-769296614
>For example, r N goes to N-th recent file, and p N goes to N-th project etc.

c02y:
Thanks, the PR works, but as https://github.com/syl20bnr/spacemacs/pull/14326#issuecomment-794891496 said, there is a bug.

And just like you said, another problem is there is no index for other items when 9 number keys are run out.

duianto:
That bug should have been fixed in:
Open home buffer recent files in expected order #14474

I don't know how we should handle shortcuts when the 9 number keys run out.

Adding a two key sequence, matches the current behavior when `o` (`spacemacs/ace-buffer-links`) is pressed.
(As stated above: https://github.com/syl20bnr/spacemacs/issues/12519#issuecomment-510564073)

c02y:
[...]

PR #14326, opened 2021-01-27
Add home buffer recent files numbers

duianto:
Prepend numbers to the Recent Files list.
and bind the number keys to open that file.

Before:

emacs_2021-01-27_16-40-00

After:

emacs_2021-01-27_16-36-16

Fixes: Is it able to show the index of the recent files/projects in spacemacs home page? #12519

And possibly: Improve keybindings on spacemacs buffer (home buffer) #14230

### DIFF
--- a/core/core-spacemacs-buffer.el
+++ b/core/core-spacemacs-buffer.el
@@ -721,21 +721,32 @@ LIST: a list of strings displayed as entries."
 (defun spacemacs-buffer//insert-file-list (list-display-name list)
   "Insert an interactive list of files in the home buffer.
 LIST-DISPLAY-NAME: the displayed title of the list.
-LIST: a list of string pathnames made interactive in this function."
+LIST: a list of string pathnames made interactive in this function.
+
+If LIST-DISPLAY-NAME is \"Recent Files:\":
+prepend each list item with a number starting at: 1
+The numbers indicate that the file can be opened,
+by pressing its number key."
   (when (car list)
     (insert list-display-name)
-    (mapc (lambda (el)
-            (insert "\n    ")
-            (widget-create 'push-button
-                           :action `(lambda (&rest ignore)
-                                      (find-file-existing ,el))
-                           :mouse-face 'highlight
-                           :follow-link "\C-m"
-                           :button-prefix ""
-                           :button-suffix ""
-                           :format "%[%t%]"
-                           (abbreviate-file-name el)))
-          list)))
+    (let ((list-nr 1))
+      (mapc (lambda (el)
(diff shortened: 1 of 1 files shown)


lebensterben:
I've tested this and it works great.
Do you think we can expand this idea to other parts?

For example, `r N` goes to `N`-th recent file, and `p N` goes to `N`-th project etc.

duianto:
Thanks for testing. I'll merge this now.

Additional features can be handled in a new PR.

For example:
Maybe the first startup list in `dotspacemacs-startup-lists`:
```
                                (recents-by-project . (7 . 5))
                                (bookmarks . 5)
                                (projects . 7)
                                (agenda . 5)
                                (todos . 5))
```
could exclude the prefix (only show the numbers).

duianto:
### Issue:
I forgot to test with more than the default 5 recent files.

The home buffer breaks when there are more than 9 recent files.
The messages buffer shows:
>progn: Key sequence 1 0 starts with non-prefix key 1

duianto:
I found a possible solution: 
Limit home buffer recent files numbers to 9 https://github.com/syl20bnr/spacemacs/commit/38a314f7ccf2f9fe702e3d5fdbe3e6eb1eac4a4f

winsphinx:
This is a pretty work. Thanks a lot.
However, I met such an issue -- when i open a file (eg No.3) and edit it and close it (with key ~spc b d~), now emacs returns back to home buffer. Then i try to open it again i press key ~3~, another file is opened. The reason is that the No3 file is last opened and is on the place of No1, but in `recent files` list, it keeps on No3.
Would this issue be checked?

duianto:
Confirmed.

The `recentf-list` has updated in the background but the home buffer still shows the previous order.

The number keys on the home buffer calls commands like this one:
`(recentf-open-most-recent-file-1)`

which opens that item in the `recentf-list`.

One solution might be to update the home buffer when a file is saved,
and the home buffer is open in another window.

possibly like this:
```elisp
    (defun spacemacs-buffer/refresh-in-other-window ()
      (when (get-buffer-window spacemacs-buffer-name)
        (save-selected-window
          (switch-to-buffer-other-window spacemacs-buffer-name)
          (spacemacs-buffer/refresh))))

    (add-hook 'after-save-hook 'spacemacs-buffer/refresh-in-other-window)
    ;; (remove-hook 'after-save-hook 'spacemacs-buffer/refresh-in-other-window)
```

duianto:
It's probably not a good idea to refresh it on every file save.

A PR has been opened with a possible solution:
Fix home buffer recent files order after file save #14472

PR #14328, opened 2021-01-29
Implement C(-S)-tab buffer switch in Helm layer

dalanicolai:
Equivalent to PR #14287 for ivy-layer

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2246,6 +2246,7 @@ Other:
   - Add ~C-d~ to create a project perspective and open a project root dired
     buffer, from the ~SPC p l~ persp switch project prompt
     (thanks to Magnus Therning)
+  - Add =C(-S)-tab= keybindings for buffer switching (thanks to Daniel Nicolai)
 - Fixes:
   - Fixed signature of =spacemacs//display-helm-window= (thanks to Jack Kamm)
   - Fixed initialization by calling =helm-mode= when Helm is initialized=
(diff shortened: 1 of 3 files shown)


Kazark (approved):

smile13241324 (approved):
Cool, thanks for adding this to helm too :)

dalanicolai:
No problem. Personally I prefer to bind this functionality to `C-j` and `C-k` as in evil those bindings do not really add useful functionality (`C-j` in normal mode is somewhat equivalent to `j` in vim or `o` in evil, `C-k` is somewhat equivalent to `dd` in evil and `enter digraph` in vim). In Helm you can then set `helm-follow-mode-persistent` and toggle follow-mode.

For `Ivy` a solution to use `C-j/k` for buffer switching has been posted [here](https://github.com/abo-abo/swiper/issues/1091#issuecomment-773962936).

Maybe we could add these bindings to Spacemacs too, if you think that would be nice then let me know.

lebensterben:
Bindings with `TAB` is more "expected" than `j/k`:
- In many desktop environments, across Linux, MacOS, and Windows, `<MODIFIER> TAB` is the default keybinding to switch between active windows.
- In Spacemacs, `SPC TAB` is the default keybinding to switch between the last two buffers.

A possible solution is to add a layer variable to `helm` and `ivy`:
- When it's non-nil, `SPC TAB` is used to switch to next buffer and `SPC S TAB` is used to switch to last buffer.
- When it's nil, it keeps the current default behaviour, that `SPC TAB` switches between last two buffers. And the new functionality in this commit is bound to `C TAB` and `C S TAB`.

dalanicolai:
@lebensterben Well it is not about expectation (although that is why I implemented the C-tab). It is about offering and suggesting to users the most convenient workflow. I think there is need to not improve things because of conventions/expectations (although it is important to consider such improvements carefully).

Well, otherwise users can find the suggestion here anyway.

PR #14330, opened 2021-01-29
Add home buffer named shortcuts functions

duianto:
Define named functions: spacemacs-buffer/jump-to-...
for the home buffer shortcuts. So that a descriptive name is shown,
in for example the view-lossage (C-h l) buffer.

Before:
```
 r                      ;; anonymous-command
 p                      ;; anonymous-command
```
After:
```
 r                      ;; spacemacs-buffer/jump-to-recent-files
 p                      ;; spacemacs-buffer/jump-to-projects
```

### DIFF
--- a/core/core-spacemacs-buffer.el
+++ b/core/core-spacemacs-buffer.el
@@ -575,21 +575,46 @@ If MESSAGEBUF is not nil then MSG is also written in message buffer."
       (when messagebuf
         (message "(Spacemacs) %s" msg)))))
 
+(defun spacemacs-buffer//startup-list-jump-func-name (str)
+  "Given a string, return a spacemacs-buffer function name.
+
+Given:           Return:
+\"[?]\"            \"spacemacs-buffer/jump-to-[?]\"
+\"Recent Files:\"  \"spacemacs-buffer/jump-to-recent-files\""
+  (let ((s (downcase str)))
+    ;; remove last char if it's a colon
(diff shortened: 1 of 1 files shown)


lebensterben:
I'm glad that my idea was adopted.
Currently the supported list types in home buffers are:
- recents
- bookmarks
- projects
- agenda
- todos
We actually can use their initials as the keybinding for each of them.

duianto:
This is unrelated to the PR:
Add home buffer recent files numbers #14326

The key bindings for jumping to the startup lists first item was already present.

This PR just adds named functions for the shortcuts.

Another startup list `recents-by-project` has been added recently:
[core] support recents-by-project in dotspacemacs-startup-lists #14259

It seems like the intention was to add it under the key binding: `R`
But it was added as:
`(spacemacs-buffer||add-shortcut "R" "Recent Files:")`

Currently if both the `recents` and `recents-by-project` lists are shown, then both `r` and `R` jumps to the `recents` lists first item.

duianto:
I "fixed" the `R` key to jump to the recents by project.

The key bindings for agenda: `c`
and todos: `d`
seems a bit strange.

I'm not sure why they were chosen.

`a` is undefined.
`t` with the evil editing style, calls `evil-find-char-to`.
It might not be that useful on the Spacemacs home buffer.

smile13241324 (approved):
Cool

PR #14333, opened 2021-01-30
[evil-collection] support for dired

thanhvg:
Hopefully ;) or at least the RET key binding. I use `ranger` so I can't tell this will cover all the cases.


duianto:

PR #14347, opened 2021-02-04
[javascript] add npm-mode

hexagonal-sun:
Add [npm-mode](https://github.com/mojochao/npm-mode) to the javascript layer.  This allows easy execution of npm
commands at the project root from any javascript file in the project.

### DIFF
--- a/layers/+lang/javascript/README.org
+++ b/layers/+lang/javascript/README.org
@@ -332,7 +332,7 @@ parse warnings:
 | Key binding | Description                                                                     |
 |-------------+---------------------------------------------------------------------------------|
 | ~SPC m n i~ | Run the =npm install= command in the project root                               |
-| ~SPC m n r~ | Show a list of avilable npm scripts, and execute the selected one               |
+| ~SPC m n r~ | Show a list of available npm scripts, and execute the selected one              |
 | ~SPC m n c~ | Run the =npm clean= command in the project root                                 |
 | ~SPC m n s~ | Prompt for the name of an npm package, install it and save to =dependencies=    |
 | ~SPC m n d~ | Prompt for the name of an npm package, install it and save to =devDependencies= |
--- a/layers/+lang/javascript/README.org
+++ b/layers/+lang/javascript/README.org
@@ -34,6 +34,7 @@
 - [[#key-bindings][Key bindings]]
   - [[#js2-mode][js2-mode]]
   - [[#folding-js2-mode][Folding (js2-mode)]]
+  - [[#running-npm-npm-mode][Running NPM (npm-mode)]]
   - [[#importing-import-js][Importing (import-js)]]
   - [[#refactoring-js2-refactor][Refactoring (js2-refactor)]]
     - [[#documentation-js-doc][Documentation (js-doc)]]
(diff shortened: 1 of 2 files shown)


layers/+lang/javascript/packages.el
```diff
@@ -65,6 +66,21 @@
...
+  (use-package npm-mode
+    :defer nil
+    :init
+    (progn
+      (add-hook 'js2-mode-hook #'npm-mode))
+    :config
+    (progn
+      (spacemacs/declare-prefix-for-mode 'js2-mode "mn" "npm")
+      (spacemacs/set-leader-keys-for-major-mode 'js2-mode
+        "ni" 'npm-mode-npm-install
```
lebensterben, line 79: How about adding other commands and keybindings?

https://github.com/mojochao/npm-mode/blob/3ee7c0bad5b7a041d4739ef3aaa06a3dc764e5eb/npm-mode.el#L34-L41

layers/+lang/javascript/README.org
```diff
@@ -326,6 +327,16 @@ parse warnings:
 | ~SPC m z F~ | toggle hide functions    |
 | ~SPC m z C~ | toggle hide comments     |
 
+** Running NPM (npm-mode)
+
+| Key binding | Description                                                                     |
+|-------------+---------------------------------------------------------------------------------|
+| ~SPC m n i~ | Run the =npm install= command in the project root                               |
+| ~SPC m n r~ | Show a list of avilable npm scripts, and execute the selected one               |
```
smile13241324, line 335: Small typo should be available

layers/+lang/javascript/packages.el
smile13241324: Please also add the following missing commands for a full integration:
Command|Binding|Description
--------------|----------|---------------
npm-mode-npm-init             | <kbd>SPC m n n</kbd> | Initialize new project
npm-mode-npm-uninstall        | <kbd>SPC m n u</kbd> | Remove project dependency           
npm-mode-npm-list             | <kbd>SPC m n l</kbd> | List installed project dependencies 
npm-mode-visit-project-file   | <kbd>SPC m n p</kbd> | Visit project package.json file

smile13241324 (changes requested):
I like the idea to get some of the NPM project maintenance work integrated into Spacemacs. However when we do that we should try to integrate all available commands to allow users to maintain their project without having to constantly switch between NPM and Spacemacs.

I have added a list of additional commands you should integrate prior to having this merged. Also I have found some smaller typos you should fix on the way.

layers/+lang/javascript/packages.el
hexagonal-sun: Added commands as requested.  Thanks for the suggestion!

hexagonal-sun:
@smile13241324  @lebensterben Thanks for the review!  Your comments should have been addressed. Please take another look and let me know.

hexagonal-sun:
One quick question.  It'd be really useful to have these commands avilable in the TypeScript mode too.  Is there a way to have it in both layers without having to duplicate the code?

lebensterben:
https://github.com/syl20bnr/spacemacs/blob/7276970c914af32f7e3d3f45e654eaa2e04bf81e/layers/%2Blang/typescript/layers.el#L12

typescript layer depends on javascript.

So if you define a function in javascript layer, and it should be callable in typescript layer.

hexagonal-sun:
@lebensterben Ah, I didn't know there was an inheritence structure with layers.  Does that mean that the TypeScript mode will also inherit the javascript layer's keybindings?

smile13241324 (approved):
Looks good now

smile13241324:
No this is not an inheritance structure, instead its a structure of dependencies i.e. when typescript-layer is loaded also the javascript-layer is loaded. You can derive modes from one another in emacs, though I have not made the best experiences with using this feature.

The best way to have a layer use a package introduced in another layer is by using the `pre-init-*` construct. In your case you would add this to the typescript layer which would only be executed if the package * is initialised in this function you would add the init code for typescript-mode or actually for the typescript modes, I think they have more than one.

More details can be found in the layers documentation, reachable from within spacemacs with <kbd>SPC h SPC layers return</kbd>

PR #14350, opened 2021-02-07
Also run `treemacs-load-theme` on `dired-mode-hook`

rayw000:
Icons in `dired-mode` buffer are inconsistent with `treemacs-use-all-the-icons-theme` implied. Currently we can only see expected icons in `dired-mode` buffer ***after running command `treemacs` explicitly***.

Better if we can see corrent icon style in `dired-mode`, no matter whether `treemacs-mode` buffer is initialized. It should just depend on the variable `treemacs-use-all-the-icons-theme`.


smile13241324 (approved):
Yep you're right it should not be necessary to explicitly call a mode to see the right icons in dired.

PR #14354, opened 2021-02-07
Fix doc org-contacts

dalanicolai:
I have checked and fixed the docs as requested in  #14297.

Also I have commented out the the org-agenda post/pre-config hook, and put back the conditional `use-package` in the after the :config keyword in `use-package org-agenda` so that it will work correctly. Of course you are welcome to find the trick to reduce the cross-referencing. I have thought about it a little also, but could not quickly come up with another fix.

### DIFF
--- a/layers/+emacs/org/README.org
+++ b/layers/+emacs/org/README.org
@@ -161,7 +161,9 @@ Alternatively, add the following line to each =.org= file you want to process:
 #+END_EXAMPLE
 
 ** Org-contacts support
-[[https://github.com/tkf/org-mode/blob/master/contrib/lisp/org-contacts.el][org-contacts]] is a simple contacts management system.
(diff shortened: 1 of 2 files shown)


smile13241324:
Manually merged

dalanicolai:
I am afraid that the closing :END: of the :PROPERTIES: block did not survive the merge...

smile13241324:
Looks like this is an issue with ` dotspacemacs//prettify-spacemacs-docs`. When I manually activate `org-mode` after having opened the buffer the text is correctly displayed.

@JAremko have you seen this before? Looks like the properties and end tag is removed during the rendering when the buffer is freshly opened via `SPC h SPC org` however when I manually start `org-mode` when the buffer is opened it just looks normal.

JAremko:
image

@smile13241324  this is how it looks for me.  We make  `#+BEGIN_SRC emacs-lisp` and `#+END_SRC` semi-invisible as usual.

Issue #10101, opened 2018-01-08
Ivy buffers is slow


AlejandroCatalina:
#### Description :octocat:
Since I installed the ivy layer I notice that the buffer list is a bit slow (```SPC b b```). 

#### Reproduction guide :beetle:
- Start Emacs
- SPC b b
- Filter some "py" buffers

*Observed behaviour:* :eyes: :broken_heart:
The response from emacs is a bit slow.

*Expected behaviour:* :heart: :smile:
Butter smooth behavior.

#### System Info :computer:
- OS: darwin
- Emacs: 25.3.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. f8ecf0cb)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
(ivy auto-completion better-defaults emacs-lisp git org
     (shell :variables shell-default-height 30 shell-default-position 'bottom)
     spell-checking syntax-checking latex bibtex python mc-column)
```
- System configuration features: JPEG RSVG IMAGEMAGICK NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS MODULES

#### Backtrace :paw_prints:

Thanks and share if this happens to you too or if it's an issue on my side.

PS: I don't have that many buffers open, about 20 or so.

zzjjzzgggg:
Is it related to the recent change in #9543? Try a different version before commit 3467aa84.

AlejandroCatalina:
It shouldn't be related, I noticed the slowness before the change. I will try tho.

syl20bnr:
We should make `ivy-rich` a layer variable.

If you use macOS be sure to have Emacs compile with the vfork patch included in `emacs-plus`. It greatly improve process spawning performance which makes everything more snappy in Emacs (especially Magit). Maybe `ivy-rich` is concerned by this depending on how it fetched info on the buffers/files.

syl20bnr:
Outch I tested it quickly on my Mac with vfork patch enabled... it is indeed really slow.

syl20bnr:
I nobody make a PR I'll do it tomorrow evening.

AlejandroCatalina:
I’ll try to address the PR today. Essentially making ivy-rich a variable (turned off by default).

syl20bnr:
Fixed in develop, thank tou @AlejandroCatalina

achikin:
[...]

PR #14356, opened 2021-02-08
[ivy-rich] enable ivy-rich-project-root-cache-mode

thanhvg:
which improves performance when many buffers opened.

Ref:
https://github.com/Yevgnen/ivy-rich#project-performance
https://github.com/syl20bnr/spacemacs/issues/10101


duianto:

PR #14357, opened 2021-02-08
Fix EAF-layer to make it work with the epc (support Windows)

dalanicolai:


smile13241324:
I have noticed you are loading specific libraries as package dependencies. I am not sure that this is such a good idea, as packages must be owned by one layer. Libraries on the other hand should be shared among them, and should normally be part of the respective package level dependencies.

Is there a specific reason you need to load them? I am not seeing you using any of their commands to be honest.

dalanicolai:
Well the `epc` package is required to make EAF work here, but the other packages I just added because it is instructed in [the EAF installation instructions](https://github.com/manateelazycat/emacs-application-framework#install). The `s` package I just assumed it came with Spacemacs already so I left it out.
Except for the `epc` package I don't really understand/know why the other packages are listed as requirements. Also, I could try to find out if, and then where, these packages are installed within Spacemacs

smile13241324:
No need to, given the current state of the packages involved I would expect to refactor that layer in the near future anyway. We can then find out what kind of libraries we really need. For now they don't hurt.

PR #14360, opened 2021-02-09
Change NPM package elm-analysis -> elm-analyse in the elm-layer README

jrahhali:
There is no such NPM package named elm-analysis, but there is an elm-analyse.


smile13241324 (approved):
You are right, thanks for fixing it.

PR #14367, opened 2021-02-10
[fsharp] Declare package eglot-fsharp

bcc32:
eglot-fsharp is now distributed as a separate MELPA package from
fsharp-mode, so it needs to be listed in `fsharp-packages`.

Also, refactor some conditions to read more clearly.


smile13241324 (approved):
Looks good

Issue #14366, opened 2021-02-10
Broken Colemak bindings for Magit status buffer


fosskers:
#### Description :octocat:

With the recent changes involving `evil-collection`, the `hneio` (i.e. colemak)
bindings for the Magit status buffer seems to have stopped working. Moving the
cursor up and down now requires `j` and `k` like normal `evil-mode`, but this
isn't so ergonomic for Colemak users.

#### Reproduction guide :beetle:

- Start Emacs.
- `SPC f e d` to open your Spacemacs config.
- Set your keyboard layer with: `(keyboard-layout :variables kl-layout 'colemak-hnei)`
- Restart Spacemacs.
- Enter a git project and do the usual `SPC g s`.
- Notice that `n` and `e` do not move the cursor up and down, but `j` and `k` do.

*Observed behaviour:* :eyes: :broken_heart:

Pressing `n` (The physical `j` on the keyboard) to move the cursor down activates the "next" functionality of searching.

*Expected behaviour:* :heart: :smile:

Pressing `n` should move the cursor down one line.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 27.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 9375d6f54)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
((keyboard-layout :variables kl-layout 'colemak-hnei)
 (lsp :variables lsp-rust-server 'rust-analyzer)
 (terraform :variables terraform-auto-format-on-save t)
 auto-completion c-c++ emacs-lisp epub git graphql helm java javascript latex markdown protobuf shell shell-scripts syntax-checking version-control yaml
 (rust :variables rust-format-on-save t lsp-rust-analyzer-diagnostics-disabled
       ["unresolved-proc-macro"])
 (haskell :variables haskell-completion-backend 'lsp lsp-haskell-process-path-hie "haskell-language-server-wrapper" haskell-stylish-on-save t)
 (go :variables go-format-before-save t go-backend 'lsp go-tab-width 4)
 (org :variables org-priority-highest 1 org-priority-lowest 15 org-priority-default 15 org-enable-github-support t org-enable-valign t))
```
[...]

PR #14369, opened 2021-02-11
[keyboard-layout] Fix colemak-hneio next/prev line in magit

duianto:
Fixes: Broken Colemak bindings for Magit status buffer #14366


fosskers:
Works like a charm!

PR #14371, opened 2021-02-11
Add eval-to-comment elisp-mode function

JAremko:
Clojure layer has similar function. I miss it in Elisp sometimes. Is it usefull?   That's basically `eval-print-last-sexp` but that also adds `;;`  which is cool since it doesn't break evaluation of the buffer :thinking:


Kazark (commented):
This is a bit like having Babel directly in an Elisp buffer, right?

JAremko:
@Kazark  It simply inserts a new comment line with the result of the evaluation.

smile13241324 (approved):
Cool, I wonder if there are more bindings in clojure which we could use for elisp...

PR #14373, opened 2021-02-12
Fix home buffer ace links

duianto:
Pressing o `(spacemacs/ace-buffer-links)`
on the Spacemacs home buffer `SPC b h`.

Adds an ace link (one or more highlighted letters),
at the menu buttons, and the startup list entries.

### problem:
The text: `[S P A C E M A C S]` gets an ace link,
but it isn't a button.

When the [?] button has been pressed, and the Quick Help
is open, then ace links appear on the open square brackets
for each key binding, but they aren't buttons.

In the lists: recent files, projects, etc.
The ace links appear before the first / (slash).
This means that in windows they appear after `c:`.

### solution:
Add the ace links at the beginning of each widget-button.

### Before
emacs_2021-02-12_15-16-56

emacs_2021-02-12_15-17-05

### After
emacs_2021-02-12_15-24-18

emacs_2021-02-12_15-24-23

### DIFF
--- a/layers/+spacemacs/spacemacs-navigation/funcs.el
+++ b/layers/+spacemacs/spacemacs-navigation/funcs.el
@@ -258,16 +258,17 @@ If the universal prefix argument is used then kill also the window."
 
 ;; ace-link
 
-(defvar spacemacs--link-pattern "~?/.+\\|\s\\[")
-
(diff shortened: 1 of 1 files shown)


smile13241324 (approved):
Looks good

Issue #11309, opened 2018-09-11
Where to add multi-line package, bindings?


gondalez:
I really like [multi-line](https://github.com/IvanMalison/multi-line) for toggling single/multiline function invocations.

Currently I have it hacked in under `jump/join/split` as below. In future it'd be nice to convert to a transient state.

What would be the best way to package it up for other spacemacs users? 
It seems too small to be in its own layer.

```
diff --git a/layers/+distributions/spacemacs-base/keybindings.el b/layers/+distributions/spacemacs-base/keybindings.el
index 230a424ce..37d4ca970 100644
--- a/layers/+distributions/spacemacs-base/keybindings.el
+++ b/layers/+distributions/spacemacs-base/keybindings.el
@@ -217,6 +217,7 @@
   "jo" 'open-line
   "j=" 'spacemacs/indent-region-or-buffer
   "jS" 'spacemacs/split-and-new-line
+  "jm" 'multi-line
   "jk" 'spacemacs/evil-goto-next-line-and-indent)

 ;; navigation/jumping ---------------------------------------------------------
diff --git a/layers/+distributions/spacemacs-base/packages.el b/layers/+distributions/spacemacs-base/packages.el
index 6d5f3c93f..2dc125eea 100644
--- a/layers/+distributions/spacemacs-base/packages.el
+++ b/layers/+distributions/spacemacs-base/packages.el
@@ -53,6 +53,7 @@
         (visual-line-mode :location built-in)
         (whitespace :location built-in)
         (winner :location built-in)
+        multi-line
         ))

 ;; Initialization of packages
@@ -260,6 +261,10 @@
       (when dotspacemacs-line-numbers
         (global-linum-mode)))))

+(defun spacemacs-base/init-multi-line()
+  :init
+  )
+
 (defun spacemacs-base/init-occur-mode ()
   (evilified-state-evilify-map occur-mode-map
     :mode occur-mode))
```

Miciah:
I would put it in the spacemacs-editing layer.  You could follow the string-inflection transient state as an example: https://github.com/syl20bnr/spacemacs/blob/c122eb6a05ddffeccfe20fc7f56cae550c463575/layers/+spacemacs/spacemacs-editing/packages.el#L384-L402
You might also consider <kbd>SPC x m</kbd> as a key binding.  <kbd>SPC j</kbd> is more movement-oriented commands (think "jumping") whereas <kbd>SPC x</kbd> has a lot of text-transformation commands.

[...]

PR #14386, opened 2021-02-17
[editing] add binding for multi-line

gondalez:
Adds support for cycling between multi line block styles via [multi-line](https://github.com/IvanMalison/multi-line).

See also #11309 - thank you @Miciah for the suggestion all those years ago.

Demo:
CleanShot 2021-02-17 at 10 01 45


smile13241324 (approved):
Looks nice

PR #14393, opened 2021-02-18
Restore evil-visual-state-map on edebug exit

duianto:
### problem:
Exiting `edebug-mode`, leaves the `evil-visual-state-map`
with only the two key bindings that are defined in:
`evilified-state--setup-visual-state-keymap`
```
y       evil-yank
escape  evil-exit-visual-state
```

### solution:
Restore the `evil-visual-state-map` when exiting `edebug-mode`.

Added an alias for the new exit function:
`evilified-state--evilified-state-on-exit`

called: `evil-evilified-state-exit`
the opposite of: `evil-evilified-state`


smile13241324 (approved):
👍

PR #14396, opened 2021-02-19
Add org-notify option to org-layer

dalanicolai:
This PR adds notifications via org-notify as an option to the org layer. Org-notify uses the Emacs [notifications package](https://www.gnu.org/software/emacs/manual/html_node/elisp/Desktop-Notifications.html), which works great with desktop notification systems that communicate via the dbus. Also, it should work well on Windows. I don't know about Mac, but it would be very easy to add some `org-notify-action-alert` function to the org-notify package (a simple copy of the `org-notify-action-message` with `message replaced by alert).

### DIFF
--- a/layers/+emacs/org/README.org
+++ b/layers/+emacs/org/README.org
@@ -11,6 +11,7 @@
 - [[#important-note][Important Note]]
 - [[#install][Install]]
   - [[#layer][Layer]]
+  - [[#agenda-notifications][Agenda notifications]]
   - [[#github-support][GitHub support]]
   - [[#twitter-bootstrap-support][Twitter Bootstrap support]]
   - [[#gnuplot-support][Gnuplot support]]
@@ -111,6 +112,37 @@ More details can be found [[https://github.com/syl20bnr/spacemacs/issues/8106][h
 To use this configuration layer: in the main Spacemacs configuration
 file (=~/.spacemacs=), to the existing =dotspacemacs-configuration-layers= list
 add the =org= entry.
+
+** Agenda notifications
+To enable notifications for agenda events, set the variable
+=org-enable-notifications= to =t=.
+
+#+BEGIN_SRC emacs-lisp
+  (setq-default dotspacemacs-configuration-layers '(
+    (org :variables
+         org-enable-notifications t
+         org-start-notification-daemon-on-startup t)))
+#+END_SRC
+
+This will install and configure the [[https://code.orgmode.org/bzg/org-mode/src/master/contrib/lisp/org-notify.el][org-notify]] package. Run the
+=org-notify-start= command to start the =org-notify= notification daemon. To
+start the notification daemon automatically on Spacemacs startup, set the
+variable =org-start-notification-daemon-on-startup= to =t= as shown in the code
+snippet above.
+
+The notifications system can be configured via the =org-notify-add= function. An
+explanation for how to use this function can be found in the initial comment of
+the =org-notify.el= file (=SPC h d f org-notify-add RET o a=) and in the
+function its docstring.
+
+The =org-notify= package will scan through your =org-agenda-files= for =TODO=
+headlines that include the =DEADLINE:= stamp. Press =M-S RET= to enter a todo
+headline and then press =, d d= to add a deadline stamp to the headline. This
+layer implements a configuration type named =default=. Org-notify automatically
+applies this type to all =TODO= entries with a =DEADLINE:= stamp. To set a
+custom notification type for an item add a =:NOTIFY:= property where its value
+should be the name of a type that you have added via =org-notify-add= function
+(see the explanation in the initial comment of the =org-notify.el=).
 
 ** GitHub support
 To install GitHub related extensions like [[https://github.com/larstvei/ox-gfm][ox-gfm]] to export to GitHub
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -23,6 +23,8 @@
     (ob :location built-in)
     (org :location built-in)
     (org-agenda :location built-in)
+    (org-notify :location built-in
+                :toggle org-enable-notifications)
     (org-contacts :location built-in
                   :toggle org-enable-org-contacts-support)
     (org-vcard :toggle org-enable-org-contacts-support)
@@ -587,6 +589,21 @@ Headline^^            Visit entry^^               Filter^^                    Da
       (kbd "M-RET") 'org-agenda-show-and-scroll-up
       (kbd "M-SPC") 'spacemacs/org-agenda-transient-state/body
       (kbd "s-M-SPC") 'spacemacs/org-agenda-transient-state/body)))
+
+(defun org/init-org-notify ()
+  (use-package org-notify
+    :commands org-notify-start
+    :init
+    (when org-start-notification-daemon-on-startup
+      (org-notify-start))
+    :config
+    (org-notify-add 'default
+                    '(:time "-1s" :period "20s" :duration 10
+                            :actions (-message -ding))
+                    '(:time "15m" :period "2m" :duration 100
+                            :actions -notify)
+                    '(:time "2h" :period "5m" :actions -message)
+                    '(:time "3d" :actions -email))))
 
 (defun org/init-org-brain ()
   (use-package org-brain
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2800,6 +2800,7 @@ files (thanks to Daniel Nicolai)
   - Added package =org-rich-yank= (thanks to Keith Pinson)
   - Added package =org-roam= (thanks to Mariusz Klochowicz)
   - Added package =org-contacts= (thanks to Daniel Nicolai)
+  - Added package =org=notify= (thanks to Daniel Nicolai)
 - Key bindings:
   - ~SPC m T i~ to toggle inline images
   - Move clock related key bindings to ~SPC a o C~
--- a/layers/+emacs/org/config.el
+++ b/layers/+emacs/org/config.el
@@ -34,6 +34,12 @@ path, one file per project is used (and the path is relative to
 the project root). If it an absolute path, one global file is
 used.")
 
+(defvar org-enable-notifications nil
+  "If non-nil org-notify is configured.")
+
+(defvar org-start-notification-daemon-on-startup nil
+  "If non-nil start the notification daemon on startup.")
+
 (defvar org-enable-org-contacts-support nil
   "If non-nil org-contacts is configured.")
 


lebensterben:
org-notify was last updated in 2012. There's a risk of incompatibility with future Emacs version.

A more recent package is org-alert https://github.com/spegoraro/org-alert, which was last updated 3 years ago.

layers/+emacs/org/README.org
```diff
@@ -112,6 +114,46 @@ To use this configuration layer: in the main Spacemacs configuration
...
+=:NOTIFY:= property where its value should be the name of a type that you have
+added via =org-notify-add= function. If you add a notification type with the
+name =default=, this type is applied to headlines without the =:NOTIFY:=
+property.
+
+#+begin_src org
+** TODO Enable org notifications in Spacemacs
+   DEADLINE: <2021-02-28 Sun 20:51>
+   :PROPERTIES:
+   :NOTIFY: appt
```
smile13241324, line 153: I have a question about this, does that mean all my todos must be tagged separately? I would expect to have the package pull in all my todos by default to be honest. Can we configure the package like this by default and give people a hint how they can change it to explicitly tagging messages?

layers/+emacs/org/packages.el
```diff
@@ -588,6 +590,19 @@ Headline^^            Visit entry^^               Filter^^                    Da
       (kbd "M-SPC") 'spacemacs/org-agenda-transient-state/body
       (kbd "s-M-SPC") 'spacemacs/org-agenda-transient-state/body)))
 
+(defun org/init-org-notify ()
+  (use-package org-notify
+    :config
```
smile13241324, line 595: Can we lazy load the package?

layers/+emacs/org/packages.el
```diff
@@ -588,6 +590,19 @@ Headline^^            Visit entry^^               Filter^^                    Da
...
+    (org-notify-add 'appt
+                    '(:time "-1s" :period "20s" :duration 10
+                            :actions (-message -ding))
+                    '(:time "15m" :period "2m" :duration 100
+                            :actions -notify)
+                    '(:time "2h" :period "5m" :actions -message)
+                    '(:time "3d" :actions -email))
+    (when org-start-notification-daemon-on-startup
+      (org-notify-start))))
```
smile13241324, line 604: What happens when you call org-notifiy-start twice? I am just wondering if we have to start that once for every buffer or once for an emacs session?

smile13241324 (changes requested):
Looks promising. I still found some smaller issues and questions you should check before this can be merged.

@lebensterben can you give a short overview about the difference between org-notification and org-alarm? To be honest I would prefer the one being better maintained to be honest.

dalanicolai:
@lebensterben Thanks for checking the PR. Actually I see that I have not been clear in the PR message, but I meant that we could easily make `org-notify` support (extend it with) [alert](https://github.com/jwiegley/alert) notifications. But that would be only required for Mac's because on Linux and Windows, the Emacs [notification](https://www.gnu.org/software/emacs/manual/html_node/elisp/Desktop-Notifications.html) package provides a better notification system. Really, I have tested both the `alert` package and the `notifications` package, and the notifications package really is more powerful because it actually communicates via the dbus, while org-alert only sends out messages. (Just quickly try and see how well the proposed feature in this PR works, at least when using linux. The dbus communication is great!).

 So I am quite confident that org-notify is the best option. There is quite some confusion about notifications in the community (which for me personally was mainly caused by the existence of the [alert](https://github.com/jwiegley/alert) package.

So let me start with the main culprit. `org-alert` is indeed a package by John Wiegley, which of course makes users (including myself) think it is probably an enhanced version of what was available. But after looking into it, I have found that the `notification` packages is a better (and more modern) package for providing desktop notifications. I have looked closely into it (I have even created [two PR's](https://github.com/jwiegley/alert/pulls) for org-alert, of which the [second one](https://github.com/jwiegley/alert/pull/95) is an update of its documentation to include a reference to Emacs its notification package

It took me actually a long research to finally get to this, what should be, most straightforward solution of using org-notify. I have checked out, and inspected many packages: 
- [orgalert](https://github.com/spegoraro/org-alert)
- and its [updated/enhanced forks](https://github.com/spegoraro/org-alert/network)
- [org-wild-notifier](https://github.com/akhramov/org-wild-notifier.el) (which already looks better than org-alert)
- [appt.el](https://www.emacswiki.org/emacs/appt.el)

But all these packages have flaws. Org-alert and org-wild-notifier use org-alert, which seems better but as I explained above, it is not. Additionally they use complicated ways to retrieve the correct headlines (e.g. scheduled and deadlines).
`org-notify` uses a much cleaner way to obtain the correct headlines, and I would say its code, despite being older, is more modern than that of org-alert and org-wild-notifier (some of the forks already improve the way org-alert retrieves headlines, but still org-notify its code is cleaner).

Also [org-alert] has a very limited notification system of only checking and showing notifications periodically (see [here](https://github.com/herbertjones/org-alert/blob/f87bff4acbd839acb4d2245b56b2c3d21f950911/org-alert.el#L43).

`org-wild-notifier` describes `org-notify` [to be complicated](https://github.com/akhramov/org-wild-notifier.el#how-does-it-differ-from-existing-packages), but I would not call it very complicated, and for sure the code of org-notify is much cleaner than that of `org-wild-notifier`. org-wild-notifier uses a very cumbersome process to retrieve correct headlines, while org-notify uses a very clean way for retrieving them (using the org-elements API)

the [appt.el](https://www.emacswiki.org/emacs/appt.el) package, I have checked out also, because the first two packages did not satisfy. But indeed this package I would say uses old and deprecated ways.

So finally about the incompatability risk, it seems that org-notify actually has the lowest risk of becoming incompatible. It might be true that `org-alert` works better for Macs but it is 1 second work to add an alert function to org notify (as I said in the PR message, just use a copy of the `org-notify-action-message` function.

layers/+emacs/org/README.org
dalanicolai: You will find in the documentation part that this can be achieved by creating a `default` notification type. So I refer to that part (the README file) of this PR.

layers/+emacs/org/packages.el
dalanicolai: Ah yes, we could lazy load it. I first loaded it `:after` org-agenda, but then I figured that users would like the notification daemon to start at Spacemacs startup. But the `org-notify-start` function is autoloadable so yes, we could lazy load it (althoguh it would still start at startup if users choose to start the daemon at startup).

layers/+emacs/org/packages.el
dalanicolai: This function should only be called once for an Emacs session.

dalanicolai:
Let me elaborate a little more.

`org-alert` and org-wild-notifier` both let users configure the type of headlines they would like to include (default is DEADLINES and SCHEDULED). So this would probably be nice for org-notify too.
Also `org-wild-notifier` uses a nice way to schedule the notifications, which could be nice to have in org-notify too, although I think it is not really worth the trouble.

But regarding the `incompatability` I think `org-notify` is the way to go, it is clean code and should be used as a starting point. Of course users are welcome to enhance/extend its features, but I think we should guide them to the right package (at least in my opinion:) to start with. They could use the `org-alert` and `org-wild-notifier` packages for inspiration.

layers/+emacs/org/packages.el
smile13241324: Are you sure, the config code is only executed when the package is loaded, so if you call org-notifiy-start in there and lazy the autoloading on that function will not trigger it. Instead you would need to load the server in the init part, if requested.

layers/+emacs/org/README.org
smile13241324: Yeah I have seen it there, however don't you think that most people will just expect all tasks with deadlines to be notified about. I mean a deadline is a tag afterall, so I would expect this as the sensible standard behaviour i.e. creating a default notification type with package loading and document how to disable this. Then giving an example how to configure your own message types and to set them explicitly for messages.

smile13241324:
Thanks for the very detailed explanation @dalanicolai. I agree with you lets start with this package, if features are missing we could still do PRs on upstream. Maybe we can contribute to solving this entire emacs notification confusion by focusing on the package with the cleanest code and extend it where necessary.

layers/+emacs/org/README.org
dalanicolai: Ah good point, on the one hand, I don't expect users to enable this feature without reading the documentation. But indeed it makes sense to rename the "appt" to "default" to make the tagging unnecessary. :flushed:

layers/+emacs/org/packages.el
dalanicolai: Yes it should be moved too :init then also :+1:

dalanicolai: Ah okay, it looks like `org-notify-start` is not autoloadable. So I think we should keep it as it is now. Do you think I should create a PR for making it autoloadable?

lebensterben:
@dalanicolai 
Thanks for your detailed explanation.

One minor issue, the link in you commit is not the original org-notify.
From org-mode's [webpage](https://orgmode.org/worg/org-contrib/), the raw file is available [here](https://code.orgmode.org/bzg/org-mode/raw/master/contrib/lisp/org-notify.el)

lebensterben:
This seems to be actively maintained.

https://code.orgmode.org/bzg/org-mode/src/master/contrib/lisp/org-notify.el

layers/+emacs/org/packages.el
smile13241324: I think you can declare it autoloadable in use-package directly using the `:commands` tag. Still an up-stream PR would be nice for the rest of the emacs world.

dalanicolai: Ah great! I did not know that feature yet

wztdream:
Only to notice that emacs function `notifications-notify` seems not compatible with gnome3.36 (ubuntu20.04 gnome-shell)
You can check it by:
1. using ubuntu20.04 os
2. modify below code, change to your sound file path, then eval it, then there should be no desktop notify.
```
(notifications-notify
      :title "some title"
      :sound-file file-of-sound-file
      )
``` 
3. remove the :sound-file item, then restart gnome-shell, then re-eval it, it should work correctly
So it seems notifications-notify have trouble if there need play sound files in gnome-shell 3.36

lebensterben:
@wztdream 
> In order to use this functionality on POSIX hosts, Emacs must have been compiled with D-Bus support, and the notifications library must be loaded. 

https://www.gnu.org/software/emacs/manual/html_node/elisp/Desktop-Notifications.html

dalanicolai:
I have checked here on Ubuntu 20.10 with gnome 3.38.3 and also on Fedora 33 with KDE plasma 5.20.5. On both systems notifications seem to work fine, except that the sound does not play even if the :sound-file keyword is included.

On Ubuntu I tried with Emacs from the default Ubuntu repo, and it worked alright (except for playing the sound, but the notifications were still working even when the :sound-file keyword was included). Anyway, seems like it has been build with D-Bus support.

wztdream:
@ lebensterben I install emacs27 with this [ppa](https://launchpad.net/~kelleyk/+archive/ubuntu/emacs)
1. it was compiled with dbus
eval `(notifications-get-capabilities)`  shows `(:actions :body :body-markup :icon-static :persistence :sound)` and it also support play sound.
2. I `(require 'notifications)` ahead but there is no difference, desktop notify seems hang, I mean no pop up window for all notifications, not only emacs but all notifications from os not showing up, no sound played too
3. if remove the :sound-file item, restart gnome-shell, then every thing works fine
4. I checked on ubuntu18.04 with lower version of gnome-shell, with emacs from same PPA, it works fine
5. So it seems this is gnome-shell bug

lebensterben:
I don't use gnome now so I cannot reproduce this.

wztdream:
I just report this bug to gnome-shell, hope this will be solved there [issue](https://gitlab.gnome.org/GNOME/libnotify/-/issues/17)

dalanicolai:
@wztdream Well I noticed that in Gnome 3.38 the notifications do get added to the notifications list (in the clock dropdown menu), therefore I said they were working alright. But actually I agree with you that it does not work alright, because it should indeed show pop-ups and play sounds also. So it is great that you created the gnome-shell bug report!

dalanicolai:
Okay guys, after using `org-notify` for the last few days, I noticed that I really prefer the notification 'configuration style' of [org-wild-notifier](https://github.com/akhramov/org-wild-notifier.el). So I continued to look for better alternatives (maybe creating one), including again `org-wild-notifier`, and now I changed my mind about that package for a few reasons:

- I got some wrong ideas about the `alert` package. At first, I preferred to use the `'notifications` package, because it comes with Emacs, and the development of org-alert seems slow. Besides that, there is still [the bug](https://github.com/jwiegley/alert/pull/94) that notifications using the `libnotify` option don't work. Also, org-notify uses the `potentially` nice feature of the notifications package that notifications can be suspended from the desktop notification pop-up. But now I found that, the `alert` package actually provides an option to use the notifications package (although suspending the notifications does not seem very useful anyway). Also, I noticed that it is actually available from Melpa (did not notice that before).
- The way `org-wild-notifier` retrieves entries looked less elegant/efficient to me because it does not use the `org-elements-api` but instead first lets the agenda format its entries, and then retrieves them from there. But now I think that is probably an easy way to do it, and moreover it is done asynchronously. (Also I did/do not like `dash` package function names that `org-wild-notifier` uses everywhere, because they are not very informative. But 'unfortunately' those functions seem very powerful/elegant. Finally the package gave an error here at first, which is gone now after some updates.)
- I already mentioned in an earlier comment that I prefer the configuration and scheduling style of `org-wild-notifier`. Just see [how it can be configured](kkjhttps://github.com/akhramov/org-wild-notifier.el) yourself (org-notify only allows notifications of `TODO` headlines that include a DEADLINE stamp).

So now, I am wondering... should we just replace the org-notify package, or should we make both packages available?

lebensterben:
using dash is plus for me.
CL simulation is too verbose.
Symbols in dash resembles the feeling of some Haskell operators.

But not using org-elemnt API is quite disappointing. Why giving up the benefit of dealing with AST....

dalanicolai:
@lebensterben Well I guess, the org-agenda uses the `org-element-api` below. So I should say it is not using it directly.

dalanicolai:
Ass you can see I have created another PR for the org-wild-notifier package. It replaces this PR.

And just for sake of documentation. There exists the [https://github.com/legalnonsense/org-timed-alerts](org-timed-alerts) package now also.

lebensterben:
I think you should add the one you actually use. So at least that one will have a maintainer.

If someone is interested in the other notification packages, he can add it later.

dalanicolai:
@lebensterben Yes I have created the [PR](#14418) for the org-wild-notifier package. That will be the one I am using for the upcoming period.

dalanicolai:
Closing this PR in favor of PR #14418

PR #14401, opened 2021-02-20
Revert to default value of helm-swoop-speed-or-color

yyoncho:
- having font-lock enabled for swoop buffers is really slow


smile13241324 (approved):
:+1:

PR #14402, opened 2021-02-20
[lsp] Do not include lsp-ui by default

yyoncho:
- lsp-ui is very opinionated so better to have it enabled/installed on demand

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -190,6 +190,7 @@ the [[file:CHANGELOG.org][CHANGELOG.org]] file.
   - Change jump back to last mark from ~SPC m g p~ to ~SPC m g b~
     to follow the keybindings convention (thanks to Magnus Therning)
   - Use ~lsp-ivy~ when ~ivy~ layer is present.
+  - added  ~lsp-use-lsp-ui~ variable to control whether ~lsp-ui~ is installed or not.
 ***** IPython notebook
 - Key bindings;
   - Change prefix from ~SPC a i~ to ~SPC a y~ (thanks to Swaroop C H)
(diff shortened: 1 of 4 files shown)


smile13241324 (approved):
👍

lebensterben:
I think if any of layer variable that relies on lsp-ui is non-nil, this new variable should default to t.
Otherwise this PR is a breaking change to many users.

smile13241324:
F***, you are right I was in a hurry and just moved it through :exploding_head:. To solve this probably we should define two profiles for the LSP layer, one for a complex and feature rich experience and a second one for a simple and fast one.

This is similar to what we do with helm vs ivy and treemacs vs neotree. We should add these choices to the initial dotfile creation workflow to enable easy access to this feature. Lsp layer was due for a revision anyway lets do this soon.

lebensterben:
After we migrate to use `defcustom` to define layer variables, we probably can do these based on hierarchical structure of variables.

For example, `Spacemacs::lsp::lsp-ui-enable` defaults to t if any of `Spacemacs::lsp::lsp-ui::*` is t.

yyoncho:
> For example, `Spacemacs::lsp::lsp-ui-enable` defaults to t if any of `Spacemacs::lsp::lsp-ui::*` is t.

that means that in order to have lsp-ui off by default everything should be off by default, right?

lebensterben:
Yep.
To be clearer, we probably should accept more fine grained settings:
- If it's `full`, all lsp-ui features is on.
- If it's `off`, explicitly disable `lsp-ui` package.
- If it's other non-nil value, `lsp-ui` package is installed and only its default features are enabled.
- If it's nil,
    - If any other lsp-ui variables are set, say lsp-hearline-breadcrumb, `lsp-ui` should be installed so it defaults to t.
    - Otherwise, defaults to `off`.

PR #14408, opened 2021-02-21
Change SPC gfd to magit-file-dispatch

agzam:
I believe `SPC g f d`  should be more "contextual", allowing user to perform operations
associated with the current file. 

For example, if you need to open the diff buffer related  to the exact line at point, there's currently no "Spacemacsy" key to do it quickly. One has to do: `C-c M-g d`.

Current `SPC g f d` keybinding opens "general" magit-diff dialog, not specifically aimed to help with operations related to the current file.

This tiny PR remaps `SPC g f d` to `magit-file-dispatch`, which feels to be exact mnemonic for this thing.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -345,6 +345,7 @@ In org-agenda-mode
 - Key bindings:
   - Changed ~SPC g h o~ to ~SPC g o~ for =browse-at-remote= (thanks to Codruț
     Constantin Gușoi)
+  - Changed ~SPC g f d~ to bind to ~magit-file-dispatch~ (thanks to Ag Ibragimov)
 ***** YAML
 - Added LSP support (thanks to Seong Yong-ju)
 ***** ycmd
--- a/layers/+source-control/git/packages.el
+++ b/layers/+source-control/git/packages.el
@@ -161,7 +161,7 @@
         "gc"  'magit-clone
         "gfF" 'magit-find-file
         "gfl" 'magit-log-buffer-file
-        "gfd" 'magit-diff
+        "gfd" 'magit-file-dispatch
         "gi"  'magit-init
         "gL"  'magit-list-repositories
         "gm"  'magit-dispatch


lebensterben:
`magit-file-dispatch` is not really `git-diff`, but `git-blame`. See: https://magit.vc/manual/magit/Blaming.html

So probably `SPC g f b` is  better.

agzam:
@lebensterben 

`magit-file-dispatch` docstring says: "Invoke a Magit command that acts on the visited file."

Have you even tried invoking it?

lebensterben:
How about `SPC g f .`?

agzam:
Why? `SPC g f d` makes complete sense here. It actually makes more sense than the current binding. Current binding binds to the "global" magit-diff dispatch, the command doesn't really makes sense here. Perhaps it should be moved to `SPC g d` key.
Normally, I strongly oppose to any changes to existing keybindings, but in this case I believe this change makes it better.
`SPC g f d` - is the perfect mnemonic for `magit-file-dispatch`.

lebensterben:
"." is usually used for transient state.

`magit-file-dispatch`
=> `SPC g` + `f` + `.`

agzam:
Yet, it is not a transient-state. `magit-file-dispatch` does not work like a Spacemacs' transient (even though visually resembles one). Besides, that creates a confusion with existing `SPC g .` - `vcs-transient-state`.

duianto:

agzam:
Oh wow. I did not expect it to be merged so soon, since we were still discussing it, but thank you @duianto. @lebensterben I just want to apologize if I come across as a passive-aggressive person, I do appreciate your feedback. I'd like to emphasize again: I rarely advocate for changing existing keybindings (people get angry when you mess with their muscle memory), but this case felt like the right thing to do.

duianto:
It seems like a very good change.

Additional magit file related commands become available,
and  `magit-diff` is still available under: `SPC g f d D`

duianto:
It was pointed in:
New binding for `magit-file-dispatch` is confusingly different from `magit-dispatch` binding #14434

That, binding `magit-file-dispatch` to `SPC g f d`
introduced a mismatch with: `magit-dispatch` `SPC g m`

It was resolved by restoring `magit-diff` to `SPC g f d`
and binding `magit-file-dispatch` to `SPC g f m`.

Restore SPC gfd for magit-diff, set SPC gfm for magit-file-dispatch #14436

Issue #14348, opened 2021-02-07
[.spacemacs.evn] env var DISPLAY is different for each `ssh -X` session, should not be stored in .spacemacs.evn


sunlin7:
The env `DISPLAY` is stored in `.spacemacs.evn` as constant value, while it's actually different for each `ssh -X ...` session, and cause error when execute `(shell-command-to-string "emacs -q")` with error message when the real `DISPLAY` for running session is `localhost:13.0`, the `localhost:10.0` is the `DISPLAY` value which is stored in `.spacemacs.evn`.
```
Display localhost:10.0 unavailable...
```

smile13241324:
Ups, this doesn't sound right, I will check later if we can exclude it from there.

JAremko:
Add `display` to  `spacemacs-ignored-environment-variables`

sunlin7:
JAremko's solution work in my local, I'll create a PR for this. Thanks

sunlin7:
Close for solution 722fa45ea13412d1edc68a454ec3ab5dfd272669 merged.
Thanks.

PR #14409, opened 2021-02-22
[core-env] #14348 ignore env "DISPLAY" for .spacemacs.env

sunlin7:
As discussed in #14348, the env "DISPLAY" should not be stored in `.spacemacs.env`.


JAremko (approved):

JAremko:
Either way it can be removed from the ignore list be end user.  Or added.  But then wrong display can be more confusing so this should be better default. :+1:

sunlin7:
Thanks

PR #14413, opened 2021-02-22
Categorize languagetool layer as "checker"

real-or-random:
I think that's fine because it's a "grammar checker" and then it appears in the category as spell checking.

Why this is a draft PR:
Does this mean the layer should be moved to the `+checker` directory? If you want, I can add this change to this commit here but I was not sure if I can simply move layers without breaking things. Do I need to change anything else?

### DIFF
--- a/layers/+tools/languagetool/README.org
+++ b/layers/+tools/languagetool/README.org
@@ -1,6 +1,6 @@
 #+TITLE: LanguageTool layer
 
-#+TAGS: layer|uncategorized
+#+TAGS: layer|checker
 
 [[file:img/languagetool.png]]
 


duianto:
Related: Merge languagetool layer with spell-checking layer #14414

real-or-random:
Oh indeed. I think merging the languagetool layer with into the spell-checking layer as proposed in #14414 is much better solution. It's more work of course but if this finds agreement, then I'd simply abandon this PR. This PR may be quicker to implement but it's rather cosmetic and not urgent. So we would not need it at all.

lebensterben:
I'm closing it for now.
I'll reopen it if @real-or-random or anyone is interested in this again.

PR #14415, opened 2021-02-23
Bind s to the dired-quick-sort hydra

duianto:
The `dired-quick-sort` hydra was bound to: `S`

But the introduction of `evil-collection-dired`
[[evil-collection] support for dired #14333](https://github.com/syl20bnr/spacemacs/pull/14333)

Now binds `S` to: `dired-do-symlink`
https://github.com/emacs-evil/evil-collection/blob/334670e29d964c5f591f75ccbf52b7b5faf4daba/modes/dired/evil-collection-dired.el#L66

It also restored the default evil behavior of: `s`
to: `evil-substitute`

But it isn't useful in the dired buffer since it is: read-only

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1121,6 +1121,7 @@ Other:
     (thanks to Ivan Yonchovski)
   - Added ~SPC x n~ for multi-line transient state
     (thanks to Tristan Harmer)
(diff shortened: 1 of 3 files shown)


xuhdev (approved):

smile13241324 (approved):
Whoa, well done dude :)
I couldn't figure out why it was not working, now I see I did not consider evil :facepalm:

PR #14422, opened 2021-02-24
[core] new function spacemacs/set-root-leader-keys-for-mode

thanhvg:
[core] This function allows to bind a mode command behind `dotspacemacs-leader-key' and
`dotspacemacs-emacs-leader-key' when that mode is activate.

[org] apply this function to org narrow commands

motivation:

Currently `spacemacs/set-leader-keys-for-major-mode` and
`spacemacs/set-leader-keys-for-minor-mode` only let us bind command under `spc
m` and `,` That works fine in general. But not all commands should go there. A
mode that deals with compilation or comments when active should have commands
bound behind `spc c` not `spc m`. Another example is org-mode narrow commands
should go to prefix `spc n` not a long and tedious `spc m s b n`.

### DIFF
--- a/core/core-keybindings.el
+++ b/core/core-keybindings.el
@@ -167,4 +167,44 @@ Example:
     (when (spacemacs//init-leader-mode-map mode map t)
       (which-key-add-keymap-based-replacements (symbol-value map) prefix name))))
 
+(defun spacemacs//init-root-leader-mode-map (mode map &optional minor)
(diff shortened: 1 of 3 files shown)


smile13241324:
@syl20bnr can you comment on this PR, I am not sure if those global bindings are not breaking our conventions. If they do the least we should do is to merge the general function to allow people to use it in their personal configs.

PR #14425, opened 2021-02-25
[lsp] add option to bind commands to lsp-command-map

thanhvg:
`lsp-use-upstream-bindings` set to `t` then Spacemacs will bind 
`lsp-command-map` behind `SPC m`, `,` and `M-m`. This way the bindings will be
managed by upstream `lsp-mode` which is documented at
https://emacs-lsp.github.io/lsp-mode/page/keybindings/.

Spacemacs only replaces the prefix `s-l` with `SPC m`. For example:
```
| lsp binding | Spacemacs binding                   |
|-------------+-------------------------------------|
| `s-l s s`   | `SPC m s s` or `, s s` or `M-m s s` |
```

It is assumed in this PR we trust @yyoncho and his gang to manage the bindings for us to save time and effort. Of course, there is a mismatch such as we have `, b` for backend but `lsp-mode` prefer `, s` for session.

### DIFF
--- a/layers/+tools/lsp/README.org
+++ b/layers/+tools/lsp/README.org
@@ -18,6 +18,7 @@
   - [[#language-specific-key-binding-extensions][Language-specific key binding extensions]]
     - [[#spacemacslsp-define-extensions-layer-name-kind-request-optional-extra-parameters][~spacemacs/lsp-define-extensions layer-name kind request &optional extra-parameters~]]
     - [[#spacemacslsp-bind-extensions-for-mode][~spacemacs/lsp-bind-extensions-for-mode~]]
+  - [[#bind-to-lsp-upstreams][Bind to lsp upstreams]]
 - [[#dap-integration][DAP integration]]
 - [[#diagnostics][Diagnostics]]
 - [[#references][References]]
@@ -319,6 +320,24 @@ whereas with ~lsp-navigation~ set to ~'peek~, this is equivalent to:
 #+END_SRC
 
 etc.
+
+** Bind to lsp upstreams
+Alternative to the binding shemes above. If you set ~lsp-use-upstream-bindings~
+to ~t~ then Spacemacs will bind all ~lsp-command-map~ behind ~SPC m~, ~,~ and
+~M-m~. This way the bidings will be managed by upstream ~lsp-mode~ which is
+documented at https://emacs-lsp.github.io/lsp-mode/page/keybindings/. Spacemacs
+only replaces the prefix ~s-l~ with ~SPC m~. For example:
+
+| lsp binding | Spacemacs binding                   |
+|-------------+-------------------------------------|
+| ~s-l w s~   | ~SPC m w s~ or ~, w s~ or ~M-m w s~ |
+
+As ~lsp-mode~ and has a deep integration into ~Spacemacs~. ~Spacemacs~ hackers
+should pay attention to avoid any binding collision with ~lsp-mode~.
+
+#+begin_src elisp
+  (lsp :variables lsp-use-upstream-bindings t)
+#+end_src
 
 * DAP integration
 =lsp-mode= integrates with =dap-mode=, which implements DAP(Debugger Adapter Protocol). See documentation on =DAP= layer for details.
--- a/layers/+tools/lsp/config.el
+++ b/layers/+tools/lsp/config.el
@@ -36,3 +36,5 @@ If `both', binds lightweight navigation functions under `SPC m g' and lsp-ui fun
 (defvar lsp-ui-sideline-ignore-duplicate t "Ignore duplicates")
 
 (defvar lsp-use-lsp-ui t "When non-nil, use `lsp-ui' package.")
+
+(defvar lsp-use-upstream-bindings nil "When non-nil, map keys to `lsp-command-map'.")
--- a/layers/+tools/lsp/funcs.el
+++ b/layers/+tools/lsp/funcs.el
@@ -37,6 +37,24 @@
     (setq key (pop bindings)
           def (pop bindings))))
 
+(defun spacemacs/lsp-bind-upstream-keys ()
+  "Bind upstream `lsp-command-map' behind \"SPC m\" and the likes."
+  (bind-map lsp-command-map
+    :minor-modes (lsp-mode)
+    :keys ((concat dotspacemacs-emacs-leader-key " m") dotspacemacs-major-mode-emacs-leader-key)
+    :evil-keys ((concat dotspacemacs-leader-key " m") dotspacemacs-major-mode-leader-key)
+    :evil-states (normal motion visual evilified))
+  (dolist (it '(("=" . "format")
+                ("F" . "folder")
+                ("T" . "toggle")
+                ("g" . "goto")
+                ("h" . "help")
+                ("r" . "refactor")
+                ("w" . "workspace")
+                ("a" . "actions")
+                ("G" . "peek")))
+    (which-key-add-keymap-based-replacements lsp-command-map (car it) (cdr it))))
+
 (defun spacemacs/lsp-bind-keys ()
   "Define key bindings for the lsp minor mode."
   (cl-ecase lsp-navigation
--- a/layers/+tools/lsp/packages.el
+++ b/layers/+tools/lsp/packages.el
@@ -36,7 +36,9 @@
     :defer t
     :config
     (progn
-      (spacemacs/lsp-bind-keys)
+      (if lsp-use-upstream-bindings
+          (spacemacs/lsp-bind-upstream-keys)
+        (spacemacs/lsp-bind-keys))
       (setq lsp-prefer-capf t)
       (add-hook 'lsp-after-open-hook (lambda ()
                                        "Setup xref jump handler"


layers/+tools/lsp/funcs.el
```diff
@@ -25,6 +25,23 @@
     (setq key (pop bindings)
           def (pop bindings))))
 
+(defun spacemacs/lsp-bind-upstream-keys ()
```
yyoncho, line 28: Here you can use `lsp-enable-which-key-integration` in `(let ((lsp-prefix-key "..."))` block.

yyoncho (approved):
This was my idea from the start but never had the time to implement it.

> Of course, there is a mismatch such as we have , b for backend but lsp-mode prefer , s for session.

I believe that `,s` is consistent with other layers that are managing connections and decided to fix that.

thanhvg:
Let's negotiate.

`, s` by convention is for interaction with repl. So current lsp binding in this PR would collide with those `, s` in javascript layer and  clojure layer and possibly  including `, T` as far as I know.

Can lsp bind all the toggles and session commands to the `l` prefix. (`s-l l blah blah`)? I think `, l` is free in Spacemacs and most of the time I don't use lsp toggles and session commands much. 

Also this is just my idea,  we need the blessing from the big bosses and users too.

yyoncho:
@thanhvg  I agree with you. But IMHO let's either use `b` like we currently do in spacemacs or use `w`? Is that free? It will be `workspace restart`, etc. I hope that other maintainers at lsp-mode side will be fine with the change. The other option is to "fix" the keymap in spacemacs for `s` key and move it under `b`.

thanhvg:
I like `w`. I ran a quick `helm-ag` search with query = `spacemacs/declare-prefix-for-mode "mw` and found none. Search for `"mb"` also returned irrelevant results. So `w` is the best of all.  

`b` is also not safe.
```
layers/+lang/idris/packages.el:54:      (spacemacs/declare-prefix-for-mode 'idris-mode "mb" "idris/build")
layers/+lang/scala/packages.el:45:      (spacemacs/declare-prefix-for-mode 'scala-mode "mb" "sbt")
layers/+tools/ansible/packages.el:81:  (spacemacs/declare-prefix-for-mode 'yaml-mode "mb" "buffer")
layers/+lang/ruby/packages.el:48:            (spacemacs/declare-prefix-for-mode mode "mb"
layers/+lang/sql/packages.el:135:      (spacemacs/declare-prefix-for-mode 'sql-mode "mb" "buffer")
layers/+lang/sql/packages.el:171:      (spacemacs/declare-prefix-for-mode 'sql-interactive-mode "mb" "buffer")
```
`l` is also claimed

```
layers/+lang/bibtex/packages.el:79:      (spacemacs/declare-prefix-for-mode 'bibtex-mode "ml" "lookup")
layers/+lang/sql/packages.el:138:      (spacemacs/declare-prefix-for-mode 'sql-mode "ml" "listing")
```

smile13241324:
I think `w` for `workspace` is a good  choice. What commands would reside in there? Is that all workspace commands and toggles or only the workspace commands?

If this has no conflicts I would suggest to make this the default choice for `lsp-layer` reducing the maintenance work on this layer.

layers/+tools/lsp/config.el
```diff
@@ -24,3 +24,5 @@ If `both', binds lightweight navigation functions under `SPC m g' and lsp-ui fun
 (defvar lsp-ui-sideline-ignore-duplicate t "Ignore duplicates")
 
 (defvar lsp-use-lsp-ui t "When non-nil, use `lsp-ui' package.")
+
+(defvar lsp-use-upstream-bindings nil "When non-nil, map keys to `lsp-commamd-map'.")
```
tko, line 28:
```suggestion
(defvar lsp-use-upstream-bindings nil "When non-nil, map keys to `lsp-command-map'.")
```

layers/+tools/lsp/funcs.el
```diff
@@ -25,6 +25,23 @@
...
+  "Bind upstream `lsp-command-map' behind \"SPC m\" and the likes."
+  (bind-map lsp-command-map
+    :minor-modes (lsp-mode)
+    :keys ((concat dotspacemacs-emacs-leader-key " m") dotspacemacs-major-mode-emacs-leader-key)
+    :evil-keys ((concat dotspacemacs-leader-key " m") dotspacemacs-major-mode-leader-key)
+    :evil-states (normal motion visual evilified))
+  (dolist (it '(("=" . "format")
+                ("F" . "fodler")
```
tko, line 36:
```suggestion
                ("F" . "folder")
```

thanhvg:
@yyoncho the response from Spacemacs side on this PR is positive. How about your lsp side's so far?

layers/+tools/lsp/funcs.el
```diff
@@ -25,6 +25,23 @@
...
           def (pop bindings))))
 
+(defun spacemacs/lsp-bind-upstream-keys ()
+  "Bind upstream `lsp-command-map' behind \"SPC m\" and the likes."
+  (bind-map lsp-command-map
+    :minor-modes (lsp-mode)
+    :keys ((concat dotspacemacs-emacs-leader-key " m") dotspacemacs-major-mode-emacs-leader-key)
+    :evil-keys ((concat dotspacemacs-leader-key " m") dotspacemacs-major-mode-leader-key)
+    :evil-states (normal motion visual evilified))
+  (dolist (it '(("=" . "format")
```
yyoncho, line 35: lsp-mode has these as well, we might make them public/reusable.

yyoncho (commented):
> How about your lsp side's so far?

Anything particular? PR looks good to me.

lebensterben:
There's no issue for this PR per se.
But there are lots of work for integration with other major-modes.

Most major-modes has their key-bindings under `SPC m`/`,`, and we need to make sure there's no collision of key-bindings.

One particular kinds of error I found is when multiple layers declared same prefix for bindings, and then some bindings wouldn't be added at all.

So I propose to make this into a project that contains this PR and a thorough review of any layers that modifies `SPC m`/`,`.

thanhvg:
@yyoncho we agreed that `, w` as `workspace` is the best prefix for session commands, so `lsp-mode` should change the following key maps in `lsp-command-map` accordingly:

``` elisp
      ;; sessions
      "sD" lsp-disconnect "disconnect" (lsp-workspaces)
      "sd" lsp-describe-session "describe session" t
      "sq" lsp-workspace-shutdown "shutdown server" (lsp-workspaces)
      "sr" lsp-workspace-restart "restart server" (lsp-workspaces)
      "ss" lsp "start server" t
``` 
to 

``` elisp
      ;; worskpace
      "wD" lsp-disconnect "disconnect" (lsp-workspaces)
      "wd" lsp-describe-session "describe session" t
      "wq" lsp-workspace-shutdown "shutdown server" (lsp-workspaces)
      "wr" lsp-workspace-restart "restart server" (lsp-workspaces)
      "ws" lsp "start server" t

Once it's done I update this PR to use `, w` and add @tko suggestions then it's ready to go.

**Comment by smile13241324** [comment]
@thanhvg can you give me a status of this? Do we already have a go from lsp-mode and should I take a look at this PR and possibly merge it or are you still working on it?

**Comment by thanhvg** [comment]
@smile13241324 I think the status is waiting for upstream. This PR binds all the sessions commands under `, s` which will collide with:

-  clojure layer which claims `, s` for all kind of repl, cider commands
- javascript layer also takes `, s` for repl too
There may be more conflict but I didn't bother to check.

@yyoncho, you and I agree that ` w` as `workspace` is  the best one  but it will require a minor change from `lsp` as I mentioned above.

**Comment by yyoncho** [comment]
@thanhvg I will try to take care next week. I am very short on time.

**Comment by thanhvg** [comment]
Take your time we all work here for free :)

**Comment by thanhvg** [comment]
Done my part, back to you @smile13241324 and the gang.

File: layers/+tools/lsp/README.org [hunk 5]
```diff
@@ -320,6 +321,24 @@ whereas with ~lsp-navigation~ set to ~'peek~, this is equivalent to:
 
 etc.
 
+** Bind to lsp upstreams
+Alternative to the binding shemes above. If you set ~lsp-use-upstream-bindings~
```
lebensterben, line 325: - shemes -> schemes
- This is not a complete sentence. For example, "Alternatively, you can use vanilla LSP binding scheme, by setting ~lsp-use-upstream-bindings~ to ~t~."

layers/+tools/lsp/README.org
```diff
@@ -320,6 +321,24 @@ whereas with ~lsp-navigation~ set to ~'peek~, this is equivalent to:
...
+to ~t~ then Spacemacs will bind all ~lsp-command-map~ behind ~SPC m~, ~,~ and
+~M-m~. This way the bidings will be managed by upstream ~lsp-mode~ which is
+documented at https://emacs-lsp.github.io/lsp-mode/page/keybindings/. Spacemacs
+only replaces the prefix ~s-l~ with ~SPC m~. For example:
+
+| lsp binding | Spacemacs binding                   |
+|-------------+-------------------------------------|
+| ~s-l w s~   | ~SPC m w s~ or ~, w s~ or ~M-m w s~ |
+
+As ~lsp-mode~ and has a deep integration into ~Spacemacs~. ~Spacemacs~ hackers
```
lebensterben, line 335: remove the third word `and`

lebensterben (changes requested):
Please see my comments, and also revise this PR in accordance to suggestions from @yyoncho .

thanhvg:
> Please see my comments, and also revise this PR in accordance to suggestions from @yyoncho .

- @yyoncho's suggestion is optional and I think he is fine with my solution. `lsp-enable-which-key-integration` is ok for vanilla emacs, but we want to support all kinds bindings: vanilla, evil, normal, motion, visual, evilified. So `bind-map` is the the best one, and I  use `which-key-add-keymap-based-replacements` on `lsp-command-map` which is supposed to perform better than `which-key-add-key-based-replacements` or `which-key-add-major-mode-key-based-replacements` which `lsp-enable-which-key-integration` uses. We had so much problem with `which-key-add-major-mode-key-based-replacements` in the past ( https://github.com/syl20bnr/spacemacs/issues/12455) until `which-key` introduced `which-key-add-keymap-based-replacements` (https://github.com/aam-at/spacemacs/commit/e32cece225ab0fd0af89a8ec07979391c4305a47) 

- I don't get your previous comment about keybinding conflict.  The whole point of this PR and all the discussions so far is to have `lsp` move from `s` as `server` to `w` as `workspace` to avoid conflict, and the final scheme reserved for `lsp` is:

```
       "w"   "workspaces"
       "F"   "folders"
       "="   "formatting"
       "T"   "toggle"
       "g"   "goto"
       "h"   "help"
       "r"   "refactor"
       "a"   "code actions"
       "G"   "peek"
```
Btw, it would be great if `lsp` expose this scheme as a public variable.

Before this PR, the `lsp` scheme is:
```
       "b"   "backend"
       "F"   "folders"
       "="   "formatting"
       "T"   "toggle"
       "g"   "goto"
       "h"   "help"
       "r"   "refactor"
       "a"   "code actions"
       "G"   "peek"
```
And we have no conflict. This PR will change ` "b" "backend"` to a free `"w" "workspace"`, so I don't see any extra PR required.

- For obvious errors please go ahead and make appropriate changes, that what @duianto and @smile13241324 often do to my PRs. The problem is my PR is usually buried inside my own fork, so come back to fix the PR and sync it with my fork is a PITA. I appreciate if you can do those chores for me.

Thanks

layers/+tools/lsp/funcs.el
smile13241324: Hmm good point, but we should add it in a later PR

smile13241324 (changes requested):
I have fixed the opened typos in the documentation and merged it.
However there seems to be something wrong with the function, when I activate the new feature I am getting below error message:

`LSP :: Error from the Language Server: AttributeError: 'PosixPath' object has no attribute 'endswith' (Invalid Parameters) [3 times]` 

Do you see the same @thanhvg?

smile13241324:
A fix for this should be opened on a separate PR.

thanhvg:
> I have fixed the opened typos in the documentation and merged it.
> However there seems to be something wrong with the function, when I activate the new feature I am getting below error message:
> 
> `LSP :: Error from the Language Server: AttributeError: 'PosixPath' object has no attribute 'endswith' (Invalid Parameters) [3 times]`
> 
> Do you see the same @thanhvg?
@smile13241324 
No I don't have this error, and it doesn't look like it was caused by this PR. What function did you call?  Probably you called a function that your lsp server has no implementation yet. Try updating your lsp package.

smile13241324:
I only have this when I activate `lsp-use-upstream-bindings`, not sure if this comes from this PR, I have never tried setting this particular var. It did happen with python layer running `pyls`. My package is up-to-date though my server definitely is not.

Is `pyls` working for you with lsp-use-upstream-bindings set to t? If it is it is most likely my outdated LSP server and all is cool. Just want to make sure.

layers/+tools/lsp/README.org
lebensterben: typo needs to be fixed

lebensterben:
> * I don't get your previous comment about keybinding conflict.

Nothing wrong with this PR per se. But we need to make sure other layers doesn’t have `SPC m w ` bounded to a prefix or key binding.

PR #14430, opened 2021-02-27
[editing] Added string-edit package to editing layer

yyoncho:


smile13241324:

duianto:
This caused an issue when trying to describe a key: `C-h k` `j`
>Symbol’s value as variable is void: string-edit-mode

I removed `:defer t` from `spacemacs-editing/init-string-edit`.

Commit: Fix void string-edit-mode https://github.com/syl20bnr/spacemacs/commit/6d8101c20e600c9498398d3a1c412ab5e68f86f4

Strange that `describe-key` was effected, since it seems to be unrelated to the `string-edit` package.

yyoncho:
> Strange that `describe-key` was effected, since it seems to be unrelated to the `string-edit` package.

I hit that too. I have no explanation.

PR #14432, opened 2021-02-27
[lsp] Bind `SPC p E` to `helm-lsp-diagnostics`

yyoncho:
IMHO this should be global binding, the alternative was `SPC e e` because it can
be used everywhere, e. g. from compilation buffer to find the error causing the
build breakage.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2610,6 +2610,7 @@ files (thanks to Daniel Nicolai)
 - Added package =helm-lsp=
   - ~SPC m g s~ to find symbol in current project
   - ~SPC m g S~ to find symbol in all projects
+  - ~SPC p E~ to show error list
 - Disabled =helm-lsp= keybindings when =ivy= layer enabled (Cormac Cannon)
 - Deleted =fix-lsp-company-prefix= since =company-lsp= is doing that handling.
 - Fixed a delay when declaring prefixes for mode (thanks to Thanh Vuong)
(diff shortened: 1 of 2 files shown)


CHANGELOG.develop
```diff
@@ -2605,6 +2605,7 @@ files (thanks to Daniel Nicolai)
 - Added package =helm-lsp=
   - ~SPC m g s~ to find symbol in current project
   - ~SPC m g S~ to find symbol in all projects
+  - ~SPC m t g e~ to show error list
```
smile13241324, line 2608: I think this statement is wrong, the new binding is `SPC p E` for `helm-diagnostics`.

layers/+tools/lsp/funcs.el
```diff
@@ -97,7 +97,8 @@
    ((configuration-layer/package-usedp 'helm)
     (spacemacs/set-leader-keys-for-minor-mode 'lsp-mode
       (concat prefix-char "s") #'helm-lsp-workspace-symbol
-      (concat prefix-char "S") #'helm-lsp-global-workspace-symbol))
+      (concat prefix-char "S") #'helm-lsp-global-workspace-symbol)
+    (spacemacs/set-leader-keys "pE" #'helm-lsp-diagnostics))
```
smile13241324, line 101: Do we have a similar function for ivy?

smile13241324 (changes requested):
I have just reviewed your PR and found some smaller points you should check.

CHANGELOG.develop
yyoncho: I will fix it.

layers/+tools/lsp/funcs.el
yyoncho: ATM no, but I guess soon we will have it implemented by lsp-mode team member that users ivy.

smile13241324 (approved):

PR #14433, opened 2021-02-27
mu4e: require mu4e for mu4e-org-link-support

agzam:
it looks like it still would throw a message:

"mu4e-org-store-link: Please load mu4e before mu4e-org" if attempted to use
`org-store-link` before loading mu4e (for any kind of link, not just mu4e) 

I'm just going to bump the version number until
someone offers a better solution


lebensterben:
This should be reported and fixed in the upstream...
It seems that the [fix](https://github.com/djcb/mu/commit/13118ae173b5adeff9bc812d16e52163235468a6) doesn't work as expected.

smile13241324:
I also think this should be reported to mu4e @agzam. Bumping up the version number will only work until the next version of mu4e is released.

Anyway I will merge it as a temp fix, if you open an issue in mu4e please link to this one so we can keep track of its merge state.

smile13241324 (approved):

PR #14437, opened 2021-02-28
Add pdf-view-restore to pdf layer

dalanicolai:
Accidentally created this commit on develop. However, the addition is so small that I just PR from here.
I think restore last visited page should probably be the default in a pdf-viewer. (pdf-view-restore)[https://github.com/007kevin/pdf-view-restore] does it simple and fine.


dalanicolai:
I have decided to [recreate the PR](#14438) using a proper fix branch

PR #14438, opened 2021-02-28
Add pdf-view-restore to pdf layer

dalanicolai:
I think restore last visited page should probably be the default in a pdf-viewer. [pdf-view-restore](https://github.com/007kevin/pdf-view-restore) does it simple and fine.

### DIFF
--- a/layers/+readers/pdf/packages.el
+++ b/layers/+readers/pdf/packages.el
@@ -10,7 +10,8 @@
 ;;
 ;;; License: GPLv3
 
-(setq pdf-packages '(pdf-tools))
+(setq pdf-packages '(pdf-tools
+                     pdf-view-restore))
 
 (defun pdf/init-pdf-tools ()
   (use-package pdf-tools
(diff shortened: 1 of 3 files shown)


layers/+readers/pdf/packages.el
```diff
@@ -118,3 +119,9 @@
         "r"              'pdf-occur-revert-buffer-with-args
         "*"              'spacemacs/enter-ahs-forward
         "?"              'evil-search-backward))))
+
+(defun pdf/init-pdf-view-restore ()
+  (use-package pdf-view-restore
+    :after pdf-tools
```
smile13241324, line 125: Please lazy load.

layers/+readers/pdf/packages.el
```diff
@@ -118,3 +119,9 @@
...
+    :config
+    (add-hook 'pdf-view-mode-hook 'pdf-view-restore-mode)))
```
smile13241324, line 127: Hooks should go into the init section to trigger lazy loading.

smile13241324 (changes requested):
Nice addition to the pdf layer. However I think you need to double check the loading of your package.

layers/+readers/pdf/packages.el
dalanicolai: Ai! I should have learned that by now... :grin:

smile13241324 (approved):
Looks fine now, will be merged soon.

Issue #12451, opened 2019-06-18
emacs daemon does not read environment variable correctly


thongpv87:
I set two env variable, by `export` command and by add a new entry in `.spacemacs.env`, then reload the daemon. After that I open `emacsclient`, and use `getenv` to get the value of those env: both variables are not existed.

If I open the normal `emacs` GUI or `emacs -nw`, I can get those variable.

Emacs version: 25.2.2 - spacemacs branch: develop

LanternD:
I have the same problem. The environment variable could not be loaded properly in `emacsclient`. Then I could not find the `pyenv` command in my emacs client mode. The standalone `emacs` does not have such problem. Btw, the `exec-path-from-shell` failed to work as well. 

Emacs 26.3, develop branch, Ubuntu 18.04

LanternD:
I had a temporal solution, but not perfect.

1. Add `exec-path-from-shell` to `dotspacemacs-additional-packages`. (although `exec-path-from-shell` is already in  `'(package-selected-packages)`, it is not actually running in `emacsclient`.)
2. Add `(exec-path-from-shell-initialize)` to `dotspacemacs/user-config`, wrapped by a system type decision.
``` lisp
  (if (string-equal system-type "gnu/linux")
      (progn
        (exec-path-from-shell-initialize))
```
Now I can have my shell environment variable in `getenv`.

jZhangTk:
I have the same problem on develop in MacOS. If I start emacs without daemon mode, it works probably. But with daemon mode, it would not load from `.spacemacs.env`.

Thanks @LanternD , `exec-path-from-shell` workaround works.

jZhangTk:
Hmm, there's still issues for me with the workaround. I exported some variables in `.bash_profile`, but it doesn't seem to be picked up by `exec-path-from-shell` unless I use `exec-path-from-shell-copy-env`. Updating `PATH` variable works fine though.

ccruquan:
A other solution with recall spacemacs/load-spacemacs-env. add under code to dotspacemacs/user-config()
``` lisp
  (spacemacs|do-after-display-system-init
   (spacemacs/load-spacemacs-env))
```

jZhangTk:
Thanks @ccruquan , that worked for me.

PR #14445, opened 2021-03-01
Core env patch

wpmcbee:
Fix issue #12451 by causing the env
file to be created and loaded when
emacs is run in daemon mode.

Also, remove a superfluous variable 
and an unnecessary option.


smile13241324:
@JAremko this looks like a controversial change to me, care to have a look?

Basically we seem to not find our env file in deamon mode, however I am not sure if this change will have side effects to non deamon users and I am not too familiar with the entire env file mechanism to be honest.

lebensterben:
Part of the bug in #12451 seems like caused by wrong settings to me.

exporting a variable in a shell only makes it available within that shell and all sub shells, hence it's found in emacs -nw.

But daemon is usually handled by system service manager, eg systemd. So that variable is not valid in any sub shell of systemd, including emacsclient.

JAremko:
@smile13241324  I don't get why it's bad to load env file multiply times? Maybe it has something to do with  portable dumper setup :thinking:

lebensterben:
Normally it's not a good idea to load shell init files for multiple times because they can be arbitrarily shell scripts, which may be slow to load or have side effects.

But current design of .spacemacs.env is not shell script, but a simple key-value storage, so there's no issue with performance.

Also, it's pure and idempotent, meaning that loading it has no side effects and always have same results.

So I think there's no big issue for that.

wpmcbee:
First of all, I stupidly conflated two separate mechanisms so the first part of my proposed change should be ignored.  Only the modifications to spacemacs/load-spacemacs-env should be considered.

Secondly, my use case is running the Cygwin version of "emacs --daemon" in Windows startup, something which is probably not very common.  This change might not be as useful for systemd or other means of starting daemon mode.

lebensterben:
> Secondly, my use case is running the Cygwin version of "emacs --daemon" in Windows startup, something which is probably not very common.  This change might not be as useful for systemd or other means of starting daemon mode.

Then you can wrap the forms in a conditional and only evaluate them in cygwin.

wpmcbee:
Seems clear this pull request is defective and should not be accepted.  The workaround from ccruquan is probably the better option.

Issue #14439, opened 2021-02-28
Keep seeing an error about `(void-function smartparens-mode)`


ilyagr:
#### Description :octocat:
I keep getting errors saying `File mode specification error: (void-function smartparens-mode)`
in the messages buffer after I load a file. Similar messages about something `smartparens`-related being `void` appear at other times as well, for instance after running some commands with `helm-M-X` (don't remember the exact wording). 

I'm not sure if they cause any harm. The message was often on the screen when something went wrong, but on further investigation, the problem was usually unrelated.

#### Reproduction guide :beetle:
- Start Emacs with `emacs -nw ~/.spacemacs`
- Open the messages buffer.

*Observed behaviour:* :eyes: :broken_heart:
The message may or may not be visible on screen after Emacs loads, but it's always in the messages buffer.

*Expected behaviour:* :heart: :smile:
No error.

**Workaround:** :smile:
Putting `(require 'smartparens)` in my `dotspacemacs/user-config` function fixes the problem. When doing so, I can no longer disable colorful parentheses, though.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 27.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 54e3cc572)
- Graphic display: nil
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(auto-completion emacs-lisp git helm html markdown multiple-cursors org treemacs)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG CAIRO SOUND GPM DBUS GSETTINGS GLIB NOTIFY INOTIFY ACL LIBSELINUX GNUTLS LIBXML2 FREETYPE HARFBUZZ M17N_FLT LIBOTF ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XDBE XIM MODULES THREADS LIBSYSTEMD JSON PDUMPER LCMS2 GMP

#### More details :paw_prints:
Here is my complete `.spacemacs`:

<details>

```
;; -*- mode: emacs-lisp; lexical-binding: t -*-
;; This file is loaded by Spacemacs at startup.
;; It must be stored in your home directory.

(defun dotspacemacs/layers ()
  "Layer configuration:
This function should only modify configuration layer settings."
  (setq-default
   ;; Base distribution to use. This is a layer contained in the directory
   ;; `+distribution'. For now available distributions are `spacemacs-base'
[...]

PR #14446, opened 2021-03-02
Fix `(void-function smartparens-mode)`

ilyagr:
Fixes https://github.com/syl20bnr/spacemacs/issues/14439.

This fixes two+ problems I had, described below. 

1) When loading `emacs -nw ~/.spacemacs`, I would get a message 
`File mode specification error: (void-function smartparens-mode)`.

2) When pressing `M-:` to call `eval-expression`, it would not work.
Instead it would show 
`Symbol’s function definition is void: smartparens-mode` 
in the minibuffer.

3) I think this happened at other random times as well. From the code,
  you can get a sense `smartparens-mode` is getting loaded at strange
  times.

Fair warning: I'm not sure why I had this problem (some others claim 
they didn't experience it), when exactly is the right time to load various
types of code with `use-package`, or whether this create any new problems 
(though it looks pretty safe to me). It does seem to make things work for 
me, though.

### DIFF
--- a/layers/+spacemacs/spacemacs-editing/packages.el
+++ b/layers/+spacemacs/spacemacs-editing/packages.el
@@ -393,12 +393,6 @@
             sp-highlight-pair-overlay nil
             sp-highlight-wrap-overlay nil
             sp-highlight-wrap-tag-overlay nil)
-      (spacemacs/add-to-hooks (if dotspacemacs-smartparens-strict-mode
-                                  'smartparens-strict-mode
-                                'smartparens-mode)
-                              '(prog-mode-hook comint-mode-hook))
-      ;; enable smartparens-mode in `eval-expression'
-      (add-hook 'minibuffer-setup-hook 'spacemacs//conditionally-enable-smartparens-mode)
       ;; toggles
       (spacemacs|add-toggle smartparens
         :mode smartparens-mode
@@ -415,6 +409,14 @@
     :config
     (progn
       (require 'smartparens-config)
+
+      (spacemacs/add-to-hooks (if dotspacemacs-smartparens-strict-mode
+                                  'smartparens-strict-mode
+                                'smartparens-mode)
+                              '(prog-mode-hook comint-mode-hook))
+      ;; enable smartparens-mode in `eval-expression'
+      (add-hook 'minibuffer-setup-hook 'spacemacs//conditionally-enable-smartparens-mode)
+
       (spacemacs|diminish smartparens-mode " ⓟ" " p")
       (spacemacs//adaptive-smartparent-pair-overlay-face)
       (add-hook 'spacemacs-post-theme-change-hook


smile13241324:
I have just checked your PR, however I am not yet sure that this is a proper solution. Basically what you have done is you have moved the mode activation to until the mode has been completely loaded, this will happen after the first call of one of the autoloaded functions which could be too late.

The root issue here is why is calling one of these hooks not causing the package to be autoloaded. I'll have a look at this today and see why the autoload is not working.

ilyagr:
Thanks for taking a look, @smile13241324 !

I've been using this fix of mine for a few weeks. It helps a lot. In particular, it fixes the worst instance of this problem (`M-:` didn't work). However, it is not a complete fix. I still get the same error when opening a Markdown file or an org file. 

If you can fix it better, I'd appreciate it!

In case it's any help, here's the stacktrace I got when opening a markdown file after putting ` (defun smartparens-mode () (debug)) ` in my `user-init`.

```
Debugger entered: nil
  smartparens-mode()
  run-hooks(change-major-mode-after-body-hook text-mode-hook markdown-mode-hook)
  apply(run-hooks (change-major-mode-after-body-hook text-mode-hook markdown-mode-hook))
  run-mode-hooks(markdown-mode-hook)
  markdown-mode()
  set-auto-mode-0(markdown-mode nil)
  set-auto-mode()
  normal-mode(t)
  after-find-file(nil t)
  find-file-noselect-1(#<buffer README.md> "~/stow/dotfiles/README.md" nil nil "~/stow/dotfiles/README.md" (919706 48))
  find-file-noselect("/home/ilyagr/stow/dotfiles/README.md" nil nil nil)
  find-file("/home/ilyagr/stow/dotfiles/README.md")
  recentf-open-most-recent-file(3)
  recentf-open-most-recent-file-3()
  funcall-interactively(recentf-open-most-recent-file-3)
  call-interactively(recentf-open-most-recent-file-3 nil nil)
  command-execute(recentf-open-most-recent-file-3)
```

smile13241324:
I have tried to reproduce this issue you are seeing with smartparens, with your original dotfile but still this is just working on my side. I have double checked the autoloading but running any of the mode hooks would activate the mode effectively loading it.

What I have noticed is that part of the custom code in your `user-load` is causing a lot of random errors especially with helm. After I have removed that code the issues were gone though.

Can you retry without this customer `user-load` code? If the issue still persist, please try to remove the "defer" and "commands" keyword from `spacemacs-editing/init-smartparens` function, if this does not solve the issue it is not related to autoloading.

smile13241324:
Given that this PR is open for quite some time and as of now you can disable smartparens completely via the dotfile it is very unlikely that this is going to get merged. Therefore I would like to close this.

ilyagr:
Sounds good, thanks for looking it over. The option to disable smartparens is great to have.

Issue #14431, opened 2021-02-27
Binding `y` and `Y` in evil-evilified-state-map 


yyoncho:
Is there a reason for not having y/Y bindings in evilified-state?

duianto:
I don't know what the reason was for choosing some and not other keys.

The `evil-evilified-state.el` header says:
https://github.com/syl20bnr/spacemacs/blob/6d8101c20e600c9498398d3a1c412ab5e68f86f4/layers/%2Bdistributions/spacemacs-bootstrap/local/evil-evilified-state/evil-evilified-state.el#L28-L41

These possibly useful keys are not bound either: `w`, `b`, `$` and `^`

But they, including `y`, are bound in the evilified visual state.

This seems to be where the evilified keys are defined:
https://github.com/syl20bnr/spacemacs/blob/721765dccb94fc89a51c4d3e4730e311fe70821d/layers/%2Bdistributions/spacemacs-bootstrap/local/evil-evilified-state/evil-evilified-state.el#L174-L194

yyoncho:
Unless someone is against these binding imho they should be added. For me, it is a PITA, e. g. I want to copy something in evilified view-mode and I cant, I have to use a visual selection

duianto:
Agreed, it's surprising when the evil keys don't work as expected.

I'll open a PR, someone might think of other useful keys that should be added.

It would also be useful to be able to apply text objects in visual state,
ex: `vio` for `evil-inner-symbol`

But I'm not quite sure how to bind those.

duianto:
A PR is pending:
[evilified] Add evilified normal keys: yank, navigate #14448

duianto:
`f`, `F`, `t` and `T` were also missing, they are included in the PR.

duianto:
The PR: [evilified] Add evilified normal keys: yank, navigate #14448

broke org-agenda:
Evilified normal keys breaks "t" keabinding in agenda #14484

And probably more modes.
That "solution" was to general.

The PR has been reverted.

Let's add the keys for each specific mode where they are missing.

One instance is the `*Backtrace*` buffer (`debugger-mode`).

You mentioned: `view-mode`
Do you have some reproduce steps, I'm still in `normal` `evil-state` state when I just call: `M-x` `view-mode`

I can't think of any other modes at the moment.

duianto:
The `debugger-mode` already has an evilified section without any bound keys:
[...]

PR #14448, opened 2021-03-02
[evilified] Add evilified normal keys: yank, navigate

duianto:
Fixes: Binding y and Y in evil-evilified-state-map #14431

evilified normal state is missing some useful keys:
`y` (`evil-yank`) and the common navigation keys:
`f`, `F`, `t`, `T`, `w`, `W`, `b`, `B`, `$`, `^`

They are available in evilified visual state,
but it's useful to be able to copy text
without having to enter visual state first.

Adding `y` (`evil-yank`), makes the inner (`i`) and outer (`a`),
text objects available in evilified normal state.

This also adds the text objects to evilified visual state.


yyoncho (approved):

smile13241324 (approved):
Cool, will be merged soon.

duianto:
This broke org-agenda (probably more modes).

I reverted it.

We might have to bind these keys for each specific mode where they are missing.

PR #14449, opened 2021-03-02
Replace move-text with drag-stuff

duianto:
Just like: https://github.com/emacsfodder/move-text

https://github.com/rejeep/drag-stuff.el
also drags one or more (region) lines up or down.

But it also allows for dragging left and right (across end of lines):
- a word: changing place with the next or previous word.
- a region: moving it one character at a time to the left or right.

Added a new key binding: `SPC x .`
that opens the:
```
Drag Stuff Transient State
[k/K] up    [h/H] left   [q] quit
[j/J] down  [l/L] right
```

The `move-text` package isn't removed, even though it isn't used anymore in
Spacemacs.

Because the `evil-unimpaired` elpa directory is generated from the local
Spacemacs `evil-unimpaired.el` file.
https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Bspacemacs/spacemacs-evil/local/evil-unimpaired/evil-unimpaired.el

Therefore the `evil-unimpaired` key bindings `[e` and `]e` still call the
`move-text` commands.

Until the `evil-unimpaired` elpa directory has been removed and regenerated by
restarting Spacemacs.

Then they will call the new `drag-stuff` commands.

I don't know if/when the `move-text` package can be removed in the future.


lebensterben:
A question not directly related:
Is there any way to move the selection out of enclosing scope, e.g. {} () etc.?

duianto:
`move-text-up` almost gets there, but the indentation isn't correct, and the line above the if block is moved to the end of the if blocks last line.

Before:
```
(foo arg1 arg2
     (bar arg3
          (if arg4
              (baz)
            (xyz))))
```
With the if block selected: `M-x` `move-text-up` or `SPC x K` (before this PR has been applied)
```
(foo arg1 arg2
(if arg4
              (baz)
            (xyz))     (bar arg3
          ))
```
`move-text-up` again.
```
(if arg4
              (baz)
            (xyz))(foo arg1 arg2
     (bar arg3
          ))
```

I haven't tried the it, but this package came up:
https://github.com/mrbig033/cool-moves
when searching for `emacs drag sexp`.

It has commands called: `sexp-forward` and `sexp-backward`

They might be able to move the if block correctly.

smile13241324 (approved):

duianto:
I'm not sure how I thought that the `evil-unimpaired` package gets updated.

But I see now that it's elpa directory: `evil-unimpaired-20210314.80252`
is generated from the local version:
`layers/+spacemacs/spacemacs-evil/local/evil-unimpaired/evil-unimpaired.el`

`evil-unimpaired` was updated automatically when checking for package updates.

Now `move-text` isn't used anywhere in Spacemacs.
I'll open a PR to remove it.

PR #14450, opened 2021-03-03
git: adds orgit-forge so org links can be stored from forge topics

agzam:
Do we need the `:toggle` here as well, as it's [done for](https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Bsource-control/git/packages.el#L18) `forge` or this is good?

### DIFF
--- a/layers/+source-control/git/packages.el
+++ b/layers/+source-control/git/packages.el
@@ -34,7 +34,7 @@
         magit-svn
         org
         (orgit :requires org)
-        (orgit-forge :requires org)
+        (orgit-forge :requires (org forge))
         smeargle
         transient))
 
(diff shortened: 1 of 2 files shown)


layers/+source-control/git/packages.el
```diff
@@ -34,6 +34,7 @@
         magit-svn
         org
         (orgit :requires org)
+        (orgit-forge :requires org)
```
lebensterben, line 37: `(orgit-forge :requires (org forge))` should be sufficient to handle the dependency and then we don't need a toggle.

agzam: Pushed a commit with the suggested change

smile13241324 (approved):

PR #14451, opened 2021-03-03
[haskell] Add a binding to change the target for the session

magthe:
Using the correct target for the Haskell session is required when working on a
project with more than one target defined in the Cabal file.

### DIFF
--- a/layers/+lang/haskell/packages.el
+++ b/layers/+lang/haskell/packages.el
@@ -178,6 +178,7 @@
           "sc"  'haskell-interactive-mode-clear
           "sS"  'spacemacs/haskell-interactive-bring
           "ss"  'haskell-interactive-switch
+          "st"  'haskell-session-change-target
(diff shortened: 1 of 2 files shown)


Kazark (approved):

layers/+lang/haskell/packages.el
```diff
@@ -178,6 +178,7 @@
           "sc"  'haskell-interactive-mode-clear
           "sS"  'spacemacs/haskell-interactive-bring
           "ss"  'haskell-interactive-switch
+          "st"  'haskell-session-change-target
```
smile13241324, line 181: New bindings must be documented in the layer readme.

smile13241324 (changes requested):
This is a sensible addition to the layer however this must also be documented in the org file of the layer.

smile13241324 (approved):
Now it looks good, will be merged in a minute.

PR #5305, opened 2016-02-29
Fix inline code blocks

johankj:
Add zero-width space because emphasis blocks can't start/end with
a comma, an apostrophe or a quote.

PR #14454, opened 2021-03-03
Delete unexpected chars within pairs of ~

rayw000:
Unexpected chars break style of quoted key bindings in documents. For example: https://develop.spacemacs.org/layers/+tools/shell/README.html#key-bindings

image

More similar commits will come later after this one gets approved.

### DIFF
--- a/layers/+tools/shell/README.org
+++ b/layers/+tools/shell/README.org
@@ -125,7 +125,7 @@ to the following variables:
     '((shell :variables shell-default-shell 'eshell)))
 #+END_SRC
 
-The default shell is quickly accessible via a the default shortcut key ~SPC '​~.
+The default shell is quickly accessible via a the default shortcut key ~SPC '~.
 
 ** Default shell position, width, and height
 It is possible to choose where the shell should pop up by setting the variable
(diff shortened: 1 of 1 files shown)


duianto:
The `ZERO WIDTH SPACE` character before the closing tilde seems to have been added to hide the tilde characters around the key bindings when the readme is read on GitHub.

Minor fix of the rendering of key binding #2699
Fix inline code block #5297
Fix inline code blocks #5305

That's why the tilde characters are not visible on the current Spacemacs shell readme page:
https://github.com/syl20bnr/spacemacs/tree/develop/layers/%2Btools/shell#key-bindings

But they are visible on your fork (this PR):
https://github.com/rayw000/spacemacs/tree/delete-unexpected-chars-within-wave-pairs/layers/%2Btools/shell

It's been suggested to remove them before:
Fix formatting in Latex layer README.org #12704

The source of the issue seems to be that the tool that GitHub uses to parse org files [org-ruby](https://github.com/wallyqs/org-ruby), doesn't handle quotes, commas, etc. before a tilde.

It's been reported upstream: Aug 13, 2020
https://github.com/wallyqs/org-ruby/issues/36#issuecomment-673423916

>I reported this issue to the Org mode developers and the spec is now updated and allows any characters but whitespace in the border - https://orgmode.org/worg/dev/org-syntax.html#Emphasis_Markers.

The latest org-ruby commit is from May 19, 2020, so they might not have updated to handle the latest org mode syntax.

rayw000:
Thank you @duianto 

I've read all you posted. No wonder every `ZWS` appears before a tilde.

How would you think if I leave this PR opened before Github upgrades its `org-ruby` to the latest version?

duianto:
Sure.
They might be working on update.
Update gem
https://github.com/wallyqs/org-ruby/issues/86

PR #14456, opened 2021-03-05
Provide a switch to auto run mu4e at emacs startup

rayw000:
Currently `mu4e` does not run automatically at startup. Users will not get email notification until they run `mu4e` command explicitly. This PR provide a switch to make `mu4e` auto run in background at emacs startup.

### DIFF
--- a/layers/+email/mu4e/config.el
+++ b/layers/+email/mu4e/config.el
@@ -51,5 +51,8 @@
 (defvar mu4e-org-compose-support nil
   "If non-nil org-mu4e is configured.")
 
+(defvar mu4e-autorun-background-at-startup nil
+  "If non-nil, mu4e will automatically run in background at emacs startup.")
+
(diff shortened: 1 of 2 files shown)


lebensterben:
https://www.reddit.com/r/spacemacs/comments/ch2txa/comment/eutj6zo

How about using these settings?
In particular, `(mu4e t)` instead of explicitly starting mu4e.

rayw000:
@lebensterben 

The link you posted is very closed to what I'm talking about. The only difference is it uses `(mu4e t)` but mine uses `(mu4e~start)`. `(mu4e t)` is equivalent to `(mu4e~start)`.

I'd like to change `(mu4e~start)` to `(mu4e t)`, since function `mu4e` is autoloaded.

Thanks for your advice!

smile13241324 (approved):
Looks good.

PR #14457, opened 2021-03-05
Add note on which jar to use for groovy

peixian:
The build system for https://github.com/prominic/groovy-language-server generates two different jars, this commit adds a note on which jar to use.


smile13241324 (approved):
Ups, could have mentioned this in the first place, I am just so used to working with fat jars that I have plainly forgotten that a thin one is created too.

Thanks for clarifying this.

PR #14458, opened 2021-03-06
[ess] add keybinding for ess-quit

benswift:
Add a keybinding (alongside all the other `SPC m s` ones) to quit the current inferior process REPL.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -137,7 +137,7 @@ the [[file:CHANGELOG.org][CHANGELOG.org]] file.
 - Add ~J/K~ for scrolling down/up (thanks to Daniel Nicolai)
 ***** ESS
 - ESS key bindings were re-organised in the following list
-  (thanks to Guido Kraemer, Yi Liu, and Jack Kamm):
+  (thanks to Guido Kraemer, Yi Liu, Jack Kamm and Ben Swift):
   - ~SPC m ,~ for =ess-eval-region-or-function-or-paragraph-and-step=
   - ~SPC m d~ for =ess-dev-map=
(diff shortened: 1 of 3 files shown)


benswift:
Hmm, looking at the log for that particular action, it doesn't look like the issue is in the PR. Do you want me to re-run the actions?

lebensterben:
It's too easy to hit `, q` by accident. The REPL very often has unsaved results which could be expensive to get. `, Q` is safer.

benswift:
Yeah, ok, fair enough. R at least does give a `Save workspace image? [y/n/c]:` prompt (so that if you hit it accidentally you wouldn't necessarily lose the workspace) but I think that <kbd>,</kbd> <kbd>Q</kbd> is a good option anyway. I've updated the PR.

lebensterben:
By default yes. But there are a few ways to disable this prompt.

benswift:
Any idea why the CI checks are failing just for `macos-latest, 27.1, layers/+distribution/spacemacs-base`? To my eyes it looks like the issue is in the boilerplate rather than in my PR, but I could be wrong about that---I don't understand the details of how it all works.

lebensterben:
I don't think it's caused by your PR.

benswift:
Yep, ok. Well, my updated PR has the requested change from <kbd>,</kbd> <kbd>q</kbd> to <kbd></kbd>, <kbd>Q</kbd>. Does that look ok?

smile13241324 (approved):
Nice addition, will be merged soon. Yeah CI has some issues lately, I will post a comment on the maintainer chat about this.

PR #14463, opened 2021-03-07
[evilified] Restore evil-surround

duianto:
Enable `evil-surround-mode` when exiting evilified state.
If `evil-surround-mode` was enabled.

Currently it only disables `evil-surround-mode`.

This was discovered while exiting `edebug-mode`.

This also removes the unused variable:
`evilified-state--evil-surround`

It might have been intended for this purpose.

Otherwise it can be added back when
a use case for it is introduced.


smile13241324 (approved):
Looks good.

PR #14466, opened 2021-03-09
[core] fix warnings about overwriting by aliasing

Kazark:
Fixes this warning which annoying pops up a warning buffer on the launching
IELM:

> Warning (defvaralias): Overwriting value of ‘inferior-emacs-lisp-mode-hook’ by aliasing to ‘ielm-mode-hook’

On the one hand, this change scares me, because it impacts a lot of modes and I
have no idea how to regression test it. Someone with deeper Spacemacs knowledge
should be able to reassure me or tell me why this is bad.

On the other hand, I cannot see how it makes sense to use a mode's hook before
the mode has even been loaded; though the problem here only seems to emerge due
to the use of an alias, in my case for `ielm-mode`, which adds
`inferior-emacs-lisp-mode-hook` as an alias for `ielm-mode-hook`.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -630,6 +630,7 @@ Other:
   - Add toggles for persistent which-key keymaps, (full)-major/minor mode
     keymaps and top-level keymap, under prefix ~SPC t k~ (thanks to Daniel Nicolai)
 - Fixes:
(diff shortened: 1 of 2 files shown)


Kazark:
This is not the right way to fix this. It breaks goto definition. `add-hook` is meant to work even with variables that aren't bound yet.

PR #14468, opened 2021-03-10
Fixes error if parsing ipython dev version

Corentin-pro:
ipython versions are usually in the form `X.X.X` but a dev version is installed it is printed as `X.X.X.dev` this makes an error in the python layer initialization.

This is useful as it has been months since `ipython` is not yet released as version 8 while `jedi` moved to version 0.18. The completion API used by `ipython` version 7 is deprecated for more than a year (for `jedi` version 0.17.2 it is still working with deprecation warnings).

Since `jedi` 0.18 the deprecation warning is removed and while spacemacs will correctly complete with `jedi` 0.18 ipython will be broken (crash).

The best course of action is to clone the master branch of https://github.com/ipython/ipython and install it (just `pip install -e .` from the cloned folder). So both ipython command and spacemacs can work.


smile13241324 (approved):
This is a pragmatic workaround lets merge it.

PR #14470, opened 2021-03-10
[Scala] Update to default behaviour

EstebanMarin:
When working long hours with Metals and saving and moving between buffers. This feature of automatically showing and writing in buffers breaks HELM, and buffer management stops, rendering Emacs useless. Furthermore, this feature is not that useful. In VIM LSP is off by default. The purpose of this is to make the experience of Scala Developers with Metals a nice one, rather than having Emacs breaking all the time, as has been happening to me. This has saved me hours.

Cheers!

### DIFF
--- a/layers/+lang/scala/README.org
+++ b/layers/+lang/scala/README.org
@@ -61,6 +61,12 @@ where =0.9.8= can be replaced with the current version of [[https://scalameta.or
     -o /usr/local/bin/metals-emacs -f
 #+END_SRC
 
+Notice that the layer by default overwrites Metals-scala tree view to nil
+#+begin_example elisp
+(defvar scala-auto-treeview nil
+#+end_example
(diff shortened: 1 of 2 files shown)


Kazark (approved):
I absolutely agree that this is a better default. While some make like this UI, and that's fine, it is rather intrusive; it feels more like VSCode than Emacs; and as noted, even if this treeview works well with other LSP projects, it does not seem to play well with Metals in particular. 👍

smile13241324 (changes requested):
I agree, if this function is not working reliably and even makes emacs freeze, it should be disabled by default.

This default change however must be documented in the readme, i.e. the current instructions to disable that feature must be changed to enabling it and a short warning should be added that this could cause performance issues.

EstebanMarin:
@smile13241324 I added the description in the `README.md` file. Please let me know if you want more changes.

smile13241324:

Issue #14471, opened 2021-03-11
incorrect order number in recent files on home buffer


winsphinx:
on commit https://github.com/syl20bnr/spacemacs/pull/14326/commits/13577927640db77d9c61cbdfadb0a78337e53d42  home buffer supports number order in recent files. That's a pretty work. 

However, I met such an issue -- when i open a file (eg No.3) and edit it and close it (with key spc b d), now emacs returns back to home buffer. Then i try to open it again i press key 3, another file is opened. The reason is that the No3 file is last opened and is on the place of No1, but in recent files list, it keeps on No3.

Would this issue be checked?

emacs18:
Hitting `SPC b R` within the home buffer seems to reset "Recent Files' list.

In order to keep this list up to date as files are being edited, then we would need to add more code in hooks or timer based callback functions  so that the home buffer is updated. In my opinion such complication is not warranted.  The home buffer I think is presented for the convenience of quickly locating a file the first time emacs is started.  I don't think it is intended to be kept up to date after the first file is opened.

winsphinx:
Yes, refreshing home buffer can have it work correctly. and I also think that keeping updating home buffer is not a good idea. Howerver, i think that keep the file list *NOT* updating. 
I don't understand how the plugins works, but i guess there would be a list stored the infomation such as (1-file_a, 2-file_b, 3-file_c ...). So keep the numbers fitting the files may be a workaround.

duianto:
This issue is caused because the number keys call the commands:
`(recentf-open-most-recent-file-1)`, `-2`, etc.
https://github.com/syl20bnr/spacemacs/blob/b7cbcb5ed5741fbe289a99efaed0350812ed85ad/core/core-spacemacs-buffer.el#L973-L982

They open the files in the order that they are listed in the variable: `recentf-list`

But the list updates when a file is saved.

The home buffer still shows the order they were in when the home buffer was drawn.

duianto:
A PR has been opened with a possible solution:
Fix home buffer recent files order after file save #14472

winsphinx:
Now it runs as expected. Thanks for your great work.

PR #14474, opened 2021-03-11
Open home buffer recent files in expected order

duianto:
### problem:
Pressing a number key on the Spacemacs home buffer,
doesn't reliably open the expected recent file.

### cause:
The number keys are bound to open the files in the
order they are listed in the `recentf-list`.

The `recentf-list` is updated when a file is saved.

The home buffers recent files list is only updated
when the home buffer is created/refreshed/resized.

### solution:
Open the recent files in the order they were listed
when the home buffer was last updated.

Fixes: incorrect order number in recent files on home buffer #14471


smile13241324 (approved):
Looks good.

PR #14480, opened 2021-03-13
Add keybindings for org-roam-alias-add

KjartanOli:
Add bindings for org-roam-alias-add as `SPC a o r a` and `SPC m r a`, to make it more convenient to add aliases to org-roam files.

### DIFF
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -901,7 +901,8 @@ Headline^^            Visit entry^^               Filter^^                    Da
         "aorI" 'org-roam-insert-immediate
         "aorl" 'org-roam-buffer-toggle-display
         "aorta" 'org-roam-tag-add
-        "aortd" 'org-roam-tag-delete)
+        "aortd" 'org-roam-tag-delete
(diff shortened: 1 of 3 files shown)


layers/+emacs/org/packages.el
```diff
@@ -918,7 +919,8 @@ Headline^^            Visit entry^^               Filter^^                    Da
         "rI" 'org-roam-insert-immediate
         "rl" 'org-roam-buffer-toggle-display
         "rta" 'org-roam-tag-add
-        "rtd" 'org-roam-tag-delete))
+        "rtd" 'org-roam-tag-delete
+        "rta" 'org-roam-alias-add))
```
Kazark, line 923: Did you mean to have the `t` here? From the rest of your PR (and the mnemonics of it), that looks like a mistake.

KjartanOli: No that t is not supposed to be there. Not sure how it got there, but it should be fixed now.

Kazark (approved):

smile13241324 (approved):
Looks good.

PR #14481, opened 2021-03-13
Bind frame keys using spacemacs|spacebind

duianto:
This names the `SPC F` which-key entry.
before: `+prefix`
after: `Frames`

And adds "`...`" to the end of the descriptions,
for the keys that require additional user interactions.

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/keybindings.el
+++ b/layers/+spacemacs/spacemacs-defaults/keybindings.el
@@ -382,15 +382,18 @@
     ("y" spacemacs/copy-file-path "File path")
     ("b" spacemacs/copy-buffer-name "Buffer name")))))
 ;; frame ----------------------------------------------------------------------
-(spacemacs/set-leader-keys
-  "Ff" 'spacemacs/find-file-other-frame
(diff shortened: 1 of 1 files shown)


Kazark (approved):

lebensterben:
question: is `spacemacs/set-leader-keys` superseded by `spacemaca|spacebind`?

JAremko:
@lebensterben   Yeah it will eventually be.  I want to change syntax a bit first.  Mainly I  want to add a group (similarly to customization groups) for example:

``` elisp
(spacemacs|spacebind
 "Frames"
:spacemacs-frames
 :global
 (("F" "Frames"
   ("f" spacemacs/find-file-other-frame "Find file other frame...")
```

The group will be used in .spacemacs  to disable default bindings. Something like:
`dotspacemacs-disabled-bindings '(:spacemacs-frames :some-other-bindings)` optionally it will be `dotspacemacs-disabled-bindings 'all` that means all default bindings should be disabled.  When I finalize the API I'll write docs for it :+1:

smile13241324 (approved):
Great, looking forward to have spacebind used more frequently in time.

PR #14465, opened 2021-03-08
[org] Add org-appear support

winsphinx:
One picture is more than many words :  

image

Update PR at #14482

PR #14482, opened 2021-03-14
[org] Add org-appear support

winsphinx:
fixes for PR #14465

### DIFF
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -50,7 +50,9 @@
     (org-sticky-header :toggle org-enable-sticky-header)
     (verb :toggle org-enable-verb-support)
     (org-roam :toggle org-enable-roam-support)
-    (valign :toggle org-enable-valign)))
+    (valign :toggle org-enable-valign)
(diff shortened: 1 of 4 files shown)


smile13241324 (approved):

layers/+emacs/org/packages.el
```diff
@@ -50,7 +50,9 @@
     (org-sticky-header :toggle org-enable-sticky-header)
     (verb :toggle org-enable-verb-support)
     (org-roam :toggle org-enable-roam-support)
-    (valign :toggle org-enable-valign)))
+    (valign :toggle org-enable-valign)
+    (org-appear :location (recipe :fetcher github :repo "awth13/org-appear")
```
tko, line 54: [org-appear seems to be in MELPA](https://melpa.org/#/org-appear) -- I wonder whether it or github is preferred from where to fetch the packages?

Kazark: Good question.

winsphinx: If `location` is set, github repo is preferred. So I think the one on melpa should be preferred.

PR #14489, opened 2021-03-15
 rust: Improved LSP Rust server switch functionality

lebensterben:
Currently, when `lsp` backend is used for `rust` layer, `SPC m s s` is bound to
`lsp-rust-switch-server`, which temporarily changes the preferred LSP server
for `rust-mode`, but it DOES NOT switch the running LSP server. Users need to
call `lsp-workspace-restart` to finish the switching.

This PR introduced a new function `spacemacs//lsp-rust-switch-server`, which is
a wrapper for the two aforementioned functions


duianto:

PR #14490, opened 2021-03-16
Move known temporary files of LSP into `spacemacs-cache-directory`

rayw000:
Move known temporary files of LSP into `spacemacs-cache-directory`.

Original post:
> Ignore auto-generated `.lsp-eslint-choices` file.
> (Or, would it be better if we move this file into `spacemacs-cache-directory`/`lsp` by setting variable `lsp-eslint-library-choices-file`?)

### DIFF
--- a/layers/+tools/lsp/packages.el
+++ b/layers/+tools/lsp/packages.el
@@ -22,6 +22,15 @@
 (defun lsp/init-lsp-mode ()
   (use-package lsp-mode
     :defer t
+    :init
+    (setq lsp-cache-dir (concat spacemacs-cache-directory "lsp/")
+          lsp-server-install-dir lsp-cache-dir
+          lsp-session-file (concat lsp-cache-dir (file-name-nondirectory ".lsp-session-v1"))
+          lsp-eslint-library-choices-file (concat lsp-cache-dir ".lsp-eslint-choices")
+          lsp-yaml-schema-store-local-db (concat lsp-cache-dir "lsp-yaml-schemas.json")
+          lsp-vetur-global-snippets-dir (concat lsp-cache-dir ".snippets/vetur")
+          ;; If you find something else should be ignored, you could also set them here
+          )
     :config
     (progn
       (spacemacs/lsp-bind-keys)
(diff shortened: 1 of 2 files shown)


smile13241324:
I think moving the file would be the better choice, than it should be automatically ignored.

rayw000:
> I think moving the file would be the better choice, than it should be automatically ignored.

I have the same idea, just a little bit unconfident when I saw `.lsp-session-*` in `.gitignore`. I'll handle this kind of `.lsp-*` files in a later PR.

layers/+tools/lsp/packages.el
```diff
@@ -22,6 +22,15 @@
...
   (use-package lsp-mode
     :defer t
+    :init
+    (setq lsp-cache-dir (concat spacemacs-cache-directory "lsp/")
+          lsp-server-install-dir lsp-cache-dir
+          lsp-session-file (concat lsp-cache-dir (file-name-nondirectory ".lsp-session-v1"))
+          lsp-eslint-library-choices-file (concat lsp-cache-dir ".lsp-eslint-choices")
+          lsp-yaml-schema-store-local-db (concat lsp-cache-dir "lsp-yaml-schemas.json")
+          lsp-vetur-global-snippets-dir (concat lsp-cache-dir ".snippets/vetur")
+          ;; If you find something else should be ignored, you could also set them here
```
smile13241324, line 32: I don't think that setting these variables in the central LSP layer is good choice. In the end we would need to maintain one row for every layer (given that each will one day have LSP support).

I would rather prefer a solution like for the auto-complete layer. This is let the affected layers inject their individual config in a `post-init-lsp-mode` function instead. By this we keep the layer specific part localised and can still define general settings like the `lsp-cache-dir` in one place.

smile13241324 (changes requested):

layers/+tools/lsp/packages.el
rayw000: Im glad to have a look at how `auto-completion` does this job.

smile13241324:

PR #14491, opened 2021-03-16
Add to docs, replacing a layer package with a local version

dalanicolai:
When contributing to packages it is often handy to replace a package, already
installed and configured by some layer with a local version of the package (i.e.
a git repo). This commit adds instructions for how to achieve that.

### DIFF
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -23,6 +23,7 @@
       - [[#declaration][Declaration]]
       - [[#initialization][Initialization]]
       - [[#exclusion][Exclusion]]
+      - [[#overriding-a-layer-package][Overriding a layer package]]
     - [[#without-a-layer][Without a layer]]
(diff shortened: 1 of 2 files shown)


doc/DOCUMENTATION.org
```diff
@@ -435,6 +436,9 @@ Example:
   (setq <layer>-excluded-packages '(package1 package2 ...)
 #+END_SRC
 
+**** Replace layer package with local version (e.g. for contributing to a package)
```
lebensterben, line 439: I think you can take the parenthesis out of the section title.
So the auto-generated TOC entry won't be so long.

lebensterben: Actually this can be named as
`Overriding a layer package`

dalanicolai: Thanks! Good feedback. I agree this would be comparably informative, while looking better

duianto:

PR #14494, opened 2021-03-17
Prefer downloading org-appear from melpa

winsphinx:
Since org-appear is [on melpa](https://melpa.org/#/org-appear), it should be preferred.


Kazark (approved):

duianto:

Issue #8363, opened 2017-02-14
Suspicious error messages when loading help system


nalzok:
#### Description :octocat:
Suspicious error messages when loading help system

#### Reproduction guide :beetle:
- Start Emacs
- Press `<SPC> h <SPC>` start help system

*Observed behaviour:* :eyes: :broken_heart:
The `*Messages*` buffer shows errors:

1. `org-mode fontification error in #<buffer DOCUMENTATION.org> at 381`
2. `Eager macro-expansion failure: (void-variable erc-spacemacs-layout-name)`
3. `Eager macro-expansion failure: (void-variable rcirc-spacemacs-layout-name)`

Despite the error messages, everything seems to work well. I can read the document and the help system didn't crash.

*Expected behaviour:* :heart: :smile:
No error messages :)

#### System Info :computer:
- OS: darwin
- Emacs: 26.0.50.1
- Spacemacs: 0.200.7
- Spacemacs branch: master (rev. c2774bc)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(auto-completion emacs-lisp helm osx scheme
                 (shell :variables shell-default-shell 'ansi-term shell-default-term-shell "/bin/bash" shell-default-position 'bottom shell-default-height 50)
                 syntax-checking)
```
- System configuration features: JPEG NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS

#### Backtrace :paw_prints:
```
<<Skipped thousands lines of "loading ...done" messages>>

Loading /Users/sunqingyao/.emacs.d/layers/+checkers/spell-checking/packages.el (source)...done
Loading /Users/sunqingyao/.emacs.d/layers/+checkers/syntax-checking/packages.el (source)...done
Loading /Users/sunqingyao/.emacs.d/layers/+chat/erc/packages.el (source)...
Eager macro-expansion failure: (void-variable erc-spacemacs-layout-name)
Loading /Users/sunqingyao/.emacs.d/layers/+chat/erc/packages.el (source)...done
Loading /Users/sunqingyao/.emacs.d/layers/+chat/jabber/packages.el (source)...done
Loading /Users/sunqingyao/.emacs.d/layers/+chat/rcirc/packages.el (source)...
Eager macro-expansion failure: (void-variable rcirc-spacemacs-layout-name)
[...]

PR #14495, opened 2021-03-17
Ensure required arguments of macro spacemacs|define-custom-layout are all defined.

rayw000:
### What happened

Macro `spacemacs|define-custom-layout` requires `name` argument and `:binding` property **bound** if they are symbols. If one of them is not defined before calling this macro, an error message
```
Eager macro-expansion failure: (void-variable <layer name>-spacemacs-layout-name)
```
will come out in `*Messages*` buffer.

### What changed

Macro `spacemacs|define-custom-layout` will check if both `name` and `binding-prop` are not unbound symbols, otherwise, do nothing.

Fix: #8363


duianto:

PR #14496, opened 2021-03-17
[java] only download meghanada package if needed

thanhvg:
currently toggle `:if` for meghanada with `use-package` only tells `use-package`
not to load `meghanada` but spacemacs still downloads `meghanada`.

also we shouldn't care about emacs 25 anymore.

### DIFF
--- a/layers/+lang/java/packages.el
+++ b/layers/+lang/java/packages.el
@@ -19,7 +19,7 @@
     helm-gtags
     (java-mode :location built-in)
     maven-test-mode
-    (meghanada :toggle (not (version< emacs-version "25.1")))
+    (meghanada :toggle (eq (spacemacs//java-backend) 'meghanada))
     mvn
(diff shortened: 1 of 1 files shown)


smile13241324:
Right now the minimum supported emacs version is still 25.1 for Spacemacs, I'll ask the others if we can bump it up to 26.3 I think this should be possible now a days.

smile13241324 (approved):
Not installing meghanada when not used is a sensible addition. Will be merged in some minutes.

smile13241324:

thanhvg:
@smile13241324 hold on, this PR will break with new change in dev, please change the line 22 to 

```
    (meghanada :toggle (eq java-backend 'meghanada))
```

smile13241324:
@thanhvg ups, thanks for pointing this out, should be fixed now.

PR #14497, opened 2021-03-17
Removed legacy codes for emacs-version < 25.1

lebensterben:
#14496 covers Java layer and it's therefore not included in this PR.


smile13241324:
Right now the minimum supported emacs version is still 25.1 for Spacemacs, I'll ask the others if we can bump it up to 26.3 I think this should be possible now a days.

lebensterben:
@smile13241324 
We could use `debian`'s stable version number as our minimum supported version.
The reason is it's a distro well-known for being barely up-to-date...
Currently it's 26.1 on `debian`.
https://tracker.debian.org/pkg/emacs

smile13241324:
@syl20bnr what's your opinion about this, do you think it would be ok to move up our minimum version to at least 26.1? We could remove some hacks this way.

lebensterben:
I'm going to merge this first.

I'll make another PR to deal with 26.1

PR #14498, opened 2021-03-17
In SPC h d P, print :toggle expression using Elisp syntax

bcc32:
This prints out an expression in valid Lisp syntax, so that, e.g.,
strings in the expression are quoted.

For example, I had the following in `dotspacemacs-additional-packages`:

```
(direnv :toggle (executable-find "direnv"))
```

Without this patch, the `:toggle` expression is printed as `(executable-find direnv)`, which looks like it is reading a variable called `direnv`.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -993,7 +993,7 @@ a new object."
         (princ (if (cfgl-package-toggled-p pkg t) "on" "off"))
         (princ " because the following expression evaluates to ")
         (princ (if (cfgl-package-toggled-p pkg t) "t:\n" "nil:\n"))
-        (princ (oref pkg :toggle))
+        (prin1 (oref pkg :toggle))
         (princ "\n"))
       (when (oref pkg :requires)
         (princ "\nThis package requires the following packages: ")


lebensterben:
This is the intended behaviour of `princ`.
> This function is intended to produce output that is readable by people, not by read, so it doesn’t insert quoting characters and doesn’t put double-quotes around the contents of strings. It does not add any spacing between calls.

bcc32:
I understand this is the intended behavior of `princ`.  This PR replaces it with `prin1` because it makes more sense to output an expression in Lisp syntax, as it would appear in somebody's spacemacs init file.

lebensterben:
This is not the intended usage of `dotspacemacs-additional-package`.
> If you need some configuration for these
packages then consider to create a layer, you can also put the
configuration in `dotspacemacs/user-config'.

bcc32:
I'm not sure how your comment applies to this pull request.  My change is merely to how the package description is printed.

You can see an example without using `dotspacemacs-additional-package`, for example, press: `SPC h d P bracketed-paste RET`.

I see:

> A toggle is defined for this package, it is currently off because the following expression evaluates to nil:
(version<= emacs-version 25.0.92)

But the actual expression in layers/+spacemacs/spacemacs-editing/packages.el is:
```
        (bracketed-paste :toggle (version<= emacs-version "25.0.92"))
```

This PR would change the `SPC h d P` output to preserve the quote characters around the version string.

lebensterben:
You are not suppose to pass complex forms to `dotspacemacs-additional-package`, which is exactly what I quoted
> If you need some configuration for these
packages then consider to create a layer, you can also put the
configuration in `dotspacemacs/user-config'.

Your issue arises from the fact that you're abusing `dotspacemacs-additional-package`.

bcc32:
Again, I don't see how your comment applies to whether or not this pull request should be merged.  As I stated in my previous comment,

> You can see an example without using dotspacemacs-additional-package, for example, press: SPC h d P bracketed-paste RET.

So, even if my `dotspacemacs-additional-packages` is completely empty, the problem I am describing occurs, where the syntax displayed by `SPC h d P` does not match the Lisp syntax used in the packages list, whether that package list is `dotspacemacs-addditional-packages` or a layer package list like `spacemacs-editing-packages`.

bcc32:
By the way:

> You are not suppose to pass complex forms to dotspacemacs-additional-package

I believe that's not correct.  `configuration-layer/make-packages-from-dotfile` passes the additional package forms to `configuration-layer/make-package`, which is capable of handling the `:toggle` and `:location` keywords, among others.

In fact, the documentation for `dotspacemacs-additional-packages` also mentions:

> To use a local version of a package, use the
   ;; `:location' property: '(your-package :location "~/path/to/your-package/")

In the sentence you quoted, "configuration" refers to the `use-package` calls and related `layer/init-<package>` functions.  It does not exclude the possibility of using `:location` (or in this case `:toggle`) in `dotspacemacs-additional-packages`, as I have been doing for a long time without issue.

lebensterben:
> So, even if my `dotspacemacs-additional-packages` is completely empty, the problem I am describing occurs, where the syntax displayed by `SPC h d P` does not match the Lisp syntax used in the packages list, whether that package list is `dotspacemacs-addditional-packages` or a layer package list like `spacemacs-editing-packages`.

So how'd this matter?
It only matters because you're abusing `SPC h d P`. You wish it outputs a Read syntax, but by design `configuration-layer/describe-package` is user-facing instead of for scripting
> Describe a package in the context of the configuration layer system.

If you're trying to use the output of `SPC h d P` and turns it to a Read syntax, you'd use `dotspacemacs/user-config`.

bcc32:
I'm not trying to use it for scripting.  It is merely my suggestion that describe-package should output read syntax, because I believe it would be far less confusing for users.

For example, `princ` can produce ambiguous outputs:

```elisp
(princ '("a b" "c"))
; prints (a b c)
```

With the quote marks missing, it is not possible to figure out whether the original expression contained one, two, or three strings, or indeed whether they were strings and not symbols.

I do not understand why you seem so opposed to this change?  I don't think there's any downside to using `prin1` over `princ`, and only a potential benefit in terms of clarity.

lebensterben:
I'm not against the change.
What I'm saying is it doesn't matter if all delimiters (") are removed in the output stream. It only matters if it's intended to be used as Read syntax.

Specifically, `:toggle` rarely takes arguments with complex string which can confuse the reader should " are removed.

bcc32:
> I'm not against the change.

Okay, thanks for clarifying 😄 

> Specifically, :toggle rarely takes arguments with complex string which can confuse the reader should " are removed.

Yeah, the `direnv` example from the original post is exactly the case in which I got confused by this printing behavior.  I agree that the output format is only for human consumption.

duianto:

This does clarify the description.
`SPC h d P` `bracketed-paste`

Before:
```
bracketed-paste is a package declared and configured by the layer `spacemacs-editing'.

A toggle is defined for this package, it is currently off because the following expression evaluates to nil:
(version<= emacs-version 25.0.92)
```

After:
```
bracketed-paste is a package declared and configured by the layer `spacemacs-editing'.

A toggle is defined for this package, it is currently off because the following expression evaluates to nil:
(version<= emacs-version "25.0.92")
```

PR #14506, opened 2021-03-17
[org] add keybinding for org-copy-tree as with org-cut-tree

Kazark:
Used `y` as a mnemonic for yank (in the Vim, not the Emacs, sense).


duianto:

PR #14509, opened 2021-03-18
Refactor

lebensterben:
- Refactored function
	- layers/+frameworks/react/func.el
	- layers/+lang/c-c++/layers.el
	- layers/+lang/elm/funcs.el
	- layers/+lang/julia/packages.el
	- layers/+lang/latex/funcs.el
	- layers/+lang/lua/funcs.el
	- layers/+lang/nim/funcs.el
	- layers/+lang/perl5/funcs.el
	- layers/+lang/purescript/funcs.el
	- layers/+spacemacs/spacemacs-editing/local/spacemacs-whitespace-cleanup/spacemacs-whitespace-cleanup.el
	- layers/+spacemacs/spacemacs-editing/packages.el
	- layers/+tools/cmake/funcs.el
	- layers/+tools/eaf/packages.el
	- layers/+tools/terraform/funcs.el
- Moved backend determination to `config.el`
	- layers/+lang/c-c++/config.el 
	- layers/+lang/clojure/config.el
	- layers/+lang/crystal/config.el
	- layers/+lang/elixir/config.el
	- layers/+lang/elm/config.el
	- layers/+lang/erlang/config.el
	- layers/+lang/ess/config.el
	- layers/+lang/fsharp/config.el
	- layers/+lang/go/config.el
	- layers/+lang/groovy/config.el
	- layers/+lang/haskell/config.el
	- layers/+lang/java/config.el
	- layers/+lang/javascript/config.el
	- layers/+lang/json/config.el
	- layers/+lang/julia/config.el
	- layers/+lang/kotlin/config.el
	- layers/+lang/latex/config.el
	- layers/+lang/lua/config.el
	- layers/+lang/nim/config.el
	- layers/+lang/purescript/config.el
	- layers/+lang/python/config.el
	- layers/+lang/ruby/config.el
	- layers/+lang/rust/config.el
	- layers/+lang/shell-scripts/config.el
 	- layers/+lang/sql/config.el
 	- layers/+lang/typescript/config.el
 	- layers/+lang/vimscript/config.el
 	- layers/+tools/cmake/config.el
 	- layers/+tools/terraform/config.el
- Moved function from `packages.el` to `func.el`
    - layers/+frameworks/react/packages.el
    - layers/+lang/fsharp/packages.el
- Replaced `(set (make-local-variable 'foo) bar)` with `(setq-local foo bar)`
    - core/core-dotspacemacs.el
    - core/libs/ido-vertical-mode.el
    - layers/+chat/rcirc/packages.el
    - layers/+intl/chinese/packages.el
	- layers/+lang/c-c++/funcs.el
  	- layers/+lang/javascript/funcs.el
  	- layers/+lang/typescript/funcs.el
  	- layers/+vim/vinegar/funcs.el
- Replaced `(when (not foo) bar)` with `(unless foo bar)`
    - layers/+chat/rcirc/funcs.el
    - layers/+distributions/spacemacs-bootstrap/funcs.el
    - layers/+distributions/spacemacs-bootstrap/packages.el
    - layers/+frameworks/react/funcs.el
    - layers/+fun/xkcd/packages.el
    - layers/+intl/keyboard-layout/funcs.el
 	- layers/+lang/c-c++/config.el 
 	- layers/+lang/clojure/packages.el
  	- layers/+lang/javascript/funcs.el
  	- layers/+lang/julia/funcs.el
 	- layers/+lang/typescript/funcs.el
 	- layers/+readers/dash/funcs.el
 	- layers/+spacemacs/spacemacs-defaults/funcs.el
 	- layers/+themes/colors/funcs.el
 	- layers/+tools/eaf/funcs.el
- Replaced unnecessary backquote construct with simple quotation
	- layers/+distributions/spacemacs-bootstrap/packages.el
	- layers/+frameworks/react/funcs.el
	- layers/+lang/c-c++/funcs.el
	- layers/+lang/c-c++/layers.el
	- layers/+lang/c-c++/packages.el
	- layers/+lang/crystal/funcs.el
	- layers/+lang/csharp/funcs.el
	- layers/+lang/dart/funcs.el
	- layers/+lang/elixir/funcs.el
	- layers/+lang/elm/funcs.el
 	- layers/+lang/go/funcs.el
 	- layers/+lang/groovy/funcs.el
 	- layers/+lang/haskell/funcs.el
 	- layers/+lang/java/funcs.el
 	- layers/+lang/javascript/funcs.el
 	- layers/+lang/latex/funcs.el
 	- layers/+lang/nim/funcs.el
 	- layers/+lang/perl5/funcs.el
 	- layers/+lang/python/funcs.el
 	- layers/+lang/python/packages.el
 	- layers/+lang/ruby/funcs.el
 	- layers/+lang/rust/funcs.el
 	- layers/+lang/typescript/funcs.el
 	- layers/+spacemacs/spacemacs-editing/local/spacemacs-whitespace-cleanup/spacemacs-whitespace-cleanup.el
- Replaced `pcase` form with only one-arm with `when` or `unless` form 
	- layers/+frameworks/react/funcs.el
	- layers/+frameworks/vue/funcs.el
	- layers/+lang/c-c++/funcs.el
	- layers/+lang/clojure/funcs.el
	- layers/+lang/crystal/funcs.el
	- layers/+lang/csharp/funcs.el
	- layers/+lang/elixir/funcs.el
	- layers/+lang/elixir/packages.el
 	- layers/+lang/erlang/funcs.el
 	- layers/+lang/fsharp/funcs.el
 	- layers/+lang/go/funcs.el
 	- layers/+lang/go/packages.el
 	- layers/+lang/groovy/funcs.el
 	- layers/+lang/haskell/funcs.el
 	- layers/+lang/java/funcs.el
 	- layers/+lang/java/packages.el
 	- layers/+lang/javascript/funcs.el
 	- layers/+lang/javascript/packages.el
 	- layers/+lang/json/funcs.el
 	- layers/+lang/julia/funcs.el
 	- layers/+lang/kotlin/funcs.el
 	- layers/+lang/latex/funcs.el
 	- layers/+lang/lua/funcs.el
 	- layers/+lang/perl5/funcs.el
 	- layers/+lang/perl5/packages.el
 	- layers/+lang/php/funcs.el
 	- layers/+lang/php/packages.el
 	- layers/+lang/purescript/funcs.el
 	- layers/+lang/python/funcs.el
 	- layers/+lang/python/packages.el
 	- layers/+lang/ruby/funcs.el
 	- layers/+lang/ruby/packages.el
 	- layers/+lang/rust/funcs.el
 	- layers/+lang/rust/packages.el
 	- layers/+lang/scala/packages.el
 	- layers/+lang/shell-scripts/funcs.el
 	- layers/+lang/sql/funcs.el
 	- layers/+lang/typescript/funcs.el
 	- layers/+lang/vimscript/funcs.el
 	- layers/+tools/cmake/funcs.el
 	- layers/+tools/dap/README.org
 	- layers/+tools/docker/funcs.el
 	- layers/+tools/terraform/funcs.el
- Removed unnecessary `progn` 
	- layers/+lang/c-c++/funcs.el
 	- layers/+lang/elixir/funcs.el
 	- layers/+lang/nim/funcs.el
- Replaccd `(cond ((eq foo) bar))` form with `(pcase (foo bar))`
  	- layers/+lang/javascript/funcs.el
 	- layers/+lang/lua/funcs.el
 	- layers/+lang/python/funcs.el
 	- layers/+lang/typescript/funcs.el
 	- layers/+lang/typescript/packages.el
- Replaced `(if foo bar nil)` with `(when foo bar)`
	- layers/+lang/typescript/packages.el


layers/+lang/perl5/packages.el
```diff
@@ -78,7 +78,7 @@
           (if (nth 4 (syntax-ppss))
               'font-lock-comment-face
             (if (= (- (match-end 2) (match-beginning 2)) 1)
-                (if (eq (char-after (match-beginning 3)) ?{)
+                (if (eq (char-after (match-beginning 3)) ?{})
```
lebensterben, line 81: Please help me to determine whether this modification is correct.
Without `}`, emacs complains about mismatching parentheses.

layers/+lang/rust/packages.el
```diff
@@ -64,9 +64,9 @@
   (spacemacs/counsel-gtags-define-keys-for-mode 'rust-mode))
 
 (defun rust/pre-init-dap-mode ()
-  (pcase (spacemacs//rust-backend)
-    (`lsp (add-to-list 'spacemacs--dap-supported-modes 'rust-mode)))
-  (add-hook 'rust-mode-local-vars-hook #'spacemacs//rust-setup-dap))
+  (when (eq rust-backend 'lsp)
+    (add-to-list 'spacemacs--dap-supported-modes 'rust-mode)
+    (add-hook 'rust-mode-local-vars-hook #'spacemacs//rust-setup-dap)))
```
lebensterben, line 70: Notice that this is not a typo:

Originally, the last `add-hook` form is executed unconditionally, but this is only meaningful when `lsp` backend is chosen, so it's now moved inside the `when` form.

layers/+lang/perl5/packages.el
rayw000: Hi @lebensterben 

`?<char>` is a syntax which Emacs represents characters.
```emacs-lisp
(eq ?{ (string-to-char "{"))
;; => t

(char-to-string ?{)
;; => "{"
```

So this change may not be necessary.

Kazark: I have seen the `<` function confuse it about mismatching parentheses as well. This is a problem with the Lisp mode code, I believe, and not with this line of code. The `?` syntax is a bit surprising sometimes: for example, `? ` is a space character (and I had thought Lisp syntax was space-agnostic).

lebensterben: Thanks.
`?C` is the special read syntax.
https://www.gnu.org/software/emacs/manual/html_node/elisp/Special-Read-Syntax.html#Special-Read-Syntax

I will fix this in a new commit later.
The closing parentheses is added by pareninfer I think.

lebensterben: Fixed.

smile13241324:
Whao this is a lot of changes. You did run some tests against these haven't you?

About `Moved backend determination to config.el` I personally like this idea however I am not sure why we haven't done this in the first place, lets ask the others first maybe there was a reason for not doing this in the original design.

For the rest I will make individual reviews and sent my feedback one by one.

lebensterben:
> About `Moved backend determination to config.el` I personally like this idea however I am not sure why we haven't done this in the first place, lets ask the others first maybe there was a reason for not doing this in the original design.

This is just a coincidence I believe.
The current behaviour of determining lsp backend at runtime, more specifically at every call of `spacemacs/XX-backend`, is introduced in this https://github.com/syl20bnr/spacemacs/commit/ac30247811d28e86c4b33d6150d678d5f5824330

Then when people adding LSP support to other layers, they just followed the existing pattern.
One thing to notice that we haven't played with default value of layer variables much, and we often determine/infer things in packages.el or funcs.el instead, including which bakcend/formatter to use and etc.

After we finished the migration to `defcustom` based config variables, we can review what else can be done like in this PR.

lebensterben:
@smile13241324 
Any feedback?

lebensterben:
I will rebase a while later

duianto:

sdwolfz pointed out that, since each commit describes it's changes, let's keep them separate.

duianto:
The `emacspace` bot wants to revert the `ido-vertical-mode.el` changes:
https://github.com/syl20bnr/spacemacs/pull/14573/files#diff-606d1dcd79a6fefd96a1f3b99eb4074f2f3ceb6a78f9d86c1720282926a4d46e

The bot might be checking for differences against the upstream file:
https://github.com/creichert/ido-vertical-mode.el/blob/b1659e967da0687abceca733b389ace24004fa66/ido-vertical-mode.el#L161-L162
```elisp
   ""                                                                        ; right bracket around the sole remaining completion
   ))
```

https://github.com/creichert/ido-vertical-mode.el/blob/b1659e967da0687abceca733b389ace24004fa66/ido-vertical-mode.el#L321
```elisp
  (set (make-local-variable 'truncate-lines) nil))
```

Those changes should probably be suggested upstream, then the bot will open a PR here when it detects that they have changed upstream.

hrkrshnn:
@lebensterben after this PR, I get an error saying that the variable `c-c++-backend` is null.

I think the problem is https://github.com/syl20bnr/spacemacs/blob/develop/layers/+lang/c-c++/layers.el#L24

A quick fix was to add `(defvar c-c++-backend 'lsp-clangd)` right above that line.

duianto:
I reverted the refactoring of the c-c++ layer.

I'll check the rest of the layers if they work as expected.
Sorry for not testing this more thoroughly.

duianto:
None of the other refactored layers break Spacemacs on startup.

lebensterben:
I will take a look of c-c++ layer

duianto:
I think I found the cause.
A single quote was missing in:
Before:
```elisp
(when (boundp c-c++-backend)
```
After:
```elisp
(when (boundp 'c-c++-backend)
```
in `c-c++/layers.el`

There is no void variable message on startup with the single quote added.

lebensterben:
I will review each of them once again.

duianto:
The python layer seems to have issues:
https://gitter.im/syl20bnr/spacemacs?at=6069fb9992a3431fd67544f9
>I do think you two should be careful with your current changes across all layers on dev branch. This is madness. My tide layer and and python just break. Python layer won't work with lsp. Because your new change in spacemacs//python-setup-lsp:
```elisp
(pcase python-lsp-server
                   ('mspyls 'lsp-python-ms)
                   ('pyright 'lsp-pyright))
will always return nil because of the extra quote
```
I get the intention but as user I can't check all other layers.

Could you prioritize it next.

thanhvg:
Hi, it works again I don't see any problem anymore. Just a bad timing to check out dev branch. Thanks.

sunlin7:
> The current behaviour of determining lsp backend at runtime, more specifically at every call of `spacemacs/XX-backend`, is introduced in this [ac30247](https://github.com/syl20bnr/spacemacs/commit/ac30247811d28e86c4b33d6150d678d5f5824330)

For some one enable/disable `lsp` feature from config layer and evaluate `.spacemacs` file again and again, the  `spacemacs/XX-backend` will keep behavior as expected, while current changes do **NOT** follow the config layer changes. 
Please consider the compatible for the situation that user change the config layer and "SPC e f R" to apply new settings without restart.
@lebensterben @smile13241324

lebensterben:
@sunlin7 
This is the expected behavior of this PR.
It's not a bug, but a feature.

You are still able to change to other backend by changing the value of the corresponding layer variable, in dotspacemacs or in IELM.

I don't understand how this causes any compatibility issue.

sunlin7:
Hi @lebensterben ,

The reproduct steps are:

1. Please Config `dotspacemacs-configuration-layers`  to `'(c-c++)`, restart spacemacs, now variable `c-c++-backend` is `nil`.

2. Open `.spacemacs` by press `SPC f e d`, and add `lsp` to  `dotspacemacs-configuration-layers`, then press `SPC f e R` to apply the new config layer value, the `c-c++-backend` is `nil` (restarting Spacemacs can get the new vlaue);  while former `spacemacs/XX-backend` can work with new configuration vlaue without require restarting.

lebensterben:
@sunlin7 
I don't get it.

So you'd prefer to compute the value for a missing layer variable on each initialization of a layer, than just compute it once with the only additional requirements of reloading dotspacemacs on inclusion of LSP layer?

Why would a typical user frequently enable and disable LSP layer or any layer?

sunlin7:
Okay, I got your point.

This behavior change will require:
1. Restarting spacemacse;
2. Re-creating the `spacemacs*.pdmp ` file if pdmp feature is actived.

Document update for changing configuration layer require restarting Spacemacs and update pdmp file will be helpful for some one doesn't know this behavior change:

```
doc/QUICK_START.org:81:_without restarting_ Spacemacs by pressing ~SPC f e R~.
doc/VIMUSERS.org:354:Emacs or press ~SPC f e R~. To view all layers and their documentation use
doc/DOCUMENTATION.org:494:layer from the =dotspacemacs-configuration-layers= and press ~SPC f e R~.
doc/DOCUMENTATION.org:547:To apply the modifications made in =~/.spacemacs= press ~SPC f e R~. It will
doc/DOCUMENTATION.org:567:tests are also run automatically when you synchronize with ~SPC f e R~.
....
```

lebensterben:
> This behavior change will require:
> 
> 1. Restarting spacemacse;
> 
> 2. Re-creating the `spacemacs*.pdmp ` file if pdmp feature is actived.

pdump file should be already generated everytime when you reload dotspacemacs.

So there's one single step.

Does this need to be documented? No.

Because whenever you add or remove a layer, you're supposed to reload dotspacemacs.

PR #14510, opened 2021-03-18
[factor] Only reload fuel mode if factor root path has changed.

timor:
Without this, fuel mode is reloaded needlessly on every Factor REPL restart.
Now we only do that if the Factor about to be invoked is a different one than
before.

### DIFF
--- a/layers/+lang/factor/funcs.el
+++ b/layers/+lang/factor/funcs.el
@@ -70,16 +70,22 @@ Will stop current fuel connection if applicable."
 
 Since unloading switches buffers which were in factor-mode back
 to fundamental mode, this re-enables factor-mode in these buffers
-afterwards."
-  (let ((factor-buffers (cl-loop for b being the buffers
-                                 if (eq (buffer-local-value 'major-mode b)
-                                        'factor-mode)
-                                 collect b)))
-    (factor//unload-fuel)
-    (factor//load-fuel-from-path path)
-    (cl-loop for b in factor-buffers do
-             (with-current-buffer b
-               (factor-mode)))))
+afterwards.
+
+Only reloads if currently loaded factor mode belongs to a different factor root.
+"
+  (when (not (string-equal (factor//fuel-elisp-dir)
+                           path))
+    (let ((factor-buffers (cl-loop for b being the buffers
+                                   if (eq (buffer-local-value 'major-mode b)
+                                          'factor-mode)
+                                   collect b)))
+      (factor//unload-fuel)
+      (message "Reloading fuel mode from %s" path)
+      (factor//load-fuel-from-path path)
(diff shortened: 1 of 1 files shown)


layers/+lang/factor/funcs.el
```diff
@@ -65,21 +65,31 @@ Will stop current fuel connection if applicable."
...
+                                   if (eq (buffer-local-value 'major-mode b)
+                                          'factor-mode)
+                                   collect b)))
+      (factor//unload-fuel)
+      (message "Reloading fuel mode from %s" path)
+      (factor//load-fuel-from-path path)
+      (cl-loop for b in factor-buffers do
+               (with-current-buffer b
+                 (factor-mode)))))
+  (setq factor--previous-fuel-path path))
```
lebensterben, line 92:
```elisp
(defun factor//reload-fuel-from-path (path)
  (when (and factor--previous-fule-path
             (not (eql factor-previous-fule-path path)))
    (factor//unload-fuel)
    (message "Reloading fuel mode from %s" path)
    (factor//load-fuel-from-path path)
    (cl-loop for b in (buffer-list)
             if (eq (buffer-local-value 'major-mode b)
                    'factor-mode)
             do (with-current-buffer b
                  (factor-mode +1))))
  (setq factor--previous-fuel-path path))
```

This is shorter. Also `cl-loop for FOO in BAR` is slightly faster than `cl-loop for FOO being the BAR`.

timor: Thanks. Will change that.

timor: Actually, that modification does not seem to work.  I need to collect the list of factor-mode buffers before unloading, because `major-mode` will not be `factor-mode` anymore after unloading.

timor: > ```emacs-lisp
> (defun factor//reload-fuel-from-path (path)
>   (when (and factor--previous-fule-path
>              (not (eql factor-previous-fule-path path)))
>     (factor//unload-fuel)
>     (message "Reloading fuel mode from %s" path)
>     (factor//load-fuel-from-path path)
>     (cl-loop for b in (buffer-list)
>              if (eq (buffer-local-value 'major-mode b)
>                     'factor-mode)
>              do (with-current-buffer b
>                   (factor-mode +1))))
>   (setq factor--previous-fuel-path path))
> ```
> 
> This is shorter. Also `cl-loop for FOO in BAR` is slightly faster than `cl-loop for FOO being the BAR`.

smile13241324 (approved):
Looks good

layers/+lang/factor/funcs.el
smile13241324: As the loop contents has not been changed this can already be merged. Optimisation of the loop should be done in a separate PR.

PR #14512, opened 2021-03-18
Enhance +lang/zig layer

rbino:
Add support to LSP via zls (see also https://github.com/emacs-lsp/lsp-mode/pull/2723)

### DIFF
--- a/layers/+lang/zig/README.org
+++ b/layers/+lang/zig/README.org
@@ -6,14 +6,59 @@
 - [[#description][Description]]
   - [[#features][Features:]]
 - [[#install][Install]]
+  - [[#layer][Layer]]
+  - [[#choosing-a-backend][Choosing a backend]]
+    - [[#lsp-zls][LSP (zls)]]
+    - [[#zig-mode][zig-mode]]
+- [[#key-bindings][Key bindings]]
+  - [[#lsp][LSP]]
 
 * Description
 This layer adds support for the zig programming language [[https://ziglang.org/]]
 
 ** Features:
 - Syntax Highlighting
+- LSP support via =zls= (see [[#lsp-zls][LSP]])
 
 * Install
+
+** Layer
 To use this configuration layer, add it to your =~/.spacemacs=. You will need to
 add =zig= to the existing =dotspacemacs-configuration-layers= list in this
 file.
+
+** Choosing a backend
+You cane choose between multiple language backends.
+This choice defines mainly which kind of IDE features spacemacs
+can deliver.
+
+Alternatively, if non is set the =lsp= backend will be automatically
+chosen if the layer =lsp= is used, otherwise =zig-mode= will be used.
+
+For best results, make sure that the =auto-completion= and =syntax-checking=
+layers are enabled as well.
+
+*** LSP (zls)
+You can use [[https://github.com/zigtools/zls][=zls=]] as a Language Server backend to add features like
+auto-completion, code navigation, formatting etc.
+
+To do so, make sure you have the =lsp= layer in your
+=dotspacemacs-configuration-layers=. If necessary, upgrade the =lsp-mode=
+package (support for Zig was added in version =20210318.923=). Install =zls=
+following the [[https://github.com/zigtools/zls/wiki/Downloading-and-Building-ZLS#using-an-official-release][instructions on the Wiki]] and then configure the =zls= layer with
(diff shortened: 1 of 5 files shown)


layers/+lang/zig/funcs.el
```diff
@@ -0,0 +1,36 @@
...
+;; lsp
+
+;; zls
+
+(defun spacemacs//zig-setup-zls ()
+  "Setup LSP zls."
+  (if (configuration-layer/layer-used-p 'lsp)
+      (progn
+        (when zls-path
+          (add-to-list 'exec-path zls-path))
```
smile13241324, line 34: This looks like a global change to the exec-path, don't you have a variable in lsp mode you can set for this? Changing this globally for the need of one layer does not look like an ideal solution to me.

layers/+lang/zig/packages.el
```diff
@@ -20,7 +20,7 @@
     :defer t
     :init
     (progn
-      ;; config goes here.
+      (spacemacs/add-to-hook 'zig-mode-hook '(spacemacs//zig-setup-backend))
```
smile13241324, line 23: If you just add one function to one hook you should consider using the standard add-to-hook function from emacs. Also please pass only the function in there, using the `#'` form, this helps the compiler to detect function passing.

layers/+lang/zig/config.el
```diff
@@ -0,0 +1,20 @@
...
+;; Author: Riccardo Binetti <rbino@gmx.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
+
+(spacemacs|define-jump-handlers zig-mode)
+
+(defvar zig-backend nil
```
smile13241324, line 14: The backend variable should support two values, of which one must be `lsp` the other one should be the simple choice for non LSP users. Nil is reserved for `not set` resulting in an automatic choice of backends.
In your case lets introduce `(lsp zig-mode)` and describe both alternative in the README.

`zls` would be a specific implementation of an LSP server for which there is a different standard variable, see the python layer for details.

However with only one server supported introducing the server choice variable is not necessary.

layers/+lang/zig/README.org
```diff
@@ -6,14 +6,43 @@
...
+  - [[#lsp-zls][LSP (zls)]]
+- [[#key-bindings][Key bindings]]
+  - [[#lsp][LSP]]
 
 * Description
 This layer adds support for the zig programming language [[https://ziglang.org/]]
 
 ** Features:
 - Syntax Highlighting
+- LSP support via =zls= (see [[#lsp-zls][=zls=]])
```
smile13241324, line 19: A local link in org will not work when HTML is generated from the org files, please link to the develop.spacemacs.org website instead.

smile13241324 (changes requested):
I have just reviewed your PR and found some issues you should resolve before this can be merged.
If you have questions or require assistance just drop me a line.

layers/+lang/zig/README.org
rbino: I took this from the [README of the Go Layer](https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Blang/go/README.org) which has a local link ("`see LSP backend`") and it [seems to work also in the website](https://develop.spacemacs.org/layers/+lang/go/README.html#lsp-gopls).

layers/+lang/zig/funcs.el
rbino: `lsp-mode` just [added said variable](https://github.com/emacs-lsp/lsp-mode/pull/2724), so I removed this

rbino:
@smile13241324 thanks for the review, I fixed the issues you raised

rbino:
Added explicit GPLv3 terms following https://github.com/syl20bnr/spacemacs/commit/97cd83e169a7d3a5633bfebe561b0084a5ed6c1a

smile13241324:
Looks like I will not be able to work on this this week, I will review again when I am back on the project.

layers/+lang/zig/README.org
smile13241324: Hmm good point, I have thought this would not be supported but it looks like I have been proven wrong, lets keep it then.

smile13241324 (approved):
Looks good, will be merged in a minute

PR #14515, opened 2021-03-18
colors: replace local nyan-mode with the one on melpa

### LINKED ISSUE #14505: localised `nyan-mode` should be removed as it's available on melpa
https://github.com/TeMPOraL/nyan-mode/

lebensterben:
closes  #14505


Kazark (approved):

smile13241324:
I am not sure about this one, here we are effectively using a fork from @syl20bnr with some smaller changes to the animation which have never been submitted to the parent repository.

Maybe we can try to port these improvements to the parent?

lebensterben:
I think the real addition is
> - the animation loop is not endless and occurs only when nyan cat moves.

We probably could open a separate repo for this modified nyan-cat, and replace the one in `colors` layer with the upstream one.
Should any user opened an issue on missing this feature, we then consider to add a layer variable to allow them to choose the modified one, or to send the patch to upstream.

syl20bnr:
If we could have a toggle upstream to choose between endless and finite animation it would do the trick. I can take a quick look to see if this is easy to implement.

syl20bnr:
I opened a PR with new config here: https://github.com/syl20bnr/spacemacs/pull/14539
See this new PR for more info.

syl20bnr:
Can be merged when https://github.com/syl20bnr/spacemacs/pull/14539 is merged.

syl20bnr (approved):

PR #14516, opened 2021-03-18
Fix void variable `mu4e-view-mode-map`

claritystorm:
Fixed, the following error, void variable `mu4e-view-mode-map`.

`Error (use-package): mu4e/:config: Symbol’s value as variable is void: mu4e-view-mode-map`


smile13241324 (approved):

PR #14518, opened 2021-03-19
Enhance shell-default-position validation spec

surgeboris:
to match values listed in it's documentation.

I prefer to have my eshell opened to the right of my current window, but it does not seem to be allowed by [recently introduced validation rule](https://github.com/syl20bnr/spacemacs/commit/369cf7519f0bdf90d9e4285d4114b0cd42a49a33). 

@JAremko have you intentionally omitted `(const left)`  and `(const right)` values there (are those values deprecated)? I suppose that it's just an oversight there, so I bothered to create this PR.


JAremko:
Thanks, my bad :+1:

PR #14520, opened 2021-03-19
Fix .spacemacs validation choices

Claude-Ray:
Add choices for variables that are allowed to be nil

### DIFF
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -265,7 +265,7 @@ can be a symbol or a list with additional properties like '(all-the-icons
 (spacemacs|defc dotspacemacs-frame-title-format "%I@%S"
   "Default format string for a frame title bar, using the
 original format spec, and additional customizations."
-  'string
+  '(choice (const nil) string)
   'spacemacs-dotspacemacs-init)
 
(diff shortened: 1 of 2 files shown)


smile13241324:
@JAremko I am not sure whether this makes sense, care to have a look?

JAremko:
`dotspacemacs-major-mode-leader-key` makes sense. Not sure about `dotspacemacs-frame-title-format` is `nil`  simply removes the title?

Claude-Ray:
> `dotspacemacs-major-mode-leader-key` makes sense. Not sure about `dotspacemacs-frame-title-format` is `nil` simply removes the title? If so then why not 🤔

When dotspacemacs-frame-title-format is nil, Spacemacs is using default frame-title-format instead of calculating title by spacemacs/title-prepare all the time.

This did help to avoid performance issues.

JAremko:
@Claude-Ray  that's cool.  Can you update `dotspacemacs-frame-title-format` with info about `nil` ?

JAremko:
And the comment above https://github.com/syl20bnr/spacemacs/blob/develop/core/templates/.spacemacs.template#L454

Claude-Ray:
Yes, my pleasure

JAremko:
Thanks :+1:

PR #14521, opened 2021-03-19
windows-scripts: replaced dos.el with bat-mode, added bmx-mode

### LINKED ISSUE #14499: `dos-mode` package in `windows-script` layer should be replaced by `bat-mode`
https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Blang/windows-scripts/local/dos/dos.el

`dos-mode` is obsolete and it's renamed to `bat-mode`, which is shipped with `emacs` now. We should replace it with `bat-mode`.

lebensterben:
- Replaced `dos.el` with `bat-mode`
  - `dos.el` is superseded by `bat-mode`
  - `dos.el` is localised package
  - `bat-mode` is a built-in mode
- Added `bmx-mode`
  - `bmx-mode` provided syntax highlighting, code-navigation and refactoring
    capabilities on top of `bat-mode`

closes #14499


lebensterben:
I need some help:

First, According to [documentation of spacebind](https://github.com/syl20bnr/spacemacs/blob/b71e5fdb179d7fdba92f80ce07aedbfb70339411/core/core-spacebind.el#L399-L400) doc-string would be automatically added to `README.org`. Does that mean it would be added by our bot automatically?

Second, There's something wrong with major-mode keybindings of `bat-mode`. After-loading, `spacemacs-bat-mode-map` shows:
```text
<which-key> e eval    nil
<which-key> h help    nil
<which-key> i insert  nil
e B                   bat-run-args
e b                   bat-run
h h                   bat-cmd-help
i t                   bat-template
z                     windows-scripts/bat-outline
```
which is what I intended to have. But
- `SPC m` shows keybindings from minor-mode `bmx-mode` (that's intended) and `z windows-scripts/bat-outline`. All the others are gone...
- `,` shows only bindings from `bmx-mode` and not even `z windows-scripts/bat-outline`.

This is the first time I'm using `spacemacs|spacebind` and I'm unsure about whether this is a bug or I'm doing it wrong....

JAremko:
> First, According to documentation of spacebind doc-string would be automatically added to README.org. Does that mean it would be added by our bot automatically?

`spacebind` forms are plain data so it can be statically parsed and used as an additional data source for the .org generator:

Currently bot fixes documentation files by extracting data out them, validates it and infers missing parts (the .org validation in PRs digs into docs and tries to figure out whether the bot will be able to fix it, this is why CI doesn't complain about the stuff that can be fixed.) 
*--- here we can add data about keybindings ---*
*--- and perhaps layer variables and their types ---*
Then normalized data is used to generate .org files.


> This is the first time I'm using spacemacs|spacebind and I'm unsure about whether this is a bug or I'm doing it wrong....

Could be a bug :thinking:  I'm planing to rewrite the binding code for `spacebind` anyway.

JAremko:
@lebensterben 


> It turns out that you have to declare the prefix for major modes (but not for minor modes) with
spacemacs/declare-prefix-for-mode.

So [spacemacs/declare-prefix-for-mode](https://github.com/syl20bnr/spacemacs/blob/eccdcbb1dbd77864c89612ba281c088a29f6aa2c/core/core-keybindings.el#L52) is bugged then?

lebensterben:
> Currently bot fixes documentation files by extracting data out of them, validates it and infers missing parts

So I assume I can be lazy and just skip the entire key-binding section and let the CI do the job?

> So spacemacs/declare-prefix-for-mode is bugged then?
Should be the cause. But I haven't debug this. This should be resolved in another PR.

Also notice that, before I manually call `spacemacs/declare-prefix-for-mode`, only the one not nested under a `<prefix>  nil`, i.e. `z windows-scripts/bat-outline` is correctly shown in `SPC m`, but not `,`. This is also odd...

JAremko:
> So I assume I can be lazy and just skip the entire key-binding section and let the CI do the job?

Well, we don't know when it will be implemented, meanwhile, we still need keybinding documentation :smile:

smile13241324:

Btw, spacebind did work flawlessly for me, maybe the issue you have described is gone?
I have also documented the keybindings, for the doc extraction is not yet live.

JAremko:
> Btw, spacebind did work flawlessly for me, maybe the issue you have described is gone?
I have also documented the keybindings, for the doc extraction is not yet live.

@smile13241324  I think Clojure mode prefixes are messed up. We set them without spacebind programmatically.  Maybe it has to something to do with collisions.  like if we override prefix, instead of showing new one it just turns into "+prefix" placeholder :thinking:

smile13241324:
I have just had a look into this prefix issue you have mentioned. For me this looks like a general issue with (`set-leader-keys-for-minor-mode)` as it happens with or without the `windows-scripting` layer.

I think we need to check this separately, maybe there is a conflict somewhere in the keymap...

PR #14523, opened 2021-03-19
python: replaced local package `nose` with `syl20bnr/nose.el`

lebensterben:


smile13241324:

PR #14528, opened 2021-03-19
[Fix] `dotspacemacs/test-dotfile` complains error on new custom vars

rayw000:
After new custom `dotspacemacs-*` variables being introduced in `core/core-dotspacemacs.el`, if one has an old `.spacemacs` without these variables set, `SPC f e R` will complain error messages like
```
Variable: "dotspacemacs-emacs-dumper-dump-file" has value: "nil" that doesn't match its type: "string"...
```
even if these variables have already had initial values.

This is because `dotspacemacs||let-init-test` macro only locally declares these variables without setting them with their current values.

This pull request fixes this issue by expanding `(dotspacemacs/get-variable-list)` into `'((var1 value1) (var2 value2) ...)` list instead of `(var1 var2 ...)`.

### DIFF
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -1078,7 +1078,11 @@ error recovery."
 (defmacro dotspacemacs||let-init-test (&rest body)
   "Macro to protect dotspacemacs variables"
   `(let ((fpath dotspacemacs-filepath)
-         ,@(dotspacemacs/get-variable-list)
+         ,@(mapcar (lambda (symbol)
+                     `(,symbol ,(let ((v (symbol-value symbol)))
+                                  (if (or (symbolp v) (listp v))
+                                      `',v v))))
+                   (dotspacemacs/get-variable-list))
          (passed-tests 0) (total-tests 0))
      (setq dotspacemacs-filepath fpath)
      (load dotspacemacs-filepath)


smile13241324:
@JAremko this looks like its related to the latest changes on validating dotfile settings. Can you have a look?

JAremko:
@smile13241324  From the look of it - makes sense.

JAremko:
This also might be a problem dotspacemacs//test-dotspacemacs/layers :thinking:

JAremko:
I'm thinking, maybe instead of doing all this protective stuff it would be simpler to run the tests in a separate Emacs process.

rayw000:
> I'm thinking, maybe instead of doing all this protective stuff it would be simpler to run the tests in a separate Emacs process.  The tests can also run `spacemacs-customization//validate-dotspacemacs-init-vars ` and `spacemacs-customization//validate-dotspacemacs-layers-var` explicitly, catch the error (if any) and dump it as a text into the report.

I like this idea. Currently when some errors break the test procedure protected by `dotspacemacs||let-init-test`, it might be confusing that we may get surprising variable values.

smile13241324:
@JAremko can you give me a hint about the state of this PR, is it ready to be merged or do you require some adjustments first? I think you have lost me somewhere in your discussion 🤔.

JAremko:
@smile13241324  It's ok.  We can merge it as a hot (well, warm-ish) fix but doing https://github.com/syl20bnr/spacemacs/pull/14528#issuecomment-803189300 would make more sense in the long run.

PR #14532, opened 2021-03-19
Add Edebug Eval List key bindings

claritystorm:
Added Edebug Eval List key bindings.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1903,6 +1903,14 @@ Other:
   - Make ~SPC m h h~ and jump handlers work in =ielm= (thanks to Keith Pinson)
   - Added ~c~ to continue in edebug mode (thanks to hodge)
   - Added ~SPC m g b~ to jump back to previous point (thanks to Magnus Therning)
+  - Added Edebug Eval List key bindings
+    - ~SPC m g w~ for =edebug-where=
+    - ~SPC m a~ for =edebug-delete-eval-item=
+    - ~SPC m k~ for =edebug-delete-eval-item=
+    - ~SPC m ,~ for =edebug-update-eval-list=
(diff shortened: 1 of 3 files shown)


smile13241324 (approved):
Looks good, though your key bindings look a bit strange to me i.e. why do you have duplicate bindings for delete-eval-item and update-eval-list but we do not have any bindings for this mode yet either so lets merge them and see how they work. We can still change them when necessary.

dalanicolai:
Nice to have these bindings, but indeed they look strange to me. @claritystorm What is the logic behind them?

I would propose to change them to
```
:config
(spacemacs/set-leader-keys-for-major-mode 'edebug-eval-mode
  "j" 'edebug-eval-print-last-sexp
  "e" 'edebug-eval-last-sexp
  "u" 'edebug-update-eval-list
  "d" 'edebug-delete-eval-item
  "w" 'edebug-where)))
```
which is more mnemonic and also is more in line with the [defaults Emacs keybindings](https://www.gnu.org/software/emacs/manual/html_node/elisp/Eval-List.html#Eval-List).

But actually I came here because I would like the `Edebug Eval` buffer to come up in a split window. For some reason in Spacemacs in opens in the currently selected window, while, looking at the code of `edebug-pop-to-buffer`, it should open in a split window as it does in vanilla Emacs.

Does it open in a split window for you? Or do you know how to achieve that?

claritystorm:
This was an attempt to follow Spacemacs key bindings conventions, mirror Emacs Lisp layer prefix keys as well as _Confirm and Abort_ conventions. Since no key bindings were defined these were added. I have no objection to the change proposed by @dalanicolai

> Does it open in a split window for you? Or do you know how to achieve that?

No; try the Spacemacs-purpose layer.

PR #14533, opened 2021-03-19
Add keybind to reload Rust-Analyzer workspace

teozkr:
Allows you to pick up new Cargo dependencies without a (slooow) full server restart.

~Should not be merged before https://github.com/emacs-lsp/lsp-mode/pull/2728 is.~

### DIFF
--- a/layers/+lang/rust/funcs.el
+++ b/layers/+lang/rust/funcs.el
@@ -62,9 +62,9 @@
         (mapcar 'lsp--client-server-id)
         (member 'rust-analyzer))
       (progn
-       (lsp-rust-analyzer-reload-workspace)
-       (message "Reloaded Rust-Analyzer workspace"))
-    (message "Can only reload Rust-Analyzer workspaces, consider restarting workspace instead")))
+        (lsp-rust-analyzer-reload-workspace)
+        (message "Reloaded workspace"))
+    (message "RLS reloads automatically, and doesn't require an explicit reload")))
 
 
 ;; racer
--- a/layers/+lang/rust/funcs.el
+++ b/layers/+lang/rust/funcs.el
@@ -47,13 +47,24 @@
         (lsp)
         (spacemacs/declare-prefix-for-mode 'rust-mode "ms" "switch")
         (spacemacs/set-leader-keys-for-major-mode 'rust-mode
-          "bR" 'lsp-rust-analyzer-reload-workspace
+          "bR" 'spacemacs//lsp-rust-analyzer-reload-workspace
           "ss" 'lsp-rust-switch-server))
     (spacemacs//lsp-layer-not-installed-message)))
 
 (defun spacemacs//rust-setup-lsp-dap ()
   "Setup DAP integration."
   (require 'dap-gdb-lldb))
+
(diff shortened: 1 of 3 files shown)


lebensterben:
One minor issue, this key binding should only be available when the LSP backend is rust-analyzer, and should emit a warning if it's not. (Preferably it also display a message when reloading workspace)

teozkr:
@lebensterben Are these diagnostics ok?

layers/+lang/rust/funcs.el
```diff
@@ -47,14 +47,25 @@
...
           "ss" 'lsp-rust-switch-server))
     (spacemacs//lsp-layer-not-installed-message)))
 
 (defun spacemacs//rust-setup-lsp-dap ()
   "Setup DAP integration."
   (require 'dap-gdb-lldb))
 
+(defun spacemacs//lsp-rust-analyzer-reload-workspace ()
+  (interactive)
+  (if (->> (lsp-workspaces)
```
lebensterben, line 60: Looks like `->>` requires `dash`.
It's best to avoid it or declare it as dependency explicitly.

lebensterben:
> @lebensterben Are these diagnostics ok?

Yes on a second thought I think we can do this:
- When it's rust-analyzer, RELOAD the workspace and emit "Reloading workspace"
- When it's rls, RESTART the workspace and emit "RLS doesn't support reloading workspace, fallback to restarting workspace" (Something like this. It's idiot-proof.)

layers/+lang/rust/funcs.el
teozkr: Looks like `dash` comes from the `spacemacs-bootstrap` layer, which implies (to me at least) that it's always available. `ivy-spacemacs-help` doesn't seem to mention any other layer explicitly depending on `dash`, from what I can tell.

lebensterben: That should be okay then.

teozkr:
RLS seems to pick up Cargo.toml changes immediately, without requiring any manual reload. So maybe the correct behaviour here would actually be to do nothing?

lebensterben:
If so, then you are right. And no need to make it an error. A message that no need to reload manually for rls should suffice.

lebensterben:
We probably can do something even better. We can add a layer variable which toggles auto-reloading of workspace after user called cargo-process-add.

teozkr:
I agree that that would be nice, but hooking into `cargo-process` seems like it'd require someone more familiar with elisp than me.. :/

lebensterben:
@teozkr No problem I will do that in a separate commit. I hope your PR to lsp-mode is merged ASAP.

teozkr:
It was, a few hours ago!

smile13241324 (approved):
👍

smile13241324:

lebensterben:
It comes to me that, when I added `lsp-rust-switch-server` under `SPC m s s`, instead of `SPC m b s`, the motivation was if otherwise, `SPC m b s` binding persists when you go to other major mode with lsp backend.

I haven't found a decent way to remove `SPC m b s`.
This PR adds a new binding `SPC m b R`, it would have the same issue.

@smile13241324 What do you think?

PR #14536, opened 2021-03-20
Minibuffer bind C-n C-p to <next>/<previous>-line-or-history-element

rayw000:
Currently `C-n` `C-p` are bound to `next-line` and `previous-line` in `minibuffer-local-map`. This PR binds them to `next-line-or-history-element` and `previous-line-or-history-element`. This may be handy when using `M-:` to call `eval-expression`.


smile13241324 (approved):
👍

PR #14538, opened 2021-03-20
[home] Add multi-digit numbers to the startup lists

duianto:
Added support for jumping to two digit numbers.

How quickly two numbers have to be typed,
can be modified with the variable:
`spacemacs-buffer-multi-digit-delay`

---

The file/dir opens as soon as the idle timer runs (default `0.4` seconds).

#### Minibuffer only
While typing: `Jump to startup list: <nr>` 
If the startup list number isn't found: `Couldn't find startup list number: <nr>`

#### Mini and Messages buffer
When the number matches a startup list line: `Opening file/dir: <nr> <file or dir name>`

emacs_2021-03-20_11-56-27

### DIFF
--- a/core/core-spacemacs-buffer.el
+++ b/core/core-spacemacs-buffer.el
@@ -61,6 +61,17 @@ Internal use, do not set this variable.")
 
 (defvar spacemacs-buffer-mode-map
   (let ((map (make-sparse-keymap)))
+    (define-key map (kbd "0") 'spacemacs-buffer/jump-to-number-startup-list-line)
+    (define-key map (kbd "1") 'spacemacs-buffer/jump-to-number-startup-list-line)
+    (define-key map (kbd "2") 'spacemacs-buffer/jump-to-number-startup-list-line)
+    (define-key map (kbd "3") 'spacemacs-buffer/jump-to-number-startup-list-line)
+    (define-key map (kbd "4") 'spacemacs-buffer/jump-to-number-startup-list-line)
+    (define-key map (kbd "5") 'spacemacs-buffer/jump-to-number-startup-list-line)
+    (define-key map (kbd "6") 'spacemacs-buffer/jump-to-number-startup-list-line)
+    (define-key map (kbd "7") 'spacemacs-buffer/jump-to-number-startup-list-line)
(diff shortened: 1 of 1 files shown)


duianto:
The previous behavior with only single numbers (1-9),
opened the file directly when a number was typed.

I'm not sure if this multi-digit solution works as expected.
Therefore this only moves the cursor to the target startup list line,
`RET` has to be pressed to open the file.

If the file should open as soon as the number has been typed and the idle timer is called.
Then I'll rewrite it to accumulate the key presses.

That might also support more than two digits.

smile13241324 (approved):
Looks good.

lebensterben:
One minor issue:
`spacemacs-buffer-multi-digit-delay` is not documented currently.

duianto:
Would this be more descriptive?
If not then any recommendations for improvements/rewrites are always welcome.

### Before
```elisp
(defvar spacemacs-buffer-multi-digit-delay 0.4
  "This determines how quickly (in seconds)
the numbers have to be typed.")
```

### After
```elisp
(defvar spacemacs-buffer-multi-digit-delay 0.4
  "The idle time in seconds, after the last number has been typed,
before the numbered home buffer startup list widget button is pressed.")
```

lebensterben:
@duianto 
I think the current doc-string is already very clear.
What I mean is this variable is not mentioned in FAQ, or appear in dotspacemacs template, which makes this not easily discovered by user.

`dotspacemacs-which-key-delay` is in the template, so we probably can add this just below it.

duianto:
>I think the current doc-string is already very clear.
What I mean is this variable is not mentioned in FAQ, or appear in dotspacemacs template, which makes this not easily discoverable by user.

Ok, yeah that makes sense.

There are four startup buffer related dotspacemacs variables:
(in my `.spacemacs` they are on the lines: 237, 245, 259 and 269)
```
   dotspacemacs-startup-buffer-show-version t
   dotspacemacs-startup-buffer-responsive t
```
The `dotspacemacs-which-key-delay` variable is on line 388.

It might be easier to find this startup number delay variable if it's grouped with the related variables.

lebensterben:
Exactly, they'd better be grouped together.

duianto:
A PR has been opened:
[dotfile] New var: dotspacemacs-startup-buffer-multi-digit-delay #14556

PR #14541, opened 2021-03-20
fixup! Validate .spacemacs variables.

lebensterben:
fixes syl20bnr#14535 syl20bnr#14537 syl20bnr#14529 syl20bnr#10638 (comment)

### DIFF
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -171,7 +171,7 @@ a layer lazily."
 wrapped in a layer. If you need some configuration for these
 packages then consider to create a layer, you can also put the
 configuration in `dotspacemacs/user-config'."
-  '(repeat symbol)
+  '(repeat (choice symbol (cons symbol sexp)))
   'spacemacs-dotspacemacs-layers)
 
 (defvar dotspacemacs--additional-theme-packages '()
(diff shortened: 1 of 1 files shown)


lebensterben:
@JAremko Here we go

core/core-dotspacemacs.el
```diff
@@ -171,7 +171,7 @@ a layer lazily."
 wrapped in a layer. If you need some configuration for these
 packages then consider to create a layer, you can also put the
 configuration in `dotspacemacs/user-config'."
-  '(repeat symbol)
+  '(repeat (choice symbol (cons symbol sexp)))
```
JAremko, line 174: maybe the type can be stricter here.

In the case:

```
   dotspacemacs-additional-packages '(
     amx
     (beacon :location (recipe :fetcher github :repo "Malabarba/beacon")))
```
it looks like in the cons case we always have first 2 elements as symbols.  So maybe '(repeat (choice symbol (cons symbol (const symbol sexp))) will work :thinking:

JAremko: On the other hands it's too cryptic.

JAremko:
It's good as it is. Thanks agan.

Issue #1781, opened 2015-05-31
Smooth scrolling [Enhancement]


ReneFroger:
[Smooth scroll plugin](http://www.emacswiki.org/emacs/smooth-scroll.el). 

With the following settings: 

```
( require 'smooth-scroll                        ) ;; Smooth scroll 
( smooth-scroll-mode 1                          ) ;; Enable it 
( setq smooth-scroll/vscroll-step-size 5        ) ;; Set the speed right
```

Try scroll option, or with page-up and page-down. It gives 

I believe this is a big improvement. Because when I'm scrolling, I have difficulties to recognize in which way the file was scrolling and where I exactly am. This package helps me with that.

tuhdo:
We had a PR a while ago: https://github.com/syl20bnr/spacemacs/pull/1670

But thank you :+1:

syl20bnr:
We are using `smooth-scrolling` plugin, is this really the same thing as `smooth-scroll` ?

sooheon:
Just tested it. They address different issues. The `smooth-scrolling` we have currently is like vim's scrolloff margin feature, which lets you scroll off screen without the cursor jumping to the center of the screen. `smooth-scroll` is literally a smoother (graphically), non-instantaneous scroll. Therefore, it cannot replace `smooth-scrolling`, but it could be an addition for those who want it. Personally I could take it or leave it, this is what I added to my config to make it work if @ReneFroger wants to improve and submit a layer:

```
(defun sooheon/init-smooth-scroll ()
  (use-package smooth-scroll
    :defer
    :init
    (progn
      (require 'smooth-scroll)
      (smooth-scroll-mode 1)
      (setq smooth-scroll/vscroll-step-size 5))))
```

tuhdo:
@sooheon you are right. I've tested it. It works nicely even on large file. I think this would make a user friendly addition for new comers.

@ReneFroger Thank you for this awesome suggestion. Sorry for not testing it and jump to the conclusion so quickly.

ReneFroger:
@tuhdo it's not a problem, I was surprised by your first comment, because I considered `smooth-scroll` to provide much more smooth scrolling than `smooth-scrolling`. But glad to read that you like it. 

By the way, everyone, when you compare it with his [Vim counterpart Smooth Scroll](https://github.com/terryma/vim-smooth-scroll) , it's even actually bad.

Let me demonstrate it for you:

https://www.youtube.com/watch?v=GCwZnBgR4DM

[...]

Issue #12293, opened 2019-05-12
Mouse wheel scrolling with control held scrolls window, rather than zooming the frame


DMRobertson:
#### Description :octocat:
Mouse wheel scrolling with control held scrolls window, rather than zooming the frame.

#### Reproduction guide :beetle:
- Start Emacs
- Open a buffer with text inside which cannot be displayed within an entire window (e.g. open your spacemacs dotfile)
- Hold control and scroll the mouse wheel up and down a few times

*Observed behaviour:* :eyes: :broken_heart:
- The window scrolls up and down, just as if control was not held.

*Expected behaviour:* :heart: :smile:
Judging by the following lines from the end of [spacemacs-ui-visual/packages.el](https://github.com/syl20bnr/spacemacs/blob/c7a103a772d808101d7635ec10f292ab9202d9ee/layers/%2Bspacemacs/spacemacs-ui-visual/packages.el), I think the intention is that this increases the font size, just like a web browser would.

```
(defun spacemacs-ui-visual/init-zoom-frm ()
      ;; ...
      ;; Font size, either with ctrl + mouse wheel
      (global-set-key (kbd "<C-wheel-up>") 'spacemacs/zoom-frm-in)
      (global-set-key (kbd "<C-wheel-down>") 'spacemacs/zoom-frm-out))))
```

I can see these bindings in the "Helm descbinds" popup (`SPC ?`). I think the problem instead is that scrolling the wheel generates a different kind of event. If I describe variable `mouse-wheel-up-event` (`C-h v mouse-wheel--up-event RET`) I can see that it's value is `mouse-5`; similarly `mouse-wheel-down-event` is `mouse-4`.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 26.2
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 7a9c455)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(spell-checking auto-completion
                (helm :variables helm-ff-lynx-style-map nil)
                (org :variables org-todo-keywords
                     '((sequence "TODO" "WIP" "REVIEW" "DONE"))
                     org-hide-leading-stars t org-bullets-bullet-list
                     '("*"))
                (c-c++ :variables c-c++-backend 'rtags c-c++-default-mode-for-headers 'c++-mode)
[...]

PR #14542, opened 2021-03-21
[defaults] Improve scrolling with mouse wheel

syl20bnr:
Thanks to Profpatsch.

Fixes https://github.com/syl20bnr/spacemacs/issues/1781

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -499,6 +499,7 @@ In org-agenda-mode
   - New variable =dotspacemacs-read-process-output-max= to optimise lsp
     performance in emacs 27 (thanks to Maximilian Wolff)
   - New variable =dotspacemacs-show-trailing-whitespace= (thanks to duianto)
+  - New variable =dotspacemacs-scroll-bar-while-scrolling= (thanks to duianto)
 - Removed Variables:
   - Removed unused variable =dotspacemacs-verbose-loading= from
     =.spacemacs.template= (thanks to Ying Qu)
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -500,6 +500,12 @@ screen."
   'boolean
   'spacemacs-dotspacemacs-init)
 
+(spacemacs|defc dotspacemacs-scroll-bar-while-scrolling t
+  "Show the scroll bar while scrolling. The auto hide time can be configured by
+setting this variable to a number."
+  '(choice boolean number)
+  'spacemacs-dotspacemacs-init)
+
 (spacemacs|defc dotspacemacs-line-numbers nil
   "Control line numbers activation.
 If set to `t' or `relative' line numbers are turned on in all `prog-mode' and
--- a/core/templates/.spacemacs.template
+++ b/core/templates/.spacemacs.template
@@ -371,6 +371,10 @@ It should only modify the values of Spacemacs settings."
    ;; when it reaches the top or bottom of the screen. (default t)
    dotspacemacs-smooth-scrolling t
 
+   ;; Show the scroll bar while scrolling. The auto hide time can be configured
+   ;; by setting this variable to a number. (default t)
+   dotspacemacs-scroll-bar-while-scrolling t
+
    ;; Control line numbers activation.
    ;; If set to `t', `relative' or `visual' then line numbers are enabled in all
    ;; `prog-mode' and `text-mode' derivatives. If set to `relative', line
--- a/layers/+spacemacs/spacemacs-defaults/config.el
+++ b/layers/+spacemacs/spacemacs-defaults/config.el
@@ -60,9 +60,6 @@ It runs `tabulated-list-revert-hook', then calls `tabulated-list-print'."
   ;; (tabulated-list-print t)
   (tabulated-list-print))
 
-;; Mouse cursor in terminal mode
-(xterm-mouse-mode 1)
-
 ;; Highlight and allow to open http link at point in programming buffers
 ;; goto-address-prog-mode only highlights links in strings and comments
 (add-hook 'prog-mode-hook 'goto-address-prog-mode)
@@ -77,6 +74,19 @@ It runs `tabulated-list-revert-hook', then calls `tabulated-list-print'."
 
 ;; Don't try to ping things that look like domain names
 (setq ffap-machine-p-known 'reject)
+
+;; ---------------------------------------------------------------------------
+;; Mouse
+;; ---------------------------------------------------------------------------
+
+;; Mouse cursor in terminal mode
+(xterm-mouse-mode 1)
+
+(when (boundp 'mouse-wheel-scroll-amount)
+  ;; scroll two line at a time (less "jumpy" than defaults)
+  (setq mouse-wheel-scroll-amount '(2)
+        ;; don't accelerate scrolling
+        mouse-wheel-progressive-speed nil))
 
 ;; ---------------------------------------------------------------------------
 ;; Edit
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -1433,6 +1433,23 @@ using a visual block/rectangle selection."
   (interactive)
   (spacemacs/sort-lines-by-column -1))
 
+;; Show scroll bar when using the mouse wheel
+(defun spacemacs//scroll-bar-hide ()
+  " Hide the scroll bar."
+  (scroll-bar-mode -1))
+(defun spacemacs//scroll-bar-show-delayed-hide (&rest _ignore)
+  "Show the scroll bar for a couple of seconds, before hiding it.
+
+This can be used to temporarily show the scroll bar when mouse wheel scrolling.
+(advice-add 'mwheel-scroll :after #'spacemacs//scroll-bar-show-delayed-hide)
+
+The advice can be removed with:
+(advice-remove 'mwheel-scroll #'spacemacs//scroll-bar-show-delayed-hide)"
+  (scroll-bar-mode 1)
+  (run-with-idle-timer 3 nil #'spacemacs//scroll-bar-hide))
+(when (fboundp 'scroll-bar-mode)
+  (advice-add 'mwheel-scroll :after #'spacemacs//scroll-bar-show-delayed-hide))
+
 ;; BEGIN linum mouse helpers
 
 (defvar spacemacs-linum-mdown-line nil
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -1437,6 +1437,7 @@ using a visual block/rectangle selection."
 (defun spacemacs//scroll-bar-hide ()
   " Hide the scroll bar."
   (scroll-bar-mode -1))
+
 (defun spacemacs//scroll-bar-show-delayed-hide (&rest _ignore)
   "Show the scroll bar for a couple of seconds, before hiding it.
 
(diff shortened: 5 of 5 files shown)


duianto:
With the cursor at the top of: `.spacemacs`

### In the screen recordings below
The scroll wheel is turned three times at "normal" speed.
The wheel turned about 5 ticks each time.


### Before applying the PR

It scrolls to quickly, it's easy to lose sight of how far one has scrolled.

The right side of the modeline shows that it scrolled to line 221 (18%).

One "benefit" might be that it doesn't take long to scroll to the bottom.

Depending on how fast one scrolls, in this 1286 line `.spacemacs` file,
it only takes about 7 turns.

before PR

### After applying the PR
It's easy to see what one has scrolled past.

The right side of the modeline shows that it scrolled to line 43 (4%).

One "issue" might be that it takes a while to scroll to the bottom.
At 14 turns, the buffer is only at 20%.

But it's most likely better to:
- be able to reliably see what one scrolls past
- having the scroll rate be similar to other applications:
3 lines at a time: `notepad`, `vim`, browsers? (probably depends on the font size)
slightly less than 3: `vscode`

after PR

duianto:
Maybe one way of mouse scrolling longer distances.
Could be to temporarily show the scroll bar for a couple of seconds after mouse wheel scrolling.
```elisp
      (defun spacemacs//scroll-bar-hide () (scroll-bar-mode -1))

      (defun spacemacs//scroll-bar-show-delayed-hide (&rest _ignore)
        "Show the scroll bar for a couple of seconds, before hiding it.

This can be used to temporarily show the scroll bar when mouse wheel scrolling.
(advice-add 'mwheel-scroll :after #'spacemacs//scroll-bar-show-delayed-hide)

The advice can be removed with:
(advice-remove 'mwheel-scroll #'spacemacs//scroll-bar-show-delayed-hide)"
        (scroll-bar-mode 1)
        (run-with-idle-timer 3 nil #'spacemacs//scroll-bar-hide))

      (advice-add 'mwheel-scroll :after #'spacemacs//scroll-bar-show-delayed-hide)
      ;; (advice-remove 'mwheel-scroll #'spacemacs//scroll-bar-show-delayed-hide)
```

duianto:
Another alternative.

Change scroll speed based on which third of the frame height, the mouse pointer is on.

The minibuffer shows the message:
`scroll speed: 4 (on frame top third)`
`scroll speed: 2 (on frame top third)`
`scroll speed: 1 (on frame top third)`

Maybe 4 isn't fast enough at the top

```elisp
      (defun spacemacs/mouse-y () (cddr (mouse-position)))

      (defun spacemacs//frame-height-third () (/ (frame-height) 3))

      (defun spacemacs//mouse-on-frame-top-third ()
        (< (spacemacs/mouse-y) (spacemacs//frame-height-third)))

      (defun spacemacs//mouse-on-frame-middle-third ()
        (and (> (spacemacs/mouse-y) (spacemacs//frame-height-third))
             (< (spacemacs/mouse-y) (* (spacemacs//frame-height-third) 2))))

      (defun spacemacs//mouse-on-frame-bottom-third ()
        (> (spacemacs/mouse-y) (* (spacemacs//frame-height-third) 2)))

      (defun spacemacs/mwheel-scroll-speed-on-frame-height-thirds (&rest ignore)
        "Set the mouse wheel scroll speed, based on which
frame third height the mouse pointer is on.
4 lines at a time on the frames top third.
2 lines at a time on the frames middle third.
1 lines at a time on the frames bottom third.

This can be used to advice the mouse wheel scrolling:
(advice-add 'mwheel-scroll :after #'spacemacs/mwheel-scroll-speed-on-frame-height-thirds)

The advice can be removed like this:
(advice-remove 'mwheel-scroll #'spacemacs/mwheel-scroll-speed-on-frame-height-thirds)"
        (cond ((spacemacs//mouse-on-frame-top-third)
               (setq mouse-wheel-scroll-amount '(4 ((shift) . 1)))
               (let (message-log-max)
                 (message "scroll speed: 4 (on frame top third)")))
              ((spacemacs//mouse-on-frame-middle-third)
               (setq mouse-wheel-scroll-amount '(2 ((shift) . 1)))
               (let (message-log-max)
                 (message "scroll speed: 2 (on frame middle third)")))
              ((spacemacs//mouse-on-frame-bottom-third)
               (setq mouse-wheel-scroll-amount '(1 ((shift) . 1)))
               (let (message-log-max)
                 (message "scroll speed: 1 (on frame bottom third)")))))

      (advice-add 'mwheel-scroll :after #'spacemacs/mwheel-scroll-speed-on-frame-height-thirds)
      ;; (advice-remove 'mwheel-scroll #'spacemacs/mwheel-scroll-speed-on-frame-height-thirds)
```

syl20bnr:
Great proposal @duianto !

I added the advice for the scroll bar which seems to be a pretty neat solution for people wanting to scroll faster.
Also I tweaked the amount to 1. 
At least on my system (macOS) it feels right like this, what about you ?

duianto:
Is this needed? Since it's `t` by default in Emacs:
```
;; scroll window under mouse
mouse-wheel-follow-mouse 't
```

---

It is smoother with a lower scroll amount.

But 1 might be on the slow side. (possibly because the default scroll speed is so fast)


Auto show/hiding the scroll bar has some issues:
- If one doesn't start dragging the handle within the idle time,
(which happens easily in large files where the scroll handle is small),
then the scroll bar will auto hide from under the cursor,
which won't be pleasant. When it happens more than once, 
then people probably will try to disable it.
- And toggling on the scroll bar widens the Emacs frame.
This means that if Emacs isn't maximized, and the frame is against the screens right edge, 
then the scroll bar opens outside the screens right edge.


Changing the scroll speed based on the mouse position relative to the frame height also has issues.

If the window is split top/bottom, then:
- the top window can't scroll at the slowest speed
- and the bottom window can't scroll at the fastest speed

It might be possible to change it to check for window thirds instead, with `window-top-line`,
but the scroll speed might feel inconsistent since the window thirds would become quite small.

syl20bnr:
Speed consistency is more important. It is not natural to have different speeds relative the file position.

I really like the scroll bar to appear when mouse scrolling, we can make a new dotspacemacs option to make it easy to disable for people who does not like it.

I'll revert the speed to 2.

Also on my system if I hold Control while scrolling it speeds up the scrolling. But I bet this is not like that on all systems.

syl20bnr:
I reverted to `2` and removed the `mouse-wheel-follow-mouse 't`.

What about the dotspacemacs variable ? I feel that we could add that and call it a day.

duianto:
>Also on my system if I hold Control while scrolling it speeds up the scrolling. But I bet this is not like that on all systems.

For me in Windows 10, control scrolling, resizes the Emacs frame.

---

There's a typo in the comment above the scroll amount, if we are going to use `2`
```
  ;; scroll one line at a time (less "jumpy" than defaults)
  (setq mouse-wheel-scroll-amount '(2)
```
---

I've got a working dotspacemacs variable: `dotspacemacs-scroll-bar-while-scrolling`
It accepts: `t`, `nil` and a number (integer or float).
Should I push the commit to this branch?

duianto:
I'll post the diff as well, if something needs to be changed, then it can be handled directly.
```diff
4 files changed, 20 insertions(+), 2 deletions(-)
CHANGELOG.develop                             |  1 +
core/core-dotspacemacs.el                     |  6 ++++++
core/templates/.spacemacs.template            |  4 ++++
layers/+spacemacs/spacemacs-defaults/funcs.el | 11 +++++++++--

modified   CHANGELOG.develop
@@ -499,6 +499,7 @@ In org-agenda-mode
   - New variable =dotspacemacs-read-process-output-max= to optimise lsp
     performance in emacs 27 (thanks to Maximilian Wolff)
   - New variable =dotspacemacs-show-trailing-whitespace= (thanks to duianto)
+  - New variable =dotspacemacs-scroll-bar-while-scrolling= (thanks to duianto)
 - Removed Variables:
   - Removed unused variable =dotspacemacs-verbose-loading= from
     =.spacemacs.template= (thanks to Ying Qu)
modified   core/core-dotspacemacs.el
@@ -500,6 +500,12 @@ screen."
   'boolean
   'spacemacs-dotspacemacs-init)
 
+(spacemacs|defc dotspacemacs-scroll-bar-while-scrolling t
+  "Show the scroll bar while scrolling. The auto hide time can be configured by
+setting this variable to a number."
+  '(choice boolean number)
+  'spacemacs-dotspacemacs-init)
+
 (spacemacs|defc dotspacemacs-line-numbers nil
   "Control line numbers activation.
 If set to `t' or `relative' line numbers are turned on in all `prog-mode' and
modified   core/templates/.spacemacs.template
@@ -371,6 +371,10 @@ It should only modify the values of Spacemacs settings."
    ;; when it reaches the top or bottom of the screen. (default t)
    dotspacemacs-smooth-scrolling t
 
+   ;; Show the scroll bar while scrolling. The auto hide time can be configured
+   ;; by setting this variable to a number. (default t)
+   dotspacemacs-scroll-bar-while-scrolling t
+
    ;; Control line numbers activation.
    ;; If set to `t', `relative' or `visual' then line numbers are enabled in all
    ;; `prog-mode' and `text-mode' derivatives. If set to `relative', line
modified   layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -1437,6 +1437,7 @@ using a visual block/rectangle selection."
 (defun spacemacs//scroll-bar-hide ()
   " Hide the scroll bar."
   (scroll-bar-mode -1))
+
 (defun spacemacs//scroll-bar-show-delayed-hide (&rest _ignore)
   "Show the scroll bar for a couple of seconds, before hiding it.
 
@@ -1446,8 +1447,14 @@ This can be used to temporarily show the scroll bar when mouse wheel scrolling.
 The advice can be removed with:
 (advice-remove 'mwheel-scroll #'spacemacs//scroll-bar-show-delayed-hide)"
   (scroll-bar-mode 1)
-  (run-with-idle-timer 3 nil #'spacemacs//scroll-bar-hide))
-(when (fboundp 'scroll-bar-mode)
+  (run-with-idle-timer
+   (if (numberp dotspacemacs-scroll-bar-while-scrolling)
+       dotspacemacs-scroll-bar-while-scrolling
+     3)
+   nil
+   #'spacemacs//scroll-bar-hide))
+(when (and (fboundp 'scroll-bar-mode)
+           dotspacemacs-scroll-bar-while-scrolling)
   (advice-add 'mwheel-scroll :after #'spacemacs//scroll-bar-show-delayed-hide))
 
 ;; BEGIN linum mouse helpers
```

syl20bnr:
> Should I force push to this branch?

Sure. Thank you.

So on Windows 10 the frame is always resized, even maximized ?
Is it acceptable or the resize is very annoying ?

duianto:
>So on Windows 10 the frame is always resized, even maximized ?
Is it acceptable or the resize is very annoying ?

In Spacemacs in Windows 10:
```
 <C-wheel-down>                                     ;; spacemacs/zoom-frm-out
 <C-wheel-up>                                       ;; spacemacs/zoom-frm-in
```

In both Spacemacs and `emacs -q` in Manjaro, and in `emacs -q` in Windows 10:
```
 <C-wheel-down>		  ;; mouse-wheel-text-scale
 <C-wheel-up>		  ;; mouse-wheel-text-scale
```

Spacemacs binds control and the scroll wheel here:
https://github.com/syl20bnr/spacemacs/blob/90cf83f21a7a8bbcbc41c81bf0fc159cb5d03dc4/layers/%2Bspacemacs/spacemacs-visual/packages.el#L145-L147

In Windows 10 the frame gets smaller when zooming both in and out.



Even when Spacemacs is maximized.



In `emacs -q` in Windows 10 (the same thing is observed in Manjaro in both Spacemacs and `emacs -q`),
only the text in the scrolled window is resized.

duianto:
The frame also shrinks in Manjaro when zooming the frame in the: `SPC z f`
Zoom Frame Transient State

This treads author is/was using linux, and reported the same behavior as you describe.
Mouse wheel scrolling with control held scrolls window, rather than zooming the frame #12293

It might be caused because the zoom package seems to have some `x` specific behaviors.
```
frame-cmds.el
693 (defcustom window-mgr-title-bar-pixel-height (cond ((eq window-system 'mac) 22)
695                                                    ((eq window-system 'ns)  50)
1313                                (if (eq window-system 'mac)
1512      (if (not (eq window-system 'x))
1561                 (if (not (eq window-system 'x)) ; Menu bar for X is not in the frame.
1575                (if (not (eq window-system 'x)) ; Menu bar for X is not in the frame.
1871                            (if (eq window-system 'mac)
```

smile13241324 (approved):
Feels much better now, will be cherry picked in a minute.

smile13241324:
Thank you for contributing to spacemacs 💜, I have cherry picked your commits into develop.

PR #14546, opened 2021-03-21
python: fix pylookup location

### LINKED ISSUE #14544: [error] can't find pylookup.py
Hi, after upgrade to last `develop` branch, there is an error when execute `pylookup-lookup` command:
>pylookup-exec-get-cache: Searching for program: No such file or directory, ~/.spacemacs.d/layers/+lang/python/local/pylookup/pylookup.py

It seems #14522 remove the `pylookup.py`.

lebensterben:
- `pylookup-program` defaults to the correct path so it doesn't needs to be set.
- Set `pylookup-db-file` to the correct value.

closes https://github.com/syl20bnr/spacemacs/issues/14544

### DIFF
--- a/layers/+lang/python/packages.el
+++ b/layers/+lang/python/packages.el
@@ -325,10 +325,7 @@
         "hH" 'pylookup-lookup))
     :config
     (progn
-      (let ((dir (configuration-layer/get-layer-local-dir 'python)))
-        (setq pylookup-dir (concat dir "pylookup/")
-              pylookup-program (concat pylookup-dir "pylookup.py")
-              pylookup-db-file (concat pylookup-dir "pylookup.db")))
+      (setq pylookup-db-file (concat pylookup-root "pylookup.db"))
       (setq pylookup-completing-read 'completing-read))))
 
 (defun python/init-pytest ()


lebensterben:
@sunlin7 would you test this?

layers/+lang/python/packages.el
```diff
@@ -325,10 +325,7 @@
         "hH" 'pylookup-lookup))
     :config
     (progn
-      (let ((dir (configuration-layer/get-layer-local-dir 'python)))
-        (setq pylookup-dir (concat dir "pylookup/")
-              pylookup-program (concat pylookup-dir "pylookup.py")
-              pylookup-db-file (concat pylookup-dir "pylookup.db")))
+      (setq pylookup-db-file (concat pylookup-root "pylookup.db"))
       (setq pylookup-completing-read 'completing-read))))
```
smile13241324, line 329: Just using pylookup-root will not work, for quelpa will not copy the python script to the respective elpa directory.
Instead you need to set the root like this:
```elisp
(setq pylookup-root (expand-file-name "pylookup" quelpa-build-dir))
```
Also setting these variables should be moved into the init section of the package.

smile13241324 (changes requested):
I think we need to postpone this till pylookup is python 3 compatible.

Apart of this I have added a comment about some smaller changes to properly set the path to the python script once it has been fetched via quelpa.

layers/+lang/python/packages.el
sunlin7: How about copy the `pylookup.py` from quelpa directory to the installation folder?

lebensterben: @sunlin7 
Setting the variable is cheaper than copying the file, and it has no side effect.
The real issue currently is the Python 3 compatibility.

sunlin7: @lebensterben  yes, agree, setting the `pylookup.py` path to quelpa directory is cheaper than copy it.
Let's focus on high priority first. 
And copying script file from source folder to destination folder is optimize for local package installization, it's on low priority. 
Thanks

Issue #14548, opened 2021-03-23
Evil keys not working for mu4e


jia-j-chen:
#### Description :octocat:
Evil keys not working for mu4e

#### Reproduction guide :beetle:
- Start Emacs
- Start mu4e
- Jump into inbox and open a message
- Try to move around inside the message using key "j" or "k"

*Observed behaviour:* :eyes: :broken_heart:
Cannot move and around and receive the error message below
"user-error: [mu4e] Unknown shortcut 'j'"

A futher check of the Messages buffer shows the following warning messages:
"Package cl is deprecated
Warning: Eager macro-expansion skipped due to cycle:
  … => (load "mu4e-utils.el") => (macroexpand-all …) => (macroexpand (eval-when-compile …)) => (load "org-mu4e.el") => (load "mu4e.el") => (load "mu4e-headers.el") => (load "mu4e-utils.el")
Auto-evilification could not remap these functions in map ‘mu4e-headers-mode-map’:
   - ‘mu4e-headers-query-prev’ originally mapped on ‘\’
Auto-evilification could not remap these functions in map ‘mu4e-headers-mode-map’:
   - ‘mu4e-headers-query-prev’ originally mapped on ‘\’
Created mu4e link.
Auto-evilification could not remap these functions in map ‘mu4e-headers-mode-map’:
   - ‘mu4e-headers-query-prev’ originally mapped on ‘\’"

*Expected behaviour:* :heart: :smile:
Move around using evil keys

#### System Info :computer:
- OS: gnu/linux
- Emacs: 27.0.50
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 3fe675d)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: helm
- Layers:
```elisp
(csv python bibtex html helm auto-completion better-defaults emacs-lisp git markdown org
     (chinese :variables chinese-enable-fcitx t chinese-fcitx-use-dbus t pyim-default-scheme 'microsoft-shuangpin)
     latex ranger pdf mu4e
     (shell :variables shell-default-height 30 shell-default-position 'bottom)
     version-control
[...]

PR #14549, opened 2021-03-23
Revert "Fix void variable `mu4e-view-mode-map`" to fix issue #14548

zzjjzzgggg:
fix issue #14548


magthe:
The issue remains on the `develop` branch, and this change fixes it.

Why isn't it a no-brainer to pick it up?

duianto:

Issue #14444, opened 2021-03-01
Proposal to improve GPL compliance


gnusupport:
After reading the GPL, and reviewing Spacemacs, I am of the opinion that Spacemacs is not merely an aggregate, it is one full work that includes various other parts. It does not matter if parts come from various places, the whole package is there.

Review:
https://www.gnu.org/licenses/gpl-faq.html#MereAggregation
https://www.gnu.org/licenses/gpl-faq.html#AggregateContainers

I am proposing that authors review the section 5. about Conveying Modified Source Versions, and to see if there is need to comply to it. As it is actually a distribution of Emacs in modified form IMHO. I have not found copy of license inside. I found the reference, but the reference like GPLv3 is not enough to comply to licensing.

Please review:

```
    a) The work must carry prominent notices stating that you modified
    it, and giving a relevant date.

    b) The work must carry prominent notices stating that it is
    released under this License and any conditions added under section
    7.  This requirement modifies the requirement in section 4 to
    "keep intact all notices".

    c) You must license the entire work, as a whole, under this
    License to anyone who comes into possession of a copy.  This
    License will therefore apply, along with any applicable section 7
    additional terms, to the whole of the work, and all its parts,
    regardless of how they are packaged.  This License gives no
    permission to license the work in any other way, but it does not
    invalidate such permission if you have separately received it.

    d) If the work has interactive user interfaces, each must display
    Appropriate Legal Notices; however, if the Program has interactive
    interfaces that do not display Appropriate Legal Notices, your
    work need not make them do so.
```

lebensterben:
The interesting question is, which of the points you listed above is not already met by Spacemacs?

gnusupport:
* Lucius Hu <notifications@github.com> [2021-03-01 12:46]:
> The interesting question is, which of the points you listed above is not already met by Spacemacs?

Thank you. Please note that this comment is with good faith. I have
stated that authors shall review its compliance, as myself I may not
be sure of each detail. I have already stated the details, but you
maybe did not pick it up.

From my side I have downloaded the package and run it by
instructions. So in my opinion the pacakge runs together with Emacs
software and modifies on the fly Emacs, so in my opinion it forms new
work. That is why IMHO, this section applies "5. Conveying Modified
Source Versions.".

Please find more guidance here:
https://www.gnu.org/licenses/gpl-howto.html

Myself, I have not verified all of your repositories, as those are
developers' hosting spaces. I have verified the master.zip that I have
[...]

PR #14552, opened 2021-03-24
Tweak licensing and update years to 2021

syl20bnr:
Should fix https://github.com/syl20bnr/spacemacs/issues/14444

- Added the LICENSE file.
- Added GPLv3 terms explicitly to each elisp file
- Updated the years to `2021`
- Added a button beneath the logo on the home buffer
- Cherry on the cake, added the GPLv3 logo to the README file and in the footer of the home buffer:

image

image


lebensterben:
The boilerplate copyright notice in each file's header is not required by GPL, it's only a recommendation.

We only need one COPYRIGHT file containing ALL license information. One example is https://github.com/rust-lang/rust/blob/master/COPYRIGHT

gnusupport:
> The boilerplate copyright notice in each file's header is not required by GPL, it's only a recommendation.
> 
> We only need one COPYRIGHT file containing ALL license information. One example is https://github.com/rust-lang/rust/blob/master/COPYRIGHT

It is however safest, as each file is software in itself and can separately from whole package, and thus would or could break the chain of proper licensing.

skangas:
This looks great.

BTW, it's very easy to avoid seeing long license statements at the top of all files by adding this to your .emacs:

```
(add-hook 'prog-mode-hook 'elide-head)
```

PR #14554, opened 2021-03-25
git: Added `magit-todos`, improved docs, etc.

lebensterben:
- Added `magit-todos` package, toggled by `git-enable-magit-todos-plugin`.
- Improved installation and loading of all magit-plugins:
  - All magit plugins are toggled by their corresponding layer variables
    PRE-INSTALLATION, instead of PRE-LOADING. That is, they are installed
    only when the toggle is non nil.
  - The `use-package` call is more specific now.
    - Hooks goes to `:hook`.
    - Removed `:defer` since it's implied by `:hook`, and it's actually not
      doing anything.
- Improved the documentation.

### DIFF
--- a/layers/+source-control/git/README.org
+++ b/layers/+source-control/git/README.org
@@ -9,12 +9,15 @@
   - [[#features][Features:]]
 - [[#install][Install]]
   - [[#layer][Layer]]
+  - [[#git][Git]]
   - [[#magit-status-fullscreen][Magit status fullscreen]]
   - [[#magit-auto-complete][Magit auto-complete]]
-  - [[#magit-svn-plugin][Magit SVN plugin]]
+  - [[#magit-plugins][Magit Plugins]]
+    - [[#magit-delta][magit-delta]]
+    - [[#magit-gitflow][magit-gitflow]]
+    - [[#magit-svn][magit-svn]]
+    - [[#magit-todos][magit-todos]]
   - [[#global-git-commit-mode][Global git commit mode]]
-  - [[#git][Git]]
-  - [[#git-flow][Git-Flow]]
   - [[#forge][Forge]]
   - [[#org-integration][Org integration]]
 - [[#working-with-git][Working with Git]]
@@ -25,7 +28,7 @@
   - [[#interactive-rebase-buffer][Interactive rebase buffer]]
   - [[#quick-guide-for-recurring-use-cases-in-magit][Quick guide for recurring use cases in Magit]]
   - [[#git-blame-transient-state][Git Blame Transient State]]
-  - [[#git-flow-1][Git-Flow]]
+  - [[#git-flow][Git-Flow]]
   - [[#git-time-machine][Git time machine]]
   - [[#git-links-to-web-services][Git links to web services]]
   - [[#repository-list][Repository list]]
@@ -53,6 +56,10 @@ To use this configuration layer, add it to your =~/.spacemacs=. You will need to
 add =git= to the existing =dotspacemacs-configuration-layers= list in this
 file.
 
+** Git
+Of course if your OS does not ship with git (!) you'll have to install it
+on your machine. You can download it from the [[http://git-scm.com/downloads][download page]].
+
 ** Magit status fullscreen
 To display the =magit status= buffer in fullscreen set the variable
 =git-magit-status-fullscreen= to =t= in your =dotspacemacs/user-init= function.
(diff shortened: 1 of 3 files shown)


lebensterben:
For reference, `:defer` is implied by
- `:commands`
- `:bind`
- `:bind*`
- `:mode`
- `:magic`
- `:hook`
- `:magic-fallback`
- `:interpreter`

Also
> In almost all cases you don't need to manually specify :defer t. This is implied whenever :bind or :mode or :interpreter is used. Typically, you only need to specify :defer if you know for a fact that some other package will do something to cause your package to load at the appropriate time, and thus you would like to defer loading even though use-package isn't creating any autoloads for you.

Kazark (approved):
Excited to try turning this on.

duianto:
magit-todos shows this message, when a magit buffer is opened:
>magit-todos: Not overriding bind of "jT" in ‘magit-status-mode-map’.

I've added two TODO comments:
- `# TODO abc`
in: `c:/Users/username/AppData/Roaming/.emacs.d/.gitignore`
- `;; TODO abc`
in: `c:/Users/username/AppData/Roaming/.emacs.d/core/core-jump.el`

There is no TODOs section in the magit buffer.
Possibly because of the error message, maybe the package doesn't load when that occurs.

The same message was reported upstream:
magit-todos: Not overriding bind of "jT" in ‘magit-status-mode-map’
https://github.com/alphapapa/magit-todos/issues/28

The package author said that it only should be activated once.
The OP removed the hook and said that it started working.

I tried the OPs suggestion of removing:
```elisp
  :hook (magit-mode . magit-todos-mode)
```
and adding:
```elisp
  :commands (magit-todos-mode)
```
But that doesn't seem to work for me.

When magit todos is started manually then the overriding message appears again.

I found an issue where the error message key binding is mentioned.

And a comment has a possible workaround:
https://github.com/alphapapa/magit-todos/issues/70#issuecomment-570092919
```elisp
(general-define-key
 :keymaps '(magit-todos-section-map magit-todos-item-section-map)
 "jT" nil
 "jl" nil
 "j" nil)
```

But the commenter also mentions that the evil-collection has a magit-todos module.
https://github.com/emacs-evil/evil-collection/blob/master/modes/magit-todos/evil-collection-magit-todos.el

It binds `gT` to jump to the todos section, and it seems to suppress the jT warning.
https://github.com/emacs-evil/evil-collection/blob/b45ec5b3156d27a18a949e4bf55643107a21abc2/modes/magit-todos/evil-collection-magit-todos.el#L50
>  ;; No need to tell me that jT isn't bound

lebensterben:
@duianto 
Yes, I can confirm there's the bug you described. It's most likely to be caused by some conflicts with other evil packages.

Let's put a hold on this PR and I will try to figure that out.

duianto:
This enables `evil-collection-magit-todos`:
```diff
1 file changed, 3 insertions(+)
layers/+source-control/git/packages.el | 3 +++

modified   layers/+source-control/git/packages.el
@@ -300,6 +300,9 @@
               (spacemacs|diminish magit-svn-mode "SVN")
               (define-key magit-mode-map "~" 'magit-svn))))
 
+(defun git/pre-init-magit-todos ()
+  (add-to-list 'spacemacs-evil-collection-allowed-list 'magit-todos))
+
 (defun git/init-magit-todos ()
   (use-package magit-todos
     :hook (magit-mode . magit-todos-mode)
```

`gT` calls `magit-todos-jump-to-todos` which is bound in the evil-collection:
https://github.com/emacs-evil/evil-collection/blob/b45ec5b3156d27a18a949e4bf55643107a21abc2/modes/magit-todos/evil-collection-magit-todos.el#L40

The `overriding` message doesn't appear when a magit buffer is started.

But I'm not getting a TODOs section in the magit status buffer.
(With TODO comments in two files (as mentioned above: https://github.com/syl20bnr/spacemacs/pull/14554#issuecomment-809330845))

duianto:
>But I'm not getting a TODOs section in the magit status buffer.
(With TODO comments in two files (as mentioned above: #14554 (comment)))

It doesn't seem to Spacemacs related, because I'm not seeing a TODOs section without Spacemacs (Emacs and `magit-todos`) either.

And I'm getting the overriding "jT" message when `magit-todos-mode` is enabled.

smile13241324 (approved):
Looks good, I will add `magit-todo` to the list of `evil-collection` controlled packages as suggested by @duianto.

With this the package works well on my side:
grafik

duianto:
@smile13241324 What's your: `nice --version`

There's an open issue upstream:
Windows issues https://github.com/alphapapa/magit-todos/issues/29

About other Windows users not seeing a TODOs section either.

It seems to be related to `nice`:
https://en.wikipedia.org/wiki/Nice_(Unix)
>nice is a program found on Unix and Unix-like operating systems such as Linux. It directly maps to a kernel call of the same name. nice is used to invoke a utility or shell script with a particular CPU priority, thus giving the process more or less CPU time than other processes.

My version of `nice` is:
>nice --version
nice (GNU coreutils) 8.26
Packaged by Cygwin (8.26-2)
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
>
>Written by David MacKenzie.

---

Evaluating this from a magit buffer:
```elisp
(magit-todos-candidates)
```
>user-error: nice failed

---

In the upstream issue they mention that `nice` can be disabled with:
```elisp
(setq magit-todos-nice nil)
```

And that seems to work for me. The `TODOs` section does appear.

---

But now when a magit buffer is opened, then the cursor can't be moved for about 10 seconds.

It seems to be yas (snippets) related, because after a couple of seconds the minibuffer shows:
>[yas] Prepared just-in-time loading of snippets successfully.

When the TODOs section has appeared and the cursor can be moved,
when the cursor has been idle for a second, then the minibuffer shows:
>Error running timer ‘imenu-list-update’: (wrong-type-argument (or eieio-object class) nil obj)

I don't know if it's related, but the HELM Mini buffer list (`SPC b b`), 
shows an `*Ilist*` buffer when a magit buffer has been opened.

It's not magit-todos related, because the `*Ilist*` buffer also appears 
before enabling the `git-enable-magit-todos-plugin` variable.

When the `*Ilist*` buffer is opened
- From a magit buffer, then it contains the recent commits:
```
+ Recent commits
   66d774cd6 develop origin/develop [java] Fix not existing fun
   5bd1603b6 [Fix] `dotspacemacs/test-dotfile` complains error 
   e54affd5a [java] Add back support for emacs 25.1
   d301a88df [java] only download meghanada package if needed
   413930048 [zig] Adjust backend variables to match new style
   be6f82025 Add missing GPLv3 headers to various files
   8c937e9a7 [zig] Fix some smaller issues in the docs
   6bafc7269 Enhance +lang/zig layer
   dc07cdbdc [windows-scripts] Add missing keybinding to docs
   96008496d windows-scripts: replaced dos.el with bat-mode, ad
```
- From `.spacemacs`, then the shows all the functions (defun) in `.spacemacs`.

The `*Ilist*` buffer doesn't appear with just `Emacs` and `magit-todos` installed.
So it seems to be some setting in Spacemacs.

The `imenu-list` layer is not enabled.

<details><summary>System Info (Click To expand)</summary>
<pre>
#### System Info :computer:
- OS: windows-nt
- Emacs: 27.2
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 66d774cd6)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(auto-completion autohotkey c-c++ emacs-lisp
                 (git :variables git-enable-magit-todos-plugin t)
                 helm html lsp markdown multiple-cursors
                      '("c:/Users/username/org/"))
                 shell spell-checking syntax-checking treemacs version-control)
```
</pre>
</details>

PR #14559, opened 2021-03-26
[init] Disable GUI elements in early-init as well

syl20bnr:
Can provide a smoother startup by reducing graphical artifacts.

### DIFF
--- a/core/core-early-funcs.el
+++ b/core/core-early-funcs.el
@@ -1,0 +1,39 @@
+;;; core-early-funcs.el --- Spacemacs Core File
+;;
+;; This file is sourced by emacs early-init.el file.
+;;
+;; Copyright (c) 2012-2021 Sylvain Benner & Contributors
+;;
+;; Author: Sylvain Benner <sylvain.benner@gmail.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+;;
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+;;
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+(defun spacemacs/removes-gui-elements ()
+  "Remove the menu bar, tool bar and scroll bars."
+  ;; removes the GUI elements
+  (when (and (fboundp 'scroll-bar-mode) (not (eq scroll-bar-mode -1)))
+    (scroll-bar-mode -1))
+  (when (and (fboundp 'tool-bar-mode) (not (eq tool-bar-mode -1)))
+    (tool-bar-mode -1))
+  (unless (memq (window-system) '(mac ns))
+    (when (and (fboundp 'menu-bar-mode) (not (eq menu-bar-mode -1)))
+      (menu-bar-mode -1)))
+  ;; tooltips in echo-aera
+  (when (and (fboundp 'tooltip-mode) (not (eq tooltip-mode -1)))
+    (tooltip-mode -1)))
+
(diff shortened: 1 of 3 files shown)


core/core-early-funcs.el
```diff
@@ -0,0 +1,39 @@
...
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+;;
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+(defun spacemacs/removes-gui-elements ()
+  "Remove the menu bar, tool bar and scroll bars."
+  ;; removes the GUI elements
+  (when (and (fboundp 'scroll-bar-mode) (not (eq scroll-bar-mode -1)))
```
lebensterben, line 28: From docstring
> To set this variable in a Lisp program, use set-scroll-bar-mode
to make it take real effect.

smile13241324 (approved):
Looks good.

cjf-666:
I've discovered that disabling GUI elements in early-init causes the `green button` on title bar of Emacs app on macOS to turn into `zoom button`, instead of `full screen button`. This can be annoying obviously.
The root cause of this is `window-system` function will return `nil` when emacs is in early-init,

> (unless (memq (window-system) '(mac ns))
>    (when (and (fboundp 'menu-bar-mode) (not (eq menu-bar-mode -1)))
>      (menu-bar-mode -1)))

and the intention was to avoid disabling `menu-bar-mode` for `mac` and `ns` window system, but this can't be done in early-init stage using `window-system` function, guessing it'd be better to allow user to configure whether to disable GUI elements in early-init stage.

@syl20bnr

lebensterben:
@cjf-666 

By `zoom button`, do you mean maximizing?

Also, can you try <kbd>⌘ cmd</kbd>-<kbd>⌃ ctrl</kbd>-<kbd>F</kbd>? That should be default keyboard shortcut for entering fullscreen mode.

lizhuohua:
> @cjf-666
> 
> By `zoom button`, do you mean maximizing?
> 
> Also, can you try ⌘ cmd-⌃ ctrl-F? That should be default keyboard shortcut for entering fullscreen mode.

I have the same issue. The green button becomes `zoom`:

image

Instead of `fullscreen`:

image

Luckily, `command-ctrl-F` still works.

lebensterben:
okay, so it doesn't break the functionality but only the UI.

cjf-666:
> okay, so it doesn't break the functionality but only the UI.

Actually `command-ctrl-F` works differently from the `full screen button`, the former only `maximizing` Emacs on desktop, but the latter makes full screen out of desktop. Also the `full screen button` is the default since OSX Yosemite, I believe it is a good idea to keep `full screen button` as all other apps do.

Plus, hitting `command-ctrl-F` twice (i.e. maximizing followed by exiting maximizing) can make `zoom button` become `full screen button`, weirdly, but it can change back to `zoom button` if clicked twice (i.e. full screen followed by exiting full screen).

lebensterben:
You can re-enable one or more of the mode in your user-config to figure out what breaks the it.

And then you can report it to upstream.

It's not a Spacemacs bug.

cjf-666:
> It's not a Spacemacs bug.

Actually it is a Spacemacs bug, as I mentioned at the beginning,

> The root cause of this is window-system function will return nil when emacs is in early-init,
> 
>     (unless (memq (window-system) '(mac ns))
>       (when (and (fboundp 'menu-bar-mode) (not (eq menu-bar-mode -1)))
>         (menu-bar-mode -1)))
> 
> and the intention was to avoid disabling menu-bar-mode for mac and ns window system, but this can't be done in early-init stage using window-system function

In other words, commenting out `(spacemacs/removes-gui-elements)` in early-init.el will fix this problem.

lebensterben:
Sorry I missed this part.
You're right.

@syl20bnr 
According to https://www.gnu.org/software/emacs/manual/html_node/emacs/Early-Init-File.html

> the early init file is read before the GUI is initialized, so customizations related to GUI features will not work reliably in early-init.el. By contrast, the normal init files are read after the GUI is initialized. If you must have customizations in the early init file that rely on GUI features, make them run off hooks provided by the Emacs startup, such as window-setup-hook or tty-setup-hook.

We should remove GUI elements via `window-setup-hook`.

cjf-666:
@lebensterben Hi, I have created a PR(https://github.com/syl20bnr/spacemacs/pull/15004) fixing this, thanks for the emacs doc reference

PR #14562, opened 2021-03-26
[Python layer] Add links to the LSP implementations

Sir4ur0n:
* Document how to make Nix + Microsoft LSP work
* Add links to Microsoft LSP and Pyright repos

### DIFF
--- a/layers/+lang/python/README.org
+++ b/layers/+lang/python/README.org
@@ -135,11 +135,11 @@ setting your =PYTHONPATH= as explained at
 ** Language Server Protocol
 The =lsp= backend can use either of the following language server implementations:
 
-| symbol   | description                              |
-|----------+------------------------------------------|
-| 'pyls    | python-language-server package (default) |
-| 'mspyls  | Microsoft python language server         |
-| 'pyright | Microsoft pyright language server        |
+| symbol   | description                                                                      |
+|----------+----------------------------------------------------------------------------------|
+| 'pyls    | python-language-server package (default)                                         |
+| 'mspyls  | [[https://github.com/emacs-lsp/lsp-python-ms][Microsoft python language server]] |
+| 'pyright | [[https://github.com/emacs-lsp/lsp-pyright][Microsoft pyright language server]]  |
 
 =pyls= is used by default - to use the Microsoft python language server, set the
 =python-lsp-server= layer variable as follows:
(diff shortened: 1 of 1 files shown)


layers/+lang/python/README.org
```diff
@@ -203,6 +203,13 @@ To use the latest built version in a cloned git repo, use the ~python-lsp-git-ro
 N.B. If you're using Arch linux or a derivative distribution, you can install the =microsoft-python-language-server=
 package from the AUR.
 
+If you're using the Nix package =python-language-server= (which is the Microsoft Python language server), you may also need to set the =lsp-python-ms-executable= variable:
```
lebensterben, line 206: Some LSP clients actively calls `executable-find` to look for the server executable, and if it doesn't exist, fallbacks to `lsp-server-install-dir` or check the PATH of registered providers. But this is largely leave to the implementation of the lsp-client.

If possible, I suggest to make feature request/PR in upstream of lsp-python-ms so that it calls `executable-find` out-of-box.

As for this PR, I think this is too particular for one package management system/distribution. I don't think we should put this kind of instruction in a layer documentation. I suggest to move this to LSP layer's documentation.

Sir4ur0n: I have removed the Nix part, but I still think linking to the LSP implementations would help users, is that ok for you?

lebensterben: @Sir4ur0n 
> I don't think we should put this kind of instruction in a layer documentation. I suggest to move this to LSP layer's documentation.

You missed the last sentence.
Nix integration certainly goes beyond Python layer. The issue is not really Python LSP per session, but the pattern that nixos packages are installed in special location and sometimes Emacs packages are not smart enough to find them.

That's why I suggested to move that part to LSP layer.

Alternatively, you can also add those information to nixos layer.

Sir4ur0n: I didn't miss it: the `lsp-python-ms` readme already has a NixOS dedicated section hinting at the `(setq lsp-python-ms-executable (executable-find "python-language-server"))` bit to make it work.

I don't really have the time nor will to start digging in the LSP MS python repo to discuss such change, all the more if they already documented it (I guess if they did so, they are already aware of this limitation)

I just wanted to save time for others, as I lost some time on it myself, thinking it would be quick :slightly_smiling_face:

smile13241324 (approved):
Adding the links here is certainly ok.

smile13241324:

PR #14564, opened 2021-03-27
bootstrap: new :spacediminish keyword to use-package

lebensterben:
Added a new keyword to `use-package`, `:spacediminish` which calls
`spacemacs|diminish`.

Supported patterns:
- nil
- SYMBOL
- STRING
- (SYMBOL STRING)
- (STRING STRING)
- (SYMBOL STRING STRING)
- List of patterns above

When `SYMBOL` is missing, it's inferred from the package name, e.g
`(use-package foo :spacediminish nil)` becomes
`(use-package foo :spacediminish foo-mode)`.

This PR aims to replace explict calls to `spacemacs|diminish` and brings more
declarative flavour to layer configuration.

For example, the following form

```elisp
(use-package highlight-indentation
  :config
  (progn
    (spacemacs|diminish highlight-indentation-mode " ⓗi" " hi")
    (spacemacs|diminish
     highlight-indentation-current-column-mode " ⓗc" " hc")))

```
can be replaced by

```elisp
(use-package highlight-indentation
  :spacediminish ((" ⓗi" " hi")
                  (highlight-indentation-current-column-mode " ⓗc" " hc")))
```


lebensterben:
There's one caveat:
The current approach of calling `spacemacs|diminish` inside `:config` is evaluated earlier than the new approach.
Specifically, `:spacediminish` is evaluated after `post-config-hook`.

smile13241324 (approved):
This looks pretty cool, lets have this merged, we can still optimise where necessary and use it step by step.

About the evaluation order, I don't think that having this evaluated after post-config-hook should be a big issue.

smile13241324:

PR #14569, opened 2021-03-28
haskell: reworked

lebensterben:
- Improved formatting capability via `format-all` and/or `lsp`.
  - Supported backends are:
    - `brittany`
    - `hindent`
    - `ormolu`
    - `stylish-haskell`
    - `fourmolu` (only work with `lsp`)
    - `floskell` (only work with `lsp`)
  - Added new layer variable `haskell-formatter-backend`.
  - The old layer variable `haskell-enable-hindent` is superseded, but for backward compatibility, if `haskell-formatter-backend` is not set explicitly, it defaults to `hindent` should `haskell-enable-hindent` is non-nil.
  - Also, the package `hindent` is superseded by `format-all`.
  - When `lsp` layer is enabled, the chosen formattr backend is also registered with `lsp`.
- Added new layer variable `haskell-enable-cmm-mode` and `haskell-enable-hlint`
  - Both integrations are available for a long time, and this PR just make them optional.
- Improved support for `haskell-literate-mode`
  - When using `dante`, company backend is enabled.
  - `gtags` integration is added (via `helm-gtags` and ``counsel-gtags` respectively).
  - `yasnippet` integration.
  - The obsolete name `literate-haskell-mode` is removed.
- Updated `yasnippet` initialization:
  - The old way of loading `yasnippet` is still valid, but this PR simplifies it by calling the the functions in upstream package.
- Fixed key-binding problems:
  - Many bindings are missing due to wrong settings in `prefix`. This PR fixed them. Prefixes should work with both `lsp` and `dante`.
- Moved `defun` forms in `packages.el` to `funcs.el`
- Other minor modifications in style.

### DIFF
--- a/layers/+lang/haskell/config.el
+++ b/layers/+lang/haskell/config.el
@@ -23,14 +23,46 @@
 
 ;; Variables
 
-(setq haskell-modes '(haskell-mode haskell-literate-mode))
+(defconst spacemacs--haskell-modes '(haskell-mode haskell-literate-mode))
 
 (spacemacs|define-jump-handlers haskell-mode haskell-mode-jump-to-def-or-tag)
+(spacemacs|define-jump-handlers haskell-literate-mode haskell-mode-jump-to-def-or-tag)
 
 (defvar haskell-completion-backend (if (configuration-layer/layer-used-p 'lsp) 'lsp 'dante)
   "Completion backend used by company.
 Available options are `dante' and `lsp'.
-If `nil' then `dante' is the default backend unless `lsp' layer is used.")
+If nil then `dante' is the default backend unless `lsp' layer is used.")
 
+;; TODO: verify whether `stylish-haskell' breaks flycheck highlighting
+;; reported by taksuyu 2015-10-06
+(defvar haskell-formatter-backend (cond
+                                   (haskell-enable-hindent 'hindent)
+                                   ((executable-find "brittany") 'brittany)
+                                   ((executable-find "ormolu") 'ormolu)
+                                   ((and (executable-find "fourmolu")
+                                         (eq haskell-completion-backend 'lsp))
+                                    'fourmolu)
+                                   ((and (executable-find "floskell")
+                                         (eq haskell-completion-backend 'lsp))
+                                    'floskell))
+  "Formatter backend for `haskell-mode' and `haskell-literate-mode'.
+Valid values are: `brittany', `hindent', `ormolu', `stylish-haskell',
+`fourmolu', `floskell' and nil. `fourmolu' and `floskell' are valid only when
+`haskell-completion-backend' is `lsp'.
+For backward compatibility, defaults to `hindent' when `haskell-enable-hindent'
+is set.
+Otherwise, use the first formatter found in `brittany', `ormolu', `fourmolu' and
+and `floskell'.
+`stylish-haskell' is never used as default due to past report of issues.")
+
+(defvar haskell-enable-cmm-mode nil
+  "When non-nil, enable cmm-mode for C-- language.")
+
+(defvar haskell-enable-hlint t
+  "When non-nil, enable hlint integration.
+When `haskell-completion-backend' is `lsp', use its builtin support.
+When it's `dante', use `hlint-refactor-mode'.")
+
+;; This should be deprecated at some point, use haskell-formatter-backend instead
 (defvar haskell-enable-hindent nil
-  "Formatting with hindent; If t hindent is enabled.")
+  "When non-nil, use hindent as formatter.")
--- a/layers/+lang/haskell/packages.el
+++ b/layers/+lang/haskell/packages.el
@@ -20,44 +20,51 @@
 ;; You should have received a copy of the GNU General Public License
 ;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+(defconst haskell-packages
+  '(
+    ;; auto-completion
+    company
+    (company-cabal :requires company)
 
-(setq haskell-packages
-      '(
-        cmm-mode
-        company
-        (company-cabal :requires company)
+    ;; syntax-checking
+    flycheck
+    (flycheck-haskell :requires flycheck)
 
-        ;; dante completion backend
-        (dante :requires company)
-        ;; dante auto refactor companion
-        (attrap :requires dante)
+    ;; gtags
+    ggtags
+    counsel-gtags
+    helm-gtags
 
-        lsp-haskell
+    ;; yassnippet
+    yasnippet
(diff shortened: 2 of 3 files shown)


lebensterben:
The layer documentation haven't been updated. I will update it if the above passes review.

layers/+lang/haskell/packages.el
```diff
@@ -68,123 +74,145 @@
...
+        (dolist (mode modes)
+          (spacemacs/set-leader-keys-for-major-mode mode
+            "==" 'format-all-buffer
+            "T=" 'format-all-mode))))
+      ;; also set lsp formattign provider
+      ;; NOTE: except of 'hindent which is not currently supported (April 2021)
+      (when (and (eq haskell-completion-backend 'lsp)
+                 (memq haskell-formatter-backend
+                       '(brittany ormolu stylish-haskell fourmolu floskell)))
+        (setq lsp-haskell-formatting-provider (symbol-name haskell-formatter-backend))))))
```
michaelpj, line 178: Does this work properly, given that `lsp-haskell-formatting-provider` is a `defcustom`? Can't we just let people set the already-existing customization variable that controls this, rather than adding another layer of abstraction here?

lebensterben: The thing is `hindent` doesn't work with `lsp-haskell`, so I need to add this test.
Also, there was an old variable `haskell-enable-hindent`, I don't want to cause a hard beak. So for compatibility ,`haskell-formatter-backend` would default to `hindent`.

layers/+lang/haskell/config.el
```diff
@@ -23,14 +23,42 @@
...
-(defvar haskell-completion-backend nil
+(defvar haskell-completion-backend (if (configuration-layer/layer-used-p 'lsp) 'lsp 'dante)
   "Completion backend used by company.
 Available options are `dante' and `lsp'.
-If `nil' then `dante' is the default backend unless `lsp' layer is used.")
+If nil then `dante' is the default backend unless `lsp' layer is used.")
 
+;; TODO: verify whether `stylish-haskell' breaks flycheck highlighting
+;; reported by taksuyu 2015-10-06
+(defvar haskell-formatter-backend (or haskell-enable-hindent
```
lebensterben, line 38: I spotted some problem here. Need changes.

layers/+lang/haskell/packages.el
michaelpj: I mean, why not do:
- `haskell-enable-hindent`: enable `hindent`, unless the backend is `lsp`, in which case give an error.
- No other formatting configuration: `lsp` users can use the configuration variables provided by `lsp-haskell`

michaelpj: Also, sorry I'm not that familiar with `format-all`, do we want it if we have the `lsp` support for formatting?

layers/+lang/haskell/packages.el
```diff
@@ -67,123 +74,145 @@
...
-       haskell-stylish-on-save nil))
+    :hook
+    (((haskell-mode-local-vars haskell-literate-mode-local-vars) . spacemacs-haskell//setup-backend)
+     (haskell-cabal-mode . spacemacs//force-haskell-mode-loading)
+     ((haskell-mode haskell-cabal-mode) . spacemacs//haskell-disable-electric-indent))
+    :custom
+    (haskell-notify-p t "Notify with notify.el, if installed, after running Cabal commands and etc")
+    (haskell-interactive-popup-errors nil "Remove annoying error popups")
+    (haskell-process-suggest-remove-import-lines t "Show unused imports")
+    (haskell-process-auto-import-loaded-modules t "Auto-import module")
```
michaelpj, line 193: `lsp` users probably don't want any of the `haskell-process-` variables set, since they won't be using that!

michaelpj (commented):
Generally looks nice.

lebensterben:
Still need some rework. There's issue with loading order.

layers/+lang/haskell/packages.el
lebensterben: @michaelpj 

The new behaviour would be:
- If `lsp` is not used, and `haskell-enable-hindent` is non-nil, use `hindent.
- If `lsp` is used, and `haskell-completion-backend` is non-nil, use its value.
- Otherwise, search for a supported formatter in the PATH.

Does this look good for you?

michaelpj: > If lsp is used, and haskell-completion-backend is non-nil, use its value.

I think you mean `haskell-formatting-provider` there?

> Otherwise, search for a supported formatter in the PATH.

I'd prefer not to have a fallback like this, it's very implicit. I might have multiple formatters, or none, and in any case I'd like to use whatever I've set up with `lsp` instead of those!

I didn't quite understand what you were suggesting, are you suggesting to keep `haskell-formatting-provider` as a variable or not?

michaelpj: I think this is an admirable attempt to make everything work. I guess I wonder if we could get away with trying less hard:

- Don't provide any formatting options when using `dante` (there are none currently!)
- Don't provide any formatting options when using `lsp` (the `lsp-haskell` options do the job)
- Remove `haskell-enable-hindent` (you can enable `hindent-mode` yourself if you still want it)

I'm biased, of course, since I hope the community will continue to converge on the LSP-based solutions! But it would be nice to actually benefit from that and just leave the bulk of the configuration to `lsp-mode` and `lsp-haskell`.

smile13241324 (changes requested):
In general I like the refactoring you did, also the separate support of the formatter is a good choice. Still I think these changes are so big we need to have a proper test with this before this can be merged.

Having a separate variable for a formatter should not be overkill, we have implemented a similar setup for other layers already (python for example) this tends to work pretty well.

We should at least go through a setup of each backend, not necessarily with each formatter. I can cover LSP with `hie` if required. But we would still have someone use `dante`. I am also not familiar with `format-all` but if it supersedes existing packages we should deprecate these i.e. show a warning if this is still used.

LuisChDev:
I can do the test run, but first I need to know if the current state of the code is good-to-go

lebensterben:
@LuisChDev 

It's almost good but there could be a few issues.
For example some forms may need to be moved from :config to :init or may be wrapped in a with-eval-after-load.

smile13241324:
I am going to restart my work with Haskell in the near future, lets take the opportunity and test this PR.

lebensterben:
I probably will close this PR and split it into a few smaller ones.

smile13241324:
Sounds cool, then I'll keep my fingers from this ticket. Just tag me if you have your first bunch of changes ready.

smile13241324:

PR #14571, opened 2021-03-28
[finance] Enable flycheck when syntax-checking-enable-by-default

bcc32:
This enables flycheck-ledger by default in ledger-mode buffers when syntax-checking-enable-by-default is non-nil.


layers/+tools/finance/packages.el
```diff
@@ -74,4 +78,9 @@
       (add-hook 'ledger-mode-hook 'evil-normalize-keymaps)
       (add-hook 'ledger-mode-hook
                 (lambda () (setq-local pcomplete-termination-string "")))
+      ;; global-flycheck-mode is enabled lazily by prog-mode-hook, but
+      ;; ledger-mode derives from text-mode
+      (spacemacs|add-transient-hook ledger-mode-hook
+        (lambda () (when syntax-checking-enable-by-default
+                     (global-flycheck-mode 1))))
```
smile13241324, line 85: Can you explain this a bit more? Looks like you will globally activate flycheck here. I have though this should be handled automatically for you, after enabling ledger-mode.

smile13241324 (changes requested):

layers/+tools/finance/packages.el
bcc32: This is basically adapted from `syntax-checking/init-flycheck`, which contains:

```elisp
(spacemacs|add-transient-hook prog-mode-hook
        (lambda () (when syntax-checking-enable-by-default
                     (global-flycheck-mode 1)))
        lazy-load-flycheck)
```

I agree this isn't the most elegant way to do this, but I'm not sure what would be.  I think this code is also broken (e.g., `syntax-checking` layer not used, and it is missing the `fname` argument to `spacemacs|add-transient-hook`).

Any suggestions?

bcc32: Okay, I thought this over a bit.  I think this change is fine.  In particular, Spacemacs already globally activates flycheck upon opening any buffer whose major mode derives from prog-mode.  `global-flycheck-mode` only has any effect in modes listed in `flycheck-global-modes`.

The only purpose of the hook here is to defer activating `global-flycheck-mode` (and thus loading flycheck) until it would actually have any effect, but in principle it would make no difference (except to load time) to just do `(global-flycheck-mode 1)` at top level.

smile13241324: You are right, I wonder when will Spacemacs stop to suprise me 🤔.
Lets keep your code and thanks for investigating.

bhipple:
I had no idea that there was flycheck support for ledger-mode.  This is a huge quality of life improvement for me, particularly on validating balance assertions.  Thanks for all of your hard work @bcc32 and @smile13241324!

PR #14583, opened 2021-03-31
 Fixed a bug in notmuch-tree which was preventing d/D bindings to work

agenbite:
The bindings `d`/`D` were not working in `notmuch-tree-mode`.

The cause is that `spacemacs/notmuch-message-delete` calls `notmuch-search-tag` and this function does not work in tree-mode. 

The solution is to duplicate `spacemacs/notmuch-message-delete` into `spacemacs/notmuch-search-message-delete` and `spacemacs/notmuch-tree-message-delete`, adapting the inner calls to the corresponding functions of each mode.


smile13241324 (approved):

PR #14587, opened 2021-04-02
Add docs agenda usage recommendations

dalanicolai:
The Emacs documentation is not very clear about the best way to use org-agenda,
especially regarding integration of the diary into org. This PR adds most
sensible configuration recommendations. More info in the commit messages.

### DIFF
--- a/layers/+emacs/org/README.org
+++ b/layers/+emacs/org/README.org
@@ -12,6 +12,7 @@
 - [[#install][Install]]
   - [[#layer][Layer]]
   - [[#agenda-notifications][Agenda notifications]]
+  - [[#agenda-recommendations][Agenda recommendations]]
   - [[#github-support][GitHub support]]
   - [[#twitter-bootstrap-support][Twitter Bootstrap support]]
   - [[#gnuplot-support][Gnuplot support]]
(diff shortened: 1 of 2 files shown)


layers/+emacs/org/README.org
```diff
@@ -125,28 +126,54 @@ To enable notifications for agenda events, set the variable
...
+
+Subsequently, to add appointments to the agenda you could use the following
+syntax:
+#+begin_src org
+  ,* Friday, 04/02/2021
+    ,** Appointment 1
+    <2021-04-02 Fri 10:10> 
+    ,** TODO Appointment 2
+    <2021-04-02 Fri 10:11>
+#+end_src
```
smile13241324, line 170: Can you double check that org snippet? I am just wondering why most of the lines start with a `,` I would expect something like this:
``` org
    * Friday, 04/02/2021
    ** Appointment 1
    <2021-04-02 Fri 10:10> 
    ** TODO Appointment 2
    <2021-04-02 Fri 10:11>
```

smile13241324 (changes requested):
Looks ok, it should be ok to temporarily move the agenda source to your fork.
However I have a question about the org snippet you have posted.

layers/+emacs/org/README.org
dalanicolai: It is not possible to use the "normal" syntax in code blocks, it would just be recognized as being part of the org document its outline (which indeed happened without those comma's). This behavior and the `,` prefix solution is documented [here](https://orgmode.org/manual/Literal-Examples.html).

Kazark: TIL

dalanicolai: TIL me too... I had to look up its meaning

smile13241324: Lol and just another kind of obscurity within the emacs world :), TIL indeed.

smile13241324 (approved):

PR #14588, opened 2021-04-03
[pocket] Add pocket layer

bcc32:
Add a layer for[ pocket-reader.el](https://github.com/alphapapa/pocket-reader.el/), an Emacs client for the [Pocket](https://getpocket.com/) reading list service.

I wasn't sure whether this would be classified under "readers" or "web services", so I picked one arbitrarily.  Happy to change.


lebensterben:
what's the license of the IMG file?

bcc32:
Hmm, that's a good question.  I downloaded this one off the internet somewhere but there's an official press kit here: https://blog.getpocket.com/press/ that seems more appropriate to use.  That being said, I downloaded that zip file and didn't see a license file (although, clearly, there is *some* license granted to use it).

Is there a good place to record the license information somewhere?  Perhaps just the README file?

lebensterben:
There's no convention for recording license information of non-code files currently.

Since there's no license, it's better not to use it.

bcc32:
That seems reasonable, but there are plenty of other logos (`+web-services/twitter/img/twitter.png`, `+chat/slack/img/slack.png`, `+lang/go/img/go.png`) used without apparent license.  I searched through the commit history and found that logo images are never attributed or really mentioned.

We should probably add license files/attribution for all of the logos in a separate change.

lebensterben:
good idea

smile13241324:
Manually merged into develop, thank you for contributing to Spacemacs.
About the pictures lets do that in a separate PR.

PR #14592, opened 2021-04-04
react: minor change

lebensterben:
This amended commit 58243c67f88ca8b54f2891b05046a119292f6b13

There is an unnecessary `when-let*` form, and it's replaced by `when`.


duianto:

PR #14594, opened 2021-04-04
fixup! python: refactor

lebensterben:
Added the missing branch `'pyls`, so it won't evaluates to `(require nil)` and causes error.
Also added a catch all branch to report the error.


duianto:

PR #14602, opened 2021-04-05
fixup! purescript: refactor

lebensterben:
spacemacs|add-company-backends is a macro, so arguments are taken as-is, instead of being evaluated.

The previous refactor would not cause error, but it would not work either.

The arguments to :backends must take the form of:

- One or multiple unquoted symbol
- A list of symbol
- Lists of symbols


duianto:
The Validate PR test failed with the message: `Please rebase your PR`

No action is required, the following command rebases it automatically.
/rebase

smile13241324 (approved):
This looks ok, will be merged in a minute.

PR #14612, opened 2021-04-06
[CI]use comment based PR rebase

JAremko:
So the cron based auto rebase action seems to be broken.  So we'll have to use the comment based one.
But to run tests on rebase it needs private token (not the default `${{ secrets.GITHUB_TOKEN }}`) 

Token permissions(`repoi:invite` and `security_enets` seems unnecessary  :thinking:  ):


JAremko:
Her's example with the private token:

```yaml
name: Automatic Rebase
on:
  issue_comment:
    types: [created]
jobs:
  rebase:
    name: Rebase
    if: github.event.issue.pull_request != '' && contains(github.event.comment.body, '/rebase')
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
        with:
          token: ${{ secrets.PAT_TOKEN }}
          fetch-depth: 0 # otherwise, you will failed to push refs to dest repo
      - name: Automatic Rebase
        uses: cirrus-actions/rebase@1.4
        env:
          GITHUB_TOKEN: ${{ secrets.PAT_TOKEN }}
```

@syl20bnr  Can you  swap the default token with the private one?  Only you can add it.

PR #14619, opened 2021-04-09
Fix bug in `spacemacs//python-setup-backend ()`

jaminsore:
The PR splits the single `when` into two so that `python-poetry-activate` is actually checked. In the current implementation `(spacemacs//poetry-activate)` is called regardless of the value of `python-poetry-activate`.


smile13241324 (approved):
Ups, good catch thanks for fixing.

jaminsore:
No problem, thanks for getting this merged so quickly!

PR #14622, opened 2021-04-12
[defaults] add qickrun

thanhvg:
bind it to `SPC x x`, not sure it's the right place but it's easy to invoke.

### DIFF
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -212,6 +212,7 @@
     - [[#error-transient-state][Error transient state]]
     - [[#custom-fringe-bitmaps][Custom fringe bitmaps]]
   - [[#compiling][Compiling]]
+  - [[#quickrun][Quickrun]]
 - [[#editorconfig][EditorConfig]]
 - [[#emacs-server][Emacs Server]]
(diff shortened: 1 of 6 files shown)


lebensterben:
What's the use case?
The functionality seems to be covered already.

doc/DOCUMENTATION.org
```diff
@@ -3688,6 +3689,14 @@ Spacemacs binds a few commands to support compiling a project.
 | ~SPC c m~   | =helm-make=                    |
 | ~SPC c r~   | recompile                      |
 
+** Quickrun
+Spacemacs can excute current buffer or region.
```
smile13241324, line 3693:
```suggestion
Spacemacs can execute current buffer or region.
```

smile13241324 (changes requested):
Looks ok, just found a small typo in the docs, the rest is ok.

I am not sure whether `SPC x x` is the right place, one could read it as `text -> execute`.
However something like eval would be better, however I don't see a better matching place in the global key map right now, maybe compile? Not sure, lets keep it in `SPC x x` and see how it works out.

smile13241324 (approved):
:+1:

smile13241324:

PR #14623, opened 2021-04-12
[visual][popwin] add keybiding to resume last popwin buffer

thanhvg:
bind it to `spc r w`

### DIFF
--- a/layers/+spacemacs/spacemacs-visual/config.el
+++ b/layers/+spacemacs/spacemacs-visual/config.el
@@ -1,0 +1,24 @@
+;;; config.el --- Spacemacs UI Visual Layer Configuration File
+;;
+;; Copyright (c) 2012-2021 Sylvain Benner & Contributors
+;;
+;; Author: Sylvain Benner <sylvain.benner@gmail.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+;;
+;; This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


Kazark:
I noticed that some Helm searches can be resumed with `SPC s l` but not always. Is this related at all? I would really love to be able to resume all Helm searches.

thanhvg:
No this just resumes popwin buffer such as *help* buffer,  lsp document at point buffer etc. 

Re helm I don't know, helm seems to prevents some commands from resuming I switched to ivy and it seems that ivy can resume almost anything. Btw I also bind `spc r L` to `C-u spc r l` not sure if it's worth a PR but it is extremely useful.

layers/+spacemacs/spacemacs-visual/config.el
```diff
@@ -0,0 +1,13 @@
+;;; config.el --- Spacemacs UI Visual Layer packages File
+;;
+;; Copyright (c) 2012-2021 Sylvain Benner & Contributors
+;;
+;; Author: Sylvain Benner <sylvain.benner@gmail.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;;; License: GPLv3
```
smile13241324, line 10: This line must be replaced with the proper GPLv3 header

smile13241324 (changes requested):
Nice change, only missing thing is the GPLv3 header we have been asked to add, otherwise we will get into trouble with the FSF.

smile13241324:
Yeah helm is a buggy beast, still I largly prefer it to ivy which I find so much lacking in comparison.

Kazark:
@smile13241324 off the top of my head, certainly I can't resume `SPC s s`. I'll post the other ones here as a re-encounter them.

smile13241324 (approved):
:+1:

smile13241324:

smile13241324:
Interesting, never noticed that there is a difference between the two. Thanks for the hint I will double check

PR #14627, opened 2021-04-12
Search Engine layer: Allow customization of Amazon site TLD.

pataquets:
Amazon search engine is of limited use if you live overseas :)

### DIFF
--- a/layers/+web-services/search-engine/packages.el
+++ b/layers/+web-services/search-engine/packages.el
@@ -38,7 +38,9 @@
       (setq search-engine-alist
             `((amazon
                :name "Amazon"
-               :url "https://www.amazon.com/s/ref=nb_sb_noss?url=search-alias%%3Daps&field-keywords=%s")
+               :url (concat "https://www.amazon."
+                            search-engine-amazon-tld
+                            "/s/ref=nb_sb_noss?url=search-alias%%3Daps&field-keywords=%s"))
               (bing
                :name "Bing"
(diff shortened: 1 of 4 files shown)


layers/+web-services/search-engine/packages.el
```diff
@@ -38,7 +38,8 @@
       (setq search-engine-alist
             `((amazon
                :name "Amazon"
-               :url "https://www.amazon.com/s/ref=nb_sb_noss?url=search-alias%%3Daps&field-keywords=%s")
+               :url (concat (format "https://www.amazon.%s" search-engine-amazon-tld)
```
lebensterben, line 41: You can combine concat and format into one format form.

pataquets: Thanks for the prompt review, @lebensterben.
I've had some problems with that because of insufficient arguments for format. I guess that last `%s` must be escaped some way. Disclaimer: Elisp newbie here :sweat_smile:
My only concern, if I should go for that, is the lost readability, which made me do it this way in the end. Any advice?

lebensterben: Well, you can (concat) everything.

pataquets: Done. I can't remember exactly which, but I was getting some type error when doing it this way in the beginning, so I worked around it but using `format`. Still finding my way around Elisp... 😅

smile13241324 (approved):
Good change, will be merged soon

Issue #14614, opened 2021-04-08
vim layer key binding of "J" in dired mode


wztdream:
In vim layer the "J" was bind to `dired-go-to-file` but this function is weak, it can only jump to the lines in dired buffer, can not jump to sub directory.
I use ivy layer, after press "J" in dired mode, the ivy completion works fine for the files in sub directory, but after press enter nothing happens, this is wired and it is due to the limited function of dired-go-to-file.
I also noticed that in in **spacemacs-bootstrap** layer, "J" was remaped to other function, and the reason is also the limited function of dired-go-to-file, so I think it is better to remap it to `spacemacs/counsel-find-file` which is more powerful and works fine with ivy
```
  ;; replace `dired-goto-file' with `helm-find-files', since `helm-find-files'
  ;; can do the same thing and with fuzzy matching and other features.
  (with-eval-after-load 'dired
    (evil-define-key 'normal dired-mode-map "J" 'spacemacs/helm-find-files)
    (define-key dired-mode-map "j" 'spacemacs/helm-find-files)
```

smile13241324:
Fix is merged care to have another look @wztdream?

PR #14629, opened 2021-04-12
[dired] Add ivy command and replace evil collection key

duianto:
Fixes: vim layer key binding of "J" in dired mode #14614

problem
The evil collection binds "J" to `dired-go-to-file`

It overrides the Spacemacs `helm` and `ivy` equivalent commands.
`spacemacs/helm-find-files' has fuzzy matching and other features
`spacemacs/counsel-find-file' has more `M-o' actions

And the `ivy` layer uses the `spacemacs/helm-find-files` command.

solution
Move the Spacemacs definitions of "J",
after the evil collection dired keys have been setup
from: `spacemacs-bootstrap/packages.el`
to: `spacemacs-evil/init-evil-collection`

And use `spacemacs/counsel-find-file` in the `ivy` layer.

### DIFF
--- a/layers/+distributions/spacemacs-bootstrap/packages.el
+++ b/layers/+distributions/spacemacs-bootstrap/packages.el
@@ -292,13 +292,13 @@
   ;; turn off evil in corelv buffers
   (add-to-list 'evil-buffer-regexps '("\\*LV\\*"))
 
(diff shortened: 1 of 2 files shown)


lebensterben (approved):

smile13241324 (approved):
👍

PR #14630, opened 2021-04-12
fixup! windows-scripts: replaced dos.el with bat-mode, added bmx-mode

### LINKED ISSUE #14626: [bug][regression] press "enter" or mouse-1 in org-mode will enter "bat-mode"
Press "enter" or mouse-1 in org-mode will enter "bat-mode", it's an regression issue.
Enable both `org` and `windows-scripts` in configuration layer, then open an `*.org` file, press `enter` will reproduce the issue.

lebensterben:
Fixes #14626 

Now the two convenient bindings are only set when the user explicitly invoke `windows-scripts/bat-outline`

### DIFF
--- a/layers/+lang/windows-scripts/funcs.el
+++ b/layers/+lang/windows-scripts/funcs.el
@@ -20,11 +20,6 @@
 ;; You should have received a copy of the GNU General Public License
 ;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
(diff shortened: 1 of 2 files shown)


choffee (commented):
That makes sense to me. It only sets the Enter/Mouse shortcuts when bat mode has switched to outline mode not all outline modes now.
Thanks.

smile13241324 (approved):
👍

Issue #5577, opened 2016-03-24
Add a help-describe for evil-ex-commands


gilch:
Unless I've completely missed something important (possible), Evil's ex commands are not discoverable enough. They are not the same as Vim's (Vim has a lot more for one thing), so even experienced Vim users don't know what's there. 

I'm proposing adding a help-describe evil-ex-command function. It would work in helm like `SPC h d f` does now. Possible mnemonics are `SPC h d x`, `SPC h d e`, `SPC h d :`. Alternatively, if possible, just add the `evil-ex-commands` to the existing `SPC h d f` search.

PR #14634, opened 2021-04-13
Add describe evil ex-command functionality

dalanicolai:
For introspection purposes it can be nice to have this functionality. Also this
function can be used to get an overview of which ex-commands are available.

### DIFF
--- a/core/core-debug.el
+++ b/core/core-debug.el
@@ -236,6 +236,22 @@ seconds to load")
                      "You can paste it in the gitter chat.\n"
                      "Check the *Messages* buffer if you need to review it"))))
 
+(defun spacemacs/describe-ex-command (ex-command)
+  (interactive (list (completing-read "Describe ex-command: " evil-ex-commands)))
+  (let* ((func (alist-get ex-command evil-ex-commands nil nil 'string=))
+         (prompt (if (stringp func)
+                     "The ex-command :%s is an alias for the ex-command :%s. Describe :%s?"
+                   "The ex-command :%s calls %s. Describe %s?")))
+    (when (y-or-n-p (format prompt
+                            ex-command
+                            func
+                            func))
+      (if (stringp func)
+          (spacemacs/describe-ex-command func)
+        (if (configuration-layer/layer-usedp 'helpful)
+            (helpful-callable func)
+          (describe-function func))))))
+
 (defun spacemacs/report-issue (arg)
   "Open a spacemacs/report-issue-mode buffer prepopulated with
    issue report template and system information.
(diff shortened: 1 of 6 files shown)


Kazark (approved):
Slick

dalanicolai:
Haha! Yeah it is a funny one...

dalanicolai:
Probably quite unnecessary to refer to the already closed request #5577

core/core-debug.el
```diff
@@ -236,6 +236,20 @@ seconds to load")
...
+         (prompt (if (stringp func)
+                     "The ex-command :%s is an alias for the ex-command :%s. Describe :%s?"
+                   "The ex-command :%s calls %s. Describe %s?")))
+    (when (y-or-n-p (format prompt
+                            ex-command
+                            func
+                            func))
+      (if (stringp func)
+          (spacemacs/describe-ex-command func)
+        (describe-function func)))))
```
lebensterben, line 251: If `helpful` layer is enabled, we can use `(helpful-callable func)` instead of `(describe-function func)`.
The output is much more readable.

lebensterben (changes requested):

core/core-debug.el
dalanicolai: Thanks for pointing out, I have added it. On the other hand, the helpful layer could make `describe-function` an `alias` for `helpful-callable` (and document it is aliased by the helpful layer in the docstring).

dalanicolai: Making it an alias, also keeps the `ivy-actions` available. Right now, with the helpful layer, I loose the ivy actions `Info` and `definition` (which I use all the time) when using `SPC h d f`.

dalanicolai: On the other hand, I already had a personal helpful layer, where I bound the helpful function under `SPC h h f` (etc.) , which is an alternative solution. The command is not called `describe-function` anymore, but `helpful-callable`, so that made more sense to me...

dalanicolai: I would vote to change those bindings in the helpful layer

dalanicolai: or provide a configuration variable like `helpful-prefer-alias`

lebensterben: I have
```elisp
  (global-set-key (kbd "C-h f") #'helpful-callable)
  (global-set-key (kbd "C-h v") #'helpful-variable)
  (global-set-key (kbd "C-h k") #'helpful-key)
  (global-set-key (kbd "C-c C-d") #'helpful-at-point)
  (global-set-key (kbd "C-h F") #'helpful-function)
  (global-set-key (kbd "C-h o") #'helpful-symbol)
```
in my `user-config`.
The issue with alias is, the first time when you call `describe-*`, it won't automatically use `helpful` package. Because it's not loaded...

smile13241324 (approved):
Looks good.

smile13241324:
Hmmm not a bad idea about the alias in helpful layer, we just need to make sure to add the new third parameter to the alias call otherwise this will break emacs 28 again.

Issue #14631, opened 2021-04-12
Emacs 28 startup error from define-obsolete-variable-alias: Wrong number of arguments: (3 . 4), 2


jmromer:
#### Description :octocat:

As of commit 85d0aa2b18c07e, Emacs raises an error on startup.

The error appears to be from `define-obsolete-variable-alias` and only in Emacs 28.

Related:

https://github.com/emacs-lsp/lsp-mode/issues/2497
https://github.com/syl20bnr/spacemacs/issues/14265

#### Reproduction guide :beetle:

1. Start Emacs.
2. Observe an error message.

*Observed behaviour:* :eyes: :broken_heart:

The following error message:

```
Warning (initialization): An error occurred while loading ‘~/.emacs.d/init.el’:

Wrong number of arguments: (3 . 4), 2

To ensure normal operation, you should investigate and remove the
cause of the error in your initialization file.  Start Emacs with
the ‘--debug-init’ option to view a complete error backtrace. Disable showing Disable logging
```

*Expected behaviour:* :heart: :smile:

No error message, successful startup.

#### System Info :computer:
- OS: darwin
- Emacs: 28.0.50
- Spacemacs: 0.300.0
- Spacemacs branch: working (rev. a4eadc97f)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: helm
- System configuration features: ACL GIF GLIB GMP GNUTLS IMAGEMAGICK JPEG JSON LCMS2 LIBXML2 MODULES NATIVE_COMP NOTIFY KQUEUE NS PDUMPER PNG RSVG THREADS TIFF TOOLKIT_SCROLL_BARS XWIDGETS ZLIB

fosskers:
Confirming that `git reset --hard a4eadc97fecaa4f7c5776d24c8ec55379fbc876e` (i.e. the commit *before* the commit hash mentioned above) fixes the issue.

MHOOO:
Fixed it locally, by adding the following code at the top of my `.emacs.d/init.el` (without the last `defadvice!` part, which is apparently required for emacs doom)
https://gist.github.com/hlissner/175c21000114d1dba7a47678191a888a

smile13241324:
I am sorry guys, didn't think about emacs 28 and above. Anyway a fix has been merged just now, please give it a try and report back.

PR #14638, opened 2021-04-14
[c-c++] Fix "Wrong number of arguments" on Emacs 28

eeshugerman:
Closes https://github.com/syl20bnr/spacemacs/issues/14631

Since Spacemacs doesn't really have versions I figure `nil` is the best we can do?


smile13241324:
F***, thanks for fixing this I have totally overlooked this.

PR #14639, opened 2021-04-14
Fix scheme layer support for scheme implementations

dalanicolai:
The configuration for various scheme implementations has been split out of the
[geiser package](https://gitlab.com/emacs-geiser), breaking the current layer support for the implementations.

This PR adds a simple fix for the scheme implementations support.
Anybody is free to improve the code, but at least the layer will be fixed

### DIFF
--- a/layers/+lang/scheme/config.el
+++ b/layers/+lang/scheme/config.el
@@ -24,3 +24,9 @@
 ;; variables
 
 (spacemacs|define-jump-handlers scheme-mode geiser-edit-symbol-at-point)
+
+(defvar scheme-implementations nil
+  "List of scheme implementations for which support should get enabled.
+The symbols of the list can be any multiple of the by geiser
+  supported scheme implementations: chez, chibi, chicken, gambit,
+  gauche, guile, kawa, mit and/or racket.")
--- a/layers/+lang/scheme/packages.el
+++ b/layers/+lang/scheme/packages.el
@@ -21,16 +21,24 @@
 ;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 
-(setq scheme-packages
-      '(
-        company
-        evil-cleverparens
-        geiser
-        ggtags
-        counsel-gtags
-        helm-gtags
-        parinfer
-        ))
+(defconst scheme-packages
+  '(company
+    evil-cleverparens
+    geiser
+    ggtags
+    counsel-gtags
+    helm-gtags
+    parinfer-rust-mode
+    (geiser-chez    :toggle (memq 'chez    scheme-implementations))
+    (geiser-chibi   :toggle (memq 'chibi   scheme-implementations))
+    (geiser-chicken :toggle (memq 'chicken scheme-implementations))
+    (geiser-gambit  :toggle (memq 'gambit  scheme-implementations))
+    (geiser-gauche  :toggle (memq 'gauche  scheme-implementations))
+    (geiser-guile   :toggle (memq 'guile   scheme-implementations))
+    (geiser-kawa    :toggle (memq 'kawa    scheme-implementations))
+    (geiser-mit     :toggle (memq 'mit     scheme-implementations))
+    (geiser-racket  :toggle (memq 'racket  scheme-implementations))))
+
 
 (defun scheme/post-init-company ()
   ;; Geiser provides completion as long as company mode is loaded.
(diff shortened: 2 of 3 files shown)


Kazark (commented):
I think this is going to be a big improvement in experience honestly. I have struggled with the bias of Geiser in the past. It worked much better for some Schemes than others.

layers/+lang/scheme/config.el
```diff
@@ -24,3 +24,30 @@
 ;; variables
 
 (spacemacs|define-jump-handlers scheme-mode geiser-edit-symbol-at-point)
+
+(defvar scheme-activate-chez-implementation nil
```
lebensterben, line 28: Why not make them a list?

`scheme-implementation`

And use `memq` to test whether a package should be installed.

dalanicolai: Sounds like a good idea. I just did not have time to think about how to implement it (or look up how other layers might have implemented something like this). Will have a look a little later, no problem. Or otherwise you are welcome to edit it.

dalanicolai:
@Kazark Ah I do not use Scheme very frequently, but I don't remember to have had any big problems. However, today I got shocked that the layer, suddenly, seemed completely broken. Fortunately it turned out to be a small fix.

layers/+lang/scheme/config.el
lebensterben:
```elisp
(defconst scheme-packages
   `(a
     b
     ,@(mapcar (lambda (x) (let ((implementations '(foo bar))))
                        (if (memq x implementations)
                            (intern (format "geiser-%s" x))
                          (user-error "Unknown scheme implmentation: %s" x)))
               scheme-implementations)))
```

Try this

lebensterben: - We should use `defconst` instead of `defvar` to make sure this is immutable.
- In backquote struct, `,@` evaluate the form and splice its elements.

lebensterben:
You'd be surprised that many other layers are installing unnecessary packages..

layers/+lang/scheme/config.el
dalanicolai: Ah yes, the splice! Nice code, but unfortunately it does not work here because the `scheme-implementations` variable is not yet defined when this gets unpacked (at least when setting it in `dotspacemacs-configuration-layers`). Any suggestion how to fix this?

dalanicolai: Okay nice! It works, but the `scheme-implementations` have to get set in `user-init` (would be nicer if it could be set in `dotspacemacs-configuration-layers` but I think its fine) . I had not used it before, but it says [here](https://develop.spacemacs.org/doc/DOCUMENTATION.html#setting-configuration-layers-variables) that it is "a perfectly valid way to configure a layer").

lebensterben: Ok, then forget about this.
Just put each of the scheme implementation in the list WITHOUT any toggle.

Then, wrap the corresponding `init` function in a test.
Without `init`, the package is defined but never downloaded.

lebensterben:
```elisp
(if (memq scheme-implementation scheme-implementations)
    (eval `(defun ,(intern (format "scheme/init-geiser-%s" scheme-implementation))
               (use-package ,(intern (format "geiser-%s" scheme-implementation))
                 :defer t)))
  (user-error "Unknown scheme implementation: %s" scheme-implementation))
```

Try this.

dalanicolai: I have found a better way. We can just set use `(memq 'implementation scheme-implementations)` for the `:toggle`. (Have updated the PR)

layers/+lang/scheme/packages.el
```diff
@@ -106,3 +114,39 @@
 
 (defun scheme/post-init-parinfer ()
   (add-hook 'scheme-mode-hook 'parinfer-mode))
```
lebensterben, line 116: Can you update this line and L31?
`parinfer-mode` has been superseded by `parinfer-rust-mode`.

Kazark (approved):
That certainly makes more sense than the first implementation.

Is it safe to have multiple of these turned on in parallel (beyond, say, Racket and something else)?

lebensterben (changes requested):

layers/+lang/scheme/packages.el
lebensterben: Well I also found three other layers are using the deprecated `parinfer-mode`.

dalanicolai: Sure... done!

dalanicolai: But it is a post-init function. It does not install parinfer-rust-mode (and also parinfer-rust-mode does not already get installed by another layer here)

smile13241324 (approved):
Looks good. I will double check the parinfer references in other layers, I wonder whether they are actually needed.

PR #14643, opened 2021-04-15
Fix call to deleted method in latex layer

simwir:
In commit f136f46 spacemacs//init-jump-handlers-MODE was removed,
which is used by the latex layer. Change to call to add-to-list instead.


smile13241324 (approved):
Ups overlooked this one, thanks for fixing.

PR #14644, opened 2021-04-15
[pdf] Change pdf-tools source to maintained fork

Phundrak:
The original repo for pdf-tools at https://github.com/politza/pdf-tools
is no longer maintained, and its maintainer seems to be unreachable,
see the issue https://github.com/politza/pdf-tools/issues/659.

Github user vedang forked the project in order to become its new
maintainer at https://github.com/vedang/pdf-tools.

This commit changes the source of the pdf-tools package and makes it
point to the new repository. This is meant to be a workaround until
either the original maintainer of pdf-tools comes back or the repo used
by the MELPA is updated.

The recipe is based on MELPA’s recipe for pdf-tools:
https://github.com/melpa/melpa/blob/master/recipes/pdf-tools

By the way, should the `setq` be replaced by a `defconst`? I noticed
some layers use the latter rather than the former, and I think it would
make more sense to switch from a `setq` to a `defconst`, but I guess 
this would make more sense for another PR.

### DIFF
--- a/layers/+readers/pdf/packages.el
+++ b/layers/+readers/pdf/packages.el
@@ -22,8 +22,16 @@
 ;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 
-(setq pdf-packages '(pdf-tools
-                     pdf-view-restore))
+(setq pdf-packages
+      '((pdf-tools :location (recipe
+                              :fetcher github
+                              :repo "vedang/pdf-tools"
+                              :files ("lisp/*.el"
+                                      "README"
+                                      ("build" "Makefile")
+                                      ("build" "server")
+                                      (:exclude "lisp/tablist.el" "lisp/tablist-filter.el"))))
+        pdf-view-restore))
 
 (defun pdf/init-pdf-tools ()
   (use-package pdf-tools


Phundrak:
I noticed there is an error in the [autobuild file](https://github.com/vedang/pdf-tools/blob/master/server/autobuild), Emacs reports:
```
/home/phundrak/.emacs.d/elpa/28.0/develop/pdf-tools-20210411.1603/build/server/autobuild -i /home/phundrak/.emacs.d/elpa/28.0/develop/pdf-tools-20210411.1603/
/usr/bin/sh: /home/phundrak/.emacs.d/elpa/28.0/develop/pdf-tools-20210411.1603/build/server/autobuild: /bin/sh^M: bad interpreter: No such file or directory
```
While running the script from a terminal reports:
```
Failed to execute process './autobuild'. Reason:
The file uses windows line endings (\r\n). Run dos2unix or similar to fix it.
```
This PR should probably not be merged until this issue is solved in pdf-tools. I will try to submit a PR there asap.

smile13241324:
Good idea, I will wait till the issue is solved upstream before merging.
Basically it is prefered to have "defconst" instead of "setq" for the layers package list. I normally change it when I do something on the layer. Feel free to change it for pdf layer in your PR if you have the time.

vedang:
I don't think the line-ending problem is a `pdf-tools` problem. I'd appreciate it if users of `spacemacs` could investigate this on their end. I've confirmed that no file in `pdf-tools` has CRLF line endings.

smile13241324:
We need someone to test this PR to make sure that it behaves correctly in
  - linux
  - windows
  - macos
prior to have this merged.

Please help by posting the results for your respective operation system.

lebensterben:
https://github.com/vedang/pdf-tools/blob/566c75f3a5af2dc4bfc8333fbafa2a48f02f9eb4/Makefile#L14

- It needs to `make Makefile`. So `gnu make` or some compatible ones is required.
- There's an undeclared dependency, `sed`.

@Phundrak 
I don't think this would even install properly on Windows unless it follows those
https://github.com/vedang/pdf-tools#compilation-and-installation-on-windows

Phundrak:
Hello, sorry for the late reply

> @Phundrak
> I don't think this would even install properly on Windows unless it follows those
> [vedang/pdf-tools#compilation-and-installation-on-windows](https://github.com/vedang/pdf-tools#compilation-and-installation-on-windows)

Then I guess these instructions should be included in the layer’s README

> [vedang/pdf-tools@`566c75f`/Makefile#L14](https://github.com/vedang/pdf-tools/blob/566c75f3a5af2dc4bfc8333fbafa2a48f02f9eb4/Makefile#L14)
> 
> * It needs to `make Makefile`. So `gnu make` or some compatible ones is required.
> * also it requires `sed`.

I think this would be solved by following the instructions above no? I’m very unfamiliar with Windows development and using Emacs under Windows, so I’m not sure how to test this.

> We need someone to test this PR to make sure that it behaves correctly in
> * linux
> * windows
> * macos
> 
> prior to have this merged.
> 
> Please help by posting the results for your respective operation system.

I just tested under macOS with macports and the latest stable release of Emacs on macports (27.2) with macOS 10.15.7, it seems the script for installing all the required dependencies forgot to install `autoconf`, otherwise it works like a charm. Likewise, I’ll test soon on Linux again, I’ll let you know how it goes, and I’ll try to get also macOS BigSur in a VM to test it too, just in case.

Phundrak:
[Melpa changed the upstream repo](https://github.com/melpa/melpa/commit/a6321fb947f6461ac9b673fe47fc7c27ffae09ee) from politza’s repo to Vedang’s, making this pull request obsolete.

lebensterben:
thanks for your effort anyway.

Phundrak:
And a big thanks to @vedang for becoming the new maintainer of pdf-tools!

PR #14652, opened 2021-04-17
[python] Fix directory specific python-backend

syl20bnr:
This is a regression we have in multiple languages. We need to add the backend setup function to `<major-mode>-local-vars-hook` so that people can override a language backend in a given project. See the Emacs documentation on this wonderfully named variable [hack-local-variables](https://www.gnu.org/software/emacs/manual/html_node/elisp/File-Local-Variables.html).

For instance with the following `.dir-locals.el` at the root of a git repository I can choose to use `anaconda` instead of `lsp` (useful for legacy python projects):

```.emacs
;;; Directory Local Variables
;;; For more information see (info "(emacs) Directory Variables")

((python-mode . ((python-test-runner . pytest)
                 (python-backend . anaconda))))
```

This feature is explained in the README.org files of all the languages supporting multiple backends. Example in the Python README: https://github.com/syl20bnr/spacemacs/tree/develop/layers/+lang/python#choosing-a-backend

This PR only fixes it for Python.

cc @lebensterben

### DIFF
--- a/layers/+lang/python/config.el
+++ b/layers/+lang/python/config.el
@@ -30,10 +30,12 @@
   "The backend to use for IDE features.
 Possible values are `anaconda'and `lsp'.
 If `nil' then `anaconda' is the default backend unless `lsp' layer is used.")
+(put 'python-backend 'safe-local-variable #'symbolp)
 
 (defvar python-lsp-server 'pyls
   "Language server to use for lsp backend. Possible values are `pyls', `pyright'
 and `mspyls'")
+(put 'python-lsp-server 'safe-local-variable #'symbolp)
 
 (defvar python-lsp-git-root nil
   "If non-nil, use a development version of the language server in this folder")
@@ -54,6 +56,7 @@ If nil then `yapf' is the default formatter unless `lsp' layer is used.")
 
 (defvar python-test-runner 'nose
   "Test runner to use. Possible values are `nose' or `pytest'.")
+(put 'python-test-runner 'safe-local-variable #'symbolp)
 
 (defvar python-save-before-test t
   "If non nil, current buffer will be save before call a test function")
--- a/layers/+lang/python/packages.el
+++ b/layers/+lang/python/packages.el
@@ -364,9 +364,8 @@
       (spacemacs/register-repl 'python
                                'spacemacs/python-start-or-switch-repl "python")
       (spacemacs//bind-python-repl-keys)
-      (spacemacs/add-to-hook 'python-mode-hook
-                             '(spacemacs//python-setup-backend
-                               spacemacs//python-default))
+      (add-hook 'python-mode-local-vars-hook 'spacemacs//python-setup-backend)
+      (add-hook 'python-mode-hook 'spacemacs//python-default)
       ;; call `spacemacs//python-setup-shell' once, don't put it in a hook
       ;; (see issue #5988)
       (spacemacs//python-setup-shell))


syl20bnr:
Oops I'll need to put the proper author info on the commit.

lebensterben:
@syl20bnr I'll take a look.

lebensterben:
Do you suggest that for file-local variables, e.g. choice of backend, formatter, etc.,
- We need to label it as safe with `(put '<VARIABLE> 'safe-local-variable #'symbolp)`
- If some functions set up the environment according to those variables, we need to add these functions to the local variables hook?

duianto (approved):
This seems to be working as expected.

With the `python` layer added to `dotspacemacs-configuration-layers`,
without any specified variables.

And two directories:
- project_1
containing:
dir: .git
file: 1.py
file: .dir-locals.el
```
;;; Directory Local Variables
;;; For more information see (info "(emacs) Directory Variables")

((python-mode . ((python-test-runner . pytest)
                 (python-backend . anaconda))))
```
- project_2
containing:
file: 2.py

### Before this PR
1.py opens with LSP enabled.
But it asks if the local variables should be applied.
Answering `y` does sets the `python-backend` variable to `anaconda`.
But LSP is still enabled.

2.py opens with LSP enabled.

### After applying this PR
1.py opens with `anaconda` enabled

2.py opens with LSP enabled.

<details><summary>
System Info, click to expand
</summary>
<pre>
#### System Info :computer:
- OS: windows-nt
- Emacs: 27.2
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 6c6062bfe)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(auto-completion autohotkey c-c++ emacs-lisp
                 (git :variables)
                 helm html lsp markdown multiple-cursors
                      '("c:/Users/username/org/"))
                 python shell spell-checking syntax-checking treemacs version-control)
```
</pre>
</details>

lebensterben:
I think we can utilize our new `spacemacs|defc` interface to automate some of the works.

Specifically, some layer variables like `*-backend` and `*-formatter` are checked in `setup-*` functions. Instead of directly calling those `setup-*` function in `MAJOR-mode-init`, it's always better to add those `setup-*` function to the `MAJOR-mode-local-vars-hook`.

Some common patterns are
- Add `MAJOR-setup-backend` to `MAJOR-mode-local-vars-hook` in `MAJOR-mode-init`
- Add `MAJOR-setup-company` to `MAJOR-mode-local-vars-hook` in `company-post-init`
- Add `MAJOR-setup-dap` to `MAJOR-mode-local-vars-hook` in `dap-pre-init`

So for these kind of buffer-local *layer* variables, we can abstract out an interface like

```elisp
(spacemacs|defc-local SYMBOL STANDARD DOC TYPE &optional HOOKS)
```

- All arguments except `HOOKS` are the same as in `spacemacs|defc`
- We don't need `GROUP-OVERRIDE` because layer variable should always belong to the layer
- `HOOKS` is one or multiple lists, (KEYWORD PACKAGE . FUNCS)
  - `KEYWORD` is one of the keyword defined in `spacemacs|use-package-add-hook`,
    i.e one of `:pre-init`, `:post-init`, `:pre-config`, `:post-config`
  - PACKAGE is the package name
  - `FUNC` is a *named* function, or a list of them. (It should require named function to force us to put it in `funcs.el` and document it)

- It always mark the variable as `safe-local-variable`.
- If `HOOKS` is provided, each of the `SETUPFUNC` is added to `MAJOR-mode-local-vars-hook` via `spacemacs|user-package-add-hook` interface.

Taking python layer as an example:

```elisp
(spacemacs|defc-local python-backend (if (configuration-layer/layer-used-p 'lsp) 'lsp 'anaconda)
  "The backend to use for IDE features.
Possible values are `anaconda'and `lsp'.
If `nil' then `anaconda' is the default backend unless `lsp' layer is used."
  '(choice (const lsp) (const anaconda))
  (:post-config python . spacemacs//python-setup-backend))
```

It's equivalent to
```elisp
(spacemacs|defc python-backend (if (configuration-layer/layer-used-p 'lsp) 'lsp 'anaconda)
  "The backend to use for IDE features.
Possible values are `anaconda'and `lsp'.
If `nil' then `anaconda' is the default backend unless `lsp' layer is used."
  '(choice (const lsp) (const anaconda))
  'spacemacs-python-layer)

(put 'python-backend 'safe-local-variable #'symbolp)

(spacemacs|use-package-add-hook python
  :post-config (add-hook 'python-mode-local-vars-hook #'spacemacs//python-setup-backend))
```

smile13241324:
I like the idea with `spacemacs|defc`, I have used the chance and switched the layer variables to defcustoms yesterday for groovy and go layer anyway, having the standard interfaces already declared during the var definition would clean up the config considerably.

Nevertheless what I am still missing is a proper safe-local-variable function to set. In my last two commits I have added a new parameter to `spacemacs|defc` to be able to set such a function. However what I actually wanted to do is to set the generic validation function for the custom var as safe-local-variable.

Not sure if we need two versions of the `defc` macro or just one with a lot of parameters though i.e. hooks could be an optional parameter  to `defc` also and we would only have to maintain one function.

@JAremko @duianto @sdwolfz whats your opinion about this? I think we should try to get a clear opinion about what `defc` need to do before we start switching the layers on a bigger scale.

smile13241324 (approved):
Looks ok.

duianto:
I'll trust your judgements. I don't know enough about this to have an informed opinion.

syl20bnr:
I also like the `defc` macro idea, it feels natural to have one. I agree with @smile13241324 that it should be one macro with all the keywords. If the implementation seems awkward and artificial then we can start splitting it. But better to start with only one.

Also in this PR I used the predicate `symbolp` to avoid duplication of supported values, if we have a macro we can be even more accurate with `safe-local-variable` as it is done in the `typescript` layer.

Do we have a GitHub discussion about this already ? If not feel free to start it and mention me on it.

lebensterben:
The solution is much simpler than I initially thought.
For any variables defined via `spacemacs|defc`, we've already have `spacemacs/customization-valid-p` to check its validity.

```elisp
(defun spacemacs/customization-valid-p (var-symbol)
  "returns true if symbol refers spacemacs custom variable with valid value.
Emits message with the result when called interactively."
  (interactive "v")
  (let* ((defc? (get var-symbol 'spacemacs-customization--variable))
         (val (symbol-value var-symbol))
         (type (custom-variable-type var-symbol))
         (valid? (and defc? (validate-value val type t))))
    (when (called-interactively-p 'interactive)
      (if valid?
          (message "symbol: \"%s\" value: \"%s\" type: \"%s\" is valid."
                   var-symbol val type)
        (if defc?
            (condition-case err (validate-value val type) (error (message err)))
          (message "%s is not Spacemacs customization variable" var-symbol))))
    valid?))
```

What we really needed is:
```elisp
(put 'foo 'safe-local-variable
     (lambda (var-symbol)
       (let* ((val (symbol-value var-symbol))
              (type (custom-variable-type var-symbol)))
         (validate-value val type t))))
```

lebensterben:
implemented in https://github.com/syl20bnr/spacemacs/pull/14679

Issue #14653, opened 2021-04-17
tracking issue: Fix directory/file specific layer configuration


lebensterben:
As reported in https://github.com/syl20bnr/spacemacs/pull/14652, there's a regression since https://github.com/syl20bnr/spacemacs/pull/14509.

> This is a regression we have in multiple languages. We need to add the backend setup function to <major-mode>-local-vars-hook so that people can override a language backend in a given project. See the Emacs documentation on this wonderfully named variable hack-local-variables.
>
> For instance with the following .dir-locals.el at the root of a git repository I can choose to use anaconda instead of lsp (useful for legacy python projects):
>
> ;;; Directory Local Variables
> ;;; For more information see (info "(emacs) Directory Variables")
>
> ((python-mode . ((python-test-runner . pytest)
>                  (python-backend . anaconda))))
>
>This feature is explained in the README.org files of all the languages supporting multiple backends. Example in the Python README: https://github.com/syl20bnr/spacemacs/tree/develop/layers/+lang/python#choosing-a-backend

 <summary>Review progress</summary>

  - [ ] haskell https://github.com/syl20bnr/spacemacs/pull/14668
  - [ ] c-c++ https://github.com/syl20bnr/spacemacs/pull/14654
  - [ ] clojure https://github.com/syl20bnr/spacemacs/pull/14655
  - [ ] crystal https://github.com/syl20bnr/spacemacs/pull/14656
  - [ ] csharp https://github.com/syl20bnr/spacemacs/pull/14657
  - [x] dart https://github.com/syl20bnr/spacemacs/pull/14658
  - [x] elixir https://github.com/syl20bnr/spacemacs/pull/14659
  - [ ] elm https://github.com/syl20bnr/spacemacs/pull/14661
  - [x] erlang https://github.com/syl20bnr/spacemacs/pull/14662
  - [ ] ess https://github.com/syl20bnr/spacemacs/pull/14663
  - [ ] fsharp https://github.com/syl20bnr/spacemacs/pull/14664
  - [x] go https://github.com/syl20bnr/spacemacs/pull/14665
  - [ ] python https://github.com/syl20bnr/spacemacs/pull/14652
  - [x] groovy https://github.com/syl20bnr/spacemacs/pull/14666
  - [ ] javascript https://github.com/syl20bnr/spacemacs/pull/14670
  - [ ] typescript
  - [ ] java https://github.com/syl20bnr/spacemacs/pull/14669
[...]

PR #14654, opened 2021-04-17
c-c++: Fix file specific layer configuration

lebensterben:
- Labelled `c-c++-backend`, `c-c++-enable-organize-includes-on-save`
  as safe local variable.
- Added local variable hooks of c/c++ modes:
  - `spacemacs//c-c++-setup-clang-format`.
  - `spacemacs//c-c++-organize-includes-on-save-hook`
- Improved `spacemacs//c-c++-setup-dap`
- Improved `c-c++-organize-includes` related functions.
- Declared `c-c++-dap-adapters` as obsolete variable alias for
  `c-c++-dap-adaptors` because of misspelling.

All other setup functions are already added to local variable hook.

-------

Cheers!

### DIFF
--- a/layers/+lang/c-c++/config.el
+++ b/layers/+lang/c-c++/config.el
@@ -29,6 +29,7 @@
 (defvar c-c++-backend (when (configuration-layer/layer-used-p 'lsp) 'lsp-clangd)
   "The backend to use for IDE features.
 Possible values are `lsp-ccls', `lsp-clangd', `rtags' and `ycmd'.")
+(put 'c-c++-backend 'safe-local-variable #'symbolp)
 
 
 ;; lsp
@@ -47,14 +48,14 @@ By default `font-lock' is used to highlight the text, set the variable to
 
 
 ;; dap
-
-(defvar c-c++-dap-adapters '(dap-cpptools)
-  "Debug adapters to use for IDE debug features.
+(define-obsolete-variable-alias 'c-c++-dap-adapters 'c-c++-dap-adaptors "April 17, 2021")
+(defvar c-c++-dap-adaptors '(dap-cpptools)
+  "Debug adaptors to use for IDE debug features.
 
 By default only `dap-cpptools' is used.
 
 Add `dap-cpptools' for the official Microsoft C/C++ Extension for VSCode.
-Add `dap-lldb' for the official LLDB project adapter.
+Add `dap-lldb' for the official LLDB project adaptor.
 Add `dap-gdb-lldb' for the WebFreak Native Debug extension.")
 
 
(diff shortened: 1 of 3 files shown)


syl20bnr:
I believe `adapter` is the right word. I looked for `adaptor` keyword in the DAP related repos and found nothing, also `DAP` means `Debug Adapter Protocol` so high chance `adapter` is the way to go.

layers/+lang/c-c++/config.el
```diff
@@ -47,14 +48,14 @@ By default `font-lock' is used to highlight the text, set the variable to
 
 
 ;; dap
-
-(defvar c-c++-dap-adapters '(dap-cpptools)
-  "Debug adapters to use for IDE debug features.
+(define-obsolete-variable-alias 'c-c++-dap-adapters 'c-c++-dap-adaptors "April 17, 2021")
```
smile13241324, line 51: I think adapters is the right term, see syl20bnr comment.

layers/+lang/c-c++/funcs.el
```diff
@@ -393,10 +396,12 @@
 (defun spacemacs//c-c++-organize-includes-on-save ()
   "Organize the includes on save when `c-c++-enable-organize-includes-on-save'
```
smile13241324, line 397: This comment is now wrong, given that this function now works without conditions.

layers/+lang/c-c++/packages.el
```diff
@@ -90,7 +89,7 @@
 
 (defun c-c++/init-clang-format ()
   (use-package clang-format
-    :init (spacemacs//c-c++-setup-clang-format)))
+    :hook ((c-mode-local-vars c++-mode-local-vars) . spacemacs//c-c++-setup-format)))
```
smile13241324, line 92: This is probably problematic, in this layer the keybindings for formating has been added to `spacemacs//c-c++-setup-format` in this case it would set the bindings each time we activate the mode. Please move the keys into the init block of `c-c++/init-clang-format`.

Also be aware that each call to this function adds more hooks to `c-c++-mode-hooks`.

layers/+lang/c-c++/funcs.el
```diff
@@ -43,7 +43,10 @@
   ;; currently DAP is only available using LSP
   (pcase c-c++-backend
     ('lsp-clangd (spacemacs//c-c++-setup-lsp-dap))
-    ('lsp-ccls (spacemacs//c-c++-setup-lsp-dap))))
+    ('lsp-ccls (spacemacs//c-c++-setup-lsp-dap))
+    (_ (user-error "Unknown `c-c++-backend': %s" c-c++-backend)))
+  (setq-local spacemacs--dap-supported-modes
+              '(c-mode c++-mode)))
```
smile13241324, line 49: This should not be necessary, the list of supported dap modes is globally set among all layers, a local version of this variable could cause havoc, i.e. go into a python file, swtich to c++-mode and switch back your buffer will not longer have dap support for python available.

Instead this list should be extended in the `pre-init-dap` section.

smile13241324 (changes requested):
I have reviewed your changes and found some points we should double check. Most of them are about setting key bindings multiple times or extending hooks multiple times.

Both we should avoid.

PR #14655, opened 2021-04-17
clojure: Fix file specific variable

lebensterben:
- Labelled `clojure-backend` as safe local variable.
- Added local variable hooks of clojure mode:
  - `spacemacs//clojure-setup-backend`
- Added new function `spacemacs//clojure-bindings-hook` to setup
  prefixes and bindings. Common prefixes and bindings are set only
  once. `cider` specific prefixes and bindings are set after
  file local variables are loaded to allow local override of
  `clojure-backend`.
- `spacemacs|forall-clojure-modes` is renamed to
  `spacemacs||forall-clojure-modes` according to naming conventions.

-------

Cheers!

### DIFF
--- a/layers/+lang/clojure/funcs.el
+++ b/layers/+lang/clojure/funcs.el
@@ -25,6 +25,24 @@
   "Conditionally setup clojure backend."
   (when (eq clojure-backend 'lsp) (lsp)))
 
+(defun spacemacs//clojure-bindings-hook
+    (common-prefix cider-prefix common-binding cider-binding)
+  "Conditinally setup clojure-modes bindings."
+  (spacemacs||forall-clojure-modes m
+    (mapc (lambda (x) (spacemacs/declare-prefix-for-mode m (car x) (cdr x)))
+          common-prefix)
+    (eval `(spacemacs/set-leader-keys-for-major-mode ,m
+             ,@(common-binding)))
+    (add-hook (intern (format "%s-local-vars-hook" m))
+              (lambda ()
+                (when (eq clojure-backend 'cider)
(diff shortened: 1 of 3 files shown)


lebensterben:
Found some issue. Hold on.

layers/+lang/clojure/funcs.el
```diff
@@ -224,7 +242,7 @@ in your Spacemacs configuration:
              clojure-enable-clj-refactor t)
     ) ")))
 
-(defmacro spacemacs|forall-clojure-modes (m &rest body)
+(defmacro spacemacs||forall-clojure-modes (m &rest body)
```
syl20bnr, line 245: Maybe we can keep it public as users might find this macro handy for their custom config.

smile13241324: Yep, btw we should add a line to the conventions defining that a macro with `||` is private. Right now we only have this for functions and variables.

layers/+lang/clojure/funcs.el
```diff
@@ -25,6 +25,24 @@
   "Conditionally setup clojure backend."
   (when (eq clojure-backend 'lsp) (lsp)))
 
+(defun spacemacs//clojure-bindings-hook
+    (common-prefix cider-prefix common-binding cider-binding)
+  "Conditinally setup clojure-modes bindings."
```
smile13241324, line 30: Small typo
```suggestion
  "Conditionally setup clojure-modes bindings."
```

smile13241324 (changes requested):
Found a small typo, the rest looks good. Can you check @syl20bnr comment? Maybe we should keep this macro non private for private layers.

PR #14656, opened 2021-04-17
crystal: Fix file local variable

lebensterben:
- Labelled `crystal-backend` and `crystal-enable-auto-format`
  as safe local variable.
- Added local variable hooks of crystal mode:
  - `spacemacs//crystal-setup-backend`
  - `spacemacs//crystal-setup-company`
  - `spacemacs//crystal-auto-format-setup`
- Imprvoed `spacemacs//crystal-setup-backend` to setup bindings
  correctly according to backend.

-------

Cheers!

### DIFF
--- a/layers/+lang/crystal/funcs.el
+++ b/layers/+lang/crystal/funcs.el
@@ -47,5 +47,9 @@
                         :variables company-tooltip-align-annotations t))))
 
 (defun spacemacs//crystal-setup-backend ()
-  "Conditionally setup crystal backend."
-  (when (eq crystal-backend 'lsp) (lsp)))
+  "Conditionally setup crystal backend and bindings."
+  (if (eq crystal-backend 'lsp)
+      (lsp)
+    (spacemacs/declare-prefix-for-mode 'crystal-mode "mg" "goto")
+    (spacemacs/declare-prefix-for-mode 'crystal-mode "mt" "test")
+    (spacemacs/declare-prefix-for-mode 'crystal-mode "ma" "action")))
--- a/layers/+lang/crystal/packages.el
+++ b/layers/+lang/crystal/packages.el
@@ -51,15 +51,12 @@
 (defun crystal/init-crystal-mode ()
   (use-package crystal-mode
     :defer t
+    :hook
+    (crystal-mode-local-vars . spacemacs//crystal-auto-format-setup)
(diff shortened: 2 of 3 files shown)


layers/+lang/crystal/funcs.el
```diff
@@ -47,5 +47,9 @@
...
 
 (defun spacemacs//crystal-setup-backend ()
-  "Conditionally setup crystal backend."
-  (when (eq crystal-backend 'lsp) (lsp)))
+  "Conditionally setup crystal backend and bindings."
+  (if (eq crystal-backend 'lsp)
+      (lsp)
+    (spacemacs/declare-prefix-for-mode 'crystal-mode "mg" "goto")
+    (spacemacs/declare-prefix-for-mode 'crystal-mode "mt" "test")
+    (spacemacs/declare-prefix-for-mode 'crystal-mode "ma" "action")))
```
syl20bnr, line 55: I know you just moved the prefix definition so this is not really related to this PR but I wonder why these prefixes are removed with the backend is `lsp`. 
Feels weird to me, `lsp` does not handle tests and it surely does not declare tests key bindings.

smile13241324: This looks wrong, except of `test` these should not be bound when LSP is used. Also the prefixes should be set one time not everytime the mode is activated, this could otherwise degrade which-key's performance.

layers/+lang/crystal/packages.el
```diff
@@ -34,7 +34,7 @@
     play-crystal))
 
 (defun crystal/post-init-company ()
-  (spacemacs//crystal-setup-company))
+  (add-hook 'crystal-mode-local-vars-hook #'spacemacs//crystal-setup-company))
```
smile13241324, line 37: Here we are possibly setting the company backends multiple times. Is there a side effect with company when we register a backend multiple times and what happens when we set two different backends for the same mode.

With this directory-local-variable support we seem to have to call some functions multiple times, the same may apply to prefixes and key bindings. However if we really have someone wanting to switch between lsp and non-lsp backends we will most probably not have a clean key map anymore i.e. entering a python project with anaconda, then switching to another one with lsp the prefixes from anaconda won't be removed possibly hiding prefixes from `lsp` layer.

I am not sure how to handle this @syl20bnr @lebensterben.

smile13241324 (changes requested):

PR #14661, opened 2021-04-17
elm: Fix local vars

lebensterben:
- Labelled `elm-backend` as safe local variable.
- Added local variable hooks of elm modes:
  - `spacemacs//elm-setup-backend`
  - `spacemacs//elm-setup-company`

-------

Cheers!

### DIFF
--- a/layers/+lang/elm/funcs.el
+++ b/layers/+lang/elm/funcs.el
@@ -40,6 +40,26 @@
     ('lsp (lsp))
     ('company-elm (elm-oracle-setup-completion))))
 
+(defun spacemacs//elm-setup-binding
+    (common-prefix company-elm-prefix common-binding company-elm-binding)
+  "Conditionally setup prefix and bindings according to `elm-backend'."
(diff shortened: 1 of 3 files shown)


layers/+lang/elm/funcs.el
```diff
@@ -40,6 +40,26 @@
...
+  (add-hook 'elm-mode-local-vars-hook
+            (lambda ()
+              (when (eq elm-backend 'company-elm)
+                (cl-loop for x on company-elm-prefix
+                         by 'cddr
+                         do (apply 'spacemacs/declare-prefix-for-mode
+                                   'elm-mode
+                                   (list (car x) (cadr x))))
+                (apply 'spacemacs/set-leader-keys-for-major-mode 'elm-mode
+                       company-elm-binding)))))
```
syl20bnr, line 61: You don't need this function because key bindings are setup in the `:config` part of `use-package` which will happen after the backend is defined and known.
Also there is no reusability of this key bindings.

PR #14664, opened 2021-04-17
fsharp: Fix local vars

lebensterben:
- Labelled `fsharp-backend` as safe local variable.
- Added local variable hooks of fsharp mode:
  - `spacemacs//fsharp-setup-backend`

-------

Cheers!

### DIFF
--- a/layers/+lang/fsharp/funcs.el
+++ b/layers/+lang/fsharp/funcs.el
@@ -35,7 +35,25 @@
   "Conditionally setup fsharp backend."
   (pcase fsharp-backend
     ('lsp (lsp))
-    ('eglot (eglot-ensure))))
+    ('eglot (eglot-ensure)
+            (require 'eglot-fsharp))))
+
+(defun spacemacs//fsharp-setup-binding
+    (common-prefix eglot-prefix common-binding)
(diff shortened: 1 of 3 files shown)


layers/+lang/fsharp/funcs.el
```diff
@@ -35,7 +35,25 @@
...
+  (apply 'spacemacs/set-leader-keys-for-major-mode 'fsharp-mode
+         common-binding)
+  (add-hook 'fsharp-mode-local-vars-hoook
+            (lambda ()
+              (when (eq fsharp-backend 'eglot)
+                (cl-loop for x on eglot-prefix
+                         by 'cddr
+                         do (apply 'spacemacs/declare-prefix-for-mode
+                                   'fsharp-mode
+                                   (list (car x) (cadr x))))))))
```
syl20bnr, line 56: Same as my other comment on the key bindings, I think we don't need this function as we define the key bindings in `:config` and there is no reusability of the key bindings.
I am aware that if someone wants for some reason use `eglot` in one project and `lsp` in another there will be a key bindings conflict. We might need to think about a way to isolate keybindings per backend at some point.

PR #14678, opened 2021-04-18
configure racket-describe-mode keybindings

dalanicolai:
`racket-describe-mode` is a read-only mode that comes with its own keybinding (i.e. `q` for quit).
To make those keybindings available, this PR makes the mode open in evilified state.
Additionally, it contains links, so this PR binds `o` to `link-hint-open-link`
(`ace-link` does not support this mode)


smile13241324 (approved):
:+1:

PR #14679, opened 2021-04-18
core-customization: improved SAFE variable handling

lebensterben:
- SAFE can either be a function or t.
  - When it's t, use a default validation function
    `(lambda (val) (validate-value val TYPE t))`
  - When it's a function, use the supplied function.

-------

Take `groovy-backend` as example, notice the `SAFE` argument is `t`.

```elisp
(spacemacs|defc groovy-backend (if (configuration-layer/layer-used-p 'lsp) 'lsp 'company-groovy)
  "The backend to use for IDE features.
Possible values are `lsp' and `company-groovy'.
If not set then 'company-groovy` is the default backend unless `lsp' layer is used"
  '(choice (const lsp) (const company-groovy)) nil t)
```

Check the safety validator function
```elisp
ELISP> (get 'groovy-backend 'safe-local-variable)
#f(compiled-function
   (&rest args2)
   #<bytecode 0x15901b8b29c9>)

ELISP> (aref (get 'groovy-backend 'safe-local-variable) 2)
[spacemacs-customization//get-variable-validator
 (groovy-backend)
 apply append]
```

which is returned from
`(apply-patially 'spacemacs-customization//get-variable-validator groovy-backend)`

This would return a validation function for the given customization variable.
Different from `validate-value`, which returns the value as-is if the value is valid
and nil/error otherwise, it returns t when it's valid and error otherwise.

The error is suppressed when it's called from `safe-local-variable-p`.
Try a few values:

```elisp
ELISP> (safe-local-variable-p 'groovy-backend 'lsp)
lsp
ELISP> (safe-local-variable-p 'groovy-backend 'company-groovy)
company-groovy
ELISP> (safe-local-variable-p 'groovy-backend 'company-xx)
nil
ELISP> (safe-local-variable-p 'groovy-backend "lsp")
nil
```

### DIFF
--- a/core/core-customization.el
+++ b/core/core-customization.el
@@ -63,8 +63,9 @@ TYPE           should be a widget type for editing the symbol's value.
                base types and useful composite types.
 GROUP-OVERRIDE should be provided if you don't want Spacemacs to infer the
                configuration group from the currently configured layer name.
-SAFE           should be filled with a function to be set to
-               safe-local-variable property.
+SAFE           should either be a function or t to be set to
+               safe-local-variable property. When it's t, use TYPE to determine
+               the safety.
 
 NOTE: Use interactive function `spacemacs/customization-valid-p' to test if a
       variable has a proper type. In interactive mode it will also `message'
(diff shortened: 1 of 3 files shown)


core/core-customization.el
```diff
@@ -91,8 +92,9 @@ NOTE: Variables defined with a group listed in
       ,(format "%s\n\nTYPE: %s\n" doc type)
       :type ,type
       :group group)
-     (when (not (null ',safe))
-       (put ',symbol 'safe-local-variable ,safe))
+     (pcase ,safe
+       ('t (put ',symbol 'safe-local-variable (lambda (val) (validate-value val ,type t))))
```
lebensterben, line 96: Note that here it's matching the symbol `'t`, not to be confused with the catch-all case `t`.

lebensterben:
By the way I found a bug in `validate.el`
You'd expect for a `(defcustom foo nil ".." :type 'boolean)`,
at default value nil, `(validate-variable 'foo)` should
evaluate to t.

But no, it returns its value, nil.

If we fix that, we can use `validate.el`'s provide function to mark safe variables.
```elisp
(defun validate-mark-safe-local (symbol)
  "Mark SYMBOL as a safe local if its custom type is obeyed."
  (put symbol 'safe-local-variable
       (lambda (val)
         (validate-value val (custom-variable-type symbol) 'noerror))))
```

smile13241324 (approved):
Great work, will be merged soon.

PR #14680, opened 2021-04-19
react: fix file auto detection

LuisChDev:
commit 58243c67f cleans up the React layer nicely but introduces
a regression where files are not auto detected correctly.
This commit simply restores the previous code where needed to make it
work again.

### DIFF
--- a/layers/+frameworks/react/funcs.el
+++ b/layers/+frameworks/react/funcs.el
@@ -63,7 +63,8 @@
 (defun spacemacs//javascript-jsx-file-p ()
   "Enable rjsx mode by using magic-mode-alist."
   (when buffer-file-name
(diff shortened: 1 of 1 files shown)


layers/+frameworks/react/funcs.el
```diff
@@ -63,7 +63,8 @@
 (defun spacemacs//javascript-jsx-file-p ()
   "Enable rjsx mode by using magic-mode-alist."
   (when buffer-file-name
-    (and (memq (file-name-extension buffer-file-name) '("js" "jsx"))
```
lebensterben, line 66: replace memq with member should fix this.

lebensterben (changes requested):

LuisChDev:
You were right. It's fixed now.

lebensterben (approved):

PR #14681, opened 2021-04-19
Fix: update eaf layer

dalanicolai:
The layer got broken due to some minor changes in the eaf package by the eaf
developers. This PR fixes the layer to a working state.

### DIFF
--- a/layers/+tools/eaf/packages.el
+++ b/layers/+tools/eaf/packages.el
@@ -131,7 +131,6 @@
               ("-" . "insert_or_zoom_out")
               ("=" . "insert_or_zoom_in")
               ("0" . "insert_or_zoom_reset")
-              ("d" . "insert_or_dark_mode")
               ("m" . "insert_or_save_as_bookmark")
               ("o" . "insert_or_open_browser")
               ;; ("y" . "insert_or_download_youtube_video")
@@ -236,12 +235,11 @@
         (lambda (prompt)
           (if (derived-mode-p 'eaf-mode)
               (pcase eaf--buffer-app-name
-                ((or
-                  (and "browser"
-                       (guard (string= (eaf-call-sync "call_function" eaf--buffer-id "is_focus") "True")))
-                  "image-viewer"
-                  "pdf-viewer")
-                 (kbd eaf-evil-leader-key))
+                ("browser" (if (string= (eaf-call-sync "call_function" eaf--buffer-id "is_focus") "True")
+                               (kbd "SPC")
+                             (kbd eaf-evil-leader-key)))
+                ("pdf-viewer" (kbd eaf-evil-leader-key))
+                ("image-viewer" (kbd eaf-evil-leader-key))
                 (_  (kbd "SPC")))
             (kbd "SPC"))))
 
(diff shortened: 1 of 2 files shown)


dalanicolai:
Or actually I think it got broken by the recent big refactor operation...

layers/+tools/eaf/packages.el
```diff
@@ -236,20 +235,19 @@
         (lambda (prompt)
           (if (derived-mode-p 'eaf-mode)
               (pcase eaf--buffer-app-name
-                ((or
-                  (and "browser"
-                       (guard (string= (eaf-call-sync "call_function" eaf--buffer-id "is_focus") "True")))
```
lebensterben, line 241:
```
(guard (string= (eaf-call-sync "call_function" eaf--buffer-id "is_focus") "True")))
```
to
```
(guard (not (string= (eaf-call-sync "call_function" eaf--buffer-id "is_focus") "True"))))
```

Should fix this.

lebensterben (changes requested):

layers/+tools/eaf/packages.el
dalanicolai: I guess you mean I should just add the `not` in the original code (before the PR). Anyway, I have tried that but that also does not work (ESC does exit insert mode).

smile13241324 (approved):
:+1:

layers/+tools/eaf/packages.el
dalanicolai: Ah sorry... I think it does work, there is another reason why it seems broken. I don't know exactly how eaf reparenting works, but somehow (on Ubuntu) the browser window seems not to get well embedded in Emacs, but instead it opens in its own separate window. But from its position it looks to be embedded in Emacs, however when I select the browser window, that window gets active instead of the Emacs window (so that the keybindings don't seem to work). On Fedora it all seems to work fine (I am switching a little between Fedora and Ubuntu, because `mit-scheme` does not build in Fedora. Anyway.. this makes things a little more confusing.)

dalanicolai: Congrats with your new role in the Spacemacs team!

lebensterben: Key binding issue aside, I've trouble with `SPC m d` for toggling dark-mode.
P.S. EAF never fully worked on my computer so I cannot fully test it.

layers/+tools/eaf/packages.el
```diff
@@ -236,20 +235,19 @@
...
+                ("pdf-viewer" (kbd eaf-evil-leader-key))
+                ("image-viewer" (kbd eaf-evil-leader-key))
                 (_  (kbd "SPC")))
             (kbd "SPC"))))
 
       ;; The following lines create the major-mode leader key emulation map
       ;; in a similar way as how it was done in the evil-integration example
       (setq eaf-evil-leader-for-major-keymap (make-sparse-keymap))
       (define-key eaf-evil-leader-for-major-keymap (kbd "h") 'eaf-open-browser-with-history)
-      (define-key eaf-evil-leader-for-major-keymap (kbd "d") 'eaf-toggle-dark-mode)
```
lebensterben, line 252: This function is defined in `funcs.el` and now it's not used.
Do we need to remove it? But it works 100% of time on my computer....

dalanicolai: Are you sure that function works, because here it did not. I did not remove it, but replace it with the working version of that function (see below)

lebensterben: I can only view PDF file in EAF on my computer. And the only working function is to toggle dark mode. After this commit that stopped working.

lebensterben: No, the it should not be this line.

lebensterben: What I meant is https://github.com/syl20bnr/spacemacs/blob/df15f8cc159b9fe4380144355af077935e7770b7/layers/%2Btools/eaf/funcs.el#L34

That is not used after this PR. And that function works for me.

dalanicolai: Haha okay, I forgot that I had added that function myself. But I am using mainly the eaf-browser, and there this function stopped working while that new function does work. So I guess maybe we require both functions. I will look into it a little more then.

PR #14685, opened 2021-04-20
Fix eslint not available for typescript-tsx-mode

ralph-cheng:
The bug is that when using the following typescript layer configuration, flycheck doesn't have any checker available in typescript-tsx-mode.
```emacs-lisp
  (typescript :variables
              typescript-backend 'lsp
              typescript-linter 'eslint
              typescript-lsp-linter nil)
```

The root cause is that the `(pcase ...)` code, in typescript/packages.el, never got executed . Removing the surrounding `(with-eval-after-load 'lsp` fixes it. 

Given  `(pcase ..)` never executed, typescript-mode still have some flycheck checkers available because flycheck has eslint configured for typescript-mode by default, see flycheck.el line [9773](https://github.com/flycheck/flycheck/blob/f8c679fff349850c80541a31de50009c3c15d4c9/flycheck.el#L9773) and line [12254](https://github.com/flycheck/flycheck/blob/f8c679fff349850c80541a31de50009c3c15d4c9/flycheck.el#L12254). So the bug is not reproducible in typescript-mode.

Using lsp as checker still work after removing `(with-eval-after-load 'lsp` in this PR, because lsp for checker is turned on in typescript/funcs.el line [52](https://github.com/syl20bnr/spacemacs/blob/b1f54b337109dd6988b102de59d43e71641f5c34/layers/+lang/typescript/funcs.el#L52).

### DIFF
--- a/layers/+lang/typescript/packages.el
+++ b/layers/+lang/typescript/packages.el
@@ -70,16 +70,15 @@
 
 (defun typescript/set-lsp-linter ()
   (with-eval-after-load 'lsp-ui
-    (with-eval-after-load 'lsp
-      (with-eval-after-load 'flycheck
-        (pcase typescript-linter
-          ('tslint (flycheck-add-mode 'typescript-tslint 'typescript-tsx-mode))
-          ;; This sets tslint unconditionally for all lsp clients which is wrong
-          ;; Must be set for respective modes only, see go layer for examples.
-          ('eslint (flycheck-add-mode 'javascript-eslint 'typescript-tsx-mode)
-                   (flycheck-add-mode 'javascript-eslint 'typescript-mode))
-          (_ (message
-              "Invalid typescript-layer configuration, no such linter: %s" typescript-linter)))))))
+    (with-eval-after-load 'flycheck
+      (pcase typescript-linter
+        ('tslint (flycheck-add-mode 'typescript-tslint 'typescript-tsx-mode))
+        ;; This sets tslint unconditionally for all lsp clients which is wrong
+        ;; Must be set for respective modes only, see go layer for examples.
+        ('eslint (flycheck-add-mode 'javascript-eslint 'typescript-tsx-mode)
+                 (flycheck-add-mode 'javascript-eslint 'typescript-mode))
+        (_ (message
+            "Invalid typescript-layer configuration, no such linter: %s" typescript-linter))))))
 
 (defun typescript/post-init-flycheck ()
   (spacemacs/enable-flycheck 'typescript-mode)


lebensterben:
That's strange.
There's still the enclosing ￼`with-eval-after-load 'lsp-ui`, which should load after `lsp`.

ralph-cheng:
> That's strange.
> There's still the enclosing ￼`with-eval-after-load 'lsp-ui`, which should load after `lsp`.

Actually, don't know why but `lsp` is not loaded at all using the provided layer configuration.

This can be verified by evaluating the following emacs lisp code in a `typescript-tsx-mode`, or `typescript-mode`, buffer
```emacs-lisp
(with-eval-after-load 'lsp-ui
   (message "Hello, spacemacs")))
;; "Hello, spacemacs"

(with-eval-after-load 'lsp-ui
 (with-eval-after-load 'lsp
     (message "Hello, spacemacs"))))
;; nil
```

Back to the code, I removed the enclosing `(with-eval-after-load 'lsp-ui`, and tested manually. Everything works just fine. 

Do you recommend to remove it as well?

lebensterben:
What if you remove the one with lsp-ui, and keep the one with lsp ?

ralph-cheng:
> What if you remove the one with lsp-ui, and keep the one with lsp ?

doesn't work. 

Like the example has shown, `lsp` is not loaded. So keeping it makes the `(pcase` not executed.

lebensterben:
this is wild....

lsp-ui requires lsp feature...

lebensterben (approved):

lebensterben:
Merged.
But we need to figure out how could this happen.
How come lsp is not loaded while lsp-ui is?

ralph-cheng:
> Merged.
> But we need to figure out how could this happen.
> How come lsp is not loaded while lsp-ui is?

Thanks.

I think it has something to do with the name `lsp`. 

When I navigate to definition for `'lsp` in the `with-eval-after-load`, I end at a `defun` in the file [`lsp-mode.el`](https://github.com/emacs-lsp/lsp-mode/blob/54cb9a81d97c9f19ebea34510844ac7b69e755bc/lsp-mode.el#L7947), but at the same time there is a file called [`lsp.el`](https://github.com/emacs-lsp/lsp-mode/blob/54cb9a81d97c9f19ebea34510844ac7b69e755bc/lsp.el).

When I try navigate to definition for `'lsp-ui` and `'flycheck`, it took me to the files under the same names. So my guess is `(with-eval-after-load 'lsp` is waiting for the file lsp.el to be loaded. But it doesn't load.

I add some `(message ...)` in the related files

1. lsp.el line 1 - `(message "load lsp.el file")`
2. lsp-mode.el, line 1 - `(message "load lsp-mode.el file")`
3. lsp-mode.el, 1st line in (defun lsp ...) - `(message "executing lsp function")`
4. lsp-ui.el line 1 - `(message "load lsp-ui file")`

and this is what I get in the message buffer:
```
load lsp-mode.el file
executing lsp function
loading lsp-ui.el file
LSP :: Connected to [ts-ls:5283/starting].
```

The message for lsp.el is not printed.

Then I tried another approach to fix, by changing `(with-eval-after-load 'lsp` into `(with-eval-after-load 'lsp-mode `. It worked.

Anyways, it might be a good idea to remove both lsp relared `with-eval-after-load` because the code it enclosing doesn't really require these two files to be loaded.

lebensterben:
this sounds like an upstream bug

lsp.el provides lsp feature and requires lsp-mode feature from lsp-mode.el

so (with-eval-after-load 'lsp) is perfectly valid.

PR #14687, opened 2021-04-20
Add support for org-roam-protocol

KjartanOli:
Org-Roam provides org-roam-protocol, a feature that allows the graphs generated by `org-roam-graph` and org-roam-server to contain cilckable links, which open the corresponding  org-roam file.

### DIFF
--- a/layers/+emacs/org/README.org
+++ b/layers/+emacs/org/README.org
@@ -27,6 +27,7 @@
   - [[#org-brain-support][Org-brain support]]
   - [[#org-roam-support][Org-roam support]]
     - [[#org-roam-server-support][Org-roam-server support]]
+    - [[#org-roam-protocol-support][Org-roam-protocol support]]
   - [[#mode-line-support][Mode line support]]
   - [[#sticky-header-support][Sticky header support]]
   - [[#epub-support][Epub support]]
@@ -422,6 +423,18 @@ More information about org-roam package (including manual) can be found at [[htt
 *** Org-roam-server support
 To install support for [[https://github.com/org-roam/org-roam-server][org-roam-server]] set the variable =org-enable-roam-server=
 to =t=.
+
+*** Org-roam-protocol support
+To enable support for [[https://www.orgroam.com/manual.html#Roam-Protocol][Org Roam Protocol]] set the variable
+=org-enable-roam-protocol= to =t=.
+
+#+begin_src emacs-lisp
+  (setq-default dotspacemacs-configuration-layers '(
+    (org :variables
+         org-enable-roam-protocol t)))
+#+end_src
+
+And create a desktop file as described in the [[https://www.orgroam.com/manual.html#Roam-Protocol][org-roam manual]].
 
 ** Mode line support
 To temporarily enable mode line display of org clock, press ~SPC t m c~.
--- a/layers/+emacs/org/config.el
+++ b/layers/+emacs/org/config.el
@@ -94,4 +94,7 @@ ATTENTION: `valign-mode' will be laggy working with tables contain more than 100
   "If non-nil, enable org-appear in org-mode buffers.")
 
 (defvar org-enable-roam-server nil
-  "if non-nil, enable org-roam-server support")
+  "If non-nil, enable org-roam-server support")
+
+(defvar org-enable-roam-protocol nil
+  "If non-nil, enable org-roam-protocol (https://www.orgroam.com/manual.html#Roam-Protocol)")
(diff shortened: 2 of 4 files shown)


layers/+emacs/org/README.org
```diff
@@ -423,6 +424,18 @@ More information about org-roam package (including manual) can be found at [[htt
...
+To install support for [[https://www.orgroam.com/manual.html#Roam-Protocol][Org Roam Protocol]] set the variable
+=org-enable-roam-protocol= to =t=.
+
+#+begin_src emacs-lisp
+  (setq-default dotspacemacs-configuration-layers '(
+    (org :variables
+         org-enable-roam-protocol t)))
+#+end_src
+
+And create a desktop application as described in the [[https://www.orgroam.com/manual.html#Roam-Protocol][org-roam manual]].
```
lebensterben, line 437: According to Desktop Entry Specification of Freedesktop, it's either Desktop Entry File or Desktop File. 

*desktop application* is a misnomer.

layers/+emacs/org/packages.el
```diff
@@ -951,7 +951,10 @@ Headline^^            Visit entry^^               Filter^^                    Da
         "ra" 'org-roam-alias-add))
     :config
     (progn
-      (spacemacs|hide-lighter org-roam-mode))))
+      (spacemacs|hide-lighter org-roam-mode)
+      (if org-enable-roam-protocol
+          (add-hook 'org-roam-mode-hook (lambda ()
+                                          (use-package org-roam-protocol)))))))
```
lebensterben, line 957: Why not `(require 'org-roam-protocol)`?

lebensterben (changes requested):

layers/+emacs/org/README.org
KjartanOli: Noted, I used the terminoligy of the org-roam manual, but if the spec says something different then that's definitively  the terminology that should be used.

layers/+emacs/org/packages.el
KjartanOli: Yeah, I guess `use-package` doesn't provide any advantage over a plain require in this situation. I feel like I read somewhere in the Spacemacs docs that `use-package` should be prefered, but that was probably meant in some other context.

lebensterben: Yes. It did say in general you should not use `require` because spacemacs deliberately want to avoid the dependencies between packages.

But `require` is much simpler.

layers/+emacs/org/packages.el
```diff
@@ -951,7 +951,10 @@ Headline^^            Visit entry^^               Filter^^                    Da
         "ra" 'org-roam-alias-add))
     :config
     (progn
-      (spacemacs|hide-lighter org-roam-mode))))
+      (spacemacs|hide-lighter org-roam-mode)
+      (if org-enable-roam-protocol
```
lebensterben, line 955: Use `when` since there's no ELSE for `if`.

layers/+emacs/org/config.el
```diff
@@ -95,3 +95,6 @@ ATTENTION: `valign-mode' will be laggy working with tables contain more than 100
 
 (defvar org-enable-roam-server nil
   "if non-nil, enable org-roam-server support")
+
+(defvar org-enable-roam-protocol nil
+  "if non-nil, enable org-roam-protocol (https://www.orgroam.com/manual.html#Roam-Protocol)")
```
lebensterben, line 100: The documentation should starts with capital letter and ends with period.

The first line of documentation should be a short sentence < 70 characters.

layers/+emacs/org/README.org
```diff
@@ -423,6 +424,18 @@ More information about org-roam package (including manual) can be found at [[htt
 To install support for [[https://github.com/org-roam/org-roam-server][org-roam-server]] set the variable =org-enable-roam-server=
 to =t=.
 
+*** Org-roam-protocol support
+To install support for [[https://www.orgroam.com/manual.html#Roam-Protocol][Org Roam Protocol]] set the variable
```
lebensterben, line 428: From the manual:
> To enable Org-roam’s protocol extensions

There's nothing to be installed. `org-roam-protocol` is included in `org-roam`.

lebensterben (changes requested):
LGTM despite small issues

layers/+emacs/org/README.org
KjartanOli: Thanks for pointing that one out. I think I copied that section from one of the other sections, and must have forgotten to change that. I'll push a fix later today.

lebensterben:
I solved the merge conflict.
Merging now.
Thanks for your contribution.

KjartanOli:
Thanks, and thank you for the help with this.

PR #14688, opened 2021-04-20
Replace function aliases with their real names

KjartanOli:
`first` and `list-length` are aliases for `car` and `cl-list-length`
respectively. Using these aliases causes emacs to error with a `symbol's function
definition is void` error, but using their real names causes no error.

### DIFF
--- a/layers/+window-management/exwm/funcs.el
+++ b/layers/+window-management/exwm/funcs.el
@@ -58,4 +58,4 @@ Can show completions at point for COMMAND using helm"
 
 ;; Other utilities
 (defun exwm//flatenum (i ls)
-  (if ls (cons i (cons (first ls) (exwm//flatenum  (1+ i) (cdr ls)))) (list)))
+  (if ls (cons i (cons (car ls) (exwm//flatenum  (1+ i) (cdr ls)))) (list)))
--- a/layers/+window-management/exwm/packages.el
+++ b/layers/+window-management/exwm/packages.el
@@ -71,8 +71,8 @@
     ;; At the user's option, begin with even more workspaces
     (setq exwm-workspace-number
           (if exwm-workspace-number
-              (max exwm-workspace-number (list-length exwm--randr-displays))
-              (list-length exwm--randr-displays)))
+              (max exwm-workspace-number (length exwm--randr-displays))
+              (length exwm--randr-displays)))
     ;; The first workspaces will match the order in RandR
     (setq exwm-randr-workspace-output-plist
           (exwm//flatenum 0 exwm--randr-displays))


lebensterben:
cl-lib is required by various package, including our core library.
If any cl emulation function is available, their aliases should also be available.
I cannot conceive a case how your commit could fix anything. Again, if the aliases is not in the namespace, the underlying function isn't as well.

There are multiple issues with this layer.
- funcs.el and config.el don't have the heading.
- https://github.com/KjartanOli/spacemacs/blob/516397a82da52d23512a5a7beabea74577fa02b2/layers/%2Bwindow-management/exwm/packages.el#L16 Is not needed for emacs > 25. Our minimum supported version is 25.1, so it should be removed.

KjartanOli:
> cl-lib is required by various package, including our core library.
> If any cl emulation function is available, their aliases should also be available.
> I cannot conceive a case how your commit could fix anything. Again, if the aliases is not in the namespace, the underlying function isn't as well.
> 
> There are multiple issues with this layer.
> 
> * funcs.el and config.el don't have the heading.
> * https://github.com/KjartanOli/spacemacs/blob/516397a82da52d23512a5a7beabea74577fa02b2/layers/%2Bwindow-management/exwm/packages.el#L16 Is not needed for emacs > 25. Our minimum supported version is 25.1, so it should be removed.

This confused me greatly to. But the fact is that my emacs will not start if the aliases are used, but starts just fine if the real names are used. As for other issues with this layer, I did not create it, but I am willing to submit a PR to fix them if you want.

lebensterben:
Can you try to remove the `cl-generic` dependency first, without using your fix?

I'm wondering whether that's the cause.
(It shouldn't.)

KjartanOli:
I'm afraid that did not fix it. Is there some way to make emacs write the backtrace from `--debug-init` to a file? Maybe there are clues in there that I'm not seeing.

lebensterben:
Do you have `dotspacemacs-distribution` set to `spacemacs-base`?

KjartanOli:
No, I have it set to `spacemacs`, should I have it set to `spacemacs-base` for development?

lebensterben:
No. Because there's no mentioning of `(require 'cl-lib)` in `spacemacs-base` distribution.
But there're plenty of it in `spacemacs` distribution.

You can just add `(require 'cl-lib)`.

KjartanOli:
In the `(use-package exwm)` `:init` block? Or should it be added to `exwm-packages`?

lebensterben:
Never mind. No need to add `(require 'cl-lib)`

Just replace `first` by `car` and `list-length` by `length`.

```elisp
(defun cl-list-length (x)
  "Return the length of list X.  Return nil if list is circular."
  (cl-check-type x list)
  (condition-case nil
      (length x)
    (circular-list)))
```

lebensterben:
Merged.
Thanks for your contribution.

KjartanOli:
Thanks for the help in getting it to work. I'll probably open a separate PR to fix the other issues you mentioned with this layer later.

PR #14693, opened 2021-04-21
Fix issues with EXWM layer

KjartanOli:
*  Add the proprer header to layer files
*  Remove unnecessary `cl-generic` from `exwm-packages`
*  Replace `defvar`s with `defcustom`s in `config.el`

### DIFF
--- a/layers/+window-management/exwm/config.el
+++ b/layers/+window-management/exwm/config.el
@@ -1,3 +1,25 @@
+;;; config.el --- EXWM Layer packages File for Spacemacs
+;;
+;; Copyright (c) 2012-2021 Sylvain Benner & Contributors
+;;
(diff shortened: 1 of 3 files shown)


lebensterben:
We don't use defcustom to declare variables. We have a specialised one `spacemacs|defc` but it's still under development.

Example usage:

https://github.com/syl20bnr/spacemacs/blob/e290b1006072cd2ee4578b1c4e2c737c78c1675e/layers/%2Blang/go/config.el#L28

Alternatively you can just use defvar

lebensterben:
Use spacemacs|defc if you are going to specify its type. Otherwise use defvar.

lebensterben (approved):

lebensterben:

PR #14694, opened 2021-04-22
small refactor

lebensterben:
- Replaced `(cadr (assq KEY ALIST))` pattern with `let-alist` construct.
- Many times we first retrieve a key-value pair from an alist, and only when the
pair exists, we need to proceed. So it's advisable to use `when-let*` form to
instead of `let*` to bind the pair.
- Another common pattern is to let-bind a key-value pair from an alist, and then
modify the value if it exist, or insert a new key-value pair if it doesn't. This
can be replaced by `(setf (alist-get KEY ALIST) VALUE)`.


syl20bnr:
I did not know about `let-alist`.
The expension of `let-alist` is a bit different as it does `(cdr (assq ...))` and here we need `(cadr (assq ...))`. So the result gives 

```
 (e:(395) r:(9) l:(17) b:(67))
```

Applying the `car` on the dotted variable might fix it.

lebensterben:
Alternatively, we can make `configuration-layer/configured-package-stats` a real alist.

lebensterben:
@syl20bnr 
I've updated it and incorporated more refactor regarding alist.

PR #14698, opened 2021-04-22
[core] Fix 'append nesting in transient-state-format-hint macro

syl20bnr:
See comment: https://github.com/syl20bnr/spacemacs/commit/bfec25422108952598010a50babb87565f9ba935#r49828503

### DIFF
--- a/core/core-transient-state.el
+++ b/core/core-transient-state.el
@@ -118,8 +118,8 @@ effect if called after that point."
                                :columns ,prop-columns
                                :foreign-keys ,prop-foreign-keys
                                :body-pre ,prop-entry-sexp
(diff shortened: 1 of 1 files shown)


syl20bnr:
@lebensterben yes, the DEPTH argument is backward compatible so behavior is unchanged. If at some point we want to make sure that something is executed last no matter what we will be able to put 91.
I don't want to abuse the DEPTH value because it can become a nightmare to maintain if we start to have a lot of different value. Up to now `'append` always did the job so lets keep this way as long as we can. We could use `91` at some point if necessary.

lebensterben (approved):

PR #14699, opened 2021-04-23
defc if valid it is also safe local

JAremko:
Mark valid values as safe local. Not sure how safe it is  :thinking:

### DIFF
--- a/core/core-customization.el
+++ b/core/core-customization.el
@@ -91,6 +91,8 @@ NOTE: Variables defined with a group listed in
          `',standard)
       ,(format "%s\n\nTYPE: %s\n" doc type)
       :type ,type
+      :safe (lambda (val)
+              (or (spacemacs-customization//validate val ,type nil) t))
       :group group)
      (pcase ,safe
        ('t (put ',symbol 'safe-local-variable


JAremko:
It would be better if we can mark only some valid values as safe. For example:

This means that the variable is valid if it's boolean
``` elisp
(spacemacs|defc dotspacemacs-ask-for-lazy-installation t
  "If non-nil then Spacemacs will ask for confirmation before installing
a layer lazily."
  'boolean
  'spacemacs-dotspacemacs-layers)

And this means that it is only safe if value is 'vim' 
``` elisp
(spacemacs|defc dotspacemacs-editing-style 'vim
  "One of `vim', `emacs' or `hybrid'.
`hybrid' is like `vim' except that `insert state' is replaced by the
`hybrid state' with `emacs' key bindings. The value can also be a list
 with `:variables' keyword (similar to layers). Check the editing styles
 section of the documentation for details on available variables."
  '(choice (safe (const vim) (cons symbol sexp))
           (const emacs) (cons symbol sexp)
           (const hybrid) (cons symbol sexp))
  'spacemacs-dotspacemacs-init)
```

lebensterben:
I need to think about this

JAremko:
@lebensterben  My concern is that a variable can keep a concert value or a function (or a function's symbol) to calculate the value. In the second case it is not safe file variable.

lebensterben:
> @lebensterben My concern is that a variable can keep a value or a function (or a function's symbol) to calculate the value. In the second case it is not safe file variable even if it's valid.

We define the layer variable and decide how the variable is used in layer configuration. So we have the liberty to avoid the second case....

JAremko:
@lebensterben  good point.  The only draw back here is that it will be less explicit so more error prone than if you forced to inject `(safe )`.

JAremko:
Actually it will also be more verbose since you'll have to  add extra check in code for "dangerous" value.  And it won't be visible to the user in the same way as a declarative type is visible.

JAremko:
But them the most of variables are safe so adding `(safe ...)` everywhere  is sub optimal :smile:

lebensterben:
then add "risky"......

How about this, `spacemacs|defc` allows one of two keyword symbol, `:safe` and `:risky` but not both.

When `:safe`, it presume valid types are safe-local, except when the type is defined with `risky`.

When `:risky`, it presume the the variable is risky, but if any branch of the type declaration is marked with `safe`, it's taken as a safe-local value.

lebensterben:
```
(spacemacs|defc foo nil
  "doc"
  (choice (const BAR) (risky string))
  nil :safe)
```
This accepts the symbol BAR as safe-local; but not any string.

Or equivalently,

```
(spacemacs|defc foo nil
  "doc"
  (safe choice (safe const BAR) (string))
  nil :risky)
```

JAremko:
@lebensterben  Seems like adding more complexity for the minimal benefit :thinking:   I'll start re-implementing  `valdiate.el` and see how it goes :+1:

lebensterben:
And :safe or :risky takes an optional argument if you want to supply your own function as an validator.

PR #14707, opened 2021-04-24
[documentation]: Improved core-load-paths and core-versions

lebensterben:
- Conform to standard ELISP file practice.
- Rearranged the top-level forms s.t. Variables and constants are declared at top.
- Added section markers and documentations.
  - Use three or more `;` as a section header. They work well with built-in outline-mode. `M-j` and `M-k` is the binding to quickly navigate between them.
  - Optionally use `^L` (form-feed) as a divider for the ELISP file. It's rendered as a horizontal divider. And it works with `backward-page` (`C-x [`) and `forward-page` (`C-x ]`)

### DIFF
--- a/core/core-load-paths.el
+++ b/core/core-load-paths.el
@@ -118,21 +118,20 @@
 
 ;;;; Load Paths
 
-(defun add-to-load-path (dir)
+(defun spacemacs//add-to-load-path (dir)
   "Prepend DIR to `load-path'."
   (add-to-list 'load-path dir))
 
 ;; FIXME: unused function (Apr 25 2021 Lucius)
-(defun add-to-load-path-if-exists (dir)
+(defun spacemacs//add-to-load-path-if-exists (dir)
   "If DIR exists in the file system, prepend it to `load-path'."
   (when (file-exists-p dir)
-    (add-to-load-path dir)))
+    (spacemacs//add-to-load-path dir)))
 
 (dolist (suffix '(nil "libs/" "libs/spacemacs-theme/" "libs/forks"))
-  (add-to-load-path (concat spacemacs-core-directory suffix)))
+  (spacemacs//add-to-load-path (concat spacemacs-core-directory suffix)))
 
 ;;;; Themes
-
 (add-to-list 'custom-theme-load-path (concat spacemacs-core-directory
                                              "libs/spacemacs-theme/"))
 
--- a/core/core-load-paths.el
+++ b/core/core-load-paths.el
@@ -1,4 +1,4 @@
-;;; core-load-paths.el --- Spacemacs Core File  -*- no-byte-compile: t -*-
+;;; core-load-paths.el --- Spacemacs Core File  -*- no-byte-compile: t; lexical-binding: t -*-
 ;;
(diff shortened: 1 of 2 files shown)


JAremko (commented):
Shouldn't make difference for those files but having `-*- lexical-binding: t -*-`  in every core file sound like a good practice :thinking:

core/core-load-paths.el
```diff
@@ -20,83 +20,121 @@
...
+

+;;;; Load Paths
+
+(defun add-to-load-path (dir)
+  "Prepend DIR to `load-path'."
+  (add-to-list 'load-path dir))
+
+;; FIXME: unused function (Apr 25 2021 Lucius)
+(defun add-to-load-path-if-exists (dir)
```
JAremko, line 126: Maybe we should mark unused functions as obsolete.  Or even better. Since this function is public Spacemacs can warn(on call) that we are going to remove it soon and that the end user should copy pasta it to their configs or tell us why we should keep it (also this warning should be disableble(is this even a word) via .spacemacs var) How about that :thinking:

JAremko: Private unused functions can be safely removed since we don't guarantee anything about them.

lebensterben: I don't plan to remove unused functions in this PR. Just mark them.

Because this can be usable in other parts.

JAremko: Sure but marking them via code instead of comment will allow us to know when/if this function is used.

JAremko: or actually anywhere :thinking:

lebensterben: User normally should not use spacemacs functions anyway...

No guarantee of API stability for <1.0 release.

So I don't bother to add any error warning.

JAremko: > No guarantee of API stability for <1.0 release.

Oh boi.

lebensterben:
@JAremko 
Good to merge?

JAremko:
@lebensterben   We might rely on file names being expanded  but lets merge and see :smile:

JAremko:
Actually, it's a bad idea. Those vars are public - we can break user configs. Lets keep them expanded.

lebensterben:
no they are already expanded

lebensterben:
the only variable that needs to be expanded is `spacemacs-start-directory`, all other variable is simple concating it with a fixed string.

JAremko:
> the only variable that needs to be expanded is `spacemacs-start-directory`, all other variable is simple concating it with a fixed string.

If they all are expanded (on all platforms)  :+1:

lebensterben:
@JAremko 
It's easier to see here https://github.com/lebensterben/spacemacs/blob/ef13ffe2a0651e01445350e662b33fd4cbf36df8/core/core-load-paths.el#L31-L38

The first two variables needs to be expanded. And all of the followings are concating a absolute path with a fixed string. So no worry here.

I re-ordered them alphabetically by their default values.

JAremko:
Funny that no one noticed it before. Good to have a fresh pair eyes.

JAremko (approved):

PR #14706, opened 2021-04-23
Fix clone workspace

duianto:
### problem
Trying to clone a workspace `SPC l w c`
Does create a new workspace,
but it just shows the scratch buffer.

### notes
In addition to fixing the cloning.

Show a minibuffer message:
`Workspace: 1, cloned to Workspace: 2`

To make it clear that the cloning worked.
Because the only visible difference, is that
one of the left side mode line numbers changed.

PR #14709, opened 2021-04-24
Add workspace action messages and open home buffer

duianto:
### Added workspace action messages
`SPC l w` followed by a number key `0-9`
or `SPC l w s` single window workspace
>Workspace 2 created

`SPC l w 1` from another workspace
>Workspace switched to 1

`SPC l w 1` from the same workspace
>Already on Workspace 1

`SPC l w c` clones the current workspace to the next free slot
>Workspace 1 cloned to 3

`SPC l w d` close the current workspace
Workspace 3 closed

### And
Open the Spacemacs home buffer when creating a new workspace
with the number keys. Previously it opened the scratch buffer.

---

Opening the home buffer when creating a new workspace, might be a better starting point.
It also matches the current behavior when opening a single window workspace `SPC l w s`.

This PR also incorporates the changes in:
Fix clone workspace #14706

With some modifications:
Removed the mention about the minibuffer message from the docstring, (it seemed unnecessary).
And it uses a separate function to get the next free slot.

Should we close the separate fix clone workspace PR.
I can mention the clone fix in this commit message.

Or is it better to keep them separate?


lebensterben:
That PR is only a few lines and easy to review.

This is likely to take longer. So we can have both open.

lebensterben (approved):
LGTM

duianto:
Rebased.
I found a misleading message when trying to close the last workspace.
It's fixed in an included commit.

lebensterben:
good catch

PR #14713, opened 2021-04-25
Use timeclock-mode-line-display in default bindings

zjp:
Currently, if a user tries to use the 'ttm' key combination and their
emacs version is >24.3, timeclock-mode complains that
timeclock-modeline-display is deprecated. Spacemacs itself has
deprecated support for Emacs < 25.1 per CHANGELOG.develop. This commit
thus calls the new function without checking emacs-version for versions
for which the old function should be called.


lebensterben (approved):

lebensterben:
Merged. Thanks for your contribution!

PR #14715, opened 2021-04-26
Ruby layer refactor and new shadowenv layer

syl20bnr:
Related issue: https://github.com/syl20bnr/spacemacs/issues/14653

* new macro `spacemacs|eval-for-enabled-ruby-mode`
* use "dispatch" pattern with local vars hook for backend, test-runner, version-manager
* use spacebind
* diminish minor modes lighter
* new layer: shadowenv
* add shadowenv layer to ruby mode, python mode and compilation buffer


layers/+tools/shadowenv/README.org
```diff
@@ -0,0 +1,31 @@
+#+TITLE: shadowenv layer
+
+#+TAGS: layer|tool
+
+[[file:img/shadowenv.png]]
```
lebensterben, line 5: I suggest to write copyright notice for all image files from now on.

So add a footnote `[[file:img/shadowenv.png]] [fn:1]`
And `[fn:1] This logo image consists only of simple geometric shapes or text. It does not meet the threshold of originality needed for copyright protection, and is therefore in the public domain.`

layers/+tools/shadowenv/packages.el
```diff
@@ -0,0 +1,40 @@
...
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+(defconst shadowenv-packages
+  '(shadowenv))
+
+(defun shadowenv/init-shadowenv ()
+  (use-package shadowenv
+    :defer t
```
lebensterben, line 29: optionally we can use
`:spacediminish " Ⓥ" " V"`

It takes 0 to 3 arguments, same as those of `spacemacs|diminish`.
When the first argument is omitted, it's inferred from the package name, which is `(intern (format "%s-mode" PKG-NAME))`.

layers/+tools/shadowenv/packages.el
```diff
@@ -0,0 +1,40 @@
...
+    :init
+    (dolist (mode spacemacs--shadowenv-modes)
+      (add-hook (intern (format "%S-hook" mode)) 'shadowenv-mode)
+      (spacemacs|diminish shadowenv-mode " Ⓥ" " V")
+      (eval `(spacemacs|spacebind
+              :major
+              (,mode
```
lebensterben, line 36: A think this is a common case. `spacemacs|spacebind` should accept either a symbol or a list of symbols as for `MODE`.

layers/+lang/ruby/funcs.el
```diff
@@ -160,9 +225,5 @@ Called interactively it prompts for a directory."
...
 ;; Prettier
 
-(defun spacemacs/ruby-format ()
-  (interactive)
-  (call-interactively 'prettier-js))
-
-(defun spacemacs/ruby-fmt-before-save-hook ()
-  (add-hook 'before-save-hook 'spacemacs/ruby-format t t))
+(defun spacemacs//ruby-add-prettier-js-before-save-hook ()
+  (add-hook 'before-save-hook 'prettier-js t t))
```
lebensterben, line 229: the third argument to `add-hook` is `DEPTH`.
It defaults to 0, and any non-nil symbol implies 90.
So this hook is likely to be run after other hooks (-100 is first and 100 is last).

Is this really the behaviour we want?

layers/+lang/ruby/funcs.el
```diff
@@ -22,11 +22,55 @@
...
+      (cl-nsubstitute-if new 'spacemacs//ruby-substitute-predicate body))))
+
+(defun spacemacs//ruby-substitute-predicate (body)
+  "Predicate to replace all `old' occurrences by `new' occurrences in BODY.
+Reccurse if BODY is a list."
+  (if (listp body)
+      (progn
+        (cl-nsubstitute-if new 'spacemacs//ruby-substitute-predicate body)
+        nil)
+    (eq body old)))
```
lebensterben, line 67: `old` is dynamically scoped. It's better to make it a function argument.

layers/+lang/ruby/funcs.el
```diff
@@ -22,11 +22,55 @@
...
+    ;; ruby-mode
+    (let ((new 'ruby-mode)
+          (old 'mode))
+      (cl-nsubstitute-if new 'spacemacs//ruby-substitute-predicate body))
+    (let ((new 'ruby-mode-hook)
+          (old 'hook))
+      (cl-nsubstitute-if new 'spacemacs//ruby-substitute-predicate body))
+    (let ((new 'ruby-mode-local-vars-hook)
+          (old 'local-vars-hook))
+      (cl-nsubstitute-if new 'spacemacs//ruby-substitute-predicate body))))
```
lebensterben, line 58: This part can be simplified.
First we let-bound a list of modes based on whether `ruby-enable-enh-ruby-mode` is t.
```elisp
(let* ((rubymode (if ruby-enable-enh-ruby-mode
                     "enh-ruby"
                  "ruby")
       (pairs (cl-loop for old in '("mode" "hook" "local-vars-hook")
                       for new = (s-join "-" (list rubymode old))
                       collect (cons old new)))
 ..)
```
Then we can simply
```elisp
  (dolist (pair pairs)
     (cl-nsubstitute-if (cdr pair) 'spacemacs//ruby-substitute-predicate (car pair)))
```

lebensterben: The real question is, how efficient is `cl-nsubstitute-if`.

layers/+tools/shadowenv/packages.el
JAremko: @lebensterben  Already planed :smile:   We have a bunch of layers doing this kind of templating.

syl20bnr: I did not think about static analysis. I guess we should drop the idea of static analysis because it will be like mission impossible to catch all the calls to `spacebind` correctly. For example here, even if I drop the macro to replace with a `dolist` (which would an inferior solution) the static analysis will have a hard time to detect each iteration of the `spacebind` call.
A better place for static analysis could be the `README.org` files once the key bindings have been rendered into them.

layers/+tools/shadowenv/packages.el
syl20bnr: Thanks for mentioning `:spacediminish`, I was not aware of it.

Done.

layers/+lang/ruby/funcs.el
```diff
@@ -97,39 +150,49 @@
...
+
+;; minitest
+(defun spacemacs//ruby-setup-minitest ()
+  "Setup minitest test runner."
+  (minitest-enable-appropriate-mode))
+
 ;; rspec
+(defun spacemacs//ruby-setup-rspec ()
+  "Setup rspec test runner."
+  (when (eq ' 'rvm)
```
lebensterben, line 177: typo?

syl20bnr:
I updated the conventions for toggles lighters in the mode line, see commit message https://github.com/syl20bnr/spacemacs/commit/c8ae548c9
These conventions were missing from the convention doc, this commit adds them.

layers/+tools/shadowenv/packages.el
syl20bnr: Makes me think that the macro here will prevent from correctly rendering in the `README.org`. 
You are right, passing an explicit list of modes is the way to do.

layers/+tools/shadowenv/README.org
```diff
@@ -0,0 +1,31 @@
...
+- [[#description][Description]]
+  - [[#features][Features:]]
+- [[#install][Install]]
+- [[#key-bindings][Key bindings]]
+
+* Description
+This layer adds support for [[https://shopify.github.io/shadowenv/][shadowenv]]
+
+** Features:
+- Automatically activatefouhnd shadowenv each supported buffer locally
```
Kazark, line 17: typo: `activatefouhnd`

layers/+lang/ruby/funcs.el
syl20bnr: I guess so. The addition of `DEPTH` is a rather new one and it is meant to be backward compatible. 
If there was `t` to append, then `DEPTH` should keep the same behavior.

layers/+lang/ruby/funcs.el
syl20bnr: I thought about it but I could not do that, the predicate for `cl-substitute` is a function with one argument.
What could be done to improve upon this ? Sort of a curried function ? Or maybe declare variables explicitly ? Another macro ?

layers/+lang/ruby/funcs.el
syl20bnr: Good question, I would assume it is the fastest as it is the destructive version `cl-substitute-if` but this is a wild guess 😅 
Going to update with your great implementation for now.

syl20bnr: Done

layers/+lang/ruby/funcs.el
syl20bnr: Fixed.

layers/+tools/shadowenv/README.org
syl20bnr: Fixed.

layers/+tools/shadowenv/README.org
syl20bnr: :+1: done.

layers/+tools/shadowenv/packages.el
syl20bnr: OK so we can keep it like it for now. When we support a list of modes we will migrate the spacebind calls out of the macro.

layers/+lang/ruby/funcs.el
lebensterben: Solution one

well. let's make it to take pair, (OLD . BODY).
it's a function taking ONE argument.
`(cl-nsubstitute-if new 'spacemacs//ruby-substitute-predicate (cons old body)`

---

Make it take two arguments, the first is `old`
Solution two `(cl-nsubstitute-if new (apply-partially old) body)`.

lebensterben:
Also it'd be nice if you can change
`(add-hook 'foo 'bar)` to `(add-hook 'foo #'bar)` where bar is a symbol of function.

This is not strictly enforced now but is easier for byte compiler to understand such symbol is a function.

smile13241324 (changes requested):
Just reviewed your PR. I have found some smaller typos which should be changed before it can be merged.

syl20bnr:
Do not merge this yet, I'm stil working on it.

layers/+spacemacs/spacemacs-defaults/keybindings.el
```diff
@@ -733,6 +734,15 @@ respond to this toggle."
   "xwc" 'spacemacs/count-words-analysis
   "xwr" 'spacemacs/randomize-words
   "x TAB" 'indent-rigidly)
+;; Version and Environment ----------------------------------------------------
+(spacemacs|spacebind
+ "Version/Environment"
+ :global
+ (("V" "Version/Environment"
```
JAremko, line 741: @syl20bnr  Extra set of `()`


@lebensterben  I think we should detect it and report as en error since it is easy to introduce when refactoring.

smile13241324:
Given that this is now older than 3 years I would like to go ahead and close it for now.

PR #14716, opened 2021-04-26
Improve quick-start guide

dalanicolai:
Although the commit message below is only my opinion, I have written it in a style more suitable for a commit message. Anyway, this PR is only a proposal and if you agree with it then you could merge it (or provide feedback first).

 The [quick-start](https://www.spacemacs.org/doc/QUICK_START.html) starts with
 explaining that users can build their own layers. Instead, it should start by
 showing how easy and straightforward it is to use one of the existing layer,
 then continue about the possibility of creating personal layers. It is only a
 small detail, but it can make a substantial difference for people who peek into
 the quick-start guide and decide if it is worth the trouble to switch to
 Spacemacs (Many newcomers think that even only trying another editor, is
 probably not worth it because they are already using vim).

A guide that starts explainining that you can build your own layers in
Spacemacs that exists of a directory containing a packages.el file, is not a
quick-start guide.

Cheers!

### DIFF
--- a/doc/QUICK_START.org
+++ b/doc/QUICK_START.org
@@ -16,11 +16,28 @@
 
 * Configuration layers
 Spacemacs divides its configuration into self-contained units called
-=configuration layers=. These layers are stacked on top of each other
-to achieve a custom configuration.
+=configuration layers=. These layers are stacked on top of each other to achieve
+a custom configuration.
 
 By default Spacemacs uses a dotfile called =~/.spacemacs= to control which
-layers to load. Within this file you can also configure certain features.
+layers to load. Within this file you can also configure certain features. This
+design makes it extremely easy to turn Spacemacs into a fully integrated
+development environment.
+
+For example to get python support, simply add =python= to the list of
+=dotspacemacs-configuration-layers= in your =~/.spacemacs=. Find the
+=~/.spacemacs= file by pressing =SPC f e d= and then after adding the layer,
+reload it with =SPC f e R=. Now open a =.py= file (=SPC f f=) to find the python
+environment has been fully configured. Some extra configuration might make the
+environment even more powerful, which is very well described in the [[https://develop.spacemacs.org/layers/+lang/python/README.html][layer's
+documentation]] that can be accessed by pressing =SPC h l= and selecting the
+=python= layer entry. For configuration of specific packages within a layer,
+=SPC h SPC= provides quick navigation functionality for jumping directly to the
+relevant location within the configuration files.
+
+A list of pre-configured layers is available [[https://develop.spacemacs.org/layers/LAYERS.html][here]]. If you still would like to
+configure anything not covered by any layer, then it is easy to build a personal
+layer.
 
 A configuration layer is a directory containing at least a =packages.el=
 file which defines and configures packages to be downloaded from Emacs


lebensterben:
This is nice. But our quick start is getting too long already... It doesn't fit the name.

doc/QUICK_START.org
```diff
@@ -16,11 +16,25 @@
...
-layers to load. Within this file you can also configure certain features.
+layers to load. Within this file you can also configure certain features. This
+design makes it extremely easy to turn Spacemacs into a fully integrated
+development environment.
+
+For example to get python support, simply add =python= to the list of
+=dotspacemacs-configuration-layers= in your =~/.spacemacs=. Find the
+=~/.spacemacs= file by pressing =SPC f e d=, and after adding the layer reload
+it with =SPC f e R=. Now open a =.py= file (=SPC f f=) to find the python
+environment has been fully configured. Some small extra configurations might
```
syl20bnr, line 31: It think `configuration` singular is better.

doc/QUICK_START.org
```diff
@@ -16,11 +16,25 @@
...
+make the environment even more powerful, which is very well described in the
+[[https://develop.spacemacs.org/layers/+lang/python/README.html][layer's documentaion]].
+
+A list of pre-configured layers is available [[https://develop.spacemacs.org/layers/LAYERS.html][here]]. If you still would like to
+configure things not covered by any of the layers then it it easy to build a
+personal layer.
```
syl20bnr, line 37: You can also mention the in-app key binding `SPC h l` or `SPC h SPC` (which combines several sources into one).

syl20bnr:
> But our quick start is getting too long already... It doesn't fit the name.

😂 we could indeed do better.
Also I really want to do a comprehensive layer tutorial at some point. The layer system is very expressive but half of the feature are not really documented. The ideal tutorial would be an interactive org file, I have yet to do a POC on this.

doc/QUICK_START.org
```diff
@@ -16,11 +16,25 @@
...
+design makes it extremely easy to turn Spacemacs into a fully integrated
+development environment.
+
+For example to get python support, simply add =python= to the list of
+=dotspacemacs-configuration-layers= in your =~/.spacemacs=. Find the
+=~/.spacemacs= file by pressing =SPC f e d=, and after adding the layer reload
+it with =SPC f e R=. Now open a =.py= file (=SPC f f=) to find the python
+environment has been fully configured. Some small extra configurations might
+make the environment even more powerful, which is very well described in the
+[[https://develop.spacemacs.org/layers/+lang/python/README.html][layer's documentaion]].
```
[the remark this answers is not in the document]
dalanicolai: This makes me realize that links don't get checked by flyspell

doc/QUICK_START.org
dalanicolai: Good point

dalanicolai:
Thanks for the feedback, I have updated the PR.

The interactive tutorial sounds very interesting; would be interesting to learn more about those hidden expressive features.

As expressed already I would suggest to make the quick-start more like a quick showcase of the beast that is Spacemacs, just my 2 cents...

Looking a bit deeper into the quick-start file, I find a section that explains how to create a dotfile or use a .spacemacs.d directory. To keep it short,  I would question also if these should be part of a quick-start (as the dotfile is created automatically at the first Spacemacs startup). But you were reconsidering the quick-start guide already of course...

lebensterben:
Ideally the quick start is literally about how to install Spacemacs and any possibility dependencies. Showcases can be put in separate files.

lebensterben:
We can even ship a few example dotfiles in addition to show cases.

dalanicolai:
Good point about the quick-start guide. Just to clear things up, I did not mean we should literally make it a showcase, but personally I would use it to make the newcomer experience some of Spacemacs its strong "selling" points (and maybe even mention some "weak" points, like longer startup time and how to deal with it). The reason is that I have seen various vim users trying out Spacemacs, but not being impressed (or patient) enough to stay with it. And I think there is a good possibility that if they understood it a little more, then they would have been more patient. I find this important because I think a larger community makes a stronger community. Just another 2 cents, all open for discussion...

lebensterben:
I don't object this PR and I think it is pretty well written.

My suggestions can be discussed in future PRs.

If all issues are addressed and resolved, we can merge this.

smile13241324 (approved):
This really looks well. 
I also think that starting our quickstart guide with something constructive as of how to use what is already there is much more useful than directly starting with personal layers.

I mean the whole idea is to have only a small maintainable config in Spacemacs in contrast to the Emacs or Vim approach.

Issue #13054, opened 2019-12-12
Missing gpg key?


JoKeyser:
During initial install on Ubuntu 18.04, I receive this gpg error:

```
Failed to verify signature archive-contents.sig:
No public key for 066DAFCB81E42C40 created at 2019-12-11T14:10:02+0100 using RSA
Command output:
gpg: Signature made Mi 11 Dez 2019 14:10:02 CET
gpg:                using RSA key C433554766D3DDC64221BFAA066DAFCB81E42C40
gpg: Can't check signature: No public key
```

And when I try to `gpg --recv-keys 066DAFCB81E42C40`, I get this:

```
gpg: key 066DAFCB81E42C40: 1 signature not checked due to a missing key
gpg: key 066DAFCB81E42C40: "GNU ELPA Signing Agent (2019) <elpasign@elpa.gnu.org>" not changed
gpg: Total number processed: 1
gpg:              unchanged: 1
```

JoKeyser:
Oh, the fix was simple:
```
$ gpg --homedir ~/.emacs.d/elpa/gnupg --receive-keys 066DAFCB81E42C40

gpg: key 066DAFCB81E42C40: public key "GNU ELPA Signing Agent (2019) <elpasign@elpa.gnu.org>" imported
gpg: no ultimately trusted keys found
gpg: Total number processed: 1
gpg:               imported: 1
```

karlicoss:
Related: https://github.com/aquamacs-emacs/aquamacs-emacs/issues/166

karlicoss:
I wonder if it's worth reopening?
Not sure what's the proper way to resolve this would be, but this must be very confusing for people new to Spacemacs (half of packages failing to install).

duianto:
Emacs 26.3 is supposed to have fixed the signature issue.

And the `ppa:kelleyk/emacs` has also updated the keys for older Emacs versions:
ELPA signing key expired https://github.com/kelleyk/ppa-emacs/issues/9

So the issue might have been fixed in linux, maybe the mac repositories need to update the key in their builds as well.

vvarik:
Not fixed in Linux (Ubuntu 18.04.4), just ran into it today. I can confirm it is confusing for new people. For instance, I don't know whether I should 1) just import the gpg key and restart; 2) remove everything in `elpa` except the `gnupg` folder and then import gpg key; 3) remove everything in `elpa` and issue `emacs --insecure`

dennismayr:
[...]

PR #14723, opened 2021-04-28
Bumped minimum emacs version to 26.1

lebensterben:
In https://github.com/syl20bnr/spacemacs/pull/14497 we removed legacy code for emacs version < 25.
There was a discussion about bumping minimum emacs version to 26.1.

This PR incorporated such change and removed all legacy codes.

26.1 is chosen as it's the latest stable version of Emacs on Debian.
If we decided to use `26.3`, I would revise this PR accordingly.


duianto (commented):
This toggle can also be removed:
https://github.com/syl20bnr/spacemacs/blob/d24249b7597a30f759703ca99ed04bb65b4f026a/layers/%2Bspacemacs/spacemacs-defaults/packages.el#L36-L37

26.3 would probably be better, because there were verification issues before 26.3.
Missing gpg key? #13054

Which required some workarounds to update the keys:
https://metaredux.com/posts/2019/12/09/dealing-with-expired-elpa-gpg-keys.html

Emacs 26.3 released
https://lists.gnu.org/archive/html/emacs-devel/2019-08/msg00577.html
>This release is mainly a maintenance release, which contains a new GPG key for
GNU ELPA packages.

smile13241324 (approved):
Looks good, I'll merge it in a minute.

smile13241324:

PR #14728, opened 2021-04-30
[ocaml] Fixed variable name to make ocamlformat properly working on save

n-raymond:
According to https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Blang/ocaml/packages.el#L160, the documentation could not be followed to make `ocamlformat` properly work when saving a file.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2821,6 +2821,7 @@ files (thanks to Daniel Nicolai)
 - Fixed backspace delete (thanks to Hasan Alirajpurwala)
 - Fixed flycheck-ocaml not initialized warning (thanks to Artur Juraszek)
 - Added =dune= support via =dune-mode= (thanks to Lyman Gillispie)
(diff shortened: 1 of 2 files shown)


lebensterben:
Looks good. 

- The meaning of format-before-save is very obvious. No ambiguity.
- But overall we name such functionality as format-on-save. (format after save doesn't make sense so this implies it's formatted BEFORE save).

What do you think?

n-raymond:
Hi @lebensterben ! Yes I prefer `format-on-save` also and find it more straightforward.
I just did the few needed modification. What do you think ?

smile13241324 (approved):
Looks good now will be merged soon.

bcc32:
FYI, I based this originally on `go-format-before-save`, so maybe we should update that variable name as well.

lebensterben:
Agree.
Just define it as an obsolete alias.

PR #14730, opened 2021-05-02
Warn about pdumper not working together with Native Compilation

sdwolfz:
Pdumper does not work with Native Compilation. Reference: https://lists.gnu.org/r/bug-gnu-emacs/2020-12/msg00589.html

Updated docs and config with a warning.


core/templates/.spacemacs.template
```diff
@@ -89,9 +89,14 @@ It should only modify the values of Spacemacs settings."
   ;; This setq-default sexp is an exhaustive list of all the supported
   ;; spacemacs settings.
   (setq-default
-   ;; If non-nil then enable support for the portable dumper. You'll need
-   ;; to compile Emacs 27 from source following the instructions in file
+   ;; If non-nil then enable support for the portable dumper. You'll need to
```
lebensterben, line 92: need to check the line length

core/templates/.spacemacs.template
```diff
@@ -89,9 +89,14 @@ It should only modify the values of Spacemacs settings."
...
+   ;; compile Emacs 27 from source following the instructions in file
    ;; EXPERIMENTAL.org at to root of the git repository.
+   ;;
+   ;; WARNING: pdumper does not work with Native Compilation, so if you get an
```
lebensterben, line 96: this is too intrusive. I prefer to keep this shorter.

Also, the better way of handling this is to disable pdump automatically when native compilation is enabled.

lebensterben (changes requested):

syl20bnr:
Is it possible to detect that Native Compilation is enabled ?
If yes, I agree that we should find a way to disable it automatically and print a message in the message buffer about it.

lebensterben:
@syl20bnr 

If `comp-ensure-native-compiler` doesn't throw error and if `comp-speed` is not -1,
then native-compilation is available and enabled.

```elisp
(defcustom comp-speed 2
  "Optimization level for native compilation, a number between -1 and 3.
 -1 functions are kept in bytecode form and no native compilation is performed.
  0 native compilation is performed with no optimizations.
  1 light optimizations.
  2 max optimization level fully adherent to the language semantic.
  3 max optimization level, to be used only when necessary.
    Warning: with 3, the compiler is free to perform dangerous optimizations."
  :type 'integer
  :safe #'integerp
  :version "28.1")
```

```elisp
 (defun comp-ensure-native-compiler ()
   "Make sure Emacs has native compiler support and libgccjit can be loaded.
 Signal an error otherwise.
 To be used by all entry points."
   (cond
    ((null (featurep 'native-compile))
     (error "Emacs was not compiled with native compiler support (--with-native-compilation)"))
    ((null (native-comp-available-p))
     (error "Cannot find libgccjit library"))))
```

sdwolfz:
Let's deal with this when emacs 28 will be released.

lebensterben:
I've just updated it so that pdumper is not enabled when native compilation is enabled.

lebensterben (approved):

sdwolfz:
Thanks for picking up the work and ressurecting this @lebensterben

lebensterben:
merging now.

PR #14732, opened 2021-05-03
[org] sane default strategies for enforcing `TODO` dependencies

Kazark:
Part of the excellence of Spacemacs is that it helps beginners to ease into the ELisp (or even mostly stay out of it). When I was first using Org, I was surprised that there was no existing one-and-done way to turn on dependency enforcement. After reading up on it and experimenting, I got a configuration represented by `'naive-auto` in this PR. I used that for a long time, but after beginning to use more complex Org workflows, it didn't work well for me anymore. By then I was much more comfortable in ELisp, and the result of much experimentation is represented by`'semiauto` here, which is slightly more intrusive to the user's workflow but surprisingly unopinionated. I envision that if someone what it to get really fancy with it, a `dwim-auto` or `smart-auto` would also be possible, that would attempt to inspect the workflows that a user had defined and transition to a state without prompting if it could be determined unambiguously. I didn't go that far, but hopefully this could ease the learning curve for the next poor soul who begins to dabble in the vast sea that is Org Mode.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2822,6 +2822,8 @@ files (thanks to Daniel Nicolai)
 - Fixed flycheck-ocaml not initialized warning (thanks to Artur Juraszek)
 - Added =dune= support via =dune-mode= (thanks to Lyman Gillispie)
 **** Org
+- Provide some sane default strategies for enforcing =TODO= dependencies via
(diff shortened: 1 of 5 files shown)


smile13241324 (approved):
This looks really good, @Kazark. Will be merged in a minute.

PR #14736, opened 2021-05-04
[ocaml] Fix name of layer variable in defvar

bcc32:
The variable was renamed in #14728 to match the documentation, but the
defvar was not updated.


lebensterben:
Define the original symbol as an obsolete alias is better.
It won't be a breaking change.

bcc32:
Good idea, thanks.

lebensterben (approved):

lebensterben:

PR #14741, opened 2021-05-05
Update pyls install command to use quotes

zjp:
This commit reflects pip's own documentation. Quotes are required
around 'python-language-server[all]' to install it correctly.


lebensterben:

Issue #5471, opened 2016-03-12
`evil--jump-forward` does not work


eltechnic0:
I have tried with both the latest master and develop branch but nothing. Apparently  `evil-jumps-struct-idx` always returns `-1`. I tried looking that symbol up but I am unable to find it anywhere... I am sure that the problem is there, since manually executing `evil--jumps-jump-to-index` does jump as expected.

jkaye2012:
+1, any update on this?

StreakyCobra:
Could anyone report more information using [this template](https://github.com/syl20bnr/spacemacs/wiki/Debugging#how-to-make-a-great-bug-report)? It will help us to help you :-)

jkaye2012:
Well, in filling out the template I realized that I was not hitting leader before C-i, so this is actually not a problem for me. Apologies :)

StreakyCobra:
Uh? If you want to use `C-i` you should set `dotspacemacs-distinguish-gui-tab` to `t` in your `.spacemacs`.

jkaye2012:
Do you recommend something else? Leader C-i seems to be the default binding (at least I don't believe I've changed it)

TheBB:
I'm not sure what you're talking about. There is no `SPC C-i` binding.

NathanLovato:
`evil-jump-forward` works here in the latest commit on the develop branch, but as explained above by @StreakyCobra, by default when you press `<C-i>` emacs acts as if you were pressing `TAB`. In your init.el file, set `dotspacemacs-distinguish-gui-tab t` in the `dotspacemacs/init` function and it should work.

PR #14744, opened 2021-05-07
[evil-better-jumper] Update layer documentation

HanshenWang:
https://github.com/syl20bnr/spacemacs/issues/5471 helps point out the non obvious default binding of `dotspacemacs-distinguish-gui-tab` to nil, this README update makes it explicit, hopefully saving future users of this new layer some troubleshooting of the `C-i` keybinding.

### DIFF
--- a/layers/+vim/evil-better-jumper/README.org
+++ b/layers/+vim/evil-better-jumper/README.org
@@ -19,7 +19,17 @@ locations.
 * Install
 To use this configuration layer, add it to your =~/.spacemacs=. You will need to
 add =evil-better-jumper= to the existing =dotspacemacs-configuration-layers=
-list in this file.
+list in this file. If =C-i= inserts an indent instead of
+=better-jumper-jump-forward=, please use =M-x describe-variable= to check the value
+of the symbol =dotspacemacs-distinguish-gui-tab= and ensure it is set to true in
+the dotspacemacs file.
+
+#+begin_src emacs-lisp
+  (defun dotspacemacs/user-config ()
+    ;; ...
+    (setq dotspacemacs-distinguish-gui-tab t)
+    )
+#+end_src
 
 * Key bindings
 
--- a/layers/+vim/evil-better-jumper/README.org
+++ b/layers/+vim/evil-better-jumper/README.org
@@ -18,17 +18,16 @@ locations.
 
 * Install
 To use this configuration layer, add it to your =~/.spacemacs=. You will need to
-add =evil-better-jumper= to the existing =dotspacemacs-configuration-layers=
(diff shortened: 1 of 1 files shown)


layers/+vim/evil-better-jumper/README.org
```diff
@@ -19,7 +19,17 @@ locations.
 * Install
 To use this configuration layer, add it to your =~/.spacemacs=. You will need to
 add =evil-better-jumper= to the existing =dotspacemacs-configuration-layers=
-list in this file.
+list in this file. If =C-i= inserts an indent instead of
```
lebensterben, line 22: C-i don't necessarily insert an indentation. It can binded to other functions.

Recommendation:

If =C-i= acts like =TAB= instead of ....

layers/+vim/evil-better-jumper/README.org
```diff
@@ -19,7 +19,17 @@ locations.
...
+=better-jumper-jump-forward=, please use =M-x describe-variable= to check the value
+of the symbol =dotspacemacs-distinguish-gui-tab= and ensure it is set to true in
```
lebensterben, line 24: remove "the symbol"

set to "t" not "true"

layers/+vim/evil-better-jumper/README.org
```diff
@@ -19,7 +19,17 @@ locations.
...
+the dotspacemacs file.
+
+#+begin_src emacs-lisp
+  (defun dotspacemacs/user-config ()
+    ;; ...
+    (setq dotspacemacs-distinguish-gui-tab t)
```
lebensterben, line 30: Doesn't look quite right.

https://github.com/syl20bnr/spacemacs/blob/a26121db4f60d49dd0ccb0bcf4a4fc538a11bc67/core/templates/.spacemacs.template#L275

lebensterben (changes requested):

layers/+vim/evil-better-jumper/README.org
HanshenWang: Yes. Done.

layers/+vim/evil-better-jumper/README.org
HanshenWang: Good catch

layers/+vim/evil-better-jumper/README.org
HanshenWang: Yeah I do my best to keep an eye on .spacemacs.template to keep it in sync with my own dotfile, but I have to manually start a diff.

Corrected snippet to its proper location, user-init

smile13241324 (approved):
Now it looks good.

PR #14750, opened 2021-05-08
[home] Only jump to visible nr lines

duianto:
problem
The number keys searches for a line starting with a number,
from the beginning to the end of the buffer,
even if they are above or below the window.

solution
Only search for the visible number lines.


lebensterben:
This is a great idea.
We should always comply principle of least astonishment

PR #14751, opened 2021-05-08
[home] New var dotspacemacs-show-startup-list-numbers

duianto:
This also defines the Spacemacs home buffer key bindings,
in the `emacs-startup-hook`.

Because the keys were being defined to early,
before the new value of: `dotspacemacs-show-startup-list-numbers`
was set in `.spacemacs`.


smile13241324 (approved):
:+1:

Issue #14577, opened 2021-03-30
core-jump: reference-handler api is not working at all


lebensterben:
We've created internal APIs similar to jump-handlers (jump to definition) for jumping to references, but its' never actually used by any modes.
It's very surprising to me this is never implemented. Below is a minimal example how we can implement this:

```elisp
(setq spacemacs-default-reference-handlers '(xref-find-references))
```

This is surely not smart, because it's just matching the input token, but this is better than not having anything.

For example, often time in `emacs-lisp` mode this is all what I need.

Any thoughts on why this issue is never addressed?

tshu-w:
Maybe not having a shortcut like `spacemacs/jump-to-definition` is a big reason that people don't know about this feature.

lebensterben:
There is, `g r`. The thing is the function that creates the binding is never called.

tshu-w:
I have no bindings on my side, maybe there is something wrong with my configuration.

lebensterben:
https://github.com/syl20bnr/spacemacs/blob/eaf2e932e6aca3fe5a53715089f9566d9dacffe4/core/core-jump.el#L54-L75

As I said,
> There is, `g r`. The thing is the function that creates the binding is never called.

tshu-w:
> https://github.com/syl20bnr/spacemacs/blob/eaf2e932e6aca3fe5a53715089f9566d9dacffe4/core/core-jump.el#L54-L75
> 
> As I said,
> 
> > There is, `g r`. The thing is the function that creates the binding is never called.

Sorry, I misunderstood your meaning.

thanhvg:
It never worked. I am glad it didn't work because xref.el in core emacs is better.

https://github.com/syl20bnr/spacemacs/commit/c825035e48c047f907968813a4bf57cace05ac21#commitcomment-37484537

Spacemacs `core-jump` was great at the time when `lsp` and `xref.el` weren't  a thing yet. Now it is a redundant feature. I also have another pr on core-jump https://github.com/syl20bnr/spacemacs/pull/13448 sitting for a year but it will never see the light of day I guess.

lebensterben:
@thanhvg 
Thus my simple solution described above just uses xref. The only real benefit is "g r" could be bounded for all prog-mode.

And yes, your unmerged PR deserves to be reviewed.

We can have both ", g d/r" and "g d/r" and both intelligently determine its source.

thanhvg:
[...]

PR #14752, opened 2021-05-08
[core] remove jump-to-reference from core-jump

thanhvg:
we don't need it and it doesn't work

refs:
https://github.com/syl20bnr/spacemacs/issues/14577
https://github.com/syl20bnr/spacemacs/commit/c825035e48c047f907968813a4bf57cace05ac21


smile13241324 (approved):
:+1:

lebensterben:
I found a way to use it but I forgot the details...

PR #14754, opened 2021-05-08
Extend spacemacs/copy-file command with extra actions

dalanicolai:
This PR implements an alternative for PR #8974.

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -1345,10 +1345,55 @@ the right."
   (interactive)
   (set-buffer-file-coding-system 'undecided-dos nil))
 
-(defun spacemacs/copy-file ()
-  "Write the file under new name."
-  (interactive)
-  (call-interactively 'write-file))
+(defun spacemacs/save-as (filename &optional visit)
+  "Save current buffer or active region as specified file.
+When called interactively, it first prompts for FILENAME, and then asks
+whether to VISIT it, and if so, whether to show it in current window or
+another window. When prefixed with a universal-argument \\[universal-argument], include
+filename in prompt.
+
+FILENAME  a non-empty string as the name of the saved file.
+VISIT     When it's `:current', open FILENAME in current window. When it's
+          `:other', open FILENAME in another window. When it's nil, only
+          save to FILENAME but does not visit it. (Default to `:current'
+          when called from a LISP program.)
+
+When FILENAME already exists, it also asks the user whether to
+overwrite it."
+  (interactive (let* ((filename (expand-file-name (read-file-name "Save buffer as: " nil nil nil
+                                                                  (when current-prefix-arg (buffer-name)))))
+                      (choices  '("Current window"
+                                  "Other window"
+                                  "Don't open"))
+                      (actions  '(:current :other nil))
+                      (visit    (let ((completion-ignore-case t))
+                                  (nth (cl-position
+                                        (completing-read "Do you want to open the file? "
+                                                         choices nil t)
+                                        choices
+                                        :test #'equal)
+                                       actions))))
+                 (list filename visit)))
+  (unless (called-interactively-p 'any)
+    (cl-assert (and (stringp filename)
+                    (not (string-empty-p filename))
+                    (not (directory-name-p filename)))
+               t "Expect a non-empty filepath, found: %s")
+    (setq filename (expand-file-name filename)
+          visit (or visit :other))
+    (let ((choices '(:current :other nil)))
+      (cl-assert (memq visit choices)
+                 t "Found %s, expect one of %s")))
+  (let ((dir (file-name-directory filename)))
+    (unless (file-directory-p dir)
+      (make-directory dir t)))
+  (if (use-region-p)
+      (write-region (region-beginning) (region-end) filename nil nil nil t)
+    (write-region nil nil filename nil nil nil t))
+  (pcase visit
+    (:current (find-file filename))
+    (:other   (funcall-interactively 'find-file-other-window filename))))
+
 
 ;; from https://www.emacswiki.org/emacs/CopyingWholeLines
 (defun spacemacs/duplicate-line-or-region (&optional n)
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -3529,6 +3529,9 @@ files (thanks to Daniel Nicolai)
 - Defer git-gutter loading (thanks to Aaron Jensen)
 - Remove redundant arg from git-gutter timer (thanks to bmag)
 - Added Git Blame Transient State (thanks to duianto)
+**** Spacemacs-defaults
+- Updated spacemacs/save-as function, see github #8974 and #14754 (thanks to
+  Daniel Nicolai and Lebensterben)
 **** Spotify
 - Replaced helm-spotify package with helm-spotify-plus
   (thanks to Leonard Lausen)


layers/+spacemacs/spacemacs-defaults/funcs.el
```diff
@@ -1350,10 +1350,43 @@ the right."
...
+    ;; non-visiting file buffers
+    (when (directory-name-p new-file-name)
+      (if buffer-file-name
+          (setq new-file-name (concat new-file-name (buffer-name)))
+        (error "Copying of non-visiting file buffers requires to specify new file name.")))
+    (when (and (file-exists-p new-file-name)
+               (not (y-or-n-p (format "File %s exists. Overwrite?: " new-file-name))))
+      (error "Canceled"))
+    (pcase
+        (read-answer "New file action: "
```
lebensterben, line 1376: In general avoid string comparison. This is inefficient.

You can just match the character code.

If you really want, match symbol, e.g. 'only-write.

layers/+spacemacs/spacemacs-defaults/funcs.el
```diff
@@ -1350,10 +1350,43 @@ the right."
...
+help menu that explains the different options."
+  (interactive "P")
+  (let* ((new-file-name (read-file-name "Save file as: "
+                                        (if arg buffer-file-name default-directory))))
+    ;; when selecting a directory name, add (buffer-name) or message error for
+    ;; non-visiting file buffers
+    (when (directory-name-p new-file-name)
+      (if buffer-file-name
+          (setq new-file-name (concat new-file-name (buffer-name)))
+        (error "Copying of non-visiting file buffers requires to specify new file name.")))
```
lebensterben, line 1371: "non-visiting file" is not easy to understand

layers/+spacemacs/spacemacs-defaults/funcs.el
```diff
@@ -1350,10 +1350,43 @@ the right."
...
+      (error "Canceled"))
+    (pcase
+        (read-answer "New file action: "
+                     '(("replace"    ?r "open new file in current buffer")
+                       ("new-buffer" ?n "open new file in new buffer")
+                       ("only-write" ?o "write new-file only. Don't open")
+                       ("quit" ?q "exit")))
+      ("quit" (print "Canceled"))
+      (code (let ((dir (file-name-directory new-file-name)))
+              (when (not (file-exists-p dir))
```
lebensterben, line 1383: unless

layers/+spacemacs/spacemacs-defaults/funcs.el
```diff
@@ -1350,10 +1350,43 @@ the right."
...
+                (make-directory dir t)))
+            (if (string= "replace" code)
+                (write-file new-file-name))
+            (write-region (point-min) (point-max) new-file-name)
```
lebensterben, line 1387: write-region after write-file?

not indented correctly

lebensterben (changes requested):
Also, replace error by user-error. These errors should not start the debugger.

layers/+spacemacs/spacemacs-defaults/funcs.el
dalanicolai: Thanks for this info, handy to be aware of. But in this case the code only runs once only when this command is called, so I guess the efficiency is not so important in this case.

So otherwise you would suggest to compare only the first (or other relevant) characters or otherwise compare with `intern`?

layers/+spacemacs/spacemacs-defaults/funcs.el
dalanicolai: I agree somewhat, although I think it is hard to misunderstand this (of course it is "non-visiting file buffer").
I would be happy if you could suggest a better term (English is not my native language).

layers/+spacemacs/spacemacs-defaults/funcs.el
dalanicolai: Good point. Interesting how I miss these things regularly...

layers/+spacemacs/spacemacs-defaults/funcs.el
dalanicolai: Good catch! And thanks for your review :+1:

layers/+spacemacs/spacemacs-defaults/funcs.el
dalanicolai: Maybe "non-visiting-file buffers"?

layers/+spacemacs/spacemacs-defaults/funcs.el
dalanicolai: Ah yes, I did not know about `user-error`. Indeed I used a catch-throw before, but I thought `error` looks nicer (and then accept the "wrong" behavior). Very useful info!

lebensterben: > So otherwise you would suggest to compare only the first (or other relevant) characters, or alternatively compare by converting to symbol with `intern`?

See https://www.gnu.org/software/emacs/manual/html_node/elisp/Multiple-Queries.html

- wrap this in a let form to set read-answer-short to t. 
- Change the long answer to a single character string. the long answer is not shown but is returned. Since it has length 1, it's fast to compare.
- Since read-answer-short is t, user only need to type one key to select a choice, instead of being required to type a sentence/phrase

lebensterben:
key-bindings for the multiple choice can be improved.

If this command is invoked via SPC f c, we can put the most possible choice under the key c.
So the user just quickly type SPC f c c to perform the default behavior.

I suggest
- c Copy and open in new buffer
- C Copy and open on current buffer
- d Don't open the new file after copying
- q Quit

dalanicolai:
I agree with you @lebensterben that generally, it is nice to put the default bindings on the quickest way to type. However, in this case after typing `SPC f c` the user first has to enter a new filename and a RET. So it is not very important here that the default action is under `c`.
Also, I always prefer downcase letters, so I would vote against using `C`.

On the other hand, I do agree that my suggestions (r, n and o) might not be the best choices. But I would suggest we choose wither the most sensible mnemonic or most consistent with other commands keybindings (which should generally most sensible mnemonic also).

So another main candidate for me was:
`c` for open in current buffer
`o` for open in other buffer
 `w` for just write (or 's' for just save)

lebensterben:
okay. c o w looks a good choice

dalanicolai:
Updated! The reason why I chose "replace" over "current-buffer" is that the former "more explicitly" implies that the old-file buffer is getting killed (or well... replaced). But I guess this `cow` will also produce our milk...

layers/+spacemacs/spacemacs-defaults/funcs.el
lebensterben: this feels like romance language, not english

layers/+spacemacs/spacemacs-defaults/funcs.el
```diff
@@ -1350,10 +1350,43 @@ the right."
...
-(defun spacemacs/copy-file ()
-  "Write the file under new name."
-  (interactive)
-  (call-interactively 'write-file))
+(defun spacemacs/copy-file (arg)
+  "Write the file under new name and select action on new file.
+This command prompts for a new filename then prompts for
+selecting an action to perform on the new file.
+
+If prefixed with the universal ARG \\[universal-argument] the
```
lebensterben, line 1358: Can you remind me why we need this prefix argument?
```elisp
(read-file-name "Save file as: " (if arg buffer-file-name default-directory))
```

Can we just write
```elisp
(read-file-name "Save file as: " (or buffer-file-name default-directory))
```

lebensterben:
Use this
```elisp
(defun spacemacs/save-as (filename &optional visit)
  "Save current buffer or active region as specified file.
When called interactively, it first prompts for FILENAME, and then asks
 whether to VISIT it, and if so, whether to show it in current window or
 another window.

FILENAME  a non-empty string as the name of the saved file.
VISIT     When it's `:current', open FILENAME in current window. When it's
          `:other', open FILENAME in another window. When it's nil, only
          save to FILENAME but does not visit it. (Default to `:current'
          when called from a LISP program.)

When FILENAME already exists, it also asks the user whether to overwrite it."
  (interactive (let* ((filename (expand-file-name (read-file-name "Save buffer as: ")))
                      (choices  '("Current window"
                                  "Other window"
                                  "Don't open"))
                      (actions  '(:current :other nil))
                      (visit    (let ((completion-ignore-case t))
                                  (nth (cl-position
                                        (completing-read "Do you want to open the file? "
                                                         choices nil t)
                                        choices
                                        :test #'equal)
                                       actions))))
                 (list filename visit)))
  (unless (called-interactively-p 'any)
    (cl-assert (and (stringp filename)
                    (not (string-empty-p filename))
                    (not (directory-name-p filename)))
              t "Expect an non-empty filepath, found: %s")
    (setq filename (expand-file-name filename)
          visit (or visit :other))
    (let ((choices '(:current :other nil)))
      (cl-assert (memq visit choices)
                 t "Found %s, expect one of %s")))
  (let ((dir (file-name-directory filename)))
    (unless (file-directory-p dir)
      (make-directory dir t)))
  (if (use-region-p)
      (write-region (region-beginning) (region-end) filename nil nil nil t)
    (write-region nil nil filename nil nil nil t))
  (pcase visit
    (:current (find-file filename))
    (:other   (funcall-interactively 'find-file-other-window filename))))
```

layers/+spacemacs/spacemacs-defaults/funcs.el
dalanicolai: This argument is to include the filename in the prompt for new filename like in the [https://github.com/syl20bnr/spacemacs/pull/8974#issuecomment-755299650](original proposal) (but then inverted). The `or` wouldn't work here, at least not for how I meant it. On the other hand `(buffer-name)` would be even better than buffer-file-name.

dalanicolai:
Haha! That's funny... it is a full rewrite. Anyway, that is fine with me. However, this doesn't yet include an option to include the (current) filename in the prompt.
Also it is not possible to select a "choice" with a single keystroke, but that is a matter of preference of course (I like this solution too).

lebensterben:
@dalanicolai 
That's a bug of helm.
If you want, `(read-file-name "Save buffer as: " nil nil nil buffer-file-name)`

lebensterben:
I already found this issue. I just didn't bother to fix it.

lebensterben:
it's possible to select with two keystroke. the initial and <enter>.

lebensterben:
Also note that I've edited the code snippet. It's certainly not final.

layers/+spacemacs/spacemacs-defaults/funcs.el
dalanicolai: Yes I was aware of this, but then I would probably have wrapped it in `read-answer-short 'auto` (and indeed make the long answers slightly shorter, but still a little more informative)

lebensterben (changes requested):
I'm going to merge this. Just fix the docstring.
Also please update CHANGELOG.develop

dalanicolai:
I got a little confused about 5d213ce. It caused this PR to consist of 2 commits, and I think it was not meant like that. Anyway, I reverted to a single commit and rebased the PR on develop.

lebensterben:
https://github.com/syl20bnr/spacemacs/pull/14754/files
This is what the changes will be like when it's squashed and merged.

dalanicolai:
You mean it is fine to have multiple commits because you will squash them when merging?

lebensterben:
yes

layers/+spacemacs/spacemacs-defaults/funcs.el
```diff
@@ -1350,10 +1350,55 @@ the right."
   (interactive)
   (set-buffer-file-coding-system 'undecided-dos nil))
 
-(defun spacemacs/copy-file ()
-  "Write the file under new name."
-  (interactive)
-  (call-interactively 'write-file))
+(defun spacemacs/save-as (filename &optional visit)
+  "Save current buffer or active region as specified file.
+ When called interactively, it first prompts for FILENAME, and then asks
```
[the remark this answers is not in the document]
dalanicolai: Done

lebensterben (approved):

lebensterben:

bcc32:
I think this key binding needs to be updated, since the name of the function was changed:

```elisp
;; file -----------------------------------------------------------------------
(spacemacs|spacebind
 "Files manipulation."
 :global
 (("f" "Files"
   ("A" spacemacs/find-file-and-replace-buffer "Set another file for buffer...")
   ("c" spacemacs/copy-file "Copy file to new file...")
```

PR #14758, opened 2021-05-10
Evilify geiser-doc-mode-map and add keybindings in scheme layer

dalanicolai:
In the scheme REPL, RET is bound to `geiser-repl--maybe-send` which most times
sends the input instead of inserting a newline. Therefore this separate
keybinding for that will always insert the newline is required.

Also the `geiser-doc-mode` is read-only and should get evilified to not shadow its default keybindings.

### DIFF
--- a/layers/+lang/scheme/README.org
+++ b/layers/+lang/scheme/README.org
@@ -16,7 +16,8 @@
   - [[#macroexpansion][Macroexpansion]]
   - [[#repl-interaction][REPL interaction]]
   - [[#evaluation][Evaluation]]
-  - [[#repl][REPL]]
+  - [[#repl-mode][REPL-mode]]
+  - [[#geiser-doc-mode][Geiser-doc-mode]]
 
 * Description
 This layer adds support for Scheme via [[http://geiser.nongnu.org][Geiser]]. Note that combined usage of racket-mode and geiser has not been tested.
@@ -129,7 +130,7 @@ When enabled the symbol =🆂= should be displayed in the mode-line.
 
 | Key binding | Description         |
 |-------------+---------------------|
-| ~SPC m g g~ | Goto Definition     |
+| ~SPC m g d~ | Goto Definition     |
 | ~SPC m g b~ | Go Back             |
 | ~SPC m g m~ | Goto Module         |
 | ~SPC m g n~ | Goto next error     |
@@ -183,15 +184,38 @@ When enabled the symbol =🆂= should be displayed in the mode-line.
 | ~SPC m e l~ | Evaluate line             |
 | ~SPC m e r~ | Evaluate region           |
 
-** REPL
+** REPL-mode
 
-| Key binding | Description                 |
-|-------------+-----------------------------|
-| ~g j~       | geiser-repl-next-prompt     |
-| ~g k~       | geiser-repl-previous-prompt |
-| ~SPC m h~   | geiser-doc-symbol-at-point  |
-| ~SPC m i~   | geiser-repl-interrupt       |
-| ~SPC m c~   | geiser-repl-clear-buffer    |
-| ~SPC m m~   | geiser-repl-import-module   |
-| ~SPC m u~   | geiser-repl-unload-function |
-| ~SPC m q~   | geiser-repl-exit            |
+*Insert state*
+| Key binding | Description    |
+|-------------+----------------|
+| ~S-RET~     | Insert newline |
+| ~C-l~       | Clear buffer   |
+| ~C-d~       | Exit           |
+
+*Normal state*
+| Key binding  | Description          |
+|--------------+----------------------|
+| ~g j~ / ~]]~ | Goto next prompt     |
+| ~g k~ / ~[[~ | Goto previous prompt |
+
+| ~SPC m h h~ | Show documentation for symbol at point      |
+| ~SPC m C~   | Clear buffer                                |
+| ~SPC m i l~ | Insert lambda                               |
+| ~SPC m i m~ | Import module                               |
+| ~SPC m u~   | Unload function                             |
+| ~SPC m s~   | Toggle surrounding parenthesis <-> brackets |
+| ~SPC m k~   | REPL interrupt                              |
+| ~SPC m q~   | REPL exit                                   |
+
+** Geiser-doc-mode
+| Key binding    | Description           |
+|----------------+-----------------------|
+| ~o~            | Follow link           |
+| ~]]/[[~        | Next/previous-section |
+| ~g p~ / ~<~    | Previous page         |
+| ~g n~ / ~>~    | Next page             |
+| ~g d~          | Goto definition       |
+| ~g z~          | Switch to repl        |
+| ~TAB~ / ~C-j~  | Next button           |
+| S-TAB~ / ~C-k~ | Previous button       |
--- a/layers/+lang/scheme/packages.el
+++ b/layers/+lang/scheme/packages.el
@@ -75,7 +75,6 @@
         "el" 'lisp-state-eval-sexp-end-of-line
         "er" 'geiser-eval-region
 
-        "gb" 'geiser-pop-symbol-stack
         "gm" 'geiser-edit-module
         "gn" 'next-error
         "gN" 'previous-error
@@ -90,7 +89,7 @@
 
         "me" 'geiser-expand-last-sexp
         "mf" 'geiser-expand-definition
-        "mx" 'geiser-expand-region
+        "mr" 'geiser-expand-region
 
         "si" 'geiser-mode-switch-to-repl
         "sb" 'geiser-eval-buffer
@@ -102,17 +101,46 @@
         "sR" 'geiser-eval-region-and-go
         "ss" 'geiser-set-scheme)
 
-      (evil-define-key 'normal 'geiser-repl-mode-map
-      "gj" 'geiser-repl-next-prompt
-      "gk" 'geiser-repl-previous-prompt)
+      (evil-define-key 'insert 'geiser-repl-mode-map
+        (kbd "S-<return>") 'geiser-repl--newline-and-indent
+        (kbd "C-l") 'geiser-repl-clear-buffer
+        (kbd "C-d") 'geiser-repl-exit)
 
+      (evil-define-key 'normal 'geiser-repl-mode-map
+        "]]" 'geiser-repl-next-prompt
+        "[[" 'geiser-repl-previous-prompt
+        "gj" 'geiser-repl-next-prompt
+        "gk" 'geiser-repl-previous-prompt)
+
+      (spacemacs/declare-prefix-for-mode 'geiser-repl-mode "mh" "help")
+      (spacemacs/declare-prefix-for-mode 'geiser-repl-mode "mi" "insert")
       (spacemacs/set-leader-keys-for-major-mode 'geiser-repl-mode
-        "q" 'geiser-repl-exit
-        "c" 'geiser-repl-clear-buffer
-        "i" 'geiser-repl-interrupt
-        "m" 'geiser-repl-import-module
+        "C" 'geiser-repl-clear-buffer
+        "k" 'geiser-repl-interrupt
+        "f" 'geiser-load-file
+        "il" 'geiser-insert-lambda
+        "im" 'geiser-repl-import-module
         "u" 'geiser-repl-unload-function
-        "h" 'geiser-doc-symbol-at-point))))
+        "hh" 'geiser-doc-symbol-at-point
+        "s" 'geiser-squarify
+        "q" 'geiser-repl-exit)
+
+      (evilified-state-evilify-map geiser-doc-mode-map
+        :mode geiser-doc-mode
+        :eval-after-load geiser-doc
+        :bindings
+        "o" 'link-hint-open-link
+
+        "]]" 'geiser-doc-next-section
+        "[[" 'geiser-doc-previous-section
+        ">" 'geiser-doc-next
+        "<" 'geiser-doc-previous
+
+        "gp" 'geiser-doc-previous
+        "gn" 'geiser-doc-next
+        "gz" 'geiser-doc-switch-to-repl
+        (kbd "C-j") 'forward-button
+        (kbd "C-k") 'backward-button))))
 
 (defun scheme/post-init-ggtags ()
   (add-hook 'scheme-mode-local-vars-hook #'spacemacs/ggtags-mode-enable))


dalanicolai:
@lebensterben RET is already bound to such function (but so that function does not work very well). Anyway, I could change the new binding from `C-RET` to `S-RET` but I am not sure. Let us discuss a little...

I agree that the Ctrl should be on the inside of the Fn (as is the default on my Lenovo keyboard. I wonder how much thought/testing manufacturers that place the Ctrl on the outside put in that design). But usually those buttons can be swapped very easily via the bios or the operating system (which we should probably recommend to users).

So for me, pressing Shift or Ctrl really makes no difference and I would more think about consistency with other modes.
E.g. in the customize menu's we use `C-j` to add a newline (so that is Ctrl`). And I think (not sure) in most repl/notebook interfaces (ipython/jupyter/Mathematica) `S-RET` is used to send to REPL.

You have any more "references"? What should we decide?

lebensterben:
> And I think (not sure) in most repl/notebook interfaces (ipython/jupyter/Mathematica) `S-RET` is used to send to REPL.

No. For vast majority of applications S-enter is to start a new line.
https://defkey.com/what-means/shift-enter

While Command-enter or Control-enter usually do something completely different
https://defkey.com/what-means/control-enter
https://defkey.com/what-means/command-enter

dalanicolai:
Ah that's funny, I never worked with any of those programs, except with Jupyter and Spyder (and Mathematica which is not in the list), and indeed they send to REPL, but now I see those are exception (although maybe for a good reason).
Also I see in your list that `C-RET` also is mostly used to insert a line break, so I would not call that completely different (`S-RET` is as much used to do completely different things).

So I still think the choice is quite arbitrary. But I have no problem to swap them if you like (and break with Mathematica and Jupyter behavior)

dalanicolai:
@lebensterben I actually tried them first (I think I took the `g j` and `g k` from there), but then for some reason I preferred to add bindings myself. But I do not remember why I preferred that, because now those binding look quite acceptable.

However, those bindings do not add much to what we have already. And for sure I prefer to use evilified-state for read-only buffers. So we could selectively adopt the few useful extra keybindings they have. But until now I did not miss any of those bindings (still might be nice to add them). I will have a quick look again tomorrow, and maybe adopt some of them.

dalanicolai:
I've added some more bindings, should be fairly complete now

layers/+lang/scheme/funcs.el
```diff
@@ -0,0 +1,4 @@
+(defun geiser-customize-implementation ()
```
lebensterben, line 1: what's this for?

layers/+lang/scheme/README.org
```diff
@@ -183,15 +184,31 @@ When enabled the symbol =🆂= should be displayed in the mode-line.
...
-| ~SPC m i~   | geiser-repl-interrupt       |
-| ~SPC m c~   | geiser-repl-clear-buffer    |
-| ~SPC m m~   | geiser-repl-import-module   |
-| ~SPC m u~   | geiser-repl-unload-function |
-| ~SPC m q~   | geiser-repl-exit            |
+** REPL-mode
+
+| Key binding | Description                      |
+|-------------+----------------------------------|
+| ~C-RET~     | geiser-repl--newline-and-indent  |
```
lebensterben, line 191: S-RET

layers/+lang/scheme/packages.el
```diff
@@ -102,17 +102,44 @@
...
+        "[[" 'geiser-repl-previous-prompt
+        "gj" 'geiser-repl-next-prompt
+        "gk" 'geiser-repl-previous-prompt)
 
+      (spacemacs/declare-prefix-for-mode 'geiser-repl-mode "mh" "help")
       (spacemacs/set-leader-keys-for-major-mode 'geiser-repl-mode
-        "q" 'geiser-repl-exit
+        "b" 'geiser-pop-symbol-stack
         "c" 'geiser-repl-clear-buffer
         "i" 'geiser-repl-interrupt
```
lebensterben, line 112: this one should be C-d.
Because, literally, ^D has the meaning of end of transmission.

And this is indeed the default key binding of most repl to signal EOT.

"i" is not really meaningful. I'd more likely to think it represents "insert".

Then insert lambda is "il", and import module is "im".

layers/+lang/scheme/packages.el
```diff
@@ -102,17 +102,44 @@
...
+        "]]" 'geiser-repl-next-prompt
+        "[[" 'geiser-repl-previous-prompt
+        "gj" 'geiser-repl-next-prompt
+        "gk" 'geiser-repl-previous-prompt)
 
+      (spacemacs/declare-prefix-for-mode 'geiser-repl-mode "mh" "help")
       (spacemacs/set-leader-keys-for-major-mode 'geiser-repl-mode
-        "q" 'geiser-repl-exit
+        "b" 'geiser-pop-symbol-stack
         "c" 'geiser-repl-clear-buffer
```
lebensterben, line 111: This can be C-l.
^l is form feed and is the default binding for clearing current screen.

the current binding is understandable enough though.

layers/+lang/scheme/packages.el
```diff
@@ -102,17 +102,44 @@
...
         "i" 'geiser-repl-interrupt
+        "f" 'geiser-load-file
+        "l" 'geiser-insert-lambda
         "m" 'geiser-repl-import-module
         "u" 'geiser-repl-unload-function
-        "h" 'geiser-doc-symbol-at-point))))
+        "hd" 'geiser-doc-symbol-at-point
+        "s" 'geiser-squarify
+        "q" 'geiser-repl-exit
+        "hc" 'geiser-customize-implementation)
```
lebensterben, line 126: I don't recall we have added any binding for customizing a particular mode.

Is this a common part of workflow?

If you really want, something in ", T" is better because that's the prefix for toggles.

layers/+lang/scheme/packages.el
```diff
@@ -102,17 +102,44 @@
...
+        "b" 'geiser-pop-symbol-stack
         "c" 'geiser-repl-clear-buffer
         "i" 'geiser-repl-interrupt
+        "f" 'geiser-load-file
+        "l" 'geiser-insert-lambda
         "m" 'geiser-repl-import-module
         "u" 'geiser-repl-unload-function
-        "h" 'geiser-doc-symbol-at-point))))
+        "hd" 'geiser-doc-symbol-at-point
+        "s" 'geiser-squarify
```
lebensterben, line 124: Is this a common part of workflow?

Some other modes may also accomplish this functionality, eg in parinfer-rust-mode if you change the opening parenthesis to bracket it changes the closing one (or vice versa).

p.s. one real issue is the command name itself is really not informative...

layers/+lang/scheme/packages.el
```diff
@@ -102,17 +102,44 @@
...
+        :eval-after-load geiser-doc
+        :bindings
+        "o" 'link-hint-open-link
+
+        "]]" 'geiser-doc-next-section
+        "[[" 'geiser-doc-previous-section
+        ">" 'geiser-doc-next
+        "<" 'geiser-doc-previous
+
+        "gb" 'geiser-doc-previous
```
lebensterben, line 139: for "gb" and "gf"

do you mean backward and forward?

normally we just use "gn" and "gp" as next and previous.

layers/+lang/scheme/packages.el
```diff
@@ -102,17 +102,44 @@
...
+        "l" 'geiser-insert-lambda
         "m" 'geiser-repl-import-module
         "u" 'geiser-repl-unload-function
-        "h" 'geiser-doc-symbol-at-point))))
+        "hd" 'geiser-doc-symbol-at-point
+        "s" 'geiser-squarify
+        "q" 'geiser-repl-exit
+        "hc" 'geiser-customize-implementation)
+
+      (evilified-state-evilify-map geiser-doc-mode-map
```
lebensterben, line 128: do you think we should also added bindings for forward-button and backward-button, as it does in evil-collections? (I've no strong preference on this one)

lebensterben (changes requested):

layers/+lang/scheme/funcs.el
dalanicolai: Ah well, as up to now I only use very few geiser functions, I just used the `geiser` menu in the menu-bar to check out the available geiser functions in the different geiser modes.
This menu has an entry to jump to the customize menu for the currently used scheme implementation.
I thought that is might be a nice service, to provide a similar function under the `SPC m h` menu, but I am just trying it out.
I am happy to hear any positive or negative opinion about it, maybe I am becoming too user friendly :laughing:

layers/+lang/scheme/packages.el
dalanicolai: Yes, nice suggestions :+1:

layers/+lang/scheme/packages.el
dalanicolai: Well generally I use `o` for this, and otherwise personally I prefer `TAB/S-TAB` which are already the default emacs (i.e. evilified) bindings. But indeed I should at least add `TAB/S-TAB` to the docs. Or do you think I should add evil-collection's `C-j/C-k` also?

layers/+lang/scheme/packages.el
dalanicolai: It [seems that the `geiser-pop-symbol-stack` is quite useless in the REPL](https://www.nongnu.org/geiser/geiser_4.html#Jumping-around), I put it here because `evil-collection` has it. But I will remove it.

layers/+lang/scheme/packages.el
dalanicolai: You mean `, T` for toggle squarify? Or to go to customize?

layers/+lang/scheme/packages.el
dalanicolai: Well it is `almost` provided by `sp-rewrap-sexp` or `evil-surround` (`c s ) ]`). But they both require you to type explicitly the new type of enclosing.
I know in racket and chez, brackets are equivalent to parenthesis (they are used interchangeably to "improve" readability. I don't know about scheme in general). So anyway, I guess this shortcut is quite handy.

layers/+lang/scheme/packages.el
lebensterben: no, I mean geiser customize implementation.

layers/+lang/scheme/packages.el
lebensterben: C-j/C-k is almost as common as TAB/S-TAB in various evil bindings.
So you can add both.

layers/+lang/scheme/packages.el
lebensterben: After thinking about this again, we probably should use more cubersome key binding than "c".

a single key is easy to press, and is easy to press by mistake.
Clearing the repl seems like a "destructive" action, so we'd better make the key binding idiot-proof.

"C-l" would be the standard binding for clearing screen, but you may use "C" ("s-c").

dalanicolai:
I have updated the keybindings (implemented suggestions) and updated the docs

layers/+lang/scheme/README.org
```diff
@@ -183,15 +184,41 @@ When enabled the symbol =🆂= should be displayed in the mode-line.
...
-| ~SPC m m~   | geiser-repl-import-module   |
-| ~SPC m u~   | geiser-repl-unload-function |
-| ~SPC m q~   | geiser-repl-exit            |
+** REPL-mode
+
+*Insert state*
+| Key binding | Description           |
+|-------------+-----------------------|
+| ~S-RET~     | Insert newline indent |
+| ~C-l~       | Clear buffer          |
```
lebensterben, line 193: ~C-l / C~ since you added both

layers/+lang/scheme/README.org
```diff
@@ -183,15 +184,41 @@ When enabled the symbol =🆂= should be displayed in the mode-line.
...
+| Key binding | Description           |
+|-------------+-----------------------|
+| ~o~         | Follow link           |
+| ~]]/[[~     | Next/previous-section |
+| ~>/<~       | Next/previous page    |
+| ~g p~       | Previous page         |
+| ~g n~       | Next page             |
+| ~g d~       | Goto definition       |
+| ~g z~       | Switch to repl        |
+| ~TAB/S-TAB~ | Next/previous button  |
```
lebensterben, line 223: It's clearer if
```
| ~TAB/C-j~   | Next button     |
| ~S-TAB/C-k~ | Previous button |
```

layers/+lang/scheme/funcs.el
lebensterben: I think this is not really useful.
If you find particular customization variables that are commonly used on a per-file/per-project bases, you can make some toggle functions and bind them under `SPC m T`.

That provides easy access to particular, concrete functionalities.

layers/+lang/scheme/packages.el
lebensterben:
```elisp
(defun geiser-edit-symbol-at-point (&optional arg)
  "Opens a new window visiting the definition of the symbol at point.
With prefix, asks for the symbol to edit.")

(defun geiser-pop-symbol-stack ()
  "Pop back to where \\[geiser-edit-symbol-at-point] was last invoked.")
```

I think both functions are not useful...
To jump to the symbol-at-point, you can do `g d` or `SPC m g d`. Then `C-o` brings you back to the original marker.

layers/+lang/scheme/README.org
```diff
@@ -183,15 +184,41 @@ When enabled the symbol =🆂= should be displayed in the mode-line.
...
+| ~g k~       | Goto previous prompt                              |
+| ~]]/[[~     | Goto next/previous prompt                         |
+
+| ~SPC m h d~ | Show documentation for symbol at point            |
+| ~SPC m h c~ | Open customization menu for scheme implementation |
+| ~SPC m C~   | Clear buffer                                      |
+| ~SPC m i l~ | Insert lambda                                     |
+| ~SPC m i m~ | Import module                                     |
+| ~SPC m u~   | Unload function                                   |
+| ~SPC m s~   | toggle surrounding parenthesis <-> brackets       |
```
lebensterben, line 209: Now I think this is a good binding.
In visual state, `s` binds to `evil-surrond-edit`, which let you change/delete surrounding charater.
Analogously `SPC m s` toggles the surround () to/from [].

lebensterben (changes requested):
Overall good. Just one issue, is `geiser-customize-implementation` really important so we give it a dedicated keybinding "SPC m h c"

If certain variables are commonly set on per-file/per-session basis, it's better to make them toggles if possible, under "SPC m T" prefix.

dalanicolai:
I think I forgot to 'submit the review'...

dalanicolai:
> Overall good. Just one issue, is `geiser-customize-implementation` really important so we give it a dedicated keybinding "SPC m h c"
> 
> If certain variables are commonly set on per-file/per-session basis, it's better to make them toggles if possible, under "SPC m T" prefix.

No it is not important, but https://github.com/syl20bnr/spacemacs/pull/14758#discussion_r635157422 is what I commented in the 'review that I forgot to submit'. Anyway, it would be no problem to remove it...

layers/+lang/scheme/README.org
```diff
@@ -157,7 +158,7 @@ When enabled the symbol =🆂= should be displayed in the mode-line.
 |-------------+------------------------------|
 | ~SPC m m e~ | Macroexpand last sexp        |
 | ~SPC m m f~ | Macroexpand surrounding sexp |
-| ~SPC m m r~ | Macroexpand region           |
+| ~SPC m m x~ | Macroexpand region           |
```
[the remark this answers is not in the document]
dalanicolai: Ah yes, strange. I think this was an accident...

layers/+lang/scheme/README.org
```diff
@@ -183,15 +184,41 @@ When enabled the symbol =🆂= should be displayed in the mode-line.
...
-| ~SPC m m~   | geiser-repl-import-module   |
-| ~SPC m u~   | geiser-repl-unload-function |
-| ~SPC m q~   | geiser-repl-exit            |
+** REPL-mode
+
+*Insert state*
+| Key binding     | Description           |
+|-----------------+-----------------------|
+| ~S-RET~         | Insert newline indent |
+| ~C-l / M-RET C~ | Clear buffer          |
```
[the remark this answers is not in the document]
dalanicolai: I indeed wanted to remove that (as mentioned in https://github.com/syl20bnr/spacemacs/pull/14758#discussion_r635149752, which I forgot to submit), but it is no typo; `M-RET` in insert mode is the equivalent of `SPC m` in normal mode

layers/+lang/scheme/README.org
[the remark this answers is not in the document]
dalanicolai: Ah it was also `x` in packages.el... then this must have been a brain shortcut :grin:

dalanicolai:
I have updated the PR. As I expect you prefer to remove the `customization shortcut` I have removed it (I have no strong opinion about it). If you like it back then I can of course simply put it back...

layers/+lang/scheme/packages.el
lebensterben: it's just not conventional. We haven't added similar bindings to other major modes.
If `SPC m h c` is a good binding for customization, it should also be applied to other major modes to maintain consistency.

It's better to open a separate PR for this. We need to be more cautious when devising a new convention.

layers/+lang/scheme/README.org
```diff
@@ -183,15 +184,40 @@ When enabled the symbol =🆂= should be displayed in the mode-line.
...
+| ~SPC m h d~ | Show documentation for symbol at point            |
+| ~SPC m C~   | Clear buffer                                      |
+| ~SPC m i l~ | Insert lambda                                     |
+| ~SPC m i m~ | Import module                                     |
+| ~SPC m u~   | Unload function                                   |
+| ~SPC m s~   | toggle surrounding parenthesis <-> brackets       |
+| ~SPC m k~   | REPL interrupt                                    |
+| ~SPC m q~   | REPL exit                                         |
+
+** Doc-mode
```
lebensterben, line 212: 'Geiser-doc-mode' is clearer.

layers/+lang/scheme/README.org
```diff
@@ -183,15 +184,40 @@ When enabled the symbol =🆂= should be displayed in the mode-line.
...
-| ~SPC m c~   | geiser-repl-clear-buffer    |
-| ~SPC m m~   | geiser-repl-import-module   |
-| ~SPC m u~   | geiser-repl-unload-function |
-| ~SPC m q~   | geiser-repl-exit            |
+** REPL-mode
+
+*Insert state*
+| Key binding | Description           |
+|-------------+-----------------------|
+| ~S-RET~     | Insert newline indent |
```
lebensterben, line 192: 'Insert newline and indent' or simply 'Insert newline'

layers/+lang/scheme/README.org
```diff
@@ -183,15 +184,40 @@ When enabled the symbol =🆂= should be displayed in the mode-line.
...
+| ~C-l~       | Clear buffer          |
+| ~C-d~       | Exit                  |
+
+*Normal state*
+| Key binding | Description                                       |
+|-------------+---------------------------------------------------|
+| ~g j~       | Goto next prompt                                  |
```
lebensterben, line 199: you may combine them
|~g j~ / ~]]~ | Go to next prompt     |
|~g k~ / ~[[~ | Go to previous prompt |

layers/+lang/scheme/README.org
```diff
@@ -183,15 +184,40 @@ When enabled the symbol =🆂= should be displayed in the mode-line.
...
+| ~g k~       | Goto previous prompt                              |
+| ~]]/[[~     | Goto next/previous prompt                         |
+
+| ~SPC m h d~ | Show documentation for symbol at point            |
```
lebensterben, line 203: if 'SPC m h h' is not already used, it's better since easier to type

layers/+lang/scheme/README.org
```diff
@@ -183,15 +184,40 @@ When enabled the symbol =🆂= should be displayed in the mode-line.
...
+| ~SPC m s~   | toggle surrounding parenthesis <-> brackets       |
+| ~SPC m k~   | REPL interrupt                                    |
+| ~SPC m q~   | REPL exit                                         |
+
+** Doc-mode
+| Key binding | Description           |
+|-------------+-----------------------|
+| ~o~         | Follow link           |
+| ~]]/[[~     | Next/previous-section |
+| ~>/<~       | Next/previous page    |
```
lebensterben, line 217:
```txt
| ~>~ / ~g n~ | Next page     |
| ~<~ / ~g p~ | Previous page |
```

layers/+lang/scheme/README.org
```diff
@@ -183,15 +184,40 @@ When enabled the symbol =🆂= should be displayed in the mode-line.
...
+| ~g p~       | Previous page         |
+| ~g n~       | Next page             |
+| ~g d~       | Goto definition       |
+| ~g z~       | Switch to repl        |
+| ~TAB/S-TAB~ | Next/previous button  |
+| ~C-j/k~     | Next/previous button  |
```
lebensterben, line 223: `~C-j/k~` is ambiguous
```txt
| ~TAB~ / ~C-j~   | Next button     |
| ~S-TAB~ / ~C-k~ | Previous button |
```

dalanicolai:
Nice feedback! I agree with all of them and I have updated the PR...

lebensterben (approved):

lebensterben:
merging now. thanks for your contributions.

PR #14759, opened 2021-05-10
[csv] Bind csv-mode keys also in tsv-mode

bcc32:
tsv-mode is a derived mode of csv-mode, and supports all of the same
commands, but didn't have any of the key bindings.

I'm not confident this is the idiomatic way of providing these key bindings for derived modes, but there wasn't anything in `core/core-keybindings.el` relating to derived modes so this seemed like the most straightforward option.

### DIFF
--- a/layers/+lang/csv/packages.el
+++ b/layers/+lang/csv/packages.el
@@ -28,20 +28,21 @@
     :defer t
     :init
     (progn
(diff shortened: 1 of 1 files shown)


bcc32:
I added an alternative PR in #14763.  Feel free to merge whichever you prefer, or let me know if there's changes you'd like me to make.

bcc32:
Closing in favor of #14763

PR #14760, opened 2021-05-10
emacs-lisp: load srefactor

agzam:
I wanted to post this as a question, but then I thought: maybe I'll make a tiny PR and ask it here.

For some reason srefactor is not being loaded, does anyone know why? It looks like the handler should be `init` and not `post-init`.

### DIFF
--- a/layers/+lang/emacs-lisp/packages.el
+++ b/layers/+lang/emacs-lisp/packages.el
@@ -317,7 +317,7 @@
   (with-eval-after-load 'semantic
     (semantic-default-elisp-setup)))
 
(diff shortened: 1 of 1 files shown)


agzam:
Ha. My bad. Looks like there's nothing wrong here. It gets enabled when `semantic` layer is added and loaded. I dunno, it feels to me there's gotta be a better way. I'm just gonna close this. Sorry for the noise.

lebensterben:
When there's pre-init or post-init but not init, this package is owned by another layer.

To use it, you need to enable that layer.

PR #14762, opened 2021-05-10
Evilify xref--xref-buffer-mode

dalanicolai:
When using `spacemacs/jump-to-definition` (`g d`) in e.g. guile source code, Spacemacs opens a xref buffer, which is a read only buffer. Generally read-only buffers should open in evilified state, and indeed here it "unshadows" many of the xref-buffer-mode's convenient default keybindings.


lebensterben (approved):

smile13241324 (approved):
Good point, thanks for fixing

PR #14763, opened 2021-05-11
[csv] Make tsv-mode inherit csv-mode's leader key bindings

bcc32:
I tried to add a sort of general way to inherit key bindings.  I would have automatically computed the parent mode but it is not known until the package is loaded, so I would have had to move the bindings to `:config`, which would have been worse (key bindings no longer autoload the package).

This is an alternative solution to #14759.

### DIFF
--- a/core/core-keybindings.el
+++ b/core/core-keybindings.el
@@ -131,6 +131,20 @@ minor-mode, the third argument should be non nil."
               :evil-states (normal motion visual evilified)))
           (boundp prefix)))))
 
+(defun spacemacs/inherit-leader-keys-from-parent-mode (mode &optional parent-mode)
+  "Make derived mode MODE inherit leader key bindings from PARENT-MODE.
+
+If omitted, PARENT-MODE defaults to the parent mode of MODE.
+Signal an error if MODE is not a derived mode."
+  (unless parent-mode
+    (setq parent-mode (or (get mode 'derived-mode-parent)
+                          (error "Mode %s has no parent" mode))))
+  (let ((map (intern (format "spacemacs-%s-map" mode)))
+        (parent-map (intern (format "spacemacs-%s-map" parent-mode))))
+    (when (and (spacemacs//init-leader-mode-map mode map)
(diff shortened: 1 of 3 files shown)


core/core-keybindings.el
```diff
@@ -131,6 +131,16 @@ minor-mode, the third argument should be non nil."
               :evil-states (normal motion visual evilified)))
           (boundp prefix)))))
 
+(defun spacemacs/inherit-leader-keys-from-parent-mode (mode)
```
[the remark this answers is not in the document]
bcc32: Ah, I misunderstood your comment.

core/core-keybindings.el
```diff
@@ -131,6 +131,20 @@ minor-mode, the third argument should be non nil."
               :evil-states (normal motion visual evilified)))
           (boundp prefix)))))
 
+(defun spacemacs/inherit-leader-keys-from-parent-mode (mode &optional parent-mode)
+  "Make derived mode MODE inherit leader key bindings from PARENT-MODE.
+
+If omitted, PARENT-MODE defaults to the parent mode of MODE.
+Signal an error if MODE is not a derived mode."
```
lebensterben, line 138: You can delete the blank line on L136. Otherwise there'll be two blank lines.

It's worth mentioning that if `mode` itself is nil, as in the case when this function is called before the package is loaded, then we cannot derive the parent mode as well.

So just add a note in docstring that people should call this function after the package is loaded.

bcc32: I added a note mentioning the package loading thing, but I'm a bit confused about your note about the blank line.  I didn't see an extra blank line in describe-function, but I also saw the same result with the blank line removed, so I went ahead and implemented your suggestion to be consistent with the rest of the file.

lebensterben: Oh sorry. There won't be two blank lines.

It just automatically adds a blank line after first sentence.
This is a tacit behavior.

lebensterben (approved):

lebensterben:
merged. Thanks for your contribution.

PR #14766, opened 2021-05-12
Not using "call-interactively" result in error

w9:
`xref-find-definitions` is an interactive function that requires one argument `identifier`. It either needs to be called with `call-interactively`, or better yet, by being forwarded the `sym-name` as the `identifier`.


lebensterben:
this entire function is problematic.

Not all arguments is used

smile13241324 (approved):
Your local change should be ok and will be merged in a minute.
I would agree to lebensterben that this function looks suspicious, it does not use `arg` and actually provides an implementation for a standard function.

PR #14714, opened 2021-04-26
SpacemacsOS: provide new layer variables for systray and autostart

be11ng:
I'm not sure how best to start doing this, but the branch over at
timor:spacemacsOS has moved ahead a bit.  I've grown used to a lot of these
changes, and would like to see them moved to the main develop branch.  But, the
pull request (formed by merging in CestDiego:spacemacs/SpacemacsOS in #3321)
started diverging.

Again, I'm not sure how to approach moving together these two changes.  Maybe
@timor can help somehow??

This at the least moves in some little things from Timor's branch:

- Can enable exwm-systemtray with `exwm-enable-systray`
- Can enable running XDG autostart applications (only in XDG_USER_HOME/autostart
  for now) with `exwm-autostart-xdg-applications`
- Support for reading `.desktop` files and XDG autostart.
- Can specify additional code appended to exwm-init hook with `exwm-custom-init`

PR #14770, opened 2021-05-13
[exwm] Remove manual XF86 keybindings

be11ng:
Recently, #14714 added `desktop-environment-mode`.  This means that we don't
need to separately define bindings like `<XF86MonBrightnessUp>`, but can instead
lean on the user to enable `desktop-environment-mode` in `init.el`.

`desktop-environment-mode` should not be loaded by default since it is
linux/unix-centric.

Looking at my previous PR, I added `desktop-environment-mode`, then never used
it.  Oops.

### DIFF
--- a/layers/+window-management/exwm/packages.el
+++ b/layers/+window-management/exwm/packages.el
@@ -53,16 +53,15 @@
       :evil-leader "TD")
 
     :config
-    (progn
-      ;; We bypass desktop-environment's locking functionality for 2 reasons:
-      ;; 1. s-l is most likely needed for window manipulation
-      ;; 2. desktop-environment's locking mechanism does not support registering as session manager
-      ;; The following line would instead assign their locking command to the default binding:
-      ;; (define-key desktop-environment-mode-map (kbd "<s-pause>") (lookup-key desktop-environment-mode-map (kbd "s-l")))
-      (setq desktop-environment-update-exwm-global-keys :prefix)
-      (define-key desktop-environment-mode-map (kbd "s-l") nil)
-      ;; If we don't enable this, exwm/switch-to-buffer-or-run won't move an X window to the current frame
-      (setq exwm-layout-show-all-buffers t))))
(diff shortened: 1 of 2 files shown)


layers/+window-management/exwm/packages.el
```diff
@@ -159,13 +158,7 @@
 
     (exwm/exwm-bind-command
      "s-'"  exwm-terminal-command
```
lebensterben, line 161: this is duplicated.
Is this intended?

be11ng: Not sure what you're asking here – if you're referring to the `exwm-bind-command`, I think that's Github displaying the two different chunks of the diff a bit weirdly.  It only actually shows up once.
Above that is a `progn` I didn't *think* we need.  Do we need it?

lebensterben: What I'm asking is there're two bindings for `exwm-terminal-command`, and
do you have both of them on purpose?

be11ng: oh I see, sorry.
This is outside of the scope of my PR, though I could remove one of them.  The `s-'` is documented in the README; `s-return` isn't.  Both make a little sense though.  `s-'` feels like a "system repl", `s-return` is more standard on other window managers.  So I could document both in the README.  Thoughts?

lebensterben: let's remove s-'.
This relies on qwerty layout. On other layout ' can be anywhere on the keyboad is not necessarily easy to type.

lebensterben:
merged. thanks for your contribution.

HanshenWang:
Be advised that calling "helm-spacemacs-help" with this commit produces 
```
edebug--read: End of file during parsing: /home/hanshen/.emacs.d/layers/+window-management/exwm/packages.el
```
In said file, I believe a missing close paren on line 64.

lebensterben:
@HanshenWang Thanks for reporting. I just pushed a fix.

PR #14771, opened 2021-05-13
[home] Show notes preview, with full note button

duianto:
Show the first 5 lines of the home buffer notes:
Quick Help and Release Notes.

Added a button to show the full note.

---

emacs_2021-05-13_16-56-08

emacs_2021-05-13_16-56-20

### DIFF
--- a/core/core-spacemacs-buffer.el
+++ b/core/core-spacemacs-buffer.el
@@ -334,6 +334,26 @@ WIDTH: current external width of the note's frame."
        (setq ,caption nil
              ,caption-length 0))))
 
+(defvar spacemacs-buffer-note-preview-lines 5
+  "If it's a positive integer, show the notes first number of lines.
+If nil, show the full note.")
+
+(defvar spacemacs-buffer--note-preview-nr-of-removed-lines nil
+  "Store the number of removed lines from the notes:
+Quick Help and Release Notes.")
+
+(defun spacemacs-buffer//if-note-preview-remove-rest-of-note ()
+  "If `spacemacs-buffer-note-preview-lines' is a positive integer,
+remove the rest of the note, after the variables line number."
+  (when (and (integerp spacemacs-buffer-note-preview-lines)
+             (> spacemacs-buffer-note-preview-lines 0))
+    (goto-line (1+ spacemacs-buffer-note-preview-lines))
+    (let* ((first-removed-line (line-number-at-pos (point)))
+           (last-removed-line (line-number-at-pos (point-max))))
+      (setq spacemacs-buffer--note-preview-nr-of-removed-lines
+            (- last-removed-line first-removed-line))
+      (delete-region (point) (point-max)))))
+
 (defun spacemacs-buffer//notes-render-framed-text
     (content &optional topcaption botcaption hpadding max-width min-width)
   "Return a formatted string framed with curved lines.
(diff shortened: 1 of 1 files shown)


core/core-spacemacs-buffer.el
```diff
@@ -334,6 +334,24 @@ WIDTH: current external width of the note's frame."
...
 
+(defvar spacemacs-buffer-note-preview-lines 5
+  "If it's a positive integer, show the notes first number of lines.
+If nil, show the full note.")
+
+(defvar spacemacs-buffer--note-preview-nr-of-removed-lines nil
+  "Store the number of removed lines from the notes:
+Quick Help and Release Notes.")
+
+(defun spacemacs-buffer//note-remove-after-line (line)
```
lebensterben, line 345: seems like this function always take spacemacs-buffer-note-preview-lines as its argument.

If so we can remove the argument.

core/core-spacemacs-buffer.el
```diff
@@ -356,9 +374,13 @@ MIN-WIDTH is the minimal width of the frame, frame included.  The frame will not
     (if (not (file-exists-p content))
         (insert content)
       (insert-file-contents content)
-      (goto-char (point-max))
-      (when (eq ?\n (char-before))    ;; remove additional newline at eof
-        (delete-char -1)))
+      (if spacemacs-buffer-note-preview-lines
+          (progn (spacemacs-buffer//note-remove-after-line
```
lebensterben, line 378: first form of progn can be moved out of if form.
Because it has no effect when spacemaca-notes-preview-lines is nil.

core/core-spacemacs-buffer.el
```diff
@@ -455,6 +476,39 @@ ADDITIONAL-WIDGETS: a function for inserting a widget under the frame."
...
+               (delete-region (point) (+ (point) (length full-note-link-text) 2))
+               (add-to-list
+                'spacemacs-buffer--note-widgets
+                (widget-create
+                 'push-button
+                 :tag (propertize
+                       full-note-link-text 'face 'font-lock-warning-face)
+                 :help-echo "Open full note."
+                 :action (lambda (&rest ignore)
+                           (let ((cursor-pos-before-showing-full-note (point)))
```
lebensterben, line 499: the let bindings can be simplified by using prog1

So instead of
(let ((var expr))
  body)
you can
(let ((var (progn1 expr body))))

prog1 runs both expr and body but the return value is expr.


Then you can combine the two let forms.

core/core-spacemacs-buffer.el
duianto: Done.

core/core-spacemacs-buffer.el
duianto: Moving the function: `spacemacs-buffer//note-remove-after-line` 
outside the condition: `(if spacemacs-buffer-note-preview-lines`
could be read as if something from the note is being removed by default.

Therefore I renamed the function to make it clear that it only removes something if a note preview is requested.
`spacemacs-buffer//if-note-preview-remove-rest-of-note`

And removed the if statement, by adding two newlines to the end of the function
`spacemacs-buffer//if-note-preview-remove-rest-of-note`

core/core-spacemacs-buffer.el
duianto: I'm not sure what the prog1's first expression should return.

So I combined the two let bindings, since it doesn't matter what the note preview variable is set to when the note is closed.

duianto:
One of the frame tests are breaking.
It didn't like the two extra newlines at the bottom of the frame.

I'll look for another solution.

smile13241324 (approved):
This looks good. I have also checked the issues mentioned by lebensterben and they seems to have been resolved by refactoring the code.

PR #14772, opened 2021-05-13
Update python layer to use pylsp (new maintained fork of pyls)

dalanicolai:
LSP warns that pyls is unmaintained and the user should install pylsp. This PR updates the python correspondingly.

If still pyls is installed instead of pylsp, then lsp shows a warning that the
user should install/update to the new fork i.e. pylsp.


lebensterben:

PR #14777, opened 2021-05-14
add key bindings for moving org table cells

emacs18:
Org mode key `,tc` was bound to `org-table-convert` command.  This is changed so that `,tc` is a prefix key to add four new key bindings for moving current org table cell left, right, up, and down.

### DIFF
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -216,6 +216,7 @@ Will work on both org-mode and any mode that accepts plain html."
       (dolist (prefix `(
                         ("mb" . "babel")
                         ("mC" . ,(org-clocks-prefix))
+                        ("mc" . "cell/convert")
                         ("md" . "dates")
                         ("me" . "export")
                         ("mf" . "feeds")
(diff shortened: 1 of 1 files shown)


lebensterben:
you need to add the prefix for which key too

emacs18:
I hope I got this right.

lebensterben:
alternatively, do you think "C" ("s-c") fits better for table convert?

Benefit is it clearly separate this from cell related commands.

---

Alternatively, is there other way to move cells?

And if swapping cells is commonly used, "SPC t c h/j/k/l" is too long to type. We'd better having something easier to type.

emacs18:
I submitted https://github.com/syl20bnr/spacemacs/pull/14778 which used `,tC` prefix key.

emacs18:
I am not aware of any org mode commands which swap contents of cells.  If there were such commands, I would use those rather than commands to move cell content from one cell to adjacent cell.

lebensterben:
@emacs18 
Actuall, `SPC m H/J/K/L` are used to move cells. So you don't need to define them again.
Check documentation of `org-shiftleft` etc.

emacs18:
@lebensterben Thanks.  Now that you mention it, I vaguely recall `org-shiftleft`.  I'll probably forget it again, but hopefully not for few months.

PR #14778, opened 2021-05-14
use ,tC prefix for org table cell movement keys

emacs18:
In org mode, ",tC" key was not used.  So set this to a prefix key for moving org table cells to an adjacent cell via the usual h, j, k, l keys.


emacs18:
This is an alternative to https://github.com/syl20bnr/spacemacs/pull/14777 which uses `,tc` rather than `,tC` prefix key.

emacs18:
As mentioned by @lebensterben we already have `SPC m H/J/K/L` key bindings for moving table cells.

PR #14781, opened 2021-05-17
[helm] add C-o keybinding to open files/buffers in other window for helm-files and helm-buffers

dschrempf:
Advantage:
- Consistent behavior with helm-files, helm-buffers and helm-bookmarks.

Remarks:
- It may be better to actually use `C-<return>`. This is what I use personally, but I thought it interferes too much with the current setup. Like so, `<return>` opens the buffer/file/bookmark in the current window and `C-<return>` in another one.

### DIFF
--- a/layers/+completion/helm/packages.el
+++ b/layers/+completion/helm/packages.el
@@ -138,12 +138,17 @@
       (with-eval-after-load 'helm-files
         (define-key helm-find-files-map
           (kbd "C-c C-e") 'spacemacs/helm-find-files-edit)
+        (define-key helm-find-files-map
+          (kbd "C-o") 'helm-ff-run-switch-other-window)
         (defun spacemacs//add-action-helm-find-files-edit ()
           (helm-add-action-to-source
            "Edit files in dired `C-c C-e'" 'spacemacs//helm-find-files-edit
            helm-source-find-files))
         (add-hook 'helm-find-files-before-init-hook
                   'spacemacs//add-action-helm-find-files-edit))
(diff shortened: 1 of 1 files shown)


lebensterben:
If `<enter>` is the default behavior that opens an item in current window, you should use `S-<enter>` for the alternative behaviour to open the item in other window.

For example, https://github.com/emacs-evil/evil-collection/blob/master/modes/helm/evil-collection-helm.el

lebensterben:
That is to say, not only should you not to use C-o for the new bindings, you'd better change all existing bindings to S-<enter> as well.

dschrempf:
Alright. Is this because the keys are defined like that in `evil-collection-helm`? I never saw the evil collection file before. In which maps do those bindings work? Do we even need this PR then?

BTW, the `C-o` binding is already defined for the bookmarks, that's why I sticked to it.

lebensterben:
> EDIT: Or did you mean to change the bookmark binding to S-Enter?

Yes.

The KEY vs S-KEY dichotomy is inherent even on typewriter keyboad. That is, if you've a pair of functionalities that are similar in nature but has slight differences in behavior, you can use KEY and S-KEY as their bindings. Just like without/with shift plus a keystroke let you input lowercase/uppercase characters.

dschrempf:
I understand but I was specifically referring to the `helm-bookmark` binding, which is and was before this PR, `C-o`. That's why I chose to go with `C-o` in the first place.

dschrempf:
See https://github.com/syl20bnr/spacemacs/pull/14785

PR #14783, opened 2021-05-17
[version-control] Show bound but unlisted VCS TS key: d

duianto:
In the VCS Transient State (`SPC g .`)
The key: `d` is bound to: `magit-ediff`
But `d` isn't shown in the VCS TS.


smile13241324 (approved):
👍

PR #14784, opened 2021-05-17
[home] Fix full release note cursor position

duianto:
problem
Opening the full release note, leaves the cursor at the
beginning of the second line after the preview.

solution
Leave the cursor at the beginning of the first line after the preview.


smile13241324 (approved):
:+1:

PR #14786, opened 2021-05-19
[+intl/chinese] fix warning for quote lambda function

sunlin7:
Fix the warning message:
`layers/+intl/chinese/packages.el: Warning: (lambda nil \.\.\.) quoted with ' rather than with #'`


lebensterben:
Merged. thanks for your contribution.

sunlin7:
Thanks

PR #14794, opened 2021-05-23
[perforce] Add p4-shelve and p4-unshelve commands with keybindings

Binary-Eater:
Shelving and unshelving changelists are operations commonly used when
choosing Perforce as a VCS solution. Package consumed by the "perforce"
layer already provides "p4-shelve" and "p4-unshelve" commands that can
be exposed by the layer. Provide keybindings for these two commands as
well.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -291,6 +291,12 @@ In org-agenda-mode
   - Add consistent keybinding ~SPC m .~ to enter transient-state (thanks to Daniel Nicolai)
   - Removed ~C-h~ binding in =org-agenda-mode= (thanks to Ag Ibragimov)
   - Rebind =org-agenda-refile= to ~r~ in transient state (thanks to Muneeb Shaikh)
+***** Perforce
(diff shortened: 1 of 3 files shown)


lebensterben:
These two bindings are not mnemonic

Binary-Eater:
Yeah, I was giving that some thought. I noticed there are some exceptions to the rule like `SPC p 4 y` maps to `p4-resolve`. The alternative I had in mind before seeing that was `SPC p 4 [` for shelving and `SPC p 4 ]` for unshelving.

Binary-Eater:
Thinking about it more, I am more inclined towards `SPC p 4 [` and `SPC p 4 ]` for the bindings related to shelving/unshelving.

lebensterben:
bindings in this layer is arbitrary and not well designed...

smile13241324 (approved):
Looks good, I also agree that `SPC p 4 [` and `SPC p 4 ]` is a good choice for now.
The layer has not been designed to follow the conventions and the bindings should
be redesigned but this should be done by someone being familiar with `perforce`.

PR #14805, opened 2021-05-28
use org and org-contrib packages rather than org-plus-contrib

emacs18:
The "contrib" portion of `org-plus-contrib` is now available as separate package named `org-contrib` within `nongnu` package archive.
This pull request replaces use of `org-plus-contrib` package with `org` and `org-contrib` packages.

I believe `org-plus-contrib` will no longer be available starting with org version 9.5, so that we need to move away from using it in the near future.


smile13241324 (approved):
:+1:

smile13241324:

PR #14815, opened 2021-05-31
Swap 'so' and 'such' in the 999 banner

Luctins:
'Such powerfull' and 'So Love' seems strange to read, swapping to 'So Powerfull' and 'Such Love' reads better.


lebensterben:
they are not strange to read.
they are downright errors

Kazark:
Aren't they deliberately dogespeak errors though? :thinking:

Luctins:
Maybe, but usually "Such" is paired with an substantive-like term.

richarddewit:
image

It's deliberate. It's from the Doge meme (it even says DOGEMACS). Revert it please :)

Luctins:
Yeah, I know the original inspiration. 
These 2 sentences were the only two that seemed 'strange' in comparison to the rest; But you do have a point, in the original meme It seems like the verbal agreement is off on purpose.

Luctins:
@richarddewit you mean like [this](https://github.com/syl20bnr/spacemacs/commit/734eb9dc7b1e5da48c0f6c54466886a434ba5bee#diff-9ec1c9ddcf294361fc64226f595c30f8c0ce675912d71c7e74c480601063ad0f) ?

richarddewit:
@Luctins wow 👍🏻

PR #14819, opened 2021-06-01
[transmission] Add new keybindings

arifer612:
This commit adds 3 sets of keybindings and organises the keybindings so that it can be read easier.

1. Marking and labelling
   =SPC m t m= marks a torrent and =SPC m s l= sets the labels for marked torrents. The old command bound to =SPC m s l=, =transmission-set-ratio=, is bound to =SPC m s r= instead.
2. Queueing
   Keybindings under a new prefix are added to ~transmission-mode~ to manipulate how the torrents are queued. A keybinding to sort logically by columns is also added.
3. Turtle mode
   A new keybinding while in =transmission-mode= toggles the =transmission-turtle-mode= minor mode. When it is active, a new set of 3 keybindings will be active to change the settings of the minor mode. The bindings used are uppercase letters since there are already defined bindings using the lowercases.

Although there is a keybinding for =transmission-remove= in =transmission-mode=, a new keybinding for =transmission-delete= is also added. This is done since =transmission-remove= merely removes the torrent from Transmission but does not delete the file(s).

The changes are added to =CHANGELOG.develop=

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -350,6 +350,17 @@ In org-agenda-mode
   - Changed ~SPC e e~ to ~SPC e b~ for =flycheck-buffer=
 ***** Tabs
 - Add auto hide tabs feature (thanks to Daniel Nicolai)
+***** Transmission
+- Key bindings (thanks to Arif Er):
+  - Organised and corrected in both =packages.el= and =README.org=.
+  - Change ~SPC m s l~ to ~SPC m s r~ for =transmission-set-ratio= and =transmission-set-torrent-ratio=.
+  - Change ~SPC m r~ to ~SPC m R~ for =transmission-remove=.
+  - Change ~SPC m m~ to ~SPC m r~ for =transmission-move=.
+  - Change ~SPC m r~ to ~SPC m X~ for =transmission-files-command=.
+  - Swap ~SPC m a a~ and ~SPC m a t~ for =transmission-add= and =transmission-trackers-add= respectively.
+  - Change ~SPC m s b~ to ~SPC m s p~ for =transmission-set-bandwidth-priority=.
+  - Change ~SPC m g t~ to ~SPC m g f~ for =transmission-files=.
+  - Change ~SPC m m u~ and ~SPC m m w~ to ~SPC m t u~ and ~SPC m t w~ for =transmission-files-unwant= and =trasnmission-files-want= respectively
 ***** Vagrant
 - Key bindings:
   - Vagrant key bindings prefix is now ~SPC a V~.
@@ -3616,6 +3627,18 @@ files (thanks to Daniel Nicolai)
 - Prevent =tmux-command= at GUI mode (thanks to Isaac Zeng)
 - Fixed regression in tmux by setting keybindings using =use-package=
   (thanks to Aaron Jensen)
+**** Transmission
+- Key bindings (thanks to Arif Er):
+  - ~H/L~ is to move torrent to the top/bottom of the queue.
+  - ~J/K~ is to move torrent down/up the queue.
+  - ~SPC m D~ is to remove and delete torrent at point.
+  - ~SPC m m~ is to mark torrent(s) at point or in region.
+  - ~SPC m S~ is to sort the torrents according to the column at point.
+  - ~SPC m s l~ is to set label(s) for marked torrent(s).
+  - ~SPC m t t~ is to toggle =transmission-turtle-mode=.
+  - ~SPC m s D~ is to set days for =transmission-turtle-mode= to be active.
+  - ~SPC m s S~ is to set speed limits for =transmission-turtle-mode=.
+  - ~SPC m s T~ is to set time range for =transmission-turtle-mode to be active.
 **** Treemacs
 - =Treemacs= replaces =NeoTree= as the default sidebar
 - Added missing ~SPC p t~ to =readme.org= (thanks to oo6)
--- a/layers/+tools/transmission/README.org
+++ b/layers/+tools/transmission/README.org
@@ -16,6 +16,7 @@
   - [[#transmission-info-mode][=transmission-info-mode=]]
   - [[#transmission-files-mode][=transmission-files-mode=]]
   - [[#transmission-peers-mode][=transmission-peers-mode=]]
+  - [[#transmission-turtle-mode][=transmission-turtle-mode=]]
 
 * Description
 This layer integrates a BitTorrent client into Spacemacs.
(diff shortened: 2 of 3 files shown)


lebensterben:
I suggest to use bindings defines here https://github.com/emacs-evil/evil-collection/blob/master/modes/transmission/evil-collection-transmission.el

arifer612:
The bindings that I have added are not in evil-collection either. Things like queue ordering, labelling, and turtle mode, I added them because I see them as fundamental API used in managing active torrents in Transmission (except perhaps =transmission-delete= but I added that just for completion).

lebensterben:
I'm not saying you shouldn't add anything not included in evil-collections already.

I suggest you to use the same bindings as in evil-collections, if a function is defined both in evil-collections and in your PR.

And also it's best to avoid conflicts with evil-collections.

CHANGELOG.develop
```diff
@@ -3615,6 +3619,21 @@ files (thanks to Daniel Nicolai)
 - Prevent =tmux-command= at GUI mode (thanks to Isaac Zeng)
 - Fixed regression in tmux by setting keybindings using =use-package=
   (thanks to Aaron Jensen)
+**** Transmission
+- Key bindings (thanks to Arif Er):
+  - ~SPC m d~ is to remove and delete torrent at point.
```
lebensterben, line 3624: this for example is strictly worse than evil-collections'.

removal is a destructive operation and is better bounded to D (S-d)

arifer612:
I see, I'll make the appropriate corrections then!
So for the example you pointed out, the remove-and-delete should be bound to `SPC m D` instead, is that right?

arifer612:
I've changed the bindings for `transmission-delete` and `transmission-remove` to use the uppercase letters since they are both destructive commands. The other prior bindings differ from the bindings used by evil-collection but they can be intuitively understood with the current bindings and are not destructive.

layers/+tools/transmission/packages.el
```diff
@@ -29,50 +29,94 @@
...
               "m"  'transmission-move
-              "r"  'transmission-remove
-              "ts" 'transmission-toggle
+              "q"  'transmission-quit
+              "R"  'transmission-remove
+              "v"  'transmission-verify
+              "S"  'tabulated-list-sort
+
+              ;; add
               "aa" 'transmission-trackers-add
```
lebensterben, line 47: "aa" and "at" are confusing to me..

Why not `aa` for `transmission-add` and `at` for `transmission-trackers-add`?

layers/+tools/transmission/packages.el
```diff
@@ -29,50 +29,94 @@
...
+              "at" 'transmission-add
+
+              ;; goto
+              "gi" 'transmission-info
+              "gp" 'transmission-peers
+              "gt" 'transmission-files
```
lebensterben, line 53: what does `t` in `gt` mean?

layers/+tools/transmission/packages.el
```diff
@@ -29,50 +29,94 @@
...
+
+              ;; set
+              "sb" 'transmission-set-bandwidth-priority
```
lebensterben, line 56: should this be "sp" as for set priority?
bandwidth cannot be set.

layers/+tools/transmission/packages.el
```diff
@@ -29,50 +29,94 @@
...
-              "v"  'transmission-verify
-              "q"  'transmission-quit
-              "sb" 'transmission-set-bandwidth-priority)
+
+              ;; toggle
+              "tm" 'transmission-toggle-mark
+              "ts" 'transmission-toggle
+              "tt" 'transmission-turtle-mode
+
+              ;; Queue
```
lebensterben, line 67: How about
J/K move down/move up
H/L move to top/move to bottom

lebensterben (changes requested):

layers/+tools/transmission/packages.el
arifer612: To be honest, I agree with you here. These bindings were already present from before and I believe the author meant for it to match the binding for `transmission-tracker-add` in `transmission-info-mode`. It would make more sense to change them right?

layers/+tools/transmission/packages.el
arifer612: This binding was also brought over as it was from before so I do not know the reason for it. I can change it to `gf` for `[g]o to [f]iles`

layers/+tools/transmission/packages.el
arifer612: This was also brought over as it was from before. I failed to spot that it does not match with the binding in `transmission-files-mode` where it is bound to `sp`. This should be changed.

layers/+tools/transmission/packages.el
arifer612: I thought of doing this but I was unsure if it would okay for me to use the movement keys here. If we're using HJKL should it be nested under a prefix like "SPC m Q H" "SPC m Q J" ... or would "SPC m H" "SPC m J" ... be enough?

arifer612:
Made the necessary corrections as recommended!

layers/+tools/transmission/packages.el
lebensterben: well, putting them under `SPC m` is still very cumbersome.
How about define them as evil normal state keys?
`(evil-define-key 'normal transmission-mode-map ..)`

lebensterben: Also, please move this entire chunk except `(spacemacs/set-leader-keys "att" 'transmission)` to config, so that they're executed only when `transmission` is loaded.

arifer612: I will make the changes soon.

arifer612: I've moved everything but the binding for "att" to `:config` and changed the queue manipulation keys to use HJKL. Please check if they're in order.

arifer612:
In the latest squash, I've moved everything but the binding for "att" to :config and changed the queue manipulation keys to use HJKL. Please check if they're in order.

layers/+tools/transmission/README.org
```diff
@@ -63,52 +64,68 @@ Add =(transmission :variables transmission-auto-refresh-all t)= to
...
-| ~SPC m r~   | Prompt to remove torrent at point or torrents in region.      |
-| ~SPC m t s~ | Toggle torrent between started and stopped.                   |
-| ~SPC m a a~ | Add announce URLs to torrent or torrents.                     |
-| ~SPC m s u~ | Set global upload speed limit in kB/s.                        |
-| ~SPC m v~   | Verify torrent at point or in region.                         |
-| ~SPC m q~   | Quit and bury the buffer.                                     |
-| ~SPC m s b~ | Set bandwidth priority of torrent(s) at point or in region.   |
+| Key binding | Description                                                       |
+|-------------+-------------------------------------------------------------------|
+| ~H~         | Move torrent to the top of the queue.                             |
```
lebensterben, line 69: You can group J/K and H/L so it's clearer

layers/+tools/transmission/packages.el
```diff
@@ -27,56 +27,99 @@
...
+      (spacemacs/set-leader-keys-for-major-mode 'transmission-mode
+        "D"  'transmission-delete
+        "m"  'transmission-move
+        "q"  'transmission-quit
+        "R"  'transmission-remove
+        "v"  'transmission-verify
+        "S"  'tabulated-list-sort
+
+        ;; add
+        "at" 'transmission-trackers-add
```
lebensterben, line 46: since "A" is not used, how about
"a" transmission-add
"A" transmission-trakers-add

layers/+tools/transmission/README.org
```diff
@@ -63,52 +64,68 @@ Add =(transmission :variables transmission-auto-refresh-all t)= to
...
+| ~J~         | Move torrent down the queue by one.                               |
+| ~K~         | Move torrent up the queue by one.                                 |
+| ~L~         | Move torrent to the bottom of the queue.                          |
+| ~SPC m m~   | Move torrent at point or in region to a new location.             |
+| ~SPC m q~   | Quit and bury the buffer.                                         |
+| ~SPC m v~   | Verify torrent at point or in region.                             |
+| ~SPC m D~   | Prompt to remove and delete torrent at poit or torrens in region. |
```
lebensterben, line 76: description of remove and delete needs to be improved
the difference between them is `transmission-remove` only remove the torrent by default, while `transmission-delete` in addition remove local data. (`transmission-remove` prompts whether to remove local data)

See https://github.com/transmission/transmission/blob/8566df069899ce8923463cadeb0ff66d4544991a/extras/rpc-spec.txt#L439

layers/+tools/transmission/README.org
```diff
@@ -63,52 +64,68 @@ Add =(transmission :variables transmission-auto-refresh-all t)= to
...
+| ~SPC m R~   | Prompt to remove torrent at point or torrents in region.          |
+| ~SPC m S~   | Sort torrents according to the column at point.                   |
+| ~SPC m a a~ | Add torrent by filename, URL, magnet link, or info hash.          |
```
lebensterben, line 79: above each prefix you can add a heading, for example
```txt
| ~SPC m S~   | Sort torrents according to the column at point.                   |
|-------------+-------------------------------------------------------------------|
| Add         |                                                                   |
|-------------+-------------------------------------------------------------------|
| ~SPC m a a~ | Add torrent by filename, URL, magnet link, or info hash.          |
```

layers/+tools/transmission/README.org
```diff
@@ -63,52 +64,68 @@ Add =(transmission :variables transmission-auto-refresh-all t)= to
...
+| ~SPC m a t~ | Add announce URLs to torrent or torrents.                         |
+| ~SPC m g i~ | Open a =transmission-info-mode= buffer for torrent at point.      |
```
lebensterben, line 81: for the group, `SPC m <i | f | p>`
the default bindings are
- `transmission-info`:  `i`
- `transmission-files`: `<RET>` or `!`
- `transmission-peers`: `e`

layers/+tools/transmission/README.org
```diff
@@ -63,52 +64,68 @@ Add =(transmission :variables transmission-auto-refresh-all t)= to
...
+| ~SPC m g f~ | Open a =transmission-files-mode= buffer for torrent at point.     |
+| ~SPC m g p~ | Open a =transmission-peers-mode= buffer for torrent at point.     |
+| ~SPC m s p~ | Set bandwidth priority of torrent(s) at point or in region.       |
+| ~SPC m s d~ | Set global download speed limit in kB/s.                          |
+| ~SPC m s l~ | Set label(s) for marked torrent(s) or torrent at point.           |
+| ~SPC m s r~ | Set global seed ratio limit.                                      |
+| ~SPC m s u~ | Set global upload speed limit in kB/s.                            |
+| ~SPC m t m~ | Toggle mark on torrent(s) at point or in region.                  |
```
lebensterben, line 89: I think we can make toggle mark `SPC m m`, whose default binding is `m`

layers/+tools/transmission/README.org
```diff
@@ -63,52 +64,68 @@ Add =(transmission :variables transmission-auto-refresh-all t)= to
...
+| ~SPC m t s~ | Toggle torrent between started and stopped.                       |
+| ~SPC m t t~ | Toggle =transmission-turtle-mode=.                                |
```
lebensterben, line 91: Needs to add key bindings for turtle-mode as well:
"sD" 'transmission-turtle-set-days
"sS" 'transmission-turtle-set-speeds
"sT" 'transmission-turtle-set-times

layers/+tools/transmission/README.org
```diff
@@ -63,52 +64,68 @@ Add =(transmission :variables transmission-auto-refresh-all t)= to
...
 
 ** =transmission-info-mode=
 
 | Key binding | Description                                                   |
 |-------------+---------------------------------------------------------------|
 | ~SPC m c~   | Copy magnet link of current torrent.                          |
```
lebensterben, line 97: we should add an evil normal state binding to copy the magnet link.
it should be "y".

layers/+tools/transmission/README.org
```diff
@@ -63,52 +64,68 @@ Add =(transmission :variables transmission-auto-refresh-all t)= to
...
-| ~SPC m s d~ | Set download limit of torrent(s) at point in kB/s.            |
-| ~SPC m g p~ | Open a =transmission-peers-mode= buffer for torrent at point. |
-| ~SPC m s l~ | Set seed ratio limit of torrent(s) at point.                  |
 | ~SPC m m~   | Move torrent at point or in region to a new location.         |
-| ~SPC m a a~ | Add announce URLs to torrent or torrents.                     |
 | ~SPC m T~   | Remove trackers from torrent at point by ID or announce URL.  |
```
lebensterben, line 99: remove tracker can just be `SPC m D`
(As mentioned in previous comment I recommend to have `SPC m a` for adding trackers in `transmission-info-mode`, then these two bindings pair very well.)

layers/+tools/transmission/README.org
```diff
@@ -63,52 +64,68 @@ Add =(transmission :variables transmission-auto-refresh-all t)= to
...
-| ~SPC m s u~ | Set upload limit of torrent(s) at point in kB/s.              |
+| ~SPC m a t~ | Add announce URLs to torrent or torrents.                     |
+| ~SPC m g p~ | Open a =transmission-peers-mode= buffer for torrent at point. |
```
lebensterben, line 101: we might as well add `SPC m a f` here

layers/+tools/transmission/README.org
```diff
@@ -63,52 +64,68 @@ Add =(transmission :variables transmission-auto-refresh-all t)= to
...
+| ~SPC m R~   | Prompt to remove torrent at point or torrents in region.          |
+| ~SPC m S~   | Sort torrents according to the column at point.                   |
+| ~SPC m a a~ | Add torrent by filename, URL, magnet link, or info hash.          |
+| ~SPC m a t~ | Add announce URLs to torrent or torrents.                         |
+| ~SPC m g i~ | Open a =transmission-info-mode= buffer for torrent at point.      |
+| ~SPC m g f~ | Open a =transmission-files-mode= buffer for torrent at point.     |
+| ~SPC m g p~ | Open a =transmission-peers-mode= buffer for torrent at point.     |
+| ~SPC m s p~ | Set bandwidth priority of torrent(s) at point or in region.       |
+| ~SPC m s d~ | Set global download speed limit in kB/s.                          |
+| ~SPC m s l~ | Set label(s) for marked torrent(s) or torrent at point.           |
```
lebensterben, line 86: the default binding for setting ratio limit and setting labels are "l" and "L" respectively.

We should just use them.

layers/+tools/transmission/README.org
```diff
@@ -63,52 +64,68 @@ Add =(transmission :variables transmission-auto-refresh-all t)= to
...
-| ~SPC m g p~ | Open a =transmission-peers-mode= buffer for torrent at point. |
-| ~SPC m s l~ | Set seed ratio limit of torrent(s) at point.                  |
 | ~SPC m m~   | Move torrent at point or in region to a new location.         |
-| ~SPC m a a~ | Add announce URLs to torrent or torrents.                     |
 | ~SPC m T~   | Remove trackers from torrent at point by ID or announce URL.  |
-| ~SPC m s u~ | Set upload limit of torrent(s) at point in kB/s.              |
+| ~SPC m a t~ | Add announce URLs to torrent or torrents.                     |
+| ~SPC m g p~ | Open a =transmission-peers-mode= buffer for torrent at point. |
+| ~SPC m s d~ | Set download limit of torrent(s) at point in kB/s.            |
+| ~SPC m s l~ | Set seed ratio limit of torrent(s) at point.                  |
```
lebensterben, line 103: we should also add `SPC m s L` for adding labels.

layers/+tools/transmission/README.org
```diff
@@ -63,52 +64,68 @@ Add =(transmission :variables transmission-auto-refresh-all t)= to
...
 | ~SPC m s p~ | Set bandwidth priority of torrent(s) at point or in region.   |
+| ~SPC m s u~ | Set upload limit of torrent(s) at point in kB/s.              |
 
 ** =transmission-files-mode=
 
 | Key binding | Description                                                   |
 |-------------+---------------------------------------------------------------|
-| ~SPC m g f~ | Visit the file at point with =find-file-read-only=.           |
 | ~SPC m r~   | Run a command on the file at point.                           |
```
lebensterben, line 111: it'd better be "!", which parallels with `SPC !` which runs shell command

layers/+tools/transmission/README.org
```diff
@@ -63,52 +64,68 @@ Add =(transmission :variables transmission-auto-refresh-all t)= to
...
-| ~SPC m g p~ | Open a =transmission-peers-mode= buffer for torrent at point. |
+| ~SPC m g f~ | Visit the file at point with =find-file-read-only=.           |
 | ~SPC m g i~ | Open a =transmission-info-mode= buffer for torrent at point.  |
-| ~SPC m m~   | Move torrent at point or in region to a new location.         |
+| ~SPC m g p~ | Open a =transmission-peers-mode= buffer for torrent at point. |
+| ~SPC m m m~ | Move torrent at point or in region to a new location.         |
```
lebensterben, line 115: we should remove `SPC m m m`. normally user won't try to remove a torrent in files view.
also by default there's no such binding in transmission-files-mode-map

layers/+tools/transmission/README.org
```diff
@@ -63,52 +64,68 @@ Add =(transmission :variables transmission-auto-refresh-all t)= to
...
 | ~SPC m m u~ | Mark file(s) at point or in region as unwanted.               |
```
lebensterben, line 116: for wanted and unwanted, we can use `SPC m t w/u` instead.
this frees up `SPC m m` for `transmission-toggle-mark`.

lebensterben:
Looks pretty good. See my new comments

layers/+tools/transmission/README.org
arifer612: "!" is already provided in evil-collection and is the default binding provided by the package as you mentioned. I will change this to `SPC m X` for "E(X)ECUTE in the major mode".

layers/+tools/transmission/README.org
arifer612: This is an easy task, I'll get this done up!

layers/+tools/transmission/README.org
arifer612: Adding headers is an easy task as well. It'll make reading the README much smoother, got it.

I'll remove `transmission-trackers-add` from `transmission-mode-map` and just leave it in `transmission-info-mode-map` where trackers are listed explicitly.

layers/+tools/transmission/README.org
arifer612: I had thought of having it in `SPC m t m` because it was a toggle. But if having it out in `SPC m m ` then I believe adding `SPC m x` for `transmission-toggle` (toggle between activating and deactivating a torrent process) should be added too.

layers/+tools/transmission/README.org
arifer612: `SPC m T` is used here to match the binding used in evil-collections, `T` for "Remove (T)racker". Using `SPC m D` here might cause some confusion as it is the same binding for `transmission-delete` in `transmission-mode-map`

layers/+tools/transmission/README.org
arifer612: Oh, it has already been added but I forgot to correct this table! What a blunder on my end!

layers/+tools/transmission/README.org
arifer612: I'll work on the description for these 2. The reference you provided is perfect! I never thought of checking the source for Transmission.

The reason why I thought of changing the binding to "R" and "D" respectively was because they're both destructive executions (removing data from the client/disk). Both `transmission-remove` and `transmission-delete` prompt the user for confirmation of the removal (and deletion) before carrying out the process.

layers/+tools/transmission/README.org
arifer612: Sure thing, this will then match up with the mode-networkings that were provided in the default bindings.

layers/+tools/transmission/README.org
arifer612: I am in favour of "r" and "l" for "(r)atio limit" and "(l)abel" because I have been using a function I made bound to "L" to set "(L)abels by regex", but this then would be different from both the default bindings and evil-collections.

They are consistent in all mode maps, only the README file was not updated properly as I mentioned earlier above.

layers/+tools/transmission/README.org
arifer612: I believe that the original author had meant for marked files of a multi-file torrent to be moved to a different location with this binding. `transmission-move` does not serve this purpose so I will remove it.

layers/+tools/transmission/README.org
arifer612: That's a good idea! This is an easy change to make.

layers/+tools/transmission/README.org
arifer612: I have just remembered that ~SPC m m~ to toggle marks would clash with ~transmission-move~.

arifer612: Tying with the comment about fixing the keybindings for ~transmission-remote~ and ~transmission-delete~, we can just have
"R": ~transmission-remove~ for "[R]emove"
"D": ~transmission-delete~ for "[D]elete"
"r": ~transmission-move~ for "[r]ename"
"m": ~transmission-toggle-mark~ for "[m]ark"

layers/+tools/transmission/README.org
arifer612: After checking on the documentation for `transmission-trackers-add`, I believe that it should remain in `transmission-mode-map` because it should be used to add trackers to the selected torrent, or multiple marked torrents even in `transmission-mode`. For that reason, we should leave `SPC m a` as a prefix for `SPC m a a` and `SPC m a t`.

layers/+tools/transmission/README.org
arifer612: This has been settled!

layers/+tools/transmission/README.org
arifer612: I've made the descriptions a little clearer. I've left the bindings as "R" and "D" for the purpose of having `transmission-move` and `transmisison-mark` not clash with each other.

    | `SPC m D`    | Delete from disk and remove torrent(s) at point or in region.     |
    | `SPC m R`    | Remove torrent(s) at point or in region.                          |

layers/+tools/transmission/README.org
arifer612: Added `SPC m g f` here!

layers/+tools/transmission/README.org
arifer612: Corrected the README file

layers/+tools/transmission/README.org
arifer612: Changed the key binding to `SPC m X` as proposed

layers/+tools/transmission/README.org
arifer612: Changed the prefixes for these 2 bindings and `SPC m m` is reserved for `transmission-toggle-mark` to maintain consistency with the other modes.

layers/+tools/transmission/README.org
arifer612: Removed this binding.

layers/+tools/transmission/README.org
arifer612: Made the changes as mentioned in the comment above this.

arifer612:
I've made several changes. Only those that were not marked resolved are left as they are, with my reasons for so in the thread. I'd like to hear more about your thoughts about my opinions for not changing them.

I will not be squashing all the commits at once anymore until everything is finalised. I realised that it gets messy if I just squash everything every time I make a change for review.

lebensterben:
@arifer612 
Don't worry about squashing. I can squash and merge.

arifer612:
Are there any comments regarding the latest version of the change?

lebensterben (approved):

lebensterben:
@arifer612 
Sorry for the delay. It's ready to merge now.
Do you want to update the commit message?

lebensterben:
@arifer612 
Terrific. That commit message is well-written.

I've just merged this PR. Thanks for your contribution.

arifer612:
Thank you for the kind words! I really appreciate it!

Issue #14820, opened 2021-06-02
devdocs now requires emacs 27.1


mbertheau:
Up until a few days ago devdocs was this package: https://github.com/xuchunyang/DevDocs.el/blob/master/devdocs.el

Now it is this package: https://elpa.gnu.org/packages/devdocs.html https://github.com/astoff/devdocs.el

The new package requires emacs 27.1: https://github.com/astoff/devdocs.el/blob/8a3f4e00e0a2113c604c631f81f7db90c1b56d5b/devdocs.el#L8

devdocs is in spacemacs-misc: https://github.com/syl20bnr/spacemacs/blob/1ce4a7fd66a4f19c8b9008798d526a754b213f94/layers/%2Bspacemacs/spacemacs-misc/packages.el#L52

How do we handle this?

lebensterben:
make the last compatible one a local package

mbertheau:
How do we handle this for all spacemacs users that use emacs 26?

lebensterben:
I said local package.
we've plenty of local packages already

lebensterben:
example of local package: https://github.com/syl20bnr/spacemacs/blob/af86a142ebeff3bfa0c5cb41d47455af3cd68eab/layers/%2Bspacemacs/spacemacs-defaults/packages.el#L41

For Emacs < 27, use a local package. Otherwise use the upstream one.

emacs18:
This was discussed recently on emacs-dev list.  The new package has far more features than the old one.
Also https://github.com/blahgeek/emacs-devdocs-browser is similar package which I test drove.
I assume that emacs-devdocs-browser and devdocs packages provide similar features.

I think these new document browsing packages may be more convenient for some use cases than https://github.com/dash-docs-el/dash-docs which I've used for a while.

So I hope that we can find a way to provide these really cool features to those spacemacs users willing to use newer emacs versions.

thanhvg:
I think we should just spin off devdocs to an optional layer for Emacs 27. The old devdocs package was very simple, all it did was to open devdocs.io with current symbol. I don't think anyone would miss it.  While local package should work, I don't like the idea that Spacemacs will take the maintenance job of a deprecate one. I also prefer `devdoc-browser` (https://github.com/blahgeek/emacs-devdocs-browser) over the new `devdocs`.

But this trouble could be avoided if we only supported the latest stable Emacs in the first place.

emacs18:
[...]

PR #14821, opened 2021-06-02
Lenient pkg delete

mbertheau:
I had the problem described in https://github.com/syl20bnr/spacemacs/issues/14820

During a package update on an emacs 26, the devdocs package failed to install an updated version because it now requires emacs 27.1. This prompted me to do a package rollback.

Upon package rollback, spacemacs tried to uninstall the current devdocs package in preparation for restoring the old devdocs package. Since the new package devdocs isn't installed at all, the deletion failed, and with it the whole rollback.

The changes in this PR allow the rollback to continue by having package-delete not fail hard, but instead just output a message.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -2398,11 +2398,12 @@ depends on it."
 
 (defun configuration-layer//package-delete (pkg-name)
   "Delete package with name PKG-NAME."
-  (let ((p (cadr (assq pkg-name package-alist))))
-    ;; add force flag to ignore dependency checks in Emacs25
-    (if (not (configuration-layer//system-package-p p))
-        (package-delete p t t)
-      (message "Would have removed package %s but this is a system package so it has not been changed." pkg-name))))
+  (if-let ((pkg (car (alist-get pkg-name package-alist))))
+      ;; add force flag to ignore dependency checks in Emacs25
+      (if (configuration-layer//system-package-p pkg)
+          (message "Would have removed package %s but this is a system package so it has not been changed." pkg-name)
+        (package-delete pkg t t))
+    (message "Can't remove package %s since it isn't installed." pkg-name)))
 
 (defun configuration-layer/delete-orphan-packages (packages)
   "Delete PACKAGES if they are orphan."


core/core-configuration-layer.el
```diff
@@ -2398,11 +2398,14 @@ depends on it."
 
 (defun configuration-layer//package-delete (pkg-name)
   "Delete package with name PKG-NAME."
-  (let ((p (cadr (assq pkg-name package-alist))))
-    ;; add force flag to ignore dependency checks in Emacs25
-    (if (not (configuration-layer//system-package-p p))
-        (package-delete p t t)
-      (message "Would have removed package %s but this is a system package so it has not been changed." pkg-name))))
+  (let ((pkg-cell (assq pkg-name package-alist)))
```
lebensterben, line 2401: use if-let + cadr-safe

lebensterben (changes requested):

core/core-configuration-layer.el
mbertheau: Done.

mbertheau: I hope I got what you mean correctly. My experience in elisp is so far measured in mere days.

core/core-configuration-layer.el
```diff
@@ -2398,11 +2398,13 @@ depends on it."
 
 (defun configuration-layer//package-delete (pkg-name)
   "Delete package with name PKG-NAME."
-  (let ((p (cadr (assq pkg-name package-alist))))
-    ;; add force flag to ignore dependency checks in Emacs25
-    (if (not (configuration-layer//system-package-p p))
-        (package-delete p t t)
-      (message "Would have removed package %s but this is a system package so it has not been changed." pkg-name))))
+  (if-let ((pkg-cell (assq pkg-name package-alist))
```
lebensterben, line 2401: You can combine the two bindings

mbertheau: Done.

core/core-configuration-layer.el
```diff
@@ -2398,11 +2398,12 @@ depends on it."
...
 (defun configuration-layer//package-delete (pkg-name)
   "Delete package with name PKG-NAME."
-  (let ((p (cadr (assq pkg-name package-alist))))
-    ;; add force flag to ignore dependency checks in Emacs25
-    (if (not (configuration-layer//system-package-p p))
-        (package-delete p t t)
-      (message "Would have removed package %s but this is a system package so it has not been changed." pkg-name))))
+  (if-let ((p (car-safe (cdr-safe (assq pkg-name package-alist)))))
+      ;; add force flag to ignore dependency checks in Emacs25
+      (if (not (configuration-layer//system-package-p p))
```
lebensterben, line 2403: let's remove `not` from the condition and swap the branches of `if`.

core/core-configuration-layer.el
```diff
@@ -2398,11 +2398,12 @@ depends on it."
 
 (defun configuration-layer//package-delete (pkg-name)
   "Delete package with name PKG-NAME."
-  (let ((p (cadr (assq pkg-name package-alist))))
-    ;; add force flag to ignore dependency checks in Emacs25
-    (if (not (configuration-layer//system-package-p p))
-        (package-delete p t t)
-      (message "Would have removed package %s but this is a system package so it has not been changed." pkg-name))))
+  (if-let ((p (car-safe (cdr-safe (assq pkg-name package-alist)))))
```
lebensterben, line 2401: well my bad. I told you to use car-safe + cdr-safe.
But we can use `alist-get` (it's a C subroutine and is very fast).
So
`(if-let ((pkg (alist-get pkg-name package-alist))) ..)`

mbertheau: Done.

core/core-configuration-layer.el
mbertheau: Done.

lebensterben:

mbertheau:
@lebensterben I'm following this https://www.spacemacs.org/CONTRIBUTING.html#commit-messages with the commit message, but the commit message body didn't end up in the commit on develop - should I restrict myself to just the subject line in the commit message?

lebensterben:
that's probably because I merged on phone and used the default commit message, which I don't know where it's coming from...

PR #14824, opened 2021-06-03
racket: Function-quote company-mode-hook function

arminfriedl:
At least in Emacs >= 28.0.50 a warning occurs at startup when enabling racket
mode:
`.emacs.d/profiles/spacemacs/layers/+lang/racket/packages.el: Warning: (lambda nil \.\.\.) quoted with ' rather than with #'`

In 27.2 the warning does not appear.

In any case, the warning is fair and the change does not really affect anything
apart from the anonymous function being compiled potentially.


lebensterben:
the quote is redundant. you can remove it

arminfriedl:
True, changed it

lebensterben:
Merged. Thanks for your contribution

Issue #9169, opened 2017-07-01
Wrong key bindings for Holy Mode in Git interactive rebase buffer


dsedivec:
#### Description :octocat:

Git interactive rebase buffer comments give incorrect key bindings in Holy Mode.

#### Reproduction guide :beetle:
- Start Emacs with Spacemacs in Holy Mode
- Start an interactive rebase in Magit (e.g. `M-x magit-status RET l l`, move down a few commits, `r i`)

*Observed behaviour:* :eyes: :broken_heart:

Comments about key bindings in the rebase buffer:

```
# Commands:
# p        pick = use commit
# r        reword = use commit, but edit the commit message
# e        edit = use commit, but stop for amending
# s        squash = use commit, but meld into previous commit
# f        fixup = like "squash", but discard this commit's log message
# x        exec = run command (the rest of the line) using shell
# d        drop = remove commit
# u        undo last change
# C-c C-c  tell Git to make it happen
# C-c C-k  tell Git that you changed your mind, i.e. abort
# k        move point to previous line
# j        move point to next line
# M-k      move the commit at point up
# M-j      move the commit at point down
# RET      show the commit at point in another buffer
```

Some of these are incorrect.  For example, `p` moves to the previous commit, it does not "pick" a commit.  `M-k` does not move a commit up.  Others may also be wrong.

*Expected behaviour:* :heart: :smile:

Normally these comments are added/corrected by `git-rebase-mode-show-keybindings`, but merely loading evil-magit removes this function from `git-rebase-mode-hook`.  If you either don't load evil-magit or else restore this function as in

```
;; This is the hook that evil-magit adds.
(remove-hook 'git-rebase-mode-hook 'evil-magit-add-rebase-messages)
(add-hook 'git-rebase-mode-hook 'git-rebase-mode-show-keybindings t)
```

then a rebase buffer will explain the correct key bindings:

```
# Commands:
# C-c C-c  tell Git to make it happen
# C-c C-k  tell Git that you changed your mind, i.e. abort
# p        move point to previous line
# n        move point to next line
# M-p      move the commit at point up
# M-n      move the commit at point down
# SPC      show the commit at point in another buffer
# RET      show the commit at point in another buffer and select its window
# C-_      undo last change
# k        drop the commit at point
# y        insert a line for an arbitrary commit
# z        add noop action at point
# <normal-state> p, c pick = use commit
# <normal-state> r, r, w reword = use commit, but edit the commit message
[...]

PR #14825, opened 2021-06-05
[git] Only load evil-collection in vim or hybrid mode

lyallcooper:
As described in #9169, when using the emacs editing style the comment inserted at the bottom of a git rebase shows the incorrect key bindings. This is apparently due to `evil-collection` [changing the `git-rebase-mode-hook`](https://github.com/emacs-evil/evil-collection/blob/3ed92cadda7ff1c2d9bc43ec2352e5b02d9e6477/modes/magit/evil-collection-magit.el#L523-L524).

We can fix the issue by only loading `evil-collection` if we're using an evil-style editing mode.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -172,6 +172,7 @@ the [[file:CHANGELOG.org][CHANGELOG.org]] file.
 - Key bindings;
   - Changed ~SPC g f h~ to ~SPC g f l~ for 'git log' of current file
     (thanks to Ag Ibragimov)
+- Fixed rebase mode displaying incorrect keybindings when using emacs editing style
 ***** gtags
 - Key bindings;
   - Regenerate tags are now under ~SPC m g C~ instead of ~SPC m g c~.
--- a/layers/+source-control/git/funcs.el
+++ b/layers/+source-control/git/funcs.el
@@ -73,11 +73,10 @@
   (let (git-link-open-in-browser)
     (call-interactively 'git-link-commit)))
 
-
-(defun spacemacs//support-evilified-buffer-p (style)
-  "Return non-nil if evil navigation should be enabled for STYLE."
-  (or (eq style 'vim)
-      (and (eq style 'hybrid)
+(defun spacemacs//support-evilified-buffer-p ()
+  "Return non-nil if evil navigation should be enabled."
+  (or (eq dotspacemacs-editing-style 'vim)
+      (and (eq dotspacemacs-editing-style 'hybrid)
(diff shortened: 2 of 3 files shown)


lebensterben (changes requested):
Can you change `spacemacs//support-evilified-buffer-p` to
```elisp
(defun spacemacs//support-evilified-buffer-p ()
  "Return non-nil if evil navigation should be enabled."
  (or (eq dotspacemacs-editing-style 'vim)
      (and (eq dotspacemacs-editing-style 'hybrid)
           hybrid-style-enable-evilified-state)))
```

lyallcooper:
> Can you change `spacemacs//support-evilified-buffer-p` to [...]

Good call, updated.

lebensterben:
merged. thanks for your contribution

winsphinx:
cause a warning for vim/hybrid user:

````
 Warnings:
     - Ignoring :toggle for package evil-collection because layer git does not own it.      
````

**Comment by dschrempf** [comment]
After this commit I get the following error during startup:

```
Warning (initialization): An error occurred while loading ‘/home/dominik/.emacs.d/init.el’:

Symbol's function definition is void: spacemacs//support-evilified-buffer-p

To ensure normal operation, you should investigate and remove the
cause of the error in your initialization file.  Start Emacs with
the ‘--debug-init’ option to view a complete error backtrace.
```

Going back to `eda7e193701742c58857e556b60db9bff6896916` solves the issue.

**Comment by lebensterben** [comment]
@dschrempf 
sorry. #14826 should fix it. can you verify that?

**Comment by kappa** [comment]
@lebensterben No, it does not.

```
Debugger entered--Lisp error: (void-function spacemacs//support-evilified-buffer-p)
eval-buffer(#<buffer  *load*-147642> nil "/home/kappa/.emacs.d/layers/+source-control/git/pa..." nil t)  ; Reading at buffer position 2167
```

which points to

```
(when (spacemacs//support-evilified-buffer-p)
  (defun git/pre-init-evil-collection ()
    (add-to-list 'spacemacs-evil-collection-allowed-list 'magit)))
```
funcs.el is loaded after packages.el.

lebensterben:
@kappa 
You're right, but I've just pushed a new commit which fixed this issue.

lyallcooper:
Ah sorry for the issue. I swear I tested this in vim mode, which is what I normally use, and didn't see any issues 🤷‍♂️

dschrempf:
I can confirm that it works no (I guess this is the second one of your commits). Thank you!

PR #14827, opened 2021-06-05
fixup fixup fixup #14715

lebensterben:
funcs.el is loaded after packages.el so `spacemacs//support-evilified-buffer-p` is undefined at this point.

moved it inside the `defun` block and it should work.

### DIFF
--- a/layers/+source-control/git/packages.el
+++ b/layers/+source-control/git/packages.el
@@ -57,8 +57,8 @@
     :post-config
     (add-to-list 'golden-ratio-exclude-buffer-names " *transient*")))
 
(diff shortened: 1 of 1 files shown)


WarFox:
Fixes issue with `spacemacs//support-evilified-buffer-p` being undefined

but now getting this
```
(void-variable spacemacs-evil-collection-allowed-list)
```

WarFox:
Pardon me, above comment is wrong. I may have done something when I copy pasted code.

Switched you your branch in it works fine 🎉

lebensterben:
@WarFox 
Thanks for testing.

PR #14830, opened 2021-06-06
[nix] Add key binding for nix-format-buffer

bcc32:
Add key binding `, = =` (nix-format-buffer) in nix-mode buffers.


smile13241324 (approved):
Looks good

PR #14836, opened 2021-06-09
Fix wrong buffer name after file rename

dalanicolai:
When providing only the new directory path (i.e. without new filename) while
moving a file to a new directory with `spacemacs/rename-current-buffer-file`,
Spacemacs currently opens the file in a new buffer with the directory name.

Additionally, the `file-renamed-p` gets set to the wrong value. To fix both
issues (with minimal work), it is easiest to just fix the `new-name`
value (alternatively fixing the `rename-buffer` and `set-visited-file-name`
forms later would not fix the wrong `file-renamed-p` value issue).

Finally when called with a universal-argument, then that argument unintended gets 
'passed' in the interactive call to `projectile-invalidate-cache`. So this PR fixes that by changing
the interactive call to a `funcall`.

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -418,7 +418,10 @@ projectile cache and updates recentf list."
   (let* ((old-short-name (buffer-name))
          (old-filename (buffer-file-name))
          (old-dir (file-name-directory old-filename))
-         (new-name (read-file-name "New name: " (if arg old-dir old-filename)))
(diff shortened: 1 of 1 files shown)


duianto:
I'm able to apply this PR as a patch without any issues, but Github says:
>This branch cannot be rebased safely
Rebasing the commits of this branch on top of the base branch cannot be performed automatically as this would create a different result than a regular merge.

Could you rebase this.

dalanicolai:
Ah strange... well rebasing was also no problem. Thanks for processing!

duianto:

PR #14838, opened 2021-06-10
[org] Fix org-habit

arifer612:
Adding `org-habit` to `org-modules` under `(with-eval-after-load 'org-agenda)` in `init-org`
will not properly load it as a module unless `org-agenda` is started before `org`.

Below are some photos of how the issue happens. I have managed to reproduce it on an Arch and Ubuntu distribution with a stock Spacemacs setup.

## Starting `org-agenda` first
Starting `org-agenda` right when Emacs is launched, the `org-habit` graph is displayed just fine. `org-habit` is also seen to be added as a module just fine.
Starting `org-agenda` first

## Starting `org-mode` first
However, starting an Org file first, the `org-habit` graph no longer appears in the agenda tab even though it appears to be added as a module.
Starting `org-mode` first

## Fix
The fix that I have made in this commit is to move the appending of `org-habit` to `org-modules` from right after initialising `org` to after initialising `org-agenda`. The reason for this is because I had realised that 

    (use-package org
      :init
      (progn
        ...
        (with-eval-after-load 'org-agenda
          (add-to-list 'org-modules 'org-habit))
        ...))

meant that after initialising `org`, Emacs will only run the code in the wrapper if `org-agenda` had already been loaded. That is not the case for situation 2 where I had first opened up an Org file before checking `org-agenda`. 

Moving it to 

    (use-package org-agenda
      :after org
      :init
       (progn
         ...
         (with-eval-after-load 'org
           (add-to-list 'org-modules 'org-habit)
         ...))

meant that since `org-agenda` would only be loaded after `org` is loaded, `org-habit` would definitely be added as a module. I had tried to do this without the `with-eval-after-load` wrapper but for some reason I know not of, it simply did not work.

### DIFF
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -160,9 +160,6 @@
                     (naive-auto #'spacemacs/org-summary-todo-naive-auto)
                     (semiauto #'spacemacs/org-summary-todo-semiauto))))
 
-      (with-eval-after-load 'org-agenda
-        (add-to-list 'org-modules 'org-habit))
-
       (with-eval-after-load 'org-indent
         (spacemacs|hide-lighter org-indent-mode))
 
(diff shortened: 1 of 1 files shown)


lebensterben (approved):
LGTM

arifer612:
Is there anything else for me to add to this change?

lebensterben:

arifer612:
Awesome, thank you for the help!

PR #14839, opened 2021-06-11
Fixed Links in README.md

electron271:
Fixes a issue where when clicking on the doc links it will redirect you to a page that firefox classifies as a "Potential Security Risk"


lebensterben:
though this fixes the problem, it's a design choice to use "spacemacs.org" without subdomain "www".

this problem also occurs to Safari. Basically some modern browsers don't trust wildcard encryption certificate

So the correct solution is to fix it on the server side.

@syl20bnr 
@JAremko

electron271:
:+1:

smile13241324:
This change is not longer necessary, we have changed the destination for all links from spacemacs.org to develop.spacemacs.org for which the certs are valid.

PR #14848, opened 2021-06-16
[core] fix failure message when `user-emacs-directory' not exists

sunlin7:
*core-load-paths.el: use function `locate-user-emacs-file` replace create
directory manually.

The Spacemacs will failed for creating `~/.emacs.d/.cache/` when `~/.emacs.d` dose *NOT* exist.
You can reproduce this issue by removing `~/.emacs.d` and installing Spacemacs in `~/.spacemacs/`, loading the `init.el` from `~/.emacs`.

### DIFF
--- a/core/core-load-paths.el
+++ b/core/core-load-paths.el
@@ -100,7 +100,7 @@
 
 ;; ~/.emacs.d/.cache
 (defconst spacemacs-cache-directory
-  (concat user-emacs-directory ".cache/")
+  (locate-user-emacs-file ".cache/")
   "Spacemacs storage area for persistent files.")
 
 ;; ~/.emacs.d/.cache/auto-save


sunlin7:
@smile13241324 , please help review this patch. Thanks.

lebensterben:
You should just set `user-emacs-directory` and it would be fine.

lebensterben (changes requested):
-1

lebensterben:
The docstring of `user-emacs-directory` reads:
> Return an absolute per-user Emacs-specific file name.
> 
> If NEW-NAME exists in user-emacs-directory, return it.
Else if OLD-NAME is non-nil and ~/OLD-NAME exists, return ~/OLD-NAME.
Else return NEW-NAME in user-emacs-directory, creating the
directory if it does not exist.

When the directory exists, it's fine. Nothing changed from current behaviour.
But when it doesn't exist, this PR would tacitly create a new directory.

If you simply set `user-emacs-directory` correctly, then everything should be fine.
If I'm mistaken, let me know why it isn't case.
Otherwise please close this PR.

sunlin7:
Hi @lebensterben , 

The chagne is updated. 

The key point is, `spacemacs-cache-directory` is more than two levels directory `~` / `.emacs.d` / `.cache`, the line https://github.com/syl20bnr/spacemacs/blob/2fd3eb3edbc7c09b825892ce53721120bb999504/core/core-load-paths.el#L116 just try make the last folder, and failed when its parent directory doese **NOT** exits.

You can reproduce the issue with the steps in my initial comment.

lebensterben:
The direct solution is to change `(make-directory spacemacs-cache-directory))` to `(make-directory spacemacs-cache-directory t))`.

But not setting the optional argument to `t` is an intentional choice.
It's only needed if parent directory doesn't exist.
Should that happen, other parts of Emacs functionality are also compromised.

The correct way to handle this is:
- set `emacs-user-directory` correctly. `spacemacs-cache-directory` and any `spacemacs-*-directory` will correctly set to desired value.
- user should make sure `emacs-user-directory` exist.

sunlin7:
User didn't have to make sure the `user-emacs-directory`. Emacs will call `locate-user-emacs-file` to ensure the `.emacs.d` be created when its not exist. 

You can remove your `~/.emacs.d` and execute `emacs -q` to observe that the `.emacs.d` folder is created by calling `locate-user-emacs-file`. And sadly the line 116 is early than vanilla emacs lisp and failed in my scenario.
https://github.com/syl20bnr/spacemacs/blob/2fd3eb3edbc7c09b825892ce53721120bb999504/core/core-load-paths.el#L116

lebensterben:
You're not listening.


> User don't have to make sure the `user-emacs-directory`.

Not creating parent directory is a design choice.
What if `user-emacs-directory` is set to system directory, e.g. `/usr/lib64`, and the user happens to have write permission?

Your PR would TACITLY create directory at where user should most likely not touch.

Should user ever change `user-emacs-directory`, he should be responsible to make sure it exists.

> You can remove your `~/.emacs.d` and execute `emacs -q` to observe that the `.emacs.d` folder is created by calling `locate-user-emacs-file`.

By running emacs with `-q` flag, you're using the default value of `user-emacs-directory`, which is `~/.emacs.d`. And of course it is safe to create such directory if it's missing.

lebensterben:
https://raw.githubusercontent.com/emacs-mirror/emacs/master/ChangeLog.3
```txt
2019-08-30  Paul Eggert  <eggert@cs.ucla.edu>

	Prefer ~/.config/emacs to ~/.emacs.d if neither exists

	That way, when Emacs starts in a fresh home directory,
	it prefers the new (XDG) convention rather than the old one.
	* lisp/files.el (locate-user-emacs-file): Make the parent
	directories of user-emacs-directory if needed.  This is useful
	if user-emacs-directory is "~/.config/emacs" and "~/.config"
	does not yet exist.
```

Clearly the intended use case for `locate-user-emacs-file` is limited to home directory and alike.
By default, vanilla Emacs does call `locate-user-emacs-file`, but it WON'T create any directory outside home directory.

In this PR,

```elisp
(locate-user-emacs-file ".cache/")
```

It would tacitly create the directory in any unexpected place, should `user-emacs-directory` is changed.

sunlin7:
The `locate-user-emacs-file` is called twice:

>emacs startup ↓
>...
>(locate-user-emacs-file) ; from Spacemacs
>...
>(locate-user-emacs-file) ; from Emacs
>...

Did you mean the `user-emacs-directory` maybe changed  between two `locate-user-emacs-file` callings?

lebensterben:
> The `locate-user-emacs-file` is called twice:
> 
> > emacs startup ↓
> > ...
> > (locate-user-emacs-file) ; from Spacemacs
> > ...
> > (locate-user-emacs-file) ; from Emacs
> > ...
> 
> Did you mean the `user-emacs-directory` maybe changed between two `locate-user-emacs-file` callings?

It's called in `info+` package, and `startup.el` of Emacs.
The former is not in our control.
The later one I've already explained above.

Essentially we should never call this function in our code base, for reasons I've already explained for multiple times.

We should always ask explicit permission from user when creating directory that's potentially not under `$HOME` or `$XDG_CONFIG_HOME`.

sunlin7:
Okay, got your point.

PR #14856, opened 2021-06-17
Adds function to run default shell in project root, with leader keys `SPC p $`

ryanprior:
Hi there! Here's a patch to add a new function, `spacemacs/projectile-shell`, which creates a buffer with the default shell and switches to it. It behaves similarly to `projectile-multi-term-in-root` which was previously bound to `SPC p $ t`, and this function is more general, so I gave it leader keys `SPC p $`.

### DIFF
--- a/layers/+tools/shell/README.org
+++ b/layers/+tools/shell/README.org
@@ -287,22 +287,25 @@ Some advanced configuration is setup for =eshell= in this layer:
 
 * Key bindings
 
-| Key binding   | Description                                                |
-|---------------+------------------------------------------------------------|
-| ~SPC '​~       | Open, close or go to the default shell                     |
-| ~SPC "​~       | Open external terminal emulator in current directory       |
-| ~SPC p '​~     | Open a shell in the project's root                         |
-| ~SPC p "​~     | Open external terminal emulator in project root            |
-| ~SPC a t s e~ | Open, close or go to an =eshell=                           |
-| ~SPC a t s i~ | Open, close or go to a =shell=                             |
-| ~SPC a t s m~ | Open, close or go to a =multi-term=                        |
-| ~SPC a t s t~ | Open, close or go to a =ansi-term=                         |
-| ~SPC a t s T~ | Open, close or go to a =term=                              |
-| ~TAB~         | browse completion with =helm=                              |
-| ~SPC m H~     | browse history with =helm= (works in =eshell= and =shell=) |
-| ~SPC a t s v~ | Open, close or go to a =vterm=                             |
-| ~C-j~         | next item in history                                       |
-| ~C-k~         | previous item in history                                   |
+| Key binding   | Description                                              |
+|---------------+----------------------------------------------------------|
+| ~SPC '​~       | Toggle pop-shell with your default shell                 |
+| ~SPC "​~       | Open external terminal emulator in current directory     |
+| ~SPC a t s e~ | Toggle pop-shell with =eshell=                           |
+| ~SPC a t s i~ | Toggle pop-shell with =shell=                            |
+| ~SPC a t s m~ | Toggle pop-shell with =multi-term=                       |
+| ~SPC a t s t~ | Toggle pop-shell with =ansi-term=                        |
+| ~SPC a t s T~ | Toggle pop-shell with =term=                             |
+| ~SPC a t s v~ | Toggle pop-shell with =vterm=                            |
+|---------------+----------------------------------------------------------|
+| ~SPC p '​~     | Toggle pop-shell with your default shell in project root |
+| ~SPC p "​~     | Open external terminal emulator in project root          |
+| ~SPC p $~     | Open a new buffer with default shell in project root     |
+|---------------+----------------------------------------------------------|
+| ~TAB~         | In a shell buffer, browse completions                    |
+| ~SPC m H~     | In =shell= or =eshell=, browse history                   |
+| ~C-j~         | Next item in history                                     |
+| ~C-k~         | Previous item in history                                 |
 
 *Note:* You can open multiple shells using a numerical prefix argument,
 for instance pressing ~2 SPC '​~ will a second default shell, the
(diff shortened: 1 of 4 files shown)


lebensterben:
Great addition.

We've a variable `shell-default-shell`, I suggest to create a new function `spacemacs/projectile-run-shell`, which is a dispatcher to various projectile-shell commands.

It can be bounded to `SPC p $ $`.

What do you think?

ryanprior:
@lebensterben thank you for your suggestion, I've attempted an implementation. I like this better than adding redundant leader keys for each different shell type.

layers/+tools/shell/funcs.el
```diff
@@ -27,6 +27,14 @@
   (let ((default-directory (projectile-project-root)))
     (call-interactively 'spacemacs/default-pop-shell)))
 
+(defun spacemacs/projectile-run-shell ()
```
ryanprior, line 30: Maybe we should call it `spacemacs/default-run-shell`?

ryanprior (commented):
To get bolder: what do you think about removing the `SPC p $ t` binding & making `SPC p $` run the default shell in the project root? This would save everybody a keystroke.

layers/+tools/shell/funcs.el
```diff
@@ -27,6 +27,14 @@
...
+  "Run the default shell in the project root directory."
+  (interactive)
+  (let ((shell (cl-case shell-default-shell
+                 ('multi-term #'projectile-multi-term-in-root)
+                 (t (intern (format "projectile-run-%S" shell-default-shell))))))
```
lebensterben, line 35: When `shell-default-shell` is `ansi-term`, it would fail.
Just add `('ansi-term nil)`
And change `let` to `when-let`

lebensterben (changes requested):

layers/+tools/shell/funcs.el
lebensterben: `spacemacs/projectile-run-default-shell`

layers/+tools/shell/funcs.el
```diff
@@ -27,6 +27,15 @@
   (let ((default-directory (projectile-project-root)))
     (call-interactively 'spacemacs/default-pop-shell)))
 
+(defun spacemacs/projectile-run-default-shell ()
+  "Run the default shell in the project root directory."
+  (interactive)
+  (when-let ((shell (cl-case shell-default-shell
+                 ('multi-term #'projectile-multi-term-in-root)
+                 ('ansi-term nil)
```
lebensterben, line 35: on second thought. maybe we can do this
```elisp
(call-interactively (if (eq shell-default-shell 'multi-term)
                        #'projectile-multi-term-in-root
                      (or (intern-soft (format "projectile-run-%s") shell-default-shell)
                          #'projectile-run-shell)))
```

And change the docstring accordingly.

lebensterben (changes requested):

lebensterben:
Also don't forget to update `CHANGELOG.develop`

layers/+tools/shell/funcs.el
```diff
@@ -27,6 +27,15 @@
   (let ((default-directory (projectile-project-root)))
     (call-interactively 'spacemacs/default-pop-shell)))
 
+(defun spacemacs/projectile-run-default-shell ()
+  "Run the default shell at the project root. If projectile
+doesn't support this, use projectile-run-shell instead."
```
lebensterben, line 32: as for docstring:
- the first line should be a short and complete sentence. So add a new line after the period.
- symbol name should be quoted with <code>&#96;&apos;</code> so  <code>&#96;projectile-run-shell&apos;</code>

lebensterben:
Well. I just realised the functionality is almost the same as `spacemacs/projectile-pop-shell`....

ryanprior:
> Also don't forget to update `CHANGELOG.develop`

I don't have a `spacemacs/projectile-pop-shell` function.

In any case, this is useful to me because sometime I want to "pop" a shell (creating a new full-width buffer along the bottom of the frame) but often I want a full-height shell buffer that replaces my current one, which is what this function provides.

lebensterben:
Okay. That sounds reasonable.
So we've `spacemacs/projectile-shell-pop` which opens a shell at projectile root dir in a popup window.
Our new addition in this PR opens a shell at projectile root dir but takes the whole window.

So I think we better match their names. For example, `spacemacs/projectile-shell` and `spacemacs/projectile-shell-pop`, or any names that clearly shows their similarities and differences.

Also we'd better document this in layer's README.org and in both functions docstring.
This is to make sure user understand their sole (obvious) difference is whether the shell takes the whole window or in a popup.

ryanprior:
I took another pass and did these things:

- renamed `spacemacs/projectile-run-default-shell` -> `spacemacs/projectile-shell`
- updated the docstrings to emphasize the differences between `spacemacs/projectile-shell-pop` and `spacemacs/projectile-shell`
- updated README to clarify the different shell options
- removed the `SPC p $` prefix & the leader keys `SPC p $ t` which specifically opens a multi-term
- adds leader keys `SPC p $` which runs `spacemacs/projectile-shell` (which will run a multi-term for those who prefer it)

layers/+tools/shell/funcs.el
```diff
@@ -22,11 +22,27 @@
...
 
 (defun spacemacs/projectile-shell-pop ()
-  "Open a term buffer at projectile project root."
+  "Pop-up a shell buffer at the project root.
+Customize `shell-default-shell' to control what type of shell
+buffer you create. This function will pop-up a full-width buffer
+and move your focus to it; to switch the current buffer view, use
+`spacemacs/projectile-shell'."
   (interactive)
   (let ((default-directory (projectile-project-root)))
```
lebensterben, line 31: There's a pre-existing bug. When it's not in a projectile directory, this command will fail.
One solution is
```elisp
(if-let* ((backup default-directory)
          (default-directory (projectile-project-root)))
    (spacemacs/default-pop-shell)
  (message "Not in a project directory")
  (setq-local default-directory backup)
  (spacemacs/default-pop-shell))
```

lebensterben (changes requested):

layers/+tools/shell/funcs.el
ryanprior: I decided instead to use `projectile-acquire-root`, which prompts the user to specify a project root when they call `spacemacs/projectile-pop-shell` outside a project directory. This matches the behavior of `projectile-run-shell` and cohorts, so will hopefully be least surprising.

layers/+tools/shell/README.org
```diff
@@ -287,22 +287,23 @@ Some advanced configuration is setup for =eshell= in this layer:
...
+| ~SPC "​~       | Open external terminal emulator in current directory               |
+| ~SPC p '​~     | Pop-up your default shell in the project root                      |
+| ~SPC p "​~     | Open external terminal emulator in project root                    |
+| ~SPC a t s e~ | Pop-up, close or visit an =eshell= buffer                          |
+| ~SPC a t s i~ | Pop-up, close or visit a =shell=  buffer                           |
+| ~SPC a t s m~ | Pop-up, close or visit a =multi-term= buffer                       |
+| ~SPC a t s t~ | Pop-up, close or visit an =ansi-term= buffer                       |
+| ~SPC a t s T~ | Pop-up, close or visit a =term= buffer                             |
+| ~SPC a t s v~ | Pop-up, close or visit a =vterm= buffer                            |
+| ~SPC p $~     | Create a default shell buffer in the project root and switch to it |
```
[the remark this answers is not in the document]
ryanprior: I've updated the organization & descriptions in this table.

layers/+tools/shell/packages.el
```diff
@@ -206,9 +206,8 @@
 
 (defun shell/post-init-projectile ()
   (spacemacs/set-leader-keys
-    "p'" 'spacemacs/projectile-shell-pop
-    "p$t" 'projectile-multi-term-in-root)
-  (spacemacs/declare-prefix "p$" "projects/shell"))
```
lebensterben, line 211: Do you accidentally removed `(spacemacs/declare-prefix "p$" "projects/shell")`

lebensterben (changes requested):

layers/+tools/shell/packages.el
ryanprior: This was intentional:`p$t` was the only leader key under prefix `p$`, so now that we have a command which uses the default shell (whether multi-term or other) so I removed this prefix and made leader keys `p$` run `spacemacs/projectile-shell`.

lebensterben: Oh right.

ryanprior:
Hi, just checking-in on this issue. Anything I can do to move it along?

lebensterben (approved):

lebensterben:
It's ready to merge.
Can you update the commit message

ryanprior:
I rewrote the history & commit messages, let me know if anything is confusing or needs another update. :)

lebensterben:

PR #14859, opened 2021-06-19
vc: use smerge-vc-next-conflict instead of smerge-next

michaelpj:
`smerge` recently learned `smerge-vc-next-conflict`, which will jump to the next file with a conflict in it if there are none remaining in the current file. This gives a nice experience, in particular it improves on the `magit` workflow since you no longer need to go back to the `magit` status buffer in order to find the next conflicted file, you can simply keep going to the next conflict.

I considered giving this a new keybinding, but in fact it subsumes the functionality of `smerge-next`, so we might as well just substitute it in.

While I was there I tweaked the help. I think we typically talk about "conflicts" in this context and "hunks" when staging. `smerge` uses "conflicts".

### DIFF
--- a/layers/+source-control/version-control/README.org
+++ b/layers/+source-control/version-control/README.org
@@ -211,12 +211,12 @@ Use ~SPC g .~ to enter a transient state for quickly navigating between hunks in
 ** Smerge Mode Transient-state
 Movement:
 
-| Key binding                | Description         |
-|----------------------------+---------------------|
-| ~SPC g r n~                | Next hunk           |
-| ~SPC g r N~ or ~SPC g r p~ | Previous hunk       |
-| ~SPC g r j~                | Go to next line     |
-| ~SPC g r k~                | Go to previous line |
+| Key binding                | Description                              |
+|----------------------------+------------------------------------------|
+| ~SPC g r n~                | Next conflict (possibly in another file) |
+| ~SPC g r N~ or ~SPC g r p~ | Previous conflict                        |
+| ~SPC g r j~                | Go to next line                          |
+| ~SPC g r k~                | Go to previous line                      |
 
 Merge Actions:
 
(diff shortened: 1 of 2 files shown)


layers/+source-control/version-control/README.org
```diff
@@ -211,12 +211,12 @@ Use ~SPC g .~ to enter a transient state for quickly navigating between hunks in
...
 
-| Key binding                | Description         |
-|----------------------------+---------------------|
-| ~SPC g r n~                | Next hunk           |
-| ~SPC g r N~ or ~SPC g r p~ | Previous hunk       |
-| ~SPC g r j~                | Go to next line     |
-| ~SPC g r k~                | Go to previous line |
+| Key binding                | Description                              |
+|----------------------------+------------------------------------------|
+| ~SPC g r n~                | Next conflict (possibly in another file) |
```
lebensterben, line 216: I think replacing `hunk` by `conflict` is fine at least, if it's not clearer.

But why do you added `(possibly in another file)`? And not adding that to `SPC g r N` or `SPC g r p`?

michaelpj: I wanted to highlight the fact that this command now may take you to another file. `n`/`P` is unchanged and only looks within the file.

michaelpj: Maybe I should add that to the transient mode hint as well...

lebensterben: I'm not sure about this.
Usually I prefer clarity, but does it really cause confusion if you didn't add `(possibly in another file)`?

michaelpj: Well, people might not discover that they can go between files otherwise. They'd just have to happen to hit `n` after the last conflict.

lebensterben:
https://github.com/emacs-mirror/emacs/commit/2eecaa28972320a1b8886ac8cde353c2a2f4aa44

smerge-vc-next-conflict requires Emacs > 27.0

Our current minimum Emacs version is 26.3.

So you need to add additional logic to prevent error.

smile13241324:
I have merged the original PR and fixed the open issues right away. 

Issue #14865, opened 2021-06-21
Error at init time after updating which-key package


bcc32:
#### Description :octocat:
Errors at init after updating which-key past a recent commit,
https://github.com/justbur/emacs-which-key/commit/a55b90844c837e157c289ad4b10f5f2e3a4d53ff

I don't know whether this is a bug in which-key or whether Spacemacs
is using a private interface that changed, so I reported it here.
Happy to open an issue in which-key instead if it turns out the change
there was broken or backward-incompatible.

#### Reproduction guide :beetle:
- Start Emacs
- Update packages and restart

*Observed behaviour:* :eyes: :broken_heart:
Lots of errors at init:

```
Error (use-package): auto-yasnippet/:init: Key sequence i S c starts with non-prefix key i S
Error (use-package): command-log-mode/:init: Key sequence a t l l starts with non-prefix key a t l
Error (use-package): counsel-spotify/:init: Key sequence a m s s T starts with non-prefix key a m s s
Error (use-package): emr/:config: Key sequence r d l starts with non-prefix key r d
Error (use-package): js-doc/:init: Key sequence r d b starts with non-prefix key r d
Error (use-package): js2-refactor/:init: Key sequence x m j starts with non-prefix key x m
Error (use-package): keycast/:init: Key sequence a t l k starts with non-prefix key a t l
Error (use-package): org/:init: Key sequence a o f i starts with non-prefix key a o f
Error (use-package): org-brain/:init: Key sequence B a c starts with non-prefix key B a
Error (use-package): org-download/:init: Key sequence i D y starts with non-prefix key i D
Error (use-package): magit/:config: replacement is neither a cons cell or a string
Warning (initialization): An error occurred while loading ‘/Users/aaron/.emacs.d/init.el’:
```

Also, layers are not properly initialized, so emacs just ends up
being thoroughly broken.

*Expected behaviour:* :heart: :smile:
No errors at init time.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 27.2
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. b0ef47a88)
- Graphic display: nil
- Distribution: spacemacs-base
- Editing style: hybrid
- Completion: ivy
- Layers:
```elisp
(spell-checking
 (syntax-checking :variables syntax-checking-use-original-bitmaps t)
 (auto-completion :variables auto-completion-idle-delay 1.0)
 (ivy :variables ivy-enable-advanced-buffer-information t)
[...]

PR #14869, opened 2021-06-23
Revert "Fix which-key entries: gr and gs in dired and magit"

duianto:
Fixes: Error at init time after updating which-key package #14865

This reverts commit ec57b21a924d88fa0f7e1d747f091c349cca0170.

A which-key change caused issues with this fix.
And the fix doesn't seem to be needed anymore.

The correct which-key entries appear without the fix.


fradav:
Thanks! Banged my head about the cryptic error `user-error : replacement is neither a cons cell or a string` on spacemacs init`.
This fixed it.

duianto:
@fradav Thanks for testing and confirming that it worked.

PR #14872, opened 2021-06-25
Add yasnippet-snippets-dir to yas-snippet-dirs in auto-compl layer

dalanicolai:
The auto-completion layer installs the yasnippet-snippets collection by default.
However, it does make the snippets available, which makes no sense.
This commit fixes it by simply making the snippets available.

### DIFF
--- a/layers/+completion/auto-completion/packages.el
+++ b/layers/+completion/auto-completion/packages.el
@@ -350,4 +350,7 @@
 
     :config (spacemacs|diminish yas-minor-mode " ⓨ" " y")))
 
-(defun auto-completion/init-yasnippet-snippets ())
+(defun auto-completion/init-yasnippet-snippets ()
+  (use-package yasnippet-snippets
+    :after yasnippet
+    :config (yasnippet-snippets-initialize)))


lebensterben:
Is it already added by the package `yasnippet-snippets-dir` upon loading/
See https://github.com/AndreaCrotti/yasnippet-snippets/blob/be823d7e1a1a46454d60a9f3dabb16b68b5dd853/yasnippet-snippets.el#L52

dalanicolai:
That looks like it does. So I have 'appropriately' updated the PR

lebensterben:
@dalanicolai 
To my understanding these two lines already added those snippets directory once `yasnippet` is loaded
https://github.com/AndreaCrotti/yasnippet-snippets/blob/be823d7e1a1a46454d60a9f3dabb16b68b5dd853/yasnippet-snippets.el#L74-L75

dalanicolai:
@lebensterben That is very sharp:). So yes it looks like it should, but for some reason it does not work. It even does not work when I only add `:after yasnippet`. I really have to run again the `(yasnippet-snippets-initialize)`.

I tried to debug it by adding `(debug-on-variable-change 'yas-snippet-dirs)` after the `(yasnippet-snippets-initialize)`, but it seems it does not get cahnged after the initialization. So I don't quickly understand why it does not work without the `:after` and the "extra" initialization.

dalanicolai:
Looking at it again, I find it is working correctly in a 'fresh` Spacemacs install (but I really don't know what causes this on my normal installation). Anyway... I am closing the issue...

PR #14876, opened 2021-06-28
Add (optional) ebib support to bibtex layer

dalanicolai:
Ebib is a nice new Emacs application to manage bibtex references.This PR adds optional (via a layer variable) installation and configuration of the ebib package.

Currently, it only adds the most basic configurations to make it fit into Spacemacs. Additional configurations are welcome but they are not essential.


arifer612 (changes requested):
Shouldn't `biblio` also have a toggle or use `bibtex-enable-ebib-support` since it is only used by `ebib-biblio`? Otherwise it would be installed even when the user does have any use for the package.

layers/+lang/bibtex/packages.el
```diff
@@ -28,7 +28,9 @@
     (ivy-bibtex :requires ivy)
     markdown-mode
     org
-    org-ref))
+    org-ref
+    biblio
```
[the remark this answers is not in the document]
dalanicolai: @arifer612 Thanks for the review. I deliberately chose to install `biblio` by default. But just as for all other packages, I refer to its docs for further instructions. I did not really emphasize that is gets installed (I just added it to the list of 'package references') because I think much of its functionality overlaps with the 'org-ref' package (I did not study in detail the features of `org-ref` and `biblio`). But for example, `biblio` provides a function to retrieve bibtex info by doi, which `org-ref` lacks, and as it is only a small package, I figured it does not hurt to install it by default and provide some extra useful features to the bibtex layer.

Also I assumed that in practice, most users like to import articles with a browser extension via JabRef, so I figured that users who really prefer to use biblio will find their way to its documentation (I already provided them with a link, which lacks for the other packages).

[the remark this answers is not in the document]
arifer612: `org-ref` also provides a means to retrieve bibtex info by doi through `doi-utils-add-bibtex-entry-from-doi` but from what I see, `biblio` does indeed provide more backends to search from like DBLP, HAL, and IEEE. The compatibility with browser extensions is a good read to include `biblio` as you mentioned. I think I'll look more into it then!

layers/+lang/bibtex/packages.el
```diff
@@ -110,3 +112,48 @@
         "ld" 'doi-utils-add-bibtex-entry-from-doi
         "li" 'isbn-to-bibtex
         "lp" 'pubmed-insert-bibtex-from-pmid))))
+
+(defun bibtex/init-ebib ()
+  (use-package ebib
+    :defer t
+    :init
+    (spacemacs/set-leader-keys "oe" 'ebib)
```
arifer612, line 120: I think the keybinding for this should be changed since "o" is a reserved binding for users, as laid out in [CONVENTIONS](https://github.com/syl20bnr/spacemacs/blob/develop/doc/CONVENTIONS.org#user-prefix). My suggestion would be to use "ate" since `ebib` is a tool, although "ae" is not a bad idea too since I don't remember seeing any layers with a binding to it.

arifer612 (changes requested):

layers/+lang/bibtex/packages.el
dalanicolai: Ah yes, good point. Indeed, this was my personal, private layer, keybinding. The keybinding "ae" [is reserved for "email" applications](https://github.com/syl20bnr/spacemacs/blob/64fed6f7daba47144b09403a4982f26eb79b8ec6/layers/%2Bspacemacs/spacemacs-defaults/keybindings.el#L36).

layers/+lang/bibtex/packages.el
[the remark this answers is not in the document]
dalanicolai: Interesting, somehow I could not find that doi function (probably I searched in the org-ref namespace), but I had noticed also the extra backends.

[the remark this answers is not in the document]
arifer612: After comparing the bibliography organisation that `org-ref` and `ebib` provide, I've come to learn that whilst `ebib` does everything that `org-ref` provides with a much prettier interface except for cross referencing. It would be good if we could 'turn off' one while the other is selected.

layers/+lang/bibtex/packages.el
arifer612: I see! Then the "ate" I suggested earlier might be the binding to go then.

layers/+lang/bibtex/packages.el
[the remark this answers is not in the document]
dalanicolai: Although I understand your point about providing either the one or the other, I don't understand what you mean with mixing components. But despite your idea sounds sensible (if the packages really are mostly two similar alternatives), it still does not seem a big issue to me if org-ref always gets installed by default. If users really don't want it, then they could add it to `dotspacemacs-excluded-packages`.

It is a little too much for me to read the full docs of `org-ref` and `ebib`, but on first sight it seems that org-ref is focused on `org-mode` while ebib is more general.

My initial intention for this PR is just to offer an easy Spacemacsey configuration for `ebib` to Spacemacs users. Personally, I had a separate private `ebib` layer. But for Spacemacs I figured it would be 'cleaner' to offer it as an option in the bibtex layer.

layers/+lang/bibtex/README.org
```diff
@@ -82,3 +117,24 @@ key binding added:
 | Key binding | Description     |
 |-------------+-----------------|
 | ~SPC m i c~ | Insert citation |
+
+*EBIB*
```
lebensterben, line 121: analogous to insert line / delete line, we can have
"o" and "d" for add / delete entry.

Evil collections have "d" and "D" for delete/kill entries. We can follow that.
https://github.com/emacs-evil/evil-collection/blob/6c97b0c18678fdb1d6370c30324bc8df9a8fb1e1/modes/ebib/evil-collection-ebib.el#L42-L43

layers/+lang/bibtex/README.org
```diff
@@ -82,3 +117,24 @@ key binding added:
...
+
+*Index buffer*
+| Key Binding     | Description                                           |
+|-----------------+-------------------------------------------------------|
+| ~J/K~           | scroll page down/up                                   |
+| ~SPC m j~       | jump to entry (searches only author and title fields) |
```
lebensterben, line 127: normally "jumping"-like functionalities are under "g" prefix.
so I suggest `SPC m g g`.

layers/+lang/bibtex/packages.el
```diff
@@ -110,3 +112,48 @@
...
+      "J" 'evil-scroll-page-down
+      "K" 'evil-scroll-page-up
+
+      "/" 'ebib-search
+      "n" 'ebib-search-next)
+
+    (spacemacs/set-leader-keys-for-major-mode 'ebib-index-mode
+      "j" 'ebib-jump-to-entry
+      "k" 'ebib-kill-entry
+      "b" 'biblio-lookup)
```
lebensterben, line 138: This is from `biblio` package. So it should not be configured in `ebib`.

layers/+lang/bibtex/packages.el
```diff
@@ -110,3 +112,48 @@
...
+
+    (evilified-state-evilify-map ebib-entry-mode-map
+      :mode ebib-entry-mode)
+
+    (evilified-state-evilify-map ebib-log-mode-map
+      :mode ebib-log-mode)
+
+    (use-package ebib-biblio
```
lebensterben, line 146: `ebib-biblio` is not a package. it's a feature of `ebib`.
So you should have `:after (biblio)` in `(use-package ebib)` instead.
Then in its `:config`, add `ebib-biblio-selection-import` key binding.

layers/+web/eww/packages.el
```diff
@@ -32,9 +32,9 @@
     :defer t
     :init
     (spacemacs//eww-setup-transient-state)
-    (spacemacs/declare-prefix "ate" "eww")
-    (spacemacs/set-leader-keys "atee" 'eww)
-    (spacemacs/set-leader-keys "atew" 'eww-switch-to-buffer)
+    (spacemacs/declare-prefix "awe" "eww")
```
lebensterben, line 35: you also need to update the commit message and documentation.

layers/+lang/bibtex/packages.el
```diff
@@ -110,3 +112,48 @@
...
+      :after (ebib biblio)
+      :config
+      (evilified-state-evilify-map biblio-selection-mode-map
+        :mode biblio-selection-mode
+        :bindings
+        "e" 'ebib-biblio-selection-import
+        (kbd "C-j") 'biblio--selection-next
```
lebensterben, line 153: these two `biblio` biindings should be added in `biblio` package's init function.

lebensterben (changes requested):

layers/+lang/bibtex/README.org
dalanicolai: This key binding is to insert a citation into org/latex/markdown buffer. I really think `, i` makes more sense here.
Also `d` is bound to delete entry already in the ebib index buffer. But it is probably handy to provide that information here also (instead of only in the ebib documentation / major-mode-describe).

layers/+lang/bibtex/README.org
dalanicolai: No need to prefix with `SPC m` I propose to add `g j` for jumping to an item and leave this keybinding here also (or you think I should remove the current one?)

layers/+lang/bibtex/packages.el
dalanicolai: But it is bound in the `ebib-index-mode` map, so if I would put it under biblio, we would have a similar argument. However, it really is ebib using biblio here as complementary package, not biblio using ebib(-index) as complementary package. So I really think it is better as it is (but other good arguments are welcome of course).

layers/+lang/bibtex/packages.el
dalanicolai: `ebib-biblio` really has to get loaded separately (see [here](https://joostkremers.github.io/ebib/ebib-manual.html#integration-with-the-biblio-package)). But indeed I could use the `:after (biblio)` in `(use-package ebib)` and then `(require ebib-biblio)` in the :config. But I think it does not make much of a difference (or maybe does it?)

layers/+lang/bibtex/packages.el
dalanicolai: I am not sure about the `ebib-biblio-selection-import` case, because that more is an ebib feature.
But for the `biblio--selection-next` I agree with you...

layers/+web/eww/packages.el
dalanicolai: I've updated the documentation, but I already put this in a separate commit, incl. commit message, for clarity of history (this PR has two commits)

dalanicolai:
I have removed the commit about migration to JabRef because its browser extension turns out not to be so good (I will probably stay with jabref myself, but don't really explicitly recommend it at the moment. It still is quite nice though)

arifer612 (changes requested):

layers/+web/eww/packages.el
dalanicolai: Regarding the commits, I prefer to have the change of eww keybindings in the primary commit message, therefore I prefer two commits.

layers/+lang/bibtex/README.org
```diff
@@ -82,3 +101,26 @@ key binding added:
...
 |-------------+-----------------|
 | ~SPC m i c~ | Insert citation |
+
+*EBIB*
+
+*Index buffer*
+| Key Binding | Description                                           |
+|-------------+-------------------------------------------------------|
+| ~J/K~       | scroll page down/up                                   |
+| ~g j~       | jump to entry (searches only author and title fields) |
```
lebensterben, line 111: you can also use `spacemacs|define-jump-handlers`. then user only needs to press `gd` and you don't need to manually bind the key.

(But the downside is it won't be able to receive universal argument. So you still need to manually bind `SPC u g j`.

layers/+lang/bibtex/packages.el
lebensterben: You're right.

layers/+lang/bibtex/packages.el
lebensterben: I think the upstream author is abusing `use-package`.
For example, Spacemacs has defined a `core-configuration-layer` feature. You can try `(use-package core-configuration-layer)`, since it's found in load path, it will succeed.

Basically, any feature found in load path can be declared as a package via `use-package`. But it's really not a package.

dalanicolai: Your argument for using the appropriate 'labels' for the various cases, makes sense. But it seems that, practically, it should not make much of a difference as `use-package` for this case simply expands to `require` (see [here](https://www.reddit.com/r/emacs/comments/643dkt/use_package_vs_require_and_maybe_some_sorrowful/dfzya9d?utm_source=share&utm_medium=web2x&context=3)). But it is fine for me to change the 'terminology'.

layers/+lang/bibtex/README.org
dalanicolai: It is a nice suggestion (also because I learn things), but here the jump is more an index search feature than a jump to definition feature (similar to swoop/swiper). It is not a jump to definition, because, I would argue that, that should jump to the .bib entry, while this jump only jumps to the 'index' entry in the index buffer. And as you mentioned already, it would be strange to have another, different, binding for the 'universal argument' only.

lebensterben:
@dalanicolai 
I'm about to merge this and want to know if you've anything to add to the PR, since a long time has passed.

dalanicolai:
As far as I remember, this PR was ready to get merged, otherwise I would have made the changes already. Thanks again!

Issue #14860, opened 2021-06-20
Link in "Spacemacs logo" section of README.md is broken


popzxc:
Readme states:

> For Ubuntu users, follow this guide to [change the logo in Unity](http://splendidabacus.com/posts/2015/03/spacemacs-unity-icon/).

However, this page seems to not exist anymore.

arifer612:
You can view what the link used to show from [an archive of the page](https://web.archive.org/web/20190128131716/http://splendidabacus.com/posts/2015/03/spacemacs-unity-icon/).

In short, it creates a .desktop file that starts Emacs using the Spacemacs logo instead. In summary:

Create a new file in `~/.local/share/applications/spacemacs.desktop` with the contents

    [Desktop Entry]
    Name=Spacemacs
    GenericName=Text Editor
    Comment=Edit text
    MimeType=text/english;text/plain;text/x-makefile;text/x-c++hdr;text/x-c++src;text/x-chdr;text/x-csrc;text/x-java;text/x-moc;text/x-pascal;text/x-tcl;text/x-tex;application/x-shellscript;text/x-c;text/x-c++;
    Exec=emacs %F
    Icon=/PATH/TO/.emacs.d/core/banners/img/spacemacs.png
    Type=Application
    Terminal=false
    Categories=Development;TextEditor;
    StartupWMClass=Emacs

where the icon path directs to your `.emacs.d` directory.

You can create a PR to fill in this information into the README file, just make sure to properly credit the original author of the page (@cpaulik).

PR #14871, opened 2021-06-24
Fix guide on logo for the Ubuntu users

popzxc:
Fixes #14860 

Replaced the dead link with the instructions mentioned in the issue.

PR #14877, opened 2021-06-28
[Readme] Fix link

arifer612:
Fix dead link. Addresses #14860.

This is meant to replace #14871 since the original author had closed the PR. Addressing the open change request by @lebensterben, I credited the author of the page that the original link referenced to because I feel that it is important to give credit where it is due.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -3834,6 +3834,7 @@ files (thanks to Daniel Nicolai)
   - README.md:
     - Fixed macOS documentation to install fonts (thanks to Sergio Morales)
     - Updated Windows section, suggest official Emacs (thanks to ghost-420)
+    - Fixed dead link for creating a Spacemacs.desktop file with the Spacemacs icon (thanks to cpaulik)
   - private/local/README.md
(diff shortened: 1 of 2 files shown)


dalanicolai:
Thanks for fixing. Just two small unnecessary comments. I would say these instructions are valid for most GNU/linux desktops in general. Also, I think the kudo's, although nice, are not so important here as it is only a small explanation for how to create a desktop file. I think it would fit better in the [CHANGELOG.develop file](https://develop.spacemacs.org/CONTRIBUTING.html#changelog-entry), because it would be too much if all small fixes like this get kudo's in the docs.
So if you like (and you agree) then you could update the PR, but otherwise, for me, it is fine like this also

arifer612:
Got it! Moving it to the CHANGELOG.develop file makes sense.

lebensterben:
merged. Thanks for your contributions.

arifer612:
Thanks for the merge! Can you also help to close #14860 since this PR resolves that issue?

lebensterben:
Closed now.

PR #14889, opened 2021-07-03
Fix face name typo: ahs-plugin-defalt-face

duianto:
problem:
Trying to switch the range in the `SPC h s`:
symbol highlight transient state

By pressing: `r`

Shows the message:
let*: Symbol’s value as variable is void: ahs-plugin-defalt-face

And the symbol highlight transient state can't be opened again,
until Emacs is restarted.

cause:
The typo in the face name: ahs-plugin-defalt-face
has been fixed upstream.


duianto:
This typo was fixed in the just merged PR: Fix symbol highlighting #14892

Issue #14880, opened 2021-06-30
How to disable auto highlighting of all matching words


lafith:
How to disable the following default highlighting of all matching words? I need to disable it globally, not just for a buffer. I prefer the normal cursor just like in vim.

![hh-2021-06-30_22 20 51](https://user-images.githubusercontent.com/39316548/124000947-bca1d400-d9f1-11eb-8437-58806ace35f2.gif)

lebensterben:
what theme?
are you using LSP?

korayal:
I think I was able to change to orange one via something like this:
```
  (set-face-attribute 'ahs-plugin-default-face nil
                      :background "#383838"
                      :foreground nil)
```
but I couldn't find the face used for other occurrences.

lebensterben:
if this only appears in LSP, then you can turn off semantic highlighting.

korayal:
@lebensterben turning off semantic highlighting doesn't affect this, I've already have it turned off. It's probably about something else.

lebensterben:
there's also symbol highlighting you can turn off in LSP.

korayal:
unfortunately `(setq lsp-enable-symbol-highlighting nil)` does not work for me, were there any other type of highlighting setting in LSP?

![image](https://user-images.githubusercontent.com/313801/124028688-f75d3980-d9fc-11eb-9ff5-e5b60497e05d.png)

lebensterben:
what if you add auto-highlight-symbol to exclude packages?

lafith:
> what theme?
> are you using LSP?

This is after the installation of spacemacs, no modifications. I have not included LSP and theme is the default one itself.

lafith:
> I think I was able to change it to something less annoying via something like this:
> 
> ```
>   (set-face-attribute 'ahs-plugin-default-face nil
>                       :background "#3f3f3f"
>                       :foreground "orange")
> 
>   (set-face-attribute 'ahs-face nil
>                       :background "#383838"
>                       :foreground "Coral3")
> 
>   (set-face-attribute 'ahs-face-unfocused nil
>                       :background "#383838"
>                       :foreground "Coral3")
> ```
> 
> ![image](https://user-images.githubusercontent.com/313801/124026812-d0056d00-d9fa-11eb-81ae-25a515e6dabc.png)
> 
> @lebensterben yes, ~I think this is caused by LSP.~ looks like this is caused by something else.

This worked, thanks! It is less irritating with this combination of colors.



[...]

PR #14892, opened 2021-07-03
Fix symbol highlighting

duianto:
Fixes: How to disable auto highlighting of all matching words #14880

Recent updates to the `auto-highlight-symbol` package,
broke the Spacemacs configurations for the package.

Problems:
- The symbols under the cursor gets highlighted on startup.
- The highlight isn't cleared when the Symbol Highlight Transient State is closed.
- The `SPC t h a` symbol highlighting toggle, shows the same message every time:
`automatic-symbol-highlight disabled.`
but the symbols are still being highlighted.

This PR fixes these issues above, but there are some issues with the toggling (see below).

There might be more issues, according to this upstream comment:
https://github.com/jcs-elpa/auto-highlight-symbol/issues/7#issuecomment-873092710
>iedit, isearch, and symbol-overlay is no longer working as expected

I have not been able to reproduce any issues with either `iedit`, `isearch` or `symbol-overlay`.
@alexalekseyenko Could you provide some reproduction steps?

### Toggling issues
There are some issues when toggling the highlighting.

The first time it's toggled on, then the minibuffer shows
a message that highlighting has been disabled.
It should be enabled the first time since the symbols weren't being highlighted before trying to toggle its state.

In the disabled state, the symbol highlight transient state key bindings:
`#`, `*`, `SPC s h` and `SPC s H` do open the transient state,
but when one tries to jump between the symbols,
then it shows this message:
`ahs-select: Wrong type argument: overlayp, nil`

Toggling to the enabled state, works as expected,
and it keeps highlighting when the cursor is moved.

But the highlight disappears when the transient state is closed.
If the symbol highlighting was enabled, then they should remain,
enabled after closing the TS.

The toggling should be fixed, but in this PR's current state,
it probably already is better than the current behavior.

### Notes:
This PR also includes the transient state exit fix:
from: `:before-exit`
to: `on-exit`
That's open in a separate PR:
Fix exit keyword in symbol-highlight TS #14890

### DIFF
--- a/layers/+spacemacs/spacemacs-navigation/funcs.el
+++ b/layers/+spacemacs/spacemacs-navigation/funcs.el
@@ -58,8 +58,8 @@ If the universal prefix argument is used then kill also the window."
   (interactive)
   (if spacemacs-last-ahs-highlight-p
       (progn (goto-char (nth 1 spacemacs-last-ahs-highlight-p))
-             (spacemacs/ahs-highlight-now-wrapper)
-             (spacemacs/symbol-highlight-transient-state/body))
+             (spacemacs/symbol-highlight-transient-state/body)
+             (spacemacs/ahs-highlight-now-wrapper))
     (message "No symbol has been searched for now.")))
 
 (defun spacemacs/integrate-evil-search (forward)
@@ -88,18 +88,9 @@ If the universal prefix argument is used then kill also the window."
               evil-ex-substitute-pattern `(,(concat isearch-string "\\C")
                                            nil (0 0))))
 
-(defun spacemacs/ensure-ahs-enabled-locally ()
-  "Ensures ahs is enabled for the local buffer."
-  (unless
-      (bound-and-true-p ahs-mode-line)
-    (auto-highlight-symbol-mode)
-    ))
-
 (defun spacemacs/ahs-highlight-now-wrapper ()
   "Safe wrapper for ahs-highlight-now"
-  (eval '(progn
-           (spacemacs/ensure-ahs-enabled-locally)
-           (ahs-highlight-now)) nil))
+  (eval '(ahs-highlight-now) nil))
 
 (defun spacemacs/enter-ahs-forward ()
   "Go to the next occurrence of symbol under point with
@@ -133,23 +124,23 @@ If the universal prefix argument is used then kill also the window."
   (if (eq forward spacemacs--ahs-searching-forward)
       (progn
         (spacemacs/integrate-evil-search t)
-        (spacemacs/ahs-highlight-now-wrapper)
         (evil-set-jump)
         (spacemacs/symbol-highlight-transient-state/body)
+        (spacemacs/ahs-highlight-now-wrapper)
         (ahs-forward))
     (progn
       (spacemacs/integrate-evil-search nil)
-      (spacemacs/ahs-highlight-now-wrapper)
       (evil-set-jump)
       (spacemacs/symbol-highlight-transient-state/body)
+      (spacemacs/ahs-highlight-now-wrapper)
       (ahs-backward))))
 
 (defun spacemacs/symbol-highlight ()
   "Highlight the symbol under point with `auto-highlight-symbol'."
   (interactive)
-  (spacemacs/ahs-highlight-now-wrapper)
   (setq spacemacs-last-ahs-highlight-p (ahs-highlight-p))
   (spacemacs/symbol-highlight-transient-state/body)
+  (spacemacs/ahs-highlight-now-wrapper)
   (spacemacs/integrate-evil-search t))
 
 (defvar-local spacemacs//ahs-was-disabled t)
--- a/layers/+spacemacs/spacemacs-navigation/funcs.el
+++ b/layers/+spacemacs/spacemacs-navigation/funcs.el
@@ -59,7 +59,7 @@ If the universal prefix argument is used then kill also the window."
   (if spacemacs-last-ahs-highlight-p
       (progn (goto-char (nth 1 spacemacs-last-ahs-highlight-p))
              (spacemacs/symbol-highlight-transient-state/body)
-             (spacemacs/ahs-highlight-now-wrapper))
+             (ahs-highlight-now))
     (message "No symbol has been searched for now.")))
 
 (defun spacemacs/integrate-evil-search (forward)
@@ -87,10 +87,6 @@ If the universal prefix argument is used then kill also the window."
         (setq evil-ex-last-was-search nil
               evil-ex-substitute-pattern `(,(concat isearch-string "\\C")
                                            nil (0 0))))
-
-(defun spacemacs/ahs-highlight-now-wrapper ()
-  "Safe wrapper for ahs-highlight-now"
-  (eval '(ahs-highlight-now) nil))
 
 (defun spacemacs/enter-ahs-forward ()
   "Go to the next occurrence of symbol under point with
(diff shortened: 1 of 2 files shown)


lebensterben:
I merged the smaller fix #14890 for exit keyword.

madand:
@duianto due to https://github.com/jcs-elpa/auto-highlight-symbol/commit/d9b83fa766078533ee712ca50cb49f0487b3a18a you need to replace `ahs-plugin-defalt-face` with `ahs-plugin-default-face` here:
https://github.com/syl20bnr/spacemacs/blob/55f1d56d8942763fe13cb5d00cf5725051e062f4/layers/%2Bspacemacs/spacemacs-navigation/funcs.el#L179 

Also, to make the highlighting slightly less annoying for folks on Spacemacs theme you might add these lines:
```elisp
     `(ahs-plugin-default-face ((,class (:background ,highlight))))
     `(ahs-plugin-default-face-unfocused ((,class (:background ,highlight))))
```
here: 
https://github.com/syl20bnr/spacemacs/blob/be74dd786f9f9b029ba9d8f472e1a18442f44303/core/libs/spacemacs-theme/spacemacs-common.el#L216-L218

duianto:
>you need to replace ahs-plugin-defalt-face with ahs-plugin-default-face here:

There's a PR pending with that change:
Fix face name typo: ahs-plugin-defalt-face #14889

>Also, to make the highlighting slightly less annoying for folks on Spacemacs theme you might add these lines

The updates to the `spacemacs-theme` are pulled from the upstream package:
https://github.com/nashamri/spacemacs-theme

If I remember correctly, another maintainer said that, 
Spacemacs has a local copy of the `spacemacs-theme`,
to make sure that the theme loads correctly on startup.

An issue has been opened upstream about the new ahs face:
auto-highlight-symbol, new default face, hard to see cursor https://github.com/nashamri/spacemacs-theme/issues/186

Could you post the suggestion there.
A screenshot would also be helpful.

madand:
I created a PR with screenshots at  nashamri/spacemacs-theme#187. Should I also open the same PR against the Spacemacs repo?

Since the upstream broke rather important part of Spacemacs' toolkit, how about temporarily including `auto-highlight-symbol` as a local package with the breaking commits reverted?

alexalekseyenko:
@madand
Yes, that would be both efficient and logical. Unfortunately it looks like people have decided to enshrine the new logic, making it more difficult for the maintainer to roll back the recent changes if they decide that's a good idea. I'm still not sure why performance isn't a consideration here or how this is supposed to work together with LSP's semantic highlighting, but I don't currently have the time to work on this, so if you have ideas for how to prevent Spacemacs from accreting cruft, I think future generations would appreciate it.

duianto:
>Should I also open the same PR against the Spacemacs repo?

The `spacemacs-theme` repo has merged your changes.
And the SpacemacsBot saw the updates, and opened a PR with those changes.
They have now been merged.

---

I've almost managed to fix the symbol highlighting issues.

The highlighting is disabled on startup.
And it works to toggle the auto highlighting on and off.

### But an issue remains
The symbol highlighting stays enabled

The first time the `Symbol Highlight Transient State` is closed.
After highlighting a symbol with either of these keys:
```
 #                      ;; spacemacs/enter-ahs-backward
 *                      ;; spacemacs/enter-ahs-forward
 SPC s h                ;; spacemacs/symbol-highlight
 SPC s H                ;; spacemacs/goto-last-searched-ahs-symbol
```
The auto highlighting can be disabled with:
```
 SPC t h a              ;; spacemacs/toggle-automatic-symbol-highlight
```
(the auto highlighting can be toggled on and off from startup, as expected)

From then on, it seems to be working as expected (until Emacs is restarted).
The symbol highlighting is disabled when the Transient State is closed.
(`Esc`, `C-g`, or by pressing a key that isn't defined in the Transient State.)

It's probably a logic problem, that I'm having trouble figuring out.
Any help would be appreciated.

This was tested with the following example:
`ahs-testing.el`
```
abc-def-ghi
abc-def-ghi
abc-def-ghi
jkl-mno-pqr
jkl-mno-pqr
jkl-mno-pqr
```

trishume:
I'm looking into the remaining bug you mention, and it appears to caused by `(bound-and-true-p auto-highlight-symbol-mode)` returning `t` after startup even when `auto-highlight-symbol-mode` is clearly not active. I'm looking for a better way to check for it.

trishume:
Okay I finished up this branch and tested the cases I could think of and things seem to work: https://github.com/trishume/spacemacs/commit/cfd849e6b9e0ea3ee1192937255cc3919e57a2c6

I fixed the issue that was causing the noted remaining bug, where ahs-mode was being enabled before the check to see if it was already active.

There was also some remaining weirdness around `SPC t h a` which I fixed by switching the toggle to a mode-based one instead of messing with the internal timer, since the idea of this PR is to properly toggle the mode instead of keeping it enabled and messing with the timer.

Either @duianto can `git cherry-pick` my commit, or I can just open a new pull request with my commit and @duianto's which supersedes this one.

madand:
@trishume I've got the error with your patch: `setq: Symbol’s function definition is void: ahs-highlight-p`. 
`ahs-highlight-p` isn't an autoloaded function and seems to be called before the package has been loaded.

duianto:
>This is caused by the listed commands calling spacemacs/ahs-highlight-now-wrapper which enables auto-highlight-symbol-mode before calling the transient state.

Your right. Great find.

>There was also some remaining weirdness around SPC t h a which I fixed by switching the toggle to a mode-based one instead of messing with the internal timer, since the idea of this PR is to properly toggle the mode instead of keeping it enabled and messing with the timer.

Removing anything that isn't needed is great.

---

I'm seeing the same message as `madand`:
>setq: Symbol’s function definition is void: ahs-highlight-p

When pressing: `SPC s h` which calls `spacemacs/symbol-highlight`
on startup, before having highlighted any symbols with either `#`, `*` or by toggling on highlighting with `SPC t h a`.

It seems to be caused by the line:
```elisp
(setq spacemacs-last-ahs-highlight-p (ahs-highlight-p))
```
in the `spacemacs/symbol-highlight` function.
https://github.com/trishume/spacemacs/commit/cfd849e6b9e0ea3ee1192937255cc3919e57a2c6#diff-bf8baa844a18fe110196f8a1f3d27ef20bb918589786ca13ebd06953fb261503R141

duianto:
The `ahs-idle-timer` is set to `0` in the `:init` section of:
`spacemacs-navigation/init-auto-highlight-symbol`

Since the timer isn't used for toggling anymore,
then it might not need to be assigned.

madand:
Spacemacs fiddles really hard with the guts of the `auto-highlight-symbol` package. We were lucky that package was semi-abandoned all this time :smile: 
 I'll try to fix the error and will also cleanup and simplify the code involved in the next hour or two. Will use @trishume patch as a starting point.

madand:
Please test on your side: 18d49c58861cb28c7c17419b7b45d78f8b42c5cb

madand:
I've found one more issue. Immediately after activating the transient state (either via `SPC s h`, `*` or `#`) the matches counter in minibuffer says `[0/0]`. After using any transient state bindings, that keep it open, the counter shows correct numbers.

matches-0

duianto:
>Please test on your side: 18d49c5

It seems to be working now. Thanks.

---

Let's fix the `[0/0]` issue in new PR.

I'll merge this one so that the symbol highlighting can be fixed as soon as possible.

Thank you @trishume and @madand for the fixes and improvements.

duianto:
A PR has been opened with a possible fix for the `[0/0]` issue.
Fix symbol highlight ts counter #14913

duianto:
I think I've found a solution to the [0/0] counter issue.
Could someone test that it still works as expected?
Fix symbol highlight ts counter #14913

duianto:
The counter fix PR has been merged.
Thanks madand for testing and suggsting a fix.

PR #14902, opened 2021-07-06
[ivy] add open in other window for ivy-xref

thanhvg:
`M-o j`


duianto:

PR #14903, opened 2021-07-06
Evilify image-dired-thumbnail/display-image-mode-(map)s

dalanicolai:
Image-dired is a quite nice image browser, but its modes really should open in evilified state.


duianto:

PR #14904, opened 2021-07-06
Html export

JAremko:
Looks like CircleCI doesn't schedule cron jobs from the generated config.   Maybe it works from setup.. Lets seee.


JAremko:
I think CircleCI just silently died from seeing cron...  😄

JAremko:
What a meme...

PR #14913, opened 2021-07-08
Fix symbol highlight ts counter

duianto:
problem
The Symbol Highlight Transient State (`SPC s h`) counter
ex: [2/7]

Shows: [0/0]
when opening the TS, either with `SPC s h`, `SPC s H`
or when navigating between symbols with: `#` or `*`

cause
The symbols haven't been counted yet, when the TS is opened.

And the counter is reset when `auto-highlight-symbol-mode` is disabled,
this happens when the TS is restarted, which happens when navigating to the
next/previous symbol with `#` and `*`.

solution
Highlight the symbol before opening the TS.
And only disable `auto-highlight-symbol-mode` when the TS is closed.
(when the TS buffer `" *LV*"` doesn't exist)

notes
This also shows the "No previously searched for symbol found" message,
when the variable `spacemacs-last-ahs-highlight-p` is void on startup.

### DIFF
--- a/layers/+spacemacs/spacemacs-navigation/funcs.el
+++ b/layers/+spacemacs/spacemacs-navigation/funcs.el
@@ -56,11 +56,11 @@ If the universal prefix argument is used then kill also the window."
   "Go to the last known occurrence of the last symbol searched with
 `auto-highlight-symbol'."
   (interactive)
-  (if spacemacs-last-ahs-highlight-p
+  (if (bound-and-true-p spacemacs-last-ahs-highlight-p)
       (progn (goto-char (nth 1 spacemacs-last-ahs-highlight-p))
-             (spacemacs/symbol-highlight-transient-state/body)
-             (ahs-highlight-now))
-    (message "No symbol has been searched for now.")))
+             (spacemacs//ahs-setup)
+             (spacemacs/symbol-highlight-transient-state/body))
+    (message "No previously searched for symbol found")))
 
 (defun spacemacs/integrate-evil-search (forward)
         ;; isearch-string is last searched item.  Next time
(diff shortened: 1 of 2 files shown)


madand:
Works ok for me.

duianto:
The `ahs-current-overlay` variable was removed upstream.
https://github.com/jcs-elpa/auto-highlight-symbol/commit/4e02c92442d6131816caa0285ae8626cf4356c4b

It's used in the `symbol-highlight-doc` function.
https://github.com/syl20bnr/spacemacs/blob/00599d325565c22553388fcad06355775a5f79e6/layers/%2Bspacemacs/spacemacs-navigation/funcs.el#L180

This seems to cause an infinite loop,
when one opens the symbol highlight TS: SPC s h
and tries to navigate to the next/prev symbol with: n, N or p.

It can be canceled with `C-g` twice.

duianto:
There seems to be an `off by one` issue in the
Symbol Highlight Transient State counter: [current/total]
with or without this PR applied.

When navigating to the next/prev symbol with `#` or `*`.

The counter is correct when navigating with the 
Symbol Highlight Transient State keys: `n`, `N` or `p`

This can be seen without this PR,
by enabling automatic symbol highlighting: `SPC t h a`

Example:
```
abc
abc
abc
```
The Symbol Highlight Transient State, counter shows the following:

When navigating with: `#    ;; spacemacs/enter-ahs-backward`
and the cursor is on the:
- first line: [3/3]
- second line: [1/3]
- third line: [2/3]

When navigating with: `*    ;; spacemacs/enter-ahs-forward`
and the cursor is on the:
- first line: [2/3]
- second line: [3/3]
- third line: [1/3]

madand:
@duianto I just tested your latest commit and it needs one final fix:
In the body of `spacemacs//ahs-setup` function, replace
``` emacs-lisp
  (unless auto-highlight-symbol-mode
    (setq-local spacemacs//ahs-was-disabled t))
```
with 

``` emacs-lisp
  (unless (bound-and-true-p auto-highlight-symbol-mode)
    (setq-local spacemacs//ahs-was-disabled t))
```

With this fix, everything seems to work ok. Good job! :+1:

duianto:
>it needs one final fix

Thanks for testing.

PR #14917, opened 2021-07-10
Add option to enable all-the-icons-ivy-rich to ivy layer

dalanicolai:
This PR adds an option to configure =all-the-icons-ivy-rich=.

I am not sure if I should add extra logic to make sure that the various required modes get activated in the right order as instructed in [these notes](https://github.com/seagle0128/all-the-icons-ivy-rich#use-package).

Currently, the `all-the-icons-ivy-rich-mode` seems to get activated before `ivy-rich-mode` as instructed in the third point of the notes.
It all seems to work fine (also, I think it will not break anything if activated in the wrong order, just it will not work correctly).

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2478,6 +2478,7 @@ Other:
 - Fixed =cl= package deprecated =letf*= (thanks to duianto)
 - Fixed counsel find-file functions to use Spacemacs check for opening large
 files (thanks to Daniel Nicolai)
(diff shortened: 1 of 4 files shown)


smile13241324 (approved):
Looks good, I think we can keep the ordering for now. If some modes are discovered where it not work properly we can fix them one by one. Right now I would not know which modes to load before this ivy extension.

PR #14924, opened 2021-07-12
[org-mode]/[org-brain] Add flag to enable org-brain support making it optional

WarFox:
Make org-brain optional using `org-enable-org-brain-support` flag

This is similar to what we have for other packages. 

Reason for this change:
- Give a choice for users who don't use `org-brain`


duianto:

PR #14927, opened 2021-07-16
Fix typos in .spacemacs.template

feltcat:
Fix some typos and also change the few instances of "non nil" to
"non-nil", so as to be consistent with the rest of the file.


Kazark (approved):

duianto:

PR #14930, opened 2021-07-18
Rename commands for org-roam v2 compatibility

### LINKED ISSUE #14929: org layer not fully compatible with org-roam v2
#### Description :octocat:
org-roam v2 replaced org-roam v1 on July 17.

- Start Emacs
- "SPC a o r f"

"Wrong type argument: org-roam-find-file"

org-roam notes browser should open.

The new org-roam does away with files and now works with "nodes". The relevant command is called "org-roam-node-find".

There are probably other incompatibilities.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 27.2
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. ba8074b60)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
((auto-completion :variables auto-completion-enable-snippets-in-popup t auto-completion-enable-sort-by-usage t)
 bm dap dtrt-indent git gtags helm helpful imenu-list lsp multiple-cursors
 (spell-checking :variables spell-checking-enable-auto-dictionary t spell-checking-enable-by-default nil)
 syntax-checking templates treemacs typography version-control unicode-fonts ibuffer colors command-log copy-as-format dash debug deft
 (org :variables org-enable-org-journal-support t org-journal-enable-agenda-integration t org-journal-dir "~/org/journal" org-enable-roam-support t org-projectile-file "TODO.org")
 outshine pandoc pdf ansible asm c-c++ csv docker emacs-lisp go graphql graphviz html javascript json markdown nginx perl5 raku plantuml python rust scheme shell-scripts terraform yaml kappa olivetti)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG CAIRO SOUND DBUS GSETTINGS GLIB NOTIFY INOTIFY ACL LIBSELINUX GNUTLS LIBXML2 FREETYPE HARFBUZZ M17N_FLT LIBOTF ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XDBE XIM MODULES THREADS XWIDGETS LIBSYSTEMD JSON PDUMPER GMP

kappa:
Fixes #14929

The main org-roam command "org-roam-find-file" was replaced with "org-roam-node-find". I decided to leave 
the old key sequence "a o r f" as an alternative for the new command alongside "a o r n" so that people who don't care about the internal naming of org-roam objects could keep using their muscle memory.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -3038,6 +3038,7 @@ files (thanks to Daniel Nicolai)
   =org-insert-drawer=, =org-insert-heading=, =org-insert-item= and
   =org-insert-structure-template= commands (thanks to Andriy Kmit')
 - Added =ox-asciidoc= as org export backend (thanks to Christian "West" Westrom)
+- Fix org-roam v2 compatibility (thanks to Alex Kapranoff and Caleb Rogers)
 **** Osx
 - Key bindings:
   - Added key bindings to use ~command-1..9~ for selecting window
@@ -3850,4 +3851,3 @@ files (thanks to Daniel Nicolai)
   Thanks to: Abhishek(Compro) Prasad, Deepak Khidia, Enze Chi, Grant Shangreaux,
   Igor Almeida, Jiahao Jiang, Miciah Dashiel Butler Masters, Songpeng Zu, Ward
   Harris
-  
--- a/layers/+emacs/org/README.org
+++ b/layers/+emacs/org/README.org
@@ -464,7 +464,7 @@ To install support for [[https://github.com/org-roam/org-roam-server][org-roam-s
 to =t=.
 
 *** Org-roam-protocol support
-To enable support for [[https://www.orgroam.com/manual.html#Roam-Protocol][Org Roam Protocol]] set the variable
+To enable support for [[https://www.orgroam.com/manual.html#Org_002droam-Protocol][Org Roam Protocol]] set the variable
 =org-enable-roam-protocol= to =t=.
 
 #+BEGIN_SRC emacs-lisp
@@ -473,7 +473,7 @@ To enable support for [[https://www.orgroam.com/manual.html#Roam-Protocol][Org R
          org-enable-roam-protocol t)))
 #+END_SRC
 
-And create a desktop file as described in the [[https://www.orgroam.com/manual.html#Roam-Protocol][org-roam manual]].
+And create a desktop file as described in the [[https://www.orgroam.com/manual.html#Org_002droam-Protocol][org-roam manual]].
 
 ** Mode line support
 To temporarily enable mode line display of org clock, press ~SPC t m c~.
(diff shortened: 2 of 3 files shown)


snuffop:
:+1:

emacs18:
`org-roam-tag-delete` also seems to have been removed.
I see `org-roam-tag-remove` instead.

emacs18:
`READEME.org` should also be updated, i.e., the key bindings at the end of the file.

Claude-Ray:
The same goes for `org-roam-dailies-find-*`. They are warned as obsolete since `org-roam-dailies-goto-*` have been created.

snuffop (approved):

emacs18:
It seems like key bindings on lnes 960 and 961 are incorrect, i.e.,
```diff
+        "rtr" 'org-roam-tag-add
+        "rtd" 'org-roam-tag-remove
```
In global key bindings, the same key commands are bound to "aorta" and "aortr".

emacs18:
Should `org-roam-setup` function be called to set things up?

magthe:
> Should `org-roam-setup` function be called to set things up?

The docs, https://www.orgroam.com/manual.html#Setting-up-Org_002droam, suggests it should be called, yes.

ZacheryPD:
Feel free to disregard, but would it also be a good idea to add a keybinding for org-roam-capture?

jason--liu:
Why hasn't this patch been merged yet? ^^

kappa:
> Feel free to disregard, but would it also be a good idea to add a keybinding for org-roam-capture?

Are there scenarios where you want org-roam-capture outside of adding a new node? And that one already has 2 keybindings -- "SPC a o r f/n".

duianto:
There are two org-roam fix PR's:
Rename commands for org-roam v2 compatibility #14930
Change org-roam keybindings to use new v2 format #14939

We should combine the changes into one.

A quick glance shows these differences.
- two node find key bindings in the first PR
```
        "aorf" 'org-roam-node-find
        "aorn" 'org-roam-node-find
```
the second PR only has the first.
- tag removal has two different key bindings
First PR:
```
"aortr" 'org-roam-tag-remove
```
Second PR:
```
"aortd" 'org-roam-tag-remove
```
- the function org/init-org-roam-server is removed in the second PR
```
(defun org/init-org-roam-server()
  (use-package org-roam-server
    :defer t
    :init
    (progn
      (spacemacs/set-leader-keys "aors" 'org-roam-server-mode)
      (spacemacs/set-leader-keys-for-major-mode 'org-mode "rs" 'org-roam-server-mode))))
```

KjartanOli:
> > Feel free to disregard, but would it also be a good idea to add a keybinding for org-roam-capture?
> 
> Are there scenarios where you want org-roam-capture outside of adding a new node? And that one already has 2 keybindings -- "SPC a o r f/n".

If I remember correctly there is a separate function `org-roam-capture` which behaves slightly differently from `org-roam-node-find`.

duianto:
Let's use this PR since it was opened first.

duianto:
@kappa can you add an entry in
https://github.com/syl20bnr/spacemacs/blob/develop/CHANGELOG.develop

That mentions both of you org-roam fix PR authors. Something like this:
```
- Fix org-roam v2 compatibility (thanks to Alex Kapranoff and Caleb Rogers)
```
at the end of the org section in:
`*** Layer changes and fixes`
`**** Org`
on line 3042, above the heading: `**** Osx`

Note, there's also a Org section in `*** Breaking Changes`, don't add it there.

ZacheryPD:
> > > Feel free to disregard, but would it also be a good idea to add a keybinding for org-roam-capture?
> 
> > 
> 
> > Are there scenarios where you want org-roam-capture outside of adding a new node? And that one already has 2 keybindings -- "SPC a o r f/n".
> 
> 
> 
> If I remember correctly there is a separate function `org-roam-capture` which behaves slightly differently from `org-roam-node-find`. 

This is correct. I'm new to roam so I may be mistaken but if I recall correctly capture can have templates like org-capture.

duianto:
@kappa After deciding if any of the other PR's changes should be incorporated.
Change org-roam keybindings to use new v2 format #14939

And if there are other changes that should be made.

Could you squash the commits into 1 when this is ready to review.
The recommendation is to keep simple PRs to one commit per PR.
https://github.com/syl20bnr/spacemacs/blob/develop/CONTRIBUTING.org#pull-request

kappa:
@duianto -- great instructions on squashing using magit, thank you!

I incorporated differences from @komali2's PR and added a new keybinding for org-roam-capture as suggested by @ZacheryPD.

duianto:
org/readme changes:
- Add an entry for the org-roam-capture key binding
- Remove the org-roam server key binding, since the package isn't used at the moment.
- Align the key binding table.
With the cursor in the table, press `TAB` to call `org-cycle`. It's docstring says:
>When inside a table, re-align the table and move to the next field.

duianto:

The readme has been updated.
[docs] Update org-roam key bindings https://github.com/syl20bnr/spacemacs/commit/3b8bb7f5b7305a04438d609036461f3ddbe1e6cd

The org switch buffer key binding was also removed.
And the alias key binding was moved up,
to group it with the other key bindings without a prefix.

Issue #14919, opened 2021-07-11
spacemacs/force-init-spacemacs-env is not executed in a fresh environment


lunalunaa:
I recently noticed that no matter how I execute `spacemacs/force-init-spacemacs-env`, some environment variables not present in my shell are always written to `.spacemacs.env`. I later realized that when this function is executing `env` from shell, it does not do so in a blank state as if `.spacemacs.env` were not loaded, but already carries all those variables from startup. So when emacs is executing `force-init-spacemacs-env`, if there are variables defined in `.spacemacs.env`, they will not disappear. 

In my opinion, this behavior defeats the purpose of this command. Of course this is easily solved by deleting .spacemacs.env and restart emacs, but (in my understanding) this command is supposed to produce this exact same result but without needing to restart.

duianto:
The following PRs changes have been applied to the `develop` branch:
fix issue #14919 (force-init-spacemacs-env fails) #14931

duianto:
@lunalunaa Can you test if it works as expected now. If it does, then you can close this issue.

lunalunaa:
Can confirm this is working as expected on my machine. Thank for fixing this!

PR #14931, opened 2021-07-18
fix issue #14919 (force-init-spacemacs-env fails)

dalanicolai:
Currently, when running the env shell-command, Emacs does so using its current
`process-environment`. Instead, we should run the command with the
`initial-environment`, which seems to fix issue #14919 (at least on GNU/linux).


duianto:

PR #14934, opened 2021-07-19
Fix ahs ts, open prompt

duianto:
problem
opening a prompt, ex: `M-x`
while the Symbol Highlight Transient State (`SPC s h`) is open,
leaves `auto-highlight-symbol-mode` enabled

cause
when the idle timer runs,
it tries to disable ahs mode in the prompt window

### DIFF
--- a/layers/+spacemacs/spacemacs-navigation/funcs.el
+++ b/layers/+spacemacs/spacemacs-navigation/funcs.el
@@ -151,31 +151,68 @@ before highlighting a symbol."
 when the Symbol Highlight Transient State is closed.
 If ahs mode was disabled before a symbol was highlighted.")
 
+(defun spacemacs//ahs-was-disabled-in-ahs-ts-exit-window-p ()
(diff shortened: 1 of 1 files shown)


thanhvg:
@duianto I have another problem probably your PR can fix too: while Symbol Highlight Transient State is on (`*`) invoking any transient command (`s`, `f`,  `/`) will leaves `auto-highlight-symbol-mode` enabled. 

What is the reason this critical PR is not merged yet?

duianto:
>I have another problem probably your PR can fix too: while Symbol Highlight Transient State is on (*) invoking any transient command (s, f, /) will leave auto-highlight-symbol-mode enabled.

Those cases are also covered by this PR.

>What is the reason this critical PR is not merged yet?

No one has reviewed it yet.

If your able to test it and can confirm that it works, then I can merge it.

thanhvg:
Tested, it works

duianto:

PR #14935, opened 2021-07-19
Use built-in get-buffer

duianto:


layers/+spacemacs/spacemacs-navigation/funcs.el
```diff
@@ -179,18 +179,12 @@ if it was disabled before a symbol was highlighted."
 
 (defun spacemacs//transient-state-buffer-title ()
   (let ((transient-state-buffer-name " *LV*"))
-    (when (spacemacs/buffer-exists transient-state-buffer-name)
+    (when (bufferp (get-buffer transient-state-buffer-name))
```
lebensterben, line 182: `get-buffer` calls `bufferp` already. So you don't need too call `bufferp` again.

```c
DEFUN ("get-buffer", Fget_buffer, Sget_buffer, 1, 1, 0,
       doc: /* Return the buffer named BUFFER-OR-NAME.
BUFFER-OR-NAME must be either a string or a buffer.  If BUFFER-OR-NAME
is a string and there is no buffer with that name, return nil.  If
BUFFER-OR-NAME is a buffer, return it as given.  */)
  (register Lisp_Object buffer_or_name)
{
  if (BUFFERP (buffer_or_name))
    return buffer_or_name;
  CHECK_STRING (buffer_or_name);

  return Fcdr (assoc_ignore_text_properties (buffer_or_name, Vbuffer_alist));
}
```

duianto: Your right, thanks for the correction.

smile13241324 (approved):

PR #14950, opened 2021-07-28
plantuml layer supports .puml extension

bostonaholic:
The title tells all!


duianto:

PR #14954, opened 2021-07-29
[java] improve binding for lsp-java

thanhvg:
just some small improvement to java layer. Removed some obsolete commands and added some. Tag @yyoncho for review

### DIFF
--- a/layers/+lang/java/README.org
+++ b/layers/+lang/java/README.org
@@ -111,40 +111,43 @@ at the root of your project first.
 
 | Key binding   | Description                              |
 |---------------+------------------------------------------|
-| ~SPC m g g~   | Go to definition                         |
-| ~SPC m g r~   | Find references                          |
-| ~SPC m g R~   | Peek references using ~lsp-ui~           |
-| ~SPC m g d~   | Goto type definition                     |
-| ~SPC m g a~   | Search type in project                   |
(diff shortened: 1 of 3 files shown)


layers/+lang/java/README.org
```diff
@@ -117,9 +117,12 @@ at the root of your project first.
 | ~SPC m g d~   | Goto type definition                     |
 | ~SPC m g a~   | Search type in project                   |
 | ~SPC m g A~   | Search type in project using ~lsp-ui~    |
+| ~SPC m g k k~ | Show type hierarchy                      |
```
yyoncho, line 120: tb is missing

yyoncho (approved):

layers/+lang/java/README.org
yyoncho: Also IMHO, type hierarchy should be under `,gh`

thanhvg: will add `t b`, `,gh` is used by call hierarchy

yyoncho: then `,gH`?

thanhvg: I'd prefer lower case letters. imo bindings having upper case letter is a pain to invoke.

smile13241324 (approved):
Looks good, I will merge it in a minute.

PR #14955, opened 2021-07-29
[defaults] Disable line numbers with `SPC t n n`

Kazark:
Toggles are great and all, but the trouble with the various toggles for line
numbers is that they force you to remember and think about the fact that there
are variant forms of line numbers. To beginners or forgetful people the various
combinations of behavior can be unintuitive here (e.g. toggling off relative
line numbers does nothing if you are actual in visual line numbers mode). Though
it doesn't fit exactly into the normal stuff we have for toggles, due to the
complexity of this situation I think it is good to just have a way to say "line
numbers be gone!"

Wasn't sure what keybinding to use, so used a double-tap mnemonic.

If this PR gets rejected, there are a couple minor fixes that should probably be
pulled out and submitted separately.

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/packages.el
+++ b/layers/+spacemacs/spacemacs-defaults/packages.el
@@ -287,18 +287,17 @@
 
       (spacemacs/declare-prefix "tn" "line-numbers")
 
-      ;; backwards compatibility of symbols:
-      ;; keep the spacemacs/toggle-line-numbers & friends around
       (spacemacs|add-toggle line-numbers
         :status (and (featurep 'display-line-numbers)
                      display-line-numbers-mode
-                     (eq display-line-numbers t))
+                     (eq display-line-numbers dotspacemacs-line-numbers))
         :on (prog1 (display-line-numbers-mode)
-              (setq display-line-numbers t))
+              (setq display-line-numbers dotspacemacs-line-numbers))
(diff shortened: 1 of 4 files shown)


duianto:
Maybe it would be better to use a variable to store the last used line number style.
Then the line numbers could be toggled on/off with one key binding: `SPC t n n`

Kazark:
@duianto what should we do if line numbers were turned off in `.spacemacs` and `SPC t n n` is invoked?

Kazark:
Oh, I see that absolute is configured via `t` and is therefore a sane default.

Kazark:
@duianto What turned on to be the most obvious solution was to make `SPC t n n` just toggle based on what was in your `.dotspacemacs`. What do you think?

layers/+spacemacs/spacemacs-defaults/packages.el
```diff
@@ -287,18 +287,18 @@
 
       (spacemacs/declare-prefix "tn" "line-numbers")
 
-      ;; backwards compatibility of symbols:
-      ;; keep the spacemacs/toggle-line-numbers & friends around
+      ;; Need a way to toggle without thinking about which mode you are in.
```
duianto, line 290: This comment can be removed.

Kazark:
@duianto removed

duianto:

PR #14957, opened 2021-07-30
Use breakpoint() for more Python versions

verhovsky:
follow up to #11490


duianto:

PR #14958, opened 2021-07-30
Add bindings for scrolling the message pane in tree-view

agenbite:
In standard notmuch-emacs, one can scroll back and forth the message
pane using `SPC`/`BKSPC`. Obviously, this is not the case in Spacemacs,
where `SPC` is reserved. There was no way to scroll the message pane in
this layer, so I added `M-d` and `M-u` wired to
`notmuch-tree-scroll-message-window` and
`notmuch-tree-scroll-message-window-back`.


duianto:

Issue #14947, opened 2021-07-26
[Feature Request] toggle pop-shell should remember its old path


c02y:
#### Description :octocat:
<<DESCRIPTION OF THE PROBLEM>>
When I use `SPC-'` to toggle pop-shell(I use the default eshell), it should remember its old path. 

For instance, I open a buffer from /path/a, after I execute `SPC-'` to activate the pop-shell, I change my PWD to `cd ../b`(this is the directory I execute commands in), then I use `SPC-'` again to hide the terminal window or just switch back to the buffer window, I activate it again using `SPC-'`, the history content is still there, but the PWD is changed according to the currently opened buffer that I execute `SPC-`', I believe it should remember its old path, or at least I can set it using a variable.
#### Reproduction guide :beetle:
- Start Emacs
- `SPC-'`
- change your PWD using `cd`
- `SPC-'` to hide the terminal or just switch to the buffer window without hiding the terminal window
- 'SPC-'` to active it again, check the path in the terminal, it is changed

*Observed behaviour:* :eyes: :broken_heart:
<<DESCRIPTION OF THE OBSERVED BEHAVIOUR>>

The PWD is changed according to the path of the opened buffer every time when I `SPC-`' 

*Expected behaviour:* :heart: :smile:
<<DESCRIPTION OF THE EXPECTED BEHAVIOUR>>

Remember its path when I toggle the pop-shell using `SPC-'`

#### System Info :computer:
- OS: gnu/linux
- Emacs: 28.0.50
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 431dfd5ad)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: ivy
- Layers:
```elisp
((auto-completion :variables auto-completion-enable-snippets-in-popup t auto-completion-enable-help-tooltip t auto-completion-enable-sort-by-usage t)
 better-defaults syntax-checking colors git
 (ivy :variables ivy-enable-advanced-buffer-information t)
 emacs-lisp markdown vimscript
 (rust :variables cargo-process-reload-on-modify t cargo-process--open-file-after-new t cargo-process--enable-rust-backtrace t cargo-process--command-flags "--color never" lsp-rust-server 'rls)
 yaml
 (treemacs :variables treemacs-position 'right treemacs-no-png-images t treemacs-use-scope-type 'Perspectives treemacs-display-in-side-window nil treemacs-show-cursor t)
 shell-scripts
[...]

PR #14959, opened 2021-07-30
Improve spacemacs/default-pop-shell docstring

dalanicolai:
Add instruction for controlling automatic change to working directory ('fixes' #14947)


duianto:

PR #14960, opened 2021-07-30
Fix google-translate functionality (',tkk' search fails)

dalanicolai:
This bug exists already since I started using Spacemacs about 1.5 year ago.
Because it was already reported upstreaim in
[this issue](https://github.com/atykhonov/google-translate/issues/52), I
implemented this fix only locally.


duianto:

PR #14963, opened 2021-08-01
[compleseus] add consult-line-multi

thanhvg:
consult-line-multi bindings: `spc s b` and `spc s B`

### DIFF
--- a/layers/+completion/compleseus/funcs.el
+++ b/layers/+completion/compleseus/funcs.el
@@ -62,6 +62,15 @@
                       (region-beginning) (region-end))
                    (thing-at-point 'symbol t))))
 
+(defun spacemacs/consult-line-multi ()
+  (interactive)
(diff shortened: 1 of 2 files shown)


duianto:

The last three lines had some extra indentation.
```elisp
(defun spacemacs/consult-line-multi ()
  (interactive)
  (consult-line-multi
   nil
   (if (region-active-p)
                     (buffer-substring-no-properties
                      (region-beginning) (region-end))
                   (thing-at-point 'symbol t))))
```

It's understandable since it matches the previous function above:
```elisp
(defun spacemacs/consult-line ()
  (interactive)
  (consult-line
   (if (region-active-p)
                     (buffer-substring-no-properties
                      (region-beginning) (region-end))
                   (thing-at-point 'symbol t))))
```

np, it's been fixed.

PR #14965, opened 2021-08-02
Add python-send-shell-with-output command

dalanicolai:
Currently when using any python-send-shell command (using either the python or ipython shell),
the output of the evaluation is not shown in the comint shell (see Emacs bug#49822).
This commit adds a simple function to send input and get the output printed in the comint process buffer.


duianto:

PR #14966, opened 2021-08-02
[shell-scripts] Add shell-scripts-format-on-save

bcc32:
Add a configuration option for formatting shell scripts using shfmt.

### DIFF
--- a/layers/+lang/shell-scripts/README.org
+++ b/layers/+lang/shell-scripts/README.org
@@ -4,7 +4,7 @@
 
 [[file:img/fish.png]]
 
-* Table of Contents                     :TOC_5_gh:noexport:
+* Table of Contents                                       :TOC_5_gh:noexport:
 - [[#description][Description]]
   - [[#features][Features:]]
 - [[#install][Install]]
@@ -30,6 +30,7 @@ Supported scripting files:
 - =Sh= scripts linting using [[https://www.shellcheck.net/][shellcheck]]
 - =Sh= scripts style checking using [[https://github.com/openstack-dev/bashate][bashate]]
 - Support for the [[https://langserver.org/][Language Server Protocol]] (experimental)
+- Automatic formatting via [[https://github.com/mvdan/sh][shfmt]]
 
 * Install
 ** Layer
--- a/layers/+lang/shell-scripts/README.org
+++ b/layers/+lang/shell-scripts/README.org
@@ -11,6 +11,7 @@
   - [[#layer][Layer]]
   - [[#linting][Linting]]
   - [[#style-checking][Style checking]]
+  - [[#format-on-save][Format on save]]
   - [[#backends][Backends]]
     - [[#shell-script-mode][Shell-script-mode]]
     - [[#lsp][LSP]]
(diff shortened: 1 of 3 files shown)


duianto:
Could you add a format on save section to the readme.

An example of how it could be written can be found in the dart layer:
https://github.com/syl20bnr/spacemacs/blob/develop/layers/+lang/dart/README.org#format-on-save-1

bcc32:
Sure thing, thanks for catching that.

duianto:

lebensterben:
We probably can also add another check when initializing this layer:

Should the program shfmt not be found on PATH, emacs's shfmt package doesn't need to be enabled. Or alternatively, throw a warning. This avoids any confusion later.

bcc32:
As it stands, emacs will display this message when saving and `shfmt` is not installed:

    Error: (file-missing "Searching for program" "No such file or directory" "shfmt")

Unfortunately, it is only displayed for a brief moment before the save message:

    Wrote /path/to/file

I think the `*Messages*` buffer is a reasonable place to have this error message appear, but if you prefer something different, feel free to edit or suggest a different mechanism.

lebensterben:
> it is only displayed for a brief moment before the save message

That's exactly how Emacs confused me every now and then.

`message` outputs to `*Message*` buffer and is not very visible.
`warn` by default opens in a new pop window. We can use `warn`, it's not a hard error so the rest part is unaffected.

bcc32:
Makes sense.  Do you want to do that in the layer (it would involve wrapping all the commands to provide the prefix arg `display-errors`)?  There is an upstream issue, purcell/emacs-reformatter#3, that seems like it would also address this problem by adding a custom variable that we could just document the existence of.

lebensterben:
@bcc32 
Just add a toggle to `shell-scripts/init-shfmt`

https://github.com/bcc32/spacemacs/blob/525e520520ee2547a28e069e2aa0e56374e55cac/layers/%2Blang/shell-scripts/packages.el#L103-L106

No need to load this package if `shfmt` is not found at runtime. And optionally issue a warning.
But meanwhile, as long as `shell-scripts-format-on-save` is non-nil, the `shfmt` package itself is downloaded. (This prevents constant uninstalling and reinstalling.)

bcc32:
Hmm, that makes sense, but what if the user sets `shfmt-command` to some non-default value?  Then, we'll have to duplicate some logic to figure out what the value of `shfmt-command` would be if we loaded the package, to determine if we should load the package.

Actually, I noticed that regular errors (i.e., the formatter ran and exited non-zero with some error messages) are displayed in a separate buffer, but "could not find formatter in $PATH" is a special error for emacs (i.e., `call-process-region` signals instead of returning non-zero).  So, maybe the right thing to do here is to change reformatter.el to display such errors in the error window.  I'm working on a new PR to do just that.

lebensterben:
great question. so `shfmt-command` should be a custom variable whose value is either one of known program names, or an arbitrary string, e.g. it's `:type` is `'(choice (const :tag "shfmt" "shfmt") (string :tag "Any other program")`

Then we test the existence of the program accordingly. That is, we test it only if it's one of the known program, and always pass if it's not.

purcell:
> Then we test the existence of the program accordingly. That is, we test it only if it's one of the known program, and always pass if it's not.

You can't reliably test for presence of the executable in advance without opening the buffer: if the user is using something like `envrc`/`direnv`, the PATH can vary from one buffer to another. So enabling/disabling certain features in response to presence of executables when you initialise a layer isn't necessarily practical.

PR #14967, opened 2021-08-03
Add .leex support for the html layer

yourpalal:
Register .leex files for web-mode in the html layer, and add .leex to the html lazy-install file extensions list.

.leex files are a variant of .eex files. They are used with the [Phoenix LiveView](https://hexdocs.pm/phoenix_live_view/assigns-eex.html#content) web framework.

### DIFF
--- a/layers/+lang/html/packages.el
+++ b/layers/+lang/html/packages.el
@@ -242,6 +242,7 @@
      ("\\.[gj]sp\\'"     . web-mode)
      ("\\.as[cp]x?\\'"   . web-mode)
      ("\\.eex\\'"        . web-mode)
+     ("\\.leex\\'"       . web-mode)
      ("\\.erb\\'"        . web-mode)
      ("\\.mustache\\'"   . web-mode)
      ("\\.handlebars\\'" . web-mode)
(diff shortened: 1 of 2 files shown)


lebensterben:
This should be feature-gated by a layer variable.
Because unfortunately, `.leex` is used by certain virus. We'd better not opening such files by default without user permission.

yourpalal:
@lebensterben interesting, I would be glad to implement the feature gate if you can give me a pointer of what it might look like. An example of an existing feature gate that I could use for inspiration would be perfect. If not, then some keywords to get me started would be really helpful too.

Looks like the virus in question is a ransomware thing that saves your encrypted files with a .leex extension. That said, I can already open .leex files right now with spacemacs, so I don't quite understand the security risk in enabling the html layer on these files.

lebensterben:
If you treat `.leex` just like other html files, all associated packages are also (auto)loaded.
It's possible certain packages parses/evaluates the buffer content in the background, and that's exploitable.

One proof-of-concept: https://github.com/lucky/bad_actor_poc
When LSP client expands the procedural macro in the background, the attack occurs.

---

To make the 'feature gate':

First define a layer config variable such as https://github.com/syl20bnr/spacemacs/blob/40ae5e2293c6edb5aed1c554ec6b825f24db45d8/layers/%2Bsource-control/git/config.el#L26-L27
, e.g. `html-enable-leex-support`.

Document it in `README.org` as well. https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Bsource-control/git/README.org#magit-delta

In `packages.el`, in `:init` section of `html/init-web-mode`, you can add something like

```elisp
(when html-enable-leex-support
  (setq auto-mode-alist
    (append '(("\\.leex\\'" . web-mode))
            auto-mode-alist)))
```

smile13241324:

PR #14977, opened 2021-08-06
[defaults] switch frame by displayed buffers names `SPC F s`

Kazark:
On my Linux box I often have at least four Emacs frames displayed in different
virtual desktops. Switching between them with `SPC F o` is disorienting because I
never know which one is next. I look in vain for a more targeted way to switch,
so I built one. The use of `completing-read` means it should integrate with
whatever completion framework is in use.

If you happen to have two frames up with the same buffers in them, this will not
help you discriminate between them. But that seems like an odd case; why would
you make multiple frames with exactly the same contents?

Used `s` as a mnemonic for switch, select or search (since it using
`completing-read`). I believe this is in line with Spacemacs conventions and it
didn't collide with anything existing.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -838,6 +838,7 @@ Other:
   - Added =string-edit= package to =spacemacs-editing= layer
     (thanks to Ivan Yonchovski)
 - Key bindings:
+  - Switch frames by displayed buffer names via ~SPC F s~ (thanks to Keith Pinson)
   - Toggle line numbers (i.e. toggle =display-line-numbers-mode=) via ~SPC t n n~
     without having to think about which variant of line numbers you have turned
     on (thanks to Keith Pinson)
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -2565,6 +2565,7 @@ Frame manipulation commands (start with ~F~):
 | ~SPC F o~   | cycle focus between frames                          |
 | ~SPC F O~   | open a dired buffer in another frame                |
 | ~SPC F n~   | create a new frame                                  |
+| ~SPC F s~   | switch to frame by displayed buffers names          |
 
 **** Emacs and Spacemacs files
 Convenient key bindings are located under the prefix ~SPC f e~ to quickly
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -133,6 +133,25 @@ If not in such a search box, fall back on `Custom-newline'."
 (defalias 'spacemacs/display-buffer-other-frame 'display-buffer-other-frame)
 (defalias 'spacemacs/find-file-and-replace-buffer 'find-alternate-file)
 
+(defun spacemacs/switch-frame-by-buffers ()
+  "Switch to frame selected by user based on names of displayed buffers."
+  (interactive)
+  (let ((frame-alist
+         (mapcar (lambda (frame)
+                   (cons (string-join (mapcar (lambda (window)
+                                                (buffer-name (window-buffer window)))
+                                              (window-list frame))
+                                      " | ")
+                         frame))
+                 (frame-list))))
+    (select-frame-set-input-focus
+     (cdr
+      (assoc (completing-read "Switch to frame: "
+                              frame-alist
+                              nil
+                              t)
+             frame-alist)))))
+
 (defun spacemacs/indent-region-or-buffer ()
   "Indent a region if selected, otherwise the whole buffer."
   (interactive)
--- a/layers/+spacemacs/spacemacs-defaults/keybindings.el
+++ b/layers/+spacemacs/spacemacs-defaults/keybindings.el
@@ -410,7 +410,8 @@
    ("B" spacemacs/display-buffer-other-frame "Display buffer other frame...")
    ("o" other-frame "Switch to other frame")
    ("O" spacemacs/dired-other-frame "Dired other frame...")
-   ("n" make-frame "Make frame"))))
+   ("n" make-frame "Make frame")
+   ("s" spacemacs/switch-frame-by-buffers "Switch frame by displayed buffer names") )))
 ;; help -----------------------------------------------------------------------
 (defalias 'emacs-tutorial 'help-with-tutorial)
 (spacemacs/set-leader-keys


lebensterben:
Each frame, window, or buffer, can be uniquely identified.
But buffer names are not unique. This means that it's possible to have multiple frames having the same collection of buffers and they're not uniquely identified in your PR.
This is the major problem.

A minor problem is the efficiency of string comparison.
This PR relies on `assoc` to retrieve the frame after `completion-read` from the `frame-alist`. `assoc` uses `string=` for string keys. Note that keys in `frame-alist` are concatenation of all buffer names, which can easily be very long. So string comparison is likely to be inefficient.

Kazark:
@lebensterben I implemented it at first with `helm-comp-read`, which didn't require the `assoc`; but I wanted it to be agnostic to completion framework, and be usable by Ivy and Compleseus folks too. I'm new to this area of Emacs Lisp; advice apprecaited.

Are you certain that buffer names are not unique? What I've observed in my use of this code is that if I have too files named `notes.org` open, they will get differentiated in the list by their directory name, like `notes.org<foo>` and `notes.org<bar>`.

Supposing that there is some case I am not thinking of here, where the names will not be thus distinct, I don't think it's fair to say this is a major problem. It's just a particular usecase for which this mode of frame switching is suboptimal. On the other hand, I don't know how it could be made optimal for that situation; if you had frames that were really so similar in what they were displaying, it's hard to say how you could distinguish them without either requiring some manual frame naming (which I am trying to avoid) or just by careful inspection of the buffers themselves. I would suggest that this usecase you bring up is a fairly hypothetic edge case, and that at most this is a minor problem.

Another note about the inefficiency of the string comparison is that there is a sort of natural boundedness for most cases on the length of these strings.
* My guess is that most people do not work with giant file names. Take Spacemacs itself as an example; one of the longer names may be `CHANGELOG.develop` (17; compare `funcs.el` 8, `keybindings.el` 14, `FAQ.org` 7, `packages.el` 11...), which in terms of modern computing power is still fairly short. I don't deny that there will be some file names like `UnsafePerformDastardlyIOProxyFactoryBeanDecoratorFacade.javaskell` out in the wild---but I consider that an edge case (those sorts of people tend to work in IDEs anyway in my experience).
* There is a natural limit to the number of windows one can have open in a frame (which is the limit on the number of buffer names per frame in this code). Of course, you could just keep splitting and splitting, but we're not trying to please the trolls here but the people who are actually trying to get work done. In my experience, even on a large external monitor, having 8 windows open in one frame is a *lot*. Most frequently I work with two or three.
* The number of frames is also naturally somewhat bounded. Of course, people using one one frame don't need this, and people using only two can just `SPC F o`. But even when I have more than three frames going, I feel like I am using a lot if I have 5 or 6.

Switching frames is a fairly top-level behavior; it's not like this is an algorithm we will be running as someone types. We are operating at the level where humans are unlikely in the majority of cases to discern any difference. Of course, I am open to any optimization suggestions you have, but I wouldn't have submitted the PR if I hadn't found this to be a nice PR that actually speeds up my workflow.

Suppose somebody is working with filenames that are somewhat long on average: say, 30 characters; and that they have a surprising number open per frame; say, 8. Then they will get frames identifiers that are 261 characters long. I do admit that this is suboptimal.

```
xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
```

Yeah, not great. That'll be off your screen for sure; it's not even so much the performance I'm worried about at that point but the usability at all. Nevertheless, if someone actually finds themselves doing that, _and_ using this feature, then they will have perspective on how it can be improved; whereas that seems speculative to me at this time. (Do we truncate buffer names? But then we raise the possibility of duplicates. Do we get clever about selecting only three buffer names to use? But what if those are not the buffers that will catch the user's eye? &c.)

My guess is most of the use cases here are people using file names that will average closer to a dozen characters depending on their zeal for brevity, and average two or 3 windows per frame, such that they will mostly be dealing with frame titles that are under 50 characters. These should fit easily on the screen, and not be a big deal for a modern processor, particularly if they are working with on average about half a dozen frames.

This is all to say I disagree that it is a "major problem". I am of course open to suggestions on how to improve this, still being far from an ELisp guru. OTOH if you want to close this and not merge it, I will not take offense; I'll just keep using it in my own [fork of Spacemacs](https://github.com/Kazark/spacemacs/) which is what I run off anyway.

lebensterben:
> @lebensterben I implemented it at first with `helm-comp-read`, which didn't require the `assoc`; but I wanted it to be agnostic to completion framework, and be usable by Ivy and Compleseus folks too. I'm new to this area of Emacs Lisp; advice apprecaited.

Helm uses both alist and hash table for different purposes. Surely if keys are short, it doesn't worth making a hash table.

> What I've observed in my use of this code is that if I have too files named `notes.org` open, they will get differentiated in the list by their directory name, like `notes.org<foo>` and `notes.org<bar>`.

What if directory name are the same?

> Supposing that there is some case I am not thinking of here, where the names will not be thus distinct, I don't think it's fair to say this is a major problem.

It is exactly a major problem that defeats the purpose of the new command, if emacs is not able to uniquely identify a frame to switch to.

> I don't know how it could be made optimal for that situation; if you had frames that were really so similar in what they were displaying, it's hard to say how you could distinguish them without either requiring some manual frame naming (which I am trying to avoid) or just by careful inspection of the buffers themselves. I would suggest that this usecase you bring up is a fairly hypothetic edge case, and that at most this is a minor problem.

I often have same files opened in different frames, where each frame is in a different virtual desktop.
- If it's Windows, there's no virtual desktop, so it's end of story.
- If it's macOS, there are unnamed virtual desktop. It's also screwed up.

But I don't care about these users.

What you can do is to utilize the standard set by EWMH.

First, for each frame, you can get its window ID with `(frame-parameter FRAME 'outer-window-id)`.

With this ID, you can get which virtual desktop its in with `xprop -id ID _NET_WM_DESKTOP`. This is an index, starts from 0. If names of virtual desktops are set, you can get that with `xprop -root _NET_DESKTOP_NAME` and get the human readable name for that desktop. Otherwise, you can just use the index.

In either cases, you have one more information to identify a frame, i.e. the virtual desktop.

So instead of `BUFFER [| BUFFER]...`, you can use `DESKTOP: BUFFER [| BUFFER]...` in `completing-read`.

>     * My guess is that most people do not work with giant file names. Take Spacemacs itself as an example; one of the longer names may be `CHANGELOG.develop`

If you're are Linux, take a look in `/usr/share/applications`. File names there easy exceeds 16 characters, which is longer than most people's password. And reverse domain name notation is very pervasive many areas.

For example, say you're debugging something like this:

In desktop 1 you have an Emacs frame with 2 of these files opened, where one of them is a correctly working one and another one is the one you're customizing but has bugs. Your current implementation yields:

`org.gnome.FontManager.desktop | org.gnome.FontViewer.desktop.bad`

And in desktop 2 you have another Emacs frame with another two files opened, the first one is the same, but the second one is an earlier version of the currently bugging file. And you have Firefox on the side to lookup the reference manual.

`org.gnome.FontManager.desktop | org.gnome.FontViewer.desktop.good`

Now what you have are two frames, each of which have very long names. Note that reverse domain name notation also makes it possible to have very long common prefix. With `string=`, you will end up traversing almost the whole string. The worst thing is, they are all the same. So the slow string comparison is futile. 

But if you attached the virtual desktop name or index, it might look like

`1: org.gnome.FontManager.desktop | org.gnome.FontViewer.desktop.bad`
`2: org.gnome.FontManager.desktop | org.gnome.FontViewer.desktop.good`

Two benefits:
- Even when the files opened in each frame are identical, as long as they're not in the same virtual desktop, they are unique.
- With virtual desktop name/index prefixed, you greatly optimized the key lookup.

Kazark:
> What if directory name are the same?

In my experience, it will show as much of the path as necessary to differentiate. Admittedly, this just converts the uniqueness problem into the length problem.

Kazark:
> if emacs is not able to uniquely identify a frame to switch to.

I think I see what you're saying---Emacs itself would get confused by the alist?

Kazark:
Thanks for the info on how to get workspace names; that was something that I wanted to do but wasn't sure how. We could degrade gracefully if not on Linux. What about Wayland though? Does `xprop` support that?

lebensterben:
@Kazark 

`xprop` works for `X11` only. But some `Wayland` desktop environment partially supports it. In general you should assume it doesn't work on `Wayland`.

Kazark:
I'm going to close this for now until I either get more time to work on it or I become more convinced that it is good in the present state. Thanks for the consideration.

PR #14983, opened 2021-08-10
Change show-all keybinding to outline-cycle-buffer in pdf layer

dalanicolai:
Currently there is no keybinding to fold the outline of a pdf. So let's replace
show-all because it is easier to remember a single keybinding for folding and
unfolding then having separate bindings for those (and consistent with
org-mode's global cycle keybinding)

### DIFF
--- a/layers/+readers/pdf/README.org
+++ b/layers/+readers/pdf/README.org
@@ -189,7 +189,7 @@ differently from the default Evil search. To go to the next match, use ~C-s~.
 
 | *Key binding* | *Description*                                         |
 |---------------+-------------------------------------------------------|
-| ~S-tab~       | Expand all trees                                      |
+| ~S-tab~       | Expand all trees (Cycle trees for >= Emacs 28)        |
 | ~RET~         | Follow link                                           |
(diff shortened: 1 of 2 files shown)


duianto:
Another reason to change the `show-all` command,
is because the first time `S-tab` is pressed, it says:
>‘show-all’ is an obsolete command (as of 25.1); use ‘outline-show-all’ instead.

However, `outline-cycle-buffer` is void.
There's no mention of `cycle` in:
`c:/Program Files/Emacs/x86_64/share/emacs/27.2/lisp/outline.el`

Were you thinking of `outshine-cycle-buffer`?
https://github.com/alphapapa/outshine/blob/9334b555aaf1426a9e405a57b80809a1615627b3/outshine.el#L1724

`outshine` is loaded in a separate Spacemacs layer with the same name.
We can probably not assume that everyone has it enabled.

dalanicolai:
Ah okay, I really did not expect this to be a new feature (introduced in Emacs 28). So I have updated the PR. And I will check my future PR's first in Emacs 27...

duianto:

PR #14994, opened 2021-08-16
Add jump to specific manual keybinding (SPC h j)

dalanicolai:
This PR is just a suggestion, because I personally frequently like to use this functionality (have it under `SPC h M`).
Of course, you are free to dismiss it...

This keybinding could also go under `SPC h M`, but that is currently taken by
`helm-switch-major-mode` (which seems a quite redundant command to me, as you
can switch major-modes directly using `M-x`/`SPC SPC`).

### DIFF
--- a/layers/+spacemacs/spacemacs-navigation/packages.el
+++ b/layers/+spacemacs/spacemacs-navigation/packages.el
@@ -31,7 +31,7 @@
         (view :location built-in)
         golden-ratio
         (grep :location built-in)
-        (info+ :location local)
+        (info+ :location (recipe :fetcher wiki))
         open-junk-file
         paradox
         restart-emacs
@@ -323,6 +323,7 @@
     :defer t
     :init
     (progn
+      (spacemacs/set-leader-keys "hj" 'info-display-manual)
       (setq Info-fontify-angle-bracketed-flag nil)
       (add-hook 'Info-mode-hook (lambda () (require 'info+))))))
 
--- a/layers/+spacemacs/spacemacs-navigation/README.org
+++ b/layers/+spacemacs/spacemacs-navigation/README.org
@@ -5,6 +5,7 @@
 * Table of Contents                     :TOC_5_gh:noexport:
 - [[#description][Description]]
   - [[#features][Features:]]
+- [[#key-bindings][Key bindings]]
 
 * Description
 This layer adds general navigation functions to all supported layers.
@@ -21,3 +22,8 @@ This layer adds general navigation functions to all supported layers.
 - Support for =paradox= a modern emacs package manager
 - Shortcuts for restarting =emacs=
 - Shortcuts for easily switching between windows
+
+* Key bindings
+| Key Binding | Description    |
+|-------------+----------------|
+| ~SPC h j~   | jump to manual |


layers/+spacemacs/spacemacs-navigation/packages.el
```diff
@@ -323,6 +323,7 @@
     :defer t
     :init
     (progn
+      (spacemacs/set-leader-keys "hj" 'info-display-manual)
```
lebensterben, line 326: It seems like `helm-info-woman` is a better choice if `helm` is being used.

Ivy supports Info nodes out-of-box but I'm not sure whether there already exists commands similar to those of helm's. In any case we can fall back to `info-display-manual`.

---

Also, I'm thinking about another possibility for key bindings:

|Command                |Current Keybinding|Proposed Keybinding|
|-----------------------|------------------|-------------------|
|helm-man-woman         |`SPC h m`         |`SPC h m`          |
|helm-switch-major-mode |`SPC h M`         |`SPC f m`          |
|helm-info-woman        |                  |`SPC h M`          |

`SPC f m` for switching the major mode seems more natural then `SPC h M`.
Alternatively, `SPC b M` is also a valid choice and probably better if we expect
more file-related keybindings to be added. (Say `SPC f m` for a hypothetical `file-modify`).

dalanicolai: Sorry, I have placed the binding on `SPC h j`. I was just commenting that it might also as well be bound under `SPC h M` (because the `helm-major-mode` thing is rather redundant)...

Also, the `info-display-manual` jumps to `info` files not to `man` files...

lebensterben: > Also, the info-display-manual jumps to info files not to man files...

I know, that what exactly helm-info-woman does.

dalanicolai: Sorry, I mean texinfo pages or whatever they are called (e.g. the emacs, calc, magit manual etc.)

I find it annoying to always have to navigate there via the `C-h i`

lebensterben: `helm-info-woman` jumps to info nodes, not man pages.

dalanicolai: Yes I see, but it is annoying to jump to nodes, because I do not know how their namer. Moreover, it seems that `helm-info-woman` includes only very few nodes (e.g. try to jump to `transient` `classes and methods`, or any `calc` node).
I would prefer to just jump to the manual (e.g. elisp, calc, magic), but `helm-info-woman` does not offer that functionality. Also, I put this binding under `SPC h j`, while you are mentioning only bindings ending with `m/M`. So, your suggestions and this PR are not interfering.

smile13241324 (approved):
I think binding this to `h j` is a good start, finding better choices depending on the current user's search tool should be something to do a next step/pr.

layers/+spacemacs/spacemacs-navigation/packages.el
dalanicolai: I am using Ivy b.t.w., but otherwise it sounds reasonable to use `woman`' instead of `man`, and to add the `info` binding. Once more, I find the `helm-switch-major-mode` binding quite redundant, and otherwise probably more logical under `SPC b M`.

lebensterben: agreed. just make a new PR to move it to SPC b

PR #14995, opened 2021-08-16
Evilify special-mode by default

### LINKED ISSUE #14988: Info isn't evil enough
This isn't a bug report, it's a request for discussion of an issue that's been frustrating to me.  The info mode is a good example for it, but affects other modes as well, specifically modes which like info mode are normally read only and thus use even in "normal" emacs use single key commands for navigation and other functions.

The problem is that info mode is partially evilified; enough so that we can use `j` `k` to move up and down, but `h` and `l` have their original non-evil meanings.  And this doesn't work very well with my finger muscle memory... in info mode `l` means go back to the last visited info node, and back is conceptually to the left, but of course in vi `l` means move to the right.  This causes and irritating level of cognitive dissonance for me and wastes a lot of time as I keep going back nodes when my fingers were just trying to move the cursor out of the way or I end up in a help window.  For a while I tried to make it a habit of switching to holy mode when in such a buffer to avoid this, but I don't really want to give up `j` and `k` for movement either.

What I'm getting at is that I think we need define a core set of vi commands that always work the vi way if a mode is even partially evilified, specifically the core movements and searches, and moves all conflicting keybindings from those keys onto a predictable alternate binding, maybe with a leader key that stands for "original binding".

What do y'all think?

dalanicolai:
Due to the `org-roam v2` warning buffer, I realized that special modes do not get `evilified` by default.
As I am not aware of any case where this is not desired, I created this PR as a suggestion.

It change should probably get documented also (before getting accepted), but I would be happy to fix that.

As special-mode is a read-only mode, it could be evilified by default.
I am not aware of any case where this is not desired.

Closes #14988

### DIFF
--- a/layers/+distributions/spacemacs-bootstrap/packages.el
+++ b/layers/+distributions/spacemacs-bootstrap/packages.el
@@ -585,7 +585,9 @@ Press \\[which-key-toggle-persistent] to hide."
 ;; pre packages
 
 (defun spacemacs-bootstrap/init-evil-evilified-state ()
-  (use-package evil-evilified-state)
+  (use-package evil-evilified-state
+    :config
+    (add-to-list 'evil-evilified-state-modes 'special-mode))
   (define-key evil-evilified-state-map (kbd dotspacemacs-leader-key)
     spacemacs-default-map))
 


lebensterben:
Evilifying all special-mode is very aggressive, because any non-file buffer theoretically should derive from special-mode. See: https://www.gnu.org/software/emacs/manual/html_node/elisp/Basic-Major-Modes.html

I'm not sure whether this will cause problem for specific modes, as `spacemacs-default-map` contains quite a lot key-bindings and many of which may not make sense for all special modes.

dalanicolai:
Theoretically yes, but what is the case practically? I know it is tricky, but if we can not find a case where it gives problems, then we could just do it (not having it causes many problems, because many buffers that should open evilified do not).

Also, I do not understand the relation between evilifying and the `spacemacs-default-map`. I thought evilified-state only remaps some keys (see https://develop.spacemacs.org/doc/CONVENTIONS.html#evilified-buffers)

lebensterben:
I'm talking about the very next line https://github.com/syl20bnr/spacemacs/blob/1abc3b56aed137955d41bbcfe090591caf11ea4e/layers/%2Bdistributions/spacemacs-bootstrap/packages.el#L591

All evilified nodes are registered with the SPC prefix.

For example, commands like renaming file name or deleting files in `SPC f` doesn't make any sense for special mode.

dalanicolai:
Ah okay. But currently `special mode` opens in `normal-state` so there is the same problem. Except that the keys that should work do not...

lebensterben (approved):
I've no strong objection to this PR.
But for future references, evilifying all `special-mode` is theoretically not ideal.

smile13241324 (approved):
Hmmm, thats a very aggressive mode. I have played a bit around with it locally and did not run into issues. So lets give it a try and fix individual modes if this breaks anything.

dalanicolai:
@smile Do you mean aggressive state? Anyway, I thought it only changes keybindings. And as far as I know currently the special-mode's are opened in **normal state** (if that is not some local bug here), which changes the keybindings even more aggressively. So I do not really expect any problems, but I agree it might be tricky. We could also show a warning about it at startup (to make users 'aware' of the "experiment").

lebensterben:
@dalanicolai 
Let's just wait and see any issues occurs and fix it when necessary.
I don't think it's very likely. But that's what make it annoying because once it happens we might all forgot this PR.

PR #14997, opened 2021-08-16
Fix file rename target path when buffer name is uniquified

bcc32:
When the buffer's name is uniquified (because multiple buffers have
the same short name), it is no longer equal to the basename of the
buffer's file name, so it is incorrect to use the buffer name in the
rename operation.


lebensterben:
good catch

lebensterben (approved):

smile13241324 (approved):
Looks good, I'll merge it in a minute.

PR #15004, opened 2021-08-21
Fix macOS GUI full screen button wreck caused by GUI setup in early-i…

cjf-666:
Fix macOS GUI full screen button wreck caused by GUI setup in early-init.el


cjf-666:
Disabling GUI elements directly can cause problem, see discussion in https://github.com/syl20bnr/spacemacs/pull/14559

smile13241324 (approved):
Looks good, I have tested this on Linux this worked pretty well.

Issue #14969, opened 2021-08-03
org-roam-mode buffer broken due to (global-)page-break-lines-mode (Spacemacs continuously switches major-mode)


dalanicolai:
#### Description :octocat:
org-roam-mode buffer is broken due to (global-)page-break-lines-mode

#### Reproduction guide :beetle:
- Start Emacs

- Install org-roam via the [org-mode layer instructions](https://develop.spacemacs.org/layers/+emacs/org/README.html#org-roam-support)
- Additionally carry out the [setting up instructions of the org-mode manual](https://www.orgroam.com/manual.html#Setting-up-Org_002droam).

- Create a node using `M-x org-roam-node-find` followed by `C-c C-c`

- Within that note create another note but now using `M-x org-roam-node-insert` (and submit with `C-c C-c`)

- follow the link to the newly created note

- in the 'new note' buffer do `M-x org-roam-buffer-toggle`

There should be a backlink, otherwise there is a bug (which probably has not yet
been reported). If there is no backlink then customize the
`org-roam-mode-section-functions`. There settings are correct, but possibly have
to get 'reapplied`. Please note down as a bug.

Finally if there is a backlink then notice that you can not follow it by placing
the cursor on it and pressing enter. If you switch off
`global-page-break-lines-mode` before creating the org-mode buffer (delete and
toggle again), then the link does work (as already reported
[here](https://github.com/syl20bnr/spacemacs/issues/14929#issuecomment-887764025).

Now the `org-roam-mode` is derived from `special-mode` (via
`magit-section-mode-hook`) and, as none of these modes is included in
`page-break-lines-modes`, should be excluded from `global-page-break-lines-mode`.
However Spacemacs, somehow designates the `org-roam-mode` buffer as (over here)
 either an `emacs-lisp-mode` buffer or an `org-mode` buffer.

Now it took me a long time to find the culprit, and still help is appreciated.
So to see what is causing the problem evaluate the following lines:

```
(defun derived-mode-p (&rest modes)
  "Non-nil if the current major mode is derived from one of MODES.
Uses the `derived-mode-parent' property of the symbol to trace backwards."
  (apply #'provided-mode-derived-p (print major-mode) modes))
```

now just observe what happens. Spacemacs is switching modes continuously.

The `derived-mode-p` function is used in `page-break-lines-mode-maybe`.

So maybe for now we can indeed first switch-off `global-page-break-lines-mode`
for users of `org-roam-mode` (as suggested
[...]

PR #15005, opened 2021-08-22
[org-roam] Add "r" key to refresh org-roam links list buffer

bhipple:
This matches the org-agenda key, and is doubly important since we need to
refresh regularly to workaround the page-break-lines bug, as described here
https://github.com/syl20bnr/spacemacs/issues/14969#issuecomment-894624932


smile13241324 (approved):

Issue #14137, opened 2020-11-15
org-roam-insert should insert link AFTER the cursor in evil 


real-or-random:
#### Description :octocat:

When inserting a link via org-roam-insert, there's no space between the link and the preceding text.
This is very annoying  because inserting links is the core feature of org-roam and fixing the space is non-trivial: Simply moving left and inserting a space inserts the space into the link, which is not what you want.

This is essentially https://github.com/org-roam/org-roam/issues/4.
A workaround is available there.
We may want to apply a variant of this here.
(I have no idea how this is done with similar packages in evil mode.)

#### Reproduction guide :beetle:
- Start Emacs
- Open an org-roam file (e.g. org-roam-dailies-today)
- Enter insert state, type "bla"
- Enter normal state
- org-roam-insert a link to "file" 

*Observed behaviour:* :eyes: :broken_heart:
 - "blafile" (no space between text and link)

*Expected behaviour:* :heart: :smile:
 - "bla file" (with space)

#### System Info :computer:
- OS: gnu/linux
- Emacs: 27.1
- Spacemacs: 0.300.0
- Spacemacs branch: evil-set-undo-system-on-upgrade (rev. 3517cbcd3)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(csv html javascript systemd helm auto-completion better-defaults emacs-lisp git github markdown
     (org :variables org-enable-github-support t org-enable-roam-support t)
     spell-checking syntax-checking treemacs
     (latex :variables latex-enable-auto-fill nil)
     pdf yaml lua gtags
     (c-c++ :variables c-c++-backend 'lsp-ccls c-c++-lsp-enable-semantic-highlight 'rainbow)
     php python rust shell major-modes)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG CAIRO SOUND GPM DBUS GSETTINGS GLIB NOTIFY INOTIFY ACL GNUTLS LIBXML2 FREETYPE HARFBUZZ M17N_FLT LIBOTF ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XDBE XIM MODULES THREADS LIBSYSTEMD JSON PDUMPER LCMS2 GMP

emacs18:
I too was annoyed by the fact that the link was insert before space character at cursor rather than affter.
Following works for me to deal with this.
```elisp

[...]

PR #15008, opened 2021-08-22
[org-roam] Fix evil org-roam-node-insert spacing

bhipple:
When we're adding links, without this they get smooshed together under the
curser. Tested to work; resolves https://github.com/syl20bnr/spacemacs/issues/14137

### DIFF
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -971,7 +971,21 @@ Headline^^            Visit entry^^               Filter^^                    Da
         :mode org-roam-mode
         :bindings
         "o" 'link-hint-open-link
-        "r" 'org-roam-buffer-refresh))))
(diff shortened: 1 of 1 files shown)


lebensterben:
I'm okay with this but this should be `co-authored` by @emacs18 .

emacs18:
Please there is no need for extra steps on my account.
I probably copied someone else's code or at least got the idea from someone when I wrote my defadvice.
I think whoever spends the time to submit a PR has done a service to us all.

bhipple:
Happy to give sole credit to @emacs18, is there somewhere I should add something besides my comment linking to the issue where he posted his code?  I think we're of the same mind: I'm just glad org-roam linking works seamlessly now :sweat_smile:

lebensterben:
Merged. Thanks for both of you.

Issue #14392, opened 2021-02-18
OCaml layer: Changes with upcoming opam version 2.1


bschommer:
With the next opam version 2.1 `opam config var share` will no longer work and be replaced by `opam var share`. I think one can support older and newer versions of opam by testing both and only failing if both return no result.

smile13241324:
This sounds like changes would be necessary to the ocaml layer. As I am oblivious to ocaml I would ask the community to provide a PR for this. Support for both versions would be cool but not a must requirement in my opinion

bschommer:
Add the moment this isn't even a problem for users trying to use the new opam version since one can set the environment variable `OPAMCLI` to 2.0 to get the old behavior.

PR #15011, opened 2021-08-23
ocaml: Allow new OPAM command-line syntax

akkuankka:
As of OPAM v2.1, `opam config var share` is no longer supported. Instead, the syntax is `opam var share`. To allow both, this checks first with the new syntax, then if that doesn't work, the old is tried. If neither produce any output, it fails as normal.
This should resolve syl20bnr/spacemacs#14392 .

P.S. I'm pretty new to Elisp so criticism, constructive or otherwise is appreciated!

### DIFF
--- a/layers/+lang/ocaml/funcs.el
+++ b/layers/+lang/ocaml/funcs.el
@@ -23,26 +23,31 @@
 
 (defun spacemacs//init-ocaml-opam ()
   (if (executable-find "opam")
-      (let ((share (string-trim-right
-                    (with-output-to-string
-                      (with-current-buffer
-                          standard-output
-                        (process-file
-                         shell-file-name nil '(t nil) nil shell-command-switch
-                         "opam config var share"))))))
-        (cond ((string= "" share)
-               (spacemacs-buffer/warning
-                "\"opam config var share\" output empty string."))
-              ((not (file-directory-p share))
-               (spacemacs-buffer/warning
-                "opam share directory does not exist."))
-              (t (setq opam-share share
-                       opam-load-path (concat share "/emacs/site-lisp"))
-                 (add-to-list 'load-path opam-load-path))))
+        (let* ((share-dir-from-command (lambda (command)
+                (string-trim-right
+                 (with-output-to-string
+                   (with-current-buffer standard-output
+                                        (process-file shell-file-name nil
+                                        '(t nil)
+                                        nil
+                                        shell-command-switch
+                                        command))))))
+               (share-new (funcall share-dir-from-command "opam var share"))
+               (share (if (string= "" share-new)
+                          (funcall share-dir-from-command "opam config var share")
+                          share-new)))
+          (cond
+           ((string= "" share)
+            (spacemacs-buffer/warning "\"opam config var share\" and \"opam var share\" both output empty string."))
+           ((not (file-directory-p share))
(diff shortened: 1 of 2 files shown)


layers/+lang/ocaml/funcs.el
```diff
@@ -23,26 +23,30 @@
...
-        (cond ((string= "" share)
-               (spacemacs-buffer/warning
-                "\"opam config var share\" output empty string."))
-              ((not (file-directory-p share))
-               (spacemacs-buffer/warning
-                "opam share directory does not exist."))
-              (t (setq opam-share share
-                       opam-load-path (concat share "/emacs/site-lisp"))
-                 (add-to-list 'load-path opam-load-path))))
+      (cl-flet ((share-dir-from-command (command)
```
lebensterben, line 26: can you combine `cl-flet` and `let*` forms and replace them by `cl-letf*`?
that kill remove one level of nesting.

layers/+lang/ocaml/funcs.el
```diff
@@ -23,26 +23,30 @@
...
+                                        (string-trim-right (with-output-to-string (with-current-buffer standard-output
```
lebensterben, line 27: here it indents too much and hard to read.

lebensterben (changes requested):

layers/+lang/ocaml/funcs.el
akkuankka: ended up expanding the `let*` to include `share-dir-from-command` as a lambda variable and then called it using `funcall`, which seems clearer than `letf` (or my original solution for that matter). if this is unidiomatic or what-have-you i can redo.

layers/+lang/ocaml/funcs.el
```diff
@@ -23,26 +23,30 @@
...
-               (spacemacs-buffer/warning
-                "\"opam config var share\" output empty string."))
-              ((not (file-directory-p share))
-               (spacemacs-buffer/warning
-                "opam share directory does not exist."))
-              (t (setq opam-share share
-                       opam-load-path (concat share "/emacs/site-lisp"))
-                 (add-to-list 'load-path opam-load-path))))
+        (let* ((share-dir-from-command (lambda (command)
+                (string-trim-right (with-output-to-string (with-current-buffer standard-output
```
lebensterben, line 27: the indentation is still not right.
try to press `= =` on the line `(process-file shell-file-name`.

lebensterben: for indentation, either write
```elisp
(foo bar
     baz
     xyz)
```
or
```elisp
(foo
 bar
 baz
 xyz)
```

but never
```elisp
(foo bar
 baz
 xyz)
```

akkuankka:
That should have fixed the formatting errors &c, `= =` was causing some weird indentation but I think that's a me problem. For future reference is there a style guide for spacemacs or emacs lisp in general that I've missed out on somewhere that goes beyond what lebensterben mentioned above?

lebensterben:
https://www.gnu.org/software/emacs/manual/html_node/elisp/Tips.html

lebensterben:
@akkuankka 
I think your last forced push has reverted some changes I requested.

akkuankka:
Yikes,, ok at long last that should be doing what i meant *and also* what i said.

lebensterben (approved):

PR #15012, opened 2021-08-23
Add hackernews layer

BenediktBroich:
This layer adds support for reading Hacker News.

### DIFF
--- a/layers/+web-services/hackernews/packages.el
+++ b/layers/+web-services/hackernews/packages.el
@@ -1,0 +1,33 @@
+;;; packages.el --- hackernews layer packages file for Spacemacs.
+;;
+;; Copyright (c) 2012-2021 Sylvain Benner & Contributors
+;;
+;; Author: Benedikt Broich <b.broich@posteo.de>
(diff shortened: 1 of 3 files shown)


layers/+web-services/hackernews/packages.el
```diff
@@ -0,0 +1,33 @@
+;;; packages.el --- hackernews layer packages file for Spacemacs.
+;;
+;; Copyright (c) 2012-2021 Sylvain Benner & Contributors
```
lebensterben, line 3:
```suggestion
;; Copyright (c) 2021 Sylvain Benner & Contributors
```

lebensterben (changes requested):
There's no significant issue with this PR. But there's also no significant feature either.

I'd prefer to see a new layer to include more features/configurations rather than just being a collection of packages.

smile13241324:
I have merged this layer for now, it looks like the other open ones are pretty similar so I will merge them and later combine them to one `news` layer providing the various web service destinations.

PR #15013, opened 2021-08-23
Add lobsters layer

BenediktBroich:
This layer adds support for reading lobsters.

### DIFF
--- a/layers/+web-services/lobsters/packages.el
+++ b/layers/+web-services/lobsters/packages.el
@@ -1,0 +1,41 @@
+;;; packages.el --- lobsters layer packages file for Spacemacs.
+;;
+;; Copyright (c) 2012-2021 Sylvain Benner & Contributors
+;;
+;; Author: Benedikt Broich <b.broich@posteo.de>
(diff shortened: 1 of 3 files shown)


layers/+web-services/lobsters/packages.el
```diff
@@ -0,0 +1,41 @@
+;;; packages.el --- lobsters layer packages file for Spacemacs.
+;;
+;; Copyright (c) 2012-2021 Sylvain Benner & Contributors
```
lebensterben, line 3:
```suggestion
;; Copyright (c) 2021 Sylvain Benner & Contributors
```

lebensterben (changes requested):
There's no significant issue with this PR. But there's also no significant feature either.

I'd prefer to see a new layer to include more features/configurations rather than just being a collection of packages.

smile13241324:
I do agree this layer is really small, still I would like to have it merged for now. If we have gathered 3-4 similar web service layers we can start to group and merge them as needed.

PR #15017, opened 2021-08-25
[shell-scripts] Add key binding for shfmt-buffer

bcc32:
#14966 added format-on-save for shell scripts but I forgot to add a key binding for manually triggering reformatting.


lebensterben:

PR #15018, opened 2021-08-25
Fix typo that prevents ivy-rich from being enabled

rgemulla:
Fixes a misplaced paren that prevented ivy-rich from being enabled when ```ivy-enable-advanced-buffer-information``` was set but ```ivy-enable-icons``` wasn't.


lebensterben:
Just use two setq for clarity

Issue #15022, opened 2021-08-29
dotspacemacs-install-packages doesn't match condition-case in dotspacemacs/sync-configuration-layers


15cm:
## Description
Got errors  when invoking `dotspacemacs/sync-configuration-layers`:
```
condition-case: Variable: "dotspacemacs-install-packages" has value: "nil" that doesn’t match its type: "(choice (const used-only) (const used-but-keep-unused) (const all))". Validator message: "(user-error Looking for `(choice (const used-only) (const used-but-keep-unused) (const all))' in `nil' failed because:
  all of the options failed
    Looking for `(const used-only)' in `nil' failed because:
      not the expected value
    Looking for `(const used-but-keep-unused)' in `nil' failed because:
      not the expected value
    Looking for `(const all)' in `nil' failed because:
      not the expected value)"
```

## System Info
System Info :computer:
- OS: gnu/linux
- Emacs: 27.2
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 59852a6ab)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: helm
- Layers:
```elisp
((spacemacs-layouts :variables layouts-autosave-delay 300)
 (shell :variables shell-default-shell ’multi-term shell-default-height 30 shell-default-position ’bottom)
 spell-checking chinese neotree
 (ranger :variables ranger-show-preview t ranger-show-literal nil ranger-ignored-extensions
         ’("mkv" "iso" "mp4" "zip" "tar")
         ranger-max-preview-size 10 ranger-parent-depth 2)
 systemd restclient
 (auto-completion :variables spacemacs-default-company-backends
                  ’((company-files company-capf)
                    (company-abbrev company-dabbrev))
                  auto-completion-enable-snippets-in-popup t auto-completion-enable-sort-by-usage t auto-completion-enable-help-tooltip t)
 semantic emacs-lisp lsp dap scheme
 (c-c++ :variables c-c++-default-mode-for-headers ’c++-mode c-c++-backend ’lsp-ccls)
 rust
 (go :variables go-backend ’lsp go-use-golangci-lint t)
 (python :variables python-backend ’lsp python-lsp-server ’pyright python-sort-imports-on-save nil)
 (java :variables java-backend nil)
 (ruby :variables ruby-enable-enh-ruby-mode t ruby-backend ’lsp ruby-version-manager ’rbenv)
[...]

PR #15023, opened 2021-08-29
Bind dotspacemacs-* variables to current values when protecting them in test

15cm:
Fix https://github.com/syl20bnr/spacemacs/issues/15022.

### DIFF
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -1085,11 +1085,11 @@ error recovery."
                    "[[file:%s::dotspacemacs/layers][Show in File]]\n")
            dotspacemacs-filepath))
   ;; protect global values of these variables
-  (let (dotspacemacs-additional-packages
-        dotspacemacs-configuration-layer-path
-        dotspacemacs-configuration-layers
-        dotspacemacs-excluded-packages
-        dotspacemacs-install-packages
+  (let ((dotspacemacs-additional-packages dotspacemacs-additional-packages)
+        (dotspacemacs-configuration-layer-path dotspacemacs-configuration-layer-path)
+        (dotspacemacs-configuration-layers dotspacemacs-configuration-layers)
+        (dotspacemacs-excluded-packages dotspacemacs-excluded-packages)
+        (dotspacemacs-install-packages dotspacemacs-install-packages)
         (passed-tests 0)
         (total-tests 0))
     (load dotspacemacs-filepath)


lebensterben:
the binding SPEC of let form is a list of:

(SYMBOL EXPR)

where a new variable named SYMBOL is bound to the return of EXPR.

If EXPR is omitted, SYMBOL gets value of nil.

Therefore this PR should be dismissed.

15cm:
The problem is: nil is not a valid value for `dotspacemacs-install-packages`. Binding it to nil causes this error as I explained in the issue https://github.com/syl20bnr/spacemacs/issues/15022.
```
condition-case: Variable: "dotspacemacs-install-packages" has value: "nil" that doesn’t match its type: "(choice (const used-only) (const used-but-keep-unused) (const all))". Validator message: "(user-error Looking for `(choice (const used-only) (const used-but-keep-unused) (const all))' in `nil' failed because:
  all of the options failed
    Looking for `(const used-only)' in `nil' failed because:
      not the expected value
    Looking for `(const used-but-keep-unused)' in `nil' failed because:
      not the expected value
    Looking for `(const all)' in `nil' failed because:
      not the expected value)"
```

The purpose of this PR is to still follow the comment to protect altering global variables but initialize the local variables with their current value. So we don't assume nil to be the default value of the global variables.

15cm:
To limit the scope of the PR, I change the commit to only fix the initialization of `dotspacemacs-install-packages`. PTAL at the updated commit https://github.com/15cm/spacemacs/commit/036784cc259bd339da246209fbb61c1311d85b5b. The PR doesn't reflect it probably because the PR has been closed.

lebensterben:
I will investigate.

15cm:
Not a bug but IMO a better to have per discussion in https://github.com/syl20bnr/spacemacs/issues/15022#issuecomment-909771382.

lebensterben:
The following variables are protected by shadowing in `dotspacemacs//test-dotspacemacs/layers`:
- `dotspacemacs-additional-packages`
- `dotspacemacs-configuration-layer-path`
- `dotspacemacs-configuration-layers`
- `dotspacemacs-excluded-packages`
- `dotspacemacs-install-packages`

---

# `dotspacemacs-additional-packages`

`dotspacemacs/sync-configuration-layers` ->
[`configuration-layer/load`](https://github.com/syl20bnr/spacemacs/blob/ed12b1118f1e6b3813dfc2a98536f2316524d0d4/core/core-dotspacemacs.el#L850) ->
[`configuration-layer/load`](https://github.com/syl20bnr/spacemacs/blob/ed12b1118f1e6b3813dfc2a98536f2316524d0d4/core/core-configuration-layer.el#L563) ->
[`configuration-layer//load`](https://github.com/syl20bnr/spacemacs/blob/ed12b1118f1e6b3813dfc2a98536f2316524d0d4/core/core-configuration-layer.el#L584) (appears in the first three branch of `cond`, but should go to the second one because of https://github.com/syl20bnr/spacemacs/blob/ed12b1118f1e6b3813dfc2a98536f2316524d0d4/core/core-dotspacemacs.el#L849) ->
[`configuration-layer//declare-used-packages`](https://github.com/syl20bnr/spacemacs/blob/ed12b1118f1e6b3813dfc2a98536f2316524d0d4/core/core-configuration-layer.el#L633) ->
[`configuration-layer/make-packages-from-dotfile`](https://github.com/syl20bnr/spacemacs/blob/ed12b1118f1e6b3813dfc2a98536f2316524d0d4/core/core-configuration-layer.el#L1675) which appends additional theme packages to `dotspacemacs-additional-packages`(https://github.com/syl20bnr/spacemacs/blob/ed12b1118f1e6b3813dfc2a98536f2316524d0d4/core/core-configuration-layer.el#L1237).

It's safe to set it to nil in test environment.

# `dotspacemacs-configuration-layer-path`

This is used more than once.

`dotspacemacs/sync-configuration-layers` ->
[`dotspacemacs/test-dotfile`](https://github.com/syl20bnr/spacemacs/blob/ed12b1118f1e6b3813dfc2a98536f2316524d0d4/core/core-dotspacemacs.el#L836) ->
[`configuration-layer/discover-layers`](https://github.com/syl20bnr/spacemacs/blob/ed12b1118f1e6b3813dfc2a98536f2316524d0d4/core/core-dotspacemacs.el#L1215)

`dotspacemacs/sync-configuration-layers` ->
[`configuration-layer/load`](https://github.com/syl20bnr/spacemacs/blob/ed12b1118f1e6b3813dfc2a98536f2316524d0d4/core/core-dotspacemacs.el#L850) ->
[`configuration-layer//load`](https://github.com/syl20bnr/spacemacs/blob/ed12b1118f1e6b3813dfc2a98536f2316524d0d4/core/core-configuration-layer.el#L593) ->
[`configuration-layer/discover-layers`](https://github.com/syl20bnr/spacemacs/blob/ed12b1118f1e6b3813dfc2a98536f2316524d0d4/core/core-configuration-layer.el#L631)

both funcall of `configuration-layer/discover-layers` uses
[`dotspacemacs-configuration-layer-path`](https://github.com/syl20bnr/spacemacs/blob/ed12b1118f1e6b3813dfc2a98536f2316524d0d4/core/core-configuration-layer.el#L1427)

It's safe to set it to nil in test environment.

PR #15024, opened 2021-08-30
Add outline minor mode bindings to navigation layer

dalanicolai:
Outline-minor-mode works better than `hide-show-minor-mode` in some (or maybe
all) cases. E.g. compare in `transient.el` `hs-hide-all` vs
`outline-cycle-buffer`.

Outline mode is useful (and should be used by default) for structuring large
single file elisp packages (which because of its simplicity is often preferable over
splitting the package into multiple files), but it can be used in any other mode
too. For an example, see `transient.el`.

It seems that the bindings don't conflict with any other mode (I have ripgrep
searched through the `layers` dir). So I guess `outline-minor-mode` might as well
be activated by default.

### DIFF
--- a/layers/+spacemacs/spacemacs-navigation/README.org
+++ b/layers/+spacemacs/spacemacs-navigation/README.org
@@ -25,6 +25,15 @@ This layer adds general navigation functions to all supported layers.
 
 * Key bindings
 
-| Key binding | Description    |
-|-------------+----------------|
-| ~SPC h j~   | jump to manual |
+| Key binding | Description               |
+|-------------+---------------------------|
+| ~SPC h j~   | jump to manual            |
+| ~SPC t o~   | toggle outline-minor-mode |
+
+*Outline minor mode*
+| ~SPC m o o~   | outline cycle buffer             |
+| ~SPC m o TAB~ | outline cycle (heading)          |
+| ~SPC m o j~   | outline next visible heading     |
+| ~SPC m o k~   | outline previous visible heading |
+| ~SPC m o J~   | outline next heading             |
+| ~SPC m o K~   | outline previous heading         |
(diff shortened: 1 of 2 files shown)


lebensterben:
I personally use outline minor mode plus outshine mode.

but it's not a replacement to hide-show. hide-show uses syntax table to show/hide code. While outline mode only recognize "headings".

Surely these two normally won't conflict with each other. And I doubt anyone use hide-show instead of evil-fold/origami etc to fold codes.

lebensterben:
one thing I'm probably over-cautious here is that this introduce new bindings which could collide with others, given that outline mode is almost certainly enabled in all buffers.

in the meantime I don't think we should use long prefixes to avoid conflicts, say "SPC a o", because navigation across headings/sections could be frequently used by some people.

once again I draw inspiration from evil-collection.
In particular, the motion bindings here seems to make sense: https://github.com/emacs-evil/evil-collection/blob/940dfe91aa1272979669785fe803d6ce8846eaf6/modes/outline/evil-collection-outline.el#L94

dalanicolai:
I have looked again into it now, and I have found that evil comes with all kinds of folding bindings by default. The actions can be configured via the `evil-fold-list`. Subsequently, the action is selected depending on what mode is active. I simply searched for bindings for `outline-cycle(-buffer)` and did not find them. But the default evil bindings are already better for sure.

We could optionally implement a binding to cycle the folding-style(/package). But I don't think it is necessary (users can just configure it in the dotfile, where users can also set different folding methods: evil, origami or vimish). Anyway, I will close this issue.

Those evil-collection motion bindings are conflicting with bindings from many packages (as you can see by grepping for "]]" in your elpa dir, but they are only activated when outline-mode is activated of course. Anyway, not so wise to use them with outline-minor-mode).

PR #15028, opened 2021-08-31
latex layer add auto configure pdf-tools and open in split window

dalanicolai:
Due to some question on Gitter, and some struggle to find the correct answer,
 I propose to add the following sane configuration options to the latex layer
(It is anyway nice to be able to allow for configuring these things with layer variables).

I did not write the documentation because I first "poll" for your opinion. So let me
 know if you think that I should finish this PR.
 
This commit makes the latex layer automatically configure pdf-tools as
latex pdf-viewer if the pdf layer is installed (or use another viewer by setting the
`latex-view-with-pdf-tools` variable to nil). Here I assume that if the pdf layer is
installed, then users would also mostly prefer to use it (otherwise I would anyway
like users to try it).

Additionally, it adds a layer variable for configuring Spacemacs to open pdf-view in a
split-window. (Strangely, finding out how to achieve that is far from trivial,
and also not easy to google, because many hits do not provide the "correct"
answer (see the typical example
[here](https://emacs.stackexchange.com/questions/28056/spacemacs-latex-how-to-make-view-command-tile-vertically-with-emacs-window)).

### DIFF
--- a/layers/+lang/latex/funcs.el
+++ b/layers/+lang/latex/funcs.el
@@ -56,7 +56,7 @@
     (lsp-deferred)))
 
 (defun spacemacs//latex-setup-pdf-tools ()
-  "Conditionally setup pdf-tools"
+  "Conditionally setup pdf-tools."
   (when latex-view-with-pdf-tools
     (setf (alist-get 'output-pdf TeX-view-program-selection) (list "PDF Tools"))
 
--- a/layers/+lang/latex/funcs.el
+++ b/layers/+lang/latex/funcs.el
@@ -55,6 +55,15 @@
     (require 'lsp-latex)
     (lsp-deferred)))
 
+(defun spacemacs//latex-setup-pdf-tools ()
+  "Conditionally setup pdf-tools"
+  (when latex-view-with-pdf-tools
+    (setf (alist-get 'output-pdf TeX-view-program-selection) (list "PDF Tools"))
+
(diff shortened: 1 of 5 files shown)


layers/+lang/latex/funcs.el
```diff
@@ -55,6 +55,15 @@
     (require 'lsp-latex)
     (lsp-deferred)))
 
+(defun spacemacs//latex-setup-pdf-tools ()
+  "Conditionally setup pdf-tools"
+  (when latex-view-with-pdf-tools
```
lebensterben, line 60: what if a user set it to `t` while `pdf-tools` is not used?

```elisp
(when latex-view-with-pdf-tools
  (if (configuration-layer/layer-used-p 'pdf)
      ;; THEN
    ;; show a warning
```

lebensterben (changes requested):
LGTM.
Just one minor issue.

Also needs to update CHANGELOG.develop and README.org

layers/+lang/latex/funcs.el
```diff
@@ -55,6 +55,15 @@
     (require 'lsp-latex)
     (lsp-deferred)))
 
+(defun spacemacs//latex-setup-pdf-tools ()
+  "Conditionally setup pdf-tools"
```
lebensterben, line 59:
```suggestion
  "Conditionally setup pdf-tools."
```

smile13241324:

PR #15030, opened 2021-09-02
Revert "Fix evil search prompt on mouse leave"

lebensterben:
Reverts syl20bnr/spacemacs#15002

reverting since it's fixed upstream https://github.com/emacs-evil/evil/pull/1330


duianto:

PR #15037, opened 2021-09-05
c-c++: Package cpp-auto-include can be fetched from Melpa

eklitzke:
The package `cpp-auto-include` is now [in Melpa](https://melpa.org/#/cpp-auto-include) so it doesn't need to be fetched from GitHub (and the github repo it points to redirects now anyway).


lebensterben:
Thanks. at first glance `syohex/emacs-cpp-auto-include` is different from the upstream repository `emacsorphanage/cpp-auto-include`, but it redirects there.

PR #15038, opened 2021-09-08
README: fix macOS prerequisites regarding grep

thvnx:
The --with-default-names option seems to no longer exist. A directory have to be added to PATH instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,10 +253,18 @@ packages may fail. In this case it is possible to install using
 be transferred using http, use at your own risk.
 
 You might also have some issues when doing some search on your projects, you
-probably want to install grep through homebrew with default names:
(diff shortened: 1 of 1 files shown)


lebensterben (approved):

lebensterben:
Great. We should recommend users to install grep's faster replacement as well.

thvnx:
> Great. We should recommend users to install grep's faster replacement as well.

Thanks. I think I don't know it, what are you talking about?

lebensterben:
> > Great. We should recommend users to install grep's faster replacement as well.
> 
> 
> 
> Thanks. I think I don't know it, what are you talking about?

rg ag ack

thvnx:
OK thanks ;)

Issue #15039, opened 2021-09-09
Feature request: add inspector to emacs-lisp layer


vigilancetech-com:
It would be really nice to add a data inspector to the emacs lisp layer, for example:

https://github.com/mmontone/emacs-inspector

I've tried adding it but I can't get it to recognize the inspector's key strokes as they seem to be completely overridden by evil mode keys.

dalanicolai:
If you would still like to add this, you would probably want to use `evilified-state-evilify-map`. You can check its docstring for how to use it, and search using `SPC s f` in your `.emacs.d` directory to find plenty of examples (also see [here](https://develop.spacemacs.org/doc/CONVENTIONS.html#evilified-buffers) for info about evilified state).

Additionally, as I am not familiar with inspectors so that I don't know how useful this is generally, I would suggest that you make it an optional package using the `:toggle` keyword. Again, you could find how to use it by searching for it in `.emacs.d`  using `SPC s f` (e.g. look at the `org` layer its `packages.el` and `config.el` files).

Finally, you should add some info about it in the documentation. As it might be that many users are not familiar with inpectors, it would be great if you could add some more info than given on that github page you've linked to.

Of course you can always ask for help here or on gitter (which you know already :).
If you prefer someone else to implement this, then please add some more info about its usefulness. But of course would be great if you manage to make it work.

Anyway, thanks for creating the issue!

lebensterben:
@dalanicolai 

the most common use case for normal user would be --debug-init.
Emacs by default prints tons of backtrace with almost no annotations and useful information.

this inspector provides a command to interactively "step in/out". so that at least you are viewing fewer information at a time.

but still Emacs lisp debugging is painful...

vigilancetech-com:
this video shows how handy an inspector can be.  In this one the inspector directly accessible in the source buffer, which would be ideal, but even having it have to be in a separate buffer is far better than nothing.   I'm working on some stuff that can have some pretty convoluted data structures so its really a necessity.

https://youtu.be/rh5J4vacG98?t=1084

All I really need it to do is give the inspector's keys presidence over all the others in the inspector buffer.    I'm really kind of surprised it doesn't just do that automatically once it's loaded, to wit, that evil would just hijack the mode (or maybe the inspector author didn't define his keys in the right map).

dalanicolai:
[...]

PR #15041, opened 2021-09-10
Implement inspector package to emacs-lisp layer see #15039

dalanicolai:
This is a PR in response to #15039.

I have just implemented it isnthe most straightforward way. We might want to remap the `inspector-pop` from `L` to a more mnemonic/consistent (with other inspectors) key.


smile13241324 (approved):
Looks good

smile13241324:

PR #15045, opened 2021-09-14
Fix autocompletion in elisp and text mode

dalanicolai:
Currently, in `emacs-lisp-mode`, most or all auto-completion, using completion-at-point (i.e. the TAB
key), is 'hijacked' by company-capf in emacs-lisp-mode, while in text mode there
is no auto-completion by default at all.

EXAMPLE: In emacs-lisp-mode try typing:
```
(defun test-function ()
  (print "hello"))

(test-fun)
```
and try to auto complete using TAB, no suggestions are given. You can use `M-x company-diag` to find out which backend is used.

or in `text-mode` try to type 
```
foobar
foob
```
and try to complete using TAB (it will insert a TAB instead of suggesting completions)

This commit adds auto-completion for text-mode and slightly improves the
situation for emacs-lisp mode (by grouping `company-capf` with `company-dabbrev(-code)`).

When my question at https://github.com/company-mode/company-mode/discussions/1186#discussioncomment-1322465 gets answered, then I will update the configuration here accordingly.

### DIFF
--- a/layers/+completion/auto-completion/packages.el
+++ b/layers/+completion/auto-completion/packages.el
@@ -114,6 +114,7 @@
     (progn
       (spacemacs|diminish company-mode " ⓐ" " a")
 
+      (spacemacs|add-company-backends :modes text-mode)
       ;; key bindings
       (defun spacemacs//company-complete-common-or-cycle-backward ()
         "Complete common prefix or cycle backward."
--- a/layers/+lang/emacs-lisp/packages.el
+++ b/layers/+lang/emacs-lisp/packages.el
@@ -80,7 +80,7 @@
           (lisp-indent-line))))))
 
 (defun emacs-lisp/post-init-company ()
-  (spacemacs|add-company-backends :backends company-capf
+  (spacemacs|add-company-backends :backends (company-capf company-dabbrev-code)
                                   :modes emacs-lisp-mode)
   (spacemacs|add-company-backends :backends (company-files company-capf)
                                   :modes ielm-mode))


layers/+completion/auto-completion/packages.el
```diff
@@ -109,11 +109,15 @@
...
       (add-hook 'company-completion-finished-hook 'company-maybe-turn-on-fci)
-      (add-hook 'company-completion-cancelled-hook 'company-maybe-turn-on-fci))
+      (add-hook 'company-completion-cancelled-hook 'company-maybe-turn-on-fci)
+
+      (add-hook 'text-mode-hook 'company-mode nil t))
     :config
     (progn
       (spacemacs|diminish company-mode " ⓐ" " a")
 
+      (spacemacs|add-company-backends :backends company-dabbrev
```
lebensterben, line 119: This [list of company backends](https://github.com/syl20bnr/spacemacs/blob/cabe650f008b07f240d3a220960e2c84c5ea9918/layers/%2Bcompletion/auto-completion/config.el#L26-L32) is always enabled by default when calling [`spacemacs|add-company-backends`](https://github.com/syl20bnr/spacemacs/blob/cabe650f008b07f240d3a220960e2c84c5ea9918/layers/%2Bcompletion/auto-completion/funcs.el#L101-L103).

So you can just write
```elisp
(spacemacs|add-company-backends :modes text-mode)
```

layers/+lang/emacs-lisp/packages.el
```diff
@@ -80,7 +80,7 @@
           (lisp-indent-line))))))
 
 (defun emacs-lisp/post-init-company ()
-  (spacemacs|add-company-backends :backends company-capf
+  (spacemacs|add-company-backends :backends (company-capf company-dabbrev-code)
```
lebensterben, line 83: from `company-diag`, you will find `company-dabbrev-code` already enabled.

lebensterben (changes requested):

layers/+lang/emacs-lisp/packages.el
dalanicolai: Really? Did you try my `test-function` example above?
I know that `company-dabbrev-code` is activated by default, but, at least in `emacs-lisp-mode` all completion is `hijacked` by `company-capf`. It is explained in a little more detail in the thread I have linked to (but also the comment itself) above.

layers/+completion/auto-completion/packages.el
dalanicolai: Updated

layers/+lang/emacs-lisp/packages.el
lebensterben: you're right.
currently `company-capf` and `company-dabbrev` are two separate entities in the `company-backends` and `compnay-dabbrev` is shadowed whenever `company-capf` or any other backend before it returns non-nil value.

grouping `company-capf` and `company-dabbrev` in one list should fix the problem.

lebensterben (approved):

lebensterben:
LGTM

layers/+completion/auto-completion/packages.el
```diff
@@ -109,11 +109,14 @@
...
       (add-hook 'company-completion-finished-hook 'company-maybe-turn-on-fci)
-      (add-hook 'company-completion-cancelled-hook 'company-maybe-turn-on-fci))
+      (add-hook 'company-completion-cancelled-hook 'company-maybe-turn-on-fci)
+
+      (add-hook 'text-mode-hook 'company-mode nil t))
     :config
     (progn
       (spacemacs|diminish company-mode " ⓐ" " a")
 
+      (spacemacs|add-company-backends :modes text-mode)
```
smile13241324, line 119: I am not sure that this is a good idea. The general design of the auto completion layer is to not decide whether a layer registers company backends for its modes. When a company-backend is registered the macro `spacemacs|add-company-backends` will activate company mode for the registered modes automatically instead. Layers have to call `post-init-` code to register handlers on a layer basis. 

With this change we break this design by having the auto completion layer activate company for all `text-modes` in general. I think we should rather change the existing layers to properly register the backends for their respective modes this includes the respective `text-modes`.

layers/+lang/emacs-lisp/packages.el
smile13241324, line 83: I am not sure that we really need this, I have copied your sample code into a sample buffer and tried to auto-complete test-fun, which did work very well after me having evaluated the new function. This behaviour did not change for me when I evaluate the new code.

Can you double check?

smile13241324 (changes requested):

layers/+completion/auto-completion/packages.el
dalanicolai: I went looking at this, but somehow I got distracted by other things. Anyway, thanks for the explanation about the general design, and the 'add-company-backend' macro. So, in any case, I will remove the hook then.

The thing I wanted to fix is auto-completion in text-mode buffers,
i.e. when I write some text in a .txt buffer, I would like to have autocompletion (and I think Spacemacs should offfer/activate it by default for this case).
As I could not find any specialized `text-layer`, I went looking for where best to add this. Maybe I should create a special layer for it, but I think that would be overkill. Also, I am not sure why we would not want basic auto-completion in all text layers. But if we for some reason don't want that, then it would still be nice to configure it for `.txt` buffers.
I find it just weird to have no auto-completion at all in such buffers.

layers/+lang/emacs-lisp/packages.el
dalanicolai: Yes, I know that, after evaluating, the auto-completion works. But evaluating should not be necessary. When I type a single function and use the same new symbol in it a few times, it should get auto-completed, and currently it does not.
This is exactly what the `company-dabbrev` backend is for, and it is already configured for `emacs-lisp-mode` (actually it is in the Spacemacs default company backends). However, the configuration is wrong for `emacs-lisp-mode` because in this mode all completion is hijacked by `company-capf`. In other modes, where `company-capf` uses different auto-completion functions (e.g. pyhton mode), `company-capf` does not hijack all auto-completion, so for such modes `company-capf` is configured correctly. I actually, "discussed" it in a company issue, and this solution was what they recommended (and indeed it works fine and does what it should do here).

dalanicolai: Okay, I wanted to create an example, and it turns out that `comapny-capf` does complete `variables within scope`.
So I have quickly made up a little more exotic, but still very relevant, example:
```
;; This is a comment about this-example-file
(defun test-function (some-list)
  (if (eq some-list nil)
      nil
    (test-f

;; this-example-f
```
So here, even though I am in the scope of test-function, it does not yet get auto-completed. Of course here is no way that I can already evaluate the function before recursively using it.
Additionally, I have added a comment mentioning already e.g. the file-name, then when I try to make a second comment the file-name does not get completed. For sure there are more examples, I started this fix because it was annoying me, and then when I tested e.g. in vim (but probably also most other editors), it just worked as expected.

layers/+completion/auto-completion/packages.el
lebensterben: it seems that org-mode and tex-mode both derive from text-mode. so changing behavior of the text-mode will leak into other modes unexpectedly. (because Tex is a programming language 🤦🏻‍♂️ not natural language)

lebensterben:
after reading your comments I think we can make a new layer dedicated for human language editing. 
- It introduces a major mode that derives from text-mode.
- Disable syntax checking.
- Enable spell checking by default.
- Enable company. Probably with https://melpa.org/#/company-wordfreq and not company-dabbrev.
- Disable UI such as rainbow parentheses, indent guide, etc.

layers/+completion/auto-completion/packages.el
dalanicolai: I wanted to test if is this messes up something (because those backends get added to the layer anyway), then I found two thing about auto-complete:

[Here](https://develop.spacemacs.org/layers/+completion/auto-completion/README.html#enable-company-globally) the documentation says that enabling company globally is not recommended because disabling company in specific layers will not work anymore.

While [here]((https://develop.spacemacs.org/doc/DOCUMENTATION.html#disabling-layer-services-in-other-layers) the documentation mentions that company can be disabled for specific modes by loading the layer as follows:
`(auto-completion :disabled-for org)`.

But testing this on an `org` file, it seems that the disabling mechanism is already broken.

Finally, as far as I know about all this, in my opinion it would make sense to make auto-completion a default in text-mode (also when no 'human language editing' layer is installed), but some disabling mechanism should work. Also, there is still `fundamental-mode` if users really don't want auto-complete (by default).

smile13241324:
I will have a look at this at the weekend, this is waiting for too long already.

dalanicolai:
No worries! Just, I do not completely remember the exact details now, but I think I wrote the essential points in the comments. As far as I remember, company did not work in text mode at all, and it did not use the abbrev backend in `emacs-lisp-mode`. So it just adds company completion for text mode, and it fixes the use of [the dabbrev backend](https://jblevins.org/log/dabbrev-expand) in `emacs-lisp-mode`.

dalanicolai:
> after reading your comments I think we can make a new layer dedicated for human language editing.
> 
>     * It introduces a major mode that derives from text-mode.
> 
>     * Disable syntax checking.
> 
>     * Enable spell checking by default.
> 
>     * Enable company. Probably with https://melpa.org/#/company-wordfreq and not company-dabbrev.
> 
>     * Disable UI such as rainbow parentheses, indent guide, etc.

As far as I know activating company for text-mode does not have any effect on other modes (maybe just activating company). And the dabbrev backend really only suggests completions of words already used in the buffer (it 'automates' [dabbrev-expand](https://jblevins.org/log/dabbrev-expand) using company). If users like further improvemnets (like using the company-wordfreq backend) then they are very welcome to create a PR (it all takes quite some time to investigate and sort out). This PR simply improves the current situation.

smile13241324 (approved):
Works as expected

smile13241324:

PR #15046, opened 2021-09-14
Fix (activate) company-emoji completion

dalanicolai:
Currently company-emoji does not get activated by the emoji layer.
Also when it is activated, then the emoji short-name does not get completed to
unicode if `company-emoji-insert-unicode` is `t`.

This PR fixes both issues, by activating company-emoji for text-mode and
advising the emoji insert function to optionally do company-emoji completion
(try e.g. in magit-commit buffers).


lebensterben (approved):
LGTM

smile13241324 (approved):
Looks ok

PR #15050, opened 2021-09-16
Add documentation about evilifying 'special-mode buffers' (+fix eww)

dalanicolai:
This PR implements an easy fix for a case (eww-mode) where the change in #14995 caused an issue.

Additionally this PR adds documentation that should have been added already in #14995.

### DIFF
--- a/doc/CONVENTIONS.org
+++ b/doc/CONVENTIONS.org
@@ -90,6 +90,10 @@ It is recommended to make sure that ~q~ allows to leave the transient-state.
 /Evilifying/ a buffer is to set the =evilified state= as the default
 state for the major mode of the buffer.
 
+Except for =eww-mode= buffers, all buffers with major-modes derived from
+=special-mode= are evilified by default. To exclude buffers of such major-modes
+from getting evilified, add a regexp to =evil-buffer-regexps=.
+
 The =evilified state= is derived from the =emacs state= and modify the
 map to:
 - add ~hjkl~ navigation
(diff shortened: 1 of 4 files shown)


doc/CONVENTIONS.org
```diff
@@ -90,6 +90,10 @@ It is recommended to make sure that ~q~ allows to leave the transient-state.
 /Evilifying/ a buffer is to set the =evilified state= as the default
 state for the major mode of the buffer.
 
+Except for =eww-mode= buffer, all buffer with major-modes derived from
+=special-mode= are evilified by default. To exclude buffers of such major-modes
```
lebensterben, line 94: The instruction here is not clear.
`evil-buffer-regexps` is a alist where each element is (REGEXP . STATE).
STATE is one of evil state and when it's nil evil is **disabled**.

But currently eww buffers has ("\\*eww\\*" . normal), which means it defaults to evil NORMAL state instead of being disabled.

dalanicolai: > The instruction here is not clear.
> `evil-buffer-regexps` is a alist where each element is (REGEXP . STATE).
> STATE is one of evil state and when it's nil evil is **disabled**.
> 

You mean I should say the buffer opens in `evilified-state` instead of saying they get `evilified`? `evilified-state` is an extra `evil state` introduced in Spacemacs. The default is to open buffers in `normal-state` (so they are not `evilified`. As the `conventions.org` is more targeted to developers, I'm hoping/assuming that this explanation clear enough.

> But currently eww buffers has ("\_eww\_" . normal), which means it defaults to evil NORMAL state instead of being disabled.

With 'such major-modes', I am referring to 'major-modes that are derived from special-mode'. To open them in `normal-state` (or any other state) instead of `evilified-state`, the developer should use the `evil-buffer-regexps`. I think this explanation is sufficiently clear for developers reading this/for whom this information is relevant.

smile13241324:

PR #15068, opened 2021-09-29
Add pydoc to python layer for better python doc functionality

dalanicolai:
Currently the python layer (without anaconda backend) provides only helm-pydoc
or pylookup-lookup-at-point functionality. Both these functions are cumbersome.
`pydoc-at-point-no-jedi` adds simpler doc functionality, therefore this commit
adds this command as a third option. It places it under `SPC m h p` because `SPC
m h h` is already taken by the anaconda doc functionality (it could be placed on
`h h` if anaconda is not available, but I did not bother to implement it)

EXAMPLE:
see [question at stack-overflow](https://emacs.stackexchange.com/questions/66803/how-can-i-access-the-python-documentation-for-the-thing-at-cursor/68707#68707)
Now try to answer it. You will probably agree that this provides the simplest/best solution.
Although both `helm-pydoc` and `pylookup-lookup-at-point` are functional, they require unnecessary (for most cases) extra steps which makes their usage 'frustrating' for most cases.


smile13241324 (approved):
Good catch, I will merge it right away

PR #15074, opened 2021-09-30
[compleseus] tuned completion-style and added support for M-:

thanhvg:
- set completion-styles '(basic partial-completion orderless) currently it is '(orderless) which is ridiculous in how it suggests company results. 
- enable consult for doing tab in `M-:`

### DIFF
--- a/layers/+completion/compleseus/packages.el
+++ b/layers/+completion/compleseus/packages.el
@@ -250,9 +250,7 @@
 (defun compleseus/init-orderless ()
   (use-package orderless)
   :init
-  (setq completion-styles '(orderless)
-        ;; for ssh completion
-        completion-category-overrides '((file (styles basic partial-completion)))
+  (setq completion-styles '(basic partial-completion orderless)
         completion-category-defaults nil
         completion-category-overrides '((file (styles . (partial-completion))))))
 
@@ -289,6 +287,10 @@
     ;; Enable recursive minibuffers
     (setq enable-recursive-minibuffers t)
 
+    ;; when vertico is used set this so tab when doing M-: will show suggestions
+    ;; https://github.com/minad/vertico/issues/24
+    (setq completion-in-region-function #'consult-completion-in-region)
+
     ;; Optionally enable cycling for `vertico-next' and `vertico-previous'.
     ;; (setq vertico-cycle t)
     (vertico-mode)
@@ -310,7 +312,6 @@
     (spacemacs/set-leader-keys
       "rl" 'vertico-repeat
       "sl" 'vertico-repeat)))
-
 
 (defun spacemacs/compleseus-wgrep-change-to-wgrep-mode ()
   (interactive)


lebensterben (approved):

dalanicolai:
I just noticed this also, and I agree that the pure `(orderless)` suggestions are/work ridiculous. I am trying to understand this ecosystem, but it is very frustrating because they recommend such weird configurations, and it is hard to find a sensible one.
The idea/philosophy of those packages sounds nice and worth it, but Ivy is definitely much more straightforward and more friendly to use.

I fully agree with these two improvements, but even with `#'consult-completion-in-region`, the completion is not yet sensible.

Also, the vertico README suggests to try `'(basic substring partial-completion flex)`, but I don't know if this would make more or less sense than what you configured here; just mentioning it.

Anyway, kudos for fighting this beast...

dalanicolai:
Actually, as long as the `#'consult-completion-in-region` does not use the `completion-styles` (e.g. provide partial-completion), I would prefer to remove it (and just use orderless with its completion-styles). But even better would be to fix the completion-styles for the `#'consult-completion-in-region`.

dalanicolai:
@minad Thank you very much for your elaborate explanations (also in https://github.com/oantolin/orderless/issues/85). Although I appreciate your work a lot, unfortunately, I did get frustrated because I was spending a lot of time on testing many configurations without success. Of course, your packages are new, and very configurable, and it is a whole ecosystem, so I totally understand it. But I hope some parts in the documentation will get (even more) clear over time, including these comments here making it to more prominent places of the documentation.  We are quite spoiled with almost perfectly fully configured 'ecosystems' like Ivy and Helm.

I did not try your suggestions yet, but I remember that I tried to get sensible autocompletion for `describe-function` and for `M-:`. So first, I did not get why for `M-:` the completion was not 'passed to vertico' (which @thanhvg cleared up in this commit). But then (if I tried it correctly), when the autocompletion did get handed to vertico, I remember the autocompletion configuration that was working for `describe-function` was not working when using it with `M-:` (I was trying to autocomplete `current-buffer` and `(current-buffer)` using the `cb` initialism(s), but I got weird results. Then I also found that orderless has a separate completion style for `orderless-initialism` and `orderless-strict-leading-initialism`, which suggested to me that I could not use `^` and `$` with initialisms (indeed they do not work by default). So I do not know if 'we' already know enough to also fix the `M-:` issue. But anyway I will look at it again soon when I find some time again (and probably @thanhvg will), so we will come back to it maybe.

Anyway again, thanks a lot for your 'support' :pray:

minad:
> But I hope some parts
> in the documentation will get (even more) clear over time, including
> these comments here making it to more prominent places of the
> documentation. 

I tried to update the documentations and Consult wiki recently. The
problem with these completion styles is that they affect *everything*.
And different commands or UIs have different requirements. Here you are
struggling with the different requirements of Company and minibuffer
completion for example.

> We are quite spoiled with almost perfectly fully
> configured 'ecosystems' like Ivy and Helm.

Yes of course. I believe Ivy doesn't pose the same problems since it
doesn't use completion styles but its own filtering machinery. So you
can just leave the Emacs completion style configuration as is. But from
my perspective, by not using the completion machinery you get
consistency issues in Emacs down the road. Orderless tries to avoid
this, it embraces the default machinery and just provides this more
advanced completion style. On the other hand Helm uses the completion
machinery (as the Orderless README points out there is a `helm`
completion style). However it could be that Helm restricts its usage of
its own completion style to Helm only, such that it doesn't affect
Company or completion-in-region.

> I did not try your suggestions yet, but I remember that I tried to get
> sensible autocompletion for |describe-function| and for |M-:|. So first,
> I did not get why for |M-:| the completion was not 'passed to vertico'
> (which @thanhvg <https://github.com/thanhvg> cleared up in this commit).

Yes, Vertico does not replace completion-at-point. This is not the
business of Vertico. In fact completing-read is subtly different from
completion-at-point/completing-read, so it makes a lot of sense that
Vertico stays out of this business.

By using consult-completion-in-region there is a way to opt-in to
completion-at-point by transferring the completion to the minibuffer
(completion-at-point -> completion-in-region ->
completion-in-region-function=consult-completion-in-region ->
completing-read). But note that this is an arbitrary choice. Personally
I am not using consult-completion-in-region, I am using Corfu which
comes with its own completion-in-region-function. Again, therefore it
makes sense that Vertico doesn't take over automatically.

But this question came up very often - many people asked why TAB
completion opens a weird `*Completions*` buffer despite Vertico being
activated. So I added pointers to the Vertico and Consult README.
Unfortunately this kind of documentation is often overlooked since users
are not aware of the distinction of
completing-read/completion-in-region/completion-at-point. Reason for
this is of course that the Ivy/Helm packages also conflate these things.

> But then (if I tried it correctly), when the autocompletion did get
> handed to vertico, I remember the autocompletion configuration that was
> working for |describe-function| was not working when using it with |M-:|

This should not be the case. As long as you didn't use consult-customize
to configure a separate completion-styles for
consult-completion-in-region, both completion-at-point and
describe-function will use the same styles.

> (I was trying to autocomplete |current-buffer| and |(current-buffer)|
> using the |cb| initialism(s), but I got weird results. Then I also found
> that orderless has a separate completion style for
> |orderless-initialism| and |orderless-strict-leading-initialism|, which
> suggested to me that I could not use |^| and |$| with initialisms
> (indeed they do not work by default).

The problem here is that Orderless does not even support initialisms by
default, so "cb" won't complete to current-buffer! Orderless has a
performance sensitive default configuration which only compiles the
separate words to regular expressions or literals.

(setq orderless-matching-styles '(orderless-literal orderless-regexp))

The input is compiled to a list of regular expressions, which the
candidates are matched against.

"word1 word2" -> ("word1" "word2")
"f.*oo bar"   -> ("f.*oo\\|f\\.\\*oo" "bar")

If you want to have initialisms then you have to enable the Orderless
*matching style* orderless-initialism. There are multiple variants
(strict initialism etc).

(setq orderless-matching-styles '(orderless-literal orderless-regexp
orderless-initialism))

"foo" -> ("foo\\|f.*-o.*-o")     (literal+initialism, approximately)

But note that this can lead to an exponential slowdown when matching
long candidate strings. The same applies to the orderless-flex matching
style which is even more prone to such a slowdown.

(setq orderless-matching-styles '(orderless-literal orderless-regexp
orderless-flex))

"foo" -> ("foo\\|f.*o.*o")

The orderless recommendation is therefore to opt-in to these special and
expensive matching styles via *style dispatchers*. Furthermore I am
defining a customized orderless-with-initialism style which includes
orderless-initialism by default. This style is only used for completions
with short completion candidates, where the initialism regexps won't
lead to slowdown. See the Consult wiki for +orderless-dispatch and
+orderless-with-initialism.

I hope this clears it up a bit? Anyways if you have suggestion on where
the documentation should be improved please point this out.

PR #15078, opened 2021-10-02
Add commands to compleseus layer (locate, yasnippet, search from)

dalanicolai:
This PR adds two (default) keybindings to compleseus layer, i.e. consult-locate on `SPC f L` and consult-yasnippet on `SPC i s`.

Furthermore, it adds an embark action to start a grep search from a candidate its path's directory.
@thanhvg Please review/approve if you agree. The PR is just a proposal.

### DIFF
--- a/layers/+completion/compleseus/README.org
+++ b/layers/+completion/compleseus/README.org
@@ -1,4 +1,4 @@
-#+TITLE: compleseus layer
+#+TITLE: Compleseus layer
 
 #+TAGS: completion|layer
 
@@ -33,7 +33,7 @@ layer that's listed last.
 
 * Key bindings
 
-| Key binding | Description   |
-|-------------+---------------|
-| ~M-o~       | embark-action |
-| ~C-r~       | history       |
+| Key binding | Description       |
+|-------------+-------------------|
+| ~M-o~       | embark-action     |
+| ~C-r~       | history           |
--- a/layers/+completion/compleseus/funcs.el
+++ b/layers/+completion/compleseus/funcs.el
@@ -96,6 +96,15 @@
   (interactive)
   (spacemacs/compleseus-search nil (projectile-project-root)))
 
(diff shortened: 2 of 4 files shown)


dalanicolai:
@thanhvg Do you know why the `consult-ripgrep` results get grouped? Because it seems that they do not get grouped by default (i.e. when trying with `emacs -Q`).

Also, can you explain how the narrow key is supposed to work (e.g. with `consult-ripgrep`)? Because pressing it just inserts it into the 'input field' (I wanted to check how it should work with `emacs -Q` but there, the results do not get grouped).

dalanicolai:
I've added a second commit for adding file classifier metadata to fasd search results.

thanhvg:
> @thanvg For `spacemacs/compleseus-search` I strongly prefer to have no initial-input when using `SPC s f` (as is the default in the Ivy and probably also helm layers. Now I have to delete the initial-input all the time) I think in those layers you can search for symbol under point in a project using `SPC /` in which case the symbol under point is used as initial-input. What do you think?
> 
I prefer `spc s f` that way but you can implement:
-  an option  to change its behavior.  
- or `spc s F` to ask for input.

`spc *` takes symbol at point to search while `spc /` prompts for input.

> Also, I prefer when the results do not get grouped. From trying with `emacs -Q` that also seems to be the default, but I can not easily find why the results (of the ripgrep search) in this layer get grouped. Is it configured/configurable in the layer somewhere?

I don't know but grouping is what the `consult` guys are very proud of and I think it is a great feature, with the search result use `C-J/K` to go by group. Speaking of going by group, `helm-ag` can do it too, I'm not sure if `counsel-rg` is capable of.

dalanicolai:
And I've added `C-h` to quickly delete parts of the input, and `?` for completion help to the `vertico-map`.

smile13241324 (approved):
Looks good

PR #15092, opened 2021-10-11
LaTeX: Fix broken key binding

sir4ur0n:
There was a sneaky `200b` character (zero width space), which breaks the display on the website:
image

The character is not visible in Github diff but can be seen by using an editor that displays those (e.g. Vim):
image


lebensterben:
@smile13241324 
https://github.com/sir4ur0n/spacemacs/commit/abac41e8d07231235cac3810019213ed7fa05a6b

Was this an accident or added on purpose to fix certain problem?

smile13241324:
@lebensterben I think I have copied it from somewhere, so this was rather an accident than anything else, lets remove it.

PR #15095, opened 2021-10-14
Add zonokai themes to themes-megapack layer

dalanicolai:
The documentation of the `themes-megapack` layer states that the layer provides
all themes included in the [theme gallery](https://themegallery.robdor.com/) by
Rob Merrell. However, this theme currently is missing from the layer.


smile13241324 (approved):
Looks good

Issue #15036, opened 2021-09-04
tracking issue: pyls is deprecated in lsp-mode


lebensterben:
See https://emacs-lsp.github.io/lsp-mode/page/CHANGELOG/#release-800

We should remove this option.

PR #15097, opened 2021-10-15
python: Update README.org

lebensterben:
It turns out that we've already migrated to pylsp from pyls a while ago, except that README.org is not updated.

closes https://github.com/syl20bnr/spacemacs/issues/15036


smile13241324 (approved):
:+1:

PR #15102, opened 2021-10-15
Update go layer docs since go get is deprecated

robbyoconnor:
See https://golang.org/doc/go-get-install-deprecation


smile13241324 (approved):
:+1:

PR #15104, opened 2021-10-18
Open special-mode buffer in motion-state (not evilified)

dalanicolai:
Special mode and its derived mode(s) buffers were made evilified in PRs #14995
and #15050. However, special-mode is too aggressive (as 'warned' for already in
 #14995); it makes it hard to overwrite its keybindings (e.g. `evil-local-set-key`,
or using `evil-evilified-state-map` on a derived map have no effect).

Therefore it makes more sense to open the buffers in motion-state instead. This
state also does not hijack the `q` keybindings as defined by special-mode (see
https://www.gnu.org/software/emacs/manual/html_node/elisp/Basic-Major-Modes.html),
although it does still hijack its `g` keybinding, but evil navigation is
considered to have higher priority than revert-buffer anyway.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -806,8 +806,8 @@ Other:
     section (thank to Daniel Nicolai)
   - New function =spacemacs/inherit-leader-keys-from-parent-mode= to
     inherit leader key bindings from another mode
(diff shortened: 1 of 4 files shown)


smile13241324 (approved):
Looks good

PR #15119, opened 2021-10-24
[doc] Fix spelling mistake

bhipple:


smile13241324:
I am not a native speaker but I think loose should be correct there. @lebensterben could you double check?

Kazark (approved):
`loose` as a verb means "to set free or cut off" but `lose` as a verb means "to no longer have that which you had"

Kazark:
Helpful: https://www.thefreedictionary.com/loose

Kazark:
Also: https://www.thefreedictionary.com/lose

lebensterben:
It's best to use subjunctive moode
>  If you chose to use =(global-company-mode)= then you would lose some advantages ..

But it's also okay to say
>  If you choose to use =(global-company-mode)= then you will lose some advantages ..

both are better than the imperative mood
> then you lose some advantages ..

smile13241324:
Thanks for the explanation guys :), I'll merge it.

PR #15125, opened 2021-10-26
helm-find-files-history requires prefix argument

kassick:
Calling spacemacs/helm-find-files with prefix would cause an error
instead of popping helm-find-files-history before helm-find-files


smile13241324 (approved):
Looks good

PR #15117, opened 2021-10-22
git: migrate git{attributes,config,ignore}-mode to git-modes

lebensterben:
fixes https://github.com/syl20bnr/spacemacs/issues/15111

PR #15127, opened 2021-10-27
[git] Remove helm-gitignore from git layer

eklitzke:
Related to #15117, this removes the helm-gitignore from the git layer because it
hasn't been updated to use the new git-modes package. The helm-gitignore
functionality isn't particularly useful anyway, it's just a utility to help
generate .gitignore files.

This change can be reverted if the upstream helm-gitignore package is fixed and
people find the functionality it provides to be useful.


smile13241324 (approved):
Looks good

PR #15135, opened 2021-11-03
fix a space issue introduced in #15098 in CHANGELOG.develop

robbyoconnor:

PR #15142, opened 2021-11-08
follow-up for #15135 with stray leading space in CHANGELOG.develop

robbyoconnor:
follow-up for #15135 with stray leading space in CHANGELOG.develop


robbyoconnor:
cc @smile13241324 @lebensterben

robbyoconnor:
That was fast @lebensterben

PR #15147, opened 2021-11-09
Fix typo doftile->dotfile

jrudel:
I didn't add anything to CHANGELOG.develop. I don't think this change warrants a bullet point and I don't mind not being on the contributors list.


lebensterben:
Thanks.

PR #15160, opened 2021-11-15
[scala] Shortcut keys for common/standard SBT commands

Kazark:
It's nice to have a way to use common SBT commands without having to enter a
Hydra, and without having to type them out or select from a history list. The
following should be standard and quite common:

+ `scalafmtAll`
+ `compile`
+ `test`

Added mnemonic keybindings that I think match Spacemacs conventions. They
behaviors also give better parity with the build capabilities of other major
modes, which frequently support compile, format and test functionality.


smile13241324 (approved):
Looks good

PR #15171, opened 2021-11-20
Fix #15169, snippets not loading correctly

### LINKED ISSUE #15169: Julia layer somehow hinders correct loading of (ya-)snippets
#### Description :octocat:
Julia layer somehow hinders correct loading of (ya-)snippets

- Start Emacs
- Install auto-completion layer
- Visit e.g. some .org buffer and type `SPC i s` note the amount of available snippets
- Install Julia layer (also press SPC f e R)
- After the Julia layer got installed, restart Spacemacs
- In the new session, again visit some .org buffer
- Again press `SPC i s`. Note that this time, no snippets are available

Missing snippets

No missing snippets

#### System Info :computer:
- OS: gnu/linux
- Emacs: 27.2
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. dee70b848)
- Graphic display: t
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(auto-completion emacs-lisp helm multiple-cursors treemacs julia)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG CAIRO SOUND DBUS GSETTINGS GLIB NOTIFY INOTIFY ACL LIBSELINUX GNUTLS LIBXML2 FREETYPE HARFBUZZ M17N_FLT LIBOTF ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XDBE XIM MODULES THREADS XWIDGETS LIBSYSTEMD JSON PDUMPER GMP

#### Backtrace :paw_prints:
```
<<BACKTRACE IF RELEVANT>>
```

dalanicolai:
The auto-completion layer seems not to have been designed/configured correctly.
So that, e.g., when using the julia layer, the `auto-completion/init-yasnippet`
function is called after the `yasnippet-snippets` package has been loaded.
Despite the `yasnippet-snippets` package containing an `eval-after-load
'yasnippet`, the form within that block is, for some reason, evaluated already
before `auto-completion/init-yasnippet` has been called.

Anyway, what happens currently is that first `(yasnippet-snippets-initialize)`
is evaluated, adding `yasnippet-snippets-dir` to the `yas-snippet-dirs` list.
Then `auto-completion/init-yasnippet` is called which contains the line `(setq
yas-snippet-dirs nil)`, setting the list to nil again.

As there is quite some logic involved (e.g. the order of evaluation is correct
when the julia layer is not being used), this commit fixes the reported issue by
simply calling (again) `(yasnippet-snippets-initialize)` from
`auto-completion/init-yasnippet` after the yas-snippet-dirs list has been
resetted to nil.

Any cleaner solution is welcome, but I would say this is a 'harmless' very
pragmatic quick fix/solution.

### DIFF
--- a/layers/+completion/auto-completion/packages.el
+++ b/layers/+completion/auto-completion/packages.el
@@ -335,7 +335,9 @@
         (when auto-completion-private-snippets-directory
           (if (listp auto-completion-private-snippets-directory)
               (setq yas-snippet-dirs (append yas-snippet-dirs auto-completion-private-snippets-directory))
-            (add-to-list 'yas-snippet-dirs auto-completion-private-snippets-directory))))
+            (add-to-list 'yas-snippet-dirs auto-completion-private-snippets-directory)))
+        ;; initialize again yasnippet-snippets, see PR #15171
+        (yasnippet-snippets-initialize))
       (spacemacs|add-toggle yasnippet
         :mode yas-minor-mode
         :documentation "Enable snippets."


smile13241324 (approved):
Good quick fix, we can look into the details later.

PR #15172, opened 2021-11-20
Add binding for markdown-table-align

deb0ch:
To make it consistent with Org layer :slightly_smiling_face:


smile13241324 (approved):
Looks good

Issue #15123, opened 2021-10-26
`evil-redirect-digit-argument` File mode specification error: (void-function evil-redirect-digit-argument) 


dhawangayash:
#### Description :octocat:
This isn't a bug per se but would like this documented as a known issue in spacemacs. 
Updating spacemacs packages leads to a broken dependency being pulled in from `evil-org-mode`.

#### Reproduction guide :beetle:
- Start Emacs
- Update spacemacs packages: `SPC f e U`
- This installs the broken version of `evil-org-mode`, package.

*Observed behaviour:* :eyes:
The following error message is printed on *Message* buffer. 
`File mode specification error: (void-function evil-redirect-digit-argument)`

*Expected behaviour:* :heart: :smile:
Everything should work without issues. :)

#### Backtrace :paw_prints:
NOT APPLICABLE.

#### Work around
This isn't a bug per se but would like this documented as a known issue in spacemacs. 

The latest version of `evil-org-mode` is broken owing to [this](https://github.com/Somelauw/evil-org-mode/issues/93)

So for any of you who has recently updated their packages will face the bug. The work around is provided in a comment [here](https://github.com/Somelauw/evil-org-mode/issues/93#issuecomment-949953955). 

I have tried it out and it works. I think this issue should be kept open to help others and closed once evil-org-mode comes with a newer fixed version.

enisozgen:
There is quick fix:

https://github.com/Somelauw/evil-org-mode/issues/93#issuecomment-949953955

bhoudebert:
There is also a very simple fix for evil-iedit-state here https://github.com/syl20bnr/evil-iedit-state/pull/37

lebensterben:
As suggested in other comments, this results from breaking change in upstream API.
Now we should wait for updates in affected packages.

smile13241324:
Added the temporary fix from @dalanicolai to heal the symptom. Still if the referenced pqckage is unmaintained we should switch to a maintained fork in the long run.

dalanicolai:
FYI, the improved fix in #15184 is great, but it requires recompilation of `evil-iedit-state.el` (recompilation after the macro from the fix has already been loaded).

gitprouser:
@dalanicolai I had to fetch the latest code change and I had to recompile all packages using `SPC r c`.

dalanicolai:
[...]

PR #15180, opened 2021-11-24
Add temporary fix for `evil-redirect-digit-argument` error

dalanicolai:
iedit does not work already for quite some time (at least 29 days, see https://github.com/syl20bnr/spacemacs/issues/15123).
A fix (see https://github.com/syl20bnr/evil-iedit-state/pull/37) has been proposed, but it
has not yet been merged (even after multiple reminders). Therefore, I propose to add
this temporary quick fix, that can/should be removed after that PR gets merged.

The fix simply defines the 'missing function' that prints that it should get removed after that PR gets merged when called.
In this way, iedit at least can be used (more or less) without problems.

### DIFF
--- a/layers/+spacemacs/spacemacs-evil/funcs.el
+++ b/layers/+spacemacs/spacemacs-evil/funcs.el
@@ -73,6 +73,17 @@ Otherwise, revert to the default behavior (i.e. enable `evil-insert-state')."
       (evil-escape-mode t)
     (evil-escape-mode -1)))
 
+(defun evil-redirect-digit-argument (_ _ _)
+  "This is a temporary fix.
+This is a temporary fix until the PR at URL
+ `https://github.com/syl20bnr/evil-iedit-state/pull/37' gets
+ merged. Please remove this function as soon as the mentioned PR
+ gets merged."
+  (message "This is a temporary fix until
+https://github.com/syl20bnr/evil-iedit-state/pull/37 gets merged.
+Please remove this function as soon as the mentioned PR gets
+merged."))
+
 
 ;; vi-tilde-fringe
 


layers/+spacemacs/spacemacs-evil/funcs.el
```diff
@@ -73,6 +73,17 @@ Otherwise, revert to the default behavior (i.e. enable `evil-insert-state')."
       (evil-escape-mode t)
     (evil-escape-mode -1)))
 
+(defun evil-redirect-digit-argument (_ _ _)
+  "This is a temporary fix.
+This is a temporary fix until the PR at URL
+ `https://github.com/syl20bnr/evil-iedit-state/pull/37' gets
+ merged. Please remove this function as soon as the mentioned PR
+ gets merged."
+  (print "This is a temporary fix until
```
lebensterben, line 82: message is better since it's recorded in message buffer

thanhvg:
Why don't just use quelpa to pull the fixed package from https://github.com/kassick/evil-iedit-state? Beside I think evil-iedit-state project should be handled over to active Spacemacs maintainers since the big boss is technically retired from Emacs :(.

layers/+spacemacs/spacemacs-evil/funcs.el
dalanicolai: Message is probably better indeed, so I have changed it. However, print also seems to get recorded in the message buffer, I don't really get the difference and can not find info about it

dalanicolai:
> Why don't just use quelpa to pull the fixed package from https://github.com/kassick/evil-iedit-state? Beside I think evil-iedit-state project should be handled over to active Spacemacs maintainers since the big boss is technically retired from Emacs :(.

Yeah, I thought about forking/using some fork too, but I assumed that the big boss was just busy. Does anybody know the real situation? Otherwise, I do agree with @thanhvg...

smile13241324 (approved):

smile13241324:
Sorry I have no idea whats the situation with @syl20bnr. Lets have this fixed in spacemacs for now.

If you are still maintaining this package please leave a short comment. If not I would also vote for switching to a maintained fork of the package.

lebensterben:
since the upstream has fixed the issue, this PR could be reverted now.

Issue #14246, opened 2020-12-30
New Layer: Emacs Tree Sitter


muhmud:
https://github.com/ubolonton/emacs-tree-sitter

Thanks!

smile13241324:
I assume this is a feature request to have tree-sitter, a native structural parsing library, integrated into Spacemacs? I am not sure this makes much sense, this library primarily targets emacs packages which in turn we group into pre-configured layers. This is to have tree sitter used in spacemacs it just has to be used by one of the packages integrated. Also binary installation of the C or Rust part of tree-sitter is covered by the standard melpa installation.

So this is no change for Spacemacs instead you should open an issue upstream on your individual emacs package you would want to have native support for.

muhmud:
You could be right, and I should've done a better write up, but I see this as providing general syntax highlighting even without any other emacs language packages installed.

The highlighting definitions provided with emacs-tree-sitter are also somewhat basic. I've done a fair few modifications on the Java one to make them more suitable for myself. So, with a layer for this you would have a place to provide and maintain better rule definitions in a central place.

lebensterben:
I agree with @smile13241324 
And I'd like to see some demonstration of this new package. It seems like a library for elisp developer (of language syntax highlighting) instead of a package for end-users.

yyoncho:
JFYI tree-sitter ships with highlighting which tends to be much faster and more precise than what emacs major modes have. I don't have experience with it but this is the feedback that I am getting(especially for cc-mode related modes). 

IMHO bare tree sitter layer should be added and potentially it should be advertised as a default layer(when user's emacs supports dynamic modules) if the aforementioned perf and precision gains are real.

PR #15064, opened 2021-09-26
[WIP] Tree Sitter Layer

sdwolfz:
There is much work to be done here, but this is the initial support for tree-sitter. You can enable it with...
```elisp
(tree-sitter
  :variables
    tree-sitter-toggle 'global)
```
... in your `dotspacemacs-configuration-layers`.

It seems to work, and appears to load files faster, but the colors are... weird, at least for me with the monokai theme, I'll probably have to deal with that upstream.

For the final PR I'll make it a default layer and have it toggle on demand for the supported languages. When I'm convinced it's mature enough I'll have it globally by default.

PR #15185, opened 2021-11-28
Add basic tree-sitter layer

eeshugerman:
Closes https://github.com/syl20bnr/spacemacs/issues/14246

In the process of opening this PR I realized there's prior art: https://github.com/syl20bnr/spacemacs/pull/15064. ~But from what I can tell it's a different approach which wouldn't make sense to incorporate here.~

EDIT: On closer inspection I realize #15064 is basically the same approach as this; the main difference is that #15064 would allow the user to turn on `tree-sitter-hl-mode` for specified major modes or globally, whereas this PR only provides the global switch.

### DIFF
--- a/layers/+misc/tree-sitter/packages.el
+++ b/layers/+misc/tree-sitter/packages.el
@@ -49,8 +49,7 @@
 (defun tree-sitter/init-tree-sitter-indent ()
   (use-package tree-sitter-indent
     :if tree-sitter-indent-enable
-    ;; missing autoload https://codeberg.org/FelipeLema/tree-sitter-indent.el/pulls/15
-    ;; :defer t
+    :defer t
     :init
     (progn
       (add-hook 'rust-mode-hook #'tree-sitter-indent-mode))))
--- a/layers/+misc/tree-sitter/packages.el
+++ b/layers/+misc/tree-sitter/packages.el
@@ -54,9 +54,7 @@
     (progn
       (add-hook 'rust-mode-hook #'tree-sitter-indent-mode))))
 
-;; how can we avoid listing these explicitly?
-;; pull them out of ts-fold at init-time somehow?
-(defconst tree-sitter//ts-fold-supported-major-modes
+(defconst tree-sitter//ts-fold-supported-major-mode-hooks
   '(agda-mode-hook
     sh-mode-hook
     c-mode-hook
@@ -91,7 +89,7 @@
     :init
     (progn
       (when tree-sitter-fold-enable
-        (dolist (mode-hook tree-sitter//ts-fold-supported-major-modes)
+        (dolist (mode-hook tree-sitter//ts-fold-supported-major-mode-hooks)
           (when (boundp mode-hook)
             (add-hook mode-hook #'ts-fold-mode)
             (when tree-sitter-fold-indicators-enable
--- a/layers/+misc/tree-sitter/packages.el
+++ b/layers/+misc/tree-sitter/packages.el
@@ -1,0 +1,103 @@
+;;; packages.el --- tree-sitter layer packages file for Spacemacs.
+;;
+;; Copyright (c) 2012-2021 Sylvain Benner & Contributors
+;;
+;; Author: Elliott Shugerman <eeshugerman@gmail.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+;;
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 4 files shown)


layers/+misc/tree-sitter/packages.el
```diff
@@ -0,0 +1,100 @@
...
+
+(defun tree-sitter/init-tree-sitter-indent ()
+  (use-package tree-sitter-indent
+    :if tree-sitter-indent-enable
+    :defer t
+    :init
+    (progn
+      (add-hook 'rust-mode-hook #'tree-sitter-indent-mode))))
+
+(defconst tree-sitter//ts-fold-supported-major-mode-hooks
```
eeshugerman, line 57: Would be nice if there were a way to avoid listing these out explicitly and still defer loading. There's `global-ts-fold-mode` but it breaks folding for unsupported major modes. Would it make sense to ask the developer to accommodate this, either in `global-ts-fold-mode` or by autoloading something like `ts-fold-supported-modes`?

In other words I'm not sure how this usually works. If this is the way to do it then it works for me but if there's a better way to do it I'm up for it.

eeshugerman:
@sdwolfz would you be interested in reviewing this?

eeshugerman:
converting to draft for now because the install/build of `tree-sitter-langs` has broken somehow

sdwolfz:
It's going to take me a bit of time as I have a lot to deal with right now.

eeshugerman:
@sdwolfz No rush! And sorry for the useless "marked as draft", "marked as ready for review" notifications -- I thought this was broken for a minute there and then realized the issue was unrelated.

layers/+misc/tree-sitter/packages.el
smile13241324: Hmmmm I think having a list of supported modes in the emacs package would be the way to go. But thats just an improvement, for the first version this is ok.

smile13241324 (approved):

eeshugerman:
@smile13241324 Thanks for the review and merge!

I noticed the new tree-sitter layer is still not listed in [the docs](https://develop.spacemacs.org/layers/LAYERS.html) :thinking:. Any idea as to why that would be? Is there a job that needs to be kicked off manually for new layers perhaps?

smile13241324:
Not sure maybe the doc generation fails again due to an invalid link I can have a look this evening, still studies comew first ;).

smile13241324:
I think I have solved the issue, CI was failing since we increased the min version from 26.3 to 27.1 for emacs. Now the docs should also update themselves in the next few days.

lebensterben:
We probably should move the layer to `+tools`.
Consider `lsp` layer which is in `+tools` now and provide similar cross-major-mode functionalities.

eeshugerman:
IMO `lsp` is out of place in `+tools` since besides `lsp`, `+tools` is mostly external/standalone tools that are typically used outside of Emacs, or in a few cases the tool is Emacs-based but is for a specific purpose and is used on an ad-hoc basis.

It doesn't matter much to me where `tree-sitter` goes, but that's why I chose `+misc` over `+tools`. I do see the logic in putting it alongside `lsp`, and I realize that at this point it would be much less disruptive to move `tree-sitter` than `lsp`.

smile13241324:
Moved the layer into tools, where also node is situated which is also a pure support layer. Lets keep it there till we reorganise.
I have also renamed the list of supported modes to be consistent with our conventions for internal variables.

eeshugerman:
@smile13241324 thanks for taking care of that cleanup!

About the docs, there's now a [tree-sitter section in LAYERS.html](https://develop.spacemacs.org/layers/LAYERS.html#tree-sitter) :tada: , but [+tools/tree-sitter/README.html](https://develop.spacemacs.org/layers/+tools/tree-sitter/README.html) doesn't exist :confused:. Any idea why that could be? Is it just a matter of time for it to get picked up?

smile13241324:
Hmmm good point I think the web page is updated automatically but I am not 100% sure, lets ask @JAremko about this.

JAremko:
Bot creates pull requests but I merge them manually (for the safety sake)

eeshugerman:
@JAremko got it, thanks. Any idea why [+tools/tree-sitter/README.html](https://develop.spacemacs.org/layers/+tools/tree-sitter/README.html) is 404ing?

JAremko:
@eeshugerman  That's strange since it's listed https://develop.spacemacs.org/layers/LAYERS.html#tree-sitter :thinking:

JAremko:
And looks like it was exported https://app.circleci.com/pipelines/github/syl20bnr/spacemacs/3446/workflows/caac66ec-e570-4bb8-be7e-eb0e7cce04cc/jobs/69104?invite=true#step-106-689 :thinking:

JAremko:
But there is no tree-sitter here https://github.com/syl20bnr/develop.spacemacs.org/tree/gh-pages/layers/%2Btools :thinking:

JAremko:
Aha ok. I see. It's not in the patch. I guess Bot stopped adding new files for some reason :thinking:

JAremko:
@eeshugerman  fixed :+1:

eeshugerman:
Nice, thanks!

PR #15189, opened 2021-11-30
Add pdf occur follow mode spacemacs binding

dalanicolai:
Follow mode in the pdf-occur buffer is very handy, and deserves a specific
Spacemacs keybinding (mainly for discoverability/documentation, as its name
`next-error-follow-minor-mode`, putting it mildly, is not very intuitive).

### DIFF
--- a/layers/+readers/pdf/packages.el
+++ b/layers/+readers/pdf/packages.el
@@ -132,7 +132,10 @@
         "g"              'pdf-occur-revert-buffer-with-args
         "r"              'pdf-occur-revert-buffer-with-args
         "*"              'spacemacs/enter-ahs-forward
-        "?"              'evil-search-backward))))
+        "?"              'evil-search-backward)
+      (spacemacs/declare-prefix-for-mode 'pdf-occur-buffer-mode "mt" "toggles")
+      (spacemacs/set-leader-keys-for-major-mode 'pdf-occur-buffer-mode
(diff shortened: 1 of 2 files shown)


layers/+readers/pdf/packages.el
```diff
@@ -131,8 +131,13 @@
         "q"              'tablist-quit
         "g"              'pdf-occur-revert-buffer-with-args
         "r"              'pdf-occur-revert-buffer-with-args
+        "C-j"            'next-error
```
lebensterben, line 134: What does "error" mean in this context?

dalanicolai: Ah, I thought that I had removed these bindings. Well, I am not sure exactly what it means, but what is does is jump to next line and 'follow`. The name of the command is very misleading (you can read in the docstring of `next-error-follow-minor-mode`, that the 'error' functionality also works for occur and diff mode).
I will probably suggest to the Emacs developers to create a more useful alias for it (in Emacs).

So the `'next-error-follow-minor-mode` is useful, because it does not select the pdf-buffer when using it. `next/previous` error will select the pdf-buffer when using them, so I don't find them very useful for `occur-mode`. I will remove them (which I had wanted to do already)

lebensterben (approved):

smile13241324 (approved):

PR #15190, opened 2021-11-30
Add documentation for most important org shortcuts

dalanicolai:
Some primary commands in org-mode are for sure `org-meta-return` and
`org-metaleft/right`. However, documentation for their shortcuts is lacking.

Additionally, using org-babel without `yasnippets` is self-torture, therefore
when using org-babel the keybinding for inserting yasnippets is highly relevant.

This commits adds documentation for these shortcuts.

### DIFF
--- a/layers/+emacs/org/README.org
+++ b/layers/+emacs/org/README.org
@@ -656,6 +656,17 @@ A [[https://github.com/syl20bnr/spacemacs/blob/develop/doc/DOCUMENTATION.org#lay
 Please see the [[https://github.com/Somelauw/evil-org-mode/blob/master/doc/keythemes.org][evil-org documentation]] for additional instructions on customizing
 =evil-org-mode=.
 
+*Insert state*
+In evil insert-state, use the introspection functions under =M-m
+h d= and =M-RET= to discover keybindings. The following table shows keybindings
+for some of the most frequently used commands
+
+| Key binding   | Description                                       |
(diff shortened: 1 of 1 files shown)


layers/+emacs/org/README.org
```diff
@@ -773,31 +784,32 @@ are also available.
...
+| ~SPC m b p~ | org-babel-previous-src-block                   |
+| ~SPC m b r~ | org-babel-goto-named-result                    |
+| ~SPC m b s~ | org-babel-execute-subtree                      |
+| ~SPC m b t~ | org-babel-tangle                               |
+| ~SPC m b u~ | org-babel-goto-src-block-head                  |
+| ~SPC m b v~ | org-babel-expand-src-block                     |
+| ~SPC m b x~ | org-babel-do-key-sequence-in-edit-buffer       |
+| ~SPC m b z~ | org-babel-switch-to-session                    |
+| ~SPC m b Z~ | org-babel-switch-to-session-with-code          |
+| ~SPC i s~   | insert yasnippet (e.g. source block templates) |
```
Kazark, line 812: I believe this binding is from the yasnippets layer---is it documented there?

dalanicolai: Yes, I was aware that I am repeating things here. But I think that does more good than it does harm. However, I do agree that it is better to point to the keybinding (i.e. the functionality) in that layer. I will fix that...

PR #15194, opened 2021-12-04
[core] less ".el" suffix for support both *.el and *.elc

sunlin7:
Hi, 

This change try to remove '.el' suffix to support loading `*.elc` or `*.el`.

I byte-compile the `Spacemacs/core/*.el` to `*.elc` files, and compress the `*.el` to `*.el.gz` (Like a emacs release).

Then Spacemacs will failed for there is no `*.el` at startup. This change can make it work for both `*.el` and `*.elc`.

Please help review. Thanks

B.R.
Lin


thanhvg:
This is great, I think your MR is fine but I never play with byte compiling of Spacemacs so my questions here are just beginner's level:

- Do you notice any performance gain?
- Why do you have to zip it? If you leave the el and elc files together I think the later one will be picked up won't it?
- Can we byte-complie the all Spacemacs files not just core ones?
- Will this work with Emacs 28 native-comp feature? Is it possible to native-compile Spacemacs els?

Thanks

sunlin7:
Hi @thanhvg 
> * Do you notice any performance gain?
Yes, there is performance enhance.
> * Why do you have to zip it? If you leave the el and elc files together I think the later one will be picked up won't it?
Yes, load elisp file without extension will try pick up `*.elc` first. And compressing `*.el` to `*.el.gz` is just for saving the disc space, like vanilla Emacs behavior.
> * Can we byte-complie the all Spacemacs files not just core ones?
It's unkown. The scope is out of this merge request. And this merge request can be the beginning of Spacemacs supporting  byte-compiling elisp files.
> * Will this work with Emacs 28 native-comp feature? Is it possible to native-compile Spacemacs els?
It's unkown, but I pretty sure that native-comp feature always require the `*.elc` files are exists, while Spacemacs is not support `*.elc` without this pull request, again, this pull request can be the beginning for Spacemace supporting native-comp feature.

sunlin7:
Hi @thanhvg ,

It's a big scope for Spacemacs supporting byte-compile (or native-comp), there're so many layers, no-one can enable all the Spacemacs layers and have a testing with the layers one by one.

So we can make sure the core files support byte-compile, then the layers users can try byte-compile by themself.

lebensterben:
also, Emacs has built-in support for viewing .el.gz.

Actually if Emacs is built from source, all .el files are compressed by default.

so nothing to worry for .el.gz

sunlin7:
@lebensterben  Yes, totally. 
And this change is safe for all users.
Could you help approve this merge request? Thanks.

sunlin7:
ping

smile13241324 (approved):
Looks good, thanks for this @sunlin7, this could indeed be the beginning of compilation support for spacemacs.

sunlin7:
@smile13241324 thank you more!

sunlin7:
And byte-compile (without native-compile) files in Spacemacs core/ folder work for me, no issue be found yet.
But native-compile are not working in my local, seems have many critical issues.
I'll create new PR to enhance Spacemacs on byte-compile or native-compile feature.

lebensterben:
@sunlin7 
What's wrong with native compile? Any error message?

PR #15199, opened 2021-12-08
Fix #15198: Apply ansi-colors to compilation-mode derived modes

### LINKED ISSUE #15198: Apply Ansi colors to "compilation-mode" derived modes
#### Description :octocat:
The output generated from running `dap-java-run-test-class` prints the ansi-colors as escaped chars, instead of applying the colors to the output.

The Spacemacs process hook `spacemacs-visual//compilation-buffer-apply-ansi-colors` is already bound by default when running the above function, but unfortunately it does work. The reason is that the above function runs for `compilation-mode` major modes only, and the dap-java uses a mode (`dap-server-log-mode`) derived from `compilation-mode`

- Start Emacs
- Enable java and lsp layers
- Create a sample maven java projects
- Create sample Java test class
- Execute the test class with `dap-java-run-test-class`
- Notice the ansi colors escaped chars in the output buffer

- Notice the ansi colors escaped chars in the output buffer

- Ansi-color codes should be interpreted as colors, so for instance, instead of printing `^[[36m` it should display the subsequent text in `Cyan`.

#### System Info :computer:
- OS: darwin
- Emacs: 27.2
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 1a4912e51)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(rust go html javascript yaml python java auto-completion better-defaults emacs-lisp git helm lsp dap markdown multiple-cursors org
      spell-checking syntax-checking treemacs)
```
- System configuration features: NOTIFY KQUEUE ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS MODULES THREADS JSON PDUMPER GMP

#### Backtrace :paw_prints:
```
```

rinconjc:
Hi,

This change should enable applying ANSI colors to `dap-java` output buffers, and in general to the output buffer of any mode that derives from `compilation-mode`. 
 
Cheers,

### DIFF
--- a/layers/+spacemacs/spacemacs-visual/funcs.el
+++ b/layers/+spacemacs/spacemacs-visual/funcs.el
@@ -25,7 +25,7 @@
 ;; ansi-colors
 
 (defun spacemacs-visual//compilation-buffer-apply-ansi-colors ()
-  (when (eq major-mode 'compilation-mode)
+  (when (memq 'compilation-mode (parent-mode-list major-mode))
     (let ((inhibit-read-only t))
       (goto-char compilation-filter-start)
       (ansi-color-apply-on-region (line-beginning-position) (point-max)))))


layers/+spacemacs/spacemacs-visual/funcs.el
```diff
@@ -25,7 +25,7 @@
 ;; ansi-colors
 
 (defun spacemacs-visual//compilation-buffer-apply-ansi-colors ()
-  (when (eq major-mode 'compilation-mode)
+  (when (member 'compilation-mode (parent-mode-list major-mode))
```
lebensterben, line 28: I think memq is sufficient

smile13241324 (changes requested):
Please have a look at the comment from lebensterben and replace the more expensive member function with memq.

smile13241324 (approved):

PR #15201, opened 2021-12-08
Make dotspacemacs config diffs a little smaller...

sr105:
Move closing parens on all dotspacemacs/user-* functions to a new line so the
function documentation and closing paren are not part of the config diff when a
user runs `SPC f e D`. This should make it faster to check config changes versus
the template.


lebensterben:
by elisp standard, closing parentheses should not take a new line.

lebensterben (approved):

sr105:
> by elisp standard, closing parentheses should not take a new line.

agreed. I think this is a special case since those functions are designed to be different on every single install, and we want to minimize the effort required to keep current.

smile13241324 (approved):

PR #15204, opened 2021-12-11
loading tex package before PDF tool setting

hackartists:
## Issue

* `PDF tools` will be set to `TeX-view-program-selection` while lsp is starting.
  * It causes the below error because the variable doesn't be loaded in advance.

```
Error (use-package): lsp-latex/:config: Symbol’s value as variable is void: TeX-view-program-selection
```

## Solution

* `TeX-view-program-selection` is defined in `auctex/tex.el` package.
  * We need to load the package before using the variable.

### DIFF
--- a/layers/+lang/latex/funcs.el
+++ b/layers/+lang/latex/funcs.el
@@ -60,6 +60,7 @@
   (when latex-view-with-pdf-tools
     (if (configuration-layer/layer-used-p 'pdf)
         (progn
+          (require 'tex)
           (setf (alist-get 'output-pdf TeX-view-program-selection) '("PDF Tools"))
           (when latex-view-pdf-in-split-window
             (require 'pdf-sync)


emiltoacs:
Hi hackartists, the documentation of Spacemacs tells us to avoid the use of `require` as much as possible,  the layer are supposed to call use-package when needed. I committed a patch to solve the same warning.

lebensterben:
@hackartists Please verify whether the PR fixed the issue.

hackartists:
> @hackartists Please verify whether the PR fixed the issue.

Yes, I've tested it. And I am using this code for my emacs.
If you have some issue when you try it, let me know what issue you faced.
@lebensterben

lebensterben:
@hackartists 
Sorry for the confusion.
Does the PR by @emiltoacs fix your issue?

hackartists:
> @hackartists Sorry for the confusion. Does the PR by @emiltoacs fix your issue?

You said about #15214 , right?
In the PR, @emiltoacs has just eliminated a part of initialization of pdf tool. 
So, the PR by @emiltoacs  can make viewing PDF files not properly work.

The root cause of `Tex-view-program-selection` is reading the variable before initialization in `tex` package.
Therefore, this PR imports the proper package `tex` in advance.

lebensterben:
@hackartists 
this is weird.
the function is called here https://github.com/syl20bnr/spacemacs/blob/8fce28cb3cfeebe397fb8a05db77488f5182413d/layers/%2Blang/latex/packages.el#L81
which is within the use-package call for Tex package
https://github.com/syl20bnr/spacemacs/blob/8fce28cb3cfeebe397fb8a05db77488f5182413d/layers/%2Blang/latex/packages.el#L51

that's to say, the function is called when the Tex package is already loaded. And there should be no need to have (require 'tex)

emiltoacs:
Yes I remove the call to `spacemacs//latex-setup-pdf-tools` inside the `init-lsp-latex` function on purpose, and it does not break break pdf tools. As explained in the comment https://github.com/syl20bnr/spacemacs/pull/15214#issue-1083140171 the initialization of pdf tools is already done in the initialization of auctex which is its proper place. 
So as @lebensterben  said https://github.com/syl20bnr/spacemacs/pull/15204#issuecomment-998481636 and it's a global recommendation of spacemacs you should not use require inside the initialization function.

```
Do not use require. If you find yourself using =require=, you are almost
certainly doing something wrong. Packages in Spacemacs should be loaded through
auto-loading, and not explicitly by you. Calls to =require= in package init
functions will cause a package to be loaded upon startup. Code in an =:init=
block of =use-package= should not cause anything to be loaded, either. If you
need a =require= in a =:config= block, that is a sign that some other package is
missing appropriate auto-loads.
```

@hackartists  if you have issues with pdf viewing in latex that maybe because of the pdf layer if you use it. It requires to build the epdfinfo program properly.

hackartists:
> @hackartists this is weird. the function is called here
> 
> https://github.com/syl20bnr/spacemacs/blob/8fce28cb3cfeebe397fb8a05db77488f5182413d/layers/%2Blang/latex/packages.el#L81
> 
> 
> which is within the use-package call for Tex package
> https://github.com/syl20bnr/spacemacs/blob/8fce28cb3cfeebe397fb8a05db77488f5182413d/layers/%2Blang/latex/packages.el#L51
> 
> that's to say, the function is called when the Tex package is already loaded. And there should be no need to have (require 'tex)

Oh, thank you. I will check it out with PR by @emiltoacs .

PR #3632, opened 2015-11-02
VCS microstate with multiple margin tools.

ralesi:
Add a microstate for staging hunks with diff-hl

PR #15213, opened 2021-12-17
git-gutter: Remove custom fringe bitmaps

real-or-random:
The config for git-gutter currently sets custom fringe bitmaps which
are optimized for half-width fringes.

I think that's an arbitrary choice. There's no spacemacs option which
enables half-width fringes and as far as I can tell, we don't optimize
for half-width fringes anywhere else. So we should optimize the
package defaults for the spacemacs default of full-size fringes.

If users set half-width fringes manually, they can customize the
bitmaps manually. Even if they don't, the bitmaps are still visible,
they just don't look as nice.

### DIFF
--- a/layers/+source-control/version-control/packages.el
+++ b/layers/+source-control/version-control/packages.el
@@ -183,30 +183,7 @@
        (with-eval-after-load 'git-gutter
          (require 'git-gutter-fringe)))
       (setq git-gutter-fr:side (if (eq version-control-diff-side 'left)
-                                   'left-fringe 'right-fringe)))
-    :config
-    (progn
-      ;; custom graphics that works nice with half-width fringes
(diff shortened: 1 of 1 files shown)


real-or-random:
Screenshots with full-width fringes (the default)

Before: Screenshot from 2021-12-17 10-39-07, After: Screenshot from 2021-12-17 10-37-20

The symbols on the left are hardly visible, I think.

lebensterben:
I've no idea why this is first added in https://github.com/syl20bnr/spacemacs/pull/3632

It was quite too long ago...

I've no issue with this PR but let's wait for approvals from other maintainers.

lebensterben (approved):

smile13241324 (approved):
No idea why custom fringes have been added but when the original ones work well why keep the exception. So lets have this merged and see how it plays out.

Issue #15212, opened 2021-12-15
Update packages doesn't finish. "Function is void read-answer".


Atrus7:
#### Description :octocat:
Update packages broken, Function is void read-answer.

My Emacs (26) doesn't have that function
Breaking on Line 2164 in ~/.emacs.d/core/core-configuration-layer.el

#### Reproduction guide :beetle:
- Start Emacs
- <SPC> f e U to update packages

*Observed behaviour:* :eyes: :broken_heart:
Doesn't update. Error Message is `Function is void: read-answer`

*Expected behaviour:* :heart: :smile:
Updates packages.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 26.1
- Spacemacs: 0.300.0
- Spacemacs branch: nil (rev. 387d16545)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
((auto-completion :variables auto-completion-return-key-behavior 'complete auto-completion-enable-snippets-in-popup t)
 helm git
 (org :variables org-want-todo-bindings t)
 (deft :variables deft-zetteldeft t)
 (shell :variables shell-default-shell 'shell)
 spell-checking version-control
 (c-c++ :variables c-c++-default-mode-for-headers 'c++-mode c-c++-enable-clang-support t)
 common-lisp emacs-lisp haskell html javascript swift markdown csv
 (spell-checking :variables spell-checking-enable-by-default nil)
 python scheme vimscript yaml dash erc graphviz ibuffer semantic cf-home syntax-checking util-funcs cf-cc cf-calendar cf-desktop cf-irc cf-ide cf-linux cf-lisp cf-mac cf-opinions cf-org cf-scheme cf-ssh cf-writing)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GSETTINGS NOTIFY ACL LIBSELINUX GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 THREADS LIBSYSTEMD LCMS2

My workaround was to avoid the confirmation with a prefix, but this isn't ideal. 

I think this is breaker:
cac010555 * Add option to select packages for update

Atrus7:
@dalanicolai hey, could you take a look. Not 100% sure if emacs 26 is supported, (first time submitting) but if it is, I think this will break others too.

[...]

PR #15215, opened 2021-12-17
Fix package update for emacs versions < 26.2 (issue #15212)

dalanicolai:
See #15212

The selective update functionality uses the function `read-answer` which is not available in Emacs version < 26.2.
Therefore, this PR restores the initial update functionality for Emacs versions < 26.2.


Atrus7 (approved):
works in my emacs (26.1).

PR #15216, opened 2021-12-17
[languagetool] fix for non standard language code

emiltoacs:
In the language tool layer, the `languagetool-get-language` function is used to
deduce the correct language code for languagetool from the variable
`ispell-local-dictionary`, `ispell-dictionary`, or from the list
`ispell-dicts-name2locale-equivs-alist` by replacing dash by underscore. However
many languagetool codes are not formatted consistently with what can be found in
ispell, check `languagetool --list <file>` for the exact list.

Therefore I created this list of conversion between ispell codes and languagetool
codes.


lebensterben:
please use hash table instead of linked list.

emiltoacs:
@lebensterben isn't it a bit overkill to use hash table for a few tens of values ? Glancing at the Spacemacs code I see that make-hash-table is usually used for hundreds of entries. 
I'm not experienced with emacs-lisp data structure so it's really a question so if it's better to use hash-table in that case I'll commit

lebensterben:
at very least you can reorder entries by usage (L1 population).

emiltoacs:
that make sense so I keep the list and just reorder the element

lebensterben:
@emiltoacs 
Are you sure you've included are languages `ispell` supports?

The following are the languages languagetool supports:
```txt
ar Arabic
ast-ES Asturian
be-BY Belarusian
br-FR Breton
ca-ES Catalan
ca-ES-valencia Catalan (Valencian)
da-DK Danish
de German
de-AT German (Austria)
de-CH German (Swiss)
de-DE German (Germany)
de-DE-x-simple-language Simple German
el-GR Greek
en English
en-AU English (Australian)
en-CA English (Canadian)
en-GB English (GB)
en-NZ English (New Zealand)
en-US English (US)
en-ZA English (South African)
eo Esperanto
es Spanish
es-AR Spanish (voseo)
fa Persian
fr French
ga-IE Irish
gl-ES Galician
it Italian
ja-JP Japanese
km-KH Khmer
nl Dutch
nl-BE Dutch (Belgium)
pl-PL Polish
pt Portuguese
pt-AO Portuguese (Angola preAO)
pt-BR Portuguese (Brazil)
pt-MZ Portuguese (Moçambique preAO)
pt-PT Portuguese (Portugal)
ro-RO Romanian
ru-RU Russian
sk-SK Slovak
sl-SI Slovenian
sv Swedish
ta-IN Tamil
tl-PH Tagalog
uk-UA Ukrainian
zh-CN Chinese
```

lebensterben:
Also, it seems that languagetool follows IETF language tag.
Specifically, each language tag is one of
- A two-letter language code from ISO-639-1
- A two-letter language code from ISO-639-1, a hyphen, and a 2 letter country code from ISO 3166-1-alpha-2
where the second form is used for regional variants only.

So it's not correct to say the naming of tags is inconsistent.

emiltoacs:
What I meant by inconsistent was the mapping between ispell and language-tool. My main motivation for this patch was because of the french dictionary that does not work with the actual code. In ispell the dictionnary is either `francais`, `fr_FR`, `fr_CA` and other variants for other countries but in languagetool it's only `fr` therefore replacing underscore by dash makes languagetool loading `fr-FR` or `fr-CA` which does not exits,

For the ispell language I just patched the code so I took only the language present in the previous variable `ispell-dicts-name2locale-equivs-alist` which is maybe a bit too restrictive I agree. So I could change the code to proceed to the translation from language that comes from this list only and let the other language code as there are.

emiltoacs:
I guess French and Spanish are some exception where the mapping from underscore to dash does not work to translate ispell code to languagetool so I think I'll write a list with only these kind of exceptions and let the other code be handled as they where.

I'll take so time later to write a correct patch in this sense.

emiltoacs:
I created this list of special case where languagetool dictionary name exits only in the form of a two-letter language code although the ispell-dictionary name exists only in the four-letter form.  I reintroduced the underscore to dash replacement from the original code as an alternative to special case, any other language code will be passed to languagetool unchanged.

smile13241324 (approved):

Issue #15219, opened 2021-12-18
ivy-occur-mode-map keybindings are shadowed by evil-mode


czqhurricnae:
#### System Info :computer:
- OS: darwin
- Emacs: 27.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 5bcedd191)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
(ivy better-defaults ranger colors prodigy search-engine graphviz
     (syntax-checking :variables syntax-checking-enable-by-default t syntax-checking-enable-tooltips t)
     (spell-checking :variables spell-checking-enable-by-default nil)
     (vinegar :variables vinegar-reuse-dired-buffer t)
     (spacemacs-layouts :variables layouts-enable-autosave nil layouts-autosave-delay 300)
     (git :variables git-magit-status-fullscreen t magit-push-always-verify nil magit-save-repository-buffers 'dontask magit-revert-buffers 'silent magit-refs-show-commit-count 'all magit-revision-show-gravatars nil)
     (ibuffer :variables ibuffer-group-buffers-by 'projects)
     (auto-completion :variables auto-completion-enable-sort-by-usage t auto-completion-enable-snippets-in-popup t auto-completion-enable-help-tooltip nil :disabled-for org markdown)
     (osx :variables osx-dictionary-dictionary-choice "Simplified Chinese - English" osx-command-as 'super)
     restclient
     (gtags :disabled-for clojure emacs-lisp javascript latex python shell-scripts)
     (shell :variables shell-default-shell 'eshell)
     latex
     (javascript :variables javascript-disable-tern-port-files nil add-node-modules-path t)
     (typescript :variables typescript-fmt-on-save nil typescript-fmt-tool 'typescript-formatter)
     html react
     (python :variables python-test-runner
             '(nose pytest))
     (c-c++ :variables c-c++-default-mode-for-headers 'c++-mode)
     emacs-lisp markdown
     (org :variables org-want-to-do-binding t org-enable-github-support t org-enable-reveal-js-support t org-enable-roam-support t)
     fasd
     (spacemacs-layouts :variables layouts-enable-autosave nil layouts-autosave-delay 300)
     theming
     (chinese :variables chinese-default-input-method 'pinyin chinese-enable-youdao-dict t)
     tern hurricane)
```
[...]

PR #15222, opened 2021-12-21
Fix keybindings in ivy-occur-grep-mode

dalanicolai:
Ivy occur grep mode is a read-only mode that is very similar to ivy-occur mode,
and has one extra keybinding under `w` to change to `wgrep-mode`.

Therefore, instead of starting it normal mode, it should start with the
`overriding-map` similar as how it is done for ivy-occur-mode.

Theoretically (this currently does not work here), this should make the `w` call
`spacemacs/ivy-wgrep-change-to-wgrep-mode` which is perfectly fine.

For some reason, after setting this, my Spacemacs opens ` ivy-occur-grep-mode`
in motion-state (despite it is not derived from `special-mode`, see  #15104`). But
anyway, this fix should work fine, and it is even working fine here with special-mode
(I have no time to debug this irrelevant behavior on my system now).

### DIFF
--- a/layers/+completion/ivy/packages.el
+++ b/layers/+completion/ivy/packages.el
@@ -276,7 +276,7 @@
       (global-set-key (kbd "C-c C-r") 'ivy-resume)
       (global-set-key (kbd "<f6>") 'ivy-resume)
       ;; Occur
-      (evil-set-initial-state 'ivy-occur-grep-mode 'normal)
+      (evil-make-overriding-map ivy-occur-grep-mode-map)
       (evil-make-overriding-map ivy-occur-mode-map 'normal)
       (dolist (mode-map (list ivy-occur-mode-map ivy-occur-grep-mode-map))
         (define-key mode-map "g" nil)


lebensterben (approved):

lebensterben:
this probably could solve other key binding shadowing problems as well.

dalanicolai:
As reported by @czqhurricnae [here](https://github.com/syl20bnr/spacemacs/issues/15219#issuecomment-998902397), `ivy-occur-grep-mode` opens in motion state for him also (I don't know why, as mentioned already in my previous comment). Anyway, I have updated the commit with the simple fix to override the keymap for all states (which should be fine as we use only a singel state in the mode anyway).

dalanicolai:
> this probably could solve other key binding shadowing problems as well.

You mean the overriding keymap? Yes, I tried to see if I could use it for the `custom-mode` also, but the `custom-mode-map` would override the `p` keybinding in `normal-state`, but I prefer 'paste' over 'previous widget', so in #15223 I set the 'mouse click' keybinding (only) explicitly.

In the example [here](https://github.com/noctuid/evil-guide#evils-tools), noctuid uses it for Info-mode map, which might be a good idea also. (Although I prefer 'full evilification' like in `evil-collection` or in [my personal layer](https://github.com/dalanicolai/info-layer)).

lebensterben:
@czqhurricnae
I'll merge this if this PR fixes your issues.

lebensterben:
I'm merging this despite that no response has been heard.

PR #15228, opened 2021-12-28
[auto-completion][julia] Revert #15171, Proper fix for #15169

thanhvg:
This reverts commit f2755533deef68cb2fb2a2d40e6a5ab6ede13204.

This adds fix for julia layer breaking yasnippet setup.

Commit f2755533deef68cb2fb2a2d40e6a5ab6ede13204 fixed a specific bug in julia
layer but it loads yasnippet and friends on start up which adds 1 sec to my
startup time even though I don't use julia.

Julia layer loads lsp-julia early on startup, which loads lsp which then loads
yasnippet behind the scene unknown to spacemacs and use-package. That's why our
nice and ordered setup for yasnippet and yasnippet-snippets failed as discussed
in https://github.com/syl20bnr/spacemacs/pull/15171. We now tell use-package to
defer its load.

Note: I didn't do a full test with a julia lsp server. If someone who actually does julia with lsp can verity it that would be great.

### DIFF
--- a/layers/+completion/auto-completion/packages.el
+++ b/layers/+completion/auto-completion/packages.el
@@ -335,9 +335,7 @@
         (when auto-completion-private-snippets-directory
           (if (listp auto-completion-private-snippets-directory)
               (setq yas-snippet-dirs (append yas-snippet-dirs auto-completion-private-snippets-directory))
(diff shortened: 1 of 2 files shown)


smile13241324 (approved):
Sounds sensible, changing the behaviour of all layers to fix a single one is not a good solution, fixing the issue in the layer itself is much better.

Issue #15230, opened 2022-01-01
core-configuration-layer.el depend on deprecated `defmethod` from EIEIO


oxyum:
EIEIO was deprecated for years now. Most spacemacs codebase was migrated to `cl-defmethod`, but not `core-configuration-layer.el`.

But at the end of 2021 `eieio-compat.el` was finally killed from emacs repo:
https://github.com/emacs-mirror/emacs/commit/d65534d254b3965ea82a9300c12c5c07f88818b7

So spacemacs became unusable on newer emacs.

sunlin7:
And the dependency package `multi-line` in https://github.com/syl20bnr/spacemacs/blob/0206197b224f67e6a4c65bb850758f4affc9cc84/layers/%2Bspacemacs/spacemacs-editing/packages.el#L46  also require the `defmethod`, and sadly the package `multi-line` is in-actived https://github.com/IvanMalison/multi-line.

lebensterben:
the problem with `multi-line` is fixed in an upstream PR https://github.com/IvanMalison/multi-line/pull/16
I will take a look on core-configuration-layer.el

cormacc:
While waiting for that PR to get merged, I've made the following edit locally to spacemacs-editing/packages.el:

```
        ;; multi-line
        ;; Workaround pending merge of PR https://github.com/IvanMalison/multi-line/pull/16
        (multi-line :location (recipe
                            :fetcher github
                            :repo "eklitzke/multi-line"
                            :branch "master"))
```

oxyum:
fixed in #15238

PR #15238, opened 2022-01-03
Change core-configuration-layer.el to not use obsolete defmethod

eklitzke:
As reported in #15230, upstream has removed `eieio-compat.el`. This means that `defmethod` is removed upstream, which means that Spacemacs is broken when using the latest commit in the upstream Emacs master branch.

I'm not really an expert in `cl` or `eieio` but this seems to fix the issue for me (would love some better code review from someone more knowledgeable though). Changing `defmethod` to `cl-defmethod` in `core-configuration-layer.el` just worked for all lines except one, the specialization of `cfgl-layer-owned-packages` for a `nil` argument on line 217 had to be changed to `null`.

There are a few other people who left thumbs up on #15230 so if anyone else who builds Emacs from upstream master can test this change please let me know if it works for you.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -203,7 +203,7 @@ subdirectory of ROOT is used."
                 "Boolean to track whether layers.el has been loaded."))
   "A configuration layer.")
 
-(defmethod cfgl-layer-owned-packages ((layer cfgl-layer) &optional props)
+(cl-defmethod cfgl-layer-owned-packages ((layer cfgl-layer) &optional props)
   "Return the list of owned packages by LAYER.
 If PROPS is non-nil then return packages as lists with their properties.
 LAYER has to be installed for this method to work properly."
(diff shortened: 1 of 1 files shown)


eklitzke:
Note that if you're using an Emacs build with the upstream master branch you're probably also going to get some errors loading `multi-line` which is initialized by the `spacemacs-editing` package, the error will look like this:

```
Error (use-package): multi-line/:catch: Symbol’s value as variable is void: multi-line-candidate-position Disable showing Disable logging
```

I created a PR for that project https://github.com/IvanMalison/multi-line/pull/17 to fix the issue, but it looks like `multi-line` may be abandoned as there's another open PR from October 2021 fixing an emacs 28 issue that the upstream author hasn't commented on or merged.

freydor:
Hi I just tested this PR works perfect on my system thanks.
Before that I just did a quick `(require 'eieio-compat)`

bhipple:
PR also works perfectly on my system with the latest emacs build from `master`; thank you for the improvement!

smile13241324 (approved):
Looks good, thanks for the fix.

PR #15220, opened 2021-12-19
Update minimal supported emacs version to 27.1

lebensterben:
Emacs 27.1 has landed in Debian's stable release.

Since Debian stable is relatively stable/outdated, it's safe to change our
minimal supported emacs version to 27.1 now.

PR #15239, opened 2022-01-03
Remove Emacs 26.3 from github workflows

eklitzke:
Since #15220 Emacs 26 shouldn't be part of the github test workflows.


smile13241324:
Ups didn't see your PR and fixed it already, thanks for the effort anyway.

Issue #14201, opened 2020-12-10
Default tab-width of 2 causes weird indentation (e.g. in 'read-only' emacs-core .el files)


dalanicolai:
#### Description :octocat:
The default Spacemacs tab-width causes weird indentation in e.g. Emacs core files, see image
![image](https://user-images.githubusercontent.com/18429791/101728596-f55e7800-3ab6-11eb-97dc-c4d7de77f8b3.png)
In the image, left is with the default tab-width of 2 while right is with tab-width of 8. I have marked the `when` form. To me it seems like the two lines below the `when` are not part of the `when` form, while on the right side it shows clearly that they are.
So I wonder it this is desired behavior, or if the default tab-width should be kept at Emacs its default of 8.

Also `SPC h L` does not seem to work/be available anymore (on develop branch), although I did not test this on a clean installation yet. It would be great
if there was an option to start/test Spacemacs in its basic state. Similar to `emacs -Q`.

#### Reproduction guide :beetle:
- Start Spacemacs (with default tab-width of 2) and open a 'read-only' emacs core .el file (e.g. `SPC h d f calc-do o`).
- Repeat with a tab-wtidth of 8.

lebensterben:
Confirmed.
Some time during this week I also noticed this behaviour in `helpful-callable`.
(Interestingly enough, the helpful buffer allows me to re-indent the whole file even though it won't be saved.)

The reason why the indentation is weird is, these files contains '\t'.
With reformatting (in helpful buffer), all '\t' are replaced by '\ ', so it appears normal.

I've no idea how to easily fix this except that we can advise the `ace-link-hlep` with a `:after`, that it will create a temporary file with that content, but reformat the whole buffer.
Alternatively, when opening this buffer, we need set buffer-local tab-width to 8.

It would be a little bit easier if you use `helpful` instead, we just need to set `helpful-switch-buffer-function` (without using advising function feature).

lebensterben:
```elisp
(defun indent-help-mode ()
    (setq buffer-read-only nil)
    (indent-region (buffer-end -1) (buffer-end 1))
    (setq buffer-read-only t))
(advice-add 'ace-link-help :after #'indent-help-mode)
```
This is a proof of concept. The can automatically format the whole buffer.

dalanicolai:
Thanks for the helpful;) answer, although one thing is not clear to me.
[...]

PR #15243, opened 2022-01-05
help/helpful: Set tab-width to 8, better integration w/ Ivy

lebensterben:
Currently we set default tab-width to 2, but in Emacs's Elisp source code,
tab-width defaults to 8.

This discrepancy results in bad alignment of Elisp source code in help-mode
and helpful-mode (when helpful layer is enabled).

This commit fixes this problem by setting tab-width to 8 in those buffers.

In addition, when both helpful and ivy are in use, this commit improved
their integration according to helpful's documentation.

### DIFF
--- a/layers/+emacs/helpful/packages.el
+++ b/layers/+emacs/helpful/packages.el
@@ -52,7 +52,7 @@
     (defalias 'describe-variable 'helpful-variable)
     (defalias 'describe-key 'helpful-key)
     (add-hook 'helpful-mode (lambda () (setq-local tab-width 8)))
-    (when (featurep 'consel)
+    (when (featurep 'counsel)
       (setq counsel-describe-function-function #'helpful-callable)
       (setq counsel-describe-variable-function #'helpful-variable))))
 
--- a/layers/+emacs/helpful/packages.el
+++ b/layers/+emacs/helpful/packages.el
@@ -50,7 +50,11 @@
     (evil-define-key 'normal helpful-mode-map (kbd "q") 'quit-window)
     (defalias 'describe-function 'helpful-callable)
     (defalias 'describe-variable 'helpful-variable)
-    (defalias 'describe-key 'helpful-key)))
+    (defalias 'describe-key 'helpful-key)
+    (add-hook 'helpful-mode (lambda () (setq-local tab-width 8)))
+    (when (featurep 'consel)
+      (setq counsel-describe-function-function #'helpful-callable)
+      (setq counsel-describe-variable-function #'helpful-variable))))
 
 (defun helpful/post-init-link-hint ()
   (with-eval-after-load 'helpful
--- a/layers/+spacemacs/spacemacs-defaults/packages.el
+++ b/layers/+spacemacs/spacemacs-defaults/packages.el
@@ -232,7 +232,8 @@
 (defun spacemacs-defaults/init-help-fns+ ()
   (use-package help-fns+
     :commands (describe-keymap)
-    :init (spacemacs/set-leader-keys "hdK" 'describe-keymap)))
+    :init (spacemacs/set-leader-keys "hdK" 'describe-keymap)
+    :config (add-hook 'help-mode (lambda () (setq-local tab-width 8)))))
 
 (defun spacemacs-defaults/init-hi-lock ()
   (with-eval-after-load 'hi-lock


lebensterben:
This PR helps for https://github.com/syl20bnr/spacemacs/issues/14201 thought not solving it.

emacs18:
What about shell-mode where output of ls command is totally messed up?
What about diff mode?
There were many others, so for me I simply added the following in `dotspacemacs/user-config`:
```elisp
(setq-default tab-width 8)
```

lebensterben:
Can you share screenshots of ls's output and diff-mode?

emacs18:
I just examined ls output in my shell buffer.  The output is not impacted by tab-width setting, because ls output does not have any tabs!  Years ago when I first started using spacemacs, this was not the case.  Perhaps ls has been modified in the past few years not to use tabs. It could also be that there is some option which has been set to tell ls not to use tabs.

I think I will remove my custom code to  just use default tab width of 2 and see if I observe other problems.

layers/+emacs/helpful/packages.el
```diff
@@ -50,7 +50,11 @@
     (evil-define-key 'normal helpful-mode-map (kbd "q") 'quit-window)
     (defalias 'describe-function 'helpful-callable)
     (defalias 'describe-variable 'helpful-variable)
-    (defalias 'describe-key 'helpful-key)))
+    (defalias 'describe-key 'helpful-key)
+    (add-hook 'helpful-mode (lambda () (setq-local tab-width 8)))
+    (when (featurep 'consel)
```
tko, line 55:
```suggestion
    (when (featurep 'counsel)
```
?

lebensterben: thanks!

emacs18:
My comment on ls output in shell buffer before was on ubuntu 20.04.
On CentOS 7, ls output in shell buffer has lots of tab characters so that if tab-width is not 8, then it is too painful to deal with.

lebensterben:
@emacs18 
I'm not familiar with CentOS and I've no idea why there are TAB characters in ls outupt.
Can you show me a screenshot?

emacs18:
I don't have access to CentOS 7 right now, but it is easy to show this. Starting with my shell buffer on Ubuntu 20.4 containing the output of `ls /usr/lib/X11/config` command, I simply evaluated this:

```elisp
(let ()
  (setq tab-width 8)
  (tabify (point-min) (point-max))
  (setq tab-width 2)
  (untabify (point-min) (point-max)))
```

The result is shown next.


```shell
$ ls /usr/lib/X11/config
Amoeba.cf  hp.cf         necLib.tmpl   sony.cf
apollo.cf  hpLib.rules   NetBSD.cf   sun.cf
bsd.cf     hpLib.tmpl    noop.rules  sunLib.rules
bsdi.cf    ibm.cf        nto.cf    sunLib.tmpl
bsdiLib.rules  ibmLib.rules  nto.rules   sv3Lib.rules
bsdiLib.tmpl   ibmLib.tmpl   Oki.cf    sv3Lib.tmpl
bsdLib.rules   Imake.cf      oldlib.rules  sv4Lib.rules
bsdLib.tmpl  Imake.rules   OpenBSD.cf  sv4Lib.tmpl
cde.rules  Imake.tmpl    OpenBSDLib.rules  svr3.cf
cde.tmpl   isc.cf        OpenBSDLib.tmpl   svr4.cf
convex.cf  Library.tmpl  os2.cf    Threads.tmpl
cray.cf    linux.cf      os2Lib.rules  ultrix.cf
cross.def  lnxdoc.rules  os2Lib.tmpl   usl.cf
cross.rules  lnxdoc.tmpl   os2.rules   version.def
cygwin.cf  lnxLib.rules  osf1.cf     Win32.cf
cygwin.rules   lnxLib.tmpl   osfLib.rules  Win32.rules
cygwin.tmpl  luna.cf       osfLib.tmpl   WinLib.tmpl
darwin.cf  lynx.cf       pegasus.cf  X11.rules
darwinLib.rules  mach.cf       QNX4.cf     X11.tmpl
darwinLib.tmpl   macII.cf      QNX4.rules  x386.cf
date.def   mingw.cf      sco5.cf     xf86.rules
DGUX.cf    mingw.rules   sco.cf    xf86site.def
dmx.cf     mingw.tmpl    scoLib.rules  xf86.tmpl
DragonFly.cf   minix.cf      sequent.cf  xfree86.cf
FreeBSD.cf   Mips.cf       sequentLib.rules  xorg.cf
fujitsu.cf   Motif.rules   ServerLib.tmpl  xorgsite.def
generic.cf   Motif.tmpl    Server.tmpl   xorg.tmpl
gnu.cf     moto.cf       sgi.cf    xorgversion.def
gnuLib.rules   ncr.cf        sgiLib.rules
gnuLib.tmpl  nec.cf        sgiLib.tmpl
host.def   necLib.rules  site.def
$ 
```

emacs18:
If it was just shell buffer, I probably  could set `tab-width` via major mode hook.  However I got so annoyed having to do that in multiple places that I set `tab-width` to 8 globally.  This was within my first days using spacemacs years ago.

lebensterben:
@emacs18 
I do see \t in shell-mode buffer now. I normally use vterm and there's not \t.

lebensterben:
@emacs18 
You may open a new issue on the major-modes where you think tab-width should be 8.

emacs18:
In my opinion `tab-width` should have been left at 8 all along, because that has been assumed in so many things.
So I'm not going to spend any time on this myself.  I may set `tab-width` in one or two major mode hooks.  However if I have to do this more than that, I may just go back to over-riding `tab-width` and set it to 8 for all buffers.

lebensterben:
@emacs18 
I kinda agree.
We probably should not change it by default. And only set it on a per-major-mode basis.

PR #15247, opened 2022-01-06
Add C-o and C-i keybindings to evil-evilified-state-map (and remove from pdf-view-mode-map) 

dalanicolai:
See motivation in commit messages. I for clarity I have kept the two commits separate. If you agree with these improvements, and you would like me to squash the commits, then just let me know.

So I think `C-o` and `C-i` should consistently be used for jumping 'global' history (or at least using evil-jump functions, I don't think that means 'global' per se) and any other keybindings for jumping 'local' history.

I forgot to provide an example. But just open some pdf in pdf-view, then go to a separate buffer and start jumping to previous locations using `C-o`. I guess you will agree that at least that behavior is quite annoying (I have added those bindings myself, because it was better then the bindings before that, but now I think these are even more logical bindings. In pdf-tools, jumping history is already available under `[` and `]`.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -814,6 +814,8 @@ Other:
     inherit leader key bindings from another mode
   - Open buffers with major-mode derived from =special-mode= in motion-state by
     default (thanks to Daniel Nicolai)
+  - Add evil-jump-backward/forward (=C-o=/=C-i=) keybindings to
+    evil-evilified-state-map (thanks to Daniel Nicolai)
 *** Distribution changes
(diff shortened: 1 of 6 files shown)


dalanicolai:
Oops... forgot to create a branch for the PR. I hope that is alright, otherwise just let me know, and I will create a branch for this PR.

dalanicolai:
I would like to add the comment about the history being available already under `[` and `[` to the commit message, but I am not sure if it is wise to force push on the `develop` branch. If it is okay then I could do it, but I guess that comment is not very important anyway (just would have been a slightly nicer commit message).

lebensterben:
there will only be one commit when "squash and merge"

lebensterben (approved):

Issue #15106, opened 2021-10-18
Glitches with pdf-tools


DrWaleedAYousef:
#### Description :octocat:
1. `pdf-tools` mode does not start when opening a pdf, unles there is this line in `.spacemacs`:
```
(add-hook 'pdf-view-mode-hook 'pdf-tools-enable-minor-modes)
```
I traced the problem, it only happend when the `org` layer is loaded.

2. When scrolling a pdf file, the following buffer always appears (also, attached as an image for clarity)

```
These default coding systems were tried to encode text
in the buffer ‘ *temp file*’:
  (utf-8 (11 . 4194274) (12 . 4194275) (13 . 4194255) (14 . 4194259) (468 .
  4194204) (469 . 4194285) (471 . 4194267) (473 . 4194267) (477 . 4194295) (480
  . 4194280) (481 . 4194199))
However, each of them encountered characters it couldn’t encode:
  utf-8 cannot encode these: � � � � � � � � � � ...

Click on a character (or switch to this window by ‘M-m w w’
and select the characters by RET) to jump to the place it appears,
where ‘M-m u C-x =’ will give information about it.

Select one of the safe coding systems listed below,
or cancel the writing with C-g and edit the buffer
   to remove or modify the problematic characters,
or specify any other coding system (and risk losing
   the problematic characters).

  raw-text no-conversion
```

#### Reproduction guide :beetle:
- Start Emacs
- open any PDF.

*Observed behaviour:* :eyes: :broken_heart:
explained upfront

*Expected behaviour:* :heart: :smile:
<<DESCRIPTION OF THE EXPECTED BEHAVIOUR>>

#### System Info :computer:
- OS: gnu/linux
- Emacs: 27.2
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 63056ecb5)
- Graphic display: t
- Distribution: spacemacs
- Editing style: emacs
- Completion: helm
- Layers:
```elisp
(dap c-c++ php systemd
     (latex :variables latex-build-command "LatexMk" latex-enable-auto-fill t latex-backend 'lsp latex-build-engine 'luatex latex-view-pdf-in-split-window t latex-enable-magic t latex-refresh-preview t latex-enable-folding t)
     (bibtex :variables bibtex-enable-ebib-support t ebib-preload-bib-files
[...]

PR #15258, opened 2022-01-10
fix evil-lisp-state prefixes & improve related functions

lebensterben:
- make sure `evil-lisp-state` related prefix keys are displayed correctly.
- `which-key-add-keymap-based-replacements` accepts multiple key-def pairs
  thus this commit allow `spacemacs/declare-prefix` to accept multiple inputs
  too.

### DIFF
--- a/core/core-keybindings.el
+++ b/core/core-keybindings.el
@@ -54,12 +54,12 @@ gui, translate to [C-i]. Otherwise, [9] (TAB)."
 ;;     [C-m] [?\C-m]))
 ;; (define-key key-translation-map [?\C-m] 'spacemacs/translate-C-m)
 
-(defun spacemacs/declare-prefix (prefix name &optional _)
+(defun spacemacs/declare-prefix (prefix name &rest more)
   "Declare a prefix PREFIX. PREFIX is a string describing a key
 sequence. NAME is a string used as the prefix command."
-  (which-key-add-keymap-based-replacements spacemacs-default-map
-    prefix name))
-(put 'spacemacs/declare-prefix 'lisp-indent-function 'defun)
+  (declare (indent defun))
+  (apply #'which-key-add-keymap-based-replacements spacemacs-default-map
+    prefix name more))
 
 (defun spacemacs/declare-prefix-for-mode (mode prefix name &optional _)
   "Declare a prefix PREFIX. MODE is the mode in which this prefix command should
--- a/layers/+spacemacs/spacemacs-defaults/keybindings.el
+++ b/layers/+spacemacs/spacemacs-defaults/keybindings.el
@@ -26,81 +26,80 @@
 ;; ---------------------------------------------------------------------------
 
 ;; We define prefix commands only for the sake of which-key
-(setq spacemacs/key-binding-prefixes `((,dotspacemacs-emacs-command-key "M-x")
-                                       ("!"   "shell cmd")
-                                       ("*"   "search project w/input")
-                                       ("/"   "search project")
-                                       ("?"   "show keybindings")
-                                       ("a"   "applications")
-                                       ("ac"   "chat")
-                                       ("ae"   "email")
-                                       ("af"   "fun")
-                                       ("ar"   "readers")
-                                       ("am"   "music")
-                                       ("at"  "tools")
-                                       ("ats"  "shells")
-                                       ("aw"  "web-services")
-                                       ("c"   "compile/comments")
-                                       ("C"   "capture/colors")
-                                       ("e"   "errors")
-                                       ("g"   "git/versions-control")
-                                       ("h"   "help")
-                                       ("hd"  "help-describe")
-                                       ("hP"  "profiler")
-                                       ("hT"  "tutorials")
-                                       ("i"   "insertion")
-                                       ("j"   "jump/join/split")
-                                       ("jj"  "avy timer")
-                                       ("jl"  "avy line")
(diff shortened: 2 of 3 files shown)


lebensterben:
@DrWaleedAYousef
Please pull this PR and restart emacs to see whether the prefix keys now correctly shows the descriptions instead of just "+prefix".

DrWaleedAYousef:
I think it is the same. The `SPC-k` shows the `+lisp` menus.

lebensterben:
@DrWaleedAYousef It should be the lisp menu, isn't it?

DrWaleedAYousef:
> @DrWaleedAYousef It should be the lisp menu, isn't it?
> 
> To clarify, SPC k is the prefix for both smartparens and evil-lisp-state.

Yes, I know what you mean. I am confirmgin that it gives only the +lisp state. And the smartparens (the +prefix) does not exist.

DrWaleedAYousef:
I have a hypothesis that I am testing now: I think the reason of this bug and of this #15106 as well is the `org` layer. I just started disabling it and every thing works fine now. I will confirm with you after some days of keeping it dissabled.

lebensterben:
@DrWaleedAYousef
Please see the screenshot below.

You would see the smartparen bindings (those started with sp-) and the evil-lisp-state bindings (under prefix keys `, d, and D, near the end).

Screenshot_2022-01-10-50_1687x306

DrWaleedAYousef:
Ok I will retry

DrWaleedAYousef:
I tried several times to retart emacs and do `SPC-k`. It is very undeterministic! sometimes it gives the image that you shared above, and sometimes it gives only the three prefix keys!

lebensterben:
@DrWaleedAYousef 
> and sometimes it gives only the three prefix keys

If you haven't opened any buffer, you would only see these three prefix. And this is not a bug.

Is that the case?

DrWaleedAYousef:
> @DrWaleedAYousef
> 
> > and sometimes it gives only the three prefix keys
> 
> If you haven't opened any buffer, you would only see these three prefix. And this is not a bug.
> 
> Is that the case?

No; I open a buffer and simply do `SPC q r` to restart on the same buffer. However, I just noticed something. The correct menu only activates when I open a lisp file (`.el`). Then, it works correctly with any other file (e.g. `.tex`). However, if start with opening any other file it does now show up, unless I open a list file.

Also, aside issue is that at the start up this message appear: `Error (use-package): gnus/:init: replacement is neither a cons cell or a string`

lebensterben:
@DrWaleedAYousef 
A few findings.

Previously I said `SPC k` is for `evil-lisp-state` and `smartparens` because I saw some bindings like `SPC k $` is bound to `sp-end-of-sexp`. This is technically correct but the twist is that those `sp-*` commands are actually `evil-lisp-state-sp-*` commands. That is, they're registered when `evil-lisp-state` configured the keymap and they are just wrapper functions calling the `sp-*` commands from `smartparens` package.

`evil-lisp-state` is *lazily loaded*. It's *fully loaded* only after you opened a buffer that derives from `prog-mode`.
- The three prefix keys are always there because they're registered via `which-key`, regardless of whether `evil-lisp-state` is loaded or not.
- If you just opened Emacs, or you opened the message buffer, scratch buffer, a `plaintext` file like .txt or .org, since none of them is `prog-mode`, `evil-lisp-state` is not loaded. And you would not see anything other than three prefixes.
- If you've opened a buffer for any programming language, it's `prog-mode`. And `evil-lisp-state would be fully loaded. And you should see the full list of commands as I showed in the screenshot.

---

> However, if start with opening any other file it does now show up, unless I open a lisp file.

Please describe what are 'any other file' you've opened.

DrWaleedAYousef:
Thanks so much for these explanations. I just made a silly typo above; I said:
> However, if start with opening any other file it does now show up, unless I open a lisp file.
I just meant:
However, if start with opening any other file it does **not** show up, unless I open a lisp file.

So, when opening any other file, e.g., `.tex`, before I open a `.el` file, the `sp-*` does not show up.

lebensterben:
@DrWaleedAYousef 
1. Open an .tex file. (Alternatively open an empty buffer and start `Tex-tex-mode` manually.
2. Evaluate `(derived-mode-p 'prog-mode)` and you will see `nil`. (It derives from text-mode instead)
3. As I said those bindings are loaded for `prog-mode`. And `.tex` file is not. That's why you don't see them.

If you're convinced and if you can confirm there's no other unexpected behaviour, then I'll merge this PR.

smile13241324 (approved):

DrWaleedAYousef:
> 2\. (derived-mode-p 'prog-mode)

Ok; thanks for your patience. I am convinced. Just one remaining question: how then I can have the `sp-*` functionalities for the tex files natively? From wht you explained, the only way is to open a dummy file for any programing, .e.g., `.el`, then close it. Is there any natural way like adding something to the `.spacemacs`?

lebensterben:
@DrWaleedAYousef 
I will merge this now but keep your original issue open.

DrWaleedAYousef:
@lebensterben Just to confirm (to keep you posted) I am able to use all the keybindings of `sp-parens` *only* after opening, e.g., a lisp file and close it again.

lebensterben:
@DrWaleedAYousef 

You should update in your original issue, not in this closed PR.

DrWaleedAYousef:
Oh sorry; I thought it is the original post. I was rushing

PR #15284, opened 2022-01-22
[org] integrate org-appear with evil when triggering manually

Kazark:
There is a little integration code that is necessary for using Evil with the latest feature of `org-appear`: manual triggering. This is exactly the kind of code that Spacemacs layer exist for: to integrate various packages. For more information, see the original [feature request](https://github.com/awth13/org-appear/issues/37) on `org-appear`.

I don't know much about the hybrid editing style so please correct me if I should only have applied this to the Vim editing style.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2889,8 +2889,11 @@ files (thanks to Daniel Nicolai)
 - Fixed ocamlformat not properly initialized by following do (thanks to Nicolas Raymond)
 - Allow new OPAM share commandline syntax (thanks to Cody McKenzie)
 **** Org
+- When =org-appear= is set to be triggered manually and Spacemacs' editing mode is
+  Vim or hybrid, register relevant =org-appear= functions in Evil hooks for Org
+  buffers (thanks to Keith Pinson).
 - Provide some sane default strategies for enforcing =TODO= dependencies via
-  =org-todo-dependencies-strategy=. (thanks to Keith Pinson)
+  =org-todo-dependencies-strategy= (thanks to Keith Pinson).
 - Load org-mode email integration (mu4e/notmuch) when org is loaded
   (thanks to Steven Allen)
 - Packages:
--- a/layers/+emacs/org/README.org
+++ b/layers/+emacs/org/README.org
@@ -548,6 +548,9 @@ To install [[https://github.com/awth13/org-appear][org-appear]], which toggles v
    '((org :variables org-enable-appear-support t)))
 #+END_SRC
 
+If you set =org-appear-trigger= to =manual= and your editing style is modal,
+=org-appear= will be on in insert mode but off in normal mode.
+
 ** Verb support
 To install [[https://github.com/federicotdn/verb][Verb]], an HTTP client based on Org mode, set the
 =org-enable-verb-support= variable to =t=:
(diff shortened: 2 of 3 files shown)


layers/+emacs/org/README.org
```diff
@@ -548,6 +548,9 @@ To install [[https://github.com/awth13/org-appear][org-appear]], which toggles v
    '((org :variables org-enable-appear-support t)))
 #+END_SRC
 
+If you set =org-appear-trigger= to =manual= and your editing style is modal,
```
lebensterben, line 551: modal? do you mean vim or hybrid?

layers/+emacs/org/packages.el
```diff
@@ -1025,6 +1025,12 @@ Headline^^            Visit entry^^               Filter^^                    Da
     :init
     (progn
       (add-hook 'org-mode-hook 'org-appear-mode)
+      (when (and (eq org-appear-trigger 'manual)
+                 (memq dotspacemacs-editing-style '(vim hybrid)))
+        (add-hook 'org-mode-hook
+                  (lambda ()
+                    (add-hook 'evil-insert-state-entry-hook #'org-appear-manual-start nil t)
```
lebensterben, line 1032: By directly modifying evil's hooks, would this also affect other buffer that's not in org-mode?

Those hooks should be made buffer-local. (via the fourth argument of add-hook)

smile13241324 (changes requested):
Please check the points from @lebensterben. I would like to be sure this will not affect non org buffers.

layers/+emacs/org/README.org
Kazark: Yes. Would it be better to phrase it that way?

layers/+emacs/org/packages.el
Kazark: I must be missing something here. I thought the `t` I was passing as the final (fourth?) argument of `add-hook` did just that.

layers/+emacs/org/README.org
lebensterben: yes, please change it.

Kazark: Will do when I get a sec.

CHANGELOG.develop
```diff
@@ -2889,8 +2889,11 @@ files (thanks to Daniel Nicolai)
 - Fixed ocamlformat not properly initialized by following do (thanks to Nicolas Raymond)
 - Allow new OPAM share commandline syntax (thanks to Cody McKenzie)
 **** Org
+- When =org-appear= is set to be triggered manually and Spacemacs' editing mode is
+  Vim or hybrid, register relevant =org-appear= functions in Evil hooks for Org
```
lebensterben, line 2893:
```suggestion
  Vim or hybrid, register relevant =org-appear= commands in Evil hooks for Org
```

layers/+emacs/org/README.org
lebensterben:
```suggestion
If you set =org-appear-trigger= to =manual= and your editing style is =vim= or =hybrid=,
```

layers/+emacs/org/README.org
```diff
@@ -548,6 +548,9 @@ To install [[https://github.com/awth13/org-appear][org-appear]], which toggles v
...
+=org-appear= will be on in insert mode but off in normal mode.
```
lebensterben, line 552:
```suggestion
=org-appear= is turned on in insert mode but not in normal mode.
```

lebensterben (changes requested):

lebensterben (approved):

lebensterben:

PR #15289, opened 2022-01-26
org: add binding for org-paste-subtree

dankessler:
There are already bindings for org-cut-subtree and org-copy-subtree; add a
binding for org-paste-subtree so that the user can make use of their freshly cut
or copied subtrees (and automatically paste them in a level-adjusted manner)


lebensterben (approved):

lebensterben:
Thanks for your contributions.

PR #15297, opened 2022-01-29
spacemacs-defaults: Added prompt to save a buffer

### LINKED ISSUE #12857: Deleting a new buffer with changes does not ask for confirmation
Given that software usually asks when one is about to throw away unsaved work I find this behavior unexpected and it should be changed.

lebensterben:
Added a new `kill-buffer-hook` that prompt the user to save a buffer
before killing it.

closes #12857


lebensterben:
@gutsle @henriquenj
Please pull in this PR and help to test whether this works as you both requested.

lbutkows:
This change makes difficult to use with treemacs and its collapse feature. Every time the folder is selected, the confirmation to kill the buffer will appear several times.
Buffer "Pfuture-Buffer (/usr/bin/python3 -O ... 3 t)" modified; kill anyway? (y or n) y

PR #15313, opened 2022-02-05
[compleseus][ivy] quote regex when search for region or symbol at point

thanhvg:
things like `C++` will be quoted as `C\+\+` and search will produce desired result


lebensterben (approved):

lebensterben:
Thank!

PR #15315, opened 2022-02-05
[elisp] add elisp-def

thanhvg:
this package provide highly accurate jump to the definition, it even works with
local variable

### DIFF
--- a/layers/+lang/emacs-lisp/packages.el
+++ b/layers/+lang/emacs-lisp/packages.el
@@ -28,6 +28,7 @@
     (debug :location built-in)
     (edebug :location built-in)
     eldoc
+    elisp-def
     elisp-slime-nav
     (emacs-lisp :location built-in)
     evil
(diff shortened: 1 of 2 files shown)


layers/+lang/emacs-lisp/packages.el
```diff
@@ -160,6 +161,10 @@
       (spacemacs/set-leader-keys-for-major-mode 'emacs-lisp-mode
         "cl" 'auto-compile-display-log))))
 
+(defun emacs-lisp/init-elisp-def ()
+  (use-package elisp-def
+    :defer (spacemacs/defer)))
```
lebensterben, line 166: normally we use :defer t
why do you use (spacemacs/defer) here?

thanhvg: I don't know somehow other packages in the elisp layers follow this convention. Its goal I assume is to work with dump feature which I never try.

lebensterben, line 166:
```suggestion
    :defer t))
```

lebensterben:
Please update CHANGELOG.develop and I will merge this.

lebensterben:

PR #15327, opened 2022-02-08
Fix consult-buffer error

ismama:
consult--source-file has been renamed/deprecated. It is now called
consult--source-recent-file.

See https://github.com/minad/consult/issues/508.


lebensterben:
Thanks.

PR #15329, opened 2022-02-09
Fix ⌘-s shortcut in macOS

### LINKED ISSUE #14166: cannot use ⌘-s  to save under macOS with osx layer
#### Description :octocat:
cannot save under macOS with osx layer

I found there is code which seems bind  ⌘-s to "\C-x\C-s" to do save, but it doesn't work
```emacs-lisp 
(global-set-key (kbd-mac-command "s")
                  (lambda ()
                    (interactive)
                    (call-interactively (key-binding "\C-x\C-s"))))
```

- Start Emacs
- open a file
- type some thing
- ⌘-s to save

cannot save, and got a message:
```
Wrong type argument: commandp, nil
```

save without error

#### System Info :computer:
- OS: darwin
- Emacs: 27.1
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 184bb2b00)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:

bebound:
Fix #14166

### DIFF
--- a/layers/+os/osx/keybindings.el
+++ b/layers/+os/osx/keybindings.el
@@ -94,10 +94,7 @@ default."
   (global-set-key (kbd-mac-command "n") 'make-frame)
   (global-set-key (kbd-mac-command "`") 'other-frame)
   (global-set-key (kbd-mac-command "z") 'undo-tree-undo)
-  (global-set-key (kbd-mac-command "s")
-                  (lambda ()
-                    (interactive)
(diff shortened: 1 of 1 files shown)


lebensterben:
My feeling on this:

https://github.com/syl20bnr/spacemacs/issues/14166#issuecomment-983645593

PR #15331, opened 2022-02-09
spacemacs-defaults: Prompt to save a buffer for certain operations

lebensterben:
- Added a new function `spacemacs//confirm-kill-buffer` which prompts the user
to save a buffer to a file before killing it.
- Added the said new function to the buffer-local `kill-buffer-hook` of scratch
buffer and any buffer created via command `spacemacs/new-empty-buffer`.

closes https://github.com/syl20bnr/spacemacs/issues/12857

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -285,7 +285,6 @@ Dedicated (locked) windows are left untouched."
   (interactive "p")
   (spacemacs/rotate-windows-forward (* -1 count)))
 
-
 (if (configuration-layer/package-used-p 'winum)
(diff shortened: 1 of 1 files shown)


lebensterben:
@gutsle @henriquenj
Please pull in this PR and help to test whether this works as you both requested.

---

@lbutkows

Hi Lukasz, this is my second attempt on this issue. This should not cause the same bug as mentioned in https://github.com/syl20bnr/spacemacs/pull/15297#issuecomment-1033571548

lbutkows:
@lebensterben It looks that now it works well. Thanks.

henriquenj:
Thanks for the implementation @lebensterben. It works as I expect it.

PR #15332, opened 2022-02-10
spacemacs-buffer: Added optional feature to show file icons

lebensterben:
- Added a new .spacmeacs config variable
`dotspacemacs-startup-buffer-show-icons`. When non-nil, it will try to display a
file icon precedes each heading and entry in the startup list.
- Adjusted the order of the sections on the startup list. Startup loading time
is moved to the end.
- Removed the `heart.png` file since it's replaced by heart emoji or "heart"
depending on the said configuration variable.
- Also moved `all-the-icons` package declaration from `auto-completion` layer to `spacemacs-visual` layer.
  - This change is due to the fact that `all-the-icons` is used in
    `core/spacemacs-buffer.el`, which doesn't belong to any layer.
  - On a new installation of Spacemacs, `all-the-icons` is not installed yet but
    also note that `spacemacs-buffer-show-icons` defaults to nil, thus no error
    will occur.
  - Once the user modified the said variable to t, `all-the-icons` would be
    installed.

Screenshot_2022-02-10-22_988x1368


Kazark:
I will enable this immediately in my `.spacemacs`. 💯

lebensterben:
There's one more issue to solve. This new feature requires `all-the-icons` but it's not included as a `core/lib` package yet.

Kazark:
This is not working for me on Mac, and when I turn it off, the buffer doesn't look 100% like it used to. (This is probably okay; I just didn't know if it was intentional. For one thing, there seems to be more space between the numbers and the files now.)

Is there a font I need to manually install on my system to get this to work? I will try on my Manjaro box too when I get a chance.

Kazark:
It is working on my Linux box. Maybe this is a case of Emacs being impoverished on Mac, which I know is constant struggle. Only one surprise on Linux: the feature turns on automatically before you've updated your `.spacemacs` from the template---I would have assumed it would have defaulted to `nil`. Doesn't much matter since you are supposed to `SPC f e D` on update and then you can control what it does after that (and I wanted it on anyway), but I just didn't know if that was expected behavior.

lebensterben:
@Kazark 

You need to install all-the-icons fonts. Just M-x (all-the-icons-install-fonts)

lebensterben:
@Kazark 
I've set `dotspacemacs-startup-buffer-show-icons` to nil and restarted Emacs, and there's no icons.

Note that if you have multiple Emacs instances, the home buffer may not be killed even after you reload Emacs. In that case, the file icons will still be there.

Kazark:
Yeah, when it's set to `nil` it turns off. Before I had diffed my `.spacemacs` with the template, it wasn't in there at all---and that's when it was on. In other words, I:

* Pulled the latest changes
* `SPC f e U`
* `SPC q r`
Then it was on.
* `SPC f e D`
Then I copied it over, after which with `nil` it was off and `t` it was on---perfect.

May not be worth looking into further.

arifer612:
> I will investigate why it's turned on by default.

I just rebased my Spacemacs and it's set to 't by default. A very weird effect.

Kazark:
No rush.

PR #15345, opened 2022-02-13
Fix tab key functionality in pdf outline mode

dalanicolai:
For some reason the current 'syntax' for setting the TAB key functionality in pdf outline work does not work.
The small syntax change in this PR fixes the issue.

### DIFF
--- a/layers/+readers/pdf/packages.el
+++ b/layers/+readers/pdf/packages.el
@@ -108,7 +108,7 @@
         "gh"               'pdf-outline-up-heading
         "gg"               'beginning-of-buffer
         "G"                'pdf-outline-end-of-buffer
-        "TAB"              'outline-toggle-children
+        (kbd "<tab>")      'outline-toggle-children
         "RET"              'pdf-outline-follow-link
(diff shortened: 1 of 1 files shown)


layers/+readers/pdf/packages.el
```diff
@@ -108,7 +108,7 @@
         "gh"               'pdf-outline-up-heading
         "gg"               'beginning-of-buffer
         "G"                'pdf-outline-end-of-buffer
-        "TAB"              'outline-toggle-children
+        (kbd "<tab>")      'outline-toggle-children
```
lebensterben, line 111: you can bind it to both (kbd "TAB") and (kbd "<tab>")

See https://www.emacswiki.org/emacs/TabKey

dalanicolai: Ah that's interesting, there is more information also [here](https://emacs.stackexchange.com/questions/17509/how-to-distinguish-c-i-from-tab).

Anyway, as pdf-tools can not be used in terminals, it makes not sense to bind `C-i` to `outline-toggle-children` so it seems to me that the current PR is 'correct',

lebensterben:

Issue #14388, opened 2021-02-17
Add warning to `evilified-state-evilify` docstring that it is deprecated


dalanicolai:
I was trying to evilify the keymap for the [ebib](http://joostkremers.github.io/ebib/) package. To remind myself how to evilify a map, I peeked in some other layer (don't remember which one). I found that it used the `evilified-state-evilify` macro.

However, when I used this macro, I noticed that it does not evilify correctly, e.g it does not move `j` -> `J` and `J` to `C-j` etc.

Now in the source code I find a comment that this is an old macro, while just below there is the new macro `evilified-state-evilify-map`, which does evilify correctly.

As it is probably discouraged to use the old macro, I guess it would be better to add that to the docstring.
I would be happy to do it (just let me know)

lebensterben:
@dalanicolai 

If you're still interested in it, you can go ahead and add it to docstring.

Alternatively, you can make it as obsolete. See https://www.gnu.org/software/emacs/manual/html_node/elisp/Obsolete-Functions.html

PR #15354, opened 2022-02-17
Add obsoletion notification evilified-state-evilify

dalanicolai:
This PR simply adds the obsoletion notification as proposed in #14388.

### DIFF
--- a/layers/+readers/pdf/packages.el
+++ b/layers/+readers/pdf/packages.el
@@ -66,7 +66,9 @@
       (evil-define-key 'visual pdf-view-mode-map "y" 'pdf-view-kill-ring-save)
 
       ;; TODO: Make `/', `?' and `n' work like in Evil
-      (evilified-state-evilify pdf-view-mode pdf-view-mode-map
+      (evilified-state-evilify-map pdf-view-mode-map
+        :mode  pdf-view-mode
+        :bindings
         ;; Navigation
         "0"  'image-bol
         "$"  'image-eol
@@ -96,7 +98,9 @@
         "o"   'pdf-links-action-perform
         "O"   'pdf-outline
         "zr"  'pdf-view-scale-reset)
-      (evilified-state-evilify pdf-outline-buffer-mode pdf-outline-buffer-mode-map
+      (evilified-state-evilify-map pdf-outline-buffer-mode-map
+        :mode  pdf-outline-buffer-mode
+        :bindings
         "-"                'negative-argument
         "j"                'next-line
         "k"                'previous-line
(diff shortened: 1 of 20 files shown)


lebensterben:
There are a few invocations of this macro in our codebase.

See https://github.com/syl20bnr/spacemacs/search?q=%22evilified-state-evilify+%22&type=

dalanicolai:
Generally, that does not matter for obsoletion, the notification is only to inform users to use the new macro in new code.
However, replacing the old macro invocations is not so much work, so I will quickly do that. I guess the old macro invocation could be removed completely then (if users use them in private packages, it is only 10 secs work to replace them).

layers/+readers/pdf/packages.el
```diff
@@ -135,9 +154,9 @@
       (spacemacs/set-leader-keys-for-major-mode 'pdf-occur-buffer-mode
         "tf" 'next-error-follow-minor-mode))))
 
-(defun pdf/init-pdf-view-restore ()
-  (use-package pdf-view-restore
-    :after pdf-tools
-    :defer t
-    :init
-    (add-hook 'pdf-view-mode-hook 'pdf-view-restore-mode)))
+;; (defun pdf/init-pdf-view-restore ()
```
lebensterben, line 157: why is this commented out?

dalanicolai: Ah... that is an accident. I use this locally, while I am working on implementing/improving continuous scroll in pdf-tools...

layers/+emacs/helpful/packages.el
```diff
@@ -36,9 +36,9 @@
       (add-hook 'emacs-startup-hook
                 (lambda ()
                   (spacemacs/set-leader-keys
-                    "hdk" #'helpful-key
-                    "hdf" #'helpful-callable
-                    "hdv" #'helpful-variable)
+                    "hhk" #'helpful-key
```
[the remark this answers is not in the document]
dalanicolai: Ah, I thought I had skipped my personal settings. I prefer helpful not to overwrite the standard functionality (but I use this also locally, although I wanted to suggest that as a better keybinding in the helpful layer anyway).

Anyway, I think that while staged selectively, I somehow accidentally staged all when extending the commit (pressing `ce`). I will fix it. As far as I know, I only had personal changes in the helpful and pdf layers...

lebensterben (approved):

dalanicolai:
Ah, I guess you did that already... thanks!

lebensterben:
@dalanicolai 

>  Additionally, I have removed some redundant progns. AFAIK they are not required within the use-package form (you can check from expanding the use-package macro twice).

They are redundant. But we require this for unknown reason...
See https://develop.spacemacs.org/doc/CONVENTIONS.html#use-package

dalanicolai:
I have again pushed another update. I encountered some more 'unknown variable: ...keymap` warnings (in the vc layer), so I have wrapped the relevant `evilified-state-evilify-map` forms within appropriate `with-eval-after-load` forms

smile13241324:
@dalanicolai I have noticed there are some merge conflicts, care to have look?

dalanicolai:
@smile Sure! Took a look, and there was conflict from some newer commit. Anyway, should be 'clean' now...

PR #15361, opened 2022-02-19
unicode-fonts: ligature.el requires Emacs 28 +

stormwatch:
From [ligatures.el](https://github.com/mickeynp/ligature.el#compatibility-and-version-requirements) README:

> You must use Emacs 28 or later, or backport a fix to Emacs 27.x;
> 
> You can check by typing M-x emacs-version.
> 
> NOTE: There are critical issues in Emacs 27.1 and 27.2. Ideally, if at all possible, you should attempt to use a build of Emacs that includes this fix. See below for details.
>

### DIFF
--- a/layers/+fonts/unicode-fonts/README.org
+++ b/layers/+fonts/unicode-fonts/README.org
@@ -17,7 +17,7 @@ install the fonts listed in the [[https://github.com/rolandwalker/unicode-fonts#
 - Easily override glyphs or sections of glyphs.
 - Display color emoji on both the macOS port version of Emacs and emacs-plus (with
   =unicode-fonts-force-multi-color-on-mac= set to non nil).
-- Enable support for font ligature in Emacs 27 + via [[https://github.com/mickeynp/ligature.el][ligatures.el]].
+- Enable support for font ligature in Emacs 28 + via [[https://github.com/mickeynp/ligature.el][ligatures.el]].
 
 * Install
 To use this configuration layer, add it to your =~/.spacemacs=. You will need to


lebensterben:
That's not precise.
Emacs 27.x supports ligatures. The  issue is it's not stable.
Emacs 27 master branch contains a fix, which is also mentioned in the repository.

stormwatch:
[ligature.el will check the emacs version and stop the show if it is not running under emacs 28.](See https://github.com/mickeynp/ligature.el/blob/9357156a917a021a87b33ee391567a5d8e44794a/ligature.el#L273)

What changes would you recommend instead of this PR?

stormwatch:
If I enable ligatures by adding `unicode-fonts-enable-ligatures t` to the `unicode-fonts` layer

I get this error during startup:
> Warning (initialization): An error occurred while loading ‘/home/eze/.config/emacs/init.el’:
> 
> Symbol's function definition is void: ligature-set-ligatures
> 
> To ensure normal operation, you should investigate and remove the
> cause of the error in your initialization file.  Start Emacs with
> the ‘--debug-init’ option to view a complete error backtrace.

and if I use --debug-init it shows:
> Debugger entered--Lisp error: (error "Package ‘emacs-28’ is unavailable")
>   signal(error ("Package ‘emacs-28’ is unavailable"))
>   error("Package `%s-%s' is unavailable" emacs "28")
>   package-compute-transaction(nil ((emacs (28))))
>   package-install-from-buffer()
>   package-install-file("/home/eze/.emacs-profiles.d/spacemacs/.cache/quelp...")

Should I report this as an issue instead?

lebensterben:
Ligatures works well until this commit https://github.com/mickeynp/ligature.el/commit/00a7a39c80aa3690a0f7f598e341665e587d4391

Indeed there are some issues on Emacs 27, but the way the upstream handles it isn't optimal.

As for now, you should not enable this package in your configuration.

PR #15363, opened 2022-02-19
Make startup banner scalable (from dotfile)

dalanicolai:
The default logo is rather big. Users might prefer to use the screen size for
other information (while still showing a banner).

This PR adds a configuration variable `dotspacemacs-startup-banner-scale`
to the dotfile

### DIFF
--- a/core/core-spacemacs-buffer.el
+++ b/core/core-spacemacs-buffer.el
@@ -27,6 +27,7 @@
   (defvar dotspacemacs-filepath)
   (defvar dotspacemacs-show-startup-list-numbers)
   (defvar dotspacemacs-startup-banner)
+  (defvar dotspacemacs-startup-banner-scale)
   (defvar dotspacemacs-startup-buffer-show-icons)
   (defvar spacemacs-badge-official-png)
   (defvar spacemacs-banner-directory)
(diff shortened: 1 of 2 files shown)


lebensterben:
@dalanicolai 

Ideally we should not let user decide what the scale to be.

Instead, we should get the height of the window and make sure the logo isn't too large relative to the window height.

dalanicolai:
Okay, that might be nice. I have updated the PR, but note that when using the `'auto` value then it scales to a minimum of 1 line height.

dalanicolai:
I have additionally set the `auto` value to a maximum of the original image size (scale 1)

dalanicolai:
Now I have changed the minimum size of the banner to be 3 line heights (1 line height looks ridiculously small)

lebensterben:
@dalanicolai 
I need some time before I can review this

dalanicolai:
No problem... there are more important things than having a scalable banner (I would just prefer new users to see the nice labels and messages at the bottom of the home buffer)...

dalanicolai:
I've added some useful comments to make the review easier... but no hurries

smile13241324:
Sounds cool and looks good still I would like to have test first.

core/core-spacemacs-buffer.el
```diff
@@ -261,15 +262,48 @@ If ALL is non-nil then truly all banners can be selected."
...
+                                                        ;; number of entries
+                                                        ;; number of lines
+                                                        (+ 2 (cdr l)))
+                                                      dotspacemacs-startup-lists)))
+         (mode)
+         ;; We determine the maximum available banner height by subtracting the
+         ;; number of lines in the home buffer contents (excl. logo and
+         ;; startup-list) and the number of lines in the startup list from the
+         ;; total available text lines
+         (image-height (- (window-text-height) 26 startup-list-line-height)))
```
smile13241324, line 282: I think adding a proper constant name is more clear than a magic number of "26".

core/core-spacemacs-buffer.el
```diff
@@ -261,15 +262,48 @@ If ALL is non-nil then truly all banners can be selected."
...
+Returns height in units of line height with a minimum of 1."
+  ;; first determine number of lines occupied by startup list
+  (let* ((startup-list-line-height (apply #'+ (mapcar (lambda (l)
+                                                        ;; each section has 2
+                                                        ;; (header and
+                                                        ;; separator) plus
+                                                        ;; number of entries
+                                                        ;; number of lines
+                                                        (+ 2 (cdr l)))
+                                                      dotspacemacs-startup-lists)))
```
smile13241324, line 276: This works, however it would be much cooler if the scale would be determined by the actual size of the list on the home buffer and not on the potential size. Right now I have 4 projects however my project list is set to store up to 40 resulting in the logo being too small for the actual content on the screen.

smile13241324: After checking the code I have noticed this won't be too easy, if you don't want to implement this please put the default scale to "1" for now and I merge it, adding the optimisation later.

smile13241324 (changes requested):

core/core-spacemacs-buffer.el
dalanicolai: The first point turned out not to be so much work, so I have implemented that.
As far as I understand your second point, a scaling of 1 is already the default maximum scaling
(see the comment `return factor with maximum of 1` in the `spacemacs-buffer//insert-image-banner` below)

core/core-spacemacs-buffer.el
dalanicolai: It is quite hard to come up with a proper concise variable name for this. However, I have added the number to the comment to make the relation between the meaning of the comment and the code 'more explicit'.

core/core-spacemacs-buffer.el
smile13241324: Yes however as long as the scale is so aggressiv I would keep the default value in the dotfile to be "1" instead of "auto".

dalanicolai: Ah okay, but what do you mean with 'aggressive'? I mean, I would call the current behavior 'gentle' (at least how it behaves here, and I hope now also there)...

smile13241324: Yeah, what I mean is that the scale of the picture is derived from the potencial size of the lists and not from the actual list content. This is agressive as it may reserve more room for the list of links than what is actually necessary.

Especially I am afraid that the default scale will be too small with a fresh dotfile.

Therefore I would like to have the default value not set to 'auto but to 1 to keep the original scale by default until the scale algorithm is improved not to reserve more space then needed.

Or have this improvement already integrated into this PR.

dalanicolai: Ah okay, but that is what I meant with 'I have implemented your first point'; I meant that I have adapted the PR so that the scaling is derived from the actual size (In my previous comment I wrote 'startup buffer size' but I meant 'startup buffer list size). So I guess the PR (with the default still set to `auto`) is okay now.

smile13241324: Ahhh sorry than this was a big misunderstanding from my side. I am merging your changes.

smile13241324 (approved):

PR #15364, opened 2022-02-20
Improve helm completion in spacemacs help

dankessler:
Several (but not all) of the functions responsible for constructing helm sources
for use with spacemacs help simply construct alists rather than using the
'helm-build-*-source' macros. The former approach is discouraged by the [helm wiki](https://github.com/emacs-helm/helm/wiki/Developing#sources).

One of the consequences of this approach is that some of the nice helm features
one might expect don't work. For example, before adopting this commit, try
bringing up the layer help with 'SPC h l', and then type 'evil space'. One would
expect this to match 'spacemacs-evil' (since space-separated "words" are treated as
separate search strings), but this match doesn't work. Even worse, just type
'evil ' (with a trailing space): there will be no candidates (presumably because
it is treating the space literally).

This issue was present with the layer help, the package help, and the dotfile
help. I noticed the weird behavior a long time ago when I was first learning
spacemacs, and it was frustrating because it limited discoverability.

One note: with the previous approach, the alists had an unspecified (nil)
`candidate-number-limit`, which caused helm to not limit the number of options.
However, setting the ':candidate-number-limit' to nil with this approach does
not work: when helm builds a source from an 'eieio' object, it *only* copies
over slots that have non-nil values, so setting the keyword to nil yields an
object with a nil slot (which is also its initform), and then it does not get
copied into the helm source. Instead, I have set the keyword to the same large
constant used internally by helm when the user intends for the list to be
unlimited.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -816,6 +816,7 @@ Other:
     default (thanks to Daniel Nicolai)
   - Add evil-jump-backward/forward (=C-o=/=C-i=) keybindings to
     evil-evilified-state-map (thanks to Daniel Nicolai)
+  - Improve helm completion in spacemacs-help (thanks to @dankessler)
 *** Distribution changes
 - Refactored =spacemacs-bootstrap=, =spacemacs-ui=, and =spacemacs-ui-visual=
   layers:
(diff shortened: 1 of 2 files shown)


lebensterben (approved):

dankessler:
Just wanted to ping to see if there was anything further I could do to get this PR accepted. It was approved back in Feb 21 but never merged. I know y'all are volunteers and very busy so I hope I don't seem impatient or ungrateful, but just wanted to make sure this didn't slip through the cracks after approval.

lebensterben:
merged. thanks for your contributions.

PR #15369, opened 2022-02-21
fix: do not setenv PAGER in shell layer

kassick:
Setting PAGER with setenv changes the environment variable for any child process
launched fom within emacs, such as terminals apps (gnome-terminal, tilix,
etc.). This can interfere with paginated tools --  pydoc, bat and others.

This commit sets PAGER only within eshell sessions.

### DIFF
--- a/layers/+tools/shell/packages.el
+++ b/layers/+tools/shell/packages.el
@@ -124,7 +124,10 @@
       ;; quick commands
       (defalias 'eshell/e 'find-file-other-window)
       (defalias 'eshell/d 'dired)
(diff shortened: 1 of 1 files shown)


lebensterben:
@kassick 
What's the motivation for setting PAGER in the first place?

kassick:
@lebensterben I guess it's because eshell did not play well with `less` in spacemacs -- I remember having some trouble with it when I tried eshell in the past.

I tried the git log, but there's not much context why this environment has been set in the past

PR #15371, opened 2022-02-22
[osx] don't set dired-listing-switches

### LINKED ISSUE #15370: [osx] including `"--group-directories-first"` in `dired-listing-switches` is problematic
At present, the osx layer includes [some logic](https://github.com/syl20bnr/spacemacs/blob/4f66e93ed94451dace943a25e3511d64a30c5495/layers/%2Bos/osx/packages.el#L44-L47) which checks if `gls` is available, and if so, configures things so that `gls` will be used by dired (which is probably a good thing), but then it goes on to set `dired-listing-switches` to `"-aBhl --group-directories-first"`, which IMHO is problematic for a few reasons.

1. It breaks the directory-grouping functionality in `dired-quick-sort`. To see this, open up a dired buffer, then hit `s` to pop the `dired-quick-sort` hydra. Regardless of whether I hit `g` (group directories) or `G` (don't group directories), directories are still grouped. This is because `dired-quick-sort` just adds the `--group-directories-first` flag to the *end* of the switches, so when it's "on" the flag is actually passed twice, and when it's "off" the flag is still passed once.

2. It can screw things up when using tramp to list a remote directory. There's more discussion of this [on this doom issue](https://github.com/hlissner/doom-emacs/issues/1703). I noticed it when using `yadm` to manage my local dotfiles and where I have [registered it as a tramp method](https://github.com/TheLocehiliosan/yadm/blob/3d82aff3e8140361fca18f664fffa64dfd33296e/yadm.md?plain=1#L119-L128); I imagine the issue would also surface if using tramp to connect to another Mac (or a BSD machine). In brief, tramp seems to always want to use `ls` on the remote (even if you've told it to use `gls` locally), and so if the remote's `ls` command doesn't support `--group-directories-first` (which the OSX `ls` does not), then it bricks dired.

Without the `osx` layer, then the `dired-listing-switches` are not modified (and are left at their default `-al`). I think it would probably be better if the `osx` layer also left them alone (although swapping to use `gls` locally is indeed nice), and then the user could enable directory grouping if they so wished using `dired-quick-sort` (which is owned by by `spacemacs-editing` layer).

There are various workarounds for this, see [SO](https://stackoverflow.com/questions/22971299/conditionally-set-dired-listing-switches-locally-but-nil-remotely), the [solution to this doom issue](https://github.com/hlissner/doom-emacs/issues/1703), or even the [TRAMP manual's FAQ](https://www.gnu.org/software/tramp/#Frequently-Asked-Questions) (search for `dired-actual-switches`). However, it seems cleanest to just not cause the problem in the first place as a side effect of the `osx` layer and to instead let users manually turn on `--group-directories-first` either through `dired-quick-sort` or their own customization (in which case they may be sophisticated enough to enable one of the linked workarounds).

I can of course submit a simple PR that just axes the expression that sets `dired-listing-switches` in the `osx` layer, but since I imagine it's a matter of opinion/taste thought it might be better to discuss as an issue first.

dankessler:
fixes #15370; see there for more explanation

Note: Actively using `dired-quick-sort` will still break things via tramp when the remote is using a version of `ls` that doesn't support `--sort=` or `--group-directories-first`, but at least now it doesn't start off broken :)

### DIFF
--- a/layers/+os/osx/packages.el
+++ b/layers/+os/osx/packages.el
@@ -43,8 +43,7 @@
   ;; not using GNU ls.
   (let ((gls (executable-find "gls")))
     (when gls
-      (setq insert-directory-program gls
-            dired-listing-switches "-aBhl --group-directories-first"))))
+      (setq insert-directory-program gls))))
 
 (defun osx/pre-init-helm ()
   ;; Use `mdfind' instead of `locate'.


lebensterben:
Thanks.

PR #15372, opened 2022-02-22
[scala] add keybinding for running `Test / compile` in SBT

Kazark:
This is another standard command (in addition to [these](https://github.com/syl20bnr/spacemacs/pull/15160)), not one specific to my build configuration. It can be useful only to compile the tests if your tests require any setup or do not run quickly.


lebensterben:
Scala's LSP (Metals) supports sbt.
Does lsp-mode expose those features?

Kazark:
To add some more detail: `sbt-mode` lets SBT build in a terminal window, and it integrations with `compilation-mode` stuff so that I can navigate my errors with `SPC c n`, &c. Metals does not do this.

PR #15376, opened 2022-02-23
reinstate cfgl-layer docstring fix

dankessler:
I was mucking around in the `cfgl-layer` class and noticed I wasn't able to get help to come up, i.e., if I did `SPC h d f cfgs-layer RET` I'd get an error about `(concat...)` not satisfying `stringp`. I did some digging, and it seems to me that `:documentation` is not evaluated and needs to be a literal string. This was actually previous fixed back in fcf186faff8afdeda51c3dd20b61943e876acae5, but was subsequently accidentally reverted in dc58801c7ddd770a1f94a17890b4efbd1f20f6e8.

This PR restores the fix of fcf186faff8afdeda51c3dd20b61943e876acae5.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -184,8 +184,7 @@ subdirectory of ROOT is used."
             :initform 'unspecified
             :type (satisfies (lambda (x) (or (listp x) (eq 'unspecified x))))
             :documentation
-            (concat "A list of layers where this layer is enabled. "
-                    "(Takes precedence over `:disabled-for'.)"))
+            "A list of layers where this layer is enabled. (Takes precedence over `:disabled-for'.)")
    ;; Note:
    ;; 'can-shadow' is a commutative relation:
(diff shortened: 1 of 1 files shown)


core/core-configuration-layer.el
```diff
@@ -184,8 +184,8 @@ subdirectory of ROOT is used."
             :initform 'unspecified
             :type (satisfies (lambda (x) (or (listp x) (eq 'unspecified x))))
             :documentation
-            (concat "A list of layers where this layer is enabled. "
-                    "(Takes precedence over `:disabled-for'.)"))
+            "A list of layers where this layer is enabled.
```
lebensterben, line 187: you added an extra newline here

dankessler: good catch. I was over-zealous about keeping things under 80 columns wide

lebensterben:
merged. thanks.

PR #15378, opened 2022-02-24
Save current buffer when building Tex document.

sinistersnare:
This is useful when you are not working in the TeX-master-file.

Im not an ELisp programmer, so please let me know if theres a better way of doing this!

This isnt a big change, so I wont edit the CHANGELOG.

Have a good day!

### DIFF
--- a/layers/+lang/latex/funcs.el
+++ b/layers/+lang/latex/funcs.el
@@ -70,6 +70,7 @@
   (interactive)
   (progn
     (let ((TeX-save-query nil))
+      (save-buffer)
       (TeX-save-document (TeX-master-file)))
     (TeX-command latex-build-command 'TeX-master-file -1)))
 ;; (setq build-proc (TeX-command latex-build-command 'TeX-master-file -1))


lebensterben:
`TeX-save-document` says:
> Save all files belonging to the current document.

So I wonder what buffer would not be included.

sinistersnare:
I am not sure, but I think it started when I renamed some of my files from `.tex` to `.inc` (so LatexMk doesnt try to compile them individually).

I looked for the definition of `Tex-save-document` but could not find it.

lebensterben:
@sinistersnare 
After opening a TeX file, hit "C-h f" and type TeX-save-document.

sinistersnare:
Alright, when I do `M-x TeX-save-document` it works, but not when I do `SPC m c` to build the document (calling the `latex/build` function im modifying in this PR).

I think the better solution is to remove `(Tex-master-file)` from the `Tex-save-document` function call. That way, it saves all document files and not just the master.

lebensterben:
what if you replaced it by

(call-interactively #'TeX-save-document)

sinistersnare:
That doesnt work either. i tried replacing the call to `TeX-save-document` with that, and also putting them both together in the `let` body. Neither saved the `.inc` file that I was working on. To be clear, I did explicitly set the `.tex` master file as master using `TeX-master-file-ask`.

lebensterben:
What I suggested is
```elisp
    (let ((TeX-save-query nil))
      (call-interactively #'TeX-save-document))
```
And when it's called in this way it's equivalent to `(TeX-save-document (TeX-master-file)` according to the source code.

sinistersnare:
Yeah, thats what I tried. And I thought it would work too, but surprisingly it didnt.

lebensterben:
Can you unset the master file?

lebensterben:
It's recommended in acutex's documentation that you should set master file to nil if you use \include.

sinistersnare:
I restarted emacs to unset the master file, and it doesnt fix it. I also dont use `\include` i use `\input`, which I think is a better choice for my use-case.

smile13241324 (approved):

smile13241324:
I have just had a look into this layer and the state is horrible. Still I think a static save is not a nice solution here given that auctex is trying to follow the file link itself. If they ignore input and only follow include this should be a bug on their side.

If the interactive call fixes it for you feel free to include the interactive call solution from Lebensterben.

smile13241324 (changes requested):
Please do not do a simple save buffer instead try to use the auctex function.

sinistersnare:
The fixes recommended by Lebensterben did not work for me in this case, and that is how I landed on the current cudgel of an approach.

If the changes need to be more extended, I would not be capable of completing them. It would be better perhaps to close this PR and make an issue then.

smile13241324:
I think we have to restart on this issue, I have upgraded to emacs 29 some days ago and noticed some bug reports with the latex layer anyway. Lets revise this layer and have this issue fixed while I am working on it. I will add a node.

PR #15381, opened 2022-02-28
[github] mark layer as deprecated - remove unsecure packages

### LINKED ISSUE #15377: github layer: docs for authentication outdated
https://develop.spacemacs.org/layers/+source-control/git/README.html explains how to generate a GitHub auth token and add it to your git config. But this is unclear in multiple ways:

-  It doesn't explain to which of the packages to auth info applies. (Note that this doesn't really spark confidence in the reader, it's a little bit like "Paste your token here and spacemacs will use it for some stuff but I don't explain what stuff"). AFAIU the config applies only packages using `gh.el` as a backend, which is true for `github-clone` but not for the other packages in the layer. `forge` uses a different backend for instance (https://magit.vc/manual/forge/Token-Creation.html)
- The docs say the token should have the `gist` permission but the gist package was removed from the layer. 

Maybe for now, it will be easier to just refer to the package documentation...

practicalli-john:
Purpose of the PR changed after discussion and further review of the `github` layer

Remove github-clone & github-search packages from the layer, their key bindings and associated documentation.  Magit in the `git` layer can support cloning in a more secure way.

Update documentation:
- add depreciated layer notice and explanation
- remove reference to Magit forge (forge is added via the `git` layer)
- remove reference to git-link (git-link is added via the `git` layer
- Remove documentation regarding a GitHub access token (no longer required)

`grip-mode` and its configuration / documentation remains as the only functionality in the layer until it is decided what to to with this package.

Resolve #15377

### DIFF
--- a/layers/+source-control/github/README.org
+++ b/layers/+source-control/github/README.org
@@ -9,19 +9,21 @@
   - [[#features][Features:]]
 - [[#install][Install]]
   - [[#layer][Layer]]
-  - [[#git-configuration][Git configuration]]
   - [[#grip-mode][grip-mode]]
 - [[#key-bindings][Key bindings]]
-  - [[#clone-repositories][Clone repositories]]
   - [[#grip-mode-1][Grip-mode]]
 
 * Description
-This layers adds support for [[http://github.com][GitHub]].
+*This layer is deprecated*.
+
+The layer had support searching for and cloning [[http://github.com][GitHub]] repositories,
+although that did not provide a secure way to store an access token
+to connect to GitHub.
+
+The git layer provides interaction with GitHub and can securely
+store the access token using PGP.
 
 ** Features:
-- [[https://github.com/magit/forge][forge]]: integration with various forges, such as GitHub and GitLab.
-- [[https://github.com/sshaw/git-link][git-link]]: quickly generate URLs for commits or files.
-- [[https://github.com/dgtized/github-clone.el][github-clone]] allows for easy cloning and forking of repositories.
 - [[https://github.com/seagle0128/grip-mode][grip-mode]] Github-flavored Markdown/Org preview using [[https://github.com/joeyespo/grip][Grip]].
 
 * Install
@@ -30,17 +32,6 @@ To use this configuration layer, add it to your =~/.spacemacs=. You will need to
 add =github= to the existing =dotspacemacs-configuration-layers= list in this
 file.
 
-** Git configuration
-You will need to generate a [[https://help.github.com/en/articles/creating-a-personal-access-token-for-the-command-line][personal access token]] on GitHub. This token should
-have the =gist= and =repo= permissions. Once this token is created, it needs to
-be added to your =~/.gitconfig=
-
-You will also need to [[https://help.github.com/articles/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent/][generate an SSH key]] and [[https://help.github.com/articles/adding-a-new-ssh-key-to-your-github-account/][add it to your GitHub account]].
-
-#+BEGIN_SRC sh
-  git config --global github.oauth-token <token>
-#+END_SRC
-
 ** grip-mode
 Grip-mode [[https://github.com/seagle0128/grip-mode#prerequisite][requires python and the python package grip]] to be installed on the
 system. Grip can usually be installed with the command =pip install grip=,
@@ -48,15 +39,6 @@ on debian based systems make sure you use the python 3 version
 of pip, it is normally called =pip3=.
 
 * Key bindings
-** Clone repositories
-
-| Key binding   | Description                                              |
-|---------------+----------------------------------------------------------|
-| ~SPC g h c /~ | search for a repository to clone it                      |
-| ~SPC g h c c~ | clone and optionally fork repository                     |
-| ~SPC g h c r~ | add a remote that is an existing fork of selected remote |
-| ~SPC g h c f~ | fork remote in current user namespace                    |
-| ~SPC g h c u~ | add upstream as remote                                   |
 
 ** Grip-mode
 
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2096,6 +2096,9 @@ Other:
 - Show user error when location is not set in theme changer
   (thanks to Boris Buliga)
 **** GitHub
+- Layer:
+  - marked as deprecated as functionality is all in =git= layer (thanks
+    to practicalli-john)
 - Packages:
   - Added new packages =forge= (thanks to Miciah Dashiel Butler Masters)
   - Do not install =forge= on Windows by default, see =README.org= of the layer
@@ -2123,6 +2126,8 @@ Other:
   - Remove package =magit-gh-pulls= as it was deprecated in favor of =magithub=
     and =forge= (thanks to Robby O'Connor)
   - Made magithub offline by default (thanks to Miciah Masters)
+- Removed =git-clone= and  =git-search= packages for not using a secure means to
+  store access tokens #1153777 (thanks to @practicalli-john)
 **** Git
 - Key bindings:
   - ~SPC g c~ to clone a repository (thanks to Steven Allen)
--- a/layers/+source-control/github/packages.el
+++ b/layers/+source-control/github/packages.el
@@ -20,32 +20,16 @@
 ;; You should have received a copy of the GNU General Public License
 ;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+;; Package deprecation notice shown at startup
+(warn "`github' layer is deprecated. See layer README.org for details.")
+
 
 (defconst github-packages
   '(
-    github-clone
-    github-search
     grip-mode
     ;; this package does not exits, we need it to wrap
     ;; the call to spacemacs/declare-prefix.
     (spacemacs-github :location built-in)))
-
-(defun github/init-github-clone ()
-  (use-package github-clone
-    :defer t
-    :init
-    (progn
-      (spacemacs/declare-prefix "ghc" "clone")
-      (spacemacs/set-leader-keys
-        "ghcc" 'github-clone
-        "ghcr" 'github-clone-add-existing-remote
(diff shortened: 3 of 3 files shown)


arifer612 (approved):
All of these changes look good! The README is much clearer and should be easy for newer users to follow and understand.

lebensterben:
yes this does solve the issue and makes the documentation much easier to follow.

but it's not a good idea to add GitHub token in `.gitconfig`.

(well, more secure alternatives are available but then they're much harder for beginners)

real-or-random:
Does storing in `.gitconfig` actually work? (Do you have a reference?) The forge docs refer to https://magit.vc/manual/ghub/Storing-a-Token.html,  which only mentions `.authinfo`.

layers/+source-control/github/README.org
```diff
@@ -31,16 +31,22 @@ add =github= to the existing =dotspacemacs-configuration-layers= list in this
...
-You will also need to [[https://help.github.com/articles/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent/][generate an SSH key]] and [[https://help.github.com/articles/adding-a-new-ssh-key-to-your-github-account/][add it to your GitHub account]].
+Add the token and GitHub account username to your =~/.gitconfig=
 
 #+BEGIN_SRC sh
+  git config --global github.user <github-username>
   git config --global github.oauth-token <token>
 #+END_SRC
 
+[[https://help.github.com/articles/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent/][Generate an SSH key]] and [[https://help.github.com/articles/adding-a-new-ssh-key-to-your-github-account/][add it to your GitHub account]] to remove the need
+to provide GitHub login details for each request from Magit to GitHub.
```
real-or-random, line 45: s/Magit/Magit Forge ?

practicalli-john:
There is no mention of using authinfo.gpg in the github-clone.el package, although it does suggest a way to help avoid checking in a dotfile with the api token - by referencing a separate file
https://github.com/dgtized/github-clone.el#securing-the-keys

This seems a useful thing to mention at least.

real-or-random:
Right, so at the moment, I don't know if this PR actually solves the issue because I believe `github-clone` and `forge` store their credentials differently. That's sad but it is how it is. So if we you want to have both packages in the layer, I think we ideally would have a separate section for each.

practicalli-john:
> Right, so at the moment, I don't know if this PR actually solves the issue because I believe `github-clone` and `forge` store their credentials differently. That's sad but it is how it is. So if we you want to have both packages in the layer, I think we ideally would have a separate section for each.

Yes, when reviewing the docs it made sense to have different config for github-clone and magit forge (and ssh key).  I've pushed an update that should make this much clearer.

layers/+source-control/github/README.org
practicalli-john: Yes, this needs a little more elaboration.  An SSH key should be used when using SSH URLs for remote repositories with Magit.  The SSH key is not needed for Forge.  
I've pushed an update to this section to clarify.

lebensterben:
I really want to remove github-clone.

It has one single purpose/functionality, but it comes with a huge security risk. (making a .gitconfig.private isn't secure at all)

Fixing it requires quite some work, but there's not much gain given that's this package isn't very useful in the first place...

practicalli-john:
> I really want to remove github-clone.
> 
> It has one single purpose/functionality, but it comes with a huge security risk. (making a .gitconfig.private isn't secure at all)
> 
> Fixing it requires quite some work, but there's not much gain given that's this package isn't very useful in the first place...

I assume it is gh.el that requires changing rather than github-clone.el, as it is the latter package that reads / writes the token to the .gitconfig file.

I've realised that I actually use `magit-clone` rather than `github-clone` so would not miss the removal of github-clone.el.  I do no use the search feature of github-clone, which seems to be the only functionality not already in Magit.

I assume the removal of github-clone would be a separate issue.  Although if we are all in agreement I could change this PR to remove github-clone and update the GitHub layer documentation.

lebensterben:
@practicalli-john yes that's a better idea.

Magit offers better security and receives more frequent maintenance.

When you remove github-clone, please also replace the key-bindings if there are equivalents in magit.

real-or-random:
I agree that `github-clone` is not worth keeping. First, most of its functionality is included in `magit` (and everyone uses `magit`), and second, cloning is rather a rare task. I mean I commit/push/pull/rebase basically daily but I don't need clone daily, and I believe this is true for most users. And then having separate config (no matter if secure or not) is a lot of overhead for a single package.

I leave the question of what changes should go to what PR to @lebensterben . ;)

practicalli-john:
@lebensterben 
After further investigation I'm of the opinion there is no reason to actually have a separate `github` layer.

1) The `github` layer does not add Magit Forge, this is part of the `git` layer (although forge may have been part of github at one point) 

2) git-link package is already included in the `git` layer and has no configuration in the github 

3) `git-clone` and `git-search` packages both require an oauth personal token in .gitconfig and we have already agreed to remove these packages along with their associated keybinding code from the GitHub layer.

4) The [grip-mode](https://github.com/seagle0128/grip-mode) package for GitHub-flavored Markdown/Org preview using [Grip](https://github.com/joeyespo/grip) should be part of the `markdown` layer.  The `markdown` layer already has a [vmd-mode](https://github.com/blak3mill3r/vmd-mode) for GitHub flavored live previews, so grip-mode could be added as optional package or simply dropped.

#### Proposal
Delete the `github` layer from Spacemacs (via a new PR)
Add Magit Forge documentation to the `git` layer showing how to create a machine configuration for a forge (using this PR)
Either move grip-mode to the markdown as an option or drop it altogether and just provide vmd-mode in the `markdown` layer (a  new PR on the markdown layer)

Thoughts?

lebensterben:
yes I agree with you that github layer is redundant.

The only obstacle is we don't have a formal way of removing a layer...

I suggest to have a github/packages.el with

    (warn "`github' layer has been deprecated. See layer README.org for details.")

This way user sees a warning on startup.

And in github/README.org mentions that user should consult git layer for Magit forge, and enable grip-mode from markdown layer.

dalanicolai:
Removing the github layer sounds good to me, however, I am not sure about where to move `grip-mode`. I have motivated its addition [here](https://github.com/syl20bnr/spacemacs/pull/14120#issuecomment-726819821) (and also explained its advantage(s) over vmd, mainly it can be used also on org-mode files directly). So maybe it should be moved to the org-layer, as grip-mode was added for users who prefer to write README's using org-mode syntax (and who prefer using pip over using npm).

practicalli-john:
A separate PR #15388 has been created to add Magit Forge configuration to the 'git` layer, as it is the 'git' layer that included Magit Forge package.

arifer612 (approved):
The changes look good to me! The deprecation warning is sufficiently self-explanatory.

arifer612 (commented):
Since everything but `grip` is removed, I would suggest either renaming this layer or moving it to a different layer as a toggled package e.g. `'(grip-mode :toggle enable-grip-mode)`. The latter is more attractive to me since a layer just for a single package that only adds a single toggle for the mode is very redundant. 

Although I understand @dalanicolai's reasoning in #14120 for having them in the github layer back then, but with it now unofficially deprecated, we should probably move `grip` into the `markdown/org tools` layer instead. The package is, at the very core, a way to render markdown/org files.

practicalli-john:
@arifer612 I suggested that grip-mode be moved in a separate PR, once someone decides which other layer it is supposed to be a part of.  I've never used grip or any intention to use it, so have no input on this package.

I would think renaming the layer is redundant if the grip-mode is going to be moved to another layer.  Also, what happens to Spacemacs on startup when `github` layer is in the configuration but the layer no longer exists (as its been renamed). 

I dont have any time to work on this for the next few days, so will circle back at the weekend to see if any decision has been taken.  Thank you.

arifer612:
@lebensterben Can we get you to verify this and merge this in?

lebensterben:
@arifer612 

Please give me some time.

arifer612:
Oh my bad 😲 no rush, sorry for the ping!

smile13241324 (approved):

smile13241324:
Adding the depreciation message should be ok therefore I have merged this change.

Moving the grip package to markdown layer would be the most sensible thing to do. This would be a breaking change though so we should first wait a bit (2 weeks?) before we remove the github layer.

We could announce running depreciation periods in the home buffers release notes message manually.

Or produce a formal mechanic in the core with depreceation dates which could warn the user about these periods in a warning buffer.

What do you think @lebensterben?

eeshugerman:
Removing the `github` layer breaks `SPC g o`/`browse-at-remote`. Will this functionality be reincorporated somewhere?

lebensterben:
@smile13241324 sorry I missed the comment. I will review the issue raised by @eeshugerman.

lebensterben:
@eeshugerman 
It's unclear to me why browse-at-remote ceases to work. It's defined in version-control layer. See https://github.com/syl20bnr/spacemacs/blob/1bd5428360bf5a45afe9109e86576b682a39ed62/layers/%2Bsource-control/version-control/packages.el#L296

And it should not rely on github layer (or any of its packages).

smile13241324:
> Removing the `github` layer appears to break `SPC g o`/`browse-at-remote`. Will this functionality be reincorporated somewhere?

Can you provide a reproduction guide to see what fumctionality is broken by removing the github layer?

Ich just tried "spc g o" in a package.el file and it was working as expected.

eeshugerman:
> Can you provide a reproduction guide to see what functionality is broken by removing the github layer?
>
> Ich just tried "spc g o" in a package.el file and it was working as expected.

I haven't tried with a vanilla config, but:
1. remove `github` from `dotspacemacs-configuration-layers`
2. restart emacs (`spacemacs/sync-configuration-layers` is *not* sufficient)
3. `browse-at-remote` is no longer defined

However, I just tried adding the `version-control` layer (still without `github`) and `SPC g o`/`browse-at-remote` is back! It looks like I should probably have this layer enabled anyway. So there's no real issue here, but folks will probably want to enable `version-control` if they haven't already.

PR #15388, opened 2022-03-02
[git] document configuration of Magit Forge and Git identity

practicalli-john:
Add documentatation to the Spacemacs Git layer describing how to configure basic git identity and more importantly how to configure Magit Forge to access GitHub / GitLab using a machine configuration and an encrypted file to hold the personal access token.

PR #15395, opened 2022-03-05
[git] Extend documentation for SSH keys

bryce-carson:
This extends the configuration information for using SSH keys.

It does not make use of authentication tokens, purely SSH. You can then decide whether you want to use tokens with git-forge stuff <kbd>SPC g m @</kbd>.

This is my first significant pull request, so any feedback on my documentation writing is welcome, as well as what I might have missed in the contribution guidelines.

---

It is dependent upon @practicalli-john's pull request (#15388). I tried to follow @lebensterben's instructions on how to make this PR, but I'm not the best at Git. It's easy to be a git when it comes to Git.

### DIFF
--- a/layers/+source-control/git/README.org
+++ b/layers/+source-control/git/README.org
@@ -116,6 +116,47 @@ Emacs to acquire SSH keys for remote logins from GNOME or KDE key-chains (or
 
    ~SPC q R~, or ~SPC f d~ (to delete the client frame) then ~systemctl --user restart emacs~.
 
+4. Add the key to the SSH configuration file, so that when an SSH connection is
+   made to a Git forge, the appropriate key is used with that host by the
+   ssh-agent.
+
+   #+begin_src emacs-lisp
+     ((lambda (forgeUsername
+               sshKeyName
+               (&optional (gitForgeProperName "Github")
+                          (gitForgeHostname   "github.com")
+                          (addKeyToAgent-p    "yes")))
+        "Add a new SSH key to the user's login keyring, and\
+             configure it for a Git forge."
+        nil
+        (append-to-file
+         (concat "\n"
+                 "Host " gitForgeProperName "\n"
+
+                 ;; The .pub file extension is not necessary; it may cause
+                 ;; issues to include it. See man 5 ssh_config
+                 "     IdentityFile ~/.ssh/" sshKeyName "\n"
+                 "     User " forgeUsername "\n"
+                 "     Hostname " gitForgeHostname "\n"
+                 "     AddKeysToAgent" addKeyToAgent-p)
+         nil
+         "~/.ssh/config"))
+
+      "Your Git forge username"
+      "The path to your SSH key"
+
+      ;; ;; Optionally specify alternative to GitHub, and not add key to agent
+      ;; ;; permanently.
+      ;; "GitLab"
+      ;; "gitlab.com"
+      ;; ;; Either "no", or "confirm".
+      ;; "no"
+      )
+   #+end_src
+
+   You can evaluate the above source block to configure SSH fully for working
+   with a Git forge. Ensure to set the variables to their appropriate values.
+
 ** Magit status fullscreen
 To display the =magit status= buffer in fullscreen set the variable
 =git-magit-status-fullscreen= to =t= in your =dotspacemacs/user-init= function.
--- a/layers/+source-control/git/README.org
+++ b/layers/+source-control/git/README.org
@@ -93,11 +93,11 @@ lifetime of Emacs", as the /GNU Emacs Manual/ states [Free Software Foundation
 2021]).
 
 1. Configure Emacs to use login or session keyrings. The default configuration of
-Emacs (and Spacemacs) does not include this. Login and session keyrings are
-managed by the ssh-agent of the desktop environment you use on Linux.
+   Emacs (and Spacemacs) does not include this. Login and session keyrings are
+   managed by the ssh-agent of the desktop environment you use on Linux.
 
-   - If you use Windows, you will need to consult extra documentation not discussed
-here.
+   - If you use Windows, you will need to consult extra documentation not
+     discussed here.
 
    Add the following to your ~spacemacs/user-config~ function, as shown below.
 
@@ -127,7 +127,7 @@ here.
         nil
         (append-to-file
          (concat "\n"
-                 "Host " gitForgeProperName "\n"
+                 "HOST " gitForgeProperName "\n"
 
                  ;; The .pub file extension is not necessary; it may cause
                  ;; issues to include it. See man 5 ssh_config
@@ -153,6 +153,22 @@ here.
    You can evaluate the above source block to configure SSH fully for working
    with a Git forge. Ensure to set the variables to their appropriate values.
 
+3. Ensure the upstream URLs for the remote repository are SSH form, according to
+   the documentation for the Git forge you are working with.
+
+   GitHub's documentation says the remote repository URL to use for SSH, rather
+   than HTTPS, is:
+
+   #+begin_example
+   git@github.com:syl20bnr/repository.git
+   #+end_example
+
+   Other Git forges may have a different URL scheme. Check the documentation for
+   the forge you would like to use.
+
+Below is a fully worked example of configuring SSH for use with Magit. You will
+be able to clone, pull, and push as normal.
+
 E.g.: Configure Emacs to use the desktop environment login key-chain, ie tell
 Emacs to acquire SSH keys for remote logins from GNOME or KDE key-chains (or
 "wallets"). This is provided by the Free Desktop standard [[https://www.gnu.org/software/emacs/manual/html_mono/auth.html#Secret-Service-API][Secret Service API]].
@@ -162,8 +178,7 @@ Emacs to acquire SSH keys for remote logins from GNOME or KDE key-chains (or
      (setq auth-sources '(default "secrets:Login"))
    #+end_src
 
-2. Add the =exampleRSAKey" to the SSH configuration.
-
+2. Add the =exampleRSAKey= to the SSH configuration.
    #+begin_src emacs-lisp
      ((lambda (forgeUsername
                sshKeyName
@@ -175,7 +190,7 @@ Emacs to acquire SSH keys for remote logins from GNOME or KDE key-chains (or
         nil
         (append-to-file
          (concat "\n"
-                 "Host " gitForgeProperName "\n"
+                 "HOST " gitForgeProperName "\n"
 
                  ;; The .pub file extension is not necessary; it may cause
                  ;; issues to include it. See man 5 ssh_config
@@ -189,6 +204,7 @@ Emacs to acquire SSH keys for remote logins from GNOME or KDE key-chains (or
       "bryce-carson"
       "~/ssh/exampleRSAKey"
       )
+   #+end_src
 
 3. Save and reload the configuration changes.
 
@@ -198,6 +214,14 @@ Emacs to acquire SSH keys for remote logins from GNOME or KDE key-chains (or
 
    ~SPC q R~, or ~SPC f d~ (to delete the client frame) then ~systemctl --user restart emacs~.
 
+5. Ensure the remote repository URL is configured.
+
+   1. Configure the remotes for the current local repository: ~SPC g m M C~.
+
+   2. Select the upstream, such as a fork of another repository, or the origin(al) repository, in the minibuffer.
+
+   3. Set the URL to the appropriate scheme for the forge (=url= & =pushurl=): ~git@github.com/syl20bnr/spacemacs.git~.
+      
 ** Magit status fullscreen
 To display the =magit status= buffer in fullscreen set the variable
 =git-magit-status-fullscreen= to =t= in your =dotspacemacs/user-init= function.
--- a/layers/+source-control/git/README.org
+++ b/layers/+source-control/git/README.org
@@ -138,7 +138,7 @@ Emacs to acquire SSH keys for remote logins from GNOME or KDE key-chains (or
                  "     IdentityFile ~/.ssh/" sshKeyName "\n"
                  "     User " forgeUsername "\n"
                  "     Hostname " gitForgeHostname "\n"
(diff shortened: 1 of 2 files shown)


real-or-random:
By the way, another nice way when using auth tokens is this:
```
(secrets-create-item "default" "auth-source magit-forge" "<auth-token>" :host "api.github.com" :port "https" :user "real-or-random^forge") 
```
I think it took me an hour to figure that command out, so I always wanted to share this. The secret should probably read via `read-string` instead to avoid storing it in command history.

bryce-carson:
> By the way, another nice way when using auth tokens is this:
> ```
> (secrets-create-item "default" "auth-source magit-forge" "<auth-token>" :host "api.github.com" :port "https" :user "real-or-random^forge") 
> ```
> I think it took me an hour to figure that command out, so I always wanted to share this. The secret should probably read via `read-string` instead to avoid storing it in command history.

Can you expand on how this relates to my contributions with SSH? I'm a little lost.

bryce-carson:
If you can write up how to use it it could be worthwhile to include as well in the changes to the git layer since its seeing a lot of attention right now.

layers/+source-control/git/README.org
```diff
@@ -59,6 +65,163 @@ file.
...
+      "The path to your SSH key"
+
+      ;; ;; Optionally specify alternative to GitHub, and not add key to agent
+      ;; ;; permanently.
+      ;; "GitLab"
+      ;; "gitlab.com"
+      ;; ;; Either "no", or "confirm".
+      ;; "no"
+      )
+   #+end_src
```
practicalli-john, line 151: It seems this script adds a host configuration to `~/.ssh/config`, which seems simpler to showing an example of that host configuration, e.g. for GitHub
```
HOST GitHub
IdentityFile ~/.ssh/rsa_id
User forgeusername
Hostname github.com
AddKeysToAgent yes
```

layers/+source-control/git/README.org
```diff
@@ -59,6 +65,163 @@ file.
...
+   ~SPC q R~, or ~SPC f d~ (to delete the client frame) then ~systemctl --user restart emacs~.
+
+5. Ensure the remote repository URL is configured.
+
+   1. Configure the remotes for the current local repository: ~SPC g m M C~.
+
+   2. Select the upstream, such as a fork of another repository, or the origin(al) repository, in the minibuffer.
+
+   3. Set the URL to the appropriate scheme for the forge (=url= & =pushurl=): ~git@github.com/syl20bnr/spacemacs.git~.
+
```
practicalli-john, line 224: Perhaps this section could be simplified to the following (unless I have misunderstood something)
```org
**** Magit Forge with SSH Key
An SSH key can be used instead of a personal access token to 

Set =auth-sources= variable to ="secrets:Login"= to use a login key-ring, unlocking
the key on login to the desktop environment.

#+begin_src emacs-lisp
   (setq auth-sources '(default "secrets:Login"))
#+end_src

Alternative, use ="secrets:session"= to open the key ring only when Emacs is running.

Add SSH configuration to =~/.ssh/config= 

   #+begin_src 
  HOST hostname              # e.g. GitHub / GitLab
  IdentityFile ~/.ssh/rsa_id
  User forgeusername          
  Hostname forge-domain.com  # e.g. github.com
  AddKeysToAgent yes
   #+end_src
```

practicalli-john:
@bryce-carson sorry, I think I am confused about the purpose of this configuration.

Initially I assumed the suggested configuration was to use an SSH key with Magit forge, avoiding the need for a `machine` entry in `.authinfo` or `authinfo.gpg`, which would be nice and simple.  However, in a quick test, this does not seem to work (Magit Forge prompts for a token after I remove .authinfo.gpg file  

As the content in this PR is added after the SSH keys for remote repositories section, which does not require the use of a personal access token, I am unclear what token you are trying to replace?

I am definitely not an expert in this area of configuration, so keen to learn.
Thank you.

bryce-carson:
@practicalli-john 
For myself, GitHub always yelled at me if I didn't setup SSH for pushing from the CLI.

As far as I was aware, I thought SSH key pairs were the preferred authentication method for working with remote repositories on GitHub. I'd seen *somewhere* that I cannot locate now that SSH was required, even.

Whereas, from my experience, HTTPS authentication was disallowed for working with repositories directly, using the API is still used with HTTPS + OAuth tokens.

It's all a little confusing, so I'm not sure.

Having documentation for how to setup SSH keys, agents, etc. with the Git layer is still helpful, even if it's not *necessary* if you're otherwise using a GPG encrypted token file for using the API with magit-forge.

practicalli-john:
I've always use SSH with Magit, but not needed the configure in this PR to do so. I am afraid I don't understand what it adds, as in my own experience it hasn't been needed.

Magit Forge however seems to require a token to access the GitHub or GitLab API, to work with issues, PRs, etc.
I haven't seen a way to use SSH key, although would be interested to know if it's possible.  I am still not clear if it's this aspect the PR is trying to solve.

Magit can clone/push/pull commits:
- an SSH key is used when the remote was added with an SSH URI
- a personal access token is used when the remote was added with HTTPS URI.

> GitHub denies access to HTTPS URI using a password (for over a year now)

The first time an SSH key is used with Magit, the desktop should prompt to save it to the key-ring.  This has worked for me on various Linux OS's, so assumed it's standard.  Is this not the case?

So I have never needed to do anything for Magit other than create the key (using ssh-keygen command line) and add the basic git identity.

The only time I've defined hosts is when I had separate SSH keys for several services (where I really should have used the same key everywhere).  Now I don't have an .ssh/config at all

So I am also unclear what this PR is trying to achieve with that part of config

real-or-random:
> > By the way, another nice way when using auth tokens is this:
> > ```
> > (secrets-create-item "default" "auth-source magit-forge" "<auth-token>" :host "api.github.com" :port "https" :user "real-or-random^forge") 
> > ```
> >     
> >   
> > I think it took me an hour to figure that command out, so I always wanted to share this. The secret should probably read via `read-string` instead to avoid storing it in command history.
> 
> Can you expand on how this relates to my contributions with SSH? I'm a little lost.

This is for storing GitHub personal authentication tokens (for HTTPS, not for SSH). The token is stored in a secret service backend (GNOME Keyring or KDE Wallet) instead of an `authinfo` file. This is hinted at in https://magit.vc/manual/ghub/Storing-a-Token.html but not fully described and it was kind of hard for me to figure out a working example just from the API docs in https://www.gnu.org/software/emacs/manual/html_mono/auth.html#Secret-Service-API (The parameters are not exactly natural,  also needs to set `auth-sources` etc...) Now that I think about it, I think the first natural place to write an elaborate version of this is the forge/ghub docs. @tarsius Do you think this will be a useful addition to the ghub docs?

bryce-carson:
> I've always use SSH with Magit, but not needed the configure in this PR to do so. I am afraid I don't understand what it adds, as in my own experience it hasn't been needed.

Yesm, as you note below that GitHub denies access to repositories when attempting to authenticate with a username and password over HTTPS, that led me to believe that SSH was the required authentication method when I was reading GitHub documentation.

> Magit Forge however seems to require a token to access the GitHub or GitLab API, to work with issues, PRs, etc. I haven't seen a way to use SSH key, although would be interested to know if it's possible. I am still not clear if it's this aspect the PR is trying to solve.

This PR is not related to magit-forge at all. It is limited to documenting how to authenticate with SSH to GitHub for basic Magit functionality.

> Magit can clone/push/pull commits:
> 
> * an SSH key is used when the remote was added with an SSH URI
> * a personal access token is used when the remote was added with HTTPS URI.
> 
> > GitHub denies access to HTTPS URI using a password (for over a year now)
> 
> The first time an SSH key is used with Magit, the desktop should prompt to save it to the key-ring. This has worked for me on various Linux OS's, so assumed it's standard. Is this not the case?
> 
> So I have never needed to do anything for Magit other than create the key (using ssh-keygen command line) and add the basic git identity.

This did not occur for me on my Fedora Linux machine. When I created my SSH key pairs for GitHub, I already had an `~/.ssh/config` because I used academic clusters. After creating an SSH key pair for GitHub, and registering the public key with GitHub, adding it to my keyring (using Seahorse), Magit would still try to authenticate with a username and password.

That may have been my fault, I had difficulties configuring the URI for a while, but once I had that resolved I had already configured the use of the Secret Service API (which I believe would be necessary anyways, since I wasn't using just a single key).

> The only time I've defined hosts is when I had separate SSH keys for several services (where I really should have used the same key everywhere). Now I don't have an .ssh/config at all
> 
> So I am also unclear what this PR is trying to achieve with that part of config

The configuration here, ultimately, is born from my personal struggle getting Magit to authenticate with GitHub through SSH. It would not prompt me for what key to use after configuring the local Git repository to use an SSH GitHub URI (`git@github.com:username/repository`).

I'm going to rework this PR so that it is more clear, more general, and also /tested/ in a clean environment to ensure that the instructions it gives are neither misleading or dependent upon any sort of system and file state.

bryce-carson:
> > > By the way, another nice way when using auth tokens is this:
> > > ```
> > > (secrets-create-item "default" "auth-source magit-forge" "<auth-token>" :host "api.github.com" :port "https" :user "real-or-random^forge") 
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > I think it took me an hour to figure that command out, so I always wanted to share this. The secret should probably read via `read-string` instead to avoid storing it in command history.
> > 
> > 
> > Can you expand on how this relates to my contributions with SSH? I'm a little lost.
> 
> This is for storing GitHub personal authentication tokens (for HTTPS, not for SSH). The token is stored in a secret service backend (GNOME Keyring or KDE Wallet) instead of an `authinfo` file. This is hinted at in https://magit.vc/manual/ghub/Storing-a-Token.html but not fully described and it was kind of hard for me to figure out a working example just from the API docs in https://www.gnu.org/software/emacs/manual/html_mono/auth.html#Secret-Service-API (The parameters are not exactly natural, also needs to set `auth-sources` etc...) Now that I think about it, I think the first natural place to write an elaborate version of this is the forge/ghub docs. @tarsius Do you think this will be a useful addition to the ghub docs?

That's a lovely addition. We should _definitely_ include it in the upstream ghub documentation, or here in Spacemacs if upstream doesn't integrate that. In the first case, we should direct users to the new ghub documentation after your PR goes through for that, if you'll be opening one after talking with @tarsius.

---

I prefer using GNOME keyring opposed to separate files (I can remember where to look easier). Storing OAuth tokens in GNOME Keyring is awesome.

I think I will re-work this documentation PR to be about the Secret Service API, rather than SSH. It can instruct users how to configure Emacs to use the API for either SSH or OAuth tokens, as you use.

---

@practicalli-john The necessity of OAuth tokens (GitHub does not document use of their API with SSH authentication) for using magit-forge, the ability to store OAuth tokens in system keyrings (`gnome-keyring`), and the implicit lack of _necessity_ for SSH with GitHub leads me to believe that I should document things in this way:

- You may use OAuth tokens or SSH to authenticate with Github, but SSH authentication is not available for use of the Github API
- The Github API is required for using features of `magit-forge`, so OAuth token authentication with https is preferential as you will not need to point to SSH URIs for remotes of local repositories

Then the instructions for the `git` layer can focus on two branches, after the generally applicable documentation:
- Using authinfo files
- Using the system keyring with the Secret Service API

real-or-random:
@bryce-carson This all sounds very reasonable to me.

---

> You may use OAuth tokens or SSH to authenticate with Github, 

One minor technical comment I have is that these are actually not OAuth tokens but they're called "Personal access tokens". The explanation at the bottom of the https://github.com/settings/tokens site makes is very precise here:

> Personal access tokens function like ordinary OAuth access tokens. They can be used instead of a password for Git over HTTPS, or can be used to [authenticate to the API over Basic Authentication](https://docs.github.com/v3/auth/#basic-authentication).

bryce-carson:
> @bryce-carson This all sounds very reasonable to me.
> 
> ---
> 
> > You may use OAuth tokens or SSH to authenticate with Github, 
> 
> One minor technical comment I have is that these are actually not OAuth tokens but they're called "Personal access tokens". The explanation at the bottom of the https://github.com/settings/tokens site makes is very precise here:
> 
> > Personal access tokens function like ordinary OAuth access tokens. They can be used instead of a password for Git over HTTPS, or can be used to [authenticate to the API over Basic Authentication](https://docs.github.com/v3/auth/#basic-authentication). 

What about this bit?

> Via OAuth and personal access tokens
>
> We recommend you use OAuth tokens to authenticate to the GitHub API. OAuth tokens include personal access tokens and enable the user to revoke access at any time.

real-or-random:
Hehe yeah, it seems even GitHub is not entirely consistent about the wording.

I So my thinking is that we should talk only about PATs and not about OAuth tokens. PATs are really for uses like bash or your script, whereas OAuth is intended for other web apps, so it's not relevant to users here. 

This is line with the overview in https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/authorizing-oauth-apps .

bryce-carson:
> Hehe yeah, it seems even GitHub is not entirely consistent about the wording.
> 
> I So my thinking is that we should talk only about PATs and not about OAuth tokens. PATs are really for uses like bash or your script, whereas OAuth is intended for other web apps, so it's not relevant to users here. 
> 
> This is line with the overview in https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/about-authentication-to-github 
> 
> (edited to correct link, I had first pasted the wrong URL)

Okay, sounds good.

I'll work on the documentation this week or weekend.

layers/+source-control/git/README.org
lebensterben: Yes. Let's don't touch ~/.ssh/config and let user edit it manually.

bryce-carson:
I don't know when I'll get back to this, but when I do I will squash the commits and make things much cleaner.

smile13241324:
Given that this ticket is open for 2 years now I will close it now.

bryce-carson:
> Given that this ticket is open for 2 years now I will close it now.

That's appropriate. Thanks for closing this and the other ticket, Smile.

PR #15403, opened 2022-03-11
Add key binding for revert-buffer in process-menu-mode

bcc32:
This PR just binds `gr` to revert-buffer in process-menu-mode.  I bound it to `gr` rather than `g` so that `gg` still works.

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/packages.el
+++ b/layers/+spacemacs/spacemacs-defaults/packages.el
@@ -393,7 +393,9 @@
   (spacemacs|hide-lighter page-break-lines-mode))
 
 (defun spacemacs-defaults/init-process-menu ()
-  (evilified-state-evilify process-menu-mode process-menu-mode-map
+  (evilified-state-evilify-map process-menu-mode-map
+    :mode process-menu-mode
+    :bindings
     "gr" 'revert-buffer))
 
 (defun spacemacs-defaults/init-quickrun ()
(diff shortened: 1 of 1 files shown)


lebensterben:
I need to check whether "gr" conflicts with other bindings.

At least not long ago "gr" was intended for finding references of current buffer. But that feature is removed.

Also, what's the rationale for "gr"?

bcc32:
It doesn't seem to conflict in process-menu-mode, at least.  `gr` is used in at least a few other places to mean something like "refresh this view", e.g., in magit-status-mode, elfeed-search-mode, notmuch-search-mode, etc.

layers/+spacemacs/spacemacs-defaults/packages.el
```diff
@@ -393,7 +393,8 @@
   (spacemacs|hide-lighter page-break-lines-mode))
 
 (defun spacemacs-defaults/init-process-menu ()
-  (evilified-state-evilify process-menu-mode process-menu-mode-map))
+  (evilified-state-evilify process-menu-mode process-menu-mode-map
```
lebensterben, line 396: `evilified-state-evilify` is not recommended.
Use `evilified-state-evilify-map` instead.

lebensterben (changes requested):

layers/+spacemacs/spacemacs-defaults/packages.el
bcc32: Do you want me to make the change in this PR?  I'm not familiar enough with the two macros to know the difference immediately, so I just made the minimal change (the existing code uses `evilified-state-evilify`).  Feel free to suggest changes or add commits as you see fit.

lebensterben (approved):

lebensterben:
merged. thanks for your contribution.

PR #15410, opened 2022-03-16
[completion] Fix readme markup.

pataquets:
Just found an error while reading the docs. A rogue `==` was breaking the markup.


lebensterben:
thanks

Issue #15089, opened 2021-10-10
Magit rebase todo keybinding not working from normal mode


muhifauzan:
#### Description :octocat:
`git-rebase-todo` default keybinding not working

#### Reproduction guide :beetle:
- Start Emacs
- Change directory to working project
- `SPC g s`
- `l l`
- Choose commit to rebase
- `r i`

*Observed behaviour:* :eyes: :broken_heart:
Usual `git-rebase-todo` buffer keybinding not working from `normal` mode. For example, picking commit, dropping commit, moving commit, etc. The keybinding somehow triggered from `insert` mode instead thus breaking the `insert` mode because the key input was treated as commands instead of regular input.

*Expected behaviour:* :heart: :smile:
Usual `git-rebase-todo` buffer keybinding working from `normal` mode.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 27.2
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 63056ecb5)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(auto-completion emacs-lisp git helm lsp markdown multiple-cursors spell-checking syntax-checking version-control treemacs)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG CAIRO SOUND GPM DBUS GSETTINGS GLIB NOTIFY INOTIFY ACL GNUTLS LIBXML2 FREETYPE HARFBUZZ M17N_FLT LIBOTF ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XDBE XIM MODULES THREADS LIBSYSTEMD JSON PDUMPER LCMS2 GMP

#### Backtrace :paw_prints:
```
```

arifer612:
I'm also facing the same problem but seeing how there hasn't been a commit editing any of this functionality on Spacemacs, it's probably an issue that arose somewhere upstream.

lebensterben:
Similarly, if you've both `git` and `helm` layers enabled,
`helm-ls-git-rebase-*` commands are defined here:
https://github.com/emacs-helm/helm-ls-git/blob/ae2202fbbbe11873ad4b393a6959da50ac250c1e/helm-ls-git.el#L1574-L1581

So for example, `(helpful-function 'helm-ls-git-rebase-pick)` shows:
>Key Bindings
helm-ls-git-rebase-todo-mode-map p

lebensterben:
Try
[...]

PR #15411, opened 2022-03-16
[helm] fix typo in helm-ls-git requires

dankessler:
helm-ls-git is specified as a package belonging to the helm layer, but it's
declared with the (non-existent) keyword :require, which should likely be
:requires. Changing it to :requires means that helm-ls-git will be installed
only if the git *package* is installed. It seems possible the original author's
intent was to install helm-ls-git only if the git *layer* was installed, but
it's hard to know for sure. As far as I know, no current spacemacs layers
install the git package, so fixing this typo will effectively remove helm-ls-git
except for cases where the git package is somehow installed (either explicitly
by the user, or through a private layer, etc).

As of writing, helm-ls-git is currently interfering with some git-rebase
keymaps (see #15089 for discussion), so fixing this typo *should* fix those
issues for most users, although it'll fix it essentially by nerfing helm-ls-git.

### DIFF
--- a/layers/+completion/helm/packages.el
+++ b/layers/+completion/helm/packages.el
@@ -30,7 +30,7 @@
         helm-ag
         helm-descbinds
         helm-flx
-        (helm-ls-git :require git)
+        (helm-ls-git :toggle (configuration-layer/layer-used-p 'git))
         helm-make
         helm-mode-manager
         helm-org


lebensterben:
I believe actually the intention is to install it when `git` layer is in used.

So

(helm-ls-git :toggle (configuration-layer/layer-used-p 'git))

is likely to be the correct one.

dankessler:
I agree that was likely the intention but had forgotten about `:toggle`; I'll amend the PR accordingly (in which case it will no longer really affect #15089)

lebensterben:
@dankessler 
If it's working correctly I will merge it. (I assume it is but I cannot test it now)

dankessler:
Yup, with this edit in place, I tried commenting out my `git` layer, and `helm-ls-git` was removed (along with lots of other things, and `SPC h d P helm-ls-git` correctly reports that it is disabled by toggle), and then re-enabling my `git` layer brought `helm-ls-git` back

PR #15413, opened 2022-03-17
Fix variable name in layers/+os/nixos/README

avdv:
I simply copied the setting form the README to my config and wondered why nothing would happen... :-)


lebensterben:
Thanks.

Issue #9850, opened 2017-11-10
evil-collection: Join forces for a universal Evil binding package


Ambrevar:
A new Evil-binding project was recently started:
[evil-collection](https://github.com/jojojames/evil-collection).
The stated goal is obvious:

> Bring universal, ubiquitous and consistent Evil bindings to Emacs.

Spacemacs has undertaken tremendous work in that area.  Unfortunately,
non-Spacemacs users cannot easily benefit of this work as of now.  This leads to
an Emacs community in a badly fragmented state, with every one here and there
re-implementing her own set of bindings.

Starting evil-collection from scratch and without coordination with the existing
effort (the other evil-* packages as well as Doom config and the hugely popular
Spacemacs, among others) would lead to duplicate efforts and, much more
importantly, an unavoidable divergeance in the bindings.  This would lead to a
lot of friction, say, when moving from Emacs to Spacemacs and vice-versa.
That would not exactly solve the problem of fragmentation...

Some divergeance is probably unavoidable, but we could limit it to a minimum:
While Spacemacs has some of its own specifics in terms of bindings, I believe it
would not be unreasonable to base itself on a universal package while adding on
top a thin layer of its own.

The invaluable knowledge base and community of Spacemacs makes it a key actor
into establishing guidelines and a rationale for a universal package.  I might
be opening a Pandora box here.... Hopefully not, and a joint effort will make
both ends meet to create a universal, well-established package.

quicknir:
I think co-operation here would be a huge benefit for everyone; non spacemacs users would benefit from some current spacemacs functionality entering packages, and I think Ambrevar will help coordinate improvements to evilification that will in turn benefit spacemacs users. Win-win!

CeleritasCelery:
My two cents:
I am afraid that this will suffer from the same problem that @Fuco1 is discussing in #9838. Namely that when a Spacemacs user finds a mode-specific config (for evil, smartparens, etc) they are going to add it to Spacemacs by default. I am willing to bet most of the time they don't even think of contacting the upstream package maintainer. So it seems this process (fix upstream rather then hack spacemacs) is not going to work well unless @syl20bnr makes that part of the official "Spacemacs" policy and it is documented. 

Contributors could try submitting pull requests to remove stuff from Spacemacs and pull requests to add stuff to upstream packages, but that feels like a losing battle unless the collaborators are behind this.

Ambrevar:
@CeleritasCelery: Completely agree.

[...]

PR #15414, opened 2022-03-17
Improve Docker layer 'dockerfile-mode' major mode keybindings

pataquets:
After several weeks using the layer for building Dockerfiles, I've came up with a few enhancements:
* I've flattened the `, c b` and `, c B` keys for build commands to just `, b` and `, B` to save keystrokes. I'm hesitant to remove existing ones because of annoying existing users, but feels redundant. Feedback on this is needed.
* Since [Silex/docker.el](https://github.com/Silex/docker.el)'s package is also installed for managing the `docker` daemon, I've added some command shortcuts for convenience (in addition to the existing `SPC a t d ...` ones):
  * `p` for `docker ps`, which can be used to delete failed builds' containers.
  * `i` for `docker images`, to delete old image builds.
  * `d` to access all the `docker` package commands (incl. `images` and `ps`): `docker {volumes,networks,...}`. Doing `, d` is a bit faster than existing `SPC a t d`.

As for the `docker` package commands, I've wrapped them inside a `(with-eval-after-load)` in case the user disables/excludes `docker` package. Not sure which is better, as opposed to `(feature-p)`, still newbie. Advice much appreciated, as I learn.

### DIFF
--- a/layers/+tools/docker/packages.el
+++ b/layers/+tools/docker/packages.el
@@ -50,12 +50,9 @@
     :defer t
     :init (add-hook 'dockerfile-mode-local-vars-hook #'spacemacs//docker-dockerfile-setup-backend)
     :config
-    (spacemacs/declare-prefix-for-mode 'dockerfile-mode "mc" "compile")
     (spacemacs/set-leader-keys-for-major-mode 'dockerfile-mode
       "b" 'dockerfile-build-buffer
       "B" 'dockerfile-build-buffer-no-cache-buffer
-      "cb" 'dockerfile-build-buffer
-      "cB" 'dockerfile-build-no-cache-buffer)
     (with-eval-after-load 'docker
       (spacemacs/set-leader-keys-for-major-mode 'dockerfile-mode
         "d" 'docker
--- a/layers/+tools/docker/packages.el
+++ b/layers/+tools/docker/packages.el
@@ -52,8 +52,15 @@
     :config
     (spacemacs/declare-prefix-for-mode 'dockerfile-mode "mc" "compile")
     (spacemacs/set-leader-keys-for-major-mode 'dockerfile-mode
+      "b" 'dockerfile-build-buffer
+      "B" 'dockerfile-build-buffer-no-cache-buffer
       "cb" 'dockerfile-build-buffer
-      "cB" 'dockerfile-build-no-cache-buffer)))
+      "cB" 'dockerfile-build-no-cache-buffer)
+    (with-eval-after-load 'docker
+      (spacemacs/set-leader-keys-for-major-mode 'dockerfile-mode
+        "d" 'docker
+        "i" 'docker-images
+        "p" 'docker-containers))))
 
 (defun docker/post-init-flycheck ()
   (spacemacs/enable-flycheck 'dockerfile-mode))
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1907,6 +1907,9 @@ Other:
 - Fixed: broken package declaration for dockerfile-mode
   (thanks to Maximilian Wolff)
 - Added LSP support
+- Dockerfile mode keybindings (thanks to Alfonso Montero):
+  - Flatten ~cb~ and ~cB~ build commands into ~b~ and ~B~ to save keystrokes.
+  - Add keybindings to some =docker= package commands for convenience.
 **** Dotnet
 - Key bindings:
   - Added key bindings for =dotnet= (thanks to Jordan Kaye):
--- a/layers/+completion/helm/README.org
+++ b/layers/+completion/helm/README.org
@@ -63,7 +63,7 @@ listed last.
 
 ** Alternative layers
 Ivy layer is a replacement layer for Helm. When you add =ivy= to the existing
-== list in your dotfile, it will completely replace the =helm= layer.
+list in your dotfile, it will completely replace the =helm= layer.
 
 To switch from Ivy to Helm, modify your =~/.spacemacs=. You will need to add
 =helm= to the existing =dotspacemacs-configuration-layers= list in this file,


layers/+tools/docker/packages.el
```diff
@@ -52,8 +52,15 @@
...
+      "b" 'dockerfile-build-buffer
+      "B" 'dockerfile-build-buffer-no-cache-buffer
       "cb" 'dockerfile-build-buffer
-      "cB" 'dockerfile-build-no-cache-buffer)))
+      "cB" 'dockerfile-build-no-cache-buffer)
+    (with-eval-after-load 'docker
+      (spacemacs/set-leader-keys-for-major-mode 'dockerfile-mode
+        "d" 'docker
+        "i" 'docker-images
+        "p" 'docker-containers))))
```
arifer612, line 63: By wrapping them in `with-eval-after-load`, these keybindings will not be accessible if `docker` is not already loaded beforehand. There may be use-cases where one spins up a new Emacs instance just to edit a Dockerfile without loading the `docker` package. I suggest that you move these under `:init` since they are autoloaded by the `docker` package anyways.

arifer612 (changes requested):
I also agree that having the extra keystroke is just wasted effort so I feel that this is a very reasonable change to make. However, there are a few changes you need to make regarding the first point on redundant keybindings. As for the second point, it may better serve being placed under `:init`. I have left some comments on the both of these points.

Besides `packages.el` you will also need to update the keybindings in the README for this layer.

thanhvg:
I my fork I reverted the following two commits which mess up special mode.

- This reverts commit 8c931b9c8648691c064e72966cb3ffb6308e9731. 

https://github.com/syl20bnr/spacemacs/pull/15050
https://github.com/syl20bnr/spacemacs/pull/14995

- This reverts commit b1fe17d370d5a3f3b7c8d69b51f7b12b77368418.

After that enable evil-collection for docker, remove evil-define-key ceremony, then enjoy all the bindings.

lebensterben:
@dalanicolai 
Please see comment above. I intend to revert those two commits. We should instead evilify specific major modes directly.

dalanicolai:
@lebensterben That second commit is a correction/improvement on that first commit. So now special-mode buffers open in motion state instead of evilified state. It still seems to me that motion state is a sensible default for special-mode buffers, because special-mode buffers are read-only and therefore it is useless to have a normal/insert state there (and motions-state specially exists for read-only buffers). I would probably prefer, to take/copy the useful keybindings from the docker layer (or add a regexp to 'evil-buffer-regexps` as I described in [spacemacs conventions](evil-buffer-regexps)). I would not say that they messes up special mode buffers. For me it is fine whatever you choose, but I think it would make sense to 'verbalize' a good argument for it (something like 'we prefer to use emacs-state bindings in special-mode buffers, instead of evil style navigation', which is more or less what it is about, or 'we prefer to use evil collection, i.e. without major-mode leader key, over creating our own configuration')

lebensterben:
@dalanicolai 

okay let me think more about it.

thanhvg:
I don't want to reinvent the wheel. I want to get benefit from upstream and give them some contributions too.

normal-mode or motion-mode is the same to me. But normal mode is what upstream evil or evil-collection adds extra bindings. Switching to motion-mode as default causes all these free bindings gone and you have this PR trying to add them back. Docker mode uses transient and if you need help, just press `?` then it shows all the bindings, but because you are in motion-mode, `?` will do a evil search. I also noticed problems with ivy buffers because of this motion-mode behavior.

And please don't get me started with `evil-evilified-state`. I despise it, why Spacemacs has its own evil state thing? why not contributing to upstream instead. It drags us down.

dalanicolai:
I would prefer to benefit and add to upstream also, but it is not Spacemacs that reinvented the wheel (AFAIK it invented 'the wheel'). There are more or less no modes that come with 'free' bindings, they always have to get configured (evil-collection is not used automatically by most packages, maybe it should, but it is not). Also, if someone/some layer wants to use evil-collection then there is really no need to revert this PR... just activate evil collection (e.g. try calendar-mode, run `evil-collection-calendar-setup`, and you will have evil collection keybindings. On the other hand, make it open in `normal-state` and then try to exit with `q`, you will be recording a macro). Evilifying states historically was a good option to quickly 'evilify' some state, in a way it is an improvement to motion state, because it does not break too many default Emacs keybindings, and otherwise it modifies them in a consistent way. But indeed it is only helpful as long as no 'better' bindings exist, so indeed now it might make more sense to use evil-collection for such modes ([this](https://github.com/syl20bnr/spacemacs/issues/9850#issuecomment-402566394) is what @syl20bnr had to say about this, and in [this](https://github.com/noctuid/evil-guide#global-keybindings-and-evil-states) section you can read what noctuid has to say about motion state, where I think his complaints are not really big issues (he just prefers evil-collection)). Anyway, as @thanhvg mentioned there is not much difference between motion and normal state, except that in normal state you can not exit with `q` (in that sense it is actually normal state that is messing up special modes). Anyway, AFAIK there is no need to revert those PR's to make possible to use evil-collection, so I would say you are handling a non-issue, and it would probably be a regression. I am not against using evil-collection, but it is perfectly possible without reverting this PR (@thanhvg points in his comments seem not very much related to these PR's).

layers/+tools/docker/packages.el
pataquets: A couple questions:
Independent of `docker` package being not yet loaded at some point, wouldn't the keybindings be accessible anyway as soon as the package is gets to load? (Thus effectively only delaying the key bindings until the last moment, but never skipping them)
What if the user excludes the `docker` package via `dotspacemacs-excluded-packages`? (This is the main corner case that made me wrap everything in `with-eval-after-load`).

smile13241324 (approved):
This is looks ok, though we need to add conditions to only bind the short commands if not lsp is used. In this specific case we still need to use the long version. I will do the necessary adaption after the merge.

layers/+tools/docker/packages.el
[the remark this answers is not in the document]
pataquets: I wasn't aware of this ([CONVENTIONS.org](https://github.com/syl20bnr/spacemacs/blob/develop/doc/CONVENTIONS.org#use-package)). Fixed on #15707.

pataquets: Finally learnt how to do it properly. Fixed on #15707.

pataquets:
Submitted #15707 to address requested changes. Sorry for not doing it here. I was awaiting for the conversation to settle, since it was way above my head to understand it (I even thought it was an unrelated cross-post!). Then the PR was unexpectedly merged.
So much to learn still...

PR #15418, opened 2022-03-21
Refactor core-load-paths to use an in-line function

bryce-carson:
- Remove the two functions that add directories to the load path.
- Add a lambda funcall to the dolist and remove the call to the first of the two functions that this commit would remove.

### DIFF
--- a/core/core-load-paths.el
+++ b/core/core-load-paths.el
@@ -115,7 +115,7 @@
 
 ;;;; Load Paths
 (dolist (subdirectory '(nil "libs/" "libs/spacemacs-theme/" "libs/forks/"))
-  (let (path (concat spacemacs-core-directory subdirectory))
+  (let ((path (concat spacemacs-core-directory subdirectory)))
     (if (file-exists-p path)
        (add-to-list 'load-path path)
      (error "The directory %s does not exist and cannot be added to the `load-path'." path))))
--- a/core/core-load-paths.el
+++ b/core/core-load-paths.el
@@ -114,20 +114,11 @@
 (setq pcache-directory (concat spacemacs-cache-directory "pcache/"))
 
 ;;;; Load Paths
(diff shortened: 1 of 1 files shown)


lebensterben:
- indentation isn't correct.
- no need to (funcall (lambda () ...). just modifying the load path suffices.

bryce-carson:
@lebensterben is the PR ready now with 91a8938?

It wasn't *necessary*, but I thought making the variable names more explicit would be good to do as well.

bryce-carson:
Wait... the `let` form is entirely unnecessary. One last commit then I'll leave it up to you.

core/core-load-paths.el
```diff
@@ -114,20 +114,11 @@
...
-  (when (file-exists-p dir)
-    (spacemacs//add-to-load-path dir)))
-
-(dolist (suffix '(nil "libs/" "libs/spacemacs-theme/" "libs/forks"))
-  (spacemacs//add-to-load-path (concat spacemacs-core-directory suffix)))
+(dolist (subdirectory '(nil "libs/" "libs/spacemacs-theme/" "libs/forks/"))
+  (let ((path (concat spacemacs-core-directory subdirectory)))
+    (if (file-exists-p path)
+       (add-to-list 'load-path path)
+     (error "The directory %s does not exist and cannot be added to the `load-path'." path))))
```
dalanicolai, line 121: I guess, as previously, the 'error' alternative wasn't there, it is better to use `display-warning`  (or the shorter `lwarn`) here

lebensterben: It's okay to use `error` here, because if any of the load paths does not exist, something is really wrong.

lebensterben: It's okay to use `error` here, because if any of the load paths does not exist, something is really wrong.

smile13241324 (approved):

smile13241324:
Looks good, still I want to do a quick test when I am back at home before merging.

smile13241324:

PR #15424, opened 2022-03-24
Add layer `translate` for doing translation jobs

rayw000:
New layer `translate` for doing article translation jobs.

### DIFF
--- a/layers/+tools/translate/README.org
+++ b/layers/+tools/translate/README.org
@@ -1,0 +1,82 @@
+#+TITLE: Translate Layer
+
+#+TAGS: layer|tool
+
+[[file:img/screen-record.gif]]
+
+* Table of Contents                     :TOC_5_gh:noexport:
+- [[#description][Description]]
+  - [[#features][Features:]]
+- [[#install][Install]]
+- [[#usage][Usage]]
+- [[#configuration][Configuration]]
+  - [[#language-pair][Language Pair]]
+  - [[#highlighting][Highlighting]]
+  - [[#read-only][Read-only]]
+  - [[#face][Face]]
+- [[#key-bindings][Key bindings]]
+
+* Description
+This layer is designed for Paragraph-oriented minor mode for
+side-by-side document translation workflow.
+
+** Features:
+- Paragraph-oriented side-by-side document translation workflow
+- Integrate word/paragraph online translation
+
+* Install
+To use this configuration layer, add it to your =~/.spacemacs=. You will need to
+add =translate= to the existing =dotspacemacs-configuration-layers= list in this file.
+
+* Usage
+It's quite simple to use this layer. Open the translation file you are currently working
+on, and call command =translate-open-reference-file= or =translate-select-reference-buffer=
+to set the reference buffer or open an existed reference file. Now you can just keep your
+cursor in your translation buffer. The cursor in the reference buffer will be kept in
+sync with the cursor in the translation buffer.
+
+* Configuration
+All layer configurations can be done by setting layer variables in your dotfile.
+No custom user config lines are necessary. For more details please see the homepage
+of package [[https://github.com/rayw000/translate-mode][translate-mode]] and [[https://github.com/lorniu/go-translate/][go-translate]].
+
+** Language Pair
+You need to set language pairs to make online translation work.
+
+#+BEGIN_SRC emacs-lisp
+  (translate :variables gts-translate-list '(("en" "zh") ("en" "fr")))
+#+END_SRC
+
+** Highlighting
+Set =translate-enable-highlight= to =nil= to disable highlighting.
+
+#+BEGIN_SRC emacs-lisp
+  (translate :variables translate-enable-highlight nil)
+#+END_SRC
+
+** Read-only
+You can set =translate-reference-buffer-read-only= to =t= to make the reference buffer read-only.
+
+#+BEGIN_SRC emacs-lisp
+  (translate :variables translate-reference-buffer-read-only t)
+#+END_SRC
+
+** Face
+The face of highlight paragraph can be customized by =translate-paragraph-highlight-face=.
+Put this following line into =custom-set-faces= in your =~/.spacemacs~ file to set the background
+color to red, for example.
+
+#+BEGIN_SRC emacs-lisp
+'(translate-paragraph-highlight ((t (:extend t :background "red"))))
+#+END_SRC
+
+* Key bindings
+
+| Key binding   | Description                                                                  |
+|---------------+------------------------------------------------------------------------------|
+| ~SPC a t T t~ | Toggle =translate-mode=                                                      |
+| ~SPC a t T p~ | Open a translation list buffer for current paragraph in the reference buffer |
+| ~SPC a t T w~ | Popup translation list for word at point, also can be used on regions        |
+| ~SPC a t T f~ | Prompt to open the reference file                                            |
+| ~SPC a t T b~ | Prompt to select a buffer and set it as the reference buffer                 |
+| ~SPC a t T h~ | Toggle paragraph highlighting                                                |
--- a/layers/+tools/translate/funcs.el
+++ b/layers/+tools/translate/funcs.el
@@ -1,0 +1,42 @@
+;;; funcs.el --- Semantic Layer functions File for Spacemacs
+;;
+;; Copyright (c) 2012-2021 Sylvain Benner & Contributors
+;;
+;; Author: Ray Wang <rayw.public@gmail.com>
+;; URL: https://github.com/syl20bnr/spacemacs
+;;
+;; This file is not part of GNU Emacs.
+;;
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+;;
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+;;
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+(when (and (configuration-layer/package-used-p 'translate-mode)
+           (configuration-layer/package-used-p 'go-translate))
+
+  (defun translate/translate-current-reference-paragraph ()
+    "Show all available translations of the reference paragraph at point in a pop-up frame."
+    (interactive)
+    (gts-translate translate//paragraph-translator))
+
+  (defun translate/translate-word-at-point ()
+    "Pop-up translations of the word at point."
+    (interactive)
+    (gts-translate translate//word-translator))
+
+  (defun translate//set-translate-mode-paragraph-functions ()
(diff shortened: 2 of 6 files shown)


layers/+tools/translate/README.org
```diff
@@ -0,0 +1,64 @@
...
+* Install
+To use this configuration layer, add it to your =~/.spacemacs=. You will need to
+add =translate= to the existing =dotspacemacs-configuration-layers= list in this file.
+
+* Usage
+It's quite simple to use this layer. Open the translation file you are currently working
+on, and call command =translate-open-original-file= or =translate-select-original-buffer=
+to set the original article for referring. Now you can just keep your cursor in your
+translation buffer. The cursor in the original buffer will keep syncing with your cursor
+moving.
```
arifer612, line 35: This sentence sounds a little awkward though it gets the point across. I would have written it as "The cursor in the original buffer will be kept in sync with the cursor in the translation buffer" or something like that.

layers/+tools/translate/README.org
```diff
@@ -0,0 +1,64 @@
...
+
+** Read-only
+You can set =translate-original-buffer-read-only= to =t= to make original buffer read-only.
+
+#+BEGIN_SRC emacs-lisp
+  (translate :variables translate-original-buffer-read-only t)
+#+END_SRC
+
+** Face
+The face of highlight paragraph can be customized by =translate-paragraph-highlight-face=.
```
arifer612, line 64: I would suggest adding an example source block like the other subsections above

layers/+tools/translate/packages.el
```diff
@@ -0,0 +1,80 @@
...
+  (use-package translate-mode
+    :demand t
+    :hook (translate-mode . set-transalte-mode-paragraph-functions)))
+
+(defun translate/init-go-translate ()
+  (use-package go-translate
+    :demand t
+    :config
+    (progn
+      (setq gts-translate-list '(("en" "zh")))
```
arifer612, line 42: The language pair should not be hard-coded into the layer since not everybody will need to use it. You have already mentioned in the README that `gts-translate-list` has to be specified as layer variables, so let users do so there.

layers/+tools/translate/funcs.el
```diff
@@ -0,0 +1,42 @@
...
+    "Show all available translations of the original paragraph at point in a pop-up frame."
+    (interactive)
+    (gts-translate paragraph-translator))
+
+  (defun spacemacs/translate-word-at-point ()
+    "Pop-up translations of the word at point."
+    (interactive)
+    (gts-translate word-translator))
+
+  (defun set-transalte-mode-paragraph-functions ()
```
arifer612, line 36: A typo in `translate`?

arifer612 (changes requested):
This is a very sweet layer! Wish I had it back when I was actively translating stuff. I notice that you have not suggested any Spacemacs keybindings for this layer, it would be better if there are some `evil'-like bindings for the interactive functions.

arifer612:
Spacemacs keybindings are mostly vim-like and follow the [guidelines from emacs-evil](https://github.com/emacs-evil/evil-collection#guidelines). Each major mode gets its own prefix, usually `SPC m` or equivalently `,` and you can define the keybindings from that using `spacemacs/set-leader-keys-for-major-mode`. This is great since your layer provides a new major-mode. 

I suggest that you have a look at the following example from the `transmission` layer:
https://github.com/syl20bnr/spacemacs/blob/18ad9759b29a56ba92fe452b62dd545329175b63/layers/%2Btools/transmission/packages.el#L27-L60

##### A brief explanation
`spacemacs/set-leader-keys` is a macro that sets a global keybinding to an interactive function. You can maybe assign `translate-open-original-file` or `translate-select-original-buffer` to something like `atT` for applications -> tools -> (T)ranslate. I'm not sure if `T` is still available but a quick check of all the layers in tools starting with 't' does not suggest that it is being used.

`spacemacs/declare-prefix-for-mode` is a macro that adds like a tree-branch structure to your bindings for the major mode. Bindings to interactive functions do not go here.

`spacemacs/set-leader-keys-for-major-mode` is a macro that sets bindings for interactive functions after the major-mode prefix `SPC m`.

rayw000:
Great suggestion!

This layer doesn't provide a new `major-mode`, that's why I was not sure which prefix should the new keybindings follow. But I like the idea of `SPC a t T`. I've checked my keybindings by seeing the suggestions from `which-key`, `SPC a t T` is not being used too. So, let's put it into `SPC a t`, i.e. application -> tools -> (T)ranslate.

Many thanks! @arifer612

layers/+tools/translate/packages.el
```diff
@@ -0,0 +1,79 @@
...
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+(defconst translate-packages
+  '(
+    (translate-mode :location (recipe :fetcher github :repo "rayw000/translate-mode"))
+    go-translate
+    popup
+    posframe
+    ))
```
lebensterben, line 29: trailing parentheses should not take a new line

layers/+tools/translate/packages.el
```diff
@@ -0,0 +1,79 @@
...
+         :render (gts-posframe-pop-render))
+        "Paragraph translator for `go-translate'.")
+      (defvar word-translator
+        (gts-translator
+         :picker (gts-noprompt-picker)
+         :engines (list (gts-google-engine) (gts-google-rpc-engine) (gts-bing-engine))
+         :render (gts-posframe-pop-render))
+        "Word translator for `go-translate'."))))
+
+(defun translate/init-popup ()
```
lebensterben, line 73: is popup or posframe used in other layer?

layers/+tools/translate/packages.el
```diff
@@ -0,0 +1,79 @@
...
+  (use-package translate-mode
+    :demand t
+    :hook (translate-mode . set-translate-mode-paragraph-functions)))
+
+(defun translate/init-go-translate ()
+  (use-package go-translate
+    :demand t
+    :config
+    (progn
+      (defclass spacemacs//original-paragraph-texter (gts-texter) ())
```
lebensterben, line 42: those new classes and methods should be moved to funcs.el

layers/+tools/translate/packages.el
```diff
@@ -0,0 +1,79 @@
...
+        (translate-get-original-paragraph-text-at-point))
+      (defclass spacemacs//original-paragraph-picker (gts-picker)
+        ((texter :initarg :texter :initform (spacemacs//original-paragraph-texter))))
+      (cl-defmethod gts-pick ((o spacemacs//original-paragraph-picker))
+        (let ((text (gts-text (oref o texter))))
+          (when (= 0 (length (if text (string-trim text) "")))
+            (user-error "Make sure there is any word at point, or selection exists"))
+          (let ((path (gts-path o text)))
+            (cl-values text path))))
+      ;; TODO Popup-el based render
```
lebensterben, line 53: either remove it or implement it

layers/+tools/translate/packages.el
```diff
@@ -0,0 +1,79 @@
...
+      ;; (defclass spacemacs//inplace-render (gts-render) ())
+      ;; (cl-defmethod gts-out ((_ spacemacs//inplace-render) task)
+      ;;   (with-slots (ecode result) task
+      ;;     (if ecode
+      ;;         (user-error "%s" result)
+      ;;       (popup-menu* (split-string result "\n")))))
+      (defvar paragraph-translator
```
lebensterben, line 60: if these two variables are not mutable, defconst are preferred.

also they should go to config.el, with correct naming.

layers/+tools/translate/funcs.el
```diff
@@ -0,0 +1,42 @@
...
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+;;
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+(when (and (configuration-layer/package-used-p 'translate-mode)
+           (configuration-layer/package-used-p 'go-translate))
+
+  (defun spacemacs/translate-current-original-paragraph ()
```
lebensterben, line 26: naming convention is layer/xxx or layer//xxx. so use translate/xx instead

layers/+tools/translate/packages.el
arifer612: A quick search tells me that `posframe` is used in the `auto-completion` and `dap` layers; `popup` in `spacemacs-purpose` layer.

arifer612:
> This layer doesn't provide a new `major-mode`

Ah right! I misread your package's header. If it's a minor mode then you just need to change the macro to `spacemacs/set-leader-keys-for-minor-mode`. The process is the same as before, but now you have to make sure that your keybindings do not overlap with the major-mode's. For both Org-mode and Markdown-mode files, you can probably file the translate-mode bindings under the prefix `x` which is meant for text manipulation, unless you want to use an alternative prefix that is free!

layers/+tools/translate/packages.el
lebensterben: yes this is what I think of.

Each package should have one and only one "owner" layer. It means that each package can only have one "init-PACKAGE" function.

arifer612: That is reasonable. For `popup` it can be removed since it is loaded up by `spacemacs-visual` and `spacemacs-purpose`. However, `posframe` would then have to be added to one of the `spacemacs-` layers.

lebensterben: it should then also be moved to spacemacs-visual layer, with a minimal configuration, like (use-package posframe :defer t)

then for various layers that want to integrate with posframe, they should add post-config hooks.

layers/+tools/translate/packages.el
rayw000: I've tried before, but it is weird that my `funcs.el` is always loaded before `go-translate` and `translate-mode` packages are ready, differing from described [here](https://develop.spacemacs.org/doc/LAYERS.html#funcsel). Checking `(featurep 'go-translate)` at the beginning of the funcs.el always gets `nil`. That's why I put the class/method definitions into `:config` in `(use-package go-translate)`. Same thing happens in `cmake` layer too. Do you have any idea?

rayw000:
Meaningful key prefixes such as `t` and `T` are all taken by `markdown-mode` or `org-mode`, so I added a new prefix `atT` named `translate`, and put all the keybindings here, as a collection of tools. For example, `translate/translate-word-at-point` is bound to `SPC a t T w`.

layers/+tools/translate/packages.el
lebensterben: I don't think load order is the issue here.

The issue is that when you define the new classes, they inherit from classes defined in the respective packages. So the new class declarations must be evaluated after the loading of the packages.

Okay just keep them here.

rayw000: Yes, I knew your point, and that's what I did at first.
Perhaps there is a bug to be investigated that the actual load order here is not following its documentation. It may not affect function definitions, like the way many other layers use `funcs.el`, but when inheriting from classes, it fails.

rayw000:
I temporarily added `posframe` into layer package list, and hooked the initialization function of `go-translate` package  on `:post-config` of `posframe`. If #15431 is merged, `posframe` initialization will be removed from this layer.

smile13241324 (approved):
Looks good

PR #15425, opened 2022-03-24
Feature/python layer code cells support

Rathur421:
- Enable [code-cells](https://github.com/emacs-straight/code-cells) support in the python layer using variable `python-use-code-cells` if `ipython-notebook` layer is not used.
- Allow to open ipython-notebook after installing [jupytext](https://github.com/mwouts/jupytext)
- allow working with code cell in a python REPL process


layers/+lang/python/packages.el
```diff
@@ -95,6 +96,22 @@
       (add-to-list 'spacemacs-jump-handlers-python-mode
                    '(anaconda-mode-find-definitions :async t)))))
 
+(defun python/init-code-cells ()
+  (use-package code-cells
+    :if (and (python-use-code-cells)
```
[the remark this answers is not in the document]
Rathur421: I assumed that people using the `ipython-notebook` layer would prefer to use its implementation for notebook, and I was not sure if all people would like to have python-cell activated by default.
But we could totally remove the variable `python-use-code-cell` and only check for the `ipython-notebook` layer.

layers/+lang/python/README.org
```diff
@@ -306,6 +309,14 @@ To use =dap-mode= for debugging do:
   pip install "ptvsd>=4.2"
 #+END_SRC
 
+** Notebook and code cells
+To use =code-cells-mode= with ipython notebook automatically you should install
+[[https://github.com/mwouts/jupytext][jupytext]]. Make sure that the
+=ipython-notebook= layer is not used.
+
+Then you can evaluate the current cell in a REPL process with ~SPC s c~ or all
+the cells above with ~SPC s a~.
```
smile13241324, line 318: These bindings don't look right, they should go under `SPC m s c` and `SPC m s a` at least that's what your code will bind them at.

layers/+lang/python/packages.el
```diff
@@ -95,6 +96,19 @@
...
+    :if (not (configuration-layer/layer-used-p 'ipython-notebook))
+    :defer t
+    :init
+    (progn
+      (add-hook 'python-mode-hook 'code-cells-mode)
+      (spacemacs/set-leader-keys-for-major-mode 'python-mode
+        "gp" 'code-cells-backward-cell
+        "gn" 'code-cells-forward-cell
+        "sc" 'code-cells-eval
+        "sa" 'code-cells-eval-above))))
```
smile13241324, line 110: These bindings should be listed in the README. Also you should check that there are no conflicts with the standard bindings from LSP layer if you bind them unconditionally.

layers/+lang/python/packages.el
```diff
@@ -95,6 +96,19 @@
       (add-to-list 'spacemacs-jump-handlers-python-mode
                    '(anaconda-mode-find-definitions :async t)))))
 
+(defun python/init-code-cells ()
+  (use-package code-cells
+    :if (not (configuration-layer/layer-used-p 'ipython-notebook))
```
smile13241324, line 101: Can you make sure that code-cells is not owned by a different layer already? Packages cannot be owned by multiple layers, if this is the case the package must be moved to one of both and this layer should then depend on the other layer via a layers.el file and configure the package using a pre/post-init hook.

smile13241324 (changes requested):

layers/+lang/python/README.org
Rathur421: Yes, that was my mistake.

layers/+lang/python/packages.el
Rathur421: There doesn't seem to be any package which own code-cells :+1:

layers/+lang/python/packages.el
```diff
@@ -95,6 +96,19 @@
...
+    :defer t
+    :init
+    (progn
+      (add-hook 'python-mode-hook 'code-cells-mode)
+      (spacemacs/set-leader-keys-for-major-mode 'python-mode
```
Rathur421, line 106: Could `spacemacs/set-leader-keys-for-minor-mode 'code-cells-mode` be more appropriate there?

lebensterben: good idea indeed!

Rathur421:
Before, accepting the merge request, could someone confirm that the bindings for lsp is okay. As I mentioned [here](https://github.com/syl20bnr/spacemacs/pull/15425#discussion_r845452533), the bindings doesn't seem to conflict, but if that's okay the documentation for the bindings in the lsp layer should be updated.
I can do it, but I'm relatively new to emacs and lisp and I want to be sure that I didn't miss something.

layers/+lang/python/packages.el
smile13241324: Hmmmmm looks like this Readme is out of date and there are mismatches between xref and peek bindings. The layer must be a revised.

Back to your PR I have quickly checked the bindings and "gn" is already taken when peek bindings replace xref bindings.

"qb" is taken by xref bindings.

I would suggest using "qN" and "qB" these are free for both jump solutions.

smile13241324: Sounds cool, lets go with your bindings 👍

Rathur421:
This should be good to merge

smile13241324 (approved):

PR #15427, opened 2022-03-25
[core] enable lexical binding for all

thanhvg:
For most of the core files the change was just ` -*- lexical-binding: t -*-` added to the top line.

Major and risky changes are in [core/core-configuration-layer.el](https://github.com/syl20bnr/spacemacs/compare/develop...thanhvg:feat/lexical?expand=1#diff-8abc7faad21a170282bd7b305596fb44024d8456e258863d40eea2be0ff6d630) which really depended on dynamic scope. For a basic test I was able to add and remove layers. And Spacemacs runs a bit faster.

I'm not very sure this change is safe, I am using this change on my boxes as a way to verify it. But I would like to push it to dev at some point to crowd source testing this change.

### DIFF
--- a/core/core-command-line.el
+++ b/core/core-command-line.el
@@ -1,4 +1,4 @@
-;;; core-command-line.el --- Spacemacs Core File
+;;; core-command-line.el --- Spacemacs Core File -*- lexical-binding: t -*-
 ;;
 ;; Copyright (c) 2012-2021 Sylvain Benner & Contributors
 ;;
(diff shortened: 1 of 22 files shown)


eklitzke:
I checked out your branch and tested it locally and as far as I can tell everything works, so this is :+1: from me.

Can I suggest adding a linter rule in CI to ensure that new files added to Spacemacs also have lexical-binding enabled? You can probable come up with something better/more portable, but this works with GNU grep and bash:

```bash
#!/bin/bash
#
# Ensure .el files in the core/ directory have lexical-binding enabled.

shopt -s globstar
rc=0
for f in core/**/*.el; do
  if ! grep -q 'lexical-binding: t' "$f"; then
    echo "error: File $f does not have lexical-binding enabled" >&2
    rc=1
  fi
done
exit "$rc"
```

PR #15428, opened 2022-03-26
[compleseus] update vertico-repeat

thanhvg:
Update commands for vertico-repeat


lebensterben:
I've an overhaul of this layer not yet pushed. I will incorporate these changes there.

PR #15431, opened 2022-03-28
move posframe package into spacemacs-visual layer

rayw000:
As a visualization package, `posframe` introduced in `dap` layer will probably be used in other layers (see https://github.com/syl20bnr/spacemacs/pull/15424#discussion_r834933516). Better if  `spacemacs-visual` layer contains this package, and other layers can do initialization stuff in `pre-init-posframe` or `post-init-posframe`.

This PR
1. moves `posframe` package from `dap` layer and `translate` layer into `spacemacs-visual` layer for package sharing
2. adds `pre-int-posframe` and `post-init-posframe` for further configurations in `dap` layer
3. deletes `posframe` dependency from `translate` layer
4. enables `posframe` package in `translate` layer only when it is available

### DIFF
--- a/layers/+tools/dap/packages.el
+++ b/layers/+tools/dap/packages.el
@@ -22,8 +22,7 @@
 
 
 (defconst dap-packages
-  '(dap-mode
-    posframe))
+  '(dap-mode))
 
 (defun dap/init-dap-mode ()
   (use-package dap-mode
@@ -137,6 +136,6 @@
           ;; Set bindings
           (apply #'spacemacs/set-leader-keys-for-major-mode mode bindings))))))
 
+(defun dap/pre-init-posframe ())
 
-(defun dap/init-posframe ()
-  (use-package posframe))
+(defun dap/post-init-posframe ())
--- a/layers/+spacemacs/spacemacs-visual/packages.el
+++ b/layers/+spacemacs/spacemacs-visual/packages.el
@@ -33,6 +33,7 @@
         hl-todo
         popup
         popwin
+        posframe
         (zoom-frm :location local)))
 
 (defun spacemacs-visual/init-all-the-icons ()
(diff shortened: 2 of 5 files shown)


layers/+tools/dap/packages.el
```diff
@@ -137,6 +136,6 @@
           ;; Set bindings
           (apply #'spacemacs/set-leader-keys-for-major-mode mode bindings))))))
 
+(defun dap/pre-init-posframe ())
 
-(defun dap/init-posframe ()
-  (use-package posframe))
+(defun dap/post-init-posframe ())
```
arifer612, line 141: Is there a need for the `post-init`, there wasn't any special configuration for the `dap` layer before this so correct me if I'm wrong but I think just having `pre-init` should be good right?

rayw000: I think you re right. Perhaps there is no need to preserve `pre-init` also? I leave these two functions here just for not surprising the maintainer of `dap` layer. I'm okay to remove `post-init`.

arifer612: That is true, but I think it'll be better to wait for more responses too.

rayw000:
Hi @smile13241324 

Since you approved #15424 a few hours ago, would you please also approve this PR?

Previously this PR moved the dependency package `posframe` from `dap` layer to `spacemacs-visual`. Now the `posframe` requirement in `translate` layer has also been removed.

BTW, I have no idea why I got test failed on `snapshot` operating systems. If there are something I missed, please feel free to let me know. Thank you!

rayw000:
This PR doesn't work well with `spacemacs-base` distribution, I'll do some further work before converting it back from draft.

rayw000:
Now this PR is ready for review. Please take a look. Thank you! @smile13241324

smile13241324 (approved):
Looks good, I will make a local merge and see how it goes.

smile13241324:

PR #15435, opened 2022-03-28
Add grip-mode to markdown layer from github layer

robbyoconnor:
Since the github layer was deprecated in
cb786cc0711e0b23790185ef79012812dcfbe768, add it to the markdown layer.

### DIFF
--- a/layers/+lang/markdown/README.org
+++ b/layers/+lang/markdown/README.org
@@ -9,6 +9,7 @@
   - [[#features][Features:]]
 - [[#bibtex][BibTeX]]
 - [[#install][Install]]
+  - [[#grip-mode][grip-mode]]
 - [[#configuration][Configuration]]
   - [[#live-preview][Live preview]]
   - [[#automatic-mmm-mode-generation][Automatic MMM-Mode Generation]]
@@ -17,6 +18,7 @@
 - [[#key-bindings][Key bindings]]
   - [[#element-insertion][Element insertion]]
   - [[#element-removal][Element removal]]
+  - [[#grip-mode-1][Grip-mode]]
   - [[#table-manipulation][Table manipulation]]
   - [[#completion][Completion]]
   - [[#following-and-jumping][Following and Jumping]]
@@ -32,6 +34,7 @@
 This layer adds markdown support to Spacemacs.
 
 ** Features:
+- [[https://github.com/seagle0128/grip-mode][grip-mode]] Github-flavored Markdown/Org preview using [[https://github.com/joeyespo/grip][Grip]].
 - markdown files support via [[http://jblevins.org/git/markdown-mode.git/][markdown-mode]]
 - [[https://github.com/mdx-js/mdx][mdx]] file support via [[http://jblevins.org/git/markdown-mode.git/][markdown-mode]]
 - Fast GitHub-flavored live preview via [[https://github.com/blak3mill3r/vmd-mode][vmd-mode]]
@@ -47,6 +50,9 @@ For more extensive support of references with BibTeX files, have a look at the
 To use this configuration layer, add it to your =~/.spacemacs=. You will need to
 add =markdown= to the existing =dotspacemacs-configuration-layers= list in this
 file.
+** grip-mode
+Grip-mode [[https://github.com/seagle0128/grip-mode#prerequisite][requires python and the python package grip]] to be installed on the
+system. Grip can usually be installed from [[https://pypi.org/project/grip/][PyPI]].
 
 * Configuration
 ** Live preview
@@ -152,6 +158,10 @@ To generate a table of contents type on top of the buffer:
 | Key binding | Description         |
 |-------------+---------------------|
 | ~SPC m k~   | kill thing at point |
+
+** Grip-mode
+
+Toggle =grip-mode= by doing ~M-x grip-mode~.
 
 ** Table manipulation
 
--- a/layers/+lang/markdown/packages.el
+++ b/layers/+lang/markdown/packages.el
@@ -27,6 +27,7 @@
     company-emoji
     emoji-cheat-sheet-plus
     gh-md
+    grip-mode
     markdown-mode
     markdown-toc
     mmm-mode
@@ -56,6 +57,10 @@
     (dolist (mode markdown--key-bindings-modes)
       (spacemacs/set-leader-keys-for-major-mode mode
         "cr" 'gh-md-render-buffer))))
+
+(defun markdown/init-grip-mode ()
+  (use-package grip-mode
+    :defer t))
 
 (defun markdown/post-init-valign ()
   (add-hook 'markdown-mode-hook 'valign-mode))
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2097,8 +2097,9 @@ Other:
   (thanks to Boris Buliga)
 **** GitHub
 - Layer:
-  - marked as deprecated as functionality is all in =git= layer (thanks
+  - Marked as deprecated as functionality is all in =git= layer (thanks
     to practicalli-john)
+  - Moved =grip-mode= to the =markdown= layer.
 - Packages:
   - Added new packages =forge= (thanks to Miciah Dashiel Butler Masters)
   - Do not install =forge= on Windows by default, see =README.org= of the layer
@@ -2792,6 +2793,7 @@ files (thanks to Daniel Nicolai)
 - Made =*lsp-help*= buffer shown in a popup window using popwin
 - Add support for mdx files via markdown-mode (thanks to robbyoconnor)
 - Improve documentation about markdown preview options (thanks to Lin Sun)
+- Moved =grip-mode= to the =markdown= layer from the =github= layer.
 **** Major modes
 - Added SPARQL-mode (thanks to Dietrich Daroch)
 - Added android logcat (thanks to jiejingzhang)


bryce-carson:
Thank you, @robbyoconnor for the pull request. This looks fine, but I did create some review comments.

I believe this should be placed within the Markdown layer, however, as the functionality provided by Grip is related to GitHub and its specific rendering of (primarily) Markdown in `README.md` files (and optionally `README.org` files). Do you agree, @lebensterben, that this should be added to the Markdown layer instead?

robbyoconnor:
> I believe this should be placed within the Markdown layer, however, as the functionality provided by Grip is related to GitHub and its specific rendering of (primarily) Markdown in `README.md` files (and optionally `README.org` files). Do you agree, @lebensterben, that this should be added to the Markdown layer instead?

No, because this is specific to GitHub technically, which goes with git since GitHub is a forge.

bryce-carson:
> > I believe this should be placed within the Markdown layer, however, as the functionality provided by Grip is related to GitHub and its specific rendering of (primarily) Markdown in `README.md` files (and optionally `README.org` files). Do you agree, @lebensterben, that this should be added to the Markdown layer instead?
> 
> No, because this is specific to GitHub technically, which goes with git since GitHub is a forge.

GitHub _is a forge_, yes. The Git layer adds functionality to work with _any_ forge, not just Github.

The Grip Python package provides functionality specific to GitHub's rendering of (primarily) Markdown, as I already said. If you `C^f` the devel version of the Git layer README you'll notice it makes no reference to markdown, and the only reference to Org is the `orgit` package included in Magit.

In short, it's not related to the `git` command line tool (Git), version control *systems* (SVN, Bazaar, Git, etc.) and should belong to the markdown layer because that is what it is related to.

GitHub != Git layer.

robbyoconnor:
I'll move it.

robbyoconnor:
@bryce-carson moved.

layers/+lang/markdown/README.org
```diff
@@ -153,6 +161,12 @@ To generate a table of contents type on top of the buffer:
 |-------------+---------------------|
 | ~SPC m k~   | kill thing at point |
 
+** Grip-mode
+
+| Key binding | Description                                                 |
+|-------------+-------------------------------------------------------------|
+| ~SPC g p~ | toggle github flavored mb/org buffer preview in web-browser |
+
```
robbyoconnor, line 169: Not sure about this keybinding

bryce-carson:
Once a maintainer responds to the comments I made regarding the keybinding for this, it'll probably be ready to merge.

You can always keep `g p` for your personal config if you prefer. :)

layers/+lang/markdown/packages.el
```diff
@@ -57,6 +58,14 @@
...
         "cr" 'gh-md-render-buffer))))
 
+(defun git/init-grip-mode ()
+  (use-package grip-mode
+    :defer t
+    :init
+    (progn
+      (spacemacs/set-leader-keys
+        "gp" 'grip-mode))))
+
```
robbyoconnor, line 68: Not sure about this keybinding

bryce-carson:
The `safe_join` function in Flask [was deprecated recently](https://github.com/pallets/flask/blob/b109b092a95036de875246f05cdc377a2c7faab8/CHANGES.rst#id5), so the current version of Grip on PyPI is broken (joeyespo/grip@be091045a1c1175850c766b9793edbfddca88afd).

This cannot be merged until it is fixed, as users without existing, outdated Flask installations will find that the package is broken.

robbyoconnor:
@bryce-carson was it deprecated or removed because deprecated doesn't equate to removal.

bryce-carson:
> @bryce-carson was it deprecated or removed because deprecated doesn't equate to removal.

It was removed, pardon me. Previously it *was* deprecated. Now it is gone and functionality in Grip is broken.

robbyoconnor:
Yeah the words matter a lot here! Let's hold off but this can be safely merged... it's functionality that exists already

bryce-carson:
> Yeah the words matter a lot here! Let's hold off but this can be safely merged... it's functionality that exists already

True. Anyone trying the currently deprecated GitHub layer, without already having Flask as a requirement on their system, will find it's broken regardless.

The merge would be good with the suggested keybinding and the fixed init function, @robbyoconnor.

layers/+lang/markdown/README.org
```diff
@@ -47,6 +50,11 @@ For more extensive support of references with BibTeX files, have a look at the
 To use this configuration layer, add it to your =~/.spacemacs=. You will need to
 add =markdown= to the existing =dotspacemacs-configuration-layers= list in this
 file.
+** grip-mode
+Grip-mode [[https://github.com/seagle0128/grip-mode#prerequisite][requires python and the python package grip]] to be installed on the
+system. Grip can usually be installed with the command =pip install grip=,
+on debian based systems make sure you use the python 3 version
```
lebensterben, line 56: No need to include installation instruction for specific distributuon.

layers/+lang/markdown/README.org
```diff
@@ -153,6 +161,12 @@ To generate a table of contents type on top of the buffer:
 |-------------+---------------------|
 | ~SPC m k~   | kill thing at point |
 
+** Grip-mode
+
+| Key binding | Description                                                 |
+|-------------+-------------------------------------------------------------|
+| ~SPC g p~ | toggle github flavored mb/org buffer preview in web-browser |
```
lebensterben, line 168: I don't think this is necessary. Let user call grip-mode via M-x.

lebensterben (changes requested):

lebensterben:
I never thought it's necessary to preview any markup language...

But since some people want it we can add it, but preferably behind a layer config variable.

layers/+lang/markdown/README.org
robbyoconnor: This is literally copy/paste from the `github` layer.

layers/+lang/markdown/README.org
robbyoconnor: I do -- I don't wanna do it -- why not have a leader key

lebensterben:
also, this should be in the markdown layer. because GitHub flavored markdown is widely used, not limiting to GitHub.

robbyoconnor:
@lebensterben 

> also, this should be in the markdown layer. because GitHub flavored markdown is widely used, not limiting to GitHub.

Check again. It is.

layers/+lang/markdown/README.org
bryce-carson: I proposed changes earlier. We may as well rewrite the description while we're moving it from the deprecated layer.

bryce-carson:
@robbyoconnor @lebensterben I didn't realize I had to "finish" a review. I wrote those review comments many hours ago.

They're there and availble now, I guess.

layers/+lang/markdown/README.org
lebensterben: SPC leader key is too precious.

layers/+source-control/git/README.org
```diff
@@ -80,6 +83,13 @@ An SSH key removes the need to provide login details for each request from Magit
 to the remote repository service.
 - [[https://docs.github.com/en/authentication/connecting-to-github-with-ssh/adding-a-new-ssh-key-to-your-github-account][GitHub SSH key documentation]]
 - [[https://docs.gitlab.com/ee/ssh/#add-an-ssh-key-to-your-gitlab-account][GitLab SSH key documentation]]
+** grip-mode
+Grip-mode [[https://github.com/seagle0128/grip-mode#prerequisite][requires python and the python package grip]] to be installed on the
+system. Grip can usually be installed with the command =pip install grip=,
+on debian based systems make sure you use the python 3 version
+of pip, it is normally called =pip3=.
```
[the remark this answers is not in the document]
lebensterben: yes. we should not include any ditro specific instruction.

lebensterben:
on the installation command,

for technical reasons it's prefered to use `python3 -m pip install` in a *certain environment*.

obviously this easily gets very complicated due to virtual environment...

So either just tell the user to use `pip install` or just put the link there and let them figure out themselves...

lebensterben:
if you really want a key binding, just make a major mode key binding for markdown-mode (which is prefixed with SPC m)

robbyoconnor:
Will do @lebensterben

layers/+lang/markdown/README.org
dalanicolai: I would prefer to 'more actively' inform users about this functionality by adding a keybinding for it. But why not leave it just under "ghp" as "gh" already exists and grip is 'more or less' github specific (users can still also call via `M-x` of course)

dalanicolai:
Actually the main reason for adding grip-mode, beside the already, in the markdown layer available `vmd-mode` that also does github markdown rendering, is that grip-mode supports previewing `org` files directly. Therefore, I would still vote for a keybinding under `SPC g h` (although I would not mind another major mode keybinding in markdown mode).

But like @robbyoconnor my idea is that this package is more github related than markdown related (and therefore I did not protest when putting it in the git layer, despite the git layer not being a github layer specifically. Because then we would not be required to install the markdown layer to preview `org` files)... Anyway, in the end, I'll probably support any choice you make.

layers/+lang/markdown/README.org
lebensterben: @dalanicolai 

nope it's not GitHub specific.
GFM is widely used and is one of top makrdown implementations.

though grip makes it sounds like it's used for GitHub only, this package can be used for previewing all GFM.

layers/+lang/markdown/README.org
robbyoconnor: Resolved.

layers/+lang/markdown/packages.el
robbyoconnor: @bryce-carson what does the T stand for?

robbyoconnor:
Okay, so I thought about this. It's also relevant to Org Mode, so what do we do about that?

layers/+lang/markdown/packages.el
robbyoconnor: Nuked the keybinding.

layers/+lang/markdown/README.org
```diff
@@ -153,6 +159,10 @@ To generate a table of contents type on top of the buffer:
 |-------------+---------------------|
 | ~SPC m k~   | kill thing at point |
 
+** Grip-mode
+
+Toggle =grip-mode= by doing ~M-x grip-mode~.
+
```
robbyoconnor, line 165: This should work.

winsphinx:
What difference is in previewing between vmd-mode and grip-mode?

lebensterben:
they use different program, vmd vs grip. some people have preference of one of NodeJS and Python over the other.

lebensterben:
vmd only supports GitHub access token from command line (and yes this is not even documented).

grip is better and you can specify that in a config file or let it read the password from a password manager.

So grip is better in this regard.

---

Both support auto refreshing so that gives you a live preview similar to other markdown editors. BUT, without access token or password, you'd be rate-limited very soon.

robbyoconnor:
I wouldn't be offended if this is rejected but how with the org layer?

lebensterben:
@robbyoconnor 

well... if you want, move it to a tool/grip layer...

robbyoconnor:
@lebensterben we could make a layer for preview tools as well

lebensterben:
that's a good idea actually.

bryce-carson:
The Org functionality is a lesser feature, no?

Could be easily solved by using a delist + set leader keys for major mode to provide a toggle, or better use the toggle functions.

robbyoconnor:
@bryce-carson The idea of moving them to their own layer is useful. What if they don't have the markdown layer?

bryce-carson:
> @bryce-carson The idea of moving them to their own layer is useful. What if they don't have the markdown layer but the org layer?

I'm not sure about that case.

A layer for previewers may be useful, but I don't know too much about how multiple layers work together.

There's also the point that if you wanted just markdown and markdown previewing, you would be getting a lot of other previewers that you're not using.

What's best for the direction to take applications like this in Spacemacs is hard to say. I don't know.

lebensterben:
just make a tools/previewer layer, with vmd-mode and grip-mode. (probably other packages)

each package is gated behind a layer config variable.

robbyoconnor:
I'll do it this weekend. Going to close this and make a new PR

Issue #15426, opened 2022-03-25
Recent undo-tree package leaves *.~undo-tree~ files


Antiarchitect:
#### Description :octocat:
After `undo-tree` upgrade spacemacs starts leaving *.\~undo-tree\~ files on fs 

#### Reproduction guide :beetle:
- Start Emacs
- Edit some file
- git status

*Observed behaviour:* :eyes: :broken_heart:
See *.\~undo-tree\~ files related to those I edit

*Expected behaviour:* :heart: :smile:
No extra files are created

#### System Info :computer:
- OS: gnu/linux
- Emacs: 28.0.91
- Spacemacs: 0.999.0
- Spacemacs branch: develop (rev. 18ad9759b)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(auto-completion ansible docker emacs-lisp erlang git
                 (go :variables go-backend 'lsp)
                 (helm :variables helm-no-header t)
                 html javascript
                 (json :variables json-fmt-tool 'prettier)
                 (lsp :variables lsp-rust-server 'rust-analyzer)
                 multiple-cursors nginx org php puppet python ruby
                 (rust :variables rust-backend 'lsp)
                 (shell :variables shell-default-height 100 shell-default-position 'top shell-default-term-shell "~/Sync/dotfiles/sway-alacritty.sh")
                 shell-scripts sql
                 (syntax-checking :variables syntax-checking-enable-tooltips nil)
                 systemd themes-megapack
                 (treemacs :variables treemacs-show-hidden-files t treemacs-sorting 'alphabetic-asc treemacs-use-filewatch-mode t treemacs-use-follow-mode t)
                 typescript
                 (yaml :variables yaml-enable-lsp nil))
```
- System configuration features: ACL CAIRO DBUS FREETYPE GIF GLIB GMP GNUTLS GPM HARFBUZZ JPEG JSON LCMS2 LIBOTF LIBSYSTEMD LIBXML2 MODULES NATIVE_COMP NOTIFY INOTIFY PDUMPER PGTK PNG RSVG SECCOMP SOUND THREADS TIFF TOOLKIT_SCROLL_BARS XIM XWIDGETS GTK3 ZLIB

#### Backtrace :paw_prints:
```
```

arifer612:
[...]

PR #15440, opened 2022-04-01
[spacemacs-editing] initialize `undo-tree-history-directory-alist`

rayw000:
Initialize `undo-tree-history-directory-alist` to stop `*.~undo-tree~` files from scattering all around.

Fix: #15426

### DIFF
--- a/layers/+spacemacs/spacemacs-editing/packages.el
+++ b/layers/+spacemacs/spacemacs-editing/packages.el
@@ -539,7 +539,16 @@
             ;; Emacs GC which truncages the undo history very aggresively
             undo-limit 800000
             undo-strong-limit 12000000
-            undo-outer-limit 120000000)
+            undo-outer-limit 120000000
+            undo-tree-history-directory-alist
+            `(("." . ,(let ((dir (expand-file-name "undo-tree-history" spacemacs-cache-directory)))
+                        (if (file-exists-p dir)
+                            (unless (file-accessible-directory-p dir)
+                              (warn "Cannot access directory `%s'.
+ Perhaps you don't have required permissions, or it's not a directory.
+ See variable `undo-tree-history-directory-alist'." dir))
+                          (make-directory dir))
+                        dir))))
       (global-undo-tree-mode))
     :config
     (progn


layers/+spacemacs/spacemacs-editing/packages.el
```diff
@@ -539,7 +539,13 @@
...
             undo-limit 800000
             undo-strong-limit 12000000
-            undo-outer-limit 120000000)
+            undo-outer-limit 120000000
+            undo-tree-history-directory-alist
+            `(("." . ,(expand-file-name
+                       (let ((dir (concat spacemacs-cache-directory "undo-tree-history")))
+                         (unless (file-exists-p dir)
+                           (make-directory dir))
+                         dir)))))
```
arifer612, line 548: The logic here will fail if there exists a *file* in the cache directory titled `undo-tree-history`. I employed the following in my own fork:
```
undo-tree-history-directory-alist (list
                                   (cons
                                    "." (file-name-as-directory
                                         (concat spacemacs-cache-directory
                                                 "undo-tree")))))

```
`undo-tree` ensures that the directory exists so you don't have to worry that it has yet to be created.

arifer612 (changes requested):
I was just about to make a PR for my own fork! There's an issue with the logic, more details in the comment.

layers/+spacemacs/spacemacs-editing/packages.el
rayw000: Thanks! It's really an issue here.

Would it be better to generate a warning when the path does exist but not a directory? I use function `file-accessible-directory-p` to achieve this in my new commit.

rayw000:
> Can you also add a comment in your commit message that it resolves #15426 to get GitHub to auto-link this PR to the issue?

With pleasure. I didn't know there had been already an open issue before. Thank you!

lebensterben:
you don't need `concat` inside `expand-file-name`, just use
`(expand-file-name "undo-tree-history" spacemacs-cache-directory)`

smile13241324 (approved):

PR #15449, opened 2022-04-06
core-jump: fix incorrect lookup of :async

corngood:
:async wasn't being retrieved properly due to this typo.

The symptom of this is if you added an async handler before another handler, it would call the async handler and then immediately call the next one as well.  In my case, this would result in `evil-goto-definition` bringing up help while the async handler was running.


lebensterben:
much appreciated!

PR #15452, opened 2022-04-07
Add documentation for HTTPS remote push URLs

bryce-carson:
This documentation provides an example of the remote push URL that must be set
to use a personal access token to authenticate with HTTPS with GitHub.

The existing documentation does not provide information on setting a specific
remote push URL for GitHub.

### DIFF
--- a/layers/+source-control/git/README.org
+++ b/layers/+source-control/git/README.org
@@ -12,6 +12,8 @@
   - [[#git][Git]]
     - [[#basic-git-configuration][Basic Git configuration]]
     - [[#using-ssh-urls-for-remote-repositories][Using SSH URLs for remote repositories]]
+    - [[#using-https-urls-for-remote-repositories][Using HTTPS URLs for remote repositories]]
+      - [[#github-https-push-urls][GitHub HTTPS push URLs]]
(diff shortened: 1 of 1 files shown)


lebensterben:
SSH is better and more secure.
There's no need for personal token.

bryce-carson:
> SSH is better and more secure.
> There's no need for personal token.

A token is necessary for using Magit Forge. This documentation guides users to the fact that if they wish they can use the same token to push to repositories.

Configuring Magit is easier using a single method.

If the push URL can be configured securely, like how the token as password for Forge can be GPG secured, then would this be acceptable?

I don't understand why we would document one usage of HTTPS authentication but not document the other.

lebensterben:
we should not document insecure methods in general unless it's the only method.

insecure methods can be documented elsewhere and then it's not our responsibility (actually it never is) should anything went wrong.

bryce-carson:
> we should not document insecure methods in general unless it's the only method.
> 
> insecure methods can be documented elsewhere and then it's not our responsibility (actually it never is) should anything went wrong.

That's reasonable. When I get around to it, I'll finish the SSH documentation.

Issue #15448, opened 2022-04-06
Unable to stage individual lines using magit


dankessler:
#### Description :octocat:
Unable to stage individual lines using magit

I believe this is due to [this recent change](https://github.com/magit/magit/commit/70b34f08ab5dc8bb7a0fb71bbf5d8cab7f8a47c4#diff-335d97ddde0824a60d53f17c711216f33e2a27860169fc2e02763781e2ae4779L2816-L2818) in `magit`.

Previously, `magit-stage` was bound to `s` using the `keymap` text property,
which has very high priority. Now, instead the text property simply rebinds
`magit-stage-file` to `magit-stage`. `s` is bound to `magit-stage-file` in the
keymap for the `magit-status` global mode, but `evil-surround` binds `s` to
`evil-surround-region` when in visual state, which has higher priority than the
global mode keymap.

This isn't really a bug in `magit` but rather a complication of using
`evil-surround` and `magit` at the same time. This could potentially be remedied
by making a change to `evil-collection` to explicitly map `s` to `magit-stage`
in a way that would override the `evil-surround-region` mapping, or really any solution that would appropriately over-shadow the `evil-surround-region` mapping with `magic-stage`. I'm happy to take this on in a PR, but am open to input on the most graceful way to fix this (e.g., in `spacemacs`, in `evil-collection`, or (somehow?) in `magit` itself).

Reverting to an older version of `magit` remedies the issue.

#### Reproduction guide :beetle:
- Start Emacs
- Edit some lines in a file under version control
- Open a magit status buffer (`SPC g s`)
- Move point to a line in the diff hunk
- Change to visual mode (`v`) and select a line or two (`j`)
- Hit `s` to try to stage just those lines

*Observed behaviour:* :eyes: :broken_heart: 
`s` doesn't stage and instead waits for another character due to `evil-surround-region` firing

*Expected behaviour:* :heart: :smile:
`s` should trigger `magit-stage` and stage the currently selected lines

#### System Info :computer:
- OS: darwin
- Emacs: 27.2
- Spacemacs: 0.999.0
- Spacemacs branch: develop (rev. d79b12e7b)
- Graphic display: t
- Distribution: spacemacs
- Editing style: hybrid
- Completion: helm
- Layers:
```elisp
((auto-completion :disabled-for org)
 better-defaults bibtex csv emacs-lisp
 (ess :variables ess-assign-key "\255")
 eww git
 (helm :variables helm-use-fuzzy 'source)
 html
[...]

PR #15462, opened 2022-04-11
disable evil-surround in magit-status buffers

dankessler:
See #15448 for full discussion.

In brief, `evil-surround` interferes with line-wise staging using `magit`, so this PR adds to the `magit-status-mode-hook` so that `evil-surround` is disabled in `magit-status` buffers.


lebensterben (approved):

lebensterben:
merged. thanks for your contributions.

PR #15471, opened 2022-04-14
[latex] add, configure, and document evil-tex package

dankessler:
Add support for the very useful [evil-tex package](https://github.com/iyefrat/evil-tex) to the latex layer.

Of particular note, this package provides numerous text objects for latex, e.g.,
latex-style quotes and double quotes, and also provides valuable toggles for
managing delimeters.

I've mapped its "magnificent toggles" (which are otherwise under a `mt` prefix)
to `SPC m m` (m for magnificent) to avoid interfering with any standard evil features.

I added brief documentation for evil-tex to the latex layer's README, which
primarily refers the reader to evil-tex's thorough documentation rather
than duplicating it in the readme.

### DIFF
--- a/layers/+lang/latex/packages.el
+++ b/layers/+lang/latex/packages.el
@@ -32,6 +32,7 @@
     (company-reftex :requires company)
     counsel-gtags
     evil-matchit
+    evil-tex
     flycheck
     flyspell
     ggtags
@@ -195,6 +196,19 @@
 
 (defun latex/post-init-flyspell ()
   (spell-checking/add-flyspell-hook 'LaTeX-mode-hook))
+
+(defun latex/init-evil-tex ()
+  (use-package evil-tex
+    :defer t
+    :init
+    (progn
+      (add-hook 'LaTeX-mode-hook #'evil-tex-mode)
+      (setq evil-tex-toggle-override-m nil)
+      (setq evil-tex-toggle-override-t nil))
+    :config
+    (spacemacs/declare-prefix-for-mode 'latex-mode "mq"
+      (cons "evil-tex toggles" evil-tex-toggle-map))))
+
 
 (defun latex/init-reftex ()
   (add-hook 'LaTeX-mode-hook 'turn-on-reftex)
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2685,6 +2685,7 @@ files (thanks to Daniel Nicolai)
   compiled PDF is changed (thanks to Lucius Hu)
 - Automatically configure =pdf-tools= as pdf viewer for =TeX-view= command (by default) (Thanks to Daniel Nicolai)
 - Add layer variable to open =pdf-tools= in split window (Thanks to Daniel Nicolai)
+- New package =evil-tex= (thanks to Dan Kessler)
 **** Lua
 - Added support for auto-completion with =company= (thanks to halfcrazy)
 - Added support for =LSP= (EmmyLua-LS-all, lua-language-server, lua-lsp) (thanks to Lin.Sun)
--- a/layers/+lang/latex/README.org
+++ b/layers/+lang/latex/README.org
@@ -24,6 +24,7 @@
 - [[#key-bindings][Key bindings]]
   - [[#folding-1][Folding]]
   - [[#reftex][RefTeX]]
+  - [[#evil-tex][evil-tex]]
 
 * Description
 This layer adds support for LaTeX files with [[https://savannah.gnu.org/projects/auctex/][AucTeX]].
@@ -34,6 +35,7 @@ This layer adds support for LaTeX files with [[https://savannah.gnu.org/projects
 - Auto-completion
 - Tags navigation on ~%~ with [[https://github.com/redguardtoo/evil-matchit][evil-matchit]]
 - Labels, references, citations and index entries management with [[http://www.gnu.org/software/emacs/manual/html_node/reftex/index.html][RefTeX]]
+- LaTeX-specific text objects and much more with [[https://github.com/iyefrat/evil-tex][evil-tex]]
 
 * BibTeX
 For more extensive support of BibTeX files than RefTeX provides, have a look at
@@ -324,3 +326,20 @@ Available only when =latex-enable-folding= is non nil.
 | ~SPC m r t~ or with LSP ~SPC m R t~     | reftex-toc                            |
 | ~SPC m r T~ or with LSP ~SPC m R T~     | reftex-toc-recenter                   |
 | ~SPC m r v~ or with LSP ~SPC m R v~     | reftex-view-crossref                  |
+
+** evil-tex
+See the [[https://github.com/iyefrat/evil-tex/blob/master/README.org][evil-tex documentation]] for more comprehensive explanation of text
+objects it provides and its other features, including its integration with
+~evil-surround~.
+
+
+| Key binding | Description                                         |
+|-------------+-----------------------------------------------------|
+| ~]]~ / ~[[~ | jump between section headings                       |
+| ~M-n~       | Move between braces, similar to ~TAB~ in ~cd-latex~ |
+| ~SPC m q~   | Prefix for [[https://github.com/iyefrat/evil-tex#toggles][evil-tex toggle commands]]                 |
+
+In order to [[https://github.com/iyefrat/evil-tex#user-options][preserve both the precious ~t~ mark and the indispensable ~ts~
+motion]], the [[https://github.com/iyefrat/evil-tex/blob/master/README.org#toggles][evil-tex "magnificent toggles"]] are bound under ~SPC m q~ rather than
+~mt~ or ~ts~. This binding can be remembered with mnemonic "quite magnificent
+toggle".


layers/+lang/latex/packages.el
```diff
@@ -196,6 +197,22 @@
...
+    :init
+    (progn
+      (add-hook 'LaTeX-mode-hook #'evil-tex-mode)
+      (setq evil-tex-toggle-override-m nil)
+      (setq evil-tex-toggle-override-t nil))
+    :config
+    (progn
+      (spacemacs/set-leader-keys-for-major-mode 'latex-mode
+        "e" evil-tex-toggle-map)
+      (spacemacs/declare-prefix-for-mode 'latex-mode "mm" "magnificent toggles")
```
lebensterben, line 212: I think this needs to be added to :init block otherwise it won't show up until it's loaded.

layers/+lang/latex/packages.el
```diff
@@ -196,6 +197,22 @@
...
+      )))
+
```
lebensterben, line 214: trialing parentheses should not take new lines.

dankessler:
Thanks for the quick review; I'll address. Also just realized that putting the "magnificent toggles" under `m` is no good because that's already mapped to `TeX-insert-macro`. I considered putting it under `SPC m e` (for evil-tex toggles) but didn't want to go against the "eval" convention. Any suggestions for where I might stick it?

layers/+lang/latex/packages.el
dankessler: I think thanks to the magic of auto-loading this "works" (since switching to a latex buffer invokes evil-tex via the hook (which was placed there in init and so then this config block runs). I could move it into init, but not sure if evil-tex-toggle-map is defined prior to evil-tex loading (though I can check)

lebensterben:
why not to use SPC m t for toggle?

dankessler:
That's not a bad idea, and one that I had considered. The drawbacks are (1) If using `lsp`, `T` is currently used for its toggles, although lower case `t` is available. Also (2), although `evil-tex` calls them "toggles" they really aren't toggles in the way that spacemacs usually means toggles. If you look at the ["toggles"](https://github.com/iyefrat/evil-tex#toggles) you'll see what I mean, e.g., one of them "toggles" (edits) the name of the current section in the latex file.

layers/+lang/latex/packages.el
dankessler: Yea, I can't move it to :init because evil--tex-toggle-map is not defined when that block runs. I suppose I could move the prefix declaration there, but seems better to keep them packaged together.

lebensterben:
how about SPC m s (, s)?
this is analogous to s for evil-surround. (and you can explain this in the documentation)

layers/+lang/latex/packages.el
lebensterben: what I meant is to move declare-prefix alone. (not the actual key bindings)

dankessler:
Unfortunately that's already bound to LaTeX-section (it's also convention for ["sending code"](https://develop.spacemacs.org/doc/CONVENTIONS.html#send-code)). Another strike against t is that by convention it is for editing tables. For now there's nothing bound there, but latex is very table-centric and so it'd be nice to keep that reserved. 

After looking carefully at the keymap, the best I can come up with is SPC m q (, q). It is not currently mapped to a command nor does it correspond to any of the documented spacemacs conventions. In the documentation, I can explain it with the (loose) mnemonic that these are for the "quite magnificent toggles" (a sort of meta-joke on the loose mnemonic of "magnificent toggles) that evil-tex used for binding it to mt.

layers/+lang/latex/packages.el
dankessler: Is "declare-prefix" the "idiomatic" spacemacs way to give pretty labels to prefix keymaps in which-key? I just followed the pattern I saw elsewhere in the file.

lebensterben: spacemacs/declare-prefix-for-mode is the idiomatic way for that in a major/minor mode specific keymap.

lebensterben:
okay. that sounds good.

layers/+lang/latex/packages.el
dankessler: Moving the declaration into :init doesn't work either. It doesn't throw an error, but when it actually gets bound it overwrites the string and just shows up as +prefix in which-key. Everywhere else I've seen it used the prefix key gets mapped first and then declare-prefix follows (presumably for this reason). I'll go back to wrapping them in progn in :config so that I can do them in the proper order.

lebensterben: > It doesn't throw an error, but when it actually gets bound it overwrites the string and just shows up as +prefix in which-key. 

well this is a known problem when the same prefix is used again...

dankessler: Sorry, I don't follow. If I want to put a given keymap behind a key *and* give it a "friendly" name in which-key, what am I supposed to do?

lebensterben: it's not documented.

```
(spacemacs/declare-prefix-for-mode 'foo-mode
  "<prefix-keys>" '("description . map))
```

something like this adds both description and key bindings.
(you can check the documentation for which-key-add-keymap-based-replacements)

dankessler: No luck. It complains about evil-tex-toggle-map failing commandp

lebensterben: check documentation of which-key-add-keymap-based-replacements

I may be mistaken on the calling convention.

dankessler: From the definition of spacemacs/declare-prefix-for-mode and what it calls, it seems like it doesn't handle any mapping, just changing descriptions (it calls which-key-add-keymap-based-replacements), so it doesn't seem to me like there is an obvious "one-shot" way of doing the mapping and the relabeling at once, and instead one needs to first do the mapping, and then do the relabeling to avoid the former overwriting the latter.

I do see a note about preferring the "cons" style for replacement being preferred by which-key, but when the command is a prefix, then command is nil and it's a trivial cons.

lebensterben: which-key-add-keymap-based-replacements should add both description and bindings.

lebensterben: > but when the command is a prefix, then command is nil and it's a trivial cons

the behavior is actually subtly different in terms of font face... that's not documented.

dankessler: > which-key-add-keymap-based-replacements should add both description and bindings.

I don't think so. I tried evaluating the example from the which-key documentation:
```
(which-key-add-keymap-based-replacements global-map
  "C-x w" '("Save as" . write-file))
```

yet `C-x w` remains bound to window selection for me

lebensterben: see this

https://github.com/justbur/emacs-which-key/blob/1217db8c6356659e67b35dedd9f5f260c06f6e99/which-key.el#L923

dankessler: Fair enough. I've gone a bit further down this rabbit hole and just can't figure out a way to get this to work. Consider the following elisp, inspired by [this section of the `which-key` documentation](https://github.com/justbur/emacs-which-key#keymap-based-replacement)

```
(define-key global-map (kbd "C-t") '("My keymap" . evil-tex-toggle-map))
(define-key global-map (kbd "C-t") '("My keymap" . (evil-tex-toggle-map)))
(define-key global-map (kbd "C-t") '("My keymap" . 'evil-tex-toggle-map))
(define-key global-map (kbd "C-t") evil-tex-toggle-map)
```

The last line works as I'd expect and causes `C-t` to be a prefix key for `evil-tex-toggle-map` (when `C-t` isn't intercepted by a higher priority keymap), but of course doesn't set the description "My keymap".

The first three lines all run but fail to effect a binding. If I try to fire `C-t`, they all fail with `Wrong type argument: commandp, x` where `x` is the cdr of the cons that I used.

I looked at the [documentation for `define-key`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Changing-Key-Bindings.html#index-define_002dkey), where the argument in question is `binding`, which referred me to the documentation for [key lookup](https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Lookup.html). Assuming that documentation is correct, it seems that if `binding` is a list, then it's `car` needs to either be the symbol `keymap` or `lambda`; there's nothing about using the car for a string and putting the function in the cdr.

dankessler: It seems I am not [the only one confused by this](https://github.com/justbur/emacs-which-key/issues/253)/

Here is the approach that works
```
(define-key global-map (kbd "C-t") (cons "My keymap" evil-tex-toggle-map))
```

I'd argue that the documentation for `which-key` is a bit confusing, but maybe my elisp-skills are just too poor to understand what's going on.

dankessler: and returning to my specific example, here is a command which in the :config block remaps the key and sets the description all in one shot, as you suggested

```
(spacemacs/declare-prefix-for-mode 'latex-mode "mq"
        (cons "evil-tex toggles" evil-tex-toggle-map))
```

dankessler:
I think this is probably good to go now (although @lebensterben is great at finding small mistakes when I think there are none left :)

smile13241324 (approved):

Issue #15472, opened 2022-04-14
TAB in an Org Agenda buffer in terminal Emacs is bound to the wrong function


swinkels:
#### Description :octocat:
When Emacs runs in the terminal and cursor is in an `*Org Agenda*` buffer on an entry, pressing TAB calls `evil-jump-forward` (and in effect does nothing). It used to call `org-agenda-goto` and open the entry in another window.

**Analysis**

In a terminal, TAB and C-i are the same. In `core-spacemacs.el`, function `spacemacs/init` specifically sets `evil-want-C-i-jump` to `nil` to "avoid TAB being overlapped in terminal mode". Commit bbd40f113fe396a2109086db0529751c3e9d0846 of January 8 2022 bypasses that logic and always binds `C-i` to `evil-jump-forward`, see the change in `evil-evilified-state.el`.

#### Reproduction guide :beetle:
- Start Emacs in a terminal, so using `emacs -nw`.
- `M-x org-agenda`.
- In the `*Org Agenda*` buffer, move point to an entry and press TAB.

*Observed behaviour:* :eyes: :broken_heart:
Nothing happens.

*Expected behaviour:* :heart: :smile:
The `*Org Agenda*` entry at point should open in another window.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 27.2
- Spacemacs: 0.999.0
- Spacemacs branch: develop (rev. 587f584eb)
- Graphic display: nil
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(yaml auto-completion c-c++
      (c-c++ :variables c-c++-backend 'lsp-ccls)
      emacs-lisp javascript git helm html lsp lua markdown nutter org outshine
      (python :variables python-backend 'lsp python-lsp-server 'pylsp python-formatter 'black python-format-on-save t python-test-runner 'pytest)
      (shell :variables shell-default-height 30 shell-default-position 'bottom)
      syntax-checking treemacs)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG CAIRO SOUND GPM DBUS GSETTINGS GLIB NOTIFY INOTIFY ACL GNUTLS LIBXML2 FREETYPE HARFBUZZ M17N_FLT LIBOTF ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XDBE XIM MODULES THREADS JSON PDUMPER GMP

#### Backtrace :paw_prints:
```
[...]

PR #15473, opened 2022-04-14
Only bind TAB when evil setting allows it

swinkels:
In a terminal TAB and C-i generate the same key press. So to "avoid TAB being overlapped in terminal mode", function spacemacs/init sets evil-want-C-i-jump to nil. Unfortunately commit bbd40f1 always binds C-i to evil-jump-forward regardless
of the value of evil-want-C-i-jump. This pull request fixes that.

This pull request addresses ticket #15472.

### DIFF
--- a/layers/+distributions/spacemacs-bootstrap/local/evil-evilified-state/evil-evilified-state.el
+++ b/layers/+distributions/spacemacs-bootstrap/local/evil-evilified-state/evil-evilified-state.el
@@ -198,7 +198,11 @@ Needed to bypass keymaps set as text properties."
 (define-key evil-evilified-state-map (kbd "C-d") 'evil-scroll-down)
 (define-key evil-evilified-state-map (kbd "C-u") 'evil-scroll-up)
 (define-key evil-evilified-state-map (kbd "C-o") 'evil-jump-backward)
-(define-key evil-evilified-state-map (kbd "C-i") 'evil-jump-forward)
+(if evil-want-C-i-jump
+    ;; In terminal mode, `TAB` and `<C-i>' generate the same key press. Only if
+    ;; `evil-want-C-i-jump' holds, `TAB' should bind to the same function as
+    ;; `<C-i>'.
+    (define-key evil-evilified-state-map (kbd "C-i") 'evil-jump-forward))
 (define-key evil-evilified-state-map (kbd "C-z") 'evil-emacs-state)
 (define-key evil-evilified-state-map (kbd "C-w") 'evil-window-map)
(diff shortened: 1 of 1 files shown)


layers/+distributions/spacemacs-bootstrap/local/evil-evilified-state/evil-evilified-state.el
```diff
@@ -198,7 +198,11 @@ Needed to bypass keymaps set as text properties."
 (define-key evil-evilified-state-map (kbd "C-d") 'evil-scroll-down)
 (define-key evil-evilified-state-map (kbd "C-u") 'evil-scroll-up)
 (define-key evil-evilified-state-map (kbd "C-o") 'evil-jump-backward)
-(define-key evil-evilified-state-map (kbd "C-i") 'evil-jump-forward)
+(if evil-want-C-i-jump
```
lebensterben, line 201: use when instead of if

swinkels: > use when instead of if

Indeed, that's better.

> also you can add a test for terminal. so in GUI Emacs, it always binds C-i to evil-jump-forward.

Just to be sure I understand correctly: when in GUI Emacs, you want the value of `evil-want-C-i-jump` to be ignored. I understand the rationale but unless I'm mistaken, this will change behavior from upstream evil.

Finally, is it OK if the pull request contains more commits or should I squash all commits to one?

swinkels: By the way, does `dotspacemacs-distinguish-gui-tab` also play a role here?

lebensterben: > Just to be sure I understand correctly: when in GUI Emacs, you want the value of evil-want-C-i-jump to be ignored. I understand the rationale but unless I'm mistaken, this will change behavior from upstream evil.

you're right on this.

I might remove C-i and C-o from evilified mode later and add them on a per major mode basis.

---

no need to squash it manually.
GitHub's default is squash and merge.

lebensterben: > By the way, does `dotspacemacs-distinguish-gui-tab` also play a role here?

it doesn't seem to.

swinkels: Done.

smile13241324 (approved):

PR #15446, opened 2022-04-06
Fix function quote warning

bryce-carson:
Change `'` to `#'`, resolves #15443

PR #15477, opened 2022-04-18
change read syntax for functions from ' -> #'

dankessler:
I just used `ripgrep` to try to find instances of `'(lambda` that weren't preceded by `#`, so it's possible I missed a few that had more exotic whitespace. 

I don't know that this change will make a huge difference, but without it I get warnings (in emacs 28) whenever the relevant bits are called.


dankessler:
This likely partially conflicts with #15446, but includes fixes not therein.

lebensterben (changes requested):
#' and (lambda () ..) are duplicated

dankessler:
> #' and (lambda () ..) are duplicated

Sorry, GitHub is telling me that you have requested changes, but I can't see any annotations on the diff to indicate what changes you might have requested or where. All of my changes edit forms like `'(lambda` to `#'(lambda`. Are you saying that there are some (all?) instances where the edit should instead be `'(lambda` to `(lambda`?

lebensterben:
> Are you saying that there are some (all?) instances where the edit should instead be '(lambda to (lambda?

All of them should just be (lambda () ..)

dankessler:
So I presume then that despite their equivalence, the `(lambda` form is preferred to `#'(lambda` (perhaps because the former is more concise)? Arguably both are preferred to `'(lambda` (which now raises warnings).

The `#'(lambda` form appears in numerous other places in the `spacemacs` code base: should I fix those too while I'm at it?

lebensterben:
> So I presume then that despite their equivalence, the `(lambda` form is preferred to `#'(lambda` (perhaps because the former is more concise)? Arguably both are preferred to `'(lambda` (which now raises warnings).

yes.


> The `#'(lambda` form appears in numerous other places in the `spacemacs` code base: should I fix those too while I'm at it?

yes if you want.

dankessler:
Ok, PR has been updated to
1.  edit `'(lambda` -> `(lambda` (since otherwise this raises warnings)
2. edit `#'(lambda` -> `(lambda` (the former is harmless but the latter makes things more consistent)

lebensterben:
files under core/lib and local should not be altered.

dankessler:
Oops; fixed

> files under core/lib and local should not be altered.

lebensterben (approved):

PR #15484, opened 2022-04-22
Permit `.spacemacs.d/init.el` to not be user init

bryce-carson:
`DOCUMENTATION.org`:
- Add documentation for Chemacs users to the Alternative Dotfile Directory section in `DOCUMENTATION.org`
- Add example Chemacs `.emacs-profiles.el` to aide understading of added documentation

`core/core-dotspacemacs.el`:
- Add comment to explain the purpose of the `let*` form used to define `dotspacemacs-directory` and `dotspacemacs-filepath`
- Add check for optional `SPACEMACS_INIT_FILE_NAME` environment variable, and fall back to `"init.el"` if the environment variable is not set
- Remove unnecessary fall-through `(t nil)` form in `dotspacemacs-directory`'s cond form: it is unnecessary, and the style of the other usage of cond makes this proposed change more consistent
- Move the fallback to `~/.spacemacs` in the cond for the definition of `dotspacemacs-filepath` to the end of the cond, so that `~/.spacemacs` can be used with or without a dotfile dir, and that dir being either set by the SPACEMACSDIR environment variable or the default location (`~/.spacemacs.d`)
- Change from a forced clause in the `dotspacemacs-filename`–setting cond to just `(default-init)`, for the same reason as the other cond change

### DIFF
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -550,6 +550,21 @@ change the location of this directory.
 so =~/.spacemacs= must not exist for =~/.spacemacs.d/init.el= to be used by
 Spacemacs.
 
+*** Chemacs users with =$CONFIG_HOME/.spacemacs.d= as =$SPACEMACSDIR=
+Users of Chemacs may use =$CONFIG_HOME/.spacemacs.d= as =$SPACEMACSDIR=, so long
+as they define an additional environment variable =SPACEMACS_INIT_FILE_NAME= to
+something other than ~init.el~. Chemacs users can still use =.spacemacs.d= as
+their dotfile directory with Chemacs as long as they retain a =.spacemacs= file
+otherwise, if they do not declare an alternative dotfile name and locate that in
+the dotfile directory.
+
+#+name: chemacs-spacemacsdir-example
+#+begin_example emacs-lisp
+  (("default"   . ((user-emacs-directory . "~/.spacemacs.d")
+		               (env . (("SPACEMACS_INIT_FILE_NAME" . "spacemacs")))))
+   ("gnu"       . ((user-emacs-directory . "~/.emacs.gnu.d"))))
+#+end_example
+
 ** Synchronization of dotfile changes
 To apply the modifications made in =~/.spacemacs= press ~SPC f e R~. It will
 re-execute the Spacemacs initialization process.
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -37,6 +37,7 @@
 - [[#dotfile-configuration][Dotfile Configuration]]
   - [[#dotfile-installation][Dotfile Installation]]
   - [[#alternative-dotdirectory][Alternative dotdirectory]]
+    - [[#chemacs-users-with-config_homespacemacsd-as-spacemacsdir][Chemacs users with =$CONFIG_HOME/.spacemacs.d= as =$SPACEMACSDIR=]]
   - [[#synchronization-of-dotfile-changes][Synchronization of dotfile changes]]
   - [[#testing-the-dotfile][Testing the dotfile]]
   - [[#dotfile-contents][Dotfile Contents]]
@@ -561,8 +562,8 @@ the dotfile directory.
 #+name: chemacs-spacemacsdir-example
 #+begin_example emacs-lisp
   (("default"   . ((user-emacs-directory . "~/.spacemacs.d")
-		               (env . (("SPACEMACS_INIT_FILE_NAME" . "spacemacs")))))
-   ("gnu"       . ((user-emacs-directory . "~/.emacs.gnu.d"))))
+                   (env . (("SPACEMACS_INIT_FILE_NAME" . "spacemacs")))))
+  ("gnu"       . ((user-emacs-directory . "~/.emacs.gnu.d"))))
 #+end_example
 
 ** Synchronization of dotfile changes
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -35,9 +35,15 @@
 Useful for users in order to given them a hint of potential bottleneck in
 their configuration.")
 
+;; Set directory variable containing effective `.spacemacs' file. Will be one of
+;; `~/.spacemacs', `~/.spacemacs.d/init.el',
+;; `~/.spacemacs.d/${SPACEMACS_INIT_FILE_NAME}', `${SPACEMACSDIR}/init.el', or
+;; `${SPACEMACSDIR}/${SPACEMACS_INIT_FILE_NAME}'.
 (let* ((env (getenv "SPACEMACSDIR"))
        (env-dir (when env (expand-file-name (concat env "/"))))
-       (env-init (and env-dir (expand-file-name "init.el" env-dir)))
+       (env-init-file-name (or (getenv "SPACEMACS_INIT_FILE_NAME")
+                               "init.el"))
+       (env-init (and env-dir (expand-file-name env-init-file-name env-dir)))
        (no-env-dir-default (expand-file-name
                             (concat user-home-directory
                                     ".spacemacs.d/")))
@@ -45,8 +51,7 @@ their configuration.")
   (defconst dotspacemacs-directory
     (cond
      ((and env (file-exists-p env-dir)) env-dir)
-     ((file-exists-p no-env-dir-default) no-env-dir-default)
-     (t nil))
+     ((file-exists-p no-env-dir-default) no-env-dir-default))
     "Optional spacemacs directory, which defaults to
 ~/.spacemacs.d. This setting can be overridden using the
 SPACEMACSDIR environment variable. If neither of these
@@ -55,18 +60,19 @@ directories exist, this variable will be nil.")
   (defvar dotspacemacs-filepath
     (let ((spacemacs-dir-init (when dotspacemacs-directory
                                 (concat dotspacemacs-directory
-                                        "init.el"))))
+                                        env-init-file-name))))
       (cond
        (env-init)
-       ((file-exists-p default-init) default-init)
        ((and dotspacemacs-directory (file-exists-p spacemacs-dir-init))
         spacemacs-dir-init)
-       (t default-init)))
-    "Filepath to the installed dotfile. If SPACEMACSDIR is given
-then SPACEMACSDIR/init.el is used. Otherwise, if ~/.spacemacs
-exists, then this is used. If ~/.spacemacs does not exist, then
-check for init.el in dotspacemacs-directory and use this if it
-exists. Otherwise, fallback to ~/.spacemacs"))
+       ((file-exists-p default-init) default-init)
+       (default-init)))
+    "Filepath to the installed user–init file (dotfile, like
+`.spacemacs'). If the SPACEMACSDIR environment variable is set
+then `${SPACEMACSDIR}/init.el' is used. If the
+`$SPACEMACS_INIT_FILE_NAME' environment variable is set its value
+is substituted for `init.el'. Otherwise `~/.spacemacs' is used if
+it exists; if it does not exist it will be created."))
 
 (spacemacs|defc dotspacemacs-distribution 'spacemacs
   "Base distribution to use. This is a layer contained in the directory


bryce-carson:
This is why, basically:
<html>
<pre><font color="#9580FF">λ  </font><font color="#50FA7B">tree</font> <font color="#FFB86C">-a</font> <font color="#FFB86C">-d</font> <font color="#FFB86C">-L</font> <font color="#F8F8F2">2</font>                                                                                                                                                                                   <font color="#D5CCFF">~</font>
<font color="#9B6BDF">.</font>
├── <font color="#9B6BDF">.emacs.doom.d</font>    ;; 👀👀👀
├── <font color="#9B6BDF">.emacs.d</font>    ;; Where Chemacs lives
├── <font color="#9B6BDF">.emacs.gnu.d</font>   ;; GNU Emacs
└── <font color="#9B6BDF">.spacemacs.d</font>   ;; Spacemacs' happy little home
    └── <font color="#9B6BDF">init.el</font>    ;; The normal `init.el`
    └── <font color="#9B6BDF">spacemacs</font>    ;; "`.spacemacs`"
</pre>
</html>

doc/DOCUMENTATION.org
```diff
@@ -550,6 +551,21 @@ change the location of this directory.
 so =~/.spacemacs= must not exist for =~/.spacemacs.d/init.el= to be used by
 Spacemacs.
 
+*** Chemacs users with =$CONFIG_HOME/.spacemacs.d= as =$SPACEMACSDIR=
```
lebensterben, line 554: CONFIG_HOME may not be set on some systems. how do you deal with that?

doc/DOCUMENTATION.org
```diff
@@ -550,6 +551,21 @@ change the location of this directory.
...
+Users of Chemacs may use =$CONFIG_HOME/.spacemacs.d= as =$SPACEMACSDIR=, so long
+as they define an additional environment variable =SPACEMACS_INIT_FILE_NAME= to
+something other than ~init.el~. Chemacs users can still use =.spacemacs.d= as
+their dotfile directory with Chemacs as long as they retain a =.spacemacs= file
+otherwise, if they do not declare an alternative dotfile name and locate that in
+the dotfile directory.
+
+#+name: chemacs-spacemacsdir-example
```
lebensterben, line 562: what does this code snippet mean?
where should it be added?
it's not correctly indented.

doc/DOCUMENTATION.org
bryce-carson: Woops! That's just a bad referent; I'm not using CONFIG_HOME in the code. I should have referred to HOME.

The purpose of the heading was to guide users who are using `spacemacs.d` as their `.emacs.d`; `CONFIG_HOME` was unnecessary.

I'll improve the clarity of this heading.

doc/DOCUMENTATION.org
bryce-carson: It's only an example for Chemacs users. If a user is not using Chemacs, is using `.spacemacs.d`, and doesn't want to use `~/.spacemacs` then they will need to declare the environment variable in their shell.

I can improve the clarity of the documentation and code comments.

bryce-carson: For Chemacs users, this is an example of the `.emacs-profiles.el` file.

lebensterben:
question: why do we need to make changes?
does chemacs's instructions on Spacemacs stopped working?

lebensterben:
have you tried this without chemacs?

bryce-carson:
> question: why do we need to make changes?
> does chemacs's instructions on Spacemacs stopped working?

See the name of this PR and my second comment in the conversation (with the file tree).

With the current code, users must either use `~/.spacemacs` or `~/.spacemacs.d/init.el` for their configuration file. This doesn't work if `~/.spacemacs.d` ***is*** their `.emacs.d`, and not just a hidden folder containing their user configuration file.

bryce-carson:
> have you tried this without chemacs?

I haven't, but Chemacs is not required. Without Chemacs, the environment variable must be set in the user's shell. If the environment variable is not set, the behaviour remains the same as now and `init.el` will be used as the configuration file  if `SPACEMACSDIR` is set and `~/.spacemacs` does not exist, or if `~/.spacemacs` does not exist and `~/.spacemacs.d/init.el` does exist and `user_home_directory` is not `~/.spacemacs.d`.

This fixes the assumption that `~/.spacemacs.d/init.el` is the user's configuration file, as the PR name states.

doc/DOCUMENTATION.org
bryce-carson: > it's not correctly indented.
Are you looking at 26c054394c727a7b868c1d60e19b2b51f3ec82e5? I already changed the indentation of the example last night.

bryce-carson:
> have you tried this without chemacs?

With Spacemacs/Emacs living in `~/.emacs.d`, this works:
```sh
SPACEMACS_INIT_FILE_NAME="config" emacs --dump-file=/Users/bryce/.emacs.d/.cache/dumps/spacemacs-27.2.pdump
```

With a symbolic link from `~/.emacs.d` to `~/.spacemacs.d`, this works:
```sh
SPACEMACSDIR="/Users/bryce/.spacemacs.d" SPACEMACS_INIT_FILE_NAME="config" emacs
```

As I said, Chemacs not required. :)

sunlin7:
There is a simple way to satisfiy your requirement, the key point is change `spacemacs-start-directory`.
```
emacs -q --eval '((lambda (x)
   (setq spacemacs-start-directory (file-name-directory x))
   (load-file x))
 (expand-file-name "~/.spacemacs.d0/init.el"))'
```
So you can put this logical into your script, or, submit common change to official `.spacemacs.d/init.el` file.

emacs18:
I have been using chemacs for a long time so that I can select among several spacemacs variants, doom, scimax, and other startup files when emacs starts.  I do have some custom changes of spacemacs on top of develop branch, e.g.,
```diff
diff --git a/init.el b/init.el
index 2ce6f5bf6..db7f8ff4e 100644
--- a/init.el
+++ b/init.el
@@ -27,6 +27,11 @@
 ;; Avoid garbage collection during startup.
 ;; see `SPC h . dotspacemacs-gc-cons' for more info
 (defconst emacs-start-time (current-time))
+(setq spacemacs-start-directory (file-name-directory load-file-name))
+
+;; Load following rather than ~/.spacemacs.
+(setq dotspacemacs-filepath (expand-file-name "core/templates/.spacemacs.template" spacemacs-start-directory))
+
 (setq gc-cons-threshold 402653184 gc-cons-percentage 0.6)
 (load (concat (file-name-directory load-file-name)
               "core/core-versions")
```
It has been too long since I made my custom changes, so I don't remember which of my changes were made to allow chemacs to work.  I don't have or need `~/.spacemacs` since above change uses `.spacemacs.template` instead.  If you are curious, all my changes can be found in https://github.com/emacs18/spacemacs/commits/sm-my branch which comes off of develop branch.  Some of the changes are to allow use of `straight.el` instead of `package.el`.

emacs18:
Emacs 29 now support  `--init-directory` option which makes chemacs no longer necessary.
See https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=8eaf04de83
Given this is this PR still relevant?

bryce-carson:
> Emacs 29 now support  `--init-directory` option which makes chemacs no longer necessary.
> See https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=8eaf04de83
> Given this is this PR still relevant?
> 

As explained, the presence of documentation changes related to Chemacs in this PR is secondary, and not necessarily related to the proposed code changes.

Regardless of Emacs 29 and Chemacs, this PR was converted to draft for the security reason that Lucius noted.

smile13241324:
I have just followed up the conversation, and I understand the need to have spacemacs being flexible when it comes to user specific file organisation.

However given that emacs 29 is kind of solving the issue for us I wonder if this is really worth the time.

I am still going to have a look at the changes @emacs18 has done on his develop and see if I can integrate parts of this into the trunk, especially as I am really interested in straight.el myself, but no promises.

lebensterben:
btw I've installed chemacs 2 and there's no need to do any configuration on Spacemacs's side.

emacs18:
How would you setup multiple spacemacs configurations?  For example vanilla develop branch of spacemacs as well as spacemacs with personal setup added to it?  For me the issue is `~/.spacemacs` file.  I need two different versions of this.  Hence my solution is to launch emacs like this:
```sh
#!/bin/bash

export SPACEMACSDIR=${HOME}/.emacs.d/$1/.spacemacs

exec emacs29 --init-directory ${HOME}/.emacs.d/$1 --debug-init &
```
and setup each config directory to use its own copy of the spacemacs init file like this:

```sh

$ file ~/.emacs.d/sm-ms/.spacemacs/*
/home/me/.emacs.d/sm-ms/.spacemacs/init.el: symbolic link to ../core/templates/.spacemacs.template
$ 
```

lebensterben:
> How would you setup multiple spacemacs configurations?

good question. i ignored this use case.

lebensterben:
@emacs18 

but that sounds like an edge case.

most users should not need to have two separate .spacemacs.

bryce-carson:
> How would you setup multiple spacemacs configurations?  For example vanilla develop branch of spacemacs as well as spacemacs with personal setup added to it?  For me the issue is `~/.spacemacs` file.  I need two different versions of this.  Hence my solution is to launch emacs like this:
> ```sh
> #!/bin/bash
> 
> export SPACEMACSDIR=${HOME}/.emacs.d/$1/.spacemacs
> 
> exec emacs29 --init-directory ${HOME}/.emacs.d/$1 --debug-init &
> ```
> and setup each config directory to use its own copy of the spacemacs init file like this:
> 
> ```sh
> 
> $ file ~/.emacs.d/sm-ms/.spacemacs/*
> /home/me/.emacs.d/sm-ms/.spacemacs/init.el: symbolic link to ../core/templates/.spacemacs.template
> $ 
> ``` 
> 

I maintained two separate configurations of Spacemacs using Chemacs2. It was convenient due to the single flag and a simple elsip config file to define profiles.

I used the *same* `.spacemacs` in my home directory, however.

@lebensterben , what about documenting the ability to predefine the dotspavemacs file used in early-init?, which overrides any searching logic for the .spacemacs file?

That'd do several things:
1. Permit .spacemacs.d/init.el to *not* be .spacemacs when early init is specifying an alternate path (the assumption is not made any more, and the init.el takes on its usual meaning)
2. Allow multiple .dotspacemacs for emacs18, and
3. Retain current behaviour *if* the user does not predefine the dotspacemacs-filepath constant in their early init.

It also avoids introducing another envvar, and is a simple implementation, simply wrap the current form defining the constant with an unless!


```elisp
(unless dotspacemacs-filepath
  (defconst dotspacemacs-filepath
    ;; ... the rest of the definition, including the INITVALUE argument
))
```

It's been some time since I have thought about this, however.

lebensterben:
that's *over engineering*.

users should not have multiple configurations.

if users still want multiple configs, they should manually hack it.

lebensterben:
one way to hack .spacmeacs is to realize that this is a normal elisp file.

for example, you have two configurations for two systems, the pseudo code  should look like:

```txt
(if (SOME-KIND-OF-TESTS) 
    (load DOTSPACEMACS-ONE)
  (load DOTSPACEMACS-OTHER))
```

emacs18:
I actually have more than two spacemacs configuration.
- one I use daily using my own customizations using straight.el and using vim style
- same one using packge.el
- one using emacs style rather than vim style
- vanilla spacemacs without any customizations

I of course use the first one most of the time, but I do sometimes use other ones for testing purposes.

bryce-carson:
> that's *over engineering*.
> 
> users should not have multiple configurations.
> 
> if users still want multiple configs, they should manually hack it.

A single `when` form is over-engineering Spacemacs?

The point is not to have multiple .spacemacs for a single Emacs process with the Spacemacs framework, but to ensure a given Spacemacs is running with the correct file.

It is a sane thing to do *when* you're hacking on Spacemacs to keep your hacks in a separate clone and using either the template dotspacemacs, the same as usual, or a different one from how you nornally configure develop.

While I was hacking on it, reading every core file (Spacemacs' core is a tidy ~20,000 lines), I needed to keep a separate clone and use my normal for writing the hacks.

I don't think providing the *option*, with a couple lines of new comments in early-init is bad, except if there is some other part of the loading procrss impacted and the assumptions made in the earlier design were intentional.

I suppose a look at the history of ~/.spacemacs vs ~/.spacemacs.d/init.el might give us insight.

I'm guessing a few users wanted a directory to house several elisp files and load them from their dotspacemacs rather than have a monolithic configuration file, given ~/.spacemacs.d/ *cannot* house Spacemacs itself.

lebensterben:
> A single when form is over-engineering Spacemacs?

what is the single when form?

bryce-carson:
> > A single when form is over-engineering Spacemacs?
> 
> what is the single when form?

I mentioned it in my previous response.

Surrounding the current constant definition of dotspacemacs-filepath in core-dotspacemacs.el (line 50) with a when form to check if it has already been defined (which would be the case if you predefine it in early init yourself).

Like so (sorry for screenshotting it, I'm on mobile and this will preserve the indentation)

...pardon me, I meant an `unless`

bryce-carson:
Currently the return value of the defconst form is unused, so wrapping it and changing the overall return value shouldn't have any ill effects.

bryce-carson:
Correction
```elisp
(unless (boundp 'dotspacemacs-filepath)
    (defconst dotspacemacs-filepath ...))
```

lebensterben:
replace defconst by defcustom should suffice.

bryce-carson:
> replace defconst by defcustom should suffice.

An interesting alternative. I'll leave a PR for it up to someone else, but after a quick test I don't see any issues with allowing this.

It worked fine in my test.

@emacs18 what do you think?

emacs18:
It appears that the point of the proposed change is to allow users to set `dotspacemacs-filepath` early on to over-ride the default value, right?  So this would be an alternate method  which avoids messing with `SPACEMACSDIR` environment variable.

What is the advantage of using 
```elisp
(setq dotspacemacs-filepath my-path)
```
verses 
```elisp
(setenv "SPACEMACSDIR" my-path)
```
Wouldn't these two be equivalent?

Also I see 
```elisp
(defconst dotspacemacs-directory
  (let* ((spacemacs-dir-env (getenv "SPACEMACSDIR"))
  ...
```
Thus if you don't set `SPACEMACSDIR`, then don't you need to set not only `dotspacemacs-filepath`, but also `dotspacemacs-directory`?

lebensterben:
environment variable can be set by 

- any shell startup script
- any scripts sourced by the shell startup script
- your scripts source by your desktop environment or VM/container
- the process that starts your emacs instance
- user overrides in the terminal
- user overrides in emacs initialization files

it's a total mess and cryptic.

emacs18:
You seem to think the many ways in which environment variables can be set is a disadvantage. I've heard some say the exact opposite arguing that the flexibility is an advantage. However I think this is off topic, because we are not discussing getting rid of `SPACEMACSDIR`.

The use of `SPACEMACSDIR` environment variable has been around many years. I doubt that anyone would consider getting rid of it, because some do not like using environment variables in general.

Assuming that `SPACEMACSDIR` will continue to exist, what is the advantage of making changes to spacemacs to allow use of

```elisp
    (setq dotspacemacs-directory some-path)
    (setq dotspacemacs-filepath (expand-file-name "init.el" dotspacemacs-directory))
```

verses the following

```elisp
    (setenv "SPACEMACSDIR" some-path)
```

Even if former was doable today, why would you use that verses the much simpler latter?

lebensterben:
> I doubt that anyone would consider getting rid of it

Me.

lebensterben:
this environment variable can be set multiple times with different values across various places. there is no way we can track where it is set for a generic setup. that's a maintenance hell us.

emacs18:
Really?  Why would you consider breaking existing uses?  What possible benefit do you see to outweigh certain outcry from long time users?

emacs18:
Can you provide example of "maintenance hell"?  Emacs variables also can be set on command line and arbitrary files loaded by the user.  So what is the difference?

lebensterben:
suppose the user has multiple spacemacs configurations and expects it to be loaded automatically by changing the said environment variable.

so each value of the variable leads to a set of expected behaviors of emacs.

now the user wrongfully think the variable has value X and expect emacs to have behaviors X, but in fact the value is Y and emacs have behaviors Y.

the user opens an issue and complains about the discrepancy. we spend time diagnosing the causes but find nothing. after some back and forth he realized that he messed up the environment variable. we tell him to fix it. he doesn't know how because otherwise he wouldn't mess it up in the first place.

now if we choose to help him to fix the environment variable. we need him to provide the scripts that could be sourced. but this cannot be explained easily and concisely because there are multiple actors that source shell scripts, as I explained earlier. 

surely we could tell the user to do `env=value emacs`. but why not let user simply set a customization variable in the first place?

bryce-carson:
> You seem to think the many ways in which environment variables can be set is a disadvantage. I've heard some say the exact opposite arguing that the flexibility is an advantage. However I think this is off topic, because we are not discussing getting rid of `SPACEMACSDIR`.
> 
> The use of `SPACEMACSDIR` environment variable has been around many years. I doubt that anyone would consider getting rid of it, because some do not like using environment variables in general.
> 
> Assuming that `SPACEMACSDIR` will continue to exist, what is the advantage of making changes to spacemacs to allow use of
> 
> ```elisp
>     (setq dotspacemacs-directory some-path)
>     (setq dotspacemacs-filepath (expand-file-name "init.el" dotspacemacs-directory))
> ```
> 
> verses the following
> 
> ```elisp
>     (setenv "SPACEMACSDIR" some-path)
> ```
> 
> Even if former was doable today, why would you use that verses the much simpler latter?
> 
> 

The advantage is permitting the user more control, in short.

That constant is usually nil. However, if the user happens to have installed Spacemacs to `.spacemacs.d` much trouble arises because they cannot specify an alternative file path for the dotspacemacs file.

With the change in Emacs allowing the use of XDG config directories as the home for init.el, ~/.emacs, or the norm, it seems sensible to be *more* permissive in where Spacemacs' config file is located.

bryce-carson:
I know i have argued for this for a while, but my heart is not particularly in it now. If I absolutely needed it I would simply maintain the necessary changes for my own needs.

In the end, its just another possible option for the configuration framework.

lebensterben:
> That constant is usually nil. However, if the user happens to have installed Spacemacs to .spacemacs.d much trouble arises because they cannot specify an alternative file path for the dotspacemacs file.

I don't know why you say that.

you can always set the variable with custom-set-variables.

bryce-carson:
> > That constant is usually nil. However, if the user happens to have installed Spacemacs to .spacemacs.d much trouble arises because they cannot specify an alternative file path for the dotspacemacs file.
> 
> I don't know why you say that.
> 
> you can always set the variable with custom-set-variables.

`dotspacemacs-filepath`? It is presently defined with `defconst` though.

Could you clarify with the steps in the process as you understand them? If it is simply poorly documented we can address that instead.

lebensterben:
https://github.com/syl20bnr/spacemacs/pull/15484?notification_referrer_id=NT_kwDOABKo0bIzNDkzMTU0NTc0OjEyMjI4NjU&notifications_query=is%3Aunread+reason%3Aparticipating#issuecomment-1510009241

lebensterben:
also note that there's no constant variable in elisp. you can always change a defconst variable.

bryce-carson:
> also note that there's no constant variable in elisp. you can always change a defconst variable.

Okay. Thanks. I reviewed that and now understand better. I had the false belief attempting to assign a value to a const would raise an error.

I will write some docs soon and make the change after getting feedback.

smile13241324:
Given that this is now open since 2 years I would like to go ahead and close it for now.

PR #15492, opened 2022-04-28
doc: vscode-json-languageserver -> vscode-langservers-extracted

sir4ur0n:
Cf https://github.com/emacs-lsp/lsp-mode/issues/3395 the Node package to install has changed.

Installing the wrong Node package leads to `lsp-mode` not detecting the LSP and suggesting to install it automatically.


lebensterben:
thanks!

PR #15494, opened 2022-05-02
Prefix obsolete aliases with "cl-" namespace.

BenedictHW:
Some low hanging fruit.

Warning: ‘case,’ is an obsolete alias (as of 27.1); use ‘cl-case’ instead.

Same for decf and assert.


lebensterben:
thanks!

PR #15496, opened 2022-05-02
[version-control] Load smerge-mode when entering smerge-transient-state

bcc32:
This fixes the following error when pressing `SPC g r n` if
smerge-mode is not loaded yet:

    Symbol's function definition is void: smerge-vc-next-conflict

Removed the `:commands` argument to use-package, which is
erroneous (:commands sets up autoloads so that invoking the command
will load the package specified, but only if the command is not
already fboundp).

Instead, explicitly load smerge-mode when entering the transient
state, so that any of the key bindings in the transient state will be
available.  This has the same intended effect as the previous code.

### DIFF
--- a/layers/+source-control/version-control/packages.el
+++ b/layers/+source-control/version-control/packages.el
@@ -247,7 +247,6 @@
   (use-package smerge-mode
     :defer t
     :diminish smerge-mode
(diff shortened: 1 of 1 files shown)


lebensterben (approved):
LGTM

smile13241324 (approved):

PR #15504, opened 2022-05-10
add org-roam-ui

winsphinx:
For `org-roam-server` doesn't support `org-roam-v2`, `org-roam-ui` is now added to replace it.


smile13241324 (approved):

smile13241324:
Strange I did not see any dangerous code change.

Can you provide an error message or stack trace? @winsphinx do you see the same?

Bost:
There are no error messages or stack trace, however the problem might have something to do with my OS. I use GuixOS. I need to investigate it deeper.

Bost:
The error has most probably something to do with the GuixOS. On Ubuntu everything works fine. Sigh.

PR #15518, opened 2022-05-14
[core] new recipe :fetcher local 

thanhvg:
with this current (package :location local) can be replaced with `(package :location (recipe :fetcher local))` then quelpa will install the local package to common location just like a normal package (eg ~/.emacs.d/elpa/28.1/develop/). We have the benefit of bytecompile and autoload.


smile13241324 (approved):

smile13241324:

PR #15523, opened 2022-05-15
[doc] Remove install instructions for old master

### LINKED ISSUE #14843: Documentation on stable installation is wrong
At the moment `README.md` suggests that cloning this repository will yield to a stable installation (aka not develop). This is an important detail, for example because the develop-installation won't receive auto-updates by default.

From the README:

> For stable releases:
> `git clone https://github.com/syl20bnr/spacemacs ~/.emacs.d`
> For development updates and participation:
> `git clone -b develop https://github.com/syl20bnr/spacemacs ~/.emacs.d`

This is incorrect since `develop` is configured as the default branch on GitHub, yielding a clone without specifying a branch to be `develop`, so basically at the moment, both commands are doing the same.

real-or-random:
Fixes #14843.

This is in line with 964f4c5af3615b648a7707a39b82b7ef02dc4807, whose message states that "we no longer support master releases directly".

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,30 +32,26 @@
 - - -
 
 
-**Quick Install:**
+**Quick Install**
 
 This assumes you don't have an existing Emacs setup and want to run Spacemacs as
 your config. If you do have one, look at
 the [full installation instructions](#install) for other options.
 
-* For stable releases:
-  ```shell
-  git clone https://github.com/syl20bnr/spacemacs ~/.emacs.d
+```shell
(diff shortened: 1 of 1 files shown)


real-or-random:
A next step would be to overhaul the "Update" section (https://github.com/syl20bnr/spacemacs/blob/develop/README.md#update). I think it should focus on `development` and simply have some migrations instructions for users who are still on `master`.

PR #15539, opened 2022-05-22
Add fountain layer, a screenwriting file format

PiOverFour:
This new layer adds support for the [fountain](https://fountain.io/) screenwriting format, using the [fountain-mode](https://github.com/rnkn/fountain-mode) by @rnkn.


rnkn:
Interesting. Thanks for putting the time/effort into this. Not a Spacemacs user myself so doesn't make a huge lot of sense to me but let me know if you need any help.

smile13241324 (approved):

smile13241324:
Thanks for adding this mode to spacemacs, the more tools we are getting integrated the more people will be using emacs and hopefully contribute back to it.

PR #15543, opened 2022-05-25
[ietf] fix auto-mode-alias for ietf-mode

sunlin7:
The RFC document is named as "rfcXYZ.txt" from ietf.org, for example: https://ietf.org/rfc/rfc7269.txt.
So here is the change for the `auto-mode-alist` to correctly associate with rfc document. 
Please help review code and apply to the repo. 
Thanks


sunlin7:
@smile13241324  please help review this change, thanks  :smile:

smile13241324 (approved):

smile13241324:
Thanks for the fix @sunlin7 :)

PR #15547, opened 2022-05-27
[scala] add keybinding for running `It / compile` in SBT

Kazark:
Next logical step after #15372 and #15160 .


smile13241324 (approved):
Looks good

PR #15558, opened 2022-05-31
[Rust] fix dap mode toggle logic

thanhvg:
rust layer doesn't own dap-mode so toggle is ignored


smile13241324 (approved):
Ups my mistake, thanks for fixing.

Issue #15567, opened 2022-06-06
Symbol's value as variable is void: dotspacemacs-startup-banner-scale


Zlopez:
#### Description :octocat:
Spacemacs start in Lisp Interaction mode and throws an error above.

#### Reproduction guide :beetle:
- Start Emacs

*Observed behaviour:* :eyes: :broken_heart:
```
Symbol's value as variable is void: dotspacemacs-startup-banner-scale
```

*Expected behaviour:* :heart: :smile:
Welcome screen is shown

#### Backtrace :paw_prints:
from emacs --debug-init
```
Debugger entered--Lisp error: (void-variable dotspacemacs-startup-banner-scale)
  (eq dotspacemacs-startup-banner-scale 'auto)
  (if (eq dotspacemacs-startup-banner-scale 'auto) (progn (let ((factor (/ (float (spacemacs-buffer//banner-fit-height-size)) height))) (min factor 1))) (let ((factor dotspacemacs-startup-banner-scale)) factor))
  (let* ((title spacemacs-buffer-logo-title) (spec (create-image banner)) (unscaled-size (image-size spec)) (height (cdr unscaled-size)) (scale (if (eq dotspacemacs-startup-banner-scale 'auto) (progn (let ((factor ...)) (min factor 1))) (let ((factor dotspacemacs-startup-banner-scale)) factor))) (size (cons (* scale (car unscaled-size)) (* scale (cdr unscaled-size)))) (width (car size)) (left-margin (max 0 (floor (- spacemacs-buffer--window-width width) 2)))) (plist-put (cdr spec) :scale scale) (insert (make-string left-margin 32)) (insert-image spec) (insert "\n\n") (insert (make-string (max 0 (floor (/ (- spacemacs-buffer--window-width (+ ... 1)) 2))) 32)) (insert (format "%s\n\n" title)))
[...]

Issue #15571, opened 2022-06-07
Spacemacs failure to start correctly even after complete removal and re-installation!


DrWaleedAYousef:
I cannot generate any bug report because spacemacs, all of the sudden, no longer starts correctly even after I fully removed the folder `.emacs.` and re-installed spacemacs. The following message always appear upon starting:
```
Warning (initialization): An error occurred while loading ‘/home/wyousef/.emacs.d/init.el’:

Symbol's value as variable is void: dotspacemacs-startup-banner-scale

To ensure normal operation, you should investigate and remove the
cause of the error in your initialization file.  Start Emacs with
the ‘--debug-init’ option to view a complete error backtrace. Disable showing Disable logging
```

**All my work is pending now. I cannot do anything**

real-or-random:
Duplicate of #15567.

You'll also need to update `~/.spacemacs`. It's not in `~/emacs.d`.

For a quick fix, try `git checkout 51686e5f1206d57a86d9d33a62ed02c9aa9c7e8b` in `~/.emacs.d`

DrWaleedAYousef:
Sorry, what do you mean by updating '.spacemacs'

sunlin7:
PR #15579 will fix this issue.

DrWaleedAYousef:
I added the line in the `.spacemacs`; it is back almost to normal. However, it seems there is a problem with cashing. It never save the latest files and projects opened.

sunlin7:
It sound like a new issue.
Dose the original issue on this ticket was resolved?  Thanks

DrWaleedAYousef:
Yes; but I just reported this here in the case that it is related or connected.

sunlin7:
Please verify with the last version in develop branch.

DrWaleedAYousef:
> Please verify with the last version in develop branch.
```
85edf6a57 * develop origin/develop org-contacts no longer built-in
```

sunlin7:
Here are the version history, especially the `216b98c` would fix this ticket.

d1bdf6c Evilify the pyim-dict-manager-mode-map after its mode be loaded
5a715c8 [pdf] Evilify the pdf-outline-buffer-mode-map after its mode be loa..
216b98c [core] Add variable `dotspacemacs-startup-banner-scale' to scale..
85edf6a org-contacts no longer built-in
a485b5a Make startup banner scalable (from dotfile)

DrWaleedAYousef:
[...]

PR #15574, opened 2022-06-07
Fix #15363: add default value scalable banner

dalanicolai:
This PR should fix issues #15567 and  #15571


dalanicolai:
I have rushed a bit too much here, not even checking what sloppy things I have been doing here. Anyway, PR #15579 by @sunlin7 adds a proper fix, and therefore I am closing this PR.

Issue #15568, opened 2022-06-06
pyim cause an error:  variable is void: pyim-dict-manager-mode-map


winsphinx:
#### Description :octocat:
pyim cause an error:
`Error (use-package): pyim/:init: Symbol’s value as variable is void: pyim-dict-manager-mode-map`

#### Reproduction guide :beetle:
- Start Emacs
- The error appears

*Observed behaviour:* :eyes: :broken_heart:
<<DESCRIPTION OF THE OBSERVED BEHAVIOUR>>

*Expected behaviour:* :heart: :smile:
<<DESCRIPTION OF THE EXPECTED BEHAVIOUR>>

#### System Info :computer:
- OS: windows-nt
- Emacs: 28.1
- Spacemacs: 0.999.0
- Spacemacs branch: develop (rev. 85edf6a57)
- Graphic display: t
- Running in daemon: nil
- Distribution: spacemacs
- Editing style: hybrid
- Completion: ivy
- Layers:
```elisp
(
 ......
 (chinese :variables chinese-enable-youdao-dict t pangu-spacing-real-insert-separtor t pyim-page-tooltip
          '(posframe popup minibuffer))
 ......         
```
- System configuration features: ACL GIF GMP GNUTLS HARFBUZZ JPEG JSON LCMS2 LIBXML2 MODULES NATIVE_COMP NOTIFY W32NOTIFY PDUMPER PNG RSVG SOUND THREADS TIFF TOOLKIT_SCROLL_BARS XPM ZLIB

#### Backtrace :paw_prints:
```
<<BACKTRACE IF RELEVANT>>
```

tshu-w:
The `pyim-dict-manager`  need to require manually after 22-05-29, see https://github.com/tumashu/pyim/blob/2f1aaaa9a7f0e6640e5bafb66857a9a0f49e80f8/README.org#2022-05-29-sun-pyim-cregexp-utils-pyim-cstring-utils-和-pyim-dict-manager-需要用户手动-require

winsphinx:
It is already autoload'ed in https://github.com/syl20bnr/spacemacs/blob/85edf6a578c81f12537ac0b1d8b98324625e4275/layers/%2Bintl/chinese/packages.el#L62-L63

sunlin7:
It's my failed, and the PR https://github.com/syl20bnr/spacemacs/pull/15577 will solve the problem.

winsphinx:
[...]

PR #15577, opened 2022-06-07
Evilify the pyim-dict-manager-mode-map after its mode be loaded

sunlin7:
Try to fix the issue https://github.com/syl20bnr/spacemacs/issues/15568.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1628,6 +1628,8 @@ Other:
   - Added =fcitx5= support (thanks to wgjak47)
 - Fixes:
   - Make =fcitx.el= work by default (thanks to AmaiKinono)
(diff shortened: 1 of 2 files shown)


sunlin7:
Hi @smile13241324 , this PR can resolve the #15568, and verified by the reporter.
Please help merge it, thank you.

dalanicolai:
@smile13241324 @sunlin7 We should probably change the text in `changelog.develop` to something like "changes implemented since/after version 0.200"...

smile13241324 (approved):

smile13241324:
> @smile13241324 @sunlin7 We should probably change the text in `changelog.develop` to something like "changes implemented since/after version 0.200"...

Jep good idea, PR would be welcome.

PR #15581, opened 2022-06-08
Fix vc-svn/git void variable warning (evilify after load)

dalanicolai:
Fix two more startup 'void variable' warnings, introduced (probably) after PR #15354.

For explanation, see PR message of #15544.


Kazark:
If I understand correctly, this may fix a problem I've been having. Thanks.

smile13241324:
I think this has already been fixed, please double check and reopen if still necessary @dalanicolai.

dalanicolai:
Ah yeah, well PR #15584 did not exist back then of course... but good that it has been fixed!

PR #15590, opened 2022-06-10
Allow setting flycheck buffer location on screen.

greydot:
Hello!

This PR adds a few config variables for syntax-checking layer, allowing to set the window location and size.

That's it. Nothing fancy :)


smile13241324:
Merged, thanks for the addition :purple_heart:

PR #15595, opened 2022-06-14
[core] fix functions missed for byte-compiled core-spacemacs-buffer.el

sunlin7:
The function `spacemacs-buffer/jump-to-recent-files` is used for jumping to the recent files by press `r` in Spacemacs home buffer.
Try to byte-compile the file `core-spacemacs-buffer.el`,
```
(byte-compile-file (concat spacemacs-core-directory "core-spacemacs-buffer.el"))
```
Restart the Spacemacs, then the function `spacemacs-buffer/jump-to-recent-files` is missed, press `r` in Spacemacs home buffer will not work, describe-function will not find the function.

This change fixed the issue with giving correct lisp for byte-compiler (avoid using the `eval` in lexical scope).

### DIFF
--- a/core/core-spacemacs-buffer.el
+++ b/core/core-spacemacs-buffer.el
@@ -838,16 +838,16 @@ in for example the `view-lossage' (C-h l) buffer:
 instead of:
  r                      ;; anonymous-command
  p                      ;; anonymous-command"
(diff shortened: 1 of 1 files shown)


sunlin7:
@smile13241324 Please help to review this change, thanks  : )

smile13241324 (approved):

PR #15614, opened 2022-06-24
[defaults] Exclude `custom-file` correctly in recentf

real-or-random:
Entries of recentf-exclude must be expanded, e.g., `~` for the home
directory won't work.


real-or-random:
Before this PR, `~/.emacs.d/.cache/.custom-settings` ends up in `recentf-exclude` (the default value).
After this PR, it's correctly expanded to `/home/user/.emacs.d/.cache/.custom-settings`.

smile13241324 (approved):

PR #15625, opened 2022-06-30
[editing] lazy binding for string-edit-mode

thanhvg:
or it will break `C-h k`


smile13241324 (approved):

rommeswi:
On emacs 29, C-h k is still broken due to void string-edit-mode.

lebensterben:
I don't understand how `C-h k` is broken with or without this change. It works perfectly fine.

PR #15631, opened 2022-07-04
Undo-region: more ergonomic keybindings

BenedictHW:
As was the case with `vim-style-remap-Y-to-y$', or rebinding "s" to
evil-surround over vim's default substitute, by rebinding "u" to undo-region
is a sensible trade-off given that prior to this, undo-region was accessible
only through SPC SPC C-_.

See https://emacs.stackexchange.com/questions/32244/undo-region-spacemacs

Comments and suggestions?

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -127,6 +127,8 @@ the [[file:CHANGELOG.org][CHANGELOG.org]] file.
     (thanks to Seong Yong-ju)
   - Changed ~SPC t t m~ from ~timeclock-modeline-display~ which was deprecated
     in Emacs 24.3 to ~timeclock-mode-line-display~ (thanks to Zach Pearson)
+  - Added =vim-style-enable-undo-region= style variables to enable undo-region
+    in Vim editing style; disabled by default
 - Other:
   - Support for multiple cursors using =evil-mc= is now encapsulated in the
     =multiple-cursors= layer (thanks to Codruț Constantin Gușoi)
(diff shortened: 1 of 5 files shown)


lebensterben:
Instead of rebinding the key, you can avoid changing evil's undo system:

https://github.com/BenedictHW/spacemacs/blob/ffe97d0da1f7e941ea5b4f4a67753a885e160692/layers/%2Bdistributions/spacemacs-bootstrap/packages.el#L83-L85

BenedictHW:
Okay, I've separated it into 3 commits. I do not wish to change
"evil-set-undo-system" from undo-tree but enable undo-region with
undo-tree.

1. 82e3b38dd Just enables undo-region.
2. fe69236d5 Link to undo-tree in spacemacs-editing layer readme
3. 153de94e3 I find this a common scenario: I undo multiple times to yank some
text and then redo multiple times to get back to my working state and
then paste the text. Much easier to take advantage of Emacs advanced
undo system and just undo-region.

In spirit with not overriding vim defaults, I've left it as a flag like
"vim-style-remap-Y-to-y$" but unlike vim-surround on s. Arguably more
useful than downcasing as "~" also changes casing in default vim I believe.

I leave it up to you guys if you want to cherry pick 2 of the 3 or some
other combination thereof.

Thanks for maintaining Spacemacs,

lebensterben:
I pushed a few small changes.
If you're okay with the current status I will merge it.

BenedictHW:
Of course. Cheers!

PR #15634, opened 2022-07-04
[spacemacs-bootstrap] Remove whitespace from kill ring.

BenedictHW:
More discussion at https://github.com/syl20bnr/spacemacs/discussions/15455

Is it alright if the advising function is anonymous?


lebensterben (approved):

lebensterben:
Just pushed a few changes.

- I replaced `advice-add` by `define-advice`. They're similar except the latter provides an easier way to define a *named* advice function (with function docstring as well).
- `(message ..)` returns the message. To prevent any unexpected behaviour I make sure it returns a `nil`.

BenedictHW:
Thank you, I did not know about define-advice.

PR #15635, opened 2022-07-04
Enable tws mode also in text-mode in addition to prog-mode.

### LINKED ISSUE #15621: Highlight trailing whitespace in text-mode instead of only in prog-mode
Currently we do this only in `prog-mode`:
https://github.com/syl20bnr/spacemacs/blob/c9db028421a74eb8339152ae3c2fe694dc322216/layers/%2Bspacemacs/spacemacs-defaults/packages.el#L513

The config option just describes it as "Show trailing whitespace":
https://github.com/syl20bnr/spacemacs/blob/c9db028421a74eb8339152ae3c2fe694dc322216/core/templates/.spacemacs.template#L495-L496

I think people who want to have this functionality, typically would also want to have it in `text-mode`. This will be consistent with line numbers, which are also enabled in `prog-mode` and in `text-mode`. But I wanted to ask here first before opening a PR, not sure if everyone agrees.

In any case, the docs could be improved. They should explain in which modes this is enables (similar to the docs for line numbering, which clearly state the modes:

https://github.com/syl20bnr/spacemacs/blob/c9db028421a74eb8339152ae3c2fe694dc322216/core/templates/.spacemacs.template#L406-L408

emacs18:
Color highlighting of trailing white spaces is currently enabled only in prog-mode derived modes.  This is changed so that it is also enabled in all text-mode derived modes.  Also documentation for the user variable was updated.

This change should fix https://github.com/syl20bnr/spacemacs/issues/15621

### DIFF
--- a/core/templates/.spacemacs.template
+++ b/core/templates/.spacemacs.template
@@ -486,8 +486,10 @@ It should only modify the values of Spacemacs settings."
    ;; (default nil - same as frame-title-format)
    dotspacemacs-icon-title-format nil
 
-   ;; Show trailing whitespace (default t)
-   dotspacemacs-show-trailing-whitespace t
+   ;; Color highlight trailing whitespace in all prog-mode and text-mode derived
+   ;; modes such as c++-mode, python-mode, emacs-lisp, html-mode, rst-mode etc.
+   ;; (default t)
+   dotspacemacs-show-trailing-whitespace t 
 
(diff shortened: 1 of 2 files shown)


real-or-random (approved):
Thanks, as I mentioned in the linked issue, I think this is an improvement overall.

lebensterben (approved):

PR #15637, opened 2022-07-06
Add keybinding for projectile-run-project

gbastkowski:
Just a new keybinding that people might find helpful.

Not sure if pC (like Compile-and-run) or px (like eXecute) would be the better choice here. I found C more intuitive.


lebensterben:
https://github.com/bbatsov/projectile/blob/00fce5a62cab56873c37629bef653bb2a1e4a4c1/projectile.el#L5711

`pu` is better for its consistency with upstream.

PR #15657, opened 2022-07-22
Update README.org

mikefarmer01:
Provide more info on plantuml-mode execution modes (mode `executable` in particular)


lebensterben:
Thanks.

PR #15662, opened 2022-07-25
fix: dotspacemacs-directory is expected to end with a slash

kassick:
Commit 861ea131e1be7963b9602e774024f687fe9d4a29 refactored the definition of
dotspacemacs-directory, but did not ensure that the path ended with a slash,
which is expected elsewhere in the code

This prevented spacemacs from loading the environment file and the user's custom layers


lebensterben:
thanks

PR #15675, opened 2022-08-03
Added documentation to set julia envornment location for lsp server

dCaples:
lsp-julia-default-envornment is set by default to ~/.julia/environments/v1.0 .
To make the lsp server show completions for installed packages, the environment
location must be the same as that of the installed julia version enviornment.
Documentation was added to notify users to set the emacs variable to their
enviornment location.

### DIFF
--- a/layers/+lang/julia/README.org
+++ b/layers/+lang/julia/README.org
@@ -46,10 +46,17 @@ for installation, and then install this layer with:
      (julia :variables julia-backend 'lsp)))
 #+END_SRC
 
+
 =LanguageServer.jl= tends to have a very long startup time. In the worst case,
 =lsp-mode= might give up on the language server before its started, but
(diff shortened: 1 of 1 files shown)


lebensterben:
I suggest to add a reference link to Julia's language server on the environment variable.

smile13241324 (changes requested):
Looks ok still a link to the respective emacs-lsp config page and the projects page of the lsp should be added, to notify users about other config vars for this particular language server.

dCaples:
ok, I implemented your suggestion! I didn't include the [emacs-lsp config page](https://emacs-lsp.github.io/lsp-mode/page/lsp-julia/) because it didn't really contain any information, and the links in it were already included in the docs. I can include it if you still think it would still be useful.

lebensterben (approved):

PR #15684, opened 2022-08-06
[evil] Use the default evil-lion binding instead

### LINKED ISSUE #15501: consider switching or adding the default `evil-lion` bindings: `gl` and `gL`
From commit 8281ed7, it showed that it's not exactly clear what the default bindings `gl` and `gL` (`evil-lion-left` and `evil-lion-right` respectively) were colliding with. from what i can tell from discussion in #8555 however, is that it seemed to collide with markup binding `gl` ([move down one level in headings](https://github.com/syl20bnr/spacemacs/blob/master/doc/CONVENTIONS.org#movement-in-normal-mode)) but i don't think `evil-lion` have much uses in markup files. imo overriding bindings `gl` and `gL` from `evil-lion`'s to markup file behaviour, *only in markup files* make more sense (which is what `evil-org` and `markdown/init-markdown-mode` already do).

i would like to ask if the maintainers can reconsider the default bindings by either deprecating the current bindings `SPC x a l` and `SPC x a L` or instead simply adding `gl` and `gL` in non-markup files (retaining the previous `SPC` bindings). if wanted, i can make a PR for this. i believe a simple `(evil-define-key '(normal visual) 'global "gl" #'evil-lion-left ...)` should be fine, but i didn't test further than JS, org and markdown files.

iliekprogrammar:
Fixes #15501.

### consideration

- retain the old bindings, `SPC x a l` and `SPC x a L`. if requested, i'll readd this.


smile13241324 (approved):
Looks good, thanks for the contribution :)

PR #15692, opened 2022-08-10
core: support native-compile the "core/*.el" files

sunlin7:
Hi,
There are many errors when I try to byte-compile all the `<spacemacs>/core/*.el` files into `*.elc` files, and it will automatically do native-compile for my local emacs has native-compile feature on.
This PR will fix the errors when running the byte-compile and native-compile the `core/*.el` files.
Please help review it. Thanks.

Just run follow line to compile the `core/` directory.
```
(byte-recompile-directory spacemacs-core-directory 0)
```

PR #15696, opened 2022-08-10
core: byte-compile the "core/libs/" folder when dotspacemacs-byte-compile is t

sunlin7:
Hi,

Currently `dotspacemacs-byte-compile` is `t` will only compile serveral files in `core/lib/*`.
After the PR #15692, all files in `core/` will be able byte-compile/native-compile.

So we'll byte compile the `core/` directory when `dotspacemacs-byte-compile` is `t`.

### DIFF
--- a/core/core-compilation.el
+++ b/core/core-compilation.el
@@ -23,6 +23,7 @@
 
 (require 'cl-lib)
 (require 'subr-x)
+(require 'bytecomp)
 
 (defvar spacemacs--last-emacs-version ""
   "This variable is set during Emacs initialization to its version.")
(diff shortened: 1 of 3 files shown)


lebensterben:
I prefer not to make haste on byte compiling everything for now.

sunlin7:
Hi @lebensterben 

What about byte-compiling the directory `core/libs/`?  I'll change it to work in `core/libs/` directory.

lebensterben:
it's okay to byte compile core/libs. but don't modify any of those files because they are 3rd party packages.

sunlin7:
Hi @lebensterben 

I had modified the code to byte-compile the `core/libs/` folder when the `dotspacemacs-byte-compile` is `t`.

Please help review and approve this change. Thanks.

smile13241324 (approved):

sunlin7:
Thank you all !

PR #15697, opened 2022-08-11
core: better way to locate the spacemacs-start-directory

sunlin7:
Hi,
There is a better way to locate the `spacemacs-start-directory`, in the change.

I didn't always put spacemacs in `~/.emacs.d`, then we should locate the `spacemacs-start-directory` based on its file path.

This change is part of #15692 .

@smile13241324 Please help review and approve this change. Thanks.

### DIFF
--- a/core/core-load-paths.el
+++ b/core/core-load-paths.el
@@ -34,7 +34,7 @@
 
 ;; ~/.emacs.d
 (defvar spacemacs-start-directory
-  (expand-file-name user-emacs-directory)
+  (concat (file-name-directory (or load-file-name buffer-file-name)) "../")
   "Spacemacs start directory.")
 
 ;; ~/.emacs.d/assets


lebensterben:
See https://github.com/syl20bnr/spacemacs#alternative-install-location

sunlin7:
Hi @lebensterben 
This change fix a bug that `spacemacs-start-directory` got wrong value when native compile the `core/core-load-paths.el`.

The installization folder is `~/.spacemacs-devel/`, and the `~/.emacs.d/` only has a customized `init.el`, it will set the `spacemacs-start-directory` and load the real init.el in the `~/.spacemacs-devel` folder.

After compile the `core-load-paths.el` to its `*.elc` form, the emacs will do `native-compile-async` in background, in the native compile routine, emacs will use initilization value  `~/.emacs.d/` for `spacemacs-start-directory`, but the desired value is `~/.spacemacs-devel/`.

When the `core-load-paths.el` is loaded, assume its parent folder to be the spacemacs start directory is more reasonable than assume it's the `~/.emacs.d/` (in previous case the `~/.emacs.d/` don't contain the Spacemacs installization files) .

lebensterben:
> This change fix a bug that spacemacs-start-directory got wrong value when native compile the core/core-load-paths.el.

What is the bug?

sunlin7:
Hi @lebensterben 
The bug is it will try to search the ~/.emacs.d/ for the library *.el files on async native compiling the *.el files.

I can wait your enhance change that you mentioned in  #15692, 
then will to do native compile core/ folder again, and report the issues if there still has error message.

lebensterben:
can you show me the exact the error?

PR #15708, opened 2022-08-20
unicode-fonts doc: hint at supported ligatures

real-or-random:
This caught me. The default config enables ligatures that are not supported by my font and then strange things happen when emacs tries to use the non-existing ligatures. 

Maybe the default config can be tuned to be somewhat more conservative but this can be done in a different PR independently of this doc change.

### DIFF
--- a/layers/+fonts/unicode-fonts/README.org
+++ b/layers/+fonts/unicode-fonts/README.org
@@ -70,8 +70,13 @@ Or to enable ligatures only for text files:
 #+END_SRC
 
 To configure the ligature set that gets used with your font there is a
-=unicode-fonts-ligature-set= variable. For example To only enable the ligatures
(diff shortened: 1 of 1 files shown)


lebensterben:
what happens if you set unsupported ligatures?

real-or-random:
> what happens if you set unsupported ligatures?

Hm, I thought I get very strange rendering artifacts when the default set of ligatures is enabled, e.g., for `!= ##`, where here both ligatures are  enabled and misrendered (but only `!=` is supported by Iosevka)
image

But after more testing, I think this setting was a red herring. At the moment I can't really tell what's causing this bug, and it happens randomly and even in some starts of emacs only in bold text etc. I'll probably report to upstream. Closing for now.

PR #15718, opened 2022-09-02
Fix `spacemacs/emacs-with-native-compilation-enabled-p`

bryce-carson:
The built-in function was apparently renamed. See the Emacs' ChangeLog.3:

> native-comp-available-p is the definitive test
> — 2021-10-05  Stephen Gildea  <stepheng+emacs@gildea.com>


lebensterben:
merged. thanks.

Issue #15724, opened 2022-09-06
org-roam-setup cause race condition with user config file


emacs18:
Following is part of `org-roam` setup within spacemacs:

    (defun org/init-org-roam ()
      (use-package org-roam
        :defer t
        :hook (after-init . org-roam-setup)
        :init
        ...))

`org-roam-setup` is an obsolete function as indicated by these lines from [org-roam](https://github.com/org-roam/org-roam/blob/main/org-roam-compat.el#L183):

    (define-obsolete-function-alias
      'org-roam-setup
      'org-roam-db-autosync-enable "org-roam 2.0")

The damage done by adding `org-roam-setup` is that automatic database update mode is enabled. This cause `org-roam-db-sync` function to be called which will then start to update sqlite database of all my org files. The problem is that `org-roam-db-sync` is called **before my init file is loaded**! Thus `org-roam-directory` has not yet been setup properly. `org-roam-db-sync` thus uses incorrect directories and happily updates the sqlite db.

After emacs comes up if I now execute `org-roam-db-sync` manually, then it thinks that all files are out of date, because `org-roam-directory` has now changed! Thus sqlite db is updated for all my files again. At this point if I restart emacs, then the cycle repeats all over again where sqlite db for all my org files are updated again. This can add up to 8 minute delay in my emacs start up time! For some unknown reason sqlite db update for some files take very long time like two minutes for some files.

What I would like to ask is that we not enable automatic update of `org-roam` database and let the user choose when this is enabled, i.e., do not add `org-roam-setup` to `after-init-hook`.

lebensterben:
I do not use org-roam (except for debugging..). Can you show me what user might need to config in Spacemacs/user-config?

emacs18:
`(org-roam-db-autosync-mode)` enables the minor mode which updates sqlite db of org files as they are modified.
This is mentioned in the link that you mentioned above.

lebensterben:
one solution is to tell users to call `org-roam-db-autosync-mode` AFTER any user configuration.

another solution is to tell users to add all user configurations in 

```elisp
(spacemacs|use-package-add-hook org-roam
  :pre-config
  ;; user configuration
  )
```

and we call `org-roam-db-autosync-mode` inside the regular `:config`.

What do you think?

emacs18:
My preference would be to simply tell to call the function in question to enable the minor mode.
This is the documented method in org-roam manual.

[...]

PR #15725, opened 2022-09-06
do not add org-roam-setup to after-init-hook

emacs18:
This is to prevent race condition between org-roam db update verse org-roam setup. See https://github.com/syl20bnr/spacemacs/issues/15724


lebensterben:
thanks

PR #15736, opened 2022-09-15
Fix cumbersome pdf-tools selection behavior

dalanicolai:
Currently, due to evil entering visual state, the selection behavior in pdf-tools is different from and more cumbersome than in vanilla Emacs (see https://github.com/emacs-evil/evil/issues/1671).

This commit prevents evil from entering evil-state in pdf-mode and moves the (only) visual-state yank binding to the evilified-map.

It doesn't look pretty, but it is working much better than before.

### DIFF
--- a/layers/+readers/pdf/packages.el
+++ b/layers/+readers/pdf/packages.el
@@ -31,6 +31,12 @@
     :mode (("\\.pdf\\'" . pdf-view-mode))
     :init
     (spacemacs//pdf-tools-setup-transient-state)
+
+    (add-hook 'pdf-view-mode-hook
+              (lambda () (add-hook 'evil-evilified-state-entry-hook
+                                   (lambda (remove-hook 'activate-mark-hook 'evil-visual-activate-hook t))
+                                   nil t)))
+
     :config
     (progn
       (pdf-tools-install)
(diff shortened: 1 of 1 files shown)


smile13241324 (approved):
Sounds sensible, I'll merge it.

sasinhe:
Did this change work? I've opened a issue inside pdf-tools about the same strange behaviour a while ago ([https://github.com/vedang/pdf-tools/issues/137](link)).
However when adding this snippet of code to my config it doesn't solve the error.
```
(add-hook 'pdf-view-mode-hook
              (lambda () (add-hook 'evil-evilified-state-entry-hook
                                   (lambda (remove-hook 'activate-mark-hook 'evil-visual-activate-hook t))
                                   nil t)))
```

dalanicolai:
@sashine The change works over here already, even though I forgot to add an argument list in the second lambda (strangely enough). Anyway, I've created another PR to add the argument list. I think you should refer to the fix here in https://github.com/vedang/pdf-tools/issues/137. Try it again in Doom, and if it works you can also create a PR for Doom.

As this behavior can simply get fixed by removing that hook, I am not sure if this has to get fixed in evil. But I have reopened [the issue there](https://github.com/emacs-evil/evil/issues/1671), because I think it is better if they decide about the best solution.

dalanicolai:
The solution only starts to work after the first time returning from visual state. However, removing the hook function directly via the `pdf-view-mode-hook` obviously (also) does not work. It would be great to find a more appropriate hook than the `evil-evilified-state-entry-hook`, but I can't come up with any better option at the moment.

lebensterben:
alternatively we can just revert these commits...

dalanicolai:
Well, I guess there exists no single person that prefers the previous behavior. If someone does prefer it, then that user can still enjoy it until making a selection for a second time ;)

Issue #15734, opened 2022-09-14
Killing *scratch* on startup breaks *spacemacs* recent files and projects


rommeswi:
#### Description :octocat:
I never use the scratch buffer so I wanted to kill it on startup. Unfortunately, this breaks the *spacemacs* buffer as it no longer shows the recent files and projects.

 Killing the message buffer on startup does not have this effect.

#### Reproduction guide :beetle:
- Place `(kill-buffer "*scratch*") ` into `.spacemacs` file under `user-config`
- Restart emacs

*Observed behaviour:* :eyes: :broken_heart:
- The `*spacemacs*` buffer no longer shows recent files and projects.
- The loading bar at the bottom also gets stuck.

*Expected behaviour:* :heart: :smile:
Spacemacs closes the `*scratch*` buffer but I can still use recent files and projects.

drlkf:
Not a contributor, but if I may suggest trying:

```emacs-lisp
(advice-add 'spacemacs-buffer/goto-buffer
            :after (lambda (&rest args)
                     (let ((buffer (get-buffer "*scratch*")))
                       (when buffer (kill-buffer buffer)))))
```

rommeswi:
Thank you for your suggestion. It works but creates the error "Warning: Unused lexical argument `args'" when compiling the dotfile.

I also think that the *spacemacs* buffer should somehow work independently of the *scratch* buffer.

lebensterben:
spacemacs buffer doesn't seem to use scratch buffer.

but the setup requires it:

https://github.com/syl20bnr/spacemacs/blob/c5c69ce239577ce1e0fb468698d7aeb2453a6cbb/core/core-spacemacs.el#L242-L245

rommeswi:
Thank you both for your helpful replies.

For now, I simply set `dotspacemacs-scratch-mode` to nil which allows me to kill the buffer just normally.

Perhaps one could insert the additional check
`(when (get-buffer "*scratch*")
  (with-current-buffer "*scratch*"
.....
))`

Perhaps it would even make sense to create a configuration option to run spacemacs without a scratch buffer. Or to kill the scratch buffer if `dotspacemacs-scratch-mode` is nil. But I don't know what the policy is on adding/changing configuration options.

This issue can be closed in case people find that no change is necessary.

rommeswi:
The above pull request solved the issue.

PR #15742, opened 2022-09-19
Add check that *scratch* buffer exists

rommeswi:
Fix bug that setting the dotspacemacs-scratch-mode can break if no scratch buffer exists.

### DIFF
--- a/core/core-spacemacs.el
+++ b/core/core-spacemacs.el
@@ -240,9 +240,10 @@ Note: the hooked function is not executed when in dumped mode."
      (run-hooks 'spacemacs-post-user-config-hook)
      (setq spacemacs-post-user-config-hook-run t)
      (when (fboundp dotspacemacs-scratch-mode)
-       (with-current-buffer "*scratch*"
-         (funcall dotspacemacs-scratch-mode)
-         (run-hooks 'spacemacs-scratch-mode-hook)))
+       (when (get-buffer "*scratch*")
+         (with-current-buffer "*scratch*"
+           (funcall dotspacemacs-scratch-mode)
+           (run-hooks 'spacemacs-scratch-mode-hook))))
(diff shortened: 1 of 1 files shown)


rommeswi:
Fix for #15734

lebensterben:
it can be fixed by `get-or-create-buffer`

lebensterben:
I'm not yet sure whether skipping spacemacs-scratch-mode-hook is safe.

lebensterben:
currently this hook seems to be used only here https://github.com/syl20bnr/spacemacs/blob/ba33309327e278a5fe7c12e095ed3a314523484f/layers/%2Bspacemacs/spacemacs-editing/packages.el#L636

If this commit doesn't break persistent scratch then it's okay.

rommeswi:
I tested for whether persistent scratch mode still works.

If one does not kill the scratch buffer on startup, persistent scratch mode works just normally.

Whenever one kills the scratch buffer, it starts next time as an empty buffer. So if one kills the scratch buffer on startup and then opens the scratch buffer manually, it is empty and requires a persistent-scratch-restore.

I think this is expected behavior.

sunlin7:
It's not a Spacemacs bug or Emacs bug. 
It should consider as a local configuration issue, wrong configuration create this issue, should change local configurations to fix the issue.

rommeswi:
I am not sure I understand you correctly. Are you saying that
`dotspacemacs-scratch-mode nil` in `dotspacemacs/init` and
`(kill-buffer "*scratch*")` in `dotspacemacs/user-config`
is the intended way of configuring spacemacs to startup without `*scratch*`?

The fix allows that `(kill-buffer "*scratch*")` in the user configuration has the intended effect on startup. This also works in plain emacs.

smile13241324:

PR #15744, opened 2022-09-22
fix: Refactor code-cells-mode setup to avoid undefined symbol errors on define-key

kassick:
Commit 54f12b66f correctly moved code-cells setup to use-package's :init to
ensure that python-mode had a hook for code-cells-mode -- otherwise, code-cells
would never be loaded automatically with python major mode.

A side-effect of this change was that now there was a minor mode keymap for
`code-cells-mode` , which may not have been loaded unless emacs has already
opened a python file.

When calling `describe-key` , the function `help--binding-locus` would iterate
over `minor-mode-map-alist` and, when it reached the entry for code-cells-mode,
a call to `(symbol-value 'code-cells-mode)` would fail with a `void-variable`
error. Despite `code-cells-mode` being an auto-loadeable symbol, `symbol-balue` would not trigger the autoload.

The solution (as pointed by @sunlin7) was to call `add-hook` on `:init` and
`spacemacs/set-leader-keys-for-minor-mode` on `:config`

### DIFF
--- a/layers/+lang/python/packages.el
+++ b/layers/+lang/python/packages.el
@@ -101,14 +101,12 @@
     :if (not (configuration-layer/layer-used-p 'ipython-notebook))
     :defer t
     :commands (code-cells-mode)
-    :init
-    (progn
-      (add-hook 'python-mode-hook 'code-cells-mode)
-      (spacemacs/set-leader-keys-for-minor-mode 'code-cells-mode
-        "gB" 'code-cells-backward-cell
-        "gF" 'code-cells-forward-cell
-        "sc" 'code-cells-eval
-        "sa" 'code-cells-eval-above))))
+    :init (add-hook 'python-mode-hook 'code-cells-mode)
+    :config (spacemacs/set-leader-keys-for-minor-mode 'code-cells-mode
+            "gB" 'code-cells-backward-cell
+            "gF" 'code-cells-forward-cell
+            "sc" 'code-cells-eval
+            "sa" 'code-cells-eval-above)))
 
 (defun python/post-init-company ()
   ;; backend specific


kassick:
Hey @smile13241324 please advise if you see a better way of solving this error.

I tried adding `code-cells-mode` as `:defines` and `:functions` in `use-package` to no avail

lebensterben:
this doesn't make sense to me.

:commands should have autoloaded it (if it's not already)

kassick:
Indeed it should -- and yet it fails somehow :/

Captura de tela de 2022-09-22 20-13-42

(already byte-recompiled everything)

#### System Info :computer:
- OS: gnu/linux
- Emacs: 28.1
- Spacemacs: 0.999.0
- Spacemacs branch: develop (rev. bb5ba70a9)
- Graphic display: t
- Running in daemon: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(csv html
     (unicode-fonts :variables unicode-fonts-enable-ligatures t)
     (ibuffer :variables ibuffer-group-buffers-by ’projects)
     nav-flash theming helm
     (auto-completion :variables auto-completion-front-end ’company auto-completion-enable-snippets-in-popup t auto-completion-enable-help-tooltip nil auto-completion-use-company-posframe t)
     multiple-cursors
     (git :variables git-enable-magit-gitflow-plugin t magit-bury-buffer-function ’magit-mode-quit-window)
     (version-control :variables version-control-diff-tool ’git-gutter version-control-global-margin nil version-control-diff-side ’left)
     (spell-checking :variables spell-checking-enable-by-default nil spell-checking-enable-auto-dictionary t)
     dtrt-indent
     (lsp :variables lsp-keymap-prefix "C-c C-l" lsp-ui-doc-enable nil lsp-ui-doc-position ’at-point lsp-ui-doc-max-height 15 lsp-use-upstream-bindings nil)
     emacs-lisp shell-scripts
     (python :variables lsp-pyright-venv-path "/home/kassick/.pyenv/versions" python-auto-set-local-pyenv-version nil python-test-runner ’pytest python-backend ’lsp python-lsp-server ’pyright python-spacemacs-indent-guess nil python-formatter ’yapf)
     rust
     (c-c++ :variables c-c++-backend ’lsp-clangd)
     cmake javascript
     (typescript)
     (groovy :variables groovy-lsp-jar-path "/home/kassick/Sources/user/groovy-language-server/build/libs/groovy-language-server-all.jar")
     terraform yaml
     (markdown :variables markdown-live-preview-engine ’vmd)
     org
     (latex :variables latex-enable-folding t latex-enable-auto-fill nil)
     (bibtex)
     (shell :variables shell-default-height 30 shell-default-position ’bottom)
     kzk-general kzk-evil kzk-window-management kzk-projectile kzk-code-tweaks kzk-mc kzk-embrace kzk-magit kzk-org kzk-lsp kzk-makefile kzk-python kzk-ahs kzk-visual-editing kzk-helm kzk-company kzk-persp spacemacs-editing)
```
- System configuration features: ACL CAIRO DBUS FREETYPE GIF GLIB GMP GNUTLS GSETTINGS HARFBUZZ JPEG JSON LIBOTF LIBSELINUX LIBSYSTEMD LIBXML2 M17N_FLT MODULES NATIVE_COMP NOTIFY INOTIFY PDUMPER PNG RSVG SECCOMP SOUND THREADS TIFF TOOLKIT_SCROLL_BARS X11 XDBE XIM XPM XWIDGETS GTK3 ZLIB

layers/+lang/python/packages.el
```diff
@@ -100,8 +100,7 @@
   (use-package code-cells
     :if (not (configuration-layer/layer-used-p 'ipython-notebook))
     :defer t
-    :commands (code-cells-mode)
-    :init
+    :config
     (progn
       (add-hook 'python-mode-hook 'code-cells-mode)
```
lebensterben: moving this line to :config is problematic. it won't be added to the hook until it's loaded, which means you need to manually load it.

kassick: But having on :init causes `describe-key` to break ... :/

The issue seems to be that `(symbol-value 'code-cells-mode)` does not triggers an auto-load for code-cells

For instance: this fails in the `(message (symbol-value mode))` statement
```
(mapcar
                                       (lambda (mode-and-map)
                                         (message "mode and map: %S" mode-and-map)
                                         (let ((mode (car mode-and-map)))
                                           (message "trying to get value for %S" mode)
                                           (message "symbol-value: %S" (symbol-value mode))
                                           (when (symbol-value mode)
                                             (intern-soft
                                              (format "%s-map" mode)))))
                                       minor-mode-map-alist)
```

lebensterben: what I really don't understand is that code-cells-mode is already autoloaded at current version

https://github.com/astoff/code-cells.el/blob/fd68a33eb43b3cbd44fed767f48e230382903592/code-cells.el#L303

what if we remove the `:command (code-cells-mode)` and keep other lines unchanged?

lebensterben:
also can you describe how to reproduce this behavior?

kassick:
> also can you describe how to reproduce this behavior?

- Enable the python layer
- Open emacs
- Use `C-h k` (`describe-key`) to get help on some key

Expected result: the help buffer describing the function mapped to the specific key comination
Result: A message complaning about undefined symbol

kassick:
If I `(require 'code-cells)` , everything works as expected
If I `M-x code-cells-mode` , the mode is autoloaded and enabled

lebensterben:
the code-cells-mode is always autoloaded even without `:command (code-cells-mode)`

what's strange is why describe-key will be affected:

- If code-cells-mode is not loaded, it should not affect the keymap
- If it is loaded, the symbol should not be undefined......

kassick:
Indeed @lebensterben -- it appears to be a contradiction

sunlin7:
Same error in my local.

The function `(spacemacs/set-leader-keys-for-minor-mode ...)` desire the `code-cells-mode` to be already loaded.

How about move the `(add-hook 'python-mode-hook 'code-cells-mode)` to the`:init`, and move the `(spacemacs/set-leader-keys-for-minor-mode...)` to the `:config` sessions?

lebensterben:
maybe spacemacs/set-leader-keys-for-minor-mode is the culprit here...

kassick:
@sunlin7 hint did it!
@lebensterben I've updated the PR

smile13241324 (approved):
looks ok

PR #15745, opened 2022-09-24
Create visual selection from transient paste

rommeswi:
Improve "Pasting Transient State" by adding a shortcut C-v to generate a visual selection from pasted test.
This allows to further manipulate the pasted text after pasting.
For example: `p` `C-j` `C-v` `U` pastes the second most recent element in the kill buffer and capitalizes everything that has been pasted.

### DIFF
--- a/layers/+distributions/spacemacs-bootstrap/packages.el
+++ b/layers/+distributions/spacemacs-bootstrap/packages.el
@@ -233,9 +233,7 @@
  same text above or below, [_C-v_] creates a visual selection \
  from last paste and exits. Anything else exits."
     :bindings
-    ("C-v" (lambda (x) (interactive "*p")
-             (evil-active-region))
-     :exit t)
+    ("C-v" (evil-active-region) :exit t)
     ("C-j" evil-paste-pop)
     ("C-k" evil-paste-pop-next)
     ("p" evil-paste-after)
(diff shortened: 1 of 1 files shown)


layers/+distributions/spacemacs-bootstrap/packages.el
```diff
@@ -229,9 +229,13 @@
   (spacemacs|define-transient-state paste
     :title "Pasting Transient State"
     :doc "\n[%s(length kill-ring-yank-pointer)/%s(length kill-ring)] \
- [_C-j_/_C-k_] cycles through yanked text, [_p_/_P_] pastes the same text \
- above or below. Anything else exits."
+ [_C-j_/_C-k_] cycles through yanked text, [_p_/_P_] pastes the \
+ same text above or below, [_C-v_] creates a visual selection \
+ from last paste and exits. Anything else exits."
     :bindings
+    ("C-v" (lambda (x) (interactive "*p")
```
lebensterben: why is "*p" necessary here?

the lambda form never uses the extra numerical prefix argument (from "p").

lebensterben:
just for clarification, C-v will select the region just pasted before?

layers/+distributions/spacemacs-bootstrap/packages.el
rommeswi: Indeed, I will remove the unused argument, a remnant from a more complicated version before I realized that (evil-active-region) does the job.

rommeswi:
Yes, with the exception when the editor adds additional whitespace during pasting. Then the additional whitespace will also be selected. I found one exception: when pasting lines selected from visual-line-mode, this moves the cursor to the start of the text rather than the beginning of the line, resulting in the whitespace not being selected.

smile13241324 (approved):

smile13241324:

PR #15758, opened 2022-10-03
Configure consult-imenu for Java using LSP backend.

ir210:
This commit configures the narrowing feature of `consult-imenu` for java-mode with entries coming from LSP.

In addition to setting `consult-imenu-config`, `lsp-imenu-index-function` must be set to `#'lsp-imenu-create-categorized-index`.


lebensterben:
I think you can use

```elisp
(use-package consult-imenu
  :after consult
  :init (..))
```

instead of with-eval-after

lebensterben:
merged. thanks.

PR #15770, opened 2022-10-11
Change default scratch mode to lisp-interaction-mode

dalanicolai:
As Spac(Emacs) users are probably primarily (office-)workers of the more adventurous kind, the initial scratch mode should really be `lisp-interaction-mode` (like in vanilla Emacs), where lexical binding is the default.

Also, this helps to prevent (new) users from accidentally setting it to `emacs-lisp-mode` (without lexical binding).

### DIFF
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -238,12 +238,16 @@ This has no effect in terminal or if \"all-the-icons\" is not installed."
   'boolean
   'spacemacs-dotspacemacs-init)
 
-(spacemacs|defc dotspacemacs-scratch-mode 'text-mode
+(spacemacs|defc dotspacemacs-scratch-mode 'lisp-interaction-mode
   "Default major mode of the scratch buffer."
   'symbol
   'spacemacs-dotspacemacs-init)
 
-(spacemacs|defc dotspacemacs-initial-scratch-message 'nil
+(spacemacs|defc dotspacemacs-initial-scratch-message
+  ";; Initial mode for this buffer is configurable via the
+;; `dotspacemacs-scratch-mode' variable.
+;; In the current mode, evaluate lisp code using the major-mode
+;; `, e'-prefix keybindings.\n\n"
   "Initial message in the scratch buffer."
   '(choice (const nil) string)
   'spacemacs-dotspacemacs-init)
--- a/core/templates/.spacemacs.template
+++ b/core/templates/.spacemacs.template
@@ -219,8 +219,8 @@ It should only modify the values of Spacemacs settings."
    ;; (default `text-mode')
    dotspacemacs-new-empty-buffer-major-mode 'text-mode
 
-   ;; Default major mode of the scratch buffer (default `text-mode')
-   dotspacemacs-scratch-mode 'text-mode
+   ;; Default major mode of the scratch buffer (default `lisp-interaction-mode')
+   dotspacemacs-scratch-mode 'lisp-interaction-mode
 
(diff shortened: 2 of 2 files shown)


robbyoconnor:
I am against this because I use the scratch buffer for well..storing text content I need temporarily and it's not always lisp/elisp. 

Just set this in your config.

dalanicolai:
Haha! Well personally, (obviously) I think it should be the other way around (text-mode should be set in the config).
and there is no reason to differ from the Emacs default. Also, I am not sure if you have seen it, but I have added an info text to the scratch buffer, that informs users about the possibility to set the mode via the `dotspacemacs-scratch-mode` variable (info that is currently lacking).

Let me state some arguments for this 'proposal':
- I would love every Emacs user to realize (s)he can enjoy some fun and use Emacs lisp also
- otherwise (new) users might wrongly set the scratch buffer to emacs-lisp-mode (which is mainly what I would like to prevent here)
- it is anyway nice to inform (new) users about how they can set their preferred mode via the dotfile
(which is a tiny operation, also, for users who prefer text mode)

I guess it comes down to if we prefer to sell Emacs the editor, or Emacs the virtual lisp machine.

Anyway, good to hear your opinions. So let democracy decide then...

EDIT
I was just thinking... one can perfectly store text in a lisp buffer also, but I agree that is not ideal, it might get auto (re)formatted

lebensterben:
we can keep the current default (text-mode) but expand the documentation.

lebensterben:
> Also, this helps to prevent (new) users from accidentally setting it to emacs-lisp-mode (without lexical binding).

I don't know about this. Do you have any source about lisp-interaction-mode on lexical binding?

dalanicolai:
It is stated [here](https://www.gnu.org/software/emacs/manual/html_node/elisp/Using-Lexical-Binding.html), just below the introduction of the lexical-binding variable, same for IELM mode...

lebensterben:
emacs-lisp-mode by default doesn't have lexical binding unless lexical-binding is t.

but this won't affect eval-print-last-sexp or most eval-* commands because they always enable lexical binding.

the only difference is for eval-buffer.

smile13241324:
To be honest I think which mode the scratch buffer is set to by default is a pretty opinionated decision. I for example have my scratch buffer set to "org-mode" as I tend to do some planning for my current task in there. For elisp evaluation I think "M-:" for inline evaluation is much more convenient for small scripts. Everything else can be done directly with the standard eval bindings or in a proper repl.

Anyway I think there is no right decision here, however I would rather like to have Spacemacs seen as a ready to use IDE or advanced editor than as a generic lisp evaluation engine. Therefore I find standard texts in the scratch buffer about how to execute lines of code in a certain language rather inconvenient.

When people want to start contributing code they can still pick up all the details about lisp but this is not the only way to contribute nor is elisp code the only code one can write emacs packages in or is preferred in anyway i.e. treemacs is partly based on python and many packages have native cores etc.

So my personal opinion is to keep Spacemacs in IDE mode by default this is just work without pointing users to internals or force them through a basic lisp tutorial hence don't fill in a buffer for personal notes with helpful messages about script evaluation which will most likely only be used when Spacemacs itself is modified. Also keep the default for the scratch buffer with a default text-mode and have users set it to something more language specific when they feel the need for.

dalanicolai:
Okay... looks like it is more opinionated than I expected. Personally, I would like to lure users into helping to fix/investigating bugs, and make nice lisp contributions. But of course your perspectives are equally 'valid'.

So, I guess we can simply close this PR then.

Thanks all for your feedback!

Issue #15674, opened 2022-08-02
repeat loading lsp layer when enable lsp for two or more language layers


sunlin7:
Enable the `lsp` and `javascript` and `c-c++` layers in dot-spacemacs, then start emacs with debug init option, checking the log in `*Message*` buffer, can find the `lsp/layers.el` is loaded more than once ( if enable more lsp-dependened layers, the lsp layer will be loaded again and again ). 
>$ emacs --debug-init
layers/+lang/c-c++/layers.el (source)...
layers/+tools/lsp/layers.el (source)...done
layers/+tools/dap/layers.el (source)...
layers/+tools/lsp/layers.el (source)...done
layers/+tools/dap/layers.el (source)...done
layers/+lang/c-c++/layers.el (source)...done
layers/+tools/dap/packages.el (source)...done
layers/+tools/dap/layers.el (source)...
layers/+tools/lsp/layers.el (source)...done
layers/+tools/dap/layers.el (source)...done
layers/+lang/javascript/packages.el (source)...done
layers/+lang/javascript/layers.el (source)...
layers/+lang/json/layers.el (source)...done
layers/+tools/lsp/layers.el (source)...done
layers/+lang/javascript/layers.el (source)...done
layers/+tools/lsp/packages.el (source)...done
layers/+tools/lsp/layers.el (source)...done
layers/+lang/typescript/packages.el (source)...done
layers/+lang/typescript/layers.el (source)...
layers/+lang/javascript/layers.el (source)...
layers/+lang/json/layers.el (source)...done
layers/+tools/lsp/layers.el (source)...done
layers/+lang/javascript/layers.el (source)...done
layers/+lang/typescript/layers.el (source)...done

lebensterben:
this looks like a misconfiguration on your side.

when starting emacs, lsp or any language layers should not be loaded at all, until they are used.

lebensterben:
how to reproduce this?

When emacs is launched with --debug-init, I don't see any of these loaded.

sunlin7:
Hi @lebensterben just enable the `lsp` layer and `c-c++` layers with lsp backend, for example, enable follow layers will be able to reproduce the issue.
``` emacs-lisp
lsp
(c-c++ :variables c-c++-backend 'lsp-clangd)
(javascript :variables javascript-backend 'lsp)
```

[...]

PR #15772, opened 2022-10-12
core/core-configuration-layer: avoid loading the dependence layers multiple times

sunlin7:
Fix the #15674.
To avoid loading the dependence layers multiple times in (configuration-layer/declare-layer-dependencies).


smile13241324 (approved):
Good find

sunlin7:
Thank you !

PR #15776, opened 2022-10-17
spacemacs-modeline: do not dependent on the neotree

sunlin7:
The `spacemacs-modeline` dependent on the `neotree` layer, but Spacemacs prefer `treemacs` as default file tree browser.
https://github.com/syl20bnr/spacemacs/blob/264a8ffc203a3215623330576941d1722bbee1e4/layers/%2Bdistributions/spacemacs/layers.el#L31

This change apply the function only when the `neotree` layer is used.


smile13241324 (approved):
looks good

sunlin7:
Thank you !

PR #15779, opened 2022-10-19
Add `shell-scripts-shfmt-args` variable

Ralph7C2:
This adds a variable to the `shell-scripts` layer to allow passing arguments through to the `shfmt` package.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -349,6 +349,8 @@ In org-agenda-mode
   - Add ~o~ to link-hint-open-link in woman-mode
 ***** Shaders
 - =shaders= layer has been moved to =gpu= layer.
+***** Shell-scripts
+- Add variable =shell-scripts-shfmt-args= to pass through arguments to =shfmt= package.
 ***** Syntax checking
 - Key bindings (thanks to Loris Lucido):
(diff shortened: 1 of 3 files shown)


sunlin7:
The `shfmt-arguments` is defined as a custom variable, shall we define a generic varaible to control the custom variable?

https://github.com/purcell/emacs-shfmt/blob/12ed73dafe0df62704c048b20ad9a31994625dbc/shfmt.el#L61

Ralph7C2:
I'm not sure I fully grasp what you mean there. Is that how one would normally configure a custom variable in a layer's package, or is it standard just to configure the variable directly and not passthrough like this PR is adding?

sunlin7:
If we didn't change the package's custom variable in Spacemacs repo, it's no necessary to define a variable and pass the value to the package's variable.
If Spacemacs prefer a non-default behavior from a package, it will define a variable and pass it to the package to change its default behavior, writing down the variable document and examples in the layer's README.org also required.

smile13241324 (approved):

smile13241324:
If this is something we expect people to configure often I see no issue in adding the variable to the layers package, however it is not strictly necessary, i.e. many config variables in the LSP layer are directly moved to the emacs-lsp package.

smile13241324:

PR #15798, opened 2022-11-03
[finance] Evilify ledger-reconcile-mode-map

bcc32:
Remove the redundant key bindings for
`ledger-reconcile-{add,quit,change-target}`, which are already bound to reasonable keys in `ledger-reconcile-mode-map`.

Leave `ledger-reconcile-toggle` key binding intact but recommend `SPC` instead of `, ,` to toggle the pending flag.

Leave `ledger-reconcile-finish` key binding intact, because the upstream key binding is `C-c C-c`---not particularly idiomatic for Spacemacs.

### DIFF
--- a/layers/+tools/finance/README.org
+++ b/layers/+tools/finance/README.org
@@ -4,7 +4,7 @@
 
 [[file:img/ledger.png]]
 
-* Table of Contents                     :TOC_5_gh:noexport:
+* Table of Contents                                       :TOC_5_gh:noexport:
 - [[#description][Description]]
   - [[#features][Features:]]
 - [[#install][Install]]
@@ -70,7 +70,7 @@ can for example use ~dix~ or ~vax~ to delete or mark the transaction at point.
 | Key binding | Description                                                               |
 |-------------+---------------------------------------------------------------------------|
 | ~SPC m ,~   | toggle the current transaction pending                                    |
-| ~SPC m a~   | quickly add a transaction, without exiting the reconciliation buffer      |
-| ~SPC m t~   | change the target amount for the selected account                         |
-| ~SPC m q~   | quit reconciliation                                                       |
+| ~a~         | quickly add a transaction, without exiting the reconciliation buffer      |
+| ~t~         | change the target amount for the selected account                         |
+| ~q~         | quit reconciliation                                                       |
 | ~SPC m RET~ | finalize the reconciliation, changing all pending transactions to cleared |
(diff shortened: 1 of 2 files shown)


JoshTRN:
My sense is that this seems like the opposite of Evilify-ing it. I don't personally use `ledger-mode` but from what I can tell, you just modify text like most other emacs. As such, if I wanted to go into insert mode after my cursor and typed `a` I would be very irritated if it didn't display that expected behavior. It also seems like you're overwriting the leader key based on the documentation you've written? I could be missing something, but as brief overview, this seems like very unexpected behavior.

bcc32:
This is the `ledger-reconcile` buffer, which you're not meant to do any text editing in, not just regular ledger ("source") buffers.  Typical usage in non-Spacemacs Emacses would be to press `n` and `p` to navigate between transactions and press `SPC` to toggle transaction status.

I agree it might be confusing to overwrite `SPC`, but FWIW that's the original key binding for toggling transaction status in `ledger-mode`.  Do you think we should not bind that key, and if so, what alternative do you suggest?

JoshTRN:
You're probably right. I'm too dissociated with the use-case to comment intelligently on it. Many standard vim keybindings are overwritten in magit buffers and it doesn't interfere with my usage there not to mention transient state buffers. Then, however, the `SPC` keybinding isn't overwritten.

If I opened `ledger-reconcile` in a split to the right or bottom etc. and wanted to switch windows, how could I do so without closing the `ledger-reconcile` window? Or if I wanted to browse `helm-M-x` would I just have to type `M-x`? That would also be disorienting. 

Again, I don't feel like I really have a place to comment if it's something you use all the time since I don't, so I don't have any alternatives, but the current bindings seem in keeping with the traditional minor mode bindings, so I wouldn't change them.

That's just my 2 cents whatever they may be worth.

bcc32:
> If I opened ledger-reconcile in a split to the right or bottom etc. and wanted to switch windows, how could I do so without closing the ledger-reconcile window? Or if I wanted to browse helm-M-x would I just have to type M-x? That would also be disorienting.

Good point!  It turns out I was wrong, and the SPC key is still bound to the Spacemacs leader and not `ledger-reconcile-toggle`, so I've fixed the documentation.

> Again, I don't feel like I really have a place to comment if it's something you use all the time since I don't, so I don't have any alternatives, but the current bindings seem in keeping with the traditional minor mode bindings, so I wouldn't change them.

Since the reconcile buffer isn't ever used for text editing, it seems unnecessary to prefix all of the useful key bindings with `,`.  For example, `q` seems clearly better than `, q` for quitting, as it is commonly used in many other special modes.  Like the example you mentioned, the reconcile buffer is more like the magit-status buffer than, say, an org-mode buffer.

smile13241324 (approved):

PR #15803, opened 2022-11-07
Use string-edit-at-point instead of string-edit

yyoncho:
- apparently this package was renamed at some point:

https://melpa.org/#/string-edit-at-point

### DIFF
--- a/layers/+spacemacs/spacemacs-editing/packages.el
+++ b/layers/+spacemacs/spacemacs-editing/packages.el
@@ -41,7 +41,7 @@
     (smartparens :toggle dotspacemacs-activate-smartparens-mode)
     (evil-swap-keys :toggle dotspacemacs-swap-number-row)
     (spacemacs-whitespace-cleanup :location (recipe :fetcher local))
-    string-edit
+    string-edit-at-point
     string-inflection
     multi-line
     undo-tree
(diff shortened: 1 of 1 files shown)


sunlin7:
I applied this patch, and manually execute the `string-edit-at-point`, but nothing happened.
Dose the package `string-edit-at-point` work for your ? Thanks
My local emacs is the emacs 29 build with native compile support.

yyoncho:
If you delete your elpa dir spacemacs will fail to install string-edit because it no longer exists(at least this is what I am getting)

sunlin7:
Got your point, then I have no other questions. Thanks !

yannziegler:
I would suggest updating the doc/log as well for consistency (I can't do it): [CHANGELOG.develop (L872)](https://github.com/syl20bnr/spacemacs/blob/9541a8e0a8234d894faffdd20fa5c920b700c868/CHANGELOG.develop#L872) and [layers/LAYERS.org (L891)](https://github.com/syl20bnr/spacemacs/blob/4f5fddc96d67503f1ab3e1c483dc8237a87dce9d/layers/LAYERS.org#L891). Those updates were in my commit, but you beat me to it! Beware not to modify the repository URL for [string-edit.el](https://github.com/magnars/string-edit.el), though, it still uses the old name.

mbld:
I can confirm the issue with a fresh installation of Emcas(28.2) and Spacemacs today. With your changes it works fine again. Anything left to get it merged? So far I did not contribute but might jump in if some work is missing to get it merged :)

smile13241324 (approved):

PR #15812, opened 2022-11-10
Create layer 'apache'

gderber:
This is a simple layer, based HEAVILY on the nginx layer.  In fact, almost an exact duplicate replacing nginx with apache.

I wasn't 100% certain what image to use for apache.png.  I used the one on the front page of their website.


layers/+tools/apache/README.org
```diff
@@ -0,0 +1,22 @@
...
+* Table of Contents                     :TOC_5_gh:noexport:
+- [[#description][Description]]
+  - [[#features][Features:]]
+- [[#install][Install]]
+
+* Description
+This layer adds support for configuring the [[https://httpd.apache.org/][Apache]] web server.
+
+** Features:
+- Syntax highlighting of nginx configuration files via [[https://github.com/emacs-php/apache-mode][apache-mode]].
```
smile13241324: This should be apache configuration files I guess

layers/+tools/apache/packages.el
```diff
@@ -0,0 +1,27 @@
+;;; packages.el --- nginx layer packages file for Spacemacs.
```
smile13241324: nginx -> apache

smile13241324 (changes requested):
I have made a quick review and found some smaller issues you need to fix before this can be merged.

layers/+tools/apache/packages.el
gderber: I've made these fixes in a new commit.  I wasn't sure if you'd want a squashed commit.  If so let me know.

smile13241324 (approved):

PR #15828, opened 2022-11-19
core/core-dotspacemacs.el: speedup dotspacemacs/get-variable-string-list

sunlin7:
Speedup the function `dotspacemacs/get-variable-string-list` for ~3x.
The original function benchmark result is ~ 7.0 seconds; 
assume the enhanced function to be `dotspacemacs/get-variable-string-list2`, the benchmark result is ~1.8 seconds. 
And the last block comprating the results from two functions show that the enhanced function return same results as the original function.
```
(benchmark-run 100
  (dotspacemacs/get-variable-string-list))
;; (7.039543576000001 4 1.1748634609999993)
(benchmark-run 100
  (dotspacemacs/get-variable-string-list2))
;; (1.875364132 0 0.0)

(cl-nset-difference
 (dotspacemacs/get-variable-string-list)
 (dotspacemacs/get-variable-string-list2))
;; nil
```

### DIFF
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -892,13 +892,11 @@ Called with `C-u C-u' skips `dotspacemacs/user-config' _and_ preliminary tests."
 
 (defun dotspacemacs/get-variable-string-list ()
   "Return a list of all the dotspacemacs variables as strings."
-  (all-completions "" obarray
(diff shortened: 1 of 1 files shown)


sunlin7:
@smile13241324  please help to review this change. :)

smile13241324 (approved):

PR #15848, opened 2022-12-20
*layers/+lang/python: enhance the virtual environment detection

sunlin7:
Enenhance the virtual environment detection.
1. when the customer value for `python-shell-interpreter` exists, don't overide it;
2. when the `ipython` not exists, fallback to python layer behavior with pyvenv.
3. the parameter for the `spacemacs/python-setup-everything` should be a root-path for the envrionment
4. other minor enhance...

### DIFF
--- a/layers/+lang/python/funcs.el
+++ b/layers/+lang/python/funcs.el
@@ -150,35 +150,46 @@ as the pyenv version then also return nil. This works around https://github.com/
           executable))
     (executable-find command)))
 
-(defun spacemacs//python-setup-shell (&rest args)
-  (if (spacemacs/pyenv-executable-find "ipython")
-      (progn
-        (setq python-shell-interpreter "ipython")
-        (let ((version (replace-regexp-in-string "\\(\\.dev\\)?[\r\n|\n]$" ""
-                                                 (shell-command-to-string
-                                                  (format "\"%s\" --version"
-                                                          (string-trim (spacemacs/pyenv-executable-find "ipython")))))))
-          (if (or (version< version "5")
-                  (string-blank-p version))
-              (setq python-shell-interpreter-args "-i")
-            (setq python-shell-interpreter-args "--simple-prompt -i"))))
-    (progn
-      (setq python-shell-interpreter-args "-i"
-            python-shell-interpreter "python"))))
+(defun spacemacs//python-setup-shell (&optional root-dir)
+  "Setup the python shell if no customer prefered value or the value be cleaned.
+ROOT-DIR should be the directory path for the environment, `nil' for clean up."
+  (when (or (null python-shell-interpreter)
+            (equal python-shell-interpreter spacemacs--python-shell-interpreter-origin))
+    (if-let* ((root-dir)
+              (default-directory root-dir))
+        (if-let* ((ipython (spacemacs/pyenv-executable-find "ipython"))
+                  (version (replace-regexp-in-string
+                            "\\(\\.dev\\)?[\r\n|\n]$" ""
+                            (shell-command-to-string (format "\"%s\" --version" ipython)))))
+            (setq-local python-shell-interpreter ipython
+                        python-shell-interpreter-args
+                        (concat "-i" (unless (version< version "5") " --simple-prompt")))
+          ;; else try python3 or python
+          (setq-local python-shell-interpreter
+                      (or (spacemacs/pyenv-executable-find "python3")
+                          (spacemacs/pyenv-executable-find "python")
+                          "python3")
+                      python-shell-interpreter-args "-i"))
+      ;; args is nil, clean up the variables
+      (setq-local python-shell-interpreter nil
+                  python-shell-interpreter-args nil))))
 
-
-(defun spacemacs//python-setup-checkers (&rest args)
+(defun spacemacs//python-setup-checkers (&optional root-dir)
+  "Setup the checkers.
+ROOT-DIR should be the path for the environemnt, `nil' for clean up"
   (when (fboundp 'flycheck-set-checker-executable)
-    (let ((pylint (spacemacs/pyenv-executable-find "pylint"))
-          (flake8 (spacemacs/pyenv-executable-find "flake8")))
-      (when pylint
-        (flycheck-set-checker-executable "python-pylint" pylint))
-      (when flake8
-        (flycheck-set-checker-executable "python-flake8" flake8)))))
+    (if-let* ((root-dir)
+              (default-directory root-dir))
+        (dolist (x '("pylint" "flake8"))
+          (when-let ((exe (spacemacs/pyenv-executable-find x)))
+            (flycheck-set-checker-executable (concat "python-" x) exe)))
+      ;; else root-dir is nil
+      (dolist (x '("pylint" "flake8"))
+        (set (flycheck-checker-executable-variable (concat "python-" x)) nil)))))
 
-(defun spacemacs/python-setup-everything (&rest args)
-  (apply 'spacemacs//python-setup-shell args)
-  (apply 'spacemacs//python-setup-checkers args))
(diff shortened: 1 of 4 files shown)


emacs18:
It appears to me that this change is not sufficient, because there are multiple ways to change virtual environments but only one method is supported in this change.  If `pyenv-mode-set` or `pyenv-mode-unset` is called to change virtual environments, appropriate code seems to be called to change states properly.  However what about `pyvenv-activate` and `pyvenv-deactivate` calls?  Until this review I was aware of only the latter pair.  I had no idea that there were alternate ways to change environments.

emacs18:
Oops.  Sorry.  I just realized that proper hooks are in place even for `pyvenv` calls.

layers/+lang/python/funcs.el
```diff
@@ -151,21 +151,22 @@ as the pyenv version then also return nil. This works around https://github.com/
...
-                  (string-blank-p version))
-              (setq python-shell-interpreter-args "-i")
-            (setq python-shell-interpreter-args "--simple-prompt -i"))))
-    (progn
-      (setq python-shell-interpreter-args "-i"
-            python-shell-interpreter "python"))))
-
+  "Setup the python shell if no customer prefered value or the value be cleaned.
+ARGS should be the directory path for the environment, `nil' for clean up."
+  (if args
```
emacs18: When would `args` be not `nil`?

sunlin7: `pyenv-mode-unset`  and `pyvenv-deactivate` will call the function `spacemacs/python-setup-everything` with `nil` param.

smile13241324 (approved):
Looks good, thanks for the great work @sunlin7!

sunlin7:
Hi @smile13241324 
Thank you more!

dlchet:
Hello, I'm encountering an issue that prevents LSP from properly loading in python since this merge. It's definitely possible I've got a weird or broken setup, but it wasn't obvious to me how to fix it after a while of poking around. It resolves immediately as soon as I reset to `b74da79dbbd8573ab4f43721c74014d6f54d8bd0`. I'm happy to create an issue and/or provide any debugging - I'm using pyenv and a .python-version file. This is the error I'm seeing:

> File mode specification error: (void-variable args)

And here's the backtrace:

```
Debugger entered--Lisp error: (void-variable args)
  (if args (progn (pyenv-mode-full-path version)))
  (spacemacs/python-setup-everything (if args (progn (pyenv-mode-full-path version))))
  (lambda (&optional version) (spacemacs/python-setup-everything (if args (progn (pyenv-mode-full-path version)))))("venv")
  apply((lambda (&optional version) (spacemacs/python-setup-everything (if args (progn (pyenv-mode-full-path version))))) "venv")
  pyenv-mode-set("venv")
  (cond ((member version (pyenv-mode-versions)) (setenv "VIRTUAL_ENV" version) (pyenv-mode-set version)) (t (message "pyenv: version `%s' is not installed (set by %s)" version file-path)))
  (if version (cond ((member version (pyenv-mode-versions)) (setenv "VIRTUAL_ENV" version) (pyenv-mode-set version)) (t (message "pyenv: version `%s' is not installed (set by %s)" version file-path))))
  (let* ((root-path (and t (locate-dominating-file default-directory ".python-version"))) (file-path (and root-path (expand-file-name ".python-version" root-path))) (version (and file-path (let ((temp-buffer (generate-new-buffer " *temp*" t))) (save-current-buffer (set-buffer temp-buffer) (unwind-protect (progn ... ...) (and ... ...))))))) (if version (cond ((member version (pyenv-mode-versions)) (setenv "VIRTUAL_ENV" version) (pyenv-mode-set version)) (t (message "pyenv: version `%s' is not installed (set by %s)" version file-path)))))
  spacemacs//pyenv-mode-set-local-version()
  run-hooks(change-major-mode-after-body-hook prog-mode-hook python-mode-hook)
  apply(run-hooks (change-major-mode-after-body-hook prog-mode-hook python-mode-hook))
  run-mode-hooks(python-mode-hook)
  python-mode()
  set-auto-mode-0(python-mode nil)
```

relevant part of .spacemacs:
```
     (python :variables
             python-format-on-save t
             python-formatter 'black
             python-lsp-server 'pyright
             python-sort-imports-on-save t
             python-test-runner 'pytest)
```

@sunlin7 any thoughts? is there some emacs pyenv config I need (or should remove?)

emacs18:
Lin made follow-on change two days go.  Do you have this fix?

dlchet:
> Lin made follow-on change two days go. Do you have this fix?

Yes, I've tried the tip of develop (`fd749704ff567c4766e2b8e192d6fa46708a7ee3`) which includes a small fix; no change in what happens in my environment. I also tried a fresh checkout of spacemacs - initially I was sure getting rid of elcs or something would fix it.

sunlin7:
Hi @dlchet ,
Could you provide the backtrace for the issue? I'll check it as soon as posible.

emacs18:
I think the culprit is the following code where `args` undefined symbol is used 4th line from bottom.

```python
(defun python/init-pyenv-mode ()
  (use-package pyenv-mode
    :if (executable-find "pyenv")
    :commands (pyenv-mode-versions)
    :init
    (progn
      (pcase python-auto-set-local-pyenv-version
        ('on-visit
         (dolist (m spacemacs--python-pyenv-modes)
           (add-hook (intern (format "%s-hook" m))
                     'spacemacs//pyenv-mode-set-local-version)))
        ('on-project-switch
         (add-hook 'projectile-after-switch-project-hook
                   'spacemacs//pyenv-mode-set-local-version)))
      ;; setup shell correctly on environment switch
      (dolist (func '(pyenv-mode-set pyenv-mode-unset))
        (advice-add func :after
                    #'(lambda (&optional version)
                        (spacemacs/python-setup-everything
                         (when args (pyenv-mode-full-path version))))))
      (spacemacs/set-leader-keys-for-major-mode 'python-mode
        "vu" 'pyenv-mode-unset
        "vs" 'pyenv-mode-set))))
```

emacs18:
Also #' prefix I don't think is needed or usually used for (lambda ...).

sunlin7:
Hi @emacs18 ,
Agree with you.

Hi @dlchet ,
Please change the `(when args (pyenv-mode-full-path version)` to `(when version (pyenv-mode-full-path version)` and try again. Thanks.

sunlin7:
This PR https://github.com/syl20bnr/spacemacs/pull/15856 should be able to fix the issue.
@dlchet please help verify it. Thanks.

dlchet:
> This PR #15856 should be able to fix the issue. @dlchet please help verify it. Thanks.

Making this change solved my problem, thanks very much to both of you!

sunlin7:
@dlchet ,  @emacs18 
Thanks you all !

layers/+lang/python/packages.el
```diff
@@ -319,7 +319,10 @@
                    'spacemacs//pyenv-mode-set-local-version)))
       ;; setup shell correctly on environment switch
       (dolist (func '(pyenv-mode-set pyenv-mode-unset))
-        (advice-add func :after 'spacemacs/python-setup-everything))
+        (advice-add func :after
+                    #'(lambda (&optional version)
+                        (spacemacs/python-setup-everything
+                         (when args (pyenv-mode-full-path version))))))
```
nl-ix-arjan, line 325: Yo, the `args` here is unbound. It should probably be `(when version ...` I'm guessing?

smile13241324:
The hotfix has been merged to develop please verify and close this issue if it works.

layers/+lang/python/packages.el
sunlin7: Yes, that's the root cause. Resolved in https://github.com/syl20bnr/spacemacs/pull/15856. Thanks.

Issue #15105, opened 2021-10-18
Cannot exit emacs


Soromeister:
Trying to exit emacs after clicking inside on the "Update packages". Commands just pile up and there is no way to simply "cancel whatever keeps piling down there. Please see screenshot for details. This would not have been an issue in VIM, where ENTER actually executes the command you're giving it.

![image](https://user-images.githubusercontent.com/45243695/137707071-6fe15377-1dca-42f7-90e7-fcf180cf8da2.png)

smile13241324:
Ok there seems to be a bug in the home buffer when emacs is run in terminal mode and a left click is done on the update button.

This results in jumping into the mini buffer where all entered keys are listed.

This only happens when the button is clicked in terminal mode.

smile13241324:
@duianto are you familiar with those push buttons? I have no idea whats going on here but all non url link buttons on the home buffer are broken when spacemacs is running in terminal mode and a mouse is used.

duianto:
This might be related to using the mouse in a Linux terminal.
I don't know if it also happens in a Mac terminal.

### Summary:
The package update doesn't start when mouse clicking on `[Update Packages]` in the Manjaro Terminal.
It just inserts minibuffer text like this, with every mouse click:
>ESC [ < 0 ; 6 0 ; 1 3 m-

At least for me, in Manjaro, `C-g` cancels the minibuffer mouse click messages.
Then Spacemacs works as expected, and `SPC q q` exits.

I'm not familiar with Linux terminals, so someone with more knowledge would probably have to investigate further.

### Manjaro Terminal
`emacs -nw`

When mouse left-clicking on the `P` in the button `[Update Packages]`.
Then the minibuffer shows:
```
ESC [ < 0 ; 6 0 ; 1 3 m-
```
Nothing else seems to happen.
The package update never starts.

Left-clicking again shows:
```
ESC [ < 0 ; 6 0 ; 1 3 m ESC [ < 0 ; 6 0 ; 1 3 M ESC [ < 0 ; 6 0 ; 1 3 m-
```

Pressing `C-g` shows: `Quit`.

`view-lossage` (`C-h l`) shows:
```
ESC [ < 0 ; 6 0 ; 1 3 M ;; widget-button-click
ESC [ < 0 ; 6 0 ; 1 3 m ESC [ < 0 ; 6 0 ; 1 3 M ESC [ < 0 ; 6 0 ; 1 3 m C-g C-h l ;; view-lossage
```
[...]

Issue #13539, opened 2020-05-01
On first screen, clicking on [links] does not open anything, inserts line break.


fidergo-stephane-gourichon:
#### Description :octocat:

On first screen, clicking on links (bracketed words "Homepage" "Documentation" etc) does not open anything, inserts line break.

#### Reproduction guide :beetle:
- Make sure you have no `.emacs.d` or `.spacemacs*`
- `git clone https://github.com/syl20bnr/spacemacs` optionally switch to `develop` branch
- Start Emacs
- Click on any of the bracketed words "Homepage" "Documentation".

*Observed behaviour:* :eyes: :broken_heart:

The word is broken as if a newline character is inserted at the position of the clicked character

See image.
![spacemacs_click_breaks](https://user-images.githubusercontent.com/2843528/80806711-b2046880-8bbb-11ea-8e7f-434c8339dd33.png)

*Expected behaviour:* :heart: :smile:

- Clicking on "Homepage" would show some spacemacs-related page.
- Clicking on "Documentation" would show some spacemacs-related documentation.
-Etc...

(%LAST_KEYS%)
%SYSTEM_INFO%

#### Backtrace :paw_prints:

```
%BACKTRACE%
```

#### Additional information

* Xubuntu 20.04 with plain X11 emacs packages, no customization.
* Right after answering the questions "What is your preferred editing style?" and "What distribution of spacemacs would you like to start with?", clicking on the "links" produces no visible effect.
* Once spacemacs has finished downloading 110 packages (wow), the same behavior can be observed.

nonducor:
You have to finish configuring it  (note the question on the bottom of the window).

jr0cket:
It would seem that at the start of the install wizard the the home buffer allows editing.

Is it simple to hide these links until the installer has completed, as Spacemacs is not installed fully until the installer has completed and created a `.spacemacs` file (or a `.spacemacs` or `.spacemacs.d/init.el` file is manually added).

lebensterben:
hiding home buffer links is a simple solution.

But this issue is of very, very, very low severity and only affect a very specific use case. that's why I think it's not fixed yet.

lebensterben:
> From your comments I finally understood that the links are to be ignored. I thought they were the main place to start and a necessary path.

No I did not say that.
[...]

PR #15859, opened 2022-12-31
Fix widget-button-click for core Spacemacs buffer

### LINKED ISSUE #14964: Spacemacs buffer terminal mouse click on links appears broken
It seem that this line: https://github.com/syl20bnr/spacemacs/blob/develop/core/core-spacemacs-buffer.el#L91

Should instead bound to `mouse-1` and not to `down-mouse-1` as it does.

The way it is now, when I click on a link from a terminal Emacs, the link gets highlighted, but it doesn't navigate, and Emacs is stuck waiting for more input as if I have not finished typing a keychord:

> down-mouse-1 ESC [ < 0 ; 8 1 ; 1 9 ; m

I'm not too sure why this happens, but it prevents the click from working, and afterwards everything I type or click on just gets added to more commands on the keychord and it only stops if I `C-g`.

I tried to simply change it to:

```
(define-key map [mouse-1] 'widget-button-click)
```

And that seems to have fixed it.

opie4624:
Fixes syl20bnr/spacemacs#14964

Possibly related to syl20bnr/spacemacs#13539

### DIFF
--- a/core/core-spacemacs-buffer.el
+++ b/core/core-spacemacs-buffer.el
@@ -122,7 +122,7 @@ It's cleared when the idle timer runs.")
       (define-key map (kbd "8") 'spacemacs-buffer/jump-to-number-startup-list-line)
       (define-key map (kbd "9") 'spacemacs-buffer/jump-to-number-startup-list-line))
 
-    (define-key map [down-mouse-1] 'widget-button-click)
+    (define-key map [mouse-1] 'widget-button-click)
     (define-key map (kbd "RET") 'spacemacs-buffer/return)
 
     (define-key map [tab] 'widget-forward)


smile13241324 (approved):

real-or-random:
This has the strange effect that I'm always in "visual select" mode... 	

See screencast:
[Screencast from 2023-01-03 11-36-25.webm](https://user-images.githubusercontent.com/1071625/210341371-7869a6de-4c16-417b-a810-4fbc31e79eff.webm)

I'm never *holding down* the mouse button during this screencast, just pressing it.

I haven't included the mode indicator but it shows grey (=visual in my theme) the entire time.

lebensterben:
@real-or-random 

Because if you use `evil`, `down-mouse-1` is now bound to `evil-mouse-drag-region` which enters `evil-visual-state`.

So in other word this PR introduced the new issue.

real-or-random:
Is there a chance that this fixed https://github.com/syl20bnr/spacemacs/issues/15105 on the way?

lebensterben:
@real-or-random 

I never reproduced the issue described there so I don't know.

PR #15867, opened 2023-01-04
fix(spacemacs-buffer): mouse click

### LINKED ISSUE #14964: Spacemacs buffer terminal mouse click on links appears broken

lebensterben:
fixes #14964

### DIFF
--- a/core/core-spacemacs-buffer.el
+++ b/core/core-spacemacs-buffer.el
@@ -122,7 +122,7 @@ It's cleared when the idle timer runs.")
       (define-key map (kbd "8") 'spacemacs-buffer/jump-to-number-startup-list-line)
       (define-key map (kbd "9") 'spacemacs-buffer/jump-to-number-startup-list-line))
 
-    (define-key map [mouse-1] 'widget-button-click)
+    (define-key map [down-mouse-1] 'spacemacs-buffer//mouse-1)
     (define-key map (kbd "RET") 'spacemacs-buffer/return)
 
     (define-key map [tab] 'widget-forward)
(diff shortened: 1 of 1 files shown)


lebensterben:
@opie4624 @real-or-random 
Please review

real-or-random:
Unfortunately this fixes it only partly.  :/

I get into visual select again by
 * double-clicking mouse left and then clicking mouse left, or 
 * clicking mouse right

lebensterben:
@real-or-random 

This PR only intends to fix the left mouse click that it should open entries on spacemacs-buffer.

what you mentioned can be reported in a separate issue.

real-or-random:
> This PR only intends to fix the left mouse click that it should open entries on spacemacs-buffer.

Okay, I see. 

The thing is that even without this PR here, opening entries with left mouse in fact works, also when I'm in visual mode after clicking somewhere.

opie4624:
Fixes it for me!

TheJJ:
thanks! it seems to be fixed for good now :)

PR #15868, opened 2023-01-04
fix(haskell): missing haskell-debug-mode-map

### LINKED ISSUE #15866: Symbol’s value as variable is void: haskell-debug-mode-map
After a recent change in `lsp-hakell`, Spacemacs throws the warning above when opening something LSP-related file. This is supposed to be a Spacemacs problem, not a `lsp-hakell` one, see https://github.com/emacs-lsp/lsp-haskell/issues/165. No idea what's going on here...

lebensterben:
fixes #15866

### DIFF
--- a/layers/+lang/haskell/packages.el
+++ b/layers/+lang/haskell/packages.el
@@ -229,20 +229,6 @@
             "hh"  'hoogle
             "hG"  'haskell-hoogle-lookup-from-local)))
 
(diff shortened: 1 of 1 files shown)


lebensterben:
@svenpanne please review

svenpanne:
This change makes the reported problem go away, thanks for the quick fix! :+1:

PR #15871, opened 2023-01-04
fix(spacemacs-buffer): fix retrieval org-agenda-files

lebensterben:
In particular, we only need to call `(org-agenda-files)` (and hence load `org`)
if the variable `org-agenda-files` is a single file name. Otherwise we can
safely use the value of the said variable directly.

This was not handled correctly before this commit.

### DIFF
--- a/core/core-spacemacs-buffer.el
+++ b/core/core-spacemacs-buffer.el
@@ -1273,27 +1273,38 @@ SEQ, START and END are the same arguments as for `cl-subseq'"
     (insert spacemacs-buffer-list-separator)))
 
 (defun spacemacs-buffer//insert-recent-files (list-size)
+  "Insert recent file entries to spacemacs-buffer.
+
+LIST-SIZE is specified in `dotspacemacs-startup-lists' for recent entries."
   (unless recentf-mode (recentf-mode))
-  (let ((agenda-files
-         (let ((default-directory
-                (or (bound-and-true-p org-directory) default-directory))
-               (files
-                (if (or (not (fboundp 'org-agenda-files))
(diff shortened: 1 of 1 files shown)


lebensterben:
@sunlin7 please review.

sunlin7:
Hi @lebensterben 
1. if the `org-agenda-files` is not a list, it will always pop a prompt for loading the org-mode at starup, it's a little annoying;
2. when the `org-agenda-files` is list of relative paths, the `expand-file-name` should base on the `org-directory`.
Thanks

lebensterben:
@sunlin7 

As for (1),

Notice that until this commit, when `org-agenda-files` is a file name instead of a string, it won't be correctly processed. Because the following always avoids calling `#'org-agenda-files` unless `org` is loaded.

```elisp
(if (or (not (fboundp 'org-agenda-files))
             (autoloadp (symbol-function 'org-agenda-files)))
    (bound-and-true-p org-agenda-files)
  (org-agenda-files))
```

It's imperative to handle this case correctly. I added the `y-or-n` question to let user decided whether to load `org`. Of course I can avoid asking this question but then loading `org` automatically slows down the start-up.

And in addition, the question itself hints the user to use a list instead of a string for `org-agenda-files`'s value.

---

As for (2),

I'm going to fix it in another commit.
I will set `default-directory` to `org-directory` if it's set or use `~/org`.

Also note that before this commit, since `default-directory` is modified in the outer `let` form, it affects all subsequent `expand-file-name` forms, which is erroneous. (we only need that when expanding file name for `org-agenda-files`)

sunlin7:
Hi @lebensterben 
I have no other questions for the latest commit. Thanks.

PR #15876, opened 2023-01-06
fix: helm-display-header-line must be set when helm-echo-input-in-header-line is t

kassick:
When the current path is longer then the current frame width, helm-ff may open with the cursor in a wrong position. Any interaction will result in a "Text is readonly" message, until helm ends up navigating to the root folder and we can interact again.

According to @thierryvolpiatto, helm-echo-input-in-header-line and helm-display-header-line must be both set.
https://github.com/emacs-helm/helm/issues/2579#issuecomment-1373158505

Updating spacemac's defaults for these variables fixes the issue.


kassick:
Updated docstring makes this clear in helm documentation:
https://github.com/emacs-helm/helm/commit/4e39df9ee024f82548028f0edde9c71afbcd3635

lebensterben:
Thanks.

PR #15888, opened 2023-01-12
Fix paradox-list-package key not being bound on init

mosquito-magnet:
The key binding of autoloaded `paradox-list-packages` to leader a k does not execute until package is autoloaded by some other means, since being moved from :init to :config in f4b04bcf228b9e9a6eb1e19a9f1f78b429e347ad. Moved it back to init


lebensterben:
thx

PR #15891, opened 2023-01-17
org-layer: Update to org 9.6.1, fixing broken install

### LINKED ISSUE #15360: SPC SPC doesn't work
#### Description
`SPC SPC` doesn't work.
(`M-x` work)

The log output to the Message buffer is as follows.
```log
SPC SPC-  which-key: There are no keys to show
SPC SPC SPC is undefined
```

`SPC SPC` stopped working at commit b7e1dee13f619646d9d419f518f6ba8d835bef41

#### Reproduction guide
– Start Emacs
– Press `SPC SPC`

*Observed behaviour:*
Pressing the `SPC SPC` will result in `SPC SPC-`, prompting me to continue entering commands.

*Expected behaviour:*
helm-M-x(or perhaps spacemacs/helm-M-x-fuzzy-matching)

#### System Info :computer:
– OS: darwin
– Emacs: 28.0.91
– Spacemacs: 0.999.0
– Spacemacs branch: develop (rev. ee3c558c8)
– Graphic display: nil
– Distribution: spacemacs
– Editing style: vim
– Completion: helm
– Layers:
```elisp
(emacs-lisp helm multiple-cursors treemacs)
```
– System configuration features: ACL GIF GLIB GMP GNUTLS JPEG JSON LCMS2 LIBXML2 MODULES NATIVE_COMP $

Le09:
This integrates the patch that fix default install: https://list.orgmode.org/87bkonzisl.fsf@gnu.org/T/#u

In particular fixes the error:
(invalid-function org-assert-version)

I don't know how emacs resolve the version, as I would expect it to choose the highest patch release satisfying the min requirement; but the 9.6 release is borked, and any fresh install would choose the 9.6 and thus break because of this issue (I tried more than 5 times without success).

I'm trying to begin into spacemacs, let me know if I messed up anything up!

### DIFF
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -33,7 +33,7 @@
     htmlize
     ;; ob, org, org-agenda and org-contacts are installed by `org-contrib'
     (ob :location built-in)
-    (org :location elpa :min-version "9.6")
+    (org :location elpa :min-version "9.6.1")
     (org-agenda :location built-in)
     (org-wild-notifier
                 :toggle org-enable-notifications)


lebensterben:
cc @dalanicolai

dalanicolai:
Thanks for the PR, would be great if this works. Indeed, I also expected Spacemacs to choose the highest patch release,
so I merely increased the min-version number from 9.5 to 9.6 (to force the update).

I am not sure how that patch works, i.e. what is the difference between "org-loaddefs.el" and "org-autoloads.el". So to me, it is not immediately clear how this patch solves the loading order issue.
But I guess the patch was created to solve exactly that loading order issue, so then it would probably work.

@smile13241324 So let's see if this indeed solves the issue, then I guess the fix proposed [here](https://github.com/syl20bnr/spacemacs/issues/15360#issuecomment-1383187770) is not necessary (I can not easily test it, because I don't know how to easily reproduce the error in the first place).

Also in that case @Le09, thanks, great find!

Le09:
Thank you, I had missed that thread. My previous fix was to remove org packages and revert to 025e9f7e325ef2c71de1aee9ba0ef3d52c8dc5b2, just before  changes to org. I can only say both seem to 'work on my machine™' .

smile13241324:
Thanks I will have a look later

smile13241324 (approved):

smile13241324:
This is merged and the side effects of the other hotfix should be solved too, do you still see version mismatches?

Issue #15907, opened 2023-01-29
unnecessary y-or-no pop-up dialog box regarding org-agenda-files being a string


emacs18:
Due to a change made in spacemacs few weeks ago (detailed below), I get the same pop-up dialog box twice each time I start emacs.  The diaglog box asks whether to continue because `org-agenda-files` is  a string! The value of this variable for me has not changed in many  years.  For now I set `dotspacemacs-startup-lists` to `nil` to prevent these dialog boxes from popping on not only at startup, but also when I switch into or away from the spacemacs buffer.

The code in question is the following lines added in `spacemacs-buffer//insert-recent-files` function a weeks ago:

```elisp
                       (if (listp org-agenda-files)
                           ;; if it's a list, we take that value directly
                           org-agenda-files
                         ;; but if it's a string, it must be file where the list
                         ;; of agenda files are stored in that file and we have
                         ;;to load `org-agenda' to process the list.
                         (when (y-or-n-p "`org-agenda-files' is a string and \
not a list. Load `org' and continue?")
                           (require 'org)
                           (org-agenda-files)))
```

lebensterben:
I made this change and it's intended.

as explained, if org-agenda-files is a string we **have to** load org during startup, which can be slow for some people.

lebensterben:
you can

1. remove agenda from the startup list
2. change the value of the said variable

emacs18:
If that is the case then why just print a warning to the user that org is being loaded due to org-agenda-files being a string?
In my case org is *always* loaded and very early on so there will be no slow down at all.
Forcing me to use mouse to click on diaglog boxes is not right.

In my opinion the change is question is not necessary which is trying to be too smart and causing problems
My hope is that it will be reverted.

emacs18:
Also you could check if org was already loaded first rather than using y-or-no call.

emacs18:
Following is what I mean.

```diff
diff --git a/core/core-spacemacs-buffer.el b/core/core-spacemacs-buffer.el
index c88192ad9..e92d68b7d 100644
--- a/core/core-spacemacs-buffer.el
+++ b/core/core-spacemacs-buffer.el
@@ -1295,8 +1295,9 @@ LIST-SIZE is specified in `dotspacemacs-startup-lists' for recent entries."
                          ;; but if it's a string, it must be file where the list
                          ;; of agenda files are stored in that file and we have
                          ;;to load `org-agenda' to process the list.
[...]

PR #15908, opened 2023-01-30
Do not call org-agenda-files if org-mode is already loaded

emacs18:
Fix for https://github.com/syl20bnr/spacemacs/issues/15907 to prevent calling y-or-n-p if org mode is already loaded.


lebensterben:
thx

PR #15911, opened 2023-02-09
spacemacs-editing: disable undo-tree history file

bb2020:
`undo-tree` history file makes copies of files so I suppose it should be disabled by default.


lebensterben:
the default behavior is to save backups.

and we save the backups in .emacs.d/.cache so it's not populating the working directory.

so I don't understand why would you want to turn it off.

bb2020:
@smile13241324

lebensterben:
other features of this package relies on the backup files.

if you don't like it, you can turn it off in your user-config or disable undo-tree.

PR #15912, opened 2023-02-09
[search-engine] add google-this package

bb2020:


layers/+web-services/search-engine/packages.el
```diff
@@ -24,8 +24,16 @@
...
-  '(engine-mode))
+  '(engine-mode google-this))
 
+(defun search-engine/init-google-this ()
+  (use-package google-this
+    :defer t
+    :init
+    (progn
+      (spacemacs/set-leader-keys "aws" 'google-this-ray)
+      (with-eval-after-load 'google-this
```
lebensterben, line 35: if you use this it means you should use :config

layers/+web-services/search-engine/packages.el
```diff
@@ -24,8 +24,16 @@
...
+        (fset 'google-this-url (lambda () "https://www.google.com/search?q=%s&pws=0&gl=us&gws_rd=cr"))))))
```
lebensterben, line 36: this is bad for two reasons:

1. you assumed language and region
2. you can define a function normally instead of using fset

lebensterben (changes requested):

lebensterben:
You can add them in a private layer or add them in user-config.

bb2020:
Why are you instantly closing pull requests? I would have fixed it with your suggestions.

lebensterben:
see the comment https://github.com/syl20bnr/spacemacs/pull/15912#issuecomment-1424715089

bb2020:
Of course I know there is user config. This is not a proper way to maintain an open source project. If you don't want contributions, you should archive this repository.

lebensterben:
also, this is far superior https://github.com/hrs/engine-mode/tree/a5f1ae2b087109f37a704c496942cec9d9f48698

lebensterben:
> If you don't want contributions

that's not true.

I closed this issue because 

1. engine-mode is far better than your new addition.
2. the code quality is awful.
3. we had internal discussion before to discourage PR that introduces trivial functionality like this.

bb2020:
Can I add the URL into `search-engine-alist` with this PR?
Because regional Google sites are pretty bad for tech related searches.

lebensterben:
There is already a Google entry. Your addition in the PR and you later suggestion is simply meaningless.

bb2020:
The Google entry yields bad results when you are abroad, because it uses regional websites. I wanted to add a Google US entry that would be helpful for many people.

lebensterben:
> The Google entry yields bad results when you are abroad, because it uses regional websites. 

It uses regional website so what?
The search region doesn't depend on which Goolge you're accessing.

Then if you've logged in your Google account, it will use your normal region and language settings.

Meanwhile your PR hardcoded geo location which is indeed erroneous for most people on this world.

bb2020:
If you can reopen the PR, you will see what I meant.
https://github.com/bb2020/spacemacs/commit/43cc0055999a890aecde6a7d9a868f7f0f32c3a8

lebensterben:
are we going to add entries for each region?

bb2020:
It literally says regionless.

Issue #15901, opened 2023-01-23
Recent Files do not work on fresh install


rromoli:
#### Description :octocat:
Recent` Files do not work on fresh install

#### Reproduction guide :beetle:
- Start Emacs
- Open several files, close Spacemacs and open it again and no recent files list is present. Do it several time but nothing changes. I see that .emacs.d/cache/recentf is present and polulated with the files I opened but it is not visualized in the home buffer.

*Observed behaviour:* :eyes: :broken_heart:
No recent files list present

*Expected behaviour:* :heart: :smile:
List of recently viewed files to be populated

#### System Info :computer:
- OS: gnu/linux
- Emacs: 28.2
- Spacemacs: 0.999.0
- Spacemacs branch: develop (rev. 58a52ae6e)
- Graphic display: t
- Running in daemon: nil
- Distribution: spacemacs-base
- Editing style: emacs
- Completion: helm
- Layers:
```elisp
(emacs-lisp helm multiple-cursors)
```
- System configuration features: ACL CAIRO DBUS FREETYPE GIF GLIB GMP GNUTLS GSETTINGS HARFBUZZ JPEG JSON LIBOTF LIBSELINUX LIBSYSTEMD LIBXML2 M17N_FLT MODULES NATIVE_COMP NOTIFY INOTIFY PDUMPER PNG RSVG SECCOMP SOUND THREADS TIFF TOOLKIT_SCROLL_BARS X11 XDBE XIM XPM XWIDGETS GTK3 ZLIB

#### Backtrace :paw_prints:
```
<<BACKTRACE IF RELEVANT>>
```

lebensterben:
Go to home buffer and SPC b d.
Then SPC b h.
Does it appear?

rromoli:
No, it doesn't appear.

emacs18:
Using vanilla spacemacs files (i.e., without making any changes whatsoever to any files), I too see this behavior.
`*spacemacs*` buffer does not show recent files.  Typing `SPC b d` followed by `SPC b h` does not show recent files.
However if now do `M-x load-library` followed with `org` so that org mode is loaded, then subsequent `SPC b d` followed by `SPC b h` shows recent files.

Thus displaying of recent files seems to require org mode to be loaded.

emacs18:
I was using few days old emacs 29 on ubuntu.

emacs18:
[...]

PR #15923, opened 2023-02-18
do not filter away all recent files when org-directory is not set

emacs18:
fix for https://github.com/syl20bnr/spacemacs/issues/15901


lebensterben:
thx

PR #15925, opened 2023-02-20
latex: Make it possible to use RefTeX completions even with LSP backend

real-or-random:
This PR adds a key binding `C-b` for `company-other-backend`, which can be pressed in the company dropdown to switch to the next completion backend.

The user can use this in the latex layer to get to RefTeX completions on demand, even with the LSP backend. This is useful because RefTeX tends to offer better completions than the LSP server texlab, at least for some things to be completed.

### DIFF
--- a/layers/+lang/latex/funcs.el
+++ b/layers/+lang/latex/funcs.el
@@ -23,31 +23,31 @@
 
 (defun spacemacs//latex-setup-company ()
   "Conditionally setup company based on backend."
-  (pcase latex-backend
-    ('lsp
-     (spacemacs|add-company-backends ;; Activate lsp company explicitly to activate
-       :backends company-capf        ;; standard backends as well
-       :modes LaTeX-mode))
-    ('company-auctex
-     (when (configuration-layer/package-used-p 'company-auctex)
-       (if (configuration-layer/package-used-p 'company-math)
-           (spacemacs|add-company-backends
-             :backends (company-math-symbols-unicode
-                        company-math-symbols-latex
-                        company-auctex-macros
-                        company-auctex-symbols
-                        company-auctex-environments)
-             :modes LaTeX-mode)
-         (spacemacs|add-company-backends
-           :backends (company-auctex-macros
-                      company-auctex-symbols
-                      company-auctex-environments)
-           :modes LaTeX-mode)))
-     (when (configuration-layer/package-used-p 'company-reftex)
-       (spacemacs|add-company-backends
-         :backends company-reftex-labels
-         company-reftex-citations
-         :modes LaTeX-mode)))))
+  ;; Always activate auctex and reftex backends so that they're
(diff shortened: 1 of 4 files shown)


layers/+lang/latex/funcs.el
```diff
@@ -23,31 +23,31 @@
...
-                      company-auctex-environments)
-           :modes LaTeX-mode)))
-     (when (configuration-layer/package-used-p 'company-reftex)
-       (spacemacs|add-company-backends
-         :backends company-reftex-labels
-         company-reftex-citations
-         :modes LaTeX-mode)))))
+  ;; Always activate auctex and reftex backends so that they're
+  ;; accessible via company-other-backend even when using lsp.
+  (when (configuration-layer/package-used-p 'company-auctex)
```
lebensterben, line 28: could we use a unified `backends` variable to store all backends. and adding individual backends to the list when conditions are met.

then there's no need to switch backend.

real-or-random: Do you mean a grouped backend (https://company-mode.github.io/manual/Backends.html#Grouped-Backends)?

Yes, I tried this locally before I ended up with this solution. I don't think a grouped backend is a good idea.

The issue is that both lsp and reftex suggest the same items (assuming they work properly), but reftex suggestions are better searchable. For example, lsp may suggest `ABC18` as a citation key with no additional info, and reftex may suggest the same citation with a much nicer text `Author One, Ben Author, Charlie Anotherautor: On The Usefulness of LSP. 2018. Bla....`. that, when you select it, still gives you `ABC18` but you have a much better fuzzy search.

So if we have a grouped backend, then we'd have every citation twice in the dropdown, which ends up being more confusing than it helps. I think the same is more or less true for references, too.

lebensterben: gotta file a feature request to lsp server.

lebensterben: actually you should file a request for everything you think reftex backend is doing better than lsp.

real-or-random: Yeah, I'm happy to do this. (That's on my list in fact.)

lebensterben (approved):

layers/+lang/latex/funcs.el
real-or-random: > gotta file a feature request to lsp server.

Okay, I looked into this further. The problem is not the language server, but lsp-mode's sorting. Setting `lsp-completion-sort-initial-results` to nil fixes this, see https://github.com/ROCKTAKEY/lsp-latex/issues/38. 

If I had known this upfront, I probably wouldn't have created this PR here. However, this PR still adds a feature without drawbacks, so I think it's still useful to merge it.

real-or-random:
Force-pushed to update the README. Now it no longer claims that reftex is better, but it's just a valid alternative.

smile13241324 (approved):

PR #15928, opened 2023-02-21
compleseus: improve performance by delaying preview

TheJJ:
- vertico is on elpa now
- adjust preview delay so the preview's startup (loading major mode, fontifying) isn't triggered instantly when selecting a candiate, but rather when lingering on it for a short time.

### DIFF
--- a/layers/+completion/compleseus/packages.el
+++ b/layers/+completion/compleseus/packages.el
@@ -36,24 +36,7 @@
     (selectrum :toggle (eq compleseus-engine 'selectrum))
     (vertico
      :toggle (eq compleseus-engine 'vertico)
-     ;; TODO remove when `vertico-repeat' on ELPA
-     :location (recipe :fetcher github
-                       :repo "minad/vertico"))
-    (vertico-directory
-     :toggle (eq compleseus-engine 'vertico)
-     ;; TODO remove when it's on ELPA
-     :location (recipe :fetcher url
-                       :url "https://raw.githubusercontent.com/minad/vertico/main/extensions/vertico-directory.el"))
-    (vertico-quick
-     :toggle (eq compleseus-engine 'vertico)
-     ;; TODO remove when it's on ELPA
-     :location (recipe :fetcher url
-                       :url "https://raw.githubusercontent.com/minad/vertico/main/extensions/vertico-quick.el"))
-    (vertico-repeat
-     :toggle (eq compleseus-engine 'vertico)
-     ;; TODO: Remove when https://github.com/minad/vertico/issues/83 solved.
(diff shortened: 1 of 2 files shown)


lebensterben (approved):

sebhahn:
is this pull request fixing this issue as well https://github.com/doomemacs/doomemacs/issues/7064?

TheJJ:
yea, that's how I even noticed :)

sebhahn:
Can this be merged soon? Without these changes it is not possible to get a buffer list

layers/+completion/compleseus/packages.el
```diff
@@ -36,24 +36,7 @@
     (selectrum :toggle (eq compleseus-engine 'selectrum))
     (vertico
      :toggle (eq compleseus-engine 'vertico)
-     ;; TODO remove when `vertico-repeat' on ELPA
```
sebhahn, line 39: why did you remove the extension of vertico, they are not an elpa or are they?

TheJJ: they are - the the extension files are part of the https://elpa.gnu.org/packages/vertico.html tar archive.

sebhahn: thanks! it seems that they are not enabled by default, is it possible to add them again?

lebensterben:
fixes #15941

lebensterben:
thx

cremacs1:
removing all of the vertico extensions from `compleseus-packages` made them all inaccessible, they seem to need to be listed. E.g. `compleseus/init-vertico-directory` doesn't work right now, but if you add `vertico-directory` back, it does.

sebhahn:
Yes, I noticed it too and added them back. However, there has been a discussion how to deal with the exentions https://github.com/minad/vertico/issues/83 and it seems they are stuck in a subfolder called "./extensions". Installing vertico from MELPA doesn't add/download the subfolder, but if it would be possible to download them too there are possibilities to load them e.g. using straight or adding them to the load-path

TheJJ:
Hmm, the ELPA archive contains the files directly in the archive root directory, not in a subdirectory, but the git repo contains it in a subdir. The previous hack by defining new packages directly from single files out of the git repo now has to be replaced by something better, the vertico-directory with the module files is in `load-path`, one just needs to load the files.

sebhahn:
the extensions were missing for me, but actually deleting the vertico folder helped, since vertico is now downloaded again including the extension

@thejj do you have an example/config where to put the load-path so that the existing init functions work in the compleseus packages.el, thanks!

TheJJ:
I've submitted #15945 which should now fix the extension loading :)

PR #15932, opened 2023-02-22
modeline: make installation of spaceline-all-the-icons optional

TheJJ:
- make the installation of `spaceline-all-the-icons` optional depending on selected theme
- font-lock+ is no longer used, apparently


smile13241324 (approved):
Looks good.

PR #15940, opened 2023-03-03
Add purpose-toggle-window-purpose-dedicated to key bindings

august-feng:
We already have the key binding to fix a purpose to a window (`SPC r d b`), and this will add the key binding to fix a buffer to that window.

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/keybindings.el
+++ b/layers/+spacemacs/spacemacs-defaults/keybindings.el
@@ -66,6 +66,7 @@
                                        "p"   "projects"
                                        "q"   "quit"
                                        "r"   "registers/rings/resume"
+                                       "rd"  "purpose-toggle-window"
                                        "s"   "search/symbol"
                                        "sa"  "ag"
                                        "sg"  "grep"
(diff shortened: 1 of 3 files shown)


layers/+spacemacs/spacemacs-purpose/packages.el
```diff
@@ -130,7 +130,8 @@
       (spacemacs/set-leader-keys
         "rb" 'purpose-switch-buffer-with-purpose
         "rB" 'switch-buffer-without-purpose
-        "rd" 'purpose-toggle-window-purpose-dedicated
+        "rdb" 'purpose-toggle-window-buffer-dedicated
+        "rdw" 'purpose-toggle-window-purpose-dedicated
```
tko, line 134: I think the `rd` prefix should be labeled somewhere so that which-key shows it with something more informative than `prefix`

august-feng: Good idea, thanks :)

layers/+spacemacs/spacemacs-defaults/keybindings.el
```diff
@@ -66,6 +66,7 @@
                                        "p"   "projects"
                                        "q"   "quit"
                                        "r"   "registers/rings/resume"
+                                       "rd"   "purpose-toggle-window"
```
lebensterben:
```suggestion
                                       "rd"  "purpose-toggle-window"
```

lebensterben (approved):

PR #15944, opened 2023-03-08
org: add org-modern support

TheJJ:
even more beautiful `org-mode` :)

### DIFF
--- a/layers/+emacs/org/README.org
+++ b/layers/+emacs/org/README.org
@@ -34,6 +34,7 @@
   - [[#epub-support][Epub support]]
   - [[#jira-support][Jira support]]
   - [[#valign-support][Valign support]]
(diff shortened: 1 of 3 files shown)


real-or-random:
Nice stuff! I tested this and I prefer to set   `org-modern-hide-stars " "` to keep the headline indentation... I believe this is what most users want, but I'm not entirely sure if we should override the package default here. If not, this should be documented at least.

TheJJ:
I like that the headline is no longer indented :) But yes, this could be documented for easier adjustment.

smile13241324 (approved):
Looks good.

junghan0611:
Its really nice. But I would like to remove org-mode -hook of bottleneck and delay a little bit. I think it is better manually use. How could I do that?

PR #15948, opened 2023-03-09
[command-log] update keybindings & modeline insert after

### LINKED ISSUE #14169: Feature request - keycast - show keybindings and commands in modeline
The https://github.com/tarsius/keycast project displays the Emacs key binding and its command in the mode line.  This is very useful for demos and screencasts to show off the features of Spacemacs, without taking up space in the Emacs frame.

keycast example

command-log (already part of Spacemacs) does show the history of key bindings and commands, although this does take up a significant part of the frame, even if the font is made small.

Should this package be added to Spacemacs directly, or would it be better as its own layer?

**A working configuration - hack**
I have keycast working using the doom-modeline theme and Spacemacs develop.
I added keycast as an `dotspacemacs-additional-package` and added the following code in `dotspacemacs/user-config` which I assume could be greatly improved.
```
  (use-package keycast
                :commands keycast-mode
                :config
                (define-minor-mode keycast-mode
                  "Show current command and its key binding in the mode line."
                  :global t
                  (if keycast-mode
                      (progn
                        (add-hook 'pre-command-hook 'keycast-mode-line-update t)
                        (add-to-list 'global-mode-string '("" mode-line-keycast " "))
                        )
                    (remove-hook 'pre-command-hook 'keycast-mode-line-update)
                    (setq global-mode-string (remove '("" mode-line-keycast " ") global-mode-string)))))
```

I took this code from https://tecosaur.github.io/emacs-config/config.html#fun,code--8 although had to tweak a few commands and remove the custom faces from that code as it interfered with the flycheck tool-tips, which I use to see the linter output in the source code (using clj-kondo).

practicalli-johnny:
- Add keybindings for header and tab bar placement of keycast
- Update command-log documentation
- Update keycast-insert-after to keycast-mode-line-insert-after

tab bar provides places the keycast output in the top right of the frame and output does not hide when using menu bar like other modes dot

header position is show on the top left of the frame (under the tab bar if visible). header keycast output is not visible when using which-key menu or in buffers such as commit_editmsg

keycast modeline replaces the whole modeline with its own output which is not the desired experience. Ideally Keycast should be shown along with other important aspects of the modeline (screenshot in issue #14169)

Resolve: #14169

Thank you :heart: for Spacemacs!

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1831,6 +1831,14 @@ Other:
     - Added ~SPC m s l~ for =coffee-send-line=
     - Added ~SPC m s r~ for =coffee-send-region=
     - Added ~SPC m T c~ for =coffee-cos-mode=
+**** Command-log
+- Key bindings:
+  -  ~SPC a t k h~ display keycast in buffer header (thanks to John Stevenson)
+  -  ~SPC a t k m~ display keycast in modeline (thanks to John Stevenson)
+  -  ~SPC a t k t~ display keycast in tab bar (thanks to John Stevenson)
+- Improvements:
+  - ~keycast-mode-line-insert-after~ variable used to include keycast in
+    mode-line-format (thanks to John Stevenson)
 **** Common Lisp
 - Improvements:
   - Added eval-thing-at-point functions to Common Lisp Layer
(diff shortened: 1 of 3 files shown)


smile13241324 (approved):
Looks ok

PR #15950, opened 2023-03-11
Remove function call of obsolete helm-gtags-define-keys-for-mode

liuxueyang:
The function `spacemacs/helm-gtags-define-keys-for-mode` is obsolete, I remove all of the function calls of it. This will remove the corresponding message in *Message* buffer when Emacs starts.

### DIFF
--- a/layers/+lang/d/packages.el
+++ b/layers/+lang/d/packages.el
@@ -72,4 +72,4 @@
   (spacemacs/counsel-gtags-define-keys-for-mode 'd-mode))
 
 (defun d/post-init-helm-gtags ()
-  (spacemacs/helm-gtags-define-keys-for-mode 'd-mode))
+  )
--- a/layers/+lang/d/packages.el
+++ b/layers/+lang/d/packages.el
@@ -32,7 +32,6 @@
         (flycheck-dmd-dub :requires flycheck)
         ggtags
         counsel-gtags
-        helm-gtags
         ))
 
 (defun d/post-init-company ()
@@ -70,6 +69,3 @@
 
 (defun d/post-init-counsel-gtags ()
   (spacemacs/counsel-gtags-define-keys-for-mode 'd-mode))
-
-(defun d/post-init-helm-gtags ()
-  )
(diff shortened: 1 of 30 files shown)


layers/+lang/d/packages.el
```diff
@@ -72,4 +72,4 @@
   (spacemacs/counsel-gtags-define-keys-for-mode 'd-mode))
 
 (defun d/post-init-helm-gtags ()
```
lebensterben: why not delete the enclosing function as well (the same applies to other changes)

lebensterben (changes requested):

lebensterben:
also please remove https://github.com/syl20bnr/spacemacs/blob/3512e43c9e6efef78b5104785c99518eff60d668/layers/%2Btags/gtags/funcs.el#L56

layers/+lang/d/packages.el
liuxueyang: Removing the enclosing function will get the following warning:
```
(Spacemacs) Warning: package helm-gtags not initialized in layer go, you may consider removing this package from the package list or use the :toggle keyword instead of a `when' form.
```

This warning is generated here: https://github.com/syl20bnr/spacemacs/blob/develop/core/core-configuration-layer.el#L914

The function `go/post-init-helm-gtags` needs to be bound in language layers. `configuration-layer/make-package` will check that.

Or do I need to add `:toggle` keyword for `helm-gtags` in the package list?

lebensterben: simply remove helm-grabs from the <layer>-package variable defined in the beginning of the packages.el files

layers/+lang/windows-scripts/packages.el
```diff
@@ -87,9 +87,6 @@
 (defun windows-scripts/post-init-counsel-gtags ()
   (spacemacs/counsel-gtags-define-keys-for-mode 'bat-mode))
 
-(defun windows-scripts/post-init-helm-gtags ()
-  (spacemacs/helm-gtags-define-keys-for-mode 'bat-mode))
```
lebensterben, line 91: need to remove helm-gtags from the packages list

lebensterben (changes requested):

lebensterben (approved):

smile13241324 (approved):
Looks good.

PR #15955, opened 2023-03-21
fix: typescript/tide don't start when switch from a none-tide project

lawrsp:
move the flycheck setup functions to `-mode-local-vars-hooks`
remove a `flycheck-disable-checker`, which is not neccesary for has `add-to-list flycheck-disabled-checkers`


smile13241324 (approved):
Looks good

PR #15959, opened 2023-03-28
[spacemacs-evil] Only load evil-lisp-state once

bcc32:
This avoids calling `require` over and over every time a new file is opened, which can get expensive if `features` is large.


smile13241324 (approved):
Looks good.

PR #15961, opened 2023-03-29
[version-control] Use upper/lower terminology in smerge-ts

bcc32:
smerge-mode.el has obsoleted the names "mine" and "other" since Emacs 26.1.  Their meanings are not consistent between different VC systems or even different operations (e.g., rebase vs. merge), and "upper" and "lower" are more obvious anyway.

This also changes the Undo key binding in smerge-ts to `U` to avoid a conflict.

### DIFF
--- a/layers/+source-control/version-control/README.org
+++ b/layers/+source-control/version-control/README.org
@@ -2,7 +2,7 @@
 
 #+TAGS: layer|versioning
 
-* Table of Contents                     :TOC_5_gh:noexport:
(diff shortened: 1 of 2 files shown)


bcc32:
BTW, I didn't remove the old key bindings in case users might be accustomed to them, but would be happy to do so if you think it's appropriate.

smile13241324 (approved):
Looks good. I think removing the old bindings would have been fine, anyway I am going to remove them during the merge.

PR #15968, opened 2023-04-07
*spacemacs-navigation/packages.el: load info+ after the info feature

sunlin7:
Error message is emitted when try `info` (press `C-h i`) first time.
> info+·is·not·available:·spaceline-info-mode·disabled

It caused by original code loading the `info+` at the mode hook, too late.

The change will load the `info+` after  the `info`, then there will no error message.
Please help review it. Thanks.


smile13241324 (approved):

smile13241324:
Looks good

sunlin7:
Thanks!

PR #15969, opened 2023-04-07
Openai layer

rommeswi:
This is a first draft for an openai layer. Github organization emacs-openai provides several packages to use OpenAI.
This layer bundles these packages and provides evilification for two of the packages (the others do not need it).
All OpenAI requests are available via the applications/web services keybindings `SPC a w o'.

Features:
send selection/buffer to various openai models
ask openai for code improvement/documentation
generate image from prompt via dall-e
chat with chatgpt

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -501,6 +501,7 @@ In org-agenda-mode
 - hackernews
 - streamlink
 - twitch
+- openai (thanks to Hendrik Rommeswinkel)
 *** Renamed layers
 - =extra-langs= renamed to =major-modes= (thanks to Eivind Fonn)
 - =pdf-tools= renamed to =pdf=
--- a/layers/+web-services/openai/README.org
+++ b/layers/+web-services/openai/README.org
@@ -1,0 +1,72 @@
+#+TITLE: OpenAI layer
+
+#+TAGS: layer|web service|general
+
+
+* Table of Contents                     :TOC_5_gh:noexport:
+- [[#description][Description]]
+  - [[#features][Features:]]
+- [[#install][Install]]
+- [[#configuration][Configuration]]
+- [[#key-bindings][Key bindings]]
+
+* Description
+This layer provides applications to make requests to openai via its API.
+
+** Features:
+- Provides access to openai via the emacs-openai suite of packages
+- Send visual selection or buffer and receive response in current buffer
+- Improve, edit, and document code
+- Chat sessions with chatgpt
+- Obtain AI generated images
+
+* Install
+To use this configuration layer, add it to your =~/.spacemacs=. You will need to
+add =openai= to the existing =dotspacemacs-configuration-layers= list in this
+file.
+
+* Configuration
+You need to set your openai key. If you have stored the key in the environment
+variable OPENAI_API_KEY, then you need to add to your user-config:
+  =(setq openai-key (getenv "OPENAI_API_KEY"))=
+Also, most requests require setting a user, which is done via:
+  =(setq openai-user "user")=
+where "user" can be substituted by other usernames.
+
+
+* Key bindings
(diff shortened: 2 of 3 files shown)


rommeswi:
Here are some possible improvements that might be worth implementing:
- the chatgpt input minibuffer requires the somewhat awkward `<DEL>' `i' to start typing
- it would be more natural to submit the chatgpt buffer content with `C-c C-c'
- the chatgpt input minibuffer is too small, it should scale well by default

lebensterben:
I prefer not to add them to Spacemacs.

This PR mainly adds key bindings to launch the various packages (under SPC a prefix). Users are better off adding their own key bindings. (because it can be much shorter)

rommeswi:
I understand. Once I have figured out the configuration issues that make these packages a bit hard to use in spacemacs, I will make a new pull request where the added value is greater.

The length of the bindings is largely due to the structure of the spacemacs application bindings; I think the "web-services", "chat", "email", and "tools" bindings are just one level too much. I don't think people run that many email clients...

NotBrianZach:
I have a spacemacs copilot layer (mostly adds some useful keybindings to zeroflx package, from https://github.com/rksm/copilot-emacsd)

that maybe could be rolled into this with some modification

 https://github.com/NotBrianZach/spacemacsCopilotLayer

lebensterben:
closing.

the first is what I already mentioned above.

the second reason is more severe, the code snippets you intended to add are not licensed.

rommeswi:
I think this makes sense. It would be best if an experienced spacemacs contributor could look into creating an AI layer or extending existing layers by AI functionality. Inexperienced contributors like myself will probably do more harm than good.

lebensterben:
the major issue is licensing. we cannot accept any code that's not compatible to GPL v3.

rommeswi:
I don't quite understand. As far as I know, I did not suggest introducing any non-GPL v3 code into the spacemacs repository. In the layer I suggested, I added all the license terms in accordance with spacemacs. The packages used are all published under GPL v3. I don't think that writing a layer for these packages would be in conflict with GPL v3 at all.

Perhaps you meant the suggestion by @NotBrianZach for a copilot package? That seems an entirely different idea. I don't think different AIs should be packed into a single layer.

Right now, using the emacs-openai packages requires (for beginners) substantially more effort than just writing "openai" into the list of layers. Given how important AI has become in coding I think it would be good if someone writes a layer for openai.

lebensterben:
> I have a spacemacs copilot layer (mostly adds some useful keybindings to zeroflx package, from https://github.com/rksm/copilot-emacsd)

this is not under a compatible license

rommeswi:
> this is not under a compatible license

and @rommeswi is not @NotBrianZach and neither did this pull request contain anything about copilot.

lebensterben:
> and @rommeswi is not @NotBrianZach 

good point. I was mistaken.

smile13241324:
I see no issue with adding GPL-v3 packages to use a web service though it is an commercial one. I'll be working on this a bit and will merge it later into develop. Keybindings are a bit cumbersome right now but this can be improvement in time.

Lets see if inspiration hits me if not I may just merge it with the original bindings and have the community find shorter ones.

smile13241324:
Lets give AI the place it deserves, I have moved the bindings to global $ which is for AI, there I have made a sub binding for openAI for now. If we add more AI services (hopefully free ones) we could also think about leaving the source binding (openAi) and just have bindings for features like code completion with users having to make a choice which provider to use similar to what we did for the search solution in Spacemacs (Ivy, Helm...) but for now lets keep it simple.

Moving the setup into individual layers, something I have recommended in the related ticket, makes no sense anymore as all languages use the same requests so lets start with this and see how it goes (feedback is welcome).

smile13241324:
Its merged onto develop @rommeswi, thanks for the contribution :)

PR #15975, opened 2023-04-12
*layers/+spacemacs/spacemacs-layouts/packages.el: loading after helm-buffer

sunlin7:
The `(spacemacs//persp-helm-setup)` should be loaded after `helm-buffer` for it using the variable `helm-buffer-list-reorder-fn`.

### DIFF
--- a/layers/+spacemacs/spacemacs-layouts/packages.el
+++ b/layers/+spacemacs/spacemacs-layouts/packages.el
@@ -117,7 +117,7 @@
 
 
 (defun spacemacs-layouts/post-init-helm ()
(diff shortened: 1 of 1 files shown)


lebensterben (approved):

cameron1729:
This change appears to have broken layouts. When checking buffers with `C-x b` I now see **all** buffers instead of buffers specific to the layout. Reverting this change restores the old behaviour.

sunlin7:
@cameron1729 please help verify with https://github.com/syl20bnr/spacemacs/pull/16076. Thanks.

PR #15984, opened 2023-04-19
*core/core-configuration-layer.el: try `package-activate-all' for speedup

sunlin7:
Hi,

Spacemacs maybe initalize more than 100+ packages at startup, it can be speedup by setting the `package-quickstart` to `t` or manually call function `(package-quickstart-refresh)` to generate the `package-quickstart-file`, then call the `(package-activate-all)`  to load the packages in batches.

Otherwise, the `(package-activate-all)` will do nothing.

Please help review this change. Thanks

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -674,6 +674,7 @@ To prevent package from being installed or uninstalled set the variable
                  (not configuration-layer-exclude-all-layers))
         (configuration-layer/delete-orphan-packages packages))))
   ;; configure used packages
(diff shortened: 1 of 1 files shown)


lebensterben:
> If you have many packages installed, you can improve startup times by setting the user option package-quickstart to t. Setting this option will make Emacs precompute many things instead of re-computing them on every Emacs startup. However, if you do this, then you have to manually run the command package-quickstart-refresh when the activations need to be changed, such as when you change the value of package-load-list.

you can change the value in your own early-init file

sunlin7:
Agree, close this PR. Thanks

PR #15986, opened 2023-04-21
*layers/+emacs/org/packages.el: better way to load org-babel langs

sunlin7:
Speedup opening *.org file with lazy loading the babel languages.
The babel languages is used to evaluate the source block in the org file, it's no mandatory for all the org file.
This change will try load the babel language only before evaluate the source block.
It can resolve the stucking when opening an org file on my local.
Please help review this change. Thanks

Spacemacs maybe load these babel languages:
>(http . t)
(restclient . t)
(shell . t)
(elasticsearch . t)
(cfengine3 . t)
(sml . t)
(sql . t)
(shell . t)
(sparql . t)
(ruby . t)
(python . t)
(plantuml . t)
(perl . t)
(js . t)
(java . t)
(hy . t)
(crystal . t)
(groovy . t)
(R . t)
(elixir . t)
(coffeescript . t)
(clojure . t)
(C . t)
(verb . t)
(dot . t)

### DIFF
--- a/layers/+emacs/org/packages.el
+++ b/layers/+emacs/org/packages.el
@@ -122,11 +122,11 @@
     :defer t
     :init
     (progn
-      (defun spacemacs//org-babel-do-load-languages ()
-        "Load all the languages declared in `org-babel-load-languages'."
+      (define-advice org-babel-execute-src-block (:before (&rest _) load-lang)
         (org-babel-do-load-languages 'org-babel-load-languages
-                                     org-babel-load-languages))
-      (add-hook 'org-mode-hook 'spacemacs//org-babel-do-load-languages)
+                                     org-babel-load-languages)
+        (advice-remove 'org-babel-execute-src-block
+                       'org-babel-execute-src-block@load-lang))
       ;; Fix redisplay of inline images after a code block evaluation.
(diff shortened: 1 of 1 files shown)


lebensterben:
this advice should only needs to be run once.

So I suggest to declare a normal named function with defun, which load the languages and remove itself from the advice.

then add the named function to the advice.

sunlin7:
Hi @lebensterben 
Agree with you, and I changed the code to make sure it running once; verified on my local.

lebensterben (approved):
LGTM

layers/+emacs/org/packages.el
```diff
@@ -122,11 +122,11 @@
     :defer t
     :init
     (progn
-      (defun spacemacs//org-babel-do-load-languages ()
-        "Load all the languages declared in `org-babel-load-languages'."
+      (define-advice org-babel-execute-src-block (:before (&rest _) ADV)
```
lebensterben: it could be better if it has a more informative name than ADV

sunlin7: Hi @lebensterben 
I had changed the suffix to "load-lang". It will be more clear. Thanks

sunlin7:
Please help review the latest change. Thanks

smile13241324 (approved):
Looks good

Issue #12418, opened 2019-06-09
Plans to merge `dev` into `master`?


karansag:
Are there any plans to merge `dev` into `master`? There are some features I would like that were merged a while back (https://github.com/syl20bnr/spacemacs/pull/9209/commits) into `dev`. It would be nice to stay on `master`, since I'm assuming that is the most stable branch.

tuxayo:
Or maybe such a release model doesn't fit anymore and the devel branch should be the recommended one?

tuxayo:
Or could there be commits on devel that a posteriori are tagged as stable enough for the most conservative users?

tuxayo:
Or could there be a testing channel which would be freezed at a devel commit.
Then if not major regression are found, this commit would be tagged for release in the stable channel. And the testing channel/branch would jump to a newer commit from devel to test it's stability.
While devel would be like today.

I don't know if these ideas are relevant ^^"

fiveNinePlusR:
they did recently get a release 

https://github.com/syl20bnr/spacemacs/issues/11741

i believe releases will be more common but in general you can just backup your config/cache and update to whatever commit you'd like to use and it's usually pretty stable. might need to update your packages because spacemacs develop is usually developed for the newer packages.

howdoicomputer:
https://github.com/syl20bnr/spacemacs/issues/11741 being closed does not constitute a release. There is no `v0.300` tag or branch. The master branch hasn't seen anything other than a README update in, like, a year. The last Twitter message from the Spacemacs Twitter is from January and there doesn't seem to be any open GitHub issues for discussing cutting a new release.

The `v0.300` [milestone](https://github.com/syl20bnr/spacemacs/milestone/7) was hit but nobody ever drafted a release from it. There is a `v0.301` bug fix milestone that exists despite there not being a `v0.300` release - which is kind of weird. Especially since the description for the `v0.301` milestone is "All issues/PRs assigned to this milestone are for regressions discovered after releasing 0.300.0 that need to be shipped out quickly."

Honestly, it's kind of hard to figure out what is going on. I saw some discussion on setting up [Spacelpa](https://github.com/syl20bnr/spacelpa) and how its progress is a blocker for `0.300` but the last commit for that project was 7 months ago when various packages were added.

[...]

Issue #11352, opened 2018-09-20
A recent release?


hyiltiz:
There have been quite a lot of fixes since the last official release, e.g. https://github.com/syl20bnr/spacemacs/issues/9374. While there is a milestone it hasn't really been updated for about a year. If simply using `develop` branch as a rolling release is officially recommended, I'd be more than willing to switch (some of my friends share this idea too). 

So, is it possible to either release one (maybe even a `-rc` or `-hotfix` release), or a recommendation for the develop branch?

gdkrmr:
I personally use the develop branch and pull every other week and update packages, sometimes things break, mostly because of my local config, but most of the times it goes smoothly.

I have also given up on fixed and stable versions---version 0.300 should be "almost out" for a year now---I rather have the maintainers put their limited time into merging all the open pull requests and keep `develop` up to date with melpa packages and emacs version.

robbyoconnor:
If you want things faster -- use `develop` -- Most consider the `develop` to be the rolling-release channel.

jr0cket:
I have found `develop` to be perfectly than stable for my development of produciton code for well over a year.  I only pull new spacemacs commits when they have something interesting that I can use and I am not on any deadline to deliver something.

I always do a comparison with the .spacemacs template `SPC f e D` and then do a package update.  I always restart Emacs at least once after all this, sometimes twice just to make sure all the packages downloaded.

I am not sure how much sense it makes to keep master if there are not enough resources to do a 6 monthly release of master.  If I knew how to help, I would be happy to do so.

robbyoconnor:
@jr0cket is right -- there were brief moments where `develop` became unstable but I use it as my daily driver -- have since I started using spacemacs. Spacemacs is far from dead.

hyiltiz:
Maybe tentatively tag a commit to be considered `beta` and test for 3 months before it is considered *production quality* and is ready to be merged into `master`? I assume a large portion of users are migrating from Vim (as I am), and are thus new to `lisp`. But we could help organize issues, find solutions etc., as most issues have solutions somewhere out there already, and some example code that shows a fix. This is essentially how it already is working as I see it, but with a production cycle in mind. Alternatively, maybe put "consider `develop` a rolling release" somewhere in FAQ/Wiki so people wouldn't worry too much about jumping into `develop` branch.

robbyoconnor:
I just jumped into emacs lisp and added the features I needed...try doing that -- I went the whole trial by fire route

[...]

PR #15994, opened 2023-04-28
readme: Get rid of deprecated master (almost entirely)

real-or-random:


lebensterben:
people has misunderstood what a rolling release is.

Pushing changes to the develop branch is not a release.

The current status is that we have no release at all.

real-or-random:
@lebensterben Okay, what do you suggest doing?

Should I remove the word rolling release? Should the PR be closed because it's fundamentally the wrong approach?

real-or-random:
To elaborate a bit, this PR here is in line with what @smile13241324 said in https://github.com/syl20bnr/spacemacs/issues/11352#issuecomment-1105708486, and that reflects the current reality of the project.

My feeling is that, if you think there should be releases (https://github.com/syl20bnr/spacemacs/issues/12418#issuecomment-1246112010), that may be a good idea, but that's a discussion that the maintainers should have independently, and this should not block this PR which only documents reality.

lebensterben:
I prefer rolling release with fixed release interval.

But from previous discussion people seems unfavour of this idea. https://github.com/syl20bnr/spacemacs/discussions/15344

bryce-carson:
> people has misunderstood what a rolling release is.
> 
> Pushing changes to the develop branch is not a release.
> 
> The current status is that we have no release at all.

Not a release per se, but commits to the develop trunk are published to the end-users immediately. We aren't witholding bugfixes, features, or new bugs in the current release model. Everything is immediately released. In the least, I agree we have _no_ GitHub releases incoming, upcoming, or planned. We don't release that way. Releasing things to public in other means, yes we continously deploy our in-development development software.

If we want to close this issue with a yes or no, then we should decide on _something_. I think the proposed changes are good, and if you disagree with the terminology used then we should adopt "continously developed, integrated, and deployed" as the moniker for Spacemacs' development and distribution model. To avoid confusion, we can refer users in the documentation the main GitHub issues and pull requests with links, and rather than formally say the proposed moniker in-prose, we could alternatively _not say anything_.

For example, the entire **Update** section of the read me file can be changed to the following without worrying about the history. People interested in the history of the project will view it using the appropriate tools, and people uninterested in it are disserviced by giving them the extra information.

Similar changes can and should be made throughout the documentation currently being published in Spacemacs (everything in the `doc` folder).

# Update
1. Close Emacs.
2. Pull the latest changes from GitHub: `git fetch; git pull --rebase`.
3. Restart Emacs.
4. Update Emacs packages: <kbd>SPC f e U</kbd>.

Would the above be okay with what you were thinking, @real-or-random? I think a more succinct instruction is better, at least in the GitHub README which should not take the place of our more mature and extensive `DOCUMENTATION.org` (that doens't mean it doesn't also need updating).

real-or-random:
> Would the above be okay with what you were thinking, @real-or-random?

Certainly. I prefer the version in my PR here slightly, but I'm of course happy to move forward with anything that updates these instructions. Let's see what others think.

README.md
```diff
@@ -477,17 +477,10 @@ Dock][icon-mac-instructions].
 
 # Update
 
-Spacemacs supports two different update schemes, the default is a rolling update
-scheme based on the latest version of packages available. This version can be
-found on the `develop` branch and is updated by a simple `git pull`.
+Spacemacs relies solely on a rolling update scheme based on the latest
```
lebensterben, line 480: there is no release. it relies solely on pulling updates from the remote.

README.md
```diff
@@ -477,17 +477,10 @@ Dock][icon-mac-instructions].
 
 # Update
 
-Spacemacs supports two different update schemes, the default is a rolling update
-scheme based on the latest version of packages available. This version can be
-found on the `develop` branch and is updated by a simple `git pull`.
+Spacemacs relies solely on a rolling update scheme based on the latest 
+version of packages available. To update Spacemacs, simply pull the latest
+changes of the default `develop` branch:
```
lebensterben, line 482: "default" causes confusion. because previously both develop and master  branches are mentioned, but now only the develop branch is. So it's not clear to user why "default" is emphasized. Is there other non-default branches?

smile13241324: lets remove default and just mention develop.

smile13241324:
Merged and adapted, I hope that's cool for everyone.

PR #15996, opened 2023-04-29
Use cl-remove-if-not instead of remove-if-not

xuhdev:
remove-if-not is obsolete. Additionally, current remove-if-not usage without requiring the cl library often causes "Symbol's function definition is void" error.

### DIFF
--- a/layers/+lang/sql/funcs.el
+++ b/layers/+lang/sql/funcs.el
@@ -25,7 +25,7 @@
   "Update Spacemacs list of sql products"
   (setq
    spacemacs-sql-highlightable sql-product-alist
-   spacemacs-sql-startable (remove-if-not
+   spacemacs-sql-startable (cl-remove-if-not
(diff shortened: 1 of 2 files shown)


layers/+lang/sql/funcs.el
```diff
@@ -20,12 +20,13 @@
 ;; You should have received a copy of the GNU General Public License
 ;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+(require 'cl-lib)
```
lebensterben: this should be avoided

lebensterben (changes requested):

lebensterben:
merged. thanks.

---

for future references:

I requested the change to remove "require" because of this convention.

https://github.com/syl20bnr/spacemacs/blob/0bbb41ec97198b19e1e44ee604caf56596506332/doc/LAYERS.org#no-require

PR #15999, opened 2023-05-01
Require a package dependency of inspector package

bryce-carson:
Fixes an function definition error caused by a missing package dependency in the Emacs Lisp layer.

### DIFF
--- a/layers/+lang/emacs-lisp/packages.el
+++ b/layers/+lang/emacs-lisp/packages.el
@@ -301,6 +301,9 @@
   ;; i.e (require 'company) will not give an error now
   (setq flycheck-emacs-lisp-load-path 'inherit))
 
+(defun emacs-lisp/init-treeview ()
+  (use-package treeview))
+
 (defun emacs-lisp/init-flycheck-package ()
   (use-package flycheck-package
     :hook (emacs-lisp-mode . flycheck-package-setup)))
--- a/layers/+lang/emacs-lisp/packages.el
+++ b/layers/+lang/emacs-lisp/packages.el
@@ -302,7 +302,8 @@
   (setq flycheck-emacs-lisp-load-path 'inherit))
 
 (defun emacs-lisp/init-treeview ()
-  (use-package treeview))
+  (use-package treeview
+    :defer t))
 
 (defun emacs-lisp/init-flycheck-package ()
   (use-package flycheck-package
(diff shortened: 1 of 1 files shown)


sunlin7:
It should consider as an issue of `tree-inspector`.
Using the `Package-Requires:` in its main file can resolve this dependency issue.
https://github.com/mmontone/emacs-inspector/pull/25

bryce-carson:
> It should consider as an issue of `tree-inspector`. Using the `Package-Requires:` in its main file can resolve this dependency issue. [mmontone/emacs-inspector#25](https://github.com/mmontone/emacs-inspector/pull/25)

Thank you for the constructive criticism.

I don't know package development, and I'm really, horribly understudied in the layering and package configuration maintenance of Spacemacs, so it's not surprising that there is a better place for this proposed fix (upstream, of course!). I'll close this pull-request, as it will be made redundant if upstream participates.

sunlin7:
Hi @bryce-carson 
Since the inspector package owner consider it's a package management issue, so neither the Spacemacs nor the inspector won't fix this issue.
Suggest you add the `treeview` to the `dotspacemacs-additional-packages`,

lebensterben:
since upstream refuses to fix the issue. I reopened this PR.

bryce-carson:
> since upstream refuses to fix the issue. I reopened this PR.

I will read the documentation for authoring layers, update this branch, and then request review.

sunlin7:
Currently Spacemacs only target on two inspector functions `inspect-expression` and `inspect-last-sexp`.
https://github.com/syl20bnr/spacemacs/blob/e4b20f797d9e7a03d9a5603942c4a51ea19047b2/layers/%2Blang/emacs-lisp/packages.el#L380-L382

Spacemacs don't use any `tree-inspector-*` functions, no need to depends on `treevew`. 

If Spacsemacs bind a function to `tree-inspector-*`, then we should add `treeview` into the dependency list.

sunlin7:
@bryce-carson  @lebensterben 
Now `tree-inspector` is a standalone package, won't be included in the `inspector` pacakge. 
And the package management can process the dependency correctly.

bryce-carson:
> @bryce-carson  @lebensterben 
> Now `tree-inspector` is a standalone package, won't be included in the `inspector` pacakge. 
> And the package management can process the dependency correctly. 

Okay. I'll upgrade my packages and recompile and see if I get the same warning about a missing function definition.

smile13241324 (approved):

PR #16006, opened 2023-05-07
[go] Enable locally installed packages

allentiak:


allentiak:
Tested on current Debian "testing/unstable".

lebensterben:
user should set PATH correctly themselves

allentiak:
Yes, but that only works if you start Emacs from your terminal. If you do it from a GUI, it does not work.

allentiak:
If I follow the current instructions from the Go layer README, every locally installed package is accessible from my terminal, but not from Spacemacs.

@lebensterben I think this PR should be discussed further. The problem is there. If this is not the best way to fix it, I would like to hear other options.

allentiak:
@lebensterben Could you please reopen this?

I think this is a good place to have this discussion.

lebensterben:
@allentiak that's entirely false.

all process inherits the environment variables from its parent process. it doesn't matter whether it's the login shell or the terminal.

check your .profile or .xprofile.

lebensterben:
there's no problem with go layer or spacemacs or emacs.

the only problem is your inability to set environment variables correctly which is entirely out of the scope of this project.

allentiak:
...

PR #16052, opened 2023-06-23
Remove superfluous progns in use-package

equwal:
Remove a level of indentation for all these progns inside use-package. The
use-package macro is already an implicit progn. To review this commit,
since there is a lot of whitespace change, use

    git diff HEAD -w

it doesn't do anything other than that.


smile13241324 (approved):

equwal:
@smile13241324 Thanks, I accidentally only did one sub-folder of layers, `+lang`. 

https://github.com/syl20bnr/spacemacs/pull/16052

fixed that here

PR #16053, opened 2023-06-24
Remove all redundant progns in layers

equwal:
In my other commit I apparently fell asleep at the keyboard and forgot all the redundant use-package progns in the layers that aren't under `+lang`. Commit here.

I also removed all the redundant progns everywhere else, which accounted for about 50% of the progns used in the layers! It's in a separate commit to make review easy.

Like last time, this is easy to review with
`git diff HEAD -w`
and since the commit for the use-packages is really big, I suggest trying
`git diff HEAD -w | grep -B1 progn `
where you can see that all the progn edits are preceded by :init or :config in use-package.

Finally spacemacs is rid of the cancerous plague of redundant progns. No newbies will copy them to make their own packages anymore!


equwal:
https://github.com/syl20bnr/spacemacs/pull/16052

equwal:
As a side-note for the reviewer, you might want to cherry pick this one next to my previous commit and squash it.

smile13241324:
Merged manually

PR #16065, opened 2023-07-07
fix: restore SPC j j to use evil-avy-goto-char-timer

pd9333:
Commit 4ec0b8b88f10fa9ac9c42811b65030532664d64d accidentally changed "SPC j j" to call evil-avy-goto-char. This commit reverts it back to call evil-avy-goto-char-timer, which was introduced by commit 4c231ae28220d28ae0a63634683b171b96c8444d

### DIFF
--- a/layers/+spacemacs/spacemacs-editing/packages.el
+++ b/layers/+spacemacs/spacemacs-editing/packages.el
@@ -55,8 +55,7 @@
     (setq avy-background t)
     (spacemacs/set-leader-keys
       "jb" 'avy-pop-mark
-      "jj" 'evil-avy-goto-char
-      "jJ" 'evil-avy-goto-char-2
+      "jj" 'evil-avy-goto-char-timer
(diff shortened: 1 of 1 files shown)


dankessler:
It looks like there were some other unintended changes that snuck in with https://github.com/syl20bnr/spacemacs/commit/4ec0b8b88f10fa9ac9c42811b65030532664d64d. Namely, it looks like the [spacemacs-editing layer's package.el](https://github.com/syl20bnr/spacemacs/commit/4ec0b8b88f10fa9ac9c42811b65030532664d64d#diff-1cb355d2a8f6cd984f656c92fdef1400c375fd9b1c50bb68af74eb812109213d) gets dramatically edited. This resulted in `SPC j j` getting remapped, but the unintended edits are far more wide-ranging than that.

I skimmed the rest of the diff, and I think that this is the only file that got accidentally clobbered.

Issue #16044, opened 2023-06-14
Python layer: activated virtual environment is ignored when starting the REPL


elatomo:
#### Description :octocat:

When starting a python REPL from an activated virtual environment, the default Python interpreter is used instead of the defined in the virtual environment.

#### Reproduction guide :beetle:
- Start Emacs
- Open a python file
- Activate a virtual environment (`pyvenv-workon`)
- Start the REPL (`spacemacs/start-or-switch-repl`)

*Observed behaviour:* :eyes: :broken_heart:
- Even though the `PATH` contains the virtual environment (`getenv PATH`), the Python interpreter used is the default Python interpreter in the system instead of the one defined in the virtual environment.

*Expected behaviour:* :heart: :smile:
- The Python interpreter launched is the one defined in the virtual environment  (the `PATH` is respected). 

#### System Info :computer:
- OS: gnu/linux
- Emacs: 28.2
- Spacemacs: 0.999.0
- Spacemacs branch: develop (rev. d0d014160)
- Graphic display: t
- Running in daemon: nil
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(auto-completion better-defaults emacs-lisp common-lisp
                 (python :variables python-test-runner 'pytest python-formatter 'black)
                 (typescript :variables typescript-indent-level 2)
                 rust git helm markdown yaml html docker
                 (org :variables org-want-todo-bindings t)
                 (shell :variables shell-default-shell 'vterm shell-default-height 30 shell-default-position 'bottom)
                 syntax-checking treemacs emoji)
```
- System configuration features: ACL CAIRO DBUS FREETYPE GIF GLIB GMP GNUTLS GPM GSETTINGS HARFBUZZ JPEG JSON LCMS2 LIBOTF LIBSYSTEMD LIBXML2 M17N_FLT MODULES NOTIFY INOTIFY PDUMPER PNG RSVG SECCOMP SOUND THREADS TIFF TOOLKIT_SCROLL_BARS X11 XDBE XIM XPM GTK3 ZLIB

elatomo:
I took a quick look to the commit history, and it appears that https://github.com/syl20bnr/spacemacs/commit/cd0eeb9f7bbcc8a752bb344e751aa31a70dc06a6 is potentially causing the issue. To confirm this, I performed a local revert of the change, and the Python REPL now launches properly without any issues.

ghost:
[...]

PR #16074, opened 2023-07-17
*layers/+lang/python/packages.el: fix default python-shell-interpreter

sunlin7:
Fix the #16044, the default `python-shell-interpreter` changed to be absolout path will always be used without switch to local value.


smile13241324:
This looks good, but we seem to have some merge conflicts, would you mind having a quick look @sunlin7?

sunlin7:
Hi @smile13241324  
I had rebased to the latest develop branch. 
Please help review it, thanks.

smile13241324 (approved):

PR #16075, opened 2023-07-17
Documents enhancement for the use-package

sunlin7:
Minor document changes for the `use-package`.


smile13241324 (approved):

smile13241324:
Good catch ;)

PR #16107, opened 2023-08-04
*layers/+lang/python/funcs.el: use the interpreter basename

sunlin7:
Try to fix the last issue discussed in the https://github.com/syl20bnr/spacemacs/issues/16044#issuecomment-1665595535, use the basename without full path python-shell-interpreter.


sunlin7:
Verified by @thedohn

smile13241324 (approved):

PR #16123, opened 2023-08-16
[go] Add key-bindings for "go generate"

lawrsp:
`go generate` is a powerfull tool in golang, these two key-bindings make it easy to use:

 ```
| ~SPC m x g~   | run "go generate" for the current file                                                |
| ~SPC m x G~   | run "go generate" for the current package                                             |
````


smile13241324 (approved):
Looks good

Issue #4242, opened 2015-12-20
Escape does not close autocomplete window.


jordwalke:
Hi, I'm brand new to Emacs so thank you for your patience. I'm only even trying Emacs thanks to Spacemacs. The issue I'm seeing is that when hitting escape, Evil goes into normal mode as expected, but any autocomplete window that is open does not close.

This is my entire config:

```
  (load-theme 'spacemacs-light t)
  (set-face-attribute 'mode-line nil :box nil)
  (set-face-attribute 'mode-line-inactive nil :box nil)
  ;; Don't highlight line
  (global-hl-line-mode -1) ; Disable current line highlight
  (setq evil-want-fine-undo nil)
  (setq evil-move-cursor-back nil)
  (global-set-key (kbd "s-p") 'helm-M-x)
  ; (setq powerline-default-separator 'utf-8)
  (setq powerline-default-separator nil)
  (global-vi-tilde-fringe-mode -1)
```

These are my enabled layers:

```
     company
     ;; To toggle autocompletion on:
     ;; space t a 
     auto-completion
     ;; better-defaults
     emacs-lisp
     themes-megapack
     ;; Basic mac keybindings. (Cmd is super, alt is meta)
     osx

```

I'm on Mac OS Yosemite.

TheBB:
Old bug, no activity. I just tried reproducing this on develop but I can't. Is this still a problem?

Stebalien:
I'm seeing the same thing, but only with some company backends (??). Worse, it doesn't really appear to fully exit insert mode (keys still get inserted until the popup closes).

Specifically, I can reproduce this with `company-emoji` and `notmuch-company`.

NJBS:
@Stebalien Can you still reproduce this? I'm unable to reproduce the issue using `company-emoji`.

#### System Info :computer:
- OS: windows-nt
- Emacs: 24.5.1
- Spacemacs: 0.200.3
- Spacemacs branch: develop (rev. b550945)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(auto-completion better-defaults bibtex c-c++ dash emacs-lisp git github gtags html java javascript latex markdown ocaml org python racket ranger rust shell shell-scripts helm syntax-checking emoji)
```

Stebalien:
Yes, I still can.

1. Open a scratch buffer.
[...]

PR #16140, opened 2023-09-07
fix(evil): Escape and evil-escape do not close Company auto-complete

niontrix:
Why?:
- Pressing Escape or calling evil-escape using "fd" or your configured keybinding does not close Company's auto-complete pop-up especially using LSP. See also: https://github.com/syl20bnr/spacemacs/issues/4242

- Add hook to call company-abort when entering evil-normal-state

I'm actually not that sure that the code is in the right place. Maybe I should move it to the evil setup in bootstrap.
Can someone point me in the right direction?

### DIFF
--- a/layers/+spacemacs/spacemacs-evil/packages.el
+++ b/layers/+spacemacs/spacemacs-evil/packages.el
@@ -138,6 +138,9 @@
                   lazy-load-evil-escape-2)))
     :config
     (add-hook 'spacemacs-editing-style-hook #'spacemacs//evil-escape-deactivate-in-holy-mode)
+    (add-hook 'evil-normal-state-entry-hook (lambda ()
+                                              (when (and (fboundp 'company--active-p) (company--active-p))
+                                                (company-abort))))
     ;; apply once when emacs starts
     (spacemacs//evil-escape-deactivate-in-holy-mode dotspacemacs-editing-style)
     (spacemacs|hide-lighter evil-escape-mode)))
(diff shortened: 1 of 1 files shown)


layers/+spacemacs/spacemacs-evil/packages.el
```diff
@@ -138,6 +138,9 @@
                   lazy-load-evil-escape-2)))
     :config
     (add-hook 'spacemacs-editing-style-hook #'spacemacs//evil-escape-deactivate-in-holy-mode)
+    (add-hook 'evil-normal-state-entry-hook (lambda ()
+                                              (when (and (fboundp 'company--active-p) (company--active-p))
```
dalanicolai: I would suggest instead of using `(fbound 'company--active-p)` to wrap the thing in a `(with-eval-after-load 'company ...)` so that the hook is only added when company gets loaded.

niontrix: Thanks for the hint. But I would rather check if `company--active-p` is bound, because it is not a public function and thus could be removed or renamed, which would then cause an error and be a bad UX. What do you think?

pataquets: Both @dalanicolai's and your point are compatible, if I'm not wrong. I see benefit of not adding the hook at all until company gets loaded.

syl20bnr:
As it has safe guards I would rather put this into evil-escape directly. What do you think ?

smile13241324:
I have moved your fix into evil-escape as this was rather a package issue.

PR #16141, opened 2023-09-07
*layers/+intl/chinese/packages.el: remove incorrect config for pyim-page-tooltip

sunlin7:
Hi,

Setting the `pyim-page-tooltip` to `t` is incorrect according the change https://github.com/tumashu/pyim/commit/92d24bca089f740c33861318eba5d90cfb07bb8d#diff-95b9775e78bc8daa95cebb53205ccec621592456d4fc14dd893d2c52234dba51L48, a symbol or a list is desired.

And the default list value works great, so this change just remove the line to untouch the default value.

Please help review it. Thanks

### DIFF
--- a/layers/+intl/chinese/packages.el
+++ b/layers/+intl/chinese/packages.el
@@ -53,8 +53,7 @@
     :if chinese-default-input-method
     :defer t
     :init
-    (setq pyim-page-tooltip t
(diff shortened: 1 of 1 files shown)


smile13241324 (approved):

PR #16149, opened 2023-09-19
Add direnv layer.

greydot:
Basically, just that. Direnv package requires no configuration and just works if direnv is present and enabled in the current project directory.

### DIFF
--- a/layers/+tools/direnv/README.org
+++ b/layers/+tools/direnv/README.org
@@ -1,0 +1,16 @@
+#+TITLE: Direnv layer
+
+#+TAGS: layer|tool
+
+* Description
+This layer adds direnv support to Spacemacs.
+
+** Features:
+  - Direnv support
+
+* Install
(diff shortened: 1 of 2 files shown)


bcc32:
I think the direnv package requires at least some configuration.  Either the user needs to enable `direnv-mode` (a global minor mode), or they need to know to invoke `direnv-update-environment` and `direnv-allow` and so on, or else their `.envrc`s will have no effect.

greydot:
> I think the direnv package requires at least some configuration. Either the user needs to enable `direnv-mode` (a global minor mode), or they need to know to invoke `direnv-update-environment` and `direnv-allow` and so on, or else their `.envrc`s will have no effect.

`direnv allow` should be called from the terminal manually, but this is an expected workflow. I wouldn't want my emacs setup to do this for me.

Sorry, I switch to doom emacs a few months ago and won't be able to follow up with this PR.

smile13241324:
I don't think having the users do these things manually is very much Spacemacsy to be honest. Given that this layer is very minimal and there is no follow up work to be expected I go ahead and close it for now.

PR #16153, opened 2023-09-24
update compleseus keybindings

cremacs1:
Update several keybindings for compleseus, replacing obsolete commands or binding keys which are present in helm but not in compleseus.

`consult-apropos` was [deprecated](https://github.com/minad/consult/commit/eacc143a4439a394b5da2124e2550816c66c6d41) and seemingly removed. Replace it with the default `apropos-command`.

`describe-mode` was never bound in compleseus.


smile13241324 (approved):

PR #16154, opened 2023-09-25
[Common Lisp] Make SPC m ' switch to repl if already connected.

einsiedlerspiel:
Make `SPC m '`  switch to REPL if slime is already connected instead of asking if another inferior process should be started.


smile13241324 (approved):

PR #16167, opened 2023-10-25
Fix bug: wrong docker build no-cache command

sfavazza:


smile13241324 (approved):
Looks ok, thanks for your contribution 👍

PR #16198, opened 2023-11-23
Update latex layer README.org

rommeswi:
Explains more clearly how to install texlab

Texlab can be built in every latex project folder or installed.
For Emacs, it is preferable to install it as Emacs then handles the build in every project folder.
It took me longer than it should have to figure this out and therefore I suggest updating the documentation.


smile13241324 (approved):

smile13241324:
Changes look good but the doc generator complains, I am going to check this before this can be merged.

PR #16199, opened 2023-11-24
[auto-completion] Do not erase yas-snippet-dirs set by other layers

bcc32:
At my site, we have an internal package that adds some entries to yas-snippet-dirs (in a with-eval-after-load on yasnippet).  These are currently getting erased by the auto-completion layer.  It seems to make more sense for the latter to simply add entries, rather than replacing the whole list.

(Normally yasnippet is not loaded until auto-completion layer is initialized, however, when starting Spacemacs for the first time, it seems to be loaded earlier, in which case the `use-package yasnippet :init` section runs _after_ yasnippet is loaded.  In any case, the `defvar` seems more correct than blindly calling `setq`)

### DIFF
--- a/layers/+completion/auto-completion/packages.el
+++ b/layers/+completion/auto-completion/packages.el
@@ -275,6 +275,7 @@
     :init
     ;; We don't want undefined variable errors
     (defvar yas-global-mode nil)
+    (defvar yas-snippet-dirs nil)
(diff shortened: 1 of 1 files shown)


sunlin7:
Moving them to the `:config` section may be better.

smile13241324 (approved):

PR #16078, opened 2023-07-19
Rust layer overhaul

whirm:
- Switch to Rustic, a more complete Rust mode
- Make rust-analizer backend the default and remove support for RLS and Racer (both unmantained and obsoleted by rust-analyzer)
- Add GDB Rust DAP template

PR #16217, opened 2023-12-23
Fix inconsistencies and broken key bindings in the Rust layer

bcc32:
The Rust layer was recently overhauled in #16078, but unfortunately the large rework caused some breaking changes.

This PR attempts to fix all the breakages I've noticed (although there may be others).

Unfortunately, Rustic does not support "repeating" the last cargo command, and also it is much more inconvenient to try to "auto-reload" rust-analyzer because Rustic runs all cargo commands asynchronously, so I have dropped that functionality (which was not working correctly anyhow).

(If folks would like me to attempt to resurrect that feature, I can try to do so, but most likely the best way to do that would be a patch to Rustic itself.)

### DIFF
--- a/layers/+lang/rust/README.org
+++ b/layers/+lang/rust/README.org
@@ -12,7 +12,7 @@
   - [[#lsp][LSP]]
     - [[#autocompletion][Autocompletion]]
     - [[#debugger-dap-integration][Debugger (dap integration)]]
-    - [[#automatically-reload-workspace][Automatically Reload Workspace]]
+    - [[#reloading-workspace][Reloading Workspace]]
   - [[#cargo][Cargo]]
     - [[#cargo-edit][cargo-edit]]
     - [[#cargo-outdated][cargo-outdated]]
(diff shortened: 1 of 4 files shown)


whirm:
There's `rustic-cargo-run-rerun`, I'm aware it doesn't do exactly the same but maybe we can bind that in its place?

`lsp-rust-analyzer-(re)run` is what I personally use and I'm not sure I see the value of repeating the last cargo command other than for `cargo run` + args, which is already implemented in rustic.

Regarding auto reloading rust-analyzer. I'd say it just works by default. If you modify `Cargo.toml`  outside of emacs (IE by calling `cargo add foobar` or by modifying it with a different editor) RA will notice and reload by itself. Just double checked on my machine.

FWIW, I've reviewed and I'm running your changes locally and everything looks fine to me.

Thanks @bcc32 and @smile13241324!

bcc32:
Thanks @whirm.  I've incorporated your suggestion w.r.t. `,c.` and also reworked the comment about auto-reloading the workspace to reflect the current state of the world.

PR #16225, opened 2023-12-30
[shell] Use explicit-shell-file-name and $ESHELL as default shells

bcc32:
M-x term and M-x ansi-term both check these two values before using
shell-file-name.  The description of the variable suggests that
explicit-shell-file-name should be used for user-requested shells,
whereas shell-file-name may be used (if different) for automatically
spawned shells for things like compilation.

Since shell-pop is more like a wrapper around M-x term, it seems
appropriate to check these variables first, the same way the
interactive form of M-x term does.

### DIFF
--- a/layers/+tools/shell/config.el
+++ b/layers/+tools/shell/config.el
@@ -49,7 +49,9 @@ for Linux/macOS), `eshell' (default for windows), `shell', `term', `vterm',
 (defvar shell-default-width 30
   "Width in percents for the shell window.")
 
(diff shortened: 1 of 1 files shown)


bcc32:
Sorry, this is actually a mistake, since `shell.el`, which defines `explicit-shell-file-name`, is not loaded at this point.

I still think the fact that this layer `setq`s a defcustom like this is problematic, but I'm withdrawing this PR for now.

PR #16245, opened 2024-01-10
Migrate most uses of defadvice to advice-add (or define-advice)

bcc32:
defadvice is more complex and has many quirks, and has long been
deprecated in favor of advice-add and related mechanisms.

I did not change the occurrences in the spacemacs-purpose layer, which
are a decent bit more complex and also appear to be derived from some
upstream.  Similarly, I did not modify frame-cmds.el, which is a copy
of an upstream file.

### DIFF
--- a/core/aprilfool/zemacs.el
+++ b/core/aprilfool/zemacs.el
@@ -109,18 +109,18 @@
          'type 'help-url
          'help-args (cdr l))))))
 
-(defadvice configuration-layer/initialize (before zemacs/initialize activate)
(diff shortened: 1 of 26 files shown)


smile13241324 (approved):

macovsky:
@bcc32 it broke `expand-region` for me with:

    symbol's value as variable is void: return-value
    
do i have to fix something in my config?

bcc32:
Sorry @macovsky!  This should be fixed by #16256

macovsky:
> Sorry @macovsky! This should be fixed by #16256

thanks so much!

PR #16247, opened 2024-01-13
Update/improve prefix functionality in window delete commands

dalanicolai:
(Spac)Emacs comes with the `kill-buffer-and-window` by default (on `SPC w x`). Also, deleting other windows (and buffers) simply by prefixing the window delete commands with a numerical prefix is very convenient. Therefore, this commit updates (or adds) the functionality of the prefix in the window delete commands.

Summary: to delete another window (and optionally its buffer), simply prefix `SPC w d` (or `SPC w x`) with the window number.

This functionality exists in the `winum` package itself (on `winum-select-window-by-number`), but for deletion that command requires a negative prefix which is cumbersome.

### DIFF
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -2338,7 +2338,7 @@ Windows manipulation commands (start with ~w~):
 | ~SPC w c C~            | toggle visual distraction free mode                                         |
 | ~SPC w c .~            | center buffer and enable centering transient state                          |
 | ~SPC w d~              | delete a window                                                             |
-| ~SPC u SPC w d~        | delete a window and its current buffer (does not delete the file)           |
+| ~NUM SPC w d~          | delete window number NUM                                                    |
 | ~SPC w D~              | delete another window using [[https://github.com/abo-abo/ace-window][ace-window]]                                      |
 | ~SPC u SPC w D~        | delete another window and its current buffer using [[https://github.com/abo-abo/ace-window][ace-window]]               |
 | ~SPC w t~              | toggle window dedication (dedicated window cannot be reused by a mode)      |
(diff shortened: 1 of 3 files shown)


doc/DOCUMENTATION.org
```diff
@@ -2338,7 +2338,7 @@ Windows manipulation commands (start with ~w~):
 | ~SPC w c C~            | toggle visual distraction free mode                                         |
 | ~SPC w c .~            | center buffer and enable centering transient state                          |
 | ~SPC w d~              | delete a window                                                             |
-| ~SPC u SPC w d~        | delete a window and its current buffer (does not delete the file)           |
```
smile13241324, line 2341: I would like to keep the old behaviour, users should have a choice between the `spc u` way and the pure `num` flavor.

smile13241324 (changes requested):
I have found a small issue, please check and ping me when you are ready for another review.

doc/DOCUMENTATION.org
dalanicolai: There already exist two bindings for deleting a window and its buffer: `SPC w x` and `SPC b x`. Therefore, I argued the binding you are referring to, is somewhat redundant.

This PR suggests that we can make better use of the `SPC w d` prefix, namely, to use it for quickly deleting other windows.
Currently, the only way to quickly delete another window is via`ace-delete-window` for which there is no default binding. However, I would argue that using a numeric prefix with `SPC w d` is a better alternative to `ace-delete-window`, as ace shows labels only after invoking it, while a window number is known already before invoking `SPC w d`. Furthermore, unlike `ace-delete-window`, this alternative does not require creating some extra (custom) binding, while also, it does not depend on ace.

fnussbaum: I think adding this numerical prefix support is a good idea and would be happy to see it merged. Though keeping the old <kbd>SPC u</kbd>-prefix behaviour would also be preferable to me.

Hence I have set up a draft PR implementing this including a few changes. It handles some edge cases like entering a non-existing window number (which should not delete the current window instead) or having the package winum excluded.
Also do we actually need to reimplement the `kill-buffer-and-window` function? Wrapping it in a `(with-selected-window ...)` form seems safe to me.
@dalanicolai Would you mind having a quick look at #16425? I have added you as a co-author if that's fine with you.

> Currently, the only way to quickly delete another window is via `ace-delete-window`

There exists <kbd>SPC u SPC NUM</kbd> as an alternative, but I still think it would be nice to support <kbd>NUM SPC w d</kbd>, especially in combination with the proposed <kbd>NUM SPC w x</kbd>.

smile13241324:
I think this was waiting too long already

PR #16261, opened 2024-01-24
fix: Spacemacs-Modeline showing duplicated Anzu indicator

niontrix:
Why?:
- As we can see in the screenshot, if we have some search results that Anzu should be indicating, Spacemacs-Modeline duplicates the indicator. This is caused by Anzu consing itself on to modeline-format.

- Setting anzu-cons-mode-line-p to nil during package initialization

Image showing the problem:

image

### DIFF
--- a/layers/+spacemacs/spacemacs-modeline/packages.el
+++ b/layers/+spacemacs/spacemacs-modeline/packages.el
@@ -31,6 +31,11 @@
         symon
         (vim-powerline :location (recipe :fetcher local))))
 
+(defun spacemacs-modeline/init-anzu ()
+  (use-package anzu
+    :defer t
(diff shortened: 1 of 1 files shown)


bcc32:
There's actually this snippet in `spcemacs-evil/init-evil-anzu`:

```
    (when (eq (spacemacs/get-mode-line-theme-name) 'spaceline)
      (setq anzu-mode-line-update-function 'spacemacs/anzu-update-mode-line
            anzu-cons-mode-line-p nil))))

```

Maybe it's just not working correctly?  What does `(spacemacs/get-mode-line-theme-name)` evaluate to in your emacs?

smile13241324:
The mentioned function `spacemacs-evil/init-evil-anzu` is not testing against the right string, I am correcting this.

PR #16263, opened 2024-01-25
Prevent M-x {enable,disable}-command from modifying ~/.emacs.d/init.el

bcc32:
This PR advises the built-in function `en/disable-command` (defined in `novice.el`) to add code to `dotspacemacs/user-config` instead of `~/.emacs.d/init.el`.  For Spacemacs users, modifications in `~/.emacs.d/init.el` are potentially inconvenient (may block `git checkout`) and likely not visible (since they expect their entire config to be contained in `~/.spacemacs`).

If you would like, I would also be okay with simplifying this advice to just append to `.spacemacs` (outside of `dotspacemacs/user-config`) instead of `~/.emacs.d/init.el`, which should still work acceptably.

### DIFF
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -1257,4 +1257,32 @@ Return non-nil if all the tests passed."
                          :initial-value t)
             (goto-char (point-min))))))))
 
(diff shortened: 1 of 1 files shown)


smile13241324 (approved):
looks really good 😊

Issue #16282, opened 2024-02-14
Bindings from latex layer are not available when editing latex files


dankessler:
#### Description :octocat:

Even with the `latex` layer enabled, when editing `latex` files (that end in `.tex`) none of the major mode bindings are available and also `lsp` falls down.

#### Reproduction guide :beetle:
- Start Emacs
- Visit a file ending in `.tex`
- Press `SPC m`

*Observed behaviour:* :eyes: :broken_heart:
RE key bindings, observe this in minibuffer
```
Minibuffer output: SPC m-  which-key: There are no keys to show
```

RE LSP: this message appears (after visiting a `.tex` file)
```
LSP :: There are no language servers supporting current mode `LaTeX-mode' registered with `lsp-mode'.
This issue might be caused by:
1. The language you are trying to use does not have built-in support in `lsp-mode'. You must install the required support manually. Examples of this are `lsp-java' or `lsp-metals'.
2. The language server that you expect to run is not configured to run for major mode `LaTeX-mode'. You may check that by checking the `:major-modes' that are passed to `lsp-register-client'.
3. `lsp-mode' doesn't have any integration for the language behind `LaTeX-mode'. Refer to https://emacs-lsp.github.io/lsp-mode/page/languages and https://langserver.org/ .
4. You are over `tramp'. In this case follow https://emacs-lsp.github.io/lsp-mode/page/remote/.
5. You have disabled the `lsp-mode' clients for that file. (Check `lsp-enabled-clients' and `lsp-disabled-clients').
You can customize `lsp-warn-no-matched-clients' to disable this message.
```

*Expected behaviour:* :heart: :smile:
`which-key` pop-up showing the many useful commands I can run when editing a latex file.

Also, I expect lsp support to work (since I have `texlab` installed)

#### System Info :computer:
- OS: darwin
- Emacs: 29.1
- Spacemacs: 0.999.0
- Spacemacs branch: develop (rev. df8fef83b)
- Graphic display: t
- Running in daemon: nil
- Distribution: spacemacs
- Editing style: hybrid
- Completion: helm
- Layers:
```elisp
((auto-completion :disabled-for org)
 better-defaults
 (bibtex :variables bibtex-enable-ebib-support t)
 csv emacs-lisp
 (ess :variables ess-assign-key "\255" ess-use-flymake nil)
 eww git graphviz haskell
 (helm :variables helm-use-fuzzy 'source helm-buffer-max-length nil)
 html json latex lsp
[...]

PR #16283, opened 2024-02-14
update latex layer to track new AUCTeX naming conventions

dankessler:
AUCTeX has changed its naming conventions, e.g., latex-mode is now LaTeX-mode. In order to support both users of the "new" AUCTeX and the "old" AUCTeX, I've configured the keybinding part of the latex layer to configure both old and new version of these modes.

See #16282 for more discussion.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2737,6 +2737,7 @@ files (thanks to Daniel Nicolai)
 - Automatically configure =pdf-tools= as pdf viewer for =TeX-view= command (by default) (Thanks to Daniel Nicolai)
 - Add layer variable to open =pdf-tools= in split window (Thanks to Daniel Nicolai)
 - New package =evil-tex= (thanks to Dan Kessler)
+- Adapt to new [[https://git.savannah.gnu.org/cgit/auctex.git/tree/doc/changes.texi][AUCTeX mode naming conventions]] (thanks to Dan Kessler)
 **** Lua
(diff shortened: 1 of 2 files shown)


arifer612:
Tested out the proposed changes and it works great on lsp v20240215.1823. The
changes are backwards compatible too (b^-^)b

dankessler:
I just noticed that my PR doesn't fix all the latex-related keybinds, e.g., I missed fixing the way `reftex` keys are bound (which are still only available in actual `latex-mode`).

I've marked the PR as a draft so I can fix this, re-push, and then I'll mark ready for review.

Rather than littering `latex`'s `packages.el` file with every more dolists and code duplication, I'm just going to add some logic at the top that figures out whether `latex-mode` or `LaTeX-mode` is appropriate, sets some private variables, and then uses those throughout. Hopefully that will still give backwards compatibility while reducing code bloat.

dankessler:
Upon further reflection, I think this is better fixed upstream in `bind-map`. I've opened a [PR there](https://github.com/justbur/emacs-bind-map/pull/9) that should fix (the keybinds part of) #16282 *without* requiring this PR. I'm going to leave this PR in draft status for now, and if the PR against `bind-map` is adopted we can scrub it. If the fix can't go into `bind-map`, then we can revisit a workaround based on this PR.

dankessler:
This has been fixed (I think?) upstream now

PR #16306, opened 2024-03-08
feat: Add a config option for flycheck-display-error-delay

niontrix:
Why?:
- If we provide an option to hide the flycheck popup after a given amount of time we should also provide an option to change flychecks delay before displaying the popup, because it can get in the way when trying to complete an unfinished statement.

- Add config option syntax-checking-tooltips-delay
- Set value of flycheck-display-errors-delay to provided value or flychecks default of 0.9 on init

### DIFF
--- a/layers/+checkers/syntax-checking/config.el
+++ b/layers/+checkers/syntax-checking/config.el
@@ -35,6 +35,12 @@
 If non-positive or nil, do not hide tooltip."
   '(number))
 
+(spacemacs|defc syntax-checking-tooltips-delay nil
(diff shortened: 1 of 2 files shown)


niontrix:
Something went wrong while trying to get the dependencies in the CI tests. Can those tests be run again?

smile13241324 (approved):

smile13241324:
no idea what went wrong with ci, but it does not look like its coming from your change, normally a force push on your branch should retrigger the jobs though.

niontrix:
@smile13241324  Thanks for merging and thanks for the hint, I'll try that next time!

PR #16316, opened 2024-03-20
Remove `''` from pdf-outline-buffer-mode-map and update documentation

dankessler:
I noticed some noise in my *Messages* buffer about keybinding whenever I tried to view a PDF, and I tracked it down.

`pdf-outline-buffer-mode-map` gets evilified. Prior to this, `SPC` is mapped to `pdf-outline-display-link`. During evilification, this gets remapped to `'` (this remapping is, AFAICT, not documented in evilification but you can see it
[here](https://github.com/syl20bnr/spacemacs/blob/4a227fc94651136a8de54bcafa7d22abe1fa0295/layers/%2Bdistributions/spacemacs-bootstrap/local/evil-evilified-state/evil-evilified-state.el#L323) (note: `32` means `SPC`).

We then try to map the key `''` to `pdf-outline-move-to-current-page`, but this isn't possible, because `'` is already mapped to a function and is thus not a prefix key.

Fortunately, this behavior is already available via double-backtick: ` `` `.

I've removed the (failed) mapping of `''` and also removed it from the documentation.

### DIFF
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -895,7 +895,9 @@ behave more like Vim in a consistent manner, they use a special state called
 =C-d=, =C-e=, =C-o=, =C-i=, =C-u=, =C-y= and =C-z=. All other keys work as
 intended by the underlying mode.
 
-Shadowed keys are moved according to the pattern: =a= → =A= → =C-a= → =C-A=
+Shadowed keys are moved according to the pattern: =a= → =A= → =C-a= → =C-A=.
+Some other keys are also moved: =SPC= -> ='=, =/= -> =\=, and =:= -> =|=.
+
 
 For example, if the mode binds a function to =n=, that is found under =C-n= in
 evilified state, since both =n= and =N= are reserved, but =C-n= is not. On the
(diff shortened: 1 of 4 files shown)


bcc32:
I think it would be great to take this opportunity to also document this behavior in `evil-evilified-state.el` :)

dankessler:
It seems it *is* mentioned in the [conventions doc](https://develop.spacemacs.org/doc/CONVENTIONS.html#evilified-buffers) but not in the [primary documentation](https://www.spacemacs.org/doc/DOCUMENTATION.html#evilified-modes), and I can rectify that.

As far as I know, "evilification" is unique to spacemacs and not a general part of evil; is this correct? I just want to make sure there isn't some "upstream" where I should also make this change. The closest I could find was this [repo](https://github.com/mohsenil85/evil-evilified-state?tab=readme-ov-file) which looks like it just spliced the code from spacemacs.

dankessler:
I've pushed another commit that updates the documentation in a few places. I didn't want to squash it, since the two commits are doing rather distinct things, but I can open a separate PR if that is preferred

bcc32:
Yeah, as far as I know, evil-evilified-state is unique to Spacemacs.

smile13241324 (approved):

PR #16345, opened 2024-03-31
[org] Enable SPC a o / key binding whenever helm-org-rifle not used

bcc32:
This enables the key binding when the completion layer is compleseus,
for example.


bcc32:
Actually, I think this should not be merged until #16346 is resolved.

bcc32:
Eh, that other PR is probably going to take a bit more time.  In the meantime, this is a clear and not too ugly improvement over the current situation.

smile13241324 (approved):

PR #16350, opened 2024-04-04
Discourage accidentally modifying packages by making such buffers read-only

bcc32:
Not sure if this would be controversial, but I do find myself frequently editing installed packages when I meant to be editing a git checkout of them.  I think it's usually not productive to edit installed packages because edits will be overwritten whenever the package is upgraded, so this should help mitigate this footgun.

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/config.el
+++ b/layers/+spacemacs/spacemacs-defaults/config.el
@@ -221,6 +221,16 @@ It runs `tabulated-list-revert-hook', then calls `tabulated-list-print'."
   (remove-hook 'kill-buffer-query-functions 'server-kill-buffer-query-function))
 (add-hook 'server-visit-hook 'server-remove-kill-buffer-hook)
 
+(defun spacemacs-defaults//discourage-editing-installed-packages ()
+  "Turn on `read-only-mode' in installed Elisp packages.
+
+This is meant as a small deterrent to accidentally editing
+installed package files."
+  (when (and buffer-file-name
+             (file-in-directory-p buffer-file-name package-user-dir))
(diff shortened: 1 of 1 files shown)


bcc32:
Hmm, I don't really understand why the CI build is failing.  I don't see any errors like this when installing packages by hand.

smile13241324:
hmm this seems to kill package installation on base

allentiak:
@bcc32
It seems there have been some changes to the elisp tests workflow file.
(See https://github.com/syl20bnr/spacemacs/commit/ffcfff353c137398d2bf3d29198dabfdef7a2bf6.)

Have you tried rebasing your changes on current `develop`?

smile13241324:
This seems to make autoloaded files read only.

```
An error occurred while installing evil (error: (error goto-chg-pkg.el:0:0: error: buffer-read-only: ( *autoload-file*)))
```

bcc32:
Aha, it seems to work fine for me because `package-generate-autoloads` has changed in Emacs 29 from calling `update-directory-autoloads` (which is now obsolete) to `loaddefs-generate`.  Only the former has trouble due to applying the directory-local `buffer-read-only=t` value.  While we probably can't require Emacs 29 as the minimum version for Spacemacs just yet, could we at least add CI for it, and maybe we can put off this PR until we don't support Emacs <29 anymore?

bcc32:
Ah, Emacs 29.3 was already added to the CI, thanks!

bcc32:
Hmm, I'm not 100% sure if there's a good way to do this that is robustly going to apply when the user is visiting a file, but won't break things that actually need to edit/recompile/generate such files.

PR #16355, opened 2024-04-04
Small improvements to core-load-paths.el

bcc32:


smile13241324 (approved):

sunlin7:
Hi @bcc32 
Why change the `.cache` folder from ~/.emacs.d/ to the spacemacs installation folder? 
https://github.com/syl20bnr/spacemacs/pull/16355/files#diff-2ed2a13c6fbd6262b08ee51062ff34063a31838546235aea0a1dceebd81a1a3aL98

bcc32:
Most of the time `~/.emacs.d/` is in fact the spacemacs installation directory.  But, if a user installs spacemacs to a different directory, Spacemacs's state should live there.  It's also consistent with all the other `spacemacs-*-` directory definitions in that file.  Is there a case where you think this is the wrong behavior?

sunlin7:
Hi @bcc32 
I deployed the Spacemacs in a readonly (shared) path, and the .cache should be a per user settings.

bcc32:
Aha, that's a great point.  Submitted #16409 to revert that particular change and add an explanatory comment.  Thanks for pointing this out, @sunlin7 .

sunlin7:
Thank you more !

Issue #12946, opened 2019-11-15
nyan-mode no longer works with doom-modeline


qizwiz:
#### Description :octocat:
`nyan-mode` no longer works with `doom-modeline`

#### Reproduction guide :beetle:
- Start Emacs
- Set `dotspacemacs-modeline-theme` to `doom`
- add `(colors :variables colors-enable-nyan-cat-progress-bar t)` to `dotspacemacs-configuration-layers`
- change buffers
- look at the *Messages* buffer

*Observed behaviour:* :eyes: :broken_heart:
- You'll see warnings about doom-modeline-segments and nyan-minimum-window-width
- no rainbow kitties :cry: 

*Expected behaviour:* :heart: :smile:
no warnings, rainbow kitties

#### System Info :computer:
- OS: gnu/linux
- Emacs: 27.0.50
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 22d200fba)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(systemd csv auto-completion
         (colors :variables colors-enable-nyan-cat-progress-bar t)
         command-log common-lisp csharp dap
         (dash :variables dash-docs-docset-newpath "~/.local/share/Zeal/Zeal/docsets" zeal-at-point-zeal-version "0.6.1")
         debug docker elm emacs-lisp erc erlang evernote geolocation git github gtags
         (haskell :variables haskell-completion-backend 'ghci haskell-process-type 'stack-ghci)
         helm html ibuffer imenu-list javascript lsp markdown multiple-cursors org pandoc parinfer
         (restclient :variables restclient-use-org t)
         (ruby :variables ruby-enable-enh-ruby-mode nil ruby-version-manager 'rvm ruby-test-runner 'minitest)
         ruby-on-rails search-engine
         (shell :variables shell-default-height 30 shell-default-position 'bottom)
         spacemacs-editing-visual
         (spacemacs-layouts :variables spacemacs-layouts-directory "~/.dotfiles/spacemacs-layouts/" layouts-enable-autosave t layouts-autosave-delay 1800)
         spacemacs-visual spotify sql syntax-checking syntax-checking templates themes-megapack
         (treemacs :variables treemacs-use-follow-mode t treemacs-use-filewatch-mode t treemacs-use-collapsed-directories 3)
         twitter version-control
[...]

PR #16361, opened 2024-04-05
Color: Removal of nyan-mode local copy

tonyfloatersu:
Hi there,

I opened up this PR as I encountered the same issue in #12946, that
nyan-mode won't show up under doom-modeline setting.  The solution
here seems to be taking advantage of the nyan-mode on MELPA, which
is essentially the one-line change in colors/packages.el.  If we go down
this way, then we can take away the local copy of nyan-mode altogether.

I also checked around and removed some text mentioning nyan-mode,
but I am unsure if we are allowed to do so (for example in COPYRIGHT).
I would refer the decision up to the code owner.

Thanks in advance for reviewing bandwidth.

Hang


smile13241324:
I am not sure that's that easy, I dimly remember that there is an open pr with improvements to the nyan cat animation from sylable which never has been merged.

I think we need to either get this merged or do a fork ourselves.

Anyway I am going to merge it for now for when it stopped working entirely for one theme its more important than the improved graphics but we should still try to get these back in time

smile13241324 (approved):

tonyfloatersu:
Oh thanks for the context of nyan cat animation improvement, I was totally unaware of that... but thanks for taking a look and thanks for merging

PR #16364, opened 2024-04-09
fix inferior R mode

gdkrmr:
R and related modes (R markdown, quarto mode, ...) were broken because spacemacs tried to assign a keybinding to `inferior-ess-r-mode-map`, which doesn't exist. I have done a quick fix by removing the assignment. In the R repl you cannot type `<-` by pressing `M--` any more, in a `.R` file it still works.

### DIFF
--- a/layers/+lang/ess/funcs.el
+++ b/layers/+lang/ess/funcs.el
@@ -87,8 +87,6 @@
 (defun spacemacs/ess-bind-keys-for-inferior ()
   (define-key inferior-ess-mode-map (kbd "C-j") #'comint-next-input)
   (define-key inferior-ess-mode-map (kbd "C-k") #'comint-previous-input)
-  (when ess-assign-key
-    (define-key inferior-ess-r-mode-map ess-assign-key #'ess-insert-assign))
 
   (dolist (mode '(inferior-ess-mode inferior-ess-r-mode))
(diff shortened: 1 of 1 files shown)


dankessler:
Thanks for this quick fix! I came across the same bug and before I noticed your PR, did a deep dive myself that led to a separate PR (#16366). 

In brief, the issue is not that `inferior-ess-r-mode-map` doesn't exist in general (it is defined in `ess-r-mode.el`), the problem is that it doesn't exist *when* `spacemacs/ess-bind-keys-for-inferior` is called.

Currently, `spacemacs/ess-bind-keys-for-inferior` is called from within an `:init` block for `ess-site` [here](https://github.com/syl20bnr/spacemacs/blob/a58a7d79b3713bcf693bb61d9ba83d650a6aba86/layers/%2Blang/ess/packages.el#L104-L105):

```elisp
    (with-eval-after-load 'ess-inf
      (spacemacs/ess-bind-keys-for-inferior))
```

However, `ess-site` is (implicitly) deferred, so it does not get loaded at startup. At some point, when the user tries to edit an `R` file (or does most things `ess`-related), the autoload system will cause `ess-inf.el` to get loaded, which will fire off this code, but this can happen *before* `ess-r-mode` has been loaded.

gdkrmr:
Nice! Thanks for taking the time and doing it right! Your PR should be merged.

PR #16367, opened 2024-04-11
[spacemacs-editing-visual] Disable vhl occur extension in Emacs 28+

bcc32:
This extension causes an error when navigating to occur matches
starting in Emacs 29, and already in Emacs 28, it became unnecessary
since occur does its own highlighting of matches when you visit them.

See k-talo/volatile-highlights.el#26.

### DIFF
--- a/layers/+spacemacs/spacemacs-editing-visual/packages.el
+++ b/layers/+spacemacs/spacemacs-editing-visual/packages.el
@@ -163,6 +163,12 @@
       :mode volatile-highlights-mode
       :documentation "Display visual feedback for some operations."
       :evil-leader "thv")
+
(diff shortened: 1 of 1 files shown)


pataquets:
vhl seems quite abandoned. [Protesilaos' Pulsar](https://github.com/protesilaos/pulsar) seems a more modern, well-maintained alternative used elsewhere (eg. [goggles.el](https://github.com/minad/goggles)).
RFC to maintainers: wdyt about switching?

smile13241324 (approved):

smile13241324:
Possible but I am unsure if its worth the effort as emacs 28 and later do not need it at all

PR #16384, opened 2024-05-02
Function "spacemacs/delete-current-buffer-file" support prefix argument

sunlin7:
Hi,

The `SPC f D` (`spacemacs/delete-current-buffer-file`) will always requests confirming by type y-or-n.
This change will enhance user experience with key bindings:
1. `SPC f d` -- delete current buffer and file, with confirming required;
2. `SPC f D` -- delete current buffer and file, assume yes (without  confirming required).
3. `C-u SPC f d` -- delete current buffer and file, assume yes (without  confirming required).

Please help review and approve this enhancement. Thanks.

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -558,16 +558,18 @@ FILENAME is deleted using `spacemacs/delete-file' function.."
   (funcall-interactively #'spacemacs/delete-file filename t))
 
 ;; from magnars
-(defun spacemacs/delete-current-buffer-file ()
-  "Removes file connected to current buffer and kills buffer."
-  (interactive)
+(defun spacemacs/delete-current-buffer-file (&optional arg)
+  "Removes file connected to current buffer and kills buffer.
+If ARG is not nil, assume yes for default."
+  (interactive "P")
   (let ((filename (buffer-file-name))
         (buffer (current-buffer))
         (name (buffer-name)))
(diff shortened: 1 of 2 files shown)


smile13241324 (approved):

smile13241324:
I think you still need to extend the bindings in Documentantion.org.

sunlin7:
Yes, totally!  I'll update the document and make a new pull request. Thank you !

pataquets:
@smile13241324 @sunlin7 
I have not pulled `develop` myself yet, but I'm afraid this might cause a quite annoying UX issue to users:
Wouldn't this change cause users using previously learnt binding to suddenly delete without confirmation? If true, this would be a blatant UX no-no, IMO.
See also: [Principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)

If proposed bindings are desirable, a more progressive rollout would prevent users to suffer it. I'd rather push initially just the binding change from `SPC f D` to `SPC f d`, leaving `SPC f D` for later, with just a "moved" message for some time to allow users to adjust.
At a later time, we can rollout the `SPC f D` binding as intended here. The latter can even just be skipped completely if extra safety is desired for users, and left to them.

sunlin7:
Hi @pataquets 
Aggree, and I'm going to enhance it, just give me minutes, will create a change to let uesr customer the behavrior.

sunlin7:
Hi @pataquets 
Please help review the PR: https://github.com/syl20bnr/spacemacs/pull/16403, thanks.

PR #16389, opened 2024-05-06
Bindings to restart python shell

rommeswi:
Thank you for the excellent python layer.

I noticed a minor issue: the bindings do not include a way to restart the shell conveniently.
It would be natural to include these in the `SPC m s` bindings.
E.g., `SPC m s` includes commands to remove unused imports or start a shell.

I would also suggest to replace the uninformative 'prefix' as the description for `SPC m s` by 'python-shell' or 'shell' but did not include this in the PR.


smile13241324 (changes requested):
I think you also need to add the new bindings to the layer readme.

smile13241324 (approved):

PR #16391, opened 2024-05-07
Add support for inf-ruby-mode ruby-quit

x3qt:
Hello!

I would like to add keybinding support for https://github.com/nonsequitur/inf-ruby/pull/185

Please let me know if any additional changes are required.

Thanks!


smile13241324 (changes requested):
I think you should add the new binding to the layer readme as well, the rest looks ok.

x3qt:
@smile13241324 thanks for suggestion, I've updated layer readme as well!

smile13241324 (approved):

Issue #13863, opened 2020-08-20
Autocompletion for Go layer is broken for imported packages with latest version


p554157atch:
With latest version of Golang Gomodules becomes the default thing! sigh.
And GOCODE and many other outdated tools won't support this stuff.

The company-go package wast last updated 4 years ago:
https://github.com/emacsmirror/company-go/blob/master/company-go.el

It does not support:
https://github.com/golang/tools/blob/master/gopls/doc/user.md

Spacemacs should move to gopls or something better and remove support for old code that calls gocode that no longer works with any kind of imports.

See this issue:
https://github.com/nsf/gocode/issues/537

p554157atch:
GOCODE IS not maintained anymore:
https://develop.spacemacs.org/layers/+lang/go/README.html

Move to GOPLS or something better,
Remove GOCODE from documentation.

smile13241324:
@p554157atch there is no moving to gopls as spacemacs already has full LSP support aka also support for latest gopls. The reason why LSP is not the default for most layers is that it requires a separate setup step which not all users are willing to go through.

If the alternative backend is not longer maintained we will shift the default to LSP like we did with the scala layer.
If you want to speed things up please have a look at the other tools loaded in that layer and point out which of these are obsolete nowadays and I can adapt it.

Remember this is a communities project, there is no way we (maintainers) can keep track of outdated packages in the individual layers, this must be fixed or at least reported by people using this language.

smile13241324:
I have started changing the layer to be favouring gopls over go-mode by default. However I am unsure how much of the given tools are actually needed when LSP is activated.

Can someone more familiar with go have a look and post a list of packages which are not longer working in this layer?
Till I have some more information I will remove company-go but keep the rest of the packages.

smile13241324:
I am through, the layer documentation now clearly explains which backends are available and deprecates `go-mode`. The layer uses the standard way to select a backend, this is if `lsp-layer` is loaded `gopls` is used if not found `go-mode` is used otherwise.

I am closing this ticket for now. If more packages are found which must be replaced in that layer this issue should be reopened.

p554157atch:
Thank you!

Issue #11217, opened 2018-08-26
golang stop working after upgrade to 1.11


jvillasante:
I'm very new to go, but after the upgrade to go 1.11 it seems that gocode stopped working.

```
Company: backend (company-go :with company-yasnippet) error "GOCODE PANIC: Please check your code by "go build"" with args (candidates ur)
Company: An error occurred in auto-begin
```

It was working fine before the upgrade.
Any thoughts/

asenyshyn:
nsf/gocode is not maintained anymore use https://github.com/mdempsky/gocode

robbyoconnor:
If you're using modules -- good luck -- it's not supported yet.

andres-lowrie:
^^ Is it on the roadmap to support modules?

robbyoconnor:
@andres-lowrie -- When tooling is updated, sure -- currently neither [go-langserver](https://github.com/sourcegraph/go-langserver), which is used in for [lsp](https://github.com/emacs-lsp/lsp-go/), nor [gocode](https://github.com/mdempsky/gocode) support it. There is an [issue](https://github.com/mdempsky/gocode/issues/46) tracking this in the gocode repo.

ihaveadrame:
I think it's time to support the module. 
Gopls already supports modules but does not correctly recognize modules on spacemac，
There are compatibility issues that need to be resolved, and these problems no longer exist on vscode.

robbyoconnor:
It's relatively painless to switch the language server.

ihaveadrame:
I solved my problem, gopls is ok!!

mdempsky:
Hi. I'm not a Spacemacs user, but I work on Go.

Is Spacemacs enabling company-go / gocode by default? If so, can it be switched to use lsp-mode / gopls by default instead?

gopls is the Go project's officially supported tool for code completion. (Unlike mdempsky/gocode, which was best effort.)

robbyoconnor:
Everything can be overridden.

mdempsky:
I understand users can override the settings. My understanding though is Spacemacs is meant to provide users with good defaults. And if company-go / gocode isn't supported and isn't working, why is it the default instead of lsp-mode / gopls?

p554157atch:
The default behavior with spacemacs and go layer is, it keeps autocomplete mode disabled, and enables the company-mode, which then enabled company-go by default.
This is incorrect behavior.

[...]

PR #16392, opened 2024-05-09
[go] Disable use of company-go when gopls is used

rawenous:
The default backend for go-layer is lsp but company-go was still added to company-backends forcing it to search for gocode which should not be used since the user wants lsp.

This change only adds company-go to company-backends if go-mode is selected as the backend.

Note: go-mode is deprecated and gocode is not maintained anymore

### DIFF
--- a/layers/+lang/go/packages.el
+++ b/layers/+lang/go/packages.el
@@ -25,7 +25,8 @@
   '(
     company
     dap-mode
-    (company-go :requires company)
+    (company-go :requires company
+                :toggle (eq go-backend 'go-mode))
     counsel-gtags
     eldoc
     flycheck


rawenous:
This may not be the optimal way of solving this so please let me know if there is a more correct emacs/spacemacs way.

This seems to solve #11217 and adds the last cleanup needed for #13863

smile13241324 (changes requested):
I think you should rather put a condition on the company go package itself so that it is not installed at all, right now we are installing it but never use it.

A good example for this is the package flycheck-golangci-lint.

rawenous:
Thanks. I will take a look at it.

layers/+lang/go/packages.el
```diff
@@ -25,7 +25,8 @@
   '(
     company
     dap-mode
-    (company-go :requires company)
+    (company-go :requires company
```
smile13241324, line 28: I think you need to also cover the case when there is no go backend.

smile13241324 (changes requested):

layers/+lang/go/packages.el
rawenous: @smile13241324, I don't fully understand. Today go-backend is either set to _lsp_ or _go-mode_ so there is always a go-backend selected. I can add _nil_ as an option that will load neither lsp or company-go.

smile13241324: Sorry my mistake some layers don't set this var reliably and have some code for inferring if lsp is selected from the layer configuration if this is not the case here its ok.

smile13241324 (approved):

PR #16397, opened 2024-05-13
Fix `describe-mode` keybinding breaking helm

### LINKED ISSUE #16395: M-x broken in the magit status buffer (and potentially others too)
#### Description :octocat:
`M-x` broken in the magit-status buffer (or everywhere??)
`M-m` broken in the spacemacs buffer (or everywhere??)

this issue has been filed by copying the template text generated in emacs after reverting spacemacs to last commit I was at (a58a7d79b) before updating my `.emacs.d` spacemacs git repo today. The broken latest dev branch of spacemacs doesn't permit me to do `M-m h I` because of the same issue being reported here.

- Make sure you have commit ca8b99552 (latest develop branch) checked out in
- Start Emacs
- Navigate to any git project in dired
- Open magit-status (`M-m g s`)
- `M-x`

You can also try to file a bug report with `M-m h I` with the latest version of spacemacs i.e. ca8b99552 and see the same issue.

There's a warning that says
"Key sequence M-m h d m starts with non-prefix key M-m"

The Helm M-x buffer opens up and I can begin to type the name of the function I want to invoke, e.g. magit-branch-and-checkout

#### System Info :computer:
- OS: darwin
- Emacs: 29.1 (installed using https://github.com/railwaycat/homebrew-emacsmacport )
- Spacemacs: 0.999.0
- Spacemacs branch: nil (rev. ca8b99552)
- Graphic display: t
- Running in daemon: nil
- Distribution: spacemacs
- Editing style: emacs
- Completion: helm
- Layers:
```elisp
(docker emacs-lisp git
        (go :variables go-backend 'lsp)
        helm html markdown multiple-cursors org php puppet python react terraform treemacs yaml)
```
- System configuration features: ACL GNUTLS JSON LIBXML2 MODULES NOTIFY KQUEUE PDUMPER SQLITE3 THREADS TOOLKIT_SCROLL_BARS TREE_SITTER XWIDGETS ZLIB

#### Backtrace :paw_prints:
```
<<BACKTRACE IF RELEVANT>>
```

fnussbaum:
Fixes #16395

### DIFF
--- a/layers/+completion/helm/packages.el
+++ b/layers/+completion/helm/packages.el
@@ -132,7 +132,6 @@
     (spacemacs||set-helm-key "sgG"  spacemacs/helm-file-do-grep-region-or-symbol)
     ;; various key bindings
     (spacemacs||set-helm-key "fel" helm-locate-library)
-    (spacemacs||set-helm-key "hdm" describe-mode)
     (spacemacs||set-helm-key "hdx" spacemacs/describe-ex-command)
     (spacemacs||set-helm-key "swg" helm-google-suggest)
     (with-eval-after-load 'helm-files
--- a/layers/+completion/ivy/funcs.el
+++ b/layers/+completion/ivy/funcs.el
@@ -343,24 +343,6 @@ that directory."
   (ignore-errors
     (call-interactively 'counsel-up-directory)))
 
-(when (configuration-layer/package-used-p 'counsel)
-  (with-eval-after-load 'counsel
-    (defun spacemacs/describe-mode ()
-      "Dummy wrapper to prevent an key binding error from helm.
-
-By default the emacs leader is M-m, turns out that Helm does this:
-   (cl-dolist (k (where-is-internal 'describe-mode global-map))
-        (define-key map k 'helm-help))
-after doing this:
-   (define-key map (kbd \"M-m\") 'helm-toggle-all-marks)
-So when Helm is loaded we get the error:
-   Key sequence M-m h d m starts with non-prefix key M-m
-
-To prevent this error we just wrap `describe-mode' to defeat the
- Helm hack."
-      (interactive)
-      (call-interactively 'describe-mode))))
-
 (defun spacemacs//counsel-with-git-grep (func x)
   "This function should be kept in sync with `counsel-git-grep-action'.
 
--- a/layers/+completion/ivy/packages.el
+++ b/layers/+completion/ivy/packages.el
@@ -88,7 +88,6 @@
       "hda" 'counsel-apropos
       "hdf" 'counsel-describe-function
       "hdF" 'counsel-describe-face
-      "hdm" 'spacemacs/describe-mode
       "hdv" 'counsel-describe-variable
       "hdx" 'spacemacs/describe-ex-command
       "hi"  'counsel-info-lookup-symbol
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -131,6 +131,17 @@ If not in such a search box, fall back on `Custom-newline'."
 (defalias 'spacemacs/display-buffer-other-frame 'display-buffer-other-frame)
 (defalias 'spacemacs/find-file-and-replace-buffer 'find-alternate-file)
 
+;; By default the emacs leader is M-m, turns out that Helm does this:
+;;    (cl-dolist (k (where-is-internal 'describe-mode global-map))
+;;         (define-key map k 'helm-help))
+;; after doing this:
+;;    (define-key map (kbd \"M-m\") 'helm-toggle-all-marks)
+;; So when Helm is loaded we get the error:
+;;    Key sequence M-m h d m starts with non-prefix key M-m
+;; To prevent this error we just alias `describe-mode' to defeat the Helm hack.
+(when (configuration-layer/package-used-p 'helm)
+  (defalias 'spacemacs/describe-mode 'describe-mode))
+
 (defun spacemacs/indent-region-or-buffer ()
   "Indent a region if selected, otherwise the whole buffer."
   (interactive)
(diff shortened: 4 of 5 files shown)


smile13241324:
If you're ready please ping me directly I would like to have this fixed soon otherwise I have to rollback some of these changes I am afraid.

smile13241324:
Maybe @bcc32 also wants to have a quick look.

fnussbaum:
Thank you @smile13241324 ! The PR is ready for review now. I added a commit message and tested basic functionality with the helm, ivy and compleseus layers (also with helm added as an additional package).

Conditioning on the helm package being used is more of a personal preference to avoid creating unnecessary aliases, one could also just always alias.

donm:
I can confirm that this fixed the `Key sequence M-m h d m starts with non-prefix key M-m` error for me in a vanilla config of spacemacs, as well as with my personal config.

And now, not to overcomplicate things, but...

If it would be preferable to others, I think that it's possible to move this workaround to the Helm layer, by defining the `spacemacs/describe-mode` alias in `+completion/helm/funcs.el` and then setting up the keybinding in `+completion/helm/packages.el` files.  

However, while testing that option, I went a bit further down a rabbit hole.  I'll mention it in case it's helpful, but feel free to tell me that you all don't want to worry about this right now.

I noticed that spacemacs is _already_ redefining the builtin `describe-mode` with a custom version, on line 867 of `layers/+spacemacs/spacemacs-defaults/local/help-fns+/help-fns+.el.`   If we renamed that function to `spacemacs/describe-mode`, then this extra `defalias` wrapper wouldn't even be needed.  

But that makes me wonder why spacemacs's custom definition of `describe-mode` (and related functions, like `describe-face` and `describe-function`) weren't already using the `spacemacs/` prefix.  Maybe there's a good reason, and even if not, maybe we don't want to worry about it right now just for the sake of fixing the current bug.

fnussbaum:
> If it would be preferable to others, I think that it's possible to move this workaround to the Helm layer, by defining the `spacemacs/describe-mode` alias in `+completion/helm/funcs.el` and then setting up the keybinding in `+completion/helm/packages.el` files.

Does this also work when not using the helm layer, but only including helm as an additional package? I tried to cover all of the cases with the above, but would be happy to change something if it is preferable.

> But that makes me wonder why spacemacs's custom definition of `describe-mode` (and related functions, like `describe-face` and `describe-function`) weren't already using the `spacemacs/` prefix. Maybe there's a good reason, and even if not, maybe we don't want to worry about it right now just for the sake of fixing the current bug.

Interesting, I don't know either.

donm:
> Does this also work when not using the helm layer, but only including helm as an additional package?

Well, that is a very good point.  Probably not--it would have to be addressed somewhere in the base spacemacs config somewhere, like you did.  (Well, as long as the`"hdm"` keybinding is also set up in the  base spacemacs config.)

fnussbaum:
> > Does this also work when not using the helm layer, but only including helm as an additional package?
> 
> Well, that is a very good point. Probably not--it would have to be addressed somewhere in the base spacemacs config somewhere, like you did. (Well, as long as the "hdm" keybinding is also set up in the base spacemacs config.)

Alright, I think we should support this case, as it seems to be common enough to keep helm around for some functionality after switching to ivy or compleseus. Having the <kbd>hdm</kbd> binding set up in the defaults is also a good thing I think.

> I noticed that spacemacs is _already_ redefining the builtin `describe-mode` with a custom version, on line 867 of `layers/+spacemacs/spacemacs-defaults/local/help-fns+/help-fns+.el.` If we renamed that function to `spacemacs/describe-mode`, then this extra `defalias` wrapper wouldn't even be needed.

This seems to only be the case when using an emacs version <= 27: 
```elisp
    (help-fns+ :location local
               :toggle (not (fboundp 'describe-keymap))) ; built in emacs28+
```

donm:
Oh, good call.  I just spent a while reading about the history of `help-fns+.el` and what is and isn't included in vanilla emacs, and...it's a whole thing, evidently.  I didn't know before that the file wasn't unique to spacemacs.  But that explains why the functions defined in that file don't have the `spacemacs/` prefix.

I'm sorry for sidetracking this PR!  Thanks for your fix.

fnussbaum:
No worries, it is quite interesting :)

sunlin7:
Acctually the https://github.com/syl20bnr/spacemacs/pull/16398 is a more simple solution.

smile13241324 (approved):

bcc32:
Hi folks, sorry for the disruption this caused.  (Oh how I wish there was just a test suite I could run.)  I think #16398 would have been a cleaner solution but I don't have a strong opinion about this.  I have some PRs incoming to clean more of this up once I get some free time---will try to remember to test with all the completion layers folks use before I submit :)

fnussbaum:
> Hi folks, sorry for the disruption this caused. (Oh how I wish there was just a test suite I could run.) 

No worries, yes a test suite would be great! 

> Acctually the #16398 is a more simple solution.

> I think #16398 would have been a cleaner solution but I don't have a strong opinion about this. 

I do not see how #16398 would have solved the issue, the error also persists in my tests with just the ivy/helm layer changes. 

The error was not caused by redefining the same binding within Spacemacs, it was caused by the Helm hack of looking up the global keybindings of `describe-mode` and trying to use the same keys to bind `helm-help` in the helm map. The problem with this is that <kbd>M-m</kbd> is already bound to `helm-toggle-all-marks` in there.

sunlin7:
Hi @fnussbaum 
Yes, you're right.
The helm has special code to deal the `describe-mode`:
https://github.com/emacs-helm/helm/blob/6d23a65ca6bcb6c0ea6f21f3cf2e58f8570ef75b/helm-core.el#L479-L481
```
(defvar helm-map
;;...
    (define-key map (kbd "M-m")        #'helm-toggle-all-marks)
;;...
    ;; Use `describe-mode' key in `global-map'.
    (dolist (k (where-is-internal #'describe-mode global-map))
      (define-key map k #'helm-help))
;;...
```
If arranged the `describe-mode` with key "M-m ..." before helm, the issue will be triggered.
Thanks

sunlin7:
As [fnussbaum](https://github.com/fnussbaum) commented in #16401, also reported this issue to Helm:
https://github.com/syl20bnr/spacemacs/pull/16401#issuecomment-2111857243

PR #16401, opened 2024-05-14
Enhance the describe-mode and more document

sunlin7:
Hi,
This PR is try to make more simple code for `describe-mode`, and more document for it. 

_Originally posted by @sunlin7 in https://github.com/syl20bnr/spacemacs/issues/16397#issuecomment-2110729312_

The helm has special code to deal the `describe-mode`:
https://github.com/emacs-helm/helm/blob/6d23a65ca6bcb6c0ea6f21f3cf2e58f8570ef75b/helm-core.el#L479-L481
```
(defvar helm-map
;;...
    (define-key map (kbd "M-m")        #'helm-toggle-all-marks)
;;...
    ;; Use `describe-mode' key in `global-map'.
    (dolist (k (where-is-internal #'describe-mode global-map))
      (define-key map k #'helm-help))
;;...
```
If arranged the `describe-mode` with key "M-m ..." before helm, the issue will be triggered.
Thanks

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -139,8 +139,9 @@ If not in such a search box, fall back on `Custom-newline'."
 ;; So when Helm is loaded we get the error:
 ;;    Key sequence M-m h d m starts with non-prefix key M-m
 ;; To prevent this error we just alias `describe-mode' to defeat the Helm hack.
-(when (configuration-layer/package-used-p 'helm)
-  (defalias 'spacemacs/describe-mode 'describe-mode))
+(defalias 'spacemacs/describe-mode 'describe-mode
+  "To avoid Helm key binding issue for `describe-mode'.
+Refer Spacemacs #16397 for details.")
 
 (defun spacemacs/indent-region-or-buffer ()
   "Indent a region if selected, otherwise the whole buffer."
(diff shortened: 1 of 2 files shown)


smile13241324:
Is this still necessary? And can @fnussbaum have a quick look also given that this touches his recent fix?

sunlin7:
Hi @smile13241324 
It makes the code simple and more document for future readers.

fnussbaum:
I see no harm in applying this patch. That being said, I think the reason for the alias is sufficiently documented in the comment above it. I deliberately chose to turn the docstring from the workaround in the ivy layer into a comment because this makes `describe-key` <kbd>SPC hdm</kbd> immediately show the documentation  of `describe-mode` (while also clearly mentioning that it is aliased). I assumed people curious about the alias would look into the source or even the `vc-region-history`, it mostly does not matter to end users I think. But if a docstring is preferred, that would of course be fine too.

Removing the condition on Helm being used is also okay, as I said before, I just subjectively prefer not defining an unnecessary alias when Helm is not used.

@sunlin7 I believe the cleanest solution would involve fixing the underlying issue in Helm and completely removing the alias. So maybe it would make sense to report this upstream, but that is not a high priority for me.

sunlin7:
Thanks for the comment, very clear.This issue took me hours to figure out the root cause, adding a doc here may save developer's hours when who runs into here.

smile13241324 (approved):

PR #16416, opened 2024-05-27
[compleseus] Devide search functions into 2 intial input variants

kenkangxgwe:
Each consult search function now comes with two variants regarding initial input. They all initialize the input with region content if a region is active, but differ on whether an empty string or thing-at-point if a region is not active.

This change tries to mimic the search behaviors configured in helm layers. Thus the default empty variants are bound to lowercase keys, and the default thing-at-point variants are bound to uppercase keys.

### DIFF
--- a/layers/+completion/compleseus/funcs.el
+++ b/layers/+completion/compleseus/funcs.el
@@ -58,37 +58,50 @@ non-nil."
      consult--source-project-buffer
      consult--source-project-recent-file)))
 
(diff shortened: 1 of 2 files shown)


kassick:
Ha! I ended up doing the same in [in my config](https://github.com/kassick/dotfiles/commit/224ad4be00192290317bab007a8c0f042257316b)

smile13241324 (approved):

Issue #16441, opened 2024-06-08
helm-spacemacs-help-faq (M-m h f) fails with "if: Wrong type argument: stringp, helm"


GreenRecycleBin:
#### Description :octocat:
helm-spacemacs-help-faq (M-m h f) fails with "if: Wrong type argument: stringp, helm".

#### Reproduction guide :beetle:
- Start Emacs
- Execute helm-spacemacs-help-faq (M-m h f)

*Observed behaviour:* :eyes: :broken_heart:
"if: Wrong type argument: stringp, helm" is printed in the *Messages* buffer.

*Expected behaviour:* :heart: :smile:
The Spacemacs FAQ buffer is displayed.

#### System Info :computer:
- OS: darwin
- Emacs: 29.3
- Spacemacs: 0.999.0
- Spacemacs branch: develop (rev. 7f5722cd8)
- Graphic display: t
- Running in daemon: nil
- Distribution: spacemacs
- Editing style: emacs
- Completion: helm
- Layers:
```elisp
(auto-completion better-defaults
                 (clojure :variables clojure-enable-fancify-symbols t)
                 docker emacs-lisp git helm lsp markdown multiple-cursors
                 (org :variables org-enable-org-journal-support t org-journal-enable-agenda-integration t)
                 parinfer ruby
                 (shell :variables shell-default-height 30 shell-default-position 'bottom)
                 spell-checking syntax-checking version-control theming treemacs tree-sitter
                 (zig :variables zls-backend 'lsp))
```
- System configuration features: ACL GLIB GMP GNUTLS JPEG JSON LIBXML2 MODULES NOTIFY KQUEUE NS PDUMPER PNG RSVG SQLITE3 THREADS TIFF TOOLKIT_SCROLL_BARS TREE_SITTER ZLIB

#### Backtrace :paw_prints:
N/A.

PR #16442, opened 2024-06-08
Fix helm-spacemacs-help-faq

GreenRecycleBin:
In https://github.com/syl20bnr/spacemacs/pull/13136, helm-spacemacs-help-faq
supporting functions were rewritten because helm-org-get-candidates was removed
upstream (https://github.com/emacs-helm/helm-org/pull/10).

One of the supporting functions, helm-spacemacs-help//get-faq-headings-list,
doesn't work. As a result, helm-spacemacs-help-faq (M-m h f) fails with "if:
Wrong type argument: stringp, helm".

It's unclear if this bug has existed since the Spacemacs PR because
helm-org-build-sources has changed since the helm-org
PR (https://github.com/emacs-helm/helm-org/commits/master/helm-org.el).

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -1511,6 +1511,8 @@ Other:
   - Set =evil-undo-system= to =undo-tree= (thanks to duianto)
   - Checked that =evil-undo-system= exists before it's called (thanks to khjcph)
   - Fixed ~SPC w TAB~ golden-ratio resizing.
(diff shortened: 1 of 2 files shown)


GreenRecycleBin:
This will fix https://github.com/syl20bnr/spacemacs/issues/16441 :)

smile13241324 (approved):

PR #16447, opened 2024-06-16
Small tweaks to Emacs Lisp format-on-save behavior

bcc32:


smile13241324 (approved):

smile13241324:
Cool, thanks for improving this function, this will make reviewing so much easier 😁

bcc32:
Thanks for adding the hook in the first place.  Should make formatting more consistent over time :)

PR #16449, opened 2024-06-17
fix: clashing narrow key for  consult--source-project-buffer and consult--source-modified-buffers

kassick:


kassick:
Kind-of-related:  The Modified Buffers source also filters by perspective -- Should it be "Perspective Modified Buffers" then?

https://github.com/syl20bnr/spacemacs/pull/16449/files#diff-e8c256fb9889aa6c9e5170d346961da3f0997e3a3f48530942c6ce5c80192522R29 

image

fnussbaum:
I think using the hidden project sources is great! 

For the narrowing keys, maybe it makes sense to stay as close as possible to `consult-buffer` which is bound to <kbd>SPC b B</kbd>? For example <kbd>p</kbd> could show both project buffers and project files. I think this is the default behaviour when using the hidden sources? 

In that case I guess we could keep <kbd>b</kbd> as the narrowing key for perspective buffers, which should already not clash with project buffers anymore, because the hidden source has a different narrowing key. Then we should also use <kbd>*</kbd> instead of <kbd>M</kbd> for modified buffers.


> Should it be "Perspective Modified Buffers" then?

I think this would be good. Maybe layout instead of perspective for both, but that does not really matter to me.

kassick:
> For the narrowing keys, maybe it makes sense to stay as close as possible to consult-buffer which is bound to SPC b B? 

I agree!

> For example p could show both project buffers and project files. I think this is the default behaviour when using the hidden sources?

Indeed it is -- consult groups sources that have the same narrowing key

> In that case I guess we could keep b as the narrowing key for perspective buffers, which should already not clash with project buffers anymore, because the hidden source has a different narrowing key. Then we should also use * instead of M for modified buffers.

As of a76c9f4e8952dc174f51163c0a6bd6977dc4f072 , it looks like this:
image

smile13241324 (approved):

smile13241324:
Looks good 😊

PR #16450, opened 2024-06-18
[completion] Tweak `*-spacemacs-help//layer-action-open-file`

fnussbaum:
The default layer action accessed through pressing <kbd>RET</kbd> from <kbd>SPC h l</kbd> currently insists on creating a README.org file if none exists for the layer which might be the case for private layers. Also, if one uses one of the actions to directly open one of the layer files, for example keybindings.el, and the file does not exist, only a message is displayed.

This commit changes the corresponding functions for Compleseus, Ivy and Helm to fall back to opening Dired at the layer directory if the file does not exist. When requesting to edit the README, it is still created if necessary. Note that the functions are exactly the same for Ivy and Compleseus, and the one for Helm only differs by additionally handling a prefix argument.

Here is a before/after comparison of the behaviour given the conditions (ignoring the special case of giving a prefix argument in Helm):

| is org file | edit | file exists | before                      | after                   |
|-------------|------|-------------|-----------------------------|-------------------------|
| nil         | nil  | nil         | **message**                 | **message+dired**      |
| nil         | nil  | t           | find-file                   | find-file               |
| nil         | t    | nil         | **message**                 | **message+dired**       |
| nil         | t    | t           | find-file                   | find-file               |
| t           | nil  | nil         | **spacemacs/view-org-file** | **message+dired**       |
| t           | nil  | t           | spacemacs/view-org-file     | spacemacs/view-org-file |
| t           | t    | nil         | find-file                   | find-file               |
| t           | t    | t           | find-file                   | find-file               |

### DIFF
--- a/layers/+completion/compleseus/local/compleseus-spacemacs-help/compleseus-spacemacs-help.el
+++ b/layers/+completion/compleseus/local/compleseus-spacemacs-help/compleseus-spacemacs-help.el
@@ -21,7 +21,7 @@
 
 ;;; Commentary:
 ;; This package adds a convenient way to discover Spacemacs configuration
-;; layers thanks to ivy.
+;; layers.
 
 ;;; Code:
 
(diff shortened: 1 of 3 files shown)


smile13241324 (approved):

smile13241324:
Nice addition to existing functionality, I like it 👍

PR #16452, opened 2024-06-18
[defaults] Tweak buffer and window transient states

fnussbaum:
Bind <kbd>x</kbd> to `kill-buffer-and-window` and make the buffer ts hint toggleable.


smile13241324 (approved):

smile13241324:
Looks nice 🙂

PR #16453, opened 2024-06-18
[defaults] Add leader key binding for `other-window-prefix`

fnussbaum:
This is handy enough to deserve a more prominent binding than the default <kbd>C-x 4 4</kbd>. 

I'm not sure whether adding the top level leader key binding <kbd>SPC O</kbd> like that is controversial, but this command is a lot more convenient on a short binding, because it is always composed with another command. I use it a lot, for example as <kbd>SPC O g d</kbd> to jump to definition in another window, or <kbd>SPC O RET</kbd> to open a link in another window.

I have checked that <kbd>SPC O</kbd> is not used yet by any layer, and only the lowercase <kbd>SPC o</kbd> is reserved for user bindings.

### DIFF
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -2375,6 +2375,16 @@ Windows manipulation commands (start with ~w~):
 | ~SPC w {~              | shrink window vertically (enter transient state)                            |
 | ~SPC w }~              | enlarge window vertically (enter transient state)                           |
 
+You can prefix many commands that display a buffer with the handy
+=other-window-prefix= to use or create another window instead of reusing the
+current one:
+
+| Key binding | Description                                              |
+|-------------+----------------------------------------------------------|
+| ~SPC O~     | display the buffer of the next command in another window |
+
+Try for example ~SPC O g d~ to jump to definition in another window.
+
 Split the current window into multiple ones, deleting all others using the
 following commands:
 
--- a/layers/+spacemacs/spacemacs-defaults/keybindings.el
+++ b/layers/+spacemacs/spacemacs-defaults/keybindings.el
@@ -649,6 +649,7 @@ respond to this toggle."
     (golden-ratio)))
 
 (spacemacs/set-leader-keys
+  "O"   'other-window-prefix
   "w TAB"  'spacemacs/alternate-window
   "w1"  'spacemacs/window-split-single-column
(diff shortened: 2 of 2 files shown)


fnussbaum:
It turns out that <kbd>SPC O g d</kbd> is not the best example, as there is <kbd>g D</kbd> for that already. I still like the consistency of `other-window-prefix` for all kinds of commands that display some buffer.

kassick:
Hmmm this made me find a bug upstream ...

`purpose-mode` sets `display-buffer-overriding-action` to `(purpose--action-function . nil)` [here](https://github.com/bmag/emacs-purpose/blob/master/window-purpose.el#L299) -- something valid, as the documentation for clearly states that the car can be either a function or a list of functions.

On the other hand, `display-buffer-override-next-command` wants to push a new element to `display-buffer-overriding-action`'s `car` ([window.el](https://github.com/emacs-mirror/emacs/blob/master/lisp/window.el#L9308)). Now `display-buffer-overriding-action`'s car is now a cons cell and not a list.

Later, when it wants to `delq` an element from it it ([window.el](https://github.com/emacs-mirror/emacs/blob/master/lisp/window.el#L9292))... well, that cons cell is not a list.

This causes following window commands to fail with a `wrong-type-argument listp ...` message until you evaluate `(setq display-buffer-overriding-action (cons (list purpose--action-function) nil))` to stop the callback from failing.

Maybe we should have this snippet somewhere until that is fixed...

``` elisp

(advice-add 'display-buffer-override-next-command :before (lambda (&rest _)
                                                            (unless (listp (car display-buffer-overriding-action))
                                                              (setcar display-buffer-overriding-action (list (car display-buffer-overriding-action))))))

```

fnussbaum:
Oh nice thanks for catching that and fixing it promptly! Seems like your fix is going to be included in Emacs 30, so I agree that we should include that advice snippet somewhere in Spacemacs for Emacs < 30.1. I think that could be a separate PR @kassick

smile13241324:
@fnussbaum your pr looks pretty good, care to remove the draft flag?

fnussbaum:
Thank you, I will come back to this in a few weeks. I have implemented a more general kind of prefix command in my configuration that optionally allows to display buffers in other workspaces and layouts and would like to propose to upstream that instead, but this is going to take some more time.

PR #16456, opened 2024-06-18
[pdf] add binding for themed view mode

dankessler:
document new binding and also add documentation for midnight-mode (which has had a binding for a while, but no documentation)


smile13241324 (approved):

smile13241324:
Doc updates are always welcome :) thanks for that.

smile13241324 (approved):

PR #16462, opened 2024-06-22
[spacemacs-defaults] Clean up old linum-related code

bcc32:
Remove a bunch of commands and hacks originally used with linum, which
Spacemacs no longer uses (instead using display-line-numbers-mode).
While we're at it, rename the related functions to refer to line
numbers generically rather than linum.


smile13241324:
Looks good 😊

PR #16463, opened 2024-06-23
Check if python-shell-interpreter is bound before null checking

august-feng:
Hello! 

The `python-shell-interpreter` variable comes from the **python** package which is not eagerly loaded, so `pyvenv-workon` will fail in those circumstances.


smile13241324 (approved):

smile13241324:
Good catch!

PR #16465, opened 2024-06-24
[tree-sitter][ts-fold] fix global mode activation

nixmaniack:
New version of `ts-fold` breaks in modes where `tree-sitter` isn't supported. Based on
[this](https://github.com/emacs-tree-sitter/ts-fold/issues/121#issuecomment-2134224132) issue and comment, update global ts-fold activation.


smile13241324 (approved):

smile13241324:
Thanks! I have missed this one.

PR #16490, opened 2024-07-16
Show leader for llm-client (gptel) layer at init

rocha:
Declare the leader at `:init` time instead of `:config` so gptel shows up in the menus at startup, and not until the package has been manually loaded.


rocha:
@aam-at

aam-at:
@rocha Thanks!

smile13241324 (approved):

PR #16491, opened 2024-07-17
[org] Prevent SPC TAB from switching to `*Calendar*`

bcc32:
This buffer is temporarily displayed whenever `org-read-date' is
called, but org-mode users do not typically use it by itself, and it
gets in the way of switching to other org buffers.

### DIFF
--- a/core/core-funcs.el
+++ b/core/core-funcs.el
@@ -322,26 +322,33 @@ buffer."
 (define-obsolete-function-alias 'spacemacs/derived-mode-p 'provided-mode-derived-p "2024-06")
 
 (defun spacemacs/alternate-buffer (&optional window)
-  "Switch back and forth between current and last buffer in the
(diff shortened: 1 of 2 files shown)


fnussbaum:
Looks good, thanks! The changes just made me wonder if we should keep declaring magit buffers as useless by default, or rather leave this to the user to customize, see also 3cb9fea0ff71430858aec5abc43d5508be06b573. What do you think?

fnussbaum:
A very minor point: 
```elisp
(with-eval-after-load 'calendar
  (cl-pushnew (regexp-quote calendar-buffer) spacemacs-useless-buffers-regexp :test 'equal))
```
may be preferable, just to avoid adding redundant entries every time the configuration is reloaded.

smile13241324 (approved):

PR #16492, opened 2024-07-17
[org] Ensure install the lastest version org from elpa

### LINKED ISSUE #16464: built-in org package used. why?
I'm running:
- emacs 29.3
- spacemacs latest develop, SHA1: 554d5cd16ee75cd03c065eb9b1e37907d54a5ee4 (Fri Jun 14 20:46:13 2024)
- Linux

org layer is enabled.
org-version returns: 
```
org-version is a variable defined in ‘org.el’.
Its value is "9.6.15"

```

in `.emacs.d/layers/+emacs/org/README.org`, I can read:

```
Spacemacs uses the =org= version from the ELPA repository
instead of the one shipped with emacs. Then, any =org= related code should not
be loaded before =dotspacemacs/user-config=, otherwise both versions will be
loaded and will conflict.

```

package-archives:
```
package-archives is a variable defined in ‘package.el’.

Its value is
(("melpa" . "https://melpa.org/packages/")
 ("gnu" . "https://elpa.gnu.org/packages/")
 ("nongnu" . "https://elpa.nongnu.org/nongnu/"))

```

from https://elpa.gnu.org/packages/org.html:
```
Description
    Outline-based notes management and organizer
Latest
    [org-9.7.5.tar](https://elpa.gnu.org/packages/org-9.7.5.tar) ([.sig](https://elpa.gnu.org/packages/org-9.7.5.tar.sig)), 2024-Jun-22, 9.69 MiB
Maintainer
    Bastien Guerry <bzg@gnu.org>

```

I also don't see org package present `elpa/29.3/develop` directory.

What did I miss?

I'd like to have newer org version installed into my spacemacs (at least newest 9.6.X) and I don't understand packages update doesn't take the newest version.

meritamen:
Straightforwardly filter `org` to the list of packages to install, however seemingly no other packages are built-in but supposed to be install from elpa. Fix #16464.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -1741,7 +1741,7 @@ RNAME is the name symbol of another existing layer."
              not-inst-count)
      t)
     (spacemacs//redisplay)
-    (unless (package-installed-p pkg-name min-version)
+    (when (or (not (package-installed-p pkg-name min-version)) (eq pkg-name 'org))
       (condition-case-unless-debug err
           (cond
            ((or (null pkg) (eq 'elpa location))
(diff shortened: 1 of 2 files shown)


fnussbaum:
This seems to reinstall Org on every restart. Also I don't think we should always automatically update to the latest version, instead a package update should usually be explicitly initiated by the user via <kbd>SPC f e U</kbd>.

We could just bump the :min-version to 9.7, although I was hesitant to submit such a change, because the upgrade had not been a smooth process for me. In particular it required recompiling some Org-related packages, but with that I also encountered some problems where the built-in Org was used to compile instead of the newly installed one. Of course there certainly exists a way to do this correctly and seamlessly.

PR #16508, opened 2024-07-26
[compleseus] Mark some consult-* navigation commands as evil jumps

bcc32:
Analogous to the corresponding counsel-* commands in the ivy layer.

### DIFF
--- a/layers/+completion/compleseus/packages.el
+++ b/layers/+completion/compleseus/packages.el
@@ -420,31 +420,31 @@
       (define-key vertico-map (kbd "C-r") #'consult-history)))
 
   (use-package vertico-directory
-      :after vertico
-      :ensure nil
-      ;; More convenient directory navigation commands
-      :init (bind-key "C-h" 'vertico-directory-up vertico-map
-                      (spacemacs//support-hjkl-navigation-p))
-      ;; tidy shadowed file names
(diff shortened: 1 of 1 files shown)


fnussbaum:
One way to circumvent having to add the jump property to all kinds of jump commands is to advise `push-mark`, as well-behaved functions usually call this before jumping. Also, for me, setting a jump property for `consult-ripgrep` and related commands did not seem to work. But it looks more like a hack, so maybe it should not be upstreamed. I'm using the following:
```elisp
(define-advice push-mark (:before (&optional location nomsg activate) evil-set-jump)
  (cl-letf (((symbol-function 'push-mark) #'ignore))
    (evil-set-jump location)))
```
In addition to working seemingly universally, it has the advantage (for commands reading a location from the minibuffer) of not setting a jump point when the minibuffer is exited without jumping anywhere.

smile13241324 (approved):

bcc32:
Hmm, perhaps, but I agree that seems hacky and potentially confusing to debug.  Might be a reasonable suggestion for evil-mode itself, but I think adding the advice to Spacemacs would not be the best way about it.

FWIW, evil-collection also just sets `:jump` properties on various commands, so this is probably the "official" way to do it, AFAIK.

fnussbaum:
Makes sense, thanks! I am thinking about suggesting something along those lines for evil-mode.

PR #16548, opened 2024-09-02
[common-lisp] Add key bindings for who binds/sets globals

meritamen:


smile13241324 (changes requested):
I think you also need to add these bindings to the layer Readme. The rest is ok.

meritamen:
Done.

smile13241324 (approved):

PR #16561, opened 2024-09-09
[defaults] disable show-paren-mode

bb2020:


fnussbaum:
I think this is a good change, thanks. Some minor comments: 
- One could consider moving this into the use-package declaration of smartparens, since the spacemacs-defaults layer can be used without the spacemacs-editing layer, for example in the spacemacs-base distribution.
- While an argument of `0` seems to work to turn it off, I think `-1` would be a safer choice considering the documentation.

bb2020:
It conflicts with other highlighting packages too, not just `smartparens`.
Maybe put in the core layer?

bb2020:
It conflicts with other highlighting packages too, not just `smartparens`.

smile13241324 (approved):

PR #16571, opened 2024-09-12
[version-control] Allow setting version-control-diff-tool to nil

bcc32:
If the user doesn't want to even install either diff-hl or git-gutter,
for example.


bcc32:
FWIW I have had this variable set to nil in my config for a long time and it works smoothly as expected, so just allowing the value in the defcustom is a sufficient patch.

smile13241324 (approved):

PR #16572, opened 2024-09-12
[notmuch] Do not error if auto-completion layer not used

bcc32:


layers/+email/notmuch/packages.el
```diff
@@ -155,9 +155,10 @@
       (kbd "gr") 'notmuch-refresh-this-buffer
       (kbd "gR") 'notmuch-refresh-all-buffers
       (kbd "G") 'notmuch-search-last-thread
-      (kbd "M") 'compose-mail-other-frame)
+      (kbd "M") 'compose-mail-other-frame)))
 
-    (spacemacs|add-company-backends :backends notmuch-company :modes notmuch-message-mode)))
+(defun notmuch/post-init-company ()
```
smile13241324, line 161: I think you still need to add company to the list of packages to have the post init event fire.

smile13241324 (changes requested):

layers/+email/notmuch/packages.el
bcc32: Ah, yep, thanks for catching that.

smile13241324 (approved):

PR #16593, opened 2024-09-21
[compleseus] Disable automatic preview by default

fnussbaum:
Keep it only for theme selection. For almost all other commands automatic preview was already disabled by default. Instead of enabling it for a few other seemingly arbitrary commands, I think it makes more sense to leave this to users to customize. 

Moving the `consult-preview-key` customization to the init block allows overriding it in the dotfile without using `spacemacs|use-package-add-hook` or `with-eval-after-load`.

Also make `C-M-j` and `C-M-k` preview without delay in consult commands (they already do that for other commands).

### DIFF
--- a/layers/+completion/compleseus/packages.el
+++ b/layers/+completion/compleseus/packages.el
@@ -152,6 +152,9 @@
 
     ;; The :init configuration is always executed (Not lazy)
     :init
+    ;; disable automatic preview by default
+    (setq consult-preview-key '("M-." "C-SPC" "C-M-j" "C-M-k"))
+
     (define-key read-expression-map (kbd "C-r") #'consult-history)
     (spacemacs/set-leader-keys
       dotspacemacs-emacs-command-key 'execute-extended-command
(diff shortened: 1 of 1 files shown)


fnussbaum:
For example, automatic preview was already disabled by default for `spacemacs/compleseus-switch-to-buffer`, but was enabled for `consult-buffer`, which is inconsistent.

mijoharas:
Ahhh nice, I missed that. Thanks!

mijoharas:
I think this looks good. It is somewhat confusing how this is inconsistent. I do occasionally want to toggle it on for a short amount of time. (which is what I've added to my other PR).

One thing to note, I did wonder whether we should provide a `defcustom` for the default `consult-preview-key` to make setting that more discoverable for people. (I like it how it is so I didn't bother).

smile13241324 (approved):

fnussbaum:
> One thing to note, I did wonder whether we should provide a `defcustom` for the default `consult-preview-key` to make setting that more discoverable for people. (I like it how it is so I didn't bother).

Perhaps we could mention the setting in the readme, or explicitly refer users to the consult/embark/etc. documentation.

PR #16600, opened 2024-09-26
[spacemacs-modeline] Fix anzu initialization

bcc32:
spacemacs-modeline/post-init-anzu doesn't do anything because no layer
owns anzu (spacemacs-evil owns evil-anzu but not anzu).  Instead, we
can just configure this while we're configuring the
spacemacs-all-the-icons theme.

I wasn't able to test this because all-the-icons doesn't seem to work
with my fonts on my machine, but I think it should be a correct
change.  In any case, before this commit,
spacemacs-modeline/post-init-anzu doesn't seem to be called by anything.

### DIFF
--- a/layers/+spacemacs/spacemacs-modeline/packages.el
+++ b/layers/+spacemacs/spacemacs-modeline/packages.el
@@ -23,17 +23,12 @@
 
 (setq spacemacs-modeline-packages
       '(
-        anzu
         (doom-modeline :toggle (eq (spacemacs/get-mode-line-theme-name) 'doom))
         fancy-battery
         spaceline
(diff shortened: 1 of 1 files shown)


fnussbaum:
I have tested this briefly, and AFAICT `anzu-cons-mode-line-p` needs to stay `t`, otherwise the anzu segment does not show up at all. With that change it seems to work correctly, thanks.

bcc32:
Thanks for testing this!  Does the behavior actually change before vs. after this commit?  I don't know what it's expected to really look like (in any case, pushing your suggested change now).

fnussbaum:
Yes, it looks better. With the change, there was a little search icon displayed in the segment, and IIRC there were parentheses added and the font might also have been more in line with the rest of the mode line.

smile13241324 (approved):

PR #16609, opened 2024-10-04
Add function for toggling compleseus previews.

mijoharas:
Also add documentation on vertico keybindings.

So, there were a few ways to go about this. We already have a call to `consult-customize` here, but we don't customize any of the `compleseus/*` functions, so none of them have any preview enabled.

We could explicitly add them all to that, and then have something like [this wiki function](https://github.com/minad/consult/wiki#toggle-preview-during-active-completion-session) to allow toggling preview off, or we can have what I did below (which doesn't support toggling in a live buffer completion session). The advantage of my option is that it works for all consult functions.

There is one other issue. I wrote some docs for the existing vertico keybindings, but on my machine the `C-M-j` and `C-M-k` bindings don't actually work? I added them to the docs anyways. Is that just a me issue?

### DIFF
--- a/layers/+completion/compleseus/README.org
+++ b/layers/+completion/compleseus/README.org
@@ -10,6 +10,7 @@
     - [[#completion-engine][Completion engine]]
 - [[#key-bindings][Key bindings]]
   - [[#edit-consult-buffer][Edit consult buffer]]
+  - [[#vertico-key-bindings][Vertico key bindings]]
 
 * Description
 This layer implements completion provided by combining the following packages:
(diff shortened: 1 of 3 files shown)


fnussbaum:
>There is one other issue. I wrote some docs for the existing vertico keybindings, but on my machine the C-M-j and C-M-k bindings don't actually work? I added them to the docs anyways. Is that just a me issue?

This should be fixed in #16593.

mijoharas:
Ahh nice, missed that, thanks! (and yep! that looks like it'll fix it).

smile13241324 (approved):

fnussbaum:
I think there is a small error in `spacemacs/consult-toggle-preview`: The list in the else branch should be quoted, and the if condition should probably check whether `consult-preview-key` is eq to what was assigned in this else branch? (Additionally it might be nice to remember and restore the original value of `consult-preview-key`, in case it was customized by users. Then the function also would not have to be updated manually with the new `C-M-j` and `C-M-k` entries introduced in #16593.)

I have been experimenting a little bit with automatically restarting the current minibuffer command within the function (similar to what `embark-become` does), but I'm not sure yet whether it is possible to do this in a nice way.

mijoharas:
Sorry, missed your comment.
* you're compeletely right about the `setq` bug (I changed it a little from what's in my config, and was a bit sloppy, I just have `'any` for mine).
* Also, agreed, we should restore it to what it was before. I was a little bit lazy. We should probably `defcustom` the initial default pick as well (and/or the toggled settings?)

I see the first has been fixed, I'll see about fixing the next when I get a minute. (EDIT: nope, first one hasn't been fixed since we still compare with `'any` in the `eq`... so it's just very broken. :joy: )

PR #16614, opened 2024-10-09
Use rustic-cargo-test-run instead of rustic-cargo-test

hariseldon78:
The command rustic-cargo-test depends on the variable rustic-test-arguments which gets polluted when running the "current-test" variant. So there is no way to run all the tests after running a single one, without clearing that variable.
Using rustic-cargo-test-run allows to go back to running all tests instead.
We could also keep both, but a meaningful description would be needed for rustic-cargo-test because the behavior is "unexpected"


hariseldon78:
Sorry for the 2 typo commits, can they be removed by squashing?

bcc32:
Does the key binding need to be changed?  I feel like `t a` for test all is pretty intuitive and would prefer to keep that one, IMO.  (If it does get changed, the README.org should be updated to reflect the new key binding).

hariseldon78:
Ok, it's also pretty intuitive, like "Test All". Fix incoming.

smile13241324 (approved):

PR #16615, opened 2024-10-09
* core/core-configuration-layer.el: remove unnecessary call (package-initialize)

sunlin7:
Hi,
It's unnecessary to call the `package-initialize` since emacs-27 according to its official document:
https://github.com/emacs-mirror/emacs/blob/9ed82c26a36caf9a9a85779cbb3a58b7ccd3dffb/etc/NEWS.27#L230-L236

And also benefit shorter startup time after remove the lines.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -441,10 +441,7 @@ cache folder.")
           (configuration-layer/elpa-directory
            configuration-layer--elpa-root-directory))
     (setq package-archives (configuration-layer//resolve-package-archives
-                            configuration-layer-elpa-archives))
-    ;; optimization, no need to activate all the packages so early
-    (setq package-enable-at-startup nil)
-    (package-initialize 'noactivate)))
+                            configuration-layer-elpa-archives))))
 
 (autoload 'quelpa "quelpa")
 (autoload 'quelpa-checkout "quelpa")


bcc32:
FYI, with this change the theme selected in .spacemacs does not load properly at startup.  (At least, in my config where the theme is part of `dotspacemacs-additional-packages`).

sunlin7:
@bcc32  Could you please give more hints on your configuration? I'm going to reproduce the issue, want to know for additional packages and how to apply the external theme.  Thanks

bcc32:
Sure, here's the relevant settings, I believe:

```
   dotspacemacs-themes '(
                         base16-material
                         )

...

   dotspacemacs-additional-packages
   '(base16-theme)
``` 

I was able to reproduce the issue from a completely default config and only changing the above two variables.

It produces the following error at startup:

```
error: (file-missing Cannot open load file No such file or directory base16-theme)
```

Some of the colors from the theme are applied correctly but not all, it seems.

sunlin7:
I tried your pasted configuration and everything is okay. My local installed emacs 31.0. 
What's your emacs version? Thanks.

smile13241324:
I think we should test this with emacs 28, 29 and 31 our config need to work for all of these

sunlin7:
@smile13241324  Sure, absolutely need it works on emacs-28 +.
I had compiled the emacs-28 on Ubuntu 20.04, this patch works.
Tried the emacs-28 on Windows, this patch also works.
@bcc32 Could you trouble shooting why it happened on your local? Thanks

bcc32:
I am on Emacs 29.4.  I tried it on both NixOS and MacOS and both were broken, as in, issue this error message at startup:

```
error: (file-missing Cannot open load file No such file or directory base16-theme)
```

You can find my full config [here](https://github.com/bcc32/dotfiles/tree/de409039a0f5585549c133e295551e1c15b2767a/emacs/.spacemacs.d)

The reason this happens is because in `spacemacs/init`, `configuration-layer/initialize` is followed shortly by `spacemacs/load-default-theme`, and the latter calls `package-activate`.  `package-activate` fails if it is called before `package-initialize`.

Looking at the news entry you linked, you can see that `package-initialize` still actually needs to be called at some point if you set `package-enable-at-startup` to nil, which Spacemacs does in `early-init.el`:

> However, if your init file changes the values of 'package-load-list'
> or 'package-user-dir', or sets 'package-enable-at-startup' to nil then
> it won't work right without some adjustment:
> - You can move that code to the early init file (see above), so those
>   settings apply before Emacs tries to activate the packages.
> - You can use the new 'package-quickstart' so activation of packages
>   does not need to pay attention to 'package-load-list' or
>   'package-user-dir' any more.

bcc32:
I just noticed a second thing that breaks---packages installed through the system package managed (such as mu4e which can only be installed this way) are no longer activateable, either.

sunlin7:
@bcc32  The issue reproduced with Emacs-28.3, I'll look it closely now. Thanks.

sunlin7:
Close ticket for it only works on Emacs 31, but not compatible with Emacs-29.

bcc32:
Thanks for looking and sorry to be the bearer of bad news!

bcc32:
Thanks for looking and sorry to be the bearer of bad news!

sunlin7:
Hi @bcc32 
Could you change your configuration https://github.com/bcc32/dotfiles/blob/de409039a0f5585549c133e295551e1c15b2767a/emacs/.spacemacs.d/layers/bcc32/packages.el#L41
From 
```(setopt corfu-auto t)```
To 
```(custom-set-variables·'(corfu-auto·t))```
And give another try? Thanks

bcc32:
That does not make a difference.  It is still broken.

PR #16622, opened 2024-10-15
Fixed bug with toggle-preview and added config.

mijoharas:
Fix bug introduced in https://github.com/syl20bnr/spacemacs/pull/16609. See comment https://github.com/syl20bnr/spacemacs/pull/16609#issuecomment-2395429379 for explanation.

NOTE, at first I tried to do:
```elisp
      (setq compleseus--previous-preview-keys consult-preview-key
            consult-preview-key (consult-preview-key '(:debounce 0.3 any)))
```
within the function, but that didn't really seem to actually work? I then noticed that you can't even do:
```elisp
  (setq consult-preview-key '("M-." "C-SPC" "C-M-j" "C-M-k" :debounce 0.5 "<up>" "<down>" :debounce 0.3 any))
```
which is confusing as it's supported in `consult-customize`. Either way, afaict that's a bug in consult, but I ignored it since this simpler thing fixed my problem.

### DIFF
--- a/layers/+completion/compleseus/README.org
+++ b/layers/+completion/compleseus/README.org
@@ -8,6 +8,7 @@
 - [[#install][Install]]
   - [[#configuration][Configuration]]
     - [[#completion-engine][Completion engine]]
+    - [[#preview-keys][Preview keys]]
 - [[#key-bindings][Key bindings]]
   - [[#narrowing][Narrowing]]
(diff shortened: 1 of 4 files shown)


mijoharas:
(I raised an issue for the behaviour I mentioned above in consult [here](https://github.com/minad/consult/issues/1116)

smile13241324 (approved):

PR #16627, opened 2024-10-26
feat(LSP): Allow users to manage Company backends manually

niontrix:
Why?:
- Currently lsp-mode aggressively inserts the company-capf backend, which contains its completion-at-point-function, as the ultimate first item of the list of company-backends. If a language server does not provide all the functionality, that could otherwise be provided by another Company backend like company-files for example, lsp-mode will practically block any other Company backend inside company-backends from running, because it tells Company that it may be able to complete at point but in the end doesn't provide any completion candidates. This is not ideal and users may want to manage company-backends by themselves. Also using this variable within some language layers; we could also be able to provide a more focused list of Company backends for that mode that extends the functionality currently not provided by the language server.

- Introduce new lsp-layer variable lsp-manage-backends-manually, which can either be :all (to manage the backends for every major-mode manually), a list of major-modes (to manage backends for these specific modes manually) or nil to let lsp-mode manage the company-backends
- Modify lsp/init-lsp-mode to set lsp-completion-provider based on the value of lsp-manage-backends-manually


niontrix:
Ok, maybe this is not ready just yet. When `company-capf` is not the first item in the `company-backends` list the sorting of the candidates is now broken - at least when using it with `gopls`. I'll try to find the issue first or create an issue upstream.

niontrix:
To make `company-files` work in conjunction with `company-capf` one needs to set `company-backends` like this for example
```elisp
(setq-local company-backends '(:separate company-capf company-files company-yasnippet))
```

niontrix:
@smile13241324 could you please take a look and/or provide feedback?

smile13241324 (approved):

niontrix:
Thanks @smile13241324 ! :+1: 
Should I maybe add what I have written in the comments above to the documentation? If so could you point me to the right place?

smile13241324:
I think the best place would be a new section in the lsp layer readme

PR #16643, opened 2024-11-05
[unicode-fonts] Do not error in after-make-frame-functions

bcc32:
If `unicode-fonts-setup` errors (such as if the eieio cache is
corrupted), emacsclient behaves confusingly by opening both a
graphical frame and a terminal frame.

You can reproduce this scenario by corrupting
`~/.emacs.d/.cache/pcache/unicode-fonts` (such that is no longer valid
Lisp syntax) and then starting Emacs with `emacsclient -ca ''`.

Also remove an unnecessary require, since `unicode-fonts-setup` is
autoloaded.


smile13241324 (approved):

michaelpass:
- [x] - - <> . (Cool)[@]()

PR #16650, opened 2024-11-09
Fix failed on the default theme and not depending on package-activate

sunlin7:
Hi,

Spacemacs has issues on theme, 
1. Error on `default` theme:
 a) write the `dotspacemacs-themes` to `(default)`, then start Spacemacs with daemon mode: `emacs --daemon`
 b) start a graphic emacsclient: `emacsclient -c .`
Then error happens: `*ERROR*: Unable to find theme file for ‘default’`.
This patch will try load theme but not try enable it first, if the theme is exists then enable the theme; otherwise set the theme to `spacemacs--delayed-user-theme` to postpone the operation. 
2. For a `<non-exists>` theme in `dotspacemacs-themes`, then Spacemacs will switch theme from `default` -> `spacemacs-dark(fallback-theme)` -> `default` -> `spacemacs-dark(fallbak-theme)` during startup:
 a) write the `dotspacemacs-themes` to `(non-exists)`, then restart the Spacemacs.
 b) the Spacemacs will flash during the startup.
It caused by the `spacemacs/load-default-theme` will try load the `non-exists` theme then fallback to the `spacemacs-dark`, and  in `emacs-startup-hook` it will load the `spacemacs--delayed-user-theme` again, it still is not exists, then fallback to `spacemacs-dark` again. 
This patch won't try the fallback-theme for first attempt, only fallback at the last one.
3. Spacemacs theme function relates to the `package-activate` function, which requests the `package-initialize` will load all the `*-pkg.el` and `*-autoloads.el` for packages; cause a heavy loading; 
This change will disconnect the theme from `package-activate`, will let the `package-quickstartup` to work, refer: https://github.com/syl20bnr/spacemacs/pull/16651 .

### DIFF
--- a/core/core-spacemacs.el
+++ b/core/core-spacemacs.el
@@ -235,7 +235,7 @@ the final step of executing code in `emacs-startup-hook'.")
     (defvar load-hints '()))   ; *-autoloads.el after the feature was toggled.
 
   ;; theme
-  (spacemacs/load-default-theme spacemacs--fallback-theme 'disable)
+  (spacemacs/load-default-theme)
   ;; font
   (spacemacs|do-after-display-system-init
    ;; If you are thinking to remove this call to `message', think twice. You'll
--- a/core/core-themes-support.el
+++ b/core/core-themes-support.el
@@ -373,16 +373,22 @@ package name does not match theme name + `-theme' suffix.")
                  pkg-name)))
     dir))
 
-(defun spacemacs/load-default-theme (&optional fallback-theme disable)
+(defun spacemacs/load-default-theme ()
   "Load default theme.
-Default theme is the car of `dotspacemacs-themes'.
-If FALLBACK-THEME is non-nil it must be a package name which will be loaded if
-THEME cannot be applied."
+Default theme is the car of `dotspacemacs-themes'. If failed to load the
+default theme, setting the `spacemacs--delayed-user-theme' to postpond
+the action."
   ;; This function is called before all packages are necessarily activated, so
-  ;; just activate the necessary one now, if available.
-  (let ((default-theme (car dotspacemacs-themes)))
-    (spacemacs//activate-theme-package default-theme)
-    (spacemacs/load-theme default-theme fallback-theme disable)))
+  ;; if failed to load the theme we can try again after the packages activated.
+  (if-let* ((default-theme (car dotspacemacs-themes))
+            (theme-name (spacemacs//get-theme-name default-theme)))
+      (condition-case err
+          (spacemacs//load-theme-internal theme-name)
+        ('error (setq spacemacs--delayed-user-theme theme-name)))
+    (spacemacs-buffer/warning
+     (concat "Please check the `dotspacemacs-themes' in your dotfile\n"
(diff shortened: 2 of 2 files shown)


smile13241324 (approved):

bcc32:
I think this made the situation worse for themes installed through `dotspacemacs-additional-packages`.  They used to be applied immediately, but after this commit, Spacemacs starts with the `default` theme and uses that theme until configuration layers are done loading, and *only then* applies the user configured theme.

sunlin7:
Hi @bcc32  This change had fixed the bugs as I listed on the first message:
1. to support `default` theme, no theme  (`default` theme) will save the users who love the Emacs default theme; 
    Beside that, Windows user are suffered on the startup slow, no theme will save the startup time.
2. UI was flashed... <as I decribed on first message>
3. Early theme functions depends on package.el FULL activation, which is too heavy and block the feature "package-quickstart".

So this change is necessary. 

And, another benefits is, we're now have chance to get ride of the variable `spacemacs-theme-name-to-package` from Spacemacs core, that variable will growth again and again until no one can maintain it. 

For your concerns that the themes in addition package will apply later, we maybe can explorer some way to apply the additional theme early.

bcc32:
> * to support `default` theme, no theme  (`default` theme) will save the users who love the Emacs default theme;
>   Beside that, Windows user are suffered on the startup slow, no theme will save the startup time.

This issue could have been fixed very easily by itself, without a large and error-prone change to how packages are activated during startup.

> * UI was flashed...

UI flashing seems like a correct (or at least reasonable) behavior here, actually.  The user may reasonably decide to have startup code that installs the package to provide the theme, and if so, Spacemacs should make a second attempt to activate the theme.  I guess you could argue that it would be better to avoid attempting to enable a theme that would not be found, but this issue, which I would describe as "Spacemacs behaves less well than it could, when the user has a bad config" should not be fixed by breaking configs that **are supposed to work.**

> * Early theme functions depends on package.el FULL activation, which is too heavy and block the feature "package-quickstart".
>
> So this change is necessary.

#16651 did not need to remove the call to `package-initialize` for all users, only those who enabled `package-quickstart`.  These recent changes seem to have been far broader than necessary and caused a lot of annoying and difficult-to-debug side effects.

sunlin7:
Hi @bcc32  Please help review the enhancement MR #16685, it will compatible the previous behavior. Thanks

bcc32:
I don't think that MR helps at all.  Before the MR, it flashed "default -> selected theme", now it flashes "spacemacs-dark -> selected theme".  Neither is desirable, and the behavior before the recent package-activate changes was correct: no flashing at all, directly activate selected theme.

sunlin7:
@bcc32  Got your key point, please give me minutes, I'll research the solution to support the theme on early stage, and then create a PR to follow the previous behavior. Thanks.

sunlin7:
@bcc32  Please help review the #16690 , it should be able to resolve your concerns. Thanks.

bcc32:
Thanks, I will try to take a look tomorrow or Wednesday (have been busy over the holiday weekend)

sunlin7:
@bcc32  Take your time, thanks!

PR #16651, opened 2024-11-09
* core/core-configuration-layer.el: support package-quickstart

sunlin7:
A draft to support `package-quickstart`. 
Request the MR https://github.com/syl20bnr/spacemacs/pull/16650 to be merged first, then this MR will works.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -441,10 +441,7 @@ cache folder.")
           (configuration-layer/elpa-directory
            configuration-layer--elpa-root-directory))
     (setq package-archives (configuration-layer//resolve-package-archives
-                            configuration-layer-elpa-archives))
-    ;; optimization, no need to activate all the packages so early
-    (setq package-enable-at-startup nil)
-    (package-initialize 'noactivate)))
+                            configuration-layer-elpa-archives))))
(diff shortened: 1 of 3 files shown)


smile13241324:
The related MR has been merged @sunlin7

smile13241324:
The change looks good, though I would like to make another review when its state has been changed to ready.

sunlin7:
Thank you, I'm adding some document to show how it works, will update later.

sunlin7:
@smile13241324  I had finished the `package-quickstart` support.  The document is in another PR: #16661.
Please help review the changes, thank you!

smile13241324 (approved):

bcc32:
This PR causes the same problem as #16615 w.r.t. themes that the user installed through `dotspacemacs-additional-packages`.  If the theme package is installed, it will at least be applied correctly after all packages have been installed and after layer loading, etc.

But the code to apply the theme immediately at startup is now broken.  See `spacemacs/load-default-theme`.

If `package-quickstart` and `dotspacemacs-enable-package-quickstart` are nil, you still need to manually call `package-initialize` during startup, because `early-init.el` sets `package-enable-on-startup` to nil.

PR #16659, opened 2024-11-13
Migrate the if-let/when-let to the if-let*/when-let*

sunlin7:
The `if-let`/`when-let` is marked as obsoleted since emacs-26,
 https://lists.gnu.org/archive/html/emacs-devel/2018-03/msg00360.html.

And Emacs-31 produce a warning for if-let/when-let:
https://github.com/emacsmirror/emacs/commit/8903106bb783c2825233c149b6799960aacdea57

So Spacemacs should migrate the if-let/when-let to if-let*/when-let*. 

Please help review the changes. Thanks.


core/libs/quelpa.el
```diff
@@ -1911,7 +1911,7 @@ version."
                   (package-delete pkg-desc 'force))))
             all-pkgs))
     ;; Only packages with same version remained. Just pick the first one.
-    (when-let (all-pkgs (alist-get name package-alist))
+    (when-let* (all-pkgs (alist-get name package-alist))
```
bcc32, line 1914: This one is wrong, it used the old single-binding syntax and needs to be updated.

bcc32: (The rest of the changes LGTM)

bcc32 (commented):
Argh, I didn't realize I had to "submit review" to make the review comments show up.

bcc32:
@smile13241324, sorry about this, there is a small typo in the PR that I tried to alert about.

sunlin7:
Hi @bcc32 Thanks for comments, the changes on `core/libs/quelpa.el` actually force restored by the https://github.com/syl20bnr/spacemacs/commit/c3269f6a3b9ce8ee57f362c2b053b5381f8727e3, so it back to original one.

core/libs/quelpa.el
sunlin7: Restored by https://github.com/syl20bnr/spacemacs/commit/c3269f6a3b9ce8ee57f362c2b053b5381f8727e3

PR #16661, opened 2024-11-13
* doc/FAQ.org: document for load-hints and package-quickstartup

sunlin7:
The document for `load-hints` and `package-quickstartup`, 
1. explained how to use the `load-hints` to speedup the Spacemacs on Windows;
2. explained how to use the `package-quickstart` to speedup the Spacemacs on Windows.

### DIFF
--- a/doc/FAQ.org
+++ b/doc/FAQ.org
@@ -52,6 +52,8 @@
   - [[#why-is-there-no-spacemacs-logo-in-the-startup-buffer][Why is there no Spacemacs logo in the startup buffer?]]
   - [[#why-are-all-packages-unavailable][Why are all packages unavailable?]]
   - [[#the-powerline-isnt-shown-correctly-when-spacemacs-is-used-within-putty][The powerline isn't shown correctly when Spacemacs is used within =PuTTY=]]
+  - [[#speedup-the-spacemacs-with-load-hints-feature][Speedup the Spacemacs with load-hints feature]]
+  - [[#speedup-the-spacemacs-with-package-quickstart][Speedup the Spacemacs with package-quickstart]]
 
 * Common
 ** Which version of Spacemacs am I running?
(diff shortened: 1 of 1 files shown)


doc/FAQ.org
```diff
@@ -609,3 +611,45 @@ directory as Emacs. See [[https://www.gnu.org/software/emacs/manual/html_mono/em
...
+seperately, while a Spacemacs-base distribution maybe installes more than 100+
+packages, and the ~load-path~ will has more than 100+ entries with packages
+paths on the head, emacs builtin paths on the tail. Then a simple
+src_elisp{(require 'org)} will trigger the emacs goes through the ~load-path~
+from the head to the tail to seach =org.dll=, =org.dll.gz=, =org.elc=,
+=org.elc.gz=, =org.el=, =org.el.gz=, that will be more 600 (100*6) attemptions
+to open the desired files, but Windows system slow on the open-file API
+(Linux/BSD are mutch faster).
+
+The petential solutions are:
```
SamHasler: Spelling: potential

SamHasler: "possible" might be a better word to use

smile13241324 (approved):

bcc32:
It should be noted that load-hints can cause correctness issues and lead to confusing errors.  I noticed that it seems to cause mixed version of files to be loaded for org-mode (some from the built-in version, some from the version installed by ELPA).  I think the docs should mention that it is still experimental and may cause issues like that (especially because it messes with load-path in surprising ways---during advice on require!)

sunlin7:
Yes, absolutely. I'm writing some document to highlight this feature is should be enabled for the users who suffered by the startup performance on Windows platform, and will push a new PR. Thanks

sunlin7:
Hi @bcc32  I create a draft for more document for `load-hints` feature on Windows. #16679

PR #16685, opened 2024-11-26
Guess fallback-theme for users

### LINKED ISSUE #16683: After commit 5ff44b Spacemacs shows light theme before selected theme is loaded
Hi folks!

I updated Spacemacs today and am using `doom-one` as my default theme. 
Since the recent commit [5ff44b](https://github.com/syl20bnr/spacemacs/commit/5ff44b284f8b251d45e7c90ddc020c3a23f6a47b) Spacemacs now starts up with the light theme until the package load is done and only after that switches to the configured theme.
The behavior is the same for all themes except when using `spacemacs-dark`.
I know it is a minor issue, but it is a regression from the previous state.
How can we fix this?

KR, niontrix

sunlin7:
To compatible with previous fallback theme behavior, applying the fallback-theme on startup,
and guess the fallback theme by name with "light" or "dark" to avoid flashing.
Fix #16683

### DIFF
--- a/core/core-themes-support.el
+++ b/core/core-themes-support.el
@@ -24,7 +24,7 @@
 (defconst emacs-built-in-themes (cons 'default (custom-available-themes))
   "List of emacs built-in themes")
 
-(defvar spacemacs--fallback-theme 'spacemacs-dark
+(defvar spacemacs--fallback-theme nil
   "Fallback theme if user theme cannot be applied.")
 
 (defvar spacemacs--delayed-user-theme nil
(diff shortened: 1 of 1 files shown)


niontrix:
Hi @sunlin7!

I checked your PR with `doom-one` and `doom-acario-light` and it works like it should.
Props for your quix fix! :+1:

sunlin7:
@niontrix Thank you for verifying both dark/light cases !

Look forward @smile13241324  to help approving this fix.

smile13241324 (approved):

smile13241324 (approved):

Issue #15499, opened 2022-05-07
[Feature request] Customize savehist, recentf, places save files locations


pataquets:
I was trying to customize some file location variables to be in my `~/.spacemacs.d/`directory.
Some examples:
* `recentf-file-name`
* `save-hist-file`
* `save-place-file`

Doing it on `user-init` ended up with the values being overwritten. And doing it on `user-config`, although somewhat worked, also created files on my `.emacs.d` dir anyway, adding unnecessary cruft.

I've found the corresponding `init-*` functions in the `spacemacs-defaults` layer, which have the directory `spacemacs-cache-directory` hard-coded, so it looks like to me that it is not possible to do this cleanly. No matter how you approach it, you'll be always working around Spacemacs' logic, which might change in the future.
(As a side note, I wonder if these files belong to the cache dir, but that's another story and it may be just my usage, which is keeping them synchronized between PCs via Dropbox).

If the feature is acceptable and someone is interested into picking this up, what would be the way to expose this?
Adding some `dotspacemacs-*` variables to be customizable on `dotspacemacs/init` would be  enough?
What would be the recommended naming?

pataquets:
Still interested

pataquets:
Still interested

Issue #12127, opened 2019-04-05
spacemacs-cache-directory is ignored in dotspacemacs/*-init


zabbal:
#### Description :octocat:
spacemacs-cache-directory variable is ignored when set in dotspacemacs/user-init or dotspacemacs/init
After closing Spacemacs the .cacheis created in the old location with auto-save subdirectory and spacemacs-buffer.el files
If that's the expected behavior and it should be set someplace else than it would be great to get this documented.

#### Reproduction guide :beetle:
- Set spacemacs-cache-directory in dotspacemacs/user-init to non-default value, for example "~/.cache/emacs/"
- Close Spacemacs

*Observed behaviour:* :eyes: :broken_heart:
.cache is re-created at old location

*Expected behaviour:* :heart: :smile:
The value of spacemacs-cache-directory should be used.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 25.2.2
- Spacemacs: 0.200.13
- Spacemacs branch: master (rev. c7a103a77)
- Graphic display: t
- Distribution: spacemacs
- Editing style: emacs
- Completion: helm
- Layers:
```elisp
(helm auto-completion better-defaults emacs-lisp git github markdown jabber semantic ibuffer mu4e emoji c-c   csv emacs-lisp ess go graphviz haskell scala shell-scripts org themes-megapack command-log finance pandoc pdf-tools restclient systemd search-engine
      (shell :variables shell-default-height 30 shell-default-position 'bottom)
      spell-checking syntax-checking version-control)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GSETTINGS NOTIFY ACL LIBSELINUX GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11

ViktorHaag:
This might be because the default location of the cache directory gets set in `core-load-paths.el` (and the location of `spacemacs-auto-save-directory` and `pcache-directory` gets set based on that initial value), so if you reset it later (in dotspacemacs/user-init or dotspacemacs/init), the auto-saves and pcache will still go into the original location, I'd expect.

zabbal:
So there's another variable in addition to spacemacs-cache-directory which governs the location, sets differently and has higher priority? That's highly confusing and seems like a bug to me - it's better to have single source of truth for this kind of stuff.

[...]

PR #16698, opened 2024-12-05
* core/core-load-paths.el: Support spacemacs-cache-directory from environment variables

### LINKED ISSUE #4018: Proposal: configurable cache directory
I like to keep my .emacs.d in my home directory which is NFS mounted, but I'd like to use a local cache for speed reasons.  Would it be possible to have an override for the cache directory.  I don't mind having to specify it as some custom elisp in a startup script, but obviously it would be cleaner in the .spacemacs file.

sunlin7:
Fix #4018 #12127 #15499 by enable customizing the `spacemacs-cache-directory`. 

These three tickets are desiring a way to config the `spacemacs-cache-directory` to split the cache directory from default folder.

This patch will support to change the cache directory with `SPACEMACSCACHE` or `XDG_CACHE_HOME`.

UPDATE: Drop this PR for uncertain change.

### DIFF
--- a/core/core-load-paths.el
+++ b/core/core-load-paths.el
@@ -27,6 +27,8 @@
 ;;; Code:
 
 ;;;; PATH variables/constants
+(when (version<= emacs-version "28")
+  (eval-when-compile (require 'subr-x))) ; for the 'if-let*'
 
 ;; ~/.emacs.d
 (defvar spacemacs-start-directory
@@ -96,7 +98,11 @@
 ;; because Spacemacs may be installed to a shared location and this directory
 ;; and its children should be per-user.
 (defconst spacemacs-cache-directory
-  (concat user-emacs-directory ".cache/")
+  (if-let* ((cache-dir (getenv "SPACEMACSCACHE")))
+      cache-dir
+    (if-let* ((xdg-cache (getenv "XDG_CACHE_HOME")))
+        (expand-file-name "spacemacs/" xdg-cache)
+      (concat user-emacs-directory ".cache/")))
   "Spacemacs storage area for persistent files.")
 
 ;; ~/.emacs.d/.cache/auto-save
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -37,6 +37,7 @@
 - [[#dotfile-configuration][Dotfile Configuration]]
   - [[#dotfile-installation][Dotfile Installation]]
   - [[#alternative-dotdirectory][Alternative dotdirectory]]
+  - [[#alternative-cache-directory][Alternative cache directory]]
   - [[#synchronization-of-dotfile-changes][Synchronization of dotfile changes]]
   - [[#testing-the-dotfile][Testing the dotfile]]
   - [[#dotfile-contents][Dotfile Contents]]
@@ -553,6 +554,16 @@ change the location of this directory.
 *Note*: =~/.spacemacs= will always take priority over =~/.spacemacs.d/init.el=,
 so =~/.spacemacs= must not exist for =~/.spacemacs.d/init.el= to be used by
 Spacemacs.
+
+** Alternative cache directory
+Spacemacs use the cache dir =.cache= under =user-emacs-directory=, normally it
+is =~/.emacs.d/.cache/=.
+
+If the envrionment variable =XDG_CACHE_HOME= exists, Spacemacs will use the
+=spacemacs/= under the CACHE directory, for example =~/.cache/spacemacs/=.
+
+To change the default cache directory, using the environment variable
+=SPACEMACSCACHE=.
 
 ** Synchronization of dotfile changes
 To apply the modifications made in =~/.spacemacs= press ~SPC f e R~. It will


bcc32:
This patch doesn't really fix any of the linked issues.  The variable (even though it was `defconst`) could already be set by the user, but other paths that are based on it (such as the `spacemacs-auto-save-directory` mentioned in #12127) would not respect the user's customized value.

core/core-load-paths.el
```diff
@@ -94,8 +94,8 @@
 ;;
 ;; This is based on `user-emacs-directory', not `spacemacs-start-directory',
 ;; because Spacemacs may be installed to a shared location and this directory
-;; and its children should be per-user.
-(defconst spacemacs-cache-directory
+;; and its children should be per-user. Allow customer define their own value.
```
bcc32: I'm a little confused by the word "customer" here, "user" here would be more idiomatic and consistent with other docs.  (FWIW I don't think this comment adds any value if there's a defcustom immediately below)

core/core-load-paths.el
```diff
@@ -94,8 +94,8 @@
...
+(defcustom spacemacs-cache-directory
   (concat user-emacs-directory ".cache/")
   "Spacemacs storage area for persistent files.")
```
bcc32, line 106: Please specify at least `:type` and `:group` (the latter may sometimes be implied by a `defgroup` in the same file, but there is none in this file).

bcc32 (changes requested):

core/core-load-paths.el
sunlin7: updated, please help review the latest one. Thanks

core/core-load-paths.el
sunlin7: Added these two fields. Thanks.

sunlin7:
Hi @bcc32 I had modified the patch toward your comments. Please help review again. Thanks,
try set the cache value in the `dotspacemacs/user-init`, it will effect on its depends values. 
Keep the variable `spacemacs-auto-save-directory` is for compatibility (some one may has code to modify the value).

bcc32:
There's plenty of other variables whose defaults depend on the value of `spacemacs-cache-directory` and which also need to be updated; `spacemacs-auto-save-directory` was just one example.

Reading the original issues more closely, I don't think actually making this variable customizable is really an appropriate solution here, and definitely merging this PR (if we do so) should not close those issues.

Here are some concerns that this PR does not attempt to address, which I think are pretty valid:

1. Files that contain persistent settings or history are stored in the cache, which is a mistake and error-prone (e.g., you lose meaningful data if you `git clean -n`).
2. Setting the value in `dotspacemacs/user-init` is _still too late_ to affect many of the variables that depend on `spacemacs-cache-directory`, and files are created during early Spacemacs loading before `dotspacemacs/user-init` has a chance to run.  Just as one example of many, `core-load-paths.el` creates the `spacemacs-cache-directory` if it does not exist---but it does so before the user has a chance to customize it, meaning that if the user does set it to something else it may break due to the directory not existing yet.

bcc32:
As an aside, I think it would be good to reconsider most of the places where Spacemacs does all this `setq`-ing of customization options.  It's fine for Spacemacs to provide opinionated defaults for things like `org-default-notes-file` but it should be easier and more obvious how to override such things, and perhaps a better way to do this would be for Spacemacs to provide its defaults in a custom "theme".

Themes are able to provide variable settings, not just face definitions, and in doing so, the custom.el machinery will automatically provide for the ability of the user's customizations to override the theme defaults or the package defaults, regardless of load order.

smile13241324:
I think the only probable way to set these path params is the users env, there is already a spacemacs path env var used for that but it must be  greatly improved.

Starting from the given defaults Spacemacs can calculate its path variables.

This must happen in the core as you don't know where your dotfile is when you need this information.

Also we have to consider the request to make spacemacs xdg conform or at least provide a compatible mode.

sunlin7:
I had tracked `spacemacs-cache-directory`, it does NOT used during the `init.el` and `dotspacemacs/user-init`, so it's safe for user changing the cache path in the `dotspacemacs/user-init` function, 

The key concern in `#4018 #12127 #15499` is customming the `spacemacs-cache-directory`, so this PR should help the user who really want separate their cache directory. 

And for more elegant way to support custom the cache related variables, need very long works, but this PR can be a very small step to start the changes.

bcc32:
My personal opinion: I actually think this PR makes things worse, because it advertises `spacemacs-cache-directory` as a variable that can be customized, but in fact customizing it through the `M-x customize` interface _does not work_, and even `setq`-ing it does not work entirely because several variables are still calculated too early before `dotspacemacs/user-init` runs.

At least with `defconst` it is clear that Spacemacs _does not yet_ support changing the cache location in any meaningful way.

sunlin7:
I had totally refactor the PR, please help review the change again. Thank you !

bcc32:
Hey @sunlin7, thanks for taking a fresh look at this issue.  Just FYI, I don't think it's helpful to edit the original PR description just because it is closed---I found it quite surprising when I was trying to remind myself which of several PRs this was.  It is useful for people who find this PR to be able to see the original context and intentions, and it is easy to just scroll down to the bottom to see the reasoning for the eventual resolution :)

sunlin7:
Hi @bcc32 Agree, and I had restored the original text and had update on the tail. Thanks.

PR #16700, opened 2024-12-06
* core/core-configuration-layer.el: User defined package has high priority on :toggle flag Fix #11872

### LINKED ISSUE #11872: Package deleted on init even if defined in dotspacemacs-additional-packages variable
I added `magic-latex-buffer` to my `dotspacemacs-additional-packages` variables like so:
 `dotspacemacs-additional-packages '(solidity-mode graphql-mode google-c-style color-theme-solarized emojify editorconfig magic-latex-buffer)` I would expect it to be installed on the next start of emacs (like the other packages that I added to the `dotspacemacs-addtional-packages` variables), but it is not.

If I add it manually via the `list-package` + i + e commands it gets installed, but on the next start deleted:
```
(Spacemacs) --> deleting magic-latex-buffer... [1/1]
Package ‘magic-latex-buffer-20170531.5’ deleted.
```

I am using the latest develop branch ( spacemacs 0.300.0 ) and emacs version 26.1 installed by `brew install emacs-plus`.

sunlin7:
Fix #11872 The package in layer configuration file was :toggle off, then user defining it in additional packages will not be installed.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -1246,9 +1246,12 @@ USEDP if non-nil indicates that made packages are used packages."
                        dotspacemacs--additional-theme-packages))
     (let* ((pkg-name (if (listp pkg) (car pkg) pkg))
            (obj (configuration-layer/get-package pkg-name)))
-      (if obj
-          (setq obj (configuration-layer/make-package pkg 'dotfile obj))
-        (setq obj (configuration-layer/make-package pkg 'dotfile)))
+      (if (null obj)
+          (setq obj (configuration-layer/make-package pkg 'dotfile))
+        ;; set :toggle to t for user defined package should be enabled default
+        (unless (and (listp pkg) (memq :toggle pkg))
+          (oset obj :toggle t))
+        (setq obj (configuration-layer/make-package pkg 'dotfile obj)))
       (configuration-layer//add-package obj usedp)))
   (dolist (xpkg dotspacemacs-excluded-packages)
     (let ((obj (configuration-layer/get-package xpkg)))


bcc32:
I believe the `:toggle` keyword is only documented for `<layer>-package` variables, so this would be a new usage if I'm understanding correctly.  Is there any need to support `:toggle` in the `dotspacemacs-additional-packages` list?

Also, I think it would be confusing for us to specifically update the `:toggle` value of the object here, because that will confuse the output of `SPC h d P`.  Reading the original issue, it seems better to just keep the existing behavior and warn the user that they have a contradictory config.

sunlin7:
Hi @bcc32 Thanks for comment, Spacemacs should respect the user configurations, a warning and stopping user using their desired package is not good enough, I believe Spacemacs users are smart and know what they want, so this PR is still valuable for Spacemacs users.

sunlin7:
And `dotspacemacs-additional-packages` does support full package specification in Spacemacs. 
Try 
```
dotspacemacs-additional-packages '((company-quickhelp :location (recipe :fetcher github :repo "company-mode/company-quickhelp")))
```

smile13241324 (approved):
I think that's a valid use case though some more documentation would be desirable

sunlin7:
Thank you !

bcc32:
We should at least fix `SPC h d P`'s behavior with these packages, then, for example by replacing the original package object with the dotfile one outright.  Otherwise, you get this weird behavior:

I added `flyspell-correct-ivy` to my dotspacemacs-additional-packages but do not use the ivy layer so its toggle should be nil:

```
flyspell-correct-ivy is a package declared and configured by the layer `spell-checking'.

You are using this package, it is currently installed but it has not been loaded yet.
```

The above is incorrect; the package is not being configured by that layer at all, and instead this should mention that the package is declared by the dotfile.

bcc32:
By the way there is a syntax error, you are calling `plist-get` with too few arguments so it breaks if you actually use the `:toggle` keyword.

sunlin7:
@bcc32  Thanks for the comment, that's happened on the 1st version in the PR, and the last version of the PR should work correctly https://github.com/syl20bnr/spacemacs/pull/16700/files .

sunlin7:
And I'm trying the `SPC h d P`, give me minutes.

sunlin7:
The description message by pressing the `SPC h d P` is confusing. 
I'm going to fix the messages by pressing `SPC h d P`. Will create another PR to enhance it.
Thank you.

smile13241324:
@sunlin7 Hmm looks like you need to adapt one of the unit tests, please take care.

sunlin7:
@smile13241324  Thank you, and sure, I'm checking it now.

sunlin7:
@smile13241324  A fixing change in #16810, please help review it. Thank you!

PR #16703, opened 2024-12-07
[compleseus] Fix escaping of search inputs

fnussbaum:
`consult-ripgrep` expects Emacs regular expressions, not PCRE. Previously, using <kbd>SPC *</kbd> on an identifier containing `|`, like `spacemacs|defc`, would incorrectly escape the `|` character, for example.


bcc32 (commented):
This works, but it might be better to `regexp-quote` directly instead of taking a trip through PCRE

fnussbaum:
Makes sense, thanks, I had somehow overlooked `regexp-quote`.

PR #16704, opened 2024-12-09
Set save-place-file before starting save-place-mode

xuhdev:
With the current order, Emacs 30 reads the default `save-lace-file` file located at `"~/.emacs.d/places"`, while it writes to `"~/.emacs.d/.cache/places"`.

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/packages.el
+++ b/layers/+spacemacs/spacemacs-defaults/packages.el
@@ -420,12 +420,9 @@
 (defun spacemacs-defaults/init-saveplace ()
   (use-package saveplace
     :init
-    (if (fboundp 'save-place-mode)
-        ;; Emacs 25 has a proper mode for `save-place'
-        (save-place-mode)
-      (setq save-place t))
     ;; Save point position between sessions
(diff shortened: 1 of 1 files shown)


layers/+spacemacs/spacemacs-defaults/packages.el
```diff
@@ -420,12 +420,12 @@
...
     :init
+    ;; Save point position between sessions
+    (setq save-place-file (concat spacemacs-cache-directory "places"))
     (if (fboundp 'save-place-mode)
         ;; Emacs 25 has a proper mode for `save-place'
         (save-place-mode)
-      (setq save-place t))
-    ;; Save point position between sessions
-    (setq save-place-file (concat spacemacs-cache-directory "places"))))
+      (setq save-place t))))
```
xuhdev: We can also simplify this a bit further if Emacs 25- is no longer supported

fnussbaum: Indeed, Spacemacs requires at least Emacs 28.2, so we don't need to set `save-place` anymore.

fnussbaum (approved):

PR #16706, opened 2024-12-09
* core-load-paths.el: Support spacemacs-cache-directory from env variable

### LINKED ISSUE #3589: support XDG Base Directory Specification
The [XDG Base Directory Specification](https://wiki.debian.org/XDGBaseDirectorySpecification) proposes where to store configuration, data and cache files. Basically spacemacs config should go to `$HOME/.config/spacemacs`, cache files to `$HOME/.cache/spacemacs` and data (like backup files) into `$HOME/.local/share/spacemacs`.

This scheme has a few advantages:
* It avoids pollution of $HOME with dotdirs
* It makes it simpler to thread different kinds of files different: don't backup cache files, make sure all config files are version controlled and auto-committed on change, keep cache files in a ramdisk, ...
* A user knows immediately which files are configuration files and which ones shouldn't be touched.

It would be nice of course, if spacemacs could add  `$HOME/.config/spacemacs` as a fallback place to look for its configuration.

Maybe you don't want to do this. In this case please make it possible at least to configure spacemacs in the way described above. It would be enough, if I could provide environment variables to override the default directories defined in [core-load-paths.el](https://github.com/syl20bnr/spacemacs/blob/master/core/core-load-paths.el):

spacemacs constant|current value|why make it configurable
-----------------------------|-----------------------|---
spacemacs-info-directory|spacemacs-core-directory "info/"|Hopefully there will be distribution packages of spacemacs soon. Then the docs will not be in the users homedir anymore.
spacemacs-release-notes-directory|spacemacs-info-directory "release-notes/"|see info-directory, also docs, banners?
spacemacs-cache-directory|user-emacs-directory ".cache/"|for xdg compliance, see above
spacemacs-auto-save-directory|spacemacs-cache-directory "auto-save/"|that's not cache data but important data to backup and my not reside on a ramdisk

sunlin7:
Fix the #12127  by supporting the `spacemacs-cache-directory` from environment variable SPACEMACSCACHE or XDG_CACHE_HOME.

NOTE: 
This PR can be the first step to start supporting the XDG_* directories to fix #3589.

### DIFF
--- a/core/core-load-paths.el
+++ b/core/core-load-paths.el
@@ -27,6 +27,8 @@
 ;;; Code:
 
 ;;;; PATH variables/constants
+(when (version<= emacs-version "28")
+  (eval-when-compile (require 'subr-x))) ; for the 'if-let*'
 
 ;; ~/.emacs.d
 (defvar spacemacs-start-directory
@@ -96,7 +98,14 @@
 ;; because Spacemacs may be installed to a shared location and this directory
 ;; and its children should be per-user.
 (defconst spacemacs-cache-directory
-  (concat user-emacs-directory ".cache/")
+  (if-let* ((spc-cache (getenv "SPACEMACSCACHE")))
+      spc-cache
+    (if-let* ((dot-cache (concat user-emacs-directory ".cache/"))
+              (xdg-cache (getenv "XDG_CACHE_HOME"))
+              (spc-cache (expand-file-name "spacemacs/" xdg-cache))
+              ((file-exists-p spc-cache)))
+        spc-cache
+      dot-cache))
   "Spacemacs storage area for persistent files.")
 
 ;; ~/.emacs.d/.cache/auto-save
--- a/doc/DOCUMENTATION.org
+++ b/doc/DOCUMENTATION.org
@@ -37,6 +37,7 @@
 - [[#dotfile-configuration][Dotfile Configuration]]
   - [[#dotfile-installation][Dotfile Installation]]
   - [[#alternative-dotdirectory][Alternative dotdirectory]]
+  - [[#alternative-cache-directory][Alternative cache directory]]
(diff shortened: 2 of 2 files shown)


fnussbaum:
I think that `XDG_CACHE_HOME/spacemacs/` should not be created and used as `spacemacs-cache-directory` by default, as long as the latter is still used to store data that should persist (and thus should rather be in XDG_{DATA/CONFIG/STATE}_HOME).

sunlin7:
@fnussbaum  Thanks for comment, I had change the code to detect the `XDG_CACHE_HOME/spacemacs/` exists or not, please help review again. CC @bcc32

bcc32:
I think @fnussbaum's comment extends beyond the changes you made.  If I understand correctly, we should not use `$XDG_CACHE_HOME/spacemacs` **even if it exists already** while persistent state data is still stored using `spacemacs-cache-directory`.  So, we'd have to first audit the places where `spacemacs-cache-directory` and create a separate variable, perhaps `spacemacs-state-directory`, for things that should be persisted, before making any of the changes in this PR.  (Please feel free to correct me if I misunderstood your comment).

fnussbaum:
This is a good point, it is indeed better to first clean up the use of the cache directory, thanks @bcc32. (I can see how my comment might have been slightly ambiguous.) 

@sunlin7 Thanks for taking the initiative with #16759! I'll need some more time to review it, as I would like to make sure that we do this correctly for all kinds of files, and document it well, such that users that would like to use the new locations can easily migrate, and only have to do it once. It might also make sense to adopt the [no-littering](https://github.com/emacscollective/no-littering) package, but I am not so sure about it yet.

sunlin7:
@fnussbaum Thank you for looking into this, look forward the progress : )

PR #16708, opened 2024-12-09
* core/core-dotspacemacs.el: Support the XDG_CONFIG_HOME

sunlin7:
Support the environment variable `XDG_CONFIG_HOME`.

### DIFF
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -23,6 +23,9 @@
 (require 'core-load-paths)
 (require 'core-customization)
 
+(when (version<= emacs-version "28")
+  (autoload 'if-let* "subr-x"))
+
 (defconst dotspacemacs-template-directory
   (concat spacemacs-core-directory "templates/")
   "Templates directory.")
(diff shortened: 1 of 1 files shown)


smile13241324 (approved):

nonducor:
This pull request introduced an issue when running emacs in Flatpak: the _sandboxed_ Flatpak environment always overrides `XDG_CONFIG_HOME` to something like `/home/user/.var/app/org.gnu.emacs/config/`, so the `(if-let* ...)` always succeeds. This is most likely not what the user wants (it can be argued that the way Flatpak deals with `XDG_` vars is not correct as well). 

(Would you prefer me to open an issue about it?)

sunlin7:
@nonducor This PR will detect the directory `$XDG_CONFIG_HOME/spacemacs/` exists or not then decide to use the XDG_* folder.

So there is a directory named `spacemacs` under `/home/user/.var/app/org.gnu.emacs/config/` ?

Any suggestion on how to fix this issue? Check the sub files exists or not?

fnussbaum:
@nonducor @sunlin7 I have pushed a small fix in 1d3a715be30796f8384ef9467a47612119e93462, does this solve the problem by any chance?

fnussbaum:
> @nonducor @sunlin7 I have pushed a small fix in [1d3a715](https://github.com/syl20bnr/spacemacs/commit/1d3a715be30796f8384ef9467a47612119e93462), does this solve the problem by any chance?

Sorry, I forgot a pair of parentheses, now fixed in https://github.com/syl20bnr/spacemacs/commit/44dde101e921909a8a897a63b7b1a38f19ca3de7.

nonducor:
Hello @fnussbaum, I have just tested it and it fixes the problem. Thanks a lot!

fnussbaum:
Thanks for reporting this!

sunlin7:
Thank you for fixing the bug.

PR #16712, opened 2024-12-11
* layers/+lang/python/packages.el: Avoid loading semantic in python layer

sunlin7:
Current python layer will call the `semantic-idle-summary-mode` with `0` to disable the feature regardless the `semantic` layer been used or not, but actually the function calling will trigger loading semantic feature (but not active the feature), it's really heavy. 

This PR can avoid the calling to avoid that heavy loading. Please help review and merge the changes. Thanks.

### DIFF
--- a/layers/+lang/python/packages.el
+++ b/layers/+lang/python/packages.el
@@ -450,7 +450,8 @@
         (set-default-toplevel-value x (symbol-value x))))))
 
 (defun python/post-init-semantic ()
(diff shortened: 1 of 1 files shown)


bcc32:
Are you sure this change does anything?  It looks like the `semantic` package is owned by the `semantic` layer, which means this `python/post-init-semantic` function will only be called if the `semantic` layer is used, making the new check, added in this PR, always return non-nil.

sunlin7:
Thanks for the comment, will check it carefully.

PR #16713, opened 2024-12-11
[python] Change imenu-create-index-function only when semantic-mode is activated

sunlin7:
Spacemacs will always choose a value for `imenu-create-index-function` in (`semantic-create-imenu-index`, `python-imenu-create-index`), but actually when user activing `python-ts-mode`, the correct function should be `python-imenu-treesit-create-index`. 

This PR try choose the `semantic-create-imenu-index` only the `semantic-mode` was activated, otherwise using the default function, which will be correct for `python-mode` and `python-ts-mode`.

### DIFF
--- a/layers/+lang/python/funcs.el
+++ b/layers/+lang/python/funcs.el
@@ -543,7 +543,7 @@ If region is not active then send line."
   "Start and/or switch to the REPL."
   (interactive)
   (if-let* ((shell-process (or (python-shell-get-process)
-                              (call-interactively #'run-python))))
+                               (call-interactively #'run-python))))
       (progn
         (pop-to-buffer (process-buffer shell-process))
         (evil-insert-state))
--- a/layers/+lang/python/funcs.el
+++ b/layers/+lang/python/funcs.el
@@ -271,20 +271,9 @@ location of \".venv\" file, then relative to pyvenv-workon-home()."
                       (setq-local pyvenv-activate virtualenv-abs-path))
                      (t (pyvenv-workon virtualenv-path-in-file)
                         (setq-local pyvenv-workon virtualenv-path-in-file))))))))
+
 
 ;; Tests
-
-(defun spacemacs//python-imenu-create-index-use-semantic-maybe ()
-  "Use semantic if the layer is enabled."
-  (setq imenu-create-index-function 'spacemacs/python-imenu-create-index))
-
-;; fix for issue #2569 (https://github.com/syl20bnr/spacemacs/issues/2569) and
-;; Emacs 24.5 and older. use `semantic-create-imenu-index' only when
-;; `semantic-mode' is enabled, otherwise use `python-imenu-create-index'
-(defun spacemacs/python-imenu-create-index ()
-  (if (bound-and-true-p semantic-mode)
-      (semantic-create-imenu-index)
-    (python-imenu-create-index)))
 
 (defun spacemacs//python-get-main-testrunner ()
   "Get the main test runner."
--- a/layers/+lang/python/packages.el
+++ b/layers/+lang/python/packages.el
@@ -454,9 +454,7 @@
   (when (configuration-layer/package-used-p 'anaconda-mode)
     (add-hook 'python-mode-hook
               'spacemacs//disable-semantic-idle-summary-mode t))
-  (spacemacs/add-to-hook 'python-mode-hook
-                         '(semantic-mode
-                           spacemacs//python-imenu-create-index-use-semantic-maybe)))
+  (add-hook 'python-mode-hook 'semantic-mode))
 
 (defun python/pre-init-smartparens ()
   (spacemacs|use-package-add-hook smartparens
--- a/layers/+emacs/semantic/packages.el
+++ b/layers/+emacs/semantic/packages.el
@@ -27,31 +27,11 @@
         srefactor
         ))
 
-;; To fix issue #2569 (https://github.com/syl20bnr/spacemacs/issues/2569),
-;; we store the `imenu-create-index-function' for every semantic buffer, and
-;; restore it when semantic been turned off.
-(defvar-local spacemacs--imenu-create-index-fcn-orig nil
-  "Save the original `imenu-create-index-function'")
-
-(define-advice semantic-new-buffer-fcn
-    (:before (&rest _) store-imenu-create-index-function)
-  (setq spacemacs--imenu-create-index-fcn-orig imenu-create-index-function))
-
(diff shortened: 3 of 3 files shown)


bcc32:
There's a subtle behavior change where, if the user toggles `semantic-mode`, the function would previously have correctly switched between imenu backends, but now it only decides once at the major-mode hook.  Do you know if toggling `semantic-mode` is otherwise a sensible thing to do, or do we not have to care about it?  (in which case you can disregard this paragraph)

Also, is it always the case that `semantic-mode` will be enabled early enough for this hook?  From reading the code it seems that `spacemacs//python-imenu-create-index-use-semantic-maybe` is placed in `python-mode-hook` before `semantic-mode`, which means it no longer works correctly with semantic enabled.

Please test both with and without semantic layer enabled, to double-check that the effective configuration is correct.

sunlin7:
Hi @bcc32  Thanks for comment, and you're right that there's a subtle behavior changed.
I had make the changes and tested with both `semantic` and `python` layers enabled, then toggling the `semantic-mode` in a opened python buffer, make sure every thing work as expected. 

Please help review the latest changes. Thank you.

layers/+lang/python/funcs.el
```diff
@@ -271,21 +271,23 @@ location of \".venv\" file, then relative to pyvenv-workon-home()."
...
-    (python-imenu-create-index)))
+  "Use semantic function if `semantic' is activated."
+  (setq-local imenu-create-index-function
+              #'(lambda ()
+                  (cond
+                   ((bound-and-true-p semantic-mode)
+                    (semantic-create-imenu-index))
+                   ((boundp 'imenu-create-index-function-orig)
+                    (funcall imenu-create-index-function-orig))
+                   (funcall (default-value 'imenu-create-index-function))))))
```
bcc32: This should still be a named function rather than a lambda.  If a user is trying to debug their imenu config it is useful to be able to describe-function.

layers/+lang/python/packages.el
```diff
@@ -454,9 +454,13 @@
   (when (configuration-layer/package-used-p 'anaconda-mode)
     (add-hook 'python-mode-hook
               'spacemacs//disable-semantic-idle-summary-mode t))
-  (spacemacs/add-to-hook 'python-mode-hook
-                         '(semantic-mode
-                           spacemacs//python-imenu-create-index-use-semantic-maybe)))
+  (add-hook 'python-mode-hook
+            #'(lambda ()
+                (setq-local imenu-create-index-function-orig
+                            imenu-create-index-function)
```
bcc32: This new variable should have a definition and also should be properly namespaced (e.g., start with `python//`)

layers/+lang/python/packages.el
```diff
@@ -454,9 +454,13 @@
...
+                (semantic-mode)))
```
bcc32: `semantic-mode` turns out to be a global mode so there isn't any benefit to enabling it in the same hook, as the ordering won't affect any buffers other than the first.  I think the previous add-hook is probably fine here.

layers/+lang/python/packages.el
```diff
@@ -454,9 +454,13 @@
...
+  (add-hook 'semantic-mode-hook
+            'spacemacs//python-imenu-create-index-use-semantic-maybe))
```
bcc32: `semantic-mode` is a global mode so this might now run in non-python buffers which is a bug.  Instead this should probably just be on python-mode-hook.

bcc32 (changes requested):

sunlin7:
Thanks for comment, will research more and then update this PR.

sunlin7:
Hi @bcc32  I had refactored the solution, save the original `imenu-create-index-function` before semantic changes it, and restore it to original value when semantic been turned off.

Here are the reproducing steps:
1. enable the `semantic`, `python`, `imenu-list` layers in Spacemacs
2. start Spacemacs and open a "a.py" write a python code
```
def hello(): pass
hello()
```
3.  press `SPC b i` to call `spacemacs/imenu-list-smart-focus`, check the imenu list, press `q` to quit the imenu list.
4. turn off semantic by pressing `M-x semantic-mode` to toggle the semantic mode, 
5. repeat step 3, the imenu content will be empty for semantic already turned off but the `imenu-create-index-function` is still semantic one.

Now this patch can fix the issue, by repeat the step 3/4 with this patch, the imenu list will work as expected.

bcc32:
Hmm, I don't seem to get the correct behavior as you described, with your branch.  In my test buffer I'm seeing `spacemacs--imenu-create-index-fcn-orig` be equal to `semantic-create-imenu-index`, so `imenu-create-index-function` is not restored when semantic-mode is disabled.  Seems like `semantic-new-buffer-fcn` may run more than once in the same buffer.
Perhaps it is futile to try to restore the index function (the body of `semantic-mode` has some comments suggesting so).

If I understand correctly, your primary motivation is to avoid unconditionally setting `imenu-create-index-function` in Python-mode, and only set it to something if semantic-mode is enabled, right?

I did a bit more research and it seems that `semantic-new-buffer-fcn` will already set up `imenu-create-index-function` to be `semantic-imenu-create-index` in Python buffers (due to `semantic-new-buffer-setup-functions`).  So actually, Spacemacs doesn't need to do anything to configure semantic, and the workaround for #2569 can simply be removed (it was for Emacs 24.5 and older).

I pushed a new commit to just remove this old code and I think it has the nice side-effect of fixing what I understand to be your motivating concern here.  Please let me know if the new patch works for you and if so it's a nice cleanup :)

sunlin7:
Hi @bcc32  Thanks for the comment, turn off semantic without restore the `imenu-create-index-function` will lead an EMPTY imenu-list, the reproducing steps were listed as my previous comment. Only restore the function to valid one then the imenu list will works again. So the restoring value is necessary.  As you mentioned toggle semantic multiple times, I'm going to research the scenario. Thanks

bcc32:
Yeah, sorry for the confusion.  After the research I did into #2569 I realized it was not designed to deal with turning off `semantic-mode`, it was instead fixing an issue where loading `semantic` modified `imenu-create-index-function` in Python buffers, regardless of whether `semantic-mode` was enabled.

This history, plus reading the comment in `semantic-mode`'s body itself, suggests to me that Spacemacs didn't attempt to handle turning off `semantic-mode` in the past, and that we actually don't necessarily have to handle that now.

Hope that makes sense.

sunlin7:
Hi @bcc32  That makes sense. Thank you.

PR #16715, opened 2024-12-11
[git] Fix forge key bindings

fnussbaum:
Both `forge-issue-mode` and `forge-pullreq-mode` are derived from `forge-topic-mode`, however, bind-map does not seem to activate maps corresponding to parent modes. Note that `forge-topic-mode` is never used directly according to its docstring.


fnussbaum:
Some of the commands got renamed, I have fixed those in the second commit.

bcc32:
Spacemacs has a function `spacemacs/inherit-leader-keys-from-parent-mode` which might be better here, so that the key bindings are still defined only in one map.

fnussbaum:
Nice, thanks!

PR #16716, opened 2024-12-11
* layers/+spacemacs/spacemacs-evil/config.el: Apply evil-collection for info-mode

sunlin7:
Apply `evil-collection` for `info-mode`, without this PR the `info-mode` has key-binding issues like pressing "RET"... keys in a info buffer does not work.


bcc32:
Nice, thanks!

sunlin7:
Thank you.

PR #16718, opened 2024-12-11
* layers/+lang/python/packages.el: Bind jump back on major mode init function

sunlin7:
The python-layer key binding `, g g` is bound by the major mode, but the `, g b` is bound by a optional package, will has issue if the optional package been excluded.

Here is the example to trigger the issue, enable the `python` layer and exclude the `anaconda-mode`, then open a python file in Spacemacs, try the `, g g` first, then try `, g b` will not work. 
```
dotspacemacs-excluded-packages '(anaconda-mode)
dotspacemacs-configuration-layers '(python ...)
```

This PR try bind the jump back `, g b` to the major mode to fix the issue.

### DIFF
--- a/layers/+lang/python/packages.el
+++ b/layers/+lang/python/packages.el
@@ -74,7 +74,6 @@
     (spacemacs/set-leader-keys-for-major-mode 'python-mode
       "hh" 'anaconda-mode-show-doc
       "ga" 'anaconda-mode-find-assignments
-      "gb" 'xref-pop-marker-stack
       "gu" 'anaconda-mode-find-references)
     ;; new anaconda-mode (2018-06-03) removed `anaconda-view-mode-map' in
     ;; favor of xref. Eventually we need to remove this part.
@@ -396,6 +395,7 @@
       "cc" 'spacemacs/python-execute-file
       "cC" 'spacemacs/python-execute-file-focus
       "db" 'spacemacs/python-toggle-breakpoint
+      "gb" 'xref-pop-marker-stack
       "ri" 'spacemacs/python-remove-unused-imports
       "sB" 'spacemacs/python-shell-send-buffer-switch
       "sb" 'spacemacs/python-shell-send-buffer


bcc32:
I don't have a strong objection here, but does `, g g` actually work in python-mode without anaconda-mode?

I think it's the case that the anaconda backend provides the setting for `spacemacs-jump-handlers-python-mode` which makes `, g g` work (through xref), so it seems to make sense to configure `, g b` in the same place (the same key bindings when using the lsp backend are configured separately in the LSP layer).

BTW if you just exclude anaconda-mode without setting `python-backend` appropriately (to not use anaconda), you will also get errors like:

```
File local-variables error: (void-function anaconda-mode)
```

So, if we want to support Python functionality without using anaconda-mode, there are further changes that need to happen.

sunlin7:
The `python-backend` is `lsp` but actually the file is in '/tmp/' which is lsp blocked.
The `spacemacs-jump-handlers-python-mode` is `nil`, so the jump-handler is `spacemacs-default-jump-handlers` (evil-goto-definition) for `python-mode`, which can jump to a function definition inside the testing file `/tmp/a.py` (pretty sure the jumping does not depends on the lsp feature).

bcc32:
Isn't `, g b` still bound to `xref-pop-marker-stack`, though, due to the lsp layer?  (I don't have python lsp set up on my machine but I'm just reading the source of `spacemacs//lsp-bind-simple-navigation-functions`).

sunlin7:
enable the lsp layer, open the testing file '/tmp/a.py' and block it to lsp workspace during the lsp question. Then ', g g' dose not working.On Dec 11, 2024 22:33, Aaron Zeng ***@***.***> wrote:
Isn't , g b still bound to xref-pop-marker-stack, though, due to the lsp layer?  (I don't have python lsp set up on my machine but I'm just reading the source of spacemacs//lsp-bind-simple-navigation-functions).

bcc32:
Sorry, I don't understand your last comment and I think the formatting got messed up, too.  Can you clarify?

sunlin7:
My previous message was replied from an email client, got wrong format, sorry about that. 

> Isn't `, g b` still bound to `xref-pop-marker-stack`, though, due to the lsp layer? (I don't have python lsp set up on my machine but I'm just reading the source of `spacemacs//lsp-bind-simple-navigation-functions`).

No, the `, g b` is not binded when the anaconda package was excluded; so press `, g b` will get an user error message says ", g b is undefined". 

At that situation the `, g g` still works via `evil-goto-definition`, then the `xref-pop-marker-stack` still work. 

That's why this PR moves the key binding `, g b` configuration from the `*/init-anaconda-mode` function to the `*/init-python` function.

bcc32:
Wait so just to be clear here, in your setup where you report that `, g b` does not work, you are using _neither_ anaconda _nor_ lsp?  What is `python-backend` set to?

sunlin7:
This anaconda was excluded, and the `python-backend` is lsp, but, the opened file is `/tmp/a.py` which does NOT added to lsp workspace (when opening the a.py, the lsp will ask to start a workspace, ignore or decline to avoid the lsp starting on the directory).

bcc32:
Ah, okay, thanks for explaining.  I guess it's harmless to have the binding in any case, so I'm going to go ahead and merge, thanks for the PR!

Separately, it occurs to me that maybe anaconda-mode shouldn't even be installed if `python-backend` is lsp, e.g., use `:toggle` in the packages list rather than `:if` in the `use-package` macro?  Not sure though, and in any case doesn't need to be in this PR.

sunlin7:
Thank you more!

PR #16724, opened 2024-12-14
* layers/+tools/shell/funcs.el: Avoid loading semantic for eshell

sunlin7:
The `eshell` layer try to exclude the `semantic-mode` but leading a heavy loading. Reproducing steps:
1. Enable the layers `semantic`, `eshell`.
2. Start Spacemacs and then start the `eshell`, it will lead the `semantic` feature been loaded (but not activated).

This PR speeds up the eshell by avoiding loading the `semantic` feature.

### DIFF
--- a/layers/+tools/shell/funcs.el
+++ b/layers/+tools/shell/funcs.el
@@ -173,7 +173,7 @@ is achieved by adding the relevant text properties."
             'spacemacs//eshell-auto-end nil t)
   (add-hook 'evil-hybrid-state-entry-hook
             'spacemacs//eshell-auto-end nil t)
-  (when (configuration-layer/package-used-p 'semantic)
+  (when (bound-and-true-p semantic-mode)
     (semantic-mode -1))
   ;; This is an eshell alias
   (defun eshell/clear ()


fnussbaum (approved):
The change is fine, though I think we should reconsider disabling a global minor mode in a mode hook like that.

sunlin7:
Agree, I'm going to research the way to disable semantic mode locally.

fnussbaum:
I am not sure why it gets disabled at all, perhaps it is does not need to be.

sunlin7:
The action turning off `semantic-mode` for `eshell` was introduced in the commit
https://github.com/syl20bnr/spacemacs/commit/40487778e697a680c762fccde5c7407cc2a68277
there is no comment why need turn off it.

I tried the eshell with turning on semantic mode, try function `semantic-mode-ia-fast-jump`, there is a message says `Cannot analyze buffers not supported by Semantic`, same behavior on Spacemacs and vanilla Emacs.

sunlin7:
I create the PR #16728 to remove the line.

PR #16725, opened 2024-12-14
* layers/+lang/emacs-lisp/packages.el: Config `elisp-def` in its own package

sunlin7:
Currently the `elisp-def` package actually is configured with `elisp-slime-nav ` package, will have issue if we add the `elisp-def` to the `dotspacemacs-excluded-packages`. 

This PR configures the `elisp-def` in it's own package initialization function, then user can put it in the excluded package list.

### DIFF
--- a/layers/+lang/emacs-lisp/packages.el
+++ b/layers/+lang/emacs-lisp/packages.el
@@ -159,7 +159,11 @@
 
 (defun emacs-lisp/init-elisp-def ()
   (use-package elisp-def
-    :defer t))
+    :defer t
+    :init
+    (dolist (mode '(emacs-lisp-mode lisp-interaction-mode))
+      (let ((jumpl (intern (format "spacemacs-jump-handlers-%S" mode))))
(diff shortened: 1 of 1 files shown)


layers/+lang/emacs-lisp/packages.el
```diff
@@ -159,7 +159,12 @@
 
 (defun emacs-lisp/init-elisp-def ()
   (use-package elisp-def
-    :defer t))
+    :defer t
+    :commands (elisp-def)
```
bcc32: This command is already autoloaded by the package itself so this line is redundant.  LGTM otherwise

bcc32 (changes requested):

layers/+lang/emacs-lisp/packages.el
sunlin7: Sure, removed this line on the latest change. Thanks

bcc32 (approved):

PR #16729, opened 2024-12-15
* layers/+lang/emacs-lisp/packages.el: Bind xref-go-back on major mode 

sunlin7:
The `emacs-lisp-mode` will no jump-back key binding if we exclude the `elisp-slime-nav` package. 
Just put `elisp-slime-nav` into the `dotspacemacs-excluded-packages`, then starting the Spacemacs and open a elisp file, press `, g g` to jump to a elisp function/variable, then press `, g b`/`xref-go-back` will not work. 

This PR will fix the issue. Please help review the change, thanks.


bcc32:
Should the related `spacemacs/declare-prefix-for-mode` call also be moved into `emacs-lisp/init-emacs-lisp`?

```
      (spacemacs/declare-prefix-for-mode mode "mg" "find-symbol")
```

bcc32 (approved):

PR #16730, opened 2024-12-17
* layers/+web/eww/funcs.el: Fix eww transient state key bindings

sunlin7:
Warning message happened when enable the `eww` layer:
```
Warning (emacs): Unrecognized key: _+_ 
Warning (emacs): Unrecognized key: _w_ 
Warning (emacs): Unrecognized key: _-_ 
Warning (emacs): Unrecognized key: _=_ 
```
It caused by the transient state for eww has post-defined keys, but the keys not available during transient hints initilizating.

The is PR will fix the issue by define the keys with `identify` function first, then replay them with the real keys.

### DIFF
--- a/layers/+web/eww/funcs.el
+++ b/layers/+web/eww/funcs.el
@@ -34,19 +34,6 @@
 
 
 ;;;; Transient State
-(defvar spacemacs--eww-ts-use-evil nil
-  "Transient State enable Evil key bindings")
-
-(defvar spacemacs--eww-ts-use-writeroom nil
-  "Transient State enable writeroot")
-
-(defvar spacemacs--eww-ts-use-zoom-frm nil
-  "Transient State enable zoom-fmr key bindings")
-
-(defun spacemacs//eww-key-placeholder ()
-  "Just a placeholder for a keybinding."
-  (interactive)
-  (message "The feature is not available/enabled."))
 
 (defun spacemacs//eww-ts-toggle-hint ()
   "Toggle full hint docstring for EWW transient state."
@@ -58,25 +45,9 @@
   "Return a condensed/full hint for the eww transient state"
   (concat
    " "
-   (if (not spacemacs--eww-ts-full-hint-toggle)
-       (concat "[" (propertize "?" 'face 'hydra-face-red) "] help")
-     (let ((result spacemacs--eww-ts-full-hint))
-       (cl-loop for (flag . reg) in
-                '((spacemacs--eww-ts-use-evil . "\\[[^\\[]+ scroll [^\\[]*")
-                  (spacemacs--eww-ts-use-writeroom . "\\[[^\\[]+writeroom[^\\[]*")
-                  (spacemacs--eww-ts-use-zoom-frm . "\\[[^\\[]+zoom-in[^\\[]*")
-                  (spacemacs--eww-ts-use-zoom-frm . "\\[[^\\[]+zoom-out[^\\[]*")
-                  (spacemacs--eww-ts-use-zoom-frm . "\\[[^\\[]+unzoom[^\\[]*"))
-                do
-                (when-let* (((not (symbol-value flag)))
-                            ((string-match reg result))
-                            (start (match-beginning 0))
-                            (end (match-end 0)))
-                  (setq result
-                        (concat (substring result 0 start)
-                                (make-string (- end start) ?\s)
-                                (substring result end)))))
-       result))))
+   (if spacemacs--eww-ts-full-hint-toggle
+       spacemacs--eww-ts-full-hint
+     (concat "[" (propertize "?" 'face 'hydra-face-red) "] help"))))
 
 (defun spacemacs//eww-setup-transient-state ()
   "Setup EWW transient state with toggleable help hint.
@@ -95,7 +66,8 @@ full hint text will not show up!"
   [_<_/_>_] history back/forw^^^^  [_c_] toggle colors            [_=_] unzoom        [_B_] list bookmarks [_d_] download
   [_[_/_]_] page previous/next^^^^ [_t_] toggle latex             ^^                  [_V_] view source    [_B_] add bookmark
   [_u_] page up^^^^^^              [_C_] cycle theme              ^^                  ^^                   [_q_] quit
-  [_t_] top url^^^^^^")
+  [_t_] top url^^^^^^"
+    '("w" "+" "-" "="))
   (spacemacs|define-transient-state eww
     :title "Eww Transient State"
     :hint-is-doc t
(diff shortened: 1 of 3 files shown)


layers/+web/eww/funcs.el
```diff
@@ -90,16 +114,16 @@ full hint text will not show up!"
     ("u" eww-up-url)
     ("t" eww-top-url)
     ;; Layout/Appearance
-    ;; ("w" writeroom-mode)
+    ("w" identity)
```
bcc32: `identity` is not a valid key binding definition.  If the user presses this key and they are not using the `writeroom-mode` package this errors:

```
wrong type argument: commandp, identity
```

layers/+web/eww/funcs.el
```diff
@@ -45,9 +53,25 @@
...
+     (let ((result spacemacs--eww-ts-full-hint))
+       (cl-loop for (flag . reg) in
+                '((spacemacs--eww-ts-use-evil . "\\[[^\\[]+ scroll [^\\[]*")
+                  (spacemacs--eww-ts-use-writeroom . "\\[[^\\[]+writeroom[^\\[]*")
+                  (spacemacs--eww-ts-use-zoom-frm . "\\[[^\\[]+zoom-in[^\\[]*")
+                  (spacemacs--eww-ts-use-zoom-frm . "\\[[^\\[]+zoom-out[^\\[]*")
+                  (spacemacs--eww-ts-use-zoom-frm . "\\[[^\\[]+unzoom[^\\[]*"))
+                do
+                (when-let* (((not flag))
+                            ((string-match reg result))
```
bcc32: I don't think this works as expected, you need to check the value of the variable whose name is stored in `flag`, not the value of `flag` itself.  Either use `symbol-value` or don't quote the symbols in the iterated list.

bcc32 (commented):
Can you provide some more details?  I'm not able to reproduce the issue you described.  I added the `eww` layer, started a new Emacs, and didn't see any warnings at startup, or when I started `eww`, or when I started using the transient state.

sunlin7:
@bcc32  Could you try the eww layer with emacs-28 branch? It constantly reproducible on Emacs 28.

sunlin7:
I tried emacs-29, this issue is also reproducible.

bcc32:
Hmm, I think I know what this is.  I can reproduce the issue you describe if I have a running Spacemacs, and I add the eww layer, and then I press `SPC f e R` to install the layer after startup is complete.  The issue **does not** occur if I add the eww layer and then start a new Spacemacs.  Is that the behavior you're describing?  It's not clear from your previous comments.

If that's what we're talking about, this is not really a bug in the eww layer, I would argue.  (Otherwise we would have to fix every single callsite of `spacemacs/transient-state-register-add-bindings` and not just eww).

`spacemacs|define-transient-state` is supposed to already deal with this by deferring the actual `defhydra` using `spacemacs/defer-until-after-user-config`, so it normally works fine (`defhydra` always runs after the `eww/post-init-zoom-frm` function, for example).

If we're talking about the same thing, then I think this more a bug in `SPC f e R`.  We should fix `SPC f e R` to run things in the correct order to avoid the spurious warning.

sunlin7:
@bcc32  It's not same as I described issue. 
I found the key point to reproduce the issue:
```
dotspacemacs-distribution 'spacemacs-base
dotspacemacs-configuration-layers '(eww)
```

bcc32:
Can you give a full config to reproduce?  I of course already tried adding eww to my layers list (and I'm already using the spacemacs-base distribution).

sunlin7:
@bcc32  The configuration is the default content of `core/templates/.spacemacs.template` with two lines changes:
```
dotspacemacs-distribution 'spacemacs-base
dotspacemacs-configuration-layers '(eww)
```

I tried emacs-28/29/30 on Ubuntu 20.04, the warning messages will be displayed on startup,
```
Warning (emacs): Unrecognized key: _+_ Disable showing Disable logging
Warning (emacs): Unrecognized key: _w_ Disable showing Disable logging
Warning (emacs): Unrecognized key: _-_ Disable showing Disable logging
Warning (emacs): Unrecognized key: _=_ Disable showing Disable logging
```

Please keep the `warning-suppress-log-types` to be `nil`.

sunlin7:
I'm not sure any other additional layers/package may hide this issue.

bcc32:
Ohhhh.  You're saying to make `eww` the *only* layer.  That wasn't clear at all from your previous messages.  (In general, for the avoidance of doubt, it is best to just paste a full `.spacemacs` that can be directly loaded to attempt to reproduce, otherwise, there is always the chance of misinterpretation).

Indeed, if the packages like `zoom-frm` and `writeroom-mode` are simply not enabled, the warning is issued.  Let me take another look at this, then.

bcc32:
I pushed a new proposed solution to your branch.  It attempts to be more general and reusable across other transient states, while requiring minimal duplicated code (such as regexes to match each key hint) in each transient state definition.

Unfortunately it doesn't work for the evil-mode key bindings here, but if I'm understanding your original comment correctly, the evil mode keys are actually never missing.  (In fact, spacemacs-bootstrap requires evil so I don't think basically anything else in Spacemacs supports working without loading evil-mode).  So, I did not attempt to handle the formatting in that case.

sunlin7:
@bcc32  Thank you, and the the changes are more elegant. 
I had test it on my local, there has warning
`Warning (emacs): Unrecognized key: _+_ Disable showing`

And the transient status menu has a `[_+_]` in the menu list, cause a layout issue. 
```
 Navigation                Layout/Appearance            Zoom              List/view          Other
 ------------------------- ---------------------------  ----------------- ------------------ ----------------------------
  [h/j/k/l] scroll l/d/u/r [v] toggle visual-line-mode  [_+_] zoom-in       [W] list buffers   [r] reload page
  [H/L] prev/next eww-buff                                                [S] list histories [x] view in external browser
```

sunlin7:
How to deal the dynamic loaded feature? I mean when the transient state menu started, a package (eg: zoom-frm) did not been loaded yet, but some how the package was loaded for some reason, then re-run the transient state menu, the menu items should be presented.

bcc32:
> I had test it on my local, there has warning `Warning (emacs): Unrecognized key: _+_ Disable showing`
> 
> And the transient status menu has a `[_+_]` in the menu list, cause a layout issue.

This should be fixed now, thank you for catching the issue.

bcc32:
> How to deal the dynamic loaded feature? I mean when the transient state menu started, a package (eg: zoom-frm) did not been loaded yet, but some how the package was loaded for some reason, then re-run the transient state menu, the menu items should be presented.

What scenario do you mean by "the package was loaded for some reason"?  I think the transient state will exclude those key bindings if the `spacemacs-visual` or `spacemacs-editing-visual` layers are disabled.  If the package is loaded somehow after startup, the `eww/post-init-*` functions will not run and so the transient state will not include the key bindings anyway.

sunlin7:
@bcc32  Got your points, and yeh, currently I have no `dynamic` example that package been loaded after `eww/post-init-*`.

I have no other questions about the changes. Thank you.

smile13241324 (approved):

PR #16734, opened 2024-12-18
[shell] Support multi-vterm in spacemacs/projectile-shell

sunlin7:
To support the multi-vterm in projectile. 
Enable the shell layer with `multi-vterm`, then open a projectile directory (eg: `~/.emacs.d/`), then try `SPC p $` to call `spacemacs/projectile-shell`, which does not work with `multi-vterm`.
```
(shell :variables shell-default-shell 'multi-vterm)
```
The PR will support the `multi-vterm` for projectile.

### DIFF
--- a/layers/+tools/shell/funcs.el
+++ b/layers/+tools/shell/funcs.el
@@ -31,12 +31,6 @@ and move your focus to it; to switch the current buffer view, use
   (let ((default-directory (projectile-acquire-root)))
     (call-interactively 'spacemacs/default-pop-shell)))
 
-(defun spacemacs//projectile-run-shell-in-root ()
-  "Invoke `shell-default-shell' in the project's root directory."
-  (interactive)
-  (projectile-with-default-dir (projectile-project-root)
-    (call-interactively shell-default-shell)))
-
 (defun spacemacs/projectile-shell ()
   "Create a shell buffer at the project root and switch to it.
 Customize `shell-default-shell' to control what type of shell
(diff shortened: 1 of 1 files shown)


layers/+tools/shell/funcs.el
```diff
@@ -31,6 +31,12 @@ and move your focus to it; to switch the current buffer view, use
   (let ((default-directory (projectile-acquire-root)))
     (call-interactively 'spacemacs/default-pop-shell)))
 
+(defun spacemacs//projectile-run-shell-in-root (&optional shell)
+  "Invoke `multi-term' in the project's root."
+  (interactive)
+  (projectile-with-default-dir (projectile-project-root)
+    (call-interactively (intern-soft (or shell shell-default-shell)))))
```
bcc32: `shell` is already known to be a symbol, there is no need to call `intern-soft`

layers/+tools/shell/funcs.el
```diff
@@ -31,6 +31,12 @@ and move your focus to it; to switch the current buffer view, use
   (let ((default-directory (projectile-acquire-root)))
     (call-interactively 'spacemacs/default-pop-shell)))
 
+(defun spacemacs//projectile-run-shell-in-root (&optional shell)
+  "Invoke `multi-term' in the project's root."
```
bcc32: Please update this docstring.

layers/+tools/shell/funcs.el
```diff
@@ -31,6 +31,12 @@ and move your focus to it; to switch the current buffer view, use
...
+  (interactive)
```
bcc32: I don't think it's a good idea to have an interactive command whose name containing `//` indicates it is a "private" function.  Probably not necessary to make this interactive anyway, and users should just use `spacemacs/projectile-shell`.

layers/+tools/shell/funcs.el
bcc32: Also, since this is an internal private function, and the `shell` argument is never used, I think we can just get rid of it.

bcc32 (changes requested):

layers/+tools/shell/funcs.el
sunlin7: Document is updated.

layers/+tools/shell/funcs.el
sunlin7: Arguments was removed, and `intern-soft` also was removed.

layers/+tools/shell/funcs.el
sunlin7: The `interactive` is required for the caller will call this function by `call-interactively`.

sunlin7:
@bcc32  I had push the latest changes, please help review again. Thanks

layers/+tools/shell/funcs.el
bcc32: Removed this need by refactoring `spacemacs/projectile-shell`

bcc32 (approved):

bcc32:
@sunlin7 please see if the latest version works for you

sunlin7:
@bcc32  it works for me. Thank you.

PR #16735, opened 2024-12-18
Remove empty post-init-counsel-gtags functions

bcc32:
Follow up to #16476.

### DIFF
--- a/layers/+lang/d/packages.el
+++ b/layers/+lang/d/packages.el
@@ -30,8 +30,7 @@
         d-mode
         flycheck
         (flycheck-dmd-dub :requires flycheck)
-        ggtags
-        counsel-gtags))
+        ggtags))
 
 
 (defun d/post-init-company ()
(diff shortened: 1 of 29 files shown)


layers/+lang/d/packages.el
```diff
@@ -30,8 +30,7 @@
         d-mode
         flycheck
         (flycheck-dmd-dub :requires flycheck)
-        ggtags
```
smile13241324, line 33: I think ggtags is still used

smile13241324 (changes requested):

smile13241324:
@bcc32 I have had a quick look, I think I have found one line which would still be necessary. The rest should be ok.

layers/+lang/d/packages.el
bcc32: Yes, it's still there after the PR, looks like GitHub just doesn't have good word-level diff rendering.

(Thank you for looking so closely at this, though, it's exactly the kind of mistake I was worried I'd make).

smile13241324 (approved):

PR #16736, opened 2024-12-18
[geolocation] Small fixes and cleanups

bcc32:
1. Remove the deprecated extensions.el file which is no longer used
1. Handle themes that have a :package property correctly


bcc32:
@smile13241324 would appreciate a quick confirmation that `extensions.el` is indeed not used by anything anymore.  (I think the deprecation I saw pre-dates me following this repo closely).

smile13241324:
I'll have a look

smile13241324 (approved):
Looks good

PR #16740, opened 2024-12-18
chore: remove initialization warning about pyright and lsp formatting

kassick:
The warning is misleading -- the user may believe that it is not possible to use lsp formatting when pyright is used -- this is wrong, since add-on servers (such as ruff) may provide code formatting and be used alongside the primary language server.

A better approach is to let the user set it up and warn only if lsp-format-buffer fails due to lack of server capabilities.

### DIFF
--- a/layers/+lang/python/README.org
+++ b/layers/+lang/python/README.org
@@ -234,8 +234,10 @@ One of [[https://github.com/google/yapf][YAPF]], [[https://github.com/ambv/black
 
 The default =python-formatter= is =yapf=, with the following exception: When
 both the =lsp= layer is used, and =python-lsp-server= is =pylsp=, the default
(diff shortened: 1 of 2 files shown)


kassick:
@fnussbaum FYI

fnussbaum (approved):

fnussbaum:
Sorry about that, and thanks!

kassick:
No worries!

What do you think about generalizing this to a `spacemacs-lsp-format-on-save` minor mode?

There is no real dependency on python or any specific server ...

fnussbaum:
> What do you think about generalizing this to a `spacemacs-lsp-format-on-save` minor mode?

I agree, that would be a definite improvement.

PR #16741, opened 2024-12-18
* layers/+lang/c-c++/packages.el: fill the :toggle to avoid unused packages

sunlin7:
Fill the :toggle to avoid install the unused packages.

### DIFF
--- a/layers/+lang/c-c++/packages.el
+++ b/layers/+lang/c-c++/packages.el
@@ -41,7 +41,7 @@
     srefactor
     xcscope
     ;; lsp
-    (ccls :requires lsp-mode)
+    (ccls :requires lsp-mode :toggle (eq c-c++-backend 'lsp-ccls))
     dap-mode
     ;; rtags
     (company-rtags :requires (company rtags))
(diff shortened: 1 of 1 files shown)


layers/+lang/c-c++/packages.el
```diff
@@ -41,23 +41,27 @@
...
     ggtags
-    (helm-rtags :requires (helm rtags))
-    (ivy-rtags :requires (ivy rtags))
-    rtags
+    (helm-rtags :requires (helm rtags)
+                :toggle (and (eq c-c++-backend 'rtags)
+                             (configuration-layer/layer-used-p 'helm)))
+    (ivy-rtags :requires (ivy rtags)
+               :toggle (and (eq c-c++-backend 'rtags)
+                            (configuration-layer/layer-used-p 'ivy)))
```
bcc32: I think the `:requires ivy` is probably sufficient, no need for `layer-used-p` here IMO.  (Similarly for helm)

layers/+lang/c-c++/packages.el
```diff
@@ -41,23 +41,27 @@
...
     xcscope
     ;; lsp
-    (ccls :requires lsp-mode)
+    (ccls :requires lsp-mode :toggle (eq c-c++-backend 'lsp-ccls))
     dap-mode
     ;; rtags
-    (company-rtags :requires (company rtags))
-    (flycheck-rtags :requires (flycheck rtags))
+    (company-rtags :requires (company rtags) :toggle (eq c-c++-backend 'rtags))
+    (flycheck-rtags :requires (flycheck rtags) :toggle (eq c-c++-backend 'rtags))
```
bcc32: These packages already depend on rtags which is owned by this layer, so I think these toggles are probably redundant.

layers/+lang/c-c++/packages.el
```diff
@@ -41,23 +41,27 @@
...
+                             (configuration-layer/layer-used-p 'helm)))
+    (ivy-rtags :requires (ivy rtags)
+               :toggle (and (eq c-c++-backend 'rtags)
+                            (configuration-layer/layer-used-p 'ivy)))
+    (rtags :toggle (eq c-c++-backend 'rtags))
     ;; ycmd
-    (company-ycmd :requires company)
-    (flycheck-ycmd :requires flycheck)
+    (company-ycmd :requires company :toggle (eq c-c++-backend 'ycmd))
+    (flycheck-ycmd :requires flycheck :toggle (eq c-c++-backend 'ycmd))
```
bcc32: I think it would be cleaner to do something similar to what we do for rtags above.

sunlin7:
@bcc32 Thank you for the comments, you're right, and I had pushed the enhanced changes. Please help review again.

bcc32 (approved):

Issue #16529, opened 2024-08-23
Opening python files is very slow


sid-kap:
#### Description :octocat:
Opening python in python-mode files is very slow

#### Reproduction guide :beetle:
- Start Emacs
- SPC f r
- select a recent file and press enter

*Observed behaviour:* :eyes: :broken_heart:
It takes like 2-3 seconds for the file to open

*Expected behaviour:* :heart: :smile:
It should open faster

#### System Info :computer:
- OS: darwin
- Emacs: 29.1.90
- Spacemacs: 0.999.0
- Spacemacs branch: develop (rev. 554d5cd16)
- Graphic display: t
- Running in daemon: nil
- Distribution: spacemacs
- Editing style: vim
- Completion: ivy
- Layers:
```elisp
(csv graphviz lua sql rust yaml python markdown docker
     (go :variables go-tab-width 4)
     (typescript :variables typescript-indent-level 2)
     (html :variables web-mode-markup-indent-offset 2 web-mode-code-indent-offset 2 web-mode-css-indent-offset 2 web-mode-sql-indent-offset 2)
     ess terraform auto-completion emacs-lisp git ivy multiple-cursors tree-sitter treemacs)
```
- System configuration features: ACL GLIB GMP GNUTLS JPEG JSON LIBXML2 MODULES NOTIFY KQUEUE NS PDUMPER PNG RSVG SQLITE3 THREADS TIFF TOOLKIT_SCROLL_BARS TREE_SITTER ZLIB

#### Profiler report
```
  1015  85% - command-execute
  1015  85%  - call-interactively
  1015  85%   - funcall-interactively
  1009  84%    - spacemacs-buffer/return
  1009  84%     - if
  1009  84%      - widget-button-press
  1009  84%       - widget-apply-action
  1009  84%        - widget-apply
  1009  84%         - #<lambda -0x18c06ecd5c426d65>
  1009  84%          - find-file-existing
  1009  84%           - find-file
  1009  84%            - find-file-noselect
  1009  84%             - apply
  1009  84%              - editorconfig--advice-find-file-noselect
  1008  84%               - apply
  1008  84%                - #<compiled -0x25ea928a4de5da0>
  1005  84%                 - find-file-noselect-1
  1003  84%                  - after-find-file
    920  77%                   - normal-mode
    920  77%                    - set-auto-mode
    919  77%                     - set-auto-mode--apply-alist
    918  77%                      - set-auto-mode-0
    892  74%                       - python-mode
[...]

PR #16742, opened 2024-12-20
New macro `spacemacs|deferred-run-in-buffer` to deferred actions

sunlin7:
A new macro `spacemacs|deferred-run-in-buffer` to defer loading the `dap-python`, it speedup the python buffer for users, partly improve on python layer user experience, like #16529.

On my local, the time reduced from ~5.1 seconds to ~3.9 seconds by running follow testing commands line and check the time in `*Messages*` buffer.
```
$ emacs -nw --eval "(require 'benchmark)" --eval '(message "time: %s" (benchmark-elapse (with-current-buffer (find-file "/tmp/a.py") (kill-buffer))))'
```
Before: time: `5.086223264`
After: time: `3.932631126`

### DIFF
--- a/core/core-funcs.el
+++ b/core/core-funcs.el
@@ -28,6 +28,28 @@
 values."
   `(if (boundp ',variable) ,variable ',default))
 
+(defmacro spacemacs|deferred-run-in-buffer (name &rest body)
+  "Entry point that defers server startup until buffer is visible.
+It will wait until the buffer is visible before executing the instructions."
+  (let ((func (intern (format "spacemacs//init-if-buffer-visible-%s" name))))
+    `(defun ,func ()
+       "Run the commands for the current buffer if the buffer is visible.
+Returns non nil if it was run for the buffer."
+       (when (or (buffer-modified-p) (get-buffer-window nil t))
+         (remove-hook 'window-configuration-change-hook
+                      ,func t)
+         ,@body
+         t))
+    `(run-with-idle-timer
+      0 nil
+      (lambda (buffer)
+        (when (buffer-live-p buffer)
+          (with-current-buffer buffer
+            (unless (funcall ,func)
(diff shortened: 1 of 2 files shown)


bcc32:
Did you test that dap-python actually still gets loaded?  From a glance at the code I don't think it works correctly.  (Your idle timer is starting when the macro expansion happens, which seems very wrong.)

FWIW, is this worth the complexity here?  It only affects the first Python file loaded in a session, right?  Can you think of a less hacky way to do this?  (E.g., maybe we could defer loading `dap-python` until the LSP is actually connected, or something like that?)

sunlin7:
@bcc32  Thanks, I correct the code, and verified on my local.

This PR try to introduce this new macro to defer heavy loading works, the `dap-python` maybe has other way to do that, but for whole Spacemacs, user should have a chance to defer heavy works, and this macro provides the way.

bcc32:
Here's my concerns and why I still don't think we should merge this:

With the use-case (and benchmarking code) you presented, it is error-prone.  Now, opening a Python buffer as part of some other Lisp function will no longer reliably have dap-python loaded---only sometimes.  Similar things like enabling minor modes or whatever, which I imagine you were envisioning as potential use-cases for this macro, would be even worse because the minor mode would not be enabled by the time that (for example) `find-file-noselect` returns.  This sounds like it would be extremely confusing to debug---tying buffer/mode initialization to window display is not a good idea in general.

Also, I'm not convinced the user experience for this use case is really much better.  With this PR, Python buffers would appear to open slightly faster, but then more-or-less immediately start doing some more work that would block user interaction anyway.

Happy for some others to weigh in and see if they disagree and/or have good use-cases for such a helper in mind.

sunlin7:
Sure, got your points. Thanks.

PR #16749, opened 2024-12-23
* core/core-dotspacemacs.el: Mark variable dotspacemacs-command-key obsoleted

sunlin7:
The `dotspacemacs-command-key` was renamed to `dotspacemacs-command-key` on Jan 9, 2016, commit:
58e5241c8d96da0a2cf04d4897a74c9d1250875e .

It should be safe to remove the alias and use the official name only.

### DIFF
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -342,12 +342,14 @@ pressing `<leader> m`. Set it to `nil` to disable it."
   'string
   'spacemacs-dotspacemacs-init)
 
-(spacemacs|defc dotspacemacs-command-key "SPC"
+(spacemacs|defc dotspacemacs-emacs-command-key "SPC"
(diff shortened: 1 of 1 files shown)


bcc32:
The old name was not made obsolete when the alias was added, so nothing would have automatically warned users of the old variable name.  It's not safe to remove one of the names immediately; it would need to be marked obsolete first (e.g., using `define-obsolete-variable-alias`).

sunlin7:
@bcc32  Sure, I marked it obsoleted now, please review the latest one, thank you.

bcc32:
Thanks.  I removed the docstring because `define-obsolete-variable-alias` already generates one which contains the same information.

bcc32 (approved):

sunlin7:
Sure, it's more elegant, thanks!

PR #16752, opened 2024-12-24
* layers/+lang/python/*: New flag python-enable-importmagic

sunlin7:
Spacemacs slow on opening a single *.py file.

I figured out the top performance killer for python layer are `importmagic` and `pyvenv`.
The `importmagic` will start a background server to process the requests from importmagic, it took ~0.5 seconds on my local; but acctually I rarely using the `importmagic` feature. 

This PR introduce a toggle flag `python-enable-importmagic` to give chance to disable this heavy feature.

PS: For `pyvenv` performance issue (leads Spacemacs slow when opening a *.py), I had an upstream PR, but it seems the project stop update for 3 years.
https://github.com/jorgenschaefer/pyvenv/pull/128

### DIFF
--- a/layers/+lang/python/README.org
+++ b/layers/+lang/python/README.org
@@ -315,7 +315,24 @@ config section:
 or as a directory-local variable (for per-project settings).
 
 ** Importmagic
-Install importmagic and epc for importmagic functionality.
+Install importmagic and epc for importmagic functionality, it will start a epc
+instance on background to serve the requests.
+
+The importmagic is a heavy feature, will take time to prepare the background
+server, especially slow with =ipython= or =ipython3=, set the
+=importmagic-python-interpreter= to =python= or =python3= (if it is the
+interpreter of ipython) will have performance benefit. Fox example:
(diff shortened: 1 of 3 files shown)


bcc32 (commented):
By the way, thank you for all of your recent investigation into spacemacs's performance!

Do you think we should maybe leave this on by default?  We should weigh the cost of confusing users with silent behavior changes.  Certainly if we change the default behavior I think we should make that clear in the PR title, description, and commit messages.

sunlin7:
Sure, I use Spacemacs for daily works, and really suffered by the slow performance for just opening a single *.py file :)

I had change the default value to `t` to not make a surprise for the users who're using the importmagic feature; and added a section on the document for users how to get performance benefit when they using the importmagic feature.

smile13241324 (approved):

smile13241324:
Thanks for all the work guys :). Given that importmagic and epc are depreciated for a long time, and are really costly background processes I would disable them by default.

sunlin7:
Thank you so much!

PR #16754, opened 2024-12-26
Fix the layers dependencies which required a lsp backend

sunlin7:
Fix the layers dependency which required a lsp backend.
These layers configuration indicate user want lsp feature, so the layer should depends on the lsp.

### DIFF
--- a/layers/+lang/go/layers.el
+++ b/layers/+lang/go/layers.el
@@ -23,4 +23,4 @@
 
 (when (and (boundp 'go-backend)
            (eq go-backend 'lsp))
-  (configuration-layer/declare-layer-dependencies '(dap)))
+  (configuration-layer/declare-layer-dependencies '(lsp)))
--- a/layers/+lang/python/layers.el
+++ b/layers/+lang/python/layers.el
@@ -23,4 +23,4 @@
 
 (when (and (boundp 'python-backend)
            (eq python-backend 'lsp))
-  (configuration-layer/declare-layer-dependencies '(dap)))
+  (configuration-layer/declare-layer-dependencies '(lsp)))
--- a/layers/+lang/ruby/layers.el
+++ b/layers/+lang/ruby/layers.el
@@ -23,7 +23,7 @@
 
 (when (and (boundp 'ruby-backend)
            (eq ruby-backend 'lsp))
-  (configuration-layer/declare-layer-dependencies '(dap)))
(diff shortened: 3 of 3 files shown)


bcc32:
The `dap` layer depends on `lsp` already.  If I understand correctly, users who do use `dap` features would see breaking behavior after this change.

sunlin7:
@bcc32 Hidden logic here for the user who want `lsp` backend but the layer depends on `dap`, just because of `dap` layer had requested the `lsp` layer.  

This patch corrects the dependencies relations.  (the `dap` depends on `lsp` from its implementation, but the `lsp` should not depends on `dap` for the `dap` will affect the user experience a lot, the `dap` is heavy and under developing).

The `dap` layer should be explicitly configured in dot spacemacs file if user really use it.

sunlin7:
It's a bug, not a feature, for the layers who actually should depend on the lsp but currently depends on the dap works for user.

bcc32:
> It's a bug, not a feature, for the layers who actually should depend on the lsp but currently depends on the dap works for user.

No, this is a deliberate feature.  I suppose in theory it could specify both `dap` and `lsp` as dependencies, but it is intentional that enabling `go` layer with `go-backend` set to `lsp` will implicitly activate `dap layer.  See [this commit](https://github.com/syl20bnr/spacemacs/commit/9bd9151c7aed5628af7c7df93dab3e2c262596a3) for an example of the historical context.

> dap will affect the user experience a lot, the dap is heavy and under developing).

If you want a way to opt out of `dap` while still using lsp for `go-backend`, maybe we should just add a new config variable instead of piggy-backing on `go-backend`.

My objection is only that we should try to stay backwards compatible with existing user's configs, and the proposed change is currently incompatible.

sunlin7:
> If you want a way to opt out of dap while still using lsp for go-backend, maybe we should just add a new config variable instead of piggy-backing on go-backend.

That's what I want, I used the lsp daily, but never use the dap, and the dap make Spacemacs slow on opening a simple *.py or *.c file.

bcc32:
Makes sense, let's add a new variable then.

smile13241324 (approved):
I think you are right @sunlin7 dap should be opt in just as syntax checking and auto complete

sunlin7:
@smile13241324  Thank you so much, it will save the user who only use the `lsp` but never use the `dap`. 

And I will work on the document to make sure user who want tasting the `dap` feature can get start with it. Thanks

PR #16755, opened 2024-12-26
* core/core-display-init.el: Bump graphic frame initilization to after-make-frame-functions

sunlin7:
Fix a bug that Spacemacs won't init the font for graphic frame (X11 frame), reproducing steps:
1. emacs -nw  # start Spacemacs on terminal
2. `M-x make-frame-on-display` and input `:0` (if you have :0 display, or input other display).

The graphic frame from Spacemacs won't apply the font from Spacemacs.

This issue caused by Spacemacs only hook the `server-create-window-system-frame`.

The patch will do works on `after-make-frame-functions`, works for both emacs daemon and previous `make-frame-on-display` command.

### DIFF
--- a/core/core-display-init.el
+++ b/core/core-display-init.el
@@ -36,15 +36,17 @@
         ;; fallback to normal loading behavior only if in a GUI
         (t (display-graphic-p))))
 
-(define-advice server-create-window-system-frame
-    (:after (&rest _) spacemacs-init-display)
-  "After Emacs server creates a frame, run functions queued in
+(defun spacemacs/init-window-frame (frame)
+  "After Emacs creates a window frame, run functions queued in
 `SPACEMACS--AFTER-DISPLAY-SYSTEM-INIT-LIST' to do any setup that needs to have
 the display system initialized."
   (when (spacemacs--display-system-initialized-p)
-    (mapc #'funcall (reverse spacemacs--after-display-system-init-list))
-    (advice-remove 'server-create-window-system-frame
-                   #'server-create-window-system-frame@spacemacs-init-display)))
+    (with-demoted-errors "Spacemacs init display error: %S"
+      (with-selected-frame frame
+        (mapc #'funcall (reverse spacemacs--after-display-system-init-list))))
+    (remove-hook 'after-make-frame-functions #'spacemacs/init-window-frame)))
+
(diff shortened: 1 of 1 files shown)


core/core-display-init.el
```diff
@@ -36,15 +36,17 @@
...
   (when (spacemacs--display-system-initialized-p)
-    (mapc #'funcall (reverse spacemacs--after-display-system-init-list))
-    (advice-remove 'server-create-window-system-frame
-                   #'server-create-window-system-frame@spacemacs-init-display)))
+    (with-selected-frame frame
+      (mapc #'funcall (reverse spacemacs--after-display-system-init-list)))
+    (remove-hook 'after-make-frame-functions #'spacemacs-init-window-frame)))
+
+(with-eval-after-load 'frame
+  (add-to-list 'after-make-frame-functions #'spacemacs-init-window-frame))
```
bcc32: I think you can just use `add-hook` here, and there's no need to `with-eval-after-load` (`frame` gets loaded extremely early IIRC)

core/core-display-init.el
```diff
@@ -36,15 +36,17 @@
...
+(defun spacemacs-init-window-frame (frame)
+  "After Emacs creates a window frame, run functions queued in
 `SPACEMACS--AFTER-DISPLAY-SYSTEM-INIT-LIST' to do any setup that needs to have
 the display system initialized."
   (when (spacemacs--display-system-initialized-p)
-    (mapc #'funcall (reverse spacemacs--after-display-system-init-list))
-    (advice-remove 'server-create-window-system-frame
-                   #'server-create-window-system-frame@spacemacs-init-display)))
+    (with-selected-frame frame
+      (mapc #'funcall (reverse spacemacs--after-display-system-init-list)))
```
bcc32: You may want to use something like `with-demoted-errors` here.  Erroring in `after-make-frame-functions` can cause confusing behavior, see #16643 for an example.

sunlin7:
@bcc32  Thank you for comments, and I had changed the code to follow you comments. Please review again. Thanks

bcc32:
Hope you don't mind, I pushed a couple of follow-up fixes to this file also.

bcc32 (approved):

sunlin7:
I'm okay with the changes, thank you.

PR #16756, opened 2024-12-27
[python] Fix toggle breakpoint behavior when invoked twice in a row

sunlin7:
Currently pressing `, d b` can toggle debug line on a *.py file, but the cursor will be moved, so user have to move cursor back to the target line then to do "Toggle" again.

This PR will keep cursor on the same line, so user can press `, d b` to do a really "Toggle" operation (eg: press `, d b` twice).


layers/+lang/python/funcs.el
```diff
@@ -205,11 +205,11 @@ ROOT-DIR should be the path for the environemnt, `nil' for clean up"
...
-      (progn
-        (back-to-indentation)
-        (insert trace)
-        (insert "\n")
-        (python-indent-line)))))
+      (back-to-indentation)
+      (insert trace ?\n)
+      (python-indent-line)
+      (forward-line -1))
+    (mwim-beginning-of-code-or-line)))
```
bcc32: This command is part of the `mwim` package, not enabled by default.

bcc32 (changes requested):
It seems weird to leave point on the debugging line rather than the line where the user originally invoked `,db`.  Perhaps we should just make `,db` also attempt to search the line before the cursor for the previously-inserted debug statement, and delete it if found?

sunlin7:
@bcc32  I agree the behavior on this patch is not perfect.
But if we search backward and remove previous line, it still not like a "Toggle" operation. 
Only keep the cursor on same line can act like a "Toggle".
And I removed the `mwim-beginning-of-code-or-line`.

bcc32:
I was thinking we could have the command check the current line *and* the previous line.  Pushed a commit to show what I mean, let me know if you think that makes sense.

sunlin7:
@bcc32 It's good for me with the behavior to try check previous line. Thanks

bcc32:
Thanks for the PR!

PR #16760, opened 2024-12-27
[github-copilot] Fix activation for holy-mode

fnussbaum:
Completions were disabled by default due to `holy-mode` permanently activating `evil-emacs-state`. See for example https://github.com/copilot-emacs/copilot.el/pull/124.


fnussbaum:
On second thought, it might be better to suggest enabling copilot in `evil-emacs-state` (or even for all evil states with non-nil `:input-method` property) upstream first. I originally rejected this idea, and am still not sure that enabling it there is necessarily the right thing to do.

EDIT: I think copilot.el is not the right place for this integration, so I will just merge it here.

PR #16762, opened 2024-12-27
[spacemacs-defaults] Tweak savehist configuration

### LINKED ISSUE #1300: Spacemacs freezes from time to time. When this happens it eats up almost all memory and CPU.
OS X 10.10 user here. I've only used Spacemacs for 3 days but it has frozen multiple times, even in one day :( ... By "freeze" I mean it just doesn't respond to anything at all, be it mouse click or keyboard input. I had to `sudo kill -9` to kill the process and restart it, losing unsaved changes in the meantime. Is there any way to get an error log or something such that I might look at it/attach it here so that it would be easier to figure out what happened? I figure it could have something to do with "persistent server" but I'm not sure.

### LINKED ISSUE #1369: Spacemacs gets stuck for a few seconds and hogs CPU(100% usage) from time to time

### LINKED ISSUE #9409: emacs eats cpu 100% frequently
#### Description :octocat:
emacs hoge cpu 100% frequently even It is idle.

- start emacs
- open any file

emacs got freeze， can not input or move。Both happened on terminal emacs and graphic.

(%LAST_KEYS%)
%SYSTEM_INFO%
- OS: gnu/linux
- Emacs: 25.2.1
- Spacemacs: 0.200.9

#### Backtrace :paw_prints:
```
- timer-event-handler                                            8765  73%
 - apply                                                         8761  73%
  + savehist-autosave                                            7504  63%
  + linum-update-current                                         1153   9%
  + which-key--update                                              60   0%
  + #<compiled 0x285199>                                           36   0%
  + auto-revert-buffers                                             5   0%
  + show-paren-function                                             2   0%
  + which-func-update                                               1   0%
   cancel-timer-internal                                            4   0%
+ redisplay_internal (C function)                                1548  13%
+ ...                                                             894   7%
+ command-execute                                                 625   5%
+ which-key--hide-popup                                            15   0%
+ evil--jump-hook                                                  11   0%
+ evil-repeat-pre-hook                                             10   0%
+ evil-repeat-post-hook                                             7   0%
+ hl-paren-initiate-highlight                                       7   0%
+ winner-save-old-configurations                                    4   0%
  company-post-command                                              4   0%
  internal-timer-start-idle                                         3   0%
+ yas--post-command-handler                                         3   0%
```
And I have repeated using the profiler to track the cpu usage, the main reason leading this is savehist-autosave.

fnussbaum:
1. Use an idle timer instead of a strict periodic autosave interval. Saving large histories can cause noticable delays, and the default `history-length` set by Spacemacs is relatively large (1000).

2. Save `kill-ring` without text properties, as they are known to cause significant performance problems and huge history files in some cases, and persisting them in the kill-ring usually brings little benefit. This should fix #1300, fix #1369 and fix #9409 again.

3. Add `kmacro-ring` and `log-edit-comment-ring` as additional variables. They are useful and usually not very large. BTW, `search-ring`, `regexp-search-ring` and `extended-command-history` are redundant here as long as long as `savehist-save-minibuffer-history` is non-nil (the default). I decided to leave them there for now to avoid any breaking changes for users that customized that option, though I guess it could be okay to change this.

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/config.el
+++ b/layers/+spacemacs/spacemacs-defaults/config.el
@@ -196,6 +196,13 @@ or `nil' to only save and not visit the file."
 ;; Session
 ;; ---------------------------------------------------------------------------
 
+(spacemacs|defc spacemacs-savehist-autosave-idle-interval 60
+  "Idle interval between autosaves of minibuffer histories and other
+variables (see `savehist-mode' and `savehist-additional-variables')."
+  'integer)
+
+(defvar spacemacs--savehist-idle-timer nil)
+
 ;; scratch buffer empty
 (setq initial-scratch-message dotspacemacs-initial-scratch-message)
 ;; don't create backup~ files
--- a/layers/+spacemacs/spacemacs-defaults/packages.el
+++ b/layers/+spacemacs/spacemacs-defaults/packages.el
@@ -413,9 +413,30 @@
           savehist-additional-variables '(search-ring
                                           regexp-search-ring
                                           extended-command-history
-                                          kill-ring)
-          savehist-autosave-interval 60)
-    (savehist-mode t)))
+                                          kill-ring
+                                          kmacro-ring
+                                          log-edit-comment-ring)
+          ;; We use an idle timer instead, as saving can cause
+          ;; noticable delays with large histories.
+          savehist-autosave-interval nil)
+    (savehist-mode t)
+    (define-advice savehist-save
+        (:around (orig &rest args) spacemacs//kill-ring-no-properties)
+      "Text properties can blow up the savehist file and lead to
+excessive pauses when saving."
+      (if (memq 'kill-ring savehist-additional-variables)
+          (let ((kill-ring (mapcar #'substring-no-properties
+                                   (cl-remove-if-not #'stringp kill-ring))))
+            (apply orig args))
+        (apply orig args)))
+    (when (and (boundp 'spacemacs--savehist-idle-timer)
+               (timerp spacemacs--savehist-idle-timer))
+      (cancel-timer spacemacs--savehist-idle-timer))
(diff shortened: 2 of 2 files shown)


fnussbaum:
> 1. Use an idle timer instead of a strict periodic autosave interval. Saving large histories can cause noticable delays, and the default `history-length` set by Spacemacs is relatively large (1000).

I think an idle timer is a reasonable choice here, note that `savehist-autosave` is also called in the `kill-emacs-hook`. So the only case where this change would cause more savehist data to be lost than before is if users work for a long time without ever pausing for a minute, and then Emacs crashes in a way that does not trigger `kill-emacs-hook` (which should be very rare). But let me know if you object to this.

smile13241324 (approved):
This looks very good

PR #16765, opened 2024-12-28
Remove temp files /tmp/fineDiff* for kill-emacs without quit ediff

sunlin7:
Fix a bug that the temp files `/tmp/fineDiff*` were left after Spacemacs ediff templates.

Reproducing steps: Pressing `SPC f e D` and press `n` to locate first diff line, then press `SPC f q` to quick Spacemacs directly. 

Then the `/tmp/fineDiff*` files left.

### DIFF
--- a/core/core-versions.el
+++ b/core/core-versions.el
@@ -27,4 +27,15 @@
 (defconst spacemacs-version          "0.999.0" "Spacemacs version.")
 (defconst spacemacs-emacs-min-version   "28.2" "Minimal version of Emacs.")
 
+(defmacro spacemacs|eval-until-emacs-min-version (version msg &rest body)
+  "Evaluate the BODY if `spacemacs-emacs-min-version' < VERSION, otherwise
+warn the MSG."
+  (declare (indent 1))
+  `(if (version< spacemacs-emacs-min-version ,version)
+       (progn ,@body)
+     (apply 'warn
+            (if ,msg '("%s" ,msg)
+              '("Minimum emacs version %s is newer than supported version %s"
+                ,version spacemacs-emacs-min-version)))))
+
 (provide 'core-versions)
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -951,6 +951,17 @@ variable."
   (when (fboundp 'outline-show-all)
     (outline-show-all)))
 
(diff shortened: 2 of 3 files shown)


bcc32:
For context, here is the [upstream bug](https://lists.gnu.org/archive/html/bug-gnu-emacs/2024-12/msg00869.html) which adds the ediff function mentioned here.

bcc32:
IMO that's a pretty trivial bug, and I don't think "upstream Emacs will make this change eventually" is a generally a good reason on its own to backport a similar patch to Spacemacs---it's not even related to Spacemacs functionality, only ediff itself.  I would vote to just close this, but will wait for others to chime in.

sunlin7:
I found the /tmp/fineDiff* files after Spacemacs `SPC f e D` function, and got the root cause, pushed the patch to emacs upstream.

So I hope Spacemacs user won't be bothered by using the `SPC f e D` function, that's why I pushed the solution here.

bcc32:
That doesn't really address my root concern: I don't think backporting random things to Spacemacs is worth the maintenance burden, especially when the impact is so low.  (I can't imagine the existence of some files in `/tmp` is really bothering anybody, especially since they get cleaned up if you properly exit the ediff session).

Can explain why these files not getting cleaned up is actually causing you problems?

sunlin7:
When using Spacemacs on a never-rebooted linux server with shared `/tmp/` folder, other users on the server can see the `fineDiff*` files were created again and again, they will find the creator and kindly reminder me with mail/chat that maybe some leak happened.

bcc32:
Okay so that's pretty specific to your site and not exactly catastrophic.  Why not just fix it in your config, or use site-start?

sunlin7:
This issue happened on using the Spacemacs `SPC f e D` function, I do fixed it on my local about a month ago. And hope the solution can help the Spacemacs users who still with emacs-28/29/30.

smile13241324 (approved):
As much as I dislike fixing emacs bugs in Spacemacs we have done this for a lot of broken packages in the past. Given how long it will take until a new emacs version is broadly accepted I think it makes sense to shield our users from those issues.

We can do cleanup once we increase the min emacs Version again.

smile13241324:
@sunlin7 can you check the conflicts on your branch? Once its ready please ping me and I will check and merge it.

sunlin7:
@smile13241324  Sure, had resolved the conflicts. Please help review again. Thank you !

smile13241324 (approved):

PR #16766, opened 2024-12-28
* layers/+spacemacs/spacemacs-defaults/funcs.el: Enhance copying file path for dired-mode

sunlin7:
Currently pressing `SPC f y y` on a `dired-mode` buffer will get an error.

This PR enhances the users experience by support copying file path on a `dired-mode` buffer (`SPC f y y` in a dired buffer).

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -861,9 +861,15 @@ ones created by `magit' and `dired'."
     (user-error "Current buffer is not visiting a file or directory")))
 
 (defun spacemacs/copy-file-path ()
-  "Copy and show the file path of the current buffer."
+  "Copy and show the file path of the current buffer. If the buffer is a
+dired buffer, the result will be the file path under cursor, otherwise
+the result is directory path."
   (interactive)
-  (if-let* ((file-path (spacemacs--file-path)))
+  (if-let* ((file-path (or (spacemacs--file-path)
+                           (and (derived-mode-p 'dired-mode)
+                                (condition-case nil
+                                    (dired-get-file-for-visit)
+                                  (error default-directory))))))
       (progn
         (kill-new file-path)
         (message "%s" file-path))
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -861,15 +861,15 @@ ones created by `magit' and `dired'."
     (user-error "Current buffer is not visiting a file or directory")))
 
 (defun spacemacs/copy-file-path ()
-  "Copy and show the file path of the current buffer. If the buffer is a
-dired buffer, the result will be the file path under cursor, otherwise
-the result is directory path."
+  "Copy and show the file path of the current buffer.
+
+In Dired, the result will be the file path under cursor if any,
+otherwise the listed directory's path."
   (interactive)
   (if-let* ((file-path (or (spacemacs--file-path)
                            (and (derived-mode-p 'dired-mode)
-                                (condition-case nil
-                                    (dired-get-file-for-visit)
-                                  (error default-directory))))))
+                                (dired-get-filename nil t))
+                           (spacemacs--directory-path))))
       (progn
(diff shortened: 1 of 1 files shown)


layers/+spacemacs/spacemacs-defaults/funcs.el
```diff
@@ -863,7 +863,11 @@ ones created by `magit' and `dired'."
 (defun spacemacs/copy-file-path ()
   "Copy and show the file path of the current buffer."
```
bcc32: Please update the docstring to reflect the change in behavior.

layers/+spacemacs/spacemacs-defaults/funcs.el
```diff
@@ -863,7 +863,11 @@ ones created by `magit' and `dired'."
...
   (interactive)
-  (if-let* ((file-path (spacemacs--file-path)))
+  (if-let* ((file-path (or (spacemacs--file-path)
+                           (and (eq major-mode 'dired-mode)
```
bcc32: Prefer to use `derived-mode-p` over `(eq major-mode ...)`

layers/+spacemacs/spacemacs-defaults/funcs.el
```diff
@@ -863,7 +863,11 @@ ones created by `magit' and `dired'."
...
+                                (condition-case nil
+                                    (dired-get-file-for-visit)
```
bcc32: If in a dired buffer and `dired-get-file-for-visit`, it should probably be an error rather than silently doing a different behavior, IMO.  Users can already press `SPC f y d` to get the current directory.

bcc32 (changes requested):

layers/+spacemacs/spacemacs-defaults/funcs.el
sunlin7: updated

layers/+spacemacs/spacemacs-defaults/funcs.el
sunlin7: Changed.

layers/+spacemacs/spacemacs-defaults/funcs.el
sunlin7: This PR wants the function be more "smart", so try best to copy the the path as much as possible.

sunlin7: On a dired buffer, function `copy-file-path` to copy dir path still makes sense for user.

bcc32: "Smart" can be confusing instead.  I would find `SPC f y y` *sometimes* copying a file and *sometimes* copying a directory to be a very unintuitive behavior.

sunlin7:
@bcc32  Thanks for the comments, new changes pushed, please help review again.

layers/+spacemacs/spacemacs-defaults/funcs.el
sunlin7: An emacs user try `find-file` with a dir "~/", or `(find-file "~/")`, then emacs open a dired buffer, that imply emacs treats a directory like a file; so user try a copy file path function on a dired buffer, copying  the dir path to user still makes sense.

sunlin7:
An emacs user try find-file with a dir "~/", or (find-file "~/")`, then emacs open a dired buffer, that imply a directory is a specific file in emacs; so user try a copy file path function on a dired buffer, copying the dir path to user still makes sense.

layers/+spacemacs/spacemacs-defaults/funcs.el
bcc32: Fair enough, I think you've convinced me.

bcc32:
Pushed some small tweaks, to (1) avoid raising a `user-error` and then immediately catching it and (2) use logic consistent with `SPC f y d` to determine what directory path to copy.  Happy to merge once confirmed these work for you.

sunlin7:
@bcc32  Thank you for the improvement, and I had tried the last one on my local, they work perfect ! Thanks

bcc32 (approved):

PR #16770, opened 2024-12-30
Add evil-fringe-mark to +spacemacs/spacemacs-evil

phillip-simons:
This PR adds the [evil-fringe-mark](https://github.com/Andrew-William-Smith/evil-fringe-mark) package.

I configured it to display both user and special marks globally.

Thanks

### DIFF
--- a/layers/+spacemacs/spacemacs-evil/packages.el
+++ b/layers/+spacemacs/spacemacs-evil/packages.el
@@ -30,6 +30,7 @@
         (evil-escape :location (recipe :fetcher github
                                        :repo "smile13241324/evil-escape"))
         evil-exchange
+        evil-fringe-mark
         evil-goggles
         evil-iedit-state
         evil-indent-plus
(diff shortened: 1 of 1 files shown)


sunlin7:
Have you try it with `holy` mode? And how about its performance on a large file with several marks? I remember this package have some issues on these two questions. correct me if I was wrong. Thank you.

layers/+spacemacs/spacemacs-evil/packages.el
```diff
@@ -155,6 +156,12 @@
       (define-key evil-visual-state-map evil-exchange-cancel-key
                   'evil-exchange-cancel))))
 
+(defun spacemacs-evil/init-evil-fringe-mark ()
+  (use-package evil-fringe-mark
+    :config
+    (setq-default evil-fringe-mark-show-special t)
```
fnussbaum, line 162: I think it would be better to set options in the `:init` or `:custom` block. It makes them easier to override by users when the package is deferred.

layers/+spacemacs/spacemacs-evil/packages.el
```diff
@@ -155,6 +156,12 @@
...
+    (global-evil-fringe-mark-mode)))
```
fnussbaum, line 163: If we add this package, it should be deferred, and the mode should not be enabled by default. It would make sense to add a toggle as for `evil-visual-mark-mode`, <kbd>SPC t '</kbd> seems logical as a key binding.

smile13241324 (changes requested):
Not a bad idea though I think it makes sense to have this configurable in the layer and to have a toggle for this feature as well.

Given the performance concerns from @sunlin7 the package should be disabled by default. Also please only install the package when it is used.

smile13241324:
Looks like evil fridge mode had not received any commit in the last 6 years. Given thst the package seems dead I am hesitant to add it to Spacemacs.

PR #16772, opened 2024-12-31
* layers/+spacemacs/spacemacs-defaults/packages.el: Avoid loading outline-mode

sunlin7:
Hi @fnussbaum I noticed two lines 
https://github.com/syl20bnr/spacemacs/blob/2d43f15e0c06fbd325fd802d428fe2244f1ed6da/layers/%2Bspacemacs/spacemacs-defaults/packages.el#L212-L213

can be improved to no-requiring the `outline` package for better performance. 

Please help review the changes, thank you!

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/packages.el
+++ b/layers/+spacemacs/spacemacs-defaults/packages.el
@@ -209,8 +209,7 @@
      ediff-merge-split-window-function 'split-window-horizontally)
     :config
     ;; show org ediffs unfolded
-    (require 'outline)
-    (add-hook 'ediff-prepare-buffer-hook #'outline-show-all)
+    (add-hook 'ediff-prepare-buffer-hook 'spacemacs/ediff-buffer-outline-show-all)
     ;; restore window layout when done
     (add-hook 'ediff-quit-hook #'winner-undo)))
 
--- a/layers/+spacemacs/spacemacs-defaults/packages.el
+++ b/layers/+spacemacs/spacemacs-defaults/packages.el
@@ -209,7 +209,7 @@
      ediff-merge-split-window-function 'split-window-horizontally)
     :config
     ;; show org ediffs unfolded
-    (add-hook 'ediff-prepare-buffer-hook 'spacemacs/ediff-buffer-outline-show-all)
+    (add-hook 'ediff-prepare-buffer-hook 'spacemacs//ediff-buffer-outline-show-all)
     ;; restore window layout when done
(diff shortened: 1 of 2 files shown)


layers/+spacemacs/spacemacs-defaults/packages.el
```diff
@@ -209,8 +209,10 @@
      ediff-merge-split-window-function 'split-window-horizontally)
     :config
     ;; show org ediffs unfolded
-    (require 'outline)
-    (add-hook 'ediff-prepare-buffer-hook #'outline-show-all)
+    (add-hook 'ediff-prepare-buffer-hook
+              #'(lambda ()
```
bcc32: Generally it is better to define named functions that you add to hooks, rather than adding lambdas.  The former are much easier for users to remove in their configs, if desired.

layers/+spacemacs/spacemacs-defaults/packages.el
```diff
@@ -209,8 +209,10 @@
...
+                  (when (featurep 'outline)
```
bcc32: Using `fboundp` instead of `featurep` would prevent the byte-compiler from complaining about `outline-show-all` possibly not being defined.

bcc32 (commented):
Thank!  Just two small nits here.

layers/+spacemacs/spacemacs-defaults/packages.el
sunlin7: Sure, I had changed to the `fboundp` on the latest changes. Thank you !

layers/+spacemacs/spacemacs-defaults/packages.el
sunlin7: This is really minor function, user can immediately know what will be done here; as a comparison, if there was a named function, the function should be in the `funcs.el`, then user have to jump from current `packages.el` to the `funcs.el` to see what had been done in the function.   That's why lambda function is used for these two lines.

fnussbaum: I would still prefer a named function here: Jumping to definitions or using `describe-function` is quick and easy, while removing anonymous functions from hooks is not as straight-forward. While the latter might not be needed as often, I still think we should support this in all cases.

sunlin7: Sure, I had pushed a named function `spacemacs/ediff-buffer-outline-show-all`. Please help review again, thank you.

fnussbaum (approved):

PR #16783, opened 2025-01-08
* core/core-display-init.el: Fix detection graphic init on Windows

sunlin7:
Fix the graphic initialization detection on the MS-Window, the font-family-list will works on a selected frame.

### DIFF
--- a/core/core-display-init.el
+++ b/core/core-display-init.el
@@ -24,13 +24,7 @@
 (defvar spacemacs--after-display-system-init-list '()
   "List of functions to be run after the display system is initialized.")
 
-(defun spacemacs--display-system-initialized-p (&optional frame)
-  "Return non-nil if the display system has been initialized."
-  (cond ((boundp 'ns-initialized) ns-initialized)
-        ((boundp 'x-initialized) x-initialized)
-        ((boundp 'pgtk-initialized) pgtk-initialized)
-        ;; fallback to normal loading behavior only if in a GUI
-        (t (display-graphic-p frame))))
+(defvar spacemacs--display-system-initialized-p nil)
 
 (defun spacemacs//init-window-frame (frame)
   "After Emacs creates a window frame FRAME, run enqueued functions.
@@ -40,7 +34,8 @@ Functions are called with FRAME selected.
 Queued functions are added to
 `spacemacs--after-display-system-init-list' and are run once,
 only after the display system has been initialized."
-  (when (spacemacs--display-system-initialized-p frame)
+  (when (display-graphic-p frame)
+    (setq spacemacs--display-system-initialized-p t)
     (dolist (f (reverse spacemacs--after-display-system-init-list))
       (with-demoted-errors "spacemacs|do-after-display-system-init: %S"
         (with-selected-frame frame
@@ -53,9 +48,12 @@ only after the display system has been initialized."
   "Call FUNC with no arguments once the display system is initialized.
 
 See `spacemacs|do-after-display-system-init'."
-  (if (not (spacemacs--display-system-initialized-p))
-      (push func spacemacs--after-display-system-init-list)
-    (funcall func)))
+  (if spacemacs--display-system-initialized-p
+      (funcall func)
+    (push func spacemacs--after-display-system-init-list)
+    ;; `spacemacs--display-system-initialized-p' may be nil even if the initial
+    ;; frame is graphical.
+    (spacemacs//init-window-frame (selected-frame))))
 
 (defmacro spacemacs|do-after-display-system-init (&rest body)
   "If the display system is initialized, run BODY.
--- a/core/core-display-init.el
+++ b/core/core-display-init.el
@@ -24,18 +24,13 @@
 (defvar spacemacs--after-display-system-init-list '()
   "List of functions to be run after the display system is initialized.")
 
-(defun spacemacs--display-system-initialized-p ()
+(defun spacemacs--display-system-initialized-p (&optional frame)
   "Return non-nil if the display system has been initialized."
   (cond ((boundp 'ns-initialized) ns-initialized)
-        ;; w32-initialized gets set too early, so
-        ;; if we're on Windows, check the list of fonts
-        ;; instead (this is nil until the graphics system
-        ;; is initialized)
-        ((boundp 'w32-initialized) (font-family-list))
         ((boundp 'x-initialized) x-initialized)
         ((boundp 'pgtk-initialized) pgtk-initialized)
         ;; fallback to normal loading behavior only if in a GUI
-        (t (display-graphic-p))))
+        (t (display-graphic-p frame))))
 
 (defun spacemacs//init-window-frame (frame)
   "After Emacs creates a window frame FRAME, run enqueued functions.
@@ -45,7 +40,7 @@ Functions are called with FRAME selected.
 Queued functions are added to
 `spacemacs--after-display-system-init-list' and are run once,
 only after the display system has been initialized."
-  (when (spacemacs--display-system-initialized-p)
+  (when (spacemacs--display-system-initialized-p frame)
     (dolist (f (reverse spacemacs--after-display-system-init-list))
       (with-demoted-errors "spacemacs|do-after-display-system-init: %S"
         (with-selected-frame frame
--- a/core/core-display-init.el
+++ b/core/core-display-init.el
@@ -49,13 +49,21 @@ only after the display system has been initialized."
 
 (add-hook 'after-make-frame-functions #'spacemacs//init-window-frame)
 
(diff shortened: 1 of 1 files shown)


core/core-display-init.el
```diff
@@ -45,7 +48,7 @@ Functions are called with FRAME selected.
 Queued functions are added to
 `spacemacs--after-display-system-init-list' and are run once,
 only after the display system has been initialized."
-  (when (spacemacs--display-system-initialized-p)
+  (when (spacemacs--display-system-initialized-p frame)
```
fnussbaum: I'm wondering whether `spacemacs--display-system-initialized-p` is necessary here. Could it not just be replaced with `(display-graphic-p frame)`? This is just a guess, I have not tested it yet.

sunlin7: No, I tried the `emacsclient -c -a=` on windows, the `display-graphic-p` returns nil in there.

The reason is when the emacs daemon creating first graphic frame, but does not selected the frame until line 1015 in follow link; before the graphic frame been selected, the default frame is not a graphic frame (it's a terminal) for emacs daemon. Here are the function lines:
https://github.com/emacsmirror/emacs/blob/b71225ed90ab28ad88c22f8fcd1b87c5681eb0a7/lisp/server.el#L994-L1019

That's why this PR uses the `with-selected-frame` with the frame to get correct result.

sunlin7: To make it clear, I copy the lines here, when emacs daemon create first graphic frame, the `make-frame` is executed and the `spacemacs--display-system-initialized-p` is called in its tail, at that time, the frame is NOT selected, and the default frame is a inner frame (terminal frame) of emacs daemon, so `display-graphic-p` will return nil for the default frame. 

```
(defun server--create-frame (nowait proc parameters)
...
  (let ((frame (make-frame....)))
...
    (select-frame frame)
...
))
```

So the PR selects the new created frame to get correct result.

bcc32: Just to be clear, @fnussbaum said:

```
(display-graphic-p frame)
```

not

```
(display-graphic-p)
```

Are you saying the former returns nil on Windows with `emacsclient -c -a=`?  I'd be surprised to find out that's true.

sunlin7: `(display-graphic-p frame)` works! I had pushed the latest, verified both normal startup and daemon startup.

> Are you saying the former returns nil on Windows with emacsclient -c -a=? I'd be surprised to find out that's true.

The change is from the #16755 that we moved the action from `:after server-create-window-system-frame` to `after-make-frame-functions`, then there is no `(select-frame frame)` until `(server--create-frame)` selects the frame.

fnussbaum: Thanks for testing this! Sorry for the confusion, what I actually meant though, is to replace the condition in the function `spacemacs//init-window-frame`
```elisp
(when (spacemacs--display-system-initialized-p frame) ...)
```
 in line 43 by 
```elisp
(when (display-graphic-p frame) ...)
```
I think that should also work and would slightly simplify the code.

sunlin7: @fnussbaum Thanks for the comment, but I don't have the NS environment to verify replacing the `spacemacs--display-system-initialized-p` with the `display-graphic-p`.

bcc32: I would vote to replace `spacemacs--display-system-initialized-p` with `display-graphic-p`.  It is simpler, almost certainly equivalent, and also matches the docstring of `spacemacs|do-after-display-system-init` better.

I did some investigating:

`ns` appears to refer to macOS.  On my MacBook, removing this line from the `cond`:

```
((boundp 'ns-initialized) ns-initialized)
```

still behaves correctly (e.g., unicode-fonts is initialized only upon the creation of the first graphical frame).

I believe `x-initialized` and `pgtk-initialized` ought to behave the same way, and `display-graphic-p` should be sufficient here.

I took a look at `make-frame`'s code, which calls `window-system-initialization` (which in turn, sets variables like `ns-initialized`).  It does this before creating the frame, so there's no chance that `(display-graphic-p frame)` could succeed too early.

sunlin7: But if we replace line 43 `spacemacs--display-system-initialized-p` with `(display-graphic-p frame)`, 
but the line 62|57 still keep using `spacemacs--display-system-initialized-p`, will lead different detecting there.

bcc32: Agreed.  I pushed a couple more commits to flesh out the overall refactor idea.  Hopefully both of you think it's an improvement.

sunlin7: @bcc32  I had verified your changes and confirm it works on both Linux/Windows for Spacemacs daemon/normal startup. Thanks!

bcc32:
This pull request now LGTM.  Will leave it to @fnussbaum to double check and merge.  Thanks!

core/core-display-init.el
fnussbaum: @bcc32 Thanks, this looks very good! I just have one question about the docstring of `spacemacs|do-after-display-system-init`: I think "BODY is evaluated with a graphical frame selected" is only guaranteed before the window system is initalized. The docstring seems ambiguous in that regard, or am I missing something?

bcc32: In the deferred case, `spacemacs//init-window-frame` uses `with-selected-frame` with a frame that just passed the `(display-graphic-p frame)` test, before calling `f`.  It uses `with-selected-frame` inside `dolist` so that a passed-in function that might accidentally select a different frame won't affect subsequent functions.  I think that is sufficient to guarantee what is promised by the docstring.

smile13241324 (approved):
Looks really good @sunlin7 I am merging it tight away

sunlin7:
@smile13241324  Thank you, appreciate it!

core/core-display-init.el
fnussbaum: I agree with the deferred case, though I did not understand from the docstring that its last sentence also only refers to the deferred case.

I was thinking of something like the following:
1. Start Spacemacs in GUI and execute `M-x server-start`.
2. Open a terminal frame. Then any evaluation of `(spacemacs|do-after-display-system-init BODY)` will immediately evaluate BODY with the terminal frame selected. For example
``` shell
emacsclient -nw -e '(spacemacs|do-after-display-system-init (message "(display-graphic-p) is %s" (display-graphic-p)))'
``` 
will print "(display-graphic-p) is nil".

sunlin7: @fnussbaum  The behavior is  correct. The function will detecting the current emacs instance is graphic ready or not for first time. So, any of the graphic frame ready, that mean current emacs is graphic ready. So the code should be run immediately.  

It's NOT related to individual frame.  

You can check the original function, it was similar.

bcc32: Ah, I agree that the docstring is not correct in the immediate case (I misunderstood your question).

I have fixed the docstring in 1e83f30fedb9dbfe845bd3b44b04798a592edfe1.  Fortunately, this PR does not break previous behavior, as the macro never guaranteed that the selected frame was graphical.

Since you mention that though, is there a reason to even bother with `with-selected-frame` in `spacemacs//init-window-frame`?  Perhaps it can safely be simplified?

fnussbaum: > Since you mention that though, is there a reason to even bother with with-selected-frame in spacemacs//init-window-frame? Perhaps it can safely be simplified?

This sounds reasonable. There seems to be at least one call site of `spacemacs|do-after-display-system-init` that needs to be fixed though: In `emoji/init-company-emoji` the selected frame is explicitly used in the macro body. But this just seems like an error anyway, it should probably be evaluated for all new (graphical?) frames. 

Also, I don't quite understand `spacemacs//load-theme-internal` (another call site) at first glance, I would need to look into this.

bcc32: `set-fontset-font`'s docstring says that the FRAME argument isn't used if the `fontset` argument is non-nil (which it is, in `spacemacs//set-emoji-font`).  So, I think it doesn't actually use the frame at all.

sunlin7: @fnussbaum  
> Also, I don't quite understand spacemacs//load-theme-internal (another call site) at first glance, I would need to look into this.

I found the `eval` statement in `spacemacs//load-theme-internal` was first introduced in this commit:
https://github.com/syl20bnr/spacemacs/commit/c913ec89f86f21b6473b8dbd2f1f4d82d30f6e36

The `eval` is necessary?

bcc32: I think it's not necessary now that we're using lexical-binding in that file and also for the macro definition.

fnussbaum: > set-fontset-font's docstring says that the FRAME argument isn't used if the fontset argument is non-nil (which it is, in spacemacs//set-emoji-font). So, I think it doesn't actually use the frame at all.

@bcc32 This makes a lot more sense, thanks for catching that.

PR #16785, opened 2025-01-10
[defaults] Improve `SPC w m`, persist `quit-restore` parameter

fnussbaum:
This is a follow-up to #16533, and in particular addresses the caveat mentioned there.

The built-in tab-bar.el similarly keeps a writable window state as well as a window configuration for each tab.

Making the `quit-restore` window parameter persistent allows quitting help windows correctly after maximizing/minimizing or switching tab-bar tabs, for example (without using popwin).

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -219,24 +219,36 @@ persistent which-key) are kept or minimized too."
          (window-already-maximal
           (eq max-target-window (selected-window))))
     (cond ((and window-already-maximal
-                (window-parameter nil 'spacemacs-max-state))
+                (window-parameter nil 'spacemacs-max-state-writable))
            ;; Restore the previously deleted windows, keeping the state of the
            ;; selected window.
            (let ((selected-win-state (window-state-get (selected-window))))
              (window-state-put
-              (window-parameter nil 'spacemacs-max-state) (selected-window))
+              ;; Prefer non-writable window states during a session,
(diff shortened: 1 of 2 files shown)


layers/+spacemacs/spacemacs-defaults/funcs.el
```diff
@@ -219,24 +219,36 @@ persistent which-key) are kept or minimized too."
...
-           (set-window-parameter nil 'spacemacs-max-state (window-state-get max-target-window t))
+           ;; Clean up ...
+           (walk-windows
+            (lambda (win)
+              (set-window-parameter win 'spacemacs-max-state nil)
+              (set-window-parameter win 'spacemacs-max-state-writable nil)))
+           ;; ... and store the current state as a window parameter of the selected window
+           ;; before deleting other windows.
+           (set-window-parameter nil 'spacemacs-max-state (window-state-get max-target-window))
+           (set-window-parameter nil 'spacemacs-max-state-writable (window-state-get max-target-window t))
```
bcc32, line 251: Might be nicer to pass `'writable` rather than `t` for the WRITABLE argument, for readability.  Everything else looks reasonable, but I'm not super familiar with this functionality.

smile13241324 (approved):
Looks good

layers/+spacemacs/spacemacs-defaults/funcs.el
fnussbaum: Makes sense, thanks! I'll keep this in mind for the future, not sure if it is worth it to push now as I can't squash it anymore.

PR #16795, opened 2025-01-13
[shell] Fix shell-pop renaming the wrong buffer for eat

### LINKED ISSUE #16794: spacemacs/shell-pop-eat renames the current buffer to `*eat-1*`
#### Description :octocat:

spacemacs/shell-pop-eat renames the current buffer to `*eat-1*`

- Enable shell layer
- Start Emacs
- Press `SPC a t s a` in some random buffer, e.g., `foo.txt`

`foo.txt` buffer gets renamed to `*eat-1*`, and the shell buffer is called `*eat*`.

`foo.txt` buffer does not get renamed to `*eat-1*`, and the shell buffer is called `*eat-1*`.

#### System Info :computer:
- OS: darwin
- Emacs: 29.4
- Spacemacs: 0.999.0
- Spacemacs branch: bcc32 (rev. ab95572cc)
- Graphic display: t
- Running in daemon: nil
- Editing style: hybrid
- Completion: compleseus
- Layers:
```elisp
(spell-checking syntax-checking compleseus mu4e ibuffer
                (org :variables org-enable-appear-support t org-enable-github-support t org-enable-modern-support t org-enable-notifications t org-project-capture-projects-file "TODO.org" org-start-notification-daemon-on-startup nil org-want-todo-bindings t)
                pdf
                (unicode-fonts :variables unicode-fonts-enable-ligatures t)
                eglot autohotkey csv emacs-lisp json markdown
                (ocaml :variables ocaml-format-on-save t)
                (rust :variables rust-backend nil)
                (scheme :variables scheme-implementations
                        '(chez chicken guile))
                (shell-scripts :variables shell-scripts-backend nil shell-scripts-format-on-save t)
                sql
                (nixos :variables nixos-format-on-save t)
                osx
                (git :variables git-enable-magit-todos-plugin t)
                (version-control :variables version-control-diff-tool nil)
                spacemacs-completion spacemacs-editing spacemacs-editing-visual spacemacs-evil spacemacs-language spacemacs-misc spacemacs-modeline spacemacs-navigation spacemacs-org spacemacs-project spacemacs-visual finance
                (shell :variables shell-default-shell 'shell shell-enable-smart-eshell t)
                (transmission :variables transmission-auto-refresh-all t)
                (bcc32 :variables bcc32-enable-auto-revert-debugging-at-startup nil bcc32-enable-explain-pause-at-startup nil)
                bcc32-ledger bcc32-org)
```
- System configuration features: ACL GLIB GMP GNUTLS JPEG JSON LIBXML2 MODULES NOTIFY KQUEUE NS PDUMPER PNG RSVG SQLITE3 THREADS TIFF TOOLKIT_SCROLL_BARS TREE_SITTER ZLIB

bcc32:
Fix #16794.

### DIFF
--- a/layers/+tools/shell/funcs.el
+++ b/layers/+tools/shell/funcs.el
@@ -120,6 +120,17 @@ SHELL is the SHELL function to use (i.e. when FUNC represents a terminal)."
        (shell-pop index)
        (spacemacs/resize-shell-to-desired-width))))
 
+(defun spacemacs//eat-for-shell-pop (&rest args)
+  "Like `eat', but make the new shell buffer current.
+
+Used to satisfy `shell-pop's assumptions."
+  ;; `eat' unexpectedly selects the window displaying the returned buffer, but
+  ;; doesn't actually leave the buffer current when it returns.  The fix is
+  ;; suggested upstream at https://codeberg.org/akib/emacs-eat/pulls/193, but
+  ;; meanwhile we work around it in Spacemacs.  We can replace this function at
+  ;; its callsite with just `eat' if/when the above is merged.
+  (set-buffer (apply #'eat args)))
+
 (defun spacemacs//toggle-shell-auto-completion-based-on-path ()
   "Deactivates automatic completion on remote paths.
 Retrieving completions for Eshell blocks Emacs. Over remote
--- a/layers/+tools/shell/packages.el
+++ b/layers/+tools/shell/packages.el
@@ -345,7 +345,7 @@
     :defer t
     :commands (eat eat-other-window eat-project eat-project-other-window)
     :init
-    (make-shell-pop-command "eat" eat)
+    (make-shell-pop-command "eat" spacemacs//eat-for-shell-pop)
     (spacemacs/set-leader-keys "atsa" 'spacemacs/shell-pop-eat)
     (spacemacs/register-repl 'eat 'eat)
     :config


bcc32:
`eat` uses `pop-to-buffer-same-window` to display the newly-created eat-mode buffer after the command terminates, but does not leave the buffer current at the end.  `shell-pop` assumes that the function passed in leaves the new buffer current, so we have to reconcile these by just using `set-buffer`.

Perhaps it would be better to advocate for `eat`, `vterm`, etc. support upstream in shell-pop?

fnussbaum:
> eat uses pop-to-buffer-same-window to display the newly-created eat-mode buffer after the command terminates, but does not leave the buffer current at the end.

This could also be considered a bug in eat, see https://debbugs.gnu.org/cgi/bugreport.cgi?bug=69259#31. Perhaps it should also be fixed there? I think it might still be good to merge this in the meantime.

> Perhaps it would be better to advocate for eat, vterm, etc. support upstream in shell-pop?

I think this is a good idea. (Additionally, it would be nice for shell-pop to switch to `display-buffer` as suggested in https://github.com/kyagi/shell-pop-el/issues/61, but this is not a tiny change.)

fnussbaum:
> This could also be considered a bug in eat, see https://debbugs.gnu.org/cgi/bugreport.cgi?bug=69259#31. 

The elisp manual at least discourages this:

"When an editing command returns to the editor command loop, Emacs
automatically calls ‘set-buffer’ on the buffer shown in the selected
window (*note Selecting Windows::).  This is to prevent confusion: it
ensures that the buffer that the cursor is in, when Emacs reads a
command, is the buffer to which that command applies (*note Command
Loop::).  Thus, you should not use ‘set-buffer’ to switch visibly to a
different buffer; for that, use the functions described in *note
Switching Buffers::.

   When writing a Lisp function, do _not_ rely on this behavior of the
command loop to restore the current buffer after an operation.  Editing
commands can also be called as Lisp functions by other programs, not
just from the command loop; it is convenient for the caller if the
subroutine does not change which buffer is current (unless, of course,
that is the subroutine’s purpose)." (Elisp manual, chapter 28.2 Current Buffer)

bcc32:
Good idea, fixing upstream in eat at https://codeberg.org/akib/emacs-eat/pulls/193.

fnussbaum (approved):

ryanprior:
@bcc32 thank you! I've been trying on and off to fix this since I introduced `eat` to the shell layer but hadn't nailed down what the problem was yet.

PR #16798, opened 2025-01-15
[markdown] Add key bindings for edit code block

sunlin7:
Hi,

This PR will make the `markdown` layer support key bindings `, '` and `, ,` as same as the `org` layer behavior.

The `org` layer has the shortcut `, '` to edit the code block, and `, ,` to finish editing code block. Here are the `org` layer key bindings, edit a code block:
https://github.com/syl20bnr/spacemacs/blob/e0c05e02f295a4124009e2d70d19312eb33edb4e/layers/%2Bemacs/org/packages.el#L245-L246

Commit/Abort the edit:
https://github.com/syl20bnr/spacemacs/blob/e0c05e02f295a4124009e2d70d19312eb33edb4e/layers/%2Bemacs/org/packages.el#L192-L197

### DIFF
--- a/layers/+lang/markdown/packages.el
+++ b/layers/+lang/markdown/packages.el
@@ -25,6 +25,7 @@
   '(
     company
     company-emoji
+    edit-indirect
     emoji-cheat-sheet-plus
     gh-md
     markdown-mode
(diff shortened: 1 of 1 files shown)


bcc32:
The `edit-indirect` package isn't mentioned by any of the Spacemacs layers yet, I think this won't work unless the user happens to add it to `dotspacemacs-additional-packages`?  Perhaps we should explicitly add it to markdown layer?

sunlin7:
@bcc32 You're right! When executing `markdown-edit-code-block` in a markdown file, it will prompt installing the package `edit-indirect`, it works in current session; And the package  `edit-indirect` will be **delete**  on next startup. 

I had pushed new changes to include the package. Please help review again. Thanks

bcc32:
Thanks!  Merging with a very minor fix.

sunlin7:
Thank you !

PR #16800, opened 2025-01-17
* layers/+emacs/org/packages.el: Fix org-appear init

radrow:
- `org-appear`'s `manual` trigger was not configured for `hybrid` editing style. Note that `insert` mode only applies to `evil` style --- `hybrid` style uses `hybrid` mode as its "insert". This PR fixes it.
- Hooks were assigned to `org-mode`, while they should be to `org-appear-mode`. In my case this caused necessity to reload `org-mode` once a buffer booted up. With this change the hooks are now properly set.

### DIFF
--- a/CHANGELOG.develop
+++ b/CHANGELOG.develop
@@ -2980,7 +2980,7 @@ files (thanks to Daniel Nicolai)
   flag (thanks to Keith Pinson).
 - When =org-appear= is set to be triggered manually and Spacemacs' editing mode is
   Vim or hybrid, register relevant =org-appear= commands in Evil hooks for Org
-  buffers (thanks to Keith Pinson).
+  buffers (thanks to Keith Pinson and Radosław Rowicki).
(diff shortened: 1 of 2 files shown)


bcc32:
>org-appear's manual trigger was not configured for hybrid editing style. Note that insert mode only applies to evil style --- hybrid style uses hybrid mode as its "insert". This PR fixes it.

Maybe it would be better to fix this in hybrid-mode itself, so that it would apply to any code which uses these hooks.

I think `evil-hybrid-state-{entry,exit}-hook` should probably just be an alias for `evil-insert-state-{entry,exit}-hook`, in the same way that `hybrid-mode` redefines `evil-insert-state-p` to also check for hybrid-state?

radrow:
Personally I would find that more intuitive. I just thought it was by design.

smile13241324 (approved):

Kazark (commented):
:+1:

PR #16802, opened 2025-01-23
compleseus: advise vertico-posframe--show to match current frame font setting

downward-funarg:
Prior to this change, vertico-posframe does not register changes in the current frame font (for example font size change via `SPC z f`. This commit advises vertico-posframe--show to set the posframe font to the current frame setting.

### DIFF
--- a/layers/+completion/compleseus/packages.el
+++ b/layers/+completion/compleseus/packages.el
@@ -468,14 +468,23 @@
   (use-package vertico-posframe
     :after vertico
     :init
-    (setq vertico-posframe-poshandler 'posframe-poshandler-frame-center)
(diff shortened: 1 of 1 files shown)


bcc32:
Seems useful for all users of vertico-posframe, not just Spacemacs.  Have you considered reporting it to https://github.com/tumashu/vertico-posframe/issues instead?  The docstring for `vertico-posframe-font` says it already falls back to the selected frame's font.

downward-funarg:
Thanks, I opened an issue [here](https://github.com/tumashu/vertico-posframe/issues/56).

smile13241324:
Opened PR in related emacs package

PR #16804, opened 2025-01-25
* core/core-configuration-layer.el: Fix bug on configuration-layer/describe-package

sunlin7:
Fix a bug when revert the buffer on `configuration-layer/describe-package` buffer.
Reproduce steps:
0. Start Spacemacs, 
1. Pressing `SPC h d P` and input a package name to show a help buffer for a package;
2. Pressing `SPC b R` to revert the buffer, will trigger an error, like
```text
Debugger entered--Lisp error: (void-function nil)
  nil()
  apply(nil nil)
  help-mode-revert-buffer(t nil)
  revert-buffer(t)
  funcall-interactively(revert-buffer t)
  call-interactively(revert-buffer nil nil)
  command-execute(revert-buffer)
```

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -966,6 +966,9 @@ a new object."
   (interactive
    (list (intern
           (completing-read "Package: " configuration-layer--used-packages))))
+  (help-setup-xref (list #'configuration-layer/describe-package
+                         pkg-symbol layer-list pkg-list)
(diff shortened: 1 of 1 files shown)


bcc32 (approved):

bcc32:
Nice, thanks.  I reworded the commit to be slightly more specific about the bug, but otherwise merged as-is.

sunlin7:
Thank you !

PR #16812, opened 2025-02-03
[doc] Fix broken links

EasternLi:


bcc32:
These pages just got moved to the same path but under https://www.spacemacs.org, could we just update the URLs to point to the new location instead?

EasternLi:
> These pages just got moved to the same path but under https://www.spacemacs.org, could we just update the URLs to point to the new location instead?

done (but use https://spacemacs.org for consistency with others)

I mistakenly thought that `layers/LAYERS.org` was rendered at the url

bcc32:
Thank you!

PR #16814, opened 2025-02-04
*layers/+spacemacs/spacemacs-editing/packages.el: Enhance the dired sorting

sunlin7:
Hi,

There is a file sorting bug when browsering a dir, reproducing steps with `dotspacemacs-distribution 'spacemacs` (not spacemacs-base):
1. `$ touch /tmp/a; sleep 1; touch /tmp/b` # create two files in shell, and b is newer than a
2. `$ emacs -nw --eval '(custom-set-variables (quote (dired-listing-switches "-alt")))' /tmp/` # start emacs and show /tmp/ in dired
3. press `s`, hydra-dired-quick-sort menu will popup, we remember the files orders in dired buffer (the `dired-actual-switches` is `-al --sort=extension` now)
4. press `t` on the hydra menu, the files in dired has no changes (the `dired-actual-switches` is `-al -t` now)

With this patch, behavior follow previous steps will be more naturally on step 3 and step 4.

### DIFF
--- a/layers/+spacemacs/spacemacs-editing/packages.el
+++ b/layers/+spacemacs/spacemacs-editing/packages.el
@@ -114,7 +114,7 @@
     (define-advice dired-noselect (:before (&rest _) quick-sort-setup)
       (let ((dired-quick-sort-suppress-setup-warning 'message))
         (dired-quick-sort-setup))
-      (advice-remove 'dired-noselect 'quick-sort-setup@dired-noselect))
+      (advice-remove 'dired-noselect 'dired-noselect@quick-sort-setup))
     :config
     (evil-define-key 'normal dired-mode-map "s" 'hydra-dired-quick-sort/body)))
 
--- a/layers/+spacemacs/spacemacs-editing/packages.el
+++ b/layers/+spacemacs/spacemacs-editing/packages.el
@@ -111,21 +111,12 @@
   (use-package dired-quick-sort
     :defer t
     :init
-    (spacemacs|add-transient-hook dired-mode-hook
-      (lambda ()
-        (let ((dired-quick-sort-suppress-setup-warning 'message))
-          (dired-quick-sort-setup))))
+    (define-advice dired-noselect (:before (&rest _) quick-sort-setup)
+      (let ((dired-quick-sort-suppress-setup-warning 'message))
+        (dired-quick-sort-setup))
+      (advice-remove 'dired-noselect 'quick-sort-setup@dired-noselect))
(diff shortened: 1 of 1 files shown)


bcc32:
I don't think what you're describing sounds like a bug, it sounds like the expected behavior of dired-quick-sort.

Can you please specify what order the files *actually* appear in after each of steps (2), (3), and (4), and what order you *expected* them to appear?

sunlin7:
@bcc32 
step 2: dired shows files in time ordered  -- OK
step 3: hydra menu was displayed, in the menu it shows files sort by "version" -- but files in dired buffer actually sorted by time.
step 4: after press t, hydra menu shows sort by time, files are ordered by time (results same as step 3).

So the key point is step 3, files ordered not follow the hydra menu.

bcc32:
Could you follow up on the upstream package issue?  It sounds like there is a suggested workaround there.

sunlin7:
@bcc32 No, there is no workaround on upstream. I reported the issue to upstream 3 years ago. The workaround on upstream ticket actually is linked to Spacemacs patch.

bcc32:
dired-quick-sort-setup should be run before dired-mode-hook, actually, not afterwards.  dired-quick-sort-setup adds something to dired-mode-hook which won't take effect if dired-quick-sort-setup is part of dired-mode-hook, and so it is behaving incorrectly on first invocation.

The intended behavior of dired-quick-sort is to _never respect_ dired-listing-switches's sorting behavior, because dired-quick-sort expects to set the sorting order itself and persist it through savehist.

I think what we should do instead is to remove the advice and call dired-quick-sort-setup immediately after dired is loaded rather than in dired-mode-hook.

sunlin7:
@bcc32  Code updated based on your comments. 
the `dired.el` is required int the `window-purpose-x.el` on Spacemacs startup, so the code uses advice with `dired-noselect` to avoid loading dired-quick-sort on startup (I had tried the `dired-load-hook` or `(with-eval-after-load 'dired...` , it will load the dired-quick-sort on Spacemacs startup, slows down the Spacemacs).

bcc32:
Pushed a small fix.  Can you please confirm this works for you?  Otherwise LGTM and happy to merge.

bcc32 (approved):

sunlin7:
@bcc32  Thanks for the fixing, and I had confirmed it works on my local. Good!

bcc32:
Terrific.  Thank you for confirming, and for reworking the PR!

PR #16818, opened 2025-02-05
* layers/+lang/python: Migrate key binding for toggle debug to avoid conflicting

sunlin7:
Hi,

The python layer has conflicting on key binding `, d b` when enabling the `python` and `lsp(dap)` layers (the `dap` layer is enabled automatically when `lsp` layer is enabled).

The `dap` layer will bind the `, d b` for break point functions, 
https://github.com/syl20bnr/spacemacs/blob/d13f96767659900d097ac021cf01a97ff48a3f9a/layers/%2Btools/dap/packages.el#L66-L72

while the `python` layer will bind `, d b` to toggle breakpoint:
https://github.com/syl20bnr/spacemacs/blob/d13f96767659900d097ac021cf01a97ff48a3f9a/layers/%2Blang/python/packages.el#L390

The patch migrate the python layer toggle breadpoint to `, d t` to avoid the conflicting.


bcc32 (approved):

bcc32:
Thanks!

PR #16819, opened 2025-02-05
* layers/+lang/python/funcs.el: Rewrite spacemacs/pyenv-executable-find to support list

sunlin7:
Rewrite the function `spacemacs/pyenv-executable-find` for better performance.
Original function will invoke the '/bin/pyenv' twice to determine one command; then it has **24** invokes for the `/bin/pyenv` triggered by the function `spacemacs/python-toggle-breakpoint` on my local, with this patch the calls will reduce to **8**. 
Please help review the changes, thanks

### DIFF
--- a/layers/+lang/python/funcs.el
+++ b/layers/+lang/python/funcs.el
@@ -128,12 +128,16 @@
   (highlight-lines-matching-regexp "trepan.api.debug()"))
 
 (defun spacemacs/pyenv-executable-find (commands)
-  "Find executable taking pyenv shims into account, return the first
-occured executable path in the COMMANDS. If the pyenv was configured
-with \"system\" then the system exectutable will be included, otherwise
-the system executable will be ignored. A non-list COMMANDS is supported
-for compatibility reason."
-  (unless (listp commands)              ; to compatible
+  "Find executable taking pyenv shims into account.
+
+Return the first executable in COMMANDS whose path was found.  If
+the pyenv was configured with \"system\" then the system
+executable will be included, otherwise the system executable
+will be ignored.
+
+COMMANDS may also be a single string, for backwards
+compatibility."
+  (unless (listp commands)
     (setq commands (list commands)))
   (if (or (bound-and-true-p pyvenv-virtual-env) ; in virtualenv
           (not (executable-find "pyenv")))      ; or no pyenv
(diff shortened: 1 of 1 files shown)


layers/+lang/python/funcs.el
```diff
@@ -127,45 +127,38 @@
...
-              (while (and (not executable)
-                          (< i (length pyenv-version-names)))
-                (if (string-match (elt pyenv-version-names i) (string-trim pyenv-string))
-                    (setq executable (string-trim pyenv-string)))
-                (if (string-match (elt pyenv-version-names i) "system")
-                    (setq executable (string-trim (executable-find command))))
-                (setq i (1+ i))))
-          executable))
-    (executable-find command)))
+(defun spacemacs/pyenv-executable-find (commands)
```
bcc32, line 130: This function might be called in user configs.  It would be nicer if it continued to accept a single command and treated it as a singleton list, for backwards compatibility.

layers/+lang/python/funcs.el
```diff
@@ -127,45 +127,38 @@
...
+  "Find executable taking pyenv shims into account. If the pyenv was
+configured with \"system\" then the system exectutable will be included,
+otherwise the system executable will be ignored."
```
bcc32: Can you add documentation about how the function handles the commands argument?  Specifically, it's important to note that it will return some path corresponding to *some* command (the first one that appears in the list?)

layers/+lang/python/funcs.el
bcc32: It would also avoid some unnecessary changes in this same file.

layers/+lang/python/funcs.el
```diff
@@ -190,22 +181,18 @@ ROOT-DIR should be the path for the environemnt, `nil' for clean up"
...
-                     ((spacemacs/pyenv-executable-find "python3.9") "breakpoint()")
-                     ((spacemacs/pyenv-executable-find "python3.10") "breakpoint()")
-                     ((spacemacs/pyenv-executable-find "python3.11") "breakpoint()")
-                     (t "import pdb; pdb.set_trace()")))
-        (prev-line (save-excursion
-                     (and (zerop (forward-line -1))
-                          (thing-at-point 'line))))
-        (line (thing-at-point 'line)))
+  (let* ((exe (spacemacs/pyenv-executable-find '("trepan3k" "wdb" "ipdb3" "pudb3" "ipdb" "pudb" "python3")))
+         (trace (pcase (file-name-nondirectory (or exe ""))
```
bcc32: Nit: `(and exe (file-name-nondirectory exe))` would be more elegant here, IMO (avoid unnecessarily calling the function `file-name-nondirectory`).

bcc32 (changes requested):
Nice, thanks for this contribution.

sunlin7:
@bcc32  I had update the code based on your comments, please help review it again. Thanks.

bcc32:
Sorry for the delay.  Made some minor fixes, otherwise LGTM, so merging.

sunlin7:
All good for me, thank you !

PR #16822, opened 2025-02-11
[python] Fix toggle breakpoint on empty line

sunlin7:
Hi,
There is a bug on toggling debug point on *.py file, reproducing steps with `python` layer enabled:
1. $ emacs /tmp/non-exists.py
2. press `, d t` (`spacemacs/python-toggle-breakpoint`) to add the debugging line; then press `, d t` again, but it does not work.

This patch will make the `, d t` works as expected. Please help review the changes. Thanks.

### DIFF
--- a/layers/+lang/python/funcs.el
+++ b/layers/+lang/python/funcs.el
@@ -196,22 +196,23 @@ ROOT-DIR should be the path for the environemnt, `nil' for clean up"
                   ((or "ipdb" "ipdb3") "import ipdb; ipdb.set_trace()")
                   ((or "pudb" "pudb3") "import pudb; pudb.set_trace()")
                   ("python3"           "breakpoint()") ; not consider the python3.6 or lower
-                  (_ "import pdb; pdb.set_trace()")))
-         (prev-line (save-excursion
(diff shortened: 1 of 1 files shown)


bcc32:
I can't reproduce the bug you're describing, at commit 77d37e87da2b8eca05b8e598f5eab8ff2eb7c620.  Could you double check what version of the code you're running?

What value does `trace` and `exe` have when you encounter this bug?

sunlin7:
Hi @bcc32 
I tried with 77d37e8 and the issue is still reproducible. 
Just open an empty file like a.py, and try `, d t` to insert debugging line, then pressing `, d t` again try toggle back (do not move your cursor). The issue won't happened if you have an exists line.
It happened on emacs-28.3, emacs-29.4, emacs-31.0.50 (I didn't try it on emacs-30).

bcc32:
Ah, I see, this only happens when point is on an empty line, or in an empty file.  Updated the title.

bcc32 (approved):

bcc32:
LGTM, thanks!

sunlin7:
Thank you, appreciate!

PR #16829, opened 2025-02-19
Enable goto-address-mode for text-mode

xuhdev:
Currently it's only enabled in prog-mode, which has missed out text-mode.


smile13241324 (approved):

bb2020:
Can we disable this for `org-mode`? It derives from `text-mode` and conflicts with its own link highlighting.

smile13241324:
Sure we can add a line to the org mode hook

bcc32:
I think having this in text-mode-hook is likely too aggressive.  Lots of modes derive from text-mode and many of them may define their own link types (e.g., markdown-mode, various other types of markup/rich text, etc.).

@xuhdev did you want this for text-mode itself, or some other derived mode?  Maybe the hook should only be added for the specific mode you needed originally.

xuhdev:
@bcc32 If it's too aggressive, I don't mind following a mode-by-mode approach. I'm also interested in using it for plain text editing, but I don't know how we can do that while not being too aggressive.

(I'm also interested in i3wm-config-mode but it's not part of spacemacs anyway)

PR #16835, opened 2025-02-20
Remove seeing-is-believing package

cody-vibe:
Removing https://github.com/JoshCheek/seeing_is_believing from the `ruby` layer. The package seems unmaintained, with breakages on newer systems for the gem dependency.

In hindsight I should have not merged this in, it's not a core editor functionality but a flashy gimmick and trivial to add to a custom setup, you just put the package name in `dotspacemacs-additional-packages`.

### DIFF
--- a/layers/+lang/ruby/README.org
+++ b/layers/+lang/ruby/README.org
@@ -27,7 +27,6 @@
   - [[#toggles][Toggles]]
   - [[#rake][Rake]]
   - [[#refactor][Refactor]]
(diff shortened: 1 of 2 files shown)


cody-vibe:
Not sure what I should do about the Changelog for this. What is the expectation?

smile13241324 (approved):

smile13241324:
Just keep it as it is I think we will have a more transparent overview when we link the fixed issues with the corresponding release.

PR #16861, opened 2025-03-04
[themes-megapack] Remove majapahit-themes package

### LINKED ISSUE #16860: Error installing majapahit-themes
Spacemacs attempts and fails to install `majapahit-themes` on startup when using the `themes-megapack` layer. The reason is that this package was removed from Melpa due to missing requirements https://github.com/melpa/melpa/commit/29822ffa96a4414d2efe050d007c3fa9fcedd2f7

xu-shawn:
This removes the `majapahit-themes` package, which was removed from melpa https://github.com/melpa/melpa/commit/29822ffa96a4414d2efe050d007c3fa9fcedd2f7

Closes #16860


bcc32:
Thanks!

PR #16863, opened 2025-03-04
[go] just call generated hooks when setup company

EasternLi:
since the func is in `go-mode-local-vars-hook`.
Or rather, this restores the PROPS removed in #13455.
Fix yasnippet unavailable in first go-mode buffer.

### DIFF
--- a/layers/+lang/go/funcs.el
+++ b/layers/+lang/go/funcs.el
@@ -37,7 +37,9 @@
                 :call-hooks t))
     ('lsp (spacemacs|add-company-backends
             :backends company-capf
-            :modes go-mode))))
+            :modes go-mode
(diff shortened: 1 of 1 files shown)


EasternLi:
btw, this requires https://github.com/emacs-lsp/lsp-mode/pull/4725, otherwise yasnippet remains unenabled in any buffer (many modes with lsp).

smile13241324:
Hi  @EasternLi can you advice whether it makes sense to merge your PR without the upstream changes you have mentioned?

EasternLi:
> Hi @EasternLi can you advice whether it makes sense to merge your PR without the upstream changes you have mentioned?

I think you can. This does right and doesn't break new.

Issue #16824, opened 2025-02-15
Support using nerd-icons instead of all-the-icons in *Spacemacs* buffer when dotspacemacs-startup-buffer-show-icons


Pancf:
Hi, is it possible to replace all-the-icons with nerd-icons in dashboard?

bcc32:
By dashboard, do you mean the `*Spacemacs*` buffer or the mode line?  If the latter, what mode line theme are you using?

Pancf:
> By dashboard, do you mean the `*Spacemacs*` buffer or the mode line? If the latter, what mode line theme are you using?

I mean \*Spacemacs\*

Pancf:
Hi, I just try to make a PR to use nerd-icons in *Spacemacs* buffer, would you mind reviewing it and giving any comments?@smile13241324 @bcc32

smile13241324:
Tested and merged your PR, thanks for the help :)

PR #16865, opened 2025-03-06
Replace all-the-icons with nerd-icons in *Spacemacs* buffer.

Pancf:
- Replace all-the-icons with nerd-icons in *Spacemacs* buffer to make more compatible when running in shell.


Pancf:
#16824

smile13241324 (approved):
Looks good

winsphinx:
But, the home buffer disappers.
> progn: Cannot open load file: No such file or directory, nerd-icons

I installed nerd fonts, and can display correctly in dired.

Pancf:
> But, the home buffer disappers.
> 
> > progn: Cannot open load file: No such file or directory, nerd-icons
> 
> I installed nerd fonts, and can display correctly in dired.

You need to install nerd-icons package, not nerd-fonts.

Issue #16841, opened 2025-02-23
Python REPL for Conda environments is broken


TheDohn:
#### Description
Python REPL does not use Python installation as determined by Conda environment.

#### Reproduction guide
- Start Emacs
- Select and activate conda env (conda-env-activate)
- Start Python REPL (run spacemacs/python-start-or-switch-repl)

*Observed behaviour:*
- Python REPL starts successfully, but points to some sort of default Python installation, implying Conda shim has not been detected in PATH
- This can be seen by running `import sys; print(sys.executable)`, which does not print the Conda installation. More problematically, the libraries in the desired Conda env are not available. 

*Expected behaviour:*
- Python REPL starts successfully, and points to the correct Python installation, as determined by the conda env

#### System Info :computer:
- OS: darwin
- Emacs: 29.4
- Spacemacs: 0.999.0
- Spacemacs branch: develop (rev. 4216cbbf0)
- Graphic display: t
- Running in daemon: nil
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(toml ibuffer
      (lua :variables lua-backend 'lsp)
      octave sql html
      (latex :variables latex-backend 'lsp)
      csv vimscript javascript yaml debug
      (dap :variables dap-python-debugger 'debugpy)
      (auto-completion :variables auto-completion-return-key-behavior 'complete auto-completion-private-snippets-directory "/Users/donbunk/dotfiles/yasnippets" auto-completion-enable-snippets-in-popup t auto-completion-enable-sort-by-usage t auto-completion-enable-help-tooltip t auto-completion-use-company-box t auto-completion-tab-key-behavior 'cycle auto-completion-complete-with-key-sequence "jk" auto-completion-minimum-prefix-length 1 auto-completion-idle-delay 0.2)
      emacs-lisp
      (git :variables git-enable-magit-delta-plugin t git-enable-magit-gitflow-plugin t)
      helm
      (lsp :variables lsp-sonarlint t lsp-headerline-breadcrumb-enable nil lsp-lens-enable t)
      markdown multiple-cursors org
      (shell :variables shell-default-height 30 shell-default-position 'bottom)
      outshine spell-checking syntax-checking version-control treemacs
      (python :variables python-backend 'lsp python-spacemacs-indent-guess 'nil python-lsp-server 'pyright)
      (ess :variables ess-indent-level 5 ess-style 'RStudio ess-r-backend 'ess)
[...]

PR #16867, opened 2025-03-06
Integrate pet-mode and remove other venv management packages

### LINKED ISSUE #16855: using uv package / virtual environment manager with Python layer
`uv ` package is getting more and more popular as a package / virtual environment manager for python.

Not sure how it can be  configured with the python layer. There is no mention of `uv `  in the documentation. 

Would appreciate any help on this.

Raghu

smile13241324:
This PR tries to integrate [pet-mode](https://github.com/wyuenho/emacs-pet) to detect and setup python virtual environments.

I have tried to remove alternative venv management packages as well. However I am unsure if this has broken someones setup, so some testing is in order :).

Should fix #16855 and #16841.


smile13241324:
@sunlin7 @RaghuNaik @fnussbaum @bcc32 care to review?

bcc32:
My site definitely uses pyvenv directly (we have both Spacemacs and non-Spacemacs users), so I would be sad to see this suddenly removed without a good migration path.  Is pet-mode a drop-in replacement?  Can it be programmatically configured?

sunlin7:
@smile13241324  Thanks for the PR, I test it on my local and some feedback here:
1. the `pet-mode` dose not auto start with the python-base-mode, should move the `:config` to `:init` for pet package?
2. the `pet-mode` dose not select the `ipython3` if there are both `python3` and `ipython3` exists, it select `python3` (pet-mode issue)
3. I found a lot of `vfork` attemptions from `pet-mode` by linux strace tool. (pet-mode issue)

smile13241324:
Looks like this is not ready to be merged.
@bcc32 the idea of pet is that it detects active venv environments automatically and setups all packages accordingly. I have had hoped this would turn into a drop in replacement for our own venv detection and setup code. Would this replace venv's functionality for you? If not I can add them back and add pet mode behind a setting to make it opt in instead.

@sunlin7 thanks for testing, I will fix the setup and will have a look at the pet package to fix the bugs you have found.

I'll put this branch to draft state until it is in a better state.

bcc32:
@smile13241324 thanks for taking a look at this.  For context, my site uses pyvenv with a default, hard-coded environment for most users (distributed via NFS or similar), and also some users require the ability to manually set the environment to something specific depending on their team or deployment environment (however, we use a monorepo and therefore having the virtualenv path for all parts of the tree be determined by a single file in the project would not be viable).

I took a brief look at pet-mode and it seems it might be able to support this, but maybe not in an especially obvious way (it seems designed to be zero-configuration and doesn't have any interactive commands to support changing the virtual environment).

I will consult with some users when I'm back at work on Monday.  I am hopeful this will not be a problem, and in the "worst case" we can obviously just still use the `pyvenv` package and maintain our own key bindings internally if needed, so you don't need to consider this concern a hard blocker for the PR.

sthysel:
This branch seems to work fairly well for my use cases anyway. Are there any particular reason why it has not been merged into develop propper ?

smile13241324:
Hi guys,

I am back and this time for good :).

Lets talk about this specific feature I have started to integrate pet mode.
From the feedback I have seen it looks like it may not work well in all
current scenarios.

I think the current PR must be changed a bit to not negatively affect users
following a manual approach to virtual environments.

For now I am getting github copilot support running, this is getting it to work on the same level
as it is done in VSCode.

When this is done I would jump back into this topic and see if I can find a solution 
to have pet integrated without having to break the use cases of other users.

Is it ok when I reference you @sthysel for some testing runs? I think this would
speed up development tremendously.

Thanks a lot to all of you for keeping Spacemacs going I cannot wait to join in once again ^^.

sunlin7:
:+1:  So glad you're back now!

sthysel:
Very happy to do some testing. I've done some work in my own branch mostly around using ruff and basedpyright. As a rule I'd like to standarise on the astral stack with basedpyright predominately as this is the stack we are moving to locally. This seems to represent the golden path for python dev today.

smile13241324:
I have finished my work on this branch, can you two have a look and give me some feedback @sthysel @sunlin7?

I have adapted the readme and removed the outdated pyvenv mode. I have also made pet-mode correctly load on python mode starting.

I have also made a PR here https://github.com/wyuenho/emacs-pet/pull/56, but I am unsure if it will be merged. Until it is I have switched to a local fork.

sunlin7:
@smile13241324  thank you for the work.
I update to the latest changes, got stuck for a while and found 
1. It runs  `apt install sqlite3`  in the background 
2. It runs `apt install dasel` on background but got error message `E: Unable to locate package dasel`
It should be triggered by `:ensure-system-package (dasel sqlite3)`, and try install dasel every time.
And I'm not sure that will work on Windows system.

Besides that, it's good for me. Thank you.

smile13241324:
I see thats this ensure system package directive I am going to change it to something just testing for the package existence and mention tne requirements in the readme. Thanks for the review:)

sthysel:
I've been using this branch as my main working branch on Arch linux on uv and poetry managed projects, for the last week or so and have seen no issues. Seems to do the business, thanks @smile13241324

wyuenho:
I was just told that spacemacs has integrated pet, I'm glad that it's becoming more useful for a wider audience. I'm aware of wyuenho/emacs-pet#56. I have some changes related to automatically syncing variable dependencies in the pipeline. Basically, I want to support a use case where you can just change `python-shell-interpreter` and `python-shell-virtualenv-root` as if pet didn't exist, and then pet will automatically detect the change, find the absolute paths for them and then automatically reassign all the other emacs package variables that need to be in sync with them. This way you can set a .dirlocal on a directory specifying the name of the python executable and pet will automatically pick it up and DWIM.

layers/+lang/python/packages.el
```diff
@@ -63,6 +64,13 @@
     ;; packages for Microsoft's pyright language server
     (lsp-pyright :requires lsp-mode :toggle (eq python-lsp-server 'pyright))))
 
+(defun python/init-pet ()
+  (use-package pet
+    :hook (python-base-mode . pet-mode)
+    :if (and (executable-find "dasel")
+             (executable-find "sqlite3"))
```
wyuenho, line 71: sqlite3 is not required unless you are running emacs < 29 or hand compiled a emacs without it, AFAIK, most emacs distributions have it. You may want to do similar [feature detection](https://github.com/wyuenho/emacs-pet/blob/92da4fa6ffd97fadb24c6fc5d46976606a0264c4/pet.el#L472-L473) before making it a requirement.

layers/+lang/python/packages.el
```diff
@@ -226,6 +234,26 @@
...
+       (dolist (m spacemacs--python-pyenv-modes)
+         (add-hook (intern (format "%s-hook" m))
+                   'spacemacs//pyenv-mode-set-local-version)))
+      ('on-project-switch
+       (add-hook 'projectile-after-switch-project-hook
+                 'spacemacs//pyenv-mode-set-local-version)))
+    ;; setup shell correctly on environment switch
+    (spacemacs/set-leader-keys-for-major-mode 'python-mode
+      "vu" 'pyenv-mode-unset
+      "vs" 'pyenv-mode-set)))
```
wyuenho, line 255: Pet is meant to be a drop-in replacement for all the dozens of python virtualenv packages including pyenv-mode. Why would you want to keep pyenv-mode support? Pet already searches for the correct environment for you depending on where you are in your directory. A few days ago, I've also introduced [`pet-search-globally`](https://github.com/wyuenho/emacs-pet/commit/af40c8bcd3c78bacaa09bdb2ea835b9dea14557e), to allow the user to configure whether they want to search outside of the project for executable.

The problem with pyenv-mode is, when pyenv-mode-set is called, pythonic-activate is called and [globally set](https://github.com/pythonic-emacs/pythonic/blob/5daa1b9f589c2faf0ee8165dbb470a4dc752122f/pythonic.el#L293) `python-shell-virtualenv-root`, after this all the [variables](https://github.com/wyuenho/emacs-pet/blob/92da4fa6ffd97fadb24c6fc5d46976606a0264c4/pet.el#L1003-L1005) that depends on it will be out of sync.

layers/+lang/python/packages.el
```diff
@@ -28,6 +28,8 @@
     company
     cython-mode
     dap-mode
+    ;; We are using a fork until pet is prefering ipython as default shell (https://github.com/wyuenho/emacs-pet/pull/56)
+    (pet :location (recipe :fetcher github :repo "smile13241324/emacs-pet"))
```
wyuenho, line 32: Does Spacemacs always prefer ipython? Before this PR, did it also change all the package variables that need a path to a python executable, such as `lsp-pyright-python-executable-cmd`?

wyuenho:
I've just reviewed the PR. As a vanilla Emacs user, I'm not sure I understand why Spacemacs would still want to keep the pyenv-mode, pipenv and poetry packages after integrating with pet. Pet just detects all of them and set up the dependent variables in various emacs packages for python automatically for you. This works so well, even Microsoft has [copied my idea](https://github.com/microsoft/python-environment-tools).

If someone can illustrate all the different use cases to me, maybe delete some more code or I can move more of the logic into pet.

layers/+lang/python/packages.el
smile13241324: Good point :), I will have a look

sunlin7:
PET has issues:
[Not work on Ubuntu22.04 for there is no "python", but "python3" exists](https://github.com/wyuenho/emacs-pet/issues/59)
[searching multiple python interpreters by breadth first ](https://github.com/wyuenho/emacs-pet/issues/60)

sunlin7:
@bcc32 [commented on Mar 8](https://github.com/syl20bnr/spacemacs/pull/16867#issuecomment-2708425520)
> ... some users require the ability to manually set the environment to something specific depending on their team or deployment environment

I met similar scenario, I have a local python folder with `.dir-locals.el` to set python infer shell to specific one; but now the pet-mode will try the "python" and give global one but it's not expected.

layers/+lang/python/packages.el
```diff
@@ -149,7 +154,10 @@
   (add-hook `python-mode-hook `turn-on-evil-matchit-mode))
 
 (defun python/post-init-flycheck ()
-  (spacemacs/enable-flycheck 'python-mode))
+  (spacemacs/enable-flycheck 'python-mode)
+  ;; Setup flycheck but only after pet is loaded.
```
wyuenho, line 158: See https://github.com/syl20bnr/spacemacs/commit/8fc2b3b26be4012ebe97a0d33e8028da7b89ed54#r162800942

PR #16868, opened 2025-03-07
Fix `dotspacemacs-startup-buffer-show-icons` documentation in core-dotspacemacs.el

xu-shawn:
`all-the-icons` has been replaced by `nerd-icons` in the last commit


smile13241324 (changes requested):
You also need to change the template

smile13241324 (approved):

Issue #16346, opened 2024-04-01
configuration-layer/package-used-p is inconsistent with cfgl-package-used-p


bcc32:
As mentioned in the title, the two functions
`configuration-layer/package-used-p` and `cfgl-package-used-p` have
inconsistent behavior with respect to certain packages.

In particular, if a package is disabled through a `:toggle` keyword
whose form evaluates to nil, `configuration-layer/package-used-p` will
return t for that package while `cfgl-package-used-p` will return
nil.

`cfgl-package-used-p`'s behavior seems more correct to me---these
functions are used frequently in Spacemacs to figure out whether some
other layer is going to configure that package and cause it to, for
example, be responsible for implementing some key bindings.  In which
case, the `:toggle` evaluating to nil should cause the code to assume
that the package in question is *not used*, for this purpose.

In any case, it seems like `configuration-layer/package-used-p` should
be a very thin wrapper around `cfgl-package-used-p` and not implement
any of its own logic.  @smile13241324, do you agree with this idea?

smile13241324:
Yep I think this makes sense

bcc32:
Still valid

PR #16873, opened 2025-03-09
[core] Do not load ido eagerly at startup

bcc32:
ido is only regularly used by Spacemacs if the user has not configured
any completion layer, or during initial setup when the dotfile
installer wizard is prompting for an editing style.

Do not bother loading and configuring it when another completion UI is
configured.

### DIFF
--- a/layers/+spacemacs/spacemacs-completion/packages.el
+++ b/layers/+spacemacs/spacemacs-completion/packages.el
@@ -25,9 +25,13 @@
       '(
         (default-helm-config :location built-in)
         (default-ivy-config :location built-in)
-        flx-ido
-        (ido :location built-in)
-        (ido-vertical-mode :location built-in)))
+        (flx-ido :requires ido-vertical-mode)
+        (ido :location built-in
+             :toggle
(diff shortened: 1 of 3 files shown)


layers/+spacemacs/spacemacs-completion/packages.el
```diff
@@ -25,9 +25,13 @@
...
         (default-ivy-config :location built-in)
-        flx-ido
-        (ido :location built-in)
-        (ido-vertical-mode :location built-in)))
+        (flx-ido :requires ido-vertical-mode)
+        (ido :location built-in
+             :toggle
+             (not (or (configuration-layer/package-used-p 'helm)
+                      (configuration-layer/package-used-p 'ivy)
+                      (configuration-layer/package-used-p 'vertico))))
```
fnussbaum, line 33: ~~I haven't tried it, but this also looks like it requires a fix for #16346 first?~~
Nevermind, in this case there are no incorrectly valued `:toggle` forms involved, sorry for the noise.

fnussbaum (approved):

Issue #12105, opened 2019-03-29
Font size not correct when running as daemon if using themes-megapack


benbot:
#### Description :octocat:
font size not being set when run as daemon

#### Reproduction guide :beetle:
- Set font size to 100 in .spacemacs
- Start Emacs as daemon

*Observed behaviour:* :eyes: :broken_heart:
font size is incorrect

*Expected behaviour:* :heart: :smile:
font size should be 100

#### System Info :computer:
- OS: darwin
- Emacs: 27.0.50
- Spacemacs: 0.300.0
- Spacemacs branch: develop (rev. 6ff9e2ccc)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(osx themes-megapack helm dash auto-completion version-control evil-snipe git org restclient syntax-checking multiple-cursors lsp ipython-notebook
     (treemacs :variables treemacs-use-filewatch-mode t treemacs-use-follow-mode t)
     (ranger :variables ranger-override-dired t)
     (colors :variables colors-enable-nyan-cat-progress-bar t colors-colorize-identifiers 'variables)
     (shell :variables shell-default-height 40 shell-default-position 'bottom shell-default-shell 'eshell)
     yaml html markdown
     (javascript :variables javascript-backend 'lsp)
     (sql :variables sql-capitalize-keywords t)
     (go :variables gofmt-command "goimports")
     emacs-lisp clojure rust elixir python ruby shell-scripts react phoenix)
```
- System configuration features: NOTIFY KQUEUE ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS THREADS GMP

benbot:
This appears to be an issue with `themes-megapack` as that doesn't happen if I remove that layer

gdkrmr:
I have the same issue.

freidrichen:
Same here. After enabling themes-megapack my font settings are ignored if I start spacemacs with `emacsclient -c -a ""`, but works fine if I start it with plain old `emacs` command.

ccruquan:
See this comment 
https://github.com/syl20bnr/spacemacs/blob/24d9c3f7ee84e03476bb8e0ddcdac548df7f05c8/core/core-spacemacs.el#L123

change this line to 
``` lisp
(message "Setting the font...")
```

then restart emacs services, the font setting will active.

philippera:
[...]

PR #16874, opened 2025-03-09
[core] Remove needless message

bcc32:
https://github.com/syl20bnr/spacemacs/issues/12105 claims this fixes
an issue for daemon Emacs when the themes-megapack layer is used, but
I can't reproduce the issue (on Darwin, matching the original post).
Perhaps one of the upstream theme packages has been fixed?

### DIFF
--- a/core/core-spacemacs.el
+++ b/core/core-spacemacs.el
@@ -244,17 +244,6 @@ the final step of executing code in `emacs-startup-hook'.")
   (spacemacs/load-default-theme)
   ;; font
   (spacemacs|do-after-display-system-init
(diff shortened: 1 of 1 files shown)


fnussbaum:
I have a very vague memory of the message being necessary on my system a long time ago (without the themes-megapack layer), but cannot reproduce a problem nowadays -- I tested with Emacs 29 and 30 and up-to-date packages. I think we should at least test with Emacs 28.2 too, though.

fnussbaum (approved):
I've tested with Emacs 28.2, too, and could not reproduce any problem.

PR #16875, opened 2025-03-09
[org] Improvements when using org-modern-mode

bcc32:
When using org-modern-mode:

- Disable org-superstar from the spacemacs-org layer.  These packages
  have overlapping functionality, and org-modern's README recommends
  not using org-superstar at the same time.

- Use `global-org-modern-mode` directly rather than reimplementing it.
  This allows it to be easily disabled globally

- Add a key binding to toggle `global-org-modern-mode`

### DIFF
--- a/layers/+spacemacs/spacemacs-org/packages.el
+++ b/layers/+spacemacs/spacemacs-org/packages.el
@@ -33,7 +33,7 @@
     ;; layer. So it is easier for users to steal the ownership of the
     ;; `org' package.
     (default-org-config :location built-in)
-    org-superstar
+    (org-superstar :toggle (not (configuration-layer/package-used-p 'org-modern)))
     (space-doc :location (recipe :fetcher local))
     toc-org
     ))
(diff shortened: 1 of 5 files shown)


layers/+spacemacs/spacemacs-org/packages.el
```diff
@@ -33,7 +33,7 @@
     ;; layer. So it is easier for users to steal the ownership of the
     ;; `org' package.
     (default-org-config :location built-in)
-    org-superstar
+    (org-superstar :toggle (not (configuration-layer/package-used-p 'org-modern)))
```
fnussbaum, line 36: This seems to depend on a solution for #16346 which is not implemented or pushed yet?

bcc32: Oh, sad.  It works on my branch because I have a WIP fix for #16346.  Thanks for catching this.  Making the PR into a draft for now.

bcc32:
> When using org-modern-mode:
> 
>     * Disable org-superstar from the spacemacs-org layer.  These packages
>       have overlapping functionality, and org-modern's README recommends
>       not using org-superstar at the same time.

Moved this to #16346.  The other two parts, I pushed directly to develop as they did not depend on it.

PR #16880, opened 2025-03-13
* core/core-spacemacs-buffer.el: Nerd icons for Spacemacs bufer on teminal

sunlin7:
Make the nerd icons works on terminal. 

Extend the `dotspacemacs-startup-buffer-show-icons` with a new flag `all` to support the icons for both graphic frame and terminal frame (the terminal should be already support a nerd font first). 

Please help review the changes. Thanks

### DIFF
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -235,9 +235,18 @@ Spacemacs buffer."
   'spacemacs-dotspacemacs-init)
 
 (spacemacs|defc dotspacemacs-startup-buffer-show-icons nil
-  "If non-nil, show file icons for entries and headings on spacemacs buffer.
-This has no effect in terminal or if \"nerd-icons\" is not installed."
-  'boolean
+  "If non-nil, show file icons for entries and headings on spacemacs
+buffer, it requires \"nerd-icons\" package been installed.
+
+For graphic frame, it also requires a nerd font been installed (Execute
+the `nerd-icons-install-fonts' to install the font file).
+
+For terminal frame, it requires a nerd font avaliable on Terminal
+(Please make sure your terminal working with nerd font first then try
+this feature).
+
+Set the value to quoted `display-graphic-p' for graphic frame only."
+  '(choice boolean (const all))
   'spacemacs-dotspacemacs-init)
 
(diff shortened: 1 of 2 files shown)


bcc32:
Maybe the setting should be `display-graphic-p` for GUI-only, and then anything else non-nil is interpreted as applying to both GUI and terminal?  This would be consistent with how `dotspacemacs-mode-line-unicode-symbols` already works.

sunlin7:
@bcc32 Thanks for the comment. Keeping the `t` for graphic frame will be better for compatible reason.
Terminal with icons is complicated, especially for the ssh to remote server cases, the terminal maybe only support one font (like mintty/putty), then the Emacs displays icons on these terminal will met error (Need the user to confirm their terminal working nerd fonts or not). 
So this PR keeps `dotspacemacs-startup-buffer-show-icons` with `t` to works on graphic frame only.

sunlin7:
And yes, this would be consistent with how `dotspacemacs-mode-line-unicode-symbols` already works.

bcc32:
> And yes, this would be consistent with how dotspacemacs-mode-line-unicode-symbols already works.

Sorry, I just don't agree with your assessment about this.  Spacemacs already accepts the possibility that the default setting for `dotspacemacs-mode-line-unicode-symbols` may not work correctly for all users:

https://github.com/syl20bnr/spacemacs/blob/develop/doc/DOCUMENTATION.org#minor-modes

> Unicode symbols are displayed by default. Setting the variable
> =dotspacemacs-mode-line-unicode-symbols= to =nil= in your =~/.spacemacs= will
> display ASCII characters instead (may be useful in terminal if you cannot set an
> appropriate font).

If the Spacemacs defaults already require users to install such fonts for their terminal or otherwise actively disable certain features, I think it's acceptable to do that here as well, and avoid having these two variables interpret their values differently.

(Then we can reuse the `dotspacemacs|symbol-value` macro for example, although as a separate point I think it should probably be renamed.)

sunlin7:
Got you point.
I'm researching the unicode layer and try to understand how it works. Maybe need some time. Thanks

sunlin7:
Pushed the changes towards to the comments.

bcc32 (approved):
Looks good, thanks!

sunlin7:
Thank you !

PR #16883, opened 2025-03-15
* layers/+spacemacs/spacemacs-defaults/funcs.el: Copy file path on eww buffer

sunlin7:
Support `SPC y y` to copy file path on a *eww*  buffer.

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -881,6 +881,8 @@ otherwise the listed directory's path."
   (if-let* ((file-path (or (spacemacs--file-path)
                            (and (derived-mode-p 'dired-mode)
                                 (dired-get-filename nil t))
+                           (and (derived-mode-p 'eww-mode)
+                                (plist-get eww-data :url))
                            (spacemacs--directory-path))))
       (progn
         (kill-new file-path)


bcc32:
Shouldn't we just add a key binding for `eww-copy-page-url` instead of doing this?  URLs are not file paths, so reusing this key binding seems confusing.

sunlin7:
Thanks for the comment.

Acctually `SPC y y` is already available on a *eww* buffer, you can try the key sequence in a *eww* buffer; but it give a strange result. This patch will give a reasonable result for the key sequence. 
So this patch is try to fix a bug, does not introduce a new feature :)

bcc32:
I feel like that's an argument for simplification, not adding more special cases to this function (i.e., we should undo the dired-specific logic).  `spacemacs--directory-path` is already kind of weird, I feel like it should just always return `default-directory` rather than doing anything special.  (`list-buffers-directory` is not guaranteed to point to an actual directory, it's just shown in buffer listings for human consumption).

It does not feel intuitive to me that `SPC f y y` should ever copy a URL.

sunlin7:
Thank you for comment, got you point and close this ticket.

sunlin7:
But got a strange result after pressing `SPC y y` in a *eww* buffer as follow:
`/home/test/https:/github.com/` <-- the result is a path mixed URL.

That's a bug, absolutely.

bcc32:
I agree that's a bug, but it's actually because `spacemacs--directory-path` returns something very weird here (because it's incorrectly assuming that `list-buffers-directory` refers to an actual directory---and tries to call `file-truename` on it).  Closing in favor of making a new issue.

PR #16886, opened 2025-03-15
[defaults] Fix restoring window configuration in ediff quit hook

### LINKED ISSUE #16885: Ediff registry can't quit properly
When using `ediff-directories` in spacemacs, the `ediff-quit` doesn't work properly and causes the session keeps alive unless you forcibly close the control panel buffer. What's worse, it populates several empty ediff errors buffers as well.

You can easily replicate the issue by comparing two similar folders. Ticket #8202 also mentioned this before.

This is caused by [this line](https://github.com/syl20bnr/spacemacs/blob/39eccb1ce78c72ee594aa88f9e3b4a87026094b6/layers/%2Bspacemacs/spacemacs-defaults/packages.el#L214).

bcc32:
Running winner-undo in ediff-quit-hook doesn't generally work, because
ediff setup may have changed the window configuration multiple times.

Also, running it too early interferes with ediff-cleanup-mess and
prevents the control buffer and other buffers from being killed
properly by ediff's own cleanup.

Fix #16885.

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -974,6 +974,16 @@ variable."
           (with-current-buffer b
             (ediff-delete-temp-files)))))))
 
+(defvar spacemacs//ediff-saved-window-configuration nil)
+
+(defun spacemacs//ediff-save-window-configuration ()
+  (setq spacemacs//ediff-saved-window-configuration
+        (current-window-configuration)))
+
+(defun spacemacs//ediff-restore-window-configuration ()
+  (when spacemacs//ediff-saved-window-configuration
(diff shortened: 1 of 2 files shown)


fnussbaum:
Looks good, thanks. I guess having multiple concurrent ediff sessions is probably niche enough such that we don't need to bother supporting it here?

bcc32:
It's pretty niche.  At the very least, the previous version of this code made no consideration of such a use-case, and the new code feels "strictly better" IMO.

FYI Spacemacs also sets `org-agenda-restore-windows-after-quit` non-nil which has a very similar effect.  Easy enough for users to opt out of either customization, anyway.

neutrous:
@bcc32 I can confirm it works, thank you :)

PR #16425, opened 2024-06-01
Add numerical prefix functionality to window delete commands

fnussbaum:
Some changes to #16247. In particular this commit restores the previous <kbd>SPC u</kbd>-prefix functionality and handles a few more edge cases.

PR #16890, opened 2025-03-16
Spring Cleaning: remove spacemacs specific buffer and window kill functions

cody-vibe:
Removes the following:
1. `spacemacs/kill-this-buffer` as Emacs has a built-in `kill-current-buffer` that has extra frame safety baked in.
2. `spacemacs/delete-window` as we already have `SPC b x` and `SPC w x` that handle the need for the universal argument.

Also, because `SPC b x` and `SPC w x` are easier to type than `SPC u SPC b d` `SPC u SPC w d` having a special function just for that part of the functionality seems like a waste to me.

This should simplify the code a bit as it removes duplication.

Aside: I've noticed that while editing `.el` code in Spacemacs, the indentation changes on save. So I had to open them literally to prevent messed up diffs. Any suggestions on how to avoid this (or are you OK with receiving PRs with indent changes)


smile13241324:
@sdwolfz hi there prs with indent changed are perfectly fine, in my opinion having everyone use the same indention saves a lot of time in the long run i.e. as in go. 

For files which are not yet indented one can use github to filter out indent changes in the diff.

So no worries and welcome back ;)

fnussbaum:
>     1. `spacemacs/kill-this-buffer` as Emacs has a built-in `kill-current-buffer` that has extra frame safety baked in.

I think this makes sense, thanks.

>     2. `spacemacs/delete-window` as we already have `SPC b x` and `SPC w x` that handle the need for the universal argument.
> 
> 
> Also, because `SPC b x` and `SPC w x` are easier to type than `SPC u SPC b d` `SPC u SPC w d` having a special function just for that part of the functionality seems like a waste to me.

I agree with this, too, though in my branch I have gotten used to this unmerged PR: https://github.com/syl20bnr/spacemacs/pull/16425. I think it is sufficiently different from #16247 such that we could make those changes without permission. But I would be interested to hear what others think, perhaps removing the function completely is indeed preferred, also over adding numerical prefixes? 

> Aside: I've noticed that while editing `.el` code in Spacemacs, the indentation changes on save. So I had to open them literally to prevent messed up diffs. Any suggestions on how to avoid this (or are you OK with receiving PRs with indent changes)

Ideally, whitespace changes would be in a separate commit within in the same PR, but it does not matter too much. In case you really want to turn off the automatic formatting, you can set `emacs-lisp-format-on-save` to nil. (But I agree with @smile13241324 that auto-formatting is generally advantageous here.)

cody-vibe:
```
~NUM SPC w d~          | delete window number NUM
```

Interesting, this is basically ace-delete-window, but you specify the number up-front rather than after executing the command.

The thing I find weird about prefixes is basically this: I first need to decide on which item I want to perform an action, then I decide what action I want to perform. Isn't this kind of backwards? Normally I think, I want to perform this action, on this item.

Anyway, I'm not against keeping the prefix for those 2 functions. Just as long as we use kill-current-buffer.

fnussbaum:
If there are no objections I think we can merge this? Upstreaming the numerical prefixes is not a priority for me.

> Isn't this kind of backwards? Normally I think, I want to perform this action, on this item.

I got used to both ways. Selecting first, before acting on the selection also seems to be common, for example in context menus (or [embark](https://github.com/oantolin/embark) actions).

bcc32 (approved):
Added a small tweak but otherwise LGTM.

PR #16891, opened 2025-03-17
Fix incorrect nerd font icon names in Spacemacs buffer

SteveLane:
Edits nerd-font call so that icons are displayed properly in startup buffer.

Fixes issue where the startup buffer is not displayed properly due to incorrect icon names (`nerd-icons-codicon: Unable to find icon with name ‘nt-cod-radio_tower’ in icon set ‘codicon’`)


bcc32:
Thanks for the fix!

PR #16893, opened 2025-03-19
Remove pdump feature from Spacemacs which is actually not work, Fix #16892

### LINKED ISSUE #16892: Does the pdump really work in Spacemacs?
I tried the pdump feature in Spacemacs, but it seems never works.
My environment is Ubuntu 22.04 x86_64, emacs-30/emacs-31, enabled/disabled native-compile feature.

sunlin7:
As the discussing on the #16892, the pdump is not recommend any more (actually I tried but it can not work).


core/core-dumper.el
```diff
@@ -1,204 +0,0 @@
...
-  "Save `load-path' variable."
-  (setq spacemacs-dump-load-path load-path))
-
-(defun spacemacs/dump-restore-load-path ()
-  "Restore the `load-path' variable from the dump. "
-  (spacemacs|unless-dumping
-    (when (not (null spacemacs-dump-load-path))
-      (setq load-path spacemacs-dump-load-path))))
-
-(defun spacemacs/defer (&optional idle-time)
```
bcc32, line 99: This function is used in lots of layers.  We need to replace its callsites (probably mostly with `t`?).  In my config at least, there is an error in the `use-package` for `recentf`.

core/core-spacemacs.el
```diff
@@ -202,8 +201,6 @@ the final step of executing code in `emacs-startup-hook'.")
     (unless (frame-parameter nil 'fullscreen)
       (toggle-frame-maximized))
     (add-to-list 'default-frame-alist '(fullscreen . maximized)))
-  (spacemacs|unless-dumping
-    (dotspacemacs|call-func dotspacemacs/user-init "Calling dotfile user init..."))
```
bcc32, line 206: `spacemacs|unless-dumping` calls should be replaced with their bodies.  This change accidentally causes `dotspacemacs/user-init` to never get called.

layers/+lang/c-c++/layers.el
```diff
@@ -24,6 +24,6 @@
 (when (boundp 'c-c++-backend)
   (configuration-layer/declare-layer-dependencies
    (pcase c-c++-backend
-     ('lsp-clangd '(lsp dap))
-     ('lsp-ccls '(lsp dap))
```
bcc32: Is this change accidental?

bcc32 (changes requested):

bcc32:
Thanks for taking the time to do this!  BTW I have added a couple of follow-up changes.

core/core-dumper.el
sunlin7: You're right, my latest push had replace them to `t`. Thanks!

core/core-spacemacs.el
sunlin7: You're correct, the statement is restore in my latest change, thanks!

layers/+lang/c-c++/layers.el
sunlin7: Restore to original lines, thank you!

layers/+spacemacs/spacemacs-defaults/packages.el
```diff
@@ -386,11 +385,10 @@
 
 (defun spacemacs-defaults/init-recentf ()
   (use-package recentf
-    :defer (spacemacs/defer)
+    :defer t
     :commands (recentf-save-list)
     :init
-    (spacemacs|require-when-dumping 'recentf)
-    (when (spacemacs/defer)
```
bcc32, line 393: I think `(when t ...)` should be simplified away.  Otherwise looks good.

sunlin7: Sure, I had updated the RP. Thank you!

bcc32 (approved):

bcc32:
The code looks good to me, thanks for your work!  Since this is a pretty large change I'd like to give the other maintainers a little time to weigh in before merging.

sunlin7:
Sure, thank you!

bryce-carson (approved):
The only thing that concerns me is the state of CHANGELOG.develop; has it been well-maintained?

When we cut a release will we have proper documentation of all changes since the last master release? Should we start a project to document the large void of time since the last release by deferring to the git log?

Apart from my concern about CHANGELOG.develop, the changes look fine. I haven't tested them myself. My concern about that isn't relevant to this PR.

bcc32:
> The only thing that concerns me is the state of CHANGELOG.develop; has it been well-maintained?

I certainly haven't been logging my own changes to the change log at all.  I don't think we're very consistent about it overall, perhaps it would be worth having some sort of CI check for PRs.

bryce-carson:
> > The only thing that concerns me is the state of CHANGELOG.develop; has it been well-maintained?
> 
> I certainly haven't been logging my own changes to the change log at all.  I don't think we're very consistent about it overall, perhaps it would be worth having some sort of CI check for PRs.

That's not a bad idea, but with our previous workflow of just working off of development *for eight years*, I think it's more appropriate to abandon the file and instead educate users on how to check the git changelog.

Ensuring all commit messages are reasonably descriptive is going to be more maintainable in the long run than maintaining an *additional* change log. We don't have the editorial capacity to maintain it, in my opinion.

If we were corporate it'd be worth it for customers who only go from n to n+1 and want to know about oddball changes that would kill their hacks, but since we're an Emacs config I think it's more reasonable to just tell people to check the git changelog.

Thoughts on abandoning the file? Otherwise, using CI to check all commits and PRs modify the file, *beginning after our next release*, would be a good idea.

Since we're cutting a new release and using master again I think it's a good opportunity to have the contributors and the maintainers decide on some breaking workflow changes or reassess what is reasonable and within our capacity.

fnussbaum (approved):
I have updated the dotfile template, everything else looks good. (This just made me wonder whether it would be beneficial to rename the dotfile template such that it gets searched by ripgrep by default.)

bcc32:
> I have updated the dotfile template, everything else looks good. (This just made me wonder whether it would be beneficial to rename the dotfile template such that it gets searched by ripgrep by default.)

Nice catch, thanks.  I agree with renaming the template, maybe just `dotspacemacs.template` would be fine?  Or perhaps `dotspacemacs-template.el` so that it gets picked up by `--type elisp` as well.

bcc32 (approved):

sunlin7:
Thank you, that's really a hidden one (The `rg` is in my local) :)

bcc32:
Thank you @sunlin7!

sunlin7:
Thank you for the passions on Spacemacs!

PR #16895, opened 2025-03-19
[core] Activate site packages

### LINKED ISSUE #16894: [mu4e] mu4e package not properly activated
mu4e must be installed as a site package, not from ELPA.  However,
unlike ELPA packages, Spacemacs does not actually call
`package-activate` on site packages, which means that their autoloads
are not loaded properly, nor are their info manuals added to
`Info-directory-list`.

This causes frequent errors because `mu4e` actually has commands
(e.g., `C-h` in headers view) that visit various Info nodes.  Besides
which, it would be better to actually activate the installed info
manuals, of course.

bcc32:
Fix #16894.

Since the `fuel` package, as comes with the Factor programming
language distribution, doesn't seem to have an autoloads file,
activating it would fail.  In order to make this work, switch to the
ELPA version of `fuel`, which does not have this problem.

cc @timor since I'm not sure any of the current maintainers are
familiar with this layer.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -1999,7 +1999,7 @@ RNAME is the name symbol of another existing layer."
             (when dir
               (add-to-list 'load-path dir)))
           ;; configuration
-          (unless (memq (oref pkg :location) '(local site built-in))
+          (unless (memq (oref pkg :location) '(local built-in))
             (configuration-layer//activate-package pkg-name))
           (cond
            ((eq 'dotfile (car (oref pkg :owners)))
--- a/layers/+email/mu4e/packages.el
+++ b/layers/+email/mu4e/packages.el
@@ -50,7 +50,6 @@
 
 (defun mu4e/init-mu4e ()
   (use-package mu4e
-    :commands (mu4e mu4e-compose-new)
     :init
     (spacemacs/set-leader-keys "aem" 'mu4e)
     (setq mail-user-agent 'mu4e-user-agent


fnussbaum:
> Since the fuel package, as comes with the Factor programming
language distribution, doesn't seem to have an autoloads file,
activating it would fail.

`package-activate` seems to just return nil in this case, so I think keeping it as a site package would maintain the status quo? (Although I guess, for users who need an older version, it would be better to use something like quelpa to install it properly. It might also make sense for factor to include proper emacs packages in their releases in the future.)

> In order to make this work, switch to the
ELPA version of fuel, which does not have this problem.

I'm not familiar with factor and fuel, but I wonder if the version on MELPA is backward compatible with older releases of factor. Especially since there is some infrastructure in `funcs.el` to unload and load different fuel versions. Hence I think it might be safer to keep it as a site package?

bcc32:
> `package-activate` seems to just return nil in this case, so I think keeping it as a site package would maintain the status quo? (Although I guess, for users who need an older version, it would be better to use something like quelpa to install it properly. It might also make sense for factor to include proper emacs packages in their releases in the future.)

Ah, I noticed the error message (from `with-demoted-messages`) in `package-activate-1` which complains that it can't load the autoloads file.  I guess this is probably fine, not the worse regression.
 
> I'm not familiar with factor and fuel, but I wonder if the version on MELPA is backward compatible with older releases of factor. Especially since there is some infrastructure in `funcs.el` to unload and load different fuel versions. Hence I think it might be safer to keep it as a site package?

Seems like a reasonable concern.  At least, given that you pointed out that `package-activate` doesn't actually fail hard on the missing autoloads, it seems better to keep it as a site package for now.

fnussbaum (approved):

PR #16899, opened 2025-03-20
Shorten "helm-\(.*?\)-smart-do-search\(.*\)" to "hsearch-\1\2"

Bost:
With old names obsoleted for backward compatibility.

### DIFF
--- a/layers/+completion/helm/funcs.el
+++ b/layers/+completion/helm/funcs.el
@@ -540,16 +540,17 @@ Removes the automatic guessing of the initial value based on thing at point. "
 (defmacro spacemacs||set-helm-key (keys func)
   "Define a key bindings for FUNC using KEYS.
 Ensure that helm is required before calling FUNC."
-  (let ((func-name (intern (format "lazy-helm/%s" (symbol-name func)))))
+  (let* ((actual-func (if (consp func) (cdr func) func))
+         (func-name (intern (format "lazy-helm/%s" (symbol-name actual-func))))
+         (func-param (if (consp func) `(,(car func) . ,func-name) func-name)))
     `(progn
        (defun ,func-name ()
          ,(format "Wrapper to ensure that `helm' is loaded before calling %s."
-                  (symbol-name func))
+                  (symbol-name actual-func))
          (interactive)
          (require 'helm)
-         (command-execute ',func))
-       (spacemacs/set-leader-keys ,keys ',func-name))))
-
+         (command-execute ',actual-func))
+       (spacemacs/set-leader-keys ,keys ',func-param))))
  ;; Find files tweaks
 
 (defun spacemacs//helm-find-files-edit (candidate)
--- a/layers/+completion/helm/packages.el
+++ b/layers/+completion/helm/packages.el
@@ -125,11 +125,17 @@
     (spacemacs||set-helm-key "sj"   spacemacs/helm-jump-in-buffer)
     ;; search with grep
     (spacemacs||set-helm-key "sgb"  spacemacs/helm-buffers-do-grep)
-    (spacemacs||set-helm-key "sgB"  spacemacs/helm-buffers-do-grep-region-or-symbol)
+    (spacemacs||set-helm-key
+     "sgB" ("grep-search buffers w/ input" .
+            spacemacs/helm-buffers-do-grep-region-or-symbol))
     (spacemacs||set-helm-key "sgf"  spacemacs/helm-files-do-grep)
-    (spacemacs||set-helm-key "sgF"  spacemacs/helm-files-do-grep-region-or-symbol)
+    (spacemacs||set-helm-key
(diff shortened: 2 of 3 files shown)


bcc32:
What benefit does this have?  Sure, the names are shorter, but now they also obscure the fact that they are specific to helm.  Is `hsearch` a popular abbreviation for helm-related search functionality outside of Spacemacs?

Bost:
> What benefit does this have?

`SPC s` is bound to search/symbol functionality. When this keybinding is pressed, the function names are abbreviated, so it's difficult to identify the desired function. That's a poor user experience. See screenshots.

The original structure is `helm-<something>-smart-do-search-region-or-symbol`. The `search-<something>` is the most important piece of information. Next in importance is the scope, which is `region-or-symbol`. The fact that the search is done by 'helm-smart-do' is less critical, so I've replaced it with a single letter `h`.

My choice of names may not be optimal, so please feel free to propose better alternatives.
spacemacs-hsearch-after
spacemacs-hsearch-before

bcc32:
How about fixing that in just the key bindings themselves, since this is primarily to affect which-key?  I feel like the new command names are a little better but still not that clear.  As an example:

```
      ;; current dir scope
      "sd"  '("search dir" . spacemacs/helm-dir-smart-do-search)
      "sD"  '("search dir w/ input" . spacemacs/helm-dir-smart-do-search-region-or-symbol)
```

Alternatively, if the other maintainers think renaming the commands is fine, we should perhaps rename to conform to the `helm/` layer prefix convention rather than `spacemacs/helm`

fnussbaum:
>How about fixing that in just the key bindings themselves, since this is primarily to affect which-key?

This is what I intended to suggest, too.

Bost:
Aaah, you're right. I reworked the patch. Please have a look.

layers/+completion/helm/funcs.el
```diff
@@ -540,16 +540,17 @@ Removes the automatic guessing of the initial value based on thing at point. "
 (defmacro spacemacs||set-helm-key (keys func)
   "Define a key bindings for FUNC using KEYS.
 Ensure that helm is required before calling FUNC."
-  (let ((func-name (intern (format "lazy-helm/%s" (symbol-name func)))))
+  (let* ((actual-func (if (consp func) (cdadr func) func))
```
bcc32: Since this macro expects unquoted function names in the normal case, I think it makes sense to also accept unquoted `("description" . func-name)` rather than with the quote.  Then this can just use `cdr` instead of `cdadr`---the latter is surprising.

layers/+completion/helm/packages.el
```diff
@@ -250,34 +256,45 @@
...
+              spacemacs/helm-buffers-do-rg-region-or-symbol)
       "stb" 'spacemacs/helm-buffers-do-pt
-      "stB" 'spacemacs/helm-buffers-do-pt-region-or-symbol
+      "stB" '("pt-search buffers w/ input" .
+              spacemacs/helm-buffers-do-pt-region-or-symbol)
       ;; current file scope
-      "ss"  'spacemacs/helm-file-smart-do-search
-      "sS"  'spacemacs/helm-file-smart-do-search-region-or-symbol
+      "ss"  'spacemacs/hsearch-file
+      "sS"  'spacemacs/hsearch-file-region-or-symbol
```
bcc32: I think this change needs to be reverted too?

bcc32 (changes requested):
Mostly LGTM now, thanks for reworking the PR.  Made a couple small suggestions.

layers/+completion/helm/packages.el
Bost: Yeah, sorry. That slipped through.

layers/+completion/helm/funcs.el
Bost: Done. BTW, we have a slight discrepancy in the quotation:
```
(spacemacs/set-leader-keys "<keys>"  '<func>)
(spacemacs/set-leader-keys "<keys>"  '("<desc>" . <func>))
```
vs.
```
(spacemacs||set-helm-key "<keys>"  <func>)
(spacemacs||set-helm-key "<keys>"  ("<desc>" . <func>))
```

It would be nice to have the same function signature for both `spacemacs||set-helm-key` and `spacemacs/set-leader-keys`.

Bost:
Please have a look again.

bcc32 (approved):

layers/+completion/helm/funcs.el
bcc32: Yeah, agreed it would be more consistent.  I guess one reason it might be difficult is that `spacemacs||set-helm-key` needs its input to be a symbol and not just any old function (or perhaps a keymap, etc.).

bcc32:
Looks good, thanks for your contribution!

PR #16900, opened 2025-03-20
* core/core-spacemacs.el: Fix regression introduced in syl20bnr#16656 for require@LOAD-HINTS

sunlin7:
Enhance the performance for the function `require@LOAD-HINTS`.


bcc32:
Can you add more context to the description/commit messages?  The changes look more like a bugfix than just "improving performance", so it would be helpful to know what was going wrong previously.

sunlin7:
Hi @bcc32 
The previous one only applied on the `(require 'X "Y")` to get a full filename to replace the "Y" to avoid searching `load-path`; 
current one will apply on the `(require 'X)` to get a full filename for the feature to get better performance.

bcc32:
Okay, I think I understand.  In that case, was load-hints basically providing no benefit at all, because `require` almost always going through the slow path?

sunlin7:
Hi @bcc32  I go through the commit logs, confirm it's a regression issue introduced in the https://github.com/syl20bnr/spacemacs/commit/298489256ccf0f525125b6816f942ba311ec7e66. 
The commit message is corrected to make it clear.
Please help review again. Thanks you!

bcc32 (approved):

bcc32:
Great, thanks for the fix.

sunlin7:
Thank you !

PR #16909, opened 2025-03-24
[evil] Don't enable evil-collection for emacs editing style

bcc32:
`holy-mode` users expect their key bindings to be largely the same as
Emacs ones.  For example, revert-buffer in special-mode should be
bound to `g`, not `g r`.

### DIFF
--- a/layers/+spacemacs/spacemacs-evil/packages.el
+++ b/layers/+spacemacs/spacemacs-evil/packages.el
@@ -25,7 +25,7 @@
       '(
         evil-anzu
         evil-args
-        evil-collection
+        (evil-collection :toggle (not (eq dotspacemacs-editing-style 'emacs)))
         evil-cleverparens
         (evil-escape :location (recipe :fetcher github
                                        :repo "smile13241324/evil-escape"))


fnussbaum:
> For example, revert-buffer in special-mode should be bound to g, not g r.

Is there a specific mode where evil-collection causes such conflicts? AFAIK in `holy-mode` all buffers should be in `evil-emacs-state` where evil-collection does not normally change key bindings. 

Disabling evil-collection based on `dotspacemacs-editing-style` would also affect users who like to change the editing style within a session using <kbd>M-m t E e</kbd>, for example.

bcc32:
> > For example, revert-buffer in special-mode should be bound to g, not g r.
> 
> Is there a specific mode where evil-collection causes such conflicts?

special-mode is a commonly-used parent mode.  Things like tabulated-list-mode and modes derived from it, expect `g` to be bound to revert-buffer.  `M-x list-timers` is an easy example of such a buffer.  (Spacemacs defines its own `g r` key bindings for many of these, but it does not mention `timer-list-mode`).

> AFAIK in `holy-mode` all buffers should be in `evil-emacs-state` where evil-collection does not normally change key bindings.

That's not what I observe without this change though.  evil-collection seems to affect holy-mode.

> Disabling evil-collection based on `dotspacemacs-editing-style` would also affect users who like to change the editing style within a session using M-m t E e, for example.

Yes, that's true, but I figure that's a least a minority use-case.  And it did not seem obvious to me how to "undo" `evil-collection-init`, meaning it would still not be easily toggle-able as a minor mode anyway.  So `dotspacemacs-editing-style` at least is a good way to determine a user's primary preference.

fnussbaum:
Ah thanks, I can reproduce it now (I should have just tried `special-mode` in the first place, instead of `helpful-mode`).

> That's not what I observe without this change though. evil-collection seems to affect holy-mode.

Indeed, this is bad. In the majority of modes, evil-collection binds keys only in normal state, but in a few cases it mutates the actual mode maps.

I think, ideally, all of those instances should get addressed in evil-collection, by making this behaviour optional or reversable.

I grepped for `evil-collection-define-key\ nil` and `define-key` in evil-collection, and compared with `spacemacs-evil-collection-allowed-list`: Only magit, magit-todos, ediff, and simple (which covers `special-mode`) seem to be affected in Spacemacs. For magit, `evil-collection-magit` provides `evil-collection-magit-init` and `evil-collection-magit-revert`. Similarly, for ediff, there is `evil-collection-ediff-setup` and `evil-collection-ediff-revert`. Unfortunately, those functions currently don't work flawlessly, otherwise we could make use of them within `spacemacs-editing-style-hook`.

> Yes, that's true, but I figure that's a least a minority use-case.

I tend to agree, though the functionality is prominently advertised in the [README](https://github.com/syl20bnr/spacemacs?tab=readme-ov-file#introduction) and [documentation](https://github.com/syl20bnr/spacemacs/blob/1aedf3e94a9c11320e097ec51b4ecdf020c16395/doc/DOCUMENTATION.org#who-can-benefit-from-this) for pair-programming. But there are quite a few edge cases already where toggling the editing style is not quite the same as changing `dotspacemacs-editing-style` and restarting, hence I think it is fine to merge this PR as is. I might revisit this later.

fnussbaum (approved):

PR #16915, opened 2025-03-25
[layers][doc][core] key-sequence format as shown by M-x describe-key

Bost:


bcc32 (approved):
LGTM.  For the context of anybody else who reads this, the return value of `kbd` on the strings changed in this PR is the same as before.  This just makes the syntax more consistent.

PR #16917, opened 2025-03-26
'replaced if-let with if-let* and when-let with when-let*'

emacs18:
Following is from NEWS file of emacs 31.

** 'if-let' and 'when-let' are now obsolete.
Use 'if-let*', 'when-let*' and 'and-let*' instead.

This effectively obsoletes the old '(if-let (SYMBOL SOMETHING) ...)' single binding syntax, which we'd kept only for backwards compatibility.


bcc32:
The libraries in core/libs are automatically updated by the bot from upstream, so these changes will simply get overwritten.  Please consider submitting a PR [upstream ](https://github.com/melpa/package-build) instead.

PR #16919, opened 2025-03-26
[navigation] Do not consider */# for evil-repeat

### LINKED ISSUE #4123: repeat (.) after search (*) repeats the search
Search with `*` then perform an action. Search again with `*` and then hit `.` to repeat the action.

Instead of repeating the action, the search is repeated. Whoops.

I think this might be new on develop, since I think I would have noticed this on master.

bcc32:
Fix #4123.

I did some testing at my site but am not super confident.  Would
appreciate a second set of eyes on this change.


fnussbaum (approved):
Thanks, this is a correct fix. Hydra heads set `this-command` as expected, so we don't need to handle them explicitly.

PR #16920, opened 2025-03-27
[defaults] Replace some key binding definitions with those from evil-collection

bcc32:
This PR switches to using evil-collection for key bindings definitions
for process-menu-mode and proced.

Spacemacs defines only a minimal evilified key binding scheme with
"gr" bound to revert-buffer for these modes, whereas evil-collection's
key bindings for these modes includes "gr" in addition to other useful
key bindings.  (Also, it lets us get rid of some unnecessary code.)


fnussbaum:
Are the evil-collection bindings actually better or more consistent here? I was under the impression that the the `evilified-state-evilify-map` macro works quite well for simple mode maps like `proced-mode-map`. It binds a few motion keys, other key bindings are left as in emacs-state (or shifted). [Here](https://github.com/syl20bnr/spacemacs/pull/14200#issue-760861577) at least one user expressed a preference of evilified bindings over evil-collection. I don't have a strong opinion about this, but in general I feel we should either use evil-collection everywhere or limit it to modes where it is strictly better.

Now I'm just cherry-picking, but evil-collection is in many cases not completely consistent: For example, in dired (where we use evil-collection), <kbd>u</kbd> is bound to `dired-unmark`. In vanilla and evilified proced, <kbd>u</kbd> is analogously bound to `proced-unmark`, while it is bound to `proced-undo` by evil-collection-proced.

bcc32:
> Are the evil-collection bindings actually better or more consistent here? I was under the impression that the the `evilified-state-evilify-map` macro works quite well for simple mode maps like `proced-mode-map`.

At least from experience at myself, it's decently confusing to users in that it looks like normal-state but doesn't bind all of the normal navigation keys; and it causes issues with derived modes.  Meanwhile, the remapping of vanilla keys is not always intutive or comlete.  Maybe these can be improved over time.

> [Here](https://github.com/syl20bnr/spacemacs/pull/14200#issue-760861577) at least one user expressed a preference of evilified bindings over evil-collection.

ping @thanhvg in case you'd like to weigh in/clarify your intended meaning in that comment.

I was not clear on whether this comment meant "I prefer evilified key bindings in general over evil-collection" or "I don't want to break all the existing Spacemacs key bindings so I'm limiting the number of modes that we enable in evil-collection".  I interpret the comment as meaning the latter.

>  I don't have a strong opinion about this, but in general I feel we should either use evil-collection everywhere or limit it to modes where it is strictly better.

I'd lean towards "prefer to use it everywhere" to minimize the amount of maintenance we have to do that duplicates other people's work.  (e.g., when comands get deleted/renamed in various packages).  (That's my main motivation for this PR.)
 
> Now I'm just cherry-picking, but evil-collection is in many cases not completely consistent: For example, in dired (where we use evil-collection), u is bound to `dired-unmark`. In vanilla and evilified proced, u is analogously bound to `proced-unmark`, while it is bound to `proced-undo` by evil-collection-proced.

Fair objection, but OTOH with evilified proced there is no vim-friendly binding for `proced-undo` at all.  It's a trade-off either way, but if you think this change is confusing I'm fine with dropping the PR.

thanhvg:
Hey, my opinion is no longer relevant for this project.

But as you see a lot of tedious work in spacemacs is adding key bindings for evil.  So I'd prefer spacemacs to use and contribute  to  evil-collections as much as possible which will benefit everyone not just spacemacs.

Btw, I'm glad to see that this project is making great progress recently with the new collaborators.

fnussbaum:
> I'd lean towards "prefer to use it everywhere" to minimize the amount of maintenance we have to do that duplicates other people's work.

Makes sense, I think we can proceed with this then.

>  So I'd prefer spacemacs to use and contribute to evil-collections as much as possible which will benefit everyone not just spacemacs.

Thanks for chiming in! Sounds good, I'll merge this as a first step.

fnussbaum (approved):

PR #16921, opened 2025-03-27
[core] More small refactors to the configuration layer

bcc32:


fnussbaum (approved):
Looks good, thanks. I suppose one could also write `(car (alist-get element alist))` instead of `(cadr (assq element alist))`, but it probably would not make a significant difference in terms of readablity.

PR #16923, opened 2025-03-27
[org] Add `, T T` binding for org-agenda-todo

jaeyeom:
The keybinding `, T T` can be used in Org Mode to cycle todo keywords (`org-todo`) as well as `C-c C-t`. There's another keybinding `t` for that purpose, but `, T T` is also good for users who have muscle memory in org files.

It may also make sense to bind `SPC f s` for `org-save-all-org-buffers`, since `C-x C-s` is bound as well, but that's overriding the exising keybinding, so it wasn't applied here.

### DIFF
--- a/layers/+emacs/org/README.org
+++ b/layers/+emacs/org/README.org
@@ -938,6 +938,7 @@ The evilified org agenda supports the following bindings:
 | ~SPC m i P~          | org-agenda-priority               |
 | ~SPC m i t~          | org-agenda-set-tags               |
 | ~SPC m s r~          | org-agenda-refile                 |
+| ~t~ or ~SPC m T T~   | org-agenda-todo                   |
 | ~M-j~                | next item                         |
 | ~M-k~                | previous item                     |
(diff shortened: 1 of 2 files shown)


bcc32:
I think it would be fine to bind `SPC f s`.  Users probably expect it to do what `C-x C-s` would do in the same buffer.

jaeyeom:
> I think it would be fine to bind `SPC f s`. Users probably expect it to do what `C-x C-s` would do in the same buffer.

Yeah.. but it can be controversial because I don't know other good use cases that `SPC ....` keybinding (unless `SPC m`) is overriden in a mode. I'll make a separate PR, so it can be reviewed separately.

bcc32:
Ah, good point, that's fair enough.  Up to you whether you want to do it.

Could you add a README.org entry (for the org agenda key bindings section) for the new key binding?

jaeyeom:
@bcc32 `README.org` is updated. Any chance you could take a look?

bcc32:
Merged, thanks for the PR!

PR #16928, opened 2025-03-30
[core] Rework dotspacemacs-sync-packages behavior for site packages

### LINKED ISSUE #16898: [mu4e] Spacemacs attempts to delete package mu4e at startup
With `dotspacemacs-install-packages` set to `'used-only`, and when
using the `mu4e` layer (which requires the `mu4e` package to be
installed in the system packages), Spacemacs attempts and fails to
delete that package at startup.  This is noisy and it should not even
attempt to do so.

This problem seems to be present in Spacemacs versions at least a year
old, but I wasn't able to immediately test anything further back
without breaking under Emacs 30.1.  I hadn't noticed because I usually
have `dotspacemacs-install-packages` set to `'used-but-keep-unused`

bcc32:
Fix #16898.

This PR consists of two commits that fix two separate bugs with site
packages.  Please see the individual commit messages for details.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -345,7 +345,7 @@ is ignored."
 
 (cl-defmethod cfgl-package-distant-p ((pkg cfgl-package))
   "Return non-nil if PKG is a distant package (i.e. not built-in Emacs)."
-  (and (not (memq (oref pkg location) '(built-in site local)))
+  (and (not (memq (oref pkg location) '(built-in local)))
        (not (stringp (oref pkg location)))))
 
 (cl-defmethod cfgl-package-get-safe-owner ((pkg cfgl-package))
@@ -1269,7 +1269,7 @@ USEDP if non-nil indicates that made packages are used packages."
 
 (defun configuration-layer//filter-distant-packages
     (packages usedp &optional predicate)
-  "Return the distant packages (ie to be intalled).
+  "Return the distant packages (i.e., to be installed).
 If USEDP is non nil then returns only the used packages; if it is nil then
 return both used and unused packages.
 PREDICATE is an additional expression that eval to a boolean."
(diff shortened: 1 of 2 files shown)


core/core-configuration-layer.el
```diff
@@ -2365,6 +2369,13 @@ When called interactively, delete all orphan packages."
     ;; (message "dependencies: %s" dependencies)
     ;; (message "implicit: %s" implicit-packages)
     ;; (message "orphans: %s" orphans)
+
+    ;; Do not noisily fail to delete system packages.
+    (setq orphans
+          (cl-delete-if (lambda (pkg-name)
+                          (let ((pkg-desc (cadr (assq pkg-name package-alist))))
+                            (configuration-layer//system-package-p pkg-desc)))
+                        orphans))
```
fnussbaum: Is this actually still necessary, after having declared site packages as distant? The PACKAGES passed to `configuration-layer/delete-orphan-packages` should never end up in the orphans list (the docstring seems misleading here). And IIUC the function should always get called with all distant packages.

fnussbaum: Ah, I finally understood that this is about unused site packages (which the commit message indeed says). An unused site package that is properly installed will end up in the orphans list when interactively calling `configuration-layer/delete-orphan-packages`. But then, if a user actually does this, perhaps they would like to be notified about such unused site packages? 

With this change, the `orphans-count` still includes unused site packages. I guess this could look like a bug, so perhaps we should change this, or print some kind of message?

Perhaps we could consider one of the following alternatives:
1. Move the `system-package-p` check to `configuration-layer//package-orphan-p`?
2. Print a single message about the skipped site packages.
3. Keep noisily trying to delete site packages (only in interactive calls, not on startup).

bcc32: My use case is for site packages that are installed by an administrator in a shared place, which the user should not be notified about because they _cannot_ deleted by the user.  I would at least like to be able to preserve this property, perhaps making it configurable.  (3) seems appealing, but I'll need to think about it a little further.

fnussbaum: Makes sense. If you prefer, we can merge this first and then possibly improve it.

bcc32: Sorry it took me so long to get back to this, I have been very busy the last few weeks.  I have refactored this a bit and also made the change proposed in (3).

smile13241324 (approved):
This looks pretty good to me!
Lets merge and improve as we go along.

Thanks for the hard work :)

PR #16934, opened 2025-04-01
[mu4e] Fallback to loading mu4e-autoloads if package not activated

### LINKED ISSUE #16931: [mu4e] Wrong type argument: commandp, mu4e
1. Install spacemacs
2. Select emacs mode and normal or base
3. Add mu4e layer and restart
4. Start mu4e: SPC-m a e m
5. Get the above (title) message error (in fact, mu4e command is not available)

Emacs 30.1, EndeavourOS fully updated (mu4e from aur, rebuilt just in case).

bcc32:
Fix #16931.

Rather than just reverting the previous change and loading mu4e at
startup, just load its autoloads.  Warn if they were not already
loaded by `package-activate`, to encourage users to fix issues in
their Emacs site-lisp distribution or report issues to their
distribution maintainers.

### DIFF
--- a/layers/+email/mu4e/packages.el
+++ b/layers/+email/mu4e/packages.el
@@ -49,6 +49,17 @@
           (persp-kill mu4e-spacemacs-layout-name))))))
 
 (defun mu4e/init-mu4e ()
+  ;; Some distributions do not construct the initial Emacs load paths properly
+  ;; so that packages are included in `package-directory-list' and thereby
+  ;; activated at startup.  To avoid breakage, now that Spacemacs defers loading
+  ;; mu4e, load `mu4e-autoloads' explicitly and issue a warning.
+  ;;
+  ;; See https://github.com/syl20bnr/spacemacs/issues/16931
+  (unless (featurep 'mu4e-autoloads)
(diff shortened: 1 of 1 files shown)


fnussbaum:
As in the factor case, some distributions (I noticed this on Ubuntu 22.04) do not even provide the autoloads file. Should we fall back to eager loading in that case? Or display a different warning, perhaps even mentioning how to generate missing autoloads: `(loaddefs-generate mu4e-installation-path "mu4e-autoloads.el")`.

bcc32:
If the distribution does not provide the autoloads file, it's unlikely the user can generate the missing autoloads with `loaddefs-generate` (usually the install directories will be read-only, right?  It's actually been a long time since I regularly used a non-NixOS Linux distribution so I don't know off-hand).

fnussbaum:
Ah indeed, one would need to do something like `(find-file mu4e-installation-path)` and <kbd>SPC f E</kbd> first, only then `loaddefs-generate` succeeds.

(Of course, this only works with root access, so perhaps we should fall back to `(require 'mu4e)` if `mu4e-autoloads` is not found.)

fnussbaum (approved):

Issue #16925, opened 2025-03-28
Unable to send mail via mu4e


Geegee:
#### Description :octocat:
<<DESCRIPTION OF THE PROBLEM>>
Recently the problem occured that when I try to send a mail
(I used the exact setup for many months now to send mails).

```
Sending... Sending via mail... Delivering message to xxx@xxx.com... Saving file /home/xxx/email/xxx/Sent/cur/1743082566.ddaefc8a0eb58423.dellxps:2,S... Wrote /home/xxx/email/xxx/Sent/cur/1743082566.ddaefc8a0eb58423.dellxps:2,S Sending...done error in process sentinel: async-when-done: Invalid read syntax: "#<" error in process sentinel: Invalid read syntax: "#<"
When sending the message to my own mail addresses I get nothing,
the message reaches the sentbox though of the mailserver ( I can
see it if I log into the webinterface of the provider).
```

I also tried to send myself a mail more barebones with
the function smtpmail-via-smtp which works.

I use (setq message-send-mail-function 'smtpmail-send-it) to set the mail sending
function. The code until smtpmail-via-smtp is called is above my lisp level.

<<BE AWARE YOU NEED A GITHUB ACCOUNT TO SEND THIS REPORT>>

#### Reproduction guide :beetle:
- Start Emacs
- start mu4e
- Compose a mail
- CTRL-C CTRL-C to send the mail

*Observed behaviour:* :eyes: :broken_heart:
The mail is not sent but lands in the sent folder
of the mailbox (even when I check the mailbox via the web api of
the provider)

*Expected behaviour:* :heart: :smile:
I would receive the mail with my other mail address.

#### System Info :computer:
- OS: gnu/linux
- Emacs: 30.1
- Spacemacs: 0.999.0
- Spacemacs branch: develop (rev. da5db88ad)
- Graphic display: t
- Running in daemon: nil
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(toml javascript csv sql deft octave
      (llm-client :variables llm-client-enable-gptel t)
      (spell-checking :variables spell-checking-enable-auto-dictionary t) elixir
      epub ranger systemd html pdf yaml
      (rust :variables lsp-rust-analyzer-cargo-auto-reload t
            rustic-format-on-save t)
      graphviz djvu haskell lsp
      (tree-sitter :variables tree-sitter-syntax-highlight-enable t
[...]

PR #16935, opened 2025-04-01
Removing the subheading describing how to use async mode.

Geegee:
According to issue #16925 the feature is broken and therefore should not be
in the documentation.

### DIFF
--- a/layers/+email/mu4e/README.org
+++ b/layers/+email/mu4e/README.org
@@ -13,7 +13,6 @@
   - [[#compose-mode][Compose mode]]
 - [[#configuration][Configuration]]
   - [[#multiple-accounts][Multiple Accounts]]
-  - [[#async-mode][Async mode]]
   - [[#attachment-directory][Attachment directory]]
   - [[#example-configuration][Example configuration]]
   - [[#notifications][Notifications]]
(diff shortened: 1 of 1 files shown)


bcc32:
I think removing the section entirely is not the best way to remedy this.  We should instead have an explicit warning in the README, and also in the docstring of the variable itself.

(If the option is simply undocumented, it looks like somebody just forgot to add it to the README and is an "attractive nuisance")

Geegee:
Alright will do it. The other option to me as you previously mentioned would be to remove the
variable completely, but it probably originates from the underlying package, so spacemacs can not
do anything about it. Can a layer fix a variables value?

bcc32:
No, this isn't an underlying mu4e variable.  If you look at
`layers/+email/mu4e/packages.el` you'll see that enabling
`mu4e-enable-async-operations` just causes the following code, which
is part of Spacemacs, to run at startup:

``` emacs-lisp
(require 'smtpmail-async)
(setq send-mail-function         'async-smtpmail-send-it
      message-send-mail-function 'async-smtpmail-send-it)
```

So the option I mentioned before would be to remove the variable and
also the code that is gated by that variable.

bcc32:
Thanks.  For future readers, this PR was superseded by #16941.

PR #16938, opened 2025-04-02
Rework treatment of built-in packages

fnussbaum:
This should prevent frequent errors after updating packages, caused by outdated versions of built-in packages being loaded. We also ensure transient and org get updated, without having to manually bump the `:min-version` parameter. More generally, this PR allows installing built-in packages from arbitrary locations, for example by adding them to `dotspacemacs-additional-packages` with a custom quelpa recipe.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -1781,6 +1781,11 @@ RNAME is the name symbol of another existing layer."
       ;; installation
       (when upkg-names
         (spacemacs-buffer/set-mode-line "Installing packages..." t)
+        ;; Prevent built-in org from being loaded when updating consult, for example.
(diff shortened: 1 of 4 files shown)


fnussbaum:
This is now ready for review. I had no problems with updating packages with these commits in the last months, whereas before I frequently encountered broken package installations. I also vaguely remember quite a few reported issues that were related or caused by these load- and installation-order subtleties.

fnussbaum:
@bcc32 @smile13241324 (There is no rush, take your time, just pinging again in case the notification got lost.)

smile13241324:
Whups this got lost i'll do a test today, thsnks for pinging :) @fnussbaum

smile13241324 (approved):
Looks good

PR #16939, opened 2025-04-02
* doc/FAQ.org: Document for optimizing Spacemacs startup

sunlin7:
Document for optimizing the Spacemacs by customize the variable `load-suffixes` and `load-file-rep-suffixes`.
On my local, Spacemacs/Emacs-30.1 startup time on Windows will be reduced from ~2.2 seconds to ~1.5 seconds.
> 353 packages loaded in 1.529s (e:266 r:19 l:5 b:63)

### DIFF
--- a/doc/FAQ.org
+++ b/doc/FAQ.org
@@ -55,6 +55,7 @@
   - [[#why-are-all-packages-unavailable][Why are all packages unavailable?]]
   - [[#the-powerline-isnt-shown-correctly-when-spacemacs-is-used-within-putty][The powerline isn't shown correctly when Spacemacs is used within =PuTTY=]]
   - [[#speedup-the-spacemacs-with-load-hints-feature][Speedup the Spacemacs with load-hints feature]]
+  - [[#speedup-the-spacemacs-by-optimizing-module-loading][Speedup the Spacemacs by optimizing module loading]]
 
 * Common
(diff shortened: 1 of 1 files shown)


fnussbaum:
Interesting, thanks! I'm a little bit hesitant to add this to our FAQ though, because I imagine users adding this to their dotfile, forgetting about it, and later installing some package that relies on dynamic modules. So I fear this might result in issues that are difficult to debug.

bcc32:
I'm inclined to agree with the hesitation.  Perhaps at least, we should include a strong warning that certain packages that Spacemacs uses **do** in fact use dynamic modules (vterm).

(I know that vterm is not enabled for Windows builds, but users may simply read the FAQ and not realize they are in a section that should **only** apply to Windows.  I made the same mistake when reviewing this PR)

sunlin7:
@fnussbaum  @bcc32 
Thanks for your comments, and sorry for late response, totally understand your concerns.  And they're narrow so I just add them to FAQ.org to let other Spacemacs users know the options.
I change the example code snippets to give message to users who want try the options to understand the impacts.

smile13241324 (approved):
Looks good

sunlin7:
Thank you so much!

bcc32:
Ah, sorry for letting this fall off my plate.  Looks reasonable to me as well.

PR #16943, opened 2025-04-04
[ivy] Remove redundant C-h key binding for counsel-find-file

bcc32:
This is already covered by `spacemacs//ivy-hjkl-navigation` which remaps
`C-h` to `DEL` (via keyboard macro).

### DIFF
--- a/layers/+completion/ivy/packages.el
+++ b/layers/+completion/ivy/packages.el
@@ -161,11 +161,6 @@
     (dolist (command '(counsel-org-goto counsel-imenu spacemacs/counsel-jump-in-buffer))
       (evil-add-command-properties command :jump t))
 
-    (when (or (eq 'vim dotspacemacs-editing-style)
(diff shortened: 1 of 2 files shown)


fnussbaum:
It seems that with this change, typing something into the <kbd>SPC f f</kbd> prompt first, and then pressing <kbd>C-h</kbd> would only delete a single character. I would probably prefer the previous behaviour, but I guess it would be better to do this inside `spacemacs//ivy-hjkl-navigation`. What do you think?

bcc32:
That's true.  I guess I don't have a preference between them---I mildly like the new behavior for consistency with `C-h = DEL` in other minibuffer prompts, but I can make the change.

fnussbaum:
Thank you! I like that the behaviour stays consistent with the compleseus and helm find-file prompts, but changing it would also be fine.

fnussbaum (approved):

PR #16946, opened 2025-04-06
[modeline] Fix error when using vanilla modeline and refactor

fnussbaum:
The call to `spacemacs//restore-buffers-powerline` in `dotspacemacs/sync-configuration-layers` would lead to an error when using the `spacemacs-modeline` layer and the vanilla mode line.

A fix for #16346 would solve this, but in any case, this rather belongs in spaceline package init function.

Also factor out the spaceline toggle condition into `spacemacs//enable-spaceline-p` as it is now used in three places.

### DIFF
--- a/layers/+spacemacs/spacemacs-modeline/funcs.el
+++ b/layers/+spacemacs/spacemacs-modeline/funcs.el
@@ -50,6 +50,10 @@ Return nil if no scale is defined."
 
 ;; spaceline
 
+(defun spacemacs//enable-spaceline-p ()
+  (memq (spacemacs/get-mode-line-theme-name)
+        '(spacemacs all-the-icons custom)))
+
 (defun spacemacs/spaceline-config-startup-hook ()
   "Install a transient hook to delay spaceline config after Emacs starts."
   (spacemacs|add-transient-hook window-configuration-change-hook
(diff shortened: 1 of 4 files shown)


layers/+spacemacs/spacemacs-modeline/funcs.el
```diff
@@ -81,13 +85,14 @@ Return nil if no scale is defined."
...
-  (powerline-set-selected-window)
-  (force-mode-line-update t))
+  (when (spacemacs//enable-spaceline-p)
+    (dolist (buffer (buffer-list))
+      (unless (string-match-p "\\*which-key\\*" (buffer-name buffer))
+        (with-current-buffer buffer
+          (setq-local mode-line-format (default-value 'mode-line-format)))))
+    (powerline-reset)
+    (powerline-set-selected-window)
+    (force-mode-line-update t)))
```
bcc32, line 101: If I understand correctly, checking `spacemacs//enable-spaceline-p` shouldn't be necessary here because this function is now only called in `spacemacs-modeline/init-spaceline` so `enable-spaceline-p` will always return non-nil.

fnussbaum: I was concerned about the possibility of changing `dotspacemacs-mode-line-theme` without restarting, only using `SPC f e R`. This currently works for most combinations, except changing from something else to `vanilla` (which doesn't restore `mode-line-format`). I have pushed a small revision.

It probably wouldn't hurt to keep `spacemacs//restore-buffers-powerline` in the hook, I'm just removing it now for consistency. Also, it might make sense to reset mode-line-format for non-powerline mode lines as well, but I don't really understand why we do that anyway, so I'd rather leave it as is for now.

bcc32: Ah, I always forget about `SPC f e R`—because it never works reliably enough for me to trust it :)

bcc32 (approved):

PR #16950, opened 2025-04-09
Update docs to recommend using golangci-lint version 2.x

liuchong:
Update docs to recommend using golangci-lint version 2.x, in the golang layer


bcc32:
Thank you!

PR #16952, opened 2025-04-11
[core] Catch errors in package init functions and when loading files

fnussbaum:
(We already similarly catch errors in pre-init and post-init functions.)

Errors can happen, for example, when accidentally introducing a typo in some (private or public) layer file; or when adding the mu4e layer without having the mu4e package installed (or mu4e-autoloads.el is missing).

Such errors currently completely break the startup process, such that most keybindings don't work, which makes it more tedious to actually solve the problems.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -1572,7 +1572,7 @@ RNAME is the name symbol of another existing layer."
                     (spacemacs-customization//validate
                      val (custom-variable-type var)))
                   (set-default var val))
-              ('error
+              (error
                (configuration-layer//error
                 (concat "An error occurred while setting layer "
                         "variable %s "
@@ -1689,7 +1689,7 @@ RNAME is the name symbol of another existing layer."
             (oset pkg :lazy-install nil))
            (t (configuration-layer//warning "Cannot install package %S."
                                             pkg-name)))
-        ('error
+        (error
          (configuration-layer//error
           (concat "An error occurred while installing %s " "(error: %s)")
           pkg-name
@@ -2004,7 +2004,7 @@ LAYER must not be the owner of PKG."
             (format "%S -> pre-init (%S)..." pkg-name layer))
            (condition-case-unless-debug err
                (funcall (intern (format "%S/pre-init-%S" layer pkg-name)))
-             ('error
+             (error
               (configuration-layer//error
                (concat "An error occurred while pre-configuring %S "
                        "in layer %S (error: %s)")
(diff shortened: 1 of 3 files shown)


core/core-configuration-layer.el
```diff
@@ -2800,7 +2806,12 @@ ARGS: format string arguments."
 
 (defun configuration-layer/load-file (file &optional noerror)
   "Load file silently except if in debug mode."
-  (load file noerror (not init-file-debug)))
+  (condition-case-unless-debug err
+      (load file noerror (not init-file-debug))
+    ('error
+     (configuration-layer//error
+      "An error occurred while loading %S (error: %s)"
+      file err))))
```
bcc32, line 2814: I'm not convinced this change is a good idea.  `configuration-layer/load-file` is used in several places that IMO should simply propagate any errors loading the file.  For example in `configuration-layer/rollback` and `configuration-layer/load-auto-layer-file` (and maybe `configuration-layer/make-layer`).

It is also unfortunately a "public" function based on naming convention, so one more reason to maybe change the callsites rather than the function itself (although I consider this a weaker reason).

core/core-configuration-layer.el
```diff
@@ -2018,7 +2018,13 @@ LAYER must not be the owner of PKG."
          (owner (car (oref pkg :owners))))
     ;; init
     (spacemacs-buffer/message (format "%S -> init (%S)..." pkg-name owner))
-    (funcall (intern (format "%S/init-%S" owner pkg-name)))))
+    (condition-case-unless-debug err
+        (funcall (intern (format "%S/init-%S" owner pkg-name)))
+      ('error
```
bcc32: Byte-compilation recommends against quoting condition names

fnussbaum: Thanks, fixed.

core/core-configuration-layer.el
fnussbaum: > configuration-layer/load-file is used in several places that IMO should simply propagate any errors loading the file.

In `configuration-layer/rollback` we should definitely propagate the error: State from a previous invocation could be left in `update-packages-alist`, which would execute those previous rollbacks again.

In `configuration-layer/load-auto-layer-file` and `configuration-layer/make-layer` I don't immediately see the necessity to propagate the errors. Do you think something bad could happen if we continue in case of errors? In particular, could failing to load a single file be realistically worse than not loading any other file after an error? I guess it could be, but I'm not convinced yet.

Issue #16871, opened 2025-03-08
Package installation error is ignored despite --debug-init


bcc32:
If package-installation fails (at least from quelpa recipes, I didn't test other packages), the error is not caught by `--debug-init` and instead startup proceeds as normal.  This seems bad, `--debug-init` should start the debugger when the package installation encounters an error rather than later down the line when the package fails to load.

For example, I deliberately uninstalled and broke my local copy of `evil-evilified-state`, then tried starting `emacs --debug-init`.  The error appears in `*Messages*` but does not trigger the debugger, which only gets tripped when a later step attempts to load the package and fails.

```
(Spacemacs) --> installing package: evil-evilified-state@spacemacs-bootstrap... [1/1]
Fetcher: file
Source: nil

(Spacemacs) Error: 
An error occurred while installing evil-evilified-state (error: (error Single file evil-evilified-state-2.el does not match package name evil-evilified-state))
```

Similarly, when `use-package` fails to find a package to load it does not trigger the debugger despite `--debug-init`.

bcc32:
Doing some digging, it appears that errors raised inside `condition-case-unless-debug` (such as inside `configuration-layer//install-package`) do not trip the debugger with `--debug-init`.  I don't happen to have an Emacs 29 install handy right now so I'm not immediately sure if this is a regression in Emacs 30.1.  Seems surprising.

fnussbaum:
This indeed seems to be regression in Emacs 30. I've submitted a bug report upstream: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=76949.

fnussbaum:
@bcc32 How should we proceed here? I'm considering the following alternatives:
1. Make `--debug-init` set `debug-on-error` in Spacemacs. We could either adhere to (and thus ignore the errors in) `debug-ignored-errors`; or try to save its value, set it to nil, and restore it later, which Emacs used to do, see also [this Emacs bug](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=65267) (note that `handler-bind` was introduced for `--debug-init` to properly fix this bug).
If we don't set `debug-ignored-errors` to nil during startup by default, users might have to set it themselves in `dotspacemacs/user-init` or early-init.el for debugging purposes (or we provide another startup option that does that...). However, I'm not sure how often one actually needs to debug errors occuring during startup that are in the default value of `debug-ignored-errors`---at least I've never needed this.
[...]

PR #16953, opened 2025-04-11
[mu4e] Catch errors when loading org

fnussbaum:
An error within `(with-eval-after-load 'org ...)` has severe consequences: Org fails to load, and also help buffers cannot be opened at all, due to the elisp-demos package requiring org.

### DIFF
--- a/layers/+email/mu4e/packages.el
+++ b/layers/+email/mu4e/packages.el
@@ -155,21 +155,22 @@ See https://github.com/syl20bnr/spacemacs/issues/16931#issuecomment-2767608202."
 (defun mu4e/pre-init-org ()
   (when mu4e-org-link-support
     (with-eval-after-load 'org
-      ;; This is a dirty hack due to mu(4e) 1.8.2 renaming mu4e-meta to
-      ;; mu4e-config.  See also
-      ;; https://github.com/djcb/mu/commit/cf0f72e4a48ac7029d7f6758b182d4bb559f8f49
-      ;; and https://github.com/syl20bnr/spacemacs/issues/15618.  This code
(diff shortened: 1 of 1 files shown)


bcc32:
What error from mu4e did you see here?

Because of #16871 I'm a little wary of introducing uses of
`with-demoted-errors`; shouldn't the user just be made to see the
error and either debug it or disable `mu4e-org-link-support`?

fnussbaum:
This raises an error when mu4e is not properly installed, but the layer is added, for example.

> Because of #16871 I'm a little wary of introducing uses of
with-demoted-errors;

Fair point, but I'd actually like to catch even more errors (see #16952), at least on startup, where they often make Spacemacs completely unusable. I think having to set `debug-on-error` to get a backtrace in some more cases is an acceptable trade-off for having a usable editor, which can be used to investigate, solve, or report those startup errors. (For me personally, at least, encountering uncaught errors on startup has been a common source of frustration in the past.)

Also, in this case the error usually occurs (or can easily be reproduced) during a session, which means `toggle-debug-on-error` will produce a backtrace.

> shouldn't the user just be made to see the
error and either debug it or disable mu4e-org-link-support?

With this change, the user should still see the error; but I can make it even more visible by using `configuration-layer//error` which would make it show up on the Spacemacs buffer, if you prefer. I don't think we should "artificially" make errors more severe by letting them break unrelated functionality such as in this case.

bcc32:
Ah, right, I misunderstood and now realize that this error does not actually occur at startup but instead afterwards.  LGTM.

bcc32 (approved):

PR #16955, opened 2025-04-11
Update documentation for debugging errors on startup

fnussbaum:
Addresses #16871.


fnussbaum:
I guess the first part should also be changed: "If any errors happen during the loading the mode-line will turn red and the
errors should appear inline in the startup buffer." I've never seen the mode-line turn red, and not all errors appear in the Spacemacs buffer.

bcc32:
Merged with a small tweak (`delq` doesn't work if the element you're removing happens to be the first one.

PR #16957, opened 2025-04-16
[core] Enhance configuration-layer/display-summary output

Bost:
- Dynamically build the statistics string from the configuration data, removing hardcoded source keys.
- Replace ambiguous abbreviations (e, r, l, b) with full source names (elpa, recipe, local, built-in) and explain their meaning.
- Update time units from "s" to "seconds" for readability and consistency.

Before:
580 packages loaded in 683.673s (e:480 r:30 l:7 b:63)

After:
580 packages loaded in 683.673 seconds. Location counts: elpa 480, recipe 30, local 7, built-in 63

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -2448,15 +2448,15 @@ Return nil if MODE does not appear in `auto-mode-alist'."
     (with-current-buffer (get-buffer-create spacemacs-buffer-name)
       (let ((buffer-read-only nil))
         (spacemacs-buffer/append
-         (format "\n%s packages loaded in %.3fs (e:%s r:%s l:%s b:%s)"
+         (format "\n%s packages loaded in %.3f seconds. Location counts: %s"
                  (cadr (assq 'total stats))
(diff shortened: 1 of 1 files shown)


sunlin7:
NO comment on the changes.
Just for the startup time, is it really take **683** seconds for Spacemacs startup on your local?
> 580 packages loaded in 683.673s (e:480 r:30 l:7 b:63)

Bost:
No. I tweaked the function to see the message every time the `(configuration-layer/display-summary)` is evaluated and then I removed the tweak.

bcc32:
I'm curious: does anybody find the breakdown of packages by location, in the startup message, helpful?  I can't think of any reason I would care about this information if all of the packages are installed (obviously, if Spacemacs startup is slow because it needed to install some missing packages, that is another matter).

(I guess I would kind of rather know, e.g., how many packages were merely autoloaded vs. actually loaded, `(length features)`, things like that)

smile13241324 (approved):
I like that change, whether or not this message is useful, this change makes the meaning more clear to me.

PR #16958, opened 2025-04-17
Add --no-dotspacemacs and --default-dotspacemacs options

bcc32:
## Summary

This PR adds two command-line options for Spacemacs startup,
`--no-dotspacemacs` and `--default-dotspacemacs`.

`--no-dotspacemacs` inhibits loading the user's dotspacemacs file
(`dotspacemacs-filepath`).

`--default-dotspacemacs` inhibits loading the user's dotspacemacs
file, but also loads `dotspacemacs-template.el` when the user's
dotspacemacs would have been loaded.

In either case, take care to avoid deleting installed packages even if
the configuration would normally cause `dotspacemacs-install-packages`
to be `used-only`.

## Motivation

To simplify debugging Spacemacs issues that might be due to a bad
config, it is useful to have an easy way to suppress loading the
user's config and instead load Spacemacs's defaults.  This is
basically supposed to be like `emacs -q` except that `emacs -q` would
of course suppress loading Spacemacs itself.

Commands for working with, modifying, or visiting the user's
dotspacemacs are deliberately broken when these command-line options
are passed, to avoid confusion.  This is similar to how
customize-option behaves when `user-init-file` is nil (`emacs -q`).

### DIFF
--- a/layers/+completion/ivy/local/ivy-spacemacs-help/ivy-spacemacs-help.el
+++ b/layers/+completion/ivy/local/ivy-spacemacs-help/ivy-spacemacs-help.el
@@ -381,7 +381,7 @@ If EDIT is false, open org files in view mode."
 
 (defun ivy-spacemacs-help//go-to-dotfile-variable (candidate)
   "Go to candidate in the dotfile."
-  (find-file dotspacemacs-filepath)
+  (find-file (dotspacemacs/location))
   (goto-char (point-min))
   ;; try to exclude comments
   (re-search-forward (format "^[a-z\s\\(\\-]*%s" candidate))
--- a/core/core-command-line.el
+++ b/core/core-command-line.el
@@ -29,6 +29,14 @@
   "If non-nil packages are synchronized when the configuration layer system is
 loaded.")
 
+(defvar spacemacs-load-dotspacemacs t
+  "If nil, suppress loading the user's Spacemacs configuration.
+
+If set to `template', load `dotspacemacs-template.el' rather than the
+user's Spacemacs configuration.
+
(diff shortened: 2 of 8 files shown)


layers/+completion/ivy/local/ivy-spacemacs-help/ivy-spacemacs-help.el
```diff
@@ -381,7 +381,7 @@ If EDIT is false, open org files in view mode."
 
 (defun ivy-spacemacs-help//go-to-dotfile-variable (candidate)
   "Go to candidate in the dotfile."
-  (find-file dotspacemacs-filepath)
+  (find-file (dotspacemacs/location))
```
fnussbaum, line 384: This change seems to be missing for compleseus. (I can't push to this branch.)

fnussbaum:
Should we add this to the documentation somewhere? Looks good to me otherwise, thanks!

I currently use multiple separate Spacemacs installations for debugging, but this seems more convenient, at least when it is not necessary to add any non-default layers.

layers/+completion/ivy/local/ivy-spacemacs-help/ivy-spacemacs-help.el
bcc32: Thanks.  Not sure why you couldn't push.

bcc32:
> Should we add this to the documentation somewhere? Looks good to me otherwise, thanks!

Good idea, thanks.  Added to the FAQ, since the other command line args are documented w.r.t. "restarting" commands but not otherwise.

fnussbaum (approved):

Issue #16959, opened 2025-04-21
Emacs 31 started throwing 'lexical-binding' warnings at each boot


korayal:
#### Description :octocat:
Starting with newer versions of Emacs 31 (my current version is from git / 20250421) Emacs is throwing the below error when there's no `lexical-binding` cookie at the top of a core file.

#### Reproduction guide :beetle:
- Start Emacs

*Observed behaviour:* :eyes: :broken_heart:
throws below error at boot:
```
⛔ Warning (files): Missing `lexical-binding' cookie in "~/.emacs.d/early-init.el".
You can add one with ‘M-x elisp-enable-lexical-binding RET’.
See ‘(elisp)Selecting Lisp Dialect’ and ‘(elisp)Converting to Lexical Binding’
for more information.
⛔ Warning (files): Missing ‘lexical-binding’ cookie in "~/.emacs.d/init.el".
⛔ Warning (files): Missing ‘lexical-binding’ cookie in "~/.emacs.d/.lock".
⛔ Warning (files): Missing ‘lexical-binding’ cookie in "~/.emacs.d/layers/+tools/nginx/packages.el".
⛔ Warning (files): Missing ‘lexical-binding’ cookie in "~/.emacs.d/layers/+lang/clojure/packages.el".
⛔ Warning (files): Missing ‘lexical-binding’ cookie in "~/.emacs.d/layers/+lang/clojure/layers.el".
⛔ Warning (files): Missing ‘lexical-binding’ cookie in "~/.emacs.d/layers/+lang/sql/packages.el".
⛔ Warning (files): Missing ‘lexical-binding’ cookie in "~/.emacs.d/layers/+lang/sql/layers.el".
⛔ Warning (files): Missing ‘lexical-binding’ cookie in "~/.emacs.d/layers/+lang/typescript/packages.el".
⛔ Warning (files): Missing ‘lexical-binding’ cookie in "~/.emacs.d/layers/+lang/typescript/layers.el".
⛔ Warning (files): Missing ‘lexical-binding’ cookie in "~/.emacs.d/layers/+tools/systemd/packages.el".
⛔ Warning (files): Missing ‘lexical-binding’ cookie in "~/.emacs.d/layers/+lang/rust/packages.el".
⛔ Warning (files): Missing ‘lexical-binding’ cookie in "~/.emacs.d/layers/+lang/rust/layers.el".
[...]

PR #16960, opened 2025-04-21
Add missing lexical binding cookies

fnussbaum:
Addresses #16959.

For most files `lexical-binding: t` should also work without changes. But for each file, someone would need examine it and check whether dynamic binding is relied on somewhere. Hence I went with the current default of `lexical-binding: nil` for most files for now to pacify the warnings in Emacs 31.


bcc32 (approved):

korayal:
With this version, the only warning I'm now getting is from `~/.emacs.d/.cache/company-statistics-cache.el` which is auto-generated at boot :/

bcc32:
> With this version, the only warning I'm now getting is from `~/.emacs.d/.cache/company-statistics-cache.el` which is auto-generated at boot :/

Consider reporting [upstream](https://github.com/company-mode/company-statistics)?

fnussbaum:
> With this version, the only warning I'm now getting is from ~/.emacs.d/.cache/company-statistics-cache.el which is auto-generated at boot :/

@korayal Thanks for testing! Until this is fixed upstream, you could try something like the following in `dotspacemacs/user-init`:

```elisp
(advice-add
 'company-statistics--load :around
 (lambda (orig-fun)
   (let ((warning-inhibit-types '((files missing-lexbind-cookie))))
     (funcall orig-fun))))
```

korayal:
@fnussbaum that works perfectly, thanks!

PR #16961, opened 2025-04-21
[core] Allow selecting packages to update via CRM or embark

fnussbaum:
Previously, selecting specific packages to update meant having to answer `yes-or-no-p` for each one. While a `list-packages` type of interface might be even better, I think this should be a clear improvement.

However, making selection work via embark requires some gymnastics (not sure if this is controversial).

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -2082,8 +2082,11 @@ to update."
        (mapconcat #'symbol-name
                   configuration-layer--check-new-version-error-packages
                   " ")))
-    ;; (message "packages to udpate: %s" update-packages)
-    (when (> upgrade-count 0)
+    ;; (message "packages to update: %s" update-packages)
+    (if (eq upgrade-count 0)
+        (progn
+          (spacemacs-buffer/append "--> All packages are up to date.\n")
+          (spacemacs//redisplay))
       (spacemacs-buffer/append
        (format (concat "--> Found %s package(s) to update"
                        (if (> skipped-count 0)
@@ -2106,8 +2109,6 @@ to update."
           (if (string= answer "no")
               (progn (spacemacs-buffer/append "Packages update has been cancelled.\n" t)
                      (user-error "Packages update has been cancelled.\n"))
-            ;; backup the package directory and construct an alist
-            ;; variable to be cached for easy update and rollback
             (when (string= answer "some")
               (setq update-packages
                     ;; 'apply nconc on list of lists' is equivalent to 'cl-remove-if nil'
@@ -2116,13 +2117,13 @@ to update."
                                                (list pkg)))
                                            update-packages))))
             (setq upgrade-count (length update-packages)))))
-      (configuration-layer//update-packages update-packages))
-    (when (eq upgrade-count 0)
-      (spacemacs-buffer/append "--> All packages are up to date.\n")
-      (spacemacs//redisplay))))
+      (configuration-layer//update-packages update-packages))))
 
 (defun configuration-layer//update-packages (update-packages)
-  "Back up and delete packages in UPDATE-PACKAGES."
+  "Back up and delete packages in UPDATE-PACKAGES.
+
+Keep records in `update-packages-alist', and store the alist in a file
+in the back-up directory."
   (let* ((date (format-time-string "%y-%m-%d_%H.%M.%S"))
          (rollback-dir (expand-file-name
                         (concat configuration-layer-rollback-directory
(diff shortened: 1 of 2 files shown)


core/core-configuration-layer.el
```diff
@@ -2102,64 +2099,88 @@ to update."
...
+            ("some"
+             ;; We define `configuration-layer//select-packages-to-update' as an
+             ;; interactive function taking a single argument, such that embark
+             ;; users have the option to use `embark-select' and
+             ;; `embark-act-all' instead of CRM. Note that we capture
+             ;; `update-packages' in the lexical environment of the closure, in
+             ;; order to avoid introducing another global variable (and because
+             ;; embark constrains the function to a single argument).
+             (require 'crm) ; for `crm-separator'
+             (fset 'configuration-layer//select-packages-to-update
```
bcc32: Could this maybe use `cl-letf` with `symbol-function` as a generalized variable, so that the symbol doesn't have a leftover function definition once this is done?

core/core-configuration-layer.el
```diff
@@ -2102,64 +2099,88 @@ to update."
...
+                   (lambda (selected-packages)
+                     (interactive (list (completing-read-multiple
+                                         (concat "Packages to update"
+                                                 (when (string-match-p crm-separator ",") " (comma separated)")
```
bcc32: Is this necessary?  Spacemacs already has advice for updating the prompt for completing-read-multiple (see `crm-indicator`).  And besides, we're prompting for a plural.

core/core-configuration-layer.el
```diff
@@ -2102,64 +2099,88 @@ to update."
...
+                                                 ": ")
+                                         update-packages nil t)))
+                     (configuration-layer//update-packages
+                      (mapcar #'intern selected-packages))))
+             (let ((this-command 'configuration-layer//select-packages-to-update))
```
bcc32: The `this-command` thing is mysterious, perhaps it deserves a comment.

bcc32 (commented):
Thanks for keeping these commits separated, this was pretty easy to review.

One question, maybe subjective: I've been doing squash merges for most PRs but perhaps doing a rebase-then-non-ff-merge is warranted for more complicated changes?  I see that Spacemacs has mostly had a linear history but I already broke that once recently.  Opinions?  @smile13241324, do you have an opinion on squashing-vs-merging?

smile13241324:
Not really @bcc32 our convention says we prefer rebasing over merging however we could change that. I personally think that sometimes not squashing all prs will make things more readable especially if it is not just micro changes.

bcc32:
If merge commits are not desired we can always rebase the PR but without squashing, and then fast-forward into the `develop` branch.

fnussbaum:
Good idea, I would prefer not to squash sometimes, too.

core/core-configuration-layer.el
fnussbaum: Unfortunately, the function definition needs to extend beyond the scope of `configuration-layer/update-packages`, because embark-act-all calls the function from a post-command-hook.

(But I actually mildly liked being able to call `configuration-layer//select-packages-to-update` again to immediately update further packages that I missed before, without having to wait for the archives getting refreshed and all of the quelpa checkouts.)

core/core-configuration-layer.el
fnussbaum: Right, so perhaps this should rather be part of this advice (and the advice moved to spacemacs-defaults?). I'll consider creating a separate PR for this.

(It seems that `completing-read-multiple` is used only once within Spacemacs (and I don't think it is commonly used in packages, either), and it is not obvious how to use it the first time. I remember that this comma-separated indication helped me in the past.)

core/core-configuration-layer.el
fnussbaum: Sure, added one now.

core/core-configuration-layer.el
bcc32: > Unfortunately, the function definition needs to extend beyond the scope of `configuration-layer/update-packages`, because embark-act-all calls the function from a post-command-hook.
Got it, thanks.

> being able to call `configuration-layer//select-packages-to-update` again to immediately update further packages that I missed before

IMO, this seems to argue in favor of just defining the command globally, and just using a global variable to stash the packages-to-update state.

Defining the command using `fset` makes `find-function` not work.

fnussbaum: > IMO, this seems to argue in favor of just defining the command globally, and just using a global variable to stash the packages-to-update state.

Makes sense, I have pushed this, along with some minor changes.

bcc32 (approved):
Thanks, added a small comment but otherwise LGTM.

bcc32:
(Merged without squashing)

PR #16970, opened 2025-04-24
* layers/+spacemacs/spacemacs-modeline/packages: Use "spaceline" for use-package

sunlin7:
The `spaceline-config` is **NOT** a package name, it's a feature in the package `spaceline`. 
This PR will use the correct package name `spaceline` for `use-package`.

### DIFF
--- a/layers/+spacemacs/spacemacs-modeline/funcs.el
+++ b/layers/+spacemacs/spacemacs-modeline/funcs.el
@@ -100,7 +100,7 @@ Excluding which-key."
                  #'spacemacs//restore-buffers-powerline)))
 
 (defun spacemacs//prepare-diminish ()
-  (when spaceline-minor-modes-p
+  (when (bound-and-true-p spaceline-minor-modes-p)
     (let ((unicodep (dotspacemacs|symbol-value
                      dotspacemacs-mode-line-unicode-symbols)))
       (setq spaceline-minor-modes-separator
--- a/layers/+spacemacs/spacemacs-modeline/packages.el
+++ b/layers/+spacemacs/spacemacs-modeline/packages.el
@@ -48,7 +48,8 @@
     (setq-default fancy-battery-show-percentage t)))
 
 (defun spacemacs-modeline/init-spaceline ()
-  (use-package spaceline-config
+  (use-package spaceline
+    :commands (spaceline-compile spaceline-define-segment)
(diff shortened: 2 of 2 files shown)


fnussbaum (changes requested):
I think we still need to require `spaceline-config` somewhere to be safe, otherwise this can break custom mode line configurations, in particular [this example from the documentation](https://github.com/syl20bnr/spacemacs/blob/0008f83b101df0acec99bb5dce804823146c0fb8/doc/DOCUMENTATION.org?plain=1#L1564). (`spaceline-config` is autoloaded through `spaceline-helm-mode` and `spaceline-info-mode`, so this problem only shows up when not using helm and info+.)

sunlin7:
Thank you for the comment, I'll research to document and update again.

sunlin7:
@fnussbaum  I had pushed the changes with `:commands`, and verified with the spaceline `custom`  in my local, please help review again. Thank you!
https://github.com/syl20bnr/spacemacs/blob/0008f83b101df0acec99bb5dce804823146c0fb8/doc/DOCUMENTATION.org?plain=1#L1564

fnussbaum:
> I had pushed the changes with :commands

Unfortunately this did not seem to fix the issue with my recipe above. I have pushed a small fix that should make the change fully backward compatible. As mentioned in the commit message, we have to load the complete package in most cases, and I don't think it is worth it to make exceptions here. Does this look okay to you?

fnussbaum (approved):

sunlin7:
@fnussbaum  Verified the changes on my local with a "custom" spaceline, works for me. Thank you!

sunlin7:
I had reproduced the issue and find the root cause, pushed the latest changes as the solution.

The issue happened for the function `spacemacs//prepare-diminish` check the variable `spaceline-minor-modes-p` but the variable actually was defined in the `spaceline-segments.el`, while this .el file was loaded by the `spaceline-config.el`. 

As my latest change, detecting the variable `spaceline-minor-modes-p` by the `bound-and-true-p` works. 

Please help review again. Thanks

smile13241324 (approved):

smile13241324:
Looks nice @sunlin7 however I think I would make a short manual test anyway.

smile13241324:
Merged manually

PR #16973, opened 2025-04-28
* layers/+source-control/version-control/config: Document for how to disabe diff-hl

sunlin7:
Minor document update for how to disable the diff light.


bcc32:
BTW the typical convention in Spacemacs is to prefix PR and commit message with `[version-control]` or whatever layer you're making changes to.  The whole file path is too verbose to see at a glance.

sunlin7:
Will do that on future PRs. Thank you!

PR #16974, opened 2025-04-28
[version-control] Support auto select diff-hl-margin mode

sunlin7:
I found a bug that `diff-hl` does not work for TUI, just enable the `version-control` in the `dotspacemacs-configuration-layers` then start Spacemacs in TUI:
```shell
$ emacs -nw # open a modified file in a git repo to verify the diff-hl does NOT work
$ emacs # the diff-hl will works with the graphic frame
```
It caused by the `diff-hl` works for the fringe (GUI) but does not work for the margin (TUI).

This PR will activate the `diff-hl-margin-local-mode` for the buffers in TUI.
NOTE: If a buffer is shared by a TUI frame and GUI frame, the `diff-hl` package only works on one of the frame.

### DIFF
--- a/layers/+source-control/version-control/README.org
+++ b/layers/+source-control/version-control/README.org
@@ -53,11 +53,11 @@ You can choose the side on which the diff appears (by default it's the right sid
 #+END_SRC
 
 To automatically enable diff margins in all buffers, set
-=version-control-global-margin=
+=version-control-margin=
 
 #+BEGIN_SRC emacs-lisp
   '(version-control :variables
-                    version-control-global-margin t)
+                    version-control-margin 'global)
 #+END_SRC
 
 ** Differences between margin packages
@@ -73,6 +73,10 @@ one over the other:
 | Extended VCS support (e.g. hg, svn) | X       | X          |
 | Stage hunks from buffer             |         | X          |
 | Dired support                       | X       |            |
+
+The =diff-hl= or =git-gutter-fringe= supports GUI/Fringe, while the
+=diff-hl-margin= or =git-gutter= supports the TUI/Margin. Check the variable
+=version-control-margin= for how Spacemace selecting the margin features.
 
 * Key bindings
 VC commands (not exhaustive):
--- a/layers/+source-control/version-control/config.el
+++ b/layers/+source-control/version-control/config.el
@@ -24,9 +24,18 @@
 (defvar spacemacs--smerge-ts-full-hint-toggle nil
   "Display smerge transient-state documentation.")
 
-(spacemacs|defc version-control-global-margin t
-  "If non-nil, will show diff margins globally."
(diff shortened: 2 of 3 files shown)


bcc32:
I'm pretty sad about the complexity of this workaround (advice, timers, doesn't work correctly with multiple emacsclients).  Can we instead just add a layer variable that users can configure to use the fringe vs. the margin, and document that the fringe doesn't work in the terminal?  (Plenty of other functionality in Spacemacs requires some small adjustments on the user's part in the terminal, especially key bindings)

sunlin7:
@bcc32 When the variable `version-control-margin` value `nil` is for fringe only, value `global` is for margin only. 
There is no solution to work with fringe and margin together. 
The maintainers/contributors may have deep experience on these, 
but for normal users it's a friendly solution in this PR.

bcc32:
Sorry, I pressed the wrong button in forge.

bcc32:
I still think the hackiness of this approach is not justified and that
we should instead just report this upstream and see if there can be a
more principled solution.  Will let the others weigh in, though.

fnussbaum:
> we should instead just report this upstream and see if there can be a more principled solution

Sounds good to me, see also https://github.com/dgutov/diff-hl/issues/212#issuecomment-1966734951 (not sure if this would be sufficient) @sunlin7

sunlin7:
@fnussbaum Thanks for letting me know the thread diff-hl#212, should sufficient if it committed in the repo and we use `diff-hl-margin` mode with the control variable to be t , but it does not commit yet, seems no plan for it.

smile13241324:
@sunlin7 I have provided a fork of diff-hl with the proposed changes from dgutov. Can you have another look and see if this helps with finishing this PR?

sunlin7:
@smile13241324 

It still **NOT** show changes flag on a TTY frame, the reproduce steps were described on my initial message. Thanks

And I confirmed it installed the `diff-hl` from your folk. 
```
(Spacemacs) --> installing package: diff-hl@version-control... [1/1]
Fetcher: github
Source: smile13241324/diff-hl
Cloning https://github.com/smile13241324/diff-hl.git to ~/.emacs.d/.cache/quelpa/build/diff-hl/
```

smile13241324 (approved):
Given that the proposed changes from the referenced pr are not sufficient I assume we have no choice than to accept the additional complexity.

The feature is really cool and I agree to @sunlin7 the new behaviour is what I would expect from a modern IDE.

sunlin7:
@smile13241324  Thank you so much, appreciate.
 Neither the git-gutter/git-gutter-fringe nor diff-hl/diff-hl-margin can cover the margin/fringe mode together, maybe need years to wait them to ready for both margin/fringe. So this patch is valuable on making Spacemacs better.

PR #16975, opened 2025-04-30
[core] Avoid duplicate calls to (spacemacs-buffer/insert-startup-lists) on startup

sunlin7:
There are duplicate calls to (spacemacs-buffer/insert-startup-lists) on Spacemacs startup.
The paths are:
spacemacs/startup-hook -> spacemacs-buffer//startup-hook --> spacemacs-buffer/insert-startup-lists
spacemacs/startup-hook ->  spacemacs-buffer/goto-buffer --> spacemacs-buffer/insert-startup-lists

This PR will avoid avoid the unnecessary calls to just call the function one time on the startup.

### DIFF
--- a/core/core-spacemacs.el
+++ b/core/core-spacemacs.el
@@ -343,7 +343,6 @@ defer call using `spacemacs-post-user-config-hook'."
      (when spacemacs--delayed-user-theme
        (spacemacs/load-theme spacemacs--delayed-user-theme
                              spacemacs--fallback-theme t))
-     (spacemacs-buffer//startup-hook)
      (configuration-layer/display-summary)
      (spacemacs/check-for-new-version nil spacemacs-version-check-interval)
      (spacemacs-buffer/goto-link-line)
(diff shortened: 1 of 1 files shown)


bcc32:
I think, the first time it is called, it doesn't actually display anything except warnings.  That seems intentional or at least harmless to me.

E.g., display warnings early on before the startup hook is called, just in case the startup hook doesn't get called due to an error, then display the remainder of the elements when Spacemacs is actually finished initializing.

sunlin7:
A previous change https://github.com/syl20bnr/spacemacs/commit/e71a6863#diff-ae3f5c3c015a35304b97cb2aba03896647b102f8dfd59f91553e3f7e3201ac3eL257-L264 is let the first call on `spacemacs-buffer/insert-startup-lists` to ignore the dotspacemacs-startup-lists for `is-org-loaded` is not t; then second call will display full list but re-display the warnings again.

bcc32:
I only see warnings appear once in the Spacemacs buffer.  Can you provide a config that has the warnings displayed twice in the buffer?

smile13241324 (approved):

smile13241324:
Looks like my workaround was not necessary after all :), thanks for double checking @sunlin7

sunlin7:
Previous change is necessary for using the org from elpa on Emacs-29; this change just postpone the function to after epla initialized :)

Thank you !

PR #16979, opened 2025-05-06
Fix consult narrow tabbing.

mijoharas:
Fix consult group tabbing feature.

When I tried to use the other one I got an error saying `consult--narrow-keys` was undefined. I don't know when it broke.

Use the `vertico-next-group` funcs instead.

### DIFF
--- a/layers/+completion/compleseus/README.org
+++ b/layers/+completion/compleseus/README.org
@@ -79,8 +79,6 @@ Delete the narrowing key again to return to all selections.
 | Key binding | Description                                      |
 |-------------+--------------------------------------------------|
 | ~<~         | View narrowing commands available                |
-| ~C-left~    | Cycle left through narrowing commands available  |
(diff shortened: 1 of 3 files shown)


mijoharas:
Actually, I just looked and we already have these available as `C-S-j` so just removing the feature.

bcc32:
`vertico-next-group` does not do the same thing as these commands.  Try `SPC b b` and compare: for me, `vertico-next-group` switches between files, buffers, and bookmarks; whereas the removed commands switch between:

- all
- workspace buffer
- window buffer
- hidden buffer
- modified buffer
- buffer
- file
- bookmark
- etc.

IMO we should just fix the commands instead. Please see #16981.

mijoharas:
Fair enough. I didn't use it enough that I was find with next group _or_ narrowing. (Also I assumed since no-one had noticed it was broken that removing was fine.)

I'll close this PR in favour of yours @bcc32 .

PR #16988, opened 2025-05-12
[version-control] Fix toggle diff-hl margin mode local/global

sunlin7:
Fix toggle the diff-hl margin mode local/global. Original code did not use the function properly. 
local margin on: (diff-hl-margin-local-mode) (diff-hl-update)
local margin off: (diff-hl-margin-local-mode -1) (diff-hl-update)
toggle global margin mode: (diff-hl-margin-mode)

### DIFF
--- a/layers/+source-control/version-control/funcs.el
+++ b/layers/+source-control/version-control/funcs.el
@@ -70,25 +70,27 @@
 (defun spacemacs/vcs-enable-margin ()
   (interactive)
   (let ((current-prefix-arg t))
-    (call-interactively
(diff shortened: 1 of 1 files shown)


bcc32:
It seems pretty hacky that these commands use `call-interactively` and `current-prefix-arg` anyway, does any of the other maintainers know why the code is written this way?  Could we maybe just rewrite these to turn on/off the appropriate minor modes the normal way?  (by just calling them as functions)

sunlin7:
I does not try the `git-gutter`, so keep the `call-interactively` for git-gutter. :)

PR #16989, opened 2025-05-12
[core] Avoid adding non-exists item into load-path for a local package

sunlin7:
Mark package `page-break-lines` to be built-in, similar to the https://github.com/syl20bnr/spacemacs/blob/85c03aef74b73300d3714d7dde30608b96c38575/layers/%2Bspacemacs/spacemacs-completion/packages.el#L34

Original code will create a non-exists entry `../[layers/+spacemacs/spacemacs-defaults/local/page-break-lines` in the `load-path`.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -1976,7 +1976,9 @@ RNAME is the name symbol of another existing layer."
                  (let* ((owner (configuration-layer/get-layer owner)))
                    (when owner (oref owner dir))))))
       (if dir
-          (file-name-as-directory (format "%slocal/%S/" dir pkg-name))
+          (and-let* ((path (format "%slocal/%S/" dir pkg-name))
+                     ((file-exists-p path)))
+            path)
         (configuration-layer//warning
          "Cannot find path location path for package %S." pkg-name)
         nil)))))
--- a/tests/core/core-configuration-layer-utest.el
+++ b/tests/core/core-configuration-layer-utest.el
@@ -20,7 +20,7 @@
 ;; You should have received a copy of the GNU General Public License
 ;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-
+(require 'ert-x)
 (require 'mocker)
 (require 'core-command-line)
 (require 'core-configuration-layer)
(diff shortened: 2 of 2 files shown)


bcc32:
I recently changed this in https://github.com/syl20bnr/spacemacs/commit/bbc604170b7b5e7cc375c618c545b1eb8ad52bdb because built in is documented to mean things that are built into Emacs itself.  Do we need to define a new category for spacemacs core libs?

sunlin7:
Hi @bcc32 I checked the "built-in", a new keyword maybe changed a lot to support just two packages `page-break-lines`, `ido-vertical-mode`. Mark them as built-in and give a comment behind them is a simple way (I admit it's not a elegant way).

sunlin7:
@bcc32 Okay, I'm going to add a new type called "core-lib" to be a clone of "built-in".  Give me some time.

bcc32:
The other thing we could do is just skip adding `load-path` for `local` if the directory does not exist.  I assume your concern is the performance impact of the unnecessary `load-path` entry?

bcc32:
Alternatively: can we just use page-break-lines as a normal package instead of having a copy imported into the Spacemacs repo?  It is used in the Spacemacs buffer but I think it would be possible to just use `page-break-lines` if it is installed and just use a line of underscore or star characters if not installed.

`ido-vertical-mode` could likely be replaced by `fido-vertical-mode` although the latter seems to have been added in 28.1 so maybe we'd have to wait a bit.

sunlin7:
Hi @bcc32 
> The other thing we could do is just skip adding load-path for local if the directory does not exist. I assume your concern is the performance impact of the unnecessary load-path entry?

Good idea, had pushed the change to skip the non-exists path. And yes,  it will affect the performance, especially on Windows.

> Alternatively: can we just use page-break-lines as a normal package instead of having a copy imported into the Spacemacs repo? It is used in the Spacemacs buffer but I think it would be possible to just use page-break-lines if it is installed and just use a line of underscore or star characters if not installed.

Will confirm and update later. Thanks

sunlin7:
> Alternatively: can we just use page-break-lines as a normal package instead of having a copy imported into the Spacemacs repo? It is used in the Spacemacs buffer but I think it would be possible to just use page-break-lines if it is installed and just use a line of underscore or star characters if not installed.

I create a PR to use the `page-break-lines` from elpa.  #16990

sunlin7:
And this PR now turn to detecting the path exists then add it into the load-path. 
Please help review it. Thanks

smile13241324 (approved):

PR #16994, opened 2025-05-15
[core] Fix spurious error installing site packages

### LINKED ISSUE #16991: Cannot install package `mu4e' from ELPA.  It is a site package and must be installed with your Emacs distribution
After rev:  d0f64709f, `[core] Consider site packages to be "distant"', Spacemacs print this error during startup:

```
Cannot install package `mu4e' from ELPA.  It is a site package and must be installed with your Emacs distribution
```

@bcc32

bcc32:
Fix #16991.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -345,7 +345,7 @@ is ignored."
 
 (cl-defmethod cfgl-package-distant-p ((pkg cfgl-package))
   "Return non-nil if PKG is a distant package (i.e. not built-in Emacs)."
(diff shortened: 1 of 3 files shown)


bcc32:
Alright, it seems my previous change to consider site packages
"distant" was not necessary in order to prevent them from being
spuriously uninstalled at startup.  And, it has undesirable downstream
consequences.

Updated this PR to revert those unnecessary changes and also add some
explanatory documentation so that I don't make this mistake again.

PR #16998, opened 2025-05-17
[core] Remove unnecessary require for pgp package

sunlin7:
Remove the unnecessary line for better performance. The `(require 'epg)` took 0.015s during Spacemacs startup on my local.
The `epg-*` functions were called in the `configuration-layer//stable-elpa-verify-archive` but this function will **NOT** be called after installation; it's called for install packages. So remove it by adding autoload lines for better performance.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -22,7 +22,6 @@
 
 
 (require 'cl-lib)
-(require 'epg)
 (require 'eieio)
 (require 'subr-x)
 (require 'package)
@@ -2693,6 +2692,8 @@ MSG is an additional message append to the generic error."
                    "Do you still want to install the stable ELPA repository ?")
            configuration-layer-stable-elpa-name
            reason)))
+
+(autoload 'epg-context-home-directory "epg") ; autoload it for it's in setf
 
 (defun configuration-layer//stable-elpa-verify-archive ()
   "Verify the downloaded stable ELPA repository archive.


core/core-configuration-layer.el
```diff
@@ -2694,6 +2693,13 @@ MSG is an additional message append to the generic error."
...
 
+;; autoload functions from epg package
+(autoload 'epg-context-home-directory "epg")
+(autoload 'epg-import-keys-from-file "epg")
+(autoload 'epg-verify-file "epg")
+(autoload 'epg-context-result-for "epg")
+(autoload 'epg-signature-status "epg")
+
 (defun configuration-layer//stable-elpa-verify-archive ()
   "Verify the downloaded stable ELPA repository archive.
```
bcc32, line 2700: Can we just do `(require 'epg)` in the first line of `configuration-layer//stable-elpa-verify-archive` instead?  It would let us avoid listing a whole bunch of functions that may go out of sync easily.

sunlin7: @bcc32  No, require 'epg in the function does not work for this case, it caused by the `epg-verify-file` is an eieio created function for accessing the epg structure member, which is not defined during `defun`,
Testing commands:
```
emacs -batch -l ert -l ~/.emacs -l tests/core/core-configuration-layer-ftest.el -f ert-run-tests-batch-and-exit
```
> Test test-stable-elpa-verify-archive--verification-failed backtrace:
    (void-function \(setf\ epg-context-home-directory\))

sunlin7:
@bcc32  I upload a simple one, please help review again. Thanks

bcc32:
Macro-expanding `setf` will actually load the autoload anyway, if the PLACE is autoloaded.  I am observing that with these changes, `epg` is still loaded.

I think it is not really possible to load core-configuration-layer without loading epg, because eager macro-expansion will need to load epg to learn how to correctly expand `(setf (epg-context-home-directory) ...)`.

Perhaps you could instead move all of the `stable-elpa` stuff into a separate file which can then just `(require 'epg)` normally, but that separate file isn't loaded until actually used.

sunlin7:
You're right, the epg still be opened and readed. And separating the stable-elpa to a new file is over-changes for this minor issue. 
Thanks a lot, I close this ticket for current implement works enough.

PR #17000, opened 2025-05-19
* layers/+misc/dtrt-indent/packages: Explicitly denote the lambda as function

sunlin7:
Got an error with latest Emacs master branch:
`error: (void-variable body)`
Mark the lambda as function explicitly will resolve the issue.


bcc32:
Did something change recently on the Emacs master branch, which caused this change to be necessary?  Do you mind providing a link or news item I can look at?

(I didn't think doing this was necessary.  If doing this is generally problematic, there are other occurrences of lambdas in use-package that need to be fixed.)

sunlin7:
Non reproducible on latest emacs, close this PR.

PR #17004, opened 2025-05-22
[github-copilot] Make commit message generation opt-in

### LINKED ISSUE #16999: Copilot commit message are annoying
I didn't find a way to disable github-copilot commit messages. They are annoying when you amend an existing commit because it deletes your existing text.
It would be nice to at least opt out of this feature.

bcc32:
Add a new layer variable, `github-copilot-enable-commit-messages`.
Only add `copilot-chat-insert-commit-message` to
`git-commit-setup-hook` if the variable is non-nil.

Close #16999.

### DIFF
--- a/layers/+web-services/github-copilot/packages.el
+++ b/layers/+web-services/github-copilot/packages.el
@@ -42,12 +42,14 @@
 (defun github-copilot/init-copilot-chat ()
   "Initialize the `copilot-chat' package and set up keybindings."
   (use-package copilot-chat
-    :hook (git-commit-setup . copilot-chat-insert-commit-message)
     :defer t
     :init
     ;; Provide our transient state in the AI menu
     (spacemacs/declare-prefix "$" "AI")
     (spacemacs/set-leader-keys "$c" 'copilot-chat-transient)
+    (when github-copilot-enable-commit-messages
+      (add-hook 'git-commit-setup-hook
+                #'copilot-chat-insert-commit-message))
     :config
     ;; Make sure that standard ,, works as confirm in the chat window
     (evil-define-key 'normal copilot-chat-prompt-mode-map ",," #'copilot-chat-prompt-send)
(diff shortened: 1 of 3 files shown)


layers/+web-services/github-copilot/packages.el
```diff
@@ -42,12 +42,14 @@
 (defun github-copilot/init-copilot-chat ()
   "Initialize the `copilot-chat' package and set up keybindings."
   (use-package copilot-chat
-    :hook (git-commit-setup . copilot-chat-insert-commit-message)
     :defer t
     :init
     ;; Provide our transient state in the AI menu
     (spacemacs/declare-prefix "$" "AI")
     (spacemacs/set-leader-keys "$c" 'copilot-chat-transient)
+    (when github-copilot-enable-commit-messages
```
phenix-factory: Setting this to `t` doesn't seem to restore the behavior? I used this:
```lisp
  (setq git-magit-status-fullscreen t
        magit-diff-refine-hunk nil
        package-install-upgrade-built-in t
        github-copilot-enable-commit-messages t)
```

bcc32: Where is that snippet, in `dotspacemacs/user-config`?  (If so, that is too late.  Layer variables, which are defined in `config.el`, need to be set [during layer configuration](https://www.spacemacs.org/doc/DOCUMENTATION.html#setting-configuration-layers-variables) using the `:variables` keyword).

phenix-factory: Well, my bad, I didn't understand that.
This is working flawlessly :)

bcc32:
Oops.  I pushed to develop before force-pushing this branch.  Anyway, it was actually merged, not closed.

phenix-factory:
> Oops. I pushed to develop before force-pushing this branch. Anyway, it was actually merged, not closed.

Thank you!

PR #17006, opened 2025-05-23
[core] Load the projects list without activating the projectile-mode

sunlin7:
Use the function `(projectile-known-projects)` to replace the function `(projectile-relevant-known-projects)` for correctly getting the project list.

This patch will:
1. don't active the `projectile-mode` in spacemacs-buffer
2. just load the projects by projectile function `projectile-known-projects` which was introduced in the commit https://github.com/bbatsov/projectile/commit/373fca78904330cc130604c9490c4e0785df7c4d.

The `projectile` is part of `+distributions/spacemacs/`:
https://github.com/syl20bnr/spacemacs/blob/cc9b605bc2d00cbb2ddeb5e026ba12deb435bbd2/layers/%2Bdistributions/spacemacs/layers.el#L43
And it's activated in the function `spacemacs-project/init-projectile`.

The `projectile-mode` is **NOT** part of `+distributions/spacemacs-base/`:
https://github.com/syl20bnr/spacemacs/blob/cc9b605bc2d00cbb2ddeb5e026ba12deb435bbd2/layers/%2Bdistributions/spacemacs-base/layers.el#L28
But it was activated in the `(spacemacs-buffer//insert-projects)` even user selected the `spacemacs-base` as `dotspacemacs-distribution`.

### DIFF
--- a/core/core-spacemacs-buffer.el
+++ b/core/core-spacemacs-buffer.el
@@ -1277,7 +1277,7 @@ LIST: list of `org-agenda' entries in the todo list."
     (when (string-prefix-p (car x) (expand-file-name recent-file))
       (setcdr x (cons (string-remove-prefix (car x) recent-file) (cdr x))))))
 
-(autoload #'projectile-known-projects "projectile")
+(autoload 'projectile-known-projects "projectile")
 (defun spacemacs-buffer//recent-files-by-project ()
   (let ((by-project (mapcar (lambda (p) (cons (expand-file-name p) nil))
                             (projectile-known-projects))))
--- a/core/core-spacemacs-buffer.el
+++ b/core/core-spacemacs-buffer.el
@@ -1277,9 +1277,10 @@ LIST: list of `org-agenda' entries in the todo list."
     (when (string-prefix-p (car x) (expand-file-name recent-file))
       (setcdr x (cons (string-remove-prefix (car x) recent-file) (cdr x))))))
 
+(autoload #'projectile-known-projects "projectile")
 (defun spacemacs-buffer//recent-files-by-project ()
   (let ((by-project (mapcar (lambda (p) (cons (expand-file-name p) nil))
-                            (projectile-relevant-known-projects))))
+                            (projectile-known-projects))))
     (dolist (recent-file recentf-list by-project)
       (spacemacs-buffer//associate-to-project recent-file by-project))))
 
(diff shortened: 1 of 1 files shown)


bcc32:
I'm not currently at my personal computer to test this, but:

If the user has selected `spacemacs-base` as the distribution, without enabling the `spacemacs-project` layer... is `projectile` even installed?  In that case, won't the spacemacs-buffer trigger an error?

If the above is correct, I feel like the better solution would be to skip the calls to `projectile-known-projects` rather than attempting to autoload it.

sunlin7:
@bcc32  Thanks for the comment. The `projectile` will be installed as dependency of `helm-porjectile` and `treemacs-projectile`, which are from the default value of `dotspacemacs-configuration-layers` included the `helm` and `treemacs`. 

If remove the `helm` and `treemacs` from layers, the `spacemacs-base` will not going to install the projectile.

Keep project list on Spacemacs startup buffer will provide consistent user experience.

bcc32:
I think if a user has gone out of their way to exclude the `spacemacs-project` layer, they are saying they don't want projectile-mode at all.  In that case, installing/loading it anyway seems unhelpful.

bcc32:
And it would be nice to take this opportunity to support Spacemacs starting up properly without projectile installed at all (e.g., if the user wants to exclude it in their dotfile)

sunlin7:
@bcc32  
The key point is, original function `(projectile-relevant-known-projects)` has two issue here:
1. will remove the project for current directory from the project list
2. require the `projectile-mode` to be enabled on the spacemacs-buffer.

This PR will use the function `(projectile-known-projects)`  which just load the project list without side effects.

sunlin7:
Got your points, mark this PR as draft first, will enhance it towards the comments.

sunlin7:
@bcc32 
It works **without** `projectile` package after setting the `dotspacemacs-distribution 'spacemacs-base` and `dotspacemacs-configuration-layers '()`.  I confirmed on my local. 

Before calling projectile function, it will detect projectile feature by `(fboundp 'projectile-mode)` on these lines:
https://github.com/syl20bnr/spacemacs/blob/7d0ddf020a989361981a5c55cd55bce0e65e07b0/core/core-spacemacs-buffer.el#L1450-L1453
https://github.com/syl20bnr/spacemacs/blob/7d0ddf020a989361981a5c55cd55bce0e65e07b0/core/core-spacemacs-buffer.el#L1456-L1459

So the changes works for projectile exists or projectile not exists cases.

bcc32:
Got it, thank you for explaning.

sunlin7:
Thank you!

PR #17007, opened 2025-05-23
[layers/+lang/python] Remove unused function, variable and key bindings

sunlin7:
Remove the unused function, variable and key bindings after the PR https://github.com/syl20bnr/spacemacs/pull/16867, which removed the `pyvenv`,


sunlin7:
Close for #17038 still need the lines.

PR #17011, opened 2025-05-28
[tests] Make the Makefile to scan the *-utest.el and *-ftest.el

sunlin7:
Minor enhance the Makefile to scan the *-utest.el and *-ftest.el.


bcc32 (approved):
Looks great, thank you!

PR #17019, opened 2025-06-04
[auto-completion] Fix function 'yas-hippie-try-expand' not defined on hippie-expand 

### LINKED ISSUE #17018: hippie-expand error with yasnippet
When `hippie-expand` (`M-/`) is called before `yasnippet` is loaded, for example before the user has opened a file which runs `prog-mode-hook`, an error is triggered: `hippie-expand: Symbol’s function definition is void: yas-hippie-try-expand`. For example, in the scratch buffer, type `test|` and press `M-/` with the cursor after the last character (represented by |). 

This is because `hippie-expand` tries the first function in `hippie-expand-try-functions-list`, which is `yas-hippie-try-expand`, which hasn't been loaded yet.

The following code in `dotspacemacs/user-config` fixes it for me, but I'm not sure where it should go for a fix in spacemacs.
```

  (setq hippie-expand-try-functions-list
        (remove 'yas-hippie-try-expand hippie-expand-try-functions-list))

  (with-eval-after-load 'yasnippet
    (add-to-list 'hippie-expand-try-functions-list 'yas-hippie-try-expand))

```

sunlin7:
Fix #17018

### DIFF
--- a/layers/+completion/auto-completion/packages.el
+++ b/layers/+completion/auto-completion/packages.el
@@ -249,10 +249,7 @@
           ;; unique.
           try-complete-lisp-symbol-partially
           ;; Try to complete word as an Emacs Lisp symbol.
-          try-complete-lisp-symbol))
-  (when (configuration-layer/package-used-p 'yasnippet)
-    ;; Try to expand yasnippet snippets based on prefix
-    (add-to-list 'hippie-expand-try-functions-list 'yas-hippie-try-expand)))
+          try-complete-lisp-symbol)))
 
 (defun auto-completion/init-ivy-yasnippet ()
   (use-package ivy-yasnippet
@@ -270,7 +267,7 @@
 
 (defun auto-completion/init-yasnippet ()
   (use-package yasnippet
-    :commands (yas-global-mode yas-minor-mode yas-activate-extra-mode)
+    :commands (yas-global-mode yas-minor-mode yas-activate-extra-mode yas-hippie-try-expand)
     :init
     ;; We don't want undefined variable errors
     (defvar yas-global-mode nil)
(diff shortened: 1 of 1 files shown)


layers/+completion/auto-completion/packages.el
```diff
@@ -270,7 +267,7 @@
 
 (defun auto-completion/init-yasnippet ()
   (use-package yasnippet
-    :commands (yas-global-mode yas-minor-mode yas-activate-extra-mode)
+    :commands (yas-global-mode yas-minor-mode yas-activate-extra-mode yas-hippie-try-expand)
```
bcc32, line 270: FYI :commands will make the autoloaded function "interactive" so it will show up in M-x, but this is wrong because it has no interactive form and will fail to be invoked this way.

layers/+completion/auto-completion/packages.el
```diff
@@ -284,6 +281,9 @@
     ;; disable yas minor mode map
     ;; use hippie-expand instead
     (setq yas-minor-mode-map (make-sparse-keymap))
+    (with-eval-after-load 'hippie-exp
+      ;; Try to expand yasnippet snippets based on prefix
+      (add-to-list 'hippie-expand-try-functions-list 'yas-hippie-try-expand))
```
bcc32, line 286: `hippie-expand-try-functions` is autoloaded (for historical reasons) so no need for `with-eval-after-load`

Also, IMO we can wait until `:config` to add this one.  It does nothing unless `yas-minor-mode` is enabled.

sunlin7:
@bcc32  Thanks for the comment, you're right, and the latest changes towards your comment. Please help review again.

sunlin7:
The github always show "Processing updates" on this PR, something went wrong. 
I try to close the ticket and re-open again.

sunlin7:
Move to #17021.

PR #17023, opened 2025-06-04
[python] Less condition for enabling pet-mode

sunlin7:
Currently Spacemacs only activates the `pet-mode` when the `dasel` and `sqlite3` installed.
But exists users may not notice this requirement, does not install these two commands, then pet-mode does not been activated automatically. 

The command `dasel` or `sqlite3` are required to parse the `pyproject.toml`, `Pipfile`, `environment*.y*ml`... in `pet-mode`.

**Without** the `dasel` or `sqlite3`, `pet-mode` still works for 
a) non-virtualenv case: there is no virtual envrionment, like a `~/proj1/test.py` works towards to `/usr/bin/python`.
b) simple local `.venv` case: `python3 -m venv ~/proj1/.venv`, then open the `~/proj1/test.py`.

So we can loosen the condition for activating the `pet-mode` to make it works on the two cases.

### DIFF
--- a/layers/+lang/python/README.org
+++ b/layers/+lang/python/README.org
@@ -354,9 +354,9 @@ This will allow you to use [[https://github.com/pwalsh/pipenv.el][pipenv]] bindi
 You can add to the other two lists analogously.
 
 * Management of Python versions and virtual environments
-In general virtual environments are automatically used when active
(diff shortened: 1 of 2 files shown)


bcc32 (approved):
I don't have the right environment to verify this, but it seems reasonable to me based on the description and a cursory reading of pet.el.

PR #17027, opened 2025-06-07
[layers/+lang/gleam] switch to gleam-ts-mode

### LINKED ISSUE #16977: +lang/gleam layer needs to be updated (tree-sitter)
[`gleam-mode`](https://github.com/gleam-lang/gleam-mode) has switched to using tree-sitter. There is only `gleam-ts-mode.el` in the repository, `gleam-mode.el` was removed in [Commit a734f84](https://github.com/gleam-lang/gleam-mode/commit/a734f847eadc34b0c9e8711c6ac190a226badb4d).

However, the [Gleam layer](https://github.com/syl20bnr/spacemacs/commits/develop/layers/%2Blang/gleam) hasn't incorporated this change. It tries to install the old `gleam-mode.el`, which no longer exists.

I tried to fix the layer locally, but it turned out to be more complicated than just changing every mention of `gleam-mode` to `gleam-ts-mode` in [`packages.el`](https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Blang/gleam/packages.el).

Also, requiring tree-sitter would raise the minimum version requirement for the layer to Emacs 29 (and compilation flag `--with-tree-sitter`), and the official minimum version for Spacemacs proper is 28.2.

The old `gleam-mode.el` is still available in [the Git history](https://github.com/gleam-lang/gleam-mode/blob/8656c4080dd2bb7dd6d6167953d6463d090509b0/gleam-mode.el).

I assume the layer now needs to check the Emacs version + the tree-sitter presence and either use the current `gleam-ts-mode.el` or the old `gleam-mode.el` from Git history? It also needs to automatically generate the tree-sitter grammar? (I have no experience working with tree-sitter.)

@2ynn, as the [original contributor](https://github.com/syl20bnr/spacemacs/pull/16427) of the Gleam layer I hope you don't mind being tagged.

### My current workaround

I ultimately failed to fix it via the layer, so right now I'm doing this.

I added `gleam-ts-mode` to `dotspacemacs-additional-packages` like this:
```elisp
   dotspacemacs-additional-packages
   '(
     (gleam-ts-mode :mode (rx ".gleam" eos)) ;; TODO: Check Gleam layer status
     )
```

To my `dotspacemacs/user-config` I added the following lines:
```
  (require 'gleam-ts-mode)
  (add-to-list 'auto-mode-alist '("\\.gleam\\'" . gleam-ts-mode))
  (add-hook 'gleam-ts-mode-hook #'lsp)
```

At first I got a few various errors, including «cannot activate tree-sitter because language grammar for gleam is unavailable». I ran `gleam-ts-install-grammar` as per `gleam-mode` [README.md](https://github.com/gleam-lang/gleam-mode?tab=readme-ov-file#install-the-tree-sitter-grammar). It generated `libtree-sitter-gleam.so` in `tree-sitter/` directory.

Now I have syntax highlighting, indentation, and LSP support, but no Gleam Major Mode leader key keybindings—only LSP-specific keybindings.

2ynn:
fix #16977 

This PR fixes the `gleam` layer  after [gleam-mode](https://github.com/gleam-lang/gleam-mode/blob/8656c4080dd2bb7dd6d6167953d6463d090509b0/gleam-mode.el) got deprecated in favor of [gleam-ts-mode](https://github.com/gleam-lang/gleam-mode/blob/main/gleam-ts-mode.el), which is now available on MELPA.

### DIFF
--- a/layers/+lang/gleam/README.org
+++ b/layers/+lang/gleam/README.org
@@ -18,7 +18,8 @@
   - [[#execution-1][Execution]]
 
 * Description
-This layers adds support for [[https://gleam.run/][Gleam]]. It relies on the official [[https://github.com/gleam-lang/gleam-mode][gleam-ts-mode]] package and [[https://github.com/emacsmirror/tree-sitter-indent][tree-sitter-indent]].
+This layers adds support for [[https://gleam.run/][Gleam]]. It relies on the official [[https://github.com/gleam-lang/gleam-mode][gleam-ts-mode]] package and requires Emacs to be compiled with treesit support (built-in with version 29+).
+To check that the treesit package is installed, you can run =M: (treesit-available-p)=.
 
 ** Features:
 - [[https://gleam.run/news/v0.21-introducing-the-gleam-language-server/][Gleam language server]] integration
--- a/layers/+lang/gleam/README.org
+++ b/layers/+lang/gleam/README.org
@@ -18,7 +18,7 @@
   - [[#execution-1][Execution]]
 
 * Description
-This layers adds support for [[https://gleam.run/][Gleam]]. It relies on the official [[https://github.com/gleam-lang/gleam-mode][gleam-mode]] package and [[https://github.com/emacsmirror/tree-sitter-indent][tree-sitter-indent]].
+This layers adds support for [[https://gleam.run/][Gleam]]. It relies on the official [[https://github.com/gleam-lang/gleam-mode][gleam-ts-mode]] package and [[https://github.com/emacsmirror/tree-sitter-indent][tree-sitter-indent]].
 
 ** Features:
 - [[https://gleam.run/news/v0.21-introducing-the-gleam-language-server/][Gleam language server]] integration
@@ -51,7 +51,7 @@ When enabled, the =gleam-lsp= server will be automatically initialized when visi
 More details can be found in the [[https://github.com/syl20bnr/spacemacs/tree/develop/layers/%2Btools/lsp#configuration][lsp layer configuration section]].
 
 ** Formatting
-=gleam-mode= provides the ~gleam-format~ command (~SPC m = =~) to format source code in the official Gleam style.
+=gleam-ts-mode= provides the ~gleam-ts-format~ command (~SPC m = =~) to format source code in the official Gleam style.
 
 To automatically apply formatting files before saving:
 
--- a/layers/+lang/gleam/packages.el
+++ b/layers/+lang/gleam/packages.el
@@ -22,34 +22,32 @@
 
 
 (defconst gleam-packages
-  '((tree-sitter-indent)
-    (gleam-mode :location (recipe
-                           :fetcher github
-                           :repo "gleam-lang/gleam-mode"
-                           :files ("*.el" "tree-sitter-gleam")))))
+  '((treesit)
+    (gleam-ts-mode)))
 
-(defun gleam/post-init-tree-sitter-indent ()
-  "Initialize tree-sitter-indent"
-  (use-package tree-sitter-indent
+(defun gleam/post-init-treesit ()
+  "Initialize treesit"
+  (use-package treesit
     :defer t))
 
-(defun gleam/init-gleam-mode ()
-  "Initialize gleam-mode"
-  (use-package gleam-mode
-    :mode ("\\.gleam\\'" . gleam-mode)
-    :hook ((gleam-mode . spacemacs//gleam-setup-format-on-save)
-           (gleam-mode . spacemacs//gleam-setup-lsp))
+(defun gleam/init-gleam-ts-mode ()
+  "Initialize gleam-ts-mode"
+  (use-package gleam-ts-mode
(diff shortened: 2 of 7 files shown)


sindikat:
It seems like the LSP and major-mode leader key commands work as intended. I think this can be merged.
Thank you so much, @2ynn!

layers/+lang/gleam/README.org
```diff
@@ -18,7 +18,7 @@
   - [[#execution-1][Execution]]
 
 * Description
-This layers adds support for [[https://gleam.run/][Gleam]]. It relies on the official [[https://github.com/gleam-lang/gleam-mode][gleam-mode]] package and [[https://github.com/emacsmirror/tree-sitter-indent][tree-sitter-indent]].
+This layers adds support for [[https://gleam.run/][Gleam]]. It relies on the official [[https://github.com/gleam-lang/gleam-mode][gleam-ts-mode]] package and [[https://github.com/emacsmirror/tree-sitter-indent][tree-sitter-indent]].
```
bcc32: It seems `tree-sitter-indent` has been removed for gleam, by this PR

layers/+lang/gleam/packages.el
```diff
@@ -22,34 +22,32 @@
 
 
 (defconst gleam-packages
-  '((tree-sitter-indent)
-    (gleam-mode :location (recipe
-                           :fetcher github
-                           :repo "gleam-lang/gleam-mode"
-                           :files ("*.el" "tree-sitter-gleam")))))
+  '((treesit)
```
bcc32, line 25: `treesit` is a built-in package so should be marked as such by the `:location` keyword (but see below about the post-init function)

layers/+lang/gleam/packages.el
```diff
@@ -22,34 +22,32 @@
...
+    (gleam-ts-mode)))
 
-(defun gleam/post-init-tree-sitter-indent ()
-  "Initialize tree-sitter-indent"
-  (use-package tree-sitter-indent
+(defun gleam/post-init-treesit ()
+  "Initialize treesit"
```
bcc32: `treesit` does not have an owning layer, so this `post-init-treesit` function will never get called.  Also, there's no useful configuration that is happening here, so this function should probably just be deleted.

bcc32 (commented):
Since tree-sitter requires Emacs 29+ and Spacemacs's minimum supported Emacs version is 28.2, perhaps it would be worth noting in the README that there is an extra requirement.

layers/+lang/gleam/packages.el
2ynn: thank you, I took the liberty to add `built-in` as a possible `:location` value in `doc/LAYERS.org`

2ynn:
Thank you for your thorough review @bcc32! I addressed your comments and it's ready for another look. Thanks!

layers/+lang/gleam/packages.el
smile13241324: Thanks for adding the hint to the docs. However I think you have mist adding the location to your package declaration.

smile13241324 (changes requested):
I think you either must add an init function for treesit or remove it from the layer.

Please also have a look about the locality of treesit this should be declared as build in if it is so

smile13241324:
Manually merged and revised.

smile13241324:
@2ynn can you have a quick look that the layer still works? I have not tried installing the gleam language server. But I assume some more information of how to do so in the readme would help new users. 

Alternatively the lsp server could be extended to support the auto install feature of lsp-mode.

2ynn:
> @2ynn can you have a quick look that the layer still works? 

Thanks @smile13241324 , I just checked and the layer works as expected!


> I have not tried installing the gleam language server. But I assume some more information of how to do so in the readme would help new users. Alternatively the lsp server could be extended to support the auto install feature of lsp-mode.

`gleam lsp` comes [pre-installed](https://gleam.run/language-server/#installation) with the `gleam` binary. To my knowledge there is no extra installation step needed than what is described in the README.

**Note**: I did however just realized there is a typo in the [instruction to enable lsp mode](https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Blang/gleam/README.org?plain=1#L47) when the `lsp` layer is not yet installed, it should be `t`, not `nil` :/
```
  (gleam :variables gleam-enable-lsp t)
```

As for the tree-sitter grammar, it gets [automatically installed](https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Blang/gleam/packages.el#L36) through the [gleam-ts-install-grammar](https://github.com/syl20bnr/spacemacs/blob/develop/layers/%2Blang/gleam/funcs.el#L36)  function [provided by gleam-ts-mode](https://github.com/emacsmirror/gleam-ts-mode/blob/main/gleam-ts-mode.el#L297)

layers/+lang/gleam/packages.el
2ynn: 🤦‍♀️

PR #17030, opened 2025-06-09
[tests] add python layer test case

sunlin7:
And a test case for python layer


bcc32:
I'm concerned about having tests which depend on automatically installing external packages, especially because we have no way of pinning packages to a specific version yet.  That would make the tests not deterministic/reproducible at a given revision.

sunlin7:
It's function level testing, should pass regardless the external packages.

smile13241324 (approved):

sunlin7:
Thank you !

Issue #17028, opened 2025-06-07
Cursor location difficult to see on a selection in light themes


duianto:
As initially reported here: 
Light theme, selection, cursor location, underscore color https://github.com/nashamri/spacemacs-theme/issues/138

It's difficult to see the cursor location (beg or end) on a selection in a light theme, because the cursor color is to close to the selection color.

![Image](https://github.com/user-attachments/assets/823704bf-b7a0-4266-b182-69774d6a2516)

It's unrelated to the spacemacs-theme package, because it also happens in other light themes.

leuven

![Image](https://github.com/user-attachments/assets/ef95bdcc-ae02-43d8-859e-c28228ac5351)

modus-operandi-deoteranopia

![Image](https://github.com/user-attachments/assets/6bc13223-c7e4-418b-9c03-1889bc5d4234)

The issue seems to be caused by the visual cursor being set to gray, hbar.

https://github.com/syl20bnr/spacemacs/blob/bd42259faf5ca573fe005ffa7054ad059027bde0/layers/%2Bdistributions/spacemacs-bootstrap/config.el#L94

smile13241324:
I have merged a fix from @sunlin7 @duianto, can you have another look?

duianto:
Confirmed fixed, thanks

<img width="90" height="173" alt="Image" src="https://github.com/user-attachments/assets/6fbf07ca-365c-4144-84fb-ff444931fcb6" />

PR #17031, opened 2025-06-10
[bootstrap] Change cursor color in evil visual mode

sunlin7:
I'm not a expert of color-theme, but try my best to enhance #17028.


bcc32:
I think making it blue is unfortunately sometimes worse, depending on the theme (e.g., leuven uses blue for the selected region).  It feels like we should not use a hardcoded color here, and instead pick something programmatically that will sufficiently contrast with the `region` face.  Something like `secondary-selection` might work---themes should set that to a color that is distinct from `region`.

sunlin7:
@bcc32  The change is only apply to the Spacemacs default themes; for the other themes, user should follow the document of `spacemacs-evil-cursors` to change on their local. 
```emacs-lisp
(defvar spacemacs-evil-cursors '()
  "Colors assigned to evil states with cursor definitions.
To add your own, use `spacemacs/add-evil-curosr'.")
```

smile13241324 (approved):

Issue #17036, opened 2025-06-16
LSP is not resolving Python dependencies in Poetry managed virtual environment


jonathanunderwood:
I am using spacemacs with LSP enabled for Python projects using pyright - relevant part of .spacemacs is below.

```emacs-lisp
   dotspacemacs-configuration-layers
   '(
     ;; ----------------------------------------------------------------
     ;; Example of useful layers you may want to use right away.
     ;; Uncomment some layer names and press `SPC f e R' (Vim style) or
     ;; `M-m f e R' (Emacs style) to install them.
     ;; ----------------------------------------------------------------
     auto-completion
     better-defaults
     emacs-lisp
     git
     ;; helm
     ivy
     lsp
     markdown
     multiple-cursors
     ;; org
     ;; (shell :variables
     ;;        shell-default-height 30
     ;;        shell-default-position 'bottom)
     spell-checking
     syntax-checking
     ;; version-control
     treemacs
     ;; Add the rust layer primarily for TOML support
     rust
     (shell :variables
            shell-default-height 30
            shell-default-position 'bottom)
     (python :variables
             python-backend 'lsp
             ;; Choose LSP server - for now we favour pylsp (python-lsp-server)
             ;; over pyright as, although slower, is more feature complete.
             ;; python-lsp-server 'pylsp
             python-lsp-server 'pyright
             ;; Assumption is we've installed the LSP server in the virtual
             ;; environment we are using. So, we need the virtual environment
             ;; activated to find the LSP server within the virtual environment.
             python-pipenv-activate t
             python-poetry-activate t
             python-formatter 'ruff
             python-test-runner 'pytest
             python-format-on-save t
             python-sort-imports-on-save t
             python-indent-def-block-scale 1)
     yaml
     groovy
     javascript
     html
     csv
     docker
     semantic-web
     sql
     (theming :variables
              theming-headings-same-size 'all))
```

I am seeing that lsp-mode is reporting in emacs that 

The `lsp-log` buffer shows this:
```
Command "semgrep lsp" is not present on the path.
Command "ty server" is not present on the path.
Command "pylsp" is not present on the path.
Command "pyls" is not present on the path.
Command "/Users/redacted/Library/Caches/pypoetry/virtualenvs/scan-etl-r6JE7vQ_-py3.12/bin/pyright-langserver --stdio" is present on the path.
Command "semgrep lsp" is not present on the path.
Command "ty server" is not present on the path.
Command "pylsp" is not present on the path.
[...]

Issue #17035, opened 2025-06-11
Python pytest runner cannot find virtualenv configuration


rocha:
Currently, [`pytest`](https://pytest.org) support is provided by the package [`pytest-el`](https://github.com/ionrock/pytest-el) . This package is not supported by [`pet-mode`](https://github.com/wyuenho/emacs-pet), which was added to spacemacs as of #16855. As is, `pytest-el` cannot find the correct virtualenv configuration for python projects that use tools such as [uv](https://github.com/astral-sh/uv) and others.

To fix the problem, we can use `pet-mode` to correctly configure `pytest-el`, or replace `pytest-el` with a newer package such as [`python-pytest`](https://github.com/wbolster/emacs-python-pytest) which is supported by `pet-mode`.

rocha:
@smile13241324 this may interest you

smile13241324:
Should be fixed, please double check

rocha:
`pytest` works now when `pet-mode` is enabled or not using `python-virtualenv-management`. Thanks @smile13241324 and @sunlin7 !

But now `dap`/`debugpy,` cannot find libraries installed in the project's `venv` when `pet-mode` is enabled. I will investigate and make a new issue if I find the problem is not on my end.

wyuenho:
pet 3.4.0 should support pytest.el now, but yes, I would highly suggest replacing a lot of these old packages with newer ones.

smile13241324:
@rocha how does it look like now?

smile13241324:
I have migrated the python layer to python-pytest and also switched to the non-forked version of pet.

Issue #17032, opened 2025-06-10
fd process spam


phenix-factory:
Since this change https://github.com/syl20bnr/spacemacs/commit/9c5b523890c77c31a448c7a569d3b10fbe362dc0 my machine is spamming `fd` process at 100% CPU  each time I open a Python file.
I moved to the previous commit, and the issue disappeared. 

I guess those removed conditions where useful?

sebhahn:
In my case there is no `fd` process spaming, but my python shell interpreter is all messed up and set to `nil`, previous commit 980939c still works fine

bcc32:
@sunlin7  could you please take a look?

sunlin7:
@phenix-factory @sebhahn 
The removed lines were detect the `sqlite3` and `dasel` commands.
Does these two commands exists or not in your cases? 
I'm try to reproduce the issue on my local, will update soon.

sunlin7:
How about turn on the `pet-mode` manually on the previous commit https://github.com/syl20bnr/spacemacs/commit/980939cd20a1cdb5d5d818b98797d37907dfa158?

The change https://github.com/syl20bnr/spacemacs/commit/9c5b523890c77c31a448c7a569d3b10fbe362dc0 will turn on `pet-mode` without detecting the`sqlite3` and `dasel`, that imply on your cases the `pet-mode` were not be activated before.

sunlin7:
I did NOT reproduced the issue with/without the `sqlite3` or `dasel` on Ubuntu 22.04.

Please provide details for the `sqlite3`, `dasel`, `fd`... python project type, emacs version to clarify the reproducing steps.

phenix-factory:
Hello,
You can find my spacemacs config here : https://gitlab.com/phenix-factory/config-spacemacs

sebhahn:
activating pet-mode changes the python interpreter from `python3` to `ipython`, which still seem to work for 980939c, but not afterwards

phenix-factory:
I don't know why pet-mode is important, but I lived without before so…

sunlin7:
@phenix-factory  Your [config-spacemacs](https://gitlab.com/phenix-factory/config-spacemacs) don't have the `python` layer, so **NO** `pet-mode` been installed. 
I downloaded your configurations and started Spacemacs with it, and tried to open a ~/test.py, I got prompt
`Support for python-mode requires installation of layer python, do you want to install it? (y or n)`
[...]

PR #17038, opened 2025-06-16
[python] Recall the pyvenv for migrating to pet smoothly

sunlin7:
Previous #16867 try replace `pyvenv` with `pet-mode` directly, but there're several issues:
1. [searching multiple python interpreters by breadth first](https://github.com/wyuenho/emacs-pet/issues/60)
2. performance issue: [pet SLOW on a large project directory wyuenho/emacs-pet#63](https://github.com/wyuenho/emacs-pet/issues/63)
3. [fd process spam](https://github.com/syl20bnr/spacemacs/issues/17032)
4. Not support manually set python interpreter: https://github.com/syl20bnr/spacemacs/pull/16867#issuecomment-2917237140
5. [Python pytest runner cannot find virtualenv configuration](https://github.com/syl20bnr/spacemacs/issues/17035)
6. [LSP is not resolving Python dependencies in Poetry managed virtual environment](https://github.com/syl20bnr/spacemacs/issues/17036)

This submit will recall the `pyvenv` and control selecting pet/pyvenv by variable `python-virtualenv-management` to try migrate from `pyvenv` to `pet` smoothly. 

The "smooth" migration plan proposal:
1. Support manually set the python interpreter (.dir-locals.el) first, it can be the obvious way to work when the rest of "auto" detection failed;
2. migrate the pyvenv to pet for detecting the `.venv` environment
3. migrate the python interpreter detection to pet, and remove pyvenv

### DIFF
--- a/layers/+lang/python/README.org
+++ b/layers/+lang/python/README.org
@@ -29,8 +29,11 @@
   - [[#fill-column][Fill column]]
   - [[#sort-imports][Sort imports]]
   - [[#importmagic][Importmagic]]
-  - [[#pyenv-pipenv-and-poetry][Pyenv, pipenv and poetry]]
+  - [[#pyvenv-pyenv-pipenv-and-poetry][Pyvenv, pyenv, pipenv and poetry]]
 - [[#management-of-python-versions-and-virtual-environments][Management of Python versions and virtual environments]]
+  - [[#manage-virtual-environments-with-pet-mode][Manage virtual environments with pet-mode]]
+  - [[#manage-virtual-environments-with-pyvenv-legacy-inactively-maintained][Manage virtual environments with pyvenv (legacy, inactively maintained)]]
+    - [[#automatic-activation-of-local-virtual-environment][Automatic activation of local virtual environment]]
   - [[#manage-multiple-python-versions-with-pyenv][Manage multiple Python versions with pyenv]]
     - [[#automatic-activation-of-local-pyenv-version][Automatic activation of local pyenv version]]
   - [[#manage-environments-and-packages-with-pipenv][Manage environments and packages with pipenv]]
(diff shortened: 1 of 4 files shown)


jonathanunderwood:
Something else I noticed a lot lately is that when I open a file in a Poetry based project the first time, the lsp mode will find the LSP in the poetry managed virtualenv. When I then open the next file in the same project, I get prompted to install the relevant lsp, even though it is present in the virtualenv. I mention that in case it's relevant to this PR.

jonathanunderwood:
OK, I have checkout out this branch and updated the packges accordingly, and #17036 is no longer happening, and, so far, opening multiple files in the same project is now using the correct pyright LSP. So, this looks really good. Note that I haven't modeified my .spacemacs file, and so I haven't added `python-virtualenv-management 'pet`.

jonathanunderwood:
I have just now tried switching to using pet, and on opening the second file in a project, I am back to seeing this:

image


I tried multiple restarts both with and without these two options:

```
             python-pipenv-activate t
             python-poetry-activate t
```

and in all cases, the pyright LSP in the poetry virtualenv isn't found. So, I think the pet functionality isn't working, at least for poetry projects.

jonathanunderwood:
@smile13241324 would be great if you could look at this, since the pet mode integration was your hard work.

stkain:
I fetched this pull request and tested it locally on my mac. lsp server is now started again without find-subprocess.
Navigation in python works. Thumbs up from me

smile13241324 (approved):

smile13241324:
Merged, thanks for fixing @sunlin7 I think we have to invest some more time into pet before it can become the default

sunlin7:
Thank you so much. And I had initialized the test cases for python layer #17030, I will collect the cases for python-layer, then make sure migrating `pyvenv` to `pet-mode` can cover the key cases.

wyuenho:
I've just released pet 3.3.0. It should now respect whatever you've set `python-shell-interpreter` to, so if it's set to `ipython`, it'll search for `ipython`. It'll now search for the LSP server binaries from your virtualenvs first as well. I don't know why you'll install LSP servers into your virtualenvs as VSCode only ever gives you the one that comes with the extension, but apparently these users do exist, so I'm supporting it...

As to performance, there are settings tunable by the users, please read the updated README to find out how.

All this also means, you can delete the pet fork from layer now.

wyuenho:
The latest 4.0.0 should have much improved performance due to more aggressive caching, fd spam should be eliminated, and it should also prefer the LSP servers from your virtualenv.

PR #17039, opened 2025-06-17
Improve contrast of evil-state mode line indicator in light themes

bcc32:
In some light themes such as modus-operandi, the normal state
indicator can be completely illegible because the mode line background
color is too similar to the hardcoded color for that state, as chosen
by Spacemacs.

Instead of always using the mode line background color as the
foreground for the state indicator, pick one of the following colors
based on the contrast ratio (as calculated by modus-themes-contrast
which is built-in to Emacs):

1. Mode line foreground
1. Mode line background
1. Black (fallback)
1. White (fallback)

### DIFF
--- a/layers/+distributions/spacemacs-bootstrap/funcs.el
+++ b/layers/+distributions/spacemacs-bootstrap/funcs.el
@@ -75,12 +75,24 @@ For evil states that also need an entry to `spacemacs-evil-cursors' use
   (set-face-attribute (spacemacs/state-color-face (intern state)) nil
                       :foreground (face-attribute 'mode-line :background)))
 
+
+
+(defun spacemacs//pick-contrasting-fg-color-from-mode-line (bg-color)
+  (let* ((ml-fg (face-attribute 'mode-line :foreground))
+         (ml-bg (face-attribute 'mode-line :background))
+         (candidates (mapcar (lambda (fg) (cons (modus-themes-contrast bg-color fg) fg))
+                             (list ml-fg ml-bg "#ffffff" "#000000"))))
+    ;; Pick the first candidate with a reasonably acceptable contrast ratio; if
+    ;; none, just pick the best one.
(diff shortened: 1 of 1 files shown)


bcc32:
Some screenshots for example.

Before:
image

After:
image

smile13241324 (approved):
This is pretty cool :)

saviola777:
> (as calculated by modus-themes-contrast
which is built-in to Emacs)

Is it? Maybe something is wrong with my config, but since today I'm getting the error `Error running timer: (void-function modus-themes-contrast)`, and checking the list of installed packages, `modus-themes` is not on it, and `modus-themes-contrast` is indeed not defined (GNU Emacs 30.1 (build 1, x86_64-pc-linux-gnu, X toolkit, cairo version 1.18.2, Xaw3d scroll bars)). Manually installing `modus-themes` fixes the problem.

smile13241324:
@bcc32 can you have a look?

fnussbaum:
Perhaps something like `(require-theme 'modus-themes)` is necessary?

bcc32:
`modus-themes` has been distributed with Emacs since 28.1, according to https://protesilaos.com/emacs/modus-themes#h:1af85373-7f81-4c35-af25-afcef490c111.  And on my box I saw that `modus-themes-contrast` was an autoloaded function.

Maybe that is a distribution-specific thing.  It would not be hard to just copy the functions to Spacemacs (which would avoid having to load the theme package for just this function)

Not at my computer presently, sorry this broke for you.

saviola777:
Don't be sorry, thanks for all the great work you and all the other contributors/maintainers are doing!

It was an easy fix, just thought I'd write here in case other people may be affected as well. You may be right about it being distribution-specific, I'm on NixOS.

bcc32:
I realized that my patch was actually broken at my site, but my personal config happens to load `modus-vivendi` at startup so I never saw the bug.  Fixed in 34a72874ced7296ac563d3e2bf21d620602b80db.

PR #17048, opened 2025-06-25
[emacs/amx] Migrate smex layer to amx layer

### LINKED ISSUE #16813: Switch from smex to amx
Spacemacs's smex and ivy layers configure package smex for persisting
M-x history among other things.  However, smex has not been updated in
a long time (10 years) and amx is an actively developed fork of it.

Amx supports at least migrating the `smex-save-file` automatically to
`amx-save-file`, but function calls would need to be updated.

sunlin7:
Fix #16813 with migrating the smex to amx.


bcc32:
The `smex` package is also used by the `ivy` layer, not just the `smex` layer.  (that's the only reason that enabling ivy will disable the `smex` layer).  I think we should probably just migrate both rather than only one?

smile13241324 (approved):

smile13241324:
Looks like the ivy layer has also been migrated

smile13241324:
I have found some smaller places in the ivy layer still and fixed them after the merge. Now smex should be gone.

sunlin7:
Thank you! The `amx` was folked from `smex`, I tried replace the smex package with amx package, everything works. After that, I migrate the smex to amx, and verified on local.

PR #17051, opened 2025-06-26
Give occur mode the same bindings as wgrep.

mijoharas:
I use `wgrep` fairly often, and always forget the keybindings now that I sometimes do multi replace in occur buffers. Give the same keybindings.

### DIFF
--- a/layers/+spacemacs/spacemacs-editing/packages.el
+++ b/layers/+spacemacs/spacemacs-editing/packages.el
@@ -36,6 +36,7 @@
     link-hint
     lorem-ipsum
     (origami :toggle (eq 'origami dotspacemacs-folding-method))
+    (occur-mode :location built-in)
     password-generator
     (persistent-scratch :toggle dotspacemacs-scratch-buffer-persistent)
     pcre2el
@@ -671,3 +672,12 @@ See variable `undo-fu-session-directory'." dir))
   (evil-define-key 'normal wgrep-mode-map ",r" #'wgrep-toggle-readonly-area)
   (evil-define-key 'normal wgrep-mode-map ",d" #'wgrep-mark-deletion)
   (evil-define-key 'normal wgrep-mode-map ",f" #'next-error-follow-minor-mode))
+
+;; copy wgrep
+;; this already exists as `i', but I keep looking for things similar to wgrep
+(defun spacemacs-editing/init-occur-mode ()
+  (spacemacs/set-leader-keys-for-major-mode 'occur-mode
+    "w" 'occur-edit-mode)
+  (spacemacs/set-leader-keys-for-major-mode 'occur-edit-mode
+    "," 'occur-cease-edit)
+  )
--- a/layers/+spacemacs/spacemacs-editing/packages.el
+++ b/layers/+spacemacs/spacemacs-editing/packages.el
@@ -36,7 +36,7 @@
     link-hint
     lorem-ipsum
     (origami :toggle (eq 'origami dotspacemacs-folding-method))
-    (occur-mode :location built-in)
+    (replace :location built-in)
     password-generator
     (persistent-scratch :toggle dotspacemacs-scratch-buffer-persistent)
     pcre2el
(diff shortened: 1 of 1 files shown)


sunlin7:
On vanilla emacs `occur-edit-mode` is bound to `e` (for "edit"), which is easy to understand than `w` (which implies "write"?)

mijoharas:
True, I'd be fine with making `wgrep` do the same, and then me relearning muscle memory, but having both be different makes me always forget it (and my `wgrep` muscle memory is stronger.)

Essentially I just want consistency. Also, extra bindings that are mode specific are nice and discoverable.

(in `vim` mode in spacemacs `i` does the `occur-edit-mode`, but doesn't for `wgrep`)

bcc32:
Can we just do the `i` key binding in that case?  It's a little unfortunate to take over navigation key bindings like `w` and `e` when `i` is available and does the right thing for another package.  (As another example, `i` also activates wdired but `w` and `e` have their usual vim meanings in dired).

mijoharas:
Hey, @bcc32 I want to double check I understand things properly, so I'm gonna reply inline.

> Can we just do the `i` key binding in that case?

To be clear, are you talking about adding the `i` keybinding to `wgrep`? because the `i` binding already does the right thing in `occur-mode`.

> It's a little unfortunate to take over navigation key bindings like `w` and `e` when `i` is available and does the right thing for another package.

This is the bit I'm not fully clear on. So, this PR only adds the leader key for major mode (i.e. `,w` for me, comma followed by w). Are you talking about what this is introducing? or something else? (also, those leader keybindings have already been added to `wgrep`)

 > (As another example, `i` also activates wdired but `w` and `e` have their usual vim meanings in dired).
 
 Yeah, it probably makes sense to add `i` to `wgrep` so that it's consistent with the other two. It still doesn't do much for discoverability (which the leader key give because you can use whichkey to find what's offered. In fact, I'd be in favour of adding a leader key to switch to `wdired` mode from `dired` mode personally).

mijoharas:
@bcc32 - or sorry, are you suggesting it should be `,i` for the key binding? should that be changed in `wgrep` as well? (I'm fine with whatever, I just want it to be consistent between both, and available via a leader key so I don't need to remember anything and can find it.)

layers/+spacemacs/spacemacs-editing/packages.el
```diff
@@ -671,3 +672,12 @@ See variable `undo-fu-session-directory'." dir))
   (evil-define-key 'normal wgrep-mode-map ",r" #'wgrep-toggle-readonly-area)
   (evil-define-key 'normal wgrep-mode-map ",d" #'wgrep-mark-deletion)
   (evil-define-key 'normal wgrep-mode-map ",f" #'next-error-follow-minor-mode))
+
+;; copy wgrep
+;; this already exists as `i', but I keep looking for things similar to wgrep
```
bcc32: I think these comments are useful context for the PR but should not be added to the code itself as they will confuse future readers.

layers/+spacemacs/spacemacs-editing/packages.el
```diff
@@ -671,3 +672,12 @@ See variable `undo-fu-session-directory'." dir))
...
+(defun spacemacs-editing/init-occur-mode ()
+  (spacemacs/set-leader-keys-for-major-mode 'occur-mode
+    "w" 'occur-edit-mode)
+  (spacemacs/set-leader-keys-for-major-mode 'occur-edit-mode
+    "," 'occur-cease-edit)
```
bcc32, line 680: Please adjust formatting (no closing paren on its own line) and also add a similar key binding for `,c`.

layers/+spacemacs/spacemacs-editing/packages.el
```diff
@@ -36,6 +36,7 @@
     link-hint
     lorem-ipsum
     (origami :toggle (eq 'origami dotspacemacs-folding-method))
+    (occur-mode :location built-in)
```
bcc32: FYI occur-mode is not the name of an actual package or file.  Perhaps this should be `replace` instead (occur-mode is defined in `replace.el`)

bcc32 (changes requested):

layers/+spacemacs/spacemacs-editing/packages.el
mijoharas: Nice catch, done in https://github.com/syl20bnr/spacemacs/pull/17051/commits/5d762830a5d2388290b167d7a2400d435a6f25fd

layers/+spacemacs/spacemacs-editing/packages.el
mijoharas: done in https://github.com/syl20bnr/spacemacs/pull/17051/commits/5d762830a5d2388290b167d7a2400d435a6f25fd

layers/+spacemacs/spacemacs-editing/packages.el
mijoharas: good point, should have just put them in the PR description. (removed in https://github.com/syl20bnr/spacemacs/pull/17051/commits/5d762830a5d2388290b167d7a2400d435a6f25fd )

bcc32:
Thank you for the contribution!

PR #17052, opened 2025-06-26
Support orderless annotation and kwd matchers.

mijoharas:
Orderless has a few new matchers that are quite useful. The kwd matcher provides a lot of different ones based on different modes, and the annotation matcher with `&` allows you to match against all marginalia annotations.

To enable this we need to change the component separators back to the default (which is a space not preceded by a backslash).

This could be left as a configuration option, but I figured the nicest solution was to use the default. People can override if they prefer the other behaviour.

It seems to have supported this for a bit over a year now.

Reference to orderless-kwd: https://github.com/oantolin/orderless/blob/master/orderless-kwd.el

and the base annotation matcher with `&`: https://github.com/oantolin/orderless?tab=readme-ov-file#style-dispatchers (`orderless-annotation`)

### DIFF
--- a/layers/+completion/compleseus/packages.el
+++ b/layers/+completion/compleseus/packages.el
@@ -347,15 +347,16 @@
         (let ((completion-styles '(basic partial-completion orderless)))
           (apply orig-fun args))))
 
-    (setq orderless-component-separator "[ &]")
-
     ;; should be all in with orderless otherwise the results are inconsistent.
     ;; the available styles are registered in `completion-styles-alist`.
     (setq completion-styles '(orderless basic)
           completion-category-defaults nil
           ;; we need to have 'basic here first in order to support tramp connections...
           ;; see `completion-styles`.
-          completion-category-overrides '((file (styles basic partial-completion))))))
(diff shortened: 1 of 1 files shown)


fnussbaum:
Thanks, these matchers are indeed useful. I would prefer to keep the `&` separator for in-buffer completion though, where a space cannot be used. I pushed a suggested change to that end, what do you think? 

(As mentioned in the commit message, it might be possible to replace the advice by defining a separate completion
style for in-buffer completion, see `orderless-define-completion-style` and the discussion in https://github.com/minad/consult/pull/1131.)

mijoharas:
Ahhhh! That's what it's for! For company. Seems good. I'll try and test this tomorrow.

smile13241324:
@mijoharas can you clarify what's the state of that PR? Is it worth checking again?

fnussbaum:
@smile13241324 I'll revisit this and merge it later this week, perhaps with a minor revision. Sorry for the delay!

fnussbaum (approved):

PR #17059, opened 2025-07-01
[core] Shift owner to dotfile if system package is off

### LINKED ISSUE #17010: Layer config should not be applied if a package is :toggled off but present in dotspacemacs-additional-packages
Thanks to #16700, packages listed in `dotspacemacs-additional-packages` are installed even if some enabled layer has a `:toggle` form for that package which evals to nil.  However, it seems that in this case, Spacemacs attempts to apply the layer configuration for that package, which can lead to errors (and seems like undesired beahvior anyway).

Consider a config which is like the default, except that `dotspacemacs-additional-packages` contains `evil-swap-keys` because the user wants to use it directly.   `dotspacemacs-swap-number-row` is nil, as is the default.

However, because `spacemacs-editing` configures `evil-swap-keys`, `spacemacs-editing/init-evil-swap-keys` runs at startup.  Since `spacemacs-editing/init-evil-swap-keys` (reasonably) assumes `dotspacemacs-swap-number-row` is non-nil, you get this error:

```
evil-swap-keys-swap-number-row: Wrong type argument: listp, "dotspacemacs-swap-number-row nil is not supported."
```

sunlin7:
Fix #17010.

In the function `configuration-layer/make-packages-from-dotfile`,
pkg --> short for user defined package in dotfile
obj --> short for the package object already created in the Spacemacs

pkg:on or obj:on means the package toggle is on;
pkg:off or obj:off means the package toggle is off;

| pkg\obj | obj:off | obj:on |
|--------|--------|--------|
| pkg:off | off | on |
| pkg:on | on, shift owner | on |

As the table, shift the owner for obj off but pkg on case.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -1200,14 +1200,25 @@ USEDP if non-nil indicates that made packages are used packages."
   (dolist (pkg (append dotspacemacs-additional-packages
                        dotspacemacs--additional-theme-packages))
     (let* ((pkg-name (if (listp pkg) (car pkg) pkg))
-           (obj (configuration-layer/get-package pkg-name)))
+           (obj (configuration-layer/get-package pkg-name))
+           pkg-toggle
+           obj-toggle)
       (if (null obj)
           (setq obj (configuration-layer/make-package pkg 'dotfile))
+        (setq pkg-toggle (if (and (listp pkg) (memq :toggle pkg))
+                             (plist-get pkg :toggle)
+                           t)       ; user-defined package is enabled by default
+              obj-toggle (cfgl-package-toggled-p obj t))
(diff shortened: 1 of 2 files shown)


smile13241324 (changes requested):
@sunlin7 please have a look at the comments from copilot, apart of these this looks good.

sunlin7:
@smile13241324  Had committed the AI spelling suggestions, Thank you !

smile13241324:
@sunlin7 strange I am not seeing anything new yet can you double check

sunlin7:
@smile13241324  I had manually merge the suggested changes, combine the changes into one commit, and force pushed to this PR. 
So there still has one commit log.

smile13241324 (approved):

PR #17060, opened 2025-07-02
[core] Fix warning for private dir not exists

sunlin7:
Spacemacs give an error message on startup buffer (Spacemacs home buffer):
> Warnings: (w)
    - Layer directory "~/.emacs.d/private/" not found. Ignoring it.

It assume the `spacemacs-private-directory` is located as `~/.emacs.d/private` but on actually the Spacemacs was cloned into  `/usr/share/spacemacs.git/` then the spacemacs init.el will be load from the `~/.emacs.d/init.el`.

This PR will detect the directory to avoid the warning on Spacemacs startup buffer.


bcc32:
Thanks

Issue #16884, opened 2025-03-15
Can't use spacemacs-installed packages in vanilla Emacs


bcc32:
When debugging Spacemacs, it is often useful to run `emacs -q` and then load only the packages whose commands are being debugged, so that it can be determined whether some Spacemacs customization is causing the issue, or whether the issue exists in the upstream package (or Emacs itself).

However, due to the fact that `hybrid-mode` uses autoloaded defcustom variables (autoloading defcustoms is not necessary and sometimes harmful) and evaluating these variables' default values requires macros defined in Spacemacs itself, it is not possible to directly use the packages installed by Spacemacs with:

```
(setq package-user-dir "~/.emacs.d/elpa/30.1/develop/")
(package-initialize)
```

or similar (hybrid-mode-autoloads fails to load, and probably other issues I haven't discovered yet).

bcc32:
When digging into this, I noticed that in b282e553ba3e7cea8902908a857fd9a2b542e348, the macro `spacemacs|dotspacemacs-backward-compatibility` was added to replace some usage of variable aliases.  I think this was likely a mistake, and we should revert the change (perhaps using `define-obsolete-variable-alias` rather than just `defvaralias`).  Variable aliases are self-documenting through `SPC h d v`, and obsolete aliases actively encourage the user to migrate their configuration.  Also, a macro for a simple `boundp` check seems overkill.

bcc32:
As an aside, it would be nice in principle if `hybrid-mode` and `holy-mode` could be made into packages independent from Spacemacs.  Their behavior depends very little on Spacemacs except for some configuration that could be factored out better.

fnussbaum:
>  the macro spacemacs|dotspacemacs-backward-compatibility was added to replace some usage of variable aliases. I think this was likely a mistake, and we should revert the change (perhaps using define-obsolete-variable-alias rather than just defvaralias)

I agree, let's use `define-obsolete-variable-alias` instead.

> it is often useful to run emacs -q and then load only the packages whose commands are being debugged

There is also `spacemacs/restart-stock-emacs-with-packages`, but it does not work for me because it does not set `package-user-dir` or `load-path`. This should be fixed, and I think it would be nice to have a similar function that uses `restart-emacs-start-new-emacs` instead of quitting Spacemacs.

bcc32:
[...]

PR #17062, opened 2025-07-04
[core] Phase out the spacemacs|dotspacemacs-backward-compatibility

sunlin7:
Phase out the `spacemacs|dotspacemacs-backward-compatibility`.

The ` Phase out the spacemacs|dotspacemacs-backward-compatibility` was compiled into the `hybrid-mode-autoloads.el`, so we can NOT delete it directly, change it to warn.

### DIFF
--- a/core/core-funcs.el
+++ b/core/core-funcs.el
@@ -26,6 +26,8 @@
 (defmacro spacemacs|dotspacemacs-backward-compatibility (variable default)
   "Return `if' sexp for backward compatibility with old dotspacemacs
 values."
+  (declare (obsolete nil "The `dotspacemacs-backward-compatibility' will be removed after 2025.
+Please reinstall the package which relies on this macro."))
   `(if (boundp ',variable) ,variable ',default))
 
 (defun spacemacs/system-is-mac ()
--- a/core/core-funcs.el
+++ b/core/core-funcs.el
@@ -26,8 +26,13 @@
 (defmacro spacemacs|dotspacemacs-backward-compatibility (variable default)
   "Return `if' sexp for backward compatibility with old dotspacemacs
 values."
-  (declare (obsolete nil "The `dotspacemacs-backward-compatibility' will be removed after 2025.
-Please reinstall the package which relies on this macro."))
+  (declare (obsolete nil "The `spacemacs|dotspacemacs-backward-compatibility' macro will be removed after 2025.
+Please reinstall the package which relies on this macro (such as `hybrid-mode')"))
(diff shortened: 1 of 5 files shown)


sunlin7:
Fix #16884 partially.

core/core-funcs.el
```diff
@@ -26,7 +26,8 @@
 (defmacro spacemacs|dotspacemacs-backward-compatibility (variable default)
   "Return `if' sexp for backward compatibility with old dotspacemacs
 values."
-  `(if (boundp ',variable) ,variable ',default))
+  (warn "The `dotspacemacs-backward-compatibility' will be removed after 2025.
+Please reinstall the package which relies on this macro."))
```
bcc32: I think directly replacing the real macro definition will break any interpreted code which relies on the macro, right?  Furthermore, this warning won't get triggered by any compiled code, which already expanded the macro.

bcc32 (changes requested):
I agree with the direction of this change, but I'm not sure this method of warning about the outdated macro works as you intended.

sunlin7:
@bcc32 You're absolutely right! I had pushed the quoted line, please help review again. Thank you!

bcc32:
(Sorry for the delay, I have been dealing with stupid NYC housing nonsense for the last two months)

The `warn` doesn't trigger any kind of warning when some code which calls this macro is compiled, only if some _interpreted_ code calls the macro (or when the compiled code is later executed).  However, such code will break immediately upon updating past this PR, because `(warn)` does not return a useful Lisp form.

So IMO the warning is not helpful, because the user's config will be broken before they have a chance to react to the warning.  It doesn't seem like much of an improvement over just deleting the macro outright.

We should either:

1. Just delete the macro immediately, if we think it is unliklely that any user configs or packages depend on it.  (I don't see anything on GitHub, although that is not definitive).
2. Add a warning to the macro, and also mark it obsolete using `(declare (obsolete ...))`, **but keep its current return value**.

sunlin7:
@smile13241324  Thanks for the comment, and I add the `obsolete` in the macro and keep original return value, as my initial comment message, the already-installed package file `hybrid-mode-autoloads.el` depends on this macro, if we remove it directly, the user will be blocked  on startup.

bcc32:
Ah, thanks, I missed that comment.  I have added the warning back so that users who have `hybrid-mode` installed will see the warning on startup (since autoloads are interpreted and not compiled).

Please take another look and see if the current version of the PR looks good to you.

bcc32 (approved):

sunlin7:
@bcc32 It's good for me, thank you!

PR #17071, opened 2025-07-26
Markdown packages.el: navigation shortcut

stradicat:
Added `p` as `'markdown-toc-follow-link-at-point` shortcut (a particular pain point of mine as Emacs is pretty much my documentation IDE), since the `p` key is both intuitive for _"point"_ and currently unused for the mode.


smile13241324 (changes requested):
Please add the new key binding to the readme as well. The rest looks good.

PR #17072, opened 2025-07-27
adapt to variable name change in new treemacs

dankessler:
in a recent commit
Alexander-Miller/treemacs@acd69f4d19b3aa7de2168e47661148d221853b81

treemacs underwent some reorganization and `treemacs--buffer-name-prefix` was renamed to `treemacs-buffer-name-prefix`. spacemacs relies on the this variable in its treemacs layer, so adjust the logic to use either depending on which is bound (so that things will work both for people who have and haven't recently updated their treemacs package)


jaeyeom:
:+1:

Ghasak:
Any new update on this please!

phenix-factory:
I copied to code and spacemacs works again, thank you for the fix

fnussbaum (approved):
Thank you!

PR #17075, opened 2025-07-29
Update README.org

mstergianis:
Fixes typo


fnussbaum (changes requested):
Looks like this should be `.golangci.yml`?

fnussbaum (approved):

PR #17078, opened 2025-08-01
restructuredtext: migrate from auto-complete to company

miyabinomuratic:
This layer is the only user of the auto-complete package. Let's migrate to company mode to make it align with other layers. We can also remove the workaround for flyspell as company can work together with flyspell.


bcc32:
Thank you!

PR #17084, opened 2025-08-13
Update funcs.el of EXWM Layer: Remove void functions `exwm-randr-enable` and `exwm-systemtray-enable`

khaalidrafi:
In funcs.el of EXWM layer, `(exwm-randr-enable)` and `(exwm-systemtray-enable)` functions don't work and give errors at startup complaining their function definitions are void. I changed them to `(exwm-randr-mode 1)` and `(exwm-systemtray-mode 1)` respectively according to EXWM Wiki.


fnussbaum (approved):
Thank you!

PR #17085, opened 2025-08-14
Fix keybindings and error of EXWM layer

khaalidrafi:
The `exwm-input-set-key` can no longer set a global keybinding unless invoked interactively. The keybindings are redefined in this commit with `exwm-input-global-keys`. `exwm-terminal-command` is no longer available and hence it's keybinding is removed. Global keybinding `C-q` to send key has been removed because it's unnecessary. Why would someone use it in char-mode? `s-N` has been remapped to switch to N-th window instead of workspace. The code to remove EXWM prefix bindings has been reduced to one line according to the EXWM wiki. The deprecated `exwm-enable` is replaced with `exwm-wm-mode`. The redundant function `exwm-init` has been removed since it's already called inside `exwm-wm-mode`. It was causing assertion failure.

### DIFF
--- a/layers/+window-management/exwm/packages.el
+++ b/layers/+window-management/exwm/packages.el
@@ -81,7 +81,26 @@
     (use-dialog-box nil "Disable dialog boxes since they are unusable in EXWM")
     (exwm-input-line-mode-passthrough t "Pass all keypresses to emacs in line mode.")
     :init
-    (exwm-enable)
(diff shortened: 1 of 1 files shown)


smile13241324 (approved):

smile13241324:
Looks good, from a first review this looks good, a detailed manual test should follow after it has been merged.

PR #17092, opened 2025-08-27
[core] Fix package :toggle flag does not support nil

sunlin7:
The package `:toggle` flag does not support `nil` directly.

Currently the `:toggle` require a variable or an expression, but sometime want set it to nil directly for torubleshooting.
For example, replace the `python-enable-importmagic` to `nil` directly in follow package.el line:
https://github.com/syl20bnr/spacemacs/blob/2bfa3555c9e63e3363d929d6f080d0056bc80b09/layers/%2Blang/python/packages.el#L40

As the example showed, directly setting a `nil` to `:toggle` in layer's package file will take no effect.

This PR will support the `nil` for the package `:toggle' flag. 

Please help review the change. Thanks


bcc32:
I added a small suggestion, to make the `toggle` variable just be `'unspecified` if missing, rather than checking `(memq :toggle pkg-spec)` in multiple places.

bcc32 (approved):

bcc32:
fyi my workaround for this issue has been to add  `:excluded t` to a package rather than `:toggle nil`, but of course it is better to fix this surprising bug if we can :)

sunlin7:
Thanks for then enhancement, I can confirm it works on my local, and more elegant.  Thank you!

PR #17099, opened 2025-08-31
[helm] Replace helm-git-grep package with helm builtin function helm-grep-git-1

sunlin7:
The [helm-git-grep](https://github.com/yasuyk/helm-git-grep) was unmaintained 8 years,
and it was also removed from melpa https://github.com/melpa/melpa/pull/9520,
Remove the `helm-git-grep` package from Spacemacs, and replace the functions with helm builtin function `helm-grep-git-1` (helm won't directly support the desired function, but suggested to use the `helm-grep-git-1`, refer https://github.com/emacs-helm/helm/pull/2734)

### DIFF
--- a/layers/+completion/helm/README.org
+++ b/layers/+completion/helm/README.org
@@ -20,6 +20,7 @@
   - [[#transient-state][Transient state]]
   - [[#files][Files]]
   - [[#buffers][Buffers]]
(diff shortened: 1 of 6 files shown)


smile13241324 (approved):
Looks good

PR #17105, opened 2025-09-05
Add lexical-binding cookie in ido-vertical-mode.el

liuchong:
Add lexical-binding cookie in ido-vertical-mode.el to suppress warning


bcc32:
This file is automatically copied from the upstream source by the bot, your changes will be overwritten.  Please make this report upstream at https://github.com/creichert/ido-vertical-mode.el instead.

PR #17120, opened 2025-09-15
[helm] Remove deprecated helm-posframe package

sunlin7:
According to the https://github.com/melpa/melpa/pull/9520, helm-posframe was removed from melpa, so it need receipt now.

### DIFF
--- a/layers/+completion/helm/config.el
+++ b/layers/+completion/helm/config.el
@@ -40,8 +40,6 @@
 (define-obsolete-variable-alias
   'dotspacemacs-helm-position 'helm-position "20180608")
 
-(defvar helm-use-posframe nil
-  "Use helm-posframe to display completions in a separate frame")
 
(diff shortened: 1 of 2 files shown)


bcc32:
Should we just remove support for this package instead?  The linked pull request says helm-posframe is unmaintained.

sunlin7:
@bcc32  The `helm-posframe` can popup a child frame in **TUI**, there is NO alternative package can do that. 
The helm has built-in feature `helm-display-buffer-in-own-frame` only works on GUI.

sunlin7:
@bcc32  I pushed the change to remove `helm-posframe`, and uses who want this feature should use `helm-posframe` as external package. Please review again. Thank you !

bcc32:
That sounds like a reasonable alternative to me.  Thanks!

PR #17122, opened 2025-09-19
Fix spacemacs-modeline/init-spaceline

Bost:
The 94b3cda51 calls `(use-package spaceline ...)` leading to:

■  Error (use-package): spaceline/:config: Symbol’s function definition is void: spaceline-spacemacs-theme

Also the `#'spaceline-helm-mode` and `#'spaceline-info-mode` are not defined.

### DIFF
--- a/layers/+spacemacs/spacemacs-modeline/packages.el
+++ b/layers/+spacemacs/spacemacs-modeline/packages.el
@@ -48,7 +48,7 @@
     (setq-default fancy-battery-show-percentage t)))
 
 (defun spacemacs-modeline/init-spaceline ()
-  (use-package spaceline
+  (use-package spaceline-config
     :commands (spaceline-compile spaceline-define-segment)
     :init
     (add-hook 'spacemacs-post-user-config-hook


sunlin7:
As discussing in https://github.com/syl20bnr/spacemacs/pull/16970, the `spaceline-config` is **NOT** a package name, it's part of `spaceline` package. 
For using the functions in `spaceline-config.el`, you can require the feature explicitly, or using autoload features.

Bost:
Be it what it may, `(use-package spaceline ...)` doesn't work, while `(use-package spaceline-config ...)` does - and that's what I **really** care about.

sunlin7:
Please share your configuration for reproducing, thanks

fnussbaum:
I think fixing this by adding autoloads would be unnecessarily verbose and messy due to the non-standard structure of spaceline package. We should either add `(require 'spaceline-config)` at the end of the use-package init block, or indeed change it back to `(use-package spaceline-config ...)`. I prefer the latter, as it is more concise. Moreover, according to the use-package manual the `foo` in  `(use-package foo)` does not have to be a package name, it can also be a feature that is part of a package.

fnussbaum (approved):

PR #17127, opened 2025-09-22
[llm-client] Key binding for gptel-rewrite

sunlin7:
Add key binding for `gptel-rewrite`, without it, we have to press `SPC $ g m` to get the gptel menu, then press `r` to call the rewrite function.

### DIFF
--- a/layers/+web-services/llm-client/README.org
+++ b/layers/+web-services/llm-client/README.org
@@ -93,16 +93,17 @@ The layer provides several key bindings to interact with LLMs efficiently.
 
 ** GPTel
 
-| Key binding | Command                  | Description                    |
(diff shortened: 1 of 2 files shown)


smile13241324 (changes requested):
I think you need to add the new keybinding to the readme still.

Btw, I have added predefined personas into a copilot_instructions.md file. In there is a dedicated persona for writing documents.

Just fire up the ai agent of your choice read in the file and ask the "documentation writer" to update the readme according to the layer and you should be good to go.

sunlin7:
@smile13241324 Key binding is added into the readme file now, please help review again.
And will try the copilot feature, thank you so much!

smile13241324 (approved):

PR #17128, opened 2025-09-22
Remove obsoleted variables from layers/ and core/

sunlin7:
Remove the obsoleted variables which were obsoleted more than 5 years.


bcc32:
Sadly, Spacemacs doesn't actually warn users if a variable they are setting in their .spacemacs file is obsolete (unless they happen to byte-compile their .spacemacs, which is not recommended).  Perhaps we should add such a mechanism first?

sunlin7:
A mechanism to notify obsoleted variable will be user friendly. 
There are other obsoleted variables, like `ocaml-format-before-save`, `c++-enable-organize-includes-on-save` less than 5 years, then we can try a way to notify user the variable are obsoleted.

Current  changes include the variables obsoleted more thank 7 years ago, should be safe to remove them.

bcc32:
> Current changes include the variables obsoleted more thank 7 years ago, should be safe to remove them.

It doesn't really matter how long ago the variables were obsoleted, if there has never been a mechanism to alert users that the variables were obsolete.  I have dozens of users at my site who have .spacemacs configurations dating from back when those were the official names of the variables, and they were obviously never prompted to update.

IMO there is no urgency to remove these aliases, I think we should at least ty tro see if there's a good notification mechanism, before potentially breaking behavior.

sunlin7:
Got your point. I don have a good idea to notify user a variable obsoleted at current time. :thinking:

smile13241324:
I totally agree with @bcc32 we first need to warn about the depreciation.

Couldn't we use the home buffer for that? There is a message buffer reserved for changes we could add breaking changes in there as well as a date when they will get into effect.

What do you both think? @bcc32  @sunlin7

sunlin7:
As @smile13241324 mentioned, I add new `core-obsolete.el` for recording/detecting the obsolete variables. 
If a user has obsolete variable, Spacemacs startup buffer will give a warning message.
Please review again. Thank you!

core/core-obsolete.el
```diff
@@ -0,0 +1,55 @@
...
+    (dotspacemacs-helm-use-fuzzy helm-use-fuzzy "20180608")
+    (dotspacemacs-helm-resize helm-enable-auto-resize "20180608")))
+
+(defun spacemacs/check-obsolete-variables ()
+  "Check the obsolated vairables and return message list for the obsolated vairable."
+  (flatten-tree
+   (mapcar (lambda (item)
+             (when (boundp (car item))
+               (apply #'format "%s is obsolete as %s, use %s instead." item)))
+           spacemacs-obsolete-variable-list)))
```
smile13241324: Thanks for this patch. I've reviewed the function and made one adjustment.

You were absolutely right to spot that the mapcar would create nil entries. The intent to filter them was correct, but flatten-tree isn't the right tool for that specific job (it's for nested lists).

I've replaced it with seq-remove #'null, which achieves the filtering you intended. I also took the liberty of revising the docstring for clarity.

The rest of the logic looks solid. Great catch!

```suggestion
(defun spacemacs/check-obsolete-variables ()
  "Check for defined obsolete variables and return a list of warning messages."
  ;; Filter out all 'nil' values generated by the mapcar.
  ;; `seq-remove` is preferred over the older `cl-remove-if #'null`.
  (seq-remove #'null
    (mapcar (lambda (item)
              (when (boundp (car item))
                (apply #'format "%s is obsolete as %s, use %s instead." item)))
            spacemacs-obsolete-variable-list)))
```

smile13241324 (changes requested):
Thank you for your work on this PR! This is a valuable contribution, and I appreciate the time you've invested.

I've reviewed the code in detail and left a few comments with change requests. These are mostly minor points to get everything perfectly aligned with our coding guidelines and the overall architecture.

Please take a look when you have a moment, and just let me know if you have any questions about the notes or if you'd like to walk through them together. As soon as those points are addressed, I'll be happy to get this merged!

sunlin7:
@smile13241324  Had changed the code to follow your comment, and tested on my local, confirmed the new changes works great on my local. 
Please help review the changes again. Thank you!

core/core-obsolete.el
```diff
@@ -0,0 +1,56 @@
...
+    (dotspacemacs-helm-use-fuzzy helm-use-fuzzy "20180608")
+    (dotspacemacs-helm-resize helm-enable-auto-resize "20180608")))
+(defun spacemacs/check-obsolete-variables ()
+  "Check for defined obsolete variables and return a list of warning messages."
+  ;; Filter out all 'nil' values generated by the mapcar.
+  ;; `seq-remove` is preferred over the older `cl-remove-if #'null`.
+  (seq-remove #'null
+    (mapcar (lambda (item)
+              (when (boundp (car item))
+                (apply #'format "%s is obsolete as %s, use %s instead." item)))
```
smile13241324: You have an issue with the order of files the message should be

X is obsolete since 2025-01-01, use Y instead.

Right now I am getting
X is obsolete as Y, use 20250101 instead.

smile13241324 (changes requested):
I have found a small bug please have a look

core/core-obsolete.el
bcc32: BTW I think this should be made a private function (use `//` naming convention).  Also, I'd expect `(delq nil (mapcar ...))` to be faster and more concise.

sunlin7: Updated the code to correct the message (X is obsolated since DATE, use Y instead), and changed the function name with double "//". Please help review again. Thank you!

smile13241324 (approved):
This looks great @sunlin7  thanks for all the hardwork I am going to merge it in a minute.

PR #17129, opened 2025-09-22
[llm-client] Apply config after gptel-context loads for better performance

sunlin7:
As the title, minor changes for performance.

### DIFF
--- a/layers/+web-services/llm-client/packages.el
+++ b/layers/+web-services/llm-client/packages.el
@@ -45,9 +45,8 @@
     :ensure t
     :init
     ;; evilify gptel-context-buffer-mode-map
-    (require 'gptel-context)
-    (evil-set-initial-state 'gptel-context-buffer-mode 'evilified)
     (evilified-state-evilify-map gptel-context-buffer-mode-map
+      :eval-after-load gptel-context
       :mode gptel-context-buffer-mode
       :bindings
       "C-c C-c" #'gptel-context-confirm


layers/+web-services/llm-client/packages.el
```diff
@@ -44,18 +44,18 @@
...
-      :bindings
-      "C-c C-c" #'gptel-context-confirm
-      "C-c C-k" #'gptel-context-quit
-      "RET"     #'gptel-context-visit
-      "n"       #'gptel-context-next
-      "p"       #'gptel-context-previous
-      "d"       #'gptel-context-flag-deletion)
+    (with-eval-after-load 'gptel-context
+      ;; evilify gptel-context-buffer-mode-map
+      (evil-set-initial-state 'gptel-context-buffer-mode 'evilified)
```
bcc32: I think `evil-set-initial-state` works fine before the mode has been defined.  But, FWIW, `evilified-state-evilify-map` also already sets up the initial state so I think this line can just be deleted.

layers/+web-services/llm-client/packages.el
```diff
@@ -44,18 +44,18 @@
...
+      (evilified-state-evilify-map gptel-context-buffer-mode-map
```
bcc32: `evilified-state-evilify-map` has an `:eval-after-load` argument which can be used to do this a little more concisely, if you'd like.

sunlin7:
@bcc32  Thanks for the comment, and you're right, rewrite with the `:eval-after-load`, and removed the evilified line. 

bcc32 (approved):

bcc32:
LGTM, thanks!

Issue #17080, opened 2025-08-07
Merge #17065 breaks `configuration-layer/package-used-p`


thanhvg:
This merge https://github.com/syl20bnr/spacemacs/pull/17065 enhances `configuration-layer/package-used-p`. However in doing so the loading order in `configuration-layer//load` has been changed. This makes  `configuration-layer/package-used-p` useless during the startup and can lead to unexpected regressions.

One example is `SPC b 1..9` `Move buffer to window 1..9` is now throwing void function `spacemacs/move-buffer-to-window`.

It's a chicken and egg problem.`configuration-layer/package-used-p` is used mostly in `funcs.el` and `packages.el`. So it is very important to have its data source `configuration-layer--indexed-packages` initialized asap for its consumers. This was ensured by calling `(configuration-layer//declare-used-packages configuration-layer--used-layers)` close to the top of `configuration-layer//loadconfiguration-layer//load`.

But the merge pushed the initialization `(configuration-layer//declare-used-packages configuration-layer--used-layers)` bellow the calls to load layers. This makes `configuration-layer/package-used-p` to return `nil` on *any package* while layers are being loaded.

The intention in the merge is to load `config.el` as well as el files `funcs.el`, `packages.el`...  so `toggle` slot can be evaluated. But with this order it won't be useful for `configuration-layer/package-used-p` consumers.

sunlin7:
Thanks for analysis, and yes, it likes a circle dependency issue. Any solution for #16346 but without regressions?

thanhvg:
I don't know. I'm just a happy Spacemacs user with my own fork expecting you guys the maintainers to carry out all the hard work :)

I would consider Spacemacs core stable, it's been battle tested for more than 10 years or less. So any change to it should be carefully considered and justified. I don't see much of benefit in fixing this `configuration-layer/package-used-p` for the sake of consistency.

However, imo, the use of `configuration-layer/package-used-p` should not be encouraged. It is an escape hatch you reach to only when there is no other solution. I think it is overused in current code base.

`pre-init`,  `post-init` and the built in `eval-after-load` should cover most of the use cases of `configuration-layer/package-used-p`.

On the other hand  your `configuration-layer/package-used-p` is perfect once Spacemacs startup process has been done, so using it after that is fine.

bcc32:
[...]

PR #17131, opened 2025-09-22
[defaults] Fix void-function move-buffer-to-window

### LINKED ISSUE #17130: Symbol’s function definition is void: spacemacs/move-buffer-to-window
After 80a9502702557fc9770e7e8eb0616258eef8479b, packages are not declared until after config.el and funcs.el, which means those files cannot use `configuration-layer/package-used-p` reliably (although their defuns maybe can, depending on when the functions are actually invoked).

This in particular causes `spacemacs/move-buffer-to-window` not to be defined at all, because even if package `winum` is used, `(configuration-layer/package-used-p)` is called too early.

## Reproduction
Just press `SPC b 2` at startup.  It errors:

```
Symbol’s function definition is void: spacemacs/move-buffer-to-window
```

bcc32:
- Don't try to avoid defining various functions at top level if some
  package doesn't exist---it's fine for them to just error.

- Replace soft message with a user-error.

- Do not check whether the winum package is configured by Spacemacs.
  It is sufficient for the winum feature to be provided by a site-lisp
  package.

- Fix up docstrings slightly, according to checkdoc.

Fix #17130.

See also #17080.

### DIFF
--- a/layers/+spacemacs/spacemacs-defaults/funcs.el
+++ b/layers/+spacemacs/spacemacs-defaults/funcs.el
@@ -356,46 +356,47 @@ Dedicated (locked) windows are left untouched."
   (interactive "p")
   (spacemacs/rotate-windows-forward (* -1 count)))
 
-(if (configuration-layer/package-used-p 'winum)
-    (progn
-      (defun spacemacs/move-buffer-to-window (windownum follow-focus-p)
-        "Moves a buffer to a window, using the spacemacs numbering. follow-focus-p
-  controls whether focus moves to new window (with buffer), or stays on current"
-        (interactive)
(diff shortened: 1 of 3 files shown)


bcc32:
Pushed another commit to also fix a couple other places where `configuration-layer/package-used-p` was being used at top-level in `funcs.el`.

The configuration layer is very complicated and stuff like this is why my branch for fixing #16346 got stalled :)

fnussbaum (approved):

fnussbaum:
Thanks! I have pushed two minor related changes.

bcc32:
Ah, good catch on the documentation especially :)

PR #17133, opened 2025-09-24
[python] Package pip-requirements follow the python-enable-tools flag

sunlin7:
The package `pip-requirements` should follow the `python-enable-tools` flag to toggle on or off.


bcc32:
I think `pip-requirements` just provides a major mode editing functionality for the `requirements.txt` file, and doesn't do, like, automatic environment management that might conflict with other tools, right?  I don't think this change is necessary.

sunlin7:
Okay, close this PR.

PR #17140, opened 2025-09-30
fix: set next-error-last-buffer when calling pytest

kassick:
python-pytest does not use `compilation-start` (which would have taken care of that for us) and neither sets the variable manually.

This kludge works around that by advising python-pytest--get-buffer. This function is called from a python buffer when running pytest -- so it looks like the easiest way to intercept the pytest-buffer.

### DIFF
--- a/layers/+lang/python/funcs.el
+++ b/layers/+lang/python/funcs.el
@@ -278,6 +278,15 @@ location of \".venv\" file, then relative to pyvenv-workon-home()."
 ;; Tests
 
 ;; python-pytest adapters
+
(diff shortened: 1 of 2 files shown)


kassick:
Without this, `next-error`/`previous-error`, `spacemacs/switch-to-compilation-buffer`, `spacemacs/show-hide-compilation-window` etc. do not work, as these functions need the buffer-local value of `next-error-last-buffer`

kassick:
Fixes https://github.com/syl20bnr/spacemacs/issues/17141

smile13241324 (approved):

PR #17149, opened 2025-10-14
[spacemacs-theme] Make diff-refine faces legible, use bg for diff faces

bcc32:
In particular, use background colors for both
diff-{added,changed,removed} and diff-refine-{added,changed,removed}. Changing the former allows diff-font-lock-syntax to function correctly and leave diffed code with proper syntax highlighting.

Change the `diff-refine-*` faces to use less harsh background colors (those actually intended for background use are named `*-bg*`), so that refined text is actually legible.

## before

image

## after

image

### DIFF
--- a/core/libs/spacemacs-theme/spacemacs-theme.el
+++ b/core/libs/spacemacs-theme/spacemacs-theme.el
@@ -314,17 +314,17 @@ to `auto', tags may not be properly aligned. "
      `(corfu-current ((,class (:background ,ttip-sl :foreground ,base))))
 
 ;;;;; diff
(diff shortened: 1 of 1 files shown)


fnussbaum:
Looks good; I think this needs to be submitted [upstream](https://github.com/nashamri/spacemacs-theme), though.

bcc32:
Yes, thank you for the reminder.  Posted at https://github.com/nashamri/spacemacs-theme/pull/212, and closing here.

It would be nice to have a CI check on new PRs to ensure they do not modify bot-updated files.

PR #17152, opened 2025-10-16
[ci] Warn PR submitters if they are modifying bot-maintained files

bcc32:
As I recently made this exact mistake in #17149, I feel it would be
nice if the CI reminded contributors (both long-time and newcomers)
that PRs making changes to any bot-maintained files will just see
their changes erased by the bot.


fnussbaum (approved):
Thanks!

PR #17154, opened 2025-10-22
[bootstrap] Use evil-undo in visual state

fnussbaum:
To be consistent with normal undo. All undo backends support undo in region, but the support in undo-fu seems buggy.

### DIFF
--- a/layers/+distributions/spacemacs-bootstrap/packages.el
+++ b/layers/+distributions/spacemacs-bootstrap/packages.el
@@ -155,7 +155,10 @@
     (advice-add 'evil-refresh-cursor :around #'spacemacs/not-in-pdf-view-mode))
 
   (when vim-style-enable-undo-region
-    (define-key evil-visual-state-map (kbd "u") 'undo))
+    (define-key evil-visual-state-map (kbd "u")
+                (if (eq dotspacemacs-undo-system 'undo-fu)
+                    'undo
+                  'evil-undo)))
 
   (evil-ex-define-cmd "enew" 'spacemacs/new-empty-buffer)
 


layers/+distributions/spacemacs-bootstrap/packages.el
```diff
@@ -155,7 +155,10 @@
     (advice-add 'evil-refresh-cursor :around #'spacemacs/not-in-pdf-view-mode))
 
   (when vim-style-enable-undo-region
-    (define-key evil-visual-state-map (kbd "u") 'undo))
+    (define-key evil-visual-state-map (kbd "u")
+                (if (eq dotspacemacs-undo-system 'undo-fu)
+                    'undo
+                  'evil-undo)))
```
bcc32, line 161: Would it be correct to just bind this to `evil-undo` always?  I believe that should respect whatever the current `evil-undo-system` is, right?

fnussbaum: Ideally, yes, but at least in my testing the undo-fu commands did not reliably work in visual state. I think we could also just remove undo-fu at some point, it does not seem to have enough benefits.

bcc32: Oh, I misunderstood.  `undo` is being used to specifically avoid going through undo-fu?

Did you have `undo-fu-allow-undo-in-region` non-nil?

> I think we could also just remove undo-fu at some point, it does not seem to have enough benefits.

I agree.  I use undo-fu personally but I don't think I really take advantage of it, it mostly just gets in the way sometimes.

fnussbaum: > Oh, I misunderstood. undo is being used to specifically avoid going through undo-fu?

Yes, exactly, to avoid breaking the functionality for undo-fu users with this change.

> Did you have undo-fu-allow-undo-in-region non-nil?

Yes, Spacemacs sets this, and I don't have it customized. Still, it seems buggy or at least counterintuitive: Repeated `evil-undo` (with undo-fu) starting in visual state can undo/redo things outside of the initial region. I can try to come up with a reproducible example, if you prefer.

bcc32: No, I've observed the thing you're talking about.  Seems reasonable to me.

bcc32 (approved):

PR #17155, opened 2025-10-22
[core] Set undo-redo as default undo system

fnussbaum:
The built-in undo/redo system is more stable and provides all the functionality nowadays that undo-fu does. There is no need to reimplement this. In particular, the undo in region feature works perfectly with undo-redo, while it seems to have issues with undo-fu.


winsphinx:
After this commit, I see the package `undo-fu-session` is still remained. 
Which one depends on it or requires it?

fnussbaum:
> Which one depends on it or requires it?

As mentioned in #17157, `undo-fu-session` also works with the built-in undo system, and disabling it could make users lose their history.

PR #17157, opened 2025-10-23
Disable undo-fu-session unless use undo-fu

winsphinx:
After commit #17155 ,also disable undo-fu-session unless undo-fu is in use.


bcc32:
undo-fu-session [doesn't require undo-fu](https://codeberg.org/ideasman42/emacs-undo-fu-session#motivation) to be used as the undo system, to work correctly.  It is maintained by the same developer but works on `undo-redo`'s history too.

bcc32:
I'm going to close this, since I think we shouldn't do it.

PR #17161, opened 2025-10-26
[navigation] Add cycling functionality to SPC TAB and SPC w TAB

fnussbaum:
Sometimes one finds oneself in the situation where one used `SPC TAB` to switch to the last buffer, but it turns out that one actually intended to switch to the second or third to last buffer. By enabling the option added in this commit, one can consecutively press `TAB` in such situations to cycle through the previous buffers. From the point of view of subsequent commands it is as though one switched directly to the destination buffer. This uses the excellent transient-cycles package under the hood (https://elpa.gnu.org/packages/transient-cycles.html).

The setup needs lexical binding in `spacemacs-navigation/packages.el`. I have audited the file and it does not look as if anything depends on dynamic binding.

I considered (and implemented a first version) using hydra/Spacemacs transient states, by the way, but the implementation using the transient-cycles package is cleaner.

The current solution uses somewhat hacky advice to lazily load the package, not sure if this is controversial. Alternatives could be to use remapping instead of advice, or load the package at startup, or perhaps defer it for a few seconds (using `:defer 3`, for example).

### DIFF
--- a/core/core-dotspacemacs.el
+++ b/core/core-dotspacemacs.el
@@ -483,6 +483,36 @@ window."
   'boolean
   'spacemacs-dotspacemacs-init)
 
+(spacemacs|defc dotspacemacs-enable-cycling nil
+  "Make consecutive tab key presses after
+`spacemacs/alternate-buffer' (SPC TAB) or
+`spacemacs/alternate-window' (SPC w TAB) cycle through previous buffers
+or windows. After arriving at the destination buffer/window, from the
+point of view of consecutive commands, it is as if the destination was
+directly switched to. By default, the backspace key cycles in the
+opposite direction.
+
+You can customize the cycling keys with the options
+`spacemacs-default-cycle-forwards-key',
+`spacemacs-default-cycle-backwards-key', or with the command-specific
+variants `spacemacs-alternate-buffer-cycle-forwards-key',
+`spacemacs-alternate-buffer-cycle-backwards-key'
+`spacemacs-alternate-window-cycle-forwards-key', and
+`spacemacs-alternate-window-cycle-backwards-key',
+
+Moreover, you can set the option `transient-cycles-show-cycling-keys' to
+nil to suppress the message specifying the cycling keys in each invocation.
+
+Note that this feature requires Emacs 29 or later.
+
+Set the option to t in order to enable cycling for all current and
+future cycling commands. Alternatively, choose a subset of the currently
+supported commands: '(alternate-buffer alternate-window). (default nil)"
+  '(choice (const t)
+           (repeat (choice (const alternate-buffer)
+                           (const alternate-window))))
+  'spacemacs-dotspacemacs-init)
+
 (spacemacs|defc dotspacemacs-maximize-window-keep-side-windows t
   "Whether side windows (such as those created by treemacs or neotree)
 are kept or minimized by `spacemacs/toggle-maximize-window' (SPC w m)."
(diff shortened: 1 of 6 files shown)


core/core-dotspacemacs.el
```diff
@@ -483,6 +483,35 @@ window."
...
+
+Moreover, you can set the option `transient-cycles-show-cycling-keys' to
+nil to suppress the message specifying the cycling keys in each invocation.
+
+Note that this feature requires Emacs 29 or later.
+
+Set the option to '(alternate-buffer alternate-window) in order to
+enable cycling for all currently supported commands; or choose a
+subset. (default nil)
+  '(repeat (choice (const alternate-buffer)
```
bcc32: IMO, it should also be allowed to set this to `t` to enable all current and future cycling commands.  (Will need to adjust the callsites of course to handle when this variable isn't a list)

layers/+spacemacs/spacemacs-navigation/packages.el
```diff
@@ -433,6 +434,88 @@
...
+                                spacemacs-default-cycle-backwards-key)
+       :cycle-forwards-key (or spacemacs-alternate-buffer-cycle-forwards-key
+                               spacemacs-default-cycle-forwards-key)))
+    (when (member 'alternate-window dotspacemacs-enable-cycling)
+      (transient-cycles-define-commands
+       (sorted-windows index)
+       (([remap spacemacs/alternate-window] ()
+         (interactive)
+         (setq sorted-windows
+               (sort (window-list)
```
bcc32, line 478: I think `sort` with keyword arguments is only usable from Emacs 30+ onwards, sadly

bcc32 (changes requested):

bcc32:
> The current solution uses somewhat hacky advice to lazily load the package, not sure if this is controversial. Alternatives could be to use remapping instead of advice, or load the package at startup, or perhaps defer it for a few seconds (using `:defer 3`, for example).

IMO just loading the package at startup is probably fine.  On my box, it takes 4-5ms to load (timed with `(benchmark-run 100 (load "transient-cycles"))`)

BTW I have pushed a couple of small tweaks, I think they should be uncontroversial but if you disagree please lmk.

core/core-dotspacemacs.el
fnussbaum: Good idea, done.

layers/+spacemacs/spacemacs-navigation/packages.el
fnussbaum: Right, thanks!

fnussbaum:
> IMO just loading the package at startup is probably fine. On my box, it takes 4-5ms to load (timed with `(benchmark-run 100 (load "transient-cycles"))`)

Makes sense, done.

> BTW I have pushed a couple of small tweaks

Perfect, thank you! I've addressed the feedback, fixed the package's toggle form, and added it to the layer's readme.

fnussbaum:
I have added another commit to prevent the tab-line from showing an inconsistent state after exiting transient cycling.

smile13241324 (approved):
This is excellent. Thanks so much for this contribution. The refactoring in core-funcs.el is clean, and the new dotspacemacs-enable-cycling variable is perfectly documented.

I noticed your PR description mentions "somewhat hacky advice," but the implementation I see in the diff (using [remap spacemacs/alternate-buffer]) is the perfectly clean, declarative use-package solution. That's exactly the right way to do it, so I'm assuming the description is just outdated.

Great work! Merging this now.

PR #17164, opened 2025-10-31
[ci] Upgrade Emacs 29.3 to 29.4

bcc32:
Emacs 29.4 contains security fixes not present in Emacs 29.3.  The actual motivation is to get a newer Windows build of Emacs 29 so that it contains a new-enough ELPA signing key.  This, in turn, fixes the Python layer functional tests because otherwise `compat` could not be installed from GNU ELPA.


bcc32:
Well, sadly, that didn't fix the issue like I thought it would, but I think this is a good idea to do anyway.

PR #17165, opened 2025-10-31
[ci] Fix python layer functional tests on windows, Emacs 29

bcc32:
compat was unable to be installed due to a failed signature
verification when downloading the archive contents from GNU ELPA.

I attempted to import the updated key directly using gpg but ran into
other issues on macOS with GitHub Actions:
https://github.com/actions/runner-images/issues/9777.  So, instead
this just disables signature verification for the package archives in
CI.


bcc32:
Alternatively, one could imagine just disabling the Windows+Emacs 29 CI combination.  I don't have a strong opinion on which we should do.

fnussbaum (approved):

PR #17173, opened 2025-11-11
update layer +lang/emacs-lisp use-package to make :config section for…

bflick21:
layer/+lang/emacs-lisp interferes with adding gptel llm client layer unless this is added for me.

Emacs 30.2 Windows 11, installed with winget.

… late bound variables or whatever the term is, causes the code in the config section for the emacs-lisp package to need to go there instead of init. Didn't encounter this until trying to load the gptel llm-client interface.

### DIFF
--- a/layers/+lang/emacs-lisp/packages.el
+++ b/layers/+lang/emacs-lisp/packages.el
@@ -103,6 +103,7 @@
       (spacemacs/set-leader-keys-for-major-mode mode
         "df" 'spacemacs/edebug-instrument-defun-on
         "dF" 'spacemacs/edebug-instrument-defun-off))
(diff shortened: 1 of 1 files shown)


bcc32:
Can you paste the actual error you got, here, or describe what the actual symptoms were?

smile13241324:
Honestly I do not see how deferring key bindings has any effect on another package @bflick21 please provide a stacktrace or log output or a step by step reproduction guide otherwise there is not much we can do here.

Delaying binding menu labels and bindings is nothing I would agree to until there is a good reason for

smile13241324:
Closed due to too long inactivity

PR #17177, opened 2025-11-14
[core] Install packages in order of declaration

fnussbaum:
In particular, do not install packages in alphabetical order. This means one can manually manage dependencies for packages installed via quelpa, addressing https://github.com/syl20bnr/spacemacs/discussions/17176. This change should be safe, also the order of package installations was approximately reversed before without obvious regressions (see this PR's first commit).

Note that we still use the alphabetically sorted list of packages in other contexts, such as when configuring packages, mostly for backward compatibility.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -1877,7 +1877,7 @@ RNAME is the name symbol of another existing layer."
             (when install-deps
               (setq result (append install-deps result))))
           (when (funcall filter pkg-name)
-            (cl-pushnew pkg-name result))))
+            (setq result (append result (list pkg-name))))))
       (delete-dups result))))
 
 (defun configuration-layer//filter-packages-with-deps
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -390,7 +390,10 @@ file. It can be overridden by users inside `dotspacemacs/user-init'.")
   "Hash map to index `cfgl-layer' objects by their names.")
 
 (defvar configuration-layer--used-packages '()
-  "An alphabetically sorted list of used package names.")
+  "A list of used package names.")
+
+(defvar configuration-layer--used-packages-sorted '()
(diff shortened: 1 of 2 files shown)


fnussbaum:
Installing packages in (reverse) alphabetic order is arbitrary, inflexible and can lead to problems as evidenced by #17176. 

Consider also the openai layer:
https://github.com/syl20bnr/spacemacs/blob/ccd3e200c2f88c3eb8b465fc00f34adc4d1d28fd/layers/%2Bweb-services/openai/packages.el#L23-L27

Here, none of the packages are available on GNU/Non-GNU ELPA or MELPA, and `chatgpt`, `codegpt` and `dall-e` all depend on `openai`. Therefore, installation of those packages currently only succeeds because `openai` comes first in reverse alphabetical order.
After this PR, the installation succeeds because `openai` is declared first in `openai-packages`.

I skimmed through all instances of packages declared to get installed via quelpa, and it does not look like this change would break anything (in any case, it would be a simple matter to change the order of package declarations). In most other cases, the installation order should not make a difference.

fnussbaum:
By the way, the reason for why we manually need to manage installation order for dependencies of quelpa-installed packages is that we cannot fetch the dependencies from `package-alist` in `configuration-layer//filter-packages-with-deps-recur`.

Perhaps we could, alternatively, first clone all of the quelpa packages without installing them, and then look up their dependencies to determine the installation order. But this would be a more involved change.

On the other hand, with the change of this PR, we would make the installation order of packages depend on the order of layers in `dotspacemacs-configuration-layers`, but I don't think this would matter in most normal use cases. Nevertheless, it might be better to fix the order of the layers in the context of package installation, for example alphabetically.

fnussbaum:
> Perhaps we could, alternatively, first clone all of the quelpa packages without installing them, and then look up their dependencies to determine the installation order.

This might indeed be better, I'll mark this PR as a draft. Sorry for the noise.

PR #17178, opened 2025-11-14
[core] Set evil-want-keybinding to nil before installing packages

fnussbaum:
Byte-compilation of packages requiring evil somewhere will load evil, for example. This would previously lead to a warning at startup.

### DIFF
--- a/core/core-spacemacs.el
+++ b/core/core-spacemacs.el
@@ -195,9 +195,14 @@ of directories to file basenames."
   (prefer-coding-system 'utf-8)
   ;; Extend use-package if installed.
   (spacemacs/use-package-extend)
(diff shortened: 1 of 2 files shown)


bcc32:
Other things could load evil, for example the user's dotfile could load a local checkout of evil-mode for testing.  I think we should instead just set `evil-want-keybinding` to nil as early as possible (e.g., in `init.el` or `spacemacs/init`), and then we can get rid of the duplicate `setq` from `spacemacs-bootstrap/init-evil` at that point.

fnussbaum:
> I think we should instead just set evil-want-keybinding to nil as early as possible (e.g., in `init.el` or `spacemacs/init`)

Yes, this is even better, done.

bcc32 (approved):

PR #17179, opened 2025-11-15
[core] Improve error message when GNU tar is missing

fnussbaum:
This has led to numerous reported issues in the past. Spacemacs can work without GNU tar, because quelpa installation only fails for some packages; so it should suffice to add the message in that case.

### DIFF
--- a/core/core-configuration-layer.el
+++ b/core/core-configuration-layer.el
@@ -1717,9 +1717,15 @@ RNAME is the name symbol of another existing layer."
             (oset pkg lazy-install nil))
            (t (configuration-layer//warning "Cannot install package %S."
                                             pkg-name)))
-        ('error
+        (error
          (configuration-layer//error
-          (concat "\nAn error occurred while installing %s " "(error: %s)\n")
+          (concat "An error occurred while installing %s " "(error: %s)"
+                  (when (and (equal err '(wrong-type-argument package-desc nil))
+                             (and (listp location) (eq 'recipe (car location)))
+                             (boundp 'quelpa--tar-type)
+                             (not (eq quelpa--tar-type 'gnu)))
+                    ". It looks like GNU tar is not installed, or `quelpa-build-tar-executable' \
+needs to be customized to a GNU tar executable."))
           pkg-name
           err)
          (spacemacs//redisplay))))))


core/core-configuration-layer.el
```diff
@@ -1717,9 +1717,15 @@ RNAME is the name symbol of another existing layer."
...
                                             pkg-name)))
-        ('error
+        (error
          (configuration-layer//error
-          (concat "\nAn error occurred while installing %s " "(error: %s)\n")
+          (concat "An error occurred while installing %s " "(error: %s)"
+                  (when (and (equal err '(wrong-type-argument package-desc nil))
+                             (and (listp location) (eq 'recipe (car location)))
+                             (boundp 'quelpa--tar-type)
+                             (not (eq quelpa--tar-type 'gnu)))
```
fnussbaum, line 1726: This does not cover the case where `(quelpa--tar-type)` evaluates to `gnu` because of the fallback:
```elisp
(defun quelpa--tar-type ()
  "Return `bsd' or `gnu' depending on type of Tar executable.
Tests and sets variable `quelpa--tar-type' if not already set."
  (or quelpa--tar-type
      (when (and quelpa-build-tar-executable
                 (file-executable-p quelpa-build-tar-executable))
        (setq quelpa--tar-type
              (let ((v (shell-command-to-string
                        (format "%s --version" quelpa-build-tar-executable))))
                (cond ((string-match-p "bsdtar" v) 'bsd)
                      ((string-match-p "GNU tar" v) 'gnu)
                      (t 'gnu)))))))
```

Not sure how common this case is. Perhaps we would need to duplicate a part of this function's implementation to cover this case.

bcc32: If I remember correctly, the issues have all been with bsdtar which is installed on MacOS, right?  AFAICT that will actually print `bsdtar` so the fallback should not happen in that scenario at least:

```
$ /usr/bin/tar --version
bsdtar 3.5.3 - libarchive 3.7.4 zlib/1.2.12 liblzma/5.4.3 bz2lib/1.0.8
```

fnussbaum: > If I remember correctly, the issues have all been with bsdtar which is installed on MacOS, right?

This should be the most common case. Though on OpenBSD, apparently the default tar does not support the `--version` flag at all. There was also an issue on Android, but I'm not sure if having a default "non-gnu" tar installed there is common.

PR #17186, opened 2025-11-26
[langtool] enable more setting methods

EasternLi:
+ use executable =languagetool= as default if available
+ change tag uncategorized to checker


smile13241324 (approved):
Looks good

PR #17187, opened 2025-11-29
[sql] Improvements to product selection UI

bcc32:
1. Remove error-prone hook used to cache product list (totally
   unnecessary, the list is very small so we should just read it
   directly and filter when needed).
2. Move commands to funcs.el where they belong, so that
   `find-function` works correctly.
3. Stop depending on `helm` and use `completing-read` to be generic


smile13241324 (approved):
Looks good to me

PR #17196, opened 2025-12-19
[layouts] Fix error in switch-to-prev-buffer when persp-mode is excluded

bcc32:
When the `persp-mode` package is excluded by
`dotspacemacs-excluded-packages`, and the `spacemacs-layouts` layer is
still enabled---as it is by default---the advice around functions
listed in `spacemacs-layouts-restricted-functions` can error, causing
the user to be unable to kill buffers, etc.

This PR just makes the advice check whether `persp-contain-buffer-p`
is actually defined before attempting to potentially call it around
`buffer-list`.

This PR also:

1. Fixes up some outdated docstrings
1. Removes a pointless private macro by just expanding it in the only
   callsite
1. ~Fixes the customization :set function to avoid an unnecessary extra
   variable, and also correctly use `set-default-toplevel-value` to
   avoid issues if the variable is currently let-bound.~ This was not essential,
   and I decided to remove it because it was not clear what to do with it during
   review.

### DIFF
--- a/layers/+spacemacs/spacemacs-layouts/funcs.el
+++ b/layers/+spacemacs/spacemacs-layouts/funcs.el
@@ -93,25 +93,19 @@ Cancels autosave on exiting perspectives mode."
   (let ((ivy-ignore-buffers (remove #'spacemacs//layout-not-contains-buffer-p ivy-ignore-buffers)))
     (ivy-switch-buffer)))
 
-(defmacro spacemacs||with-persp-buffer-list (&rest body)
-  "This one is a brute force version of `with-persp-buffer-list'.
-It maintains the order of the original `buffer-list'"
-  `(cl-letf* ((org-buffer-list
-               (symbol-function 'buffer-list))
-              ((symbol-function 'buffer-list)
-               #'(lambda (&optional frame)
-                   (seq-filter
-                    #'persp-contain-buffer-p
-                    (funcall org-buffer-list frame)))))
-     ,@body))
-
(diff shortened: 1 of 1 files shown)


layers/+spacemacs/spacemacs-layouts/config.el
```diff
@@ -70,28 +70,31 @@ layout, the 4th for the 4th, and so on until the 10th (aka layout
...
-               spacemacs-layouts-restricted-functions value)
-         (dolist (fn spacemacs-layouts-restricted-functions)
+  :set (lambda (var new-value)
+         (ignore-error void-variable
+           (let ((old-value (default-toplevel-value var)))
+             (dolist (fn old-value)
+               (advice-remove fn 'spacemacs-layouts//advice-with-persp-buffer-list))))
+         (set-default-toplevel-value var new-value)
+         (dolist (fn new-value)
            (advice-add fn :around 'spacemacs-layouts//advice-with-persp-buffer-list))))
```
fnussbaum, line 95: When reloading the configuration, the custom set function is run after the value is already set to the `new-value` (if the variable is set as a layer variable in `dotspacemacs-configuration-layers`). Hence in this case, with this change, the advice will not be removed correctly. (I should have added a comment here.)

bcc32: That's pretty confusing, but good catch.  How about instead of trying to handle the case where the variable is set but the :setter isn't run, in the :setter---which seems really hard in the general case---we just make Spacemacs call the setter (if any) for layer variables set by `:variables`?

PR #17198, opened 2025-12-22
Fix gptel advice function to return original function's value

lululau:
Use prog1 to ensure `llm-client/disable-purpose-mode-around-for-gptel` returns the return value of the original gptel function while still properly managing purpose-mode state.

As `llm-client/disable-purpose-mode-around-for-gptel` originally does not return the original return value of `gptel`, it causes a parameter type error `(string-p t)` in the following scenario:

	1. The gptel session buffer has not yet been created
	2. In gptel-menu, set the response to as gptel session
	3. Press Enter to send the gptel request

Cheers!

### DIFF
--- a/layers/+web-services/llm-client/packages.el
+++ b/layers/+web-services/llm-client/packages.el
@@ -86,7 +86,8 @@
     (let ((purpose-mode-was-enabled (bound-and-true-p purpose-mode)))
       (when purpose-mode-was-enabled
         (purpose-mode -1))
-      (apply orig-func args)
-      (when purpose-mode-was-enabled
-        (purpose-mode 1))))
+      (unwind-protect
+          (apply orig-func args)
+        (when purpose-mode-was-enabled
+          (purpose-mode 1)))))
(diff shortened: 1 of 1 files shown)


layers/+web-services/llm-client/packages.el
```diff
@@ -86,7 +86,8 @@
     (let ((purpose-mode-was-enabled (bound-and-true-p purpose-mode)))
       (when purpose-mode-was-enabled
         (purpose-mode -1))
-      (apply orig-func args)
-      (when purpose-mode-was-enabled
-        (purpose-mode 1))))
+      (prog1
+          (apply orig-func args)
+        (when purpose-mode-was-enabled
+          (purpose-mode 1)))))
```
bcc32, line 92: It seems to me that maybe this should use `unwind-protect` instead of `prog1`, in case orig-func errors.  Thoughts?

lululau: Agreed, `unwind-protect` is definitely the better solution. The code has been updated.

bcc32 (approved):

bcc32:
Thanks for your contribution!

PR #17205, opened 2026-01-07
[modeline] Add Spacemacs state colors to vanilla mode line evil-state tag

bcc32:
This PR improves the usability of the vanilla mode line theme by
colorizing the evil-state indicator (see evil-mode-line-format) in the
vanilla mode line format.  It uses the same colors that Spacemacs
would otherwise use with Spaceline (and also the same ones that
are used for the cursor in graphical Emacs, regardless of mode line
theme).

It includes a few preliminary cleanups and improvements as well.


fnussbaum (approved):
Looks good, thanks!

